PR #20290, opened 2023-12-28
checker: cleanup array and alias of array method call

yuyi98:
This PR cleanup array and alias of array method call.


spytheman (approved):
Excellent work.

PR #20294, opened 2023-12-28
Make public map_from in json2

esquerbatua:
module x.json2

It's useful to make public as map function to transform any structure to modify data in the fly and transform the map into json directly


spytheman:
The CI failure is for `vlib/x/json2/json2.v:432:0:pub fn map_from[T](t T) map[string]Any `, since it is now public, it needs to be documented.
Please add a short comment above it, starting with: `// map_from` to fix it.

esquerbatua (commented):
Add docs to map_from

spytheman (approved):

PR #20297, opened 2023-12-28
pref: disable -macosx-version-min clang flag by default

edam:
Disable the `-macosx-version-min` clang flag which is poorly documented, but which seems to affect STL (and some other headers) and also which runtime is linked against.

This flag causes problems (on my Intel-based mac, at least), such as
* use of `C.environ` causes segfault
* `os.Process._spawn()` segfaults

This results in total breakage of some v commands, including `v run` and `v test`. E.g., `v run` segfaults as follows:
```
signal 11: segmentation fault
0   libsystem_platform.dylib            0x00007ff8009cf37d _sigtramp + 29
1   ???                                 0x0000000000000000 0x0 + 0
2   v                                   0x000000010bd3e5db os__Process__spawn + 155
3   v                                   0x000000010bd3e503 os__Process_wait + 35
4   v                                   0x000000010c072aa5 v__builder__Builder_run_compiled_executable_and_exit + 3701
5   v                                   0x000000010c07027c v__builder__compile + 124
6   v                                   0x000000010c07bd19 main__rebuild + 121
7   v                                   0x000000010c07af25 main__main + 2181
8   v                                   0x000000010c080830 main + 64
9   dyld                                0x00007ff8006193a6 start + 1942
```

This PR changes default behaviour of V compiler so that it does not pass a min macosx version to clang by default.

### DIFF
--- a/vlib/v/pref/pref.c.v
+++ b/vlib/v/pref/pref.c.v
@@ -157,7 +157,7 @@ pub mut:
 	use_cache              bool   // when set, use cached modules to speed up subsequent compilations, at the cost of slower initial ones (while the modules are cached)
 	retry_compilation      bool = true // retry the compilation with another C compiler, if tcc fails.
 	use_os_system_to_run   bool   // when set, use os.system() to run the produced executable, instead of os.new_process; works around segfaults on macos, that may happen when xcode is updated
-	macosx_version_min     string = '10.7' // relevant only for macos and ios targets
+	macosx_version_min     string = '0' // relevant only for macos and ios targets
 	// TODO Convert this into a []string
 	cflags  string // Additional options which will be passed to the C compiler *before* other options.
 	ldflags string // Additional options which will be passed to the C compiler *after* everything else.


spytheman:
Please try `v run examples/tetris` with the modified compiler. I think the main reason for the flag, was to allow gg apps to run, if that is no longer needed, that would be great.

edam:
I tried running examples/tetris...

| Run | Result | Output |
|--------|--------|--------|
| With this PR | Blank (white) window | Screenshot 2023-12-30 at 10 11 23 am |
| Without this PR (stock V) | Compiler error | Screenshot 2023-12-30 at 10 30 50 am |
| With this PR and `-macosx-version-min=10.7` | Same compiler error | Screenshot 2023-12-30 at 10 29 03 am |

...baring in mind this is just on my mac, which is:
* 2.4 GHz 8-Core Intel Core i9
* AMD Radeon Pro 5500M 8 GB
* Intel UHD Graphics 630 1536 MB
* 64 GB 2667 MHz DDR4
* Sonoma 14.1.2 (23B92)

The compiler error is interesting, though. Without this PR, doing a `v run` always segfaults when it tries to spawn the subprocess. But for tetris, it doesn't even get that far.

edam:
Actually, all gg examples seem to be broken for me. I can't run 2048, flappylearning-v, or anything in examples/gg. They are all blank (white, or black, it seems) windows.

So I can't run any of them in with any branch or command line arguments at the moment!

spytheman:
Well, at least with your changes, it compiled, which is better than a cgen error, due to the missing library 🤔 .

spytheman:
On m1 with Sonoma 14.0, it still works with the changes, so it seems that it is a net positive:
image

spytheman (approved):
Excellent find.

PR #20299, opened 2023-12-28
compiler: remove .ctags.d (ctags 6.1 now supports V)

edam:
[ctags v6.1.0](https://github.com/universal-ctags/ctags/releases/tag/v6.1.0) was released today and now supports V out of the box. (Release notes refer to it only as "V for merging", due to a ctags dev branching the original "V: new parser" branch before merging!)

This PR removes V's ctags optlib config, in top-level `.ctags.d` dir, for a V parser.


hholst80:
Is this not a bit preemptive? Not everyone runs arch ;)

spytheman:
Thank you @edam . 

@hholst80, without removing the folder, the new universal ctags does:
```sh
#0 10:21:39 ᛋ master /v/oo❱./v -print-v-files examples/hello_world.v | /usr/local/bin/ctags -e -L -
ctags: Language "V" already defined
#1 10:21:40 ᛋ master /v/oo❱
```
Note that it exits with code 1, but without a clear message about what is wrong, which is a problem, if you want to use the newer version with V. It also does not produce any TAGS file in this case, even though the runtime seems proportional to the size of the input files, so it clearly does a lot, before the error message, and exiting with 1.

In my opinion, the native V parser, that the latest ctags has, is a very welcome inclusion, and the value of the improved support for generating much more accurate tags for V sources, that it provides, justifies the deletion of the old hacky regex based patterns, that I added in .ctags.d/v.ctags .

Ctags is easy to compile from source, for people that want to use it right away, and I have no doubt that it will be packaged in the various distros very quickly, if it has not already happened.

The tarball for the new release is here:
https://github.com/universal-ctags/ctags/releases/download/v6.1.0/universal-ctags-6.1.0.tar.gz

spytheman:
@edam can you please also just make a small modification to .gitignore, to ignore `.ctags.d/v.ctags` too, so that in case someone does want to use an older ctags, he/she can do it, by just copying it from an older V release.

spytheman (approved):

Larpon:
@edam thanks a lot for this! Huge developer experience boost in `nvim` and `kate` also :heart:

PR #20305, opened 2023-12-29
ORM: Fix example in readme

xjzh123:
According to the discord group, someone pointed out the example codes in the readme of ORM has `@` and some quotes missing. Ref to [this message](https://discord.com/channels/592103645835821068/631271043876519967/1183823916851019887). So I tried to fix it so that the syntax looks right. As I don't develop with ORM or databases, I haven't tested the code, but I think it is better than before with syntax error.

### DIFF
--- a/vlib/orm/README.md
+++ b/vlib/orm/README.md
@@ -32,16 +32,16 @@ import time
 
 @[table: 'foos']
 struct Foo {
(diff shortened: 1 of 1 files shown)


spytheman (approved):
Good work.

PR #20315, opened 2023-12-31
v: improved comptime var checking with `is` operator and smartcasting

### LINKED ISSUE #20312: checker: sumtype error: $for expects a type name or variable name to be used here, but T is not a type or variable name
### Describe the bug

error: $for expects a type name or variable name to be used here, but T is not a type or variable name

### Reproduction Steps

```v
type TestSum = int | string

struct Abc {
	s TestSum
}

fn gen[T, R](sum T) R {
	$if T is $sumtype {
		$for v in sum.variants {
			if sum is v { // smartcast
				dump(sum)
				return sum
			}
		}
	}
	return R{}
}

fn main() {
	gen[TestSum, string](TestSum('foo'))
}
```

### Expected Behavior

No error

### Current Behavior

```bash
discardable.v:9:13: error: $for expects a type name or variable name to be used here, but T is not a type or variable name
    7 | fn gen[T, R](sum T) R {
    8 |     $if T is $sumtype {
    9 |         $for v in sum.variants {
      |                   ~~~
   10 |             if sum is v { // smartcast
   11 |                 dump(sum)
discardable.v:10:11: error: `is` can only be used with interfaces and sum types
    8 |     $if T is $sumtype {
    9 |         $for v in sum.variants {
   10 |             if sum is v { // smartcast
      |                    ~~
   11 |                 dump(sum)
   12 |                 return sum
discardable.v:10:11: error: cannot use type `int` as type `T`
    8 |     $if T is $sumtype {
    9 |         $for v in sum.variants {
   10 |             if sum is v { // smartcast
      |                    ~~
   11 |                 dump(sum)
   12 |                 return sum
discardable.v:12:12: error: cannot use `TestSum` as type `string` in return argument
   10 |             if sum is v { // smartcast
   11 |                 dump(sum)
   12 |                 return sum
      |                        ~~~
   13 |             }
   14 |         }
discardable.v:10:11: error: cannot use type `string` as type `T`
    8 |     $if T is $sumtype {
    9 |         $for v in sum.variants {
   10 |             if sum is v { // smartcast
      |                    ~~
   11 |                 dump(sum)
   12 |                 return sum
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.3 4690fec

Ubuntu
> [!NOTE]
>

felipensp:
Fix #20312

```v
type TestSum = string | int

fn gen[T, R](sum T) R {
	$if T is $sumtype {
		$for v in sum.variants {
			if sum is v {
				dump(sum)
				$if sum is R {
					return sum
				}
			}
		}
	}
	return R{}
}

fn main() {
	dump(gen[TestSum, string](TestSum('foo')))
}
```

### DIFF
--- a/vlib/v/ast/scope.v
+++ b/vlib/v/ast/scope.v
@@ -129,6 +129,13 @@ pub fn (mut s Scope) update_ct_var_kind(name string, kind ComptimeVarKind) {
 	}
 }
 
+pub fn (mut s Scope) update_smartcasts(name string, typ Type) {
+	mut obj := unsafe { s.objects[name] }
+	if mut obj is Var {
+		obj.smartcasts = [typ]
+	}
+}
+
 // selector_expr:  name.field_name
 pub fn (mut s Scope) register_struct_field(name string, field ScopeStructField) {
 	if f := s.struct_fields[name] {
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -3852,7 +3852,7 @@ fn (mut c Checker) smartcast(mut expr ast.Expr, cur_type ast.Type, to_type_ ast.
 					orig_type = expr.obj.typ
 				}
 				is_inherited = expr.obj.is_inherited
-				ct_type_var = if expr.obj.ct_type_var == .field_var {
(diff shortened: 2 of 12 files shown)


spytheman (approved):
Excellent work.

PR #20321, opened 2023-12-31
orm: fix "not equal" operator

elliotchance:
While most SQL databases support "!=" for not equal, "<>" is the standard operator.


Delta456 (approved):
Excellent work.

spytheman (approved):

PR #20326, opened 2024-01-01
db.pg: fix error with msvc compiler +instructions

ldedev:
Whenever it is compiled on Windows and there are no .dlls in the executable folder, it simply closes with the error code `-1073741515` equivalent to `0xC0000135`.

1. I'm adding to the readme which .dlls are needed.
2. Correcting the flag that is passed to the msvc compiler.
3. Explained where to place libpq if using the msvc compiler.


spytheman (approved):
Excellent. Thank you.

PR #20333, opened 2024-01-01
sokol: Fix build on OpenBSD

jasperla:
It needs this include path in order to find `X11/Xlib.h`.

This was tested on OpenBSD/amd64 with a local X11 display using various programs in `examples/`.


spytheman (approved):
Thank you!

PR #20336, opened 2024-01-01
net: support only ip and ip6 in net.tcp_listener

### LINKED ISSUE #20307: net.tcp_listener should only accept ip and ipv6 (not unix) address family
### Describe the bug

TCP/IP is streaming IP and as such requires an IP based protocol.  We could perhaps refactor the whole thing and call this "streaming" and merge unix.listen_stream into a single net.listen_stream (and net.listen_datagram (for UDP in the specific instance of IP).)

ANYWHO, that big refactor can wait.  For now I think it is sufficient that tcp_listener throws an error indicating that ip or ipv6 address family is the only supported ones.

Also, disable the misplaced test net.tcp_test.v `test_tcp_unix` which is just a mess right now until it is cleaned up on a bigger architectural level.

### References

```v
pub fn listen_tcp(family AddrFamily, saddr string, options ListenOptions) !&TcpListener {
	mut s := new_tcp_socket(family) or { return error('${err.msg()}; could not create new socket') }
	s.set_dualstack(options.dualstack) or {}

	addrs := resolve_addrs(saddr, family, .tcp) or {
		return error('${err.msg()}; could not resolve address ${saddr}')
	}
	// TODO(logic to pick here)
	addr := addrs[0]

	// cast to the correct type
	alen := addr.len()
	socket_error_message(C.bind(s.handle, voidptr(&addr), alen), 'binding to ${saddr} failed')!
	socket_error_message(C.listen(s.handle, options.backlog), 'listening on ${saddr} with maximum backlog pending queue of ${options.backlog}, failed')!
	return &TcpListener{
		sock: s
		accept_deadline: no_deadline
		accept_timeout: infinite_timeout
	}
}
```

```v
// listen_stream creates an unix domain socket at `socket_path`
pub fn listen_stream(socket_path string, options ListenOptions) !&StreamListener {
	if socket_path.len >= max_sun_path {
		return error('Socket path too long! Max length: ${max_sun_path - 1} chars.')
	}
	mut s := new_stream_socket(socket_path) or {
		return error('${err.msg()}; could not create new unix stream socket')
	}

	addrs := net.resolve_addrs(socket_path, .unix, .tcp) or {
		return error('${err.msg()}; could not resolve path ${socket_path}')
	}
	addr := addrs[0]

	// cast to the correct type
	alen := addr.len()

	// try to unlink/remove an existing filesystem object at `socket_path`. Ignore errors,
	// because it's ok if the path doesn't exists and if it exists, but can't be unlinked
	// then `bind` will generate an error
	$if windows {
		os.rm(socket_path) or {}
	} $else {
		C.unlink(&char(socket_path.str))
	}

	net.socket_error_message(C.bind(s.handle, voidptr(&addr), alen), 'binding to ${socket_path} failed')!
	net.socket_error_message(C.listen(s.handle, options.backlog), 'listening on ${socket_path} with maximum backlog pending queue of ${options.backlog}, failed')!
	return &StreamListener{
		sock: s
		accept_deadline: no_deadline
		accept_timeout: infinite_timeout
	}
}
```

test_tcp.v section to be removed:

```v
fn test_tcp_unix() {
	eprintln('\n>>> ${@FN}')
	// TODO(emily):
	// whilst windows supposedly supports unix sockets
	// this doesnt work (wsaeopnotsupp at the call to bind())
	$if !windows {
		address := os.real_path('tcp-test.sock')
		// address := 'tcp-test.sock'
		println('${address}')

		mut l := net.listen_tcp(.unix, address) or { panic(err) }
		start_echo_server(mut l)
		echo(address) or { panic(err) }
		l.close() or {}

		os.rm(address) or { panic('failed to remove socket file') }
	}
}
```

### Reproduction Steps

Run 

```VTEST_FAIL_FLAKY=1 v test vlib/net```

### Expected Behavior

The tests should not fail.  Yet, Unix datagrams should work. And they do, if the right API is used.

### Current Behavior

Ignored tests, and developer confusion on how to use Unix datagrams (UDS) properly with V.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.3 02c0f3f

NAME="Artix Linux"
PRETTY_NAME="Artix Linux"
ID=artix
BUILD_ID=rolling
ANSI_COLOR="0;36"
HOME_URL="https://www.artixlinux.org/"
DOCUMENTATION_URL="https://wiki.artixlinux.org/"
SUPPORT_URL="https://forum.artixlinux.org/"
BUG_REPORT_URL="https://bugs.artixlinux.org/"
PRIVACY_POLICY_URL="https://terms.artixlinux.org/docs/privacy-policy/"
LOGO=artixlinux-logo
> [!NOTE]
>

### PR DESCRIPTION
**Pull request by hholst80**
Closes #20307

### DIFF
--- a/vlib/net/tcp.c.v
+++ b/vlib/net/tcp.c.v
@@ -301,6 +301,9 @@ pub:
 }
 
 pub fn listen_tcp(family AddrFamily, saddr string, options ListenOptions) !&TcpListener {
+	if family != .ip && family != .ip6 {
+		return error('listen_tcp only supports ip and ip6')
+	}
 	mut s := new_tcp_socket(family) or { return error('${err.msg()}; could not create new socket') }
 	s.set_dualstack(options.dualstack) or {}
 

### CODE REVIEW

File: vlib/net/tcp.c.v [hunk 1]
```diff
@@ -301,6 +301,9 @@ pub:
 }
 
 pub fn listen_tcp(family AddrFamily, saddr string, options ListenOptions) !&TcpListener {
+	if family != .ip && family != .ip6 {
+		return error('listen_tcp only supports ip and ip6')
```
JalonSolov, line 305:
```suggestion
		return error('listen_tcp only supports ipv4 and ipv6')
```

vlib/net/tcp_test.v
```diff
@@ -76,23 +62,25 @@ fn test_tcp_ip() {
...
+fn test_tcp_unix() {
+	eprintln('\n>>> ${@FN}')
+	address := 'tcp-test.sock'
+	eprintln('${address}')
 
-		os.rm(address) or { panic('failed to remove socket file') }
-	}
+	mut l := net.listen_tcp(.unix, address) or { return }
+	assert false
 }
```
spytheman, line 84: imho make a new _test.v file for this specific test (that directly tests the new check)
and revert vlib/net/tcp_test.v . The other changes for vlib/net/tcp_test.v can be done in a separate PR, specialized just for that.

spytheman: I know that it is tempting to do several things in the same PR, but it also makes it so that when some of the changes fail for some reason, to delay the other good ones.
Finding the root cause of a problem when regression testing/bisecting also becomes harder, when a single commit (and each PR gets squashed to a single commit in the V repo), contains unrelated changes.

Last but not least, reviewing becomes harder too, and Alex, Joe and me, have to review and merge sometimes dozens of PRs a day.

hholst80: I see your point.  If you do not mind I'll skip the new test file for this case because it will be orthogonal to the existing test: one checks that it should work (but is disabled) and the new one checks that it should fail.... better just keep the single broken test instead of adding confusion in the commit log.
This PR now adds just the address check and the other PR includes the changes to the _tests. With that PR the tcp_listener tests should be fully working and I can focus on the more complicated PR I am working on with the buffered_reader for http requests. :)

spytheman (approved):
Good work.

PR #20339, opened 2024-01-01
net: refactor tcp_test.v

hholst80:
Fix a broken test and sort _test functions together.

Note that this PR depends on #20336


hholst80:
Fix fails on FreeBSD Code CI so I need to rework this and make it more robust so we can remove the flaky flag without causing instabilities.

PR #20342, opened 2024-01-01
vlib: fix syntax error in go_array_wait_test.v

hholst80:
Fix compiler syntax error.  At some point in time this syntax might have been legal, but now it is neither legal nor required.


spytheman (approved):
Good work.

PR #20349, opened 2024-01-02
cgen: fix option initialization with default struct initialization to not be `none`

### LINKED ISSUE #20345: Option not initialised to default values
### Describe the bug

When using options, a struct may not be initalised with default parameters.

### Reproduction Steps

If a is not set to 10, then the struct test is not set with initial values.

```
pub struct Test {
    a int = 10
    b int = 5 
}
pub struct ABC {
    test ?&Test
}
fn main() {
    abc := ABC{
        test: &Test{
            a: 10    <---------- Try commenting this line out
        }
    }

    if ttt := abc.test {
        println(" ABC ${abc}  ${ttt}")  // with it  b = 5,  without test = 0  !!
    }
}
```

### Expected Behavior

If test was not an option then  &Test{} has all default values as expected. But not using an Option.
I expected that the default values would be set with an option

### Current Behavior

The default parameters are not set using an empty constructor : `Test { }`

### Possible Solution

_No response_

_No response_

### V version

V 0.4.3 9694024

V full version: V 0.4.3 ab4c1f4.9694024
Processor: 8 cpus, 64bit, little endian,

getwd: C:\Users\myname
vexe: D:\Documents\_Projects\_VLang\v\v.exe
vexe mtime: 2023-12-30 23:43:48

vroot: OK, value: D:\Documents\_Projects\_VLang\v
VMODULES: OK, value: C:\Users\myname\.vmodules
VTMP: OK, value: C:\Users\myname\AppData\Local\Temp\v_0

Git version: git version 2.9.0.windows.1
Git vroot status: weekly.2023.52-30-g96940246 (17 commit(s) behind V master)
.git/config present: true

thirdparty/tcc status: thirdparty-windows-amd64 e90c2620
> [!NOTE]
>

felipensp:
Fix #20345

```V
pub struct Test {
	a int = 10
	b int = 5
}

pub struct ABC {
	test ?&Test
}

fn test_main() {
	abc := ABC{
		test: &Test{}
	}

	if ttt := abc.test {
		assert ttt.a == 10
		assert ttt.b == 5
	} else {
		assert false
	}
}
```

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -1968,7 +1968,7 @@ fn (mut g Gen) expr_with_tmp_var(expr ast.Expr, expr_typ ast.Type, ret_typ ast.T
 		}
 		if ret_typ.has_flag(.option) {
 			if expr_typ.has_flag(.option) && expr in [ast.StructInit, ast.ArrayInit, ast.MapInit] {
-				if expr is ast.StructInit {
+				if expr is ast.StructInit && expr.init_fields.len > 0 {
 					g.write('_option_ok(&(${styp}[]) { ')
 				} else {
 					g.write('_option_none(&(${styp}[]) { ')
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Very elegant.
Excellent work.

Issue #20348, opened 2024-01-02
cgen error: _ := voidptr(a)


enghitalo:
### Describe the bug

https://play.vlang.io/p/51acb00f49

### Reproduction Steps

https://play.vlang.io/p/51acb00f49
```v
a := ''
unsafe {
_ := voidptr(a)
}
```

### Expected Behavior

not cgen error
maybe `voidptr(a) == 0` ?

### Current Behavior

```bash
Running code...
Can't run code. The server returned an error:
/tmp/v_60000/../../../../../../box/code.v:3: error: cannot convert 'struct string' to 'void *'
builder error: 
==================
C error. This should never happen.

This is a compiler bug, please report it using `v bug file.v`.

https://github.com/vlang/v/issues/new/choose

You can also use #help on Discord: https://discord.gg/vlang

Exited with error status 1
Please try again.
```

### V version

https://play.vlang.io/ version

### Environment details (OS name and version, etc.)

https://play.vlang.io/ OS
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

shove70:
I think, `string -> voidptr`  Should be disallowed, whether it's in unsafe or not.

Delta456:
A C error is also thrown when `byteptr` and `charptr` casts are used instead.

dumblob:
@Delta456 @JalonSolov did you get any feedback on disallowing cast of `string` to any pointer (`byteptr` `charptr` ...)?

Delta456:
> @Delta456 @JalonSolov did you get any feedback on disallowing cast of `string` to any pointer (`byteptr` `charptr` ...)?

Nope.

JalonSolov:
It shouldn't be allowed, because `string` is a struct.  It's not a `charptr` (which is `char *`) or a `byteptr (which is basically the same thing).  Casting `string.str` to one of these will be fine, but not `string` itself.

PR #20351, opened 2024-01-02
 checker: disallow `string` to `voidptr` cast entirely

Delta456:
Fixes https://github.com/vlang/v/issues/20348


JalonSolov:
Shouldn't it fail for any cast of string to pointer?  voidptr, charptr, etc.  They should all give a similar message:  cannot cast string to <whatever>, use string.str instead

vlib/v/checker/checker.v
```diff
@@ -3181,7 +3181,7 @@ fn (mut c Checker) cast_expr(mut node ast.CastExpr) ast.Type {
 		tt := c.table.type_to_str(to_type)
 		c.error('cannot cast string to `${tt}`, use `${snexpr}[index]` instead.', node.pos)
 	} else if final_from_sym.kind == .string && to_type.is_voidptr()
-		&& !node.expr_type.has_flag(.generic) {
+		&& !node.expr_type.has_flag(.generic) && !from_type.is_ptr() {
 		c.error('cannot cast string to `voidptr`, use voidptr(s.str) instead', node.pos)
```
JalonSolov, line 3185:
```suggestion
		c.error('cannot cast string to a pointer, use pointer_type(s.str) instead', node.pos)
```
Unless you know the exact type of pointer used in the cast.  Just saying can't cast to voidptr doesn't make sense if it was a cast to another pointer type.

Delta456: Good point but I only check for `voidptr` by using `to_type.is_voidptr()`. I will do a generalised check when the discussion is done.

spytheman (approved):
Good work.

PR #20353, opened 2024-01-02
os: small cleanup in the FreeBSD branch of os.executable/0: use fixed array for the sysctl params, instead of allocating a dynamic one

spytheman:


jasperla:
While you're at it, you could replace the magic constants with `C.CTL_KERN` etc since `sys/sysctl.h` is already included.

PR #20357, opened 2024-01-02
fix: avoid unnecessary push

enghitalo:

### DIFF
--- a/vlib/strings/builder.c.v
+++ b/vlib/strings/builder.c.v
@@ -223,7 +223,10 @@ pub fn (b &Builder) after(n int) string {
 // The returned string *owns* its own separate copy of the accumulated data that was in
 // the string builder, before the .str() call.
 pub fn (mut b Builder) str() string {
-	b << u8(0)
+	if b.last() != 0 {
+		// vlang need zero at end or not?
+		b << u8(0)
+	}
 	bcopy := unsafe { &u8(memdup_noscan(b.data, b.len)) }
 	s := unsafe { bcopy.vstring_with_len(b.len - 1) }
 	b.trim(0)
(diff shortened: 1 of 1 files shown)


vlib/strings/builder.c.v
```diff
@@ -223,7 +223,10 @@ pub fn (b &Builder) after(n int) string {
 // The returned string *owns* its own separate copy of the accumulated data that was in
 // the string builder, before the .str() call.
 pub fn (mut b Builder) str() string {
-	b << u8(0)
+	if b.last() != 0 {
+		// vlang need zero at end or not?
```
JalonSolov, line 227: Yes, if `b` is a []u8.

From the docs for `vstring_with_len`:

// vstring_with_len converts a C style 0 terminated string to a V string.

spytheman, line 227: V itself does not need it, but you do not know here, how the resulting string will be used, and its .str pointer can be passed to a C API, that expects a C string, which will need the 0 terminator.

spytheman: Also, please use discord for discussions and questions like this one (or github's discussions), not inline comments in the code, that will later need editing.

spytheman: We can add a separate `@[unsafe]` method, that can do the conversion, without adding a 0 at the end, and document it well. Then it can be used by callers that are well aware of the risks.

JalonSolov: I vote against that.  V strings are supposed to have NUL terminated u8 arrays.  Saving that one extra byte is not worth the headaches.

vlib/strings/builder.c.v
```diff
@@ -226,6 +226,15 @@ pub fn (mut b Builder) str() string {
...
 		b << u8(0)
+	} else {
+		for i := b.len - 1; i > 0; i-- {
+			if (b[i - 1] & 0xff) != 0 {
+				unsafe {
+					b.len = i + 1 // include zero
+				}
+				break
+			}
+		}
```
spytheman, line 237: This is useless :-|. Imho just do not push the 0s in the first place, if you do not intend to have them.

PR #20359, opened 2024-01-03
checker: rename `check_expr_opt_call` to `check_expr_result_call`

steven-mathew:
Renames `check_expr_opt_call` function to match the behavior since only the `.result` flag is cleared.

### DIFF
--- a/vlib/v/checker/assign.v
+++ b/vlib/v/checker/assign.v
@@ -37,7 +37,7 @@ fn (mut c Checker) assign_stmt(mut node ast.AssignStmt) {
 			if i == 0 {
 				right_first_type = right_type
 				node.right_types = [
-					c.check_expr_opt_call(right, right_first_type),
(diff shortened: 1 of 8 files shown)


shove70:
No, check both.
For historical reasons (splitting optional into optional and result), there are places in the system where `opt` is used to refer to both.

spytheman:
@shove70, @steven-mathew how about `check_expr_opt_call` -> `check_expr_option_or_result_call` ?

shove70:
> @shove70, @steven-mathew how about `check_expr_opt_call` -> `check_expr_option_or_result_call` ?

yep, this is reasonable

shove70:
@steven-mathew You can change it to `check_expr_option_or_result_call`, as spytheman's suggests

spytheman:
The CI failure is unrelated. Excellent work.

spytheman (approved):

PR #20367, opened 2024-01-03
cgen: fix code generation when the function returns mut fixed array(fix #20366)

### LINKED ISSUE #20366: cgen: C error when the function returns mut fixed array
### Describe the bug

```v
pub fn foo(mut fixed_array [3]u8) [3]u8 {
	return fixed_array
}

fn main() {
	mut fixed := [3]u8{}
	mut res := foo(mut fixed)
	println(res)
}
```

```
==================
              ^
/tmp/v_501/a.1882641659735005790.tmp.c:12674:22: error: initializing '_v_Array_fixed_u8_3' (aka 'struct _v_Array_fixed_u8_3') with an expression of incompatible type 'Array_fixed_u8_3' (aka 'unsigned char[3]')
        _v_Array_fixed_u8_3 _t1 = *fixed_array;
                            ^     ~~~~~~~~~~~~
1 warning and 1 error generated.
...
```

### Reproduction Steps

ditto

### Expected Behavior

no errors.

### Current Behavior

C error.

### Possible Solution

_No response_

_No response_

### V version

latest master

all
> [!NOTE]
>

shove70:
1. Fixed #20366
2. Add tests.

```v
pub fn foo(mut fixed [3]int) [3]int {
	return fixed
}

fn main() {
	mut fixed := [3]int{}
	mut res := foo(mut fixed)
	println(res)
}
```
outputs:
```
[0, 0, 0]
```

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5284,7 +5284,7 @@ fn (mut g Gen) return_stmt(node ast.Return) {
 			g.autofree_scope_vars(node.pos.pos - 1, node.pos.line_nr, true)
 			g.write('return ')
 		}
-		if expr0.is_auto_deref_var() {
+		if expr0.is_auto_deref_var() && !fn_return_is_fixed_array {
 			if g.fn_decl.return_type.is_ptr() {
 				var_str := g.expr_string(expr0)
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #20369, opened 2024-01-03
diff: Add support for OpenBSD's diff implementation

jasperla:
OpenBSD's diff doesn't have the long option names like FreeBSD does.


spytheman (approved):
Excellent work.

PR #20370, opened 2024-01-03
tests: Also skip native tests on OpenBSD

jasperla:
Just like on FreeBSD, skip the native tests for now.


spytheman (approved):
Good work.

PR #20377, opened 2024-01-03
x.vweb, picoev: fix timeout event

### LINKED ISSUE #20323: x.vweb missing response on Content-Length > body payload
### Describe the bug

x.vweb should await data up to the timeout and then respond with an 400 status code (Invalid payload).

The current behavior is at best, weird.

### Reproduction Steps

```v
import x.vweb

const port = 1337

pub struct App {
}

pub struct Context {
	vweb.Context
}

fn main() {
	mut app := &App{}
	vweb.run_at[App, Context](mut app, port: port, timeout_in_seconds: 2, family: .ip) or {
		panic('could not start vweb app')
	}
}
```

```
[I] root@trump ~
# curl -v -H Content-Length:10 http://127.0.0.1:1337/ -d123456789
*   Trying 127.0.0.1:1337...
* Connected to 127.0.0.1 (127.0.0.1) port 1337
> POST / HTTP/1.1
> Host: 127.0.0.1:1337
> User-Agent: curl/8.5.0
> Accept: */*
> Content-Length: 10
> Content-Type: application/x-www-form-urlencoded
> 
* Empty reply from server
* Closing connection
curl: (52) Empty reply from server
[I] root@trump ~ [52]
# curl -v -H Content-Length:10 http://127.0.0.1:1337/ -d123456789
*   Trying 127.0.0.1:1337...
* Connected to 127.0.0.1 (127.0.0.1) port 1337
> POST / HTTP/1.1
> Host: 127.0.0.1:1337
> User-Agent: curl/8.5.0
> Accept: */*
> Content-Length: 10
> Content-Type: application/x-www-form-urlencoded
> 
< HTTP/1.1 400 Bad Request
< Content-Type: text/plain
< Server: VWeb
< Connection: close
< Content-Length: 49
< 
* Closing connection
Mismatch of body length and Content-Length header⏎                 
[I] root@trump ~
#
```

### Expected Behavior

HTTP/1.1 400 Bad Request

should be returned after a specified timeout (in x.vweb). Now the first connection is closed after (2 seconds, specified in vtest.v) a timeout and the second attempt to connect will add the new payload to the existing request.  This can be proven by changing Content-Length in the first request to something much larger like 1000.

### Current Behavior

See repro.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.3 ac06458

NAME="Artix Linux"
PRETTY_NAME="Artix Linux"
ID=artix
BUILD_ID=rolling
ANSI_COLOR="0;36"
HOME_URL="https://www.artixlinux.org/"
DOCUMENTATION_URL="https://wiki.artixlinux.org/"
SUPPORT_URL="https://forum.artixlinux.org/"
BUG_REPORT_URL="https://bugs.artixlinux.org/"
PRIVACY_POLICY_URL="https://terms.artixlinux.org/docs/privacy-policy/"
LOGO=artixlinux-logo
> [!NOTE]
>

Casper64:
fixes #20323 

This pr adds a missing call to the raw callback in picoev on a timeout event. 

When a timeout occurs `x.vweb` sends a HTTP 408 Request Timeout response.

This pr probably also fixes the flakiness issues in the `large_payload_test.v`, but I'll check that in some time

### DIFF
--- a/vlib/x/vweb/tests/large_payload_test.v
+++ b/vlib/x/vweb/tests/large_payload_test.v
@@ -91,8 +91,8 @@ fn test_bigger_content_length() {
 		data: data
 	})!
 
-	assert x.status() == .bad_request
-	assert x.body == 'Mismatch of body length and Content-Length header'
+	// Content-length is larget than the data sent, so the request should timeout
+	assert x.status() == .request_timeout
 }
 
 fn test_smaller_content_length() {
--- a/vlib/x/vweb/tests/large_payload_test.v
+++ b/vlib/x/vweb/tests/large_payload_test.v
@@ -91,7 +91,7 @@ fn test_bigger_content_length() {
 		data: data
 	})!
 
-	// Content-length is larget than the data sent, so the request should timeout
+	// Content-length is larger than the data sent, so the request should timeout
 	assert x.status() == .request_timeout
 }
 
--- a/vlib/picoev/picoev.v
+++ b/vlib/picoev/picoev.v
@@ -216,6 +216,12 @@ fn raw_callback(fd int, events int, context voidptr) {
 		$if trace_fd ? {
 			eprintln('timeout ${fd}')
 		}
+
+		if !isnil(pv.raw_cb) {
+			pv.raw_cb(mut pv, fd, events)
+			return
+		}
+
 		pv.close_conn(fd)
 		return
 	} else if events & picoev.picoev_read != 0 {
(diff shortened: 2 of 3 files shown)


vlib/x/vweb/tests/large_payload_test.v
```diff
@@ -91,8 +91,8 @@ fn test_bigger_content_length() {
 		data: data
 	})!
 
-	assert x.status() == .bad_request
-	assert x.body == 'Mismatch of body length and Content-Length header'
+	// Content-length is larget than the data sent, so the request should timeout
```
spytheman, line 94:
```suggestion
	// Content-length is larger than the data sent, so the request should timeout
```

spytheman (commented):
Excellent work.

PR #20379, opened 2024-01-04
cgen: fix returns struct with mut fixed array init(with/without generics)(fix #20361)

### LINKED ISSUE #20361: fixed array on struct field is overwriting
### Describe the bug

first element is changed

### Reproduction Steps

```v
struct Builder[T] {
mut:
	buf T
}

pub fn new_builder[F](mut fixed_array F) !Builder[F] {
	$if F is $array_fixed {
		return Builder[F]{fixed_array}
	} $else {
		return error('fixed_buf need be a fixed array')
	}
}

fn main() {
	mut fixed := [3]u8{}
	mut sb := new_builder(mut fixed) or {
		eprintln(err)
		return
	}
	println(sb)
}
```

### Expected Behavior

```bash
╰─ v -prod crun e.v ─╯
Builder[[3]u8]{
    buf: [0, 0, 0]
}

╰─ v -prod crun e.v ─╯
Builder[[3]u8]{
    buf: [0, 0, 0]
}

╰─ v -prod crun e.v─╯
Builder[[3]u8]{
    buf: [0, 0, 0]
}
```

### Current Behavior

```bash
╰─ v -prod crun e.v ─╯
Builder[[3]u8]{
    buf: [5, 0, 0]
}

╰─ v -prod crun e.v ─╯
Builder[[3]u8]{
    buf: [69, 0, 0]
}

╰─ v -prod crun e.v─╯
Builder[[3]u8]{
    buf: [197, 0, 0]
}
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.3 bf7b29a

Ubuntu
> [!NOTE]
>

shove70:
1. Fixed #20361
2. Both the cases with and without generics are fixed
3. Add tests.

```v
struct Builder {
mut:
	buf [3]int
}

pub fn new_builder(mut fixed [3]int) !Builder {
	return Builder{fixed}
}

fn main() {
	mut fixed := [3]int{}
	mut sb := new_builder(mut fixed) or {
		return
	}
	println(sb)
}
```
outputs:
```
[0, 0, 0]
```

```v
struct Builder[T] {
mut:
	buf T
}

pub fn new_builder[F](mut fixed_array F) !Builder[F] {
	$if F is $array_fixed {
		return Builder[F]{fixed_array}
	} $else {
		return error('fixed_buf need be a fixed array')
	}
}

fn main() {
	mut fixed := [3]u8{}
	mut sb := new_builder(mut fixed) or {
		eprintln(err)
		return
	}
	println(sb)
}
```
outputs:
```
Builder[[3]u8]{
    buf: [0, 0, 0]
}
```

### DIFF
--- a/vlib/v/tests/struct_init_with_fixed_array_field_test.v
+++ b/vlib/v/tests/struct_init_with_fixed_array_field_test.v
@@ -26,3 +26,35 @@ fn test_struct_init_with_fixed_array_field() {
 	println(game.board)
 	assert '${game.board}' == '[x, o, x, o, o, x, x, x, o]'
 }
+
+// for issue 20361(part 1): returns struct with init mut fixed array fields without generics
+struct Foo {
+mut:
+	buf [3]int
+}
+
+pub fn returns_struct_with_mut_fixed_array_init(mut fixed [3]int) Foo {
+	return Foo{fixed}
+}
+
+fn test_returns_struct_with_mut_fixed_array_init() {
+	mut fixed := [3]int{}
+	mut foo := returns_struct_with_mut_fixed_array_init(mut fixed)
+	assert foo.buf == [0, 0, 0]!
+}
+
+// for issue 20361(part 2): returns struct with init mut fixed array fields with generics
+struct Bar[T] {
+mut:
+	buf T
+}
+
+pub fn returns_struct_with_mut_fixed_array_init_with_generics[T](mut fixed T) Bar[T] {
+	return Bar[T]{fixed}
+}
+
+fn test_returns_struct_with_mut_fixed_array_init_with_generics() {
+	mut fixed := [3]int{}
+	mut bar := returns_struct_with_mut_fixed_array_init_with_generics(mut fixed)
+	assert bar.buf == [0, 0, 0]!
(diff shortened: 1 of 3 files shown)


vlib/v/tests/struct_init_with_fixed_array_field_test.v
```diff
@@ -26,3 +26,35 @@ fn test_struct_init_with_fixed_array_field() {
...
+mut:
+	buf [3]int
+}
+
+pub fn returns_struct_with_mut_fixed_array_init(mut fixed [3]int) Foo {
+	return Foo{fixed}
+}
+
+fn test_returns_struct_with_mut_fixed_array_init() {
+	mut fixed := [3]int{}
```
spytheman, line 41: It is a bit better to use non zero values for such tests, like for example `mut fixed := [5,99,1]!`, so that later the assertion can verify that they remained intact, and were copied.

With just 0s, since that is the default, it is not clear if the function copied them, or did it just initialized a new fixed array value, due to a bug/regression.

spytheman (approved):
Excellent work.

PR #20381, opened 2024-01-04
add try_lock to mutex/rwmutex

kbkpbot:
sync: add try_lock support for mutex/rwmutex


spytheman (approved):
Excellent work.

spytheman:
Can you also add a small test for each function?

PR #20384, opened 2024-01-04
cgen: fix option/result fixed array init and related tests

shove70:
Many thanks @spytheman for a review suggest. I found a problem while adjusting the related test case! It is now cleanup and fixed!
Adore very much 👍


spytheman (approved):
Excellent work.

Less code, less bugs :-)

Issue #20385, opened 2024-01-04
for in range type isn't checked properly 


Delta456:
### Describe the bug

Run the following program

```v
for _ in  map[string]string..`9`{
}
```

### Reproduction Steps

Same as above

### Expected Behavior

```
file.v:3:11: error: range type cannot be `map[string]string`
    1 | 
    2 | 
    3 | for _ in  map[string]string..`9`{
      |           ~~~
    4 |   
    5 |
```

### Current Behavior

A C error

### Possible Solution

A checker fix which I am working on

### V version

Latest

### Environment details (OS name and version, etc.)

Not relevant to the issue.

PR #20386, opened 2024-01-04
checker: enhance `for` range loop check and add better positioning

Delta456:
Fixes https://github.com/vlang/v/issues/20385


Delta456:
> Hmm... instead of separate messages for each type it can't be, why not one message to say the only types it can be?  Something like
> ```
> error: range type can only be an integer type
> ```
> or whatever.  That will cover strings, `none`, structs, enums, etc., etc., etc.

I thought separate messages for each type would be better but also others are handled the way you have already suggested.

I will do the changes ASAP.

Delta456:
Done with changes.

spytheman (approved):
Excellent work.

PR #20389, opened 2024-01-04
vlib: fix tcp_test.v

hholst80:
The listen_tcp method should not accept .unix address family.

Test that it indeed fails when invoked.

Also, remove the flaky flag on tcp_test.v as none of the tests here has any real reason to be flaky.

Tested on Linux 6.6 (amd64) and FreeBSD 13.2 (amd64).

### DIFF
--- a/vlib/net/tcp_test.v
+++ b/vlib/net/tcp_test.v
@@ -1,20 +1,17 @@
-// vtest flaky: true
-// vtest retry: 8
 import net
-import os
 
 const test_port = 45123
 
-fn handle_conn(mut c net.TcpConn) {
+fn handle_conn(mut c net.TcpConn) ! {
 	for {
 		mut buf := []u8{len: 100, init: 0}
 		read := c.read(mut buf) or {
-			println('Server: connection dropped')
-			return
+			eprintln('Server: connection dropped')
+			return err
 		}
 		c.write(buf[..read]) or {
-			println('Server: connection dropped')
-			return
+			eprintln('Server: connection dropped')
+			return err
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


JalonSolov:
Panics on Windows... maybe part of why it was marked flaky...

Casper64:
> 	0.0.0.0:45123: net: op timed out; code: 9

Why is it trying to dial to an ipv4 address when an ipv6 address is given? Seems like a bug in the address resolve code to me

hholst80:
> > 0.0.0.0:45123: net: op timed out; code: 9
> 
> Why is it trying to dial to an ipv4 address when an ipv6 address is given? Seems like a bug in the address resolve code to me

I found some funky looking code in tcp.c.v that I removed.  It was responsible for replacing ::: with localhost which resolved to an IPv6 address.  Having a generic resolve which magically figures out the address family is probably not very safe, but yet common.  I've ran into a similar issue with Node.js too.

vlib/net/tcp_test.v
```diff
@@ -103,6 +85,6 @@ fn test_bind() {
 	$if !network ? {
 		return
 	}
-	mut conn := net.dial_tcp_with_bind('vlang.io:80', '127.0.0.1:0')!
-	conn.close()!
```
hholst80, line 107: How could this ever have worked? test_bind() is not declared with `!` so I thought conn.close()! would be illegal (nothing is returned)?

hholst80:
OK. That leads into a rabbit hole.  What does *not* work on Windows is to open a connection to 0.0.0.0.  That is only a valid IP address to bind to for accept.  If we know for sure that the server is listening on all devices on a specific port, then we know for sure that it is listening on 127.0.0.1 (within reason).  The fix should be to not use `${server.addr}` but `127.0.0.1` in the fetch in `server_test.v`:

image

vlib/net/tcp_test.v
spytheman: `test_` functions act similar to `main`. They are both exception for the general rule, that functions should themselves return ! or ?, if you want to use f()! or f()? inside them.

The reasoning for that, is that both main() and the test_...() functions are not called by V code themselves, but by generated code, so there is nowhere for the propagation to go, and instead in these top level functions, propagation just panics on the spot.

It also makes them a bit easier to write, since you only have to remember `fn main() {`, instead of having to remember to change it to `fn main() ! {` or `fn main() ? {`.

spytheman: Personally, I also dislike this decision, even when I know the motivation and reasons behind it.

In my opinion, having *all functions* behave uniformly, without syntax exceptions, would have been more consistent and easier to remember 🤷🏻‍♂️ .

hholst80:
The failing Docker tests were really good! It showed me that the tcp_test.v IPv6 tests *actually uses* IPv6 now. ;-)

I want to disable to IPv6 tests on environments where there is known not to exist IPv6 stack, or validate that the test fails.  Any ideas on how to do that?  I know this PR is getting out of hand but first I will fix it then I will shop it up into reasonable commit.

esquerbatua:
This also seems related with this: https://github.com/vlang/v/issues/20441

spytheman:
Closing for inactivity.

PR #20392, opened 2024-01-05
parser: fix checking for invalid PrefixExpr(fix #20388)

### LINKED ISSUE #20388: C error when compiling an invalid `x - - -y` expression
**V doctor:**
```
V full version: V 0.4.3 5306469.7d29afe

vexe mtime: 2024-01-04 15:05:36

Git vroot status: weekly.2024.01.test2
.git/config present: true

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg main.v`
```v
fn main() {
	y := 5 - - -5
}

```

Compile-time error because of an invalid expression

**What did you see instead?**
```
main.v:2:2: warning: unused variable: `y`
    1 | fn main() {
    2 |     y := 5 - - -5
      |     ^
    3 | }
==================
C:/Users/user/AppData/Local/Temp/v_0/main.16516255675734833237.tmp.c:6822: warning: implicit declaration of function 'tcc_backtrace'
C:/Users/user/AppData/Local/Temp/v_0/main.16516255675734833237.tmp.c:12810: error: lvalue expected
...
==================

builder error: 
==================

```

Generated C code:

image

> [!NOTE]
>

shove70:
1. Fixed #20388 
2. Add tests.

```v
_ = 5 - - -5
_ = 5 - - (-5)
_ = 5 - - ((((-5))))
```

```
a.v:2:11: error: invalid expression: unexpected token `-`
    1 | fn main() {
    2 |     _ = 5 - - ((((-5))))
      |              ^
    3 | }
    4 |
```


spytheman:
That is a good stop gap measure for now. In the long term though, I think we should allow `5 - - -5` and fix cgen. It is allowed in both C and Go:

C:
```c
#include <stdio.h>
int main(){
	printf("x: %d\n", 5 - - -5);
	return 0;   
}
```
Go:
```go
package main

import "fmt"

func main() {
	fmt.Println("Hello", 5 - - -5)
}
```

shove70:
> That is a good stop gap measure for now. In the long term though, I think we should allow `5 - - -5` and fix cgen. It is allowed in both C and Go:
> 
> C:
> 
> ```c
> #include <stdio.h>
> int main(){
> 	printf("x: %d\n", 5 - - -5);
> 	return 0;   
> }
> ```
> 
> Go:
> 
> ```go
> package main
> 
> import "fmt"
> 
> func main() {
> 	fmt.Println("Hello", 5 - - -5)
> }
> ```

Oh, Ok, I've got it on record, and I'll deal with it at the right time.

PR #20393, opened 2024-01-05
builtin, checker: fix aliases of u8 array clone() call (related #20290)

yuyi98:
This PR fix aliases of u8 array clone() call (related #20290).


spytheman (approved):
Amazing work.

Less code, and more general.

PR #20399, opened 2024-01-05
cgen: fix assert expr handling + print with comptime smartcast

### LINKED ISSUE #20396: cgen: sumtype error when working with array, string and assert
### Describe the bug

`string _t1 = Array_main__Any_str((*val._Array_main__Any));` declared on wrong place? (for assert case)

### Reproduction Steps

```v
pub type Any = []Any
	| bool
	| f32
	| f64
	| i16
	| i64
	| i8
	| int
	| map[string]Any
	| string
	| u16
	| u32
	| u64
	| u8

fn main() {
	ana := Any([Any('')])

	name(ana) // error
	match ana {
		[]Any {
			for i := 0; i < ana.len; i++ {
				name(ana[i]) // error
			}
		}
		else {
			assert false
		}
	}
}

fn name[T](val T) {
	$for v in val.variants {
		if val is v {
			val.str() // error: cannot convert 'struct array' to 'unsigned char' // Terminated by signal 11 (SIGSEGV) // Any
			assert val.str() == '' //  error: '_t1' undeclared
			println(val) // error: cannot convert 'struct array' to 'unsigned char'
			dump(val) // works
		}
	}
}
```

### Expected Behavior

not cgen error

### Current Behavior

`error: cannot convert 'struct array' to 'unsigned char'`, `Terminated by signal 11 (SIGSEGV)`, `error: '_t1' undeclared`

### Possible Solution

_No response_

_No response_

### V version

V 0.4.3 bf7b29a

Ubuntu
> [!NOTE]
>

### LINKED ISSUE #20398: feat: $if (int(typeof(val).idx) >> 16) & 0xff == 0 {
### Describe the feature

Every info there is knowed at compile time there are not why it not be allowed

### Use Case

`json2 encode/decode`
```v
pub fn (e &Encoder) encode_value[T](val T, mut buf []u8) ! {
	  $if (int(typeof(val).idx) >> 16) & 0xff == 0 {

	}
}
```

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### Version used

V 0.4.3 bf7b29a

Ubuntu

felipensp:
Fix #20396

`assert_stmt()` was changing node.expr.left and node.expr.right inside. 

```V
pub type Any = []Any
	| bool
	| f32
	| f64
	| i16
	| i64
	| i8
	| int
	| map[string]Any
	| string
	| u16
	| u32
	| u64
	| u8

fn test_main() {
	ana := Any([Any('')])
	name(ana)
	match ana {
		[]Any {
			for i := 0; i < ana.len; i++ {
				name(ana[i])
			}
		}
		else {
			assert false
		}
	}
	assert true
}

fn name[T](val T) {
	$for v in val.variants {
		if val is v {
			dump(val.str())
			$if val is []Any {
				assert val.str() == "[Any('')]"
			} $else {
				assert val.str() == ''
			}
			dump(val)
		}
	}
}
```

### DIFF
--- a/vlib/v/gen/c/assert.v
+++ b/vlib/v/gen/c/assert.v
@@ -17,11 +17,17 @@ fn (mut g Gen) assert_stmt(original_assert_statement ast.AssertStmt) {
 	}
 	mut node := original_assert_statement
 	g.writeln('// assert')
+
+	mut save_left := ast.empty_expr
+	mut save_right := ast.empty_expr
+
 	if mut node.expr is ast.InfixExpr {
 		if subst_expr := g.assert_subexpression_to_ctemp(node.expr.left, node.expr.left_type) {
+			save_left = node.expr.left
 			node.expr.left = subst_expr
 		}
 		if subst_expr := g.assert_subexpression_to_ctemp(node.expr.right, node.expr.right_type) {
+			save_right = node.expr.right
 			node.expr.right = subst_expr
 		}
 	}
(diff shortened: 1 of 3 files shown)


felipensp:
> I believe you linked to the wrong issue. It should be #20396 instead.

Ooops, right. Thanks.

spytheman (approved):
Good work.

PR #20403, opened 2024-01-05
Make tetris example window not resizable

iacore:
On systems with tiling window manager, all resizable windows will be tiled.


spytheman:
The `resizable` field in gg.Config is currently ignored afaik.

Issue #20402, opened 2024-01-05
C Error in Struct Construction when not handling Results


splashsky:
### Describe the bug

When calling functions that have Results within the creation of a Struct, a C error results from not handling the Result of that function. This does not occur when creating a normal variable with a function that creates a Result. I feel that the error across both these cases should be identical; C errors are cryptic and don't really point out how to fix the mistake.

### Reproduction Steps

```v
payload := Payload{
    sub: "1234567890",
    ext: json.encode(claims),
    exp: time.parse("2019-01-01 00:00:00")
}
// C error: /tmp/v_1000/jwt_test.6624048298048836489.tmp.c:26827: error: cannot convert 'struct _result_time__Time' to 'int'

new := time.parse("2006-01-02 15:04:05")
// error: time.parse() returns a Result, so it should have either an `or {}` block, or `!` at the end
```

### Expected Behavior

I expect both cases to have the same compiler (not C) error:
```
error: time.parse() returns a Result, so it should have either an `or {}` block, or `!` at the end
```

### Current Behavior

Not handling the Result within a Struct causes a cryptic C error:
```
// C error: /tmp/v_1000/jwt_test.6624048298048836489.tmp.c:26827: error: cannot convert 'struct _result_time__Time' to 'int'
```

### V version

V 0.4.3 7d29afe

### Environment details (OS name and version, etc.)

OS: linux, Linux version 5.15.133.1-microsoft-standard-WSL2 (gcc (GCC) 11.2.0, GNU ld (GNU Binutils) 2.37) (WSL 2)
Processor: 8 cpus, 64bit, little endian, 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #20404, opened 2024-01-05
 checker: disallow `Result` type being used in struct init fields 

Delta456:
Fixes https://github.com/vlang/v/issues/20402


spytheman (approved):
Excellent work.

PR #20410, opened 2024-01-06
checker: fix checking of expression as mut receiver when calling methods of arrays/maps

shove70:
When calling the methods of array/map, the checking that expr cannot be used as mut receiver is missed,
This PR fixes it, as well as adding and modifying test cases.

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -2685,6 +2685,9 @@ fn (mut c Checker) map_builtin_method_call(mut node ast.CallExpr, left_type_ ast
 			}
 			if method_name[0] == `m` {
 				c.fail_if_immutable(mut node.left)
+				if !node.left.is_lvalue() {
+					c.error('cannot pass expression as `mut`', node.left.pos())
+				}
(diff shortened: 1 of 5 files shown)


vlib/v/checker/fn.v
```diff
@@ -2944,12 +2953,18 @@ fn (mut c Checker) array_builtin_method_call(mut node ast.CallExpr, left_type as
...
+			}
 			node.receiver_type = node.left_type.ref()
 		} else {
 			node.receiver_type = node.left_type
 		}
 	} else if method_name == 'delete' {
 		c.fail_if_immutable(mut node.left)
+		if !node.left.is_lvalue() {
+			c.error('cannot pass expression as `mut`', node.left.pos())
+		}
```
spytheman, line 2967: hm, that is better extracted in a separate method

spytheman (approved):
Good work.

vlib/v/checker/fn.v
shove70: done. 😃

PR #20411, opened 2024-01-06
parser: implement MyEnum.from generic static method

spytheman:
- parser: implement `Enum.from(123)!` and `Enum.from(abc)!`
- parser: cleanup the implementation of p.codegen/1
- debug and add tests

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -1331,7 +1331,7 @@ pub mut:
 pub struct EnumField {
 pub:
 	name          string // just `lock`, or `abc`, etc, no matter if the name is a keyword or not.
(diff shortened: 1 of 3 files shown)


vlib/v/ast/ast.v
```diff
@@ -1330,7 +1330,8 @@ pub mut:
 // enum field in enum declaration
 pub struct EnumField {
 pub:
-	name          string
+	name          string // just `lock`, or `abc`, etc, no matter if the name is a keyword or not.
+	source_name   string // The name in the source, for example `@lock`, and `abc`. Note that `@lock` is a keyword in V.
```
JalonSolov, line 1334:
```suggestion
	source_name   string // The name in the source, for example `@lock`, and `abc`. Note that `lock` is a keyword in V.
```

spytheman: Thank you.

PR #20413, opened 2024-01-06
net: fix socket non-blocking read/write

kbkpbot:
TcpConn non-blocking read/write will not work, even you set_blocking(false)

### DIFF
--- a/examples/buf_reader.v
+++ b/examples/buf_reader.v
@@ -6,6 +6,7 @@ import io
 fn main() {
 	// Make a new connection
 	mut conn := net.dial_tcp('google.com:80')!
+	conn.set_blocking(true)!
 	// Simple http HEAD request for a file
 	conn.write_string('GET /index.html HTTP/1.0\r\n\r\n')!
 	// Wrap in a buffered reader
--- a/examples/net_raw_http.v
+++ b/examples/net_raw_http.v
@@ -4,6 +4,7 @@ import io
 fn main() {
 	// Make a new connection
 	mut conn := net.dial_tcp('google.com:80')!
+	conn.set_blocking(true)!
 	defer {
 		conn.close() or {}
 	}
--- a/examples/net_udp_server_and_client.v
+++ b/examples/net_udp_server_and_client.v
@@ -13,6 +13,7 @@ fn main() {
 	if is_server {
 		println('UDP echo server, listening for udp packets on port: ${port}')
 		mut c := net.listen_udp('0.0.0.0:${port}')!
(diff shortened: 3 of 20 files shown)


shove70:
net module is non-blocking by default, so why change it to blocking?


> TcpConn non-blocking read/write will not work, even you set_blocking(false)

There should be other reasons, and I think should rule out other reasons first

JalonSolov:
Changing the default from non-blocking to blocking is a serious breaking change.  Even if that's the way it works now, people are expecting non-blocking from the current description, and that's the way they would have written their code.

shove70:
> Changing the default from non-blocking to blocking is a serious breaking change. Even if that's the way it works now, people are expecting non-blocking from the current description, and that's the way they would have written their code.

yes 👍

Casper64:
> I suggest change the default state of a tcp connection , from non-blocking mode to blocking mode.
> 
> As most examples, tests use a blocking mode tcp connection.
> 
> 

Everything in the net module uses non-blocking socket and `select` to check whether sockets are ready to read/write, so all examples use non-blocking sockets.

But the example code acts blocking, since we only use `select` on one socket. That is the difference.

I don't think this change is necessary and I agree with @JalonSolov that this change is too breaking

spytheman:
I agree that this change is too breaking.

Note, that you can achieve more or less the same, by just compiling with `-d net_blocking_sockets`.

In my personal opinion, blocking sockets are a better default (since blocking programs require less debugging), but others can reasonably disagree, and changing everything now, should be done after a lot of thought :-|, not by just modifying every program to have `server.set_blocking(true)` calls in it.

Casper64:
Hm I think I get what you mean. Why are there 2 read calls in the current implementation of `read_ptr`? 

This line

```v
if code == int(error_ewouldblock) { 
```

Could also check wether the socket is non-blocking. If it is non-blocking just return 0 (read bytes). The function exits immediately and that is what you expect from a non-blocking connection. If the connection is blocking we do want to wait for the socket to become ready to read.

It's a matter of documentation imo, because `read_ptr` and `read` will ALWAYS return some data, or an error. Even if it involves waiting for the socket to be ready.

Casper64:
@kbkpbot currently you can do a non-blocking read the way you described:

```v
fn non_blocking_read(mut c net.TcpConn, mut buf []u8) int {
    res := wrap_read_result(C.recv(c.sock.handle, voidptr(buf.data), buf.len, 0))!
    if res > 0 { return res }
    
    // handle errors
    return 0
}
```

PR #20426, opened 2024-01-07
net: add non-blocking read/write support to TCP connection

kbkpbot:
add non-blocking read/write support to TCP connection
read_nb: read non-blocking
write_nb: write non-blocking


spytheman:
Is not just calling `c.set_blocking(false)` enough (which is the default)?

JalonSolov:
It should be enough, which means there's a bug.  Shouldn't need completely separate `_nb` routines...

PR #20427, opened 2024-01-07
V init: Add all prompts as flags as well

exapsy:
## Adds

Flags on `v init`:

```
--description (if '' - empty string, then it's gonna prompt)
--project-version (--version conflicted with v's version, default: 0.0.0)
--license (default: MIT)
```

## Reason / What it fixes:

As a user, I wanted to make a quick bin/sh command that would auto-create a project with `v` without any prompts, I wanted all to be automatically and the prompts would auto-fill themselves with the information I already had from the script. But prompts preventing that. 

Even if there is another way, it feels much more straightforward to just have the prompts accessible as flags, so the user can notice them immediately.

### DIFF
--- a/cmd/tools/vcreate/vcreate.v
+++ b/cmd/tools/vcreate/vcreate.v
@@ -48,6 +48,21 @@ fn main() {
 			name: 'web'
 			description: 'Use the template for a vweb project.'
 		},
+		Flag{
+			flag: .string
+			name: 'description'
+			description: 'Description of the project.'
+		},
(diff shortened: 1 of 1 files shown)


ttytm:
To suggest a possible alternative, there could be a no-interactive flag to create a project with just the name. Desc, license and version would have to be changed manually in the mod file. Something like this might be used more often than the single flags.

spytheman:
No, I prefer the flags version that is here.

PR #20434, opened 2024-01-08
ast: add has_option_or_result() and cleanup all the related calls

yuyi98:
This PR add has_any_flag() and cleanup all the related calls.

### DIFF
--- a/vlib/v/ast/types.v
+++ b/vlib/v/ast/types.v
@@ -395,6 +395,16 @@ pub fn (t Type) has_flag(flag TypeFlag) bool {
 	return int(t) & (1 << (int(flag) + 24)) > 0
 }
 
+@[inline]
+pub fn (t Type) has_any_flag(flags ...TypeFlag) bool {
+	for flag in flags {
+		if int(t) & (1 << (int(flag) + 24)) > 0 {
+			return true
+		}
+	}
+	return false
+}
+
 // debug returns a verbose representation of the information in ts, useful for tracing/debugging
 pub fn (ts TypeSymbol) debug() []string {
 	mut res := []string{}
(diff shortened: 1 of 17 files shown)


vlib/v/ast/types.v
```diff
@@ -395,6 +395,16 @@ pub fn (t Type) has_flag(flag TypeFlag) bool {
 	return int(t) & (1 << (int(flag) + 24)) > 0
 }
 
+@[inline]
+pub fn (t Type) has_any_flag(flags ...TypeFlag) bool {
```
spytheman, line 399: Passing `...TypeFlag` has performance and memory implications (the callsites generate a new array each time, which allocates), for such a frequently called method.

Given that the callsites are usually of the form `.has_any_flag(.option, .result)`, it may be beneficial, to use a specialized .has_option_and_result instead.

spytheman:
```
#0 12:55:47 ᛋ add_has_any_flag /v/oo❱xtime ./v -o x cmd/v
CPU: 3.01s      Real: 2.60s     Elapsed: 0:02.60        RAM: 343988KB   ./v -o x cmd/v
#0 12:55:52 ᛋ add_has_any_flag /v/oo❱xtime ./vnew -o x cmd/v
CPU: 3.09s      Real: 2.66s     Elapsed: 0:02.66        RAM: 353972KB   ./vnew -o x cmd/v
#0 12:55:57 ᛋ add_has_any_flag /v/oo❱
```

shove70: geek mind 👍

felipensp:
Another option... Why not just make work `.has_flag(.option|.result)`?

spytheman:
> Another option... Why not just make work `.has_flag(.option|.result)`?

Because `.option|.result` will be an invalid enum value, since `TypeFlag` is not a flagged enum currently.

PR #20435, opened 2024-01-08
strings: use faster levenshtein distance algorithm

zeozeozeo:
This algorithm is about 2.8x faster compared to the old implementation (added @[direct_array_access], early array initialization and use a faster implementation of the 2 loops)

### DIFF
--- a/vlib/strings/similarity.v
+++ b/vlib/strings/similarity.v
@@ -1,5 +1,10 @@
 module strings
 
+@[inline]
+fn imin(x u16, y u16) u16 {
+	return if x < y { x } else { y }
+}
+
 // levenshtein_distance uses the Levenshtein Distance algorithm to calculate
 // the distance between between two strings `a` and `b` (lower is closer).
 @[direct_array_access]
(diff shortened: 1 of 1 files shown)


spytheman (approved):
Excellent work.

penguindark:
Good work!
Does it work for string longer then 32768 bytes?

zeozeozeo:
> Good work! Does it work for string longer then 32768 bytes?

yes:

```v
import strings
import rand

fn main() {
	a := rand.string(100_000) + 'x'
	b := a + 'x'
	dist := strings.levenshtein_distance(a, b)
	println("dist: ${dist}")
	assert dist == 2
}
```

penguindark:
my doubt is about this line: `mut row := []u16{len: a.len + 1, init: u16(index)}`
where the index list is of `u16` type, this mean that you can not compare more than 65536 bytes of distances.
If you need a confrontation wider than 2^16 it probably fail.

zeozeozeo:
> my doubt is about this line: `mut row := []u16{len: a.len + 1, init: u16(index)}` where the index list is of `u16` type, this mean that you can not compare more than 65536 bytes of distances. If you need a confrontation wider than 2^16 it probably fail.

yeah, this is true. might want to change it to `int` instead of `u16`

spytheman:
It was fixed in https://github.com/vlang/v/pull/20474 @sisyphus-ppcg .

PR #20438, opened 2024-01-08
net: fix non-blocking read/write

kbkpbot:
Adding set_blocking(true) to net.dial_tcp(), net.dial_tcp_with_bind(), net.accept(). 
This will ensure that the upper net apps do not need to modify.

For non-blocking usage, after create a TCPConn via net.dial_tcp(), net.dial_tcp_with_bind(), net.accept(), should use set_blocking(false) change to non-blocking mode.

Both blocking/non-blocking apps can use the same read/write now.

### DIFF
--- a/vlib/net/tcp.c.v
+++ b/vlib/net/tcp.c.v
@@ -49,11 +49,14 @@ pub fn dial_tcp(oaddress string) !&TcpConn {
 			continue
 		}
 
-		return &TcpConn{
+		mut conn := &TcpConn{
 			sock: s
 			read_timeout: net.tcp_default_read_timeout
 			write_timeout: net.tcp_default_write_timeout
+			is_blocking: true
 		}
+		conn.set_blocking(true)!
+		return conn
 	}
 
 	// Once we've failed now try and explain why we failed to connect
@@ -92,11 +95,14 @@ pub fn dial_tcp_with_bind(saddr string, laddr string) !&TcpConn {
 			continue
 		}
 
-		return &TcpConn{
+		mut conn := &TcpConn{
 			sock: s
 			read_timeout: net.tcp_default_read_timeout
 			write_timeout: net.tcp_default_write_timeout
+			is_blocking: true
 		}
+		conn.set_blocking(true)!
+		return conn
 	}
 	// failed
 	return error('dial_tcp_with_bind failed for address ${saddr}')
@@ -111,9 +117,9 @@ pub fn (mut c TcpConn) close() ! {
 
 pub fn (c TcpConn) read_ptr(buf_ptr &u8, len int) !int {
 	mut res := $if is_coroutine ? {
-		wrap_read_result(C.photon_recv(c.sock.handle, voidptr(buf_ptr), len, 0, c.read_timeout))!
+		C.photon_recv(c.sock.handle, voidptr(buf_ptr), len, 0, c.read_timeout)
 	} $else {
-		wrap_read_result(C.recv(c.sock.handle, voidptr(buf_ptr), len, 0))!
+		C.recv(c.sock.handle, voidptr(buf_ptr), len, 0)
 	}
 	$if trace_tcp ? {
 		eprintln('<<< TcpConn.read_ptr  | c.sock.handle: ${c.sock.handle} | buf_ptr: ${ptr_str(buf_ptr)} len: ${len} | res: ${res}')
@@ -127,12 +133,12 @@ pub fn (c TcpConn) read_ptr(buf_ptr &u8, len int) !int {
 		return res
 	}
 	code := error_code()
-	if code == int(error_ewouldblock) {
+	if code == int(error_ewouldblock) && c.is_blocking {
 		c.wait_for_read()!
 		res = $if is_coroutine ? {
-			wrap_read_result(C.photon_recv(c.sock.handle, voidptr(buf_ptr), len, 0, c.read_timeout))!
+			C.photon_recv(c.sock.handle, voidptr(buf_ptr), len, 0, c.read_timeout)
 		} $else {
-			wrap_read_result(C.recv(c.sock.handle, voidptr(buf_ptr), len, 0))!
+			C.recv(c.sock.handle, voidptr(buf_ptr), len, 0)
 		}
 		$if trace_tcp ? {
 			eprintln('<<< TcpConn.read_ptr  | c.sock.handle: ${c.sock.handle} | buf_ptr: ${ptr_str(buf_ptr)} len: ${len} | res: ${res}')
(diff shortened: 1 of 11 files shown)


vlib/net/tcp_test.v
```diff
@@ -76,25 +73,6 @@ fn test_tcp_ip() {
...
-		println('${address}')
-
-		mut l := net.listen_tcp(.unix, address) or { panic(err) }
-		start_echo_server(mut l)
-		echo(address) or { panic(err) }
-		l.close() or {}
-
-		os.rm(address) or { panic('failed to remove socket file') }
-	}
-}
```
spytheman, line 96: I do not want to lose that test function. It can be extracted to a new `_test.v` file though, and then that new file can be skipped on the platforms, where it is flaky.

spytheman:
> Why I can't see the full log of FreeBSD Code CI ? https://github.com/vlang/v/pull/20438/checks?check_run_id=20289394271

See it here: https://cirrus-ci.com/task/4822104995004416 (it is linked at the bottom of the CI check page).

hholst80:
I can test this later this afternoon on FreeBSD and Alpine without ipv6 and comment. I would like to avoid the call to set_blocking(true) if possible and have it be blocking by default.

vlib/net/tcp.c.v
```diff
@@ -49,11 +49,13 @@ pub fn dial_tcp(oaddress string) !&TcpConn {
 			continue
 		}
 
-		return &TcpConn{
+		mut conn := &TcpConn{
 			sock: s
 			read_timeout: net.tcp_default_read_timeout
 			write_timeout: net.tcp_default_write_timeout
 		}
+		conn.set_blocking(true)!
```
spytheman, line 57:
```suggestion
		$if !net_nonblocking_sockets ? {
			conn.set_blocking(true)!
		}
```

spytheman: `-d net_nonblocking_sockets` will not be defined by default, but will provide a way for people to opt in for the old behavior for their old projects, if needed.

vlib/net/tcp.c.v
```diff
@@ -92,11 +94,13 @@ pub fn dial_tcp_with_bind(saddr string, laddr string) !&TcpConn {
...
```
spytheman, line 102:
```suggestion
		$if !net_nonblocking_sockets ? {
			conn.set_blocking(true)!
		}
```

vlib/net/tcp.c.v
```diff
@@ -328,12 +306,13 @@ pub fn listen_tcp(family AddrFamily, saddr string, options ListenOptions) !&TcpL
 pub fn (mut l TcpListener) accept() !&TcpConn {
 	mut res := l.accept_only()!
 	res.set_sock()!
+	res.set_blocking(true)!
```
spytheman, line 309:
```suggestion
	$if !net_nonblocking_sockets ? {
		res.set_blocking(true)!
	}
```

vlib/net/tcp.c.v
kbkpbot:
```v
		$if !net_blocking_sockets {
			conn.set_blocking(true)!
		}

```
net_blocking_sockets ?

Casper64:
If the default behaviour is changed from non-blocking to blocking sockets it should add `conn.set_blocking(false)` for code that depends on non-blocking sockets. For example the `picoev` and `x.vweb` modules will break after this is merged.

vlib/net/tcp.c.v
```diff
@@ -54,7 +54,9 @@ pub fn dial_tcp(oaddress string) !&TcpConn {
 			read_timeout: net.tcp_default_read_timeout
 			write_timeout: net.tcp_default_write_timeout
 		}
-		conn.set_blocking(true)!
+		$if !net_blocking_sockets {
```
spytheman: No. Use `!net_nonblocking_sockets`, not `!net_blocking_sockets`.
The goal is for people that want the old behavior to be able to do:
`-d net_nonblocking_sockets`.

`net_blocking_sockets` is for the exact opposite result - to enforce the compilation of an app, that WILL have a blocking behavior for its connections.

vlib/net/tcp.c.v
```diff
@@ -92,11 +96,15 @@ pub fn dial_tcp_with_bind(saddr string, laddr string) !&TcpConn {
 			continue
 		}
 
-		return &TcpConn{
+		mut conn := &TcpConn{
 			sock: s
 			read_timeout: net.tcp_default_read_timeout
 			write_timeout: net.tcp_default_write_timeout
 		}
+		$if !net_blocking_sockets ? {
```
spytheman, line 104:
```suggestion
		$if !net_nonblocking_sockets ? {
```

vlib/net/tcp.c.v
```diff
@@ -328,12 +310,15 @@ pub fn listen_tcp(family AddrFamily, saddr string, options ListenOptions) !&TcpL
 pub fn (mut l TcpListener) accept() !&TcpConn {
 	mut res := l.accept_only()!
 	res.set_sock()!
+	$if !net_blocking_sockets ? {
```
spytheman, line 313:
```suggestion
	$if !net_nonblocking_sockets ? {
```

vlib/net/tcp.c.v
```diff
@@ -310,7 +310,7 @@ pub fn listen_tcp(family AddrFamily, saddr string, options ListenOptions) !&TcpL
 pub fn (mut l TcpListener) accept() !&TcpConn {
 	mut res := l.accept_only()!
 	res.set_sock()!
-	$if !net_blocking_sockets ? {
+	$if !net_nonblocking_sockets ? {
 		res.set_blocking(true)!
```
JalonSolov, line 314:
```suggestion
		res.set_blocking(false)!
```

spytheman: No, it should be `res.set_blocking(true)!` , the PR is changing the default.

spytheman:
@kbkpbot I've merged #20472 . If you rebase, I think it may fix part of the CI fail on FreeBSD.

vlib/net/tcp.c.v
JalonSolov: Doesn't make sense for `-d net_nonblocking_sockets` to call `res.set_blocking(true)`, since that makes it block... `res.set_blocking(false)` would make it not block, which would match the `-d` option.

spytheman: Note that there is inversion, it is `$if !net_nonblocking_sockets ? {`, not `$if net_nonblocking_sockets ? { .
In other words, that code inside the comptime if body, will be executed only if `-d net_nonblocking_sockets` *is not* passed, i.e. by default, which is what we want.

JalonSolov: ... ok.  I was never a fan of double negatives, but if it works, it works, I suppose.

spytheman:
(rebased over master, should solve FreeBSD failure)

PR #20443, opened 2024-01-09
ast: add clear_option_and_result() and cleanup all the related calls

yuyi98:
This PR add clear_option_and_result() and cleanup all the related calls.


spytheman (approved):
Excellent work.

PR #20455, opened 2024-01-09
x.vweb: add error when static directory does not exist

Casper64:
This pr improves error handling when the directory passed to `StaticHandler` can't be found.

It now returns an error when the static directory does not exist, instead of `false`.

### DIFF
--- a/vlib/x/vweb/static_handler.v
+++ b/vlib/x/vweb/static_handler.v
@@ -56,7 +56,7 @@ pub fn (mut sh StaticHandler) handle_static(directory_path string, root bool) !b
 // ```
 pub fn (mut sh StaticHandler) host_handle_static(host string, directory_path string, root bool) !bool {
 	if !os.exists(directory_path) {
-		return false
+		return error('directory "${directory_path}" does not exist. The directory should be relative to the current working directory')
 	}
 	dir_path := directory_path.trim_space().trim_right('/')
 	mut mount_path := ''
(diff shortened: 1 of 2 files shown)


vlib/x/vweb/static_handler.v
```diff
@@ -56,7 +56,7 @@ pub fn (mut sh StaticHandler) handle_static(directory_path string, root bool) !b
 // ```
 pub fn (mut sh StaticHandler) host_handle_static(host string, directory_path string, root bool) !bool {
 	if !os.exists(directory_path) {
-		return false
+		return error('directory "${directory_path}" does not exist. The directory should be relative to the current working directory')
```
Delta456, line 59: I think backticks are better here as they are a convention used in V too.

```suggestion
		return error('directory `${directory_path}` does not exist. The directory should be relative to the current working directory')
```

spytheman: It will be helpful for diagnosing, to put `${os.getwd()}` too after `the current working directory`.

Casper64: yes will add this

spytheman (approved):
Excellent work.

vlib/x/vweb/tests/static_handler_test.v
```diff
@@ -49,6 +49,13 @@ fn run_app_test() {
 	}
 
 	app.static_mime_types['.what'] = vweb.mime_types['.txt']
+
+	if _ := app.handle_static('not_found', true) {
+		assert true == false, 'should throw directory not found error'
```
spytheman, line 54:
```suggestion
		assert false, 'should throw directory not found error'
```

Issue #20447, opened 2024-01-09
str_intp/StrIntpData_process_str_intp_data crash in dynamic library (Windows10)


kbkpbot:
### Describe the bug

When try to println in a .dll file dynamic loaded, it crash:

```v
    name := 3
    println('hello from ${@FN} ${name}')
```
will crash in a dynamic library(.dll)
but
```v
    name := '3'
    println('hello from ${@FN} ${name}')
```
will not crash.

### Reproduction Steps

library.v
```v
module library

@[export: 'add_1']
pub fn add_1(x int, y int) int {
    name := 3
    println('hello from ${@FN} ${name}')
	return x + y
}
```

main.v
```v
module main

import os
import dl
import dl.loader

type FNAdder = fn (int, int) int

const cfolder = os.dir(@FILE)
const default_paths = [
	os.join_path(cfolder, 'library${dl.dl_ext}'),
]

fn main() {
	mut dl_loader := loader.get_or_create_dynamic_lib_loader(
		key: cfolder + '/library'
		paths: default_paths
	)!
	defer {
		dl_loader.unregister()
	}
	sym := dl_loader.get_sym('add_1')!
	f := FNAdder(sym)
	eprintln('f: ${ptr_str(f)}')
	res := f(1, 2)
	eprintln('res: ${res}')
}

```

```shell
v -o library -shared library.v
v . -o test

./test
```

### Expected Behavior

```shell
PS D:\v\D2\test\dynamic_library_loader> .\test.exe
f: 1003ccae
hello from add_1 3
res: 3
```

### Current Behavior

```shell
PS D:\v\D2\test\dynamic_library_loader> .\test.exe
f: 1003cc7d
Unhandled Exception 0xC0000005
C:/Users/DDT/AppData/Local/Temp/v_0/test.12200780638927629242.tmp.c:7996: at print_backtrace_skipping_top_frames_tcc: Backtrace
C:/Users/DDT/AppData/Local/Temp/v_0/test.12200780638927629242.tmp.c:7963: by print_backtrace_skipping_top_frames
C:/Users/DDT/AppData/Local/Temp/v_0/test.12200780638927629242.tmp.c:8773: by unhandled_exception_handler
7ffdbc668b1c : by ???
C:/Users/DDT/AppData/Local/Temp/v_0/library.17962249940916516741.tmp.so.c:8193: at i64_str: RUNTIME ERROR: invalid memory access
[...]

PR #20459, opened 2024-01-10
cgen, dl: add windows dll support, fix (#20447)

kbkpbot:
As Windows LoadLibrary/FreeLibrary do not support constructor/destructor,  add vlib/dl_windows.c.v support a workaround.
Inside the dl.open(), call the _vinit_caller().
Inside the dl.close(), call the _vcleanup_caller().
This will build the v runtime environment for the DLL.

fix (#20447 )


spytheman (approved):
Excellent work.

PR #20461, opened 2024-01-10
parser, checker: check invalid lambda expr (fix #20414)

### LINKED ISSUE #20414: Cannot use closures as a final statement in an if/else block
**V doctor:**
```
V full version: V 0.4.3 5306469.10d738c

vexe mtime: 2024-01-06 12:34:25

Git vroot status: weekly.2024.01.test2-7-g10d738c7
.git/config present: true

```

**What did you do?**
```v
fn main() {
	x := 5
	f := if x == 5 { |x| x - 5 } else { |x| x }
	println(f(x))
}

```

0 in stdout

**What did you see instead?**
```
test.v:3:19: error: a lambda expression was used, but `void` was expected
    1 | fn main() {
    2 |     x := 5
    3 |     f := if x == 5 { |x| x - 5 } else { |x| x }
      |                      ^
    4 |     println(f(x))
    5 | }
test.v:3:19: error: a lambda expression was used, but `none` was expected
    1 | fn main() {
    2 |     x := 5
    3 |     f := if x == 5 { |x| x - 5 } else { |x| x }
      |                      ^
    4 |     println(f(x))
    5 | }
test.v:3:19: error: the final expression in `if` or `match`, must have a value of a non-void type
    1 | fn main() {
    2 |     x := 5
    3 |     f := if x == 5 { |x| x - 5 } else { |x| x }
      |                      ^
    4 |     println(f(x))
    5 | }
test.v:3:4: error: assignment mismatch: 1 variable(s) 0 value(s)
    1 | fn main() {
    2 |     x := 5
    3 |     f := if x == 5 { |x| x - 5 } else { |x| x }
      |       ~~
    4 |     println(f(x))
    5 | }
test.v:3:19: error: a lambda expression was used, but `string` was expected
    1 | fn main() {
    2 |     x := 5
    3 |     f := if x == 5 { |x| x - 5 } else { |x| x }
      |                      ^
    4 |     println(f(x))
    5 | }
test.v:4:10: error: unknown function: f
    2 |     x := 5
    3 |     f := if x == 5 { |x| x - 5 } else { |x| x }
    4 |     println(f(x))
      |             ~~~~
    5 | }
test.v:4:2: error: `println` can not print void expressions
    2 |     x := 5
    3 |     f := if x == 5 { |x| x - 5 } else { |x| x }
    4 |     println(f(x))
      |     ~~~~~~~~~~~~~
    5 | }
If the code of your project is in multiple files, try with `v .` instead of `v test.v`
```

Example of this expression in Rust: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=60a8125c5277fb69cf22f9272f60aba7
> [!NOTE]
>

yuyi98:
This PR check invalid lambda expr (fix #20414).

- Check invalid lambda expr.
- Add test.

```v
fn main() {
	x := 5
	f := if x == 5 { |x| x - 5 } else { |x| x }
	println(f(x))
}

PS D:\Test\v\tt1> v run .
tt1.v:3:19: error: lambda expressions are allowed only in places expecting function callbacks
    1 | fn main() {
    2 |     x := 5
    3 |     f := if x == 5 { |x| x - 5 } else { |x| x }
      |                      ~~~~~~~~~
    4 |     println(f(x))
    5 | }
```

### DIFF
--- a/vlib/v/checker/errors.v
+++ b/vlib/v/checker/errors.v
@@ -41,6 +41,16 @@ fn (mut c Checker) error(message string, pos token.Pos) {
 	c.warn_or_error(msg, pos, false)
 }
 
+fn (mut c Checker) fatal(message string, pos token.Pos) {
+	if (c.pref.translated || c.file.is_translated) && message.starts_with('mismatched types') {
+		// TODO move this
+		return
+	}
+	msg := message.replace('`Array_', '`[]')
+	c.pref.fatal_errors = true
+	c.warn_or_error(msg, pos, false)
+}
+
 fn (mut c Checker) note(message string, pos token.Pos) {
 	if c.pref.message_limit >= 0 && c.nr_notices >= c.pref.message_limit {
 		c.should_abort = true
--- a/vlib/v/checker/lambda_expr.v
+++ b/vlib/v/checker/lambda_expr.v
@@ -7,8 +7,8 @@ pub fn (mut c Checker) lambda_expr(mut node ast.LambdaExpr, exp_typ ast.Type) as
 	if node.is_checked {
 		return node.typ
 	}
-	if exp_typ == 0 {
-		c.error('lambda expressions are allowed only in places expecting function callbacks',
+	if exp_typ in [0, ast.void_type] {
+		c.fatal('lambda expressions are allowed only in places expecting function callbacks',
 			node.pos)
 		return ast.void_type
 	}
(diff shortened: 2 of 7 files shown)


spytheman (approved):
Excellent work.

zeozeozeo:
I think this should be allowed in the future, lambdas should act as functions.

For example, this is allowed:

```v
import term

fn print_colored(is_red bool, text string) {
    color := if is_red { term.red } else { term.white }
    print(color(text))
}
```

but this is not:

```v
import term

fn print_colored(is_red bool, text string) {
    color := if is_red { |text| term.red(text) } else { |text| term.white(text) }
    print(color(text))
}
```

PR #20468, opened 2024-01-10
vlib: add new Dynamic Template Manager (DTM) module

GGRei:
This pull request introduces a **Dynamic Template Manager (DTM)** to the Vweb module, add capability to manage HTML templates dynamically. The DTM allows for direct modifications to HTML templates without the need for recompiling the application.

_Key Features:_

- Dynamic Template Management: Facilitates real-time editing and updating of HTML templates.
- Integrated Cache Manager: Optimizes performance by managing the caching of processed templates.

_Tests Included:_

- Internal Test: Evaluate the DTM functionality itself.
- External Test: Look the DTM in action through a minimalistic web application that dynamically displays simulated data.

The codebase for the Dynamic Template Manager is documented, offering insights into its functionality and usage. 

In developing the DTM, I focused on maintaining simplicity and adhering to the Vlang principles and of the Vweb module. Exemple, this involved keeping HTML files in a 'templates' directory at the project's root.

In the code, there is a TODO list, but I personally do not consider this work to be blocking. However, I leave it to the core team to determine its significance.
I am available for any requests for modifications to the project/code.

If, however, this does not interest the Vlang core team, or does not align with their plans for including dynamic systems in Vweb (preferring an approach closer to the Vlang compiler than to the Vweb module itself), that's okay. I might consider proposing it as an extension module instead.

If accepted, I am donating this work to the Vlang project, which is why there is no mention of a license in the code.

Regards,


spytheman:
I think that is a very good addition, but with one change - moving it to a separate module, like for example `vlib/vweb/dtm`.
The reason is that, if it is part of the `vweb` module directly, it will be compiled in for every vweb app, even if the app does not use it.

Having it as an independent separate module, will also make it possible for other apps, outside of vweb, to benefit from it, for example static website generators.

Looking at the code, that move will only require removing the .dtm field from the Context, and making some of the new Context methods, normal functions that accept `mut ctx Context` instead.

JalonSolov:
Agreed.  Templates aren't always used (or even wanted), so forcing every vweb app to have the support just makes the default larger than it needs to be.

Other than that, this feature has been requested more than once in the past, so it will be a good addition.

spytheman:
> If you confirm that this consideration is not a concern

Yes, it is not a concern, go ahead. 

The need for the `@[vweb_global]` attribute, is more of a legacy hack at this point, needed due to the design of `vweb`, that is fixed in `x.vweb`.

Mentioning `x.vweb`, I think, that a good place will be `x.templating.dtm`, i.e. `vlib/templating/dtm/`, which is independent from `vweb` and `x.vweb`, and everyone will be able to `import x.templating.dtm` .

In time, when it is more feature complete, it can be moved out of `x.templating` to a more permanent location outside `vlib/x`.

GGRei:
I've made the changes you requested, and the Dynamic Template Manager (DTM) is now a module. 

To maintain DTM's filter functionality, I copied the filter function from x.vweb to the templating directory, making it public there. If this approach isn't suitable, please let me know, and I'll make the necessary adjustments.

I would like to add, however, that for the moment, it is intended for use with vweb exclusively.

spytheman:
I think, that it should be self contained, i.e. move vlib/x/templating/escape_html_strings_in_templates.v into `vlib/x/templating/dtm` too.

Calls to the filter function are generated by the V compiler, in the case of vweb, since it needs to do everything at compile time. In DTM's case, the calls can be done at runtime, and to me it makes no sense for that function to be separate.

GGRei:
The changes have been implemented. If there are any more suggestions or modifications needed, please let me know.

Thanks!

vlib/vweb/tests/dynamic_template_manager_test_server/dynamic_template_manager_test_server.v
```diff
@@ -0,0 +1,62 @@
+import vweb
+import time
+import x.templating.dtm
+
+struct App {
+	vweb.Context
+pub mut:
+	dtm                &dtm.DynamicTemplateManager = dtm.create_dtm() @[vweb_global]
+	shared_data_string []string = []string{}                    @[vweb_global]
+	shared_data_int    []int    = []int{}                       @[vweb_global]
```
spytheman, line 10:
```suggestion
	shared_data_string []string @[vweb_global]
	shared_data_int    []int @[vweb_global]
```

vlib/x/templating/dtm/dynamic_template_manager_test.v
```diff
@@ -0,0 +1,402 @@
...
+const temp_dtm_dir = 'dynamic_template_manager_test'
+const temp_cache_dir = 'vcache'
+const temp_templates_dir = 'templates'
+const temp_html_fp = 'temp.html'
+const temp_html_n = 'temp'
+const vtmp_dir = os.vtmp_dir()
+
+fn testsuite_begin() {
+	temp_folder := os.join_path(dtm.vtmp_dir, dtm.temp_dtm_dir)
+	os.mkdir_all(temp_folder) or { panic(err) }
```
spytheman, line 15:
```suggestion
	os.mkdir_all(temp_folder)!
```

spytheman: propagation with ! works in all test functions, and it is implemented as `or {panic(err)}`, similar to propagation in `fn main() {`

vlib/x/templating/dtm/dynamic_template_manager.v
```diff
@@ -0,0 +1,1136 @@
...
+	max_size_data_in_mem int  = dtm.max_size_data_in_memory
+	test_cache_dir       string
+	test_template_dir    string
+}
+
+// pub fn create_dtm() return &DynamicTemplateManager
+//
+// Initializes and returns a reference to a new instance of DynamicTemplateManager. This instance is allocated on the heap.
+//
+pub fn create_dtm() &DynamicTemplateManager {
```
spytheman, line 148: The documentation tool `v doc` needs a comment above each public function, starting with
`// <name_of_the_fn> ` and continuing with the purpose/goal of it.

In this case, that would be something like:
```v
// create_dtm initializes and returns a reference to a new instance of DynamicTemplateManager on the heap.
```

vlib/x/templating/dtm/dynamic_template_manager.v
```diff
@@ -0,0 +1,1136 @@
...
+// to substitute placeholders with their corresponding values defined in the map.
+@[params]
+pub struct TemplateCacheParams {
+	placeholders           &map[string]DtmMultiTypeMap = &map[string]DtmMultiTypeMap{}
+	cache_delay_expiration i64 = dtm.cache_delay_expiration_by_default
+}
+
+// These parameters are used with 'initialize_dtm' function. (See below at initialize_dtm comment)
+@[params]
+pub struct DynamicTemplateManagerInitialisationParams {
```
spytheman, line 136:
```suggestion
// DynamicTemplateManagerInitialisationParams is used with 'initialize_dtm' function. (See below at initialize_dtm comment)
pub struct DynamicTemplateManagerInitialisationParams {
```

vlib/x/templating/dtm/dynamic_template_manager.v
spytheman: The same applies to all public symbols, like constants, structs etc.

vlib/x/templating/dtm/dynamic_template_manager.v
```diff
@@ -0,0 +1,1136 @@
...
+// The Dynamic Template Manager (DTM) requires this initialization function to be operational.
+// A "vcache" directory must be created at the root of the project (where the executable is located) to use the DTM.
+// Initalisation params are :
+// - max_size_data_in_mem 'type int' Maximum size of data allowed in memory for caching. The value must be specified in kilobytes. ( Default is: 500KB / Limit max is : 500KB)	
+// - compress_html: 'type bool' Light compress of the HTML ouput. ( default is true )
+// - active_cache_server: 'type bool' Activate or not the template cache system. ( default is true )
+// - test_cache_dir: 'type string' Used only for DTM test file, parameter is ignored otherwise.
+// - test_template_dir: 'type string' Used only for DTM test file, parameter is ignored otherwise.
+//
+// vfmt off
```
spytheman, line 164: Why is there a need to turn off vfmt for just the function declaration?

vlib/x/templating/dtm/dynamic_template_manager.v
```diff
@@ -0,0 +1,1136 @@
...
+        if request_path.contains(cache_dir) {
+            ctx.text('Access to the cache directory is not allowed.')
+            return false
+        }
+        return true
+    }
+}
+*/
+
+// pub fn (mut tm DynamicTemplateManager) serve_dynamic_template(string, TemplateCacheParams) return string
```
spytheman, line 232: // serve_dynamic_template

vlib/x/templating/dtm/dynamic_template_manager.v
GGRei: I used the 'vfmt off' directive to preserve the '&' in the parameter `mut dtmref &DynamicTemplateManager`. I suppose the V compiler try at first to passes parameters by reference and that vfmt might be considering the explicit '**&**' unnecessary. 

However, I find that including enhances the code's readability and clarifies its intention as a reference. Nevertheless, I can let vfmt format this if you prefer.

GGRei:
New commit.

vlib/x/templating/dtm/dynamic_template_manager.v
spytheman: It is ok as it is. Currently the compiler just ignores the `&` in `mut x &Type`, but soon it will be needed, for exactly the reason you are using it. I was just curious about the vfmt tags, since I have not seen it used that way before.

spytheman (approved):

JalonSolov:
Now we just need documentation on how to use it...

medvednikov:
Great work @GGRei !

GGRei:
@medvednikov Thanks you! And thank you for this fantastic language that is V!

Thanks at @spytheman too!

PR #20470, opened 2024-01-10
Small enchancement to the core functionality of ChaCha20 stream cipher

blackshirt:
This pull request adds some minor enhancement to the `xor_key_stream` and other core functionality, especially in speed of execution.
Local bench on my low end laptop shows up small nice improvement, like i paste on [discord](https://discord.com/channels/592103645835821068/592294828432424960/1194488820540899338)
Before patched
```vlang
$ v run bench.v
 SPENT  7206.581 ms in chacha20.encrypt 40 bytes of plaintext with 500000 iterations, total bytes encrypted : 20000000 bytes
 SPENT  8056.834 ms in chacha20.encrypt 64 bytes of plaintext with 500000 iterations, total bytes encrypted : 32000000 bytes
 SPENT 11103.743 ms in chacha20.encrypt 128 bytes of plaintext with 500000 iterations, total bytes encrypted : 64000000 bytes
 SPENT 22957.365 ms in chacha20.encrypt 200 bytes of plaintext with 500000 iterations, total bytes encrypted : 100000000 bytes
```
And after being patched
```vlang
$ v run bench.v
 SPENT  5193.056 ms in chacha20.encrypt 40 bytes of plaintext with 500000 iterations, total bytes encrypted : 20000000 bytes
 SPENT  5925.972 ms in chacha20.encrypt 64 bytes of plaintext with 500000 iterations, total bytes encrypted : 32000000 bytes
 SPENT 11119.856 ms in chacha20.encrypt 128 bytes of plaintext with 500000 iterations, total bytes encrypted : 64000000 bytes
 SPENT 19557.270 ms in chacha20.encrypt 200 bytes of plaintext with 500000 iterations, total bytes encrypted : 100000000 bytes
```
This PR adds some changes by :
- Adds `@[direct_array_access]` attributes to the core function that likely needed
- Reduces allocation of output buffer by using dst buffer directly
- Use builtin `vmemset` to reset cipher's key and nonce
- Some other minor changes

Please gives a review, comment and feedback.
Best regards


vlib/x/crypto/chacha20/chacha_test.v
```diff
@@ -1,7 +1,24 @@
 module chacha20
 
+import crypto.rand
```
spytheman, line 3: Use `rand` for tests. It is much faster, and it will not deplete the entropy on the CI runners.

spytheman: Then `key := rand.bytes(32)!` will be the same as `key := rand.read(32)!`

spytheman (approved):
Excellent work.

vlib/x/crypto/chacha20/chacha_test.v
blackshirt: good suggestion...i would change it

PR #20472, opened 2024-01-10
net.listen_tcp af .unix expect to return error

hholst80:
The vlib listen_tcp API only allows IP based address family.
The UNIX domain stream- and datagram is tested in
net/unix/unix_tests.v.


hholst80:
Ping @kbkpbot @spytheman for PR #20438

PR #20482, opened 2024-01-11
sync: add mutex.try*lock functions for FreeBSD too

bakul:
mutex: add try*lock functions for freebsd. This fixes the vlib/sync/mutex_test.v test.


shove70:
The CI failure should be unrelated, so I restarted the failed CI tasks.

PR #20484, opened 2024-01-11
log: enhance log time format setting

kbkpbot:
enhance log time format setting
support most time format in time module, and custom time format also.


spytheman (approved):
Excellent enhancement.

PR #20485, opened 2024-01-11
parser: fix close_scope() missing when field.name is sort, sorted(fix#20436)

### LINKED ISSUE #20436: If use `sort_with_compare`, it will report an error
### Describe the bug

```vlang

pub fn (mut api App) api_account_position() vweb.Result {
    mut postions := []AccountPosition{} // some items

    for i, mut postion in postions {
		postion.recid = postion.inst_id
		if postion.mgn_ratio.f64() * 100 <= 300 {
			postion.w2ui.style = 'background-color: #f3d6e3'
		} else if postion.upl_ratio.f64() * 100 < -50 {
			postion.w2ui.style = 'background-color: rgb(242, 250, 140)'
		} else {
			mut ratio := 0
			if postion.lever.f64() >= 20 {
				ratio = 50
			} else if postion.lever.f64() >= 10 {
				ratio = 30
			} else if postion.lever.f64() >= 3 {
				ratio = 20
			}
			if postion.upl_ratio.f64() * 100 > ratio {
				postion.w2ui.style = 'background-color: #C2F5B4'
			}
		}
		postions[i] = postion
	}

	request := api.query['request'] or { '{}' }
	dto := json.decode(RequestDto, request) or { RequestDto{} }
      // If you use sort_with_compare, it will report an error on the last line
	postions.sort_with_compare(fn [dto] (a &AccountPosition, b &AccountPosition) int {
		if dto.sort.len == 0 {
			return 0
		}

		mut left := ''
		mut right := ''
		left, right = match dto.sort[0].field {
			'uplRatio' {
				a.upl_ratio, b.upl_ratio
			}
			'upl' {
				a.upl, b.upl
			}
			else {
				'', ''
			}
		}
		if left == right {
			return 0
		}

		if dto.sort[0].direction == 'desc' {
			return if left < right { -1 } else { 1 }
		} else {
			return if left > right { -1 } else { 1 }
		}
	})

	return api.json(postions) <- here 
}
```

image

> It will work properly after the annotation sorting function is implemented

image

> Is it a bug or my usage error? How can I fix this problem?

### Reproduction Steps

--

### Expected Behavior

success compile

### Current Behavior

failed

### Possible Solution

_No response_

_No response_

### V version

V 0.4.3 e2334d8

```
V full version: V 0.4.3 f705897.e2334d8
OS: macos, macOS, 13.5.1, 22G90
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz

getwd: /Users/xiusin/projects/
vexe: /opt/v/v
vexe mtime: 2024-01-05 10:50:47

vroot: OK, value: /opt/v
VMODULES: OK, value: /Users/xiusin/.vmodules
VTMP: OK, value: /tmp/v_501

Git version: git version 2.23.0
Git vroot status: weekly.2023.50-176-ge2334d8b
.git/config present: true

CC version: Apple clang version 14.0.0 (clang-1400.0.29.202)
thirdparty/tcc status: thirdparty-macos-amd64 46662e20-dirty
```
> [!NOTE]
>

shove70:
1. Fixed #20436
2. Add tests.

```v
struct Foo {
	sort []int
}

fn main() {
	f := Foo{}

	fn [f] () {
		println(f.sort.len)
		assert f.sort.len == 0
	}()
}
```
outputs:
```
0
```

### DIFF
--- a/vlib/v/parser/parser.v
+++ b/vlib/v/parser/parser.v
@@ -3205,6 +3205,9 @@ fn (mut p Parser) dot_expr(left ast.Expr) ast.Expr {
 	is_filter := field_name in ['filter', 'map', 'any', 'all']
 	if is_filter || field_name == 'sort' || field_name == 'sorted' {
 		p.open_scope()
+		defer {
+			p.close_scope()
+		}
 	}
 	// ! in mutable methods
 	if p.tok.kind == .not && p.peek_tok.kind == .lpar {
@@ -3267,9 +3270,6 @@ fn (mut p Parser) dot_expr(left ast.Expr) ast.Expr {
 			scope: p.scope
 			comments: comments
 		}
-		if is_filter || field_name == 'sort' || field_name == 'sorted' {
-			p.close_scope()
-		}
 		return mcall_expr
 	}
 	mut is_mut := false
@@ -3315,9 +3315,6 @@ fn (mut p Parser) dot_expr(left ast.Expr) ast.Expr {
 		next_token: p.tok.kind
 	}
 
-	if is_filter || field_name == 'sort' || field_name == 'sorted' {
-		p.close_scope()
-	}
 	return sel_expr
 }
 
--- a/vlib/v/parser/parser.v
+++ b/vlib/v/parser/parser.v
@@ -3315,7 +3315,7 @@ fn (mut p Parser) dot_expr(left ast.Expr) ast.Expr {
 		next_token: p.tok.kind
 	}
 
-	if is_filter {
+	if is_filter || field_name == 'sort' || field_name == 'sorted' {
 		p.close_scope()
 	}
 	return sel_expr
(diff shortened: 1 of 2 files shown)


vlib/v/parser/parser.v
```diff
@@ -3315,7 +3315,7 @@ fn (mut p Parser) dot_expr(left ast.Expr) ast.Expr {
 		next_token: p.tok.kind
 	}
 
-	if is_filter {
+	if is_filter || field_name == 'sort' || field_name == 'sorted' {
 		p.close_scope()
 	}
```
spytheman, line 3320: I think that using `defer { p.close_scope() }`, at line 3201, right after the `p.open_scope()` call, will be more robust.

Then both these 3 lines, and the ones before the other return branch in 3270-3272, can be removed.

spytheman (approved):
Good work.

vlib/v/parser/parser.v
shove70: hmm, right away.

shove70: Good suggestion 👍 

done.   :)

spytheman: Amazing speed! Thank you.

PR #20487, opened 2024-01-11
examples: fix "call v from ruby" example for Windows and Mac

omnikron:
I noticed that the "call v from ruby" example doesn't work on MacOS, and likely also not on Windows, since the shared library artifact generated by `v -d no_backtrace -shared examples/call_v_from_ruby/test.v` has a different, platform-specific extension (`.dylib` and `.dll` respectively - [source](https://github.com/vlang/v/blob/757929392e0e7a75fc1272116460981e589737d5/vlib/dl/dl.v#L9-L20)) but the ruby code assumed `.so`.  

While I was there, I also improved the user feedback from the Ruby script if the shared library is missing.

- [x] Before submitting a PR, please run `v test-all`
  - Quite a few failures on my machine, but this also happens on `master` and I'm only changing ruby code, so I hope this is OK.

### DIFF
--- a/examples/call_v_from_ruby/test.rb
+++ b/examples/call_v_from_ruby/test.rb
@@ -22,7 +22,11 @@ end
 module Lib
   extend FFI::Library
 
-  ffi_lib File.join(File.dirname(__FILE__), 'test' + shared_library_extension)
+  begin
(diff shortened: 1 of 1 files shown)


spytheman (approved):
Excellent work.

ylluminate:
Thank you.

PR #20495, opened 2024-01-11
os: make os.cache_dir() and os.vtmp_dir() more robust to parallel test executions, extract common code

spytheman:
Intended to fix failures similar to this one:
https://github.com/vlang/v/actions/runs/7487373820/job/20387668781
```
 FAIL  [ 153/1701]  2033.642 ms vlib/flag/usage_example_test.v
V panic: mkdir failed for "C:\Users\runneradmin\.cache", because CreateDirectory 
returned: Cannot create a file when that file already exists.
```

### DIFF
--- a/vlib/os/os.v
+++ b/vlib/os/os.v
@@ -738,6 +738,21 @@ pub fn mkdir_all(opath string, params MkdirParams) ! {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


JalonSolov:
Seems like an odd error message, to see "cannot create a file" when you called CreateDirectory.  Is that coming from Windows?

PR #20496, opened 2024-01-11
repl: support executing shell commands on REPL with `!sh [COMMAND]`

viniciusfdasilva:
### Implementation code change on vrepl.v

```v
fn repl_help() {
	println(version.full_v_version(false))
	println('
	|help                   Displays this information.
	|list                   Show the program so far.
	|reset                  Clears the accumulated program, so you can start a fresh.
	|Ctrl-C, Ctrl-D, exit   Exits the REPL.
	|clear                  Clears the screen.
	|pin                    Pins the entered program to the top.
	|!sh [COMMAND]          Execute on REPL shell commands.
'.strip_margin())
}
```

```v
fn run_shell(command string) {
	if command.len >= 2 && command[0..2] == 'cd' {
		command_splited := command.split(' ')
		assert command_splited.len >= 2
		dir := command_splited[command_splited.len - 1]

		os.chdir(dir) or { eprintln('`${command}` failed, err: ${err}') }
	} else {
		os.system(command)
	}
}
```

```v
if r.line.len > 4 && r.line[0..3] == '!sh' {
	run_shell(r.line[4..r.line.len])
	continue
}
```

### Implementation working

repl

### v fmt -w cmd/tools/vrepl.v Execution result

file

### DIFF
--- a/cmd/tools/vrepl.v
+++ b/cmd/tools/vrepl.v
@@ -100,7 +100,21 @@ fn repl_help() {
 	|Ctrl-C, Ctrl-D, exit   Exits the REPL.
 	|clear                  Clears the screen.
 	|pin                    Pins the entered program to the top.
+	|!sh [COMMAND]          Execute on REPL shell commands.
 '.strip_margin())
+}
+
+fn get_shell(command string) {
+	if command.contains('cd') {
+		
+		command_splited := command.split(' ')
+		assert command_splited.len >= 2
+		dir := command_splited[command_splited.len-1]
+		
+		os.chdir(dir) or { panic('Switch directory failed!') }
+	}else{
+		println(os.execute('timeout 1s ${command}').output)
+	}
 }
 
 fn (mut r Repl) checks() bool {
@@ -361,6 +375,12 @@ fn run_repl(workdir string, vrepl_prefix string) int {
 			repl_help()
 			continue
 		}
+
+		if r.line.len > 4 && r.line[0..3] == '!sh' {
+			get_shell(r.line[4..r.line.len])
+			continue
+		}
+		
 		if r.line.contains(':=') && r.line.contains('fn(') {
 			r.in_func = true
 			r.functions_name << r.line.all_before(':= fn(').trim_space()
--- a/cmd/tools/vrepl.v
+++ b/cmd/tools/vrepl.v
@@ -110,7 +110,7 @@ fn run_shell(command string) {
 		assert command_splited.len >= 2
 		dir := command_splited[command_splited.len - 1]
 
-		os.chdir(dir) or { println('Switch directory failed!') }
+		os.chdir(dir) or { eprintln('`${command}` failed, err: ${err}') }
 	} else {
 		os.system(command)
 	}
--- a/cmd/tools/vrepl.v
+++ b/cmd/tools/vrepl.v
@@ -104,16 +104,15 @@ fn repl_help() {
 '.strip_margin())
 }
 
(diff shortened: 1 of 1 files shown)


Delta456:
I don't understand the need for this. One can always import `os` and use the `system` function to achieve the same functionality.

viniciusfdasilva:
Well, its more easy type direct shell commands than call every time os library functions!

spytheman:
@Delta456 It is a good feature. `ipython` for example has `!` for this, try `! ls -la` for example or `! df /`.

spytheman:
@viniciusfdasilva please do `v fmt -w cmd/tools/vrepl.v` and push the result too.

cmd/tools/vrepl.v
```diff
@@ -100,9 +100,23 @@ fn repl_help() {
...
+fn get_shell(command string) {
+	if command.contains('cd') {
+		
+		command_splited := command.split(' ')
+		assert command_splited.len >= 2
+		dir := command_splited[command_splited.len-1]
+		
+		os.chdir(dir) or { panic('Switch directory failed!') }
+	}else{
+		println(os.execute('timeout 1s ${command}').output)
```
spytheman, line 116: Why the timeout? It is not portable afaik (windows), and some commands can legitimately take more than a second.
Imho just use `os.system(command)` . It is better in the sense that you will get immediate results in your terminal, without having to wait for the whole command to complete, and then print its output.

cmd/tools/vrepl.v
```diff
@@ -100,9 +100,23 @@ fn repl_help() {
 	|Ctrl-C, Ctrl-D, exit   Exits the REPL.
 	|clear                  Clears the screen.
 	|pin                    Pins the entered program to the top.
+	|!sh [COMMAND]          Execute on REPL shell commands.
 '.strip_margin())
 }
 
+fn get_shell(command string) {
```
spytheman, line 107:
```suggestion
fn shell_cmd(command string) {
```

spytheman: Personally, I associate `get_` with something that retrieves some immutable information, but this does arbitrary things, depending on the command, not just retrieve information.

cmd/tools/vrepl.v
```diff
@@ -100,9 +100,23 @@ fn repl_help() {
...
+	if command.contains('cd') {
+		
+		command_splited := command.split(' ')
+		assert command_splited.len >= 2
+		dir := command_splited[command_splited.len-1]
+		
+		os.chdir(dir) or { panic('Switch directory failed!') }
```
spytheman, line 114:
```suggestion
		os.chdir(dir) or { eprintln('Switch directory failed, err: $err') }
```

JalonSolov: Definitely.  It's bad to panic out of a REPL.

cmd/tools/vrepl.v
viniciusfdasilva: I added the timeout to stop long-running executions like ping! However, you are right! I will remove the timeout!

cmd/tools/vrepl.v
viniciusfdasilva: Perfect! I will modify that as well! Any suggestion?

cmd/tools/vrepl.v
viniciusfdasilva: No problem! I will modify as well!

spytheman (approved):
Excellent work. 
Thank you for your first PR.

PR #20502, opened 2024-01-12
parser, cgen: a minor cleanup

shove70:
This PR does some minor cleanup work based on the comments of several previous PRs.

- Moved some `close_scope()` into defer (parser).
- Get fixed array size from sym.info.size (cgen, suggestions from previous PR)


spytheman (approved):
Excellent work 🥳 !

PR #20503, opened 2024-01-12
checker: cleanup the const variable evaluate for fixed array fields of structs

shove70:
This PR:

- The fixed array size evaluation logic for the struct fields did not take into account the multidimensional case, so the code has been cleaned up and adjusted to be consistent with the fixed array size evaluation logic in the `containers.v`.
- `fn array_fixed_has_unresolved_size()` is added to avoid unwanted evaluation.
- The related test cases were modified.


spytheman (approved):
Excellent work 🚀 🥳 ❣️

PR #20515, opened 2024-01-13
checker: multi return should check option is handled

### LINKED ISSUE #17874: the checker allows for `a, b := abc()` where `fn abc() ?(int,int)`, but it should require the `none` to be handled instead
### Describe the bug

There are some situation where `Option` type does not handled correctly,  When some function (methods) returns `Option` type and called without proper error propagation, its result in `C error` compiler bug.

### Expected Behavior

error propagation should be handled correctly, similar to `Result` type ..minimal looks like this >
```
code.v:9:10: error: abc() returns an option, so it should have either an `or {}` block, or `?` at the end
    7 | 
    8 | fn main() {
    9 |     a, b := abc(10)
      |             ~~~~~~~
   10 |     dump(a)
   11 |     dump(b)
Exited with error status 1
```

### Current Behavior

```

Exited with error status 1
```

### Reproduction Steps

Use this snippet :
```
fn abc(n int) ?(int,int) {
    if n == 5 {
        return 5, 10
    }
    return none
}

fn main() {
    // call function without propagate the error
    a, b := abc(10)
    dump(a)
    dump(b)
}
```

### Possible Solution

I dont know exactly, but, maybe in syntax checker

_No response_

### V version

```V 0.3.3 6756d28```

```
V full version: V 0.3.3 6756d28
OS: linux, Ubuntu 20.04.5 LTS (VM)
Processor: 2 cpus, 64bit, little endian, Intel(R) Xeon(R) Platinum 8272CL CPU @ 2.60GHz

getwd: /workspaces/asn1
vexe: /home/codespace/v/v
vexe mtime: 2023-03-29 02:16:58

vroot: OK, value: /home/codespace/v
VMODULES: OK, value: /home/codespace/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.40.0
Git vroot status: weekly.2023.13-9-g6756d285 (35 commit(s) behind V master)
.git/config present: true

```

### PR DESCRIPTION
**Pull request by steven-mathew**
Fixes #17874

This fix seems to have uncovered a bug in [vlib/orm/orm.v](https://github.com/vlang/v/blob/6c016e51b77c477051cdaa900f4548a95d03b70b/vlib/orm/orm.v#L511C9-L511C9) where the option is not handled which I've gone ahead and fixed.

### DIFF
--- a/vlib/orm/orm.v
+++ b/vlib/orm/orm.v
@@ -505,12 +505,13 @@ pub fn orm_table_gen(table string, q string, defaults bool, def_unique_len int, 
 						}
 					} else {
 						if attr.arg.trim(' ') == '' {
-							return error("references attribute needs to be in the format [references], [references: 'tablename'], or [references: 'tablename(field_id)']")
+							return error("references attribute needs to be in the format @[references], @[references: 'tablename'], or @[references: 'tablename(field_id)']")
 						}
 						if attr.arg.contains('(') {
-							ref_table, ref_field := attr.arg.split_once('(')
+							err := error("explicit references attribute should be written as @[references: 'tablename(field_id)']")
+							ref_table, ref_field := attr.arg.split_once('(') or { return err }
 							if !ref_field.ends_with(')') {
-								return error("explicit references attribute should be written as [references: 'tablename(field_id)']")
+								return err
 							}
 							references_table = ref_table
 							references_field = ref_field[..ref_field.len - 1]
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: vlib/orm/orm.v [hunk 1]
```diff
@@ -508,9 +508,10 @@ pub fn orm_table_gen(table string, q string, defaults bool, def_unique_len int,
 							return error("references attribute needs to be in the format [references], [references: 'tablename'], or [references: 'tablename(field_id)']")
 						}
 						if attr.arg.contains('(') {
-							ref_table, ref_field := attr.arg.split_once('(')
+							err := error("explicit references attribute should be written as [references: 'tablename(field_id)']")
```
StunxFS, line 511:
```suggestion
							err := error("explicit references attribute should be written as @[references: 'tablename(field_id)']")
```

I think the other error messages need to be updated to the new attribute syntax

steven-mathew: Fixed! I also adjusted the other error messages as well for the new attribute syntax.

JalonSolov:
Handled in #21106 which has been merged.

PR #20519, opened 2024-01-13
example: change_color_by_id

enghitalo:
### Why?
- Smaller than mandelbrot
- Better string manipulation
- HTML elements manipulation (Js using V function)

### Discussion
Would be cool have some template processing when using wasm

`@v 'change_color_by_id.wasm.v'` Would compile `change_color_by_id.wasm.v` file and generate the actual script with all public function 

Might following the logic bellow

```
     @for func in v_functions
            @func.name = res.instance.exports["${func.mod_name}.${func.name}"];
      @end
```
like

```v
reflection
    .get_funcs()
    .filter(it.mod_name == "main")
    .map(fn (f reflection.Function) string {
		return "${f.mod_name}.${f.name}"
	})
```


spytheman:

spytheman (approved):
Excellent example.

PR #20520, opened 2024-01-13
tests: show compile time and runtime for each _test.v file in separate columns

spytheman:
- benchmark: support passing an optional preparation time for the step messages
- pref: support a -skip-running option
- testing: add `compile_begin` and `compile_end` messages to the testing.Reporter interface
- testing: support the new messages in the output_normal.v Reporter
- benchmark: add Benchmark.step_restart/0 method
- tools: fix `v test` to not skip _test.c.v files by default
- testing,benchmark: use the new -skip-running and the enhanced Reporter to implement a separate compilation time measurement for each test

### DIFF
--- a/cmd/tools/modules/testing/common.v
+++ b/cmd/tools/modules/testing/common.v
@@ -137,7 +137,7 @@ pub fn (mut ts TestSession) print_messages() {
 		// first sent *all events* to the output reporter, so it can then process them however it wants:
 		ts.reporter.report(ts.nmessage_idx, rmessage)
 
-		if rmessage.kind in [.cmd_begin, .cmd_end] {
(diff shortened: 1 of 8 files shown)


spytheman:
The output for `v test vlib/builtin/` will look like this:
image

felipensp:
Nice improvement.

spytheman:
> Nice improvement.

thank you

shove70:
wow, It's beautiful 👍

ttytm:
@spytheman since this commit the `-stats` flag won't include prints in the test output.

For example, if you run one of the vpm tests without `-d network` but with `-stats` the tests run without information such as the following, which used to be provided.

```
> skipping /Users/t/Nextcloud/Dev/vlang/v/cmd/tools/vpm/remove_test.v, when `-d network` is missing
```

PR #20521, opened 2024-01-13
Add field shadowing comment

lazalong:


doc/docs.md
```diff
@@ -480,7 +480,24 @@ fn main() {
 	}
 }
 ```
+While variable shadowing is not allowed, field shadowing is allowed. 
+```v failcompile nofmt
```
spytheman, line 484:
```suggestion
```v nofmt
```
or even better, remove the `nofmt` too, then run `VAUTOFIX=1 ./v check-md -hide-warning doc/docs.md` 2 times, and commit the result.

PR #20522, opened 2024-01-14
refactor: JS create_sourcemap

enghitalo:
type and better use of json2 encode

### DIFF
--- a/vlib/v/gen/js/sourcemap/mappings.v
+++ b/vlib/v/gen/js/sourcemap/mappings.v
@@ -3,7 +3,7 @@ module sourcemap
 import v.gen.js.sourcemap.vlq
 import io
 
-struct Empty {}
+pub struct Empty {}
 
 pub struct SourcePosition {
 	source_line   u32
@@ -11,7 +11,7 @@ pub struct SourcePosition {
 }
 
 type IndexNumber = u32
-type SourcePositionType = Empty | SourcePosition
+pub type SourcePositionType = Empty | SourcePosition
 type NameIndexType = Empty | IndexNumber
 
 struct GenPosition {
@@ -25,7 +25,7 @@ struct MappingInput {
 	source_position SourcePositionType
 }
 
-struct Mapping {
+pub struct Mapping {
 	GenPosition
 	sources_ind     u32
 	names_ind       NameIndexType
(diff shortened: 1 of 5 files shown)


vlib/v/gen/js/sourcemap/mappings.v
```diff
@@ -25,7 +25,7 @@ struct MappingInput {
 	source_position SourcePositionType
 }
 
-struct Mapping {
+pub struct Mapping {
```
spytheman, line 28: why is that made public?

vlib/v/gen/js/sourcemap/sets.v
```diff
@@ -1,6 +1,6 @@
 module sourcemap
 
-struct Sets {
+pub struct Sets {
```
spytheman, line 3: why?

vlib/v/gen/js/sourcemap/mappings.v
```diff
@@ -3,15 +3,15 @@ module sourcemap
 import v.gen.js.sourcemap.vlq
 import io
 
-struct Empty {}
+pub struct Empty {}
```
spytheman, line 6: why?

vlib/v/gen/js/sourcemap/mappings.v
enghitalo: Json2 encoding needs public struct

vlib/v/gen/js/sourcemap/sets.v
enghitalo: Json2 encoding needs public struct

vlib/v/gen/js/sourcemap/mappings.v
enghitalo: Json2 encoding needs public struct

spytheman:
I think that is not an improvement at all.

It adds more complexity and verbosity, just so that `json2.encode()` could be used here:
```v
-	out += base64.encode(sm.to_json().str().bytes())
+	sourcemap_json := json2.encode(sm.to_sourcemap_json())
+	out += base64.encode(sourcemap_json.bytes())
```

The added structure for `SourceMapJson`, also is not very valuable to me, given that type will be only used for output formatting purposes in one single place, and that is already tested (typos in the map keys will be caught by that).

Issue #20524, opened 2024-01-14
REPL: `print()` or `println()` functions do not display the entire value, or only part of it, after executing a `for` or `if` statement.


viniciusfdasilva:
### Describe the bug

When executing a for or an if statement with a 'print' or 'println' inside, and later, executing a 'println' outside, if the size of the new output is smaller than the size of the previous output, the result is not displayed or is shown in parts, as shown in the image below.

![for](https://github.com/vlang/v/assets/32939036/c91cd11d-48e9-41d8-a03c-70f2326ef903)

And 

![for2](https://github.com/vlang/v/assets/32939036/6b782ca7-5242-49de-9d45-dc4365d07b26)

And 

![if](https://github.com/vlang/v/assets/32939036/a3011574-21e6-48b2-a715-2c6c11b2afe9)

### Reproduction Steps

"I debugged the code and found out what was causing the issue!

Note: When mentioning 'print', consider 'print' or 'println' as well!"

```v
if r.line.starts_with('print') {
	source_code := r.current_source_code(false, false) + '\n${r.line}\n'
	os.write_file(temp_file, source_code) or { panic(err) }
	s := repl_run_vfile(temp_file) or { return 1 }
	
	println('OUTPUT ${s.output}')
	println('LAST OUTPUT ${r.last_output}')

	if s.output.len > r.last_output.len {
		
		cur_line_output := s.output[r.last_output.len..]
		print_output(cur_line_output)
	}
}
```

```v
if s.output.len > r.last_output.len
```

This code line checks if the current output size is smaller than the previous output size. When running a 'for' loop and an 'if' statement with a 'print' inside, the value is saved. When executing a 'print' outside the loop, if the output string is smaller, the 'print' is not executed, or only part of the output is displayed, as shown in the images below.

![error](https://github.com/vlang/v/assets/32939036/ff88bb81-3969-403a-8dcd-6b38091f8b94)

### Expected Behavior

The value displayed on the screen after the execution of the 'print()' or 'println()' function.

### Current Behavior

![if](https://github.com/vlang/v/assets/32939036/69467214-d1a9-46a0-bc39-e947833655f0)

![for](https://github.com/vlang/v/assets/32939036/2a489f45-cd55-4310-8000-a1a60ee9daf5)

![for2](https://github.com/vlang/v/assets/32939036/e69c1d6d-adf2-47b7-998f-d2523b5e824f)

[...]

PR #20525, opened 2024-01-14
[Fix] - repl: Fixing an issue with `print` and println after the execution of for or `if` - ISSUE #20524

viniciusfdasilva:
### Outcome after implementing the solution:

solved

solved2

solvedfor

### v fmt -w cmd/tools/vrepl.v Execution result

test

### DIFF
--- a/cmd/tools/vrepl.v
+++ b/cmd/tools/vrepl.v
@@ -432,10 +432,7 @@ fn run_repl(workdir string, vrepl_prefix string) int {
 			source_code := r.current_source_code(false, false) + '\n${r.line}\n'
 			os.write_file(temp_file, source_code) or { panic(err) }
 			s := repl_run_vfile(temp_file) or { return 1 }
-			if s.output.len > r.last_output.len {
-				cur_line_output := s.output[r.last_output.len..]
-				print_output(cur_line_output)
-			}
+			print_output(s.output)
 		} else {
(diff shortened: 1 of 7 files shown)


spytheman:
Try `v vlib/v/slow_tests/repl/repl_test.v` locally.

spytheman:
Note the content of vlib/v/slow_tests/repl/print_and_fn_call.repl, which is:
```
println('hello')
fn abc(x int) { println(x) }
abc(123)
abc(456)
println('hello')
===output===
hello
123
456
hello
```
i.e. the goal of the deleted heuristic is to emulate the behavior of normal interpreter based REPLs, by suppressing the already known output, of the compiled V program, and only showing the change (the last part).

The way it does it on master (the deleted if), is far from perfect, since it does use just the size of the output, but not the actual content, but the content is important/changed too in the cases you mentioned.

spytheman:
The problem with:
```v
mut numbers := [1,2,3,4,5]
for number in numbers {
println(number)
}
```
is that the body of the loop, was being ignored (treated like just `println(1)`, which the REPL does not store in the program).
You can check that with `list` inside the REPL.

spytheman:
I think a more general solution would be to *always* store the output, and to not use heuristics to ignore statements like `println`.
Then the output of each subsequent program can only grow, and cutting the previous lines will work reliably.

PR #20529, opened 2024-01-14
docs: document the return type restrictions for operator overloading

thesombady:
Documentation for operator overloading with return type limitation.

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -6191,6 +6191,7 @@ To improve safety and maintainability, operator overloading is limited.
 
 - When overriding `<` and `==`, the return type must be strictly `bool`.
 - Both arguments must have the same type (just like with all operators in V).
+- Overloaded operators has to return the same type as the argument (with the expection `<` and `==` which returs boolean expressions).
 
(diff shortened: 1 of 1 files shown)


Delta456 (approved):
Good catch

Delta456:
I will be doing a follow-up PR to make it a checker error when this is merged.

doc/docs.md
```diff
@@ -6191,6 +6191,7 @@ To improve safety and maintainability, operator overloading is limited.
 
 - When overriding `<` and `==`, the return type must be strictly `bool`.
 - Both arguments must have the same type (just like with all operators in V).
+- Overloaded operators has to return the same type as the argument (with the expection `<` and `==` which returns boolean expressions).
```
Delta456, line 6194: As this line goes beyond the 100 words. I think

```suggestion
- Overloaded operators have to return the same type as the argument (with the exception `<` and `==).
```

works better

thesombady: Wow, trying to write to fast hehe :/

PR #20530, opened 2024-01-14
encoding.csv: add a new utility fn `new_reader_from_file/2`

koplenov:
Q: What is the best way to organize high-level helper functions?


Delta456:
Format the files which you have added.

spytheman:
> Q: What is the best way to organize high-level helper functions?

The way you did it is fine imho.

PR #20532, opened 2024-01-14
orm: Ensure table creation before instance insertion [Issue #20017]

### LINKED ISSUE #20017: [orm] Child table not created and Child instance ignored
### Describe the bug

when creating a entity with a list of children entity.
if the child entity table is not created , inserting a parent entity with child will just silently ignore the children persistence.

### Reproduction Steps

create a file with
```
import db.pg

struct Member{
    id int @[primary]
    children []Child @[fkey:parent_id]
}

struct Child{
	id int @[primary]
	parent_id int
}

fn main() {
	db := pg.connect(pg.Config{
		host: 'localhost' 
		user: 'meveo'
		password: 'meveo'
		dbname: 'meveo'
	}) or {
		println('failed to connect')
		println(err)
		return
	}
	
	//sql db { create table Child } or {
	//	println(err)
	//}
	sql db { create table Member } or {
		println(err)
	}
	new_member := Member{
		id: 1
		children: [Child{
			id: 1
			parent_id: 1
		}]
	}
	sql db { insert new_member into Member}!
}
```
and run it

### Expected Behavior

either the child table should be automatically created when the parent table is created
or an error/warning should be given when ignoring the persistence of the children

### Current Behavior

children are silently ignored

### Possible Solution

_No response_

_No response_

### V version

V 0.4.3 b5ba122

ubuntu 22.04.2 in wsl2 (win 11)
> [!NOTE]
>

GGRei:
Hello,

This PR will fix #20017. 

Check for the existence of a table before attempting insertion. 

If the table is not present, it triggers a panic with an explanatory message. 

Given that a panic is triggered, creating a test for this doesn't really make sense to me. However, if triggering a panic is not the desired approach (allowing code execution to continue while returning only an error message), please let me know and I will make the necessary modifications.

Similarly, if a test is absolutely required, please let me know. I am available for any necessary modifications.

### DIFF
--- a/vlib/db/mysql/orm.c.v
+++ b/vlib/db/mysql/orm.c.v
@@ -129,6 +129,7 @@ pub fn (db DB) @select(config orm.SelectConfig, data orm.QueryData, where orm.Qu
 
 // insert is used internally by V's ORM for processing `INSERT ` queries
 pub fn (db DB) insert(table string, data orm.QueryData) ! {
+	db.insert_if_table_exist(table) or { panic(err) }
 	mut converted_primitive_array := db.convert_query_data_to_primitives(table, data)!
 
 	converted_primitive_data := orm.QueryData{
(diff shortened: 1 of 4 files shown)


spytheman:
No. Checking whether a table exist before each insertion is not right :-| .
An insert should just fail, if the table does not exist, and that failure should result an error, not a panic. The error should be propagated, so that it can be handled in the user's code. Panics can not.

PR #20533, opened 2024-01-14
v: $dbg statement - native V debugger REPL

felipensp:
This idea is to provide an core debugger for V devs when using `$dbg;` on their code.

```
vdbg> h
vdbg commands:
  anon?                 check if the current context is anon
  bt                    prints a backtrace
  c, continue           continue debugging
  generic?              check if the current context is generic
  heap                  show heap memory usage
  h, help, ?            show this help
  l, list [lines]       show some lines from current break (default: 3)
  mem, memory           show memory usage
  method?               check if the current context is a method
  m, mod                show current module name
  p, print <var>        prints an variable
  q, quit               exits debugging session in the code
  scope                 show the vars in the current scope
  u, unwatch <var>      unwatches a variable
  w, watch <var>        watches a variable
```

#### Example

```V
fn debugger(a string, b bool) {
    $dbg;
}

struct Test{}


fn (t Test) debugger[T](z T) {
    $dbg;
    dump(z)
}

fn test(params map[string]string) {
    $dbg;    
}

fn test2() {
    a := 1
    if a != 0 {
        b := 2 
        if b > 1 {
            $dbg;
        }
    }
}

fn main() {
    a := {'a': 'b', 'c': 'd'}
    d := Test{}
    d.debugger(a)

    b := 'foo'
    c := true
    println('testing vdbg')
    $dbg;
    println('after debugger;')
    test(a)
    println('after test() call')
    test2()
    debugger(b, c)
    $dbg;
}
```

When using `$dbg` in the V code, the debugger REPL will be invoked when such statement is reached in the execution, bringing to you a debugger REPL.

#### list

```
Break on [main] main in bug.v:35
bug.v:35 vdbg> l
0032      b := 'foo'
0033      c := true
0034      println('testing vdbg')
0035>     $dbg;
0036      println('after debugger;')
0037      test(a)
0038      println('after test() call')
bug.v:35 vdbg> l 2
0033      c := true
0034      println('testing vdbg')
0035>     $dbg;
0036      println('after debugger;')
0037      test(a)
bug.v:35 vdbg> l 1
0034      println('testing vdbg')
0035>     $dbg;
0036      println('after debugger;')
```

#### Inspecting variables

```
Break on [main] test2 in bug.v:22
bug.v:22 vdbg> l
0019      if a != 0 {
0020          b := 2 
0021          if b > 1 {
0022>             $dbg;
0023          }
0024      }
0025  }
bug.v:22 vdbg> p a
a = 1 (int)
bug.v:22 vdbg> p b
b = 2 (int)
bug.v:22 vdbg> scope
b = 2 (int)
a = 1 (int)
```


vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
...
+fn print_memory_use() {
+	println('${gc_memory_use() / 1024} MB')
+}
+
+// print_heap_usage prints the GC heap usage
+fn print_heap_usage() {
+	h := gc_heap_usage()
+	println('heap size: ${h.heap_size / 1024} MB')
+	println('free bytes: ${h.free_bytes / 1024} MB')
+	println('total bytes: ${h.total_bytes / 1024} MB')
```
spytheman, line 105: In a debugger context, the exact amounts in bytes will be preferable.
Imho show them too, or instead of the rounded MB values.

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
...
+	} else {
+		s.write_string('[${d.mod}] ${d.fn_name}')
+	}
+	if d.is_generic {
+		s.write_string(' [generic]')
+	}
+	return s.str()
+}
+
+// print_current_file prints 5 lines before current location
```
spytheman, line 85: 3 lines before, and 3 after may be better

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
...
+	println('heap size: ${h.heap_size / 1024} MB')
+	println('free bytes: ${h.free_bytes / 1024} MB')
+	println('total bytes: ${h.total_bytes / 1024} MB')
+}
+
+// debugger is the implementation for C backend's debugger statement (debugger;)
+@[unsafe]
+pub fn debugger(info DebugContextInfo) ! {
+	mut static profile := u64(0)
+	mut static exited := 0
```
spytheman, line 112: if that is static, and you exit the debugger, you will not be able to enter it anymore

vlib/debug/debug.v
```diff
@@ -102,7 +102,7 @@ fn print_heap_usage() {
 	h := gc_heap_usage()
 	println('heap size: ${h.heap_size / 1024} MB')
 	println('free bytes: ${h.free_bytes / 1024} MB')
-	println('total bytes: ${h.total_bytes / 1024} MB')
+	println('total bytes: ${h.total_bytes} MB')
```
JalonSolov:
```suggestion
	println('total bytes: ${h.total_bytes}')
```

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
...
+fn print_help() {
+	println('vdbg commands:')
+	println('  anon?\t\t\tcheck if the current context is anon')
+	println('  bt\t\t\tprints a backtrace')
+	println('  c, continue\t\tcontinue debugging')
+	println('  f,file\t\tshow current file name')
+	println('  fn,func\t\tshow current function name')
+	println('  generic?\t\tcheck if the current context is generic')
+	println('  heap\t\t\tshow heap memory usage')
+	println('  h, help\t\tshow this help')
```
spytheman, line 23: please add `?` too

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
...
+	println('  list\t\t\tshow 5 lines from current file')
+	println('  l, line\t\tshow current line number')
```
spytheman, line 25: `l` is usually a shortcut for `list` in all debuggers I have used, not for showing the line number.

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
...
+	println('  mem,memory\t\tshow memory usage')
+	println('  method?\t\tcheck if the current context is a method')
+	println('  m,mod\t\t\tshow current module name')
+	println('  p,print <arg>\t\tprints an variable')
+	println('  q,quit\t\texits debugging session in the code')
+	println('  scope\t\t\tshow the vars in the inner most scope')
+	println('  s,profile\t\tstart CPU profiling session')
```
spytheman, line 32: `s` is usually for stepping over a statement, imho remove this shortcut for now. It is better for it, to not work, than to do something unrelated, that people will have to unlearn later.

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
...
+
+const prompt = 'vdbg> '
+
+// print_help prints the debugger REPL commands help
+fn print_help() {
+	println('vdbg commands:')
+	println('  anon?\t\t\tcheck if the current context is anon')
+	println('  bt\t\t\tprints a backtrace')
+	println('  c, continue\t\tcontinue debugging')
+	println('  f,file\t\tshow current file name')
```
spytheman, line 19: `f` is usually for selecting a stack frame. If it is not supported, it is better for it to not work, than to trigger something else. Perhaps `F` can be used instead.

spytheman: Or the prompt could be modified to always show the current file and line number, then both commands will not need a shortcut, since the information will be always visible.

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
...
+			'scope' {
+				info.show_scope()
+			}
+			's', 'profile' {
+				profile = time.sys_mono_now()
+				println('profiler :: starting profiler')
+			}
+			'e', 'profileEnd' {
+				println('profiler :: elapsed time: ${time.Duration(time.sys_mono_now() - profile)}')
+			}
```
spytheman, line 199: I am not sure that I understand the utility.
Suppose you have a long loop, that you want to measure.
Will you have to put 2 $dbg statements before and after it, and then play a key chord?

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
+// Copyright (c) 2019-2024 V devs. All rights reserved.
+// Use of this source code is governed by an MIT license that can be found in the LICENSE file.
+module debug
```
spytheman, line 3: I think the module should be inside `vlib/v/debug`, not in `vlib/debug`.
It is tightly integrated with the V compiler, and I think it should not clutter the common space. People may reasonably want to have their own `debug` modules in their projects.

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
+// Copyright (c) 2019-2024 V devs. All rights reserved.
```
spytheman, line 1: Imho either put your name, or Alex's name here. I think there is no legal entity named `V devs` currently, which is needed for the Copyright to work afaik.

vlib/v/fmt/fmt.v
```diff
@@ -513,6 +513,7 @@ pub fn (mut f Fmt) stmt(node ast.Stmt) {
 		ast.ConstDecl {
 			f.const_decl(node)
 		}
+		ast.DebuggerStmt {}
```
spytheman, line 516: vfmt should preserve the source, if $dbg is there, it should produce it in the output.

vlib/debug/debug.v
```diff
@@ -0,0 +1,214 @@
...
+// print_heap_usage prints the GC heap usage
+fn print_heap_usage() {
+	h := gc_heap_usage()
+	println('heap size: ${h.heap_size / 1024} MB')
+	println('free bytes: ${h.free_bytes / 1024} MB')
+	println('total bytes: ${h.total_bytes / 1024} MB')
+}
+
+// debugger is the implementation for C backend's debugger statement (debugger;)
+@[unsafe]
```
spytheman, line 109:
```suggestion
@[unsafe; markused]
```

spytheman: Calls to this function will only be made by cgen, and `-skip-unused` (the `markused` module is ran before cgen) will not know about it, so it may decide to just strip it.

vlib/debug/debug.v
felipensp: Ok, fixing.

vlib/debug/debug.v
felipensp: I like the idea for `file:number vdbg> ` prompt. 👍🏻

vlib/debug/debug.v
felipensp: Yes, I think it works in this way on nodejs debugger. Or I missed something heh.

vlib/debug/debug.v
felipensp: Ok

vlib/v/fmt/fmt.v
felipensp: Oh, my fault.

vlib/debug/debug.v
felipensp: Good catch.

vlib/debug/debug.v
felipensp: Ok, fixing.

vlib/debug/debug.v
felipensp: Ok, fixing.

vlib/debug/debug.v
felipensp: Yes, this is the plan. When hitting 'q' it does not breaks anymore.

vlib/debug/debug.v
felipensp: Ok fixed. I've also added a parameter to l/line cmd.

smartiniOnGitHub:
Hi guys, I think this is a great feature.
Sorry didn't found related issue (if any), just a question (sorry to post here): would it be possible to have this feature disabled by default when the code has been built with the production flag (to avoid block running applications maybe for mistake) ? Thank s a lot. Bye

felipensp:
> Hi guys, I think this is a great feature. Sorry didn't found related issue (if any), just a question (sorry to post here): would it be possible to have this feature disabled by default when the code has been built with the production flag (to avoid block running applications maybe for mistake) ? Thank s a lot. Bye

File a feature request. 👍🏻

odiroot:
>  anon?                 check if the current context is anon
May I ask, what's this?

felipensp:
> > anon?                 check if the current context is anon
> 
> May I ask, what's this?

Check if the current context is an anon function.

odiroot:
> > > anon?                 check if the current context is anon
> > 
> > 
> > May I ask, what's this?
> 
> Check if the current context is an anon function.

Thanks. That's useful!

PR #20538, opened 2024-01-15
vweb: add an optional before_accept_loop/0 method, to make testing easier and more robust

spytheman:


enghitalo:
x.vweb will need/have something like that?

spytheman:
> `x.vweb` will need/have something like that?

Probably yes, but that will be another PR.

PR #20539, opened 2024-01-15
ast, markused, checker: modified comments on generic receiver type storage

shove70:
The comment of PR 20527 was not clear enough to be misleading, and this PR rewrites it.


spytheman (approved):
Those comments are indeed much clearer.
Thank you.

PR #20544, opened 2024-01-15
checker: fix return map index with or_block (fix #20391)

### LINKED ISSUE #20391: map access `or` block behavior is inconsistent with the rest of the language
### Describe the bug

Please see the code snippets with commentary below.

### Reproduction Steps

Try to compile the following code:
```v
fn get_value(m map[string]int, key string) ?int {
	return m[key] or { none }
}
```

### Expected Behavior

The program to compile successfully. This would be consistent with how `or` blocks behave in other contexts, e.g.:
```v
fn even_or_none(i int) ?int {
	return if i % 2 == 0 { i } else { none }
}

fn useless_fn(i int) ?int {
	return even_or_none(i) or { none }
}
```

### Current Behavior

The compiler gives an error.
```
map.v:2:21: error: `or` block must provide a value of type `int`, not `none`
    1 | fn get_value(m map[string]int, key string) ?int {
    2 |     return m[key] or { none }
      |                        ~~~~
    3 | }
```

### Possible Solution

_No response_

One of the users on Discord pointed out, I can use `or { return none }` instead of just `or { none }` and it'll work fine, but the second `return` seems redundant.
```v
fn get_value(m map[string]int, key string) ?int {
	return m[key] or { return none }
}
```

### V version

V 0.4.3 e19b2dd

V full version: V 0.4.3 5b96d8d.e19b2dd
OS: macos, macOS, 14.2.1, 23C71
Processor: 11 cpus, 64bit, little endian, Apple M3 Pro

vexe mtime: 2024-01-05 00:03:27

vroot: OK
VMODULES: OK
VTMP: OK

Git version: git version 2.39.3 (Apple Git-145)
Git vroot status: weekly.2024.01.test2-1-ge19b2dd4
.git/config present: true

CC version: Apple clang version 15.0.0 (clang-1500.1.0.2.5)
thirdparty/tcc status: thirdparty-macos-arm64 a668e5a0
> [!NOTE]
>

yuyi98:
This PR fix return map index with or_block (fix #20391).

- Fix return map index with or_block.
- Add test.

```v
fn get_value1(m map[string]int, key string) ?int {
	return m[key] or { none }
}

fn get_value2(m map[string]int, key string) !int {
	return m[key] or { error('not found') }
}

fn main() {
	m := {
		'aaa': 111
	}
	if ret1 := get_value1(m, 'aaa') {
		println(ret1)
		assert ret1 == 111
	} else {
		assert false
	}

	if ret2 := get_value2(m, 'aaa') {
		println(ret2)
		assert ret2 == 111
	} else {
		assert false
	}
}

PS D:\Test\v\tt1> v run .
111
111
```

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -80,6 +80,7 @@ pub mut:
 	inside_lambda              bool // true inside `|...| ...`
 	inside_ref_lit             bool // true inside `a := &something`
 	inside_defer               bool // true inside `defer {}` blocks
+	inside_return              bool // true inside `return ...` blocks
 	inside_fn_arg              bool // `a`, `b` in `a.f(b)`
 	inside_ct_attr             bool // true inside `[if expr]`
 	inside_x_is_type           bool // true inside the Type expression of `if x is Type {`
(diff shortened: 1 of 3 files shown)


spytheman:
I can not decide on whether the example should work.

From one point of view:
```v
return expr
```
should behave exactly the same as:
```v
tmp := expr
return tmp
```

and you can not do `tmp := expr_res or { error('x') }`,
but you can do: `tmp := expr_opt or { none }`.

From another point of view, it seems clear what is wanted by just:
```v
fn f() ? { return expr or { none } }
```
and by:
```v
fn f() ! { return expr or { error('x') } }
```

@medvednikov, what do you think, should V work like in this PR?

spytheman:
(It is excellent work, anyways, whatever is decided.)

yuyi98:
Here is to omit return, equivalent to `return none` and `return error('....')`.

medvednikov:
I think this should be merged.

PR #20546, opened 2024-01-15
checker: fix checking give static map as default or init value to struct fields(fix #20512)

### LINKED ISSUE #20512: mutable map field, initialised with const, makes const mutable
### Describe the bug

When a mutable map field in a struct is initialised with a const map, modifications to existing keys in the mutable field affect the const map.

### Reproduction Steps

```v
const xxx = { 'a': 4 } // const map

struct Foo {
mut:
	foo map[string]int = xxx // init with const
}

fn main() {
	mut a := Foo{}
	a.foo['a'] = 99 // should mutate copy, right?
	println(xxx) // { 'a': 99 } ...but const is affected
}
```

### Expected Behavior

Const maps should be immutable.

### Current Behavior

Const maps can be mutated

### Possible Solution

initialisation with `.clone()` fixes the problem.

E.g.:
```v
struct Foo {
mut:
	foo map[string]int = xxx.clone()
}
```

if the struct field is mutable, then a copy-on-write approach could be taken, so that the map is only actually `clone()`ed when mutated.

### V version

V 0.4.4 bcd77ed

MacOS Sonoma 14.1.2

shove70:
1. Fixed #20512 
2. Add tests.

```v
const const_map = {
	'key': 4
}

struct Foo {
mut:
	field map[string]int = const_map
}

fn main() {
	_ = Foo{
		field: const_map
	}
}
```
outputs:
```
a.v:7:25: error: cannot copy map: call `move` or `clone` method (or use a reference)
    5 | struct Foo {
    6 | mut:
    7 |     field map[string]int = const_map
      |                            ~~~~~~~~~
    8 | }
    9 |
a.v:12:10: error: cannot assign a const map to mut struct field, call `clone` method (or use a reference)
   10 | fn main() {
   11 |     _ = Foo{
   12 |         field: const_map
      |                ~~~~~~~~~
   13 |     }
   14 | }
```

### DIFF
--- a/vlib/v/checker/tests/struct_field_init_and_default_is_map_err.out
+++ b/vlib/v/checker/tests/struct_field_init_and_default_is_map_err.out
@@ -1,0 +1,14 @@
+vlib/v/checker/tests/struct_field_init_and_default_is_map_err.vv:7:25: error: cannot copy map: call `move` or `clone` method (or use a reference)
+    5 | struct Foo {
+    6 | mut:
+    7 |     field map[string]int = const_map
+      |                            ~~~~~~~~~
+    8 | }
+    9 |
+vlib/v/checker/tests/struct_field_init_and_default_is_map_err.vv:12:10: error: cannot assign a const map to mut struct field, call `clone` method (or use a reference)
+   10 | fn main() {
+   11 |     _ = Foo{
+   12 |         field: const_map
+      |                ~~~~~~~~~
+   13 |     }
+   14 | }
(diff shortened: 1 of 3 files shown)


vlib/v/checker/tests/struct_field_init_and_default_is_map_err.out
```diff
@@ -0,0 +1,14 @@
+vlib/v/checker/tests/struct_field_init_and_default_is_map_err.vv:7:25: error: cannot copy map: call `move` or `clone` method (or use a reference)
+    5 | struct Foo {
+    6 | mut:
+    7 |     field map[string]int = const_map
```
spytheman, line 4: I think that only `clone` should be suggested here, because `move` will make the `const_map` contain invalid data.
Consider the case, where you have several fields, all doing `= const_map.move()` for their default. One of them will get the data of the original map, the rest will have rubbish.

shove70: hmm...  yes.

PR #20549, opened 2024-01-15
coroutines: fix segfaults with GC

joe-conigliaro:


joe-conigliaro:
not ready to be merged, working on some fixes currently

felipensp:
> not ready to be merged, working on some fixes currently

just set it as Draft :+1:

spytheman:
@joe-conigliaro can it be merged now?

I am afraid that the more we wait, the more chances of conflicts will emerge, and I do not see it breaking anything already working.

PR #20556, opened 2024-01-16
trace_calls: fix `v -m32 -trace-calls run vlib/v/tests/testdata/trace_calls/simple.vv`

spytheman:


smalltalkman:
Only tested `C.fprintf(C.stderr, c'> trace %8d %8lld %8lld %s\n', tid, ns, ssize, fname.str) `, which works in both 32-bit and 64-bit environments.

PR #20558, opened 2024-01-16
picoev: refactor

enghitalo:


vlib/picoev/picoev.v
```diff
@@ -96,44 +98,55 @@ pub fn (mut pv Picoev) add(fd int, events int, timeout int, cb voidptr) int {
 	target.events = 0
 
 	if pv.update_events(fd, events | picoev.picoev_add) != 0 {
-		pv.del(fd)
+		if pv.del(fd) != 0 {
+			eprintln('Error during del')
+		}
+		close_socket(fd) // Close fd on failure
```
spytheman, line 104: Imho closing here, is a violation of layering. You are passing the file descriptor from somewhere else.
That somewhere else should be responsible for checking for errors, and closing it, if it decides to, not Picoev.

vlib/picoev/picoev.v
```diff
@@ -96,44 +98,55 @@ pub fn (mut pv Picoev) add(fd int, events int, timeout int, cb voidptr) int {
...
+	}
+
 	mut target := pv.file_descriptors[fd]
 
 	$if trace_fd ? {
 		eprintln('delete ${fd}')
 	}
 
 	if pv.update_events(fd, picoev.picoev_del) != 0 {
+		eprintln('Error during update_events')
```
spytheman, line 126: If you are going to print that, please also add `picoev.picoev_del` in the message.

vlib/picoev/picoev.v
```diff
@@ -96,44 +98,55 @@ pub fn (mut pv Picoev) add(fd int, events int, timeout int, cb voidptr) int {
...
 	target.loop_id = -1
 	target.fd = 0
+	target.cb = unsafe { nil } // Clear callback to prevent accidental invocations
 	return 0
 }
 
-fn (mut pv Picoev) loop_once(max_wait int) int {
-	pv.loop.now = get_time()
+fn (mut pv Picoev) loop_once(max_wait_in_sec int) int {
+	pv.loop.now = get_time() // Update loop start time
```
spytheman, line 138: Is that comment needed?
To me it is clear from the code itself.

vlib/picoev/picoev.v
```diff
@@ -157,52 +170,54 @@ fn (mut pv Picoev) set_timeout(fd int, secs int) {
...
-
 	for fd, timeout in pv.timeouts {
 		if timeout <= pv.loop.now {
-			to_remove << fd
+			target := pv.file_descriptors[fd]
+			assert target.loop_id == pv.loop.id
+			if target.cb != unsafe { nil } { // Check for valid callback
+				unsafe { target.cb(fd, picoev.picoev_timeout, &pv) }
+			}
+			pv.timeouts.delete(fd)
```
spytheman, line 180: No, deleting inside the `for fd, timeout in pv.timeouts {` is a terrible idea, and it will backfire in long debugging sessions.
There are 2 explicit loops here for a very good reason.
Imho restore the code as it was.

vlib/picoev/picoev.v
```diff
@@ -157,52 +170,54 @@ fn (mut pv Picoev) set_timeout(fd int, secs int) {
...
-			pv.err_cb(pv.user_data, picohttpparser.Request{}, mut &picohttpparser.Response{},
-				err)
-			return
-		}
-		pv.add(newfd, picoev.picoev_read, pv.timeout_secs, raw_callback)
+	setup_sock(accepted_fd) or {
+		eprintln('setup_sock failed, fd: ${accepted_fd}, listen_fd: ${listen_fd}, err: ${err.code()}')
+		pv.err_cb(pv.user_data, picohttpparser.Request{}, mut &picohttpparser.Response{},
+			err)
+		close_socket(accepted_fd) // Close fd on failure
```
spytheman, line 208: good

vlib/picoev/picoev.v
```diff
@@ -157,52 +170,54 @@ fn (mut pv Picoev) set_timeout(fd int, secs int) {
...
 
 // close_conn closes the socket `fd` and removes it from the loop
 @[inline]
 pub fn (mut pv Picoev) close_conn(fd int) {
-	pv.del(fd)
-	close_socket(fd)
+	if pv.del(fd) != 0 {
+		eprintln('Error during del')
+	}
+	close_socket(fd) // Close fd regardless of del outcome
```
spytheman, line 220: Now, that is much more sensible. The method itself is named `close_conn`, and it does what it describes.

vlib/picoev/picoev.v
```diff
@@ -333,30 +347,33 @@ pub fn new(config Config) &Picoev {
...
 
 // serve starts the Picoev server
 pub fn (mut pv Picoev) serve() {
-	spawn update_date(mut pv)
-
 	for {
 		pv.loop_once(1)
+
+		// Update date within the main loop
+		pv.date = update_date_string()
```
spytheman, line 368: Why? Updating it each time is potentially much slower, since you can have thousands of requests served per second, and you usually you do not need a high resolution for the time.

spytheman: imho a thread updating it periodically, like it was before, is a better solution.

spytheman:
@enghitalo please avoid refactoring/renaming in the same PR, as functional changes. I know that it is sometimes unavoidable, but here, more than half of the changes are renames.

Intermixing both in the same PR makes it much harder to read and approve, and the cost in time is not linear, especially if there are problems found. You (and I) will have to read those diffs over and over and over again, till everything is good to merge, while if those were 2 separate PRs (one pure refactoring, and another with potentially breaking, but surely functional changes), the pure refactoring one would have been approved and merged pretty much immediately, and the diff for the second one, would have been much smaller and focused.

spytheman:
Think of it as wading through a pile of straw, to search for a pearl, or a needle potentially hidden in it.

enghitalo:
@spytheman Thank you for your review and tips. I will try finish it next week

vlib/picoev/picoev.v
enghitalo: By what I understandood `fn (mut pv Picoev) poll_once(max_wait_in_sec int) int {` will hold the loop for `max_wait_in_sec seconds (In this case, 1 second)` then process all events get in this time.

https://man7.org/linux/man-pages/man2/epoll_wait.2.html

spytheman: It will timeout in that time, if there are no events (idle server), but it will also return, once there are any events (say a new connection/request) (after handling them). If you have many requests, I think that loop here, will execute many times a second.

Casper64: Yes that is right, it will wait for a maximum of `max_wait_in_sec` seconds for any events. If no events have occurred the function returns. If there is an event to be handled it will immediately handle that event and return. So in most cases the loop will execute many times per second.

PR #20559, opened 2024-01-16
readline: completion support

felipensp:
Usage:

```V
import readline

fn main() {
	list := ['mem', 'memory', 'method']


	mut r := readline.Readline{
		// completion_list: list
		// or   
		completion_callback: fn [list] (prefix string) []string {
			return list.filter(it.starts_with(prefix))
		}
	}

	dump(r.read_line('prompt> ')!)
}
```

### DIFF
--- a/vlib/readline/readline_nix.c.v
+++ b/vlib/readline/readline_nix.c.v
@@ -458,12 +458,14 @@ fn (mut r Readline) delete_word_left() {
 
 	r.current.delete_many(r.cursor, orig_cursor - r.cursor)
 	r.refresh_line()
+	r.completion_clear()
 }
 
 fn (mut r Readline) delete_line() {
 	r.current = []
 	r.cursor = 0
 	r.refresh_line()
+	r.completion_clear()
 }
 
 // suppr_character removes (suppresses) the character in front of the cursor.
@@ -585,6 +587,7 @@ fn (mut r Readline) history_next() {
 	r.refresh_line()
 }
 
+// completion implements the tab completion feature
 fn (mut r Readline) completion() {
 	// check if completion is used
 	if r.completion_list.len == 0 && r.completion_callback == unsafe { nil } {
@@ -625,6 +628,7 @@ fn (mut r Readline) completion() {
 	r.refresh_line()
 }
 
+// completion_clear resets the completion state
 fn (mut r Readline) completion_clear() {
 	r.last_prefix_completion = []rune{}
 	r.last_completion_offset = 0
--- a/vlib/readline/readline_nix.c.v
+++ b/vlib/readline/readline_nix.c.v
@@ -36,6 +36,7 @@ enum Action {
 	overwrite
 	clear_screen
 	suspend
+	completion
 }
 
 // enable_raw_mode enables the raw mode of the terminal.
@@ -174,7 +175,7 @@ pub fn read_line(prompt string) !string {
 }
 
 // analyse returns an `Action` based on the type of input byte given in `c`.
-fn (r Readline) analyse(c int) Action {
+fn (mut r Readline) analyse(c int) Action {
 	if c > 255 {
 		return Action.insert_character
 	}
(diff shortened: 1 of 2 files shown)


vlib/readline/readline_nix.c.v
```diff
@@ -183,8 +184,12 @@ fn (r Readline) analyse(c int) Action {
 			return .eof
 		} // NUL, End of Text, End of Transmission
 		`\n`, `\r` {
+			r.last_prefix_completion = []rune{}
```
spytheman, line 187:
```suggestion
			r.last_prefix_completion.clear()
```

spytheman: (it will allocate less)

vlib/readline/readline_nix.c.v
```diff
@@ -211,6 +216,7 @@ fn (r Readline) analyse(c int) Action {
 		} // CTRL + Z, SUB
 		else {
 			if c >= ` ` {
+				r.last_prefix_completion = []rune{}
```
spytheman, line 219:
```suggestion
				r.last_prefix_completion.clear()
```

vlib/readline/readline_nix.c.v
```diff
@@ -577,6 +587,53 @@ fn (mut r Readline) history_next() {
...
+
+	// set the possible completion match
+	r.current = opts[r.last_completion_offset].runes()
+	r.cursor = r.current.len
+	r.refresh_line()
+}
+
+// completion_clear resets the completion state
+fn (mut r Readline) completion_clear() {
+	r.last_prefix_completion = []rune{}
```
spytheman, line 633:
```suggestion
	r.last_prefix_completion.clear()
```

vlib/readline/readline_nix.c.v
```diff
@@ -577,6 +587,53 @@ fn (mut r Readline) history_next() {
...
+	prefix := if r.last_prefix_completion.len > 0 { r.last_prefix_completion } else { r.current }
+	// no prefix set
+	if prefix.len == 0 {
+		return
+	}
+	// filtering by prefix
+	opts := if r.completion_list.len > 0 {
+		r.completion_list.filter(it.starts_with(prefix.string()))
+	} else {
+		r.completion_callback(prefix.string())
```
spytheman, line 606: this is not protected by `if r.completion_callback != unsafe { 0 } {`

felipensp: > this is not protected by `if r.completion_callback != unsafe { 0 } {`

On the beginning of the fn has the check ensuring such cond which cover this else part.

spytheman: I think the check is required, otherwise it will crash, when `r.completion_callback` is 0.

spytheman: The start of the function is:
`if r.completion_list.len == 0 && r.completion_callback == unsafe { nil } {`

when the first expression is false, the second does not matter, you will still not return and go here.

spytheman: imho just add separate check here, closer to when it is used

spytheman: hm, I am wrong, and understand what you mean.
I still do not like it, the check is very removed from the place it is protecting (the call).

vlib/readline/readline_nix.c.v
```diff
@@ -577,6 +587,53 @@ fn (mut r Readline) history_next() {
...
+	}
+
+	prefix := if r.last_prefix_completion.len > 0 { r.last_prefix_completion } else { r.current }
+	// no prefix set
+	if prefix.len == 0 {
+		return
+	}
+	// filtering by prefix
+	opts := if r.completion_list.len > 0 {
+		r.completion_list.filter(it.starts_with(prefix.string()))
```
spytheman, line 604: `prefix.string()` can be called a lot, if there are lots of potential candidates.
`sprefix := prefix.string()` before the check, can eliminate it.

vlib/readline/readline_nix.c.v
felipensp: yeah, I've improved. thanks

spytheman (approved):
Excellent work, done very quickly.
Hats off to you @felipensp 🙇🏻‍♂️ .

PR #20561, opened 2024-01-17
v: add map update-init syntax

edam:
Add support for map update-init syntax to v:

```v
foo := {
    ...base_map
    'a': 55
    'b': 66
}
```

which is functionally equivalent to:
```v
mut foo := base_map.clone()
foo['a'] = 55
foo['b'] = 66
```
except that it yields an expression which does not have to be mutable.

This PR updates:
* ast, checker, parser and cgen
* adds `new_map_update_init()` to builtin/map.v to support cgen
* docs
* v fmt
* v tests & checker tests

### DIFF
--- a/vlib/builtin/map.v
+++ b/vlib/builtin/map.v
@@ -284,6 +284,20 @@ fn new_map_init(hash_fn MapHashFn, key_eq_fn MapEqFn, clone_fn MapCloneFn, free_
 	return out
 }
 
+fn new_map_update_init(update &map, n int, key_bytes int, value_bytes int, keys voidptr, values voidptr) map {
+	mut out := unsafe { update.clone() }
+	mut pkey := &u8(keys)
+	mut pval := &u8(values)
+	for _ in 0 .. n {
+		unsafe {
+			out.set(pkey, pval)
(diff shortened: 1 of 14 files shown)


edam:
@spytheman are you able to retry the failing test? It seems unrelated?

shove70:
> @spytheman are you able to retry the failing test? It seems unrelated?

restarted.

vlib/builtin/map.v
```diff
@@ -284,6 +284,20 @@ fn new_map_init(hash_fn MapHashFn, key_eq_fn MapEqFn, clone_fn MapCloneFn, free_
 	return out
 }
 
+fn new_map_update_init(update &map, n int, key_bytes int, value_bytes int, keys voidptr, values voidptr) map {
+	mut out := unsafe { update.clone() }
```
spytheman, line 288:
```suggestion
fn new_map_update_init(original &map, n int, key_bytes int, value_bytes int, keys voidptr, values voidptr) map {
	mut out := unsafe { original.clone() }
```

spytheman (approved):
Excellent work.

PR #20567, opened 2024-01-17
picoev: renaming, doc

enghitalo:


vlib/picoev/loop_default.c.v
```diff
@@ -7,6 +7,8 @@ $if windows {
 	#include <sys/select.h>
 }
 
+// Represents the select-based event loop implementation.
+// Contains an ID and a timestamp for tracking time.
```
spytheman, line 11: Those fields are not public, there is no need to mention them.

vlib/picoev/picoev.v
```diff
@@ -4,40 +4,48 @@ import net
...
 pub struct Target {
 pub mut:
-	fd      int
-	loop_id int = -1
-	events  u32
-	cb      fn (int, int, voidptr) = unsafe { nil }
+	fd       int // file descriptor
+	loop_id  int = -1
+	events   u32
+	callback fn (int, int, voidptr) = unsafe { nil }
```
spytheman, line 38: That is a breaking change for no good reason at all. Revert to the old field name.

vlib/picoev/picoev.v
```diff
@@ -105,14 +115,21 @@ pub fn (mut pv Picoev) add(fd int, events int, timeout int, cb voidptr) int {
...
+// removes a file descriptor from the event loop
+@[deprecated: 'use remove() instead']
 @[direct_array_access]
 pub fn (mut pv Picoev) del(fd int) int {
+	return pv.remove(fd)
+}
+
+// removes a file descriptor from the event loop
+@[direct_array_access]
+pub fn (mut pv Picoev) remove(fd int) int {
```
spytheman, line 127: Renaming a public function is a breaking change, revert the old name.

spytheman: Although you do have `@[deprecated]` which mitigates that.

vlib/picoev/picoev.v
```diff
@@ -105,14 +115,21 @@ pub fn (mut pv Picoev) add(fd int, events int, timeout int, cb voidptr) int {
...
 
-// del removes a file descriptor from the loop
+// removes a file descriptor from the event loop
+@[deprecated: 'use remove() instead']
 @[direct_array_access]
 pub fn (mut pv Picoev) del(fd int) int {
+	return pv.remove(fd)
+}
+
+// removes a file descriptor from the event loop
```
spytheman, line 125:
```suggestion
// remove removes a file descriptor from the event loop
```
otherwise `v doc` will not pick up the comment for the new public method

vlib/picoev/picoev.v
```diff
@@ -122,17 +139,18 @@ pub fn (mut pv Picoev) del(fd int) int {
 	pv.set_timeout(fd, 0)
 	target.loop_id = -1
 	target.fd = 0
+	// REVIEW - should not be better unsafe { free(target.fd) }?
```
spytheman, line 142: You can not free a file descriptor. A file descriptor, is not a pointer to a block of memory that has been allocated, and that needs to be freed, it is just a simple integer, that is used as an index to some internal data structure in the kernel, that the kernel itself manages.

spytheman: btw, it is better to ask such questions in Discord, or as a comment to the issue, not as an internal source comment, since the later requires also cleaning, and re-triggering CI, while doing so :-| .

enghitalo: Sorry. I write while coding and forget to delete it all later. :/

vlib/picoev/picoev.v
```diff
@@ -48,11 +56,13 @@ pub:
...
+// Core structure for managing the event loop and connections.
+// Contains event loop, file descriptor table, timeouts, buffers, and configuration.
 @[heap]
 pub struct Picoev {
-	cb     fn (voidptr, picohttpparser.Request, mut picohttpparser.Response) = unsafe { nil }
-	err_cb fn (voidptr, picohttpparser.Request, mut picohttpparser.Response, IError) = default_err_cb
-	raw_cb fn (mut Picoev, int, int) = unsafe { nil }
+	cb             fn (voidptr, picohttpparser.Request, mut picohttpparser.Response) = unsafe { nil }
+	error_callback fn (voidptr, picohttpparser.Request, mut picohttpparser.Response, IError) = default_error_callback
+	raw_callback   fn (mut Picoev, int, int) = unsafe { nil }
```
spytheman, line 65: this is a breaking change, restore those names

vlib/picoev/picoev.v
```diff
@@ -84,19 +94,19 @@ pub fn (mut pv Picoev) init() {
 	}
 }
 
-// add adds a file descriptor to the loop
+// adds a file descriptor to the event loop
```
spytheman, line 97:
```suggestion
// add a file descriptor to the event loop
```
The doc comment should start with the name of the method. The method is `add`, not `adds`.

vlib/picoev/picoev.v
```diff
@@ -335,11 +353,11 @@ pub fn new(config Config) !&Picoev {
 
 	pv.init()
 
-	pv.add(listen_fd, picoev.picoev_read, 0, accept_callback)
+	pv.add(listening_socket_fd, picoev.picoev_read, 0, accept_callback)
 	return pv
 }
 
-// serve starts the Picoev server
+// serve starts the event loop for listening and handling connections
```
spytheman, line 360:
```suggestion
// serve starts the event loop for accepting new connections
// See also picoev.new().
```

vlib/picoev/loop_default.c.v
spytheman: Remove the comments. The fields are not public. There is no need of documenting them, in fact, it can be misleading.

vlib/picoev/picoev.v
JalonSolov: Alternatively (since the new names do make better sense), add them as 2 new fields, and mark the old ones deprecated.

vlib/picoev/picoev.v
```diff
@@ -105,14 +118,21 @@ pub fn (mut pv Picoev) add(fd int, events int, timeout int, cb voidptr) int {
...
+
+// remove a file descriptor from the event loop
+@[direct_array_access]
+pub fn (mut pv Picoev) remove(fd int) int {
 	assert fd < picoev.max_fds
 	mut target := pv.file_descriptors[fd]
 
 	$if trace_fd ? {
-		eprintln('delete ${fd}')
+		eprintln('remove ${fd}')
```
Casper64, line 135: I don't think this function should be renamed to `remove`.  The delete event is a common term in event loops: e.g. `EPOLL_CTL_DEL`, kqueue: `EV_DELETE`. Why should it be changed?

For me remove means that something is not fully gone, but data could exists/be recovered. While delete completely erases all data, which is the case.

Issue #20568, opened 2024-01-17
C error when trying to append ?Foobar into an array of Foobar


zeozeozeo:
**V doctor:**
```
V full version: V 0.4.4 5306469.a79a9cb
OS: windows, Microsoft Windows 10 Pro v19045 64-bit
Processor: 20 cpus, 64bit, little endian, 

getwd: C:\Users\user\Desktop\Code\leaflang
vexe: C:\Users\user\Desktop\Code\v\v.exe
vexe mtime: 2024-01-17 15:35:51

vroot: OK, value: C:\Users\user\Desktop\Code\v
VMODULES: OK, value: C:\Users\user\.vmodules
VTMP: OK, value: C:\Users\user\AppData\Local\Temp\v_0

Git version: git version 2.42.0.windows.1
Git vroot status: weekly.2024.03-7-ga79a9cb5
.git/config present: true

CC version: Error: 'cc' is not recognized as an internal or external command,

operable program or batch file.


thirdparty/tcc status: thirdparty-windows-amd64 a39eb79b

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg test2.v`
```v
pub enum Foobar {
	foo
	bar
}

fn get_foo_or_none(fail bool) ?Foobar {
	return if fail {none} else {.foo}
}

fn main() {
	mut foobars := []Foobar{}

	// append a ?Foobar into an array of Foobar
	// results in C error
	foobars << get_foo_or_none(false)
	foobars << get_foo_or_none(true)
}

```

**What did you expect to see?**

Checker error

**What did you see instead?**
```
==================
C:/Users/user/AppData/Local/Temp/v_0/test2.01HMC11JHX95P8NNNGF59M8VNW.tmp.c:6862: warning: implicit declaration of function 'tcc_backtrace'
C:/Users/user/AppData/Local/Temp/v_0/test2.01HMC11JHX95P8NNNGF59M8VNW.tmp.c:12987: error: cannot convert 'struct _option_main__Foobar' to 'enum <anonymous>'
...
==================
(Use `v -cg` to print the entire error message)

builder error: 
==================
C error. This should never happen.

This is a compiler bug, please report it using `v bug file.v`.

https://github.com/vlang/v/issues/new/choose

You can also use #help on Discord: https://discord.gg/vlang

```

Relevant generated C code:

```c
VV_LOCAL_SYMBOL void main__main(void) {
[...]

PR #20573, opened 2024-01-17
 checker: check right type is option before appending to array 

Delta456:
Fixes https://github.com/vlang/v/issues/20568


spytheman (approved):
Excellent work.

PR #20574, opened 2024-01-17
vast,ast: output map init update expression

edam:
PR adds:

* `v ast` support for map update-init syntax
* `ast.Expr.str()` support for for map update-init syntax
  * also added `StructInit.update_expr_pos`, which was missing


spytheman (approved):
Excellent work.

PR #20575, opened 2024-01-17
checker: fix .variant smartcast on non-comptime variables

felipensp:
```V
type TestSum = int | bool

fn main() {
    a := TestSum(true)
    $for v in TestSum.variants {
        if a is v {
                dump(a)
        }
    }    
}
```


spytheman (approved):
Excellent work.

PR #20580, opened 2024-01-18
orm: fix orm insert issue if table missing [Issue : #20017]

### LINKED ISSUE #20017: [orm] Child table not created and Child instance ignored

### LINKED ISSUE #20019: [orm] Children persistence ignore if fkey sql type mismatch
### Describe the bug

When creating an entity with id of type string  then children have a parent_id of sql_type 'uuid', then persisting an parent with clidren will silently ignore the persistence of the children

### Reproduction Steps

run this file
```
import db.pg

struct Member{
    id string @[primary]
    children []Child @[fkey:parent_id]
}

struct Child{
	id string @[primary]
	parent_id string @[sql_type:'uuid']
}

fn main() {
	db := pg.connect(pg.Config{
		host: 'localhost' 
		user: 'meveo'
		password: 'meveo'
		dbname: 'meveo'
	}) or {
		println('failed to connect')
		println(err)
		return
	}
	
	sql db { create table Child } or {
		println(err)
	}
	sql db { create table Member } or {
		println(err)
	}
	new_member := Member{
		id: '1'
		children: [Child{
			id: '1'
			parent_id: '1'
		}]
	}
	sql db { insert new_member into Member}!
}
```

### Expected Behavior

An error should notify that the foreign key `parent_id` has not the same type as the parent primary column.

### Current Behavior

no error thrown but children are not persisted

### Possible Solution

_No response_

_No response_

### V version

V 0.4.3 b5ba122

ubuntu 22.04.2 in wsl2 (win 11)
> [!NOTE]
>

GGRei:
Hello,

Fixed #20017 
Fixed #20019 

To achieve this, a condition has been added in cgen.orm to first verify if the primary element has been successfully inserted. Following this, it checks if the sub-element insertions are successful. 

If not, the error is captured and returned to the user code for handling as desired.

A test function has also been added.

Would it be necessary to includes files in v/slow_tests/ for this update?

I am available for any suggestions/modifications.

### DIFF
--- a/vlib/orm/orm_insert_test.v
+++ b/vlib/orm/orm_insert_test.v
@@ -393,7 +393,7 @@ fn test_orm_insert_with_child_element_and_no_table() {
 		select from Parent
 	}!
 
-	assert p_table[1].children[0].name == 'Sophia'
+	assert p_table[2].children[0].name == 'Sophia'
 }
 
 @[table: 'customers']
--- a/vlib/orm/orm_insert_test.v
+++ b/vlib/orm/orm_insert_test.v
@@ -348,6 +348,54 @@ fn test_orm_insert_with_multiple_child_elements() {
 	assert parent.notes[2].text == 'Third note'
 }
 
+fn test_orm_insert_with_child_element_and_no_table() {
+	mut db := sqlite.connect(':memory:')!
+
+	sql db {
+		create table Parent
+	}!
+
+	new_parent := Parent{
+		name: 'test'
+		children: [
+			Child{
+				name: 'Lisa'
+			},
+		]
+	}
+
+	sql db {
+		insert new_parent into Parent
+	} or { assert true }
+
+	sql db {
+		create table Child
+	}!
+
+	sql db {
+		insert new_parent into Parent
+	} or { assert false }
+
+	new_parent_two := Parent{
+		name: 'retest'
+		children: [
+			Child{
+				name: 'Sophia'
+			},
+		]
+	}
+
+	sql db {
+		insert new_parent_two into Parent
+	} or { assert false }
+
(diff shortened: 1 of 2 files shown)


shove70:
It would be better to avoid the child table loop when there is known to be an error in the main table

GGRei:
Yes, you are right. Its modified.

GGRei:
@spytheman Are there any tasks you'd like me to work on here, or should I close this PR?

medvednikov:
Looks good, thanks.

PR #20504, opened 2024-01-12
vweb: fix parsing url error when querypath is '//'(fix #20476)

shove70:
Close #20476

PR #20584, opened 2024-01-19
picoev, x.vweb: small fixes and add changes from vweb

Casper64:
Add changes from `vweb` to `x.vweb`  and fix notices.

This pr is a combination of #20548, #20538, #20504 and #20422

File responses on linux are now handled using the `sendfile` function for speed improvements.

**Edit:**
Make Picoev's `listen` function behave similair to `net.listen_tcp`: set Picoev's listening socket to dualstack mode so it can accept connections from both ipv4 and ipv6 addresses and make the address family ipv6 by default.


vlib/picoev/picoev.v
```diff
@@ -44,8 +44,8 @@ pub:
 	max_headers  int            = 100
 	max_read     int            = 4096
 	max_write    int            = 8192
-	family       net.AddrFamily = .ip
-	host         string = 'localhost'
+	family       net.AddrFamily = .ip6
```
JalonSolov, line 47: ip4 is still a _lot_ more prevalent than ip6, so I think ip4 should remain the default.

spytheman (approved):
Stellar work ⭐, very good improvements!

PR #20586, opened 2024-01-19
docs: fix link for `an example of a module that calls C code from V`

ZihxS:
change from https://github.com/vlang/v/blob/master/vlib/v/tests/project_with_c_code/mod1/wrapper.v to https://github.com/vlang/v/blob/master/vlib/v/tests/project_with_c_code/mod1/wrapper.c.v


Delta456 (approved):
Good catch

PR #20591, opened 2024-01-19
checker: fix non dereferenced enum in match statements (fixes #10045)

### LINKED ISSUE #10045: Forgetting dereferencing enum reference fails
**V version:** V 0.2.2 273655e.e9d7ff7
**OS:** Microsoft Windows 10 Pro v19041 64 bit

**What did you do?**
```v
enum SomeType {
	a
}
fn f(t &SomeType) ?int {
	return match
		t	// note the missing asterisk
			{
				.a {
					panic('This does not happen!')
					3
				}
			}
}
fn main() {
	t := SomeType.a
	f(&t)?
	assert false // should not happen, but does
}
```

A compiler error: `missing '*' dereferencing in match statement`. Or at least auto-dereferencing matching the `.a` in the example.
 
**What did you see instead?**
A failed match, as well as `Assertion failed` in the above example.

pierrec:
Add a check for non-dereferenced enum values when used in a match statement and a corresponding test.
This fixes issue #10045.


spytheman (approved):
Good work.

Issue #20606, opened 2024-01-20
cgen.v:5862:2: error: expression evaluated but not used


Lycs-D:
### Describe the bug

![c1834c09-83d3-495a-a609-bf96551d0c2b](https://github.com/vlang/v/assets/139197676/c1834c09-83d3-495a-a609-bf96551d0c2b)

### Reproduction Steps

```
module main

import v.ast
import v.parser
import v.pref
import os

fn main() {
	file := os.join_path(@VEXEROOT, 'vlib/v/gen/c/cgen.v')
	table:= ast.new_table()
	pref_:= pref.new_preferences()
	dump(parser.parse_file(file, table, .parse_comments, pref_))
}
```

### Expected Behavior

no error

### Current Behavior

error: expression evaluated but not used

### V version

V 0.4.4 df3591c

### Environment details (OS name and version, etc.)

V full version: V 0.4.4 d744314.df3591c
OS: windows, Microsoft Windows Server 2022 Datacenter v26010 64 λ
Processor: 8 cpus, 64bit, little endian
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

spytheman:
`cmd/tools/measure/parser_speed.v` uses:

```v
    mut pref_ := pref.new_preferences()
    pref_.is_fmt = true
    pref_.skip_warnings = true
    pref_.output_mode = .silent
```

spytheman:
`is_fmt = true` in particular, makes the parser a lot more relaxed

StunxFS:
Wouldn't it be nice to detect that in the checker?

spytheman:
> Wouldn't it be nice to detect that in the checker?

Yes, it would be nicer, for that error to be in the checker, instead of inside the parser. The current parser has several such places, where ideally, it should not error, but instead the checker should.

PR #20607, opened 2024-01-20
parser: Fixer #20606 

Lycs-D:


vlib/v/gen/native/comptime.v
```diff
@@ -73,143 +73,143 @@ fn (mut g Gen) comptime_is_truthy(cond ast.Expr) bool {
 }
 
 fn (mut g Gen) comptime_ident(name string, is_comptime_option bool) bool {
-	return match name {
```
JalonSolov, line 76: Why change this?  Returning the branch should be the same effect as a return in every branch.  If not, there's a different bug.

vlib/v/gen/native/elf.v
```diff
@@ -938,16 +938,17 @@ pub fn (mut g Gen) link_elf_file(obj_file string) {
 	crtn := g.find_o_path('crtn.o')
 	lpaths := g.get_lpaths()
 
-	arch := match g.pref.arch {
```
JalonSolov, line 941: Same as previous... should work fine to assign the results of the match.

vlib/v/gen/c/cgen.v
```diff
@@ -5856,12 +5856,12 @@ fn (mut g Gen) global_decl(node ast.GlobalDecl) {
 	// was static used here to to make code optimizable? it was removed when
 	// 'extern' was used to fix the duplicate symbols with usecache && clang
 	// visibility_kw := if g.pref.build_mode == .build_module && g.is_builtin_mod { 'static ' }
-	visibility_kw := if
```
JalonSolov, line 5859: Why change this structure?

JalonSolov:
The only thing I see when I run this (after a bunch of AST stuff) is
```
git/v/thirdparty/stdatomic/nix/cpp/gen.v:28:3: error: expression evaluated but not used
   26 |             exit(2)
   27 |         }
   28 |         'no_file' // bits/atomic_base.h is only used on linux
      |         ~~~~~~~~~
   29 |     }
   30 |
```

PR #20617, opened 2024-01-21
docs: update range `for`: 0-based index rationale

verulian:
Added a note to the Range `for` section in the documentation to explain the rationale behind using exclusive range notation and zero-based indexing. From spytheman in Discord https://discord.com/channels/592103645835821068/592115457029308427/1194711919572033576, This update references Edsger W. Dijkstra's 'Why Numbering Should Start at Zero' to provide theoretical backing and clarity on this design choice.

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -2058,6 +2058,15 @@ for i in 0 .. 5 {
 `low..high` means an *exclusive* range, which represents all values
 from `low` up to *but not including* `high`.
 
+> [!NOTE]
+> This exclusive range notation and zero-based indexing follow principles of
+logical consistency and error reduction. As Edsger W. Dijkstra outlines in
+'Why Numbering Should Start at Zero' ([EWD831](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html)), zero-based indexing aligns the 
+index with the preceding elements in a sequence, simplifying handling and 
+minimizing errors, especially with adjacent subsequences. This logical and 
+efficient approach shapes our language design, emphasizing clarity and reducing 
+confusion in programming.
+
 #### Condition `for`
 
 ```v
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -2061,7 +2061,9 @@ from `low` up to *but not including* `high`.
 > [!NOTE]
 > This exclusive range notation and zero-based indexing follow principles of
 logical consistency and error reduction. As Edsger W. Dijkstra outlines in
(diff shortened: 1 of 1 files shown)


verulian:
Not sure if there is a formatter used for .md files. I don't think `v fmt` works for these. Seems line length is off in what I have done here, especially due to the URL. Anyone able to clear this up?

doc/docs.md
```diff
@@ -2058,6 +2058,15 @@ for i in 0 .. 5 {
 `low..high` means an *exclusive* range, which represents all values
 from `low` up to *but not including* `high`.
 
+> [!NOTE]
+> This exclusive range notation and zero-based indexing follow principles of
+logical consistency and error reduction. As Edsger W. Dijkstra outlines in
+'Why Numbering Should Start at Zero' ([EWD831](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html)), zero-based indexing aligns the
```
StunxFS, line 2064:
```suggestion
'Why Numbering Should Start at Zero' 
([EWD831](https://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html)), 
zero-based indexing aligns the 
```

JalonSolov:
You can check before submitting changes with `v check-md <file>`.

verulian:
Thank you for the command @JalonSolov. `v check-md doc/docs.md` only returned:
```

Warnings: 0 | Errors: 0 | OKs: 248
```
@StunxFS maybe just manually pushed that url down one line and didn't use any mechanism to check?

spytheman:
Lines with links can now be longer (up to 250 characters currently, determined by `too_long_line_length_link` in `cmd/tools/vcheck-md.v`) to avoid false positives, but sometimes .md files are read in a terminal, or in a text editor pane, so it is good to have a document that does not require horizontal scrolling too much.

spytheman (approved):
Excellent addition.

verulian:
Thank you. Maybe there is a certain length we should use for rule? Maybe `v check-md` could throw a warning or some notification that line length should be adjusted for easier standardization?

spytheman:
It already does.

verulian:
Mhmm. Thanks @spytheman. Spose I need to look into this more to see how this works. Haven't had success making it give the above result from @StunxFS so I think I will just need to use my eyes when it doesn't catch something itself. I see a lot of mixed line lens. Some are over 100 chars and some are around 90. Maybe just eyeballing is the best to keep around similar surrounding contents.

Issue #20499, opened 2024-01-11
JavaScript backend treats u32 as i32


justinas:
### Describe the bug

When compiled to JS, `u32` variables seem to be limited to the range of `i32` (signed) instead.

### Reproduction Steps

```
a := u32(0x7FFFFFFF)
b := u32(0xFFFFFFFF)
println(a)
println(b)
```

### Expected Behavior

Expected the code, when run, to output

```
2147483647
4294967295
```

using any backend.

### Current Behavior

```console
$ v run u32.v
2147483647
4294967295
$ v -o u32.js u32.v
$ node u32.js
2147483647
-1
```

### V version

V 0.4.4 ac2dcc2

### Environment details (OS name and version, etc.)

NixOS 23.11 (Tapir)
gcc (GCC) 12.3.0

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #20618, opened 2024-01-21
jsgen: Fix javascript backend treats u32 as i32 [issue #20499]

GGRei:
Fix issue #20499 

Addresses the issue where u32 variables in the JavaScript backend were incorrectly treated as i32, leading to incorrect arithmetic results and comparisons for values exceeding the i32 range.

These adjustments ensure that u32 values are correctly managed in the JavaScript backend.

I'm always available for any suggestions/modifications.

**Edit :**

my test files is a bit usless but I can remove if needed...


enghitalo:
What about u16? 🤔

spytheman:
> What about u16? 🤔

`u16` fits entirely into a signed JS number and it is already handled.

spytheman (approved):
Excellent work.

spytheman:
Hm, `u16` is not handled entirely correctly.
```v
a := u16(0x7FFF)
b := u16(0xFFFF)
println(a)
println(b)
```
this does work, producing:
```
32767
65535
```
with both the default C backend, and the JS one.

However this produces different results with `-b c` and `-b js`, due to the default mask:
```v
a := u16(0x7FFFFFFF)
b := u16(0xFFFFFFFF)
println(a)
println(b)
```
produces:
```
0[14:44:36]@m1: (master) ~/v $ v run x.v
65535
65535
0[14:44:38]@m1: (master) ~/v $ v -b js run x.v
2147483647
-1
```

GGRei:
@spytheman i'm going to work on it.

PR #20624, opened 2024-01-22
examples/term.ui: add casts to term.ui functions to satisfy stricter type checking

kimshrier:
When using clang version 14, the function types for init_fn, frame_fn, event_fn, and cleanup_fn were not strictly enforced and the example programs in examples/term.ui would all compile.

With clang version 16, these function types are strictly checked and the added type definitions clean up the errors.

### DIFF
--- a/examples/term.ui/cursor_chaser.v
+++ b/examples/term.ui/cursor_chaser.v
@@ -82,12 +82,15 @@ fn event(e &tui.Event, mut app App) {
 	}
 }
 
+type EventFn = fn (&tui.Event, voidptr)
+type FrameFn = fn (voidptr)
+
 fn main() {
 	mut app := &App{}
 	app.tui = tui.init(
 		user_data: app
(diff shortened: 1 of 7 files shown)


kimshrier:
I have tested these changes on FreeBSD versions 13.2, 14.0, and current, as well as on macOS 14.2.1.  I don't know why tests of files that I did not touch are failing.  I do not have any windows or linux machines that I can test on.

JalonSolov:
Sometimes the CI is just flaky, and the tests need to be re-run.

spytheman:
The CI failures should be fixed by #20631 soon.

spytheman (approved):
Excellent work.

spytheman:
I've tested on Ubuntu 20.04, with clang 12.0, and `./v -cc clang-12 -cstrict examples/term.ui/term_drawing.v` works on this PR, but it did fail on master.

PR #20625, opened 2024-01-22
debug: add command to clear the terminal in the $dbg prompt

ldedev:
I have a little OCD 😁, I missed cleaning the screen.

### DIFF
--- a/vlib/v/debug/debug.v
+++ b/vlib/v/debug/debug.v
@@ -274,7 +274,7 @@ pub fn (mut d Debugger) interact(info DebugContextInfo) ! {
 					}
 				}
 			}
-			'cls', 'clear' {
+			'clear' {
 				term.erase_clear()
 			}
 			'' {
--- a/vlib/v/debug/debug.v
+++ b/vlib/v/debug/debug.v
@@ -7,6 +7,7 @@ import os
 import math
 import readline
 import strings
+import term
 
 __global g_debugger = Debugger{}
 
@@ -37,6 +38,7 @@ mut:
 			'scope',
 			'unwatch',
 			'watch',
+			'clear',
 		]
 	}
 }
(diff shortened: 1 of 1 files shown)


vlib/v/debug/debug.v
```diff
@@ -271,6 +274,9 @@ pub fn (mut d Debugger) interact(info DebugContextInfo) ! {
 					}
 				}
 			}
+			'cls', 'clear' {
```
JalonSolov, line 277:
```suggestion
			'clear' {
```

vlib/v/debug/debug.v
```diff
@@ -117,6 +119,7 @@ fn (mut d Debugger) print_help() {
 	println('  scope\t\t\tshow the vars in the current scope')
 	println('  u, unwatch <var>\tunwatches a variable')
 	println('  w, watch <var>\twatches a variable')
+	println('  cls, clear\tClears the screen.')
```
JalonSolov, line 122: In the completion list, you only have `clear`, while you have `cls, clear` here.  I suggest this change:
```suggestion
	println('  clear\tClears the terminal window.')
```
since it doesn't clear the whole screen, only the terminal.

vlib/v/debug/debug.v
```diff
@@ -271,6 +274,9 @@ pub fn (mut d Debugger) interact(info DebugContextInfo) ! {
 					}
 				}
 			}
+			'clear' {
+				term.erase_clear()
```
spytheman, line 278: Why `term.erase_clear()`, instead of `term.clear()` ?

spytheman: `term.clear()` will have the same visible effect at the end, but will also allow you to scroll upwards, and see the previous lines.
With `term.erase_clear()`, all the lines in the current screen are erased, and if you scroll, you will not see them, you will only see the lines that were out of sight.

spytheman (approved):
Good work.

Issue #9838, opened 2021-04-21
String interpolation with brackets in templates


fnetX:
**V version:** V 0.2.2 0da827f.7b8957b
**OS:** linux, Debian GNU/Linux bullseye/sid

**What did you do?**
Call a function from a template which included a string interpolation with brackets (`{}`), looked somewhat like
~~~
@{string.replace("{{thing}}","${array[0]")}
~~~

I also tried to reproduce this with a less complex situation, you can find the attempt over here: https://codeberg.org/eventbike/vlang_plg/src/branch/main/misc/string_interpolation

**What did you expect to see?**
The string is interpolated, sent to the function and the return value printed in the template.
 
**What did you see instead?**
With my app: 
~~~
:21:1: error: unexpected token `}`, expecting `,`
~~~
With my less complex example:
~~~
:11:1: error: unexpected name `_tmpl_res_main__main`, expecting `,
~~~

kahsa:
still issue? if so, a minimally reproducible example would be helpful.

felipensp:
It is fixed.

```
felipe@~/github/v (master)$ v run bug.v
test

felipe@~/github/v (master)$ cat bug.v 
module main

fn main () {
        var := 'test'
        println($tmpl('test.tmpl'))
}

fn function (str string) string {
        return str
}
felipe@~/github/v (master)$ cat test.tmpl 
@{function("${var}")}
```

PR #20627, opened 2024-01-22
$tmpl: add test for #9838

pierrec:
Nothing to fix as it works with current master (0.4.4 1141432).


spytheman:
(rebased over current master)

spytheman (approved):

PR #20629, opened 2024-01-22
v: implement @[_linker_section] attribute

### LINKED ISSUE #20608: Enable V function placement into a desired object file section
### Describe the feature

A function attribute called `link_section` that enables specifying a section name in the resulting object file that this function's code should be placed in.

Eg:
```
@[linker_section = ".text.interrupt"]
fn interrupt_vector() {
        .
        .
}
```

The above should generate the following C declaration:
```
extern void interrupt_vector (void) __attribute__ ((section (".text.interrupt")));
```

### Use Case

Embedded systems programming often requires precise control of code (and data) placement in the final executable image. Examples are interrupt vectors, boot code etc.

C compilers support annotating functions with attributes that specify section names for this purpose.
Eg: `extern void foobar (void) __attribute__ ((section ("bar")));`

Modern programming languages that support systems programming - like Rust for example - have similar support.
Eg:
```
#[no_mangle]
#[link_section = ".example_section"]
fn example_function() {
    .
    .
}
```

### Proposed Solution

This probably requires mods to `v/vlib/v/gen/c/fn.v` perhaps in `gen_fn_decl`.
The implementation should include suitable test code etc.

### Other Information

_No response_

### Acknowledgements

### Version used

Current V version: V 0.4.4 df3591c, timestamp: 2024-01-20 19:54:12 +0200

This feature should be available on all targets where V can be run.
> [!NOTE]
>

felipensp:
Fix #20608

```V
@[linker_section:'foo']
fn test() {

}

fn main() {

}
```

```
$ objdump -t bug | grep test
000000000003bd1c g     F foo    0000000000000007              .hidden main__test
0000000000052be0 g     O .bss   0000000000000008              GC_greatest_plausible_heap_addr
00000000000183c6 g     F .text  0000000000000058              Array_u8_bytestr
00000000000171ce g     F .text  0000000000000007              .hidden float_test
```


vlib/v/gen/c/fn.v
```diff
@@ -323,7 +323,11 @@ fn (mut g Gen) gen_fn_decl(node &ast.FnDecl, skip bool) {
 	if node.params.len == 0 {
 		g.definitions.write_string('void')
 	}
-	g.definitions.writeln(');')
+	if attr := node.attrs.find_first('linker_section') {
```
raw-bin, line 326: I think the convention is to use an '_' prefix for such options (eg '_naked').

felipensp: Hmm, ok. thanks

spytheman (approved):
Excellent work 🥳 .

PR #20630, opened 2024-01-22
v.gen.c: initialize closures in shared library mode

### LINKED ISSUE #19446: call V from C not working when closure.
### Describe the bug

i tried calling v from c via shared lib. but it doesn't work on closure of v. `signal 11: segmentation fault`. but if not closure is not a problem.

### V code
```v
// foo.v
[export: 'foo']
fn show_foo(s &char) {
	my_fn := fn [s]() {
		println(unsafe { cstring_to_vstring(s) })
	}
	my_fn()
}
```
command: `v -cc gcc -shared foo.v`

### C code
```c
extern void foo(const char* s);

int main() {
	foo("bar");
}
```
compile: `gcc foo.c foo.so -o foo -Wl,-rpath=.`

next i call `./foo` in terminal.

### Expected Behavior

Expected output is `bar` in terminal.

### Current Behavior

```bash
signal 11: segmentation fault
| 0x7fdf57847090 | /lib/x86_64-linux-gnu/libc.so.6(+0x43090) 
```

### Possible Solution

_No response_

_No response_

### V version

Current V version: V 0.4.1 a8a3d3a, timestamp: 2023-09-26 07:02:39 +0300

```bash
V full version: V 0.4.1 20bce37.a8a3d3a
OS: linux, Pop!_OS 20.04 LTS
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-4030U CPU @ 1.90GHz

getwd: /home/herudi/deno/ffi-deno
vexe: /home/herudi/vlang/v
vexe mtime: 2023-09-26 04:11:20

vroot: OK, value: /home/herudi/vlang
VMODULES: OK, value: /home/herudi/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.25.1
Git vroot status: weekly.2023.28-458-ga8a3d3a6
.git/config present: true

```
> [!IMPORTANT]
> You can vote for this issue using the 👍 reaction. More votes increase the issue's priority
> for developers.
>
> Take into account that only the 👍 reaction counts as a vote.
> Only reactions to the issue itself will be counted as votes, not comments.

pierrec:
Fixes #19446.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -6185,6 +6185,9 @@ fn (mut g Gen) write_init_function() {
 		}
 		g.writeln('void _vinit_caller() {')
 		g.writeln('\tstatic bool once = false; if (once) {return;} once = true;')
+		if g.nr_closures > 0 {
+			g.writeln('\t__closure_init(); // vinit_caller()')
+		}
 		g.writeln('\t_vinit(0,0);')
 		g.writeln('}')
 
--- a/vlib/v/gen/c/cmain.v
+++ b/vlib/v/gen/c/cmain.v
@@ -98,7 +98,7 @@ fn (mut g Gen) gen_c_main_function_header() {
 	g.writeln('\tg_main_argc = ___argc;')
 	g.writeln('\tg_main_argv = ___argv;')
 	if g.nr_closures > 0 {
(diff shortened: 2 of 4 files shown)


spytheman (approved):
Excellent work.
The CI failures are not related.

PR #20633, opened 2024-01-22
os: Fix File.read() in JS backends [Issue : #20501]

### LINKED ISSUE #20501: File.read() always returns 0 bytes in JS backends
### Describe the bug

In both `js` and `js_browser` backends, `File.read()` call seems to read 0 bytes

### Reproduction Steps

```v
import os

mut buf := []u8{len: 10}
f := os.open('/etc/hosts') or { panic(err) }
n := f.read(mut &buf) or { panic(err) }
println(buf[..n])
```

### Expected Behavior

Using the `js` backend: expected to either read a non-zero amount of bytes, or raise an error.

Using the `js_browser`: expected to either fail compilation (since File I/O is unavailable in the browser), or for `os.open()` to always return an error.

### Current Behavior

```console
$ v run read_file.v    # runs as expected
[49, 50, 55, 46, 48, 46, 48, 46, 49, 32]
$  v -o read_file.js read_file.v && node read_file.js
[]
$ v -b js_browser -o read_file.js read_file.v && node read_file.js
[]
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.4 ac2dcc2

NixOS 23.11 (Tapir)
Node v18.18.2
> [!NOTE]
>

GGRei:
Fixed #20501 File.read() function for the JavaScript backend. 

The issue was due to incorrect array access in the JavaScript generated from V code.

To respond at ticket expectation : 

- _Using the js backend: expected to either read a non-zero amount of bytes, or raise an error._

The original function already includes a try-catch block to manage errors. Unless specifically requested, I don't plan to add additional error handling

_- Using the js_browser: expected to either fail compilation (since File I/O is unavailable in the browser), or for os.open() to always return an error._

the existing code already handles errors. If **open_file()** is called in a non-NodeJS environment, an error is generated. However, it appears that the test in ticket was conducted using Node.js, which could explain the observed results.

I'm always available for any suggestions/modifications.

### DIFF
--- a/vlib/os/file.js.v
+++ b/vlib/os/file.js.v
@@ -84,9 +84,10 @@ pub fn (f &File) read(mut buf []u8) !int {
 	}
 	mut nbytes := 0
 	#try {
-	#let buffer = $fs.readFileSync(f.val.fd.valueOf());
+	#let readBuffer = Buffer.alloc(buf.val.len.valueOf());
+	#nbytes = $fs.readSync(f.val.fd.valueOf(), readBuffer, 0, buf.val.len.valueOf(), null);
 	#
-	#for (const val of buffer.values()) { buf.val.arr.arr[nbytes++] = new u8(val); }
+	#for (let i = 0; i < nbytes; i++) { buf.val.arr.arr[i] = new u8(readBuffer[i]); }
 	#}
 	#catch (e) { return error('' + e); }
 
--- a/vlib/os/file.js.v
+++ b/vlib/os/file.js.v
@@ -86,7 +86,7 @@ pub fn (f &File) read(mut buf []u8) !int {
 	#try {
 	#let buffer = $fs.readFileSync(f.val.fd.valueOf());
 	#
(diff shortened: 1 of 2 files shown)


spytheman:
image

The observed behavior for the C backend and for the JS backend is still different.
You can use this file `open_and_read_from_file_test.js.v` to try (and please add it to vlib/os/ for example, to prevent future regressions).
```v
import os

fn test_read_from_file() {
	mut buf := []u8{len: 10}
	f := os.open(@FILE)!
	n := f.read(mut &buf)!
	println(buf[..n])
}	
```

spytheman:
Note that the C version https://github.com/vlang/v/blob/master/vlib/os/file.c.v#L233 , reads up to `buf.len` bytes, and it returns how many bytes it could read.
It does not read the whole file.

GGRei:
Done. let me know if everything is okay. Thanks.

spytheman:
I can confirm, that the fix for File.read/1 is working.

It still prints differently, but fixing that is out of scope for this PR.

Excellent work.

PR #20642, opened 2024-01-23
x.vweb: add new sessions module

Casper64:
Add new sessions module for `x.vweb`.

The main goal of this session module is to provide a way of using sessions easily with vweb while also making it usable outside vweb.

Documentation is inside the README file.
Tests are also included.

## Stores

The session store handles the saving, storing and retrieving of data. There are currently two implementations:
1. The `MemoryStore[T]` stores session data in memory only using the `map` datatype. 
2. The `DBStore[T]` stores session data in a database by encoding the session data to JSON.

It is possible to create your own implementation of a session store. You only need to follow the `Store` interface. See "Custom Stores" in the README.

## Usage

This module provides a `Sessions` struct which the user can embed in their vweb app.  It makes the usage of sessions easier, for example session data is retrieved automatically when the provided middleware is used. The struct stores session id's in a cookie and you can use any session store.

See the Usage section in the README file for more information and on how to use this module outside vweb, or without the `Sessions` struct.


vlib/x/vweb/sessions/README.md
```diff
@@ -0,0 +1,330 @@
...
+and set as cookie. Assuming there wasn't already a session going on. If you want to 
+be sure that a new session id is generated when you save data, you can use the `resave`
+method. This method will save the data and *always* set a new session id.
+
+**Example:**
+```v
+pub fn (mut app App) login(mut ctx Context) vweb.Result {
+	// use resave, because the authentication status of the session changes.
+	// this function will set a new session id and destroy old session data
+	app.sessions.save(mut ctx, User{
```
xqliang, line 155: save -> resave

vlib/x/vweb/sessions/db_store.v
```diff
@@ -0,0 +1,108 @@
...
+pub struct DBStore[T] {
+pub mut:
+	db orm.Connection @[required]
+}
+
+// create a new Database store with a connection to a database
+pub fn DBStore.create[T](db orm.Connection) !DBStore[T] {
+	sql db {
+		create table DBStoreSessions
+	}!
```
spytheman, line 25: What happens if the table `DBStoreSessions` already existed?

vlib/x/vweb/sessions/tests/db_store_test.v
```diff
@@ -0,0 +1,94 @@
...
+	// verify that data is deleted
+	rows := sql db {
+		select from sessions.DBStoreSessions where session_id == 'c'
+	}!
+	assert rows.len == 0
+}
+
+fn testsuite_end() {
+	os.rm(db_path) or {}
+}
```
spytheman, line 90: It is a bit easier to keep setup/cleanup actions in sync, when they are spatially close in the source code. In this case, I think that moving testsuite_end right after testsuite_begin will be beneficial for maintainance.

vlib/x/vweb/sessions/sessions.v
```diff
@@ -0,0 +1,204 @@
...
+// This function also destroys the data associtated to the old session id.
+pub fn (mut s Sessions[T]) resave[X](mut ctx X, data T) {
+	if sid := s.get_session_id(ctx) {
+		s.store.destroy(sid)
+	}
+
+	s.save(mut ctx, data)
+}
+
+// get the current session id, if it is set
```
spytheman, line 153:
```suggestion
// get_session_id retrieves the current session id, if it is set
```

vlib/x/vweb/sessions/sessions.v
```diff
@@ -0,0 +1,204 @@
...
+			// when data is saved. So we create the session here
+			sid := s.set_session_id(mut ctx)
+			s.store.set(sid, data)
+			ctx.CurrentSession.session_data = data
+		}
+		eprintln('[vweb.sessions] error: trying to save data without a valid session!')
+	}
+}
+
+// Save `data` for the current session and reset the session id.
```
spytheman, line 141:
```suggestion
// resave saves `data` for the current session and resets the session id.
```

vlib/x/vweb/sessions/sessions.v
```diff
@@ -0,0 +1,204 @@
...
+	})
+	// indicate that the response should not be cached: we don't want the session id cookie
+	// to be cached by the browser, or any other agent
+	// https://datatracker.ietf.org/doc/html/rfc7234#section-5.2.1.4
+	ctx.res.header.add(.cache_control, 'no-cache="Set-Cookie"')
+
+	return sid
+}
+
+// validate the current session, returns the session id and the validation status
```
spytheman, line 92:
```suggestion
// validate_session validates the current session, returns the session id and the validation status
```

vlib/x/vweb/sessions/sessions.v
```diff
@@ -0,0 +1,204 @@
...
+	// max age of session data and id, default is 30 days
+	max_age time.Duration = time.hour * 24 * 30
+	// set to true if you want to create a session if there isn't any data stored yet.
+	// Also called pre-sessions
+	save_uninitialized bool
+pub mut:
+	store Store[T] @[required]
+}
+
+// generate a new session id and set a Set-Cookie header on the response
```
spytheman, line 69:
```suggestion
// set_session_id generates a new session id and sets a Set-Cookie header on the response
```

vlib/x/vweb/sessions/sessions.v
```diff
@@ -0,0 +1,204 @@
...
+	cookie_name string = 'sid'
+	domain      string
+	http_only   bool          = true
+	path        string        = '/'
+	same_site   http.SameSite = .same_site_strict_mode
+	secure      bool
+}
+
+@[heap]
+pub struct Sessions[T] {
```
spytheman, line 56: The structure needs a comment too. It is public.

vlib/x/vweb/sessions/sessions.v
```diff
@@ -0,0 +1,204 @@
...
+	return sid, hmac.equal(actual_hmac, new_hmac)
+}
+
+pub struct CurrentSession[T] {
+pub mut:
+	session_id   string
+	session_data ?T
+}
+
+pub struct CookieOptions {
```
spytheman, line 46: New public structures should be documented.

vlib/x/vweb/sessions/sessions.v
```diff
@@ -0,0 +1,204 @@
...
+pub fn new_session_id(secret []u8) (string, string) {
+	sid := rand.hex(sessions.session_id_length)
+
+	hashed := hmac.new(secret, sid.bytes(), sha256.sum, sha256.block_size)
+
+	// seperate session id and hmac with a `.`
+	return sid, '${sid}.${base64.url_encode(hashed)}'
+}
+
+// verify the signed session id with `secret`. This function returns the session id
```
spytheman, line 24:
```suggestion
// verify_session_id verifies the signed session id with `secret`. This function returns the session id
```

vlib/x/vweb/sessions/sessions.v
```diff
@@ -0,0 +1,204 @@
...
+import crypto.hmac
+import encoding.base64
+import net.http
+import rand
+import time
+import x.vweb
+
+const session_id_length = 32
+
+// create and return a random session id and its signed version. You can directly use
```
spytheman, line 13:
```suggestion
// new_session_id creates and returns a random session id and its signed version. You can directly use
```

vlib/x/vweb/sessions/db_store.v
```diff
@@ -0,0 +1,108 @@
...
+
+	return DBStore[T]{
+		db: db
+	}
+}
+
+// get data from all sessions
```
spytheman, line 32:
```suggestion
// all gets the data from all sessions
```

vlib/x/vweb/sessions/db_store.v
```diff
@@ -0,0 +1,108 @@
+module sessions
+
+import json
+import orm
+import time
+
+pub struct DBStoreSessions {
```
spytheman, line 7: Public structs should be documented.

vlib/x/vweb/sessions/sessions.v
```diff
@@ -0,0 +1,204 @@
...
+		return none
+	}
+}
+
+// You can add this middleware to your vweb app to ensure a valid session always
+// exists. If a valid session exists the session data will be loaded into
+// `session_data`, else a new session id will be generated.
+// You have to pass the Context type as the generic type
+// Example: app.use(app.sessions.middleware[Context]())
+pub fn (mut s Sessions[T]) middleware[X]() vweb.MiddlewareOptions[X] {
```
spytheman, line 181: This needs a doc comment too (starting with the method name).

I am a bit more concerned about the utility of this method though.
It is the only thing that couples `Sessions` to `vweb.MiddlewareOptions` afaik.

spytheman: Instead, can we add this `vlib/x/vweb/sessions/middleware/create.v`:
```v
module middleware

import x.vweb
import x.vweb.sessions

// You can add this middleware to your vweb app to ensure a valid session always
// exists. If a valid session exists the session data will be loaded into
// `session_data`, else a new session id will be generated.
// You have to pass the Context type as the generic type
// Example: app.use(app.sessions.middleware[Context]())
pub fn create[T,X](mut s sessions.Sessions[T]) vweb.MiddlewareOptions[X] {
        return vweb.MiddlewareOptions[X]{
                handler: fn [mut s] [T, X](mut ctx X) bool {
                        // a session id is retrieved from the client, so it must be considered
                        // untrusted and has to be verified on every request
                        sid, valid := s.validate_session(ctx)

                        if !valid {
                                if s.save_uninitialized {
                                        // invalid session id, so create a new one
                                        s.set_session_id(mut ctx)
                                }
                                return true
                        }

                        ctx.CurrentSession.session_id = sid
                        if data := s.store.get(sid, s.max_age) {
                                ctx.CurrentSession.session_data = data
                        }

                        return true
                }
        }
}
```
(the name can be different of course, the goal is just extracting it into its own separate module)

spytheman: Then the patch for `vlib/x/vweb/sessions/tests/session_app_test.v` becomes:
```v
diff --git vlib/x/vweb/sessions/tests/session_app_test.v vlib/x/vweb/sessions/tests/session_app_test.v
index 4d1c4530b..99ef4e671 100644
--- vlib/x/vweb/sessions/tests/session_app_test.v
+++ vlib/x/vweb/sessions/tests/session_app_test.v
@@ -1,8 +1,8 @@
 import net.http
-import os
 import time
 import x.vweb
 import x.vweb.sessions
+import x.vweb.sessions.middleware
 
 const port = 13010
 const localserver = 'http://localhost:${port}'
@@ -98,7 +98,7 @@ fn testsuite_begin() {
 		}
 	}
 
-	app.use(app.sessions.middleware[Context]())
+	app.use( middleware.create[User,Context](mut app.sessions) )
 
 	spawn vweb.run_at[App, Context](mut app, port: port, timeout_in_seconds: 2)
 	// app startup time
```

spytheman: Given that decoupling, the whole `x.vweb.sessions` module can be moved to `x.sessions`, and potentially reused by vweb or other web frameworks as well?

Casper64: Decoupling the sessions module from vweb could be an option. Just to make sure we're on the same page:

1. Yes the `middleware` method is the only thing that couples `Session` to `vweb.MiddlewareOptions`. I intended it as an easy to use method and I think 9/10 of the users will end up using it; that's why I added it as a method instead of a separate function.
2. If the sessions module and vweb are to be decoupled, would the `Sessions` struct be in `x,.sessions` or `x.vweb.sessions`? Since it has to be used in combination with `vweb`.
3. It makes sense to me to put `Sessions` in `x.sessions`. But then I don't know what the best place for the middleware function is. Maybe inside `vlib/x/vweb/middleware.v`, but then anyone who imports `x.vweb` relies on `x.sessions`, so separating in a it's own module makes sense, but you end up with a submodule with a single function :/

I cannot clearly picture the decoupling in my head, so I'm a bit biased against it. But on the other hand a general sessions module makes room for a lot of new features and improvement, which I'm excited about!

What do you think @JalonSolov  @spytheman ?

vlib/x/vweb/sessions/db_store.v
Casper64: Nothing, it's a `CREATE TABLE IF NOT EXISTS`  the warning will be ignored by `orm`. I will add to the documentation that the `DBStore` creates a table with the name `DBStoreSessions`.

vlib/x/vweb/sessions/sessions.v
JalonSolov: Wouldn't it live in `net.sessions` (or `x.net.sessions` to start), since it is network related, but could be used in direct picoev apps (for example), rather than just vweb.

spytheman: > you end up with a submodule with a single function :/

That is not a problem. Users that do not want it, can just not import it.

spytheman: > If the sessions module and vweb are to be decoupled, would the Sessions struct be in x,.sessions or x.vweb.sessions? Since it has to be used in combination with vweb.

The sessions module can be `vlib/x/sessions` for now, until it matures.

The middleware helper module, with the create function, can go to `vlib/x/sessions/vweb2_middleware/` .

spytheman: > Wouldn't it live in net.sessions (or x.net.sessions to start), since it is network related, but could be used in direct picoev apps (for example), rather than just vweb.

No, it is not directly network related.
In many systems, you may want to do background server tasks (for example cronjobs), that also want to access user sessions, without necessarily doing network operations themselves.

spytheman: >  I intended it as an easy to use method and I think 9/10 of the users will end up using it; that's why I added it as a method instead of a separate function.

You are right, that sessions can be useful for many web server users.
With your middleware helper, that boils down to just 2 lines more(one for the import, and one to call the helper).

With the coupling however, 100% of users of x.sessions, will also get the code for x.vweb, even if they are not using that web framework, but another.
It practically ensures that module will not be used, except for x.vweb apps specifically.

Casper64: I moved the sessions module to `x.sessions` and decoupled it from `x.vweb`, by making the middleware into a separate module like you suggested. Looks better now 👍 . Thanks for your feedback!

spytheman (approved):
Excellent work 🚀 .

PR #20646, opened 2024-01-24
v.checker: disallow assigning none to _

### LINKED ISSUE #20645: cgen error for ` _ = none`
```bash
v run a.v
```
```
==================
/tmp/v_2625/a.01HMWV3D4QDHES1C1B00P00B2E.tmp.c:12858: error: '{' expected (got ";")
...
==================
builder error:
==================

```
```bash
cat a.v
```
```v
fn x() ?bool {
        return none
}                                                                  
_ = x() or {
        _ = none
        true
}
```
I have no idea what's happening here, I was experimenting as I'm still learning the language and then it asked me to report.

and I want to leverage this opportunity to ask an off-topic little doubt,
in structs, what `__global:` allows that `pub mut:` doesn't?
I tried them in various forms and both worked fine,
the docs talks about "parent module only" stuff with `pub mut:` but I don't get it they both work no matter what I try.
> [!NOTE]
>

pierrec:
Fixes #20645.

### DIFF
--- a/vlib/v/checker/tests/assign_blank_none_err.out
+++ b/vlib/v/checker/tests/assign_blank_none_err.out
@@ -1,5 +1,12 @@
-vlib/v/checker/tests/assign_blank_none_err.vv:2:2: error: cannot assign none to blank `_` identifier
+vlib/v/checker/tests/assign_blank_none_err.vv:2:7: error: cannot assign a `none` value to a variable
     1 | fn main() {
-    2 |     _ = none
-      |     ^
-    3 | }
+    2 |     _ := none
+      |          ~~~~
+    3 |     _ = none
+    4 | }
+vlib/v/checker/tests/assign_blank_none_err.vv:3:6: error: cannot assign a `none` value to blank `_` identifier
+    1 | fn main() {
+    2 |     _ := none
+    3 |     _ = none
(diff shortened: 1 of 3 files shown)


vlib/v/checker/tests/assign_blank_none_err.out
```diff
@@ -0,0 +1,5 @@
+vlib/v/checker/tests/assign_blank_none_err.vv:2:2: error: cannot assign none to blank `_` identifier
+    1 | fn main() {
+    2 |     _ = none
+      |     ^
+    3 | }
```
Delta456, line 5: Add new line at the end

pierrec: @Delta456 why is a newline required here?

shove70 (changes requested):

Delta456:
Run `v fmt -w vlib/v/checker/assign.v`

shove70 (approved):

vlib/v/checker/tests/assign_blank_none_err.out
JalonSolov: It's only needed in a .out file if it helps match against the test output.

PR #20648, opened 2024-01-24
Strings/builder with buffer

enghitalo:
### Why

- 3.75x faster than `strings.new_builder` (Can be improved)
```
 SPENT   308.734 ms in new_builder_with_buffer([]u8{cap: n})
 SPENT  1191.929 ms in strings.new_builder(n)
```
- No breaking change
- Explicit use of the []u8
- Possibility of reusing an existing buffer


vlib/strings/builder_with_buffer.c.v
```diff
@@ -0,0 +1,48 @@
...
+
+// write_string
+@[direct_array_access]
+pub fn (mut b BuilderWithBuffer) write_string(s string) {
+	if s.len == 0 {
+		return
+	}
+	// try keep at leat one zero at end
+	if b.char_count + s.len + 1 > b.initial_buffer_cap {
+		eprintln('string overflowing inital capacity. To performance improvement increase array capacity from []u8{cap: ${b.initial_buffer_cap}} to []u8{cap: ${
```
spytheman, line 27: Either remove that eprintln, or put it behind a comptime flag like `$if trace_write_string ? {`.

spytheman:
I've tested a lot with a variation of the test here (needed to make both measurements as similar to each other):
```v
import strings
import benchmark

const max_iterations = 10_000_000

fn test_compatibility() {
	mut b := benchmark.start()
	mut new_sb := strings.new_builder_with_buffer([]u8{cap: max_iterations * 2 + 1})
	for _ in 0 .. max_iterations {
		new_sb.write_string('87')
	}
	str_from_new_sb := new_sb.str()
	b.measure('new_builder_with_buffer([]u8{cap: ${max_iterations}})')

	mut old_sb := strings.new_builder(3 * max_iterations)
	for _ in 0 .. max_iterations {
		old_sb.write_string('87')
	}
	str_from_old_sb := old_sb.str()
	b.measure('strings.new_builder(${max_iterations})')

	assert str_from_old_sb == str_from_new_sb

	unsafe { old_sb.free() }
	unsafe { new_sb.free() }
}
```

The results on my machine were inconclusive:
image

Note that the only combination, where the loop using `strings.new_builder_with_buffer` was faster than the loop using `strings.new_builder`, was when the test was compiled with `-cc gcc -prod`, and that with `-cc clang -prod` the results were reversed.

I've also tested another variation, with:
```v
import os
import strings
import benchmark

const max_iterations = os.getenv_opt('MAX_ITERATIONS') or { '1_000_000' }.int()

fn test_buf_memory() {
        mut new_sb := strings.new_builder_with_buffer([]u8{cap: 3})
        mut new_sb2 := strings.new_builder_with_buffer([]u8{cap: 3})

        new_sb.write_string('87')
        new_sb2.write_string('87')

        assert ptr_str(new_sb.str().str) != ptr_str(new_sb2.str().str)

        unsafe { new_sb.free() }
}

fn test_compatibility() {
        mut b := benchmark.start()
        mut new_sb := strings.new_builder_with_buffer([]u8{cap: max_iterations * 2 + 1})
        for _ in 0 .. max_iterations {
                new_sb.write_string('87')
        }
        str_from_new_sb := new_sb.str()
        b.measure('new_builder_with_buffer([]u8{cap: ${max_iterations}})')

        mut old_sb := strings.new_builder(3 * max_iterations)
        for _ in 0 .. max_iterations {
                old_sb.write_string('87')
        }
        str_from_old_sb := old_sb.str()
        b.measure('strings.new_builder(${max_iterations})')

        assert str_from_old_sb == str_from_new_sb

        unsafe { old_sb.free() }
        unsafe { new_sb.free() }
}
```
where the meaningful change was using:
`const max_iterations = os.getenv_opt('MAX_ITERATIONS') or { '1_000_000' }.int()`
instead of `const max_iterations = 10_000_000`, i.e. a constant whose value is calculated at runtime, instead of at comptime, and the results were:
image


The above tests were done on Ubuntu 20.04, Intel(R) Core(TM) i3-3225 CPU @ 3.30GHz, with gcc 10.5, and clang 10.0 . I've also did test on my m1 with clang 15.0 , and the results were ~ the same.

spytheman:
As for:
> Possibility of reusing an existing buffer

That can be achieved by just adding another constructor function, there is no need of a separate implementation.

Issue #20565, opened 2024-01-17
How should I obtain the field names of SQL?


xiusin:
### Describe the bug

like :
```vlang
mut db := sqlite.connect(':memory:')!
result := db.exec('select max(id) as max_id from db')!



```
How do I know the max_id in the field list?

xiusin:
How do I know the max_id in the field list?

shove70:
V don't have that yet, I think the field names are already known, and for some aggregate functions we should use "as" to determine the field names

xiusin:
@shove70 If you want to write a custom SQL statement, it is almost impossible to get the field list directly, unless you use some analytical SQL statements to disassemble the field list

shove70:
hmm, It's already marked as a `feature request`

xiusin:
@shove70 MySQL can use the following function to get the list of fields:

```vlang
result := db.query('SELECT max(id) as max_id FROM jobs')!
dump(result.fields())
```

result:
```
[db.v:163] result.fields(): [
{
        name: "max_id"
        org_name: ""
        table: ""
        org_table: ""
        db: ""
        catalog: "def"
        def: ""
        length: 11
        max_length: 0
        name_length: 6
        org_name_length: 0
        table_length: 0
        org_table_length: 0
        db_length: 0
        catalog_length: 3
        def_length: 0
        flags: 32896
        decimals: 0
        charsetnr: 63
        type: long
}
]
```

JalonSolov:
Is that a list of fields, or a description of the `max_id` field?

The proper way to show fields in a table is `SHOW COLUMNS FROM <table name>`

shove70:
> ```v
> result := db.query('SELECT max(id) as max_id FROM jobs')!
> dump(result.fields())
> ```

Yes, currently mysql, sqlite, mssql, pg several databases have different wrappe ways for query results, sqlite query results only contain []row data, no column information.

This feature request may be scheduled for processing in due course

xiusin:
@shove70 `C.sqlite3_column_name` I think this should be a function to get the field name, but it's not a public function

xiusin:
@JalonSolov Most of our usage scenarios are not simply viewing the structure of a data table, but complex queries, which cannot be applied in this case.

shove70:
> @shove70 `C.sqlite3_column_name` I think this should be a function to get the field name, but it's not a public function

Yes, it is a field name query against the query result set.
[...]

PR #20650, opened 2024-01-25
sqlite: Functions thats returns the values resulting from a 'SELECT' query and name of column - Feature ISSUE #20565

viniciusfdasilva:
#### This pull request addresses the feature mentioned in issue #20565  and implements a function that returns the values resulting from a  'SELECT' query and the name of each column. If an alias is provided  either through the 'as' command or not, the returned value becomes  the alias.

['SELECT SYNTAX' - Documentation](https://www.sqlite.org/lang_select.html)
['JOIN SYNTAX' - Documentation](https://www.sqlite.org/syntax/join-clause.html)

[Source code](https://github.com/viniciusfdasilva/v/blob/master/vlib/db/sqlite/sqlite.c.v#L258-L358)

#### All fields

```v
r := db.get_queryset('SELECT * FROM repository')! 
```

query1

#### All names with explicit names

```v
r := db.get_queryset('SELECT name, id FROM repository')!
```

query2

#### All fields with alias

```v
r := db.get_queryset('SELECT name as n, id i FROM repository')!
```

query3

#### Specific field

```v
r := db.get_queryset('SELECT name FROM repository')! 
```

query4

#### Specific field with alias and 'as'

```v
r := db.get_queryset('SELECT name as n FROM repository')! 
```

query5e6

#### Specific field with alias without 'as'

```v
r := db.get_queryset('SELECT name n FROM repository')! 
```

query5e6

#### OUTPUT TESTS

test

### DIFF
--- a/vlib/db/sqlite/sqlite.c.v
+++ b/vlib/db/sqlite/sqlite.c.v
@@ -1,5 +1,7 @@
 module sqlite
 
+import regex { regex_opt }
+
 $if freebsd || openbsd {
 	#flag -I/usr/local/include
 	#flag -L/usr/local/lib
@@ -83,6 +85,11 @@ pub fn (db &DB) str() string {
 pub struct Row {
 pub mut:
 	vals []string
+}
+
+pub struct QuerySet {
+pub mut:
+	vals map[string]string
 }
 
 //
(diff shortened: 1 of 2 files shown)


vlib/db/sqlite/sqlite.c.v
```diff
@@ -248,6 +255,108 @@ pub fn (db &DB) exec(query string) ![]Row {
 	return rows
 }
 
+@[manualfree]
+pub fn (db &DB) get_queryset(query string) ![]QuerySet {
```
spytheman, line 259: The new public function should be documented with a comment, that `v doc` can understand.
Use something like this above the header:
```v
// get_queryset returns ...
```

spytheman: It is only important for `v doc`, that it starts with the name of the method, the explanation after that can be whatever makes sense.

vlib/db/sqlite/sqlite_test.v
```diff
@@ -79,6 +79,72 @@ fn test_sqlite() {
 	db.exec("delete from users where name='Sam'")!
 	assert db.get_affected_rows_count() == 1
 
+	db.get_queryset('SELECT * FROM users') or { panic(err) }
```
spytheman, line 82: Inside test_ functions, you can just use `db.get_queryset('SELECT * FROM users')!`, without the repetitive `or { panic(err) }`. The test_ functions act as `fn main() {` in that regard, since they are called by generated code, not by other user functions.

vlib/db/sqlite/sqlite.c.v
```diff
@@ -248,6 +255,108 @@ pub fn (db &DB) exec(query string) ![]Row {
...
+	query_lower := query.to_lower()
+
+	// 'select' syntax verified on: https://www.sqlite.org/lang_select.html and
+	// https://www.sqlite.org/syntax/join-clause.html
+	mut select_header := regex_opt(r'select((\s)+(all)|(distinct)(\s)+)|(\s)+')!
+	mut from := regex_opt(r'(\s)+from(\s)+')!
```
spytheman, line 265: These seem static to me. Extracting them to a const, may be beneficial.

spytheman, line 265: hm, why is regex matching needed at all?

vlib/db/sqlite/sqlite.c.v
```diff
@@ -248,6 +255,108 @@ pub fn (db &DB) exec(query string) ![]Row {
...
+
+	// This check eliminates queries that do not have the SELECT statement
+	if query_lower.count('select') == 1 {
+		// or do not include 'FROM', just like 'SELECT 1 + 1'
+		if select_header.matches_string(query_lower) && query_lower.contains(r'from') {
+			// The execution of this function indicates that the passed
+			// query is syntactically correct, which is why no additional verification
+			rows := db.exec(query)!
```
spytheman, line 273: `rows := db.exec(query)!` will return an error itself, if the query has invalid syntax.

vlib/db/sqlite/sqlite.c.v
```diff
@@ -248,6 +255,108 @@ pub fn (db &DB) exec(query string) ![]Row {
...
+
+			defer {
+				unsafe { rows.free() }
+			}
+
+			if rows.len == 0 {
+				return SQLError{
+					msg: 'No rows'
+					code: sqlite.sqlite_done
+				}
```
spytheman, line 283: Why is that an error?
Returning an empty set is perfectly valid thing to do in SQL.

vlib/db/sqlite/sqlite.c.v
viniciusfdasilva: I use the string query itself to obtain the names of the columns if they have been specified, as well as the name of the table! Since there are several ways to initiate the 'SELECT' command [doc](https://www.sqlite.org/lang_select.html), I constructed a regex based on the syntax to facilitate obtaining the column names and the table name!

vlib/db/sqlite/sqlite.c.v
viniciusfdasilva: Correct, that's why I don't perform any verification other than simply checking if the query is a "select" statement beforehand. If the query isn't a select statement, I don't execute it at all! 

That way, I save execution time for a query that isn't even a 'select'.

vlib/db/sqlite/sqlite.c.v
```diff
@@ -248,6 +255,101 @@ pub fn (db &DB) exec(query string) ![]Row {
...
+
+		defer {
+			unsafe { rows.free() }
+		}
+
+		if rows.len == 0 {
+			return []QuerySet{}
+		} else {
+			// Finding final index of select((\s)+(all)|(distinct)(\s)+)|(\s)+ inside query_lower string
+			_, end_select := select_header.match_string(query_lower)
```
ttytm, line 282: As an aspect of readability, I would mention avoiding unnecessary nested code, with a reference to the book [100 Go Mistakes and How to Avoid Them book](https://100go.co/book/). 
> When an if block returns, we should omit the else block [...]

JalonSolov:
Any plans to finish this PR?

spytheman:
Closing for inactivity.

PR #20654, opened 2024-01-25
io: Add read_line_until() to BufferedReader

syrmel:
Enable BufferedReader to work consistently with lines terminated by characters other than \n (such as \0 when using coreutils with the zero-terminated option). read_line() will call read_line_until(`\n`) to behave as before.

Test for the new function has been added in reader_test.v; tests for existing read_line() pass as before.

### DIFF
--- a/examples/xvweb/todo/main.v
+++ b/examples/xvweb/todo/main.v
@@ -6,6 +6,7 @@ module main
 
 import x.vweb
 import db.sqlite
+import os
 import time
 
 struct Todo {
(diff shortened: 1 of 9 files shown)


JalonSolov:
Hmm... it would've been more "the V way" to make a Params struct, with a delimiter as the only parameters, and a default value of `\n`.  Then someone could do `read_line()` to get the current, or `read_line(delim: \`0\`)`.  Perhaps even just `read_line(\`0\`)`.

syrmel:
Live and learn... I like it. Will create an updated PR.

PR #20658, opened 2024-01-25
x.vweb: support HTTP 1.1 persistent connections

Casper64:
Allow persistent HTTP connections so multiple requests and responses can be sent over a single connection.

Remove the default behaviour of vweb to always send the `Connection: close` header. The connection will be closed when the client sends the `Connection: close` header in the request, or if a read on the socket returns `0` indicating that the client has closed the connection. According to HTTP 1.1 standard.

This pr greatly improves the speed of `x.vweb` for clients that support HTTP 1.1 and higher


spytheman (approved):
Excellent work.

PR #20671, opened 2024-01-27
os: Add proper process termination in Linux

SewerynKaminski:
There is only signal_kill() option to terminate a process wich is not working.
The proper way is to send TERM signal.

### DIFF
--- a/vlib/os/process.c.v
+++ b/vlib/os/process.c.v
@@ -302,7 +302,7 @@ fn (mut p Process) _signal_kill() {
 // _signal_term - should not be called directly, except by p.signal_term
 fn (mut p Process) _signal_term() {
 	$if windows {
-		p.win_kill_process()
+		p.win_term_process()
 	} $else {
 		p.unix_term_process()
 	}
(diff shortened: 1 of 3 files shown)


spytheman:
You have to also add shims for `unix_kill_process` to `vlib/os/process_windows.c.v`, and then `v fmt -w vlib/os` .

spytheman:
A shim for `unix_term_process` should be also added to `vlib/os/process_windows.c.v`, and a shim for `win_term_process` should be also added to `vlib/os/process_nix.c.v`. See the bottom of those files.

The current implementation is a bit convoluted, but it is needed to make the generation of the vc/v.c and vc/v_win.c (bootstrapping) reliable.

SewerynKaminski:
OK. I missed it :) Corrected.

vlib/os/process.c.v
```diff
@@ -17,6 +17,15 @@ pub fn (mut p Process) signal_kill() {
 	return
 }
 
+// signal_term - terminate the process
+pub fn (mut p Process) signal_term() {
+	if p.status !in [.running, .stopped] {
+		return
+	}
+	p._signal_term()
+	return
```
StunxFS, line 26: That "return" is not necessary

SewerynKaminski: Yes there are 7 more `return` not necessary

spytheman (approved):
Excellent work.

PR #20672, opened 2024-01-27
v.fmt: drop newline in end comments for const

### LINKED ISSUE #20657: vfmt adds unintended? line breaks after const inline comments
### Describe the bug

```v
const a = 'a'
const b = 'b' // comment
const c = 'c'
```

will become

```v
const a = 'a'
const b = 'b' // comment

const c = 'c'
```

### Reproduction Steps

-

### Expected Behavior

won't add a line break between constants.

### Current Behavior

vfmt adds a line break.

### Possible Solution

_No response_

_No response_

### V version

0.4.4

linux, amd64
> [!NOTE]
>

pierrec:
Fixes #20657.


spytheman:
Please also add the sample from the issue to a `_keep.vv` file under `vlib/v/fmt/tests/` .

spytheman (approved):
Excellent work.

PR #20674, opened 2024-01-27
crypto: opt aes performance

kbkpbot:
Performance opt for AES module.

Loop 100000 times test_aes()
From 800.566 ms down to 261.793 ms;
From 112.662 ms down to 47.344 ms ( -prod)


spytheman (approved):
Excellent work.

PR #20677, opened 2024-01-27
v.gen.c: write profile out upon CTRL-C or kill

### LINKED ISSUE #8264: Profiler needs to write output no matter how a program exits.
The profiler built into V is useful, but limited, as it only gives output if a program exits cleanly.
```
OS: linux, Ubuntu 20.04.1 LTS (WSL 2)
Processor: 32 cpus, 64bit, little endian, AMD Ryzen Threadripper 1950X 16-Core Processor
CC version: cc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0

getwd: /home/jalon
vmodules: /home/jalon/.vmodules
vroot: /home/jalon/git/v
vexe: /home/jalon/git/v/v
vexe mtime: 2021-01-21 21:13:34
is vroot writable: true
is vmodules writable: true
V full version: V 0.2.1 fafd949.a569dc1

Git version: git version 2.25.1
Git vroot status: weekly.2020.49.5-930-ga569dc17
.git/config present: true
thirdparty/tcc status: thirdparty-linux-amd64 7543de81
```
**What did you do?**
Doesn't matter what you run... just hit Ctrl-C to exit.
```
jalon@Ripper:~/git/v$ v -profile prof.out run examples/tetris
^CTerminated by signal  2 (SIGINT)
jalon@Ripper:~/git/v$ ls -l prof.out
ls: cannot access 'prof.out': No such file or directory
jalon@Ripper:~/git/v$
```

Profile output written to prof.out file. 

**What did you see instead?**

Just came back to a prompt, and no file written.

pierrec:
Fixes #8264.

### DIFF
--- a/vlib/v/gen/c/cmain.v
+++ b/vlib/v/gen/c/cmain.v
@@ -220,6 +220,8 @@ pub fn (mut g Gen) gen_failing_return_error_for_test_fn(return_stmt ast.Return, 
 pub fn (mut g Gen) gen_c_main_profile_hook() {
 	if g.pref.is_prof {
 		g.writeln('')
+		g.writeln('\tsignal(SIGINT, vprint_profile_stats);')
+		g.writeln('\tsignal(SIGTERM, vprint_profile_stats);')
 		g.writeln('\tatexit(vprint_profile_stats);')
 		g.writeln('')
 	}
--- a/vlib/v/gen/c/cmain.v
+++ b/vlib/v/gen/c/cmain.v
@@ -220,8 +220,8 @@ pub fn (mut g Gen) gen_failing_return_error_for_test_fn(return_stmt ast.Return, 
 pub fn (mut g Gen) gen_c_main_profile_hook() {
 	if g.pref.is_prof {
 		g.writeln('')
-		g.writeln('\tsignal(SIGINT, vprint_profile_stats);')
-		g.writeln('\tsignal(SIGTERM, vprint_profile_stats);')
+		g.writeln('\tsignal(SIGINT, vprint_profile_stats_on_signal);')
+		g.writeln('\tsignal(SIGTERM, vprint_profile_stats_on_signal);')
 		g.writeln('\tatexit(vprint_profile_stats);')
 		g.writeln('')
 	}
--- a/vlib/v/gen/c/profile.v
+++ b/vlib/v/gen/c/profile.v
@@ -48,7 +48,7 @@ fn (mut g Gen) profile_fn(fn_decl ast.FnDecl) {
 }
 
 pub fn (mut g Gen) gen_vprint_profile_stats() {
-	g.pcs_declarations.writeln('void vprint_profile_stats(int sig){')
+	g.pcs_declarations.writeln('void vprint_profile_stats(){')
 	fstring := '"%14llu %14.3fms %14.0fns %s \\n"'
 	if g.pref.profile_file == '-' {
 		for pc_meta in g.pcs {
@@ -62,7 +62,6 @@ pub fn (mut g Gen) gen_vprint_profile_stats() {
 		}
 		g.pcs_declarations.writeln('\tfclose(fp);')
 	}
-	g.pcs_declarations.writeln('if (sig == SIGINT || sig == SIGTERM) { exit(0); }')
 	g.pcs_declarations.writeln('}')
 	g.pcs_declarations.writeln('')
 	g.pcs_declarations.writeln('void vreset_profile_stats(){')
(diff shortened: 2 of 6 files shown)


vlib/v/gen/c/profile.v
```diff
@@ -62,6 +62,7 @@ pub fn (mut g Gen) gen_vprint_profile_stats() {
 		}
 		g.pcs_declarations.writeln('\tfclose(fp);')
 	}
+	g.pcs_declarations.writeln('if (sig == SIGINT || sig == SIGTERM) { exit(0); }')
```
spytheman, line 65:
```suggestion
	g.pcs_declarations.writeln('if (sig == SIGINT || sig == SIGTERM) { exit(130); }')
```

On Ctrl-C, of an existing command like `sleep 99`, the observable exit code is not 0.

vlib/v/gen/c/cmain.v
```diff
@@ -217,6 +217,8 @@ pub fn (mut g Gen) gen_failing_return_error_for_test_fn(return_stmt ast.Return,
 pub fn (mut g Gen) gen_c_main_profile_hook() {
 	if g.pref.is_prof {
 		g.writeln('')
+		g.writeln('\tsignal(SIGINT, vprint_profile_stats);')
+		g.writeln('\tsignal(SIGTERM, vprint_profile_stats);')
 		g.writeln('\tatexit(vprint_profile_stats);')
```
spytheman, line 222: what will be the `sig` parameter, when atexit calls that function?

spytheman: For me:
```c
int atexit(void (*function)(void));
```
> The  atexit()  function  registers the given function to be called at normal process termination, either via exit(3) or via return from the program's main().  Functions so registered are called in the reverse order of their registration; no arguments are passed.

spytheman: We may need 2 separate functions 🤔 , one for the atexit callback, that takes no parameters, and one for the signal handler, accepting 1 parameter.

spytheman: The one that takes no parameters, can just call the other with -1.

pierrec: Ah yes good point.

spytheman: Sorry, I just tested, and `vprint_profile_stats` gets called 2 times now (the direct call, and then indirectly the exit(130), because of it being used as a callback in atexit). Commenting out the direct call works fine.

I tested with: `./v -profile x.txt t.v`, then `rm -rf x.txt; ./t`, waiting 3 seconds, and pressing Ctrl-C, where t.v is:
```v
import time
for i in 1..99 {
        time.sleep(1*time.second)
        println(i)
}
```

The following patch, made it so the profile data is still there, but its writing is done just once:
```v
diff --git vlib/v/gen/c/profile.v vlib/v/gen/c/profile.v
index fb84476ac..ee7fb7d15 100644
--- vlib/v/gen/c/profile.v
+++ vlib/v/gen/c/profile.v
@@ -49,6 +49,7 @@ fn (mut g Gen) profile_fn(fn_decl ast.FnDecl) {

 pub fn (mut g Gen) gen_vprint_profile_stats() {
        g.pcs_declarations.writeln('void vprint_profile_stats(){')
+       g.pcs_declarations.writeln('\tfprintf(stderr, "vprint_profile_stats called\\n");')
        fstring := '"%14llu %14.3fms %14.0fns %s \\n"'
        if g.pref.profile_file == '-' {
                for pc_meta in g.pcs {
@@ -73,7 +74,8 @@ pub fn (mut g Gen) gen_vprint_profile_stats() {
        g.pcs_declarations.writeln('')

        g.pcs_declarations.writeln('void vprint_profile_stats_on_signal(int sig){')
-       g.pcs_declarations.writeln('\tvprint_profile_stats();')
+       g.pcs_declarations.writeln('\tfprintf(stderr, "vprint_profile_stats_on_signal sig: %d\\n", sig);')
+//     g.pcs_declarations.writeln('\tvprint_profile_stats();')
        g.pcs_declarations.writeln('\texit(130);')
        g.pcs_declarations.writeln('}')
        g.pcs_declarations.writeln('')
```

spytheman:
```
#0 12:13:50 ᛋ issue8264 /v/oo❱./v -profile x.txt t.v
#0 12:13:56 ᛋ issue8264 /v/oo❱rm -rf x.txt; ./t
1
2
3
^Cvprint_profile_stats_on_signal sig: 2
vprint_profile_stats called
#130 12:14:10 ᛋ issue8264 /v/oo❱wc x.txt
  18   72 1128 x.txt
#0 12:14:14 ᛋ issue8264 /v/oo❱
```

spytheman: This also works:
```
#0 12:15:47 ᛋ issue8264 /v/oo❱rm -rf x.txt; timeout 3.2 ./t
1
2
3
vprint_profile_stats_on_signal sig: 15
vprint_profile_stats called
#124 12:15:54 ᛋ issue8264 /v/oo❱wc x.txt
  18   72 1128 x.txt
#0 12:15:59 ᛋ issue8264 /v/oo❱
```

spytheman: Not sure why the exit code in the second case is 124 🤔 .

spytheman: Oh, I see ... 124 is the code from the `timeout` program.

PR #20683, opened 2024-01-28
bitfield: enhance multiple flags operation

kbkpbot:
`set_if()` sets bit number 'bit_nr' to 1 (count from 0) if `cond` is true or clear the bit.
`toggle_bits()` changes the value (from 0 to 1 or from 1 to 0) of bits
`set_bits()` sets multiple bits in the array to 1.
`clear_bits()` clear multiple bits in the array to 0.
`has()` test if *at least one* of the bits is set
`all()` test if *all* of the bits are set


spytheman (approved):
Excellent work.

PR #20685, opened 2024-01-28
v.gen.c: allow returning option/result in or blocks

### LINKED ISSUE #20682: cgen error when returning an Option/Result from an `or {}` block
**V doctor:**
```
V full version: V 0.4.4 5306469.7e5ffb7

vexe mtime: 2024-01-28 08:36:51

Git vroot status: weekly.2024.03-91-g7e5ffb71
.git/config present: true

```

**What did you do?**
```v
fn find_startswith_string(a []string, search string) ?string {
	for s in a {
		if s.starts_with(search) {
			return s
		}
	}
	return none
}

fn find_any_startswith_string(a []string, b []string, search string) ?string {
	// cannot convert 'struct _option_string' to 'struct string'
	// V wants the or {} block to return a string, but find_startswith_string returns ?string
 	return find_startswith_string(a, search) or { find_startswith_string(b, search) }
}

fn main() {
	deadbeef := find_any_startswith_string(['foobar', 'barfoo'], ['deadbeef', 'beefdead'],
		'dead')
	assert deadbeef or { panic('unreachable') } == 'deadbeef'
}

```

Successful compilation, assert succeeds

**What did you see instead?**
```
==================
C:/Users/user/AppData/Local/Temp/v_0/test.01HN7KE7PZSE8B70ZE078KJQ15.tmp.c:6861: warning: implicit declaration of function 'tcc_backtrace'
C:/Users/user/AppData/Local/Temp/v_0/test.01HN7KE7PZSE8B70ZE078KJQ15.tmp.c:13018: error: cannot convert 'struct _option_string' to 'struct string'
...
==================

builder error: 
==================

```

Relevant generated C code:

```c
VV_LOCAL_SYMBOL _option_string main__find_any_startswith_string(Array_string a, Array_string b, string search) {
	_option_string _t2 = main__find_startswith_string(a, search);
	if (_t2.state != 0) {
		IError err = _t2.err;
		*(string*) _t2.data = main__find_startswith_string(b, search);
	}
	
 	_option_string _t1;
	_option_ok(&(string[]) {  (*(string*)_t2.data) }, (_option*)(&_t1), sizeof(string));
	return _t1;
}
```

image

> [!NOTE]
>

pierrec:
Fixes #20682.

### DIFF
--- a/vlib/v/gen/c/testdata/return_optional_or_block.vv
+++ b/vlib/v/gen/c/testdata/return_optional_or_block.vv
@@ -1,0 +1,20 @@
+fn find_startswith_string(a []string, search string) ?string {
+	for s in a {
+		if s.starts_with(search) {
+			return s
+		}
+	}
+	return none
+}
+
+fn find_any_startswith_string(a []string, b []string, search string) ?string {
+	// cannot convert 'struct _option_string' to 'struct string'
+	// V wants the or {} block to return a string, but find_startswith_string returns ?string
+ 	return find_startswith_string(a, search) or { find_startswith_string(b, search) }
+}
+
+fn main() {
+	deadbeef := find_any_startswith_string(['foobar', 'barfoo'], ['deadbeef', 'beefdead'],
+		'dead')
+	println(deadbeef or { panic('unreachable') })
+}
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -6614,7 +6614,11 @@ fn (mut g Gen) gen_or_block_stmts(cvar_name string, cast_typ string, stmts []ast
 						is_array_fixed = expr_stmt.expr is ast.ArrayInit
 							&& g.table.final_sym(return_type).kind == .array_fixed
 						if !is_array_fixed {
-							g.write('*(${cast_typ}*) ${cvar_name}.data = ')
+							if expr_stmt.typ.has_option_or_result() {
+								g.write('${cvar_name} = ')
+							} else {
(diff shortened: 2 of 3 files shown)


vlib/v/gen/c/testdata/return_optional_or_block.vv
```diff
@@ -0,0 +1,20 @@
...
+	return none
+}
+
+fn find_any_startswith_string(a []string, b []string, search string) ?string {
+	// cannot convert 'struct _option_string' to 'struct string'
+	// V wants the or {} block to return a string, but find_startswith_string returns ?string
+ 	return find_startswith_string(a, search) or { find_startswith_string(b, search) }
+}
+
+fn main() {
```
spytheman, line 16:
```suggestion
fn test_main() {
```
and rename the file to `vlib/v/tests/return_optional_or_block_test.v` .

The pair of .vv/.out files is slower to test (they are run by `coutput_test.v`, which in turn tests them sequentially, while `v test .` runs the _test.v files in a thread pool.
`coutput_test.v` also relies on reading the generated output, and comparing it to a prerecorded known good value from the .out file, while the _test.v files are expected to assert themselves, and just exit with != 0, when there are failures.

For a program, that would have produced a cgen error on master, but will compile and run fine in the PR that fixes it, a _test.v is cheaper.

pierrec: OK thanks for the explanation. The CI found an issue that I am still looking at and I will update the new test.

PR #20700, opened 2024-02-01
runtime: add cpuinfo support

kbkpbot:
Retrieve CPU information from sys regs /cpuid.
So, in your code, you can get CPU info like /proc/cpuinfo.
```v
		mut cpu := runtime.detect_x86()
		assert cpu.amd64_level() > 1
		assert cpu.amd64_level() < 5
		assert cpu.feature_set().len > 1
		assert cpu.feature_set().len < 10000
		assert cpu.all(.fpu, .mmx, .sse)
		assert cpu.has(.fpu, .mmx, .sse, .sse3)
```


JalonSolov:
Hmm... you are correct that `asm arm64` isn't currently supported.  `asm amd64` and `asm i386` are.

Terrible oversight that `arm64` isn't done yet.  :-(

From `vlib/v/gen/native/arm64.v`:
```
335:fn (mut g Gen) gen_asm_stmt_arm64(asm_node ast.AsmStmt) {
336:    g.v_error('The asm statement for arm64 not yet implemented', asm_node.pos)
```

spytheman:
The code here is very platform specific, and does not present a unified interface to the rest of the system. Imho keep it out of vlib for now, until it evolves a common interface.

PR #20701, opened 2024-02-01
Use standard default RM make variable to fix MSYS2 env

nsauzede:
Trivial patch to make V build fine within an MSYS2 shell


spytheman (approved):
Thank you.

PR #20706, opened 2024-02-02
v.debug: fix code generated for stringfy ptr with non-pointer str() 

### LINKED ISSUE #20702: `$dbg` statement in `sdl/examples/basic_window` results in compile error
### Describe the bug

`$dbg` statement in `sdl/examples/basic_window` results in compile error

### Reproduction Steps

`basic_window.v` with the following contents:
```v
module main

import sdl

fn main() {
	sdl.init(sdl.init_video)
	window := sdl.create_window('Hello SDL2'.str, 300, 300, 500, 300, 0)
	renderer := sdl.create_renderer(window, -1, u32(sdl.RendererFlags.accelerated) | u32(sdl.RendererFlags.presentvsync))

	mut should_close := false
	for {
		evt := sdl.Event{}
		for 0 < sdl.poll_event(&evt) {
			match evt.@type {
				.quit { should_close = true }
				else {}
			}
		}
		if should_close {
			break
		}
    $dbg // <- adding this yields a compile error

		sdl.set_render_draw_color(renderer, 255, 55, 55, 255)
		sdl.render_clear(renderer)
		sdl.render_present(renderer)
	}

	sdl.destroy_renderer(renderer)
	sdl.destroy_window(window)
	sdl.quit()
}
```

### Expected Behavior

Clean compile + run - and the debugger invoked

### Current Behavior

```
  v run examples/basic_window/
==================
/dev/shm/v_1000/basic_window.01HNJT6T4C374YQ9WZ2C5DPYX7.tmp.c:23075: warning: assignment discards qualifiers from pointer target type
/dev/shm/v_1000/basic_window.01HNJT6T4C374YQ9WZ2C5DPYX7.tmp.c:31646: error: cannot convert 'struct SDL_Window **' to 'struct SDL_Window'
...
==================
```

### Possible Solution

???

_No response_

### V version

V 0.4.4 fb0efc0

Linux Arch based (Endeavor OS)
> [!NOTE]
>

felipensp:
Fix #20702

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -141,63 +141,62 @@ mut:
 	inside_dump_fn            bool
 	expected_fixed_arr        bool
 	inside_for_c_stmt         bool
-	// inside_comptime_for_field bool
-	inside_cast_in_heap     int // inside cast to interface type in heap (resolve recursive calls)
-	inside_cast             bool
-	inside_const            bool
-	inside_array_item       bool
-	inside_const_opt_or_res bool
-	inside_lambda           bool
-	inside_cinit            bool
-	inside_interface_deref  bool
-	last_tmp_call_var       []string
-	loop_depth              int
-	ternary_names           map[string]string
-	ternary_level_names     map[string][]string
-	arraymap_set_pos        int      // map or array set value position
-	stmt_path_pos           []int    // positions of each statement start, for inserting C statements before the current statement
-	skip_stmt_pos           bool     // for handling if expressions + autofree (since both prepend C statements)
-	left_is_opt             bool     // left hand side on assignment is an option
-	right_is_opt            bool     // right hand side on assignment is an option
-	assign_ct_type          ast.Type // left hand side resolved comptime type
-	indent                  int
-	empty_line              bool
-	assign_op               token.Kind // *=, =, etc (for array_set)
-	defer_stmts             []ast.DeferStmt
-	defer_ifdef             string
-	defer_profile_code      string
-	defer_vars              []string
-	str_types               []StrType       // types that need automatic str() generation
-	generated_str_fns       []StrType       // types that already have a str() function
-	str_fn_names            []string        // remove duplicate function names
-	threaded_fns            shared []string // for generating unique wrapper types and fns for `go xxx()`
-	waiter_fns              shared []string // functions that wait for `go xxx()` to finish
-	needed_equality_fns     []ast.Type
-	generated_eq_fns        []ast.Type
-	array_sort_fn           shared []string
-	array_contains_types    []ast.Type
-	array_index_types       []ast.Type
-	auto_fn_definitions     []string // auto generated functions definition list
-	sumtype_casting_fns     []SumtypeCastingFn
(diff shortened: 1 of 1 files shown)


felipensp:
Can you test using this branch @Larpon ?

spytheman:
I've tested with this branch. 

It now compiles with this PR, but when running, I get:
```
#0 13:26:31 ᛋ fix_deref_debug /v/oo❱./v -g -keepc -gc none run b.v
7fd016148d58 : at ???: RUNTIME ERROR: invalid memory access
/tmp/v_1000/b.tmp.c:6415: by SDL_SysWMEvent_str
/tmp/v_1000/b.tmp.c:6317: by indent_sdl__SysWMEvent_str
/tmp/v_1000/b.tmp.c:6821: by indent_SDL_Event_str
/tmp/v_1000/b.tmp.c:5982: by SDL_Event_str
/tmp/v_1000/b.tmp.c:5946: by indent_sdl__Event_str
/tmp/v_1000/b.tmp.c:5943: by sdl__Event_str
/tmp/v_1000/../../../../../../space/v/oo/b.v:25: by main__main
/tmp/v_1000/../../../../../../tmp/v_1000/b.tmp.c:17345: by main
#255 13:26:47 ᛋ fix_deref_debug /v/oo❱
#255 13:26:48 ᛋ fix_deref_debug /v/oo❱
```

spytheman:
`./v -cg -keepc -cc clang-12 b.v` also fails to compile, on both master, and here.

spytheman:
I think what is happening, is that since it tries to stringify all variables in scope, that includes things like the locals `window` and `renderer`, and the automatic str methods for those do not work.

felipensp:
> I think what is happening, is that since it tries to stringify all variables in scope, that includes things like the locals `window` and `renderer`, and the automatic str methods for those do not work.

Yes, probably some struct members are not proper initialized before the stringify.

spytheman:
Adding those manually in my `~/.vmodules/sdl/fixes.v`:
```v
module sdl

pub fn (r &Renderer) str() string { return 'a renderer' }
pub fn (e &Event) str() string { return 'an event' }
pub fn (w &Window) str() string { return 'a window' }
```
fixes compilation.

Note that the manual methods above are defined on `x &Type`, not for `x Type`.

All of the 3 types above are of the form:
```v
@[noinit; typedef]
pub struct C.SDL_Renderer {
}

pub type Renderer = C.SDL_Renderer
```
i.e. a V type alias to an incomplete C declaration, so the compiler can only know how to handle a pointer to it, not how to pass it by value.

spytheman:
I think we have to fix the automatic str() method generation for those, to also pass them by reference.

spytheman (approved):
Good work. It does fix compilation, even though it does not fix the whole problem.

Larpon:
Confirmed. The fix avoids the compile error and it now results in a compile error:
```
v run ~/.vmodules/sdl/examples/basic_window/
73224028c287 : at ???: RUNTIME ERROR: invalid memory access
/dev/shm/v_1000/basic_window.01HNPZ42YER0676RQTPR78QD3S.tmp.c:32178: by main
```

Issue #20708, opened 2024-02-02
CORS in x.vweb


MedLabs:
### Describe the feature

Enabling CORS in a REST API is very important, there is an example created by @spytheman in vweb/cors `app.add_header("Access-Control-Allow-Origin", "*")`
but in x.vweb it doesn't work, and I think it could be a better and faster way to configure CORS, instead of adding a preflight function on @['/:path...'; options]

Why not add a Context function that takes the CORS config and could be added to before_request if we want the same config to all endpoints, or to be added to some endpoints as an argument...

### Use Case

Usually in other frameworks, I import cors library and set the cors options in a variable then app.use(corsOptions)

### Proposed Solution

a Context function that takes Origin, Methods etc as arguments...
`ctx.set_cors("*", "GET, POST, PUT, DELETE")`

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

x.vweb 0.4.4

### Environment details (OS name and version, etc.)

V full version: V 0.4.4 8e47c21.804a7bd
OS: linux, "ArcoLinux"
Processor: 16 cpus, 64bit, little endian, AMD Ryzen 7 7840HS with Radeon 780M Graphics
Git version: git version 2.43.0
Git vroot status: weekly.2024.05-2-g804a7bdd (10 commit(s) behind V master)
.git/config present: true

CC version: cc (GCC) 13.2.1 20230801
thirdparty/tcc status: thirdparty-linux-amd64 12f392c3
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

metif12:
hi.
i looked iside x.vweb and there is a complete and powerfull cors middleware for handling cors that you can use.

Example:
```v
app.use(vweb.cors[Context](vweb.CorsOptions{
    origin: '*'
    allowed_methods: [.get, .head, .patch, .put, .post, .delete]
}))
 ```

Casper64:
It was added for this issue, I'll close it

PR #20713, opened 2024-02-02
x.vweb: add cors middleware

Casper64:
fix for #20708 

This pr adds a middleware function in x.vweb to handle CORS requests.
It also includes a small test and a modified example of `examples/vweb/cors`

CORS request can be traced by adding `-d vweb_trace_cors` when you compile

## Usage

You only need to add `vweb.cors` as middleware to your app and configure CORS

**Example:**
```v
import x.vweb

pub struct Context {
	vweb.Context
}

pub struct App {
	vweb.Middleware[Context]
}
fn main() {
	mut app := &App{}
	// use the cors middleware
	app.use(vweb.cors[Context](vweb.CorsOptions{
		origin: '*'
		allowed_methods: [.get, .head, .put, .post, .patch, .delete]
	}))

	vweb.run[App, Context](mut app, 8080)
}
```

### DIFF
--- a/examples/xvweb/cors/vweb_cors_example.v
+++ b/examples/xvweb/cors/vweb_cors_example.v
@@ -1,0 +1,65 @@
+import time
+import x.vweb
+
+// See https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS
+// and https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#preflighted_requests
+// >  Cross-Origin Resource Sharing (CORS) is an HTTP-header based mechanism that allows
+// >  a server to indicate any origins (domain, scheme, or port) other than its own from
+// >  which a browser should permit loading resources...
+
+// Usage: do `./v run examples/xvweb/cors/` to start the app,
+// then check the headers in another shell:
+//
+// 1) `curl -vvv -X OPTIONS http://localhost:45678/time`
+// 2) `curl -vvv -X POST http://localhost:45678/time`
+
+pub struct Context {
+	vweb.Context
+}
+
+pub struct App {
+	vweb.Middleware[Context]
+}
+
+// time is a simple POST request handler, that returns the current time. It should be available
+// to JS scripts, running on arbitrary other origins/domains.
+@[post]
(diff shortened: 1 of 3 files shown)


examples/xvweb/cors/vweb_cors_example.v
```diff
@@ -0,0 +1,65 @@
...
+pub fn (app &App) time(mut ctx Context) vweb.Result {
+	return ctx.json({
+		'time': time.now().format_ss_milli()
+	})
+}
+
+// no_cors is a simple POST request handler, that returns a string.
+// For this example we don't want to expose the private route to any other origins/domains.
+@[post]
+pub fn (app &App) no_cors(mut ctx Context) vweb.Result {
```
spytheman, line 36: To me, it is not clear, how is `no_cors` different than the `time` method above.

Casper64: Left it in by accident, I'll remove it

enghitalo:
How will multiple origins cors works?

Casper64:
> How will multiple origin cors works?

Good suggestion, I can make the origin into an array. 

From the MDN docs:
> Only a single origin can be specified. If the server supports clients from multiple origins, it must return the origin for the specific client making the request.

xqliang:
Yes should allow a list of hosts, and return the requested origin(even when the allowed host is *).

vlib/x/vweb/middleware.v
```diff
@@ -198,7 +198,15 @@ pub:
 
 // set_headers adds the CORS headers on the response
 pub fn (options &CorsOptions) set_headers(mut ctx Context) {
-	ctx.set_header(.access_control_allow_origin, options.origin)
+	// A browser will reject a CORS request when the Acess-Control-Allow-Origin header
```
JalonSolov, line 205:
```suggestion
	// A browser will reject a CORS request when the Access-Control-Allow-Origin header
```

spytheman:
@Casper64 , please run `git pull --rebase origin master` .

spytheman:
Sorry, it was easier for me to do it.

PR #20716, opened 2024-02-02
dtm: First version of user documentation

GGRei:
First version of DTM documentation.

Feel free to suggest modifications or corrections if necessary. Thank you in advance!

### DIFF
--- a/vlib/x/templating/dtm/README.md
+++ b/vlib/x/templating/dtm/README.md
@@ -1,6 +1,6 @@
 # dtm - Dynamic Template Manager
 
-A simple template manager integrated into the V project, designed to combine the power of templates with vweb, without the need to recompile the application with every change.
+A simple template manager integrated into the V project, designed to combine the power of V templates with Vweb, without the need to recompile the application with every change.
 
 **_This module is in the experimental phase, it lacks some features and may contain bugs_**
 
@@ -13,13 +13,17 @@ Before starting, at the root directory of your vweb project, you need to create 
 
 Next, you must add your HTML file into the folder you previously created. ***Be aware that if the HTML templates are not placed in the correct directory, the DTM will return an error message indicating that it cannot find the template!!***
 
-Now a minimal quick start example:
+A minimal quick start example:
 
 ```
+import vweb
+import x.templating.dtm
+
 struct App {
 	vweb.Context
 pub mut:
-    // The vweb_global attribute must be applied in order to use the dtm (Dynamic Template Manager) throughout your application.
+    // The vweb_global attribute must be applied in order to use the dtm 
+    // (Dynamic Template Manager) throughout your application.
     // However, you do not need the attribute if you are using x.vweb
 	dtm &dtm.DynamicTemplateManager = dtm.create_dtm() @[vweb_global] 
 }
@@ -54,7 +58,7 @@ There are two types of option possibilities:
 Three parameters are available:
 
 - `max_size_data_in_mem` : ( **Int** value ) Maximum size of data allowed in memory for each cached template. The value must be specified in kilobytes. ( Default is: 500KB / Limit max is : 500KB )
-- `compress_html` : ( **Bool** value ) Light compress of the HTML ouput, to remove all unnecessary spacing.
+- `compress_html` : ( **Bool** value ) Light compress of the HTML ouput, to remove all unnecessary spacing. ( Default is true )
 - `active_cache_server` : ( **Bool** value ) Activate or not the template cache system. ( Default is true, ***_Highly recommended to keep it enabled for optimal performance_*** )
 
 Using like this :
@@ -70,8 +74,9 @@ initialize_dtm(mut app.dtm,
 
 ### Defined for each web page
 
-- `placeholders` ( **&map[string]DtmMultiTypeMap** value ) Used to map placeholders within the template to their corresponding content, facilitating dynamic content insertion. By specifying values in the placeholders map, templates can dynamically display content.
-- `cache_delay_expiration` ( **i64** value ) Specifies the cache expiration time for the concerned page in seconds. ( Minimum value is **300** seconds and maximum is **31536000** seconds, equivalent to five minutes and one year. Default value is **86400** seconds or one day ). 
+- `placeholders` ( **&map[string]DtmMultiTypeMap** value ) Used to map placeholders within the template to their corresponding content, facilitating dynamic content insertion, by specifying values in the placeholders map. Templates can dynamically display content.
+
+- `cache_delay_expiration` ( **i64** value ) Specifies the cache expiration time for the concerned page in seconds. ( Minimum value is **300** seconds and maximum is **31536000** seconds, equivalent to five minutes and one year. Default value is **86400** seconds or one day ). You can add any value you want in seconds as long as it remains within the previously indicated range. 
 
 Possibility to use already defined cache delay constants like: `cache_delay_expiration_at_min`, `cache_delay_expiration_at_max`, `cache_delay_expiration_by_default`
 
@@ -86,9 +91,9 @@ serve_dynamic_template('path/of/template.html',
 
 ## The PlaceHolders System
 
-### On the user's side code :
+### On The User's Side Code :
 
-The placeholder system allows for the insertion of dynamic content into your template. As of the current state of the module, it accepts the following primitive types:
+The placeholder system allows for the insertion of dynamic content into your template. As of the current state of the module, it accepts the following types like:
 
 ```
 - string
@@ -123,7 +128,7 @@ Pay attention to this particular tag: "**_#includehtml**", it enables you to inc
 '</summary>'
 ```
 
-### On the template side :
+### On The Template Side :
 
 An example of a template, corresponding to the previous subsection:
 
@@ -142,3 +147,21 @@ An example of a template, corresponding to the previous subsection:
   </body>
 </html>
 ```
+You will note that the `'_#includehtml'` directive is not found in the template with `'@placeholder_name_3'`, and this is entirely normal. Directives are specially handled by the DTM, and including them in the name of your placeholders within the template will result in the placeholder not being found because it does not match the key name defined in the map containing the dynamic content.
+
+
+Like the traditional template system in V, inclusions or placeholders start with the '**@**' character. The traditional inclusion system is still perfectly usable, such as:
+
+```
+- @include 'my/html/path.html'
+- @css 'my/css/path.css'
+- @js 'my/js/path.js'
+```
+
+## In The Future
+
+As you've understood, the DTM is still under development and optimization. There are functionalities to be added, such as data compression, managing loops or conditions within the template itself. Able to be used in contexts other than HTML, such as text files, and for modules other than vweb.
+
+This will come in time.
+
+Feel free to report any issues or contribute to the project!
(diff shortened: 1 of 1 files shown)


JalonSolov:
Since you separated it from vweb, perhaps some documentation on how to use it without vweb?

GGRei:
@JalonSolov Indeed, the DTM has now become its own module; however, for the time being, even though it has all the elements needed to utilize files other than .html, it does not accept them. This is because, originally, the module was designed for use with Vweb. Its functionality will certainly be expanded to be applicable in other types of projects, but as of now, I'm unsure what uses to suggest outside of Vweb.


Keeping this in mind, if you have any ideas to suggest, please don't hesitate to give them to me so that I can incorporate them into the documentation. Thank you for your help. :)

spytheman:
A simple static site generator can be a good example, i.e. iterate over a folder, containing template files, and produce another folder, containing their expansion.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
+# dtm - Dynamic Template Manager
+
+A simple template manager integrated into the V project, designed to combine the power of V
+templates with Vweb, without the need to recompile the application with every change.
+
+**_This module is in the experimental phase, it lacks some features and may contain bugs_**
```
spytheman, line 6: There is no need for that sentence, all software is.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Quick Start
+
+Using the dynamic template manager ( named '**dtm**' in this readme) is relatively straightforward.
+You just need to create an instance and initialize it. Then, call the '**serve_dynamic_template**'
```
spytheman, line 11: Is there a use case, where you may want to create an instance, and use it without initializing it?
If no, merging the initialization into the creation can simplify uses.

spytheman: Would not just `expand` be clearer as a name? The instance on which you call it, is already called `dtm`, there is little point in repeating the `dynamic_template` in the method name too, and `serve` is too generic imho.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+function in the code that manages your web pages. 
+
+Before starting, at the root directory of your vweb project, you need to create a '**templates**'
+folder as well as a '**vcache**' folder. The absence of these directories at the root of your
```
spytheman, line 15: The absence of a `vcache` folder should not prevent `dtm` from working. If it is not there, it can just not cache anything at all, which in many cases will be preferable anyway.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+```
+import vweb
+import x.templating.dtm
+
+struct App {
+	vweb.Context
+pub mut:
+    // The vweb_global attribute must be applied in order to use the dtm 
+    // (Dynamic Template Manager) throughout your application.
+    // However, you do not need the attribute if you are using x.vweb
```
spytheman, line 33: Just use `x.vweb` as an example directly. Commenting here on `@[vweb_global]` is just making the example less clear, and the purpose of that example is to showcase `dtm`, not to troubleshoot and point out quirks in `vweb`.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+	dtm &dtm.DynamicTemplateManager = dtm.create_dtm() @[vweb_global] 
+}
+
+fn main() {
+	mut app := &App{}
+
```
spytheman, line 39: No need for these empty lines, they just make users scroll more.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+folder as well as a '**vcache**' folder. The absence of these directories at the root of your
+project will prevent you from using the dtm.
+
+Next, you must add your HTML file into the folder you previously created. ***Be aware that if the
+HTML templates are not placed in the correct directory, the DTM will return an error message
+indicating that it cannot find the template!!***
+
+A minimal quick start example:
+
+```
```
spytheman, line 24: End these 3 backslashes with `v`. This will make sure that `v check-md .` will try to compile the example, and thus prevent silent regressions, when something in dtm or vweb changes.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Defined for each web page
+
+- `placeholders` ( **&map[string]DtmMultiTypeMap** value ) Used to map placeholders within the
+  template to their corresponding content, facilitating dynamic content insertion, by specifying
+  values in the placeholders map. Templates can dynamically display content.
+
+- `cache_delay_expiration` ( **i64** value ) Specifies the cache expiration time for the concerned
+  page in seconds. ( Minimum value is **300** seconds and maximum is **31536000** seconds,
+  equivalent to five minutes and one year. Default value is **86400** seconds or one day ). You can
```
spytheman, line 93:
```suggestion
  equivalent to one year and five minutes. Default value is **86400** seconds or one day ). You can
```

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+  add any value you want in seconds as long as it remains within the previously indicated range. 
+
+Possibility to use already defined cache delay constants like: `cache_delay_expiration_at_min`,
+`cache_delay_expiration_at_max`, `cache_delay_expiration_by_default`
+
+Using like this :
```
spytheman, line 99:
```suggestion
Use it like this:
```

vlib/x/templating/dtm/README.md
spytheman: It will also make github syntax highlight the example code.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+- `max_size_data_in_mem` : ( **Int** value ) Maximum size of data allowed in memory for each cached
+  template. The value must be specified in kilobytes. ( Default is: 500KB / Limit max is : 500KB )
+- `compress_html` : ( **Bool** value ) Light compress of the HTML ouput, to remove all unnecessary
+  spacing. ( Default is true )
+- `active_cache_server` : ( **Bool** value ) Activate or not the template cache system. ( Default is
+  true, ***_Highly recommended to keep it enabled for optimal performance_*** )
+
+Using like this :
+
+```
```
spytheman, line 76: use this here:
```
```v ignore
```

spytheman: It will still get highlighted, but since it is a snippet, it will not get compiled, and `v check-md .` will not complain about it.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+  page in seconds. ( Minimum value is **300** seconds and maximum is **31536000** seconds,
+  equivalent to five minutes and one year. Default value is **86400** seconds or one day ). You can
+  add any value you want in seconds as long as it remains within the previously indicated range. 
+
+Possibility to use already defined cache delay constants like: `cache_delay_expiration_at_min`,
+`cache_delay_expiration_at_max`, `cache_delay_expiration_by_default`
+
+Using like this :
+
+```
```
spytheman, line 101:
```
```v ignore
```

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+- string
+- i8, i16, int, i64
+- u8, u16, u32, u64
+- f32, f64
+```
+
+it is used as follows:
+
+```
```
spytheman, line 124:
```
```v ignore
```

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,193 @@
...
+'</footer>', '<main>', '</main>', '<section>', '</section>', '<article>', '</article>', '<aside>',
+'</aside>', '<details>', '</details>', '<dialog>', '</dialog>', '<data>', '</data>', '<summary>',
+'</summary>'
+```
+
+### On The Template Side :
+
+An example of a template, corresponding to the previous subsection:
+
+```
```
spytheman, line 154:
```
```html
```

GGRei:
Thank you for your feedbacks. However, we're not just talking about modifying the documentation; some changes need to be made within the DTM itself. Some things will be quick to adjust, such as renaming functions or combining instance creation and initialization into a single function.

On the other hand, omitting the "vcache" directory touches on the core functioning of the DTM since it uses this space to work, even temporarily. This leads to two potential issues:

If there's no "vcache" directory, then the files, if any, will be created at the root of the project (which I personally find less than ideal).

If we don't want any cache files and, therefore, no disk access, meaning the data can be stored in memory if it doesn't exceed the allowed size, or users do not disable the cache system. This implies a significant overhaul of portion of the DTM, and it's no longer the same task.

I will adapt to your requests, but it's essential to be fully aware of these considerations before deciding on a direction.

JalonSolov:
If there's no `vcache` subdir, use https://modules.vlang.io/os.html#temp_dir to find the system temp dir, and create a subdir under that, and clean it up on exit (`defer` can help with that).  Most would likely prefer this, with a `vcache` possibility as something for those who need it.

GGRei:
That's a good idea; I hadn't thought of that. Thank you!

@spytheman its' ok for you?

spytheman:
I do not care much about performance in this case (with the cache turned off). I do want *predictability* and *simplicity*, i.e. just do the expansion, without starting threads, using channels, caching in memory etc. Just the pure raw action of the system, that given an input, it produces an output. Such a mode of operation is extremely useful for debugging, while developing.

spytheman:
As for priority, I think that fixing the rest of the docs here, should be done first. The other suggestions/ideas that need code modifications can follow later. I just thought about them, while reading the condensed example, and implementing them here is out of scope of this PR.

GGRei:
> Just the pure raw action of the system, that given an input, it produces an output. Such a mode of operation is extremely useful for debugging, while developing.

I assure you, this is already the case. Choosing to disable the cache results in exactly the behavior you've described, without the need for threads, channels, or other mechanisms. The only downside is potentially reduced performance under heavy system load. However, as you correctly noted, this is irrelevant in a debugging context.

It should be noted, and I admit this was an oversight in the documentation, that even with the caching system activated, it's possible for a specific page by setting the '**cache_delay_expiration**' parameter to '**-1**' to ignore the cache. this will have the effect ensures that only the pure raw action is executed for a page specified with this parameter.

 I will add this clarification to the documentation.

spytheman:
> it's possible for a specific page by setting the 'cache_delay_expiration' parameter to '-1' to ignore the cache

Excellent. That should be documented indeed; it will help pinpointing/troubleshooting problems.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,278 @@
...
+Title of text : @title
+Value in the text : @non_string_type
+HTML tags is always escaped in text file : @html_section
+```
+
+### 2. Minimal Vweb example:
+
+```v ignore
```
spytheman, line 65: Why `v ignore`?

A full example should compile, and it should be as presentable as possible, including formatting etc. It should be ready to be copy/pasted.

Using just v, ensures that `v check-md .` will check that all such examples compile, and are v fmt-ed.

vlib/x/templating/dtm/README.md
```diff
@@ -0,0 +1,278 @@
...
+The DTM currently only works with 2 types of files:
+
+- html
+- raw text
+
+Below you will find 2 usage contexts:
+
+### 1. Minimal static generator example :
+
+```v ignore
```
spytheman, line 33: Use just v here, not `v ignore` . This is a complete example, that should compile, and that should be v fmt-ed.

GGRei:
Ok, sorry, I'm not very used to working on README.md files for the project. I hope it will be better. Thanks for the information, anyway!

PR #20721, opened 2024-02-03
cgen: fix interface on multi return func

### LINKED ISSUE #20720: Compilation fails with Interface option
### Describe the bug

Compilation fails with error: cannot convert 'struct main__ITest' to 'unsigned char' when using an option type return with an interface

### Reproduction Steps

```
interface ITest {
	test()
}

struct Test {}

fn (p Test) test() {
	println('p.test()')
}

fn get_test(id int) (?ITest, string) {
	match id {
		0 { return Test{}, 'Value 0' }
		else { return none, 'Value not 0' }
	}
}

t, _ := get_test(0)

if tt := t {
	tt.test()
}
```

### Expected Behavior

Compilation should work

### Current Behavior

Compilation fails

### Possible Solution

_No response_

_No response_

### V version

V 0.4.4 1dc5d36

V full version: V 0.4.4 8f76a06.1dc5d36
OS: linux, Ubuntu 22.04.3 LTS
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-7100 CPU @ 3.90GHz
> [!NOTE]
>

felipensp:
Fix #20720

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5273,8 +5273,7 @@ fn (mut g Gen) return_stmt(node ast.Return) {
 				call_expr := expr as ast.CallExpr
 				expr_sym := g.table.sym(call_expr.return_type)
 				mut tmp := g.new_tmp_var()
-				if !call_expr.return_type.has_flag(.option)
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #20722, opened 2024-02-03
test: add test for cast as selector fix

felipensp:
Missing test to push.

### DIFF
--- a/vlib/v/tests/as_cast_selector_test.v
+++ b/vlib/v/tests/as_cast_selector_test.v
@@ -1,0 +1,18 @@
+struct Foo {
+mut:
+	x int
+}
+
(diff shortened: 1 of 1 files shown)


vlib/v/tests/as_cast_selector_test.v
```diff
@@ -0,0 +1,18 @@
...
+	y int
+}
+
+type Foobar = Bar | Foo
+
+fn test_main() {
+	mut bar := Foobar(Bar{
+		y: 123
+	})
+	assert bar as Bar.y == 123
```
spytheman, line 17: That will need a change, when the checker error is added, but it is good for now.
Excellent work.

PR #20723, opened 2024-02-03
checker: fix comptime ident checking on defer stmt

### LINKED ISSUE #20719: Comptime conditional in defer is parser error
### Describe the bug

If you attempt to use a comptime conditional `$if` statement within a `defer` block, the parser gives an error.

### Reproduction Steps

```v
defer {
  $if foo ? { // error: undefined ident: `foo`
    // whatever
  }
}
```

### Expected Behavior

This should be allowed by the parser

### Current Behavior

Parser gives error
```
tim.v:18:7: error: undefined ident: `foo`
   16 | fn main() {
   17 |     defer {
   18 |         $if foo ? {
      |             ~~~
   19 |             println('foo')
   20 |         }
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.4 10aaeeb

Mac
> [!NOTE]
>

felipensp:
Fix #20719

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -2036,7 +2036,8 @@ fn (mut c Checker) stmt(mut node ast.Stmt) {
 			for i, ident in node.defer_vars {
 				mut id := ident
 				if mut id.info is ast.IdentVar {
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #20718, opened 2024-02-02
cgen: fix as cast as selector

felipensp:
Fix #20710

```v
struct Foo {
mut:
	x int
}

struct Bar {
mut:
	y int
}

type Foobar = Foo | Bar

fn main() {
	mut bar := Foobar(Bar{y: 123})
	assert bar as Bar.y == 123
}
```

PR #20725, opened 2024-02-03
checker: add check if `as` cast used incorrectly for `selector-expr`

Delta456:
Adds a checker error as proposed at https://github.com/vlang/v/pull/20718#issuecomment-1925272965

```
vlib/v/checker/tests/as_cast_selector_expr_err.vv:15:14: error: indeterminate `as` cast, use parenthesis to clarity
   13 | fn main() {
   14 |     mut bar := Foobar(Bar{y: 123})
   15 |     println(bar as Bar.y == 123)
      |                 ~~
   16 | }
Details: for example `(bar as main.Bar).y`
```


spytheman (approved):
Excellent work.

PR #20731, opened 2024-02-04
cgen: Fix premature variable release by autofree

### LINKED ISSUE #20690: autofree - variable undeclared - C builder error
### Describe the bug

Compiling the example code with autofree option results in a compiler error

### Reproduction Steps

```v
module main

fn f(s string) ?[]u8 {
    if s == '' {
        return none
    }
    return [u8(104), 105]
}

fn main() {
    text := if temp := f('') {
        temp.str()
    } else if temp := f('something') {
        temp.str()
    } else {
        return
    }
    println(text)
}
```

```
v -autofree .
```

### Expected Behavior

No error

### Current Behavior

```
❯ v -cg -autofree .
/tmp/v_1000/autofree_error.01HNBMATFZWHF0JMW8F2KNQ4SC.tmp.c:12816: error: 'text' undeclared
builder error:
==================

```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.4 804a7bd

Arch Linux x86
> [!NOTE]
>

GGRei:
Fix #20690

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -2970,7 +2970,9 @@ fn (mut g Gen) autofree_scope_vars2(scope &ast.Scope, start_pos int, end_pos int
 				// continue
 				// }
 				// if v.pos.pos > end_pos {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #20735, opened 2024-02-05
dtm: redesign certain parts

GGRei:
To adjust certain behaviors recommended in this PR: #20716 , the DTM has undergone some minor modifications:

- The creation and initialization of a DTM instance have been merged into the same function
- Some function names have been revised.
- Now, for the cache folder : User can define cache folder path. If not exist or problem with it, an attempt will be made to create the directory in the temporary storage area of the OS. Otherwise, cache system will be desactived and the problem is reported if the user had activated the cache service..
- Internal and external test files have been updated accordingly.
- Raw text files are now managed by the dtm

@spytheman  I will update the documentation in due course. However, this PR needs to be reviewed and, if accepted, merged as a priority; otherwise, the documentation will not match. Thank you!

### DIFF
--- a/vlib/x/templating/dtm/dynamic_template_manager.v
+++ b/vlib/x/templating/dtm/dynamic_template_manager.v
@@ -53,6 +53,11 @@ enum CacheRequest {
 	delete
 }
 
+enum TemplateType {
+	html
+	text
+}
+
 // DynamicTemplateManager
 @[heap]
 pub struct DynamicTemplateManager {
@@ -82,7 +87,7 @@ mut:
 	//	Dtm clock
 	c_time            i64
 	ch_stop_dtm_clock chan bool = chan bool{cap: 5}
-	// Store small informations about already cached pages to improve the verification speed of the check_html_and_placeholders_size function.
+	// Store small informations about already cached pages to improve the verification speed of the check_tmpl_and_placeholders_size function.
 	html_file_info shared map[string]HtmlFileInfo = map[string]HtmlFileInfo{}
 	// Indicates whether the cache file storage directory is located in a temporary OS area
 	cache_folder_is_temporary_storage bool
@@ -145,6 +150,7 @@ mut:
 struct HtmlFileInfo {
 	file_full_path string
 	file_name      string
+	file_type      TemplateType
 }
 
 // TemplateCacheParams are used to specify cache expiration delay and provide placeholder data for substitution in templates.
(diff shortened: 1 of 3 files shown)


JalonSolov:
A suggestion... instead of

* Now, if the vcache_dtm folder is not present in the project's root directory, an attempt will be made to create the directory in the temporary storage area of the OS. Otherwise, cache system will be desactived.

perhaps it would be even more flexible as

* By default, an attempt will be made to create the directory in the temporary storage area of the OS.  If you want to choose a fixed location elsewhere, pass the "cache_dir" parameter with the path to use.

GGRei:
Interesting. So for you, it's better never have the vcache folder in the project root at all?

Thanks for you support!

JalonSolov:
Well, there might be a time you want it in the root project (at least one Linux distro prevents users from writing to /tmp, for example... they may even prevent writing to anywhere other than the user's home dir), which is why there should be a parameter to tell it where to put the cache dir.

GGRei:
Ok, so let me know if you agree with this based on your recommendations, and then I'll get to work:

1) User's choice (However, the DTM will not create the directory; it will be the user's responsibility). The DTM will just return whether it can utilize the directory or not. (Which is what it actually does already).

2) If choice 1 is not specified by the user or fails (for example, due to a permission issue), we attempt creation in the OS's temporary area.

3) If everything fails, we disable the cache.

Does this seem correct to you?

JalonSolov:
I would swap 1 & 2.

By default, most people will want a cache, but they won't want to make a special directory for it.  So if the code attempts to create a directory in the system temp area, that will give them what they want.

If that directory can't be created, for whatever reason (and hopefully the reason will be communicated to the user), the code will work without a cache.

If the user wants the cache somewhere else, be that in the project root, the home dir, etc., etc., then an option to set the cache path will be needed... and again, if that can't be created/used, the reason needs to be reported, and the code will work without a cache.

GGRei:
Alright, but if the user wants to control the cache directory for any reason, by doing as you suggest, priority will be given to the temporary area, and if creation in this area is successful, the user's parameter will be ignored.

Are you sure with this approach?

> Perhaps there could be a "-no-cache" option as well, to force running without a cache, without even trying to create/read a cache.

This is already the case, this option exists in two variations:

- General parameter: cache disabled
- Specific parameter per template: can disable the cache on the specific template even if the cache system is active.

GGRei:
It's done. Now, dtm manage raw text files in addition to HTML files and the changes from our previous discussion have been implemented.

vlib/x/templating/dtm/dynamic_template_manager.v
```diff
@@ -150,84 +160,116 @@ pub struct TemplateCacheParams {
...
+// initialize create and init the 'DynamicTemplateManager' with the storage mode, cache/templates path folders.
+// A cache directory can be created by the user for storage. If it is not defined or encounters issues such as permission problems,
+// the DTM will attempt to create it in the OS's temporary area. If this proves impossible, the cache system will be deactivated and the user will be informed if cache system was required.
 // Initalisation params are :
+// - def_cache_path 'type string' User can define the path of cache folder.
 // - max_size_data_in_mem 'type int' Maximum size of data allowed in memory for caching. The value must be specified in kilobytes. ( Default is: 500KB / Limit max is : 500KB)	
 // - compress_html: 'type bool' Light compress of the HTML ouput. ( default is true )
 // - active_cache_server: 'type bool' Activate or not the template cache system. ( default is true )
 // - test_cache_dir: 'type string' Used only for DTM internal test file, parameter is ignored otherwise.
 // - test_template_dir: 'type string' Used only for DTM internal test file, parameter is ignored otherwise.
```
spytheman, line 183: Move those comments, to the fields of the `@[params]` struct.
Documenting them separately leads to needless redundancy :-| .

spytheman (approved):
Good work, I guess 🤔 . It was very hard to review, given all the renames and moving of code all over the place.

In general, it is better to split pure refactoring PRs, from those that do functional changes. Smaller, and targeted PRs have a much better chance of not being delayed.

GGRei:
Ok, thank you, I will organize myself better in the future!

PR #20736, opened 2024-02-06
builtin: add string to u8 array support

kbkpbot:
u8_array returns the value of the hex/bin string as u8 array.
hex string example: `'0x11223344ee'.u8_array() == [u8(0x11),0x22,0x33,0x44,0xee]`.
bin string example: `'0b1101_1101'.u8_array() == [u8(0xdd)]`.
 underscore in the string will be stripped.

This will help convert a hex/bin format string into a u8 array.

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -621,6 +621,9 @@ pub fn (s string) u8() u8 {
 pub fn (s string) u8_array() []u8 {
 	// strip underscore in the string
 	mut tmps := s.replace('_', '')
+	defer {
+		unsafe { tmps.free() }
+	}
 	if tmps.len == 0 {
 		return []u8{}
 	}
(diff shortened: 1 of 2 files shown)


vlib/builtin/string.v
```diff
@@ -613,6 +613,63 @@ pub fn (s string) u8() u8 {
 	return u8(strconv.common_parse_uint(s, 0, 8, false, false) or { 0 })
 }
 
+// u8_array returns the value of the hex/bin string as u8 array.
+// hex string example: `'0x11223344ee'.u8_array() == [u8(0x11),0x22,0x33,0x44,0xee]`.
+// bin string example: `'0b1101_1101'.u8_array() == [u8(0xdd)]`.
+// underscore in the string will be stripped.
+@[inline]
```
spytheman, line 620: I am not sure that marking it `@[inline]` is beneficial. The code contains multiple branches and a loop. The overhead of those, is probably bigger than the savings from inlining.

PR #20739, opened 2024-02-06
os: add os.stat() and helpers

syrmel:
Adds a platform-agnostic Stat struct and os.stat() function to provide stat() functionality as it is found in other languages and to make it possible to simplify other functions within os like file_size() by consistently using Stat.

Helper functions for the Stat struct have been added to make working with stat info dev-friendly, specifically get_{a|c|m}time() to return a Time struct instead of just the unix epoch and get_filetype() and get_mode() which use existing os structs (FileType, FilePermissions, and FileMode).

Handling of stat-related info is quite different in Windows than in *nix; the required C code has been split up accordingly.


spytheman (approved):
Excellent work.

PR #20740, opened 2024-02-06
x.vweb: add serve_static example in README

Casper64:
Add a small example in the README on how to use the `serve_static` method on the `StaticHandler` struct.

At the start of the `Static files` doc section is an example file structure. That one is also used for this example


spytheman (approved):
Good work.

PR #20751, opened 2024-02-07
dtm: Optimization of parser/gen of template 

GGRei:
The template parsing and rendering system has been completely overhauled for significant performance improvements.

### DIFF
--- a/vlib/x/templating/dtm/dynamic_template_manager.v
+++ b/vlib/x/templating/dtm/dynamic_template_manager.v
@@ -1,7 +1,6 @@
 module dtm
 
 import os
-import v.parser
 import crypto.md5
 import hash.fnv1a
 import time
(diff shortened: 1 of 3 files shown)


vlib/x/templating/dtm/dynamic_template_manager.v
```diff
@@ -963,17 +963,11 @@ fn (mut tm DynamicTemplateManager) chandler_remaining_cache_template_used(cr Cac
 
 // fn (mut DynamicTemplateManager) parse_tmpl_file(string, string, &map[string]DtmMultiTypeMap, bool, TemplateType) return (string, string)
 //
-// The V compiler's template parser 'vlib/v/parser/tmpl.v', parses and transforms template file content.
+// Parses and generate template file content.
```
JalonSolov, line 966:
```suggestion
// Parses and generates template file content.
```

vlib/x/templating/dtm/dynamic_template_manager.v
```diff
@@ -224,11 +225,16 @@ pub fn initialize(dtm_init_params DynamicTemplateManagerInitialisationParams) &D
 			}
 			// If it is impossible to use a cache directory, the cache system is deactivated, and the user is warned."
 			if !active_cache_handler {
+				// If the cache server is not used, the HTML minifier is disabled, which increases the performance of pure generation without caching
+				minified_html = false
```
spytheman, line 229: No, revert that change. 

Such inter-dependencies between subsystems should be avoided.
Debugging a system, that decides on its own to turn on/off unrelated subsystems, based on the state of other subsystems is a nightmare. Each such decision doubles the amount of edge cases that you have to test and consider, for practically no benefit.

Performance concerns are better addressed by *explicit* and clear configuration.

GGRei:
Ok, I've reverted back by one commit. Do you think instead I should mention in the documentation that not using the minifier when the cache is disabled can improve performance, especially for smaller templates?

spytheman:
Yes, that is good information to document. Then the user can decide for themselves what is suitable.

GGRei:
In this PR? Or open a new?

spytheman (approved):
Excellent work.

PR #20759, opened 2024-02-08
os: refactor to use os.stat and os.lstat instead of unsafe C calls

syrmel:
Goal: simplify `os` codebase and reduce number of unsafe blocks.

Changes:
- Introduce os.lstat() to call C.lstat explicitly to be consistent with the behavior of C.stat() and C.lstat()
- Refactor the following functions to use os.stat() or os.lstat(): file_size, cp, is_executable, is_dir, is_link, kind_of_existing_path, file_last_mod_unix, posix_set_permission_bit, and inode
- Add comments to the Stat struct

When appropriate because of Win API instead of C lib calls, split functions between os_stat_default.c.v or os_stat_windows.c.v.

Regression tested with existing os tests.

### DIFF
--- a/vlib/os/os.c.v
+++ b/vlib/os/os.c.v
@@ -169,52 +169,15 @@ pub fn truncate(path string, len u64) ! {
 	}
 }
 
-fn eprintln_unknown_file_size() {
-	eprintln('os.file_size() Cannot determine file-size: ' + posix_get_error_msg(C.errno))
-}
-
 // file_size returns the size of the file located in `path`.
 // If an error occurs it returns 0.
 // Note that use of this on symbolic links on Windows returns always 0.
 pub fn file_size(path string) u64 {
-	mut s := C.stat{}
-	unsafe {
-		$if x64 {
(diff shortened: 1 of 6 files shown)


vlib/os/os.c.v
```diff
@@ -169,52 +169,15 @@ pub fn truncate(path string, len u64) ! {
...
 }
 
-fn eprintln_unknown_file_size() {
-	eprintln('os.file_size() Cannot determine file-size: ' + posix_get_error_msg(C.errno))
-}
-
 // file_size returns the size of the file located in `path`.
 // If an error occurs it returns 0.
 // Note that use of this on symbolic links on Windows returns always 0.
 pub fn file_size(path string) u64 {
```
spytheman, line 175: Can you check, if that works with files larger than 4GB on windows?

syrmel: Confirmed; works on 32- and 64-bit Windows

spytheman (approved):
Excellent cleanup.

Issue #20763, opened 2024-02-09
cgen: using incomplete `?bool` type


MCausc78:
### Describe the bug

<title>

### Reproduction Steps

```v
type Foo = ?bool

@[params]
struct Bar {
        a ?Foo
}

fn f(bar Bar) {
        dump(bar)
}

fn main() {
        f()
        f(a: none)
        f(a: true)
        f(a: false)
}
```

### Expected Behavior

Successful compile

### Current Behavior

```c
==================
/data/data/com.termux/files/usr/tmp/v_10272/issue090224.01HP5P3S3A8R2QHSD5HMD4S9DZ.tmp.c:1077:12: error: invalid application of 'sizeof' to an incomplete type 'main__Foo' (aka 'struct _option_bool')
 1077 |         byte data[sizeof(main__Foo) > 1 ? sizeof(main__Foo) : 1];
      |                   ^     ~~~~~~~~~~~                                   /data/data/com.termux/files/usr/tmp/v_10272/issue090224.01HP5P3S3A8R2QHSD5HMD4S9DZ.tmp.c:82:16: note: forward declaration of 'struct _option_bool'
   82 | typedef struct _option_bool _option_bool;
      |                ^
/data/data/com.termux/files/usr/tmp/v_10272/issue090224.01HP5P3S3A8R2QHSD5HMD4S9DZ.tmp.c:1077:36: error: invalid application of 'sizeof' to an incomplete type 'main__Foo' (aka 'struct _option_bool')
 1077 |         byte data[sizeof(main__Foo) > 1 ? sizeof(main__Foo) : 1];
      |                                           ^     ~~~~~~~~~~~           /data/data/com.termux/files/usr/tmp/v_10272/issue090224.01HP5P3S3A8R2QHSD5HMD4S9DZ.tmp.c:82:16: note: forward declaration of 'struct _option_bool'
   82 | typedef struct _option_bool _option_bool;
      |                ^
...                                                                           ==================
(Use `v -cg` to print the entire error message)

builder error:
==================
C error. This should never happen.

This is a compiler bug, please report it using `v bug file.v`.

https://github.com/vlang/v/issues/new/choose

You can also use #help on Discord: https://discord.gg/vlang
```

### V version

V 0.4.4 a4d0d4e

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.4 14afda7.a4d0d4e
[...]

PR #20769, opened 2024-02-09
 parser: disallow option alias with option parent type 

Delta456:
Fixes https://github.com/vlang/v/issues/20763


vlib/v/parser/parse_type.v
```diff
@@ -534,6 +534,11 @@ fn (mut p Parser) parse_type() ast.Type {
 		if is_option && sym.info is ast.SumType && sym.info.is_anon {
 			p.error_with_pos('an inline sum type cannot be an Option', option_pos.extend(p.prev_tok.pos()))
 		}
+
+		if is_option && sym.info is ast.Alias && sym.info.parent_type.has_flag(.option) {
+			p.error_with_pos('alias type cannot be option as parent type is already option',
+				option_pos.extend(p.prev_tok.pos()))
```
Delta456, line 540: I can't think of a better error message, @JalonSolov can you suggest a better error message?

JalonSolov (approved):

Issue #18324, opened 2023-06-01
orm: segfault when having nested foreign keys


mak3e:
### Describe the bug

If I define a foreign key relation to a table that has foreign key relation of its own the code compiles but results a runtime segfault even without a database connection.

### Expected Behavior

No segfault

### Current Behavior

`signal 11: segmentation fault`

### Reproduction Steps

```v
module main

import db.pg

[table: 'example']
pub struct Example {
    id        int        [primary; sql: serial]
    example2s []Example2 [fkey: 'example_id']
}

[table: 'example2']
pub struct Example2 {
    id         int        [primary; sql: serial]
    example_id int
    example3s  []Example3 [fkey: 'example2_id']
}

[table: 'example3']
pub struct Example3 {
    id          int [primary; sql: serial]
    example2_id int
}

fn main() {
    db := pg.connect(pg.Config{}) or { exit(0) }
    sql db {
        create table Example3
        create table Example2
        create table Example
    } or { print(err) }
}
```

### V version

V 0.3.4 a647a71

### Environment details (OS name and version, etc.)

V full version: V 0.3.4 a938dcd.a647a71
OS: linux, "Gentoo Linux"
Processor: 16 cpus, 64bit, little endian, Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz

getwd: /home/make
vexe: /home/make/v/v
vexe mtime: 2023-05-30 18:54:28

vroot: OK, value: /home/make/v
VMODULES: OK, value: /home/make/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.40.1
Git vroot status: weekly.2023.21-53-ga647a71c
.git/config present: true

CC version: cc (Gentoo 13.1.1_p20230527 p3) 13.1.1 20230527
thirdparty/tcc status: thirdparty-linux-amd64 12f392c3

spytheman:
https://github.com/vlang/v/pull/20774 fixes the compilation of this example.

PR #20774, opened 2024-02-10
ast: make Scope.find methods more robust, when called on default initialised `scope &Scope = unsafe { nil }` fields

spytheman:
Fix https://github.com/vlang/v/issues/18324 .

### DIFF
--- a/vlib/v/tests/orm_create_several_tables_test.v
+++ b/vlib/v/tests/orm_create_several_tables_test.v
@@ -1,0 +1,46 @@
+module main
+
+import os
+import db.sqlite
+
+@[table: 'visits']
+struct Visit {
+	id   int    @[primary; sql: serial]
+	site string
+}
+
+@[table: 'sites']
+struct Site {
+	hostname string  @[primary]
+	owner    int
+	visits   []Visit @[fkey: 'site']
+}
+
+@[table: 'users']
+struct User {
+	id    int    @[primary; sql: serial]
+	name  string
+	sites []Site @[fkey: 'owner']
+}
+
+fn init_db(path string) !sqlite.DB {
+	mut db := sqlite.connect(path)!
+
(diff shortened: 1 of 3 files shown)


vlib/v/tests/orm_create_several_tables_test.v
```diff
@@ -0,0 +1,46 @@
...
+
+fn test_creating_db() {
+	db_folder := os.join_path(os.vtmp_dir(), 'creating_db')
+	os.mkdir_all(db_folder)!
+	db_path := os.join_path(db_folder, 'abc.db')
+	dump(db_path)
+	mut db := init_db(db_path)!
+	dump(db)
+	db.close()!
+}
```
nkoehring, line 46: Afaik, you can simplify (and accelerate) this, by using an in-memory DB.
```suggestion
fn test_creating_db() ! {
	mut db := sqlite.connect(':memory:')!

	sql db {
		create table User
		create table Site
		create table Visit
	}!

	db.close()!
}
```

spytheman: Thank you.
Yes, it does exhibit the same problem on master, and it compiles here.

spytheman: Simplified in 4eca32b .

nkoehring:
One could also add a test, if all the tables are included. Not sure, if that is not already exhaustively tested, somewhere. One could otherwise add something like:
```v
nr_tables := db.exec('select count(*) from sqlite_master')!.first()

assert nr_tables.vals.first().int() >= 2 // result is probably bigger than 2, because of auto index tables
```

spytheman:
This is a test for a compiler error, its purpose is not to test actual behavior of the ORM or db.sqlite. It is enough as it is.

PR #20776, opened 2024-02-10
debug: fix scope cmd with variable in if scope

### LINKED ISSUE #20755: C error when using `$dbg` in a branch of an if/match expression on the right side of :=
**V doctor:**
```
V full version: V 0.4.4 5306469.b7b47fe

vexe mtime: 2024-02-08 11:43:40

Git vroot status: weekly.2024.05-47-gb7b47fe1
.git/config present: true

```

**What did you do?**
```v
fn main () {
	x := if true {
		$dbg // this tries to access `x`, which is undeclared at this point
		1
	} else {
		0
	}
	println(x)
}

```

vdbg breaking on `$dbg` breakpoint and then `1` printed to stdout

**What did you see instead?**
```
==================
C:/Users/user/AppData/Local/Temp/v_0/test.01HP48GV0212V7X79WSHQR09MT.tmp.c:18574: warning: implicit declaration of function 'CreateSymbolicLinkW'
C:/Users/user/AppData/Local/Temp/v_0/test.01HP48GV0212V7X79WSHQR09MT.tmp.c:22916: error: 'x' undeclared
...
==================

builder error: 
==================

```

Relevant generated C code:
```c
	int _t1; /* if prepend */
	if (true) {
		{
			Map_string_string _scope = new_map_init(&map_hash_string, &map_eq_string, &map_clone_string, &map_free_string, 1, sizeof(string), sizeof(v__debug__DebugContextVar),
				_MOV((string[1]){_SLIT("x")}),
				_MOV((v__debug__DebugContextVar[1]){{.typ=_SLIT("int"),.value=int_str(x)}}));
			v__debug__Debugger_interact(&g_debugger, (v__debug__DebugContextInfo){.is_anon=false,.is_generic=false,.is_method=false,.receiver_typ_name=_SLIT(""),.line=3,.file=_SLIT("test.v"),.mod=_SLIT("main"),.fn_name=_SLIT("main"),.scope=_scope});
		}_t1 = 1;
	} else {
		_t1 = 0;
	}
	int x =  _t1;
	println(int_str(x));
}
```

**Additional info**

This is also the case with `match`:
```v
fn main () {
	x := match true {
		true { $dbg 1 }
		false { 0 }
	}
	println(x)
}
```

> [!NOTE]
>

felipensp:
Fix #20755


vlib/v/ast/scope.v
```diff
@@ -182,6 +182,19 @@ pub fn (s &Scope) innermost(pos int) &Scope {
 	return s
 }
 
+// is_inner verifies if the supplied scope is inner the scope
+pub fn (s &Scope) is_inner(s2 &Scope) bool {
+	for sc := unsafe { s }; true; sc = sc.parent {
```
spytheman, line 187: Check `if s == unsafe { nil } {` first, and return early in this case.
Scopes are used in a lot of places through reference fields, that are not always != 0,
and in that case, code calling .is_inner would segfault at comptime, which is not always easy to debug (on windows especially, backtraces do not work reliably).

spytheman (approved):

PR #20784, opened 2024-02-11
x.vweb: add full static website support

davlgd:
If there is an `index.html` or an `index.htm` file at the root of a folder served static, it will be sent even if the URL ends with `/folder_name` or `/folder_name/`, without the file name. It allows to use `x.vweb` to be used to serve a full static website.

A test has been added and checked.


MCausc78:
Please run `v fmt -w vlib/x/vweb/vweb.v`

JalonSolov:
Are there _any_ systems that are less the 40 years old that have to be constrained to the letter extensions?  I'd vote for only working with `.html` and letting `.htm` die with the older, not longer supported versions of Windows.

spytheman (approved):
Excellent work.

PR #20785, opened 2024-02-11
cgen: use C99 short fixed array init syntax

zeozeozeo:
Before this change, cgen would initialize fixed arrays like this:

```c
Array_fixed_u8_5 buffer = {0, 0, 0, 0, 0};
```

This means that large arrays would generate *a lot* of C code just for initializing the array, and that significantly increased compilation times. C99 allows this short initialization syntax (see C99 6.7.8/21):

```c
Array_fixed_u8_2048 buffer = {0}; // array of 2048 zeroes
```

Size of v.c before this change: 6,35 MB (6 665 526 bytes)
Size of v.c after this change: 6,11 MB (6 411 950 bytes) **(-253 kb)**


spytheman:
Closing, because #23074 implements something similar.

Issue #20428, opened 2024-01-07
checker: interface casting fails


MCausc78:
### Describe the bug

<title>

### Reproduction Steps

```v
interface Foo {
	f()
}

struct Bar {}

struct Baz {}

fn (_ Bar) f() {
	println('bar')
}

fn (_ Bar) secret1() {
	println('secret 1')
}

fn (_ Baz) f() {
	println('baz')
}

fn (_ Bar) secret2() {
	println('secret 2')
}

fn h(foo Foo, i int) {
	foo.f()
	match i {
		0 { Bar(foo).secret1() }
		1 { Baz(foo).secret2() }
		else {}
	}
}

fn main() {
	h(Bar{}, 0)
	h(Baz{}, 1)
}

```

### Expected Behavior

Successful compile when using `x as T`, and output:
```
bar
secret 1
baz
secret 2
```

### Current Behavior

```v
test.v:28:7: error: cannot cast `Foo` to struct
   26 |     foo.f()
   27 |     match i {
   28 |         0 { Bar(foo).secret1() }
      |             ~~~~~~~~
   29 |         1 { Baz(foo).secret2() }
   30 |         else {}
test.v:28:7: error: cannot cast `Foo` interface value to `Bar`, use `foo as Bar` instead
   26 |     foo.f()
   27 |     match i {
   28 |         0 { Bar(foo).secret1() }
      |             ~~~~~~~~
   29 |         1 { Baz(foo).secret2() }
   30 |         else {}
test.v:29:7: error: cannot cast `Foo` to struct
   27 |     match i {
   28 |         0 { Bar(foo).secret1() }
   29 |         1 { Baz(foo).secret2() }
      |             ~~~~~~~~
   30 |         else {}
   31 |     }
test.v:29:7: error: cannot cast `Foo` interface value to `Baz`, use `foo as Baz` instead
   27 |     match i {
   28 |         0 { Bar(foo).secret1() }
   29 |         1 { Baz(foo).secret2() }
      |             ~~~~~~~~
   30 |         else {}
   31 |     }
```
If i change `Bar(foo)`/`Baz(foo)` to `(foo as Bar)`/`(foo as Baz)`, i will get another error:
```v
test.v:28:12: error: cannot cast non-sum type `Foo` using `as`
   26 |     foo.f()
   27 |     match i {
   28 |         0 { (foo as Bar).secret1() }
      |                  ~~
   29 |         1 { (foo as Baz).secret2() }
   30 |         else {}
[...]

PR #20789, opened 2024-02-11
checker,cgen: fix interface casting

MCausc78:
1. Fix https://github.com/vlang/v/issues/20428
2. Add tests


vlib/v/tests/cast_interface_to_impl.v
spytheman, line 1: Test files should end with `_test.v` .

spytheman (approved):
Excellent work.

PR #20796, opened 2024-02-11
x.sessions: change session Store interface to use results instead of options

Casper64:
Change the session store implementation to use results instead of options. This way (database) errors can be propagated and handled by the end users instead of forcing a session store to handle the error.

Could not find any reason why the `db_store_test` failed very rarely (see [pr checks of 20774](https://github.com/vlang/v/actions/runs/7855494077/job/21437262845?pr=20774)). Adding error propagation will also help debug future issues.


spytheman (approved):
Excellent work.

PR #20800, opened 2024-02-12
Add dependabot

Lycs-D:
auto update action version


spytheman:
Can you explain in more details (or link to an explanation) on what will that do?

spytheman:
I found this:
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

spytheman:
It seems to be working already 🥳 ...

PR #20810, opened 2024-02-12
vtest,pref: add ability to have platform specific _test.v files

syrmel:
Allow OS-specific _test.v files analogous to the logic re: `foo_nix.c.v` to make it easier to write platform-specific tests without a lot of `$if !windows {}` type code.
 
`foo_nix_test.v` would only run on nix OSes and `bar_windows_test.v` only on Windows, etc. On a non-targeted host OS, v test will skip the test with the `[SKIP]` message.

Technically, this could affect existing tests (though I didn't find any in the V repo) that are unfortunately named something like `<something>_windows_test.v` without targeting intention but this would be caught by the dev seeing the [SKIP] message.

`should_test()` in cmd/tools/vtest.v has been modified to check for an OS suffix and skip the test if the suffix is not a target of the host OS.

`pub fn (this_os &OS) is_target_of(target string) bool` has been added to vlib/pref/should_compile.v to make a reusable comparison between a selected OS and the target strings, for example, to have generic logic for 'nix' will never match Windows or to deal with `android_outside_termux` as a valid Android target.

Future work to use `is_target_of` to simplify existing code has been marked with TODO.

### DIFF
--- a/vlib/v/tests/filtering_macos_test.v
+++ b/vlib/v/tests/filtering_macos_test.v
@@ -2,6 +2,6 @@ fn test_is_macos() {
 	$if macos {
 		assert true
 	} $else {
-		assert false
+		assert false, 'platform-specific test filtering failed'
 	}
 }
--- a/vlib/v/tests/filtering_macos_test.v
+++ b/vlib/v/tests/filtering_macos_test.v
@@ -1,0 +1,7 @@
+fn test_is_macos() {
+	$if macos {
+		assert true
+	} $else {
+		assert false
+	}
+}
--- a/vlib/v/tests/filtering_nix_test.v
+++ b/vlib/v/tests/filtering_nix_test.v
@@ -1,0 +1,7 @@
+fn test_is_nix() {
+	$if !windows {
+		assert true
+	} $else {
+		assert false
+	}
+}
--- a/vlib/v/tests/filtering_nix_test.v
+++ b/vlib/v/tests/filtering_nix_test.v
@@ -2,6 +2,6 @@ fn test_is_nix() {
 	$if !windows {
 		assert true
 	} $else {
-		assert false
+		assert false, 'platform-specific test filtering failed'
 	}
 }
--- a/vlib/v/tests/filtering_windows_test.v
+++ b/vlib/v/tests/filtering_windows_test.v
@@ -1,0 +1,7 @@
+fn test_is_windows() {
+	$if windows {
+		assert true
+	} $else {
+		assert false
+	}
+}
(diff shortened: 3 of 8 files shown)


spytheman:
Imho add `filtering_nix_test.v` and `filtering_windows_test.v` under `vlib/v/tests/` .

syrmel:
The `filtering_*_test.v` files have been added. The platform-specific logic does not apply to `v test-self` at this time so they are skipped in the self-test.

vlib/v/tests/filtering_macos_test.v
```diff
@@ -0,0 +1,7 @@
+fn test_is_macos() {
+	$if macos {
```
JalonSolov, line 2: Why?  If it's not run on macos, why the comptime to verify macos?

vlib/v/tests/filtering_nix_test.v
```diff
@@ -0,0 +1,7 @@
+fn test_is_nix() {
+	$if !windows {
```
JalonSolov, line 2: Why?  If it's not run on 'nix, why the comptime to verify !windows?

vlib/v/tests/filtering_windows_test.v
```diff
@@ -0,0 +1,7 @@
+fn test_is_windows() {
+	$if windows {
```
JalonSolov, line 2: Why?  If it's only run on windows, why the comptime to verify windows?

vlib/v/tests/filtering_macos_test.v
syrmel: To provide that the platform-specific test filtering works. If it doesn't, it will run on the wrong platform and the comptime logic will make it `assert false`. I'll add a comment to the negative assert to clarify.

vlib/v/tests/filtering_nix_test.v
syrmel: To provide that the platform-specific test filtering works. If it doesn't, it will run on the wrong platform and the comptime logic will make it `assert false`. I'll add a comment to the negative assert to clarify.

vlib/v/tests/filtering_windows_test.v
syrmel: To prove that the platform-specific test filtering works. If it doesn't, it will run on the wrong platform and the comptime logic will make it `assert false`. I'll add a comment to the negative assert to clarify.

syrmel:
This PR includes a fix for a subtle issue with skip_files sometimes not working as intended because they were matching on relative instead of absolute paths and ended up sometimes not matching when they should have.

spytheman (approved):
Excellent work 🙇🏻‍♂️ .

PR #20817, opened 2024-02-13
initial addition of AEAD ChaCha20Poly1305 algorithm in pure v

blackshirt:
This pull request is the next step in two modules that have been merged into the experimental `x.crypto` namespace. This effort also adds to the completeness and capabilities of v crypto modules in the form of additional support for the authenticated encryption with additional data (AEAD) algorithm in pure V. See [wiki](https://en.wikipedia.org/wiki/Authenticated_encryption) about AEAD.

In this pull request opportunity, I included the `AEAD` interface which is actually more appropriate in `crypto.cipher`, but it doesn't matter where it is here, because this still needs review from the community and from experts. See the discussion on [discord](https://discord.com/channels/592103645835821068/592320321995014154/1206029352412778577) to find out his background and motivation.

This module is mostly inspired by the Golang version and attempts to align and conform to the RFC 8439 standard [ChaCha20 and Poly1305 for IETF Protocols](https://datatracker.ietf.org/doc/html/rfc8439), with a few additions, likes supports two nonce sizes, 12 bytes length of RFC 8439 and 24 bytes length of extended nonce size, wich is backed up by `x.crypto.chacha20` which supports both sizes. For extended nonce size, see [XChaCha](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-xchacha-03) for more detail.

Please give a review, feedback or suggestion.
All are welcome
Thanks


blackshirt:
> Following RFC is good. Is there an RFC mentioning the 2 nonce sizes?

Its already in `x.crypto.chacha20` especially in `xchacha20.v` file for supporting extended ChaCha20 construction with 24 bytes nonce

vlib/x/crypto/chacha20poly1305/usage_test.v
```diff
@@ -0,0 +1,64 @@
...
+	// the data mostly based on rfc
+	// gives yours secure key
+	key := hex.decode('1c9240a5eb55d38af333888604f6b5f0473917c1402b80099dca5cbc207075c0')!
+	// gives your additional data
+	aad := hex.decode('f33388860000000000004e91')!
+	// gives your nonce
+	nonce := hex.decode('000000000102030405060708')!
+
+	// your message to be encrypted and authenticated
+	plaintext := hex.decode('496e7465726e65742d4472616674732061726520647261667420646f63756d656e74732076616c696420666f722061206d6178696d756d206f6620736978206d6f6e74687320616e64206d617920626520757064617465642c207265706c616365642c206f72206f62736f6c65746564206279206f7468657220646f63756d656e747320617420616e792074696d652e20497420697320696e617070726f70726961746520746f2075736520496e7465726e65742d447261667473206173207265666572656e6365206d6174657269616c206f7220746f2063697465207468656d206f74686572207468616e206173202fe2809c776f726b20696e2070726f67726573732e2fe2809d')!
```
spytheman, line 15:
```suggestion
	plaintext := 'Internet-Drafts are draft documents valid for a maximum of six months and may be updated, replaced, or obsoleted by other documents at any time. It is inappropriate to use Internet-Drafts as reference material or to cite them other than as /“work in progress./”'.bytes()
```

vlib/x/crypto/chacha20poly1305/README.md
```diff
@@ -0,0 +1,38 @@
...
+Examples
+--------
+```v
+import encoding.hex
+import x.crypto.chacha20poly1305
+
+fn main() {
+    // plaintext message to be encrypted and authenticated
+    message := "Ladies and Gentlemen of the class of '99: If I could offer you only \
+    one tip for the future, sunscreen would be it.".bytes()
```
spytheman, line 22: nice :-)

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -0,0 +1,248 @@
...
+// it into `crypto.cipher` later.
+// Authenticated Encryption with Additional Data (AEAD) interface
+pub interface AEAD {
+	// nonce_size return the nonce size (in bytes) used by this AEAD algorithm that should be
+	// passed to `.encrypt` or `.decrypt`.
+	nonce_size() int
+	// tag_size returns the authenticated tag size (in bytes) produced by this AEAD algorithm.
+	tag_size() int
+	// overhead returns the maximum difference between the lengths of a plaintext and its ciphertext.
+	overhead() int
```
spytheman, line 32: why both?

spytheman: would not just `overhead()` be enough?

vlib/x/crypto/chacha20poly1305/usage_test.v
blackshirt: i dont know, `v fmt -w` does not format it nicely

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -0,0 +1,248 @@
...
+	// encrypt encrypts and authenticates the provided plaintext along with a nonce, and associated data.
+	// It returns ciphertext bytes where its encoded form is up to implementation and not dictated by the interfaces.
+	// Usually its contains encrypted text plus some authentication tag, and maybe some other bytes.
+	encrypt(plaintext []u8, nonce []u8, aad []u8) ![]u8
```
spytheman, line 36:
```suggestion
	encrypt(plaintext []u8, nonce []u8, ad []u8) ![]u8
```

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -0,0 +1,248 @@
...
+	// decrypt decrypts and authenticates (verifies) the provided ciphertext along with a nonce, and
+	// associated data. If verified successfully, it returns the plaintext and error otherwise.
+	decrypt(ciphertext []u8, nonce []u8, aad []u8) ![]u8
```
spytheman, line 39:
```suggestion
	decrypt(ciphertext []u8, nonce []u8, ad []u8) ![]u8
```

vlib/x/crypto/chacha20poly1305/usage_test.v
spytheman: It does for me, and the plaintext is actually readable:

image

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: `.tag_size()` tells exactly what size the of the tag (mac) of this AEAD produced ... some of aead based CCM, supports authentication tag sizes of different size, ie, 4, 6, 8, 10, 12, 14, and 16 bytes.
where `.overhead()` is intended to be more general and expandable in the semantic when its needed
consider the cases when this AEAD algorithm is not produces encrypted text + tag,  but more bytes for some reason.

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: commonly, `aad` comes from the term `authenticated additional data` because existing additional data is also authenticated ....is it should be changed ? or just updates the comments ?

blackshirt: i would change it

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: when strictly speaking with aead, i think its can just be `.tag_size()` and discard the `.overhead()`

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: Applied recently in [3b0e396](https://github.com/vlang/v/pull/20817/commits/3b0e396b95dfa980b54866cb3d24eedba228c936)

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: Done in [3b0e396](https://github.com/vlang/v/pull/20817/commits/3b0e396b95dfa980b54866cb3d24eedba228c936)

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: > would not just `overhead()` be enough?

after rethinking it again,  i thought if `.overhead()` is enough, for some reasons
- the ciphertext output from `.encrypt()` is in raw bytes and not dictated how this was encoded, its up to implementor to interprete this bytes. In current standard, the output are ciphertext associated with plaintext plus message authentication code (tag/mac).
- with this reason, semantically `.overhead()` is exactly same with `.tag_size()`., and we can drop one of  them ...
I would change this

blackshirt:
Thank for merging it ...
Have a nice day

PR #20818, opened 2024-02-13
v.debug: implement tracing hook feature

felipensp:
This PR adds a feature to v.debug to allow tracing fn call. Activated by `-d trace` V flag.

#### `debug.add_before_call(HookFnCall) HookFnCall`
Adds a new hook fn that will be called before each fn call and return its idx

#### `debug.add_after_call(HookFnCall) HookFnCall`
Adds a new hook fn that will be called after each fn call and return its idx

#### `debug.remove_before_call(HookFnCall)`
Remove a hook fn from before hook list

#### `debug.remove_after_call(HookFnCall)`
Remove a hook fn from after hook list

### Example

```V
import time
import v.debug

@[heap]
struct Profiler {
mut:
	fn_map map[string]time.StopWatch
}

fn (mut p Profiler) trace_after(fn_name string) {
	println('>>> called after ${fn_name} - ${p.fn_map[fn_name].elapsed()}')
}

fn (mut p Profiler) trace_before(fn_name string) {
	p.fn_map[fn_name] = time.new_stopwatch()
	println('>>> called before ${fn_name}')
}

fn test_call() {
	println('${@FN} got called')
	time.sleep(100)
}

fn main() {
	mut myprofiler := Profiler{}
	debug.add_after_call(myprofiler.trace_after)
	debug.add_before_call(myprofiler.trace_before)
	test_call()
}
```

```
$ v -d trace run trace.v
>>> called before main.test_call
test_call got called
>>> called before time.sleep
>>> called after time.sleep - 35.800us
>>> called after main.test_call - 137.600us
```


spytheman:
Why is it still draft, do you plan to add more features/bug fixes in the same PR?

felipensp:
> Why is it still draft, do you plan to add more features/bug fixes in the same PR?

No. I was just waiting for some feedback/tip/comment about to improved something on It.

vlib/v/debug/trace.v
```diff
@@ -0,0 +1,40 @@
...
+	g_trace.in_hook = false
+}
+
+pub fn add_after_call(func HookFnCall) {
+	g_trace.trace_after_call << func
+}
+
+pub fn add_before_call(func HookFnCall) {
+	g_trace.trace_before_call << func
+}
```
spytheman, line 40: There are no APIs for removing functions from the 2 arrays.

There are also no examples or comments for the public functions.

vlib/v/debug/trace.v
```diff
@@ -0,0 +1,66 @@
...
+	g_trace.in_hook = true
+	for func in g_trace.trace_before_call {
+		func(fn_name)
+	}
+	g_trace.in_hook = false
+}
+
+// add_after_call adds a fn hook to after hook list and returns its id
+@[inline]
+pub fn add_after_call(func &HookFnCall) int {
```
spytheman, line 40: The returned index is not safe to use, since if you delete a previous function, the indexes will change.

Imho pass `func &HookFnCall` to the remove calls too, and do a linear search based on `func`.

spytheman: Another option is to document that limitation, and require deleting in the reverse order of addition.

spytheman: Another is to not delete anything, but keep a flag in addition to the fn pointer, and just mark the entry as unused.

vlib/v/debug/tests/trace/trace_test.v
spytheman, line 1: Name the file something else, != the `trace_test.v` in the the debug folder.

vlib/v/debug/trace.v
felipensp: This was my first, but it does not works for methods at least. Since it creates closures on each mention.

Like

```C
v__debug__add_after_call((voidptr)__closure_create(_V_closure_main__Profiler_trace_after_538, &(*(myprofiler))));
_v__trace__main__main_v__debug__remove_after_call__35((voidptr)__closure_create(_V_closure_main__Profiler_trace_after_758, &(*(myprofiler))));
```

PR #20822, opened 2024-02-13
picohttpparser: add tests for u64toa

enghitalo:
## Question1 

`u64toa` function is taking `0.000 ms` to process `1 billion` of iterations. Is this right?

```bash
SPENT     0.000 ms in picohttpparser.u64toa 
```

```v
import picohttpparser

import benchmark

const max_iterations = 1_000_000_000

fn main() {
	v := u64(9199)

	mut b := benchmark.start()

	for i := 0; i < max_iterations; i++ {

		mut buf := [4]u8{}
		_ := unsafe { picohttpparser.u64toa(&buf[0], v) or { 0 } }
	}
	b.measure('picohttpparser.u64toa ')
}

```

## Question2

I have a C function `C.send` being used at 

```v
@[inline]
pub fn (mut r Response) end() int {
	n := int(i64(r.buf) - i64(r.buf_start))
	// use send instead of write for windows compatibility
	if C.send(r.fd, r.buf_start, n, 0) != n {
		return -1
	}
	return n
}
```

This function came from module `net`. But I am only using this funtion from this module.

What make the CI fail.

How can I solve it?

### DIFF
--- a/vlib/picoev/loop_default.c.v
+++ b/vlib/picoev/loop_default.c.v
@@ -76,14 +76,13 @@ fn (mut pv Picoev) poll_once(max_wait_in_sec int) int {
 		// Iterates through file descriptors and calls their callbacks for triggered events
 		for target in pv.file_descriptors {
 			if target.loop_id == pv.loop.id {
-				mut read_events := 0
-				if C.FD_ISSET(target.fd, &readfds) != 0 {
-					read_events |= picoev_read
-				}
-				if C.FD_ISSET(target.fd, &writefds) != 0 {
-					read_events |= picoev_write
-				}
-
+				// vfmt off
+				read_events := (
+					(if C.FD_ISSET(target.fd, &readfds) != 0 { picoev_read } else { 0 })
+						|
+					(if C.FD_ISSET(target.fd, &writefds) != 0 { picoev_write } else { 0 })
+				)
+				// vfmt on
 				if read_events != 0 {
 					$if trace_fd ? {
 						eprintln('do callback ${target.fd}')
--- a/vlib/picoev/loop_default.c.v
+++ b/vlib/picoev/loop_default.c.v
@@ -76,13 +76,14 @@ fn (mut pv Picoev) poll_once(max_wait_in_sec int) int {
 		// Iterates through file descriptors and calls their callbacks for triggered events
 		for target in pv.file_descriptors {
 			if target.loop_id == pv.loop.id {
-				// vfmt off
-				read_events := (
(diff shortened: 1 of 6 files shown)


Casper64:
`C.send` is the only C function that is used. It would be better if only that function is defined in `response.c.v`. So you don't need to import net and the compiler error also disappears

Casper64:
the `u64toa` function is a very fast implementation, so the time could be right. But more likely is that the compiler optimized the code, since the loop essentially does nothing and can't have any side effects.

spytheman:
When doing timing measurements, it is good to ensure that the compiler will not optimize out the calls in the loops, by using the results *after the loop*. Otherwise sufficiently smart compilers are free to ignore the loop entirely, if they decide, that it ultimately does nothing that will affect anything, except burn CPU time.

Try for example:
```v
import picohttpparser
import benchmark

const max_iterations = 1_000_000_000

fn main() {
	v := u64(9199)
	mut buf := [4]u8{}
	mut b := benchmark.start()
	mut sum := 0
	for i := 0; i < max_iterations; i++ {
		sum += unsafe { picohttpparser.u64toa(&buf[0], v) or { 0 } }
	}
	b.measure('picohttpparser.u64toa sum: ${sum} | buf: ${buf}')
}
```

enghitalo:
> When doing timing measurements, it is good to ensure that the compiler will not optimize out the calls in the loops, by using the results _after the loop_. Otherwise sufficiently smart compilers are free to ignore the loop entirely, if they decide, that it ultimately does nothing that will affect anything, except burn CPU time.
> 
> Try for example:
> 
> ```v
> import picohttpparser
> import benchmark
> 
> const max_iterations = 1_000_000_000
> 
> fn main() {
> 	v := u64(9199)
> 	mut buf := [4]u8{}
> 	mut b := benchmark.start()
> 	mut sum := 0
> 	for i := 0; i < max_iterations; i++ {
> 		sum += unsafe { picohttpparser.u64toa(&buf[0], v) or { 0 } }
> 	}
> 	b.measure('picohttpparser.u64toa sum: ${sum} | buf: ${buf}')
> }
> ```

There is a extrem difference between TCC and GCC here
image

JalonSolov:
Part of that "extreme" difference its the compile time.  Try compiling, _then_ timing the resulting executable.

Another "extreme" difference is that tcc _does not optimize_ anything, which gcc/clang will.  So if gcc/clang are optimizing the code away, you'll get the extremely short runtimes you're seeing, while with tcc, you'll see whatever time it takes if it isn't optimized.

PR #20840, opened 2024-02-15
dtm: separated the template generation from the cache system in test files

GGRei:


spytheman (approved):
Good choice of splitting the tests to several files.
That way only one of them could be marked as flaky as needed.

PR #20841, opened 2024-02-15
pref: use net.http.download_file to download photonwrapper.so instead of wget

### LINKED ISSUE #20834: `-use-coroutines` fails if wget isn't installed
### Describe the bug

I tried to use coroutines instead of native threads on Linux, but my computer doesn't have wget installed (I generally substitute it for curl).
On compilation with the `-use-coroutines` flag, an error occurs because it tries to use wget to download photonwrapper.so.

### Reproduction Steps

- Create a simple hello world in a `main.v` file
- Compile it with `v -use-coroutines main.v`

### Expected Behavior

The program is compiled successfully.

### Current Behavior

The compiler doesn't detect photonwrapper.so, so it tries to download it using wget, which fails as wget isn't found.
```
$ v -use-coroutines main.v
coroutines .so not found, downloading...
failed    cmd: wget -O "[...]/vlang/v/thirdparty/photon/photonwrapper.so" "https://github.com/vlang/photonbin/raw/master/photonwrapper_linux_amd64.so"
failed   code: 127
V panic: sh: line 1: wget: command not found

v hash: 4145669
/tmp/v_1000/v2.01HPMZBJW4B1YDHVR7G8XK741E.tmp.c:21529: at _v_panic: Backtrace
/tmp/v_1000/v2.01HPMZBJW4B1YDHVR7G8XK741E.tmp.c:35533: by os__execute_or_panic
/tmp/v_1000/v2.01HPMZBJW4B1YDHVR7G8XK741E.tmp.c:41956: by v__pref__parse_args_and_show_errors
/tmp/v_1000/v2.01HPMZBJW4B1YDHVR7G8XK741E.tmp.c:53823: by main__main
/tmp/v_1000/v2.01HPMZBJW4B1YDHVR7G8XK741E.tmp.c:54612: by main
```

### Possible Solution

The error comes from [vlib/v/pref/pref.c.v](https://github.com/vlang/v/blob/414566966d92d3e4d9ab2d36f6297fe830f044ed/vlib/v/pref/pref.c.v#L905), where wget is used directly from an os.execute command.

I can see that the previous line tries to use http.download_file, I'm not sure why it's commented out but maybe it is an option.
Otherwise, there needs to be a check on whether wget exists, potentially use curl as a fallback, and if neither are present, instructions to manually download and install the required file.

_No response_

### V version

0.4.4 4145669

V full version: V 0.4.4 a374d25.4145669
OS: linux, "Arch Linux"
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
> [!NOTE]
>

joe-conigliaro:
use net.http.download_file to download photonwrapper.so instead of wget

fixes #20834

### DIFF
--- a/vlib/v/pref/pref.c.v
+++ b/vlib/v/pref/pref.c.v
@@ -8,7 +8,7 @@ import os.cmdline
 import os
 import v.vcache
 import rand
-// import net.http // TODO can't use net.http on arm maccs: bignum.c arm asm error
+import net.http
 
 pub enum BuildMode {
 	// `v program.v'
@@ -901,8 +901,9 @@ pub fn parse_args_and_show_errors(known_external_commands []string, args []strin
 					so_url := 'https://github.com/vlang/photonbin/raw/master/photonwrapper_${os.user_os()}_${arch}.so'
 					if !os.exists(so_path) {
 						println('coroutines .so not found, downloading...')
-						// http.download_file(so_url, so_path) or { panic(err) }
-						os.execute_or_panic('wget -O "${so_path}" "${so_url}"')
+						http.download_file(so_url, so_path) or {
+							panic('error downloading ${so_url}')
+						}
 						println('done!')
 					}
 					res.compile_defines << 'is_coroutine'


joe-conigliaro:
oops

JalonSolov:
Heh.  Yeah, some odd reports there.

Looks like the macos-cross test needs another package installed.

The path testing test... we know tcc doesn't like non-ASCII paths, so why are those being used?

Bootstrapping... umm...

Linux-cross... same as bootstrapping.

Wild stuff.

vlib/v/pref/pref.c.v
```diff
@@ -8,7 +8,7 @@ import os.cmdline
 import os
 import v.vcache
 import rand
-// import net.http // TODO can't use net.http on arm maccs: bignum.c arm asm error
+import net.http
```
spytheman, line 11: No, v.pref is used by the compiler, and many other language tools.
Importing net.http here is bad. You can make a tool like wget in `cmd/tools/` and then invoke it here instead if you want.

joe-conigliaro: Yeah I saw the comment, I wanted to see what the errors were. Which is why I made the PR :)

PR #20842, opened 2024-02-15
test: pass json2 tests to test folder

enghitalo:


spytheman:
I do not get the point of moving them, but ok.

PR #20846, opened 2024-02-16
cgen, ast, checker: fix auto deref arg when fn expects ref

### LINKED ISSUE #20703: C error / UB when smartcasting a sumtype to a reference of a summed type
**V doctor:**
```
V full version: V 0.4.4 5306469.fb0efc0

vexe mtime: 2024-02-01 19:06:51

Git vroot status: weekly.2024.05-11-gfb0efc0a
.git/config present: true

```

**What did you do?**
```v
struct Foo {
	foo i32
}

struct Bar {
	bar f32
}

type Foobar = Foo | Bar

fn match_sum(sum Foobar) {
	match sum {
		Foo {
			sum_foo(sum)
		}
		Bar {
			sum_bar(sum)
		}
	}
}

fn sum_foo(i &Foo) {
	// if Foo were a primitive type (e.g. int), this would be UB (i think?)
	println('foo: ${i}')
}

fn sum_bar(i Bar) {
	println('bar: ${i}')
}

fn main() {
	match_sum(Foo{foo: 5})
}

```

Foo printed to stdout

**What did you see instead?**
```
==================
C:/Users/user/AppData/Local/Temp/v_0/test.01HNK135B307WA0WA0D6CDH9GB.tmp.c:6929: warning: implicit declaration of function 'tcc_backtrace'
C:/Users/user/AppData/Local/Temp/v_0/test.01HNK135B307WA0WA0D6CDH9GB.tmp.c:13073: error: cannot convert 'struct main__Foo' to 'struct main__Foo *'
...
==================

builder error: 
==================

```
> [!NOTE]
>

felipensp:
Fix #20703

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -2248,6 +2248,7 @@ fn (mut g Gen) call_args(node ast.CallExpr) {
 		if is_variadic && i == expected_types.len - 1 {
 			break
 		}
+		mut is_smartcast := false
 		if arg.expr is ast.Ident {
 			if arg.expr.obj is ast.Var {
 				if arg.expr.obj.smartcasts.len > 0 {
@@ -2260,6 +2261,7 @@ fn (mut g Gen) call_args(node ast.CallExpr) {
 						if cast_sym.info is ast.Aggregate {
 							expected_types[i] = cast_sym.info.types[g.aggregate_type_idx]
 						}
+						is_smartcast = true
 					}
 				}
 			}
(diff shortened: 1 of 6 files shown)


vlib/v/gen/c/fn.v
```diff
@@ -2525,7 +2528,10 @@ fn (mut g Gen) ref_or_deref_arg(arg ast.CallArg, expected_type ast.Type, lang as
 			}
 		}
 	}
+	// check if the argument must be dereferenced or not
+	g.arg_no_auto_deref = is_smartcast && !arg_is_ptr && !exp_is_ptr && arg.should_be_ptr
```
spytheman, line 2532: hm, should not we remember the previous value, and restore it later?
What happens when nested callexprs are used?

felipensp: This is not a problem, this one is an Ident-only flag.

spytheman: ok

PR #20855, opened 2024-02-17
pref: allow fetching the photonwrapper .so (for the coroutines) with curl too, or print details for manual download

Varpie:
Fix for #20834 

This PR tries to use `curl` to download the coroutine's .so file for systems that do not use `wget`, and prints a message with manual download instructions if both `wget` and `curl` fail.


spytheman (approved):
Good work.

PR #20868, opened 2024-02-18
x.vweb: add configuration options and handle incomplete requests

Casper64:
This pr adds a `Config` struct to `x.vweb` to change the default behaviour of vweb.

When vweb encounters an incomplete request it will now wait until a timeout occurs for new data instead of immediately closing the connection. The `Config` struct contains an option to limit the number of incomplete requests per connetion that are allowed before the connection is closed.

The `Config` struct also adds the option to limit the maximum HTTP header size and the maximum body length for a request.

### DIFF
--- a/vlib/x/vweb/vweb.v
+++ b/vlib/x/vweb/vweb.v
@@ -216,6 +216,39 @@ pub struct RunParams {
 	port                 int  = 8080
 	show_startup_message bool = true
 	timeout_in_seconds   int  = 30
+	config               Config
+}
+
+// Config contains options that can change the internal behaviour of vweb
+pub struct Config {
+pub mut:
+	// maximum size of HTTP headers in bytes, default is 96KB
+	max_header_len int = 96_000
+	// maximum number of incomplete HTTP requests that can be send over a connection
+	// before vweb closes the connection
+	max_nr_of_incomplete_requests int = 3
+	// the maximum size of a request body in bytes, default is 1MB
+	max_request_body_len int = 1_000_000
+}
+
+struct IncomingRequest {
+pub mut:
+	req                    http.Request
+	buf                    []u8
+	nr_incomplete_requests int
+}
+
+// done frees the current incomplete request header buffer and reset the HTTP request.
+// use this function when the connection is kept open, but the request is handled
+@[manualfree]
+pub fn (mut ir IncomingRequest) done() {
+	ir.buf.clear()
+	ir.req = http.Request{}
+}
+
+// reset set the incomplete request counter to 0
+pub fn (mut ir IncomingRequest) reset() {
+	ir.nr_incomplete_requests = 0
 }
 
 struct FileResponse {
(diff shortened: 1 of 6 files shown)


vlib/x/vweb/vweb.v
```diff
@@ -216,6 +216,39 @@ pub struct RunParams {
...
+pub mut:
+	req                    http.Request
+	buf                    []u8
+	nr_incomplete_requests int
+}
+
+// done frees the current incomplete request header buffer and reset the HTTP request.
+// use this function when the connection is kept open, but the request is handled
+@[manualfree]
+pub fn (mut ir IncomingRequest) done() {
```
spytheman, line 244: Why is this public? The structure itself is not, and it is used only for the `incoming_requests` field of RequestParams, which itself is not public.

vlib/x/vweb/vweb.v
```diff
@@ -216,6 +216,39 @@ pub struct RunParams {
 	port                 int  = 8080
 	show_startup_message bool = true
 	timeout_in_seconds   int  = 30
+	config               Config
```
spytheman, line 219: why add a separate `config` field here?
Would not embedding Config be a better choice, from the point of view of the callers?
i.e. should they care, that some of the named values are directly accessible like `timeout_in_seconds: 10`, and some should use the more verbose `config: Config{ ... }` syntax?

vlib/x/vweb/vweb.v
```diff
@@ -216,6 +216,39 @@ pub struct RunParams {
...
+}
+
+// Config contains options that can change the internal behaviour of vweb
+pub struct Config {
+pub mut:
+	// maximum size of HTTP headers in bytes, default is 96KB
+	max_header_len int = 96_000
```
spytheman, line 226: why 96KB?

vlib/x/vweb/vweb.v
Casper64: Then I think I forgot to use it, I'll check

vlib/x/vweb/vweb.v
JalonSolov: Apache - 8K
    Nginx - 4K-8K
    IIS - 8K-16K
    Tomcat - 8K – 48K
    Node (<13) - 8K; (>13) - 16K

So... 49_152 (48K) should be the maximum needed to cover every eventuality, though 16_384 (16K) would handle most things.

vlib/x/vweb/vweb.v
Casper64: embedding makes more sense. All properties except for the `timeout_in_seconds` property are more common and will probably used most often. Timeout is going to be split into a read timeout, write timeout and request handling timeout in a next pr, similar to the old behaviour of vweb.

Casper64:
`io.BufferedReader` will erase any existing data from its buffer when it refills. Another solution is needed to store the already read data from an incomplete request. Or a custom "`StringReader`" which keeps the read data in the buffer

JalonSolov:
It shouldn't be a problem to cache the header until it's completely read.

Casper64:
> It shouldn't be a problem to cache the header until it's completely read.

No, but the problem is that I can't access the data to cache it. And can't read the data first, because the net library only takes a buffered reader has argument when parsing only the request headers.

So I'll probably need to add another function as well

JalonSolov:
Now that `x.vweb` has become `veb` (and `x.vweb` is deprecated), perhaps this should be adapted to `veb` if needed.

JalonSolov:
`x.vweb` was renamed to `veb`. Is this still an issue in `veb`, or can this PR be closed?

spytheman:
Closing for inactivity.

Issue #20858, opened 2024-02-17
net: net.listen_tcp fail in -prod mode


kbkpbot:
### Describe the bug

when compile a simple net app, it will fail in prod mode

### Reproduction Steps

test.v 
```v
module main
import net

fn main() {
	net.listen_tcp(.ip6, ':8081') or {
		ecode := err.code()
		panic('failed to listen ${ecode} ${err}')
	}
       println('hello world')
}
```

Compile in prod mode
```sh
v test.v -prod
```
Then run ./test

### Expected Behavior

should run without panic
```sh
hello world
```

### Current Behavior
V panic: failed to listen 0 net: socket error: 88; code: 88; binding to :8081 failed
v hash: 298a2a2
                                                        | 0x7f3953c29d90 | /lib/x86_64-linux-gnu/libc.so.6(+0x29d90)
                                                        | 0x7f3953c29e40 | /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80)
                                                        | 0x560c9ac42745 | ./test(+0x3745)

### Possible Solution

compile with tcc, will OK
```sh
v test.v
```

### V version

V 0.4.4 298a2a2

### Environment details (OS name and version, etc.)

V full version: V 0.4.4 349741d.298a2a2
OS: linux, Ubuntu 22.04.4 LTS
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz

getwd: /home/mars/v/wrk
vexe: /HD/github/kbkpbot/v/v
vexe mtime: 2024-02-17 11:53:11

vroot: OK, value: /HD/github/kbkpbot/v
VMODULES: OK, value: /home/mars/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.34.1
Git vroot status: weekly.2024.06-92-g298a2a2d
.git/config present: true

CC version: cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
thirdparty/tcc status: thirdparty-linux-amd64 99683af0

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

kbkpbot:
change 
```v
net.listen_tcp(.ip6, ':8081') 
```
to 
```v
net.listen_tcp(.ip, ':8081') 
```

The same panic.

[...]

PR #20872, opened 2024-02-19
net: fix -prod mode gcc bug

kbkpbot:
This fix issue #20858


spytheman (approved):
Excellent investigation.

PR #20879, opened 2024-02-20
tests: fix `init_global_test.v` failing occasionally

smalltalkman:

### DIFF
--- a/vlib/v/tests/init_global_test.v
+++ b/vlib/v/tests/init_global_test.v
@@ -154,6 +154,7 @@ fn test_global_shared_map() {
 fn switch2() u64 {
 	mut cnt := u64(0)
 	for {
+		cnt++
 		mtx.@lock()
 		f1, f2 = f2, f1
 		if f1 == 17.0 || f2 == 17.0 {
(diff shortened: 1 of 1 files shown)


spytheman:
I am not sure if that was the only reason to mark the test as flaky in 5277ce7 2022-04-30 ... We can perhaps remove `// vtest flaky: true` line at the top, and see if it will fail again in the near future on the CI.

smalltalkman:
> I am not sure if that was the only reason to mark the test as flaky in 5277ce7 2022-04-30 ... 

I'm only encountering this error in my current testing and can't determine if this error is the sole cause.

> We can perhaps remove `// vtest flaky: true` line at the top, and see if it will fail again in the near future on the CI.

My current idea is to observe it for a period of time in local testing

In addition, I didn't figure out the real intention of `assert n > 0`, so I'm worried that this PR destroyed the original author's code logic

spytheman:
I think the idea was to check, that it passed through at least one lock/unlock cycle.

spytheman:
> My current idea is to observe it for a period of time in local testing

ok

PR #20881, opened 2024-02-20
(W.I.P.) Time: offset from utc in minutes

enghitalo:
`offset_from_utc_in_minutes` time field to solve performance problems


spytheman:
That is a breaking change. Please test such changes in your own private repo, before submitting them here. The CI here is shared :-| .

PR #20884, opened 2024-02-21
sha256/512: fix behavioral bug in .checksum

blackshirt:
This PR basically fix behavioral bug in `crypto.sha256` and `crypto.sha512` especially in future deprecated `.checksum()` method of the Digest. This is discussed previously with @joe-conigliaro on discord channel, see this [thread](https://discord.com/channels/592103645835821068/592106336838352923/1209734382357577800) for detail.

Basically,  behavioral `.checksum()` bug comes from the facts the `.checksum()` just calls private `.checksum_internal()` without returning proper size of the underlying digest. Its affect `sha256` or `sha512` Digest when operates on non standard size of the module, ie., `size224` for `sha256` and the size other than `sha512.size` that sha512 support them.
This snippet from discord copied here for sake of clarity
```v
import encoding.hex
import crypto.sha256

fn main() {
    text := 'hello world\n'
    data := text.bytes()


    expected := hex.decode('95041dd60ab08c0bf5636d50be85fe9790300f39eb84602858a9b430')!
    hash := sha256.sum224(data)
    // with function based, it should produces correct behavioral result
    assert hash == expected

    // lets use digest based hash
    mut d := sha256.new224()

    _ := d.write(data)!
    out := d.sum([])
    // with sum, also produces correct behavioural result
    assert out == expected

    // lets use checksum 
    d.reset()
    _ := d.write(data)!
    out2 := d.checksum()
    // this assert would fail because out2 length is normal size (32 bytes) in this case
    assert out2 == expected
}
```

This PR comes with the fix for `sha256.Digest.checksum` and `sha512.Digest.checksum` and additional test confirm it.
Please give it a review...

Thanks @joe-conigliaro

### DIFF
--- a/vlib/crypto/sha256/sha256.v
+++ b/vlib/crypto/sha256/sha256.v
@@ -201,7 +201,13 @@ fn (mut d Digest) checksum_internal() []u8 {
 @[deprecated: 'checksum() will be changed to a private method, use sum() instead']
 @[deprecated_after: '2024-04-30']
 pub fn (mut d Digest) checksum() []u8 {
-	return d.checksum_internal()
+	out := d.checksum_internal()
+	// if this digest has `size224` length, return the correct `size224` checksum
+	if d.is224 {
+		return out[0..sha256.size224]
+	}
+	// otherwise, returns a normal size
+	return out
 }
 
 // sum returns the SHA256 checksum of the bytes in `data`.
--- a/vlib/crypto/sha256/sha256_test.v
+++ b/vlib/crypto/sha256/sha256_test.v
@@ -28,3 +28,24 @@ fn test_crypto_sha256_writer_reset() {
 	sum := digest.sum([])
 	assert sum.hex() == 'dc7163299659529eae29683eb1ffec50d6c8fc7275ecb10c145fde0e125b8727'
 }
+
+fn test_crypto_sha256_224() {
+	data := 'hello world\n'.bytes()
+	mut digest := sha256.new224()
+	expected := '95041dd60ab08c0bf5636d50be85fe9790300f39eb84602858a9b430'
+
+	// with sum224 function
+	sum224 := sha256.sum224(data)
+	assert sum224.hex() == expected
+
+	// with sum
+	_ := digest.write(data)!
+	sum := digest.sum([])
+	assert sum.hex() == expected
+
+	// with checksum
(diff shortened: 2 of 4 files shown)


JalonSolov:
I would rather make `checksum()` private than worry about fixing it.

As I (vaguely) recall, it was only made public in the first place so it could be called from some other routine, instead of duplicating the code.

If it is kept as public, then at least deprecate it and come up with a better name.

blackshirt:
> I would rather make `checksum()` private than worry about fixing it.
> 
> As I (vaguely) recall, it was only made public in the first place so it could be called from some other routine, instead of duplicating the code.
> 
> If it is kept as public, then at least deprecate it and come up with a better name.

I think, even its a private, they should produce correct behavioral result.
I agree if it would be deprecated, maybe @joe-conigliaro have more better thing on this

JalonSolov:
If it's private, and the results are handled properly in the caller that has access, there's no need to modify it.

blackshirt:
> If it's private, and the results are handled properly in the caller that has access, there's no need to modify it.

Actually, this PR is useless in the sense the `.checksum` would be deprecated, i have no opinion on it ..

joe-conigliaro:
It's probably a good idea for it to have the correct behaviour even though its getting deprecated. probably no one has come across the issue yet besides @blackshirt which is good. what do you think @spytheman ?

blackshirt:
> It's probably a good idea for it to have the correct behaviour even though its getting deprecated. probably no one has come across the issue yet besides @blackshirt which is good. what do you think @spytheman ?

maybe the other people has some issues too..i found this after play a bit and experiment with this topic

spytheman:
> I would rather it had stayed private, but if it must remain public, it's better to have it return the correct values.

The reason, that it is deprecated, is to force people to not use it after a while, so that later we can make it private / delete it.

spytheman:
> It's probably a good idea for it to have the correct behaviour even though its getting deprecated. probably no one has come across the issue yet besides @blackshirt which is good. what do you think @spytheman ?

I agree. I do think that functions/methods that are easy to misuse, should be either private, or marked very well in the documentation, so that people can avoid them, and know what the pitfalls are.

spytheman (approved):
Good work.

blackshirt:
Thank you

PR #20885, opened 2024-02-22
os: don't check rdev equality on FreeBSD

kimshrier:
On FreeBSD, the stat.rdev value of a regular file and its parent directory are not necessarily the same.

### DIFF
--- a/vlib/os/os_stat_test.v
+++ b/vlib/os/os_stat_test.v
@@ -56,8 +56,10 @@ fn test_stat() {
 	assert dstat.get_filetype() == .directory
 	assert fstat.dev == dstat.dev, 'File and directory should be created on same device'
 	$if !freebsd {
+		assert fstat.rdev == dstat.rdev, 'File and directory should have same device ID'
+	} $else {
 		// On FreeBSD, the rdev values are not necessarily the same for non-devices
 		// such as regular files and directories.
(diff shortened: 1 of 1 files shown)


vlib/os/os_stat_test.v
```diff
@@ -55,5 +55,9 @@ fn test_stat() {
 	dstat := os.stat(temp_dir)!
 	assert dstat.get_filetype() == .directory
 	assert fstat.dev == dstat.dev, 'File and directory should be created on same device'
-	assert fstat.rdev == dstat.rdev, 'File and directory should have same device ID'
+	$if !freebsd {
+		// On FreeBSD, the rdev values are not necessarily the same for non-devices
+		// such as regular files and directories.
+		assert fstat.rdev == dstat.rdev, 'File and directory should have same device ID'
```
spytheman, line 61: Is there an example, that will reliably fail this assertion, given that both `temp_dir` and `test_file` are created on the fly for the test?

kimshrier: These values always look different on the test runs I have done.  In particular, the minor device number encoded into rdev always looks different.  Sometimes, the major device number encoded into rdev is the same.  The net effect is that rdev is not the same for the test file and directory.

spytheman: thanks

spytheman (approved):

PR #20886, opened 2024-02-22
ui: handle dpi change when moving window to another monitor

jkvatne:
Update the dpi scaling in ctx and ctx.ft when a window is dragged to a monitor with a different dpi scaling.

### DIFF
--- a/vlib/gg/gg.c.v
+++ b/vlib/gg/gg.c.v
@@ -407,10 +407,7 @@ fn gg_event_fn(ce voidptr, user_data voidptr) {
 		}
 		.resized {
 			ctx.scale = dpi_scale()
-			mut ft_scale := &ctx.ft.scale
-			unsafe {
-				*ft_scale = ctx.scale
-			}
+			ctx.ft.scale = ctx.scale
 			if ctx.config.resized_fn != unsafe { nil } {
 				ctx.config.resized_fn(e, ctx.config.user_data)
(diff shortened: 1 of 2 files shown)


vlib/gg/gg.c.v
```diff
@@ -406,6 +406,11 @@ fn gg_event_fn(ce voidptr, user_data voidptr) {
 			}
 		}
 		.resized {
+			ctx.scale = dpi_scale()
+			mut ft_scale := &ctx.ft.scale
+			unsafe {
+				*ft_scale = ctx.scale
+			}
```
spytheman, line 413: why not just `ctx.ft.scale = ctx.scale` ?

jkvatne: Because the compiler will not accept it:
`C:/Doc/v/vlib/gg/gg.c.v:419:11: error: field scale of struct &gg.FT is immutable`
And without the unsafe block:
`:/Doc/v/vlib/gg/gg.c.v:416:14: error: modifying variables via dereferencing can only be done in unsafe blocks`

But it is perhaps better to change the declaration of FT and use:
```
pub mut:
	scale       f32 = 1.0
```

spytheman: I think that changing the declaration is better. I see no reason for it, to not be mutable, and then changed in unsafe blocks.

Issue #20890, opened 2024-02-22
[x.json2] Add skippable struct fields @[json: '-']


johnpgr:
### Describe the feature

The main `json` module has a skippable struct fields feature, so fields on the struct don't get encoded / decoded

```v
struct Foo {
    x int @[json: '-'] // This field doesn't get encoded/decoded
    y int
    z int    
}
```

### Use Case

I wanted a way to encode a struct that didn't envolve mapping it to another struct, because only one field had to be hidden from json.
I could've used the main `json` module, but the way it encodes `time.Time` as a number doesn't work for me

### Proposed Solution

Implement the `@[json: '-']` attribute to skip struct fields

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.4 c9933da.10c2f9e

### Environment details (OS name and version, etc.)

OS: linux, Ubuntu 22.04.3 LTS (WSL 2)
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

johnpgr:
I can work on this

PR #20892, opened 2024-02-22
x.json2: add skippable field attr @json: '-'

johnpgr:
Adds the `@json: '-'` Attribute to skip struct fields in the encoding/decoding process of `x.json2` module
This feature already exists in the main `json` module
Adds tests to ensure the fields are correctly skipped from encoding/decoding

Addresses: #20890

### DIFF
--- a/vlib/x/json2/decoder.v
+++ b/vlib/x/json2/decoder.v
@@ -142,113 +142,120 @@ fn decode_struct[T](_ T, res map[string]Any) !T {
 	mut typ := T{}
 	$if T is $struct {
 		$for field in T.fields {
+			mut skip_field := false
 			mut json_name := field.name
+
 			for attr in field.attrs {
 				if attr.contains('json: ') {
 					json_name = attr.replace('json: ', '')
+					if json_name == '-' {
+						skip_field = true
+					}
 					break
 				}
 			}
 
-			$if field.is_enum {
-				if v := res[json_name] {
-					typ.$(field.name) = v.int()
-				} else {
-					$if field.is_option {
-						typ.$(field.name) = none
+			if !skip_field {
+				$if field.is_enum {
+					if v := res[json_name] {
+						typ.$(field.name) = v.int()
+					} else {
+						$if field.is_option {
+							typ.$(field.name) = none
+						}
 					}
-				}
-			} $else $if field.typ is u8 {
-				typ.$(field.name) = res[json_name]!.u64()
-			} $else $if field.typ is u16 {
-				typ.$(field.name) = res[json_name]!.u64()
-			} $else $if field.typ is u32 {
-				typ.$(field.name) = res[json_name]!.u64()
-			} $else $if field.typ is u64 {
-				typ.$(field.name) = res[json_name]!.u64()
-			} $else $if field.typ is int {
-				typ.$(field.name) = res[json_name]!.int()
-			} $else $if field.typ is i8 {
-				typ.$(field.name) = res[json_name]!.int()
-			} $else $if field.typ is i16 {
-				typ.$(field.name) = res[json_name]!.int()
-			} $else $if field.typ is i32 {
-				typ.$(field.name) = i32(res[field.name]!.int())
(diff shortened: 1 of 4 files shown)


JalonSolov:
Personally, I prefer the Go `omitempty`.  `-` doesn't mean anything to me.

Though I suppose they could have different meanings... `omitempty` means `skip if it's empty` where `-` could mean `always skip - period`.

johnpgr:
> Personally, I prefer the Go `omitempty`.  `-` doesn't mean anything to me.
> 
> Though I suppose they could have different meanings... `omitempty` means `skip if it's empty` where `-` could mean `always skip - period`.

`json: '-'` is a feature from golang, and it's already implemented in the main `json` module

enghitalo:
> > Personally, I prefer the Go `omitempty`.  `-` doesn't mean anything to me.
> > 
> > Though I suppose they could have different meanings... `omitempty` means `skip if it's empty` where `-` could mean `always skip - period`.
> 
> `json: '-'` is a feature from golang, and it's already implemented in the main `json` module

It's not because it was used in others it means that is good. The feature is good, basically, it skip fields to be processed by json. I just think you need to change the attribute name, since using the same attribute to rename the field and skip the field, it can be difficult to understand, for code maintenance and is also weird. Probably, change it to `skip_json` or something, should be better

johnpgr:
> > > Personally, I prefer the Go `omitempty`.  `-` doesn't mean anything to me.
> > > 
> > > Though I suppose they could have different meanings... `omitempty` means `skip if it's empty` where `-` could mean `always skip - period`.
> > 
> > `json: '-'` is a feature from golang, and it's already implemented in the main `json` module
> 
> It's not because it was used in others it means that is good. The feature is good, basically, it skip fields to be processed by json. I just think you need to change the attribute name, since using the same attribute to rename the field and skip the field can be difficult to understand, for code maintenance and is also weird. Probably, change it to `skip_json` or something, should be better 

I agree that using the same attribute to do two different things is awkward, but since: 

1. The main `json` module already uses this name to do that

2. There's no conflict in using the `json` name to achieve this because it makes no sense to to remap a struct field to a json field that you'd want to skip encoding/decoding

 I think it's best to use the same name, to have a easier replacement and keep the same behaviour between the two modules.

enghitalo:
@johnpgr IMHO, make sense have differences between json2 and json already json2 is in the experimental stage, this is why it is in x folder for now. It's a great place to try new things to get the best possible version of the library. json2 already have big differences from json all (features done) as an improved version from json.

spytheman:
@johnpgr please run `v fmt -w vlib/x/json2` .

spytheman:
> I think it's best to use the same name, to have a easier replacement and keep the same behaviour between the two modules.

I agree. Unless there is a reason to break compatibility, `json2` should support all field attributes that `json` does. It can support additional ones, but it is good to not make people that are transitioning to it, do busy work.

spytheman:
> It's a great place to try new things to get the best possible version of the library.

Trying new things, should not mean intentionally breaking compatibility, unless there is a good reason for that.

johnpgr:
> @johnpgr please run `v fmt -w vlib/x/json2` .

Done.

enghitalo:
> > It's a great place to try new things to get the best possible version of the library.
> 
> Trying new things, should not mean intentionally breaking compatibility, unless there is a good reason for that.

Well... I think that we should use `@[skip_json]` or something else and use a compiler warning to say that `@[json: '-']` is deprecaded. IMHO

spytheman:
> Well... I think that we should use @[skip_json] or something else and use a compiler warning to say that @[json: '-'] is deprecaded. IMHO

Currently, attributes do not have syntax for declaring them.
Any compiler checks for them, are implemented as special cases in the compiler.

I do not see a reason to expand this PR's scope, to add a compiler change too.
It currently does one thing (improves the compatibility with the older `json` module, and it does it well, as evidenced by the tests => it can be merged.

spytheman (approved):
Excellent work.

PR #20893, opened 2024-02-23
io: add new stringreader struct

Casper64:
This pr adds a new StringReader struct to the `io` module.

The `StringRedeader` is able to read data from a Reader interface and/or source string to a dynamically growing buffer using a string builder. 

Unlike the `BufferedReader`, `StringReader` will keep the entire contents of the buffer in memory, allowing the incoming data to be reused and read in an efficient matter.

The `StringReader` also implements the `io.ReaderWriter` interface

### DIFF
--- a/vlib/io/string_reader.v
+++ b/vlib/io/string_reader.v
@@ -285,7 +285,7 @@ pub fn (mut r StringReader) flush() string {
 // a new string builder with size 0 is initialized and the stringreaders offset is reset to 0
 @[unsafe]
 pub fn (mut r StringReader) free() {
-	r.builder.free()
+	unsafe { r.builder.free() }
 	r.builder = strings.new_builder(0)
 	r.offset = 0
 }
--- a/vlib/io/string_reader.v
+++ b/vlib/io/string_reader.v
@@ -15,7 +15,8 @@ pub struct StringReaderParams {
 // StringReader is able to read data from a Reader interface and/or source string to a dynamically
 // growing buffer using a string builder. Unlike the BufferedReader, StringReader will
 // keep the entire contents of the buffer in memory, allowing the incoming data to be reused
-// and read in an efficient matter.
+// and read in an efficient matter. The StringReader will not set a maximum capacity to the string
+// builders buffer and could grow very large.
 pub struct StringReader {
 mut:
 	reader ?Reader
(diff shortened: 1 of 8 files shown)


vlib/io/string_reader.v
```diff
@@ -0,0 +1,274 @@
...
+	reader ?Reader
+	// initialize the builder with this source string
+	source ?string
+	// if no source is given the string builder is initialized with this size
+	initial_size int
+}
+
+// StringReader is able to read data from a Reader interface and/or source string to a dynamically
+// growing buffer using a string builder. Unlike the BufferedReader, StringReader will
+// keep the entire contents of the buffer in memory, allowing the incoming data to be reused
```
blackshirt, line 17: If this reader keep the entire buffers content in memory, i think its a good to place limit on them, or it would grow uncontrollable

JalonSolov: Since it's a Reader, it's on the caller to control the size of the string.  If it was a Writer, then yes, it would have to handle not allowing the string to run you out of memory.

Casper64: That said, it would be useful to provide methods to clear the buffer and offset. I'll add them

Casper64: I've added a line in the doc comment to make it more clear: 
> The StringReader will not set a maximum capacity to the string builders buffer and could grow very large.

vlib/strings/builder.c.v
```diff
@@ -253,6 +254,20 @@ pub fn (mut b Builder) ensure_cap(n int) {
 	}
 }
 
+// grow_len grows the length of the buffer by `n` bytes if necessary
+@[unsafe]
+pub fn (mut b Builder) grow_len(n int) {
```
spytheman, line 259: Please add tests for the new public functions on `strings.Builder`.

vlib/io/string_reader.v
```diff
@@ -0,0 +1,291 @@
+module io
+
+import strings
```
spytheman, line 3: Nothing in `io` imports `strings` so far (only its tests indirectly through `os`).

I think, that it would be better, for keeping `io` itself small, if the new functionality, is in a separate module, for example `io.string_reader`, not directly in `io`.

Issue #10748, opened 2021-07-10
Alias array type requires cast


elliotchance:
**V doctor:**
```
OS: macos, macOS, 11.2.3, 20D91
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz
CC version: Apple clang version 12.0.0 (clang-1200.0.32.29)

getwd: /Users/elliot/v/src/github.com/elliotchance/v-bigint
vmodules: /Users/elliot/.vmodules
vroot: /Users/elliot/v
vexe: /Users/elliot/v/v
vexe mtime: 2021-07-05 22:56:24
is vroot writable: true
is vmodules writable: true
V full version: V 0.2.2 9e61321

Git version: git version 2.24.3 (Apple Git-128)
Git vroot status: Error: fatal: not a git repository (or any of the parent directories): .git
.git/config present: false
thirdparty/tcc status: thirdparty-macos-amd64 689c8a02-dirty
```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg bug/alloc_ints.v`
```v
type Ints = []int

nums := Ints{len: 3}
// nums := Ints([]int{len: 3})

println(nums)

```

**What did you expect to see?**

[0, 0, 0]

**What did you see instead?**
```
bug/alloc_ints.v:3:9: error: alias type name: []int is not struct type
    1 | type Ints = []int
    2 | 
    3 | nums := Ints{len: 3}
      |         ~~~~~~~~~~~~
    4 | // nums := Ints([]int{len: 3})
    5 |
```

I know the error message is clear. However, the compiler knows the aliased type so I shouldn't need to also provide the underlying type.

yuyi98:
I personally disagree with this usage.

felipensp:
Already fixed.

PR #20898, opened 2024-02-24
ast, parser, fmt: fix alias array no cast init (fix #20715 #10748)

### LINKED ISSUE #20715: type alias for 2D array generates compile time error
### Describe the bug

This does not compile:
```
module main

pub type Labels = [][]int

pub fn new_labels(width int, height int) Labels {
	//mut labels := [][]int{len: height, init: []int{len: width}}  // compiles OK
	mut labels := Labels{len: height, init: []int{len: width}}  // does not compile
	return labels
}

fn main() {
	mut labels := new_labels(2,2)
}
```

### Reproduction Steps

Try to compile the above code.

### Expected Behavior

I would expect it to compile with no error.

### Current Behavior

Get compiler error:
```
issues/complier_error.v:6:16: error: alias type name: [][]int is not struct type
    4 | 
    5 | pub fn new_labels(width int, height int) Labels {
    6 |     mut labels := Labels{len: height, init: []int{len: width}}
      |                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    7 |     return labels
    8 | }

```

### Possible Solution

Either fix or update documentation to say its not allowed.

_No response_

### V version

V 0.4.4 ac2dcc2

EndeavourOS
> [!NOTE]
>

yuyi98:
This PR fix alias array no cast init (fix #20715 #10748).

- Fix alias array no cast init.
- Add test.

```v
pub type Labels = [][]int

pub fn new_labels(width int, height int) Labels {
	mut labels := Labels{len: height, init: []int{len: width}}
	return labels
}

fn main() {
	mut labels := new_labels(2, 2)
	println(labels)
	assert labels == [[0, 0], [0, 0]]
}

PS D:\Test\v\tt1> v run .
Labels([[0, 0], [0, 0]])
```

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -1492,6 +1492,7 @@ pub mut:
 	elem_type  Type   // element type
 	init_type  Type   // init: value type
 	typ        Type   // array type
+	alias_type Type   // alias type
 }
 
 pub struct ArrayDecompose {
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -1710,7 +1710,11 @@ pub fn (mut f Fmt) array_init(node ast.ArrayInit) {
 	if node.exprs.len == 0 && node.typ != 0 && node.typ != ast.void_type {
 		// `x := []string{}`
 		f.mark_types_import_as_used(node.typ)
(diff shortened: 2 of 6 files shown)


spytheman (approved):
Excellent work.

PR #20906, opened 2024-02-25
picohttpparser: add tests for the Request methods

enghitalo:


vlib/picohttpparser/request_test.v
```diff
@@ -0,0 +1,73 @@
...
+	assert req.headers[0].name == 'Host'
+	assert req.headers[0].value == 'example.com'
+}
+
+pub fn test_handles_invalid_requests() {
+	mut req := Request{}
+	err := req.parse_request('INVALID REQUEST') or {
+		assert false, 'error while parse request: ${err}'
+		assert err.msg.contains('parse error')
+		assert err.code == 55
```
spytheman, line 22: That does not make sense.
Once the first `assert false` fails, the rest will not be executed.

spytheman (approved):

PR #20914, opened 2024-02-26
time: speed up offset from utc

enghitalo:
- public `offset_from_utc_in_seconds` to avoid unnecessary calls to the `localtime` function, without break changes


vlib/time/time.c.v
```diff
@@ -24,25 +24,14 @@ fn C.gmtime(t &C.time_t) &C.tm
 fn C.gmtime_r(t &C.time_t, res &C.tm) &C.tm
 fn C.strftime(buf &char, maxsize usize, const_format &char, const_tm &C.tm) usize
 
+pub const offset_from_utc_in_seconds = init_offset_from_utc_in_seconds_const()
+
```
spytheman, line 28: You can not just cache that unfortunately in a constant. It can change while the program is running, due to daylight savings time, or due to the system physically moving to another location, or due to the user changing his timezone.

enghitalo: One option is to use globals. So, the user or system can update it whenever.

spytheman: No, keeping state/cache should not be the default.

PR #20916, opened 2024-02-27
[x.json2] [FIX] Encoding Structs with Skipped fields not generating commas correctly

johnpgr:
### Yesterday my PR #20892 adding a missing feature in the x.json2 module got merged.
Although all tests were passing, the encoding of structs were not working properly for some reason, even if they were on the tests.
However my struct encoding tests were only using generic structs that only had fields of the same type:
```v
struct StructTypeSkippedFields[T] {
mut:
	val  T @[json: '-']
	val1 T
	val2 T @[json: '-']
	val3 T
}
```

These kind of structs were passing on the tests. However as soon as I tested them with non generic structs with different types on each field, they were not encoding correctly, having missing commas

This is what it would look like:

```v
struct StructTypeSkippedField5 {
pub mut:
	val string @[json: '-']
	val1 int @[json: '-']
	val2 f64
	val3 bool
	val4 string
	val5 string
}

json2.encode(StructTypeSkippedField5{
	val: 'foo'
	val1: 999
	val2: 999.0
	val3: true
	val4: 'bar'
	val5: 'baz'
}) // The result would be: {"val2": 999.0,"val3":true"val4":"bar""val5":"baz"}
```

### This PR Fixes this bug, and adds new tests to ensure they are always encoded correctly

### DIFF
--- a/vlib/x/json2/encoder.v
+++ b/vlib/x/json2/encoder.v
@@ -177,17 +177,27 @@ fn (e &Encoder) encode_struct[U](val U, level int, mut buf []u8) ! {
 	mut fields_len := 0
 
 	$for field in U.fields {
-		$if field.is_option {
-			if val.$(field.name) != none {
+		mut @continue := false
+		for attr in field.attrs {
+			if attr.contains('json: ') {
+				if attr.replace('json: ', '') == '-' {
+					@continue = true
+				}
+				break
+			}
+		}
+		if !@continue {
+			$if field.is_option {
+				if val.$(field.name) != none {
+					fields_len++
+				}
+			} $else {
 				fields_len++
 			}
-		} $else {
-			fields_len++
 		}
 	}
 	$for field in U.fields {
 		mut ignore_field := false
-		mut skip_field := false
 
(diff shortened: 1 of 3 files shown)


vlib/x/json2/encoder.v
```diff
@@ -177,17 +177,27 @@ fn (e &Encoder) encode_struct[U](val U, level int, mut buf []u8) ! {
 	mut fields_len := 0
 
 	$for field in U.fields {
-		$if field.is_option {
-			if val.$(field.name) != none {
+		mut @continue := false
```
spytheman, line 180:
```suggestion
		mut should_continue := false
```

spytheman: or `should_skip`

spytheman: `continue` on its own is both too generic, and a keyword requiring an escape with `@` as you did. A longer compound name can be more descriptive, and avoid the use of an escape.

spytheman (approved):
Excellent work.

PR #20934, opened 2024-03-01
cgen: fix global initializer of fixed array on gcc

felipensp:
Fix code generated to build `vcc` on with `-cc gcc`


spytheman (approved):
Excellent work.
Can you please add a small test in `vlib/v/gen/c/testdata/` too?

spytheman (approved):

Issue #19159, opened 2023-08-17
cgen error for `Name[int](123)`, when `type Name[T] = string | int | T`


spytheman:
**V doctor:**
```
V full version: V 0.4.0 2fec2fa.c5e1c8e
OS: linux, Ubuntu 20.04.6 LTS
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-3225 CPU @ 3.30GHz

getwd: /v/vnew
vexe: /v/vnew/v
vexe mtime: 2023-08-17 13:54:55

vroot: OK, value: /v/vnew
VMODULES: OK, value: /home/delian/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.41.0
Git vroot status: weekly.2023.33-5-gc5e1c8ee
.git/config present: true

CC version: cc (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
thirdparty/tcc status: thirdparty-linux-amd64 12f392c3

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg a.v`
```v
type Name[T] = string | int | T
_ := Name[int](123)

```

**What did you expect to see?**

either a compiled program, or a checker error

**What did you see instead?**
```
==================
/tmp/v_1000/a.13613374991411628734.tmp.c:948: error: duplicate member '_int'
...
==================
(Use `v -cg` to print the entire error message)

builder error: 
==================
C error. This should never happen.

This is a compiler bug, please report it using `v bug file.v`.

https://github.com/vlang/v/issues/new/choose

You can also use #help on Discord: https://discord.gg/vlang

```

PR #20936, opened 2024-03-01
checker,cgen: fix duplicates in generic sumtype

MCausc78:
1. Fix https://github.com/vlang/v/issues/19159
2. Add tests
```v
PS D:\Games\Proekti\V\vl> type test.v
@[on_duplicate: 'thats sentinel lol'; notice_if_duplicate]
type Name[T] = string | int | T

fn main() {
        _ := Name[int](123)
}
PS D:\Games\Proekti\V\vl> .\v2 run test.v
test.v:5:7: notice: duplicate type: thats sentinel lol
    3 |
    4 | fn main() {
    5 |     _ := Name[int](123)
      |          ~~~~~~~~~~~~~~
    6 | }
```

Note: you can control, whether to raise a error, notice or ignore if user used type, which is present as T in generic sumtype


vlib/v/checker/checker.v
```diff
@@ -3082,6 +3082,22 @@ fn (mut c Checker) cast_expr(mut node ast.CastExpr) ast.Type {
 			ft := c.table.type_to_str(from_type)
 			tt := c.table.type_to_str(to_type)
 			c.error('cannot cast `${ft}` to `${tt}`', node.pos)
+		} else if to_sym_info.variants.len != to_sym_info.get_deduplicated_variants().len {
+			// not a option type because of autofree bug; see https://github.com/vlang/v/actions/runs/8116684464/job/22187255774?pr=20936
```
spytheman, line 3086: There is no point in linking to job runs permanently in the source code, since they expire after a week or so.

vlib/v/checker/checker.v
```diff
@@ -3082,6 +3082,22 @@ fn (mut c Checker) cast_expr(mut node ast.CastExpr) ast.Type {
...
+				if attr.name == 'on_duplicate' && attr.has_arg {
+					msg = attr.arg
+				}
+			}
+			if msg.len > 0 {
+				if print_notice {
+					c.note('duplicate type: ${msg}', node.pos)
+				} else {
+					c.error('duplicate type: ${msg}', node.pos)
+				}
```
spytheman, line 3099: hm, there is no existing precedent for using an attribute for toggling between an error and a notice 🤔. I am not sure if it is worth the complexity.
Would not either just a notice, or either just an error be simpler?

MCausc78: sometimes this is bad idea to give error. Consider that:
`firstapi`:
```v
pub struct Row {
pub mut:
	// children components
	components []Component
}

@[on_duplicate: 'T is present in sumtype']
@[notice_if_duplicate]
pub type Component[T] = Row
	| Button
	| StringSelect
	| TextInput
	| T
```
`secondapi`:
```v
struct Default[T] {
	v T
}

pub fn c[T]() Component[T] {
	return Component[T](Default[T]{}.v)
}
```
user (not dev of `firstapi` and `secondapi` packages), not knowing that e.g. `Row` is present in `Component` type, uses `c[Row]()`, and gets an notice, instead of hard error. That also would allow deprecating duplicates in generic sumtypes.
Yet one example:
```v
struct Undefined {}

@[on_duplicate: 'Sentinels are internal']
pub type UndefinedOr[T] = Undefined | T
```
That would give a error if user will do `UndefinedOr[Undefined]`.

vlib/v/checker/checker.v
MCausc78: you're right, i will link issue url instead, once ci passes

vlib/v/checker/checker.v
spytheman: @medvednikov what do you think?

medvednikov: I agree with @spytheman here. This complicates the language and gives too many options. I'd just stick with one option.

MCausc78: removed attribute and made `error` by default

spytheman: There are still 2 attributes:
`ignore_generic_duplicates` and `on_generic_duplicate`.

I do not know how to document/explain that in `doc/docs.md`.

medvednikov:
Sorry, we'll figure something out to merge this.

enghitalo:
I don't think using attributes or custom error messages is a good idea.

spytheman:
Closing for inactivity.

PR #20952, opened 2024-03-03
cgen: fix comptime `is` condition when using interface

### LINKED ISSUE #20918: Optional type matches incorrectly in $for field
### Describe the bug

Accessing optional field in compile time code causes C error

### Reproduction Steps

```v
module main

interface TestInterface {
	test_func() f64
}

struct Struct1 {
	num f64
}

fn (o &Struct1) test_func() f64 {
	return o.num
}

struct Struct2 {
	s1 ?&Struct1
}

fn do_thing[T]() T {
	t := T{}
	$for field in T.fields {
		$if field.typ is TestInterface {
			i := TestInterface(t.$(field.name))
			println('${i.test_func()}')
		}
	}
	return t
}

pub fn main() {
	do_thing[Struct2]()
}
```

### Expected Behavior

no C error

### Current Behavior

C error:
```
PS C:\Users\jcwea\Documents\git\vtest> v run .
==================
C:/Users/jcwea/AppData/Local/Temp/v_0/vtest.01HQPQ5SW9NC7GKNV47M8KWAVT.tmp.c:6995: warning: implicit declaration of function 'tcc_backtrace'
C:/Users/jcwea/AppData/Local/Temp/v_0/vtest.01HQPQ5SW9NC7GKNV47M8KWAVT.tmp.c:13239: error: cannot convert 'struct _option_main__Struct1_ptr' to 'struct main__Struct1 *'
...
==================

builder error:
==================

```

### Possible Solution

The best solution is probably to not match `?T`'s with `$if field.typ is T`, thus never calling the code in the $if statement. But we would likely still want to be able to match these optional field types somehow, maybe with a `$if field.underlying_typ is T` or something similar. I image it could work the same way as unwrapping an alias or a sum type during compile time.

affects vgdextension

### V version

V 0.4.4 7df5646

```
V full version: V 0.4.4 1733dab.7df5646

getwd: C:\Users\jcwea\Documents\git\vtest
vexe mtime: 2024-02-27 22:26:52

Git vroot status: weekly.2024.09-4-g7df56462
.git/config present: true

```
> [!NOTE]
>

felipensp:
Fix #20918

### DIFF
--- a/vlib/v/checker/if.v
+++ b/vlib/v/checker/if.v
@@ -10,6 +10,10 @@ fn (mut c Checker) check_compatible_types(left_type ast.Type, right ast.TypeNode
 	right_type := c.unwrap_generic(right.typ)
 	sym := c.table.sym(right_type)
 
+	if right_type.has_flag(.option) != left_type.has_flag(.option) {
+		return .skip
+	}
+
 	if sym.kind == .interface_ {
 		checked_type := c.unwrap_generic(left_type)
 		return if c.table.does_type_implement_interface(checked_type, right_type) {
--- a/vlib/v/checker/postfix.v
+++ b/vlib/v/checker/postfix.v
@@ -19,7 +19,8 @@ fn (mut c Checker) postfix_expr(mut node ast.PostfixExpr) ast.Type {
 			node.expr.pos())
 	}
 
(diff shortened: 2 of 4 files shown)


spytheman (approved):
Excellent work.

PR #20961, opened 2024-03-04
checker: fix missing check for interface cast of option type

felipensp:
Fix missing interface cast checking

```v
interface IFoo {
	a int
}

struct Foo {
	a int
}

fn main() {
	a := ?Foo{}
	dump(a)
	b := IFoo(a?)
	c := IFoo(a) // must be an error
}
```


spytheman (approved):
Excellent work.

PR #20966, opened 2024-03-05
json decode

enghitalo:
benchmark

```v
 SPENT    47.116 ms in new_json.decode[Stru](json_data)!
 SPENT    48.742 ms in old_json.decode(Stru, json_data)!

 SPENT     9.631 ms in new_json.decode[StructType[string]](json_data1)!
 SPENT    13.537 ms in old_json.decode(StructType[string], json_data1)!

 SPENT     9.970 ms in new_json.decode[StructTypeOption[string]](json_data1)!
 SPENT    13.733 ms in old_json.decode(StructTypeOption[string], json_data1)!

 SPENT    18.907 ms in new_json.decode[map[string]string](json_data1)!
 SPENT    32.577 ms in old_json.decode(map[string]string, json_data1)!
```
Next PRs
- [ ] skip `fulfill_nodes` when map
- [ ] use Sumtype methods with callback to decode Sumtypes
- [ ] handle escape


spytheman:
I do not like at all the `vlib/x/new_json/` folder, because it will confuse people that it is a 3rd json module (in addition to `json` and `x.json2`. 

Move the folder under `vlib/x/json2/`, for example `vlib/x/json2/decoder2`, or just use different names for its APIs and keep it directly inside `vlib/x/json2` .

enghitalo:
> Move the folder under vlib/x/json2/, for example vlib/x/json2/decoder2, or just use different names for its APIs and keep it directly inside vlib/x/json2 .

ok. make sense.

enghitalo:
I will reopen. I need solve a problem in my V repository

PR #20973, opened 2024-03-07
cgen: fix missing scope enclosing for const init which needs temp variables

### LINKED ISSUE #20970: builder error: Compiler confuses declarations of const expressions when they are located in different files
### Describe the bug

The bug happens when using expressions that result in different types for constants and the consts are declared in different files.

### Reproduction Steps

```v
// foo/main.v
import rand

struct MyStruct {
	foo int
}

const my_struct = MyStruct{rand.intn(6) or { panic(err) }}
// No issue when using a single file for the constants.
// const cfg_dir = os.join_path(os.config_dir() or { panic(err) }, 'foo') 

fn main() {
	dump('hello')
}
```
```v
// foo/config.v
import os

const cfg_dir = os.join_path(os.config_dir() or { panic(err) }, 'foo')
```
```
v foo
```

### Expected Behavior

Program is able to compile and run.

### Current Behavior

Builder error

```
==================
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/v_501/foo.01HRBP61MACMMT3V4G8Y5RKWV0.tmp.c:22751:13: error: redefinition of '_t1' with a different type: '_result_int' (aka 'struct _result_int') vs '_result_string' (aka 'struct _result_string')
_result_int _t1 = rand__intn(6);
            ^
/tmp/v_501/foo.01HRBP61MACMMT3V4G8Y5RKWV0.tmp.c:22742:16: note: previous definition is here
_result_string _t1 = os__config_dir();
               ^
3 warnings and 1 error generated.
...
==================

builder error: 
==================
```

### Possible Solution

_No response_

Tested on macOS arm and ubuntu arm.

### V version

-

```
V full version: V 0.4.4 790ea2f
OS: macos, macOS, 14.2.1, 23C71
Processor: 8 cpus, 64bit, little endian, Apple M1
```
> [!NOTE]
>

felipensp:
Fix #20970

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5722,7 +5722,7 @@ fn (mut g Gen) const_decl(node ast.ConstDecl) {
 						has_unwrap_opt_res)
 				} else {
 					g.const_decl_init_later(field.mod, name, field.expr, field.typ, false,
-						false)
+						true)
 				}
 			}
 		}
--- a/vlib/v/tests/modules/consts_with_complex_init/config.v
+++ b/vlib/v/tests/modules/consts_with_complex_init/config.v
@@ -1,0 +1,13 @@
+module main
+
+import rand
+
+struct MyStruct {
+	foo int
+}
+
+const my_struct = MyStruct{rand.intn(6) or { panic(err) }}
+
+fn main() {
+	dump('hello')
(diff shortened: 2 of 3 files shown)


spytheman (approved):
Excellent work.

PR #20974, opened 2024-03-07
cgen: fix fixed array return on fn with option generic return

### LINKED ISSUE #20969: Fixed Array Generic Function with Option/Result return type returns unexpected results
### Describe the bug

On V Playground:
The first element of the return value of a generic function with fixed array as the return type wrapped in option/result is a seemengly random big number that's then serialized into the array's type whilst all of the rest are 0

On Windows:
Same as before but the number always seems to be 21888480

### Reproduction Steps

```v
fn main() {
    println(example[[4]int]([1,2,3,4]!)?) //[409208832, 0, 0, 0], [-1935261600, 0, 0, 0], [2079242128, 0, 0, 0]
    println(example[[4]string](['A','B','C','D']!)?) //['', '', '', '']
    println(example[[4]bool]([true,false,true,false]!)?) //[true, false, false, false]
    println(example[[4]f64]([1.0,2.0,3.0,4.0]!)?) //[4.09207872e+08, 0.0, 0.0, 0.0], [-1.93526256e+09, 0.0, 0.0, 0.0], [2.079241168e+09, 0.0, 0.0, 0.0] 
}

fn example[T](arr T) ?T {
    return arr
}
```
Commented parts are what the function returned after multiple runs

### Expected Behavior

The array to be returned as it was passed

### Current Behavior

On V Playground:
The first element is a seemengly random big number that's then serialized into the array's type whilst all of the rest are 0

On Windows:
Same as before but the number always seems to be 21888480

### Possible Solution

_No response_

_No response_

### V version

V 0.4.4 6279e8a.60b4fb3

V full version: V 0.4.4 6279e8a.60b4fb3
Processor: 16 cpus, 64bit, little endian, 

getwd: C:\Users\nicof\Desktop\Projects\V\tyest
vexe: C:\Users\nicof\Desktop\v\v.exe
vexe mtime: 2024-03-04 20:23:22

vroot: OK, value: C:\Users\nicof\Desktop\v
VMODULES: OK, value: C:\Users\nicof\.vmodules
VTMP: OK, value: C:\Users\nicof\AppData\Local\Temp\v_0

Git version: git version 2.44.0.windows.1
.git/config present: false

thirdparty/tcc: N/A

felipensp:
Fix #20969

### DIFF
--- a/vlib/v/gen/c/array.v
+++ b/vlib/v/gen/c/array.v
@@ -151,15 +151,20 @@ fn (mut g Gen) fixed_array_init(node ast.ArrayInit, array_type Type, var_name st
 		ret_typ := g.typ(node.typ)
 		g.write('(${ret_typ})')
 	}
-	g.write('{')
+	elem_type := (array_type.unaliased_sym.info as ast.ArrayFixed).elem_type
+	mut elem_sym := g.table.final_sym(elem_type)
+
+	is_struct := g.inside_array_fixed_struct && elem_sym.kind == .struct_
+
+	if !is_struct {
+		g.write('{')
+	}
 	if node.has_val {
 		tmp_inside_array := g.inside_array_item
 		g.inside_array_item = true
 		defer {
 			g.inside_array_item = tmp_inside_array
 		}
-		elem_type := (array_type.unaliased_sym.info as ast.ArrayFixed).elem_type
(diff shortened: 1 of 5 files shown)


spytheman (approved):
Excellent work.

PR #20977, opened 2024-03-08
fmt: fix removal of used selective and alias imports in modules in `$VMODULES` dirs

ttytm:
Fixes issues the formatter introduces(/false negative formatting checks) in modules that are located in a `$VMODULES` directory.

E.g. the following, when located in `~/.vmodules`,...
```v
// ~/.vmodules/foo/main.v
import x.json2 as json
import datatypes { Stack }

const my_datatype = Stack[string]{}
```

will currently be formatted into:
```v
// ~/.vmodules/foo/main.v
import x.json2
import datatypes

const my_datatype = Stack[string]{}
```


spytheman (approved):
Excellent work.

PR #20984, opened 2024-03-09
ast: cleanup and simplify `shorten_user_defined_typenames` method

ttytm:
The PR updates the mentioned method. The change should also bring be a slight performance improvement (not tested). It will help to add a simple fix in a follow up for further issues when running `v fmt` on modules inside a `$VMODULES` directory.


spytheman (approved):
Good work.

PR #20989, opened 2024-03-09
ast, fmt: improve submodule type alias lookup; fix formatting of modules in `$VMODULES`

ttytm:
Fixes an issue where

```v
// ~/.vmodules/foo/foo.v || ~/.vmodules/foo/src/foo.v
module foo

import bar.baz

type MyAlias = baz.Baz
// Will be formatted into:
// type MyAlias = bar.baz.Baz
``` 

```v
// ~/.vmodules/foo/bar/baz/baz.v
module baz

enum Baz {
	bar
	baz
}
```


spytheman (approved):
Good work.

PR #20994, opened 2024-03-10
cgen: fix code generated to comptime passed to generic arg

### LINKED ISSUE #20990: cgen erro: wrong sumtype cast function?
### Reproduction Steps

```v
struct StructType[T] {
mut:
	val T
}

type SumTypes = StructType[string] | []SumTypes | []string | bool | int | string

pub struct Count {
mut:
	total int
}

// count_chars count json sizen whithout new encode
pub fn (mut count Count) count_chars[T](val T) {
	$if T is $sumtype {
		$for v in val.variants {
			if val is v {
				// dump(typeof(val).name)
				count.count_chars(val)
			}
		}
	} $else $if T is $struct {
		count.chars_in_struct(val)
	} $else {
	}
}

// chars_in_struct
fn (mut count Count) chars_in_struct[T](val T) {
	$for field in T.fields {
		va := val.$(field.name)
		count.count_chars(va)  // Not works
		count.count_chars(val.$(field.name)) // Not works
	}
}

fn main() {
	mut count := Count{}
	count.count_chars(StructType[SumTypes]{ val: '' })
	println('ok')
	assert true
}

```

### Expected Behavior

assert pass, print ok

### Current Behavior

```bash
==================
/tmp/v_1000/a.01HRM75XMWA9WKHPFBFAQVD0EY.tmp.c:13325: error: cannot convert 'struct main__SumTypes' to 'struct string'
...
==================

builder error: 
==================

```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.4 386bd77

Linux
> [!NOTE]
>

felipensp:
Fix #20990

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -2514,7 +2514,6 @@ fn (mut g Gen) expr_with_cast(expr ast.Expr, got_type_raw ast.Type, expected_typ
 					unwrapped_got_type = unwrapped_got_sym.info.types[g.aggregate_type_idx]
 					unwrapped_got_sym = g.table.sym(unwrapped_got_type)
 				}
-
 				fname := g.get_sumtype_casting_fn(unwrapped_got_type, unwrapped_expected_type)
 
 				if expr is ast.ArrayInit && got_sym.kind == .array_fixed {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #20995, opened 2024-03-11
fmt: update tests to run all fmt also in a vmodules context

ttytm:
The PR updates v fmt tests to additionally run as if they were running for code in `$VMODULES`. This expands the tested scenarios for vmodules and makes it easier to add new tests for future fixes, as it follows the current schema that we use to test formatting.

The change makes the current separate tests for vmodules that I added in #20977 obsolete. The tests from this PR here covers the fix.


vlib/v/fmt/fmt_keep_test.v
```diff
@@ -9,43 +9,42 @@ import v.parser
...
-	basepath := vroot + '/'
-	tmpfolder := os.temp_dir()
-	diff_cmd := diff.find_working_diff_command() or { '' }
-	mut fmt_bench := benchmark.new_benchmark()
-	keep_input_files := os.walk_ext('vlib/v/fmt/tests', '_keep.vv')
-	expected_input_files := os.walk_ext('vlib/v/fmt/tests', '_expected.vv')
 	mut input_files := []string{}
-	input_files << keep_input_files
-	input_files << expected_input_files
-	input_files = vtest.filter_vtest_only(input_files, basepath: vroot)
```
spytheman, line 39: Removing vtest.filter_vtest_only here, loses functionality.
Compare `VTEST_ONLY=anon_struct ./v vlib/v/fmt/fmt_keep_test.v` on master and here.

ttytm: Thanks for the review! Gonna make sure this works as it should.

vlib/v/fmt/fmt_keep_test.v
```diff
@@ -9,43 +9,42 @@ import v.parser
...
 const fpref = &pref.Preferences{
 	is_fmt: true
 }
-const vexe = os.getenv('VEXE')
+
+fn get_vroot() string {
+	vexe := pref.vexe_path()
+	if vexe == '' || !os.exists(vexe) {
+		eprintln('VEXE must be set')
+		exit(ecode_missing_vexe)
```
ttytm, line 26: can the exits be brought down to asserts? 
```
assert false, 'VEXE must be set'
```

JalonSolov: From the docs:

"All `assert` statements are removed, when you compile your program with the `-prod` flag."

So, it wouldn't work if someone compiled vfmt with `-prod` - which I can assure you, some do.

spytheman: The code in question is in a test file, not in vfmt itself.

spytheman: > can the exits be brought down to asserts?
> 
> ```
> assert false, 'VEXE must be set'
> ```

yes

spytheman: The idea of the separate exit codes at the time, was to signal information to the test runner, so that it could show `SKIP` for some tests, that detected themselves, that they are run in an unsuitable environment (like lacking running mysql or postgres, or working network, or an `expect` binary etc), but it never materialized, and in this specific case of fmt_keep_test.v, I do not see now value from having separate exit codes, while simplifying it will make the code a bit more easier to maintain.

vlib/v/fmt/fmt_keep_test.v
```diff
@@ -72,6 +73,12 @@ fn test_fmt() {
 	eprintln(term.h_divider('-'))
 	eprintln(fmt_bench.total_message(fmt_message))
 	if fmt_bench.nfail > 0 {
-		exit(error_failed_tests)
+		exit(ecode_failed_tests)
 	}
```
spytheman, line 77: That can be turned into an assert too.

ttytm:
Further testing and cleaning up my local test files I realized that vmodules specific test cases will be required for full coverage.

I'm adding the required test and to make the changes reliable and give easy acces to the tested fix, 6b61bf932a4d66d3ebec1ec86fefd884baf5051f temporary removes the recent fix. Checking out the commit, there should be a failing test when running
```
v vlib/v/fmt/fmt_keep_test.v
```

At the final state all tests should pass.

vlib/v/fmt/fmt_keep_test.v
ttytm: Nice, thanks for the insights!

vlib/v/fmt/fmt_keep_test.v
```diff
@@ -11,67 +11,77 @@ import v.pref
 import v.util.diff
 import v.util.vtest
 
-const error_missing_vexe = 1
-const error_failed_tests = 2
+const vroot = @VEXEROOT
```
ttytm, line 14: As far as I can see, having checked that the AtExpr uses `os.dir(pref.vexe_path())` under the hood, this shouldn't cause problems :thinking:. The tests would now assert that input_files are present as well.

PR #20996, opened 2024-03-11
doc: fix multiline codeblock comments

### LINKED ISSUE #20921: documentation generator seems to be off
### Describe the bug

image

used to produce a code block but now we get 

image

unrelated we also have some double entries

image

### Reproduction Steps

see https://github.com/freeflowuniverse/crystallib

there is a doc.sh script to generate

### Expected Behavior

code block in the docu

### Current Behavior

see above

### Possible Solution

_No response_

_No response_

### V version

newest

m.1 osx
> [!NOTE]
>

ttytm:
Fixes #20921

```v
// foo.v
module foo

// foo does stuff
// ```
// this is a multiline codeblock.
// second line
// third line.
// ```
pub fn foo() { }
```

```sh
v self // after checkout
v doc -comments foo.v
```

- Current
  ```sh
  module foo
      foo.v
  
  fn foo()
      foo does stuff
      ```
      this is a multiline codeblock. second line third line.
      ```
  ```

- Change
  ```sh
  module foo
      foo.v
  
  fn foo()
      foo does stuff
      ```
      this is a multiline codeblock.
      second line
      third line.
      ```
  ```
  
  image


spytheman (approved):
Excellent work.

Issue #20975, opened 2024-03-08
StringReader using io module private constants


daddycocoaman:
### Describe the bug

Hi. I'm attempting to use the new `string_reader` module but it looks like it's trying to use private constants from `io`. Would it be a good idea to make those constants public?

### Reproduction Steps

```v
module main

import os
import io
import io.string_reader

pub fn file_magic(filename string) !string {
	file_data := os.open(filename)!
	mut reader := string_reader.StringReader.new(reader: file_data)
	magic := reader.read_string(3)!
	return magic
}

fn main() {
	println(file_magic("v.mod")!)
}
```

### Expected Behavior

Read 3 bytes from the file (should output "Mod" in this example)

### Current Behavior

```
C:/Users/daddycocoaman/dev/_langs/v/vlib/io/string_reader/string_reader.v:71:33: error: constant `io.read_all_len` is private
   69 |
   70 |     // make sure there is enough room in the string builder
   71 |     unsafe { r.builder.grow_len(io.read_all_len) }
      |                                    ~~~~~~~~~~~~
   72 |     defer {
   73 |         // shrink the length of the buffer to the total of bytes read
C:/Users/daddycocoaman/dev/_langs/v/vlib/io/string_reader/string_reader.v:87:35: error: constant `io.read_all_grow_len` is private
   85 |             break
   86 |         } else if r.builder.len == end {
   87 |             unsafe { r.builder.grow_len(io.read_all_grow_len) }
      |                                            ~~~~~~~~~~~~~~~~~
   88 |         }
   89 |     }
C:/Users/daddycocoaman/dev/_langs/v/vlib/io/string_reader/string_reader.v:109:12: error: constant `io.read_all_len` is private
  107 |     start := r.builder.len
  108 |     // make sure there is enough room in the string builder
  109 |     if n > io.read_all_len {
      |               ~~~~~~~~~~~~
  110 |         unsafe { r.builder.grow_len(io.read_all_len) }
  111 |     } else {
C:/Users/daddycocoaman/dev/_langs/v/vlib/io/string_reader/string_reader.v:110:34: error: constant `io.read_all_len` is private
  108 |     // make sure there is enough room in the string builder
  109 |     if n > io.read_all_len {
  110 |         unsafe { r.builder.grow_len(io.read_all_len) }
      |                                        ~~~~~~~~~~~~
  111 |     } else {
  112 |         unsafe { r.builder.grow_len(n) }
[...]

PR #20997, opened 2024-03-12
io.reader: Make read_all constants public

daddycocoaman:
The `io.string_reader.StringReader` class uses private constants from `io.string_reader`. This PR makes those constants public so that the newly added StringReader class (and any future reader-like classes) can use them. Addresses #20975.

### DIFF
--- a/vlib/io/reader.v
+++ b/vlib/io/reader.v
@@ -29,8 +29,8 @@ mut:
 	read(mut buf []u8) !int
 }
 
-const read_all_len = 10 * 1024
-const read_all_grow_len = 1024
+pub const read_all_len = 10 * 1024
+pub const read_all_grow_len = 1024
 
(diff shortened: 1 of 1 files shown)


spytheman:
Hm, that is interesting indeed 🤔 . Good find.

The following, when saved as `a_test.v` compiles cleanly (but it should not, since the consts are private, as you noticed).

At the same time, the *same source*, when saved as `a.v`, produces errors.
```v
import io.string_reader
import os

fn file_magic(filename string) !string {
        file_data := os.open(filename)!
        mut reader := string_reader.StringReader.new(reader: file_data)
        magic := reader.read_string(3)!
        return magic
}

fn test_main() ! {
        println( file_magic("v.mod")! )
}

fn main() {
        test_main()!
}
```

spytheman (approved):

Issue #20998, opened 2024-03-12
StringReader off-by-one panic


daddycocoaman:
### Describe the bug

When attempting to read one byte at the end of the buffer with StringReader, V will panic. This is because the `StringReader.needs_fill_until` and `StringReader.needs_fill` will return true if the requested length is **_equal to_** or greater than the length of the buffer. The buffer should only be filled if the requested length is greater than the length of the buffer, not equal to. 

### Reproduction Steps

```v
import io.string_reader

fn main() {

    source := "ABCD"
	mut reader := string_reader.StringReader.new(source: source)

	println(reader.read_all_bytes(true)!)

	reader = string_reader.StringReader.new(source: source)
        println(reader.read_bytes(1)!)
	println(reader.read_bytes(1)!)
	println(reader.read_bytes(1)!)
	println(reader.read_bytes(1)!)
}
```

### Expected Behavior

StringReader will read the last byte of the buffer if the length to read requested is equal to the end of the buffer.

### Current Behavior

```
[65, 66, 67, 68]
[65]
[66]
[67]
V panic: result not set (reader is not set)
v hash: a867ed6
print_backtrace_skipping_top_frames is not implemented
```

### Possible Solution

I actually don't understand why the reader object disappears in the `fill_buffer` functions. Even if the buffer is at the end of the stream and too many bytes are requested, shouldn't it still be there? At any rate, the `needs_fill` functions should probably be changed to check for only greater than, not greater than or equal to.

### V version

`V 0.4.4 f789874.a867ed6`

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.4 f789874.a867ed6
OS: windows, Microsoft Windows 11 Enterprise v26058 ARM 64-bit Processor
Processor: 8 cpus, 64bit, little endian, 
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #21005, opened 2024-03-12
fix needs_fill_until check

daddycocoaman:
This PR is meant to address https://github.com/vlang/v/issues/20998. 

It changes the condition so that if you request the last byte individually, then StringReader won't fail when the current offset is equal to the length of the builder before it increases the offset to EOF.


daddycocoaman:
In my mind, the builder length should only be increased if

- the offset is at EOF and you need to get space from that point
- the offset is before EOF and you need get additional space past EOF

Maybe @Casper64 can add input as module creator.

spytheman (approved):
For me, the change is logical. @Casper64 what do you think?

Casper64:
Yeah this is better

Casper64 (approved):

PR #21007, opened 2024-03-12
checker: remove unnecessary string interpolation in deprecation method calls

ttytm:


spytheman (approved):

spytheman:
It may be a good idea to make `v vet` detect such places.

PR #21008, opened 2024-03-13
db.sqlite: Fix exec_param_many bug

jacksonmowry:
Fixes a bug where sqlite errors would not be caught through `exec_param_many` and it's associated methods.
Previously runtime failures like unique constraints and others would not be caught when using `exec_param_many`.


spytheman (approved):
Excellent work.

PR #21011, opened 2024-03-13
 checker: disallow void return value lambdas for map 

Delta456:
Fixes https://github.com/vlang/v/issues/21006


vlib/v/checker/fn.v
```diff
@@ -2739,6 +2739,14 @@ fn (mut c Checker) check_map_and_filter(is_map bool, elem_typ ast.Type, node ast
...
 			}
 		}
+		ast.LambdaExpr {
+			if arg_expr.expr is ast.CallExpr {
+				if is_map && arg_expr.expr.return_type in [ast.void_type, 0] {
+					c.error('type mismatch, `${arg_expr.expr.name}` does not return anything',
+						arg_expr.expr.pos)
+				}
+			}
+		}
```
ttytm, line 2749: Good fix :+1: 

As it is the if statement can be merged to reduce a nesting level and make immediately visible that it's just a single block here
```suggestion
			if arg_expr.expr is ast.CallExpr && is_map
				&& arg_expr.expr.return_type in [ast.void_type, 0] {
				c.error('type mismatch, `${arg_expr.expr.name}` does not return anything',
					arg_expr.expr.pos)
			}
```

Delta456: Indeed a good suggestion

PR #21014, opened 2024-03-13
ci,tests: enable vdoc tests on linux; fix unintended line break

ttytm:
The PR enables vdoc tests.

vdoc is currently not test covered as it is skipped on all platforms. 

Enabling the tests on a working Linux platform - also what we use to deploy V docs - reveals that there currently is one failing test to fix.

- [x] fix tests

### DIFF
--- a/cmd/tools/vdoc/tests/testdata/newlines/main.comments.out
+++ b/cmd/tools/vdoc/tests/testdata/newlines/main.comments.out
@@ -14,7 +14,8 @@ fn funky()
     - bar
     # test
     ########### deep test
-    #a shouldn't have a newline test
+    #a shouldn't have a newline
+    test
     
     | foo bar   |  yes   |
     |-----------|--------|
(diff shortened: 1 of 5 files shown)


cmd/tools/vdoc/tests/testdata/newlines/main.comments.out
```diff
@@ -14,7 +14,8 @@ fn funky()
     - bar
     # test
     ########### deep test
-    #a shouldn't have a newline test
+    #a shouldn't have a newline
+    test
```
ttytm, line 18: Used suboptimal words in the todo with `fix tests`. 
Because it is not the test that needs to be corrected, but the code so that the test passes again.

Afaik the `.out` was correct and as the text says the word should not be on a new line.
If updating the test is fine for this PR, the behavior with line breaks should be investigated separately.

spytheman (approved):

Issue #17311, opened 2023-02-14
We are allowed to create structs time.Time 


brandonpille:
### Describe the bug

It is possible to create time.Time objects which might introduce problems when using the object. For example the code below does not do what you want it to do:
```
now := time.now()
today := time.Time { year: now.year, month: now.month, day: now.day }
today_5am := today.add(time.hour*5)
```
You expect the object today_5am to be today 5 AM but it is actually 1970-01-01 05:00:00. That is because the unix property was not filed in while creating the object. 
Suggestion: don't allow the construction of the object outside of the module

### Expected Behavior

See description

### Current Behavior

See description

### Reproduction Steps

See description

### Possible Solution

See description

### V version

V 0.3.2

### Environment details (OS name and version, etc.)

OS: linux, Ubuntu 22.04.1 LTS
Processor: 16 cpus, 64bit, little endian, 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
CC version: cc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0

getwd: /work/farmerbot
vmodules: /home/brandon/.vmodules
vroot: /work/v
vexe: /work/v/v
vexe mtime: 2023-01-18 13:03:15
is vroot writable: true
is vmodules writable: true
V full version: V 0.3.2 f0a252d.2c78078

Git version: git version 2.34.1
Git vroot status: weekly.2023.03-15-g2c780788 (152 commit(s) behind V master)
.git/config present: true
thirdparty/tcc status: thirdparty-linux-amd64 12f392c3

JalonSolov:
Easily accomplished by adding the `[required]` attribute to the Time fields.

medvednikov:
@JalonSolov there's `[noinit] struct Time {}` for that

JalonSolov:
Or that.  :-)

paulyelesin:
I'll pick it up

visrut-at-incubyte:
We can add `[noinit]` before the struct, but it could introduce breaking changes and I can see there are some tests that are failing.

- If the user is intending the behavior like `today:= time.Time { year: now.year, month: now.month, day: now.day }`, they are assuming in variable `today`, `hours` are also being automatically set.
- We can implement this behavior but, I don't know how to run some operation after the user initiates the object, something like a constructor function where we can set value for `hour`

Issue #17162, opened 2023-01-29
time.add_days(1) jumps from 2013-05-01 to 1970-01-02


gmlewis:
### Describe the bug

Using `V 0.3.2 9a86456`, I wrote the following program:

```v
module main

import time

fn main() {
	t := time.Time{
		year: 2013
		month: 5
		day: 1
	}
	println('BEFORE: ${t} is a ${t.long_weekday_str()}: ${t.debug()}')
	t2 := t.add_days(1)
	println('AFTER: ${t2} is a ${t2.long_weekday_str()}: ${t2.debug()}')
}
```

and ran it on my Linux Mint Cinnamon 20.3 box:

```bash
v version && v run bug.v
V 0.3.2 9a86456
BEFORE: 2013-05-01 00:00:00 is a Wednesday: Time{ year: 2013 month: 05 day: 01 hour: 00 minute: 00 second: 00 microsecond: 000000 unix: 0000000 }
AFTER: 1970-01-02 00:00:00 is a Friday: Time{ year: 1970 month: 01 day: 02 hour: 00 minute: 00 second: 00 microsecond: 000000 unix: 0086400 }
```

### Expected Behavior

I expected to see `2013-05-02`.

### Current Behavior

I got: `1970-01-02`.

### Reproduction Steps

```v
module main

import time

fn main() {
	t := time.Time{
		year: 2013
		month: 5
		day: 1
	}
	println('BEFORE: ${t} is a ${t.long_weekday_str()}: ${t.debug()}')
	t2 := t.add_days(1)
	println('AFTER: ${t2} is a ${t2.long_weekday_str()}: ${t2.debug()}')
}
```

### V version

V 0.3.2 9a86456

### Environment details (OS name and version, etc.)

```
OS: linux, Linux Mint 20.3
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
CC version: cc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0

getwd: /home/glenn/exercism/vlang/meetup
vmodules: /home/glenn/.vmodules
vroot: /home/glenn/src/github.com/vlang/v
vexe: /home/glenn/src/github.com/vlang/v/v
vexe mtime: 2023-01-29 02:00:22
is vroot writable: true
is vmodules writable: true
V full version: V 0.3.2 148d578.9a86456

Git version: git version 2.25.1
[...]

PR #21017, opened 2024-03-13
checker: add deprecation messages for structs and struct fields

ttytm:
Wants to add the ability to inform about deprecation of structs and struct fields when initiating them.

Screenshot_20240313_213109

Then trying to init elements with the deprecated attribute would look something like 

Screenshot_20240313_233134

The change might help with informing about moving to `@[noinit]` structs for #17162 #17311. Else for other deprecation use cases.


spytheman (approved):
Excellent work.

PR #21022, opened 2024-03-14
parser: allow lock prefix instructions and numbered reg in asm inline

felipensp:
Allow asm inline cases like:

```c
__asm__ volatile("lock orl $0, (%esp)");
__asm__ volatile("lock orq $0, (%rsp)");
```

```v
	$if i386 {
		asm i386 {
			lock orl '(%esp)', 0
		}
	} $else $if amd64 {
		asm amd64 {
			lock orq '(%rsp)', 0
		}
        }
```

```v
		asm amd64 {
			lock cmpxchgb '%1', '%2'
			; =a (rv)
			+m (*&u8(atom))
			; q (xchg)
			0 (cmp)
			; memory
		}
```


spytheman:
Please add tests in `vlib/v/slow_tests/assembly/asm_test.amd64.v`.

vlib/v/parser/parser.v
```diff
@@ -1253,6 +1253,11 @@ fn (mut p Parser) asm_stmt(is_top_level bool) ast.AsmStmt {
 		}
 		if p.tok.kind in [.key_in, .key_lock, .key_orelse, .key_select, .key_return] { // `in`, `lock`, `or`, `select`, `return` are v keywords that are also x86/arm/riscv/wasm instructions.
 			name += p.tok.kind.str()
+			if p.tok.kind == .key_lock && arch in [.i386, .amd64] {
+				p.next()
+				name += ' '
+				name += p.tok.lit
+			}
```
spytheman, line 1260: According to https://www.felixcloutier.com/x86/lock:
> The LOCK prefix can be prepended only to the following instructions and only to those forms of the instructions where the destination operand is a memory operand: ADD, ADC, AND, BTC, BTR, BTS, CMPXCHG, CMPXCH8B, CMPXCHG16B, DEC, INC, NEG, NOT, OR, SBB, SUB, XOR, XADD, and XCHG.

What will check that? The C compiler?

vlib/v/parser/parser.v
```diff
@@ -1764,7 +1769,11 @@ fn (mut p Parser) asm_ios(output bool) []ast.AsmIO {
 			if p.tok.kind == .at {
 				p.next()
 			} else {
-				p.check(.name)
+				if p.tok.kind == .number {
+					p.check(.number)
+				} else {
+					p.check(.name)
+				}
```
spytheman, line 1776: Why? What syntax is needed to trigger the number check?

spytheman:
This sample:
```v
fn main() {
        asm amd64 {
                lock orq (%rsp), 0
        }
}
```
does not compile cleanly on both `master`, and with this PR.
```
#0 09:22:35 ᛋ fix_asm_inline /v/oo❱git co master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
#0 09:22:36 ᛋ master /v/oo❱./v self
V self compiling ...
#0 09:22:40 ᛋ master /v/oo❱cat a.v
fn main() {
        asm amd64 {
                lock orq (%rsp), 0
        }
}
#0 09:22:44 ᛋ master /v/oo❱./v a.v
a.v:3:12: error: unexpected token `(`, expecting name
    1 | fn main() {
    2 |     asm amd64 {
    3 |         lock orq (%rsp), 0
      |                  ^
    4 |     }
    5 | }
#1 09:22:46 ᛋ master /v/oo❱git switch -
Switched to branch 'fix_asm_inline'
#0 09:22:49 ᛋ fix_asm_inline /v/oo❱./v self
V self compiling ...
#0 09:22:53 ᛋ fix_asm_inline /v/oo❱./v a.v
a.v:3:12: error: invalid token in assembly block
    1 | fn main() {
    2 |     asm amd64 {
    3 |         lock orq (%rsp), 0
      |                  ^
    4 |     }
    5 | }
#1 09:22:56 ᛋ fix_asm_inline /v/oo❱
```

Only the parser error is different with this PR.

spytheman:
This compiles cleanly with this PR, but cgen errors on master:
```v
fn main() {
        asm amd64 {
                lock inc eax
        }
}
```

spytheman:
So does this: `asm amd64 { lock orq rsp, 0 }`

felipensp:
> This is a parser error on master, but it is fine here: `asm amd64 { lock orq rsp, 0 }`.
> 
> Later with tcc, it generates code that leads to a runtime error:
> 
> ```
> /tmp/v_1000/x.01HS0GH0PXCQ6FMGR3QCGPNPK7.tmp.c:12951: at main__main: RUNTIME ERROR: illegal instruction
> ```
> 
> With gcc, it is a C compiler error:
> 
> ```
> /tmp/v_1000/x.01HS0GHPANSP1XEQER4TFYXJDY.tmp.c:12976: Error: expecting lockable instruction after `lock'
> ```
> 
> since it generated:
> 
> ```c
> void main__main(void) {
>     __asm__ (
>         "lock orq $0, %rsp;"
>     );
> }
> ```
> 
> and apparently gcc does check the inline assembly.

Oh sorry. I missed to update the PR description. vfmt just showed me the proper syntax for the (%reg) stuff, so I have declined some changes on the PR and so I focused on the numbered constraint and lock-prefix.

The correct test case for the PR is:

```v
	$if amd64 {
		asm amd64 {
			lock cmpxchgb '%1', '%2'
			; =a (rv)
			+m (*&u8(atom))
			; q (xchg)
			0 (cmp)
			; memory
		}
	}
```

felipensp:
And:

```v
	$if i386 {
		asm i386 {
			lock orl '(%esp)', 0
		}
	} $else $if amd64 {
		asm amd64 {
			lock orq '(%rsp)', 0
		}
        }
```

vlib/v/parser/parser.v
felipensp: > According to https://www.felixcloutier.com/x86/lock:
> 
> > The LOCK prefix can be prepended only to the following instructions and only to those forms of the instructions where the destination operand is a memory operand: ADD, ADC, AND, BTC, BTR, BTS, CMPXCHG, CMPXCH8B, CMPXCHG16B, DEC, INC, NEG, NOT, OR, SBB, SUB, XOR, XADD, and XCHG.
> 
> What will check that? The C compiler?

I think so. But we can add a startswith checking though.

vlib/v/parser/parser.v
felipensp: It is about the constraint part.

image

vlib/v/parser/parser.v
```diff
@@ -1253,6 +1253,19 @@ fn (mut p Parser) asm_stmt(is_top_level bool) ast.AsmStmt {
 		}
 		if p.tok.kind in [.key_in, .key_lock, .key_orelse, .key_select, .key_return] { // `in`, `lock`, `or`, `select`, `return` are v keywords that are also x86/arm/riscv/wasm instructions.
 			name += p.tok.kind.str()
+			if p.tok.kind == .key_lock && arch in [.i386, .amd64] {
+				p.next()
+				allowed_ins := ['add', 'adc', 'and', 'btc', 'btr', 'bts', 'cmpxchg', 'cmpxch8b',
+					'cmpxchg16b', 'dec', 'inc', 'neg', 'not', 'or', 'sbb', 'sub', 'xor', 'xadd',
+					'xchg']
```
Delta456, line 1260: Can this be a const instead? I think this may be used in future as well.

spytheman: Yes, it would be a bit nicer as a constant, for code that uses lots of inline assembly, it will avoid some allocations. On the other hand, such code is rare, while a constant will be initialized every time the compiler is ran 🤔 .

Still, I think that the benefit of not doing allocations each time is greater.

vlib/v/parser/parser.v
```diff
@@ -1764,7 +1777,14 @@ fn (mut p Parser) asm_ios(output bool) []ast.AsmIO {
 			if p.tok.kind == .at {
 				p.next()
 			} else {
-				p.check(.name)
+				if p.tok.kind == .number {
```
spytheman, line 1780: Please add a comment with a link to the docs you mentioned in the screenshot, or with a small example. It will really help with the maintenance of the parser.

felipensp: Done.

spytheman (approved):
Excellent work.

PR #21023, opened 2024-03-14
os: fix Process struct fields documentation applied to wrong field

bramvbilsen:
The fields of the `Process` struct in the `os` modules incorrectly have their documentation behind the field, instead of before. Example of what it currently looks like:
```V
pub struct Process {
pub mut:
        // ...
	pid      int    // the PID of the process
	code     int = -1
	// the exit code of the process, != -1 *only* when status is .exited *and* the process was not aborted
	// ...
}
```
This results in `code` incorrectly having the documentation associated with `pid`. This specific case is confusing in particular as both are integers and `code` could be mistaken for the `pid` after reading the associated comments.

### DIFF
--- a/vlib/os/process.v
+++ b/vlib/os/process.v
@@ -18,34 +18,20 @@ pub enum ProcessState {
 @[heap]
 pub struct Process {
 pub mut:
-	 // the process's command file path
-	filename string
-	// the PID of the process
-	pid      int
-	// the exit code of the process, != -1 *only* when status is .exited *and* the process was not aborted
-	code     int = -1
-	// the current status of the process
-	status ProcessState = .not_started
-	// if the process fails, contains the reason why
-	err              string
-	// the arguments that the command takes
-	args             []string
-	// the initial working folder of the process. When '', reuse the same folder as the parent process.
-	work_folder      string
-	// true, when the environment was customized with .set_environment
-	env_is_custom    bool
-	// the environment with which the process was started  (list of 'var=val')
-	env              []string
-	// when true, then you can use p.stdin_write(), p.stdout_slurp() and p.stderr_slurp()
-	use_stdio_ctl    bool
-	// when true, the process will create a new process group, enabling .signal_pgkill()
-	use_pgroup       bool
-	// the stdio file descriptors for the child process, used only by the nix implementation
-	stdio_fd         [3]int
-	// the WProcess; used only by the windows implementation
-	wdata            voidptr
-	// sets a value indicating whether to start the process in a new window, The default is false; used only by the windows implementation
-	create_no_window bool
+	filename         string       // the process's command file path
+	pid              int          // the PID of the process
+	code             int = -1 // the exit code of the process, != -1 *only* when status is .exited *and* the process was not aborted
+	status           ProcessState = .not_started // the current status of the process
+	err              string       // if the process fails, contains the reason why
+	args             []string     // the arguments that the command takes
+	work_folder      string       // the initial working folder of the process. When '', reuse the same folder as the parent process.
+	env_is_custom    bool     // true, when the environment was customized with .set_environment
+	env              []string // the environment with which the process was started  (list of 'var=val')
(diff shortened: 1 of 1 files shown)


vlib/os/process.v
```diff
@@ -18,22 +18,34 @@ pub enum ProcessState {
...
+	// when true, then you can use p.stdin_write(), p.stdout_slurp() and p.stderr_slurp()
+	use_stdio_ctl    bool
+	// when true, the process will create a new process group, enabling .signal_pgkill()
+	use_pgroup       bool
+	// the stdio file descriptors for the child process, used only by the nix implementation
+	stdio_fd         [3]int
+	// the WProcess; used only by the windows implementation
+	wdata            voidptr
+	// sets a value indicating whether to start the process in a new window, The default is false; used only by the windows implementation
+	create_no_window bool
```
ttytm, line 48: Good catch, thanks for the contribution :blue_heart:.

Personally, I think keeping the comments at the end of the line is cleaner and in terms of the code better readable.

Moving the fields with the default fields down would result in something like:

```v
pub struct Process {
pub mut:
	filename         string       // the process's command file path
	pid              int          // the PID of the process
	err              string       // if the process fails, contains the reason why
	args             []string     // the arguments that the command takes
	work_folder      string       // the initial working folder of the process. When '', reuse the same folder as the parent process.
	env_is_custom    bool         // true, when the environment was customized with .set_environment
	env              []string     // the environment with which the process was started  (list of 'var=val')
	use_stdio_ctl    bool         // when true, then you can use p.stdin_write(), p.stdout_slurp() and p.stderr_slurp()
	use_pgroup       bool         // when true, the process will create a new process group, enabling .signal_pgkill()
	stdio_fd         [3]int       // the stdio file descriptors for the child process, used only by the nix implementation
	wdata            voidptr      // the WProcess; used only by the windows implementation
	create_no_window bool         // sets a value indicating whether to start the process in a new window, The default is false; used only by the windows implementation
	status           ProcessState = .not_started // the current status of the process
	code             int = -1 // the exit code of the process, != -1 *only* when status is .exited *and* the process was not aborted
}

```

Screenshot_20240314_184252


Though this is opinionated and just a suggestion I'm throwing in here.

spytheman: I prefer keeping the comments on the same line as well.

The other style (comment before field), should be reserved only for extremely long lines/explanations.

spytheman: In this particular case, keeping all the comments on the same line will result in a smaller diff too.

spytheman: First edit the comments as shown above, then `v fmt -w vlib/os/process.v` and commit the result.

spytheman: Then, when the PR is merged, the documentation on <https://modules.vlang.io/os.html> will be regenerated automatically.

spytheman: You can test locally with: `./v doc -m -f html vlib/` then open the generated `vlib/_docs/os.html` in your browser.

bramvbilsen: I added the requested changes to make field defaults cleaner.
I also moved the comments back to next of the fields. Thus this will again result in the comments being associated with the wrong field. But if from a style pov, we would like to keep it this way, we will have to change the functionality related to associating comments/docs with struct fields.

Let me know if there's anything else!

bramvbilsen: Or should I have left the comments above the fields (as initially proposed) until a fix for comments next to fields is pushed out?

spytheman: > Thus this will again result in the comments being associated with the wrong field.

They should not. If they are, then that is a bug, potentially in `v-analyzer`, which is a separate project.

spytheman: > Or should I have left the comments above the fields (as initially proposed) until a fix for comments next to fields is pushed out?

No, the current state of the PR is fine.

bramvbilsen: @ttytm That's exactly the issue I was encountering. The dialog that opens up when hovering over the field.

PR #21029, opened 2024-03-15
scanner: disallow shebang outside of *.vsh files

ttytm:

### DIFF
--- a/vlib/v/scanner/scanner.v
+++ b/vlib/v/scanner/scanner.v
@@ -959,14 +959,15 @@ pub fn (mut s Scanner) text_scan() token.Token {
 				if nextc == `!` {
 					// treat shebang line (#!) as a comment
 					comment := s.text[start - 1..s.pos].trim_space()
-					if !s.file_base.ends_with('.vsh') {
+					if s.line_nr != 1 {
 						comment_pos := token.Pos{
 							line_nr: s.line_nr - 1
-							len: comment.len + 2
+							len: comment.len
 							pos: start
-							col: s.current_column() - comment.len - 2
+							col: s.current_column() - comment.len
 						}
-						s.error_with_pos('a shebang is only valid in *.vsh files', comment_pos)
+						s.error_with_pos('a shebang is only valid at the top of the file',
+							comment_pos)
 					}
 					// s.fgenln('// shebang line "$s.line_comment"')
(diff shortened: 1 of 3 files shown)


spytheman:
why?

ttytm:
Good question. Could be better to keep it valid it for .v files and treat it as a shebang is in script files. Recognized as shebang when it's at the beginning of the file.

Currently this compiles. Which got me into this PR.
```v
module main

type Bar = string

fn main() {
	#!/usr/bin/env -S v
	dump('abc')
}
```

ttytm:
In case this shouldn't be a valid comment.

spytheman:
I think it should be valid, only on the first line, no matter the extension.

spytheman:
Note that you can:
```v
#!/usr/bin/env -S v -raw-vsh-tmp-prefix tmp run
module main

type Bar = string

fn main() {
        dump('abc')
}
```
Save this file as just `abc` (without any extension), then do `chmod 755 abc`, and then do: `./abc`

spytheman:
You can also skip the `run` in the first line, in which case, v will generate a `tmp.abc` executable just once, and doing `./abc` after that, will just start `./tmp.abc` directly, without having to compile it again, which will be much faster.
That is a documented feature, see https://github.com/vlang/v/blob/master/doc/docs.md#vsh-scripts-with-no-extension, and making the scanner depend on the extension will break it.

ttytm:
With the new changes a shebang that's not on the first line is disallowed.

Shebangs are not recognized when it has empty lines beforehand.

`script.sh`: won't use cat if there's an empty line beforehand:
```

#!/bin/cat

echo 'hello'
```

`script.vsh`: with an empty line
```

#!/usr/bin/env -S v

println('hello world')
```

Currently results in:
```
./test.vsh: line 4: syntax error near unexpected token `'hello world''
./test.vsh: line 4: `println('hello world')'
```

With the changes it would be:
```
bugs/shebang.v:3:1: error: a shebang is only valid at the top of the file
    1 | 
    2 | 
    3 | #!/usr/bin/env -S v
      | ^
    4 | 
    5 | println('hello')
```

The new error will stop `v fmt` to remove the empty lines beforehand

spytheman (approved):
Excellent work.

PR #21035, opened 2024-03-15
cgen: add possibility to have a global hidden

felipensp:
Makes possible to declare:

```v
@[hidden]
__global __dso_handle = 0
```


spytheman (approved):
Good work, the CI failure is unrelated.

PR #21037, opened 2024-03-15
rand: add PRNG.fill_buffer_from_set/2

spytheman:


vlib/rand/rand.c.v
```diff
@@ -93,14 +93,15 @@ fn internal_ulid_at_millisecond(mut rng PRNG, unix_time_milli u64) string {
...
 
+@[direct_array_access]
 fn internal_string_from_set(mut rng PRNG, charset string, len int) string {
 	if len == 0 {
 		return ''
 	}
 	mut buf := unsafe { malloc_noscan(len + 1) }
 	for i in 0 .. len {
 		unsafe {
-			buf[i] = charset[intn(charset.len) or { 0 }]
```
spytheman, line 103: Calling intn() here is a bug, since intn() will use the default rng, not the passed one.

PR #21042, opened 2024-03-16
cgen: fix _str name generated for C struct defined str() method

### LINKED ISSUE #21015: cgen error for `pub fn (d &C.Display) str() string {`, where Display is from X11/Xlib.h (incomplete struct)
**V doctor:**
```
V full version: V 0.4.4 f789874.74b8c9d
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-03-13 11:19:25

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.11-19-gc29b64e0
.git/config present: true

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg /v/misc/2024_03_13__18/c.v`
```v
#include <X11/Xlib.h>

@[typedef]
pub struct C.Display {}

pub fn (d &C.Display) str() string {
	return 'C.Display{}'
}

fn main() {
	x := unsafe { &C.Display(1) }
	println(x)
}

```

compiled program with no cgen errors

**What did you see instead?**
```
==================
/tmp/v_1000/c.01HRWD2K94MQ3JFTGA0B33M979.tmp.c:12987: warning: implicit declaration of function 'Display_str'
/tmp/v_1000/c.01HRWD2K94MQ3JFTGA0B33M979.tmp.c:12987: error: type mismatch in conditional expression (have 'struct string' and 'int')
...
==================

```

Note: this should be tested on Linux/FreeBSD, but I think it would affect other cases, where on the C side, we have structs with incomplete definitions in the header files.
> [!NOTE]
>

felipensp:
Fix #21015

### DIFF
--- a/vlib/v/gen/c/auto_str_methods.v
+++ b/vlib/v/gen/c/auto_str_methods.v
@@ -83,6 +83,9 @@ fn (mut g Gen) get_str_fn(typ ast.Type) string {
 			else {}
 		}
 	}
+	if sym.language == .c && !typ.has_flag(.option) && sym.has_method('str') {
+		str_fn_name = util.no_dots(g.cc_type(unwrapped, false)) + '_str'
+	}
 	g.str_types << StrType{
 		typ: unwrapped
 		styp: styp


spytheman:
We have to think of a way to test it 🤔. Depending on <X11/Xlib.h> (as in the issue) for it, is not nice, since it will fail on platforms where those headers are not installed.

spytheman:
I think the relevant part, is something like this:
```v
struct _XDisplay;
typedef struct _XDisplay Display;
```

felipensp:
> I think the relevant C part, is something like this:
> 
> ```c
> struct _XDisplay;
> typedef struct _XDisplay Display;
> ```

Test added. 👍🏻

spytheman:
Your test does not test the typedef over an incomplete struct type.

felipensp:
> Your test does not test the typedef over an incomplete struct type.

But It seems it does not need to be a incomplete type, I got the same error with this test case.

spytheman:
Good point. I checked more thoroughly, and it is indeed so 🙇🏻‍♂️ - the PR fixes #21015 as well, even though the test is not for the exact situation.

spytheman (approved):
Excellent work.

Issue #21041, opened 2024-03-16
C builder error when omitting error handling in use of Result type function in the range expression of a for-statement


thomashansen123:
### Describe the bug

Hello.

As the title says, omitting the ! operator or an 'or panic{}' block when using a function like time.days_in_month() !int or arrays.idx_min() !int, when used in the range part of a for statement, results in a builder error.

### Reproduction Steps

Example:
```
module main

import arrays

fn main() {
	arr := [1, 2, 3, 4, 5, 6, 7, 8, 9]
	for _ in 0 .. arrays.idx_min(arr) {
	}
}
```
Output (from V playground):
```
/tmp/v_60000/../../../../../../box/code.v:7: error: '{' expected (got ";")
builder error: 
==================
C error. This should never happen.
```
The offending line here seems to be
```
	for _ in 0 .. arrays.idx_min(arr) {
```
### Expected Behavior

V compiler should catch the omission of error handling when using a function that returns a Result type in a for statement. 

### Current Behavior

C builder error, see above

### Additional Information/Context

This seems to be only an issue with functions that return !int.
```
module main


fn main() {
	s := 'this is a string'
	for _ in s.substr_with_check(2, 5) {
	}
}
```
Lack of error handling here when using !string in a for statement is correctly caught by V:
```
Running code...
Can't run code. The server returned an error:
code.v:6:13: error: for in: cannot index `!string`
    4 | fn main() {
    5 |     s := 'this is a string'
    6 |     for _ in s.substr_with_check(2, 5) {
      |                ~~~~~~~~~~~~~~~~~~~~~~~
    7 |     }
    8 | }
Exited with error status 1
```

I am guessing the issue is more precisely the use of a !int value as the lower or upper limit in the range expression of the for statement

Edit: Additionally, functions returning ?int produce the same builder error

### V version

V 0.4.4 4221522

### Environment details (OS name and version, etc.)

V full version: V 0.4.4 1c10d2f.4221522
OS: windows, Microsoft Windows 10 Home v19045 64 bits
Processor: 4 cpus, 64bit, little endian
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #21043, opened 2024-03-16
 checker: disallow `Optional` and `Result` high val in for range loop 

Delta456:
Fixes https://github.com/vlang/v/issues/21041


spytheman (approved):
Excellent work.

Issue #21047, opened 2024-03-17
for/in cannot iterate over C structs


spiveeworks:
### Describe the bug

When defining a `next` method for a C struct, for/in loops don't generate valid C code.

### Reproduction Steps

struct C.MyCStruct {
    x int
}

fn (mut self C.MyCStruct)next() ?int {
    return self.x
}

fn main() {
    iter := C.MyCStruct{}
    for each in iter {
    }
}

### Expected Behavior

This should produce C code that calls C__MyCStruct_next in a loop.

### Current Behavior

The following code is generated:

	MyCStruct iter = ((MyCStruct){.x = 0,});
	MyCStruct _t1 = iter;
	while (1) {
		_option_int _t2 = struct MyCStruct_next(&_t1);
		if (_t2.state != 0) break;
		int each = *(int*)_t2.data;
	}

The problem with this code is it tries to call `struct MyCStruct_next` as a function, instead of `C__MyCStruct_next`.

Adding a `@[typedef]` attribute unsurprisingly drops the `struct` part, just giving `MyCStruct_next` without the `C__` namespace.

### Possible Solution

It seems like the code that generates for loops, that tries to generate the function name `Module__Typename_next`, is calling something that emits the C type directly, instead of something that emits the actual namespace associated with the type.

### Additional Information/Context

Similar to another bug I just reported, this came up while trying to add iterator implementations to a type, but this time when I tried to iterate over the type itself, rather than a wrapper. (I had two different types, one of them suited being wrapped where the other one suited being iterated over directly.)

### V version

V 0.4.4 549654a

### Environment details (OS name and version, etc.)

V full version: V 0.4.4 386bd77.549654a
OS: windows, Microsoft Windows 10 Pro v19045 64-bit
Processor: 4 cpus, 64bit, little endian,

getwd: C:\Users\Owner
vexe: C:\Program Files\v\v.exe
vexe mtime: 2024-03-17 05:49:07

vroot: contains spaces, value: C:\Program Files\v
VMODULES: OK, value: C:\Users\Owner\.vmodules
VTMP: OK, value: C:\Users\Owner\AppData\Local\Temp\v_0

Git version: git version 2.37.2.windows.2
Git vroot status: weekly.2024.11-39-g549654ae
.git/config present: true

CC version: Error: 'cc' is not recognized as an internal or external command,
operable program or batch file.

thirdparty/tcc: N/A
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
[...]

PR #21052, opened 2024-03-18
cgen: fix for/in codegen when iterating over C structs

spiveeworks:
When implementing `next` for a C struct, and then looping over it with for/in, it was generating `MyCStruct_next(&_t)` or even `struct MyCStruct_next(&_t)` instead of `C__MyCStruct_next(&_t)`. See issue #21047 for more details.


spytheman:
It can be tested similarly to <https://github.com/vlang/v/pull/21042/files>, i.e.
1) create vlib/v/tests/c_structs/iterator.c:
```c
struct MyCStruct {
        int x;
};
```
2) create vlib/v/tests/c_structs/cstruct_iterator_test.c.v:
```v
#include "@VMODROOT/cstruct.h"

struct C.MyCStruct {
mut:
	x int
}

fn (mut self C.MyCStruct) next() ?int {
	if self.x >= 10 {
		return none
	}
	self.x++
	return self.x
}

fn test_iterating_over_cstructs() {
	iter := C.MyCStruct{}
	for x in iter {
		println(x)
		assert true
	}
}
```

spytheman (approved):

PR #21054, opened 2024-03-18
cgen: force C struct types which does not implement str() to be passed as ptr

### LINKED ISSUE #21053: cgen error with gcc, but not with tcc, for a program, printing an incomplete typedef C struct
**V doctor:**
```
V full version: V 0.4.4 a36c693
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-03-18 09:30:45

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.12
.git/config present: true

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg y.v`
```v
#include <X11/Xlib.h>

@[typedef]
pub struct C.Display {}
// fn (d &C.Display) str() string { return 'C.Display{}' }

fn main() {
    x := unsafe { &C.Display( malloc(1024) ) }
    println(x)
}

```

A compiled program, but with gcc. It does compile with tcc.

**What did you see instead?**
With tcc, which is OK:
```
&C.Display{}
```

With gcc, there is an error:
```
#0 11:35:51 ᛋ master /v/oo❱v -cc gcc run y.v
==================
/tmp/v_1000/y.01HS8ENRJ7YKPB6BXD22QC67E8.tmp.c:2129:35: error: parameter 1 (‘it’) has incomplete type
 2129 | static string Display_str(Display it) { return indent_Display_str(it, 0);}
      |                           ~~~~~~~~^~
/tmp/v_1000/y.01HS8ENRJ7YKPB6BXD22QC67E8.tmp.c: In function ‘Display_str’:
/tmp/v_1000/y.01HS8ENRJ7YKPB6BXD22QC67E8.tmp.c:2129:67: error: type of formal parameter 1 is incomplete
 2129 | static string Display_str(Display it) { return indent_Display_str(it, 0);}
      |                                                                   ^~
/tmp/v_1000/y.01HS8ENRJ7YKPB6BXD22QC67E8.tmp.c: At top level:
/tmp/v_1000/y.01HS8ENRJ7YKPB6BXD22QC67E8.tmp.c:2135:42: error: parameter 1 (‘it’) has incomplete type
 2135 | static string indent_Display_str(Display it, int indent_count) {
      |                                  ~~~~~~~~^~
/tmp/v_1000/y.01HS8ENRJ7YKPB6BXD22QC67E8.tmp.c: In function ‘main__main’:
...
==================
```
> [!NOTE]
>

felipensp:
Fix #21053

### DIFF
--- a/vlib/v/gen/c/auto_str_methods.v
+++ b/vlib/v/gen/c/auto_str_methods.v
@@ -1111,7 +1111,7 @@ fn struct_auto_str_func(sym &ast.TypeSymbol, lang ast.Language, _field_type ast.
 		if !field_type.is_ptr() && field_type.has_option_or_result() {
 			method_str = '(*(${sym.name}*)it${op}${final_field_name}.data)'
 		} else {
-			method_str += '${prefix}it${op}${final_field_name}'
+			method_str = '${prefix}it${op}${final_field_name}'
 		}
 		if sym.kind == .bool {
 			return '${method_str} ? _SLIT("true") : _SLIT("false")', false
--- a/vlib/v/gen/c/auto_str_methods.v
+++ b/vlib/v/gen/c/auto_str_methods.v
@@ -256,7 +256,7 @@ fn (mut g Gen) gen_str_for_alias(info ast.Alias, styp string, str_fn_name string
 	}
 	mut clean_type_v_type_name := util.strip_main_name(styp.replace('__', '.'))
 
-	is_c_struct := parent_sym.is_c_struct()
+	is_c_struct := parent_sym.is_c_struct() && str_method_expects_ptr
 	arg_def := if is_c_struct { '${styp}* it' } else { '${styp} it' }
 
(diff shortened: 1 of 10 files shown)


spytheman (approved):
Excellent work.

vlib/v/gen/c/auto_str_methods.v
```diff
@@ -1065,6 +1076,25 @@ fn (mut g Gen) gen_str_for_struct(info ast.Struct, lang ast.Language, styp strin
 	fn_body.writeln('\t}));')
 }
 
+// c_struct_ptr handles the C struct argument for .str() method
+@[inline]
```
spytheman, line 1080: imho that does not need to be inline, and it does not need to be pub too. It is only called internally in the cgen module.

Issue #21049, opened 2024-03-17
c error while trying to map it.wait() on a thread of arrays


not-ivy:
**V doctor:**
```
V full version: V 0.4.4 ac2dcc2
OS: linux, "Arch Linux"
Processor: 16 cpus, 64bit, little endian, 12th Gen Intel(R) Core(TM) i7-1260P

getwd: /home/nora/Documents/foxglove-world/fox
vexe: /usr/lib/vlang/v
vexe mtime: 2024-03-16 22:23:00

vroot: NOT writable, value: /usr/lib/vlang
VMODULES: OK, value: /home/nora/.vmodules
VTMP: OK, value: /tmp/v_60271

Git version: git version 2.44.0
Git vroot status: Error: fatal: not a git repository (or any of the parent directories): .git
.git/config present: false

CC version: cc (GCC) 13.2.1 20230801
thirdparty/tcc: N/A

```

**What did you do?**
`v run .`
```v
module cmds

import os
import cli
import net.urllib

pub fn sync() cli.Command {
	return cli.Command{
		name: 'sync'
		description: 'sync local dens from remote'
		execute: execsync
	}
}

fn execsync(cmd cli.Command) ! {
	dens := os.read_lines('/etc/fox/dens')!.map(urllib.parse(it)!)
	mut threads := []thread !{}
	for den in dens {
		threads << spawn update(den)
	}
	threads.map(it.wait()!)
}

fn update(den urllib.URL) ! {
	println(den.str())
	return
}

```

**What did you expect to see?**

code works as expected

**What did you see instead?**
```
==================
/tmp/v_60271/fox.6109388202681245468.tmp.c:24611: error: declaration of void object
...
==================
(Use `v -cg` to print the entire error message)

builder error: 
==================
C error. This should never happen.

This is a compiler bug, please report it using `v bug file.v`.

https://github.com/vlang/v/issues/new/choose

You can also use #help on Discord: https://discord.gg/vlang
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

not-ivy:
using regular `for..in` loop works fine, like the following
```v
fn execsync(cmd cli.Command) ! {
	dens := os.read_lines('/etc/fox/dens')!.map(urllib.parse(it)!)
	mut threads := []thread !{}
	for den in dens {
		threads << spawn update(den)
[...]

PR #21055, opened 2024-03-18
 checker: disallow `Result` callbacks functions like `map/filter/all/any`

Delta456:
Fixes https://github.com/vlang/v/issues/21049


spytheman:
`.map(f()!)` should be allowed, since `f()!` is sugar for `f() or { return err }`, which is well defined.

spytheman:
The checker however should check whether `f()!` returns a valid type . 
In other words, it should allow:
`f1()!`, for `fn f1() !Type {`
... but it should not allow:
`f2()!` for `fn f2() ! {`

Delta456:
> The checker however should check whether `f()!` returns a valid type . In other words, it should allow: `.map( f1()! )`, for `fn f1() !Type {` ... but it should not allow: `.map( f2()! )` for `fn f2() ! {`

Fair

spytheman (approved):
Excellent work.
Sorry for the delay.

ttytm:
Wanted to try to utilize it instead of an or block, I'm still running into a builder error here:

https://github.com/vlang/v/blob/bccf660456713122e1db652a0b729ff814ba86c4/vlib/x/json2/decoder.v#L292-L294

PR #21057, opened 2024-03-18
vdoc: cleanup `get_module_list`, add unit tests

ttytm:
Cleans up the `get_module_list` function and adds simple unit tests for it and `get_ignore_paths`. Hoping this creates a better base for further development of such features. ~~E.g. I'd like to add the functionality for a `.vfmtignore` - similar to a `.prettierignore`, but for now with the simple functionality of the `.vdocignore` - in a follow up.~~


JalonSolov:
Did you mean `.vdocignore`?  Alex does not want `.vfmtignore` - he doesn't want _ANY_ options to vfmt, as it is supposed to stay opinionated.

ttytm:
I'm coming to the `.vfmtignore` as you currently can't run `v fmt` on the. v-analzyer project. 
V has some ignoring formatted for some internal files hardcoded. Instead of adding `vfmt off` to every file a `.vfmtignore` would be just logical. There would be no new options.

spytheman:
> The commit is related to cleaning up tests the `.vdocignore` functionality.
> 
> A feature I would like to implement is a `.vfmtignore`. There already is `// vfmt off` which is the same functionality that a `.vfmtignore` would do.

No, it is not. The difference is scope.
A feature like .vfmtignore will shift the balance towards making it extremely easy to just put a `*` pattern in it. I want the balance to be towards most projects using vfmt, since that is the easiest.

To do that using `// vfmt off`, you would have to put that in every single .v file, since it was designed for *local* changes, not for global ones, like a .vfmtignore file would allow.

spytheman:
>> V is ignoring formatting for some internal files hardcoded. For other projects and for the V project itself, instead of hardcoding / adding vfmt off to every problematic file`

It is better for the ecosystem to gradually fix those, not to make it easier for projects to use a global .vfmtignore file.

cmd/tools/vdoc/tests/testdata/module_list_with_ignore/testdata/echo/echo.v
```diff
@@ -0,0 +1 @@
+
```
spytheman, line 1: Why put a project with empty files, in a testdata/ folder, inside another testdata/ folder?

spytheman: `!it.contains('testdata')` - this will be affected by the parent folder too, not just by the inner one.

ttytm: Yep the test files need an update. I didn't made it to continue working on the PR yet.

> !it.contains('testdata') - this will be affected by the parent folder too, not just by the inner one.

The path is strip to the relative folder so the parent folder is not contained here.

ttytm:
The decision to not have a .vfmtignore is comprehensible. I'll won't submit a PR. The issue of not beeing able to easily run `v fmt -verify/-c .` on a project like v analyzer  and tell if there are formatting errors next to internal errors was solved by a recent PR extending the exit codes.

spytheman:
v-analyzer can just use .vv for those files, similar to what V itself does.
For example, after:
```sh
rename s/.v$/.vv/gm metadata/stubs/*.v
rename s/.v$/.vv/gm editors/code/syntaxes/tests/*.v
rename s/.v$/.vv/gm tests/testdata/documentation/*.v
rename s/.v$/.vv/gm tests/testdata/types/*.v
```
vfmt does not complain anymore:
```
#0 11:58:29 ᛋ main /v/v-analyzer❱v fmt -c .
#0 11:59:28 ᛋ main /v/v-analyzer❱
```

cmd/tools/vdoc/tests/testdata/module_list_with_ignore/testdata/echo/echo.v
ttytm: The testdata and tests directory in the testdata directory should serve purpose of providing a testcase for the files that are ignored by default by vdoc. So it should test that modules in such directories won't be included. 

But I can write a better and simpler test for everything that should be tested in this PR when the testing mechanism for vdoc gets improved first. I'll open a separate PR for this.

ttytm: Ended up adding the necessary test files as "virtual-files" and creating them inside the instead of adding them to the testdata directory. Should also help to understand the test better

spytheman (approved):

PR #21058, opened 2024-03-18
tools.fmt: extend exit codes to allow spotting unformatted files next to internal errors

ttytm:
This extends v fmts exit codes in case of errors. This way it would allow to spot if there are unformatted files next to internal errors, at least it would be easily checkable by a script based on the error code.  
E.g. a current use case would be using v fmt with with v-analyzer and encountering internal vfmt erros.

### DIFF
--- a/cmd/tools/vfmt.v
+++ b/cmd/tools/vfmt.v
@@ -99,7 +99,7 @@ fn main() {
 		}
 	}
 	mut errors := 0
-	mut internal_error_code := 0
+	mut has_internal_error := false
 	for file in files {
 		fpath := os.real_path(file)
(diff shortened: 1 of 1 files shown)


cmd/tools/vfmt.v
```diff
@@ -111,6 +112,7 @@ fn main() {
 			eprintln(worker_result.output)
 			if worker_result.exit_code == 1 {
 				eprintln('Internal vfmt error while formatting file: ${file}.')
+				internal_error_code = 128
```
spytheman, line 115: Use a smaller offset, codes above 127 are not portable.

PR #21059, opened 2024-03-18
vtest-self: add sandboxed packaging case

davlgd:
I'm working on a package for Exherbo Linux (a source based derivative of Gentoo). Packaging is done in a sandbox, without access to network and where compilers should be launched in a specific way. So, some essentials tests always fail and should be skipped. I added a variable for that.

### DIFF
--- a/cmd/tools/vtest-self.v
+++ b/cmd/tools/vtest-self.v
@@ -6,6 +6,7 @@ import testing
 const github_job = os.getenv('GITHUB_JOB')
 
 const just_essential = os.getenv('VTEST_JUST_ESSENTIAL') != ''
+	|| os.getenv('VTEST_SANDBOXED_PACKAGING') != ''
 
 const essential_list = [
 	'cmd/tools/vvet/vet_test.v',
@@ -338,6 +339,12 @@ const skip_on_non_amd64_or_arm64 = [
 	'vlib/sync/many_times_test.v',
 	'do_not_remove',
 ]
+const skip_on_sandboxed_packaging = [
+	'do_not_remove',
+	'vlib/v/slow_tests/inout/compiler_test.v',
+	'vlib/v/compiler_errors_test.v',
+	'vlib/v/gen/c/coutput_test.v',
+]
 
 // Note: musl misses openssl, thus the http tests can not be done there
 // Note: http_httpbin_test.v: fails with 'cgen error: json: map_string_string is not struct'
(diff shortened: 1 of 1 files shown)


cmd/tools/vtest-self.v
```diff
@@ -338,6 +339,12 @@ const skip_on_non_amd64_or_arm64 = [
 	'vlib/sync/many_times_test.v',
 	'do_not_remove',
 ]
+const skip_on_sandoxed_packaging = [
```
Larpon, line 342:
```suggestion
const skip_on_sandboxed_packaging = [
```

cmd/tools/vtest-self.v
```diff
@@ -444,6 +451,9 @@ fn main() {
 	if os.getenv('V_CI_UBUNTU_MUSL').len > 0 {
 		tsession.skip_files << skip_on_ubuntu_musl
 	}
+	if os.getenv('VTEST_SANDBOXED_PACKAGING').len > 0 {
+		tsession.skip_files << skip_on_sandoxed_packaging
```
Larpon, line 455:
```suggestion
		tsession.skip_files << skip_on_sandboxed_packaging
```

Larpon (commented):
Spotted a few typos

spytheman (approved):

cmd/tools/vtest-self.v
```diff
@@ -338,6 +339,12 @@ const skip_on_non_amd64_or_arm64 = [
 	'vlib/sync/many_times_test.v',
 	'do_not_remove',
 ]
+const skip_on_sandboxed_packaging = [
+	'do_not_remove',
+	'vlib/v/slow_tests/inout/compiler_test.v',
+	'vlib/v/compiler_errors_test.v',
+	'vlib/v/gen/c/coutput_test.v',
```
spytheman, line 346: Just curious, why is the `coutput_test.v` excluded?
Afaik it does not make network calls, it just runs programs and compares their output, or the generated c source for them, to known patterns.

spytheman: Can you share the output when it fails in the sandbox?

spytheman: `vlib/v/slow_tests/inout/compiler_test.v` should be the same - it just compares the output of running programs to known patterns.

spytheman: I see. `coutput_test.v` does try to detect whether `gcc` is in PATH, and if it is not, it skips the test with it, but since it is, it tries to use it.

Thanks for the example.

PR #21060, opened 2024-03-19
fmt: fix formatting for imports of submodule from module src dir

### LINKED ISSUE #19799: `fmt` breaks strange module import
### Describe the bug

I have the following import in my project

```
import artemkakun.textproc.src.punctuation
```

This code works but looks pretty invalid. Anyway, `fmt` formats this import as 

```
import punctuation
```

which is invalid for sure

### Reproduction Steps

```
import artemkakun.textproc.src.punctuation
```

### Expected Behavior

Either error about invalid module setup/path or nothing

### Current Behavior

`fmt` formats import in the wrong way

### Possible Solution

_No response_

_No response_

### V version

V 0.4.2 f54f156.f82529e

```
V full version: V 0.4.2 f54f156.f82529e
OS: linux, "Garuda Linux"
Processor: 16 cpus, 64bit, little endian, AMD Ryzen 7 3800X 8-Core Processor

getwd: /home/yuart/Projects/seven-concurrency-models-v
vexe: /home/yuart/Projects/v/v
vexe mtime: 2023-11-05 12:24:30

vroot: OK, value: /home/yuart/Projects/v
VMODULES: OK, value: /home/yuart/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.42.1
Git vroot status: weekly.2023.42-147-gf82529e0 (15 commit(s) behind V master)
.git/config present: true

```
> [!NOTE]
>

ttytm:
Resolves #19799

With the linked PR vfmt won't break the import anymore. I think, since it's a working import it's okay when fmt just keeps it for now. Eventually we can check if a module placed in the `src` dir of a parent module is recognize when importing it without the src dir like `import artemkakun.textproc.punctuation` and let vfmt do the job of removing the `src.` part - currently, the module would not be found.

### DIFF
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -382,12 +382,14 @@ pub fn (mut f Fmt) imports(imports []ast.Import) {
 pub fn (f Fmt) imp_stmt_str(imp ast.Import) string {
 	normalized_mod := if f.inside_vmodules {
 		imp.source_name
+	} else if imp.mod.len == 0 {
+		imp.alias
 	} else {
-		mod := if imp.mod.len == 0 { imp.alias } else { imp.mod }
-		if mod.starts_with('src.') {
-			mod.all_after('src.')
+		mod_last := imp.mod.all_after_last('.')
+		if imp.source_name == mod_last {
+			mod_last
 		} else {
-			mod
+			imp.mod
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #21063, opened 2024-03-19
 parser: add better error for mut variadic fn argument 

Delta456:
Code snippet:

```v
fn load(mut filepaths ...string) {
	println(filepaths[0])
}
```

Before:

(before this PR)

```
file.v:1:26: error: mutable arguments are only allowed for arrays, interfaces, maps, pointers, structs or their aliases
return values instead: `fn foo(mut n string) {` => `fn foo(n string) string {`
    1 | fn load(mut filepaths ...string) {
      |                          ~~~~~~
    2 |     println(filepaths[0])
    3 | }

```

After:

(after this PR)

```
file.v:1:26: error: variadic arguments cannot be `mut`, `shared` or `atomic`
    1 | fn load(mut filepaths ...string) {
      |                          ~~~~~~
    2 |     println(filepaths[0])
    3 | }
```


spytheman (approved):
Excellent work.

PR #21065, opened 2024-03-19
ci: add testing with clang on windows too

spytheman:


spytheman:
Closed for now, till the CI failure is fixed on my branch, to not consume resources from the shared CI queue.

PR #21066, opened 2024-03-19
fmt: update determining of struct field comments

ttytm:
The PR updates how struct field comments are determined. The goal is to increase robustness and to better resemble what is happening to improve maintainability. It uses a single instead of multiple loops.


spytheman (approved):
Excellent work.

PR #21069, opened 2024-03-20
examples: port sokol/clear.v and sokol/sdf/sdf.v

spytheman:
Port https://github.com/floooh/sokol-samples/blob/master/sapp/clear-sapp.c .
image

Port https://github.com/floooh/sokol-samples/blob/master/sapp/sdf-sapp.c .
image


Larpon:
Nice! Good work!

PR #21070, opened 2024-03-21
parser,fmt: inform about invalid interop function bodies instead of removing them

### LINKED ISSUE #12966: v fmt removes function bodies without warning
**What did you do?**
**V version:** V full version: V 0.2.4 3528239
**OS:** OS: macos, Mac OS X, 10.15.7, 19H1419

the js backend doesnt error and the `v fmt` "removes" the function body without warning.

```v
$ cat j.v
fn JS.state_is_ready_on_leave(retval int) {
       println('Hooked callback')
}
$ v fmt j.v
fn JS.state_is_ready_on_leave(retval int)
$
```

so using the cgen results in proper error

```v
$ v fmt j.v
j.v:1:42: error: interop functions cannot have a body
    1 | fn C.state_is_ready_on_leave(retval int) {
      |                                          ^
    2 |        println('Hooked callback')
    3 | }

Internal vfmt error while formatting file: j.v.
Encountered a total of: 1 errors.
$ cat j.v
fn C.state_is_ready_on_leave(retval int) {
       println('Hooked callback')
}
$
```

## Test

see above

 proper error at format and compile times because it could remove sensitive code by accident

**What did you see instead?**

silence and void

ttytm:
Fixes #12966

Where formatting
```v
fn JS.state_is_ready_on_leave(retval int) {
	println('Hooked callback')
}
```
results in
```v
fn JS.state_is_ready_on_leave(retval int)
```

With the changes it would throw an error instead
```
error: interop functions cannot have a body
    1 | fn JS.state_is_ready_on_leave(retval int) {
      |                                           ^
    2 |     println('Hooked callback')
    3 | }
```

### DIFF
--- a/vlib/v/parser/fn.v
+++ b/vlib/v/parser/fn.v
@@ -577,7 +577,7 @@ run them via `v file.v` instead',
 	mut stmts := []ast.Stmt{}
 	body_start_pos := p.tok.pos()
 	if p.tok.kind == .lcbr {
-		if language != .v && language != .js {
+		if language != .v {
 			p.error_with_pos('interop functions cannot have a body', body_start_pos)
 		}
 		p.inside_fn = true


vlib/v/parser/fn.v
```diff
@@ -577,7 +577,7 @@ run them via `v file.v` instead',
 	mut stmts := []ast.Stmt{}
 	body_start_pos := p.tok.pos()
 	if p.tok.kind == .lcbr {
-		if language != .v && language != .js {
+		if language != .v {
```
spytheman, line 580: That was deliberately changed in 6c32c54, but there does not seem to exist a test for it 🤔 .

ttytm: Ty! Gonna look info fixing it for the interface methods and to add a test.

spytheman: It seems to work on both master and here with `-b js_browser`

PR #21074, opened 2024-03-21
vdoc,tests: detect expected files from input instead of hardcoding names

ttytm:
The PR extends vdoc tests while also simplifying the process of adding test files with modules/filenames other than main. This allows for a simpler inclusion of imho more "natural" use cases, as hared modules like libraries are the larger user base for public functions with generated documentation. It also updates one of the tests to make use of this new functionality. The changes should allow to tests vdoc functionality more easily when solidifying and extending it.


spytheman (approved):
Excellent work.

PR #21083, opened 2024-03-23
utils: fix module lookup if module import parts end with the word `modules`

### LINKED ISSUE #21078: Cannot compile `examples/submodules`
### Describe the bug

First time using v, I was looking around the examples and I cannot compile and run it. Using the latest relaese of `v` - `V 0.4.5 8a9def6`.

The LSP also gives squiggly lines saying it cannot find the modules. I think the example is broken. 

### Reproduction Steps

1. `cd examples/submodule`
2. `v run main.v` 

### Expected Behavior

Program runs without issues.

### Current Behavior

Error:
```sh
main.v:1:1: builder error: cannot import module "mymodules.mymodule" (not found)
    1 | import mymodules.mymodule { add_xy }
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    2 | import mymodule.submodule { sub_xy }
    3 |

```

### Possible Solution

I don't know

_No response_

### V version

V 0.4.5 8a9def6

Ubuntu Linux 6.2.0-39-generic x86_64 GNU/Linux

ttytm:
Fixes #21078 

Contains a little module import name correction.

Should we run the related example run in CI and add a info comment to not change the modules name?

### DIFF
--- a/.github/workflows/linux_ci.yml
+++ b/.github/workflows/linux_ci.yml
@@ -57,6 +57,8 @@ jobs:
       #   run: V_CI_CSTRICT=1 ./v -cstrict test-self
       - name: Build examples
         run: ./v -W build-examples
+      - name: Run the submodule example, using a relative path
+        run: ./v -W run examples/submodule
       - name: Build v tools
         run: ./v -W build-tools
       - name: Build v binaries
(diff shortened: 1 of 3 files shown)


spytheman:
> Should we run the related example run in CI and add a info comment to not change the modules name?

Yes, that would be best. `v build-examples` uses only the absolute paths, and thus worked, while compiling with the relative path was a problem.

spytheman (approved):
Excellent work.

PR #21084, opened 2024-03-24
get UdpSocket local address 

wilesun:
UdpSocket l field has local address,but it's not public, add the local_addr()


spytheman:
Why was it closed?

PR #21088, opened 2024-03-24
parser, fmt: parse methods on JS interfaces, write JS method bodies

ttytm:
Builds on #21070 and it's comments.

Adds a test for the scenario and for #12447.


spytheman (approved):
Excellent work.

PR #21090, opened 2024-03-24
ast: cleanup

ttytm:
Simplifies `ast.v`

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -685,14 +685,11 @@ pub mut:
 }
 
 pub fn (p &Param) specifier() string {
-	if p.is_shared {
-		return 'shared'
-	} else if p.is_atomic {
-		return 'atomic'
-	} else if p.is_mut {
-		return 'mut'
-	} else {
-		return ''
+	return match true {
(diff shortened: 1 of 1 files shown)


vlib/v/ast/ast.v
```diff
@@ -685,14 +685,11 @@ pub mut:
...
-		return 'atomic'
-	} else if p.is_mut {
-		return 'mut'
-	} else {
-		return ''
+	return match true {
+		p.is_shared { 'shared' }
+		p.is_atomic { 'atomic' }
+		p.is_mut { 'mut' }
+		else { '' }
```
Delta456, line 692: Can this be a short hand if statement instead?

ttytm: If I may ask, why not a match statement?

Delta456: I don't think this has any advantage over a short-hand-if statement.

spytheman (approved):
I am a bit ambivalent towards the change here 🤔 .

Imho, the code was clearer before (more imperative). 
However, it is clear also, that it is much shorter, and easier to read, after the change.

Excellent work.

PR #21092, opened 2024-03-25
doc: fix vup/vself replacement scripts

davlgd:
In the documentation, there are scripts to replace `vup.v` and `vself.v` for packaging. They are composed of multiple `echo` to fill in a file line by line. 

But after the first line we should use `echo "content" >> dest_file.v` to append content and not `echo "content" > dest_file.v` to replace it. 

This PR fixes this.


spytheman (approved):
Good spot.

PR #21095, opened 2024-03-25
vtest self: add native_test to skippable tests for sandboxed packaging

davlgd:
For packaging purposes, I've added the `native_test` in skipped essential tests from `v test-self`, as Exherbo Linux packages we can't use external compilers/linkers during install. This test use `ld` and then leads to an unfixable error.


spytheman:
What is the error that you get?

spytheman (approved):

PR #21101, opened 2024-03-26
v: fix sumtype option support

### LINKED ISSUE #21093: c error on unwrap of an option type in a match
### Describe the bug

Unwrapping an option type using an if statement in a sum type match results in a `cannot convert` c error.

### Reproduction Steps

```v
fn main() {
	f := Foo(Bar{})
	match f {
		?Bar {
			if i := f {
				println(i)
			}
		}
		else {}
	}
}

type Foo = ?Bar | Baz
struct Baz {}
struct Bar {}

```

### Expected Behavior

To compile and print `Bar{}`.

### Current Behavior

results in `error: cannot convert 'struct main__Foo' to 'struct _option_main__Bar'`

### Possible Solution

_No response_

_No response_

### V version

V 6b23b1d

V full version: V 0.4.5 9d7fe51.6b23b1d
OS: linux, "Arch Linux"
Processor: 12 cpus, 64bit, little endian, AMD Ryzen 5 3600X 6-Core Processor

vroot: OK
VMODULES: OK
VTMP: OK, value: /tmp/v_1000

Git vroot status: 0.4.5-19-g6b23b1dd
.git/config present: true

> [!NOTE]
>

felipensp:
Fix #21093

### DIFF
--- a/.github/workflows/v_apps_and_modules_compile_ci.yml
+++ b/.github/workflows/v_apps_and_modules_compile_ci.yml
@@ -48,7 +48,7 @@ jobs:
           echo "Clone https://github.com/DarpHome/discord.v/"
           git clone https://github.com/DarpHome/discord.v/ discord && cd discord
           echo "Checkout last known good commit"
-          git checkout 65448b687ae759e385c127c8739b97b6ac62d3e3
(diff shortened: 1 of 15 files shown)


spytheman (approved):
Excellent work.

PR #21105, opened 2024-03-26
checker: optimize option and result typ check, add more typinfo to error details

ttytm:
The main change of the PR is improving error messages to provide more details on return types. The changes serve as a base for a bugfix in a followup but take up their own scope so they are submitted separately.

The PR intentionally makes minimal changes to the error messages, keeping it similar to the current state. Imho the error messages can generally be optimized here by using error details to split things up into showing an expressive error message and providing a recommendation for action. But that's a scope for a separate PR as well.

There are also minor optimization to the codeblock of the check. It moves variables assignments deeper into the block scope where they are used and removes a duplicate conditional check. Modern compilers should be smart enough to optimize such variable assignments, so the optimization likely mainly helps with read-/maintainability.


spytheman (approved):
Excellent work.

PR #21106, opened 2024-03-26
checker, orm: add error for unchecked option multi return types, fix undefined behavior

### LINKED ISSUE #17874: the checker allows for `a, b := abc()` where `fn abc() ?(int,int)`, but it should require the `none` to be handled instead

### LINKED ISSUE #21117: undefined values for an option multireturning function, that executes `return none`, instead of a checker error
**V doctor:**
```
V full version: V 0.4.5 e6b43a1.8511731
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-03-27 13:36:18

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.13-27-g7e0e2aaa
.git/config present: true

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg x.v`
```v
fn f() ?(int, string) {
	// return 42, 'the answer'
	return none
}

x, y := f()
dump(x)
dump(y)

```

A checker error, telling me that `x,y := f()` should be `if x, y := f() {` or `x, y := f()?` instead .

**What did you see instead?**
```
[x.v:7] x: 0
[x.v:8] y: 
```
> [!NOTE]
>

ttytm:
Fixes #17874  
Fixes #21117
Due to the changes, a fix to the vlib orm module is included.

```v
fn split(str string) ?(string, string) {
	return none
}

fn main() {
	foo, bar := split('foo/bar')
	println('${foo}.${bar}')
}
```

Will now throw:
```
error: split() returns `?(string, string)`, so it should have either an `or {}` block, or `?` at the end
    4 | 
    5 | fn main() {
    6 |     foo, bar := split('foo/bar')
      |                 ~~~~~~~~~~~~~~~~
    7 |     println('${foo}.${bar}')
    8 | }
```

Based on #21105 (After resolving it first, the PR should rebased)

Superseeds #20515. The PR is stale for a while, but took good steps towards the solution.

### DIFF
--- a/vlib/orm/orm.v
+++ b/vlib/orm/orm.v
@@ -508,12 +508,13 @@ pub fn orm_table_gen(table string, q string, defaults bool, def_unique_len int, 
 							return error("references attribute needs to be in the format [references], [references: 'tablename'], or [references: 'tablename(field_id)']")
 						}
 						if attr.arg.contains('(') {
-							ref_table, ref_field := attr.arg.split_once('(')
-							if !ref_field.ends_with(')') {
-								return error("explicit references attribute should be written as [references: 'tablename(field_id)']")
+							if ref_table, ref_field := attr.arg.split_once('(') {
+								if !ref_field.ends_with(')') {
+									return error("explicit references attribute should be written as [references: 'tablename(field_id)']")
+								}
+								references_table = ref_table
+								references_field = ref_field[..ref_field.len - 1]
 							}
-							references_table = ref_table
-							references_field = ref_field[..ref_field.len - 1]
 						} else {
 							references_table = attr.arg
 							references_field = 'id'
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -1161,17 +1161,20 @@ fn (mut c Checker) check_expr_option_or_result_call(expr ast.Expr, ret_type ast.
 				}
 			}
 			if expr_ret_type.has_option_or_result() {
-				if expr.or_block.kind == .absent && expr_ret_type.has_flag(.result) {
+				if expr.or_block.kind == .absent {
 					ret_sym := c.table.sym(expr_ret_type)
-					ret_typ_tok := if expr_ret_type.has_flag(.option) { '?' } else { '!' }
-					if c.inside_defer {
-						c.error('${expr.name}() returns `${ret_typ_tok}${ret_sym.name}`, so it should have an `or {}` block at the end',
-							expr.pos)
-					} else {
-						c.error('${expr.name}() returns `${ret_typ_tok}${ret_sym.name}`, so it should have either an `or {}` block, or `${ret_typ_tok}` at the end',
-							expr.pos)
+					if expr_ret_type.has_flag(.result)
+						|| (expr_ret_type.has_flag(.option) && ret_sym.kind == .multi_return) {
+						ret_typ_tok := if expr_ret_type.has_flag(.option) { '?' } else { '!' }
+						if c.inside_defer {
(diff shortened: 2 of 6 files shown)


JalonSolov:
Not a good idea mention `tuples` at all.  V doesn't support tuples, only multi-returns.

ttytm:
Ok thanks @JalonSolov gonna update. 
Wasn't sure as we have `ast.Tuple` and `ast.TupleType` and also the changelog mentions 
> - Fix returning an option tuple - `fn f() ?(int,int) { return g() }` (#18851)

vlib/orm/orm.v
```diff
@@ -508,12 +508,13 @@ pub fn orm_table_gen(table string, q string, defaults bool, def_unique_len int,
 							return error("references attribute needs to be in the format [references], [references: 'tablename'], or [references: 'tablename(field_id)']")
 						}
 						if attr.arg.contains('(') {
-							ref_table, ref_field := attr.arg.split_once('(')
-							if !ref_field.ends_with(')') {
-								return error("explicit references attribute should be written as [references: 'tablename(field_id)']")
+							if ref_table, ref_field := attr.arg.split_once('(') {
```
spytheman, line 511: hm, that `x,y := f()`, may need to become a checker error for `fn f() ?(Type1, Type2) {`, if it is not already, since `f()` can return `none`, and then `x` and `y` will not be well defined.

spytheman: I've filed https://github.com/vlang/v/issues/21117 for it.

spytheman (approved):
Excellent work 🚀 .

vlib/orm/orm.v
ttytm: Should be an error after the PR. That's why this fix here became mandatory for test to pass in this PR

ttytm: ****

PR #21107, opened 2024-03-26
parser: improve expected token detection in error messages for unterminated fn calls

### LINKED ISSUE #21100: "bad" error message for unterminated `println`
### Describe the bug

If you forget the closing `)` on a `println` call, V tells you there is a missing comma.

This is incorrect, since `println` only takes a single parameter, not a comma-separated list.

Unknown if this is happening with other single parameter functions, but likely.

### Reproduction Steps

Filename: `x.v`
```v
import os

println(os.file_last_mod_unix('x.v')
```
Try to run it with `v run x.v`

### Expected Behavior

An error message stating that the final closing `)` is missing.

### Current Behavior

```
x.v:4:1: error: unexpected eof, expecting `,`
    2 | 
    3 | println(os.file_last_mod_unix('x.v')
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.5 b01e091

```
V full version: V 0.4.5 efa98d9.b01e091
OS: linux, "Manjaro Linux"
Processor: 32 cpus, 64bit, little endian, AMD Ryzen 9 7950X 16-Core Processor

getwd: /home/jalon
vexe: /home/jalon/git/v/v
vexe mtime: 2024-03-25 14:04:33

vroot: OK, value: /home/jalon/git/v
VMODULES: OK, value: /home/jalon/.vmodules
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.13-8-gb01e0914
.git/config present: true

```

ttytm:
Fixes #21100

So that the expected token is shown as `)` instead of `,`.

```v
import os

println(os.file_last_mod_unix('x.v')
```

```v
import os

fn foo(i i64) {
	println(i)
}

foo(os.file_last_mod_unix('x.v')
```

### DIFF
--- a/vlib/v/parser/tests/fn_call_unexpected_eof_err.out
+++ b/vlib/v/parser/tests/fn_call_unexpected_eof_err.out
@@ -1,0 +1,3 @@
+vlib/v/parser/tests/fn_call_unexpected_eof_err.vv:5:1: error: unexpected eof, expecting `)`
+    3 | println(os.file_last_mod_unix('x.v')
+    4 |
--- a/vlib/v/parser/tests/fn_call_unexpected_eof_err.vv
+++ b/vlib/v/parser/tests/fn_call_unexpected_eof_err.vv
@@ -1,0 +1,4 @@
+import os
+
+println(os.file_last_mod_unix('x.v')
+


ttytm:
Closing for now. The fix is harder then expected. 
Currently, accessing the number of params of a function when checking the args of it's call expression would add quite some complication. Afaik, it's nowhere done during parsing yet. 

The number of params would help to tell if a comma - as it is currently printent - or right bracket is the correct token to check for.

Issue #21003, opened 2024-03-12
VC: _M_ARM64 check missing for endianness


daddycocoaman:
### Describe the bug

Hi,

I posted about this (https://github.com/vlang/v/discussions/13768#discussioncomment-8725449) and while updating my V build, I realized that `v_win.c` actually does `_M_ARM64` checks a few times to set some values, but that check isn't defined when it comes to the endianess on line 857. I know V isn't officially supported on Windows ARM64, but I think that check should at least be there based on other checks done in the same file.

Right now, I'm adding this manually when updating V, but this is _**probably**_  a bug. Just not sure where to fix it.

 

### Reproduction Steps

`./make.bat -msvc` on Windows ARM64

### Expected Behavior

Modifying the check in `v_win.c` lets me successfully compile the x64 version of V on Windows ARM64 using MSVC with `./make.bat -msvc --local`:

```
PS v-dev\v> .\make.bat -msvc --local
Building V...
 > Attempting to build "./v_win_bootstrap.exe" (from v_win.c) with MSVC
v_win.c
 > Compiling "./v.exe" with "./v_win_bootstrap.exe"
<OMITTED>

Note: `tcc` was not used, so unless you install it yourself, your backend
C compiler will be `cc`, which is usually either `clang`, `gcc` or `msvc`.

These C compilers, are several times slower at compiling C source code,
compared to `tcc`. They do produce more optimised executables, but that
is done at the cost of compilation speed.

 > V built successfully
 > To add V to your PATH, run `./v.exe symlink`.

V version: V 0.4.4 365bd18
```

### Current Behavior

```
PS v-dev\v> .\make.bat -msvc --local
Building V...
 > Attempting to build "./v_win_bootstrap.exe" (from v_win.c) with MSVC
v_win.c
vc/v_win.c(860): fatal error C1189: #error:  "Unknown architecture endianness"
In some cases, compile errors happen because of the MSVC compiler version
Microsoft (R) C/C++ Optimizing Compiler Version 19.39.33519 for ARM64
Copyright (C) Microsoft Corporation.  All rights reserved.

usage: cl [ option... ] filename... [ /link linkoption... ]

Backend compiler error

Exiting from error
ERROR: please follow the instructions in https://github.com/vlang/v/wiki/Installing-a-C-compiler-on-Windows
```

### Possible Solution

Add the _M_ARM64 check like this, wherever this code gets generated from:

```c
[...]

PR #21109, opened 2024-03-26
v.gen: add _M_ARM64 to endianness check

daddycocoaman:
This PR addresses #21003 by adding a check that will allow for Windows ARM64 users to compile the X64 version of V with `.\make.bat -msvc` . The only remaining issue is that there is a harmless but loud error that shows up from `detect_tcc` because TCC is downloaded even when another compiler is specified (which maybe shouldn't be the case?). But everything works:

```
Updating TCC
 > Syncing TCC from https://github.com/vlang/tccbin

Updating vc...
 > Sync with remote https://github.com/vlang/vc

Building V...
 > Attempting to build "./v_win_bootstrap.exe" (from v_win.c) with MSVC
v_win.c
 > Compiling "./v.exe" with "./v_win_bootstrap.exe"
> C compiler cmd: "C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.39.33519\bin\Hostx64\x64\cl.exe" "@C:\Users\*\AppData\Local\Temp\v_0\v_up.exe.tmp.c.rsp"
> C compiler response file "C:\Users\*\AppData\Local\Temp\v_0\v_up.exe.tmp.c.rsp":
-w /we4013 /volatile:ms /Fo"C:\Users\*\AppData\Local\Temp\v_0\v_up.exe.tmp.c.obj" /F 16777216 /MDd /D_DEBUG /Zi /Fd"C:\Users\*\AppData\Local\Temp\v_0\v_up.exe.tmp.c.pdb" "C:\Users\*\AppData\Local\Temp\v_0\v_up.exe.tmp.c" -I "C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\ucrt" -I "C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.39.33519\include" -I "C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\um" -I "C:\Program Files (x86)\Windows Kits\10\Include\10.0.22621.0\shared" -I "C:\Users\*\dev\v-dev\thirdparty\stdatomic\win" kernel32.lib user32.lib advapi32.lib ws2_32.lib dbghelp.lib ws2_32.lib advapi32.lib shell32.lib /link /NOLOGO /OUT:"C:\Users\*\dev\v-dev\v_up.exe" /LIBPATH:"C:\Program Files (x86)\Windows Kits\10\Lib\10.0.22621.0\ucrt\X64" /LIBPATH:"C:\Program Files (x86)\Windows Kits\10\Lib\10.0.22621.0\um\X64" /LIBPATH:"C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.39.33519\lib\X64" /DEBUG:FULL
==================
C:/Users/*/AppData/Local/Temp/v_0/detect_tcc.01HSYEXNM7BRXKQF38GTJMN8XC.tmp.c:6895: warning: implicit declaration of function 'tcc_backtrace'
tcc: error: unrecognized file type
tcc: error: library 'dbghelp' not found
...
==================

builder error:
==================

 > V built successfully
 > To add V to your PATH, run `./v.exe symlink`.

V version: V 0.4.5 80a4249
```

### DIFF
--- a/vlib/v/gen/c/cheaders.v
+++ b/vlib/v/gen/c/cheaders.v
@@ -540,7 +540,7 @@ void v_free(voidptr ptr);
 
 #if defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__ || defined(__BYTE_ORDER) && __BYTE_ORDER == __BIG_ENDIAN || defined(__BIG_ENDIAN__) || defined(__ARMEB__) || defined(__THUMBEB__) || defined(__AARCH64EB__) || defined(_MIBSEB) || defined(__MIBSEB) || defined(__MIBSEB__)
 	#define TARGET_ORDER_IS_BIG 1
-#elif defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__ || defined(__BYTE_ORDER) && __BYTE_ORDER == __LITTLE_ENDIAN || defined(__LITTLE_ENDIAN__) || defined(__ARMEL__) || defined(__THUMBEL__) || defined(__AARCH64EL__) || defined(_MIPSEL) || defined(__MIPSEL) || defined(__MIPSEL__) || defined(_M_AMD64) || defined(_M_X64) || defined(_M_IX86)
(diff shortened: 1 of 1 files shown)


JalonSolov:
Yes, tcc should still be downloaded... if there were a proper Windows ARM64 version of tcc.

The reason is that `.\make.bat -msvc` only builds V itself with msvc, it doesn't force everything built with V to be built the same.  To do that, you would have to set `VFLAGS='-cc msvc'` or just let the normal "rollover" to msvc happen when a working tcc can't be found - which is slower than setting the env var so it doesn't try tcc at all.

tcc is _so_ much faster than msvc that it will always be the default... unless we replace it with vtcc (the V port of tcc).

spytheman (approved):

PR #21110, opened 2024-03-27
parser: improve token detection and error messages for unexpected EOFs, add tests

ttytm:
Fixes https://github.com/vlang/v/issues/21100, Second attempt of #21107.


vlib/v/checker/tests/like_operator_outside_orm_error.out
```diff
@@ -1,4 +1,4 @@
-vlib/v/checker/tests/like_operator_outside_orm_error.vv:4:15: error: unexpected name `like`, expecting `,`
+vlib/v/checker/tests/like_operator_outside_orm_error.vv:4:15: error: only `c`, `r`, `js` are recognized string prefixes, but you tried to use `like`
     2 |     name := 'Luke'
     3 | 
     4 |     println(name like 'L%')
```
ttytm, line 4: The only test that is updated. Though imho an improvement, the old error informing about an expecting a comma which would be wrong to use here.

ttytm:
How token checking is currently handled is more imperative. I've mixed feeling about this change. Feel free to close the PR 👍

vlib/v/parser/fn.v
```diff
@@ -52,8 +52,23 @@ fn (mut p Parser) call_expr(language ast.Language, mod string) ast.CallExpr {
...
+				col: ok_arg_pos.col + ok_arg_pos.len
+			}
+			p.unexpected_with_pos(pos.extend(p.tok.pos()), expecting: '`)`')
+		} else {
+			p.unexpected_with_pos(p.prev_tok.pos(), expecting: '`)`')
+		}
+	}
 	last_pos := p.tok.pos()
-	p.check(.rpar)
+	p.next()
```
spytheman, line 71: This is less strict, it will accept any token in that position.

PR #21112, opened 2024-03-27
examples: fix `simple_shader` compile error, add 2 shader examples to CI

Larpon:
Locally I get the following compile error when doing `v run examples/sokol/simple_shader_glsl/`:
```
examples/sokol/simple_shader_glsl/simple_shader.v:146:35: error: undefined ident: `state`
  144 |     mut app := unsafe { &App(user_data) }
  145 | 
  146 |     pass := sapp.create_default_pass(state.pass_action)
      |                                      ~~~~~
  147 |     gfx.begin_pass(&pass)
  148 |
examples/sokol/simple_shader_glsl/simple_shader.v:146:41: error: `state` does not return a value
  144 |     mut app := unsafe { &App(user_data) }
  145 | 
  146 |     pass := sapp.create_default_pass(state.pass_action)
      |                                            ~~~~~~~~~~~
  147 |     gfx.begin_pass(&pass)
  148 |
examples/sokol/simple_shader_glsl/simple_shader.v:146:35: error: `state.pass_action` (no value) used as value in argument 1 to `sokol.sapp.create_default_pass`
  144 |     mut app := unsafe { &App(user_data) }
  145 | 
  146 |     pass := sapp.create_default_pass(state.pass_action)
      |                                      ~~~~~~~~~~~~~~~~~
  147 |     gfx.begin_pass(&pass)
  148 |
```
Looking further into it, it looks like `sdf` and `simple_shader` is not included in the CI step and thus not tested for regressions in this line:
https://github.com/vlang/v/blob/master/.github/workflows/other_ci.yml#L103-L106

I have fixed the example and renamed the example folders so the CI now includes them in the `should-compile-all` step.

Maybe we should think about dropping the `0N_` prefix and put them in e.g. `examples/sokol/shader_examples` or `examples/sokol_shaders` subfolder instead to minimize the chance of contributions getting "overlooked" by the CI job?


spytheman (approved):

spytheman:
I do not like the renaming much, since it will break all the links that people made for those examples, but having them checked on the CI is important too.

Issue #21072, opened 2024-03-21
sql Delete Compiler bug


rodabt:
**V doctor:**
```
V full version: V 0.4.5 561fd94.24d1572
OS: linux, Debian GNU/Linux 11 (bullseye) (WSL 2)
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz

getwd: /home/rabt/devel/prisma2/OLD/tmp
vexe: /home/rabt/devel/v/v
vexe mtime: 2024-03-21 16:20:10

vroot: OK, value: /home/rabt/devel/v
VMODULES: OK, value: /home/rabt/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.30.2
Git vroot status: 0.4.5-7-g24d15720 (2 commit(s) behind V master)
.git/config present: true

CC version: cc (Debian 10.2.1-6) 10.2.1 20210110
thirdparty/tcc status: thirdparty-linux-amd64 12f392c3

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg main.v`
```v
import db.sqlite
import rand
import time

@[table: 'parameter_tables']
struct ParameterTable {
	id          string = rand.ulid()    @[primary]
	name  		string    @[unique: 'client_table']
	description string
	table_type  string = 'parameter'
	client_id   string    @[unique: 'client_table']
	created     time.Time @[default: 'CURRENT_TIMESTAMP'; sql_type: 'datetime']
}

fn main() {
	mut db := sqlite.connect('test.db')!
	db.synchronization_mode(sqlite.SyncMode.off)!
	db.journal_mode(sqlite.JournalMode.memory)!	
	sql db {
		create table ParameterTable
	}!
	sql db {
		delete from ParameterTable where client_id == client_id && name == name
	} or { panic(err) }	
}
```

**What did you expect to see?**

An error of variables not defined instead of the cryptic error

**What did you see instead?**
```
================ V panic ================
   module: v.ast
 function: default_table_panic_handler()
  message: sym: invalid type (typ=ast.Type(0x0 = 0) idx=0). Compiler bug. This should never happen. Please report the bug using `v bug file.v`.

     file: /home/rabt/devel/v/vlib/v/ast/table.v:85
   v hash: 24d1572
=========================================
/tmp/v_1000/../../../../../../home/rabt/devel/v/vlib/builtin/builtin.c.v:66: at panic_debug: Backtrace
[...]

PR #21113, opened 2024-03-27
 parser: disallow invalid infix for where clause in `delete` and `update` 

Delta456:
Fixes https://github.com/vlang/v/issues/21072


spytheman (approved):
Excellent work.

PR #21118, opened 2024-03-27
thirdparty,net.mbedtls: update mbedtls to latest compatible version v3.3.0

ttytm:
Big update of the mbedtls library to it's latest compatible release. The current latest release `3.5.2` works on linux and macos, but runs into general compatibility issues on Windows. Version 3.4.x on encounters issues on Windows with tcc. It can likely be used after our tcc upgrade.

Notes to the reviewer:
1. The commit `update thirdpary/mbedtls to 3.3.0` only updates thirdparty files. If it helps with the review process / verifying legitimacy, add the git directory of the library at the state of the release. 
   For example, pull the library to a local directory, then add its `.git` directory to Vs thirdparty/mbedtls. There should be no changes to the libraries files besides deleting unused files (for Vs codebase it follows same practice as before - not including superfluous files).
   ```sh
   # pull Mbed-TLS/mbedtls.git at v3.3.0 to access its .git dir 
   git clone --depth=1 --recursive --shallow-submodules --single-branch -b v3.3.0 https://github.com/Mbed-TLS/mbedtls.git
   ```


ttytm:
I "slightly dislike" Windows... Gonna paddle one version back, it is before a move towards a newer Windows API. This might be less headache with Windows compatibility.

Issue #21119, opened 2024-03-27
Tools not built in subfolders with build-tools


lcheylus:
### Describe the bug

With the last V version (`V 0.4.5 f6a00fa`), I have an error when I'm trying to build tools (`cmd/tools`/) with `build-tools` target. Tools with sources in sub-directories (`vast`, `vcreate`, `vdoc`, `vpm`, `vvet` and `vwhere`) are **not built**.

Same error with build on Linux Debian/amd64 and OpenBSD-current/amd64.

### Reproduction Steps

- Clone sources via `git clone`
- Build V compiler via `make`
- Build tools with `./v build-tools`

=> tools in sub-directories are not built.

### Expected Behavior

Every tool in `cmd/tools` and sub-directories are built with `./v build-tools`

### Current Behavior

Tools with sources in sub-directories (`vast`, `vcreate`, `vdoc`, `vpm`, `vvet` and `vwhere`) are not built with `./v build-tools`.

### Possible Solution

After some code analysis, this issue was not present in V version 0.4.4. It seems that commit 6a4f293 is the root cause.

- `cmd/tools/vbuild-tools.v`: `vast`, `vcreate`, `vdoc`, `vpm`, `vvet` and `vwhere` (sources in sub folders) are initially skipped, then added into a test session.
- `cmd/tools/modules/testing/common.v`: function `worker_trunner`, line 546, build is skipped for the tools list above.

**I have a working fix** : line 546 => `if !ts.build_tools && (abs_path in ts.skip_files)`

With this fix, every tool are built during `./v build-tools`

**I am going to submit a PR with this fix.**

### V version

V 0.4.5 f6a00fa

### Environment details (OS name and version, etc.)

- Linux Debian testing/amd64
- OpenBSD current/amd64
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #21120, opened 2024-03-27
tools: fix build for tools in subfolders

lcheylus:
- In 'cmd/tools/vbuild-tools.v', vast, vcreate, vdoc, vpm, vvet and vwhere tools (sources in subfolders) are initially skipped, then added into a test session.
- During session run (function 'worker_trunner'), build was skipped for the tools list above.

- Fix vlang/v#21119

---
Build with `./v build-tools` OK on Linux Debian testing/amd64 and OpenBSD current/amd64

### DIFF
--- a/cmd/tools/modules/testing/common.v
+++ b/cmd/tools/modules/testing/common.v
@@ -543,7 +543,7 @@ fn worker_trunner(mut p pool.PoolProcessor, idx int, thread_id int) voidptr {
 	}
 	ts.benchmark.step()
 	tls_bench.step()
-	if abs_path in ts.skip_files {
+	if !ts.build_tools && abs_path in ts.skip_files {
 		ts.benchmark.skip()
 		tls_bench.skip()
 		if !testing.hide_skips {


cmd/tools/modules/testing/common.v
```diff
@@ -543,7 +543,7 @@ fn worker_trunner(mut p pool.PoolProcessor, idx int, thread_id int) voidptr {
 	}
 	ts.benchmark.step()
 	tls_bench.step()
-	if abs_path in ts.skip_files {
+	if !ts.build_tools && (abs_path in ts.skip_files) {
```
ttytm, line 546:
```suggestion
	if !ts.build_tools && abs_path in ts.skip_files {
```
v fmt might warn about this one

lcheylus: Thanks for your review, forgot to check my code with `v fmt`

lcheylus: Thanks for your review, forgot to check my code with `v fmt`

spytheman (approved):

PR #21127, opened 2024-03-28
stbi: update `thirdparty/stb_image` to latest commit `ae721c5`

ttytm:
Update stbi to latest commit https://github.com/nothings/stb/tree/ae721c50eaf761660b4f90cc590453cdb0c2acd0.

Vab currently misses a cflag for andorid compilation to work.
Instead of making it a requirement to add it manually when compiling or running CI, this PR to adds it directly into vab: https://github.com/vlang/vab/pull/278. 

Tested on a fork with a temporary change to use the vab fork of the linked PR above
```diff
# vab_ci.yml
# ...
# Line 28
      - name: Install vab
        run: |
-         .github/workflows/retry.sh v install vab
+         .github/workflows/retry.sh v install https://github.com/ttytm/vab@update-stbi-cflags
```
Result: https://github.com/ttytm/v/actions/runs/8469154352

So the only requirement should be merging https://github.com/vlang/vab/pull/278 then re-running the full VAB workflow should be green.


thirdparty/stb_image/stb_image.h
```diff
@@ -365,11 +366,6 @@ RECENT REVISION HISTORY:
 //    default this is set to (1 << 24), which is 16777216, but that's still
 //    very big.
 
-#ifdef __TINYC__
-#define STBI_NO_SIMD
-#define STBI_NO_THREAD_LOCALS
-#endif
```
spytheman, line 371: This was important for compiling with tcc on macos. I am not sure, if it is needed anymore 🤔, but imho it better stay for now.

spytheman: Just checked, it is still needed:
image

... while with fd78504:
image

ttytm: Nice, thanks for the review!

This would work. 
```
v -gc none -cc tcc -showcc -cflags -mfpu=neon-vfpv4 vlib/stbi/stbi_test.v
```

Is there a chance to intrinsically pass that flag on macos with tcc to use the lib in more comprehensively? Similar to what I do in the vab PR? Then the vab failure would fix itself without a change required there.

spytheman: No, it does not, it still needs the patches:
image

With them, it works with and without `-mfpu=neon-vfpv4`.

spytheman: Anyway, the benefit of tcc on all platforms, is not optimizations, but fast compilation, and I prefer to keep it working, rather than make the header lib more fiddly to deal with.

spytheman: I do not like the vab PR that makes it pass that specific option, since it is needed only for some targets (some Androids are not ARMs), but passed for all, and then we rely on the C compiler to ignore it, instead of error, which is brittle (such things change between compiler versions).

ttytm: Yes I wanted a quick fix a little too quick for this. Gonna update the vab PR when I have a free moment

Issue #21123, opened 2024-03-27
cannot return an option from function that returns alias to `Option` as `Result`


MCausc78:
### Describe the bug

<title>

### Reproduction Steps

```v
struct Foo {
	x int
}

type Bar = ?Foo

enum Baz {
	nothing
	bad
	value
}

fn f(b Baz) !Bar {
	match b {
		.nothing {
			return none
		}
		// .bad { return error('hmm?') }
		.value {
			return Foo{
				x: 42
			}
		}
		else {
			return none
		}
	}
}

fn main() {
}
```

### Expected Behavior

either checker error, or successful compile

### Current Behavior

```
==================
C:/Users/mclr/AppData/Local/Temp/v_0/test.01HT0PV8ZMF3XE50KJ9J44T7RM.tmp.c:6922: warning: implicit declaration of function 'tcc_backtrace'
C:/Users/mclr/AppData/Local/Temp/v_0/test.01HT0PV8ZMF3XE50KJ9J44T7RM.tmp.c:13162: error: cannot convert 'struct _option_main__Foo' to 'struct _result_main__Bar'
...
==================
(Use `v -cg` to print the entire error message)

builder error: 
==================
C error. This should never happen.

This is a compiler bug, please report it using `v bug file.v`.

https://github.com/vlang/v/issues/new/choose

You can also use #help on Discord: https://discord.gg/vlang
```

### V version

V 0.4.5 f6a00fa

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.5 57a33c3.f6a00fa
OS: windows, Microsoft Windows 11 Pro v22000 64-bit
Processor: 12 cpus, 64bit, little endian, 

getwd: D:\Games\Proekti\V\interactions
vexe: D:\Games\Proekti\V\v\v.exe
vexe mtime: 2024-03-27 19:39:55

vroot: OK, value: D:\Games\Proekti\V\v
VMODULES: OK, value: C:\Users\mclr\.vmodules
VTMP: OK, value: C:\Users\mclr\AppData\Local\Temp\v_0

Git version: git version 2.37.0.windows.1
Git vroot status: weekly.2024.09-133-gf6a00fa5
.git/config present: true

CC version:
thirdparty/tcc status: thirdparty-windows-amd64 a39eb79b
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
[...]

PR #21128, opened 2024-03-28
 checker: disallow using aliases of ?Type as !Type

Delta456:
Fixes https://github.com/vlang/v/issues/21123


spytheman (changes requested):
That should be a checker error, not a parser one.
The parser does not have complete type info (the alias may be defined in another file, in the same module), while the checker does.

spytheman:
(rebased over master)

PR #21129, opened 2024-03-28
cgen: fix C struct sumtype support

### LINKED ISSUE #21122: Sumtypes used as C union types not working
### Describe the bug

I'm trying to do the uing bindings for vlang, but i don't want to cast all the controls to C.uiControl everything those are boxed. So i though that using a sumtype would do the job, but seems like this doesn't play well with C types.

The situation is, that doing

```c
struct C.uiControl {}
type C.uiControl = C.uiButton | C.uiLabel | C.uiBox | C.uiControl
```
i get:
```sh
uing/uing.c.v:54:1: error: sum type `C.uiControl` must begin with capital letter
uing/uing.c.v:54:55: error: sum type cannot hold itself
```

if i defined like this:

```c
struct C.uiControl {}
type CiControl = C.uiButton | C.uiLabel | C.uiBox | C.uiControl
```

i get this
```sh
uing/box.v:12:26: warning: unknown type `C.Control` (all virtual C types must be defined, this will be an error soon).
uing/box.v:14:30: error: cannot cast `C.Control` to `&C.uiControl`
main.v:10:13: error: cannot use `&C.uiLabel` as `C.Control` in argument 1 to `uing.Box.append`
```

### Reproduction Steps

```sh
$ git clone https://github.com/trufae/v-uing.git
$ cd v-uing
$ git checkout sumtype
$ make
```

### Expected Behavior

building and running withuot segfaults

### Current Behavior

```sh
$ make
v -o main .
uing/box.v:12:26: warning: unknown type `C.Control` (all virtual C types must be defined, this will be an error soon).
Did you mean `C.uiControl`?
   10 | }
   11 |
   12 | pub fn (b &Box) append(c &C.Control, pos int) {
      |                          ~~~~~~~~~~
   13 |     C.uiBoxAppend(&C.uiBox(b), &C.uiControl(c), pos)
   14 | }
main.v:10:13: error: cannot use `&C.uiLabel` as `&C.Control` in argument 1 to `uing.Box.append`
    8 |     box := uing.new_vbox()
    9 |     // box.append(&C.uiControl(label), 1)
   10 |     box.append(label, 1)
      |                ~~~~~
   11 |     /*
   12 |     b := uing.new_button('ok')
make: *** [all] Error 1
2$
```

### Possible Solution

Well. for now , i created a branch named "wip" , where i'm just using voidptr, and that makes it work. but i think the language should provide a way to use V sumtypes as an alias for C union types.

_No response_

### V version

V 0.4.5 f6a00fa

```sh
$ v doctor
V full version: V 0.4.5 f6a00fa
OS: macos, macOS, 14.4.1, 23E224

getwd: /Users/pancake/prg/v-uing
vexe: /Users/pancake/prg/v/v
vexe mtime: 2024-03-27 16:22:45

vroot: OK, value: /Users/pancake/prg/v
VMODULES: OK, value: /Users/pancake/.vmodules
VTMP: OK, value: /tmp/v_501

Git vroot status: weekly.2023.33-1470-gf6a00fa5
.git/config present: true

thirdparty/tcc status: thirdparty-macos-arm64 5c1d002f
```
> [!NOTE]
>

felipensp:
Fix #21122

```v
#include "@VMODROOT/foo.h"

struct C.Foo {
	a int
}

struct C.Bar {
	a int
}

type Baz = C.Foo | C.Bar

fn main() {
	mut a := Baz{}
	a = C.Foo{a: 1000}
	dump(a)	
}
```

### DIFF
--- a/vlib/v/tests/c_structs/cstruct.h
+++ b/vlib/v/tests/c_structs/cstruct.h
@@ -20,8 +20,3 @@ typedef struct Foo {
 typedef struct Bar {
     int a;
 } Bar;
-
-typedef union Baz {
-    Foo foo;
-    Bar bar;
-} Baz;
--- a/vlib/v/tests/c_structs/cstruct.h
+++ b/vlib/v/tests/c_structs/cstruct.h
@@ -10,3 +10,18 @@ typedef struct Test2 {
 typedef struct Test1 {
     Test2 a;
 } Test1;
+
+/////
+
+typedef struct Foo {
+    int a;
+} Foo;
+
+typedef struct Bar {
+    int a;
+} Bar;
+
+typedef union Baz {
+    Foo foo;
+    Bar bar;
+} Baz;
--- a/vlib/v/gen/c/auto_eq_methods.v
+++ b/vlib/v/gen/c/auto_eq_methods.v
@@ -72,7 +72,7 @@ fn (mut g Gen) gen_sumtype_equality_fn(left_type ast.Type) string {
 	for typ in info.variants {
 		variant := g.unwrap(typ)
 		fn_builder.writeln('\tif (${left_typ} == ${int(variant.typ)}) {')
-		name := '_${g.get_sumtype_variant_name(variant.typ, variant.sym.cname)}'
+		name := '_${g.get_sumtype_variant_name(variant.typ, variant.sym)}'
 
 		left_arg := g.read_field(left_type, name, 'a')
 		right_arg := g.read_field(left_type, name, 'b')
--- a/vlib/v/gen/c/auto_str_methods.v
+++ b/vlib/v/gen/c/auto_str_methods.v
@@ -465,7 +465,7 @@ fn (mut g Gen) gen_str_for_union_sum_type(info ast.SumType, styp string, typ_str
 			fn_builder.write_string('\t\tcase ${int(typ)}: return ${res};\n')
 		} else {
 			mut val := '${func_name}(${deref}(${typ_name}*)x._${g.get_sumtype_variant_name(typ,
-				sym.cname)}'
+				sym)}'
 			if should_use_indent_func(sym.kind) && !sym_has_str_method {
 				val += ', indent_count'
 			}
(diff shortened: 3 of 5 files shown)


trufae:
ive tested this pr with the uing code and it builds but crashes. the resulting C code looks like this:

```
  uing__Window_set_child(w, HEAP(uing__Control, uing__Box_to_sumtype_uing__Control/*KEK*/(box)));
```

so it seems there's an extra abstraction for V sumtypes but it should just do a plain cast. Because i'm using a V struct as the public interface for those C.uiX structs which makes the public interface V-friendly while keeping the C structs internal to the module. So not sure if that's intended to be used that way

felipensp:
> ive tested this pr with the uing code and it builds but crashes. the resulting C code looks like this:
> 
> ```
>   uing__Window_set_child(w, HEAP(uing__Control, uing__Box_to_sumtype_uing__Control/*KEK*/(box)));
> ```
> 
> so it seems there's an extra abstraction for V sumtypes but it should just do a plain cast. Because i'm using a V struct as the public interface for those C.uiX structs which makes the public interface V-friendly while keeping the C structs internal to the module. So not sure if that's intended to be used that way

Yes, thanks for testing. I'll need to change a bit the sumtype implementation after the major PR about sumtype get merged.

trufae:
still seems to not make my code to work, sumtypes contain the data and the type, while i think for those cases it should just pass the data :? but at least its getting closer. also another issue is that in my case i want Control to be a sumtype for Box, Label, Button.. but at the same time all those Box Label Button types should be castable to C.uiControl. so there's an extra level of indirection in my case..

vlib/v/tests/c_structs/cstruct.h
```diff
@@ -9,4 +9,19 @@ typedef struct Test2 {
...
+} Foo;
+
+typedef struct Bar {
+    int a;
+} Bar;
+
+typedef union Baz {
+    Foo foo;
+    Bar bar;
+} Baz;
```
spytheman, line 27: This is not needed. The `main.Baz` type on the V side, has no relation at all to this C `Baz`, and imho it is confusing to have that union here.

spytheman: Imho remove those lines.

spytheman (approved):
Very good work.

vlib/v/tests/c_structs/cstruct.h
felipensp: OH, I forgot to remove this one.

PR #21134, opened 2024-03-28
fmt: improve module detection when formatting imports

ttytm:
The PR improves vfmts import detection. It makes it obsolete to keep a special clause to handle modules in a `$VMODULES` directory. So vmft now should be able can treat all modules accordingly independent of their location accordingly.

It also fixes scenarios where vmft adds a `src.` prefixes to imports. E.g. it would happen in any regular directory when trying to format a module like the `src_import` that is added to the testdata in this PR.

Running the updated test before the "fix" commit also shows other issues that are resolved with the changes.


vlib/v/fmt/testdata/vmodules/src_import/src/a/aa/import_without_src_expected.vv
```diff
@@ -0,0 +1,4 @@
+// NOTE: A test that follows a directory strucutre and module naming like this one
+// also tests whether vfmt would add a prefix to the import.
+
+import a.ab
```
spytheman, line 4: Why is the `src.` removed at all?

The less vfmt modifies the paths, the better.

spytheman:
(rebased over master)

vlib/v/fmt/testdata/vmodules/src_import/src/a/aa/import_without_src_expected.vv
ttytm: So you say to allow for both without a formatting error?
For usage of the module both paths work.

ttytm: I'd prefer the "one way" philo here when thinking about it.

spytheman: My main concern about it, is preventing oscillations, where it is removed by one run of vfmt, based on a heuristic, and then added in another run of vfmt, based on another heuristic. It seems to me safer to not modify the imports at all/not have special cases for it.

spytheman: > So you say to allow for both without a formatting error?

Imho vfmt should not try to outsmart the person writing the code. It should just format the code as it is written.

spytheman: Removing the special case for the vmodules is very important though, and `src.` is an edge case, that will not be encountered very often, so overall, excellent work.

ttytm: > Personally, I would have preferred ALWAYS stripping `src.`. It is supposed to be a "silent" subdir, same as "modules".

There are edge cases where `src.` is part of the import and required to be kept for the module to work. E.g. importing from a module that is structured like this issue is mentioning https://github.com/vlang/v/issues/19799. We might improve module lookup so that it works without it then we can always strip `src.`.

PR #21137, opened 2024-03-29
checker: fix C struct embedded init fields checking

### LINKED ISSUE #21046: False errors when initializing an embedded C struct that contains a reference.
### Describe the bug

When declaring a `struct C.MyCStruct` with any name, that has a reference inside it, and embedding that struct inside a wrapper, initializing the wrapper doesn't work unless both the MyCStruct field _and_ the inner fields are initialized.

### Reproduction Steps

```v
struct C.MyCStruct {
    data &u8
}

struct MyWrapper {
    C.MyCStruct
}

fn (it C.MyCStruct)wrap() MyWrapper {
    return MyWrapper {
        MyCStruct: it
        data: it.data
    }
}
```

### Expected Behavior

The above code should the following error:
> error: duplicate field name in struct literal: `data`

Removing either field initializer should make it wrap the struct as expected.

### Current Behavior

The incorrect code sample given actually compiles without error, generating the designated initializer `((main__MyWrapper){.MyCStruct = it,})`

What's worse, though, is if you *remove* `MyCStruct` from the struct literal, leaving only the initializer for the `data` field you get this error message:
> error: reference field `MyWrapper.MyCStruct.data` must be initialized (part of struct `C.MyCStruct`)

And if you remove the `data` field, leaving the `MyCStruct: it` field, you get two error messages:
> error: reference field `MyWrapper.data` must be initialized
> error: reference field `C.MyCStruct.data` must be initialized

If the inner type is not a C struct, e.g. replacing `C.MyCStruct` with `MyNormalStruct` then all three of these cases work as expected of course, but it is worth noting that when both fields are removed from the `MyWrapper` struct literal it also gives two errors rather than one, but I can't tell whether that is intentional or not. The errors:
> error: reference field `MyWrapper.data` must be initialized
> error: reference field `MyNormalStruct.data` must be initialized

### Possible Solution

The check that all reference fields were initialized seems to have a bug in it, that only comes up when the embedded struct is a C struct.

I was taking a C struct from my old codebase and giving it some iterator wrappers. The wrappers worked, but I had to write this seemingly incorrect struct literal to make it work!

### V version

V 0.4.4 549654a

V full version: V 0.4.4 386bd77.549654a

getwd: C:\Users\Owner
vexe mtime: 2024-03-17 05:49:07

Git vroot status: weekly.2024.11-39-g549654ae
.git/config present: true

thirdparty/tcc: N/A
> [!NOTE]
>

felipensp:
Fix #21046

```v
#include "@VMODROOT/foo.h"

struct C.MyCStruct {
	data &u8
}

struct MyWrapper {
	C.MyCStruct
}

fn (it C.MyCStruct) wrap() MyWrapper {
	return MyWrapper{
		//MyCStruct: it
		data: it.data
	}
}

fn main() {
	dump(C.MyCStruct{data: 123}.wrap())
}
```


spytheman:
Can you add a test please?

felipensp:
> Can you add a test please?

Done.

spytheman (approved):
Excellent work.

PR #21138, opened 2024-03-29
vet: skip testdata dirs

ttytm:
Same reason as sated in the block of the code for the `/tests/` directory. 
Currently even e.g. `_input.vv` a testdata dir are vetted.
It will allow to extend automatic code vetting on actual module code.

> ```
> // skip all /tests/ files, since usually their content is not
> // important enough to be documented/vetted, and they may even
> // contain intentionally invalid code.
> ```

### DIFF
--- a/cmd/tools/vvet/vvet.v
+++ b/cmd/tools/vvet/vvet.v
@@ -96,7 +96,8 @@ fn main() {
 
 // vet_file vets the file read from `path`.
 fn (mut vt Vet) vet_file(path string) {
-	if !vt.opt.is_force && (path.contains('/tests/') || path.contains('/slow_tests/')) {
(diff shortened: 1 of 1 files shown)


JalonSolov:
By that reasoning, all `test*` dirs should probably be skipped.

`tests`
`testdata`
`test_data`
`test data`
etc., etc.

Different people choose different names for places to store tests.

ttytm:
Some dirs containing `test` might be valid modules in project e.g. `testing` in v-analyzer. So I think we should be a bit cautious here. `testdata` is a special dir for V already and excluded for several cases.

ttytm:
After this PR is resolved, I'll submit a next change to allow to overwrite default exclusions so that users can run vet over test and testdata dirs if explicitly specifying them. Personally I see value in this as well.

spytheman (approved):

PR #21139, opened 2024-03-29
vet: optimize performance for path detection when vetting files

ttytm:
Uses a single instead of three consecutive loops for vetting files.


spytheman (approved):
Excellent work.

PR #21140, opened 2024-03-29
fmt: use fixed size array for max_len const

ttytm:
Makes minor code vetting / optimization in `fmt/fmt.v`. An array with a max len for a const named `max_len`.

### DIFF
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -12,7 +12,7 @@ import v.pref
 const bs = '\\'
 
 // when to break a line depending on the penalty
-const max_len = [0, 35, 60, 85, 93, 100]
+const max_len = [0, 35, 60, 85, 93, 100]!
 
 @[minify]
 pub struct Fmt {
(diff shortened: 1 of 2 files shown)


vlib/v/fmt/struct.v
```diff
@@ -330,7 +330,7 @@ pub fn (mut f Fmt) struct_init(node ast.StructInit) {
 				if single_line_fields && (init_field.comments.len > 0
 					|| init_field.next_comments.len > 0
 					|| !expr_is_single_line(init_field.expr)
-					|| f.line_len > max_len.last()) {
+					|| f.line_len > fmt.max_len[fmt.max_len.len - 1]) {
```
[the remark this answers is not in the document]
ttytm: After realizing I pushed unformatted, vfmt even removes the prefix here.

vlib/v/fmt/fmt.v
```diff
@@ -1696,7 +1696,8 @@ pub fn (mut f Fmt) sum_type_decl(node ast.SumTypeDecl) {
 	for variant in variants {
 		// 3 = length of ' = ' or ' | '
 		line_length += 3 + variant.name.len
-		if line_length > fmt.max_len.last() || (variant.id != node.variants.len - 1
+		if line_length > fmt.max_len[fmt.max_len.len - 1]
```
spytheman, line 1699: Imho store that `fmt.max_len[fmt.max_len.len - 1]` into another constant. There is no need to calculate it each time.

ttytm: Good point

spytheman (approved):
Excellent work, the new const names are much clearer.

PR #21141, opened 2024-03-29
checker: detect redundant ref when assigning call expr with ref return

ttytm:
Fixes:
```v
@[heap]
struct Foo {}

fn foo() &Foo {
	return &Foo{}
}

_ := &foo()
```
throwing:
```
error: lvalue expected
...
==================

builder error: 
==================
```


spytheman (approved):
Excellent work.

PR #21142, opened 2024-03-29
vet: allow to overwrite excluded dirs

ttytm:
Allows to run vet over paths that include e.g. `tests` or `testdata` when explictly passing them. Currently, they will be excluded without any information.

```sh
# Current
v vet vlib/v/fmt/testdata # Silently skipped

# With changes
v vet vlib/v/fmt/testdata # Vetted when explicitly specified
v vet vlib/v/fmt # Subdir /testdata will skipped
```

This feature also removes the need for keeping a flag that was used for vet development / test purposes.

E.g. checking out the first commit that just removes the flag before extending the functionality and running below will show the problems solved.
```
v cmd/tools/vvet/vet_test.v
```

### DIFF
--- a/cmd/tools/vvet/vet_test.v
+++ b/cmd/tools/vvet/vet_test.v
@@ -34,8 +34,7 @@ fn check_path(vexe string, dir string, tests []string) int {
 	for path in paths {
 		program := path
 		print(path + ' ')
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #21144, opened 2024-03-29
checker: fix multi return using nil and voidptr

### LINKED ISSUE #17343: multi_struct redefinition when returning !(int, voidptr) result in an or block
### Describe the bug

when returning two values (an `int`, `u32` or any and a `voidptr`) in an or block, a second  multi_return struct is created with the same values (in this case `multi_return_int_voidptr`) wich causes a redefinition error.

### Expected Behavior

the multi_return struct should not be recreated

### Current Behavior

tcc error:
`error: struct/union/enum already defined`

### Reproduction Steps

~~~v
fn ret_int_ptr(arg int) !(int, voidptr) {
	if arg < 0 {
		return error('argument is smaller then zero')
	}
	return 1, [1, 2, 3].data
}

mut val1 := 0
mut val2 := unsafe { nil }

val1, val2 = ret_int_ptr(0) or {
	println(err)
	val1, val2
}
~~~

### Possible Solution

probably a compiler bug.

in the example, changing `val2` to an int type resolves the error, it just creates a warning because of the pointer to int conversion.

### V version

V 0.3.3 e066d1d

OS: linux, "Manjaro Linux"
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
CC version: cc (GCC) 12.2.0

getwd: /home/uhrbaan/Documents/code/vulkan/learn_vk
vmodules: /home/uhrbaan/.vmodules
vroot: /home/uhrbaan/vlang/v
vexe: /home/uhrbaan/vlang/v/v
vexe mtime: 2023-02-16 19:20:09
is vroot writable: true
is vmodules writable: true
V full version: V 0.3.3 ef5be22.e066d1d

Git version: git version 2.39.0
Git vroot status: weekly.2023.07-14-ge066d1d3-dirty
.git/config present: true

felipensp:
Fix #17343

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -3855,7 +3855,12 @@ fn (mut c Checker) ident(mut node ast.Ident) ast.Type {
 fn (mut c Checker) concat_expr(mut node ast.ConcatExpr) ast.Type {
 	mut mr_types := []ast.Type{}
 	for mut expr in node.vals {
-		mr_types << c.expr(mut expr)
+		mut typ := c.expr(mut expr)
+		if typ == ast.nil_type {
+			// nil and voidptr produces the same struct type name
+			typ = ast.voidptr_type
+		}
+		mr_types << typ
 	}
 	if node.vals.len == 1 {
 		typ := mr_types[0]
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #21146, opened 2024-03-29
parser, checker: move error handling for `any` type to the checker to resolve parsing issues

ttytm:
This can be a checker error. The type error during parsing causes parsing interruption and internal errors. E.g.:

```v
// any.v
type Any = any
```
```sh
v fmt any.v
```
```
error: cannot use `any` type here
    1 | type Any = any
      |            ~~~

Internal vfmt error while formatting file: any.v.
```

similar to how `type MyType = unknown_type` is treated


spytheman (approved):
Excellent work.

JalonSolov:
V still has `any`?  I thought it was decided that that was going away...  <sigh>

PR #21147, opened 2024-03-30
parser, checker: move error handling for user defined type duplicates to the checker to resolve parsing issues

ttytm:
Moves toward a checker error for user declared type duplicates. I think this similar too #21146. An error that has less to do with actual parsing.

It allows to resolve internal errors with e.g., vfmt and it will include more error details / other errors that the checker detected as it won't interrupt.


spytheman (approved):
Excellent work.

PR #21148, opened 2024-03-30
fmt: don't change paths when formatting imports

ttytm:
@spytheman your point made here https://github.com/vlang/v/pull/21134#discussion_r1544440533  confirms itself.

With our current module lookup there are more scenarios where it is required to keep the a `src.` part in the import. E.g. in v-analyzer for `build.vsh` (`src.metadata`) or in `tools/project-checker.v`. By removing the `src.` prefix vfmt would break the functionality, since the imported modules are not found anymore. 
Now vfmt would not try to "outsmart" anymore. And for scenarios where an import prefixed with `src.` and as well as the path without `src.` are valid, both will be also seen as valid by vfmt. 

General formatting will still happen ofc, unused import symbols will be cleared, just the path of the import won't be changed anymore. 

The tests that check for the import path update were removed. All other behavior is tested due to older and recently added tests, the changes should be covered.

### DIFF
--- a/.github/workflows/v_apps_and_modules_compile_ci.yml
+++ b/.github/workflows/v_apps_and_modules_compile_ci.yml
@@ -107,7 +107,7 @@ jobs:
           # v -cc gcc run /tmp/gitly/tests/first_run.v
           # # /tmp/gitly/gitly -ci_run
 
-      - name: Build V Language Server (v-analyzer) vlang/v-analyzer
+      - name: Build vlang/v-analyzer
         run: |
           echo "Clone v-analyzer"
           .github/workflows/retry.sh git clone --depth=1 --filter=blob:none --recursive --shallow-submodules https://github.com/vlang/v-analyzer /tmp/v-analyzer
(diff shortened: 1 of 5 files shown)


ttytm:
To further improve test coverage and to better prevent regressions, I would add formatting check over an official extensive project like v-analyzer to Vs regular CI.

ttytm:
CI step added isolated on a separate branch - to verify it would recognize formatting issues https://github.com/ttytm/v/actions/runs/8489249682/job/23259136533

.github/workflows/v_apps_and_modules_compile_ci.yml
```diff
@@ -119,6 +119,18 @@ jobs:
...
           v build.vsh release
 
+      - name: Format vlang/v-analyzer
+        run: |
+          cd /tmp/v-analyzer
+          set +e
+          v fmt -c .
+          exit_code=$?
+          if [[ $exit_code -ne 0 && $exit_code -ne 5 ]]; then
+            # Don't fail on on internal errors
```
spytheman, line 129:
```suggestion
            # Don't fail on internal errors
```

spytheman (approved):

Issue #9507, opened 2021-03-29
[Bug] `println` refuses to print large number


guekho64:
# What did you do?

``` v

module main

fn main () {
 println(i64(0x8000000000000000))
}

```

# What did you expect to see?

``` v

-9223372036854775808

```

# What did you see instead?

Literally nothing

# Additional Info

* No warnings / errors when executing with `v run Test.v` nor `v Test.v && ./Test`
* Tested against `Weekly V 0.2.2 7664964` and `Stable V 0.2.2 46cdf4f`
* Using `Ubuntu 20.04.2 LTS -> Kernel: 5.8.0-48-lowlatency`
* Possibly related to #3751 (Although this bug persists even when executing a successfully compiled binary)

Carpall:
I'm not a fan of vlang, I found this issue in VLang Memes (wip), so I don't known if vlang compiles also to LLVM Bitcode, anyway the problem is linked to native version of binary, it seems to work perfectly in playground with wasm interpreter (executed in msedge)
![image](https://user-images.githubusercontent.com/65513900/113178695-1ce87480-924f-11eb-91f0-c2264f02fb5a.png)

guekho64:
Apparently, you've just discovered a new/related bug: [Playground](https://v-wasm.now.sh/) seems to be broken as well, probably even more than native `v`. Haven't checked on Windows, but this doesn't look promising...

These should be the correct results:

``` v

int(0xffffffff) == -1
i64(0x8000000000000000) == -9223372036854775808

```

![image](https://user-images.githubusercontent.com/8020999/113322446-edc71700-92d1-11eb-918b-1e56afd87f81.png)

ArtemkaKun:
Actual V (0.3.2):

```v
module main

fn main() {
	println(i64(0x8000000000000000) == -9223372036854775808)
}
```

```
true
```

Fixed

PR #21152, opened 2024-03-31
toml: update https://github.com/alexcrichton/toml-rs.git to https://github.com/toml-lang/toml-test.git, record new exceptions

spytheman:


spytheman:
@Larpon, sorry for the noise. I realized too late, that the successor repo for alexcrichton, can be used directly, since it also contains the commit 499e8c4, which we need.

I think the PR should be stable enough for review now.

Larpon (commented):
Hi @spytheman thanks for taking care of this. The changes look fine to me - the CI job definitely looks a lot nicer now. Did you have a specific reason to not rename burntsushi/alexcrihton, if those tests are now the official tests?

I only have one issue locally:

All separate tests (`alexcrichton_toml_rs_test.v`, `burntsushi_toml_test.v` `iarna_toml_spec_test.v` and `large_toml_file_test.v`) pass on my system (EndeavourOS Arch/rolling) when run *separately*, but running:
```bash
  VTEST_TOML_DO_YAML_CONVERSION=1 VTEST_TOML_DO_LARGE_FILES=1 ./v test vlib/toml
  ```
...  locally gives me this:
image

Note the `e` vs. `E` I'm not sure if that is because I'm on newer versions of the software stack or what :thinking: 

I remember we had to do some shoehorning / tweaking of some test data/results because of the YAML conversions I'm not sure if this might be one of those :thinking:

Larpon:
image

From https://github.com/vlang/v/pull/21152/files#diff-f9e436e6d2d57cf88c41f996f2c4be49302cc0559cd810f254cc7fcb7536bae1R343-R391

So it might be a tool update that might have fixed/changed the scientific notation.

Also seems we've closed the Inf special case: https://github.com/vlang/v/issues/9507
But it's probably better to look into those in  a separate clean up

spytheman:
I could not get the problem to show up locally @Larpon , and it seems it is also passing on the CI. I suspect that the used `jq` version may be the culprit. My local copy is `jq-1.6`.

spytheman:
> Did you have a specific reason to not rename burntsushi/alexcrihton, if those tests are now the official tests?

Just minimization of the commit diff. I did not want to have both renames and code changes in the same PR.

PR #21155, opened 2024-03-31
vweb, x.vweb: edit some descriptions

kimshrier:
I noticed that the bullet points under the Features heading on both vweb and x.vweb were not formatting right due to the line for "Easy to deploy" being continued on a second line.

It seems like that would be a bug in vdoc but I was unable to find it.  The expedient fix is to just have that bullet point all on one line.

I also re-worded some other descriptive text to make it easier (at least for me) to understand.

I am not sure of the status of x.vweb vs. vweb.  Using `v new --web` generates files referencing vweb so I assume that is still the preferred version.

### DIFF
--- a/vlib/vweb/README.md
+++ b/vlib/vweb/README.md
@@ -20,8 +20,7 @@ in the browser. No need to quit the app, rebuild it, and refresh the page in the
 
 - **Very fast** performance of C on the web.
 - **Small binary** hello world website is <100 KB.
-- **Easy to deploy** just one binary file that also includes all templates. No need to install any
-  dependencies.
+- **Easy to deploy** just one binary file that also includes all templates. No need to install any dependencies.
 - **Templates are precompiled** all errors are visible at compilation time, not at runtime.
 - **Multithreaded** by default
 
@@ -69,7 +68,7 @@ fn new_app() &App {
 
 @['/']
 pub fn (mut app App) page_home() vweb.Result {
-	// all this constants can be accessed by src/templates/page/home.html file.
+	// all these constants can be accessed by src/templates/page/home.html file.
 	page_title := 'V is the new V'
 	v_url := 'https://github.com/vlang/v'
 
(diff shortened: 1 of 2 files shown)


kimshrier:
Looks like CI doesn't like the length of the "Easy to deploy" line.  So the real fix is to track down the problem in vdoc.

If the other changes are deemed useful, I'll modify the merge request without the changes to the "Easy to deploy" line.

vlib/vweb/README.md
```diff
@@ -20,8 +20,7 @@ in the browser. No need to quit the app, rebuild it, and refresh the page in the
 
 - **Very fast** performance of C on the web.
 - **Small binary** hello world website is <100 KB.
-- **Easy to deploy** just one binary file that also includes all templates. No need to install any
-  dependencies.
+- **Easy to deploy** just one binary file that also includes all templates. No need to install any dependencies.
```
spytheman, line 23: You can wrap the line, so that the 2 pieces individually will be < 100 characters long.

spytheman:
For me on master, github shows:
image
(from https://github.com/vlang/v/blob/master/vlib/vweb/README.md)

While module.vlang.io shows:
image
(from https://modules.vlang.io/vweb.html)

spytheman:
It is formatted like this, because of this:
image

spytheman:
> I am not sure of the status of x.vweb vs. vweb. 

Afaik, `x.vweb` is intended to move to `veb`, and `vweb`, will stay for a while too, then it will be moved eventually to `https://github.com/vlang/vweb` as a module that is outside of vlib, and installable with `v install vweb`, for people that still depend on it.

spytheman:
The markdown parser that vdoc uses for the .md files is here: https://github.com/vlang/markdown . It is a wrapper of the https://github.com/mity/md4c library .

spytheman:
Imho what we can do for now, without a fix to the markdown parser, is move the problem item that needs wrapping, at the end of the list.

PR #21156, opened 2024-04-01
parser: fix case of falsely registering imports as used, remove unused imports

ttytm:
Fixes e.g,
```v
import term as te
import te.ui // alias `te` as part of another import would register the import `term` as used, despite its never being used.

fn main() {}
```
The change reveals a lot of unused imports that are removed in this PR as a requirement to pass CI.

### DIFF
--- a/examples/pendulum-simulation/modules/sim/anim/worker.v
+++ b/examples/pendulum-simulation/modules/sim/anim/worker.v
@@ -1,7 +1,6 @@
 module anim
 
 import benchmark
-import sim
 import sim.img
 
 fn pixels_worker(mut app App) {
--- a/examples/sokol/fonts.v
+++ b/examples/sokol/fonts.v
@@ -1,4 +1,3 @@
-import sokol
 import sokol.sapp
 import sokol.gfx
 import sokol.sgl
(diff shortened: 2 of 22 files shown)


JalonSolov:
In this particular case, I'd think it would report both `term` and `term.ui` as being unused.

Will that happen with this change?

ttytm:
There is a test added that correctly recognizes them as not being used

```v
import term as te
import te.ui
```

```
vlib/v/checker/tests/unused_import_err.vv:1:8: warning: module 'te (term)' is imported but never used
    1 | import term as te
      |        ~~~~
    2 | import te.ui
vlib/v/checker/tests/unused_import_err.vv:2:8: warning: module 'ui (te.ui)' is imported but never used
    1 | import term as te
    2 | import te.ui
      |        ~~~~~
vlib/v/checker/tests/unused_import_err.vv:2:1: builder error: cannot import module "te.ui" (not found)
    1 | import term as te
    2 | import te.ui
      | ~~~~~~~~~~~~
```

ttytm:
Specifying an alias for an import should be imo detached from using the alias in other imports to import submodules.

spytheman (approved):
Good work.

PR #21160, opened 2024-04-01
breaking,vlib: update handling of imports whose symbols are not directly used in imported file, remove `pub const is_used = 1` workarounds

ttytm:
The changes move from using a workaround with public constants to a less verbose, imho more elegant way to handle imports whose symbols are not directly used (a case mostly for our C wrappers).

This will also remove the "workaround-constants" - that might be confusing - from module documentations. E.g. json, mbedtls and several others
|||
| -- | -- |
| Screenshot_20240401_184439 | Screenshot_20240401_184446 |

Changes are covered by our current test suite.

### DIFF
--- a/examples/sokol/drawing.v
+++ b/examples/sokol/drawing.v
@@ -1,4 +1,3 @@
-import sokol
 import sokol.sapp
 import sokol.gfx
 import sokol.sgl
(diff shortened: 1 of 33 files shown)


spytheman (approved):
Excellent work.

PR #21163, opened 2024-04-01
x.json2: add ability to decode arrays

### LINKED ISSUE #21150: x.json2 can not decode properly a struct with a fixed array field
**V doctor:**
```
V full version: V 0.4.5 872bcbc.13a2efd
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-03-30 15:59:29

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.13-57-g13a2efd6
.git/config present: true

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg bug_fixed_array_json2.v`
```v
import x.json2
struct Pos { pos [3]f64 }
input := '{"pos": [0.5, 0.0, 0.5]}'
p := json2.decode[Pos](input)!
dump(p)

```

the pos field should be: `pos: [0.5, 0.0, 0.5]`

**What did you see instead?**
```
[bug_fixed_array_json2.v:5] p: Pos{
    pos: [0.0, 0.0, 0.0]
}
```

The equivalent program, using `json` works:
```v
import json
struct Pos { pos [3]f64 }
input := '{"pos": [0.5, 0.0, 0.5]}'
p := json.decode(Pos, input)!
dump(p)
```
producing:
```
[bug_fixed_array_json_working.v:5] p: Pos{
    pos: [0.5, 0.0, 0.5]
}
```
> [!NOTE]
>

ttytm:
Fixes #21150

Adds the ability to decode arrays, as currently it is an empty block:
- Current master
  https://github.com/vlang/v/blob/5da1fccaed6ca699ecce79018b29b123c3e19969/vlib/x/json2/decoder.v#L251-L252

The feature is integrated in a simple enough manner that it should create little interference with #21039 . It might even help with a TODO there.

### DIFF
--- a/vlib/x/json2/decoder.v
+++ b/vlib/x/json2/decoder.v
@@ -249,6 +249,39 @@ fn decode_struct[T](_ T, res map[string]Any) !T {
 						typ.$(field.name) = res[json_name]!.to_time()!
 					}
 				} $else $if field.is_array {
+					arr := res[field.name]! as []Any
+					match typeof(typ.$(field.name)).name {
+						'[]bool' { typ.$(field.name) = arr.map(it.bool()) }
+						'[]?bool' { typ.$(field.name) = arr.map(?bool(it.bool())) }
+						'[]f32' { typ.$(field.name) = arr.map(it.f32()) }
+						'[]?f32' { typ.$(field.name) = arr.map(?f32(it.f32())) }
+						'[]f64' { typ.$(field.name) = arr.map(it.f64()) }
+						'[]?f64' { typ.$(field.name) = arr.map(?f64(it.f64())) }
+						'[]i8' { typ.$(field.name) = arr.map(it.i8()) }
+						'[]?i8' { typ.$(field.name) = arr.map(?i8(it.i8())) }
+						'[]i16' { typ.$(field.name) = arr.map(it.i16()) }
+						'[]?i16' { typ.$(field.name) = arr.map(?i16(it.i16())) }
+						'[]i64' { typ.$(field.name) = arr.map(it.i64()) }
+						'[]?i64' { typ.$(field.name) = arr.map(?i64(it.i64())) }
+						'[]int' { typ.$(field.name) = arr.map(it.int()) }
+						'[]?int' { typ.$(field.name) = arr.map(?int(it.int())) }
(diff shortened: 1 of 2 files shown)


Delta456:
Isn't there a more efficient way to do this? It feels too lengthy to me.

ttytm:
I hope there is one but didn't found one yet. Unfortunately there is mess with accessing the comptime of the array. Had to integrate a similar approach in toml back then as last option :/.

https://github.com/vlang/v/blob/c086bee5bebb8b9374605fc65469f629d42ae18d/vlib/toml/toml.v#L59-L87

If there is none, I hope there will be one soon.

Delta456:
> I hope there is one but didn't found one yet. Unfortunately there is mess with accessing the comptime type of the array. Had to integrate a similar approach in toml back then as last option :/. There it is now used for arrays and maps.
> 
> https://github.com/vlang/v/blob/c086bee5bebb8b9374605fc65469f629d42ae18d/vlib/toml/toml.v#L59-L87
> 
> If there is none, I hope there will be one soon.

I think we need to support checking the array in a much functional way.

enghitalo:
Very good initiative!! But I think this will cause a lot of headaches. In the end, the biggest problem today would be finding a good solution to decode Sumtypes (This will probably have to be done more "manually/handled" by the user, calling some method). After that, array decodes became much simpler.

enghitalo:
I can't wait to have fun again with the new json2 decode. I hope that within the next few months I will have time to finish this project.

ttytm:
> Very good initiative!! But I think this will cause a lot of headaches. In the end, the biggest problem today would be finding a good solution to decode Sumtypes (This will probably have to be done more "manually/handled" by the user, calling some method...). After that, array decodes became much simpler.

The change just adds decoding for all the types that the current type is decoding. Sumtypes might be in their own scope.

spytheman (approved):
Excellent work.

spytheman:
@enghitalo JSON does not have sumtypes. JSON does have arrays.

-> For the completion of `x.json2`, first class support for arrays is much more important.

PR #21164, opened 2024-04-01
net.http: fix panic on empty addr, cleanup `listen_and_serve` Server method

ttytm:
Fixes a panic when the server gets an empty address. Does a cleanup to the related scope.

```v
import net.http

mut server := &http.Server{
	addr: ''
	on_running: fn (mut server http.Server) {
		spawn fn [mut server] () {
			println('server started')
			server.stop()
			println('server stopped')
		}()
	}
}
server.listen_and_serve()
```

Instead it will now behave like before passing 0 to the deprecated `port` field, or `:0` to `addr` - using a random port.

### DIFF
--- a/vlib/net/http/server.v
+++ b/vlib/net/http/server.v
@@ -62,10 +62,10 @@ pub fn (mut s Server) listen_and_serve() {
 		return
 	}
 	if l.family() == net.AddrFamily.unspec {
-		listen_addr := if s.addr == '' || s.addr == ':0' { 'localhost:0' } else { s.addr }
-		listen_fam := net.AddrFamily.ip
-		// listen_fam := $if windows { net.AddrFamily.ip } $else { net.AddrFamily.ip6 }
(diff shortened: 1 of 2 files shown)


vlib/net/http/server.v
```diff
@@ -61,16 +61,11 @@ pub fn (mut s Server) listen_and_serve() {
...
-		if listening_address == ':0' {
-			listening_address = 'localhost:0'
-		}
-		mut listen_family := net.AddrFamily.ip
-		//		$if !windows {
-		//			listen_family = net.AddrFamily.ip6
-		//		}
-		s.listener = net.listen_tcp(listen_family, listening_address) or {
+		listen_addr := if s.addr == '' || s.addr == ':0' { 'localhost:0' } else { s.addr }
+		listen_fam := net.AddrFamily.ip
```
spytheman, line 66: Why not use the original names `listen_family` and `listening_address` ?

spytheman (approved):
Excellent work.

PR #21166, opened 2024-04-02
parser: add support for inclusive match range with omited zero `...5`

ttytm:
Add support to omit the `0` for an inclusive range expression. Similar to what is already supported when accessing an `array[..5]`. E.g.,
```v
n := 3
match n {
	...5 {}
	else {}
}

x := `5`
match x {
	...`3` { println(n) }
	else { println('other') }
}

```


Delta456:
This shouldn't work.

enghitalo:
`array[..5]` makes sense because it is about positions `...5` doesn't because before 0 there's infinity numbers

PR #21167, opened 2024-04-02
parser: add support for inclusive match range with omiited zero ...5

ttytm:
Removes obsolete zeros from range expressions. Based on #21166. Includes updates tests that cover the new feature.


spytheman:
I do not get it. Why remove the 0? The result is imho less readable.

Delta456:
> I do not get it. Why remove the 0? The result is imho less readable.

We discussed this in https://github.com/vlang/v/pull/21166 and the PR is now closed.

Let's also close this PR.

PR #21169, opened 2024-04-02
vfmt: remove unused imports

ttytm:
Resolves a TODO for vfmt adding/fixing removal of unused imports. It does so respecting user comments. 
https://github.com/vlang/v/blob/a1b6360da5b5d21da207536f1810abfe34796c3b/vlib/v/fmt/fmt.v#L343-L346

I think `vlib/v/fmt/tests/import_comments_expected.vv` is the best file in the updated test showcasing the behavior.

Includes minor quality improvements. Next step would fixing/extending auto imports.


spytheman:
Automatically removing unused imports is very breaking for prototyping. Imho the warning that is currently present, is more than enough.

spytheman:
Autoimporting is imho also a bad idea, because it is a heuristic, that can not work reliably in all or even most cases, since vfmt works on single files, and thus has no idea, whether the `os.abc` that you wrote somewhere in a file, means that you want to `import os` or `import myproject.os` or `import xyz as os`.

spytheman:
I think, we should at some point, do a cleanup of such `TODO:` items, that may have seemed good on the surface initially, but that turned out to have big negatives revealed later.

ttytm:
Yes it is better to solve this via import completion of a ls. V-analyzer could support it.

ttytm:
@spytheman should we than remove the current auto imports?

PR #21171, opened 2024-04-02
x.json2,checker,toml: allow field.typ compile-time checking with MatchExpr and add array of option checking

### LINKED ISSUE #21170: have a better to way to detect arrays from compile time
https://github.com/vlang/v/blob/c086bee5bebb8b9374605fc65469f629d42ae18d/vlib/toml/toml.v#L59-L87

To currently match arrays via compile time, one has to use `typeof(typ.$(field.name)).name ` which is extremely hacky and inconvenient. There needs to be a better way to check the array type.

`typeof(typ.$(field.name))` should be used instead and match with the appropriate type.
> [!NOTE]
>

felipensp:
Fix #21170
Improve type checking from #21163

### DIFF
--- a/vlib/x/json2/decoder.v
+++ b/vlib/x/json2/decoder.v
@@ -250,36 +250,86 @@ fn decode_struct[T](_ T, res map[string]Any) !T {
 					}
 				} $else $if field.is_array {
 					arr := res[field.name]! as []Any
-					match typeof(typ.$(field.name)).name {
-						'[]bool' { typ.$(field.name) = arr.map(it.bool()) }
-						'[]?bool' { typ.$(field.name) = arr.map(?bool(it.bool())) }
-						'[]f32' { typ.$(field.name) = arr.map(it.f32()) }
-						'[]?f32' { typ.$(field.name) = arr.map(?f32(it.f32())) }
-						'[]f64' { typ.$(field.name) = arr.map(it.f64()) }
-						'[]?f64' { typ.$(field.name) = arr.map(?f64(it.f64())) }
-						'[]i8' { typ.$(field.name) = arr.map(it.i8()) }
-						'[]?i8' { typ.$(field.name) = arr.map(?i8(it.i8())) }
-						'[]i16' { typ.$(field.name) = arr.map(it.i16()) }
-						'[]?i16' { typ.$(field.name) = arr.map(?i16(it.i16())) }
-						'[]i64' { typ.$(field.name) = arr.map(it.i64()) }
-						'[]?i64' { typ.$(field.name) = arr.map(?i64(it.i64())) }
-						'[]int' { typ.$(field.name) = arr.map(it.int()) }
-						'[]?int' { typ.$(field.name) = arr.map(?int(it.int())) }
-						'[]string' { typ.$(field.name) = arr.map(it.str()) }
-						'[]?string' { typ.$(field.name) = arr.map(?string(it.str())) }
+					match field.typ {
+						[]bool {
+							typ.$(field.name) = arr.map(it.bool())
+						}
+						[]?bool {
+							typ.$(field.name) = arr.map(?bool(it.bool()))
+						}
+						[]f32 {
+							typ.$(field.name) = arr.map(it.f32())
+						}
+						[]?f32 {
+							typ.$(field.name) = arr.map(?f32(it.f32()))
+						}
+						[]f64 {
+							typ.$(field.name) = arr.map(it.f64())
+						}
+						[]?f64 {
+							typ.$(field.name) = arr.map(?f64(it.f64()))
+						}
+						[]i8 {
+							typ.$(field.name) = arr.map(it.i8())
+						}
+						[]?i8 {
+							typ.$(field.name) = arr.map(?i8(it.i8()))
+						}
+						[]i16 {
+							typ.$(field.name) = arr.map(it.i16())
+						}
+						[]?i16 {
+							typ.$(field.name) = arr.map(?i16(it.i16()))
+						}
+						[]i64 {
+							typ.$(field.name) = arr.map(it.i64())
+						}
+						[]?i64 {
+							typ.$(field.name) = arr.map(?i64(it.i64()))
+						}
+						[]int {
+							typ.$(field.name) = arr.map(it.int())
+						}
+						[]?int {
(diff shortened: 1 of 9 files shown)


ttytm:
Awesome! The same can be applied to toml then!

vlib/x/json2/decoder.v
```diff
@@ -250,36 +250,86 @@ fn decode_struct[T](_ T, res map[string]Any) !T {
...
 						// vfmt on
-						'[]u8' { typ.$(field.name) = arr.map(it.u64()) }
-						'[]?u8' { typ.$(field.name) = arr.map(?u8(it.u64())) }
-						'[]u16' { typ.$(field.name) = arr.map(it.u64()) }
-						'[]?u16' { typ.$(field.name) = arr.map(?u16(it.u64())) }
-						'[]u32' { typ.$(field.name) = arr.map(it.u64()) }
-						'[]?u32' { typ.$(field.name) = arr.map(?u32(it.u64())) }
-						'[]u64' { typ.$(field.name) = arr.map(it.u64()) }
-						'[]?u64' { typ.$(field.name) = arr.map(?u64(it.u64())) }
+						[]u8 {
```
ttytm, line 309: The `vfmt off` part was added to prevent the line breaks - thinking of achieving an  overall overview of the handled types. E.g., below code is formatted. But it's opinionated, if adding the line breaks, `vfmt off` can be removed.
```v
match field.typ {
	[]bool { typ.$(field.name) = arr.map(it.bool()) }
	[]?bool { typ.$(field.name) = arr.map(?bool(it.bool())) }
	[]f32 { typ.$(field.name) = arr.map(it.f32()) }
	[]?f32 { typ.$(field.name) = arr.map(?f32(it.f32())) }
	[]f64 { typ.$(field.name) = arr.map(it.f64()) }
	[]?f64 { typ.$(field.name) = arr.map(?f64(it.f64())) }
	[]i8 { typ.$(field.name) = arr.map(it.i8()) }
	[]?i8 { typ.$(field.name) = arr.map(?i8(it.i8())) }
	[]i16 { typ.$(field.name) = arr.map(it.i16()) }
	[]?i16 { typ.$(field.name) = arr.map(?i16(it.i16())) }
	[]i64 { typ.$(field.name) = arr.map(it.i64()) }
	[]?i64 { typ.$(field.name) = arr.map(?i64(it.i64())) }
	[]int { typ.$(field.name) = arr.map(it.int()) }
	[]?int { typ.$(field.name) = arr.map(?int(it.int())) }
	[]string { typ.$(field.name) = arr.map(it.str()) }
	[]?string { typ.$(field.name) = arr.map(?string(it.str())) }
	// NOTE: Using `!` on `to_time()` inside the array method causes a builder error - 2024/04/01.
	[]time.Time { typ.$(field.name) = arr.map(it.to_time() or { time.Time{} }) }
	// vfmt off
	[]?time.Time { typ.$(field.name) = arr.map(?time.Time(it.to_time() or { time.Time{} })) }
	// vfmt on
	[]u8 { typ.$(field.name) = arr.map(it.u64()) }
	[]?u8 { typ.$(field.name) = arr.map(?u8(it.u64())) }
	[]u16 { typ.$(field.name) = arr.map(it.u64()) }
	[]?u16 { typ.$(field.name) = arr.map(?u16(it.u64())) }
	[]u32 { typ.$(field.name) = arr.map(it.u64()) }
	[]?u32 { typ.$(field.name) = arr.map(?u32(it.u64())) }
	[]u64 { typ.$(field.name) = arr.map(it.u64()) }
	[]?u64 { typ.$(field.name) = arr.map(?u64(it.u64())) }
	else {}
}
```

vlib/x/json2/decoder.v
```diff
@@ -250,36 +250,86 @@ fn decode_struct[T](_ T, res map[string]Any) !T {
...
+						[]?i8 {
+							typ.$(field.name) = arr.map(?i8(it.i8()))
+						}
+						[]i16 {
+							typ.$(field.name) = arr.map(it.i16())
+						}
+						[]?i16 {
+							typ.$(field.name) = arr.map(?i16(it.i16()))
+						}
+						[]i64 {
```
JalonSolov, line 284: Looks like `[]i32` is missing...

felipensp:
> what about maps?

It does works.

```v
struct Test {
	a int
	b []int
	c map[int]string
}

fn main() {
	$for f in Test.fields {
		type_name := typeof(f.$(f.name)).name
		match f.typ {
			int { println('1. ${f.name} is ${type_name}') }
			[]int { println('2. ${f.name} is ${type_name}') }
			map[int]string { println('3. ${f.name} is ${type_name}') }
			else {}
		}
	}
}
```

Delta456:
> > what about maps?
> 
> It does works.
> 
> ```v
> struct Test {
> 	a int
> 	b []int
> 	c map[int]string
> }
> 
> fn main() {
> 	$for f in Test.fields {
> 		type_name := typeof(f.$(f.name)).name
> 		match f.typ {
> 			int { println('1. ${f.name} is ${type_name}') }
> 			[]int { println('2. ${f.name} is ${type_name}') }
> 			map[int]string { println('3. ${f.name} is ${type_name}') }
> 			else {}
> 		}
> 	}
> }
> ```

Is there a way to check the key and the value type of a map separately?

felipensp:
> > > what about maps?
> > 
> > 
> > It does works.
> > ```v
> > struct Test {
> > 	a int
> > 	b []int
> > 	c map[int]string
> > }
> > 
> > fn main() {
> > 	$for f in Test.fields {
> > 		type_name := typeof(f.$(f.name)).name
> > 		match f.typ {
> > 			int { println('1. ${f.name} is ${type_name}') }
> > 			[]int { println('2. ${f.name} is ${type_name}') }
> > 			map[int]string { println('3. ${f.name} is ${type_name}') }
> > 			else {}
> > 		}
> > 	}
> > }
> > ```
> 
> Is there a way to check the key and the value type of a map separately?

No yet.

spytheman (approved):

vlib/x/json2/decoder.v
spytheman: Should be fine after a519fa5 .

vlib/x/json2/decoder.v
spytheman: Should be fine after a519fa5 .

PR #21172, opened 2024-04-02
fmt: improve import processing

ttytm:
While deciding not to follow through on removing unused imports (for comprehensible reasons), working on #21169 provided insights allowing for general improvements. Picking the cherries, they should be integrated with this PR. They include better detection of used imports and improvements in terms of performance, quality and maintainability.


spytheman (approved):
Excellent work. Thank you.

Issue #20699, opened 2024-02-01
Incorrect typecheck in sql expressions


koplenov:
**V doctor:**
```
V full version: V 0.4.4 cae40ad
OS: linux, Debian GNU/Linux 12 (bookworm)
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i5-3337U CPU @ 1.80GHz

getwd: /home/koplenov/dev/zone/cjer
vexe: /home/koplenov/dev/v/v
vexe mtime: 2024-01-29 11:59:16

vroot: OK, value: /home/koplenov/dev/v
VMODULES: OK, value: /home/koplenov/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.39.2
Git vroot status: Error: fatal: не найден git репозиторий (или один из родительских каталогов): .git
.git/config present: false

CC version: cc (Debian 12.2.0-14) 12.2.0
thirdparty/tcc: N/A

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg sql_bug.v`
```v
module main

import db.sqlite

pub struct SeoBlock {
    pub:
    seo_f1 string
    seo_is_f1_capitalize bool
    seo_params string
}

fn main() {
	db := sqlite.connect('test.db') or { panic(err) }

    sql db {
        update SeoBlock set seo_f1 = seo_f1 where seo_f1 == ""
    } or { panic(err) }
}

```

**What did you expect to see?**

Correct error in sql expression

**What did you see instead?**
```
================ V panic ================
   module: v.ast
 function: default_table_panic_handler()
  message: sym: invalid type (typ=ast.Type(0x0 = 0) idx=0). Compiler bug. This should never happen. Please report the bug using `v bug file.v`.

     file: /home/koplenov/dev/v/vlib/v/ast/table.v:85
   v hash: cae40ad
=========================================
/tmp/v_1000/../../../../../../home/koplenov/dev/v/vlib/builtin/builtin.c.v:66: at panic_debug: Backtrace
/tmp/v_1000/../../../../../../home/koplenov/dev/v/vlib/v/ast/table.v:85: by v__ast__default_table_panic_handler
/tmp/v_1000/../../../../../../home/koplenov/dev/v/vlib/v/ast/table.v:91: by v__ast__Table_panic
/tmp/v_1000/../../../../../../home/koplenov/dev/v/vlib/v/ast/table.v:679: by v__ast__Table_sym
[...]

PR #21174, opened 2024-04-03
 checker.orm: check if update expr is invalid ident 

Delta456:
Fixes https://github.com/vlang/v/issues/20699


Delta456:
This PR may change according to what the syntax is decided for `set`.

KamkoAmoh (approved):

PR #21091, opened 2024-03-25
v: add `@VMODHASH` comptime variable to store the commit sha of a V module

ttytm:
The PR should implement the possibility to store a commit hash of a vmodule as a compile time variable. Similar to Similar to `@VHASH` / `@VCURRENTHASH` for V itself. The reason for it is that I encountered multiple scenarios where it would be useful and function calls in constants are not yet evaluated at comptime.

PR #21178, opened 2024-04-03
v.utils: update `githash` to be able to get the githash of every passed project

ttytm:
Reworks `githash` to extend its use cases by allowing it to return the githash of any passed path. Currently, it is used for `@VCURRENTHASH` and only returns the path of the V hash. Chances that `githash` was called directly outside the V codebase itself are low, but it is a public function, so this change can be considered breaking.

The changes:
- Simplifies: the function no longer uses a bool param to return a static value if it is set to false; instead, we assign static values directly.
- Uses early returns instead of a less safe workaround with a for loop.
- Utilizes Result values to be able to trace down misbehavior instead of assigning a fallback value directly in the function.

The new functionality will enable an easy implementation of `@VMODHASH` in #21091.

### DIFF
--- a/vlib/v/util/version/version.c.v
+++ b/vlib/v/util/version/version.c.v
@@ -17,7 +17,7 @@ pub fn vhash() string {
 
 pub fn full_hash() string {
 	build_hash := vhash()
-	current_hash := githash(false)
+	current_hash := @VCURRENTHASH
 	if build_hash == current_hash {
 		return build_hash
 	}
@@ -29,8 +29,7 @@ pub fn full_v_version(is_verbose bool) string {
 	if is_verbose {
 		return 'V ${version.v_version} ${full_hash()}'
 	}
-	hash := githash(false)
-	return 'V ${version.v_version} ${hash}'
+	return 'V ${version.v_version} ${@VCURRENTHASH}'
 }
 
 // githash(x) returns the current git commit hash.
@@ -75,6 +74,5 @@ pub fn githash(should_get_from_filesystem bool) string {
 		}
 		break
 	}
-
 	return @VCURRENTHASH
 }
--- a/vlib/v/util/version/version.c.v
+++ b/vlib/v/util/version/version.c.v
@@ -38,26 +38,26 @@ pub fn githash(path string) !string {
 	// .git/HEAD
 	git_head_file := os.join_path(path, '.git', 'HEAD')
 	if !os.exists(git_head_file) {
-		return error('failed to find HEAD in: ${path}')
+		return error('failed to find `${git_head_file}`')
 	}
 	// 'ref: refs/heads/master' ... the current branch name
 	head_content := os.read_file(git_head_file) or {
-		return error('failed to read HEAD in: ${path}')
+		return error('failed to read `${git_head_file}`')
 	}
 	current_branch_hash := if head_content.starts_with('ref: ') {
-		gcbranch_rel_path := head_content.replace('ref: ', '').trim_space()
-		gcbranch_file := os.join_path(path, '.git', gcbranch_rel_path)
+		rev_rel_path := head_content.replace('ref: ', '').trim_space()
+		rev_file := os.join_path(path, '.git', rev_rel_path)
 		// .git/refs/heads/master
-		if !os.exists(gcbranch_file) {
-			return error('failed to find branch in: ${path}')
+		if !os.exists(rev_file) {
+			return error('failed to find revision file `${rev_file}`')
 		}
 		// get the full commit hash contained in the ref heads file
-		os.read_file(gcbranch_file) or { return error('failed to read branch in: ${path}') }
+		os.read_file(rev_file) or { return error('failed to read revision file `${rev_file}`') }
 	} else {
 		head_content
 	}
 	desired_hash_length := 7
 	return current_branch_hash[0..desired_hash_length] or {
-		error('failed to trim hash: ${current_branch_hash}')
+		error('failed to limit hash `${current_branch_hash}` to ${desired_hash_length} characters')
 	}
 }
--- a/vlib/v/util/version/version.c.v
+++ b/vlib/v/util/version/version.c.v
@@ -32,47 +32,32 @@ pub fn full_v_version(is_verbose bool) string {
 	return 'V ${version.v_version} ${@VCURRENTHASH}'
 }
 
-// githash(x) returns the current git commit hash.
-// When x is false, it is very fast - it just returns a predefined C constant.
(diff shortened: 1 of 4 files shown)


vlib/v/util/version/version.c.v
```diff
@@ -29,52 +29,35 @@ pub fn full_v_version(is_verbose bool) string {
...
-			if current_branch_hash.len > desired_hash_length {
-				return current_branch_hash[0..desired_hash_length]
-			}
+// githash tries to find the current git commit hash for the specified
+// project path by parsing the relevant files in its `.git/` folder.
+pub fn githash(path string) !string {
+	// .git/HEAD
+	git_head_file := os.join_path(path, '.git', 'HEAD')
+	if !os.exists(git_head_file) {
+		return error('failed to find HEAD in: ${path}')
```
spytheman, line 41:
```suggestion
		return error('failed to find ${path}/.git/HEAD')
```

vlib/v/util/version/version.c.v
```diff
@@ -29,52 +29,35 @@ pub fn full_v_version(is_verbose bool) string {
...
+	}
+	// 'ref: refs/heads/master' ... the current branch name
+	head_content := os.read_file(git_head_file) or {
+		return error('failed to read HEAD in: ${path}')
```
spytheman, line 45:
```suggestion
		return error('failed to read ${path}/.git/HEAD')
```

vlib/v/util/version/version.c.v
```diff
@@ -29,52 +29,35 @@ pub fn full_v_version(is_verbose bool) string {
...
+	}
+	current_branch_hash := if head_content.starts_with('ref: ') {
+		gcbranch_rel_path := head_content.replace('ref: ', '').trim_space()
+		gcbranch_file := os.join_path(path, '.git', gcbranch_rel_path)
+		// .git/refs/heads/master
+		if !os.exists(gcbranch_file) {
+			return error('failed to find branch in: ${path}')
```
spytheman, line 52:
```suggestion
			return error('failed to find branch file ${gcbranch_file}')
```

vlib/v/util/version/version.c.v
```diff
@@ -29,52 +29,35 @@ pub fn full_v_version(is_verbose bool) string {
...
 		}
-		break
+		// get the full commit hash contained in the ref heads file
+		os.read_file(gcbranch_file) or { return error('failed to read branch in: ${path}') }
+	} else {
+		head_content
+	}
+	desired_hash_length := 7
+	return current_branch_hash[0..desired_hash_length] or {
+		error('failed to trim hash: ${current_branch_hash}')
```
spytheman, line 61:
```suggestion
		error('failed to limit hash: ${current_branch_hash} to ${desired_hash_length characters}')
```

vlib/v/util/version/version_test.v
```diff
@@ -0,0 +1,23 @@
...
+
+fn test_githash() {
+	sha := version.githash(@VMODROOT)!
+	assert sha == @VCURRENTHASH
+
+	git_proj_path := os.join_path(os.vtmp_dir(), 'test_githash')
+	defer {
+		os.rmdir_all(git_proj_path) or {}
+	}
+	os.execute_opt('git init ${git_proj_path}')!
```
spytheman, line 12: imho do `os.chdir(git_proj_path)!` after the `git init`, so that the potential `git config user.email` etc later will operate in the new repo.

vlib/v/util/version/version_test.v
```diff
@@ -0,0 +1,23 @@
...
+	os.execute_opt('git config user.name') or {
+		os.execute_opt('git config user.email "ci@vlang.io"')!
+		os.execute_opt('git config user.name "V CI"')!
+	}
+	os.chdir(git_proj_path)!
+	os.write_file('v.mod', '')!
+	os.execute_opt('git add .')!
+	os.execute_opt('git commit -m "test"')!
+	test_rev := os.execute_opt('git rev-parse HEAD')!.output[..7]
+	assert version.githash(git_proj_path)! == test_rev
```
spytheman, line 22: imho add a second assert, with the same condition, but after doing `os.chdir(os.vtmp_dir())!`

vlib/v/util/version/version_test.v
```diff
@@ -0,0 +1,23 @@
+import v.util.version
+import os
+
+fn test_githash() {
+	sha := version.githash(@VMODROOT)!
+	assert sha == @VCURRENTHASH
```
spytheman, line 6: That may fail on the release zip files, since they do not pack the .git/ folder in them.

spytheman: imho check for the .git folder existence first, and assert only if it is present, otherwise return early in that test

vlib/v/util/version/version_test.v
spytheman: I think that will fix the CI error too.

vlib/v/util/version/version_test.v
ttytm: Thanks for the great review! Gonna update when I have a free moment.

spytheman:
Excellent work.

vlib/v/util/version/version_test.v
```diff
@@ -27,12 +27,12 @@ fn test_githash() {
...
 	os.execute_opt('git add .')!
 	os.execute_opt('git commit -m "test1"')!
-	test_rev := os.execute_opt('git rev-parse HEAD')!.output[..7]
+	test_rev := os.execute_opt('git rev-parse --short HEAD')!.output.trim_space()
 	assert test_rev == version.githash(git_proj_path)!
 	os.write_file('README.md', '')!
 	os.execute_opt('git add .')!
 	os.execute_opt('git commit -m "test2"')!
-	test_rev2 := os.execute_opt('git rev-parse HEAD')!.output[..7]
+	test_rev2 := os.execute_opt('git rev-parse --short HEAD')!.output.trim_space()
```
spytheman, line 35: That is a bit brittle, since if you do not specify the length explicitly with `--short=7` for example, it will depend on https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreabbrev which says:
> core.abbrev
>    Set the length object names are abbreviated to. If unspecified or set to "auto",
> an appropriate value is computed based on the approximate number of packed objects in 
> your repository, which hopefully is enough for abbreviated object names to stay unique 
> for some time. If set to "no", no abbreviation is made and the object names are shown 
> in their full length. The minimum length is 4.

i.e. in the future, when the v repo grows, it may decide to use 8 characters.

ttytm: Good point, we should add that specification to that

PR #21181, opened 2024-04-03
parser, checker: move more match validation into checker, add error for match without branches

ttytm:
Moves validation for match expression into the checker. We already do most of the related validation in the checker so the mantra "Parse, don't validate" might work especially well here.

The change includes adding an error for empty matches. 
So instead of:
```
error: invalid expression: unexpected token `}`
    1 | match true {
    2 | }
      | ^
```
We'll get:
```
error: `match` must have at least two branches
    1 | match true {
      | ~~~~~
    2 | }
```


vlib/v/checker/match.v
```diff
@@ -535,17 +535,27 @@ fn (mut c Checker) match_exprs(mut node ast.MatchExpr, cond_type_sym ast.TypeSym
 			}
 		}
 	}
+	if node.branches.len == 0 {
+		c.error('`match` must have at least two branches', node.pos)
```
JalonSolov, line 539:
```suggestion
		c.error('`match` must have at least one branch and `else`', node.pos)
```

ttytm: Ty @JalonSolov

PR #21183, opened 2024-04-04
breaking,checker: disallow initializing private struct fields outside structs module

ttytm:
Currently, it's possible to init a struct and assign values to it's private fields. Only accessing fields directly will trigger an error.

Currently following is possible
```v
import amod

b := amod.Bcg{ // public struct
	x: 3 // private field
}
```
After the changes an error will be visible on initialization (ref. to added test:`struct_field_private_err.out`). Eventually the change should allow for better accessibility control and improved security.

### DIFF
--- a/.github/workflows/v_apps_and_modules_compile_ci.yml
+++ b/.github/workflows/v_apps_and_modules_compile_ci.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Test vsql compilation and examples
         run: |
           echo "Install vsql"
-          .github/workflows/retry.sh v install elliotchance.vsql ; cd ~/.vmodules/elliotchance/vsql
+          .github/workflows/retry.sh v install https://github.com/ttytm/vsql@temp/vi-ci ; cd ~/.vmodules/vsql
           echo "Generate vsql/grammar.v"
           make vsql/grammar.v
           echo "Compile vsql"
@@ -53,10 +53,10 @@ jobs:
 
       - name: Test discord.v
         run: |
-          echo "Clone https://github.com/DarpHome/discord.v/"
-          .github/workflows/retry.sh git clone https://github.com/DarpHome/discord.v/ discord && cd discord
-          echo "Checkout last known good commit"
-          git checkout 533485c08f21df91ff62fea9477e7017d21f91c4
+          echo "Clone https://github.com/DarpHome/discord.v"
+          .github/workflows/retry.sh v install https://github.com/ttytm/discord.v@temp/v-ci && cd ~/.vmodules/discord
+          # echo "Checkout last known good commit"
+          # git checkout 789de8ad35ee2683fbcd950fc07936f052088d0c
           echo "Execute Tests"
           v test .
 
--- a/vlib/cli/flag_test.v
+++ b/vlib/cli/flag_test.v
@@ -25,21 +25,6 @@ fn test_if_string_flag_parses() {
 	flag.parse(['-flag=value2'], false) or { panic(err) }
 	mut values := flag.get_strings() or { panic(err) }
 	assert values == ['value1', 'value2']
-
-	flags := [
-		cli.Flag{
-			flag: .string_array
-			name: 'flag'
-			value: ['a', 'b', 'c']
-		},
-		cli.Flag{
-			flag: .string
-			name: 'flag2'
-		},
-	]
-
-	values = flags.get_strings('flag') or { panic(err) }
-	assert values == ['a', 'b', 'c']
 }
 
 fn test_if_bool_flag_parses() {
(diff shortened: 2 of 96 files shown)


ttytm:
Some of the remaining errors might require discussion as not all fields of them might be meant for a `pub mut`.

vlib/cli/flag_test.v
```diff
@@ -25,21 +25,6 @@ fn test_if_string_flag_parses() {
 	flag.parse(['-flag=value2'], false) or { panic(err) }
 	mut values := flag.get_strings() or { panic(err) }
 	assert values == ['value1', 'value2']
-
-	flags := [
-		cli.Flag{
-			flag: .string_array
-			name: 'flag'
-			value: ['a', 'b', 'c']
```
ttytm, line 33: Referring to the test above, the removed parts didn't extend coverage or test any new/other api functionality.

vlib/net/tcp.c.v
```diff
@@ -9,8 +9,7 @@ pub const tcp_default_write_timeout = 30 * time.second
 @[heap]
 pub struct TcpConn {
 pub mut:
-	sock TcpSocket
-mut:
+	sock           TcpSocket
 	handle         int
```
ttytm, line 13: imho the `handle` should eventually not be publicly mutable. But to not introduce a big breaking change for how TcpConn structs are currently created, it's is likely the only way have it pub mut.

ttytm:
@spytheman you have push access to https://github.com/elliotchance/vsql?

vlib/v/checker/struct.v
```diff
@@ -740,11 +740,28 @@ or use an explicit `unsafe{ a[..] }`, if you do not want a copy of the slice.',
...
-								c.deprecate('field', field.name, field.attrs, init_field.pos)
-								break
+					if c.mod != type_sym.mod {
+						if !field.is_pub {
+							parts := type_sym.name.split('.')
+							mod_type := if parts.len > 1 {
+								parts#[-2..].join('.')
+							} else {
+								parts.last()
+							}
```
spytheman, line 750: imho move this right next to the error in the loop below, not outside it. The common case is that there will be no errors, and all those splits, slices and joins can be avoided.

spytheman (approved):
Excellent work.

spytheman:
> @spytheman you have push access to https://github.com/elliotchance/vsql? If not maybe it's okay to comment it's CI step out for now as the developer is not active for the last months.

I do not, but @elliotchance is on Discord, so I think we can coordinate a change with a PR.

JalonSolov:
I have a question about this PR... how am I supposed to initialize a private field in a struct from another module, if I can't do it at instantiation time?

With this change in place, every module that has a struct with a private field will now have to supply a `new` or `create` or whatever method for that struct, in order to initialize those fields.

Either that or some sort of setter method just for those fields.

ttytm:
Private struct fields are usually not not meant to be initialised from other modules.
For the cases where a private struct field should be initialised a factory function like it is with `@[noinit]` will be required.

vlib/v/checker/struct.v
ttytm: :100: That's where it should be! Failed on moving it when did the update to the error position!

ttytm:
I temporary added a fork of external V apps to the CI. 

The only one remaining (with more restrictive merge access) is the PR at gitly.

.github/workflows/v_apps_and_modules_compile_ci.yml
```diff
@@ -53,10 +53,10 @@ jobs:
 
       - name: Test discord.v
         run: |
-          echo "Clone https://github.com/DarpHome/discord.v/"
-          .github/workflows/retry.sh git clone https://github.com/DarpHome/discord.v/ discord && cd discord
-          echo "Checkout last known good commit"
-          git checkout 533485c08f21df91ff62fea9477e7017d21f91c4
+          echo "Clone https://github.com/DarpHome/discord.v"
+          .github/workflows/retry.sh v install https://github.com/ttytm/discord.v@temp/v-ci && cd ~/.vmodules/discord
```
spytheman, line 57: https://github.com/DarpHome/discord.v/ was updated, and now has 789de8ad35ee2683fbcd950fc07936f052088d0c as well.

spytheman:
The `gitly` and the `ui` failures are now fixed too.

I think it is good to go.

.github/workflows/v_apps_and_modules_compile_ci.yml
ttytm: Yes, unfortunately I missed one thing in the PR. The commit would not work yet. It would require to change one more field to pass. I'll submit another PR soon.

vlib/v/checker/struct.v
```diff
@@ -740,11 +740,28 @@ or use an explicit `unsafe{ a[..] }`, if you do not want a copy of the slice.',
...
+						if !field.is_pub {
+							parts := type_sym.name.split('.')
+							for init_field in node.init_fields {
+								if field.name == init_field.name {
+									mod_type := if parts.len > 1 {
+										parts#[-2..].join('.')
+									} else {
+										parts.last()
+									}
+									c.error('cannot access private field `${field.name}` on `${mod_type}`',
```
spytheman, line 753: I think, that it will be better, for this to be a warning for now, not an error (for a month or two).
It is too breaking otherwise.

ttytm: Good idea :+1:

ttytm: Missed to update the error message itself to be more appropriate for a warning. It'll take a few hours until I'm back at my desk.

PR #21185, opened 2024-04-04
breaking,net.ftp: allow to choose a different port than port 21 (change FTP.connect to accept `host:port`, not just a `host` address)

4zv4l:
ftp: allow to choose a different port than port 21 by simply removing the hardcoded port and using the same syntax for the address as for dial_tcp.

so instead of doing this
```v
ftp.connect("127.0.0.1")
```
would do this
```v
ftp.connect("127.0.0.1:2021")
```
this would allow to connect to ftp server using another port (for example due to security reason).


spytheman (approved):
Good work.

PR #21186, opened 2024-04-04
x.json2: decode array

### LINKED ISSUE #21168: [x.json2] Decoding array type panics
### Describe the bug

When json decoding an array type, V panics

### Reproduction Steps

```
import x.json2

pub struct Value {
	value string
}

fn main() {
	println(json2.decode[[]Value]('[{"value": "abc"}, {"value": "def"}]')!)
}
```

### Expected Behavior

This is printed
```
[
  Value {
    value: "abc"
  },
  Value {
    value: "def"
  },
]
```

### Current Behavior

```
   module: main
 function: main()
  message: The type `[]Value` can't be decoded.
     file: code.v:8
   v hash: a1b6360
/tmp/v_60000/../../../../../../home/admin/v/vlib/builtin/builtin.c.v:66: at panic_debug: Backtrace
/tmp/v_60000/../../../../../../box/code.v:10: by main__main
/tmp/v_60000/../../../../../../tmp/v_60000/code.01HTEKB9M24DZTMKFASK5NH047.tmp.c:17052: by main
Exited with error status 1
```

### Possible Solution

Handle array type as type argument

_No response_

### V version

V 0.4.5 a1b6360

V full version: V 0.4.5 4687f8c.a1b6360
Processor: 12 cpus, 64bit, little endian, 

getwd: C:\dev\v\winmd
vexe: C:\dev\_tools\v\v.exe
vexe mtime: 2024-04-02 04:56:12

vroot: OK, value: C:\dev\_tools\v
VMODULES: OK, value: C:\Users\simen\.vmodules
VTMP: OK, value: C:\Users\simen\AppData\Local\Temp\v_0

Git vroot status: weekly.2024.13-70-ga1b6360d
.git/config present: true

> [!NOTE]
>

felipensp:
Fix #21168

Added generic way to decode a JSON array string into a array of generic type.

```v
import x.json2


pub struct Value {
	value string
}

pub struct Value2 {
	value string
}

fn main() {
	println(json2.decode_array[Value]('[{"value": "abc"}, {"value": "def"}]')!)
	println(json2.decode_array[Value2]('[{"value": "abc"}, {"value": "def"}]')!)
}
```

### DIFF
--- a/vlib/x/json2/decoder.v
+++ b/vlib/x/json2/decoder.v
@@ -137,6 +137,25 @@ pub fn decode[T](src string) !T {
 	return decode_struct[T](T{}, res)
 }
 
+// decode_array is a generic function that decodes a JSON string into the array target type.
+pub fn decode_array[T](src string) ![]T {
+	res := raw_decode(src)!.as_map()
+	return decode_struct_array(T{}, res)
+}
+
+// decode_struct_array is a generic function that decodes a JSON map into array struct T.
+fn decode_struct_array[T](_ T, res map[string]Any) ![]T {
+	$if T is $struct {
+		mut arr := []T{}
+		for v in res.values() {
+			arr << decode_struct[T](T{}, v.as_map())!
+		}
+		return arr
+	} $else {
+		return error("The type `${T.name}` can't be decoded.")
+	}
+}
+
 // decode_struct is a generic function that decodes a JSON map into the struct T.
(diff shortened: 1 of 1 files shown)


ttytm:
Why not making it part of `decode`?

felipensp:
> Why not making it part of `decode`?

Must investigate a way to decode []T to T, it seems we can't do it at the moment. This was the point.

spytheman (approved):

PR #21190, opened 2024-04-05
xml: update entity parse test

ttytm:
Upgrades an xml test that assigns values to the private field `parsed_reverse_entities` which in an external test. Otherwise, the fields data is only used internally.

https://github.com/vlang/v/blob/254250d375acff5a92c2a0b1a3079c98e7e16244/vlib/encoding/xml/types.v#L61-L63

The new test might be more robust and easier to follow as well, as it compares the parses files with entities with an an expected file that have the entities already resolved. 

The upgrade is also required for #21183


spytheman (approved):
Good work.

PR #21193, opened 2024-04-05
vdoc: select first result when entering search term, use more explict query for `header.doc-nav`

ttytm:
Currently, there are some situations where search results are sub-optimally presented. For example, when opening a module site like https://modules.vlang.io/sync.html where the sidebar navigation on the left is already scrolled down. Then, upon entering a search term, such as `str`, the navigation remains scrolled down, the most relevant search results are not visible without manually scrolling up in the side bar. With the changes, upon entering a search term, the search navigation scrolls to the top.


spytheman (approved):
Good work.

PR #21197, opened 2024-04-06
cgen: add `autofree` comptime check

### LINKED ISSUE #21188: Double free with `-autofree` and `os.input`
### Describe the bug

when using `os.input` I get a double free on the returned string

### Reproduction Steps

With this code
```v
import os

#flag -nostartfiles -Wl,-z,now

@[export: '_start']
fn main() {
    name := os.input("Your name: ")
    print("Hi ")
    print(name)
    println(" !")

    exit(0)
}

// garbage because with GCC I get undefined symboles so I need to define them somehow
@[export: '_init']
pub fn a() {}
@[export: '_fini']
pub fn b() {}
```

compiling with:
`v -cc gcc -gc none hello.v`
it works

but with:
`v -cc gcc -gc none -autofree hello.v`
I get a double free

image

### Expected Behavior

No double free, to work like without the `-autofree`

### Current Behavior

image

it double free

### Possible Solution

_No response_

_No response_

### V version

V 0.4.5 2c646bf

Kali 2024.1
> [!NOTE]
>

felipensp:
Add possibility to check if autofree is enabled.

```v
$if autofree ? {

}
```

### DIFF
--- a/vlib/os/os.c.v
+++ b/vlib/os/os.c.v
@@ -539,9 +539,11 @@ pub fn get_raw_line() string {
 		nr_chars := unsafe { C.getline(&buf, &max, C.stdin) }
 		str := unsafe { tos(&u8(buf), if nr_chars < 0 { 0 } else { nr_chars }) }
 		ret := str.clone()
-		unsafe {
-			if nr_chars > 0 && buf != 0 {
-				C.free(buf)
+		$if !autofree ? {
+			unsafe {
+				if nr_chars > 0 && buf != 0 {
+					C.free(buf)
+				}
 			}
 		}
 		return ret
--- a/vlib/os/os.c.v
+++ b/vlib/os/os.c.v
@@ -539,11 +539,9 @@ pub fn get_raw_line() string {
 		nr_chars := unsafe { C.getline(&buf, &max, C.stdin) }
 		str := unsafe { tos(&u8(buf), if nr_chars < 0 { 0 } else { nr_chars }) }
 		ret := str.clone()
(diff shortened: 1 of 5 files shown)


vlib/os/os.c.v
```diff
@@ -539,9 +539,11 @@ pub fn get_raw_line() string {
 		nr_chars := unsafe { C.getline(&buf, &max, C.stdin) }
 		str := unsafe { tos(&u8(buf), if nr_chars < 0 { 0 } else { nr_chars }) }
 		ret := str.clone()
-		unsafe {
-			if nr_chars > 0 && buf != 0 {
-				C.free(buf)
+		$if !autofree ? {
```
spytheman, line 542: this should not need the `?`, since the compiler knows about it. The ? is for user defines.

spytheman: Imho implement the feature in this PR, and add its usage in a test or in a later PR. 

Modifying os.c.v in the same PR will make bootstrapping much harder.

felipensp: OK.

spytheman (approved):
Excellent work.

spytheman:
There seems to be one problem:
```v
println('start')

$if autofree {
    println('autofree was passed')
}
$if !autofree {
    println('no autofree')
}

$if abc ? {
    println('passed `-d abc`')
}

$if !abc ? {
    println('no `-d abc` passed')
}

println('done')
```

The `no autofree` case is never triggered for some reason.

PR #21200, opened 2024-04-06
fmt: extend import import alias reference map for submodules

ttytm:
Should fix and cover the current fmt fails on vls.

### DIFF
--- a/vlib/v/fmt/fmt_test.v
+++ b/vlib/v/fmt/fmt_test.v
@@ -72,6 +72,9 @@ fn run_fmt(mut input_files []string) {
 fn test_fmt() {
 	mut input_files := os.walk_ext(os.join_path(tdir, 'tests'), '_input.vv')
 	run_fmt(mut input_files)
+
+	input_files = os.walk_ext(os.join_path(tdir, 'testdata', 'vmodules'), '_input.vv')
+	run_fmt(mut input_files)
 }
 
 fn test_fmt_vmodules() {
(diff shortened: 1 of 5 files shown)


vlib/v/fmt/fmt_test.v
```diff
@@ -72,9 +72,6 @@ fn run_fmt(mut input_files []string) {
 fn test_fmt() {
 	mut input_files := os.walk_ext(os.join_path(tdir, 'tests'), '_input.vv')
 	run_fmt(mut input_files)
-
-	input_files = os.walk_ext(os.join_path(tdir, 'testdata', 'vmodules'), '_input.vv')
-	run_fmt(mut input_files)
```
spytheman, line 77: why is not that needed anymore?

spytheman: Imho it should stay, since it is a common situation, when you make your own modules, to have a module, that you develop locally, outside of `~/vmodules`, and then other people install it through vpm in their ~/.vmodules.

In both cases, formatting the files should behave the same (mostly the imports should not change).

ttytm: Agreed. I thought this needs more of an update to fmt. But for some reason I cannot reproduce the local failure of the tests anymore. I would have left it if it would not have failed right away.

ttytm: Tests fail on master so all should be good.

spytheman (approved):
Excellent work.

PR #21201, opened 2024-04-06
x.vweb: accept query params as method arguments

Casper64:
Add ability to accept method arguments for query parameters in a `get` request the same way form values are passed as arguments in a `post` request.

### DIFF
--- a/vlib/x/vweb/tests/vweb_test.v
+++ b/vlib/x/vweb/tests/vweb_test.v
@@ -249,6 +249,12 @@ fn test_login_with_multipart_form_data_send_by_fetch() {
 	assert x.body == 'username: xmyusernamex | password: xmypassword123x'
 }
 
+fn test_query_params_are_passed_as_arguments() {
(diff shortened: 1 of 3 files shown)


vlib/x/vweb/tests/vweb_test.v
```diff
@@ -249,6 +249,12 @@ fn test_login_with_multipart_form_data_send_by_fetch() {
 	assert x.body == 'username: xmyusernamex | password: xmypassword123x'
 }
 
+fn test_query_params_are_passed_as_arguments() {
+	x := http.get('http://${localserver}/query_echo?c=3&a="test"&b=20')!
```
spytheman, line 253: What is the `c` used for?

spytheman: To test that they are passed by name, and not by position?

spytheman (approved):

vlib/x/vweb/tests/vweb_test.v
Casper64: yes

PR #21202, opened 2024-04-06
parser: update file path fields; use more expressive name for file path, remove obsolete `file_name_dir` field

ttytm:
Updates file field names on the parser. Main concern was `file_name` holding the path and same data field is e.g. used on the scanner named `file_path`. Imho simpler to understand during development when reading file_path and understanding assignments to path related fields then doing so with `file_name`.


spytheman (approved):
Excellent work.

PR #21206, opened 2024-04-06
checker: disallow structs with `@[params]` attribute as mutable function parameters

### LINKED ISSUE #21205: mutable params struct not optional as function argument
### Describe the bug

```v
module main

@[params]
struct Params {
mut:
	a bool
}

fn foo(mut opts Params) {
	// ... processing, goal is to eventually have it accessible inside the function, recursively using it.
	// opts.a = true
	// foo(opts)
	println('')
}

fn main() {
	foo()
	// Below works.
	// mut p := Params{a: 1}
	// foo(mut p)

}
```

### Reproduction Steps

above

### Expected Behavior

params optionally passing params struct

### Current Behavior

```
error: function `foo` parameter `opts` is `mut`, so use `mut Params{....}` instead
    1 | module main
      | ^
    2 | 
    3 | @[params]
```

### Possible Solution

_No response_

_No response_

### V version

v0.4.5

linux, amd64
> [!NOTE]
>

ttytm:
Fixes #21205


spytheman:
I am not sure if that should work.

ttytm:
I feel like I should have left out the recursive example and that it might be source of confusion 😅 
The change should fix a params struct that is marked as mutable being not optional and the confusing error that currently comes with it.

ttytm:
Looking into it again. It should not work at that point. `has_arg` is always empty at the point of the condition and it would not guard when args are passed.

So, if a mutable params struct should be allowed after all, below should trigger an error that the arg would have to be passed as part of a mutable struct.
```v
foo(a: true)
```

medvednikov:
@spytheman I agree, I think mutable params simply shouldn't be allowed.

ttytm:
Okay, I updated the PR to disallow it when declaring a fn with it.

But I don't fully see through where it would cause problems. Why not allowing it to be handled like any other mutable struct and to be optional?

spytheman:
> Okay, I updated the PR to disallow it when declaring a fn with it.
> 
> But I don't fully see through where it would cause problems. Why not allowing it to be handled like any other mutable struct and to be optional?

For normal non `@[params]` mut arguments, you have to use `mut` in all call sites, i.e. do `f(mut s)`. You can not just `f(Struct{})` or `f(s)`.

If you allow `f()` or `f(some_field: some_value)` for `fn f(mut opts Params) {`, then you would not have to use `mut` inside the calls - they will look just like normal calls, that do not pass any mutable state down to `f()`.

That can be valuable, but also inconsistent.

-> I can see it both ways, thus I could not decide how it should behave.

vlib/v/tests/struct_test.v
```diff
@@ -277,17 +276,14 @@ fn test_struct_literal_args() {
 	bar_config(Config{}, 10)
 	bar_config(Config{ def: 4 }, 4)
 
-	mut c_ := Config{
-		def: 10
-	}
-	c := mut_bar_config(mut c_, 10)
```
spytheman, line 283: This should remain to be allowed, the call has `mut c_`.

spytheman: The inconsistency, is only when you use the default `f()`, and f gets a mutable parameter, that was created implicitly. If you pass it explicitly, you and the reader will know where it originated.

ttytm: > The inconsistency, is only when you use the default f(), and f gets a mutable parameter, that was created implicitly. If you pass it explicitly, you and the reader will know where it originated.

The test used `Config` (a struct with a `@[params]` attribute) to test passing mutable structs. That's why I added another test passing a mut struct that isn't a params struct.

Else wouldn't this disagree with alex, disallowing mut params, and lead us where I wanted to go in the second comment? Allowing a params struct like this:
> ```v
> mut p := Params{a: true}
> foo(mut p)
> ```

Or is what is meant to be done here to only disallow trailing mut params structs?

spytheman: I do not know. @medvednikov what do you think?

PR #21210, opened 2024-04-07
faster clone tcc using "git fetch"

cakyus:
faster clone using "git fetch --depth=1" since we do not need entire history of tcc project.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -15,6 +15,7 @@ LEGACYLIBS := $(VROOT)/thirdparty/legacy
 TMPLEGACY := $(LEGACYLIBS)/source
 TCCOS := unknown
 TCCARCH := unknown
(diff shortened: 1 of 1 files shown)


spytheman (commented):
No. It is much more complex than the current solution, and the savings are minimal.
Note that currently it does:
`git clone --filter=blob:none --quiet --branch thirdparty-linux-amd64 https://github.com/vlang/tccbin ./thirdparty/tcc`

The `--filter=blob:none` makes sure that older commits are available on demand, and are not fetched if not needed.

`git clone --depth=1 --branch thirdparty-linux-amd64 https://github.com/vlang/tccbin ./thirdparty/tcc` could be also used instead of 4 separate commands, like here, but has some other disadvantages.

PR #21216, opened 2024-04-07
vpm: optimize performance by adding filter when cloning

ttytm:

### DIFF
--- a/cmd/tools/vpm/vcs.v
+++ b/cmd/tools/vpm/vcs.v
@@ -23,7 +23,7 @@ const vcs_info = {
 	VCS.git: VCSInfo{
 		dir: '.git'
 		args: struct {
-			install: 'clone --depth=1 --recursive --shallow-submodules'
+			install: 'clone --depth=1 --recursive --shallow-submodules --filter=blob:none --also-filter-submodules'
 			version: '--single-branch -b'
 			update: 'pull --recurse-submodules' // pulling with `--depth=1` leads to conflicts when the upstream has more than 1 new commits.
 			path: '-C'


spytheman (approved):
Good work.

Wertzui123:
This breaks VPM on systems with an older Git version as `--also-filter-submodules` is fairly new.
Maybe we should revert this change or at least check for the Git version before adding the flag to the clone command?

ttytm:
Good catch. The version check should be good or removing just this flag. We can keep the filtering benefit then.

Wertzui123:
> Version check would be best. Nobody should be using old versions of Git - if their package manager doesn't have the newer version, they can always clone it and build it themselves.
> 
> Too many things can go wrong with older versions of Git. :-\

  While I agree with you in principle, the "old version" you're referring to is from just 2022. And yes, you could theoretically build Git manually (which I have tried earlier today), but especially if you're not very familiar with C build systems it might be a little (and mostly unnecessary) challenge. Furthermore, I need to use VPM in a Debian 10 CI/CD script (for BC reasons), which is why building Git manually is even more unsuited.

JalonSolov:
The version check I'm referring to would check the version of Git that's available, and add the extra option if it was "new enough".

As for the age of Git... it doesn't matter.  They don't do new versions on a whim... the new versions are to fix problems found in the previous version.  Rarely, the new version will have a new feature (such as this new option).

Either way, you're always better off with the latest version of Git, since your entire codebase depends on it working properly.

If Debian made this hard to keep up with, the fault is with Debian.

PR #21217, opened 2024-04-07
time: cleanup module

ttytm:
Splits up a cleanup from some fixes to the time module. 
Updates the `unix` param name from `abs` to `epoch` for a cleaner signature (e.g. name used in cpp docs)


vlib/time/time.c.v
```diff
@@ -28,14 +24,13 @@ fn C.strftime(buf &char, maxsize usize, const_format &char, const_tm &C.tm) usiz
...
-	$if windows {
+	} $else $if windows {
 		return win_now()
-	}
-	$if solaris {
+	} $else $if solaris {
 		return solaris_now()
+	} $else {
+		return linux_now()
 	}
```
spytheman, line 33: I am a bit afraid, to merge this, because `time` is used in bootstrapping, and the `$if x {}` statements are simpler and have less edge cases than a comptime if-else ladder.

Can you please restore just those and the ones $if statements in utc, as they were, leaving only the extraction of the code to duration.v in this PR?

ttytm: It came easier to me splitting things up by reverting the duration changes here and pulling them onto their own branch.

PR #21220, opened 2024-04-08
vdoc: fix indentation in module readmes

ttytm:
<table>
<tr>
	<th>
		Current
	</th>
	<th>
		Updated
	</th>
</tr>
<tr>
	<td>

Screenshot_20240408_022330
	</td>
	<td>

Screenshot_20240408_022332
	</td>
</tr>
	<tr>
 <td>

Screenshot_20240408_022325
 </td>
 <td>

Screenshot_20240408_022334
	</td>
</tr>
</table>

ref: https://modules.vlang.io/arrays.html, https://modules.vlang.io/os.html

Regarding commit No. 2: We can let the markdown lib do the escaping.
Without the second commit a `<script>` tag in html will be wrongly displayed as `` ` `` . E.g., generating the docs of https://github.com/ttytm/webview and viewing the `Usage Example` in the readme.

I'm thankful to have the tests to not implement regressions. Though they also included invalid markdown that didn't made it through the changes.


cmd/tools/vdoc/tests/testdata/output_formats/main.html
```diff
@@ -1,54 +1,54 @@
 <section id="readme_main" class="doc-node">
 			<div class="title"><h1> main <a href="#readme_main">#</a></h1></div>
-<h2>Description</h2><p>This is an example of a an .md file, used for adding more rich text documentation in a project or module.</p><p>This is a <a href="https://vlang.io/">link</a> to the main V site.</p><p>This is a <b>bold text</b>.</p><p>This is a script <code>console.log('hi from README.md');</code> .</p><h2>Examples</h2><h3>Processing command line args</h3><pre><code class="language-v"><span class="token keyword">import</span> os
+<h2>Description</h2><p>This is an example of a an .md file, used for adding more rich textdocumentation in a project or module.</p><p>This is a <a href="https://vlang.io/">link</a> to the main V site.</p><p>This is a <b>bold text</b>.</p><p>This is a script <code>&lt;script&gt;console.log('hi from README.md');&lt;/script&gt;</code> .</p><h2>Examples</h2><h3>Processing command line args</h3><pre><code class="language-v"><span class="token keyword">import</span> os
```
spytheman, line 3: why is it `textdocumentation` here, but `text documentation` in the md file?

spytheman: is it because of merge_comments?

ttytm: Thanks for looking properly during the review @spytheman 

Likely not because of `merge_comments` since it is not used for the readme text. I think newlines need to be escaped, adding a space character.

Also, I should have kept `merge_comments_without_examples` here. This changed during testing yesterday and I forgot to revert it.

Gonna update and extend tests for both.

PR #21223, opened 2024-04-08
docs: Add Exherbo Linux install instructions to README.md

davlgd:


README.md
```diff
@@ -160,6 +160,15 @@ shell/editor after that, so that it can pick up the new PATH variable.
...
 
+### Exherbo Linux
+
+<details><summary>Expand Exherbo Linux instructions</summary>
+
+```bash
+# cave resolve -x repository/CleverCloud
+# cave resolve -x v
+```
+
```
spytheman, line 171: These instruction seem to be platform specific, not just for the V dependencies, like `git`, `make`, `libatomic`, but for V itself.

In my opinion, it is preferable, that `git clone https://github.com/vlang/v && cd v && make` (i.e. installing from source), remain as the main installation method, until V reaches version 1.0 .

PR #21224, opened 2024-04-08
vdoc: improve line merging for readme contents

ttytm:
Fixes an issue where readme markdown lines would be displayed without spacing.

```md
This is an example of a an .md file, used for adding more rich text
documentation in a project or module.
```

```html
<p>This is an example of a an .md file, used for adding more rich textdocumentation [...]
```

Discovered in https://github.com/vlang/v/pull/21220#discussion_r1555354018


spytheman (approved):
That processing seems a bit complex, but the output is definitely better.
Thank you.

PR #21226, opened 2024-04-08
vdoc: fix regression on example comments

ttytm:
Fix for a regression that happend due to accidentally leaving merge methods changed when working #21220. The symptom is that examples are part of the comment text
Screenshot_20240408_160452

Adds tests for doc comment examples. The main coverage here is that there are no duplicates in of the example comments.


ttytm:
@spytheman Would the CI failure here resolve itself after a rebase/merge?

spytheman:
> @spytheman Would the CI failure here resolve itself after a rebase/merge?

Not sure, I do not understand why it failed in the first place, since I do not see anything platform specific in the change here.

Lets rebase it, and see if it will repeat.

ttytm:
If thats okay not adding the test here due to the conflicts with Ubuntu Docker MUSL CI and just restoring the correct behaviour then lets do it. A simple change that can get a little of coverage due to it's blame history.

PR #21231, opened 2024-04-08
cgen, checker: allow `none` as arg to `?&type`

### LINKED ISSUE #21213: cannot use `none` as a `?&Type`
### Describe the bug

When a function takes an optional pointer, `?&T`, passing `none` to the function gives a type error.

### Reproduction Steps

```
fn takes_optional_pointer(maybe_ptr ?&int) int {
    if ptr := maybe_ptr {
        return *ptr
    } else {
        return 0
    }
}

fn main() {
    val := takes_optional_pointer(none)
    println("${val}")
}
```

### Expected Behavior

The program should compile and run fine.

### Current Behavior

It gives a type error.
```
optional_ptr_bug.v:10:35: error: cannot use `none` as `?&int` in argument 1 to `takes_optional_pointer`
    8 |
    9 | fn main() {
   10 |     val := takes_optional_pointer(none)
      |                                   ~~~~
   11 |     println("${val}")
   12 | }
```

### Possible Solution

The way that type checking only requires `Type` rather than `&Type` for function arguments might be tripping up the type inference, making it think that only `?Type` is needed, and thus treating `none` as a `?Type`.

In fact I think before #21081 was fixed, it used to actually accept `none` and complete type checking and cgen, but then the c compiler would complain about the resulting C code, its complaint was something along the lines of `__t0` was an `__option_MyTypeHere` when it should be an `__option_MyTypeHere_ptr`.

Given that `nil` is considered unsafe, one would expect `?&Type` to be a natural and valid way of passing optional pointers around. In Rust `?&Type` takes up just as much memory as `&Type` of course, but in V it should at least behave as expected, even if it takes more memory and requires manipulation of unaligned memory to work.

Optional pointers are especially useful when making C bindings for functions that take multiple optional pointers.
```
fn C.my_fun(arg1 &Type1, arg2 &Type2)

fn my_fun_wrapper(arg1 ?&Type1, arg2 ?&Type2) {
    arg1_ptr := arg1 or { unsafe { nil } }
    arg2_ptr := arg2 or { unsafe { nil } }
    C.my_fun(arg1_ptr, arg2_ptr)
}

fn main() {
    my_fun_wrapper(none, none) // This gives an error! How annoying!
    my_fun_wrapper(?&Type1(none), ?&Type2(none)) // This works, but is so much less convenient than option types should be for a language like V
}
```
Changing the V wrapper to take `?Type1` and `?Type2` instead of `?&Type1` and `?&Type2` would add unnecessary stack allocation and copying, and potentially complicated default-initialization and cleanup, in order to simply call a C function that takes nullable pointers - nullable pointers are very common in C APIs!

### V version

V 0.4.5 ad17be5

Git vroot status: weekly.2024.14-26-gad17be5d (5 commit(s) behind V master)
.git/config present: true

thirdparty/tcc: N/A

felipensp:
Fix #21213

### DIFF
--- a/vlib/v/tests/option_ptr_arg_none_test.v
+++ b/vlib/v/tests/option_ptr_arg_none_test.v
@@ -1,0 +1,12 @@
+fn takes_optional_pointer(maybe_ptr ?&int) int {
+	if ptr := maybe_ptr {
+		return *ptr
+	} else {
+		return 0
+	}
+}
+
+fn test_main() {
+	val := takes_optional_pointer(none)
+	assert val == 0
+}
--- a/vlib/v/checker/check_types.v
+++ b/vlib/v/checker/check_types.v
@@ -228,6 +228,7 @@ fn (mut c Checker) check_expected_call_arg(got ast.Type, expected_ ast.Type, lan
 		if expected.has_flag(.option) {
 			got_is_ptr := got.is_ptr()
 				|| (arg.expr is ast.Ident && (arg.expr as ast.Ident).is_mut())
+				|| arg.expr is ast.None
 			if (expected.is_ptr() && !got_is_ptr) || (!expected.is_ptr() && got.is_ptr()) {
 				got_typ_str, expected_typ_str := c.get_string_names_of(got, expected)
 				return error('cannot use `${got_typ_str}` as `${expected_typ_str}`')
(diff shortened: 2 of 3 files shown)


vlib/v/tests/option_ptr_arg_none_test.v
```diff
@@ -0,0 +1,12 @@
+fn takes_optional_pointer(maybe_ptr ?&int) int {
+	if ptr := maybe_ptr {
+		return *ptr
+	} else {
+		return 0
```
spytheman, line 5: Please return (and assert) a different value. `0` is a default value for many things, and thus the test with it, is a bit less sensitive than what it could be.

spytheman (approved):
Excellent work.

PR #21233, opened 2024-04-08
vdoc: don't add additional _docs directory when an out path is specified

ttytm:
Running a command that specifies an output path for the documentation like `v doc -m -f html -o docs .` currently creates a `docs/_docs` directory. The change removes adding the additional directory and uses the output path as final directory for the document files. This is likely the more expected result and the behavior that is easier to work with.


spytheman (approved):
Excellent work.

Issue #21234, opened 2024-04-09
unable to concat strings in for loop "operator `+=` not defined on left operand type `string`"


ttytm:
### Describe the bug

```v
mut arr := ['a', 'b', 'c']
for mut c in arr {
	c += 'd'
}
```

### Reproduction Steps

above

### Expected Behavior

Working code. Same as arr[i] += 'str'

### Current Behavior

```
error: operator `+=` not defined on left operand type `string`
    1 | mut arr := ['a', 'b', 'c']
    2 | for mut c in arr {
    3 |     c += 'd'
      |     ^
    4 | }
```

### V version

v0.4.5

### Environment details (OS name and version, etc.)

Linux x64

PR #21237, opened 2024-04-09
 checker, cgen: allow `for mut v in [12, 13, 14] { v+= 2 }` 

Delta456:
Fixes https://github.com/vlang/v/issues/21234


spytheman (approved):
Excellent work.

PR #21239, opened 2024-04-09
examples: examples: add pidigits.v; optimise math.big's Integer divide a bit

spytheman:
- examples: add pidigits.v
- optimise math.big's Integer divide a bit


spytheman:
The example works like this:
image

PR #21240, opened 2024-04-09
v.util: improve code related to diff tool specified via environment, add check if the diff tool exists

ttytm:
The PR works on the `util.diff` module's code related to the `VDIFF_TOOL` environment variable. 
It cleans up to help with maintainability and slightly extends by adding an error that should improve usability. 

Regarding the changes: 
At the beginning of the function we already checked if there is an `env_difftool` and return early. How  `env_difftool` was used after this in the function, was quasi dead code. 
https://github.com/vlang/v/blob/4dcc9265e16e74dc3691e6da427121e87bd05341/vlib/v/util/diff/diff.v#L11-L13

Then, instead of just returning the strings specified in the environment string it is also ensured here that specified command is executable. If not an error is returned to the user. E.g. when it isn't installed or if a typo or something.


spytheman (approved):
Excellent work.

PR #21242, opened 2024-04-09
v.util: improve find_diff_cmd: don't add spaces to result without env opts

ttytm:
- Adds test for the diff util module.

- Fixes an issue where `find_working_diff_command()!` adds spaces to the result, potentially making it unavailable to other commands. E.g. `os.find_abs_path_of_executable` would not find the result of the function.

  ```v
  cmd := util.find_working_diff_command()!
  dump(os.find_abs_path_of_executable(cmd)!)
  ```
  
Might even be required to make the changes in #21243 robust.

### DIFF
--- a/vlib/v/util/diff/diff.v
+++ b/vlib/v/util/diff/diff.v
@@ -40,8 +40,8 @@ pub fn find_working_diff_command() !string {
 				}
 			}
 			if diffcmd in ['code', 'code.cmd'] {
-				cmd := '${diffcmd} ${if env_diffopts != '' { env_diffopts } else { '-d' }}'
-				return cmd
+				// Make sure the diff flag `-d` is included in any case.
+				return '${diffcmd} ${env_diffopts} -d'
 			}
 			// Don't add spaces to the cmd if there are no `env_diffopts`.
 			return if env_diffopts != '' { '${diffcmd} ${env_diffopts}' } else { diffcmd }
--- a/vlib/v/util/diff/diff.v
+++ b/vlib/v/util/diff/diff.v
@@ -40,11 +40,11 @@ pub fn find_working_diff_command() !string {
 				}
 			}
 			if diffcmd in ['code', 'code.cmd'] {
-				// there is no guarantee that the env opts exist
-				// or include `-d`, so (harmlessly) add it
-				return '${diffcmd} ${env_diffopts} -d'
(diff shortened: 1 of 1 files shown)


vlib/v/util/diff/diff.v
```diff
@@ -41,11 +41,11 @@ pub fn find_working_diff_command() !string {
 		}
 		if p.exit_code == 0 { // success
 			if diffcmd in ['code', 'code.cmd'] {
-				// there is no guarantee that the env opts exist
-				// or include `-d`, so (harmlessly) add it
-				return '${diffcmd} ${env_diffopts} -d'
+				cmd := '${diffcmd} ${if env_diffopts != '' { env_diffopts } else { '-d' }}'
+				return cmd
```
spytheman, line 45: The `-d` is needed, if `code` is used as a diff viewer:
```
code --help
...
Options
  -d --diff <file> <file>                    Compare two files with each other.
...
```

spytheman: With -d:
image

Without it, the editor just opens the 2 files in 2 separate tabs.

ttytm: Why is this a place where the function should try to "outsmart" the user and add flags to the `env_diffopts`? It would be added if there are no env_diffopts

vlib/v/util/diff/diff.v
```diff
@@ -41,11 +41,11 @@ pub fn find_working_diff_command() !string {
...
 			}
-			return '${diffcmd} ${env_diffopts}'
+			// Don't add spaces to the cmd if there are no `env_diffopts`.
+			return if env_diffopts != '' { '${diffcmd} ${env_diffopts}' } else { diffcmd }
```
spytheman, line 48: Nice.

vlib/v/util/diff/diff.v
ttytm: Okay I see, this should be added as it's just the diff command for code. Sorry

ttytm:
Let's please prefer #21247 since conflicts will have to be resolved due to test extension. Should cover potential mac CI failure as well.

spytheman:
https://github.com/vlang/v/pull/21247 is merged now.

spytheman (approved):

PR #21243, opened 2024-04-09
v.util: improve `color_compare_files`

ttytm:

### DIFF
--- a/vlib/v/util/diff/diff.v
+++ b/vlib/v/util/diff/diff.v
@@ -65,22 +65,19 @@ fn opendiff_exists() bool {
 	return false
 }
 
-pub fn color_compare_files(diff_cmd string, file1 string, file2 string) string {
-	if diff_cmd != '' {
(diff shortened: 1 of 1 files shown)


spytheman (approved):

spytheman:
I've reverted the changes here, since they break the ability to customize the diff command and options by setting VDIFF_TOOL and VDIFF_OPTIONS.

ttytm:
Might it be that it would work with: https://github.com/vlang/v/pull/21242?

spytheman:
No, I've tested it.
The problem is the `os.find_abs_path_of_executable(diff_cmd) or {` check, which makes an early return, when the passed diff_cmd is not just a name, but something like `code -d`, composed from VDIFF_TOOL and VDIFF_OPTIONS .

spytheman:
I should have been more thorough when reviewing it yesterday, and just not merge it in the first place.

ttytm:
Good point. But easily fixed with keeping the check.

PR #21244, opened 2024-04-09
v.util: ensure output of `color_compare_files` is colorized

ttytm:
Makes the function do what it's name implies. I was really missing this when there is just diff installed and working in the codebase.

|Current|Updated|
|-|-|
| Screenshot_20240409_195735|Screenshot_20240409_195741|

Due to the ansi codes in the output, it will conflict with #21242. Depending on which one is resolved first conflicts in the other will need to be taken care of.

### DIFF
--- a/vlib/v/util/diff/diff.v
+++ b/vlib/v/util/diff/diff.v
@@ -40,6 +40,11 @@ pub fn find_working_diff_command() !string {
 			return error('could not find specified VDIFF_TOOL ${diffcmd}')
 		}
 		if p.exit_code == 0 { // success
+			if diffcmd in ['gdiff', 'diff'] {
+				if p.output.contains('GNU diffutils') && env_diffopts == '' {
+					return '${diffcmd} --color=always'
+				}
+			}
 			if diffcmd in ['code', 'code.cmd'] {
 				// there is no guarantee that the env opts exist
 				// or include `-d`, so (harmlessly) add it


vlib/v/util/diff/diff.v
```diff
@@ -72,9 +72,9 @@ pub fn color_compare_files(diff_cmd string, path1 string, path2 string) string {
 	flags := $if openbsd {
 		['-d', '-a', '-U', '2']
 	} $else $if freebsd {
-		['--minimal', '--text', '--unified=2']
+		['--minimal', '--text', '--unified=2', '--color=always']
```
spytheman, line 75: FreeBSD's diff tool does not support `--color=always`:
```
0[09:12:55]ec2-user@freebsd: (util/fix-diff-color) ~/v $ v run diff.v
diff: unrecognized option `--color=always'
usage: diff [-aBbdilpTtw] [-c | -e | -f | -n | -q | -u] [--ignore-case]
            [--no-ignore-case] [--normal] [--strip-trailing-cr] [--tabsize]
            [-I pattern] [-L label] file1 file2
       diff [-aBbdilpTtw] [-I pattern] [-L label] [--ignore-case]
            [--no-ignore-case] [--normal] [--strip-trailing-cr] [--tabsize]
            -C number file1 file2
       diff [-aBbdiltw] [-I pattern] [--ignore-case] [--no-ignore-case]
            [--normal] [--strip-trailing-cr] [--tabsize] -D string file1 file2
       diff [-aBbdilpTtw] [-I pattern] [-L label] [--ignore-case]
            [--no-ignore-case] [--normal] [--tabsize] [--strip-trailing-cr]
            -U number file1 file2
       diff [-aBbdilNPprsTtw] [-c | -e | -f | -n | -q | -u] [--ignore-case]
            [--no-ignore-case] [--normal] [--tabsize] [-I pattern] [-L label]
            [-S name] [-X file] [-x pattern] dir1 dir2
       diff [-aBbditwW] [--expand-tabs] [--ignore-all-blanks]
            [--ignore-blank-lines] [--ignore-case] [--minimal]
            [--no-ignore-file-name-case] [--strip-trailing-cr]
            [--suppress-common-lines] [--tabsize] [--text] [--width]
            -y | --side-by-side file1 file2
0[09:12:57]ec2-user@freebsd: (util/fix-diff-color) ~/v $
```

spytheman:
I've reverted bfd53a9 in 056a443. The reason is that `find_working_diff_command` can be used to form a diff command, customizable by setting VDIFF_TOOL and VDIFF_OPTIONS, that later can be passed to `util.color_compare_files`:

```v
import v.util
import v.util.diff
diff_cmd := diff.find_working_diff_command()!
res := util.color_compare_files(diff_cmd, 'v.mod', 'v2.mod')
println(res)
```

image

ttytm:
I think this still can be used with https://github.com/vlang/v/pull/21242 and that the PR does not need to be reverted or this one closed

ttytm:
What command does it use for you? Do you have colordiff additionally installed?
It would be nice to get color with the default system diff tool as well that is just `diff`

spytheman:
> I think this still can be used with #21242 and that the PR does not need to be reverted or this one closed

I disagree. It represents a regress in functionality, and a break on FreeBSD (whose diff tool does not support color output at all).

spytheman:
> What command does it use for you? Do you have colordiff additionally installed? It would be nice to get color with the default system diff tool as well that is just `diff`

The list of commonly used diff commands is in `known_diff_tools` in vlib/v/util/diff/diff.v:18:
['colordiff', 'gdiff', 'diff', 'colordiff.exe', 'diff.exe', 'opendiff', 'code', 'code.cmd']

It would be nice, for the diff to work *when it is supported*, and to be customizable when the user has more sophisticated tools and needs (which currently can be done by setting VDIFF_TOOL and VDIFF_OPTIONS).

That can be done by changing `find_working_diff_command`, which can add options for the default commands, if it detects that they are supported.

spytheman:
image

Something like this for example works on FreeBSD too (if the user has installed `diffutils`).

ttytm:
> > What command does it use for you? Do you have colordiff additionally installed? It would be nice to get color with the default system diff tool as well that is just `diff`
> 
> The list of commonly used diff commands is in `known_diff_tools` in vlib/v/util/diff/diff.v:18: ['colordiff', 'gdiff', 'diff', 'colordiff.exe', 'diff.exe', 'opendiff', 'code', 'code.cmd']
> 
> It would be nice, for the diff to work _when it is supported_, and to be customizable when the user has more sophisticated tools and needs (which currently can be done by setting VDIFF_TOOL and VDIFF_OPTIONS).
> 
> That can be done by changing `find_working_diff_command`, which can add options for the default commands, if it detects that they are supported.

Yes I wanted to take the diff module there. Though thought it's organic and sophisticated doing it step by step

spytheman:
> Do you have colordiff additionally installed?

Yes, on both Linux and on FreeBSD, but I renamed it while testing.

ttytm:
Awesome fix thanks 👍

ttytm:
I was sloppy here with reviewing as well. With two things 

From research and the tests before I knew that it won't work to put the flag at the beginning. So the check would have to be in the color_compare function.
My initial implementation was on mac which I tried to cover. It's not covered here. 

Improving the module while adding test we can eventually get there I believe.

PR #21246, opened 2024-04-10
vdoc: fix out path if `-o` is not passed but input has a path, add test

ttytm:
docs should still land in `<path>/_docs` in such cases. Sorry for the regression


spytheman (approved):
Good work.

PR #21251, opened 2024-04-11
ci: prefer dedicated tiggers for platform workflows 

ttytm:
Had a chat about this recently on discord with spy. 

One big reason to prefer this is that it's currently not possible to rerun single platform CI while the caller (cross-platform) workflow is running. It should outweigh the grouping / use of one concurrency group. In any case, the changes are uncomplicated. It can certainly be good to go with it at least for a while and draw further conclusions.


spytheman (approved):
Thank you.
Excellent work.

PR #21256, opened 2024-04-11
cgen: optimize .map(), .any(), .filter() and .all() when using closure

### LINKED ISSUE #21199: optimisation: array filter with a closure, may generate 1 closure per iteration, instead of just 1 closure and then calling it multiple times
**V doctor:**
```
V full version: V 0.4.5 872bcbc.b931e83
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-04-06 13:26:11

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.14-20-ga7d68cbc (5 commit(s) behind V master)
.git/config present: true

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg a_new_closure_fn_in_filter_created_on_each_iteration.v`
```v
x := 42
a := []int{len: 20, init: index*5}
res := a.filter(fn [x](i int) bool {
	println('x: $x | i: $i')
	return i < 20
})
dump(res)

```

A compiled program, where only 1 closure is created, and then it is called multiple times by the a.filter()

**What did you see instead?**
```
x: 42 | i: 0
x: 42 | i: 5
x: 42 | i: 10
x: 42 | i: 15
x: 42 | i: 20
x: 42 | i: 25
x: 42 | i: 30
x: 42 | i: 35
x: 42 | i: 40
x: 42 | i: 45
x: 42 | i: 50
x: 42 | i: 55
x: 42 | i: 60
x: 42 | i: 65
x: 42 | i: 70
x: 42 | i: 75
x: 42 | i: 80
x: 42 | i: 85
x: 42 | i: 90
x: 42 | i: 95
[a_new_closure_fn_in_filter_created_on_each_iteration.v:7] res: [0, 5, 10, 15]
```

It works, but currently it is not efficient, since the closure is created once per each iteration that the filter call generates, even though the context is not changed:
```c
        Array_int _t2 = {0};
        Array_int _t2_orig = a;
        int _t2_len = _t2_orig.len;
        _t2 = __new_array_noscan(0, _t2_len, sizeof(int));

        for (int _t3 = 0; _t3 < _t2_len; ++_t3) {
                int it = ((int*) _t2_orig.data)[_t3];
                bool (*_t4) (int) =             __closure_create(anon_fn_86964c176b0df002_int__bool_78, (struct _V_anon_fn_86964c176b0df002_int__bool_78_Ctx*) memdup_uncollectable(&(struct _V_anon_fn_86964c176b0df002_int__bool_78_Ctx){.x = x,
                }, sizeof(struct _V_anon_fn_86964c176b0df002_int__bool_78_Ctx)));
                if (_t4(it)) {
                        array_push_noscan((array*)&_t2, &it);
                }
        }
        Array_int res =_t2;
```

It would have been more efficient, to do the closure creation just once,
before the loop starts, then the loop can just call it, without creating more closures.

In this context, it is also possible to free the closure too, after the loop ends, since it will not be needed anymore, and it is guaranteed, that it does not escape:
```c
        bool (*_t4) (int) = __closure_create(anon_fn_86964c176b0df002_int__bool_78, (struct _V_anon_fn_86964c176b0df002_int__bool_78_Ctx*) memdup_uncollectable(&(struct _V_anon_fn_86964c176b0df002_int__bool_78_Ctx){.x = x, }, sizeof(struct _V_anon_fn_86964c176b0df002_int__bool_78_Ctx)));

        for (int _t3 = 0; _t3 < _t2_len; ++_t3) {
                int it = ((int*) _t2_orig.data)[_t3];
                if (_t4(it)) {
                        array_push_noscan((array*)&_t2, &it);
                }
        }
```

> [!NOTE]
>

felipensp:
Fix #21199

### DIFF
--- a/vlib/v/gen/c/array.v
+++ b/vlib/v/gen/c/array.v
@@ -463,15 +463,19 @@ fn (mut g Gen) array_init_with_fields(node ast.ArrayInit, elem_type Type, is_amp
 	}
 }
 
-fn (mut g Gen) write_closure_fn(mut expr ast.AnonFn, var_name string) {
-	past := g.past_tmp_var_new()
-	fn_ptr_name := g.fn_var_signature(expr.decl.return_type, expr.decl.params.map(it.typ),
-		past.tmp_var)
-	g.write('${fn_ptr_name} = ')
-	g.gen_anon_fn(mut expr)
-	g.writeln(';')
-	g.past_tmp_var_done(past)
-	g.write('(${var_name})') // usually `it`
+fn (mut g Gen) write_closure_fn(mut expr ast.AnonFn, var_name string, declared_var string) {
+	if declared_var == '' {
+		past := g.past_tmp_var_new()
+		fn_ptr_name := g.fn_var_signature(expr.decl.return_type, expr.decl.params.map(it.typ),
+			past.tmp_var)
+		g.write('${fn_ptr_name} = ')
+		g.gen_anon_fn(mut expr)
+		g.writeln(';')
+		g.past_tmp_var_done(past)
+		g.write('(${var_name})') // usually `it`
+	} else {
+		g.write('${declared_var}(${var_name})')
+	}
 }
 
 // `nums.map(it % 2 == 0)`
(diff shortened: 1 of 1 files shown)


spytheman (approved):
Excellent work.

PR #21259, opened 2024-04-11
vpm: check for git version before adding `--also-filter-submodules` flag

ttytm:
The PR should resolve https://github.com/vlang/v/pull/21216#issuecomment-2049974869


Wertzui123:
Works like a charm, thank you!

PR #21260, opened 2024-04-12
v.util: fix diff coloring, add test

ttytm:
The PR should fix coloring when using a system default diff comparison and adds a test if the output actually includes color codes.

### DIFF
--- a/vlib/v/util/diff/diff.v
+++ b/vlib/v/util/diff/diff.v
@@ -45,7 +45,7 @@ pub fn find_working_diff_command() !string {
 pub fn color_compare_files(diff_cmd string, path1 string, path2 string) string {
 	cmd := diff_cmd.all_before(' ')
 	os.find_abs_path_of_executable(cmd) or { return 'comparison command: `${cmd}` not found' }
-	mut flags := $if openbsd {
+	flags := $if openbsd {
 		['-d', '-a', '-U', '2']
 	} $else $if freebsd {
 		['--minimal', '--text', '--unified=2']
@@ -53,10 +53,10 @@ pub fn color_compare_files(diff_cmd string, path1 string, path2 string) string {
 		['--minimal', '--text', '--unified=2', '--show-function-line="fn "']
 	}
 	if cmd == 'diff' {
-		color_flag := '--color=always'
-		if !os.execute('${cmd} ${color_flag}').output.contains(color_flag) {
-			// If the flag is unknown, it will be reprinted in the output.
-			flags << color_flag
+		color_diff_cmd := '${diff_cmd} --color=always ${flags.join(' ')} ${os.quoted_path(path1)} ${os.quoted_path(path2)}'
+		color_result := os.execute(color_diff_cmd)
+		if !color_result.output.starts_with('diff: unrecognized option') {
+			return color_result.output.trim_right('\r\n')
 		}
 	}
 	full_cmd := '${diff_cmd} ${flags.join(' ')} ${os.quoted_path(path1)} ${os.quoted_path(path2)}'
--- a/vlib/v/util/diff/diff.v
+++ b/vlib/v/util/diff/diff.v
@@ -31,14 +31,6 @@ pub fn find_working_diff_command() !string {
 			continue
 		}
 		if p.exit_code == 0 { // success
-			// TODO: proper implemenation of --color flag
-			$if !macos {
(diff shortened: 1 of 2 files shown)


vlib/v/util/diff/diff_test.v
```diff
@@ -60,3 +64,21 @@ fn test_compare_files() {
 	assert res.contains("+\tversion: '0.1.0'"), res
 	assert res.contains("+\tlicense: 'MIT'"), res
 }
+
+fn test_coloring() {
+	color_flag := '--color=always'
+	if os.execute('diff ${color_flag}').output.contains(color_flag) {
+		// If the flag is unknown, it will be reprinted in the output.
+		eprintln('> skipping test, since `diff` does not support `${color_flag}`')
```
ttytm, line 72: We could assert false if it reaches here in Linux CI. This might guard even better against regressions.

vlib/v/util/diff/diff.v
```diff
@@ -51,16 +43,22 @@ pub fn find_working_diff_command() !string {
...
 	} $else {
 		['--minimal', '--text', '--unified=2', '--show-function-line="fn "']
 	}
+	if cmd == 'diff' {
+		color_flag := '--color=always'
+		if !os.execute('${cmd} ${color_flag}').output.contains(color_flag) {
+			// If the flag is unknown, it will be reprinted in the output.
+			flags << color_flag
+		}
+	}
```
spytheman, line 61: on FreeBSD:
```
0[10:37:40]ec2-user@freebsd:  ~ $ diff --color=always
diff: unrecognized option `--color=always'
usage: diff [-aBbdilpTtw] [-c | -e | -f | -n | -q | -u] [--ignore-case]
            [--no-ignore-case] [--normal] [--strip-trailing-cr] [--tabsize]
            [-I pattern] [-L label] file1 file2
       diff [-aBbdilpTtw] [-I pattern] [-L label] [--ignore-case]
            [--no-ignore-case] [--normal] [--strip-trailing-cr] [--tabsize]
            -C number file1 file2
       diff [-aBbdiltw] [-I pattern] [--ignore-case] [--no-ignore-case]
            [--normal] [--strip-trailing-cr] [--tabsize] -D string file1 file2
       diff [-aBbdilpTtw] [-I pattern] [-L label] [--ignore-case]
            [--no-ignore-case] [--normal] [--tabsize] [--strip-trailing-cr]
            -U number file1 file2
       diff [-aBbdilNPprsTtw] [-c | -e | -f | -n | -q | -u] [--ignore-case]
            [--no-ignore-case] [--normal] [--tabsize] [-I pattern] [-L label]
            [-S name] [-X file] [-x pattern] dir1 dir2
       diff [-aBbditwW] [--expand-tabs] [--ignore-all-blanks]
            [--ignore-blank-lines] [--ignore-case] [--minimal]
            [--no-ignore-file-name-case] [--strip-trailing-cr]
            [--suppress-common-lines] [--tabsize] [--text] [--width]
            -y | --side-by-side file1 file2
```

spytheman: On Linux:
```
#2 10:38:30 ᛋ master /v/oo❱diff --color=always
diff: missing operand after '--color=always'
diff: Try 'diff --help' for more information.
```

ttytm: That's the way it should be

spytheman: No, note how on freebsd, the flag *is* still present in the output, even though the diff command does not support it.

ttytm: That means is available on your mac

spytheman: The presence or absence of it is not a reliable way to do feature detection.

spytheman: An analysis of the output of `diff --help` may be better.

spytheman: Both Linux and Freebsd do have `--color=always` in the output of `diff --color=always`. macos's diff does not have it.

ttytm: Thanks for the review. I see it's problematic on linux.

Then running the command with `--color=always` once and if it fails running it again without and returning the output would likely be the best thing.

spytheman: Yes, I also think, that running it twice will be more reliable.

PR #21262, opened 2024-04-12
 checker, cgen: allow op overload for type with generic parent 

Delta456:
This PR compiles the following code:

```v
import math.vec


type Vec3 = vec.Vec3[f64]

fn main() {
	a := Vec3{1.0, 2.0, 3.0}
	b := Vec3{0.5, 1.5, 2.5}
	println(a - b)
}
```

Previously it wasn't working.


spytheman (approved):
Excellent work.

PR #21268, opened 2024-04-12
ci: remove duplicate `test-cleancode` in linux tcc runs

ttytm:

### DIFF
--- a/.github/workflows/linux_ci.yml
+++ b/.github/workflows/linux_ci.yml
@@ -39,7 +39,6 @@ jobs:
         run: |
           echo $VFLAGS
           make
-          ./v test-cleancode
           ./v -d debug_malloc -d debug_realloc -o v cmd/v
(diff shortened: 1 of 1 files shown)


.github/workflows/linux_ci.yml
```diff
@@ -39,7 +39,6 @@ jobs:
         run: |
           echo $VFLAGS
           make
-          ./v test-cleancode
```
spytheman, line 42: No. That is here on purpose, so that it can fail as fast as possible.

It only adds a few seconds to the run, while it prevents much more expensive checks done for code that is not even formatted properly.

ttytm: I see. That's the other option I though why it might be there. But the more expensive checks still run for gcc and clang. Failing here in tcc won't interrupt them. 

If it makes sense to have formatting verification run before the other cc CI I'd submit something here.

PR #21271, opened 2024-04-12
util/diff: return diff options with the diff command for FreeBSD/OpenBSD

kimshrier:
When returning the diff command for FreeBSD or OpenBSD, the diff options from the environment were not included as they are for other OS's.

This fixes a broken unit test.

### DIFF
--- a/vlib/v/util/diff/diff.v
+++ b/vlib/v/util/diff/diff.v
@@ -23,7 +23,7 @@ pub fn find_working_diff_command() !string {
 		}
 		$if freebsd || openbsd {
 			if diffcmd == 'diff' { // FreeBSD/OpenBSD diff have no `--version` option
-				return '${diffcmd} ${env_diffopts}'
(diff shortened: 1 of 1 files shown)


vlib/v/util/diff/diff.v
```diff
@@ -23,7 +23,7 @@ pub fn find_working_diff_command() !string {
 		}
 		$if freebsd || openbsd {
 			if diffcmd == 'diff' { // FreeBSD/OpenBSD diff have no `--version` option
-				return diffcmd
+				return '${diffcmd} ${env_diffopts}'
```
ttytm, line 26:
```suggestion
				return if env_diffopts != '' { '${diffcmd} ${env_diffopts}' } else { diffcmd }
```
to not add spaces to the diffcmd if there are not env_diffopts as it can lead to unpredicted behaviour.

ttytm (approved):

PR #21276, opened 2024-04-13
cgen: fix code generated for `none` passed to generic option casting

### LINKED ISSUE #21215: generic code struggles with optional function args
### Describe the bug

Calling generic higher order functions with optional callbacks leads to all kinds of C compile and runtime errors. I have tried to provide a concrete example that is easy to understand, but this gives issues completely different to the issues I was originally running into. This case still seem like a simpler place to start, before I try and build SSCCEs for the other issues I found.

### Reproduction Steps

```v
pub type MyFn[T] = fn (mut my_state T)

fn higher_order_fn[T](initial_state T, maybe_callback ?MyFn[T]) State {
    mut my_state := initial_state
    if callback := maybe_callback {
        callback[T](mut my_state)
        callback[T](mut my_state)
    }
    return my_state
}

pub struct State {
mut:
    x int
}

fn handler(mut my_state State) {
    my_state.x += 1
}

fn main() {
    mut state := State { }
    state = higher_order_fn[State](state, ?MyFn[State](handler))
    state = higher_order_fn[State](state, ?MyFn[State](none))
    state = higher_order_fn[State](state, handler)
    state = higher_order_fn[State](state, none)
    println("${state}")
}
```

### Expected Behavior

I expect the above code to work, and to print `State { x: 4 }`. I also expect any subset of those four lines to work.

### Current Behavior

Many things go wrong.

The first thing seems intentional:
```sh
higher_order_fn_error2.v:26:43: error: cannot use `none` as generic argument
   24 |     state = higher_order_fn[State](state, ?MyFn[State](none))
   25 |     state = higher_order_fn[State](state, handler)
   26 |     state = higher_order_fn[State](state, none)
      |                                           ~~~~
   27 |     println("${state}")
   28 | }
```
Fine, so let's remove line 26.

Now it compiles, but crashes:
```sh
Unhandled Exception 0xC0000005
C.CaptureStackBackTrace returned less than 2 frames
signal 11: segmentation fault
6 : print_backtrace_skipping_top_frames  ?? : address = 0x2a1510514d0
5 : seh_filter_exe             ?? : address = 0x2a1510515a0
4 : `__scrt_common_main_seh'::`1'::filt$0  D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:304
3 : _C_specific_handler        ?? : address = 0x2a151051690
2 : _chkstk                    ?? : address = 0x2a151051510
1 : RtlRaiseException          ?? : address = 0x2a1510516a0
0 : KiUserExceptionDispatcher  ?? : address = 0x2a151051580
```

so remove line 24, and run again:
```v
State{
    x: 4
}
```
It works. Okay, let's remove line 23 now, and just call the function once.

```sh
higher_order_fn_error2.01HTVR8KKCEQBVVH81E32QQZ4E.tmp.c
C:\Users\Owner\AppData\Local\Temp\v_0\higher_order_fn_error2.01HTVR8KKCEQBVVH81E32QQZ4E.tmp.c(12786): error C2440: '=': cannot convert from '_option_anon_fn_mut_main__state' to '_option_anon_fn_mut_t'

builder error: msvc error
```
cgen seems to have screwed something up now. weird!

This may turn out to be one bug or two, but these were just the things I ran into this time. Other times I find these higher order functions work fine, and other times I think I run into other issues. I can experiment more and give more specific bug reports once the general chaos has been fixed a little.

### Possible Solution

Good luck.

Similar to #21214 I encountered these issues trying to make a wrapper for `gg` that handled streaming images automatically. Some apps provide event callbacks, other apps just provide do_frame callbacks with no event callbacks. I got it working, with lots of `[T]` annotations everywhere, but it was a minefield.

### V version

V 0.4.5 ad17be5

Git vroot status: weekly.2024.14-26-gad17be5d (6 commit(s) behind V master)
.git/config present: true

thirdparty/tcc: N/A
> [!NOTE]
>

felipensp:
Fix #21215

### DIFF
--- a/vlib/v/gen/c/assign.v
+++ b/vlib/v/gen/c/assign.v
@@ -71,8 +71,12 @@ fn (mut g Gen) expr_opt_with_cast(expr ast.Expr, expr_typ ast.Type, ret_typ ast.
 		styp := g.base_type(ret_typ)
 		decl_styp := g.typ(ret_typ).replace('*', '_ptr')
 		g.writeln('${decl_styp} ${past.tmp_var};')
-		g.write('_option_ok(&(${styp}[]) {')
-
+		is_none := expr is ast.CastExpr && expr.expr is ast.None
+		if is_none {
+			g.write('_option_none(&(${styp}[]) {')
+		} else {
+			g.write('_option_ok(&(${styp}[]) {')
+		}
 		if expr is ast.CastExpr && expr_typ.has_flag(.option) {
 			g.write('*((${g.base_type(expr_typ)}*)')
 			g.expr(expr)
--- a/vlib/v/tests/option_generic_cast_none_test.v
+++ b/vlib/v/tests/option_generic_cast_none_test.v
@@ -1,0 +1,27 @@
+pub type MyFn[T] = fn (mut my_state T)
+
+fn higher_order_fn[T](initial_state T, maybe_callback ?MyFn[T]) State {
+	mut my_state := initial_state
+	if callback := maybe_callback {
+		callback[T](mut my_state)
+		callback[T](mut my_state)
+	}
+	return my_state
+}
+
+pub struct State {
+mut:
+	x int
+}
+
+fn handler(mut my_state State) {
+	my_state.x += 1
+}
+
+fn test_main() {
+	mut state := State{}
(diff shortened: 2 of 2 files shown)


spytheman (approved):
Excellent work.

PR #21281, opened 2024-04-14
vdoc: improve creation of trimmed node descriptions, extend tests

ttytm:
The PR makes creating trimmed doc node descriptions more flexible by improving detecting the first descriptive line. Currently, it only works with `## Description` - which is not uniformly used in all vlib modules and likely won't be used in all modules outside stdlib that generate their docs wit v doc. 

It also solidifies the mechanism of omitting the module name if the description starts with it, adds documentation to the function and organizes and extend tests.

Results in the rendered docs:

|Current|Updated|
|-|-|
|Screenshot_20240414_191828|Screenshot_20240414_191810|
|Screenshot_20240414_191849|Screenshot_20240414_191803|

### DIFF
--- a/.github/workflows/module_docs_ci.yml
+++ b/.github/workflows/module_docs_ci.yml
@@ -13,6 +13,12 @@ jobs:
       - uses: actions/checkout@v4
       - name: Build V
         run: make
+      - name: Test v doc
+        run: |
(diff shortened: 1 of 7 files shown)


spytheman (approved):
Excellent work.

PR #21283, opened 2024-04-14
checker: allow pass array as mut param to spawn fn

### LINKED ISSUE #21280: spawn a function with mut array not allowed
### Describe the bug

When trying to pass a mutable array to a `spawn` function call, the compiler tells to use references and when trying to use a reference of the array the compiler tells that arrays are already references so we need to use mut instead.

### Reproduction Steps

These are the two cases:
```v
fn mutable_array_arg(mut a []int) {
  println(a)
}

fn main() {
  mut a := []int{}
  spawn mutable_array_arg(mut a)
}
```

```v
fn ref_array_arg(a &[]int) {
  println(a)
}

fn main() {
  mut a := []int{}
  spawn ref_array_arg(&a)
}
```

### Expected Behavior

To be able to pass the mut array as an argument to the function. (or the compiler telling me how to do it)

### Current Behavior

```sh
t.v:7:31: error: function in `spawn` statement cannot contain mutable non-reference arguments
    5 | fn main() {
    6 |   mut a := []int{}
    7 |   spawn mutable_array_arg(mut a)
      |                               ^
    8 | 
    9 | }
```

```sh
t.v:1:25: error: V arrays are already references behind the scenes,
there is no need to use a reference to an array (e.g. use `[]string` instead of `&[]string`).
If you need to modify an array in a function, use a mutable argument instead: `fn foo(mut s []string) {}`.
    1 | fn ref_array_arg(mut a &[]int) {
      |                         ^
    2 |   println(a)
    3 | }
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.5 149cbba

V full version: V 0.4.5 294f7e4.149cbba
OS: linux, Linux version 6.6.13-200.fc39.x86_64 (mockbuild@72f11b2996ed4699b0f705186172808f) (gcc (GCC) 13.2.1 20231205 (Red Hat 13.2.1-6), GNU ld version 2.40-13.fc39) #1 SMP PREEMPT_DYNAMIC Sat Jan 20 18:03:28 UTC 2024
Processor: 4 cpus, 64bit, little endian, Intel(R) Pentium(R) Gold 7505 @ 2.00GHz
> [!NOTE]
>

felipensp:
Fix #21280

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -2514,8 +2514,12 @@ fn (mut c Checker) spawn_expr(mut node ast.SpawnExpr) ast.Type {
 			node.call_expr.or_block.pos)
 	}
 	// Make sure there are no mutable arguments
-	for arg in node.call_expr.args {
+	for mut arg in node.call_expr.args {
 		if arg.is_mut && !arg.typ.is_ptr() {
+			if c.table.final_sym(arg.typ).kind == .array {
+				// allow mutable []array to be passed as mut
+				continue
+			}
 			c.error('function in `spawn` statement cannot contain mutable non-reference arguments',
 				arg.expr.pos())
 		}
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -2514,7 +2514,7 @@ fn (mut c Checker) spawn_expr(mut node ast.SpawnExpr) ast.Type {
 			node.call_expr.or_block.pos)
 	}
 	// Make sure there are no mutable arguments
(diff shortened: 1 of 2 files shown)


JalonSolov:
The only problem with making this work, is that it makes V _less_ safe.  If multiple threads start modifying arrays at the same time, it could wreak all sorts of havok... the _least_ being that it can cause very hard to debug issues, since they won't hit unless multiple threads beat on the array at the same time.

Eliyaan:
But this seem to already work with structs so a choice is to be made

PR #21286, opened 2024-04-15
os: format readme, fix markdown inside html

ttytm:
Auto-formats the os modules readme. Doing so it fixes an issue with markdown inside html tags.
The formatting changes match with what is used for the bigger part other markdown files in the codebase.

Following commonmarks recommended formatting for markdown inside html tags works here as a fix. Stackoverflow answer with link to commonmark example: https://stackoverflow.com/a/29378250

|Current|After formatting|
|-|-|
|Screenshot_20240415_084810|Screenshot_20240415_084800|

<details>

Putting markdown immediately after a tag **never works**(at least as far as I know). It even breaks highlighting in the example below. It **can work** to have a tag immediately after markdown. The latter stopped working for vdoc after the recent fixes regarding indentation and escape characters. The recommended way is to keep a line-break before and after tags. Maybe following recommended commonmark formatting (which is applied by markdown formatters) is good enough as a fix now.

|MD immediatly after opening tag|MD followed by tag|
|-|-|
|Screenshot_20240415_090704|Screenshot_20240415_090711|

</details>

### DIFF
--- a/vlib/os/README.md
+++ b/vlib/os/README.md
@@ -8,16 +8,16 @@ handling processes etc.
 
 ### Security advice related to TOCTOU attacks
 
-A few `os` module functions can lead to the <b>TOCTOU</b> vulnerability if used incorrectly.
-<b>TOCTOU</b> (Time-of-Check-to-Time-of-Use problem) can occur when a file, folder or similar
+A few `os` module functions can lead to the **TOCTOU** vulnerability if used incorrectly.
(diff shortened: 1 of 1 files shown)


ttytm:
In terms of a "fix" the PR will not be required anymore after https://github.com/vlang/v/pull/21287#issuecomment-2056342825. If desired, the formatting here can still be done for the sake of formatting.

spytheman (approved):

PR #21293, opened 2024-04-16
time: update unix time acces, fix issues related to deviating unix times

ttytm:
The PR fixes issues like #17162 and #17311

The fix does NOT add the `@[noinit]` attribute to the Time struct, as doing so would break its use in other structs, such as:

```v
struct Foo {
	t time.Time
}

foo := Foo{} // would not be possible anymore if Time is turned into a noinit struct
```

- The problem is fixed by ensuing explicitness - while still keeping the possibility the to create Time structs. 

  Only the `unix` field is turned into a private field. Not being able to init it with other fields resolves their interferes. Accessing and explicitly setting the `unix` field is possible via `unix()` function and method. This approach should provide a better predictable result and more general flexibility in comparison to a `@[noinit]` solution.

- The next change is making sure that the unix time is present or calculated when a time is used.  

- Unix methods have been updated in the go spirit. Former methods have been deprecatd.
  ```
  `unix_time` -> `unix`
  `unix_time_milli` -> `unix_milli`
  ```
  
While initially planned making it into a `v fmt` autofix, types of selector expressions are not parsed and not accessible via vfmt. Therefore I'm making the access of a `.unix` file into a checker warning, informing that it'll become an error and recommending to use `unix()` instead.

ref. https://discord.com/channels/592103645835821068/592320321995014154/1229503437029380107

### DIFF
--- a/vlib/time/time.v
+++ b/vlib/time/time.v
@@ -159,20 +159,19 @@ pub fn (t Time) unix_time_nano() i64 {
 
 // add returns a new time with the given duration added.
 pub fn (t Time) add(duration_in_nanosecond Duration) Time {
-	t_with_unix := time_with_unix(t)
 	// This expression overflows i64 for big years (and we do not have i128 yet):
 	// nanos := t.unix * 1_000_000_000 + i64(t.nanosecond) <-
 	// ... so instead, handle the addition manually in parts ¯\_(ツ)_/¯
 	mut increased_time_nanosecond := i64(t.nanosecond) + duration_in_nanosecond.nanoseconds()
 	// increased_time_second
-	mut increased_time_second := t_with_unix.unix + (increased_time_nanosecond / second)
+	mut increased_time_second := t.unix() + (increased_time_nanosecond / second)
 	increased_time_nanosecond = increased_time_nanosecond % second
 	if increased_time_nanosecond < 0 {
 		increased_time_second--
 		increased_time_nanosecond += second
 	}
 	res := unix_nanosecond(increased_time_second, int(increased_time_nanosecond))
-	return if t_with_unix.is_local { res.as_local() } else { res }
+	return if t.is_local { res.as_local() } else { res }
 }
 
 // add_seconds returns a new time struct with an added number of seconds.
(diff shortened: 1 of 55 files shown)


vlib/time/time.v
```diff
@@ -99,56 +104,85 @@ pub fn (t Time) smonth() string {
...
+// unix returns the UNIX time with second resolution.
 @[inline]
+pub fn (t Time) unix() i64 {
+	return new_time(t).unix
+}
+
+// unix_milli returns the UNIX time with millisecond resolution.
+@[inline]
+pub fn (t Time) unix_milli() i64 {
+	return new_time(t).unix * 1_000 + (i64(t.nanosecond) / 1_000_000)
```
JalonSolov, line 116:
```suggestion
	return new_time(t).unix() * 1_000 + (i64(t.nanosecond) / 1_000_000)
```

vlib/time/time.v
```diff
@@ -99,56 +104,85 @@ pub fn (t Time) smonth() string {
...
+	// TODO: use i128 here, when V supports it, since the following expression overflows for years like 3001:
+	return new_time(t).unix * 1_000_000_000 + i64(t.nanosecond)
+}
+
+// unix_time returns the UNIX time with second resolution.
+@[deprecated: 'use `t.unix()` instead']
+@[deprecated_after: '2024-05-31']
 pub fn (t Time) unix_time() i64 {
-	return t.unix
+	return t.unix()
```
JalonSolov, line 136: Should still work without the `()` since it's in the same module... correct?

ttytm: Yes, correct

ttytm: But we can keep it here then it calls the function to ensure it is calculated. So we don't need to wrap it in new() / new_time.

JalonSolov: Would it make sense, then, to change the other `t.unix` to `t.unix()`?

ttytm: Where we can't be sure if an expression has a unix time, this definitively makes sense.

I hope that I applied it now everywhere where it is required to ensure robustness for all cases and only left where internal direct access makes total sense (e.g. when using an internal variable has a `now()` time value - which certainly will have the unix time on it).

PR #21298, opened 2024-04-17
vdoc: cleanup `vdoc_file_test.v`

ttytm:
Clarifies things to simplify maintaining and extending the doc tests. 
Initially, part of #21281 but was ultimately omitted for a simpler diff. 
Now, I'm submitting it separately from an update/fix that includes adding new tests.


spytheman (approved):
Excellent work.

PR #21301, opened 2024-04-17
vpm: revert deep clones

ttytm:
The suggestion keeps shallow cloning since it can make a tremendous difference to clone a repo just one commit deep.

The issue of not being able to run the sdl script after a `v install sdl` is addressed here: https://github.com/vlang/sdl/pull/742

### DIFF
--- a/cmd/tools/vpm/outdated_test.v
+++ b/cmd/tools/vpm/outdated_test.v
@@ -15,7 +15,6 @@ fn testsuite_begin() {
 	test_utils.set_test_env(test_path)
 	os.mkdir_all(test_path)!
 	os.chdir(test_path)!
-	println('test_path: ${test_path}')
 }
 
 fn testsuite_end() {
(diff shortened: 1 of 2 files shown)


spytheman:
I strongly prefer the simplicity of not using `--depth=1` as a default. 

It is much more reliable, in many ways, and makes working within modules installed with `v install name` more convenient, for people that do not necessarily know git well enough, to know how to fix their problems by using `git fetch --unshallow` etc. 

The few seconds, that will be saved with `--depth=1` are an optimization that is costly, and may pay for bigger repos with many commits, but imho should be opt in (with something put in v.mod for example), and *not* the default.

spytheman:
You can also see my comments about `--depth=1` here: <https://github.com/vlang/v/commit/27cc27737a7e63745f6aa9239996e3b1098c8097> .

ttytm:
Yep, it's okay. Lets do it this way.

> You can also see my comments about --depth=1 here: https://github.com/vlang/v/commit/27cc27737a7e63745f6aa9239996e3b1098c8097 .

Responded there.

PR #21302, opened 2024-04-17
ci: continue testing other modules if one fails

ttytm:
It makes sense to continue testing other modules if one fails to build. This should give a better understanding of what is wrong and have more insight where fixes can happen.

Example run:
https://github.com/ttytm/v/actions/runs/8727507855/job/23945173601


Delta456:
Indeed, it was a required change.

Delta456 (approved):

spytheman (approved):
Excellent 🚀 🥳.

PR #21307, opened 2024-04-18
cli: update `command_test.v`

ttytm:

### DIFF
--- a/vlib/cli/command_test.v
+++ b/vlib/cli/command_test.v
@@ -6,7 +6,7 @@ fn test_if_command_parses_empty_args() {
 		execute: empty_func
 	}
 	cmd.parse(['command'])
-	assert cmd.name == 'command' && compare_arrays(cmd.args, [])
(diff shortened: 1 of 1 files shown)


vlib/cli/command_test.v
```diff
@@ -6,7 +6,7 @@ fn test_if_command_parses_empty_args() {
 		execute: empty_func
 	}
 	cmd.parse(['command'])
-	assert cmd.name == 'command' && compare_arrays(cmd.args, [])
+	assert cmd.name == 'command' && cmd.args == []string{}
```
spytheman, line 9:
```suggestion
	assert cmd.name == 'command' && cmd.args.len == 0
```

spytheman (approved):
Excellent work. Deleting code, while keeping the functionality, and improving the readability 🥳.

vlib/cli/command_test.v
ttytm: Thanks !

PR #21308, opened 2024-04-18
cli: extend control over defaults

ttytm:
The PR extends functionality of the cli module to allow more control over defaults.

Creating an application with the cli module and default commands and will result in something like:

```
Usage: <cool_upcoming_app> [flags] [commands] <paths>

Flags:
  -i  --include       Lorem ipsum dolor.
  -h  --help          Prints help information.
  -v  --version       Prints version information.
      --man           Prints the auto-generated manpage.

Commands:
  help                Prints help information.
  version             Prints version information.
  man                 Prints the auto-generated manpage.
```

An example result with more direct control over defaults could look like (keeps some default flags):
```
Usage: <cool_upcoming_app> [flags] <paths>

Flags:
  -i  --include       Lorem ipsum dolor.
  -h  --help          Prints help information.
  -v  --version       Prints version information.
```

Current customization possibilities are limited as they only disable flag and command at once.
```v
mut app := cli.Command{
	// ...
	disable_man: true
	disable_help: true
	disable_version: true
}
```
The updated functionality would allow full control
```v
mut app := cli.Command{
	// ...
	defaults: struct{
		man: false
		help: cli.CommandFlag{false,true} // Fine grained control is also possible.
		version: cli.CommandFlag{false,true}
	}

}
```

The current `disable_man`, `disable_help`, `disable_version` flags are deprecated but will continue to work.

I had this potential change sitting since October but didn't submit it because there were some issues regarding its implementation. After the recent development and since private fields are now actually treated as private to a module, this change can work out. Also, I find myself in the situation again where I'd like to be able to easily disable default flags or commands but not both when working with the cli module. So I'd like to give it a shot.

Would just need to be refined and covered with a test.

### DIFF
--- a/vlib/cli/command.v
+++ b/vlib/cli/command.v
@@ -21,9 +21,9 @@ pub mut:
 	pre_execute     FnCommandCallback = unsafe { nil }
 	execute         FnCommandCallback = unsafe { nil }
 	post_execute    FnCommandCallback = unsafe { nil }
-	disable_help    bool
-	disable_man     bool
-	disable_version bool
+	disable_help    bool              @[deprecated: 'use defaults.help instead'; deprecated_after: '2024-06-31']
+	disable_man     bool              @[deprecated: 'use defaults.man instead'; deprecated_after: '2024-06-31']
+	disable_version bool              @[deprecated: 'use defaults.version instead'; deprecated_after: '2024-06-31']
 	disable_flags   bool
 	sort_flags      bool
 	sort_commands   bool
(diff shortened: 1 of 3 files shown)


vlib/cli/command.v
```diff
@@ -129,12 +150,60 @@ pub fn (mut cmd Command) add_flag(flag Flag) {
 	cmd.flags << flag
 }
 
-// parse parses `args` into this structured `Command`.
-pub fn (mut cmd Command) parse(args []string) {
+// TODO: remove deprecated `disable_<>` switches after deprecation period.
+fn (mut cmd Command) parse_defaults() {
+	// Help
+	if cmd.defaults.help is bool {
```
spytheman, line 156: A exhaustive type match like this `match cmd.defaults.help { bool {...} CommandFlag {} }`, is usually preferable, since if we later change the sumtype to `type Defaults = CommandFlag | bool | string`, the code will break in exactly the places, that need modification.

PR #21311, opened 2024-04-18
tools: update vretry.v, add test

ttytm:
Next to minor updates to `vretry` it uses a default retry count of 10 as it can be a more sensible default in comparison to getting stuck in 999_999_999 loop(/ for 300sec) when passing an invalid command. This default count is similar to `.github/workflows/retry.sh`.
It might be a better fit the other defaults (1 sec delay and timeout of 300sec) as well.

The added test will pass after #21310

### DIFF
--- a/cmd/tools/vretry_test.v
+++ b/cmd/tools/vretry_test.v
@@ -1,0 +1,29 @@
+import os
+
+const vexe = @VEXE
+
+fn test_retry() {
+	tpath := os.join_path(os.vtmp_dir(), 'vretry_test')
+	os.rmdir_all(tpath) or {}
+	os.mkdir_all(tpath)!
+	defer {
+		os.rmdir_all(tpath) or {}
+	}
+
+	fail_cmd := os.execute('${vexe} retry git asdf')
+	assert fail_cmd.exit_code != 0
+	assert fail_cmd.output.contains('error: exceeded maximum number of retries')
+
+	with_flags_fail_cmd := os.execute('${vexe} retry -d 0.2 -r 3 git asdf')
+	assert with_flags_fail_cmd.exit_code != 0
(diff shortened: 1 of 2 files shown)


cmd/tools/vretry_test.v
```diff
@@ -10,14 +10,18 @@ fn test_retry() {
 		os.rmdir_all(tpath) or {}
 	}
 
-	fail_cmd := os.execute('${vexe} retry git asdf')
-	assert fail_cmd.exit_code != 0
-	assert fail_cmd.output.contains('error: exceeded maximum number of retries')
+	if os.getenv('CI') != 'true' {
+		// Skip longer taking test in CI runs.
```
JalonSolov, line 14:
```suggestion
		// Skip longer running test in CI runs.
```

ttytm: ty

cmd/tools/vretry_test.v
```diff
@@ -0,0 +1,36 @@
...
+	os.rmdir_all(tpath) or {}
+	os.mkdir_all(tpath)!
+	defer {
+		os.rmdir_all(tpath) or {}
+	}
+
+	fail_cmd := 'git asdf'
+	if os.getenv('CI') != 'true' {
+		// Skip longer running test in CI runs.
+		res := os.execute('${vexe} retry ${fail_cmd}')
```
spytheman, line 16: That will take ~10 seconds, which is very long for such a test.
I often do just `v test cmd/`, and with it, on my 2 core system, it will in effect double the time needed.

spytheman: The CI check, is also imho in the wrong direction - I would not mind 10 seconds more there, since there are checks that already take hours. I do mind 10 seconds a lot more in an interactive local session, where my attention is focused on the result.

PR #21312, opened 2024-04-18
vdoc: refactor `get_ignore_paths()`

ttytm:
Covered by the recently added tests this PR should refactor the function to a better readable and more performant version prior to extending functionality.


cmd/tools/vdoc/files.v
```diff
@@ -7,23 +7,19 @@ fn get_ignore_paths(path string) ![]string {
...
-			final << rule
+	for rule in rules {
+		if rule.contains('*.') || rule.contains('**') {
+			println('vdoc: Wildcards in ignore rules are not allowed for now.')
+			continue
 		}
-		res = final.map(os.join_path(path, it.trim_right('/')))
-	} else {
-		mut dirs := os.ls(path) or { return []string{} }
-		res = dirs.map(os.join_path(path, it)).filter(os.is_dir(it))
```
spytheman, line 24: what happens for this case in the new version?

ttytm: Thats some empty .vdocignore file magic that isn't applied if the file has contents. So I thought that the same logic would need to be there for doc files with contents if it has sense. Therefore just returning an empty array if the ignore file is empty.

Ofc I might miss a case and I'm open for input to get it covered.

ttytm: An empty vdocignore file should just Ignore all in a directory? If thats it's special power then there is the miss understanding I had.

spytheman: I think you are right - it does ignore everything in this case.
It is not documented, not very intuitive (an empty .gitignore for example does not act like that), and not tested, so I do not think it is used much.

spytheman: In any case, keeping it for now, may be better, so that the PR is a pure refactoring.
We can remove it later in another PR.

ttytm: Yes, maybe moving towards the more common for empty ignore files when there is some kind pattern support implemented.

PR #21313, opened 2024-04-18
ci: prevent unintended deployment workflow steps

ttytm:
- Limits the vc deployment workflow to run when relevant instead of
  > ```yml
  > # Note that failure below may happen, due to vlang/vc rejecting the push
  > # from forked repos; that is not usually a problem.
  > ```
- Extends the doc_ci workflow by adding the default condition that we currently use to make sure it runs when relevant.


spytheman (approved):
Excellent work.

PR #21314, opened 2024-04-19
vdoc: rewrite and extend vdocignore

ttytm:
After working on an improved status quo to also have a good fallback, this PR approaches to rewrite and to extend `.vdocignore`.

- It adds pattern support. 
  
  Until now to ignore `a` and the same pattern in a sub-directory (`b/a`) it is required to set the exact path for every subdirectory. Now it is sufficient to set the pattern

<table width="400">
<tr width="100%">
	<th>
	Ignore Goal
	</th>
	<th>
	Current
	</th>
	<th>
	Updated
	</th>
</tr>

<tr width="100%">
 <td>

<pre>
a
a/b
b/c/a
</pre>

 </td>

 <td>

<pre>
a
a/b
b/c/a
</pre>
 </td>
 <td>

<pre>
a
</pre>
 </td>
</tr>

</table>

  - Similar to gitignore a pattern starting with `/` will be treated as relative pattern.
    -> `/a` will ignore `/a` but not `/b/a`.  
    -> `/c/d` will ignore `/c/d` but not `/b/c/d`

- Treats lines starting with `#` as comments

- Removes the functionality that an empty ignore file ignores all in its directory.

### DIFF
--- a/cmd/tools/vdoc/files.v
+++ b/cmd/tools/vdoc/files.v
@@ -22,24 +22,29 @@ fn get_modules(opath string, mut ignore_rules IgnoreRules) []string {
 	mut res := []string{}
 	for p in os.ls(opath) or { return [] } {
 		fp := os.join_path(opath, p)
-		if fp in ignore_rules.paths
-			|| ignore_rules.patterns.keys().any(p.contains(it.trim_right('/'))) {
+		if fp in ignore_rules.paths {
 			continue
 		}
-		if os.is_dir(os.join_path(opath, p)) {
+		is_dir := os.is_dir(fp)
+		if ignore_rules.patterns.keys().any(p.contains(it)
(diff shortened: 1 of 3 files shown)


JalonSolov:
Would be nice, of course, to get full .gitignore paths, so you could have

**/a/

meaning every `a` subdir... `/a`, `/b/a`, `/1/2/3/4/5/6/7/a`, etc.

At the very least, support trailing slash means subdir, while no trailing slash means file.

ttytm:
> Yep, but a trailing slash forces it to _only_ match subdirs.

Yep, a test for this is added.

To give more info: Our `.vdocignore` is in general a little simpler then a `.gitignore`  since it only looks for modules that end with `.v`. So the test has an ignore pattern `name.v/`, a directory with the name `name.v` and also file `name.v`, testing  that only the directory is ignored.

PR #21317, opened 2024-04-20
mysql: exec_one invalid memory access

sibkod:
When executing the exec_one() command, if a field with a NULL value came across, it caused the RUNTIME ERROR: invalid memory access error


spytheman (approved):
Excellent work.

PR #21321, opened 2024-04-20
v: update linguist languages, add .vdocignore

ttytm:
For highlighting / in some editors comment toggle.

Example in viewing a .vdocignore file on github 
|Current|Updated|
|-|-|
|Screenshot_20240420_131307|Screenshot_20240420_131305|

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,5 +3,6 @@
 *.vsh linguist-language=V text=auto eol=lf
 **/v.mod linguist-language=V text=auto eol=lf
 *.bat text=auto eol=crlf
- Dockerfile.* linguist-language=Dockerfile
 *.toml text eol=lf
+Dockerfile.* linguist-language=Dockerfile
+.vdocignore linguist-language=ignore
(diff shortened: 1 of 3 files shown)


.gitattributes
```diff
@@ -1,7 +1,10 @@
+* text=auto eol=lf
+*.bat eol=crlf
+
 *.v linguist-language=V text=auto eol=lf
 *.vv linguist-language=V text=auto eol=lf
 *.vsh linguist-language=V text=auto eol=lf
 **/v.mod linguist-language=V text=auto eol=lf
-*.bat text=auto eol=crlf
- Dockerfile.* linguist-language=Dockerfile
-*.toml text eol=lf
```
spytheman, line 7: why is the *.toml entry removed?

ttytm: since `* text=auto eol=lf` which we also add during default scaffolding is added.

spytheman (approved):
Excellent work.

PR #21323, opened 2024-04-20
ci: update deployment workflow

ttytm:
Sorry about the recent oopsi with the deployment workflow. Using the commit message from environment variables that were not added to the github environment in a separate step caused missing information in the commit messages.

With this PR I want to take the chance to fix it properly and also further improve the workflow.
- The changes allow to correctly test vc generation on forks and PRs and deploy on vlang/v master.
- Adds a concurrency setting so interference in deployment will be prevented and ensure sure the latest state will be pushed in such edge cases
- For better traceability it prints the log after creating the commit and rebasing

Example run: https://github.com/ttytm/v/actions/runs/8767565529/job/24060957263

The same applies for module doc generating workflow, though the last list point is excluded here since it's not necessary.


ttytm:
This still can be improved. Sometimes it's hard to coordinate changes step by step. 
Would make a refinement part of one of the follow-ups there would be one where it fits well.

PR #21326, opened 2024-04-21
tools: update `test-self` to allow to choose running tests for cmd and vlib

ttytm:
With the changes, it will be possible to choose to run `v test-self` for project parts in `cmd` and `vlib`. By default - as it already is - both will be included, the same when passing both flags `v test-self --vlib --cmd`.

This can also help with local development but is mainly motivated to further improve our current CI - running parts when and where required.

As far as I can see, when changes are made just to a cmd tool, it **should usually not** require to run tests for the whole vlib. While changes to the vlib **should** include running tests for cmd tools. On the CI part, this can skyrocket tool development and reduce stress on vlib development. As PRs including updates just made to a tool without changes on the vlib can skip the most heavy part of CI.

Workflow update suggestions would happen in a follow-up.

### DIFF
--- a/cmd/tools/vtest-self.v
+++ b/cmd/tools/vtest-self.v
@@ -341,11 +341,9 @@ fn main() {
 	vroot := os.dir(vexe)
 	os.chdir(vroot) or { panic(err) }
 	args_idx := os.args.index('test-self')
-	cmd_prefix := os.args[1..args_idx]
+	cmd_prefix := os.args[1..args_idx].join(' ')
 	mut has_vlib_opt := false
 	mut has_cmd_opt := false
-	mut asan_compiler := false
-	mut msan_compiler := false
 	for arg in os.args#[args_idx + 1..] {
 		match arg {
 			'--vlib' {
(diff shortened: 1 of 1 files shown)


spytheman:
`v test cmd/` is almost the same as `v test-self --cmd` , except for the builtin skipping that `v test-self` does in certain conditions.


What do you think of modifying `v test-self` so that it supports passing *folders*, instead of flags?

Then `v test-self --vlib` could become `v test-self vlib/` and `v test-self --cmd` can be `v test-self cmd/`.
i.e. passing a folder (or several), can just modify the existing `tsession.files` lists, filtering only what is in the passed folders, so that `tsession.test()` will only process what is left.

spytheman (approved):

ttytm:
> `v test cmd/` is almost the same as `v test-self --cmd` , except for the builtin skipping that `v test-self` does in certain conditions.
> 
> What do you think of modifying `v test-self` so that it supports passing _folders_ (relative to vroot), instead of flags (since each flag needs special edge case handling in the source)?
> 
> Then `v test-self --vlib` could become `v test-self vlib/` and `v test-self --cmd` can be `v test-self cmd/`. i.e. passing a folder (or several), can just modify the existing `tsession.files` lists, filtering only what is in the passed folders, so that `tsession.test()` will only process what is left.
> 
> It will be also a bit more flexible, since in some cases, you may want to do: `v test-self vlib/db/` or `v test-self vlib/context/`, which will run all tests, for which v test-self does not have explicit skips, similar to `v test folder/` .

Awesome idea. Thank you for the input!

spytheman (approved):
Excellent work :partying_face: :rocket: !

PR #21327, opened 2024-04-21
tools: rewrite test-self arg handling

ttytm:
Approaches to handle args more safe and maintainable.

After: #21326. Alternatively, it can supersede the PR

### DIFF
--- a/cmd/tools/vtest-self.v
+++ b/cmd/tools/vtest-self.v
@@ -351,7 +351,7 @@ const skip_on_sandboxed_packaging = [
 	'vlib/v/gen/c/coutput_test.v',
 ]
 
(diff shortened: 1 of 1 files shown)


ttytm:
Adding `compiler_errors_test.v` to the skip list when testing for santized undefined as it can't work in that test setting.

To verify:
```sh
./v -cc clang -cflags -fsanitize=undefined -o v2 cmd/v && ./v2 -cc clang -cflags -fsanitize=undefined vlib/v/compiler_errors_test.v
```

In CI there just is no test setting that tests for full sanitized runs, that's why it likely is missing until now.

spytheman (approved):

PR #21330, opened 2024-04-21
tools: skip file that is not compatible with sanitized=undefined tests during v test-self

ttytm:
```sh
./v -cc clang -cflags -fsanitize=undefined -o v2 cmd/v && ./v2 -cc clang -cflags -fsanitize=undefined vlib/v/compiler_errors_test.v
```


spytheman:
Why is this skip needed?

Afaik it works, and can reveal real problems - most of the .vv files, are tested by `compiler_errors_test.v` .

spytheman:
https://github.com/vlang/v/pull/21331 should fix the root cause (cgen for .sort(), called on an empty array).

ttytm:
Yep that's the better solution.

PR #21331, opened 2024-04-21
cgen: fix `-fsanitize=undefined` used with `[]int{}.sort()`

spytheman:
This makes `./v -cc clang -cflags -fsanitize=undefined -o v2 cmd/v && ./v2 -cc clang -cflags -fsanitize=undefined vlib/v/compiler_errors_test.v` pass.


ttytm:
nice :+1:

PR #21339, opened 2024-04-23
test, ci: fix network tests; update and activate tests in periodic ci

ttytm:
The PR fixes network tests and reactivates them in CI
- updates the sslcon shutdown test so that the scenario spawns a test function thread without reading input, since it will currently block finishing the test.
  ```
  v -d network vlib/net/mbedtls/mbedtls_sslconn_shutdown_does_not_panic_test.v
  ```
- adds a currently broken ftp test to the skip files of test-self
  ```
  v -d network vlib/net/ftp/ftp_test.v
  ```
- skips a net.unix test on windows
- updates periodic CI and reactivates tests. On current master it only builds, tests were commented out in https://github.com/vlang/v/commit/e2822356c2875e468ad541961ee87ae78cc8c607
  (regarding macos runners: I'm making the experience that macos-13 runner availability is usually better in comparison to the new m1 runners. Tests often complete before a m1 runner is available. So a macos-13 atm could be the better overall choice when just choosing for performance.)

### DIFF
--- a/.github/workflows/periodic_ci.yml
+++ b/.github/workflows/periodic_ci.yml
@@ -10,9 +10,6 @@ on:
     paths:
       - '**/periodic_ci.yml'
 
-env:
-  V_CI_PERIODIC: 1
-
 jobs:
   network:
     strategy:
@@ -28,14 +25,15 @@ jobs:
     runs-on: ${{ matrix.os }}
     if: github.event_name != 'push' || (github.ref == 'refs/heads/master' && github.repository == 'vlang/v')
     env:
-      VFLAGS: -cc ${{ matrix.cc }} -d network -showcc
+      VFLAGS: -cc ${{ matrix.cc }}
+      V_CI_PERIODIC: 1
     steps:
       - uses: actions/checkout@v4
       - name: Build
         if: runner.os != 'Windows'
-        run: make -j4 && ./v doctor
+        run: make -j4 && ./v -showcc -o v cmd/v && ./v doctor
       - name: Build V (Windows)
         if: runner.os == 'Windows'
-        run: ./make.bat -msvc && ./v doctor
+        run: ./make.bat -msvc && ./v -showcc -o v cmd/v && ./v doctor
       - name: Test
-        run: ./v test-self vlib/net
+        run: ./v -d network test-self vlib/net
--- a/.github/workflows/periodic_ci.yml
+++ b/.github/workflows/periodic_ci.yml
@@ -24,6 +24,7 @@ jobs:
       fail-fast: false
     runs-on: ${{ matrix.os }}
     if: github.event_name != 'schedule' || (github.ref == 'refs/heads/master' && github.repository == 'vlang/v')
+    timeout-minutes: 30
     env:
       VFLAGS: -cc ${{ matrix.cc }}
       V_CI_PERIODIC: 1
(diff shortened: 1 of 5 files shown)


.github/workflows/periodic_ci.yml
```diff
@@ -1,58 +1,41 @@
...
+            cc: msvc
+          - os: macos-13
+            cc: clang
+      fail-fast: false
+    runs-on: ${{ matrix.os }}
+    if: github.event_name != 'push' || (github.ref == 'refs/heads/master' && github.repository == 'vlang/v')
     env:
-      V_CI_PERIODIC: 1
-      VFLAGS: -cc msvc
+      VFLAGS: -cc ${{ matrix.cc }} -d network -showcc
```
spytheman, line 31: I prefer that `-d network` is passed explicitly only when needed like `v -d network test .`, `v -d network test-self`, and not in VFLAGS.

The reason, is that way it is faster and more intuitive to reproduce the CI failures locally, by just copy/pasting the command from the CI logs, without having to construct it from several different places, or copy several lines.

It is also easier to search where `-d network` is used.

A similar argument applies to `-showcc`. Setting it in VFLAGS, makes it apply to the output of `make` and other unrelated v commands, where it just adds noise 99% of the time.

vlib/net/unix/use_net_and_net_unix_together_test.v
```diff
@@ -8,6 +8,9 @@ const tfolder = os.join_path(os.vtmp_dir(), 'net_and_unix_together')
 const test_port = os.join_path(tfolder, 'unix_domain_socket')
 
 fn testsuite_begin() {
+	$if windows {
+		exit(0)
+	}
```
spytheman, line 13: why? it currently passes on windows:
изображение

spytheman: It passed with `-d network` as well 🤔 :
изображение

spytheman:
> macos-13 runner availability is usually better in comparison to the new m1 runners

Good observation. Lets use the macos-13 runner then, when m1 is not needed.

vlib/net/unix/use_net_and_net_unix_together_test.v
ttytm: Oh, I didn't checked the CI error properly enough then. Looks like it's not a Windows but a Windows CI issue. Should not be to hard to resolve at first glance:

```
r.exit_code: 1 ; trimmed_output.len: 619
V panic: net: resolve_addrs Unix socket address is too long; could not resolve path C:\Users\runneradmin\AppData\Local\Temp\v_0\tsession_13dc_7792263700\net_and_unix_together\unix_domain_socket
```

https://github.com/ttytm/v/actions/runs/8806677175/job/24172021083

vlib/net/http/http_httpbin_test.v
```diff
@@ -1,3 +1,4 @@
+// vtest retry: 3
```
ttytm, line 1: I think a 9 would be okay here and likely what I should have put in the first place. In the test, we send requests to httpbin.org, which can be quite overused. I would not put `flaky: true` since regressions might go unnoticed.

To not re-run the whole thing I would wait for the current CI and update just this value after.

spytheman: I agree completely.

vlib/net/unix/use_net_and_net_unix_together_test.v
spytheman: You are right, there is a path length limit, that is rather low:
https://unix.stackexchange.com/questions/367008/why-is-socket-path-length-limited-to-a-hundred-chars

spytheman (approved):
Excellent work.

PR #21342, opened 2024-04-24
ci: update detection of workflow cancellation scenarios

ttytm:
The change should improve and extend our current concurrency and cancellation management. In doing this, it should also allow for simplifications.

The main change is about using concurrency groups that are more tailored to relevant scenarios:

- It will make adding conditions to every job obsolete, which are currently required to control workflow runs for forks and to prevent CI pollution on the original repository.
- It will make it more pleasant to work on branches on forks before opening PRs. Currently, pushing changes to the same branch on a fork will not cancel running and waiting workflows, unlike in a PR. Therefore, it's easy to run into blocked CI that would require long waiting times or manual cancellation to free the CI pipeline. Now, making the cancellation policy on a fork branch the same as for a PR should be the most helpful CI behavior.

### DIFF
--- a/.github/workflows/bootstrapping_works_ci.yml
+++ b/.github/workflows/bootstrapping_works_ci.yml
@@ -10,9 +10,12 @@ on:
     paths-ignore:
       - "**.md"
 
+concurrency:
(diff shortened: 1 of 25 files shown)


spytheman (approved):
Excellent work.

PR #21344, opened 2024-04-24
ci: add tools ci

ttytm:
With the PR, CI is intended to run as follows:

- Updates only made to cmd/tools -> run tool workflows (takes about 3 minutes), don't run regular vlib workflows
  (exception: changes to builder tools will also trigger regular workflows)
- Changes including vlib -> run regular platform workflows (now focusing on vlib) + tool workflows

Other notable changes:
- For most situations, there should also be faster workflow runs for the current platform/sanitized CI, since the tools workflow is run in parallel and tool tests are extracted from regular platform workflows.
 (there will be situations where the benefit of the parallel run will be relativized due to limited runner availability, e.g., when there are many open PRs / active workflow runs. However, there is more potential to optimize our workflows for better runner availability to maximize the occasions of the benefit - I would utilize it with further evolution of the workflows.)
- Tool tests are extended to always include sanitized test runs and to test optimization on all platforms.

### DIFF
--- a/.github/workflows/linux_ci.yml
+++ b/.github/workflows/linux_ci.yml
@@ -6,11 +6,16 @@ on:
       - '**.md'
       - '**.yml'
       - '!**/linux_ci.yml'
+      - 'cmd/tools'
+      - '!cmd/tools/builders/**.v'
   pull_request:
     paths-ignore:
       - '**.md'
       - '**.yml'
       - '!**/linux_ci.yml'
+      - 'cmd/tools'
+      - '!cmd/tools/builders/**.v'
+
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.sha || github.ref }}
@@ -64,9 +69,9 @@ jobs:
       - name: Test pure V math module
         run: ./v -exclude @vlib/math/*.c.v test vlib/math/
       - name: Self tests
-        run: ./v test-self
+        run: ./v test-self vlib
       # - name: Self tests (-cstrict)
-      #   run: V_CI_CSTRICT=1 ./v -cstrict test-self
+      #   run: V_CI_CSTRICT=1 ./v -cstrict test-self vlib
       - name: Build examples
         run: ./v -W build-examples
       - name: Run the submodule example, using a relative path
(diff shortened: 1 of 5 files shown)


.github/workflows/linux_ci.yml
```diff
@@ -135,7 +140,7 @@ jobs:
       - name: All code is formatted
         run: ./v test-cleancode
       - name: Self tests with `-gc boehm` with V compiler using Boehm-GC itself
-        run: ./v -gc boehm test-self
+        run: ./v -gc boehm test-self vlib
```
spytheman, line 143: imho that step can be removed completely, since `-gc boehm` is the default now

.github/workflows/tools_ci.yml
```diff
@@ -0,0 +1,101 @@
...
+        run: |
+          if [[ ${{ matrix.cc }} == "clang" ]]; then
+            cmd="./v -cc clang -cflags -fsanitize=undefined test-self cmd/tools"
+            echo $cmd && $cmd
+            cmd="./v -cc clang -cflags -fsanitize=memory test-self cmd/tools"
+            echo $cmd && $cmd
+          else
+            cmd="./v -cc gcc -cflags -fsanitize=thread test-self cmd/tools"
+            echo $cmd && $cmd
+            cmd="./v -cc gcc -cflags -fsanitize=address,leak,undefined,shift,shift-exponent,shift-base,integer-divide-by-zero,unreachable,vla-bound,null,return,signed-integer-overflow,bounds,bounds-strict,alignment,object-size,float-divide-by-zero,float-cast-overflow,nonnull-attribute,returns-nonnull-attribute,bool,enum,vptr test-self cmd/tools"
```
spytheman, line 57: Are all these compatible with each other?

.github/workflows/windows_ci.yml
```diff
@@ -99,9 +103,7 @@ jobs:
       - name: Test pure V math module
         run: ./v -exclude @vlib/math/*.c.v test vlib/math/
       - name: Self tests
-        run: |
-          ./v -cg cmd\tools\vtest-self.v
-          ./v test-self
+        run: ./v -cg cmd\tools\vtest-self.v && ./v test-self vlib
```
spytheman, line 106: Are you sure that && works in the default windows shell for that runner?
Imho isolate/avoid such changes in PRs that do other things.

ttytm: Yes it works, but can keep such changes into PR that does a general cleanup of workflows.

ttytm:
@spytheman I currently added test runs with `-cstrict` and `-prod`.  Is it enough to just test compilation of the tools with those instead of running all tests with them?

spytheman:
> @spytheman I currently added test runs with `-cstrict` and `-prod`. Is it enough to just test compilation of the tools with those instead of running all tests with them?

Imho just do `v -W -cstrict test-self cmd` for the tools, without a separate `-prod` step (which is very slow, and usually will not reveal problems in the tools themselves without running them).

ttytm:
> Imho just do `v -W -cstrict test-self cmd` for the tools, without a separate `-prod` step (which is very slow, and usually will not reveal problems in the tools themselves without running them).

Thanks for the opinion. I agree. Updated accordingly. It gives use really quick tool tests now.

PR #21345, opened 2024-04-24
vlib: fix veb

ttytm:
- commit no.1 should include the fix and extends renaming
- commit no.2 copies to module again to bring it up to date, since the current uses an older state of x.vweb


medvednikov:
Thanks! I tried it on a fork, but was not successful.

PR #21346, opened 2024-04-25
test, ci: skip same files for `veb` on ubuntu musl that are skipped for `x.vweb`

ttytm:
should fix currently failing ubuntu musl ci


spytheman:
Sorry, did not see this PR this morning.
I've already added it in fe9be96 .

ttytm:
No problem. Even better since one file less is skipped

PR #21347, opened 2024-04-25
tools: fix compiling vwhere with `-cc gcc -cstrict`

ttytm:
fixes:
```
v -cc gcc -cstrict cmd/tools/vwhere
```

### DIFF
--- a/cmd/tools/vwhere/finder_utils.v
+++ b/cmd/tools/vwhere/finder_utils.v
@@ -107,7 +107,7 @@ fn invalid_option(invalid ParamOption, arg string) {
 
 fn valid_args_quantity_or_show_help(args []string) {
 	if true in [
-		args.len < 1,
+		(args.len < 1),
 		'-help' in args,
 		'--help' in args,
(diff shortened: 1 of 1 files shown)


cmd/tools/vwhere/finder_utils.v
```diff
@@ -107,7 +107,7 @@ fn invalid_option(invalid ParamOption, arg string) {
 
 fn valid_args_quantity_or_show_help(args []string) {
 	if true in [
-		args.len < 1,
+		(args.len < 1),
```
spytheman, line 110: Imho that is a general problem with cgen imho:
```
/tmp/v_1000/vwhere.01HW9N7XJ6BZ847VS6F5VJX8A1.tmp.c:22575:23: error: suggest parentheses around comparison in operand of ‘==’ [-Werror=parentheses]
22575 |  if (true == args.len < 1 || true == (Array_string_co...
      |              ~~~~~~~~~^~~
```

Good find.

PR #21351, opened 2024-04-25
cgen: fix hot code reloading on windows

### LINKED ISSUE #21020: Unhandled exception when hot reloading with @[live] attribute on Windows (Invalid Memory Access)
### Describe the bug

using the `@[live]` attribute on a function and compiling with `v -live run .` results in an exception stating `Unhandled Exception 0xC0000005 print_backtrace_skipping_top_frames is not implemented signal 11: segmentation fault print_backtrace_skipping_top_frames is not implemented` 

### Reproduction Steps

```v
module main

@[live]
fn testing_method()
{
	println('testing here!')
}

fn main() 
{
	for 
	{
		testing_method()
	}
}

 ```
 then run `v -live run main.v`
 
 Alternatively, using `v -live run message.v` in v examples

### Expected Behavior

after running the command `v -live run .` or `v -live -cc clang run .`, it should compile and allow modifications to that method at runtime

### Current Behavior

at the moment, it fails to compile altogether, leaving the message `RUNTIME ERROR: invalid memory access`

### Possible Solution

Although I have not tested this myself, it does seem to work for those using Linux or MacOS, I have been told that tcc for Windows is relatively outdated and may need some updating. Also, since the error I am seeing is `print_backtrace_skipping_top_frames is not implemented` and `0xC0000005` it makes me believe it is trying to access memory that has either not been allocated or has already been freed? of course, I am not an expert on this though. 

_No response_

### V version

V 0.4.4 8cb7e75

Windows 10 Enterprise Version 22H2 (OS Build 19045.2728)
> [!NOTE]
>

### PR DESCRIPTION
**Pull request by xenoken**
Fixes #21020

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -352,8 +352,8 @@ fn (mut g Gen) gen_fn_decl(node &ast.FnDecl, skip bool) {
 		}
 		if is_liveshared {
 			if g.pref.os == .windows {
-				g.definitions.write_string('__declspec(dllexport) ${type_name} ${impl_fn_name}(')
-				g.write('__declspec(dllexport) ${type_name} ${impl_fn_name}(')
+				g.definitions.write_string('VV_EXPORTED_SYMBOL ${type_name} ${impl_fn_name}(')
+				g.write('VV_EXPORTED_SYMBOL ${type_name} ${impl_fn_name}(')
 			} else {
 				g.definitions.write_string('${type_name} ${impl_fn_name}(')
 				g.write('${type_name} ${impl_fn_name}(')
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: vlib/v/gen/c/fn.v [hunk 1]
```diff
@@ -351,8 +351,13 @@ fn (mut g Gen) gen_fn_decl(node &ast.FnDecl, skip bool) {
 			g.write('${type_name} no_impl_${name}(')
 		}
 		if is_liveshared {
-			g.definitions.write_string('${type_name} ${impl_fn_name}(')
-			g.write('${type_name} ${impl_fn_name}(')
+			if g.pref.os == .windows {
+				g.definitions.write_string('__declspec(dllexport) ${type_name} ${impl_fn_name}(')
+				g.write('__declspec(dllexport) ${type_name} ${impl_fn_name}(')
```
spytheman, line 356:
```suggestion
				g.definitions.write_string('VV_EXPORTED_SYMBOL ${type_name} ${impl_fn_name}(')
				g.write('VV_EXPORTED_SYMBOL ${type_name} ${impl_fn_name}(')
```

spytheman (approved):
Excellent work.

PR #21355, opened 2024-04-25
ci: update bootstrapping workflow, reactivate `v up` testing, cleanup

ttytm:

### DIFF
--- a/.github/workflows/bootstrapping_ci.yml
+++ b/.github/workflows/bootstrapping_ci.yml
@@ -1,4 +1,4 @@
-name: Bootstrapping works
+name: Bootstrapping CI
 
 on:
   schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/bootstrapping_ci.yml
```diff
@@ -1,24 +1,28 @@
...
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.sha || github.ref }}
   cancel-in-progress: true
 
 jobs:
-  bootstrapping-works:
+  bootstrap-v:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-14]
```
spytheman, line 21: macos-14 here is on purpose, to get the CI to use the new m1 runners

.github/workflows/bootstrapping_ci.yml
```diff
@@ -29,10 +33,11 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 10
-      - name: Build v
-        run: make
-      - name: v.c can be compiled and run with -os cross (bootstrapping works)
+      - name: Build V
+        run: make -j4
+      - name: Test bootstrapping (self-hosting compiler)
```
spytheman, line 38: The previous task name was more descriptive. Please restore it.

.github/workflows/bootstrapping_ci.yml
```diff
@@ -29,10 +33,11 @@ jobs:
...
         run: |
+          ## Ensure v.c can be compiled and run with -os cross.
```
spytheman, line 40: There is no need for this comment, if the task name is restored.

.github/workflows/bootstrapping_ci.yml
ttytm: I've updated the suggestion to combine them. Basically putting that shorter general concept description to the front.

Imho it could help making that mental assignment when thinking/talking about what V's bootstrapping actually does.

Please leave a note fully restoring it is preferred! The CI is quick here.

spytheman (approved):

PR #21360, opened 2024-04-27
tools: utilize environment specific files for vsymlink 

ttytm:
~~To keep the review simpler this PR does an update of the current state prior to an extraction. Also, this can be good to have in the history.~~

~~The followup would be a~~ structural update that allows for better organization and maintainability:

```
├── tools
│   ├── vsymlink
│   │   ├── vsymlink.v
│   │   ├── vsymlink_nix.c.v
│   │   └── vsymlink_windows.c.v
│   ├── ...

```

edit: Splitting the changes runs into some issues. To save some time, I'd directly include the extraction. The conceptual idea of the beforehand simplification would remain part of this PRs history.

### DIFF
--- a/.github/workflows/check_symlink_works.yml
+++ b/.github/workflows/check_symlink_works.yml
@@ -4,11 +4,11 @@ on:
   workflow_dispatch:
   push:
     paths:
-      - 'cmd/tools/vsymlink.v'
+      - 'cmd/tools/vsymlink/**.v'
       - '.github/workflows/check_symlink_works.yml'
   pull_request:
     paths:
-      - 'cmd/tools/vsymlink.v'
+      - 'cmd/tools/vsymlink/**.v'
       - '.github/workflows/check_symlink_works.yml'
 
 jobs:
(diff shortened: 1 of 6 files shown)


spytheman:
The tool is 210 lines on master, and is not going to change much in functionality. Only bug fixes for it will happen.

I do not see the value of splitting it into several files 🤔 .

ttytm:
> The tool is 210 lines on master, and is not going to change much in functionality. Only bug fixes for it will happen.
> 
> I do not see the value of splitting it into several files 🤔 .

It would make a direct distinction and the use of all the `$if windows` in the current file obsolete. Alternatively if it's prefered to keep it in one file I'd just removing the antipatterns of wrapping whole functions in if statements and reduce unnecessary nesting.

spytheman:
All true, and yet, I feel a bit uneasy about splitting a 210 line file... Anyway that is a personal feeling, I'll get over it.

Good work.

ttytm:
Thanks

PR #21363, opened 2024-04-27
parser: add check for result type on chan init

### LINKED ISSUE #21337: "c := chan !string{}"     ->     error: unexpected token `!`, expecting `{`
### Describe the bug

The expression is not compiled:

`c := chan !string{}`

Returns an error:

```
error: unexpected token `!`, expecting `{`
    1 | 
    2 | fn main() {
    3 |     c := chan !string{}
      |               ^

```

### Reproduction Steps

``` v
fn main() {
	c := chan !string{}
	println(c.len)
}
```

### Expected Behavior

a working program

### Current Behavior

```
v run test.v
test.v:3:12: error: unexpected token `!`, expecting `{`
    1 | 
    2 | fn main() {
    3 |     c := chan !string{}
      |               ^
    4 |     println(c.len)
    5 | }

```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.5 f124d03

OS: linux, "Manjaro Linux"
> [!NOTE]
>

felipensp:
Fix #21337

### DIFF
--- a/vlib/v/parser/parser.v
+++ b/vlib/v/parser/parser.v
@@ -2654,6 +2654,9 @@ fn (mut p Parser) name_expr() ast.Expr {
 		first_pos := p.tok.pos()
 		mut last_pos := first_pos
 		mut elem_type_pos := p.peek_tok.pos()
+		if p.peek_tok.kind in [.question, .not] {
+			return p.error_with_pos('cannot use chan with Option or Result type', p.peek_tok.pos())
+		}
 		chan_type := p.parse_chan_type()
 		elem_type_pos = elem_type_pos.extend(p.prev_tok.pos())
 		mut has_cap := false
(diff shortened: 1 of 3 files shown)


vlib/v/parser/parser.v
```diff
@@ -2654,6 +2654,9 @@ fn (mut p Parser) name_expr() ast.Expr {
 		first_pos := p.tok.pos()
 		mut last_pos := first_pos
 		mut elem_type_pos := p.peek_tok.pos()
+		if p.peek_tok.kind in [.question, .not] {
+			return p.error_with_pos('cannot use chan with Option or Result type', p.peek_tok.pos())
```
spytheman, line 2658: I get the restriction for `!Type` values. However I think that `?Type` values should be allowed to be passed in channels. They do have first class support everywhere else.

Delta456: Yes

felipensp: Ok, I've changed the PR to treat result only.

spytheman (approved):
Excellent work.

PR #21364, opened 2024-04-27
ci: update `binary_artifact.yml`

ttytm:
- Adds tiggers to include PRs updating this workflow, while the publishing and releasing part is limited to tag pushes. 
- Simplifes the current state without compromising on functionality where it can be appropriate, makes more consistent.
- For macos-x86 `macos-13` was set instead of `macos-latest` because `macos-latest` is migrating towards arm runners.

The workflow run under this PR should allow to manually inspect the crated binary artifacts.
edit: https://github.com/vlang/v/actions/runs/8860588557?pr=21364

### DIFF
--- a/.github/workflows/binary_artifact.yml
+++ b/.github/workflows/binary_artifact.yml
@@ -139,6 +139,8 @@ jobs:
            rm -rf vc/
            rm -rf v_old.exe
            rm -rf vlib/v/tests/bench/gcboehm/*.svg
+           find |grep pdb$|xargs rm -rf
+           find |grep ilk$|xargs rm -rf
       - name: Create archive
         shell: msys2 {0}
         run: |
--- a/.github/workflows/binary_artifact.yml
+++ b/.github/workflows/binary_artifact.yml
@@ -29,11 +29,11 @@ jobs:
           ./v -skip-unused -cc $CC -prod cmd/tools/vdoctor.v
       - name: Remove excluded
         run: |
-           rm -rf .git/
-           rm -rf thirdparty/tcc/.git/
-           rm -rf vc/
-           rm -rf v_old
-           rm -rf vlib/v/tests/bench/gcboehm/*.svg
+          rm -rf .git/
+          rm -rf thirdparty/tcc/.git/
+          rm -rf vc/
+          rm -rf v_old
+          rm -rf vlib/v/tests/bench/gcboehm/*.svg
       - name: Create ZIP archive
         run: |
           cd ..
@@ -93,11 +93,11 @@ jobs:
           git clone --branch thirdparty-macos-arm64 --depth=1 https://github.com/vlang/tccbin thirdparty/tcc
       - name: Remove excluded
         run: |
-           rm -rf .git/
-           rm -rf thirdparty/tcc/.git/
-           rm -rf vc/
-           rm -rf v_old
-           rm -rf vlib/v/tests/bench/gcboehm/*.svg
+          rm -rf .git/
+          rm -rf thirdparty/tcc/.git/
+          rm -rf vc/
+          rm -rf v_old
+          rm -rf vlib/v/tests/bench/gcboehm/*.svg
       - name: Create ZIP archive
         run: |
           cd ..
@@ -129,11 +129,11 @@ jobs:
       - name: Remove excluded
         shell: msys2 {0}
         run: |
-           rm -rf .git/
-           rm -rf thirdparty/tcc/.git/
-           rm -rf vc/
-           rm -rf v_old
-           rm -rf vlib/v/tests/bench/gcboehm/*.svg
+          rm -rf .git/
+          rm -rf thirdparty/tcc/.git/
+          rm -rf vc/
+          rm -rf v_old.exe
+          rm -rf vlib/v/tests/bench/gcboehm/*.svg
       - name: Create archive
         shell: msys2 {0}
         run: |
--- a/.github/workflows/binary_artifact.yml
+++ b/.github/workflows/binary_artifact.yml
@@ -36,10 +36,10 @@ jobs:
            rm -rf vlib/v/tests/bench/gcboehm/*.svg
       - name: Create ZIP archive
         run: |
-          cd ..
-          zip -r9 --symlinks $ZIPNAME v/
-          mv $ZIPNAME v/
-          cd v/
+           cd ..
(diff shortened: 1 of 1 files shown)


.github/workflows/binary_artifact.yml
```diff
@@ -1,71 +1,67 @@
...
           ./v -skip-unused -cc $CC -prod cmd/tools/vup.v
           ./v -skip-unused -cc $CC -prod cmd/tools/vdoctor.v
       - name: Remove excluded
-        run: |
-           rm -rf .git/
-           rm -rf thirdparty/tcc/.git/
-           rm -rf vc/
-           rm -rf v_old
-           rm -rf vlib/v/tests/bench/gcboehm/*.svg
+        run: rm -rf .git/ thirdparty/tcc/.git vc/ v_old vlib/v/tests/bench/gcboehm/*.svg
```
spytheman, line 31: Restore the previous script. It is deliberately written that way, so that new entries can be added cleanly, the resulting diff is smaller, and the resulting line is not too long.

.github/workflows/binary_artifact.yml
```diff
@@ -1,71 +1,67 @@
...
```
spytheman, line 58: Restore the line by line script. It is easier to maintain, and has smaller diffs.

.github/workflows/binary_artifact.yml
```diff
@@ -78,32 +74,26 @@ jobs:
...
+          rm -rf thirdparty/tcc
+          git clone --branch thirdparty-macos-arm64 --depth=1 https://github.com/vlang/tccbin thirdparty/tcc
       - name: Remove excluded
-        run: |
-           rm -rf .git/
-           rm -rf thirdparty/tcc/.git/
-           rm -rf vc/
-           rm -rf v_old
-           rm -rf vlib/v/tests/bench/gcboehm/*.svg
+        run: rm -rf .git/ thirdparty/tcc/.git vc/ v_old vlib/v/tests/bench/gcboehm/*.svg
```
spytheman, line 90: Restore the previous script.

.github/workflows/binary_artifact.yml
```diff
@@ -128,51 +118,39 @@ jobs:
           .\v.exe -skip-unused -prod -cc msvc cmd\tools\vdoctor.v
       - name: Remove excluded
         shell: msys2 {0}
-        run: |
-           rm -rf .git/
-           rm -rf thirdparty/tcc/.git/
-           rm -rf vc/
-           rm -rf v_old.exe
-           rm -rf vlib/v/tests/bench/gcboehm/*.svg
+        run: rm -rf .git/ thirdparty/tcc/.git vc/ v_old.exe vlib/v/tests/bench/gcboehm/*.svg
```
spytheman, line 121: Restore the previous script.

.github/workflows/binary_artifact.yml
```diff
@@ -128,51 +118,39 @@ jobs:
...
-          msg: ${{ github.ref }}
-          separator: /
       - name: Create Release
         id: create_release
         uses: ncipollo/release-action@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          tag: ${{ steps.split.outputs._2 }}
-          name: ${{ steps.split.outputs._2 }}
+          tag: ${{ github.ref_name }}
```
spytheman, line 161: what is the content of `github.ref_name` ?

ttytm: The tag name. Using `ref_nam`e for release tags for like 15 projects.

ttytm: Or do we create tags for V releases that need some special stripping here?

.github/workflows/binary_artifact.yml
```diff
@@ -132,47 +132,40 @@ jobs:
            rm -rf .git/
            rm -rf thirdparty/tcc/.git/
            rm -rf vc/
-           rm -rf v_old.exe
+           rm -rf v_old
```
spytheman, line 135: I am not sure about that, have you tested in a VM or actual windows, in msys2, that the resulting file is not `v_old.exe` ?

.github/workflows/binary_artifact.yml
ttytm: For the tags that are used in V it should be the same. 
If would be formated like `release_with_slash/<tagname>` than `GITHUB_BASE_REF` would be still the better way to get the first part than using the split action.

spytheman: Is it `weekly.2024.17` for `git tag -s -m "releases: weekly.2024.17" weekly.2024.17` ?

.github/workflows/binary_artifact.yml
ttytm: Good point. Was going to here being in a rush. Sorry.

.github/workflows/binary_artifact.yml
ttytm: In that case split splited the ref to get the ref name.

For a ref `GITHUB_REF` and `GITHUB_REF_NAME` can be checked here:
https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

.github/workflows/binary_artifact.yml
```diff
@@ -41,25 +47,20 @@ jobs:
...
           ./v -skip-unused -cc $CC -prod cmd/tools/vup.v
           ./v -skip-unused -cc $CC -prod cmd/tools/vdoctor.v
       - name: Remove excluded
-        run: |
-           rm -rf .git/
-           rm -rf thirdparty/tcc/.git/
-           rm -rf vc/
-           rm -rf v_old
-           rm -rf vlib/v/tests/bench/gcboehm/*.svg
+        run: rm -rf .git/ thirdparty/tcc/.git vc/ v_old vlib/v/tests/bench/gcboehm/*.svg
```
spytheman, line 63: This rm script still needs restoring.

spytheman (approved):
Excellent work. It is now simpler and cleaner.

spytheman:
Can you please also add:
`find |grep pdb$|xargs rm -rf`
`find |grep ilk$|xargs rm -rf`
to the windows rm script?

image

It will reduce the size of the windows .zip file.

ttytm:
Good observation. Done :+1:

spytheman:
Thank you 🙇🏻‍♂️ .

PR #21365, opened 2024-04-27
wasm,test,ci: prevent false negatives, activate windows in workflow tests

ttytm:
- Currently, when running wasm test without a runtime available, they pass without additional  info about the missing wasm runtime. This makes it easy to miss malfunctions or regressions.

  ```sh
  # pass without info
  ./v test vlib/v/gen/wasm/tests/
  # directly calling a test file would already include the error message and pass.
  ./v vlib/v/gen/wasm/tests/wasm_test.v
  ```
  
  With the updates `exit(1)` instead of `exit(0)` would be used when tests are called with `VTEST_ONLY=wasm` (env variable that is also used in CI). This will give a more appropriate test result when explicitly testing wasm code and prevent false negatives.
  
  ```sh
  # would now error when no runtime is available
  VTEST_ONLY=wasm ./v test vlib/v/gen/wasm/tests/
  ```

- Windows tests are activated as far as they work to prevent regressing further (3 tests where disabled for windows ci runs).

All updates should be easy to follow per commit.

### DIFF
--- a/.github/workflows/wasm_backend_ci.yml
+++ b/.github/workflows/wasm_backend_ci.yml
@@ -42,50 +42,27 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.sha || github.ref }}
   cancel-in-progress: true
 
-env:
-  VTEST_ONLY: wasm
-
 jobs:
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #21367, opened 2024-04-27
tools: update check_os_api_parity.v

ttytm:
The PR does required maintenance work on the check_os_api_parity tool.

### DIFF
--- a/cmd/tools/check_os_api_parity.v
+++ b/cmd/tools/check_os_api_parity.v
@@ -16,6 +16,7 @@ const skip_modules = [
 	'builtin.bare',
 	'builtin.linux_bare.old',
 	'builtin.js',
+	'builtin.wasm',
 	'strconv',
 	'strconv.ftoa',
 	'hash',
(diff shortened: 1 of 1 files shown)


cmd/tools/check_os_api_parity.v
```diff
@@ -99,29 +101,16 @@ fn (app App) gen_api_for_module_in_os(mod_name string, os_name string) string {
...
-
-fn (mut app App) compare_api(api_base string, api_os string, mod_name string, os_base string, os_target string) {
-	res := diff.color_compare_strings(app.diff_cmd, rand.ulid(), api_base, api_os)
-	if res.len > 0 {
-		summary := 'Different APIs found for module: `${mod_name}`, between OS base: `${os_base}` and OS: `${os_target}`'
-		eprintln(term.header(summary, '-'))
-		eprintln(res)
-		eprintln(term.h_divider('-'))
-		app.api_differences[mod_name] = 1
-	}
```
spytheman, line 126: Previously, it checked, and only printed the message, when there were differences.
Please preserve that order, so the output is less noisy.

cmd/tools/check_os_api_parity.v
```diff
@@ -26,43 +27,44 @@ const skip_modules = [
...
 			}
-			api_os := app.gen_api_for_module_in_os(mname, oname)
-			app.compare_api(api_base, api_os, mname, base_os, oname)
+			api_os := gen_api_for_module_in_os(m, oname)
+			if api_base == api_os {
+				continue
+			}
+			diff_modules[m] = true
+			summary := 'Different APIs found for module: `${m}`, between OS base: `${base_os}` and OS: `${api_os}`'
+			eprintln(term.header(summary, '-'))
```
spytheman, line 55: This should only happen after the diff.color_compare_strings call. See the deleted compare_api .

cmd/tools/check_os_api_parity.v
ttytm: Thats why I added before the diff check. 
```
if api_base == api_os {
```
That's how it is done in other palces where we use the diff tool and even cheaper since we don't call the diff fn.

cmd/tools/check_os_api_parity.v
ttytm: Afaik the loop already continued here if there is no diff. So we would get the message that there are differences when there are some. 

Gonna check gain.

cmd/tools/check_os_api_parity.v
spytheman: That is not enough. Checking `if api_base == api_os {` just prevents more work for the non interesting case.

The actual use of this tool, is to discover APIs that *differ* between say linux and windows. In this case, you could have for example `pub fn abc() {` in a nix platform file, and `fn abc() {` in the windows version, or a totally different declaration.
The tool should make a summary of the found changes, and *not print anything*, if there are no differences.

In the new version, the tool will always print `Different APIs found for module:`, even if there are no actual differences found. Those lines are just noise, and not a useful signal.

cmd/tools/check_os_api_parity.v
spytheman: I checked, by running the tool in both branches, and you are right, and I was wrong.

I overlooked that `api_base` and `api_os` are the output that already contains the differences, so checking them for equality is enough 🤦🏻‍♂️ .

spytheman (approved):

ttytm:
Thanks for the reviewing work.

I had a tiny mistake in there, I didn't manged to push:
```diff
- summary := 'Different APIs found for module: `${m}`, between OS base: `${base_os}` and OS: `${api_os}`'
+ summary := 'Different APIs found for module: `${m}`, between OS base: `${base_os}` and OS: `${oname}`'
```

You can update locally and check the output.

PR #21368, opened 2024-04-28
ci: add docker to tools workflow, update ignore paths

ttytm:
improves running tests when they are relevant and skipping them when they are not.

### DIFF
--- a/.github/workflows/module_docs_ci.yml
+++ b/.github/workflows/module_docs_ci.yml
@@ -2,19 +2,13 @@ name: vlib modules CI
 
 on:
   pull_request:
-    paths-ignore:
-      - '**.vv'
-      - '**.out'
-      - '**.yml'
-      - '**.md'
-      - '!**/module_docs_ci.yml'
+    paths:
+      - 'vlib/**.v'
+      - '**/module_docs_ci.yml'
   push:
-    paths-ignore:
-      - '**.vv'
-      - '**.out'
-      - '**.yml'
-      - '**.md'
-      - '!**/module_docs_ci.yml'
+    paths:
+      - 'vlib/**.v'
+      - '**/module_docs_ci.yml'
 
 # Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
 # However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
--- a/.github/workflows/module_docs_ci.yml
+++ b/.github/workflows/module_docs_ci.yml
@@ -4,10 +4,12 @@ on:
   pull_request:
     paths:
       - 'vlib/**.v'
+      - 'vlib/**.md'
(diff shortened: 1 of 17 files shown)


.github/workflows/module_docs_ci.yml
```diff
@@ -2,19 +2,13 @@ name: vlib modules CI
...
   pull_request:
-    paths-ignore:
-      - '**.vv'
-      - '**.out'
-      - '**.yml'
-      - '**.md'
-      - '!**/module_docs_ci.yml'
+    paths:
+      - 'vlib/**.v'
+      - '**/module_docs_ci.yml'
```
spytheman, line 7: Changes to .md files should trigger that too, since they are included in the generation.

.github/workflows/sdl_ci.yml
```diff
@@ -3,15 +3,19 @@ name: sdl CI
...
-    paths-ignore:
-      - "**.md"
-      - "**.yml"
-      - '!**/sdl_ci.yml'
+    paths:
+      - 'vlib/**'
+      - 'thirdparty/**'
+      - 'cmd/tools/builders/**.v'
+      - '**/sdl_ci.yml'
+      - '!**.md'
```
spytheman, line 18: We use `v shader sdl/examples/sdl_opengl_and_sokol` so `cmd/tools/vshader.v` should be added too.

.github/workflows/tools_ci.yml
```diff
@@ -92,3 +92,28 @@ jobs:
...
+      - name: Build V
+        run: make -j4 && ./v -cg -o v cmd/v
+      - name: Verify `v test` works
+        run: |
+          ./v cmd/tools/test_if_v_test_system_works.v
+          ./cmd/tools/test_if_v_test_system_works
+      - name: Test
+        run: ./v test-self cmd
+      - name: Test (-cstrict)
+        run: ./v -W -cstrict test-self cmd
```
spytheman, line 119: Why add that here 🤔 ?
The tools_ci.yml has jobs that are very fast now.
In contrast, the docker job will run very slowly, and check practically everything.

spytheman: Oh, you do `./v test-self cmd`, not `./v test-self` .

spytheman: image
That is nice actually: +2min in parallel to all the rest, and a quick fail if there is a problem.

.github/workflows/module_docs_ci.yml
ttytm: Yep good point. They were ignored before so I didn't add them.

.github/workflows/sdl_ci.yml
ttytm: Thanks!

spytheman (approved):

PR #21370, opened 2024-04-28
builder: use cc enum in CcompilerOptions, fix cc detection, enable cc guessing without prod flag

ttytm:
- The PR utilizes an enum for the CC instead of multiple boolean fields on the CcompilerOptions struct. Since only one CC can be used at a time,  structuring things this way can make sense.

  A nice side effect of the change was discovering a bug where the CC is not set when compiling V on macOS. This was fixed by enabling compiler guessing on macOS even when V is not compiled in prod mode

  Example CI after swapping to use an enum and panicking on an unknown compiler:
  - macos-12: https://github.com/ttytm/v/actions/runs/8870500885/job/24352354898
  - macos-14: https://github.com/ttytm/v/actions/runs/8870500874/job/24352354543

- To pass the sanitized workflow, a vpath in a step (which was enabled in a recently made change) was fixed. Additionally, a trigger was added for changes to this sanitized workflow.

---
Edit:
Refs. for development / review purposes
- Commit no. 3 9b58466 contains the first full CI run after the initial changes from which further improvements were started.

### DIFF
--- a/.github/workflows/sanitized_ci.yml
+++ b/.github/workflows/sanitized_ci.yml
@@ -39,6 +39,7 @@ on:
       - 'vlib/v/markused/**.v'
       - 'vlib/v/preludes/**.v'
       - 'vlib/v/embed_file/**.v'
+      - '**/sanitized_ci.yml'
   pull_request:
     paths:
       - '!**'
@@ -64,6 +65,7 @@ on:
       - 'vlib/v/markused/**.v'
       - 'vlib/v/preludes/**.v'
       - 'vlib/v/embed_file/**.v'
+      - '**/sanitized_ci.yml'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.sha || github.ref }}
@@ -161,7 +163,7 @@ jobs:
           .\make.bat -msvc
           .\v.exe self
       - name: Ensure code is well formatted
-        run: v test-cleancode
+        run: ./v test-cleancode
       # - name: Install dependencies
       #   run: |
       #     .\v.exe setup-freetype
--- a/.github/workflows/sanitized_ci.yml
+++ b/.github/workflows/sanitized_ci.yml
@@ -163,7 +163,7 @@ jobs:
           .\make.bat -msvc
           .\v.exe self
       - name: Ensure code is well formatted
-        run: ./v test-cleancode
+        run: .\v.exe test-cleancode
       # - name: Install dependencies
       #   run: |
       #     .\v.exe setup-freetype
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -186,6 +186,8 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
 				cc_ver.output.contains('clang version ') { .clang }
 				else { panic('failed to detect C compiler from version info `${cc_ver.output}`') }
 			}
+		} else {
+			panic('unknown C compiler')
 		}
 	} else {
 		cc_file_name := os.file_name(ccompiler)
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -178,7 +178,8 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
 	}
 	ccoptions.debug_mode = v.pref.is_debug
 	ccoptions.guessed_compiler = v.pref.ccompiler
-	if ccoptions.guessed_compiler == 'cc' && v.pref.is_prod {
+	guess_compiler := $if macos { true } $else { v.pref.is_prod }
+	if ccoptions.guessed_compiler == 'cc' && guess_compiler {
 		// deliberately guessing only for -prod builds for performance reasons
 		ccversion := os.execute('cc --version')
 		if ccversion.exit_code == 0 {
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -180,20 +180,14 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
 	ccoptions.guessed_compiler = v.pref.ccompiler
 	if ccoptions.guessed_compiler == 'cc' {
 		if cc_ver := os.execute_opt('cc --version') {
-			if gcc_ver := os.execute_opt('gcc --version') {
-				if cc_ver.output == gcc_ver.output {
-					ccoptions.cc = .gcc
-				}
-			} else if clang_ver := os.execute_opt('clang --version') {
-				if cc_ver.output == clang_ver.output {
-					ccoptions.cc = .clang
-				}
-			} else if gpp_ver := os.execute_opt('g++ --version') {
-				if cc_ver.output == gpp_ver.output {
-					ccoptions.cc = .gcc
-				}
+			if cc_ver.output.contains('Free Software Foundation')
+				&& cc_ver.output.contains('This is free software') {
+				// Also covers `g++-9` and `g++-11`.
+				ccoptions.cc = .gcc
+			} else if cc_ver.output.contains('clang version ') {
+				ccoptions.cc = .clang
 			} else {
-				panic('unknown C compiler')
+				panic('failed to detect C compiler from version info `${cc_ver.output}`')
 			}
 		}
 	} else {
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -180,14 +180,11 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
 	ccoptions.guessed_compiler = v.pref.ccompiler
 	if ccoptions.guessed_compiler == 'cc' {
 		if cc_ver := os.execute_opt('cc --version') {
-			if cc_ver.output.contains('Free Software Foundation')
-				&& cc_ver.output.contains('This is free software') {
-				// Also covers `g++-9` and `g++-11`.
-				ccoptions.cc = .gcc
-			} else if cc_ver.output.contains('clang version ') {
-				ccoptions.cc = .clang
-			} else {
-				panic('failed to detect C compiler from version info `${cc_ver.output}`')
+			ccoptions.cc = match true {
+				// Also covers `g++`, `g++-9`, `g++-11` etc.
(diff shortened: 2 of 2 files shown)


ttytm:
Checking the other fails that I couldn't spot before the PR:
If moving towards using an enum, should all supported compilers be part of the enum, or should there be a `other` enum field that is used when the cc is an unknown compiler?

JalonSolov:
`other` makes more sense.  Especially when the `cc` might be a cross-compiler, zig, or anything else that's C99 compatible.

JalonSolov:
For example, you can also choose `-cc g++`, and it usually works fine.

In other words, there is no current way to know _all_ possible compilers.

ttytm:
Thanks for the input.

I let the current CI finish to be able to analyze the full run before making an update.

vlib/v/builder/cc.v
```diff
@@ -178,28 +178,36 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
...
+			} else if clang_ver := os.execute_opt('clang --version') {
+				if cc_ver.output == clang_ver.output {
+					ccoptions.cc = .clang
+				}
+			} else if gpp_ver := os.execute_opt('g++ --version') {
+				if cc_ver.output == gpp_ver.output {
+					ccoptions.cc = .gcc
+				}
+			} else {
+				panic('unknown C compiler')
```
JalonSolov, line 196: Thought you were going to have an `other` type?

vlib/v/builder/cc.v
```diff
@@ -178,28 +178,36 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
...
+	} else {
+		cc_file_name := os.file_name(ccompiler)
+		ccoptions.cc = match true {
+			// vfmt off
+			cc_file_name.contains('tcc') || ccoptions.guessed_compiler == 'tcc' { .tcc }
+			cc_file_name.contains('gcc') || cc_file_name.contains('g++') || ccoptions.guessed_compiler == 'gcc' { .gcc }
+			cc_file_name.contains('clang') || ccoptions.guessed_compiler == 'clang' { .clang }
+			cc_file_name.contains('msvc') || ccoptions.guessed_compiler == 'msvc' { .msvc }
+			cc_file_name.contains('icc') || ccoptions.guessed_compiler == 'icc' { .icc }
+			else { panic('unknown C compiler `${cc_file_name}`') }
```
JalonSolov, line 208: Thought you were going to have an `other` type?

ttytm: That's why the commit message 
> improve info on unknown compiler (for now keep panic for CI, add .unknown to enum later)
I forced pushed, but the message was the same.

It helps to work with the errors for now instead of silencing them.

Though of adding an `unknown` field rather than `other` then as it still can be a `gcc` or `clang` just not detected. `other` implies its something else. Also an error message but not a panic when it's an unknown compiler.

ttytm:
Enabling guessing the compiler outside of prod. Reason is was done only in prod was performance. Not tested yet but the gain through saving the redundant `ccompiler_file_name.contains(...)` checks and due to the improved code structure should relativize performance losses.

vlib/v/builder/cc.v
ttytm: Looks like we can pass now without a panicking if we keep the panic. Keeping it would make it immediately clear when there is an unknown compiler and potentially prevent regressions and unknown behaviour. If keeping the panic is not an option I'd follow through with what was stated in the comment above.

.github/workflows/sanitized_ci.yml
```diff
@@ -161,7 +163,7 @@ jobs:
           .\make.bat -msvc
           .\v.exe self
       - name: Ensure code is well formatted
-        run: v test-cleancode
+        run: ./v test-cleancode
```
spytheman, line 166:
```suggestion
        run: .\v.exe test-cleancode
```

It is already fixed on master.

vlib/v/builder/cc.v
```diff
@@ -175,31 +178,33 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
...
-			}
-			if ccversion.output.contains('clang version ') {
-				ccoptions.guessed_compiler = 'clang'
+	if ccoptions.guessed_compiler == 'cc' {
+		if cc_ver := os.execute_opt('cc --version') {
+			ccoptions.cc = match true {
+				// Also covers `g++`, `g++-9`, `g++-11` etc.
+				cc_ver.output.replace('\n', '').contains('Free Software Foundation, Inc.This is free software;') { .gcc }
+				cc_ver.output.contains('clang version ') { .clang }
+				else { panic('failed to detect C compiler from version info `${cc_ver.output}`') }
```
spytheman, line 187: Do not panic here, that is needlessly restrictive.

vlib/v/builder/cc.v
```diff
@@ -175,31 +178,33 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
...
+	} else {
+		cc_file_name := os.file_name(ccompiler)
+		ccoptions.cc = match true {
+			// vfmt off
+			cc_file_name.contains('tcc') || ccoptions.guessed_compiler == 'tcc' { .tcc }
+			cc_file_name.contains('gcc') || cc_file_name.contains('g++') || ccoptions.guessed_compiler == 'gcc' { .gcc }
+			cc_file_name.contains('clang') || ccoptions.guessed_compiler == 'clang' { .clang }
+			cc_file_name.contains('msvc') || ccoptions.guessed_compiler == 'msvc' { .msvc }
+			cc_file_name.contains('icc') || ccoptions.guessed_compiler == 'icc' { .icc }
+			else { panic('unknown C compiler `${cc_file_name}`') }
```
spytheman, line 201: Remove the panic.

vlib/v/builder/cc.v
```diff
@@ -175,31 +178,33 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
 	}
 	ccoptions.debug_mode = v.pref.is_debug
 	ccoptions.guessed_compiler = v.pref.ccompiler
-	if ccoptions.guessed_compiler == 'cc' && v.pref.is_prod {
-		// deliberately guessing only for -prod builds for performance reasons
-		ccversion := os.execute('cc --version')
-		if ccversion.exit_code == 0 {
```
spytheman, line 181: The is_prod check is there for a reason. Restore it.

spytheman: I do not like how you are doing refactorings (the replacement of the strings and several bool fields -> a single enum is good), mixed with functional changes in behavior (which is questionable) :-| .

Please do keep such PRs separate. One can be merged, and the other closed. Now, you have to do more work to separate them, and I to review it.

vlib/v/builder/cc.v
spytheman: > Also increase the likelihood of making us aware through reports when we can add a not yet supported compiler.

The likelyhood is imho that people experimenting with such compilers will just not report anything and skip V.

The only requirement that we do have, is that the C compiler should support C99. Not that it is named in a particular "supported" way. That is good, and it should stay that way.

vlib/v/builder/cc.v
ttytm: It's a point was kept up for discussion and where an alternative is provided. Can you comment there, if the alternative is appropriate? https://github.com/vlang/v/pull/21370#discussion_r1582498154

vlib/v/builder/cc.v
ttytm: > The is_prod check is there for a reason. Restore it.

- If I'm not forbidden to tust the comment. The reason why it's there is performance.
  But there is no performance regression with the update. Instead there is a benefit of several fixes.

Please read:

- It fixes the compiler being detected. E.g. on macos in general. This was stated in the issue description:
  <details><summary>screenshot macos</summary>
  
  Screenshot_20240429_122657
  </details>
  
  So before opening the PR I just enabled guessing outside prod on macos. After opening it, it revealed that there are scenarios running into the same. I classified them that detecting a compiler of priority - even if there would be a performance regression:
  <details><summary>screenshot freestanding</summary>

  > Screenshot_20240429_122712
  </details>
  
  <details><summary>screenshot vsl/vtl test</summary>
  
  Screenshot_20240429_122751
  </details>
  

   <details><summary>screenshot freebsd</summary>

  Screenshot_20240429_122831
  </details

Another fix that was made, unrelated to enabling guessing without prod:
Screenshot_20240429_122817


So it is mainly a fix, not a refactor.

I know there is a lot of descriptions in the PR, I try to keep them simple and down to what is important to simplify reviewing. But doesn't seem like anyone was read before writing the review comments. The changes pointed out in the reviews were mentioned discussed. I see your point but on some code it's hard sometimes impossible to keep changes very simple or would take a lot of time and PRs so I try to find a middlegroud.

spytheman: All of those are panics, that I do insist should not be there at all.

ttytm: Which I have no problem agreeing up on. Literally zero preference to having them over not having them.

spytheman: The portability of V to new platforms and OSes, will be limited by having a panic for no good reason at all.
The failures that you claim to be fixed, are self inflicted, by the very panics that I do insist should not be there. The check for -prod is also related - without it, the restrictions to `cc` are practically non existent, and that facilitates bootstrapping on new platforms, where it can be anything.

Again, V requires a C99 compiler, not a specific named one, and that should continue to be so. It should not panic when it can not decide what the compiler is, it should just not add compiler specific flags.

ttytm: Yes we don't need to dwell down so much on the panics. I made the experience often enough that it's good to turn something into an error that should be a notice or warning in the result, just to be able to better work with it during development.

I would update the panic like stated here: https://github.com/vlang/v/pull/21370#discussion_r1582498154. Would that make the PR a merge candidate?

spytheman: Even an error is not appropriate imho, since it should not prevent the compilation and bootstrapping on a new system.

At best, it can be a notice.

ttytm: Updated. As - with the PRs changes - we didn't panicked with all our currently tested options, the notice should only become visible when there is an unknown compiler.

spytheman (approved):

PR #21376, opened 2024-04-29
ci: update symlink ci, add matrix

ttytm:
Makes use of a matrix for jobs that share the same steps cross-platform or have marginal differences. This allows for easier and more consistent changes, and should simplify maintainability by reducing redundancy and eliminating the need to update and review the same things in multiple places.

Notable changes:
- While the changes manage to mainly simplify things, they also include an extension in functionality: testing sudo symlinking on macOS.
- Summary messages were removed. Although they are a nice gimmick, they have no real technical benefit. But please don't hesitate to let me know if summary messages should be preserved.

<br>

The changes are part of finalizing on planed rougher-/maintenance-changes to the CI, before moving towards suggestions that are more enhancement-based.

### DIFF
--- a/.github/workflows/check_symlink_works.yml
+++ b/.github/workflows/check_symlink_works.yml
@@ -11,14 +11,23 @@ on:
       - 'cmd/tools/vsymlink/**.v'
       - '.github/workflows/check_symlink_works.yml'
 
+concurrency:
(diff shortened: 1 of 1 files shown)


spytheman:
>     Summary messages were removed. Although they are a nice gimmick, they have no real technical benefit. But please don't hesitate to let me know if summary messages should be preserved.

I agree. I hoped, that they would show in other places (like the `gh` command line monitoring system) but currently they do not offer advantages in that regard.

PR #21378, opened 2024-04-29
ci: update binary artifact workflow, add matrix

ttytm:
Same as with a recent matrix related CI update:
> Makes use of a matrix for jobs that share the same steps cross-platform or have marginal differences. This allows for easier and more consistent changes, and should simplify maintainability by reducing redundancy and eliminating the need to update and review the same things in multiple places.

While the changes manage to mainly simplify things, they also include improvements.

Notable changes concern only Windows release archives:
- data structure that allows for more platform consistency / independency when downloading and viewing the contents of a Windows archive.
  Example screenshots showing directories and archives in comparison on a Linux system:
  - Current - Windows and Linux directories after extraction on the left and archives on the right (windows is flat) 
    Screenshot_20240429_191308

  - Updated - Windows and Linux directories after extraction on the left and windows archive on the right.
    Screenshot_20240429_190102
    
- building is more than twice as fast (secondary for that workflow, but still notable)

Ref. CI runs:
- Run at current state: https://github.com/vlang/v/actions/runs/8876119311
- Test run with the updates: https://github.com/ttytm/v/actions/runs/8882381733 (alternatively the run below the PR)

<br>

It's one of the last planed rougher-/maintenance-changes to the CI, before moving towards suggestions that are more enhancement-based.

### DIFF
--- a/.github/workflows/binary_artifact.yml
+++ b/.github/workflows/binary_artifact.yml
@@ -17,151 +17,90 @@ concurrency:
   cancel-in-progress: ${{ github.ref_name != 'master' }}
 
 jobs:
-  build-linux:
-    runs-on: ubuntu-20.04
-    env:
-      CC: gcc
-      ZIPNAME: v_linux.zip
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-13, macos-14, windows-latest]
+        include:
+          - os: ubuntu-20.04
+            cc: gcc
+          - os: ubuntu-20.04
+            target: linux
+            artifact: v_linux.zip
+          - os: macos-13
+            cc: clang
+          - os: macos-13
+            target: macos_x86_64
+            artifact: v_macos_x86_64.zip
+          - os: macos-14
+            cc: clang
+          - os: macos-14
+            target: macos_arm64
+            artifact: v_macos_arm64.zip
+          - os: macos-14
+            cflags: -cflags "-target arm64-apple-darwin"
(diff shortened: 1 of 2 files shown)


spytheman:
No. I do not want any drastic changes in this CI script.

It works as it is (after fixing it today), and given that it is used once per week, if it breaks again in subtle ways (which is very likely given the changes), we will understand with a huge delay next week.

ttytm:
@spytheman the only change here is related to the artifact creation. 
There are no changes to how things are uploaded or releases created. So as long as the release archives are the same there will be no difference.

Would it be an option to leave it open until next week. Or if I create a release on a fork to show you that it works?

spytheman:
> Would it be an option to leave it open until next week. Or if I create a release on a fork to show you that it works?

You can create a release on a fork, but I do prefer for this PR to stay closed, until next Monday.

spytheman:
I will test it more thoroughly today.

ttytm:
Ref. run and release at latest state after 82b79b2

- https://github.com/ttytm/v/actions/runs/896902311
- https://github.com/ttytm/v/releases/tag/weekly.test.1

ttytm:
What about also renaming it into `Release CI`? This would be the common name for this kind of action.

spytheman:
> What about also renaming it into `Release CI`? This would be the common name for this kind of action.

Yes, that is a good idea. It will simplify searching for it.

spytheman:
Both the name: and the .yml file?

spytheman:
Excellent work.

spytheman (approved):

PR #21380, opened 2024-04-29
builder: simplify generic cc detection

ttytm:
Reduces nesting / loc by handling a potential fatal error directly.

### DIFF
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -180,15 +180,15 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
 	ccoptions.debug_mode = v.pref.is_debug
 	ccoptions.guessed_compiler = v.pref.ccompiler
 	if ccoptions.guessed_compiler == 'cc' {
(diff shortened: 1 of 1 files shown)


vlib/v/builder/cc.v
```diff
@@ -180,21 +180,18 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
...
-			} else if cc_ver.output.contains('clang version ') {
-				ccoptions.cc = .clang
-			} else {
-				if v.pref.is_verbose {
-					eprintln('failed to detect C compiler from version info `${cc_ver.output}`')
-				}
-				eprintln('Compilation with unknown C compiler')
-				ccoptions.cc = .unknown
-			}
+		cc_ver := os.execute_opt('cc --version') or { panic('unknown C compiler') }
```
spytheman, line 183: do not use panics here please

spytheman: Without that panic, if I do not have `cc`:
image

With that panic:
image

I really do prefer the first message.

spytheman (approved):

PR #21387, opened 2024-04-30
v.pref: extract architecture related code into `arch.c.v`; rename `pref.c.v` to `pref.v`

ttytm:
We similarly structure with pref/os. Since the arch code contains the only C code in pref.c.v it can be renamed to pref.v.

### DIFF
--- a/vlib/v/pref/pref.v
+++ b/vlib/v/pref/pref.v
@@ -73,21 +73,6 @@ pub enum CompilerType {
 	cplusplus
 }
 
-pub enum Arch {
-	_auto
-	amd64 // aka x86_64
-	arm64 // 64-bit arm
-	arm32 // 32-bit arm
-	rv64 // 64-bit risc-v
-	rv32 // 32-bit risc-v
-	i386
-	js_node
-	js_browser
-	js_freestanding
-	wasm32
-	_max
-}
-
 pub const list_of_flags_with_param = ['b', 'd', 'e', 'o', 'define', 'backend', 'cc', 'os', 'cflags',
 	'ldflags', 'path', 'arch']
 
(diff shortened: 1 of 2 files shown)


vlib/v/pref/pref.v
```diff
@@ -73,21 +73,6 @@ pub enum CompilerType {
 	cplusplus
 }
 
-pub enum Arch {
-	_auto
-	amd64 // aka x86_64
-	arm64 // 64-bit arm
-	arm32 // 32-bit arm
-	rv64 // 64-bit risc-v
-	rv32 // 32-bit risc-v
```
spytheman, line 82: The comments here made it easier to search for some patterns; please restore them in the extracted enum in arch.c.v too.

vlib/v/pref/pref.v
```diff
@@ -1115,53 +1100,6 @@ pub fn (pref &Preferences) should_output_to_stdout() bool {
...
-		}
-		'rv64', 'riscv64', 'risc-v64', 'riscv', 'risc-v' { // rv64 recommended
-
-			return .rv64
-		}
-		'rv32', 'riscv32' { // rv32 recommended
-
-			return .rv32
-		}
-		'x86_32', 'x32', 'i386', 'IA-32', 'ia-32', 'ia32' { // i386 recommended
```
spytheman, line 1140: The comments are missing in the extracted version. Please restore them.

vlib/v/pref/pref.v
ttytm: Oh okay gonna restore them. I've seen it the Carmack way and wanted to avoid comments where they seem unnecessary as it's also duplicate work when reading code and can make things more confusing.

PR #21389, opened 2024-04-30
pref: cleanup `Backend` and `CompilerType` functions, make minor performance adjustments

ttytm:
The PR does a cleanup and makes minor adjustments. E.g., sorting match arms for more relevant matches first, adding a `gcc` arm instead of using it only in the fallback.

### DIFF
--- a/vlib/v/pref/pref.v
+++ b/vlib/v/pref/pref.v
@@ -1136,8 +1136,8 @@ pub fn cc_from_string(s string) CompilerType {
 		cc.contains('tcc') || cc.contains('tinyc') { .tinyc }
 		cc.contains('gcc') { .gcc }
 		cc.contains('clang') { .clang }
-		cc.contains('mingw') { .mingw }
 		cc.contains('msvc') { .msvc }
+		cc.contains('mingw') { .mingw }
 		cc.contains('++') { .cplusplus }
 		else { .gcc }
 	}
--- a/vlib/v/pref/pref.v
+++ b/vlib/v/pref/pref.v
@@ -1131,7 +1131,7 @@ pub fn cc_from_string(s string) CompilerType {
 	if s == '' {
 		return .gcc
 	}
-	cc := s.all_after_last('\\').all_after_last('/')
+	cc := os.file_name(s).to_lower()
 	return match true {
 		cc.contains('tcc') || cc.contains('tinyc') { .tinyc }
 		cc.contains('gcc') { .gcc }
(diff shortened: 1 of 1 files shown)


vlib/v/pref/pref.v
```diff
@@ -846,13 +846,12 @@ pub fn parse_args_and_show_errors(known_external_commands []string, args []strin
 				b := backend_from_string(sbackend) or {
 					eprintln_exit('Unknown V backend: ${sbackend}\nValid -backend choices are: c, go, interpret, js, js_node, js_browser, js_freestanding, native, wasm')
 				}
-				if b.is_js() {
-					res.output_cross_c = true
-				}
 				if b == .wasm {
```
JalonSolov, line 849: Need a `b.is_wasm()` to align with the other backends.

ttytm: Okay checking, it's correct as it is. For the backends there only is `.wasm`. For the oses there is:
```
		.wasm32,
		.wasm32_emscripten,
		.wasm32_wasi,
		// Native wasm options:
		.wasi,
		.browser,
```

vlib/v/pref/pref.v
```diff
@@ -1114,46 +1113,34 @@ fn is_source_file(path string) bool {
...
-	}
-	if cc.contains('clang') {
-		return .clang
-	}
-	if cc.contains('mingw') {
-		return .mingw
-	}
-	if cc.contains('msvc') {
-		return .msvc
+	cc := s.all_after_last('\\').all_after_last('/')
```
spytheman, line 1134:
```suggestion
	cc := os.file_name(s).to_lower()
```

spytheman: The / and \ searches are from before os.file_name existed afaik.

ttytm: Oh, that's what it looks like. Nice observation!

vlib/v/pref/pref.v
JalonSolov: There are several js output variations as well...
```
Note that `js` defaults to the `node` codegen backend but it's also possible to
pick another:

* `js_browser`        - V outputs JS source code ready for the browser.
* `js_node`           - V outputs JS source code to run with nodejs.
* `js_freestanding`   - V outputs JS source code with no hard runtime dependency.
```
wasm is definitely set up different from the others, if it differentiates via `os` rather than `backend`.

spytheman (approved):

PR #21390, opened 2024-04-30
pref: update `os_from_string`, add missing `qnx`, remove deprecated wasm options that used `-` instead of `_`

ttytm:
Would extract a functional change from other updates I'd like to suggest.

As a change that is quasi related to the scope of the update - can the deprecated options be removed here? I've seen removing deprecated code is on the list, and these have been deprecated for quite some time.

### DIFF
--- a/vlib/v/pref/os.v
+++ b/vlib/v/pref/os.v
@@ -86,6 +86,9 @@ pub fn os_from_string(os_str string) !OS {
 		'serenity' {
 			return .serenity
 		}
+		'qnx' {
+			return .qnx
+		}
 		'plan9' {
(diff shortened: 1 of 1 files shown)


spytheman:
Yes, you can remove the messages for `wasm32-wasi` and `wasm32-emscripten`, they were deprecated in 2022-09-20, so it is safe to assume that they are not used anymore.

JalonSolov:
I think current policy is to allow removal 6 months after the deprecation date.

Should be more than enough time for someone to update their code, if they're paying any attention to it at all.

spytheman (approved):

PR #21391, opened 2024-04-30
v.pref: error for `v file.v --unknown-option`

JoeyShapiro:
## Describe the Problem
Compiling a single code file with flags does not give a warning, and continues the compile.
This will work on real or fake flags
This does not look nice and could cause unknown usage.
```sh
$ ./v examples/hello_world.v -autofree
$ ./v examples/hello_world.v -garbage
$ echo $?
0
```

## Show the Change
This change reports an error and exits as such. It will report the same error if the argument is valid or not, but it is clear what the problem is.
```sh
$ ./v examples/hello_world.v -autofree
Build flags must be before the source file.
$ ./v examples/hello_world.v -garbage
Build flags must be before the source file.
$ echo $?
1
```

### DIFF
--- a/cmd/tools/vdoc/vdoc_file_test.v
+++ b/cmd/tools/vdoc/vdoc_file_test.v
@@ -145,7 +145,6 @@ fn check_output(cmd string, out_path string, opts CheckOutputParams) int {
 	mut fails := 0
 	os.setenv('VDOC_SORT', opts.should_sort.str(), true)
 	expected := os.read_file(out_path) or { panic(err) }.replace('\r\n', '\n').trim_space()
-	dump(cmd)
 	res := os.execute_opt(cmd) or { panic(err) }
 	found := res.output.replace('\r\n', '\n').trim_space()
 	if expected != found {
--- a/cmd/tools/vdoc/vdoc_file_test.v
+++ b/cmd/tools/vdoc/vdoc_file_test.v
@@ -145,6 +145,7 @@ fn check_output(cmd string, out_path string, opts CheckOutputParams) int {
 	mut fails := 0
 	os.setenv('VDOC_SORT', opts.should_sort.str(), true)
 	expected := os.read_file(out_path) or { panic(err) }.replace('\r\n', '\n').trim_space()
+	dump(cmd)
 	res := os.execute_opt(cmd) or { panic(err) }
 	found := res.output.replace('\r\n', '\n').trim_space()
 	if expected != found {
--- a/vlib/v/pref/pref.v
+++ b/vlib/v/pref/pref.v
@@ -948,6 +948,10 @@ pub fn parse_args_and_show_errors(known_external_commands []string, args []strin
 					// Allow for `script.vsh abc 123 -option`, because -option is for the .vsh program, not for v
 					continue
 				}
+				if command == 'doc' {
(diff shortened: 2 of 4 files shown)


spytheman:
I agree that not reporting an error for `./v examples/hello_world.v -garbage` is a problem.
Note however that recognized options do work after the source file currently, and will not after your change, which is a breaking change:
```
#0 09:47:17 ᛋ master /v/oo❱./v examples/hello_world.v -gc none
#0 09:49:15 ᛋ master /v/oo❱ls -l examples/hello_world
-rwxrwxr-x 1 delian delian 502792 May  1 09:49 examples/hello_world
#0 09:49:20 ᛋ master /v/oo❱./v examples/hello_world.v
#0 09:49:23 ᛋ master /v/oo❱ls -l examples/hello_world
-rwxrwxr-x 1 delian delian 699344 May  1 09:49 examples/hello_world
```

spytheman:
Testing on your branch, shows that it behaves differently than your demonstration:
```
#0 09:54:22 ᛋ master /v/oo❱
#0 09:54:22 ᛋ master /v/oo❱git switch -
Switched to branch 'v-file-flags'
#0 09:54:23 ᛋ v-file-flags /v/oo❱./v self
V self compiling ...
#0 09:54:27 ᛋ v-file-flags /v/oo❱./v examples/hello_world.v -autofree
#0 09:54:30 ᛋ v-file-flags /v/oo❱./v examples/hello_world.v -garbage
Build flags must be before the source file.
#1 09:54:34 ᛋ v-file-flags /v/oo❱echo $?
1
#0 09:54:38 ᛋ v-file-flags /v/oo❱
```
... which is nice.

JoeyShapiro:
wow a response, and yeah, it is acting different, my bad. ok give me a bit to work on this, I might have to retitle this issue.

JalonSolov:
Personally, I would like for it to warn about options after the filename.  This is one way `v build` is different from `v run`.  In `v run`, all options after the filename are passed to the target program after it is built.  In `v build`, they're just treated as options to the build.

It can be a bit confusing to new users as to why it is ok to have `v file.v -gc none` but `v run file.v -gc none` doesn't work the same.

JoeyShapiro:
yes I agree. that is why I want this worked out. I realized what I was doing. doing a double dash `--skip-unused` is not the same as `-skip-unused`. This is why in my testing it appeared like they were not working.
So let me reformat my statement.

I want to make some statements, and we can discuss, before I start making any changes
If an option is valid, on either side, it works as intended. however, an invalid option on the right side will not be caught.
I think this examples rounds out all the problems
```sh
$ ./v -o hello-skip.c -skip-unused examples/hello_world.v
$ ./v -o hello-maybe-skip.c examples/hello_world.v -skip-unused
$ ./v -o hello-wont-skip.c examples/hello_world.v --skip-unused
$ ./v -o nothing.c --skip-unused examples/hello_world.v
Unknown argument `--skip-unused`

$ diff -q hello-skip.c hello-maybe-skip.c
$ diff -q hello-skip.c hello-wont-skip.c 
Files hello-skip.c and hello-wont-skip.c differ
```

spytheman:
(rebased over master and resolved conflicts)

spytheman:
(rebased again, since master had an unrelated error, that is now fixed)

PR #21392, opened 2024-04-30
cli: simplify flag parsing

ttytm:
Attempts to create an easier-to-understand base for working with the processes updated in the PR.

### DIFF
--- a/vlib/cli/flag.v
+++ b/vlib/cli/flag.v
@@ -213,31 +213,27 @@ pub fn (flags []Flag) get_strings(name string) ![]string {
 // parse parses flag values from arguments and return
 // an array of arguments with all consumed elements removed.
 pub fn (mut flag Flag) parse(args []string, posix_mode bool) ![]string {
-	if flag.matches(args, posix_mode) {
-		if flag.flag == .bool {
-			new_args := flag.parse_bool(args)!
-			return new_args
-		} else {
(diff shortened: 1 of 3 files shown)


vlib/cli/flag.v
```diff
@@ -213,31 +213,27 @@ pub fn (flags []Flag) get_strings(name string) ![]string {
...
 // matches returns `true` if first arg in `args` matches this flag.
-fn (mut flag Flag) matches(args []string, posix_mode bool) bool {
+fn (mut flag Flag) matches(arg string, posix_mode bool) bool {
 	prefix := if posix_mode { '--' } else { '-' }
-	return (flag.name != '' && args[0] == '${prefix}${flag.name}')
-		|| (flag.name != '' && args[0].starts_with('${prefix}${flag.name}='))
-		|| (flag.abbrev != '' && args[0] == '-${flag.abbrev}')
-		|| (flag.abbrev != '' && args[0].starts_with('-${flag.abbrev}='))
+	return (flag.name != '' && arg.starts_with(prefix + flag.name))
+		|| (flag.abbrev != '' && arg.starts_with('-' + flag.abbrev))
```
spytheman, line 236: What is the value of `arg` at this point?
Is it guaranteed, it would not be for example `--abc-def`, matched against flag.name == `abc` ?

Issue #66, opened 2019-04-05
Redeclaring same variable more than one in same scope


Ivo-Balbaert:
This is rightly forbidden: the following code does indeed that, but should perhaps give a better message signalling that: "Redeclaring same variable more than one in same scope is not allowed"

fn main() {
	a := 42
	a := 50
}

You just found a bug. V can't compile this program, but it should. Please create a GitHub issue.

medvednikov:
Fixed.

PR #21393, opened 2024-04-30
ci: pass a good commit to oldv.v in `bootstrapping_ci.yml`

ttytm:
The changes should be able to address: https://github.com/vlang/v/commit/fd66067bc02dba9fc8198d1fea04c9e3c071de74#commitcomment-141513977

A small change included here is a resolved linter complaint for the empty
```
B_CFLAGS:
```
`B_CFLAGS: ''` Would address it, but as the var is apparently not used it was removed.

### DIFF
--- a/.github/workflows/bootstrapping_ci.yml
+++ b/.github/workflows/bootstrapping_ci.yml
@@ -59,9 +59,8 @@ jobs:
           ls -la v_from_vc2 v_from_vc_produced_native_v2
       - name: Test `v up`
         run: |
-          # Get commit sha from the last successful, fast workflow that could build V on master.
-          # The workflow used below is `Path Testing CI` (18477644).
-          last_good_commit=$(curl -s "https://api.github.com/repos/vlang/v/actions/workflows/18477644/runs?branch=master&status=success&per_page=1" | jq -r '.workflow_runs[0].head_sha')
+          # Derive commit sha from an older successful fast workflow on master that was able to build V.
+          last_good_commit=$(curl -s "https://api.github.com/repos/vlang/v/actions/workflows/18477644/runs?branch=master&status=success&per_page=2" | jq -r '.workflow_runs[1].head_sha')
           echo "last_good_commit=$last_good_commit"
           ./v run cmd/tools/oldv.v -v $last_good_commit
           cd ~/.cache/oldv/v_at_$last_good_commit
--- a/.github/workflows/bootstrapping_ci.yml
+++ b/.github/workflows/bootstrapping_ci.yml
@@ -21,6 +21,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-14]
+      fail-fast: false
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
     env:
@@ -60,7 +61,12 @@ jobs:
       - name: Test `v up`
         run: |
           # Derive commit sha from an older successful fast workflow on master that was able to build V.
-          last_good_commit=$(curl -s "https://api.github.com/repos/vlang/v/actions/workflows/18477644/runs?branch=master&status=success&per_page=2" | jq -r '.workflow_runs[1].head_sha')
+          last_good_commit=$(curl -L \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            "https://api.github.com/repos/vlang/v/actions/workflows/18477644/runs?branch=master&status=success&per_page=2" \
+            | jq -r '.workflow_runs[1].head_sha')
           echo "last_good_commit=$last_good_commit"
           ./v run cmd/tools/oldv.v -v $last_good_commit
           cd ~/.cache/oldv/v_at_$last_good_commit
--- a/.github/workflows/bootstrapping_ci.yml
+++ b/.github/workflows/bootstrapping_ci.yml
@@ -14,7 +14,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.sha || github.ref }}
-  cancel-in-progress: false
+  cancel-in-progress: true
 
 jobs:
   bootstrap-v:
--- a/.github/workflows/bootstrapping_ci.yml
+++ b/.github/workflows/bootstrapping_ci.yml
@@ -25,7 +25,6 @@ jobs:
     timeout-minutes: 30
     env:
       VFLAGS: -no-parallel
-      B_CFLAGS:
       B_LFLAGS: -lm -lpthread
     steps:
       - uses: actions/checkout@v4
@@ -37,7 +36,7 @@ jobs:
         run: |
           ls -la v vc/v.c
           ./v -os cross -o vc/v.c cmd/v
-          cc $B_CFLAGS -o v_from_vc vc/v.c $B_LFLAGS
+          cc -o v_from_vc vc/v.c $B_LFLAGS
           ls -lart v_from_vc
           ./v_from_vc version
           ./v_from_vc run examples/hello_world.v
@@ -48,7 +47,7 @@ jobs:
           ls -la v vc/v.c v_from_vc v_from_vc_produced_native_v
           ./v_from_vc_produced_native_v -os cross -o vc/v.c cmd/v
           ### do it a second time, just in case:
-          clang $B_CFLAGS -o v_from_vc2 vc/v.c $B_LFLAGS
+          clang -o v_from_vc2 vc/v.c $B_LFLAGS
           ls -lart v_from_vc2
           ./v_from_vc2 version
           ./v_from_vc2 run examples/hello_world.v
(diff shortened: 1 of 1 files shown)


.github/workflows/bootstrapping_ci.yml
```diff
@@ -14,18 +14,18 @@ on:
...
   group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.sha || github.ref }}
-  cancel-in-progress: false
+  cancel-in-progress: true
 
 jobs:
   bootstrap-v:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-14]
+      fail-fast: false
```
spytheman, line 24: From <https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures>:
> jobs.<job_id>.strategy.fail-fast applies to the entire matrix. If jobs.<job_id>.strategy.fail-fast is set to true or its expression evaluates to true, GitHub will cancel all in-progress and queued jobs in the matrix if any job in the matrix fails. This property defaults to true.

What happens when the jobs themselves do not fail, but they are in the same concurrency group, and are started at the same time?

See <https://github.com/vlang/v/actions/runs/8890038625/attempts/1> .
Both jobs got canceled.

.github/workflows/bootstrapping_ci.yml
```diff
@@ -14,18 +14,18 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.sha || github.ref }}
-  cancel-in-progress: false
+  cancel-in-progress: true
```
spytheman, line 17: Imho, leave that to `false`.

.github/workflows/bootstrapping_ci.yml
```diff
@@ -60,8 +60,15 @@ jobs:
...
+          # Derive commit sha from an older successful fast workflow on master that was able to build V.
+          last_good_commit=$(curl -L \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            "https://api.github.com/repos/vlang/v/actions/workflows/18477644/runs?branch=master&status=success&per_page=2" \
+            | jq -r '.workflow_runs[1].head_sha')
+          echo "last_good_commit=$last_good_commit"
+          ./v run cmd/tools/oldv.v -v $last_good_commit
+          cd ~/.cache/oldv/v_at_$last_good_commit
```
spytheman, line 72: No, the purpose is not to check that `v up` from a specific commit works, and even if it was, just hardcoding the sha of the known good commit would have been much much simpler, than that curl thing.

The purpose of this task is to test `v up` for recent commits. That way, it will check more things, preferably without needing more updates.

Note, that I do agree that it would be best to test from a good commit, where V could be at least built (to avoid delayed failures like after the recent merge conflict).

However, imho a better tactic for it, to avoid hard coding a known good sha (which will need constant updates to keep it recent), would be to start from say 3 commits ago, then if that does not work, try from 4 commits ago, then 5 etc, up to say 10 commits ago, logging everything.

spytheman: btw, it was not obvious to me, that the `18477644` workflow id is for the `Path Testing CI`, i.e. `paths_ci.yml`.

Is there a convenient way, to see it directly on the github site, without using `gh workflow list` and searching in it, or `gh workflow view 18477644` ?

spytheman: I was wrong - the curl thing does retrieve a known good commit dynamically, just in a more obscure way than I thought initially.
I tested locally with:
```
last_good_commit=$(curl -L \
            -H "Accept: application/vnd.github+json" \
            -H "X-GitHub-Api-Version: 2022-11-28" \
            "https://api.github.com/repos/vlang/v/actions/workflows/18477644/runs?branch=master&status=success&per_page=10" \
            | jq -r '.workflow_runs[5].head_sha') && echo $last_good_commit
```
and it retrieved `00dd0bf9383c601b3b96a57191f6b76335f44538` , which is a recent commit, but not the latest, which is very good.

spytheman: It did not need the `Authorization: Bearer` header.

.github/workflows/bootstrapping_ci.yml
ttytm: Okay I'm happy to be able to address all the concerns. Gonna start here:

- The `fail-fast: false` is only if mac fails it won't cancel the ubuntu matrix, its unrelated to the concurrency group. To get the result for both platforms to better analyze why a failed happened if it did so. `fail-fast: false` is what we use afaik in any other matrix.

.github/workflows/bootstrapping_ci.yml
ttytm: If it would fix something false would correct. Tho false is false here 😅. It should not be used. 

In the review comment above you mention a workflow.
> See [vlang/v/actions/runs/8890038625/attempts/1](https://github.com/vlang/v/actions/runs/8890038625/attempts/1) .
Both jobs got canceled.

The reason for the fail was that it was run scheduled. 
Checking run #66:
Screenshot_20240501_101348
You removed scheduled runs. So this was solved there. If scheduled runs were kept, a update to concurrency group name for the master branch that considers this would be the right solution.  

The main reason why it should be false I mentioned in comment linked in the description:

> When pushing to master the concurrency group uses the sha of the commit, so there will be no cancelation anyway.
-> In other words, on master it is a separate concurrency group without a possible cancellation.
>
> Settings this to always false will only stack up the CI pipline with this workflow when there are repeated pushes to PRs. Then there will waiting for old irrelevant runs of this workflow to finish before the relevant of a PR starts.
Therefore it should stay true.

.github/workflows/bootstrapping_ci.yml
ttytm: Is it going up just 3 `^^^` parents better vs always going up a few "good" parents (can also be 3)?

In the first commit I used a simpler version, which also included a comment about the used workflow.
```yml
# Derive commit sha from an older successful fast workflow on master that was able to build V.
# The workflow used below is `Path Testing CI` (18477644).
last_good_commit=$(curl -s "https://api.github.com/repos/vlang/v/actions/workflows/18477644/runs?branch=master&status=success&per_page=1" | jq -r '.workflow_runs[0].head_sha')
echo "last_good_commit=$last_good_commit"
```

> It did not need the Authorization: Bearer header, which is also nice.
Yes it's possible to check the URL without authorization.

Those api endpoints have fairly low rate limits though. So I tried to give the version without the auth token a few consecutive pushes. It started to fail then, and could be resolved by adding the auth token.

ttytm: - Regarding the variable name: maybe using something better than `last_good_commit` as it's not exactly the `last`. `old_good_commit`? 

- Regarding rate limits: Idk exactly under which ip address runners are tracked. It can be aspects outside this repository why they already come with a lower rate limit.


-  For local testing: https://api.github.com/rate_limit
-  Check again after opening the url to get the good commmit: https://api.github.com/repos/vlang/v/actions/workflows/18477644/runs?branch=master&status=success&per_page=1

Authorized, the rate limit is 5000 instead of 60.

.github/workflows/bootstrapping_ci.yml
spytheman: i.e. you think that the <https://github.com/vlang/v/actions/runs/8890038625/attempts/1> cancelations, were caused by an older periodic/scheduled run?

spytheman: > Checking the linked run 66
I do not know where do you see that.
Can you please link to it?

spytheman: I see this:
image

.github/workflows/bootstrapping_ci.yml
spytheman: > In the first commit I used a simpler version, which also included a comment about the used workflow.

Please restore that comment. In several weeks/months, it would be hard again to re-acquire that knowledge, and it may be needed, in the worst possible time for little side quests.

.github/workflows/bootstrapping_ci.yml
ttytm: > i.e. you think that the [vlang/v/actions/runs/8890038625/attempts/1](https://github.com/vlang/v/actions/runs/8890038625/attempts/1) cancelations, were caused by an older periodic/scheduled run?

Based on the available data I'm confident that is the reason.

> Can you please link to it?

Sure, here a link to the overview of the bootstrapping_ci runs on master:
https://github.com/vlang/v/actions/workflows/bootstrapping_ci.yml?query=branch%3Amaster
When scrolling down a bit you find:
Screenshot_20240501_134752

The overview shows the triggers next to the workflow name. In the summary of a single workflow it's not immediately visible.

.github/workflows/bootstrapping_ci.yml
ttytm: Thanks for the constructive review.

If there are further unclear points lets address them.

spytheman: No, I think it is good to merge as it is now.

spytheman (approved):
Excellent work.

ttytm:
Looks like additional verification if a sha is still present on master is needed. In case of force pushes or why ever they are not available anymore?

What do you think @spytheman

PR #21395, opened 2024-04-30
pref: cleanup conditions in `os.v` and `arch.c.v`

ttytm:


vlib/v/pref/os.v
```diff
@@ -7,9 +7,8 @@ import os
 
 pub enum OS {
 	_auto // Reserved so .macos cannot be misunderstood as auto
-	ios
-	macos
 	linux
+	macos
```
spytheman, line 11: why reorder the `ios` next to android?
imho it is much closer to macos in behavior

spytheman: Reordering macos after linux also does not make much sense, given the comment on line 9.

spytheman: imho just restore the order, the rest of the PR seems very good.

ttytm: Yes gonna revert the sorting. The order was a minor change just regarding consistency. 
I put `ios` - a likely lesser often used platform - next to android. I'm good with keeping it as it is.

spytheman (approved):

spytheman:
I think the CI failure was caused by a bug in the $if implementation, but I am not sure yet.
We would have to do it in steps at a later date (first only the non comptime changes, and then the $if ones).

ttytm:
Yes, after seeing that a revert was necessary I also wanted to suggest to do it step by step to allow to isolate the bug.

So I'd also say that's the way.

spytheman:
I think the main difference is from the generated code here:

https://github.com/vlang/vc/commit/663b4cf403ead57e0285038450dd9420f7e86253#diff-433300c4cce601bb34f081942aecd3dea93e6273442da2d38f6008ead4f43648L41935

image

ttytm:
Hard to tell for me. Best will be to try and check the smaller diffs.

V code that thing might have caused it is.
```v
pub fn res_ret_fn(arch_str string) !Arch {
	return match arch_str {
		'amd64', 'x86_64', 'x64', 'x86' { .amd64 } // amd64 recommended
		// ...
		else { error('invalid arch: ${arch_str}') }
	}
```

Using the else `{ error(... }` instead of `else { return error(... }`. Even when there is a return before the expression `return match ...`
I had some builder errors using it like this locally.

In the PR I had to use `return error` in `os_from_string(os_str string) !OS` else it would have already failed. Maybe `arch_from_string` would require it as well.

When those refactorings have the side effect of disclosing bugs and extending tested cases it's a very good side effect I think.

spytheman:
> When those refactorings have the side effect of disclosing bugs and extending tested cases it's a very good side effect I think.

Yes, and no.
Yes, because the effect is improving the compiler.
No, because doing it like this, in the most critical and less easily testable parts of the compiler, instead of in separate isolated tests, is the most stressful and time consuming for both the you and me.

I would have preferred for bugs of the corresponding language constructs, to be isolated in their own _test.v files and fixed in their own targeted PRs to the checker and cgen, and not to block the whole CI by doing in `v.pref`, where they can have the widest impact.

spytheman:
v.c broke on macos after https://github.com/vlang/v/pull/21396 too :
https://github.com/vlang/v/actions/runs/8913644903/job/24479556291

I've reverted it again in <https://github.com/vlang/v/commit/794e238c7af99803636f88d36e421207a8269be0> .

ttytm:
I mostly agree. 

After all, the way it was not was not like going Rambo and pushing untested or known bugs to master. But to the best of our knowledge, getting changes CI covered.
In regards of efficiency I think it's okay too. Some things are hard or not possible to isolate locally or would take way more time. On top the changes here depend on other repos as well.
So I would still call it good and part of the game, especially for a young project. It will result in a better product eventually.

> v.c broke on macos after https://github.com/vlang/v/pull/21396 too :

That helps to isolate it. Shouldn't be to hard to reproduce it locally now.

PR #21398, opened 2024-05-02
v.ast: extract `ast.File` into `ast/file.v`, add more `ast.EmbeddedFile` code to `ast/embed_file.v`

ttytm:
Currently, `ast/embed_file.v` has minimal content. The PR renames it to `ast/file.v` to then extract `ast.File` code and more `ast.EmbeddedFile` code into it.


spytheman:
Why extract them in a separate file? 

`ast.File` and `ast.EmbeddedFile` are not related at all.

ttytm:
Okay, for simplicity I reopen and do the split here.

spytheman:
I think, that I was not clear - I do not want any split of `ast.v` .

All AST nodes are currently in that file `ast.v` . If I want to modify something related to their declarations, I know where to find it.

Splitting declarations from it into their own files, is completely arbitrary, and in the extreme, will lead to ast.v splitting to ~150 small files (the current number of ast nodes).

-> I do not see what is the point? There is no objective technical or organizational value in the current PR.

ttytm:
Got you.

ttytm:
Regarding the refactorings and as an explanation fits into the context of the opened diff rewrite and I can use it as an example. 
Preparing or getting everything together for such changes takes a lot of additional work.
And this is still a fairly "small" change. When trying to make bigger improvements via PRs it can be multiple the amounts of work sometimes.
What could be done in three days can take weeks or longer. This doesn't mean I want extended access or anything.
It's okay as it is, I even prefer it that way. It results in better quality. Just want remind that it is a totally different dynamic, I'm doing my best to find a good approach. Usually, I the refactorings done are in the context of a fix or feature or when I think that they make a lot of sense.

spytheman:
Yes. You are 100% right, that it is much slower, compared to just making big changes at one go, without splitting into PRs.

I do appreciate your efforts, and your understanding, that it (hopefully) results in better quality very much.

I know, that it can be frustrating to do work, that is rejected, for seemingly not good reason from your point of view, and that is why, I try to provide explanations for closing PRs like the current one. I may not always succeed in that, and if you really do think, that I am wrong in some decisions, just tell me, so that I can re-evaluate, or ping Alex, so that he can do so.

Doing big changes though, especially in modules, that are core to the language implementation like `v.pref`, `builtin`, `os`, `v.ast` etc, imho should be justified by bringing positives, either technical or organizational. The cost/risk for breaks otherwise is too big.

PR #21400, opened 2024-05-02
sync.stdatomic: add paths for compilation with musl on Gentoo

spytheman:


JalonSolov:
Shouldn't those paths be added in reverse order?

What if someone has multiple versions installed - this way, the oldest will be found and used, instead of the newest.

PR #21403, opened 2024-05-02
v.util: rewrite diff module, deprecate old functions

ttytm:
The rewritten module should extend functionality, make usage easier/less verbose, fix issues and resolve friction points that may stand in the way of using it in whatever context.

## Some of the changes:

An overview of the changes is given below. However, it should not be necessary for the review, it was attempted to keep the code changes compact and easy to understand. Feel free to skip the overview.

### General reliability and predictability of return values:

- Use of known diff tools that can actually return a string value.

  To make the function call itself as safe and reliable as possible, tools like `code` and `opendiff` that open non-conforming files in a gui and do return a result string is limited to the overwrite via the environment variable. Additionally, it is possible to disable and to customize overwrites via the env variable.

### Error handling:

- Use of Result return types
  - The "old" module handles errors via unrecoverable panics or returns them formatted as string.
    Both should be resolved through using `!string` return types.

### Other changes:

- Fix  `\ No newline at end of file` in the return values of the string comparing function.
  Which is suboptimal for a function that by definition compares strings.
  ```v
  // Current
  diff.color_compare_strings(diff.find_working_diff_command()!, 'unique_smthn', 'abc', 'cde')
  // New
  diff.compare_text('abc', 'cde')!
  ```
  Comparing the output (assuming the automatically available diff command is `diff`)
  ```sh
  # Current
  ❯ v run file.v
  # ...
  -abc
  \ No newline at end of file
  +cde
  \ No newline at end of file

  # New
  ❯ v run file.v
  # ...
  -abc
  +cde
  ```

- Include [delta](https://github.com/so-fancy/diff-so-fancy) as diff tool. Which is a feature rich tool for human readable diffs that can help to analyze outputs. It can also emulate diff-so-fancy.

- Extend tests.

- For now, `vfmt.v` was updated as the only module that the diff module internally - also to make it easier testing the changes during a review locally. Other that use the current diff module could be updated separately.

### Better handled, more user oriented options:

- Handle options in a non-conflicting way. The current module has some defaults options that are hard coded to be always added, which might conflict with other args or tools.
  - **This is also a fix for the statement in `v fmt --help`**, which shows `VDIFF_OPTIONS="-W 80 -y"` (using the `-y` flag for side-by-side comparison).
    | Current | Updated |
    | - | - |
    | cur_diff_y | Screenshot_20240504_130540 |

    As a reference to verify that it's not a regression 87270e9936f44ee0eccf9da9eb56d6aed297467e is a commit before the recent fixes and preparations of the diff module.
  - Allows to use other tools
    | Current | Updated |
    | - | - |
    | cur_delta | Screenshot_20240504_130445 |

### DIFF
--- a/vlib/v/util/diff/diff.v
+++ b/vlib/v/util/diff/diff.v
@@ -3,7 +3,128 @@ module diff
 import os
 import time
 
+pub enum DiffTool {
+	auto
+	diff // core package on Unix-like systems.
+	colordiff // `diff` wrapper.
+	delta // viewer for git and diff output.
+	// fc // built-in tool on windows. // TODO: enable when its command output can be read.
+	@none
+}
+
+@[params]
+pub struct CompareOptions {
+pub:
+	cmd DiffTool
+	// Custom args used with the diff command.
+	args string
+	// Sets the environment variable whose value can overwrite a diff command passed to a compare function.
+	// It also enables the use of commands that are not in the list of known diff tools.
+	// Set it to 'none' to disable it.
+	env_overwrite_var ?string = 'VDIFF_CMD'
+}
+
+// Default options for `diff` and `colordiff`.
+// Short `diff` args are supported more widely (e.g. on OpenBSD, ref. https://man.openbsd.org/diff.1).
+// `-d -a -U 2` ^= `--minimal --text --unified=2`
+const default_diff_args = $if openbsd || freebsd { '-d -a -U 2' } $else { '-d -a -U 2 -F="fn "' }
+const known_diff_tool_defaults = {
+	// When searching for an automatically available diff tool, the tools are searched in this order.
+	DiffTool.delta: ''
+	.colordiff:     default_diff_args
+	.diff:          default_diff_args
+	// .fc:        '/lnt'
+}
+
+pub const available_tool = find_working_diff_tool() or {
+	// Allows public checking for the available tool and prevents repeated searches
+	// when using compare functions with automatic diff tool detection.
+	dbg('${@LOCATION}: No working comparison command found')
+	DiffTool.@none
+}
+
+// compare_files returns a string displaying the differences between two files.
+pub fn compare_files(path1 string, path2 string, opts CompareOptions) !string {
+	if v := opts.env_overwrite_var {
+		env_cmd := os.getenv(v)
+		if env_cmd != '' {
+			tool, args := env_cmd.split_once(' ') or { env_cmd, opts.args }
+			os.find_abs_path_of_executable(tool) or {
+				return error('error: failed to find comparison command `${tool}`')
+			}
+			return run_tool('${tool} ${args} ${os.quoted_path(os.real_path(path1))} ${os.quoted_path(os.real_path(path2))}',
+				@LOCATION)
+		}
+	}
+	if diff.available_tool == .@none {
+		return error('error: failed to find comparison command')
+	}
+	tool := match true {
+		opts.cmd != .auto { opts.cmd }
+		else { diff.available_tool }
+	}
+	tool_cmd := $if windows { '${tool.str()}.exe' } $else { tool.str() }
+	os.find_abs_path_of_executable(tool_cmd) or {
+		return error('error: failed to find comparison command `${tool_cmd}`')
+	}
+	p1, p2 := os.quoted_path(os.real_path(path1)), os.quoted_path(os.real_path(path2))
+	if opts.args == '' {
+		args := if defaults := diff.known_diff_tool_defaults[tool] { defaults } else { '' }
+		if tool == .diff {
+			// Ensure that the diff command supports the color option.
+			// E.g., some BSD installations do not include `diffutils` alongside `diff` by default.
+			res := run_tool('${tool} ${args} --color=always ${p1} ${p2}', @LOCATION)
+			if !res.contains('unrecognized option') {
+				return res
+			}
+		}
+	}
+	return run_tool('${tool} ${opts.args} ${p1} ${p2}', @LOCATION)
+}
+
+// compare_text returns a string displaying the differences between two strings.
+pub fn compare_text(text1 string, text2 string, opts CompareOptions) !string {
+	ctime := time.sys_mono_now()
+	tmp_dir := os.join_path_single(os.vtmp_dir(), ctime.str())
+	os.mkdir(tmp_dir)!
+	defer {
+		os.rmdir_all(tmp_dir) or {}
+	}
+	path1 := os.join_path_single(tmp_dir, 'text1.txt')
+	path2 := os.join_path_single(tmp_dir, 'text2.txt')
+	// Add `\n` when comparing strings to prevent `\ No newline at end of file` in the output.
+	os.write_file(path1, text1 + '\n')!
+	os.write_file(path2, text2 + '\n')!
+	return compare_files(path1, path2, opts)!
+}
+
+fn find_working_diff_tool() !DiffTool {
+	for tool in diff.known_diff_tool_defaults.keys() {
+		cmd := $if windows { '${tool.str()}.exe' } $else { tool.str() }
+		os.find_abs_path_of_executable(cmd) or { continue }
+		if tool == .delta {
+			// Sanity check that the `delta` executable is actually the diff tool.
+			res := os.execute('${cmd} --help').output.split_into_lines()[0]
+			if !res.contains('diff') {
+				dbg('${@LOCATION} delta does not appear to be the diff tool `${res}`')
+				continue
+			}
+		}
+		return tool
+	}
+	return error('No working "diff" command found')
+}
+
+fn run_tool(cmd string, dbg_location string) string {
+	dbg('${dbg_location}: cmd=`${cmd}`')
+	res := os.execute(cmd)
+	dbg('${dbg_location}: res=`${res}`')
+	return res.output.trim_right('\r\n')
+}
+
 // find_working_diff_command returns the first available command from a list of known diff cli tools.
+@[deprecated_after: '2024-06-30']
+@[deprecated]
 pub fn find_working_diff_command() !string {
 	env_difftool := os.getenv('VDIFF_TOOL')
 	env_diffopts := os.getenv('VDIFF_OPTIONS')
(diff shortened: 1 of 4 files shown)


vlib/v/util/diff/diff.v
```diff
@@ -3,7 +3,128 @@ module diff
...
+	tmp_dir := os.join_path_single(os.vtmp_dir(), ctime.str())
+	os.mkdir(tmp_dir)!
+	defer {
+		os.rmdir_all(tmp_dir) or {}
+	}
+	path1 := os.join_path_single(tmp_dir, 'text1.txt')
+	path2 := os.join_path_single(tmp_dir, 'text2.txt')
+	// Add `\n` when comparing strings to prevent `\ No newline at end of file` in the output.
+	os.write_file(path1, text1 + '\n')!
+	os.write_file(path2, text2 + '\n')!
```
spytheman, line 97: The default should not change the content of what is passed.

The `No newline at end of file` message is a valuable signal.

ttytm: Can you add more info why? This would help me agree.

I would agree immediately if we would compare files. But this is comparing strings and the `No newline at end of file` seems inappropriate when comparing strings.

ttytm: Or let me phrase it like this. I want to make aware that this is comparing strings.
This like would be there for most comparisons made. If you are aware and this is the preferred way, I have no trouble updating this.

spytheman: Consider the most frequent usage - comparing text, generated by vfmt, to input source, that was written by a human, using an unknown text editor.
The input may not have an ending line. The output will.

Consider also that such diffs, will be used for `assert txt1 == txt2`.
When those 2 differ, it is very valuable to know what the exact difference was, and that is easier to discern, when you know that the tool does not lie to you, by manipulating and hiding data.

ttytm: Hmm, I still think there can be the possibility for a misunderstanding here - on both sides. So lets try to get things clear.

There still is `compare_files`, which is used by `v fmt ` and it also still prints `No newline at end of file` if it is the case for the formatted file. I would include a test case for this as well and already should have.

In regards of asserting before, how there will be a difference? the + '\n' is added to both asserted strings, and and it is added for a "quasi"-formatting-reasons of formatting the colored diff, so that the result resembles the function signature better.

spytheman: Spotting `\n` in the output will be harder (especially if the report also has its own newlines, so that it will result in 2 newlines vs 1 newline + "no newline" message + 1 newline).

I do agree with the other objection though. As long as vfmt is not affected, it would not be such a problem.

Can you please just add a small comment in the function description, that the function does add newlines to the strings, before comparing them? Then people that do not want that, can just not use that function.

ttytm: How it already works:
- **Text** comparison
  ```v
  println(diff.compare_text('abc', 'cde', tool: .diff)!)
  ```
  Screenshot_20240503_155710

  Adding additional `\n` to one of the texts
  ```v
  println(diff.compare_text('abc\n', 'cde', tool: .diff)!)
  ```
  Screenshot_20240503_155658

- **File** comparison
  ```v
  println(diff.compare_files('abc.txt', 'cde.txt', tool: .diff)!)
  ```
  Screenshot_20240503_155811


Would adding something below make sense?
```v
@[params]
pub struct CompareTextOptions {
	CompareOptions
pub:
	// Preserves original linefeed. By default, LF is ensured for line endings in compared texts.
	keep_input_linefeed bool
}
```

I have trouble finding the use-case yet but this could make sure there don't have to be compromises if the use-case for this comes up.

spytheman: The output is fine, and I do not think that adding an option is needed at the moment.

I was thinking about just a small comment above the function, something like:
`// the inputs are compared after adding a newline to each`

JalonSolov: However, shouldn't the newline only be added if there isn't one there already?  Otherwise, you could end up with an extra on one, adding a difference that wasn't in the original.

ttytm: > However, shouldn't the newline only be added if there isn't one there already? Otherwise, you could end up with an extra on one, adding a difference that wasn't in the original.

Not necessarily, as it is added to both texts it is not a difference that the tools will track and highlight as a difference. So is will not change what is highlighted in the output. 

I temporarily had a condition for this before opening the PR but thought to skip the additional check when it doesn't create a diff. Now thinking again, re-adding it could remove one additional non-differing empty line after the diffs, depending on the used configuration.

To give examples of what is highlighted with things as they currently are:

<details>

```v
println(diff.compare_text('abc\n\n', 'cde\n', tool: .colordiff)!)
```
Screenshot_20240503_180441

```v
println(diff.compare_text('abc\n\n', 'cde\n', tool: .colordiff, args: '-y -W 80')!)
```
Screenshot_20240503_181739

```v
println(diff.compare_text('abc\n\n\n', 'cde\n\n')!)
```
Screenshot_20240503_181339

</details>

JalonSolov: That extra "non-different empty line" is what I was talking about.  If it isn't there in the original text, it shouldn't show up in the output, either.

ttytm: Agreed, and added it. It ensures a cleaner visualization when using custom options. The results of the comparison itself that shows the differences stays ofc the same.

vlib/v/util/diff/diff.v
```diff
@@ -92,11 +84,37 @@ pub fn compare_text(text1 string, text2 string, opts CompareOptions) !string {
...
 	// Add `\n` when comparing strings to prevent `\ No newline at end of file` in the output.
-	os.write_file(path1, text1 + '\n')!
-	os.write_file(path2, text2 + '\n')!
+	if !text1.ends_with('\n') && !text2.ends_with('\n') {
+		os.write_file(path1, text1 + '\n')!
+		os.write_file(path2, text2 + '\n')!
+	} else {
+		os.write_file(path1, text1)!
+		os.write_file(path2, text2)!
+	}
```
JalonSolov:
```suggestion
	if !text1.ends_with('\n') {
		os.write_file(path1, text1 + '\n')!
	} else {
		os.write_file(path1, text1)!
	}

	if !text2.ends_with('\n'){
		os.write_file(path2, text2 + '\n')!
	} else {
		os.write_file(path2, text2)!
	}
```
... since one might have it, and the other might not.

vlib/v/util/diff/diff.v
ttytm: You know what, I removed the line handling. I can live with that.

Personally, when there would be a comment like that, and I would read it as a user IT would make the function confusing to me. The handling of `\n` line endings in the string comparison function was meant as something internal.

Now if the **string** comparison function is printing about line endings in a **file** and users don't like that, they can still handle the line endings themselves. That's would I would do in my application then.

vlib/v/util/diff/diff.v
ttytm: But this would add a diff

ttytm: You can't add something just to one file and compare it with another.

ttytm: The condition was working as it should. Testing it with different custom configurations it produced the right results. 

removed th eol handling though.

spytheman:
How is `VDIFF_TOOL` and `VDIFF_OPTIONS` supported in the new code?

From what I can see, they are only read now in `find_working_diff_command`, which is deprecated in this PR after 2024-06-30.

What will happen, when it expires?

spytheman:
Why is find_working_diff_command deprecated at all, if there is not a suitable replacement, offering the same functionality?

ttytm:
> Why is find_working_diff_command deprecated at all, if there is not a suitable replacement, offering the same functionality?

Because it is not required. Or do I miss something?
As far as I could see, there isn't a need to make it necessary to use a module function to find the diff command to then pass it to another module function. This can be handled internally. 

```v
// master
diff.color_compare_strings(diff.find_working_diff_command()!, 'unique_smthn', 'abc', 'cde')

// updated
diff.compare_text('abc', 'cde')!
```
or 
```
// master
diff_cmd := diff.find_working_diff_command()!
// ...
diff.color_compare_files(diff_cmd, file_path_1, file_path2)

// updated
diff.compare_text(file_path1, file_path_2)!
```

The automatically available diff tool is still available as constant `available_tool`. Which can be `none` if none is available.

> What will happen, when it expires?

It can removed then since it's not required.

ttytm:
> How is VDIFF_TOOL and VDIFF_OPTIONS supported in the new code?

They are part of the old code and replaced by an alternative.

By default, it is `VDIFF_CMD` that can be used to pass a custom tool and options. 

To give developers full control, and make the module attractive to use anywhere, the env variable it self is also configurable in `CompareOptions`. It can be customized. Also set to `none`, which would disable overwrites - in case developers don't want to allow to overwrite values that were passed to a diff function.

The new variable is handled in a way should better support custom options.

E.g. current master (this is stated to be working in `v fmt --help`, but doesn't)
```sh
VDIFF_TOOL="diff" VDIFF_OPTIONS="-W 120 -y" v fmt -diff file.v
```
No issues with the new command
```sh
VDIFF_CMD="diff -W 120 -y" v fmt -diff file.v
```
**Screenshots and another example for this are in the last section of the PRs description.**

When starting to use the new functions, of course documentation should be updated in the corresponding locations, like `v fmt --help`.

vlib/v/util/diff/diff.v
ttytm: Having had a long day yesterday and other things to close off, I just wanted to get done with things in the PR. But I see, that I wanted to go to fast here. This PR is a small "bigger" change. So it's okay to go as slow as required. 

To get the best result. I'll re-add preventing of `\ No newline at end of file`, as I think it's the right thing to do for that function. Then we can continue discussing it. If there should be any concerns with that, it would help me to address them if you test the function locally. Maybe this resolves concerns in itself or you can find an example and we would have something concrete to discuss.

ttytm:
It is intended that everything is intuitive to use, please bare with me when explanations make things seem complicated. I'm open for questions if there are more concerns.

PR #21404, opened 2024-05-02
Fix os.execute stderr redirection

### LINKED ISSUE #20986: vlib: os.execute and arguments with linefeeds causes invalid exit_code
### Describe the bug

`os.execute` does not sanitize arguments properly and the exit_code cannot be trusted.

### Reproduction Steps

test1() and test2() should present the same exit code (assuming there is no process 19935)

```v
import os

fn test1()
{
	eprintln("probling 19935")
	pid := "19935"
	status := os.execute("ps $pid")
	eprintln(status)
	
}

fn test2()
{
	eprintln("probing 19935 with a line feed at the end")
	pid := "19935\n"
	status := os.execute("ps $pid\n")
	eprintln(status)
}


fn main() {
	test1()
	test2()
}
```

### Expected Behavior

test1 and test2 should both return the same exit code (1) given that there is no process 19935.

### Current Behavior

test2 function returns exit code 0 even though test1 returns 1 (the expected exit code)

### Possible Solution

_No response_

_No response_

### V version

V 0.4.4 cf7dcfe

V full version: V 0.4.4 cf7dcfe
OS: linux, "Artix Linux"
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz

getwd: /home/root
vexe: /home/root/Apps/v/v
vexe mtime: 2024-02-24 09:52:28

VTMP: OK, value: /tmp/v_0

Git vroot status: weekly.2024.10-3-ga867ed6a (6 commit(s) behind V master)
.git/config present: true

> [!NOTE]
>

hholst80:
Closes #20986

### DIFF
--- a/vlib/os/os_nix.c.v
+++ b/vlib/os/os_nix.c.v
@@ -339,10 +339,7 @@ pub fn mkdir(path string, params MkdirParams) ! {
 // execute starts the specified command, waits for it to complete, and returns its output.
 @[manualfree]
 pub fn execute(cmd string) Result {
-	// if cmd.contains(';') || cmd.contains('&&') || cmd.contains('||') || cmd.contains('\n') {
-	// return Result{ exit_code: -1, output: ';, &&, || and \\n are not allowed in shell commands' }
-	// }
-	pcmd := if cmd.contains('2>') { cmd.clone() } else { '${cmd} 2>&1' }
+	pcmd := 'exec 2>&1;${cmd}'
 	defer {
 		unsafe { pcmd.free() }
 	}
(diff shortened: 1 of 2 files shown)


hholst80:
We probably need to tune this on non-posix platforms like Windows

$if windows [??? don't change cmd ] $else [the logic we have]

hholst80:
The full CI/CD suite takes over two hours? Jikes. It would be interesting to see the entire trace of it where potential bottlenecks are.

JalonSolov:
The bottleneck is GitHub throttling.  Sometimes the entire CI suite runs in just a few minutes.  Other times, I've seen it take 4 - 5 hours.

spytheman (approved):

PR #21407, opened 2024-05-02
time: fix string representation of negative Duration

### LINKED ISSUE #20965: vlib: time.Duration and `-` operator leaves spurious ns reminder for a - b < 0 case.
### Describe the bug

For a, b : time.Time the `-` operator will floor the components of the time.Duration, except the ns field. 

### Reproduction Steps

```v
import time
fn main() {
  now := time.now()
  yesterday := time.parse("2024-03-04 00:00:00")!
  delta := yesterday - now
  if delta == time.Duration(0) {
    println("we will most likely not see this")
  } else {
    println("delta: $delta")
  }
}
```

### Expected Behavior

time.Duration(0) identically.

### Current Behavior

ns field is nonzero.

### Possible Solution

Floor the ns to identically zero.

_No response_

### V version

V 0.4.4 cf7dcfe

```
V full version: V 0.4.4 cf7dcfe
OS: linux, "Artix Linux"
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz

getwd: /home/root
vexe: /home/root/Apps/v/v
vexe mtime: 2024-02-27 14:30:38

VTMP: OK, value: /tmp/v_0

Git version: git version 2.43.1
Git vroot status: weekly.2024.08-15-gee045eda (35 commit(s) behind V master)
.git/config present: true

```

hholst80:
Closes #20965

### DIFF
--- a/vlib/time/duration.v
+++ b/vlib/time/duration.v
@@ -63,7 +63,12 @@ pub fn (d Duration) str() string {
 	if d == time.infinite {
 		return 'inf'
 	}
+	mut sign := ''
 	mut t := i64(d)
+	if t < 0 {
+		sign = '-'
+		t = -t
+	}
 	hr := t / time.hour
 	t -= hr * time.hour
 	min := t / time.minute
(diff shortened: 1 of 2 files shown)


hholst80:
@JalonSolov how do you think negative Duration should be represented as strings? just prefix with a '- ' like in debug string representation?

vlib/time/duration.v
```diff
@@ -63,7 +63,12 @@ pub fn (d Duration) str() string {
 	if d == time.infinite {
 		return 'inf'
 	}
+	mut sign := ''
 	mut t := i64(d)
+	if t < 0 {
+		sign = '- '
```
spytheman, line 69: Why use '- ' and not just '-'?

hholst80: I think the debug string used an extra space so I took that.

I am not too satisfied with either of these.

What about adding parenthesis like they do in accounting?


https://trumpexcel.com/show-negative-numbers-parentheses-brackets-excel/

JalonSolov: Because most developers aren't accountants?

Definitely should just be `-` not `- `.

JalonSolov:
> @JalonSolov how do you think negative Duration should be represented as strings? just prefix with a '- ' like in debug string representation?

It should look like a negative number you typed in, so `-10`, not `- 10` - that looks like you're trying to subtract 10 from something.

spytheman (approved):
Good work.

PR #21408, opened 2024-05-02
builder, ci: fix compilation on macos-arm with cstrict; run macos ci also on arm runner

ttytm:
- Fixes compilation with `-cstrict` on macos-13 and macos-arm. This is done by only adding `-x none` when there would be actual paths that follow it.
  ```sh
  ./v -cg -cstrict -o v cmd/v
  ```
- Adds arm runner to macos wokflow.

### DIFF
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -415,8 +415,8 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
 			}
 			dir := os.dir(if flag.starts_with('"') && flag.ends_with('"') {
 				flag[1..flag.len - 1]
-				} else {
-					flag
+			} else {
+				flag
 			})
 			if os.is_dir(dir) {
 				ccoptions.source_args << '-x none'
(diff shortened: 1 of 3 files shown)


ttytm:
@spytheman can I add the let the openssl test be skipped when testing with macos > 12?

I couldn't resolve it here by adding the modern openssl path to the library path.
Having tested and checked log `export LIBRARY_PATH="$LIBRARY_PATH:$(brew --cellar)/openssl@3/3.3.0/lib/"` should add it, though the test failure will remain the same.

spytheman:
> @spytheman can I let the openssl test be skipped in CI runs when testing with macos > 12?

Yes, just skip it for that job.

vlib/v/builder/cc.v
```diff
@@ -415,6 +404,38 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
...
+		}
+	}
+	if !v.pref.no_std {
+		if v.pref.os == .linux {
+			ccoptions.source_args << '-std=${builder.c_std_gnu}'
+		} else {
+			ccoptions.source_args << '-std=${builder.c_std}'
+		}
+		ccoptions.source_args << '-D_DEFAULT_SOURCE'
+	}
```
spytheman, line 438: This should be before the `$if trace_ccoptions ? {` line.
Doing it after it, defeats its purpose (to have an easy way to inspect the results for diagnostics).

spytheman (approved):

vlib/v/builder/cc.v
ttytm: Whops. Yes that how it should be. Thanks you :+1:

PR #21409, opened 2024-05-03
os: fix `dir` function for quoted paths

ttytm:
When passing a quoted path to `os.dir` the change will result in: 

```v
os.dir('"foo/bar/baz"') // -> `"foo/bar"` instead of `"foo/bar`
```

### DIFF
--- a/vlib/os/os.v
+++ b/vlib/os/os.v
@@ -291,9 +291,9 @@ pub fn dir(opath string) string {
 	if pos == 0 && path_separator == '/' {
 		return '/'
 	}
-	quote := match path[0] {
-		`'` { "'" }
-		`"` { '"' }
+	quote := match true {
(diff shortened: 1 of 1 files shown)


spytheman:
One should not pass quoted paths to os.dir().

vlib/os/os.v
```diff
@@ -291,7 +291,12 @@ pub fn dir(opath string) string {
 	if pos == 0 && path_separator == '/' {
 		return '/'
 	}
-	return path[..pos]
+	quote := match true {
+		opath[0] == `'` && opath[opath.len - 1] == `'` { "'" }
+		opath[0] == `"` && opath[opath.len - 1] == `"` { '"' }
+		else { '' }
+	}
+	return path[..pos] + quote
```
spytheman, line 299: no, os.dir should not have to care about " or ' at all.

PR #21412, opened 2024-05-04
vlib: add port() for net.Addr

vcker:
When I start write a reverse proxy for my android / ios app, I need a port that is always available, so I set address `':0'`, then I try to obtain port that automatically assigned, but i can't find a function like `a.port()`.  Although I can parse the address to get the port, but this is inelegant, so i add function `.port()`.

```v
fn main() {
	mut server := net.listen_tcp(.ip, ':0') or { panic(err) }
	println(server.addr()!)
	println(server.addr()!.port()!) // add port func
	mut socket := server.accept()!
	// ...
}
```

```v
// port returns the ip or ip6 port of the given address `a`
pub fn (a Addr) port() !u16 {
	match unsafe { AddrFamily(a.f) } {
		.ip {
			unsafe {
				return conv.ntoh16(a.addr.Ip.port)
			}
		}
		.ip6 {
			unsafe {
				return conv.ntoh16(a.addr.Ip6.port)
			}
		}
		.unix {
			return error('unix addr has no port')
		}
		.unspec {
			return error('unspec addr family when obtain port')
		}
	}
}
```

```v
// address_port_test.v
import net
import strconv

fn test_ip_port(){
	mut server := net.listen_tcp(.ip, ':0') or {panic(err)}
	port_str := server.addr()!.str().split(':').last()
	port  := strconv.parse_uint(port_str, 10, 16)!
	assert port == server.addr()!.port()!
}

fn test_ip6_port(){
	mut server := net.listen_tcp(.ip6, ':0') or {panic(err)}
	port_str := server.addr()!.str().split(':').last()
	port  := strconv.parse_uint(port_str, 10, 16)!
	assert port == server.addr()!.port()!
}

fn test_unix_port(){
	a := net.Addr{
		f: u8(net.AddrFamily.unix)
	}

	p := a.port() or {
		assert err.str() == 'unix addr has no port'
		0
	}
	assert p == 0
}

fn test_unspec_port(){
	a := net.Addr{
		f: u8(net.AddrFamily.unspec)
	}

	p := a.port() or {
		assert err.str() == 'unspec addr family when obtain port'
		0
	}
	assert p == 0
}
```

### DIFF
--- a/vlib/net/address.c.v
+++ b/vlib/net/address.c.v
@@ -60,6 +60,28 @@ pub fn (a Addr) family() AddrFamily {
 	return unsafe { AddrFamily(a.f) }
 }
 
+// port returns the ip or ip6 port of the given address `a`
+pub fn (a Addr) port() !u16 {
+	match unsafe { AddrFamily(a.f) } {
+		.ip {
+			unsafe {
+				return conv.ntoh16(a.addr.Ip.port)
+			}
+		}
+		.ip6 {
+			unsafe {
+				return conv.ntoh16(a.addr.Ip6.port)
+			}
+		}
+		.unix {
+			return error('unix addr has no port')
+		}
+		.unspec {
+			return error('unspec addr family when obtain port')
+		}
+	}
+}
+
 const max_ip_len = 24
 const max_ip6_len = 46
 
--- a/vlib/net/address_test.c.v
+++ b/vlib/net/address_test.c.v
@@ -1,5 +1,10 @@
 module net
 
+fn test_ip_port() {
+	assert new_ip(1234, addr_ip_any).port()! == 1234
+	assert new_ip6(1234, addr_ip6_any).port()! == 1234
+}
+
 fn test_diagnostics() {
 	dump(aoffset)
 	eprintln('--------')


Delta456 (commented):
Run `v fmt -w` on this file.

vcker:
> Run `v fmt -w` on this file.

no difference after do this

vcker:
@Delta456 it looks `v up` auto checks  faild, is it my wrong? what should i do next?

JalonSolov:
The only 2 failed checks were `v up` checks, which were changed recently.  @spytheman  @ttytm

spytheman:
It think, that it will need rebasing over current master for all CI checks to pass.

The most likely reason for it to fail like that with:
> Commit: "206441c9df3413dd2e877d35bd3d4b8f3edb69ac" does not exist in the current repository.

... even though that commit is present on master, is that `bootstrapping_ci.yml` checkouts the branch of the PR, which was probably made before that `206441c9` was merged on master.

We would have to correct bootstrapping_ci.yml so that it is more reliable, however that is an unrelated change, and in the meantime, rebasing with `git pull --rebase origin master`, then pushing with `git push -f` should fix it.

@ttytm what do you think?

spytheman:
That 
image
cancellation is a bit more puzzling, since our FreeBSD runner script in `.cirrus.yml` does not do anything that is against their rules: <https://cirrus-ci.org/faq/#repository-is-blocked> 🤔.

It may be just due to hitting some limit, which unfortunately does not seem to be documented.

I'll test manually the PR on a FreeBSD vps.

spytheman:
@ttytm, about the bootstrapping CI, what do you think of changing it to fetch an older master commit, say 20 or 30 commits older than the current head?

We do merge 10 commits per day relatively frequently, but very rarely over 12. If it checkouts a commit that is over 20 steps removed from the present, I think it will minimize the chance of similar failures (and people that work for several days before making their PRs usually already do rebase or merge before that).

vlib/net/address.c.v
```diff
@@ -60,6 +60,28 @@ pub fn (a Addr) family() AddrFamily {
...
+		.ip6 {
+			unsafe {
+				return conv.ntoh16(a.addr.Ip6.port)
+			}
+		}
+		.unix {
+			return error('unix addr has no port')
+		}
+		.unspec {
+			return error('unspec addr family when obtain port')
```
spytheman, line 80: @JalonSolov does that read well?

spytheman:
As expected, tests FreeBSD on pass (except for `ftp_test.v`, but that is a known failure and is not related to this PR), when run manually.

spytheman:
@vcker everything looks good to me. 

Can you please just add a small test for the new method in `vlib/net/address_test.c.v`, something like:
```v
fn test_ip_port() {
        assert new_ip(1234, addr_ip_any).port()! == 1234
        assert new_ip6(1234, addr_ip6_any).port()! == 1234
}
```

ttytm:
> I think, that it will need rebasing over current master for all CI checks to pass.
> 
> The most likely reason for it to fail like that with:
> 
> > Commit: "206441c9df3413dd2e877d35bd3d4b8f3edb69ac" does not exist in the current repository.
> 
> ... even though that commit is present on master, is that `bootstrapping_ci.yml` checkouts the branch of the PR, which was probably made before that `206441c9` was merged on master.
> 
> We would have to correct bootstrapping_ci.yml so that it is more reliable, however that is an unrelated change, and in the meantime, rebasing with `git pull --rebase origin master`, then pushing with `git push -f` should fix it.
> 
> @ttytm what do you think?

Already addressed here: #21414. So running into this issue should not happen anymore.

ttytm:
If re-running doesn't solve this. Then it shouldn't happen anymore after a rebase.

spytheman:
> Already addressed here: #21414. So running into this issue should not happen anymore.

Excellent. Thank you.

vlib/net/address.c.v
JalonSolov: Hmm... how about 
```suggestion
			return error('cannot find port for unspec addr family')
```

spytheman:
The bootstrapping CI passed, and the Cirrus FreeBSD job also passed after the rebasing.

Imho it is good to merge.

vlib/net/address.c.v
spytheman: oops, I saw that too late ... I'll fix it on master

spytheman: Fixed in 44c84f6 .

PR #21414, opened 2024-05-04
ci: ensure `master` is available when trying to check out it's commits to build oldv

ttytm:
Fixes workflow fails of the `Bootstrapping CI` on forks and occasionally in PRs. 

Failures happen, because oldv receives a commit that is available on master, but the master branch itself is not available in the workflow due to a shallow checkout of the checkout action.

The fix is simple by ensuring the master branch is available.

### DIFF
--- a/.github/workflows/bootstrapping_ci.yml
+++ b/.github/workflows/bootstrapping_ci.yml
@@ -40,7 +40,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 10
+          fetch-depth: 0
       - name: Build V
         run: make -j4
       - name: Test bootstrapping (v.c can be compiled and run with -os cross)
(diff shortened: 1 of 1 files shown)


.github/workflows/bootstrapping_ci.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          fetch-depth: 10
+          fetch-depth: 0
```
ttytm, line 43: `0` to fetch other branches. Ref. https://github.com/actions/checkout

ttytm:
Fork workflow refs.:
Current: https://github.com/ttytm/v/actions/runs/8950146423
Updated: https://github.com/ttytm/v/actions/runs/8950336144

PR #21417, opened 2024-05-04
vet, parser: rewrite vet error handling (improve parser performance extend vvet) p1

ttytm:
Currently, vet errors are validated and filtered in the parser.
 
The changes will help with to take vet to the next level while also simplifying the parser and improving it's performance. Similar to recent changes, this also fits under the mantra "Parse, don't validate".

This will also resolve TODOS, which should not be TODOs in those modules.

> ```v
> // TODO: `$if vet {` for performance`
> ```

To simplify reviewing things I'll try to submit the changes in a few parts that can hopefully stand on their own without falling appart. The parts should become simpler when things progress. 

The first part creates a base for vets evolution and extends it's ability to handle  trailing space and indentation errors on it's own. This already takes load of the parser.


ttytm:
vet tests and even more so`v test-cleancode` should cover the changes.

PR #21421, opened 2024-05-04
tools.vet: add notice for empty strings conditions

ttytm:
After getting a response from Alex that it's fine to add the suggestion made in https://github.com/vlang/v/pull/21300  this should be the frist extension after #21417. 

```v
fn main() {
	foo := 'foo'
	foo.len == 0
	foo.len != 0
	foo.len < 1
	foo.len > 0
}
```

will result in:

```
cmd/tools/vvet/tests/empty_string.vv:3: notice: Use `foo == ''` instead of `foo.len == 0`
cmd/tools/vvet/tests/empty_string.vv:4: notice: Use `foo != ''` instead of `foo.len != 0`
cmd/tools/vvet/tests/empty_string.vv:5: notice: Use `foo == ''` instead of `foo.len < 1`
cmd/tools/vvet/tests/empty_string.vv:6: notice: Use `foo != ''` instead of `foo.len > 0`
```

---

To keep this as a ref.: as mentioned here(https://github.com/vlang/v/pull/21523#issuecomment-2118922595), adding this check for globally scoped symbols is still a *TODO*.


spytheman:
The Vinix failure is not related, and already fixed on master.

cmd/tools/vvet/vvet.v
```diff
@@ -318,6 +322,41 @@ fn (mut vt Vet) const_decl(stmt ast.ConstDecl) {
...
 }
 
+fn (mut vt Vet) vet_empty_str(expr ast.InfixExpr) {
+	if expr.left is ast.InfixExpr {
+		vt.expr(expr.left)
+	} else if expr.right is ast.InfixExpr {
+		vt.expr(expr.right)
+	} else if expr.left is ast.SelectorExpr && expr.right is ast.IntegerLiteral
+		&& expr.left.field_name == 'len' {
+		if expr.right.val == '0' && expr.op != .lt {
```
spytheman, line 332: Would not that trigger for `a.len == 0` too, where `a` is an array, and not a string?

ttytm: Important point. It would. It's not strict enough and needs to be limited to stings. I missed that.

ttytm: Similar to what I was encountering when I trying to implement a `v fmt` autofix when recently moving from `time.unix` to `time.unix()` I'm running into the issue that the type of the selector expression is not accessible.

Likely it will be necessary first to look at parsing those types in order to enable such fixes and the feature I'm trying to implement here.

spytheman: The checker has that information, but currently `v vet` is run on single files, not whole programs, thus it cannot use the checker.

spytheman: I think that a potential solution, is to move the notice to the checker.

spytheman (approved):
Excellent work.

PR #21422, opened 2024-05-04
vet, parser: move array_init_one_val checks from parser into vet

ttytm:
Due to similar locations changed in vvet this might conflict with #21421.
It's recommended to await the re-scan after one is resolved.


ttytm:
The re-scan I was referring two is the short one that is done after another merge master. Should things be safe for departure :steam_locomotive:

spytheman:
Thanks for checking.

PR #21424, opened 2024-05-04
vet, parser: remove vet_errors and vet_notices from parser

ttytm:
After #21422, #21423 this PR can already remove vet_errors and vet_notices from the parser. Extending vvet to be able to handle the scanner errors as well would be the next step.


spytheman (approved):
Excellent work.

PR #21425, opened 2024-05-04
vlib.os: fix join-path

### LINKED ISSUE #20231: v fails to run and generate binaries from / folder
### Describe the bug

The compiler will not generate a binary if it is invoked on a file in the root.

### Reproduction Steps

```
/ # cat hello.v
fn main() {
	println("Hello, world!");
}
/ # /v/v hello.v
/ # ls -a
.           bin         foo         home        mnt         root        srv         usr
..          dev         hello.c     lib         opt         run         sys         v
.dockerenv  etc         hello.v     media       proc        sbin        tmp         var
/ # /v/v run hello.v
No such file or directory; code: 2
/ # 
```

### Expected Behavior

It should compile a `hello` binary placed in / and it should run the code in the `v run hello.v` instance.

### Current Behavior

Silent ignores to generate binary and No such file or directory; code 2 error message (not helpful)

### Possible Solution

Fix the path assumptions that the source file is located in a sub directory.

_No response_

### V version

V 0.4.3 5306469

Docker container:

```
/ # cat /etc/os-release 
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.19.0
PRETTY_NAME="Alpine Linux v3.19"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
/ # 
```
> [!NOTE]
>

hholst80:
Closes #20231

### DIFF
--- a/vlib/os/os.v
+++ b/vlib/os/os.v
@@ -568,6 +568,8 @@ pub fn is_file(path string) bool {
 
 // join_path joins any number of path elements into a single path, separating
 // them with a platform-specific path_separator. Empty elements are ignored.
+// Windows platform output will rewrite forward slashes to backslash.
+// Consider looking at the unit tests in os_test.v for semi-formal API.
 @[manualfree]
 pub fn join_path(base string, dirs ...string) string {
 	// TODO: fix freeing of `dirs` when the passed arguments are variadic,
@@ -576,25 +578,36 @@ pub fn join_path(base string, dirs ...string) string {
 	defer {
 		unsafe { sb.free() }
 	}
-	sbase := base.trim_right('\\/')
-	defer {
-		unsafe { sbase.free() }
+	mut needs_sep := false
+	if base != '' {
+		$if windows {
+			sb.write_string(base.replace('/', '\\'))
+		} $else {
+			sb.write_string(base)
+		}
+		needs_sep = !base.ends_with(path_separator)
 	}
-	sb.write_string(sbase)
-	for d in dirs {
-		if d != '' {
-			sb.write_string(path_separator)
-			sb.write_string(d)
+	for od in dirs {
+		if od != '' && od != '.' {
+			mut md := od
+			$if windows {
+				md = md.replace('/', '\\')
+			}
+			// NOTE(hholst80): split_any not available in js backend,
+			// which could have been more clean way to implement this.
+			nestdirs := md.split(path_separator)
+			for id in nestdirs {
+				if id != '' && id != '.' {
+					if needs_sep {
+						sb.write_string(path_separator)
+					}
+					sb.write_string(id)
+					needs_sep = !id.ends_with(path_separator)
+				}
+			}
 		}
 	}
-	mut res := sb.str()
-	if sbase == '' {
-		res = res.trim_left(path_separator)
-	}
-	if res.contains('/./') {
-		// Fix `join_path("/foo/bar", "./file.txt")` => `/foo/bar/./file.txt`
-		res = res.replace('/./', '/')
-	}
+	res := sb.str()
 	return res
 }
 
--- a/vlib/os/os_test.c.v
+++ b/vlib/os/os_test.c.v
@@ -619,11 +619,16 @@ fn test_join() {
 		assert os.join_path('v', 'vlib', 'os') == 'v\\vlib\\os'
 		assert os.join_path('', 'f1', 'f2') == 'f1\\f2'
 		assert os.join_path('v', '', 'dir') == 'v\\dir'
+		assert os.join_path('v', 'foo/bar', 'dir') == 'v\\foo\\bar\\dir'
+		assert os.join_path('v', 'foo/bar\\baz', '/dir') == 'v\\foo\\bar\\baz\\dir'
+		assert os.join_path('C:', 'f1\\..', 'f2') == 'C:\\f1\\..\\f2'
 	} $else {
 		assert os.join_path('v', 'vlib', 'os') == 'v/vlib/os'
 		assert os.join_path('/foo/bar', './file.txt') == '/foo/bar/file.txt'
 		assert os.join_path('', 'f1', 'f2') == 'f1/f2'
 		assert os.join_path('v', '', 'dir') == 'v/dir'
+		assert os.join_path('/', 'test') == '/test'
+		assert os.join_path('/foo/bar', './.././file.txt') == '/foo/bar/../file.txt'
 	}
 }
 
--- a/vlib/os/dir_expansions_test.v
+++ b/vlib/os/dir_expansions_test.v
@@ -23,7 +23,7 @@ fn test_ensure_folder_is_writable() {
 
 fn test_expand_tilde_to_home() {
 	os.setenv('HOME', '/tmp/home/folder', true)
-	os.setenv('USERPROFILE', '/tmp/home/folder', true)
+	os.setenv('USERPROFILE', r'\tmp\home\folder', true)
 	//
 	home_test := os.join_path(os.home_dir(), 'test', 'tilde', 'expansion')
 	home_expansion_test := os.expand_tilde_to_home(os.join_path('~', 'test', 'tilde',


vlib/os/os.v
```diff
@@ -570,54 +570,55 @@ pub fn is_file(path string) bool {
...
-		sb.write_string(sbase)
-		sb.write_string(path_separator)
+	if base != '' && elem != '' {
+		return '${base}${path_separator}${elem}'
+	} else if base != '' {
+		return base
+	} else if elem != '' {
+		return elem
+	} else {
+		return ''
```
ttytm, line 620: A match handy when having for such `if - else if - else`s 
```suggestion
	return match true {
		base != '' && elem != '' { '${base}${path_separator}${elem}' }
		base != '' { base }
		elem != '' { elem }
		else { '' }
```

vlib/os/os_test.c.v
```diff
@@ -621,9 +621,10 @@ fn test_join() {
 		assert os.join_path('v', '', 'dir') == 'v\\dir'
 	} $else {
 		assert os.join_path('v', 'vlib', 'os') == 'v/vlib/os'
-		assert os.join_path('/foo/bar', './file.txt') == '/foo/bar/file.txt'
+		assert os.join_path('/foo/bar', './file.txt') == '/foo/bar/./file.txt'
```
spytheman, line 624: That is a regression.

vlib/os/os.v
```diff
@@ -570,54 +570,52 @@ pub fn is_file(path string) bool {
 // them with a platform-specific path_separator. Empty elements are ignored.
 @[manualfree]
 pub fn join_path(base string, dirs ...string) string {
-	// TODO: fix freeing of `dirs` when the passed arguments are variadic,
-	// but do not free the arr, when `os.join_path(base, ...arr)` is called.
-	mut sb := strings.new_builder(base.len + dirs.len * 50)
+	mut sb := strings.new_builder(256)
```
spytheman, line 573: The previous estimate for the capacity was usually better.

vlib/os/os.v
```diff
@@ -570,54 +570,52 @@ pub fn is_file(path string) bool {
...
-		// Fix `join_path("/foo/bar", "./file.txt")` => `/foo/bar/./file.txt`
-		res = res.replace('/./', '/')
-	}
+	res := sb.str()
 	return res
 }
 
 // join_path_single appends the `elem` after `base`, separated with a
 // platform-specific path_separator. Empty elements are ignored.
-@[manualfree]
```
spytheman, line 603: This is important. The function should not leak when used with `-autofree`. Removing it is a regression.

vlib/os/os.v
```diff
@@ -570,54 +570,52 @@ pub fn is_file(path string) bool {
...
-	}
-	sbase := base.trim_right('\\/')
-	defer {
-		unsafe { sbase.free() }
-	}
-	if base != '' {
-		sb.write_string(sbase)
-		sb.write_string(path_separator)
+	return match true {
+		base != '' && elem != '' { '${base}${path_separator}${elem}' }
```
spytheman, line 614: The interpolation here will leak, when used with `-autofree` currently.
Using a string builder instead, looks less compact in the source, but prevents it.

vlib/os/os.v
hholst80: Better how? 256 or some max-path thing is sufficient I think, that what we would find in C implementation. A os.join_path is unlikely to be part of a tight inner loop in a performance critical loop.

vlib/os/os.v
hholst80: But we don't have anything to free? I agree that a function should not leak, but how does this leak compared to any other function in the library now that return an interpolated string?

vlib/os/os.v
hholst80: I would rather then remote the _single variant completely fromt he stdlib.  It serves no purpose what so ever as it stands.

vlib/os/os_test.c.v
hholst80: should os.join_path do normalization or not? That is a pretty big task. I think real_path can be used to cleanup it.  Compare to os.path.join in Python they also do not normalize ./ it is important that the file will be interpreted correctly, not how it looks like. "foo/./bar" is perfectly valid path and is equivalent to "foo/bar"

Any kind of logic we expose here will be maintained forever. Less is more.

vlib/os/os.v
spytheman: Yes, it does, until the compiler improves.

vlib/os/os.v
spytheman: Usually the calculated capacity will be smaller than 256 .

vlib/os/os_test.c.v
spytheman: It is better to have less edge cases in the output for this particular function.
The call sites can then be simpler.

vlib/os/os.v
spytheman: `os.join_path_single` is used in `vlib/v/slow_tests/valgrind/import_os_and_use_its_constants.v` . The programs in that folder are compiled and tested with `-autofree` . 
I do not want to chase regression bugs in them, caused by a `@[manualfree]` tagged function, not being `@[manualfree]` anymore.

vlib/os/os.v
hholst80: Would it be OK just to defer the call to the main function in this PR and then I add another PR to factor it out (that is, remove the calls to _single with a direct call?)

It looks kinda silly to use a string builder to join two strings. Calling the main handler will at least look less silly. :)

vlib/os/os.v
hholst80: Does it matter? It is a heuristic that also takes time to compute because of the .len.  If we need actual precision could just run through the strings and actually compute the need.  counting every token that will go into the builder.

vlib/os/os.v
spytheman: I do not care what looks silly. I care that tests pass, and that `-autofree` bugs related bugs will not have to be chased in complex places like the compiler itself.

Until the compiler improves and can distinguish between varargs and arrays in relation to autofree, os.join_path_single has its usage, and it should stay.

vlib/os/os.v
spytheman: Usually it does not, but os.join_path and os.join_path_single are called a lot by the compiler, during a normal compilation. The calculation based on .len, is cheap, compared to allocating 256 bytes, when a smaller piece of memory could be used.

vlib/os/os.v
hholst80: Sure, so which test does this break?

vlib/os/os.v
hholst80: So is there a test that breaks now? valgrind sounds like something would pop out if I introduced a regression, no? Can we add a test that fails due to my change?

vlib/os/os.v
hholst80: I can revert this change if you feel strongly about your code that it offers true benefit besides being more complex than just a common max_path dictum. People will look at this code and wonder why 50, where does this comes from, what would the implications be of changing this to 40, how low can we go, where is this measured what the optimal constant is.

vlib/os/os_test.c.v
hholst80: Neither C++ nor Python does this. I think this would be confusing and hard to maintain properly.  If the caller needs proper normalization it should call normalize path after the call.

```c++
# cat t.cpp
#include <iostream>
#include <filesystem>

int main() {
  std::filesystem::path p1 = "/";
  std::filesystem::path p2 = "./test";
  std::filesystem::path p = p1 / p2;
  std::cout << p << std::endl;
}
[I] root@trump ~ 
# g++ -o t t.cpp
[I] root@trump ~ 
# ./t
"/./test"
[I] root@trump ~ 
# 
```

vlib/os/os.v
spytheman: It will break, once you trigger the interpolation in a valgrind tested program.
I do not want to have to go back to it, and then change it to use a string builder again. I've done that previously, and chasing such problems is not fun.

vlib/os/os.v
spytheman: I think that an accurate loop for pre-estimating the length, will be better, than a rough estimate like `base.len + dirs.len * 50`, and a rough estimate, is better than a hard coded value like 256 in this case.

vlib/os/os.v
spytheman: > So is there a test that breaks now? valgrind sounds like something would pop out if I introduced a regression, no? Can we add a test that fails due to my change?

Currently no, there is not - all the valgrind tests passed in the CI.

vlib/os/os.v
hholst80: fair enough I will revert those changes.  I did not see anything break in the PR related to valgrind though, perhaps something we can add then to make it damn obvious what needs to be handle with delicate care.

vlib/os/os_test.c.v
spytheman: It does help code that uses it, and it is not proper to make such a change, that breaks an existing test (which then needed changing), in a PR that fixes an unrelated problem (appending to /) .

hholst80: I am happy with the changes as-is. Feel free to reject this PR if you want.

spytheman: > Do not fud my PR with "breaks tests" when obviously no tests are failing.

Obviously, I am commenting under *that very change*.

However, I do not want to argue anymore, and justify my opinion, and I am not happy with your PR as it is. I also asked you something in the other PR, which you are ignoring.

vlib/os/os.v
spytheman: I agree, adding more tests will be better for everyone.

vlib/os/os.v
```diff
@@ -576,25 +576,34 @@ pub fn join_path(base string, dirs ...string) string {
...
 	}
-	mut res := sb.str()
-	if sbase == '' {
-		res = res.trim_left(path_separator)
-	}
-	if res.contains('/./') {
-		// Fix `join_path("/foo/bar", "./file.txt")` => `/foo/bar/./file.txt`
-		res = res.replace('/./', '/')
-	}
+	res := sb.str()
```
ttytm, line 606: As a simpler variant, could this work too?
```suggestion
	mut needs_sep := false
	if base != '' {
		sb.write_string(base)
		needs_sep = !base.ends_with(path_separator)
	}
	for d in dirs {
		if d != '' {
			if needs_sep {
				sb.write_string(path_separator)
			}
			sb.write_string(d)
			needs_sep = !d.ends_with(path_separator)
		}
	}
	res := sb.str()
```

ttytm: The suggestion looks buggy.

I'm referring to the part starting below the defer block:
Current:
```v
	mut i := 0
	mut needs_sep := false
	if base != '' {
		sb.write_string(base)
		needs_sep = !base.ends_with(path_separator)
	} else {
		for i < dirs.len {
			d := dirs[i]
			i++
			if d != '' {
				sb.write_string(d)
				needs_sep = !base.ends_with(path_separator)
				break
			}
		}
	}
	for i < dirs.len {
		d := dirs[i]
		if d != '' {
			if needs_sep {
				sb.write_string(path_separator)
			}
			sb.write_string(d)
			needs_sep = !d.ends_with(path_separator)
		}
		i++
	}
	res := sb.str()
```
Suggested:
```v
	mut needs_sep := false
	if base != '' {
		sb.write_string(base)
		needs_sep = !base.ends_with(path_separator)
	}
	for d in dirs {
		if d != '' {
			if needs_sep {
				sb.write_string(path_separator)
			}
			sb.write_string(d)
			needs_sep = !d.ends_with(path_separator)
		}
	}
	res := sb.str()
```

vlib/os/os_test.c.v
ttytm: I agree with @spytheman about the regression. 

Please also refer to how it's done in go:
```go
package main

import (
	"path/filepath"
)

func main() {
	println(filepath.Join("/foo/bar", "./file.txt")) // -> /foo/bar/file.txt
}
```

vlib/os/os.v
hholst80: Totally agree the default false value makes it possible to merge the loops and get rid of the index i.

spytheman: I think that the approach in <https://github.com/vlang/v/pull/21429> is cleaner.

spytheman: @medvednikov what do you think?

vlib/os/os_test.c.v
hholst80: That is reasonable argument, only mentioned now, that the API should be compatible with Golang. Given how similar the API are to go in other areas like network that V should try and offer few surprises.

See, I like a good discussion, but I am not unreasonable. ;)

ttytm: Idk if they should be compatible. I wanted to put an example of more mature project that also has good engineers. If we find good reasons for it lets break the rules :D.

I think spy did a good work and also extends a lot of test cases. Personally, I like achieving a better product and finding solutions. So when another approaches arise that solve a problem in a more complete way it I usually have no problem giving into it. So we all get a better language :P

Things were just a little unlucky here that the approaches are so different and hardly compatible.

vlib/os/os.v
hholst80: Sure, but it is still incomplete with respect to Golangs "Clean" method:

- https://pkg.go.dev/path#Clean
- https://9p.io/sys/doc/lexnames.html
- https://cs.opensource.google/go/go/+/master:src/internal/filepathlite/path.go

If everyone thinks that this is the gold standard, ok then. I have tried my best to argue for a smaller scope. :)

ttytm:
Since I'm mentioned as a co-author in the full diff, because the parts I tried to help with during development got squashed, I have to say it's not signed off by me.

I think it's honorable that you leave me in there, I thank you for that, @hholst80, but the way it's done might also distort the situation.

Now, aiming to keep things technical and to choose the solution that best serves the project, why not go with the suggestion in Spy's PR and have this one updated? Or what issues are there when doing it like this?

hholst80:
> Since I'm mentioned as a co-author in the full diff, because the parts I tried to help with during development got squashed, I have to say it's not signed off by me.
> 
> I think it's honorable that you leave me in there, I thank you for that, @hholst80, but the way it's done might also distort the situation.
> 
> Now, aiming to keep things technical and to choose the solution that best serves the project, why not go with the suggestion in Spy's PR and have this one updated? Or what issues are there when doing it like this?

I don't think that matter. I should add Spy as a co-author as well because I stole code from his PR but he already maxed out on street cred.

Ultimately it is a team effort to get it done and a PR process is just an exercise an proving ground if the communication works does the developers listen and respect each other.

PR #21429, opened 2024-05-05
os: fix `os.join_path('/', 'file.txt')` and `os.join_path_single('/', 'file.txt')` (fix #20231)

### LINKED ISSUE #20231: v fails to run and generate binaries from / folder

spytheman:
Fix #20231 .

### DIFF
--- a/vlib/os/dir_expansions_test.v
+++ b/vlib/os/dir_expansions_test.v
@@ -23,7 +23,7 @@ fn test_ensure_folder_is_writable() {
 
 fn test_expand_tilde_to_home() {
 	os.setenv('HOME', '/tmp/home/folder', true)
-	os.setenv('USERPROFILE', '/tmp/home/folder', true)
+	os.setenv('USERPROFILE', r'\tmp\home\folder', true)
 	//
 	home_test := os.join_path(os.home_dir(), 'test', 'tilde', 'expansion')
 	home_expansion_test := os.expand_tilde_to_home(os.join_path('~', 'test', 'tilde',
--- a/vlib/os/join_path_test.v
+++ b/vlib/os/join_path_test.v
@@ -6,6 +6,7 @@ fn test_join_path() {
 		assert os.join_path('', 'f1', 'f2') == r'f1\f2'
 		assert os.join_path('v', '', 'dir') == r'v\dir'
 		assert os.join_path('foo\\bar', '.\\file.txt') == r'foo\bar\file.txt'
+		assert os.join_path('/opt/v', './x') == r'\opt\v\x'
 	} $else {
 		assert os.join_path('v', 'vlib', 'os') == 'v/vlib/os'
 		assert os.join_path('/foo/bar', './file.txt') == '/foo/bar/file.txt'
@@ -13,6 +14,8 @@ fn test_join_path() {
 		assert os.join_path('v', '', 'dir') == 'v/dir'
 		assert os.join_path('/', 'test') == '/test'
 		assert os.join_path('foo/bar', './file.txt') == 'foo/bar/file.txt'
+		assert os.join_path('/opt/v', './x') == '/opt/v/x'
+		assert os.join_path('./a', './b') == './a/b'
 	}
 }
 
@@ -21,9 +24,12 @@ fn test_join_path_single() {
 		assert os.join_path_single('/foo/bar', './file.txt') == r'\foo\bar\file.txt'
 		assert os.join_path_single('/', 'test') == r'\test'
 		assert os.join_path_single('foo\\bar', '.\\file.txt') == r'foo\bar\file.txt'
+		assert os.join_path_single('/opt/v', './x') == r'\opt\v\x'
 	} $else {
 		assert os.join_path_single('/foo/bar', './file.txt') == '/foo/bar/file.txt'
 		assert os.join_path_single('/', 'test') == '/test'
 		assert os.join_path_single('foo/bar', './file.txt') == 'foo/bar/file.txt'
+		assert os.join_path_single('/opt/v', './x') == '/opt/v/x'
+		assert os.join_path_single('./a', './b') == './a/b'
 	}
 }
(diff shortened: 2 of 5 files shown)


hholst80:
@spytheman what are you doing? are you stealing my PR?

spytheman:
> @spytheman what are you doing? are you stealing my PR? Let's not waste double efforts into solving the SAME ISSUE on multiple fronts there are 850+ issues to work on why not work together?

I am fixing it in a different way, that will require way less edits and review changes.

I am sorry, but `os` and join_path especially, is not the place for learning about V, and I am not in the mood of breaking bootstrapping and then fixing it slowly today.

spytheman:
@hholst80, the CI passed for this PR, and imho it will be a bit less breaking, while ensuring the future stability of the behavior of os.join_path and os.join_path_single (which are now more unified in their output, and better tested). 

Do you want me to push the changes here to your PR, so that I can merge them there?

vlib/os/dir_expansions_test.v
```diff
@@ -23,7 +23,7 @@ fn test_ensure_folder_is_writable() {
 
 fn test_expand_tilde_to_home() {
 	os.setenv('HOME', '/tmp/home/folder', true)
-	os.setenv('USERPROFILE', '/tmp/home/folder', true)
+	os.setenv('USERPROFILE', r'\tmp\home\folder', true)
```
hholst80, line 26: If I did this I am fairly sure the word "regression" would come up. ;-)

Anyways, I took this verbatim into my PR.  The extended suite of tests and refactoring is probably a good idea anyways, if we now are going forward with the path normalization.

vlib/os/join_path_test.v
```diff
@@ -0,0 +1,62 @@
...
+		assert os.join_path('./b', '') == './b'
+		assert os.join_path('/b', '') == '/b'
+		assert os.join_path('v', 'vlib', 'os') == 'v/vlib/os'
+		assert os.join_path('/foo/bar', './file.txt') == '/foo/bar/file.txt'
+		assert os.join_path('', 'f1', 'f2') == 'f1/f2'
+		assert os.join_path('v', '', 'dir') == 'v/dir'
+		assert os.join_path('/', 'test') == '/test'
+		assert os.join_path('foo/bar', './file.txt') == 'foo/bar/file.txt'
+		assert os.join_path('/opt/v', './x') == '/opt/v/x'
+		assert os.join_path('./a', './b') == './a/b'
```
hholst80, line 35: why not `a/b` here? I see no reason to keep the initial ./ prefix.

vlib/os/join_path_test.v
```diff
@@ -0,0 +1,62 @@
...
+	$if windows {
+		assert os.join_path('./b', '') == r'.\b'
+		assert os.join_path('/b', '') == r'\b'
+		assert os.join_path('v', 'vlib', 'os') == r'v\vlib\os'
+		assert os.join_path('', 'f1', 'f2') == r'f1\f2'
+		assert os.join_path('v', '', 'dir') == r'v\dir'
+		assert os.join_path('foo\\bar', '.\\file.txt') == r'foo\bar\file.txt'
+		assert os.join_path('/opt/v', './x') == r'\opt\v\x'
+	} $else {
+		assert os.join_path('./b', '') == './b'
```
hholst80, line 26: should be just `b` imho

vlib/os/dir_expansions_test.v
spytheman: `USERPROFILE` is a variable that is only defined on windows, and our `home_dir` only uses it on windows.
On windows, it will contain `\`, not `/` .
The difference between the regression in your PR, and the change here, is that here, the change in the test, makes the test *closer* to the environment, that it tested.
In your PR, the change was done just to make the CI pass with the changes in the PR.

vlib/os/join_path_test.v
spytheman: Good point. It will be simpler without it.

vlib/os/join_path_test.v
spytheman: I agree.

vlib/os/join_path_test.v
spytheman: It will have to be done in a separate focused PR, since it will break asserts like this one `cmd/tools/vcomplete_test.v:229`:
```
   > assert line == sorted_expected[i]
     Left value: sub0/
    Right value: ./sub0/
```
, and is not directly related to the `join_path('/', 'x')` problem in the issue.

spytheman: It is implemented in <https://github.com/spytheman/v/commit/a82dc289ddbe24c7930831dbf9dd5ab1306f578f> for now.

spytheman:
Obsoleted by https://github.com/vlang/v/pull/21494 .

PR #21431, opened 2024-05-05
v: utilize new diff functions p1

ttytm:
Regarding the utilization of the new diff functions. 

Currently, printing diffs has different approaches in the code base. The two with the main differences I try to point out are shown below.

Since things can also be improved and made more consistent when integrating the new functions, this PR takes two places that allow us to decide on an updated that might be more suited then the other. Then it could serve as preferred way when updating the new diff functions in other places.

- Approach 1. (visible in the updated `vlib/v/compiler_errors_test.v`)
  Prints diffs **OR** expected and found.
- Approach 2. (visible in the updated `slow_tests/inout/compiler_test.v`)
  Prints first expected then found **AND** then differences. This currently used more frequently. 
  
  Atm. the PR updates slow_tests/inout/compiler_test.v from A2 to A1. Resulting in:
  | Current | Updated |
  | - | - |
  | Screenshot_20240505_134055 | Screenshot_20240505_134927 |

### DIFF
--- a/vlib/v/compiler_errors_test.v
+++ b/vlib/v/compiler_errors_test.v
@@ -1,5 +1,4 @@
 import os
-import rand
 import term
 import v.util.diff
 import v.util.vtest
@@ -409,10 +408,10 @@ fn chunka(s []u8, chunk_size int) string {
 
 fn diff_content(expected string, found string) {
 	println(term.bold(term.yellow('diff: ')))
-	if diff_cmd := diff.find_working_diff_command() {
-		println(diff.color_compare_strings(diff_cmd, rand.ulid(), expected, found))
+	if diff_ := diff.compare_text(expected, found) {
+		println(diff_)
 	} else {
-		println('>>>> could not find a working diff command; dumping bytes instead...')
+		println('>>>> `${err}`; dumping bytes instead...')
 		println('expected bytes:\n${chunka(expected.bytes(), 25)}')
 		println('   found bytes:\n${chunka(found.bytes(), 25)}')
 		println('============')
--- a/vlib/v/slow_tests/inout/compiler_test.v
+++ b/vlib/v/slow_tests/inout/compiler_test.v
@@ -95,16 +95,18 @@ fn test_all() {
 		}
 		if expected != found {
 			println(term.red('FAIL'))
-			println(term.header('expected:', '-'))
-			println(expected)
-			println(term.header('found:', '-'))
-			println(found)
-			if diff_cmd != '' {
+			if diff_ := diff.compare_text(expected, found) {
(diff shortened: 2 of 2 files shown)


vlib/v/compiler_errors_test.v
```diff
@@ -409,10 +408,10 @@ fn chunka(s []u8, chunk_size int) string {
 
 fn diff_content(expected string, found string) {
 	println(term.bold(term.yellow('diff: ')))
-	if diff_cmd := diff.find_working_diff_command() {
-		println(diff.color_compare_strings(diff_cmd, rand.ulid(), expected, found))
+	if diff_ := diff.compare_text(expected, found) {
+		println(diff_)
 	} else {
-		println('>>>> could not find a working diff command; dumping bytes instead...')
+		println('>>>> `${err}`; dumping bytes instead...')
```
ttytm, line 414: To give a quick testable local test example, regarding the usage of `err`:

```v
// file.v
import v.util.diff

if diff_ := diff.compare_text('abcd', 'cde') {
	println(diff_)
} else {
	println('>>>> `${err}`; dumping bytes instead...')
}
```
```sh
❯ VDIFF_CMD='zzz' v run file.v
>>>> `error: failed to find comparison command `zzz``; dumping bytes instead...
```

vlib/v/slow_tests/inout/compiler_test.v
```diff
@@ -95,16 +95,18 @@ fn test_all() {
...
+				total_errors++
 			} else {
-				println(term.h_divider('-'))
+				println(err)
+				println(term.header('expected:', '-'))
+				println(expected)
+				println(term.header('found:', '-'))
+				println(found)
 			}
+			println(term.h_divider('-'))
```
ttytm, line 109: Below, the alternative update when keeping `expected`, `found` and `differences` (A2 mentioned in the PR description)

```v
println(term.header('expected:', '-'))
println(expected)
println(term.header('found:', '-'))
println(found)
println(term.header('difference:', '-'))
println(diff.compare_text(expected, found) or { err.msg() })
println(term.h_divider('-'))
```

JalonSolov:
Personally, I prefer brevity.  If everything expected was found, why waste time/space reporting the 2 lists?

Perhaps it is useful sometimes.  For those times, a `verbose` option would be handy, but I don't like verbosity as the default.

ttytm:
@JalonSolov I'm in the same boat. When the expected and found are plainly printed, it's easy to lose track. Especially when they get longer.

Having it less verbose and straight to the point when there is a working diff tool should help to spot what is important more quickly.

spytheman (approved):

PR #21432, opened 2024-05-05
docs: fix links

xenoken:
Fixes the links on the new documentation website.

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -40,7 +40,7 @@ by using the V's built-in self-updater.
 To do so, run the command `v up`.
 
 ## Packaging V for distribution
-See the [notes on how to prepare a package for V](packaging_v_for_distributions.md) .
+See the [notes on how to prepare a package for V](https://github.com/vlang/v/blob/master/doc/packaging_v_for_distributions.md) .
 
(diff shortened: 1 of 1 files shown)


doc/docs.md
```diff
@@ -251,7 +251,8 @@ In this case `main` doesn't return anything, so there is no return type.
 
 As in many other languages (such as C, Go, and Rust), `main` is the entry point of your program.
 
-[`println`](#println) is one of the few [built-in functions](#builtin-functions).
+[`println`](builtin-functions.html#println) is one of the few 
+[built-in functions](builtin-functions.html).
```
spytheman, line 255: This will not work in `docs.md`. There is not a `builtin-functions.html` file in the main repo. That file is only present in the generated output from https://docs.vlang.io/ .

spytheman:
> @JalonSolov Yes, they do, but when you click on links in the 'docs.md' file, it will redirect you to the website.

No, they do not.

I can not merge this. The docs generator will have to become more intelligent.

xenoken:
@spytheman I see. Is it still correct to let links point to the website from the docs.md file?

PR #21437, opened 2024-05-05
vet, parser, scanner: remove vetting for spaces after / before parens

ttytm:
Instead of moving the scanner errors for spaces after / before parenthesis to vvet, this PR would remove it as it might be something merely handled by v fmt. 

The reason is that it seems like overkill for vet. It doesn't point out other similar formatting errors.
If the current errors like `` Looks like you are adding a space before `)` `` should be vet errors, I'm gonna update accordingly. 

I'm also noticing that the line numbers are off that it would currently.


spytheman:
The CI failure is not related. Good work.

vlib/v/parser/parser.v
```diff
@@ -257,7 +256,7 @@ pub fn parse_file(path string, mut table ast.Table, comments_mode scanner.Commen
 	return res
 }
 
-pub fn parse_vet_file(path string, mut table_ ast.Table, pref_ &pref.Preferences) (&ast.File, []vet.Error) {
+pub fn parse_vet_file(path string, mut table_ ast.Table, pref_ &pref.Preferences) &ast.File {
```
spytheman, line 259: Can we move `parse_vet_file` completely out of parser and into vvet.v, or is it better for that to be done in another PR?

ttytm: Yes, planned to do so. However it is preferred.

spytheman: Both can work, and if you have already plans for it, I will merge it as it is now, to save some time on the CI.

PR #21438, opened 2024-05-05
v: add @[_linker_section] for global variable

felipensp:
Add support to specific the section for global var. (function is already supported)


vlib/v/checker/checker.v
```diff
@@ -2240,6 +2240,14 @@ fn (mut c Checker) branch_stmt(node ast.BranchStmt) {
 }
 
 fn (mut c Checker) global_decl(mut node ast.GlobalDecl) {
+	required_args_attr := ['_linker_section']
```
medvednikov, line 2243: this is going to allocate an array on each global_decl() call
better to have it in a global const

medvednikov: I'll fix it

PR #21443, opened 2024-05-06
v.util: make all available diff tools accessible, extend tests

ttytm:
Providing information in the public constant should further extend developer possibilities. 

A test for this is added in a way that makes the tests for the full module more regression safe:

Ref., regarding the availability of `diff` as tool on all CI runners:
Screenshot_20240506_120540


ttytm:
Thank you @spytheman. I'm always happy to work alongside a master.

PR #21446, opened 2024-05-06
ci: merge docker_alpine and docker_ubuntu workflows in `docker_ci.yml`

ttytm:
Those two relatives are nearly twins, they might get by with living together in a common workflow. It could result in less verbosity easier maintenance.


.github/workflows/docker_ci.yml
```diff
@@ -44,16 +44,36 @@ jobs:
...
       - name: Build V
         run: CC=gcc make
-
       - name: All code is formatted
         run: VJOBS=1 ./v test-cleancode
-
       - name: Run only essential tests
         run: VTEST_JUST_ESSENTIAL=1 ./v test-self
+
+  ubuntu-musl:
```
spytheman, line 56:
```suggestion
  ubuntu-docker-musl:
```

spytheman: The longer job name makes it easier to search and identify.
The normal runner uses `ubuntu` too, and that environment is different than the the one from `thevlang/vlang:ubuntu-build` .

.github/workflows/docker_ci.yml
```diff
@@ -1,41 +1,41 @@
...
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref == 'refs/heads/master' && github.sha || github.ref }}
   cancel-in-progress: true
 
+env:
+  V_CI_MUSL: 1
+
 jobs:
-  alpine-docker-musl-gcc:
+  alpine-musl-gcc:
```
spytheman, line 27: There is no possibility for confusion here, but imho `alpine-docker-musl` will fit better with `ubuntu-docker-musl` .

PR #21448, opened 2024-05-07
ci: do not allow new notices for `v build-tools` and `v build-examples` on the Linux CI jobs

spytheman:


spytheman:
The vinix failure is fixed on master in 5c2466c .

PR #21451, opened 2024-05-07
tools: improve environment detection when creating symlinks

ttytm:
The PR tries to bring goodies from all worlds for the `symlink` tool. 

- Simplifies maintenance.
- Makes the tool safer against regressions.
- Easier general usage for `v symlink` in CI runners (the `-githubci` flag might not be obvious, when using V in an external workflow).
- Relieves / frees runners in our current CI (as separate tests are no longer required).

**The diff may seem complicated but the change is actually simple:**

- `-githubci` flag becomes obsolete.
  1. It was already required for Windows only. Therefore handling of symlink_github was moved to the windows code.
  2. An improvement made alongside is that instead of adding only the current directory to the GITHUB_PATH, `new_sys_env_path` is added, which is also used on regular systems. **This brings the Windows test closer to a regular scenario outside CI.**
  3. This was automated when it is used in a Windows CI runner.

- To not introduce a breaking change, a warning message was added when the `-githubci` flag is used in workflows.

  Screenshot_20240507_142113
  Sample run with warning: https://github.com/ttytm/v/actions/runs/8985093496

### DIFF
--- a/.github/workflows/symlink_ci.yml
+++ b/.github/workflows/symlink_ci.yml
@@ -1,15 +1,15 @@
-name: V Symlink Works
+name: Symlink CI
 
 on:
   workflow_dispatch:
   push:
     paths:
       - 'cmd/tools/vsymlink/**.v'
-      - '.github/workflows/check_symlink_works.yml'
+      - '.github/workflows/symlink_ci.yml'
   pull_request:
     paths:
       - 'cmd/tools/vsymlink/**.v'
-      - '.github/workflows/check_symlink_works.yml'
+      - '.github/workflows/symlink_ci.yml'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name == 'master' && github.sha || github.ref_name }}
--- a/.github/workflows/c2v_ci.yml
+++ b/.github/workflows/c2v_ci.yml
@@ -37,7 +37,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build V
-        run: make && ./v symlink -githubci
+        run: make -j4 && ./v symlink
       - name: Install C2V
         run: |
           .github/workflows/retry.sh v install --git https://github.com/vlang/c2v
@@ -72,7 +72,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Build V
-        run: make && ./v symlink -githubci
+        run: make -j4 && ./v symlink
       - name: Install C2V
         run: |
           .github/workflows/retry.sh v install --git https://github.com/vlang/c2v
--- a/.github/workflows/check_symlink_works.yml
+++ b/.github/workflows/check_symlink_works.yml
@@ -16,35 +16,11 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test-sudo:
+  symlink:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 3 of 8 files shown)


.github/workflows/symlink_ci.yml
```diff
@@ -1,48 +1,26 @@
...
-          v version
-          echo 'println(123)' > hi.v
-          v run hi.v
-
-  test-githubci:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-20.04, macos-13, windows-2019]
+        os: [ubuntu-20.04, macos-12, windows-2019]
```
spytheman, line 23: Why are you deleting the sudo test?

What will happen on the CIs that already do use `sudo ./v symlink` ?

spytheman:
> It was already required for Windows only.

It was needed on other environments too, to avoid the need for `sudo ./v symlink`.
It should stay just as it is for all environments, since its functionality is independant on the runner, but only on whether it is a github one or not (i.e. whether it supports `GITHUB_PATH`).

spytheman:
> Easier general usage for v symlink in CI runners (the -githubci flag might not be obvious, when using V in an external workflow).

That is not a reason to remove it, it is documented in `v help symlink` .

ttytm:
@spytheman the passing CI proves that sudo was not required here for github runners.

If you prefer to also have test that run the symlink with sudo they can be kept.
But this is an improvement without downsides. I think you closed this too early.

ttytm:
If I missed something about the other runners the updated can be extended to cover all cases and be still simpler bottom line.

> That is not a reason to remove it, it is documented in v help symlink .

Yep, but this is also the weakest point. We can strike that one and talk about the others.

.github/workflows/symlink_ci.yml
ttytm: They will still work, because for them nothing changed.

spytheman:
> If you prefer to also have test that run the symlink with sudo they can be kept.

I do. That test was added for a reason. Removing it, *in the same PR*, that replaces `-githubci`, with just a warning, while moving its functionality, to the platform specific implementation for just windows, while that functionality is not at all related to windows, is bad.

spytheman:
I did not like at all the removal of the CI test. Especially in combination with moving platform independent functionality to a platform dependent file.

As it is on master, `v symlink -githubci` works the same, everywhere.
It is one less thing to worry about - it integrates cleanly, with what Github provides.

Just `v symlink` depends on `/usr/local/bin` being writable, and it does not work on Windows.

ttytm:
`v symlink` calls `setup_symlink()` in either `_nix.v` or `_windows.v`. In the latter `/usr/local/bin` has no business. 

But trying to create a theoretical map of what is happening is harder than having the concrete changes. I can make the changes in steps, then I think there should be less worries.

spytheman:
Ok, do it in steps, as long as `v symlink -githubci` continues to work in the cross platform way it does, even with a warning. CIs that already do use it, should not break suddenly.

JalonSolov:
Once everything is converted, `v symlink` can just print out a deprecation message for the `-githubci` option for a while, _then_ remove it.

JalonSolov:
Smaller PRs (when possible) are always better than large one.  Fewer changes means fewer things to review, easier to review, faster to review, etc., etc.

PR #21453, opened 2024-05-07
tools: make `./v symlink` work platform independent in CI

ttytm:
Splits up changes to make `./v symlink` work platform independent in CI.

The PR:
- Moves handling of github symlink to the `vsymlink_windows.c.v` code. On unix-like systems regular symlinking will be used, whether or not `-githubci` was passed.
- An improvement made alongside is that instead of adding `os.getwd()` to the GITHUB_PATH, `new_sys_env_path` is used. It is the path used on a regular windows systems -> Brings tests in Windows CI closer to a regular scenario, makes them more reliable.
- Now `./v symlink` should already work platform independent. The `-githubci` flag is not required anymore. Cases with `sudo` and the `-githubci` flag are still covered by the current, unchanged CI jobs. They should continue to work and there are no plans in changing this inside the PRs that are implementing this enhancement. Only adding an info that the flag is not required anymore.

### DIFF
--- a/.github/workflows/check_symlink_works.yml
+++ b/.github/workflows/check_symlink_works.yml
@@ -16,18 +16,20 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  test-sudo:
+  test:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, macos-13]
+        os: [ubuntu-20.04, macos-13, windows-2019]
       fail-fast: false
     steps:
       - uses: actions/checkout@v4
-      - name: Build V
-        run: make -j4
-      - name: Symlink
-        run: sudo ./v symlink
+      - name: Build and symlink (Windows)
+        if: runner.os == 'Windows'
+        run: ./make.bat && ./v symlink
+      - name: Build and symlink
+        if: runner.os != 'Windows'
+        run: make -j4 && sudo ./v symlink
       - name: Check if V is usable
         run: |
           pwd
--- a/cmd/tools/vsymlink/vsymlink.v
+++ b/cmd/tools/vsymlink/vsymlink.v
@@ -4,34 +4,9 @@ const vexe = os.real_path(os.getenv_opt('VEXE') or { @VEXE })
 
 fn main() {
 	at_exit(|| os.rmdir_all(os.vtmp_dir()) or {}) or {}
-
 	if os.args.len > 3 {
-		print('usage: v symlink [OPTIONS]')
+		println('usage: v symlink')
 		exit(1)
 	}
-
-	if '-githubci' in os.args {
-		setup_symlink_github()
-	} else {
-		setup_symlink()
-	}
-}
-
-fn setup_symlink_github() {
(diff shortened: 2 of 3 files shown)


spytheman:
`-githubci` should *always* invoke `setup_symlink_github`, and that `setup_symlink_github` should stay just as it is for now, in the `vsymlink.v`.

spytheman:
`new_sys_env_path` can be added in addition to the rest, if it is needed on windows.

spytheman:
> Now ./v symlink should already work platform independent. The -githubci flag is not required anymore. Cases with sudo and the -githubci flag are still covered by the current, unchanged CI jobs. They should continue to work and there are no plans in changing this inside the PRs that are implementing this enhancement. Only adding an info that the flag is not required anymore.

That will be nice, if true. I am not confident in it.
I want the behavior of `v symlink -githubci` to stay the same, until it is no longer needed. The point of doing things in steps, instead of in one big change, is to minimize/spread the risk of breakages of CIs that already do use `v symlink -githubci` . It is not just to make the diff smaller, while containing the exact same modifications, to which I already objected once.

ttytm:
The goal is to make the it work without the flag. And it works without the flag. 
It should also be made sure that it continue to works with the flag. This is prooven by the passed CI jobs that still use `-githubci` flags even when the flags don't don anything prove that it is working. 

As it is a simple change idk what is wrong. Trying to understand the problem if you would share what is not tested here, it would help. 

In the theoretical case of scenarios where nix systems would require V in the `GITHUB_PATH` it could still be done automatically - similar to how the updates does it in windows. This automation could be made in a testable way.

ttytm:
Too small steps for me, considering the relation to time, chance of breakage and easy possibilities to fix. The request of making the change of removing the `-githubci` flag from other runners to have this tested makes the changes in this PR eventually untested, which makes it for me personally illogical.

spytheman:
> This is proven to work by the passed CI jobs that still use -githubci flags even when the flags don't do anything.

Many of the CI jobs do use `./v` inside them, which will work even without any `v symlink` at all, instead of `v`, which will break, if v is not on the PATH.

They are written like that, in part because of how flaky `v symlink` and `v symlink -githubci` were historically.

spytheman:
> From all available infront of us, regular symlinking on nix systems is sufficient.

Symlinking to `/usr/local/bin`, requires that you have write permissions to that folder.

Writing to a file, whose location is described in `GITHUB_PATH`, when it is set, as described by github's docs, does not require any additional write permissions for your runner, and will continue to work, no matter what happens with their runners, and images.

spytheman:
The current state, after e012ca8 is fine.

ttytm:
> Symlinking to /usr/local/bin, requires that you have write permissions to that folder.

That is true but of relevance for local symlinking. In GitHub CI the folder should have write permissions in the same way as it is not required to add credentials when running commands as sudo. When those default permissions are changed it's a custom case where the one doing those permission changes will need to take care of it. 

I'll try that patience thing when making the smallest possible steps so that data can better cover things than words can.

PR #21456, opened 2024-05-07
cli: make program outputs using the cli module testable in `cli/testdata`

ttytm:
Changes in regard of an issue I'm encountering with the `cli` module and like to fix and test cover. To make development (especially testing) more accessible, this PR adds testability of program outputs that use the cli module. 

Changes per commit:
- Makes the current test in `cli_test.v` namely `fn test_long_description` also testable as executable in `vlib/cli/testdata/long_description.vv` / `../long_description.out`.
- Moves the cli modue tests that currently reside in `vlib/v/slow_tests/inout/` into `cli/testdata` (all the cli tests take usually less than 300ms)

Testing locally after checking out the PR and making a change to one of the test / out files to ensure the tests work as they should might help to simplify reviewing.


spytheman (approved):
Good work. It makes the module more self contained.

The slowdown is not too bad, even on my slow i3:
image

spytheman:
The change detection also works, and produces easily readable results:
image

vlib/cli/cli_test.v
```diff
@@ -19,10 +19,11 @@ fn test_cli_programs() {
 			has_err = true
 			continue
 		}
-		expected_out := os.read_file(out_path)!
+		mut expected_out := os.read_file(out_path)!
 		mut test_out := os.execute('${vexe} run ${test}').output
 		$if windows {
-			test_out = test_out.replace('\r\n', '\n').trim_space()
+			expected_out = expected_out.replace('\r\n', '\n')
+			test_out = test_out.replace('\r\n', '\n')
```
spytheman, line 26: imho this should be always done, even on unix

spytheman: The reason, is that having separate paths here, in the test code, that otherwise is not concerned with platform differences, may make reproducing bugs a bit harder, and requiring access to a windows machine/vm, or the reverse, which is not always convenient.

ttytm: Gonna update when adding the new tests

PR #21458, opened 2024-05-07
v.util: remove fast path in `diff.compare_text`

ttytm:
The equality check of the strings that should be compared is done in most cases before the function is called in the V code base. It can be expected that it will be the same for external projects if they use the module.

Due to this, what was thought as fast-path can become a slow-down in cases where the compiler is not able to optimize the double check. Therefore it is removed and the function merely cares about returning the result of the comparison result.


spytheman (approved):
It will only matter for long texts, that are nearly identical.

However simplifying code where possible, is always good.

Excellent work.

PR #21459, opened 2024-05-07
ci: add a linter for the .yml workflow files

ttytm:
- Adds a formatting check to help with keeping a consistent code style. 
- Resolves non-conformity formatting.

It would only run on changes to .yml files and takes a few seconds.


spytheman (approved):
Excellent.

I do hate working on .yml files. 

At least now, I will hate them consistently.

PR #21461, opened 2024-05-07
math/unsigned: permit _ separators in Uint128 decimal strings

kimshrier:
Permit '_' separator characters in strings that can be converted to Uint128 values.

This is like the other string -> numeric value conversions for the builtin number types.

I was thinking of doing the same for Uint256 if this change is acceptable for Uint128.


kimshrier:
I was following the precedent set in the line following my code addition where `'0'.bytes()[0]` was used.  I am more than willing to change my code in order to be both more efficient and more V like.

vlib/math/unsigned/uint128.v
```diff
@@ -423,10 +423,16 @@ pub fn uint128_new(lo u64, hi u64) Uint128 {
...
 	mut res := unsigned.uint128_zero
 	for b_ in value.bytes() {
+		// allow _ as a separator in decimal strings
+		if b_ == '_'.bytes()[0] {
+			continue
+		}
+
 		b := b_ - '0'.bytes()[0]
 		if b > 9 {
 			return error('invalid character "${b}"')
```
JalonSolov, line 438:
```suggestion
                match b_ {
                        // allow _ as a separator in decimal strings
                        `_` { continue }
                        `0`...`9` { }
                        else { return error('invalid character "${b_}"') }
```

kimshrier: I made a different change.  I hoisted the code for getting the value of an _ character and just indexed a string to get it.  I made the same change for the character being converted, so as to avoid calling bytes().   I also improved the validity check to make sure that only the digits 0-9 were legal since it is possible for b to be a negative value.  Only in the event of a non-digit do we check to see if it is an _ and permit that.

vlib/math/unsigned/uint128.v
```diff
@@ -423,12 +423,20 @@ pub fn uint128_new(lo u64, hi u64) Uint128 {
...
+// unint_from_dec_str returns an error or new Uint128 from given string.
+// The `_` character is allowed as a separator.
 pub fn uint128_from_dec_str(value string) !Uint128 {
 	mut res := unsigned.uint128_zero
+	underscore := '_'[0]
+
 	for b_ in value.bytes() {
-		b := b_ - '0'.bytes()[0]
-		if b > 9 {
+		b := b_ - '0'[0]
```
spytheman, line 433:
```suggestion
		b := b_ - `0`
```

vlib/math/unsigned/uint128.v
```diff
@@ -423,12 +423,20 @@ pub fn uint128_new(lo u64, hi u64) Uint128 {
 	return Uint128{lo, hi}
 }
 
-// unint_from_dec_str returns an error or new Uint128 from given string
+// unint_from_dec_str returns an error or new Uint128 from given string.
+// The `_` character is allowed as a separator.
 pub fn uint128_from_dec_str(value string) !Uint128 {
 	mut res := unsigned.uint128_zero
+	underscore := '_'[0]
```
spytheman, line 430:
```suggestion
	underscore := `_`
```

vlib/math/unsigned/uint128.v
```diff
@@ -423,12 +423,20 @@ pub fn uint128_new(lo u64, hi u64) Uint128 {
...
 }
 
-// unint_from_dec_str returns an error or new Uint128 from given string
+// unint_from_dec_str returns an error or new Uint128 from given string.
+// The `_` character is allowed as a separator.
 pub fn uint128_from_dec_str(value string) !Uint128 {
 	mut res := unsigned.uint128_zero
+	underscore := `_`
+
 	for b_ in value.bytes() {
```
spytheman, line 432:
```suggestion
	for b_ in value {
```

vlib/math/unsigned/uint128.v
```diff
@@ -423,12 +423,20 @@ pub fn uint128_new(lo u64, hi u64) Uint128 {
...
 pub fn uint128_from_dec_str(value string) !Uint128 {
 	mut res := unsigned.uint128_zero
-	for b_ in value.bytes() {
-		b := b_ - '0'.bytes()[0]
-		if b > 9 {
+	underscore := `_`
+
+	for b_ in value {
+		b := b_ - `0`
+		if b < 0 || b > 9 {
```
spytheman, line 434:
```suggestion
		if b > 9 {
```
(since `b` will be u8, it can not get negative; for values of `b_` < ASCII `0`, it will wrap to values > 200)

vlib/math/unsigned/uint128_test.v
```diff
@@ -97,3 +97,74 @@ fn test_leading_zeros() {
 		assert zeros == tc.zeros
 	}
 }
+
+fn test_separators() {
+	// numbers of varying lengths and a random
+	// scattering of '_' throughout.
+	test_strings := [
+		'_0',
```
spytheman, line 105: Testing just `_` and `__` may be also interesting as an edge case.

PR #21466, opened 2024-05-08
ci: update symlink ci, extend tested cases

ttytm:
Regarding: 6cda618ead43bfb7ad4bb9ad7ef24aaf2dfbf6db

- Extends tested cases 
- Second is an incremental refactor to the related file and only concerns updating the workflow name to what is most consistently used for other workflows.

### DIFF
--- a/.github/workflows/symlink_ci.yml
+++ b/.github/workflows/symlink_ci.yml
@@ -1,15 +1,15 @@
-name: V Symlink Works
+name: Symlink CI
 
 on:
   workflow_dispatch:
   push:
     paths:
       - 'cmd/tools/vsymlink/**.v'
-      - '.github/workflows/check_symlink_works.yml'
+      - '.github/workflows/symlink_ci.yml'
   pull_request:
     paths:
       - 'cmd/tools/vsymlink/**.v'
-      - '.github/workflows/check_symlink_works.yml'
+      - '.github/workflows/symlink_ci.yml'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name == 'master' && github.sha || github.ref_name }}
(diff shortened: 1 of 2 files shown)


.github/workflows/symlink_ci.yml
```diff
@@ -0,0 +1,76 @@
...
+        run: |
+          sudo ./v symlink
+          cd /tmp/ && v version
+          cd ~ && v version
+          echo 'println(123)' > hi.v
+          v run hi.v
+      - name: Unlink
+        run: |
+          rm $(which v)
+          v --version && exit 1 || exit 0
```
ttytm, line 50: Ref. ensuring that unlinking worked worked. 
Trying to execute v --version while style linked.

Screenshot_20240508_083530

spytheman:
The changes seem good to me, but please do not rename files, in the same PR that changes their contents a lot. That needlessly necessitates checking out the changes locally.

ttytm:
K understood. 
Reason was that I thought some changes might have trouble getting through on their own, so sneaking them in as an "incremental refactor" could be a way. I choose them better.

spytheman:
I usually do not care much about the individual commits in a PR (since they often contain trivial stuff like fixing typos, quoting styles, word wrapping etc, to accommodate the different CI jobs and linters).

It frequently saves time, from my perspective, when possible, to review only the final diff/result, that is going to be merged, using Github's "Files changed" pane, after the CI passes, and a PR is no longer a draft.

The negative of that, is that my feedback and reviews are often delayed, by the time the CI finishes running (which is now minimized, in many cases, thanks to your efforts), and I may be an ass, from the perspective of contributors, who did care a lot about the individual commits.

Putting functional changes in PRs, that at the same time mention the word 'refactor', and that have file renames, is definitely something to be avoided.

PR #21469, opened 2024-05-08
cli: fix default flags when their command equivalents are disabled

ttytm:
Fixes default flags not working as expected when their commands are disabled.

Example:
```v
// foo.v
import cli { Command, CommandFlag }
import os

fn main() {
	mut cmd := Command{
		name: 'foo'
		version: '0.1.0'
		posix_mode: true
		defaults: struct {
			version: CommandFlag{false, true}
			help: CommandFlag{false, true}
		}
	}
	cmd.parse(os.args)
}
```
Current:
```sh
❯ v run foo.v --version
# no output
```

Updated:
```sh
❯ v run foo.v --version
foo version 0.1.0
```

### DIFF
--- a/vlib/cli/version.v
+++ b/vlib/cli/version.v
@@ -19,29 +19,22 @@ fn version_cmd() Command {
 }
 
 fn print_version_for_command(cmd Command) ! {
-	version_cmd := if cmd.args.len > 0 {
-		mut cmd_ := &Command{}
-		for arg in cmd.args {
-			mut found := false
-			for sub_cmd in cmd.commands {
-				if sub_cmd.name == arg {
-					cmd_ = unsafe { &sub_cmd }
-					found = true
-					break
-				}
-			}
-			if !found {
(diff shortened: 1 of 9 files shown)


vlib/cli/version.v
```diff
@@ -14,12 +14,34 @@ fn version_cmd() Command {
...
 		description: 'Prints version information.'
-		execute: version_func
+		execute: print_version_for_command
 	}
 }
 
-fn version_func(version_cmd Command) ! {
-	cmd := version_cmd.parent
-	version := '${cmd.name} version ${cmd.version}'
+fn print_version_for_command(cmd Command) ! {
```
ttytm, line 21: The fix is based on the other default command:
https://github.com/vlang/v/blob/2e9b4cf42c7cedd4b14cbfb86a953924852b6f1d/vlib/cli/help.v#L30

spytheman (approved):
Excellent work.

PR #21478, opened 2024-05-09
vdoc/vdoc_file_test.v: fix test_out_path

smalltalkman:
It's not clear why the error doesn't appear in ci, but the local test keeps appearing

```
# ./v run cmd/tools/vdoc/tests/vdoc_file_test.v
D:/gym/test/mingw-w64-packages/mingw-w64-v/src/v-weekly.2024.19/cmd/tools/vdoc/tests/testdata/basic/basic.v OK
D:/gym/test/mingw-w64-packages/mingw-w64-v/src/v-weekly.2024.19/cmd/tools/vdoc/tests/testdata/comments/main.v OK
D:/gym/test/mingw-w64-packages/mingw-w64-v/src/v-weekly.2024.19/cmd/tools/vdoc/tests/testdata/multiline/main.v OK
D:/gym/test/mingw-w64-packages/mingw-w64-v/src/v-weekly.2024.19/cmd/tools/vdoc/tests/testdata/newlines/main.v OK
D:/gym/test/mingw-w64-packages/mingw-w64-v/src/v-weekly.2024.19/cmd/tools/vdoc/tests/testdata/output_formats/main.v OK
D:/gym/test/mingw-w64-packages/mingw-w64-v/src/v-weekly.2024.19/cmd/tools/vdoc/tests/testdata/readme_in_project_root/src/main.v OK
D:/gym/test/mingw-w64-packages/mingw-w64-v/src/v-weekly.2024.19/cmd/tools/vdoc/tests/testdata/run_examples_bad/main.v OK
D:/gym/test/mingw-w64-packages/mingw-w64-v/src/v-weekly.2024.19/cmd/tools/vdoc/tests/testdata/run_examples_good/main.v OK
D:/gym/test/mingw-w64-packages/mingw-w64-v/src/v-weekly.2024.19/cmd/tools/vdoc/tests/testdata/unsorted/main.v OK
cmd/tools/vdoc/tests/vdoc_file_test.v:105: ✗ fn test_out_path
    assert os.exists(output_path), output_path
        Message: D:\gym\test\mingw-w64-packages\mingw-w64-v\tmp\mingw-w64-i686-v-weekly.2024.19-1\v_0\vdoc_test_01HXE2FCMGCA3H48WQYMRGXX6H\bitfield\_docs\bitfield.html

```
In fact the generated file exists in the directory `os.join_path(vroot, 'vlib', small_pure_v_vlib_module)`, not `os.join_path(test_path, small_pure_v_vlib_module)`.

However, due to `os.cp_all(mod_path, test_mod_path, true) or {}`, the generated file will appear under `os.join_path(test_path, small_pure_v_vlib_module)` during the second test.

### DIFF
--- a/cmd/tools/vdoc/vdoc_file_test.v
+++ b/cmd/tools/vdoc/vdoc_file_test.v
@@ -98,11 +98,12 @@ fn test_out_path() {
 		os.rmdir_all(test_path) or {}
 	}
 	os.chdir(test_path)!
-	os.cp_all(os.join_path(vroot, 'vlib', small_pure_v_vlib_module), test_mod_path, true) or {}
+	mod_path := os.join_path(vroot, 'vlib', small_pure_v_vlib_module)
+	os.cp_all(mod_path, test_mod_path, true) or {}
 
 	// Relative input with default output path.
 	os.execute_opt('${vexe} doc -f html -m ${small_pure_v_vlib_module}')!
-	output_path := os.join_path(test_mod_path, '_docs', '${small_pure_v_vlib_module}.html')
+	output_path := os.join_path(mod_path, '_docs', '${small_pure_v_vlib_module}.html')
 	assert os.exists(output_path), output_path
 
 	// Custom out path (no `_docs` subdir).


spytheman:
Does the problem persist after 8eaaba6 , in which the path to the test is now: `cmd/tools/vdoc/vdoc_file_test.v` ?

spytheman:
For me, that specific test file, fails sporadically locally too, but so far I could not find a reliable way to reproduce it 🤔 . Running it a second time succeeds every time.

smalltalkman:
> For me, that specific test file, fails sporadically locally too (on Ubuntu 20.04), but so far I could not find a reliable way to reproduce it 🤔 . 

When I tested locally, I only needed to delete `vlib\bitfield\_docs` to reproduce.

>  Running it a second time succeeds every time.

The reason for the second success is that the code `os.cp_all(mod_path, test_mod_path, true) or {}` copies the last generated file(not files generated this time) to the temporary directory, so on the surface it seems that the test is successful.


> Does the problem persist after [8eaaba6](https://github.com/vlang/v/commit/8eaaba6094456bb21833fddf20d22514e7dadcb2) , in which the path to the test is now: `cmd/tools/vdoc/vdoc_file_test.v` ?

For some reason, I'm banned from using git, so I can't test the latest code now, maybe wait until next week when I can test the new tags.

spytheman:
> When I tested locally, I only needed to delete vlib\bitfield\_docs to reproduce.

That indeed works for me too. Thank you. I assumed wrongly that my repo is clean, but it was not, since `_docs/` was ignored 🤦🏻‍♂️ . I should have known better.

spytheman:
Imho it is good to merge as it is.
I've tested it locally, and it does solve the problem for me.

Do you want to change it for some reason?

smalltalkman:
I didn't figure out why the code needs to be copied to a temporary folder when testing. 
In addition, should _docs be deleted after assert is successful to prevent future errors?

spytheman:
> I didn't figure out why the code needs to be copied to a temporary folder when testing. 

It tests several modes of operation.

The first is `v doc -m -f html bitfield`, which generates documentation for the vlib module `bitfield` in the default location (which happens to be inside in your `{VROOT}/vlib/bitfield` folder, even if your working folder is somewhere else). This is a legacy mode, that should be supported until we have something better. The test just verifies that it does not regress.

The second is `v doc -m -f html -o result/ bitfield`, which generates the same, but in the local folder `result/` .

The rest is tests, for whether the tool works for a local module `bitfield` (i.e. not the one in vlib). It could have written its own module code, but the easiest way at the time of writing the test, was to just copy the module.

> In addition, should _docs be deleted after assert is successful to prevent future errors?

The one in vlib/bitfield/ should. The one in the temp folder, may or may not, since the entire folder is deleted by `v test` after a successfully run.

smalltalkman:
Thank you for your detailed explanation. I'm very sorry, I don't have enough time to look at the code in detail recently.

Issue #21490, opened 2024-05-11
auto deref not working


Eliyaan:
### Describe the bug

V forgot to dereference the `mut`  from what I can understand.

### Reproduction Steps

This does not work
```v
interface Ennemi {
}

struct Map {
    ennemis       []Ennemi
}

fn main() {
    mut maap := Map{}

    for mut ennemi in maap.ennemis {
        maap.ennemis.index(ennemi)
    }
}
```
While this works :
```v
interface Ennemi {
}

struct Map {
    ennemis       []Ennemi
}

fn main() {
    mut maap := Map{}
    
    for mut ennemi in maap.ennemis {
        maap.ennemis.index(*ennemi)
    }
}
```

### Expected Behavior

That V dereferences the mut to give it to .index()

### Current Behavior

```
/tmp/v_1000/../../../../../../home/nopana/projects/Vely/test.v:12: error: pointer expected
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
```

### V version

V 0.4.5 0e2b604

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.5 294f7e4.0e2b604
OS: linux, Linux version 6.6.13-200.fc39.x86_64 (mockbuild@72f11b2996ed4699b0f705186172808f) (gcc (GCC) 13.2.1 20231205 (Red Hat 13.2.1-6), GNU ld version 2.40-13.fc39) #1 SMP PREEMPT_DYNAMIC Sat Jan 20 18:03:28 UTC 2024
Processor: 4 cpus, 64bit, little endian, Intel(R) Pentium(R) Gold 7505 @ 2.00GHz
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #21491, opened 2024-05-11
 cgen: handle auto deref var for index when array type is interfaces and sum types

Delta456:
Fixes https://github.com/vlang/v/issues/21490


vlib/v/tests/interface_arr_for_mut_iter_index_test.v
```diff
@@ -0,0 +1,14 @@
...
+
+struct Map {
+	ennemis []Ennemi
+}
+
+fn test_interface_arr_for_mut_iter_index() {
+	mut maap := Map{}
+
+	for mut ennemi in maap.ennemis {
+		assert maap.ennemis.index(ennemi) == 0
```
Delta456, line 12: I want to know if there is a better way to test this line is code.

spytheman: Better in what way?

The test fails to compile on master with a cgen error, but succeeds here.
Imho that is enough to prove, that it fixed the problem, and that a regression will not happen, for that exact situation in the future.

Delta456: I mean a better way to test by not checking the length.

spytheman (approved):
Excellent work.

vlib/v/tests/interface_arr_for_mut_iter_index_test.v
spytheman: I think it is good as it is.

PR #21510, opened 2024-05-15
cgen: fix C struct init when it has default expr

### LINKED ISSUE #21501: using a C struct as a struct field does not adopted default values
### Describe the bug

```c
// header.h
typedef struct Bar {
  int a;
} Bar;
```

```v
#include "@VMODROOT/header.h"

struct Foo {
	a int = 3
}

struct C.Bar {
	a int = 3
}

struct FooBar {
	foo Foo
	bar C.Bar
}

dump(Foo{}) // has default value

dump(C.Bar{}) // has default value

dump(FooBar{}) // Foo has default value while C.Bar does not
```

### Reproduction Steps

above

### Expected Behavior

default values work for the C struct also when it is used as a field of another struct.

### Current Behavior

no default values when the C struct is used as a struct field.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.5 58a8fc6

all
> [!NOTE]
>

felipensp:
Fix #21501

### DIFF
--- a/vlib/v/tests/c_structs/cstruct_default_value_test.v
+++ b/vlib/v/tests/c_structs/cstruct_default_value_test.v
@@ -1,0 +1,25 @@
+#include "@VMODROOT/cstruct.h"
+
+struct Foo {
+	a int = 3
+}
+
+struct C.Bar {
+	a int = 3
+	b f64
+}
+
+struct FooBar {
+	foo Foo
+	bar C.Bar
+}
+
+fn test_main() {
+	a := dump(Foo{})
+	b := dump(C.Bar{})
+	c := dump(FooBar{})
+
+	assert a.a == b.a
+	assert b.a == c.bar.a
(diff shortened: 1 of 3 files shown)


vlib/v/tests/c_structs/cstruct_default_value_test.v
```diff
@@ -0,0 +1,25 @@
...
+}
+
+fn test_main() {
+	a := dump(Foo{})
+	b := dump(C.Bar{})
+	c := dump(FooBar{})
+
+	assert a.a == b.a
+	assert b.a == c.bar.a
+	assert b.b == c.bar.b
```
spytheman, line 24: I think that if you test against the actual known concrete values here, the tests will be clearer, and failures will be easier to diagnose.

spytheman (approved):
Excellent work.

PR #21521, opened 2024-05-17
vdoc: improve vdocignore file handling by walking all .vdocignore sub-paths in IgnoreRules.get, add test

ttytm:
With the changes, all ignore rules for a specified directory and its subdirectories are gathered first. This removes coupling with the process of getting modules and should simplify  and improve testability. A test is also added.


spytheman (approved):
Excellent work.

PR #21523, opened 2024-05-18
parser: parse string and array typ idx of `ScopeVar` and `Ident`

ttytm:
Will extend capabilities and e.g. allow to finish #21421.

Current types in `scope` and `info`
```sh
  scope: &# 3 - 311
* var: foo - ast.Type(0x0 = 0) # e.g. string type
* var: baz - ast.Type(0x0 = 0) # e.g. array type
# ...
info: ast.IdentInfo(ast.IdentVar{
    typ: ast.Type(0x0 = 0)
    share: mut
    is_mut: false
    is_static: false
    is_volatile: false
    is_option: false
})
# ...
```
Updated:
```sh
  scope: &# 3 - 311
* var: foo - ast.Type(0x15 = 21)
* var: baz - ast.Type(0x17 = 23)
# ...
info: ast.IdentInfo(ast.IdentVar{
    typ: ast.Type(0x17 = 23)
    share: mut
    is_mut: false
    is_static: false
    is_volatile: false
    is_option: false
})
# ...
```

The parsing can be limited to `p.pref.is_vet` and `p.pref.is_fmt` if it is preferred.

### DIFF
--- a/vlib/v/parser/assign.v
+++ b/vlib/v/parser/assign.v
@@ -226,6 +226,11 @@ fn (mut p Parser) partial_assign_stmt(left []ast.Expr) ast.Stmt {
 						pos: lx.pos
 						is_stack_obj: p.inside_for
 					}
+					if p.prev_tok.kind == .string {
+						v.typ = ast.string_type_idx
+					} else if p.prev_tok.kind == .rsbr {
+						v.typ = ast.array_type_idx
+					}
 					if p.pref.autofree {
 						r0 := right[0]
 						if r0 is ast.CallExpr {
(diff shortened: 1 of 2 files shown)


spytheman:
This is an improvement, but note that the problem, is that parsing is done on individual .v files, while the type information requires a pass over the whole module AST in the general case.
Consider, what will `v vet` do for a file that uses a constant or a symbol, defined in the same module, but in another .v file - its .typ will be still 0 ...

ttytm:
Right, for the global scope this won't work.

Also, this makes me remember something.

Some months ago, I started working on parsing all files of a module to at least get all declarations into global scope to fix the `modprefix.constant` formatting for files in a module that use a constant declared in another file of the module. I made some good progress but didn't follow through with it. I'll try to dig these changes out again and finish them. Unfortunately, my time is more limited currently.

spytheman (approved):

PR #21526, opened 2024-05-18
vlib: rename file for pure V check

JalonSolov:
Files with `C.` contents should have `.c.v` extension to pass pure V checks.


spytheman:
Done in 6389da7 . I also spotted it when running the tests for the coreutils :-| .

PR #21538, opened 2024-05-20
cgen: fix assert rvalue printing for float numbers

### LINKED ISSUE #21536: v test does not print float value
### Describe the bug

When asserting against a wrong f32 or f64 value, the output of `v test` show `*unknown value*` instead of the actual value.

### Reproduction Steps

run v test on

```V
pub fn test_format() {

	f := 0.2
	assert f == 0.1
}
```

### Expected Behavior

```
   > assert f == f64(0.1)
     Left value: 0.2
    Right value: 0.1
```

### Current Behavior

```
   > assert f == f64(0.1)
     Left value: 0.2
    Right value: *unknown value*
```

### Possible Solution

_No response_

_No response_

### V version

0.4.5

V full version: V 0.4.5 c0afe5d.f33b712
OS: linux, Ubuntu 22.04.4 LTS
> [!NOTE]
>

felipensp:
Fix #21536

```v
f := 0.2
assert f == 0.3
```


JalonSolov:
The problem with IEEE-754 floats is they are not guaranteed to be exact to any particular number of digits.  Assert for floats will only ever accidentally match exactly.

It might be better to simply not allow `==` checks on pure floats... only `<`, `>`, `>=`, etc.  Inexact matches.

Delta456:
> The problem with IEEE-754 floats is they are not guaranteed to be exact to any particular number of digits.  Assert for floats will only ever accidentally match exactly.
> 
> It might be better to simply not allow `==` checks on pure floats... only `<`, `>`, `>=`, etc.  Inexact matches.

If `==` is allowed for all other operations of `float` then why disallow for `assert` statements? If you are going to disallow equality operations for floats, may just disallow it completely.

JalonSolov:
For example, a _very_ simple test might always work...
```v
pub fn test_format() {
        f := 0.2
        assert f == 0.2
}
```
However, if you did several calculations before the assert, you might instead end up with
```
  > assert f == f64(0.2)
     Left value: 0.2
    Right value: 0.200001
```

Delta456:
I can work with a `notice` alert. Not an error.

JalonSolov:
That would be fine, I suppose.  The problem is that many (if not most) _expect_ floats to act the same as ints (as in, precision the same, no round-off errors, etc.), but they don't.

And when they don't, we get questions about why they don't.

So letting them know up front that using `==` on floats may not always work would be a good thing.

medvednikov:
I remember I wanted to ban `float1 == float2` way back, but there was a discussion with some arguments against this.

Can't find it now.

I think it makes sense.

spytheman (approved):
Excellent work.

spytheman:
The PR is only slightly related to `x == 0.11111` etc.

It fixes printing of floating point values for *all assertions*, including for `assert f > x`, see for example:
image

```
fn test_abc() {
	f := 0.0
	assert f > 0.3
}
```

PR #21540, opened 2024-05-20
v compiler optimisations

spytheman:
- **checker: minimise the amount of string interpolation inside selector_expr**
- **builtin: remove string.clone() calls, in the most common case of interpolating strings (with no uppercasing or padding)**
- **v.ast: use != 0 instead of > 0, for commonly used methods like Type.has_flag() etc**
- **v.ast: eliminate array_get calls from Table.final_sym and Table.sym (make profiling data more useful)**
- **cgen: reduce interpolations in Gen.string_inter_literal**
- **cgen: small optimisation of frequently called cescape_nonascii**
- **v.util: check more strictly for a valid input in util.tabs/1**

### DIFF
--- a/vlib/builtin/string_interpolation.v
+++ b/vlib/builtin/string_interpolation.v
@@ -212,38 +212,45 @@ fn (data &StrIntpData) process_str_intp_data(mut sb strings.Builder) {
 	unsafe {
 		// strings
 		if typ == .si_s {
-			mut s := ''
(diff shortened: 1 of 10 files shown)


spytheman:
The RAM usage of the compiler is improved a bit, as well as the performance of mainly the checker:
image

image

felipensp:
Excellent work. 🥇

PR #21545, opened 2024-05-22
v.scanner: optimise performance, use precomputed tables, and s.substr_unsafe/2 instead of s.substr/2 in .indent_name/0 and .num_lit/2

spytheman:
- **v.scanner: use precomputed character tables instead of checks**
- **v.scanner: use util.func_char_table in Scanner.ident_name/0**
- **v.scanner: restore the checks in ident_name, instead of using util.func_char_table**
- **v.scanner: split long condition in ident_name into several smaller ones**
- **v.scanner: use substr_unsafe in Scanner.ident_name/0 and Scanner.num_lit/2**

### DIFF
--- a/vlib/v/scanner/scanner.v
+++ b/vlib/v/scanner/scanner.v
@@ -239,7 +239,7 @@ fn (mut s Scanner) ident_name() string {
 	s.pos++
 	for s.pos < s.text.len {
 		c := s.text[s.pos]
-		if (c >= `a` && c <= `z`) || (c >= `A` && c <= `Z`) || (c >= `0` && c <= `9`) || c == `_` {
+		if util.func_char_table[c] {
 			s.pos++
 			continue
 		}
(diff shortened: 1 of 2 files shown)


vlib/v/scanner/scanner.v
```diff
@@ -239,20 +239,32 @@ fn (mut s Scanner) ident_name() string {
...
+		}
+		if c == `_` {
+			s.pos++
+			continue
+		}
+		if c >= `A` && c <= `Z` {
+			s.pos++
+			continue
+		}
+		if c >= `0` && c <= `9` {
```
ttytm, line 254: This change makes me think whether it could make sense to preserve line breaks in conditions if they meet certain criteria in vfmt.

```v
	if (c >= `a` && c <= `z`)
		|| (c >= `A` && c <= `Z`)
		|| (c >= `0` && c <= `9`)
		|| c == `_` {
		s.pos++
		continue
	}
```

spytheman:
That will have to wait, until we have a bit better tooling. I could not find the bug causing the remaining CI failures.

PR #21557, opened 2024-05-23
examples: cleanup `process/execve.v`

ttytm:
Cleans up an example that imho can become unclear to the viewer due to unnecessary complexity and confusing comments.

Alternatively, removing the file and adding the examples to the function comment could also work here:
```v
// Example:
// ```v
// execve('/usr/bin/ls', ['-la', '-h'], [])! // Passes only an array of args.
// execve('/usr/bin/bash', ['-c', 'ls -lah -s'], [])! // Considers args that would need to be passed within quotes.
// execve('/usr/bin/man', ['true'], ['MANWIDTH=60'])! // Passes an environment variable that affects the command.
// ```
```

### DIFF
--- a/examples/process/execve.v
+++ b/examples/process/execve.v
@@ -1,17 +1,14 @@
-module main
-
 import os
 
-fn exec(args []string) {
-	os.execve('/bin/bash', args, []) or {
-		// eprintln(err)
-		panic(err)
-	}
-}
-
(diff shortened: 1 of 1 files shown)


spytheman:
(rebased over master, to fix the CI failure on FreeBSD after ef758a7)

PR #21558, opened 2024-05-24
examples/audio: fix audio example to work on FreeBSD

kimshrier:
This just adds the configuration needed for compiling the audio example.


Larpon:
@kimshrier good work! Any chance these changes could end up in sokol upstream?

kimshrier:
I haven't tried pushing this change upstream.  I'll look into it.

kimshrier:
I pushed this change upstream and it was merged. [1055](https://github.com/floooh/sokol/pull/1055)

JalonSolov:
So instead of this PR, we need one to update to the latest sokol, right?

kimshrier:
yes

PR #21565, opened 2024-05-24
os: make minor improvement to utime and uname C fn semantics and their related code

ttytm:


vlib/os/os_nix.c.v
```diff
@@ -214,8 +214,8 @@ fn native_glob_pattern(pattern string, mut matches []string) ! {
 }
 
 pub fn utime(path string, actime int, modtime int) ! {
-	mut u := C.utimbuf{actime, modtime}
-	if C.utime(&char(path.str), voidptr(&u)) != 0 {
+	u := &C.utimbuf{actime, modtime}
```
spytheman, line 217: `&C.Struct{}` makes a heap allocation.

`C.utimbuf{actime, modtime}` creates a structure on the stack, without additional allocations.

spytheman (approved):

vlib/os/os_nix.c.v
ttytm: won't `&u` also create a heap alloc?

PR #21566, opened 2024-05-24
builtin, strings: improve consistency and behavior regarding explicit mutability

ttytm:
Will require to pass a buffer arg that is muted as mutable.

The variables that were passed to the updated utf32_ fns were were already declared as mutable, the change makes sure the arg needs to be passed as mutable as well.


ttytm:
@spytheman this one would would be need to sync merged with the PR at vlang/ved to allow for green CI.

PR #21567, opened 2024-05-24
ci: make use of apt flags consistent, remove obsolete

ttytm:
`-qq` implies `-y`. 
Ref. https://linux.die.net/man/8/apt-get


spytheman (commented):
I prefer to have the explicit `-y`.

PR #21571, opened 2024-05-25
toml: fix `@[toml: ]`, support `@[skip]`

heyimtcn:
fix `toml.decode_struct[T]()` to support `field @[toml: name]`


ttytm:
Nice contrib @heyimtcn.

Can you add a test? I think that the attribs would need to be respected during encoding as well to have consistent behavior.

heyimtcn:
patched @[toml: ] in encode, added @[skip] and wrote tests

spytheman (approved):
Excellent work.

heyimtcn:
why do formatting checks keep failing?? i even tested it locally and it works fine...

vlib/toml/tests/toml_attrs_test.v
```diff
@@ -0,0 +1,15 @@
...
+	baz string = 'def' @[toml: barbaz]
+}
+
+fn test_toml_attr_encode() {
+	assert toml.encode(TestStruct{}) == 'foo = 0\nbarbaz = "def"'
+}
+
+fn test_toml_attr_decode() {
+	assert toml.decode[TestStruct]('foo = 0\nbarbaz = "def"')! == TestStruct{}
+}
```
ttytm, line 15: Screenshot_20240526_021641


Because of windows crlf.

I"m wondering how they even got in there since there is an editorconfig and gitattributes that should prevent them.


```sh
❯ v fmt -diff vlib/toml/tests/toml_attrs_test.v
warning: in the working copy of 'vlib/toml/tests/toml_attrs_test.v', CRLF will be replaced by LF the next time Git touches it
```

`v fmt -w vlib/toml/tests/toml_attrs_test.v` fixes them on my end though.

```suggestion
import toml

struct TestStruct {
	foo int
	bar bool   @[skip]
	baz string = 'def' @[toml: barbaz]
}

fn test_toml_attr_encode() {
	assert toml.encode(TestStruct{}) == 'foo = 0\nbarbaz = "def"'
}

fn test_toml_attr_decode() {
	assert toml.decode[TestStruct]('foo = 0\nbarbaz = "def"')! == TestStruct{}
}
```

ttytm: Looks like it won't work with the suggestion, an update needs to be pushed directly.

spytheman: Done in f571fa0 .

spytheman:
> why do formatting checks keep failing?? i even tested it locally and it works fine...

It may have been because of some interaction between vfmt, your editor or your git settings, as @ttytm noted.

`v fmt -w vlib/toml` done manually in the terminal, fixed it for me.

I have `core.autocrlf=input` and `core.safecrlf=false` in my .gitconfig file, if that is of any help.

ttytm:
Right, there might have been an overwrite due to a local setting. Let us know if you spot what was causing it 😊

PR #21584, opened 2024-05-27
ci: extend coverage, move app test from macos_ci to v_apps_and_modules_ci

ttytm:
Extends coverage on macOS and Linux

- V Apps and Modules get tested on macOS in `v_apps_and_modules_compile_ci.yml`
- macOS app-tests (ved and vlang/ui) that were tested in `macos_ci.yml` are now part of the above workflow
- vlang/ui examples that were tested on mac are now also tested on Linux
- vpm test runs include macos arm


spytheman (approved):
Excellent work.

PR #21600, opened 2024-05-29
fmt: don't break `it` variable name in match expression function calls

ttytm:
Fixes

```v
import arrays

struct Release {
	version    string
	prerelease bool
}

releases := [Release{'1.0.0-dev', true}, Release{'0.2.0', false}, Release{'0.1.0', false}]

version := 'nightly'
v := match version {
	'nightly' {
		arrays.find_first(releases, fn (it Release) bool {
			return it.prerelease
		})?
	}
	else {
		Release{}
	}
}

dump(v)
```

Where v fmt does
```diff
v := match version {
	'nightly' {
		first_l_match := arrays.find_first(arr, fn (it Foo) bool {
-			return !it.prerelease
+			return !version.prerelease
		})?
	}
}
```

The fix is achieved by removing code whose use was likely superseded and it remained as a hotbed for errors. This assumption is based no breaking tested cases and not being able to trigger what the condition apparently should cover.

### DIFF
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -39,7 +39,7 @@ pub mut:
 	mod2alias          map[string]string // for `import time as t`, will contain: 'time'=>'t'
 	mod2syms           map[string]string // import time { now } 'time.now'=>'now'
 	use_short_fn_args  bool
-	single_line_fields bool   // should struct fields be on a single line
+	single_line_fields bool // should struct fields be on a single line
 	in_lambda_depth    int
 	inside_const       bool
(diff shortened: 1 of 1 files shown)


spytheman:
The original code was added in https://github.com/vlang/v/pull/11967 , which was concerned with cases like `a.all(it.len > 0 && it[1..].bytes().all( it.is_alnum() || it == 5 ) )` , but there were no fmt tests added, beside the code in `scanner.v` itself.

Given that `scanner.v` still formats the same, I think it is safe to say that this PR works.

spytheman (approved):
Good work.

PR #21601, opened 2024-05-29
Update docs.md (Error in example with avoiding reallocation of the array)

Zerg2003:
The array numbers has 1000 elements, not 1001

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -995,7 +995,7 @@ as reallocations can be avoided:
 mut numbers := []int{cap: 1000}
 println(numbers.len) // 0
 // Now appending elements won't reallocate
-for i in 0 .. 1000 {
+for i in 0 .. 999 {
 	numbers << i
 }
 ```


ttytm:
I think the current example is correct. 

```
mut numbers := []int{cap: 1000}
println(numbers.len) // 0
// Now appending elements won't reallocate
for i in 0 .. 1000 {
	numbers << i
}
dump(numbers.first()) // 0
dump(numbers.last()) // 999
```

With with update it would be 999 elements - where `numbers.last()` would be `998`.

Zerg2003:
yes, sorry, you are right

JalonSolov:
Yes, the example is correct.  The array is allocated with 1000 elements of space.

Then 1000 elements (0 -> 999) are added.

Zerg2003:
nums := [1, 2, 3]
println(nums.any(it == 2)) // true //???????????
println(nums.all(it >= 2)) // false

why this? ANY means any: 1, 2 or 3! Why is it true? 1 != 3. 3 != 2.
Is it more logical to use println(nums.some(it == 2)) // true

ttytm:
> Another suspicious example:
> 
> nums := [1, 2, 3] println(nums.any(it == 2)) // true //??????????? println(nums.all(it >= 2)) // false
> 
> why this? ANY means any: 1, 2 or 3! Why is it true? 1 != 2. 3 != 2. Is it more logical to use println(nums.some(it == 2)) // true

Why "Another"? Was the first one suspicious :P? 

Well you have `nums := [1, 2, 3]` and check if any of it is `2` and it obviously is so it must be `true`. And and whe nyou check that all of them are `2` or larger it's false because you have a 1.

Zerg2003:
<"it" is a builtin variable which refers to the element currently being processed in filter/map methods.
Additionally, .any() and .all() can be used to conveniently test **for elements** that satisfy a condition.>

So, in English <"any" of the elements of the array has value of 2> is not equivalent to <"some" of the elements of the array has value of 2>. And it seams to me obvious. So, using "any" for such purpose is an obvious misuse!

JalonSolov:
`any` means "if **_any_** of the values in the array matches the given condition"
`all` means "if **_all_** of the values in the array match the given condition"

So both examples are correct.  In the first, one of the values does match `2`.  In the second, one of the values does NOT match `>= 2`, which means they do not `all` match.

JalonSolov:
(In case you're wondering... I am an American, and got 98% as my daily average grade in English class... so I do know a bit about what things mean in English.)

Zerg2003:
Ok, thank you.

PR #21604, opened 2024-05-30
builtin:  update `last_index_u8`, deprecate `index_u8_last` string methods, make consistent with `last_index`

ttytm:
Updates the `index_u8_last` => `last_index_u8` method and deprecates the former to be consistent with the recent switch made by Alex from `index_last` to `last_index`. It is also consistent with the Go func equivalents.

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -1338,33 +1338,23 @@ pub fn (s string) index_u8(c u8) int {
 	return -1
 }
 
-// TODO: this no longer needs to be a separate method after index_u8_last has been removed.
-@[direct_array_access]
-fn (s string) last_index_u8_(c u8) int {
+// index_u8_last returns the index of the *last* occurrence of the byte `c` (if found) in the string.
+// It returns -1, if `c` is not found.
+@[deprecated: 'use `.last_index_u8(c u8)` instead']
+@[deprecated_after: '2024-06-30']
+pub fn (s string) index_u8_last(c u8) int {
+	return s.last_index_u8(c)
+}
+
+// last_index_u8 returns the index of the last occurrence of byte `c` if it was found in the string.
+@[inline]
+pub fn (s string) last_index_u8(c u8) int {
 	for i := s.len - 1; i >= 0; i-- {
 		if s[i] == c {
 			return i
 		}
 	}
 	return -1
-}
-
-// index_u8_last returns the index of the *last* occurrence of the byte `c` (if found) in the string.
-// It returns -1, if `c` is not found.
-@[deprecated: 'use `.last_index_u8(c u8)` instead']
-@[deprecated_after: '2024-06-30']
-pub fn (s string) index_u8_last(c u8) int {
-	return s.last_index_u8_(c)
-}
-
-// last_index_u8 returns the index of the last occurrence of byte `c` if it was found in the string.
-@[inline]
-pub fn (s string) last_index_u8(c u8) ?int {
-	idx := s.last_index_u8_(c)
-	if idx == -1 {
-		return none
-	}
-	return idx
 }
 
 // count returns the number of occurrences of `substr` in the string.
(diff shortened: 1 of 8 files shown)


spytheman (approved):

vlib/builtin/string.v
```diff
@@ -1338,25 +1338,33 @@ pub fn (s string) index_u8(c u8) int {
...
+@[deprecated_after: '2024-06-30']
+pub fn (s string) index_u8_last(c u8) int {
+	return s.last_index_u8_(c)
+}
+
+// last_index_u8 returns the index of the last occurrence of byte `c` if it was found in the string.
 @[inline]
-pub fn (s string) last_index_u8(c u8) int {
-	return s.index_u8_last(c)
+pub fn (s string) last_index_u8(c u8) ?int {
```
spytheman, line 1362: That is a breaking change.
The previous variant on master returned `int`. The new one returns `?int` .

ttytm: Right, `last_index_u8` that was the deprecated before returned an int. Now, the `revived` version matches `last_index` returning an `?int`.

spytheman: I can not merge this. Unlike the `index_u8_last` deprecation, this is breaking.

It would have to be another PR, and then we can decide on it there separately.

ttytm: I disagree. From my perspective (with the users in mind), there are a few reasons why it can (and even should) be merged in one go when making this change. Doing so with a few days' delay (on 2024-06-15) could fit. Some thoughts:

- The deprecation is already a breaking change, whether with `int` or `?int`.
- Moving from `int` to `?int` would be a second breaking change users have to deal with, so splitting it is adding hassle not just for ourselves.
- The deprecation period for `last_index_u8` is 2024-06-15; it would be an error after this anyway. `last_index_u8` with `?int` can then be classified as a "new" function that is not just consistent with `last_index` but also more V-style, returning `none` if a value is not found to allow better and safer handling of that case instead of `-1`, which is not necessarily checked. This makes it imho less of a breaking change than splitting it would do.
- The deprecation time of `index_u8_last` could also be further extended so that users have more time to adapt to `last_index_u8`.

ttytm: Of course, if it's still preferred to split it, please let me know. Then I'll push the update accordingly.

spytheman: Please note, that deprecations do have a grace period, when they are notices, then warnings, then errors, so no, it is not an immediate breaking change, and people have time to update their usages.

> Moving from int to ?int will be a second breaking change ...

It would be, *if* it is done, which is subject to another decision.
Note that all other _u8 methods do return `int` too, so there is a strong argument in favor of keeping that for `last_index_u8` too. Keeping them return `int` instead of `?int` is also preferable for performance reasons.

> The deprecation period for last_index_u8(deprecated on 2023-12-18) is running out on 2024-06-15

That date is still in the future, is not it?

> It's not an immediate break, index_u8_last continues to work for a while. 

I commented under:
```v
- pub fn (s string) last_index_u8(c u8) int {
+ pub fn (s string) last_index_u8(c u8) ?int {
```
That is not `index_u8_last`.


I do insist on my decision to split the breaking change, from the deprecation.
The deprecation can be merged without problems. I will not merge the breaking change as it is here.

ttytm: > Please note, that deprecations do have a grace period, when they are notices, then warnings, then errors, so no, it is not an immediate breaking change, and people have time to update their usages.

I'm aware, it was part of my comment. I mentioned that a merge could fit after the grace period. 
Currently, the grace period of **`last_index_u8`** is running out on **`2024-06-15`** and it would become an error in about two weeks for everyone who is using it.

> Keeping them return int instead of ?int is also preferable for performance reasons.

A good argument! It can be significantly faster. This makes me prefer the `int`. 
In any case, imho we should opt for consistency. So updating `last_index` to use an int would be an option, like the go LastIndex and LastIndexByte do.

> I commented under:
> [...]
> That is not index_u8_last.

**`index_u8_last`** is deprecated in favor of `last_index_u8`, which is why the deprecation period of the current **`index_u8_last`** and how long it will continue to work is of relevance.

PR #21612, opened 2024-05-30
builtin,js: implement JS `string.split_any`

### LINKED ISSUE #21440: backend js: no method string.split_any
### Describe the bug

The js backend lacks a string.split_any method.

### Reproduction Steps

```v
fn main() {
  println('foo bar'.split_any(' '))
}
```

### Expected Behavior

Should compile and output:
```
['foo', 'bar']
```

### Current Behavior

```
# v -b js test.v
test.v:2:21: error: unknown method or field: `string.split_any`.
Did you mean `split`?
    1 | fn main() {
    2 |   println('foo bar'.split_any(' '))
      |                     ~~~~~~~~~~~~~~
    3 | }
test.v:2:3: error: `println` can not print void expressions
    1 | fn main() {
    2 |   println('foo bar'.split_any(' '))
      |   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    3 | }
If the code of your project is in multiple files, try with `v .` instead of `v test.v`
[I] root@trump
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.5 387af74

```
V full version: V 0.4.5 387af74
OS: linux, "Artix Linux"
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i7-6600U CPU @ 2.60GHz

getwd: /home/root
vexe: /home/root/Apps/v/v
vexe mtime: 2024-05-04 22:42:37

VTMP: OK, value: /tmp/v_0

Git version: git version 2.45.0
Git vroot status: weekly.2024.18.2-13-g2e64e284-dirty (9 commit(s) behind V master)
.git/config present: true

CC version: cc (GCC) 13.2.1 20240417

```
> [!NOTE]
>

juan-db:
Implement `string.split_any` for the JS backend.

This PR fixes #21440.

### DIFF
--- a/vlib/builtin/js/string_test.js.v
+++ b/vlib/builtin/js/string_test.js.v
@@ -252,12 +252,11 @@ fn test_split_any() {
 	assert a[4] == '5'
 	s = '12131415'
 	a = s.split_any('2345')
-	assert a.len == 5
+	assert a.len == 4
 	assert a[0] == '1'
 	assert a[1] == '1'
 	assert a[2] == '1'
 	assert a[3] == '1'
-	assert a[4] == ''
 	s = 'a,b,c'
 	a = s.split_any('],')
 	assert a.len == 3
--- a/vlib/builtin/js/string_test.js.v
+++ b/vlib/builtin/js/string_test.js.v
@@ -232,6 +232,52 @@ fn test_split() {
 	assert vals[1] == ''
 }
 
+fn test_split_any() {
+	mut s := 'aaa'
+	mut a := s.split_any('')
+	assert a.len == 3
+	assert a[0] == 'a'
+	assert a[1] == 'a'
+	assert a[2] == 'a'
+	s = ''
+	a = s.split_any('')
+	assert a.len == 0
+	s = '12131415'
+	a = s.split_any('1')
+	assert a.len == 5
+	assert a[0] == ''
+	assert a[1] == '2'
+	assert a[2] == '3'
+	assert a[3] == '4'
+	assert a[4] == '5'
+	s = '12131415'
+	a = s.split_any('2345')
+	assert a.len == 5
+	assert a[0] == '1'
+	assert a[1] == '1'
+	assert a[2] == '1'
+	assert a[3] == '1'
+	assert a[4] == ''
+	s = 'a,b,c'
+	a = s.split_any('],')
+	assert a.len == 3
+	assert a[0] == 'a'
+	assert a[1] == 'b'
+	assert a[2] == 'c'
+	s = 'a]b]c'
+	a = s.split_any('],')
+	assert a.len == 3
+	assert a[0] == 'a'
+	assert a[1] == 'b'
+	assert a[2] == 'c'
+	s = 'a]b]c'
+	a = s.split_any('],\\')
+	assert a.len == 3
+	assert a[0] == 'a'
+	assert a[1] == 'b'
+	assert a[2] == 'c'
+}
+
 /*
 fn test_trim_space() {
 	a := ' a '
--- a/vlib/builtin/js/jsfns.js.v
+++ b/vlib/builtin/js/jsfns.js.v
@@ -19,6 +19,12 @@ pub interface JS.Number {
 	JS.Any
 }
 
+pub interface JS.RegExp {
+	JS.Any
+}
+
+pub type SplitSeparator = JS.String | JS.RegExp
+
 @[single_impl]
 pub interface JS.String {
 	JS.Any
@@ -32,7 +38,7 @@ pub interface JS.String {
 	endsWith(substr JS.String) JS.Boolean
 	startsWith(substr JS.String) JS.Boolean
 	slice(a JS.Number, b JS.Number) JS.String
-	split(dot JS.String) JS.Array
+	split(delim SplitSeparator) JS.Array
 	indexOf(needle JS.String) JS.Number
 	lastIndexOf(needle JS.String) JS.Number
 }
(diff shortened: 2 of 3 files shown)


vlib/builtin/js/string_test.js.v
```diff
@@ -232,6 +232,52 @@ fn test_split() {
...
+	a = s.split_any('1')
+	assert a.len == 5
+	assert a[0] == ''
+	assert a[1] == '2'
+	assert a[2] == '3'
+	assert a[3] == '4'
+	assert a[4] == '5'
+	s = '12131415'
+	a = s.split_any('2345')
+	assert a.len == 5
```
spytheman, line 255: The pure V version, makes: `assert a.len == 4` in this case, not 5.

spytheman: The result is: `['1', '1', '1', '1']` .

spytheman: The JS version produces: `['1', '1', '1', '1', '']` .

juan-db: That's a bit unintuitive, it's not really "splitting" on the '5' then.
The default split function would produce the empty string:
```v
'12131415'.split('5')
// ['1213141', '']
```

I don't really know where else to reference but Kotlin produces the same result as JS.
https://pl.kotl.in/HxZ8xuz_i

JalonSolov: There's nothing to split on the 5 - it's the end of the string.

juan-db: I disagree. Whenever it encounters a delimiter, it should split the string. It also means the behaviour is inconsistent with the regular `split` function.

juan-db: It's also inconsistent how it handles the start and end of the string, since `'1213'.split_any('1')` returns `['', '2', '3']`.

JalonSolov:
In the string `2345`, there is nothing after the 5, therefore nothing to split.  The only splits would be between `2` and `3, `3` and `4`, and `4` and `5`.

The only thing after the 5, is the `NUL` at the end, which is not a "real" character.

vlib/builtin/js/string_test.js.v
spytheman: I can not decide which is the correct behavior, but in any case, the pure V and the JS version should produce the same results.

Allowing differences between them, will lead to hard to diagnose bugs for user programs that use them, across the backends.

JalonSolov:
Doing it like this is just as bad as the Go `utils.TrimSplit()` function... if you call that on an empty string, it does NOT return an empty array... it returns an array with one element, and that element is `''`.

juan-db:
> In the string `2345`, there is nothing after the 5, therefore nothing to split. The only splits would be between `2` and `3`, `3` and `4`, and `4` and `5`.
> 
> The only thing after the 5, is the `NUL` at the end, which is not a "real" character.

What is the V version splitting in the `1213.split_any('1')` case then to give `['', '2', '3']`? Nothing precedes the first character in the string.

vlib/builtin/js/string_test.js.v
juan-db: I agree. In my opinion the V version should be updated to match the JS version. Every other instance I can think of follows the JS/Regex logic. [.NET](https://learn.microsoft.com/en-us/dotnet/api/system.string.split?view=net-8.0#example) is another example.

JalonSolov:
It split between with `1` as the delimiter.  Since there was only an "empty string" before the first delimiter, that's all it gave.

juan-db:
I understand that reasoning, but it feels like that's more driven by implementation result than actual intention. The most common use case I know for split is for CSV values. In these cases, you expect it to **split the string** every time it encounters the delimiter. E.g. if you had a CSV with `person,start,end,role,earnings,manager` and a value row of `john,2015,,founder,1USD,`, you want to extract each field. `split_any` is a more niche version so I don't really know what the most common use case for it would be over regular `split`, but either way I believe the behavior should be consistent.

JalonSolov:
The difference is in the description of `split_any`...
```
Split a string using the chars in the delimiter string as delimiters chars.
```
Meaning, you can include several different delimiters, and it will split on all of them.
```v
println('12;34:56^78,'.split_any(';:^,'))
```
output is
```
['12', '34', '56', '78']
```
Obviously contrived, but suppose you had a timestamp in a log file, and wanted to split the pieces... with `split_any` you could split them all in one call, instead of separate calls with each different separator, and trying to figure it out from there...
```v
println('[2024-05-31_19:39:23]'.split_any('[-_:]'))
```
output is
```
['', '2024', '05', '31', '19', '39', '23']
```
... and in this case, I'd have to agree... the `''` as the first entry looks like a bug to me.

juan-db:
I'm not sure how to proceed in this case because the suggested behavior is contrary to all my experiences with other languages and isn't consistent with the regular split behavior where it overlaps in concept.

In rough pseudocode, this seems like the correct simplified example to me.
```
output := []
buffer := ''

for c in str:
    if c == delim
        // *any* time we encounter delimiter, *split the string*
        output << buffer
        buffer = ''
    else
        buffer += c

output << buffer // make sure everything after the last delimiter is included

// for the suggested solution, instead of the above, we would
// have to handle the end of the string as a special case
/*
if buffer.len != {
    output << buffer
}
*/

return output
```

juan-db:
I don't hold much authority here, so I think either someone needs to decide it's more important to keep the existing V behavior, or we need more input on what the expected behavior should be.

spytheman:
Here is a plan of action:
1) Make the JS version match the V version, so that the existing PR can be merged, without breaking existing code
2) Make *another PR*, that changes the behavior of both the pure V and the JS versions to what you think is a more useful behavior (that would be a breaking change), but it will be caught by the CI, and we can test it extensively there.

spytheman (approved):
Excellent work.

PR #21613, opened 2024-05-30
parser: register selective import aliases as used

### LINKED ISSUE #21611: Selective imports are not selective.
### Describe the bug

The feature currently known as `selective imports` means:

* Import the module: `import module`.
* Allow for specific names to be used without a namespace: `{ Name }`.

The module is still imported. This is understood by the current implementation (see below) but the import is not selective as the entire module is imported.

### Reproduction Steps

```v
// thing.v
module thing

pub struct Thing {}
```

```v
// main.v
import thing { Thing }

fn main() {
        thing := Thing{}
        dump(thing)
}
```

### Expected Behavior

Semantics change across module boundaries. Selectively importing should allow the call site to decide on all symbols outside the set of the imported symbols.

### Current Behavior

```
$ v .
main.v:5:2: error: duplicate of an import symbol `thing`
    3 | 
    4 | fn main() {
    5 |     thing := Thing{}
      |     ~~~~~
    6 |     dump(thing)
    7 | }
```

### Possible Solution

A primary use case is data modeling. Module names are likely similar or identical to structures that they contain (above: `thing.v` models structure `Thing`). The call site that makes use of a model could just as likely prefer a matching name, for example an application `session` variable of type `Session`.

In the current implementation, the programmer must work around V's behavior:

```v
// main.v
import thing { Thing }

fn main() {
        th := Thing{}
        dump(th)
}
```

Selectively importing without importing the module eliminates astonishment and the need for workarounds.

It would then be explicit if the module itself is needed (as in e.g. Python):

```v
import thing
import thing { Thing }
```

If after reviewing this issue, the consensus is to keep the behavior of importing the module when also importing symbols within it, then I humbly request that the V project rename "selective import" to something else.

#12414 has consensus that module names are not to be shadowed in V.

The behavior of importing the module on selective import is [understood by the current implementation](https://github.com/vlang/v/blob/52dfef1/doc/docs.md?plain=1#L1543).

### V version

V 0.4.6 52dfef1

```
CC version: cc (Gentoo 13.2.1_p20240210 p14) 13.2.1 20240210
```
> [!NOTE]
>

ttytm:
Fixes #21611

### DIFF
--- a/vlib/v/parser/module.v
+++ b/vlib/v/parser/module.v
@@ -33,6 +33,12 @@ fn (mut p Parser) register_used_import(alias string) {
 
 fn (mut p Parser) register_used_import_for_symbol_name(sym_name string) {
 	short_import_name := sym_name.all_before_last('.').all_after_last('.')
+	for alias, mod in p.imports {
+		if mod == short_import_name {
+			p.register_used_import(alias)
+			return
+		}
+	}
 	p.register_used_import(short_import_name)
 }
 
--- a/vlib/v/parser/tests/import_alias_selective_fn_used.vv
+++ b/vlib/v/parser/tests/import_alias_selective_fn_used.vv
@@ -1,0 +1,3 @@
+import os as vos { execute }
+
+execute('ls')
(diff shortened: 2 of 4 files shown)


spytheman (approved):
Excellent work.

Issue #21617, opened 2024-06-01
sum type of result causes compiler error


shuangsilab:
### Describe the bug

sum type of result causes compiler error

### Reproduction Steps

```
type ResultInt = !i32 | !i64

fn main() {
	
}
```

### Expected Behavior

The compiler accepts or refuses this code

### Current Behavior

```
PS C:\Users\LENOVO\Documents\VSCode\v\src> v run main.v
==================
C:/Users/LENOVO/AppData/Local/Temp/v_0/main.01HZ8SQG3K8AJAM7WSX44G9S47.tmp.c:514: warning: WINVER redefined
C:/Users/LENOVO/AppData/Local/Temp/v_0/main.01HZ8SQG3K8AJAM7WSX44G9S47.tmp.c:993: error: field '_i32' has incomplete type
...
==================
(Use `v -cg` to print the entire error message)

builder error:
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```

### V version

V 0.4.6 19c08f5

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.6 19c08f5
OS: windows, Microsoft Windows 10 Pro v19045 64-bit    
Processor: 8 cpus, 64bit, little endian, 

getwd: C:\Users\LENOVO\Documents\VSCode\v\src
vexe: C:\Program Files\v\v.exe
vexe mtime: 2024-06-01 02:04:25

vroot: contains spaces, value: C:\Program Files\v      
VMODULES: OK, value: C:\Users\LENOVO\.vmodules
VTMP: OK, value: C:\Users\LENOVO\AppData\Local\Temp\v_0

Git version: git version 2.40.1.windows.1
Git vroot status: weekly.2024.22-18-g19c08f5e
.git/config present: true

CC version: cc (Rev3, Built by MSYS2 project) 14.1.0
thirdparty/tcc status: thirdparty-windows-amd64 b425ac82
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Delta456:
This should be a checker error.

PR #21620, opened 2024-06-01
 checker: disallow sum type with `Result` variants

Delta456:
Fixes https://github.com/vlang/v/issues/21617


spytheman (approved):
Good work.

Issue #21607, opened 2024-05-30
regex.regex_opt doesn't handle \x escape codes


zadeviggers:
### Describe the bug

regex.regex_opt doesn't handle \x escape codes

### Reproduction Steps

This regex should strip terminal colour codes,
```v
mut remove_colours_re := regex.regex_opt(r'\x1b\\[[0-9;]*[a-zA-Z]') or { panic(err) }
```

### Expected Behavior

The regex works properly and strips the colour codes without erroring

### Current Behavior

It errors at runtime with this message:
```
query: \x1b\\[[0-9;]*[a-zA-Z]
err  : --^
ERROR: err_syntax_error
```

### V version

V 0.4.6 a4afcba

### Environment details (OS name and version, etc.)

V full version: V 0.4.6 6b2d527.a4afcba
OS: macos, macOS, 14.5, 23F79
Processor: 8 cpus, 64bit, little endian, Apple M2

CC version: Apple clang version 15.0.0 (clang-1500.3.9.4)
thirdparty/tcc: N/A
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

penguindark:
I'll check this in the next days.

penguindark:
I've made a PR with the \x \X implementation, I Hope it will be merged ASAP.

PR #21632, opened 2024-06-01
Regex: implementation of the \x \X BSLS, issue #21607

penguindark:
**What's inside**
Implementation on the `\x` and `\X` BSLS

- regex added `\xAF` and `\X1234` BLSL match
- `\x12` match a byte with the hex value 0x12
- `\X1234` match two consecutive bytes with the hex values 0x12 0x34
- more tests
- some code cleaning
- solve the issue #21607


spytheman (approved):
Excellent work.

PR #21647, opened 2024-06-06
pref: fix regression of command flags not working

ttytm:
Fixes a regression where it's not possible to run some V commands with flags anymore after #21391. For example, if there is directory with the same name as the command, like having a `fmt` directory in the in the cwd and trying to run v fmt -diff file.v.

```sh
❯ v fmt -diff file.v
Unknown argument `-diff` for command `fmt`
```

### DIFF
--- a/vlib/v/pref/pref.v
+++ b/vlib/v/pref/pref.v
@@ -964,9 +964,6 @@ pub fn parse_args_and_show_errors(known_external_commands []string, args []strin
 				}
 				if command_idx < i && (res.is_vsh || (is_source_file(command)
 					&& command in known_external_commands)) {
-					if is_source_file(command) && os.is_dir(command) {
-						eprintln('Found `${command}` directory the in the working directory.\nUse `v ${command}/` to target it instead of running the V command.')
-					}
 					// When running programs, let them be responsible for the arguments passed to them.
 					// E.g.: `script.vsh cmd -opt` or `v run hello_world.v -opt`.
 					// But detect unknown arguments when building them. E.g.: `v hello_world.v -opt`.
--- a/vlib/v/pref/pref.v
+++ b/vlib/v/pref/pref.v
@@ -969,10 +969,6 @@ pub fn parse_args_and_show_errors(known_external_commands []string, args []strin
 					// But detect unknown arguments when building them. E.g.: `v hello_world.v -opt`.
 					continue
 				}
-				if command == 'doc' {
-					// Allow for `v doc -comments file.v`
-					continue
-				}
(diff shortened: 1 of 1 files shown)


JalonSolov:
What happens if you have a `fmt` directory that you want to compile?

ttytm:
> What happens if you have a `fmt` directory that you want to compile?

Requires
```
v fmt/
```
Shlash

JalonSolov:
So _force_ people to type the `/` at the end, when they don't normally need to do that.

ttytm:
Yep.

How else would you solve that?

If you have a builtin v command and there is a directory with the same name?

JalonSolov:
Ok, try this.  Suppose I'd been working on a program named `bleh`, that was in directory `bleh`.  I've been working on this program for a long time.

Then one day I do `v up` and V has added a new `bleh` command.

Suddenly my build script, my testing, etc., all fail to work any longer, because `v bleh` now runs the new command instead of building my directory.

I would be very... unamused.

ttytm:
Again, how else would you solve it?

ttytm:
The regression might be more unamusing.

JalonSolov:
It would have been better if V had required the `/` from the start.  Then there would be no problem.  Or requiring the `build` command, such as `v build fmt`.

As it is, no matter which way you do this, there can be problems.  I think it might be nicer to the user to at least warn them of the ambiguity... perhaps tell them the alternatives.

ttytm:
Something like an info message when there is a directory with the same name to target it with the `/`?

JalonSolov:
Yes.  There shouldn't be any surprises.  It's ok if the command "wins", but it should also tell you about it instead of silently doing that.

ttytm:
Would that be okay for a separate PR? This one would just fix the regression and keep the behavior as it was.

I think we can agree on improving behavior when there is a file/dir with the same name as a command; it's just that this PR doesn't introduce a new issue. It was the way things were handled before the commands stopped working in the described way.

JalonSolov:
I don't understand... wouldn't it be as simple as adding a notice before the `continue` in your changes?

If it's more involved than that, then yes, a separate PR would work... as long as it's not forgotten.

ttytm:
Yes, adding it is that simple. Just felt that would be a separate change.

ttytm:
Reverting adding the print here as it impacts other tests.
I think having it in a dedicated scopes can go with least friction.
Then if and what to add, can be handled separately.

spytheman (approved):
It would need a test to prevent future regressions, but currently I can not see a good way to do it in a stable and portable way.

Good work.

PR #21649, opened 2024-06-06
vpm: don't write logfiles in CI runs

ttytm:
since they are never accessed

### DIFF
--- a/cmd/tools/vpm/settings.v
+++ b/cmd/tools/vpm/settings.v
@@ -29,21 +29,18 @@ fn init_settings() VpmSettings {
 	cmds := cmdline.only_non_options(args)
 
 	vmodules_path := os.vmodules_dir()
-	no_inc_env := os.getenv('VPM_NO_INCREMENT')
-	dbg_env := os.getenv('VPM_DEBUG')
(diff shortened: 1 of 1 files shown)


cmd/tools/vpm/settings.v
```diff
@@ -31,11 +31,16 @@ fn init_settings() VpmSettings {
 	vmodules_path := os.vmodules_dir()
 	no_inc_env := os.getenv('VPM_NO_INCREMENT')
 	dbg_env := os.getenv('VPM_DEBUG')
+	is_ci := os.getenv('CI') != ''
 
 	mut logger := &log.Log{}
-	if dbg_env != '0' {
+	if (dbg_env != '0' && !is_ci) || (is_ci && dbg_env !in ['', '0']) {
```
spytheman, line 37: That seems a bit complex, for what it does.
Can you please split it into 2 separate independent sequential, and non nested conditional statements:
```
1) if ... { logger.set_level(.debug) }
2) if ... { os.mkdir...; logger.set_output_path... }
```
The logic for each can be a bit duplicated, but imho will be clearer to follow, without needing the extra comments.

ttytm: Thanks for taking over on the update, I'm short on atm.

PR #21650, opened 2024-06-06
v.utils: allow to set names of compared items when diffing strings

ttytm:
Allows further customization when comparing strings.

It adds the ability to set the names that are used for the temporary files when comparing strings directly.

Two main reason:

- Some names in the output can fit better
- Several diff viewers colorize based on on the file extension. E.g. when using delta or piping output like `v fmt -diff file.v | diff-so-fancy` or using other viewers set via `VDIFF_CMD`.

<details>

Filename samples:
|Current default (not customizable)|Updated default|Updated names passed|
|-|-|-|
|Screenshot_20240606_045034|Screenshot_20240606_035334|Screenshot_20240606_035502|
  

E.g. using vs not using file extensions:

```v
import v.util.diff

fn main() {
	txt1 := '
sum = num1 + num2
print("The sum of {0} and {1} is {2}".format(num1, num2, sum))
	'
	txt2 := '
sum = num1+num2
print("The sum of {0} and {1} is {2}".format(num1, num2, sum))
	'
	println(diff.compare_text(txt1, txt2)!)
	// println(diff.compare_text(txt1, txt2, base_name: 'old.py', target_name: 'new.py')!)
}
```
|filename not set|colored based on filename|
|-|-|
|Screenshot_20240606_035942|Screenshot_20240606_040000|

Suggestions for other new default names than `base` and `target` are welcome. Those should be general names similar to `text1` and `text2` since it's not always a comparison of e.g. `expected` and `output`.

</details>

### DIFF
--- a/vlib/v/util/diff/diff.v
+++ b/vlib/v/util/diff/diff.v
@@ -24,6 +24,14 @@ pub:
 	env_overwrite_var ?string = 'VDIFF_CMD'
 }
 
+@[params]
+pub struct CompareTextOptions {
+	CompareOptions
+pub:
+	base_name string = 'base'
+	target_name string = 'target'
+}
+
 // Default options for `diff` and `colordiff`.
 // Short `diff` args are supported more widely (e.g. on OpenBSD, ref. https://man.openbsd.org/diff.1).
 // `-d -a -U 2` ^= `--minimal --text --unified=2`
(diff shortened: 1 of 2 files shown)


spytheman (approved):

spytheman:
Can you please make a test for it?

PR #21655, opened 2024-06-07
Use "segoeui.ttf" font if exist on Windows

malisipi:
The PR will change default font to "segoeui.ttf" on Windows. To support legacy Windows versions, it will use Arial font when "segoeui.ttf" is not exist.

It's important to make UI consistent between system and V applications.


spytheman:
(the ci failure is not related)

spytheman (approved):

PR #21661, opened 2024-06-09
cgen: fix or expr with non option fn call return

### LINKED ISSUE #21660: Compiler bug when trying to return an interface-optional with a fallback default value
### Describe the bug

```
/tmp/v_1000/v-option-bug.01HZZ86M38ET1YCNSFHJVPHZQ6.tmp.c:13154: error: cannot convert 'struct src__example__IFoo' to 'struct _option_src__example__IFoo'
builder error: 
==================
```

The function in question tries to return a result of the interface directly from the struct, which itself is an optional.  If the struct value is not set it invokes a function to create a default.  The produced C code does not seem to realize that the value is no longer an optional.

### Reproduction Steps

There is a full reproducer here:
https://github.com/dasidiot/v-option-bug

```bash
~/src/v-option-bug$ v -cg run .
/tmp/v_1000/v-option-bug.01HZZ86M38ET1YCNSFHJVPHZQ6.tmp.c:13154: error: cannot convert 'struct src__example__IFoo' to 'struct _option_src__example__IFoo'
builder error: 
==================

### Expected Behavior

The program should compile without an error.

### Current Behavior

The C compiler fails to convert from the optional form of the interface to the concrete form of an implementation.

### Possible Solution

_No response_

_No response_

### V version

Current V version: V 0.4.6 1af7b7c, timestamp: 2024-06-09 17:52:56 +0300

```
getwd: /home/user/src/v
vexe: /home/user/src/v/v
vexe mtime: 2024-06-09 19:42:11

vroot: OK, value: /home/user/src/v
VMODULES: OK, value: /home/user/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.39.2
Git vroot status: weekly.2024.22-47-g1af7b7c3
.git/config present: true

```
> [!NOTE]
>

### PR DESCRIPTION
**Pull request by felipensp**
Fix #21660

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -6956,7 +6956,7 @@ fn (mut g Gen) gen_or_block_stmts(cvar_name string, cast_typ string, stmts []ast
 							&& g.table.final_sym(return_type).kind == .array_fixed
 						if !is_array_fixed {
 							if g.inside_return && !g.inside_struct_init
-								&& expr_stmt.expr is ast.CallExpr
+								&& expr_stmt.expr is ast.CallExpr&& (expr_stmt.expr as ast.CallExpr).return_type.has_option_or_result()
 								&& g.cur_fn.return_type.has_option_or_result()
 								&& return_type.has_option_or_result()
 								&& expr_stmt.expr.or_block.kind == .absent {
--- a/vlib/v/tests/option_or_expr_with_non_opt_test.v
+++ b/vlib/v/tests/option_or_expr_with_non_opt_test.v
@@ -1,0 +1,40 @@
+pub struct Foo {
+	a int
+	b int
+}
+
+pub fn (f Foo) hello() ! {
+	println('Hello')
+}
+
+fn default_foo() IFoo {
+	return Foo{}
+}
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: vlib/v/gen/c/cgen.v [hunk 1]
```diff
@@ -6956,7 +6956,7 @@ fn (mut g Gen) gen_or_block_stmts(cvar_name string, cast_typ string, stmts []ast
 							&& g.table.final_sym(return_type).kind == .array_fixed
 						if !is_array_fixed {
 							if g.inside_return && !g.inside_struct_init
-								&& expr_stmt.expr is ast.CallExpr
+								&& expr_stmt.expr is ast.CallExpr&& (expr_stmt.expr as ast.CallExpr).return_type.has_option_or_result()
```
spytheman, line 6959: Hm, that `ast.CallExpr&& (` seems a bit strangely formatted 🤔 .

spytheman (approved):
Good work.

vlib/v/gen/c/cgen.v
felipensp: Yeah. It seems a vfmt bug.

PR #21664, opened 2024-06-11
crypto: add a `crypto.sha3` hash and extended output functions

kimshrier:
This is a pure v implementation of the sha3 functions in the FIPS 202 specification.


spytheman (approved):

vlib/crypto/sha3/sha3.v
```diff
@@ -0,0 +1,362 @@
...
+		d.s.kaccak_p_1600_24()
+		bytes_remaining = unsafe { bytes_remaining[d.rate..] }
+	}
+
+	if bytes_remaining.len > 0 {
+		d.input_buffer = bytes_remaining
+	}
+
+	return
+}
```
spytheman, line 261: That seems superfluous.

JalonSolov: Perhaps something for the `transform` step to deal with?

PR #21670, opened 2024-06-12
builtin: fix string.find_between(), when not found end string return ''

kbkpbot:
`find_between` returns the string found between start string and end string.

When `end` string not found, should return '', not the rest of string.


spytheman:
Please add a test case for it too.

spytheman (approved):
Good work.

PR #21679, opened 2024-06-15
builtin.string: add CamelCase to/from snake_case

kbkpbot:
`camel_to_snake` convert string from camelCase to snake_case
`snake_to_camel` convert string from snake_case to camelCase

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -2614,32 +2614,31 @@ pub fn (s string) is_ascii() bool {
 // Example: assert 'aa_BB'.camel_to_snake() == 'aa_bb'
 @[direct_array_access]
 pub fn (s string) camel_to_snake() string {
-	mut i := 0
-	mut idx := 0
 	if s.len == 0 {
 		return ''
 	}
 	mut prev_is_upper := false
 	mut prev_char := ` `
-	mut c := `_`
 	mut lower_c := `_`
 	mut c_is_upper := false
-	unsafe {
-		mut b := malloc_noscan(2 * s.len + 1)
-		for {
-			c = s.str[idx]
-			c_is_upper = c >= `A` && c <= `Z`
-			lower_c = if c_is_upper { c + 32 } else { c }
-			if i > 0 {
-				if prev_is_upper == false && c_is_upper {
-					// aB => a_b, if prev has `_`, then do not add `_`
+	mut b := unsafe { malloc_noscan(2 * s.len + 1) }
+	mut i := 0
+	for c in s {
+		c_is_upper = c >= `A` && c <= `Z`
+		lower_c = if c_is_upper { c + 32 } else { c }
+		if i > 0 {
+			if prev_is_upper == false && c_is_upper {
+				// aB => a_b, if prev has `_`, then do not add `_`
+				unsafe {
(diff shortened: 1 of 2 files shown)


Delta456:
Do we need this?

vlib/builtin/string.v
```diff
@@ -2606,3 +2606,101 @@ pub fn (name string) match_glob(pattern string) bool {
...
+	}
+	mut prev_is_upper := false
+	mut prev_char := ` `
+	mut c := `_`
+	mut lower_c := `_`
+	mut c_is_upper := false
+	unsafe {
+		mut b := malloc_noscan(2 * s.len + 1)
+		for {
+			c = s.str[idx]
```
spytheman, line 2630:
```suggestion
        for c in s {
```
then the code for `idx` and initializing `c` can be deleted

vlib/builtin/string.v
```diff
@@ -2606,3 +2606,101 @@ pub fn (name string) match_glob(pattern string) bool {
...
+	if s.len == 1 {
+		return s
+	}
+	mut need_upper := true
+	mut c := `_`
+	mut upper_c := `_`
+	unsafe {
+		mut b := malloc_noscan(s.len + 1)
+		for {
+			c = s[idx]
```
spytheman, line 2686: same cleanup as above can be applied:
```suggestion
		for c in s {
```

vlib/builtin/string.v
```diff
@@ -2606,3 +2606,101 @@ pub fn (name string) match_glob(pattern string) bool {
...
 }
+
+// camel_to_snake convert string from camelCase to snake_case
+// Example: assert 'Abcd'.camel_to_snake() == 'abcd'
+// Example: assert 'aaBB'.camel_to_snake() == 'aa_bb'
+// Example: assert 'BBaa'.camel_to_snake() == 'b_baa'
+// Example: assert 'aa_BB'.camel_to_snake() == 'aa_bb'
+@[direct_array_access]
+pub fn (s string) camel_to_snake() string {
+	mut i := 0
```
spytheman, line 2617: move this closer to the loop please

vlib/builtin/string.v
```diff
@@ -2606,3 +2606,101 @@ pub fn (name string) match_glob(pattern string) bool {
...
+	mut idx := 0
+	if s.len == 0 {
+		return ''
+	}
+	mut prev_is_upper := false
+	mut prev_char := ` `
+	mut c := `_`
+	mut lower_c := `_`
+	mut c_is_upper := false
+	unsafe {
```
spytheman, line 2627: In general it is better to have multiple shorter `unsafe{}` blocks, than one large one.
The reason, is that reviewing and writing unsafe code requires more attention, since a lot of the compiler checks are turned off in such blocks.

Please split that big block into several, for example:
`mut b := unsafe { malloc_noscan(2 * s.len + 1) }`
etc

spytheman:
> Do we need this?

I think they are good additions.

spytheman (approved):

autf:
I noticed that the string `ThisIsATestCall` is being converted to `this_is_at_est_call`. 

Is this behavior unavoidable due to the automatic processing, or is it a bug?

PR #21680, opened 2024-06-15
cgen: generate pragmas for gcc-14, to make it more relaxed by default. Support `-no-relaxed-gcc14` for turning them off.

spytheman:


spytheman:
before/after on Arch:
```
![image](https://github.com/vlang/v/assets/26967/bcb6c946-96a7-4286-a2d1-798b0be309a3)

PR #21684, opened 2024-06-15
all: add support for alias bool assign operators 

Delta456:
Implements `&&=` and `||=` for alias with bool parent type

```v
type Bool = bool

fn test_alias_bool_assign_operator() {
	mut flag := Bool(true)
	flag = flag || false
	assert flag == true

	flag ||= false
	assert flag == true

	flag &&= false
	assert flag == false
}
```


spytheman (approved):
Excellent work.

PR #21685, opened 2024-06-16
comptime: support `-d ident=value` and `var := $d('ident', 0)`

larpon:
**Updated**: Initial suggestion was to use `-cv ident=value` and `$compile_value('ident', <value>)`, it is now an extension of `-d` using `$d('ident', <value>)` as retriever.

This PR adds support for the following:

`v -d my_i64=890 -d my_string="five six" run .`

```v
const my_i64 = $d('my_i64',4)
const my_string = $d('my_string','five')

fn main() {
    println(my_i64)
    println(my_string)
}
```
Resulting in:
```
890
five six
```
.. and leaving out the `-d` flags:
`v run .`, outputs:
```
4
five
```
Supported types are "pure" literals, and can be used also without `const`:
```v
my_f64 := $d('my_f64', 1.0)
my_i64 := $d('my_i64', 2)
my_string := $d('my_string', 'three')
my_bool := $d('my_bool', false)
my_char := $d('my_char', `f`)
```

I'm suggesting this, primarily, because I need to be able to tweak fixed array sizes at compile time - so that they have a default, but so I can decide to let my users tweak the sizes if they have reasons or needs to do so.

As @spytheman mentioned on Discord; It has potential to replace both (current) `-d` and `$env`.

I plan on adding support for `#flag -I $d('my_flag','flag_value')/xyz` and `#include "$d('my_include','/usr/include')/stdio.h"` when time permits. (**DONE**)


.github/workflows/comptime_ci.yml
```diff
@@ -0,0 +1,51 @@
+name: Time CI
```
StunxFS, line 1:
```suggestion
name: Comptime CI
```
Wouldn't that be a more appropriate name?

.github/workflows/comptime_ci.yml
```diff
@@ -0,0 +1,51 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build V
+        run: make
+      - name: Test -cv values
+        run: ./v -cv my_f64=2.0 -cv my_int=3 -cv my_string="a four" -cv my_bool=true -cv my_char=g run vlib/v/checker/tests/comptime_value/use_flag_comptime_values.vv
+      - name: Test -compile-value values
+        run: ./v -compile-value my_f64=2.0 -compile-value my_int=3 -compile-value my_string="a four" -compile-value my_bool=true -compile-value my_char=g run vlib/v/checker/tests/comptime_value/use_flag_comptime_values.vv
+
+  test-time-windows:
```
StunxFS, line 42:
```suggestion
  test-comptime-windows:
```
the same here

StunxFS (commented):
Great job!

.github/workflows/comptime_ci.yml
larpon: Indeed, copy-pasta error 😄

.github/workflows/comptime_ci.yml
larpon: Good catch

vlib/v/pref/pref.v
```diff
@@ -193,6 +193,8 @@ pub mut:
 	// -d vfmt and -d another=0 for `$if vfmt { will execute }` and `$if another ? { will NOT get here }`
 	compile_defines     []string // just ['vfmt']
 	compile_defines_all []string // contains both: ['vfmt','another']
+	// -compile_value key=value
+	compile_compile_values map[string]string
```
spytheman, line 197: `compile_values` is less repetitive

larpon: Indeed 🤔 might be a search/replace whoopsie (it was initially called `dval`)

vlib/v/pref/pref.v
```diff
@@ -1170,6 +1178,15 @@ pub fn cc_from_string(s string) CompilerType {
 	}
 }
 
+fn (mut prefs Preferences) parse_compile_value(define string) {
+	define_parts := define.split('=')
```
spytheman, line 1182: That will fail for `s="y=b=c"`.

Something like this may be more robust, without using a full state machine:
if `=` !in define {
     eprintln_exit('V error: Define argument value missing for ${define}.')
}
name := define.all_before('=')
value := define.all_after_first('=')
prefs.compile_compile_values[ name ] = value

vlib/v/fmt/fmt.v
```diff
@@ -2186,6 +2186,15 @@ pub fn (mut f Fmt) comptime_call(node ast.ComptimeCall) {
...
 			}
+			node.method_name == 'compile_value' {
+				if node.args_var.contains("'") {
+					f.write('\$${node.method_name}("${node.args_var}", ')
+				} else {
+					f.write("\$${node.method_name}('${node.args_var}', ")
+				}
+				f.expr(node.args[0].expr)
+				f.write(')')
+			}
```
spytheman, line 2197: Why distinguish between ' and " here?
It is used only for the name of the value, which is always an identifier.
Is there a case where you would want a " or a ' in the name?

vlib/v/checker/tests/comptime_value/use_flag_comptime_values.vv
```diff
@@ -0,0 +1,15 @@
...
+const my_bool = $compile_value('my_bool', false)
+const my_char = $compile_value('my_char', `f`)
+
+fn main() {
+	assert my_f64 == 2.0
+	assert my_int == 3
+	assert my_string == 'a four'
+	assert my_bool == true
+	assert my_char == `g`
+}
```
spytheman, line 15: move this file in `vlib/v/gen/c/testdata/`, then add a comment at the top:
```
// vtest vflags: -cv my_f64=2.0 -cv my_int=3 -cv my_string="a four" -cv my_bool=true -cv my_char=g
```
then add in the same folder, a file named: `use_flag_comptime_values.out` .

Then, it will be tested automatically by `v vlib/v/gen/c/coutput_test.v` which is run by `v test-self`, and there will be no need for the separate .yml job.

It will be also easier to test locally.

.github/workflows/comptime_ci.yml
```diff
@@ -0,0 +1,51 @@
+name: Comptime CI
```
spytheman, line 1: This file can be eliminated completely. See the comment for `use_flag_comptime_values.vv`.

spytheman:
It is an excellent and long needed addition.
Great work!

larpon:
Awesome, thanks @spytheman

enghitalo:
What you guys think about use `--compile-values` instead of `-cv` flag? Or using both?

JalonSolov:
It would be `-compile-values` - V only uses 1 `-`.

The longer name would be more explicit, for sure.  `-cv` could be almost anything... Want V to write a Ciriculam Vitae for you?  `-cv`.  :-)

larpon:
> What you guys think about use `--compile-values` instead of `-cv` flag? Or using both?

Currently most, if not all, v flags is denoted via single dash `-`.
The PR supports both a short and long version:
`-cv` and `-compile-value`

spytheman (approved):
@medvednikov what do you think about this new language feature?

medvednikov:
I like it.

But we do have `-d compvar`, which does the same thing, so it would probably have to be removed for "one way"?

medvednikov:
(well, not the same,  -d is a subset of this new feature)

larpon:
If we decide to remove `-d` it probably needs an extended deprecation period since it's a very "old" and very frequently used feature.

spytheman:
We can also extend `-d ident` to support `-d ident=value`, then it can work as before for the first case (except that `$compile_value(ident, false)` will be true).

For the second case, it can have the new behavior, that is described here for `-cv ident=value`.

That way, there will be no need for a deprecation, it will continue to work, while having more functionality.

felipensp:
> We can also extend `-d ident` to support `-d ident=value`, then it can work as before for the first case (except that `$compile_value(ident, false)` will be true).
> 
> For the second case, it can have the new behavior, that is described here for `-cv ident=value`.
> 
> That way, there will be no need for a deprecation, it will continue to work, while having more functionality.

+1 on keeping `-d` parameter.

spytheman:
gcc/clang's `-D` option also has 2 forms:
`-Dname`
`-Dname=value`
so there is an existing precedent for that.

spytheman:
It seems to work, but I have to do more tests:
```
#0 15:56:47 ^ comptime/key-value /v/oo>./v -d dynamic_boehm -keepc -showcc examples/hello_world.v
> C compiler cmd: '/space/v/oo/thirdparty/tcc/tcc.exe' '@/tmp/v_1000/hello_world.tmp.c.rsp'
> C compiler response file "/tmp/v_1000/hello_world.tmp.c.rsp":
  -fwrapv -o "/space/v/oo/examples/hello_world" -D GC_THREADS=1 -I "/usr/local/include" -L "/usr/local/lib" "/tmp/v_1000/hello_world.tmp.c" -std=gnu99 -D_DEFAULT_SOURCE -bt25 -lgc -lpthread -ldl
#0 15:56:52 ^ comptime/key-value /v/oo>./v -keepc -showcc examples/hello_world.v
> C compiler cmd: '/space/v/oo/thirdparty/tcc/tcc.exe' '@/tmp/v_1000/hello_world.tmp.c.rsp'
> C compiler response file "/tmp/v_1000/hello_world.tmp.c.rsp":
  -fwrapv -o "/space/v/oo/examples/hello_world" -D GC_BUILTIN_ATOMIC=1 -D GC_THREADS=1 -I "/space/v/oo/thirdparty/libgc/include" "/tmp/v_1000/hello_world.tmp.c" -std=gnu99 -D_DEFAULT_SOURCE -bt25 "/space/v/oo/thirdparty/tcc/lib/libgc.a" -ldl -lpthread
#0 15:56:59 ^ comptime/key-value /v/oo>
#0 15:57:00 ^ comptime/key-value /v/oo>
#0 15:57:00 ^ comptime/key-value /v/oo>./v -d my_f64=2.0 -d my_int=3 -d my_string='a four' -d my_bool -d my_char=g run ./vlib/v/gen/c/testdata/use_flag_comptime_values.vv
2.0
3
a four
true
g
#0 15:57:04 ^ comptime/key-value /v/oo>
```

larpon:
I'm ok with keeping `-d` but then the name `$compile_value` will be less fitting IMO. Then maybe `$compile_define` would be a better name?

felipensp:
> I'm ok with keeping `-d` but then the name `$compile_value` will be less fitting IMO. Then maybe `$compile_define` would be a better name?

Maybe `$get()`, `$ct_define()`, `$def()` or `$compile_define`. 👍🏻

JalonSolov:
Why not just `-d name=value` and forget about `$<anything>`?

If `-d name` continues to work for booleans, and `-d name=value` works for everything else...

spytheman:
$define looks nice to me

larpon:
> Why not just `-d name=value` and forget about `$<anything>`?
> 
> If `-d name` continues to work for booleans, and `-d name=value` works for everything else...

If `$<anything>` is removed how do we use the value it carries from the `-d x=y` assignment?

We need to be able to use it as consts or literals in the code

spytheman:
`$compile_define` is also nice, since we already have other `$compile_` functions, like `$compile_warn` etc.

larpon:
> `$define` looks nice to me, but it can be indeed mistaken with `#define`, which we also support.
> 
> Perhaps `$value('name', 123)` ?
> 
> Or `$d('name', 123)` - that will mirror the user side `-d name=value` ?

+1 for `$d`.

I also thought of that after @felipensp s suggestions also.

spytheman:
> Why not just `-d name=value` and forget about `$<anything>`?

The question is how to access the value in the source.

Currently that is with `$compile_value('name', some_default)` in this PR.

StunxFS:
So, `$if` would remain the same? Or would it be different?

```v
// v run main.v -d say_hi=true

$if say_hi ? {
    println('hi!')
}

// or

$if $compile_value('say_hi', false) {
    println('hi!')
}
```

JalonSolov:
When I suggested `-d name=value`, I actually meant for the command line option.  Instead of introducing `-cv` or `-compile-value` or `-compile-define`.

If they code uses `$compile_define()` or `$compile_value()`, that's fine by me.  I'd be partial to `$compile_define` since that matches the meaning of `-d` better.

spytheman:
The PR is already updated to allow for `-d name=value`, and if everyone agrees, we can remove the code handling `-cv/-compile-value` in v.pref.

felipensp:
> > So, $if would remain the same? Or would it be different?
> 
> The syntax of $if is not affected at all by this PR (and it would be a breaking change if it did). It is still `$if name ? {`.
> 
> The new `$compile_value('name', default)` syntax allows you to get the actual passed value, which was not possible before.
> 
> I think, that it would be too verbose to be used as `$if $compile_value('say_hi', false) {`, even when that is possible (it currently is not evaluated in that case, which is a bug):
> 
> ```v
> $if $compile_value('say_hi', false) {
>         println('say_hi is true')
> } $else {
>         println('say_hi is false')
> }
> ```
> 
> always prints `say_hi is true`, even for `-d say_hi=false` .

yes, `$compile_value()` is too verbose on if conditions. We just need to document about using -d to handling conditional code and accessing its values. How retrive value and how use it on conditions.

larpon:
I'm ok with `-d ident=value` and like any of `$d` `$define` and `$compile_value`. `$d` is very nice since it matches `-d`.

Maybe it will be possible to do:
`$if name == value ? {...}`
For comptime logic?

larpon:
Limited hashStmt support is added in [e1be3eb](https://github.com/vlang/v/pull/21685/commits/e1be3eb263bd6ebf6da1fd5d5129eaef698ed7b4). Currently we only support `string` values like with `$env()`.

The whole `#` comptime setup could probably benefit from some proper refactoring at some point.

But now this works:
```v
#flag -I $d('my_flag','flag_value')/xyz
#include "$d('my_include','/usr/include')/stdio.h"
```

larpon:
Resolved conflict with `master`, hence the merge with `master`

doc/docs.md
```diff
@@ -6120,6 +6120,56 @@ V can bring in values at compile time from environment variables.
...
+provided as the *second* argument.
+
+Supported default values are the following *pure* literals:
+```v
+fn main() {
+	val_str := $d('id_str', 'value') // can be changed by providing `-d id_str="my id"`
+	val_f64 := $d('id_f64', 42.0) // can be changed by providing `-d id_f64=84.0`
+	val_i64 := $d('id_i64', 56) // can be changed by providing `-d id_i64=123`
+	val_bool := $d('id_bool', false) // can be changed by providing `-d id_bool=true`
+	val_char := $d('id_char', `f`) // can be changed by providing `-d id_char=v``
```
StunxFS, line 6160:
```suggestion
	val_char := $d('id_char', `f`) // can be changed by providing `-d id_char=v`
```

larpon: Whoopsie :smile: , fixed

vlib/v/util/util.v
```diff
@@ -115,6 +115,71 @@ pub fn resolve_env_value(str string, check_for_presence bool) !string {
...
+		all_parsed += ch.ascii_str()
+		if !(ch == `'`) {
+			d_default_value += ch.ascii_str()
+		}
+		if ch == `$` {
+			return error('cannot use string interpolation in compile time \$d() expression')
+		}
+	}
+	if d_default_value == '' {
+		return error('second argument of `\$d` must be a string identifier')
```
larpon, line 170:
```suggestion
		return error('second argument of `\$d` must be a pure literal')
```

PR #21691, opened 2024-06-17
builtin: improve camel_to_snake

ttytm:
The PR improves the performance of the method by removing the `i > 0 and i > 1` checks from the loop.


spytheman (approved):
Good work.

PR #21693, opened 2024-06-18
veb.auth: use constant time compare

Le0Developer:
also added small section about security considerations

### DIFF
--- a/vlib/veb/auth/README.md
+++ b/vlib/veb/auth/README.md
@@ -92,8 +92,9 @@ pub fn (mut app App) find_user_by_name(name string) ?User {
 
 ## Security considerations
 
-`hash_password_with_salt` and its related functions use `sha256` for hashing with a single iteration.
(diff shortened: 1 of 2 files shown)


spytheman:
There is no `compare_digest` function. The CI fails because of it.

vlib/veb/auth/README.md
```diff
@@ -89,3 +89,12 @@ pub fn (mut app App) find_user_by_name(name string) ?User {
 	return User{}
 }
 ```
+
+## Security considerations
+
+`hash_password_with_salt` and related use `sha256` for hashing with a single iteration.
```
stickyburn, line 95:
```suggestion
`hash_password_with_salt` and its related functions use `sha256` for hashing with a single iteration.
```

nitpick, but i think it makes it clearer :)

spytheman (approved):

PR #21697, opened 2024-06-18
orm: update readme with more specifics

splashsky:
While working on my project I found that time.Time has non-obvious behavior when creating tables in the ORM. Some additional specificity regarding passing the structs as the table was appreciated, and so the expression examples have been updated to be more explicit about this.

Some wording was updated as well to make concepts more obvious.

This PR only updates the README file in `vlib/orm`, so no tests are necessary.

### DIFF
--- a/vlib/orm/README.md
+++ b/vlib/orm/README.md
@@ -1,31 +1,44 @@
 # ORM
 
-## Null
+V has a powerful, concise ORM baked in! Create tables, insert records, manage relationships, all regardless of the DB
+driver you decide to use.
 
-Use option fields in V structs for fields which can be NULL.  Regular,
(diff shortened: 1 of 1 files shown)


vlib/orm/README.md
```diff
@@ -47,23 +60,55 @@ struct Child {
 }
 ```
 
-### Create
+To use the ORM, there is a special interface that lets you use the structs and V itself in queries.
```
spytheman, line 63:
```suggestion
To use the ORM, there is a special expression that lets you use the structs and V itself in queries.
```
The word `interface` has a special meaning in V, and imho can be a bit confusing for the readers.

spytheman (approved):
Good work.

PR #21698, opened 2024-06-19
checker: update checker.v

eltociear:
explictly -> explicitly


spytheman:
You will need to run `VAUTOFIX=1 ./v vlib/v/compiler_errors_test.v` 2 times in a row, to fix the .out files too, and commit the results.

PR #21701, opened 2024-06-19
comptime, ast: support `[$d('s', 4)]int{}`, move resolving to method on `ComptimeCall`

larpon:
Add support for `my_fixed_size_array := [$d('my_size', 4)]int{}`.

... also moved `is_escape_sequence/1` from `scanner` to `util` for reuse (if that is not ok it can be moved back)


spytheman (approved):
Excellent work.

PR #21704, opened 2024-06-19
cgen,x.json2: fix generic map decoding

### LINKED ISSUE #21683: x.json2 fails to decode a map with number keys
**V doctor:**
```
V full version: V 0.4.6 c689f80.3ad77b2
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-06-15 17:29:20

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.43.2
Git vroot status: weekly.2024.24-11-g3ad77b25
.git/config present: true

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg j4.v`
```v
module main

import x.json2

const data = '
{
  "comments": {
    "26788945": {
      "id": "26788945",
      "message": "some comment 1"
    },
    "26788946": {
      "id": "26788946",
      "message": "some comment 2"
    },
    "26788947": {
      "id": "26788947",
      "message": "some comment 3"
    }
  }
}
'

struct Comment {
	id      string
	message string
}

struct Comments {
mut:
	comments map[string]Comment
}

fn main() {
	mut root := json2.decode[Comments](data)!
	println(root)
}

```

A non empty structure

**What did you see instead?**
```
Comments{
    comments: {}
}
```
> [!NOTE]
>

felipensp:
Fix #21683

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -4606,8 +4606,8 @@ fn (mut c Checker) index_expr(mut node ast.IndexExpr) ast.Type {
 			info := typ_sym.info as ast.Map
 			c.expected_type = info.key_type
 			index_type := c.expr(mut node.index)
-			if !c.check_types(index_type, info.key_type) {
-				err := c.expected_msg(index_type, info.key_type)
+			if !c.check_types(index_type, c.unwrap_generic(info.key_type)) {
+				err := c.expected_msg(index_type, c.unwrap_generic(info.key_type))
 				c.error('invalid key: ${err}', node.pos)
 			}
 			value_sym := c.table.sym(info.value_type)
@@ -4920,7 +4920,8 @@ fn (mut c Checker) ensure_type_exists(typ ast.Type, pos token.Pos) bool {
 		return false
 	}
 	sym := c.table.sym(typ)
-	if !c.is_builtin_mod && sym.kind == .struct_ && !sym.is_pub && sym.mod != c.mod {
+	if !c.is_builtin_mod && sym.kind == .struct_ && !sym.is_pub && sym.mod != c.mod
+		&& c.mod != 'x.json2' {
 		c.error('struct `${sym.name}` was declared as private to module `${sym.mod}`, so it can not be used inside module `${c.mod}`',
 			pos)
 		return false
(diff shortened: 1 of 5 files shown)


vlib/v/checker/checker.v
```diff
@@ -4606,8 +4606,8 @@ fn (mut c Checker) index_expr(mut node ast.IndexExpr) ast.Type {
 			info := typ_sym.info as ast.Map
 			c.expected_type = info.key_type
 			index_type := c.expr(mut node.index)
-			if !c.check_types(index_type, info.key_type) {
-				err := c.expected_msg(index_type, info.key_type)
+			if !c.check_types(index_type, c.unwrap_generic(info.key_type)) {
```
StunxFS, line 4609:
```suggestion
			key_type := c.unwrap_generic(info.key_type)
			if !c.check_types(index_type, key_type) {
```

I think saving the result of calling `c.unwrap_generic` in a variable is better, thus avoiding the double call

vlib/v/checker/checker.v
```diff
@@ -4920,7 +4920,8 @@ fn (mut c Checker) ensure_type_exists(typ ast.Type, pos token.Pos) bool {
 		return false
 	}
 	sym := c.table.sym(typ)
-	if !c.is_builtin_mod && sym.kind == .struct_ && !sym.is_pub && sym.mod != c.mod {
+	if !c.is_builtin_mod && sym.kind == .struct_ && !sym.is_pub && sym.mod != c.mod
+		&& c.mod != 'x.json2' {
```
spytheman, line 4924: Why special case `x.json2` here?

spytheman: As an alternative, the test can be modified to have `pub` for the structs that `x.json2` needs to access/use.

spytheman: Removing `&& c.mod != 'x.json2'` does not seem to affect `v test vlib/x/json2/`.

felipensp: > As an alternative, the test can be modified to have `pub` for the structs that `x.json2` needs to access/use.

hmm, I was getting an error about user struct not being public, I'll check again.

vlib/v/checker/checker.v
felipensp: Yeah, thanks.

spytheman (approved):
Excellent work.

Issue #21687, opened 2024-06-16
eof  ignored because of <>


Eliyaan:
### Describe the issue

`<p>input returns a one-line string from stdin, after printing a prompt. Returns '<EOF>' in case of an error (end of input).</p>`

EOF does not show because of the <>

### Links

https://modules.vlang.io/os.html#input

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Delta456:
Issue fixed https://modules.vlang.io/os.html#input

PR #21705, opened 2024-06-19
docs: fix comment considered as a token

stickyburn:
Tiny change closing #21687

### DIFF
--- a/vlib/os/os.v
+++ b/vlib/os/os.v
@@ -337,7 +337,7 @@ pub fn input_opt(prompt string) ?string {
 }
 
 // input returns a one-line string from stdin, after printing a prompt.
-// Returns '<EOF>' in case of an error (end of input).
+// Returns `'<EOF>'` in case of an error (end of input).
 pub fn input(prompt string) string {
 	res := input_opt(prompt) or { return '<EOF>' }
 	return res


Delta456:
Is this tested locally?

vlib/os/os.v
```diff
@@ -337,7 +337,7 @@ pub fn input_opt(prompt string) ?string {
 }
 
 // input returns a one-line string from stdin, after printing a prompt.
-// Returns '<EOF>' in case of an error (end of input).
+// Returns `'<EOF>'` in case of an error (end of input).
```
JalonSolov, line 340:
```suggestion
// Returns `EOF` in case of an error (end of input).
```

Delta456: I think the `<EOF>` notation is **important** here.

stickyburn:
> Is this tested locally?

It is :)

vlib/os/os.v
spytheman: it is not, and it breaks the documentation

spytheman: The problem is that the actual return value is `<EOF>`. I think that the original was better (the one that used backticks and apostrophes), as long as it produces valid HTML, when the comments are rendered at https://modules.vlang.io .

JalonSolov: If the actual, concrete return includes the `<` and `>` then yes, it's better to leave them.

It's a different discussion as to _why_ the `<` and `>` are included.  :-\

Issue #21707, opened 2024-06-20
c-style for loop with type alias provokes build error


pd-giz-dave:
### Describe the bug

A c-style for loop with a type alias variable results in a build error

### Reproduction Steps

Attempt to compile this:
```
module main
type Float = f32
// for loop with a type alias var of Float get build error
// replacing Float() with f32() works
fn main() {
	min_value := Float(1)
	max_value := Float(10)
	step := Float(1)
	for n := min_value; n <= max_value; n += step {
		print(',${n}')
	}
}
```

### Expected Behavior

I expect to it to compile

### Current Behavior

Get a build error:
```
/home/dave/v/v /home/dave/precious/fellsafe/v/compiler_bugs -color -o /home/dave/precious/fellsafe/v/bin/main
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
==================
/tmp/v_1000/main.01J0V23EWTE568C2JCCP4PK2CR.tmp.c:13147: error: ')' expected (got ";")
...
==================
(Use `v -cg` to print the entire error message)

```

### Possible Solution

Use a DIY loop

### Additional Information/Context

Replacing the type alias with a native type works.

### V version

V 0.4.6 537fa1b

### Environment details (OS name and version, etc.)

```
dave@dave-mini-pc~/.../fellsafe/v >>> v doctor
V full version: V 0.4.6 294f7e4.537fa1b
OS: linux, "EndeavourOS Linux"
Processor: 16 cpus, 64bit, little endian, AMD Ryzen 7 5800H with Radeon Graphics

getwd: /home/dave/precious/fellsafe/v
vexe: /home/dave/v/v
vexe mtime: 2024-06-20 14:51:25

vroot: OK, value: /home/dave/v
VMODULES: OK, value: /home/dave/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.45.2
Git vroot status: weekly.2024.23-43-g537fa1bb
.git/config present: true

CC version: cc (GCC) 14.1.1 20240522
thirdparty/tcc status: thirdparty-linux-amd64 40e5cbb5
dave@dave-mini-pc~/.../fellsafe/v >>>       
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
[...]

PR #21708, opened 2024-06-20
cgen: support inc cond for c style for loop with alias types 

Delta456:
Fixes https://github.com/vlang/v/issues/21707


vlib/v/slow_tests/inout/alias_for_c_loop_assign_op.vv
```diff
@@ -0,0 +1,10 @@
+type Float = f32
```
spytheman, line 1: Why is that in vlib/v/slow_tests/inout/ ?

It can be a simple _test.v file in `vlib/v/tests/` imho.

vlib/v/gen/c/assign.v
```diff
@@ -520,7 +520,9 @@ fn (mut g Gen) assign_stmt(node_ ast.AssignStmt) {
 					g.expr(left)
 					g.write(' ${extracted_op} ')
 					g.expr(val)
-					g.write(';')
+					if !g.inside_for_c_stmt {
+						g.write(';')
+					}
```
spytheman, line 525: That is short, but does not seem the right place for the fix.
Why does the non aliased case work, without such check?

vlib/v/slow_tests/inout/alias_for_c_loop_assign_op.vv
Delta456: Good idea

vlib/v/gen/c/assign.v
Delta456: It is because the assignment operator overloading here assumes to be a statement ending with the `;` but in a for c loop it is not the case. I also can't find a better place to make it work because it returns after code generation.

I tried removing the return and adding the `g.write(';')` but the compiler failed.

PR #21710, opened 2024-06-21
Perf: arrays.sum

esquerbatua:
Remove the check inside the loop, initializing value with 0

### DIFF
--- a/vlib/arrays/arrays.v
+++ b/vlib/arrays/arrays.v
@@ -220,16 +220,10 @@ pub fn sum[T](array []T) !T {
 	if array.len == 0 {
 		return error('Cannot sum up array of nothing.')
 	} else {
-		mut head := array[0]
-
+		mut head := 0
 		for i, e in array {
-			if i == 0 {
-				continue
-			} else {
-				head += e
-			}
+			head += e
 		}
-
 		return head
 	}
 }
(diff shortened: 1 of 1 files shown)


spytheman:
`mut head := 0` will work only for integers, but not for other types like Structs etc, unlike
`mut head := array[0]` .

spytheman:
Try this for example:
```v
import arrays

struct Abc {
	x int = 5
}

fn (a Abc) + (b Abc) Abc {
	return Abc{
		x: a.x + b.x
	}
}

a := [1, 2, 3, 5]
dump(arrays.sum(a)!)

s := [Abc{
	x: 30
}, Abc{
	x: 20
}, Abc{
	x: 10
}]
dump(arrays.sum(s)!)
```

printing:
```
[a.v:14] arrays.sum(a)!: 11
[a.v:23] arrays.sum(s)!: Abc{
    x: 60
}
```

esquerbatua:
> `mut head := 0` will work only for integers, but not for other types like Structs etc, unlike `mut head := array[0]` .

Right, sorry, I'll think another solution to optimize it. Ty!

spytheman:
if your goal is to just avoid the check with the `continue`, a C style for loop: `for i:=1; i<a.len;i++ { e := a[i] ...` can work, skipping the 0th element.

PR #21713, opened 2024-06-23
checker: fix if cond using const with $d() def

### LINKED ISSUE #21709: cannot use constants with values ​​obtained using `$d` as a condition in comptime `if`
### Describe the bug

Cannot use constants with values ​​obtained using `$d` as a condition in comptime `if`.

### Reproduction Steps

```v
const x = $d('x', true)

$if x {
	println('x == true')
}

```

### Expected Behavior

```sh
$ v run main.v
x == true
```

### Current Behavior

```sh
main.v:3:5: error: definition of `x` is unknown at compile time
    1 | const x = $d('x', true)
    2 | 
    3 | $if x {
      |     ^
    4 |     print('x == true')
    5 | }
```

### Possible Solution

_No response_

If a literal is used, it works:

```v
const x = true

$if x {
	println('x == true')
}
```

### V version

V 0.4.6 be9d2b6

V Playground
> [!NOTE]
>

felipensp:
Fix #21709

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -4215,6 +4215,13 @@ fn (mut c Checker) find_obj_definition(obj ast.ScopeObject) !ast.Expr {
 	if mut expr is ast.Ident {
 		return c.find_definition(expr)
 	}
+	if mut expr is ast.ComptimeCall && expr.is_compile_value {
+		if expr.result_type == ast.bool_type {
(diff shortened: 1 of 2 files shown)


Delta456 (approved):

spytheman (approved):

spytheman:
Excellent work.

PR #21717, opened 2024-06-23
checker: fix result call or_block with multi-statements (fix #21504)

### LINKED ISSUE #21504: Storing a function result value in a variable does not work as expected in an `or` block
### Describe the bug

Storing the result value of a function call in a variable behaves differently than directly passing the function multiple times.

### Reproduction Steps

No variable usage, expected result:
```v
fn str_ret_fn(str string) string {
	return str
}

fn foo(str string) !string {
	if str.contains('foo') {
		return str
	}
	return error('error')
}

var := foo('bar') or {
	// foo_var := str_ret_fn('foo') // uncomment for unexpected behaviour
	if str_ret_fn('foo') == 'foo' {
		foo(str_ret_fn('foo')) or {
			eprintln(err)
			exit(1)
		}
	} else {
		eprintln(err)
		exit(1)
	}
}

println(var) // `foo`
```
Variable usage, unexpected result:
```v
fn str_ret_fn(str string) string {
	return str
}

fn foo(str string) !string {
	if str.contains('foo') {
		return str
	}
	return error('error')
}

var := foo('bar') or {
	foo_var := str_ret_fn('foo')
	if foo_var == 'foo' {
		foo(foo_var) or {
			eprintln(err)
			exit(1)
		}
	} else {
		eprintln(err)
		exit(1)
	}
}

println(var) // `println(NIL)`
```

### Expected Behavior

No difference weather using a variable or not.

### Current Behavior

See repro above.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.5 58a8fc6

macos, linux arm64
> [!NOTE]
>

yuyi98:
This PR fix result call or_block with multi-statements (fix #21504).

- Fix result call or_block with multi-statements.
- Add test.

```v
fn str_ret_fn(str string) string {
	return str
}

fn foo(str string) !string {
	if str.contains('foo') {
		return str
	}
	return error('error')
}

fn main() {
	var := foo('bar') or {
		foo_var := str_ret_fn('foo')
		if foo_var == 'foo' {
			foo(foo_var) or {
				eprintln(err)
				exit(1)
			}
		} else {
			eprintln(err)
			exit(1)
		}
	}

	println(var)
	assert var == 'foo'
}

PS D:\Test\v\tt1> v run .
foo
```

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -1647,7 +1647,7 @@ fn (mut c Checker) selector_expr(mut node ast.SelectorExpr) ast.Type {
 		node.typ = field.typ
 		if node.or_block.kind == .block {
 			c.expected_or_type = node.typ.clear_option_and_result()
-			c.stmts_ending_with_expression(mut node.or_block.stmts)
+			c.stmts_ending_with_expression(mut node.or_block.stmts, c.expected_or_type)
 			c.check_or_expr(node.or_block, node.typ, c.expected_or_type, node)
 			c.expected_or_type = ast.void_type
 		}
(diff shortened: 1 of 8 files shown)


spytheman (approved):
Excellent work.

PR #21722, opened 2024-06-24
builtin: fix `snake_to_camel` when passing a `SCREAMING_SNAKE_CASE` string

ttytm:
Ref.: https://github.com/vlang/v/pull/21691#issuecomment-2184905360

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -2647,26 +2647,38 @@ pub fn (s string) camel_to_snake() string {
 	if s.len == 0 {
 		return ''
 	}
-	lower_first_char := if s[0] >= `A` && s[0] <= `Z` { s[0] + 32 } else { s[0] }
 	if s.len == 1 {
-		return lower_first_char.ascii_str()
+		return s.to_lower()
 	}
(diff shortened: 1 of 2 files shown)


ttytm:
Would a `.is_upper` / `.is_lower` method for chars make it as a builtin?

spytheman:
We have u8.is_capital/0 already, which does c >= `A` && c <= `Z`, but I could not find a separate implementation for u8.is_lower/0 .

spytheman (approved):

ttytm:
Good catch. Thanks a lot for the test cases @kbkpbot

spytheman:
Why `assert '_ISspace'.camel_to_snake() == '_i_sspace'` ?

I would have expected `assert '_ISspace'.camel_to_snake() == '_is_space' 🤔 .

ttytm:
Good question. From the short research I did before, the initial implementation (without extracting handling of the first characters to reduce the load in the loop) was very similar to the implementation at https://github.com/iancoleman/strcase. So the initial impl and Go's strcase gives `_i_sspace` :thinking: .

PR #21725, opened 2024-06-24
vlib: simplify byte character conditions by utilizing methods

ttytm:


spytheman:
What is the performance impact?

yuyi98:
Because `is_alnum()/is_capital()/is_letter()` are inline functions, there is no performance impact and the overall clarity is much clearer.

yuyi98 (approved):

ttytm:
Yep, should be only only a cosmetic / read- / maintainability  change.

- ```v
  (c >= `a` && c <= `z`) || (c >= `A` && c <= `Z`)
  ```
  occurrences of those are replaced with
  https://github.com/vlang/v/blob/c5c49d36852ff2346d5bbd900cb64d1493635df9/vlib/builtin/string.v#L2065-L2068
  
- ```v
  c >= `0` && c <= `9`
  ```
  replaced with
  https://github.com/vlang/v/blob/c5c49d36852ff2346d5bbd900cb64d1493635df9/vlib/builtin/string.v#L2037-L2040
  
- ```v
  (c >= `a` && c <= `z`) || (c >= `A` && c <= `Z`) || (c >= `0` && c <= `9`)
  ```
  replaced with
  https://github.com/vlang/v/blob/c5c49d36852ff2346d5bbd900cb64d1493635df9/vlib/builtin/string.v#L2072-L2075

- To not change anything else, e.g. in case of below
  ```v
  (c >= `a` && c <= `z`) || (c >= `A` && c <= `Z`) || c == `_` || (c >= `0` && c <= `9`)
  ```

  I'm replacing it with

  ```v
  c.is_letter() || c == `_` || c.is_digit()
  // instead of 
  c.is_alnum() || c == `_`
  ```

spytheman:
> Because `is_alnum()/is_capital()/is_letter()` are inline functions, there is no performance impact and the overall clarity is much clearer.

In theory I agree, but have you measured, for example, recompiling V itself before/after the change?

spytheman:
V executables compiled with `-prod -cc gcc-11`, are consistently a bit faster with the changes here:
image

spytheman:
V executables compiled with just `v self` (i.e. using tcc), are in contrast a bit slower, but the relative difference is much smaller, and sometimes the new version is faster too:
image

spytheman (approved):
Excellent work.

ttytm:
Thanks for verifying @spytheman ! Didn't found the a free moment to do and share those tests on time.

JalonSolov:
I'm glad the extra performance testing was done.  Too many times in the past I've seen someone say "It gets inlined, so it'll be fast."

2 things to always keep in mind:

1. Inline is a **_suggestion_**, not a guarantee.  The C compiler is free to ignore it.

2. Inline only works (sometimes, currently, at least) with the C backend.  It has no affect on the JavaScript, WASM, native, etc. backends.

However, with the tcc tests, it appears the changes don't have any significant impact, so should be fine all around.

PR #21736, opened 2024-06-26
builtin: fix `assert '_ISspace'.camel_to_snake() == '_i_sspace'`

ttytm:
Ref.: https://github.com/vlang/v/pull/21722#issuecomment-2190443321

### DIFF
--- a/vlib/builtin/string_test.v
+++ b/vlib/builtin/string_test.v
@@ -1529,6 +1529,7 @@ fn test_camel_to_snake() {
 	assert 'BBaa'.camel_to_snake() == 'b_baa'
 	assert 'aa_BB'.camel_to_snake() == 'aa_bb'
 	assert 'JVM_PUBLIC_ACC'.camel_to_snake() == 'jvm_public_acc'
+	assert '_ISspace'.camel_to_snake() == '_i_sspace'
 }
 
 fn test_snake_to_camel() {
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -2662,9 +2662,13 @@ pub fn (s string) camel_to_snake() string {
 		}
 		lower_first_c, lower_second_c
 	} else {
-		lower_first_c := s[0]
-		second_c := if s[1].is_capital() { u8(`_`) } else { s[1] }
-		lower_first_c, second_c
+		first_c := s[0]
+		second_c := if s[1].is_capital() {
+			if first_c == `_` { s[1] + 32 } else { u8(`_`) }
+		} else {
+			s[1]
+		}
+		first_c, second_c
 	}
 	unsafe {
 		b[0] = first_char


vlib/builtin/string_test.v
```diff
@@ -1529,6 +1529,7 @@ fn test_camel_to_snake() {
 	assert 'BBaa'.camel_to_snake() == 'b_baa'
 	assert 'aa_BB'.camel_to_snake() == 'aa_bb'
 	assert 'JVM_PUBLIC_ACC'.camel_to_snake() == 'jvm_public_acc'
+	assert '_ISspace'.camel_to_snake() == '_i_sspace'
```
spytheman, line 1532: Is not `IS` a separate word from `space`?
But then `BBaa` became `b_baa` 🤔, so it is consistent.

spytheman: I have these in my `/usr/include/ctype.h`:
```c
# include <bits/endian.h>
# if __BYTE_ORDER == __BIG_ENDIAN
#  define _ISbit(bit)   (1 << (bit))
# else /* __BYTE_ORDER == __LITTLE_ENDIAN */
#  define _ISbit(bit)   ((bit) < 8 ? ((1 << (bit)) << 8) : ((1 << (bit)) >> 8))
# endif

enum
{
  _ISupper = _ISbit (0),    /* UPPERCASE.  */
  _ISlower = _ISbit (1),    /* lowercase.  */
  _ISalpha = _ISbit (2),    /* Alphabetic.  */
  _ISdigit = _ISbit (3),    /* Numeric.  */
  _ISxdigit = _ISbit (4),   /* Hexadecimal numeric.  */
  _ISspace = _ISbit (5),    /* Whitespace.  */
  _ISprint = _ISbit (6),    /* Printing.  */
  _ISgraph = _ISbit (7),    /* Graphical.  */
  _ISblank = _ISbit (8),    /* Blank (usually SPC and TAB).  */
  _IScntrl = _ISbit (9),    /* Control character.  */
  _ISpunct = _ISbit (10),   /* Punctuation.  */
  _ISalnum = _ISbit (11)    /* Alphanumeric.  */
};
#endif /* ! _ISbit  */
```

And later:
```c
/usr/include/ctype.h:197:# define isspace(c)    __isctype((c), _ISspace)
```

I think the intended usage of the macro is as a predicate (`is space`), for checking whether a letter is a space character, an ASCII code of 32 (2^5).

ttytm: Right, a conversion to `_is_space` would be more intuitive is the impression I'm getting too. 

As you mention the implementation to handle a `BBaa as b_baa` should be updated to be consistent then. Since we handle the first separately: the general implementation for capitals followed by lowercase characters should be updated then: `aaBBc` currently becomes `aa_b_bc`, while `aa_bb_c` would be consistent.

spytheman: I think the rule may be that if there is `camelCase`, i.e. a single capital letter `C`, followed by lower case letters, then the capital is part of the following word.

However, if there are several capitals one after the other, then the whole span of capitals, is its own independent word (perhaps from an acronym, or an abbreviation, or the product of a deranged mind), and the following lower case letters form their own independent word.

ttytm: Doing a short search for commonalities and conventions regarding camel to snake conversions, it would `_i_sspace` for for a most widely used rust crate to:

https://crates.io/crates/convert_case

```rs
use convert_case::{Case, Casing};

fn main() {
	let camel_str = "_ISspace".from_case(Case::Camel).to_case(Case::Snake);
	dbg!(camel_str); // `[main.rs:5:5] camel_str = "_i_sspace"`
	let camel_str = "AAbb".from_case(Case::Camel).to_case(Case::Snake);
	dbg!(camel_str); // `[main.rs:7:5] camel_str = "a_abb"`
}
```

ttytm: Time to break a convention?

JalonSolov: Looks like Rust took the "easy" way out... counting the capitals is a bit more work.

I'd prefer the method that looks more correct... it can always be changed if it causes problems.

spytheman: > If it's time to break a convention, maybe first merge the fix to have it in the history for a potential rollback, then make the change?

Yes, that is good idea. The version on master is definitely broken, since it loses a letter, while the version here does not.

PR #21740, opened 2024-06-26
cgen: fix infix expr and/or operate unnecessary evaluate (fix #21590)

### LINKED ISSUE #21590: performance, robustness: don't evaluate expressions in conditions unnecessarily
### Describe the bug

A logical AND (&&) with an intentional order of conditional statements could behave the same creating deeper nesting levels.

### Reproduction Steps

```v
import semver

version := 'nightly'
if version != 'nightly' && semver.from(version)! < semver.from('2.4.1')! {
} else {
	// condition
}
```

Result:
`version != 'nightly` is checked, still getting the error:
```
V panic: result not set (Invalid version format for input "nightly")
```

### Expected Behavior

Same as using a nested conditon:

```v
import semver

version := 'nightly'
if version != 'nightly' {
	if semver.from(version)! < semver.from('2.4.1')! {
	} else {
		// condition
	}
} else {
	// same else condition
}
```
Result:
Won't error

### Current Behavior

```
V panic: result not set (Invalid version format for input "nightly")
```

### Possible Solution

_No response_

_No response_

### V version

👇

```
V full version: V 0.4.6 09eaae5
OS: linux, "EndeavourOS Linux"
Processor: 16 cpus, 64bit, little endian, 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz

getwd: /home/t/Nextcloud/Dev/vlang/playground
vexe: /home/t/Nextcloud/Dev/vlang/v/v
vexe mtime: 2024-05-28 18:24:11

vroot: OK, value: /home/t/Nextcloud/Dev/vlang/v
VMODULES: OK, value: /home/t/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.45.1
Git vroot status: weekly.2024.22-4-g09eaae5f
.git/config present: true

CC version: cc (GCC) 14.1.1 20240522
```
> [!NOTE]
>

yuyi98:
This PR fix infix expr and/or operate unnecessary evaluate (fix #21590).

- Fix infix expr and/or operate unnecessary evaluate.
- Add test.

```v
fn f(s string) !int {
	if s == '' {
		return error('invalid s')
	}
	return s.len
}

fn main() {
	v := ''
	x := v != 'xyz' || f(v)! < f('abc')!
	dump(x)
	assert x

	y := v == 'xyz' && f(v)! < f('abc')!
	dump(y)
	assert !y
}

PS D:\Test\v\tt1> v run .
[.\\tt1.v:11] x: true
[.\\tt1.v:15] y: false
```

### DIFF
--- a/vlib/v/gen/c/infix.v
+++ b/vlib/v/gen/c/infix.v
@@ -1057,6 +1057,22 @@ fn (mut g Gen) infix_expr_and_or_op(node ast.InfixExpr) {
 				return
 			}
 		}
+	} else if node.right is ast.InfixExpr && g.need_tmp_var_in_expr(node.right) {
+		prev_inside_ternary := g.inside_ternary
+		g.inside_ternary = 0
+		tmp := g.new_tmp_var()
+		cur_line := g.go_before_last_stmt().trim_space()
+		g.empty_line = true
+		g.write('bool ${tmp} = (')
+		g.expr(node.left)
+		g.writeln(');')
+		g.set_current_pos_as_last_stmt_pos()
+		g.write('${cur_line} ${tmp} ${node.op.str()} ')
+		g.infix_left_var_name = if node.op == .and { tmp } else { '!${tmp}' }
+		g.expr(node.right)
+		g.infix_left_var_name = ''
+		g.inside_ternary = prev_inside_ternary
+		return
 	}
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

Issue #21751, opened 2024-06-27
os.join_path() returns a non-rooted path when second argument is rooted path


mike-ward:
### Describe the bug

`os.join_path` returns a non rooted path when the first argument is an empty string and the second argument is a rooted path.

### Reproduction Steps

```
path := os.join_path('', '/usr/bin/x')
println(path) // => usr/bin/x
```

### Expected Behavior

Expected path to equal `/usr/bin/x`

### Current Behavior

path equals `usr/bin/x` // missing root path separator

### Possible Solution

Briefly looked at the code for os.join_path. It appears it needs an additional check to see if the path is rooted.

### V version

V 0.4.6 fccd7cd

### Environment details (OS name and version, etc.)

V full version: V 0.4.6 8215f21.fccd7cd
OS: macos, macOS, 14.5, 23F79
Processor: 8 cpus, 64bit, little endian, Apple M2

getwd: /Users/mike/Documents/github/coreutils/src/sort
vexe: /Users/mike/Documents/github/v/v
vexe mtime: 2024-06-27 19:39:41

vroot: OK, value: /Users/mike/Documents/github/v
VMODULES: OK, value: /Users/mike/.vmodules
VTMP: OK, value: /tmp/v_501

Git version: git version 2.45.2
Git vroot status: weekly.2024.26-11-gfccd7cd0
.git/config present: true

CC version: Apple clang version 15.0.0 (clang-1500.3.9.4)
thirdparty/tcc status: thirdparty-macos-arm64 5c1d002f
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

spytheman:
That is deliberate. The later arguments *are relative* to the first, and that is encoded in the existing tests of os.join_path .

PR #21752, opened 2024-06-28
os: fix join_path and join_path_singe to respect absolute paths

mike-ward:
Fixes issue #21751 where by absolute paths in second and subsequent arguments in `join_path` and `join_path_single` were treated as relative paths instead of absolute paths.

Required changing several tests in join_path_text.v, which made incorrect assertions about join_path responses.


vlib/os/join_path_test.v
```diff
@@ -26,7 +26,7 @@ fn test_join_path() {
 		assert os.join_path('foo/bar', './file.txt') == r'foo\bar\file.txt'
 		assert os.join_path('/opt/v', './x') == r'\opt\v\x'
 		assert os.join_path('v', 'foo/bar', 'dir') == r'v\foo\bar\dir'
-		assert os.join_path('v', 'foo/bar\\baz', '/dir') == r'v\foo\bar\baz\dir'
+		assert os.join_path('v', 'foo/bar\\baz', '/dir') == r'/dir'
```
JalonSolov, line 29: Shouldn't it be `foo/bar/baz/dir`?

vlib/os/join_path_test.v
```diff
@@ -40,14 +40,14 @@ fn test_join_path() {
 		assert os.join_path('foo/bar', './file.txt') == 'foo/bar/file.txt'
 		assert os.join_path('/opt/v', './x') == '/opt/v/x'
 		assert os.join_path('/foo/bar', './.././file.txt') == '/foo/bar/../file.txt'
-		assert os.join_path('v', 'foo/bar\\baz', '/dir') == r'v/foo/bar/baz/dir'
+		assert os.join_path('v', 'foo/bar\\baz', '/dir') == '/dir'
```
JalonSolov, line 43: And here `v/foo/bar/baz/dir`.

vlib/os/join_path_test.v
```diff
@@ -40,14 +40,14 @@ fn test_join_path() {
...
 	}
 }
 
 fn test_join_path_single() {
 	assert os.join_path_single('', '') == ''
 	assert os.join_path_single('', './b') == 'b'
-	assert os.join_path_single('', '/b') == 'b'
+	assert os.join_path_single('', '/b') == '/b'
```
Delta456, line 50: I believe it should be `r'/b'` instead.

spytheman:
The later arguments to join_path *are* relative and not absolute.

PR #21757, opened 2024-06-28
checker: disallow incomplete thread type struct fields 

Delta456:
Fixes https://github.com/vlang/v/issues/14043


spytheman:
hm, why disallow them?

spytheman:
The original code in https://github.com/vlang/v/issues/14043 which is:
```v
struct Str {
	t thread
	a [0xF]bool
}
```
currently compiles cleanly, while before the parser could not handle it.

Imho the issue can be just closed as completed, without introducing a special checker error.

spytheman:
Keep in mind, that one may want to represent threads, in a generic way, without caring about the type of the work they do, and just call `.wait()` on them, to complete.

Delta456:
> Keep in mind, that one may want to represent threads, in a generic way, without caring about the type of the work they do, and just call `.wait()` on them, to complete.

I noticed that a while ago, so I forgot to close the PR.

PR #21764, opened 2024-06-30
cgen: fix array fixed empty struct code generated on clang

### LINKED ISSUE #21761: Compiler error when building global arrays of empty Structs
### Describe the bug

Consider the following code:

```
❯ cat 1.v
module main

const num_elements = 10

struct DummyStruct {}

__global (
        d [num_elements]DummyStruct
)

fn main() {}
```

When built with the following invocation, the compilation fails:

```
❯ v -enable-globals 1.v
==================
/tmp/v_501/1.01J1JR6W82J8GJBYBQHK0EWBH8.tmp.c:2068:41: error: initializer for aggregate with no elements requires explicit braces
 Array_fixed_main__DummyStruct_10  d = {0}; // global4
                                        ^
1 error generated.
...
==================

builder error:
==================

```

On the other hand, if DummyStruct is non-empty then the compilation succeeds. 

Consider the following:

```
❯ cat 2.v
module main

const num_elements = 10

struct DummyStruct {
        dummy_item i32
}

__global (
        d [num_elements]DummyStruct
)

fn main() {}
```

The only change is that DummyStruct now has a dummy_item.

The compilation succeeds as follows:

```
❯ ~/Work/repos/vlang/v/v -enable-globals 2.v
```

**NOTE**:

```
❯ v --version
V 0.4.6 23e3894

❯ neofetch --off
spinlock@deadlock.local
-------------------
OS: macOS 14.5 23F79 arm64
Host: Mac14,9
Kernel: 23.5.0
Uptime: 11 days, 11 hours, 48 mins
Packages: 250 (port)
Shell: fish 3.7.1
Resolution: 5120x1440 , 3024x1964
DE: Aqua
WM: Quartz Compositor
WM Theme: Blue (Dark)
Terminal: tmux
CPU: Apple M2 Pro
GPU: Apple M2 Pro
Memory: 4652MiB / 32768MiB

```

### Reproduction Steps

See above

### Expected Behavior

Structs containing no members is a perfectly valid use case and is often referenced in the official documentation. There is nothing special about them that warrants a compiler build failure in the case of global arrays.

### Current Behavior

See bug description above.

### Possible Solution

_No response_

_No response_

### V version

0.4.6 23e3894

macOS 14.5 arm64
> [!NOTE]
>

felipensp:
Fix #21761

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -4291,7 +4291,7 @@ fn (mut g Gen) debugger_stmt(node ast.DebuggerStmt) {
 fn (mut g Gen) enum_decl(node ast.EnumDecl) {
 	enum_name := util.no_dots(node.name)
 	is_flag := node.is_flag
-	if g.pref.ccompiler == 'msvc' {
+	if g.is_cc_msvc {
 		mut last_value := '0'
 		enum_typ_name := g.table.get_type_name(node.typ)
 		g.enum_typedefs.writeln('')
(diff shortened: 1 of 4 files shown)


vlib/v/gen/c/cgen.v
```diff
@@ -7157,7 +7159,13 @@ fn (mut g Gen) type_default(typ_ ast.Type) string {
 		.string {
 			return '(string){.str=(byteptr)"", .is_lit=1}'
 		}
-		.interface_, .sum_type, .array_fixed, .multi_return, .thread {
+		.array_fixed {
+			if g.pref.ccompiler_type == .clang && sym.is_empty_struct_array() {
```
spytheman, line 7163: The generated code will be more portable (which is important for vc/v.c bootstrapping) if there is not a special check for the compiler type here (the macro `EMPTY_STRUCT_INITIALIZATION` should handle it):
```suggestion
			if sym.is_empty_struct_array() {
```

spytheman:
(rebased over master, to fix an unrelated CI error)

PR #21774, opened 2024-07-01
checker: change comptime_if_branch() to comptime_if_cond()

yuyi98:
This PR change `comptime_if_branch()` to `comptime_if_cond()`.


spytheman (approved):
The new name is indeed better.

PR #21781, opened 2024-07-01
checker: fix assign generic if expr

### LINKED ISSUE #21776: "assignment mismatch: 1 variable(s) 0 value(s)" error for a simple generic if expression
### Describe the bug

```v
fn clamp[T](a T, x T, b T) T {
    min := if x < b { x } else { b }
    return if min < a { a } else { min }
}
```

```bash
error: assignment mismatch: 1 variable(s) 0 value(s)
  130 | fn clamp[T](a T, x T, b T) T {
  131 |     min := if x < b { x } else { b }
      |         ~~
  132 |     return if min < a { a } else { min }
  133 | }
```

this works:
```v
        mut min := T(0)
        min = if x < b { x } else { b }
```
so the problem is that the type does not get inferred, for generics
but if it is known, conditional expressions work fine

perhaps it is just a matter of placing `c.unwrap_generic()` somewhere in the checker

> [!NOTE]
>

felipensp:
Fix #21776

### DIFF
--- a/vlib/v/checker/assign.v
+++ b/vlib/v/checker/assign.v
@@ -52,7 +52,16 @@ fn (mut c Checker) assign_stmt(mut node ast.AssignStmt) {
 				node.right_types = right_type_sym.mr_info().types
 				right_len = node.right_types.len
 			} else if right_type == ast.void_type {
-				right_len = 0
+				if mut right is ast.IfExpr {
+					last_branch := right.branches.last()
+					last_stmts := last_branch.stmts.filter(it is ast.ExprStmt)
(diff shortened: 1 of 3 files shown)


vlib/v/checker/assign.v
```diff
@@ -53,6 +53,14 @@ fn (mut c Checker) assign_stmt(mut node ast.AssignStmt) {
 				right_len = node.right_types.len
 			} else if right_type == ast.void_type {
 				right_len = 0
+				if mut right is ast.IfExpr {
+					last_branch := right.branches.last()
```
spytheman, line 57: Why does it use the last branch, instead of the first here?

spytheman (approved):
Excellent work.

PR #21784, opened 2024-07-01
examples: add `gg/fire.v`

shadowninja55:
this PR adds a `gg` example showcasing a pixelated animation of fire

### DIFF
--- a/examples/gg/fire.v
+++ b/examples/gg/fire.v
@@ -1,0 +1,145 @@
+import gg
+import gx
+import rand
+
+const width = 100
+const height = 140
+const scale = 4
+const palette = [
+	gx.rgb(0x07, 0x07, 0x07),
+	gx.rgb(0x1f, 0x07, 0x07),
+	gx.rgb(0x2f, 0x0f, 0x07),
+	gx.rgb(0x47, 0x0f, 0x07),
+	gx.rgb(0x57, 0x17, 0x07),
+	gx.rgb(0x67, 0x1f, 0x07),
+	gx.rgb(0x77, 0x1f, 0x07),
+	gx.rgb(0x8f, 0x27, 0x07),
+	gx.rgb(0x9f, 0x2f, 0x07),
+	gx.rgb(0xaf, 0x3f, 0x07),
+	gx.rgb(0xbf, 0x47, 0x07),
+	gx.rgb(0xc7, 0x47, 0x07),
+	gx.rgb(0xdf, 0x4f, 0x07),
+	gx.rgb(0xdf, 0x57, 0x07),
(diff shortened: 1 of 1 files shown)


larpon:
It is running as expected on my machine:
image

(I ran the example `v -d show_fps run examples/gg/fire.v` to see framerate in top-left corner)

examples/gg/fire.v
```diff
@@ -0,0 +1,145 @@
...
+
+	app.gg = gg.new_context(
+		event_fn: event
+		frame_fn: frame
+		init_fn: retile
+		window_title: 'Fire Animation'
+		user_data: app
+		bg_color: palette[0]
+		width: 1200
+		height: 600
```
larpon, line 70: 1200 width may be a bit much on older setups. Examples like 2048, flappylearning and many others are around the 500 - 600 pixels in width.

Also width and height should be constants :slightly_smiling_face:

JalonSolov:
Just to prove optimization isn't _always needed... first, I ran it with `v -d show_fps run fire.v` and it showed 60 fps.

Then I compiled it with `-prod -d show_fps` and ran the compiled version... and it was still 60 fps.

I like how short the V source is - shows that there's a lot of power in `gg`.  :-)

medvednikov:
Indeed :)

Sokol limits the fps to 60, that's why -prod had no difference.

spytheman:
You can try `LIBGL_ALWAYS_SOFTWARE=1 v -d show_fps run fire.v` too, if you have a fast CPU. In this case, there will be no fps limit imposed by the hardware.

spytheman:
A small variation is https://play.vlang.io/p/18c27fca56 (it is drawing in streaming image buffer, which is then drawn once per tile, instead of a lot of individual app.gg.draw_square_filled calls).

shadowninja55:
> A small variation is https://play.vlang.io/p/18c27fca56 (it is drawing in streaming image buffer, which is then drawn once per tile, instead of a lot of individual app.gg.draw_square_filled calls).

should i implement this into my code, or were you simply providing it as an example ?

spytheman:
I am providing it as an example, since on my machine (with an i3-3225 processor with an integrated Intel HD4000 GPU), it is more efficient - 45% CPU usage vs 10% CPU usage, when the app is maximized.

It is also more complex, so it is not clear to me, whether it should be implemented in the example, which is currently shorter/clearer, and demonstrates another technique (splitting the rendering into several passes, to overcome the Sokol limit).

examples/gg/fire.v
spytheman: Fixed in 8ee3e89 .

Issue #21767, opened 2024-06-30
after v fmt cause compile error


kbkpbot:
### Describe the bug

`v fmt` create a new v file, will cause compile error

### Reproduction Steps

test.v
```v
module main

const spnames = ['abc', (voidptr(0))]!

fn main() {
        println(spnames)
}
```

v test.v

v fmt -w test.v

v test.v

### Expected Behavior

compile ok

### Current Behavior

```sh
==================
/tmp/v_1000/test_fixed_array.01J1MJ2S78KQ8AEFD7J4VNYMKE.tmp.c:13226: error: cannot convert 'void *' to 'struct string'
...
==================
(Use `v -cg` to print the entire error message)

builder error:
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .

```

### Possible Solution

before `v fmt`, the generated c code :
```c
Array_fixed_string_2 _const_main__spnames = {_SLIT("abc"), (((voidptr)(0)))}; // fixed array const
```

after `v fmt`, the generated c code:
```c
Array_fixed_string_2 _const_main__spnames; // inited later
```

### Additional Information/Context
```v
const spnames = ['abc', (voidptr(0))]!
```
will create a fixed array const

but , after `v fmt`, 
```v
const spnames = ['abc', (unsafe { nil })]!
```
will  create a variable length array ?

_No response_

### V version

V full version: V 0.4.6 96751ed.4a7c70c

### Environment details (OS name and version, etc.)

V full version: V 0.4.6 96751ed.4a7c70c
OS: linux, Ubuntu 22.04.4 LTS
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz

getwd: /home/mars/v/9
vexe: /media/HD/github/lang/v/v
vexe mtime: 2024-06-29 03:29:17

vroot: OK, value: /media/HD/github/lang/v
VMODULES: OK, value: /home/mars/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.34.1
Git vroot status: weekly.2024.26-14-g4a7c70c9 (4 commit(s) behind V master)
.git/config present: true

CC version: cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
[...]

PR #21786, opened 2024-07-01
checker: disallow nil in non-nil arrays and vice versa 

Delta456:
Fixes https://github.com/vlang/v/issues/21767


vlib/v/checker/containers.v
```diff
@@ -205,6 +205,15 @@ fn (mut c Checker) array_init(mut node ast.ArrayInit) ast.Type {
...
 				}
+				if elem_type != ast.nil_type && typ.idx() == ast.nil_type_idx {
+					c.error('cannot have `unsafe { nil }` inside a `${c.table.type_to_str(elem_type)}` array',
+						expr.pos())
+				}
+
+				if elem_type == ast.nil_type && typ.idx() != ast.nil_type_idx {
+					c.error('cannot have `${c.table.type_to_str(typ)}` inside a `${c.table.type_to_str(elem_type)}` array',
+						expr.pos())
+				}
```
Delta456, line 216: I believe this isn't the right way to do this, but I can't think of any alternative. `check_expected` has special usage, so I can't add the check directly.

I would love to know alternatives.

medvednikov:
It should be in the checker.

This code is fine since it's an array of &char. c2v doesn't translate those to strings.

medvednikov:
@Delta456 I think this check should be in check_types(), so that it works everywhere. Strings shouldn't be compatible with nil.

Delta456:
> @Delta456 I think this check should be in check_types(), so that it works everywhere. Strings shouldn't be compatible with nil.

I will try doing this again. Also should we allow nil be compatible with c string? @felipensp's vtcc breaks.

spytheman:
C strings are represented as *pointers* to their start, so yes, `unsafe { nil }` is valid everywhere where a C string is expected.

spytheman:
Unlike C strings, V strings are represented as *structures*, not pointers, so `unsafe { nil }` is not compatible with them.

vlib/v/checker/check_types.v
```diff
@@ -129,6 +129,19 @@ fn (mut c Checker) check_types(got ast.Type, expected ast.Type) bool {
 			return true
 		}
 	}
+
+	got_sym := c.table.sym(got)
```
medvednikov, line 133: Please add a block comment to explain what this block does

medvednikov: also since you only do this for nil checks, the costly lookups should only be done inside the `if`.

you can check whether it's a string without table.sym(), just use got like you use exp_idx

Delta456: True. I will move this.

medvednikov:
Good job.

Delta456:
> So, ```t:=[c'def', unsafe{nil}]!``` create a fixed length array?

Yes

PR #21792, opened 2024-07-03
vrepl: fix method call (fix #21788)

### LINKED ISSUE #21788: array.pop() ineffective in repl without attribution of the result
### Describe the bug

If I use `array.pop()` in the repl, it's not effective except if the result is attributed to a variable. 

### Reproduction Steps

In the V repl, I get this: 

```
>>> mut values := [1, 2, 3]
>>> values.pop()
3
>>> values
[1, 2, 3]
>>> removed := values.pop()
>>> values
[1, 2]
```

In a V application, it's ok:

```
fn main() {
	mut values := [1, 2, 3]
	println('before: ${values}')
	values.pop()
	println('after: ${values}')
}
```

```
v run main.v
before: [1, 2, 3]
after: [1, 2]
```

### Expected Behavior

Get the same behavior from code and repl

### Current Behavior

In the repl, `values.pop()` is ineffective, `removed := values.pop()` is ok

### Possible Solution

_No response_

_No response_

### V version

0.4.6 c412b9f

V full version: V 0.4.6 c412b9f
OS: macos, macOS, 14.5, 23F79
Processor: 12 cpus, 64bit, little endian, Apple M3 Pro

getwd: /Users/davlgd/Documents/code/clever/vTest
vexe: /opt/homebrew/Cellar/vlang/0.4.6_1/libexec/v
vexe mtime: 2024-05-20 18:06:54

vroot: OK, value: /opt/homebrew/Cellar/vlang/0.4.6_1/libexec
VMODULES: OK, value: /Users/davlgd/.vmodules
VTMP: OK, value: /tmp/v_501

Git vroot status: 4.3.7 (58345 commit(s) behind V master)
.git/config present: false

thirdparty/tcc: N/A
> [!NOTE]
>

yuyi98:
This PR fix method call (fix #21788).

- Fix method call.
- Add test.

```v
PS D:\Vlang\v> v
 ____    ____
 \   \  /   /  |  Welcome to the V REPL (for help with V itself, type  exit , then run  v help ).
  \   \/   /   |  Note: the REPL is highly experimental. For best V experience, use a text editor,
   \      /    |  save your code in a  main.v  file and execute:  v run main.v
    \    /     |  V 0.4.6 209063f . Use  list  to see the accumulated program so far.
     \__/      |  Use Ctrl-C or  exit  to exit, or  help  to see other available commands.

>>> mut values := [1, 2, 3]
>>> values.pop()
>>> values
[1, 2]
```

### DIFF
--- a/cmd/tools/vrepl.v
+++ b/cmd/tools/vrepl.v
@@ -442,7 +442,9 @@ fn run_repl(workdir string, vrepl_prefix string) int {
 			filter_line := r.line.replace(r.line.find_between("'", "'"), '').replace(r.line.find_between('"',
 				'"'), '')
 			mut is_statement := false
-			if filter_line.count('=') % 2 == 1
+			if func_call {
+				is_statement = true
+			} else if filter_line.count('=') % 2 == 1
 				&& (filter_line.count('!=') + filter_line.count('>=') + filter_line.count('<=')) == 0 {
 				is_statement = true
 			} else {
--- a/vlib/v/slow_tests/repl/comptime_tmpl.repl
+++ b/vlib/v/slow_tests/repl/comptime_tmpl.repl
@@ -1,3 +1,3 @@
-$tmpl('./tmpl/hello.txt')
+print($tmpl('./tmpl/hello.txt'))
 ===output===
 hello
--- a/vlib/v/slow_tests/repl/fn_calls.repl
+++ b/vlib/v/slow_tests/repl/fn_calls.repl
@@ -1,4 +1,4 @@
-math.sinf(50.0)
+println(math.sinf(50.0))
 println(1+math.sinf(50.0))
 fn test() { println('foo') } fn test2(a int) { println(a) }
 test()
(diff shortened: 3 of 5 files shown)


spytheman:
This changes the way that the REPL shows function calls, returning values.
Previously it tried to detect such cases, and wrap the corresponding calls with a println(), in effect showing the result:
```
>>> mut values := [1,2,3]
>>> values.map(it*10)
[10, 20, 30]
>>>
```

Now, the behavior is closer to how V behaves normally (ignoring the resulting value), which while more reliable and predictable, when used in combination with methods that mutate their receivers, is a bit less useful for interactive experimentation, since it requires an explicit `println()`:
```
>>> mut values := [1,2,3]
>>> values.map(it*10)
>>>
>>> println(values.map(it*10))
[10, 20, 30]
>>>
```

spytheman:
I do not have a solution though, given its nature (lack of actual interpreter, that can keep runtime state). Perhaps it is better, to err on the side of being closer to actual V, reducing the surprises, at the cost of a bit higher inconvenience.

spytheman (approved):
Good work.

PR #21798, opened 2024-07-03
examples: add interpolated movement and input queue to `snek/snek.v`

shadowninja55:
this PR updates the `snek` example to use interpolated movement (a smoothly animated snake) and an input queue (more responsive input, no dropped keypresses). it also simplifies some of the logic by changing the `Direction` enum to simply use the `Pos` struct (now named `Vec`).

### DIFF
--- a/examples/snek/snek.v
+++ b/examples/snek/snek.v
@@ -1,9 +1,9 @@
-import os
+import datatypes
 import gg
 import gx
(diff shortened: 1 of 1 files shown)


spytheman (approved):

spytheman:
The new animation does look a lot better indeed.

PR #21805, opened 2024-07-05
token: make is_assign() contain .decl_assign

yuyi98:
This PR make token.is_assign() contain .decl_assign.

- Make `token.is_assign()` contain `.decl_assign`.
- Modify related calls.


spytheman (approved):
Very nice.

PR #21808, opened 2024-07-05
doc: update full list of builtin compile options

kimshrier:
I noticed that the documentation didn't mention some of the legal values for builtin compile-time options.  I think these updates are accurate and complete.


spytheman (approved):
Good work.

PR #21809, opened 2024-07-05
v/tests: fix handling of string type closure test

kimshrier:
I noticed that on macos, the max_params value was limited to 9 instead of 16 with the comment that 10 or more was not supported on m1.

In my experimentation, only the closure test on strings was limited and that the other, integer types, worked when max_params was 16.

This compiler limitation also did not appear to be specific to macos.  I found that on FreeBSD, when using the clang compiler on an arm64 machine, I needed to limit the string tests to 9 but the other integer types would work all the way up to 16, and even a little beyond 16 but I left the upper bound at 16 for them.

When using gcc on FreeBSD, I could let max_params be 16 for the string tests.  Also testing with clang on FreeBSD on amd64, I could let max_params be 16 for the string tests.

Since there was some special handling of the string type in the existing closure test, and that only the string type needed to be limited to 9 in special cases, I separated out the string closure tests and defined its own max value, max_string_params, and made that conditional on running on macos or running with the clang compiler on arm64.  I don't have an arm32 machine to test on so I couldn't test that scenario out.

By separating the string tests from the other integer tests, the code became simpler and easier to understand.  Or, at least, so it seems to me.

### DIFF
--- a/vlib/v/tests/closure_generator_test.v
+++ b/vlib/v/tests/closure_generator_test.v
@@ -2,14 +2,19 @@ import strings
 import os
 
 const max_params = get_max_params()
+const max_string_params = get_max_string_params()
 const all_param_names = []string{len: max_params, init: '${`a` + index}'}
 const all_param_values = []string{len: max_params, init: '${index + 1}'}
 
 fn get_max_params() int {
-	$if macos {
-		// on m1, 10 or more parameters are not supported
+	return 16
+}
+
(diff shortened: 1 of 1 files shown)


spytheman:
> By separating the string tests from the other integer tests, the code became simpler and easier to understand. Or, at least, so it seems to me.

It does seem cleaner, yes.
Thank your for the detailed investigation.

Issue #20636, opened 2024-01-23
Insufficient support for va_list type


arkt8:
### Describe the bug

There is no documented `va_list` alternative for use when using C bindings, but looks like there is a partial support when using `v translate`.

### Reproduction Steps

File test.c:

```c
#include <stdarg.h>

int sum(int qtd, ...) {
  va_list va;
  int s = 0;
  va_start(va, qtd);
  for (int i=0; i < qtd; i++) {
    s += (va_arg(va, int));
  }
  va_end(va);
  return s;
}

int main() {
  printf("Result %d\n", sum(3,4,5,6));
  return 0;
}
```

Then run:
`v translate test.c`

### Expected Behavior

Work seamlessly converting to V and, after running the resulting binary print on screen:
`Result 15`

### Current Behavior

The command `v translate test.c` gives the following error:

```
C to V translator 0.4.0
  translating /tmp/test.c     ... test.v:10:9: error: invalid expression: unexpected token `)`
    8 |     __builtin_va_start(va, qtd)
    9 |     for i := 0 ; i < qtd ; i ++ {
   10 |         s += ()
      |               ^
   11 |     }
   12 |     __builtin_va_end(va)

Internal vfmt error while formatting file: /tmp/test.v.
Encountered a total of: 1 errors.
 took    35 ms ; output .v file: test.v
Translated   1 files in    35 ms.
```

### Possible Solution

Partial / insufficient implementation of `va_list` or `stdarg.h`.

### Additional Information/Context

Even with the error, V still generates a `test.v` file with the following content:

```v
@[translated]
module main

@[c2v_variadic]
fn sum(qtd int) int {
	va := Va_list{}
	mas := 0
	__builtin_va_start(va, qtd)
	for i := 0 ; i < qtd ; i ++ {
		s += ()
	}
	__builtin_va_end(va)
	return s
}

fn main()  {
	C.printf(c'Result %d\n', sum(3, 4, 5, 6))
	return 
}
```

### V version

0.4.4 0205154

### Environment details (OS name and version, etc.)

v doctor output:

```
V full version: V 0.4.4 65a1d2f.0205154
OS: linux, Debian GNU/Linux bookworm/sid
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz

getwd: /tmp
vexe: /opt/v/v
vexe mtime: 2024-01-23 01:53:07

vroot: OK, value: /opt/v
[...]

PR #21812, opened 2024-07-06
v: Improve C var args interop

felipensp:
This PR:
- Introduce simple `...` syntax for function declaration (to use C var args)
- Makes possible to declare `C.va_list{}` without initialization.

```v
@[typedef]
struct C.va_list {}
fn C.va_start(voidptr, voidptr)
fn C.va_end(voidptr)

fn C.vfprintf(&C.FILE, &char, C.va_list) int

fn t2(fmt voidptr, ...) {
	ap := C.va_list{}
	C.va_start(ap, fmt)
	C.vfprintf(C.stderr, fmt, ap)
	C.va_end(ap)
}

fn main() {
	a := 1
	b := 2.5
	t2(c'%s : %d : %.2f\n', c'foo', a, b)
}
```

Outputs `foo : 1 : 2.50`


vlib/v/gen/c/cheaders.v
```diff
@@ -342,7 +342,7 @@ const c_common_macros = '
 	#define TCCSKIP(x)
 	// #include <byteswap.h>
 	#ifndef _WIN32
-		int tcc_backtrace(const char *fmt, ...);
+		int tcc_backtrace(char *fmt, ...);
```
spytheman, line 345: Why is that needed?

spytheman: Did it change in latest tcc?

spytheman: Hm, it seems to be still declared with const there:
```c
int tcc_backtrace(const char *fmt, ...);
```

https://repo.or.cz/tinycc.git/blob/HEAD:/lib/bcheck.c#l448

felipensp: Yeah, this will not be needed to change. Reverting this part.

spytheman:
Excellent work, but it misses a test, and it does not fix #20636 yet (which uses also `va_arg`, and `v translate` does not handle well neither va_start, va_end nor va_arg ... va_arg also accepts a *type name* as its second argument, which can not be modeled currently in V, except perhaps as a generic).

I suggest adding a test .vv file, in `vlib/v/gen/c/testdata/`, that includes a .c file containing the definition of a variadic function printing its arguments, then declaring and calling it in the .vv file, and having an .out file with the printed output.

Then `./v vlib/v/gen/c/coutput_test.v` will pick up the file and check that it continues to work locally and on the CIs.

spytheman:
The current example can also be copy/pasted as a _keep.vv file in `vlib/v/fmt/tests/`, to ensure that `vfmt` will continue to handle it.

PR #21813, opened 2024-07-06
v/pref: handle using gcc on FreeBSD

kimshrier:
When testing with VFLAGS='-cc gcc' I ran in to a similar problem that happened when I was testing with VFLAGS='-cc clang'.

The particular problem I am seeing is:
```
$ echo $VFLAGS
-cc gcc

$ v test vlib/v/pref/pref_test.v
---- Testing... ---------------------------------------------------------------------------------------
 FAIL     0.000 ms vlib/v/pref/pref_test.v
>> compilation failed:
==================
tcc: error: Unknown relocation type for got: 541
```

I had previously put in a test for freebsd and clang.  Now, I am adding a test for freebsd and gcc.

Is it possible to instead test for !tinyc and then use the currently specified compiler from the -cc parameter?  Otherwise, I am reduced to testing for every possible c compiler by name.

### DIFF
--- a/vlib/v/pref/pref_test.v
+++ b/vlib/v/pref/pref_test.v
@@ -38,6 +38,9 @@ fn test_version_flag() {
 	$if freebsd && clang {
 		compiler = 'clang'
 	}
+	$if freebsd && gcc {
+		compiler = 'gcc'
+	}
 
 	v_verbose_cmd_with_additional_args_res := os.execute_opt('${vexe} -cc ${compiler} -v run ${example_path}')!.output
 	assert v_verbose_cmd_with_additional_args_res != v_ver_cmd_res
--- a/vlib/v/pref/pref_test.v
+++ b/vlib/v/pref/pref_test.v
@@ -32,17 +32,7 @@ fn test_version_flag() {
 	assert v_verbose_cmd_res != v_ver_cmd_res
 	assert v_verbose_cmd_res.contains('v.pref.lookup_path:')
 
-	// tcc does not handle the symver assembly directive which is
-	// a problem on FreeBSD 14
-	mut compiler := 'tcc'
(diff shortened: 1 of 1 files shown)


spytheman:
> Is it possible to instead test for !tinyc

Try for example:
```v
println('hi')
$if tinyc {
	println('tcc is used')
}
$if !tinyc {
	println('tcc is not used')
}
println('done')
```

spytheman:
You can also use `os.getenv('VFLAGS')`, if you need it.

spytheman:
> use the currently specified compiler from the -cc parameter

VFLAGS is passed implicitly to all child processes, but setting `-cc` on the command line takes priority, so you may have to change the test code to do `mut compiler := '-cc tcc'` instead, then do:
`$if freebsd && !tinyc { compiler = '' }` .

spytheman:
Given the nature of the test and the output `v_verbose_cmd_with_additional_args_res`, imho the `-cc ...` argument can be removed completely too, and just rely on what is passed through VFLAGS (the test only checks that `-v` was passed, and that the output is thus different than before).

spytheman:
Or perhaps just use something safe, like `-skip-unused` or `-g`, instead of `-cc ...` there.

kimshrier:
I like the idea of replacing `-cc` with `-g` since that removes potential compiler idiosyncrasies from causing the test to fail while still testing `-v` with an additional argument.

I'll push that update in a minute.

kimshrier:
It looks like the failed check is not related to the change done in this merge request.

spytheman:
> It looks like the failed check is not related to the change done in this merge request.

Yes. It is unrelated.

spytheman (approved):
Good work.

PR #21814, opened 2024-07-06
examples: add vmatrix.v showing a Matrix like terminal screen effect

mike-ward:
A fun example of the, "Matrix" screen effect in V.

### DIFF
--- a/examples/vmatrix.v
+++ b/examples/vmatrix.v
@@ -1,6 +1,7 @@
 import os
 import rand
 import term
+import term.termios
 import time
 
 const snooze = time.millisecond * 70
@@ -47,7 +48,6 @@ fn random_rain_column(max_col int, max_height int) RainColumn {
 	return RainColumn{
 		col: rand.int_in_range(1, max_col + 1) or { 1 }
 		len: rand.int_in_range(4, max_height / 4 * 3) or { 4 }
-		head: 0
 	}
 }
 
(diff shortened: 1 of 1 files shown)


examples/vmatrix.v
Delta456, line 1: You forgot to `import rand`.

Delta456:
The program tears apart when scrolling.

image

examples/vmatrix.v
spytheman: What do you mean?
It is line 2?

mike-ward: Doesn't tear on Mac when I scroll, but then again I didn't make any claim that it would not tear.

mike-ward: image

Delta456: > What do you mean? It is line 2?

I copied the code which didn't have it :facepalm:

spytheman:
> The program tears apart when scrolling.

A path like this seems to work:

```v
import term.termios
...
fn init_terminal() ! {
    mut old_state := termios.Termios{}
    termios.tcgetattr(0, mut old_state)
    // restore cursor, exit alternate buffer mode on Ctrl+C
    os.signal_opt(os.Signal.int, fn [mut old_state] (sig os.Signal) {
        println('\e[?1049l\e[?25h')
        termios.set_state(0, old_state)
        exit(0)
    })!
    // turn off cursor, enter alternate buffer mode
    print('\e[?1049h\e[?25l')
    mut new_state := old_state
    new_state.disable_echo()
    termios.set_state(0, new_state)
}
```

spytheman:
https://github.com/abishekvashok/cmatrix/blob/master/cmatrix.c#L468C5-L468C11
also calls `noecho()`

mike-ward:
@spytheman Thanks for the patch. I'll add it.

spytheman:
@mike-ward happy to help.

Good work on the effect, the code is very readable.

medvednikov:
https://x.com/v_language/status/1810380646348632508

PR #21816, opened 2024-07-07
checker: check expr evaluated but not used (fix #21436)

### LINKED ISSUE #21436: Standalone brace block containing only literals
**V doctor:**
```
V full version: V 0.4.5 387af74.a1f3624
OS: linux, "Gentoo Linux"
Processor: 4 cpus, 64bit, little endian, AMD Ryzen 3 3200G with Radeon Vega Graphics

getwd: /home/stew/Projects/vtech
vexe: /home/stew/Gitclones/tools/v/v
vexe mtime: 2024-05-05 17:59:50

vroot: OK, value: /home/stew/Gitclones/tools/v
VMODULES: OK, value: /home/stew/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.43.0
Git vroot status: weekly.2024.18.2-19-ga1f36240
.git/config present: true

CC version: cc (Gentoo 13.2.1_p20240113-r1 p12) 13.2.1 20240113
thirdparty/tcc: N/A

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg src/main.v`
```v
fn main() {
  {1, 2, 3, 4}
  println("works")
}

```

Syntax error? The code is unusual, and from what I've been reading in the docs so far, functions require a return statement when returning things instead of just dropping them like in rust. The main() function doesn't even have a return type.

**What did you see instead?**
```
works
```
> [!NOTE]
>

yuyi98:
This PR check expr evaluated but not used (fix #21436).

- Check expr evaluated but not used.
- Add test.

```v
fn main() {
	{1, 2, 3, 4}
	println('works')
}

PS D:\Test\v\tt1> v run .
tt1.v:2:3: error: expression evaluated but not used
    1 | fn main() {
    2 |     {1, 2, 3, 4}
      |      ^
    3 |     println('works')
    4 | }
```

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -2219,6 +2219,12 @@ fn (mut c Checker) block(mut node ast.Block) {
 		c.stmts(mut node.stmts)
 		c.inside_unsafe = prev_unsafe
 	} else {
+		if node.stmts.len > 0 && node.stmts.last() is ast.ExprStmt {
+			last_stmt := node.stmts.last() as ast.ExprStmt
+			if last_stmt.expr !in [ast.CallExpr, ast.IfExpr, ast.MatchExpr, ast.InfixExpr] {
+				c.error('expression evaluated but not used', node.stmts.last().pos)
+			}
+		}
 		c.stmts(mut node.stmts)
 	}
 }
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -2222,7 +2222,7 @@ fn (mut c Checker) block(mut node ast.Block) {
 		if node.stmts.len > 0 && node.stmts.last() is ast.ExprStmt {
 			last_stmt := node.stmts.last() as ast.ExprStmt
 			if last_stmt.expr !in [ast.CallExpr, ast.IfExpr, ast.MatchExpr, ast.InfixExpr] {
-				c.error('expression evaluated but not used', node.stmts.last().pos)
+				c.warn('expression evaluated but not used', node.stmts.last().pos)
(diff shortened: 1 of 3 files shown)


vlib/v/checker/tests/expr_evaluated_but_not_used.vv
```diff
@@ -0,0 +1,4 @@
+fn main() {
+	{1, 2, 3, 4}
+	println('works')
+}
```
spytheman, line 4: This may be a better test input, since it will make the error point the last expression statement in the block:
```v
fn main() {
    {
        println(1)
        println(2)
        println(3)
        println(4)
        5
    }
    println('works')
}
```

vlib/v/checker/checker.v
```diff
@@ -2219,6 +2219,12 @@ fn (mut c Checker) block(mut node ast.Block) {
 		c.stmts(mut node.stmts)
 		c.inside_unsafe = prev_unsafe
 	} else {
+		if node.stmts.len > 0 && node.stmts.last() is ast.ExprStmt {
+			last_stmt := node.stmts.last() as ast.ExprStmt
+			if last_stmt.expr !in [ast.CallExpr, ast.IfExpr, ast.MatchExpr, ast.InfixExpr] {
+				c.error('expression evaluated but not used', node.stmts.last().pos)
```
spytheman, line 2225: I think that a warning may be better than an error, at least for now.

yuyi98: Ok，I'll change it.

spytheman (approved):

PR #21818, opened 2024-07-07
vrepl: fix arbitrary script execute (fix #13981)

### LINKED ISSUE #13981: V REPL passes input directly into println, allowing arbitrary script execution
**V version:** V 0.2.4 509367b
**OS:** windows, Microsoft Windows 10 Pro Education v19043 64-bit

**What did you do?** I found that the V REPL passes the input into `println` (reproduction instructions: typing `foo` will call `println(foo)`, which will return an error as `foo` is undefined), so in an attempt to see if code injection was possible, I typed `"test") exit(-1)`.

**What did you expect to see?** A syntax error: there are mismatched parentheses in the above expression.

**What did you see instead?** The code was substituted into `println("test") exit(-1)`, which **IS** valid V. This prints `test` with a newline and exits the REPL with a code of `-1`.

yuyi98:
This PR fix arbitrary script execute (fix #13981).

- Fix arbitrary script execute.
- Add test.

```v
PS D:\Vlang\v> v
 ____    ____
 \   \  /   /  |  Welcome to the V REPL (for help with V itself, type  exit , then run  v help ).
  \   \/   /   |  Note: the REPL is highly experimental. For best V experience, use a text editor,
   \      /    |  save your code in a  main.v  file and execute:  v run main.v
    \    /     |  V 0.4.6 df18047 . Use  list  to see the accumulated program so far.
     \__/      |  Use Ctrl-C or  exit  to exit, or  help  to see other available commands.

>>> "text") exit(-1)
error: expression evaluated but not used
    5 | import math
    6 |
    7 | "text") exit(-1)
      | ~~~~~~
>>>
```

### DIFF
--- a/cmd/tools/vrepl.v
+++ b/cmd/tools/vrepl.v
@@ -353,15 +353,6 @@ fn run_repl(workdir string, vrepl_prefix string) int {
 		if line.len <= -1 || line == 'exit' {
 			break
 		}
-		if exit_pos := line.index('exit') {
-			oparen := line[(exit_pos + 4)..].trim_space()
-			if oparen.starts_with('(') {
-				if closing := oparen.index(')') {
-					rc := oparen[1..closing].parse_int(0, 8) or { panic(err) }
-					return int(rc)
-				}
-			}
-		}
 		r.line = line
 		if r.line == '\n' {
 			continue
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

Issue #21807, opened 2024-07-05
net.peer_ip() returns ip address and port number 


fleximus:
### Describe the bug

From the official documentation:
> ```v
> fn (c &TcpConn) peer_ip() !string
> ```
> peer_ip retrieves the ip address used by the peer, and returns it as a string

In fact it returns the same result as if `peer_addr()` was called.

See: https://github.com/vlang/v/blob/56d2ba6281536f098c6de9976bcde3a1e83be923/vlib/net/tcp.c.v#L298-L301

### Reproduction Steps

```v
import net

const addr = "127.0.0.1:45123"

fn main() {
    mut listener := net.listen_tcp(.ip, addr)!
    println("Listening on ${addr}, wating for connection.")
    mut conn := listener.accept()!

    ip := conn.peer_ip()!
    if ip == "127.0.0.1" {
        println("OK")
    }
    else {
        eprintln("ERR: IP returned '${ip}' and '127.0.0.1' was the expected result")
    }
}
```
Compile, run and check with `curl http://127.0.0.1:45123`

### Expected Behavior

OK

### Current Behavior

ERR: IP returned '127.0.0.1:xxx' and '127.0.0.1' was the expected result

### Possible Solution

1. Omit colon and port from output in `net.peer_ip()` function
2. Consider renaming `examples/net_peer_ip.v` to `examples/net_peer_addr.v`

### V version

V 0.4.6 56d2ba6

### Environment details (OS name and version, etc.)

V full version: V 0.4.6 96751ed.56d2ba6
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #21831, opened 2024-07-09
net: fix peer_ip to only return the ip address 

Delta456:
Fixes https://github.com/vlang/v/issues/21807


vlib/net/tcp.c.v
```diff
@@ -297,7 +297,15 @@ pub fn (c &TcpConn) peer_addr() !Addr {
 
 // peer_ip retrieves the ip address used by the peer, and returns it as a string
 pub fn (c &TcpConn) peer_ip() !string {
-	return c.peer_addr()!.str()
+	address := c.peer_addr()!.str()
+	if address.contains(']:') {
+		// ipv6 addresses similar to this: '[::1]:46098'
+		ip := address.all_before(']:').all_after('[')
+		return ip
+	}
```
Delta456, line 305: I forgot to consider this case. thanks @spytheman :smile:

spytheman: Happy to help.

PR #21841, opened 2024-07-11
Add digital_rain.v to examples/gg

mike-ward:
This pull request implements the digital rain effect from the movie, "The Matrix", using `gg`. It is not feature complete, but decided getting some feedback was worth starting a pull request. Features needed:
- Window resizing logic
- Start the program in full screen. `gg.toggle_fullscreen()` didn't work for me. Maybe I'm using it wrong
- Keyboard shortcuts to change the rain speed
- Keyboard shortcut to toggle fullscreen
- Other ideas?

I'm OK with merging as is, but I'll likely add these other features in a few days.

### DIFF
--- a/examples/gg/digital_rain.v
+++ b/examples/gg/digital_rain.v
@@ -17,6 +17,7 @@ mut:
 	char_width   int
 	char_height  int
 	screen_size  gg.Size
+	should_calc  bool = true
 	rain_columns []RainColumn
 }
 
@@ -34,8 +35,6 @@ fn main() {
 }
 
 fn rain(mut app App) {
-	// calc rows and columns to use as matrix
-	app.screen_size = gg.screen_size()
 	// Create the drawing context
 	app.ctx = gg.new_context(
 		bg_color: gx.rgb(0, 0, 0)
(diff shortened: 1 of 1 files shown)


examples/gg/digital_rain.v
```diff
@@ -0,0 +1,142 @@
...
+	}
+	for _ in 0 .. max_height {
+		rc.drops << random_rain_drop()
+	}
+	return rc
+}
+
+fn random_rain_drop() u8 {
+	idx := rand.int_in_range(0, rain_drops.len) or { 0 }
+	return rain_drops[idx]
```
spytheman, line 141:
```suggestion
	return rand.element(rain_drops) or { rain_drops.first() }
```

spytheman: Although that would require something like:
`const rain_drops = '0123456789!@#$%^&*()-=+[]{}|;:<>?~bdjpqtvz'.runes()`

spytheman:
> Keyboard shortcuts to change the rain speed
What do you mean by that? Changing the sleep delay, or?

mike-ward:
Correct. Change the snooze

mike-ward:
@spytheman Thanks for help. It likely would have taken a day or two for me to figure out some of the nuances of `gg`.

spytheman:
You are welcome.

spytheman:
About that:
> no retained graphics in gg, store entire column

You can try `app.ctx.end(how: .passthru)`, instead of `ctx.end()` in the frame callback.
Then the content of the buffer will be kept as it was, and you will draw over it in the next frame.

PR #21845, opened 2024-07-11
flag: add `flag.to_struct[T]()!` and `flag.to_doc[T]()!` + tests

larpon:
This PR adds the functionality of https://github.com/larpon/flags to vlib's `flag` module.

Making it possible to map different styles of command line flags/options directly to a V `struct` via reflection.
It can also generate (customizable) "usage" documentation from the `struct`.

I've included two examples - one showing how you can make animated usage/help and one which is a small `WYSIWYG` editor, demoing most of the documentation layout features.

The flag styles supported are:
* POSIX (short and repeatable short) `-v`, `-vvv vv`
* GNU (long) `--long` and `--long=value`
* POSIX/GNU (long+short) in combination
* Go `flag` module style `-l`, `-long`, `-long value` + `--long` and `--long=value`
* V style `-l`, `-long`, `-long value`
* CMD.EXE style `/m`, `/M` or `/Moo` (activated by passing `style: .cmd_exe, delimiter: '/'` config params)

Please note that currently the POSIX/GNU style **is the default**.
I've settled on this since the vast majority of command line tools I've worked with use this style combination.
I'm not sure if we should make the V style (or something else) the default instead :shrug:

### DIFF
--- a/vlib/flag/flag_to.v
+++ b/vlib/flag/flag_to.v
@@ -30,7 +30,8 @@ pub enum Style {
 }
 
 struct StructInfo {
-	name   string
+	name   string // name of the struct itself
+	attrs  map[string]string // collection of `@[x: y]` sat on the struct, read via reflection
 	fields map[string]StructField
 }
 
@@ -85,7 +86,7 @@ pub:
 	delimiter string = '-' // delimiter used for flags
 	style     Style  = .short_long // expected flag style
 pub mut:
-	name        string = os.file_name(os.executable()) // application name
+	name        string            // application name
 	version     string            // application version
 	description string            // application description
 	footer      string            // application description footer written after auto-generated flags list/ field descriptions
@@ -153,11 +154,18 @@ fn (fm FlagMapper) dbg_match(flag_ctx FlagContext, field StructField, arg string
 
 fn (fm FlagMapper) get_struct_info[T]() !StructInfo {
 	mut struct_fields := map[string]StructField{}
+	mut struct_attrs := map[string]string{}
 	mut struct_name := ''
 	mut used_names := []string{}
 	$if T is $struct {
 		struct_name = T.name
 		trace_println('${@FN}: mapping struct "${struct_name}"...')
+
+		$for st_attr in T.attributes {
+			if st_attr.has_arg && st_attr.kind == .string {
+				struct_attrs[st_attr.name.trim_space()] = st_attr.arg.trim(' ')
+			}
+		}
 		// Handle positional first so they can be marked as handled
 		$for field in T.fields {
 			mut hints := FieldHints.zero()
@@ -168,7 +176,7 @@ fn (fm FlagMapper) get_struct_info[T]() !StructInfo {
 				trace_println('\tattribute: "${attr}"')
 				if attr.contains(':') {
 					split := attr.split(':')
-					attrs[split[0].trim(' ')] = split[1].trim(' ')
+					attrs[split[0].trim_space()] = split[1].trim(' ')
 				} else {
 					attrs[attr.trim(' ')] = 'true'
 				}
@@ -274,6 +282,7 @@ fn (fm FlagMapper) get_struct_info[T]() !StructInfo {
 
 	return StructInfo{
 		name: struct_name
+		attrs: struct_attrs
 		fields: struct_fields
 	}
 }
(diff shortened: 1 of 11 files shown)


JalonSolov:
The argument could be made that since it's a V module, it should use V style by default, but I agree that the POSIX/GNU style is a lot more prevalent.  Then again, there's a lot more GNU code than there is V code.  :-)

larpon:
> The argument could be made that since it's a V module, it should use V style by default, but I agree that the POSIX/GNU style is a lot more prevalent. Then again, there's a lot more GNU code than there is V code. :-)

Yeah 🙂 it is a bit tricky to decide 😅

JalonSolov:
When in doubt... it's a V module, so make the default V style.

As long as the others are documented well enough it's easy for someone to switch between them, there shouldn't be any problem doing so.

larpon:
> When in doubt... it's a V module, so make the default V style.
> 
> As long as the others are documented well enough it's easy for someone to switch between them, there shouldn't be any problem doing so.

We could also make it so the user have to explicitly choose a style via `@[required]`?

larpon:
I can not figure out why the `-autofree` self compile test fails 🤔

spytheman:
I've filed https://github.com/vlang/v/issues/21849 with a minimal example of the autofree problem.

spytheman:
> We could also make it so the user have to explicitly choose a style via @[required]?

I think that there is no need for that, it will just create more verbosity.
 The POSIX style is more prevalent for a reason, it is a good default.

spytheman:
> Yeah 🙂 it is a bit tricky to decide 😅

The trick imho, is to be bold and arrogant, when making proclamations. Also, to be prepared to fight in long bike shedding wars 😄 .

spytheman:
I really wish, that in time, we can migrate the existing v command line parser to this declarative approach. It can reduce a lot the amount of boilerplate in `parse_args_and_show_errors` in vlib/v/pref/pref.v .

Unlike the previous time, when a declarative approach was attempted, where the complexity of adding new options *increased*, I think that it can now reduce it.
(besides, v already does `import flag`, through `v.pkgconfig`)

vlib/flag/flag_to.v
```diff
@@ -0,0 +1,1535 @@
...
+}
+
+// to_doc returns a "usage" style documentation `string` generated from
+// the internal data structures generated via the `parse()` function.
+pub fn (fm FlagMapper) to_doc(dc DocConfig) !string {
+	mut docs := []string{}
+
+	// resolve name and version
+	if dc.options.show.has(.name_and_version) {
+		mut app_name := os.file_name(os.executable())
```
spytheman, line 567: I do not like that, since it makes `flag` dependent on `os` for just this line alone. 
Can it be avoided?

spytheman: os.file_name will also return `v.exe` vs `v`, depending on the platform.

spytheman: The FlagMapper already does have access to the input, can we just use .input[0] instead?

medvednikov:
> Go flag module style (-flag, -flag-name and GNU long)
> V style (-v,-version)

@larpon I copied the flag style from Go, how is V different?

vlib/flag/flag_to.v
larpon: > I do not like that, since it makes `flag` dependent on `os` for just this line alone. Can it be avoided?

Good point. After adding the struct attribute support you can also use:
``````v
@[name: my_exe]
struct Conf {}
```
Let me see what I can do about it

larpon: > The FlagMapper already does have access to the input, can we just use .input[0] instead?

Inferring via `.input[0]` will be an odd side-effect to what you can pass as `input` since right now you can pass a `[]string` array that contains anything, even *just* flags if you for some reason only have an array with flags, like: `['--flag, '-g', 'xyz']`

larpon:
> > Go flag module style (-flag, -flag-name and GNU long)
> > V style (-v,-version)
> 
> @larpon I copied the flag style from Go, how is V different?

Go's `flag` module also supports GNU long `--long`/`--long=val` - and, historically, from what I understand we've been trying to get rid of those in V (only some left, like `--version`  and `-v` due to very frequently used or historical reasons).

vlib/flag/flag_to.v
larpon: I'll think about it - maybe it could be worth to dublicate `os.executable()`?

spytheman: > Inferring via .input[0] will be an odd side-effect

I agree. Tagging is better.

larpon: Ouch `os.executable` is very C lib dependent and quite complex... :thinking:

spytheman: > I'll think about it - maybe it could be worth to dublicate os.executable() (and make it strip the .exe part)?

No, it has *a lot* of complexity in it.

(sorry, instead of responding, initially, I edited your answer 🤦🏻‍♂️ . Not sure why github allowed me to do it, it is not part of the PR, but an opinion 🤔 )

larpon: We could make the [`name`](https://github.com/vlang/v/pull/21845/files#diff-51189a54e0ae5b8dd3bf332084aed94d9d34f8f712da44289071825381f8c271R89) field on `DocConfig` `@[required]` :thinking:

larpon: ... so ``flag.to_doc(name: 'xyz')` is required to be specified

JalonSolov:
> > When in doubt... it's a V module, so make the default V style.
> > As long as the others are documented well enough it's easy for someone to switch between them, there shouldn't be any problem doing so.
> 
> We could also make it so the user have to explicitly choose a style via `@[required]`?

Except that would be a breaking change for everything that uses the current `flags`.

spytheman:
> Except that would be a breaking change for everything that uses the current flags.

It would not break. The existing imperative system is still there, and is not modified by this change.

The new addition is a *separate* declarative parser, with 2 new generic APIs: `flag.to_struct` and `flag.to_doc`. Existing programs that do not call them, will not have to be changed.

spytheman:
Is the argument, that the default styles for the 2 parser systems should match?

JalonSolov:
No, I didn't clue in on it being 2 separate parser systems.  That being the case, no they don't have to match.

vlib/flag/flag_to.v
larpon: I have pushed a [864c16b](https://github.com/vlang/v/pull/21845/commits/864c16b815b60c67b70d5b8766cab0a474c18169) that removes the `os` module dependency and have made `name_and_version` two separate things. I do not see how we can provide a default name without `os` - so right now the behavior is the same as with the other optional `footer` or `description` fields

larpon: Requiring `name` in `flag.to_doc(name: 'xyz')` would make the `@[name: x]` struct attribute superfluous so I think settling on making name and version optional is a better compromise at this point.

spytheman:
I think it is good to merge as it is now.

Excellent work @larpon .

felipensp:
Excellent contribution @larpon  👍🏻

PR #21851, opened 2024-07-12
net.http, veb, vweb, x.vweb: add mime type `text/x-vlang`, for `.v` and `.vsh` file extensions

spytheman:


spytheman:
> Not `.vv`? Tsk.

They are intended only for compiler tests, i.e. in this repo. 

Do you think it would be useful to serve them through a webserver for some reason?

JalonSolov:
Off-hand, I can only think of one reason... if we had a web site that showed examples of how to set up these test files for the compiler.  However, why not include them in general, since it is a very minimal change?

Issue #21547, opened 2024-05-22
examples/sokol/02_cube_glsl VALIDATE_ABND_FS_EXPECTED_SAMPLER_BINDING


janEntikan:
### Describe the bug

Running `examples/sokol/02_cube_glsl` triggers in error each frame. 

### Reproduction Steps

Run the example:
`v shader .`
`v run .`

### Expected Behavior

The shader contains code for a raycaster, I was expecting to see it.

### Current Behavior

Nothing from the buffer/glsl is rendered, only the direct drawing from the previous example.

```
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 0, item_id: 285, fname: /home/enti/v/thirdparty/sokol/sokol_gfx.h, line: 15557, message: VALIDATION_FAILED: validation layer checks failed
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 1, item_id: 267, fname: /home/enti/v/thirdparty/sokol/sokol_gfx.h, line: 16328, message: VALIDATE_ABND_FS_EXPECTED_SAMPLER_BINDING: sg_apply_bindings: sampler binding on fragment stage is missing or the sampler handle is invalid
```

### V version

Latest master

### Environment details (OS name and version, etc.)

V full version: V 0.4.6 1b5af1f.7a36b44
OS: linux, "Arch Linux"
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i7-4510U CPU @ 2.00GHz

getwd: /home/enti/v/examples/sokol/02_cubes_glsl
vexe: /home/enti/v/v
vexe mtime: 2024-05-22 14:25:25

vroot: OK, value: /home/enti/v
VMODULES: OK, value: /home/enti/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.45.1
Git vroot status: weekly.2024.20-42-g7a36b44c-dirty
.git/config present: true

CC version: cc (GCC) 14.1.1 20240507
thirdparty/tcc status: thirdparty-linux-amd64 40e5cbb5

Bug was also confirmed by another user.
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Artemixed:
[...]

PR #21853, opened 2024-07-12
sokol: added sampler to the 02_cube code, bugfix: #21547

penguindark:
**What's inside**
- fix bug #21547 for `sokol\02_cubes_glsl` where glsl shader is not running

The new version of sokol seem need an explicit declaration of the sampler to pass ot shader.
Adding the line: `app.cube bind.fs.samplers[C.SLOT_smp] = app.sampler` solve the problem


spytheman (approved):

spytheman:
Thank you 🙇🏻‍♂️!

That change, saved me a lot of frustration and unsuccessful future attempts to fix it on my own (I already tried 3 times).

larpon:
@penguindark thank you! I gave it a try a while ago and could not figure it out either 😅

penguindark:
Only experience nothing more. 
The only important thing is that we have solved the bug. 😄

PR #21858, opened 2024-07-12
docs: add missing @ on export attributes

JalonSolov:
Noticed a few places where `@` was missing from `@[export]` attributes in docs/help and one comment.


vlib/v/ast/ast.v
Delta456, line 1: This file will be modified by PR https://github.com/vlang/v/pull/21860

larpon (approved):
Nice catch!

vlib/v/ast/ast.v
spytheman: That PR will have to be updated anyway, since many of the CI jobs are failing.

PR #21865, opened 2024-07-13
flag: add `flag.using[T]()!` that uses an existing instance of `T`

larpon:
This PR adds the ability to map flags to an *existing* instantiated ("defaults") struct. This makes it possible to map flags in cases where the user [might do exotic things](https://github.com/vlang/vab/blob/48c4093a86f3ee3763296772e697fc076b1307ff/cli/cli.v#L103) or simply needs a bit more flexibility while handling flags

**EDIT** other alternative names, I've thought about, for the function could also be `user_struct`, `use_struct`, `on_struct`, `onto_struct`, `merge_struct`, `via_struct`, `to_default_struct`

`from_struct` might imply that we are generating flags from a struct 🤔 (which may even be desirable to have?! 😬 )

### DIFF
--- a/vlib/flag/flag_to.v
+++ b/vlib/flag/flag_to.v
@@ -297,7 +297,8 @@ fn (m map[string]FlagData) query_flag_with_name(name string) ?FlagData {
 }
 
 // to_struct returns `T` with field values sat to any matching flags in `input`.
-pub fn to_struct[T](input []string, config ParseConfig) !(T, []int) {
+// to_struct also returns an array of strings with flags, in order of appearance, in `input` that could not be matched.
+pub fn to_struct[T](input []string, config ParseConfig) !(T, []string) {
 	mut fm := FlagMapper{
 		config: config
 		input: input
@@ -307,10 +308,11 @@ pub fn to_struct[T](input []string, config ParseConfig) !(T, []int) {
 	return st, fm.no_matches()
 }
 
-// from_struct returns `defaults` with field values sat to any matching flags in `input`.
-// Thus any field(s) that could not be matched with a flag will have the same value as the
+// using returns `defaults` with field values overwritten with any matching flags in `input`.
+// Any field that could *not* be matched with a flag will have the same value as the
 // field(s) passed as `defaults`.
-pub fn from_struct[T](defaults T, input []string, config ParseConfig) !(T, []int) {
+// using also returns an array of strings with flags, in order of appearance, in `input` that could not be matched.
+pub fn using[T](defaults T, input []string, config ParseConfig) !(T, []string) {
 	mut fm := FlagMapper{
 		config: config
 		input: input
@@ -334,10 +336,14 @@ pub fn to_doc[T](dc DocConfig) !string {
 	return fm.to_doc(dc)!
 }
 
-// no_matches returns an array of indicies from the `input` (usually `os.args`),
-// that could not be matched against any fields.
-pub fn (fm FlagMapper) no_matches() []int {
-	return fm.no_match
+// no_matches returns an array of flags, in order of appearance, that could *not* be
+// matched against any fields.
+pub fn (fm FlagMapper) no_matches() []string {
+	mut non_matching := []string{}
+	for i in fm.no_match {
+		non_matching << fm.input[i]
+	}
+	return non_matching
 }
 
 // parse parses `T` to an internal data representation.
--- a/vlib/flag/flag_to.v
+++ b/vlib/flag/flag_to.v
@@ -297,7 +297,8 @@ fn (m map[string]FlagData) query_flag_with_name(name string) ?FlagData {
 }
 
 // to_struct returns `T` with field values sat to any matching flags in `input`.
-// to_struct also returns an array of strings with flags, in order of appearance, in `input` that could not be matched.
+// to_struct also returns any flags from `input`, in order of appearance, that could *not* be matched
+// with any field on `T`.
 pub fn to_struct[T](input []string, config ParseConfig) !(T, []string) {
 	mut fm := FlagMapper{
 		config: config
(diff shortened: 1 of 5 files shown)


vlib/flag/flag_to.v
```diff
@@ -303,7 +303,20 @@ pub fn to_struct[T](input []string, config ParseConfig) !(T, []int) {
 		input: input
 	}
 	fm.parse[T]()!
-	st := fm.to_struct[T]()!
+	st := fm.to_struct[T](none)!
```
larpon, line 306: I was trying to avoid this to not break existing userland code (must be few given it was merged *very* recently), but I could not get V to accept no argument as the `none`? Maybe I had other errors shadowing it at the time

vlib/flag/flag_to.v
```diff
@@ -303,7 +303,20 @@ pub fn to_struct[T](input []string, config ParseConfig) !(T, []int) {
...
+	return st, fm.no_matches()
+}
+
+// from_struct returns `defaults` with field values sat to any matching flags in `input`.
+// Thus any field(s) that could not be matched with a flag will have the same value as the
+// field(s) passed as `defaults`.
+pub fn from_struct[T](defaults T, input []string, config ParseConfig) !(T, []int) {
```
spytheman, line 313: what is the `[]int` for?

spytheman: Please expand on its usage/value/purpose, both here and in the documentation of `pub fn (fm FlagMapper) no_matches() []int {`

spytheman:
> This makes it possible to map flags in cases where the user might do exotic things

The link is to code, looking like this:
```v
archs: fp.string('archs', 0, defaults.archs.join(','), 'Comma separated string with any of ${android.default_archs}').split(',')
```

I do not see how the new API will help in that situation. Can you expand on it a bit?

vlib/flag/flag_to.v
spytheman: > must be few given it was merged very recently

Yes, do not worry about it. It is a breaking change, but one that will not affect many.

If you do want to avoid it, you can potentially do something like this:
```v
@[params]
struct FMToStructParams[T] {
    default ?T
}
pub fn (fm FlagMapper) to_struct[T](params FMToStructParams) !T {
...
    if params.default == none {
```

but imho it is not needed, and over complicates things.

vlib/flag/flag_to.v
larpon: Good catch, I will ASAP. (it is an array of indices in the `input` array that could not be matched - it is a flexible way to allow passing unknown flags, that may or may not be known later on in more complex setups)

larpon:
> > This makes it possible to map flags in cases where the user might do exotic things
> 
> The link is to code, looking like this:
> 
> ```v
> archs: fp.string('archs', 0, defaults.archs.join(','), 'Comma separated string with any of ${android.default_archs}').split(',')
> ```
> 
> I do not see how the new API will help in that situation. Can you expand on it a bit?

Rushed explanation, sorry. It was a from-here-and-down marker. The new function allows you to pass a struct that already values assigned (what I refer to as default values), be it default init values or values that have been assigned in other ways. `to_struct` instantiates a new struct. `from_struct` allows you to keep whatever values already assigned on your struct, but overwriting the them if they match a flag.

In vab we first assign any ENV values to `Options`, *then* we expand on it (overwriting if matched) with flags passed from userland. With the current `to_struct` approach thus requires you to merge/overwrite/assign matching flags values manually, whereas `from_struct` will allow to reuse the struct you have already instantiated and any values that are assigned to it's fields already. The link was to show that the old flag parser allows for a flexible default value something `to_struct` makes harder since it instantiates a new struct where only default values can be used as defaults. I hope that clears my intentions up a bit

spytheman:
`config = flag.merge_with[Config](config, some_args_1)!` ?

spytheman:
or `config = flag.using[Config](config, some_args_1)!` ?

vlib/flag/flag_to.v
spytheman: ok, but why an array of indices, instead of an array of the unused string values themselves?

spytheman: Is there a practical use case, that will be easier to handle, using those indices?

larpon: I opt'ed for indices because they preserve the most information, which will give end users the best chance of troubleshooting why a flag/arg didn't get matched, if they so wish - I know it comes with the slight cost of having to look up entries in the original input, which is less intuitive that other possibilities.

It preserves an exact position of unrecognized input which is valuable for e.g. generating clearer error messages.
Positions are also very nice to know if you are dealing with duplicate non-matching entries.

Secondarily, it is also very valuable to know for developers when dealing with any parsing/matching bugs.

I'm not strongly against changing it to let `no_match`/`[]int` be e.g. `[]string` - we just loose position info that can be nice to have in some cases :shrug:

larpon:
> or `config = flag.using[Config](config, some_args_1)!` ?

I like `using`

vlib/flag/flag_to.v
larpon: I have pushed [09961e7](https://github.com/vlang/v/pull/21865/commits/09961e7816acf4caea5284ba09640cf8f7cd4280) that changes `[]int` to (and documents) `[]string` used as holding any non-matching flags. After more thought I think you have good point in letting the non-matches be `[]string` since this makes the call site less complex and I can't think of any cases where you can't deduct the original index in the input array if you need it.

The commit also renames `from_struct` to the more clear `using` you suggested.

vlib/flag/flag_to.v
larpon: Yes, I also ran into errors that forced me to specify `T` even though I omitted `params` completely. I think that may be a bug - but it needs more investigation and a smaller MRE :slightly_smiling_face:

vlib/flag/flag_to.v
larpon: I also try to clarify the doc strings for the `[]string` returns :sweat_smile:

spytheman (approved):

PR #21871, opened 2024-07-14
table: fix high order generic func resolution

### LINKED ISSUE #21214: generic higher order functions can only be instantiated once
### Describe the bug

When defining higher order functions with a generic type parameter, the first instance of the higher order function works fine, but subsequent instances get confused what type they should expect.

### Reproduction Steps

```
pub type MyFn[T] = fn (mut my_state T)

fn higher_order_fn[T](initial_state T, callback MyFn[T]) T {
    mut my_state := initial_state

    callback[T](mut my_state)
    callback[T](mut my_state)

    return my_state
}

pub struct Stateless { }

pub fn higher_order_fn_stateless(callback MyFn[Stateless]) {
    initial_state := Stateless { }
    higher_order_fn[Stateless](initial_state, callback)
}

pub struct State {
mut:
    x int
}

fn handler(mut my_state State) {
    my_state.x += 1
}

fn main() {
    initial_state := State { }
    new_state := higher_order_fn[State](initial_state, handler)
    println("${new_state}")
}
```

### Expected Behavior

The program should compile and run fine, as if `higher_order_fn_stateless` were never defined.

### Current Behavior

```
higher_order_fn_error1.v:30:56: error: cannot use `fn (mut State)` as `fn (mut Stateless)` in argument 2 to `higher_order_fn`
   28 | fn main() {
   29 |     initial_state := State { }
   30 |     new_state := higher_order_fn[State](initial_state, handler)
      |                                                        ~~~~~~~
   31 |     println("${new_state}")
   32 | }
```

### Possible Solution

It works fine if you remove `higher_order_fn_stateless`.

I discovered this by trying to make a little wrapper around `gg` that handles the logic associated with `new_streaming_image` and `update_pixel_data` automatically, so that multiple apps could simply write to the image given to them, and have a working app. Some of these apps didn't have any state, so I wanted to make a simpler version of `start_app` that just took callbacks without worrying about defining a different dummy state datastructure for each stateless app.

### V version

V 0.4.5 ad17be5

Git vroot status: weekly.2024.14-26-gad17be5d (6 commit(s) behind V master)
.git/config present: true

thirdparty/tcc: N/A
> [!NOTE]
>

felipensp:
Fix #21214

### DIFF
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -2284,7 +2284,9 @@ pub fn (mut t Table) generic_insts_to_concrete() {
 				FnType {
 					mut parent_info := parent.info as FnType
 					mut function := parent_info.func
-					for mut param in function.params {
+					mut params := function.params.clone()
+					function.params = params
+					for mut param in params {
 						if param.typ.has_flag(.generic) {
 							if t_typ := t.resolve_generic_to_concrete(param.typ, function.generic_names,
 								info.concrete_types)
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -2284,9 +2284,8 @@ pub fn (mut t Table) generic_insts_to_concrete() {
 				FnType {
 					mut parent_info := parent.info as FnType
 					mut function := parent_info.func
-					mut params := function.params.clone()
-					function.params = params
-					for mut param in params {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #21872, opened 2024-07-14
cgen: fix generic type init syntax for primitive types 

### LINKED ISSUE #17028: C error when trying to use an illegal type in a Generic
### Describe the bug

C error when trying to use an illegal type in a Generic

### Expected Behavior

Some error message that `int` is not a valid argument to that generic function (as, after substituting `I{0}`, the result is not a valid V expression: `int { 0 }`)

### Current Behavior

```
==================
/tmp/v_1000/bug.5255347210653372059.tmp.c:12133: error: identifier expected
...
==================

builder error: 
==================
```

### Reproduction Steps

```
type Int = int
type MyIndex = struct { Int }

fn func[I]() I{
	return I{0}
}

fn main() {
	println(func[int]())
}

```

### Possible Solution

_No response_

_No response_

### V version

V 0.3.2 a22dfe0.5dbdb21

Linux

felipensp:
Fix #17028

### DIFF
--- a/vlib/v/tests/generic_init_syntax_test.v
+++ b/vlib/v/tests/generic_init_syntax_test.v
@@ -1,0 +1,14 @@
+type Int = int
+type MyIndex = struct {
+	Int Int
+}
+
+fn func[I]() I {
+	return I{0}
+}
+
+fn test_main() {
+	assert func[f64]() == 0.0
+	assert func[int]() == 0
+	assert func[bool]() == false
+}
--- a/vlib/v/gen/c/struct.v
+++ b/vlib/v/gen/c/struct.v
@@ -148,7 +148,7 @@ fn (mut g Gen) struct_init(node ast.StructInit) {
 			}
 		}
 		inited_fields[field_name] = i
-		if sym.kind != .struct_ {
+		if sym.kind != .struct_ && (sym.kind == .string || !sym.is_primitive()) {
 			if init_field.typ == 0 {
 				g.checker_bug('struct init, field.typ is 0', init_field.pos)
 			}


vlib/v/tests/generic_init_syntax_test.v
```diff
@@ -0,0 +1,14 @@
+type Int = int
+type MyIndex = struct {
+	Int Int
```
spytheman, line 3: what does `Int Int` do?

spytheman: does it embed Int 2 times, or?

felipensp: Just copied from original report... We can remove it.

spytheman: Please remove it. `struct { Int Int }` to me, seems like something, that deserves a checker error. The original example also used a single `Int`.

JalonSolov: Since `Int` is an alias, it is trying to define `Int Int`, which should be an error.  Not only is it trying to define a variable with the same name as an alias, if the alias is resolved, it would be the same as typing `int int`, which is also an error.

If it was a single `Int`, it would still be an error - you can't embed an alias, only another struct.  And if it is resolved, it is really `struct { int }` which is a missing variable name.

felipensp: Yeah, surely we have a missing checker error on this construction.

Delta456: I will make a separate issue for this and PR the fix.

felipensp: > I will make a separate issue for this and PR the fix.

Thanks.

JalonSolov: Well, I suppose you _could_ embed an alias if it was an alias to a struct, but that's the only thing that would work.

felipensp: > Please remove it. `struct { Int Int }` to me, seems like something, that deserves a checker error. The original example also used a single `Int`.

Done.

felipensp: > Please remove it. `struct { Int Int }` to me, seems like something, that deserves a checker error. The original example also used a single `Int`.

Removed. 👍🏻

felipensp: I was testing originally:

```v
fn func[I]() I{
	return I{0}
}

fn main() {
	println(func[int]())
}
```

felipensp: > The current test however does not test the condition in the issue (it runs on master, as it was before merging the PR):
> 
> ```
> #0 20:26:30 ^ master /space/v/oo>git pull
> Already up to date.
> #0 21:31:59 ^ master /space/v/oo>cat > x.v
> fn func[I]() I {
>         return I{0}
> }
> 
> fn test_main() {
>         assert func[f64]() == 0.0
>         assert func[int]() == 0
>         assert func[bool]() == false
> }
> #0 21:41:28 ^ master /space/v/oo>mv x.v x_test.v
> #0 21:41:32 ^ master /space/v/oo>./v x_test.v
> #0 21:41:34 ^ master /space/v/oo>git pull
> remote: Enumerating objects: 8, done.
> remote: Counting objects: 100% (8/8), done.
> remote: Compressing objects: 100% (7/7), done.
> remote: Total 8 (delta 1), reused 1 (delta 1), pack-reused 0
> Receiving objects: 100% (8/8), 43.47 KiB | 856.00 KiB/s, done.
> Resolving deltas: 100% (1/1), completed with 1 local object.
> From github.com:vlang/v
>    aa340472e..615a9a08f  master     -> origin/master
> Updating aa340472e..615a9a08f
> Fast-forward
> #0 21:41:40 ^ master /space/v/oo>cat x_test.v
> fn func[I]() I {
>         return I{0}
> }
> 
> fn test_main() {
>         assert func[f64]() == 0.0
>         assert func[int]() == 0
>         assert func[bool]() == false
> }
> #0 21:42:12 ^ master /space/v/oo>./v self
> V self compiling ...
> V built successfully as executable "v".
> #0 21:42:31 ^ master /space/v/oo>./v x_test.v
> #0 21:42:35 ^ master /space/v/oo>
> ```

Originally I was testing:

```v
fn func[I]() I{
	return I{0}
}

fn main() {
	println(func[int]())
}
```

Issue #21876, opened 2024-07-15
anonymous struct embedding lacks checks


Delta456:
The following code lacks check and compiles successfully.

```v
type Int = int
type MyIndex = struct {
	Int Int
}
```

```v
type Int = int
type MyIndex = struct {
	Int
}
```

These programs must be a checker error because the parent type is not a structure.
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #21877, opened 2024-07-15
checker: improve checks for embed in anon struct 

Delta456:
Fixes https://github.com/vlang/v/issues/21876


vlib/v/parser/struct.v
```diff
@@ -239,6 +239,10 @@ fn (mut p Parser) struct_decl(is_anon bool) ast.StructDecl {
 					typ = p.table.find_type_idx(p.anon_struct_decl.name)
 				} else {
 					start_type_pos := p.tok.pos()
+					if field_name == p.tok.lit && p.tok.lit[0].is_capital() {
+						p.error('field-name `${p.tok.lit}` cannot be same as the type-name')
+						return ast.StructDecl{}
+					}
```
Delta456, line 245: CI parser-silent fails because of this check. I can't find a better way to check or place this check.

Delta456: should be fixed now

vlib/v/checker/checker.v
```diff
@@ -561,6 +561,22 @@ fn (mut c Checker) alias_type_decl(node ast.AliasTypeDecl) {
 						c.error('unknown type `${ct_sym.name}`', node.type_pos)
 					}
 				}
+				for embed_type in parent_typ_sym.info.embeds {
```
medvednikov, line 564: A comment for what this block does

Delta456: Done

vlib/v/checker/checker.v
```diff
@@ -561,6 +561,23 @@ fn (mut c Checker) alias_type_decl(node ast.AliasTypeDecl) {
...
+				}
+				if parent_typ_sym.info.is_anon {
+					// Disallow cases like `type Alias = struct { Int Int }` where Int is `type Int = int`
+					for field in parent_typ_sym.info.fields {
+						if field.name != '' && field.anon_struct_decl.fields.len == 0
+							&& field.anon_struct_decl.name == '' {
+							c.error('invalid anonymous struct', node.type_pos)
+						}
+					}
+				}
```
spytheman, line 580: The first error is enough imho, and the error here is more confusing, since it checks for a very specific condition, but prints a very generic 'invalid anonymous struct' without details, about why it is failing.

spytheman: It also produces an error, but it should not, for this valid example:
```v
struct Abc{}
struct Xyz{}

struct TwoEmbeds {
    Abc
    Xyz
}

type MyIndex = struct {
    Abc
    Xyz
}
```

spytheman: On master, both produce these similar C declarations:
```c
struct main__TwoEmbeds {
    main__Abc Abc;
    main__Xyz Xyz;
};

struct main___VAnonStruct1 {
    main__Abc Abc;
    main__Xyz Xyz;
};
```

Delta456:
I have converted this PR into a draft and would require help.

vlib/v/checker/checker.v
Delta456: done

vlib/v/parser/tests/struct_anon_invalid_embed_err.out
```diff
@@ -1,4 +1,4 @@
-vlib/v/parser/tests/struct_anon_invalid_embed_err.vv:2:16: error: invalid anonymous struct
+vlib/v/parser/tests/struct_anon_invalid_embed_err.vv:2:16: error: cannot embed non-struct `Int`
     1 | type Int = int
     2 | type MyIndex = struct {
       |                ~~~~~~~~
```
spytheman, line 4: It still underlines `struct {`. Can it underline the field that was not a struct instead?

Delta456: Done

spytheman (approved):
Excellent work.

PR #21878, opened 2024-07-15
builtin: support for Cyrillic characters in string functions

hedgeg0d:
Adding support for Cyrillic characters in functions to_lower(), is_lower(), to_upper(), is_upper(), capitalize(), uncapitalize(), is_capital(), starts_with_capital(), title()

These function can not recognize Cyrillic symbols because they consist of 2 bytes. For example 
```println('слово'.to_upper) //"Output: "слово"```
After changes these functions can work with these characters
```
println('СЛОВО'.to_lower())
println('привет мир'.capitalize())
println('test тест'.title())
/*
Output:
слово
Привет мир
Test Тест
*/
```


medvednikov:
Thanks!

Go has a more generic implementation that works with all langes.

(btw love your chess game, I hope it gets finished)

Delta456:
Can you add for these?

hedgeg0d:
OK, will do this now

hedgeg0d (commented):
final commit
sorry for mess, tried to pull from browser first

PR #21890, opened 2024-07-18
cgen: fix missing & when map has value ref

### LINKED ISSUE #21889: string interpolation does not correctly reports pointer when used with maps
### Describe the bug

When using string interpolation with maps, the element printed is returned as a value the & of the pointer is not included in the returned string.

### Reproduction Steps

```v
module main

struct Inner {
	value int
}

struct Outer {
	value map[string]&Inner
}

fn main() {
	outer := Outer{
		value: {
			'test': &Inner{
				value: 1
			}
		}
	}
	unexpected := "Outer{
    value: {'test': Inner{
        value: 1
    }}
}"
	expected := "Outer{
    value: {'test': &Inner{
        value: 1
    }}
}"
	assert '${outer}' == unexpected
	assert '${outer}' == expected
}
```

### Expected Behavior

the `expected` string of the example to be returned

### Current Behavior

the `unexpected` string of the example is returned

### Possible Solution

_No response_

_No response_

### V version

V 0.4.6 aa34047

V full version: V 0.4.6 aa34047
OS: macos, macOS, 14.5, 23F79

getwd: /Users/thomas
vexe: /Users/thomas/Unix/local/v/master/v
vexe mtime: 2024-07-15 13:49:47

vroot: OK, value: /Users/thomas/Unix/local/v/master
VMODULES: OK, value: /Users/thomas/Unix/data/v/modules
VTMP: OK, value: /tmp/v_501

Git vroot status: weekly.2024.27-61-gaa340472 (7 commit(s) behind V master)
.git/config present: true

> [!NOTE]
>

felipensp:
Fix #21889

### DIFF
--- a/vlib/v/tests/map_value_ref_interp_test.v
+++ b/vlib/v/tests/map_value_ref_interp_test.v
@@ -1,0 +1,30 @@
+module main
+
+struct Inner {
+	value int
+}
+
+struct Outer {
+	value map[string]&Inner
+}
+
+fn test_main() {
+	outer := Outer{
+		value: {
+			'test': &Inner{
+				value: 1
+			}
+		}
+	}
+	unexpected := "Outer{
+    value: {'test': Inner{
+        value: 1
+    }}
+}"
+	expected := "Outer{
+    value: {'test': &Inner{
+        value: 1
+    }}
+}"
+	assert '${outer}' == expected
+}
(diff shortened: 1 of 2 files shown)


vlib/v/tests/map_value_ref_interp_test.v
```diff
@@ -0,0 +1,30 @@
...
+			'test': &Inner{
+				value: 1
+			}
+		}
+	}
+	unexpected := "Outer{
+    value: {'test': Inner{
+        value: 1
+    }}
+}"
```
Delta456, line 23: Seems unused. Maybe better to remove it.

spytheman (approved):
Excellent work.

PR #21894, opened 2024-07-18
time: increase http_header_string is performance

enghitalo:


spytheman:
What is the performance impact? Have you measured?

medvednikov:
It's definitely going to be faster than string interpolation, but yeah would be nice to measure it.

enghitalo:
```v
import time
import os
import benchmark

const max_iterations = os.getenv_opt('MAX_ITERATIONS') or { '1000' }.int()

const time_to_test = time.Time{
	year: 1980
	month: 7
	day: 11
	hour: 21
	minute: 23
	second: 42
}

fn main() {
	mut b := benchmark.start()

	for _ in 0 .. 1000000 {
		_ := 	time_to_test.http_header_string_old()
	}
	b.measure('http_header_string_old')

	for _ in 0 .. 1000000 {
		_ := 	time_to_test.http_header_string()
	}
	b.measure('http_header_string')

}
```
image

vlib/time/format.v
```diff
@@ -655,10 +655,27 @@ pub fn (t Time) utc_string() string {
 }
 
 // http_header_string returns a date string in the format used in HTTP headers, as defined in RFC 2616.
+// e.g. "Sun, 06 Nov 1994 08:49:37 GMT"
 pub fn (t Time) http_header_string() string {
-	day_str := t.weekday_str()
-	month_str := t.smonth()
-	http_header_string := '${day_str}, ${t.day} ${month_str} ${t.year} ${t.hour:02d}:${t.minute:02d}:${t.second:02d} GMT'
+	day_str := t.weekday_str().bytes()
+	month_str := t.smonth().bytes()
```
spytheman, line 661:
```suggestion
	day_str := t.weekday_str()
	month_str := t.smonth()
```
there is no need for the bytes() calls; later s[0] will give you bytes anyway

vlib/time/format.v
```diff
@@ -655,10 +655,27 @@ pub fn (t Time) utc_string() string {
 }
 
 // http_header_string returns a date string in the format used in HTTP headers, as defined in RFC 2616.
+// e.g. "Sun, 06 Nov 1994 08:49:37 GMT"
 pub fn (t Time) http_header_string() string {
```
spytheman, line 659:
```suggestion
@[manualfree]
pub fn (t Time) http_header_string() string {
```
later you do call buf.free(), and that will not work with -autofree, unless you mark the method here

spytheman:
The vinix CI failure is not related.

spytheman (approved):

PR #21898, opened 2024-07-20
cgen: fix wrong missing unsafe block checking when calling generic functions with @[unsafe] attr

### LINKED ISSUE #21893: carray_to_varray inside unsafe block warns about requiring an unsafe block
### Describe the bug

A warning is emitted when carray_to_varray is used to convert an array of references.

### Reproduction Steps

```v
module main

import arrays

struct Blah {}

fn main() {
	pointers := [&int(1234)]
	length := 1
	array := unsafe { arrays.carray_to_varray[&Blah](pointers, length) }
	println(array)
}
```

### Expected Behavior

No warning when inside unsafe block.

### Current Behavior

> /opt/v/vlib/arrays/arrays.v:705:17: warning: arrays of references need to be initialized right away, therefore `len:` cannot be used (unless inside `unsafe`, or if you also use `init:`)
  703 | @[unsafe]
  704 | pub fn carray_to_varray[T](c_array_data voidptr, items int) []T {
  705 |     mut v_array := []T{len: items}
      |                    ~~~~
  706 |     total_size := items * isize(sizeof(T))
  707 |     unsafe { vmemcpy(v_array.data, c_array_data, total_size) }
[&Blah{}]

### Possible Solution

_No response_

_No response_

### V version

V 0.4.6 510500d

not relevant
> [!NOTE]
>

felipensp:
Fix #21893

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -49,7 +49,7 @@ fn (mut c Checker) fn_decl(mut node ast.FnDecl) {
 	c.fn_level++
 	c.in_for_count = 0
 	c.inside_defer = false
-	c.inside_unsafe = false
+	c.inside_unsafe = node.is_unsafe
 	c.returns = false
 	defer {
 		c.stmt_level = prev_stmt_level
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #21899, opened 2024-07-20
cgen,checker: fix builtin fn var resolve

### LINKED ISSUE #21875: cgen error: '__v_error' undeclared
### Describe the bug

```v
module main

import os
import io
import coachonko.luuid

fn prepare_response(s string) []u8 {
	return '${s}\n'.bytes()
}

// commands end with `\n`
fn main() {
	mut generator := luuid.new_generator()
	stdin := os.stdin()
	mut b_reader := io.new_buffered_reader(reader: stdin)
	mut stdout := os.stdout()
	// TODO this sort of loop kills the processor, another ipc technique must be utilized instead.
	for {
		line := b_reader.read_line() or {
			dump(error)
			break
		} // TODO exit loop when eof and pipe closed. should only be when pipe closed.
		println(line)
		if line == 'v1' {
			id := generator.v1() or { 'error' } // TODO handle error
			res := prepare_response(id)
			_ := stdout.write(res) or { 0 } // TODO when could it fail?
		}
	}
}
```

### Reproduction Steps

Accidentally wrote `dump(error)` instead of `dump(err)` and got the cgen error when compiling the program.

### Expected Behavior

Compiler should tell me error is undeclared instead of producing bad c code

### Current Behavior

bad c code

### Possible Solution

_No response_

_No response_

### V version

`V 0.4.6 aa34047, timestamp: 2024-07-15 15:18:49 +0300`

```
V full version: V 0.4.6 c2f7afd.aa34047
OS: linux, Linux version 5.14.0-427.24.1.el9_4.x86_64 (mockbuild@pp-el9) (gcc (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3), GNU ld version 2.35.2-43.el9) #1 SMP PREEMPT_DYNAMIC Tue Jul 9 00:30:23 CEST 2024
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz

getwd: /home/coachonko/Documents/projects/frontend/wareme/packages/luuid
vexe: /home/coachonko/.local/lib64/v/v
vexe mtime: 2024-07-15 16:55:58

vroot: OK, value: /home/coachonko/.local/lib64/v
VMODULES: OK, value: /home/coachonko/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.43.5
Git vroot status: weekly.2023.45.1-1519-gaa340472
.git/config present: true

CC version: cc (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3)
```
> [!NOTE]
>

felipensp:
Fix #21875

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -4756,7 +4756,7 @@ fn (mut g Gen) ident(node ast.Ident) {
 		g.write(util.no_dots(node.name[2..]))
 		return
 	}
-	mut name := c_name(node.name)
+	mut name := if node.kind == .function { c_fn_name(node.name) } else { c_name(node.name) }
 	if node.kind == .constant {
 		if g.pref.translated && !g.is_builtin_mod
 			&& !util.module_is_builtin(node.name.all_before_last('.')) {
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -38,6 +38,8 @@ pub const reserved_type_names = ['byte', 'bool', 'char', 'i8', 'i16', 'int', 'i6
 pub const reserved_type_names_chk = token.new_keywords_matcher_from_array_trie(reserved_type_names)
 pub const vroot_is_deprecated_message = '@VROOT is deprecated, use @VMODROOT or @VEXEROOT instead'
 
+const builtin_v_fns = ['error']
+
 @[heap; minify]
 pub struct Checker {
 pub mut:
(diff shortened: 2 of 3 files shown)


vlib/v/gen/c/cgen.v
```diff
@@ -4756,7 +4756,7 @@ fn (mut g Gen) ident(node ast.Ident) {
 		g.write(util.no_dots(node.name[2..]))
 		return
 	}
-	mut name := c_name(node.name)
+	mut name := if node.kind == .function { c_fn_name(node.name) } else { c_name(node.name) }
```
spytheman, line 4759: The test passes with just this change in cgen, without the one in the checker.
Why is the checker change needed?

felipensp: Yeah, I realized that the checker part is not really needed.

spytheman (approved):
Excellent work.

PR #21901, opened 2024-07-20
checker,cgen: fix typeof(var.typ) with comptime $for variables

felipensp:
This PR makes possible to read the type name from comptime for loop vars.

```v
type SumType = int | string

struct Test {
	a int
	b string
}

fn get_as_sumtype[T,R](struct_ T, name string, typname string) R {
	$for f2 in T.fields {
		println('field> ${f2.typ} ${typeof(f2.typ).name}')
		if f2.name == name {
			val := struct_.$(f2.name)
			$for f in R.variants {
				println('variant> ${f.typ} ${typeof(f.typ).name}')
				if typeof(f.typ).name == typname {
					return R(val)
				}
			}
		}
	}
	return R{}
}

fn main() {
	println(get_as_sumtype[Test,SumType](Test{a: 123}, 'a', 'int'))
	println(get_as_sumtype[Test,SumType](Test{b: 'foo'}, 'b', 'string'))
}
```

### DIFF
--- a/vlib/v/checker/tests/typeof_comptime_test.vv
+++ b/vlib/v/checker/tests/typeof_comptime_test.vv
@@ -1,0 +1,27 @@
+type SumType = int | string
+
+struct Test {
+	a int
+	b string
+}
+
+fn get_as_sumtype[T, R](struct_ T, name string, typname string) R {
+	$for f2 in T.fields {
+		println('field> ${f2.typ} ${typeof(f2.typ).name}')
+		if f2.name == name {
+			val := struct_.$(f2.name)
+			$for f in R.variants {
+				println('variant> ${f.typ} ${typeof(f.typ).name}')
+				if typeof(f.typ).name == typname {
+					return R(val)
+				}
+			}
+		}
+	}
+	return R{}
+}
+
+fn test_main() {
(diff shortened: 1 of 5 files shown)


vlib/v/checker/tests/typeof_comptime_test.vv
```diff
@@ -0,0 +1,27 @@
...
+				}
+			}
+		}
+	}
+	return R{}
+}
+
+fn test_main() {
+	println(get_as_sumtype[Test, SumType](Test{ a: 123 }, 'a', 'int'))
+	println(get_as_sumtype[Test, SumType](Test{ b: 'foo' }, 'b', 'string'))
```
spytheman, line 26: There are no assertions here.

spytheman: Without assertions, the code is free to change, and the test suite will still pass, even though the result may not be valid anymore, but that will require manual inspection.

felipensp: Ok, thanks. 👍🏻

spytheman (approved):
Excellent work.

PR #21904, opened 2024-07-21
Add raycaster.v  demo to gg examples

mike-ward:
Demonstrates how raycasting works. The left side shows
the 2D layout of the walls and player. The green lines
represent the field of view of the player.

The right side is a simple 3D projection of the field
of view.

There is no collision detection so yes, you can walk
through walls.

Watch https://www.youtube.com/watch?v=gYRrGTC7GtA to
learn more on how this code works. There are some silly
digression's in the video but the tech content is spot on.

image


spytheman (approved):
Very nice example.

PR #21910, opened 2024-07-23
fmt: change align threshold from 8 to 16

yuyi98:
This PR change align threshold from 8 to 16.

- Change align threshold as dynamic and default is 16.
- Modify all the related files.

### DIFF
--- a/.github/workflows/v_apps_and_modules_compile_ci.yml
+++ b/.github/workflows/v_apps_and_modules_compile_ci.yml
@@ -153,18 +153,18 @@ jobs:
           echo "Build v-analyzer release"
           v build.vsh release
 
-      - name: Format vlang/v-analyzer
-        if: ${{ !cancelled() && steps.build.outcome == 'success' }}
-        run: |
-          cd /tmp/v-analyzer
-          set +e
-          v fmt -c .
-          exit_code=$?
-          if [[ $exit_code -ne 0 && $exit_code -ne 5 ]]; then
-            # Don't fail if there are only internal errors (exit code 5).
-            v fmt -diff .
-            exit 1
-          fi
+      # - name: Format vlang/v-analyzer
+      #   if: ${{ !cancelled() && steps.build.outcome == 'success' }}
+      #   run: |
+      #     cd /tmp/v-analyzer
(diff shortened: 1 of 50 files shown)


spytheman:
Why?

spytheman:
I can not decide on merging this 🤔 .

On one hand, the bigger align threshold does produce results that will be more stable to small changes in the surrounding context (type names, comments etc), i.e. future diffs will be smaller.

On another, this is a purely visual/aesthetic change, that can be a bit breaking to existing projects, that have `v fmt -verify` in their CI, although all they will have to do is `v fmt -w .` to fix them again, which is easy.

medvednikov:
I think it's fine, but why not make this number dynamic?

What if it reaches the new limit, then it's off again.

JalonSolov:
I agree.  Most other languages will simply make all the `=` align with each other, or all the comments, or both.  For example, from the examples in this PR
```v
	free_part        fn (voidptr, voidptr, usize, usize) bool          = unsafe { nil }
	free_            fn (voidptr, voidptr, usize) bool = unsafe { nil }
	can_release_part fn (voidptr, u32) bool            = unsafe { nil }
	allocates_zeros  fn (voidptr) bool  = unsafe { nil }
	page_size        fn (voidptr) usize = unsafe { nil } // not a constant field because some platforms might have different page sizes depending on configs
```
could be formatted as
```v
	free_part        fn (voidptr, voidptr, usize, usize) bool = unsafe { nil }
	free_            fn (voidptr, voidptr, usize) bool        = unsafe { nil }
	can_release_part fn (voidptr, u32) bool                   = unsafe { nil }
	allocates_zeros  fn (voidptr) bool                        = unsafe { nil }
	page_size        fn (voidptr) usize                       = unsafe { nil } // not a constant field because some platforms might have different page sizes depending on configs
```
I know some might not like all the extra spaces, but at least it's consistent across the given lines.

medvednikov:
@yuyi98 I missed this fix
`fmt: change align threshold as dynamic and default is 16`

I think it's good to merge with this one.

medvednikov:
Ah I see you made a different PR which has been merged :)

Issue #21912, opened 2024-07-23
cgenerror when trying to use enum key prematurely


ratkingsminion:
### Describe the bug

When trying to find out how to have enum aliases, I found a way by using `key2 = int(MyEnum.key1)`, after testing it a bit further a cgen error appeared when putting key2 before key1.

### Reproduction Steps

Code: https://play.vlang.io/p/371a7f8fd6

```v
enum MyEnum {
	a = 1
	b = 2
	c = int(MyEnum.a) // works
}

enum MyEnum2 {
	c = int(MyEnum2.a) // cgen error
	a = 1
	b = 2
}
```

### Expected Behavior

A checker error is expected

### Current Behavior

Output:

```
/tmp/v_60000/code.01J3FEXFSYXDYWVPHW0EHY56K7.tmp.c:847: error: 'main__MyEnum2__a' undeclared
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
Exited with error status 1
```

### Possible Solution

---

### Additional Information/Context

---

### V version

V 0.4.6 fd7986c

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.6 fd7986c
OS: linux, Debian GNU/Linux 11 (bullseye) (VM)
Processor: 1 cpus, 64bit, little endian, Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz

getwd: /home/admin/playground
vexe: /home/admin/v/v
vexe mtime: 2024-07-23 09:25:26

vroot: OK, value: /home/admin/v
VMODULES: OK, value: .vmodules
VTMP: OK, value: /tmp/v_0

Git version: git version 2.30.2
Git vroot status: Error: fatal: detected dubious ownership in repository at '/home/admin/v'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v
.git/config present: true

CC version: cc (Debian 10.2.1-6) 10.2.1 20210110
thirdparty/tcc status: Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v/thirdparty/tcc
 Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
To add an exception for this directory, call:

[...]

PR #21913, opened 2024-07-23
checker: require enum values to be declared, before using them in other enum values

Delta456:
Fixes https://github.com/vlang/v/issues/21912


spytheman (approved):
Excellent work.

PR #21914, opened 2024-07-23
skip unused mark explicitly used array and map methods

spytheman:
- **markused: mark explicitly used map and array methods**
- **fix `v -skip-unused test vlib/builtin`, add it to `v test-all`**


felipensp:
Very nice approach.

spytheman:
> Very nice approach.

Thank you.

PR #21924, opened 2024-07-24
checker: check enum field value duplicate 2

yuyi98:
This PR check enum field value duplicate 2.

- Check enum field value duplicate 2.
- Add test.

```v
enum MyEnum {
	a = one
	b
	c = one
}

const one = 1 + 1

fn main() {
	$for val in MyEnum.values {
		println('> name: ${val.name} | value: ${val.value}')
	}
}

PS D:\Test\v\tt1> v run .
tt1.v:4:6: error: enum value `2` already exists
    2 |     a = one
    3 |     b
    4 |     c = one
      |         ~~~
    5 | }
    6 |
```


spytheman (approved):
Excellent work. 
I failed to think of this situation 🙇🏻‍♂️ .

felipensp:
Good catch :-)

PR #21928, opened 2024-07-24
checker: allow embed of type alias anon struct 

Delta456:
This PR overtook the fixes that the https://github.com/vlang/v/pull/21877 merging caused.

```v
type Foo2 = struct { a int }

struct Foo {
	Foo2
}
```

The above code didn't work because the check gave an error for every type (like above).

This PR fixes that and allows the code below to be compiled.

```v
type Int = struct {
	a   int
	min int
	max int
}

struct Abc {
	Int
}

fn main() {
	abc := Abc{Int(struct {10, 20 , 30})}
	println(abc.a) // 10
}
```

Tests are also added so that this doesn't happen again.


spytheman (approved):
Good work.

PR #21931, opened 2024-07-25
v.scanner: use table lookups for very frequently done character checks

spytheman:
This has only the lookup changes from https://github.com/vlang/v/pull/21545 .
The substr_unsafe commit, turned out to be a problem.

- **v.scanner: use precomputed character tables instead of checks**
- **v.scanner: use util.func_char_table in Scanner.ident_name/0**
- **v.scanner: restore the checks in ident_name, instead of using util.func_char_table**
- **v.scanner: split long condition in ident_name into several smaller ones**
- **v.scanner: use substr_unsafe in Scanner.ident_name/0 and Scanner.num_lit/2**
- **Revert "v.scanner: use substr_unsafe in Scanner.ident_name/0 and Scanner.num_lit/2"**
- **v.scanner: move non_whitespace_table to v.util**
- **v.scanner: use util.func_char_table[c] and util.name_char_table[c] directly**
- **v.util: cleanup the table generating functions, by using the builtin u8 methods**

### DIFF
--- a/vlib/v/scanner/scanner.v
+++ b/vlib/v/scanner/scanner.v
@@ -528,6 +528,18 @@ fn (mut s Scanner) ident_number() string {
 	}
 }
 
+const nonwhite_space_bytes_table = fill_nonwhite_space_bytes_table()
+
+fn fill_nonwhite_space_bytes_table() [255]bool {
+	mut bytes := [255]bool{}
+	for c in 0 .. 255 {
+		if !(c == 32 || (c > 8 && c < 14) || c == 0x85 || c == 0xa0) {
+			bytes[c] = true
+		}
+	}
+	return bytes
+}
+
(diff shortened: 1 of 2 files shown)


spytheman:
With clang:
image

spytheman:
With gcc, the speedup is very modest, and sometimes (rarely), the old version is slightly faster (i.e. it is on the margin for error on my machine).

With tcc, the speedup is consistently around 1-2% for the scanner.

felipensp:
Nice, good job!

medvednikov:
Great job! Hyperfine reports ~4% though, not 1-2%?

JalonSolov:
Probably dependent on the hardware where you run it.  CPU speed, memory speed, etc. can all affect code running speed.

PR #21932, opened 2024-07-25
cgen: fix struct ref field with no ref structinit

### LINKED ISSUE #21489: Struct and Reference Struct type mismatch in Struct Default Value
### Describe the bug

_Note: Bar must at least have 1 field or it will be interpreted as nil and work_
```v
struct Foo {
	bar &Bar = Bar{}
}

struct Bar {
	field int
}

fn main() {
	dump(Foo{})
}
```
If type Foo.bar is &Bar but is provided Bar it will throw this C error:
`/tmp/v_60000/../../../../../../box/code.v:10: error: cannot convert 'struct main__Bar' to 'struct main__Bar *'`

If type Foo.bar is Bar but is provided &Bar it will convert &Bar to Bar but Bar's field values will be corrupted like #20969

### Reproduction Steps

```v
struct Foo {
	bar &Bar = Bar{}
}

struct Bar {
	field int
}

fn main() {
	dump(Foo{})
}
```

### Expected Behavior

Proper serialisation of the default field values or V error

### Current Behavior

If type Foo.bar is &Bar but is provided Bar it will throw this C error:
`/tmp/v_60000/../../../../../../box/code.v:10: error: cannot convert 'struct main__Bar' to 'struct main__Bar *'`

If type Foo.bar is Bar but is provided &Bar it will convert &Bar to Bar but Bar's field values will be corrupted like #20969

### Possible Solution

Whilst default field values don't work, setting them at runtime does, so adding those checks at comptime for default field values might help

Found and tested in V playground

### V version

V 0.4.5 0e2b604

```
V full version: V 0.4.5 0e2b604

getwd: /home/admin/playground
vexe: /home/admin/v/v
vexe mtime: 2024-05-11 08:25:23

VTMP: OK, value: /tmp/v_0

.git/config present: true

```
> [!NOTE]
>

felipensp:
Fix #21489


spytheman:
I think it should be a checker error instead.

For example this already is:
```v
struct Bar {
    field int
}

struct Foo {
    bar &Bar
}

mut a := Foo{
    bar:  Bar{}
}
```

... producing:
```
z.v:15:3: error: reference field must be initialized with reference
    9 |     mut a := Foo{
   10 |         bar:  Bar{}
      |         ~~~~~~~~~~~
```

spytheman:
For this:
```v
struct Foo {
    bar &Bar = Bar{}
}
```
... it could require using `bar &Bar = &Bar{}` instead, which is both more explicit, and easier to handle for the compiler (no need to do automatic (de)referencing).

felipensp:
Yes, I do agree. But unfortunately there is code using such syntax... (even on vlib) so it will produce a break.

spytheman:
> Yes, I do agree. But unfortunately there is code using such syntax... (even on vlib) so it will produce a break.

Where are those places? If there are small in number, we can change them manually 🤔 .

spytheman:
If they are many, we could add a checker notice, then fix them gradually, then turn the notice into a warning/error.

felipensp:
> If they are many, we could add a checker notice, then fix them gradually, then turn the notice into a warning/error.

I think just this one:

image

spytheman:
Given that the issue code triggers a cgen error on master, I think they will not be too many.

spytheman:
Are all of them in `@[params]` structs?

felipensp:
> Are all of them in `@[params]` structs?

On V project is just this one, no idea on userland cases.

spytheman:
There is not a difference in the generated code with or without the & at all, 
for `downloader &Downloader = TerminalStreamingDownloader{}`:

```
#0 00:17:06 ^ master /space/v/oo>git diff
diff --git vlib/net/http/download_progress.v vlib/net/http/download_progress.v
index 9811f3e70..f8c9d87e6 100644
--- vlib/net/http/download_progress.v
+++ vlib/net/http/download_progress.v
@@ -23,7 +23,7 @@ mut:
 pub struct DownloaderParams {
        FetchConfig
 pub mut:
-       downloader &Downloader = TerminalStreamingDownloader{}
+       downloader &Downloader = &TerminalStreamingDownloader{}
 }

 // download_file_with_progress will save the URL `url` to the filepath `path` .
#0 00:17:08 ^ master /space/v/oo>./v -o new.c cmd/tools/vdownload.v
#0 00:17:17 ^ master /space/v/oo>git stash
Saved working directory and index state WIP on master: 97432c359 v.scanner: use table lookups for very frequently done character checks (#21931)
#0 00:17:19 ^ master /space/v/oo>./v -o old.c cmd/tools/vdownload.v
#0 00:17:21 ^ master /space/v/oo>colordiff old.c new.c
#0 00:17:25 ^ master /space/v/oo>
```

I think, that it may be because `Downloader` is an interface, not a struct 🤔 .

spytheman:
In any case, I think, that changing the checker to error for `field &Struct = Struct{}`, will not break it.

spytheman (approved):

PR #21943, opened 2024-07-27
parser: update `@include` in templates, to work with relative paths & prevent recursive calls

kylepritchard:
Allows calling a partial template directly especially if the partial was previously included as part of a root template. Reasoning added to discussion [#21868](https://github.com/vlang/v/discussions/21868)

TLDR;

Using ```@include!``` will reference the template relative to the base template directory i.e. `templates` folder. Normal action is to include it relative to the template being called. Allows the partial to be called from any other template within the templates directory and its nested directories.


vlib/v/TEMPLATES.md
```diff
@@ -138,6 +138,33 @@ Project root
...
+    /components
+        /todos
+            -todo.html
+```
+
+When calling `todo.html` directly, then to include `/headers/base.html` use the the following
+
+```html
+
+<div>@include! 'header/base'</div>
```
spytheman, line 165:
```suggestion
<div>@include! 'headers/base'</div>
```

spytheman: The path does not match the description otherwise.

vlib/v/parser/tmpl.v
```diff
@@ -226,6 +226,76 @@ fn vweb_tmpl_${fn_name}() string {
...
+			idx := templates_folder.last_index('templates/')
+			if idx != none {
+				templates_folder = templates_folder.substr(0, idx + 'templates/'.len)
+			}
+
+			file_path := os.real_path(os.join_path_single(templates_folder, '${file_name}${file_ext}'))
+
+			println('templates folder => ${templates_folder}')
+			println('join => ${os.join_path_single(templates_folder, '${file_name}${file_ext}')}')
+			println('file_path => ${file_path}')
```
spytheman, line 270: These calls to println should be removed, commented, or put in `$if trace_... ? {` blocks. In normal operation, the compiler should be silent, only producing notices, warnings or errors when needed.

vlib/v/parser/tmpl.v
```diff
@@ -226,6 +226,76 @@ fn vweb_tmpl_${fn_name}() string {
...
+			// Allow single or double quoted paths.
+			mut file_name := if line.contains('"') {
+				line.split('"')[1]
+			} else if line.contains("'") {
+				line.split("'")[1]
+			} else {
+				s := '@include! '
+				position := line.index(s) or { 0 }
+				p.error_with_error(errors.Error{
+					message: 'path for @include! must be quoted with \' or "'
```
spytheman, line 240: If the only difference between the `@include` and `@include!` is the base path handling, imho extract the common stuff into a separate method, and call that in both, parametrized on the base path, instead of duplicating all this.

spytheman:
Please add a test too. See `vlib/v/tests/tmpl_test.v` which uses `vlib/v/tests/tmpl/base.txt` and `vlib/v/tests/tmpl/include.txt` etc.

vlib/v/parser/tmpl.v
kylepritchard: My bad, forgot to delete them from my own testing

kylepritchard:
I was looking into this again and have a few changes which I did not push yet. I got thrown when making additional tests and found that the test templates where in a folder which was named 'tmpl' instead of 'templates'. The logic I originally put down was to search for the 'templates' folder which is only applicable to vweb/veb programs. I was currently looking into using the p.template_paths so that any included template would be relative to it's calling template. That may remove the need to have an extra directive completely. The alternative was to have the root folder named in the @include line so it is easy to find but the syntax could start to get a little verbose ```@include! [tmpl] 'path/to/template' ```. It would be good to have it working for both vweb calls and normal $tmpl calls too.

kylepritchard:
This PR is a rethink on the [original](https://github.com/vlang/v/pull/21943#issue-2433418223) PR. When an `@include` is called, it will now call a recursive function which will import the called template and scan to see if it in turn imports any more templates. Using recursion it is possible to track the current directory which will allow nested templates to call templates in a higher directory. To provide compatibility with the current implementation, it will walk back up the directory path to find the 'templates' folder if the called file is not in the same directory as the template being rendered.

Furthermore there was a bug in the code for circular references. If a template called itself or one of its children called it again then it would result in a circular reference that would crash the program. Now all called templates are entered into a dependency tree which will check if a circular reference exists and will fail with error message should one be encountered.

Finally all called templates are put in a temporary cache which will be used instead of costly IO operations if a template is reused in the code. This should benefit complex web applications were templates may represent web components and could be used many times in the app.

Note: The original idea of `@include!` has been removed in favour of the fallback to templates folder when finding files.

vlib/v/parser/tmpl_test.v
```diff
@@ -10,3 +10,32 @@ fn test_tmpl_comptime() {
...
+	child_tmpl := $tmpl('templates/nested/child.html')
+	assert child_tmpl.contains(expected)
+}
+
+// Test the grandchild templates calling both parent and grandparent templates
+fn test_tmpl_include_grandchild() {
+	expected := [base, base, base, child, grandchild].join('\n')
+	child_tmpl := $tmpl('templates/nested/nested_deeper/grandchild.html')
+	assert child_tmpl.contains(expected)
+}
```
spytheman, line 41: Are not those the same as the new ones in `./vlib/v/tests/tmpl_test.v` ?

Why have duplicate tests?

kylepritchard: Yes, they are. I added those tests originally to the test file in parser folder then found more tests in the `./vlib/v/tests/tmpl_test.v` folder. Maybe the test file in parser folder could be removed in favour of the main tests file?

spytheman: Yes, please move the `test_tmpl_comptime` that is in `vlib/v/parser/tmpl_test.v`, and delete the file + the related html files.

There is no need to have duplicate tests and setups. It will only make the maintenance harder.

kylepritchard: Tests have been transferred 👍

spytheman: Thank you @kylepritchard 🙇🏻‍♂️ .

spytheman (approved):
Excellent work. It makes the `@include` directive a lot more usable, while it improves the error handling too.

medvednikov:
Sounds good to me.

kylepritchard:
Thanks for the consideration and advice. Hopefully I can help out more in time to come

PR #21944, opened 2024-07-27
checker: fix wrong error message for invalid array attr key

felipensp:
```v
bug.v:1:34: error: wrong field `init1`, expecting `len`, `cap`, or `init`
    1 | arr := []int{len: 100, cap: 200, init1: index * 3}
      |                                  ~~~~~
    2 | println(arr)
```


spytheman (approved):
Good work.

vlib/v/checker/tests/wrong_arr_field_err.out
```diff
@@ -0,0 +1,4 @@
+vlib/v/checker/tests/wrong_arr_field_err.vv:1:34: error: wrong field `init1`, expecting `len`, `cap`, or `init`
+    1 | arr := []int{len: 100, cap: 200, init1: index * 3}
+      |                                  ~~~~~
+    2 | println(arr)
```
spytheman, line 4: Should not this be in vlib/v/parser/tests/ instead?

Issue #21918, opened 2024-07-23
When the length of the array exceeds 1073741824, it will cause the current thread to be in a blocked state


tzSharing:
### Describe the bug

adding an element to an array, when its length exceeds 1073741824, causes the current thread to block without any exception information

### Reproduction Steps

V code：

```v
fn main() {
	mut arr := []u8{}
	buf := [1024]u8{}
	for {
		if arr.len < 1073741400 {
			arr << buf[..]
		} else {
			arr << 1 // when len > 1073741824    thread blocking
		}
		println('len=${arr.len}')
		if arr.len >= max_int { // max_int=2147483647
			break
		}
	}
	println('main  len=${arr.len}')
}
```

output:
```
...
len = 1073729536
len = 1073730560
len = 1073731584
len = 1073732608
len = 1073733632
len = 1073734656
len = 1073735680
len = 1073736704
len = 1073737728
len = 1073738752
len = 1073739776
len = 1073740800
len = 1073741824 
^ C
C:\Users\TZ\Desktop>
```

The main thread is blocked and can only be forcibly interrupted.
I tried running the same code with Kotlin to see the results, and it can be seen that Kotlin did not encounter any thread blocking issues and provided exception information. It may be that the memory requested by the Java virtual machine is too small. Most importantly, it provided location information, which is crucial for locating errors, especially when troubleshooting in complex projects.

kotlin code：
```kotlin
fun main(){
	val arr=mutableListOf<Int>()
	val buf=List(1024,{it})
	while (true){
		arr.addAll(buf)
		println(arr.size)
		if(arr.size>=Int.MAX_VALUE){
			println("Break out of the loop")
			break
		}
	}
	println("main  arr size=${arr.size}")
}
```
output：
```
...
34466816
34467840
34468864
34469888
34470912
34471936
34472960
34473984
34475008
34476032
Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
        at java.base/java.util.Arrays.copyOf(Arrays.java:3512)
        at java.base/java.util.Arrays.copyOf(Arrays.java:3481)
        at java.base/java.util.ArrayList.grow(ArrayList.java:238)
        at java.base/java.util.ArrayList.addAll(ArrayList.java:761)
        at TestKt.main(test.kt:5)
        at TestKt.main(test.kt)
        at java.base/java.lang.invoke.LambdaForm$DMH/0x00000141d600f800.invokeStatic(LambdaForm$DMH)
[...]

PR #21947, opened 2024-07-27
builtin: panic on trying to grow arrays with capacity bigger than 2^31, instead of overflowing a.cap (partial fix for #21918)

spytheman:
This PR limits dynamic array capacity growth to max_int , and makes sure that a panic happens if that needs to be exceeded, instead of silently overflowing the .cap field (thus breaking the invariant about .len <= .cap).

It is a temporary workaround for #21918 , providing a better experience (allows for growing/processing till the actual limit, and a panic with a stacktrace, when the limit is reached, for improved debugging), until V transitions to using 64 bit numbers for the .len and .cap array fields.

### DIFF
--- a/vlib/builtin/array.v
+++ b/vlib/builtin/array.v
@@ -182,9 +182,17 @@ fn (mut a array) ensure_cap(required int) {
 	if a.flags.has(.nogrow) {
 		panic('array.ensure_cap: array with the flag `.nogrow` cannot grow in size, array required new size: ${required}')
 	}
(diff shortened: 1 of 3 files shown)


spytheman:
The performance impact for the V compiler is ~0.9%-1.6% slowdown, when tcc is used :-| .
With -prod -cc clang-18, there is no performance impact.

image

I still think, that the more extensive checks are worth it, since with them, V developers will diagnose problems earlier, with clearer diagnostic messages.

medvednikov:
I'm making int 64 bit right now, fixing tests.

PR #21948, opened 2024-07-27
cgen: fix generic sumtype with repeated concrete type

### LINKED ISSUE #19159: cgen error for `Name[int](123)`, when `type Name[T] = string | int | T`

felipensp:
Fix #19159

### DIFF
--- a/vlib/v/gen/c/auto_str_methods.v
+++ b/vlib/v/gen/c/auto_str_methods.v
@@ -447,6 +447,7 @@ fn (mut g Gen) gen_str_for_union_sum_type(info ast.SumType, styp string, typ_str
 		if typ.idx() in idxs {
 			continue
 		}
+		idxs << typ.idx()
 		typ_name := g.typ(typ)
 		mut func_name := g.get_str_fn(typ)
 		sym := g.table.sym(typ)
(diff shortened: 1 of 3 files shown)


vlib/v/gen/c/auto_str_methods.v
```diff
@@ -482,6 +486,7 @@ fn (mut g Gen) gen_str_for_union_sum_type(info ast.SumType, styp string, typ_str
 			}))'
 			fn_builder.write_string('\t\tcase ${int(typ)}: return ${res};\n')
 		}
+		idxs << typ.idx()
```
spytheman, line 489: imho move it right after the check, so that it is easier to keep in context

felipensp: Makes sense.

spytheman (approved):
Excellent work.

PR #21955, opened 2024-07-28
checker: fix sumtype checking for voidptr variant

felipensp:
Allow passing voidptr to sumtype.

```v
type MySum = string | voidptr
type Alias = voidptr

fn t(a MySum) MySum {
	return dump(a)
}

fn test_main() {
	a := 'foo'
	assert t(a) == MySum('foo')

	b := voidptr(0x100)
	assert t(b) == MySum(voidptr(0x100))
}

```


spytheman (approved):
Excellent work.

PR #21957, opened 2024-07-29
parser: check using comptime veb/vweb without import veb/vweb

yuyi98:
This PR check using comptime veb/vweb without import veb/vweb.

- Check using comptime veb/vweb without import veb/vweb.
- Add test.

```v
fn main() {
	$veb.html('index.html')
}

PS D:\Test\v\tt1> v run .
tt1.v:2:2: error: `$veb` cannot be used without import veb
    1 | fn main() {
    2 |     $veb.html('index.html')
      |     ~~~~
    3 | }
```
```v
fn main() {
	$vweb.html('index.html')
}

PS D:\Test\v\tt1> v run .
tt1.v:2:2: error: `$vweb` cannot be used without import vweb
    1 | fn main() {
    2 |     $vweb.html('index.html')
      |     ~~~~~
    3 | }
```

### DIFF
--- a/examples/vweb/veb_example.v
+++ b/examples/vweb/veb_example.v
@@ -33,6 +33,7 @@ pub fn (mut app App) user_endpoint(mut ctx Context, user string) veb.Result {
 	})
 }
 
+// vfmt off
 pub fn (mut app App) index(mut ctx Context) veb.Result {
 	mut c := 0
 	lock app.state {
@@ -46,13 +47,13 @@ pub fn (mut app App) index(mut ctx Context) veb.Result {
 	show := true
 	hello := 'Hello world from veb, request number: ${c}'
 	numbers := [1, 2, 3]
-	return $vweb.html()
+	return $veb.html()
 }
 
 pub fn (mut app App) custom_template(mut ctx Context) veb.Result {
-	return $vweb.html('custom.html')
+	return $veb.html('custom.html')
 }
-
+// vfmt on
 pub fn (mut app App) show_text(mut ctx Context) veb.Result {
 	return ctx.text('Hello world from veb')
 }
--- a/examples/vweb/veb_example.v
+++ b/examples/vweb/veb_example.v
@@ -33,7 +33,6 @@ pub fn (mut app App) user_endpoint(mut ctx Context, user string) veb.Result {
 	})
 }
 
-// vfmt off
 pub fn (mut app App) index(mut ctx Context) veb.Result {
 	mut c := 0
 	lock app.state {
@@ -53,7 +52,7 @@ pub fn (mut app App) index(mut ctx Context) veb.Result {
 pub fn (mut app App) custom_template(mut ctx Context) veb.Result {
 	return $veb.html('custom.html')
 }
-// vfmt on
+
 pub fn (mut app App) show_text(mut ctx Context) veb.Result {
 	return ctx.text('Hello world from veb')
 }
--- a/vlib/v/parser/comptime.v
+++ b/vlib/v/parser/comptime.v
@@ -106,6 +106,7 @@ fn (mut p Parser) comptime_call() ast.ComptimeCall {
 	}
 	start_pos := p.tok.pos()
 	p.check(.dollar)
+	mut is_veb := false
 	error_msg := 'only `\$tmpl()`, `\$env()`, `\$embed_file()`, `\$pkgconfig()`, `\$vweb.html()`, `\$compile_error()`, `\$compile_warn()`, `\$d()` and `\$res()` comptime functions are supported right now'
 	if p.peek_tok.kind == .dot {
 		name := p.check_name() // skip `vweb.html()` TODO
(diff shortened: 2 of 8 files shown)


vlib/v/parser/comptime.v
```diff
@@ -113,6 +113,14 @@ fn (mut p Parser) comptime_call() ast.ComptimeCall {
...
 			return err_node
 		}
+		import_mods := p.ast_imports.map(it.mod)
+		if name == 'vweb' && 'vweb' !in import_mods && 'x.vweb' !in import_mods {
+			p.error_with_pos('`\$vweb` cannot be used without import vweb', start_pos.extend(p.prev_tok.pos()))
+			return err_node
+		} else if name == 'veb' && 'veb' !in import_mods {
+			p.error_with_pos('`\$veb` cannot be used without import veb', start_pos.extend(p.prev_tok.pos()))
+			return err_node
+		}
```
Delta456, line 123: This can be shifted to the checker stage as well

examples/vweb/veb_example.v
```diff
@@ -33,6 +33,7 @@ pub fn (mut app App) user_endpoint(mut ctx Context, user string) veb.Result {
 	})
 }
 
+// vfmt off
```
spytheman, line 36: Why turn off vfmt?

spytheman: Oh, I see ... It converts $veb back to $vweb ...

vlib/v/parser/comptime.v
spytheman: Why? It seems like a check that does not need type information at all. The parser can do it on its own 🤔 .

spytheman (approved):
Excellent work, but I think that vfmt will need to be changed in this PR too, to keep `$veb.html()` and `$vweb.html()` as they are.

vlib/v/parser/comptime.v
Delta456: Ah ok

yuyi98:
Ok, I'll fix it.

spytheman:
> Ok, I'll fix it.

Thank you 🙇🏻‍♂️ .

vlib/v/parser/comptime.v
```diff
@@ -106,13 +106,25 @@ fn (mut p Parser) comptime_call() ast.ComptimeCall {
...
 			p.error(error_msg)
 			return err_node
 		}
+		import_mods := p.ast_imports.map(it.mod)
+		if name == 'vweb' && 'vweb' !in import_mods && 'x.vweb' !in import_mods {
+			p.error_with_pos('`\$vweb` cannot be used without import vweb', start_pos.extend(p.prev_tok.pos()))
+			return err_node
+		} else if name == 'veb' {
+			if 'veb' !in import_mods {
+				p.error_with_pos('`\$veb` cannot be used without import veb', start_pos.extend(p.prev_tok.pos()))
```
JalonSolov, line 123:
```suggestion
				p.error_with_pos('`\$veb` cannot be used unless importing veb', start_pos.extend(p.prev_tok.pos()))
```

JalonSolov:
Question:  Isn't it past time to tell people to import veb instead of vweb?  vweb was supposed to be deprecated as soon as veb was "live".

vlib/v/parser/comptime.v
yuyi98: Thanks! I'll change it.

vlib/v/parser/comptime.v
```diff
@@ -106,13 +106,25 @@ fn (mut p Parser) comptime_call() ast.ComptimeCall {
...
 	error_msg := 'only `\$tmpl()`, `\$env()`, `\$embed_file()`, `\$pkgconfig()`, `\$vweb.html()`, `\$compile_error()`, `\$compile_warn()`, `\$d()` and `\$res()` comptime functions are supported right now'
 	if p.peek_tok.kind == .dot {
 		name := p.check_name() // skip `vweb.html()` TODO
 		if name != 'vweb' && name != 'veb' {
 			p.error(error_msg)
 			return err_node
 		}
+		import_mods := p.ast_imports.map(it.mod)
+		if name == 'vweb' && 'vweb' !in import_mods && 'x.vweb' !in import_mods {
+			p.error_with_pos('`\$vweb` cannot be used without import vweb', start_pos.extend(p.prev_tok.pos()))
```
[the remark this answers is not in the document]
yuyi98: Thanks! I'll change it.

JalonSolov:
veb was just a copy of vweb at the time it was renamed...

spytheman:
V does support:
```v
@[deprecated: 'use something_else']
@[deprecated_after: '2999-01-01']
module xyz
```

... and you are right that we probably should.

The PR however is unrelated - it only concerns the `$vweb.html()` and `$veb.html()` constructs and their handling.

We can do the deprecation later, in another PR, or just extract `vweb` to its own repo outside vlib, in <https://github.com/vlang/vweb>`, so that it could still be used by the people with legacy apps based on it, while emphasizing that `veb` is the preferred one with recent V.

JalonSolov:
The main reason I brought it up is that the error message about `$vweb.html()` could be changed to mention using veb instead of vweb, as another vector to let people know they "should" switch.
```v
p.error_with_pos('`\$vweb` cannot be used without importing vweb - NOTE: vweb is deprecated, please switch to veb', start_pos.extend(p.prev_tok.pos()))
```
or similar.

PR #21958, opened 2024-07-29
json: increased test cases before enabling sumtype decode in all json libraries

enghitalo:
Increased test cases before enabling sumtype decode in all json libraries


spytheman:
There is no `Dog` defined in your new test.

PR #21965, opened 2024-07-30
docs: add more C interop notes and examples

spytheman:
This PR aims to improve the C interop documentation by providing explanations about some not entirely obvious details for the V implementation and how it currently interacts with the used C backend compiler.

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -7188,7 +7188,7 @@ structs declared in `stdio.h`, but if you try to compile the .v file,
 it will add the include in the generated C code, so that if that header file
 is missing, you will get a C error (you will not in this specific case, if you
 have a proper C compiler setup, since `<stdio.h>` is part of the
-standart C library).
+standard C library).
 
 To overcome that limitation (that V does not have a C parser), V needs you to
 redeclare the C functions and structs, on the V side, in your `.c.v` files.
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -7199,7 +7199,7 @@ Note also that they *do not have* to be complete, unlike the ones in the .h file
 
 **C. struct redeclarations**
 For example, if a struct has 3 fields on the C side, but you want to only
-refer to just 1 of them, you can just declare it like this:
+refer to 1 of them, you can declare it like this:
 
 **Example of C struct redeclaration**
 ```v oksyntax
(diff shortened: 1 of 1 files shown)


doc/docs.md
```diff
@@ -7173,6 +7173,104 @@ For all supported options check the latest help:
...
+**Example of #include**
+```v oksyntax
+#include <stdio.h>
+```
+After this statement, V will *not* know anything about the functions and
+structs declared in `stdio.h`, but if you try to compile the .v file,
+it will add the include in the generated C code, so that if that header file
+is missing, you will get a C error (you will not in this specific case, if you
+have a proper C compiler setup, since `<stdio.h>` is part of the
+standart C library).
```
JalonSolov, line 7191:
```suggestion
standard C library).
```

doc/docs.md
```diff
@@ -7173,6 +7173,104 @@ For all supported options check the latest help:
...
+To overcome that limitation (that V does not have a C parser), V needs you to
+redeclare the C functions and structs, on the V side, in your `.c.v` files.
+Note that such redeclarations only need to have enough details about the
+functions/structs that you want to use.
+Note also that they *do not have* to be complete, unlike the ones in the .h files.
+
+
+**C. struct redeclarations**
+For example, if a struct has 3 fields on the C side, but you want to only
+refer to just 1 of them, you can just declare it like this:
```
JalonSolov, line 7202:
```suggestion
refer to 1 of them, you can declare it like this:
```

doc/docs.md
```diff
@@ -7173,6 +7173,104 @@ For all supported options check the latest help:
...
+@[typedef]
+pub struct C.TypeName {
+}
+```
+Note that the name of the `C.` struct in V, is the one *after* the `struct SomeName {...}`.
+
+**C. function redeclarations**
+The situation is similar for `C.` functions. If you are going to call just 1 function in a
+library, but its .h header declares dozens of them, you will only need to declare that single
+one function like this for example:
```
JalonSolov, line 7225:
```suggestion
function, for example:
```

doc/docs.md
```diff
@@ -7173,6 +7173,104 @@ For all supported options check the latest help:
...
+
+**Example of C function redeclaration**
+```v oksyntax
+fn C.name_of_the_C_function(param1 int, const_param2 &char, param3 f32) f64
+```
+... and then later, you will be able to call it like this:
```
JalonSolov, line 7231:
```suggestion
... and then later, you will be able to call the same way you would V function:
```

JalonSolov (approved):

larpon (approved):
Very nice additions!

spytheman:
Thank you.

Issue #21969, opened 2024-07-31
Possible to define a static type method without defining a type


dadooda:
### Describe the bug

It's possible to define a static type method, as described in [Static type methods](https://docs.vlang.io/structs.html#static-type-methods), without defining the struct or interface.

### Reproduction Steps

This works:

```v
fn main() {
  Mystery.hey()
}

// No definition of `Mystery` here.

fn Mystery.hey() {
  println('hey, ${@METHOD}')
}
```

### Expected Behavior

I'd expect an error, or a warning at least. Most probably, the developer has mistyped the type name or has left out the implementation somehow.

### Current Behavior

The code compiles and gets executed.

### V version

0.4.7 7baff15

### Environment details (OS name and version, etc.)

Ubuntu 18 under WSL.
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

felipensp:
Bug or feature? :)

Delta456:
> Bug or feature? :)

It is bug

PR #21970, opened 2024-07-31
checker: disallow static fn call when receiver type is unknown 

Delta456:
Fixes https://github.com/vlang/v/issues/21969


vlib/v/checker/fn.v
```diff
@@ -923,6 +923,15 @@ fn (mut c Checker) fn_call(mut node ast.CallExpr, mut continue_check &bool) ast.
...
 		}
 	}
+
+	if _ := fn_name.index('__static__') {
+		if sym := c.table.find_sym(node.name.all_before('__static__')) {
+			if sym.kind == .placeholder {
+				c.error('unknown type `${sym.name}`', node.name_pos)
+			}
+		}
+	}
```
spytheman, line 933: Checking this inside `fn_call`, does not seem right to me.

I think it will be more efficient to check this in `fn_decl`, since there are a lot of function calls, and comparatively a lot less function declarations usually.

spytheman: i.e. error on `fn Mystery.hey() {`, and underline `Mystery`, not on `Mystery.hey()` .

Delta456: Ah true. Done!

spytheman (approved):
Excellent work.

PR #21971, opened 2024-07-31
sokol: update to match upstream at c0e0563

larpon:
This PR updates all our sokol headers to upstream commit [c0e0563](https://github.com/floooh/sokol/commit/c0e0563e9026b71cc90eafa3005e514e29e54e84) (latest as per today). All previous patches *should* have been carried over except for the patch in `sokol_audio.h` which has been merged upstream (thanks to @kimshrier :heart:) - support for the new shader versions has also been updated in `vshader`.

It is worth noting that, with this update, the minimum required version of OpenGL is now 4.3 on Linux and 4.1 on macOS (previous version was 3.3). The bump in requirements is, as I understand it, because of the ongoing work towards compute shader support. So with this update, there's introduced support for [storage buffers](https://github.com/floooh/sokol/blob/master/CHANGELOG.md#09-may-2024) in shaders which is the main cause of the bumped OpenGL version requirements. Support for storage buffers on Android is not done according to floooh (the author of sokol).

I do not have a full overview of how much this version change will impact our graphics support in general.

I have tested the sokol, sokol shader and `gg` examples - which all work on my distro. I am unsure if older distros will break :shrug:


larpon:
According to https://en.wikipedia.org/wiki/OpenGL#OpenGL_4.3 OpenGL 4.3 is from August 6, 2012

... it also says that OpenGL 3.3 (previous version) was released two years prior, on March 11, 2010

spytheman:
For me, that updates breaks `sokol` and `gg` apps on my older Linux system:
```
#1 19:36:57 ^ master /v/oo>gh pr checkout 21971
remote: Enumerating objects: 21, done.
remote: Counting objects: 100% (21/21), done.
remote: Compressing objects: 100% (4/4), done.
remote: Total 21 (delta 16), reused 21 (delta 16), pack-reused 0
Receiving objects: 100% (21/21), 19.95 KiB | 6.65 MiB/s, done.
Resolving deltas: 100% (16/16), completed with 16 local objects.
From github.com:vlang/v
 * [new ref]             refs/pull/21971/head -> sokol/update-to-c0e0563
Switched to branch 'sokol/update-to-c0e0563'
#0 19:43:55 ^ sokol/update-to-c0e0563 /v/oo>v run examples/gg/rectangles.v
sokol.memory.slog | user_data: (nil), const_tag: sapp, level: 0, item_id: 35, fname: /space/v/oo/thirdparty/sokol/sokol_app.h, line: 10409, message: LINUX_GLX_CREATE_CONTEXT_FAILED: Failed to create GL context via glXCreateContextAttribsARB
X Error of failed request:  BadMatch (invalid parameter attributes)
  Major opcode of failed request:  152 (GLX)
  Minor opcode of failed request:  5 (X_GLXMakeCurrent)
  Serial number of failed request:  152
  Current serial number in output stream:  152
#1 19:44:02 ^ sokol/update-to-c0e0563 /v/oo>vjed /space/v/oo/thirdparty/sokol/sokol_app.h:10409

[3]+  Stopped                 vjed /space/v/oo/thirdparty/sokol/sokol_app.h:10409
#148 19:44:37 ^ sokol/update-to-c0e0563 /v/oo>v run examples/gg/rectangles.v
X Error of failed request:  BadMatch (invalid parameter attributes)
  Major opcode of failed request:  152 (GLX)
  Minor opcode of failed request:  5 (X_GLXMakeCurrent)
  Serial number of failed request:  152
  Current serial number in output stream:  152
#1 19:44:39 ^ sokol/update-to-c0e0563 /v/oo>fg
```

spytheman:
Everything looks fine on macos 14.5, on my m1.

spytheman:
Everything looks fine on Windows 11 with latest drivers for NVIDIA GeForce RTX 3050 too.

spytheman:
`LIBGL_ALWAYS_SOFTWARE=1 v -d show_fps run examples/gg/spirograph.v` continues to work (at 20fps, instead of 60fps, because of the emulation).

spytheman:
With a small patch, it worked on my old Linux desktop as well:
```
#0 20:24:22 ^ sokol/update-to-c0e0563 /v/oo>git diff
diff --git vlib/gg/gg.c.v vlib/gg/gg.c.v
index fa445f54c..307583a7d 100644
--- vlib/gg/gg.c.v
+++ vlib/gg/gg.c.v
@@ -471,6 +471,7 @@ pub fn new_context(cfg Config) &Context {
                ui_mode: cfg.ui_mode
                native_rendering: cfg.native_rendering
                window: sapp.Desc{
+                       gl_major_version: 3
                        init_userdata_cb: gg_init_sokol_window
                        frame_userdata_cb: gg_frame_fn
                        event_userdata_cb: gg_event_fn
diff --git vlib/sokol/sapp/sapp_structs.c.v vlib/sokol/sapp/sapp_structs.c.v
index c0bb9dc5c..d3c9e5dbd 100644
--- vlib/sokol/sapp/sapp_structs.c.v
+++ vlib/sokol/sapp/sapp_structs.c.v
@@ -66,6 +66,8 @@ pub:
        max_dropped_file_path_length int      // max length in bytes of a dropped UTF-8 file path (default: 2048)
        icon                         IconDesc // the initial window icon to set
        // backend-specific options
+       gl_major_version              int // override GL major and minor version (the default GL version is 4.1 on macOS, 4.3 elsewhere)
+       gl_minor_version              int
        win32_console_utf8            bool  // if true, set the output console codepage to UTF-8
        win32_console_create          bool  // if true, attach stdout/stderr to a new console window
        win32_console_attach          bool  // if true, attach stdout/stderr to parent process
#0 20:24:24 ^ sokol/update-to-c0e0563 /v/oo>
#0 20:24:24 ^ sokol/update-to-c0e0563 /v/oo>
#0 20:24:24 ^ sokol/update-to-c0e0563 /v/oo>
#0 20:24:24 ^ sokol/update-to-c0e0563 /v/oo>v -d show_fps run examples/gg/spirograph.v
#0 20:24:33 ^ sokol/update-to-c0e0563 /v/oo>
```

spytheman:
Shader related sokol examples do not work too on Linux, even with the patch above:
```
#139 20:40:05 ^ sokol/update-to-c0e0563 /v/oo>v shader examples/sokol/05_instancing_glsl/
#0 20:40:09 ^ sokol/update-to-c0e0563 /v/oo>v run examples/sokol/05_instancing_glsl/
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 1, item_id: 5, fname: /space/v/oo/thirdparty/sokol/sokol_gfx.h, line: 8462, message: GL_SHADER_COMPILATION_FAILED: shader compilation failed (gl)
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 3, item_id: 5, fname: /space/v/oo/thirdparty/sokol/sokol_gfx.h, line: 8463, message: 0:1(10): error: GLSL 4.30 is not supported. Supported versions are: 1.10, 1.20, 1.30, 1.40, 1.50, 3.30, 4.00, 4.10, 4.20, 1.00 ES, and 3.00 ES

sokol.memory.slog | user_data: (nil), const_tag: sg, level: 1, item_id: 5, fname: /space/v/oo/thirdparty/sokol/sokol_gfx.h, line: 8462, message: GL_SHADER_COMPILATION_FAILED: shader compilation failed (gl)
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 3, item_id: 5, fname: /space/v/oo/thirdparty/sokol/sokol_gfx.h, line: 8463, message: 0:1(10): error: GLSL 4.30 is not supported. Supported versions are: 1.10, 1.20, 1.30, 1.40, 1.50, 3.30, 4.00, 4.10, 4.20, 1.00 ES, and 3.00 ES

sokol.memory.slog | user_data: (nil), const_tag: sg, level: 1, item_id: 160, fname: /space/v/oo/thirdparty/sokol/sokol_gfx.h, line: 16455, message: VALIDATE_PIPELINEDESC_SHADER: sg_pipeline_desc.shader missing or invalid
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 0, item_id: 298, fname: /space/v/oo/thirdparty/sokol/sokol_gfx.h, line: 16102, message: VALIDATION_FAILED: validation layer checks failed
GLSL March init DONE!
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 1, item_id: 231, fname: /space/v/oo/thirdparty/sokol/sokol_gfx.h, line: 16703, message: VALIDATE_APIP_PIPELINE_VALID: sg_apply_pipeline: pipeline object not in valid state
05_instancing_glsl: /space/v/oo/thirdparty/sokol/sokol_gfx.h:16705: _sg_validate_apply_pipeline: Assertion `pip->shader' failed.
7ffff7a3a00b : at ???: RUNTIME ERROR: abort() called
4173257325203a75 : by ???
007bc41c : at ???: RUNTIME ERROR: invalid memory access
007bbd95 : by ???
007bc0d7 : by ???
007bc20e : by ???
7ffff7f5e420 : by ???
4173257325203a75 : by ???
Terminated by signal 11 (SIGSEGV)
#139 20:40:15 ^ sokol/update-to-c0e0563 /v/oo>
```

spytheman:
They do work, after manually running:
```
/home/delian/.cache/v/vshader/sokol-shdc.exe --input /space/v/oo/examples/sokol/05_instancing_glsl/rt_glsl_instancing.glsl --output /space/v/oo/examples/sokol/05_instancing_glsl/rt_glsl_instancing.h --slang glsl410:glsl300es:hlsl5:metal_macos:metal_ios:metal_sim:wgsl
```

(The default that was used is: `glsl430`. That made the assertions break. Using `glsl410` instead restored the old behaviour.)

spytheman:
This patch to vshader.v, restores the old behaviour (fixes it for my older Linux desktop):
 ```
+++ cmd/tools/vshader.v
@@ -50,7 +50,7 @@ const default_slangs = $if macos {
        ]
 } $else {
        [
-               'glsl430',
+               'glsl410',
```

**Comment by spytheman** [comment]
I think that this is the minimum patch to restore the old state on older Linux:
image

**Comment by spytheman** [comment]
What is weird, is that even just settting `gl_major_version int = 4` works 🤔 .

Perhaps setting an explicit version != 0, makes it request that version on the C side, instead of trying to autodetect the maximum possible.

**Comment by spytheman** [comment]
Yep, changing `thirdparty/sokol/sokol_app.h:3158` from:
```c
    // NOTE: can't patch the default for gl_major_version and gl_minor_version
    // independently, because a desired version 4.0 would be patched to 4.2
    // (or expressed differently: zero is a valid value for gl_minor_version
    // and can't be used to indicate 'default')
    if (0 == res.gl_major_version) {
        #if defined(_SAPP_APPLE)
            res.gl_major_version = 4;
            res.gl_minor_version = 1;
        #else
            res.gl_major_version = 4;
            res.gl_minor_version = 3;
        #endif
    }
```

to using `res.gl_minor_version = 1;` in both branches, is enough to fix the breakage for me.

(the vshader.v patch is still needed)

larpon:
Thanks for the help 🙇🏼‍♂️

I think the ```v
const default_slangs = $if macos { ... } $else { ... }``` is not needed at all then?

JalonSolov:
Are they only going to need to be identical for older versions of linux?  Ubuntu 20.04 is fairly old... they're stopping standard support for it next year.

spytheman:
I think that the OpenGL level of support, depends on the hardware/drivers, more than on the OS itself. Setting `LIBGL_ALWAYS_SOFTWARE=1` for example, works by using an emulated (by Mesa) "hardware", that supports OpenGL 4.3 even on Ubuntu 20.04 (just a lot slower).

spytheman:
> is not needed at all then?

hm, good spot; the 2 branches are identical and can be merged, as they were before indeed.

spytheman:
> Are they only going to need to be identical for older versions of linux? Ubuntu 20.04 is fairly old... they're stopping standard support for it next year.

`glsl410` is the default for macos as well, and works consistently everywhere, unlike `glsl430`, which does not on macos and on linux machines with older GPUs.

I think, that keeping the same default, and minimizing the breakage is better for the already existing sokol and gg apps.

spytheman:
We could probably add an opt in mechanism, based on `-d`, to choose a higher target OpenGL version, while keeping the backwards compatibility.

spytheman:
@larpon thank you very much for the update 🙇🏻‍♂️ .

I know how time consuming, it could be to update a thirdparty dependency like Sokol , and how hard it is to test it well, given all the combinations.

penguindark:
Thanks a lot @larpon 🙇‍♂️

lmptg:
> I think that the OpenGL level of support, depends on the hardware/drivers, more than on the OS itself. Setting `LIBGL_ALWAYS_SOFTWARE=1` for example, works by using an emulated (by Mesa) "hardware", that supports OpenGL 4.3 even on Ubuntu 20.04 (just a lot slower).

This is also my understanding - it has a lot to do with the hardware, vendor and available drivers.

lmptg:
> Are they only going to need to be identical for older versions of linux? Ubuntu 20.04 is fairly old... they're stopping standard support for it next year.

"Back in the day" you could sometimes be lucky that some brave soul had time to maintain a ppa on [launchpad](https://launchpad.net/ubuntu/+ppas) that provided newer drivers + libGL versions. But alas, as time goes by, these ppa's often go offline and/or become unavailable/unmaintained etc.

I'm not sure what camp I'm in when it comes to supporting old linux distros. A part of me think it is a major shame that support for older distros is not better/easier and a part of me can see the benefits of moving forward in some cases. I think I'm mostly in the "keep-things-running-for-as-long-as-possible -camp", though. So in this case @spytheman 's changes makes sense to me. I was on Ubuntu 20.04 as well not long ago (some of the first examples I contributes to V was written on 20.04 :smiling_face_with_tear: )

lmptg:
> @larpon thank you very much for the update 🙇🏻‍♂️ .
> 

> Thanks a lot @larpon 🙇‍♂️

I'm glad to help when I can. None of this would be without any of us :heart:

larpon:
> Are they only going to need to be identical for older versions of linux? Ubuntu 20.04 is fairly old... they're stopping standard support for it next year.

"Back in the day" you could sometimes be lucky that some brave soul had time to maintain a ppa on [launchpad](https://launchpad.net/ubuntu/+ppas) that provided newer drivers + libGL versions. But alas, as time goes by, these ppa's often go offline and/or become unavailable/unmaintained etc.

I'm not sure what camp I'm in when it comes to supporting old linux distros. A part of me think it is a major shame that support for older distros is not better/easier and a part of me can see the benefits of moving forward in some cases. I think I'm mostly in the "keep-things-running-for-as-long-as-possible -camp", though. So in this case @spytheman 's changes makes sense to me. I was on Ubuntu 20.04 as well not long ago (some of the first examples I contributes to V was written on 20.04 :smiling_face_with_tear: )

larpon:
> @larpon thank you very much for the update 🙇🏻‍♂️ .
> 

> Thanks a lot @larpon 🙇‍♂️

I'm glad to help when I can. None of this would be without any of us :heart:

larpon:
(Sorry for the user confusion `lmptg` is a work account, I always mix it up)

Issue #21973, opened 2024-07-31
c.crypto.chacha20: Cipher incorrectly implements method `xor_key_stream` of interface `crypto.cipher.Stream`


Coachonko:
### Describe the bug

[crypto.cipher.Stream xor_key_stream](https://github.com/vlang/v/blob/master/vlib/crypto/cipher/cipher.v#L29) expects immutable implementer.
[x.crypto.chacha20.Cipher xor_key_stream](https://github.com/vlang/v/blob/master/vlib/x/crypto/chacha20/chacha.v#L80) defines the method with mutable implementer.

### Reproduction Steps

```v
struct StreamImplementer {
	cipher cipher.Stream
}

fn test_cipher_implements_stream() ! {
	mut key := []u8{len: 32}
	mut nonce := []u8{len: 12}
	rand.read(mut key)
	rand.read(mut nonce)
	mut c := chacha20.new_cipher(key, nonce)!
	implementer := StreamImplementer{
		cipher: c
	}
	// fails to compile with error:
	// `&chacha20.Cipher` incorrectly implements method `xor_key_stream` of interface `crypto.cipher.Stream`: expected `cipher.Stream` which is immutable, not `mut &chacha20.Cipher`
	// Details: crypto.cipher.Stream has `fn xor_key_stream(x cipher.Stream, mut dst []u8, src []u8)`
        // x.crypto.chacha20.Cipher has `fn xor_key_stream(mut c chacha20.Cipher, mut dst []u8, src []u8)`
}
```

### Expected Behavior

Compiles fine

### Current Behavior

Fails to compile

### Possible Solution

Both `chacha20` and `rc4` implement the interface wrong. Given the complex operations done in the chacha20 Chipher struct in the invocation of this method, reworking chacha20 may be unfeasible. Perhaps the correct approach is to re-define the crypto.cipher.Stream xor_key_stream to have a mutable implementer.

### V version

V 0.4.7 340535f

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.7 c2f7afd.340535f
OS: linux, Linux version 5.14.0-427.26.1.el9_4.x86_64 (mockbuild@pp-el9) (gcc (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3), GNU ld version 2.35.2-43.el9) #1 SMP PREEMPT_DYNAMIC Wed Jul 17 15:28:16 CEST 2024
[...]

PR #21974, opened 2024-07-31
crypto.cipher: make Stream.xor_key_stream implementers require a mutable receiver

Coachonko:
Addresses #21973 

This may impact stream ciphers that implement xor_key_stream with immutable implementers, and may therefore be breaking.
Further checks should be made so that all stream ciphers implement xor_key_stream correctly, tests are missing for this scenario in all stream cipher implementations.

I am not sure how to perform such tests, help is needed to add such tests.

### DIFF
--- a/vlib/crypto/cipher/cfb.v
+++ b/vlib/crypto/cipher/cfb.v
@@ -63,34 +63,34 @@ fn new_cfb(b Block, iv []u8, decrypt bool) Cfb {
 	return x
 }
 
-pub fn (mut x Cfb) xor_key_stream(mut dst_ []u8, src_ []u8) {
+pub fn (mut x Cfb) xor_key_stream(mut dst []u8, src []u8) {
 	unsafe {
-		mut dst := *dst_
-		mut src := src_
-		if dst.len < src.len {
+		mut local_dst := *dst
+		mut local_src := src
+		if local_dst.len < local_src.len {
 			panic('crypto.cipher.xor_key_stream: output smaller than input')
 		}
 
-		if subtle.inexact_overlap(dst[..src.len], src) {
+		if subtle.inexact_overlap(local_dst[..src.len], local_src) {
 			panic('crypto.cipher.xor_key_stream: invalid buffer overlap')
 		}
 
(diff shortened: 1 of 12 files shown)


Coachonko:
Please review tests in particular.

Coachonko (commented):
Seems like errors are thrown in `des_ofb_test.v`, these should be fixed before merge

Coachonko:
Should be ready now

spytheman:
> I am not sure how to perform such tests, help is needed to add such tests.

They way you did it is fine. Another approach is:
```v
fn implements(x cipher.Ctr) ` { return x }
```
... which relies that the compiler will then that `cipher.Ctr` can be converted to a `cipher.Stream` interface value.

vlib/crypto/cipher/cfb.v
```diff
@@ -63,34 +63,34 @@ fn new_cfb(b Block, iv []u8, decrypt bool) Cfb {
...
 
-pub fn (mut x Cfb) xor_key_stream(mut dst_ []u8, src_ []u8) {
+pub fn (mut x Cfb) xor_key_stream(mut dst []u8, src []u8) {
 	unsafe {
-		mut dst := *dst_
-		mut src := src_
-		if dst.len < src.len {
+		mut local_dst := *dst
+		mut local_src := src
+		if local_dst.len < local_src.len {
```
spytheman, line 70: Those are good changes, just a bit unrelated to the topic of the PR.

vlib/crypto/rc4/rc4.v
```diff
@@ -65,7 +65,7 @@ pub fn (mut c Cipher) reset() {
 
 // xor_key_stream sets dst to the result of XORing src with the key stream.
 // Dst and src must overlap entirely or not at all.
-pub fn (mut c Cipher) xor_key_stream(mut dst []u8, mut src []u8) {
+pub fn (mut c Cipher) xor_key_stream(mut dst []u8, src []u8) {
```
spytheman, line 68: Good find.

spytheman (approved):
Excellent work.

PR #21983, opened 2024-08-02
tmpl: documentation

martinskou:
Documentation updated: Described how to use @ as a normal char inside a template.

### DIFF
--- a/vlib/v/TEMPLATES.md
+++ b/vlib/v/TEMPLATES.md
@@ -160,6 +160,5 @@ It's also possible to use properties of structs here like `@{my_struct.prop}`.
 
 # Escaping
 
-The `@` sign is used to begin a template directive. 
-To be able to use the `@` sign as a normal character inside a 
-template use `$$`.
+The `@` symbol starts a template directive. If you need to use `@` as a regular 
+character within a template, escape it by using a double `@` like this: `@@`.
--- a/vlib/v/TEMPLATES.md
+++ b/vlib/v/TEMPLATES.md
@@ -153,6 +153,10 @@ where you can insert your src
 @js 'myscripts.js'
 ```
 
+## Escaping
+
+The `@` sign is used to begin a template directive. To be able to use the `@` sign as a normal character inside a template use `$$`.
+
 # Variables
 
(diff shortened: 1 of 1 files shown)


vlib/v/TEMPLATES.md
```diff
@@ -157,3 +157,9 @@ where you can insert your src
 
 All variables, which are declared before the $tmpl can be used through the `@{my_var}` syntax.
 It's also possible to use properties of structs here like `@{my_struct.prop}`.
+
+# Escaping
+
+The `@` sign is used to begin a template directive. 
+To be able to use the `@` sign as a normal character inside a 
+template use `$$`.
```
spytheman, line 165: Hm, can you show a more complete example?

martinskou: Yes. Here are an example. If you place this in a template:

<link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css" crossorigin=""/>

it will render as:

<link rel="stylesheet" href="https://unpkg.com/leaflet$1.9.4/dist/leaflet.css" crossorigin=""/>

because of vlib/v/parser/tmpl.v line 80-82.

If you need to use @ as a normal @ char in a template you must use $$ instead. That is, if you place this in a template:

<link rel="stylesheet" href="https://unpkg.com/leaflet$$1.9.4/dist/leaflet.css" crossorigin=""/>

if renders as:

<link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css" crossorigin=""/>

JalonSolov: Did you try `@@` to get a single `@` in the output?

martinskou: I have now, @@ works as well. I have updated the description to indicate that @@ should be used.

spytheman: Thank you.

JalonSolov: It is strange to me that `$$` worked and turned into `@`... :man_shrugging:

medvednikov:
I feel like V should follow the "one way" rule and only have `$` in templates. No `@`

Issue #16502, opened 2022-11-21
A function's multiple return values cannot be arguments to another function


eptx:
### Describe the bug

A function's multiple return values cannot be arguments to another function. The work around requires 1) the return values to be stored in intermediate variables prior to passing to the next function or 2) to avoid multi-value returns in favor of Structs as V does not plan to support Tuples.

### Expected Behavior

A function's multiple return values can be arguments to another function. 

### Current Behavior
```
code.v:15:9: error: expected 2 arguments, but got 1
   13 | 
   14 | println(process2(1,'b'))
   15 | println(process2(get2()))
      |         ~~~~~~~~~~~~~~~~
   16 | println(process3(get2(),'a'))
code.v:16:9: error: expected 3 arguments, but got 2
   14 | println(process2(1,'b'))
   15 | println(process2(get2()))
   16 | println(process3(get2(),'a'))
      |         ~~~~~~~~~~~~~~~~~~~~
Exited with error status 1
```
### Reproduction Steps

```
fn get2() (int, string) {
    return 0, 'a'
}

fn process2(n int, s string) string {
    return '${n}:${s}'
}

fn process3(n int, s string, s2 string) string {
    return '${n}:${s}:${s2}'
}

println(process2(1,'b'))
println(process2(get2()))
println(process3(get2(),'a'))
```

### Additional Information/Context

Some discussion as to whether this is a bug or feature request. See https://github.com/vlang/v/discussions/16476

### V version

Current playground

### Environment details (OS name and version, etc.)

Current playground

PR #21991, opened 2024-08-03
v: allow multi return as fn argument

### LINKED ISSUE #21956: Can't pass multiple values returned by a function to another function
### Describe the bug

This works in Go:
```go
package main

func my_func(a1 int, a2 int) {}

func my_func2() (int, int) { return 0, 0 }

func main() {
	my_func(my_func2())
}
```

Doesn't work in V:

```v
fn my_func(a1 int, a2 int)  {}

fn my_func2() (int, int) { return 0,0}

fn main() {
	my_func(my_func2())
}

```

```bash
/tmp $ v run a.v
a.v:6:2: error: expected 2 arguments, but got 1
    4 |
    5 | fn main() {
    6 |     my_func(my_func2())
      |     ~~~~~~~~~~~~~~~~~~~
    7 | }
```

> [!NOTE]
>

felipensp:
Fix #21956

```v 
fn my_func(a1 int, a2 int)  {}

fn my_func2() (int, int)  { return 0,0 }

fn main() {
	my_func(my_func2())
}
```

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -777,7 +777,7 @@ pub mut:
 	receiver_type          Type // User / T, if receiver is generic, then cgen requires receiver_type to be T
 	receiver_concrete_type Type // if receiver_type is T, then receiver_concrete_type is concrete type, otherwise it is the same as receiver_type
 	return_type            Type
-	return_type_generic    Type   // the original generic return type from fn def
+	return_type_generic    Type // the original generic return type from fn def
 	nr_ret_values          int = -1 // amount of return values
 	fn_var_type            Type   // the fn type, when `is_fn_a_const` or `is_fn_var` is true
 	const_name             string // the fully qualified name of the const, i.e. `main.c`, given `const c = abc`, and callexpr: `c()`
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -778,6 +778,7 @@ pub mut:
 	receiver_concrete_type Type // if receiver_type is T, then receiver_concrete_type is concrete type, otherwise it is the same as receiver_type
 	return_type            Type
 	return_type_generic    Type   // the original generic return type from fn def
+	nr_ret_values          int    // amount of return values
 	fn_var_type            Type   // the fn type, when `is_fn_a_const` or `is_fn_var` is true
 	const_name             string // the fully qualified name of the const, i.e. `main.c`, given `const c = abc`, and callexpr: `c()`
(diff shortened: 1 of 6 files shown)


spytheman:
In my opinion, that makes the code that uses that new feature, harder and more confusing.

`my_func3(my_func2(), 'foo')` by the look of it, until now, meant that the `my_func3` has 2 arguments, and the first one is whatever the return value of `my_func2()` is (a single value).

Now, instead, it means that my_func3 accepts an arbitrary amount of arguments, the first of which are formed by the returned multiple values/tuple of `my_func2()`.

Given that we do not support tuples as first class values, it makes it confusing and inconsistent.

For example, the following code, that looks equivalent, will not work:
```v
val := my_func2()
my_func3( val, 'foo' )
```

medvednikov:
@spytheman you're right, and it shouldn't work like that. Only what was posted in the bug report.

In Go we get

```
./a.go:12:11: multiple-value my_func2() (value of type (int, float64)) in single-value context
./a.go:12:23: not enough arguments in call to my_func3
	have ((int, float64), string)
	want (int, int, string)
./a.go:13:18: multiple-value my_func2() (value of type (int, float64)) in single-value context
./a.go:13:18: not enough arguments in call to my_func4
	have (string, (int, float64))
	want (string, int, int)
```

felipensp:
Adjusted to just works on single-argument context.

spytheman:
If the goal for this feature, is to emulate Go's rule, it may need an extension for the case where more returned values are passed, but the outer function (that accepts them as arguments), have its last argument be a variadic one:

> As a special case, if the return values of a function or method g are equal in number and individually assignable to the parameters of another function or method f, then the call f(g(parameters_of_g)) will invoke f after binding the return values of g to the parameters of f in order. The call of f must contain no parameters other than the call of g, and g must have at least one return value. If f has a final ... parameter, it is assigned the return values of g that remain after assignment of regular parameters.

spytheman:
In my personal opinion, the whole thing is not very useful, and will encourage people to use function that return multiple values, more than they should.

In Go, that rule/special case, may have been more useful, because of their pattern of returning errors as a second element in the return tuple, so for them it could be a convenient way to write wrappers to handle the errors in a common way, like logging them for example.

In V, we do not have that, but instead use `!Type`, and require handling with `f() or { ... }` . I think that this special case rule, will be pretty much used only for translated Go code.

bakul:
I agree 100% with @spytheman and consider this a misfeature. Go's "feature" was forced on it due to it not having sum types. Something like go2v can use an idiom that doesn't require this feature.

eptx:
This also closed the bug I had opened: https://github.com/vlang/v/issues/16502

PR #21995, opened 2024-08-05
fmt: fix and simplify align of struct fields

yuyi98:
This PR fix and simplify align of struct fields.

### DIFF
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -32,7 +32,7 @@ pub mut:
 	single_line_if     bool
 	cur_mod            string
 	did_imports        bool
-	import_pos         int // position of the imports in the resulting string
+	import_pos         int               // position of the imports in the resulting string
 	auto_imports       map[string]bool   // potentially hidden imports(`sync` when using channels) and preludes(when embedding files)
 	used_imports       map[string]bool   // to remove unused imports
 	import_syms_used   map[string]bool   // to remove unused import symbols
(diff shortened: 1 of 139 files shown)


vlib/v/fmt/fmt.v
```diff
@@ -1409,31 +1410,62 @@ pub fn (mut f Fmt) interface_decl(node ast.InterfaceDecl) {
 	f.writeln('}\n')
 }
 
-pub fn (mut f Fmt) calculate_alignment(fields []ast.StructField, mut field_aligns []AlignInfo, mut comment_aligns []AlignInfo,
-	mut default_expr_aligns []AlignInfo, mut field_types []string) {
+pub fn (mut f Fmt) calculate_alignment(fields []ast.StructField, mut type_aligns []AlignInfo, mut comment_aligns []AlignInfo,
+	mut default_expr_aligns []AlignInfo, mut attr_aligns []AlignInfo, mut field_types []string) {
 	// Calculate the alignments first
+	mut prev_state := 0
```
spytheman, line 1416: What do the 0,1,2,3 mean?
I think, that an enum for those values can be easier to read/understand.

spytheman: Something like this for example?
```v
enum AlignState {
   default
   has_attributes
   has_default_expression
   has_everything
}
```

yuyi98: Thanks! I'll change it.

spytheman:
The timeout-ed windows gcc job cancelation is not related.

Good work.

PR #21998, opened 2024-08-06
ast: fix const field str()

yuyi98:
This PR fix const field str().
fmt before:
```v
module foo

pub const endline_cr_len = 1

pub struct RandomAccessReader {
pub mut:
	index i64

	start_index i64
	end_index   i64 = -1

	end_line      u8  = `\n`
	end_line_len  int = foo.endline_cr_len // size of the endline rune \n = 1, \r\n = 2
	separator     u8  = `,`            // comma is the default separator
	separator_len int = 1              // size of the separator rune
	quote         u8  = `"`            // double quote is the standard quote char
	quote_remove  bool // if true clear the cell from the quotes
	comment       u8 = `#` // every line that start with the quote char is ignored

	default_cell string = '*' // return this string if out of the csv boundaries
	empty_cell   string = '#' // retunrn this if empty cell
	// ram buffer
	mem_buf_type  u32     // buffer type 0=File,1=RAM
	mem_buf       voidptr // buffer used to load chars from file
	mem_buf_size  i64     // size of the buffer
	mem_buf_start i64 = -1 // start index in the file of the read buffer
	mem_buf_end   i64 = -1 // end index in the file of the read buffer
	// csv map for quick access
	csv_map [][]i64
	// header
	header_row  int = -1 // row index of the header in the csv_map
	header_list []HeaderItem   // list of the header item
	header_map  map[string]int // map from header label to column index
}
```
fmt now:
```v
module foo

pub const endline_cr_len = 1

pub struct RandomAccessReader {
pub mut:
	index i64

	start_index i64
	end_index   i64 = -1

	end_line      u8  = `\n`
	end_line_len  int = foo.endline_cr_len // size of the endline rune \n = 1, \r\n = 2
	separator     u8  = `,`                // comma is the default separator
	separator_len int = 1                  // size of the separator rune
	quote         u8  = `"`                // double quote is the standard quote char
	quote_remove  bool // if true clear the cell from the quotes
	comment       u8 = `#` // every line that start with the quote char is ignored

	default_cell string = '*' // return this string if out of the csv boundaries
	empty_cell   string = '#' // retunrn this if empty cell
	// ram buffer
	mem_buf_type  u32     // buffer type 0=File,1=RAM
	mem_buf       voidptr // buffer used to load chars from file
	mem_buf_size  i64     // size of the buffer
	mem_buf_start i64 = -1 // start index in the file of the read buffer
	mem_buf_end   i64 = -1 // end index in the file of the read buffer
	// csv map for quick access
	csv_map [][]i64
	// header
	header_row  int = -1 // row index of the header in the csv_map
	header_list []HeaderItem   // list of the header item
	header_map  map[string]int // map from header label to column index
}
```

### DIFF
--- a/vlib/encoding/csv/csv_reader_random_access.v
+++ b/vlib/encoding/csv/csv_reader_random_access.v
@@ -56,9 +56,9 @@ pub mut:
 
 	end_line      u8  = `\n`
 	end_line_len  int = csv.endline_cr_len // size of the endline rune \n = 1, \r\n = 2
-	separator     u8  = `,`            // comma is the default separator
-	separator_len int = 1              // size of the separator rune
-	quote         u8  = `"`            // double quote is the standard quote char
+	separator     u8  = `,`                // comma is the default separator
+	separator_len int = 1                  // size of the separator rune
+	quote         u8  = `"`                // double quote is the standard quote char
 	quote_remove  bool // if true clear the cell from the quotes
 	comment       u8 = `#` // every line that start with the quote char is ignored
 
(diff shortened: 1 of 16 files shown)


Delta456 (approved):

spytheman:
Is there an issue that this fixes?

Can it be added as a test case please, so that it does not regress in the future?

yuyi98:
Ok, I'll add test for it.

PR #22001, opened 2024-08-07
hash: Update hash.Hash methods

Coachonko:
The hash.Hash interface was likely partially ported from Go. 
Over time Vlang evolved and structs that implement the hash.Hash interface also gained new methods to bridge the differences between the Go version and the V version.
These new methods were not added to the hash.Hash interface.
This pull request adds these new methods to the hash.Hash interface.

### DIFF
--- a/vlib/hash/hash.v
+++ b/vlib/hash/hash.v
@@ -4,11 +4,15 @@
 module hash
 
 pub interface Hash {
+mut:
 	// Sum appends the current hash to b and returns the resulting array.
 	// It does not change the underlying hash state.
 	sum(b []u8) []u8
 	size() int
 	block_size() int
+	free()
+	reset()
+	write(p []u8) !int
 }
 
 interface Hash32er {


Coachonko:
In the future this interface may be improved a little bit:
I have not checked the source code of each implementation of `sum`, `size`, `block_size` method, I am not sure these are meant to have a mut receiver, however right now they are accepting a reference to the receiver, which should be about the same.

spytheman:
Why do you want to add the new methods? Do you have a specific example that would be enabled by them?
Keep in mind, that narrow interfaces are more widely useful than interfaces with lots of details.

Coachonko:
The structs that implement the hash.Hash interface already implement these methods, as you can see by the implementation of md5, sha1, sha256 and sha512.
In Go, these methods aren't needed because Go does not abstract these operations: The write method for example isn't needed there, but in V it is needed. This difference has surfaced after the port of the interface.

Consider the following code:
```v
	new_digest := fn [plugin_name] () hash.Hash {
		if plugin_name == 'Srp' {
			return sha1.new()
		}
		if plugin_name == 'Srp256' {
			return sha256.new()
		}
		panic('srp protocol error')
	}

	digest := new_digest()
	// TODO hash.Hash needs to have a write method.
	// hash.Hash is a port of Go's hash.Hash, but it differs significantly
	digest.write(n3.bytes())
	digest.write(n4.bytes())
	digest.write(salt)
	digest.write(key_a.bytes())
	digest.write(key_b.bytes())
	digest.write(key_k)
	key_m = digest.sum([]u8{})
```
In this situation, the function `new_digest` returns either a `sha1.Digest` or a `sha256.Digest`, these implement the `hash.Hash` interface. However, this Digest cannot use the `.write` method because the type `hash.Hash` does not have such method.

In Go, this is not an issue because `hash.Hash` embeds `io.Writer`
```
	if pluginName == "Srp" {
		digest = sha1.New()
	} else if pluginName == "Srp256" {
		digest = sha256.New()
	} else {
		panic("srp protocol error")
	}

	digest.Write(n3.Bytes())
	digest.Write(n4.Bytes())
	digest.Write(salt)
	digest.Write(keyA.Bytes())
	digest.Write(keyB.Bytes())
	digest.Write(keyK)
	keyM = digest.Sum(nil)
```

The change I propose bridges the difference and allows the `hash.Hash` interface to be used in a similar manner to Go's implementation

Coachonko:
Note for future improvements: 
For the `.write` method. we also have a `io.Writer` interface that could be embedded in `hash.Hash`.

PR #22003, opened 2024-08-07
time: `d`,`c`,`dd`,`ddd`,`dddd` pattern support for parse_format()

jeffangelion:
Allow weekday name to be parseable (doesn't affect Time struct)

### DIFF
--- a/vlib/time/time_test.v
+++ b/vlib/time/time_test.v
@@ -406,7 +406,7 @@ fn test_parse_ordinal_weekday_d() {
 }
 
 fn test_parse_ordinal_weekday_c() {
-	format := 'd MMM DD HH:mm:ss YYYY'
+	format := 'c MMM DD HH:mm:ss YYYY'
 	dt := '7 Jan 01 00:00:00 1970'
 	tm := time.parse_format(dt, format)!
 	tm_s := tm.custom_format(format)
--- a/vlib/time/time_test.v
+++ b/vlib/time/time_test.v
@@ -396,3 +396,43 @@ fn test_parse_three_letters_month() {
 	tm_tm := time.parse_format(tm_s, format)!
 	assert tm_tm.month == tm.month
 }
+
+fn test_parse_ordinal_weekday_sunday() {
+	format := 'd MMM DD HH:mm:ss YYYY'
+	dt := '0 Jan 01 00:00:00 1970'
+	tm := time.parse_format(dt, format)!
+	tm_s := tm.custom_format(format)
+	assert tm_s != dt
+}
+
+fn test_parse_ordinal_weekday_sunday() {
+	format := 'd MMM DD HH:mm:ss YYYY'
+	dt := '7 Jan 01 00:00:00 1970'
+	tm := time.parse_format(dt, format)!
+	tm_s := tm.custom_format(format)
+	assert tm_s != dt
+}
+
+fn test_parse_two_letters_weekday() {
+	format := 'dd MMM DD HH:mm:ss YYYY'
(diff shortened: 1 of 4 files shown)


vlib/time/time_test.v
```diff
@@ -396,3 +396,43 @@ fn test_parse_three_letters_month() {
 	tm_tm := time.parse_format(tm_s, format)!
 	assert tm_tm.month == tm.month
 }
+
+fn test_parse_ordinal_weekday_d() {
+	format := 'd MMM DD HH:mm:ss YYYY'
+	dt := '0 Jan 01 00:00:00 1970'
+	tm := time.parse_format(dt, format)!
+	tm_s := tm.custom_format(format)
+	assert tm_s != dt
```
spytheman, line 405: It is much better, to write a positive assert against the actual result, instead of one, that just states that tm_s should be != dt.

I.e. use `assert tm_s == '.....'`

jeffangelion: thanks for advice, I should make tests more readable

vlib/time/time_test.v
```diff
@@ -402,37 +402,37 @@ fn test_parse_ordinal_weekday_d() {
 	dt := '0 Jan 01 00:00:00 1970'
 	tm := time.parse_format(dt, format)!
 	tm_s := tm.custom_format(format)
-	assert tm_s != dt
+	assert tm_s == '4 Jan 01 00:00:00 1970'
 }
 
 fn test_parse_ordinal_weekday_c() {
 	format := 'd MMM DD HH:mm:ss YYYY'
```
spytheman, line 409: Given the fn name, should not that be using `c` too, instead of `d`?

jeffangelion: from time.custom_format() docs:
|Token|Output|
|------|--------|
|d|0..6 (Sun-Sat)|
|c|1..7 (Mon-Sun)|

spytheman: Ok, but `'d MMM DD HH:mm:ss YYYY'` is used 2 times, once on line 401, and another time here, on line 409. My question is whether the `d` here, should not be `c` instead?

jeffangelion: oh well you are right

also found what `d` and `c` tokens didn't work as documented

JalonSolov:
```v
long_days.index(weekday)
```
should be
```v
long_days[weekday]
```

jeffangelion:
@JalonSolov yeah you're right

already changed fn output from !int to !string (don't see much difference since fn result is always ignoring)

JalonSolov:
One possibility is that you declared the function as returning a Result, but there are no error returns.  If either of the `if` statements "fail" (as in the check doesn't let the code go into the body of the `if`), you could wind up with no answer at all, which should probably be an error...

spytheman:
> One possibility is that you declared the function as returning a Result, but there are no error returns. If either of the `if` statements "fail" (as in the check doesn't let the code go into the body of the `if`), you could wind up with no answer at all, which should probably be an error...

The function in question ends with:
 `return error_invalid_time(0, 'invalid weekday, at: ${p.current_pos_datetime}')`

If I comment that out, I get:
image

I think it works fine.

The only usage of that function is currently this one:
```v
            'dddd' {
                p.must_be_valid_week_day() or { return err }
            }
```

JalonSolov:
And of course, that can be shortened to
```v
            'dddd' {
                p.must_be_valid_week_day()!
            }
```

jeffangelion:
> And of course, that can be shortened to

IMHO writing functions for handling weekdays is overengineering since we're dropping them right away

PR #22007, opened 2024-08-08
fmt: fix enum fields with one empty line

yuyi98:
This PR fix enum fields with one empty line.

- Fix enum fields with one empty line.
- Add a test.

enum_fields_with_empty_line_keep.vv
```v
enum Info {
	aa = 1 // aa
	bbb // bbb

	cccc = 5 // cccc
	//
	ddddd = 10 // ddddd
}

fn main() {}
```

### DIFF
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -1072,9 +1072,9 @@ pub fn (mut f Fmt) enum_decl(node ast.EnumDecl) {
 		if i > 0 {
 			// keep one empty line between fields
 			last_field := node.fields[i - 1]
-			before_last_line := if last_field.comments.len > 0
-				&& last_field.pos.line_nr < last_field.comments.last().pos.last_line {
-				last_field.comments.last().pos.last_line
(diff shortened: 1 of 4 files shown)


vlib/v/fmt/fmt.v
```diff
@@ -1068,7 +1068,30 @@ pub fn (mut f Fmt) enum_decl(node ast.EnumDecl) {
...
+				&& field.pos.line_nr > field.comments[0].pos.line_nr {
+				field.comments[0].pos.line_nr
+			} else {
+				field.pos.line_nr
+			}
+
+			if next_first_line - before_last_line > 1 {
+				f.writeln('')
+			}
+		}
```
spytheman, line 1094: I think that we can modify the parser, so that it remembers for each field, whether, there was an empty line before it (I think that information, is easier to retrieve there, based on just the previous token line, (the one before the field name token), without caring about whether the previous field had comments, or expressions etc).
Then you can write a new line here, based on that info, without having to do all this.

yuyi98: Ok，I see.

PR #22009, opened 2024-08-08
hash: replace `.write` method with `io.Writer` embedding

Coachonko:
Minor change as suggested in #22001


spytheman:
It is a single method. We do not need to `import io` just for that.
To borrow one of Go's aphorisms, "a little copying is better than a little dependency".

Coachonko:
Sounds good to me!

PR #22010, opened 2024-08-08
cgen: skip .placeholder when writing result generic types

### LINKED ISSUE #22008: error C too
```V
module parselona

pub struct ParserResult[O] {
	i []u8
	o O
}

type TParser[O] = fn([]u8) !ParserResult[O]
type TF[O] = fn([]u8) O

fn take_part(count u32) TParser[[]u8] {
	return fn [count] (i []u8) !ParserResult[[]u8] {
		if i.len<count { return error('error len: ${i}') }
    	        return ParserResult[[]u8] {
    		    i: i[count..],
    		    o: i[..count],
    	       }
	}
}

fn map<O>(p TParser[u8], f TF[O]) TParser[O] {
	return fn [p, f] <O>(i []u8) !ParserResult[O] {
    	     r:=p(i)!
    	     return ParserResult{i:r.i, o:f(r.o)}
	}
} 


fn test_1() {
input := '10_ttt_uuuu_qqq_'.bytes()

  nnn:=take_part(2)  
  println(' it is: ${nnn(input)!}')
  fnf := fn(i[]u8) int { return 1 }
  n := map[int](nnn, fnf)

assert ['nn'] == ['ttt_', 'uuuu_', 'qqq_'] 
}

`
--- Testing... -----------------------------------------------------------------------------------------------------------------------------------------------------------
 FAIL     0.000 ms /home/prof/projects/v/parselona/src/parselona_test.v
>> compilation failed:
cc: /tmp/v_1000/tsession_704f1d69f740_01J4Q5YZCTYXCVAK9ESWAGAX7W/parselona_test.01J4Q5Z0T9ZS2Q9JCXM6E0YW4D.tmp.c:106: error: identifier expected
builder error: 
==================


`

_Originally posted by @Cergoo in https://github.com/vlang/v/issues/21989#issuecomment-2274227938_

### PR DESCRIPTION
**Pull request by felipensp**
Fix #22008

### DIFF
--- a/vlib/v/tests/result_generic_test.v
+++ b/vlib/v/tests/result_generic_test.v
@@ -19,16 +19,6 @@ fn take_part(count u32) TParser[[]u8] {
 	}
 }
 
-fn map[O](p TParser[u8], f TF[O]) TParser[O] {
-	return fn [p, f] [O](i []u8) !ParserResult[O] {
-		r := p(i)!
-		return ParserResult{
-			i: r.i
-			o: f(r.o)
-		}
-	}
-}
-
 fn test_main() {
 	input := '10_ttt_uuuu_qqq_'.bytes()
 
--- a/vlib/v/tests/result_generic_test.v
+++ b/vlib/v/tests/result_generic_test.v
@@ -1,0 +1,39 @@
+pub struct ParserResult[O] {
+	i []u8
+	o O
+}
+
+type TParser[O] = fn ([]u8) !ParserResult[O]
+
+type TF[O] = fn ([]u8) O
+
+fn take_part(count u32) TParser[[]u8] {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: vlib/v/tests/result_generic_test.v [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+			return error('error len: ${i}')
+		}
+		return ParserResult[[]u8]{
+			i: i[count..]
+			o: i[..count]
+		}
+	}
+}
+
+fn map[O](p TParser[u8], f TF[O]) TParser[O] {
```
spytheman, line 22: imho this should be caught earlier. Will you be able to call that function, given that `map[..]` is the start of a type declaration?

felipensp: Yes, probably this should be a error.

PR #22012, opened 2024-08-08
net.mbedtls: Server Name Indication (SNI)

martinskou:
net.mbedtls: Makes it possible to use Server Name Indication (SNI). The mbedtls.SSLConnectConfig is extended with a callback get_certificate which on connection receives a hostname and can return the certificate which the connection should use.

Server Name Indication (SNI) is an extension which is designes to enable usage of multiple domains be the same server on the same port. This extension is used be fx Nginx and Caddy.

Server Name Indication (SNI) is defined in RFC 6066. See https://mbed-tls.readthedocs.io/en/latest/kb/how-to/use-sni/

Example:
```
import io
import os
import net.mbedtls

fn get_cert(host string) !mbedtls.SNICerts {
	// callback to return the certificates to use for the connection, based on the host
	sni_certs := mbedtls.SNICerts{
		cert: os.read_file('cert/${host}.crt') or { return error('Failed to read certificate') }
		key: os.read_file('cert/${host}.key') or { return error('Failed to read key') }
	}
	return sni_certs
}

fn main() {
	mut server := mbedtls.new_ssl_listener('0.0.0.0:8443', mbedtls.SSLConnectConfig{
		verify: os.resource_abs_path('cert/ca.crt')
		cert: os.resource_abs_path('cert/defaultserver.crt')
		cert_key: os.resource_abs_path('cert/defaultserver.key')
		validate: false
		get_certificate: get_cert // set the callback to enable SNI
	})!

	mut client := server.accept()!	
}
```


spytheman:
The example `examples/ssl_server/server_sni.v`, just hangs for me.
Should not the client try to at least send the server a host header?

martinskou:
> The example `examples/ssl_server/server_sni.v`, just hangs for me. Should not the client try to at least send the server a host header?

Yes it should. But be aware that the browser does not send any host name if a domain is not used. So 127.0.0.1 will not work.

It can be tested using fx:
```
curl -vik --resolve 1x.dk:8443:0.0.0.0 https://1x.dk:8443/
```

or

```
openssl s_client -connect 0.0.0.0:8443 -servername 1x.dk -debug
```

vlib/net/mbedtls/ssl_connection.c.v
```diff
@@ -174,6 +174,43 @@ fn (mut l SSLListener) init() ! {
...
+fn (mut l SSLListener) init_sni(get_cert_callback fn (string) !SNICerts) {
+	$if trace_ssl ? {
+		eprintln(@METHOD)
+	}
+	C.mbedtls_ssl_conf_sni(&l.conf, fn [get_cert_callback, l] (param voidptr, ssl &C.mbedtls_ssl_context, name &char, lng int) int {
+		host := unsafe { name.vstring_literal_with_len(lng) }
+		if c := get_cert_callback(host) {
+			// got a certificate from the callback
+			C.mbedtls_x509_crt_init(&l.certs.client_cert)
+			C.mbedtls_pk_init(&l.certs.client_key)
```
spytheman, line 195: What happens, if that is executed on multiple threads?
Is accessing `l.certs.client_cert` thread safe?

martinskou: You are rigth, it probertly should not you the l.certs but its own instance, i'll try to fix.

spytheman: Thank you 🙇🏻‍♂️ .

martinskou:
I have updated the code.
See examples/ssl_server/server_sni_advanced.v for an example using multiple certificates.
The original (default) certificates in the mbedtls.SSLListener is no longer changed (the get_certificate callback must supply a new mbedtls.SSLCerts and this eliminates the issue about thread safety.

spytheman:
These new public APIs will need doc comments too, so that the CI can pass:
 ```
vlib/net/mbedtls/ssl_connection.c.v:94:0:pub fn (mut c SSLCerts) cleanup() 
vlib/net/mbedtls/ssl_connection.c.v:34:0:pub fn new_sslcerts() &SSLCerts 
vlib/net/mbedtls/ssl_connection.c.v:68:0:pub fn new_sslcerts_from_file(verify string, cert string, cert_key string) !&SSLCerts 
vlib/net/mbedtls/ssl_connection.c.v:42:0:pub fn new_sslcerts_in_memory(verify string, cert string, cert_key string) !&SSLCerts 
```

spytheman:
I've added some comments in 894350f . They may need to be edited later, but should be enough for the CI for now.

spytheman (approved):
Excellent work.

PR #22018, opened 2024-08-10
fmt: cleanup fields alignment

yuyi98:
This PR cleanup fields alignment.

### DIFF
--- a/vlib/v/fmt/align.v
+++ b/vlib/v/fmt/align.v
@@ -16,38 +16,55 @@ pub:
 	threshold     int = 25
 }
 
-fn (mut infos []AlignInfo) add_new_info(len int, line int) {
-	infos << AlignInfo{
+struct FieldAlign {
+mut:
+	infos   []AlignInfo
+	cur_idx int
+}
+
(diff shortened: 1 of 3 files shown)


spytheman:
Weird, I do not understand the vsl/vtl errors on the CI 🤔 .

spytheman:
```
/home/runner/.vmodules/vsl/errors/errors.v:8:1: notice: `[attr]` has been deprecated, use `@[attr]` instead
    6 | }
    7 | 
    8 | [inline]
      | ^
```

It behaves, as if it is using an older `vsl` version for some reason 🤔 ... The most recent version on `main`, does not have that `[inline]` there, but `@[inline]` already.

spytheman:
It is also not specific to this PR, but affects https://github.com/vlang/v/pull/22019 too 🤔 .

spytheman:
I've pushed a new formatted version of the main branch of vlang/vsl . Hopefully that will fix it, by invalidating a cache somewhere 🤔 .

spytheman (approved):
Good work.

PR #22024, opened 2024-08-11
flag: add optional value description to string parameters

kimshrier:
When using command line flags with string parameters, I found that the default val_desc values used were not as descriptive as I would like.  And there was no way to override val_desc to a more descriptive value.

Using trailing struct literal arguments, I was able to add this capability without breaking any existing code.

I only did this for string parameters since that seemed to me to be the place where it would be most useful.  It could easily be added to the other types if desired.

### DIFF
--- a/vlib/flag/flag.v
+++ b/vlib/flag/flag.v
@@ -400,19 +400,39 @@ pub fn (mut fs FlagParser) float(name string, abbr u8, fdefault f64, usage strin
 	return value
 }
 
+@[params]
+pub struct FlagConfig {
+pub:
+	val_desc string // descriptive string for an argument
+}
+
 // string_multi returns all string values, associated with the flag named `name`.
 // When no values for that flag are found, it returns an empty array.
 // This version supports abbreviations.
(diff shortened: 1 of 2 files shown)


vlib/flag/flag.v
```diff
@@ -400,19 +400,39 @@ pub fn (mut fs FlagParser) float(name string, abbr u8, fdefault f64, usage strin
 	return value
 }
 
+@[params]
+pub struct FlagConfig {
+pub:
+	val_desc string // descriptive string for an argument
```
spytheman, line 406:
```suggestion
	description string
```

spytheman: `description` is longer, but does not need an additional comment, and would be clearer and more mnemonic in the usage sites too.

spytheman:
> I only did this for string parameters since that seemed to me to be the place where it would be most useful. It could easily be added to the other types if desired.

I think it would be useful for the others too.

vlib/flag/flag.v
spytheman: While reading this more carefully, I think that I was wrong the first time.
`description` would be misleading, and `val_desc` is good, given the functionality.

PR #22032, opened 2024-08-11
flag: add custom value descriptions for bool, int, and float flags too

kimshrier:
Added the same sort of value description to bool, int, and float flags as was added for string flags.  I also tweaked the code for string value descriptions.


Delta456 (approved):
LGTM

PR #22034, opened 2024-08-11
cmd/tools: add amalgamate tool and description of usage

kimshrier:
I wrote a V tool program to do the source amalgamtion for libgc, replacing the need to use the amalgamator.js script.

The existing amalgamator.js script was specific to doing the amalgamation for libgc.  I made this tool more generic so that it could handle libgc and anything else that one might want to amalgamate.  I also updated the description of how to generate gc.c and put it, along with updated header files, into the right place in the thirdparty/libgc directory.

This amalgamate tool is the same one I used for my previous update to thirdparty/libgc/gc.c.  I have a newer version of gc.c based on a newer commit of bdwgc, but I didn't want to include it in this pull request.  Hopefully, with this tool, it will be easier to keep libgc more current.

### DIFF
--- a/cmd/tools/amalgamate.v
+++ b/cmd/tools/amalgamate.v
@@ -17,6 +17,9 @@ import regex
 const app_name = 'amalgamate'
 const app_version = '0.0.1'
 
+// pre-compile the include statement regex
+const re = regex.regex_opt(r'^\s*#\s*include\s*"([^"]+)"')!
+
 struct Config {
 mut:
 	input_files []string
@@ -169,9 +172,6 @@ fn (c Context) find_file(file string) !string {
 fn (mut c Context) handle_includes(local_dir string, input_source string) !string {
 	source_lines := input_source.split_into_lines()
 	mut output_lines := []string{}
-
-	include := r'^\s*#\s*include\s*"([^"]+)"'
-	mut re := regex.regex_opt(include) or { panic(err) }
 
 	for line in source_lines {
 		start, _ := re.match_string(line)
--- a/cmd/tools/amalgamate.v
+++ b/cmd/tools/amalgamate.v
@@ -1,0 +1,221 @@
+// amalgamate multiple C source files into a single
+// C source file.  See https://sqlite.org/amalgamation.html
+// for a description of file amalgamation.
+//
+// If an input file is not specified, source is read
+// from stdin.
+//
+// If an output file is not specified, source is output
+// to stdout.
+
+module main
+
+import flag
+import os
+import regex
+
+const app_name = 'amalgamate'
+const app_version = '0.0.1'
+
+struct Config {
+mut:
+	input_files []string
+	output_file string
+	search_dirs []string
+	blacklist   []string
+}
+
+struct Context {
+	config Config
+mut:
+	processed_files []string
+}
+
+fn parse_arguments() Config {
+	mut cfg := Config{}
+
+	mut parser := flag.new_flag_parser(os.args)
+	parser.skip_executable()
+	parser.application(app_name)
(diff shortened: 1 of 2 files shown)


spytheman:
Thank you 🙇🏻‍♂️!

spytheman (approved):

kimshrier:
Given the complications around updating gc.c and libgc.a, I thought I would go ahead and push up this tool and then see about updating gc.c and libgc.a.  

Based on our earlier discussion, I don't know if I should be looking into automating that process or not.  I don't have much experience with configuring CI stuff at github and I don't know if I would be more of a hindrance than a help.

spytheman:
> I thought I would go ahead and push up this tool

Excellent.

> Based on our earlier discussion, I don't know if I should be looking into automating that process or not. I don't have much experience with configuring CI stuff at github and I don't know if I would be more of a hindrance than a help.

Do not worry. I will take care of it.

kimshrier:
Thanks.  Hopefully, I haven't added to your work load too much. :)

cmd/tools/amalgamate.v
```diff
@@ -0,0 +1,221 @@
...
+		'the current working directory and\n' + flag.space +
+		'if not found, the directories in this\n' + flag.space +
+		'list are searched, in order, until the\n' + flag.space +
+		'file is found or the search list is\n' + flag.space +
+		'exhausted.  This can be specified\n' + flag.space + 'more that once.\n',
+		val_desc: '<search_dir>'
+	)
+
+	cfg.input_files = parser.finalize() or {
+		// this only reports the first unrecognized argument
```
spytheman, line 81: would you like it to report all instead?

kimshrier: That would be nice.  I guess I missed a setting.

cmd/tools/amalgamate.v
```diff
@@ -0,0 +1,221 @@
...
+// and inserts the lines from the named include file.
+//
+// The pattern matches file names for local header files,
+// not system header files as are denoted by < and >.
+fn (mut c Context) handle_includes(local_dir string, input_source string) !string {
+	source_lines := input_source.split_into_lines()
+	mut output_lines := []string{}
+
+	include := r'^\s*#\s*include\s*"([^"]+)"'
+	mut re := regex.regex_opt(include) or { panic(err) }
```
spytheman, line 174: It would have been nice, if that regex compilation could be extracted in a constant 🤔 ... However, it probably does not matter much, because it would not be run very frequently.

kimshrier: I have no problem with changing this.

kimshrier:
I'll make a few changes and push up a new version.

cmd/tools/amalgamate.v
kimshrier: I might be blind.  I don't see how the finalize function would return more than 1 UnknownFlagError.  Am I misinterpreting the following code?
```
    if !fs.allow_unknown_args {
        for a in remaining {
            if (a.len >= 2 && a[..2] == '--') || (a.len == 2 && a[0] == `-`) {
                return &UnknownFlagError{
                    flag: a
                }
            }
        }
    }
```

spytheman: No, sorry. I was not clear :-| - I was asking, if that is something that would be desirable, so that it can be added later.

kimshrier: I usually like to return as many errors as possible or practical from a single run so that they don't have to be discovered one-at-a-time.  However, I am also mindful of not overwhelming the user with too many error messages.  It's a balancing act.

PR #22039, opened 2024-08-12
cgen: make possible to initialize sumtype with default type (first variant)

### LINKED ISSUE #22030: sum types should have a default 0 value for the first type mentioned in the list of types.
### Describe the bug

sum types do not have a default value.  This can lead to errors if you try something as simple as
```v
x := MyType{}
```
this could be easily fixed if the default was to assign the default 0 value for the first type in the list.  Users could change which type was assigned by changing the order of types in the sum type.

Note that this change has been approved by Alex:  https://discord.com/channels/592103645835821068/592106336838352923/1272226313314242673

### Reproduction Steps

```v
type MyType = int | string

x := MyType{}

println(x)
```

### Expected Behavior

```
0
```

### Current Behavior

```
x.v:3:6: notice: direct sum type init (`x := SumType{}`) will be removed soon
    1 | type MyType = int | string
    2 | 
    3 | x := MyType{}
      |      ~~~~~~~~
    4 | 
    5 | println(x)
unknown sum type value
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.7 20a525a

```
V full version: V 0.4.7 efa98d9.20a525a
OS: linux, "Manjaro Linux"
Processor: 32 cpus, 64bit, little endian, AMD Ryzen 9 7950X 16-Core Processor

getwd: /home/jalon
vexe: /home/jalon/git/v/v
vexe mtime: 2024-08-11 16:45:16

vroot: OK, value: /home/jalon/git/v
VMODULES: OK, value: /home/jalon/.vmodules
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.32-24-g20a525ae
.git/config present: true

CC version: cc (GCC) 14.1.1 20240720
```
> [!NOTE]
>

felipensp:
Fix #22030

### DIFF
--- a/vlib/x/json2/tests/encode_struct_test.v
+++ b/vlib/x/json2/tests/encode_struct_test.v
@@ -252,7 +252,7 @@ fn test_pointer() {
 }
 
 fn test_sumtypes() {
-	assert json.encode(StructType[SumTypes]{}) == '{}' // is_none := val.$(field.name).str() == 'unknown sum type value'
+	assert json.encode(StructType[SumTypes]{}) == '{"val":{"val":""}}' // is_none := val.$(field.name).str() == 'unknown sum type value'
 	assert json.encode(StructType[SumTypes]{ val: '' }) == '{"val":""}'
 	assert json.encode(StructType[SumTypes]{ val: 'a' }) == '{"val":"a"}'
 
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -18,7 +18,8 @@ pub const int_type_name = $if new_int ? {
 	'int'
 }
 
-pub type Expr = AnonFn
+pub type Expr = NodeError
+	| AnonFn
 	| ArrayDecompose
 	| ArrayInit
 	| AsCast
@@ -53,7 +54,6 @@ pub type Expr = AnonFn
 	| MapInit
 	| MatchExpr
 	| Nil
-	| NodeError
 	| None
 	| OffsetOf
 	| OrExpr
--- a/vlib/v/ast/types.v
+++ b/vlib/v/ast/types.v
@@ -17,7 +17,10 @@ import v.token
 
 pub type Type = int
 
-pub type TypeInfo = Aggregate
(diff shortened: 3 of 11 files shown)


spytheman:
You will have to change vfmt too, so that it would not sort the variants anymore.

StunxFS:
And what happens if the first variant of the sumtype is a struct that has some field that is of pointer type?  (the compiler should force explicit initialization)

vlib/x/json2/tests/encode_struct_test.v
```diff
@@ -252,7 +252,7 @@ fn test_pointer() {
 }
 
 fn test_sumtypes() {
-	assert json.encode(StructType[SumTypes]{}) == '{}' // is_none := val.$(field.name).str() == 'unknown sum type value'
+	assert json.encode(StructType[SumTypes]{}) == '{"val":{"val":""}}' // is_none := val.$(field.name).str() == 'unknown sum type value'
```
spytheman, line 255: The comment at the end, may be now obsolete imho, since the sumtype value is known.

spytheman (approved):
Excellent work.

spytheman:
@felipensp about @StunxFS's comment, consider this:
```v
struct Node {
    parent &Node
    left &Node
    right &Node
}
type Sumtype = Node | int

// n := Node{} /* will produce a compiler error, because the reference fields will be nil pointers */

s := Sumtype{} // currently, the compiler does not care, even though the node that will be created by default will be the same as `n` above
dump(s)
```

producing (without any compiler messages):
```
#0 02:36:52 ^ fix_sumtype_init /v/oo>v run ss.v
[ss.v:11] s: Sumtype(Node{
    parent: &nil
    left: &nil
    right: &nil
})
```

spytheman:
i.e. the compiler currently acts on `s := Sumtype{}`, as if I had written:
```v
struct Node {
    parent &Node = unsafe { nil }
    left &Node = unsafe { nil }
    right &Node = unsafe { nil }
}
```

Issue #22044, opened 2024-08-14
C error when comparing array to nil


ratkingsminion:
### Describe the bug

Code: https://play.vlang.io/p/8cd927fdbb

```v
struct Test {
	a []int
}
t := Test {}
unsafe {
	if t.a != nil {
		dump(t.a)
	}
}
```

### Reproduction Steps

See code

### Expected Behavior

A better error message

### Current Behavior

Output:

```
/tmp/v_60000/../../../../../../box/code.v:6: error: cannot convert 'struct array' to 'unsigned long'
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
Exited with error status 1
```

### V version

V 0.4.7 21339fe

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.7 21339fe
OS: linux, Debian GNU/Linux 11 (bullseye) (VM)
Processor: 1 cpus, 64bit, little endian, Intel(R) Xeon(R) CPU E5-2686 v4 @ 2.30GHz

getwd: /home/admin/playground
vexe: /home/admin/v/v
vexe mtime: 2024-08-14 09:25:24

vroot: OK, value: /home/admin/v
VMODULES: OK, value: .vmodules
VTMP: OK, value: /tmp/v_0

Git version: git version 2.30.2
Git vroot status: Error: fatal: detected dubious ownership in repository at '/home/admin/v'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v
.git/config present: true

CC version: cc (Debian 10.2.1-6) 10.2.1 20210110
thirdparty/tcc status: Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v/thirdparty/tcc
 Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v/thirdparty/tcc

```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #22045, opened 2024-08-14
checker: add infix checks for nil 

Delta456:
Fixes https://github.com/vlang/v/issues/22044


vlib/v/checker/infix.v
```diff
@@ -197,6 +197,36 @@ fn (mut c Checker) infix_expr(mut node ast.InfixExpr) ast.Type {
 					}
 				}
 			}
+
```
medvednikov, line 200: Thanks. Can you a simple 
`// Do not allow comparing nil to non-pointers`
here
would make it much more readable

PR #22046, opened 2024-08-15
fmt: keep empty newlines in between interface fields/methods

yuyi98:
This PR keep empty newlines in between interface fields/methods.

- Keep empty newlines in between interface fields/methods.
- Add test.

```v
interface Foo {
	// a
	a string

	bb  int
	ccc bool

	// m1, m2
	m1() string
	m2() int

	// m3.1
	// m3.2
	m3() bool
}

fn main() {
}
```


vlib/v/parser/parser.v
```diff
@@ -501,6 +501,26 @@ fn (p &Parser) is_fn_type_decl() bool {
...
+	for {
+		if tok.line_nr - prev_tok.line_nr - prev_tok.lit.count('\n') > 1 {
+			return true
+		}
+		if prev_tok.kind == .comment {
+			idx--
+			tok = prev_tok
+			prev_tok = p.peek_token(idx)
+			continue
+		}
```
spytheman, line 518: That seems weird ... Why there should it care about the comments here?

yuyi98:
```v
        ccc bool

	// m1, m2
	m1() string
	m2() int
```
this case cannot be handled properly.

spytheman (approved):

PR #22047, opened 2024-08-15
crypto: add a crypto.pbkdf2 module

sibkod:
Added pbkdf2


Delta456:
A reference to the original implementation and the need to explain why it has to be in `vlib` will also be helpful.

vlib/crypto/pbkdf2/pbkdf2.v
```diff
@@ -0,0 +1,55 @@
...
+		else {
+			panic('Unsupported hash')
+		}
+	}
+
+	hash_length := size
+	block_count := (key_length + hash_length - 1) / hash_length
+	mut output := []u8{}
+	mut last := []u8{}
+	mut buf := []u8{len: 4, init: 0}
```
spytheman, line 34:
```suggestion
	mut buf := []u8{len: 4}
```

vlib/crypto/pbkdf2/pbkdf2.v
```diff
@@ -0,0 +1,55 @@
...
+		mut last_hash := xorsum.clone()
+		for j := 1; j < count; j++ {
+			last_hash = hmac.new(password, last_hash, fun, block_size)
+			for k in 0 .. xorsum.len {
+				xorsum[k] ^= last_hash[k]
+			}
+		}
+		output << xorsum
+	}
+	return output[..key_length]
```
spytheman, line 54: What happens, if `key_length` is very big?
Is there a guarantee, that this will not panic?

vlib/crypto/pbkdf2/pbkdf2_test.v
```diff
@@ -0,0 +1,113 @@
+module pbkdf2
+
+import crypto.sha512
+import crypto.sha256
+
+const data = 'test'
+const password = '123456'
+
+fn test_sha512() {
+	assert key(pbkdf2.data.bytes(), pbkdf2.password.bytes(), 1000, 64, sha512.new())! == [
```
spytheman, line 10: Public functions are better tested in normal tests, since they better reflect how users will be using them.

vlib/crypto/pbkdf2/pbkdf2_test.v
```diff
@@ -0,0 +1,113 @@
...
+		u8(149),
```
spytheman, line 11: This assertion, can benefit from `// vfmt off` before it, and `// vfmt on` after it;
then the bytes could be all on the same line, without wrapping.

vlib/crypto/pbkdf2/pbkdf2_test.v
```diff
@@ -0,0 +1,113 @@
...
+		205,
+		187,
+		247,
+		160,
+	]
+}
+
+fn test_sha256() {
+	assert key(pbkdf2.data.bytes(), pbkdf2.password.bytes(), 1000, 32, sha256.new())! == [
+		u8(110),
```
spytheman, line 80: Same - using `// vfmt off` and `// vfmt on` here, can cleanup the wrapping of the long array.

vlib/crypto/pbkdf2/pbkdf2.v
```diff
@@ -0,0 +1,55 @@
+module pbkdf2
+
+import crypto.hmac
+import crypto.sha256
+import crypto.sha512
+import hash
+
+pub fn key(password []u8, salt []u8, count int, key_length int, h hash.Hash) ![]u8 {
```
spytheman, line 8: It would need a doc comment (starting with `// key`), to pass the CI (https://github.com/vlang/v/actions/runs/10400669990/job/28801659491?pr=22047).

JalonSolov:
Simplest would be in a comment block at the top of the `pbkdf2.v` file.

vlib/crypto/pbkdf2/pbkdf2.v
spytheman: Taking a long time is fine, the question is can it panic?

sibkod: I don't think so

spytheman (approved):

PR #22050, opened 2024-08-15
x.json2.decode2: add another internal test

enghitalo:
Some json2 test

Next PR: include test bellow

```v
	nodes = []

	decoder = Decoder{
		json: '{"_type": "Cat", "val": 2}'
	}
	decoder.fulfill_nodes(mut nodes)

	dump(nodes)

	assert nodes.len == 1
	assert nodes[0].key_pos == 18
	assert nodes[0].key_len == 3
	assert nodes[0].children == none
```

### DIFF
--- a/vlib/x/json2/decoder2/decode_test.v
+++ b/vlib/x/json2/decoder2/decode_test.v
@@ -2,7 +2,6 @@ module decoder2
 
 fn test_nodes() {
 	mut nodes := []Node{}
-	empty_nodes_workaround := []Node{}
 
 	mut decoder := Decoder{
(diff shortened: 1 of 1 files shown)


vlib/x/json2/decoder2/decode_test.v
```diff
@@ -0,0 +1,43 @@
+module decoder2
+
+fn test_nodes() {
+	mut nodes := []Node{}
+	empty_nodes_workaround := []Node{}
```
spytheman, line 5: `empty_nodes_workaround` is not used at all later, remove it.

PR #22051, opened 2024-08-15
fmt: fix enum/struct_decl/struct_init fields with empty newlines

yuyi98:
This PR fix enum/struct_decl/struct_init fields with empty newlines.

- Fix enum/struct_decl/struct_init fields with empty newlines.
- Add tests.

```v
module abcde

pub struct Builder {
pub mut:
	// inline before field
	buf          []u8
	str_calls    int
	len          int
	initial_size int = 1
}

pub fn new_builder(initial_size int) Builder {
	return Builder{
		// buf: make(0, initial_size)
		buf: []u8{cap: initial_size}

		// before comment
		str_calls: 0 // after str_calls
		len:       0 // after len

		// before comment1
		// before comment2
		initial_size: initial_size // final
	}
}
```
```v
module main

fn main() {
}

pub struct OperateInfo { // implements IperateInfo
pub mut:
	title string // title

	// msg, id
	msg string // msg
	id  string // id

	// other1
	// other2
	other string
} // operate info
```
```v
enum Info {
	aa = 1 // aa
	bbb // bbb

	cccc = 5 /* cccc
	--- cccc
	*/
	ddddd = 10 // ddddd

	// before
	ee  = 20
	fff = 30

	// before comment1
	// before comment2
	gg
}

fn main() {}
```


JalonSolov:
Hmm... I'd disagree with
```v
	aa = 1 // aa
	bbb // bbb
```
I think it should be
```v
	aa = 1 // aa
	bbb    // bbb
```
Even though there is no initializer on `bbb`, it is still next to `aa` with no blank lines or comments between them.

PR #22055, opened 2024-08-16
fmt: fix interface fields/methods alignment

yuyi98:
This PR fix interface fields/methods alignment.

- Fix interface fields/methods alignment.
- Modify the related test.

```v
interface TestRunner {
mut:
	file_test_info      VTestFileMetaInfo // filled in by generated code, before .start() is called.
	fn_test_info        VTestFnMetaInfo   // filled in by generated code, before .fn_start() is called.
	fn_assert_passes    u64               // reset this to 0 in .fn_start(), increase it in .assert_pass()
	fn_passes           u64               // increase this in .fn_pass()
	fn_fails            u64               // increase this in .fn_fails()
	total_assert_passes u64               // increase this in .assert_pass()
	total_assert_fails  u64               // increase this in .assert_fail()

	start(ntests int) // called before all tests, you can initialise private data here. ntests is the number of test functions in the _test.v file.
	finish()          // called after all tests are finished, you can print some stats if you want here.
	exit_code() int   // called right after finish(), it should return the exit code, that the test program will exit with.

	fn_start() bool // called before the start of each test_ function. Return false, if the function should be skipped.
	fn_pass()       // called after the end of each test_ function, with NO failed assertion.
	fn_fail()       // called after the end of each test_ function, with a failed assertion, *or* returning an error.
	fn_error(line_nr int, file string, mod string, fn_name string, errmsg string) // called only for `fn test_xyz() ? { return error('message') }`, before .fn_fail() is called.

	assert_pass(i &VAssertMetaInfo) // called after each `assert true`.
	assert_fail(i &VAssertMetaInfo) // called after each `assert false`.

	free() // you should free all the private data of your runner here.
}

struct VTestFileMetaInfo {
	file  string
	tests int
}
```

### DIFF
--- a/cmd/tools/modules/testing/output.v
+++ b/cmd/tools/modules/testing/output.v
@@ -30,13 +30,13 @@ pub:
 
 pub interface Reporter {
 mut:
-	session_start(message string, mut ts TestSession) // called once per test session, in the main thread, suitable for setting up supporting infrastructure.
-	session_stop(message string, mut ts TestSession) // called once per test session, in the main thread, after everything else, suitable for summaries, creating .xml reports, uploads etc.
+	session_start(message string, mut ts TestSession)        // called once per test session, in the main thread, suitable for setting up supporting infrastructure.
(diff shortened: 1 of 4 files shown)


spytheman (approved):
Excellent work.

PR #22065, opened 2024-08-17
builtin,cgen: remove allocations for initialising as_cast_type_indexes and the none__ constant

spytheman:


spytheman:
Needs more work 🤔 .

PR #22067, opened 2024-08-17
checker: add error when initializing sumtype with struct as first type

felipensp:
This PR add a checker error when initializing struct via sumtype as first variant type.

```v
struct Node {
    parent &Node
    left &Node
    right &Node
}
type Sumtype = Node | int

fn main() {
    s := Sumtype{}
    dump(s)
}
```

```
bug.v:9:10: error: reference field `Node.parent` must be initialized
    7 | 
    8 | fn main() {
    9 |     s := Sumtype{}
      |          ~~~~~~~~~
   10 |     dump(s)
   11 | }
bug.v:9:10: error: reference field `Node.left` must be initialized
    7 | 
    8 | fn main() {
    9 |     s := Sumtype{}
      |          ~~~~~~~~~
   10 |     dump(s)
   11 | }
bug.v:9:10: error: reference field `Node.right` must be initialized
    7 | 
    8 | fn main() {
    9 |     s := Sumtype{}
      |          ~~~~~~~~~
   10 |     dump(s)
   11 | }
```

### DIFF
--- a/vlib/v/checker/struct.v
+++ b/vlib/v/checker/struct.v
@@ -777,15 +777,8 @@ or use an explicit `unsafe{ a[..] }`, if you do not want a copy of the slice.',
 			// and the second part is all fields embedded in the structure
 			// If the return value data composition form in `c.table.struct_fields()` is modified,
 			// need to modify here accordingly.
-			c.check_uninitialized_struct_fields(node, type_sym, mut inited_fields)
-
-			for embed in info.embeds {
-				mut zero_struct_init := ast.StructInit{
-					pos: node.pos
-					typ: embed
-				}
-				c.struct_init(mut zero_struct_init, true, mut inited_fields)
-			}
+			c.check_uninitialized_struct_fields_and_embeds(node, type_sym, mut info, mut
+				inited_fields)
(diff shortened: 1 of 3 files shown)


vlib/v/checker/struct.v
```diff
@@ -777,137 +777,24 @@ or use an explicit `unsafe{ a[..] }`, if you do not want a copy of the slice.',
...
 		}
+		.sum_type {
+			first_typ := (type_sym.info as ast.SumType).variants[0]
+			first_sym := c.table.final_sym(first_typ)
+			if first_sym.kind == .struct_ {
+				// Check uninitialized refs/sum types
+				// The variable `fields` contains two parts, the first part is the same as info.fields,
+				// and the second part is all fields embedded in the structure
+				// If the return value data composition form in `c.table.struct_fields()` is modified,
+				// need to modify here accordingly.
```
spytheman, line 792: The comments here should be in the extracted method.
Please also remove the redundant ones above, at lines 776-779.

felipensp: Ok

PR #22069, opened 2024-08-17
cgen: fix for loop with array fixed returned from fn

### LINKED ISSUE #22062: C error - multiple returns values and fixed arrays
### Describe the bug

When returning multiple values, if one of those is a fixed size array, it seems to be the cause of a C error.

### Reproduction Steps

```v
fn double() (bool, [4]i16) {
        return true, [4]i16{}
}

fn foo() {
        _, b := double()
        for i in b {
        }
}
``` 
with `v bug.v`

### Expected Behavior

No C error.

### Current Behavior

```
❯ v bug.v
bug.v:7:6: warning: unused variable: `i`
    5 | fn foo() {
    6 |     _, b := double()
    7 |     for i in b {
      |         ^
    8 |     }
    9 | }
cc: /tmp/v_1000/bug.01J5EG61QWPZ7RVQCYVP6FH9BK.tmp.c:13289: error: invalid operand types for binary operation
builder error:
==================
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.7 294f7e4.8210593

`OS: linux, Linux version 6.7.11-200.fc39.x86_64 (mockbuild@bdb8d24f206645e2bda44c1cf867612d) (gcc (GCC) 13.2.1 20240316 (Red Hat 13.2.1-7), GNU ld version 2.40-14.fc39) #1 SMP PREEMPT_DYNAMIC Wed Mar 27 16:50:39 UTC 2024`
> [!NOTE]
>

felipensp:
Fix #22062

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5784,7 +5784,13 @@ fn (mut g Gen) return_stmt(node ast.Return) {
 						} else {
 							node.types[0]
 						}
-						g.write('memcpy(${tmpvar}.ret_arr, ${g.expr_string(node.exprs[0])}, sizeof(${g.typ(typ)}))')
+						typ_sym := g.table.final_sym(typ)
+						if typ_sym.kind == .array_fixed
+							&& (typ_sym.info as ast.ArrayFixed).is_fn_ret {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #22071, opened 2024-08-17
cgen,checker: fix map generic fn arg passing

### LINKED ISSUE #22035: Passing a map to a generic argument changes a key value from int to an empty string
### Describe the bug

1. Let this map: `{ 'a': 1 }`
2. Enter the method `marshal`
3. Which calls the method `marshal_map`
4. And see it received as `{ 'a': '' }`

This is simplified code. Runnable file is further below.

```v
fn marshal[T](val T) {
  // received { 'a': 1 }
  marshal_map(val)
}

fn marshal_map[T](val map[string]T) {
  // received { 'a': '' }
}

marshal({ 'a': 1 })
```

Apparently, the value `map[string]int` was passed as `map[string]string` from `marshal` to `marshal_map` and the value of the key `'a'` became the default value for `string` - `''`.

### Reproduction Steps

1. Save the code below to the file `marshal_test.v`
2. Run the command `v marshal_test.v`.
3. Observe the following output:

```
test_marshal
 marshal: {'a': 1}
 marshal_map: {'a': ''}
test_marshal_map
 marshal_map: {'a': 1}
```

When `marshal_map[T](val map[string]T)` is called directly, the input map enters it correctly as `map[string]int`.

When `marshal_map[T](val map[string]T)` is called from `marshal[T](val T)`, the input map enters it wrong as `map[string]string`.

The code of `marshal_test.v`, which is very simplified from a JSON parser and formatter:

```v
module json

fn marshal[T](val T) {
	println(' marshal: ${val}')
	assert val['a'] == 1
	marshal_map(val)
}

fn marshal_map[T](val map[string]T) {
	println(' marshal_map: ${val}')
	// assert val['a'] == 1
}

fn test_marshal() {
	println('test_marshal')
	marshal({
		'a': 1
	})
}

fn test_marshal_map() {
	println('test_marshal_map')
	marshal_map({
		'a': 1
	})
}
```

### Expected Behavior

When `marshal_map[T](val map[string]T)` is called from `marshal[T](val T)` with an input parameter of the type `map[string]int`, this type will be passed properly via `marshal` to `marshal_map` and the value of the key `'a'` will be `1`.

### Current Behavior

I guess that `map[string]int` was converted to `map[string]string` and all key values were replaced by strings with default string values - `''`.

### Possible Solution

I wish I knew :-)

When enabling the commented out `assert` in `marshal_map`:

```v
`fn marshal_map[T](val map[string]T) {
	println(' marshal_map: ${val}')
	assert val['a'] == 1
}
```

The compiler fails with a complaint that the right argument was supposed to be a string, as expected from my assumption above:

```❯ v marshal_test.v
marshal_test.v:11:12: error: infix expr: cannot use `int literal` (right expression) as `string`
    9 | fn marshal_map[T](val map[string]T) {
   10 |     println(' marshal_map: ${val}')
   11 |     assert val['a'] == 1
      |               ~~~~~~~~~~
   12 | }
   13 |
```

However, when changing the argument from `1` to `'1'`:

```v
fn marshal_map[T](val map[string]T) {
	println(' marshal_map: ${val}')
	assert val['a'] == '1'
}
```

The compiler fails again, now desiring an integer:

```
❯ v marshal_test.v
marshal_test.v:11:12: error: infix expr: cannot use `string` (right expression) as `int`
    9 | fn marshal_map[T](val map[string]T) {
   10 |     println(' marshal_map: ${val}')
   11 |     assert val['a'] == '1'
      |               ~~~~~~~~~~~~
   12 | }
   13 |
```

It doesn't know what it wants :-)

### V version

V 0.4.7 ddb6685

V full version: V 0.4.7 ddb6685
OS: macos, macOS, 14.5, 23F79
Processor: 12 cpus, 64bit, little endian, Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz

getwd: /Users/ferdipr/Sources/github/v-bug
vexe: /Users/ferdipr/.vup/weekly.2024.32/v
vexe mtime: 2024-08-06 07:59:48

vroot: OK, value: /Users/ferdipr/.vup/weekly.2024.32
VMODULES: OK, value: /Users/ferdipr/.vmodules
VTMP: OK, value: /tmp/v_501

.git/config present: false

thirdparty/tcc: N/A
> [!NOTE]
>

felipensp:
Fix #22035

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -1709,7 +1709,7 @@ fn (mut c Checker) resolve_comptime_args(func ast.Fn, node_ ast.CallExpr, concre
 									if param_sym.info.value_type.has_flag(.generic) {
 										comptime_args[k] = ctyp
 										k++
-										ctyp = c.unwrap_generic(arg_sym.info.value_type)
+										ctyp = c.unwrap_generic(arg_sym.info.key_type)
 									}
 								} else if param_sym.info.value_type.has_flag(.generic) {
 									ctyp = c.unwrap_generic(arg_sym.info.value_type)
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -1698,22 +1698,21 @@ fn (mut c Checker) resolve_comptime_args(func ast.Fn, node_ ast.CallExpr, concre
 						mut ctyp := c.comptime.get_comptime_var_type(call_arg.expr)
 						if ctyp != ast.void_type {
 							arg_sym := c.table.sym(ctyp)
+							param_sym := c.table.final_sym(param_typ)
 							if arg_sym.info is ast.Array && param_typ.has_flag(.generic)
-								&& c.table.final_sym(param_typ).kind == .array {
+								&& param_sym.kind == .array {
 								ctyp = arg_sym.info.elem_type
 							} else if arg_sym.info is ast.Map && param_typ.has_flag(.generic)
-								&& c.table.final_sym(param_typ).kind == .map {
-								map_info := c.table.final_sym(param_typ).info as ast.Map
-								key_is_generic := map_info.key_type.has_flag(.generic)
-								if key_is_generic {
+								&& param_sym.info is ast.Map {
+								if param_sym.info.key_type.has_flag(.generic) {
 									ctyp = c.unwrap_generic(arg_sym.info.key_type)
-								}
-								if map_info.value_type.has_flag(.generic) {
-									if key_is_generic {
+									if param_sym.info.value_type.has_flag(.generic) {
 										comptime_args[k] = ctyp
 										k++
+										ctyp = c.unwrap_generic(arg_sym.info.value_type)
 									}
-									ctyp = c.unwrap_generic(arg_sym.info.key_type)
+								} else if param_sym.info.value_type.has_flag(.generic) {
+									ctyp = c.unwrap_generic(arg_sym.info.value_type)
 								}
 							}
 							comptime_args[k] = ctyp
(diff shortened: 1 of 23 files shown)


vlib/v/checker/fn.v
```diff
@@ -1756,11 +1755,19 @@ fn (mut c Checker) resolve_comptime_args(func ast.Fn, node_ ast.CallExpr, concre
...
+									if key_is_generic {
+										comptime_args[k] = cparam_type_sym.info.key_type
+									}
+									if param_typ_sym.info.value_type.has_flag(.generic) {
+										k2 := if key_is_generic { k + 1 } else { k }
+										comptime_args[k2] = cparam_type_sym.info.value_type
+										if key_is_generic {
+											k++
+										}
+									}
```
spytheman, line 1770: If you do `if key_is_generic {` just once, and split those in the 2 branches, the code of each will be clearer.

vlib/v/gen/c/fn.v
```diff
@@ -1343,11 +1343,21 @@ fn (mut g Gen) resolve_comptime_args(func ast.Fn, mut node_ ast.CallExpr, concre
...
 									comptime_args[k] = cparam_type_sym.info.elem_type
-								} else if param_typ_sym.kind == .map
+								} else if param_typ_sym.info is ast.Map
 									&& cparam_type_sym.info is ast.Map {
-									comptime_args[k] = cparam_type_sym.info.key_type
-									comptime_args[k + 1] = cparam_type_sym.info.value_type
-									k++
+									key_is_generic := param_typ_sym.info.key_type.has_flag(.generic)
+									println(param_typ_sym)
+									println(cparam_type_sym)
```
Delta456, line 1350:
```suggestion


```

Just in case

vlib/v/checker/fn.v
felipensp: ok

spytheman (approved):
Excellent work.

Issue #22021, opened 2024-08-10
Allow to omit commas in multiline function signature block


dadooda:
### Describe the feature

It's cool that Vlang lets us omit the comma when using multiline syntax in array enumeration and function calls. Like this:

```v
// (1) ✅
greet('Mr.', 'Joe')

// (2) ✅
greet(
  'Mr.'
  'Joe'
)
```

At the same time, it's not yet possible to omit the comma in `greet()` definition (signature):

```
// (3) ✅
fn greet(salutation string, name string) {
  println('Hey, ${salutation} ${name}!')
}

// (4) ❌
fn greet(
  salutation string
  name string
) {
  println('Hey, ${salutation} ${name}!')
}
```

I'd propose example (4) above to be made possible.

### Use Case

Comma omission in multiline enumerations looks good, feels good. And it's partially supported already. Let's make it consistently available in all parts of the language.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.7 7baff15

### Environment details (OS name and version, etc.)

Ubuntu 18 under WSL.
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

JalonSolov:
V seems to allow `greet('Mr.' 'Joe')` as well.  I think that should actually be an error.  It is working in V because V still sees it as 2 separate strings, but it shouldn't work, as there really should be a `,` separator between the parameters, whether they are on the same line or not.

For example, this works, too, and it looks really odd...
```v
fn greet(i1 int, i2 int) {
  println('${i1} ${i2}')
}

greet(17 23)
```

dadooda:
> For example, this works, too, and it looks really odd...

Oh yes, it does… The parser definitely needs some reinforcement…

medvednikov:
Good find, will fix.

I like no comma when there're newlines, that's why I implemented this way back:

```v
arr := [
'foo'
'bar'
]
```

Many people are against this, and vfmt was changed to insert a `,`.

I still think a comma shouldn't be needed here. Just like ; Looks cleaner this way.

JalonSolov:
As long as it's documented that either newline or comma works.  But then what about OP's comment about function declarations?  Shouldn't they work the same way, then?

As long as it doesn't work on same line without the comma.

PR #22075, opened 2024-08-19
parser: check fn call args without comma between them (related #22021)

yuyi98:
This PR check fn call args without comma between them (related #22021).

- Check fn call args without comma between them.
- Add test.

```v 
fn greet(i1 int, i2 int) {
	println('${i1} ${i2}')
}

fn main() {
	greet(17 23)
}

PS D:\Test\v\tt1> v run .       
tt1.v:6:11: error: unexpected number `23`, expecting `,`
    4 | 
    5 | fn main() {
    6 |     greet(17 23)
      |              ~~
    7 | }
```
```v
fn greet(i1 int, i2 int) {
	println('${i1} ${i2}')
}

fn main() {
	greet(
		17
		23
	)
}

PS D:\Test\v\tt1> v run .    
tt1.v:8:3: error: unexpected number `23`, expecting `,`
    6 |     greet(
    7 |         17
    8 |         23
      |         ~~
    9 |     )
   10 | }
```

### DIFF
--- a/cmd/tools/vdoc/testdata/basic/basic.out
+++ b/cmd/tools/vdoc/testdata/basic/basic.out
@@ -1,7 +1,3 @@
-Warning: `markdown` exists, but is not updated.
-V will continue, since updates can fail due to temporary network problems,
-and the existing module `markdown` may still work.
---------------------------------------------------
 module basic
 
 const source_root = 'temp' // some const
--- a/cmd/tools/vdoc/testdata/basic/basic.out
+++ b/cmd/tools/vdoc/testdata/basic/basic.out
@@ -1,3 +1,7 @@
+Warning: `markdown` exists, but is not updated.
+V will continue, since updates can fail due to temporary network problems,
+and the existing module `markdown` may still work.
+--------------------------------------------------
 module basic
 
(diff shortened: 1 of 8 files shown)


cmd/tools/vdoc/testdata/basic/basic.out
```diff
@@ -1,3 +1,7 @@
+Warning: `markdown` exists, but is not updated.
+V will continue, since updates can fail due to temporary network problems,
+and the existing module `markdown` may still work.
+--------------------------------------------------
```
Delta456, line 4: How did this get generated?

spytheman: Weird ... Perhaps there was a temporary network error while running `v install markdown`, or perhaps there was a local change in ~/.vmodules/markdown for some reason, and `v update markdown` failed (afaik it is run when doing `v doc`, when vdoc is not yet compiled (or after `v self`).

Imho it should disappear after doing `rm -rf ~/.vmodules/markdown`, followed by `v install markdown` again.

Imho that .out file should not be changed by this PR.

PR #22077, opened 2024-08-19
json2: add compatibility of @[skip] in json2

esquerbatua:
In order to make the v version of json (x.json2) compatible with json, make the same functionality of skip working in json2

### DIFF
--- a/vlib/x/json2/decoder.v
+++ b/vlib/x/json2/decoder.v
@@ -165,6 +165,9 @@ fn decode_struct[T](_ T, res map[string]Any) !T {
 			mut json_name := field.name
 
 			for attr in field.attrs {
+				if attr.contains('skip') {
+					skip_field = true
+				}
 				if attr.contains('json: ') {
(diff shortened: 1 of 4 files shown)


spytheman (approved):
Excellent work.

enghitalo:
What is the difference between `skip` and `json:"-"`?

JalonSolov:
Less typing?

esquerbatua:
> What is the difference between `skip` and `json:"-"`?

Actually in json the anon @[json: '-'] (or I don't see them in the code) do not skip the field, it's new in x.json2, can be removed maybe in json2, but to keep the current functionality I didn't removed

JalonSolov:
It worked in `json`, so needs to be supported in `x.json2` for compatibility, if nothing else.

Personally, I don't know why `json:'-'` was added to `x.json2`, instead of `skip`.

esquerbatua:
True, it's also in `json`, maybe in the future we shoud remove it.

PR #22078, opened 2024-08-19
bechmark: Add min, max & avg to recorded measurements

esquerbatua:
Add more bechmark information to recorded measurements.

old:
image

new:
image


spytheman (approved):
Excellent work.

PR #22079, opened 2024-08-20
test: add like_operator_outside_orm_error.vv/.out (related #22075)

yuyi98:
This PR fix like_operator_outside_orm_error.vv/.out (related #22075).

```v
fn main() {
	name := 'Luke'
	ret := name like 'L%'
	println(ret)
}

PS D:\Test\v\tt1> v run .
tt1.v:3:14: error: only `c`, `r`, `js` are recognized string prefixes, but you tried to use `like`
    1 | fn main() {
    2 |     name := 'Luke'
    3 |     ret := name like 'L%'
      |                 ~~~~
    4 |     println(ret)
    5 | }
```


spytheman:
Why replace the existing test? 

The old message does not fail on master, running `git grep 'unexpected name.*like.*expecting'` shows the message is only tested in that .vv/.out pair.

I think that both the old and the new variants, are valuable to have, and creating a new .vv/.out pair, instead of replacing the existing one, may be better.

Is there something that I miss?

yuyi98:
Ok, I see. I'll add a new test and restore it.

spytheman (approved):
Good work.

PR #22083, opened 2024-08-20
v: fix generic lambda type binding and resolution

### LINKED ISSUE #19426: Lambda not working with generics
### Describe the bug

Lambdas are supposed to work for all call sites taking in a function as a param. However, it's not working when there are generics

### Reproduction Steps

```v
import arrays
a := [1, 2, 3, 4]
//b:= arrays.join_to_string(a, '-', fn (it int) string {return it.str()}) // works
b := arrays.join_to_string(a, '-', |x| x.str()) // fails
println(b)
```

### Expected Behavior

Expected to print the following
```
1-2-3-4
```

### Current Behavior

compilation error 

```
error: generic function declaration must specify generic type names
```

### Possible Solution

_No response_

_No response_

### V version

0.4.1 1c9ab77

MacOS Ventura 13.4.1
> [!IMPORTANT]
> You can vote for this issue using the 👍 reaction. More votes increase the issue's priority
> for developers.
>
> Take into account that only the 👍 reaction counts as a vote.
> Only reactions to the issue itself will be counted as votes, not comments.

felipensp:
Fix #19426

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -1872,6 +1872,7 @@ pub mut:
 	func       &AnonFn = unsafe { nil }
 	is_checked bool
 	typ        Type
+	call_ctx   &CallExpr = unsafe { nil }
 }
 
 pub struct Likely {
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -1585,6 +1585,19 @@ fn (mut c Checker) fn_call(mut node ast.CallExpr, mut continue_check &bool) ast.
 						if c.comptime.type_map.len > 0 {
 							continue
 						}
+						if mut call_arg.expr is ast.LambdaExpr {
+							// Calling fn is generic and lambda arg also is generic
+							if node.concrete_types.len > 0
+								&& call_arg.expr.func.decl.generic_names.len > 0 {
+								call_arg.expr.call_ctx = unsafe { node }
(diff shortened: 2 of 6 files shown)


spytheman (commented):
Excellent work.

Will the same approach work for generic functions, calling other generic functions, with a LambdaExpr ?

Try adding such an example in the test case too, with `|x,y| x + y`, where x and y are in one call 2 ints, and in another call, 2 strings .

JalonSolov:
Perhaps a 3rd case where one is an int and the other a string, for a known error case?

felipensp:
> > Excellent work.
> > Will the same approach work for generic functions, calling other generic functions, with a LambdaExpr ?
> > Try adding such an example in the test case too, with `|x,y| x + y`, where x and y are in one call 2 ints, and in another call, 2 strings .
> 
> Yeah, Just tested:
> 
> ```v
> fn ddd[T](call fn (arg1 T, arg2 T) string, arg1 T, arg2 T) string {
> 	return call(arg1, arg2)
> }
> 
> fn abc[T](arg1 T, arg2 T) string {
> 	return ddd(|x,y| (x+y).str(), arg1, arg2)
> }
> 
> fn main() {
> 	abc('3', '3')
> 	abc(3, 3)	
> 	abc(3.1, 3.1)	
> }
> ```
> 
> It needs more work to make it possible... I'll check it.

I've handled this case. Have you any other test case?

spytheman:
I could not think of any other @felipensp .

Excellent work.

PR #22087, opened 2024-08-21
fmt: cleanup alignment using string.repeat() instead of strings.repeat()

yuyi98:
This PR cleanup alignment using `' '.repeat(n)` instead of `strings.repeat(' ', n)`.


spytheman (approved):
Excellent work.

PR #22090, opened 2024-08-22
vlib: add an `arrays.parallel` module, containing `parallel.run/3` and `parallel.amap/3` implementations

prashanth-hegde:
arrays.run_parallel
arrays.map_parallel

## Description 
When we want to run functions on input array in parallel, a simple for loop does not suffice. 
This implementation uses channels and waitgroups to let the user run a lambda function to run tasks in parallel.
It also allows the user to specify the max workers to achieve parallelism


prashanth-hegde:
Why is this failing on `crypto/pem`? I don't have any changes or dependencies on that.

spytheman:
Not sure about the exact reason, but `vlib/crypto/pem/encode.v` does `import arrays` 🤔 .

spytheman:
Imho move the implementation in `vlib/arrays/parallel/` .
Then users will be able to: `import arrays.parallel`, and `parallel.map(...)` + `parallel.run(...)`.
The dependency problem for people that do not want all of `sync`, `math` and `runtime`, but still want to use the rest of `arrays`, will be also eliminated.

vlib/arrays/parallel_test.v
```diff
@@ -0,0 +1,36 @@
+module arrays
```
spytheman, line 1: It is better for this test to not be an internal one.

Internal tests that do `module somename`, are for testing internal/private functionality. This test only tests the public one (the 2 pub functions), and can just not have `module` at its top, and later use `arrays.run_parallel` etc (or even better `parallel.run` etc).

That way, it can also serve as a run-able example for the module too, enabling people to search for usages more easily.

vlib/arrays/parallel_test.v
```diff
@@ -0,0 +1,36 @@
...
+
+import rand
+import time
+
+fn test_run_parallel() {
+	count := []int{len: 10, init: index}
+	mut res := []string{len: 10}
+	run_parallel[int](count, 2, fn [mut res] (i int) ! {
+		delay := rand.intn(1000) or { 1000 }
```
spytheman, line 10:
```suggestion
		delay := rand.intn(250) or { 250 }
```
There is no need to pessimise the runtime of the test.
With 1000/1000, it runs for over 4 seconds on my machine, because of all the delays.
With 250/250, it averages to a 1 second runtime.

vlib/arrays/parallel_test.v
```diff
@@ -0,0 +1,36 @@
...
+		res << 'task ${i}, delay=${delay}ms'
+	})
+	assert res.len == count.len
+}
+
+// todo: rename the function once the implementation is complete
+fn test_map_parallel() {
+	input := [1, 2, 3, 4, 5, 6, 7, 8, 9]
+	output := map_parallel[int, int](input, 4, fn (i int) int {
+		delay := rand.intn(1000) or { 1000 }
```
spytheman, line 21: same:
```suggestion
		delay := rand.intn(250) or { 250 }
```

prashanth-hegde:
> Imho move the implementation in `vlib/arrays/parallel/` 

Good Idea, will do that

prashanth-hegde:
@spytheman thanks for the review. The review comments are good.
I have addressed your review comments. Please take a look when you can.

vlib/arrays/parallel/parallel.v
```diff
@@ -37,48 +38,57 @@ pub fn run_parallel[T](input []T, max_workers int, worker fn (T) !) {
...
+			task: inp
+		}
 		ch <- task
 	}
 
 	// wait for all tasks to complete
 	wg.wait()
-	ch.close()
-	return output
+	tasks.sort(a.idx < b.idx)
```
spytheman, line 92: Why do you sort here?
imho the previous approach without sorting was better, since it gives the user the option to not do it. 

(with V it is very easy to just slap .sorted() at the end of the call if needed)

vlib/arrays/parallel/parallel.v
```diff
@@ -37,48 +38,57 @@ pub fn run_parallel[T](input []T, max_workers int, worker fn (T) !) {
...
-// map_parallel lets the user run an array of input with a
+// amap lets the user run an array of input with a
 // user provided function in parallel. It limits the number of
-// worker threads to max number of cpus. The function makes use of channels to achieve it
-// The worker function can return a value. The returning array maintains the input order
-// Any error handling should have happened within the worker function
-// Example: squares := arrays.map_parallel([1, 2, 3, 4, 5], 2, fn (i) { return i * i })
-pub fn map_parallel[T, R](input []T, max_workers int, worker fn (T) R) []R {
+// worker threads to max number of cpus. The function makes use of channels to achieve it.
+// The worker function can return a value. The returning array maintains the input order.
```
spytheman, line 52: There is no need to maintain the input order imho, and it can be expensive for larger input sizes.

vlib/arrays/parallel/parallel.v
```diff
@@ -37,48 +38,57 @@ pub fn run_parallel[T](input []T, max_workers int, worker fn (T) !) {
...
 
-// map_parallel lets the user run an array of input with a
+// amap lets the user run an array of input with a
 // user provided function in parallel. It limits the number of
-// worker threads to max number of cpus. The function makes use of channels to achieve it
-// The worker function can return a value. The returning array maintains the input order
-// Any error handling should have happened within the worker function
-// Example: squares := arrays.map_parallel([1, 2, 3, 4, 5], 2, fn (i) { return i * i })
-pub fn map_parallel[T, R](input []T, max_workers int, worker fn (T) R) []R {
+// worker threads to max number of cpus. The function makes use of channels to achieve it.
```
spytheman, line 51: The use of channels is an implementation detail. Imho there is no need to mention it in the fn docs.

vlib/arrays/parallel/parallel.v
```diff
@@ -1,19 +1,20 @@
...
 
-// run_parallel lets the user run an array of input with a
+// run lets the user run an array of input with a
 // user provided function in parallel. It limits the number of
-// worker threads. The function makes use of channels to achieve it
-// The worker function is supposed to not return any error or a return value
-// Any error handling should have happened within the worker function
-// Example: arrays.run_parallel([1, 2, 3, 4, 5], 2, fn (i) { println(i) })
-pub fn run_parallel[T](input []T, max_workers int, worker fn (T) !) {
+// worker threads. The function makes use of channels to achieve it.
```
spytheman, line 9: Same - the use of channels is an implementation detail.

vlib/arrays/parallel/parallel.v
```diff
@@ -22,7 +23,7 @@ pub fn run_parallel[T](input []T, max_workers int, worker fn (T) !) {
 			for {
 				task := <-ch or { break }
 				worker(task) or {
-					eprintln('error executing parallel task...')
+					eprintln('error executing parallel task, aborting')
```
spytheman, line 26: Instead of using eprintln here, make the run function return `!`, and just do `worker(task)!` .

This is a library function, and it is better for library functions to delegate error handling to their callers, which best know the context, and what is appropriate - logging, panicing, retrying etc.

spytheman:
The windows tcc CI failure is unrelated to this PR. The other (the ubuntu musc one) is (it happens for the new test). Not sure why yet.

vlib/arrays/parallel/parallel.v
prashanth-hegde: Sorting later is not possible with this signature because `T` and `R` could be unrelated. 
We could potentially split the implementation into two signatures `map_async()` and `map_parallel()` 
where `map_async()` doesn't keep the order while the other does.

vlib/arrays/parallel/parallel.v
prashanth-hegde: So if we return like this, `wg` will still be waiting at the end on an error and will be stuck indefinitely

vlib/arrays/parallel/parallel.v
spytheman: Splitting is fine too, then the user can use one or the other, depending on what exactly is needed.

vlib/arrays/parallel/parallel.v
spytheman: You can use `defer { wg.done() }`, in the spawned function, so that early returns will still do it.

spytheman:
(rebased over current master, to pickup the windows CI fix, and add some changes to make it easier to diagnose the rest of the CI failures)

spytheman:
The macos failures happen very seldom locally (~20% of the time).
Here is one:
```
[vlib/arrays/parallel/parallel_test.v:20] input: [1, 2, 3, 4, 5, 6, 7, 8, 9]
[vlib/arrays/parallel/parallel_test.v:26] output: [1, 4, 9, 16, 36, 49, 64, 81]
vlib/arrays/parallel/parallel_test.v:29: ✗ fn test_parallel_amap
   > assert output[i] == input[i] * input[i]
     Left value: 36
    Right value: 25
```

The input has 9 elements, the output has 8 (the 25 for 5*5 is missing).
Perhaps there is a race somewhere.

spytheman:
The race was for the `a << e` operations, done in the worker functions. Doing array pushes, is not thread safe: if 2 threads try to do them in a race, `a` will contain in the end only one of the new `e` elements, depending on which thread was last to update `a.data`.

Instead <<, the fix (among other things), pre-allocates the result task array in `amap`, and then just does `tasks_ref[task.idx] = t`, which is safe to races, because each task is only processed by 1 worker thread).

medvednikov:
Would be nice to detect this in runtime in debug mode or with a separate flag.

spytheman:
> Would be nice to detect this in runtime in debug mode or with a separate flag.

Yes, although I have no idea how will it work. V currently does not track what array is used by which threads.

prashanth-hegde:
Thank you! I was busy with work for a couple days. Appreciate the changes!

PR #22097, opened 2024-08-22
v: parser: Fixed #22095

### LINKED ISSUE #22095: V parser segfault with `-silent`, discovered with radamsa, for a very long `-----n----m------` pattern.
**V doctor:**
```
V full version: V 0.4.7 c92577e.18f8999
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-08-22 10:31:40

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.34-21-g18f89991-dirty
.git/config present: true

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg -silent bug7_shortened.v`

The compiler not segfaulting, even with `-silent`

**What did you see instead?**
```
#0 15:33:07 ^ master /space/v/oo>./v -silent bug7_shortened.v
Segmentation fault (core dumped)
```

bug7_shortened.v:
```v
----------n----------------------------------------------------------------------------------m-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------M'Hellprintln('Hel-o, World!')
```
> [!NOTE]
>

theaino:
Apparently, the compiler segfaults when the nested open scope limit is reached in silent mode (as seen in #22095).

This is because a Parser.error doesn't exit when in silent mode.

### DIFF
--- a/vlib/v/parser/parser.v
+++ b/vlib/v/parser/parser.v
@@ -550,6 +550,7 @@ fn (p &Parser) is_array_type() bool {
 fn (mut p Parser) open_scope() {
 	if p.opened_scopes > p.max_opened_scopes {
 		p.error('nested opened scopes limit reached: ${p.max_opened_scopes}')
+		exit(1)
 	}
 	p.scope = &ast.Scope{
 		parent:    p.scope


spytheman:
This is not a good solution, the parser is used as a component in larger apps, especially in `-silent` mode.
It should not call exit(1) in arbitrary places, because that will exit the containing process as well, which may well want to use a parser to parse many v files, and then collect stats on how many errors there were, or to act as a language server etc. 

Instead of calling exit(1), it can call p.error and return early for example.

spytheman:
@theaino, thank you for the PR, however imho https://github.com/vlang/v/issues/22095 , while more complex, is a better long term solution, given all the constraints that affect the parser module and its users.

PR #22098, opened 2024-08-22
parser,ast: protect against more overflows/panics, by forcing early returns on deeply nested expressions and scopes

spytheman:
Fixes https://github.com/vlang/v/issues/22095 .


spytheman:
The CI failure is a bit puzzling... It passes locally for me, with tcc on windows:
изображение

PR #22100, opened 2024-08-23
builtin: cleanup `u8.repeat()` and `rune.repeat()`

yuyi98:
This PR cleanup `u8.repeat()` and `rune.repeat()`.


spytheman (approved):
Excellent work.

PR #22101, opened 2024-08-23
ast: cache ident lookups for consts in ast Expr str

spytheman:
- **ast: cache Ident scope lookups for consts in Expr.str**
- **ast,cgen: remove instrumentation for the cached ident lookup, cleanup**

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -1025,6 +1025,7 @@ pub mut:
 	mod            string
 	name           string
 	full_name      string
+	cached_name    string
 	kind           IdentKind
 	info           IdentInfo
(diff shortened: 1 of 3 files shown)


spytheman:
Updating the .cached_name without synchronization, may cause some of the sanitizers to trigger 🤔 .

I think that since the cached result will be the same, even in the case of a race, it may be safe to add `v__ast__Expr_str` in the suppression lists.

esquerbatua:
Great job!

spytheman:
Thanks.

medvednikov:
Getting a stable 1.5% just from this tiny change is insane.

Makes you think how many more things can be optimized.

PR #22106, opened 2024-08-23
util: cleanup check_module_is_installed(), do not run automatically updates of modules, just because their v.mod is present.

yuyi98:
This PR cleanup check_module_is_installed().

- `pub fn check_module_is_installed(modulename string, is_verbose bool, need_update bool) !bool {...}`
- Cancel the update operation every time if `need_update = false`.

### DIFF
--- a/vlib/v/util/util.v
+++ b/vlib/v/util/util.v
@@ -232,7 +232,7 @@ pub fn launch_tool(is_verbose bool, tool_name string, args []string) {
 	if should_compile {
 		emodules := util.external_module_dependencies_for_tool[tool_name]
 		for emodule in emodules {
-			check_module_is_installed(emodule, is_verbose) or { panic(err) }
+			check_module_is_installed(emodule, is_verbose, false) or { panic(err) }
 		}
 		mut compilation_command := '${os.quoted_path(vexe)} '
 		if tool_name in ['vself', 'vup', 'vdoctor', 'vsymlink'] {
@@ -484,7 +484,7 @@ fn non_empty(arg []string) []string {
 	return arg.filter(it != '')
 }
 
-pub fn check_module_is_installed(modulename string, is_verbose bool) !bool {
+pub fn check_module_is_installed(modulename string, is_verbose bool, need_update bool) !bool {
 	mpath := os.join_path_single(os.vmodules_dir(), modulename)
 	mod_v_file := os.join_path_single(mpath, 'v.mod')
 	murl := 'https://github.com/vlang/${modulename}'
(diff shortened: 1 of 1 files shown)


vlib/v/util/util.v
```diff
@@ -232,7 +232,7 @@ pub fn launch_tool(is_verbose bool, tool_name string, args []string) {
 	if should_compile {
 		emodules := util.external_module_dependencies_for_tool[tool_name]
 		for emodule in emodules {
-			check_module_is_installed(emodule, is_verbose) or { panic(err) }
+			check_module_is_installed(emodule, is_verbose, false) or { panic(err) }
```
spytheman, line 235: the default is `true`, not `false`

yuyi98: Local vdoc_test is always reporting errors due to network problems.
basic.out
```v
Warning: `markdown` exists, but is not updated.
V will continue, since updates can fail due to temporary network problems,
and the existing module `markdown` may still work.
--------------------------------------------------
module basic

const source_root = 'temp' // some const
const another = int(5)
const windowpos_undefined_mask = C.SDL_WINDOWPOS_UNDEFINED_MASK //   0x1FFF0000u
const windowpos_undefined = C.SDL_WINDOWPOS_UNDEFINED
fn funky()
```

spytheman: Try deleting `~/.vmodules/markdown` or renaming it. You may have local changes in it.

spytheman: Another thing to try to diagnose it is: `v install -v markdown` .

yuyi98: Due to network problems.
```v
PS D:\Vlang\v> v install -v markdown
Scanning `markdown`...
Retrieving metadata for `markdown` from `https://vpm.vlang.io/api/packages/markdown`...
error: failed to install `markdown`.
details: Cloning into 'C:\Users\yuyi9\AppData\Local\Temp\v_0\vpm_modules\markdown'...
         error: RPC failed; curl 28 OpenSSL SSL_read: Connection was reset, errno 10054
         fatal: expected 'packfile'
```

spytheman: > Because access to github in China is very unstable.

Oh ... I am sorry, I see :-| .

spytheman: Well, ok then; if it makes it easier to work with less reliable internet, it is a good thing, the cost is just that people will have to manually run `v update` when it is more convenient.

spytheman (approved):
Good work.

yuyi98:
Thanks !!!

Issue #22072, opened 2024-08-18
const name shadowing


Eliyaan:
### Describe the bug

(I'm not sure if it's called like that)
This code is able to reuse the name of a constant on another unrelated variable, and it can cause some mistakes.

### Reproduction Steps

```v
const item = 24

fn main() {
        for i, item in ["abc", "def"] {
                println(item)
        }
}
```

### Expected Behavior

An error telling me that `item` is already used.

### Current Behavior

It just worked

### V version

V 0.4.7

### Environment details (OS name and version, etc.)

linux
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Delta456:
This code compiles successfully. 

```v
const item = 24

fn main() {
	item := 34
	println(item)
}
```

JalonSolov:
V doesn't allow shadowing of other things, so I agree that this is a bug that it allows shadowing of a const.

larpon:
Good find @Eliyaan

PR #22108, opened 2024-08-23
checker: disallow const name in for loop 

Delta456:
Fixes https://github.com/vlang/v/issues/22072


spytheman (approved):
Good work.

medvednikov:
I'm not sure why there needs to be separate logic just in the for loop. Variable creation logic with all the checks for duplicity should be defined once.

PR #22115, opened 2024-08-25
checker: fix generic lambda type binding resolution

### LINKED ISSUE #22109: can not use generic lambda expression as parameter to generic function over 2 generic types
**V doctor:**
```
V full version: V 0.4.7 38ca2e5.27ef543
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-08-24 08:43:03

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.34-41-g27ef5439
.git/config present: true

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg a.v`
```v
pub fn mymap[T, R](input []T, f fn (T) R) []R {
    mut results := []R{cap: input.len}
    for x in input {
        results << f(x)
    }
    return results
}

dump( mymap([1,2,3,4,5], fn (i int) int { return i*i }) ) // works
dump( mymap([1,2,3,4,5], |x| x * x) ) // checker error, but should work
dump( mymap([1,2,3,4,5], |x| u16(x * x)) ) // checker error, but should work too

```

a compiler program without errors

**What did you see instead?**
```
a.v:11:30: error: cannot use `u16` as type `R` in return argument
    9 | dump( mymap([1,2,3,4,5], fn (i int) int { return i*i }) ) // works
   10 | dump( mymap([1,2,3,4,5], |x| x * x) ) // checker error, but should work
   11 | dump( mymap([1,2,3,4,5], |x| u16(x * x)) ) // checker error, but should work too
      |                              ~~~~~~~~~~
a.v:10:30: error: cannot use `int` as type `R` in return argument
    8 | 
    9 | dump( mymap([1,2,3,4,5], fn (i int) int { return i*i }) ) // works
   10 | dump( mymap([1,2,3,4,5], |x| x * x) ) // checker error, but should work
      |                              ~~~~~
   11 | dump( mymap([1,2,3,4,5], |x| u16(x * x)) ) // checker error, but should work too
```
> [!NOTE]
>

felipensp:
Fix #22109

### DIFF
--- a/vlib/v/checker/check_types.v
+++ b/vlib/v/checker/check_types.v
@@ -1069,6 +1069,15 @@ fn (mut c Checker) infer_fn_generic_types(func ast.Fn, mut node ast.CallExpr) {
 							if param_sym.info.func.return_type.nr_muls() > 0 && typ.nr_muls() > 0 {
 								typ = typ.set_nr_muls(0)
 							}
+							// resolve lambda with generic return type
+							if arg.expr is ast.LambdaExpr && typ.has_flag(.generic) {
+								typ = c.comptime.resolve_generic_expr(arg.expr.expr, typ)
+								if typ.has_flag(.generic) {
+									lambda_ret_gt_name := c.table.type_to_str(typ)
+									idx := func.generic_names.index(lambda_ret_gt_name)
+									typ = node.concrete_types[idx]
+								}
+							}
 						}
 					}
 				} else if arg_sym.kind in [.struct_, .interface_, .sum_type] {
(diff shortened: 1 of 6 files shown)


spytheman (approved):
Excellent work 🥳 🚀 ❗

spytheman:
I especially love ❤️, how more and more functionality, is migrated from cgen to the checker. The simpler the code generation is, the better it would be for all the other backends.

PR #22128, opened 2024-08-29
examples: add examples/veb/websocket, to show how to use http connection upgrade to a websocket, from a `veb` route

spytheman:


spytheman:
image

PR #22130, opened 2024-08-29
vweb: add json_raw

esquerbatua:
Add json_raw to prevent encode an encoded body content of response


esquerbatua:
mmm strange
image

JalonSolov:
`vweb` is deprecated now.  Properly deprecated.  Should we be making changes to it?

esquerbatua:
> no, only veb

maybe then we sould create a run version of veb with multithreading?

medvednikov:
what do you mean? veb is built on top of picoev, it's parallel

PR #22131, opened 2024-08-29
vweb: mark deprecated

esquerbatua:


medvednikov:
yeah that's going to break a lot of tests :)

I wanted to deprecate it after adding implicit contexts to V, so that migration from vweb to veb is easier

I'll finish it this week

spytheman:
Can we please instead, just extract `vlib/vweb` to <https://github.com/vlang/vweb> , so that it can still be used by people that want it, by just doing `v install vweb`?

`vweb` -> `veb` is a breaking change, and can not be automated in the general case, without introducing lots of hacks and edge cases.

Making `vweb` deprecated, while it stays in vlib, is forcing people to migrate their projects to `veb`, without giving them time to do it.

spytheman:
> adding implicit contexts to V

I think that doing this should be well thought out, and for general use, not just hacks to make the `veb` routes look more like the `vweb` ones.

esquerbatua:
> Can we please instead, just extract `vlib/vweb` to https://github.com/vlang/vweb , so that it can still be used by people that want it, by just doing `v install vweb`?
> 
> `vweb` -> `veb` is a breaking change, and can not be automated in the general case, without introducing lots of hacks and edge cases.
> 
> Making `vweb` deprecated, while it stays in vlib, is forcing people to migrate their projects to `veb`, without giving them time to do it.

It's also a breaking change, maybe mark it as deprecated when veb it's ready? or only mark it at level of docs?

spytheman:
> It's also a breaking change

Moving to a separate repo, is a breaking change, but with much less impact - with it, after `v install vweb`, all existing code that depends on it, will still compile, with no warnings at all.

esquerbatua:
Removed the deprecated tags

spytheman (approved):

PR #22132, opened 2024-08-29
veb: extract constants into consts.v

esquerbatua:
Move consts of veb to another file


vlib/veb/veb.v
```diff
@@ -314,7 +166,10 @@ pub fn run_at[A, X](mut global_app A, params RunParams) ! {
 	// reserve space for read and write buffers
 	pico_context.buf = unsafe { malloc_noscan(picoev.max_fds * veb.max_read + 1) }
 	defer {
-		unsafe { free(pico_context.buf) }
+		unsafe {
+			free(pico_context.buf)
+			free(pico_context.idx)
+		}
 	}
```
spytheman, line 173: Please do not do functional changes, at the same time as refactoring ones.

spytheman: You can open another PR for the adding of `free(pico_context.idx)`.

vlib/veb/consts.v
```diff
@@ -0,0 +1,149 @@
+module veb
```
Delta456, line 1: fornat this file

vlib/veb/veb.v
```diff
@@ -439,8 +293,8 @@ fn handle_write_file(mut pv picoev.Picoev, mut params RequestParams, fd int) {
 fn handle_write_string(mut pv picoev.Picoev, mut params RequestParams, fd int) {
 	mut bytes_to_write := int(params.string_responses[fd].str.len - params.string_responses[fd].pos)
 
-	if bytes_to_write > veb.max_write {
-		bytes_to_write = veb.max_write
```
medvednikov, line 443: vfmt is so silly here

moving the definition to a different file in the same module changes its logic

PR #22135, opened 2024-08-30
veb: improve memory free

esquerbatua:


vlib/veb/veb.v
```diff
@@ -314,7 +314,10 @@ pub fn run_at[A, X](mut global_app A, params RunParams) ! {
 	// reserve space for read and write buffers
 	pico_context.buf = unsafe { malloc_noscan(picoev.max_fds * veb.max_read + 1) }
 	defer {
-		unsafe { free(pico_context.buf) }
+		unsafe {
+			free(pico_context.idx)
```
spytheman, line 318:
```suggestion
			pico_context.idx.free()
```
There are also other arrays in that struct instance, not just the idx one.
Note however, that the whole defer can be deleted completely - it will never be executed, since `pico.serve()` will not return and has an endless loop in it.

esquerbatua:
closed until have more improvements, to prevent use the github actions

PR #22140, opened 2024-08-31
builtin: fix map.clear() not resetting map's metas and keys blocks (fix #22139)

### LINKED ISSUE #22139: Can't add same element to cleared map
### Describe the bug

Code: https://play.vlang.io/p/b906a74650

```v
mut ints := map[int]int

ints[5] = 5
assert(ints.len == 1)
ints.clear()
assert(ints.len == 0)
ints[3] = 3
assert(ints.len == 1)
ints.clear()
assert(ints.len == 0)
ints[3] = 3
assert(ints.len == 1) // panics! does not panic if line before is "ints[5] = 5"
```

### Reproduction Steps

Run the code above. It will panic on the last assert()

### Expected Behavior

There should be no panic. `ints[3] = 3` should add an element to the cleared (empty) map, but it does not.

### Current Behavior

Output:

```
code.v:12: FAIL: fn main.main: assert (ints.len == 1)
V panic: Assertion failed...
v hash: 217b191
/tmp/v_60000/../../../../../../home/admin/v/vlib/builtin/builtin.c.v:136: at _v_panic: Backtrace
/tmp/v_60000/../../../../../../box/code.v:14: by main__main
/tmp/v_60000/../../../../../../tmp/v_60000/code.01J6MJA57JHHGAQJ61AMMGVTY3.tmp.c:26253: by main
Exited with error status 1
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.7 217b191

```
V full version: V 0.4.7 217b191

getwd: /home/admin/playground
vexe: /home/admin/v/v
vexe mtime: 2024-08-31 15:25:26

VTMP: OK, value: /tmp/v_0

.git/config present: true

```
> [!NOTE]
>

spytheman:
Fix #22139 .

### DIFF
--- a/cmd/tools/bench/map_clear.v
+++ b/cmd/tools/bench/map_clear.v
@@ -1,17 +1,19 @@
 import benchmark
 
-max_iterations := arguments()[1] or { '1_000_000' }.int()
-assert max_iterations > 0
-mut m := {
-	123: 456
-	789: 321
+fn main() {
+	max_iterations := arguments()[1] or { '1_000_000' }.int()
+	assert max_iterations > 0
+	mut m := {
+		123: 456
+		789: 321
+	}
+	mut volatile sum := u64(0)
+	mut b := benchmark.start()
+	for i in 0 .. max_iterations {
+		m.clear()
+		m[i] = i * 2
+		sum += u64(m.len)
+	}
+	assert m.len == 1
+	b.measure('m.clear(), iterations: ${max_iterations}, sum: ${sum}')
 }
-mut volatile sum := u64(0)
-mut b := benchmark.start()
-for i in 0 .. max_iterations {
-	m.clear()
-	m[i] = i * 2
-	sum += u64(m.len)
-}
-assert m.len == 1
-b.measure('m.clear(), iterations: ${max_iterations}, sum: ${sum}')
--- a/cmd/tools/bench/map_clear.v
+++ b/cmd/tools/bench/map_clear.v
@@ -1,0 +1,17 @@
+import benchmark
+
+max_iterations := arguments()[1] or { '1_000_000' }.int()
+assert max_iterations > 0
+mut m := {
+	123: 456
+	789: 321
+}
+mut volatile sum := u64(0)
+mut b := benchmark.start()
+for i in 0 .. max_iterations {
+	m.clear()
+	m[i] = i * 2
+	sum += u64(m.len)
+}
+assert m.len == 1
+b.measure('m.clear(), iterations: ${max_iterations}, sum: ${sum}')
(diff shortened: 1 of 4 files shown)


ka-weihe:
Hi @spytheman 😄 

I think you also have to clear/free the memory of the denseArray. I think that right now it will just leak.

spytheman:
@ka-weihe thanks for reviewing.

I think, that the explicit goal of that .clear() method, rather than just `m = []`, was to *not free* the main memory blocks, used for the keys and values, so that they would be available right away, for reuse, in subsequent `m[key]=val` operations.

Is there a mechanism, where the old blocks would just leak (rather than being reused), and new ones will be allocated instead, when the user does this for example:
```v
mut m := map[int]int{}
m[123] = 456
m[789] = 321
m.clear()
m[1] = 10
m[2] = 20
```

ka-weihe:
> Is there a mechanism, where the old blocks would just leak (rather than being reused), and new ones will be allocated instead, when the user does this for example:

> ```v
> mut m := map[int]int{}
> m[123] = 456
> m[789] = 321
> m.clear()
> m[1] = 10
> m[2] = 20
> ```

Hmm, after looking at the code again, I think it might actually be fine, but it would be interesting to run it in a loop and see if the memory usage keeps increasing. I don't think it will, but it's a good test.

JalonSolov:
Yes, the `.clear()` was to mark the map as "empty", but not change the amount of memory already allocated.  Too much memory churn that way.

You can always do `m = map[int]int{}` again to completely wipe it, free the memory, and start over.  Or just `m = {}` if that works.

spytheman:
> Hmm, after looking at the code again, I think it might actually be fine, but it would be interesting to run it in a loop and see if the memory usage keeps increasing. I don't think it will, but it's a good test.

I've added `cmd/tools/bench/map_clear.v` and `cmd/tools/bench/map_clear_runner.vsh` to make it easier to test that, for different options and iteration counts.
The results is that there is no increase in the memory usage, for me on Ubuntu 20.04:
image

I've tested with these commands on my latest commit, if someone else wants to check:
`FLAGS=''          cmd/tools/bench/map_clear_runner.vsh 1_000_000 100_000_000 10_000_000`
`FLAGS='-gc none'  cmd/tools/bench/map_clear_runner.vsh 1_000_000 100_000_000 10_000_000`
`FLAGS='-prod -cc clang'          cmd/tools/bench/map_clear_runner.vsh`
`FLAGS='-gc none -prod -cc clang' cmd/tools/bench/map_clear_runner.vsh`

(you will need the latest commit 63a51ff)

spytheman:
Here is a heaptrack view after running:
`v -gc none -cc gcc -cg -keepc cmd/tools/bench/map_clear.v`
`heaptrack cmd/tools/bench/map_clear 50_000_000` (the run takes ~4.6 seconds)
`heaptrack --analyze "/space/v/oo/heaptrack.map_clear.87529.zst"`
image

spytheman:
And the same, but after a longer run `heaptrack cmd/tools/bench/map_clear 99_000_000` (takes 8.8s):
image

ka-weihe (approved):

PR #22142, opened 2024-08-31
cgen, scanner: reduce allocations for the most common cases

spytheman:
- **scanner: avoid new allocations, while scanning strings with no escapes**
- **cgen: avoid string interpolation in Gen.new_tmp_var/0**


medvednikov:
The re-write is going to use arena allocation, like the D compiler.

JalonSolov:
Still good to improve what we have, since the rewrite is taking a while.

PR #22152, opened 2024-09-03
flag: add support for parsing `flag.FlagParser` style flags in `to_struct[T]`

larpon:
This PR adds support for parsing flags that `flag.FlagParser` also supports (`--flag value`) - it is something I overlooked when initially implementing the functionality. I am working on simplifying `vab`'s flag handling and bumped into this :slightly_smiling_face:

### DIFF
--- a/vlib/flag/v_long_style_flags_test.v
+++ b/vlib/flag/v_long_style_flags_test.v
@@ -5,6 +5,7 @@ const exe_and_v_long_args = ['/path/to/exe', '--version', '--d', 'ident=val', '-
 	'--test', 'abc', '--done', '--define', 'two', '--live']
 const exe_and_v_long_args_with_tail = ['/path/to/exe', '--version', '--d', 'ident=val', '--test',
 	'abc', '--done', '--d', 'two', '--live', 'run', '/path/to']
+const error_wrong_assignment_flags = ['--o=error']
 
 struct Prefs {
 	version    bool @[short: v]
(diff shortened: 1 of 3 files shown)


vlib/flag/v_long_style_flags_test.v
```diff
@@ -0,0 +1,88 @@
+// Test .v (V) parse style
+import flag
+
+const exe_and_v_long_args = ['/path/to/exe', '--version', '--d', 'ident=val', '--o', '/path/to',
+	'--test', 'abc', '--done', '--define', 'two', '--live']
+const exe_and_v_long_args_with_tail = ['/path/to/exe', '--version', '--d', 'ident=val', '--test',
```
spytheman, line 6: Why is there a `--d` instead of a `-d` flag here?

spytheman (approved):

vlib/flag/v_long_style_flags_test.v
larpon: Just test data (admittedly copied and modified from `v_style_flags_test.v`) - it could be mapped to some other single character. It is just to test single char flags do not mix up when there's a `=` found in the argument.

larpon:
I just pushed a minor update + test case to an error message that was worded badly

vlib/flag/v_long_style_flags_test.v
larpon: Changed to `--p` / `--pop` in [b348fff](https://github.com/vlang/v/pull/22152/commits/b348fff2ff4863486f7f6c926219bddb5f870914)

PR #22153, opened 2024-09-03
Added mandelbrot set generator in example.

Bruno-Vdr:
This PR introduce in v/examples a new example of V program under v/examples/mandelbrot directory, that compute a mandelbrot's set. It was made for learning and testing purposes.

### DIFF
--- a/examples/mandelbrot/src/complex/complex.v
+++ b/examples/mandelbrot/src/complex/complex.v
@@ -1,0 +1,39 @@
+module complex
+
+import math
+
+/**
+ * Minimalist and incomplete implementation of a Complex type, for learning purposes.
+ * This code contains the bare minimum to allow Mandelbrot's set computation.
+ */
+pub struct Complex {
+pub:
+	re f64
+	im f64
+}
+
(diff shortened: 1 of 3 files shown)


lcheylus (changes requested):
Complex type and methods are already available in `vlib/math/complex/complex.v` => no need to reimplement them.

Bruno-Vdr:
The minimalistic re-implementation was done to allow comparison against  other language, with also custom Complex types and respective operator overloading. I could use the vlib/math/complex/complex.v but the **example** will no more show operator overload. The vlib complex type is a slower, I didn't investigate why. Just tell me if you still prefer an implementation with vlib complex type, I'll do it.

spytheman:
There are already 2 Mandelbrot implementations, that demonstrate different V language features, and that produce immediate visual results in:
1) `examples/gg/mandelbrot.v`
2) `examples/wasm/mandelbrot/`

I do not see the value in adding a third one in the main repo 🤔 .
Imho it would be better as a separate repository.

PR #22162, opened 2024-09-04
flag: correct bool logic, add test

larpon:


vlib/flag/flag_to.v
```diff
@@ -907,7 +907,7 @@ pub fn (fm FlagMapper) to_struct[T](defaults ?T) !T {
 					if arg := f.arg {
 						return error('can not assign `${arg}` to bool field `${field.name}`')
 					}
-					result.$(field.name) = true
+					result.$(field.name) = !result.$(field.name)
```
spytheman, line 910: I think that it should take into account the default value, not just the current one.

Otherwise, for repeated flags like `--no-parallel --no-parallel`, you will toggle it 2 times, effectively canceling the effect.

spytheman: Hm, interesting ... I get this, if I try to add another `'--no-parallel'` flag to the input:
```
vlib/flag/flag_to_bool_test.v:14: ✗ fn test_bool_flags failed propagation with error: flag `--no-parallel ` is already mapped to field `parallel` via `--no-parallel `
```

spytheman: Does that mean that passing `-v -v -v` is not allowed currently?

spytheman: Hm, also interesting, but an unrelated problem...
Doing `default := defaults or { T{} }` leads to a cgen error 🤔, but `the_default` works.

spytheman:
(rebased over the merged fix from #22160

vlib/flag/flag_to.v
larpon: Thanks for the fix and good catch. I need to look at it with fresh eyes. For `--no-parallel --no-parallel` to work we'd need to implement either support for `[]bool` in the comptime mapper, which is currently missing or via some other mechanism like keeping a count in an `int` or implement support for `repeats` for flags other that short (`-vvvvv`) 🤔

larpon: > Does that mean that passing `-v -v -v` also is not allowed currently?

It think it is allowed when using POSIX short (`.short` and `.short_long`) when the field is of type `int` but not in other parse modes IIRC

PR #22169, opened 2024-09-06
docs: add a section for the `-compress` flag

esquerbatua:


doc/docs.md
```diff
@@ -6054,7 +6054,7 @@ Full list of builtin options:
 | `solaris`, `termux`            |                  |                               | `js_node`, `js_browser`, `js_freestanding`    |
 | `serenity`, `vinix`, `plan9`   |                  |                               | `interpreter`, `es5`, `profile`, `wasm32`     |
 |                                |                  |                               | `wasm32_emscripten`, `wasm32_wasi`            |
-|                                |                  |                               | `native`, `autofree`                          |
+|                                |                  |                               | `native`, `autofree`, `compress`              |
```
spytheman, line 6057: `compress` is not a builtin option name, i.e. you can not do `$if compress {` .
Imho revert this change.

doc/docs.md
```diff
@@ -6920,6 +6921,20 @@ with identical hardware.
...
 
+#### `-compress`
+
+This flag executes `upx` to compress the resultant executable,
+reducing its size by around 50%-70%.
+
+**When to Use**
+
+- For really tiny envs that the size it's a first, but at runtime everything will run as
+if it had no compression.
```
spytheman, line 6932:
```suggestion
- For really tiny environments, where the size of the executable on the file system, or when deploying is important. The executable will be uncompressed at runtime, so it will take a bit more time to start.
```

doc/docs.md
```diff
@@ -6920,6 +6921,20 @@ with identical hardware.
...
+
+**When to Avoid**
+
+- When you need to debug the application.
```
spytheman, line 6936:
```suggestion
- When you need to debug the application, or when your app's startup time
is extremely important (where 1-2ms can be meaningful for you).
```

doc/docs.md
esquerbatua: done

spytheman (approved):
Good work.

esquerbatua:
> Good work.

ty too for your modifications

PR #22172, opened 2024-09-07
flag: fix parsing `flag.FlagParser` style short flags in `to_struct[T]`

larpon:
This PR corrects some misunderstandings about what flags the `flag.FlagParser` supports.
Before: `--s <value>`
After: `-s <value>` (or `-s` for `bool` fields)

I decided to rename the `style` `enum` from `.v_long` to the more clear `.v_flag_parser` - I do not think the `.v_long` parsing style is much used in the  wid used - so breakage in userland should be minimal.

### DIFF
--- a/vlib/flag/v_flag_parser_style_flags_test.v
+++ b/vlib/flag/v_flag_parser_style_flags_test.v
@@ -1,16 +1,19 @@
-// Test .v (V) parse style
+// Test .v_flag_parser (V flag.FlagParser) parse style
 import flag
 
-const exe_and_v_long_args = ['/path/to/exe', '--version', '--p', 'ident=val', '--o', '/path/to',
+const exe_and_v_flag_parser_args = ['/path/to/exe', '--version', '-p', 'ident=val', '--o', '/path/to',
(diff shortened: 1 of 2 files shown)


vlib/flag/v_flag_parser_style_flags_test.v
```diff
@@ -47,48 +52,74 @@ fn test_long_v_style_no_exe() {
...
 	assert prefs.not_mapped == 'not changed'
-	assert prefs.tail.len == 2
+	assert prefs.tail.len == 3
 	assert prefs.tail[0] == 'run'
 	assert prefs.tail[1] == '/path/to'
+	assert prefs.tail[2] == 'platforms;android-21'
+}
+
+fn test_long_v_style_with_exe_and_short_alias() {
+	prefs, _ := flag.to_struct[Prefs](exe_and_v_flag_parser_help_short_arg,
```
spytheman, line 98: This `exe_and_v_flag_parser_help_short_arg` is only used right here.
Imho just inline the const. The indirection is not useful.

larpon: Good point. Fixed in #22174

PR #22184, opened 2024-09-09
net.mbedtls: add stores client ip and shutdown on handshake failure

martinskou:
Stores clients ip in the SSLConn struct.
Calls conn.shutdown() in event of handshake failure to avoid hanging CLOSE_WAIT sockets.

### DIFF
--- a/vlib/net/mbedtls/ssl_connection.c.v
+++ b/vlib/net/mbedtls/ssl_connection.c.v
@@ -114,6 +114,7 @@ pub mut:
 	handle    int
 	duration  time.Duration
 	opened    bool
+	ip        string
 
 	owns_socket bool
 }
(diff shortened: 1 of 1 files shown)


spytheman (approved):

vlib/net/mbedtls/ssl_connection.c.v
```diff
@@ -256,14 +257,18 @@ pub fn (mut l SSLListener) accept() !&SSLConn {
...
+
+	mut ret := C.mbedtls_net_accept(&l.server_fd, &conn.server_fd, &ip, 16, &iplen)
 	if ret != 0 {
 		return error_with_code("can't accept connection", ret)
 	}
 	conn.handle = conn.server_fd.fd
 	conn.owns_socket = true
+	if iplen == 4 {
+		conn.ip = '${ip[0]}.${ip[1]}.${ip[2]}.${ip[3]}'
+	}
```
spytheman, line 271: What happens for ipv6 addresses?

martinskou: not sure, should prob. be joined like y:y:y:y:y:y:y:y i'll fix when i get to it.

PR #22183, opened 2024-09-09
vfmt: fix const name was not prefixed with module name (fix #14400)

yuyi98:
This PR fix const name was not prefixed with module name (fix #14400).

- Fix const name was not prefixed with module name.
- Add test.

```v
module tt1

const full_name = 'Robert the Tomato'

fn print_full_name() {
	println(tt1.full_name)
	println(tt1.full_name)
}

fn print_name() {
	println(tt1.name)
	println(tt1.name)
}
```
```v
module tt1

const name = 'Bob'
```

PR #22185, opened 2024-09-09
fmt: remove the prefixed module name of the const name in same module. (related #22183)

yuyi98:
This PR remove the prefixed module name of the const name in same module.(related #22183)

- Remove the prefixed module name of the const name in same module.
- Modify all the related files and tests.

```v
module tt1

const full_name = 'Robert the Tomato'

fn print_full_name() {
	println(full_name)
	println(full_name)
}

fn print_name() {
	println(name)
	println(name)
}
```
```v
module tt1

const name = 'Bob'
```


spytheman (approved):
Excellent work 🥳 !
Less arbitrary rules and exceptions in v.fmt, producing shorter and cleaner code ❤️ .

PR #22188, opened 2024-09-09
transformer: optimise calls to write string with simple interpolations

spytheman:
- **transformer: add support for instrumenting the V compiler with `-d trace_transformer`**
- **transformer: optimise out simple string interpolations inside sb.write_string/1 calls**


spytheman:
(rebased over current master)

spytheman:
Comparison with current master, for compiling the V compiler to a .c file:
image

PR #22191, opened 2024-09-10
flag: add parsing mode that turn flag match errors into `no_match` instead

larpon:
This PR adds a parsing mode for `FlagMapper` that makes it easier to identify what flags the `FlagMapper` would otherwise error on.

### DIFF
--- a/vlib/flag/flag_to.v
+++ b/vlib/flag/flag_to.v
@@ -19,8 +19,8 @@ struct FlagContext {
 }
 
 pub enum ParseMode {
-	strict            // return errors for unknown or malformed flags per default
-	error_as_no_match // ignore what would have been errors and add to `no_match` list
+	strict         // return errors for unknown or malformed flags per default
+	relax_no_match // pass thru flag match errors and add them to `no_match` list instead
 }
 
 pub enum Style {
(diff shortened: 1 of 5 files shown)


larpon:
An alternative name for the mode I considered was `.relaxed` since it relaxes specific errors, that may be a shorter/better name for the mode :shrug:

spytheman:
An alternative is `allow_non_matching`, but it is a lot longer.

`relaxed` is more mnemonic and shorter. 

In my opinion, `relaxed` is more ergonomic overall, and contrasts well with the default `strict` mode.

larpon:
> An alternative is `allow_non_matching`, but it is a lot longer.
> 
> `relaxed` is more mnemonic and shorter.
> 
> In my opinion, `relaxed` is more ergonomic overall, and contrasts well with the default `strict` mode.

You are right. Let me change to that

spytheman:
`passthru` has too many consonants and a double `s`, which will make it prone to typos.

larpon:
Renamed in [21fe88a](https://github.com/vlang/v/pull/22191/commits/21fe88a2e1286033157475d816ae40d0c8ce6326)

Issue #22195, opened 2024-09-10
Show where bad union code is located


jorgeluismireles:
### Describe the bug

Next code yields correctly an union use error, but the compiler message doesn't indicate where the offending code is located. Specially when the bad code is in another imported module, it is very difficult to find it.
```v
pub struct Lines {
pub:
	from u8
	to   u8
	dst  u8
	src  u8
}

pub union Head {
	Lines
pub:
	serial [3]u8
}

pub struct Message {
	Head
pub:
	cmd     u8
	payload []u8
}

fn new_code_using_union_correctly() {
	m := Message{
		Head: Head{ serial: [u8(1),2,3]! }
	}
	assert unsafe { m.serial[0] == 1 }
	n := Message{
		Head: Head{ 
			Lines: Lines{ dst:11, src:10 } 
		} 
	}
	assert unsafe { n.dst == 11 }
}

// ...

fn old_code_that_worked_before_changing_message_to_union() {
	m := Message{
		from: 1  // <- INDICATE HERE IS THE PROBLEM
		src:  1
	}
	println(m)
}

// ...

```
Error without the details to find the problem:
```bash
$ v run union.v 
cgen error: union must not have more than 1 initializer
```

### Reproduction Steps

Run the above code and get the incomplete (to say) compiler message.

### Expected Behavior

The description of the error, in this case `cgen error: union must not have more than 1 initializer` but add the file name and the lines where the problem was found, like in the rest of compiler indications.

### Current Behavior

Simple compiler message without helping to find the bad code.

### V version

V 0.4.7 6aeef5e

### Environment details (OS name and version, etc.)

V full version: V 0.4.7 2cde320.6aeef5e
OS: linux, Ubuntu 20.04.6 LTS
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz

Git version: git version 2.25.1
Git vroot status: weekly.2024.29-254-g6aeef5e4
.git/config present: true

CC version: cc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0
thirdparty/tcc status: thirdparty-linux-amd64 a0799a5b

medvednikov:
I didn't even know you could embed a union like that :) This feature needs testing.

jorgeluismireles:
[...]

PR #22196, opened 2024-09-10
checker: restrict multiple union fields at once 

Delta456:
Fixes https://github.com/vlang/v/issues/22195


spytheman (approved):
Excellent work.

PR #16144, opened 2022-10-22
builtin: create string.trim_indexes as a general function to be used in string.trim

l1mey112:
A refactor over `string.trim` functions to allow for getting the indexes to what parts of the string was chopped off.

This PR creates `string.trim_indexes`, similar to `string.trim` but returns the new start and end indexes without manipulating the string. The `string.trim` function is now implemented with a call to this function and passing it's return values to `string.substr` unchanged.

```v
hello_string := ' ffHello V ffff'

left, right := hello_string.trim_indexes(' f')

assert hello_string.trim(' f') == hello_string.substr(left, right)
```

This is incredibly helpful if you want an indication on what has been trimmed off, it is used inside my project where I need to return correct locations for error reporting on trimmed strings.

\- l-m

PR #22205, opened 2024-09-12
builtin: fix string.trim() (fix #13021)

### LINKED ISSUE #13021: string.trim() does not work with unicode correctly
**V version:0.2.4 4b55800** 
**OS:ArchLinux** 

**What did you do?**
```v
fn main(){
	s := 'あいうえお'
	println(s)
	ts := s.trim('あい')
	println(ts)
}
```

```
> v run main3.v
あいうえお
うえお
```
 
**What did you see instead?**
```
> v run main3.v
あいうえお
�えお
```

yuyi98:
This PR fix string.trim() (fix #13021).

- Fix string.trim().
- Add test.

```v
fn main(){
	s := 'あいうえお'
	println(s)
	ts := s.trim('あい')
	println(ts)
}

PS D:\Test\v\tt1> v run .
あいうえお
うえお
```

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -1720,14 +1720,15 @@ pub fn (s string) trim(cutset string) string {
 	if s == '' || cutset == '' {
 		return s.clone()
 	}
-	left, right := s.trim_indexes(cutset)
-	return s.substr(left, right)
+	if cutset.len_utf8() == cutset.len {
+		return s.trim_chars(cutset)
+	} else {
+		return s.trim_runes(cutset)
+	}
 }
 
-// trim_indexes gets the new start and end indices of a string when any of the characters given in `cutset` were stripped from the start and end of the string. Should be used as an input to `substr()`. If the string contains only the characters in `cutset`, both values returned are zero.
-// Example: left, right := '-hi-'.trim_indexes('-')
 @[direct_array_access]
-pub fn (s string) trim_indexes(cutset string) (int, int) {
+pub fn (s string) trim_chars(cutset string) string {
 	mut pos_left := 0
 	mut pos_right := s.len - 1
 	mut cs_match := true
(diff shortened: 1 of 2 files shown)


vlib/builtin/string.v
```diff
@@ -1720,14 +1720,15 @@ pub fn (s string) trim(cutset string) string {
...
+	} else {
+		return s.trim_runes(cutset)
+	}
 }
 
-// trim_indexes gets the new start and end indices of a string when any of the characters given in `cutset` were stripped from the start and end of the string. Should be used as an input to `substr()`. If the string contains only the characters in `cutset`, both values returned are zero.
-// Example: left, right := '-hi-'.trim_indexes('-')
 @[direct_array_access]
-pub fn (s string) trim_indexes(cutset string) (int, int) {
+fn (s string) trim_chars(cutset string) string {
```
spytheman, line 1731: why make this private?

spytheman: `pub fn (s string) trim_indexes(cutset string) (int, int) {` is public since a139bed which is from 2022-10-22. 

The commit was titled:
`builtin: add string.trim_indexes method, that can be used in string.trim, but also separately from it (#16144)`

Even though it is not currently used by anything else in vlib, we do not know that for other V projects. Imho making it private and changing the signature of it, would be a breaking change.

I think that it will be better to keep the existing trim_indexes method public, and its functionality, and to add a new private function instead, that is used for trim.

vlib/builtin/string_test.v
```diff
@@ -921,16 +921,7 @@ fn test_trim() {
...
-fn test_trim_indexes() {
-	mut left, mut right := 0, 0
-	left, right = '- -- - '.trim_indexes(' -')
-	assert left == 0 && right == 0
-	left, right = '- hello-world!\t'.trim_indexes(' -\t')
-	assert left == 2 && right == 14
-	left, right = 'abc'.trim_indexes('ac')
-	assert left == 1 && right == 2
+	assert 'あいうえお'.trim('あい') == 'うえお'
 }
```
spytheman, line 925: The existing tests for `trim_indexes` should stay imho.

yuyi98: Ok, I'll restore them.

spytheman (approved):
Excellent work.

Issue #12183, opened 2021-10-14
Short enum syntax doesn't work across modules (`http.CommonHeader.content_type` instead of just `.content_type`)


medvednikov:
This is supposed to work:

```
header := http.new_header_from_map({
        .content_type: 'application/json'
})
```

Instead we have to do this:

```
header := http.new_header_from_map({
        http.CommonHeader.content_type: 'application/json'
})
```

yuyi98:
It's right now.

```v
import net.http

fn main() {
	header := http.new_header_from_map({
		.content_type: 'application/json'
	})

	println(header)
}

PS D:\Test\v\tt1> v run .       
Content-Type: application/json
```

PR #22210, opened 2024-09-13
test: add a test for short enum syntax across modules (related #12183)

yuyi98:
This PR add a test for short enum syntax across modules (related #12183).

- Add a test for short enum syntax across modules.
- Modify all the related files.

```v
import net.http

fn main() {
	header := http.new_header_from_map({
		.content_type: 'application/json'
	})

	println(header)
}

PS D:\Test\v\tt1> v run .       
Content-Type: application/json
```


spytheman (approved):
Nice.

Issue #13233, opened 2022-01-20
vfmt incorrectly makes `const_abc` prefix in C binding separate parameter


ChewingBever:
**V version:** V 0.2.4 758d1f4.a291dca
**OS:**  linux, "EndeavourOS Linux"

**What did you do?**
I tried to write a C binding that required taking in a const struct pointer.

`fn C.archive_entry_copy_stat(&C.archive_entry, const_stat &C.stat)`

**What did you expect to see?**

Nothing, just leave this line as is.
 
**What did you see instead?**

vfmt changed the line to
`fn C.archive_entry_copy_stat(&C.archive_entry, const_stat, &C.stat)`
causing the function to be interpreted as having three arguments instead of two.

yuyi98:
There is no problem now.

PR #22212, opened 2024-09-13
fmt: add a test for fn with c binding type args (related #13233)

yuyi98:
This PR add a test for fn with c binding type args (related #13233).

```v
fn C.archive_entry_copy_stat(&C.archive_entry, const_stat &C.stat)
```


spytheman (approved):
Thank you, for going back, to close those old issues.

PR #22213, opened 2024-09-13
all: add support for `@BUILD_DATE`, `@BUILD_TIME` and `@BUILD_TIMESTAMP`, all using v.util.get_build_time(), and overridable through SOURCE_DATE_EPOCH

spytheman:
With this PR, this code:
```v

dump(@BUILD_DATE)
dump(@BUILD_TIME)
dump(@BUILD_TIMESTAMP)
```
will produce something like this (fixed at compile time), each time it is run:
```
[/v/misc/2024_09_13__12/a.v:2] @BUILD_DATE: 2024-09-13
[/v/misc/2024_09_13__12/a.v:3] @BUILD_TIME: 09:31:25
[/v/misc/2024_09_13__12/a.v:4] @BUILD_TIMESTAMP: 1726219885
```

... which eliminates one more common use of a separate build system.

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -5859,6 +5859,13 @@ A program that prints its own source code (a quine):
 print($embed_file(@FILE).to_string())
 ```
 
+A program that prints the time when it was built:
+```v
+import time
+
+println('This program, was compiled at ${time.unix(@BUILD_TIMESTAMP.i64()).format_ss_milli()} .')
+```
+
 > [!NOTE]
 > you can have arbitrary source code in the file, without problems, since the full file
 > will be embedded into the executable, produced by compiling it. Also note that printing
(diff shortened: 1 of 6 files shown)


larpon (approved):

koplenov:
I think we still need a git commit hash

koplenov:
Hmm, I did it this way

```v
@["/utils/get/server/version"]
pub fn (mut server Server) utils_get_server_version(mut ctx Context) veb.Result {
	return ctx.json[map[string]string](
		{
			"commit": $embed_file('.git/refs/heads/master', .zlib).to_string()
			"build_date": @BUILD_DATE
			"build_time": @BUILD_TIME
			"build_timestamp": @BUILD_TIMESTAMP
		}
	)
}
```

PR #22215, opened 2024-09-13
vlib.x: add a go like benchmark, that estimates automatically how many iterations are needed, to get a statistically significant result

skeris:
For my task, I need a benchmark that measures memory consumption. I checked the default benchmark library and didn't find that feature. So I decided to write it myself.

As advised in the Telegram chat, "just rewrite the golang library to vlang".

I checked the golang benchmark code and decided it was strange and overly complicated due to its use as a command-line tool.

So I decided to create my own library.

Basically, I copied the memory measurement logic and the prediction of the number of iterations from golang to estimate the expected benchmark evaluation time. And from vlang, I got the logic of using the benchmark as a library, not as a command-line tool.

How to use it:

- Create a Benchmark using the constructor, provide a closure to benchmark, optionally provide the expected number of iterations or estimation time.
- Run the run_benchmark method.
- Wait until it ends.
- Use BenchmarkResult as you need - by default, it prints, but you can send it to a chat or to a database.

But I'm not sure if I'm using the gc_heap_usage fields correctly. I was confused by the boehm_gc documentation, it's too hard for me. So if anyone can help me with that - you are welcome

### DIFF
--- a/vlib/x/benchmark/benchmark.v
+++ b/vlib/x/benchmark/benchmark.v
@@ -9,18 +9,18 @@ const default_duration = time.second
 // Benchmark represent all significant data for benchmarking. Provide clear way for getting result in convinient way by exported methods
 @[noinit]
 pub struct Benchmark {
-	pub mut:
-	n 								i64 						// Number of iterations. Set explicitly or computed from expected time of benchmarking
-	bench_func 				fn()! @[required] // function for benchmarking
-	bench_time 				time.Duration 	// benchmark duration
-	is_parallel 			bool 						// if true every bench_func run in separate coroutine
-	benchmark_result 	BenchmarkResult // accumulator of benchmark metrics
-	timer_on          bool						// inner flag of time recording
-	start_time    		time.Time				// start timestamp of timer
-	duration      		time.Duration		// expected time of benchmark process
-	failed 						bool						// flag of bench_func failure. true if one of bench_func run failed
-	start_memory 			usize						// memory status on start benchmark
-	start_allocs 			usize						// size of object allocated on heap
+pub mut:
+	n                i64 // Number of iterations. Set explicitly or computed from expected time of benchmarking
+	bench_func       fn () ! @[required] // function for benchmarking
+	bench_time       time.Duration   // benchmark duration
+	is_parallel      bool            // if true every bench_func run in separate coroutine
+	benchmark_result BenchmarkResult // accumulator of benchmark metrics
+	timer_on         bool            // inner flag of time recording
+	start_time       time.Time       // start timestamp of timer
+	duration         time.Duration   // expected time of benchmark process
+	failed           bool            // flag of bench_func failure. true if one of bench_func run failed
+	start_memory     usize           // memory status on start benchmark
+	start_allocs     usize           // size of object allocated on heap
 }
 
 // Benchmark.new - constructor of benchmark
@@ -29,8 +29,8 @@ pub struct Benchmark {
 // - n - number of iterations. set if you know how many runs of function you need. if qou don't know how many you need - set 0
 // - duration - by default 1s. expecting duration of all benchmark runs. doesn't work if is_parallel == true
 // - is_parallel - if true, every bench_func run in separate coroutine
-pub fn Benchmark.new(bench_func fn()!, n i64, duration time.Duration, is_parallel bool) !Benchmark{
-	if bench_func == voidptr(0) {
+pub fn Benchmark.new(bench_func fn () !, n i64, duration time.Duration, is_parallel bool) !Benchmark {
+	if bench_func == unsafe { nil } {
 		return error('Benchmark function cannot be empty')
 	}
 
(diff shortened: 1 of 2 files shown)


Delta456:
There's already a benchmark library. How different is it from that library?

JalonSolov:
The current benchmark library only measures speed, not memory.

spytheman:
There are other differences too - the existing `benchmark` does not auto calibrate the number of iterations, till a given time passes, it relies on your code to run the loop and iterate manually.

vlib/x/benchmark/benchmark.v
```diff
@@ -0,0 +1,246 @@
...
+	start_memory     usize           // memory status on start benchmark
+	start_allocs     usize           // size of object allocated on heap
+}
+
+// Benchmark.new - constructor of benchmark
+// arguments:
+// - bench_func - function to benchmark. required, if you have no function - you don't need benchmark
+// - n - number of iterations. set if you know how many runs of function you need. if qou don't know how many you need - set 0
+// - duration - by default 1s. expecting duration of all benchmark runs. doesn't work if is_parallel == true
+// - is_parallel - if true, every bench_func run in separate coroutine
```
spytheman, line 31: The UX can be improved by using a separate parameters struct, for the optional parameters, that can describe the desired defaults. Something like this:
```v
@[params]
pub struct BenchmarkDefaults {
pub:
    duration time.Duration = time.second
    is_parallel bool
    n i64
}

pub fn setup(bench_func fn () !,  params BenchmarkDefaults) !Benchmark {
```
which then later can be called with just:
```v
b := benchmark.setup( fn () ! { println(2+2) } )
```

vlib/x/benchmark/benchmark.v
```diff
@@ -0,0 +1,246 @@
...
+		n:           n
+		bench_func:  bench_func
+		bench_time:  if duration > 0 { duration } else { default_duration }
+		is_parallel: is_parallel
+	}
+}
+
+// run_benchmark - function for start benchmarking
+// run benchmark n times, or duration time
+pub fn (mut b Benchmark) run_benchmark() {
```
spytheman, line 51:
```suggestion
pub fn (mut b Benchmark) run() {
```

vlib/x/benchmark/benchmark.v
spytheman: or this, which is a bit shorter, but still repeats the benchmark word 3 times: 
```v
import x.benchmark { Benchmark }
b := Benchmark.new(...)
```

vlib/x/benchmark/benchmark.v
```diff
@@ -0,0 +1,246 @@
...
+		b.run_n(b.n - 1)
+	}
+
+	// if n is zero then we should run bench_func enough time for estimate duration time of execution
+	if b.n == 0 {
+		b.n = 1
+		// if one of runs failed - bench_func is not valid
+		// but 1e9 times of evaluation is too much
+		// so we need to repeat prediction-execition process while elapsed time less then expected time
+		for !b.failed && b.duration < b.bench_time && b.n < 1000000000 {
```
spytheman, line 73: Why `1_000_000_000` ?

vlib/x/benchmark/benchmark.v
skeris: I agree. In  documentation mentioned that we should use static methods instead of factory method `new_<something>()` (https://docs.vlang.io/structs.html#static-type-methods). So I thought, that it is new preferable way. But to be clear factory function is more convenient for me. Should I rewrite to factory function?

spytheman: Another benefit of the parameters struct, is that the callsites will have the names of the fields, which is useful to distinguish between the 2 numbers (the desired time.Duration which is an alias to i64, and the number of iterations, which is i64).

vlib/x/benchmark/benchmark.v
skeris: let me answer by quote of Alex
```
just go to go sources and rewrite them to vlang
```
idk, why go team took this number. imho in golang sources too much magic numbers.

vlib/x/benchmark/benchmark.v
spytheman: In my personal opinion, the convenience/usage for such a useful module, should be as clear and short as possible.

spytheman: Static methods are good when you have several different types in the same module, so that you can distinguish between them. In this case, there is only one public type in the module (`BenchmarkResult` is private), so the namespace benefit of static methods is moot, while the verbosity in the callsites would be big.

skeris: thnx, I'll fix this soon. may be at night

vlib/x/benchmark/benchmark.v
spytheman: It is as good as any big number I guess.

spytheman: If it works for them, it should work for us 🤷🏻‍♂️ .

vlib/x/benchmark/benchmark.v
```diff
@@ -0,0 +1,246 @@
...
+
+// Benchmark.new - constructor of benchmark
+// arguments:
+// - bench_func - function to benchmark. required, if you have no function - you don't need benchmark
+// - n - number of iterations. set if you know how many runs of function you need. if qou don't know how many you need - set 0
+// - duration - by default 1s. expecting duration of all benchmark runs. doesn't work if is_parallel == true
+// - is_parallel - if true, every bench_func run in separate coroutine
+pub fn Benchmark.new(bench_func fn () !, n i64, duration time.Duration, is_parallel bool) !Benchmark {
+	if bench_func == unsafe { nil } {
+		return error('Benchmark function cannot be empty')
```
Delta456, line 34:
```suggestion
		return error('Benchmark function cannot be `nil`')
```

JalonSolov: How about
```suggestion
		return error('valid benchmark function must be supplied.')
```

Delta456: Even better :smile:

PR #22216, opened 2024-09-13
crypto/scrypt: add scrypt to crypto

kimshrier:
This is what prompted me to do the other work in the other crypto modules.

This adds scrypt to the library of key derivation functions.  I have only 1 scrypt function test in scrypt_test.v because the other tests from RFC-7914 take a long time, 3 minutes on my machine.  The other tests in scrypt_test.v are exercising internal functionality that is not public.  I will put these slower scrypt function tests in the slower_tests repo after this gets committed (assuming that it does get accepted, of course).

The fact that scrypt is time and/or memory intensive is not a bug.  It is for the purpose of making scrypt generated keys more resistant to brute force attacks.


spytheman (approved):
Excellent work.

vlib/crypto/scrypt/scrypt.v
```diff
@@ -0,0 +1,243 @@
...
+import encoding.binary
+import math.bits
+
+pub const max_buffer_length = ((u64(1) << 32) - 1) * 32
+pub const max_blocksize_parallal_product = u64(1 << 30)
+
+// salsa20_8 applies the salsa20/8 core transformation to a block
+// of 64 u8 bytes.  The block is modified in place.
+fn salsa20_8(mut block []u8) {
+	mut block_words := []u32{len: 16, cap: 16}
```
spytheman, line 20: This can be just:
```suggestion
	mut block_words := []u32{len: 16}
```

vlib/crypto/scrypt/scrypt.v
```diff
@@ -0,0 +1,243 @@
...
+	mut scratch := []u32{len: 16, cap: 16}
```
spytheman, line 21: Can this be a fixed array? I did not saw usages that will grow/shrink it:
```suggestion
	mut scratch := [16]u32{}
```

vlib/crypto/scrypt/scrypt.v
```diff
@@ -0,0 +1,243 @@
...
+		scratch[15] ^= bits.rotate_left_32(scratch[14] + scratch[13], 18)
+	}
+
+	for i in 0 .. 16 {
+		scratch[i] += block_words[i]
+		binary.little_endian_put_u32_at(mut block, scratch[i], i * 4)
+	}
+
+	return
+}
```
spytheman, line 78: why?

vlib/crypto/scrypt/scrypt.v
kimshrier: I'll make that change.

vlib/crypto/scrypt/scrypt.v
kimshrier: You are right.  We can use a fixed array here.

vlib/crypto/scrypt/scrypt.v
kimshrier: I program in too many languages and I forget which one(s) don't require the final return statement.

I appreciate you taking the time to improve the code.

spytheman: Happy to help, when I can. Unfortunately when it comes to the crypto modules, I am a total noob on the substance, so I can only nitpick on the less significant language trivia.

PR #22217, opened 2024-09-14
make files: makes clone faster by depth=1

mengzhuo:
Clone with blob:none still fetch a lots of commit, we can set it to 1 make clone faster.


spytheman:
That was deliberately not done, because it creates more problems when doing `v up` later.

PR #22228, opened 2024-09-15
json: allow anon struct as decode type

### LINKED ISSUE #21703: Allows annonymous struct type to be passed as args
### Describe the feature

Allows passing anonymous struct type where a method parameter requires a type

### Use Case

For example the json decoding:
```v
json_text := '{ "a": "b" }'
b := json.decode(struct {
	a string
}, json_text)!.a
```
there's no need to create a named struct type if the decoding of this type only exists here

### Proposed Solution

_No response_

### Other Information

https://discord.com/channels/592103645835821068/700746775962714232/1252377577083895919

### Acknowledgements

### Version used

V 0.4.6 58f4e12

V full version: V 0.4.6 30d6f7b.58f4e12
OS: windows, Microsoft Windows 11 רҵ�� v22631 64 λ
Processor: 12 cpus, 64bit, little endian, 

getwd: xxx
vexe: E:\Programs\v\v.exe
vexe mtime: 2024-06-19 12:43:43

vroot: OK, value: E:\Programs\v
VMODULES: OK, value: C:\Users\xxx\.vmodules
VTMP: OK, value: C:\Users\xxx\AppData\Local\Temp\v_0

Git version: git version 2.38.1.windows.1
Git vroot status: weekly.2023.24-2387-g58f4e126
.git/config present: true

> [!NOTE]
>

felipensp:
Fix #21703

```v
import json

fn main() {
	json_text := '{ "a": "b" }'
	b := json.decode(struct {
		a string
	}, json_text)!.a
	assert dump(b) == 'b'
}
```

### DIFF
--- a/vlib/json/json_decode_anon_struct_test.v
+++ b/vlib/json/json_decode_anon_struct_test.v
@@ -1,0 +1,9 @@
+import json
+
+fn test_main() {
+	json_text := '{ "a": "b" }'
+	b := json.decode(struct {
+		a string
+	}, json_text)!.a
+	assert dump(b) == 'b'
+}
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -130,6 +130,7 @@ pub:
 	pos token.Pos
 pub mut:
 	typ          Type
+	stmt         Stmt = empty_stmt // for anon struct
 	end_comments []Comment // comments that after current type node
 }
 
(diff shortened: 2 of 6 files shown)


JalonSolov:
Really should be improving x.json2 so json can be retired...

Delta456:
It is better to support passing anonymous structs as parameters in general not just `json`.

JalonSolov:
If they're going to be supported at all, then yes, they should be supported in general.

felipensp:
But it's about anon struct declaration, not initialization.

spytheman (approved):
Excellent work.

PR #22229, opened 2024-09-15
feat: traverse method args in comptime

### LINKED ISSUE #20926: add .args to compile time method
### Describe the feature

I would like to do:
```v
$for arg in method.args {

}
```

### Use Case

For vgdextension and working with godot, we need a way for godot to call our methods. In order to avoid code generation, we need to be able to loop through args at compile time. We need to build structures defining our methods and create closures that are called by godot that transform the arguments and properly call our methods that are defined in V.

### Proposed Solution

add .args to compile time method

### Other Information

_No response_

### Acknowledgements

### Version used

V 0.4.4 70c277d

V full version: V 0.4.4 1733dab.70c277d

getwd: C:\Users\jcwea\Documents\git\jvnet
vexe mtime: 2024-02-29 15:47:47

Git vroot status: weekly.2024.09-6-g70c277d4
.git/config present: true

> [!NOTE]
>

felipensp:
Fix #20926

```v
fn register[T]() []string {
	mut args := []string{}
	$for method in T.methods {
		$for arg in method.args {
			$if arg.typ is f32 {
				args << 'f32: ${arg.name} ${typeof(arg.typ).name}'
			} $else $if arg.typ is voidptr {
				args << '&void: ${arg.name} ${typeof(arg.typ).name}'
			} $else {
			}
		}
	}
	return args
}
```

### DIFF
--- a/vlib/builtin/builtin.v
+++ b/vlib/builtin/builtin.v
@@ -90,8 +90,8 @@ fn __print_assert_failure(i &VAssertMetaInfo) {
 	}
 }
 
-// MethodArgs holds type information for function and/or method arguments.
-pub struct MethodArgs {
+// MethodParam holds type information for function and/or method arguments.
+pub struct MethodParam {
 pub:
 	typ  int
 	name string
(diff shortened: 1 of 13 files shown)


medvednikov:
There's another args => params name change needed in comptime.

spytheman (approved):

PR #22230, opened 2024-09-15
checker: check duplicate interfaces when implements for `struct`

Delta456:
This PR disallows duplicate interfaces when structs have explicit `implements` for interfaces.


Krotki:
What happens if you use `walker.Visitor` on top of tripe `Visitor` types in test ? Will recognize it as a same type ?

Delta456:
> What happens if you use `walker.Visitor` on top of tripe `Visitor` types in test ? Will recognize it as a same type ?

It does recognise the same type. I have formatted the file so it truncated the explicit import symbol.

vlib/v/checker/tests/struct_implements_interface_more_than_once_err.vv
```diff
@@ -0,0 +1,10 @@
+import v.ast.walker { Visitor }
+import v.ast { Node }
```
spytheman, line 2: Please use another interface from a module not from the `v.` ones.
`import hash { Hash }` for example, is much lighter to import, and will serve the same exact purpose.

spytheman: Or `Hash32er`, since it is smaller (it has just one method).

spytheman: Ok, `import io { Writer }` seems more suitable.
It has just one method, and it is public.

spytheman: Something like this for example checks for a lot less time:
```
CPU: 98% 0.11s	Real: 0.12s	Elapsed: 0:00.12	RAM: 34592KB	v vlib/v/checker/tests/struct_implements_interface_more_than_once_err.vv
```

compared to the version here:
```
CPU: 86% 0.31s	Real: 0.40s	Elapsed: 0:00.40	RAM: 91568KB	v vlib/v/checker/tests/struct_implements_interface_more_than_once_err.vv
```

spytheman: image

the underline for the `io.Writer` is wrong though

spytheman (approved):

PR #22244, opened 2024-09-18
strconv: fix format_fl()/format_es() (fix #13210)

### LINKED ISSUE #13210: Interpolating float into string with too large of a width returns unexpected result
**V version:** 0.2.4 d826317
**OS:** Linux

**What did you do?**
```
n := 0.1
println('${n:36.2f}')
println('${n:37.2f}')
println('${n:38.2f}')
```

```
                               0.10
                                0.10
                                 0.10
                                  0.10
```

**What did you see instead?**

```
                               0.10
                                 .10
                                   10
                                     0
```

After a certain point, the padding character starts to overwrite the formatted float value and corrupt the expected output. I can't reproduce this with integers, just floats.
I think this is likely related to the [hardcoded](https://github.com/vlang/v/blob/7a2705d8ce756dff19f080216311b946c95f5567/vlib/strconv/format_mem.c.v#L318) length 32 output buffer in `format_fl`; the tipping point where the bug starts to appear is at `32 + '0.10'.len + 1`.

yuyi98:
This PR fix format_fl()/format_es() (fix #13210).

- Fix format_fl()/format_es().
- Add test.

```v
fn main() {
	n := 0.1
	println('${n:36.2f}')
	println('${n:37.2f}')
	println('${n:38.2f}')
	println('${n:70.3f}')
}

PS D:\Test\v\tt1> v run .
                                0.10
                                 0.10
                                  0.10
                                                                 0.100
```


vlib/strconv/format_mem.c.v
```diff
@@ -328,8 +326,8 @@ pub fn format_fl(f f64, p BF_param) string {
 			tmp.free()
 		}
 
-		mut buf := [max_size_f64_char]u8{} // write temp float buffer in stack
-		mut out := [max_size_f64_char]u8{} // out buffer
+		mut buf := []u8{len: p.len0 + 32} // write temp float buffer
+		mut out := []u8{len: p.len0 + 32} // out buffer
```
spytheman, line 330: The new construct allocates memory that is not freed later.

Imho a better solution, would have been increasing max_size_f64_char to a bigger fixed size, for example 256 or 512 bytes.

Alternatively, `buf` and `out` will have to be freed manually after they are used, but that is harder to get right.

medvednikov: You're right.

yuyi98: Ok, I'll change it.

PR #22248, opened 2024-09-18
docs: improve format in the conditional compilation section

lcheylus:
In section "Conditional compilation / Environment specific files" (https://docs.vlang.io/conditional-compilation.html#environment-specific-files), improve Markdown format for filenames and output strings.


spytheman (approved):
Good work.

PR #22251, opened 2024-09-18
cgen: fix wrong type of vint_t and const riscv64

mengzhuo:
The size of vint_t for riscv64 should be 64.


spytheman (approved):
Good work.

Issue #21975, opened 2024-07-31
io.BufferedWriter


einar-hjortdal:
### Describe the feature

A buffered writer.
Just like a `BufferedReader` it should come with a `new_buffered_writer` factory function. Additionally, a `flush` method should be provided.

### Use Case

Many, stream ciphers for example.

### Other Information

I am surprised it isn't here yet.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.7 340535f

### Environment details (OS name and version, etc.)

V full version: V 0.4.7 c2f7afd.340535f
OS: linux, Linux version 5.14.0-427.26.1.el9_4.x86_64 (mockbuild@pp-el9) (gcc (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3), GNU ld version 2.35.2-43.el9) #1 SMP PREEMPT_DYNAMIC Wed Jul 17 15:28:16 CEST 2024
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz

getwd: /home/coachonko/Documents/projects/vlang/v
vexe: /home/coachonko/.local/lib64/v/v
vexe mtime: 2024-07-31 14:53:27

vroot: OK, value: /home/coachonko/.local/lib64/v
VMODULES: OK, value: /home/coachonko/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.43.5
Git vroot status: weekly.2024.31-13-g340535fe
.git/config present: true

CC version: cc (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3)
thirdparty/tcc status: thirdparty-linux-amd64 a0799a5b
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

spytheman:
Please do not close issues in advance, until a PR that addresses them is merged (after that they will get closed automatically, if they are mentioned in the PR description with `Fix #XYZ`).

einar-hjortdal:
> Please do not close issues in advance, until a PR that addresses them is merged (after that they will get closed automatically, if they are mentioned in the PR description with `Fix #XYZ`).

Alright. The previous one wasn't closed automatically after merging

PR #22265, opened 2024-09-20
io: add a BufferedWriter and supporting methods

einar-hjortdal:
Address #21975

### DIFF
--- a/vlib/io/buffered_writer.v
+++ b/vlib/io/buffered_writer.v
@@ -1,0 +1,78 @@
+module io
+
+pub struct BufferedWriter {
+mut:
+	n  int
+	wr Writer
+pub mut:
+	buf []u8
+}
+
+@[params]
+pub struct BufferedWriterConfig {
+pub:
+	writer Writer
+	cap    int = 128 * 1024
+}
+
+pub fn new_buffered_writer(o BufferedWriterConfig) !&BufferedWriter {
+	if o.cap < 1 {
+		return error('`o.cap` must be a positive integer')
+	}
+
+	return &BufferedWriter{
+		buf: []u8{len: o.cap, cap: o.cap}
+		wr:  o.writer
+	}
+}
+
+pub fn (mut b BufferedWriter) reset() {
+	cap := b.buf.len
+	b.buf = []u8{len: cap, cap: cap}
+	b.n = 0
+}
+
+pub fn (b BufferedWriter) buffered() int {
+	return b.n
+}
+
+pub fn (mut b BufferedWriter) flush() ! {
+	if b.buffered() == 0 {
+		return
+	}
+
+	n := b.wr.write(b.buf[0..b.n])!
+	if n < b.n {
+		return error('Writer accepted less bytes than expected without returning any explicit error.')
+	}
+
+	b.n = 0
+	return
+}
+
+pub fn (b BufferedWriter) available() int {
+	return b.buf.len - b.n
+}
+
+pub fn (mut b BufferedWriter) write(src []u8) !int {
+	mut p := src.clone()
+	len := p.len
+	mut nn := 0
+	for len > b.available() {
+		mut n := 0
+		if b.buffered() == 0 {
+			n = b.wr.write(p)!
+		} else {
+			n = copy(mut b.buf[b.n..], p)
+			b.n += n
+			b.flush()!
+		}
+		nn += n
+		p = p[n..].clone()
+	}
+
+	n := copy(mut b.buf[b.n..], p)
+	b.n += n
+	nn += n
+	return nn
+}
--- a/vlib/io/buffered_writer.v
+++ b/vlib/io/buffered_writer.v
@@ -21,14 +21,14 @@ pub fn new_buffered_writer(o BufferedWriterConfig) !&BufferedWriter {
 	}
 
 	return &BufferedWriter{
-		buf: []u8{len: o.cap, cap: o.cap}
+		buf: []u8{len: o.cap}
 		wr:  o.writer
 	}
 }
 
 pub fn (mut b BufferedWriter) reset() {
 	cap := b.buf.len
-	b.buf = []u8{len: cap, cap: cap}
+	b.buf = []u8{len: cap}
 	b.n = 0
 }
 
--- a/vlib/io/buffered_writer.v
+++ b/vlib/io/buffered_writer.v
@@ -66,9 +66,8 @@ pub fn (b BufferedWriter) available() int {
 // number of bytes written.
 pub fn (mut b BufferedWriter) write(src []u8) !int {
 	mut p := src.clone()
-	len := p.len
 	mut nn := 0
-	for len > b.available() {
+	for p.len > b.available() {
 		mut n := 0
 		if b.buffered() == 0 {
 			n = b.wr.write(p)!
--- a/vlib/io/buffered_writer.v
+++ b/vlib/io/buffered_writer.v
@@ -15,6 +15,8 @@ pub:
 	cap    int = 128 * 1024
 }
 
+// new_buffered_writer creates a new BufferedWriter with the specified BufferedWriterConfig.
+// Returns an error when cap is 0 or negative.
 pub fn new_buffered_writer(o BufferedWriterConfig) !&BufferedWriter {
 	if o.cap < 1 {
 		return error('`o.cap` must be a positive integer')
@@ -26,16 +28,21 @@ pub fn new_buffered_writer(o BufferedWriterConfig) !&BufferedWriter {
 	}
 }
 
+// reset resets the buffer to its initial state.
 pub fn (mut b BufferedWriter) reset() {
 	cap := b.buf.len
 	b.buf = []u8{len: cap}
 	b.n = 0
 }
 
+// buffered returns the number of bytes currently stored in the buffer.
 pub fn (b BufferedWriter) buffered() int {
 	return b.n
 }
 
+// flush writes the buffered data to the underlying writer and clears the buffer, ensures all data is
+// written.
+// Returns an error if the writer fails to write all buffered data.
 pub fn (mut b BufferedWriter) flush() ! {
 	if b.buffered() == 0 {
 		return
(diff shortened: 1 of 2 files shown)


vlib/io/buffered_writer_test.v
```diff
@@ -0,0 +1,60 @@
...
+pub mut:
+	result []u8
+}
+
+// implement io.Writer
+fn (mut aw ArrayWriter) write(buf []u8) !int {
+	len := buf.len
+	mut res := 0
+	for i := 0; i < len; i++ {
+		aw.result = arrays.concat(aw.result, buf[i])
```
spytheman, line 14:
```suggestion
		aw.result << buf[i]
```

vlib/io/buffered_writer_test.v
```diff
@@ -0,0 +1,60 @@
...
+		res++
+	}
+	return res
+}
+
+fn create_data(n int) []u8 {
+	mut res := []u8{}
+	for i := 0; i < n; i++ {
+		res = arrays.concat(res, 88) // x
```
spytheman, line 23:
```suggestion
		res << `x`
```

vlib/io/buffered_writer_test.v
```diff
@@ -0,0 +1,60 @@
...
+	}
+	res = arrays.concat(res, 10) // \n
```
spytheman, line 25:
```suggestion
	res << `\n`
```

vlib/io/buffered_writer.v
```diff
@@ -0,0 +1,78 @@
...
+	cap    int = 128 * 1024
+}
+
+pub fn new_buffered_writer(o BufferedWriterConfig) !&BufferedWriter {
+	if o.cap < 1 {
+		return error('`o.cap` must be a positive integer')
+	}
+
+	return &BufferedWriter{
+		buf: []u8{len: o.cap, cap: o.cap}
```
spytheman, line 24:
```suggestion
		buf: []u8{len: o.cap}
```
(setting cap: is implied by setting len:)

vlib/io/buffered_writer.v
```diff
@@ -0,0 +1,78 @@
...
+		wr:  o.writer
+	}
+}
+
+pub fn (mut b BufferedWriter) reset() {
+	cap := b.buf.len
+	b.buf = []u8{len: cap, cap: cap}
```
spytheman, line 31:
```suggestion
	b.buf.reset()
```

vlib/io/buffered_writer.v
```diff
@@ -0,0 +1,78 @@
...
+		mut n := 0
+		if b.buffered() == 0 {
+			n = b.wr.write(p)!
+		} else {
+			n = copy(mut b.buf[b.n..], p)
+			b.n += n
+			b.flush()!
+		}
+		nn += n
+		p = p[n..].clone()
```
spytheman, line 71: why are you cloning here?
`p` is already a clone of `src`; imho you can freely just slice it with
`p = unsafe{ p[n..] }`, since you do not modify the data in it all in the loop

spytheman: In fact the first clone at the start, is also not necessary imho, if `p` is going to just get sliced thinner and thinner later.

einar-hjortdal: I got the
```
Details: vlib/io/buffered_writer.v:71:8: details: To silence this notice, use either an explicit `a[..].clone()`,
or use an explicit `unsafe{ a[..] }`, if you do not want a copy of the slice.
   69 |         }
   70 |         nn += n
   71 |         p = p[n..]
      |              ~~~~~
   72 |     }
   73 |
```
I went for clone() because I try to avoid unsafe blocks.

spytheman: imho saving the memory allocations and the copying is a better goal, for a library module. Using `unsafe{}` is not a taboo (and V is not Rust). `unsafe{}` blocks just signal to both the developers and the compiler, that what is inside should be reviewed more carefully.

spytheman: It can be optimized out in a later PR though, if you are not comfortable with changing it. It is not a big deal.

vlib/io/buffered_writer.v
einar-hjortdal: I think this requires an unsafe block

einar-hjortdal:
oops. I messed up the rebase.

vlib/io/buffered_writer.v
```diff
@@ -0,0 +1,88 @@
...
+	return b.buf.len - b.n
+}
+
+// write writes `src` in the buffer, flushing it to the underlying writer as needed, and returns the
+// number of bytes written.
+pub fn (mut b BufferedWriter) write(src []u8) !int {
+	mut p := src.clone()
+	len := p.len
+	mut nn := 0
+	for len > b.available() {
```
spytheman, line 71: What will happen, if `len` is bigger than the buffer size?
Would not that loop turn into an infinite one?

spytheman: I think that the loop condition should use `p.len` instead of `len` 🤔 .

vlib/io/buffered_writer.v
```diff
@@ -0,0 +1,88 @@
...
+		mut n := 0
+		if b.buffered() == 0 {
+			n = b.wr.write(p)!
```
spytheman, line 74: That does not seem right ... if there is nothing buffered yet, it will just skip the buffer entirely, and write directly to the writer.
What is the point of the buffering then?

spytheman: I think that in general, buffering provides a way to decouple the speeds, at which producers and consumers of data work.

A buffered writer, should allow a fast producer of data, to fill a large buffer quickly and/or in bursts, and then drain that buffer, at whatever pace the consumer of the data can handle.

Thus in the write() method, if there is still space available, the data should go into the buffer, without concerning with `b.wr` at all, and *only* when the buffer is filled, it should be drained with a loop doing b.wr.write() calls.

vlib/io/buffered_writer.v
einar-hjortdal: You are correct. This will be fixed and a test will be added.

vlib/io/buffered_writer.v
einar-hjortdal: Ideally, the buffer size should be big enough to handle large writes, writes should be done manually using the flush method.
When the write is bigger than the buffer, I think writing directly would prevent needlessly copying data from source to buffer, and then give it to the underlying writer.

spytheman: The check is just `if b.buffered() == 0 {`, it does not take into account how large the block is at all. It will trigger for even small writes, negating the useful feature of 
the buffer *actually buffering*.

It is just buggy and has to be fixed.

spytheman: I think that a new module should focus on having correct and useful behavior first, and only then with optimizing edge cases, when there are suitable tests to guarantee that the optimizations will not break the correctness and usefulness.

einar-hjortdal: > The check is just `if b.buffered() == 0 {`, it does not take into account how large the block is at all. It will trigger for even small writes, negating the useful feature of the buffer _actually buffering_.
> 
> It is just buggy and has to be fixed.

The check is inside the `for p.len > b.available()` block, the loop checks if the source data is larger than the available space in the buffer.
Am I misunderstanding you?

spytheman: Please add a test, that does several small writes in a loop, with a total size < the buffer length. The buffered writer should not call its writer. Then call the flushing method, and it should write all the accumulated data from all small writes.

That test is the *basic functionality*, that a buffered writer should support, to justify its existence in the first place.

vlib/io/buffered_writer_test.v
```diff
@@ -1,5 +1,5 @@
-import arrays
 import io
+import crypto.rand
```
spytheman, line 2: Why do you use `crypto.rand` here?

There is no need to exhaust the entropy of the system in an unrelated test like that.
Use either `rand` (which uses a pseudo random generator, that will not block, or exhaust anything), or just string literals.

vlib/io/buffered_writer_test.v
```diff
@@ -31,6 +31,34 @@ fn test_create_buffered_writer() {
...
+fn test_basic_write() {
+	mut aw := ArrayWriter{}
+	max := 65536
+	mut bw := io.new_buffered_writer(writer: aw, cap: max)!
+
+	// write less data than buffer capacity, the underlying writer should receive no data.
+	safe_max := 65536 - 255
+	mut total := 0
+	for total < safe_max {
+		a := rand.read(1)!
```
spytheman, line 43: `rand.u8()` or a literal is a better choice

vlib/io/buffered_writer_test.v
einar-hjortdal: That was the first approach I could think of to solve this problem, I shall rewrite the test without crypto.rand

einar-hjortdal:
Test is failing sometimes, I think I wrote a bad utility function in the test and just discovered it.

spytheman:
The windows CI failures are unrelated and are already fixed on master.

spytheman (approved):
Excellent work.

I could not think about any other tests.

vlib/io/buffered_writer_test.v
```diff
@@ -0,0 +1,133 @@
+import io
+import rand
+
+struct ArrayWriter {
```
spytheman, line 4:
```suggestion
struct ArrayWriter implements Writer {
```

spytheman: although that could wait a bit more 🤔

vlib/io/buffered_writer.v
```diff
@@ -0,0 +1,87 @@
...
+// Returns an error if the writer fails to write all buffered data.
+pub fn (mut b BufferedWriter) flush() ! {
+	if b.buffered() == 0 {
+		return
+	}
+
+	n := b.wr.write(b.buf[0..b.n])!
+	if n < b.n {
+		return error('Writer accepted less bytes than expected without returning any explicit error.')
+	}
```
spytheman, line 54: we could perhaps loop here, till everything is flushed

spytheman: or there could be a separate flush_all method that does it 🤔

Issue #22206, opened 2024-09-12
random big.Integer function


einar-hjortdal:
### Describe the feature

Generate random `big.Integer` in the range [0, n)
`fn random_integer(n Integer) Integer`
Similar to `integer_from_bytes`, `integer_from_i64`, `integer_from_int`, etc.

### Use Case

Needed for Secure Remote Password and other crypto implementations.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

`V 0.4.7 79bedff`

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.7 ace4e93.79bedff
OS: linux, Linux version 5.14.0-427.33.1.el9_4.x86_64 (mockbuild@pp-el9) (gcc (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3), GNU ld version 2.35.2-43.el9) #1 SMP PREEMPT_DYNAMIC Thu Aug 29 12:00:49 CEST 2024
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz

getwd: /home/einar/Documents/projects/vlang/active/firebird
vexe: /home/einar/.local/lib64/v/v
vexe mtime: 2024-09-12 07:04:54

vroot: OK, value: /home/einar/.local/lib64/v
VMODULES: OK, value: /home/einar/.vmodules
VTMP: OK, value: /tmp/v_1001

Git version: git version 2.43.5
Git vroot status: weekly.2024.35-55-g79bedff1 (5 commit(s) behind V master)
.git/config present: true

CC version: cc (GCC) 11.4.1 20231218 (Red Hat 11.4.1-3)
thirdparty/tcc status: thirdparty-linux-amd64 0134e9b9
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

MarcusMWilliams:
👍

Coachonko:
@hungrybluedev you're the big.Integer man, right? :pray:

Coachonko:
```v
// new_random_big_integer creates a random `big.Integer` with range [0, n)
// panics if `n` is 0 or negative.
pub fn new_random_big_integer(n big.Integer) !big.Integer {
	if n.signum < 1 {
[...]

PR #22266, opened 2024-09-20
crypto: add random big integer generation

einar-hjortdal:
Addresses #22206

### DIFF
--- a/vlib/crypto/rand/utils.v
+++ b/vlib/crypto/rand/utils.v
@@ -5,6 +5,7 @@
 module rand
 
 import math.bits
+import math.big
 import encoding.binary
 
 // int_u64 returns a random unsigned 64-bit integer `u64` read from a real OS source of entropy.
@@ -52,4 +53,47 @@ fn bytes_to_u64(b []u8) []u64 {
 		z[z.len - 1] = d
 	}
 	return z
+}
+
+// int_big creates a random `big.Integer` with range [0, n)
+// panics if `n` is 0 or negative.
+pub fn int_big(n big.Integer) !big.Integer {
+	if n.signum < 1 {
+		return error('`n` cannot be 0 or negative.')
+	}
+
+	max := n - big.integer_from_int(1)
+	len := max.bit_len()
+
+	if len == 0 {
+		// max must be 0
+		return max
+	}
+
+	// k is the maximum byte length needed to encode a value < n
+	k := (len + 7) / 8
+
+	// b is the number of bits in the most significant byte of n-1
+	get_b := fn [len] () u64 {
+		b := u64(len % 8)
+		if b == 0 {
+			return 8
+		}
+		return b
+	}
+	b := get_b()
+
+	mut result := big.Integer{}
+	for found := false; found == false; {
+		mut bytes := read(k)!
+
+		// Clear bits in the first byte to increase the probability that the result is < max
+		bytes[0] &= u8(int(1 << b) - 1)
+
+		result = big.integer_from_bytes(bytes)
+		if result < max {
+			found = true
+		}
+	}
+	return result
 }
--- a/vlib/crypto/rand/utils.v
+++ b/vlib/crypto/rand/utils.v
@@ -56,7 +56,7 @@ fn bytes_to_u64(b []u8) []u64 {
 }
 
 // int_big creates a random `big.Integer` with range [0, n)
-// panics if `n` is 0 or negative.
+// returns an error if `n` is 0 or negative.
 pub fn int_big(n big.Integer) !big.Integer {
 	if n.signum < 1 {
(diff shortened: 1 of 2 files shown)


vlib/crypto/rand/utils.v
```diff
@@ -53,3 +54,46 @@ fn bytes_to_u64(b []u8) []u64 {
...
+
+	// b is the number of bits in the most significant byte of n-1
+	get_b := fn [len] () u64 {
+		b := u64(len % 8)
+		if b == 0 {
+			return 8
+		}
+		return b
+	}
+	b := get_b()
```
spytheman, line 84: why is this defining a closure, just to call it right away?
imho just inline the code

vlib/crypto/rand/utils.v
```diff
@@ -53,3 +54,46 @@ fn bytes_to_u64(b []u8) []u64 {
...
+
+	mut result := big.Integer{}
+	for found := false; found == false; {
```
spytheman, line 87:
```suggestion
	for {
```

vlib/crypto/rand/utils.v
```diff
@@ -53,3 +54,46 @@ fn bytes_to_u64(b []u8) []u64 {
...
+		mut bytes := read(k)!
+
+		// Clear bits in the first byte to increase the probability that the result is < max
+		bytes[0] &= u8(int(1 << b) - 1)
+
+		result = big.integer_from_bytes(bytes)
+		if result < max {
+			found = true
```
spytheman, line 95:
```suggestion
			break
```

vlib/crypto/rand/utils.v
```diff
@@ -53,3 +54,46 @@ fn bytes_to_u64(b []u8) []u64 {
...
+		return error('`n` cannot be 0 or negative.')
+	}
+
+	max := n - big.integer_from_int(1)
+	len := max.bit_len()
+
+	if len == 0 {
+		// max must be 0
+		return max
+	}
```
spytheman, line 71: That seems weird. Why return a definite number, from a function that is said to return a random one?

vlib/crypto/rand/utils.v
```diff
@@ -53,3 +54,46 @@ fn bytes_to_u64(b []u8) []u64 {
 	}
 	return z
 }
+
+// int_big creates a random `big.Integer` with range [0, n)
+// panics if `n` is 0 or negative.
```
spytheman, line 59: It does not panic. It returns an error.

vlib/crypto/rand/utils_test.v
```diff
@@ -0,0 +1,14 @@
+import math.big
+import crypto.rand
+
+fn test_int_big() {
+	z := big.integer_from_int(0)
+	rand.int_big(z) or { assert true }
```
spytheman, line 6: That would not fail in any case, error or no error.

You can use the full form instead, to assure that the test works and covers both:
`if _ := rand.int_big(z) { assert false } else { assert true }`

vlib/crypto/rand/utils_test.v
```diff
@@ -0,0 +1,14 @@
...
+
+	rand.int_big(big.integer_from_int(-1)) or { assert true }
```
spytheman, line 8: same - use the full form, with assertions in both branches

vlib/crypto/rand/utils_test.v
```diff
@@ -0,0 +1,14 @@
...
+
+	m := big.integer_from_int(1).left_shift(128)
+	l := rand.int_big(m)! // actual large number
+	assert l < m
+	assert l > z
```
spytheman, line 13: That could fail with a false positive, although the chance of that is very small.
The function is defined to return a number in `[0, n)`, i.e. 0 is an allowed output value.
Imho compare with -1 here

spytheman:
The vab failure is unrelated, and now fixed on master. 
The other CI failure is sporadic (flaky test), and not related to this PR.

vlib/crypto/rand/utils.v
einar-hjortdal: Done, I tend to define closures over using mutable variables. A mut should be more efficient.

vlib/crypto/rand/utils_test.v
einar-hjortdal: Thank you, I threw tests together quick and I am also no good at writing tests in V, yet :grimacing:

vlib/crypto/rand/utils.v
einar-hjortdal: I added a little more explanation in the comment.
If the user-given max is 1, the only valid integer in [0, 1) is 0. So 0 is returned early.
I think allowing users to give 1 as max is ok.

spytheman (approved):
Excellent work.

PR #22270, opened 2024-09-21
$tmpl: fix .${var}

### LINKED ISSUE #22231: v template small bug something.${name}, will not work (because of . in front)
### Describe the bug

v template small bug something.${name}, will not work (because of . in front)

### Reproduction Steps

code 

```
import freeflowuniverse.crystallib.installers.something.${args.name}
```

if args.name given it will still output

```
import freeflowuniverse.crystallib.installers.something.@{args.name}

```

### Expected Behavior

import freeflowuniverse.crystallib.installers.something.aname

### Current Behavior

see above

### Possible Solution

_No response_

_No response_

### V version

latest

osx

> [!NOTE]
>

esquerbatua:
Fixes #22231


spytheman:
Excellent work @esquerbatua .

PR #22271, opened 2024-09-21
cgen: fix codegen for alias to charptr passed as ptr

### LINKED ISSUE #9679: Type alias cannot be a reference
**V version:** 0.2.2
**OS:** Linux

In this example I have two types that are equal:
1. `&char`
1. `ALcharptr`

When translated into C, both types (although equal) will be passed to `char_vstring` differently:
1. `char_vstring(s)`
1. `char_vstring(&s)`

Full exampled [here](https://github.com/vlang/v/files/6292006/alias_bug.zip)(zip).

### V snippet
```v
type ALcharptr = &char
fn C.return_text_one() ALcharptr
fn C.return_text_two() &char

fn main() {
	println('type &char\talias ALcharptr\t:\t${get_one()}')
	println('type &char\talias none\t:\t${get_two()}')
}

fn get_one() string {
	s := C.return_text_one()
	return unsafe { cstring_to_vstring(s) }
}

fn get_two() string {
	s := C.return_text_two()
	return unsafe { cstring_to_vstring(s) }
}
```

### C snippet
```c
void main__main(void) {
	println(_STR("type &char\talias ALcharptr\t:\t%.*s", 1, main__get_one()));
	println(_STR("type &char\talias none\t:\t%.*s", 1, main__get_two()));
}

string main__get_one(void) {
	main__ALcharptr s = return_text_one();
	 string _t101 = char_vstring(&s); // <- Nein!
	return _t101;
}

string main__get_two(void) {
	char* s = return_text_two();
	 string _t102 = char_vstring(s);
	return _t102;
}
```

### Output
```
type &char      alias ALcharptr :       �lf
type &char      alias none      :       text
```

felipensp:
Fix #9679

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -2722,7 +2722,8 @@ fn (mut g Gen) ref_or_deref_arg(arg ast.CallArg, expected_type ast.Type, lang as
 						return
 					} else if arg.expr is ast.Ident && arg.expr.language == .c {
 						g.write('(voidptr)')
-					} else {
+					} else if !(!arg.is_mut && arg_sym.kind == .alias
+						&& g.table.unaliased_type(arg_typ).is_any_kind_of_pointer()) {
 						g.write('(voidptr)&/*qq*/')
 					}
 				} else {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #22279, opened 2024-09-22
net:websocket EINTR handling

ismyhc:
I am working on a TUI that connections to websocket servers. I am getting premature ws closures specifically when using net.websocket client in conjunction with term.ui. By checking for EINTR error code and continuing rather than returning ws.read_next_message() fails we avoid closing prematurely. EINTR calls should not cause ws closures in this case.

This change makes the websocket behave as expected.

I hope this helps others as for me it has fixed the problem.

### DIFF
--- a/vlib/net/websocket/websocket_client.v
+++ b/vlib/net/websocket/websocket_client.v
@@ -133,12 +133,15 @@ pub fn (mut ws Client) listen() ! {
 	}
 	for ws.get_state() == .open {
 		msg := ws.read_next_message() or {
-			if ws.get_state() in [.closed, .closing] {
+			if err.code() == C.EINTR { // Check for EINTR and retry
+				continue
+			} else if ws.get_state() in [.closed, .closing] {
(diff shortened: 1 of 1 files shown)


JalonSolov:
The only thing I see off the bat here is that you've "polluted" a pure V file with C code.

I would suggest either creating an EINTR const with the actual value, or create a .c.v file with the change.

The other possibility is to simply rename the current file, but that could be a breaking change if any of the other backends are using this code.

ismyhc:
> The only thing I see off the bat here is that you've "polluted" a pure V file with C code.
> 
> I would suggest either creating an EINTR const with the actual value, or create a .c.v file with the change.
> 
> The other possibility is to simply rename the current file, but that could be a breaking change if any of the other backends are using this code.

Okay, that makes sense. I've changed it to use the built in **net.error_eintr** const already established in net module.

spytheman (approved):
Good work.

PR #22280, opened 2024-09-22
checker, cgen: fix codegen for option fntype

### LINKED ISSUE #22278: Optional Function returns break everything
**V doctor:**
```
V full version: V 0.4.7 63957f2.3307340
OS: linux, Ubuntu 24.04.1 LTS

getwd: ~/vlang/option_bug_maybe
vexe: ~/vlang/v/v
vexe mtime: 2024-09-21 23:21:32

vroot: OK, value: ~/vlang/v
VMODULES: OK, value: ~/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.43.0
Git vroot status: weekly.2024.37-36-g33073407
.git/config present: true

CC version: cc (Ubuntu 13.2.0-23ubuntu4) 13.2.0

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg src/main.v`
```v
module main

type DataFn = fn (name string) string

fn which_lang(name string) string {
	return name
}

fn find_func(name string) ?DataFn {
	return match name {
		'vlang' { which_lang }
		else { none }
	}
}

fn main() {
	req := find_func('vlang')
	fun := req('options')
	println(fun)
}

```

a print of 'options'

**What did you see instead?**
```
cc: /tmp/v_1000/main.01J8BE9J6F2YCCRKS27WSRJ220.tmp.c:13346: error: lvalue expected
builder error: 
==================
```
> [!NOTE]
>

felipensp:
Fix #22278

### DIFF
--- a/vlib/v/gen/c/match.v
+++ b/vlib/v/gen/c/match.v
@@ -92,10 +92,10 @@ fn (mut g Gen) match_expr(node ast.MatchExpr) {
 		cur_line = g.go_before_last_stmt().trim_left(' \t')
 		tmp_var = g.new_tmp_var()
 		mut func_decl := ''
-		if g.table.final_sym(node.return_type).kind == .function {
-			func_sym := g.table.final_sym(node.return_type)
-			if func_sym.info is ast.FnType {
-				def := g.fn_var_signature(func_sym.info.func.return_type, func_sym.info.func.params.map(it.typ),
+		ret_final_sym := g.table.final_sym(node.return_type)
+		if !node.return_type.has_option_or_result() && ret_final_sym.kind == .function {
+			if ret_final_sym.info is ast.FnType {
+				def := g.fn_var_signature(ret_final_sym.info.func.return_type, ret_final_sym.info.func.params.map(it.typ),
 					tmp_var)
 				func_decl = '${def} = &${g.typ(node.return_type)};'
 			}
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -1197,6 +1197,18 @@ fn (mut c Checker) check_expr_option_or_result_call(expr ast.Expr, ret_type ast.
 					expr_ret_type = unaliased_ret_type
 				}
 			}
+			// var with option function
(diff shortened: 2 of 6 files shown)


StunxFS (approved):

vlib/v/gen/c/match.v
```diff
@@ -92,10 +92,10 @@ fn (mut g Gen) match_expr(node ast.MatchExpr) {
...
 		mut func_decl := ''
-		if g.table.final_sym(node.return_type).kind == .function {
-			func_sym := g.table.final_sym(node.return_type)
-			if func_sym.info is ast.FnType {
-				def := g.fn_var_signature(func_sym.info.func.return_type, func_sym.info.func.params.map(it.typ),
+		ret_final_sym := g.table.final_sym(node.return_type)
+		if !node.return_type.has_option_or_result() && ret_final_sym.kind == .function {
+			if ret_final_sym.info is ast.FnType {
+				def := g.fn_var_signature(ret_final_sym.info.func.return_type, ret_final_sym.info.func.params.map(it.typ),
 					tmp_var)
```
spytheman, line 99: Does it work with if expressions too?

spytheman: it does ... excellent

spytheman (approved):
Excellent work.

spytheman:
There is a small variation, that still produces a cgen error:
```v
module main

type DataFn = fn (name string) string

fn which_lang(name string) string {
	return name
}

fn find_func(name string) ?DataFn {
	a := if name == 'vlang' { ?DataFn(which_lang) } else { none }
	return a
}

fn test_main() {
	req := find_func('vlang')?
	fun := req('options')
	println(fun)
	assert fun == 'options'
}

fn test_ifguard() {
	if req2 := find_func('vlang') {
		assert req2('options') == 'options'
	}
}
```

spytheman:
@felipensp should I file a separate issue for it?

spytheman:
Another one is:
```v
fn find_func(name string) ?DataFn {
    a := ?DataFn( which_lang )
    return a
}
```
with the same setup

PR #22281, opened 2024-09-22
examples,os: add an os.asset module, use it to simplify code in examples/, by removing `$if android {` checks

spytheman:


larpon (approved):
Nice addition. I like how it simplifies code in userland :heart:

spytheman:
The CI failures are unrelated, and fixed on master.

examples/clock/clock.v
```diff
@@ -163,8 +156,6 @@ fn main() {
 		frame_fn:     on_frame
 		event_fn:     on_event
 		init_fn:      on_init
-		font_path:    font_path
```
StunxFS, line 166: Isn't there a missing `font_path: asset.get_path` here?

examples/fireworks/fireworks.v
```diff
@@ -110,8 +103,6 @@ fn main() {
 		user_data:    app
 		frame_fn:     on_frame
 		event_fn:     on_event
-		font_path:    font_path
```
StunxFS, line 113: And here too

StunxFS (commented):
Perfect! 💯

examples/clock/clock.v
spytheman: This example only animates figures, and does not draw any text. I think that having it let `gg` use its default font, would be clearer to the readers.

examples/fireworks/fireworks.v
spytheman: Same - the example shows fireworks, with no text involved. Removing the explicit setting of the font path, makes it clearer imho.

PR #10790, opened 2021-07-13
checker: prohibit using non-lvalue as mut receiver

shadowninja55:
this pr ensures that mut method receivers are lvalues. closes #10788.

PR #22286, opened 2024-09-22
checker: allow method chaining call on mutable var

### LINKED ISSUE #11123: [BUG] Error: expression evaluated but not used when declaring a struct
**V version:** V 0.2.2 1588fe7
**OS:** Linux; Arch Linux; 64 bit

**What did you do?**

Run the following program to test if chaining functions works:

```v
struct App {
	name string
	version string
mut:
	commands []Command
	flags []Flag
}

struct Command {
	name string
	execute fn (cmd Command)
}

struct Flag {
	name string
	abbrev string
	required bool
}

fn (mut app App) command(cmd Command) App {
	app.commands << cmd
	return app
}

fn (mut app App) flag(flag Flag) App {
	app.flags << flag
	return app
}

fn main() {
	mut app := mut App {
		name: 'amazing cli'
		version: '0.1.0'
		commands: []
		flags: []
	}.command(Command {
		name: 'test'
		execute: fn (cmd Command) {
			println('Testing')
		}
	}).flag(Flag {
		name: 'verbose'
		abbrev: 'v'
		required: true
	})

	println(app)
}
```

The program to run, and print out the `App` struct.
 
**What did you see instead?**

The following error:

```
chaining.v:32:9: error: expression evaluated but not used
   30 | fn main() {
   31 |     app := mut App {
   32 |         name: 'amazing cli'
      |               ~~~~~~~~~~~~~
   33 |         version: '0.1.0'
   34 |         commands: []
```

felipensp:
Fix #11123

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -3366,7 +3366,7 @@ fn (mut c Checker) array_builtin_method_call(mut node ast.CallExpr, left_type as
 
 fn (mut c Checker) check_for_mut_receiver(mut expr ast.Expr) (string, token.Pos) {
 	to_lock, pos := c.fail_if_immutable(mut expr)
-	if !expr.is_lvalue() {
+	if !expr.is_lvalue() && expr !is ast.CallExpr {
 		c.error('cannot pass expression as `mut`', expr.pos())
 	}
 	return to_lock, pos
--- a/vlib/v/tests/fns/method_call_chain_mut_test.v
+++ b/vlib/v/tests/fns/method_call_chain_mut_test.v
@@ -1,0 +1,54 @@
+struct App {
+	name    string
+	version string
+mut:
+	commands []Command
+	flags    []Flag
+}
+
+struct Command {
+	name    string
+	execute fn (cmd Command) = unsafe { nil }
+}
+
+struct Flag {
+	name     string
+	abbrev   string
+	required bool
+}
+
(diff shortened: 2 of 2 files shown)


vlib/v/checker/fn.v
```diff
@@ -3366,7 +3366,7 @@ fn (mut c Checker) array_builtin_method_call(mut node ast.CallExpr, left_type as
 
 fn (mut c Checker) check_for_mut_receiver(mut expr ast.Expr) (string, token.Pos) {
 	to_lock, pos := c.fail_if_immutable(mut expr)
-	if !expr.is_lvalue() {
+	if !expr.is_lvalue() && expr !is ast.CallExpr {
 		c.error('cannot pass expression as `mut`', expr.pos())
```
spytheman, line 3370: No, that error is deliberate. It should be changed only when we have something like:
 `fn f() mut &Type {`, so that it is explicitly known that `f` was designed to return a mutable reference.

spytheman: See https://github.com/vlang/v/pull/10790 and search for it in Discord to see the previous discussions about it.

felipensp: Okay. Got it.

medvednikov: I wonder if it's as simple as adding `mut` to return value: `fn f() mut &Type {`

If so, I can do this quickly to allow method chaining.

PR #22302, opened 2024-09-24
Build docker images action

Krotki:
Github action to checkout vlang/docker repo and build docker images for amd64 and arm64 platforms.

Manual trigger for now - can be adjusted latter.


.github/workflows/hub_docker_ci.yml
```diff
@@ -0,0 +1,81 @@
+name: hub_docker_ci
+
+on:
+  workflow_dispatch:
+
+jobs:
+  example_matrix:
```
spytheman, line 7:
```suggestion
  publish-new-docker-images:
```

PR #22303, opened 2024-09-24
tmpl: add escape for `@` using `@@`

### LINKED ISSUE #22299: Templates don't accept @@ anymore
**V doctor:**
```
V full version: V 0.4.7 8a0fc25.4ee948f
OS: linux, Linux Mint 21.3
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz

getwd: /home/friedrich/Dev/Tests/v_tmpl_test
vexe: /home/friedrich/Tools/v/v
vexe mtime: 2024-09-24 18:46:16

vroot: OK, value: /home/friedrich/Tools/v
VMODULES: OK, value: /home/friedrich/.vmodules
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.39-5-g4ee948fe
.git/config present: true

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg tmpl_tst.v`
```v
module main

fn main() {
	$tmpl("tmpl.txt")
}
```

tmpl.txt:
```
Hello @@world!
```

Nothing, it should just compile, because @@ should escape an @.

**What did you see instead?**
```
tmpl.txt:9:40: error: undefined ident: `world` (veb action: main__main)
tmpl.txt:9:40: error: expression does not return a value (veb action: main__main)
```
Instead, the @@ is ignored.

felipensp:
Fix #22299

### DIFF
--- a/vlib/v/parser/tmpl.v
+++ b/vlib/v/parser/tmpl.v
@@ -77,7 +77,7 @@ fn insert_template_code(fn_name string, tmpl_str_start string, line string) stri
 	// HTML, may include `@var`
 	// escaped by cgen, unless it's a `vweb.RawHtml` string
 	trailing_bs := tmpl_str_end + 'sb_${fn_name}.write_u8(92)\n' + tmpl_str_start
-	replace_pairs := ['\\', '\\\\', r"'", "\\'", r'@', r'$', r'$$', r'\@']
+	replace_pairs := ['\\', '\\\\', r"'", "\\'", r'@@', r'@', r'@', r'$', r'$$', r'\@']
 	mut rline := line.replace_each(replace_pairs)
 	comptime_call_str := rline.find_between('\${', '}')
 	if comptime_call_str.contains("\\'") {
--- a/vlib/v/tests/tmpl/escape.txt
+++ b/vlib/v/tests/tmpl/escape.txt
@@ -1,0 +1,3 @@
+Hello @@world
+Hello @world
+Hello @ @world @
--- a/vlib/v/tests/tmpl_escape_test.v
+++ b/vlib/v/tests/tmpl_escape_test.v
@@ -1,0 +1,9 @@
+fn test_escape() {
+	world := 'world!'
+	res := $tmpl('tmpl/escape.txt')
+
+	assert res == 'Hello @world
+Hello world!
+Hello $ world! $
+'
+}


vlib/v/parser/tmpl.v
```diff
@@ -77,7 +77,7 @@ fn insert_template_code(fn_name string, tmpl_str_start string, line string) stri
 	// HTML, may include `@var`
 	// escaped by cgen, unless it's a `vweb.RawHtml` string
 	trailing_bs := tmpl_str_end + 'sb_${fn_name}.write_u8(92)\n' + tmpl_str_start
-	replace_pairs := ['\\', '\\\\', r"'", "\\'", r'@', r'$', r'$$', r'\@']
+	replace_pairs := ['\\', '\\\\', r"'", "\\'", r'@@', r'@', r'@', r'$', r'$$', r'\@']
```
JalonSolov, line 80: I'm confused... it will now turn @@ into @, @ into $, but $$ into \@?

It was always supposed to have @@ into @, but why $$ into \@?  Especially since it's a raw string so will literally be `\` followed by `@`.

JalonSolov: Since it would have to start from the beginning of the string for each pair, it will also wind up turning `@@$` into `@$`, then `@$` into `$$`, then `$$` into `\@`.

Sounds like far too many unintentional changes could be made.

felipensp: > I'm confused... it will now turn `@@` into `@`, `@` into `$`, but `$$` into `\@`?
> 
> It was always supposed to have `@@` into `@`, but why `$$` into `\@`? Especially since it's a raw string so will literally be `\` followed by `@`.

Hehe It is weird to me too. No idea if this was really intended to work as is. My chance here is only about escaping @@.

spytheman: It is not, since there is not a real parser with defined tokens, grammar, etc, but an adhoc one :-|, that gets more complex and is full of edge cases.

At least now, there are more tests for them, so it will not fail silently, and we have a chance to review the more insane behavior, and potentially clean it up consciously.

spytheman: (well, I can hope)

spytheman (approved):
Good work, especially the test, so that it will not regress silently anymore, like it did already before.

The CI failure is unrelated and fixed on master.

PR #22308, opened 2024-09-25
checker: add missing check for duplicated items on in expr

### LINKED ISSUE #22305: missing checker notice/warning for a known duplicate value in a `if x in [val1, val1, val2] {`
**V doctor:**
```
V full version: V 0.4.7 4ee948f.1a4ad55
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-09-25 08:03:20

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.39-12-g1a4ad556
.git/config present: true

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg /v/misc/2024_09_25__11/a.v`
```v
if 1 in [1, 2, 2, 2] { println('ok') }
println('done')

```

A compiler notice or warning, that 2 is a duplicate value in the condition list, used for the `x in list` check .

**What did you see instead?**
```
ok
done
```
> [!NOTE]
>

felipensp:
Fix #22305

### DIFF
--- a/cmd/tools/vdoc/highlight.v
+++ b/cmd/tools/vdoc/highlight.v
@@ -83,7 +83,7 @@ fn color_highlight(code string, tb &ast.Table) string {
 						&& (next_tok.kind != .lpar || prev.kind !in [.key_fn, .rpar]) {
 						tok_typ = .builtin
 					} else if
-						(next_tok.kind in [.lcbr, .rpar, .eof, .comma, .pipe, .name, .rcbr, .assign, .key_pub, .key_mut, .pipe, .comma, .comment, .lt, .lsbr]
+						(next_tok.kind in [.lcbr, .rpar, .eof, .name, .rcbr, .assign, .key_pub, .key_mut, .pipe, .comma, .comment, .lt, .lsbr]
 						&& next_tok.lit !in highlight_builtin_types)
 						&& (prev.kind in [.name, .amp, .lcbr, .rsbr, .key_type, .assign, .dot, .not, .question, .rpar, .key_struct, .key_enum, .pipe, .key_interface, .comment, .ellipsis, .comma]
 						&& prev.lit !in highlight_builtin_types)
--- a/vlib/v/checker/infix.v
+++ b/vlib/v/checker/infix.v
@@ -964,7 +964,7 @@ fn (mut c Checker) check_duplicated_items(node &ast.ArrayInit) {
 		if item.str() !in unique_items {
 			unique_items << item_str
 		} else {
-			c.error('item `${item_str}` is duplicated in the list', item.pos())
+			c.note('item `${item_str}` is duplicated in the list', item.pos())
 		}
 	}
 }
--- a/vlib/v/checker/infix.v
+++ b/vlib/v/checker/infix.v
@@ -957,14 +957,13 @@ fn (mut c Checker) check_div_mod_by_zero(expr ast.Expr, op_kind token.Kind) {
 }
 
 fn (mut c Checker) check_duplicated_items(node &ast.ArrayInit) {
-	mut unique_items := []string{}
+	mut unique_items := []string{cap: node.exprs.len}
 	for item in node.exprs {
 		item_str := item.str()
-		// println('>>> ${item_str}')
-		if item.str() !in unique_items {
-			unique_items << item_str
-		} else {
+		if item_str in unique_items {
 			c.note('item `${item_str}` is duplicated in the list', item.pos())
+		} else {
+			unique_items << item_str
 		}
 	}
 }
--- a/vlib/v/checker/infix.v
+++ b/vlib/v/checker/infix.v
@@ -241,6 +241,9 @@ fn (mut c Checker) infix_expr(mut node ast.InfixExpr) ast.Type {
 							c.error('left operand to `${node.op}` does not match the array element type: ${err.msg()}',
 								left_right_pos)
 						}
+						if mut node.right is ast.ArrayInit {
+							c.check_duplicated_items(node.right)
+						}
 					} else {
 						if mut node.right is ast.ArrayInit {
 							for i, typ in node.right.expr_types {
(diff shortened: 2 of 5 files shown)


vlib/v/checker/infix.v
```diff
@@ -953,6 +956,19 @@ fn (mut c Checker) check_div_mod_by_zero(expr ast.Expr, op_kind token.Kind) {
...
 
+fn (mut c Checker) check_duplicated_items(node &ast.ArrayInit) {
+	mut unique_items := []string{}
+	for item in node.exprs {
+		item_str := item.str()
+		// println('>>> ${item_str}')
+		if item.str() !in unique_items {
+			unique_items << item_str
+		} else {
+			c.error('the item `${item_str}` is duplicated in the list', item.pos())
```
StunxFS, line 967:
```suggestion
			c.error('item `${item_str}` is duplicated in the list', item.pos())
```

I think it would look better without "the"

esquerbatua:
Great job!

vlib/v/checker/str.v
```diff
@@ -96,8 +96,7 @@ fn (mut c Checker) string_inter_lit(mut node ast.StringInterLiteral) ast.Type {
 			}
 			if ((typ.is_unsigned() && fmt !in [`u`, `x`, `X`, `o`, `c`, `b`])
 				|| (typ.is_signed() && fmt !in [`d`, `x`, `X`, `o`, `c`, `b`])
-				|| (typ.is_int_literal()
-				&& fmt !in [`d`, `c`, `x`, `X`, `o`, `u`, `x`, `X`, `o`, `b`])
+				|| (typ.is_int_literal() && fmt !in [`d`, `c`, `x`, `X`, `o`, `u`, `b`])
```
spytheman, line 99: Very nice, it is immediately useful.

cmd/tools/vdoc/highlight.v
```diff
@@ -83,7 +83,7 @@ fn color_highlight(code string, tb &ast.Table) string {
 						&& (next_tok.kind != .lpar || prev.kind !in [.key_fn, .rpar]) {
 						tok_typ = .builtin
 					} else if
-						(next_tok.kind in [.lcbr, .rpar, .eof, .comma, .pipe, .name, .rcbr, .assign, .key_pub, .key_mut, .pipe, .comma, .comment, .lt, .lsbr]
+						(next_tok.kind in [.lcbr, .rpar, .eof, .name, .rcbr, .assign, .key_pub, .key_mut, .pipe, .comma, .comment, .lt, .lsbr]
```
spytheman, line 86: Nice.

vlib/v/checker/infix.v
spytheman: I think that it will be best for that to be a notice first, for 1-2 months.

Given that it already found 2 unrelated cases of duplication in the main repo, I think the chance is high, that it can break the CIs of other existing projects too, unless it is a notice for a while.

felipensp: > I think it would look better without "the".
> 
> It would also be nice to add a detail indicating that this error occurs on the right side of the `in` operator, so that the user does not get confused and think that arrays cannot have duplicate values.

The compiler error already has an "arrow" indicating the position of the error.

image

StunxFS: > > I think it would look better without "the".
> > It would also be nice to add a detail indicating that this error occurs on the right side of the `in` operator, so that the user does not get confused and think that arrays cannot have duplicate values.
> 
> The compiler error already has an "arrow" indicating the position of the error.
> 
> image

@felipensp Yes, but that indicates the position of an element within the array, plus the error message only says that there is a duplicate item in the list, that can easily cause confusion

vlib/v/checker/tests/infix_dup_in_err.vv
```diff
@@ -0,0 +1,6 @@
+if 1 in [1, 2, 2, 2] { println('ok') }
+
+a := `a`
+if `a` in [`c`, a, a] { println('ok') }
```
enghitalo, line 4: What about 

``` 
 a := `a` 
 b := `a` // same value of 'a' variable to ensure that checker is for variable name and not variable value
 if `a` in [`c`, b, a, a] { println('ok') } 

```

StunxFS: and with the constants too

Delta456: Yes

felipensp: It just does static analysis, not dynamic analysis. Because we could have more complex conditionals to handle (as ternary, function call, etc).

spytheman: The simple static analysis that it does currently, is fine to cover most cases where `x in [...]` is used.

vlib/v/checker/infix.v
spytheman: @StunxFS I understand your viewpoint, and I agree somewhat with it. In isolation, reading just that message, you can get confused, that V does not allow duplicates at all.

However, that notice/error happens in a very specific place, and it is visible from the context, that in that place, a duplicate would not be useful. 

The message also says: "... is duplicated in the list". That message is 100% true on its face value. The word `list` would not suggest directly to me, that all arrays everywhere, do not allow duplicates. It *can* be interpreted that way, but it is kind of a leap of logic to make that inference.

In my opinion, the current variant is good enough, and can be merged as it is. If there are people that will get confused, and ask about it, we can always go back and add more details.

vlib/v/checker/infix.v
```diff
@@ -953,6 +956,19 @@ fn (mut c Checker) check_div_mod_by_zero(expr ast.Expr, op_kind token.Kind) {
 	}
 }
 
+fn (mut c Checker) check_duplicated_items(node &ast.ArrayInit) {
+	mut unique_items := []string{}
+	for item in node.exprs {
+		item_str := item.str()
+		// println('>>> ${item_str}')
+		if item.str() !in unique_items {
```
spytheman, line 964:
```suggestion
		if item_str !in unique_items {
```

spytheman: Not sure if a map will not be faster to check ... Probably not, since the lists are usually very short 🤔 .

spytheman: image
The vast majority of checks, at least for V itself, will be for lists that are no longer than 8 items. 

On the other end, there is one place that checks for 28 of them, which seems weird 🤔 .

spytheman: image
I think that both are fine.

spytheman (approved):
Excellent work.

PR #22309, opened 2024-09-25
cgen: fix codegen for assign from unsafeexpr resulting fixed array

### LINKED ISSUE #22301: cgen error for chaining a union field access, after a union pointer cast
**V doctor:**
```
V full version: V 0.4.7 4ee948f.e974460
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-09-24 19:59:00

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.39-6-ge974460b
.git/config present: true

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg a.v`
```v
union Convertor {
	su8_array_p [20]u8
	sint_array_p [5]int
}

a := [1,2,3,4,5]!
p := voidptr( unsafe { &Convertor( &a[0] ) } )
c := unsafe { &Convertor( p ).su8_array_p }
dump(a)
dump(p)
dump(c)


```

a compiled program

**What did you see instead?**
```
cc: /tmp/v_1000/a.01J8JSVFAZYGDSVMC5AX0B7X9T.tmp.c:13476: error: '{' expected (got ";")
```
> [!NOTE]
>

felipensp:
Fix #22301


vlib/v/tests/assign/assign_array_fixed_from_union_test.v
```diff
@@ -0,0 +1,16 @@
...
+fn test_main() {
+	a := [1, 2, 3, 4, 5]!
+	p := voidptr(unsafe { &Convertor(&a[0]) })
+	c := unsafe { &Convertor(p).su8_array_p }
+	dump(a)
+	dump(p)
+	dump(c)
+	assert a == [1, 2, 3, 4, 5]!
+	assert p != 0
+	assert c == [u8(1), 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0, 4, 0, 0, 0, 5, 0, 0, 0]!
```
spytheman, line 15: That can fail on big endian machines, but unfortunately we do not have one on the CI.

A more robust assertion could be for a sum of all values in the array, that will be the same `1+2+3+4+5`, for big endian machines too, or asserting that 1 is one of the first 4 bytes (not necessarily the first byte).

spytheman: The actual test however is if it compiles at all, so imho just comment out that last line.

spytheman, line 15:
```suggestion
	$if little_endian {
		assert c == [u8(1), 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0, 4, 0, 0, 0, 5, 0, 0, 0]!
	}
	$if big_endian {
		assert c == [u8(0), 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0, 4, 0, 0, 0, 5]!
	}
```

spytheman (approved):
Excellent work.

PR #22314, opened 2024-09-25
Hub docker action - add cache and trigger

Krotki:
Use cache images from docker repository during build process to store intermediate layers. This change should speed up build process.

Trigger on new release and weekly tags. Others will be ignored.


spytheman (approved):
Good work.

PR #22317, opened 2024-09-26
cgen: fix variadic arg var passed to another call which expects variadic

### LINKED ISSUE #22315: C error: cannot convert 'struct array' to 'struct string *'
**V doctor:**
```
V full version: V 0.4.7 7baff15

CC version: cc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0
thirdparty/tcc: N/A

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg liveview_bug.v`
```v


type Either = string | int

fn tag(params ...Either) int {
	return 0
}

fn div(params ...Either) int {
	return tag(params)
}


```

compilation complete ok

**What did you see instead?**
```
cc: /tmp/v_1000/liveview_bug.01J8NVCRNQZEBSJ5FMBMQZ2DW7.tmp.c:13255: error: cannot convert 'struct array' to 'struct string *'
builder error: 
==================
```
> [!NOTE]
>

felipensp:
Fix #22315

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -2607,8 +2607,9 @@ fn (mut g Gen) call_args(node ast.CallExpr) {
 						}
 					}
 				} else {
-					if args.len == 1
-						&& g.table.type_to_str(arr_info.elem_type) == g.table.type_to_str(varg_type) {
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #22331, opened 2024-09-27
cgen: fix codegen for interface method call which returns array fixed

### LINKED ISSUE #22326: c compiler error on interface method with type alias
I think this is related to the fix in https://github.com/vlang/v/pull/22320 
It seems to cause a new issue:

**V doctor:**
```
V full version: V 0.4.7 18eee34.c7bdc0a

getwd: D:\vbug
vexe: C:\v\v.exe
vexe mtime: 2024-09-27 05:14:05

vroot: OK, value: C:\v
VMODULES: OK, value: C:\Users\info\.vmodules
VTMP: OK, value: C:\Users\info\AppData\Local\Temp\v_0

Git version: git version 2.42.0.windows.2
Git vroot status: weekly.2024.39-29-gc7bdc0a7-dirty
.git/config present: true

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg src\main.v`
```v
module main

pub type Mat4 = [16]f32

pub fn (a Mat4) * (b Mat4) Mat4 {
	mut res := Mat4{}
	for i := 0; i < 4; i++ {
		for j := 0; j < 4; j++ {
			res[j * 4 + i] = 0
			for k := 0; k < 4; k++ {
				res[j * 4 + i] += a[k * 4 + i] * b[j * 4 + k]
			}
		}
	}
	return res
}

interface IGameObject {
	world_transform() Mat4
	parent	?&IGameObject
	transform Mat4
}

struct GameObject implements IGameObject {
mut:
	transform Mat4
	parent    ?&IGameObject
}

fn (gameobject GameObject) world_transform() Mat4 {
	if mut p := gameobject.parent {
		return p.world_transform() * gameobject.transform
	}
	return gameobject.transform
}

fn main() {}
```

no c compiler error

**What did you see instead?**
```
cc: C:/Users/info/AppData/Local/Temp/v_0/main.01J8RZD1SRCTWSMV5JBM3H1DCN.tmp.c:7097: warning: implicit declaration of function 'tcc_backtrace'
cc: C:/Users/info/AppData/Local/Temp/v_0/main.01J8RZD1SRCTWSMV5JBM3H1DCN.tmp.c:13583: error: cannot convert 'struct _v_Array_fixed_f32_16' to 'float *'
... (the original output was 3 lines long, and was truncated to 2 lines)
builder error: 
==================
```
> [!NOTE]
>

felipensp:
Fix #22326

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -1563,6 +1563,11 @@ fn (mut g Gen) method_call(node ast.CallExpr) {
 			g.call_args(node)
 		}
 		g.write(')')
+		if !node.return_type.has_option_or_result() {
+			if g.table.final_sym(node.return_type).kind == .array_fixed {
+				g.write('.ret_arr')
+			}
+		}
 		return
 	}
 	left_sym := g.table.sym(left_type)
--- a/vlib/v/tests/fns/iface_method_fixed_arr_test.v
+++ b/vlib/v/tests/fns/iface_method_fixed_arr_test.v
@@ -1,0 +1,41 @@
+module main
+
+pub type Mat4 = [16]f32
+
+pub fn (a Mat4) * (b Mat4) Mat4 {
+	mut res := Mat4{}
+	for i := 0; i < 4; i++ {
+		for j := 0; j < 4; j++ {
+			res[j * 4 + i] = 0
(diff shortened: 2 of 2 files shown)


spytheman (approved):
Excellent work.

PR #22332, opened 2024-09-27
add encoding.iconv

kbkpbot:
Module `iconv` provides functions convert between vstring(UTF8) to/from different encodings.

```v
assert iconv.vstring_to_encoding('V大法好', 'GB18030' ) == [u8(86), 180, 243, 183, 168, 186, 195]
assert iconv.encoding_to_vstring([u8(86), 180, 243, 183, 168, 186, 195], 'GB18030') == 'V大法好'
```


vlib/encoding/iconv/iconv_windows.c.v
```diff
@@ -0,0 +1,549 @@
...
+pub fn encoding_to_vstring(fromcode string, str []u8) string {
+	dst := conv('UTF-8', fromcode, str.data, str.len)
+	defer {
+		unsafe { dst.free() }
+	}
+	if dst.len == 0 {
+		return ''
+	}
+	return unsafe { cstring_to_vstring(dst.data) }
+}
```
spytheman, line 549: Please extract the public interface that has no platform specific code into its own .v file.

vlib/encoding/iconv/iconv_nix.c.v
```diff
@@ -0,0 +1,52 @@
...
+	return dst
+}
+
+// vstring_to_encoding convert vstring `str` to `tocode` encoding string
+// tips: use `iconv --list` check for supported encodings
+pub fn vstring_to_encoding(tocode string, str string) []u8 {
+	return conv(tocode, 'UTF-8', str.str, str.len)
+}
+
+// encoding_to_vstring convert `fromcode` encoding string to vstring
```
spytheman, line 43:
```suggestion
// encoding_to_vstring convert the given `bytes` in `fromcode` encoding to V string (UTF-8).
```

spytheman:
> It seems missing libiconv on macOS?

Yes, I will install it through `brew` on the CI.

vlib/encoding/iconv/iconv.v
```diff
@@ -0,0 +1,34 @@
...
+// vstring_to_encoding convert V string `str` to `tocode` encoding string
+// tips: use `iconv --list` check for supported encodings
+pub fn vstring_to_encoding(str string, tocode string) []u8 {
+	// convert vstring to []u8
+	mut src := []u8{len: str.len}
+	unsafe { vmemcpy(src.data, str.str, str.len) }
+	defer {
+		unsafe { src.free() }
+	}
+	return conv(tocode, 'UTF-8', mut src)
```
spytheman, line 14: Hmm, why does it make a copy now?

Before, it passed `str.str` and `str.len` directly.

vlib/encoding/iconv/iconv.v
```diff
@@ -0,0 +1,34 @@
...
+
+// encoding_to_vstring converts the given `bytes` using `fromcode` encoding, to a V string (encoded with UTF-8)
+// tips: use `iconv --list` check for supported encodings
+pub fn encoding_to_vstring(bytes []u8, fromcode string) string {
+	// clone bytes => src
+	mut src := bytes.clone()
+	defer {
+		unsafe { src.free() }
+	}
+	mut dst := conv('UTF-8', fromcode, mut src)
```
spytheman, line 25: Same question - the previous version was simpler and did not have to make an additional copy.

vlib/encoding/iconv/iconv_nix.c.v
```diff
@@ -0,0 +1,39 @@
...
+// conv convert `fromcode` encoding string to `tocode` encoding string
+fn conv(tocode string, fromcode string, mut src []u8) []u8 {
+	if src.len == 0 {
+		return []u8{}
+	}
+
+	mut dst := []u8{len: (src.len + 1) * 4} // this should be enough to hold the dst encoding string
+	cd := C.iconv_open(tocode.str, fromcode.str)
+	if isize(cd) == -1 {
+		unsafe { dst.free() }
```
spytheman, line 21: If the function is going to call `dst.free()`, then the function should be marked as `@[manualfree]`, otherwise you will get double frees with the `-autofree` mode.

vlib/encoding/iconv/iconv_nix.c.v
```diff
@@ -0,0 +1,39 @@
...
+
+#include <iconv.h>
+#flag darwin -liconv
+
+fn C.iconv_open(tocode &u8, fromcode &u8) voidptr
+fn C.iconv_close(cd voidptr) int
+fn C.iconv(cd voidptr, inbuf &&u8, inbytesleft &usize, outbuf &&u8, outbytesleft &usize) usize
+
+// conv convert `fromcode` encoding string to `tocode` encoding string
+fn conv(tocode string, fromcode string, mut src []u8) []u8 {
```
spytheman, line 13: why is `src` mut?

vlib/encoding/iconv/iconv_test.v
```diff
@@ -0,0 +1,33 @@
+import encoding.iconv
+
+fn test_vstring_to_encoding() {
+	assert iconv.vstring_to_encoding('abc', 'UTF8') == [u8(97), 98, 99]
+	$if linux {
+		assert iconv.vstring_to_encoding('abc', 'UTF16') == [u8(255), 254, 97, 0, 98, 0, 99, 0]
+	}
+	$if windows {
+		assert iconv.vstring_to_encoding('abc', 'UTF16') == [u8(97), 0, 98, 0, 99, 0]
```
spytheman, line 9: That seems weird ... should not the result be the same on all platforms?

spytheman: The linux version has a BOM at the front 🤔 .
https://en.wikipedia.org/wiki/Byte_order_mark

spytheman: Can it be skipped, so that the output would be the same everywhere?
Does it produce BOMs for other utf encodings?

vlib/encoding/iconv/iconv_test.v
```diff
@@ -0,0 +1,33 @@
...
+	}
+	$if !macos {
+		// CI macos fail, maybe not support UTF16LE
+		assert iconv.vstring_to_encoding('abc', 'UTF16LE') == [u8(97), 0, 98, 0, 99, 0]
+	}
+	// CI ubuntu-docker-musl fail with this
+	// assert iconv.vstring_to_encoding('V大法好abc', 'GB2312') == [u8(86), 180, 243, 183, 168,
+	//	186, 195, 97, 98, 99]
+	assert iconv.vstring_to_encoding('abc', 'encoding_not_exist') == []
+	assert iconv.vstring_to_encoding('', 'UTF16LE') == []
```
spytheman, line 19: It would be more ergonomic if the function returned an error, if it fails a conversion, not return `[]`.

vlib/encoding/iconv/iconv_test.v
```diff
@@ -0,0 +1,33 @@
...
+	assert iconv.encoding_to_vstring([u8(97), 98, 99], 'UTF8') == 'abc'
+	$if !macos {
+		// CI macos fail, maybe not support UTF16LE
+		assert iconv.encoding_to_vstring([u8(97), 0, 98, 0, 99, 0], 'UTF16LE') == 'abc'
+	}
+	// CI ubuntu-docker-musl fail with this
+	// assert iconv.encoding_to_vstring([u8(86), 180, 243, 183, 168, 186, 195, 97, 98, 99],
+	//	'GB2312') == 'V大法好abc'
+	assert iconv.encoding_to_vstring([u8(97), 98, 99], 'encoding_not_exist') == ''
+	assert iconv.encoding_to_vstring([], 'UTF16LE') == ''
```
spytheman, line 32: same, returning an error instead of `[]` would be more ergonomic for the users, so that they can choose to handle the failure right away as they wish, without having to check the .len of the result and the input to detect it.

vlib/encoding/iconv/iconv_windows.c.v
```diff
@@ -0,0 +1,564 @@
...
+	if name_lower.is_int() {
+		return name_lower.int()
+	}
+	// XX123 for debug
+	if name_lower.starts_with('xx') {
+		return name_lower[2..].int()
+	}
+
+	for x in codepage_alias {
+		if x.name.to_lower() == name_lower {
```
spytheman, line 501: Storing the lower-cased versions in the constant will save the .to_lower() calls in the loop.

kbkpbot: Yes, this is copy from MS website, I will convert to lower-case

vlib/encoding/iconv/iconv.v
kbkpbot: En, I will fix it, as WinAPI need a WideChar Num, not byte length.  So in the function, it need to padding zero at end of string, to let the API auto detect the end of string.

But I think WideCharNum = bytes_len/2 , it that correct?  If so, that the padding is no need

spytheman: A wide char is 2 bytes.
From https://learn.microsoft.com/en-us/cpp/c-language/multibyte-and-wide-characters?view=msvc-170: "Wide characters are multilingual character codes that are always 16 bits wide. "

vlib/encoding/iconv/iconv_test.v
JalonSolov: BOMs are only really needed at the beginning of files, not individual strings.  If you don't know what you've just converted a single string to, you really need remedial programming help.

vlib/encoding/iconv/iconv_windows.c.v
kbkpbot: Convert to uppercase is better...

vlib/encoding/iconv/iconv_test.v
kbkpbot: OK, for UTF16, UTF32, remove the BOM

vlib/encoding/iconv/iconv_windows.c.v
```diff
@@ -0,0 +1,563 @@
...
+    Codepage_Alias{57009, 'X-ISCII-MA'}, // ISCII Malayalam 
+    Codepage_Alias{57010, 'X-ISCII-GU'}, // ISCII Gujarati 
+    Codepage_Alias{57011, 'X-ISCII-PA'}, // ISCII Punjabi 
+    Codepage_Alias{65000, 'UTF-7'}, // Unicode (UTF-7) 
+    Codepage_Alias{65001, 'UTF-8'}, // Unicode (UTF-8) 
+	// vfmt on
+]
+
+fn name_to_codepage(name string) int {
+	name_upper := name.to_upper()
```
spytheman, line 476: A special case like this here:
`if name == 'UTF-8' { return 65001 }` 
put before anything else, would be beneficial since half of the checks will be for it.

vlib/encoding/iconv/iconv_nix.c.v
```diff
@@ -0,0 +1,61 @@
...
+#include <iconv.h>
+#flag darwin -liconv
+
+fn C.iconv_open(tocode &u8, fromcode &u8) voidptr
+fn C.iconv_close(cd voidptr) int
+fn C.iconv(cd voidptr, inbuf &&u8, inbytesleft &usize, outbuf &&u8, outbytesleft &usize) usize
+
+// conv convert `fromcode` encoding string to `tocode` encoding string
+fn conv(tocode string, fromcode string, src &u8, src_len int) ![]u8 {
+	if src_len <= 0 {
```
spytheman, line 14:
```suggestion
	if src_len < 0 {
```
passing '' should just return []. 
An empty string is a perfectly valid input in any encoding.

vlib/encoding/iconv/iconv_test.v
```diff
@@ -0,0 +1,113 @@
...
+	{
+		assert ch_str == 'V大法好abc'
+	} else {
+		// some platforms do not support GB2312, skip
+		assert true
+	}
+}
+
+fn test_vstring_to_encoding_endian() {
+	$if big_endian || macos {
```
spytheman, line 44: I understand better now why the BOM was produced, for the UTF16 and UTF32 outputs 🤔 .

With UTF-16le/UTF-16be and UTF-32le/UTF-32be, you specify the byte order explicitly in the encoding name, so the output is just the encoded bytes.

With UTF16 and UTF32 both the little and big endian encoding variants are valid, depending on the platform, and thus the need for the BOM at the start, to show what the produced result uses.

spytheman: We can preserve the BOM for `UTF16` and `UTF32` too, no matter the platform (unix or windows), and assert in the test, that the output will be either the little endian version or the big endian version:
```v
assert abc_utf8 == [u8(254), 255,  0, 97,  0, 98,  0, 99] || 
       abc_utf8 == [u8(255), 254,  97, 0,  98, 0,  99, 0]
```

spytheman: Similarly for the UTF32, we could then assert:
```v
assert abc_utf8 == [u8(0),0,254,255,  0,0,0,97,  0,0,0,98,  0,0,0,99] || 
       abc_utf8 == [u8(255),254,0,0,  97,0,0,0,  98,0,0,0,  99,0,0,0]
```

spytheman: (the goal is for the usages, in the tests to have no platform-dependent code at all)

spytheman:
The windows gcc CI failure is not related to the PR.

vlib/encoding/iconv/iconv_test.v
kbkpbot: Preserve the BOM, will cause some problem when try to contract 2 strings.
I think BOM is only needed when try to create  a new file.
BTW, when use WIN API convert to UTF16 encoding, it has no BOM.

spytheman: My wish, is that all those details, are handled in the module, so that the *users* of the module, will not have to care about them. I.e. to abstract the platform details, so that the users could then focus on just their program, not on how each platform implements the conversion.

If I write code, that does `import encoding.iconv` in a V program, I expect that program to run the same, given the same inputs, no matter what the platform it is running on.

As a user, I would rather have the module not support conversions to `UTF16` and `UTF32` at all, than support them in an inconsistent way, that would then leak to my program, and thus become my problem.

JalonSolov:
Making the BOM an option is probably the best way to go... when saving the data to a file.  If you're writing a file specifically for us by something that doesn't need/want a BOM, then it makes sense to leave it out.  Very handy to have, though if the file might be read by things that do know how to handle a BOM properly.

While a string is in in memory, it really doesn't do anything other than take up extra space.  Your program should know what is in the strings it creates, so it doesn't need a BOM to tell it what format they are.

vlib/encoding/iconv/iconv.v
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// read_file_encoding reads the file in `path` with `encoding` and returns the contents
+pub fn read_file_encoding(path string, encoding string) !string {
+	encoding_name := encoding.to_upper()
+	if encoding_name in ['UTF16', 'UTF32', 'UTF-16', 'UTF-32']! {
+		return error('please use UTF16LE/UTF16BE/UTF32LE/UTF32BE instead')
+	}
+	encoding_bytes := os.read_file_array[u8](path)
+	println(encoding_bytes)
```
spytheman, line 132:
```suggestion
```

vlib/encoding/iconv/iconv.v
```diff
@@ -0,0 +1,135 @@
...
+
+// encoding_to_vstring converts the given `bytes` using `fromcode` encoding, to a V string (encoded with UTF-8)
+// tips: use `iconv --list` check for supported encodings
+pub fn encoding_to_vstring(bytes []u8, fromcode string) !string {
+	encoding_name := fromcode.to_upper()
+	if encoding_name in ['UTF16', 'UTF32', 'UTF-16', 'UTF-32']! {
+		return error('please use UTF16LE/UTF16BE/UTF32LE/UTF32BE instead')
+	}
+	mut dst := conv('UTF-8', fromcode, bytes.data, bytes.len)!
+	dst << [u8(0)] // Windows: add tail zero, to build a vstring
```
spytheman, line 24:
```suggestion
	dst << 0 // add a tail zero, to build a vstring
```

spytheman (approved):
I could not find anything more substantial to nitpick anymore.

Good work.

spytheman (approved):

PR #22335, opened 2024-09-27
cgen: fix interface method call after smartcast

### LINKED ISSUE #17056: Casting interface as itself generates invalid C code
### Describe the bug

Smartcasting an interface to itself creates invalid C-code and generates a casting function (in C) which unnecessarily duplicates the interface struct.

### Expected Behavior

Smartcasting an interface to its self should have no effect!

### Current Behavior

Causes C compiler error.

### Reproduction Steps

The following code produces the problem:

```v
interface Foo {}

fn (f &Foo) func() {}

struct Bar {}

fn main() {
    mut f := Foo(Bar{})
    if mut f is Foo {
        f.func()
    }
}
```

C compiler error:

```
error: cannot take the address of an rvalue of type 'main__Foo' (aka 'struct main__Foo')
                main__Foo_func(&(I_main__Foo_as_I_main__Foo(f)));
                               ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Possible Solution

_No response_

In addition, the following unnecessary code is generated in C by V:

```c
static inline bool I_main__Foo_is_I_main__Foo(main__Foo x) {
	return (x._typ == _main__Foo_main__Bar_index);
}
static inline main__Foo I_main__Foo_as_I_main__Foo(main__Foo x) {
	if (x._typ == _main__Foo_main__Bar_index) return I_main__Bar_to_Interface_main__Foo(x._main__Bar);
	_v_panic(string__plus(string__plus(tos3("`as_cast`: cannot convert "), tos3(v_typeof_interface_main__Foo(x._typ))), tos3(" to Foo")));
	return (main__Foo){0};
}
```

Issues with above generated code:
* `I_main__Foo_is_I_main__Foo(main__Foo x)` is always `True`, isn't it? Or can `x._typ` be 0?
* `I_main__Foo_as_I_main__Foo(main__Foo x)` is unnecessary, I think? 
* `I_main__Foo_as_I_main__Foo(main__Foo x)` calls `I_main__Bar_to_Interface_main__Foo()` which makes a NEW `main.Foo` and copies values in to it!

### V version

V 0.4.0 207203f

macos

felipensp:
Fix #17056

### DIFF
--- a/vlib/v/tests/interfaces/interface_smartcast_call_test.v
+++ b/vlib/v/tests/interfaces/interface_smartcast_call_test.v
@@ -1,0 +1,12 @@
+interface Foo {}
+
+fn (f &Foo) func() {}
+
+struct Bar {}
+
+fn test_main() {
+	mut f := Foo(Bar{})
+	if mut f is Foo {
+		f.func()
+	}
+}
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -1389,6 +1389,12 @@ pub fn (t &Table) is_interface_var(var ScopeObject) bool {
 		&& t.sym(var.smartcasts.last()).kind != .interface_
 }
 
+@[inline]
+pub fn (t &Table) is_interface_smartcast(var ScopeObject) bool {
+	return var is Var && var.orig_type != 0 && t.sym(var.orig_type).kind == .interface_
+		&& var.smartcasts.len > 0
+}
+
 // only used for debugging V compiler type bugs
 pub fn (t &Table) known_type_names() []string {
 	mut res := []string{cap: t.type_idxs.len}
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -1743,6 +1743,9 @@ fn (mut g Gen) method_call(node ast.CallExpr) {
 			if !node.left.is_lvalue() {
 				g.write('ADDR(${rec_cc_type}, ')
 				cast_n++
+			} else if node.left is ast.Ident && g.table.is_interface_smartcast(node.left.obj) {
+				g.write('ADDR(${rec_cc_type}, ')
+				cast_n++
 			} else if !is_array_method_first_last_repeat && !(left_type.has_flag(.shared_f)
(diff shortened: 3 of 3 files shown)


StunxFS:
Hm, why is `foo is Foo` valid? I don't see any sense in it (already, `foo` is `Foo`)

vlib/v/tests/interfaces/interface_smartcast_call_test.v
```diff
@@ -0,0 +1,12 @@
...
+fn (f &Foo) func() {}
+
+struct Bar {}
+
+fn test_main() {
+	mut f := Foo(Bar{})
+	if mut f is Foo {
+		f.func()
+	}
+}
```
spytheman, line 12: The test compiles and runs on master too 🤔 .
Can you use a more specific one, that fails on master either at compile time, or at runtime?

spytheman: Perhaps a `.c.must_have` one in `vlib/v/gen/c/testdata/` ?

felipensp: > The test compiles and runs on master too 🤔 . Can you use a more specific one, that fails on master either at compile time, or at runtime?

It does fails on clang-17. 

```c
cc: /tmp/v_1000/bug.01J8W6JG61Z3FPTMDHS5XWVPF8.tmp.c:13417:18: error: cannot take the address of an rvalue of type 'main__Foo' (aka 'struct main__Foo')
cc:                 main__Foo_func(&(I_main__Foo_as_I_main__Foo(f)));
cc:                                ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc: 1 error generated.
```

spytheman: Yes, it does fail with clang-18 on master. Thanks.

spytheman:
It could be another interface @StunxFS , in which case it will do a smartcast. The issue was that the generated C code compiled with tcc, but not with newer clang or gcc.

PR #22338, opened 2024-09-28
checker: cleanup checking array method call

yuyi98:
This PR cleanup checking array method call.

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -2015,44 +2015,9 @@ fn (mut c Checker) method_call(mut node ast.CallExpr) ast.Type {
 		// c.error('`void` type has no methods', node.left.pos())
 		return ast.void_type
 	}
-	if final_left_sym.kind == .array {
-		if array_builtin_methods_chk.matches(method_name) && (left_sym.kind == .array
-			|| (left_sym.kind == .alias && !left_sym.has_method(method_name))) {
-			return c.array_builtin_method_call(mut node, left_type)
-		} else if method_name in ['insert', 'prepend'] {
-			if method_name == 'insert' {
-				if node.args.len != 2 {
-					c.error('`array.insert()` should have 2 arguments, e.g. `insert(1, val)`',
-						node.pos)
-					return ast.void_type
-				} else {
-					arg_type := c.expr(mut node.args[0].expr)
-					if arg_type !in [ast.int_type, ast.int_literal_type] {
-						c.error('the first argument of `array.insert()` should be integer',
-							node.args[0].expr.pos())
-						return ast.void_type
-					}
-				}
-			} else {
-				if node.args.len != 1 {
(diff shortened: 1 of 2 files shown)


vlib/v/checker/fn.v
```diff
@@ -2015,44 +2015,9 @@ fn (mut c Checker) method_call(mut node ast.CallExpr) ast.Type {
...
-			}
-			arg_type := c.expr(mut arg_expr)
-			arg_sym := c.table.sym(arg_type)
-			if !c.check_types(arg_type, info.elem_type) && !c.check_types(left_type, arg_type) {
-				c.error('cannot ${method_name} `${arg_sym.name}` to `${left_sym.name}`',
-					arg_expr.pos())
-			}
-		}
+	if final_left_sym.kind == .array && array_builtin_methods_chk.matches(method_name)
+		&& !(left_sym.kind == .alias && left_sym.has_method(method_name)) {
```
spytheman, line 2019: `||  (left_sym.kind == .alias && !left_sym.has_method(method_name))`
is different from:
`&& !(left_sym.kind == .alias && left_sym.has_method(method_name))`

I have not analyzed the impact in detail though.
Are you sure about that change?

vlib/v/checker/fn.v
```diff
@@ -3163,6 +3128,52 @@ fn (mut c Checker) array_builtin_method_call(mut node ast.CallExpr, left_type as
...
+		node.return_type = ast.void_type
+		if method := c.table.find_method(left_sym, method_name) {
+			for i, arg in node.args {
+				c.check_expected_call_arg(arg.typ, method.params[i + 1].typ, node.language,
+					arg) or {
+					c.error('${err.msg()} in argument ${i + 1} to `${left_sym.name}.${method_name}`',
+						node.args[i].pos)
+				}
+			}
+		}
```
spytheman, line 3176: This is not in the original deleted code.

It should have its own error tests, and imho should not be added in a PR titled just "cleanup" - it introduces a functional/feature change, not just refactoring.

vlib/v/checker/fn.v
yuyi98: Yes, This is to exclude aliases and define this function, similar to the map_buildin_method_call() judgment later.

vlib/v/checker/fn.v
yuyi98: This test has been included in vlib\v\checker\tests\option_fn_err.vv/out.

spytheman: Thank you. 

I think, that now I understand this better - it is extracting the logic towards `array_builtin_method_call`, but since it is called by `return c.array_builtin_method_call()`, it has to also add the loop for the arguments, to preserve the existing checks 🤔 .

spytheman (approved):
Excellent work.

PR #22344, opened 2024-09-28
orm: allow comptime field.name as field name

### LINKED ISSUE #22329: Compile time variables not useable in ORM
### Describe the bug

It is not possible to use compile time variables in the ORM's sql statements.

### Reproduction Steps

```
module main

import db.sqlite

@[table: 'blog']
pub struct Blog {
	id          int       @[primary; sql: serial]
	slug        string 
	language    string 
}

fn records_by_field[T](db sqlite.DB, fieldname string, value string) ![]T {
	$for field in T.fields {
		if field.name == fieldname {
			entries := sql db {
				select from Blog where field.name == value
			} or { return err }
			return entries 
		}
	}
	return error("fieldname not found")
}


fn main() {
	db := sqlite.connect(':memory:')!
	rows:=records_by_field[Blog](db,'language','da') or {
		println(err)
		return
	}
	println("got ${rows.len}")
}``
```

### Expected Behavior

I expect the code to expand to:
```
fn records_by_field[T](db sqlite.DB, fieldname string, value string) ![]T {
		if 'id' == fieldname {
			entries := sql db {
				select from Blog where id == value
			} or { return err }
			return entries 
		}
		if 'slug' == fieldname {
			entries := sql db {
				select from Blog where slug == value
			} or { return err }
			return entries 
		}
		if 'language' == fieldname {
			entries := sql db {
				select from Blog where language == value
			} or { return err }
			return entries 
		}

	return error("fieldname not found")
}
```

### Current Behavior

A compile error:

```
test_orm.v:16:34: error: ORM: left side of the `==` expression must be one of the `Blog`'s fields
   14 |         if field.name == fieldname {
   15 |             entries := sql db {
   16 |                 select from Blog where field.name == value
      |                                              ~~~~
   17 |             } or { return err }
   18 |             return entries

```

### Possible Solution

_No response_

_No response_

### V version

0.4.7 2db53a4

MacOs
> [!NOTE]
>

felipensp:
Fix #22329


vlib/db/sqlite/sqlite_comptime_field_test.v
```diff
@@ -0,0 +1,43 @@
...
+	id       int @[primary; sql: serial]
+	slug     string
+	language string
+}
+
+fn records_by_field[T](db sqlite.DB, fieldname string, value string) ![]T {
+	$for field in T.fields {
+		if field.name == fieldname {
+			entries := sql db {
+				select from Blog where language == value
```
spytheman, line 16: The original issue had `field.name == value` directly in the query 🤔 .

felipensp: Ooops, my fault.

spytheman (approved):
Excellent work.

PR #22371, opened 2024-09-30
scanner: improve unknown comptime var error

StunxFS:
Before:
```v
main.v:2:11: error: @ must be used before keywords or compile time variables (e.g. `@type string` or `@FN`)
Available compile time variables:
['@VROOT', '@VMODROOT', '@VEXEROOT', '@FN', '@METHOD', '@MOD', '@STRUCT', '@VEXE', '@FILE', '@LINE', '@COLUMN', '@VHASH', '@VCURRENTHASH', '@VMOD_FILE', '@VMODHASH', '@FILE_LINE', '@LOCATION', '@BUILD_DATE', '@BUILD_TIME', '@BUILD_TIMESTAMP']
    1 | fn main() {
    2 |     print(@DEF)
      |              ^
    3 | }
```

After:
```v
main.v:2:11: error: @ must be used before keywords or compile time variables (e.g. `@type string` or `@FN`)
    1 | fn main() {
    2 |     print(@DEF)
      |              ^
    3 | }
Details: available compile time variables: @VROOT, @VMODROOT, @VEXEROOT, @FN, @METHOD, @MOD,
@STRUCT, @VEXE, @FILE, @LINE, @COLUMN, @VHASH, @VCURRENTHASH, @VMOD_FILE, @VMODHASH,
@FILE_LINE, @LOCATION, @BUILD_DATE, @BUILD_TIME, @BUILD_TIMESTAMP
```

In addition to improving the format of the error message, a new method is also added to `string` called `wrap`, with which it is possible to adjust the width of a string.


vlib/builtin/string.v
```diff
@@ -2820,3 +2821,37 @@ pub fn (s string) snake_to_camel() string {
...
+
+@[params]
+pub struct WrapConfig {
+pub:
+	width int    = 80
+	end   string = '\n'
+}
+
+// wrap wraps the given string within `width` in characters.
+pub fn (s string) wrap(config WrapConfig) string {
```
spytheman, line 2833: Please add tests for this, if you consider it valuable as a public method on the builtin `string` in a separate PR, or just move it to a private function in the scanner code for now in this PR.

spytheman: Public functions and methods are much harder and costly to change, and thus they have to be added much more carefully.

StunxFS: Ok, I'll add a test later

StunxFS: Ok, I'll add a test later

StunxFS:
@spytheman I moved the `wrap` method to `v.scanner`, in another PR I will add it as a `string` method and add tests too.

spytheman (approved):
Good work.

PR #22390, opened 2024-10-02
scanner: remove redundancy from the message suggesting to use quotes instead of backticks for characters

StunxFS:
When using "`" for literals that are not characters, the compiler recommends using quotes for string, so far everything is fine, the problem is that when the compiler recommends that, it returns to showing the position where the error occurs, which is unnecessary, since the position is already shown previously. 

This PR removes that part of the suggestion.

Before:
```v
main.v:2:9: error: invalid character literal `ABC` => `ABC` ([`A`, `B`, `C`]) (more than one character)
    1 | fn main() {
    2 |     _ = `ABC`
      |         ^
    3 | }
Details: 
main.v:2:9: details: use quotes for strings, backticks for characters
    1 | fn main() {
    2 |     _ = `ABC`
      |         ^
    3 | }
```

After:
```v
main.v:2:9: error: invalid character literal `ABC` => `ABC` ([`A`, `B`, `C`]) (more than one character)
    1 | fn main() {
    2 |     _ = `ABC`
      |         ^
    3 | }
Details: use quotes for strings, backticks for characters
```

### DIFF
--- a/vlib/v/scanner/scanner.v
+++ b/vlib/v/scanner/scanner.v
@@ -1640,19 +1640,16 @@ pub fn (mut s Scanner) ident_char() string {
 				s.error_with_pos('invalid character literal `${orig}` => `${c}` ([${err_info.join(', ')}]) (escape sequence did not refer to a singular rune)',
 					lspos)
 			} else if u.len == 0 {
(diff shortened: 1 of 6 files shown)


spytheman (approved):
Good find. I do not know how we reached the current state, but it is showing redundant info indeed.

PR #22400, opened 2024-10-03
json2: decoder2 json validation and precalc for memory allocation

enghitalo:

### DIFF
--- a/vlib/x/json2/decoder2/decode.v
+++ b/vlib/x/json2/decoder2/decode.v
@@ -1,6 +1,7 @@
 module decoder2
 
 import time
+import strconv
 
 // Node represents a node in a JSON decoder tree.
 struct Node {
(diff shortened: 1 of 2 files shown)


vlib/x/json2/decoder2/decode.v
```diff
@@ -16,54 +17,493 @@ mut:
...
+
+	// check if generic type matches the JSON type
+	value_kind := get_value_kind(val[0])
+
+	$if T is $option {
+		// TODO
+	} $else $if T is $sumtype {
+		// TODO
+	} $else $if T is $alias {
+		// TODO
```
spytheman, line 52: Please remove such todos. Having a clear error, when something is not supported is better, than having to debug why something fails or is missing silently.

spytheman:
Split this into several PRs.
As it is, it is too large, and it is not clear what it does at all.
It looks more like an info dump, than a reasonable and targeted change with a clear purpose.

PR #22388, opened 2024-10-02
cgen: fix enum with comptime const value (fix #22386)

yuyi98:
This PR fix enum with comptime const value (fix #22386).

- Fix enum with comptime const value.
- Add test.

```v
const enum_value = $if linux { 1 } $else { 2 }

pub enum Test {
	a = enum_value
}

fn main() {
	println(Test.a)
}

PS D:\Test\v\tt1> v run .
a
```

PR #22401, opened 2024-10-03
cgen: perfect enum with comptime const value (related #22388)

yuyi98:
This PR perfect enum with comptime const value (related #22388).

- Avoid non-precomputed const variable names that contain const.


felipensp:
> This PR perfect enum with comptime const value (related #22388).
> 
> * Avoid non-precomputed const variable names that contain const.

Good catch. 👌

Issue #21718, opened 2024-06-23
Cannot compile v compiler in Windows with an account name containing Unicode characters


shnorbluk:
### Describe the bug

I am a French user, and my name contains an accented character (`Noé`). On my Windows PC, with an account having this name, I cloned the v repository and called the `make.bat` script to compile it, as described in installation documentation.  The C compilation fails with an error about writing in temp directory.

### Reproduction Steps

Clone and build the v compiler, following the procedure described [here](https://docs.vlang.io/installing-v-from-source.html):
```batchfile
git clone https://github.com/vlang/v
cd v
make
``` 

### Expected Behavior

It should build the V package without error, or at least inform the user about this issue about having special characters in `%TEMP%` directory.

### Current Behavior

The following output is displayed:
```
Updating TCC
 > Syncing TCC from https://github.com/vlang/tccbin

Updating vc...
 > Sync with remote https://github.com/vlang/vc
remote: Enumerating objects: 2, done.
remote: Counting objects: 100% (2/2), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 2 (delta 1), reused 0 (delta 0), pack-reused 0
Receiving objects: 100% (2/2), 1.03 MiB | 7.45 MiB/s, done.
Resolving deltas: 100% (1/1), done.
Updating files: 100% (2/2), done.

Building V...
 > Attempting to build "./v_win_bootstrap.exe" (from v_win.c) with "thirdparty/tcc/tcc.exe"
 > Compiling "./v.exe" with "./v_win_bootstrap.exe"
> C compiler cmd: "thirdparty/tcc/tcc.exe" "@C:\Users\Noé\AppData\Local\Temp\v_0\v_up.exe.tmp.c.rsp"
> C compiler response file "C:\Users\Noé\AppData\Local\Temp\v_0\v_up.exe.tmp.c.rsp":
  -Bthirdparty/tcc -fwrapv -g -o "C:\\v\\v_up.exe" -Wl,-stack=16777216 -Werror=implicit-function-declaration -I "C:\\v\\thirdparty\\stdatomic\\win" "C:\\Users\\Noé\\AppData\\Local\\Temp\\v_0\\v_up.exe.tmp.c" -std=c99 -D_DEFAULT_SOURCE -municode -bt25 -ldbghelp -lws2_32 -ladvapi32 -lshell32
[...]

Issue #21502, opened 2024-05-15
v does not support Chinese character path name under Windows 10


kbkpbot:
### Describe the bug

v does not support Chinese character path name under Windows 10, it will fail when compile v program.

### Reproduction Steps

create any v file, such as 
```v
module main
fn main() {
    println('hello,world')
}
```

Place this hello.v into a directory, which contain Chinese character, such as 
```sh
D:\v\测试\hello.v
```

Then compile it, 
```sh
v -o hello hello.v
```
It will fail to compile the hello.v

### Expected Behavior

compile and run without error

### Current Behavior

```sh
D:\v\测试>v -o hello hello.v
==================
C:/Users/DDT/AppData/Local/Temp/v_0/hello.01HXXJ9FKJCDVCS554MPZP4VFA.tmp.c:6903: warning: implicit declaration of function 'tcc_backtrace'
tcc: error: could not create 'D:\v\测试\hello.def': No such file or directory
...
==================
(Use `v -cg` to print the entire error message)

builder error:
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```

### Additional Information/Context

compile with msvc , will fail also.
```sh
D:\v\测试>v -o hello hello.v -cc msvc
用于 x64 的 Microsoft (R) C/C++ 优化编译器 19.39.33523 版
版权所有(C) Microsoft Corporation。保留所有权利。

[...]

Issue #20743, opened 2024-02-07
compiler bug


sanrentai:
### Describe the bug

In most cases, compile errors happen because the version of Clang installed is too old
(built by Brecht Sanders, r5) clang version 17.0.6
Target: x86_64-w64-windows-gnu
Thread model: posix
InstalledDir: D:/mingw64/bin

Backend compiler error

### Reproduction Steps

git clone https://github.com/vlang/v
cd v
make

### Expected Behavior

v.exe

### Current Behavior

PS D:\v> .\make.bat
Updating TCC
 > Syncing TCC from https://github.com/vlang/tccbin

Updating vc...
 > Sync with remote https://github.com/vlang/vc

Building V...
 > Attempting to build "./v_win_bootstrap.exe" (from v_win.c) with "thirdparty/tcc/tcc.exe"
 > Compiling "./v.exe" with "./v_win_bootstrap.exe"
> C compiler cmd: "thirdparty/tcc/tcc.exe" "@C:\Users\曹春\AppData\Local\Temp\v_0\v_up.exe.tmp.c.rsp"
> C compiler response file "C:\Users\曹春\AppData\Local\Temp\v_0\v_up.exe.tmp.c.rsp":
  -Bthirdparty/tcc -fwrapv -g -o "D:\\v\\./v_up.exe" -Wl,-stack=16777216 -Werror=implicit-function-declaration -I "D:\\v\\thirdparty\\stdatomic\\win" "C:\\Users\\曹春\\AppData\\Local\\Temp\\v_0\\v_up.exe.tmp.c" -std=c99 -D_DEFAULT_SOURCE -municode -bt25 -ldbghelp -lws2_32 -ladvapi32 -lshell32
> C compiler cmd: gcc "@C:\Users\曹春\AppData\Local\Temp\v_0\v_up.exe.tmp.c.rsp"
> C compiler response file "C:\Users\曹春\AppData\Local\Temp\v_0\v_up.exe.tmp.c.rsp":
  -Bthirdparty/tcc -fwrapv -g -no-pie -o "D:\\v\\./v_up.exe" -Wl,-stack=16777216 -Werror=implicit-function-declaration -I "D:\\v\\thirdparty\\stdatomic\\win" "C:\\Users\\曹春\\AppData\\Local\\Temp\\v_0\\v_up.exe.tmp.c" -std=c99 -D_DEFAULT_SOURCE -municode -ldbghelp -lws2_32 -ladvapi32 -lshell32
[...]

PR #22405, opened 2024-10-04
builder: fix non-ASCII pathname under Windows

kbkpbot:
This PR may help solve #20743, #21502, #21718

Because Windows use CP_ACP encoding, not UTF-8. So the generated rsp file should be encoded in CP_ACP/ANSI encoding under Windows.

This PR will help users build non-ASCII filename/pathname:
```sh
D:\测试目录\你好.v

你好.exe
```

### DIFF
--- a/vlib/v/builder/builder_test.v
+++ b/vlib/v/builder/builder_test.v
@@ -4,9 +4,10 @@ import os
 
 const vexe = @VEXE
 const test_path = os.join_path(os.vtmp_dir(), 'run_check')
+const test_path2 = os.join_path(test_path, '测试目录')
 
 fn testsuite_begin() {
-	os.mkdir_all(test_path) or {}
+	os.mkdir_all(test_path2) or {}
 }
 
 fn testsuite_end() {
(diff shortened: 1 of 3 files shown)


spytheman:
Importing `encoding.iconv` in the compiler is not a good idea.
It will make a module, that depends on the presence of a thirdparty library, that we do not ship with V, a dependency for bootstrapping the compiler.

Besides, the V compiler already has no problem with non ASCII pathnames under windows or anything else.

What does have problem with those, on windows, is the prebuilt tcc executable that we ship in vlang/tccbin .

spytheman:
As for:
> Because Windows use CP_ACP encoding, not UTF-8. So the generated rsp file should be encoded in CP_ACP/ANSI encoding under Windows.

That is worth doing in a private function in v.builder, that has a `$if windows {` branch, containing just a stripped down version of the code needed for converting utf8 to CP_ACP, not the entire `encoding.iconv` module (that also has a lot of unrelated stuff in it, that can break bootstrapping on not just windows).

spytheman:
It is interesting though if only the encoding of the .rsp file has to change 🤔 .
Does passing `-no-rsp` fix building with tcc on windows?

spytheman:
On windows-11 this works, both with and without -no-rsp 🤔 : 
изображение

spytheman:
The `你好.v` file also works:
изображение

PR #22409, opened 2024-10-04
builder: use ANSI .rsp file under Windows

kbkpbot:
…nsi_not_null_terminated

As PR #22405 said, tcc will fail with non-ASCII filename under Windows 10.
So this PR try to make the .rsp file encoded in ANSI.
Also, builtin provide some new funcs help do the job UTF-8=>ANSI.

### DIFF
--- a/vlib/builtin/utf8.c.v
+++ b/vlib/builtin/utf8.c.v
@@ -2,6 +2,7 @@ module builtin
 
 import strings
 
+const cp_acp = 0
 const cp_utf8 = 65001
 
 // to_wide returns a pointer to an UTF-16 version of the string receiver.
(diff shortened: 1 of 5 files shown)


vlib/builtin/utf8.c.v
```diff
@@ -92,3 +93,27 @@ pub fn string_from_wide2(_wstr &u16, len int) string {
...
+		} else {
+			return []u8{}
+		}
+	} $else {
+		s := unsafe { string_from_wide(_wstr) }
+		mut str_to := []u8{len: s.len + 1}
+		unsafe { vmemcpy(str_to.data, s.str, s.len) }
+		return str_to
+	}
+	return []u8{} // TODO: remove this, bug?
```
spytheman, line 118: just `return []` should be enough (and removing lines 109-111)

spytheman: (the cleanup can be done separately though - the CI has already passed a lot of the CI jobs, and I want to merge it ASAP)

Issue #22181, opened 2024-09-08
toml: Provide a way to set default values when decoding TOML


syobocat:
### Describe the feature

Currently, this code:

```v
import toml

struct Object {
	text    string
	integer int
}

fn main() {
	null_toml := toml.parse_text('')!
	object := null_toml.decode[Object]()!
	println(object.text)
	println(object.integer)
}
```

returns

```
toml.Any(toml.Null{})
0
```

and I believe that there is no way to change this behavior (You can use `.reflect()` instead of `.decode()` to get an empty string instead of `toml.Any(toml.Null{})`, though).

I think it is very useful if Vlang has a way to provide default values:

option 1) by an attribute, like
```v
struct Object {
	text    string [toml_default: 'hello']
	integer int    [toml_default: 123]
}
```

or

option 2) by respecting the default value of the field, like
```v
struct Object {
	text    string = 'hello'
	integer int    = 123
}
```

### Use Case

This should be useful, for example, when building an application that uses TOML format for its configuration file.

### Other Information

`json.decode()` respects the default value of the field:

```v
import json

struct Object {
	text    string = 'hello'
	integer int    = 123
}

fn main() {
	object := json.decode(Object, '{}')!
	println(object.text)
	println(object.integer)
}
```

```
hello
123
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [X] This feature might incur a breaking change

### Version used

V 0.4.7 620e064

### Environment details (OS name and version, etc.)

V full version: V 0.4.7 e9c9580.620e064
OS: macos, macOS, 14.6.1, 23G93
Processor: 8 cpus, 64bit, little endian, Apple M3

getwd: /Users/syobon
vexe: /Users/syobon/.v/v
vexe mtime: 2024-09-07 12:37:14

vroot: OK, value: /Users/syobon/.v
VMODULES: OK, value: /Users/syobon/.vmodules
VTMP: OK, value: /tmp/v_501

Git version: git version 2.46.0
Git vroot status: weekly.2024.36-18-g620e0641 (2 commit(s) behind V master)
.git/config present: true

CC version: Apple clang version 15.0.0 (clang-1500.3.9.4)
thirdparty/tcc status: thirdparty-macos-arm64 713692d4
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

[...]

PR #22412, opened 2024-10-05
toml: add reflect/decode struct default value support

kbkpbot:
As #22181 suggested, add default values support when decoding/reflecting TOML to struct.

```v
struct Test {
        val_string   string        = 'abcd'
        val_bool     bool          = true
        val_int      int           = 123
        val_i64      i64           = 1234
        val_u64      u64           = 12345
        val_f32      f32           = 100.1
        val_f64      f64           = 1000.1
        val_datetime toml.DateTime = toml.DateTime{'1980-07-11 21:23:42.123'}
        val_date     toml.Date     = toml.Date{'1977-07-07'}
        val_time     toml.Time     = toml.Time{'11:11:11.111'}
}
```

### DIFF
--- a/vlib/toml/any.v
+++ b/vlib/toml/any.v
@@ -330,186 +330,186 @@ pub fn (a Any) reflect[T]() T {
 	mut reflected := T{}
 	$for field in T.fields {
 		mut toml_field_name := field.name
-		mut default_value := ''
+		mut skip := false
 		// Remapping of field names, for example:
 		// TOML: 'assert = "ok"'
 		// V:    User { asrt string @[toml: 'assert'] }
 		// User.asrt == 'ok'
 		for attr in field.attrs {
+			if attr == 'skip' {
+				skip = true
+				break
+			}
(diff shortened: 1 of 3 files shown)


JalonSolov:
So in this case, you _only_ want a default value if reading TOML?

Why not simply use the V default value, which would apply whether you read TOML or not?
```v
struct Test {
	val_string   string        = 'abcd'
	val_bool     bool          = true
	val_int      int           = 123
	val_i64      i64           = 1234
	val_u64      u64           = 12345
	val_f32      f32           = 100.1
	val_f64      f64           = 1000.1
	val_datetime toml.DateTime = '1980-07-11 21:23:42.123'
	val_date     toml.Date     = '1977-07-07'
	val_time     toml.Time     = '11:11:11.111'
}
```

JalonSolov:
I would rather see support for default values added in builtin.v, than to see separate ways to set defaults for structs based on the type of data being read/written.

spytheman (approved):
Good work.

PR #22415, opened 2024-10-05
checker: disallow blank ident in const and global decl 

Delta456:
Disallows using `_` blank ident in global and const decl.


vlib/v/checker/checker.v
```diff
@@ -2317,7 +2324,7 @@ fn (mut c Checker) global_decl(mut node ast.GlobalDecl) {
 		if sym.kind == .placeholder {
 			c.error('unknown type `${sym.name}`', field.typ_pos)
 		}
-		if field.has_expr {
+		if field.has_expr && field.name != '_' {
```
Delta456, line 2327: If  I don't add this check and have a condition `if field.name == '_'` above. It would still panic.

vlib/v/checker/checker.v
```diff
@@ -1776,6 +1776,13 @@ fn (mut c Checker) const_decl(mut node ast.ConstDecl) {
 			}
 			c.error('duplicate of a module name `${field.name}`', name_pos)
 		}
+		if const_name == '_' {
+			name_pos := token.Pos{
+				...field.pos
+				len: util.no_cur_mod(field.name, c.mod).len
+			}
+			c.error('cannot use `_` as a const name', name_pos)
```
spytheman, line 1784: You can just return early here, after the error.

vlib/v/checker/checker.v
```diff
@@ -2337,6 +2344,9 @@ fn (mut c Checker) global_decl(mut node ast.GlobalDecl) {
 				}
 				v.typ = ast.mktyp(field.typ)
 			}
+			if field.name == '_' {
+				c.error('cannot use `_` as a global name', field.pos)
+			}
```
spytheman, line 2349: Move this check above the others, and return early after the error.
Then you would not have to modify line 2327.

vlib/v/checker/checker.v
```diff
@@ -2337,6 +2344,9 @@ fn (mut c Checker) global_decl(mut node ast.GlobalDecl) {
 				}
 				v.typ = ast.mktyp(field.typ)
 			}
+			if field.name == '_' {
```
spytheman, line 2347:
```suggestion
			if field.has_expr && field.name == '_' {
```

vlib/v/checker/checker.v
Delta456: Fair

vlib/v/checker/checker.v
Delta456: No need if I move the check above and return

vlib/v/checker/checker.v
Delta456: That works

spytheman (approved):
Excellent work.

PR #22416, opened 2024-10-05
checker: fix missing checker for generic array type without concrete types

### LINKED ISSUE #22414: missing checker error for array of generic sumtype
### Describe the bug

Missing checker error, see above.

### Reproduction Steps

```v

// callback types
type CBnoret[T] = fn (val T)
type CBnoret2[T] = fn (val T, prev T)
type CBvret[T] = fn (val T) T
type CBvret2[T] = fn (val T, prev T) T

type Callback[T] = CBnoret[T] | CBnoret2[T] | CBvret[T] | CBvret2[T]

interface IObv[T] {
	v T
	prev T
	cb []Callback[T]
}

struct Obv[T] {
mut:
	v T
	prev T
	a int
	cb []Callback
}
```

### Expected Behavior

Check error

`error: `Callback` type is generic sumtype, must specify the generic type names, e.g. Callback[T], Callback[int]`

### Current Behavior

No error

### Possible Solution

_No response_

_No response_

### V version

current

V full version: V 0.4.8 2db53a4.9b17ed1
OS: linux, Debian GNU/Linux 10 (buster) (VM) (WSL)
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz

getwd: /home/felipe/github/v
vexe: /home/felipe/github/v/v
vexe mtime: 2024-10-05 13:44:13

vroot: OK, value: /home/felipe/github/v
VMODULES: OK, value: /home/felipe/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.20.1
Git vroot status: 0.4.8-40-g9b17ed14-dirty (3 commit(s) behind V master)
.git/config present: true

CC version: cc (Debian 8.3.0-6) 8.3.0

felipensp:
Fix #22414

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzAxODhkZmI5ODBmNDFkYjY2NjY1ODgiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.vy_NpibFEZqL2T9ozE_YOpvOYeG9FmzquZ8n0-qCsLI">Huly&reg;: <b>V_0.6-20878</b></a></sub>


spytheman:
Good work, but please add a test for the error.

spytheman:
Should not it be in the checker, instead of the parser?

spytheman:
The problem with it being in the parser, is that the parser has limited type information.
If the types are defined in another .v file in the same module, it will trigger different behaviour, when the check is done in the parser, depending on whether the other .v files are already parsed or not.

If the check is done in the checker, that is much less likely to happen, since everything is already parsed, and the checker will make sure that its analysis is done based on the full type information from all declarations, from all files.

felipensp:
> Should not it be in the checker, instead of the parser?

Yeah, I moved to the parser because on checker the .typ was already resolved to `[]Callback[T]` even when it was used `[]Callback`.

Issue #22411, opened 2024-10-05
compile time reflection crash when using .params


kbkpbot:
### Describe the bug

When compile a compile time reflection v code, compiler will crash.

### Reproduction Steps

hello.v
```v
struct Test {
}

fn main() {
	$for f in Test.params {
		println(f)
	}
}
```

### Expected Behavior

Compiler should not crash.

### Current Behavior

Compiler crash :
```sh
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:46768: at v__gen__c__Gen_comptime_for: RUNTIME ERROR: invalid memory access
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:39297: by v__gen__c__Gen_stmt
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:39119: by v__gen__c__Gen_stmts_with_tmp_var
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:39025: by v__gen__c__Gen_stmts
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:48167: by v__gen__c__Gen_gen_fn_decl
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:47755: by v__gen__c__Gen_fn_decl
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:39335: by v__gen__c__Gen_stmt
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:39119: by v__gen__c__Gen_stmts_with_tmp_var
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:39025: by v__gen__c__Gen_stmts
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:37917: by v__gen__c__Gen_gen_file
/tmp/v_1000/v2.01J9BZ6PA7SRDB5Q0WPWYXXZ05.tmp.c:37859: by v__gen__c__cgen_process_one_file_cb
[...]

PR #22428, opened 2024-10-06
checker: disallow iterating through .params if not a function

Delta456:
Fixes https://github.com/vlang/v/issues/22411


vlib/v/checker/comptime.v
```diff
@@ -326,11 +326,17 @@ fn (mut c Checker) comptime_for(mut node ast.ComptimeFor) {
...
-		c.comptime.comptime_for_method_param_var = node.val_var
-		c.stmts(mut node.stmts)
-		c.pop_comptime_info()
+		if sym.kind == .function {
+			c.push_new_comptime_info()
+			c.comptime.inside_comptime_for = true
+			c.comptime.comptime_for_method_param_var = node.val_var
+			c.stmts(mut node.stmts)
+			c.pop_comptime_info()
+		} else {
```
medvednikov, line 335: Let's do an early exit instead. Reduces number of indents.

Delta456: Done

vlib/v/checker/comptime.v
```diff
@@ -326,16 +326,16 @@ fn (mut c Checker) comptime_for(mut node ast.ComptimeFor) {
 			c.pop_comptime_info()
 		}
 	} else if node.kind == .params {
-		if sym.kind == .function {
+		if !(sym.kind == .function || sym.name == 'FunctionData') {
+			c.error('iterating over `.params` is supported only for functions, and `${sym.name}` is not a function',
+				node.typ_pos)
+			return
+		} else {
```
spytheman, line 333: Just remove the `} else {` completely, and then reduce the indent level of the code in the branch.

spytheman: i.e. because of the early return:
```
if cond {
   ...
   return
} else {
    ...
}
```
will become:
```
if cond {
   ...
   return
}
...
```

Delta456: done

spytheman (approved):
Good work.

PR #22431, opened 2024-10-07
ast, parser: fix generic fn type with different generic type call (fix #22306)

### LINKED ISSUE #22306: function type generic cannot have a different generic letter when used as a parameter
### Describe the bug

A generic function type with type `T` errors if used as a parameter for another function that has a different generic letter to `T`. A version of this with a struct does work.
```v
struct Foo[T] {
	a T
}

fn f[X](foo Foo[X]) X {
	return foo.a
}

fn main() {
	foo := Foo[int]{10}
	println(f[int](foo))

}
```
however the function type version does not. This leads me to believe even though it is a V error instead of a builder or c error that this is a bug, as I don't see any reason why function types should be excluded from this behavior that structs have.

### Reproduction Steps

```v
type Fn[T] = fn () T

fn problem[X](f Fn[X]) X {
	return f[X]()
}

fn foo() int {
	return 1
}

fn main() {
	t := problem[int](foo)
	println(t)
}

```

### Expected Behavior

To print `1`

### Current Behavior

error: generic type name `T` is not mentioned in fn `problem[X]`

### Possible Solution

_No response_

_No response_

### V version

V 0.4.7 9088970.4248fc5

```
OS: linux, "Arch Linux"
Processor: 8 cpus, 64bit, little endian, 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz

vroot: OK
VMODULES: OK
VTMP: OK
Git vroot status: weekly.2024.37-64-g4248fc55
.git/config present: true

CC version: cc (GCC) 14.2.1 20240805
```
> [!NOTE]
>

yuyi98:
This PR fix generic fn type with different generic type call (fix #22306).

- Fix generic fn type with different generic type call.
- Add test.

```v
type Fn[T] = fn () T

fn problem[X](f Fn[X]) X {
	return f[X]()
}

fn foo() int {
	return 1
}

fn main() {
	t := problem[int](foo)
	println(t)
}

PS D:\Test\v\tt1> v run .
1
```

### DIFF
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -1828,16 +1828,7 @@ pub fn (mut t Table) generic_type_names(generic_type Type) []string {
 			names << t.generic_type_names(sym.info.elem_type)
 		}
 		FnType {
-			mut func := sym.info.func
-			if func.return_type.has_flag(.generic) {
-				names << t.generic_type_names(func.return_type)
-			}
-			func.params = func.params.clone()
-			for mut param in func.params {
-				if param.typ.has_flag(.generic) {
-					generic_names_push_with_filter(mut names, t.generic_type_names(param.typ))
-				}
-			}
+			names << sym.info.func.generic_names
 		}
 		MultiReturn {
 			for ret_type in sym.info.types {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #22434, opened 2024-10-07
checker: reduce allocations part 2

spytheman:
- **checker: reduce allocations in Checker.register_trace_call/2**
- **checker: pass all ast.Fn args by reference explicitly**


spytheman:
image

PR #22444, opened 2024-10-07
checker: don't cast match branch structs to IError when in sumtype

### LINKED ISSUE #22435: match expressions expecting `Result(SumType)` error on yielding `SumType`'s struct values
### Describe the bug

When
* a match-expression is expecting to yield a `Result(SumType)`, and
* a match clause yields a `Struct` which is one of `SumType`'s types
Then the compiler complains that the value does not meet the requirements of the `Result` interface.

### Reproduction Steps

```
type SumType = int | Struct

struct Struct {}

fn foo(x SumType) !SumType {
	return match x {
		int { 5 }
		else { Struct{} }
	}
}
```

### Expected Behavior

Should cause no error.

### Current Behavior

Error:
```
main.v:6:9: error: `Struct` doesn't implement method `msg` of interface `IError`
    4 |
    5 | fn foo(x SumType) !SumType {
    6 |     return match x {
      |            ~~~~~~~~~
    7 |         int { 5 }
    8 |         else { Struct{} }
main.v:6:9: error: `Struct` doesn't implement method `code` of interface `IError`
    4 |
    5 | fn foo(x SumType) !SumType {
    6 |     return match x {
      |            ~~~~~~~~~
    7 |         int { 5 }
    8 |         else { Struct{} }
```

### Possible Solution

_No response_

Replacing the value in the `else{}` clause with either of the following fixes the error:
* 7
* `SumType(Struct{})`

However, replacing the else clause value with
* `foo()`
Where foo is `fn foo() Struct`

Still fails in the same way.

So it would seem that the error is triggered specifically when using one of the sumtype's struct types.

This is a fairly recent error. This code compile in Feb 2023.

### V version

V 0.4.8 5a33c36

MacOS
> [!NOTE]
>

edam:
Fixes #22435.

When checking match branch expression branch expression type, there was a check to cast yielded value to an `IError` when match return type is a Result and branch expression type is struct.

I added an additional test so that the cast (and check to ensure the struct conforms to the `IError` interface) does not occur when the return type is a sumtype and the struct is one of the sumtype types.

Tests updated to cover this

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzA0NDYyMmI5ODBmNDFkYjY2ZDYwOTEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.lbvkaNkxRJ-eV-Ua7mKCPlfOytDK-cu3fHrRSOhGecU">Huly&reg;: <b>V_0.6-20905</b></a></sub>

### DIFF
--- a/vlib/v/tests/return_match_expr_of_sumtype_result_test.v
+++ b/vlib/v/tests/return_match_expr_of_sumtype_result_test.v
@@ -1,25 +1,73 @@
-type Foo = int | string
+type Foo = int | string | Bar
 
-fn foo1() !Foo {
+struct Bar {}
+
+fn foo_result_struct_second() !Foo {
+	return match true {
+		true { 1 }
+		else { Bar{} }
+	}
+}
+
+fn foo_option_struct_second() ?Foo {
+	return match true {
+		true { 1 }
+		else { Bar{} }
+	}
+}
+
+fn foo_result_string_second() !Foo {
 	return match true {
 		true { 1 }
 		else { '' }
 	}
 }
 
-fn foo2() ?Foo {
+fn foo_option_string_second() ?Foo {
 	return match true {
 		true { 1 }
 		else { '' }
+	}
+}
+
+fn foo_result_struct_first() !Foo {
+	return match true {
+		true { Bar{} }
+		else { 7 }
+	}
+}
+
+fn foo_option_struct_first() ?Foo {
+	return match true {
+		true { Bar{} }
+		else { 7 }
 	}
 }
 
 fn test_return_match_expr_of_sumtype_opt_res() {
-	ret1 := foo1() or { return }
-	println(ret1)
-	assert '${ret1}' == 'Foo(1)'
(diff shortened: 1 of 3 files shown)


vlib/v/tests/return_match_expr_of_sumtype_result_test.v
spytheman, line 1: Please keep the old _test.v file, and add a new one.

Refactoring the old one (not just adding new test cases) has a non-trivial chance of masking/changing what it tested before.

PR #22446, opened 2024-10-08
string buffer: string_buffer_to_generic method

enghitalo:
With this PR I intend to drastically reduce redundant codes, which are abundant in json2, like the one below.

```v
		$for field in T.fields {
				$if field.typ is u8 {
					typ.$(field.name) = res[json_name]!.u64()
				} $else $if field.typ is u16 {
					typ.$(field.name) = res[json_name]!.u64()
				} $else $if field.typ is u32 {
					typ.$(field.name) = res[json_name]!.u64()
				} $else $if field.typ is u64 {
					typ.$(field.name) = res[json_name]!.u64()
				} $else $if field.typ is int {
					typ.$(field.name) = res[json_name]!.int()
				} $else $if field.typ is i8 {
					typ.$(field.name) = res[json_name]!.int()
				} $else $if field.typ is i16 {
					typ.$(field.name) = res[json_name]!.int()
				} $else $if field.typ is i32 {
					typ.$(field.name) = i32(res[field.name]!.i32())
				} $else $if field.typ is i64 {
					typ.$(field.name) = res[json_name]!.i64()
				} 
		}
```

to only
```v
		$for field in T.fields {
				$if field.typ in [$int, $float] {
					unsafe { bytes.string_buffer_to_generic(&typ.$(field.name) ) }
				} 
		}
```

### DIFF
--- a/vlib/builtin/array.v
+++ b/vlib/builtin/array.v
@@ -1044,10 +1044,24 @@ pub fn (data &u8) vbytes(len int) []u8 {
 	return unsafe { voidptr(data).vbytes(len) }
 }
 
+// string_buffer_to_generic converts a buffer of bytes (data) into a generic type T and stores the result in the provided result pointer.
+// The function supports conversion to the following types:
+// - Signed integers: i8, i16, int, i64
+// - Unsigned integers: u8, u16, u32, u64
+// - Floating-point numbers: f32, f64
+//
+// For signed integers, the function handles negative numbers by checking for a '-' character.
+// For floating-point numbers, the function handles decimal points and adjusts the result accordingly.
+//
+// If the type T is not supported, the function will panic with an appropriate error message.
+//
+// Parameters:
+// - data []u8: The buffer of bytes to be converted.
+// - result &T: A pointer to the variable where the converted result will be stored.
+//
+// NOTE: This aims works with not new memory allocated data, to more efficient use `vbytes` before.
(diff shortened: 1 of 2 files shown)


spytheman:
`builtin` can not use generics. You can put it in `arrays` if you want.

vlib/builtin/array.v
```diff
@@ -1043,3 +1043,64 @@ pub fn (data voidptr) vbytes(len int) []u8 {
...
+// For signed integers, the function handles negative numbers by checking for a '-' character.
+// For floating-point numbers, the function handles decimal points and adjusts the result accordingly.
+//
+// If the type T is not supported, the function will panic with an appropriate error message.
+//
+// Parameters:
+// - data []u8: The buffer of bytes to be converted.
+// - result &T: A pointer to the variable where the converted result will be stored.
+//
+// NOTE: This aims works with not new memory allocated data, to more efficient use `vbytes` before.
```
spytheman, line 1062: That does not look like a generally useful function to me - it is full of special cases,
it panics, and it passes a buffer by an immutable reference, while writing into it :-| .

enghitalo: Support to others generics will be included un future like, enum, bool, Alias...

enghitalo: I think a more generally useful function will be json decode

PR #22463, opened 2024-10-09
picoev: fix wrong return on failure

### LINKED ISSUE #22292: When a port is unavailable, veb segfaults
**V doctor:**
```
V full version: V 0.4.7 0117c7f.03e0b9e
OS: macos, macOS, 14.6.1, 23G93

getwd: /Users/Z002DCM
vexe: /Users/Z002DCM/apps/v/v
vexe mtime: 2024-09-24 03:42:14

vroot: OK, value: /Users/Z002DCM/apps/v
VMODULES: OK, value: /Users/Z002DCM/.vmodules
VTMP: OK, value: /tmp/v_503

Git vroot status: weekly.2024.39-3-g03e0b9e6
.git/config present: true

CC version: Apple clang version 16.0.0 (clang-1600.0.26.3)

```

**What did you do?**
Run the same veb instance twice on the same machine. 
The first run works fine, and the port 8080 will be serving.
The second run fails to acquire the port 8080. Instead of failing gracefully, it throws a segfaiult

`./v -g -o vdbg cmd/v && ./vdbg /tmp/x.v`
```v
import veb

pub struct User {
pub mut:
    name string
    id   int
}

// Our context struct must embed `veb.Context`!
pub struct Context {
    veb.Context
pub mut:
    // In the context struct we store data that could be different
    // for each request. Like a User struct or a session id
    user       User
    session_id string
}

pub struct App {
pub:
    // In the app struct we store data that should be accessible by all endpoints.
    // For example, a database or configuration values.
    secret_key string
}

// This is how endpoints are defined in veb. This is the index route
    return ctx.text('Hello V! The secret key is "${app.secret_key}"')
}

fn main() {
    mut app := &App{
        secret_key: 'secret'
    }
    // Pass the App and context type and start the web server on port 8080
}

```

When a port is unavailable, veb should fail gracefully

**What did you see instead?**
```shell
[veb] Running app on http://localhost:8080/
Error during listen: net: socket error: 48; binding to :8080 failed
signal 11: segmentation fault
0   libsystem_platform.dylib            0x000000018ba72584 _sigtramp + 56
1   x                                   0x000000010493b64c picoev__Picoev_loop_once + 24
2   x                                   0x000000010493d5f8 picoev__Picoev_serve + 204
3   x                                   0x000000010495cfe8 veb__run_at_T_main__App_main__Context + 1924
4   x                                   0x000000010495c808 veb__run_T_main__App_main__Context + 116
5   x                                   0x0000000104961ee8 main__main + 68
6   x                                   0x0000000104963a6c main + 88
7   dyld                                0x000000018b6b7154 start + 2476
```
> [!NOTE]
>

felipensp:
Fix #22292

Just noticed running `examples/veb/veb_assets/vweb_assets.v`

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzA2NjdkZGI5ODBmNDFkYjY4ODcyNDIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.BxEjstnqB3OkW18WEmgzptDoBFvO16FB2-PmK2Rmk2w">Huly&reg;: <b>V_0.6-20922</b></a></sub>

### DIFF
--- a/vlib/picoev/picoev.v
+++ b/vlib/picoev/picoev.v
@@ -352,7 +352,7 @@ fn default_error_callback(data voidptr, req picohttpparser.Request, mut res pico
 pub fn new(config Config) !&Picoev {
 	listening_socket_fd := listen(config) or {
 		eprintln('Error during listen: ${err}')
-		return unsafe { nil }
+		return err
 	}
 
 	mut pv := &Picoev{


spytheman (approved):
Good find.
Would not just `listening_socket_fd := listen(config)!` work, or is the error hidden otherwise?

esquerbatua:
> Good find. Would not just `listening_socket_fd := listen(config)!` work, or is the error hidden otherwise?

The error should happens also, because it's from the picoev spawned thread, when makes the first event loop without the socket created

PR #22470, opened 2024-10-10
cgen: fix option if expr with noreturn fn call (fix #22467)

### LINKED ISSUE #22467: cgen error for an if expression, with a panic() call in the else branch, assigned to a `?string` variable
**V doctor:**
```
V full version: V 0.4.8 7565841.72b4804
OS: linux, Ubuntu 20.04.6 LTS

getwd: /space/v/oo
vexe: /space/v/oo/v
vexe mtime: 2024-10-10 01:31:53

vroot: OK, value: /space/v/oo
VTMP: OK, value: /tmp/v_1000

Git vroot status: 0.4.8-75-g72b48047
.git/config present: true

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg x.v`
```v
mut x := ?string(none)
x = if true {
	''
} else {
	panic('')
}
println(x)

```

a compiled program

**What did you see instead?**
```
cc: /tmp/v_1000/x.01J9T0VW2FV63VCS5NVGQ8NEYN.tmp.c:13450: error: declaration of an array of incomplete type elements
```
> [!NOTE]
>

yuyi98:
This PR fix option if expr with panic() call (fix #22467).

- Fix option if expr with panic() call.
- Add test.

```v
fn main() {
	mut x := ?string(none)
	x = if true {
		''
	} else {
		panic('')
	}
	println(x)
}

PS D:\Test\v\tt1> v run .
Option('')
```

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -1926,9 +1926,14 @@ fn (mut g Gen) stmts_with_tmp_var(stmts []ast.Stmt, tmp_var string) bool {
 								g.fn_decl.return_type.clear_flag(.option)
 							}
 							styp = g.base_type(ret_typ)
-							g.write('_option_ok(&(${styp}[]) { ')
-							g.expr_with_cast(stmt.expr, stmt.typ, ret_typ)
-							g.writeln(' }, (${option_name}*)(&${tmp_var}), sizeof(${styp}));')
+							if stmt.expr is ast.CallExpr && stmt.expr.name == 'panic' {
+								g.expr(stmt.expr)
+								g.writeln(';')
+							} else {
+								g.write('_option_ok(&(${styp}[]) { ')
+								g.expr_with_cast(stmt.expr, stmt.typ, ret_typ)
+								g.writeln(' }, (${option_name}*)(&${tmp_var}), sizeof(${styp}));')
+							}
 						}
 					}
 				}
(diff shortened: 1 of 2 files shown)


vlib/v/gen/c/cgen.v
```diff
@@ -1926,9 +1926,14 @@ fn (mut g Gen) stmts_with_tmp_var(stmts []ast.Stmt, tmp_var string) bool {
 								g.fn_decl.return_type.clear_flag(.option)
 							}
 							styp = g.base_type(ret_typ)
-							g.write('_option_ok(&(${styp}[]) { ')
-							g.expr_with_cast(stmt.expr, stmt.typ, ret_typ)
-							g.writeln(' }, (${option_name}*)(&${tmp_var}), sizeof(${styp}));')
+							if stmt.expr is ast.CallExpr && stmt.expr.name == 'panic' {
```
spytheman, line 1929: It is not just `panic`. Any `@[noreturn]` function will trigger the same problem. 
Try for example `exit(0)`.

yuyi98: OK, I see.

yuyi98:
```v
if stmt.expr is ast.CallExpr && stmt.expr.is_noreturn {
```

spytheman (approved):
Excellent work.

PR #22483, opened 2024-10-10
veb,vweb: MIME types should be case insensitive

### LINKED ISSUE #22425: Unknown MIME type for uppercase JPG
### Describe the bug

I used a file named P123.JPG in a veb project and V panics about the MIME type.

### Reproduction Steps

Load a picture with an uppercase extension inside project using veb.

### Expected Behavior

MIME types should be case insensitive.

### Current Behavior

I get this error:
 `V panic: result not set (unknown MIME type for file extension ".JPG". You can register your MIME type in app.static_mime_types)`

### Possible Solution

_No response_

_No response_

### V version

0.4.8

Linux Mint
> [!NOTE]
>

alex-semenyuk:
Close #22425

As mentioned at #22425 MIME types should be case insensitive. So if we have .TXT or .JPG it should be valid as .txt and .jpg


spytheman (approved):
Good work.

JalonSolov:
Question:  Shouldn't veb be using the `net.http.mime` module, instead of partially duplicating the MIME types that are already in there?

spytheman:
> Question: Shouldn't veb be using the `net.http.mime` module, instead of partially duplicating the MIME types that are already in there?

Ideally yes, but realistically - no. 

The DB in `net.http.mime` is huge when uncompressed, and with the current implementation of `net.http.mime` and the current compiler, means both much bigger `veb` executables (+2.1MB without -prod, +396KB with -prod), and much more time, spend during optimization (compilation with -prod), for producing each of them:
image

The current smaller `mime` map in `veb` and `vweb` before that, favors much faster iteration and smaller executables, which I think is the right balance for both modules, until they mature.

PR #22501, opened 2024-10-12
parser: implement fn name using keyword

yuyi98:
This PR implement fn name using keyword.

- Implement fn name using keyword.
- Add test.

```v
struct Foo {}

fn (f Foo) lock() int {
	return 22
}

fn (f Foo) mut() bool {
	return true
}

fn select() bool {
	return true
}

fn lock[T](t T) T {
	return t
}

fn match() bool {
	return true
}

fn main() {
	f := Foo{}
	assert select()
	assert match()

	assert lock[int](22) == 22
	assert lock(22) == 22
	assert lock[string]('hello') == 'hello'
	assert lock('hello') == 'hello'

	assert f.lock() == 22
	assert f.mut()
}

PS D:\Test\v\tt1> v run .  
```


spytheman:
Why though?

That is introducing *more* context sensitivity in the parser.

It will make all tooling harder to make, and the parsing potentially slower.

Is the benefit of being able to `fn mut() {}` for example, worth it?

yuyi98:
Sorry, I just don't like:
```v
	cb.mutex.@lock()
        ...
```
```v
fn (mut s UdpSocket) @select(test Select, timeout time.Duration) !bool {
	return @select(s.handle, test, timeout)
}
```

medvednikov:
I agree with both of you.

I hate `mutex.@lock` and `@select`.

I also don't like the possible confusion it can introduce.

I'm going to merge it however. Great job!

PR #22502, opened 2024-10-12
cgen: allow for `@[c_extern] fn C.somename(param1 int, param2 &char) &char`

spytheman:
- **cgen: allow for `@[c_extern] fn C.somename(param1 int, param2 &char) &char`**
- **add a test .vv/.c.must_have pair**


medvednikov:
Can you document the new attribute please?

spytheman:
It is not a new attribute, it is already documented for usage with C globals.

I would like to avoid documenting its usage with C functions for now, until it is used a bit more in the field by people with actual projects, that need it (see the messages by @raojm in `#compiler-and-vlib-dev` on Discord), and bugs related to it fixed.

PR #22512, opened 2024-10-13
parser: allow for variable name using keyword 'type'

yuyi98:
This PR allow for variable name using keyword 'type'.

- Allow for variable name using keyword 'type'.
- Add test.

```v
fn info(type int) int {
	return type
}

fn print_info() {
	for type in [1, 2, 3] {
		println(type)
	}
}

fn main() {
	ret := info(22)
	println(ret)
	assert ret == 22

	type := 33
	println(type)
	assert type == 33

	print_info()
	assert true
}

PS D:\Test\v\tt1> v run .    
22
33
1
2
3
```

### DIFF
--- a/vlib/v/parser/expr.v
+++ b/vlib/v/parser/expr.v
@@ -519,8 +519,14 @@ fn (mut p Parser) check_expr(precedence int) !ast.Expr {
 					p.next()
 					return p.struct_init('', .anon, false)
 				}
-			}
-			if p.tok.kind != .eof && !(p.tok.kind == .rsbr && p.inside_asm) {
+			} else if p.tok.kind == .key_type {
+				// variable name: type
+				ident := p.ident(ast.Language.v)
+				node = ident
+				p.mark_var_as_used(ident.name)
+				p.add_defer_var(ident)
+				p.is_stmt_ident = is_stmt_ident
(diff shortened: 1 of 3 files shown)


vlib/v/parser/expr.v
```diff
@@ -519,8 +519,14 @@ fn (mut p Parser) check_expr(precedence int) !ast.Expr {
 					p.next()
 					return p.struct_init('', .anon, false)
 				}
-			}
-			if p.tok.kind != .eof && !(p.tok.kind == .rsbr && p.inside_asm) {
+			} else if p.tok.kind == .key_type {
+				// variable name: type
+				ident := p.ident(ast.Language.v)
```
StunxFS, line 524: Why not just use `.v`?

```v
ident := p.ident(.v)
```

yuyi98: It is currently only used in v.

medvednikov: he meant using `p.ident(.v)` instead of `p.ident(ast.Language.v)`

indeed that's simpler and I thought vfmt simplified it already

yuyi98: I see. 
Yes, it can. Thank you!

PR #22529, opened 2024-10-15
thirdparty: update `stb_image_resize2.h` from `2.06` to `2.11`

larpon:


thirdparty/stb_image/stb_image_resize2.h
```diff
@@ -1244,7 +1274,7 @@ static stbir__inline stbir_uint8 stbir__linear_to_srgb_uchar(float in)
 
 #else // STBIR_SIMD
 
-#if defined(STBIR_SSE2) && !defined(__TINYC__)
+#ifdef STBIR_SSE2
```
spytheman, line 1277: this is also needed for tcc

larpon:
Sorry I missed those patches... maybe we could find a way that somehow warns if pathes are missing in thirdparty code on PRs 🤔 (a bit like `missdoc` does a comparison)

spytheman:
That would be nice, but I do not know how would it work?
Counting the number of `// __v_ start` or ?

spytheman:
One easy strategy would be, to store the output of `git grep -c '// __v_ start'`:
image

to a checked in file in `thirdparty/`, for example:
`git grep -c '// __v_ start' > thirdparty/custom_patches.txt`

Then the CI can just run that same grep command, and bail, if the output is different from the stored one.

spytheman:
People that make updates, can also easily update the file too, if they are sure, that everything is fine, to silence the CI.

spytheman:
It would not have discovered these ones though, since they were not marked with the `// __v_ start` . 

(I knew where to search for them, because I merged the changes before, and also because of the failing CI jobs)

larpon:
Yeah, hmm 🤔 I haven't thought the actual method through. I would just be nice to have. And you are right that it would not have triggered in this case anyway 🤷🏼‍♂️

PR #22530, opened 2024-10-15
v: allow `C.va_arg(Type, va)`

### LINKED ISSUE #20636: Insufficient support for va_list type

felipensp:
Fix #20636

Improving C interop 👍🏻

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzBlNWRiMThkY2MwMDM2MTU1YjAyMWQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.oRJFuFrBDcLjb21_hH6ys8CqOVBbOtOoItWu04JhtlA">Huly&reg;: <b>V_0.6-20984</b></a></sub>

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -2026,8 +2026,7 @@ fn (mut g Gen) fn_call(node ast.CallExpr) {
 	} else if is_va_arg {
 		ast_type := node.args[0].expr as ast.TypeNode
 		typ := g.typ(ast_type.typ)
-		g.write(util.no_dots(name[2..]))
-		g.write('(')
+		g.write('va_arg(')
 		g.expr(node.args[1].expr)
 		g.write(', ${typ})')
 		return
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -1976,6 +1976,7 @@ fn (mut g Gen) fn_call(node ast.CallExpr) {
 	is_json_encode_pretty := name == 'json.encode_pretty'
 	is_json_decode := name == 'json.decode'
 	is_json_fn := is_json_encode || is_json_encode_pretty || is_json_decode
+	is_va_arg := name == 'C.va_arg'
 	mut json_type_str := ''
 	mut json_obj := ''
 	if is_json_fn {
@@ -2022,6 +2023,14 @@ fn (mut g Gen) fn_call(node ast.CallExpr) {
 		g.write('\n${cur_line}')
 		name = ''
 		json_obj = tmp2
+	} else if is_va_arg {
+		ast_type := node.args[0].expr as ast.TypeNode
+		typ := g.typ(ast_type.typ)
+		g.write(util.no_dots(name[2..]))
+		g.write('(')
+		g.expr(node.args[1].expr)
+		g.write(', ${typ})')
+		return
 	}
 	if name == '__addr' {
 		name = '&'
--- a/vlib/v/tests/va_arg_test.v
+++ b/vlib/v/tests/va_arg_test.v
@@ -12,7 +12,7 @@ fn sum(qtd int, ...) int {
 	mut s := 0
 	C.va_start(va, qtd)
 	for i := 0; i < qtd; i++ {
-		s += int(C.va_arg(int, va))
+		s += C.va_arg(int, va)
 	}
 	return s
 }
--- a/vlib/v/tests/va_arg_test.v
+++ b/vlib/v/tests/va_arg_test.v
@@ -1,0 +1,22 @@
+module main
+
+@[typedef]
+struct C.va_list {}
+
+fn C.va_start(voidptr, voidptr)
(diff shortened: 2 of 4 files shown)


vlib/v/gen/c/fn.v
```diff
@@ -2022,6 +2023,14 @@ fn (mut g Gen) fn_call(node ast.CallExpr) {
 		g.write('\n${cur_line}')
 		name = ''
 		json_obj = tmp2
+	} else if is_va_arg {
+		ast_type := node.args[0].expr as ast.TypeNode
+		typ := g.typ(ast_type.typ)
+		g.write(util.no_dots(name[2..]))
+		g.write('(')
```
StunxFS, line 2030:
```suggestion
		g.write('va_arg(')
```

I think it would be better to write `va_arg(` at once, instead of using `util.no_dots(name[2..])` (because we already know that function is being used)

spytheman: Yes, it also makes it easier to search later, if there is a bug in the implementation, by using patterns derived by the generated C source code.

vlib/v/gen/c/fn.v
```diff
@@ -2022,6 +2023,14 @@ fn (mut g Gen) fn_call(node ast.CallExpr) {
 		g.write('\n${cur_line}')
 		name = ''
 		json_obj = tmp2
+	} else if is_va_arg {
+		ast_type := node.args[0].expr as ast.TypeNode
+		typ := g.typ(ast_type.typ)
+		g.write(util.no_dots(name[2..]))
```
spytheman, line 2029:
```suggestion
```

vlib/v/tests/va_arg_test.v
```diff
@@ -0,0 +1,22 @@
+module main
+
+@[typedef]
+struct C.va_list {}
+
+fn C.va_start(voidptr, voidptr)
+fn C.va_arg(voidptr, voidptr) voidptr
+fn C.va_end(voidptr)
```
spytheman, line 8: Should be declare those in `vlib/builtin/cfns.c.v` ?

PR #22533, opened 2024-10-15
checker: disallow C/JS functions and structs to be declared as generic

StunxFS:
Currently the compiler allows you to declare C/JS functions and structs as generics, this PR prevents that from happening.

```v
$ cat main.v
struct C.Ptr[T] {
	ptr &T
}

fn C.alloc[T]() &T

fn JS.encode[T](data string) T

struct JS.Some[T] {
	i T
}

fn main() {}
```

```bash
$ v main.v
main.v:1:1: error: C structs cannot be declared as generic
    1 | struct C.Ptr[T] {
      | ~~~~~~~~~~~~
    2 |     ptr &T
    3 | }
main.v:5:1: error: C functions cannot be declared as generic
    3 | }
    4 | 
    5 | fn C.alloc[T]() &T
      | ~~~~~~~~~~~~~~~~~~
    6 | 
    7 | fn JS.encode[T](data string) T
main.v:7:1: error: JS functions cannot be declared as generic
    5 | fn C.alloc[T]() &T
    6 | 
    7 | fn JS.encode[T](data string) T
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    8 | 
    9 | struct JS.Some[T] {
main.v:9:1: error: JS structs cannot be declared as generic
    7 | fn JS.encode[T](data string) T
    8 | 
    9 | struct JS.Some[T] {
      | ~~~~~~~~~~~~~~
   10 |     i T
   11 | }
```


vlib/v/checker/fn.v
```diff
@@ -13,6 +13,9 @@ fn (mut c Checker) fn_decl(mut node ast.FnDecl) {
 			eprintln('>>> post processing node.name: ${node.name:-30} | ${node.generic_names} <=> ${c.table.cur_concrete_types}')
 		}
 	}
+	if node.language == .c && node.generic_names.len > 0 {
```
Delta456, line 16: I believe the check can simply be `if node.language != .v` which will disallow all backends except the V one.

vlib/v/checker/struct.v
```diff
@@ -11,6 +11,9 @@ fn (mut c Checker) struct_decl(mut node ast.StructDecl) {
 	defer {
 		util.timing_measure_cumulative(@METHOD)
 	}
+	if node.language == .c && node.generic_types.len > 0 {
```
Delta456, line 14: Here too

vlib/v/checker/fn.v
JalonSolov: Hmm... probably good idea for most of the backends, but the native backend (if ever completed) should be fine.  WASM... I don't know.  Go... latest Go supposedly supports generics, but I haven't messed with that.

vlib/v/checker/struct.v
spytheman: same - look at the following message

vlib/v/checker/fn.v
JalonSolov: At most, perhaps `if (node.language == .c || node.language == .js) && ...` since those 2 may be the only supported backends that don't support generics.

StunxFS: now it is also no longer allowed to declare JS structs and functions as generic

PR #22534, opened 2024-10-15
checker: clarify error message about using empty map literal to initialize structs

### LINKED ISSUE #22498: Confusing error message when trying to initialize an anonymous struct with `{}`
### Describe the bug

Running the code below, the compiler produces a confusing error message and a completely unrelated message just after.

### Reproduction Steps

```v
struct AStructName {
	foo int
	anon struct {
		bar int
	}
}

fn main() {
	s := AStructName{5, {}}
	println(s)
}
```

### Expected Behavior

An error that says to use `struct {}` to initialize the anonymous struct.

### Current Behavior

```
test.v:9:22: error: `{}` can not be used for initialising empty structs any more. Use `_VAnonStruct1{}` instead.
    7 |
    8 | fn main() {
    9 |     s := AStructName{5, {}}
      |                         ~~
   10 |     println(s)
   11 | }
test.v:9:22: error: `map{  }` (no value) used as value
    7 |
    8 | fn main() {
    9 |     s := AStructName{5, {}}
      |                         ~~
   10 |     println(s)
   11 | }
```

### Possible Solution

Change the error message to suggest using `struct {}` to initialize the embedded struct.

The code in the error message does work, which is a little bit weird:
```v
struct AStructName {
	foo int
	anon struct {
		bar int
	}
}

fn main() {
	s := AStructName{5, _VAnonStruct1{}}
	println(s)
}
```
the above compiles and runs, but it probably shouldn't (since `_VAnonStruct1` is a type defined by the compiler). Could be related to a different issue.

### V version

V 0.4.8 ab862c4

```
V full version: V 0.4.8 469b56e.ab862c4

vexe mtime: 2024-10-12 07:01:18

Git vroot status: weekly.2024.41-16-gab862c42
.git/config present: true

```
> [!NOTE]
>

StunxFS:
This PR clarifies the error message about using empty map literal to initialize structs, in addition to that it prevents another error message from appearing when using empty map literal with anonymous structs.

Fix #22498.

```v
$ cat main.v
struct AStructName {
        foo  int
        anon struct {
                bar int
        }
}

fn main() {
        s := AStructName{5, {}}
        println(s)
}
```

```sh
$ v main.v
main.v:9:22: error: `{}` cannot be used to initialize anonymous structs. Use `struct{}` instead.
    7 | 
    8 | fn main() {
    9 |     s := AStructName{5, {}}
      |                         ~~
   10 |     println(s)
   11 | }
```

### DIFF
--- a/vlib/v/checker/containers.v
+++ b/vlib/v/checker/containers.v
@@ -432,16 +432,21 @@ fn (mut c Checker) map_init(mut node ast.MapInit) ast.Type {
 			node.key_type = info.key_type
 			node.value_type = info.value_type
 			return node.typ
-		} else {
-			if sym.kind == .struct {
-				c.error('`{}` can not be used for initialising empty structs any more. Use `${c.table.type_to_str(c.expected_type)}{}` instead.',
-					node.pos)
+		} else if sym.info is ast.Struct {
+			msg := if sym.info.is_anon {
+				'`{}` cannot be used to initialize anonymous structs. Use `struct{}` instead.'
 			} else {
-				c.error('invalid empty map initialisation syntax, use e.g. map[string]int{} instead',
-					node.pos)
+				'`{}` can not be used for initialising empty structs any more. Use `${c.table.type_to_str(c.expected_type)}{}` instead.'
 			}
-			return ast.void_type
+			c.error(msg, node.pos)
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #22539, opened 2024-10-16
checker: fix if expr with compound conditions (fix #22496)

### LINKED ISSUE #22496: Type narrowing doesn't work in compound conditionals
### Describe the bug

I was expecting type narrowing to occur within compound conditionals, e.g. `if` statements with conditional checks combined using `&&`. These do not work, requiring nested `if` statements to check the types of multiple variables.

### Reproduction Steps

```
type Foo = int | []int
a := Foo([3]) 
if a is []int && a[0] == 3 {
  println("works")
}

error: type `Foo` does not support indexing
   10 | a := Foo([3])
   11 | if a is []int && a[0] == 3 {
      |                   ~~~
   12 |  println("works")
   13 | }
```

### Expected Behavior

Code compiles, and runs as expected.

### Current Behavior

Fails compilation as it doesn't recognize that `a` can only be an of type `[]int` in the second clause of the conditional.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.8 0cf3a44

OS: windows, Microsoft Windows 11 Home v22631 64-bit
Processor: 12 cpus, 64bit, little endian
> [!NOTE]
>

yuyi98:
This PR fix if expr with compound conditions (fix #22496).

- Fix if expr with compound conditions.
- Add test.

```v
type Foo = int | []int

fn main() {
	a := Foo([3])
	if a is []int && a[0] == 3 {
		println('works')
	}
}

PS D:\Test\v\tt1> v run .
works
```

### DIFF
--- a/vlib/v/checker/infix.v
+++ b/vlib/v/checker/infix.v
@@ -1052,6 +1052,9 @@ fn (mut c Checker) autocast_in_if_conds(mut right ast.Expr, from_expr ast.Expr, 
 			c.autocast_in_if_conds(mut right.left, from_expr, from_type, to_type)
 			c.autocast_in_if_conds(mut right.right, from_expr, from_type, to_type)
 		}
+		ast.IndexExpr {
+			c.autocast_in_if_conds(mut right.left, from_expr, from_type, to_type)
+		}
 		else {}
 	}
 }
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #22541, opened 2024-10-16
checker: perfect if expr with compound conditions

yuyi98:
This PR perfect if expr with compound conditions.

- Perfect if expr with compound conditions.
- Add tests.

```v
type Foo = int | []int

fn works() bool {
	arr := [1, 2, 3]
	a := Foo(2)
	if a is int && arr[a] == 3 {
		println('works')
		return true
	}
	return false
}

fn test_if_expr_with_compound_conds() {
	assert works()
}
```
```v
type Foo = int | []int

fn works() bool {
	a := Foo(2)
	if a is int && '${a}' == '2' {
		println('works')
		return true
	}
	return false
}

fn test_if_expr_with_compound_conds() {
	assert works()
}
```
```v
type Foo = int | []int

fn works() bool {
	arr := [1, 2, 3]
	a := Foo(2)
	if a is int && unsafe { arr[..a] == [1, 2] } {
		println('works')
		return true
	}
	return false
}

fn test_if_expr_with_compound_conds() {
	assert works()
}
```

### DIFF
--- a/vlib/v/checker/infix.v
+++ b/vlib/v/checker/infix.v
@@ -1054,6 +1054,19 @@ fn (mut c Checker) autocast_in_if_conds(mut right ast.Expr, from_expr ast.Expr, 
 		}
 		ast.IndexExpr {
 			c.autocast_in_if_conds(mut right.left, from_expr, from_type, to_type)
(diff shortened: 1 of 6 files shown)


spytheman (approved):
Excellent work.

PR #22549, opened 2024-10-17
v: allow checking method comptime signature with `is` operator

### LINKED ISSUE #20928: typed fns do not match correctly in compile time code
### Describe the bug

When checking if a method matches a certain signature in compile time code, it does not return true even when it matches

### Reproduction Steps

```v
module main

struct Struct1 {
	num f64
}

pub type DoThingFunc = fn(f32) bool

fn (mut s Struct1) do_thing(a f32) bool {
	return false
}

fn register[T](){
	$for method in T.methods {
		$if method is DoThingFunc {
			println("method ${method.name} is do thing")
		}$else{
			println("method ${method.name} is not do thing")
		}
	}
}

pub fn main() {
	register[Struct1]()
}

```

### Expected Behavior

It should print`method do_thing is do thing`

### Current Behavior

It prints `method do_thing is not do thing`

### Possible Solution

_No response_

I've also tried `$if method.typ is DoThingFunc` and it doesn't match either

### V version

V 0.4.4 70c277d

V full version: V 0.4.4 1733dab.70c277d

getwd: C:\Users\jcwea\Documents\git\vtest
vexe mtime: 2024-02-29 15:47:47

Git vroot status: weekly.2024.09-6-g70c277d4
.git/config present: true

> [!NOTE]
>

felipensp:
Fix #20928

### DIFF
--- a/vlib/v/ast/types.v
+++ b/vlib/v/ast/types.v
@@ -1604,7 +1604,7 @@ pub fn (t &Table) fn_signature_using_aliases(func &Fn, import_aliases map[string
 		sb.write_string(func.name)
 	}
 	sb.write_string('(')
-	start := int(opts.skip_receiver)
+	start := int(func.is_method && opts.skip_receiver)
 	for i in start .. func.params.len {
 		if i != start {
 			sb.write_string(', ')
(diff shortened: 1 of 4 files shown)


spytheman (approved):
Excellent work.

PR #22551, opened 2024-10-17
time: fix clock_gettime return type

### LINKED ISSUE #8843: Re-declaration of C function in different module is ignored
**V version:** `V 0.2.2 3a65ccd`
**OS:** Void Linux

**What did you do?**
Attempted to declare `C.clock_gettime` with the correct type signature:
```v
fn C.clock_gettime(C.clockid_t, mut C.timespec) C.int
```

Being able to check the return code, eg:
```v
if C.clock_gettime(C.CLOCK_MONOTONIC, mut &ts) != 0 {
    // ...
}
```
 
**What did you see instead?**
V uses the incorrect declaration from the `time` module's `time_nix.c.v`, resulting in:
```
error: non-bool type `void` used as if condition
```

felipensp:
Fix #8843

### DIFF
--- a/vlib/time/time_nix.c.v
+++ b/vlib/time/time_nix.c.v
@@ -50,7 +50,7 @@ pub mut:
 }
 
 // the first arg is defined in include/bits/types.h as `__S32_TYPE`, which is `int`
-fn C.clock_gettime(int, &C.timespec)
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Good find.

PR #22552, opened 2024-10-17
regex, vfmt: optimize receiver with reference

yuyi98:
This PR optimize receiver with reference because struct `RE` and struct `Fmt` are relatively large structures.


spytheman (approved):
Excellent work.

PR #22553, opened 2024-10-17
cgen: fix map_set codegen for sumtype array variant with autofree

### LINKED ISSUE #12008: Compilation error: sumtype array && map with -autofree
**V version:** Latest
**OS:** Ubuntu 20.04

**What did you do?**
`main.v`
```
module main
pub type Value = f64 | []Value
fn main(){
	mut m := map[string]Value
	var := Value([Value(1.0),Value(2.0),Value(3.0)])
	arr := &(var as []Value)
	m['var'] = arr
}
```
v -autofree main.v

**What did you see instead?**
```
error: passing 'Array_main__Value *' (aka 'struct array *') to parameter of incompatible type 'array' (aka 'struct array'); dereference with *
        map_set(&m, &(string[]){_SLIT("var")}, &(main__Value[]) {  array_clone_static_to_depth(arr, 0) });
                                                                                               ^~~
                                                                                               *
/tmp/v/main.12085532964353196708.tmp.c:5711:57: note: passing argument to parameter 'a' here
VV_LOCAL_SYMBOL array array_clone_static_to_depth(array a, int depth) {
                                                        ^
1 error generated.
...
==================

```

felipensp:
Fix #12008

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -3070,7 +3070,17 @@ fn (mut g Gen) gen_clone_assignment(var_type ast.Type, val ast.Expr, typ ast.Typ
 			if typ.share() == .shared_t {
 				g.write('(${shared_styp}*)__dup_shared_array(&(${shared_styp}){.mtx = {0}, .val =')
 			}
-			g.write(' array_clone_static_to_depth(')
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #22554, opened 2024-10-17
sync: C.pthread_cond_t

enghitalo:
```v
fn main() {
	mut queue := new_queue[int]()

	// Start producer and consumer threads
	producer_thread := spawn producer(mut queue)
	consumer_thread := spawn consumer(mut queue)

	// Wait for threads to complete
	producer_thread.wait()
	consumer_thread.wait()

	// Clean up
	queue.destroy()
}
```

### DIFF
--- a/examples/thread_safety/queue.v
+++ b/examples/thread_safety/queue.v
@@ -64,7 +64,6 @@ fn consumer(mut queue ThreadSafeQueue[int]) {
 	}
 }
 
-// Main function
 fn main() {
 	mut queue := new_queue[int]()
 
--- a/examples/thread_safety/queue.v
+++ b/examples/thread_safety/queue.v
@@ -1,0 +1,81 @@
+import sync
+
+struct ThreadSafeQueue[T] {
+mut:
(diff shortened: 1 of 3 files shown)


spytheman:
Why not just use a channel?

enghitalo:
> Why not just use a channel?

How can I make the consumer always listen for changes on the channel?

enghitalo:
It is a better example

```v
import time

fn producer(ch chan int) {
	for i in 1 .. 100 {
		ch <- i
		println('Produced: ${i}')
	}
}

fn consumer(ch chan int) {
	for {
		select {
			item := <-ch {
				println('Consumed: ${item}')
			}

			500 * time.millisecond {
				println('Timeout: No producers were ready within 0.5s')
				break
			}
		}
	}
}

fn main() {
	ch := chan int{cap: 10}

	producer_thread := spawn producer(ch)
	consumer_thread := spawn consumer(ch)

	producer_thread.wait()
	consumer_thread.wait()
}
```

PR #22558, opened 2024-10-17
utf8, readline: clarify UTF-8 vs UTF-32 usage

peppergrayxyz:
This PR only changes the docs and comments. 

It follows up on https://github.com/vlang/v/discussions/22461

The goal of this PR is to make it easier to understand where V uses UTF-8 and where UTF-32

### DIFF
--- a/vlib/builtin/utf8.v
+++ b/vlib/builtin/utf8.v
@@ -3,7 +3,7 @@
 // that can be found in the LICENSE file.
 module builtin
 
-// lenght in bytes of a UTF-8 encoded codepoint
+// lenght in bytes of a UTF-8 encoded codepoint that starts with the byte `b`
 pub fn utf8_char_len(b u8) int {
 	return ((0xe5000000 >> ((b >> 3) & 0x1e)) & 3) + 1
 }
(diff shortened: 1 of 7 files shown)


vlib/builtin/utf8.v
```diff
@@ -3,6 +3,7 @@
 // that can be found in the LICENSE file.
 module builtin
 
+// lenght in bytes of a UTF-8 encoded codepoint
```
spytheman, line 6:
```suggestion
// lenght in bytes of a UTF-8 encoded codepoint that starts with the byte `b`
```

vlib/builtin/utf8.v
```diff
@@ -3,6 +3,7 @@
 // that can be found in the LICENSE file.
 module builtin
 
+// lenght in bytes of a UTF-8 encoded codepoint that starts with the byte `b`
```
spytheman, line 6:
```suggestion
// utf8_char_len returns the length in bytes of a UTF-8 encoded codepoint that starts with the byte `b`
```

spytheman (approved):
Good work.

PR #22560, opened 2024-10-17
encoding.utf8: replace uchar with rune

peppergrayxyz:
`ustring`  has been replaced with `[]rune` in V 0.3 https://github.com/vlang/v/discussions/14895

`uchar`  should therefore be replaced with `rune` in the `encoding.utf8` module:
- added  `rune`-method --><br />deprecated `uchar` -method

public: 
-  `get_rune(s string, index int) rune` --><br />`get_uchar(s string, index int) int`
- `is_rune_punct(r rune) bool` --><br />`is_uchar_punct(uchar int) bool`
- `is_rune_global_punct(r rune) bool` --><br />`is_uchar_global_punct(uchar int) bool`

private:
- `find_punct_in_table(in_code rune, in_table []rune) rune`  --><br />`find_punct_in_table(in_code int, in_table []int) int`
- `unicode_punct_western rune[]` --><br />`unicode_punct_western int[]`
- `unicode_punct rune[]` --><br />`unicode_punct int[]`

### DIFF
--- a/vlib/encoding/utf8/utf8_util.v
+++ b/vlib/encoding/utf8/utf8_util.v
@@ -29,7 +29,7 @@ pub fn len(s string) int {
 
 // get_uchar convert a UTF-8 unicode codepoint in string[index] into a UTF-32 encoded int unicode char
 @[deprecated: 'use `.get_rune(s string, index int)` instead']
-@[deprecated_after: '2024-10-17']
+@[deprecated_after: '2024-11-17']
 pub fn get_uchar(s string, index int) int {
 	return int(get_rune(s, index))
 }
@@ -183,7 +183,7 @@ pub fn is_number(r rune) bool {
 
 // is_uchar_punct return true if the input unicode is a western unicode punctuation
 @[deprecated: 'use `.is_rune_punct(r rune)` instead']
-@[deprecated_after: '2024-10-17']
+@[deprecated_after: '2024-11-17']
 pub fn is_uchar_punct(uchar int) bool {
 	return is_rune_punct(rune(uchar))
 }
@@ -202,7 +202,7 @@ pub fn is_global_punct(s string, index int) bool {
 
 // is_uchar_global_punct return true if the input unicode is a global unicode punctuation
 @[deprecated: 'use `.is_rune_global_punct(r rune)` instead']
-@[deprecated_after: '2024-10-17']
+@[deprecated_after: '2024-11-17']
 pub fn is_uchar_global_punct(uchar int) bool {
 	return is_rune_global_punct(rune(uchar))
 }
--- a/vlib/encoding/utf8/utf8_util.v
+++ b/vlib/encoding/utf8/utf8_util.v
@@ -27,8 +27,15 @@ pub fn len(s string) int {
 	return count
 }
 
-// get_uchar convert a unicode glyph in string[index] into a int unicode char
+// get_uchar convert a UTF-8 unicode codepoint in string[index] into a UTF-32 encoded int unicode char
+@[deprecated: 'use `.get_rune(s string, index int)` instead']
+@[deprecated_after: '2024-10-17']
 pub fn get_uchar(s string, index int) int {
+	return int(get_rune(s, index))
+}
+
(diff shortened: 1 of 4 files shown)


vlib/encoding/utf8/utf8_util.v
```diff
@@ -27,8 +27,15 @@ pub fn len(s string) int {
 	return count
 }
 
-// get_uchar convert a unicode glyph in string[index] into a int unicode char
+// get_uchar convert a UTF-8 unicode codepoint in string[index] into a UTF-32 encoded int unicode char
+@[deprecated: 'use `.get_rune(s string, index int)` instead']
+@[deprecated_after: '2024-10-17']
```
spytheman, line 32:
```suggestion
@[deprecated_after: '2024-11-17']
```

vlib/encoding/utf8/utf8_util.v
```diff
@@ -175,20 +182,34 @@ pub fn is_number(r rune) bool {
 }
 
 // is_uchar_punct return true if the input unicode is a western unicode punctuation
+@[deprecated: 'use `.is_rune_punct(r rune)` instead']
+@[deprecated_after: '2024-10-17']
```
spytheman, line 186:
```suggestion
@[deprecated_after: '2024-11-17']
```

vlib/encoding/utf8/utf8_util.v
```diff
@@ -175,20 +182,34 @@ pub fn is_number(r rune) bool {
...
 
 // is_global_punct return true if the string[index] byte of is the start of a global unicode punctuation
 pub fn is_global_punct(s string, index int) bool {
-	return is_uchar_global_punct(get_uchar(s, index))
+	return is_rune_global_punct(get_rune(s, index))
 }
 
 // is_uchar_global_punct return true if the input unicode is a global unicode punctuation
+@[deprecated: 'use `.is_rune_global_punct(r rune)` instead']
+@[deprecated_after: '2024-10-17']
```
spytheman, line 205:
```suggestion
@[deprecated_after: '2024-11-17']
```

spytheman (approved):
Good work.

peppergrayxyz:
The [docs](https://github.com/vlang/v/edit/master/doc/docs.md#L5701) are somewhat unclear regarding this point:
- What is a good buffer for a future date? Another 6 months?
- When should we use dates and when not? I found that most deprecated functions have a date, but the docs only say it can be used.

I would also create a PR to update the docs with this info (when a date should be used and that a future date should be used here).

spytheman:
I usually use a date, that is close to 1 month ahead in the future, since it is easy to just bump the month number, without having to do date calculations.

spytheman:
The goal of it is to avoid immediate CI breakage, for people with actively developed V projects, i.e. it is assumed that in the next 30 days, they would have compiled their projects manually at least once, and thus they would have seen the deprecation notices and fixed the uses.

spytheman:
> Another 6 months?
A longer buffer like that may also work, for stuff that is harder to change, like the IError interface change or the ?!Type split, but for just renaming functions, it seems excessive to me.

spytheman:
> When should we use dates and when not? I found that most deprecated functions have a date, but the docs only say it can be used.

Adding a date is always preferable in my opinion, for stuff in the vlib modules.

A change with a deprecation date, to me, it is like we are giving an immediate ultimatum to all V users after the merge. It should probably be done for functions that are found to be conceptually broken, and obsoleted by much better functionality, but other than that, I do not see a situation where it would not be just rude to users.

spytheman:
As for documenting the current deprecation policy - that can be done, but perhaps as suggestions, not as rules. Other V projects with other maintainers, can reasonably choose another policy if they wish to 🤷🏻‍♂️ .

spytheman:
There is a tool btw, that can help with finding old deprecations, that we run 1-2 times per year: `v run cmd/tools/show_ancient_deprecations.v 180` for example, if ran now, will show all deprecations before 180 days ago, i.e. before 2024-04-21 .

peppergrayxyz:
Thanks a lot for taking the time and the great explanations! ☺️

PR #22561, opened 2024-10-17
examples: thread_safety

enghitalo:


examples/thread_safety/readme.md
```diff
@@ -0,0 +1,17 @@
+### Run
+
+```sh
+v -prod -autofree ./queue.v -o ./queue.c && \
+gcc ./queue.c -o ./queue.out && \
+./queue.out
```
spytheman, line 6: There is no need for that complexity.

`v -prod -cc gcc ./queue.v` is enough to produce a `queue` executable, compiled with gcc.

spytheman: If you use gcc manually you will also have to add `-lpthread` depending on the platform etc (it is needed on Ubuntu for example).

Using just V, it can take care of that for you.

examples/thread_safety/atomic_counter.v
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Declare the atomic functions
+fn C.atomic_fetch_add_u32(&u32, u32) u32
+fn C.atomic_load_u32(&u32) u32
+
+// Function to increment the atomic counter
+fn increment(atomic_counter &u32) {
+	C.atomic_fetch_add_u32(atomic_counter, 1)
+}
```
spytheman, line 36: `import sync as _` is enough

Issue #22562, opened 2024-10-17
vsh_envbang_test uses wrong v executable


peppergrayxyz:
### Describe the bug

`vsh_envbang_test` is testing the wrong executable

### Reproduction Steps

This snippet follows the [Example Workflow for Contributing](https://github.com/vlang/v/blob/master/CONTRIBUTING.md#example-workflow-for-contributing):
- System version of v for compiling v code in `~\v`
- Dev repo for  working on the compiler itself in `~\nv`

```bash
cd ~
git clone https://github.com/vlang/v
cd v
make
sudo ./v symlink

cd ~
git clone https://github.com/vlang/v nv
cd nv
make
make check
```
`vsh_envbang_test`  will check if `~/v/v` (as fetched from env) works, but not if  `~/nv/v` works, which we want to test.

### Expected Behavior

 the binary being tested by `m̀ake check` is the binary built by `make`

### Current Behavior

 the binary being tested by `m̀ake check` is the binary from `$PATH`

### Possible Solution

use `os.quoted_path(@VEXE)` as described here: https://docs.vlang.io/testing.html

### V version

0.4.8 8054808

### Environment details (OS name and version, etc.)

linux
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #22563, opened 2024-10-17
v: fix vsh_envbang_test

peppergrayxyz:
fix https://github.com/vlang/v/issues/22562

### DIFF
--- a/vlib/v/pref/vsh_envbang_test.v
+++ b/vlib/v/pref/vsh_envbang_test.v
@@ -1,13 +1,15 @@
 import os
 import rand
 
+const vexe = @VEXE
+
 fn testsuite_begin() {
 	$if windows {
 		skip_test('windows does not support envbang lines')
 	}
 	// Make sure, that we will be using the local v, instead of the globally installed one,
 	// for the rest of the tests here:
-	local_v_path := os.dir(@VEXE)
+	local_v_path := os.dir(vexe)
 	old_path := os.getenv('PATH')
 	os.setenv('PATH', '${local_v_path}:${old_path}', true)
 	println('Changed PATH: ${os.getenv('PATH')}')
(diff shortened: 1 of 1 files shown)


vlib/v/pref/vsh_envbang_test.v
```diff
@@ -2,7 +2,7 @@ import os
 import rand
 
 fn test_envbang_script_runs() {
-	env_location := '/usr/bin/env'
+	env_location := os.quoted_path(@VEXE)
```
spytheman, line 5: That test is explicitly named like that: `test_envbang_script_runs`, because it tests,
that scripts made with `#!/usr/bin/env v`  will continue to work.

This should stay as it is.

spytheman: What you can do though, if you want to make it test the local v executable, is to modify the PATH env variable, before the test, so that it will add the path of the folder, where the local executable is placed, *before* every other path in the list.

Then `find_abs_path_of_executable` will find the local executable.
The generated script later, when run, will also use the local executable, because `/usr/bin/env` also will use the PATH env variable to locate it.

spytheman: Something like this should work, without other modifications to the existing file:

```v
fn testsuite_begin() {
    local_v_path := os.dir(@VEXE)
    $if windows {
        skip_test('windows does not support envbang lines')
    }
    os.setenv('PATH', '${local_v_path}:${old_path}', true)
}
```

peppergrayxyz: Thanks for the explanation, I did not differentiate between shebang and envbang. Also thanks for picking this up and fixing the issue!

spytheman: Thank you for spotting it.

PR #22568, opened 2024-10-18
doc: explain deprecation process

peppergrayxyz:
Be more verbose about how deprecation works. Document learnings and missing hints discovered through https://github.com/vlang/v/pull/22560


spytheman (approved):
Thank you.

PR #22571, opened 2024-10-18
examples: fix some of the instructions in `examples/thread_safety/`

enghitalo:


examples/thread_safety/queue.v
```diff
@@ -16,33 +16,40 @@ that data is safely passed between threads.
...
 			item := <-ch {
 				println('Consumed: ${item}')
 			}
 			500 * time.millisecond {
 				println('Timeout: No producers were ready within 0.5s')
+				mtx.unlock()
 				break
 			}
 		}
+		mtx.unlock()
```
spytheman, line 43: This is not needed; reading from the channel is blocking, and you may want to have multiple consumers.

doc/docs.md
```diff
@@ -7134,17 +7134,8 @@ To use these functions the C header for the used OS has to be included and the f
...
-} $else {
-	#include "@VEXEROOT/thirdparty/stdatomic/nix/atomic.h"
-}
 
-// declare functions we want to use - V does not parse the C header
-fn C.atomic_store_u32(&u32, u32)
-fn C.atomic_load_u32(&u32) u32
-fn C.atomic_compare_exchange_weak_u32(&u32, &u32, u32) bool
-fn C.atomic_compare_exchange_strong_u32(&u32, &u32, u32) bool
+import sync as _
```
spytheman, line 7139: The docs should stay as they are. They show a more manual low level of interoperating with C, and V's atomics should be improved instead (that is just not a current priority, since it also involves upgrading tcc and potentially adding support for atomics to it too).

examples/thread_safety/queue.v
```diff
@@ -16,33 +16,40 @@ that data is safely passed between threads.
...
 
-fn producer(ch chan int) {
+fn producer(ch chan int, mut mtx sync.Mutex) {
 	for i in 1 .. 100 {
+		mtx.lock()
 		ch <- i
 		println('Produced: ${i}')
+		mtx.unlock()
 	}
 }
```
spytheman, line 28: The mutex is not needed here. Writing to the channel is buffered for the first 100 slots (the capacity of the channel is 100 items), and after that, if the consumer is too slow, writing will block.

enghitalo: Removing mutex from producer, I will have `==1182242== ERROR SUMMARY: 1096 errors from 3 contexts (suppressed: 2 from 2)` from Valgrind

examples/thread_safety/queue.v
enghitalo: This will give me some erros when I run 
```sh
v -prod -gc none -cc gcc ./queue.v && valgrind --tool=helgrind ./queue
```

spytheman: That should be investigated, not just hidden, by slapping mutexes everywhere.

examples/thread_safety/queue.v
spytheman: Same - understand why that is (for example, using println may not be a good idea in such a code, since writing to stdout is inherently racy), and fix the actual problem, not just add global mutexes like that, that practically eliminate the actual benefit of using threads in the first place.

PR #22578, opened 2024-10-18
cgen: fix codegen for if with struct init that contains option type

### LINKED ISSUE #22577: Trying to avoid writing too many returns in if/else or match.
### Discussed in https://github.com/vlang/v/discussions/22565

<div type='discussions-op-text'>

<sup>Originally posted by **jorgeluismireles** October 17, 2024</sup>
The combination of objects with options and created within a function with a **single** `return` outside cases `if/else` or `match` causes problems. The program:
```v
struct Number {
	natural ?string
	integer string
}

fn number(n int) Number {
	return if n >= 0 { // a single return here
		Number{ natural:'${n}', integer:'${n}' }
	} else {
		Number{ integer:'${n}' }
	}
}

fn main() {
	for n in [1,0,-1] {
		println('${number(n)}')
	}
}
```

https://play.vlang.io/p/78094800a4

The problem:
```
Running code...
/tmp/v_60000/../../../../../../box/code.v:7: error: ')' expected (got "_t2")
builder error: 
==================
Exited with error status 1
Please try again.

```
A workaround is to use a `return` for **each case** (apart changing the `?string` by `string`):
```v
fn number(n int) Number {
	if n >= 0 {
		return Number{ natural:'${n}', integer:'${n}' } // a return here
	} else {
		return Number{ integer:'${n}' } // another return here
	}
}
```
The expected output:
```
Number{
    natural: Option('1')
    integer: '1'
}
Number{
    natural: Option('0')
    integer: '0'
}
Number{
    natural: Option(none)
    integer: '-1'
}
```

I have found this problem more than once trying to avoid writing "too many" `return`s in functions with a returning `match` with too many cases and the when the returned object is not *trivial*.

But now I wonder if is worth to solve the *issue* mentioned above for this particular case and then later more complex objects could cause new problems, or just encourage to program "the easy way" as in the workaround.

Thoughts?</div>
> [!NOTE]
>

felipensp:
Fix #22577

### DIFF
--- a/vlib/v/gen/c/if.v
+++ b/vlib/v/gen/c/if.v
@@ -157,9 +157,7 @@ fn (mut g Gen) need_tmp_var_in_expr(expr ast.Expr) bool {
 				}
 			}
 			sym := g.table.sym(expr.typ)
-			if sym.info is ast.Struct {
-				return sym.info.fields.any(it.typ.has_flag(.option))
-			}
+			return sym.info is ast.Struct && sym.info.has_option
 		}
 		else {}
 	}
--- a/vlib/v/gen/c/if.v
+++ b/vlib/v/gen/c/if.v
@@ -156,6 +156,10 @@ fn (mut g Gen) need_tmp_var_in_expr(expr ast.Expr) bool {
 					return true
 				}
 			}
+			sym := g.table.sym(expr.typ)
+			if sym.info is ast.Struct {
+				return sym.info.fields.any(it.typ.has_flag(.option))
+			}
 		}
 		else {}
 	}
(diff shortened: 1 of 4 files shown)


vlib/v/gen/c/if.v
```diff
@@ -156,6 +156,10 @@ fn (mut g Gen) need_tmp_var_in_expr(expr ast.Expr) bool {
 					return true
 				}
 			}
+			sym := g.table.sym(expr.typ)
+			if sym.info is ast.Struct {
+				return sym.info.fields.any(it.typ.has_flag(.option))
```
spytheman, line 161: imho cache the information in another field in `ast.Struct`, about whether there are any option fields in a struct, during parsing, so that it is done just once, and then just use it here, instead of doing the loop for `any` potentially many times for each expression.

felipensp: Yeah, makes sense.

spytheman (approved):
Excellent work.

PR #22580, opened 2024-10-18
optimizing: builtin sym marking on parser-time

felipensp:
This PR aims add optimizations in as sym is checked as builtin.

image

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -3613,9 +3613,12 @@ fn (mut g Gen) expr(node_ ast.Expr) {
 		ast.LambdaExpr {
 			if node.call_ctx != unsafe { nil } {
 				save_cur_concrete_types := g.cur_concrete_types
+				save_cur_unwrap_cache := g.table.cur_unwrap_cache.clone()
 				g.cur_concrete_types = node.call_ctx.concrete_types
+				g.table.cur_unwrap_cache.clear()
 				g.gen_anon_fn(mut node.func)
 				g.cur_concrete_types = save_cur_concrete_types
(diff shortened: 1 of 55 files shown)


esquerbatua:
Great job!

vlib/v/gen/c/cgen.v
```diff
@@ -1669,13 +1670,7 @@ pub fn (mut g Gen) write_interface_typedef(sym ast.TypeSymbol) {
 }
 
 pub fn (mut g Gen) write_interface_typesymbol_declaration(sym ast.TypeSymbol) {
-	if sym.info !is ast.Interface {
-		return
-	}
```
spytheman, line 1674: why delete this?

spytheman: you will get garbage if you pass a sym, that is not an interface without it, because of the next line that casts it unconditionally

spytheman:
I could not reproduce the result on my machine :-| .
image

Note the standard deviations for the results.

spytheman:
@felipensp, can you please split the `Immediate identical type checking` part in a separate PR?

vlib/v/gen/c/cgen.v
felipensp: Yes, but I ensure that this method is only called with proper Interface type. i.e. I moved the check to the caller.

felipensp:
> @felipensp, can you please split the `Immediate identical type checking` part in a separate PR?
> 
> I think it can be beneficial on its own, and it may be masking the effects of the other change (which is more complex, and I suspect may be slowing things).

Ok done.

spytheman:
Thank you.

PR #22589, opened 2024-10-20
cgen: cleanup in fixed_array_init()

yuyi98:
This PR cleanup in fixed_array_init().

### DIFF
--- a/vlib/v/gen/c/array.v
+++ b/vlib/v/gen/c/array.v
@@ -86,16 +86,17 @@ fn (mut g Gen) fixed_array_init(node ast.ArrayInit, array_type Type, var_name st
 	defer {
 		g.inside_lambda = prev_inside_lambda
 	}
+	array_info := array_type.unaliased_sym.array_fixed_info()
 	if node.has_index {
 		past := g.past_tmp_var_from_var_name(var_name)
 		defer {
 			g.past_tmp_var_done(past)
 		}
 
-		ret_typ := g.typ(node.typ)
-		elem_typ := g.typ(node.elem_type)
+		ret_typ_str := g.typ(node.typ)
+		elem_typ_str := g.typ(node.elem_type)
 		if var_name == '' {
-			g.write('${ret_typ} ${past.tmp_var} =')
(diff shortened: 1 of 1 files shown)


vlib/v/gen/c/array.v
```diff
@@ -86,16 +86,17 @@ fn (mut g Gen) fixed_array_init(node ast.ArrayInit, array_type Type, var_name st
...
+	array_info := array_type.unaliased_sym.array_fixed_info()
 	if node.has_index {
 		past := g.past_tmp_var_from_var_name(var_name)
 		defer {
 			g.past_tmp_var_done(past)
 		}
 
-		ret_typ := g.typ(node.typ)
-		elem_typ := g.typ(node.elem_type)
+		ret_typ_str := g.typ(node.typ)
```
spytheman, line 96: What do you think about renaming `g.typ` to `g.stype` or something else that suggests that the result is a string?

yuyi98: I agree.

vlib/v/gen/c/array.v
```diff
@@ -136,26 +136,12 @@ fn (mut g Gen) fixed_array_init(node ast.ArrayInit, array_type Type, var_name st
...
-	mut stmt_str := ''
-	mut tmp_var := ''
-	if need_tmp_var {
-		tmp_var = g.new_tmp_var()
-		stmt_str = g.go_before_last_stmt()
-		g.empty_line = true
-
-		ret_typ := g.typ(node.typ)
-		g.write('${ret_typ} ${tmp_var} = ')
-	}
```
spytheman, line 149: hm, it is great to delete that, but are you sure it will not be needed?
I do not understand at first glance, what made it obsolete.

yuyi98: `node.exprs.len == 0` can not inside fixed array init.

spytheman: Thanks for explaining it.

spytheman (approved):
Excellent work.

PR #22579, opened 2024-10-18
cgen: fix array contains codegen for fixed array

felipensp:
Fix #22559

PR #22590, opened 2024-10-20
cgen: improve fixed array literal in operation and index calls (related #22579)

yuyi98:
This PR improve fixed array literal in operation and index calls (related #22579).

- Improve fixed array literal in operation and index calls.
- Add tests.

```v
fn test_array_of_fixed_array_in_op() {
	mut a := [][2]int{}
	a << [1, 2]!
	println([1, 2]! in a)
	ret := [1, 2]! in a
	println(ret)
	assert ret
}

fn test_fixed_array_of_fixed_array_in_op() {
	mut a := [2][2]int{}
	a[0] = [1, 2]!
	println([1, 2]! in a)
	ret := [1, 2]! in a
	println(ret)
	assert ret
}

fn test_array_of_fixed_array_index() {
	mut a := [][2]int{}
	a << [1, 2]!
	println(a.index([1, 2]!))
	ret := a.index([1, 2]!)
	println(ret)
	assert ret == 0
}
```


spytheman (approved):
Excellent work.

PR #22592, opened 2024-10-20
ci: benchmark footprint

enghitalo:


spytheman:
Benchmarking on the CI tends to produce unreliable results, since we are using the free tier runners of Github Actions.

I am skeptical about the usefulness of adding it, but I guess we can try it for several days/weeks.

PR #22594, opened 2024-10-20
ci: benchmark footprint

enghitalo:


vlib/v/tests/bench/bench_json_vs_json2_out.md
spytheman, line 1: No, those results are not reproducible and specific to your machine.
More over they would need to be updated all the time, as work on x.json2 continues.

Generated artifacts, with variable data, that can be easily produced locally *when needed*, should not be checked in version control.

vlib/x/json2/decoder2/tests/bench_out.md
spytheman, line 1: same - this file should be removed

PR #22606, opened 2024-10-20
ci: fix failing jobs after 4ed9d13

spytheman:


koplenov:
the fix is broken work veb.tr in templates))0))0
https://t.me/vlang_rus/54977

Issue #17003, opened 2023-01-17
vlib documents: Broken links due to missing READMEs


km19809:
### Describe the issue

I can click `db` or `encoding` from the sidebar, but it leads to the same, current page.

![image](https://user-images.githubusercontent.com/20869932/212830491-afb23452-07af-4d0e-b4b9-8e10391d07db.png)
Image: The link refers to the same page.

It is because there is no README in `vlib/db` or `vlib/encoding`.
However, meaningless anchors may interrupt the browsing experience.

I think there are some possible options:
1. Provide READMEs for modules.
2. Eliminate the anchor tag for missing documents.
3. Link the anchor to a default error page.

Which one is the most admirable?
Or, is the issue an intended behavior?

### Links

https://modules.vlang.io/

spytheman:
Fixed by #22636 .

PR #22636, opened 2024-10-23
docs: fix broken links by adding missing READMEs

alex-semenyuk:
As mentioned at #17003 missing README broke the links so add some short READMEs for such cases

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzE5NjAxM2YwMmQ4YTAzZmIyZmViNmMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.ZiJl2cEHpP-UKUMKoLzSUjGxYpC3uRTB4jxdeAMIuow">Huly&reg;: <b>V_0.6-21085</b></a></sub>

### DIFF
--- a/vlib/coroutines/README.md
+++ b/vlib/coroutines/README.md
@@ -1,0 +1,3 @@
+## Description
+
+`coroutines` is a namespace for a wrapper around third party/photon.
--- a/vlib/db/README.md
+++ b/vlib/db/README.md
@@ -1,0 +1,4 @@
+## Description
+
+`db` is a namespace that contains several useful modules 
+for operating with databases (SQLite, MySQL, MSQL, etc.)
--- a/vlib/v2/README.md
+++ b/vlib/v2/README.md
@@ -1,0 +1,3 @@
+## Description
+
+`v2` is a namespace containing a new version of the compiler (still in development).


JalonSolov:
The `vlib/v2/README.md` is incorrect, at the least.  It is code for an intermediate change to the V compiler, not ready for regular usage.  Not sure about the status of the others...

spytheman:
All of them are incorrect and misleading. I do not know what LLM generated them, but it is not appreciated.

Even an empty readme, or just a readme that has an explicit "placeholder" text in it, is better.

alex-semenyuk:
if you're well know what they're doing, can make your suggestion that fix this right away   

For example, as for v2 something like?

`v2` provides code for an intermediate change to the V compiler, not ready for regular usage.

spytheman:
Both vlib/v/ and vlib/db/ are just container modules. They do not do anything by themselves, but only provide an organization structure for the modules under them.

As for vlib/coroutines/ - it is just a wrapper around thirdparty/photon .
It is used by the compiler, to implement `go f()`. That is all that can be safely documented about it. The reason for that, since it is still experimental and can change, if we decide to use another green threads library instead of photon for example, or integrate changes that make it work on windows etc.

spytheman:
The README for vlib/v, starts with this:
```md
## Description

`v` is a namespace for all of the V compiler modules.
```

I think something similar can be done for `vlib/v2/README.md` and `vlib/db/README.md` .

alex-semenyuk:
> `coroutines` is a namespace for a wrapper around third party/photon.

> `v2` is a namespace for an intermediate change to the V compiler, not ready for regular usage.

> `db` is a namespace for interacting with various databases modules.

How about something like this

StunxFS:
`coroutines` is not a namespace, it is a module that works as a wrapper for `photon`.

`v2` is a namespace containing a new version of the compiler (still in development).

`db` is a namespace that contains several useful modules for operating with databases (SQLite, MySQL, MSQL, etc.)

spytheman:
The new versions seem fine to me.

alex-semenyuk:
thanks, updated

spytheman (approved):
Excellent work.

PR #22647, opened 2024-10-24
encoding.utf8: fix is_punct func

### LINKED ISSUE #22646: `!` is not punct char
### Describe the bug

Possible bug in `encoding.utf8` `is*punct` functions, `!` recognized as non-punct char. `ispunct` from `libc` returns true for all chars.

### Reproduction Steps

Run Code:

```v
module main
import encoding.utf8 as u8

fn check_char(s &string) int
{
    println("${*s}")
    println("is punct ?")
    for i, ch in s.runes()
    {
         is_p := u8.is_punct(ch.str(), 0)
	 okch := ch.str()
	 println("${if is_p {' is punct'} else {'not punct'}}, ${u8.get_uchar(s, i)}, ${i} ${okch}")
    }
    return 0
}

fn main()
{
    a := "!.,:;"
    println("${typeof(a).name}\n${check_char(a)}\n")
}
```

### Expected Behavior

```
!.,:;
is punct ?
 is punct, 33, 0 !
 is punct, 46, 1 .
 is punct, 44, 2 ,
 is punct, 58, 3 :
 is punct, 59, 4 ;
string
0
```

### Current Behavior

**Output:**

```
!.,:;
is punct ?
not punct, 33, 0 !
 is punct, 46, 1 .
 is punct, 44, 2 ,
 is punct, 58, 3 :
 is punct, 59, 4 ;
string
0
```

### Possible Solution

Not discovered yet.

_No response_

### V version

4.4, 4.5, 4.8

Manjaro linux, kernel 6.4

> [!NOTE]
>

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzFhOTM4ZmYwMmQ4YTAzZmIzM2EwOGMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.4fcuz0wsWy03FVl_DmiswhIxFxSm-WsRlg4OfiN-ZDY">Huly&reg;: <b>V_0.6-21095</b></a></sub>

felipensp:
Fix #22646

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzFhYjZjY2YwMmQ4YTAzZmIzM2EyZTgiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.ubRUp4TTVcm0iP23ElDdoLMYClJizBgQKioKIV2n0ZU">Huly&reg;: <b>V_0.6-21096</b></a></sub>

### DIFF
--- a/vlib/encoding/utf8/utf8_is_punct_test.v
+++ b/vlib/encoding/utf8/utf8_is_punct_test.v
@@ -1,21 +1,33 @@
 module main
 
-import encoding.utf8 as u8
+import encoding.utf8
 
-fn check_char(s &string) int {
-	for i, ch in s.runes() {
-		is_p := u8.is_punct(ch.str(), 0)
-		okch := ch.str()
-		if !is_p {
-			println('${if is_p { ' is punct' } else { 'not punct' }}, ${u8.get_rune(s,
-				i)}, ${i} ${okch}')
-			assert false
-		}
-	}
-	return 0
+fn test_true() {
+	assert utf8.is_punct('!', 0)
+	assert utf8.is_rune_punct(`!`)
+	assert utf8.is_punct('.', 0)
+	assert utf8.is_rune_punct(`.`)
+	assert utf8.is_punct(',', 0)
+	assert utf8.is_rune_punct(`,`)
+	assert utf8.is_punct(';', 0)
+	assert utf8.is_rune_punct(`;`)
+	assert utf8.is_punct("'", 0)
+	assert utf8.is_rune_punct(`'`)
+	assert utf8.is_punct(',', 0)
+	assert utf8.is_rune_punct(`,`)
+	assert utf8.is_punct('/', 0)
+	assert utf8.is_rune_punct(`/`)
+	assert utf8.is_punct('*', 0)
+	assert utf8.is_rune_punct(`*`)
 }
 
-fn test_main() {
-	a := "!.,:;',/*"
-	assert check_char(a) == 0
+fn test_false() {
+	assert !utf8.is_punct('a', 0)
+	assert !utf8.is_rune_punct(`a`)
+	assert !utf8.is_punct('ç', 0)
+	assert !utf8.is_rune_punct(`ç`)
+	assert !utf8.is_punct('á', 0)
+	assert !utf8.is_rune_punct(`á`)
+	assert !utf8.is_punct('-', 0)
+	assert !utf8.is_rune_punct(`-`)
 }
--- a/vlib/encoding/utf8/utf8_is_punct_test.v
+++ b/vlib/encoding/utf8/utf8_is_punct_test.v
@@ -1,0 +1,21 @@
+module main
+
+import encoding.utf8 as u8
+
+fn check_char(s &string) int {
+	for i, ch in s.runes() {
+		is_p := u8.is_punct(ch.str(), 0)
+		okch := ch.str()
+		if !is_p {
+			println('${if is_p { ' is punct' } else { 'not punct' }}, ${u8.get_rune(s,
+				i)}, ${i} ${okch}')
+			assert false
+		}
+	}
+	return 0
+}
+
+fn test_main() {
+	a := "!.,:;',/*"
+	assert check_char(a) == 0
+}
--- a/vlib/encoding/utf8/utf8_util.v
+++ b/vlib/encoding/utf8/utf8_util.v
@@ -190,7 +190,7 @@ pub fn is_uchar_punct(uchar int) bool {
 
 // is_rune_punct return true if the input unicode is a western unicode punctuation
 pub fn is_rune_punct(r rune) bool {
-	return find_punct_in_table(r, unicode_punct_western) != -1
+	return find_punct_in_table(r, unicode_punct_western) != rune(-1)
 }
 
 // Global
@@ -209,7 +209,7 @@ pub fn is_uchar_global_punct(uchar int) bool {
 
 // is_rune_global_punct return true if the input unicode is a global unicode punctuation
 pub fn is_rune_global_punct(r rune) bool {
-	return find_punct_in_table(r, unicode_punct) != -1
+	return find_punct_in_table(r, unicode_punct) != rune(-1)
 }
 
 // Private functions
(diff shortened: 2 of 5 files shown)


vlib/encoding/utf8/utf8_util.v
```diff
@@ -190,7 +190,7 @@ pub fn is_uchar_punct(uchar int) bool {
 
 // is_rune_punct return true if the input unicode is a western unicode punctuation
 pub fn is_rune_punct(r rune) bool {
-	return find_punct_in_table(r, unicode_punct_western) != 0
+	return find_punct_in_table(r, unicode_punct_western) != -1
```
spytheman, line 193:
```suggestion
	return find_punct_in_table(r, unicode_punct_western) != rune(-1)
```

vlib/encoding/utf8/utf8_util.v
```diff
@@ -209,7 +209,7 @@ pub fn is_uchar_global_punct(uchar int) bool {
 
 // is_rune_global_punct return true if the input unicode is a global unicode punctuation
 pub fn is_rune_global_punct(r rune) bool {
-	return find_punct_in_table(r, unicode_punct) != 0
+	return find_punct_in_table(r, unicode_punct) != -1
```
spytheman, line 212:
```suggestion
	return find_punct_in_table(r, unicode_punct) != rune(-1)
```

vlib/encoding/utf8/utf8_util.v
spytheman: I am not sure, what is the case, where the function will return -1 or 0xffff_ffff ?

vlib/encoding/utf8/utf8_is_punct_test.v
```diff
@@ -0,0 +1,21 @@
+module main
+
+import encoding.utf8 as u8
```
spytheman, line 3: That is a bit misleading, since `u8` is also a type name, and saves only 2 characters. Please use the full module name `utf8` below, and remove the module alias.

vlib/encoding/utf8/utf8_is_punct_test.v
```diff
@@ -0,0 +1,21 @@
...
+			println('${if is_p { ' is punct' } else { 'not punct' }}, ${u8.get_rune(s,
+				i)}, ${i} ${okch}')
+			assert false
+		}
+	}
+	return 0
+}
+
+fn test_main() {
+	a := "!.,:;',/*"
```
spytheman, line 19: Please do write several direct assertions, for `utf8.is_rune_punct` or `utf8.find_punct_in_table` (since they are changed), rather than a loop like this - it only obfuscates the intent of the tests, without contributing anything to the diagnostic role of a potentially failing regression test.

spytheman: Please, write also negative tests in addition to the positive ones (for the same functions), i.e. not just for punctuation runes, but for runes of normal letters and numbers too.

vlib/encoding/utf8/utf8_is_punct_test.v
```diff
@@ -0,0 +1,21 @@
...
+
+import encoding.utf8 as u8
+
+fn check_char(s &string) int {
+	for i, ch in s.runes() {
+		is_p := u8.is_punct(ch.str(), 0)
+		okch := ch.str()
+		if !is_p {
+			println('${if is_p { ' is punct' } else { 'not punct' }}, ${u8.get_rune(s,
+				i)}, ${i} ${okch}')
```
StunxFS, line 11:
```suggestion
			println('${if is_p { ' is punct' } else { 'not punct' }}, ${ch}, ${i} ${okch}')
```

why not just use `ch`, instead of using `utf8.get_rune(s, i)`?

AKonia: @StunxFS This test based on my sample in bug report. I've started to learn V not long ago.

StunxFS: oh ok

vlib/encoding/utf8/utf8_is_punct_test.v
felipensp: Yeah, totally. But I just copy&paste the report case.

vlib/encoding/utf8/utf8_is_punct_test.v
felipensp: Done @spytheman

vlib/encoding/utf8/utf8_is_punct_test.v
felipensp: > why not just use `ch`, instead of using `utf8.get_rune(s, i)`?

Thanks, I've done a cleanup in the test.

vlib/encoding/utf8/utf8_is_punct_test.v
```diff
@@ -0,0 +1,33 @@
+module main
+
+import encoding.utf8
+
+fn test_true() {
```
spytheman, line 5:
```suggestion
fn test_is_punct() {
```
that will have all the cases for that specific public function will be a better scheme, so that if an assertion fails, the message will be more informative (the test function name is included in the failure)

spytheman: However, the assertions themselves are already much better and direct, so the test function name does not matter as much 🤷🏻‍♂️ .

spytheman (approved):
Good work.

PR #22684, opened 2024-10-28
orm: fix orm.Connection only allowing immutable Connections

daniel-le97:
This PR allows implementing the orm.Connection interface with methods that take in mutable or immutable Connections

Before the change, the following would result in an error similar to:

 error: `Database` incorrectly implements method `select` of interface `orm.Connection`: expected `orm.Connection` which is immutable, not `mut &Database`

```v
fn (mut db Database) drop(table string) ! {
	query := 'DROP TABLE ${table};'
	db.query(query)!
}
```

This can be tested with

```bash
v -stats test ./vlib/orm/orm_mut_connection_test.v
```

### DIFF
--- a/vlib/orm/orm.v
+++ b/vlib/orm/orm.v
@@ -193,6 +193,7 @@ pub:
 // Every function without last_id() returns an optional, which returns an error if present
 // last_id returns the last inserted id of the db
 pub interface Connection {
+mut:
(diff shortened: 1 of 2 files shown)


spytheman (approved):

spytheman:
Good work.

PR #22689, opened 2024-10-28
checker: fix selector with prefixed `&` structinit

### LINKED ISSUE #22688: defect &Object{}.a
**V doctor:**

```
V full version: V 0.4.8 df4c6ac.5d76ee4
OS: linux, Debian GNU/Linux 12 (bookworm) (VM)
Processor: 1 cpus, 64bit, little endian, DO-Premium-AMD

getwd: /home/user/github.com/markfirmware/vlister/src
vexe: /home/user/github.com/vlang/v/v
vexe mtime: 2024-10-28 20:57:38

vroot: OK, value: /home/user/github.com/vlang/v
VMODULES: OK, value: /home/user/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.39.5
Git vroot status: weekly.2024.43-58-g5d76ee46
.git/config present: true

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg field_ref_main.v`

```v

struct Object {
  a int
}

fn main() {
  field := &Object{}.a
  println('${field}')
}
```

compile ok

**What did you see instead?**

```
cc: /tmp/v_1000/field_ref_main.01JBAEQVFERFTWSV304MH6SME1.tmp.c:13476: error: lvalue expected
builder error: 
==================
```

> [!NOTE]
>

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzFmZmIyMjI3Y2M3NjgxYzYyMzI0NmIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.T6ciPZpn4l1MWrj15YK0_34tAYsjqfTX-GzrFKp80f8">Huly&reg;: <b>V_0.6-21137</b></a></sub>

felipensp:
Fix #22688

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzIwMDAzOTI3Y2M3NjgxYzYyMzI2MDgiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.K1pgYXZxl9eZYzOn9o2B7CSY2d3VRGexpLCfB6bMhvo">Huly&reg;: <b>V_0.6-21138</b></a></sub>


vlib/v/tests/selector_struct_init_amp_test.v
```diff
@@ -0,0 +1,8 @@
+struct Object {
+	a int = 123
+}
+
+fn test_main() {
+	field := &Object{}.a
```
StunxFS, line 6: Hm, I think this might be confusing. At first glance, it looks like `field` takes a reference to field `a`.

spytheman: It creates an object instance on the heap, and then gets the `a` field from it.

I think it can not be reduced, to another less confusing form, since that is what the issue is - https://github.com/vlang/v/issues/22688, that leads to a cgen error.
The test is pretty much just the code sample from the issue.

spytheman: Do you think we should have a checker error for this instead?

That also seems fine to me, since I can not see a useful situation (beside code generation/translation from another language), where you would want to both create a heap instance, and then get a field from it immediately, since all the rest of the instance will be discarded/become garbage right after that.

StunxFS: @spytheman I think there should be an error checker, instead of allowing such code to compile.

If someone wants to use a value on the heap, they can use `malloc` or assign the created instance to a temporary variable.

felipensp: > On the other hand, it is valid V code, using existing V features, just in a weird way, so it should not produce a cgen error.

Yeah, to me it is just someone can do in another lang like `new foo()->prop`

medvednikov:
I think it should result in checker error.

The V compiler promotes writing correct code.

phcreery:
Would enforcing `(&Object{}).a` (`int`) vs `&(Object{}.a)` (`&int`) make it less confusing?

felipensp:
image
Done. 👍🏻

medvednikov:
Nice :)

PR #22696, opened 2024-10-30
flag: add missing short flag match in `flag.to_struct`, add test

larpon:


vlib/flag/flag_to_edge_case_1_test.v
```diff
@@ -0,0 +1,25 @@
...
+	dump_usage   bool @[long: help; short: h; xdoc: 'Show this help message and exit']
+	show_version bool @[long: version; xdoc: 'Output version information and exit']
+pub mut:
+	input  string   @[tail]
+	output string   @[short: o; xdoc: 'Path to output (dir/file)']
+	assets []string @[short: a; xdoc: 'Asset dir(s) to include in build']
+}
+
+fn test_edge_case() {
+	aio, no_matches := flag.to_struct[AppImageOptions](edge_case, skip: 1)!
```
spytheman, line 17: There is no need to extract a constant, since it is not used several times.
Moving it back to the test will make it easier to read/maintain.

PR #22715, opened 2024-10-31
net: fix resolving of `::1:48872` (pass the `*should_listen_on_both_ipv4_and_ipv6_by_default_test.v` tests)

spytheman:


felipensp:
Good catch! Thanks for repairing it. :-)

PR #22740, opened 2024-11-02
v: allow `map[k]()?` and `map[k]()()`

felipensp:
This PR allows to do:
`map['f']()?`
`map['f'](arg)(arg2)`

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzI2NDEyMzI3Y2M3NjgxYzYyNWY4ZmEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.nTQFvmkXUbnl2WYFYAiLrms4zDEXNpG758JWjc7jPHQ">Huly&reg;: <b>V_0.6-21187</b></a></sub>

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -864,16 +864,47 @@ fn (mut g Gen) call_expr(node ast.CallExpr) {
 		} else if node.or_block.kind == .absent {
 			g.expr(node.left)
 		}
-	} else if node.left is ast.IndexExpr && node.name == '' {
-		g.is_fn_index_call = true
-		g.expr(node.left)
-		g.is_fn_index_call = false
+	} else if !g.inside_curry_call && node.left is ast.IndexExpr && node.name == '' {
+		if node.or_block.kind == .absent {
+			old_is_fn_index_call := g.is_fn_index_call
+			g.is_fn_index_call = true
+			g.expr(node.left)
+			g.is_fn_index_call = old_is_fn_index_call
+		} else {
+			line := g.go_before_last_stmt()
+			g.empty_line = true
+
+			left_typ := g.table.value_type(node.left.left_type)
+			tmp_res := g.new_tmp_var()
+			fn_sym := g.table.sym(left_typ).info as ast.FnType
+			fn_type := g.fn_var_signature(fn_sym.func.return_type, fn_sym.func.params.map(it.typ),
+				tmp_res)
+
+			old_is_fn_index_call := g.is_fn_index_call
+			g.is_fn_index_call = true
+			g.write('${fn_type} = ')
+			g.expr(node.left)
+			g.is_fn_index_call = old_is_fn_index_call
+			g.writeln(';')
+
+			tmp_res2 := g.new_tmp_var()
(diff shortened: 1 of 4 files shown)


vlib/v/parser/expr.v
```diff
@@ -699,6 +693,34 @@ fn (mut p Parser) expr_with_left(left ast.Expr, precedence int, is_stmt_ident bo
...
+		if p.inside_defer {
+			p.error_with_pos('error propagation not allowed inside `defer` blocks', p.prev_tok.pos())
+		}
+		return ast.OrExpr{
+			kind: if is_not { ast.OrKind.propagate_result } else { ast.OrKind.propagate_option }
+			pos:  or_pos
+		}
+	} else {
+		return ast.OrExpr{
+			kind: ast.OrKind.absent
```
spytheman, line 719: please add the pos too from `p.tok.pos()`

vlib/v/parser/expr.v
```diff
@@ -699,6 +693,34 @@ fn (mut p Parser) expr_with_left(left ast.Expr, precedence int, is_stmt_ident bo
 	return node
 }
 
+fn (mut p Parser) gen_or_block() ast.OrExpr {
```
spytheman, line 696: Why `gen_or_block` instead of `or_expr` ?

vlib/v/gen/c/fn.v
```diff
@@ -864,16 +864,47 @@ fn (mut g Gen) call_expr(node ast.CallExpr) {
...
-		g.expr(node.left)
-		g.is_fn_index_call = false
-	} else if !g.inside_curry_call && node.left is ast.CallExpr && node.name == '' {
+	} else if !g.inside_curry_call && node.left is ast.IndexExpr && node.name == '' {
 		if node.or_block.kind == .absent {
+			old_is_fn_index_call := g.is_fn_index_call
+			g.is_fn_index_call = true
 			g.expr(node.left)
+			g.is_fn_index_call = old_is_fn_index_call
 		} else {
```
spytheman, line 873: Please, add a short comment here, with a small sample snippet, that will trigger this branch.

It is not clear to me, what it will be, by just reading it, even knowing the test, and the new else{} branch is a lot over 10 lines long.

spytheman: For example `// assert map3['f']('test')('test') == 2` (or whatever it is), will make it a lot clearer.

vlib/v/parser/expr.v
felipensp: Because the CallExpr member is named `or_block`.

vlib/v/parser/expr.v
felipensp: Good catch. :-)

spytheman (approved):
Excellent work.

spytheman (approved):
Good work.

PR #22752, opened 2024-11-03
cgen: fix codegen for generated tmp var to fit number of indirections from fn argument

### LINKED ISSUE #22749: cgen doesn't create temporaries for arguments with multiple references
**V doctor:**
```
V full version: V 0.4.8 5c65e58.335bb63

vexe mtime: 2024-11-03 09:59:01

Git vroot status: weekly.2024.43-96-g335bb635
.git/config present: true

```

**What did you do?**
```v
fn takes_ref(x &&&&int) {
	unsafe { ****x = 5 }
}

fn main() {
	y := 4;
	takes_ref(y);
	println(y);
}
```

`5` in stdout

**What did you see instead?**
```
C:/Users/user/AppData/Local/Temp/v_0/test.01JBRWGXT25H4QKMQ8M9RWFQTW.tmp.c:7204: at print_backtrace_skipping_top_frames_tcc: Backtrace
C:/Users/user/AppData/Local/Temp/v_0/test.01JBRWGXT25H4QKMQ8M9RWFQTW.tmp.c:7171: by print_backtrace_skipping_top_frames
C:/Users/user/AppData/Local/Temp/v_0/test.01JBRWGXT25H4QKMQ8M9RWFQTW.tmp.c:8116: by unhandled_exception_handler
7ffd9d7c8b5c : by ???
C:/Users/user/AppData/Local/Temp/v_0/test.01JBRWGXT25H4QKMQ8M9RWFQTW.tmp.c:13738: at main__takes_ref: RUNTIME ERROR: invalid memory access
C:/Users/user/AppData/Local/Temp/v_0/test.01JBRWGXT25H4QKMQ8M9RWFQTW.tmp.c:13745: by main__main
C:/Users/user/AppData/Local/Temp/v_0/test.01JBRWGXT25H4QKMQ8M9RWFQTW.tmp.c:13784: by wmain
00467060 : by ???
004671c3 : by ???
7ffd9d447614 : by ???
```

**Additional information**
This equivalent program in Rust prints `5`:
```rs
fn takes_ref(x: &mut &mut &mut &mut i32) {
    ****x = 5;
}

fn main() {
    let mut y = 4;
    takes_ref(&mut &mut &mut &mut y);
    println!("{y}");
}
```

The generated C code looks like this:
```c
	int y = 4;;
	main__takes_ref((voidptr)&/*qq*/y);;
	println(int_str(y));;
}
```

felipensp:
Fix #22749

```v 
fn takes_ref(x &&&&int) {
	unsafe { ****x = 5 }
}

fn main() {
	y := 4
	takes_ref(y)
	println(y)
	assert y == 5
}
```

```c
        int y = 4;
        int *_t1 = &y;
        int **_t1_1 = &_t1;
        int ***_t1_2 = &_t1_1;
        main__takes_ref((voidptr)&/*qq*/_t1_2);
        println(int_str(y));
}
```

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzI3ZGNhYTI3Y2M3NjgxYzYyNjRlODUiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.XHThDBPIN9QbPunG-drkD59yq-BmIKe74kRNSnidlPU">Huly&reg;: <b>V_0.6-21198</b></a></sub>


medvednikov:
It's a hard philosophical question. Many people in the community think there should be no autoref/deref, I think that a modern language should handle it.

medvednikov:
In 2025+ I don't want to remember, do I have to pass Config or &Config etc.

medvednikov:
But indeed maybe it should only be done for &Foo, but not for &&&&&Foo, which is too low level and rare.

zeozeozeo:
> But indeed maybe it should only be done for &Foo, but not for &&&&&Foo, which is too low level and rare.

what about just `&&Foo`? having references to references is pretty common

spytheman:
Aall the intermediate temporary variables here, are generated on the stack, in cgen, thus the checker knows nothing about them.

If the produced higher level pointer, is then stored in a struct or returned back, or otherwise escape the function, you will get a very hard to debug situation, that can not be checked at compile time, and are hard to catch at runtime too ...

Imho strict counting of pointer levels, without *any* guessing, and auto ~~(de)referencing~~ enshitification, that has errors for mismatches, is superior in most ways (except that it is perhaps a bit more verbose):
 - it is easier to understand for people
 - it is easier to implement in the compiler without special cases
 - it is easier to make meaningful and helpful error checks for it

felipensp:
I'm against such feature as well. I don't like introducing new magic behaviors.

PR #22754, opened 2024-11-04
v: add `in` expr with numbers (`var in 1..4`)

### LINKED ISSUE #20352: Implement `if num in start..end`
### Describe the feature

There was a discussion happening on the discord server to implement.

```v
i := 42
if i in 10..100 {
    println('i is in range between 10 and 100')
} else {
    println('i dont know')
}
```

### Use Case

This feature is handy when checking whether a specific int or character is in that range. We also have this in the match but it uses an inclusion i.e. `[start, end]` rather than `[start, end)`. The same has to be implemented but for `[start, end)`.

### Proposed Solution

This is easy to implement. The generated code will just compare whether the num or character is between that range or not. Just like it is implemented for match statements.

### Other Information

Only constants, untyped numbers and characters are allowed to be here. Also check if the given range is valid or not.

### Acknowledgements

### Version used

V Latest

```
V full version: V 0.4.3 c0d7c58.5d75d06
OS: linux, Pop!_OS 22.04 LTS
Processor: 12 cpus, 64bit, little endian, AMD Ryzen 5 5600H with Radeon Graphics

getwd: /home/delta
vexe: /home/delta/Documents/git/v_dev/v
vexe mtime: 2024-01-02 14:27:22

vroot: OK, value: /home/delta/Documents/git/v_dev
VMODULES: OK, value: /home/delta/.vmodules
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2023.52-54-g6bcb47ef
.git/config present: true

```
> [!NOTE]
>

felipensp:
Fix #20352

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzI4YWU0YTI3Y2M3NjgxYzYyNjk3NDkiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.GnvheXAaZOFLKtEE6lL5VFhJxJAkBPnVzHk17YNk47I">Huly&reg;: <b>V_0.6-21200</b></a></sub>


vlib/v/tests/range_expr_with_int_test.v
```diff
@@ -0,0 +1,29 @@
...
+	return 1.22
+}
+
+fn r() int {
+	return 5
+}
+
+fn test_main() {
+	y := 4
+	assert y in 1..20
```
spytheman, line 17: test the edges too

spytheman: note also that in for loops, the upper limit is not inclusive:
`for i in 0..5 { println(i) }` will produce the numbers 0,1,2,3,4 ... but not 5.

vlib/v/gen/c/infix.v
```diff
@@ -644,12 +644,40 @@ fn (mut g Gen) infix_expr_in_op(node ast.InfixExpr) {
...
+			g.expr(node.right.high)
+			g.write(')')
+		} else {
+			g.write('(')
+			g.expr(node.left)
+			g.write(' >= ')
+			g.expr(node.right.low)
+			g.write(' && ')
+			g.expr(node.left)
+			g.write(' <= ')
```
spytheman, line 677: should this be <= or just < ?

vlib/v/gen/c/infix.v
```diff
@@ -644,12 +644,40 @@ fn (mut g Gen) infix_expr_in_op(node ast.InfixExpr) {
...
+			tmp_var := g.new_tmp_var()
+			g.write('${g.styp(node.left.return_type)} ${tmp_var} = ')
+			g.expr(node.left)
+			g.writeln(';')
+			g.write(line)
+			g.write('(')
+			g.write('${tmp_var} >= ')
+			g.expr(node.right.low)
+			g.write(' && ')
+			g.write('${tmp_var} <= ')
```
spytheman, line 667: <= or < ?

vlib/v/tests/range_expr_with_int_test.v
felipensp: Good point.

vlib/v/tests/range_expr_with_int_test.v
```diff
@@ -0,0 +1,47 @@
...
+fn r() int {
+	return 5
+}
+
+fn test_main() {
+	y := 4
+
+	assert y in 1..20
+	assert y !in 5..20
+	assert y !in 4..20
```
spytheman, line 23: `4` should pass here; `4..20` is `[4,20)`, i.e. the lower bound is inclusive; only the upper bound is exclusive.

vlib/v/tests/range_expr_with_int_test.v
```diff
@@ -0,0 +1,47 @@
...
+	assert r() in 4..6
+	assert 1.22 !in z()?..t()
+	assert 1.23 !in z()?..t()
+	assert 1.221 in z()?..t()
+
+	a := {
+		'f': 2
+	}
+	assert a['f'] in 1..3
+	assert a['f'] !in 2..3
```
spytheman, line 37: this should be: `	assert a['f'] !in 2..3`

vlib/v/tests/range_expr_with_int_test.v
```diff
@@ -0,0 +1,47 @@
...
+	assert a['f'] in min..max
+}
+
+fn test_rune() {
+	assert `a` !in `a`..`h`
```
spytheman, line 42:
```suggestion
	assert `a` in `a`..`h`
```

vlib/v/tests/range_expr_with_int_test.v
```diff
@@ -0,0 +1,47 @@
...
+
+	assert y in 1..20
+	assert y !in 5..20
+	assert y !in 4..20
+	assert y !in min..max
+
+	assert (t() !in 1..3) == false
+	assert r() !in 5..6
+	assert r() in 4..6
+	assert 1.22 !in z()?..t()
```
spytheman, line 29:
```suggestion
	assert 1.22 in z()?..t()
```

vlib/v/tests/range_expr_with_int_test.v
```diff
@@ -0,0 +1,48 @@
...
+
+fn test_main() {
+	y := 4
+
+	assert y in 1..20
+	assert y !in 5..20
+	assert y in 4..20
+	assert y !in min..max
+
+	assert (t() !in 1..3) == false
```
spytheman, line 26:
```suggestion
	assert t() in 1..3
```

spytheman (approved):
Excellent work.

PR #22755, opened 2024-11-04
builtin: fix runes.to_upper() (fix #22742)

### LINKED ISSUE #22742: string to_upper() latin (utf8?) characters error
### Describe the bug

Not all characters are capitalized

### Reproduction Steps

'água'.to_upper()  // <= 'áGUA'

### Expected Behavior

'água'.to_upper()  // <= 'ÁGUA'

### Current Behavior

'água'.to_upper()  // <= 'áGUA'

### Possible Solution

I don't know

_No response_

### V version

V 0.4.8 5d76ee4

Linux linux 6.8.0-48-generic

> [!NOTE]
>

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzI2NGMyMzI3Y2M3NjgxYzYyNWZiN2MiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.OcF9fxJ6GQBqenB5XDL50u79WtqW6d1XuCrFRrPtdlI">Huly&reg;: <b>V_0.6-21189</b></a></sub>

yuyi98:
This PR fix runes.to_upper() (fix #22742).

- Implement rune.to_upper()/to_lower()/to_title().
- Fix runes.to_upper().
- Add tests.

```v
fn main() {
	s1 := 'água'.to_upper()
	println(s1)
	s2 := 'ÁGUA'.to_lower()
	println(s2)
}

PS D:\Test\v\tt1> v run .        
ÁGUA
água
```


felipensp (approved):

vlib/builtin/rune.v
```diff
@@ -81,3 +428,74 @@ pub fn (c rune) length_in_bytes() int {
...
+			}
+			if offset == upper_lower {
+				is_odd := (c - cur_map.start) % 2 == 1
+				if mode == .to_upper && is_odd {
+					return c - 1
+				} else if mode == .to_lower && !is_odd {
+					return c + 1
+				} else {
+					return c
+				}
```
StunxFS, line 489:
```suggestion
				return c
```

`else` branch is unnecessary

vlib/builtin/rune.v
```diff
@@ -81,3 +428,74 @@ pub fn (c rune) length_in_bytes() int {
...
+			} else {
+				return c + offset
+			}
```
StunxFS, line 492:
```suggestion
			return c + offset
```
same here

StunxFS (approved):
Excellent!

vlib/builtin/rune.v
yuyi98: ok, thanks !

spytheman:
The change is nice, but unfortunately it affects `builtin` in a non trivial way, by bloating the generated .c source for it, and its binary size :-|, and `builtin` is used by practically all V programs.

Here are some comparisons between master and it, for `hello_world.v` with and without `-skip-unused`:
```
#0 13:03:13 ^ master /space/v/oo>./v -skip-unused -o hw_master_skip_unused examples/hello_world.v
#0 13:03:27 ^ master /space/v/oo>./v -o hw_master examples/hello_world.v
#0 13:03:32 ^ master /space/v/oo>ls -larS hw_*
-rw-rw-r-- 1 delian delian 221988 Nov  5 12:48 hw_master_skip_unused.c
-rw-rw-r-- 1 delian delian 246839 Nov  5 12:56 hw_pr_shortened_skip_unused.c
-rw-rw-r-- 1 delian delian 246839 Nov  5 13:01 hw_pr_shortened_skip_unused_2.c
-rw-rw-r-- 1 delian delian 260311 Nov  5 12:49 hw_pr_skip_unused.c
-rwxrwxr-x 1 delian delian 445184 Nov  5 13:03 hw_master_skip_unused
-rw-rw-r-- 1 delian delian 450586 Nov  5 12:48 hw_master.c
-rw-rw-r-- 1 delian delian 476070 Nov  5 12:56 hw_pr_shortened.c
-rw-rw-r-- 1 delian delian 476152 Nov  5 13:01 hw_pr_shortened_2.c
-rwxrwxr-x 1 delian delian 482008 Nov  5 13:01 hw_pr_shortened_skip_unused_2
-rwxrwxr-x 1 delian delian 482008 Nov  5 13:02 hw_pr_shortened_skip_unused
-rw-rw-r-- 1 delian delian 489542 Nov  5 12:49 hw_pr.c
-rwxrwxr-x 1 delian delian 719432 Nov  5 13:03 hw_master
-rwxrwxr-x 1 delian delian 757320 Nov  5 13:02 hw_pr
-rwxrwxr-x 1 delian delian 757552 Nov  5 13:01 hw_pr_shortened_2
#0 13:03:34 ^ master /space/v/oo>
```

I've managed to reduce the .C size expansion a bit, by renaming `upper_lower` to `ul` and `start` to `s`, `end` to `e`, `upper_offset` to `uo`, `lower_offset` to `lo`, `title_offset` to `to` (the biggest contributor to the .c bloat is the initialization of the `rune_maps` constant in _vinit, that repeats the field names a lot), but still ... as it is, a simple program will grow with ~37KB, even though it does not use that new functionality at all :-| .

Can we move the new versions, that require that table, outside of `builtin`, to `encoding.utf8` where they will not affect all V programs?

JalonSolov:
Does the size of the generated C file really matter that much?  I don't usually pay any attention at all the the size of the .c files.  I am more interested in the speed of the final code than the size of the intermediate steps.

If it also affects the size of the final executable, that's not too big a deal right now, either.  At some point, we will have better unused code deletion, which will help tremendously.

spytheman:
Yes, it does matter, because it slows down compilation for short programs.
Most tests are very short, and there are thousands of them.
Increasing their size and time requirements with 5-10%, for a feature that will not be used much, does not seem like a good trade off to me.

spytheman:
>  At some point, we will have better unused code deletion, which will help tremendously.

Ideally yes. Pragmatically, it does increase the size (and slows down compilation) right now.

spytheman:
You could say, that features like this, can serve as motivators for improving the code generation, as a second order effect, and you will be right 🤷🏻‍♂️ .

spytheman:
The speed of the final code (at runtime) is not affected much - the size is increased mainly by the new const table for the rune rules.

spytheman:
Here is a small comparison while compiling/running tests, between master and the PR here.
I've chosen `vlib/builtin/`, `vlib/math/`, `vlib/flag/`, `vlib/encoding/`, because those are relatively small tests, that do not do much IO, or network operations (i.e. their runtime is small compared to the compile time).
```
#0 16:46:23 ^ master /space/v/oo>
#0 16:46:23 ^ master /space/v/oo>./v self
V self compiling ...
#0 16:46:28 ^ master /space/v/oo>./v self
V self compiling ...
#0 16:46:31 ^ master /space/v/oo>./v -progress test vlib/builtin/ vlib/math/ vlib/flag/ vlib/encoding/
---- Testing... ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 OK    [117/117] C:   520.1 ms, R:     3.035 ms vlib/math/vec/vec4_test.v
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Summary for all V _test.v files: 117 passed, 117 total. Elapsed time: 55923 ms, on 1 job. Comptime: 49685 ms. Runtime: 6179 ms.
#0 16:47:30 ^ master /space/v/oo>./v -progress test vlib/builtin/ vlib/math/ vlib/flag/ vlib/encoding/
---- Testing... ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 OK    [117/117] C:   516.2 ms, R:     3.041 ms vlib/math/vec/vec4_test.v
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Summary for all V _test.v files: 117 passed, 117 total. Elapsed time: 53822 ms, on 1 job. Comptime: 47581 ms. Runtime: 6183 ms.
#0 16:48:26 ^ master /space/v/oo>./v -progress test vlib/builtin/ vlib/math/ vlib/flag/ vlib/encoding/
---- Testing... ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 OK    [117/117] C:   520.4 ms, R:     2.985 ms vlib/math/vec/vec4_test.v
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Summary for all V _test.v files: 117 passed, 117 total. Elapsed time: 53661 ms, on 1 job. Comptime: 47442 ms. Runtime: 6163 ms.
#0 16:49:26 ^ master /space/v/oo>
#0 16:49:28 ^ master /space/v/oo>git switch -
Switched to branch 'implement_rune_to_upper'
#0 16:49:33 ^ implement_rune_to_upper /space/v/oo>git pull --rebase origin master
From github.com:vlang/v
 * branch                master     -> FETCH_HEAD
Current branch implement_rune_to_upper is up to date.
#0 16:49:38 ^ implement_rune_to_upper /space/v/oo>./v self
V self compiling ...
#0 16:49:42 ^ implement_rune_to_upper /space/v/oo>./v self
V self compiling ...
#0 16:49:46 ^ implement_rune_to_upper /space/v/oo>./v -progress test vlib/builtin/ vlib/math/ vlib/flag/ vlib/encoding/
---- Testing... ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 OK    [117/117] C:   537.2 ms, R:     3.167 ms vlib/math/vec/vec4_test.v
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Summary for all V _test.v files: 117 passed, 117 total. Elapsed time: 57700 ms, on 1 job. Comptime: 51363 ms. Runtime: 6276 ms.
#0 16:50:46 ^ implement_rune_to_upper /space/v/oo>./v -progress test vlib/builtin/ vlib/math/ vlib/flag/ vlib/encoding/
---- Testing... ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 OK    [117/117] C:   532.8 ms, R:     2.996 ms vlib/math/vec/vec4_test.v
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Summary for all V _test.v files: 117 passed, 117 total. Elapsed time: 55495 ms, on 1 job. Comptime: 49182 ms. Runtime: 6255 ms.
#0 16:51:44 ^ implement_rune_to_upper /space/v/oo>./v -progress test vlib/builtin/ vlib/math/ vlib/flag/ vlib/encoding/
---- Testing... ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 OK    [117/117] C:   535.7 ms, R:     3.065 ms vlib/math/vec/vec4_test.v
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Summary for all V _test.v files: 117 passed, 117 total. Elapsed time: 55455 ms, on 1 job. Comptime: 49141 ms. Runtime: 6255 ms.
#0 16:52:41 ^ implement_rune_to_upper /space/v/oo>
```

spytheman:
I also tried testing with `v run .github/workflows/compare_pr_to_master.v` , but that is not very suitable for measuring changes like this, because it runs the old and new versions of the compiler, on the same codebase (the one from the PR), where both will have to deal with the new table etc.

(for what is worth, the new compiler version is slightly faster in that particular situation, compared to the version on master, even though both produce bigger binaries)

spytheman:
On the other hand, this PR does fix the correctness of the methods, and one can argue, that is more important, compared to increased resources needed for compilation and storage.

spytheman:
I wonder if we can compress the rule table somehow 🤔 .

vlib/builtin/rune.v
```diff
@@ -4,6 +4,353 @@ module builtin
 
 import strings
 
+const upper_lower = 5000000
```
spytheman, line 7: Why is `upper_lower` so big, compared to all the other values?
Can it be some other unused value, like `-3` for example?

JalonSolov:
> I wonder if we can compress the rule table somehow 🤔 .

Don't see why not.  Make it a fixed array of fixed arrays of 5 ints.

Slightly more complex code to deal with it, but it removes the overhead of all the `structs`, instead making it one large block of memory.

It'll load faster, as well.

medvednikov:
I'm fine with making `str.to_upper()` work only with ASCII and print a warning for non-ASCII: `use unicode.to_upper()`

vlib/builtin/rune.v
spytheman: There are offsets bigger than 2^16, so `i16` will not be enough to represent them.

spytheman:
The table is ~2.3KB when compressed with xz, so there are is a lot of redundancy in it.

spytheman:
Turning it into a fixed array did not work (I hit a cgen error edge case, which perhaps needs a separate issue).

spytheman:
A flat array of just integers works fine, and reduces the overhead (compared to master, for hello_world) a lot:
* the overhead in generated .c is now down to 12KB  (from 38KB) with a flat array of plain integers
* the overhead in the binary size is down to 9.25KB (from ~36KB)

spytheman:
```
#0 18:12:49 ^ implement_rune_to_upper /space/v/oo>ls -larS mhw* ohw* nhw* fhw*
-rw-rw-r-- 1 delian delian 221620 Nov  5 17:41 mhw_su.c
-rw-rw-r-- 1 delian delian 233769 Nov  5 18:08 fhw_su.c
-rw-rw-r-- 1 delian delian 243791 Nov  5 17:52 nhw_su.c
-rw-rw-r-- 1 delian delian 259943 Nov  5 17:31 ohw_su.c
-rwxrwxr-x 1 delian delian 444384 Nov  5 17:41 mhw_su
-rw-rw-r-- 1 delian delian 449928 Nov  5 17:41 mhw.c
-rwxrwxr-x 1 delian delian 452760 Nov  5 18:08 fhw_su
-rw-rw-r-- 1 delian delian 462710 Nov  5 18:08 fhw.c
-rw-rw-r-- 1 delian delian 472732 Nov  5 17:52 nhw.c
-rwxrwxr-x 1 delian delian 481208 Nov  5 17:31 ohw_su
-rwxrwxr-x 1 delian delian 481208 Nov  5 17:52 nhw_su
-rw-rw-r-- 1 delian delian 488884 Nov  5 17:31 ohw.c
-rwxrwxr-x 1 delian delian 718192 Nov  5 17:41 mhw
-rwxrwxr-x 1 delian delian 727664 Nov  5 18:08 fhw
-rwxrwxr-x 1 delian delian 756080 Nov  5 17:31 ohw
-rwxrwxr-x 1 delian delian 756080 Nov  5 17:52 nhw
#0 18:12:50 ^ implement_rune_to_upper /space/v/oo>
```

* the `mhw` variants are from master
* the `fhw` variants are from my latest changes (using a flat fixed array of ints), on the PR
* the `ohw` variants are from the original code in the PR
* the `nhw` variants are from an intermediate version, where only the fields and const name were shortened

spytheman:
As a bonus, the flat fixed array of ints, is initialized statically, not in `_vinit()`, and it does not need to allocate at runtime.

StunxFS:
Hm, if the problem is the size of the .c file, then why not have the compiler generate a .c file for each module?

StunxFS:
Hm, if the problem is the size of the .c file, then why not have the compiler generate a .c file for each module?

JalonSolov:
Flat fixed array is even better, though it makes the code even more interesting, but pretty standard for C-style programming.

spytheman:
> Hm, if the problem is the size of the .c file, then why not have the compiler generate a .c file for each module?

That is planned too (as an extension of `-usecache`), but does not work as well as it could right now, since the parser and the checker still have to process all .v files to get the type information, even when ultimately the produced .c that will get compiled and linked is smaller.

I.e. you will get a smaller cost for the C compilation with `-usecache`, but nearly the full cost for parsing/checking, even though none of the .v files for the module dependencies have changed.

spytheman:
Just tested `v test-all` on the same machine, on master:
```
---- Summary of `v test-all`: -----------------------------------------------------------------
Total runtime: 1010710 ms
```

on the PR:
```
---- Summary of `v test-all`: -----------------------------------------------------------------
Total runtime: 1114622 ms
```

i.e. for me, the full test suite is still 10% slower now :-| .

JalonSolov:
Still not that bad, considering the CI is what mainly runs `test-all`.  Even making a compiler change locally, you don't run `test-all` after every minor change, only when you're ready to push to a PR.

spytheman:
> Still not that bad, considering the CI is what mainly runs `test-all`. Even making a compiler change locally, you don't run `test-all` after every minor change, only when you're ready to push to a PR.

Yes, and after re-running it again, I got a much closer result (+ ~2% overhead) ... Also some of the tests do network ops, so the diff is not so bad after all, and it can be optimized later, after merging.

spytheman:
I'll wait till the full CI passes here (because of the unsafe{} pointer arithmetic involved in the latest commit), then merge it.

spytheman:
I think also that some of the overhead, can be explained not just by the increased .c file size by the new table, but also by the change to string.to_upper and string.to_lower to use a linear scan for non ASCII characters, and then the new rune based implementation to handle the non utf8 strings. 

The compiler may have to be changed, to selectively use a .to_upper_ascii and .to_lower_ascii methods, that will work as before, to avoid the non ASCII check completely (since we do not allow non ASCII characters for the identifiers/keywords etc)

vlib/builtin/rune_map.v
```diff
@@ -0,0 +1,356 @@
...
+	l int // lower offset
+	t int // title offset
+}
+*/
+// The rune_maps table below, has rows, each containing 5 integers, equivalent to the RuneMap struct from above.
+// It is represented that way, instead of the more natural array of structs, to save on the .c encoding used for the initialisation.
+// The overhead for representing it as an array of structs was ~36KB in .c, while with the flat array of ints, it is ~12KB.
+// Given that xz can compress it to ~2.2KB, it could be probably represented in an even more compact way...
+const rune_maps = [
+	0x41, 0x5A, 0, 32, 0,
```
JalonSolov, line 27: Perhaps make the first entry an i32, to fix the size of all the entries?  To guard against the `int` changing size when that's implemented.

spytheman: Done in 5c8084d .

yuyi98:
@spytheman  I think we can get rid of the to_title item and shrink it.

spytheman:
There were some entries, for which the upper offset was != than the title one:
```
       0x1C4, 0x1C4, 0, 2, 1,
       0x1C5, 0x1C5, -1, 1, 0,
       0x1C6, 0x1C6, -2, 0, -1,
       0x1C7, 0x1C7, 0, 2, 1,
       0x1C8, 0x1C8, -1, 1, 0,
       0x1C9, 0x1C9, -2, 0, -1,
       0x1CA, 0x1CA, 0, 2, 1,
       0x1CB, 0x1CB, -1, 1, 0,
       0x1CC, 0x1CC, -2, 0, -1,
       0x1F1, 0x1F1, 0, 2, 1,
       0x1F2, 0x1F2, -1, 1, 0,
       0x1F3, 0x1F3, -2, 0, -1,
       0x10D0, 0x10FA, 3008, 0, 0,
       0x10FD, 0x10FF, 3008, 0, 0,
```

spytheman:
hm, interesting ... tcc can not handle const integer values, used in const fixed array inits, failing with: `error: initializer element is not constant`, for code looking like this:
```c
typedef int Array_fixed_int_3 [3];
const int _const_ul = -3;
Array_fixed_int_3 _const_abc = {0, 1, _const_ul};
```

`gcc` and `clang` have no problems with the above.

yuyi98:
@spytheman  Thank you very much for doing a lot of optimization!

Issue #22758, opened 2024-11-04
cgen: `x() in [...]` calls `x()` for every element of array


zeozeozeo:
**V doctor:**

```
V full version: V 0.4.8 5c65e58.3108820
OS: windows, Microsoft Windows 10 Pro v19045 64-bit
Processor: 10 cpus, 64bit, little endian, 

getwd: C:\Users\user\Downloads
vexe: C:\Users\user\Desktop\Code\v\v.exe
vexe mtime: 2024-11-04 16:24:59

vroot: OK, value: C:\Users\user\Desktop\Code\v
VMODULES: OK, value: C:\Users\user\.vmodules
VTMP: OK, value: C:\Users\user\AppData\Local\Temp\v_0

Git version: git version 2.42.0.windows.1
Git vroot status: weekly.2024.44-37-g3108820d
.git/config present: true

CC version: Error: 'cc' is not recognized as an internal or external command,

operable program or batch file.


thirdparty/tcc status: thirdparty-windows-amd64 b425ac82
```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg test.v`

```v
fn test() bool {
	println("i should only be called once")
	return true
}

fn main() {
	println(test() in [false, true])
}
```

**What did you expect to see?**

```
i should only be called once
true
```

**What did you see instead?**

```
i should only be called once
i should only be called once
true
```

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzI4ZjZhYWU3Y2M1YTc4NTQyNzc1YWQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.9spoKbUCnRUiwITWOTDhTK1z5RNgUzpx7Au_Zssip9c">Huly&reg;: <b>V_0.6-21205</b></a></sub>

HectorPeeters:
[...]

PR #22763, opened 2024-11-04
v: fix redundant emission of left expression in in infix expressions

HectorPeeters:
This PR fixes the issue addressed in #22758. It prevents the left expression of an `in` infix expression from being emitted multiple times. This occurs in the following code example:

```v
x() in [12, 13]
```

This gets optimized into:

```c
x() == 12 || x() == 13
```

In this case this optimization is incorrect.

The fix I implemented checks if the left expression is an `ast.Ident`. However, I'm unsure if there is side effect information present in the compiler. Having this information available would allow this optimization to be applied in more situations.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzI5NDljNWU3Y2M1YTc4NTQyNzg2ZDQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.0fmf9tYw_X9GuTIjy0IcYVVI1m6K669tgfWJ5HZYqkg">Huly&reg;: <b>V_0.6-21210</b></a></sub>

### DIFF
--- a/vlib/v/gen/c/infix.v
+++ b/vlib/v/gen/c/infix.v
@@ -517,7 +517,8 @@ fn (mut g Gen) infix_expr_in_op(node ast.InfixExpr) {
 		if node.right is ast.ArrayInit {
 			elem_type := node.right.elem_type
 			elem_sym := g.table.sym(elem_type)
-			if node.right.exprs.len > 0 {
+			// TODO: replace ast.Ident check with proper side effect analysis
+			if node.right.exprs.len > 0 && node.left is ast.Ident {
 				// `a in [1,2,3]` optimization => `a == 1 || a == 2 || a == 3`
 				// avoids an allocation
 				g.write('(')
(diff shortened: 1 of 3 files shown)


spytheman:
> However, I'm unsure if there is side effect information present in the compiler.

There is not yet afaik.

spytheman (approved):
Excellent work.

medvednikov:
This kills the allocation optimization though and now `x() in [12, 13]` will result in an allocation for [12, 13]?

PR #22764, opened 2024-11-04
cgen: fix codegen to emit callexpr one time for `in` expr optimization

### LINKED ISSUE #22758: cgen: `x() in [...]` calls `x()` for every element of array

felipensp:
Fix #22758

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzI5NTJlNWU3Y2M1YTc4NTQyNzg4YTIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.kpe1y_WnxwqN70OaP7sMk82Mn6CM9WcH08_KnxSkGoU">Huly&reg;: <b>V_0.6-21211</b></a></sub>

### DIFF
--- a/vlib/v/slow_tests/inout/in_expr_callexpr.vv
+++ b/vlib/v/slow_tests/inout/in_expr_callexpr.vv
@@ -1,10 +1,10 @@
 fn t() bool {
-	println('${@FN}')
+	println(@FN)
 	return true
 }
 
 fn t2() int {
-	println('${@FN}')
+	println(@FN)
 	return 2
 }
 
--- a/vlib/v/slow_tests/inout/in_expr_callexpr.vv
+++ b/vlib/v/slow_tests/inout/in_expr_callexpr.vv
@@ -1,0 +1,14 @@
+fn t() bool {
+	println('${@FN}')
+	return true
+}
+
+fn t2() int {
+	println('${@FN}')
+	return 2
+}
+
+fn main() {
+	println(t() in [false, true])
+	println(t2() in [0, 1, 2, 3, 4, 5])
+}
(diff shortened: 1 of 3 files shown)


spytheman:
The issue was also fixed in https://github.com/vlang/v/pull/22763 . The fix there disabled the optimization for anything but `ident in [x,y,z]`, which works, but is slower 🤔 by ~10% for `v -o vnew cmd/v` .

I'll have to measure the performance impact of this PR before merging.

vlib/v/slow_tests/inout/in_expr_callexpr.vv
```diff
@@ -0,0 +1,14 @@
+fn t() bool {
+	println('${@FN}')
```
spytheman, line 2: @FN is already a string, there is no need to interpolate it

spytheman (approved):
Good work.

Issue #22792, opened 2024-11-08
`$if T is $address`


enghitalo:
### Describe the feature

Verify if generic is pointer

### Use Case

insert or write value from pointer

```v
a := 5
b := &a

mut expected := false

$if b is $pointer {
    expected = true
} $else {
    expected = false
}

assert expected
```

```v
mut a := 5
mut b := &a
mut c := &b

fn insert_value_in_pointer[T](mut pointer T) {
	$if b is $pointer {
		insert_value_in_pointer(*pointer)
	} $else {
		pointer = 10
	}
}

insert_value_in_pointer(c)

assert a == 10
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

latest

### Environment details (OS name and version, etc.)

all

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzJkNjY1MGU3Y2M1YTc4NTQyOTY0NzciLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.RmPNkoQMU-fhQI17SsFAQpps1qAP1ornZrUI6JcKTtE">Huly&reg;: <b>V_0.6-21238</b></a></sub>

felipensp:
Already implemented.

PR #22805, opened 2024-11-08
v: add `typeof(var).indirections` and `T.indirections`

felipensp:
This PR adds a way to retrieve the number of indirections related to `typeof(expr)` and `T` generic type.

```v
fn indirections[T](val T) int {
	return T.indirections
}

fn test_main() {
	a := 0
	assert typeof(a).indirections == 0
	assert indirections(a) == 0
	b := &a
	assert typeof(b).indirections == 1
	assert indirections(b) == 1
	c := [1]
	assert typeof(c).indirections == 0
	assert indirections(c) == 0
	d := &c
	assert typeof(d).indirections == 1
	assert indirections(d) == 1
}
```

Related to #22792

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzJlMDRhOGU3Y2M1YTc4NTQyOTdmOWEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.JzKDL2jfLYNYbQLoA0GpdCh_P_DTaOCaEP299uY_IIA">Huly&reg;: <b>V_0.6-21251</b></a></sub>

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -287,6 +287,7 @@ pub enum GenericKindField {
 	name
 	typ
 	unaliased_typ
+	indirections
 }
 
 // `foo.bar`
--- a/vlib/v/gen/js/js.v
+++ b/vlib/v/gen/js/js.v
@@ -3262,6 +3262,13 @@ fn (mut g JsGen) gen_selector_expr(it ast.SelectorExpr) {
 				g.write(')')
 				return
 			}
+			.indirections {
+				g.write('new int(')
+				g.write('${int(g.unwrap_generic(it.name_type).nr_muls())}')
+				g.write(')')
+				g.write(')')
+				return
+			}
 			.unknown {
 				if node.field_name == 'name' {
 					g.type_name(it.name_type)
--- a/vlib/v/gen/js/js.v
+++ b/vlib/v/gen/js/js.v
@@ -3252,20 +3252,17 @@ fn (mut g JsGen) gen_selector_expr(it ast.SelectorExpr) {
 				g.write('new int(')
 				g.write('${int(g.unwrap_generic(it.name_type))}')
 				g.write(')')
-				g.write(')')
 				return
 			}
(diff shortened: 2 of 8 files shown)


vlib/v/ast/ast.v
```diff
@@ -287,6 +287,7 @@ pub enum GenericKindField {
 	name
 	typ
 	unaliased_typ
```
spytheman, line 289: I think we should use `type` and `unaliased_type` now that enums allow it.

vlib/v/gen/js/js.v
```diff
@@ -3252,13 +3252,17 @@ fn (mut g JsGen) gen_selector_expr(it ast.SelectorExpr) {
 				g.write('new int(')
 				g.write('${int(g.unwrap_generic(it.name_type))}')
 				g.write(')')
-				g.write(')')
```
spytheman, line 3255: good find

vlib/v/tests/typeof_indirections_test.v
```diff
@@ -0,0 +1,18 @@
...
+	assert indirections(a) == 0
+	b := &a
+	assert typeof(b).indirections == 1
+	assert indirections(b) == 1
+	c := [1]
+	assert typeof(c).indirections == 0
+	assert indirections(c) == 0
+	d := &c
+	assert typeof(d).indirections == 1
+	assert indirections(d) == 1
```
spytheman, line 17: Please also add an assert for `dd := &d` and `bb := &b` .

spytheman (approved):
Excellent work.

PR #22814, opened 2024-11-09
orm: add plain serial attribute support

nunsez:
This PR allows to use `serial` attribute separately from `sql` attribute.
What does it give?

Now we can't change the column name and mark it as a serial, since `sql` attribute is used for both tasks

```v
struct Record {
}
// or
struct Record {
  id int @[primary; sql: 'custom_id']
}
```

The PR makes this possible while maintaining backwards compatibility with `@[sql: serial]`

```v
struct Record {
  id int @[primary; serial; sql: 'custom_id']
}
// still valid
struct Record {
}
```

What do you think about this?

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzJmM2U0YWU3Y2M1YTc4NTQyOWRlOWYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.4gTh76WB2E6F0qAnGTvzXFJ8n6pXy1Var7uf_Jh5Is4">Huly&reg;: <b>V_0.6-21260</b></a></sub>


spytheman (approved):
Excellent work.

PR #22840, opened 2024-11-12
checker: fix missing or-block check for callexpr

### LINKED ISSUE #22835: cgen error for `Sumtype( f() or { variant_value } )`
**V doctor:**
```
V full version: V 0.4.8 a132d97.ec1f95a
OS: linux, Ubuntu 20.04.6 LTS

getwd: /home/delian/code/v
vexe: /home/delian/code/v/v
vexe mtime: 2024-11-11 17:27:28

vroot: OK, value: /home/delian/code/v
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.46-1-gec1f95ac-dirty
.git/config present: true

emcc version: N/A

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg oo.v`
```v
type Aa = string | int

fn f() !string {
	return ''
}

y := Aa(f() or { 2 })
println(y)

```

a compiled program, or a checker error

**What did you see instead?**
```
cc: /tmp/v_1000/oo.01JCE5SGAZ2BEB2ATKXPA41MXY.tmp.c:13771: error: cannot convert 'int' to 'struct string'
```

felipensp:
Fix #22835

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzMyYzdjZmU3Y2M1YTc4NTQyYjQxMjciLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.T7WUCwRNLpLrTpGaT8uPtX3loOmEk-mkbhuK9zwsETQ">Huly&reg;: <b>V_0.6-21285</b></a></sub>


vlib/v/checker/checker.v
```diff
@@ -1436,6 +1450,7 @@ fn (mut c Checker) check_or_last_stmt(mut stmt ast.Stmt, ret_type ast.Type, expr
 				}
 			}
 			ast.Return {}
+			ast.AssertStmt {}
```
spytheman, line 1453: hm, what happens for `assert true` ?

spytheman: also keep in mind that asserts will be removed completely with `-prod`, so they can not serve as reliable terminating statements - the control flow can go past them.

vlib/v/parser/expr.v
```diff
@@ -783,9 +783,13 @@ fn (mut p Parser) infix_expr(left ast.Expr) ast.Expr {
...
 	old_assign_rhs := p.inside_assign_rhs
-	p.inside_assign_rhs = true
+	if op in [.decl_assign, .assign] {
+		p.inside_assign_rhs = true
+	}
 	right = p.expr(precedence)
-	p.inside_assign_rhs = old_assign_rhs
+	if op in [.decl_assign, .assign] {
+		p.inside_assign_rhs = old_assign_rhs
+	}
```
spytheman, line 792: that is brittle - imho remove the condition here

spytheman: (in the future if/when those places get more distant, the risk of de-synchronizing the 2 conditions can grow)

spytheman (approved):
Good work.

vlib/v/checker/checker.v
felipensp: Without this line it will gives `error: last statement in the `or {}` block should be an expression of type `int` or exit parent scope`

vlib/v/parser/expr.v
felipensp: Yeah. I do agree.

vlib/v/checker/checker.v
spytheman: My point is that it *should* get to that error for a reason - if it does not, the problem will happen at runtime, when the assert condition is true, or when the assert statement is optimized out completely.

spytheman: For example:
```v
fn f() ?int { return none }
a := f() or { assert true }
dump(a)
```
what will be shown?

spytheman: Currently it prints:
image

and that is correct.

felipensp: > My point is that it _should_ get to that error for a reason - if it does not, the problem will happen at runtime, when the assert condition is true, or when the assert statement is optimized out completely.

Currently we have code using such "feature".
image
I just keep the code working. What do you think?

spytheman: With this PR:
image

spytheman: > I just keep the code working. What do you think?

`cm.exists('.o', 'abc') or { assert true }`
That does not use the last value - it should continue to work.

`a := f() or { assert true }` does use the value, and when it is not defined, it will lead to bugs.

spytheman: it seems that it is the same situation with the rest - they do not use the value at all, so there is no possibility of undefined values after the execution

felipensp: Should we keep it working for result?

felipensp: > it seems that it is the same situation with the rest - they do not use the value at all, so there is no possibility of undefined values after the execution

Ok, got it. Let me fix it.

spytheman: Yes, it should work for both options and results, as long as the value of the whole `f() or { ... }` is not used.

felipensp: Sure. I added a checker test to state the current behavior. 👍🏻 Thanks

spytheman: Excellent. Thank you.

vlib/v/checker/checker.v
```diff
@@ -1450,11 +1450,12 @@ fn (mut c Checker) check_or_last_stmt(mut stmt ast.Stmt, ret_type ast.Type, expr
...
-			ast.AssertStmt {}
 			else {
-				expected_type_name := c.table.type_to_str(ret_type.clear_option_and_result())
-				c.error('last statement in the `or {}` block should be an expression of type `${expected_type_name}` or exit parent scope',
-					stmt.pos)
+				if stmt !is ast.AssertStmt || c.inside_assign {
+					expected_type_name := c.table.type_to_str(ret_type.clear_option_and_result())
+					c.error('last statement in the `or {}` block should be an expression of type `${expected_type_name}` or exit parent scope',
+						stmt.pos)
+				}
```
spytheman, line 1458: That does not seem right - it has to check whether the value is used or not. Getting the value assigned to a variable is only one kind of using it.

`g(  f() or { assert true }  )`

This for example should trigger the same error, since the value of `f()or{}` is used as a parameter to g().

spytheman: Try this on master, and here:
```v
fn g(x int) { dump(x) }
fn f() ?int { return none }
f() or { assert true } // OK, should continue to work, the result is not used.
// a := f() or { assert true } dump(a) // on master triggers checker error -> ok
// g( f() or { assert true } ) // on master triggers checker error -> ok
```

felipensp: Fixed. Thanks.

vlib/picohttpparser/misc_test.v
```diff
@@ -38,7 +41,10 @@ pub fn test_u64toa_edge_cases() {
 
 	// Test zero value
 	len := unsafe {
-		u64toa(&buf[0], 0) or { assert false }
+		u64toa(&buf[0], 0) or {
+			assert false
+			0
+		}
 	}
```
spytheman, line 48: those should not need fixing - they should work as they are

felipensp: Why? It's in an assignment, its value is expected to be an unwrapped value.

spytheman: oh ... I am wrong, sorry. I should have read it more carefully.
it does use the value in the `len :=` assignment.

vlib/v/checker/or_block_assert_err.out
```diff
@@ -0,0 +1,12 @@
...
+    8 | 
+    9 | f() or { assert true }
+   10 | a := f() or { assert true }
+      |                      ~~~~~~
+   11 | dump(a)
+   12 | g(f() or { assert true })
+vlib/v/checker/or_block_assert_err.vv:12:19: error: last statement in the `or {}` block should be an expression of type `int` or exit parent scope
+   10 | a := f() or { assert true }
+   11 | dump(a)
+   12 | g(f() or { assert true })
```
spytheman, line 11: the .out files have to be updated to match the .vv ones

spytheman: imho just run `VAUTOFIX=1 ./v vlib/v/compiler_errors_test.v` 2 times

vlib/v/checker/tests/call_empty_or_block_err.out
```diff
@@ -0,0 +1,56 @@
...
+      |     ^
+   30 |         bar() or {}
+   31 |         false
+vlib/v/checker/tests/call_empty_or_block_err.vv:35:2: warning: unused variable: `b`
+   33 | 
+   34 |     // ok
+   35 |     b := foo() or {
+      |     ^
+   36 |         foo() or {}
+   37 |         false
```
spytheman, line 35: imho just add dump(a), dump(b) etc at the end to silence the warnings - they are not essential to the fix

PR #22845, opened 2024-11-12
checker: improve pos for if expr 

Delta456:
Improves position for If-Expr conditions

Before

```
code.v:5:14: error: `if` expression requires an expression as the last statement of every branch
    3 |     mut start := 0
    4 | 
    5 |     start += if val < 20 { start += 1  } else { start }
      |              ~~~~~~~~~~~
    6 |     dump(start)
    7 | }
 ```
 
 
 After
 
 ```
vlib/v/checker/tests/if_expr_last_branch_stmt_err.vv:5:30: error: `if` expression requires an expression as the last statement of every branch
    3 |     mut start := 0
    4 | 
    5 |     start = if val < 20 { start += 1 } else { start }
      |                                 ~~
    6 |     dump(start)
    7 | }
```


spytheman (approved):
Excellent work.

PR #22847, opened 2024-11-13
Small cleans up of `x.encoding.asn1` module

blackshirt:
This PR contains some smalls cleans up to the `x.encoding.asn1`, ie,

- Removes `@[noinit]` tag from public struct, Its can makes an issue when custom structure contains the standard universal element tagged with `@[noinit]` tag. For examples, when developing certificates parser, its requires to just instantiate some element with `Certificate{}` call directly and validates it later, but currently this would gives us an error similar with ```struct x.encoding.asn1.Integer` is declared with a `@[noinit]` attribute, so it cannot be initialized with `x.encoding.asn1.Integer{}` ``` and other error with other standard element tagged with `noinit`.  Removes this tag makes it more flexible. Standard universal class has builtin check when this element was serialized, so, i think its safe to remove this tag.
- Adds method to allow creating universal `UtcTime` and `GeneralizedTime` from standard `time.Time` module. Its still rather a hack thing, but can be improved later. 
- Changes definition of `Any` element into `AnyDefinedBy` element, to reflect underlying ASN.1 syntax for this, ie `ANY DEFINED BY` schema. `Any` maybe good for arbitrary other name, also removes internal marker string, its not relevant now. 
- Some small improvement to the readme files (wrong link for examples and the docs when accessed from `modules.vlang.io`)

Best regards and thanks.


vlib/x/encoding/asn1/time.v
```diff
@@ -46,6 +50,19 @@ pub fn UtcTime.new(s string) !UtcTime {
 	}
 }
 
+fn UtcTime.from_time(t time.Time) !UtcTime {
+	// changes into utc time
+	utime := t.local_to_utc()
+	s := utime.custom_format(default_utctime_format) //  20241113060446+0
+	// TODO: add supports for this
+	// Its rather a hack, not efieient as should be
+	// todo: makes its better
```
spytheman, line 59:
```suggestion
	// todo: make it better
```

vlib/x/encoding/asn1/time_test.v
```diff
@@ -73,3 +87,15 @@ fn test_serialize_decode_generalizedtime() ! {
 	assert back.value == s
 	assert back.tag().tag_number() == int(TagType.generalizedtime)
 }
+
+fn test_create_generalizedtime_from_std_time() ! {
+	now := time.utc()
```
spytheman, line 92: Same - use `time.new(year: 2020, month: 12, day: 22, hour: 7, minute: 23)` for example, instead of `time.utc()`, because the later will change each time the test is run.

vlib/x/encoding/asn1/time_test.v
```diff
@@ -20,6 +22,18 @@ fn test_serialize_utctime_basic() ! {
 	assert back.value == inp
 }
 
+fn test_create_utctime_from_std_time() ! {
+	now := time.utc()
+	utb := UtcTime.from_time(now)!
+	utc := UtcTime.from_time(now)!
```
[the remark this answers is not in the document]
blackshirt: Okey..i would change it

vlib/x/encoding/asn1/time_test.v
blackshirt: Would be addressed.. thank for the review

spytheman: Happy to help.

spytheman (approved):
Excellent work.

PR #22848, opened 2024-11-13
v: some branch prediction adjusts

felipensp:
This PR aims improve performance by doing branch prediction adjusts.

Reduced 26.004.738 (Bcm) to 25.334.846 (Bcm) on `v -o cmd.c cmd/v`

Bcm = branch predictor misses

```
$ v -cflags -O0 -cc gcc -cg -cflags -g self
$ valgrind --tool=callgrind --branch-sim=yes --cache-sim=yes  --dump-instr=yes ./v -no-parallel -o cmd.c cmd/v
$ callgrind_annotate --show=Bc,Bcm --sort=Bcm --tree=both --inclusive=no --auto=yes --context=1 callgrind.out.17213 > tree.17213
```

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzM0YThhY2U3Y2M1YTc4NTQyYzIyNzYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.2yRqDiCxv7LkYk2JdJxS9LOloVRdUz5SUejk2DJDkSA">Huly&reg;: <b>V_0.6-21293</b></a></sub>

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -2182,10 +2182,6 @@ fn (mut g Gen) stmt(node ast.Stmt) {
 			g.stmts(node.stmts)
 			g.writeln('}')
 		}
-		ast.AsmStmt {
-			g.write_v_source_line_info_stmt(node)
-			g.asm_stmt(node)
-		}
 		ast.AssignStmt {
 			g.write_v_source_line_info_stmt(node)
 			g.assign_stmt(node)
@@ -2350,6 +2346,10 @@ fn (mut g Gen) stmt(node ast.Stmt) {
 		ast.GotoStmt {
 			g.write_v_source_line_info_stmt(node)
 			g.writeln('goto ${c_name(node.name)};')
+		}
+		ast.AsmStmt {
+			g.write_v_source_line_info_stmt(node)
+			g.asm_stmt(node)
 		}
 		ast.HashStmt {
 			g.hash_stmt(node)
@@ -3555,7 +3555,11 @@ fn (mut g Gen) expr(node_ ast.Expr) {
 			g.ident(node)
 		}
 		ast.IfExpr {
-			g.if_expr(node)
+			if !node.is_comptime {
+				g.if_expr(node)
+			} else {
+				g.comptime_if(node)
+			}
 		}
 		ast.IfGuardExpr {
 			g.write('/* guard */')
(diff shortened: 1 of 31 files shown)


vlib/v/parser/parser.v
```diff
@@ -809,6 +806,9 @@ fn (mut p Parser) top_stmt() ast.Stmt {
 					return p.unexpected()
 				}
 			}
+			.key_asm {
+				return p.asm_stmt(true)
+			}
```
spytheman, line 811: it can be last even - the vast majority of V programs do not use asm statements at all

vlib/v/parser/parser.v
```diff
@@ -1014,9 +1014,6 @@ fn (mut p Parser) stmt(is_top_level bool) ast.Stmt {
 				is_used:   p.inside_test_file || !p.pref.is_prod
 			}
 		}
-		.key_for {
```
spytheman, line 1017: perhaps move the assert branch too - it is not common at all outside of `_test.v` programs

spytheman: Although I wonder if the `switch() { case ... }` construct that will be generated, will not use a short jump table anyway, so moving branches here may be a bit meaningless.

spytheman (approved):
Excellent work.

vlib/v/gen/c/cgen.v
```diff
@@ -2169,31 +2169,26 @@ fn (mut g Gen) stmt(node ast.Stmt) {
...
-			g.write_v_source_line_info_stmt(node)
-			g.asm_stmt(node)
-		}
-		ast.AssertStmt {
-			g.write_v_source_line_info_stmt(node)
-			g.assert_stmt(node)
-		}
-		ast.AssignStmt {
-			g.write_v_source_line_info_stmt(node)
-			g.assign_stmt(node)
```
spytheman, line 2182: Why move that?
It seems more common than function declarations for most programs to me?

spytheman:
Note that compiling with `-O0 -cc gcc` will turn off *all optimizations*, including the ones that turn switches into jump tables.

vlib/v/gen/c/cgen.v
felipensp: I used to do measure on cmd/v

PR #22856, opened 2024-11-14
v: improved `-skip-unused`

felipensp:
This PR remove some unused code that current `-skip-unused` does not currently.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzM1ZDE5ZWU3Y2M1YTc4NTQyY2NlNDEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.NtlXCzOdfvRaHScE7AhrDs_QAHX4q34J9zxf-vJDCyU">Huly&reg;: <b>V_0.6-21301</b></a></sub>

### DIFF
--- a/vlib/v/markused/markused.v
+++ b/vlib/v/markused/markused.v
@@ -71,7 +71,6 @@ pub fn mark_used(mut table ast.Table, mut pref_ pref.Preferences, ast_files []&a
 			// ref_array_idx_str + '.last',
 			// ref_array_idx_str + '.pop',
 			// ref_array_idx_str + '.insert_many',
-			// ref_array_idx_str + '.prepend_many',
 			// ref_array_idx_str + '.reverse',
 			// ref_array_idx_str + '.set_unsafe',
 			// TODO: process the _vinit const initializations automatically too
@@ -82,7 +81,7 @@ pub fn mark_used(mut table ast.Table, mut pref_ pref.Preferences, ast_files []&a
 			// 'os.getwd',
 			// 'v.embed_file.find_index_entry_by_path',
 		]
-		if table.used_features.builtin_types {
+		if table.used_features.builtin_types || table.used_features.as_cast {
 			// println('used builtin')
 			core_fns << '__new_array'
 			core_fns << '__new_array_with_default'
@@ -131,6 +130,9 @@ pub fn mark_used(mut table ast.Table, mut pref_ pref.Preferences, ast_files []&a
 			}
 			if table.used_features.auto_str_ptr {
 				core_fns << 'isnil'
+			}
+			if table.used_features.arr_prepend {
+				core_fns << ref_array_idx_str + '.prepend_many'
 			}
 		} else {
 			// hello world apps
--- a/vlib/v/markused/markused.v
+++ b/vlib/v/markused/markused.v
@@ -13,31 +13,10 @@ pub fn mark_used(mut table ast.Table, mut pref_ pref.Preferences, ast_files []&a
 	defer {
 		util.timing_measure(@METHOD)
 	}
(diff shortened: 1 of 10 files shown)


medvednikov:
From 5903 to 1316, insane! Great work!

JalonSolov:
Is there a good reason to introduce a completely new flag, instead of simply improving on what the current one is supposed to do?

`-skip-unused` should always remove code that isn't used, whether it is V "core functionality", "third-party modules", the user's own code, or anything else.

Eventually, we would even want to have `-skip-unused` be the default, perhaps replacing it with `-no-skip-unused` in case anyone is crazy enough to want all the dross in the output.

felipensp:
> Is there a good reason to introduce a completely new flag, instead of simply improving on what the current one is supposed to do?
> 
> `-skip-unused` should always remove code that isn't used, whether it is V "core functionality", "third-party modules", the user's own code, or anything else.
> 
> Eventually, we would even want to have `-skip-unused` be the default, perhaps replacing it with `-no-skip-unused` in case anyone is crazy enough to want all the dross in the output.

It is temporary flag to do experimental work until is is ok to be integrated to current `-skip-unused` one.

medvednikov:
And eventually `-skip-unused` will be removed as well since it'll be the default.

We'll only have `-no-skip-unused`

medvednikov:
I'm working on a flag to use libc for formatting instead of strconv

felipensp:
With latest changes:

```
$ v -cc gcc -skip-unused -o - examples/hello_world.v  | wc -l
5503
$ v -cc tcc -skip-unused -o - examples/hello_world.v  | wc -l
2037
```

JalonSolov:
Hmm... was it removing too much before?

felipensp:
> Hmm... was it removing too much before?

On tcc we can remove a lot... but on another ones no yet.. because the callstack code that uses string.Builder+strconv.

felipensp:
Now it cleans too on clang/gcc using `-d no_backtrace`:

```
$ v -cc tcc -d no_backtrace -skip-unused -o -  examples/hello_world.v   | wc -l
1935
$ v -cc gcc -d no_backtrace -skip-unused -o -  examples/hello_world.v   | wc -l
1935
$ v -cc gcc -d no_backtrace -skip-unused run  examples/hello_world.v  
Hello, World!
$ v -cc tcc -d no_backtrace -skip-unused run  examples/hello_world.v  
Hello, World!
```

vlib/v/markused/markused.v
```diff
@@ -25,119 +26,156 @@ pub fn mark_used(mut table ast.Table, mut pref_ pref.Preferences, ast_files []&a
...
-			// map methods
-			map_idx_str + '.get',
-			map_idx_str + '.set',
+			// array_idx_str + '.get_unsafe',
+			// array_idx_str + '.set_unsafe',			
+			// array_idx_str + '.first',
+			// array_idx_str + '.last',
+			// array_idx_str + '.pointers', // TODO: handle generic methods calling array primitives more precisely in pool_test.v
+			// array_idx_str + '.reverse',
+			// array_idx_str + '.repeat_to_depth',
```
spytheman, line 69: there is no point in keeping these as comments - just remove them

spytheman (approved):
Excellent work.

PR #22875, opened 2024-11-16
parser, checker: notify about unused function parameters

ttytm:
With changes code like the follow will result in a notification like shown below:

```v
fn foo(unused_var string) {
	println('foo')
}
```

```v
❯ v run foo.v
foo.v:1:8: notice: unused parameter: `unused_var`
    1 | fn foo(unused_var string) {
      |        ~~~~~~~~~~
    2 |     println('foo')
    3 |     a := 'a'
```

The change should allow for better detection of dead code. There are many situations where knowing about unused parameters can help resolve unintended behavior or simplify code, such as when removing historical parameter/argument residues after refactorings.

In the standard library, dead parameters are not removed in this PR; they were excluded from the notification. Removing this exclusion could help with the aforementioned benefits when working on the internal code.

Only the current tests have been updated without creating an additional explicit test, as I thought it might suffice to verify the behavior.

Since the change affects a series of test files, a note for the reviewer:
Actual code changes are only in `parser/fn.v` and `checker/checker.v`. Selecting those files or reviewing the test update commit (`Run autofix for compiler error tests`) separately might make things easier.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzM4OTJmZWU3Y2M1YTc4NTQyZDQ5MzYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.7yinEwkYiF_lqlMk6DASD--aneomIVIRlupUeeAIoDw">Huly&reg;: <b>V_0.6-21318</b></a></sub>

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -288,7 +288,9 @@ pub fn (mut c Checker) check_scope_vars(sc &ast.Scope) {
 			match obj {
 				ast.Var {
 					if !obj.is_used && obj.name[0] != `_` {
-						if !c.pref.translated && !c.file.is_translated {
+						if obj.is_arg {
+							c.note('unused parameter: `${obj.name}`', obj.pos)
+						} else if !c.pref.translated && !c.file.is_translated {
 							c.warn('unused variable: `${obj.name}`', obj.pos)
 						}
 					}
--- a/vlib/v/checker/tests/alias_map_unknown_op_overloading_err.out
+++ b/vlib/v/checker/tests/alias_map_unknown_op_overloading_err.out
@@ -1,3 +1,10 @@
+vlib/v/checker/tests/alias_map_unknown_op_overloading_err.vv:9:5: notice: unused parameter: `a`
+    7 | }
+    8 | 
+    9 | fn (a Map) + (b Map) Map {
+      |     ^
+   10 |     str := b['23']
+   11 |     return Map({
 vlib/v/checker/tests/alias_map_unknown_op_overloading_err.vv:19:10: error: undefined operation `Map` - `Map`
    17 |     mut a := new_map()
    18 |     b := new_map()
--- a/vlib/v/checker/tests/any_return_type_err.out
+++ b/vlib/v/checker/tests/any_return_type_err.out
@@ -1,3 +1,24 @@
+vlib/v/checker/tests/any_return_type_err.vv:4:5: notice: unused parameter: `f`
+    2 | }
+    3 | 
+    4 | fn (f Test) abc() any {
+      |     ^
+    5 |     return 0
+    6 | }
+vlib/v/checker/tests/any_return_type_err.vv:8:5: notice: unused parameter: `f`
+    6 | }
+    7 | 
+    8 | fn (f Test) abc2() ?any {
+      |     ^
+    9 |     return 0
+   10 | }
+vlib/v/checker/tests/any_return_type_err.vv:12:5: notice: unused parameter: `f`
+   10 | }
+   11 | 
+   12 | fn (f Test) abc3() !any {
+      |     ^
+   13 |     return 0
+   14 | }
 vlib/v/checker/tests/any_return_type_err.vv:4:19: error: cannot use type `any` here
     2 | }
     3 | 
--- a/vlib/v/checker/tests/array_builtin_redefinition.out
+++ b/vlib/v/checker/tests/array_builtin_redefinition.out
@@ -1,3 +1,10 @@
+vlib/v/checker/tests/array_builtin_redefinition.vv:7:5: notice: unused parameter: `a`
+    5 | // fn (a []Abc) repeat() int { return 0 }
+    6 | // fn (a []Abc) reverse() int { return 0 }
+    7 | fn (a []Abc) map() int {
+      |     ^
+    8 |     return 0
+    9 | }
 vlib/v/checker/tests/array_builtin_redefinition.vv:7:1: error: method overrides built-in array method
     5 | // fn (a []Abc) repeat() int { return 0 }
     6 | // fn (a []Abc) reverse() int { return 0 }
--- a/vlib/v/checker/tests/array_filter_fn_err.out
+++ b/vlib/v/checker/tests/array_filter_fn_err.out
@@ -1,3 +1,10 @@
+vlib/v/checker/tests/array_filter_fn_err.vv:19:16: notice: unused parameter: `b`
+   17 | }
+   18 | 
+   19 | fn fil1(a int, b int) bool {
+      |                ^
+   20 |     return a > 0
+   21 | }
 vlib/v/checker/tests/array_filter_fn_err.vv:2:28: error: function needs exactly 1 argument
     1 | fn main() {
     2 |     a1 := [1, 2, 3, 4].filter(fn (a int, b int) bool {
--- a/vlib/v/checker/tests/array_init_without_init_value_err.out
+++ b/vlib/v/checker/tests/array_init_without_init_value_err.out
@@ -1,3 +1,10 @@
+vlib/v/checker/tests/array_init_without_init_value_err.vv:35:5: notice: unused parameter: `a`
+   33 | }
+   34 | 
+   35 | fn (a LeStruct) give_string() string {
+      |     ^
+   36 |     return 'V'
+   37 | }
 vlib/v/checker/tests/array_init_without_init_value_err.vv:5:7: warning: arrays of sumtypes need to be initialized right away, therefore `len:` cannot be used (unless inside `unsafe`, or if you also use `init:`)
     3 | 
     4 | fn main_sum_type() {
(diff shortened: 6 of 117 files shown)


felipensp:
Good catch. :-)

felipensp:
Imho the receiver var must be skiped though.

spytheman:
That should not be the default.
It can be behind a custom flag.

spytheman:
Revert everything *except the check*, and add a custom flag for it `-check-unused-fn-args` or something similar.

JalonSolov:
Go _always_ notifies about unused parameters, but allows you to use `_` as a placeholder to suppress the warning, as in
```v
fn foo(_ string) {
	println('foo')
}
```
While you can argue that you shouldn't list a parameter at all if you aren't going to use it, sometimes this is needed if you're using an API from a package, and need to have the parameters listed, but don't use them yourself.

ttytm:
> Go _always_ notifies about unused parameters, but allows you to use `_` as a placeholder to suppress the warning, as in

The same would be possible with this change. Rust also notifies about them.

> While you can argue that you shouldn't list a parameter at all if you aren't going to use it, sometimes this is needed if you're using an API from a package, and need to have the parameters listed, but don't use them yourself.

Thats the case for public functions for go as well. Also I handled it that way. It would trigger for public functions.

ttytm:
> So this sounds like a v vet enhancement, rather than an extra option on V itself.

Yep rather a `v vet` instead of an additional flag. But it should be the default imo. 

Go errors for unused variables and warns about params. We do warn about variables (error in prod). The change would notify about unused params (allow them in prod). 

 @spytheman what do you think?

JalonSolov:
Warn regular, error in prod is a good thing.  If V is already doing that for variables, it really should be doing the same for parameters, too.

ttytm:
> Warn regular, error in prod is a good thing. If V is already doing that for variables, it really should be doing the same for parameters, too.

I think so too. This wouldn't change here. With the notify for params we would allow compilation in the same situations as Go does.

JalonSolov:
Yeah... I'm not really a fan of Go's way.  Probably because I have to check everything in via PR at work, and the PR always run linters, and the linters refuse my code if there are unused params, even if the code compiled clean locally.

That means I have to take the extra step to run the linters locally if I want a clean PR.  :-\

ttytm:
Having a notify by default should help to achieve that.

spytheman:
It should not be a default for the compiler.
Function/method declarations are API boundaries - they are a contract between the calling code (users), and the implementing code.
What happens inside a function, should not dictate what that API boundary looks like.

spytheman:
Local variable are *LOCAL*. They are not API boundaries, and changing/removing/adding more of them, does not change the API, and force cascading changes in unrelated parts of your codebase.

spytheman:
> Rust also notifies about them.

Can you show a specific example program, where it does so?

spytheman:
> Go linters always notify about unused parameters, but allows you to use _ as a placeholder to suppress the warning

If you can use `_` as a placeholder, so that the API is not changed, that is more acceptable.

spytheman:
Although replacing a meaningful parameter name with `_` to silence the notice/warning/error still loses information 🤔 .

spytheman:
Ok, I think I missed the nuance of what it does, and misunderstood it to introduce a very annoying noise, with no way to opt out, for a situation that happens often, if you have to adhere to a fixed API.

ttytm:
Btw regarding the pub visibility modifier: rust would warn also only warn about a local unused struct:

```rust
struct Foo {} // struct `Foo` is never constructed `#[warn(dead_code)]` on by default [dead_code]
pub struct Foo {} // no warning
```

spytheman:
The diff though is still too big - most of the .out files are now polluted with that unrelated notice :-| .

spytheman:
Go's defaults for warnings/errors are far for optimal - especially the unused local/import one is super annoying while prototyping.

spytheman:
> You taking into account that it would trigger only for local functions? If it's a pub fn the unused param name would not trigger the compiler message. 

Yes, I am now, but that is exactly what I missed on first reading, and why I closed the PR initially.

ttytm:
> The diff though is still too big - most of the .out files are now polluted with that unrelated notice :-| .

It would go down quite a bit if applying @felipensp s suggestion. Skipping it for receiver variables.

spytheman:
I still think that it should be behind a flag initially though.
It will help to smoothing out the transition period, as well as help with adjusting the behavior if needed.

JalonSolov:
Having that notice in all the .out files is just a sign that we've been missing this for too long.

ttytm:
> I still think that it should be behind a flag initially though.
It will help to smooth out the transition period, as well as help with adjusting the behavior if needed, without causing diff churn all the time that happens.

It will be hard to decide when the transition period ends. I think it's as @JalonSolov says in his last comment, getting that info is likely just overdue. I'll see how much the diff is reduced by applying the receiver var change.

Maybe then an as an on-by-default but adding a temporary opt-out vflag / env variable for a year?

spytheman:
No, I want an command line option, and for it to be off by default initially.

spytheman:
> Having that notice in all the .out files is just a sign that we've been missing this for too long.

Not it is not. The people that wrote those small examples were not careless, and did not miss/introduce bugs. The code examples in the .vv files are such, to introduce structure/setup an AST in such a way, as to trigger checker/parser errors.

It is entirely unrelated to this change, except that it shows, that it is not a default to be introduced without a grace period, in which it can be fine tuned and in which only people that know about it and actively work on it, will use it. 

I am sure that a lot of other codebases will have similar examples, and they are not errors there either - I do not want to break their CI jobs or workflows either.

spytheman:
> It will be hard to decide when the transition period ends when getting the info is opt-in.

I am thinking of a much shorter period, like a week or so - it is enough time to test it without breaking stuff too much, and to fine tune it, and to fix official V projects like v-analyzer.

After that, the flag will be removed, and it will be a default.

spytheman:
(that is because unlike `-Wimpure-v`, or the ?!Type split, which can force people to refactor/move code a lot, the notices here, will be mostly local in scope, and easy to fix with local changes)

JalonSolov:
I never said anything about people being careless writing tests.  I meant that if we had this sooner, we wouldn't "pollute" all those files now, with the extra message.

medvednikov:
I think @JalonSolov is right.

We can have an option not do this, and enable it for certain tests until the tests are updated (param name replaced with _).

medvednikov:
But it should be on by default.

ttytm:
With the linked PR running `v -check-unused-fn-args self` there are about 20 notifications for the vlib internals. For tests and examples there are a bit more ofcourse, but having applied Felipe's suggestion reduced the diff quite a bit.

---

What the compiler currently doesn't disallow for params is prefixing them with `_`. So it would also work to use `_param_name` to disable the notification for an unused `param_name`. This is how it would work on rust too. Personally I like that. What are your thoughts on that @medvednikov @spytheman @JalonSolov ?

JalonSolov:
We already use `_` alone to mean an "ignored" variable.  The same should be used for unused parameters.  There's no need for yet another convention just to handle this.

ttytm:
It could be limited to check the whole param name. Currently only the first char was checked since prefixing a variable with `_` is prohibited.

Giving into consideration: What I would like about it is that in a functions like 
https://github.com/vlang/v/blob/be4aec882a629cacb2836197d8e5c6b147115245/vlib/os/process_nix.c.v#L153-L155

I could keep the function signature intact with the prefixes. E.g. for the above:
```v
fn (mut p Process) win_read_string(_idx int, _maxbytes int) (string, int) {
	return '', 0
}
```
an alternative when the prefix is not possible to silence the notification would be:
```v
fn (mut p Process) win_read_string(idx int, maxbytes int) (string, int) {
	_, _ := idx, maxbytes
	return '', 0
}
```

JalonSolov:
How does having a name for a parameter that's never used make it "more meaningful"?

It doesn't matter at all what they were intended to be if they're never used in the code... not using them implies they were always meant to be ignored anyway, so why bother naming them?

If you are doing this to match an "outside" API (for example, a callback for a function in another module), you can always check the callback definition in the other module.

However, if it won't slow things down too much, then it really doesn't matter if you allow `_param_name`... although that still has to be processed by the parser, at least (even if just to skip to the end), which means it will be at least a bit slower than `_` by itself.

ttytm:
> How does having a name for a parameter that's never used make it "more meaningful"?
> 
> It doesn't matter at all what they were intended to be if they're never used in the code... not using them implies they were always meant to be ignored anyway, so why bother naming them?
> 
> If you are doing this to match an "outside" API (for example, a callback for a function in another module), you can always check the callback definition in the other module.
> 
> However, if it won't slow things down too much, then it really doesn't matter if you allow `_param_name`... although that still has to be processed by the parser, at least (even if just to skip to the end), which means it will be at least a bit slower than `_` by itself.

In general, I agree that `_` is sufficient for most cases. However, there are some instances where `_name` can be helpful. If there is a slowdown it should be negligible for those cases.

PR #22878, opened 2024-11-16
v: minor optimizations on `cmd/v`

felipensp:
This PR aims provide some optimization on cmd/v code.

image

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzM4YjI3ZmU3Y2M1YTc4NTQyZDRiNjIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.lqJnc70Jhk0COSJJAUF-Odtx7LCTnzgnU03TkzJPVhg">Huly&reg;: <b>V_0.6-21321</b></a></sub>

### DIFF
--- a/vlib/builtin/builtin.v
+++ b/vlib/builtin/builtin.v
@@ -24,6 +24,7 @@ struct VCastTypeIndexName {
 // will be filled in cgen
 __global as_cast_type_indexes []VCastTypeIndexName
 
+@[direct_array_access]
 fn __as_cast(obj voidptr, obj_type int, expected_type int) voidptr {
 	if obj_type != expected_type {
 		mut obj_name := as_cast_type_indexes[0].tname.clone()
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -2245,9 +2245,19 @@ pub fn (expr Expr) is_pure_literal() bool {
 
 pub fn (expr Expr) is_auto_deref_var() bool {
 	return match expr {
-		Ident { expr.obj is Var && expr.obj.is_auto_deref }
-		PrefixExpr { expr.op == .amp && expr.right.is_auto_deref_var() }
-		else { false }
+		Ident {
+			if expr.obj is Var {
+				expr.obj.is_auto_deref
+			} else {
+				false
+			}
+		}
+		PrefixExpr {
+			expr.op == .amp && expr.right.is_auto_deref_var()
+		}
+		else {
+			false
+		}
 	}
 }
 
@@ -2669,7 +2679,6 @@ pub fn (e Expr) is_nil() bool {
 	return e is Nil || (e is UnsafeExpr && e.expr is Nil)
 }
 
-@[direct_array_access]
 pub fn type_can_start_with_token(tok &token.Token) bool {
 	return match tok.kind {
 		.name {
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -2669,6 +2669,7 @@ pub fn (e Expr) is_nil() bool {
 	return e is Nil || (e is UnsafeExpr && e.expr is Nil)
 }
 
+@[direct_array_access]
 pub fn type_can_start_with_token(tok &token.Token) bool {
 	return match tok.kind {
 		.name {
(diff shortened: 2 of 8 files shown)


vlib/builtin/builtin.v
```diff
@@ -24,6 +24,7 @@ struct VCastTypeIndexName {
 // will be filled in cgen
 __global as_cast_type_indexes []VCastTypeIndexName
 
+@[direct_array_access]
```
spytheman, line 27: The [] only happens on the error path, where the end result is a panic, and outside a loop.
I am not sure that making it a few nanoseconds faster is useful 🤔 .

vlib/v/ast/ast.v
```diff
@@ -2245,9 +2245,19 @@ pub fn (expr Expr) is_pure_literal() bool {
...
+			} else {
+				false
+			}
+		}
+		PrefixExpr {
+			expr.op == .amp && expr.right.is_auto_deref_var()
+		}
+		else {
+			false
+		}
```
spytheman, line 2260: Why though? Is not the generated code the same?

vlib/v/checker/struct.v
```diff
@@ -177,8 +177,9 @@ fn (mut c Checker) struct_decl(mut node ast.StructDecl) {
 				c.check_valid_snake_case(field.name, 'field name', field.pos)
 			}
 			sym := c.table.sym(field.typ)
+			field_name, field_name_len := field.name, field.name.len
 			for j in 0 .. i {
-				if field.name == node.fields[j].name {
+				if field_name_len == node.fields[j].name.len && field_name == node.fields[j].name {
```
spytheman, line 182: That is a very small micro optimization, are you sure that it matters?
How many times is it executed for recompiling V itself?

vlib/v/gen/c/struct.v
```diff
@@ -229,8 +229,9 @@ fn (mut g Gen) struct_init(node ast.StructInit) {
 			g.is_shared = field.typ.has_flag(.shared_f)
 			if mut sym.info is ast.Struct {
 				mut found_equal_fields := 0
+				field_name, field_name_len := field.name, field.name.len
 				for mut sifield in sym.info.fields {
-					if sifield.name == field.name {
+					if sifield.name.len == field_name_len && sifield.name == field_name {
```
spytheman, line 234: Good, that is probably executed a lot more often than the other place for the declarations.

vlib/builtin/builtin.v
spytheman: Or is the goal to avoid the `v_fixed_index` generation for more programs?

vlib/v/checker/struct.v
felipensp: We can leave as is.

vlib/v/ast/ast.v
felipensp: Not. The current one calls __as_cast()

vlib/builtin/builtin.v
felipensp: Avoid array_get call is the goal.

vlib/v/checker/struct.v
JalonSolov: Even micro optimizations are good, and checking length before doing a string comparison is always a good optimization.  What is the objection to keeping this one?  That it makes the code slightly bigger?

vlib/v/ast/ast.v
spytheman: The current one is also clearer. Restore the code.

vlib/v/checker/struct.v
spytheman: > What is the objection to keeping this one? 

It makes the code less readable and more prone to bugs (keeping the names/branches synchronized), without a clear benefit in performance, since struct declarations are usually relatively few, compared to say function calls, or struct initialisations.

PR #22879, opened 2024-11-16
parser, checker, pref: allow getting notified about unused function params

ttytm:
The PR should implement #22875 as desired.

Example:

Screenshot_20241116_212257

NOTE: the v-analyzer hints won't work until this is on by default. For now, when compiling with `-check-unused-fn-args` you'll get:

```v
fn (unused_receiver Foo) method(unused_param int) {}

fn func(unused_param int) {}
```

```
❯ v -check-unused-fn-args run vlib/v/checker/tests/unused_param.vv
vlib/v/checker/tests/unused_param.vv:5:33: notice: unused parameter: `unused_param`
    3 | struct Foo {}
    4 | 
    5 | fn (unused_receiver Foo) method(unused_param int) {}
      |                                 ~~~~~~~~~~~~
    6 | 
    7 | fn func(unused_param int) {}
vlib/v/checker/tests/unused_param.vv:7:9: notice: unused parameter: `unused_param`
    5 | fn (unused_receiver Foo) method(unused_param int) {}
    6 | 
    7 | fn func(unused_param int) {}
      |         ~~~~~~~~~~~~
    8 | 
    9 | pub fn pub_func(unused_param int) {}
```

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzM4ZTIxNWU3Y2M1YTc4NTQyZDUxMjEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.nGz_aOKL4lO7OxrpWGkYgDEC0atMwC-1mZVkyuNh_tU">Huly&reg;: <b>V_0.6-21322</b></a></sub>

### DIFF
--- a/vlib/v/compiler_errors_test.v
+++ b/vlib/v/compiler_errors_test.v
@@ -8,6 +8,7 @@ import benchmark
 
 const skip_files = [
 	'non_existing.vv', // minimize commit diff churn, do not remove
+	'vlib/v/checker/tests/unused_param.vv',
 ]
 
 const skip_on_cstrict = [
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -289,7 +289,13 @@ pub fn (mut c Checker) check_scope_vars(sc &ast.Scope) {
 				ast.Var {
 					if !obj.is_used && obj.name[0] != `_` {
 						if !c.pref.translated && !c.file.is_translated {
-							c.warn('unused variable: `${obj.name}`', obj.pos)
+							if obj.is_arg {
+								if c.pref.show_unused_params {
+									c.note('unused parameter: `${obj.name}`', obj.pos)
+								}
+							} else {
+								c.warn('unused variable: `${obj.name}`', obj.pos)
+							}
 						}
 					}
 					if obj.is_mut && !obj.is_changed && !c.is_builtin_mod && obj.name != 'it' {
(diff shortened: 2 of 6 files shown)


StunxFS:
It would be great if `v fmt` could detect unused parameters and use `_` instead, thus automatically fixing all code with unused parameters.

Or, that there was a command, something like `v fix`, that could fix that problem automatically.

StunxFS:
> > It would be great if `v fmt` could detect unused parameters and use `_` instead, thus automatically fixing all code with unused parameters.
> 
> With this I see some situations where a format on save would replace the param name that is not in use yet but I want to use a few moments later.
> 
> 

So it would be better to have a special command for automatic fixes: `v fix`, and when the compiler finds a problem that can be fixed using `v fix`, it would recommend using it.

vlib/v/compiler_errors_test.v
```diff
@@ -8,6 +8,7 @@ import benchmark
 
 const skip_files = [
 	'non_existing.vv', // minimize commit diff churn, do not remove
+	'vlib/v/checker/tests/unused_param.vv',
```
ttytm, line 11: @spytheman 

This currently skips the test for automatic runs.

```
v -check-unused-fn-args run vlib/v/checker/tests/unused_param.vv
```

Can you give me a recommendation on how to include it in automatic test runs?

medvednikov:
> So it would be better to have a special command for automatic fixes: `v fix`, and when the compiler finds a problem that can be fixed using `v fix`, it would recommend using it.

That's a great idea, I've always wanted something like that. Can fix `mut params` as well etc

spytheman (approved):
Excellent work.

Issue #22859, opened 2024-11-14
arrays.sum([]) or {panic(err)} produces a misleading/obscure error message: 


holder66:
### Describe the bug

arrays.sum([]) or {panic(err)} returns "/v/vlib/arrays/arrays.v:253:4: error: operator += not defined on left operand type T"

### Reproduction Steps

arrays.sum([]) or {panic(err)}

### Expected Behavior

I would like to see an error message about the array being empty

### Current Behavior

v/vlib/arrays/arrays.v:253:4: error: operator `+=` not defined on left operand type `T`

### V version

Current V version: V 0.4.8 df51e84, timestamp: 2024-11-14 13:58:50 +0200

### Environment details (OS name and version, etc.)

V full version: V 0.4.8 dad8e3c.df51e84
OS: macos, macOS, 15.0.1, 24A348
Processor: 16 cpus, 64bit, little endian, Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz

getwd: /Users/henryolders
vexe: /Users/henryolders/v/v
vexe mtime: 2024-11-14 15:14:24

vroot: OK, value: /Users/henryolders/v
VMODULES: OK, value: /Users/henryolders/.vmodules
VTMP: OK, value: /tmp/v_507

env VFLAGS: "-use-os-system-to-run"

Git version: git version 2.47.0
Git vroot status: weekly.2024.06-1635-gdf51e840
.git/config present: true

CC version: Apple clang version 16.0.0 (clang-1600.0.26.4)
emcc version: N/A
thirdparty/tcc status: thirdparty-macos-amd64 975f1ad8

jorgeluismireles:
I think you need to specify the generic type of the argument. Once specified the type of an empty array then you got a panic with more information: `Cannot sum up array of nothing`. The generic type must overload the addition operator `+`. Types int, f32 and the like will work. For other types you need to do something like this:
```v
import arrays

struct Str {
	s string
}

fn (m Str) + (n Str) Str {
	return Str{ m.s + n.s }
}

s1 := arrays.sum([Str{'a'},Str{'b'},Str{'c'}])!
println('${s1.s}') // prints: 'abc'

if s2 := arrays.sum([]Str{}) {
	println('${s2.s}') //
} else {
	println('${err}') // prints: Cannot sum up array of nothing
}
```
See https://play.vlang.io/p/594631bd45

JalonSolov:
[...]

PR #22885, opened 2024-11-17
`checker`: error if cannot generic infer type from empty array `([]) `

Delta456:
Fixes https://github.com/vlang/v/issues/22859

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzM5YjhkZWU3Y2M1YTc4NTQyZDgyMDAiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.BlCDrzhqQLU4KEXJn8Q6nOx7frprbF7fghw28nk2VcQ">Huly&reg;: <b>V_0.6-21328</b></a></sub>


spytheman:
This should not be solved like this. The main problem is that the generic function expansion, produced an error, that is far removed from the actual source that the user wrote.

The compiler should just add more details to the error, so that the user can know what was the origin of the problem.

It was discussed here before: https://discord.com/channels/592103645835821068/592114487759470596/1306501688454287451 .

PR #21184, opened 2024-04-04
checker, json: return error on invalid string input instead of empty string result

ttytm:
With the PR we get errors for invalid Cjson input.

Currently, there are many occasions where decoding panics without an error. 
Or the main concern in the related issue, is an empty result return without an any error in case of using quotation marks when they shouldn't be used.

Atm:
```v
import json

struct TestStruct {
	host string
}

json.decode(TestStruct, "abc")! // Panics, but without an error value

//quote ⬇️ should not be there
txt := '"{"host": "localhost"}'
json.decode(TestStruct, txt)! // Currently an empty string
```
New errors:
```
error: json.decode: invalid json string
    5 | }
    6 | 
    7 | txt := '"{"host": "localhost"}'
      |        ~~~~~~~~~~~~~~~~~~~~~~~~
    8 | json.decode(TestStruct, txt)!
```

```
error: json.decode: invalid json string
    5 | }
    6 | 
    7 | json.decode(TestStruct, 'abc')!
      |                         ~~~~~
    8 | 
```
The change is simple but should hopefully do most validation needed for the module in regards to the nice progress that is being made on the json2 module.

Resolves #19801

PR #22898, opened 2024-11-18
cgen, json: make errors more informative (resolve empty panics)

ttytm:
The changes make error messages for `json.decode` more informative. The reason is that error messages we'll get when walking back on the cJSON error_ptr can currently be empty and become hard to localize. The example below would currently just print an empty line.

```v
json_txt := 'abc'
_ := json.decode(Foo2, json_txt) or {
	eprintln(err)
}
```

With the changes the errors message `failed to decode JSON string` is added. Similar to `failed to open file` during a `os.read_file` error.

---

This should help with the first of the two problems initially attempted to solve in #21184.
The second example in the linked PR could be addressed in a separate PR, where cgen is extended to add a check to verify the variable that should be decoded for validity.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzNhYzc5M2U3Y2M1YTc4NTQyZGEyODYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.fKmjsPLGXhGozekEA4iGg3JOp84kGic8WhUYxcgIGss">Huly&reg;: <b>V_0.6-21341</b></a></sub>

### DIFF
--- a/vlib/json/tests/json_decode_test.v
+++ b/vlib/json/tests/json_decode_test.v
@@ -90,7 +90,7 @@ struct DbConfig {
 
 fn test_decode_error_message_should_have_enough_context_empty() {
 	json.decode(DbConfig, '') or {
-		assert err.msg().len < 2
+		assert err.msg() == 'failed to decode JSON string'
 		return
 	}
 	assert false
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #22904, opened 2024-11-18
ci: check that more apps/modules do compile with -skip-unused

spytheman:


felipensp:
Cool. `vtcc` also uses `-skip-unused` on its `make.vsh`.

PR #22913, opened 2024-11-19
parser: fix block position's last line

yuyi98:
This PR fix block position's last line and modify the related call.

### DIFF
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -444,7 +444,7 @@ fn (f &Fmt) should_insert_newline_before_node(node ast.Node, prev_node ast.Node)
 				return false
 			}
 			ast.Block {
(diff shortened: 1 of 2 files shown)


spytheman:
Is there a specific example, that is formatted wrongly because of it?

yuyi98:
```v
			ast.Block {
				if node is ast.Block && !node.is_unsafe && node.pos.line_nr - prev_line_nr > 1 {
					return true
				} else {
					return false
				}
			}
```
`node.pos.line_nr - prev_line_nr > 1` Indicate blank line, originally it was `node.pos.line_nr - prev_line_nr > 0`.

yuyi98:
I see, I'll add one.

yuyi98:
```v
fn main() {
	{
		println('(1/3) My first small scope')
		//
	}

	println('hello')
}
```

spytheman:
Thank you.

PR #22920, opened 2024-11-20
os(windows): Use _wputenv instead of _putenv to stay in sync with _wgetenv.

Ekopalypse:
The use of _wputenv (os.setenv) ensures that environment variables read by _wgetenv (os.getenv) are returned correctly when ACP is used instead of the utf8 code page.

### DIFF
--- a/vlib/builtin/cfns.c.v
+++ b/vlib/builtin/cfns.c.v
@@ -326,7 +326,7 @@ fn C._wsystem(command &u16) int
 
 fn C._wgetenv(varname &u16) voidptr
 
-fn C._putenv(envstring &char) int
+fn C._wputenv(envstring &u16) int
 
 fn C._waccess(path &u16, mode int) int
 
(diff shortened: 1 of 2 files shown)


vlib/builtin/cfns.c.v
```diff
@@ -326,7 +326,7 @@ fn C._wsystem(command &u16) int
 
 fn C._wgetenv(varname &u16) voidptr
 
-fn C._putenv(envstring &char) int
+fn C._wputenv(envstring &u16) int
```
spytheman, line 329: imho restore the old declaration, there is no need for the new to replace it

spytheman: i.e. keep both

spytheman:
Please also add some tests for setting/getting env variables in `vlib/os/environment_test.v`, that put non ASCII characters in the values.

vlib/builtin/cfns.c.v
Ekopalypse: I can certainly do that, but the question is why? It's not used anywhere else. To make it more convenient for users when it is already defined?

spytheman: Because we do not know, if it is not used (except for the projects that are checked in the CI).

Ekopalypse: Done, but I think this is not really an advantage for these projects, as they might run into the original problem or they should rather use os.setenv instead of C._putenv.

PR #22931, opened 2024-11-21
cgen: fix option unwrap for fields of interface type (fixes #22930)

### LINKED ISSUE #22930: option unwrapping an interface results in compile error
### Describe the bug

Option unwrapping an interface results in a compile error.

### Reproduction Steps

```v
interface Interface {
	val int
}

struct Struct {
	val int
}

struct Foo {
	a ?Interface
	b ?Struct
}

fn main() {
	foo := Foo{
		a: Struct{val: 123}
		b: Struct{val: 456}
	}
	if foo.a != none {
		println(foo.a.val)
	}
	if foo.b != none {
		println(foo.b.val)
	}

}

```

### Expected Behavior

```
123
456
```

### Current Behavior

C Compile Error.

```
cc: \Temp\v_0\proj.c: In function 'main__main':
cc: \Temp\v_0\proj.c:14100:54: error: implicit declaration of function 'I_main__Interface_as_I_main__Interface'; did you mean 'I_main__App_to_Interface_main__Interface'? [-Werror=implicit-function-declaration]
cc: 14100 |                 main__Interface b = (*(main__Interface*)(I_main__Interface_as_I_main__Interface(app->my_opt_interface));
cc:       |                                                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
builder error:
==================
```

### Possible Solution

Check ``!is_option_unwrap`` in the interface cast if-statement in cgen.v

https://github.com/pisaiah/v/tree/patch-1

_No response_

### V version

V 0.4.8

Windows 11 x86-64
> [!NOTE]
>

pisaiah:
Fixes #22930 .

Adds a ```!is_option_unwrap``` check.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -4059,7 +4059,8 @@ fn (mut g Gen) selector_expr(node ast.SelectorExpr) {
 							g.write('*')
 						}
 						cast_sym := g.table.sym(g.unwrap_generic(typ))
-						if field_sym.kind == .interface && cast_sym.kind == .interface && !is_option_unwrap {
+						if field_sym.kind == .interface && cast_sym.kind == .interface
+							&& !is_option_unwrap {
 							ptr := '*'.repeat(field.typ.nr_muls())
 							dot := if node.expr_type.is_ptr() { '->' } else { '.' }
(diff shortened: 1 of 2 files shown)


Delta456 (approved):
Excellent work

PR #22942, opened 2024-11-22
net.http: build Host headers with port (fix #22941)

### LINKED ISSUE #22941: Request header Host ignores port
### Describe the bug

Currently the request header for Host is ignoring the Port, and only sending Hostname, this creates issues with servers that validate the headers to block access.

I'm running a Caddy instance, a freshly installed Caddy instance has the admin API located in `localhost:2019`, and it forbids access if the Host headers don't correctly match, so when I tried fetching it failed with status Forbidden.

### Reproduction Steps

http_bug.v:

```v
import net.http

fn main() {
  mut response := http.get('http://localhost:2019/config/') or { panic(err) }

  if response.status() != .ok {
  eprintln('Error executing start action. Status: ${response.status()}\n${response.body}')
  }
}
```

When ran against a Caddy instance:

```sh
./http_bug
Error executing start action. Status: Forbidden
{"error":"host not allowed: localhost"}
```

### Expected Behavior

The `Host` header should be `Host localhost:2019`, with that the response should be status 200 and the json with the Caddy current configuration.

### Current Behavior

It's sending `Host localhost` and failing validation.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.8 7d5e513

V full version: V 0.4.8 7d5e513
OS: macos, macOS, 15.1.1, 24B91
Processor: 12 cpus, 64bit, little endian, Intel(R) Core(TM) i5-10400F CPU @ 2.90GHz

getwd: /Users/hazer/v
vexe: /Users/hazer/v/v
vexe mtime: 2024-11-22 15:38:02

vroot: OK, value: /Users/hazer/v
VMODULES: OK, value: /Users/hazer/.vmodules
VTMP: OK, value: /tmp/v_501

Git vroot status: 7d5e513c
.git/config present: true

emcc version: N/A
thirdparty/tcc status: thirdparty-macos-amd64 975f1ad8

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzQwYTllZGQyZWY0Y2JjYzQ2ODAyZDQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.Qt2Spas_NbhBkeI0xXLQunxvjJbI54i2YQLgcEbLGi8">Huly&reg;: <b>V_0.6-21383</b></a></sub>
> [!NOTE]
>

Hazer:
This PR tries to fix the `Host` header, making sure that the port is included when we are not using ports 80 and 443. 

Ignoring the port breaks allow rules for servers that check the Host header, Caddy for instance uses it for the Admin API `localhost:2019/config`.

## Fixes
 - https://github.com/vlang/v/issues/22941

## Reproduction and manual testing

```v
import net.http

fn main() {
	mut response := http.get('https://anything.vithorio.codes:2033/path') or { panic(err) }

	if response.status() != .ok {
		eprintln('Error executing start action. Status: ${response.status()}\n${response.body}')
	}

	eprintln('result: ${response.body}')
}
```
Should print:
```
result: yay \o/ host anything.vithorio.codes 2033 anything.vithorio.codes:2033
```

While a bad result will print:
```
result: NO PORT anything.vithorio.codes  anything.vithorio.codes
```

And you can also double check with curl:
```sh
curl -X GET "https://anything.vithorio.codes:2033/path"
```

OBS: I should probably open it as a Draft, but I actually want some feedback and help planning the Unit Tests, it's actually my first day coding V, I'm loving it, but I'm still a bit lost with how to do stuff.

## TBD

- [x] Create unit tests (I really need help here, is the only way pointing to a real server?)
- [x] Test on Windows
- [x] Check the proxy settings, should they send the port too or they shouldn't

### DIFF
--- a/vlib/net/http/backend.c.v
+++ b/vlib/net/http/backend.c.v
@@ -33,7 +33,7 @@ fn net_ssl_do(req &Request, port int, method Method, host_name string, path stri
 		break
 	}
 
-	req_headers := req.build_request_headers(method, host_name, path)
+	req_headers := req.build_request_headers(method, host_name, port, path)
 	$if trace_http_request ? {
 		eprint('> ')
 		eprint(req_headers)
--- a/vlib/net/http/backend_vschannel_windows.c.v
+++ b/vlib/net/http/backend_vschannel_windows.c.v
@@ -20,7 +20,7 @@ fn vschannel_ssl_do(req &Request, port int, method Method, host_name string, pat
 	C.vschannel_init(&ctx)
 	mut buff := unsafe { malloc_noscan(C.vsc_init_resp_buff_size) }
 	addr := host_name
-	sdata := req.build_request_headers(method, host_name, path)
+	sdata := req.build_request_headers(method, host_name, port, path)
 	$if trace_http_request ? {
 		eprintln('> ${sdata}')
 	}
--- a/vlib/net/http/http_proxy.v
+++ b/vlib/net/http/http_proxy.v
@@ -85,9 +85,9 @@ fn (pr &HttpProxy) build_proxy_headers(host string) string {
 }
 
 fn (pr &HttpProxy) http_do(host urllib.URL, method Method, path string, req &Request) !Response {
-	host_name, _ := net.split_address(host.hostname())!
+	host_name, port := net.split_address(host.hostname())!
 
-	s := req.build_request_headers(req.method, host_name, path)
+	s := req.build_request_headers(req.method, host_name, port, path)
 	if host.scheme == 'https' {
 		mut client := pr.ssl_dial('${host.host}:443')!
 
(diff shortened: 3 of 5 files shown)


spytheman:
I've added a test in df63952, that uses V's own simple http server implementation, that allows setting a custom handler for the requests.

spytheman:
Most proxies do not care about the data that is actually sent/received, after the initial setup of the connection. They could inspect the passed headers and reject requests of course, but that is their policy decision, and http clients should not be concerned about that.

spytheman:
> Test on Windows
The CI already runs on macos, linux and windows.
That includes all `_test.v` files. 
`vlib/net/http/server_test.v` where I added the new test, is also run on all platforms.

Hazer:
Thank you! Yeah, using the server is a good idea, looks good to me then

PR #22952, opened 2024-11-23
gg: fix of issue ##21610

hedgeg0d:
Adding fontstash buffer cleanup when processing resize event. Increasing buffer size to 2048x2048.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzQyMDllYWQyZWY0Y2JjYzQ2ODQ4ZGQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.E2urX0oQEwJat7WBopePUIIktZf7LA-CUq5ofgT5_9E">Huly&reg;: <b>V_0.6-21394</b></a></sub>

### DIFF
--- a/cmd/tools/vvet/tests/array_init_one_val.out
+++ b/cmd/tools/vvet/tests/array_init_one_val.out
@@ -1,3 +1,3 @@
 cmd/tools/vvet/tests/array_init_one_val.vv:2: error: Use `1 == 1` instead of `1 in [1]`
 cmd/tools/vvet/tests/array_init_one_val.vv:6: error: Use `'foo' != bar` instead of `'foo' !in [bar]`
-Note: You can run `v fmt -w file.v` to fix these errors automatically
+Note: You can run `v fmt -w cmd/tools/vvet/tests/array_init_one_val.vv` to fix these errors automatically
--- a/cmd/tools/vvet/tests/indent_with_space.out
+++ b/cmd/tools/vvet/tests/indent_with_space.out
@@ -3,4 +3,4 @@ cmd/tools/vvet/tests/indent_with_space.vv:10: error: Looks like you are using sp
 cmd/tools/vvet/tests/indent_with_space.vv:17: error: Looks like you are using spaces for indentation.
 cmd/tools/vvet/tests/indent_with_space.vv:20: error: Looks like you are using spaces for indentation.
 cmd/tools/vvet/tests/indent_with_space.vv:22: error: Looks like you are using spaces for indentation.
-Note: You can run `v fmt -w file.v` to fix these errors automatically
+Note: You can run `v fmt -w cmd/tools/vvet/tests/indent_with_space.vv` to fix these errors automatically
--- a/cmd/tools/vvet/vvet.v
+++ b/cmd/tools/vvet/vvet.v
@@ -9,6 +9,7 @@ import v.parser
 import v.ast
 import v.help
 import term
+import arrays
 
 struct Vet {
 mut:
@@ -94,7 +95,8 @@ fn main() {
 		eprintln(vt.e2string(err))
 	}
 	if vfmt_err_count > 0 {
-		eprintln('Note: You can run `v fmt -w file.v` to fix these errors automatically')
+		filtered_out := arrays.distinct(vt.errors.map(it.file_path))
+		eprintln('Note: You can run `v fmt -w ${filtered_out.join(' ')}` to fix these errors automatically')
 	}
 	if vt.errors.len > 0 {
 		exit(1)


vlib/gg/gg.c.v
```diff
@@ -418,6 +418,7 @@ fn gg_event_fn(ce voidptr, user_data voidptr) {
 			if ctx.config.resized_fn != unsafe { nil } {
 				ctx.config.resized_fn(e, ctx.config.user_data)
 			}
+			C.fonsResetAtlas(ctx.ft.fons, 2048, 2048)
```
spytheman, line 421: Why should that happen on every resize event, and not just once?

hedgeg0d: because the font size may change after every resize event

spytheman: It may as well not change too - that is specific to the user code.
`examples/gg/minimal.v` for example or `examples/gg/raycaster.v`, do not scale their fonts on each resize event.

I think that calling that indiscriminately, while the user drags the border of the window will be too slow for no good benefit 🤔 .

spytheman: I think that a better approach would be to use this:
```v
pub fn (s &Context) set_error_callback(callback fn (voidptr, int, int), uptr voidptr) {
    C.fonsSetErrorCallback(s, callback, uptr)
}
```

so that we can hook into the situations, where the atlas becomes unusable, and reset it only in those cases.

hedgeg0d: It seems to me that the window size does not change so often that it somehow significantly affects performance, but this bug really makes itself felt in applications where a lot of labels are rendered at the same time, especially in large font (the buffer stores pixels, so the larger the font, the faster it overflows). But if this is important, maybe it's worth just increasing the buffer size without clearing it?

spytheman: That can be done in `fn new_ft(c FTConfig) ?&FT {` in vlib/gg/text_rendering.c.v .

hedgeg0d: > I think that a better approach would be to use this:
> 
> ```v
> pub fn (s &Context) set_error_callback(callback fn (voidptr, int, int), uptr voidptr) {
>     C.fonsSetErrorCallback(s, callback, uptr)
> }
> ```
> 
> so that we can hook into the situations, where the atlas becomes unusable, and reset it only in those cases.

Indeed, I did not think that it was possible to track the error. It's really more optimal.

spytheman: No, clearing is good - it acts as a cache, and in time caches accumulate all kinds of cruft, that may not be already needed (the previous frames for example, during resizing, or after animations involving the font size, may have caused the atlas to become filled with rendered glyphs that are now obsolete).

The only problem though is that it should not happen just because the window is resized, but because the atlas is full, or there was a bug otherwise, that hopefully will not exist anymore in the next frames, since the atlas will be reset.

spytheman: Try for example `v run examples/gg/minimal.v` on master and on this PR to see the effect.

hedgeg0d: Sorry, i don't know how to cast voidptr to C.FONScontext. How do I do this?

```
pub fn (s &C.FONScontext) set_error_callback(callback fn (voidptr, int, int), uptr voidptr) {
    C.fonsSetErrorCallback(s, callback, uptr)
}

pub fn clear_atlas_callback(uptr voidptr, error int, val int) {
	ft := C.FONScontext{uptr} // doesn't work
	fons := ft.fons
	C.fonsResetAtlas(fons, 512, 512)
}```

Comment (StunxFS) [reply vlib/gg/gg.c.v:421]: > Sorry, i don't know how to cast voidptr to C.FONScontext. How do I do this?
> 
> ```
> pub fn (s &C.FONScontext) set_error_callback(callback fn (voidptr, int, int), uptr voidptr) {
>     C.fonsSetErrorCallback(s, callback, uptr)
> }
> 
> pub fn clear_atlas_callback(uptr voidptr, error int, val int) {
> 	ft := C.FONScontext{uptr} // doesn't work
> 	fons := ft.fons
> 	C.fonsResetAtlas(fons, 512, 512)
> }
> ```

`&C.FONScontext(uptr)`

Comment (hedgeg0d) [reply vlib/gg/gg.c.v:421]: Ok, I made a callback to clear the buffer when it's full, but now I'm getting messages from sokol:

...
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 0, item_id: 298, fname: sokol_gfx.h, line: 16102, message: VALIDATION_FAILED: validation layer checks failed
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 1, item_id: 270, fname: sokol_gfx.h, line: 16874, message: VALIDATE_ABND_FS_IMG_EXISTS: sg_apply_bindings: image bound to fragment stage no longer alive
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 0, item_id: 298, fname: sokol_gfx.h, line: 16102, message: VALIDATION_FAILED: validation layer checks failed
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 1, item_id: 270, fname: sokol_gfx.h, line: 16874, message: VALIDATE_ABND_FS_IMG_EXISTS: sg_apply_bindings: image bound to fragment stage no longer alive
sokol.memory.slog | user_data: (nil), const_tag: sg, level: 0, item_id: 298, fname: sokol_gfx.h, line: 16102, message: VALIDATION_FAILED: validation layer checks failed
```

Code:
```
pub fn (s &C.FONScontext) set_error_callback(callback fn (voidptr, int, int), uptr voidptr) {
    C.fonsSetErrorCallback(s, callback, uptr)
}

pub fn clear_atlas_callback(uptr voidptr, error int, val int) {
	if error == 1 {
		ft := unsafe{&FT(uptr)}
		fons := ft.fons
		C.fonsResetAtlas(fons, buff_size, buff_size)
	}
}

fn new_ft(c FTConfig) ?&FT {
    ...
    ft := &FT{
		fons:        fons
		font_normal: fons.add_font_mem('sans', bytes, false)
		font_bold:   fons.add_font_mem('sans', bytes_bold, false)
		font_mono:   fons.add_font_mem('sans', bytes_mono, false)
		font_italic: fons.add_font_mem('sans', bytes_italic, false)
		scale:       c.scale
	}
	fons.set_error_callback(clear_atlas_callback, ft)
	return ft
}

...

But apps works just fine

hedgeg0d: the effect of resizing the window in v run `examples/gg/minimal.v` is gone

spytheman: excellent, please push your changes

Issue #22947, opened 2024-11-23
C error on `Option` casting in `assert`  statement


gechandesu:
**V doctor:**

```
V full version: V 0.4.8 5ec9bb5.393e4ea
OS: linux, Linux version 6.6.8-arch1-1 (linux@archlinux) (gcc (GCC) 13.2.1 20230801, GNU ld (GNU Binutils) 2.41.0) #1 SMP PREEMPT_DYNAMIC Thu, 21 Dec 2023 19:01:01 +0000
Processor: 16 cpus, 64bit, little endian, 12th Gen Intel(R) Core(TM) i5-1240P

getwd: /home/ge/Code/v/netaddr/issues
vexe: /home/ge/.vlang/v
vexe mtime: 2024-11-23 13:05:19

vroot: OK, value: /home/ge/.vlang
VMODULES: OK, value: /home/ge/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.46.1
Git vroot status: weekly.2024.47-20-g393e4ea8
.git/config present: true

CC version: cc (GCC) 14.2.1 20240910
emcc version: N/A
thirdparty/tcc status: thirdparty-linux-amd64 0134e9b9
```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg assert_option_cast.v`

```v
module main

fn get_opt(a int) ?string {
	if a < 0 {
		return none
	}
	return 'success'
}

fn main() {
	x := get_opt(1)
	println('${x as string}') // OK
	assert x as string == 'success' // compilation error
}
```

**What did you expect to see?**

Assertion passed, no C errors

**What did you see instead?**

```
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/assert_option_cast.01JDCMNCJXMVPQ05YX7R8SGTA6.tmp.c:13734: error: cannot convert 'struct _option_string' to 'struct string'
=================================================================
(You can pass `-cg`, or `-show-c-output` as well, to print all the C error messages).
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

[...]

PR #22953, opened 2024-11-23
cgen: allow unwrapping of `x as string` expr 

Delta456:
Fixes https://github.com/vlang/v/issues/22947

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzQyMjJmY2QyZWY0Y2JjYzQ2ODRiZDkiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0._2YzKl56B4_Hh55if6vYLruFEFsgkKjlu8781bKygZc">Huly&reg;: <b>V_0.6-21395</b></a></sub>


vlib/v/gen/c/utils.v
```diff
@@ -145,3 +145,27 @@ fn (mut g Gen) dot_or_ptr(val_type ast.Type) string {
 		'.'
 	}
 }
+
+@[inline]
```
felipensp, line 149: I think it is too long for `@[inline]`

vlib/v/gen/c/cgen.v
```diff
@@ -7801,7 +7782,17 @@ fn (mut g Gen) as_cast(node ast.AsCast) {
 			g.as_cast_type_names[idx] = variant_sym.name
 		}
 	} else {
-		g.expr(node.expr)
+		if node.expr is ast.Ident {
+			if node.expr.info is ast.IdentVar {
```
felipensp, line 7786: What happens if it is not a IdentVar?

vlib/v/gen/c/utils.v
Delta456: Right

vlib/v/gen/c/cgen.v
Delta456: Not sure if `ast.IdentFn` needs to unwrap the option type as this already works

```v
fn get_opt_fn(a int) ?fn(int) int {
	if a < 0 {
		return none
	}
	return fn (n int) int {
		return n + n
	}
}

fn main() {
	d := get_opt_fn(12)
	dump(d as fn(int) int)
	
	e := d as fn(int) int
	dump(e(12))
	
}
```

though `dump((d as fn(int) int)(12))` doesn't work due to the parser error. I don't think this would be valid.

spytheman: The problem is that the expression will not get checked at all in that case @Delta456 , while previously it was.

spytheman: i.e. add `} else { g.expr(node.expr) }` after the `if node.expr.info is ast.IdentVar {` block.

spytheman:
(rebased to resolve the merge conflict with master)

vlib/v/tests/option_unwrap_as_cast_test.v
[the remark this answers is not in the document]
Delta456: Good catch. I will PR this later.

Issue #21610, opened 2024-05-30
Fonts breaks when resizing gg window


Cieciak:
### Describe the bug

When resizing window font changes shape or vanishes

### Reproduction Steps

After running, resize window using mouse and font breaks, small changes have dramatic effect
```
module main

import gg
import gx

struct App {
    pub mut:
        context &gg.Context

        vw int
        vh int
}

fn main() {
    mut app := App{
        context: &gg.Context{}
    }

    app.context = gg.new_context(
        bg_color: gx.rgb(0x44, 0x44, 0x44)
        window_title: 'Text test'
        frame_fn: frame
        resized_fn: resize_handler
        user_data: &app
    )
    app.context.run()
}

fn resize_handler(event &gg.Event, mut app App){
    app.vw = event.window_width
    app.vh = event.window_height
}

fn frame(mut app App) {
    app.context.begin()

    cfg := gx.TextCfg{
        size: app.vh / 5
    }

    app.context.draw_text(100, 100, 'Hello There!', cfg)
    app.context.end()
}
```
![pic1](https://github.com/vlang/v/assets/60304529/9403a303-fda8-4727-91ee-901adf9dae0d)
![pic2](https://github.com/vlang/v/assets/60304529/0821e144-af1b-4204-b2dd-5346a536b507)
![pic3](https://github.com/vlang/v/assets/60304529/2e5ed51b-996e-428b-a990-ade107dad174)
Each image has slightly different resolution

### Expected Behavior

The font should stay consistent

### Current Behavior

Font does not stay consistent, as presented on pic1 to pic3

### Additional Information/Context

I use 4K display, but checked it on windows, linux and mac on different resolutions and fonts, all broken

### V version

V 0.4.6 52dfef1

### Environment details (OS name and version, etc.)

V full version: V 0.4.6 40da0f7.52dfef1
OS: linux, Zorin OS 17.1
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i5-4690K CPU @ 3.50GHz

getwd: /home/niko
vexe: /home/niko/v/v
vexe mtime: 2024-05-30 15:55:44

vroot: OK, value: /home/niko/v
VMODULES: OK, value: /home/niko/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.34.1
Git vroot status: weekly.2023.51-1050-g52dfef1a
.git/config present: true

[...]

PR #22959, opened 2024-11-24
gg: fix of issue #21610

hedgeg0d:
Adding a callback to clear the fontstash buffer, when it is full. The buffer size increases to 2048x2048

Updated https://github.com/vlang/v/pull/22952

### DIFF
--- a/vlib/gg/text_rendering.c.v
+++ b/vlib/gg/text_rendering.c.v
@@ -21,10 +21,23 @@ pub mut:
 	scale     f32 = 1.0
 }
 
+const buff_size = 2048
+
+pub fn (s &C.FONScontext) set_error_callback(callback fn (voidptr, int, int), uptr voidptr) {
+	C.fonsSetErrorCallback(s, callback, uptr)
+}
+
+pub fn clear_atlas_callback(uptr voidptr, error int, val int) {
+	if error == 1 { // atlas overflow error code
+		fons := unsafe { &fontstash.Context(uptr) }
+		C.fonsResetAtlas(fons, buff_size, buff_size)
+	}
+}
+
 fn new_ft(c FTConfig) ?&FT {
 	if c.font_path == '' {
 		if c.bytes_normal.len > 0 {
-			fons := sfons.create(512, 512, 1)
+			fons := sfons.create(buff_size, buff_size, 1)
 			bytes_normal := c.bytes_normal
 			bytes_bold := if c.bytes_bold.len > 0 {
 				c.bytes_bold
@@ -44,7 +57,7 @@ fn new_ft(c FTConfig) ?&FT {
 				debug_font_println('setting italic variant to normal')
 				bytes_normal
 			}
-
+			fons.set_error_callback(clear_atlas_callback, fons)
 			return &FT{
 				fons:        fons
 				font_normal: fons.add_font_mem('sans', bytes_normal, false)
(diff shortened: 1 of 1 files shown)


vlib/gg/text_rendering.c.v
```diff
@@ -21,10 +21,23 @@ pub mut:
 	scale     f32 = 1.0
 }
 
+const buff_size = 2048
+
+pub fn (s &C.FONScontext) set_error_callback(callback fn (voidptr, int, int), uptr voidptr) {
+	C.fonsSetErrorCallback(s, callback, uptr)
+}
```
spytheman, line 28: This is not needed, `set_error_callback` is already defined in `vlib/fontstash/fontstash.c.v:45:pub fn (s &Context) set_error_callback(callback fn (voidptr, int, int), uptr voidptr) {`.

vlib/gg/text_rendering.c.v
```diff
@@ -21,10 +21,23 @@ pub mut:
...
+
+pub fn clear_atlas_callback(uptr voidptr, error int, val int) {
+	if error == 1 { // atlas overflow error code
+		fons := unsafe { &fontstash.Context(uptr) }
+		C.fonsResetAtlas(fons, buff_size, buff_size)
```
spytheman, line 33:
```suggestion
		fons.reset_atlas(buff_size, buff_size)
```

vlib/gg/text_rendering.c.v
spytheman, line 28:
```suggestion
```

vlib/gg/text_rendering.c.v
```diff
@@ -21,10 +21,19 @@ pub mut:
 	scale     f32 = 1.0
 }
 
+const buff_size = 2048
+
+pub fn clear_atlas_callback(uptr voidptr, error int, val int) {
```
spytheman, line 26:
```suggestion
fn clear_atlas_callback(uptr voidptr, error int, val int) {
```

vlib/gg/text_rendering.c.v
```diff
@@ -21,10 +21,19 @@ pub mut:
 	scale     f32 = 1.0
 }
 
+const buff_size = 2048
```
spytheman, line 24:
```suggestion
const buff_size = int($d('gg_text_buff_size', 2048))
```

vlib/gg/text_rendering.c.v
```diff
@@ -21,10 +21,19 @@ pub mut:
 	scale     f32 = 1.0
 }
 
+const buff_size = int($d('gg_text_buff_size', 2048))
```
spytheman, line 24: I am not sure if 2048 is a good default. Running `examples/gg/minimal.v`, it seems to result in apps that use ~8MB more RAM by default on my machine, compared to the same but with 512 (on master):
```
delian     14166  3.5  0.2 141112 40524 pts/0    Sl+  14:24   0:00          |       |       \_ /home/delian/code/v/rrrrrrrrrrrrrr
```
vs
```
delian     14082  3.6  0.3 166648 47672 pts/0    Sl+  14:23   0:00          |       |       \_ /home/delian/code/v/rrrrrrrrrrrrrr
```

spytheman: On the other hand, it is certainly better in the sense that apps will not have weird artifacts and disappearing texts in much more cases, and with $d(), now people can tweak it easily, if they need something smaller or bigger.

spytheman (approved):
Excellent work, gg apps will have less text rendering problems by default because of it.

PR #22964, opened 2024-11-24
cgen: fix codegen for option on concatexpr

### LINKED ISSUE #22951: Wrong behavior of `!(Type, ?Type)` result tuple
### Describe the bug

Subject

### Reproduction Steps

```v
module main

fn foo(f int) !(int, ?int) {
	if f < 0 {
		return error('f is negative')
	}
	if f > 0 {
		return 1, f
	}
	return 0, none
}

fn main() {
	a, b := foo(-1) or { 2, 2 } // since foo arg is negative int, error must occurs
	c := b as int
	assert a == 2
	assert c == 2

	a2, b2 := foo(0) or { 2, 2 }
	c2 := b2 as int
	assert a2 == 0
	assert c2 == 0

	a3, b3 := foo(1) or { 2, 2 }
	c3 := b3 as int
	assert a3 == 1
	assert c3 == 1
}
```

### Expected Behavior

In shot: all assertions must be passed.

Values listed in `or {}` must be applied here since `foo(-1)` raises an error:

```v
a , b := foo(-1) or { 2, 2 }

c := b as int
assert a == 2
assert c == 2
```

In `foo(0)` case `b` must be `Option(none)`:

```v
a , b := foo(0) or { 2, 2 }

c := b as int // Option(none) casts to 0
assert a == 0
assert c == 0
```

Finally:

```v
a , b := foo(1) or { 2, 2 }

c := b as int
assert a == 1
assert c == 1
```

### Current Behavior

```v
a, b := foo(-1) or { 2, 2 } // since foo arg is negative int, error must occurs
println(a) // output: 2
println(b) // output: Option(none) but Option(2) expected
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.8 393e4ea

V full version: V 0.4.8 5ec9bb5.393e4ea
OS: linux, Linux version 6.6.8-arch1-1 (linux@archlinux) (gcc (GCC) 13.2.1 20230801, GNU ld (GNU Binutils) 2.41.0) #1 SMP PREEMPT_DYNAMIC Thu, 21 Dec 2023 19:01:01 +0000
Processor: 16 cpus, 64bit, little endian, 12th Gen Intel(R) Core(TM) i5-1240P

getwd: /home/ge/Code/v/netaddr/issues
vexe: /home/ge/.vlang/v
vexe mtime: 2024-11-23 13:05:19

vroot: OK, value: /home/ge/.vlang
VMODULES: OK, value: /home/ge/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.46.1
Git vroot status: weekly.2024.47-20-g393e4ea8
.git/config present: true

CC version: cc (GCC) 14.2.1 20240910
emcc version: N/A

> [!NOTE]
>

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzQyMDI3OWQyZWY0Y2JjYzQ2ODQ4M2EiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.lbp89whkozqR-zPkqg92ZbAWTyQ5sq5GriH1Dd7Xf-M">Huly&reg;: <b>V_0.6-21393</b></a></sub>

felipensp:
Fix #22951

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzQzNzI2M2QyZWY0Y2JjYzQ2ODcyOWYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.mqeeNdg3gDMJ0s69W3-jOwuvtI_B4dmYoKJyfuM0tnI">Huly&reg;: <b>V_0.6-21405</b></a></sub>

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5306,24 +5306,32 @@ fn (mut g Gen) cast_expr(node ast.CastExpr) {
 }
 
 fn (mut g Gen) concat_expr(node ast.ConcatExpr) {
-	mut styp := g.styp(node.return_type.clear_option_and_result())
+	mut typ := node.return_type.clear_option_and_result()
 	if g.inside_return {
-		styp = g.styp(g.fn_decl.return_type.clear_option_and_result())
+		typ = g.fn_decl.return_type.clear_option_and_result()
 	} else if g.inside_or_block {
-		styp = g.styp(g.or_expr_return_type.clear_option_and_result())
+		typ = g.or_expr_return_type.clear_option_and_result()
 	}
+	styp := g.styp(typ)
 	sym := g.table.sym(node.return_type)
 	is_multi := sym.kind == .multi_return
 	if !is_multi {
 		g.expr(node.vals[0])
 	} else {
+		types := (g.table.sym(typ).info as ast.MultiReturn).types
 		g.write('(${styp}){')
 		for i, expr in node.vals {
 			g.write('.arg${i}=')
-			old_left_is_opt := g.left_is_opt
-			g.left_is_opt = true
-			g.expr(expr)
-			g.left_is_opt = old_left_is_opt
+			if types[i].has_flag(.option) {
+				old_left_is_opt := g.left_is_opt
+				g.left_is_opt = true
+				g.write('{.data=')
+				g.expr(expr)
+				g.write('}')
+				g.left_is_opt = old_left_is_opt
+			} else {
+				g.expr(expr)
+			}
 			if i < node.vals.len - 1 {
 				g.write(',')
 			}
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5322,15 +5322,15 @@ fn (mut g Gen) concat_expr(node ast.ConcatExpr) {
 		g.write('(${styp}){')
 		for i, expr in node.vals {
 			g.write('.arg${i}=')
-			if types[i].has_flag(.option) {
-				old_left_is_opt := g.left_is_opt
-				g.left_is_opt = true
+			if types[i].has_flag(.option) && expr.is_literal() {
 				g.write('{.data=')
 				g.expr(expr)
 				g.write('}')
-				g.left_is_opt = old_left_is_opt
 			} else {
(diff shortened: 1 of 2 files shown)


Delta456:
This shouldn't be working as the language already disallows `!?type`. We also discussed this in discord.

spytheman:
> This shouldn't be working as the language already disallows `!?type`. We also discussed this in discord.

This is not `!?Type` however.

It is a result tuple, that just happens to contain an option type. Option type values should have first class support in V, and so should be able to be used everywhere other primitive types are allowed, including in result tuples.

spytheman:
The problem with `!?Type`, is that it has 3 possible kinds of values, while the language has  constructs like or blocks and `if x := f() {`, that can deal well with 2 at a time, and it is very confusing to use them, when there are 3 - one of those 3 will get forgotten, or cause bugs.

The only construct that can deal with 3 at the same time, is `match x {`, but that is very verbose to be used in all places that universal support of the old `?!Type` combination would need.

For result tuples/multiple result values, with options inside them, there are no such problems, because the outer result, will get handled right away in one construct, and then what remains will be a normal option value, that can then be handled with the usual V constructs:
```v
fn f() !(int, ?int) { return 1,2 }

a,b := f()!
// a will be `int(1)` and b will be `?int(2)`, and you can do `b or { ... }` etc
```

spytheman (approved):
Excellent work.

PR #22948, opened 2024-11-23
x.encoding.asn1: improve performance

blackshirt:
After some preliminary profiling, i think there are some overhead on the `field_options` parsing and common parsing routine.
This PR tries to improve this situation by

- Removes some redundant check on `FieldOptions` parsing, especially on `trim_space` call
- Removes redundant check on common parsing routine,
- Call underlying internal routine directly on encode (decode) when no-options string was passed
- Removes redundant check on wrapping (unwrapping) routine

In my foolish benchmark below, there are show up on a (minor) improvement. This is my result numbers on benchmarking 4 functons, ie, `encode`, `encode_with_options`, `decode` and `decode_with_options`

**Before patch**

```v
Iterations: 1000                Total Duration: 809.674us       ns/op: 809      B/op: 61        allocs/op: 65
Iterations: 1000                Total Duration: 15.912ms        ns/op: 15912    B/op: 32        allocs/op: 19
Iterations: 1000                Total Duration: 1.122ms 	ns/op: 1122     B/op: 49        allocs/op: 62
Iterations: 1000                Total Duration: 17.895ms        ns/op: 17895    B/op: 118       allocs/op: 164
```

**After patch**

```v
Iterations: 1000                Total Duration: 554.411us       ns/op: 554      B/op: 61        allocs/op: 65
Iterations: 1000                Total Duration: 7.219ms 	ns/op: 7219     B/op: 299       allocs/op: 302
Iterations: 1000                Total Duration: 813.360us       ns/op: 813      B/op: 49        allocs/op: 62
Iterations: 1000                Total Duration: 8.606ms 	ns/op: 8606     B/op: 348       allocs/op: 401
```

I think the number show and tells us where is the small improvement , ie, reduced total duration of execution, reduced ns/op, and relativetly increasing number in B/op and allocs/op.
Especially when your routine calls `encode_with_options` and/or `decode_with_options`, its relatively improved with this patch.

The benchmark;s code :

```v
import x.benchmark
import x.encoding.asn1

const number = asn1.Integer.from_int(0xff)
const number_encoded = [u8(2), 2, 0, 255]
const number_encoded_with_options = [u8(160), 4, 2, 2, 0, 255]

fn encode_integer() ! {
	_ := asn1.encode(number)!
}

fn encode_integer_with_options() ! {
	_ := asn1.encode_with_options(number, 'context_specific:0;explicit;inner:2')!
}

fn decode_integer() ! {
	el := asn1.decode(number_encoded)!
[...]

PR #22932, opened 2024-11-21
Cleans up unneceesary bytes from `x.encoding.asn` module

blackshirt:
This PR contains some small cleans up of unnecessary bytes from `x.encoding.asn1` module, includes :

- Cleaning up some examples to align with recent fixed issues by greats work by @felipensp , [22901](https://github.com/vlang/v/issues/22901), [22855](https://github.com/vlang/v/issues/22855), [22721](https://github.com/vlang/v/issues/22721)
- Removed extra test (this is not really needed), it test is for x509 certificates, where i have it on other branch.
- Clearing some bits of unnecessary pieces of code.

Thanks,

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzNlZjBlNGU3Y2M1YTc4NTQyZWE4NTkiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.wwWjutrQ0bBspmHNoz4ZBVVf_fWfTbOjW8U4PFkAZQk">Huly&reg;: <b>V_0.6-21374</b></a></sub>

PR #22861, opened 2024-11-14
encoding/asn1: fix time creation to also accommodate negative timezone offsets

kimshrier:
When splitting the string representation of the time, the code did not previously allow for negative timezone offsets.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzM2MmRmMWU3Y2M1YTc4NTQyY2ViYmEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.dtHSbpN4XE_CBKgTuN3iaOQsWbLx36X-6_mwGi1u7rM">Huly&reg;: <b>V_0.6-21304</b></a></sub>

PR #22895, opened 2024-11-17
v: add selector option unwrapping on if `!= none`

felipensp:
Fix #22892

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzNhNTBlM2U3Y2M1YTc4NTQyZDk1OGYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.9sxQVkpZg4nCDnHbLj5ftvGGPh3yprk_jHKesUnXXio">Huly&reg;: <b>V_0.6-21338</b></a></sub>

PR #22968, opened 2024-11-25
`x.encoding.asn1`: another minor  cleanup

blackshirt:
This PR is rather of
follow-up from previous PR in attempt to improves `x.encoding.asn1` module, as availables  in [22948](https://github.com/vlang/v/pull/22948), [22932](https://github.com/vlang/v/pull/22932), [22861](https://github.com/vlang/v/pull/22861), and [22847](https://github.com/vlang/v/pull/22847) PRs.
This PR contains some small changes to the module:

- Refinement to align with the latest fix of Option unwrapping. (mostly from @felipensp, thank guys)
- Just removes `.str()` from sequence and set for now, its needed to allow just dump current sequence fields.
- Unifies some of `RawElement` methods
- Some small refinements.

The latest bench run with `-prod` flag results in

```v
Benchmarking ASN.1 encode...
Average example encode time: 1 µs
Benchmarking ASN.1 decode (with asn1.decode)...
Average (asn1.decode) decode time: 1 µs
Benchmarking ASN.1 decode with Example.decode)...
Average (Example.decode) decode time: 1 µs
```

Its likely on par with the `go` version bench availables on `bench` dir. The previous patch help reduces some overhead of it.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzQ0MTM5NWQyZWY0Y2JjYzQ2ODk2NjYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.F1jx8tLoVe3St_QIHHTjvAFqKjq-1RRKt3XW4S81Pkg">Huly&reg;: <b>V_0.6-21409</b></a></sub>

### DIFF
--- a/vlib/x/encoding/asn1/field_options.v
+++ b/vlib/x/encoding/asn1/field_options.v
@@ -115,10 +115,11 @@ pub fn FieldOptions.from_attrs(attrs []string) !FieldOptions {
 
 	// The item has space-trimmed
 	for item in filtered {
-		if !is_tag_marker(item) && !is_optional_marker(item) && !is_default_marker(item)
-			&& !is_mode_marker(item) && !is_inner_tag_marker(item) {
-			return error('unsupported keyword')
-		}
+		// no need to check this again, its already filtered
+		// if !is_tag_marker(item) && !is_optional_marker(item) && !is_default_marker(item)
+		//		&& !is_mode_marker(item) && !is_inner_tag_marker(item) {
+		//		return error('unsupported keyword')
+		// }
 		if is_tag_marker(item) {
 			cls, num := parse_tag_marker(item)!
 			tag_ctr += 1
--- a/vlib/x/encoding/asn1/field_options.v
+++ b/vlib/x/encoding/asn1/field_options.v
@@ -115,11 +115,6 @@ pub fn FieldOptions.from_attrs(attrs []string) !FieldOptions {
 
 	// The item has space-trimmed
 	for item in filtered {
-		// no need to check this again, its already filtered
-		// if !is_tag_marker(item) && !is_optional_marker(item) && !is_default_marker(item)
-		//		&& !is_mode_marker(item) && !is_inner_tag_marker(item) {
-		//		return error('unsupported keyword')
-		// }
 		if is_tag_marker(item) {
 			cls, num := parse_tag_marker(item)!
 			tag_ctr += 1
--- a/vlib/x/encoding/asn1/other_element.v
+++ b/vlib/x/encoding/asn1/other_element.v
@@ -104,9 +104,9 @@ pub fn (mut r RawElement) set_inner_tag(inner_tag Tag, force bool) ! {
 	if r.tag.class == .universal {
 		return error('No need it on universal class')
 	}
-
+	mode := r.mode or { return error('You dont set any mode') }
 	// when its in explicit mode, compares the provided tag with tag from the inner element.
-	if r.mode? == .explicit {
+	if mode == .explicit {
 		if !r.tag.constructed {
 			return error('unmeet_requirement, explicit should be constructed')
 		}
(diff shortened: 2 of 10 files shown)


vlib/x/encoding/asn1/field_options.v
```diff
@@ -115,10 +115,11 @@ pub fn FieldOptions.from_attrs(attrs []string) !FieldOptions {
...
 	for item in filtered {
-		if !is_tag_marker(item) && !is_optional_marker(item) && !is_default_marker(item)
-			&& !is_mode_marker(item) && !is_inner_tag_marker(item) {
-			return error('unsupported keyword')
-		}
+		// no need to check this again, its already filtered
+		// if !is_tag_marker(item) && !is_optional_marker(item) && !is_default_marker(item)
+		//		&& !is_mode_marker(item) && !is_inner_tag_marker(item) {
+		//		return error('unsupported keyword')
+		// }
```
spytheman, line 122: In general, it is better to remove commented code blocks like this one - if it is needed, it can be viewed in the git history, while keeping it present in the current working copy, makes tooling less useful - if you search for `is_tag_marker` for example, tools like `grep` or `rg` will find false positives in the comments.

vlib/x/encoding/asn1/time.v
```diff
@@ -226,7 +227,7 @@ pub fn GeneralizedTime.new(s string) !GeneralizedTime {
 	}
 }
 
-// from_time creates GeneralizedTime element from tine.Time (as an UTC time).
+// from_time creates GeneralizedTime element from standard `time.Time` (as an UTC time).
```
spytheman, line 230: Oops, good find ... I missed that typo in the previous review too.

vlib/x/encoding/asn1/field_options.v
blackshirt: okey, should be be removed in next commit

vlib/x/encoding/asn1/time.v
```diff
@@ -50,7 +50,8 @@ pub fn UtcTime.new(s string) !UtcTime {
 	}
 }
 
-fn UtcTime.from_time(t time.Time) !UtcTime {
+// from_time creates a new UtcTime element from standard `time.Time` in UTC time format.
+pub fn UtcTime.from_time(t time.Time) !UtcTime {
```
spytheman, line 54: Is there a test for this new public function?

vlib/x/encoding/asn1/time.v
blackshirt: Nope, its my foolish typo

vlib/x/encoding/asn1/time.v
blackshirt: Not yet, i will add the test for it

vlib/x/encoding/asn1/other_element.v
```diff
@@ -230,8 +182,8 @@ fn (r RawElement) check_inner_tag() ! {
 	if r.tag.class == .universal {
 		return error('Universal class does not have inner tag')
 	}
-	mode := r.mode or { return error('You dont set any mode') }
-	if mode != .explicit {
+	// This option unwrap handling already has been fixed in latest v.
+	if r.mode? != .explicit {
```
spytheman, line 186: That seems sketchy ... the current function returns `!`, not `?Type`, but there is 
`if r.mode? != .explicit {` .

What would the `?` do, in this context @felipensp ?

spytheman: It can not propagate the `none`, so will it:
a) panic
b) propagate some kind of an autogenerated error
c) something else?

spytheman (approved):
Excellent work.
Less code, more performance 🥳 !

vlib/x/encoding/asn1/other_element.v
blackshirt: I think it related with the fix [22895](https://github.com/vlang/v/pull/22895), its already support it. With the previous one, you should extract optional field with `mode := r.mode or { return error('You dont set any mode') }` .
Is this `r.mode? != .explicit` illegal construct ? or i'm missing something ...

blackshirt: Done, revert it

Issue #22960, opened 2024-11-24
C error on casting optional struct field


gechandesu:
**V doctor:**

```
V full version: V 0.4.8 5ec9bb5.1345eac
OS: linux, Linux version 6.6.8-arch1-1 (linux@archlinux) (gcc (GCC) 13.2.1 20230801, GNU ld (GNU Binutils) 2.41.0) #1 SMP PREEMPT_DYNAMIC Thu, 21 Dec 2023 19:01:01 +0000
Processor: 16 cpus, 64bit, little endian, 12th Gen Intel(R) Core(TM) i5-1240P

getwd: /home/ge/Code/v/netaddr/issues
vexe: /home/ge/.vlang/v
vexe mtime: 2024-11-24 11:11:18

vroot: OK, value: /home/ge/.vlang
VMODULES: OK, value: /home/ge/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.46.1
Git vroot status: weekly.2024.47-25-g1345eac5
.git/config present: true

CC version: cc (GCC) 14.2.1 20240910
emcc version: N/A
thirdparty/tcc status: thirdparty-linux-amd64 0134e9b9
```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg casting_optional_field_struct_method.vv`

```v
module main

struct Foo {
	foo string
	bar ?string
}

fn (f Foo) str() string {
	mut ret := f.foo
	bar_value := f.bar as string
	if bar_value == '' {
		return ret
	}
	return ret + '%' + bar_value
}

fn main() {
	ff := Foo{
		foo: 'Fooooo'
	}
	println(ff)
}
```

**What did you expect to see?**

'Fooooo' printed

**What did you see instead?**

```
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/casting_optional_field_struct_method.01JDEXY0DVN2RRQBB8CKSSYVTN.tmp.c:13719: error: '{' expected (got ";")
=================================================================
(You can pass `-cg`, or `-show-c-output` as well, to print all the C error messages).
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
[...]

PR #22973, opened 2024-11-25
cgen: allow unwrapping of `foo.bar as string`, where `foo.bar` is `?string`

Delta456:
Fixes https://github.com/vlang/v/issues/22960

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzQ0Y2Q2N2QyZWY0Y2JjYzQ2OGU4MDAiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.GbQXzAsCVTazORa8FJoNYwpODbXu93Yazq3aK8sOvZ0">Huly&reg;: <b>V_0.6-21414</b></a></sub>


vlib/v/gen/c/cgen.v
```diff
@@ -7795,10 +7795,18 @@ fn (mut g Gen) as_cast(node ast.AsCast) {
 	} else {
 		mut is_optional_ident_var := false
 		if node.expr is ast.Ident {
-			if node.expr.info is ast.IdentVar && node.expr.info.is_option {
+			if node.expr.info is ast.IdentVar && node.expr.info.is_option
+				&& !unwrapped_node_typ.has_flag(.option) {
```
Delta456, line 7799: Added this check so that  ` var.a as ?time.Time` doesn't fail

vlib/v/gen/c/cgen.v
```diff
@@ -7795,10 +7795,18 @@ fn (mut g Gen) as_cast(node ast.AsCast) {
...
 				g.unwrap_option_type(unwrapped_node_typ, node.expr.name, node.expr.is_auto_heap())
 				is_optional_ident_var = true
 			}
+		} else if node.expr is ast.SelectorExpr {
+			if node.expr.expr is ast.Ident && node.expr.typ.has_flag(.option)
+				&& !unwrapped_node_typ.has_flag(.option) {
```
Delta456, line 7805: Here for the same reason

vlib/v/tests/options/option_unwrap_selector_expr_as_cast_test.v
```diff
@@ -0,0 +1,20 @@
+struct Foo {
+	foo string
+	bar ?string
+}
+
+fn (f Foo) str() string {
+	mut ret := f.foo
+	bar_value := f.bar as string
```
spytheman, line 8: What happens if it is not a string?
A panic?

spytheman: Oh, it becomes `''` 🤔 ... I am not sure that is consistent with what other `x as Type` casts do.

Delta456: if used other type like `int`

```
file3.v:10:21: error: cannot cast non-sum type `string` using `as`
    8 | fn (f Foo) str() string {
    9 |     mut ret := f.foo
   10 |     bar_value := f.bar as int
      |                        ~~
```

PR #22987, opened 2024-11-27
v: fix missing gettid on glibc < 2.30

felipensp:
This can be reproduced by running `v vlib/v/tests/trace_calls_test.v`.


spytheman:
It needs an older Linux distro, like Debian 10 .
glibc 2.30 is from 2019.

spytheman (approved):
Excellent work.

PR #23001, opened 2024-11-28
strings.textscanner, examples: add TextScanner .skip_whitespace/0, .peek_u8/0, .peek_n_u8/0, add examples/mini_calculator_recursive_descent.v

spytheman:


spytheman:
Note: unlike `examples/mini_calculator.v`, `examples/mini_calculator_recursive_descent.v` can handle/skip whitespace, instead of erroring.

It is also shorter (since it reuses strings.textscanner for the trivial parts), and clearer to read/maintain (because of it using the recursive descent technique, instead of an explicit stack/array).

PR #23018, opened 2024-11-30
x.json2.decoder2: minor fixes (typo)

enghitalo:


JalonSolov:
Dang... causing `v vet` to crash...

spytheman:
Yeah, that is interesting 🤔 . Well, it is an opportunity to make vet more robust 🤷🏻‍♂️ .

Issue #23028, opened 2024-12-01
add array.count


Le0Developer:
### Describe the feature

count() function on array, similar to string.count

### Use Case

counting how often an element is in an array

https://github.com/vlang/adventofcode/blob/b5eabb1722543b45f632e25cba151a89204ebfc1/2024/01/leodev.v#L33

### Proposed Solution

array.count()

### Other Information

workaround is `arr.filter(it == what).len`, but it allocates a new array.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

N/A

### Environment details (OS name and version, etc.)

N/A

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzRjMzIwOGQyZWY0Y2JjYzQ2Y2ZkNWMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.9Mlneyz4mtsQodF31GaD8Uf2GicHXaVsDX_GCjFlsAk">Huly&reg;: <b>V_0.6-21469</b></a></sub>

felipensp:
I've submitted a PR to optimize `.len` operation on array.filter() 👍🏻

Le0Developer:
While a nice optimization, I don't think it's intuitive and assumes people know what the cgen is doing.

spytheman:
imho it should be a method, just like s.count()

spytheman:
it is a common enough and general operation, to be in `builtin`

medvednikov:
I'm fine with a new method similar to any(), filter() etc

PR #23030, opened 2024-12-01
v: optimize arr.filter(...).len operation

felipensp:
This PR removes the array allocation for result on `array.filter()` when its result is used to check only its `.len`.

Related to #23028

```v
a := []int{len: 20000000, init:index}
assert a.filter(it%2).len > 0
```

816ms (master)
422ms (pr)

Before:

```c
Array_int _t2 = {0};
        Array_int _t2_orig = a;
        int _t2_len = _t2_orig.len;
        _t2 = __new_array_noscan(0, _t2_len, sizeof(int));

        for (int _t3 = 0; _t3 < _t2_len; ++_t3) {
                int it = ((int*) _t2_orig.data)[_t3];
                if ((int)(it % 2)) {
                        array_push_noscan((array*)&_t2, &it);
                }
        }
        println(int_str(_t2.len));
```

After:

```c
int _t2 = 0;
        Array_int _t2_orig = a;
        int _t2_len = _t2_orig.len;
        for (int _t3 = 0; _t3 < _t2_len; ++_t3) {
                int it = ((int*) _t2_orig.data)[_t3];
                if ((int)(it % 2)) {
                        ++_t2;
                }
        }
```

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzRjNDc4NGQyZWY0Y2JjYzQ2Y2ZlYTkiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.idehsCidtK8Bs4-hy9iwoXXLfQMiD1A_H6Ii9gXyFGs">Huly&reg;: <b>V_0.6-21471</b></a></sub>

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -832,6 +832,7 @@ pub mut:
 	from_embed_types       []Type // holds the type of the embed that the method is called from
 	comments               []Comment
 	is_return_used         bool // return value is used for another expr
+	is_return_discarded    bool // return value will not be stored (e.g. a.filter(...).len)
 	//
 	is_expand_simple_interpolation bool // true, when the function/method is marked as @[expand_simple_interpolation]
 	// Calls to it with an interpolation argument like `b.f('x ${y}')`, will be converted to `b.f('x ')` followed by `b.f(y)`.
--- a/vlib/v/parser/parser.v
+++ b/vlib/v/parser/parser.v
@@ -3432,6 +3432,7 @@ fn (mut p Parser) dot_expr(left ast.Expr) ast.Expr {
 	mut left_node := unsafe { left }
 	if mut left_node is ast.CallExpr {
 		left_node.is_return_used = true
+		left_node.is_return_discarded = true
 	}
 	return sel_expr
 }
--- a/vlib/v/gen/c/array.v
+++ b/vlib/v/gen/c/array.v
@@ -880,11 +880,11 @@ fn (mut g Gen) gen_fixed_array_reverse_in_place(node ast.CallExpr) {
 
 // `nums.filter(it % 2 == 0)`
 fn (mut g Gen) gen_array_filter(node ast.CallExpr) {
+	is_return_discarded := node.is_return_discarded
 	past := g.past_tmp_var_new()
 	defer {
 		g.past_tmp_var_done(past)
 	}
-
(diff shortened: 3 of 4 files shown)


vlib/v/ast/ast.v
```diff
@@ -832,6 +832,7 @@ pub mut:
 	from_embed_types       []Type // holds the type of the embed that the method is called from
 	comments               []Comment
 	is_return_used         bool // return value is used for another expr
+	is_return_discarded    bool // return value will not be stored (e.g. a.filter(...).len)
```
spytheman, line 835: that general name is misleading - it only applies to a.filter(), when you also use .len

vlib/v/parser/parser.v
```diff
@@ -3432,6 +3432,7 @@ fn (mut p Parser) dot_expr(left ast.Expr) ast.Expr {
 	mut left_node := unsafe { left }
 	if mut left_node is ast.CallExpr {
 		left_node.is_return_used = true
+		left_node.is_return_discarded = true
```
spytheman, line 3435: the parser can not know that

spytheman:
Imho that approach is very partial to that specific situation in the issue, and can lead to a lot of subtle bugs later, for other expressions involving .filter(), but without .len and for other array methods like .map(), .any() etc.

spytheman:
The more general issue, is not that .filter() is slow because it does allocations - it is that we lack a .count() method .

vlib/v/ast/ast.v
felipensp: We can find another places to apply such knowledge checked on the parser.

vlib/v/parser/parser.v
felipensp: In a few cases we could Ensure It is discarded.

felipensp:
> The more general issue, is not that .filter() is slow because it does allocations - it is that we lack a .count() method .

I know. But this case could be optimized.

PR #23058, opened 2024-12-03
crypto.sha3: add support for Keccak-256 and Keccak-512

### LINKED ISSUE #23057: crypto.sha3: Support Keccak-256 and Keccak-512
### Describe the feature

Keccak-256 and Keccak-512 is basically same as SHA3-256 and SHA3-512, but uses `0x01` instead of `0x06` for padding.

### Use Case

I need it to interact with some software that uses Keccak-256 for hashing.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### Version used

V 0.4.8 d3b5dbd

V full version: V 0.4.8 c4aaa2e.d3b5dbd
OS: macos, macOS, 15.1.1, 24B91
Processor: 8 cpus, 64bit, little endian, Apple M3

getwd: /Users/syobon/Git/asynchro
vexe: /Users/syobon/.v/v
vexe mtime: 2024-12-03 10:07:44

vroot: OK, value: /Users/syobon/.v
VMODULES: OK, value: /Users/syobon/.vmodules
VTMP: OK, value: /tmp/v_501

Git vroot status: weekly.2024.39-563-gd3b5dbd5-dirty (1 commit(s) behind V master)
.git/config present: true

emcc version: N/A
> [!NOTE]
>

syobocat:
Closes #23057

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzRlZmQwOWQyZWY0Y2JjYzQ2ZGM4YWMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.6ChDw2g0eLr7RGCI2vBvpHbGAlv1iDQ-VLz0BTfp0zU">Huly&reg;: <b>V_0.6-21498</b></a></sub>


spytheman:
@blackshirt, @kimshrier can you please review this PR?

kimshrier:
It would be good to add keccak-256 and keccak-512 tests to test_200_length_hash() so that there is more than one validation of the algorithm.  Also, since the existing tests are against an empty string, that is potentially a degenerate case. Testing against a non-empty string will give us better confidence in the correctness of the implementation.

blackshirt:
Is it good to this `sha3` and the kecak version to adhere `hash.Hash` interface?

vlib/crypto/sha3/sha3.v
```diff
@@ -166,7 +181,7 @@ pub fn new_digest(absorption_rate int, hash_size int) !&Digest {
 
 	d := Digest{
 		rate:       absorption_rate
-		suffix:     hash_pad
```
blackshirt, line 169: Other cases of byte of `suffix` input parameters was not handled. What is happens when we pass `u8(0xff)` or other non-predefined byte values as an suffix's input, ? the behaviours was not defined, and maybe subtle to bogus thing

vlib/crypto/sha3/sha3_test.v
```diff
@@ -79,6 +79,16 @@ const test_200_message_sha3_512 = [u8(0xE7), 0x6D, 0xFA, 0xD2, 0x20, 0x84, 0xA8,
 	0x84, 0x73, 0x9A, 0x2D, 0xF4, 0x6B, 0xE5, 0x89, 0xC5, 0x1C, 0xA1, 0xA4, 0xA8, 0x41, 0x6D, 0xF6,
 	0x54, 0x5A, 0x1C, 0xE8, 0xBA, 0x00]
 
+const test_200_message_keccak_256 = [u8(0x3A), 0x57, 0x66, 0x6B, 0x04, 0x87, 0x77, 0xF2, 0xC9,
+	0x53, 0xDC, 0x44, 0x56, 0xF4, 0x5A, 0x25, 0x88, 0xE1, 0xCB, 0x6F, 0x2D, 0xA7, 0x60, 0x12, 0x2D,
+	0x53, 0x0A, 0xC2, 0xCE, 0x60, 0x7D, 0x4A]
+
```
blackshirt, line 85: I think its good to add a comment where its test data come from, from standard or other verified-sources, so we can check it

blackshirt (commented):
There are many duplicates here, likes `new_keccak_digest` and other specialized module level fns, even its just a thin wrapper for other function. Its seem (for me) there are many ways to do a same thing here, ..
I think @spytheman and others has more thinking in this area's 
I'm personally like optional parameters with default to `sha3` part.  For example from standard `math.big.integer_from_bytes` accepts optional `IntegerConfig` as the last parameters. when its not supplied, its assumed signum to be positive 1.

vlib/crypto/sha3/sha3_test.v
syobocat: I took the value from the output of Go's sha3 library: https://go.dev/play/p/r155QSwZeFq

(Their test code says `// TestKeccak does a basic test of the non-standardized Keccak hash functions.`, so I believe there is no standardized test for Keccak)

blackshirt (approved):

vlib/crypto/sha3/sha3.v
```diff
@@ -108,13 +112,26 @@ fn (err XOFSizeError) msg() string {
...
+}
+
+@[params]
+pub struct PaddingConfig {
+pub:
+	padding Padding = .sha3
+}
+
 // new_digest creates an initialized digest structure based on
 // the hash size and whether or not you specify a MAC key.
```
blackshirt, line 136: Maybe you should add the function docs to elaborate the current behaviour of the `new_digest`, so the user know the created `Digest` was expected digest.
@spytheman what do you think ?

blackshirt, line 136: The term `whether or not you specify a MAC key.` is bits missleading for me, what is `MAC key` refers to ? i dont think new_digest accepts some of  `mac key`  bytes or string

syobocat: That is an existing part of the code, so I don't know either.
Do you think what kind of comments should be there? Is it enough to just delete `and whether or not you specify a MAC key`?

spytheman: I agree.

vlib/crypto/sha3/sha3.v
```diff
@@ -163,27 +196,9 @@ pub fn new_digest(absorption_rate int, hash_size int) !&Digest {
...
-// the absorption rate and how many bytes of output you need
-//
-// absorption_rate is the number of bytes to be absorbed into the
-//     sponge per permutation.  Legal values are xof_rate_128 and
-//     xof_rate_256.
-//
-// hash_size - the number if bytes in the generated hash.
-//     Legal values are positive integers.
-pub fn new_xof_digest(absorption_rate int, hash_size int) !&Digest {
+fn validate_xof(absorption_rate int, hash_size int) ! {
```
spytheman, line 201: This is a breaking change too - `new_xof_digest` was public, and  is now both changed in name, and private.

spytheman: Imho keep the new private function, but also add back the public one as it was + its comments, just calling the private one.

vlib/crypto/sha3/sha3.v
spytheman: @kimshrier what do you think?

vlib/crypto/sha3/sha3_test.v
spytheman: Why does the Go API calls the constructors `NewLegacyKeccak256` and `NewLegacyKeccak512` ?

vlib/crypto/sha3/sha3.v
blackshirt: I think @kimshrier has depth of knowledge what thats mean ...
This single of `sha3` module has already contains support for multiples of variant, like the support for standard `sha3`, `shake128`, `shake256`, extended vesion in `xof` variant and now 'keccak` variant. I think its should be documented in the `README` or internal fn docs, its make clear for user

kimshrier: Sorry to be AWOL for a bit.  I'm really hammered at my day job at the moment.

I forget some of the details of the sha3 code and will need a little time to get my head back in to it.  It will probably be a few days before I can contribute anything.

spytheman: That is understandable, please excuse me for adding to your cognitive load.

I think we can delete that comment for now, as @blackshirt suggested, and revise it later if needed.

vlib/crypto/sha3/sha3_test.v
spytheman: Although that is just a curiosity on my part, since the naming seems a bit weird.
It does not stop merging this PR in the V repo.

syobocat: Keccak has basically been superseded by SHA3, so I think that is the reason for the word `Legacy`.
However, some protocols (like Etherium) are still using Keccak, and that's why I opened this PR.

blackshirt:
Can you please also adds some normal tests, that do import `crypto.sha3` and only use the public API of the module?
Its usually good to know its not breaks anything and working normally

blackshirt (commented):
I think overall changes is mostly acceptable now, with some notes on unhandled issues (maybe can be addressed on next iteration of pr), ie, 
- is this `sha3.Digest` should adhere to `hash.Hash` interface? I think its good if yes, because the standard stock of `sha1`,  `sha256` and `sha512` has already did it.
- Its good to have `README` file for this `sha3` module. Its not yet availables at this time of writing, and for more crucial reasons, this module has support for several variant of hashing instead of standard `sha3` thats represented by this module name.

Good jobs guys...
Waiting for other review

JalonSolov:
Yay - more `int <-> u32` issues.  @spytheman

spytheman:
> Yay - more `int <-> u32` issues. @spytheman

it is fixed in master - the CI runs just happened to interleave

spytheman (approved):
Excellent work @syobocat 🙇🏻‍♂️ .

blackshirt:
> I think that it is good to be merged - the remaining things are out of scope of this particular PR, and require changes to existing code. They are better to bepr  done separately imho.

Its good, the other concerns can be done on next pr

PR #23078, opened 2024-12-06
json2.decoder2: prepare decoder in json2 to be replaced by json2.decode2

enghitalo:
<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzUyZWU2ZWQyZWY0Y2JjYzQ2ZjU4ZDMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.A6wAb6wokYGzqzTB9CxaAlp8w3kCMxzt6M0zzwTn_HA">Huly&reg;: <b>V_0.6-21517</b></a></sub>

### DIFF
--- a/vlib/x/json2/decoder2/decode.v
+++ b/vlib/x/json2/decoder2/decode.v
@@ -777,10 +777,25 @@ fn (mut decoder Decoder) decode_value[T](mut val T) ! {
 										}
 
 										if current_field_info.value.is_raw {
-											$if field.typ is $enum {
+											$if field.unaliased_typ is $enum {
 												// workaround to avoid the error: enums can only be assigned `int` values
 												return error('`raw` attribute cannot be used with enum fields')
-											} $else $if field.typ is string || field.typ is ?string {
+											} $else $if field.typ is ?string {
+												position := decoder.current_node.value.position
+												end := position + decoder.current_node.value.length
+
+												val.$(field.name) = decoder.json[position..end]
+												decoder.current_node = decoder.current_node.next
+
+												for {
+													if decoder.current_node == unsafe { nil }
+														|| decoder.current_node.value.position + decoder.current_node.value.length >= end {
+														break
+													}
+
+													decoder.current_node = decoder.current_node.next
+												}
+											} $else $if field.typ is string {
 												position := decoder.current_node.value.position
 												end := position + decoder.current_node.value.length
 
--- a/vlib/x/json2/tests/decoder_test.v
+++ b/vlib/x/json2/tests/decoder_test.v
@@ -1,48 +1,48 @@
 import x.json2 as json
 
 fn test_raw_decode_string() {
-	str := json.decode[json.Any]('"Hello!"')!
+	str := json.raw_decode('"Hello!"')!
 	assert str.str() == 'Hello!'
 }
 
 fn test_raw_decode_string_escape() {
-	jstr := json.decode[json.Any]('"\u001b"')!
+	jstr := json.raw_decode('"\u001b"')!
 	str := jstr.str()
 	assert str.len == 1
 	assert str[0] == 27
 }
 
 fn test_raw_decode_number() {
-	num := json.decode[json.Any]('123')!
+	num := json.raw_decode('123')!
 	assert num.int() == 123
 }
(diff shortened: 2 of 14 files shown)


vlib/x/json2/decoder2/decode.v
```diff
@@ -777,10 +777,25 @@ fn (mut decoder Decoder) decode_value[T](mut val T) ! {
...
+
+												for {
+													if decoder.current_node == unsafe { nil }
+														|| decoder.current_node.value.position + decoder.current_node.value.length >= end {
+														break
+													}
+
+													decoder.current_node = decoder.current_node.next
+												}
+											} $else $if field.typ is string {
```
enghitalo, line 798: This is a workaround to avoid the bug error: enums can only be assigned `int` values

enghitalo: A cleanup will be made after this decoder replace actual decoder

vlib/x/json2/tests/decoder_test.v
```diff
@@ -1,61 +1,61 @@
 import x.json2 as json
 
 fn test_raw_decode_string() {
-	str := json.raw_decode('"Hello!"')!
+	str := json.decode[json.Any]('"Hello!"')!
 	assert str.str() == 'Hello!'
 }
```
spytheman, line 6: No. Keep the existing test, and add your own if you want.
Replacing things like this is bad - it removes coverage for the existing functionality of `json.raw_decode`.
It is also very sloppy - the test name is still `fn test_raw_decode_string() {` ...

Just keeps things working as much as possible,
and add *new tests* please.

vlib/x/json2/tests/encoder_test.v
```diff
@@ -19,7 +19,7 @@ fn test_json_string_characters() {
 
 	assert json.encode(r'\n\r\b\f\t\\\"\/') == r'"\\n\\r\\b\\f\\t\\\\\\\"\\\/"'
 
-	text := json.raw_decode(r'"\n\r\b\f\t\\\"\/"') or { '' }
+	text := json.decode[json.Any](r'"\n\r\b\f\t\\\"\/"') or { '' }
```
spytheman, line 22: same ... keep the code that tests `json.raw_decode` and *add* new tests.

vlib/x/json2/tests/json2_test.v
```diff
@@ -16,12 +16,12 @@ pub mut:
 
 fn test_fast_raw_decode() {
 	s := '{"name":"Peter","age":28,"salary":95000.5,"title":2}'
-	o := json.fast_raw_decode(s) or {
+	o := json.decode[json.Any](s) or {
```
spytheman, line 19: Same - keep the current code, and add *new* tests/asserts.

spytheman: If it acts the same, even better - decode 2 times, once the old way, and once the new way, and compare the results.

spytheman:
No, deprecations of public functions takes time. Add at least 1 month starting from today.
See https://docs.vlang.io/attributes.html, search for the section `Function/method deprecations` .

Alternatively, just do not deprecate the method, and leave it as it is, including its tests.

enghitalo:
> No, deprecations of public functions take time. Add at least 1 month starting from today.
> See https://docs.vlang.io/attributes.html, search for the section `Function/method deprecations` .
> 
> Alternatively, just do not deprecate the method, and leave it as it is, including its tests.

I prefer to make them deprecated. When x.json2 becomes just json, probably none of these methods will be part of the library. I hope

spytheman:
They are public API, public APIs should have a deprecation period. It is very rude to users of your libraries, to make breaking changes like this, especially if you have an easy way to avoid it.

PR #23106, opened 2024-12-09
cgen: fix printing fn call of returning c struct value (fix #23104)

### LINKED ISSUE #23104: C error when printing C struct in V.
### Describe the bug

A wrapper function for a C function returns a struct which can only be printed when the value of the wrapper function is first set to a variable. A C error is produced when you try to directly print the function output.

### Reproduction Steps

Using this module (https://github.com/islonely/mouse) with this main function.

```v
import islonely.mouse

fn main() {
	// this prints
	size := mouse.screen_size()
	println(size)

	// this makes C error
	println(mouse.screen_size())
}
```

### Expected Behavior

Expected the screen size to be printed twice.

### Current Behavior

Variable `size` is printed. Printing the output of the same function used to set the `size` variable produces a C error. `v -cg` output:

```
/tmp/v_501/test.01JEM7FE4RPN0KSJS08MKHEXDT.tmp.c:5938:36: error: cannot take the address of an rvalue of type 'islonely__mouse__Size' (aka 'struct Size')
 5938 |         println(islonely__mouse__Size_str(&islonely__mouse__screen_size()));
      |                                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
builder error: 
==================
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.8 2911f29.d257a1f

```
V full version: V 0.4.8 2911f29.d257a1f
OS: macos, macOS, 15.1, 24B83
Processor: 8 cpus, 64bit, little endian, Apple M3

getwd: /Users/adamoates
vexe: /Users/adamoates/v/v
vexe mtime: 2024-12-08 22:42:00

vroot: OK, value: /Users/adamoates/v
VMODULES: OK, value: /Users/adamoates/.vmodules
VTMP: OK, value: /tmp/v_501

Git vroot status: weekly.2024.49-42-gd257a1f0
.git/config present: true

emcc version: N/A
```

> [!NOTE]
>

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzU2MjNiMzFlMDYzMmRkMDhlMjYwZDQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.n-UHDJD7Cp_NXnbwXt9wNA_RQ8ZYJLM-OB7O05g--qo">Huly&reg;: <b>V_0.6-21541</b></a></sub>

yuyi98:
This PR fix printing fn call of returning c struct value (fix #23104).

- Fix printing fn call of returning c struct value.
- Add test.

csize.h

```v
struct Size {
   int width;
	int height;
};
```

```v
#include "@VMODROOT/csize.h"

struct C.Size {
	width  int
	height int
}

type Size = C.Size

fn get_size() C.Size {
	return C.Size{11, 22}
}

fn get_csize() Size {
	return Size(C.Size{11, 22})
}

fn main() {
	println(get_size())
	assert true
	println(get_csize())
	assert true
}
```

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzU2NzdiYTFlMDYzMmRkMDhlMjZkNTQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.c-Ghf8Iom_rocSryn0Wk9_zP2eDXee947yq1BqflXS4">Huly&reg;: <b>V_0.6-21543</b></a></sub>

### DIFF
--- a/vlib/v/tests/c_structs/return_csize_test.v
+++ b/vlib/v/tests/c_structs/return_csize_test.v
@@ -1,0 +1,15 @@
+#include "@VMODROOT/csize.h"
+
+struct C.Size {
+	width  int
+	height int
+}
+
+fn get_size() C.Size {
+	return C.Size{11, 22}
+}
+
+fn test_return_csize() {
+	println(get_size())
+	assert true
+}
--- a/vlib/v/tests/c_structs/return_csize_test.v
+++ b/vlib/v/tests/c_structs/return_csize_test.v
@@ -5,11 +5,19 @@ struct C.Size {
 	height int
 }
 
+type Size = C.Size
+
 fn get_size() C.Size {
 	return C.Size{11, 22}
 }
 
+fn get_csize() Size {
+	return Size(C.Size{11, 22})
+}
+
 fn test_return_csize() {
 	println(get_size())
+	assert true
+	println(get_csize())
 	assert true
 }
--- a/vlib/v/gen/c/str.v
+++ b/vlib/v/gen/c/str.v
@@ -112,14 +112,14 @@ fn (mut g Gen) gen_expr_to_string(expr ast.Expr, etype ast.Type) {
 			g.write('")')
 		}
 	} else if sym_has_str_method
-		|| sym.kind in [.array, .array_fixed, .map, .struct, .multi_return, .sum_type, .interface] {
+		|| sym.kind in [.alias, .array, .array_fixed, .map, .struct, .multi_return, .sum_type, .interface] {
 		unwrap_option := expr is ast.Ident && expr.or_expr.kind == .propagate_option
 		exp_typ := if unwrap_option { typ.clear_flag(.option) } else { typ }
 		is_dump_expr := expr is ast.DumpExpr
 		is_var_mut := expr.is_auto_deref_var()
 		str_fn_name := g.get_str_fn(exp_typ)
 		temp_var_needed := expr is ast.CallExpr
-			&& (expr.return_type.is_ptr() || g.table.final_sym(expr.return_type).is_c_struct())
+			&& (expr.return_type.is_ptr() || g.table.sym(expr.return_type).is_c_struct())
 		mut tmp_var := ''
 		if temp_var_needed {
 			tmp_var = g.new_tmp_var()
--- a/vlib/v/gen/c/str.v
+++ b/vlib/v/gen/c/str.v
@@ -118,7 +118,8 @@ fn (mut g Gen) gen_expr_to_string(expr ast.Expr, etype ast.Type) {
 		is_dump_expr := expr is ast.DumpExpr
 		is_var_mut := expr.is_auto_deref_var()
 		str_fn_name := g.get_str_fn(exp_typ)
-		temp_var_needed := expr is ast.CallExpr && expr.return_type.is_ptr()
+		temp_var_needed := expr is ast.CallExpr
+			&& (expr.return_type.is_ptr() || g.table.final_sym(expr.return_type).is_c_struct())
 		mut tmp_var := ''
 		if temp_var_needed {
 			tmp_var = g.new_tmp_var()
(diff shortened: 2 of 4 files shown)


spytheman (approved):
Excellent work.

vlib/v/tests/c_structs/return_csize_test.v
spytheman, line 1: This example however is not equivalent to the issue code, and does not check the same issue (on linux with clang):
image

spytheman: Note, the issue code, compiles with tcc on both master and in this PR.

The code from `return_csize_test.v` compiles with both tcc, gcc and several versions of clang on both master and on this PR.

spytheman: image

spytheman: The issue code however, compiles with tcc on both master and the PR:
image

spytheman: The issue code needed this to be reproduced:
`git clone https://github.com/islonely/mouse ~/.vmodules/islonely/mouse`
then
```v
import islonely.mouse

fn main() {
	// this prints
	size := mouse.screen_size()
	println(size)

	// this makes C error
	println(mouse.screen_size())
}
```

then compiling with clang (or gcc), but not tcc (tcc compiles it fine).

spytheman: I think that a temporary variable can be used for such structs, so that later taking the address of it will work.

spytheman: The issue also has this function on the C side, not just the struct, perhaps this is the main difference:
```c
struct Size screen_size() {
    Display *dpy = XOpenDisplay(0);
    int screen = DefaultScreen(dpy);
    struct Size sz = {
        DisplayWidth(dpy, screen),
        DisplayHeight(dpy, screen)
    };
    XCloseDisplay(dpy);
    return sz;
}
```

yuyi98: ok, I see, I'll try to fix it.

spytheman: Hm, no ... the generated C code for the V function is more or less the same in structure 🤔 .

Perhaps it is different because the C wrapper is in a different module in the issue ?

spytheman: The repo also does:
```v
struct C.Position {
__global:
    x int
    y int
}
pub type Size = C.Size
@[inline]
pub fn screen_size() Size {
    return C.screen_size()
}
```

It may be because of the V alias of the C struct type 🤔 .

spytheman: Yep, the root cause is the `pub type Size = C.Size` alias in the issue.

If I change `pub fn screen_size() Size {` to `pub fn screen_size() C.Size {`, in the module code, it compiles on master with the same issue code.

spytheman: @yuyi98 I think it just needs an additional condition to handle the alias case in the fix.

spytheman: Yep, it works for me too for the original issue.
Excellent work.

PR #23113, opened 2024-12-09
v: allow option array element comparison `==` and `!=`

### LINKED ISSUE #23108: allow "unwrapped Option cannot be used in an infix expression"
### Describe the feature

This currently produces a checker error:

```v
fn main() {
    x := []?int{len: 5, init: 1}

    println(x[0] == x[1])
}
```

(`error: unwrapped Option cannot be used in an infix expression`)

### Use Case

Came up in Advent of Code day 9 when dealing with a `[]?int{}`.

### Proposed Solution

Fix the underlying cgen issue and then remove the checker error.

### Other Information

_No response_

### Acknowledgements

### Version used

V 0.4.8 602b097.e32e9f7

V full version: V 0.4.8 602b097.e32e9f7
OS: macos, macOS, 15.1.1, 24B2091
Processor: 10 cpus, 64bit, little endian, Apple M4

getwd: /Users/leodev
vexe: /Users/leodev/p/v/v/v
vexe mtime: 2024-12-07 15:07:56

vroot: OK, value: /Users/leodev/p/v/v
VMODULES: OK, value: /Users/leodev/.vmodules
VTMP: OK, value: /tmp/v_501

Git vroot status: weekly.2024.49-33-ge32e9f70 (10 commit(s) behind V master)
.git/config present: true

emcc version: N/A

> [!NOTE]
>

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzU2YjA0NjFlMDYzMmRkMDhlMjc1ZWIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.tczBe33yiYiNZdt4hshEpUAorkzM3eK4NGFD5dvhb1M">Huly&reg;: <b>V_0.6-21545</b></a></sub>

felipensp:
Fix #23108

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzU2ZDNkZjFlMDYzMmRkMDhlMjdiOTAiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.FMtTxVtLg9MMHG-mwIQz615q3GHUBQStJu7L2gcTbhA">Huly&reg;: <b>V_0.6-21550</b></a></sub>

### DIFF
--- a/vlib/v/tests/options/option_indexexpr_eq_test.v
+++ b/vlib/v/tests/options/option_indexexpr_eq_test.v
@@ -1,0 +1,7 @@
+fn test_main() {
+	x := []?int{len: 5, init: 1}
+	assert x[0] == x[1]
+
+	y := []?int{len: 4, init: 2}
+	assert !(x[0] == y[1])
+}
--- a/vlib/v/checker/infix.v
+++ b/vlib/v/checker/infix.v
@@ -893,9 +893,9 @@ fn (mut c Checker) infix_expr(mut node ast.InfixExpr) ast.Type {
 	right_is_option := right_type.has_flag(.option)
 	if left_is_option || right_is_option {
 		opt_infix_pos := if left_is_option { left_pos } else { right_pos }
-		if (node.left !in [ast.Ident, ast.SelectorExpr, ast.ComptimeSelector]
+		if (node.left !in [ast.Ident, ast.IndexExpr, ast.SelectorExpr, ast.ComptimeSelector]
 			|| node.op in [.eq, .ne, .lt, .gt, .le, .ge]) && right_sym.kind != .none
-			&& !c.inside_sql {
+			&& !right_is_option && !c.inside_sql {
 			c.error('unwrapped Option cannot be used in an infix expression', opt_infix_pos)
 		}
 	}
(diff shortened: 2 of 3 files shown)


vlib/v/tests/options/option_indexexpr_eq_test.v
```diff
@@ -0,0 +1,7 @@
+fn test_main() {
+	x := []?int{len: 5, init: 1}
+	assert x[0] == x[1]
+
+	y := []?int{len: 4, init: 2}
+	assert !(x[0] == y[1])
+}
```
spytheman, line 7: Please also add a test for elements of `[]?string` in the same function.
Another assertion that would be nice, is for elements of `[]?int{len: 5}`, i.e. where both are `none`.
Another assertion would be to compare 1 element that is `none`, with another which is not.

spytheman:
I've merged it without the additional tests for now, since I do not want to wait for their CI, before making a release, but I really do think that the tests should be added. 

There were times in the past already, where the implementation for the options and generics types had hysteresis, and things that normally worked in isolation, broke when used in the same function due to that.

vlib/v/tests/options/option_indexexpr_eq_test.v
felipensp: Ok. I'll add more tests for another types.

PR #23130, opened 2024-12-11
checker: fix missing check on range expr when high var is same iteration value var

felipensp:
Fix bug spotted by @Bruno-Vdr

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzU5NjNjNDFlMDYzMmRkMDhlMzE3NzMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.NJByJcoKfXeioCYv2d1cMhB3sfkg8ZTTjwsYYRedNFA">Huly&reg;: <b>V_0.6-21567</b></a></sub>

### DIFF
--- a/vlib/v/checker/for.v
+++ b/vlib/v/checker/for.v
@@ -68,10 +68,19 @@ fn (mut c Checker) for_in_stmt(mut node ast.ForInStmt) {
 		} else if high_type.has_option_or_result() {
 			c.error('the `high` value in a `for x in low..high {` loop, cannot be Result or Option',
 				node.high.pos())
+		} else if node.cond is ast.Ident && node.val_var == node.cond.name {
+			if node.is_range {
+				c.error('in a `for x in <range>` loop, the key or value iteration variable `${node.val_var}` can not be the same as the low variable',
+					node.cond.pos())
+			} else {
+				c.error('in a `for x in array` loop, the key or value iteration variable `${node.val_var}` can not be the same as the low variable',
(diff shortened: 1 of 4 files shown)


vlib/v/checker/for.v
```diff
@@ -68,6 +68,9 @@ fn (mut c Checker) for_in_stmt(mut node ast.ForInStmt) {
 		} else if high_type.has_option_or_result() {
 			c.error('the `high` value in a `for x in low..high {` loop, cannot be Result or Option',
 				node.high.pos())
+		} else if node.high is ast.Ident && node.val_var == node.high.name {
+			c.error('the `high` variable cannot be the same value iteration variable',
+				node.high.pos())
```
spytheman, line 73: The check for the low value `for v in v..10 {`, is implemented in the parser 🤔, but it is indeed better to have it in the checker, like you did for the high one ... 
Can we move the parser error from `vlib/v/parser/for.v:150`, here too?

spytheman (approved):

PR #23154, opened 2024-12-13
parser: allow `type` as plain attr value

### LINKED ISSUE #23150: Keyword 'type' cannot be used in json attribute
### Describe the bug

```
pub struct Token {
	type     Type     @[json: type]
}
```

This throws an error as `type` is a keyword.
I would expect the attribute to udnerstand we are not entering vlang keywords in this context

### Reproduction Steps

Create a struct like so:

```
pub struct Token {
	type     Type     @[json: type]
}
```

### Expected Behavior

Attribute json should allow any entry as there is no limitation in the JSON naming scheme (type, if, else, ect are all valid JSON entries)

### Current Behavior

Vlang keywords are not allowd in the JSON attribute.
`unexpected keyword 'type', an argument is expected after ':'compiler`

### Possible Solution

_No response_

_No response_

### V version

0.4.8 610558d

macOS - intel

> [!NOTE]
>

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzVjMzJhNjFlMDYzMmRkMDhlMzdjZDEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.Fe7c6kE85v2LJ8i5gkcGMYendix8_uDW2185YeOkAZg">Huly&reg;: <b>V_0.6-21587</b></a></sub>

felipensp:
Fix #23150

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzVjYjkwYjFlMDYzMmRkMDhlMzhiNmQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.DG1fSKB_kX2K5SQqKXO6k5qggTkH1zu6jTPOzVrFHCQ">Huly&reg;: <b>V_0.6-21591</b></a></sub>

### DIFF
--- a/vlib/v/parser/parser.v
+++ b/vlib/v/parser/parser.v
@@ -2045,7 +2045,7 @@ fn (mut p Parser) parse_attr(is_at bool) ast.Attr {
 		if p.tok.kind == .colon {
 			has_arg = true
 			p.next()
-			if p.tok.kind == .name || p.tok.kind == .key_type { // `name: arg`
+			if p.tok.kind == .name || (p.tok.kind != .string && token.is_key(p.tok.lit)) { // `name: arg`
 				kind = .plain
 				arg = p.check_name()
 			} else if p.tok.kind == .number { // `name: 123`
--- a/vlib/v/parser/parser.v
+++ b/vlib/v/parser/parser.v
@@ -2045,7 +2045,7 @@ fn (mut p Parser) parse_attr(is_at bool) ast.Attr {
 		if p.tok.kind == .colon {
 			has_arg = true
 			p.next()
-			if p.tok.kind == .name { // `name: arg`
+			if p.tok.kind == .name || p.tok.kind == .key_type { // `name: arg`
 				kind = .plain
 				arg = p.check_name()
 			} else if p.tok.kind == .number { // `name: 123`
--- a/vlib/v/fmt/tests/keyword_attr_keep.vv
+++ b/vlib/v/fmt/tests/keyword_attr_keep.vv
@@ -4,4 +4,6 @@ struct Type {
 
 pub struct Token {
 	foo Type @[json: type]
+	bar Type @[fn; json: if]
+	baz Type @[for; json: while; spawn]
 }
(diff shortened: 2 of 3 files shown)


vlib/v/parser/parser.v
```diff
@@ -2045,7 +2045,7 @@ fn (mut p Parser) parse_attr(is_at bool) ast.Attr {
 		if p.tok.kind == .colon {
 			has_arg = true
 			p.next()
-			if p.tok.kind == .name { // `name: arg`
+			if p.tok.kind == .name || p.tok.kind == .key_type { // `name: arg`
```
spytheman, line 2048: by that logic, any other keyword should be allowed here too

spytheman: As the issue states:

> Attribute json should allow any entry as there is no limitation in the JSON naming scheme (type, if, else, ect are all valid JSON entries)

`type` is there used just as a concrete example.

felipensp: Yeah. Makes sense. Thanks.

felipensp: Fixed.

spytheman (approved):
Excellent work.

PR #23158, opened 2024-12-14
cgen: -parallel-cc cleanup

felipensp:
Clean up

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzVkNjMwYzFlMDYzMmRkMDhlMzk3ZmQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.2-TnlSzhE-j6EIR7fRaO_jvb8GveSveCODfVdBiwGKY">Huly&reg;: <b>V_0.6-21595</b></a></sub>

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -52,7 +52,6 @@ pub struct Gen {
 	timers_should_print bool
 mut:
 	out        strings.Builder
-	out0       strings.Builder // on -parallel-cc is the file where auto fns are placed in
 	extern_out strings.Builder // extern declarations for -parallel-cc
 	// line_nr                   int
 	cheaders                  strings.Builder
@@ -670,12 +669,13 @@ pub fn gen(files []&ast.File, mut table ast.Table, pref_ &pref.Preferences) GenO
 	header = '#ifndef V_HEADER_FILE\n#define V_HEADER_FILE' + header
 	header += '\n#endif\n'
 
+	mut helpers := strings.new_builder(50_000)
 	// Code added here (after the header) goes to out_0.c in parallel cc mode
 	// Previously it went to the header which resulted in duplicated code and more code
 	// to compile for the C compiler
 	if g.anon_fn_definitions.len > 0 {
 		if g.nr_closures > 0 {
-			g.out0.writeln2('\n// V closure helpers', c_closure_fn_helpers(g.pref))
+			helpers.writeln2('\n// V closure helpers', c_closure_fn_helpers(g.pref))
 		}
 		/*
 		b.writeln('\n// V anon functions:')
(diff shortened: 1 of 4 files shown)


vlib/v/gen/c/cgen.v
```diff
@@ -274,7 +273,18 @@ mut:
 	vweb_filter_fn_name  string // vweb__filter or x__vweb__filter, used by $vweb.html() for escaping strings in the templates, depending on which `vweb` import is used
 }
 
-pub fn gen(files []&ast.File, mut table ast.Table, pref_ &pref.Preferences) (string, string, string, string, []int) {
+@[heap]
+pub struct GenOutput {
```
spytheman, line 277: good ... it was getting unwieldy
I had something similar in https://github.com/vlang/v/commit/7a3e717061a2a13199ff67f2324add89b17d73ef#diff-1327b3543e8530b1f5a8352284a6ad6acd3565e323c845af1fb372d34faafcecR46-R53, just named `struct CgenResult {`, but the exact name does not matter.

vlib/v/gen/c/cgen.v
```diff
@@ -658,14 +668,14 @@ pub fn gen(files []&ast.File, mut table ast.Table, pref_ &pref.Preferences) (str
 	mut header := b.last_n(b.len)
 	header = '#ifndef V_HEADER_FILE\n#define V_HEADER_FILE' + header
 	header += '\n#endif\n'
-	g.out0_start = b.len
-	b.writeln('// ZULUL1')
+
+	mut helpers := strings.new_builder(50_000)
```
spytheman, line 672: why 50_000 ?

felipensp: Just a shoot, what do you suggest?

JalonSolov: I usually use powers of 2, since the computer is more "comfortable" with those.  It doesn't really matter, though.  It should be based on max line length, to avoid re-alloc of the buffer.

felipensp: Well, for cmd/v it is 175225.

felipensp: hmm. `b` is 539610 for cmd/v, we could adjust it too.

spytheman: I was just curious, because of the round number. You are right that it can be adjusted later if needed to avoid reallocations for the common cases.

spytheman (approved):
Excellent work.

Issue #21826, opened 2024-07-08
Add input_character_immediate() function to Stdlib


mike-ward:
### Describe the feature

Add new function to `Stdib` that will get a character from `stdin` if available or return -1 if not. 

### Use Case

I want to add keyboard commands to the recently submitted `vmatrix` example. Currently, there does not appear to be to console input function that does not wait for user input. 

### Proposed Solution

Add an input function similar to `wgetch()` from the `ncursors` lib.

`fn input_character_no_wait() int` or `fn input_character_immediate() int` are possible names for this function. If no character is waiting in the input buffer, the function returns -1.

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.6 f6de36d

### Environment details (OS name and version, etc.)

V full version: V 0.4.6 8215f21.f6de36d
OS: macos, macOS, 14.5, 23F79
Processor: 8 cpus, 64bit, little endian, Apple M2

getwd: /Users/mike/Documents/github/vmatrix
vexe: /Users/mike/Documents/github/v/v
vexe mtime: 2024-07-07 15:50:01

vroot: OK, value: /Users/mike/Documents/github/v
VMODULES: OK, value: /Users/mike/.vmodules
VTMP: OK, value: /tmp/v_501

Git version: git version 2.45.2
Git vroot status: weekly.2024.27-30-g740be17c (3 commit(s) behind V master)
.git/config present: true

CC version: Apple clang version 15.0.0 (clang-1500.3.9.4)
thirdparty/tcc status: thirdparty-macos-arm64 5c1d002f
Mikes-Air➜  vmatrix  ᐅ  

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #23171, opened 2024-12-15
term: add key_pressed, enable_echo 

kbkpbot:
[Feature/Enhancement Request](https://github.com/vlang/v/issues/21826)
#21826

```v
// enable_echo enable/disable echo input characters
pub fn enable_echo(enable bool) 
// key_pressed gives back a single character, read from the standard input.
// It returns -1 on error or no character in non-blocking mode
pub fn key_pressed(blocking bool, echo bool) int 
```

TODO: add tests, but I don't know how ....

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzVlNTA2ZmYxNDRmNTg1YWU0ZDUyZGQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.fTq2MF2Gyw1FmFkrXU57w9Gu4HKGFmGpnJQcEol6W90">Huly&reg;: <b>V_0.6-21607</b></a></sub>


vlib/term/term_nix.c.v
```diff
@@ -215,3 +216,60 @@ pub fn graphics_num_colors() u16 {
...
+		state.c_lflag |= C.ECHO
+	} else {
+		state.c_lflag &= ~C.ECHO
+	}
+	termios.tcsetattr(0, C.TCSANOW, mut state)
+}
+
+// key_pressed gives back a single character, read from the standard input.
+// It returns -1 on error or no character in non-blocking mode
+pub fn key_pressed(blocking bool, echo bool) int {
```
spytheman, line 234: imho there is no need for the second argument, since it essentially duplicates what `enable_echo`, but for each key check.

That `state.c_lflag &= ~C.ICANON` can be moved inside enable_echo too.

spytheman: I also think that it will be clearer and easier for maintenance, if we have 2 functions:
`pub fn key_pressed() int {`
and
`pub fn key_pressed_blocking() int {`
instead of a single one, with a bool argument to choose it - the places where you want each variant, are imho different, and having a mandatory boolean argument for each callsite does not seem very ergonomic to me.

vlib/term/term_windows.c.v
```diff
@@ -156,3 +159,34 @@ pub fn graphics_num_colors() u16 {
...
+pub fn key_pressed(blocking bool, echo bool) int {
+	mut ret := u8(0)
+	for {
+		pressed := C.kbhit()
+		if pressed {
+			if echo {
+				return C.getche()
+			} else {
+				return C.getch()
+			}
```
spytheman, line 183: now I see why you want the `echo bool` argument, because on windows that can not be extracted into enable_echo 🤔 .

spytheman:
I'm using this variation to test:
```v
module main

import term
import time

fn main() {
	unbuffer_stdout()
	term.enable_echo(false)
	defer {
		term.enable_echo(true)
	}
	for {
		// non-blocking mode, no echo
		x := term.key_pressed(false, true)
		if x == 0 {
			break
		}
		if x > 0 {
			println(x)
		}
		time.sleep(16 * time.millisecond)
		print('\r${time.now()} | ')
	}
	println('done')
}
```

(pressing Ctrl-D exits the loop, without having to reset the terminal, which is needed if I stop it with Ctrl-C)

spytheman:
It seems that the keycodes that are used for F1 + arrows are different than the ones that terminals on linux/macos use 🤔 .

spytheman:
`examples/term.ui/event_viewer.v` (and `term.ui`) already does a good job of providing an uniform API to user programs, when it comes to keyboard events/key sym codes.

PR #23172, opened 2024-12-15
json: fix memory leak on result messages

felipensp:
For test just check with:
`v -keepc -g -cc clang-17 -cflags -fsanitize=address,pointer-compare,pointer-subtract test vlib/json/tests/json_option_raw_test.v `


spytheman (approved):
Excellent work.

Issue #21467, opened 2024-05-08
Port and add Cuid2 (secure uuid generator) in V standard lib


b-bruno:
### Describe the feature

Hello,

V use uuid_v4 to generate uuid, but using V4 UUIDs is also not safe, because there are known  possibile attacks on several id generating algorithms that a sophisticated attacker can use to predict next ids.

If you can port Cuid2 (Secure, collision-resistant ids optimized for horizontal scaling and performance. Next generation UUIDs) in V and  add inside standard library, it's will be nice.

JS original sources and Go port :
JS : https://github.com/paralleldrive/cuid2
Go: https://github.com/nrednav/cuid2

### Use Case

I would like to avoid any problems performances (uuid database) and collision using uuid and use another secure uuid generator.

### Proposed Solution

Port Cuid2 in V
JS original sources and Go port :
JS : https://github.com/paralleldrive/cuid2
Go:  https://github.com/nrednav/cuid2

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

0.4.5

### Environment details (OS name and version, etc.)

Linux , Windows, Mac Os
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #23181, opened 2024-12-16
vlib: add new `rand.cuid2` module

kbkpbot:
[Feature/Enhancement Request](https://github.com/vlang/v/issues/21467)
issue #21467

Ported from https://github.com/paralleldrive/cuid2

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzYwMGM1ZWYxNDRmNTg1YWU0ZDgzN2YiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.w477bsdRAmoPXNw47sEQZ6jcGYgUlkCfmUkW2y3dF7c">Huly&reg;: <b>V_0.6-21618</b></a></sub>

### DIFF
--- a/vlib/rand/cuid2/cuid2.v
+++ b/vlib/rand/cuid2/cuid2.v
@@ -31,6 +31,29 @@ pub mut:
 	length int = default_id_length
 }
 
+@[params]
+pub struct Cuid2Param {
+pub mut:
+	// A PRNG that has a PRNG interface
+	prng &rand.PRNG = rand.get_current_rng()
+	// Length of the generated Cuid, min = 2, max = 32, default = 24
+	length int = default_id_length
+}
+
+// new Create a cuid2 UUID generator.
+pub fn new(param Cuid2Param) Cuid2Generator {
+	return Cuid2Generator{
+		prng:   param.prng
+		length: param.length
+	}
+}
+
+// generate Generate a new cuid2 UUID.
+// It is an alias to function `cuid2()`
+pub fn (mut g Cuid2Generator) generate() string {
+	return g.cuid2()
+}
+
 // cuid2 generates a random (cuid2) UUID.
 // Secure, collision-resistant ids optimized for horizontal
 // scaling and performance. Next generation UUIDs.
@@ -58,6 +81,33 @@ pub fn (mut g Cuid2Generator) cuid2() string {
 	hash_input := now + salt + count + g.fingerprint
 	hash_digest := first_letter + hash(hash_input)[1..g.length]
 	return hash_digest
+}
+
+// next Generate a new cuid2 UUID.
+// It is an alias to function `cuid2()`
+pub fn (mut g Cuid2Generator) next() ?string {
+	return g.cuid2()
+}
+
+// is_cuid Checks whether a given `cuid` has a valid form and length
+pub fn is_cuid(cuid string) bool {
+	if cuid.len < min_id_length || cuid.len > max_id_length {
+		return false
+	}
+
+	// first letter should in [a..z]
+	if cuid[0] < u8(`a`) || cuid[0] > u8(`z`) {
+		return false
+	}
+
+	// other letter should in [a..z,0..9]
+	for letter in cuid[1..] {
+		if (letter >= u8(`a`) && letter <= u8(`z`)) || (letter >= u8(`0`) && letter <= u8(`9`)) {
+			continue
+		}
+		return false
+	}
+	return true
 }
 
 fn create_entropy(length int, mut prng rand.PRNG) string {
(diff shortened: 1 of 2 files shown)


vlib/rand/cuid2/cuid2.v
```diff
@@ -14,7 +14,7 @@ __global session_counter = i64(0)
 // A unique string that will be used by the Cuid generator
 // to help prevent collisions when generating Cuids in a
 // distributed system.
-__global fingerprint = ''
+__global fingerprint = string('')
```
Delta456, line 17:
```suggestion
__global fingerprint string
```

vlib/rand/cuid2/cuid2_test.v
```diff
@@ -0,0 +1,44 @@
...
+	// musl prng
+	musl_prng := &musl.MuslRNG{}
+	uuid_musl := cuid2(prng: musl_prng, length: 32)
+	assert check_valid_cuid2(uuid_musl, 32)
+
+	// mt19937 prng
+	mt19937_prng := &mt19937.MT19937RNG{}
+	uuid_mt19937 := cuid2(prng: mt19937_prng, length: 32)
+	assert check_valid_cuid2(uuid_mt19937, 32)
+}
```
hungrybluedev, line 44: Can we add another function that tests a few successive calls to cuid2 in a row and check that they are unique?

vlib/rand/cuid2/cuid2.v
```diff
@@ -0,0 +1,106 @@
...
+import math.big
+import os
+
+// A counter that will be used to affect the entropy of
+// successive id generation calls
+__global session_counter = i64(0)
+// A unique string that will be used by the Cuid generator
+// to help prevent collisions when generating Cuids in a
+// distributed system.
+__global fingerprint string
```
spytheman, line 17: There is no need for these globals.

Create a structure (for example `pub struct CuidGenerator {`),that has fields instead of the globals, and add an API, that returns instances of that structure.

Then the functions here, that use the globals, can become methods on that structure, using/modifying the fields of the struct instead, and you can have several independent generators.

vlib/rand/cuid2/cuid2.v
```diff
@@ -0,0 +1,106 @@
...
+
+pub const default_id_length = 24
+pub const min_id_length = 2
+pub const max_id_length = 32
+// ~22k hosts before 50% chance of initial counter collision
+pub const max_session_count = 476782367
```
spytheman, line 23: Why are these constants public?
Are not they just implementation details for the module?

spytheman:
Is the result without customization, i.e. with just `mut g := cuid2.Cuid2Generator{}` good enough?

spytheman:
For me it generated:
```
[24]==>goxxcgiqufthxu1sk3y48qj5
[24]==>q9n22bn6wy7a5qklsd7ukmjr
[24]==>zehhczg6s1zjsgryqsmhh42p
[24]==>xs9pjbitorlpy6pr6yqx6dwr
[24]==>gub6hw1irmnlrnijd657cyye
[24]==>h4yyf1xm7az90e9f9hdirzn9
[24]==>v80h4xfm58rkea1oqbdpr0m8
[24]==>dp3fdkc8zdv8f6fjzqt2ye1w
[24]==>onzaftvtah6im0824re9yfoq
[24]==>jcpbcj3g7z8mru3mindih689
```

spytheman:
I've just ran a modified version of the example from https://github.com/nrednav/cuid2 (a Go implementation), and it also generates a 24 id by default 🤔 .

spytheman:
Note, Go's version also makes the validation function public (while we only have it in the test). Can it be moved in the module code too?

JalonSolov:
Good point.  If you want to use cuid2's, you will probably want to validate them at some point, so it would be good to have that as another method, instead of having to write it yourself every time.

spytheman:
Go's implementation of the validation function is just:
```go
// Checks whether a given Cuid has a valid form and length
func IsCuid(cuid string) bool {
	length := len(cuid)
	hasValidForm, _ := regexp.MatchString("^[a-z][0-9a-z]+$", cuid)
	if hasValidForm && length >= MinIdLength && length <= MaxIdLength {
		return true
	}
	return false
}
```

spytheman (approved):
Excellent work.

PR #23182, opened 2024-12-16
cgen, checker: fix option unwrapping and call from option struct field

felipensp:
Make works:

```v
struct Foo {
mut:
	func ?fn (voidptr) ?bool = unsafe { nil }
}

fn callback(foo &Foo) ?bool {
	return foo.func? == callback
}

fn test_main() {
	t := Foo{
		func: callback
	}
	assert t.func? == callback
	if t.func != none {
		assert t.func(&t)
	} else {
		assert false
	}
}
```

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzYwMWEwZGYxNDRmNTg1YWU0ZDg1NWMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.RdjTIOlzQoGbLltZtzsF3jZC49AGek8GkTHNEkpkExI">Huly&reg;: <b>V_0.6-21619</b></a></sub>

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -834,6 +834,7 @@ pub mut:
 	is_return_used         bool // return value is used for another expr
 	//
 	is_expand_simple_interpolation bool // true, when the function/method is marked as @[expand_simple_interpolation]
+	is_unwrapped_fn_selector       bool // true, when the call is from an unwrapped selector (e.g. if t.foo != none { t.foo() })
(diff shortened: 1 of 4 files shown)


spytheman (approved):
Excellent work.

PR #23191, opened 2024-12-17
cgen,checker: minor optimizations

felipensp:
- Removed repetitive .has_flag() calls on fns
- Inline markused.walker or_block()
- Spread .has_option_or_result() instead of double check .has_flag(.option) + .has_flag(.result) calls.

image

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzYxNzU2MGYxNDRmNTg1YWU0ZGM2MjkiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.JKl35d2-H77ELwqaGU2ReqATb2xSX_cDlk3rFTuAsRE">Huly&reg;: <b>V_0.6-21628</b></a></sub>

### DIFF
--- a/vlib/v/checker/struct.v
+++ b/vlib/v/checker/struct.v
@@ -6,7 +6,6 @@ import v.ast
 import v.util
 import v.token
 
-@[direct_array_access]
 fn (mut c Checker) struct_decl(mut node ast.StructDecl) {
 	util.timing_start(@METHOD)
 	defer {
--- a/vlib/v/checker/struct.v
+++ b/vlib/v/checker/struct.v
@@ -6,6 +6,7 @@ import v.ast
 import v.util
 import v.token
 
+@[direct_array_access]
 fn (mut c Checker) struct_decl(mut node ast.StructDecl) {
 	util.timing_start(@METHOD)
 	defer {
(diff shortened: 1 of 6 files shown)


vlib/v/gen/c/fn.v
```diff
@@ -2578,6 +2577,7 @@ fn (mut g Gen) autofree_call_postgen(node_pos int) {
 	// g.doing_autofree_tmp = false
 }
 
+@[direct_array_access]
 fn (mut g Gen) call_args(node ast.CallExpr) {
```
spytheman, line 2581: That does not look like a good trade off to me, because the method is very long, and changed relatively frequently, thus the possibility for introducing bugs that the bounds checker is intended to catch is high.

The core of the method is also not a hot loop, doing exclusively array index accesses, thus the overhead of the bound checks is relatively low.

-> adding the tag is most likely to tradeoff segfaults, instead of panics, when bugs happen, for a not entirely clear performance optimisation.

vlib/v/checker/struct.v
```diff
@@ -6,6 +6,7 @@ import v.ast
 import v.util
 import v.token
 
+@[direct_array_access]
```
spytheman, line 9: Same - the same argument as for `call_args` apply.
Imho remove the attribute.

spytheman (approved):
Excellent work.

Delta456 (approved):
Good optimization

PR #23200, opened 2024-12-18
v: fix all memleaks to enable memleak detection

felipensp:
This PR aims to fix all memleak found on sanitize tests for enabling detect_leaks=1

PS: Some sqlite test were fixed because indirect leak was spotted on gcc sanitizer.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzYyYTdhYmYxNDRmNTg1YWU0ZGVhNGQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.uqgmbpzCfIxqpEj-EULPtm4Jnp4KGFfY9glHU9RKKoQ">Huly&reg;: <b>V_0.6-21637</b></a></sub>

### DIFF
--- a/.github/workflows/sanitized_ci.yml
+++ b/.github/workflows/sanitized_ci.yml
@@ -145,13 +145,11 @@ jobs:
           ./v retry -- sudo apt install clang
       - name: Recompile V with -cstrict
         run: ./v -cg -cstrict -o v cmd/v
-      - name: Create LSAN suppression file for ignoring websocket test
-        run: echo "leak:mbedtls_ssl_setup" > suppress.txt
       - name: Self tests (-fsanitize=address)
-        run: ASAN_OPTIONS=detect_leaks=1:fast_unwind_on_malloc=0 LSAN_OPTIONS=suppressions=suppress.txt ./v -cflags "-fsanitize=address,pointer-compare,pointer-subtract" test-self vlib
+        run: ASAN_OPTIONS=detect_leaks=1:fast_unwind_on_malloc=0 LSAN_OPTIONS=suppressions=.github/workflows/run_sanitizers_leak.suppressions ./v -cflags "-fsanitize=address,pointer-compare,pointer-subtract" test-self vlib
       - name: Self tests (V compiled with -fsanitize=address)
         run: ./v -cflags -fsanitize=address -o v cmd/v &&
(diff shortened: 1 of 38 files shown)


.github/workflows/sanitized_ci.yml
```diff
@@ -145,13 +145,15 @@ jobs:
           ./v retry -- sudo apt install clang
       - name: Recompile V with -cstrict
         run: ./v -cg -cstrict -o v cmd/v
+      - name: Create LSAN suppression file for ignoring websocket test
+        run: echo "leak:mbedtls_ssl_setup" > suppress.txt
```
spytheman, line 149: imho add it as a file in the repo instead of creating it inline here, see for example the existing `.github/workflows/run_sanitizers_thread.suppressions` and `.github/workflows/run_sanitizers_undefined.suppressions` .

felipensp: Oh great.

vlib/crypto/ecdsa/ecdsa_test.v
```diff
@@ -12,13 +12,15 @@ fn test_ecdsa() {
 	is_valid := pub_key.verify(message, signature) or { panic(err) }
 	println('Signature valid: ${is_valid}')
 	assert is_valid
+	C.EC_KEY_free(priv_key.key)
```
spytheman, line 15: Please make a V wrapper for that C function in `vlib/crypto/ecdsa/ecdsa.v` and call it in the test instead. There are no C. calls other than it in the test file.

spytheman (approved):
Excellent work.
Thank you 🙇🏻‍♂️.

spytheman:
I think we may have to skip the orm tests again for now (the ones that use sqlite) 🤔 .

felipensp:
> I think we may have to skip the orm tests again for now (the ones that use sqlite) 🤔 .

Ok, I'll treat the known cases in the .suppressions file.

PR #23203, opened 2024-12-18
veb: fix key value and translation file name

Ddiidev:
The translation is not working on Windows because it considers folder separation based on the Unix "/" separator.
It looks something like:
❌ `{'/pt-br': {'msg_hello': 'Olá'}}`
Instead of
✅ `{'pt-br': {'msg_hello': 'Olá'}}`

And correct the value so as not to catch the line break in the key value.


spytheman:
Can you please also add a small test?

I am not comfortable with merging changes like this one, without tests, since they have platform dependent changes, and tomorrow, some one else, may come and change it again to fix it on say macos/linux, and then break your fix for windows :-| .

spytheman:
With a test, we will know right away when/if it is broken in the future.

PR #23230, opened 2024-12-21
parallel-cc: fix build when closure is used

felipensp:
sFix
image
when building ui/examples/users.v

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzY2OTc3ZGYxNDRmNTg1YWU0ZWFkOTciLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.ZHzRPlbpdB_h9uxvjQdssOBFNGWT__qrBM9xGZQ822Q">Huly&reg;: <b>V_0.6-21664</b></a></sub>

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -4413,9 +4413,11 @@ fn (mut g Gen) gen_closure_fn(expr_styp string, m ast.Fn, name string) {
 		sb.write_string('${g.styp(param.typ)} a${i}')
 	}
 	sb.writeln(') {')
-	fn_header := sb.str()
-	g.extern_out.writeln('${fn_header.all_before(' {')};')
-	sb.writeln(fn_header)
+	if g.pref.parallel_cc {
+		fn_header := sb.str()
+		g.extern_out.writeln('${fn_header.all_before(' {')};')
+		sb.writeln(fn_header)
+	}
 	sb.writeln('\t${data_styp}* a0 = __CLOSURE_GET_DATA();')
 	if m.return_type != ast.void_type {
 		sb.write_string('\treturn ')
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -4413,6 +4413,9 @@ fn (mut g Gen) gen_closure_fn(expr_styp string, m ast.Fn, name string) {
 		sb.write_string('${g.styp(param.typ)} a${i}')
 	}
 	sb.writeln(') {')
+	fn_header := sb.str()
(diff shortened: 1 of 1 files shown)


medvednikov:
Thanks! Also please add the compilation of the example to CI.

felipensp:
> Thanks! Also please add the compilation of the example to CI.

The another issue still needing to be fixed yet. Adding now it will break ci.

vlib/v/gen/c/cgen.v
```diff
@@ -4413,6 +4413,11 @@ fn (mut g Gen) gen_closure_fn(expr_styp string, m ast.Fn, name string) {
 		sb.write_string('${g.styp(param.typ)} a${i}')
 	}
 	sb.writeln(') {')
+	if g.pref.parallel_cc {
+		fn_header := sb.str()
+		g.extern_out.writeln('${fn_header.all_before(' {')};')
+		sb.writeln(fn_header)
+	}
```
spytheman, line 4420: That seems very weird - you have already written the `{`, and then you are cutting it back.

Why not just generate different code depending on the flag, instead of doing that maneuver?

felipensp: It is cutting it back to g.extern_out, for `sb` it still. I'm just reusing already written buffer.

spytheman: You are doing `sb.str()`, which empties `sb`, followed by `sb.writeln(fn_header)` which fills it back up.
It is needless copying.

spytheman: At the very least, you can call `sb.bytestr()` to get a string, without emptying it :-( .

spytheman: And you can do that before line 4415, so that you would not have to do the .all_before() etc.

felipensp: > At the very least, you can call `sb.bytestr()` to get a string, without emptying it :-( .

Oh, I didn't have seen .bytestr() when looking for stuff like this on strings.builder impl. Thanks.

spytheman: `strings.Builder` is an alias to `[]u8`, and has all the methods that are defined on `[]u8` too.

PR #23231, opened 2024-12-21
parser,cgen: fix `index` variable type on array initialization to be compatible with element integer type

### LINKED ISSUE #23225: issue with array initialization using index
### Discussed in https://github.com/vlang/v/discussions/23224

<div type='discussions-op-text'>

<sup>Originally posted by **Lord-Vlad** December 20, 2024</sup>
I just started working with V, and while experimenting with arrays I stumbled upon the following.
When you initialize an `i32` array, you can't use index directly, you have to cast to `i32`.
Aren't `int` and `i32` the same thing?
This also applies to unsigned numbers, although I suspect it's due to the lack of promotion? Floating point numbers do work.

```v
array := []i32{len: 3, init: index}

error: expected `i32`, not `int`
```</div>
> [!NOTE]
>

### PR DESCRIPTION
**Pull request by felipensp**
Fix #23225

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzY2YTlhOGYxNDRmNTg1YWU0ZWFmOGEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.-K6hwOhqVeTy_nX8LHoXGv7Ek97AyiTgXprx8SRqs4A">Huly&reg;: <b>V_0.6-21665</b></a></sub>

### CODE REVIEW

**Review by Delta456 — COMMENTED** [review]
Does it work for strings and complex types? If not then we should have a checker error

**Comment by JalonSolov** [comment]
From the looks of the generated code, it will only work for integer types, not strings, floats, etc.

**Comment by JalonSolov** [comment]
Perhaps this is the wrong way to handle things.  Instead of changing the type of the index, it should cast the index to the proper type of value to put into the array.

**Comment by felipensp** [comment]
> Does it work for strings and complex types? If not then we should have a checker error

It just promote integer types.

**Comment by felipensp** [comment]
> Perhaps this is the wrong way to handle things. Instead of changing the type of the index, it should cast the index to the proper type of value to put into the array.

But it expected in the end to be numeric type only.

**Comment by JalonSolov** [comment]
Yes, I know it currently only support integer types.  However, if `index` were cast to the correct type, it would also work for floats (cast as `f32(index)` or `f64(index)`) and strings (`index.str()`), at least.  Still wouldn't work for an array of structs...

**Comment by spytheman** [comment]
`index` is always int, for the same reason, that `for i in 0..a.len {` causes `i` to be an int.

It should not be casted automatically to what the element type is.
You can have for example:
```v
dump( []u8{ len: 1000, init: u8(if index > 500 { 1 } else {0}) } )
```
which produces 0s, followed by 1s,
vs
```v
dump( []u8{ len: 1000, init: u8(if u8(index) > 500 { 1 } else {0}) } )
```
which produces only 0s.

As for `int` and `i32` - yes, `int` should be currently able to be used everywhere that expects `i32`, and `i32` should be able to be used everywhere that expects `int`, without needing more casts.

felipensp:
So why closing it?

spytheman:
Because the title is wrong, and the issue it is solving is invalid too.

spytheman:
A new issue and a new PR will be less confusing.

PR #23233, opened 2024-12-21
json: fix codegen for struct default expr

### LINKED ISSUE #23216: json.decode: code generation bug
I was unearthing an old project and ran into the following bug:

The reduced version below, as well as the version depended on by my actual main module, generates a C file containing the line `		res.results = main__Result* _t4 = (main__Result*) _v_malloc((0) * sizeof(main__Result));`. Which definitely has faulty assignment syntax but may also have a semantic issue in that it's seemingly allocating nothing.

It occurs in the `json__decode_main__Response` function definition.

The not-so-bad news is that I can still decode into `map[string][]map[string]string`.

Not sure when I'll have time to dig deeper, but if anybody can point out if I'm missing a better approach or even another workaround, I'd be much obliged!

If you want to test the whole project, it can be found at <https://github.com/kendfss/urdi>

**V doctor:**

```
V full version: V 0.4.8 afc07f4
OS: linux, "Void Linux"
Processor: 4 cpus, 64bit, little endian, Intel(R) Pentium(R) Gold 7505 @ 2.00GHz

getwd: /home/kendfss/gitclone/clones/kendfss/urdi
vexe: /home/kendfss/gitclone/clones/vlang/v/v
vexe mtime: 2024-12-18 06:20:12

vroot: OK, value: /home/kendfss/gitclone/clones/vlang/v
VMODULES: OK, value: /home/kendfss/.vmodules
VTMP: OK, value: /tmp/v_1000

Git vroot status: weekly.2024.51-12-gafc07f46
.git/config present: true

CC version: cc (GCC) 13.2.0
emcc version: N/A
thirdparty/tcc status: thirdparty-linuxmusl-amd64 a3e24da2
```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg debug.v`

```v
module main

import net.http
import net.urllib
import json

const endpoint = 'https://api.urbandictionary.com/v0/define'

// Response describes the data structure that comes back from the
// Urban Dictionary API.
pub struct Response {
pub:
	results []Result = []Result{len: 0} @[json: list]
	tags    []string = []string{len: 0} @[json: tags]
	kind    string   @[json: result_type]
}

// Result holds the information for a given definition.
pub struct Result {
pub:
	id            int      @[json: defid]
	author        string   @[json: author]
	definition    string   @[json: definition]
	link          string   @[json: permalink]
	thumbs_down   int      @[json: thumbs_down]
	thumbs_up     int      @[json: thumbs_up]
	word          string   @[json: word]
	date          string   @[json: written_on]
	audio_samples []string @[json: sound_urls]
	example       string   @[json: example]
}

// popularity states the Result.thumbs_up-Result.thumbs_down
pub fn (r Result) popularity() int {
	return r.thumbs_up - r.thumbs_down
}

// define returns the first page of results
pub fn define(word string) !&Response {
	link := '${endpoint}?page=1&term=${urllib.query_escape(word)}'
	resp := http.get(link) or { return err }
	response := json.decode(Response, resp.body) or { return err }
	return &response
}

@[inline]
fn fstr(def Result) string {
	return '${def.definition}\n--(${def.author}) <${def.link}> ${def.thumbs_up - def.thumbs_down}'
}

pub fn (r Result) str() string {
	return fstr(r)
}

fn main() {
  r := define("peng")!
  println(r)
}
```

that the code would execute when I `v run ...`

**What did you see instead?**

```
cc: /tmp/v_1000/debug.01JFCCSW6433T7NHG4FEWD0063.tmp.c:6128: error: '_t4' undeclared
builder error: 
==================
```

> [!NOTE]
>

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzY0ZDcwM2YxNDRmNTg1YWU0ZTY3ZmIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.rZBe2-M1BoHcPEvLnZnuOSdNnIX6AdSYFYyQXY6cSLY">Huly&reg;: <b>V_0.6-21652</b></a></sub>

felipensp:
Fix #23216

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzY3Mzk2NmYxNDRmNTg1YWU0ZWI5YzMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.E3u6R-ywdnSsZoZNxe9Bu_C64HAwebNtzjNsexTSbbg">Huly&reg;: <b>V_0.6-21667</b></a></sub>

### DIFF
--- a/vlib/json/tests/json_decode_struct_with_default_test.v
+++ b/vlib/json/tests/json_decode_struct_with_default_test.v
@@ -1,5 +1,3 @@
-module main
-
 import json
 
 pub struct Response {
@@ -11,14 +9,26 @@ pub:
 
 pub struct Result {
 pub:
-	id         int    @[json: defid]
-	author     string @[json: author]
-	definition string @[json: definition]
+	id            int      @[json: defid]
+	author        string   @[json: author]
+	definition    string   @[json: definition]
+	link          string   @[json: permalink]
+	thumbs_down   int      @[json: thumbs_down]
+	thumbs_up     int      @[json: thumbs_up]
+	word          string   @[json: word]
+	date          string   @[json: written_on]
+	audio_samples []string @[json: sound_urls]
+	example       string   @[json: example]
+}
+
+pub fn define(word string) !&Response {
+	resp := '{"list":[{"defid":3439287}]}'
+	response := json.decode(Response, resp) or { return err }
+	return &response
 }
 
 fn test_main() {
-	resp := '{"list": [{"defid":123, "author": "", "definition": "" }]}'
-	response := json.decode(Response, resp)!
+	response := define('')!
 	assert response.results.len > 0
-	assert response.results[0].id == 123
+	assert response.results[0].id == 3439287
 }
--- a/vlib/json/tests/json_decode_struct_with_default_test.v
+++ b/vlib/json/tests/json_decode_struct_with_default_test.v
@@ -1,0 +1,24 @@
+module main
+
+import json
+
+pub struct Response {
+pub:
+	results []Result = []Result{len: 0} @[json: list]
+	tags    []string = []string{len: 0} @[json: tags]
+	kind    string   @[json: result_type]
+}
+
+pub struct Result {
+pub:
+	id         int    @[json: defid]
+	author     string @[json: author]
+	definition string @[json: definition]
+}
+
+fn test_main() {
+	resp := '{"list": [{"defid":123, "author": "", "definition": "" }]}'
+	response := json.decode(Response, resp)!
+	assert response.results.len > 0
+	assert response.results[0].id == 123
+}
--- a/vlib/v/gen/c/json.v
+++ b/vlib/v/gen/c/json.v
@@ -846,7 +846,14 @@ fn (mut g Gen) gen_struct_enc_dec(utyp ast.Type, type_info ast.TypeInfo, styp st
 				}
 				if field.has_default_expr {
 					dec.writeln('\t} else {')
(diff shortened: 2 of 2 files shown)


vlib/json/tests/json_decode_struct_with_default_test.v
```diff
@@ -0,0 +1,24 @@
+module main
```
spytheman, line 1: The test passes on master too.

spytheman: image

spytheman: imho change the test to reflect the fix, so that it can serve better as a regression one

felipensp: oops, my fault.

spytheman (approved):
Excellent work.
Thank you 🙇🏻‍♂️ .

Issue #23228, opened 2024-12-21
Use freestanding mode, but encounter incompatible size_t redefinition error by C


PegasusPlusUS:
**V doctor:**
```
V full version: V 0.4.8 2ab1523
OS: windows, Microsoft Windows Server 2022 Standard v20348 64-bit
Processor: 8 cpus, 64bit, little endian, 

getwd: E:\Export\tinyos\ch01_bootsector\v
vexe: C:\Users\ping\scoop\apps\v\0.4.8\v.exe
vexe mtime: 2024-09-28 16:45:40

vroot: OK, value: C:\Users\ping\scoop\apps\v\0.4.8
VMODULES: OK, value: C:\Users\ping\.vmodules
VTMP: OK, value: C:\Users\ping\AppData\Local\Temp\2\v_0

Git version: git version 2.45.1.windows.1
Git vroot status: Error: fatal: not a git repository (or any of the parent directories): .git
.git/config present: false

CC version: Error: 'cc' is not recognized as an internal or external command,

operable program or batch file.


thirdparty/tcc: N/A

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg bootsector.v`
```v
fn main() {
    message := "Hello, World!"

    for ch in message {
        asm x86 {
            mov ah, 0x0e
            mov al, ch
            int 0x10
        }
    }

    // Infinite loop to keep the program running
    for {}
}

```

**What did you expect to see?**

Compile OK

**What did you see instead?**
```
bootsector.v:4:9: warning: unused variable: `ch`
    2 |     message := "Hello, World!"
    3 | 
    4 |     for ch in message {
      |         ~~
    5 |         asm x86 {
    6 |             mov ah, 0x0e
C:/Users/ping/AppData/Local/Temp/2/v_0/bootsector.01JFM5TKDDTX2P0VENQFFRDHXB.tmp.c:7059: at print_backtrace_skipping_top_frames_tcc: Backtrace
C:/Users/ping/AppData/Local/Temp/2/v_0/bootsector.01JFM5TKDDTX2P0VENQFFRDHXB.tmp.c:7026: by print_backtrace_skipping_top_frames
C:/Users/ping/AppData/Local/Temp/2/v_0/bootsector.01JFM5TKDDTX2P0VENQFFRDHXB.tmp.c:7927: by unhandled_exception_handler
7ffe189fbfc2 : by ???
[...]

PR #23237, opened 2024-12-22
pref:  freestanding supported only on  linux targets currently.

kbkpbot:
As issue #23228

Maybe we can add windows support later...

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzY3NmI3MGYxNDRmNTg1YWU0ZWJiNTgiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.T4Dn10v7VOPoSXVrt62wDXuLZ0X-wQt1YUX1QO8fbRw">Huly&reg;: <b>V_0.6-21671</b></a></sub>

### DIFF
--- a/vlib/v/pref/pref.v
+++ b/vlib/v/pref/pref.v
@@ -612,8 +612,13 @@ pub fn parse_args_and_show_errors(known_external_commands []string, args []strin
 				res.compress = true
 			}
 			'-freestanding' {
-				res.is_bare = true
-				res.build_options << arg
+				$if macos || linux {
+					res.is_bare = true
+					res.build_options << arg
+				} $else {
+					eprintln('`-freestanding` supported only on `linux` targets currently.')
+					exit(1)
+				}
 			}
 			'-no-retry-compilation' {
 				res.retry_compilation = false


vlib/v/pref/pref.v
```diff
@@ -612,8 +612,13 @@ pub fn parse_args_and_show_errors(known_external_commands []string, args []strin
...
 			'-freestanding' {
-				res.is_bare = true
-				res.build_options << arg
+				$if macos || linux {
+					res.is_bare = true
+					res.build_options << arg
+				} $else {
+					eprintln('`-freestanding` supported only on `linux` targets currently.')
+					exit(1)
+				}
```
spytheman, line 621: No, that is not true, and you can use -freestanding for freebsd and other OSes too.

If you want, you can single out windows, but imho even that is dubious.
The option exists, and it can be useful there too, I do not want for it to be limited artificially.

PR #23296, opened 2024-12-28
log: set stderr as default log output, add .set_output_stream()

gechandesu:
This PR:

- Adds `set_output_stream()` fn to `log.Log` struct which allows setting `io.Writer` as log output. It makes possible send logs to arbitrary targets such as sockets. Use cases: write logs to system journal e.g. syslog, send logs to remote server over UDP.
- Makes STDERR the default output stream (be STDOUT before). Sending logs to STOUT is not generally used in logging libraries e.g. Go, Python stdlibs etc.
- Edit test, since `log.Log` now have `output_stream` field with file address which can change at runtime and therefore cannot be tested in "inout" tests.

Example:

```v
import log
import os
import net

fn main() {
	mut logger := log.Log{}
	logger.set_level(.info)
	log.set_logger(logger)

	logger.info('hello stderr')
	logger.set_output_stream(os.stdout())
	logger.info('hello stdout')

	mut udp_sock := net.dial_udp('127.0.0.1:8889')!
	logger.set_output_stream(udp_sock)
	logger.info('hello udp')
}
```

Screenshot From 2024-12-28 17-52-38

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzcwMTZiZGYxNDRmNTg1YWU1MTM4NDIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.2rSSSthRbF0_I19-1HsJMgVWY3W0s-JBJO6aIP5OYLo">Huly&reg;: <b>V_0.6-21728</b></a></sub>

### DIFF
--- a/vlib/log/log.v
+++ b/vlib/log/log.v
@@ -37,9 +37,9 @@ mut:
 	custom_time_format string     = 'MMMM Do YY N kk:mm:ss A' // timestamp with custom format
 	short_tag          bool
 	always_flush       bool // flush after every single .fatal(), .error(), .warn(), .info(), .debug() call
+	output_stream      io.Writer = os.stderr()
 pub mut:
 	output_file_name string // log output to this file
-	output_stream    io.Writer = os.stderr()
 }
 
 // get_level gets the internal logging level.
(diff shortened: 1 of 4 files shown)


vlib/log/log.v
```diff
@@ -38,6 +39,7 @@ mut:
 	always_flush       bool // flush after every single .fatal(), .error(), .warn(), .info(), .debug() call
 pub mut:
 	output_file_name string // log output to this file
+	output_stream    io.Writer = os.stderr()
```
spytheman, line 42: why is this in the `pub mut:` section, instead of in the `mut:` one?

gechandesu: fixed, now `output_stream` is private.

spytheman:
It failed locally the valgrind test, I am reopening it after the revert.

For some reason the test was not triggered by the last commit, which led me to believe that it was fixed. I should have investigated it closer.

PR #23319, opened 2024-12-30
json: support `@[json_null]` tag to enforce `null`, when encoding `none` option values

felipensp:
Implementing feature request.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzcyYmM3NmYxNDRmNTg1YWU1MTZmMmMiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.IuqTMFe-F_ZEWbZgfl9dlS5Z2iuZ5myuO48HUP_kkqA">Huly&reg;: <b>V_0.6-21751</b></a></sub>

### DIFF
--- a/vlib/json/tests/json_is_null_attr_test.v
+++ b/vlib/json/tests/json_is_null_attr_test.v
@@ -1,0 +1,26 @@
+import json
+
+struct Bar {
+	name ?string @[json_null]
+}
+
+struct Foo {
+	name   ?string @[json_null]
+	age    ?int    @[json_null]
+	text   ?string
+	other  ?Bar
+	other2 ?Bar @[json_null]
+}
+
+fn test_main() {
+	assert json.encode(Foo{}) == '{"name":null,"age":null,"other2":null}'
+	assert json.encode(Foo{ name: '' }) == '{"name":"","age":null,"other2":null}'
+	assert json.encode(Foo{ age: 10 }) == '{"name":null,"age":10,"other2":null}'
+	assert json.encode(Foo{
+		age:    10
+		other2: Bar{
+			name: none
+		}
+	}) == '{"name":null,"age":10,"other2":{"name":null}}'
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

enghitalo:
What about just `@[nulltable]` so this can make sense to others encoders as well

enghitalo:
IMHO, `@[required; skip]` should do the same work (What I think would be better)

spytheman:
> IMHO, `@[required]`, or `@[required; skip]`, or `@[required; omitempty]` should do the same work (What I think would be better)

No. The current name @[json_null] is better - it has `json` in it, and so it is restricted to only the `json` modules. This avoids the conflict with the decisions of other modules, that may also do serialization.

spytheman:
`omitempty` in particular does the opposite, and is only valid for non option fields.

JalonSolov:
Logically, `@[omitempty]` should apply to Option fields as well.  If the Option is set to `none` (the default), then it is "empty".  Same as a string field set to `''`.

spytheman:
The *default* for option fields, with a value of `none`, is that they are already omitted when serialized to json => adding a `@[omitempty]` tag to them is redundant to that default, and non effective for them.

JalonSolov:
That sounds like "magic".  It would make more sense to output `null` unless `omitempty` was used.

spytheman:
No, it does not sound like "magic" - the default behavior is to not produce anything for option values that are `none`, and that saves resources, and that is not changed by the PR.

PR #23321, opened 2024-12-30
markused: fix for iterator method not marked as used

### LINKED ISSUE #23312: C compiler bug when running this code
### Describe the bug

I'm a total V beginner, but bear with me, i was trying to make a language analysis tool for a simple project, but it just spits a C error

### Reproduction Steps

Here's the (simplified) code:

```v
// main.v
module main

import os

fn main() {
    file := os.read_file(...) or { panic("Couldn't open") }
    
    for node in Parser.new(file) { dump(node) }
}

pub fn todo() { panic("todo") }

pub fn first[T, U](iter T, f fn(usize, U) bool) ?U {
    mut ret := ?U(none)
    for i, e in iter { if f(usize(i), e) {
        ret = e
        break
    } }
    return ret
}

pub fn take_while(s string, f fn(usize, rune) bool) string {
    mut runes := [] rune {}
    for i, r in s.runes() {
        if !f(usize(i), r) { break }
        runes << r
    }
    
    return runes.string()
}

// lexer.v
module main

import encoding.utf8

pub type Token = TokenType1 | TokenType2...

pub struct Lexer {
    str    string
    mut: i usize
}

pub fn Lexer.new(str string) Lexer { return Lexer { str, 0 } }

pub fn (self Lexer) peek() ?Token {
    tkn, _ := self.lex()?
    return tkn
}

pub fn (mut self Lexer) next() ?Token {
    tkn, i := self.lex()?
    self.i += i
    return tkn
}

fn (self Lexer) lex() ?(Token, usize) { ... }

// parser.v
module main

pub type Node = NodeTypes...

pub fn (self Node) is_expr() bool { ... }

pub struct Parser { mut: lexer Lexer }

pub fn Parser.new(str string) Parser { return Parser { Lexer.new(str) } }

pub fn (mut self Parser) next() ?Node { ... }
```

### Expected Behavior

It should just work or at least not cause c errors

### Current Behavior

```
cc: /data/data/com.termux/files/usr/tmp/v_10246/hex.01JGAX0GMQ7AY9BBNGWHZ5Y2EK.tmp.c:8745:29: error: call to undeclared function 'main__Lexer_next'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
cc:  8745 |                 _option_main__Token _t2 = main__Lexer_next(&_t1);
cc:       |                                           ^
cc: /data/data/com.termux/files/usr/tmp/v_10246/hex.01JGAX0GMQ7AY9BBNGWHZ5Y2EK.tmp.c:8745:29: note: did you mean 'main__Lexer_lex'?
cc: /data/data/com.termux/files/usr/tmp/v_10246/hex.01JGAX0GMQ7AY9BBNGWHZ5Y2EK.tmp.c:8658:56: note: 'main__Lexer_lex' declared here
cc:  8658 | VV_LOCAL_SYMBOL _option_multi_return_main__Token_usize main__Lexer_lex(main__Lexer self) {
cc:       |                                                        ^
cc: /data/data/com.termux/files/usr/tmp/v_10246/hex.01JGAX0GMQ7AY9BBNGWHZ5Y2EK.tmp.c:8745:23: error: initializing '_option_main__Token' (aka 'struct _option_main__Token') with an expression of incompatible type 'int'
cc:  8745 |                 _option_main__Token _t2 = main__Lexer_next(&_t1);
cc:       |                                     ^     ~~~~~~~~~~~~~~~~~~~~~~
cc: 2 errors generated.
builder error:
==================
```

### Possible Solution

_No response_

In the function `first`, the original code just returned `e` if `f(...)` were true, but that was yielding a V error and i did this instead. This is a helper function that gets the first element matching some criteria, if any. I use it on the `Parser.next` function to ignore skippable tokens

### V version

V 0.4.9 4225a34

V full version: V 0.4.9 b487986.4225a34
OS: termux, 3.18.140-SIMPLE-KERNEL_V1.1, #1 SMP PREEMPT Sun Apr 18 03:22:59 -03 2021
Processor: 4 cpus, 64bit, little endian

getwd: /storage/emulated/0/Documents/hex
vexe: /data/data/com.termux/files/home/.v/v
vexe mtime: 2024-12-30 04:25:37

vroot: OK, value: /data/data/com.termux/files/home/.v
VMODULES: OK, value: /data/data/com.termux/files/home/.vmodules
VTMP: OK, value: /data/data/com.termux/files/usr/tmp/v_10246

Git version: git version 2.47.1
Git vroot status: 4225a346
.git/config present: true

CC version: clang version 19.1.6
emcc version: N/A
thirdparty/tcc status: thirdparty-unknown-unknown de82a130

> [!NOTE]
>

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzcyMjU3YmYxNDRmNTg1YWU1MTVhZTYiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.kdtfGW7_JwVVNMdk4QbtAojBDmqBZmygncMR1DRr100">Huly&reg;: <b>V_0.6-21744</b></a></sub>

felipensp:
Fix #23312

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzcyZGNjZWYxNDRmNTg1YWU1MTcyNmQiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.tIQheEJawr3iGPNOaSbvDFfXXvY3xfApbAtOtzvbCOY">Huly&reg;: <b>V_0.6-21753</b></a></sub>

### DIFF
--- a/vlib/v/checker/for.v
+++ b/vlib/v/checker/for.v
@@ -173,6 +173,8 @@ fn (mut c Checker) for_in_stmt(mut node ast.ForInStmt) {
 			unwrapped_typ := c.unwrap_generic(typ)
 			unwrapped_sym := c.table.sym(unwrapped_typ)
 
+			c.table.used_features.comptime_calls['${int(unwrapped_typ)}.next'] = true
+
 			if node.key_var.len > 0 {
 				key_type := match unwrapped_sym.kind {
 					.map { unwrapped_sym.map_info().key_type }
--- a/vlib/v/tests/skip_unused/generic_iterator_loop.vv
+++ b/vlib/v/tests/skip_unused/generic_iterator_loop.vv
@@ -1,0 +1,32 @@
+module main
+
+interface IFoo {
+	next() ?Foo
+}
+
+struct Foo {}
+
+fn (f Foo) next() ?Foo {
+	return none
+}
+
+struct Bar {}
+
+fn (f Bar) next() ?Bar {
+	return none
+}
+
+fn loop[T](iter T) {
+	for _ in iter {
+	}
+}
+
+fn loop2(iter IFoo) {
+	for _ in iter {
+	}
+}
+
+fn main() {
+	loop(Foo{})
+	loop(Bar{})
+}
--- a/vlib/v/tests/skip_unused/generic_iterator_loop.vv
+++ b/vlib/v/tests/skip_unused/generic_iterator_loop.vv
@@ -1,9 +1,5 @@
 module main
 
-interface IFoo {
-	next() ?Foo
-}
-
 struct Foo {}
 
 fn (f Foo) next() ?Foo {
@@ -17,11 +13,6 @@ fn (f Bar) next() ?Bar {
 }
 
 fn loop[T](iter T) {
-	for _ in iter {
-	}
-}
-
-fn loop2(iter IFoo) {
 	for _ in iter {
 	}
 }


vlib/v/checker/for.v
```diff
@@ -173,6 +173,8 @@ fn (mut c Checker) for_in_stmt(mut node ast.ForInStmt) {
 			unwrapped_typ := c.unwrap_generic(typ)
 			unwrapped_sym := c.table.sym(unwrapped_typ)
 
+			c.table.used_features.comptime_calls['${int(unwrapped_typ)}.next'] = true
```
spytheman, line 176: hm ... interesting ... 

Why is this in this branch, and not in the `if sym.kind in [.struct, .interface] {` one ?

spytheman: Oh ... it is in a generic function, that has `fn loop[T](iter T) {` ...

spytheman: Is there some code, that marks the .next() methods for the non generic case?

felipensp: Yes. On the walker.
image

spytheman: Thanks.

spytheman: Will it not be better for maintainability, to have the setting here, moved there too?

felipensp: The generic resolving part only can be done on the checker. The walker does not handle all concrete types for the same FnDecl.

spytheman: ok

spytheman: What happens, if the generic functions, are called by concrete ones, with concrete types, but then none of those are called by main() ?

Will the .next() methods still be generated, even if they will not be called transitively by anything?

felipensp: Call resolved on checker-time are always generated disregarding it is used or not.

felipensp: > Currently yes.

Oops. Actually it is not generated if not used by main.

spytheman: thanks

PR #23326, opened 2024-12-31
gg: add linux support for `fn screen_size() Size` fix (#23146)

### LINKED ISSUE #23146: Update gg.c.v screen_size function for Linux and BSD-derivatives
### Describe the feature

The screen_size function of gg.c.v is non-functional for Linux and BSD-derivative OSes. Suggest using similar code as done by [islonely](https://github.com/islonely/mouse/blob/master/src/c/mouse_linux.c), which comes from Xlib.h (DefaultScreen(dpy)).

Example:

```
struct Size {
	int width, height;
};

struct Size screen_size() {
	Display *dpy = XOpenDisplay(0);
	int screen = DefaultScreen(dpy);
	struct Size sz = {
		DisplayWidth(dpy, screen),
		DisplayHeight(dpy, screen)
	};
	XCloseDisplay(dpy);
	return sz;
}
```

### Use Case

Will give screen size for Linux and BSD-type OSes.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### Version used

V 0.4.8

 Linux and BSD-derivatives
> [!NOTE]
>

islonely:
Fixes #23146 by adding support for Linux users using X11.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2Nzc0NGM4ZGYxNDRmNTg1YWU1MWNlOWEiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.nsEUjcC-vRZwG3V-Njl7nawqDrPsAP6Uzr71s55SKo8">Huly&reg;: <b>V_0.6-21757</b></a></sub>

### DIFF
--- a/vlib/gg/gg.c.v
+++ b/vlib/gg/gg.c.v
@@ -778,7 +778,7 @@ pub fn screen_size() Size {
 		if display == unsafe { nil } {
 			return Size{}
 		}
-		defer { _ := C.XCloseDisplay(display) }
+		defer { C.XCloseDisplay(display) }
 		root := C.DefaultRootWindow(display)
 		resources := C.XRRGetScreenResources(display, root)
 		if resources == unsafe { nil } {
--- a/vlib/gg/gg.c.v
+++ b/vlib/gg/gg.c.v
@@ -10,7 +10,7 @@ import sokol.sapp
 import sokol.sgl
 import sokol.gfx
 
-struct C.Display {}
+pub struct C.Display {}
 
 fn C.XOpenDisplay(int) &C.Display
 fn C.DefaultScreen(&C.Display) int
--- a/vlib/gg/gg.c.v
+++ b/vlib/gg/gg.c.v
@@ -750,7 +750,7 @@ pub fn screen_size() Size {
 		display := C.XOpenDisplay(0)
 		screen := C.DefaultScreen(display)
 		return Size{
-			width: C.DisplayWidth(display, screen)
+			width:  C.DisplayWidth(display, screen)
 			height: C.DisplayHeight(display, screen)
 		}
 	}
--- a/vlib/gg/gg.c.v
+++ b/vlib/gg/gg.c.v
@@ -10,6 +10,13 @@ import sokol.sapp
 import sokol.sgl
 import sokol.gfx
 
+struct C.Display {}
+
+fn C.XOpenDisplay(int) &C.Display
+fn C.DefaultScreen(&C.Display) int
+fn C.DisplayHeight(&C.Display, int) int
+fn C.DisplayWidth(&C.Display, int) int
+
 $if windows {
 	#flag -lgdi32
 	#include "windows.h"
@@ -739,7 +746,14 @@ pub fn screen_size() Size {
 			height: int(C.GetSystemMetrics(C.SM_CYSCREEN))
 		}
 	}
-	// TODO: linux, etc
+	$if linux {
+		display := C.XOpenDisplay(0)
+		screen := C.DefaultScreen(display)
+		return Size{
+			width: C.DisplayWidth(display, screen)
+			height: C.DisplayHeight(display, screen)
+		}
+	}
 	return Size{}
 }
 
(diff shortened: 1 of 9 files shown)


vlib/gg/gg.c.v
```diff
@@ -739,7 +746,14 @@ pub fn screen_size() Size {
 			height: int(C.GetSystemMetrics(C.SM_CYSCREEN))
 		}
 	}
-	// TODO: linux, etc
+	$if linux {
+		display := C.XOpenDisplay(0)
```
spytheman, line 750: I think that it needs `defer { C.XCloseDisplay(display) }` too.

spytheman: `XOpenDisplay` is also documented to return null on failure, that should be checked too

vlib/gg/gg.c.v
```diff
@@ -10,6 +10,13 @@ import sokol.sapp
 import sokol.sgl
 import sokol.gfx
 
+pub struct C.Display {}
```
spytheman, line 13: that may need `@[typedef]`

spytheman:
This program:
```v
#flag -lX11
#include <X11/Xlib.h>

@[typedef]
pub struct C.Display {}
fn C.XOpenDisplay(int) &C.Display
fn C.XCloseDisplay(display &C.Display) int
fn C.DefaultScreen(&C.Display) int
fn C.DisplayHeight(&C.Display, int) int
fn C.DisplayWidth(&C.Display, int) int

fn main() {
	println('start')
	display := C.XOpenDisplay(0)
	screen := C.DefaultScreen(display)
	dump( C.DisplayWidth(display, screen) )
	dump( C.DisplayHeight(display, screen) )
	C.XCloseDisplay(display)
	println('done')
}
```

locally produces:
```
start
[x.v:16] C.DisplayWidth(display, screen): 3840
[x.v:17] C.DisplayHeight(display, screen): 1080
done
```

I have 2 displays, both with a resolution of 1920x1080, according to `xfce4-display-settings`:
image

spytheman:
If I turn off the second monitor, I get:
```
start
[x.v:16] C.DisplayWidth(display, screen): 1920
[x.v:17] C.DisplayHeight(display, screen): 1080
done
```

islonely:
> If I turn off the second monitor, I get:
> 
> ```
> start
> [x.v:16] C.DisplayWidth(display, screen): 1920
> [x.v:17] C.DisplayHeight(display, screen): 1080
> done
> ```

It seems there's not a good quick way to the size of a single screen. I'll be back with a solution soon.

islonely:
`libxrandr-dev` is required on ubuntu based systems for this fix.

vlib/gg/gg.c.v
```diff
@@ -748,10 +775,37 @@ pub fn screen_size() Size {
...
 	$if linux {
 		display := C.XOpenDisplay(0)
-		screen := C.DefaultScreen(display)
-		return Size{
-			width:  C.DisplayWidth(display, screen)
-			height: C.DisplayHeight(display, screen)
+		if display == unsafe { nil } {
+			return Size{}
+		}
+		defer { _ := C.XCloseDisplay(display) }
```
JalonSolov, line 781: Does it need the `_ :=` here?  Do you get a warning/error without it?

JalonSolov: There's no problem having it there, and I highly recommend it in most places.  It's just that when it's in a `defer`, there's not much you can do if it goes wrong, so no need for the extra chars.

islonely: That makes sense. Just habitually typed it in there 😅

spytheman (approved):
Excellent work.

Issue #23214, opened 2024-12-19
Hot reloading invalid memory access on Windows with `-cc tcc`


CodeOptimist:
### Describe the bug

Hot reloading example seems to fail with `RUNTIME ERROR: invalid memory access`

### Reproduction Steps

`v -live run examples/hot_reload/message.v`
on a fresh V installation; Windows 11

### Current Behavior

```cmd
PS C:\Dropbox\vlang\examples\hot_reload> v -live run message.v
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
Unhandled Exception 0xC0000005
C:/Users/Chris/AppData/Local/Temp/v_0/message.01JFGGEH73ZXN1SG5KJGJEW3G1.tmp.c:5222: at print_backtrace_skipping_top_frames_tcc: Backtrace
C:/Users/Chris/AppData/Local/Temp/v_0/message.01JFGGEH73ZXN1SG5KJGJEW3G1.tmp.c:5208: by print_backtrace_skipping_top_frames
C:/Users/Chris/AppData/Local/Temp/v_0/message.01JFGGEH73ZXN1SG5KJGJEW3G1.tmp.c:5778: by unhandled_exception_handler
7fffb76ded36 : by ???
C:/Users/Chris/AppData/Local/Temp/v_0/tmp.1.message.01JFGGEHKS30AVVRA0GZ9PPZJR.tmp.so.c:4996: at strconv__format_dec_sb: RUNTIME ERROR: invalid memory access
C:/Users/Chris/AppData/Local/Temp/v_0/tmp.1.message.01JFGGEHKS30AVVRA0GZ9PPZJR.tmp.so.c:12319: by StrIntpData_process_str_intp_data
C:/Users/Chris/AppData/Local/Temp/v_0/tmp.1.message.01JFGGEHKS30AVVRA0GZ9PPZJR.tmp.so.c:12721: by str_intp
C:/Users/Chris/AppData/Local/Temp/v_0/tmp.1.message.01JFGGEHKS30AVVRA0GZ9PPZJR.tmp.so.c:15383: by impl_live_main__print_message
C:/Users/Chris/AppData/Local/Temp/v_0/message.01JFGGEH73ZXN1SG5KJGJEW3G1.tmp.c:2883: by main__print_message
C:/Users/Chris/AppData/Local/Temp/v_0/message.01JFGGEH73ZXN1SG5KJGJEW3G1.tmp.c:9676: by main__main
[...]

PR #23350, opened 2025-01-03
cgen: fix windows hot reload

kbkpbot:
Fix issue #23214

Force generate `_vinit_caller`, `_vcleanup_caller` , these are needed under Windows, because dl.open() / dl.close() will call them when loading/unloading shared dll.

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2Nzc3NDY4NWYxNDRmNTg1YWU1MjQxNDciLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.FiYiE_5Q5UQQVdy-ADsvLoIk8lh1_DD0d8SmsYJTsT4">Huly&reg;: <b>V_0.6-21781</b></a></sub>

### DIFF
--- a/vlib/v/live/common.c.v
+++ b/vlib/v/live/common.c.v
@@ -59,7 +59,7 @@ pub fn info() &LiveReloadInfo {
 	mut x := &LiveReloadInfo{}
 	unsafe {
 		mut p := &u64(&C.g_live_info)
-		*p = &u64(x)
+		*p = u64(x)
 		_ = p
 	}
 	return x
--- a/vlib/v/live/common.c.v
+++ b/vlib/v/live/common.c.v
@@ -48,13 +48,13 @@ pub mut:
 // live.info - give user access to program's LiveReloadInfo struct,
 // so that the user can set callbacks, read meta information, etc.
 pub fn info() &LiveReloadInfo {
-	if C.g_live_info == 0 {
+	if g_live_reload_info == 0 {
 		// When the current program is not compiled with -live, simply
 		// return a new empty struct LiveReloadInfo in order to prevent
 		// crashes. In this case, the background reloader thread is not
 		// started, and the structure LiveReloadInfo will not get updated.
 		// All its fields will be 0, but still safe to access.
-		C.g_live_info = int(&LiveReloadInfo{})
+		g_live_reload_info = &LiveReloadInfo{}
 	}
-	return &LiveReloadInfo(C.g_live_info)
+	return &LiveReloadInfo(g_live_reload_info)
 }
--- a/vlib/v/live/common.c.v
+++ b/vlib/v/live/common.c.v
@@ -54,9 +54,7 @@ pub fn info() &LiveReloadInfo {
 		// crashes. In this case, the background reloader thread is not
 		// started, and the structure LiveReloadInfo will not get updated.
 		// All its fields will be 0, but still safe to access.
-		unsafe {
-			C.g_live_info = int(&LiveReloadInfo{})
-		}
+		C.g_live_info = int(&LiveReloadInfo{})
 	}
-	return unsafe { &LiveReloadInfo(C.g_live_info) }
+	return &LiveReloadInfo(C.g_live_info)
 }
--- a/vlib/v/live/common.c.v
+++ b/vlib/v/live/common.c.v
@@ -48,19 +48,15 @@ pub mut:
 // live.info - give user access to program's LiveReloadInfo struct,
 // so that the user can set callbacks, read meta information, etc.
 pub fn info() &LiveReloadInfo {
-	if C.g_live_info != 0 {
-		return unsafe { &LiveReloadInfo(C.g_live_info) }
+	if C.g_live_info == 0 {
+		// When the current program is not compiled with -live, simply
+		// return a new empty struct LiveReloadInfo in order to prevent
+		// crashes. In this case, the background reloader thread is not
+		// started, and the structure LiveReloadInfo will not get updated.
+		// All its fields will be 0, but still safe to access.
+		unsafe {
+			C.g_live_info = int(&LiveReloadInfo{})
+		}
 	}
-	// When the current program is not compiled with -live, simply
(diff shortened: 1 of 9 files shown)


vlib/v/live/common.c.v
```diff
@@ -59,7 +59,7 @@ pub fn info() &LiveReloadInfo {
 	mut x := &LiveReloadInfo{}
 	unsafe {
 		mut p := &u64(&C.g_live_info)
-		*p = &u64(x)
+		*p = u64(x)
```
spytheman, line 62: why change this?

kbkpbot: This the generated C code:
```c
	v__live__LiveReloadInfo* x = ((v__live__LiveReloadInfo*)memdup(...}, sizeof(v__live__LiveReloadInfo)));
	{ // Unsafe block
		u64* p = ((u64*)(&g_live_info));
		*p = ((u64)(x));
		{u64* _ = p;}
		;
	}

```
This will create a result:
```sh
D:\v\v\v\examples\hot_reload>v -live run message.v
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
```

And I think this is wrong:
```c
	v__live__LiveReloadInfo* x = ((v__live__LiveReloadInfo*)memdup(...}, sizeof(v__live__LiveReloadInfo)));
		u64* p = ((u64*)(&g_live_info));
		*p = *((u64*)(x));
		{u64* _ = p;}
		;
	}

```
which will create a wrong result:
```sh
D:\v\v\v\examples\hot_reload>v -live run message.v
OK reloads:    0 | Total reloads:    0 | Hello! Modify this message while the program is running.
OK reloads: 1869376609 | Total reloads: 544501614 | Hello! Modify this message while the program is running.
OK reloads: 1869376609 | Total reloads: 544501614 | Hello! Modify this message while the program is running.
OK reloads: 1869376609 | Total reloads: 544501614 | Hello! Modify this message while the program is running.
OK reloads: 1869376609 | Total reloads: 544501614 | Hello! Modify this message while the program is running.
```

vlib/v/live/common.c.v
```diff
@@ -48,19 +48,13 @@ pub mut:
...
 pub fn info() &LiveReloadInfo {
-	if C.g_live_info != 0 {
-		return unsafe { &LiveReloadInfo(C.g_live_info) }
+	if C.g_live_info == 0 {
+		// When the current program is not compiled with -live, simply
+		// return a new empty struct LiveReloadInfo in order to prevent
+		// crashes. In this case, the background reloader thread is not
+		// started, and the structure LiveReloadInfo will not get updated.
+		// All its fields will be 0, but still safe to access.
+		C.g_live_info = int(&LiveReloadInfo{})
```
spytheman, line 57:
```suggestion
		C.g_live_info = voidptr(&LiveReloadInfo{})
```
int is 32 bit, and `g_live_info` is `void *`, which is 64 bit on most systems

kbkpbot: Yes, I know, but this will cause a warning message:
```sh
/media/HD/github/kbkpbot/v/vlib/v/live/common.c.v:57:17: warning: cannot assign a reference to a value (this will be an error soon) left=int false right=voidptr true ptr=false
   55 |         // started, and the structure LiveReloadInfo will not get updated.
   56 |         // All its fields will be 0, but still safe to access.
   57 |         C.g_live_info = voidptr(&LiveReloadInfo{})
      |                       ^
   58 |     }
   59 |     return &LiveReloadInfo(C.g_live_info)
```

even put it in a unsafe {}.

kbkpbot: BTW, why `C.g_live_info` is recognized as `int` not `voidptr`?
```c
// g_live_info is used by live.info()
static void* g_live_info = NULL;
```

spytheman: That is a problem indeed.
I'll merge it as it is, and then cleanup later - the problem is that this was implemented relatively early in V's history, and now we have a way to do it with V globals directly.

spytheman: hm, no ... I will have to do it here, because clang on macos is newer and stricter, and will not allow the `int` cast.

spytheman: > BTW, why `C.g_live_info` is recognized as `int` not `voidptr`?

V just assumes, that all C.symbols that lack declaration, are ints .

spytheman: It allows for `C.some_fn(C.some_enum_value)` , without having to declare all enum values.

spytheman: `g_live_info` was/is not declared on the V side, but only as a C global.

I'll clean it up later after merging this PR, once there is no code that uses it anymore on master.

phcreery:
I see this overlaps most of everything from an old PR except https://github.com/vlang/v/pull/19621/files#diff-c8aafc93f5a132c1afb63f4c56a73731d8368f0076262fc9fe26a209661690b6

spytheman:
There are some similarities, but this PR fixes a much deeper problem, and in a more maintainable way.

Issue #14994, opened 2022-07-08
Add level parameter | gzip.v


Ddiidev:
https://github.com/vlang/v/blob/70890b27a5bb5bb49876326701488571a916411c/vlib/compress/gzip/gzip.v#L11-L12

Currently, gzip compression level 0 is fixed.
I would like a flexible compression level.

PR #23370, opened 2025-01-04
compress.gzip: add compression flags

kbkpbot:
Fix issue #14994

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2Nzc5MzA2OWYxNDRmNTg1YWU1MjdiNjAiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.KgVodTiJAjCnSuFn-2SFFvkgSft2wGHECOkSYh2caj4">Huly&reg;: <b>V_0.6-21801</b></a></sub>

### DIFF
--- a/vlib/compress/gzip/gzip.v
+++ b/vlib/compress/gzip/gzip.v
@@ -7,35 +7,42 @@ import compress as compr
 import hash.crc32
 
 // CompressFlags
+// TODO: These flags have no use now
+@[flag]
 pub enum CompressFlags {
-	// 0: Huffman only
-	// 1: Huffman+LZ (fastest/crap compression)
-	// 4095: Huffman+LZ (slowest/best compression)
-	huffman_only       = 0
-	default_max_probes = 128
-	max_probes_mask    = 0xfff
+	// The low 12 bits will be overwritten by `compression_level`
+	compression_level_overwrite_flag01
+	compression_level_overwrite_flag02
+	compression_level_overwrite_flag03
+	compression_level_overwrite_flag04
+	compression_level_overwrite_flag05
+	compression_level_overwrite_flag06
+	compression_level_overwrite_flag07
+	compression_level_overwrite_flag08
+	compression_level_overwrite_flag09
+	compression_level_overwrite_flag10
+	compression_level_overwrite_flag11
+	compression_level_overwrite_flag12
 
 	// If set, the compressor outputs a zlib header before the deflate data, and the Adler-32 of the source data at the end. Otherwise, you'll get raw deflate data.
-	// write_zlib_header = 0x01000
+	write_zlib_header //= 0x01000
 	// Always compute the adler-32 of the input data (even when not writing zlib headers).
-	// compute_adler32 = 0x02000
+	compute_adler32 //= 0x02000
 	// Set to use faster greedy parsing, instead of more efficient lazy parsing.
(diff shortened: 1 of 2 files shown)


vlib/compress/gzip/gzip.v
```diff
@@ -6,10 +6,50 @@ module gzip
...
+// CompressParams set flags for compression:
+// `flags`:
+// 0: Huffman only;
+// 1: Huffman+LZ (fastest/crap compression);
+// 128: default_max_probes;
+// 4095: Huffman+LZ (slowest/best compression)
+@[params]
+pub struct CompressParams {
+pub:
+	flags int = int(CompressFlags.default_max_probes)
```
spytheman, line 46: Why is this an int, instead of `CompressFlags`?
That will make all usage sites clumsier.

spytheman: V supports `@[flag]` enums, and the different combinations can be expressed with them.

vlib/compress/gzip/gzip.v
```diff
@@ -40,12 +80,27 @@ pub fn compress(data []u8) ![]u8 {
 	return result
 }
 
+// DecompressFlags
+pub enum DecompressFlags {
```
spytheman, line 84: This can be a `@[flag]` enum.

vlib/compress/gzip/gzip.v
```diff
@@ -6,10 +6,50 @@ module gzip
 import compress as compr
 import hash.crc32
 
+// CompressFlags
+pub enum CompressFlags {
```
spytheman, line 10: This can be a flag enum.

vlib/compress/gzip/read_gz_files_test.v
```diff
@@ -39,3 +39,24 @@ fn test_reading_gzip_files_compressed_with_different_options() {
 	assert content9 == content5
 	assert content5 == content1
 }
+
+fn test_compress_with_deferent_level() {
+	compressed := os.read_bytes(s('readme_level_9.gz'))!
+	content9 := gzip.decompress(compressed)!
+
+	// compression: Huffman only=0
+	compress_0 := gzip.compress(content9, flags: int(gzip.CompressFlags.huffman_only))!
```
spytheman, line 48: If `huffman_only` was `pub const huffman_only = 0`, this could become just:
```suggestion
	compress_0 := gzip.compress(content9, flags: gzip.huffman_only)!
```

vlib/compress/gzip/gzip.v
spytheman: The usage sites will require both an int() cast per each flag 🤔 .

vlib/compress/gzip/gzip.v
kbkpbot: `error: when an enum is used as a bit field, you can not assign custom values`

spytheman (approved):
Excellent work.

PR #23442, opened 2025-01-12
db.pg: add support for prepared statements

louis77:
PostgreSQL has two separate calls to support prepared statements. This PR adds the following two public functions to `db.pg`:

```vlang
pub fn (db DB) prepare(name string, query string, num_params int) !
pub fn (db DB) exec_prepared(name string, params []string) ![]Row
```

plus one to handle an empty result to check for an error when `prepare` is called.

Also an appropriate test has been added.


spytheman (approved):
Excellent work.

felipensp:
Nice work!

PR #23450, opened 2025-01-12
ecdsa: fix memleaks

felipensp:
Fix memleak spotted by sanitizers CI.

### DIFF
--- a/vlib/crypto/ecdsa/util.v
+++ b/vlib/crypto/ecdsa/util.v
@@ -93,6 +93,7 @@ pub fn pubkey_from_bytes(bytes []u8) !PublicKey {
 		&& nidgroup != nid_secp256k1 {
 		return error('Unsupported group')
 	}
+	C.EVP_PKEY_free(pub_key)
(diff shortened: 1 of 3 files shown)


vlib/crypto/ecdsa/util.v
```diff
@@ -93,6 +93,7 @@ pub fn pubkey_from_bytes(bytes []u8) !PublicKey {
 		&& nidgroup != nid_secp256k1 {
 		return error('Unsupported group')
 	}
+	C.EVP_PKEY_free(pub_key)
```
spytheman, line 96: imho add it to a defer {} block, right after `mut pub_key := C.EVP_PKEY_new()`, so that it is executed on errors too.

spytheman:
Very good work.

I wonder why did the sanitizer jobs not trigger for the PR itself, and later on master, but they did for the next merge 🤔 .

spytheman:
The CI jobs should be more sensitive now, after 89d405e .

PR #23456, opened 2025-01-13
type_resolver: fix fn detection for comptime arg type

### LINKED ISSUE #23454: cgen error, when calling a generic function in a `$for v in dc.variants { if dc is v {` block
<details>
<summary>V version: V 0.4.9 d2b30df, press to see full `v doctor` output</summary>

|V full version      |V 0.4.9 f23ae9a.d2b30df
|OS                  |linux, Ubuntu 20.04.6 LTS
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-3225 CPU @ 3.30GHz
|Memory              |1.77GB/15.05GB
|                    |
|V executable        |/home/delian/code/v/v
|V last modified time|2025-01-12 16:55:08
|                    |
|V home dir          |OK, value: /home/delian/code/v
|VMODULES            |OK, value: /home/delian/.vmodules
|Current working dir |OK, value: /home/delian/code/v
|                    |
|Git version         |git version 2.47.1
|V git status        |weekly.2025.1-41-g89d405ec
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
|gcc version         |gcc (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
|clang version       |clang version 10.0.0-4ubuntu1 
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.31-0ubuntu9.16) 2.31

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  xx.v && xx`
```v
pub type DesiredCapabilities = FireFox | Edge

struct FireFox {
	browser_name          string = 'firefox'
	accept_insecure_certs bool   = true
	moz_debugger_address  bool   = true
}

struct Edge {
	browser_name string = 'MicrosoftEdge'
}

fn struct_values[T](s T) map[string]string {
	mut res := map[string]string{}
	$if T is $struct {
		$for field in T.fields {
			res[field.name] = s.$(field.name).str()
		}
	}
	return res
}

fn useit(dc DesiredCapabilities) {
	$for v in dc.variants {
		if dc is v {
			$if v is $struct {
				result := struct_values(dc)
				dump(result)
			}
		}
	}
}

fn main() {
	useit(Edge{})	
}
```

**What did you see?**
```
cc: /tmp/v_1000/xx.01JHF178SRTVZ4YE60ZY8B8XZ6.tmp.c:7178: error: cannot convert 'struct main__FireFox' to 'struct main__Edge'
```

a compiled program

Note: the generated C code around the error, looks like this:
```c
        /* variant 0 */ {
            v.typ = 103;
            if ((dc)._typ == 103) {
                #if 1
                {
                    Map_string_string result = main__struct_values_T_main__Edge((*dc._main__FireFox));
                     _v_dump_expr_Map_string_string(_SLIT("xx.v"), 28, _SLIT("result"), result);
                }
                #endif
            }
        }
```

i.e. V chose to use the `main__struct_values_T_main__Edge` variant for the generic function, even though the value was of type `main__FireFox` .
> [!NOTE]
>

felipensp:
Fix #23454

### DIFF
--- a/vlib/v/type_resolver/comptime_resolver.v
+++ b/vlib/v/type_resolver/comptime_resolver.v
@@ -47,7 +47,7 @@ pub fn (t &ResolverInfo) is_comptime(node ast.Expr) bool {
 			}
 		}
 		ast.SelectorExpr {
-			return node.expr is ast.Ident && node.expr.ct_expr
+			return node.expr is ast.Ident && node.expr.ct_expr && node.field_name != 'len'
 		}
 		ast.InfixExpr {
 			return node.left_ct_expr || node.right_ct_expr
--- a/vlib/v/type_resolver/comptime_resolver.v
+++ b/vlib/v/type_resolver/comptime_resolver.v
@@ -55,6 +55,9 @@ pub fn (t &ResolverInfo) is_comptime(node ast.Expr) bool {
 		ast.ParExpr {
 			return t.is_comptime(node.expr)
 		}
+		ast.ComptimeSelector {
+			return true
+		}
 		else {
 			false
 		}
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -861,6 +861,7 @@ pub mut:
 	pos             token.Pos
 	should_be_ptr   bool // fn expects a ptr for this arg
 	// tmp_name        string // for autofree
+	ct_expr bool // true, when the expression is a comptime/generic expression
 }
 
 // function return statement
--- a/vlib/v/checker/assign.v
+++ b/vlib/v/checker/assign.v
@@ -453,8 +453,7 @@ fn (mut c Checker) assign_stmt(mut node ast.AssignStmt) {
 										&& c.table.cur_fn != unsafe { nil }
 										&& c.table.cur_fn.generic_names.len != 0
 										&& !right.comptime_ret_val
-										&& c.type_resolver.is_generic_expr(right)
-										&& right.return_type_generic.has_flag(.generic) {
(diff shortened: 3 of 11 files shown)


vlib/v/type_resolver/comptime_resolver.v
```diff
@@ -47,14 +47,17 @@ pub fn (t &ResolverInfo) is_comptime(node ast.Expr) bool {
 			}
 		}
 		ast.SelectorExpr {
-			return node.expr is ast.Ident && node.expr.ct_expr
+			return node.expr is ast.Ident && node.expr.ct_expr && node.field_name != 'len'
```
spytheman, line 50: why is there an exception for `len` ?

spytheman (approved):
Excellent work.

vlib/v/type_resolver/comptime_resolver.v
felipensp: Usually it coming from a generic param means it is an []T.len one, that do not change its type from int.

spytheman: what will happen if a user defined generic struct has a `len T` field?

felipensp: Probably it will stick on the last resolved type from generic, as usual generic misresolved bugs. I'll try reproduce it asap and provide the fix+test. Thanks.

Issue #22955, opened 2024-11-23
example: One Billion Rows Challenge in V


enghitalo:
### Describe the feature

Inspirations
- https://www.dannyvankooten.com/blog/2024/1brc/
- https://github.com/dannyvankooten/1brc
- https://blog.jacobvosmaer.nl/0020-1-billion-row-challenge/
- https://github.com/jacobvosmaer/1brc

### Use Case

example, CI

### Acknowledgements

- [X] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

latest

### Environment details (OS name and version, etc.)

all
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

JalonSolov:
Here's one of the files converted... I did it back in March...

create-sample.v
```v
import benchmark
import math
import os
import rand

struct City {
        city string
        mean f64
}

const cities = [
        City{'Abha', 18.0},
        City{'Abidjan', 26.0},
        City{'Abéché', 29.4},
        City{'Accra', 26.4},
        City{'Addis Ababa', 16.0},
        City{'Adelaide', 17.3},
        City{'Aden', 29.1},
        City{'Ahvaz', 25.4},
        City{'Albuquerque', 14.0},
        City{'Alexandra', 11.0},
        City{'Alexandria', 20.0},
        City{'Algiers', 18.2},
        City{'Alice Springs', 21.0},
        City{'Almaty', 10.0},
        City{'Amsterdam', 10.2},
        City{'Anadyr', -6.9},
        City{'Anchorage', 2.8},
        City{'Andorra la Vella', 9.8},
        City{'Ankara', 12.0},
        City{'Antananarivo', 17.9},
        City{'Antsiranana', 25.2},
        City{'Arkhangelsk', 1.3},
        City{'Ashgabat', 17.1},
        City{'Asmara', 15.6},
        City{'Assab', 30.5},
        City{'Astana', 3.5},
        City{'Athens', 19.2},
        City{'Atlanta', 17.0},
        City{'Auckland', 15.2},
        City{'Austin', 20.7},
        City{'Baghdad', 22.77},
        City{'Baguio', 19.5},
        City{'Baku', 15.1},
        City{'Baltimore', 13.1},
        City{'Bamako', 27.8},
        City{'Bangkok', 28.6},
[...]

PR #23458, opened 2025-01-13
Example: 1brc

syrmel:
This adds a solution to the 1 Billion Row Challenge as an example to show the potential of V for high-volume processing and to have something to measure against since 1BRC solutions exists in several other languages.


enghitalo:
#22955

spytheman:
mmap is not portable, you will have to edit also `cmd/tools/modules/testing/common.v` around line 280, to skip building the example on platforms where it is not available (windows).

spytheman:
(rebased over current master, to pickup the CI fixes)

examples/1brc/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+```
+cd solution
+v -cc gcc -cflags "-std=c17 -O2 -m64 -march=native -mtune=native -flto" .
+./solution ~/measurements.txt
+```
+
+On Linux, you can time the solution using the `time` command:
+
+`time ./solution ~/measurements.txt`
```
spytheman, line 37: Using `v time ./solution ~/measurements.txt` instead, should work consistently everywhere V works.

examples/1brc/README.md
```diff
@@ -0,0 +1,52 @@
...
+A solution to the [1 Billion Row Challenge](https://www.morling.dev/blog/one-billion-row-challenge/), 
+written in [the V programming language](https://vlang.io/).
+
+Read more about the challenge here: https://www.morling.dev/blog/one-billion-row-challenge/
+
+
+## Running instructions
+
+Suggested compiler options for performance: 
+`v -cc gcc -cflags "-std=c17 -O2 -m64 -march=native -mtune=native -flto"`
```
spytheman, line 12: `-prod` should turn on `-O3 -flto`, so perhaps just this:
```suggestion
`v -cc gcc -prod -cflags "-std=c17 -march=native -mtune=native" .`
```

spytheman: was there a problem without `-fno-inline-small-functions`?

syrmel: I have not encountered it since using mmap instead of the buffered reader.

examples/1brc/README.md
syrmel: Perfect, I updated the README.

examples/1brc/README.md
spytheman: hm, that is good to know ... so the problem is triggered by something in it + gcc's -Os 🤔

spytheman:
There is no need to merge master here manually, unless you add more changes to the PR.

spytheman (approved):
Excellent work.

Issue #23446, opened 2025-01-12
C compile error in option field in a struct


penguindark:
### Describe the bug

Using an option field with map in structs generate C compiler error

### Reproduction Steps

compile:
```v
// module comconf

import os

struct BatchCommonConfig {
	ingestor_db map[string]string
	working_db map[string]string
	remote_db map[string]string
	ready_db ?map[string]string
	mut: client_db ?map[string]string
}

fn main(){

	new_era_db_env := os.getenv_opt('NEW_ERA_DB_ENV') or {'local_container'}
	mut config := match new_era_db_env {
		'local_container' {BatchCommonConfig{

			ingestor_db: {
				'username': 'username_data',
				'password': 'psw_data',
				'hostname': 'postgres_wh',
				'database': 'ingestor_db_test',
				'port': '5432'
        	}

			working_db: {
				'username': 'username_data',
				'password': 'psw_data',
				'hostname': 'postgres_wh',
				'database': 'working_db_test',
				'port': '5432'
			}

			remote_db: {
				'username': 'username_data',
				'password': 'psw_data',
				'hostname': 'postgres_wh',
				'database': 'remote_db_test',
				'port': '5432'
			}
			client_db: {
				'test' : 'test'
			}

		}}
		'local' {BatchCommonConfig{

			ingestor_db: {
				'username': 'postgres',
				'password': 'postgres',
				'hostname': '127.0.0.1',
				'database': 'ingestor_db_test',
				'port': '5432'
        	}

			working_db: {
				'username': 'postgres',
				'password': 'postgres',
				'hostname': '127.0.0.1',
				'database': 'working_db_test',
				'port': '5432'
			}

			remote_db: {
				'username': 'postgres',
				'password': 'postgres',
				'hostname': '127.0.0.1',
				'database': 'remote_db_test',
				'port': '5432'
			}

			ready_db: {
				'username': 'postgres',
				'password': 'postgres',
				'hostname': '127.0.0.1',
				'database': 'acme_db_test',
				'port': '5432'
			}
			
			client_db: {
				'username': 'postgres',
				'password': 'postgres',
				'hostname': '127.0.0.1',
				'database': 'remote_db_test_remote',
				'port': '5432'
			}

		}}
		else {panic('what is the target db?')} 
	}

	b := 'print test ${config.client_db['test'] or {'test'}}' 
	println(b)

}

```

### Expected Behavior

The print of the field or its option

### Current Behavior

```================== C compilation error (from tcc): ==============
[...]

PR #23459, opened 2025-01-13
checker: check if unwrapped `m[key]` if m is `Option`

Delta456:
Fixes https://github.com/vlang/v/issues/23446


vlib/v/checker/checker.v
```diff
@@ -4914,6 +4914,9 @@ fn (mut c Checker) index_expr(mut node ast.IndexExpr) ast.Type {
 		} else if node.left is ast.CallExpr {
 			c.error('type `?${typ_sym.name}` is an Option, it must be unwrapped with `func()?`, or use `func() or {default}`',
 				node.left.pos())
+		} else if node.left is ast.SelectorExpr && node.left.or_block.kind == .absent {
+			c.error('type `?${typ_sym.name}` is an Option, it must be unwrapped first; use `struct.var?[]` to do it',
```
spytheman, line 4918: Can you show the actual expression in the error too?
i.e.:
```
... use `x.m?` to do it
```
instead of the fixed `struct.var?[]` ?

Delta456: Sure

spytheman (approved):
Excellent work.

penguindark:
good job. 😄

PR #23407, opened 2025-01-08
crypto.ecdsa: expand ecdsa module to support another curve.

blackshirt:
There are lack of functionality commonly
needed in the `crypto.ecdsa` module.
Some of them are

- Currently, only support for NIST  P-256 curve, we need more curve on the stock.
- Current `ecdaa.PrivateKey.sign(message)`  does not work on hash (digest) of the message, but its work on message directly.. Its not recommended working on signing messages directly.

This PR  address some of them issue,  by adding and expand support for another popular ecc curve, ie,

- Add support for NIST P-384, ie, `secp384r1` curve
- Add support for NIST P-521, ie, `secp521r1` curve
- Add support for famous `|secp256k1` curve used in bitcoin ecosystem.

This PR also add a routine to sign message with options to use irecommended hash function or custom hash (maybe its can be unified with the current sign routine)

This PR also add a routine to load a public key part from bytes array, commonly parsed from pem-formatted public key file.

This PR was a first time i was working with c wrapping, so, i'm totally noob in the sense, please give its a review.

PR #23463, opened 2025-01-14
crypto.ecdsa: improvement to safety checking, unified signing (and verifying) api to accept options.

blackshirt:
This PR was next iteration from previous PR [23407](https://github.com/vlang/v/pull/23407) thats already merged.
In this current PR, the main changes was integration of signing and verifying API to support options. The current behaviour was not changed with this iteration, This PR also contains some bits of improvement to the `crypto.ecdsa` module, in the form:

- Adds some bits of example to the readme file.
- Improves some documentation to the exposed routine(s).
- Add safety check to the `new_key_from_seed`, and also add some random test to check safety.
- Fix `new_key_from_seed` for handling seed with leading zeros bytes and align with `PrivateKey.seed()`. See [this](https://discord.com/channels/592103645835821068/592114487759470596/1328207871665770599)  bug.
- Unifies signing and verifying api to accepts `SignerOpts`. The current behaviour was not changed.
- Changes of default value of `hash_config` of `SignerOpts` into `.with_no_hash` to align with current behaviour of signing (verifying)
- Add `privkey_from_string` routine to allow loading private key from pem-formatted string.
- Add `pubkey_from_string` routine to allow loading public key from pem-formatted string.
- Split key_free into explicitly `PublicKey.free` and `PrivateKey.free` methods. I have found issues with this current approach.
- Add an example file.
- Some bits of code restructurization.

Please give its deep of review @medvednikov, @spytheman, @JalonSolov, @felipensp and others .
Thank you


blackshirt:
@spytheman : i think the failing CI jobs was similar to the previous PR ..should we remove the additional test ?

vlib/crypto/ecdsa/README.md
```diff
@@ -2,7 +2,27 @@
...
\ No newline at end of file
+- A famous Bitcoin curve, commonly referred as secp256k1
+
+# Example
+```codeblock
+import crypto.ecdsa
+
+fn main() {
+	// create default NIST P-256 secp256r1 curve key pair
+    // If you wish generate another curve, use following manner (or similar): `pbkey, pvkey := ecdsa.generate_key(nid: .secp521r1)!`
```
spytheman, line 17:
```suggestion
    // If you wish to generate another curve, use: `pbkey, pvkey := ecdsa.generate_key(nid: .secp521r1)!`
```

vlib/crypto/ecdsa/README.md
```diff
@@ -2,7 +2,27 @@
...
+	pbkey, pvkey := ecdsa.generate_key()!
+
+	message_tobe_signed := 'Hello ecdsa'.bytes()
+	// create signature with recommended hash
+	signature := pvkey.sign(message_tobe_signed, hash_config: .with_recommended_hash)!
+
+	// verified the message with signature
```
spytheman, line 24:
```suggestion
	// verify the message with the signature
```

vlib/crypto/ecdsa/README.md
```diff
@@ -2,7 +2,27 @@
...
+```codeblock
+import crypto.ecdsa
+
+fn main() {
+	// create default NIST P-256 secp256r1 curve key pair
+    // If you wish generate another curve, use following manner (or similar): `pbkey, pvkey := ecdsa.generate_key(nid: .secp521r1)!`
+	pbkey, pvkey := ecdsa.generate_key()!
+
+	message_tobe_signed := 'Hello ecdsa'.bytes()
+	// create signature with recommended hash
```
spytheman, line 21:
```suggestion
	// create a signature with the recommended hash
```

vlib/crypto/ecdsa/README.md
```diff
@@ -2,7 +2,27 @@
...
+	signature := pvkey.sign(message_tobe_signed, hash_config: .with_recommended_hash)!
```
spytheman, line 22: afaik, `hash_config` already has a default of `.with_recommended_hash`.
Why not just skip it?

spytheman: i.e.:
```suggestion
	signature := pvkey.sign(message_tobe_signed)!
```

vlib/crypto/ecdsa/README.md
```diff
@@ -2,7 +2,27 @@
...
+
 - NIST P-256 curve, commonly referred as prime256v1 or secp256r1
 - NIST P-384 curve, commonly referred as secp384r1 
 - NIST P-521 curve, commonly referred as secp521r1
-- A famous Bitcoin curve, commonly referred as secp256k1
\ No newline at end of file
+- A famous Bitcoin curve, commonly referred as secp256k1
+
+# Example
+```codeblock
```
spytheman, line 12: Instead of `codeblock`, use `v` here.
The benefit is that the tool `v check-md .` will will check the example on the CI,
and that way it will be kept valid, and also consistent with the rest of the V examples in documentation (formatted, vetted, etc).

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -8,6 +8,9 @@ import crypto
 import crypto.sha256
 import crypto.sha512
 
+// See https://docs.openssl.org/master/man7/openssl_user_macros/#description
+#define OPENSSL_API_COMPAT 0x10100000L
```
spytheman, line 12: No. Use instead `#define OPENSSL_API_COMPAT 0x30000000L` if you are going to define it at all.
That is already used in `vlib/net/openssl/openssl.c.v:3:#define OPENSSL_API_COMPAT 0x30000000L` .
If you do not use the same define, then programs that `import net.openssl` and `import crypto.ecdsa` will not be congruent, and that can lead to hard to diagnose problems.

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -74,7 +78,12 @@ pub enum Nid {
 @[params]
 pub struct CurveOptions {
 pub mut:
-	nid Nid = .prime256v1 // default to NIST P-256 curve
+	// default to NIST P-256 curve
+	nid Nid = .prime256v1
+	// by default, using current behavior with allowing arbitrary size of seed bytes as key.
+	// Set into true when you need fixed one using curve key size.
+	// Its mainly purposes for support within `.new_key_from_seed` call.
```
spytheman, line 85:
```suggestion
	// by default, allow arbitrary size of seed bytes as key.
	// Set it to `true`, when you need a fixed size, using the curve key size.
	// Its main purpose is to support the `.new_key_from_seed` call.
```

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -74,7 +78,12 @@ pub enum Nid {
...
+	with_fixed_size bool
```
spytheman, line 86: Just `fixed_size` may read better in the callsites, but I do not have a strong opinion on that.

@JalonSolov, what do you think?

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -95,42 +104,165 @@ struct C.ECDSA_SIG {}
...
+// contains private key and public key parts.
 pub struct PrivateKey {
 	key &C.EC_KEY
+mut:
+	// ks_flag with .flexible value allowing
+	// flexible-size seed bytes as key.
+	// When it .fixed will use underlying key size.
+	ks_flag KeyFlag = .flexible
+	// ks_size stores size of the seed bytes when ks_flag was .flexible.
+	// You should set it to non null value
```
spytheman, line 125: ints do not have null values, this may be better:
```suggestion
	// You should set it to a non zero value
```

spytheman: Why not give it a default of something reasonable?
Is for example 256 good?

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -95,42 +104,165 @@ struct C.ECDSA_SIG {}
...
+// PublicKey represents ECDSA public key for verifying message.
 pub struct PublicKey {
 	key &C.EC_KEY
 }
 
-// Generate a new key pair. If opt was not provided, its default to prime256v1 curve.
-pub fn generate_key(opt CurveOptions) !(PublicKey, PrivateKey) {
+// PrivateKey.new creates a new ECDSA PrivateKey key pair. By default, it would create a prime256v1 curve.
+// You can create another supported key by supplying it with options, for example `.new(nid: .secp256k1)`.
+// If second options, ie, with_fixed_size was not set, it would be discarded and treated as fixed one.
```
spytheman, line 136: I do not understand this 🤔 .
@JalonSolov can it be reformulated better?

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -95,42 +104,165 @@ struct C.ECDSA_SIG {}
...
 	}
 	return pub_key, priv_key
 }
 
-// Create a new private key from a seed. If opt was not provided, its default to prime256v1 curve.
+// new_key_from_seed creates a new private key from the seed bytes. If opt was not provided,
+// its default to prime256v1 curve.
+//
+// Notes on the seed:
+// You should make sure, the seed bytes was comes from cryptographically secure random generators,
```
spytheman, line 223:
```suggestion
// You should make sure, the seed bytes come from a cryptographically secure random generator,
```

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -95,42 +104,165 @@ struct C.ECDSA_SIG {}
...
+		C.EC_KEY_free(ec_key)
 		return error('Failed to create new EC_KEY')
 	}
+	// Retrieve the EC_GROUP object associated with the EC_KEY
+	// Note:
+	// Its cast-ed with voidptr() to workaround the strictness of the type system,
+	// ie, cc backend with `-cstrict` option behaviour. Without this cast,
+	// C.EC_KEY_get0_group expected to return `const EC_GROUP *`,
+	// ie expected to return pointer into constant of EC_GROUP on C parts,
+	// so, its make cgen not happy with this and would fail with error.
```
spytheman, line 245: That comment is not needed and just obscures what is happening. It is related to the language implementation, not to the main content of the module. If comments like that are added to each place that uses voidptr casts, maintaining will become a lot harder
.

Something shorter like `// cast with voidptr() to allow -cstrict checks to pass` is fine.

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -363,52 +548,54 @@ fn (pv PrivateKey) recommended_hash() !crypto.Hash {
...
 	with_custom_hash
 }
 
 @[params]
 pub struct SignerOpts {
 pub mut:
-	hash_config HashConfig = .with_recomended_hash
-	// make sense when HashConfig != with_recomended_hash
+	// default to .with_no_hash to align with the current behaviour of signing.
+	hash_config HashConfig = .with_no_hash
```
spytheman, line 560: That seems counter intuitive to me - if there is a "recommended" hash, it should be the default imho. The previous default also used it, so changing it is a breaking change as well :-| .

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -363,52 +548,54 @@ fn (pv PrivateKey) recommended_hash() !crypto.Hash {
...
+	// make sense when HashConfig != with_recommended_hash
 	allow_smaller_size bool
 	allow_custom_hash  bool
 	// set to non-nil if allow_custom_hash was true
 	custom_hash &hash.Hash = unsafe { nil }
 }
 
-// sign_with_options sign the message with the options. By default, it would precompute
-// hash value from message, with recommended_hash function, and then sign the hash value.
-pub fn (pv PrivateKey) sign_with_options(message []u8, opts SignerOpts) ![]u8 {
```
spytheman, line 384: `sign_with_options` was public, renaming it to a private function is a breaking change.

Is it necessary?

vlib/crypto/ecdsa/util.v
```diff
@@ -3,13 +3,21 @@ module ecdsa
...
 const nid_ec_publickey = C.NID_X9_62_id_ecPublicKey
 
 @[typedef]
 struct C.EVP_PKEY {}
 
+@[typedef]
+struct C.BIO_METHOD {}
+
+@[typedef]
+struct C.BIO {}
```
spytheman, line 19:
```suggestion
pub struct C.BIO {}
```
otherwise it may conflict with the same in `vlib/net/openssl/openssl.c.v:49:pub struct C.BIO {`

vlib/crypto/ecdsa/util.v
```diff
@@ -100,9 +127,10 @@ pub fn pubkey_from_bytes(bytes []u8) !PublicKey {
 	}
 }
 
-// bytes gets the bytes of public key parts of this keypair.
-pub fn (pbk PublicKey) bytes() ![]u8 {
+// seed gets the bytes of public key parts of this key.
+pub fn (pbk PublicKey) seed() ![]u8 {
```
spytheman, line 131: This is a breaking change. Is it needed?

spytheman: If it is going to be renamed, why not name it `public_key_bytes`, instead of `seed()`, and then commenting that it retrieves the public key parts?

vlib/crypto/ecdsa/util.v
```diff
@@ -134,7 +163,175 @@ pub fn (pbk PublicKey) bytes() ![]u8 {
 		return error('bad conversion format')
 	}
 	n := C.EC_POINT_point2oct(group, point, conv_form, buf.data, buf.len, ctx)
-
+	if n == 0 {
+		return error('Fails on EC_POINT_point2oct')
```
spytheman, line 167:
```suggestion
		return error('EC_POINT_point2oct failed')
```

vlib/crypto/ecdsa/util.v
```diff
@@ -134,7 +163,175 @@ pub fn (pbk PublicKey) bytes() ![]u8 {
...
+// const pubkey_sample = '-----BEGIN PUBLIC KEY-----
+// MHYwEAYHKoZIzj0CAQYFK4EEACIDYgAE+P3rhFkT1fXHYbY3CpcBdh6xTC74MQFx
+// cftNVD3zEPVzo//OalIVatY162ksg8uRWBdvFFuHZ9OMVXkbjwWwhcXP7qmI9rOS
+// LR3AGUldy+bBpV2nT306qCIwgUAMeOJP
+// -----END PUBLIC KEY-----'
+//
+// pubkey := ecdsa.pubkey_from_string(pubkey_sample)!
+// verified := pubkey.verify(digest, signature)!
+// assert verified == true
+// ```
```
spytheman, line 189: imho extract such longer snippets into `_test.v` files, and put links to them here.

`_test.v` files *are checked on the CI* on each commit, and thus are guaranteed to be useful.

Commented code examples, especially inside `codeblock` marked blocks, are not checked on the CI, and *will* go stale, as the module changes.

vlib/crypto/ecdsa/util.v
```diff
@@ -134,7 +163,175 @@ pub fn (pbk PublicKey) bytes() ![]u8 {
...
+	C.EVP_PKEY_free(evpkey)
+	C.BIO_free_all(bo)
+	// Its OK to return
+	return PublicKey{
+		key: eckey
+	}
+}
+
+// privkey_from_string loads a PrivateKey from valid PEM-formatted string in s.
+// Underlying wrapper support for old secg and pkcs8 private key format, but this not heavily tested.
```
spytheman, line 246:
```suggestion
// Underlying wrapper support for old secg and pkcs8 private key format, but this was not heavily tested.
```

vlib/crypto/ecdsa/util.v
```diff
@@ -134,7 +163,175 @@ pub fn (pbk PublicKey) bytes() ![]u8 {
...
+// This routine also does not handling for pkcs8 EncryptedPrivateKeyInfo format,
```
spytheman, line 247:
```suggestion
// This routine also does not have handling for the pkcs8 EncryptedPrivateKeyInfo format,
```

vlib/crypto/ecdsa/util.v
```diff
@@ -134,7 +163,175 @@ pub fn (pbk PublicKey) bytes() ![]u8 {
...
+// the callback handler was discarded and not supported currently.
```
spytheman, line 248: I have no idea what callback handler means in this context.
The function accepts just a single string.

vlib/crypto/ecdsa/util_test.v
```diff
@@ -45,7 +46,84 @@ fn test_for_pubkey_bytes() ! {
...
-	assert pbkey.bytes()!.hex() == pb
+	assert pbkey.seed()!.hex() == pb
 	key_free(pbkey.key)
 	key_free(pvkey.key)
 }
+
+// above pem-formatted private key read with
+// `$openssl ec -in vlib/crypto/ecdsa/example.pem -text -param_out -check`
+// produces following result:
+// ```codeblock
```
spytheman, line 57: This is a good use of a codeblock, since it does not contain V code 👍🏻 .

vlib/crypto/ecdsa/README.md
blackshirt: Its changed to `.with_no_hash` to align with the current `.sign()` work

spytheman:
As a side note @blackshirt, I would appreciate it *a lot*, if future changes to this module, are done gradually, in smaller PRs, that contain focused changes.

Those are much easier to make, easier to review, faster to check on the CI (the chances of encountering edge cases and bugs are much smaller), and ultimately faster to merge.

vlib/crypto/ecdsa/ecdsa.v
blackshirt: I have not check it deeply, but the most of low level ec_key method  likely would be deprecated, on recent version of openssl ..
I would check it

vlib/crypto/ecdsa/README.md
blackshirt: Okey .. would be changed

vlib/crypto/ecdsa/ecdsa.v
blackshirt: Okey

vlib/crypto/ecdsa/ecdsa.v
blackshirt: I preferly stick with `with_recommended_hash` so the signing (verifying) work on hash instead of direct messages. i can changed it later

vlib/crypto/ecdsa/ecdsa.v
blackshirt: Its recently addition with previous PR, i think its preferably work with single `sign()` api gone changed now ..

vlib/crypto/ecdsa/util.v
blackshirt: PrivateKey has `.seed()` method, i think `PublicKey.seed()` is more align and consistent with the rest.

vlib/crypto/ecdsa/ecdsa.v
blackshirt: done

vlib/crypto/ecdsa/util.v
```diff
@@ -57,9 +84,9 @@ fn C.EC_GROUP_free(group &C.EC_GROUP)
 // -----END PUBLIC KEY-----'
 //
 // block, _ := pem.decode(pubkey_sample) or { panic(err) }
-// pubkey := pubkey_from_bytes(block.data)!
+// pubkey := ecdsa.pubkey_from_bytes(block.data)!
 // ```
-pub fn pubkey_from_bytes(bytes []u8) !PublicKey {
+fn pubkey_from_bytes(bytes []u8) !PublicKey {
```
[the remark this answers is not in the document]
blackshirt: > That is a breaking change. Is it not needed anymore? The commented example above suggests otherwise. Can you extract the example as a separate _test.v file, that does not start with `module ecdsa`, and thus can only access its public API?
Its gone be replaced with `pubkey_from_string` and the counterpart `privkey_from_string`. Likely the user easily can get PEM-formatted as an input for them.

I would extract the example as separate test file.

vlib/crypto/ecdsa/util.v
blackshirt: > This is a breaking change. Is it needed?
Mostly for consistencies..

vlib/crypto/ecdsa/util.v
blackshirt: was done

vlib/crypto/ecdsa/util.v
blackshirt: Its deep reside underlying wrapper ```EVP_PKEY *PEM_read_bio_PUBKEY(BIO *bp, EVP_PKEY **x, pem_password_cb *cb, void *u)``` where `pem_password_cb ` supposed to be callback handler for encrypted PrivateKey opaque.

blackshirt: And also for `EVP_PKEY *PEM_read_bio_PrivateKey(BIO *bp, EVP_PKEY **x,  pem_password_cb *cb, void *u);`

vlib/crypto/ecdsa/util_test.v
blackshirt: Thanks

blackshirt:
> As a side note @blackshirt, I would appreciate it _a lot_, if future changes to this module, are done gradually, in smaller PRs, that contain focused changes.
> 
> Those are much easier to make, easier to review, faster to check on the CI (the chances of encountering edge cases and bugs are much smaller), and ultimately faster to merge (and less painful to revert, if needed).

Yeah, i think i would be stick with small changes ..Thank you for your suggestion brother..
I think a lot of thing when it first written was using deprecated-recently with the current release of openssl, its need a big rewrite to use new higher-level supported API ..i can do that, but i think its can be done in slowly

vlib/crypto/ecdsa/README.md
blackshirt: Its done

vlib/crypto/ecdsa/README.md
blackshirt: Its done

vlib/crypto/ecdsa/ecdsa.v
blackshirt: Changed to `fixed_size`

vlib/crypto/ecdsa/ecdsa.v
```diff
@@ -95,42 +104,165 @@ struct C.ECDSA_SIG {}
...
+}
+
+// PrivateKey represents ECDSA private key. Actually its a key pair,
+// contains private key and public key parts.
 pub struct PrivateKey {
 	key &C.EC_KEY
+mut:
+	// ks_flag with .flexible value allowing
+	// flexible-size seed bytes as key.
+	// When it .fixed will use underlying key size.
```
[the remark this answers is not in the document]
blackshirt: Changed accordingly

vlib/crypto/ecdsa/ecdsa.v
blackshirt: Changed

vlib/crypto/ecdsa/util.v
blackshirt: I think the last line can be removed ...Its can be misleading

vlib/crypto/ecdsa/util.v
blackshirt: Its gone be changed

vlib/crypto/ecdsa/util.v
spytheman: The private key's seed is documented as:
`// Get the seed (private key bytes)`

I think that `private_key_bytes` is a better name for it too - it will be harder to get the wrong thing, compared to, if you have `key.seed()` for both.

spytheman: Consistencies are important for things that should behave the same. Afaik the whole point of the separation between the public/private key types, is that they should be kept separate.

vlib/crypto/ecdsa/util.v
blackshirt: Done. Its extracted on separated `usage_test.v` file

vlib/crypto/ecdsa/ecdsa.v
blackshirt: Its facing issues until deprecated low level wrapper migrated into higher level recommened in 3.0. 
I dont know the best way to deal with it ...

blackshirt:
@spytheman @medvednikov : i sent changes as suggested review, but some of them can not be applied ...i'm also add a warning to readme about tha facing issues. Feel free to hold this PR if its not good enough to be merged..

vlib/crypto/ecdsa/util.v
blackshirt: I mostly prefer just `.bytes()` to get the bytes. Lets see on the stock. `big.Integer.bytes()`, `rand.PRNG.bytes()`, `http.Response.bytes()`, `rune.bytes()`, `string.bytes()`, `crypto.rand.bytes()`, `rand.bytes()`

vlib/crypto/ecdsa/util.v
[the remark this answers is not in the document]
blackshirt: Likely places in new file also should be added into `cmd/tools/modules/testing/common.v`. Is it enough to add into already publicly facing example test file,  in `example/ecdsa_seed_test.v` file ?

vlib/crypto/ecdsa/ecdsa.v
blackshirt: I think the feasible way for this workaround just not defined it at all here (in this module). Its still work with `openssl` and `ecdsa` together,
but will fail if we use `-cstrict` option on the backend.

vlib/crypto/ecdsa/ecdsa.v
blackshirt: changed

vlib/crypto/ecdsa/util.v
blackshirt: is it usable to just do import from openssl instead `import openssl { C.BIO }` ?

vlib/crypto/ecdsa/README.md
blackshirt: Done

vlib/crypto/ecdsa/util.v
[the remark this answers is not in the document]
spytheman: It can be added in example/ecdsa_seed_test.v, but then it will be less clear about what is it demonstrating, and linking to it specifically from the comments here will be a bit harder. That said, if you prefer it that way - sure.

vlib/crypto/ecdsa/util.v
spytheman: That is a good name as well - `bytes` is a clear noun, and can not be interpreted as a verb too, like `seed` can, so it will show the users that it just retrieves stuff, without modifying things.

vlib/crypto/ecdsa/util.v
```diff
@@ -134,7 +163,175 @@ pub fn (pbk PublicKey) bytes() ![]u8 {
...
+	}
+	mut evpkey := C.EVP_PKEY_new()
+	bo := C.BIO_new(C.BIO_s_mem())
+	if bo == 0 {
+		return error('Failed to create BIO_new')
+	}
+	n := C.BIO_write(bo, s.str, s.len)
+	if n == 0 {
+		// todo: retry the write
+	}
```
spytheman, line 186: that code does nothing - just leave the TODO comment in place of it

vlib/crypto/ecdsa/util.v
```diff
@@ -134,7 +163,175 @@ pub fn (pbk PublicKey) bytes() ![]u8 {
...
```
spytheman, line 246: Same - only the TODO comment is meaningful and with that capitalization as well, i.e. use `TODO` and not `todo`, since it makes it easier, to search for places like this, that need more work.
The check itself is just noise and has to go.

vlib/crypto/ecdsa/util.v
spytheman: As a long term solution - yes, extracting the openssl wrapper in its own module, that is not directly oriented to networking or to cryptography, but exposes both APIs to others, will be best.

For this PR, however no. Just add the `pub`. The PR is already *a lot bigger*, than what is comfortable to process and merge, and I do not want to make it bigger still, and bike-shed every single decision to death :-| .

spytheman:
However I will not close it, since I know how demoralizing it can be to have a big PR closed :-|, so I will just strongly oppose any attempt to expand the scope of it even further, and wait for the remaining issues to be fixed.

vlib/crypto/ecdsa/ecdsa.v
blackshirt: Its getting changed

vlib/crypto/ecdsa/util.v
blackshirt: Okey

vlib/crypto/ecdsa/util.v
blackshirt: Checked

vlib/crypto/ecdsa/util.v
blackshirt: Its gotten with `.bytes()` ..done

vlib/crypto/ecdsa/util.v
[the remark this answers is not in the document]
blackshirt: Okey

blackshirt:
> However I will not close it, since I know how demoralizing it can be to have a big PR closed :-|, so I will just strongly oppose any attempt to expand the scope of it even further, and wait for the remaining issues to be fixed.

I think its the last commit i made. One thing I myself am not satisfied with is with issue related with  ```OPENSSL_API_COMPAT```. Its hard to be done correctly without rewriting most of the deprecated wrappers, so i left as is. 
I think its still compilable and works seamlessly even with `net.openssl` together.
```v
import net.http
import crypto.ecdsa

fn main() {
    resp := http.get('https://url4e.com/unknown')!
    dump(resp.body)

    pvkey := ecdsa.PrivateKey.new()!
    signature := pvkey.sign(resp.body.bytes())!
    
    pubkey := pvkey.public_key()!
    verified := pubkey.verify(resp.body.bytes(), signature)!
    dump(verified)
}
```
works as expected
```bash
$ openssl --version
OpenSSL 3.3.2 3 Sep 2024 (Library: OpenSSL 3.3.2 3 Sep 2024)
gitpod ~ $ v -d use_openssl -cc clang run http.v
[http.v:6] resp.body: <html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx</center>
</body>
</html>

[http.v:13] verified: true
```
The still remaining problems was comes from when build with `-cstrict` option

vlib/crypto/ecdsa/example/example1.v
spytheman, line 1: Having an example is good, but if you make it a test, it will be even better, because it will be checked that it compiles and runs on each commit.

There is no such convention for module examples, and only the programs in the folder `examples/` in the top level, are checked on each commit.

Issue #23461, opened 2025-01-14
C error printing the reference of a not integer variable.


jorgeluismireles:
### Describe the bug

We have a variable other than int, we create a reference and tried to print the reference.

### Reproduction Steps

```v
fn main() {
	s := 'Hello'
	p := &s
	println('${p}')
}
```

If above `s` is `int` or `u8` there is no error and prints a number. If `s` is `rune` or `f32` also there is a C error.

### Expected Behavior

Get the number of the address (like with an integer).

### Current Behavior

```
Running code...
Can't run code. The server returned an error:
/tmp/v_60000/../../../../../../box/code.v:4: warning: implicit declaration of function 'isnil'
/home/admin/v/thirdparty/tcc/lib/libgc.a: error: 'GC_noop1_ptr' defined twice
tcc: error: undefined symbol 'isnil'
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
Exited with error status 1
Please try again.
```

### Possible Solution

Related with this issue https://github.com/vlang/v/issues/23460

### Additional Information/Context

In REPL is harmless:

```
____    ____ 
 \   \  /   /  |  Welcome to the V REPL (for help with V itself, type  exit , then run  v help ).
  \   \/   /   |  Note: the REPL is highly experimental. For best V experience, use a text editor, 
   \      /    |  save your code in a  main.v  file and execute:  v run main.v 
    \    /     |  V 0.4.9 6f7cbd1 . Use  list  to see the accumulated program so far.
     \__/      |  Use Ctrl-C or  exit  to exit, or  help  to see other available commands.

>>> s := 'Hello'
>>> p := &s
>>> println('${p}')
&Hello
>>> 

```

### V version

V 0.4.9 a20fff9

### Environment details (OS name and version, etc.)

https://play.vlang.io/p/04a4c8aab1

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

jorgeluismireles:
Commits linked here solved the C error problem by printing `&` followed by the value of the object referenced. I noticed printing the references with `:x` format we can get the numbers as with integers:
```v
[...]

PR #23464, opened 2025-01-14
checker: fix isnil() markused, use c.table.used_features.auto_str_ptr = true

kbkpbot:
This PR is an improved version of PR #23462
Fix issue #23461


medvednikov:
@felipensp looks good to me. What do you think?

felipensp:
We need a test on vlib/v/tests/skip_unused/, the run.out, .out and .vv file.

spytheman (approved):
Excellent work.

PR #23467, opened 2025-01-14
parallel-cc: several fixes

felipensp:
Building a project with a lot of json and threads, such fixes was needed to build it.

It works for clang-18 and gcc.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -600,9 +600,6 @@ pub fn gen(files []&ast.File, mut table ast.Table, pref_ &pref.Preferences) GenO
 	if g.out_results.len > 0 {
 		b.write_string2('\n// V result_xxx definitions:\n', g.out_results.str())
 	}
-	if g.json_forward_decls.len > 0 {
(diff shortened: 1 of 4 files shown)


felipensp:
I've added PR #23469 for adding herolib CI after this pr gets merged. 👍🏻

vlib/v/gen/c/cgen.v
```diff
@@ -659,10 +647,13 @@ pub fn gen(files []&ast.File, mut table ast.Table, pref_ &pref.Preferences) GenO
 	header = '#ifndef V_HEADER_FILE\n#define V_HEADER_FILE' + header
 	header += '\n#endif\n'
 
-	mut helpers := strings.new_builder(200_000)
+	mut helpers := strings.new_builder(300_000)
```
spytheman, line 650: interesting ... most small programs/tests fit within 200KB 🤔
why 300_000 ?

spytheman (approved):
Excellent work.

vlib/v/gen/c/cgen.v
felipensp: We can adjust again. It increased it because was increased the buffer written to it.

PR #23473, opened 2025-01-14
type_resolver: fix generic selector field type resolving

felipensp:
```v
struct Foo[T] {
	len T
}

fn t[T](val T) string {
	a := val.len + val.len
	println(a)
	return a.str()
}

fn test_main() {
	assert t(Foo[string]{ len: 'hello' }) == 'hellohello'
	assert t(Foo[int]{ len: 123 }) == '246'
	assert t([1, 2, 3]) == '6'
	assert t([1.2, 2.2, 3.3]) == '6'
	assert t(['', '', '']) == '6'
}
```


spytheman (approved):
Excellent work.

Issue #23474, opened 2025-01-15
checker: too many levels of Checker.ensure_generic_type_specify_type_names calls: 41 | In recursive types


koplenov:
<details>
<summary>V version: V 0.4.9 83184c3, press to see full `v doctor` output</summary>

|V full version      |V 0.4.9 4fcd94a.83184c3
|:-------------------|:-------------------
|OS                  |windows, Њ ©Єа®б®дв Windows 10 Pro 19045 64-а §ап¤­ п
|Processor           |28 cpus, 64bit, little endian, Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz  
|Memory              |16.38GB/31.86GB
|                    |
|V executable        |C:\Dev\Progs\v\v.exe
|V last modified time|2025-01-10 16:30:19
|                    |
|V home dir          |OK, value: C:\Dev\Progs\v
|VMODULES            |OK, value: C:\Users\koplenov\.vmodules
|VTMP                |OK, value: C:\Users\koplenov\AppData\Local\Temp\v_0
|Current working dir |OK, value: C:\Dev\Zone\mam_v\treevee\src\ast\repro
|                    |
|Git version         |git version 2.42.0.windows.2
|V git status        |weekly.2025.1-43-g83184c36 (31 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |gcc (MinGW-W64 x86_64-ucrt-posix-seh, built by Brecht Sanders, r4) 13.2.0
|clang version       |clang version 18.1.2
|msvc version        |N/A
|tcc version         |tcc version 0.9.27 (x86_64 Windows)
|tcc git status      |thirdparty-windows-amd64 b425ac82
|emcc version        |N/A
|glibc version       |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  .\main.v && C:\Dev\Zone\mam_v\treevee\src\ast\repro\main.exe`
See attached file `.\main.v`

**What did you see?**
```
[C:\\Dev\\Progs\\v\\vlib\\v\\checker\\checker.v:5235] typ: ast.Type(0x7a = 122)
[C:\\Dev\\Progs\\v\\vlib\\v\\checker\\checker.v:5236] sym.kind: array
[C:\\Dev\\Progs\\v\\vlib\\v\\checker\\checker.v:5237] pos: token.Pos{
    len: 7
    line_nr: 29
    pos: 574
    col: 63
    last_line: 29
[...]

PR #23476, opened 2025-01-15
cgen: fix array decomposing on variadic call

felipensp:
Fix issue found on sample code of issue #23474


spytheman (approved):
Excellent work.

PR #23479, opened 2025-01-15
all: remove ancient deprecations

EmmaTheMartian:
Removes @[deprecated] things older than 180 days.


Delta456:
Run `v fmt -w` on the changed files for formatting

EmmaTheMartian:
Looks like a few things still used the deprecated functions: https://github.com/vlang/v/actions/runs/12800323201/job/35687909996?pr=23479#step:4:10

vlib/math/big/special_array_ops.v
spytheman, line 1: `import math.bits` is no longer used after the removal of the `bits.leading_zeros_32` call in `pub fn bit_length(a Integer) int {`, so the import can be removed too, to silence the warning

spytheman (approved):
Excellent work.

PR #23499, opened 2025-01-17
markused: fix `eprintln(err)` on imported module on short program

felipensp:
Fix build of  v-r2pipe project.


spytheman:
Please add a test too, to prevent future silent regressions.

spytheman (approved):
Excellent work.

Issue #23496, opened 2025-01-17
`int.str()` causes Unhandled page fault when built with `-shared` on Windows


syobocat:
### Describe the bug

Calling `int.str()` causes Unhandled page fault on Windows, when the code is built with `-shared`

### Reproduction Steps

I have created [a repository for demonstration](https://github.com/syobocat/demo), and prepared [sample binaries](https://github.com/syobocat/demo/actions/runs/12827352183) built on GitHub Actions.

`code.v`:
```v
@[export: 'func']
fn func() {
	println('Begin DLL')
	for i in 0 .. 10 {
		println('${i}')
	}
	println('End DLL')
}
```

`loader.c`:
```c
#include <stdio.h>
#include <windows.h>

typedef void(*FUNC)();

int main(void) {
	HMODULE hModule = LoadLibraryW(L"code.dll");
	if (hModule == NULL) {
		DWORD e = GetLastError();
		printf("Failed: %d\r\n", e);
		return 1;
	}
	FUNC func = (FUNC)GetProcAddress(hModule, "func");
	printf("Begin\r\n");
	func();
	printf("End\r\n");
	return 0;
}
```

### Expected Behavior

The following should be printed:
```
Begin
Begin DLL
0
1
2
3
4
5
6
7
8
9
End DLL
End
```

### Current Behavior

On Windows:
```
Begin
Begin DLL
0
```

On Wine:
```
Begin
Begin DLL
0
wine: Unhandled page fault on read access to 0000000000000002 at address 000000001001CCFE (thread 0024), starting debugger...
```

### V version

V 0.4.9 d5aa37d

### Environment details (OS name and version, etc.)

GitHub Actions:

|V full version      |V 0.4.9 d5aa37d
|:-------------------|:-------------------
|OS                  |windows, Microsoft Windows Server 2022 Datacenter 20348 64-bit
|Processor           |4 cpus, 64bit, little endian, AMD EPYC 7763 64-Core Processor  
|Memory              |13.57GB/16GB
|                    |
|V executable        |C:\hostedtoolcache\windows\v\0.4.9\x64\v.exe
|V last modified time|2025-01-17 10:34:01
|                    |
|V home dir          |OK, value: C:\hostedtoolcache\windows\v\0.4.9\x64
|VMODULES            |OK, value: C:\Users\runneradmin\.vmodules
|VTMP                |OK, value: C:\Users\runneradmin\AppData\Local\Temp\v_0
[...]

PR #23507, opened 2025-01-18
cgen,cheaders: move _vinit_caller()/_vcleanup_caller() fn declare forword

kbkpbot:
Fix issue #23496

This PR will make V's dll can be called by other program.
But the v function should call `_vinit_caller()/_vcleanup_caller()` directly....

example:

```v
fn C._vinit_caller()
fn C._vcleanup_caller()

@[export: 'func']
fn func() {
	C._vinit_caller()          // must call first
	println('Begin DLL')
	for i in 0 .. 10 {
		println('${i}')
	}
	println('End DLL')
	C._vcleanup_caller()    // must call last,  note, currently , it is an empty function....
}
```


spytheman:
(the CI failure is unrelated and already fixed on master)

vlib/v/gen/c/cheaders.v
```diff
@@ -536,6 +536,12 @@ typedef int (*qsort_callback_func)(const void*, const void*);
 int load_so(byteptr);
 void _vinit(int ___argc, voidptr ___argv);
 void _vcleanup(void);
+#ifdef _WIN32
+	// workaround for windows, export _vinit_caller/_vcleanup_caller, let dl.open()/dl.close() call it
+	// NOTE: This is hardcoded in vlib/dl/dl_windows.c.v!
+	VV_EXPORTED_SYMBOL void _vinit_caller();
+	VV_EXPORTED_SYMBOL void _vcleanup_caller();
+#endif
```
spytheman, line 544: I fail to see, what the benefit of this is, compared to the previous approach, of only generating those lines conditionally.

kbkpbot: This will make user's v function can call the `_vinit_caller/_vcleanup_caller`, or the compiler will issue error message "implicit declaration of function"

spytheman:
(the CI errors are not related)

PR #23508, opened 2025-01-18
checker: fix missing check for concrete type on match branch expr

### LINKED ISSUE #23506: C compilation error while matching generic type
<details>
<summary>V version: V 0.4.9 1b9f15d, press to see full `v doctor` output</summary>

|V full version      |V 0.4.9 1b9f15d
|OS                  |macos, macOS, 14.4, 23E214
|Processor           |12 cpus, 64bit, little endian, Apple M2 Pro
|                    |
|V last modified time|2025-01-18 06:25:43
|                    |
|Git version         |git version 2.39.3 (Apple Git-145)
|V git status        |1b9f15d6
|.git/config present |true
|                    |
|cc version          |Apple clang version 15.0.0 (clang-1500.1.0.2.5)
|gcc version         |Apple clang version 15.0.0 (clang-1500.1.0.2.5)
|clang version       |Apple clang version 15.0.0 (clang-1500.1.0.2.5)
|emcc version        |N/A
|glibc version       |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  ./src/main.v && ./src/main`
```v
module main

import strconv

fn main() {
	println(to_int('1'))
	println(to_int(1))
}

fn to_int[T](v T) i64 {
	return match typeof(v).name {
		'string' {
			strconv.atoi(v) or { 0 }
		}
		'int' {
			v
		}
		else {
			0
		}
	}
}
```

**What did you see?**
```
cc: /tmp/v_501/main.01JHW86QMJ2K76ZPQNW05P61W8.tmp.c:5804:7: error: assigning to 'int' from incompatible type 'string' (aka 'struct string')
cc:                 _t2 = v;
cc:                     ^ ~
cc: /tmp/v_501/main.01JHW86QMJ2K76ZPQNW05P61W8.tmp.c:5814:33: error: passing 'int' to parameter of incompatible type 'string' (aka 'struct string')
cc: _result_int _t3 = strconv__atoi(v);
cc:                                 ^
cc: /tmp/v_501/main.01JHW86QMJ2K76ZPQNW05P61W8.tmp.c:1906:34: note: passing argument to parameter 's' here
cc: _result_int strconv__atoi(string s) {
cc:                                  ^
cc: 2 errors generated.
builder error: 
==================
```

no errors, as string should never take the int branch and vice versa

**Equivalent Go code**
```go
package main

import (
	"fmt"
	"strconv"
)

func main() {
	fmt.Println(to_int("1"))
	fmt.Println(to_int(1))
}

func to_int[T any](v T) int64 {
	switch v := any(v).(type) {
	case string:
		i, err := strconv.Atoi(v)
		if err != nil {
			return 0
		}
		return int64(i)
	case int:
		return int64(v)
	default:
		return 0
	}
}
```

> [!NOTE]
>

felipensp:
Fix #23506

### DIFF
--- a/vlib/v/checker/tests/match_generic_case_err.out
+++ b/vlib/v/checker/tests/match_generic_case_err.out
@@ -1,4 +1,4 @@
-vlib/v/checker/tests/match_generic_case_err.vv:16:4: error: return type mismatch, it should be `int`
+vlib/v/checker/tests/match_generic_case_err.vv:16:4: error: return type mismatch, it should be `int`, but it is instead `string`
    14 |         }
    15 |         'int' {
    16 |             v
--- a/vlib/v/checker/tests/match_generic_case_err.out
+++ b/vlib/v/checker/tests/match_generic_case_err.out
@@ -1,0 +1,7 @@
+vlib/v/checker/tests/match_generic_case_err.vv:16:4: error: return type mismatch, it should be `int`
+   14 |         }
+   15 |         'int' {
+   16 |             v
+      |             ^
+   17 |         }
+   18 |         else {
--- a/vlib/v/tests/generics/generic_match_expr_test.v
+++ b/vlib/v/tests/generics/generic_match_expr_test.v
@@ -1,0 +1,24 @@
+struct Foo[T] {
+	a T
+}
+
+fn r[T]() Foo[T] {
+	return Foo[T]{}
+}
+
+fn t[T](v T) !Foo[T] {
+	return match typeof(v).name {
+		'string' {
+			r[T]()
+		}
+		else {
+			r[T]()
+		}
+	}
+}
+
+fn test_main() {
+	t(1)!
+	t('')!
+	assert true
+}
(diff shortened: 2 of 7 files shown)


spytheman:
(the CI failure is unrelated and already fixed on master)

vlib/v/checker/tests/match_generic_case_err.out
```diff
@@ -0,0 +1,7 @@
+vlib/v/checker/tests/match_generic_case_err.vv:16:4: error: return type mismatch, it should be `int`
```
spytheman, line 1: That seems a bit unclear 🤔 
Can the error also show the type that the compiler thinks `v` is too?
Something like this can be less puzzling:
```
vlib/v/checker/tests/match_generic_case_err.vv:16:4: error: return type mismatch, it should be `int`, but it is instead `string`
```

spytheman: (not in the same PR, but in general)

spytheman: especially for generic code, I think it will help a lot

vlib/v/tests/generics/generic_match_expr_test.v
spytheman, line 1: This test already passes on master though?
image

felipensp: I added It because looks like It is just covered by vtl.

vlib/v/checker/tests/match_generic_case_err.out
felipensp: I do agree.

spytheman (approved):
Excellent work.

Issue #23486, opened 2025-01-16
option sumtype `is` checking breaks


felipensp:
<details>
<summary>V version: V 0.4.9 d680c42, press to see full `v doctor` output</summary>

|V full version      |V 0.4.9 ac71f9e.d680c42
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.1 LTS (WSL 2)
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8365U CPU @ 1.60GHz
|Memory              |5.54GB/7.67GB
|                    |
|V executable        |/home/felipe/v/v
|V last modified time|2025-01-16 11:36:32
|                    |
|V home dir          |OK, value: /home/felipe/v
|VMODULES            |OK, value: /home/felipe/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/felipe/v
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.1-62-gd49b8f9f (2 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 0134e9b9
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.3) 2.39

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  bug.v && /home/felipe/v/bug`
```v
type Foo = int | string
struct Struct {
    a ?Foo
}

fn r(a string){}

fn t(s ?Foo) {
    mut t := Struct{a: s}
    if t.a is string {
        r(t.a)
    }
}
    
fn main() {
    s := ?Foo('hello')
    t(s)
}
```

**What did you see?**
```
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/bug.01JHQFSW2KWAQ1HHR895159XBV.tmp.c:5641: error: field not found: _typ
=================================================================
(You can pass `-cg`, or `-show-c-output` as well, to print all the C error messages).
builder error: 
==================
[...]

PR #23510, opened 2025-01-18
checker: disallow `expr is Type` if expr is Optional

Delta456:
Fixes https://github.com/vlang/v/issues/23486


vlib/v/checker/tests/orm_op_with_option_and_none.out
```diff
@@ -47,3 +47,17 @@ vlib/v/checker/tests/orm_op_with_option_and_none.vv:45:30: warning: comparison w
...
+      |                               ~~~~
+   50 |     }!
+   51 |
+vlib/v/checker/tests/orm_op_with_option_and_none.vv:53:25: error: name is an Optional, it needs to be unwrapped first
+   51 | 
+   52 |     _ := sql db {
+   53 |         select from Foo where name !is none
+      |                               ~~~~
+   54 |     }!
+   55 | }
```
Delta456, line 63: I don't think this should be included as an error because these are inside the ORM expr. Still not sure :thinking:

Delta456: Ok this shouldn't error

spytheman (approved):
Excellent work.

PR #23532, opened 2025-01-20
os: add split_path()

kbkpbot:
This will help user get the filename without ext.

`split_path` will split `opath` into (`dir`,`filename`,`ext`).
Examples:

```v
dir,filename,ext := os.split_path('/usr/lib/test.so')
assert [dir,filename,ext] == ['/usr/lib','test','.so']
```

Bug fix:
`os.base()`, `os.file_name()`, `os.dir()` all support both windows/non-windows path at the same time.

### DIFF
--- a/vlib/os/os.v
+++ b/vlib/os/os.v
@@ -288,8 +288,8 @@ pub fn dir(opath string) string {
 	other_separator := if path_separator == '/' { '\\' } else { '/' }
 	path := opath.replace(other_separator, path_separator)
 	pos := path.last_index(path_separator) or { return '.' }
-	if pos == 0 && path_separator == '/' {
-		return '/'
+	if pos == 0 {
+		return path_separator
 	}
 	return path[..pos]
 }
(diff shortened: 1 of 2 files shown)


vlib/os/os_test.c.v
```diff
@@ -684,6 +685,76 @@ fn test_file_name() {
...
+		dir, filename, ext = os.split_path('/lib/x.c.v/')
+		assert [dir, filename, ext] == ['/lib/x.c.v', '', '']
+
+		dir, filename, ext = os.split_path('/lib/../x.c.v/')
+		assert [dir, filename, ext] == ['/lib/../x.c.v', '', '']
+
+		dir, filename, ext = os.split_path('/lib/x.c.')
+		assert [dir, filename, ext] == ['/lib', 'x.c.', '']
+	}
+}
```
spytheman, line 756: imho extract the common cases outside the platform check

spytheman: also put a small section (1-2 cases) inside the platform check, that tests what happens if the function gets a path with the opposite separator (i.e. / on windows and \ on !windows)

vlib/os/os.v
spytheman, line 1: imho move the new function to `vlib/os/os.c.v`, to workaround the JS failure
(the JS backend is not complete, and does not support all V constructs, that the C one does)

spytheman (approved):
Excellent work.

PR #23534, opened 2025-01-20
vet: notice potentially fn to be inlined

felipensp:
Using `-F`

image

### DIFF
--- a/cmd/tools/vvet/analyze.v
+++ b/cmd/tools/vvet/analyze.v
@@ -86,14 +86,20 @@ fn (mut vt VetAnalyze) expr(vet &Vet, expr ast.Expr) {
 		}
 		ast.CallExpr {
 			if expr.is_static_method || expr.is_method {
-				vt.save_expr(callexpr_cutoff, '${expr.left}.${expr.name}(${expr.args.map(it.str()).join(', ')})',
+				left_str := expr.left.str()
+				lock vt.call_counter {
+					if vt.cur_fn.receiver.name == left_str {
+						vt.call_counter['${int(vt.cur_fn.receiver.typ)}.${expr.name}']++
+					}
+				}
+				vt.save_expr(callexpr_cutoff, '${left_str}.${expr.name}(${expr.args.map(it.str()).join(', ')})',
 					vet.file, expr.pos)
 			} else {
+				lock vt.call_counter {
+					vt.call_counter[expr.name]++
+				}
 				vt.save_expr(callexpr_cutoff, '${expr.name}(${expr.args.map(it.str()).join(', ')})',
 					vet.file, expr.pos)
-			}
-			lock vt.call_counter {
-				vt.call_counter[expr.name]++
 			}
 		}
 		ast.AsCast {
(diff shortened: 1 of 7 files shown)


cmd/tools/vvet/analyze.v
```diff
@@ -7,26 +7,32 @@ import v.token
...
+const selectorexpr_cutoff = $d('VET_SELECTOREXPR_CUTOFF', 10)
+const callexpr_cutoff = $d('VET_CALLEXPR_CUTOFF', 10)
+const stringinterliteral_cutoff = $d('STRINGINTERLITERAL_CUTOFF', 10)
+const stringliteral_cutoff = $d('STRINGLITERAL_CUTOFF', 10)
+const ascast_cutoff = $d('ASCAST_CUTOFF', 10)
+const stringconcat_cutoff = $d('STRINGCONCAT_CUTOFF', 10)
+
+// possibly inline fn cutoff
+const fns_call_cutoff = $d('VET_FNS_CALL_CUTOFF', 10) // at least N calls
+const short_fns_cutoff = $d('VET_SHORT_FNS_CUTOFF', 3) // lines
```
spytheman, line 21: Use instead:
`const indexexpr_cutoff := os.getenv_opt('VET_INDEXEXPR_CUTOFF') or { '10' }.int()`

The reason that os.getenv is better, is that with $d() those will be fixed once `v vet` is compiled, while with getenv_opt, they can be changed more conveniently at startup/runtime in most shells, with just: `VET_INDEXEXPR_CUTOFF=99 v vet .` for example.

cmd/tools/vvet/analyze.v
```diff
@@ -7,26 +7,32 @@ import v.token
...
 // long functions cutoff
-const long_fns_cutoff = 300
+const long_fns_cutoff = $d('VET_LONG_FNS_CUTOFF', 300)
 
 struct VetAnalyze {
 mut:
-	repeated_expr_cutoff shared map[string]int // repeated code cutoff	
-	repeated_expr        shared map[string]map[string]map[string][]token.Pos // repeated exprs in fn scope
-	cur_fn               ast.FnDecl // current fn declaration
+	repeated_expr_cutoff  shared map[string]i64 // repeated code cutoff
```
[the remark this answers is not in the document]
spytheman: is there a case, where it will be reached for human code?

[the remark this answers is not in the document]
felipensp: I just changed it because $d() was asking for i64.

cmd/tools/vvet/analyze.v
felipensp: Got it, thanks for the tips.

vlib/v/help/common/vet.txt
```diff
@@ -12,7 +12,8 @@ Options:
 
   -v, -verbose        Enable verbose logging.
 
-  -F                  Report empty and long function declaration (>300 lines).
+  -F                  Report potential function to be inlined, empty, long function
+                      declaration (>300 lines).
```
spytheman, line 16: Why not 2 separate option flags, instead of reusing the same option for both?

felipensp: Done.

spytheman:
Excellent work.

PR #23535, opened 2025-01-20
util, parser: inline some fns

felipensp:
image


spytheman (approved):
Excellent finds.

felipensp:
> Excellent finds.

vvet founds it 💯

PR #23536, opened 2025-01-20
v: cache repeated `node.args.len`, `method.generic_names.len`, `node.concrete_types.len` and `node.op.str()`

felipensp:
Minor optimization.


spytheman:
Nice. What is the performance impact?

PR #23538, opened 2025-01-20
markused: cleanup as_cast handling

felipensp:
Cleanup


spytheman:
As casting works fine.
I do not see what should be cleaned up about it.

felipensp:
> As casting works fine. I do not see what should be cleaned up about it.

All this logic with map to just ensure that new_array... must be created is not needed.

spytheman:
It is - see the failing CI.
Besides, all that logic is triggered only for ast.AsCast expressions, which are relatively rare.

felipensp:
> It is - see the failing CI. Besides, all that logic is triggered only for ast.AsCast expressions, which are relatively rare.

Just need a fix.

spytheman (approved):
Excellent work.

PR #23553, opened 2025-01-22
cgen: fix auto str which expects ptr for ptr type

felipensp:


spytheman (approved):
Very good work, but please add a test too, to prevent regressions.

Something like this works (with caveats):
```v
module main

import net.html

fn main() {
	mut doc := html.parse('<body><div class="Box-footer"><div class="truncate">abc</div></div></body>')
	footer := doc.get_tags_by_class_name('Box-footer')[0]
	hrefs := footer.get_tag_by_class_name('Truncate')
	//dump(hrefs.str()) // cgen error
	println(hrefs) // ok
	// println(hrefs.str()) // runtime error, invalid memory access
	res := '${hrefs}'
	assert res == '&Option(&nil)'
}
```

spytheman:
What is interesting is that `println(hrefs.str())` works, when `<div class="truncate">` is changed to `<div class="Truncate">`, because then the output of the string method changes.

felipensp:
> Very good work, but please add a test too, to prevent regressions.
> 
> Something like this works (with caveats):
> 
> ```v
> module main
> 
> import net.html
> 
> fn main() {
> 	mut doc := html.parse('<body><div class="Box-footer"><div class="truncate">abc</div></div></body>')
> 	footer := doc.get_tags_by_class_name('Box-footer')[0]
> 	hrefs := footer.get_tag_by_class_name('Truncate')
> 	//dump(hrefs.str()) // cgen error
> 	println(hrefs) // ok
> 	// println(hrefs.str()) // runtime error, invalid memory access
> 	res := '${hrefs}'
> 	assert res == '&Option(&nil)'
> }
> ```

Test added.

PR #23564, opened 2025-01-23
doc: document keyword use struct fields

mike-ward:
Document keyword use struct fields with example

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -2408,6 +2408,21 @@ p = Point{10, 20}
 assert p.x == 10
 ```
 
+Struct fields can re-use reserved keywords:
+
+```v
+struct Employee {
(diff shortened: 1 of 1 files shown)


spytheman (approved):
Thanks.

doc/docs.md
```diff
@@ -2424,6 +2424,21 @@ p := &Point{10, 10}
...
+	name string
+}
+
+employee := Employee{
+	type: 'FTE'
+	name: 'John Doe'
+}
+println(employee.type)
+```
+
```
spytheman, line 2441: The description is very good, but I think that it should be moved at line 2410, before the `### Heap structs` heading, since it applies to all structs, not just to the heap tagged ones.

Issue #23581, opened 2025-01-25
vlib/runtime: free_memory function not implemented on FreeBSD


lcheylus:
`free_memory` function is defined in `vlib/runtime/runtime_nix.c.v` and implementations are available in `vlib/runtime/free_memory_impl_*.c.v`.

For now, `free_memory` implementation is only available for:

- **Linux**: `vlib/runtime/free_memory_impl_linux.c.v`
- **Darwin**: `vlib/runtime/free_memory_impl_darwin.c.v`

Otherwise `free_memory_impl` function returns 1 (see `vlib/runtime/free_memory_impl_default.c.v`).

**`free_memory` function is not available on FreeBSD** and must be implemented in `vlib/runtime/free_memory_impl_freebsd.c.v`

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #23594, opened 2025-01-26
runtime: add freebsd free_memory() support

kbkpbot:
Fix issue #23581

### DIFF
--- a/vlib/builtin/cfns.c.v
+++ b/vlib/builtin/cfns.c.v
@@ -232,7 +232,6 @@ fn C.isatty(fd int) int
 fn C.syscall(number int, va ...voidptr) int
 
 fn C.sysctl(name &int, namelen u32, oldp voidptr, oldlenp voidptr, newp voidptr, newlen usize) int
-fn C.sysctlnametomib(name charptr, mib &int, len &usize) int
 
 @[trusted]
 fn C._fileno(int) int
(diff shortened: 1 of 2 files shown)


medvednikov:
```
FAIL  [45/69] C:   491.4 ms, R:     2.697 ms vlib/runtime/runtime_test.v
total memory: 8552550400
free  memory: 0
/tmp/cirrus-ci-build/vlib/runtime/runtime_test.v:8: fn test_physical_memory
   > assert total > 0 && free > 0
     Left value (len: 4): `true`
    Right value (len: 5): `false`

     nr cpus: 2
     nr jobs: 1
    is_32bit: false
    is_64bit: true
       is_le: true
       is_be: false
```

vlib/builtin/cfns.c.v
```diff
@@ -232,6 +232,7 @@ fn C.isatty(fd int) int
 fn C.syscall(number int, va ...voidptr) int
 
 fn C.sysctl(name &int, namelen u32, oldp voidptr, oldlenp voidptr, newp voidptr, newlen usize) int
+fn C.sysctlnametomib(name charptr, mib &int, len &usize) int
```
spytheman, line 235: Why put it here - it is only available on BSDs afaik (unlike C.sysctl)?

I think that `vlib/runtime/free_memory_impl_freebsd.c.v` will be a better place for that declaration.

spytheman (approved):
Excellent work.

PR #23610, opened 2025-01-29
doc: rework strings section to clarify

JalonSolov:
Rearranged `Strings` section a small amount, clarified some descriptions, added a small amount of detail to some examples.

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -577,14 +577,28 @@ d := b + x     // d is of type `f64` - automatic promotion of `x`'s value
 
 ### Strings
 
-```v nofmt
+In V, strings are encoded in UTF-8, and are immutable (read-only) by default:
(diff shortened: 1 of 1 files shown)


doc/docs.md
```diff
@@ -577,14 +577,28 @@ d := b + x     // d is of type `f64` - automatic promotion of `x`'s value
 
 ### Strings
 
-```v nofmt
+In V, strings are encoded in UTF-8, and are immutable (read-only) by default:
+
+```v
+s := 'hello 🌎' // emoji takes 4 bytes, and string length is reported in bytes
```
spytheman, line 583:
```suggestion
s := 'hello 🌎' // this emoji takes 4 bytes, and string length is reported in bytes
```

PR #23612, opened 2025-01-30
crypto.ecdsa: fix bug when signing with `.with_no_hash` options.

blackshirt:
This PR contains small fix bug when signing message with .`with_no_hash` options. The bug more clearly self-explained in this snippet

```v
import crypto.ecdsa

fn main() {
        // default prime256v1 key was 32 bytes length
	pv := ecdsa.PrivateKey.new()!
        // creates message with length over key size
	msg_a := 'a'.repeat(34).bytes()
	opt := ecdsa.SignerOpts{
		hash_config: .with_no_hash
	}
	signed := pv.sign(msg_a, opt)!
	pb := pv.public_key()!
	st := pb.verify(msg_a, signed, opt)!
        // this should be true
	dump(st) 

	// different msg should not be verified under the same signature.
	msg_b := 'a'.repeat(392).bytes()
	ds := pb.verify(msg_b, signed, opt)!

	// should false
	dump(ds)
}
```

but the output was

```
[code.v:14] st: true
[code.v:19] ds: true
```

Its changed `.with_no_hash` to have the same behaviour with `.with_recommended_hash`.  Its can be deprecated on next cycled.

Thanks

### DIFF
--- a/vlib/crypto/ecdsa/ecdsa.v
+++ b/vlib/crypto/ecdsa/ecdsa.v
@@ -510,36 +510,47 @@ pub mut:
 	custom_hash &hash.Hash = unsafe { nil }
 }
 
+fn calc_digest_with_recommended_hash(key &C.EC_KEY, msg []u8) ![]u8 {
+	h := recommended_hash(key)!
+	match h {
+		.sha256 {
+			return sha256.sum256(msg)
+		}
+		.sha384 {
+			return sha512.sum384(msg)
+		}
+		.sha512 {
+			return sha512.sum512(msg)
+		}
+		else {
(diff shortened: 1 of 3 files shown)


spytheman:
> @spytheman : why the old commit [23612](https://github.com/vlang/v/pull/23612/commits/56479a838cf5d87af907d63d5337de485f9c03d1) thats already merged still included in this queue of PR ?

Most likely because you continued your development over the same branch, or on a new branch, based on the old.

You can try doing: `git pull --rebase origin master`, then `git push -f`, while being on your `ext_ecdsa` branch locally, to fix it.

PR #23622, opened 2025-01-31
potential fix to the bug im having

ciaranmcguinness:

### DIFF
--- a/vlib/net/http/response.v
+++ b/vlib/net/http/response.v
@@ -35,7 +35,21 @@ pub fn (resp Response) bytestr() string {
 
 // Parse a raw HTTP response into a Response object
 pub fn parse_response(resp string) !Response {
-	version, status_code, status_msg := parse_status_line(resp.all_before('\r\n'))!
+	sline := resp.all_before('\r\n')
+	mut version := "" 
+	mut status_code := ""
+	mut status_msg := ""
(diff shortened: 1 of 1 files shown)


ciaranmcguinness:
what? i hate github

felipensp:
> what? i hate github

I have reopened it.

spytheman:
Can you please describe in more details what problem this PR solves, and potentially provide a regression test, that fails on master, but passes here?

vlib/net/http/response.v
```diff
@@ -35,7 +35,21 @@ pub fn (resp Response) bytestr() string {
 
 // Parse a raw HTTP response into a Response object
 pub fn parse_response(resp string) !Response {
-	version, status_code, status_msg := parse_status_line(resp.all_before('\r\n'))!
+	sline := resp.all_before('\r\n')
+	mut version := "" 
+	mut status_code := ""
+	mut status_msg := ""
+	//this may be a hack, but it means i can get free completions from ai.hackclub.com
+	//fix sites that send malformed responses with leading newlines
```
enghitalo, line 43: Well... I think we must avoid handling things out of documented standards.

enghitalo: Returning an error would be more interesting to respond to a bad request error. IMHO

ciaranmcguinness:
this was supposed to be a pr on my own repo, this is not at all ready for prod

Issue #23412, opened 2025-01-08
Single selective import will produce a warning "module is imported but never used"


khalyomede:
### Describe the bug

When selectively importing a single statement inside of a module, a warning is raised.

### Reproduction Steps

1. Create a folder "http" and a file "http/Mime.v" with this content:

```v
module http

enum Mime {
  text_html
}
```

2. Create a "main.v" file and have this content:

```v
module main

import http { Mime }

fn main() {
  mime := Mime.text_html.str()

  println(mime)
}

```

Interpret or compile the main file:

```v
> v run main.v
main.v:3:8: warning: module 'http' is imported but never used
    1 | module main
    2 | 
    3 | import http { Mime }
      |        ~~~~
    4 | 
    5 | fn main() {
text_html
```

### Expected Behavior

No warnings when selectively importing one module.

### Current Behavior

It produce a warning when compiing or interpreting the code:

```bash
> v main.v
main.v:3:8: warning: module 'http' is imported but never used
    1 | module main
    2 | 
    3 | import http { Mime }
      |        ~~~~
    4 | 
    5 | fn main() {
```

### Possible Solution

In the mean time I just import the module without selective import, and access my statement (see below).

### Additional Information/Context

Simple module import produces no warnings when used in place of a single selective import. This code below will produce no warnings:

```v
module main

import http

fn main() {
  mime := http.Mime.text_html.str()

  println(mime)
}
```

### V version

0.4.9

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.9 9fc8352
|:-------------------|:-------------------
|OS                  |linux, N/A
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz
|Memory              |6.8GB/15.43GB
|                    |
|V executable        |/opt/v/v
|V last modified time|2025-01-08 03:26:40
|                    |
|V home dir          |OK, value: /opt/v
|VMODULES            |OK, value: /root/.vmodules
|VTMP                |OK, value: /tmp/v_0
|Current working dir |OK, value: /home/v
|                    |
|env VFLAGS          |"-cc gcc"
|                    |
|Git version         |git version 2.30.6
|V git status        |9fc83526 (3 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |cc (Alpine 10.2.1_pre1) 10.2.1 20201203
|gcc version         |gcc (Alpine 10.2.1_pre1) 10.2.1 20201203
[...]

PR #23626, opened 2025-02-01
`parser`: do not warn on imported sym as long as it is used 

Delta456:
Fixes https://github.com/vlang/v/issues/23412


spytheman (approved):
Excellent work.

PR #23634, opened 2025-02-02
os: add disk_usage()

kbkpbot:
disk_usage(path) returns disk usage of `path`
Examples:

```v
pub struct DiskUsage {
pub:
        total     u64
        available u64
        used      u64
}

usage := os.disk_usage('.')!
assert usage.total > 0
assert usage.available > 0
assert usage.used > 0
```

### DIFF
--- a/vlib/os/os_nix.c.v
+++ b/vlib/os/os_nix.c.v
@@ -547,7 +547,7 @@ pub fn disk_usage(path string) !DiskUsage {
 	mut vfs := C.statvfs{}
 	ret := unsafe { C.statvfs(&char(mpath.str), &vfs) }
 	if ret == -1 {
-		return error('can\`t get disk usage of path')
+		return error('cannot get disk usage of path')
 	}
 	f_bsize := u64(vfs.f_bsize)
 	f_blocks := u64(vfs.f_blocks)
--- a/vlib/os/os_nix.c.v
+++ b/vlib/os/os_nix.c.v
@@ -7,6 +7,7 @@ import strings
 #include <fcntl.h>
 #include <sys/utsname.h>
 #include <sys/types.h>
+#include <sys/statvfs.h>
 #include <utime.h>
 
 // path_separator is the platform specific separator string, used between the folders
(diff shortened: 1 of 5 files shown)


vlib/os/os_nix.c.v
```diff
@@ -530,3 +531,33 @@ fn C.sysconf(name int) i64
...
+
+fn C.statvfs(path charptr, vfs &C.statvfs) int
+
+// disk_usage returns disk usage of `path`
+// Examples:
+// ```v
+// total,available,used := os.disk_usage('.')
+// ```
+@[manualfree]
+pub fn disk_usage(path string) !(u64, u64, u64) {
```
spytheman, line 550: Please make a public named struct, with `total`, `available` and `used` fields, and return an instance of it. 

Returning a tuple of 3 equally sized u64s is error prone - the chances of them being confused in the caller is much higher, and sometimes you may only want to:
 `if os.disk_usage('.').available > 10_000_000 {`, which is harder to do with a tuple.

spytheman (approved):
Excellent work.

vlib/os/os_nix.c.v
```diff
@@ -530,3 +531,31 @@ fn C.sysconf(name int) i64
...
+
+// disk_usage returns disk usage of `path`
+@[manualfree]
+pub fn disk_usage(path string) !DiskUsage {
+	mpath := if path == '' { '.' } else { path }
+	defer { unsafe { mpath.free() } }
+	mut vfs := C.statvfs{}
+	ret := unsafe { C.statvfs(&char(mpath.str), &vfs) }
+	if ret == -1 {
+		return error('can\`t get disk usage of path')
```
JalonSolov, line 550: Slightly cleaner... no escapes needed.
```suggestion
		return error("can't get disk usage of path")
```

vlib/os/os_windows.c.v
```diff
@@ -601,3 +601,26 @@ pub fn page_size() int {
...
+	mut ret := false
+	if path == '.' || path == '' {
+		ret = C.GetDiskFreeSpaceExA(&char(0), &free_bytes_available_to_caller, &total,
+			&available)
+	} else {
+		ret = C.GetDiskFreeSpaceExA(&char(path.str), &free_bytes_available_to_caller,
+			&total, &available)
+	}
+	if ret == false {
+		return error('can\`t get disk usage of path')
```
JalonSolov, line 619:
```suggestion
		return error("can't get disk usage of path")
```

vlib/os/os_nix.c.v
JalonSolov: Alternative could be
```suggestion
		return error('cannot get disk usage of path')
```

vlib/os/os_windows.c.v
JalonSolov: Or alternative
```suggestion
		return error('cannot get disk usage of path')
```

PR #23636, opened 2025-02-02
tools: add vreduce

Eliyaan:
Adds the `v reduce` tool that helps doing minimal reproducible examples.
The tool works: `v reduce [options]` and outputs the rpdc.v file in the current directory.

I dont know if I missed something I should have done.

### DIFF
--- a/cmd/tools/vreduce.v
+++ b/cmd/tools/vreduce.v
@@ -33,24 +33,23 @@ fn main() {
 		exit(1)
 	}
 
-	println("Starting to reduce the file '${file_path}' with command `${command}` reproducing `${error_msg}`")
+	println("Starting to reduce the file: '${file_path}'\n    with command: `${command}`,\n    trying to reproduce: `${error_msg}`")
 
 	if do_fmt {
-		println('Will do v fmt to the output rpdc.v file')
+		println('Will do `v fmt -w rpdc.v` after the reduction.')
 	} else {
-		println('Will not do v fmt to the output rpdc.v file (use the `--fmt` or `-w` flag to enable it)')
+		println('Will NOT do `v fmt -w rpdc.v` (use the `--fmt` or `-w` flag to enable it)')
 	}
 
-	// Opening the file
-	file := os.read_file(file_path)!
-	assert string_reproduces(file, error_msg, command)
-	println('Original code size: ${file.len}')
-	mut tree := parse(file)
+	content := os.read_file(file_path)!
+	assert string_reproduces(content, error_msg, command)
+	show_code_stats(content, label: 'Original code size')
+	mut tree := parse(content)
 
 	// start tests
 	tmp_code := create_code(tree)
 	assert string_reproduces(tmp_code, error_msg, command)
-	println('Code size without comments: ${tmp_code.len}')
+	show_code_stats(tmp_code, label: 'Code size without comments')
 
 	// reduce the code
 	reduce_scope(mut tree, error_msg, command, do_fmt)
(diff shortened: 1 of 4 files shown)


cmd/tools/vreduce.v
```diff
@@ -0,0 +1,286 @@
...
+	fp.skip_executable()
+	fp.application('v reduce')
+	fp.description('This tool will reduce the code file and try to make the smallest one it can that reproduces the error when the command is executed')
+	fp.version('')
+
+	file_path := fp.string('file_path', `f`, '', 'the path of the file you want to reduce')
+	error_msg := fp.string('error_msg', `e`, default_error_msg, 'the error message you want to reproduce, default: \'${default_error_msg}\'')
+	command := fp.string('command', `c`, default_command, 'the command used to try to reproduce the error, default: \'${default_command}\'')
+	do_fmt := fp.bool('fmt', `w`, false, 'enable v fmt for the output (rpdc.v)')
+	_ := fp.finalize() or {
```
spytheman, line 23:
```suggestion
	file_paths := fp.finalize() or {
```
then you can `v reduce file.v`, without needing `-f`
asserting file_paths.len == 1 may be a good idea too

spytheman: or checking it and doing `println(fp.usage())` if the check fails

cmd/tools/vreduce.v
```diff
@@ -0,0 +1,286 @@
...
+const tmp_folder = '.tmp_v_reduce'
+const tmp_reduced_code_file_name = '__v_reduced_code.v'
+const path = '${tmp_folder}/${tmp_reduced_code_file_name}'
+
+fn main() {
+	mut fp := flag.new_flag_parser(os.args)
+	fp.skip_executable()
+	fp.application('v reduce')
+	fp.description('This tool will reduce the code file and try to make the smallest one it can that reproduces the error when the command is executed')
+	fp.version('')
```
spytheman, line 17:
```suggestion
	fp.version('0.0.1')
```

cmd/tools/vreduce.v
```diff
@@ -0,0 +1,286 @@
+import os
+import flag
+import math
+
+const default_command = 'v -no-skip-unused' // Command used to compile the program, using -no-skip-unused to ease the reducing
+const default_error_msg = 'C error found' // the pattern to reproduce
+// Temporary files
+const tmp_folder = '.tmp_v_reduce'
```
spytheman, line 8:
```suggestion
const tmp_folder = os.join_path(os.vtmp_dir(), 'vreduce')
```

spytheman: (the benefit of that is that you will not need to add an exception for the `.tmp_v_reduce` folder in your .gitignore file)

cmd/tools/vreduce.v
```diff
@@ -0,0 +1,286 @@
+import os
+import flag
+import math
+
+const default_command = 'v -no-skip-unused' // Command used to compile the program, using -no-skip-unused to ease the reducing
```
spytheman, line 5:
```suggestion
const default_command = '${os.quoted_path(@VEXE)} -no-skip-unused' // Command used to compile the program, using -no-skip-unused to ease the reducing
```

spytheman: In this way, it will work even for v setups, where there is no global `v` (i.e. from a .zip file, where the user has not run `v symlink`)

spytheman: I am not sure that passing `-no-skip-unused` is a good idea - the default is currently the opposite (i.e. as if -skip-unused is passed).

Most people will have bugs show up, for projects where the default is used, not for cases, where they have passed `-no-skip-unused` .

Eliyaan: If `-no-skip-unused` is not used, vreduce will not be able to reduce as efficiently functions that are calling the function causing trouble. If it tries to remove the body of the function where the call to the function causing trouble is located it will not show the error as the function will get pruned before cgen.

spytheman:
Excellent work @Eliyaan .

I've tried it on one of the latest issues, and it cut the source in half 🥳:
image

spytheman (approved):

felipensp:
Nice work!

PR #23637, opened 2025-02-02
all: update copyright year to 2025

ZihxS:
This PR updates the copyright year to 2025 across multiple files to ensure consistency and reflect the current year.

No functional changes were made; this is purely a maintenance update.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019-2024 Alexander Medvednikov
+Copyright (c) 2019-2025 Alexander Medvednikov
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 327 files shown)


JalonSolov:
Not necessary... and no longer recommended in any case.

Older copyright law said to set the date when the file was created, and leave it alone.

Current copyright law says a date is not needed at all.

ZihxS:
@JalonSolov does that mean it's better to change it to "Copyright (c) 2019/created year" ?

JalonSolov:
"The general rule is that the year to include in a copyright notice is the year of first publication of the work. First publication is when the work is made available to the public without restriction."

This means there should only be a single year mentioned, and it should be the year it was first merged into the repo.

The Berne Convention recently said that a copyright notice isn't needed at all, but not everyone has caught up with this idea, and it doesn't hurt to leave it in the files.

It simply isn't necessary to change the date after the file has been created.

spytheman:
I also think, that we should remove the years - https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/ , and leave the date tracking to Git .

JalonSolov:
I would leave the copyright as-is for now.  New files can be added without a date, but there's no need to modify all the existing files.  That would be thousands(?) of files across all the repos...

spytheman:
ok, incrementally updating only newly modified files also works, and will avoid the big bulk changes

PR #23649, opened 2025-02-03
tools: let `v test .` show the running _test.v files each minute (set by `VTEST_REPORT_RUNNING_PERIOD_MS`); diagnose stuck windows gcc CI jobs

spytheman:


spytheman:
0d53e3e is mostly unrelated, but is here to trigger more of the CI jobs to run, since changes to `cmd/tools/modules/testing/`, do not affect them :-| ...

PR #23662, opened 2025-02-07
crypto.ecdsa: migrate generate_key and makes it more simpler.

blackshirt:
This is a 3rd PR sent in efforts to migrate `crypto.ecdsa` module to use high level api. In this current PR, its basically migrates two thing, ie, in the form:

- Migrates `PrivateKey.public_key()` to use the new one. Its approach was done by serializing this high level opaque (if availables) to der serialized bytes and get its back into other opaque and use this as public key part. If this high level opaque (`evpkey` field) was not availables, it will use the old one.
- Simplify `generate_key` function to use this new approach , by just more simpler calling to the `.new()` and `.public_key()` .
- Removes `.key_free()` .. Its has been replaced by its explicit counterparts of `.free()` of the desired key.
- Just some cleansup of bits

Happy days

### DIFF
--- a/vlib/crypto/ecdsa/ecdsa.c.v
+++ b/vlib/crypto/ecdsa/ecdsa.c.v
@@ -22,7 +22,9 @@ module ecdsa
 #include <openssl/x509.h>
 #include <openssl/bio.h>
 #include <openssl/pem.h>
-#include <openssl/core.h>
+
+// The following header is available on OpenSSL 3.0, but not in OpenSSL 1.1.1f
+//#include <openssl/core.h>
 
 // The new opaque of public key pair high level API
 @[typedef]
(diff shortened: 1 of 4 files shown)


blackshirt:
@felipensp , @spytheman : can you look to the jobs failed? Its run fine in my test box

spytheman:
It should be fine now. I've commented out the `openssl/core.h` inclusion (that particular header is relatively new, see https://docs.openssl.org/3.4/man7/openssl-core.h/ ).

spytheman (approved):

blackshirt:
> It should be fine now. I've commented out the `openssl/core.h` inclusion (that particular header is relatively new, see https://docs.openssl.org/3.4/man7/openssl-core.h/ ).

Thank..its my miss..

PR #23665, opened 2025-02-07
cgen: fix `$if typeof[T]().idx`, `$if typeof[T]().unaliased_typ` checking

felipensp:


vlib/v/type_resolver/comptime_resolver.v
```diff
@@ -82,6 +82,27 @@ pub fn (mut t TypeResolver) typeof_type(node ast.Expr, default_type ast.Type) as
...
+		'idx' {
+			return t.resolver.unwrap_generic(typ)
+		}
+		'unaliased_typ' {
+			return t.table.unaliased_type(t.resolver.unwrap_generic(typ))
+		}
+		'indirections' {
+			return ast.int_type
+		}
+		else {
```
spytheman, line 100: there is also 'typ', but that may have been handled elsewhere?

spytheman: see `ast.GenericKindField` from `vlib/v/ast/ast.v:286`

spytheman (approved):
Excellent work.

vlib/v/type_resolver/comptime_resolver.v
felipensp: Yes, I think there is a room for improvement there.

PR #23670, opened 2025-02-08
checker: cleanup check_expected_call_arg

felipensp:
Remove repeated fn call to `got.is_ptr()` and `expected.is_ptr()`.


spytheman (approved):
Excellent work.

PR #23673, opened 2025-02-08
Doc: remove paragraph about vfmt renaming go to spawn.

mike-ward:
Removed the paragraph and about `vfmt` renaming `go` to `spawn`. Add a sentence about using `go` keyword. More about `go` coroutines should be added but for now this pull request removes the outdated stuff.


spytheman (approved):
Excellent work.

PR #23677, opened 2025-02-08
encoding.csv: add support for multithreading to `encoding.csv.RandomAccessReader`

penguindark:
**What's inside**

- Added support for multithreading to `encoding.csv.RandomAccessReader`
- Added multithreading tests

The multithreading was performed with file up to 6GB with 60M of rows.

*Note:* if the file is huge the use of `-thread-stack-size` flag is needed to adjust the stack size

### DIFF
--- a/vlib/encoding/csv/csv_reader_random_access.v
+++ b/vlib/encoding/csv/csv_reader_random_access.v
@@ -454,13 +454,17 @@ pub fn (mut cr RandomAccessReader) get_cellt(cfg GetCellConfig) !CellValue {
 	if cr.header_row >= 0 && cfg.x < cr.header_list.len {
 		h := cr.header_list[cfg.x]
 		res := cr.get_cell(cfg)!
-		if h.htype == .int {
-			return res.int()
+		match h.htype {
+			.int {
+				return res.trim_space().int()
+			}
+			.string {
+				return res
+			}
+			.f32 {
+				return res.trim_space().f32()
+			}
 		}
-		if h.htype == .f32 {
-			return res.f32()
-		}
-		return res
 	}
 	return cr.get_cell(cfg)!
 }
--- a/vlib/encoding/csv/csv_reader_random_access.v
+++ b/vlib/encoding/csv/csv_reader_random_access.v
@@ -43,6 +43,7 @@ pub mut:
 	htype  ColumType = .string
 }
 
+@[heap]
 pub struct RandomAccessReader {
 pub mut:
 	index i64
@@ -71,7 +72,8 @@ pub mut:
 	mem_buf_start i64 = -1 // start index in the file of the read buffer
 	mem_buf_end   i64 = -1 // end index in the file of the read buffer
 	// csv map for quick access
-	csv_map [][]i64
+	create_map_csv bool = true // flag to enamble the csv map creation
+	csv_map        [][]i64
 	// header
 	header_row  int = -1 // row index of the header in the csv_map
 	header_list []HeaderItem   // list of the header item
@@ -81,19 +83,20 @@ pub mut:
 @[params]
 pub struct RandomAccessReaderConfig {
 pub:
-	scr_buf      voidptr // pointer to the buffer of data
-	scr_buf_len  i64     // if > 0 use the RAM pointed from scr_buf as source of data
-	file_path    string
-	start_index  i64
-	end_index    i64    = -1
-	mem_buf_size int    = 1024 * 64 // default buffer size 64KByte
-	separator    u8     = `,`
-	comment      u8     = `#` // every line that start with the quote char is ignored
-	default_cell string = '*' // return this string if out of the csv boundaries
-	empty_cell   string // return this string if empty cell
-	end_line_len int = endline_cr_len // size of the endline rune
-	quote        u8  = `"`            // double quote is the standard quote char
-	quote_remove bool // if true clear the cell from the quotes
+	scr_buf        voidptr // pointer to the buffer of data
+	scr_buf_len    i64     // if > 0 use the RAM pointed from scr_buf as source of data
+	file_path      string
+	start_index    i64
+	end_index      i64    = -1
+	mem_buf_size   int    = 1024 * 64 // default buffer size 64KByte
+	separator      u8     = `,`
+	comment        u8     = `#` // every line that start with the quote char is ignored
+	default_cell   string = '*' // return this string if out of the csv boundaries
+	empty_cell     string // return this string if empty cell
+	end_line_len   int = endline_cr_len // size of the endline rune
+	quote          u8  = `"`            // double quote is the standard quote char
+	quote_remove   bool // if true clear the cell from the quotes
+	create_map_csv bool = true // make the map of the csv file if true
 }
 
 /******************************************************************************
@@ -177,7 +180,10 @@ pub fn csv_reader(cfg RandomAccessReaderConfig) !&RandomAccessReader {
 	cr.quote_remove = cfg.quote_remove
 	cr.quote = cfg.quote
 
-	cr.map_csv()!
+	cr.create_map_csv = cfg.create_map_csv
+	if cr.create_map_csv {
+		cr.map_csv()!
+	}
 
 	return cr
 }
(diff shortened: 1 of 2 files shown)


vlib/encoding/csv/csv_reader_random_access.v
```diff
@@ -71,7 +72,8 @@ pub mut:
 	mem_buf_start i64 = -1 // start index in the file of the read buffer
 	mem_buf_end   i64 = -1 // end index in the file of the read buffer
 	// csv map for quick access
-	csv_map [][]i64
+	create_map_csv bool = true // flag to enamble the csv map creation
```
spytheman, line 75:
```suggestion
	create_map_csv bool = true // flag to enable the csv map creation
```

vlib/encoding/csv/csv_reader_random_access.v
```diff
@@ -81,19 +83,20 @@ pub mut:
...
+	end_index      i64    = -1
+	mem_buf_size   int    = 1024 * 64 // default buffer size 64KByte
+	separator      u8     = `,`
+	comment        u8     = `#` // every line that start with the quote char is ignored
+	default_cell   string = '*' // return this string if out of the csv boundaries
+	empty_cell     string // return this string if empty cell
+	end_line_len   int = endline_cr_len // size of the endline rune
+	quote          u8  = `"`            // double quote is the standard quote char
+	quote_remove   bool // if true clear the cell from the quotes
+	create_map_csv bool = true // make the map of the csv file if true
```
spytheman, line 99:
```suggestion
	create_map_csv bool = true // if true make the map of the csv file
```

vlib/encoding/csv/csv_reader_random_access.v
```diff
@@ -226,6 +232,18 @@ fn (mut cr RandomAccessReader) fill_buffer(i i64) !i64 {
 	return i64(-1)
 }
 
+// copy_configuration copy the configuration from another csv RandomAccessReader using reference
```
spytheman, line 235:
```suggestion
// copy_configuration copies the configuration from another csv RandomAccessReader
```

vlib/encoding/csv/csv_reader_random_access.v
```diff
@@ -226,6 +232,18 @@ fn (mut cr RandomAccessReader) fill_buffer(i i64) !i64 {
...
+// this function is an helper to user the RandomAccessReader in multi thread applicatiuons
```
spytheman, line 236:
```suggestion
// this function is a helper for using the RandomAccessReader in multi threaded applications
```

vlib/encoding/csv/csv_reader_random_access.v
```diff
@@ -428,13 +454,17 @@ pub fn (mut cr RandomAccessReader) get_cellt(cfg GetCellConfig) !CellValue {
...
 		h := cr.header_list[cfg.x]
 		res := cr.get_cell(cfg)!
-		if h.htype == .int {
-			return res.int()
-		}
-		if h.htype == .f32 {
-			return res.f32()
+		match h.htype {
+			.int {
+				return res.trim_space().int()
```
spytheman, line 459: Is the trim_space() here needed?

vlib/encoding/csv/csv_reader_test.v
```diff
@@ -347,3 +348,125 @@ fn test_coherence() {
...
+
+		mut threads := []thread{}
+		for task_index in 0 .. slices {
+			threads << spawn read_lines(task_index, csvr[task_index], mut &data, start,
+				end)
+			start = end
+			end = if (start + step) > n_rows { n_rows } else { start + step }
+		}
+		threads.wait()
+	}
```
spytheman, line 456: why is the whole block unsafe?

vlib/encoding/csv/csv_reader_test.v
```diff
@@ -347,3 +348,125 @@ fn test_coherence() {
 fn main() {
 	test_csv_string()
 }
+
+// Multithreaded tests
+
+fn create_csv(file_path string, size int) !i64 {
+	// create csv in ram
```
spytheman, line 355: That comment seems incorrect, given that the function is writing to a file 🤔?

vlib/encoding/csv/csv_reader_random_access.v
spytheman: hm ... it seems to be, although that is not checked in the tests (the generated data there contains numbers without spaces)

vlib/encoding/csv/csv_reader_random_access.v
```diff
@@ -226,6 +232,18 @@ fn (mut cr RandomAccessReader) fill_buffer(i i64) !i64 {
...
 
+// copy_configuration copy the configuration from another csv RandomAccessReader using reference
+// this function is an helper to user the RandomAccessReader in multi thread applicatiuons
+// pay attention to the free process
+pub fn (mut cr RandomAccessReader) copy_configuration(src_cr RandomAccessReader) {
+	cr.header_row = src_cr.header_row
+	unsafe {
+		cr.header_list = &src_cr.header_list
+		cr.header_map = &src_cr.header_map
+		cr.csv_map = &src_cr.csv_map
```
spytheman, line 243: That seems wrong 🤔.
The type of .header_list is `[]HeaderItem`.
The type of .header_map is `map[string]int`.
The type of .csv_map is `[][]i64`.

All of them are value types, without pointers involved.
V's `cr.header_list = &src_cr.header_list` generates:
C's `cr->header_list = *&(*(src_cr)).header_list;`
i.e. it copies the array descriptor content, without cloning the data of the elements in it.

The same can be achieved by using just: `cr.header_list = src_cr.header_list` which in the unsafe block will generate the following simpler C:
 `cr->header_list = (*(src_cr)).header_list;`

spytheman: Doing `cr.header_map = src_cr.header_map` however leads to a V error:
```
vlib/encoding/csv/csv_reader_random_access.v:242:26: error: cannot copy map: call `move` or `clone` method (or use a reference)
  240 |     unsafe {
  241 |         cr.header_list = src_cr.header_list
  242 |         cr.header_map = src_cr.header_map
      |                                ~~~~~~~~~~
```

spytheman: I think that we should allow `cr.header_map = src_cr.header_map` in unsafe blocks, to keep the symmetry with the array assignment.

spytheman: However, given that this operation is done just once per file, in the non threaded sequential part of the tests, another change can be simpler and clearer:
image

The tests still pass after it.

vlib/encoding/csv/csv_reader_test.v
spytheman: The only line that needs the unsafe{} to avoid a warning, is:
```
data = unsafe { [][]csv.CellValue{len: csvr[0].header_list.len, init: []csv.CellValue{len: n_rows}} }
``

vlib/encoding/csv/csv_reader_test.v
penguindark: right!

vlib/encoding/csv/csv_reader_test.v
penguindark: you are right, that `unsafe` was large because I was doing tests.

vlib/encoding/csv/csv_reader_random_access.v
penguindark: I've tried different solutions, but this was the one working I found,
for sure there are issues with the internal V use of the pointers in this situation with array of arrays.
Have you any suggestion?

penguindark:
@spytheman I modified the code, but the part for "copy configuration" that I want an alias not a copy 
work only in the code that we already have.
Maybe we have issues with pointers for array of array in the compiler, maybe... 🤔

vlib/encoding/csv/csv_reader_random_access.v
penguindark: to be clear, I want a "pointer"  to the configuration, not copy.
Avoiding copy data is the purpose of this part of code.

spytheman: > to be clear, I want a "pointer" to the configuration, not copy
ok, but why?

The .header_map and .header_list fields afaik represent meta data for the header line, which is minuscule for big .csv files. Is sharing it worth the potential concurrency problems, when it is updated by multiple threads?

The .csv_map could be bigger, but even for it, `unsafe { cr.csv_map = src_cr.csv_map }` will make a copy of the `array` descriptor struct, and *NOT* of the data elements.

In my view, there is simply no reason for using `unsafe { cr.csv_map = &src_cr.csv_map }` and then relying on V autodereferencing the & that you did.

spytheman: You can try `-o x.c` over the test, and see the generated C code for yourself.

Issue #23346, opened 2025-01-02
`@[sql: serial]` and `@[serial]` are not the same.


islonely:
**V doctor:**

```
V full version: V 0.4.9 4225a34.b79257b
OS: macos, macOS, 15.1.1, 24B91
Processor: 8 cpus, 64bit, little endian, Apple M3

getwd: /Users/adamoates/.vmodules/islonely/hex
vexe: /Users/adamoates/v/v
vexe mtime: 2025-01-02 15:24:10

vroot: OK, value: /Users/adamoates/v
VMODULES: OK, value: /Users/adamoates/.vmodules
VTMP: OK, value: /tmp/v_501

Git version: git version 2.39.5 (Apple Git-154)
Git vroot status: weekly.2024.53-14-gb79257b7
.git/config present: true

CC version: Apple clang version 16.0.0 (clang-1600.0.26.4)
emcc version: N/A
thirdparty/tcc status: thirdparty-macos-arm64 713692d4
```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  /Users/adamoates/Documents/test.v && /Users/adamoates/Documents/test`

```v
import db.sqlite

struct Parent {
	id       int     @[primary; serial]
	children []Child @[fkey: 'parent_id']
}

struct Child {
	id        int @[primary; sql: serial]
	parent_id int
}

fn main() {
	mut db := sqlite.connect(':memory:')!
	parent := Parent{
		children: [Child{}, Child{}]
	}
	sql db {
		create table Child
		create table Parent
		insert parent into Parent
	}!
	parents := sql db {
		select from Parent limit 1
	}!

	assert parent.children.len == parents[0].children.len
}
```

**What did you see?**

```
/Users/adamoates/Documents/test.v:27: FAIL: fn main.main: assert parent.children.len == parents[0].children.len
  left value: parent.children.len = 2
  right value: parents[0].children.len = 0
V panic: Assertion failed...
v hash: b79257b
0   test                                0x0000000102cb5658 main__main + 5388
1   test                                0x0000000102cb5a60 main + 84
2   dyld                                0x00000001948fc274 start + 2840
```

**What did you expect to see?**

Expected code to assert true. Changing the `Parent` struct to this below fixes the issue. But [ORM docs](https://modules.vlang.io/orm.html) say that `@[sql: serial]` and `@[serial]` do the same thing.

```v
struct Parent {
	id       int     @[primary; sql: serial]
	children []Child @[fkey: 'parent_id']
[...]

PR #23678, opened 2025-02-09
orm,cgen: fix `@[sql: serial]` and `@[serial]` are not the same

nunsez:
Fix https://github.com/vlang/v/issues/23346


spytheman (approved):
Excellent work.

Issue #23683, opened 2025-02-10
net.http.file: can't handle Chinese filename


kbkpbot:
### Describe the bug

net.http.file: can't handle Chinese filename

### Reproduction Steps

file.v:

```v
import net.http.file

fn main() {
    file.serve(folder: '.', on: ":8080")
}
```

create a new file name "测试.txt" in current directory:

```
测试测试
```

Use browser `http://127.0.0.1:8080/测试.txt`

### Expected Behavior

show the file content of "测试.txt"

```
测试测试
```

### Current Behavior

```
no such file
```

### Possible Solution

uri handle in `vlib/net/http/file/static_server.v` has problem
It must use `urllib` decode the uri first.

### V version

V 0.4.9 adbc869.e129ec5

### Environment details (OS name and version, etc.)

|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.1 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |1.29GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-02-10 14:14:51
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /media/HD/github/kbkpbot/v
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.1-193-ge129ec5b
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 0134e9b9
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.4) 2.39

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #23684, opened 2025-02-10
net.http.file: use urllib decode uri

kbkpbot:
Fix issue #23683


spytheman (approved):
Excellent work.
Thank you 🙇🏻‍♂️ .

PR #23686, opened 2025-02-10
Added os.quoted_path to os.system calls for the compress flag

### LINKED ISSUE #23685: -compress flag fails for paths that require quoting
### Describe the bug

File paths with characters the require quoting, e.g. space, fail when compiled with the -compress flag.

### Reproduction Steps

create file "te st.v"
`v te\ st.v -compress`

### Expected Behavior

No Errors

### Current Behavior

strip: '/home/lars/te': No such file
strip: 'st': No such file
strip failed

### Possible Solution

add os.quoted_path for the compress flag

_No response_

### V version

V 0.4.9 e129ec5

|V full version      |V 0.4.9 d680c42.e129ec5
|OS                  |linux, "Void Linux"
|Processor           |12 cpus, 64bit, little endian, AMD Ryzen 5 5600G with Radeon Graphics
|Memory              |4.19GB/15.52GB
|                    |
|V executable        |/home/lars/v/v
|V last modified time|2025-02-10 16:41:30
|                    |
|V home dir          |OK, value: /home/lars/v
|VMODULES            |OK, value: /home/lars/.vmodules
|Current working dir |OK, value: /home/lars
|                    |
|V git status        |weekly.2025.03-109-ge129ec5b
|.git/config present |true
|                    |
|cc version          |cc (GCC) 13.2.0
|gcc version         |gcc (GCC) 13.2.0
|clang version       |clang version 17.0.6
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.39
> [!NOTE]
>

Larsimusrex:
builder: Fix #23685

Added os.quoted_path for all os.system calls for the compress option.

### DIFF
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -810,16 +810,16 @@ pub fn (mut v Builder) cc() {
 		break
 	}
 	if v.pref.compress {
-		ret := os.system('strip ${v.pref.out_name}')
+		ret := os.system('strip ${os.quoted_path(v.pref.out_name)}')
 		if ret != 0 {
 			println('strip failed')
 			return
 		}
 		// Note: upx --lzma can sometimes fail with NotCompressibleException
 		// See https://github.com/vlang/v/pull/3528
-		mut ret2 := os.system('upx --lzma -qqq ${v.pref.out_name}')
+		mut ret2 := os.system('upx --lzma -qqq ${os.quoted_path(v.pref.out_name)}')
 		if ret2 != 0 {
-			ret2 = os.system('upx -qqq ${v.pref.out_name}')
+			ret2 = os.system('upx -qqq ${os.quoted_path(v.pref.out_name)}')
 		}
 		if ret2 != 0 {
 			println('upx failed')


JalonSolov:
Output paths should probably always be quoted, no matter what operations are being done.  :-\

JalonSolov:
In other words, when the `v.pref.out_name` field is first populated, perhaps it should be quoted then, so you don't have to worry about quoting it later.

Larsimusrex:
I don't know. I think that would probably break other stuff that uses `v.pref.out_name` 
e.g. 
```v
$if windows {
	if v.ccoptions.cc == .tcc {
		def_name := v.pref.out_name[0..v.pref.out_name.len - 4]
		v.pref.cleanup_files << '${def_name}.def'
	}
}
```

JalonSolov:
Yes, that would have to be changed.  My main thought was that there are likely other places where the path should be quoted, and isn't right now.  Instead of adding this extra code to every place that needs it, it would be better to have it done in one place.

spytheman (approved):
Excellent work.
Thank you.

PR #23694, opened 2025-02-11
tools: reduce: improve parser for functions and add more loops to reduce more

Eliyaan:
Improve the parser by enabling full function removal by considering it as a scope (before it left the signature and the brackets out of the scope)
Add an outer loop to the reducing process to clean as much code it can.

Before: lots leftover unnecessary functions that needed to be removed by hand
for my last MRE made by reduce for a 3k lines file: 66 lines
now: made by reduce for the same file: 9 lines (a const leftover, a struct used by the error producing function, and the function in itself)


spytheman (approved):
Excellent work, very big improvement!

PR #23697, opened 2025-02-12
tools: improve `v reduce` output; bump version, warn on failed string_reproduces/3, but continue to run

spytheman:
Make the output of `v reduce` easier to scan, let it continue after most current assertions fail, since the output is still useful.

Given this code in `a.c.v`:

```v
#define FAIL_TO_COMPILE abc()

fn C.FAIL_TO_COMPILE()

fn main() {
	C.FAIL_TO_COMPILE()
}
```

with this PR, it produces:
image

The resulting `rpdc.v` is shorter and still works for a reproduction of the intentional problem 🥳 .

### DIFF
--- a/cmd/tools/vreduce.v
+++ b/cmd/tools/vreduce.v
@@ -1,7 +1,9 @@
 import os
 import flag
 import math
+import log
 
+const version = '0.0.2'
 const default_command = '${os.quoted_path(@VEXE)} -no-skip-unused' // Command used to compile the program, using -no-skip-unused to ease the reducing
 const default_error_msg = 'C compilation error' // the pattern to reproduce
 // Temporary files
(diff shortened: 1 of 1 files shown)


spytheman:
@Eliyaan what do you think?

spytheman:
I am merging it, since I did not saw any ways it could break - the output will always be either equal or shorter than the input.

Eliyaan:
The asserts for the scopes show that the parser did not handle the program well (it thinks there will always be {} after a function)
But it is good that now these are warnings

spytheman:
> The asserts for the scopes show that the parser did not handle the program well

Yes, but it is still good to produce what it had achieved so far, even if incomplete, it will be smaller or equal to the original.

PR #23705, opened 2025-02-13
crypto.ecdsa: migrate core routines for signing (and verifying)

blackshirt:
This PR is the main part of the migration efforts to the `crypto.ecdsa` module to support of the high level unobsoleted API. The old OpenSSL 1.1.1.  series has reached its End of Life (EOL). As such it will no longer receive publicly available security fixes, see [eol-of-ossl.1.1.1](https://openssl-library.org/post/2023-09-11-eol-111/) for more detail, so, maybe better to just upgrade your openssl to the 3.x series.

In this current PR, the main concerns was adding support to `.sign()` and `.verify()` routines to support the new api. Its done by using some changes to the internal detail of the current one, ie, in the form:
- Adds support to use a new high level opaque into `.sign()` and `.verify()` methods.
- Renames internal `.sign_message` into `.sign_digest` to reflect underlying its wrapped. Its sign the digest internally.
- Adds some helpers to cooperate with the rest. Some helpers are just redundant with the old one, because we supporting two opaque here, but, its can be removed (cleaned) on later pr.
- Adds some C required declarations.
- Some bits of cleans up 

Thats its. thanks

### DIFF
--- a/vlib/crypto/ecdsa/ecdsa.c.v
+++ b/vlib/crypto/ecdsa/ecdsa.c.v
@@ -15,7 +15,7 @@ module ecdsa
 #flag darwin -L/usr/local/opt/openssl/lib
 #include <openssl/ecdsa.h>
 #include <openssl/obj_mac.h>
-#include <openssl/objects.h>
+#include <openssl/types.h>
 #include <openssl/bn.h>
 #include <openssl/evp.h>
 #include <openssl/ec.h>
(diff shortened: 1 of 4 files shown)


spytheman:
(should be merged after https://github.com/vlang/v/pull/23706, which changes the runners to Ubuntu 24.04, which has OpenSSL 3)

blackshirt:
> (should be merged after #23706, which changes the runners to Ubuntu 24.04, which has OpenSSL 3)

Okey ..i saw `v2` related errors i cant reproduce.. is this related with the ci runners?

JalonSolov:
I only see `crypto/ecdsa` test failures in the last run.

spytheman:
(rebased over current master, which now runs the CI jobs on Ubuntu 22.04 and Ubuntu 24.04, which should have OpenSSL 3, instead of OpenSSL 1.1)

blackshirt:
> (rebased over current master, which now runs the CI jobs on Ubuntu 22.04 and Ubuntu 24.04, which should have OpenSSL 3, instead of OpenSSL 1.1)

Thanks, i think its good to move the new one start from here

PR #23706, opened 2025-02-13
ci: update linux image runners from `ubuntu-20.04` to `ubuntu-24.04`

spytheman:
- **ci: use ubuntu-20.04 instead of ubuntu-24.04 in all the CI runners**


spytheman:
(rebased over 2f700c0, to get the clang-18 sanitize fixes)

PR #23709, opened 2025-02-13
tools: reduce: fix warnings

Eliyaan:
fix: warn_on_false did not check if it was false

Now it does not warn for nothing.


spytheman (approved):
My bad. Thanks for fixing it.

PR #23724, opened 2025-02-14
v: fix compiler for -no-builtin

felipensp:
Before this patch, building such code with `-no-builtin` was failing because GC code (that belongs to builtin)

```
a := [2]int{}
```

With the following code:
```
module main

import v.builder.nativebuilder

fn main() {
	nativebuilder.start()
}
```
Now it displays:
`error: V builtin usage detected, but -no-builtin was supplied`


spytheman:
I am not sure that is better than the existing behavior, that at least produces something. The new error message is not specific (no position information), and it will trigger even for situations and code that could previously be compiled with `-no-builtin`. The native backend is still in development and imho it is reasonable to expect that people that are involved with it, can handle the resulting occasional panics during development better, than some non descriptive and non specific error message.

PR #23725, opened 2025-02-14
veb.csrf: Make default printing of verbose data optional via config flag

xD0135:
We introduce a new `verbose` option in `CsrfConfig` to move the default print statements behind this flag in order to allow users to make it configurable and select when it should be enabled.


spytheman (approved):
Excellent work.

PR #23727, opened 2025-02-14
veb.request_id: New middleware that implements request ID tracking

xD0135:
# Request ID Middleware

This module implements request ID tracking functionality for V web applications. Request IDs are unique identifiers assigned to each HTTP request, which is essential for request tracing, debugging, and maintaining distributed systems.

## Purpose

Request IDs help in:

- Tracking requests across distributed systems
- Correlating logs from different services
- Debugging and troubleshooting
- Performance monitoring
- Request chain tracing

## Usage

To enable request ID tracking in your veb app, you must embed the `RequestIdContext` struct in your `Context` struct.

**Example:**

```v
import veb
import veb.request_id

pub struct Context {
    veb.Context
    request_id.RequestIdContext
}
```

### Basic Configuration

Here's a simple configuration example:

```v
const request_id_config := request_id.Config{
    header: 'X-Request-ID'
    generator: rand.uuid_v4
}
```

### Middleware Setup

Enable request ID tracking for all routes or specific routes using veb's middleware system.

**Example:**

```v
pub struct App {
    veb.Middleware[Context]
}

fn main() {
    mut app := &App{}
    // Register the RequestID middleware with custom configuration
    app.use(request_id.middleware[Context](request_id_config))
}
```

### Accessing the Request ID

You can access the request ID in your route handlers:

```v
fn (app &App) handler(mut ctx Context) veb.Result {
    // Get the current request ID
    request_id := ctx.request_id
    // Use the request ID for logging, etc.
    return ctx.text('Request ID: $request_id')
}
```

## Configuration Options

The `Config` struct provides several configuration options:

```v
pub struct Config {
pub:
    // Next defines a function to skip this middleware when returned true
    next ?fn (ctx &veb.Context) bool
    // Generator defines a function to generate the unique identifier
    generator fn () string = rand.uuid_v4
    // Header is the header key where to get/set the unique request ID
    header string = 'X-Request-ID'
    // Allow empty sets whether to allow empty request IDs
    allow_empty bool
    // Force determines whether to always generate a new ID even if one exists
    force bool
}
```

### Configuration Options Explained

- `next`: Optional function to conditionally skip the middleware
- `generator`: Function to generate unique IDs (defaults to UUID v4)
- `header`: HTTP header name for the request ID (defaults to "X-Request-ID")
- `allow_empty`: Whether to allow empty request IDs
- `force`: Whether to generate a new ID even when one already exists

### Custom ID Generator

You can provide your own ID generator function:

```v
fn custom_id_generator() string {
    return 'custom-prefix-${rand.uuid_v4()}'
}

config := request_id.Config{
    generator: custom_id_generator
}
```

### Conditional Middleware Execution

Use the `next` function to skip the middleware based on custom logic:

```v
config := request_id.Config{
    next: fn (ctx &veb.Context) bool {
        // Skip for health check endpoints
        return ctx.req.url.starts_with('/health')
    }
}
```

### Forcing New IDs

When you want to ensure a new ID is generated regardless of existing headers:

```v
config := request_id.Config{
    force: true
}
```

## Best Practices

1. **Consistent Headers**: Use consistent header names across your services
2. **ID Propagation**: Forward request IDs to downstream services
3. **Logging Integration**: Include request IDs in your logging system
4. **ID Format**: Use a reliable ID generator (UUID v4 is recommended)

## Security Considerations

While request IDs are not security features, consider these points:

- Don't include sensitive information in request IDs
- Validate request ID format if using custom generators
- Be cautious with request ID length (recommended: 8-128 characters)

### Basic Integration

```v
module main

import veb
import veb.request_id

pub struct Context {
    veb.Context
    request_id.RequestIdContext
}

pub struct App {
    veb.Middleware[Context]
}

@['/request-id'; get]
    return ctx.text("Current request ID: ${ctx.request_id}")
}

fn main() {
    mut app := &App{}
    config := request_id.Config{
        header:      'X-Request-ID'
        force:       false
        allow_empty: false
	}
    app.use(request_id.middleware[Context](config))
}
```

### With Custom Generator and Conditional Execution

```v
config := request_id.Config{
    generator: fn () string {
        return 'app-${rand.uuid_v4()}'
    }
    next: fn (ctx &veb.Context) bool {
        return ctx.req.url.starts_with('/public')
    }
}
```

### DIFF
--- a/vlib/veb/request_id/README.md
+++ b/vlib/veb/request_id/README.md
@@ -1,0 +1,202 @@
+# Request ID Middleware
+
+This module implements request ID tracking functionality for V web applications. Request IDs are unique identifiers assigned to each HTTP request, which is essential for request tracing, debugging, and maintaining distributed systems.
+
+## Purpose
+
+Request IDs help in:
+- Tracking requests across distributed systems
+- Correlating logs from different services
+- Debugging and troubleshooting
+- Performance monitoring
+- Request chain tracing
+
+## Usage
+
+To enable request ID tracking in your veb app, you must embed the `RequestIdContext` struct in your `Context` struct.
+
+**Example:**
+
+```v
+import veb
+import veb.request_id
+
+pub struct Context {
+    veb.Context
+    request_id.RequestIdContext
+}
+```
+
+### Basic Configuration
+
+Here's a simple configuration example:
+
+```v
+const request_id_config := request_id.Config{
+    header: 'X-Request-ID'
+    generator: rand.uuid_v4
+}
+```
+
+### Middleware Setup
+
+Enable request ID tracking for all routes or specific routes using veb's middleware system.
+
+**Example:**
+
+```v
+pub struct App {
+    veb.Middleware[Context]
+}
+
+fn main() {
+    mut app := &App{}
+    // Register the RequestID middleware with custom configuration
+    app.use(request_id.middleware[Context](request_id_config))
+    veb.run[App, Context](mut app, 8080)
+}
+```
+
+### Accessing the Request ID
+
+You can access the request ID in your route handlers:
+
+```v
+fn (app &App) handler(mut ctx Context) veb.Result {
+    // Get the current request ID
+    request_id := ctx.request_id
+    // Use the request ID for logging, etc.
+    return ctx.text('Request ID: $request_id')
+}
+```
+
+## Configuration Options
+
+The `Config` struct provides several configuration options:
+
+```v
+pub struct Config {
+pub:
+    // Next defines a function to skip this middleware when returned true
+    next ?fn (ctx &veb.Context) bool
+    // Generator defines a function to generate the unique identifier
+    generator fn () string = rand.uuid_v4
+    // Header is the header key where to get/set the unique request ID
+    header string = 'X-Request-ID'
+    // Allow empty sets whether to allow empty request IDs
+    allow_empty bool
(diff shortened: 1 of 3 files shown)


spytheman:
Please run `v check-md vlib/veb/request_id/README.md` locally.

spytheman:
Use ```v ignore  ... for code sections containing partial code.

spytheman:
(you can see the rest of the modifiers with `v check-md` without other arguments)

spytheman:

KeitoTobi1:
@xD0135 
You should ask those things on Discord, not in the comments section.

spytheman:
> Any updates on merging

I think that it looks good, however I do not understand the goal very much (I do not have much  experience in building distributed systems with http webservices in them; personally I try to optimize my own stuff enough, so that it *is not* needed to be distributed...), and thus I am not a good arbiter, as to whether it should be merged or not.

I've asked Alex to review and merge it, but it seems that he is busy :-| .

enghitalo:
@xD0135, From what I understand, this PR would be about inserting some information in the header, to facilitate the tracking of some requests, which would be very useful in some observability systems. I understand your intention, but I think this added some unnecessary complexity, since all of this could be done with features already existing in V (Let me know if I'm wrong). If that's the case, I think a tutorial or some documentation describing how to handle this with existing resources would be enough.

medvednikov:
Looks good, thanks @xD0135

PR #23731, opened 2025-02-15
checker, cgen: fix sumtype smartcasted var as inherited var

### LINKED ISSUE #23716: Sum Type Field Access Fails Inside closure
### Describe the bug

When using a sum type (Type = Aa | Bb) inside a function literal (fn[t]), the compiler fails to recognize the specific type variant of t and treats it as the general Type, rather than the matched variant (Aa).

### Reproduction Steps

```v

struct Aa {a int}
struct Bb {b int}
type Type = Aa | Bb

fn main() {
	t := Type(Aa{})
	match t {
		Aa {
			println(t.a)            // works
			func := fn[t] () {
				println(typeof(t))  // works, prints "Aa"
				// println(t.a) // error: field `a` does not exist or have the same type in these sumtype `Type` variants: Bb
			}
			func()
		}
		Bb {}
	}
}

```

### Expected Behavior

The expected behavior is that inside the function literal (`fn[t]`), the compiler should preserve the type information from the surrounding match branch. Since t is already matched as `Aa`, it should be treated as an `Aa` inside the function, allowing direct access to `t.a` without requiring additional casting.

### Current Behavior

`t`'s type inside the closure is the sumtyp (`Type`), and not `Aa`

### Possible Solution

_No response_

_No response_

### V version

V 0.4.9 5be2fca

|V full version      |V 0.4.9 5be2fca
|OS                  |linux, Ubuntu 24.10
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 8840U w/ Radeon 780M Graphics
|Memory              |9.51GB/22.78GB
|                    |
|V executable        |/home/sharp/dev/v/v
|V last modified time|2025-02-13 12:21:16
|                    |
|V home dir          |OK, value: /home/sharp/dev/v
|VMODULES            |OK, value: /home/sharp/.vmodules
|                    |
|Git version         |git version 2.45.2
|V git status        |weekly.2025.03-35-g5be2fcab-dirty (99 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 14.2.0-4ubuntu2) 14.2.0
|gcc version         |gcc (Ubuntu 14.2.0-4ubuntu2) 14.2.0
|clang version       |Ubuntu clang version 19.1.1 (1ubuntu1)
|emcc version        |emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.6 ()
|glibc version       |ldd (Ubuntu GLIBC 2.40-1ubuntu3.1) 2.40
> [!NOTE]
>

felipensp:
Fix #23716

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -634,8 +634,13 @@ fn (mut c Checker) anon_fn(mut node ast.AnonFn) ast.Type {
 			c.error('original `${parent_var.name}` is immutable, declare it with `mut` to make it mutable',
 				var.pos)
 		}
+		ptyp := if parent_var.smartcasts.len > 0 {
+			parent_var.smartcasts.last()
+		} else {
+			parent_var.typ
+		}
 		if parent_var.typ != ast.no_type {
-			parent_var_sym := c.table.final_sym(parent_var.typ)
+			parent_var_sym := c.table.final_sym(ptyp)
 			if parent_var_sym.info is ast.FnType {
 				ret_typ := c.unwrap_generic(parent_var_sym.info.func.return_type)
 				if ret_typ.has_flag(.generic) {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #23737, opened 2025-02-16
strconv: fix of atoi() and its tests.

Bruno-Vdr:
Revwrite of atoi(s string) !int

**Positive points:**

-Tests are more complete.
-Exact 32bits overflow/underflow
-Corrected x = (x * 10) + c (may fail on *10 and + c operation)
-Follow as much as possible V scanner with underscores (https://discord.com/channels/592103645835821068/592114487759470596/1339276917383364683).
-Underscore are supported whatever the lenght of the input string.
-Error messages are very similar to original atoi()
-Doesn't rely any more on parse_int() (https://discord.com/channels/592103645835821068/592114487759470596/1340260943325691904).

**Drawback:**

-Still considers type int as 32bits as stated in documentation (may change in the future).
-On previous implementation, atoi() called parse_int() method if string length was greater than 10 chars. Parse_int()  supports different radix than 10, unlike atoi. Code relying on this behaviour may be broken.


vlib/strconv/atoi.v
```diff
@@ -9,6 +9,8 @@ module strconv
 // max_u64 = u64(u64(1 << 63) - 1)
 const int_size = 32
 const max_u64 = u64(18446744073709551615)
+const int32_max = 2147483647
```
spytheman, line 12: `builtin` already has a const named `max_int` for that number

vlib/strconv/atoi.v
```diff
@@ -9,6 +9,8 @@ module strconv
...
+const int32_min = -2147483648
```
spytheman, line 13: there is `min_int` as well in `builtin`

spytheman: also `min_i32` and `max_i32` (that will not change, even when `int` becomes 64 bit)

vlib/strconv/atoi.v
```diff
@@ -218,34 +220,77 @@ pub fn parse_int(_s string, base int, _bit_size int) !i64 {
...
+			underscored = true
+			continue // Skip underscore
+		} else {
+			if c > 9 {
 				return error('strconv.atoi: parsing "${s}": invalid syntax')
 			}
-			n = n * 10 + int(ch)
+			underscored = false
+			x = safe_mul10_32bits(x) or { return error('strconv.atoi: parsing "${s}": ${err}') }
+			x = safe_add_32bits(x, int(c * sign)) or {
```
spytheman, line 264: imho do the intermediate calculations with i64, and only do one check at the end of the loop, right before the return of the result

Bruno-Vdr: I did this with 32bits machines in mind, to avoid using soft 64bit arithmetic. I plan to implement later atoi 8/16/32/64, the problem will exist then on 64bits. Solving the problem of under/overflow with a maximum type length, had to be done. But if you want I can implements it using i64. Should I ?

spytheman: I am concerned about the performance for the most common case, which is currently 64 bit machines.

Bruno-Vdr: Ok, I'm on it.

spytheman: imho use i64 (or just one higher size if you prefer), for the cases of numbers that are < 64 bit. See also `vlib/strconv/utilities.v` and `vlib/strconv/structs.v`, for the already defined operations on `Uint128` for the 64 bit case.

spytheman: soft 64bit arithmetic on 32 bit machines, is still preferable in terms of performance than checking for overflows on each operation in the loop, generating errors, and then handling them in the caller (in the loop)

vlib/strconv/atoi.v
Delta456, line 1: Can we have an explanation with each error message instead of the generic message `invalid syntax s`?

spytheman: what would you want them to be?

Delta456: Like how we have different error messages when using `_` in the scanner stage.

Bruno-Vdr: I return now better error messages, but still simple. Feel free to make them clearer. This change may break code that rely on parsing theses messages (if any).

Delta456: It is much better now!

vlib/strconv/atoi.v
```diff
@@ -218,34 +218,56 @@ pub fn parse_int(_s string, base int, _bit_size int) !i64 {
...
+		c := s[i] - `0`
+		if c == 47 { // 47 = Ascii(`_`) -  ascii(`0`) = 95 - 48.
+			if underscored == true { // Two consecutives underscore
+				return error('strconv.atoi: parsing "${s}": consecutives underscores are not allowed')
+			}
+			underscored = true
+			continue // Skip underscore
+		} else {
+			if c > 9 {
+				return error('strconv.atoi: parsing "${s}":  invalid radix 10 character')
```
spytheman, line 259:
```suggestion
				return error('strconv.atoi: parsing "${s}": invalid radix 10 character')
```

vlib/strconv/atoi.v
```diff
@@ -218,34 +218,56 @@ pub fn parse_int(_s string, base int, _bit_size int) !i64 {
 }
 
 // atoi is equivalent to parse_int(s, 10, 0), converted to type int.
+// It follows V scanner as much as observed.
+// @[direct_array_access]
```
spytheman, line 222:
```suggestion
```

vlib/strconv/atoi.v
```diff
@@ -9,8 +9,6 @@ module strconv
 // max_u64 = u64(u64(1 << 63) - 1)
 const int_size = 32
 const max_u64 = u64(18446744073709551615)
```
JalonSolov, line 11: This is also defined in builtin (vlib/builtin/int.v, line 64).

spytheman: yes, it can be safely deleted

vlib/strconv/atoi.v
```diff
@@ -218,34 +218,55 @@ pub fn parse_int(_s string, base int, _bit_size int) !i64 {
...
-				// return 0, &NumError{fnAtoi, s0, ErrSyntax}
-				return error('strconv.atoi: parsing "${s}": invalid syntax')
+	}
+
+	if s.len - start_idx < 1 {
+		return error('strconv.atoi: parsing "${s}": invalid syntax')
+	}
+
+	if s[start_idx] == `_` || s[s.len - 1] == `_` {
+		return error('strconv.atoi: parsing "${s}": invalid syntax')
```
JalonSolov, line 242:
```suggestion
		return error('strconv.atoi: parsing "${s}": values cannot start or end with underscores')
```

spytheman:
To me, the PR looks good, and ready to merge, after the CI jobs pass.
Excellent work @Bruno-Vdr 🙇🏻‍♂️ .

spytheman (approved):

vlib/strconv/atoi.v
```diff
@@ -218,34 +218,55 @@ pub fn parse_int(_s string, base int, _bit_size int) !i64 {
...
-		mut n := 0
-		for i in start_idx .. s.len {
-			ch := s[i] - `0`
-			if ch > 9 {
-				// return 0, &NumError{fnAtoi, s0, ErrSyntax}
-				return error('strconv.atoi: parsing "${s}": invalid syntax')
+	}
+
+	if s.len - start_idx < 1 {
+		return error('strconv.atoi: parsing "${s}": invalid syntax')
```
JalonSolov, line 238:
```suggestion
		return error('strconv.atoi: parsing "${s}": no number after sign')
```

PR #23743, opened 2025-02-16
native: fix convert_int_to_string, add comments, more verbose

Eliyaan:
Adds comments to help understanding the code.
Adds more `c.g.println` calls to have an easier time in verbose native mode.
Fixes the new test by terminating the buffer used in `convert_int_to_string` with `\0` because it is assumed after when reversing it and when priting it (it tries to find it's lenght by searching `\0`). (I did not know how to name my test function.)
Adds CLD when reversing the string to prevent the case where DF would be set and REPNE SCAS would do the opposite of what is expected.

### DIFF
--- a/vlib/v/gen/native/amd64.v
+++ b/vlib/v/gen/native/amd64.v
@@ -466,7 +466,7 @@ fn (mut c Amd64) jmp(addr i32) i32 {
 	c.g.write8(0xe9)
 	pos := c.g.pos()
 	c.g.write32(addr) // 0xffffff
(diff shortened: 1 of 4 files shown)


Eliyaan:
I fixed something but did not commit it yet

Eliyaan:
I dont think spydr is the good person to review this PR as he is busy and not working on this backend anymore.

JalonSolov (approved):
All tests run fine here.  Changes seem simple enough.

spytheman (approved):
Excellent work.

PR #23747, opened 2025-02-17
native: fix int prints

Eliyaan:
Fix `int_to_string` overwriting other variables on the stack while creating the string representation of the integer.
(Some of the tests added (the for loop) were going on an infinite loop before.)
Make the size of the lenght variable of arrays allocated with `allocate_array` independent of the size of the stored items, now 32bits.


vlib/v/gen/native/amd64.v
```diff
@@ -1323,6 +1323,7 @@ fn (mut c Amd64) gen_print(s string, fd i32) {
 	c.g.println('; print }')
 }
 
+// write a string of size n stored in r to fd
```
spytheman, line 1326:
```suggestion
// gen_print_reg writes a string of size n stored in r to fd
```

vlib/v/gen/native/gen.v
```diff
@@ -859,7 +859,7 @@ fn (mut g Gen) allocate_string(s string, opsize i32, typ RelocType) i32 {
 // allocates a buffer variable: name, size of stored type (nb of bytes), nb of items
 fn (mut g Gen) allocate_array(name string, size i32, items i32) i32 {
 	g.println('; allocate array `${name}` item-size:${size} items:${items}:')
-	pos := g.code_gen.allocate_var(name, size, items) // store the lenght of the array on the stack
+	pos := g.code_gen.allocate_var(name, 4, items) // store the lenght of the array on the stack in a 4 byte var
```
spytheman, line 862:
```suggestion
	pos := g.code_gen.allocate_var(name, 4, items) // store the length of the array on the stack in a 4 byte var
```

spytheman: or just `len`

vlib/v/gen/native/tests/print.vv
```diff
@@ -110,6 +110,18 @@ fn test_intermediate() {
 	println(c)
 }
 
+fn test_for_range() {
+	for i in 0 .. 10 {
+		println(i)
+	}
```
spytheman, line 116: add also a separate loop for `for i in 7 .. 9 {`
and also a separate loop for `for i in -2 .. 3 {`
to check for more common cases

spytheman (approved):
Excellent work.

Eliyaan:
Sadly the for loop from -2 to 3 does not work yet

spytheman:
> Sadly the for loop from -2 to 3 does not work yet

no problem, just add it as a commented block + a TODO

spytheman:
oh, you did already; excellent

PR #23752, opened 2025-02-17
native: for statement : add support for all expressions handled by g.condition

Eliyaan:
Change the way `for statement's` conditions are handled:
Replace manual match by `g.condition` (like cgen) to avoid duplicate work. (now for statements support all the expressions that are already supported by conditions)
Add tests and move tests in forin.vv to for.vv because they concern all types of for loops not just for in.
Add more comments and prints.


spytheman (approved):
Excellent work.

PR #23753, opened 2025-02-17
native: fix inc and improve support for i32

Eliyaan:
Replace the inc implementation by the INC instruction (previously ADD 1) and make it adapt depending on var size.
Add i32 in matches and add right size for i32 in get_type_size.
Add tests.


spytheman (approved):
Excellent work 🚀 .

PR #23755, opened 2025-02-18
native: negative integer literals

Eliyaan:
Use 64 bits instruction for all the mov64 registers and add the same function for unsigned ints.
Support negative integer literals.
Throw an native error when printing unsigned integers as it is not supported.
Add tests for negative integer literals.


vlib/v/gen/native/amd64.v
```diff
@@ -532,37 +532,66 @@ fn (mut c Amd64) mov32(reg Amd64Register, val i32) {
...
+		else {
+			eprintln('unhandled mov64u ${reg}')
+		}
+	}
+	c.g.write64(val)
+	c.g.println('mov64 ${reg}, ${val}')
+}
+
+fn (mut c Amd64) mov64(reg Register, val i64) {
+	// see AMD64 Architecture Programmer's Manual Volume 3 about the MOV instruction to have a detailed explaination
```
spytheman, line 571:
```suggestion
	// see AMD64 Architecture Programmer's Manual Volume 3 about the MOV instruction to have a detailed explanation
```

vlib/v/gen/native/tests/for.vv
```diff
@@ -33,6 +33,9 @@ fn simple_for_test() {
 		println('loop7')
 		i++
 	}
+	for j in -2 .. 3 {
+		println('loop8')
```
spytheman, line 37: can it print `j` in the loop too?

spytheman (approved):
Excellent work.

medvednikov:
Wonderful!

PR #23757, opened 2025-02-18
strconv : Addition of Atoi8/16/32/64 helper, with their test.

Bruno-Vdr:
Added atoi8, atoi16, atoi32 and atoi64 method. Added respective tests.Refactor some atoi() code to avoid code duplication. atoi64() uses a slightly different code, to keep using i64 types. Spytheman suggested to use math.i128 aritmetic. I didn't do it, but this can be done by anyone else.
I didn' t replicate Spytheman' s work around for MSVC compiler (msvc has a bug that treats just i64(min_int) as 2147483648).

### DIFF
--- a/vlib/strconv/atoi_test.v
+++ b/vlib/strconv/atoi_test.v
@@ -83,8 +83,8 @@ fn test_atoi() {
 		StrInt{'123_456_789', 123456789},
 		StrInt{'00000006', 6},
 		StrInt{'0_0_0_0_0_0_0_6', 6},
-		StrInt{'2147483647', 2147483647}, // Signed 32bits max.
-		StrInt{'-2147483648', -2147483648}, // Signed 32bits min.
+		StrInt{'2147483647', max_int},
+		StrInt{'-2147483648', min_int},
 	]
 
 	// Check that extracted int value matches its string.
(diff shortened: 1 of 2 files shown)


vlib/strconv/atoi_test.v
```diff
@@ -1,30 +1,90 @@
...
+		StrInt{'-1', -1},
+		StrInt{'0', 0},
+		StrInt{'+3_14159', 314159},
+		StrInt{'-1_00_1', -1001},
+		StrInt{'-1_024', -1024},
+		StrInt{'123_456_789', 123456789},
+		StrInt{'00000006', 6},
+		StrInt{'0_0_0_0_0_0_0_6', 6},
+		StrInt{'2147483647', 2147483647}, // Signed 32bits max.
+		StrInt{'-2147483648', -2147483648}, // Signed 32bits min.
```
Delta456, line 87: I feel it is good to have all cases for the sake of it. No need to remove.

Bruno-Vdr: Some tests have been moved. I think they don' t need to be redundant.

Delta456: Oh right... Just checked it is moved above...

Bruno-Vdr: I tried to provide specifics tests for each functions,  and avoid testing case already checked by another function call before.  [AFK for next 16h].

spytheman (approved):
Excellent work.

PR #23766, opened 2025-02-19
strconv : atou, atou8/16/32/64 function added, with tests.

Bruno-Vdr:
New file in strconv module: atou.v and atou_test.v.

atou / atou32 are provided to keep implementation symmetric with atoi / atoi32.
I choose to partially duplicate code, to avoid parametrizing atoi_common_check
and atoi_common for unsigned case.  This choice let the code simpler.

### DIFF
--- a/vlib/strconv/atoi_test.v
+++ b/vlib/strconv/atoi_test.v
@@ -5,7 +5,7 @@ struct StrInt { // test struct
 	int_value int
 }
 
-// test what should be catched by atoi_common_check
+// test what should be caught by atoi_common_check
 fn test_common_check() {
 	// Parsing of these strings should fail on all types.
 	ko := [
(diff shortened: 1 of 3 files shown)


vlib/strconv/atoi_test.v
```diff
@@ -1,11 +1,11 @@
 import strconv
 
-struct StrInt { // Inner test struct
+struct StrInt { // test struct
 	str_value string
 	int_value int
 }
 
-// test what should be catch by atoi_common_check
+// test what should be catched by atoi_common_check
```
medvednikov, line 8: caught

spytheman:
The remaining CI failures are unrelated, and fixed on master.

spytheman (approved):
Excellent work.

Issue #23773, opened 2025-02-20
alias + generics inconsistent compiler behaviour


pd-giz-dave:
### Describe the bug

Arithmetic on an alias of a numeric type compiles OK but not if the operation is buried in a generic function.

### Reproduction Steps

Attempt to compile this:

```
module main
type MyAlias = f32
const two = MyAlias(2)
fn mul[T](a T, b T) T {
	return a * b
}
fn main() {
	println(mul[MyAlias](two,two))  //compile fail
	println(mul[f32](two,two))  //compile OK - prints 4.0
	println(two * two)  //compile OK - prints 4.0
}
```

### Expected Behavior

Expect it to compile and run.

### Current Behavior

Get a compile error of:

```
compiler_bugs/main.v:5:9: error: undefined operation `MyAlias` * `MyAlias`
    3 | const two = MyAlias(2)
    4 | fn mul[T](a T, b T) T {
    5 |     return a * b
      |            ~~~~~
    6 | }
    7 | fn main() {
```

### Additional Information/Context

None

### V version

V 0.4.9 6d017f3

### Environment details (OS name and version, etc.)

```
~/precious/fellsafe/v git:[master]
v doctor
|V full version      |V 0.4.9 3108820.6d017f3
|:-------------------|:-------------------
|OS                  |linux, "EndeavourOS Linux"
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 5800H with Radeon Graphics
|Memory              |14.3GB/30.75GB
|                    |
|V executable        |/home/dave/v/v
|V last modified time|2025-02-20 18:16:39
|                    |
|V home dir          |OK, value: /home/dave/v
|VMODULES            |OK, value: /home/dave/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/dave/precious/fellsafe/v
|                    |
|Git version         |git version 2.48.1
|V git status        |weekly.2024.45-756-g6d017f3a
|.git/config present |true
|                    |
|cc version          |cc (GCC) 14.2.1 20240910
|gcc version         |gcc (GCC) 14.2.1 20240910
|clang version       |clang version 19.1.7
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 0134e9b9
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.40
```

> [!NOTE]
[...]

PR #23774, opened 2025-02-20
checker: allow generic operators to be called in fn 

Delta456:
Fixes https://github.com/vlang/v/issues/23773


spytheman (approved):
Excellent work.

PR #23778, opened 2025-02-20
native: cleanup assign, fix nested struct

Eliyaan:
Cleanup the `assign_stmt` function (and some functions called in it)
Add file:line on some n_error calls for faster debugging
Fix nested struct expressions, ex: `foo.bar.a` did not work
Add tests for fixed behavior and for more already supported behavior


JalonSolov (approved):
Structs and enums, too?  Excellent!

vlib/v/gen/native/amd64.v
```diff
@@ -674,6 +674,7 @@ fn (mut c Amd64) mov_reg_to_var(var Var, r Register, config VarConfig) {
 				}
 				Register {
 					// TODO
+					c.g.n_error('${@FILE}:${@LINE} unsupported Ident Register in mov_reg_to_var')
```
spytheman, line 677:
```suggestion
					c.g.n_error('${@LOCATION} unsupported Ident Register in mov_reg_to_var')
```

spytheman: `@LOCATION` will contain both the file, the line, and the method/function name

vlib/v/gen/native/amd64.v
```diff
@@ -2085,69 +2089,148 @@ fn (mut c Amd64) assign_var(var IdentVar, raw_type ast.Type) {
...
-					c.mov_reg_to_var(ident, Amd64Register.rax)
-				}
-				else {
-					tn := node.left[i].type_name()
-					dump(node.left_types)
-					c.g.n_error('unhandled assign type: ${tn}')
-				}
-			}
+			c.mov(Amd64Register.rax, i32(int_lit.val.int()))
+			dump(left)
```
spytheman, line 2126:
```suggestion
```

If it is already handled, there is no need to dump/print anymore

vlib/v/gen/native/expr.v
```diff
@@ -445,11 +446,17 @@ fn (mut g Gen) gen_print_from_expr(expr ast.Expr, typ ast.Type, name string) {
...
+	g.println('; .${expr.field_name} {')
 	main_reg := g.code_gen.main_reg()
 	g.expr(expr.expr)
 	offset := g.get_field_offset(expr.expr_type, expr.field_name)
-	g.code_gen.add(main_reg, offset)
-	g.code_gen.mov_deref(main_reg, main_reg, expr.typ)
+	if offset != 0 {
+		g.code_gen.add(main_reg, offset)
+	}
+	if expr.next_token != token.Kind.dot { // the deref needs to be on the last selector (that has no . after it)
```
spytheman, line 456:
```suggestion
	if expr.next_token != .dot { // the deref needs to be on the last selector (that has no . after it)
```

spytheman: The compiler should already know that .next_token is `token.Kind`.

vlib/v/gen/native/tests/assign.vv
```diff
@@ -18,15 +18,29 @@ fn test_int() {
...
-	b += 0.5
+	b += 1.5
+	b -= 1.0
 	b *= 4
 	b /= 2
 
+	b += a
+	b -= a
+	b *= a
+	b /= a
```
spytheman, line 42: please put additional prints between the operations, and modify asign.vv.out as well
some of them cancel out, but that can also mean that *both* operations failed, and without the intermediate prints, there will be no way to check it automatically :-|

spytheman: You can start with 10.0 instead of 1.0, and then scale up all factors by 10, in order to have changing int results, if printing f64 does not work yet.

Eliyaan: Would asserts (between each line) be good?

spytheman: Yes, they can check that too, but I am not sure if `==` is implemented for f64 yet.

If it is, then yes.

If it is not, then casting to int and then printing is a better temporary fix, that will still guarantee that regressions are not silent.

spytheman: Although casting to int, and then comparing with == to another int, may already work in the asserts 🤔 .

Eliyaan:
Oh true, == is not implemented for floats. `assert 2.0 == 1.0` is passing...

vlib/v/gen/native/tests/float.vv
```diff
@@ -28,6 +28,17 @@ fn float_assign_test() {
...
+	mut b := &a
+        unsafe{*b += 2.5}
+        assert a == 3.5
+        unsafe{*b -= 1.25}
+        assert a == 2.25
+        unsafe{*b *= 6.0}
+        assert a == 13.5
+        unsafe{*b /= 4.5}
+        assert a == 3.0	
 }
```
spytheman, line 42: this may need a change too

spytheman: however it can also serve as a check when the asserts and == work for f64, and the rest of the tests can cross check it, so imho it is good to merge as it is

spytheman (approved):
Excellent work.

vlib/v/gen/native/tests/float.vv
Eliyaan: True, I'll change that for the next PR

Eliyaan: It will require to change all the tests in the file, as it is using asserts between floats for all checks.

Eliyaan: I may wait until assert between floats are fixed as some of the tests would be interesting to keep as such.

PR #23786, opened 2025-02-22
native: add support for enums of different types

Eliyaan:
Add support for enums with a user-chosen type like `enum MyEnum as u8`
Instead of duplicating functions for u64 and i64, use a sumtype
Add tests for enums of different types

### DIFF
--- a/vlib/v/gen/native/amd64.v
+++ b/vlib/v/gen/native/amd64.v
@@ -532,42 +532,7 @@ fn (mut c Amd64) mov32(reg Amd64Register, val i32) {
 	c.g.println('mov32 ${reg}, ${val}')
 }
 
-// moves unsigned 64 bits, explained better in mov64
-fn (mut c Amd64) mov64u(reg Register, val u64) {
-	match reg as Amd64Register {
-		.rax {
-			c.g.write8(0x48)
-			c.g.write8(0xb8)
-		}
-		.rcx {
-			c.g.write8(0x48)
-			c.g.write8(0xb9)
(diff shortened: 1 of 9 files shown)


vlib/v/gen/native/amd64.v
```diff
@@ -3387,51 +3377,83 @@ pub fn (mut c Amd64) allocate_var(name string, size i32, initial_val i32) i32 {
...
+				}
+			}
 		}
 		4 {
 			// DWORD
 			c.g.write8(0xc7)
 			c.g.write8(i32(0x45) + far_var_offset)
+			// Generate N in `[rbp-N]`
+			if is_far_var {
+				c.g.write32(i32((0xffffffff - i64(n) + 1) % 0x100000000))
```
spytheman, line 3429: hm, why % 0x100000000, instead of masking with `& (0x100000000 - 1)` ?

spytheman (approved):
Good work, but in general, please keep in mind, that it is better to split refactoring PRs (especially ones that affect a lot of code), from ones that add functional changes/features/bugfixes.

Mixing too many things in the same one, makes many things harder than they could be otherwise, and increases the chances of conflicts with other PRs.

Eliyaan:
Okay I'll keep that in mind.

vlib/v/gen/native/amd64.v
Eliyaan: I did not touch this part of the code as I dont know much about machine code addressing

PR #23808, opened 2025-02-25
native: fix unsigned and signed int comparison

Eliyaan:
Fix unsigned / signed integer comparison
Add test coming from vlib/v/tests with prefix `vtest_`

I took the test from the cgen backend because it's quite complete for edge cases


spytheman (approved):
Excellent work.

PR #23815, opened 2025-02-26
fix strconv.atof64(), return error when detect extra char after number

kbkpbot:
This PR help detect extra char after number during the `atof64()` conversion.

`strconv.atof64('123xyz')` will get error `extra char after number` now.

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -636,13 +636,13 @@ pub fn (s string) i64() i64 {
 // f32 returns the value of the string as f32 `'1.0'.f32() == f32(1)`.
 @[inline]
 pub fn (s string) f32() f32 {
-	return f32(strconv.atof64(s) or { 0 })
+	return f32(strconv.atof64(s, allow_extra_chars: true) or { 0 })
 }
 
 // f64 returns the value of the string as f64 `'1.0'.f64() == f64(1)`.
 @[inline]
 pub fn (s string) f64() f64 {
-	return strconv.atof64(s) or { 0 }
+	return strconv.atof64(s, allow_extra_chars: true) or { 0 }
 }
 
 // u8_array returns the value of the hex/bin string as u8 array.
--- a/vlib/strconv/atof.c.v
+++ b/vlib/strconv/atof.c.v
@@ -393,7 +393,7 @@ fn converter(mut pn PrepNumber) u64 {
 
 @[params]
 pub struct AtoF64Param {
-pub mut:
+pub:
 	no_extra_char bool // exact convertion, no extra char after number
 }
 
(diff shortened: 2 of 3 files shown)


spytheman (approved):
Excellent work.

spytheman:
Can that be expanded to the other atoXXX functions (in another PR)?

spytheman:
The problem is that creates discrepancy between the default behaviors and supported options between otherwise similar functions.

JalonSolov:
I agree.  All the `ato*` routines should act the same.  I'd prefer if `atof64` worked like the original `atoi` - fail if there are non-float characters either before or after the digits.

("non-float" instead of "non-digit", since `e` is a valid part of a float... just not as the very first or very last character.)

spytheman:
```c
#include <stdio.h>
#include <stdlib.h>
int main(void) {
  char *str = "123xyz";
  double res = atof(str);
  printf("res = %5.4f\n", res);
}
```
```
#0 08:14:48 ^ master ~/code/v>gcc -o x x.c
#0 08:14:56 ^ master ~/code/v>./x
res = 123.0000
```

spytheman:
On my system it is documented in its man page as:
```
DESCRIPTION
       The atof() function converts the initial portion of the string pointed to by nptr to double.
       The behavior is the same as

           strtod(nptr, NULL);

       except that atof() does not detect errors.

RETURN VALUE
       The converted value.
```

spytheman:
atoi is similar:
```

DESCRIPTION
       The atoi() function converts the initial portion of the string 
       pointed to by nptr to int. The behavior is the same as

           strtol(nptr, NULL, 10);

       except that atoi() does not detect errors.

       The  atol()  and  atoll()  functions  behave the same as atoi(), 
       except that they convert the initial portion of the string to
       their return type of long or long long.

RETURN VALUE
       The converted value or 0 on error.
```

spytheman:
To clarify - I am not arguing about failing or non failing by default.

I just think that all of the strconv.atoXXX functions should have *the same default*, and also have the same options that make them behave in the opposite way.

If you introduce (as you do in the current state of the PR):
```v
@[params]
pub struct AtoF64Param {
pub:
	allow_extra_chars bool // allow extra characters after number
}
```

imho similar params structs should be added, for parametrising the other strconv.atoXXX functions as well.

spytheman:
```v
import strconv

a := strconv.atoi('123xyz') or { dump(err) 0 }
dump(a)

f := strconv.atof64('123xyz') or { dump(err) 0 }
dump(f)
```

On master, that produces inconsistent behavior:
```
[a.v:3] err: strconv.atoi: parsing "123xyz": invalid radix 10 character
[a.v:4] a: 0
[a.v:7] f: 123.0
```

On this PR, both work in a similar way, which is good (that makes the whole family more mnemonic):
```
[a.v:3] err: strconv.atoi: parsing "123xyz": invalid radix 10 character
[a.v:4] a: 0
[a.v:6] err: extra char after number
[a.v:7] f: 0.0
```

Issue #23802, opened 2025-02-24
Windows: os.get_raw_line trim right `\r\n` interpolation unhandled exception


jorgeluismireles:
### Describe the bug

New bug derived from issue https://github.com/vlang/v/issues/23764

We read a file line by line with `os.get_raw_line` and use `.trim_right('\r\n')` and after print interpolated we get an **Unhandled Exception** with rows complex and large. Tested on windows with both `tcc` and `mingw`.

### Reproduction Steps

```v
module main

import os

fn main() {
	mut i:= 0
	for {
		i += 1
		mut line := os.get_raw_line()
		if line.len == 0 {
			break
		}
		line2 := line.trim_right('\r\n')
		// println('${i}: ${line2.len}') OK prints n: 1034 n=1,2,3
		println('${i}: ${line2}') // error
	}
}
```
Note: the size of the line is calculated and printed ok but not he whole line content.

### Expected Behavior

Print three lines with text
```
1: ...
2: ...
3: ...
```

### Current Behavior

We pass the file using window command `C:\type {file} | {v app}`

```
C:\Users\jmireles\issues>type yyy.txt | v run issue.v
[...]

Issue #23764, opened 2025-02-19
Windows: os.get_raw_line - buffer overflow


DamianFekete:
Root cause described here: https://github.com/vlang/v/issues/23764#issuecomment-2671750788

---

### Describe the bug

On Windows 10 (no issue on WSL) I get a `RUNTIME ERROR: invalid memory access`.

**LE.**

* I don't think this is related to the parsing of jsons when the Log struct doesn't have the fields defined. I've removed them to make the test case smaller, this is not the original app
* If I make any of these changes, the error is not reproducible (but the memory may still be corrupted):
  * Remove the print from the `or { }` block
  * Read from the standard input but parse another string (with the same value) to json.decode

### Reproduction Steps

Create a file yyy:

```
[...]

PR #23816, opened 2025-02-26
os.get_raw_line fix buffer overflow under Windows

kbkpbot:
Fix issue #23802, #23764


Delta456:
Is this possible to test?

PR #23829, opened 2025-02-28
checker: add option type inference on if expr

### LINKED ISSUE #23827: Optional inference in if expression
### Describe the feature

The compiler should be able to infer an optional type without cast, in if expression with an optional in first or second branch.

v := if cond { **none** } else { 1 }
v := if cond { 1 } else { **none** }

Should work in place of :

v:= if cond { **?int(none)** } else { 1 }
v:= if cond { **?int(1)** } else { none }

### Use Case

This feature would allow a a clearer, simple and a more intuitive syntax.

### Proposed Solution

_No response_

### Other Information

Spytheman & Tcn  remind us that none has its own type and that the compiler should error if both branches are none.

### Acknowledgements

### Version used

V 0.4.9 5f5e48e

|V full version      |V 0.4.9 89d1aac.5f5e48e
|OS                  |linux, Debian GNU/Linux trixie/sid
|Processor           |24 cpus, 64bit, little endian, 12th Gen Intel(R) Core(TM) i9-12950HX
|Memory              |4.33GB/15.31GB
|                    |
|V executable        |/home/vedder/v/v
|V last modified time|2025-02-28 08:28:44
|                    |
|V home dir          |OK, value: /home/Bruno-VDR/v
|VMODULES            |OK, value: /home/Bruno-VDR/.vmodules
|Current working dir |OK, value: /home/Bruno-VDR
|                    |
|Git version         |git version 2.47.2
|V git status        |5f5e48ea
|.git/config present |true
|                    |
|cc version          |cc (Debian 14.2.0-16) 14.2.0
|gcc version         |gcc (Debian 14.2.0-16) 14.2.0
|clang version       |N/A
|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.40-7) 2.40

> [!NOTE]
>

felipensp:
Fix #23827

### DIFF
--- a/vlib/v/checker/if.v
+++ b/vlib/v/checker/if.v
@@ -510,6 +510,15 @@ fn (mut c Checker) if_expr(mut node ast.IfExpr) ast.Type {
 							stmt.typ = ast.error_type
 							continue
 						}
+						if (node.typ == ast.none_type && stmt.typ != ast.none_type)
+							|| (stmt.typ == ast.none_type && node.typ != ast.none_type) {
+							node.typ = if stmt.typ != ast.none_type {
+								stmt.typ.set_flag(.option)
+							} else {
+								node.typ.set_flag(.option)
+							}
+							continue
+						}
 						c.error('mismatched types `${c.table.type_to_str(node.typ)}` and `${c.table.type_to_str(stmt.typ)}`',
 							node.pos)
 					} else {
--- a/vlib/v/tests/options/option_if_infer_test.v
+++ b/vlib/v/tests/options/option_if_infer_test.v
@@ -1,0 +1,8 @@
+fn test_main() {
+	v := if true { none } else { 1 }
+	assert '${v}' == 'Option(none)'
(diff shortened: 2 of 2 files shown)


vlib/v/checker/if.v
```diff
@@ -510,6 +510,15 @@ fn (mut c Checker) if_expr(mut node ast.IfExpr) ast.Type {
...
 						}
+						if (node.typ == ast.none_type && stmt.typ != ast.none_type)
+							|| (stmt.typ == ast.none_type && node.typ != ast.none_type) {
+							node.typ = if stmt.typ != ast.none_type {
+								stmt.typ.set_flag(.option)
+							} else {
+								node.typ.set_flag(.option)
+							}
+							continue
+						}
```
spytheman, line 521: it needs to handle the case of:
```v
    n := if true { none } else { none }
```
too (imho it should be a checker error; it is currently a cgen one)

spytheman (approved):
Very good work.

PR #23832, opened 2025-02-28
checker: add missing check for IfExpr and MatchExpr with no valid type

felipensp:
Fix current crash when using:

```v
fn main() {
	a := if true { none } else { none }
}
```

### DIFF
--- a/vlib/v/checker/tests/match_expr_with_none_only.out
+++ b/vlib/v/checker/tests/match_expr_with_none_only.out
@@ -4,7 +4,7 @@ vlib/v/checker/tests/match_expr_with_none_only.vv:2:2: warning: unused variable:
       |     ^
     3 |         1 { none }
     4 |         else { none }
-vlib/v/checker/tests/match_expr_with_none_only.vv:2:7: error: invalid if expression, must supply at least one value other than `none`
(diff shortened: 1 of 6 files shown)


vlib/v/checker/tests/match_expr_with_none_only.out
```diff
@@ -0,0 +1,12 @@
+vlib/v/checker/tests/match_expr_with_none_only.vv:2:2: warning: unused variable: `a`
+    1 | fn main() {
+    2 |     a := match 1 {
+      |     ^
+    3 |         1 { none }
+    4 |         else { none }
+vlib/v/checker/tests/match_expr_with_none_only.vv:2:7: error: invalid if expression, must supply at least one value other than `none`
```
JalonSolov, line 7:
```suggestion
vlib/v/checker/tests/match_expr_with_none_only.vv:2:7: error: invalid match expression, must supply at least one value other than `none`
```

felipensp: Oh, I've fixed on same time than you. Thanks for reviewing. 👍🏻

spytheman (approved):
Excellent work.

PR #23833, opened 2025-02-28
checker: fix MatchExpr type inferring when `none` is used  

### LINKED ISSUE #23831: Optional inference in match expression
### Describe the feature

This is an extension of issue https://github.com/vlang/v/issues/23827

Infer as possible the `match` cases for optionals

### Use Case

For returning `none` we need explicity set `?int(none)`

```v
fn a(cond int) {
	v:= match cond {
		0 {
			?int(none)
		}
		1 {
			1
		}
		else {
			1000
		}
	}
	println('${v}')
}

a(2)
```

### Proposed Solution

Infer the `none` from `?int(none)` to have this simpler form:

```v
fn a(cond int) {
	v:= match cond {
		0 {
			none
		}
		1 {
			1
		}
		else {
			1000
		}
	}
	println('${v}')
}

a(2)
```

### Other Information

_No response_

### Acknowledgements

### Version used

V 0.4.9 5f5e48e

https://play.vlang.io/p/04cc0c924c

> [!NOTE]
>

felipensp:
Fix #23831

### DIFF
--- a/vlib/v/tests/options/option_match_none_test.v
+++ b/vlib/v/tests/options/option_match_none_test.v
@@ -11,4 +11,11 @@ fn test_main() {
 		else { 1 }
 	}
 	assert '${b}' == 'Option(none)'
+
+	c := match 2 {
+		0 { 1 }
+		2 { 2 }
+		else { none }
+	}
+	assert '${c}' == 'Option(2)'
 }
--- a/vlib/v/tests/options/option_match_none_test.v
+++ b/vlib/v/tests/options/option_match_none_test.v
@@ -1,0 +1,14 @@
+fn test_main() {
+	a := match 1 {
+		0 { none }
+		else { 1 }
+	}
+	assert '${a}' == 'Option(1)'
+
+	b := match 2 {
+		0 { none }
+		2 { none }
+		else { 1 }
+	}
+	assert '${b}' == 'Option(none)'
+}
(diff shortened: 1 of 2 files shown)


JalonSolov:
Make sure it fails with a similar message to the recent `if` change, if someone does this:
```v
b := match 2 {
	0 { none }
	2 { none }
	else { none }
}
```
In other words, just like `if`, `match` has to have at least one value that is something other than `none`.

felipensp:
> Make sure it fails with a similar message to the recent `if` change, if someone does this:
> 
> ```v
> b := match 2 {
> 	0 { none }
> 	2 { none }
> 	else { none }
> }
> ```
> 
> In other words, just like `if`, `match` has to have at least one value that is something other than `none`.

Oh yes, I've updated the PR to includes `match` as well.

JalonSolov:
Yep, I saw the other PR after adding that comment - that's why I deleted it.  :-)

vlib/v/tests/options/option_match_none_test.v
```diff
@@ -0,0 +1,14 @@
...
+	}
+	assert '${a}' == 'Option(1)'
+
+	b := match 2 {
+		0 { none }
+		2 { none }
+		else { 1 }
+	}
+	assert '${b}' == 'Option(none)'
+}
```
spytheman, line 14: please also add a case, where the number is the first branch, and the `none` is in some of the others

felipensp: Sure. Good catch. It really was not working, fixed now.

spytheman (approved):
Excellent work.

PR #23839, opened 2025-03-01
json: fix json.decode autofree codegen

### LINKED ISSUE #23834: [c compilation bug] when using -autofree with json library
<details>
<summary>V version: V 0.4.9 f83af8a, press to see full `v doctor` output</summary>

|V full version      |V 0.4.9 f83af8a
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 Pro 7735U with Radeon Graphics
|Memory              |7.32GB/30.12GB
|                    |
|V executable        |/home/polyface/git/v/v
|V last modified time|2025-03-01 09:25:10
|                    |
|V home dir          |OK, value: /home/polyface/git/v
|VMODULES            |OK, value: /home/polyface/.vmodules
|Current working dir |OK, value: /home/polyface/Documents/v/bug
|                    |
|V git status        |weekly.2025.09-23-gf83af8a1
|.git/config present |true
|                    |
|cc version          |cc (GCC) 14.2.1 20250207
|gcc version         |gcc (GCC) 14.2.1 20250207
|emcc version        |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg -autofree src/main.v && src/main`
```v
// main.v
module main

import json

fn main() {
	input := '{"methode":"test"}'
	decode_message(input) or {}
}

pub fn decode_message(msg string) !string {
	decoded_json := json.decode(BaseMessage, msg.str())!
	return decoded_json.str()
}

type BaseMessage = struct {
	methode string @[json: methode]
}

```

**What did you see?**
```
cc: /tmp/v_1000/main.01JN8GDSBDE2C9DZA5SNJJAHF4.tmp.c:6913: error: '_arg_expr_json_decode_1_177' undeclared
builder error: 
==================
```

A normal compilation. The bug only happens when passing -autofree to the compiler.

The bug seems to happen when you call .str() inside json.decode function. If you  do something like this below, the bug doesn't happen 
```v
s := msg.str()
decoded_json := json.decode(BaseMessage, s)!
```
> [!NOTE]
>

felipensp:
Fix #23834

### DIFF
--- a/vlib/json/tests/json_decode_autofree_test.v
+++ b/vlib/json/tests/json_decode_autofree_test.v
@@ -1,0 +1,19 @@
+// main.v
+module main
+
+import json
+
+fn test_main() {
+	input := '{"methode":"test"}'
+	decode_message(input) or {}
+	assert true
+}
+
+type BaseMessage = struct {
+	methode string @[json: methode]
+}
+
+pub fn decode_message(msg string) !string {
+	decoded_json := json.decode(BaseMessage, msg.str())!
+	return decoded_json.str()
+}
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -2477,9 +2477,10 @@ fn (mut g Gen) call_args(node ast.CallExpr) {
 			}
 		} else {
 			if use_tmp_var_autofree {
+				n := if node.name == 'json.decode' { i + 2 } else { i + 1 }
 				// TODO: copypasta, move to an inline fn
 				fn_name := node.name.replace('.', '_')
-				name := '_arg_expr_${fn_name}_${i + 1}_${node.pos.pos}'
+				name := '_arg_expr_${fn_name}_${n}_${node.pos.pos}'
 				g.write('/*af arg2*/' + name)
 			} else {
 				g.expr(arg.expr)
--- a/vlib/v/gen/c/testdata/autofree_json_decode.c.must_have
+++ b/vlib/v/gen/c/testdata/autofree_json_decode.c.must_have
@@ -1,0 +1,2 @@
+string _arg_expr_json_decode_2_271 = string_str(msg);
+cJSON* _t2 = json__json_parse(/*af arg2*/_arg_expr_json_decode_2_271);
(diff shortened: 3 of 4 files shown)


spytheman (approved):
Excellent work.
I've checked it manually, and the test indeed fails on master with -autofree, but compiles and runs cleanly with -autofree on this PR.

However, normally, the tests are not compiled with -autofree.
We do have some infrastructure for programs, that have to be compiled with specific options, but that requires moving the test to `vlib/v/gen/c/testdata/` and turning it into a .vv/.out pair, where the .vv file starts with `// vtest vflags: -autofree` .

For an example, see `vlib/v/gen/c/testdata/autofree_json_encode.vv` and make the test print the decoded result, so that could be then stored in the .out file, and compared later automatically.

You can run the test locally using `./v vlib/v/gen/c/coutput_test.v` to check that the new .vv/.out pair will be picked up, and that it will have the `-autofree` flag applied during compilation.

spytheman:
On a side note, I think that it may be a good idea, to extend the support for `// vtest vflags: ` to the normal test runner as well, so that such tests can stay where their logical place is, without having to be moved, just to take advantage of the logic in `vlib/v/gen/c/coutput_test.v` .

felipensp:
> Excellent work. I've checked it manually, and the test indeed fails on master with -autofree, but compiles and runs cleanly with -autofree on this PR.
> 
> However, normally, the tests are not compiled with -autofree. We do have some infrastructure for programs, that have to be compiled with specific options, but that requires moving the test to `vlib/v/gen/c/testdata/` and turning it into a .vv/.out pair, where the .vv file starts with `// vtest vflags: -autofree` .
> 
> For an example, see `vlib/v/gen/c/testdata/autofree_json_encode.vv` and make the test print the decoded result, so that could be then stored in the .out file, and compared later automatically.
> 
> You can run the test locally using `./v vlib/v/gen/c/coutput_test.v` to check that the new .vv/.out pair will be picked up, and that it will have the `-autofree` flag applied during compilation.

Oh right, I just missed this particularity for autofree test.

Issue #22464, opened 2024-10-09
veb, server drops connection, error parsing request


timurgordon:
### Describe the bug

Sometimes, the server drops connection after certain requests, logging `[veb] error parsing request: io.Eof:`. This is a difficult bug to isolate, as it happens almost randomly. A few refreshes and sometimes the request works again. Sometimes switching browsers works, until it doesn't. More so this is observed with post requests. The bug is experienced with different browser clients (tried with Firefox and Safari), and server running both in linux and macOS.

### Reproduction Steps

Run the following, try submitting the register form on different tabs, browsers, until the server hangs up. Sometimes it doesnt happen in a streak, sometimes it happens and lingers.

```
module main

import veb

pub struct App {}

pub struct Context {
	veb.Context
}

pub fn new() !&App {
	mut app := &App{}
	return app
}

pub fn run() ! {
	mut app := new()!
	veb.run[App, Context](mut app, 8090)
}

pub fn (app &App) index(mut ctx Context) veb.Result {
	return ctx.html('
<!DOCTYPE html>
<html lang="en" data-theme="light">
<body>
    <main>
        <h2>Login</h2>
        <form action="/register" method="POST">
            <label for="email">
                Use your Email to Register or Login.
                <input type="email" id="email" name="email" placeholder="Enter your email" required>
            </label>
            <button type="submit">Log In</button>
        </form>
    </main>
</body>
</html>	
')
}

@[post; '/register']
pub fn (mut app App) register(mut ctx Context) veb.Result {
	return ctx.text('authentication mail sent')
}

pub fn main() {
	run() or {panic(err)}
}
```

### Expected Behavior

For the server to be able to parse the request. If not, for the server to allow the developer to handle the error instead of spontaneously dropping connection. Consistency in behavior and error occurrence under the same conditions.

### Current Behavior

`veb` error when parsing incoming request, being thrown seemingly randomly.

`[veb] error parsing request: io.Eof:`

### Possible Solution

Could be a memory problem due to the random nature of the occurrences of the error. The buffered reader may have memory issues because the size of the buffer to read and parse the incoming request is set correctly, and the request is formatted correctly with the correct length header, so there shouldn't be an unexpected end of file. Or maybe a picoev error?

### Additional Information/Context

[...]

PR #23842, opened 2025-03-01
veb: fix "error parsing request: io.Eof" error when a request body is expected, but the data is not ready yet

Casper64:
Possible fix for #22464

Currently, if a client sends request with the Content-Length > 0, veb expects some data of the body to be ready after the headers are parsed in the same picoev callback, else it shows an `Eof` error.

It is possible that a client sends the headers first and the body has not been received yet. In this case veb should wait for the connection to provide more data, but it errors. I believe this is what's causing the randomness in the error message.

This pr changes the behaviour to wait for the body data, if the headers are parsed in the same piceov callback (callback that the connection has data available).

### DIFF
--- a/vlib/veb/tests/veb_test.v
+++ b/vlib/veb/tests/veb_test.v
@@ -96,6 +96,22 @@ fn test_a_simple_tcp_client_zero_content_length() {
 	assert watch.elapsed() < 1 * time.second
 }
 
+fn test_timeout_after_delayed_body() {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

medvednikov:
Thank you so much for fixing this!

syrmel:
Great fix; I could finally convert my remaining vweb to veb. Thanks!

Issue #23843, opened 2025-03-01
os.get_raw_line buffer overflow


DamianFekete:
After [the ](https://github.com/vlang/v/issues/23764) was recently closed I took another look at the method.

From https://github.com/vlang/v/blob/master/vlib/os/os.c.v#L506C1-L518C5

```v
			max_line_chars := 256
			mut old_size := max_line_chars * 2
			mut buf := malloc_noscan(old_size)
			h_input := C.GetStdHandle(C.STD_INPUT_HANDLE)
			mut bytes_read := u32(0)
			if is_atty(0) > 0 {
				x := C.ReadConsole(h_input, buf, max_line_chars * 2, voidptr(&bytes_read),
					0)
				if !x {
					return tos(buf, 0)
				}
				return string_from_wide2(&u16(buf), int(bytes_read))
			}
```

buf is 512 bytes long, but ReadConsole can read up 512 chars (of 2 bytes each), which means it will write up to 1024 bytes in buf.

From https://learn.microsoft.com/en-us/windows/console/readconsole

> The number of characters to be read. The size of the buffer pointed to by the lpBuffer parameter should be at least nNumberOfCharsToRead * sizeof(TCHAR) bytes.

Also os.get_raw_line doesn't handle user presses of Ctrl-Z (the Ctrl-D equivalent from Linux). See how this issue was handled in Go: https://github.com/golang/go/commit/610d522189ed3fcf0d298609a248a3283bde62cd

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #23846, opened 2025-03-02
os: fix get_raw_line() under windows

kbkpbot:
Fix issue #23843

1. read 1 byte/char from file pipe/console every time, and check for `\n` and `Ctrl+Z`.
2. add a `example/get_raw_line.v` as I don't know how to test it.

A discussion can be found at https://github.com/golang/go/commit/610d522189ed3fcf0d298609a248a3283bde62cd

### DIFF
--- a/vlib/os/os.c.v
+++ b/vlib/os/os.c.v
@@ -499,91 +499,72 @@ fn print_c_errno() {
 	eprintln('errno=${e} err=${se}')
 }
 
-fn to_utf8(input []u8, source_cp u32) string {
-	$if windows {
-		if source_cp == 65001 { // already in UTF-8
-			return unsafe { tos_clone(&input[0]) }
-		}
-		// Multi Byte -> Wide Char (UTF-16)
-		required_wchars := C.MultiByteToWideChar(C.CP_ACP, 0, input.data, input.len, C.NULL,
-			0)
-		mut wbuffer := []u16{len: required_wchars}
-		C.MultiByteToWideChar(C.CP_ACP, 0, input.data, input.len, wbuffer.data, required_wchars)
-
-		// Wide Char -> UTF-8
-		required_bytes := C.WideCharToMultiByte(C.CP_UTF8, 0, wbuffer.data, required_wchars,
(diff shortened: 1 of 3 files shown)


vlib/os/os.c.v
```diff
@@ -500,44 +500,72 @@ fn print_c_errno() {
...
-				return string_from_wide2(&u16(buf), int(bytes_read))
-			}
+			initial_size := 256 * wide_char_size
+			mut buf := malloc_noscan(initial_size)
+			mut capacity := initial_size
 			mut offset := 0
+
 			for {
+				required_space := offset + wide_char_size
+				if required_space > capacity {
```
spytheman, line 520: what happens is `required_space > capacity * 2`, i.e. not just bigger than `capacity`, and after the reallocation, the buffer will be still smaller than `required_space` ?

vlib/os/os.c.v
```diff
@@ -500,44 +500,72 @@ fn print_c_errno() {
...
+					}
 				}
+
+				offset += wide_char_size
+			}
+
+			return if is_console {
+				string_from_wide2(&u16(buf), offset / 2)
+			} else {
+				buf.vstring_with_len(offset)
```
spytheman, line 567: `string_from_wide2` will allocate a new string, and `buf` will remain allocated but unused/leaked.
`buf.vstring_with_len` in contrast will reuse the buffer for backing up the returned string, so `buf` will not leak.

I think that it will be cleaner, to unify that, and thus to use `buf.vstring_with_len(offset).clone()` here too, so that `buf` can be later safely freed in a `defer{}` block, put right after it is allocated above in the `mut buf := malloc_noscan(initial_size)` line.

vlib/os/os.c.v
kbkpbot: As `offset` increase 2 every loop at most, and as the  `capacity` >> 2, so the `required_space ` will never beyond `capacity * 2`, because `capacity` update it self to `capacity*2`.

spytheman (approved):
Excellent work.

PR #23851, opened 2025-03-03
cgen: fix codegen for match on return

### LINKED ISSUE #23661: C error when init struct fields with both local match and function with result
### Describe the bug

We initialize a struct with two fields. Field one with a local `match` and field two with a function returning a result. This combination can't be compiled.

### Reproduction Steps

```v
struct Bauds {
	standard bool
	bauds    u32
}

fn max(v u32) ! u32 {
	if v > 38400 {
		return error('too fast')
	}
	return v
}

fn new_bauds(bauds u32) ! &Bauds {
	return &Bauds{
		standard: match bauds {
			300,600,1200,2400,4800,9600,14400,19200,38400 { true }
			else { false }
		}
		bauds: max(bauds)!
	}
}

println(new_bauds(310)!)
```

### Expected Behavior

No C error

### Current Behavior

```
Running code...
/tmp/v_60000/../../../../../../box/code.v:15: error: ')' expected (got "_result_u32")
builder error: 
==================
```

### Possible Solution

Workarounds:

- Precalculate the first field with match before init.
- Propagate the error of the second field before init.

_No response_

### V version

V 0.4.9 b91bbad

https://play.vlang.io/p/6e15c8929d

> [!NOTE]
>

felipensp:
Fix #23661

### DIFF
--- a/vlib/v/gen/c/match.v
+++ b/vlib/v/gen/c/match.v
@@ -8,6 +8,9 @@ import v.util
 
 fn (mut g Gen) need_tmp_var_in_match(node ast.MatchExpr) bool {
 	if node.is_expr && node.return_type != ast.void_type && node.return_type != 0 {
+		if g.inside_return {
+			return true
+		}
 		if g.table.sym(node.return_type).kind in [.sum_type, .multi_return]
 			|| node.return_type.has_option_or_result() {
 			return true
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #23853, opened 2025-03-03
build,examples: add build system

EmmaTheMartian:
This PR introduces a small `build` module into the Vlib, based on the API in [Clockwork](https://github.com/emmathemartian/clockwork). It's not quite 1:1 but is still similar.

I am seeking feedback, just leave a comment here if you have any!

This PR will remain a draft while I am receiving feedback, once I feel like it's in a more comfortable state then I will un-draft it.

### DIFF
--- a/examples/build_system/build.vsh
+++ b/examples/build_system/build.vsh
@@ -1,0 +1,45 @@
+#!/usr/bin/env v
+
+import build
+
+// Define variables that can be used to change tasks in the buildscript
+const app_name = 'hello'
+const program_args = 'World'
+const build_dir = 'target'
+
+// Make the build context
+mut context := build.context()
+
+// Add a few simple tasks
+context.task(name: 'doc', run: || system('echo "Nothing to do"'))
+context.task(name: 'run', run: || system('v run . ${program_args}'))
+context.task(name: 'build', run: || system('v .'))
+context.task(name: 'build.prod', run: || system('v -prod -o ${app_name} .'))
+
+// `_` to denote "private" tasks. Nothing stops the user from using it, but
+// this tells them that the task is not meant to be used by them.
+context.task(
+	name: '_mkdirs'
+	// The `help` field is displayed in `--tasks` to give a short summary of what the task does.
+	help: 'Makes the directories used by the application'
+	run:  fn () ! {
+		if !exists(build_dir) {
+			mkdir_all(build_dir) or { panic(err) }
+		}
+	}
+)
+
+// Add a more complex task
+context.task(
+	name:    'release'
+	help:    'Build the app in production mode, generates documentation, and releases the build on Git'
+	depends: ['_mkdirs', 'doc']
+	run:     fn () ! {
+		system('v -prod -o ${build_dir}/${app_name} .')
+		// Pretend we are using Git to publish the built file as a release here.
+	}
+)
+
+// Run the build context. This will iterate over os.args and each corresponding
+// task, skipping any arguments that start with a hyphen (-)
+context.run()
--- a/examples/build_system/build.vsh
+++ b/examples/build_system/build.vsh
@@ -1,6 +1,7 @@
-#!/usr/bin/env v
+#!/usr/bin/env -S v run
 
 import build
+import time
 
 // Define variables that can be used to change tasks in the buildscript
 const app_name = 'hello'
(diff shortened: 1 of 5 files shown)


phcreery:
I would be nice if a task could be run dependent on the existence of a file or artifact, similar to GNU make. This would be nice for c dependencies and would not run again if the file already exists.

```v
context.artifact(
    name: 'cimgui', 
    file: 'thirdparty/cimgui/libcimgui.a', 
    run: || system('cd thirdparty/cimgui && make static')
)
context.task(
    name: 'build', 
    depends: ['cimgui'], 
    run: || system('v .')
)
```

EmmaTheMartian:
> I would be nice if a task could be run dependent on the existence of a file or artifact, similar to GNU make. This would be nice for c dependencies and would not run again if the file already exists.
> 
> ```v
> context.artifact(
>     name: 'cimgui', 
>     file: 'thirdparty/cimgui/libcimgui.a', 
>     run: || system('cd thirdparty/cimgui && make static')
> )
> context.task(
>     name: 'build', 
>     depends: ['cimgui'], 
>     run: || system('v .')
> )
> ```

You could also just use an if statement in the task's code, like so:
```v
context.task(
        name: 'cimgui', 
        run:  fn () {
                if !exists('thirdparty/cimgui/libcimgui.a') {
                        system('cd thirdparty/cimgui && make static')
                }
        }
)
```

I do see the benefit in there being a shorthand, although the name `artifact` feels a little awkward. Maybe that's just because I'm not extremely familiar with Make though :P

JalonSolov:
`object` would do just as well.  Built objects are "artifacts" of the build system.

examples/build_system/build.vsh
```diff
@@ -0,0 +1,45 @@
...
+	depends: ['_mkdirs', 'doc']
+	run:     fn () ! {
+		system('v -prod -o ${build_dir}/${app_name} .')
+		// Pretend we are using Git to publish the built file as a release here.
+	}
+)
+
+// Run the build context. This will iterate over os.args and each corresponding
+// task, skipping any arguments that start with a hyphen (-)
+context.run()
```
spytheman, line 45: What happens for just `v build.vsh` ?
Is there a concept of a default task?

In `make`, the first rule is the default one.

examples/build_system/main.v
```diff
@@ -0,0 +1,5 @@
+import os
```
spytheman, line 1: imho add a `v.mod` file here too, as well as a `common/` folder, with a `f1.v` and `f2.v` files in there, each starting with `module common`, and having 2 different public functions, both used by `main.v` (after it does `import common`).

The general idea, is to make it easy to test, that the build system supports dependencies, and imports, so that later, modifying `common/f1.v`, will lead to invalidating the tasks, that depend on `main.v` too, and re-running them.

spytheman: Note: `v -print-watched-files main.v` will output a list of all the .v files, that are  involved in the compilation of `main.v`

examples/build_system/build.vsh
EmmaTheMartian: It errors and tells you that you didn't provide any tasks to run, along with mentioning that passing `--tasks` will show a list of tasks.

A default task would be good to have, although I'd rather it not just implicitly be the first task. I didn't even know Make done that...

Maybe we could have a `default` parameter in the params for `task()`, and when true it will function as the default. If multiple are defined then it would error out with a message saying that there can only be one default task.

vlib/build/README.md
```diff
@@ -0,0 +1,32 @@
+## Description
+
+`build` provides a small build system leveraging V(SH) for the buildscript.
+
+## Example
+
+```v
+#!/usr/bin/env v
```
spytheman, line 8: Run `VAUTOFIX=1 v check-md vlib/build/README.md` 2 times to fix the formatting issue.

spytheman:
A good resource for ideas, may be https://www.microsoft.com/en-us/research/wp-content/uploads/2020/04/build-systems-jfp.pdf .

examples/build_system/main.v
EmmaTheMartian: The build system is not doing anything to change how `import` would normally function, nor dependencies. I can add automatic dependency downloading (i.e, automatically running `v install` to install dependencies in the `v.mod` file) if that's what you mean

spytheman: Note: I do not care about the task runner aspect, as much as about the build system aspect.

JalonSolov: He's not saying anything about imports.  He's saying that an example for a build system should show examples that make sense for a build system.  Only having a single file doesn't really show anything except that the single file can be built.

With multiple files, you can show how if one has already been built, it doesn't get built the next time if it hasn't changed, while those that have been changed are rebuilt, etc.

spytheman: If I want to rebuild a project, that has *no modifications* to its source files, and that was already build, the build system should do nothing (since the target executable will already exist).

If I have a project that has modified files in it, the question of whether the target executable should be rebuild, becomes more complicated, and to be solved, it requires the build system to know/record **which .v files were used, for the building of the executable**, and to only rebuild it, if *at least* one of them has been changed.

For example, I may have a project that builds 2 executables, one that is build from `tool1.v` + `common/*.v` + `somemodule1/*.v`, and another one, that is build from `tool2.v` + `common/*.v` + `somemodule2/*.v` .

Note how `common/*` is used in both, but `somemodule1/*.v` is used only by `tool1.v` .

If I make a change in one of the .v files in `common/`, the executable `tool1.exe` should be rebuild, even if `tool1.v` itself was not changed at all. `tool2.exe` should also be rebuild.

If I make a change in `somemodule1/*.v`, then only `tool1.exe` should be rebuild.

EmmaTheMartian: Ohhh okay I see what you mean now

medvednikov:
@EmmaTheMartian what's left to implement?

EmmaTheMartian:
> @EmmaTheMartian what's left to implement?

I need to implement a "correct" way to build a multi-module project (i.e, two modules that are separate applications, such as an app with a GUI frontend, CLI frontend, and common API).

Currently it's really just a fancy command runner. Merging it now would still work, it's stable in my uses so far.

EmmaTheMartian:
I forgot to move this from draft to review, whoops. You should be able to merge it now if everything looks good.

PR #23864, opened 2025-03-05
crypto: add missing docs

virus-rpi:
This pull request introduces several documentation improvements across multiple files in the `vlib/crypto` directory. 

* [`vlib/crypto/cipher/cfb.v`](diffhunk://#diff-a014fe16900459b26c77bff51d7d5bde2f6e1e798b6f0a0b1a0f91e8427e585bR66): Added a comment to the `xor_key_stream` function explaining its purpose.
* [`vlib/crypto/cipher/ctr.v`](diffhunk://#diff-d20f0dddfe0886d7b62f038565111a13bcf10ed8f94e1ef3d10f18fd675dd599R52): Added a comment to the `xor_key_stream` function explaining its purpose.
* [`vlib/crypto/cipher/ofb.v`](diffhunk://#diff-938af236930e4ceb99b11fa4c57710d75993d44baa29cd9699ab3cb9bed33e2eR38): Added a comment to the `xor_key_stream` function explaining its purpose.
* [`vlib/crypto/des/des.v`](diffhunk://#diff-ee952fac29dc5207217a44bdcc542fb5924327b7ee9d521c21f565e3eefbf53cR59): Added comments to the `encrypt` and `decrypt` functions in both `DesCipher` and `TripleDesCipher` classes explaining their purposes. [[1]](diffhunk://#diff-ee952fac29dc5207217a44bdcc542fb5924327b7ee9d521c21f565e3eefbf53cR59) [[2]](diffhunk://#diff-ee952fac29dc5207217a44bdcc542fb5924327b7ee9d521c21f565e3eefbf53cR73) [[3]](diffhunk://#diff-ee952fac29dc5207217a44bdcc542fb5924327b7ee9d521c21f565e3eefbf53cR99) [[4]](diffhunk://#diff-ee952fac29dc5207217a44bdcc542fb5924327b7ee9d521c21f565e3eefbf53cR136)
* [`vlib/crypto/rand/rand.v`](diffhunk://#diff-007bf08348dcea1403673ccda290dc1c9bc9ada968ac2c930062dde9c147e5ccR11): Added a comment to the `msg` function in the `ReadError` struct explaining its purpose.


vlib/crypto/cipher/cfb.v
```diff
@@ -63,6 +63,7 @@ fn new_cfb(b Block, iv []u8, decrypt bool) Cfb {
 	return x
 }
 
+// xor_key_stream xors each byte in the given slice with a byte from the key stream.
```
spytheman, line 66: Are you sure that *each* byte is xor-ed?
To me it is not obvious 🤔 .

@blackshirt, @kimshrier what do you think?

spytheman: I just saw the comments in the interface:
```v
// A Stream represents a stream cipher.
pub interface Stream {
mut:
    // xor_key_stream XORs each byte in the given slice with a byte from the
    // cipher's key stream. Dst and src must overlap entirely or not at all.
    //
    // If len(dst) < len(src), xor_key_stream should panic. It is acceptable
    // to pass a dst bigger than src, and in that case, xor_key_stream will
    // only update dst[:len(src)] and will not touch the rest of dst.
    //
    // Multiple calls to xor_key_stream behave as if the concatenation of
    // the src buffers was passed in a single run. That is, Stream
    // maintains state and does not reset at each xor_key_stream call.
    xor_key_stream(mut dst []u8, src []u8)
}
```

spytheman: Given that method implements that Stream interface, I think it is safe to assume that the same comments should apply.

spytheman: I'll merge it as it is for now.

PR #23872, opened 2025-03-06
checker: add fntype casting validations

### LINKED ISSUE #23852: V should require unsafe for type castings of zero
### Describe the bug

V allows type casting zero without requiring unsafe.

### Reproduction Steps

Example

`x := ClickFn(0)`

### Expected Behavior

Compiler should warn that unsafe is required (e.g. `unsafe { ClickFn(0) }`)

### Current Behavior

No warning is issued

### Possible Solution

_No response_

_No response_

### V version

V 0.4.9 1b136e2

|V full version      |V 0.4.9 da3112e.1b136e2
|OS                  |macos, macOS, 15.3.1, 24D70
|Memory              |0.07GB/8GB
|                    |
|V last modified time|2025-03-02 19:05:12
|                    |
|Current working dir |OK, value: /Users/mike/Documents/github/kite
|                    |
|V git status        |weekly.2025.07-129-g1b136e21 (4 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |Apple clang version 16.0.0 (clang-1600.0.26.6)
|gcc version         |Apple clang version 16.0.0 (clang-1600.0.26.6)
|clang version       |Apple clang version 16.0.0 (clang-1600.0.26.6)
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

felipensp:
Fix #23852

### DIFF
--- a/vlib/os/process_windows.c.v
+++ b/vlib/os/process_windows.c.v
@@ -14,7 +14,7 @@ type FN_NTSuspendResume = fn (voidptr) u64
 fn ntdll_fn(name &char) FN_NTSuspendResume {
 	ntdll := C.GetModuleHandleA(c'NTDLL')
 	if ntdll == 0 {
-		return FN_NTSuspendResume(0)
+		return unsafe { FN_NTSuspendResume(0) }
 	}
 	the_fn := FN_NTSuspendResume(C.GetProcAddress(ntdll, voidptr(name)))
 	return the_fn
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -3592,13 +3592,20 @@ fn (mut c Checker) cast_expr(mut node ast.CastExpr) ast.Type {
 		c.error('casting a function value from one function signature, to another function signature, should be done inside `unsafe{}` blocks',
 			node.pos)
 	} else if final_to_sym.kind == .function && final_from_sym.kind != .function {
-		if !(c.inside_unsafe || c.file.is_translated) {
+		if to_type.has_flag(.option) && node.expr !is ast.None {
+			c.error('casting number to Option function is not allowed, only compatible function or `none`',
+				node.pos)
+		} else if !(c.inside_unsafe || c.file.is_translated) {
 			if node.expr is ast.IntegerLiteral {
 				c.warn('casting number to function value should be done inside `unsafe{}` blocks',
 					node.pos)
 			} else if node.expr is ast.Nil {
 				c.warn('casting `nil` to function value should be done inside `unsafe{}` blocks',
 					node.pos)
+			} else if node.expr is ast.None {
+				if from_type.has_flag(.option) {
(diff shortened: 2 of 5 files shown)


spytheman:
`none` should be allowed for option values:
```v
type Foo = fn (bar string)

fn test_main() {
    mut func := ?Foo(none)
    dump(func)
}
```

spytheman:
On master that passes, but it fails here currently (which is the CI's problem):
image


(as an aside, the error is not shown on the CI, because of the `-silent` flag that was passed to reduce the scrolling for the normally `OK` lines in the test)

felipensp:
> On master that passes, but it fails here currently (which is the CI's problem): image
> 
> (as an aside, the error is not shown on the CI, because of the `-silent` flag that was passed to reduce the scrolling for the normally `OK` lines in the test)

Right, I'll add the option test too. 👍🏻

medvednikov:
Thanks!

medvednikov:
> casting number to function value should be done inside `unsafe{}` blocks

This should not be allowed at all. We recently forced requiring nil instead of 0 for the same reason.

spytheman:
> This should not be allowed at all. We recently forced requiring nil instead of 0 for the same reason.

No, it should be allowed in `unsafe {}`.

medvednikov:
Why?

spytheman:
Because there are cases (for example with -freestanding, you can write code that can then be booted on bare metal), when you write code, where you do know the actual address of a function (because it is an external requirement), and do want to cast it to a function pointer.

Restricting what can be done in `unsafe{}` blocks in general kind of defeats the whole purpose of the unsafe expressions.

medvednikov:
Then just use `Func(voidptr(1234))`

V positions itself as a safe language. You can't cast numbers to functions in a safe language :)

spytheman:
> Then just use `Func(voidptr(1234))`
> 
> It's more clear this way too. `Func(1234)` looks like a function call.
> 
> V positions itself as a safe language. You can't cast numbers to functions in a safe language :)

We are talking then about:
 `unsafe { Func(voidptr(1234)) }`

I think that is very verbose for no gain at all.

spytheman:
> V positions itself as a safe language. 

V itself does not do anything on its own. It is a language, not a person.

You, however position it as a language, that I will not want to use for certain tasks, because of the verbosity and edge cases.

Unsafe blocks are in V for, to allow people to write unsafe code in them, not to make them jump through hoops to avoid weird quirks in the compiler.

felipensp:
I've added V UI PR fix https://github.com/vlang/ui/pull/599

spytheman (approved):
Excellent work @felipensp, thank you.

However, I will not merge it, since @medvednikov expressed another opinion, and he is ultimately the arbiter of what should happen with his language.

I do hope, that he will reconsider, but that is of course up to him.

medvednikov:
V is a very readable language. Readability suffers with `Func(1234)`.

PR #23874, opened 2025-03-06
strings.textscanner: add rune mode support; add next_line() , read_until(), substr() funcs

kbkpbot:
Introduce `force_rune_mode` params to the `new(input string, config TextScannerConfig)` and this will help scan runes instead of bytes of a `input` string.

```v

@[params]
pub struct TextScannerConfig {
pub mut:
	force_rune_mode bool
}

// new returns a stack allocated instance of TextScanner.
pub fn new(input string, config TextScannerConfig) TextScanner

// next_line advances the scanner’s position to the start of
// the next line, and return the line.
// Returns true if successful, or false if the end of the input
// is reached.

pub fn (mut ss TextScanner) next_line() (string, bool) 

// read_until reads characters from the current scanning position
// until a delimiter (from the provided string `delimiters`) is encountered.
// The returned string includes all characters from the starting
// position up to (but ​not​ including) the first encountered
// delimiter. The scanner's position is advanced to the character
// immediately after the delimiter (or to the end of the input if
// no delimiter is found).

pub fn (mut ss TextScanner) read_until(delimiters string) !string

// substr return a sub string of input string from start to end.
pub fn (mut ss TextScanner) substr(start int, end int) string 
```

### DIFF
--- a/vlib/strings/textscanner/textscanner.v
+++ b/vlib/strings/textscanner/textscanner.v
@@ -132,14 +132,15 @@ pub fn (ss &TextScanner) peek_u8() u8 {
 // ts.peek_n(1) == ts.peek() .
 @[direct_array_access; inline]
 pub fn (ss &TextScanner) peek_n(n int) int {
-	if ss.pos + n < ss.ilen {
-		if ss.config.force_rune_mode {
-			return int(ss.input_runes[ss.pos + n])
-		} else {
-			return ss.input_bytes[ss.pos + n]
-		}
+	new_pos := ss.pos + n
+	if new_pos < 0 || new_pos >= ss.ilen {
+		return -1
 	}
-	return -1
+	if ss.config.force_rune_mode {
+		return int(ss.input_runes[ss.pos + n])
+	} else {
+		return ss.input_bytes[ss.pos + n]
+	}
 }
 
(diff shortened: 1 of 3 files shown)


vlib/strings/textscanner/textscanner.v
```diff
@@ -185,3 +186,74 @@ pub fn (mut ss TextScanner) skip_whitespace() {
...
+
+// read_until reads characters from the current scanning position
+// until a delimiter (from the provided list `delimiters`) is encountered.
+// The returned string includes all characters from the starting
+// position up to (but ​not​ including) the first encountered
+// delimiter. The scanner's position is advanced to the character
+// immediately after the delimiter (or to the end of the input if
+// no delimiter is found).
+@[direct_array_access]
+pub fn (mut ss TextScanner) read_until(delimiters []rune) !string {
```
spytheman, line 236: Why `[]rune`, instead of just `string`?

The iteration inside happens byte by byte anyway, and you can use: `if delimiters.index_u8(r) != -1 {`

spytheman:
May be there is a misunderstanding - I *do not want* rune mode support, and if you do want it, you can just write your own scanner module, outside of vlib's `strings`. Supporting a rune mode only complicates and slows down the scanner. 

My comment was just to point out, that you can pass a string for the separators, which is more compact.

PR #23876, opened 2025-03-07
crypto.ecdsa: migrate new_key_from_seed to use high opaque

blackshirt:
This PR migrates mostly for `new_key_from_seed` function to use high level EVP_PKEY opaque. Its also migrates equality check on PrivateKey and PublicKey.
- Its also adding some helpers to simplify the logic
- Adds some required definition

There are some bits thats not fully migrated, like the `.bytes()` methods on the both, maybe its ready on the next PR.
After that, we can clean its up to remove the obsoleted bits from the module.

Thanks 

.

PR #23877, opened 2025-03-08
crypto.ecdsa: the last bits of migration steps

blackshirt:
This follow-up PR from the last commited [pr](https://github.com/vlang/v/pull/23876). In this PR, its done by migrating the last bits required to be changed, ie `.bytes()` method on the `PrivateKey` and `PublicKey`.
Its was done by using high level function. All info's of `EVP_PKEY` opaque was already availables on both of them by previous migrations. This series contains the bits forms:

- The changes on `.bytes()` on the both of `PrivateKey` and `PublicKey`
- Adds required definitions
- Some fixs of private and public key loader

After this PR, we can start out to cleanup some deprecated code. Should be this cleanup phase done in gradually step ?

### DIFF
--- a/vlib/crypto/ecdsa/ecdsa.c.v
+++ b/vlib/crypto/ecdsa/ecdsa.c.v
@@ -44,6 +44,8 @@ fn C.EVP_PKEY_bits(pkey &C.EVP_PKEY) int
 fn C.EVP_PKEY_size(key &C.EVP_PKEY) int
 fn C.EVP_PKEY_eq(a &C.EVP_PKEY, b &C.EVP_PKEY) int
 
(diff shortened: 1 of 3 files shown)


JalonSolov:
V "policy" is to add `@[deprecated]` attributes, and set a date for 6 months in the future before the deprecation warning turns into an error.

blackshirt:
> V "policy" is to add `@[deprecated]` attributes, and set a date for 6 months in the future before the deprecation warning turns into an error

Actually, this series was not about deprecation,  .. I think its can be done in the last step after cleans up with the reasons,

spytheman:
Deprecating public API takes time.

Code that is not public, can be just deleted, when it is no longer used.

PR #23885, opened 2025-03-08
examples, breakout: add subtle light and shadow effects for feeling of depth

larpon:
Small visual tweaks to give the elements a little more depth:
image

### DIFF
--- a/examples/breakout/breakout.v
+++ b/examples/breakout/breakout.v
@@ -10,6 +10,9 @@ const designed_width = 600
 const designed_height = 800
 const brick_width = 53
 const brick_height = 20
+const bevel_size = int(brick_height * 0.18)
+const highlight_color = gx.rgba(255, 255, 255, 65)
+const shade_color = gx.rgba(0, 0, 0, 65)
 
 struct Brick {
 mut:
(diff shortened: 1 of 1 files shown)


spytheman:
It looks much nicer 🥳!

Thank you!

spytheman:
I can sometimes put the ball in a vertical line, that is right near the edge of a column of bricks, and then it seems like the ball passes partially *below* the edge of the column, until it hits the top wall.

It is hard to reproduce though, I would have to change the code to make it easier.

spytheman:
we may need to expand the hitboxes of the bricks, to include the new decorations 🤔

spytheman:
image

That is with:
```v
                               ._0 { g.ball_dx = 0 }
                               ._1 { g.ball_x-- }
                               ._2 { g.ball_x++ }
```
inside the `keydown_fn:` handler, and then pressing `0` and `1` and `2` to adjust the x offset of the ball

spytheman:
another option is to just draw the ball last 🤔, not sure it that is better

spytheman:
It turns out, that it is unrelated to the PR, it does it on master too:
image

PR #23887, opened 2025-03-09
crypto.ecdsa: cleans out old deprecated keypair opaque from the module.

blackshirt:
After this [PR](https://github.com/vlang/v/pull/23877) has been merged, all info's of high level opaque was already availables on the public and private key object, so, we can start out the removal of the deprecated ec_key opaque.
This PR contains this cleansup step in one-shoot commit.  This cleansup significantly reduces the overall size line of the code, ie,

- `ecdsa.v` file, reduced from 905 lines down to 621 lines
- `ecdsa.c.v` file reduced from 200 down to 172 lines
- `util.v` gets increased from 233 into 247, not too much addition, mostly for checking functionality
- the rest test files remain the same, not changed in the count

This PR contains some bits in the forms of :

- Removes out of `EC_KEY` opaque from definitions and private and public key structure
- Removes out of the alls of an old deprecated of `EC_KEY` based functionality and relies on the new high level `EVP_PKEY` opaque.
- Removes out unneeded old helpers function
- Removes out unneeded C declarations
- Add needed C declarations to the file
- Add some required helpers
- Some bits of fixs on the the test and code.

The good news is, on my test box, its now buildable with `-cstrict` options, where previously, its would fail with these option because of deprecation error.

Good days..thx

### DIFF
--- a/vlib/crypto/ecdsa/util.v
+++ b/vlib/crypto/ecdsa/util.v
@@ -241,7 +241,7 @@ fn key_group_name(key &C.EVP_PKEY) !string {
 		unsafe { gname.free() }
 		return error('fail to get group name')
 	}
-	group := gname[..gname_len].clone().bytestr()
+	group := gname[..gname_len].bytestr()
 	unsafe { gname.free() }
 	return group
 }
--- a/vlib/crypto/ecdsa/util.v
+++ b/vlib/crypto/ecdsa/util.v
@@ -37,26 +37,34 @@ pub fn pubkey_from_bytes(bytes []u8) !PublicKey {
 		return error('Get an nid of non ecPublicKey')
 	}
 
-	eckey := C.EVP_PKEY_get1_EC_KEY(pub_key)
-	if eckey == 0 {
-		C.EC_KEY_free(eckey)
-		return error('Failed to get ec key')
+	// check the key
+	pctx := C.EVP_PKEY_CTX_new(pub_key, 0)
+	if pctx == 0 {
+		C.EVP_PKEY_CTX_free(pctx)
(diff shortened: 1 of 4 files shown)


vlib/crypto/ecdsa/util.v
```diff
@@ -168,66 +187,61 @@ pub fn privkey_from_string(s string) !PrivateKey {
...
+// key_group_name returns underlying group name of the key as a string.
+fn key_group_name(key &C.EVP_PKEY) !string {
+	gname := []u8{len: default_groupname_size}
+	gname_len := usize(0)
+	mut s := C.EVP_PKEY_get_group_name(key, gname.data, u32(gname.len), &gname_len)
+	if s == 0 {
+		unsafe { gname.free() }
+		return error('fail to get group name')
+	}
+	group := gname[..gname_len].clone().bytestr()
```
spytheman, line 244:
```suggestion
	group := gname[..gname_len].bytestr()
```
`.bytestr()` creates a copy too, i.e. the clone() is superfluous

spytheman (approved):
Excellent work!
Now the module tests, pass with -cstrict, for both gcc and clang, on my system.
Screenshot_2025-03-09_09-57-46

vlib/crypto/ecdsa/util.v
blackshirt: Thanks for fixing it

blackshirt:
> Excellent work! Now the module tests, pass with -cstrict, for both gcc and clang, on my system. Screenshot_2025-03-09_09-57-46

Good to have finally we reach out the migration  and clean it up. I would look into it maybe there are something can be improved (cleaned)

PR #23889, opened 2025-03-09
docs, cmd: fix broken docs source links regression in `vdoc`

larpon:
Locally running `v doc -m -f html -o ./tmpdocs vlib/ && firefox tmpdocs/builtin.html` now results in source links showed:
image

I'm unsure about how to add a regression test :thinking: 
Maybe we could do an `assert` somewhere if `cfg.is_vlib == true`  :shrug:

### DIFF
--- a/cmd/tools/vdoc/vdoc.v
+++ b/cmd/tools/vdoc/vdoc.v
@@ -314,7 +314,7 @@ fn (mut vd VDoc) generate_docs_from_file() {
 			vd.manifest = manifest
 		}
 	} else if cfg.is_vlib {
-		assert false, 'vdoc: manifest does not exists for vlib'
+		assert false, 'vdoc: manifest does not exist for vlib'
 	}
 	if cfg.include_readme || cfg.is_vlib {
 		mut readme_name := 'README'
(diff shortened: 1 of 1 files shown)


cmd/tools/vdoc/vdoc.v
```diff
@@ -303,12 +303,18 @@ fn (mut vd VDoc) generate_docs_from_file() {
...
+	} else {
+		os.join_path(dir_path, 'v.mod')
+	}
 	if os.exists(manifest_path) {
 		vd.vprintln('Reading v.mod info from ${manifest_path}')
 		if manifest := vmod.from_file(manifest_path) {
 			vd.manifest = manifest
 		}
+	} else if cfg.is_vlib {
+		assert false, 'vdoc: manifest does not exists for vlib'
```
Delta456, line 317: I believe `eprintln` can function better here, followed by an exit as done above.

larpon: The idea is to ensure that we do not let this happen again when the code base is worked upon - it is not (/should not be) a `vdoc` runtime error IMO. Passing an invalid / unreadable manifest_file results in bad side-effects when vlib's own docs are generated. I just don't  currently see how to put his in a test where it would probably fit better?

spytheman (approved):
Excellent work.

Issue #23888, opened 2025-03-09
Create .dll under Windows will result in strange filename.


kbkpbot:
### Describe the bug

When create a .dll , it depend on the output name, and sometime create strange filename.

### Reproduction Steps

testdll.v
```v
module testdll

const r = 0.14
const s = 3 + r

@[callconv: stdcall]
@[export: get_pi]
pub fn return_pi() f32 {
	return s
}
```

```
v -shared testdll.v -o lib.so
```

Note: we try to create a `lib.so` instead of `lib.dll`.

### Expected Behavior

create a output filename `lib.so`

### Current Behavior

create a output filename `lib.so.dll` and a def filename `li.def`

### Possible Solution

fix the output filename and def filename

### V version

V 0.4.9 17fc31b.4236baf

### Environment details (OS name and version, etc.)

V full version      |V 0.4.9 17fc31b.4236baf
|:-------------------|:-------------------
|OS                  |windows, Microsoft Windows 10 企业版 LTSC 19044 64 位
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 7840H with Radeon 780M Graphics
|Memory              |0.95GB/27.69GB
|                    |
|V executable        |D:\v\v\v\v.exe
|V last modified time|2025-03-08 08:18:15
|                    |
|V home dir          |OK, value: D:\v\v\v
|VMODULES            |OK, value: C:\Users\DDT\.vmodules
|VTMP                |OK, value: C:\Users\DDT\AppData\Local\Temp\v_0
|Current working dir |OK, value: D:\v\bug\shared
|                    |
|Git version         |git version 2.43.0.windows.1
|V git status        |weekly.2025.10-32-g4236baf5-dirty (17 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |N/A
|clang version       |N/A
|msvc version        |用于 x64 的 Microsoft (R) C/C++ 优化编译器 19.39.33523 版
|tcc version         |tcc version 0.9.27 (x86_64 Windows)
|tcc git status      |thirdparty-windows-amd64 b425ac82
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

kbkpbot:
[...]

Issue #6983, opened 2020-11-27
Missing main entry point in shared libraries


Ekopalypse:
Unlike executable files libraries do not use a main entry point, at least on Windows.
No entry point means _vinit is not called at all, hence consts etc... do not get initialized.
On Windows I would suggest to generate the DllMain function which calls then _vinit and possibly a user-main function, which gets the parameters from DllMain forwarded.

**V version:** 
OS: windows, Microsoft Windows 7 Professional v7601 64-bit
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i5-2500 CPU @ 3.30GHz
CC version: N/A

vroot: D:\ProgramData\Compiler\v
vexe: D:\ProgramData\Compiler\v\v.exe
vexe mtime: 2020-11-27 18:44:30
is vroot writable: true
V full version: V 0.1.29 e0d6490.1891f55

Git version: git version 2.19.1.windows.1
Git vroot status: latest-commit-414-g1891f55c
.git/config present: true
thirdparty/tcc status: master 17d18c13

**What did you do?**
build shared library

**What did you expect to see?**
The function DllMain created
 
**What did you see instead?**
DllMain is missing

danieldaeschle:
```v
// v -cc msvc -shared FILE_NAME
module testdll

fn C.MessageBoxW(voidptr, &u16, &u16, u32) int

[windows_stdcall]
[export: greet]
pub fn send_greetings() {
    title := 'DLL Application'.to_wide()
    message := 'Hello DLL'.to_wide()
    C.MessageBoxW(0, title, message, 0)    
}
```

Ekopalypse:
Here is a sample code that shows the effect of the missing _vinit function a bit more clearly.

code for dll build:
~~~v
// v -cc msvc -shared FILE_NAME
module testdll

// fn C._vinit(int, voidptr)
// fn C._vcleanup()
// fn C.GC_INIT()

const (
	r = 0.14
	s = 3 + r
)

[windows_stdcall]
[export: get_pi]
pub fn return_pi() f32 {
	return s
}

// [windows_stdcall]
// [export: DllMain]
// fn main(hinst voidptr, fdw_reason int, lp_reserved voidptr) bool{
	// match fdw_reason {
		// C.DLL_PROCESS_ATTACH {
			// $if static_boehm ? {
				// C.GC_INIT()
			// }
			// C._vinit(0, 0)
		// }
		// C.DLL_THREAD_ATTACH {
		// }
[...]

PR #23890, opened 2025-03-09
builder,cgen: fix msvc build filename, remove temp files

kbkpbot:
1. Partially fix issue #23888 . If user specify a output name other than `.exe` or `.dll`, the output name will be appended with `.exe` or `.dll`;
2. Remove `msvc` generate temp files.

### DIFF
--- a/vlib/v/builder/msvc_windows.v
+++ b/vlib/v/builder/msvc_windows.v
@@ -256,6 +256,15 @@ pub fn (mut v Builder) cc_msvc() {
 	}
 	out_name_pdb := os.real_path(v.out_name_c + '.pdb')
 	out_name_cmd_line := os.real_path(v.out_name_c + '.rsp')
+	// testdll.01JNX9W7JAV4FKMZ6KDXT67QYV.tmp.so.c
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

Ekopalypse:
> > wait a moment, try fix issue #6983
> 
> As `tcc` doesn't has a method that change an export function's name, so my suggestion is not use the `@[callconv: stdcall]`.
> 
> `msvc` can use a `.def` file change the export function's name, so it can use `@[callconv: stdcall]`.

I don't understand what the calling convention has to do with the exported function names?

JalonSolov:
In V, names are usually exported as `<module name>__<function name>` for normal functions, `<module name>__<struct name>_<receiver method name>`, etc.  This is very similar to the `name mangling` that C++ does for symbols.

So if you want a specific name for external things to call, you need to export said specific name, instead of others having to figure out the naming scheme.

Ekopalypse:
So does this mean that there is no longer an export attribute?
Afaik, a calling convention does NOT only define what names should be, but above all how they are called and how parameters are passed or handled.

PR #23891, opened 2025-03-09
veb: handle sendfile errors

zUltimaPG3D:
fix [the issue in the discord server help channel](https://discord.com/channels/592103645835821068/592294828432424960/1348248565646102569) where cancelling a download prompted by `return ctx.file("filename")` occasionally crashed the program due to a SIGPIPE signal being caught (no github issue for this exists as far as i'm aware)

no test comes with this PR due to the issue being seemingly random(?) and most likely impossible to automate. if needed, create a simple server in veb that serves a file, then repeat the download request, cancelling it each time until the program crashes (which shouldn't happen with this fix)


spytheman:
Thank you 🙇🏻‍♂️ .

PR #23893, opened 2025-03-09
cli: add missing struct members to str() method and fix some comments

kimshrier:
I added some missing structure members to the str() method.  I did not remove the deprecated members as I didn't know if that was something we should do now or postpone until later.

I am also considering writing up more descriptive and potentially tutorial type documentation in the README.md file as well as adding to the comments in the source.  This would probably be fairly long and I didn't know if that level of documentation was appropriate for being in the source of vlib.


spytheman:
I do not see a problem in adding more documentation to the `README.md` or in other .md files in the modules.

PR #23895, opened 2025-03-09
v.util: fix stack overflow due to unlimited recursion in resolve_d_value

spytheman:


larpon (approved):
Good work!
Looks good to me

PR #23900, opened 2025-03-11
tools: support `// vtest build: !do_not_test ?`, `// vtest build: !windows && tinyc` to skip files during testing on specific platforms, without having to keep centralised skip lists

spytheman:


JalonSolov:
YES!  Excellent change.  Keeping hard-coded lists in files is cumbersome, error prone, and causes more churn than necessary.

cmd/tools/modules/testing/common.v
```diff
@@ -568,9 +574,23 @@ fn worker_trunner(mut p pool.PoolProcessor, idx int, thread_id int) voidptr {
 	} else {
 		os.quoted_path(generated_binary_fpath)
 	}
+	mut details := get_test_details(file)
+	mut should_be_build := true
```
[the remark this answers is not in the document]
spytheman: thanks

cmd/v/v.v
```diff
@@ -221,7 +221,7 @@ fn setup_vbuild_env_vars(prefs &pref.Preferences) {
 	}
 	github_job := os.getenv('GITHUB_JOB')
 	if github_job != '' {
-		facts << github_job
+		facts << os.getenv('GITHUB_JOB')
```
JalonSolov, line 224: Why go to the env var again?  The value has already been stored in the `github_job` var, so it looks like the correct code was change to something less efficient.

spytheman: Tere is a compiler bug, due to which calling `facts.free()` later, just frees the `facts` array, but not the elements in it.
There is an overriden version of the free method for `[]string`, in `builtin`, that loops over the elements, and frees them, but cgen does not choose it over the generic `fn (a &array) free() {`, while it should.

The change here was an attempt to workaround the bug, before I understood it completely.

I'll push the more general fix in a separate PR, and then I will rebase this one over it.

Issue #23898, opened 2025-03-10
import crypto.aes


tankf33der:
### Describe the bug

doing crypto audit against test vectors.

### Reproduction Steps

```
$ v
>>> import crypto.aes
```

### Expected Behavior

just works.

### Current Behavior
```
<A LOT OF WARNING and ERRORS from builder>
```

### V version

V 0.4.9 6623ac2

### Environment details (OS name and version, etc.)

```
|V full version      |V 0.4.9 52cd627.6623ac2
|:-------------------|:-------------------
|OS                  |linux, N/A
|Processor           |20 cpus, 64bit, little endian, 13th Gen Intel(R) Core(TM) i5-13500
|Memory              |32.55GB/62.61GB
|                    |
|V executable        |/root/v/v
|V last modified time|2025-03-10 13:03:08
|                    |
|V home dir          |OK, value: /root/v
|VMODULES            |OK, value: /root/.vmodules
|VTMP                |OK, value: /tmp/v_0
|Current working dir |OK, value: /root
|                    |
|Git version         |git version 2.47.2
|V git status        |weekly.2025.10-56-g6623ac21
|.git/config present |true
|                    |
|cc version          |cc (Alpine 14.2.0) 14.2.0
|gcc version         |gcc (Alpine 14.2.0) 14.2.0
|clang version       |Alpine clang version 19.1.4
|tcc version         |tcc version 0.9.27 mob:d3e940c (x86_64 Linux)
|tcc git status      |thirdparty-linuxmusl-amd64 a3e24da2
|emcc version        |N/A
|glibc version       |Error: musl libc (x86_64)
Version 1.2.5
Dynamic Program Loader
Usage: /lib/ld-musl-x86_64.so.1 [options] [--] pathname
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

JalonSolov:
Yup.  A new warning added to V.  The aes files just need to be updated.  This should be a pretty easy PR for someone...
```
vlib/crypto/aes/block_generic.v:73:57: notice: ambiguous expression. use `()` to ensure correct order of operations
   71 |     }
   72 |     // Last round uses s-box directly and XORs to produce output.
[...]

PR #23902, opened 2025-03-11
crypto.aes: fix warnings about order of operations 

tankf33der:
fix warnings in crypto.aes for #23898


spytheman (approved):
Thank you 🙇🏻‍♂️ .

PR #23915, opened 2025-03-12
checker: add missing check for casting generic type to literal values

felipensp:
Fix bug mentioned by @kbkpbot


spytheman (approved):
Excellent work.

PR #23919, opened 2025-03-12
Justify signed integer literal overflow error when most significant bit occupies signed bit

### LINKED ISSUE #23782: Signed integer limits overflow cause unexpected results
V signed integers `i8` ... `i64` limits are those of unsigned versions. But signed range should be the half of the unsigned range because one bit goes for the sign.

### Go

As a comparison Golang rejects for the `int8` type values as large as `128` and `-129`.

```go
package main

import "fmt"

func main() {
	i := int8(128) // +0x80
	fmt.Println("Hello", i)
}
```

```
./prog.go:6:12: constant 128 overflows int8

Go build failed.
```

```go
package main

import "fmt"

func main() {
	i := int8(-129) // -0x81
	fmt.Println("Hello", i)
}
```

```
./prog.go:6:12: constant -129 overflows int8

Go build failed.
```

### V

V permits to create both `i8` positives and negatives in the full range of 8 bits each so for the 9 bits total window there are unexpected overflows:

```v
ints := [
	i8(0xff)  // converted to -1
	i8(128)   // converted to -128
	i8(-129)  // converted to +127
	i8(-0xff) // converted to +1
]

println('${ints}')

/*
_ := i8(0x100) // value `0x100` overflows `i8`
_ := i8(-0x100) // value `-0x100` overflows `i8`
*/
```

```
[-1, -128, 127, 1]
```

Is until a 9th bit is entered when V mark overflow as if it were unsigned that you can see in the commented section of the code above.

The same goes for V  signed types `i16` ... `i64`. I think V should limit the signed integers as Go already does.

> [!NOTE]
>

ChAoSUnItY:
Fix #23782.


spytheman (approved):
Excellent work.

Issue #23927, opened 2025-03-13
Generics and Custom Iterators


mecolosimo:
<details>
<summary>V version: V 0.4.9 9f3f129, press to see full `v doctor` output</summary>

|V full version      |V 0.4.9 2684ef9.9f3f129
|:-------------------|:-------------------
|OS                  |linux, Linux Mint 22
|Processor           |8 cpus, 64bit, little endian, AMD Ryzen 5 7520U with Radeon Graphics
|Memory              |0.66GB/14.86GB
|                    |
|V executable        |/home/mcolosimo/github/v/v
|V last modified time|2025-03-13 16:39:39
|                    |
|V home dir          |OK, value: /home/mcolosimo/github/v
|VMODULES            |OK, value: /home/mcolosimo/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mcolosimo/SyncProjects/mecolosimo/generics
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.11-25-g9f3f1291
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |N/A
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 0134e9b9
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.4) 2.39

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  src/main.v && src/main`
```v
module main

struct SquareIterator {
    arr []int
mut:
    idx int
}

fn (mut iter SquareIterator) next[K]() ?K {
    if iter.idx >= iter.arr.len {
        return none
    }
    defer {
        iter.idx++
    }
    return iter.arr[iter.idx] * iter.arr[iter.idx]
}

fn main() {
	nums := [1, 2, 3, 4, 5]
	iter := SquareIterator{
		arr: nums
	}

	for squared in iter {
		println(squared)
	}
}
```

**What did you see?**
```
[...]

PR #23932, opened 2025-03-14
 checker: check if next() method infers generic type correctly 

Delta456:
Fixes https://github.com/vlang/v/issues/23927


spytheman (approved):
Excellent work.

Issue #23928, opened 2025-03-14
The push() in datatypes.LinkedList[T] too slow


forchid:
### Describe the bug

I found a performance issue in `datatypes.LinkedList[T] push()`(but it's prepend() quick). After I reviewed it's code, found that there is a loop existing in `push()` for skipping to the list tail. This is where the problem lies. So I suggest adding a field `tail` in LinkedList[T] to solving this problem.

### Reproduction Steps

```v
for i in 0 .. 10_000_000 {
  list.push(i)
}
```

### Expected Behavior

Quickly as LinkedList[T].prepend().

### Current Behavior

Too slow ..

### V version

V 0.4.9 6623ac2

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.9 6623ac2
|:-------------------|:-------------------
|OS                  |windows, Microsoft Windows 10
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM)
|Memory              |1.97GB/7.84GB
|                    |
...
|V last modified time|2025-03-10 09:13:02
|                    |
...
|                    |
|Git version         |git version 2.47.0.windows.2
|V git status        |f802a8be
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |gcc 14.2.0
|clang version       |N/A
|msvc version        |N/A
|tcc version         |tcc version 0.9.27 (x86_64 Windows)
|tcc git status      |N/A
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #23934, opened 2025-03-14
vlib.datatypes: optimize linkedlist

kbkpbot:
Fix issue #23928

1. add a field `tail` in `LinkedList[T]`;
2. use `nil` instead of `0`, use `isnil` replace some `unsafe` blocks.
3. add `@[heap]`, `@[inline]`

list.v

```v
module main

import datatypes
import benchmark

fn main() {
        mut b := benchmark.start()
        mut list := datatypes.LinkedList[int]{}

        for i in 0 .. 10_000_000 {
                list.push(i)
        }
        b.measure('list.push')
}
```

```
v -prod list.v
./list
 SPENT   777.296 ms in list.push
```

### DIFF
--- a/vlib/datatypes/linked_list.v
+++ b/vlib/datatypes/linked_list.v
@@ -1,66 +1,61 @@
 module datatypes
 
+@[heap]
 pub struct ListNode[T] {
 mut:
 	data T
-	next &ListNode[T] = unsafe { 0 }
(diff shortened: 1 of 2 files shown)


medvednikov:
Thanks!

Just wondering, what was the old number?

```
v -prod list.v
./list
 SPENT   777.296 ms in list.push
 ```

PR #23938, opened 2025-03-15
vast: add --check flag

kbkpbot:
As `v.checker` may modify the AST tree, so introduce a new flag `--check` flag, which default is `false`. If set to `true`, then `vast` will product a checker-modified AST tree.

### DIFF
--- a/cmd/tools/vast/vast.v
+++ b/cmd/tools/vast/vast.v
@@ -100,9 +100,9 @@ fn (ctx Context) skip_empty(child &Node) bool {
 
 fn (ctx Context) write_file_or_print(file string) {
 	if ctx.is_print {
-		println(ctx.json(file))
+		println(json(file))
 	} else {
-		println('${time.now()}: AST written to: ' + ctx.json_file(file))
+		println('${time.now()}: AST written to: ' + json_file(file))
 	}
 }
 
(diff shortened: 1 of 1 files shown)


cmd/tools/vast/vast.v
```diff
@@ -43,6 +45,7 @@ fn main() {
 	ctx.is_compile = fp.bool('compile', `c`, false, 'watch the .v file for changes, rewrite the .json file, *AND* generate a .c file too on any change')
 	ctx.is_terse = fp.bool('terse', `t`, false, 'terse output, only with tree node names (AST structure), no details')
 	ctx.is_skip_defaults = fp.bool('skip-defaults', `s`, false, 'skip properties that have default values like false, 0, "", etc')
+	ctx.is_skip_checker = fp.bool('skip-checker', `t`, false, 'skip v.checker, as it will modify the AST tree')
```
spytheman, line 48: use a positive flag:
```suggestion
	ctx.check = fp.bool('check', `t`, false, 'run v.checker as well (it will modify the AST)')
```

spytheman: ... and later, just `if context.check { /* create and run a checker */ }`

spytheman: That will keep the current default (not break the current usage) of the tool, and be simpler to reason about (double negatives are harder to interpret, and easier to get wrong).

Delta456: I empathise on using `may modify the AST` instead of `will modify the AST`

spytheman (approved):
Excellent work.

PR #23945, opened 2025-03-15
`test/enum_on_array_init`: add edge test case 

Delta456:
Adds an edge test case from https://github.com/vlang/v/commit/725e25991d7445f5609958816abc3370a5c0dac3#commitcomment-153779001 which covers not using comma


spytheman:
Thanks.

PR #23960, opened 2025-03-17
crypto.ecdsa: improves internal `sign_digest` routine

blackshirt:
This PR improves `sign_digest`  internal routine used by `PrivateKey.sign` method. Its done by eleminating one step of `C.EVP_PKEY_sign` calls with null signature buffer to know the length (size) of required buffer. This maximal sizes of the buffer was already known by calling `C.EVP_PKEY_size`, so its not needed to call `C.EVP_PKEY_sign` just to know the size of required buffer and we can allocate the buffer instead and call `C.EVP_PKEY_sign` directly  with this allocated buffer.

In my benchmark test, there are small improvements on this part

```v
Benchmarking PrivateKey.sign() (before patch)...
Average PrivateKey.sign() (before patch) time: 37 µs
```

and after patch

```v
Benchmarking PrivateKey.sign() (after patch)...
Average PrivateKey.sign() (after patch) time: 27 µs
```

Thanks


spytheman (approved):
Excellent work.

Issue #8313, opened 2021-01-24
Different behavior of const


Ekopalypse:
const behaves differently when used in an executable or in a shared library.
Is this a Windows "feature" ?

**V version:** V 0.2.2 29d6e40, timestamp: 2021-01-24 10:54:27 +0100
**OS:** Windows7 x64

**What did you do?**
```
// shared library
module test

const (
	r = 1
	s = r + 1
)

[windows_stdcall]
[export: Main]
fn main() {
	println('r from dll: $r')
	println('s from dll: $s')
}


// executable
module main
import dl

type FUNCPTR = fn()

const (
	r = 1
	s = r + 1
)

fn main() {
	handle := dl.open('./testdll/test.dll', 0)
	f := FUNCPTR(dl.sym(handle, 'Main'))
	println('r from exe: $r')
	println('s from exe: $s')
	f()
}
```

**What did you expect to see?**
```
r from exe: 1
s from exe: 2
r from dll: 1
s from dll: 2
```
 
**What did you see instead?**
```
r from exe: 1
s from exe: 2
r from dll: 1
s from dll: 0
```

ntrel:
>	r = 1
> 	s = r + 1

`r` is known at compile-time, but currently `s` is evaluated when `main` is called by the OS. For a shared library, `dl.open` probably should be changed to call a function (`DllMain`?) in the library to initialize constants.

Ekopalypse:
@ntrel - I'm not sure I understand this correctly. How should I initialize consts in DllMain?
I have a function that is exported as DllMain and this function is called as soon as the program loads the dll, but as far as I know I cannot define const, for example, during the DLL_PROCESS_ATTACH step.
If this behavior is also related to a library not calling _vinit on Windows, then I assume this issue can be closed since I already reported this in another issue.

JalonSolov:
That would be the proper place for the const init to happen - in _vinit.

Ekopalypse:
okay, it seems I can workaround it by using the following (need to do more tests to be sure).
~~~
module test

fn C._vinit()

const (
	r = 1
	s = r + 1
)

[windows_stdcall]
[export: Main]
fn output() {
	println('r from dll: $r')
	println('s from dll: $s')
}

[windows_stdcall]
[export: DllMain]
fn main(hinst voidptr, fdw_reason int, lp_reserved voidptr) bool{
	// https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain
	// Perform actions based on the reason for calling.
[...]

PR #23961, opened 2025-03-17
dl: remove workaround; cgen: add DllMain()

kbkpbot:
Fix issue #8313

When create `.dll` under windows, create a default `DllMain()` for it, which will call `_vinit(0, (voidptr)0);` and `_vcleanup();`.
Then we can remove the workaround in `dl.open()` and `dl.close()`

This is helpful when try to create dll can be called by non-v APPs.

### DIFF
--- a/vlib/dl/dl_windows.c.v
+++ b/vlib/dl/dl_windows.c.v
@@ -20,25 +20,11 @@ type FN_vcleanup_caller = fn ()
 // open loads a given module into the address space of the calling process.
 pub fn open(filename string, flags int) voidptr {
 	res := C.LoadLibrary(filename.to_wide())
-	// Because LoadLibrary has no constructor, this is a workaround
-	if !isnil(res) {
-		vinit_caller := FN_vinit_caller(sym(res, '_vinit_caller'))
-		if !isnil(vinit_caller) {
-			vinit_caller()
-		}
-	}
 	return res
 }
 
 // close frees the loaded a given module.
 pub fn close(handle voidptr) bool {
-	// Because FreeLibrary has no destructor, this is a workaround
-	if !isnil(handle) {
-		vcleanup_caller := FN_vcleanup_caller(sym(handle, '_vcleanup_caller'))
-		if !isnil(vcleanup_caller) {
-			vcleanup_caller()
-		}
-	}
 	return C.FreeLibrary(handle)
 }
 
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -1082,6 +1082,9 @@ pub fn (mut g Gen) finish() {
 	g.handle_embedded_files_finish()
 	if g.pref.is_test {
 		g.gen_c_main_for_tests()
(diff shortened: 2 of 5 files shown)


Ekopalypse:
I can't understand how this is supposed to work in practice? How is the DLL informed that it has been loaded? In other words, when can the internal initialization take place?
Also, such a change is significant, why wasn't the current workaround announced as deprecated to detect any issues?
Yes, I know it wasn't V functionality - it was a workaround, but it worked ...

JalonSolov:
From the Windows docs:
```
An optional entry point into a dynamic-link library (DLL). When the system starts or terminates a process or thread, it calls the entry-point function for each loaded DLL using the first thread of the process. The system also calls the entry-point function for a DLL when it is loaded or unloaded using the LoadLibrary and FreeLibrary functions.
```
For a DLL created by V, this function is _always_ needed, so that `_vinit()` is called - this is the routine that initializes V's `const`s, etc.

Since this was not created by default in the past, people who ran into problems had to create their own DllMain() that called _vinit(), or use some sort of other hacky workaround.

If there was a "standard" workaround, then I agree, it should've been marked deprecated.  I don't recall there being such a standard, but I don't do much with Windows (except complain about it).

Ekopalypse:
I know that, but the thing is that this PR killed all dlls that need to do an internal initialization, because with this PR the "V-internal" DllMain is called and thus the author of a dll can no longer tell when this happens. 
I know I can work around that too, but then it's a different, and this time really hacky way to get it working again.
I already pointed out in the issue that this, a way to define how the dll's own code can be initialized, needs to be considered.

> If there was a "standard" workaround ...

I would argue that the previous way was the standard workaround

JalonSolov:
I think what should've been done was to make this auto DllMain() _optional_, only to be generated if there was no existing DllMain().  That way, if you wanted to do your own initialization, you would write your own DllMain().

This change is _necessary_ if there is no DllMain() already, otherwise, DLLs created by V just silently don't work as expected.  Of course, if you provide your own DllMain() and don't call _vinit(), you'll still get the same problem, but at least it could then be documented, and your own fault if you didn't read the docs.

One could argue that doing initialization any other way than with a DllMain() goes against the Windows specs, so is definitely _not_ standard.

Ekopalypse:
> I think what should've been ... _optional_ ...

This would have been a possibility, yes.

> This change is _necessary_ 

I would not say that _this_ change is necessary, but yes, a change is necessary.
A change to the documentation would also have done the trick too.
 
> One could argue that doing initialization any other way than with a DllMain() goes against the Windows specs, so is definitely _not_ standard.

I don't know what is being referred to here. To my comment that the previous way was the quasi-standard? If so, then you have overlooked the fact that the previous way ultimately did exactly the same thing as this PR.
If you mean that V absolutely has to define a DllMain function to meet the Windows standard, then no, the Windows documentation explicitly says that the name DllMain is just a placeholder and therefore not mandatory.  From the V point of view, you could define a separate entry point where vinit is run, just like for executable programs. Furthermore, a dll-specific function should definitely be called, which allows a dll author to perform dll-specific initialization. In other words, the V-own entry point should call a dll-specific function or functions with the same parameters.

Issue #23965, opened 2025-03-17
Operator `+=` is not implemented for generic type alias `type V2 = vec.Vec2[f32]`, even though `vlib/math/vec/vec2.v:75` implements it for the parent type `Vec2[T]`


spytheman:
<details>
<summary>V version: V 0.4.9 21874f9, press to see full `v doctor` output</summary>

|V full version      |V 0.4.9 c69b1251624fb559d844bf92fc8d5963deaac8f1.21874f9
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 20.04.6 LTS
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-3225 CPU @ 3.30GHz
|Memory              |0.42GB/15.05GB
|                    |
|V executable        |/home/delian/v/v
|V last modified time|2025-03-17 07:08:53
|                    |
|V home dir          |OK, value: /home/delian/v
|VMODULES            |OK, value: /home/delian/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/delian/v
|                    |
|env VFLAGS          |"-no-parallel"
|                    |
|Git version         |git version 2.48.1
|V git status        |weekly.2025.11-48-g5978cbcf-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
|gcc version         |gcc (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
|clang version       |clang version 10.0.0-4ubuntu1 
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 0134e9b9
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.31-0ubuntu9.16) 2.31

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  b.v && b`
```v
import math.vec

type V2 = vec.Vec2[f32]

///// When this is uncommented, the program compiles and runs properly
//fn (a V2) + (b V2) V2 { return a.add(b) }

fn main() {
	a := V2{1,1}
	b := V2{10,20}
	mut c := a + b
	dump(c)
	c += a
	dump(c)
	assert c.x == 12
	assert c.y == 22
}
	

```

**What did you see?**
```
[...]

PR #23967, opened 2025-03-17
checker cgen: implement alias operator overloading for generic struct parent type

Delta456:
Fixes https://github.com/vlang/v/issues/23965


spytheman (approved):
Excellent work.

PR #23975, opened 2025-03-19
db.mysql: add SSL support; turn ConnectionFlag into @[flag]

kbkpbot:
This PR make ssl connection work.

Tested with `tidbcloud.com` mysql db.

And turn ConnectionFlag into @[flag], so we can write code like this:

```v
	mut mysql_config := mysql.Config{
		host:     '127.0.0.1'
		port:     4000
		username: 'root'
		password: '123'
		dbname:   'admin'
		ssl_ca : './client-cert.pem'
		flag:     .client_ssl | .client_ssl_verify_server_cert
	}
```


spytheman (approved):
Excellent work.

Avey777:
The importance was resolved.😁

PR #23988, opened 2025-03-20
cgen: fix codegen to make mutable sumtype working

### LINKED ISSUE #23982: Mutable sum types passed to a function can't be modified
### Describe the bug

We pass mutable sum types to a function to update the types fields and cannot be modified.

### Reproduction Steps

```v
struct Int { mut: i int }

struct String { mut: s string }

type Sum = Int | String

fn init(mut s Sum) {
	match mut s {
		Int    { s.i = 0 }
		String { s.s = '' }
	}
}

fn init_int(mut i Int) {
	i.i = 0
}

fn main() {
	mut i := Int{ i: 333 }
	mut s := String{ s: 'string' }
	init(mut i)
	init(mut s)
	println('${i.i}') // prints 333, should be 0
	println('${s.s}') // prints 'string', should be ''
	init_int(mut i)
	println('${i.i}') // prints 0 as expected
}
```

Function `init` above tries to initialize every sum type received but can't do it. Function `init_int` works.

### Expected Behavior

Mutable fields passed to init function should be set to zero for integer sum type and empty for string sum type.

### Current Behavior

Mutable fields passed to init function are not set to 0 or empty string.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.9 86bb2d7

https://play.vlang.io/p/cfbb7907ac

> [!NOTE]
>

felipensp:
Fix #23982

### DIFF
--- a/vlib/v/gen/c/assign.v
+++ b/vlib/v/gen/c/assign.v
@@ -127,8 +127,8 @@ fn (mut g Gen) expr_opt_with_cast(expr ast.Expr, expr_typ ast.Type, ret_typ ast.
 				if ret_sym.kind == .sum_type {
 					exp_sym := g.table.sym(expr_typ)
 					fname := g.get_sumtype_casting_fn(expr_typ, ret_typ)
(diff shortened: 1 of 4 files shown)


spytheman (approved):
Excellent work.

Issue #23990, opened 2025-03-20
[os] set_redirect_stdio() cause process hang when child process output large text


kbkpbot:
### Describe the bug

When try to use `set_redirect_stdio()` get the output of a child process, it will hang when child process try to output large text.

### Reproduction Steps

create two file:

sub.v
```v
module main
fn main() {
        // Windows' limit is 4095
        // Linux's limit is 65535
	x := '1'.repeat(65536)    // change to 65535 will not hang under Linux
	println(x)
}
```

pp.v
```v
module main
import os

fn main() {
	mut p := os.new_process('./sub')
	p.set_redirect_stdio()
	p.wait()
	assert p.code == 0
	output := p.stdout_slurp().trim_space()
	p.close()
	dump(output)
}
```

then compile files:

```
v sub.v
v pp.v
```

run `pp`

```
./pp
```

### Expected Behavior

dump the output child:
```
1111111111111111111111111111111111111111111111....11111
```

### Current Behavior

run `./pp` will just hang.

### Possible Solution

modify `repeat()` to a smaller number.
Maybe use a separate thread read the stdout/stderr, before we wait the child finish. 

### V version

V 0.4.10 21874f93dcb168395b4b716df6d8bde2c266e201.a9a96b8

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 21874f93dcb168395b4b716df6d8bde2c266e201.a9a96b8
|:-------------------|:-------------------
|OS                  |windows, Microsoft Windows 10 企业版 LTSC 19044 64 位
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 7840H with Radeon 780M Graphics
|Memory              |1.32GB/27.69GB
|                    |
|V executable        |D:\v\v\v\v.exe
|V last modified time|2025-03-20 13:20:58
|                    |
|V home dir          |OK, value: D:\v\v\v
|VMODULES            |OK, value: C:\Users\DDT\.vmodules
|VTMP                |OK, value: C:\Users\DDT\AppData\Local\Temp\v_0
|Current working dir |OK, value: D:\v\bug\process
|                    |
|Git version         |git version 2.43.0.windows.1
|V git status        |0.4.10-1-ga9a96b89-dirty
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |N/A
|clang version       |N/A
|msvc version        |用于 x64 的 Microsoft (R) C/C++ 优化编译器 19.39.33523 版
[...]

PR #23993, opened 2025-03-21
os: windows add is_pending()

kbkpbot:
Fix issue #23990

1. add `win_is_pending()`;
2.  modify windows stdout pipe / stderr pipe buffer length from 4KB to 64KB, this is the same of Linux version.


spytheman (approved):
Excellent work.
Thank you 🙇🏻‍♂️.

Issue #10628, opened 2021-06-30
process.set_environment() not working in windows


axetroy:
**V version:**  V 0.2.2 9d02ca5
**OS:** widnows

**What did you do?**

main.v

```v
	mut ps := os.new_process(python_path)
	py_file_path := os.join_path(cwd, 'process', 'test_data', 'server.py')
	ps.set_args([py_file_path])
	ps.set_environment(map{
		'PORT': port
	})
	ps.wait()

	if ps.code != 9 && ps.code != 15 {
		return error('process exit with code $ps.code')
	}
```

server.py

```python
import http.server
import socketserver
from http import HTTPStatus
from os import getpid, environ
import atexit
import sys

port = 9999

try:
    port = int(environ["PORT"])
except KeyError:
    print("Please set the environment variable PORT")
    sys.exit(1)

class Handler(http.server.SimpleHTTPRequestHandler):
    def do_GET(self):
        self.send_response(HTTPStatus.OK)
        self.send_header("x-pid", getpid())
        self.end_headers()
        self.wfile.write(b'Hello world')

print("listen on port " + str(port))

def exit_handler():
    print('http server exit!')

atexit.register(exit_handler)

httpd = socketserver.TCPServer(('0.0.0.0', port), Handler)
httpd.serve_forever()
```

**What did you expect to see?**

The environment variables of the child process should be set correctly.

Linux and macOS work fine but windows not.
 
**What did you see instead?**

Run the python program and found that the environment variable cannot be read.

`set_environment()` seem did not works. and python throw error **Please set the environment variable PORT**

PR #23996, opened 2025-03-21
os: windows add set_environment() support

kbkpbot:
Fix issue #10628

Update `process_test.v`, then it work now.

### DIFF
--- a/vlib/os/process_windows.c.v
+++ b/vlib/os/process_windows.c.v
@@ -124,8 +124,38 @@ fn (mut p Process) win_spawn_process() int {
 		to_be_freed << work_folder_ptr
 	}
 
+	mut env_block := []u16{}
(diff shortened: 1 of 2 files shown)


vlib/os/process_windows.c.v
```diff
@@ -124,8 +124,38 @@ fn (mut p Process) win_spawn_process() int {
...
+				i++
+			}
+			env_block << u16(0)
+			to_be_freed << env_ptr
+		}
+		env_block << u16(0)
+		creation_flags |= C.CREATE_UNICODE_ENVIRONMENT
+		defer {
+			unsafe { env_block.free() }
+		}
```
spytheman, line 153: please move this defer right after `mut env_block := []u16{}`

Issue #23976, opened 2025-03-19
Compiler bug: math operation with number|string|any with a function that return void causes C error instead of type error


AQMpolyface:
<details>
<summary>V version: V 0.4.9 87b1de8, press to see full `v doctor` output</summary>

|V full version      |V 0.4.9 f83af8a.87b1de8
|:-------------------|:-------------------
|OS                  |linux, "CachyOS"
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 Pro 7735U with Radeon Graphics
|Memory              |2.82GB/30.12GB
|                    |
|V executable        |/home/polyface/git/v/v
|V last modified time|2025-03-19 15:28:03
|                    |
|V home dir          |OK, value: /home/polyface/git/v
|VMODULES            |OK, value: /home/polyface/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/polyface/Documents/v/bug
|                    |
|Git version         |git version 2.49.0
|V git status        |weekly.2025.12-6-g87b1de82
|.git/config present |true
|                    |
|cc version          |cc (GCC) 14.2.1 20250207
|gcc version         |gcc (GCC) 14.2.1 20250207
|clang version       |clang version 19.1.7
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 0134e9b9
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.41

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  src/main.v && src/main`
```v
module main

fn main() {
	a := 1 / println('')
}

```

**What did you see?**
```
src/main.v:4:2: warning: unused variable: `a`
    2 | 
    3 | fn main() {
    4 |     a := 1 / println('')
      |     ^
    5 | }
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/main.01JPQG9QA1CVBXSRKHXKXJNW00.tmp.c:1897: error: declaration of void object
=================================================================
(You can pass `-cg`, or `-show-c-output` as well, to print all the C error messages).
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```

**What did you expect to see?**

[...]

PR #24001, opened 2025-03-21
 checker: error if right side of infix is void 

Delta456:
Fixes https://github.com/vlang/v/issues/23976


spytheman (approved):
Excellent work.
Thank you 🙇🏻‍♂️ .

PR #24004, opened 2025-03-22
net: make `close/0`, `select/2` and `remote/0` methods of `UdpSocket` `pub`

larpon:
The return signature of `remote/0` is also changed to use an `Option` (`none`) instead of a `Result` (`error('none')`). This oddity is likely a workaround for older versions of V.


spytheman:
> This oddity is likely a workaround for older versions of V.

It is the result of the `!?Type` split.

PR #24006, opened 2025-03-22
cgen: cleanup write_sumtype_casting_fn

felipensp:
Removed unneeded comment


spytheman:
Excellent work, good find.

PR #24020, opened 2025-03-23
ops decode.v

enghitalo:


Delta456:
Why?

JalonSolov:
`mut &var` is useless.  `mut var` or `&var`, but not both together.  `mut var` (the V way) is implicit `&var` (the C way), since it has to be passed by reference, or the original can't be modified.

PR #24030, opened 2025-03-24
cgen: fix codegen for option return unwrapping on last statement

### LINKED ISSUE #24026: C error when `option` is propagated in `match` as its expression value when operators are used
### Describe the bug

C error when `option` is propagated in `match` as its expression value when operators are used

### Reproduction Steps

```v
fn main() {
	dump(some_func2(12)? + 'c')
}

fn some_func(i int) ?string {
	return 'a'
}

fn some_func2(i int) ?string {
	return match i {
		12 { some_func(1)? + 'b' }
		else { none }
	}
}
```

### Expected Behavior

```
abc
```

### Current Behavior

```
cc: C:/Users/***/AppData/Local/Temp/v_0/test.01JQ2BM2T5X85ZGSP3WBFHYFMH.tmp.c:3817: warning: implicit declaration of function 'tcc_backtrace'
cc: C:/Users/***/AppData/Local/Temp/v_0/test.01JQ2BM2T5X85ZGSP3WBFHYFMH.tmp.c:5860: error: identifier expected
...
cc: C:/Users/***/AppData/Local/Temp/v_0/test.01JQ2BM2T5X85ZGSP3WBFHYFMH.tmp.c:5860: error: identifier expected
(note: the original output was 3 lines long; it was truncated to its first 2 lines + the last line)
builder error:
==================
```

### Possible Solution

_No response_

The reason is that an arg is missing, thus resulting in the following C code:
```c
_option_ok(&(string[]) { string__plus(, _SLIT("b")) }, (_option*)(&_t4), sizeof(string));
```
If we replace `string` with `int`, and `'a'`, `'b'`, `'c'` with `1`, `2`, `3`, the code can compile and run, but with wrong results(`5` instead of `6`), because the following code is generated:
```c
_option_ok(&(int[]) { (int)( + 2) }, (_option*)(&_t2), sizeof(int));
```
This code works for `int`, but it is wrong code

### V version

V 0.4.10 1f07431

|V full version      |V 0.4.10 725e25991d7445f5609958816abc3370a5c0dac3.1f07431
|Processor           |12 cpus, 64bit, little endian, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz  
|Memory              |2.39GB/15.86GB
|                    |
|V executable        |E:\Programs\v\v.exe
|V last modified time|2025-03-23 20:12:31
|                    |
|V home dir          |OK, value: E:\Programs\v
|VMODULES            |OK, value: C:\Users\***\.vmodules
|VTMP                |OK, value: C:\Users\***\AppData\Local\Temp\v_0
|Current working dir |OK, value: ***
|                    |
|Git version         |git version 2.38.1.windows.1
|V git status        |0.4.10-24-g1f07431e
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |N/A
|clang version       |N/A
|msvc version        |N/A
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

felipensp:
Fix #24026

### DIFF
--- a/vlib/v/parser/parser.v
+++ b/vlib/v/parser/parser.v
@@ -601,6 +601,20 @@ fn (mut p Parser) mark_last_call_return_as_used(mut last_stmt ast.Stmt) {
 						}
 					}
 				}
+				ast.InfixExpr {
+					// last stmt has infix expr with CallExpr: foo()? + 'a'
+					mut left_expr := last_stmt.expr.left
+					for {
+						if mut left_expr is ast.InfixExpr {
+							left_expr = left_expr.left
+							continue
+						}
+						if mut left_expr is ast.CallExpr {
+							left_expr.is_return_used = true
+						}
+						break
+					}
+				}
 				else {}
 			}
 		}
--- a/vlib/v/tests/options/option_last_call_test.v
+++ b/vlib/v/tests/options/option_last_call_test.v
@@ -1,0 +1,16 @@
+fn test_main() {
+	assert some_func2(14)? + 'c' == 'aabcac'
+}
+
+fn some_func(i int) ?string {
+	return 'a'
+}
+
+fn some_func2(i int) ?string {
+	return match i {
+		12 { some_func(1)? + 'b' }
+		13 { some_func(1)? + 'b' + 'c' }
+		14 { 'a' + some_func(1)? + 'b' + 'c' + some_func(1)? }
+		else { none }
+	}
+}


vlib/v/parser/parser.v
```diff
@@ -601,6 +601,20 @@ fn (mut p Parser) mark_last_call_return_as_used(mut last_stmt ast.Stmt) {
 						}
 					}
 				}
+				ast.InfixExpr {
```
spytheman, line 604: the whole `mark_last_call_return_as_used` walking should happen in the checker, not in the parser

spytheman (approved):
Excellent work.

PR #24033, opened 2025-03-24
os: fix swap_bytes_u64

tankf33der:
Found via s390x port.


spytheman:
Please add a test case too in `file_le_be_test.v`, to prevent future regressions.

spytheman:
I should have probably used the same approach as in `reverse_bytes_u64` in `vlib/net/conv/conv.v` 🤔 , using `vfmt off/vfmt on` to keep the stair aligned:
```v
    // vfmt off
    return ((a >> 56) & 0x00000000_000000FF) |
           ((a >> 40) & 0x00000000_0000FF00) |
           ((a >> 24) & 0x00000000_00FF0000) |
           ((a >> 8)  & 0x00000000_FF000000) |
           ((a << 8)  & 0x000000FF_00000000) |
           ((a << 24) & 0x0000FF00_00000000) |
           ((a << 40) & 0x00FF0000_00000000) |
           ((a << 56) & 0xFF000000_00000000)
    // vfmt on
```

PR #24034, opened 2025-03-24
v.vmod: add manifest encoder fn

gechandesu:
This PR:

- Adds encoder function for v.mod
- Adds tests

### DIFF
--- a/vlib/v/vmod/encoder.v
+++ b/vlib/v/vmod/encoder.v
@@ -1,0 +1,69 @@
+module vmod
+
+import strings
+
+fn quote(input string) string {
+	if input.contains("'") {
+		return '"' + input + '"'
+	}
+	return "'" + input + "'"
+}
+
+fn encode_array(mut b strings.Builder, input []string) {
+	if input.join('').len > 60 {
+		b.writeln('[')
+		for item in input {
+			b.write_string('\t\t')
+			b.write_string(quote(item))
+			b.writeln(',')
+		}
+		b.writeln('\t]')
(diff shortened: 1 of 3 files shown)


spytheman:
Why is it needed?

gechandesu:
> Why is it needed?

I plan to use it in vpm for editing v.mod (with `v install`, `v remove`), it can also be used in scaffolding utilities like `v init`.

spytheman:
Done, although it may need rebasing over the current master too.

spytheman:
It failed again, I will rebase it.

gechandesu:
Hmm, failing again, I have a look bit later. Make MR a draft for now

vlib/v/vmod/encoder.v
```diff
@@ -0,0 +1,63 @@
...
+	if input.contains("'") {
+		return '"' + input + '"'
+	}
+	return "'" + input + "'"
+}
+
+pub fn encode(manifest Manifest) string {
+	mut b := strings.new_builder(512)
+	b.writeln('Module {')
+	$for field in manifest.fields {
```
spytheman, line 15: imho just use a runtime for, instead of `$for`; the fields are a fixed set, and there is no need for the additional flexibility of a general comptime construct.

spytheman: Code, that is part of the compiler, should be as simple as possible, on general principle, that helps a lot with bootstrapping.

spytheman (approved):
Excellent work.

Issue #23904, opened 2025-03-11
[x.crypto.chacha20] 64bit internal counter


einar-hjortdal:
### Describe the feature

*chacha20* uses a 32bit internal counter which limits the stream to 256GB of data. To work around this limitation of the [IETF RFC8439](https://datatracker.ietf.org/doc/html/rfc8439) algo, many programs and libraries decided to utilize a 64bit counter instead.

I request this non-standard variant be available. If not in vlib I request help, from those who know how chacha20 works, in order to implement this variant as a third party library so that the V language community can make use of it.

libtomcrypt, libsodium, bouncy castle, and more crypto libraries offer chacha20 with 64bit nonce and 64bit counter. What changes should be made to the current implementation to make it work the same way?

Refs:

- [libsodium](https://doc.libsodium.org/advanced/stream_ciphers/chacha20)
- [libtomcrypt](https://github.com/libtom/libtomcrypt)
- [Bouncy Castle Java](https://github.com/bcgit/bc-java)
- [firebird #7065](https://github.com/FirebirdSQL/firebird/issues/7065)

### Use Case

Connect to services that use non-IETF-standard chacha20 cipher

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.9 5f5e48e

### Environment details (OS name and version, etc.)

```
|V full version      |V 0.4.9 537605a
|:-------------------|:-------------------
|OS                  |linux, "openmamba release 2024.6 for x86_64 (rolling)"
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz
|Memory              |4.63GB/31.17GB
|                    |
|V executable        |/home/einar/.local/lib64/v/v
|V last modified time|2025-02-25 15:36:48
|                    |
|V home dir          |OK, value: /home/einar/.local/lib64/v
|VMODULES            |OK, value: /home/einar/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/einar/Documents/projects/vlang/active/firebird
|                    |
|Git version         |git version 2.48.1
|V git status        |weekly.2025.09-2-g537605a0 (17 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |cc (GCC) 14.2.1 20240909
|gcc version         |gcc (GCC) 14.2.1 20240909
|clang version       |N/A
[...]

PR #24039, opened 2025-03-25
x.crypto.chacha20: changes internal cipher representation to support 64 bit ctr

blackshirt:
Related to this [issues](https://github.com/vlang/v/issues/23904), here is the plan to add support for original ChaCha20 stream cipher with 64 bit counter. 
- The changes was needed to the internal `chacha20.Cipher` representation (was done in this PR). Its changed to the raw 128 bytes of```[4]u32``` opaque where the counter was contained internally and handles transparently.  Previously, its separated in the `nonce` and `counter` of the `Cipher` field's.
- Adding 64 bit counter handling and related keystream generation that aware of new size of counter (not ready)
- Make stream cipher creation support for 64-bit counter variant. Its partially was done in this PR, but still not fully supported.
- Make `xor_key_stream` and other routines awares for 64-bit counter variant
- Adds 64-bit counter test 
- Integrates the changes with the rest 

This PR contains some bits of them, but its not fully activated support for 64-bit counter variant. 
The fundamental changes in this PR was the change of internal representation of the `nonce` and `counter` of `chacha20.Cipher` field's into raw `nonce` and make enrcyption routines work and aware of this changes..
Its also contains some cleansup bits and code safety checking

Thanks

### DIFF
--- a/vlib/x/crypto/chacha20/chacha.v
+++ b/vlib/x/crypto/chacha20/chacha.v
@@ -9,12 +9,14 @@ import math.bits
 import crypto.internal.subtle
 import encoding.binary
 
-// size of ChaCha20 key, ie 256 bits size, in bytes
+// The size of ChaCha20 key, ie 256 bits size, in bytes
 pub const key_size = 32
-// size of ietf ChaCha20 nonce, ie 96 bits size, in bytes
+// The size of standard IETF ChaCha20 nonce, ie 96 bits size, in bytes
 pub const nonce_size = 12
-// size of extended ChaCha20 nonce, called XChaCha20, 192 bits
+// The size of extended variant of standard ChaCha20 (XChaCha20) nonce, 192 bits
 pub const x_nonce_size = 24
+// The size of original ChaCha20 nonce, 64 bits
+pub const orig_nonce_size = 8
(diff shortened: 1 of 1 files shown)


vlib/x/crypto/chacha20/chacha.v
```diff
@@ -24,19 +26,39 @@ const cc1 = u32(0x3320646e) // nd 3
 const cc2 = u32(0x79622d32) // 2-by
 const cc3 = u32(0x6b206574) // te k
 
+// CipherMode was enumeration of ChaCha20 supported variant.
+enum CipherMode {
+	// The standard IETF ChaCha20 (and XChaCha20), with 32-bit internal counter.
+	standard
+	// The original ChaCha20 with 64-bit internal counter.
+	original
```
spytheman, line 34: Why name it `original`, instead of `extended`?

spytheman: although that is a private enum, and thus it does not matter; it can be changed without breaking the public API

blackshirt: `extended` refers to `xchacha20` construction already supported from the beginning. Basically its a standard version with bigger nonce size,

PR #24045, opened 2025-03-25
crypto.sha3: be big-endian friendly

tankf33der:
And as a result of these change, the results of sha3 will be the same across all systems, including s390x.


spytheman (approved):
Perfect. Less code, with more platforms supported 🥳.

medvednikov:
Can you please explain why this change works for big endian systems?

spytheman:
Because it always writes in the same order.

spytheman:
The methods that are fixed by the PR are `to_bytes` and `from_bytes`, i.e. they are serialization ones.

medvednikov:
I see, thanks.

Issue #24043, opened 2025-03-25
x.crypto.chacha20 xor_key_stream fails after a while


einar-hjortdal:
### Describe the bug

x.crypto.chacha20 xor_key_stream works as expected for a while, then fails to decode after a few consecutive calls.

### Reproduction Steps

```v
module main

import x.crypto.chacha20

const key = [u8(225), 2, 1, 178, 238, 127, 187, 188, 27, 237, 18, 62, 181, 65, 67, 152, 13, 247,
	147, 148, 101, 220, 185, 120, 234, 58, 144, 173, 3, 218, 193, 130]
const nonce = [u8(153), 221, 244, 134, 99, 135, 243, 247, 169, 121, 69, 54]
const encoded_data = [
	[u8(231), 121, 9, 28],
	[u8(178), 221, 62, 9, 153, 189, 106, 12, 117, 47, 192, 81, 65, 112, 85, 57],
	[u8(155), 202, 56, 16],
	[u8(227), 47, 226, 137],
	[u8(162), 77, 218, 52],
	[u8(42), 250, 184, 196],
	[u8(2), 129, 13, 136, 6, 12, 235, 183, 38, 178, 151, 243, 27, 88, 97, 40],
	[u8(248), 170, 168, 206],
	[u8(181), 220, 223, 139],
	[u8(95), 108, 201, 227],
	[u8(38), 221, 147, 230],
	[u8(98), 229, 5, 130, 13, 103, 248, 159, 240, 246, 56, 119, 160, 130, 82, 222],
]

// expected data obtained with an equivalent Java program
const expected_data = [
	[u8(0), 0, 0, 9],
	[u8(0), 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
	[u8(0), 0, 0, 1],
	[u8(0), 0, 0, 0],
	[u8(0), 0, 0, 0],
	[u8(0), 0, 0, 9],
	[u8(0), 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
	[u8(0), 0, 0, 1],
	[u8(0), 0, 0, 0],
	[u8(0), 0, 0, 0],
	[u8(0), 0, 0, 9],
	[u8(0), 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
]

fn main() {
	mut cipher := chacha20.new_cipher(key, nonce)!
[...]

PR #24046, opened 2025-03-25
`x.crypto.chacha20`: fix for xor_key_stream fails after a while

blackshirt:
This PR fixes bug in [24043](https://github.com/vlang/v/issues/24043) issues.
The problem comes from the fact that we forgot to clear out of  internal buffer for partial block src within `xor_key_stream` operation.

Thanks


spytheman (approved):
Excellent work.

Issue #24052, opened 2025-03-26
[orm] C error while field default value include ""


kbkpbot:
### Describe the bug

When create a field with default value include `""` in it, it will generate a  C error.

### Reproduction Steps

o.v
```v
module main

import db.mysql


fn main() {
        db := db_mysql() or { panic('failed to connect to database') }

        sql db {
                create table User
        } or { panic(err) }
}

@[table: 'sys_users']
struct User {
pub:
        home_path ?string @[default: '"dashboard"'; sql_type: 'VARCHAR(64)']   // "dashboard" will compile fail.
}

fn db_mysql() !mysql.DB {

        mut mysql_config := mysql.Config{
                host:     '127.0.0.1'
                port:     3306
                username: 'root'
                password: '12345678'
                dbname:   'vcore'
        }

        mut conn := mysql.connect(mysql_config) or {
                return err
        }

        return conn
}

```

```
v o.v
```

### Expected Behavior

compile ok

### Current Behavior

```
$ v o.v
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/o.01JQ9CGNBWZD724RNJ8FXMAJF7.tmp.c:10572: warning: assignment discards qualifiers from pointer target type
cc: /tmp/v_1000/o.01JQ9CGNBWZD724RNJ8FXMAJF7.tmp.c:10612: error: ')' expected (got "dashboard")
...
cc: /tmp/v_1000/o.01JQ9CGNBWZD724RNJ8FXMAJF7.tmp.c:10612: error: ')' expected (got "dashboard")
(note: the original output was 4 lines long; it was truncated to its first 2 lines + the last line)
=================================================================
(You can pass `-cg`, or `-show-c-output` as well, to print all the C error messages).
builder error:
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .

```

### V version

V 0.4.10 bd2ec679f0325e1f78cb446b213d6d3182aa9c31.f7f89b3

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 bd2ec679f0325e1f78cb446b213d6d3182aa9c31.f7f89b3
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.2 LTS
[...]

PR #24057, opened 2025-03-27
orm:fix default value quote

kbkpbot:
Fix issue #24052


vlib/v/gen/c/orm.v
```diff
@@ -177,9 +177,11 @@ fn (mut g Gen) write_orm_create_table(node ast.SqlStmtLine, table_name string, c
 				for attr in field.attrs {
 					g.write('(VAttribute){')
 					g.indent++
-					g.write(' .name = _SLIT("${attr.name}"),')
+					g.write(' .name = _SLIT("${cescape_nonascii(util.smart_quote(attr.name,
```
spytheman, line 180: Try extracting `aname := cescape_nonascii(util.smart_quote(attr.name,
						false))` and then use `${aname}`, to workaround the -autofree problem. If that does not help, try using more temporary local variables for the inner calls too.

vlib/v/gen/c/orm.v
```diff
@@ -177,9 +177,11 @@ fn (mut g Gen) write_orm_create_table(node ast.SqlStmtLine, table_name string, c
...
+						false))}"),')
 					g.write(' .has_arg = ${attr.has_arg},')
-					g.write(' .arg = _SLIT("${attr.arg}"),')
+					g.write(' .arg = _SLIT("${cescape_nonascii(util.smart_quote(attr.arg,
```
spytheman, line 183: same, try simplifying the expression, by using more temporary locals for the intermediate results

spytheman (approved):
Excellent work.

PR #24061, opened 2025-03-27
cgen: multithreading support for the new profiler column

Eliyaan:
Fix the calculations for the new profiler column by using a thread local.  (works with gcc)
Known issues: does not fix for tcc as it does not provide thread locals


spytheman (approved):
Excellent work.

Issue #15478, opened 2022-08-21
ORM does not recognise type alias


OliverCWY:
It compiles correctly if I replace ```Connection``` with ```sqlite.DB```.

Additionally, if I use the orm interface orm.Connection for easier port of database, it shows the same problem.

**V doctor:**
```
OS: linux, Ubuntu 18.04.6 LTS (VM)
Processor: 2 cpus, 64bit, little endian, AMD EPYC 7551 32-Core Processor
CC version: cc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0

getwd: /home/ubuntu/workspace/im_server
vmodules: /home/ubuntu/.vmodules
vroot: /home/ubuntu/v
vexe: /home/ubuntu/v/v
vexe mtime: 2022-08-20 23:52:35
is vroot writable: true
is vmodules writable: true
V full version: V 0.3.0 c356882.eb6627f

Git version: git version 2.17.1
Git vroot status: weekly.2022.33-38-geb6627fe
.git/config present: true
thirdparty/tcc status: thirdparty-linux-amd64 827f7452

```

**What did you do?**
`v -g -o vdbg cmd/v && vdbg backend/host.v`
```v
module main

// import orm
// type Connection = orm.Connection

import sqlite
type Connection = sqlite.DB

pub struct User {
	pub:
		id		int			[primary; sql: serial]
		name	string
}

pub type Content = []u8 | string

struct Host {
	pub: 
		db Connection
}

pub fn (back Host) get_users() []User {
	return []
}

pub fn create_host(db Connection) Host {
	
	sql db {
		create table User
	}

	return Host {
		db: db
	}
}

fn main(){
	println(create_host(sqlite.connect("./data.sqlite3")?))
}
```

**What did you expect to see?**

It should compile correctly since Connection is a alias of sqlite.DB

**What did you see instead?**
```
cgen error: This database type `unknown` is not implemented yet in orm
```

enghitalo:
@enghitalo

PR #24062, opened 2025-03-27
orm: fix type alias not supported in table columns

mid1221213:
Corollary to #15478, which fixes the type alias support for db type, this commit allows the use of an alias to a supported type in a column.

The fix is mainly replacing `field.typ` occurrences with `table.final_type(field.typ)` or equivalent.

Example use:

```
type CountryCode = string

@[table: 'addresses']
pub struct Address {
	id      int @[primary; sql: serial]
	name    string
	address string
	country CountryCode
}

// …

sql db {
	create table Address
} or { panic("don't panic!") }
```

Created as a draft because all comments welcome and and tests / doc not yet written because of this.


spytheman:
Please add a test, that checks the intended effect, and that fails on master.

PR #24065, opened 2025-03-28
veb:HTTP status response API.

KeitoTobi1:
```
import os
import veb
import veb.status { Status, status }

pub struct RespStatus {
	Status
}

@[heap]
pub struct App {
	veb.Controller
}

pub struct Context {
	veb.Context
}

@['/'; get]
pub fn (mut app App) status_handle() veb.Result {
	resp := RespStatus{
		Status: status('Success.', .ok)
	}
	return ctx.json(resp)
}

pub fn main() {
	os.chdir(os.dir(@FILE))!
	mut app := &App{}
	veb.run[App, Context](mut app, 8082)
}
```


spytheman:
Hm, why have a separate module, for one function and a struct?

vlib/veb/status/status.v
```diff
@@ -0,0 +1,25 @@
...
+	status string
+}
+
+// status returns a new Status struct with the given message, status, and code
+pub fn status(msg string, status http.Status) Status {
+	return Status{
+		msg:    msg
+		resp:   status.str()
+		code:   status.int()
+		status: 'success'
```
spytheman, line 23: Why is it always `success` here?
What happens for `status` != .ok ?

spytheman:
Please also add a unit test for the new public function, passing different statuses.

vlib/veb/status/status.v
```diff
@@ -0,0 +1,25 @@
...
+// Status is a struct that contains the message, status, code, and status of the request
+pub struct Status {
+	// msg is the message that will be returned to the user
+	msg string
+	// resp is the status that will be returned to the user
+	resp string
+	// code is the status code that will be returned to the user
+	code int
+	// status is the status of the request
+	status string
```
Delta456, line 14: Better have `status` as an enum

vlib/veb/status/status.v
```diff
@@ -0,0 +1,72 @@
...
+
+// is_valid returns true if the status is valid
+// this is used to check if the request was valid
+pub fn (s Status) is_valid() bool {
+	if s.status.is_valid() {
+		return true
+	} else {
+		return false
+	}
+}
```
spytheman, line 72: Those methods have no utility at all, since they just delegate to methods of the underlying field, that could be just accessed directly, if it is public.

The code of the implementation also seems extremely verbose for what it does.
```v
	if s.status.is_valid() {
		return true
	} else {
		return false
	}
}
```
^ that could be just: `return s.status.is_valid()` ...

spytheman:
I do not see any abstraction value in such filler code; the users can just `import net.http`, and do whatever this is doing directly and more clearly.

Issue #19327, opened 2023-09-11
[ORM] RUNTIME ERROR: invalid memory access


Ddiidev:
### Describe the bug

ORM cannot handle nested foreign keys, example:

### Reproduction Steps

```v
pub struct Address {
	id int [primary; sql: serial; nonull]
	create_at string [default: 'now()'; nonull; sql_type: 'TIMESTAMP']
	update_at string [default: 'now()'; nonull; sql_type: 'TIMESTAMP']
	street string [nonull]
	city string [nonull]
	state string [nonull]
	zip_code string [nonull]
	proximity string [nonull]
}
```

```v
[table: 'RealStates']
pub struct RealState {
	id string [primary; sql_type: 'uuid'; default: "uuid_generate_v4()"; nonull]
	parent_id int
	name string [sql_type: 'varchar(80)'; nonull]
	cnpj string [sql_type: 'varchar(14)'; nonull]
	address []Address [fkey: 'id']
}
```

```v
[table: 'Realtors']
pub struct Realtor {
	id string [primary; sql_type: 'uuid'; default: "uuid_generate_v4()"; nonull]
	first_name string [sql_type: "VARCHAR(30)"; nonull]
	last_name string [sql_type: "VARCHAR(30)"; nonull]
	creci string [sql_type: "VARCHAR(8)"; nonull]
	cnpj string [sql_type: "VARCHAR(15)"]
	cpf string [sql_type: "VARCHAR(12)"]
	phone string [sql_type: "VARCHAR(15)"; nonull]
	real_state []RealState [fkey: 'id']
}
```

```v
mut db := pg.connect(host: pg_host, user: pg_user, password: pg_pass, dbname: pg_db)!
defer {
	db.close()
}

dump(sql db {
	select from Realtor
}!)
```

### Expected Behavior

something like...

```v
[./src/main.v:37] ast.SqlExpr: [Realtor{
	id: 'ffce0e60-084d-49c0-9e19-a272f257b1a4',
	first_name: 'asdsd',
	last_name: 'qweqwe',
	creci: '4312',
	cnpj: '',
	cpf: '',
	phone: '1241233',
	real_state: '1'
}]
```

### Current Behavior

```
V panic: as cast: cannot cast `unknown` to `v.ast.InfixExpr`
v hash: 12dd6e8
/tmp/v_1000/v2.2271908984048564447.tmp.c:20702: at _v_panic: Backtrace
[...]

PR #24080, opened 2025-03-29
orm: fix nested array

kbkpbot:
Fix issue #19327

TODO: Just fix compile error, but I think the data select from DB has some problem, maybe fix in other PR.


spytheman (approved):
Excellent work.

PR #24096, opened 2025-03-31
cgen: fix codegen for option unwrapped var passed to generic option type

### LINKED ISSUE #23972: C error when unwrapping option arrays
<details>
<summary>V version: V 0.4.9 90c638e, press to see full `v doctor` output</summary>

|V full version      |V 0.4.9 725e25991d7445f5609958816abc3370a5c0dac3.90c638e
|Processor           |12 cpus, 64bit, little endian, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz  
|Memory              |3.57GB/15.86GB
|                    |
|V executable        |***
|V last modified time|2025-03-18 16:32:21
|                    |
|V home dir          |OK, value: ***
|VMODULES            |OK, value: ***
|VTMP                |OK, value: ***
|Current working dir |OK, value: ***
|                    |
|Git version         |git version 2.38.1.windows.1
|V git status        |weekly.2024.24-1728-g90c638e6
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |N/A
|clang version       |N/A
|msvc version        |N/A
|emcc version        |N/A
|glibc version       |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  test.v && F:\***\test.exe`
```v
fn main() {
	mut arr := ?[]int(none);
	if arr == none {
		arr = []int{};
	} else {
		unwrap(arr) << 1;
	}
}

@[inline]
pub fn unwrap[T](t ?T) T {
	return t or { panic('unexpected `none`') }
}
```

**What did you see?**
```
cc: C:/Users/***/AppData/Local/Temp/v_0/test.01JPN6Q6G3A7QRMWMR8EF3VRJP.tmp.c:3859: warning: implicit declaration of function 'tcc_backtrace'
cc: C:/Users/***/AppData/Local/Temp/v_0/test.01JPN6Q6G3A7QRMWMR8EF3VRJP.tmp.c:5901: error: cannot convert 'struct array' to 'struct _option_Array_int'
...
cc: C:/Users/***/AppData/Local/Temp/v_0/test.01JPN6Q6G3A7QRMWMR8EF3VRJP.tmp.c:5901: error: cannot convert 'struct array' to 'struct _option_Array_int'
(note: the original output was 3 lines long; it was truncated to its first 2 lines + the last line)
builder error: 
==================
```

maybe works fine, or maybe an error message, idk

> [!NOTE]
>

felipensp:
Fix #23972

### DIFF
--- a/vlib/v/tests/options/option_unwrapped_arg_generic_opt_test.v
+++ b/vlib/v/tests/options/option_unwrapped_arg_generic_opt_test.v
@@ -3,7 +3,8 @@ fn test_main() {
 	if arr == none {
 		arr = []int{}
 	} else {
-		unwrap(arr)
+		ret := unwrap(arr)
+		assert ret == arr
 	}
 	assert arr?.len == 0
 }
--- a/vlib/v/tests/options/option_unwrapped_arg_generic_opt_test.v
+++ b/vlib/v/tests/options/option_unwrapped_arg_generic_opt_test.v
@@ -1,0 +1,16 @@
+fn test_main() {
+	mut arr := ?[]int(none)
+	if arr == none {
+		arr = []int{}
+	} else {
+		unwrap(arr) << 1
+	}
+	assert arr?.len == 0
+}
+
+fn t(a ?int) {}
+
+@[inline]
+pub fn unwrap[T](t ?T) T {
+	return t or { panic('unexpected `none`') }
+}
--- a/vlib/v/tests/options/option_unwrapped_arg_generic_opt_test.v
+++ b/vlib/v/tests/options/option_unwrapped_arg_generic_opt_test.v
@@ -3,7 +3,7 @@ fn test_main() {
 	if arr == none {
 		arr = []int{}
 	} else {
-		unwrap(arr) << 1
+		unwrap(arr)
 	}
 	assert arr?.len == 0
 }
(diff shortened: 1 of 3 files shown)


vlib/v/tests/options/option_unwrapped_arg_generic_opt_test.v
```diff
@@ -0,0 +1,16 @@
+fn test_main() {
+	mut arr := ?[]int(none)
+	if arr == none {
+		arr = []int{}
+	} else {
+		unwrap(arr) << 1
```
spytheman, line 6: I think that should not compile, but be a checker error, since `arr` in this place is already unwrapped, and has type of `[]int`, while later you have `pub fn unwrap[T](t ?T) T {`, i.e. the argument is an option.

spytheman: Where does it push the `1` to?
The output of unwrap() is an immutable array value.

spytheman: Yes, I also think that changing it to assign to a temporary in the test, will be better and will remain valid, even if later we add a checker error.

felipensp: Right.

spytheman (approved):
Excellent work.

PR #24106, opened 2025-04-01
encoding.binary: add encode_binary()/decode_binary() generic functions

kbkpbot:
This PR provide  Serialization/Deserialization functions for generic type `T`.
We can serialize generic data into a []u8 byte array and deserialize it back to its original form.


spytheman (approved):
Excellent work.

einar-hjortdal:
This is awesome!

PR #24107, opened 2025-04-01
new architecture: s390x

tankf33der:
New architecture: s390x.

97% tests passed in `vlib/v/tests`.
Additional efforts will be required to bring everything to proper functioning inside `vlib` dir too.

`Whats next:`
Inline assembly basically works and will require some more tests. It will be a different PR.


spytheman:
Excellent start on supporting s390x.
Thank you 🙇🏻‍♂️.

PR #24114, opened 2025-04-02
comptime: enable s390x + docs

tankf33der:
enable "$if s390x {...}" + mention in the doc.

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -6286,7 +6286,7 @@ Full list of builtin options:
 |--------------------------------|------------------|-------------------------------|-----------------------------------------------|
 | `windows`, `linux`, `macos`    | `gcc`, `tinyc`   | `amd64`, `arm64`, `aarch64`   | `debug`, `prod`, `test`                       |
 | `darwin`, `ios`, `bsd`         | `clang`, `mingw` | `i386`, `arm32`               | `js`, `glibc`, `prealloc`                     |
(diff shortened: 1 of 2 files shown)


felipensp:
Please add a test using such identifier in the code.

tankf33der:
> Please add a test using such identifier in the code.

I didn't find such code(s) in tests for other architectures. 
How do you see this implementation in the `comptime` directory?

My suggestion is to do it like this: new file `comptime/comptime_arch.v` with content conditionally like 
[this](http://pb1n.de/?14bdfa)

Delta456:
> Please add a test using such identifier in the code.

We currently don't have a way to test this on actions runner.

Delta456 (commented):
Is this how it would be implemented? How will the current changes make sure that it runs on `s390x` systems?

felipensp:
> Not sure this _can_ be tested. Unless the CI can test it against an S390 image.

I'm just talking about to test if 
```
$if s390x {
}
```
do not emit `error: undefined ident: s390x`

JalonSolov:
I don't think that's necessary.  The ident was added in #24107 and would've failed there if it wasn't properly defined.

spytheman:
Can it be merged now, after #24129 ?

tankf33der:
> Can it be merged now, after #24129 ?

test file `vlib/v/slow_tests/assembly/asm_test.s390x.v` merged as you suggested before, so now it's time for this merge too.

spytheman:
Excellent. Thank you 🙇🏻‍♂️ .

PR #24118, opened 2025-04-02
cgen: Don't emit mman.h and pthreads related code for freestanding bu…

raw-bin:
* This PR addresses an issue encountered with freestanding builds used for bare metal embedded targets that use anonymous functions.
* The issue is to do with the code generated for closure management - which are inherently associated with anonymous functions. The closure management code assumes that the run-time environment has support for mmap and posix locking primitives.
* Mmap and posix locking primitives are typically not available on bare metal embedded systems.
* This PR ensures that the generated code for closure managment doesn't contain mmap and posix locking for freestanding builds.
* Although this code solves the immediate problem a proper solution needs to parameterise things like page size assumptions etc.
* Unfortunately I could not figure out how to write a test for freestanding builds just yet.

### DIFF
--- a/vlib/v/gen/c/cheaders.v
+++ b/vlib/v/gen/c/cheaders.v
@@ -56,7 +56,7 @@ static inline void __sort_ptr(uintptr_t a[], bool b[], int l) {
 
 fn c_closure_helpers(pref_ &pref.Preferences) string {
 	mut builder := strings.new_builder(2048)
-	if pref_.os != .windows {
+	if pref_.os != .windows && pref_.is_bare == false {
 		builder.writeln('#include <sys/mman.h>')
 	}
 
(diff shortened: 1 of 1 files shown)


spytheman (approved):

spytheman:
> Unfortunately I could not figure out how to write a test for freestanding builds just yet.

You can put a pair of `.vv/.out` files in `vlib/v/gen/c/testdata/` , then run:
`./v vlib/v/gen/c/coutput_test.v` .

Note: the .vv file has to start with:
`// vtest vflags: -cc gcc -freestanding` .
The corresponding .out file should contain the output of the .vv program, which will be compiled with the flags in that comment.

See for example https://github.com/vlang/v/blob/master/vlib/v/gen/c/testdata/freestanding_module_import_2.vv .

PR #24123, opened 2025-04-03
checker: fix wrong type hint on predicate arg type mismatch

### LINKED ISSUE #24122: Incorrect type mismatch hint
### Describe the bug

When filtering an array of references, an incorrect type mismatch hint is shown by the complier

### Reproduction Steps

It can be reproduced by the following MWE:

```vlang
import net.http
import net.html

fn main() {
	response := http.get('https://webscraper.io/test-sites/e-commerce/allinone') or { panic(err) }
	html_doc := html.parse(response.body)

	all_products := html_doc.get_tags(name: 'div')
		.filter(fn (div html.Tag) bool {
			return (div.attributes['class'] or { '' }) == 'row'
		})

	println(all_products)
}
```

### Expected Behavior

The compiler should tell the user something like this:

```terminal
test.v:9:11: error: type mismatch, should use `fn(a &net.html.Tag) bool {...}`
    7 | 
    8 |     all_products := html_doc.get_tags(name: 'div')
    9 |         .filter(fn (div html.Tag) bool {
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~
   10 |             return (div.attributes['class'] or { '' }) == 'row'
   11 |         })
```

### Current Behavior

It tells the user this (notice the `&` char denoting the reference is missing):

```terminal
test.v:9:11: error: type mismatch, should use `fn(a net.html.Tag) bool {...}`
    7 | 
    8 |     all_products := html_doc.get_tags(name: 'div')
    9 |         .filter(fn (div html.Tag) bool {
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~
   10 |             return (div.attributes['class'] or { '' }) == 'row'
   11 |         })
```

### Possible Solution

I don't know :/

_No response_

### V version

0.4.8 (because of NixOS - but I could not find a similar bug report)

V full version: V 0.4.8 2ab1523
OS: linux, Linux version 6.14.0 (nixbld@localhost) (gcc (GCC) 13.3.0, GNU ld (GNU Binutils) 2.43.1) #1-NixOS SMP PREEMPT_DYNAMIC Mon Mar 24 14:02:41 UTC 2025
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz

getwd: /home/sceptri/Documents/Dev/Vlang/vkacr
vexe: /nix/store/h6l77qqj4vpv509yb1fmv6vbfzk5aav9-vlang-0.4.8/lib/v
vexe mtime: 1970-01-01 00:00:01

vroot: NOT writable, value: /nix/store/h6l77qqj4vpv509yb1fmv6vbfzk5aav9-vlang-0.4.8/lib
VMODULES: OK, value: /home/sceptri/.vmodules
VTMP: OK, value: /tmp/v_1000

Git version: git version 2.47.2
.git/config present: false

> [!NOTE]
>

felipensp:
Fix #24122

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -2957,7 +2957,6 @@ fn (mut c Checker) check_predicate_param(is_map bool, elem_typ ast.Type, node as
 		// Finish early so that it doesn't fail later
 		return
 	}
-	elem_sym := c.table.sym(elem_typ)
 	arg_expr := node.args[0].expr
 	match arg_expr {
 		ast.AnonFn {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #24126, opened 2025-04-03
db.mysql: use mysql datatype for alloc string_binds_map, not orm's

kbkpbot:
Fix the issue in `mysql_orm_test.v`.

for example:

```v
struct TestTimeType {
mut:
	id         int @[primary; sql: serial]
	username   string
	created_at time.Time @[sql_type: 'DATETIME']
	updated_at string    @[sql_type: 'DATETIME']
	deleted_at time.Time
}
```

The field `updated_at` has a orm type of `string`, but in mysql db, it is a `DATETIME`, and it should not use dynamically allocation.


spytheman (approved):
Excellent work.

PR #24129, opened 2025-04-03
add s390x assembly support + test

tankf33der:
My attempt to add s390x assembly.

Highlights:

- Intel syntax
- I tried creating asm in C code and gcc only accepted integer literals as naked 5
- The test file is copied from a neighboring file. Not bad for a start


JalonSolov:
Oops.  Need to run `v fmt -w` on your files.

spytheman (approved):
Excellent work.

Issue #24130, opened 2025-04-04
ORM & Mysql : time.Time return err data


Avey777:
### Describe the bug

Null will be overwritten by data with previous values

### Reproduction Steps

orm_mysql.v

```
module main

import db.mysql
import time

@[table: 'sys_users']
struct Users {
        id         string     @[immutable; primary; sql: 'id'; sql_type: 'VARCHAR(255)'; unique]
        created_at ?time.Time @[omitempty; sql_type: 'TIMESTAMP']
        updated_at ?time.Time @[omitempty; sql_type: 'TIMESTAMP']
}

fn main() {
        mut mysql_config := mysql.Config{
                host:     '127.0.0.1'
                port:     3306
                username: 'root'
                password: '12345678'
                dbname:   'vcore'
        }
        mut db := mysql.connect(mysql_config) or { panic(err) }
        defer {
                db.close()
        }

        sql db {
                drop table Users
                create table Users
        }!
        user1 := Users{
                id:         '123'
                created_at: time.now()
                // updated_at defaults to none
        }
        user2 := Users{
                id:         'abc'
                created_at: time.now()
                updated_at: time.now()
        }
        user3 := Users{
                id:         'abc'
                created_at: time.now()
                // updated_at: time.now()
        }


        sql db {
                insert user1 into Users
                insert user2 into Users
                insert user2 into Users
        }!

        data := sql db {
                select from Users
        }!
        dump(data)
}
```

### Expected Behavior

[schema.SysUser{
    id: '1'
    updated_at: 2025-06-04 11:16:07
    created_at: 2025-04-15 11:16:18
    deleted_at: Option(none)
}, schema.SysUser{
    id: '2'
    updated_at: 2025-06-05 11:16:07
    created_at: 2025-04-16 11:16:18
    deleted_at: Option(2025-04-05 07:57:12)
}, schema.SysUser{
    id: '3'
    updated_at: 2025-06-06 11:16:07
    created_at: 2025-04-17 11:16:18
    deleted_at: Option(none)
}]

### Current Behavior

[schema.SysUser{
    id: '1'
    updated_at: 2025-06-04 11:16:07
    created_at: Option(none)
}, schema.SysUser{
    id: '2'
    updated_at: 2025-06-05 11:16:07
    created_at: Option(2025-04-16 11:16:18)
}, schema.SysUser{
    id: '3'
    updated_at: 2025-06-06 11:16:07
    created_at:  Option(2025-04-05 07:57:12)
}]

### V version

[...]

PR #24131, opened 2025-04-04
db.mysql: add null result support

kbkpbot:
Fix issue #24130

This PR will check fetched field is `NULL` or not.
If the field is `NULL`, it will return `orm.Null{}` to `orm`, avoid use the old value stored in field.


spytheman (approved):
Excellent work.

PR #24137, opened 2025-04-04
add fork in windows.

CryingN:
I tried adding Windows. h to solve the problem of forking on Windows and getting a `pid=-1`. Currently, it can run normally on Windows 10.

### DIFF
--- a/vlib/v/tests/os_fork.v
+++ b/vlib/v/tests/os_fork.v
@@ -1,0 +1,9 @@
+import os
+
+$if windows {
+	#include <windows.h>
+}
+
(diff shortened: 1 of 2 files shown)


medvednikov:
/home/runner/work/v/v/vlib/v/tests/os_fork.v is not vfmt'ed

vlib/v/tests/os_fork.v
```diff
@@ -0,0 +1,9 @@
+import os
+
+$if windows {
+	#include <windows.h>
```
medvednikov, line 4: this is not needed in the test

PR #24141, opened 2025-04-05
checker: simplify some check types logic

felipensp:
Refactor + minor clean up/optimization

### DIFF
--- a/vlib/v/checker/check_types.v
+++ b/vlib/v/checker/check_types.v
@@ -125,12 +125,10 @@ fn (mut c Checker) check_types(got ast.Type, expected ast.Type) bool {
 	if exp_idx == got_idx {
 		return true
 	}
-	if exp_idx == ast.voidptr_type_idx || exp_idx == ast.nil_type_idx
-		|| exp_idx == ast.byteptr_type_idx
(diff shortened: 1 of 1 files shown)


vlib/v/checker/check_types.v
```diff
@@ -11,6 +11,11 @@ fn (mut c Checker) check_types(got ast.Type, expected ast.Type) bool {
 	if got == expected {
 		return true
 	}
+	exp_idx := expected.idx()
+	got_idx := got.idx()
+	if exp_idx == got_idx {
+		return true
+	}
```
spytheman, line 18: I am not sure if that is valid - one of the types can be `&&Type` or `!Type` for example , while the other just `Type`, and their type indexes will be the same (the index is just the 16 least significant bits of the type).

felipensp: This check already exists on the code, I just moved it there.

spytheman: yes, but I am not sure if that keeps all invariants the same ... I guess the CI will tell us 🤷🏻‍♂️

spytheman:
The CI `sanitize-memory-clang` job failure is unrelated, and fixed/skipped on master.

PR #24144, opened 2025-04-05
cgen: fix multi return with option type

### LINKED ISSUE #24142: s390x: corrupted Option() from fn
This code fails on s390x, because the structures are filled differently by `?int(10)` and the `?int` inside `(int, ?int)` pair.

```
fn fails(i int) !(int, ?int) {
        return error('fails')
}

fn main() {
        // works
        x := ?int(10)
        y := x? as int
        assert y == 10

        // fails
        a2, b2 := fails(2) or { 22, 22 }
        c2 := b2? as int
        assert a2 == c2
}
```

> [!NOTE]
>

felipensp:
Fix #24142


tankf33der:
all tests passed now in `vlib/v/tests/options`.

PR #24145, opened 2025-04-06
fix wide tests for big_endian

tankf33der:
My attempt fix tests for UTF16 in `vlib/builtin/` for big_endian system(s).

### DIFF
--- a/vlib/builtin/utf8_test.v
+++ b/vlib/builtin/utf8_test.v
@@ -55,19 +55,31 @@ fn test_to_wide_cyrillic() {
 	}
 }
 
-const wide_serial_number = [u8(67), 0, 76, 0, 52, 0, 54, 0, 73, 0, 49, 0, 65, 0, 48, 0, 48, 0,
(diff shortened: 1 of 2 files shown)


tankf33der:
I need a hint on how to make it work everywhere.

tankf33der:
> I need a hint on how to make it work everywhere.

update: second attempt testing now.

spytheman:
Is it ok to merge it now?

spytheman:
We should research a way to run at least some tests (including the ones in `vlib/builtin/`) on a big endian system on the CI 🤔 .

tankf33der:
> We should research a way to run at least some tests (including the ones in `vlib/builtin/`) on a big endian system on the CI 🤔 .

These patches will completely fix `vlib/builtin/` for s390x (latest issues) and that's why pulsar was integrated. 
And although this is not a real CI system, it's better than nothing.

tankf33der:
> Is it ok to merge it now?

If patches are ok (not ugly) for you than `yes`.
From now `vlib/builtin` will be working for all systems.

spytheman:
Thank you 🙇🏻‍♂️ .

Issue #24121, opened 2025-04-03
mysql compile file on RockyLinux 8


kbkpbot:
### Describe the bug

Try to compile mysql code on RockyLinux 8 will fail.

### Reproduction Steps

goto `vlib/db/mysql`

```sh
v mysql_test.v -showcc -keepc
```

### Expected Behavior

compile ok

### Current Behavior

```
# v mysql_test.v -showcc -keepc
> C compiler cmd: '/home/github/lang/v/thirdparty/tcc/tcc.exe' '@/tmp/v_0/mysql_test.tmp.c.rsp'
> C compiler response file "/tmp/v_0/mysql_test.tmp.c.rsp":
  -fwrapv -o "/home/github/lang/v/vlib/db/mysql/mysql_test" -D GC_THREADS=1 -D GC_BUILTIN_ATOMIC=1 -I "/home/github/lang/v/thirdparty/libgc/include" -I "/usr/include/mysql -m64" -I "/usr/include" -L "/usr/lib64/mysql" -L "/usr/lib64" "/tmp/v_0/mysql_test.tmp.c" -std=gnu99 -D_DEFAULT_SOURCE -bt25 "/home/github/lang/v/thirdparty/tcc/lib/libgc.a" -ldl -lpthread -lmysqlclient -lssl -lcrypto
> C compiler cmd: 'cc' '@/tmp/v_0/mysql_test.tmp.c.rsp'
> C compiler response file "/tmp/v_0/mysql_test.tmp.c.rsp":
  -fwrapv -o '/home/github/lang/v/vlib/db/mysql/mysql_test' -D GC_THREADS=1 -D GC_BUILTIN_ATOMIC=1 -I "/home/github/lang/v/thirdparty/libgc/include" -I "/usr/include/mysql -m64" -I "/usr/include" -L "/usr/lib64/mysql" -L "/usr/lib64" "/tmp/v_0/mysql_test.tmp.c" -std=gnu99 -D_DEFAULT_SOURCE "/home/github/lang/v/thirdparty/tcc/lib/libgc.a" -ldl -lpthread -lmysqlclient -lssl -lcrypto
builder error: Header file <mysql.h>, needed for module `db.mysql` was not found. Please install the libmysqlclient-dev development headers.

```

### Possible Solution

remove `-m64` from compile flags.

### V version

V 0.4.10 a298fb8aba367cf7e877ac9636d1635014dab56a.a298fb8

### Environment details (OS name and version, etc.)

[...]

PR #24146, opened 2025-04-06
ast: fix parse_cflag() support other flags between allowed_flags

kbkpbot:
Fix issue #24121
Parse cflags like '-I/usr/include/mysql -m64 -I/usr/include', if the first flag is in `allowed_flags`, it will treat all before next `allowed_flags` as a value, which is not correct. The old code will create `name=-I`, `value=/usr/include/mysql -m64`.


spytheman (approved):
Excellent work.

PR #24155, opened 2025-04-07
doc: improve Shared and Channels's topics, add more examples

enghitalo:


JalonSolov:
Need to block the s390 step(s) when it's only a doc change.

doc/docs.md
```diff
@@ -4495,52 +4495,79 @@ fn main() {
...
+
+	for {
+		if ch.closed {
+			break
+		}
+		value := <-ch or { // receive values from the channel
+			println('Channel closed')
+			break
+		}
+		println('Received: ${value}') // pop values from the channel
```
spytheman, line 4549: that does not pop anything, it just prints a value

spytheman: the popping/receiving is done by the `value := <- ch` line

doc/docs.md
```diff
@@ -4495,52 +4495,79 @@ fn main() {
...
+}
+
+fn clock(ch chan int) {
+	for i in 0 .. 5 {
+		time.sleep(1 * time.second)
+		println('Clock tick')
+		ch <- i // push a value into the channel
+	}
+	ch.close() // close the channel when done
 }
```
spytheman, line 4534: Why are both functions pushing to the same channel, and apparently very similar values?
What does it demonstrate?

doc/docs.md
```diff
@@ -4495,52 +4495,79 @@ fn main() {
...
-	// ...
+	ch := chan int{cap: 5}
+	spawn worker(ch)
+	spawn clock(ch)
+
+	for {
+		if ch.closed {
+			break
+		}
+		value := <-ch or { // receive values from the channel
```
spytheman, line 4545:
```suggestion
	for {
		value := <-ch or { // receive values from the channel
```
the break in the or block is enough

doc/docs.md
```diff
@@ -4635,23 +4662,26 @@ if select {
...
-res := ch.try_push(a) // try to perform `ch <- a`
-println(res)
-l := ch.len // number of elements in queue
-c := ch.cap // maximum queue length
-is_closed := ch.closed // bool flag - has `ch` been closed
-println(l)
-println(c)
-mut b := Abc{}
-ch2 := chan Abc{}
-res2 := ch2.try_pop(mut b) // try to perform `b = <-ch2`
```
spytheman, line 4654: These were more informative compared to a whole example for just `.try_pop()`, with a comment that duplicates the code ...
Remove the new example, and restore the old code.

spytheman:
> Need to block the s390 step(s) when it's only a doc change.

Done in https://github.com/vlang/v/pull/24160 .

doc/docs.md
enghitalo: Both have the same amount of information.🤔 It is the same information, however, without the use of variables.

doc/docs.md
enghitalo: True! I was using `{<-ch}` here, before. I will remove the comment. Thanks

doc/docs.md
enghitalo: That the variable is used by 2 different threads, Using the `close` method to close the Channel and terminate the application.

doc/docs.md
enghitalo: True, but I wanted to demonstrate the use of the `closed` attribute, set by the `close()` method. I'll try to find another way to show this, with two different examples.

doc/docs.md
spytheman: ok, can one of them use, say `ch <- 1000 + i`, so that the values can be distinct from the main thread, that observes them?

doc/docs.md
spytheman: No. The use of `.closed` should be avoided, not advocated for and put in a reference example. Besides, it is already documented later.

It is useful in some very isolated cases, not as a general mechanism.
Just consider that it made the code here, more verbose for no good reason.

doc/docs.md
spytheman: No, it is not the same information at all. The code sample before showed several channel features, while later in your example, shows just this, that is related to the channel use:
```v
res := ch.try_pop(mut b) // try to perform `b = <-ch`
```

Do delete the new code, and restore the old in this particular case, I do not want to argue about it.

spytheman: The other examples are mostly fine, but this one loses information, that is important for understanding how to use channels.

doc/docs.md
```diff
@@ -4495,52 +4495,79 @@ fn main() {
...
+	for i in 0 .. 5 {
+		ch <- i // push values into the channel
+	}
+}
+
+fn clock(ch chan int) {
+	for i in 0 .. 5 {
+		time.sleep(1 * time.second)
+		println('Clock tick')
+		ch <- i // push a value into the channel
```
spytheman, line 4531:
```suggestion
		ch <- i + 1000 // push a value into the channel
```

doc/docs.md
spytheman, line 4549:
```suggestion
		println('Received: ${value}')
```

doc/docs.md
```diff
@@ -4635,23 +4662,26 @@ if select {
...
-mut b := Abc{}
-ch2 := chan Abc{}
-res2 := ch2.try_pop(mut b) // try to perform `b = <-ch2`
+fn main() {
+	mut b := Abc{}
+	ch := chan Abc{}
+	res := ch.try_pop(mut b) // try to perform `b = <-ch`
+	println(res)
+	println(b)
+}
```
spytheman, line 4684:
```suggestion
a := 2.13
ch := chan f64{}
res := ch.try_push(a) // try to perform `ch <- a`
println(res)
l := ch.len // number of elements in queue
c := ch.cap // maximum queue length
is_closed := ch.closed // bool flag - has `ch` been closed
println(l)
println(c)
mut b := Abc{}
ch2 := chan Abc{}
res2 := ch2.try_pop(mut b) // try to perform `b = <-ch2`
```

Issue #22086, opened 2024-08-20
mysql bug on macosx


lorenko:
**V doctor:**
```
V full version: V 0.4.7 7baff15
OS: macos, macOS, 14.5, 23F79
Processor: 10 cpus, 64bit, little endian, Apple M1 Pro

getwd: /Users/lorenko/v/examples/database
vexe: /opt/homebrew/Cellar/vlang/0.4.7/libexec/v
vexe mtime: 2024-07-26 04:41:00

vroot: OK, value: /opt/homebrew/Cellar/vlang/0.4.7/libexec
VMODULES: OK, value: /Users/lorenko/.vmodules
VTMP: OK, value: /tmp/v_501

Git version: git version 2.39.2
Git vroot status: 4.3.17 (59578 commit(s) behind V master)
.git/config present: false

CC version: Apple clang version 15.0.0 (clang-1500.3.9.4)
thirdparty/tcc: N/A

```

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg mysql.v`
```v
import db.mysql

fn main() {
	mut conn := mysql.connect(
		host:     'localhost'
		port:     3306
		username: 'root'
		password: ''
		dbname:   'mysql'
	)!
	res := conn.query('show tables')!
	for row in res.rows() {
		println(row.vals.join(', '))
	}
	conn.close()
}

```

**What did you expect to see?**

connection

**What did you see instead?**
```
================== C compilation error (from cc): ==============
cc:                       ^
cc: /tmp/v_501/mysql.01J5RP27RW6P5HHEWPQCPD9YSY.tmp.c:19263:43: error: use of undeclared identifier 'REFRESH_HOSTS'
cc:         _const_db__mysql__refresh_hosts = ((u32)(REFRESH_HOSTS));
cc:                                                  ^
cc: /tmp/v_501/mysql.01J5RP27RW6P5HHEWPQCPD9YSY.tmp.c:19265:45: error: use of undeclared identifier 'REFRESH_THREADS'
cc:         _const_db__mysql__refresh_threads = ((u32)(REFRESH_THREADS));
cc:                                                    ^
cc: /tmp/v_501/mysql.01J5RP27RW6P5HHEWPQCPD9YSY.tmp.c:19266:43: error: use of undeclared identifier 'REFRESH_SLAVE'
cc:         _const_db__mysql__refresh_slave = ((u32)(REFRESH_SLAVE));
cc:                                                  ^
[...]

PR #24162, opened 2025-04-08
db.mysql: use hardcoded const declare

kbkpbot:
Fix issue #22086


JalonSolov:
Nice.  One less C header to deal with.  The file should also be renamed to just `consts.v` since there are no more `C.` references.

Final suggestion:  add end of line comments showing the C constant, for continuity.

spytheman (approved):
Excellent work.

PR #24166, opened 2025-04-09
sync: remove pthread_rwlockattr_setpshared

tankf33der:
During the porting of `vlang` to `NetBSD`, it was discovered that `pthread_rwlockattr_setpshared()` is hidden, and `PTHREAD_PROCESS_PRIVATE` is actually the default value in all systems.
Let's remove it completely and save `NetBSD`.

[FreeBSD 14](https://man.freebsd.org/cgi/man.cgi?query=pthread_rwlockattr_setpshared&apropos=0&sektion=3&manpath=FreeBSD+14.2-RELEASE&arch=default&format=ascii)
[MacOS 15](https://man.freebsd.org/cgi/man.cgi?query=pthread_rwlockattr_setpshared&apropos=0&sektion=3&manpath=macOS+15.0&arch=default&format=ascii)
[POSIX](https://linux.die.net/man/3/pthread_rwlockattr_setpshared)
[GNU Lib](https://www.gnu.org/software/gnulib/manual/html_node/pthread_005frwlockattr_005fsetpshared.html)

### DIFF
--- a/vlib/sync/sync_darwin.c.v
+++ b/vlib/sync/sync_darwin.c.v
@@ -17,7 +17,6 @@ fn C.pthread_mutex_unlock(voidptr) int
 fn C.pthread_mutex_destroy(voidptr) int
 fn C.pthread_rwlockattr_init(voidptr) int
 fn C.pthread_rwlockattr_setkind_np(voidptr, int) int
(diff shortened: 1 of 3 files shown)


spytheman:
Thank you for investigating this 🙇🏻‍♂️ .

spytheman (approved):

PR #24173, opened 2025-04-09
GNUmakefile: add initial NetBSD support

tankf33der:
This is next step for porting `vlang` to `NetBSD`.

Tested on `NetBSD 10` so far, my goal is working `vlang` on `NetBSD 8-10` out of box.


spytheman (approved):

spytheman:
You may have to do the same to the `Makefile` (it is used by `make` programs != GNU make)

PR #24176, opened 2025-04-09
pkgconfig, termios: NetBSD support

tankf33der:
Main step for NetBSD build:

- `-d dynamic_boehm` and static builds;
- tested on `NetBSD 8-10`;
- gate for `sparc64` support is open.


spytheman (approved):
Excellent work.

Issue #24190, opened 2025-04-12
[encoding.binary] error decoding map[K]Struct


kravlad:
### Describe the bug

Now ```V``` panics if I try to call ```binary.decode_binary``` with a map that has a struct as a value. (For example ```map[string]Struct```)

### Reproduction Steps

```
module main
import encoding.binary

struct St {
	i int
}

fn main() {
	b := binary.encode_binary({"s": St{1}})!
	b2 := binary.decode_binary[map[string]St](b)!
	println('Decoded binary file: ${b2}')
}
```

### Expected Behavior

```
"Decoded binary file: {'s': St{
    i: 1
}}"
```

### Current Behavior

```
/Users/user/dev/v/vlib/encoding/binary/serialize.v:402:8: error: cannot assign to `v`: expected `binary.Any`, not `St`
  400 |             v = decode_string(mut s)!
  401 |         } $else $if V is $struct {
  402 |             v = decode_struct(mut s, V{})!
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~
  403 |         } $else $if V is $map {
  404 |             v = decode_map(mut s, V{})!
/Users/user/dev/v/vlib/encoding/binary/serialize.v:408:17: error: cannot cast `encoding.binary.Any` to `St`
  406 |             v = decode_primitive(mut s, unsafe { V(0) })!
  407 |         }
  408 |         m[k as K] = v as V
      |                       ~~
  409 |     }
  410 |     return m
```

### V version

V 0.4.10 d54a823

### Environment details (OS name and version, etc.)

```
|V full version      |V 0.4.10 b114254.d54a823
|:-------------------|:-------------------
|OS                  |macos, macOS, 15.3.2, 24D81
|Processor           |8 cpus, 64bit, little endian, Apple M1
|Memory              |0.18GB/16GB
|                    |
|V executable        |/Users/user/dev/v/v
|V last modified time|2025-04-11 03:43:02
|                    |
|V home dir          |OK, value: /Users/user/dev/v
|VMODULES            |OK, value: /Users/user/.vmodules
|VTMP                |OK, value: /tmp/v_501
|Current working dir |OK, value: /Users/user/projects/proj
|                    |
|Git version         |git version 2.49.0
|V git status        |weekly.2024.41-1243-gd54a8232 (2 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |Apple clang version 16.0.0 (clang-1600.0.26.6)
[...]

PR #24192, opened 2025-04-12
encoding.binary: fix serialize map struct

kbkpbot:
Fix issue #24190

### DIFF
--- a/vlib/encoding/binary/serialize.v
+++ b/vlib/encoding/binary/serialize.v
@@ -311,7 +311,7 @@ fn decode_primitive[T](mut s DecodeState, value T) !T {
 		// TODO: `any` type support?
 		return error('${@FN}(): unsupported type ${typeof(value).name}')
 	}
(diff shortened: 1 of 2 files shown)


vlib/encoding/binary/serialize.v
```diff
@@ -309,15 +309,15 @@ fn decode_primitive[T](mut s DecodeState, value T) !T {
 		return T(s.get_u64()!)
 	} $else {
 		// TODO: `any` type support?
-		return error('${@FN}unsupported type ${typeof(value).name}')
+		return error('${@FN}(): unsupported type ${typeof(value).name}')
 	}
-	return error('impossiable error')
+	return error('${@FN}(): impossiable error')
```
spytheman, line 314:
```suggestion
	return error('${@FN}(): impossible error')
```

spytheman (approved):
Excellent work.

Issue #24178, opened 2025-04-10
ORM: Dynamic condition filterin  (feature)


Avey777:
### Describe the feature

ORM Support Dynamic condition filtering

V's ORM currently does not support dynamic conditional filtering, which has forced me to transfer some of my business to Go.
I hope this can be supported as soon as possible

### Use Case

`query := sql db {select from Table  where("id = ?", id) }`

Group Conditions in GORM provide a more readable and maintainable way to write complex SQL queries involving multiple conditions.

```
// Complex SQL query using Group Conditions
db.Where(
  db.Where("pizza = ?", "pepperoni").Where(db.Where("size = ?", "small").Or("size = ?", "medium")),
).Or(
  db.Where("pizza = ?", "hawaiian").Where("size = ?", "xlarge"),
).Find(&Pizza{})
// SQL: SELECT * FROM `pizzas` WHERE (pizza = "pepperoni" AND (size = "small" OR size = "medium")) OR (pizza = "hawaiian" AND size = "xlarge")
```

Gorm has a completely different implementation, and I hope V's ORM can provide similar functional support! Don't worry about Gorm's grammar, V should have its own grammar

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.10 2f7cf4d

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 86536e452e6248056666c750668334c7da87202e.2f7cf4d
|:-------------------|:-------------------
|OS                  |linux, Deepin 23
|Processor           |6 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
|Memory              |0.26GB/7.61GB
|                    |
|V executable        |/home/Jengro/.vmr/versions/v_versions/v_latest/v
|V last modified time|2025-04-09 07:02:59
|                    |
|V home dir          |OK, value: /home/Jengro/.vmr/versions/v_versions/v_latest
|VMODULES            |OK, value: /home/Jengro/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/Jengro/Documents/Dev/v-product/vprod-workspase/example
|                    |
|Git version         |git version 2.45.2
|V git status        |weekly.2025.14-28-g2f7cf4df
|.git/config present |true
|                    |
|cc version          |cc (Deepin 12.3.0-17deepin8) 12.3.0
[...]

PR #24196, opened 2025-04-12
orm: add function call api

kbkpbot:
Feature Request #24178


Avey777:
https://github.com/vlang/v/issues/24178
It solves this.

Avey777:
Looking forward to this merger.

Avey777:
@mindplay-dk 
You seem to have said that you also need dynamic filtering of orm. Check this out

vlib/orm/README.md
```diff
@@ -246,3 +248,106 @@ fn main() {
...
 ```
+
+## Function Call API
+You can utilize the `Function Call API` to work with `ORM`. It provides the
+capability to dynamically construct SQL statements. The Function Call API
+supports common operations such as `Create Table`/`Drop Table`/`Insert`/`Delete`/`Update`/`Select`,
+and offers convenient yet powerful features for constructing `WHERE` clauses,
+`SET` clauses, `SELECT` clauses, and more.
+
+A complete example is available in `vlib/orm/orm_func_test.v`.
```
Delta456, line 259: Better hyperlink the file

medvednikov:
@spytheman @felipensp what do you think?

felipensp:
> @spytheman @felipensp what do you think?

Nice work @kbkpbot! Looks a great addition. I just don't like mixing different statements via chaining calls.

Avey777:
@mindplay-dk 
You seem to have said that you also need dynamic filtering of orm. Check this out

I wonder which of these and the current orm syntax will be retained in the future? Give us directions for future use.

spytheman:
> The code call panic everywhere, but if not, the chain function call will be horrible.

Panics can not be intercepted by the callers, while returned errors can.

spytheman:
> I want to know. Which of the PR grammar and the existing ORM grammar will be retained in the end?Give us directions for future use.

The use cases are different - a static query, made with the ORM syntax, can be analyzed by the compiler, and you can get an error right away, if you mess up something.

A dynamic query can not be analyzed statically, and so you will get an error at runtime, but only if you execute that piece of code, written using the query builder.

If you look at it from that perspective, imho it is pretty clear what you should prefer.

Avey777:
> > I want to know. Which of the PR grammar and the existing ORM grammar will be retained in the end?Give us directions for future use.
> 
> The use cases are different - a static query, made with the ORM syntax, can be analyzed by the compiler, and you can get an error right away, if you mess up something.
> 
> A dynamic query can not be analyzed statically, and so you will get an error at runtime, but only if you execute that piece of code, written using the query builder.
> 
> If you look at it from that perspective, imho it is pretty clear what you shOkay, i probably get it! Both will exist at the same time! 
I think I know how to adjust my code after the pr merge.

spytheman:
> @spytheman @felipensp what do you think?

I think it is a very good compromise (except of using panics instead of returning errors, since while that may make the call sites a bit more verbose, because of the `!` signs, it also makes them free to decide for themselves how to handle the potential runtime errors, not just panic and exit the entire program).

It gives an additional freedom and convenience, to people that want to construct dynamic queries at runtime, without having to drop to the underlying driver specific APIs. At the same time, it does it without having to modify the language itself, using an API that is more or less convenient, and checked by the compiler as much as possible, which is also nice.

Avey777:
> > I want to know. Which of the PR grammar and the existing ORM grammar will be retained in the end?Give us directions for future use.
> 
> The use cases are different - a static query, made with the ORM syntax, can be analyzed by the compiler, and you can get an error right away, if you mess up something.
> 
> A dynamic query can not be analyzed statically, and so you will get an error at runtime, but only if you execute that piece of code, written using the query builder.
> 
> If you look at it from that perspective, imho it is pretty clear what you should prefer.



Ok, I think I know how to adjust my code after this pr merge

spytheman (approved):
Excellent work.

spytheman:
Oh, you already approved it ... Ok, time to merge then.

PR #24202, opened 2025-04-12
vlib: vanilla_http_server

enghitalo:
```v
module main

import enghitalo.vanilla.http_server
import enghitalo.vanilla.request_parser

fn handle_request(req_buffer []u8, client_conn_fd int) ![]u8 {
	req := request_parser.decode_http_request(req_buffer)!
	method := request_parser.slice_to_string(req.buffer, req.method)
	path := request_parser.slice_to_string(req.buffer, req.path)

	match method {
		'GET' {
			match path {
				'/' {
					return home_controller([])
				}
				'/users' {
					return get_users_controller([])
				}
				else {
					if path.starts_with('/user/') {
						id := path[6..]
						return get_user_controller([id])
					}
					return http_server.tiny_bad_request_response
				}
			}
		}
		'POST' {
			if path == '/user' {
				return create_user_controller([])
			}
			return http_server.tiny_bad_request_response
		}
		else {
			return http_server.tiny_bad_request_response
		}
	}

	return http_server.tiny_bad_request_response
}

fn main() {
	mut server := http_server.Server{
		port:            3000
		request_handler: handle_request
	}
	server.run()
}
```

### DIFF
--- a/vlib/vanilla_http_server/CONTRIBUTING.md
+++ b/vlib/vanilla_http_server/CONTRIBUTING.md
@@ -1,0 +1,32 @@
+# Contributing
+
+## Rules
+
+- Don't slow down performance
+- Always try to keep abstraction to a minimum
+- Don't complicate it
+
+## Benchmarking & Testing
+
+### CURL
+
+```sh
+curl -X GET --verbose http://localhost:3000/ &&
+curl -X POST --verbose http://localhost:3000/user &&
+curl -X GET --verbose http://localhost:3000/user/1
+
+```
+
+### WRK
+
+```sh
+wrk  -H 'Connection: "keep-alive"' --connection 512 --threads 16 --duration 10s http://localhost:3000
+```
+
+### Valgrind
+
+```sh
+# Race condition check
+v -prod -gc none .
+valgrind --tool=helgrind ./vanilla
+```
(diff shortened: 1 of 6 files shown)


Avey777:
This is not merged, is there any problem?

spytheman:
> This is not merged, is there any problem?

Initially it had failing CI jobs.

After reviewing it, I think that it should stay in its own repository - https://github.com/enghitalo/vanilla_http_server . The instructions in the PR are already written with that repository in mind.

medvednikov:
We've had some issues with picoev + veb lately. I think this is a great alternative in pure V + it supports multithreading. I think it should not only be included into the examples directory, but it should become the web server Veb uses.

enghitalo:
I think we can replace `client_conn_fd int` with `client_conn ClientConn` in case we make this cross-platform in the future

medvednikov:
vweb is dead, will be removed
veb with the current pico backend seems to be a dead end (single thread only), and the maintainer is not active

medvednikov:
It also shows great performance (top1)

image

Great work @enghitalo !

medvednikov:
Next step is to migrate veb to it.

koplenov:
Perfect

Jengro777:
> Next step is to migrate veb to it.

I hope the syntax of Veb itself can be kept to a minimum without unnecessary changes
@enghitalo

PR #24208, opened 2025-04-13
cgen: fix array fixed init on push

### LINKED ISSUE #24152: s390x: wrong initialization
Modified code from test `vlib/v/tests/arrays_closure_fixed_array_test.v`.

Variable `mike` have wrong values.
```
import arrays

fn main() {
        mut rr := [][2]int{}
        rr << [1, 2]!

        mm := [3, 4]
        arrays.index_of_first(rr, fn [mm] (idx int, mike [2]int) bool {
                println(idx)
                println(mike)
                println(mm)

                return true
        })
}
```
`x86_64`:
```
0
[1, 2]
[3, 4]
```

`s390x`:
```
0
[-339742632, 2417648]
[3, 4]
```

felipensp:
Fix #24152

### DIFF
--- a/vlib/v/gen/c/for.v
+++ b/vlib/v/gen/c/for.v
@@ -254,7 +254,7 @@ fn (mut g Gen) for_in_stmt(node_ ast.ForInStmt) {
 			} else if val_sym.kind == .array_fixed && !node.val_is_mut {
 				right := '((${styp}*)${cond_var}${op_field}data)[${i}]'
 				g.writeln('\t${styp} ${c_name(node.val_var)};')
-				g.writeln('\tmemcpy(*(${styp}*)${c_name(node.val_var)}, (byte*)${right}, sizeof(${styp}));')
+				g.writeln('\tmemcpy(&${c_name(node.val_var)}, ${right}, sizeof(${styp}));')
 			} else {
 				needs_memcpy := !node.val_type.is_ptr()
 					&& g.table.final_sym(node.val_type).kind == .array_fixed
--- a/vlib/v/gen/c/infix.v
+++ b/vlib/v/gen/c/infix.v
@@ -1047,13 +1047,19 @@ fn (mut g Gen) infix_expr_left_shift_op(node ast.InfixExpr) {
 			if needs_addr {
 				g.write(')')
 			}
+			mut must_close_braces := false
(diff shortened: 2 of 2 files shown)


tankf33der:
i see the `same` bug with a fix.
lets assume i'have applied patch correctly, sigh.


there is created file - https://envs.sh/KgN.c
i have commented local CLOSURE to be sure i do not corrupt locally.
`gcc KgN.c -lg` -> the same bug.
```
// V anon functions:
VV_LOCAL_SYMBOL bool anon_fn_766ad3a762b40f01_int_array_fixed_int_2__bool_142(int idx, Array_fixed_int_2 mike) {
	//struct _V_anon_fn_766ad3a762b40f01_int_array_fixed_int_2__bool_142_Ctx* _V_closure_ctx = __CLOSURE_GET_DATA();
	println(int_str(idx));
	println(Array_fixed_int_2_str(mike));
	//println(Array_int_str(_V_closure_ctx->mm));
	return true;
}
```
That's why it seems that `mike` is already corrupted when passed to this function, and this is done by line `KgN.c:7583`.

tankf33der:
By coincidence, I've been dealing with another similar problem on `s390x` where the second function argument gets corrupted, and I spent two days in `gdb` on `sparc64` inside the closure raw bytes and came to roughly the same conclusion. Suspicious.

felipensp:
> By coincidence, I've been dealing with another similar problem on `s390x` where the second function argument gets corrupted, and I spent two days in `gdb` on `sparc64` inside the closure raw bytes and came to roughly the same conclusion. Suspicious.

What is the V code to reproduce It?

tankf33der:
> > By coincidence, I've been dealing with another similar problem on `s390x` where the second function argument gets corrupted, and I spent two days in `gdb` on `sparc64` inside the closure raw bytes and came to roughly the same conclusion. Suspicious.
> 
> What is the V code to reproduce It?

from issue.

felipensp:
I pushed another try to mitigate the bug @tankf33der Can you check it?

tankf33der:
@felipensp - the same wrong output.

tankf33der:
This was a closure `thunk` bug, new draft version is testing and will be in the tree asap.

PR #24238, opened 2025-04-16
doc: remove only reference to `any`  from docs

JalonSolov:
Step 1 for removing `any` from V.


JalonSolov:
Will re-do with better commit message.

Issue #24240, opened 2025-04-16
vfmt import error


kbkpbot:
### Describe the bug

vfmt should check for proper `import` keyword and parse it correctly.

### Reproduction Steps

oneline.v
```v
import os import math
```

```
v fmt oneline.v
```

### Expected Behavior

```
import os
import math
```

### Current Behavior

```
oneline.v:1:11: error: cannot import multiple modules at a time
    1 | import os import math
      |           ~~~~~~

Internal vfmt error while formatting file: oneline.v.
```

### Possible Solution

parser should check for `import` keyword, and add a line break before it.

### Additional Information/Context

a special case :

suppose we has a module name `import`....
```v
import import import os
```

### V version

V 0.4.10 c2e27ef2b782d18655d573f24674f6876c4fc542.c2e27ef

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 c2e27ef2b782d18655d573f24674f6876c4fc542.c2e27ef
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.2 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |5.08GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-04-12 05:31:19
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/fmt
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.14-51-ga4b2de60 (2 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.4) 2.39

> [!NOTE]
[...]

PR #24241, opened 2025-04-16
parser: fix mutiple imports at one line

kbkpbot:
Fix issue #24240


JalonSolov:
I'm not sure this is the correct fix.

Because V treats whitespace as optional,
```v
import import import os
```
should be treated the same as
```v
import import
import os
```
The actual problem is that the `import` statement _could_ be followed by other things, as in
```v
import import as im
```
So what probably should be happening is that as soon as the 3rd token is not `as`, it should treat it as a completely separate statement, not as an error.

PR #24259, opened 2025-04-17
js: fix array type checking in sum type match expressions (fix #24237 )

### LINKED ISSUE #24237: JS Backend: sum type doesn't work well with array types
### Describe the bug

When compiling this code using `v -b js_browser .`:

```v
type SumType = string | []string

fn main() {
	x := SumType('aaa')
	match x {
		string { println('string') }
		[]string { println('[]string') }
	}
}
```

throws the following error:
```
Uncaught SyntaxError: Unexpected identifier 'string' (at main.js:7185:27)

/** @type {main__SumType} */
const x = new string("aaa");
if (x instanceof string) {
	builtin__println (new string("string"));
}
else if (x instanceof []string) { // <!-- HERE!
	builtin__println (new string("[]string"));
}
```

### Reproduction Steps

As above.

### Expected Behavior

The sum types should work well with array types.

### Current Behavior

It throws a JavaScript error.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.10 d559a62

|V full version      |V 0.4.10 30ececc.d559a62
|OS                  |macos, macOS, 15.3.2, 24D81
|Memory              |2.16GB/16GB
|                    |
|V executable        |/Users/gonzalo/v/v
|V last modified time|2025-04-13 10:51:40
|                    |
|V home dir          |OK, value: /Users/gonzalo/v
|Current working dir |OK, value: /Users/gonzalo/Projects/Personal/v/rx
|                    |
|V git status        |weekly.2025.14-39-gd559a62c (12 commit(s) behind V master)
|.git/config present |true
|                    |
|tcc git status      |thirdparty-macos-arm64 713692d4
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

gchumillas:
This commit fixes a bug in the JavaScript code generator where array types in sum type match expressions were not being properly checked. The issue was that the generated code was checking if array elements were instances of the array type itself, rather than the element type of the array.

Fixes #24237

### DIFF
--- a/vlib/v/gen/js/js.v
+++ b/vlib/v/gen/js/js.v
@@ -561,7 +561,7 @@ fn (mut g JsGen) js_name(name_ string) string {
 		name = name[3..]
 		return name
 	}
-	name = name_.replace('.', '__')
+	name = name_.replace('[]', '').replace('.', '__')
 	if name in js_reserved {
 		return '_v_${name}'
 	}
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #24268, opened 2025-04-19
log: add local time / utc time selection support

kbkpbot:
This PR enable `log` to set log time in local time or utc time(default).

ll.v

```v
module main

import log

fn main() {
        mut l := log.Log{}

        l.set_time_format(.tf_default)
        l.info('utc time')

        l.set_local_time(true)
        l.info('local time')
}
```

run :

```
$ ./ll
2025-04-19 06:47 [INFO ] utc time
2025-04-19 14:47 [INFO ] local time
```

### DIFF
--- a/vlib/log/log.v
+++ b/vlib/log/log.v
@@ -38,6 +38,7 @@ mut:
 	time_format        TimeFormat = .tf_rfc3339_micro
 	custom_time_format string     = 'MMMM Do YY N kk:mm:ss A' // timestamp with custom format
 	short_tag          bool
+	local_time         bool // use local time or utc time in log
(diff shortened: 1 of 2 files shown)


vlib/log/log.v
```diff
@@ -313,6 +314,16 @@ pub fn (l Log) get_short_tag() bool {
 	return l.short_tag
 }
 
+// set_local_time will set the log using local time
```
medvednikov, line 317: getters and setters is not really the V way... but I see the entire module is designed this way. Will have to be fixed.

medvednikov:
Useful feature, thanks.

Issue #24272, opened 2025-04-20
`undefined symbol '__fpclassify'` on Windows


kbkpbot:
### Describe the bug

when compile some code, it will result in `undefined symbol '__fpclassify'`

### Reproduction Steps

On windows, run :
```
v vlib\crypto\blake3\blake3_test.v -showcc -show-c-output
```

### Expected Behavior

Compile and test ok

### Current Behavior

```
>v vlib\crypto\blake3\blake3_test.v -showcc -show-c-output
> C compiler cmd: "D:\v\v\v\thirdparty\tcc\tcc.exe" "@C:\Users\DDT\AppData\Local\Temp\v_0\blake3_test.01JS8K2CAGHNZNWSV6FHD48DTG.tmp.c.rsp"
> C compiler response file "C:\Users\DDT\AppData\Local\Temp\v_0\blake3_test.01JS8K2CAGHNZNWSV6FHD48DTG.tmp.c.rsp":
  -fwrapv "C:\\Users\\DDT\\.vmodules\\.cache\\35\\35c0869978bbbb91ceb3a74f2c2e2185.module.json.o" -o "D:\\v\\v\\v\\vlib\\crypto\\blake3\\blake3_test.exe" -Wl,-stack=16777216 -Werror=implicit-function-declaration -D GC_THREADS=1 -D GC_NOT_DLL=1 -D GC_WIN32_THREADS=1 -D GC_BUILTIN_ATOMIC=1 -I "D:\\v\\v\\v\\thirdparty\\libgc\\include" -I "D:\\v\\v\\v\\thirdparty\\cJSON" "C:\\Users\\DDT\\AppData\\Local\\Temp\\v_0\\blake3_test.01JS8K2CAGHNZNWSV6FHD48DTG.tmp.c" -std=c99 -D_DEFAULT_SOURCE -municode -bt25 -ldbghelp -luser32 "D:\\v\\v\\v/thirdparty/tcc/lib/libgc.a" -lws2_32 -ladvapi32
======== Output of the C Compiler (D:\v\v\v\thirdparty\tcc\tcc.exe) ========
tcc: error: undefined symbol '__fpclassify'
============================================================================
> C compiler cmd: gcc "@C:\Users\DDT\AppData\Local\Temp\v_0\blake3_test.01JS8K2CAGHNZNWSV6FHD48DTG.tmp.c.rsp"
[...]

PR #24273, opened 2025-04-20
json: link with libm

kbkpbot:
Fix issue #24272

It also need to fix `tccbin` `math.h` to fix this bug.
https://github.com/vlang/tccbin/issues/59


spytheman (approved):
Excellent work.

PR #24277, opened 2025-04-20
checker: add deprecate notice for `any` arg, prevent `any` from being used as map key,value or array type (currently crashing)

### LINKED ISSUE #24224: `any` cannot be used as value of `map`
### Describe the bug

`any` cannot be used as value of `map`, the compilation passed with no error but generated C code has undefined type `any`.

### Reproduction Steps

Run this code:
```v
fn main() {
    mut a := map[string]any{}
    a['key'] = 'value'
    a['urmom'] = 69
}
```

### Expected Behavior

No error at all **or** give correct diagnostics by V compiler

### Current Behavior

C compiler complains:
```
cc: /tmp/v_501/wtf.01JRW8VYQ36SH733MMCBQQXAXP.tmp.c:6336:52: error: use of undeclared identifier 'any'
cc:  6336 |         Map_string_any a = new_map(sizeof(string), sizeof(any), &map_hash_string, &map_eq_string, &map_clone_string, &map_free_string)
cc:       |                                                           ^
cc: /tmp/v_501/wtf.01JRW8VYQ36SH733MMCBQQXAXP.tmp.c:6338:47: error: expected expression
cc:  6338 |         map_set(&a, &(string[]){_SLIT("key")}, &(any[]) { _SLIT("value") });
cc:       |                                                      ^
cc: /tmp/v_501/wtf.01JRW8VYQ36SH733MMCBQQXAXP.tmp.c:6338:43: error: use of undeclared identifier 'any'
cc:  6338 |         map_set(&a, &(string[]){_SLIT("key")}, &(any[]) { _SLIT("value") });
cc:       |                                                  ^
cc: /tmp/v_501/wtf.01JRW8VYQ36SH733MMCBQQXAXP.tmp.c:6339:49: error: expected expression
cc:  6339 |         map_set(&a, &(string[]){_SLIT("urmom")}, &(any[]) { 69 });
cc:       |                                                        ^
...
cc: 5 errors generated.
(note: the original output was 16 lines long; it was truncated to its first 12 lines + the last line)
builder error: 
==================


```

### Possible Solution

Don't allow `any` as `map` value at all (which is a bad solution but still possible) and give error at **V compiler's level`, or fix the codegen.

_No response_

### V version

V 0.4.10 b76997a

Image
V doctor stuck somehow. Use `fastfetch` screenshot instead for a moment
> [!NOTE]
>

felipensp:
Fix #24224

### DIFF
--- a/vlib/v/checker/containers.v
+++ b/vlib/v/checker/containers.v
@@ -18,7 +18,7 @@ fn (mut c Checker) array_init(mut node ast.ArrayInit) ast.Type {
 		}
 		if node.elem_type != 0 {
 			elem_sym := c.table.sym(node.elem_type)
-
+			c.check_any_type(node.elem_type, elem_sym, node.pos)
 			if node.typ.has_flag(.option) && (node.has_cap || node.has_len) {
 				c.error('Option array `${elem_sym.name}` cannot have initializers', node.pos)
 			}
@@ -507,6 +507,7 @@ fn (mut c Checker) map_init(mut node ast.MapInit) ast.Type {
 				c.error('cannot use Result type as map value type', node.pos)
 			}
 			val_sym := c.table.sym(info.value_type)
+			c.check_any_type(info.value_type, val_sym, node.pos)
 			if val_sym.kind == .struct {
 				val_info := val_sym.info as ast.Struct
 				if val_info.generic_types.len > 0 && val_info.concrete_types.len == 0
@@ -531,6 +532,8 @@ fn (mut c Checker) map_init(mut node ast.MapInit) ast.Type {
 		c.ensure_type_exists(info.value_type, node.pos)
 		node.key_type = info.key_type
 		node.value_type = info.value_type
+		key_sym := c.table.sym(node.key_type)
+		c.check_any_type(info.value_type, key_sym, node.pos)
 		return node.typ
 	}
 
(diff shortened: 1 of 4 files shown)


vlib/v/checker/containers.v
```diff
@@ -18,7 +18,7 @@ fn (mut c Checker) array_init(mut node ast.ArrayInit) ast.Type {
 		}
 		if node.elem_type != 0 {
 			elem_sym := c.table.sym(node.elem_type)
```
beakash, line 20: add arraay

vlib/v/checker/tests/deprecate_any.out
```diff
@@ -0,0 +1,25 @@
...
+vlib/v/checker/tests/deprecate_any.vv:4:7: error: cannot use type `any` here
+    2 | 
+    3 | fn main() {
+    4 |     _ := []any{}
+      |          ~~~~~~
+    5 |     _ := map[int]any{}
+    6 |     _ := map[any]int{}
+vlib/v/checker/tests/deprecate_any.vv:5:7: error: cannot use type `any` here
+    3 | fn main() {
+    4 |     _ := []any{}
```
beakash, line 28: remove `any```

JalonSolov: The message is correct as-is, and has been in V for a while.  It shouldn't be changed.

Also, this is in a `.out` file, which is exactly what the compiler outputs.  It is used to make sure the compiler output hasn't changed.

PR #24278, opened 2025-04-20
checker: add check for recursive array init on struct

### LINKED ISSUE #21195: Structure field default value crash, on initializer with length 0
### Describe the bug

# Example

Code: https://play.vlang.io/p/f638c371d1

```v
// example bug

// buggy struct 
struct Node {
	next []Node = []Node {len: 0}
}

struct BetterNode {
	next [] BetterNode
}

fn main() {
	// tested with lastest V playground 
	
	// if we create variable of node
	buggy := Node{} // we get segmentation fault 
	// this will work just fine 
	normal := BetterNode{} 
	
	
	bug := []int {len: 0} // works just fine
}
```

### Reproduction Steps

Recently I started working on the first game and it's engine, when doing so I found an edge case that causes program to segfault with no compile time errors.

I won't put original code snippet since it's irrelevant and issue is really easy to reproduce.

When a structure with field that has this self referring structure array with initializer of length 0, structure creation causes crash with no error.

### Expected Behavior

I'm aware of that the following code is incorrect (at least in my opinion) which means compiler should throw an error in such a case:
```
variable []Type = []Type {len: 0}
                          Error.        ^^^^^^
Initializer of length 0 in struct field default value is forbidden
```
Or compiler has to warn you about such a possible error case scenario 

### Current Behavior

Output:

```
code.v:16:1: warning: unused variable: `buggy`
   14 |     
   15 |     // if we create variable of node
   16 |     buggy := Node{} // we get segmentation fault 
      | ~~~~~
   17 |     // this will work just fine 
   18 |     normal := BetterNode{}
code.v:18:1: warning: unused variable: `normal`
   16 |     buggy := Node{} // we get segmentation fault 
   17 |     // this will work just fine 
   18 |     normal := BetterNode{} 
      | ~~~~~~
   19 |     
   20 |
code.v:21:2: warning: unused variable: `bug`
   19 |     
   20 |     
   21 |     bug := []int {len: 0} // works just fine
      |     ~~~
   22 | }
Caught fatal signal 11
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.5 31fdf58

```
V full version: V 0.4.5 31fdf58

getwd: /home/admin/playground
vexe: /home/admin/v/v
vexe mtime: 2024-04-06 07:23:24

VTMP: OK, value: /tmp/v_0

.git/config present: true

```
> [!NOTE]
>

felipensp:
Fix #21195

### DIFF
--- a/vlib/v/checker/struct.v
+++ b/vlib/v/checker/struct.v
@@ -96,7 +96,7 @@ fn (mut c Checker) struct_decl(mut node ast.StructDecl) {
 							symfield.is_recursive = true
 						} else {
 							c.error('recursive struct is only possible with optional pointer (e.g. ?&${c.table.type_to_str(field.typ.clear_flag(.option))})',
-								node.pos)
+								field.pos)
 						}
 					}
 				}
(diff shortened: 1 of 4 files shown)


spytheman (approved):
Excellent work.

Issue #24284, opened 2025-04-21
Capitals in the struct field names with type aliasing


maiolika:
### Describe the bug

Compiler enforces you to use snake_case in struct field names.
But when you make one with `type` you can use capitals.

### Reproduction Steps

```
struct Struct1 {
 	field_One int _// no, you can't_
 	field_two int
}
// error: field name `fieldOne` cannot contain uppercase letters, use snake_case instead

type Struct2 = struct {
	field_One int // yes, you can
	field_two int
}
```

### Expected Behavior

For Struct2 // `error: field name `fieldOne` cannot contain uppercase letters, use snake_case instead`

### Current Behavior

You get

```
Struct2(struct {
    field_One: 1
    field_two: 0
})
```

### V version

Current V version: V 0.4.10 188d68d, timestamp: 2025-04-20 03:25:56 +0300

### Environment details (OS name and version, etc.)

Windows 10 Home 22H2 19045.5737

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

JalonSolov:
Definitely a bug.  It should now allow different rules for anonymous structs than for regular ones.

JalonSolov:
So it is definitely tied to creating an alias for an anonymous struct... which seems weird, because you've now given a name to something anonymous...

spytheman:
I think we should just ban `type Struct = struct {` .

I do not see any utility in it, vs `struct Struct {`, and the second is clearer and better checked.

spytheman:
@Delta456 - currently it works for me too, when initializing an instance of an anonymous struct:
![Image](https://github.com/user-attachments/assets/f30ac062-e308-4d54-9881-d7efd7fe01b2)

PR #24286, opened 2025-04-21
 checker: check anon struct field valid case

Delta456:
Fixes https://github.com/vlang/v/issues/24284


spytheman (approved):
Excellent work.

PR #24287, opened 2025-04-21
v: add initial support for ppc64le

tankf33der:
Initial `ppc64le` support. Works in `qemu`, needs hardware testing.

Highlights:

- no stack frame(s)
- thunk uses 1 floating point register
- all `vlib/v/tests` and all `*closure*_test.v` passed
- full bootstrap tested on `Debian` (POWER10) and `Alma` (POWER9)
- `wiki` page will be later
- assembler support will be later


spytheman (approved):
Excellent work.

PR #24289, opened 2025-04-21
cgen: fix codegen for indexexpr on for loop with branchstmt

### LINKED ISSUE #22760: the error `expression evaluated but not used` is not reported for `arr[i] or { break }` inside a for loop
**V doctor:**

```
V full version: V 0.4.8 5c65e58.3108820

vexe mtime: 2024-11-04 16:24:59

Git vroot status: weekly.2024.45
.git/config present: true

```

**What did you do?**

```v
fn main() {
	arr := ["hello", "world"]
	for {
		arr[69] or { break }
	}
	println("0_o")
}
```

`error: expression evaluated but not used`

**What did you see instead?**

```
cc: C:/Users/user/AppData/Local/Temp/v_0/test.01JBW2ABZYYKVJ60MYXSQFBG1J.tmp.c:7156: warning: implicit declaration of function 'tcc_backtrace'
cc: C:/Users/user/AppData/Local/Temp/v_0/test.01JBW2ABZYYKVJ60MYXSQFBG1J.tmp.c:13762: error: ';' expected (got "}")
... (the original output was 3 lines long, and was truncated to 2 lines)
builder error: 
==================
```

**Additional information**
If another expression or statement is added after the faulty `arr[69] or { break }` line, the compiler correctly recognizes that the previous expression is unused, however doesn't report the newly inserted one:

```v
fn main() {
	arr := ["hello", "world"]
	for {
		arr[69] or { break }
		1
	}
	println("0_o")
}
```

errors out with

```
test.v:4:11: error: expression evaluated but not used
    2 |     arr := ["hello", "world"]
    3 |     for {
    4 |         arr[69] or { break }
      |                 ~~~~~~~~~~~~
    5 |         1
    6 |     }
```

If expressions like `arr[69] or { break }` are not supposed to be reported as unused, this should probably be considered a cgen error, as the generated C code does not compile.

This will print `0_o`, but should be an error:

```rs
fn main() {
    arr := ["hello", "world"]
    for {
        arr[69] or { break }.len
    }
    println("0_o")
}
```

> [!NOTE]
>

felipensp:
Fix #22760

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -2200,8 +2200,9 @@ fn (mut g Gen) stmts_with_tmp_var(stmts []ast.Stmt, tmp_var string) bool {
 				}
 			}
 			g.stmt(stmt)
-			if (g.inside_if_option || g.inside_if_result || g.inside_match_option
-				|| g.inside_match_result) && stmt is ast.ExprStmt {
+			if stmt is ast.ExprStmt && (g.inside_if_option || g.inside_if_result
+				|| g.inside_match_option || g.inside_match_result
+				|| (stmt.expr is ast.IndexExpr && stmt.expr.or_expr.kind != .absent)) {
 				g.writeln(';')
 			}
 		}
--- a/vlib/v/tests/index_on_loop_test.v
+++ b/vlib/v/tests/index_on_loop_test.v
@@ -1,0 +1,7 @@
+fn test_main() {
+	arr := ['hello', 'world']
+	for {
+		arr[69] or { break }
+	}
+	assert true
+}


vlib/v/gen/c/cgen.v
```diff
@@ -2200,8 +2200,9 @@ fn (mut g Gen) stmts_with_tmp_var(stmts []ast.Stmt, tmp_var string) bool {
 				}
 			}
 			g.stmt(stmt)
-			if (g.inside_if_option || g.inside_if_result || g.inside_match_option
-				|| g.inside_match_result) && stmt is ast.ExprStmt {
+			if stmt is ast.ExprStmt && (g.inside_if_option || g.inside_if_result
+				|| g.inside_match_option || g.inside_match_result
+				|| (stmt.expr is ast.IndexExpr && stmt.expr.or_expr.kind != .absent)) {
 				g.writeln(';')
 			}
```
spytheman, line 2207: that seems to be getting very complex; I wonder if it would not be simpler, to just always append the ;

spytheman (approved):
Good work.

Issue #24280, opened 2025-04-21
compiler bug, fixed array


sibkod:
1st case

```
module main

pub fn a(x1 [32]u8, y1 [32]u8, z1 [32]u8, x2 [32]u8, y2 [32]u8, z2 [32]u8) ([32]u8, [32]u8, [32]u8) {
	return [32]u8{}, [32]u8{}, [32]u8{}
}

pub fn b(x [32]u8, y [32]u8, z [32]u8) ([32]u8, [32]u8) {
	return [32]u8{}, [32]u8{}
}

fn main() {
	x1 := [32]u8{}
	y1 := [32]u8{}
	z1 := [32]u8{}
	x2 := [32]u8{}
	y2 := [32]u8{}
	z2 := [32]u8{}

	x3, y3, z3 := a(x1, y1, z1, x2, y2, z2)

	xx, yy := b(x3, y3, z3)

	println('Done: $xx $yy')
}
```

cannot convert 'struct _v_Array_fixed_u8_32' to 'unsigned char *'

2nd case

```
type Bytes32 = [32]u8

pub fn a(x1 Bytes32, y1 Bytes32, z1 Bytes32, x2 Bytes32, y2 Bytes32, z2 Bytes32) (Bytes32, Bytes32, Bytes32) {
	return Bytes32{}, Bytes32{}, Bytes32{}
}

pub fn b(x Bytes32, y Bytes32, z Bytes32) (Bytes32, Bytes32) {
	return Bytes32{}, Bytes32{}
}

fn main() {

	x1 := Bytes32{}
	y1 := Bytes32{}
	z1 := Bytes32{}
	x2 := Bytes32{}
	y2 := Bytes32{}
	z2 := Bytes32{}

	x3, y3, z3 := a(x1, y1, z1, x2, y2, z2)

	xx, yy := b(x3, y3, z3)

	println('Done: $xx $yy')
}
```

error: expression expected before '{'

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

koplenov:
@felipensp second case doesn't work https://play.vlang.io/p/11fc06a610

felipensp:
> [@felipensp](https://github.com/felipensp) second case doesn't work https://play.vlang.io/p/11fc06a610

This requires a different fix. Can you file a bug about?

PR #24295, opened 2025-04-22
cgen: fix codegen for multi return with aliased fixed array

felipensp:
Fix second bug case posted at #24280


spytheman (approved):
Excellent work.

PR #24299, opened 2025-04-22
cgen,ast: add ppc64le assembly support + test

tankf33der:
Lets add `asm` support.

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -1834,6 +1834,12 @@ pub const s390x_with_number_register_list = {
 	'v#': 32
 }
 
+pub const ppc64le_no_number_register_list = []string{}
+pub const ppc64le_with_number_register_list = {
+	'f#': 32
+	'r#': 32
+}
+
 pub struct DebuggerStmt {
 pub:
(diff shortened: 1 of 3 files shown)


vlib/v/ast/ast.v
```diff
@@ -1834,6 +1834,12 @@ pub const s390x_with_number_register_list = {
 	'v#': 32
 }
 
+pub const ppc64le_no_number_register_list = []string{}
```
spytheman, line 1837: Why add an empty array as a constant?

spytheman: Imho just use `[]` in the only `gen_all_registers` call that uses it.

spytheman (approved):
Good work.

spytheman:
The CI failure is unrelated, and already fixed on the master branch.

vlib/v/ast/ast.v
tankf33der: > Why add an empty array as a constant?

I really wanted to maintain continuity and make it exactly the same as in other architectures even if there are no special registers.

PR #24301, opened 2025-04-23
os: fix windows filelock

kbkpbot:
Use `cfile` instead of `fd` for windows.
Avoid compiler warnings:

```
Warning: C:\Users\runneradmin\AppData\Local\Temp\v_0\v_prod_cg.01JSFX343YW33PCJ7ETMJQF2CH.tmp.c:27250:16: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
   CloseHandle(((voidptr)(l->fd)));
                ^
C:\Users\runneradmin\AppData\Local\Temp\v_0\v_prod_cg.01JSFX343YW33PCJ7ETMJQF2CH.tmp.c: In function 'os__filelock__FileLock_try_acquire':
Warning: C:\Users\runneradmin\AppData\Local\Temp\v_0\v_prod_cg.01JSFX343YW33PCJ7ETMJQF2CH.tmp.c:27266:9: warning: comparison between pointer and integer
  if (fd == -1) {
         ^~
Warning: C:\Users\runneradmin\AppData\Local\Temp\v_0\v_prod_cg.01JSFX343YW33PCJ7ETMJQF2CH.tmp.c:27269:11: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
  l->fd = ((int)(fd));
           ^
```

### DIFF
--- a/vlib/os/filelock/lib_windows.c.v
+++ b/vlib/os/filelock/lib_windows.c.v
@@ -4,6 +4,7 @@ fn C.DeleteFileW(&u16) bool
 fn C.CreateFileW(&u16, u32, u32, voidptr, u32, u32, voidptr) voidptr
 fn C.CloseHandle(voidptr) bool
 
+// unlink unlink the lock file and release it.
 pub fn (mut l FileLock) unlink() {
 	if !isnil(l.cfile) {
 		C.CloseHandle(l.cfile)
@@ -13,6 +14,7 @@ pub fn (mut l FileLock) unlink() {
 	C.DeleteFileW(t_wide)
 }
 
+// acquire acquire the lock file.
 pub fn (mut l FileLock) acquire() ! {
 	if !isnil(l.cfile) {
 		return error_with_code('lock already acquired by this instance', 1)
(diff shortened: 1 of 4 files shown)


vlib/os/filelock/lib_windows.c.v
```diff
@@ -6,11 +6,11 @@ fn C.CloseHandle(voidptr) bool
 
 pub fn (mut l FileLock) unlink() {
 	if !isnil(l.cfile) {
-		ret := C.CloseHandle(l.cfile)
+		C.CloseHandle(l.cfile)
```
JalonSolov, line 10: Technically, it would be better to have
```suggestion
		_ = C.CloseHandle(l.cfile)
```
to explicitly show that the return code is being ignored, rather than not knowing there was a return code at all.

vlib/os/filelock/lib_windows.c.v
```diff
@@ -6,11 +6,11 @@ fn C.CloseHandle(voidptr) bool
...
 		l.cfile = unsafe { nil }
 	}
 	t_wide := l.name.to_wide()
-	ret := C.DeleteFileW(t_wide)
+	C.DeleteFileW(t_wide)
```
JalonSolov, line 14: Same as above
```suggestion
	_ = C.DeleteFileW(t_wide)
```

spytheman:
I've made another fix in 87d3a1d , using i64 for the fd field.
However it seems that it is only fixing the symptom of the CI failure, not the actual cause.
I suspect that github updated the gcc version on their runners, and there may be a timing issue with the new one, because I could not reproduce the failure both before and after the change in 87d3a1d.

PR #24308, opened 2025-04-23
rv64 inline assembly support

raw-bin:
Please consider this PR that:

1. Adds support for parsing rv64 inline assembly

2. Adds a local test runner specifiable via the following comments in the test file (for example):
// vtest vflags: -cc riscv64-linux-gnu-gcc -gc none
// vtest runner: qemu-riscv64
// vtest runner_flags:  -L /usr/riscv64-linux-gnu

   The runner invocation in the above example assumes that the [qemu-user](https://packages.ubuntu.com/search?suite=noble&arch=any&mode=exactfilename&searchon=contents&keywords=qemu-user) package is installed on the host system.

3. Adds support for parsing rv64 inline assembly using the following addressing modes:

   - displacement from base register (eg: sd a0, 0(a1))
   - composite displacement from base register (eg: sd a0, (8 * 1)(a1))

4. Adds a test for those new addressing modes

Note that I have run the following successfully:

   - `make check`
   - `v vlib/v/gen/c/coutput_test.v`


medvednikov:
FAIL  [ 530/2359] C:  1298.4 ms, R: 58504.996 ms vlib/v/gen/c/coutput_test.v

raw-bin:
Closing PR to address issues raised, in a new PR.

PR #24310, opened 2025-04-24
ci: revert commit a0a8324, test ci windows

kbkpbot:
Just a try for fixing CI windows.


spytheman:
The fix is in https://github.com/vlang/v/pull/24304 .

Issue #24311, opened 2025-04-24
Support uuid v7


Avey777:
### Describe the feature

Efficient cursor-based pagination relies on lexicographical and chronological sorting support, which is not provided by older UUID versions.

Suggested solution
UUID v7 fills a big gap in the database space by addressing the need for the lexicographically and chronologically sortable identifiers. CUID and ULID are strong contenders, but they come with a significant adoption friction involved, due to not being UUID-compatible, and hence not qualifying as a drop-in replacement. 

Implementing UUID v7 support on V  level would have a huge impact for addressing this need and enabling the adoption of the new standard. Since it is storage-compatible with UUID v4, it could already be adopted in existing databases with very little changes.

UUID v7 was already published as an RFC, see https://www.rfc-editor.org/info/rfc9562, so it should be safe to implement it already.

There is an existing Rust implementation: https://crates.io/crates/uuid7 - it implements the published RFC, and can be used to provide support.

### Use Case

rand.uuid_v7()

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.10 87d3a1d

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 86536e452e6248056666c750668334c7da87202e.87d3a1d
|:-------------------|:-------------------
|OS                  |linux, Deepin 23.1
|Processor           |6 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
|Memory              |0.38GB/7.61GB
|                    |
|V executable        |/home/Jengro/.vmr/versions/v_versions/v_latest/v
|V last modified time|2025-04-24 00:28:36
|                    |
|V home dir          |OK, value: /home/Jengro/.vmr/versions/v_versions/v_latest
|VMODULES            |OK, value: /home/Jengro/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/Jengro/Documents/Dev/v-product/vprod-workspase/core-api/internal/logic/admin/role
|                    |
|Git version         |git version 2.47.2
|V git status        |weekly.2025.16-34-g87d3a1d2
|.git/config present |true
|                    |
|cc version          |cc (Deepin 12.3.0-17deepin12) 12.3.0
|gcc version         |gcc (Deepin 12.3.0-17deepin12) 12.3.0
|clang version       |Deepin clang version 17.0.6 (5deepin5)
[...]

PR #24313, opened 2025-04-24
rand: add uuid_v7(), session function, simplify uuid_v4()

kbkpbot:
Feature request #24311


spytheman (approved):
Good work.
What is the performance of the new uuid_v4 implementation compared to the old one?

JalonSolov:
Ooo... gotta love it.  Add feature AND increase performance at the same time?  Nice...

PR #24314, opened 2025-04-25
v: improve support for rv64 inline asm

raw-bin:
Please consider this PR that:

1. Adds a local test runner specifiable via the following comments in the test file (for example):
// vtest vflags: -cc riscv64-linux-gnu-gcc -gc none
// vtest runner: qemu-riscv64
// vtest runner_flags: -L /usr/riscv64-linux-gnu

   The runner invocation in the above example requires that: 
      a. The [qemu-user](https://packages.ubuntu.com/search?suite=noble&arch=any&mode=exactfilename&searchon=contents&keywords=qemu-user) package is installed on the host system
      b. The compiler indicated by the '-cc' option (riscv64-linux-gnu-gcc in this example) is installed on the host system

2. Adds support for parsing rv64 inline assembly using the following addressing modes:

   a. Displacement from base register (eg: sd a0, 0(a1))
   b. Domposite displacement from base register (eg: sd a0, (8 * 1)(a1))
   c. Adds a test for those new addressing modes

Note that I have run the following successfully:

- make check
- v vlib/v/gen/c/coutput_test.v


vlib/v/gen/c/cgen.v
```diff
@@ -3168,7 +3168,11 @@ fn (mut g Gen) asm_stmt(stmt ast.AsmStmt) {
 		}
 
 		if !template.is_label {
-			g.write(';')
+			if stmt.arch == .rv64 {
+				g.write('\\n')
+			} else {
+				g.write(';')
+			}
```
spytheman, line 3175: why is the architecture relevant here?

vlib/v/gen/c/testdata/inline_asm_rv64_displacement_with_base_addressing.vv
```diff
@@ -0,0 +1,57 @@
...
+	element_0 u64
+	element_1 u64
+}
+
+fn main() {
+	if check_displacement_with_base() == true && check_composite_displacement_with_base() == true {
+		println('pass')
+	} else {
+		println('fail')
+	}
```
spytheman, line 16: It will be more readable, and easier to diagnose the failed condition on the CI remotely, if you have this instead:
```suggestion
	if check_displacement_with_base() {
		println('pass displacement_with_base')
	} else {
		println('fail displacement_with_base')
	}
	if check_composite_displacement_with_base() {
		println('pass composite_displacement_with_base')
	} else {
		println('fail composite_displacement_with_base')
	}
```

spytheman: (the code is more verbose, but also easier to follow)

vlib/v/gen/c/coutput_test.v
spytheman, line 1: Excellent work on adding support for `vtest runner` and `vtest runner_flags`.
Thank you 🙇🏻‍♂️ .

vlib/v/builder/cc.v
```diff
@@ -341,7 +341,7 @@ fn (mut v Builder) setup_ccompiler_options(ccompiler string) {
 		if ccoptions.cc != .tcc && current_os == 'macos' {
 			ccoptions.linker_flags << '-Wl,-export_dynamic' // clang for mac needs export_dynamic instead of -rdynamic
 		} else {
-			if v.pref.ccompiler != 'x86_64-w64-mingw32-gcc' {
+			if v.pref.ccompiler != 'x86_64-w64-mingw32-gcc' && !v.pref.is_bare {
```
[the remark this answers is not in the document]
raw-bin: It's not. It shouldn't have been a part of the series. I'll split it out.

vlib/v/gen/c/cgen.v
raw-bin: I found that for some reason ricv64 gas wouldn't accept the semicolon. Without this the generated C code containing the inline asm bombs out at compile time.

vlib/v/parser/parser.v
```diff
@@ -1410,6 +1425,63 @@ fn (mut p Parser) asm_stmt(is_top_level bool) ast.AsmStmt {
...
+									.plus {
+										'+'
+									}
+									.minus {
+										'-'
+									}
+									.mul {
+										'*'
+									}
+									else {
```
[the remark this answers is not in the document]
raw-bin: Will do!

vlib/v/gen/c/coutput_test.v
raw-bin: Thank you! :)

raw-bin:
I'll spin things around in a bit and raise another PR. Thanks for your review!

PR #24319, opened 2025-04-25
tests: enable gcc and clang for closure

tankf33der:
The generated file with multiple closure variants must always be successful for all architectures under both gcc and clang.

Very important.

p.s. Made a PR and realized that I don't know if CI will break because `clang` is not installed :(

### DIFF
--- a/vlib/v/tests/fns/closure_generator_test.v
+++ b/vlib/v/tests/fns/closure_generator_test.v
@@ -200,9 +200,11 @@ fn test_closure_return_${styp}_${i}() ! {
 	full_path_to_target := os.join_path(wrkdir, 'closure_return_test.v')
 	os.write_file(full_path_to_target, code)!
 	vexe := os.getenv('VEXE')
-	cmd := '${os.quoted_path(vexe)} -keepc -cg -showcc ${full_path_to_target}'
+	mut cmd := '${os.quoted_path(vexe)} -cc gcc -keepc -cg -showcc ${full_path_to_target}'
 	res := os.execute(cmd)
-	if res.exit_code != 0 {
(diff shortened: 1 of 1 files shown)


spytheman:
There may be a misunderstanding, not sure 🤔 - currently, if `-cc gcc` or `-cc clang` is not passed explicitly, then VFLAGS still is, and in `.github/workflows/linux_ci.yml`, there are several different jobs, that make sure to use `VFLAGS: -cc clang` and `VFLAGS: -cc gcc` .

spytheman:
Have a look at <https://github.com/vlang/v/actions/runs/14670731749/job/41176435869?pr=24319> and <https://github.com/vlang/v/actions/runs/14670731749/job/41176435853?pr=24319> .

tankf33der:
1. Now i've installed aarch alpine.
2. Proved arm64 closure is wrong.
3. I need to clarify my idea - besides Tier1, there are small and incomplete CIs (riscv64, s390x or future ones) that also need to check `all` closures. This is the kernel of arch. Must have.

After this PR with or without modifications future PRs will contain:
1. add to small CIs like s390x and riscv64 one line:
```
./v test $(find vlib/ -type f -name "*closure*_test.v")
```
2. fix arm64 closure
3. remove the stub without workaround for arm64's closure bug
4. there was something else, but I forgot

Let's have verification of all closures for all compilers for all architectures.

spytheman:
You can do `VFLAGS='-cc gcc' VTEST_ONLY=closure ./v test .` etc, in the corresponding .yml files, if you want to filter for only closure related test files:
image

spytheman:
In any case, I think that this specific PR is not useful in its current state - instead of increasing testing, it actually reduces it, and limits it to specific compilers (gcc and clang), while otherwise it currently works for all C compilers, set by VFLAGS, including msvc.

PR #24332, opened 2025-04-26
cgen: fix arm64 closure + remove stub in test

tankf33der:
Let closure work with any number of arguments in arm64 too:
o) use floating point register
o) remove stub in test


spytheman (approved):
Excellent work.
Thank you 🙇🏻‍♂️.

PR #24343, opened 2025-04-27
v: add initial support for loongarch64

tankf33der:
Highlights:
o) dev&test on Alpine
o) all closure works
o) `vlib/v/tests/` all passed too


spytheman (approved):
Excellent work.

PR #24350, opened 2025-04-28
builtin: add string is_identifier()

kbkpbot:
function to check if a string is a valid identifier (starts with letter/underscore, followed by letters, digits, or underscores)

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -2767,11 +2767,29 @@ pub fn (name string) match_glob(pattern string) bool {
 	return true
 }
 
-// is_ascii returns true  if all characters belong to the US-ASCII set ([` `..`~`])
+// is_ascii returns true if all characters belong to the US-ASCII set ([` `..`~`])
 @[inline]
 pub fn (s string) is_ascii() bool {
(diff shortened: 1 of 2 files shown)


vlib/builtin/string.v
```diff
@@ -2778,6 +2778,24 @@ pub fn (s string) is_ascii() bool {
 	return true
 }
 
+// is_identifier checks if a string is a valid identifier (starts with letter/underscore, followed by letters, digits, or underscores)
+@[inline]
```
spytheman, line 2782:
```suggestion
@[direct_array_access; inline]
```

spytheman: Without it, if you have a file that is just:
```v
dump('abc'.is_identifier())
```
it will fail on macos with a cgen error:
image

It works with `-no-skip-unused`.

spytheman: I am not sure if the inline tag is needed at all - the function is a bit long and complex (several ifs and a loop), to benefit much from it.

spytheman (approved):
Excellent work.

Issue #24331, opened 2025-04-26
test fails under linux arm64


tankf33der:
`vlib/v/tests/fixed_array_update_expr_test.v` fails brutal way

o) fails under alpine
o) hangs under ubuntu and debian

```
6580948a74c3:~# ~/v/v run fixed_array_update_expr_test.v 
/tmp/v_0//root/v/vlib/v/gen/c/struct.v:319: at v__gen__c__Gen_struct_init: RUNTIME ERROR: invalid memory access
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:3894: by v__gen__c__Gen_expr
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:3852: by v__gen__c__Gen_expr
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:3081: by v__gen__c__Gen_expr_with_cast
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:6197: by v__gen__c__Gen_return_stmt
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:2523: by v__gen__c__Gen_stmt
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:2203: by v__gen__c__Gen_stmts_with_tmp_var
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:2040: by v__gen__c__Gen_stmts
/tmp/v_0//root/v/vlib/v/gen/c/fn.v:497: by v__gen__c__Gen_gen_fn_decl
/tmp/v_0//root/v/vlib/v/gen/c/fn.v:139: by v__gen__c__Gen_fn_decl
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:2415: by v__gen__c__Gen_stmt
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:2203: by v__gen__c__Gen_stmts_with_tmp_var
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:2040: by v__gen__c__Gen_stmts
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:928: by v__gen__c__Gen_gen_file
/tmp/v_0//root/v/vlib/v/gen/c/cgen.v:478: by v__gen__c__gen
/tmp/v_0//root/v/vlib/v/builder/cbuilder/cbuilder.v:80: by v__builder__cbuilder__gen_c
[...]

PR #24354, opened 2025-04-29
cgen: workaround tcc aarch64 bug #24331

kbkpbot:
This PR provide a workaround for issue #24331


spytheman (approved):
Excellent find.
Thank you 🙇🏻‍♂️.

PR #24355, opened 2025-04-29
cgen: fix riscv32 closure

tankf33der:
I managed to set up my `riscv32` in qemu and fix the closure

Highlights:

- using a floating register
- now all architectures use a floating register
- the same order of lines as in `riscv64`
- supporting any number of arguments in closure
- verified in gcc14 and clang19
- verified personally


spytheman (approved):
Excellent work.
Thank you 🙇🏻‍♂️.

PR #24359, opened 2025-04-29
tools: vreduce timeout, vreduce custom run command

Eliyaan:
- adds a flag to specify a custom command to run the project with 
- adds a flag to specify to vreduce it needs to copy the whole project, not just the specified file
- adds a flag to set a timeout in case the command takes too long (the command could be stuck in an infinite loop at runtime)

### DIFF
--- a/cmd/tools/vreduce.v
+++ b/cmd/tools/vreduce.v
@@ -1,4 +1,5 @@
 import os
+import v.vmod
 import flag
 import math
 import log
@@ -8,8 +9,6 @@ const default_command = '${os.quoted_path(@VEXE)} -no-skip-unused' // Command us
 const default_error_msg = 'C compilation error' // the pattern to reproduce
 // Temporary files
 const tmp_folder = os.join_path(os.vtmp_dir(), 'vreduce')
-const tmp_reduced_code_file_name = '__v_reduced_code.v'
-const path = '${tmp_folder}/${tmp_reduced_code_file_name}'
 
 fn main() {
 	log.use_stdout()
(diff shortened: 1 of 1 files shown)


cmd/tools/vreduce.v
```diff
@@ -46,39 +48,107 @@ fn main() {
...
+
+	// cleanse
+	if os.exists(tmp_folder) {
+		os.rmdir_all(tmp_folder)!
+	}
 }
 
 // Return true if the command ran on the file produces the pattern
-fn string_reproduces(file string, pattern string, command string) bool {
+fn string_reproduces(file string, pattern string, command string, path string, debug bool, timeout int) bool {
```
spytheman, line 97: In general, it is better to have different names than a pair of `file`/`path` used as parameters, since that can be a bit confusing, because both can mean the same or not, depending on the context.

I suggest using `folder` or `dir` here, instead of `path` or use `filename` instead of `file`.

cmd/tools/vreduce.v
```diff
@@ -46,39 +48,107 @@ fn main() {
...
 	if !os.exists(tmp_folder) {
 		os.mkdir(tmp_folder) or { panic(err) }
 	}
 	os.write_file(path, file) or { panic(err) }
-	res := os.execute(command + ' ' + path)
-	if res.output.contains(pattern) {
+	mut output := ''
```
spytheman, line 102: You can just `res := os.execute('v timeout 2.5 ${command} ${path}')` for example, instead of using os.new_process here, if the only goal is to have a timeout.

spytheman: It will exit with exit code 124, if the timeout passes first.

Eliyaan: neat, I did not know V had a timeout command

PR #24369, opened 2025-04-30
tools: vreduce: breadth first search

Eliyaan:
- uses breadth first search instead of depth first search

### DIFF
--- a/cmd/tools/vreduce.v
+++ b/cmd/tools/vreduce.v
@@ -94,11 +94,11 @@ fn main() {
 }
 
 // Return true if the command ran on the file produces the pattern
-fn string_reproduces(file string, pattern string, command string, path string, debug bool, timeout int) bool {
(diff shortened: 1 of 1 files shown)


cmd/tools/vreduce.v
```diff
@@ -94,11 +94,11 @@ fn main() {
 }
 
 // Return true if the command ran on the file produces the pattern
-fn string_reproduces(file string, pattern string, command string, path string, debug bool, timeout int) bool {
+fn string_reproduces(file_content string, pattern string, command string, dir string, debug bool, timeout int) bool {
 	if !os.exists(tmp_folder) {
 		os.mkdir(tmp_folder) or { panic(err) }
 	}
-	os.write_file(path, file) or { panic(err) }
+	os.write_file(dir, file_content) or { panic(err) }
```
spytheman, line 101: Renaming `file` to `file_content` is good, but `path` to `dir`, given the usage is not, since write_file expects a path or file name, not a dir in its first argument.

PR #24384, opened 2025-05-01
tests: enable closure_test.v for every CPU arch

tankf33der:
Every CPU arch have `closure` implemented.
Enable good tests.


spytheman (approved):
Excellent.
Thank you 🙇🏻‍♂️.

PR #24388, opened 2025-05-02
table: fix array of scoped struct

### LINKED ISSUE #24387: Scoping escape when unit testing using anonymous structs and arrays of test cases
### Describe the bug

When using anonymous structs in test methods together with arrays of the struct instances, the scope of the structs is escaping and crossing test methods.

### Reproduction Steps

```
module main

fn test_a() {
	struct Ak {
		a int
		b int
	}
	cases := [
		Ak{1,1},
		Ak{2,2},
	]
	for _, k in cases {
		assert k.a == k.b
	}
}

fn test_b() {
	struct Ak {
		a int
		b int
		c int
    }
	cases := [
		Ak{1,2,2},
		Ak{2,2,2},
	]
	for _, k in cases {
		assert k.b == k.c
	}
}
```

### Expected Behavior

I would expect, for this example, for the tests to pass, or at least fail because of an assertion failure; instead, what I'm seeing is a compiler failure as a result of the wrong struct definition being used in the second test case.

### Current Behavior

This is trimmed, without the V test header/footer, and without the additional test failure which is just a side-effect of this issue.

Running `v test .` on the provided code produces:

```
ta_sample_test.v:28:19: error: type `Ak` has no field named `c`.
2 possibilities: `a`, `b`.
   26 |     ]
   27 |     for _, k in cases {
   28 |         assert k.b == k.c
      |                         ^
   29 |     }
   30 | }
```

### Possible Solution

I can only guess that the way `assert` is implemented is following a branch that's allowing the scope from one test case to leak into another, possibly because of an optimization that's happening because the struct instances are being tucked away in an array. Again, this error does not happen if the variables are defined outside of an array. It also does not happen if the structs are named differently; change the struct in `test_b()` to `struct Am {` and the error disappears.

It's evident that the anonymous struct from `test_a()` is escaping the method scope and being used by the V testing code instead of the anonymous struct by the same name in `test_b()`. A work-around is to change the names of the structs in every method.

This also does not happen if the test cases are _not_ in an array; if the structs are used to create objects _outside_ of an array, then the attribute errors do not occur. So, another work-around is to not use loops to test cases, but to manually write every assertion.

### V version

0.4.10

Linux 6.14.3-arch1-1 x86_64

> [!NOTE]
>

felipensp:
Fix #24387

### DIFF
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -966,7 +966,11 @@ pub fn (t &Table) array_name(elem_type Type) string {
 	ptr := if elem_type.is_ptr() { '&'.repeat(elem_type.nr_muls()) } else { '' }
 	opt := if elem_type.has_flag(.option) { '?' } else { '' }
 	res := if elem_type.has_flag(.result) { '!' } else { '' }
-	return '[]${opt}${res}${ptr}${elem_type_sym.name}'
+	mut name := elem_type_sym.name
+	if elem_type_sym.info is Struct && elem_type_sym.info.scoped_name != '' {
+		name = elem_type_sym.info.scoped_name
+	}
+	return '[]${opt}${res}${ptr}${name}'
 }
 
 @[inline]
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #24393, opened 2025-05-02
comptime: fix `$if var.return_type == 1 {`

### LINKED ISSUE #24391: Cannot test for void return_type
### Describe the bug

You seem to be unable to test whether a `FunctionData#return_type` is `void`.

### Reproduction Steps

```v
struct Struct {}

fn (s Struct) func() {}

fn main() {
	$for method in Struct.methods {
		println('${method.name}: ${method.return_type}')

		$if method.return_type {
			println('this runs but it is true regardless of the return type')
		}
		$if method.return_type is ? {
			println('this never runs but I think it should')
		}
		$if method.return_type !is ? {
			println('this never runs either and that makes sense')
		}
		$if method.return_type == 1 {
			println('this runs which is kind of expected')
		}
		$if method.return_type == 2 {
			println('this runs too which is REALLY WEIRD')
		}
	}
}
```

```shell
$ v run .
func: 1
this runs but it is true regardless of the return type
this runs which is kind of expected
this runs too which is REALLY WEIRD
```

### Expected Behavior

Based on [this](https://github.com/vlang/v/blob/ac7512b2c285d64d4c9a36c67920f963408e9157/vlib/v/ast/types.v#L660-L698) I think `method.return_type is void` should be a test that works, or at least `is ?` which is what the compiler yells at you into using.

I also don't understand why `return_type` is true when compared against any integer. It should either work as expected with `return_type == 1` working in this case (and `== 42` being made to _not_ work), or if that was never intended than no integer comparison should pass.

This should probably be the output of the above example:

```shell
$ v run .
func: 1
this runs but it is true regardless of the return type
this never runs but I think it should
this runs which is kind of expected
```

### Current Behavior

- `return_type is void` says to "use `?` instead of `?void`"
- `return_type is ?void` says the same
- `return_type is ?` compiles but does not pass
- `return_type == 1` is a false positive, because it turns out comparing return_type to _any_ int passes

### Possible Solution

_No response_

_No response_

### V version

V 0.4.10 875c165

|V full version      |V 0.4.10 b76b1ee76328ce62a6a7e149f89b83ba786ad603.875c165
|OS                  |linux, "Artix Linux"
|Processor           |32 cpus, 64bit, little endian, AMD Ryzen 9 7950X3D 16-Core Processor
|Memory              |4.14GB/62GB
|                    |
|V executable        |/home/ross/x/v/v
|V last modified time|2025-05-02 14:46:13
|                    |
|V home dir          |OK, value: /home/ross/x/v
|VMODULES            |OK, value: /home/ross/.vmodules
|Current working dir |OK, value: /tmp/ciea
|                    |
|V git status        |weekly.2025.17-39-g875c1654
|.git/config present |true
|                    |
|cc version          |cc (GCC) 14.2.1 20250207
|gcc version         |gcc (GCC) 14.2.1 20250207
|emcc version        |N/A
> [!NOTE]
>

felipensp:
Fix #24391

### DIFF
--- a/vlib/v/checker/if.v
+++ b/vlib/v/checker/if.v
@@ -182,11 +182,10 @@ fn (mut c Checker) if_expr(mut node ast.IfExpr) ast.Type {
 								} else {
 									.skip
 								}
-							} else if comptime_field_name == c.comptime.comptime_for_method_var {
-								if left.field_name == 'return_type' {
-									skip_state = c.check_compatible_types(c.unwrap_generic(c.comptime.comptime_for_method_ret_type),
-										right as ast.TypeNode)
-								}
+							} else if comptime_field_name == c.comptime.comptime_for_method_var
+								&& left.field_name == 'return_type' {
+								skip_state = c.check_compatible_types(c.unwrap_generic(c.comptime.comptime_for_method_ret_type),
+									right as ast.TypeNode)
 							} else if comptime_field_name in [
 								c.comptime.comptime_for_variant_var,
 								c.comptime.comptime_for_enum_var,
--- a/vlib/v/checker/if.v
+++ b/vlib/v/checker/if.v
@@ -286,6 +286,14 @@ fn (mut c Checker) if_expr(mut node ast.IfExpr) ast.Type {
 									}
 								}
 							}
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

PR #24405, opened 2025-05-04
tools: vreduce: more frequent saves, fix timeout leaks

Eliyaan:
- make a save at each big step, in case the user does not want to wait more
- timeout option now kills all the children processes, to avoid children leaks when using `v run file.v` with timeout option


spytheman (approved):
Excellent work.

PR #24410, opened 2025-05-04
cgen: fix `if mut var != none {` generated code

### LINKED ISSUE #24351: assigning to mutable optional give v error: `next` is immutable, declare it with `mut` to make it mutable
<details>
<summary>V version: V 0.4.10 880f216, press to see full `v doctor` output</summary>

|V full version      |V 0.4.10 5c65e58.880f216
|Processor           |24 cpus, 64bit, little endian, AMD Ryzen 9 3900X 12-Core Processor  
|Memory              |18.12GB/31.9GB
|                    |
|V executable        |C:\v\v.exe
|V last modified time|2025-04-28 07:34:43
|                    |
|V home dir          |OK, value: C:\v
|VMODULES            |OK, value: C:\Users\info\.vmodules
|VTMP                |OK, value: C:\Users\info\AppData\Local\Temp\v_0
|Current working dir |OK, value: D:\vbug2
|                    |
|Git version         |git version 2.42.0.windows.2
|V git status        |weekly.2025.17-12-g880f2169-dirty
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |N/A
|clang version       |clang version 17.0.1
|msvc version        |N/A
|emcc version        |N/A
|glibc version       |ldd (cygwin) 3.4.7

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  main.v && D:\vbug2\main.exe`
```v
module main

@[heap]
interface IGameObject {
mut:
	name       string
	parent     ?&IGameObject
	next       ?&IGameObject
	child      ?&IGameObject
	last_child ?&IGameObject
}

@[heap]
struct GameObject implements IGameObject {
mut:
	name       string
	parent     ?&IGameObject
	next       ?&IGameObject
	child      ?&IGameObject
	last_child ?&IGameObject
}

fn main() {
	mut v1 := &GameObject{
		name: 'v1'
	}

	mut next := v1.next
	for {
		if next != none {
			eprintln(next.name)
			next = next.next
		} else {
			break
		}
	}
}

```

**What did you see?**
```
main.v:32:4: error: `next` is immutable, declare it with `mut` to make it mutable
   30 |         if next != none {
   31 |             eprintln(next.name)
   32 |             next = next.next
      |             ~~~~
   33 |         } else {
   34 |             break
```

no v error

> [!NOTE]
>

felipensp:
Fix #24351

### DIFF
--- a/vlib/v/tests/options/option_mut_if_none_test.v
+++ b/vlib/v/tests/options/option_mut_if_none_test.v
@@ -1,0 +1,42 @@
+module main
+
+@[heap]
+interface IGameObject {
+mut:
+	name       string
+	parent     ?&IGameObject
+	next       ?&IGameObject
+	child      ?&IGameObject
+	last_child ?&IGameObject
+}
+
+@[heap]
+struct GameObject implements IGameObject {
+mut:
+	name       string
+	parent     ?&IGameObject
+	next       ?&IGameObject
+	child      ?&IGameObject
+	last_child ?&IGameObject
+}
+
+fn test_main() {
+	mut v1 := &GameObject{
+		name: 'v1'
+	}
+	v1.next = &GameObject{
+		name: 'v2'
+	}
+
+	mut next := v1.next
+	for {
+		if mut next != none {
+			eprintln(next.name)
+			assert next.name == 'v2'
+			next = next.next
+		} else {
+			break
+		}
+	}
+	assert next == none
+}
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5265,7 +5265,11 @@ fn (mut g Gen) ident(node ast.Ident) {
 						cast_sym := g.table.sym(g.unwrap_generic(typ))
 						if obj_sym.kind == .interface && cast_sym.kind == .interface {
 							ptr := '*'.repeat(node.obj.typ.nr_muls())
-							g.write('I_${obj_sym.cname}_as_I_${cast_sym.cname}(${ptr}${node.name})')
+							if cast_sym.cname != obj_sym.cname {
+								g.write('I_${obj_sym.cname}_as_I_${cast_sym.cname}(${ptr}${node.name})')
+							} else {
+								g.write(node.name)
+							}
 						} else {
 							mut is_ptr := false
 							if i == 0 {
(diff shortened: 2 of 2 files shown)


JalonSolov:
Is this actually valid syntax?  `if mut var` seems more like a check to see if the var is mutable.  There shouldn't be any need to have the `mut` keyword... it should be invalid syntax.

felipensp:
> Is this actually valid syntax? `if mut var` seems more like a check to see if the var is mutable. There shouldn't be any need to have the `mut` keyword... it should be invalid syntax.

It's following the `if mut var is Foo {` approach.

JalonSolov:
So as long as bad syntax is allowed in one place, it should be allowed everywhere?  <sigh>

Might as well continue.

vlib/v/tests/options/option_mut_if_none_test.v
```diff
@@ -0,0 +1,42 @@
...
+	v1.next = &GameObject{
+		name: 'v2'
+	}
+
+	mut next := v1.next
+	for {
+		if mut next != none {
+			eprintln(next.name)
+			assert next.name == 'v2'
+			next = next.next
```
spytheman, line 36: Can we add a simpler test, that just uses `field ?int` for example?
I am afraid that the code from the issue is becoming too complex for the feature is, and the only other place that uses this syntax seems to be:
`vlib/v/tests/generics/generic_selector_test.v:29:if mut curr_node != none {`
which is also a bit more complex (it involves generics as well).

spytheman:
```v
struct Abc {
	field ?int
}

fn test_if_mut_opt() {
	dump('start')
	mut s := ?string('abc')
	dump(s)
	if mut s != none {
		dump(s)
		dump(typeof(s).name)
		assert s == 'abc'
	} else {
		assert false
	}
	mut a := Abc{
		field: 123
	}
	dump(a.field)
	if mut a.field != none {
		dump(a.field)
		dump(typeof(a.field).name)
		assert a.field == 123
	} else {
		assert false
	}
	dump('end')
}
```

Something like this for example?

spytheman: Although that does not fail on master... It may be better to add the new test separately, outside this PR.

PR #24411, opened 2025-05-04
cgen: fix interface `unsafe {nil}` comparison and initialization

### LINKED ISSUE #24374: Check reference to interface against nil requires cast.
### Describe the bug

The assert in the code below should not fail

### Reproduction Steps

```v
interface Cfg { id string }
struct XX { cfg &Cfg = unsafe {nil} }
xx := XX{}
assert xx.cfg == unsafe{nil}
```

### Expected Behavior

Assert should not fail

### Current Behavior

Assert fails
```
code.v:4:NaN: fn main.main: assert xx.cfg == unsafe { nil }
  left value: xx.cfg = &Cfg(0x0)
  right value: unsafe { nil } = 0x0
V panic: Assertion failed...
v hash: 80a9d27
/home/admin/v/vlib/builtin/builtin.c.v:146: at _v_panic: Backtrace
/box/code.v:6: by main__main
/tmp/v_60000/code.01JT3CRGG6MB1VHTAEE51X1G9V.tmp.c:11879: by main
Exited with error status 1
```

### Possible Solution

```
assert xx.cfg == unsafe{ &Cfg(nil) }
```

_No response_

### V version

V 0.4.10 ead4d90

|V full version      |V 0.4.10 da3112e.ead4d90
|OS                  |macos, macOS, 15.4.1, 24E263
|Memory              |0.19GB/8GB
|                    |
|V last modified time|2025-04-29 18:40:43
|                    |
|Current working dir |OK, value: /Users/mike/Documents/github/gui/src/examples
|                    |
|env VFLAGS          |"-message-limit 5 -path /Users/mike/Documents/github/gui/src|@vlib|@vmodules"
|                    |
|V git status        |weekly.2025.17-21-gead4d90a (8 commit(s) behind V master)
|.git/config present |true
|                    |
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

felipensp:
Fix #24374

### DIFF
--- a/vlib/v/gen/c/auto_eq_methods.v
+++ b/vlib/v/gen/c/auto_eq_methods.v
@@ -258,7 +258,13 @@ fn (mut g Gen) gen_struct_equality_fn(left_type ast.Type) string {
 				&& (!field.typ.has_flag(.option) || !field.typ.is_ptr()) {
 				ptr := if field.typ.is_ptr() { '*'.repeat(field.typ.nr_muls()) } else { '' }
 				eq_fn := g.gen_interface_equality_fn(field.typ)
+				if ptr != '' {
+					fn_builder.write_string('((${left_arg} == (void*)0 && ${right_arg} == (void*)0) || (${left_arg} != (void*)0 && ${right_arg} != (void*)0 && ')
+				}
 				fn_builder.write_string('${eq_fn}_interface_eq(${ptr}${left_arg}, ${ptr}${right_arg})')
+				if ptr != '' {
+					fn_builder.write_string('))')
+				}
 			} else if field.typ.has_flag(.option) {
 				fn_builder.write_string('!memcmp(&${left_arg}.data, &${right_arg}.data, sizeof(${g.base_type(field.typ)}))')
 			} else {
--- a/vlib/v/gen/c/infix.v
+++ b/vlib/v/gen/c/infix.v
@@ -1217,6 +1217,8 @@ fn (mut g Gen) gen_plain_infix_expr(node ast.InfixExpr) {
 	}
 	is_ctemp_fixed_ret := node.op in [.eq, .ne] && node.left is ast.CTempVar
 		&& node.left.is_fixed_ret
+	is_nil_cmp := node.right is ast.UnsafeExpr && node.right.expr is ast.Nil
+		&& g.table.final_sym(node.left_type).kind == .interface
 	if is_ctemp_fixed_ret {
 		if node.op == .eq {
 			g.write('!')
@@ -1224,9 +1226,9 @@ fn (mut g Gen) gen_plain_infix_expr(node ast.InfixExpr) {
 		g.write('memcmp(')
 	}
 	g.expr(node.left)
-	if !is_ctemp_fixed_ret {
+	if !is_ctemp_fixed_ret && !is_nil_cmp {
 		g.write(' ${node.op.str()} ')
-	} else {
+	} else if !is_nil_cmp {
 		g.write(', ')
 	}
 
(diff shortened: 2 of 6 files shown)


vlib/v/tests/interfaces/interface_nil_cmp_test.v
```diff
@@ -0,0 +1,12 @@
...
+	id string
+}
+
+struct XX {
+	cfg &Cfg = unsafe { nil }
+}
+
+fn test_main() {
+	xx := XX{}
+	assert xx.cfg == unsafe { nil }
```
spytheman, line 11: `assert xx.cfg == unsafe { nil }`
works, but `assert unsafe { nil } == xx.cfg` fails
I think that both should work the same.

vlib/v/gen/c/infix.v
```diff
@@ -1237,7 +1239,12 @@ fn (mut g Gen) gen_plain_infix_expr(node ast.InfixExpr) {
 		g.write('*')
 		g.expr(node.right)
 	} else {
-		g.expr_with_cast(node.right, node.right_type, node.left_type)
+		if is_nil_cmp {
+			sym := g.table.final_sym(node.left_type)
+			g.write('->_typ == _${sym.cname}_voidptr_index')
```
spytheman, line 1244: Why is this comparing _typ, instead of the pointer of the field itself?

vlib/v/tests/interfaces/interface_nil_cmp_test.v
spytheman: Here, the assert generates `unsafe { nil } == xx.cfg` generates:
`if (((void*)0) == xx.cfg) {` which imho is correct.

What is not correct imho is the initialization for `xx := XX{}`, which generated:
```c
main__XX xx = ((main__XX){.cfg = HEAP(main__Cfg, I_voidptr_to_Interface_main__Cfg(HEAP(voidptr, (((void*)0))))),});
```

Given `cfg &Cfg = unsafe { nil }`, I think this should have produced just something similar to:
```c
main__XX xx = ((main__XX){.cfg = (void*)0 })
```
i.e. a NULL pointer for the .cfg field, not a pointer to an interface value, that has itself a NULL pointer as a value.

vlib/v/gen/c/infix.v
spytheman: I saw the reason (mentioned in the comment above). I think this is the wrong place for the fix. Comparing a pointer to an interface to nil, should compare the pointer itself, not the pointer value, that is stored in the interface value.

vlib/v/tests/interfaces/interface_nil_cmp_test.v
felipensp: Hmm, Ok, good point. Thanks.

spytheman, line 11: please check also for:
```v
assert unsafe { nil } == xx.cfg
```

spytheman (commented):
Excellent work.

vlib/v/gen/c/auto_eq_methods.v
```diff
@@ -258,7 +258,13 @@ fn (mut g Gen) gen_struct_equality_fn(left_type ast.Type) string {
 				&& (!field.typ.has_flag(.option) || !field.typ.is_ptr()) {
 				ptr := if field.typ.is_ptr() { '*'.repeat(field.typ.nr_muls()) } else { '' }
 				eq_fn := g.gen_interface_equality_fn(field.typ)
+				if ptr != '' {
+					fn_builder.write_string('((${left_arg} == (void*)0 && ${right_arg} == (void*)0) || (${left_arg} != (void*)0 && ${right_arg} != (void*)0 && ')
```
spytheman, line 262: hm, I could not find what kind of V source would generated this 🤔

felipensp: I got failure on some tests without handling them.

Issue #24415, opened 2025-05-05
C compiler type error


martinskou:
<details>
<summary>V version: V 0.4.10 a1d9450, press to see full `v doctor` output</summary>

|V full version      |V 0.4.10 610558d.a1d9450
|:-------------------|:-------------------
|OS                  |macos, macOS, 13.7.3, 22H417
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i7-7567U CPU @ 3.50GHz
|Memory              |1.01GB/16GB
|                    |
|V executable        |/Users/martin/Development/v/v/v
|V last modified time|2025-05-05 08:43:45
|                    |
|V home dir          |OK, value: /Users/martin/Development/v/v
|VMODULES            |OK, value: /Users/martin/.vmodules
|VTMP                |OK, value: /tmp/v_501
|Current working dir |OK, value: /Users/martin/Development/bugs
|                    |
|Git version         |git version 2.49.0
|V git status        |weekly.2025.17-54-ga1d94500
|.git/config present |true
|                    |
|cc version          |Apple clang version 15.0.0 (clang-1500.1.0.2.5)
|gcc version         |Apple clang version 15.0.0 (clang-1500.1.0.2.5)
|clang version       |Apple clang version 15.0.0 (clang-1500.1.0.2.5)
|tcc version         |tcc version 0.9.27 (x86_64 Darwin)
|tcc git status      |thirdparty-macos-amd64 975f1ad8
|emcc version        |N/A
|glibc version       |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  bug1.v && bug1`
```v
module main

struct Record {
mut:
    f1 int
}

fn map_data[T](data map[string]string) T {
    mut t:=T{}
    $for field in T.fields {
        t.$(field.name)=data[field.name]
    }
    return t
}


fn main() {
    mut data:=map[string]string{}
    data["f1"]="123"
    t:=map_data[Record](data)
    println("bug1 : ${t}")
}

```

**What did you see?**
```
================== C compilation error (from cc): ==============
cc: /tmp/v_501/bug1.01JTFXQQWXKW4PPGMY8RQF05HQ.tmp.c:6967:9: error: assigning to 'int' from incompatible type 'string' (aka 'struct string')
cc:                         t.f1 = (*(string*)map_get(ADDR(map, data), &(string[]){field.name}, &(string[]){ (string){.str=(byteptr)"", .is_lit=1} }));
[...]

PR #24421, opened 2025-05-05
checker: check invalid comptime field name assignment 

Delta456:
Fixes https://github.com/vlang/v/issues/24415


vlib/v/checker/match.v
```diff
@@ -417,6 +417,7 @@ fn (mut c Checker) match_exprs(mut node ast.MatchExpr, cond_type_sym ast.TypeSym
 			match mut expr {
 				ast.TypeNode {
 					key = c.table.type_to_str(expr.typ)
+					c.inside_x_matches_type = true
```
spytheman, line 420: You need to remember the old state, and restore it after the branch scope is finished.
If you do not do that, `c.inside_x_matches_type` will remain stuck to `true` after the first type match branch in a .v file, since the only other place that resets it to false is inside the `reset_checker_state_at_start_of_new_file` method.

Delta456: Something like this, then?

```v
ast.TypeNode {
		key = c.table.type_to_str(expr.typ)
		c.inside_x_matches_type = true
		expr_types << expr
		c.inside_x_matches_type = false
}
```

spytheman: No, because you can have it been set to true by an outer branch (matches can be nested).

Instead, you can do something like this:
```v
old_state := c.inside_x_matches_type
defer { c.inside_x_matches_type = old_state }
```
at the start of `fn (mut c Checker) match_expr(mut node ast.MatchExpr) ast.Type {` in `vlib/v/checker/match.v`

vlib/v/checker/match.v
```diff
@@ -438,7 +439,15 @@ fn (mut c Checker) match_exprs(mut node ast.MatchExpr, cond_type_sym ast.TypeSym
...
 			c.expected_type = node.cond_type
+			mut old_inside_x_matches_type := c.inside_x_matches_type
+			if is_type_node {
+				c.inside_x_matches_type = true
+				old_inside_x_matches_type = c.inside_x_matches_type
+			}
 			expr_type := c.expr(mut expr)
+			if is_type_node {
+				c.inside_x_matches_type = old_inside_x_matches_type
+			}
```
spytheman, line 450: That essentially still sets `c.inside_x_matches_type` to true permanently for the rest of the file, but in a more complicated way, compared to before.

spytheman: What will set it back to false, once it becomes true?

If it should be permanently true, why have the bool field at all?

Delta456: Right. I am overthinking. Mind if you can solve this issue?

spytheman: I just removed the field completely.

Delta456: The checks are still failing

https://github.com/vlang/v/actions/runs/15005148584/job/42161765731?pr=24421

spytheman: should be fixed now

Delta456: Can we also add a test case for https://github.com/vlang/v/pull/24421#discussion_r2075110428 ?

spytheman: I'll add a test tomorrow, since it is pretty late here, and it will need several nested $for and matches.

PR #24422, opened 2025-05-05
x.json2.decoder2: fix array decoding in struct is field, when array is initialized

enghitalo:
```v
import x.json2 as json
import x.json2.decoder2

struct Bar {
	b []int = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
}

struct Foo {
	Bar
	a []int = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
}

fn test_main() {
	str := json.encode(Foo{})
	assert decoder2.decode[Foo](str)!.str() == 'Foo{
    Bar: Bar{
        b: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
    }
    a: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
}'
}
```

### DIFF
--- a/vlib/x/json2/decoder2/decode.v
+++ b/vlib/x/json2/decoder2/decode.v
@@ -618,7 +618,9 @@ fn (mut decoder Decoder) decode_value[T](mut val T) ! {
 		decoder.decode_map(mut val)!
 		return
 	} $else $if T.unaliased_typ is $array {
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Good work.

PR #24425, opened 2025-05-06
builtin: add diff_array() functions

kbkpbot:
This PR add `diff_array[T]()` functions. This will help build portable `diff` utility.

A directly conversion from https://github.com/covrom/diff
Fast diff library for Myers algorithm.
The algorithm is described in "An O(ND) Difference Algorithm and its Variations", Eugene Myers, Algorithmica Vol. 1 No. 2, 1986, pp. 251-266


spytheman:
That should not be in `builtin`.
You can add a separate `diff` module under `vlib/arrays/`, or in it.

vlib/v/checker/tests/globals/assign_global_to_shared_err.out
spytheman, line 1: this, and the other files in the folder, are not related to the PR :-|

PR #24428, opened 2025-05-06
arrays,tools: add diff(), add v diff

kbkpbot:
This PR add `diff[T]()`  function. This will help build portable diff utility.
Add a simple `v diff` tool.

A directly conversion from https://github.com/covrom/diff
Fast diff library for Myers algorithm.
The algorithm is described in "An O(ND) Difference Algorithm and its Variations", Eugene Myers, Algorithmica Vol. 1 No. 2, 1986, pp. 251-266

```
$ v diff examples/hello_world.v examples/log.v
@@ -1,4 +1,33 @@
-println('Hello, World!')
+import log
+
+fn main() {
+       // Note: you *do not* need to create a logger instance, and pass it around, just to use the `log` module.
+       // The log module already creates an instance of a thread safe Logger, and utility functions to work with it:
+       log.set_level(.debug)
+       log.debug('simple debug message')
+       log.warn('simple warning message')
+       log.info('simple information message')
+       log.error('simple error message')
+
+       mut l := log.Log{}
+       l.set_level(.info)
+       // Make a new file called info.log in the current folder
+       l.set_full_logpath('./info.log')
+       l.log_to_console_too()
+       println('Please check the file: ${l.output_file_name} after this example crashes.')
+
+       l.info('info')
+       l.warn('warn')
+       l.error('error')
+       l.debug('no output for debug')
+       l.set_level(.debug)
+       l.debug('debug now')
+       l.set_level(log.level_from_tag('INFO') or { log.Level.disabled }) // set level from string, sample
+       l.info('info again')
+       l.set_level(log.level_from_tag('') or { log.Level.disabled }) // set level from string, sample
+       l.error('no output anymore')
+       l.fatal('fatal') // panic, next statements won't be executed
+}
```


vlib/arrays/diff/diff.v
```diff
@@ -0,0 +1,296 @@
+module diff
+
+import strings
+
+// A Change contains one or more deletions or inserts
```
spytheman, line 5:
```suggestion
// DiffChange contains one or more deletions or inserts
```

examples/diff.v
spytheman, line 1: I think that it will be more useful, to have a general utility, similar to what https://github.com/covrom/diff/blob/master/cmd/diff/main.go does, i.e. have it receive CLI arguments, so that it can be used on arbitrary files, not just to compare hard coded content.

spytheman: It will be shorter too.

vlib/arrays/diff/diff.v
```diff
@@ -0,0 +1,296 @@
...
+	}
+}
+
+@[inline]
+fn max(a int, b int) int {
+	return if a > b { a } else { b }
+}
+
+@[inline]
+fn min(a int, b int) int {
```
spytheman, line 294: We have `int_min/2` in `builtin`.

vlib/arrays/diff/diff.v
```diff
@@ -0,0 +1,296 @@
...
+		if param.colorful {
+			sb.writeln('\033[32m+${line}\033[0m')
+		} else {
+			sb.writeln('+${line}')
+		}
+	}
+}
+
+@[inline]
+fn max(a int, b int) int {
```
spytheman, line 289: We have `int_max/2` in `builtin`.

vlib/arrays/diff/diff.v
```diff
@@ -0,0 +1,296 @@
...
+	b   int // position in input b []T
+	del int // delete Del elements from input a
+	ins int // insert Ins elements from input b
+}
+
+pub struct DiffContext[T] {
+mut:
+	a     []T
+	b     []T
+	flags []u8 // element bits 1 delete, 2 insert
```
spytheman, line 19: We can use a `@[flag] enum ContextFlag as u8 { delete insert }` and later `flags []ContextFlag` here instead. It can clarify the bit ops later, by changing for example:
`c.flags[y] & u8(2) != u8(0)` to `c.flags[y].has(.insert)` .
I am not sure of the performance impact though.

vlib/arrays/diff/diff.v
```diff
@@ -0,0 +1,296 @@
...
+pub struct DiffGenStrParam {
+pub mut:
+	colorful     bool
+	unified      int = 3 // how many context lines before/after diff block
+	block_header bool // output `@@ -3,4 +3,5 @@` or not
+}
+
+// gen_str generate a colorful diff string of two arrays.
+@[direct_array_access]
+pub fn (mut c DiffContext[T]) gen_str(param DiffGenStrParam) string {
```
spytheman, line 205: Why `gen_str`?
It is a public method. I think that a longer, and more descriptive name like `generate_patch` may be better.

vlib/arrays/diff/diff_test.v
spytheman, line 1: Please also add a test similar to the one in https://github.com/covrom/diff/blob/master/diff_test.go#L164 : 
```go
func TestDiffRunes(t *testing.T) {
	a := []rune("brown fox jumps over the lazy dog")
	b := []rune("brwn faax junps ovver the lay dago")
...
```, i.e. finding differences within a single string's runes. 

We could have for example something like this:
```v
fn test_diff_runes() {
    a := 'brown fox jumps over the lazy dog'.runes()
    b := 'brwn faax junps ovver the lay dago'.runes()
...
```v

File: examples/diff.v
Comment (spytheman) [reply examples/diff.v:1]: Another option is to have it as `cmd/tools/vdiff.v` and then modify `cmd/v/v.v` to enable `v diff` as a general platform independent tool, similar to `v timeout`, `v repeat` etc.

**Review by spytheman — APPROVED** [review]
Excellent work.
It will enable a lot of other improvements in the V infrastructure, that were blocked before.
Thank you 🙇🏻‍♂️.

File: vlib/arrays/diff/diff_test.v [hunk 6]
```diff
@@ -0,0 +1,133 @@
...
+-5
+-3
++max
++5
+'
+}
+
+fn test_diff_runes() {
+	mut aa := 'brown fox jumps over the lazy dog'.runes()
+	mut bb := 'brwn faax junps ovver the lay dago'.runes()
```
JalonSolov, line 88:
```suggestion
	mut bb := 'brwn faax junps ovver the lay dag'.runes()
```
simply to avoid triggering the overly sensitive...

JalonSolov: Could also be
```suggestion
	mut bb := 'brwn faax junps ovver the lay daog'.runes()
```

PR #24438, opened 2025-05-08
jsgen: cannot assign unsafe nil values (fix #24407)

### LINKED ISSUE #24407: jsgen: cannot assign unsafe nil values
### Describe the bug

Cannot assign unsafe `nil` values to mutable variables.

### Reproduction Steps

Create a small `main.v` file:
```v
fn main() {
	str := &string('hello!')
	unsafe {
		str = nil
	}
}
```

Transpile it to JS and execute it:
```bash
v -b js main.v
node main.js
```

Output:
```
Cannot read properties of null (reading 'val')
```

### Expected Behavior

The compiled application should be able to run.

### Current Behavior

The compiled application doesn't run properly.

### Possible Solution

_No response_

_No response_

### V version

0.4.10 0cb1366

|V full version      |V 0.4.10 dead5e69b44b4f8c4999ac82b730c21250cb0583.0cb1366
|OS                  |macos, macOS, 15.4.1, 24E263
|Memory              |0.32GB/16GB
|                    |
|V executable        |/Users/gonzalo/Projects/Personal/v/nv/v
|V last modified time|2025-05-03 07:31:04
|                    |
|V home dir          |OK, value: /Users/gonzalo/Projects/Personal/v/nv
|Current working dir |OK, value: /Users/gonzalo/Projects/Personal/v/myapp
|                    |
|V git status        |weekly.2025.18-6-g0cb1366c
|.git/config present |true
|                    |
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

gchumillas:
Fixes #24407


vlib/v/gen/js/tests/unsafe.v
```diff
@@ -0,0 +1,82 @@
...
+
+fn returns_nil_ptr() &int {
+	unsafe {
+		return nil
+	}
+}
+
+fn main() {
+	// Basic test of assigning nil to a string pointer
+	mut a := &string('hi!')
```
spytheman, line 32: that should not compile - `'hi'` is a string, not a pointer

vlib/v/gen/js/tests/unsafe.v
```diff
@@ -0,0 +1,82 @@
...
+	unsafe {
+		a = nil
+	}
+	JS.console.log(a)
```
spytheman, line 36: Please use `println()`, unless you are testing `JS.console.log()` itself.

It is much easier to compare the common behavior of the different backends, if the tests avoid using platform dependent features.

spytheman: If `assert a == unsafe { nil }` works, use that instead of println, until it is fixed.

spytheman: My main concern is avoiding/minimizing the divergence, between what is allowed in the C and the JS backends.

gchumillas: It's not possible either :) The following code generates a JS error:

```v
mut a := 'hi!'
mut b := &a
unsafe {
	b = nil
}

// Cannot read properties of null (reading 'valueOf')
assert b == unsafe { nil }
```

Delta456: Is it possible to generate an exception for `unsafe { nil }`, which states the same?

gchumillas: I think the problem is here:
https://github.com/vlang/v/blob/master/vlib/v/gen/js/js.v#L968

It should return `nil__` and that constant should be defined somewhere (similar to how `none__` was defined).

Please do not merge the current PR. The solution is incomplete.
Work in progress :)

PR #24440, opened 2025-05-09
ast,cgen,parser,pref: enable loongarch64 assembler + test

tankf33der:
I just wanted to add support for `loongarch64` assembly and accumulated so many changes.
If needed, I can split this into different PRs.

Highlights:
o) `ast.AsmRegister` part - 90% of the time was spent proving that everything works exactly as intended.
Did reverse engineering of the parser to find asm inline code lines that check and work for this architecture.
o) in this architecture, assembly instructions also have `dots`.
o) test file is simple.
o) the rest is trivial and was based on my similar previous patches.
o) developed and tested on x64 and loongarch64 under Alpine on qemu.

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -1839,6 +1839,12 @@ pub const ppc64le_with_number_register_list = {
 	'r#': 32
 }
 
+pub const loongarch64_no_number_register_list = []string{}
+pub const loongarch64_with_number_register_list = {
+	'f#': 32
+	'r#': 32
+}
+
 pub struct DebuggerStmt {
 pub:
 	pos token.Pos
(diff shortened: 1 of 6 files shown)


vlib/v/ast/ast.v
```diff
@@ -1839,6 +1839,12 @@ pub const ppc64le_with_number_register_list = {
 	'r#': 32
 }
 
+pub const loongarch64_no_number_register_list = []string{}
```
spytheman, line 1842: There is no need for an empty list constant, imho just inline `[]` instead of it, in the single call that uses it.

vlib/v/slow_tests/assembly/asm_test.loongarch64.v
```diff
@@ -0,0 +1,37 @@
+// vtest build: gcc
```
spytheman, line 1: Why restrict it to just gcc?
What about clang?

spytheman (approved):
Excellent work.

Issue #24291, opened 2025-04-22
`BufferedReader.read` returns `io.NotExpected: invalid copy of buffer`


einar-hjortdal:
### Describe the bug

I have encountered this issue when running [this test](https://github.com/einar-hjortdal/firebird/blob/pending/src/statement_test.v#L500).
Following spytheman's recommendation, I have added this line `println('r.offset: ${r.offset}, r.len : ${r.len}')` [here](https://github.com/vlang/v/blob/ccb3d5cfcd12f44375aed8832714398f7915251d/vlib/io/buffered_reader.v#L59) and it looks like in some circumstances `r.offset == r.len`

### Reproduction Steps

1. `git@github.com:einar-hjortdal/firebird.git && cd firebird`
2. 
```
sudo docker run \
  --rm \
  --detach \
  --name=firebird-server \
  --env=FIREBIRD_ROOT_PASSWORD=rootpwd \
  --env=FIREBIRD_USER=fbusr \
  --env=FIREBIRD_PASSWORD=fbpwd \
  --env=FIREBIRD_DATABASE=firebird.fdb \
  --env=FIREBIRD_DATABASE_DEFAULT_CHARSET=UTF8 \
  --volume=firebird-data:/var/lib/firebird/data \
  --publish=3050:3050 \
  firebirdsql/firebird
```
3. run test `v test src/statement_test.v`

### Expected Behavior

No errors

### Current Behavior

`io.NotExpected: invalid copy of buffer`

### V version

V 0.4.10 b76b1ee

### Environment details (OS name and version, etc.)

```
|V full version      |V 0.4.10 537605a.b76b1ee
|:-------------------|:-------------------
|OS                  |linux, "openmamba release 2024.6 for x86_64 (rolling)"
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz
|Memory              |15.88GB/31.16GB
|                    |
|V executable        |/home/einar/.local/lib64/v/v
|V last modified time|2025-04-22 07:10:54
|                    |
|V home dir          |OK, value: /home/einar/.local/lib64/v
|VMODULES            |OK, value: /home/einar/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/einar/Documents/projects/vlang/active/firebird
|                    |
|Git version         |git version 2.49.0
|V git status        |weekly.2025.16-33-gb76b1ee7
|.git/config present |true
|                    |
[...]

PR #24444, opened 2025-05-09
Addresses https://github.com/vlang/v/issues/24291

einar-hjortdal:
Fix suggested by spytheman.
Seems to solve the issue, can't find negatives.
I am not the author of BufferedReader, I do not know what the purpose of the check was.
Running tests might show issues, then we can investigate further.
Adding a test to prevent a regression would be nice, but I do not know what causes the issue in the first place.


einar-hjortdal:
A seemingly related test failed, so maybe this is a bad pull request. Do not merge for now

PR #24450, opened 2025-05-10
checker: add check for `nil` assign on option type

### LINKED ISSUE #24447: cgen: cannot assign nil to optional pointer fields
### Describe the bug

Example 1: https://play.vlang.io/p/849c6ad20f
Example 2: https://play.vlang.io/p/61c57af0ff

```v
// Example 1
struct Foo {
	name ?&string
}

fn main() {
	_ := Foo{
		name: unsafe { nil }
	}
}
```

```v
// Example 2
struct Foo {
mut:
	name ?&string
}

fn main() {
	mut foo := Foo{}
	unsafe {
		foo.name = nil
	}
}
```

### Reproduction Steps

Visit example links.

### Expected Behavior

Should be able to assign `nil` to optional pointer fields.

### Current Behavior

Output:

```
/box/code.v:7: error: lvalue expected
builder error: 
==================
Exited with error status 1
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.10 ed84fb9ce8beb86a8ff02d0325cfdfae5a5c58c7.ed84fb9

```
|V full version      |V 0.4.10 ed84fb9ce8beb86a8ff02d0325cfdfae5a5c58c7.ed84fb9
|Memory              |0.66GB/2.02GB
|                    |
|V last modified time|2025-05-10 08:25:28
|                    |
|                    |
|V git status        |N/A
|.git/config present |true
|                    |
|cc version          |cc (Debian 12.2.0-14) 12.2.0
|gcc version         |gcc (Debian 12.2.0-14) 12.2.0

|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.36-9+deb12u8) 2.36
```

> [!NOTE]
>

felipensp:
Fix #24447

### DIFF
--- a/vlib/v/checker/assign.v
+++ b/vlib/v/checker/assign.v
@@ -560,6 +560,10 @@ fn (mut c Checker) assign_stmt(mut node ast.AssignStmt) {
 			}
 		}
 
+		if right_type == ast.nil_type && left_type.has_option_or_result() {
+			c.error('cannot assin `nil` to Option type (use `none` instead`)', node.pos)
+		}
+
 		if left_sym.kind == .array && right_sym.kind == .array {
 			right_info := right_sym.info as ast.Array
 			right_elem_type := c.table.unaliased_type(right_info.elem_type)
--- a/vlib/v/checker/struct.v
+++ b/vlib/v/checker/struct.v
@@ -851,10 +851,13 @@ or use an explicit `unsafe{ a[..] }`, if you do not want a copy of the slice.',
 					}
 				}
 				expected_type_sym := c.table.final_sym(init_field.expected_type)
+				expected_is_option := init_field.expected_type.has_option_or_result()
 				if expected_type_sym.kind in [.string, .array, .map, .array_fixed, .chan, .struct]
-					&& init_field.expr.is_nil() && !init_field.expected_type.is_ptr()
+					&& init_field.expr.is_nil()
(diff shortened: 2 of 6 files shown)


vlib/v/checker/assign.v
```diff
@@ -560,6 +560,10 @@ fn (mut c Checker) assign_stmt(mut node ast.AssignStmt) {
 			}
 		}
 
+		if right_type == ast.nil_type && left_type.has_option_or_result() {
+			c.error('cannot assin `nil` to Option type (use `none` instead`)', node.pos)
```
spytheman, line 564:
```suggestion
			c.error('cannot assign `nil` to Option type (use `none` instead`)', node.pos)
```

spytheman: Although why?

if you have `field ?&Type`, doing `field: unsafe{ nil }` is a valid value imho

felipensp: For consistency ok, it makes sense to accept `nil`. Otherwise we have `none` and `nil` semantically equals to do same thing and casually making mistakes when mixing them.

spytheman: The `unsafe { nil }` is a valid value for the pointer, that later can be wrapped in an option value and stored in the field.
The `none` is a valid value of the option field itself.

PR #24451, opened 2025-05-10
slow_tests: fix good broken test

tankf33der:
A good and important test that didn't work at all and returned garbage.
I decided to fix it in this batch while minimally touching everything else.

Now everything is ok and the result will always start with a capital letter.

### DIFF
--- a/vlib/v/slow_tests/assembly/asm_test.amd64.v
+++ b/vlib/v/slow_tests/assembly/asm_test.amd64.v
@@ -100,7 +100,7 @@ fn test_inline_asm() {
 
 	assert util.add(8, 9, 34, 7) == 58 // test .amd64.v imported files
 
-	mut o := Manu{}
+	mut o := &Manu{}
 	asm amd64 {
 		mov eax, 0
 		cpuid
@@ -108,7 +108,7 @@ fn test_inline_asm() {
 		  =d (o.edx) as edx0
 		  =c (o.ecx) as ecx0
 	}
-	o.str()
+	assert o.str()[0].is_capital()
 }
 
 @[packed]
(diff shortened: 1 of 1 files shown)


vlib/v/slow_tests/assembly/asm_test.amd64.v
```diff
@@ -120,11 +120,11 @@ mut:
...
 
-fn (m Manu) str() string {
+fn (m &Manu) str() string {
 	return unsafe {
 		string{
-			str:    &u8(&m)
-			len:    24
+			str:    m
+			len:    12
 			is_lit: 1
```
spytheman, line 128: that is not valid

spytheman: hm, it is actually, my bad

tankf33der:
I see this warning from gcc and clang.

```
lambda:~/vc-ppc64le$ v -o cpuid.c cpuid.v && clang -g cpuid.c -lgc
cpuid.c:5941:26: warning: incompatible pointer types initializing 'u8 *' (aka 'unsigned char *') with an expression of type 'main__Manu *' (aka 'struct main__Manu *') [-Wincompatible-pointer-types]
 5941 |         return ((string){.str = m, .len = 12, .is_lit = 1});
      |                                 ^
1 warning generated.
```

spytheman:
> I see this warning from gcc and clang.
> 
> ```
> lambda:~/vc-ppc64le$ v -o cpuid.c cpuid.v && clang -g cpuid.c -lgc
> cpuid.c:5941:26: warning: incompatible pointer types initializing 'u8 *' (aka 'unsigned char *') with an expression of type 'main__Manu *' (aka 'struct main__Manu *') [-Wincompatible-pointer-types]
>  5941 |         return ((string){.str = m, .len = 12, .is_lit = 1});
>       |                                 ^
> 1 warning generated.
> ```


Try `str: voidptr(m)` .

tankf33der:
Both `voidptr(m)` and `&u8(m)` works.
I would prefer `&u8` form because of `string` type format:
```
pub struct string {
pub:
        str &u8 = 0
        len int
mut:
        is_lit int
}
```

What do you think?

spytheman:
Sure, `&u8(m)` is stricter.

PR #24452, opened 2025-05-10
Addresses #24291

einar-hjortdal:
Seems to solve the issue, can't find negatives.
I am not the author of BufferedReader, I do not know what the purpose of the check was.
Adding a test to prevent a regression would be nice, but I do not know what causes the issue in the first place.

### DIFF
--- a/vlib/io/buffered_reader.v
+++ b/vlib/io/buffered_reader.v
@@ -85,21 +85,23 @@ fn (mut r BufferedReader) fill_buffer() bool {
 	}
 	r.offset = 0
 	r.len = 0
-	r.len = r.reader.read(mut r.buf) or {
-		// end of stream was reached
-		r.end_of_stream = true
-		return false
-	}
-	if r.len == 0 {
-		r.fails++
-	} else {
-		r.fails = 0
-	}
-	if r.fails >= r.mfails {
-		// When reading 0 bytes several times in a row, assume the stream has ended.
-		// This prevents infinite loops ¯\_(ツ)_/¯ ...
-		r.end_of_stream = true
-		return false
+	for r.len == 0 {
+		r.len = r.reader.read(mut r.buf) or {
+			// end of stream was reached
+			r.end_of_stream = true
+			return false
+		}
+		if r.len == 0 {
+			r.fails++
+			if r.fails == r.mfails {
+				// When reading 0 bytes several times in a row, assume the stream has ended.
+				// This prevents infinite loops ¯\_(ツ)_/¯ ...
+				r.end_of_stream = true
+				return false
+			}
+		} else {
+			r.fails = 0
+		}
 	}
 	// we got some data
 	return true


spytheman:
> Adding a test to prevent a regression would be nice, but I do not know what causes the issue in the first place.

If you do not know what the cause of the issue is, and can not make a test, how are you sure, that you have fixed it, and that your changes would not regress or cause problems for others?

einar-hjortdal:
I have investigated this issue. The way I discovered it is by talking to a server that disconnects unexpectedly. When the server disconnects, you would expect Eof to be returned when attempting to read, instead I got Unexpected.
This is because instead of reaching Eof, the algorithm was just plain wrongly handling read attempts from the underlying TCPConn reader, returning Unexpected immediately when getting no data, instead of trying to get data a second time and reaching Eof.
Now I know that when I am getting Unexpected, the error is actually a Eof. I have reported the error and provided a fix, which admittedly looks rough, but seems to work correctly. Other than testing changes with a real service, I don't know how to write a test for it

spytheman:
I do not want to merge it without a test, since this part of the code is brittle and already had a period of instability, where different people wanted different behaviors.

See the existing `vlib/io/reader_test.v` as an example on how to create a struct instance of the interface, that you can later control in whatever ways you need to simulate the situation you were in, without having to re-create a real service.

einar-hjortdal:
> I do not want to merge it without a test, since this part of the code is brittle and already had a period of instability, where different people wanted different behaviors.

I agree, however, I do not know when I'll have time to write a test.
For now, a user can assume that `Unexpected` is actually `Eof`.

spytheman:
@einar-hjortdal do you have a way to test it now?

einar-hjortdal:
> @einar-hjortdal do you have a way to test it now?

No, I'm sorry, I do not have time to work on this.

spytheman:
> No, I'm sorry, I do not have time to work on this.

Thanks for the PR, but without a test, I will not merge it unfortunately, and since you do not have time, its status is unlikely to change.

Hopefully in the future, there will be an opportunity to find a more reproducible way to reveal the problem, if it exists, and add a fix + test for it.

I am closing it for now.

einar-hjortdal:
The problem does exist; it is clear if you read the source code. What matters is that users are informed that the `Unexpected` error is `Eof` and not something else.

spytheman:
I've read the source code and your patch, and I could not infer a test that would reproduce the problem that you saw.

To me, that means that the problem is not clear, or I would have added the test myself a long time ago, and this would not have waited in a limbo for 2 months.

-> I wanted a test from you, in order to prevent future regressions, as I already stated above. If you do not have time to make it, and if there is not some other developer that can, that understands it better than me, and has more time than you, I do think it is better, to have this closed, and wait until there is a better reproduction/fix.

einar-hjortdal:
Just to be clear, the issue is here:
[this line](https://github.com/vlang/v/blob/52ae3f2476b70e5590846b176fb176fe1475d13b/vlib/io/buffered_reader.v#L54) calls `fill_buffer`, which reads 0 bytes [here](https://github.com/vlang/v/blob/52ae3f2476b70e5590846b176fb176fe1475d13b/vlib/io/buffered_reader.v#L88) without returning any error, therefore function execution continues and the failure is marked at [this line](https://github.com/vlang/v/blob/52ae3f2476b70e5590846b176fb176fe1475d13b/vlib/io/buffered_reader.v#L94), but because there is no loop, the function [returns true wrongly assuming it got data](https://github.com/vlang/v/blob/52ae3f2476b70e5590846b176fb176fe1475d13b/vlib/io/buffered_reader.v#L104), and this causes the caller to [return the error](https://github.com/vlang/v/blob/52ae3f2476b70e5590846b176fb176fe1475d13b/vlib/io/buffered_reader.v#L60) instead of attempting another read and returning Eof

spytheman:
ok, hopefully that explanation, clears it up for the developer that will write the test, and we could then reopen the PR or make a new one.

PR #24456, opened 2025-05-10
cgen: fix assign from mut `for` variable to pointer

### LINKED ISSUE #24432: Reference assigned inside for loop
### Describe the bug

References assigned inside for loop does not work after the loop.

### Reproduction Steps

```
	mut rl := [&Demo{'A'}]

	println(rl)

	mut p := &Demo{}
	for mut e in rl {
		p = e	// does not work
	}

	println(p) // p is empty

	for i in 0 .. rl.len {
		mut e := rl[i]
		p = e	// work
	}

	println(p) // p is A
```

### Expected Behavior

I expect p to have the value it was assigned in the loop.
I expect the 2 loop type to behave identical.

### Current Behavior

in this loop type:
for mut e in rl { p=... }
the reference p does not work after the loop/

### Possible Solution

_No response_

_No response_

### V version

V 0.4.10 a1d9450

macos

felipensp:
Fix #24432

### DIFF
--- a/vlib/v/gen/c/assign.v
+++ b/vlib/v/gen/c/assign.v
@@ -969,6 +969,11 @@ fn (mut g Gen) assign_stmt(node_ ast.AssignStmt) {
 					defer {
 						g.is_option_auto_heap = old_is_auto_heap
 					}
+					if val is ast.Ident && val.is_mut() && var_type.is_ptr() {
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #24457, opened 2025-05-10
checker: fix generic option array arg passing to `[]T`

### LINKED ISSUE #24423: checker: wrong method check for array and array of options
### Describe the bug

`decode_array` think that `[]?int` is `[]int`

### Reproduction Steps

```v
struct Decoder { }

	mut decoder := Decoder{}
	mut result := T{}
	return result
}

		// checking wrongly. `decode_array` think that `[]?int` is `[]int`
		return
	} $else $if T.unaliased_typ is $struct {
		$for field in T.fields {
			decoder.decode_value(mut val.$(field.name))!
		}
	}
}

fn (mut decoder Decoder) decode_array[T](mut val []T) ! {}

struct Foo {
	int  []int
	oint []?int
}

fn main() {
	decode[Foo]('')!
}
```

### Expected Behavior

Not error

### Current Behavior

```sh
code.v:13:27: error: cannot use `&[]?int` as `&[]int` in argument 1 to `Decoder.decode_array`
   11 |     $if T.unaliased_typ is $array {
   12 |         // checking wrongly
   13 |         decoder.decode_array(mut val)!
      |                                 ~~~
   14 |         return
   15 |     } $else $if T.unaliased_typ is $struct {
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.10 d9afebc

Ubuntu

> [!NOTE]
>

felipensp:
Fix #24423

### DIFF
--- a/vlib/v/tests/generics/generic_array_pass_test.v
+++ b/vlib/v/tests/generics/generic_array_pass_test.v
@@ -1,0 +1,31 @@
+struct Decoder {}
+
+pub fn decode[T](val string) !T {
+	mut decoder := Decoder{}
+	mut result := T{}
+	decoder.decode_value(mut result)!
+	return result
+}
+
+fn (mut decoder Decoder) decode_value[T](mut val T) ! {
+	$if T.unaliased_typ is $array {
+		// checking wrongly. `decode_array` think that `[]?int` is `[]int`
+		decoder.decode_array(mut val)!
+		return
+	} $else $if T.unaliased_typ is $struct {
+		$for field in T.fields {
+			decoder.decode_value(mut val.$(field.name))!
+		}
+	}
+}
+
+fn (mut decoder Decoder) decode_array[T](mut val []T) ! {}
+
+struct Foo {
+	int  []int
+	oint []?int
+}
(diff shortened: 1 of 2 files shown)


eyestorefivemstore:
Great implementation of this feature. I particularly like the approach to error handling.

eyestoreoye:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

vlib/v/tests/generics/generic_array_pass_test.v
```diff
@@ -0,0 +1,31 @@
...
+pub fn decode[T](val string) !T {
+	mut decoder := Decoder{}
+	mut result := T{}
+	decoder.decode_value(mut result)!
+	return result
+}
+
+fn (mut decoder Decoder) decode_value[T](mut val T) ! {
+	$if T.unaliased_typ is $array {
+		// checking wrongly. `decode_array` think that `[]?int` is `[]int`
```
spytheman, line 12: if the PR fixes it, please remove the comment - it will be confusing after it is merged, if it will be working

PR #24463, opened 2025-05-12
term: add writeln_color()

kbkpbot:
The PR add `write_color()` and `writeln_color()` to `term` module, enable output ANSI colorful string.

example:

```v
module main
import term
import strings

fn main() {
        mut sb := strings.new_builder(100)
        term.write_color(mut sb, '[Warning]', fg : .red)
        sb.write_string('Please make sure ')
        term.write_color(mut sb, 'value > 0', fg: .cyan)
        sb.write_string(' and ')
        term.write_color(mut sb, 'format is string', fg : .cyan)
        sb.writeln('')

        output := sb.str()
        println(output)
}
```

will get output:
image


Delta456:
But why? We already have `term` module for that

spytheman:
It makes no logical sense for a specif formatting function (using ANSI escape codes) to be part of the string builder.

spytheman:
> And I prefer build the string in string builder, not just output it directly.

I agree - that is why the first argument is a mutable string builder.

That:
```v
term.writeln(mut sb, 'hello', styles :[.bold, .italic, .underline], fg: .red, bg : .cyan)
```

is a direct equivalent to your:
```v
sb.writeln_color('hello', styles :[.bold, .italic, .underline], fg: .red, bg : .cyan)
```

It will not print anything on its own to stdout, it will just append to the string builder like yours does.

spytheman:

medvednikov:
I agree with spy here

spytheman (approved):

PR #24467, opened 2025-05-12
vtest: remove ts.vroot only from beginning of the path (fix #24363)

### LINKED ISSUE #24363: vtest cannot work from /
how to repeat:
```
cd /
git clone https://github.com/vlang/v
cd /v
make
./v -version
./v test vlib/v/tests
# all tests will fail
```

Example of error:
```
 FAIL  [  69/1687] C:   172.9 ms, R:     0.000 ms vlibtests/assign/assign_bitops_with_type_aliases_test.v
>> compilation failed:
builder error: vlibtests/assign/assign_bitops_with_type_aliases_test.v does not exist
```

Path of file lost `v` between `vlib` and `tests`.

tankf33der:
To transform an absolute path into a relative one, you need to remove the `root` only at the beginning of the path, not everywhere.
I'm curious how this will work in `CI`.

### DIFF
--- a/cmd/tools/modules/testing/common.v
+++ b/cmd/tools/modules/testing/common.v
@@ -417,7 +417,7 @@ fn worker_trunner(mut p pool.PoolProcessor, idx int, thread_id int) voidptr {
 		cmd_options << ' -enable-globals'
 	}
 	if ts.root_relative {
-		relative_file = relative_file.replace(ts.vroot + os.path_separator, '')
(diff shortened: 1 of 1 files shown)


spytheman:
Afaik the relative path is only used for displaying the results.

spytheman (approved):
Excellent work.
Thank you 🙇🏻‍♂️.

PR #24468, opened 2025-05-12
term/ui: fix fatal from missing cursor pos resp

tauraamui:
There are some cases I've encountered when running a V terminal/TUI application where the host does not respond with actual cursor position data when requested. If this occurs we get an OOB panic, even if the TUI application doesn't need or care what the current cursor position is.

This very simple PR just adds an additional check to prevent an OOB based panic/fatal crash and also includes a test to demonstrate that this is working as expected.

### DIFF
--- a/vlib/term/ui/termios_nix_test.v
+++ b/vlib/term/ui/termios_nix_test.v
@@ -1,0 +1,65 @@
+module ui
+
+fn test_get_cursor_position_reads_valid_row_column_data() ! {
+	mut pipeset := [0, 0]
+	mut original_stdin_fd := -1
+	unsafe {
+		if C.pipe(&pipeset[0]) == -1 {
+			return error("unable to create pipe: ${C.strerror(C.errno)}")
+		}
+
+		fake_cursor_pos_data := '\033[45;70R'
+		written_bytes := C.write(pipeset[1], fake_cursor_pos_data.str, fake_cursor_pos_data.len)
+		if written_bytes == -1 {
+			C.close(pipeset[0])
+			C.close(pipeset[1])
+			return error("error writing into pipe: ${C.strerror(C.errno)}")
+		}
+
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.
Thank you 🙇🏻‍♂️.

vlib/term/ui/termios_nix_test.v
```diff
@@ -0,0 +1,63 @@
+module ui
+
+fn test_get_cursor_position_reads_valid_row_column_data() ! {
```
Delta456, line 3: Why have a Result value here when no function has a Result value?

vlib/term/ui/termios_nix_test.v
```diff
@@ -0,0 +1,63 @@
...
+
+		C.close(pipeset[0])
+
+		cursor_pos_x, cursor_pos_y := get_cursor_position()
+		assert cursor_pos_x == 45
+		assert cursor_pos_y == 70
+	}
+}
+
+fn test_get_cursor_position_reads_empty_position_data() ! {
```
Delta456, line 34: Same here

vlib/term/ui/termios_nix_test.v
tauraamui: Error is a result, what do you mean?

Delta456: Oh right. I didn't look :sweat:

tauraamui: No worries! ☺️

vlib/term/ui/termios_nix_test.v
tauraamui: Again, error response here ☺️

tauraamui:
> Excellent work. Thank you 🙇🏻‍♂️.

The honour is all mine! 🫡

Issue #23924, opened 2025-03-13
`make` stuck  under ARM/Linux


kbkpbot:
### Describe the bug

When try to `make` create a `v` under ARM Linux, it will stuck at `v2` stage.

### Reproduction Steps

```sh
make
```

### Expected Behavior

create a new `v`.

### Current Behavior

make stuck at `v2` stage.

```sh
$ make
cd ./vc && git clean -xf && git pull --quiet
remote: Enumerating objects: 2, done.
remote: Counting objects: 100% (2/2), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 2 (delta 1), reused 0 (delta 0), pack-reused 0 (from 0)
接收对象中: 100% (2/2), 1.04 MiB | 360.00 KiB/s, 完成.
处理 delta 中: 100% (1/1), 完成.
正在更新文件: 100% (2/2), 完成.
cd ./thirdparty/tcc && git clean -xf && git pull --quiet
cc  -std=gnu99 -w -o v1 ./vc/v.c -lm -lpthread  || cmd/tools/cc_compilation_failed_non_windows.sh
./v1 -no-parallel -o v2  cmd/v
./v2 -nocache -o ./v  cmd/v

```

### Possible Solution

By editing the `GNUMakefile`, add a `-no-parallel` flag to the `v2` command, it will create a new `v` successful.

```sh
$ make
cd ./vc && git clean -xf && git pull --quiet
cd ./thirdparty/tcc && git clean -xf && git pull --quiet
cc  -std=gnu99 -w -o v1 ./vc/v.c -lm -lpthread  || cmd/tools/cc_compilation_failed_non_windows.sh
./v1 -no-parallel -o v2  cmd/v
./v2 -nocache -no-parallel -o ./v  cmd/v
rm -rf v1 v2
```

But `v` repl can't start...

### Additional Information/Context

It may related to the parallel compiling under ARM/Linux

### V version

V 0.4.9 89d1aac

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.9 a694918.89d1aac
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 20.04.6 LTS
|Processor           |4 cpus, 64bit, little endian, aarch64
|Memory              |0.61GB/3.78GB
|                    |
|V executable        |/home/vpn/github/lang/v/v_old
|V last modified time|2025-02-23 11:35:00
|                    |
|V home dir          |OK, value: /home/vpn/github/lang/v
[...]

PR #24472, opened 2025-05-13
sync.stdatomic: workaround libatomic.a indirect symbols tcc bug

kbkpbot:
Fix issue #23924

This is due to a `tcc` bug handling indirect(IFUNC) symbols in `.a`.
I think it should generate a `GOT/PLT` in output elf file. 
It can't determine the symbol's address at compile time.
https://lists.nongnu.org/archive/html/tinycc-devel/2025-05/msg00010.html


spytheman (approved):
Excellent work.

PR #24476, opened 2025-05-13
math.big: new function com() + tests

tankf33der:
"Bitwise complement" is like `~` but for math.big.
Complete test coverage of edge cases + tested against `mpz_com()` from `libgmp` on 1B loop.


spytheman (approved):
Excellent work.

Issue #24482, opened 2025-05-14
[linux/aarch64] spawn not work when compile with tcc


kbkpbot:
### Describe the bug

a simple program , use spawn , can not work when compile with `tcc`.
when compile with `clang`, it can work.

### Reproduction Steps

```
v vlib/v/tests/fns/go_call_fn_return_test.v

```

This is `vlib/v/tests/fns/go_call_fn_return_test.v`
```v
const text = 'Hello world'

fn test_go_call_fn_return() {
        hex_go := spawn text.hex()
        hex := text.hex()

        assert hex == '48656c6c6f20776f726c64'
        assert hex_go.wait() == hex
}

```

### Expected Behavior

pass test

### Current Behavior

```
$ v vlib/v/tests/fns/go_call_fn_return_test.v
/tmp/v_1001/go_call_fn_return_test.01JV7CZEQ8C0WG8TR8YRABMX8Y.tmp.c:3017: at string_hex_thread_wrapper: RUNTIME ERROR: invalid memory access
0043fdd8 : by ???
00434b20 : by ???
ffffba684624 : by ???
ffffba77f66c : by ???
Terminated by signal 11 (SIGSEGV)

```

### Possible Solution

Fix `spawn_and_go.v` to ensure it can work with tcc/aarch64.

### Additional Information/Context

```c
void* string_hex_thread_wrapper(thread_arg_string_hex *arg) {
	string* ret_ptr = (string*) _v_malloc(sizeof(string));
	*ret_ptr = arg->fn(arg->arg0);                 // crash here
	_v_free(arg);
	return ret_ptr;
}
```

And can be workaround by following code:
```c
void* string_hex_thread_wrapper(thread_arg_string_hex *arg) {
	string* ret_ptr = (string*) _v_malloc(sizeof(string));
	string _t1 = arg->fn(arg->arg0);
	*ret_ptr = _t1; 
	_v_free(arg);
	return ret_ptr;
}
```

### V version

V 0.4.10 b9fe26cbf18c746256414ae7a6d09cd80b4fa61d.b9fe26c

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 b9fe26cbf18c746256414ae7a6d09cd80b4fa61d.b9fe26c
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 20.04.6 LTS
|Processor           |4 cpus, 64bit, little endian, aarch64
|Memory              |0.8GB/3.78GB
|                    |
[...]

PR #24483, opened 2025-05-14
cgen: workaround tcc aarch64 spawn call bug

kbkpbot:
Fix issue #24482

This PR introduce a tmp var, a workaround for the `tcc/aarch64` bug.

before :

```c
void* string_hex_thread_wrapper(thread_arg_string_hex *arg) {
	string* ret_ptr = (string*) _v_malloc(sizeof(string));
	*ret_ptr = arg->fn(arg->arg0); 
	_v_free(arg);
	return ret_ptr;
}
```

after:

```c
void* string_hex_thread_wrapper(thread_arg_string_hex *arg) {
        string* ret_ptr = (string*) _v_malloc(sizeof(string));
        string _t3 = arg->fn(arg->arg0);
        *ret_ptr = _t3;
        _v_free(arg);
        return ret_ptr;
}
```

### DIFF
--- a/vlib/v/gen/c/spawn_and_go.v
+++ b/vlib/v/gen/c/spawn_and_go.v
@@ -336,10 +336,8 @@ fn (mut g Gen) spawn_and_go_expr(node ast.SpawnExpr, mode SpawnGoMode) {
 			}
 		}
 		g.gowrappers.writeln(');')
-		if call_ret_type != ast.void_type {
-			if g.pref.os != .windows {
-				g.gowrappers.write_string('\t*ret_ptr = ${tcc_bug_tmp_var};\n')
-			}
+		if g.pref.os != .windows && call_ret_type != ast.void_type {
+			g.gowrappers.writeln('\t*ret_ptr = ${tcc_bug_tmp_var};')
 		}
 		if is_spawn {
 			g.gowrappers.writeln('\t_v_free(arg);')
--- a/vlib/v/gen/c/spawn_and_go.v
+++ b/vlib/v/gen/c/spawn_and_go.v
@@ -241,12 +241,13 @@ fn (mut g Gen) spawn_and_go_expr(node ast.SpawnExpr, mode SpawnGoMode) {
 		thread_ret_type := if g.pref.os == .windows { 'u32' } else { 'void*' }
 		g.waiter_fn_definitions.writeln('${g.static_non_parallel}${thread_ret_type} ${wrapper_fn_name}(${wrapper_struct_name} *arg);')
 		g.gowrappers.writeln('${thread_ret_type} ${wrapper_fn_name}(${wrapper_struct_name} *arg) {')
+		tcc_bug_tmp_var := g.new_tmp_var()
 		if call_ret_type != ast.void_type {
 			if g.pref.os == .windows {
 				g.gowrappers.write_string('\t*((${s_ret_typ}*)(arg->ret_ptr)) = ')
 			} else {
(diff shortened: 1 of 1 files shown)


vlib/v/gen/c/spawn_and_go.v
```diff
@@ -241,12 +241,13 @@ fn (mut g Gen) spawn_and_go_expr(node ast.SpawnExpr, mode SpawnGoMode) {
...
 		g.waiter_fn_definitions.writeln('${g.static_non_parallel}${thread_ret_type} ${wrapper_fn_name}(${wrapper_struct_name} *arg);')
 		g.gowrappers.writeln('${thread_ret_type} ${wrapper_fn_name}(${wrapper_struct_name} *arg) {')
+		tcc_bug_tmp_var := g.new_tmp_var()
 		if call_ret_type != ast.void_type {
 			if g.pref.os == .windows {
 				g.gowrappers.write_string('\t*((${s_ret_typ}*)(arg->ret_ptr)) = ')
 			} else {
 				g.gowrappers.writeln('\t${s_ret_typ}* ret_ptr = (${s_ret_typ}*) _v_malloc(sizeof(${s_ret_typ}));')
-				g.gowrappers.write_string('\t*ret_ptr = ')
+				g.gowrappers.write_string('\t${s_ret_typ} ${tcc_bug_tmp_var} = ')
```
JalonSolov, line 250: Perhaps wrap this in `$if tinyc {} else {}`, so the extra variable is only generated if tcc is used?

kbkpbot: This can't work?

```v
module main

fn main() {
        $if tinyc {
                mut var := '1'
        }


        $if tinyc {
                var = '2'
        }

        println(var)
}

```


```
$ v t.v
t.v:5:21: warning: unused variable: `var`
    3 | fn main() {
    4 |         $if tinyc {
    5 |                 mut var := '1'
      |                     ~~~
    6 |         }
    7 |
t.v:10:17: error: undefined ident: `var` (use `:=` to declare a variable)
    8 |
    9 |         $if tinyc {
   10 |                 var = '2'
      |                 ~~~
   11 |         }
   12 |
t.v:10:23: error: cannot assign to `var`: expected `void`, not `string`
    8 |
    9 |         $if tinyc {
   10 |                 var = '2'
      |                       ~~~
   11 |         }
   12 |
t.v:13:17: error: undefined ident: `var`
   11 |         }
   12 |
   13 |         println(var)
      |                 ~~~
   14 | }
   15 |
t.v:13:9: error: `println` can not print void expressions
   11 |         }
   12 |
   13 |         println(var)
      |         ~~~~~~~~~~~~
   14 | }
   15 |

```

JalonSolov: No, because it is 2 separate contexts.  You would have to declare it an use it in the same context.

It's not that big a problem if you can't get it to work.  I just thought since it only seems to be a problem with a specific compiler on one specific CPU, it would be good to isolate the extra code to only be produced in that situation.

However, since it's only a single temp var that's not used in a time-critical loop (hopefully), then it won't be a big problem for other systems.

spytheman: It will be harder to test the output, if it is platform/compiler dependent. I think that currently `vlib/v/gen/c/coutput_test.v` fails because of that (the patterns in `vlib/v/gen/c/testdata/spawn_stack_nix.c.must_have` will have to change).

Delta456 (approved):

PR #24484, opened 2025-05-14
os: use XDG tmpdir before the POSIX env if defined

trufae:
some tests fail because make wrong assumptions like considering the temporary directories are one path above the root.

```
   > assert document.get_parent_mod(os.temp_dir())? == ''
     Left value (len: 4): `user`
    Right value (len: 0): ``
```

### DIFF
--- a/vlib/os/os.v
+++ b/vlib/os/os.v
@@ -943,7 +943,10 @@ pub fn local_bin_dir() string {
 
 // temp_dir returns the path to a folder, that is suitable for storing temporary files.
 pub fn temp_dir() string {
-	mut path := getenv('TMPDIR')
+	mut path := getenv('XDG_RUNTIME_DIR')
+	if path == '' {
+		path = getenv('TMPDIR')
+	}
 	$if windows {
 		if path == '' {
 			// TODO: see Qt's implementation?


spytheman:
Please fix the tests too.

trufae:
I didnt moved the pr forward because im afk until next week but i have some questions before fixing the tests:

- should we priorize TMPDIR or XDG if both are set?
- The testsuite may probably define its own tmpdir env values instead of assuming all the systems provides the same

because in the case of xdg, each user have its own tmpdir in the home, so running the testsuite will end up into different paths depending on the uid or even if the system reboots may use a different tmpdir for the cache.

spytheman:
Reading https://wiki.archlinux.org/title/XDG_Base_Directory and https://specifications.freedesktop.org/basedir-spec/latest/ : 
> $XDG_RUNTIME_DIR defines the base directory relative to which user-specific non-essential runtime files and other file objects (such as sockets, named pipes, ...) should be stored. The directory MUST be owned by the user, and they MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700. 

The spec does not mention temporary files, nor any relationships with TMPDIR (which afaik is defined on macos).

Further, looking at other language implementations, I could not find a precedent that links the two.

Go for example, has: https://github.com/golang/go/blob/master/src/os/file.go#L481
> // TempDir returns the default directory to use for temporary files.
//
// On Unix systems, it returns $TMPDIR if non-empty, else /tmp.
// On Windows, it uses GetTempPath, returning the first non-empty
// value from %TMP%, %TEMP%, %USERPROFILE%, or the Windows directory.
// On Plan 9, it returns /tmp.

For another example, Qt also uses TMPDIR, but does not mention XDG_RUNTIME_DIR in https://github.com/qt/qtbase/blob/dev/src/corelib/io/qfilesystemengine_unix.cpp#L1861 .

Given those, and absent a good motivation about the change, and given the failing tests, I do not see a good way for this PR to proceed :-| .

trufae:
Considering systemd systems don’t have tmpdir env var defined, maybe it makes sense to decouple that functionality into an xdg module that exposes the needed directories for that purpose.

I understand the behaviour should be aligned with Go or other frameworks. So i’m also fine with that. The standard doesnt specify one should override the other and i understand non xdg apps wont be assuming the tmpdir is user-specific, with the security implications that this can have.

PR #24487, opened 2025-05-14
math.big: fix sign machinarium in .+/2 function, add test

tankf33der:
The `+` function has incorrect `sign` management and cannot pass the new test.
The new version also passes the 1M loop test against `libgmp`.


spytheman (approved):
Excellent work.

PR #24489, opened 2025-05-15
math.big: reminder must respect the sign of dividend in % operator, a…

tankf33der:
`%` operator of `math.big` does not respect sign of `dividend`.

Just copy-paste from `/` operator for `sign` management.
Now it can pass new test and 1M loop against `libgmp`.


spytheman (approved):
Excellent work.

PR #24494, opened 2025-05-15
checker, cgen: fix generic type resolving when unwrapped is an option

### LINKED ISSUE #24466: cgen, generics: C gen error when using optional element in array
### Describe the bug

`[]T` unwrap option elements, which will not work in real world

### Reproduction Steps

```v
struct Decoder { }

	mut decoder := Decoder{}
	mut result := T{}
	return result
}

		return
	} $else $if T is $option {
		println('passed here 2')
		val = none
	} 
}

fn (mut decoder Decoder) decode_array[T](mut val []T) ! {
	$if T is $option {
		println('passed here')
		val << none
		val << 1
	} $else {
		val << 1
	}
	
	mut array_element := T{}
	decoder.decode_value(mut array_element)!
	val << array_element
}

fn main() {
	assert decode[[]int]('')! == [1]
	assert decode[[]?int]('')! == [?int(none), 1, none]
}
```

### Expected Behavior

assert pass

### Current Behavior

```sh
/tmp/v_60000/code.01JV23N10FWK0SD8320EBNH30R.tmp.c:1816: error: invalid operand types for binary operation
builder error: 
==================
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.10 e93c344

Ubuntu

felipensp:
Fix #24466

### DIFF
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -1724,7 +1724,12 @@ pub fn (mut t Table) convert_generic_type(generic_type Type, generic_names []str
 		if typ.has_flag(.generic) {
 			return typ.derive_add_muls(generic_type).set_flag(.generic)
 		} else {
-			return typ.derive_add_muls(generic_type).clear_flag(.generic)
+			if !typ.has_flag(.option) {
+				return typ.derive_add_muls(generic_type).clear_flag(.generic)
+			} else {
+				// Option concrete type where T is expected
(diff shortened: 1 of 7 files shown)


felipensp:
We have to fix some issues separately before fixing this one. Closing for now.

enghitalo:
> We have to fix some issues separately before fixing this one. Closing for now.

Sorry to bother you, can you list what needs to be fixed first?

PR #24495, opened 2025-05-15
v.util: use internal diff

kbkpbot:
Use internal `arrays.diff` instead of external `diff` tool.
This make `diff` platform independent.


spytheman (approved):
Excellent work.

PR #24513, opened 2025-05-18
tools:vup_println_latest_commit_time_as_local

get200:
Convert commit time to local time.

## 1.Before
Updating V...
> git_command: git pull https://github.com/vlang/v master
V is already updated.
Current V version: V 0.4.10 f533b5e, timestamp: **2025-05-18 16:36:41 +0300**

## 2.After
Updating V...
> git_command: git pull https://github.com/vlang/v master
V is already updated.
Current V version: V 0.4.10 f533b5e, timestamp: **2025-05-18 21:36:41 +0800**

### DIFF
--- a/cmd/tools/vup.v
+++ b/cmd/tools/vup.v
@@ -3,6 +3,7 @@ module main
 import os
 import v.util.version
 import v.util.recompilation
+import time
 
 const vexe = os.real_path(os.getenv_opt('VEXE') or { @VEXE })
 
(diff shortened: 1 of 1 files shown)


enghitalo:
In my opinion, the time has to be UTC

enghitalo:
In my opinion, the time has to be UTC

spytheman:
The shown timestamp timezone is not essential to the works of `v up`, while it adds complexity to the tool, that should be simple (to reduce the failure/edge cases).

Thanks for the PR, but I will not merge it.

PR #24516, opened 2025-05-18
net.http: remove hardcoded read_timeout

martinskou:
net.http: you can now set read_timeout on a http request.

### DIFF
--- a/vlib/net/http/http.v
+++ b/vlib/net/http/http.v
@@ -38,9 +38,9 @@ pub mut:
 	on_progress_body RequestProgressBodyFn = unsafe { nil }
 	on_finish        RequestFinishFn       = unsafe { nil }
 
-	stop_copying_limit   i64 = -1 // after this many bytes are received, stop copying to the response. Note that on_progress and on_progress_body callbacks, will continue to fire normally, until the full response is read, which allows you to implement streaming downloads, without keeping the whole big response in memory
-	stop_receiving_limit i64 = -1 // after this many bytes are received, break out of the loop that reads the response, effectively stopping the request early. No more on_progress callbacks will be fired. The on_finish callback will fire.
-	read_timeout         i64 = 1000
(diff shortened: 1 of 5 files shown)


vlib/net/http/http.v
```diff
@@ -40,6 +40,7 @@ pub mut:
 
 	stop_copying_limit   i64 = -1 // after this many bytes are received, stop copying to the response. Note that on_progress and on_progress_body callbacks, will continue to fire normally, until the full response is read, which allows you to implement streaming downloads, without keeping the whole big response in memory
 	stop_receiving_limit i64 = -1 // after this many bytes are received, break out of the loop that reads the response, effectively stopping the request early. No more on_progress callbacks will be fired. The on_finish callback will fire.
+	read_timeout         i64 = 1000
```
spytheman, line 43: 1000 of what? ms, ns, seconds ?

spytheman: keep in mind that this is a public config struct, and it is preferable that all the fields are commented or named in such a way, that the possible values and measurement units are clear

martinskou: yes, i'll update and make a new submission.

vlib/net/mbedtls/ssl_connection.c.v
```diff
@@ -135,6 +133,7 @@ mut:
 	opened    bool
 	// handle		int
 	// duration	time.Duration
+	read_timeout i64
```
spytheman, line 136: please add a comment on the unit, or rename the field so that it is clear what the possible values are

vlib/net/openssl/ssl_connection.c.v
```diff
@@ -15,7 +15,8 @@ pub mut:
 	handle   int
 	duration time.Duration
 
-	owns_socket bool
+	owns_socket  bool
+	read_timeout i64
```
spytheman, line 19: same - please rename the field, or add a comment on the measurement unit

vlib/net/http/http.v
spytheman: there is no need for a new PR, just make the change and push the new commit, it will update the PR automatically

Issue #23966, opened 2025-03-17
V search tool does not find modules


jorgeluismireles:
### Describe the bug

Seems `v search` does not find modules installed or not.

### Reproduction Steps

Trying to list known modules like `ui` or `markdown` already installed.
```
$ v search help
Searches https://vpm.vlang.io/ for matching KEYWORDs and displays the details

Usage:
  v search KEYWORD[...]

Options:
  -v           Print more details about the performed operation.
  -server-url  When doing network operations, use this vpm server. Can be
               given multiple times.
$
$ v search -v
error: specify at least one keyword to search for.
$ v search -v ui
Trying server url: https://vpm.vlang.io
No module(s) found for `ui` .
$
$ v search -v markdown
Trying server url: https://vpm.vlang.io
No module(s) found for `markdown` .
$
$ v search -server-url http://github.com/vlang
No module(s) found for `http://github.com/vlang` .
$ 
```

### Expected Behavior

Display modules

### Current Behavior

No module(s) found for `ui` .

### Additional Information/Context

After reading: https://github.com/vlang/v/discussions/23939

### V version

V 0.4.9 d970a8f

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.9 8683e63.d970a8f
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 22.04.5 LTS
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz
|Memory              |0.78GB/3.7GB
|                    |
|V executable        |/home/jorge/v/v
|V last modified time|2025-03-12 16:40:34
|                    |
|V home dir          |OK, value: /home/jorge/v
|VMODULES            |OK, value: /home/jorge/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/jorge
|                    |
|Git version         |git version 2.34.1
|V git status        |weekly.2025.08-149-gd970a8fc-dirty (39 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
|gcc version         |gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
|clang version       |Ubuntu clang version 14.0.0-1ubuntu1.1
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
[...]

PR #24535, opened 2025-05-21
vpm: make search work again

kbkpbot:
Fix issue #23966

### DIFF
--- a/cmd/tools/vpm/search_test.v
+++ b/cmd/tools/vpm/search_test.v
@@ -1,0 +1,20 @@
+// vtest retry: 3
+module main
+
+import test_utils { cmd_ok }
+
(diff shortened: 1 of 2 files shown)


cmd/tools/vpm/search_test.v
```diff
@@ -0,0 +1,20 @@
...
+
+fn testsuite_begin() {
+	$if !network ? {
+		eprintln('> skipping ${@FILE}, when `-d network` is missing')
+		exit(0)
+	}
+}
+
+fn testsuite_end() {
+}
```
spytheman, line 14: there is no need for an empty function; it is fine if it is not defined

spytheman (approved):

PR #24541, opened 2025-05-21
math.big: fix Karatsuba's add_in_place() function, add carry handler on exit

tankf33der:
Found via `pow()` loop vs. `gmp`.
Without this patch, these two tests will be unsuccessful.
The 100M loop vs. `gmp` will be after another patch because numbers have already been `found` where Karatsuba is unsuccessful.


spytheman (approved):
Excellent work.

spytheman:
The CI failures are not related, and should be fixed on master.

Issue #24552, opened 2025-05-23
error: cannot import sync into a module with the same name


kbkpbot:
### Describe the bug

When developing a `sync` condition module, I found there is a strange bug. It may related to `chan` in `sync` module.

### Reproduction Steps

create a file, named `condition.v` and place it in `vlib/sync/`

condition.v
```v
module sync

@[heap]
pub struct Cond {
mut:
    mu      &sync.Mutex
    waiters [] chan bool             // comment out this line will pass the test
}
```

then run a test:

```
v test vlib/sync/once_test.v
```

### Expected Behavior

pass test

### Current Behavior

```
$ v test vlib/sync/once_test.v
---- Testing... --------------------------------------------------------------------------
 FAIL     0.000 ms vlib/sync/once_test.v
>> compilation failed:
vlib/sync/condition.v:1:1: error: cannot import `sync` into a module with the same name
    1 | module sync
      | ^
    2 |
    3 | @[heap]
vlib/sync/condition.v:1:1: error: cannot import `sync` as `sync` into a module with the same name
    1 | module sync
      | ^
    2 |
    3 | @[heap]

------------------------------------------------------------------------------------------
To reproduce just failure 1 run:    '/media/HD/github/kbkpbot/v/v'  '/media/HD/github/kbkpbot/v/vlib/sync/once_test.v'
Summary for all V _test.v files: 1 failed, 1 total. Elapsed time: 698 ms, on 1 job. Comptime: 634 ms. Runtime: 0 ms.

```

### V version

V 0.4.10 be49e2bf6c9c61b7707e2f1e487b3714c8e1ba89

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 be49e2bf6c9c61b7707e2f1e487b3714c8e1ba89
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.2 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |9.21GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-05-21 13:04:19
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /media/HD/github/kbkpbot/v
|                    |
[...]

PR #24559, opened 2025-05-23
parser: fix duplicate mod imports

kbkpbot:
Fix issue #24552

This bug is because when code has `chan` in it, `parser` will `register_auto_import('sync')`, and then `checker` will fail with error.


spytheman (approved):
Excellent work.

PR #24561, opened 2025-05-24
sync.stdatomic: add atomic types

kbkpbot:
This PR create atomic types and add/sub load/store operation on them.


vlib/sync/stdatomic/atomic_test.v
```diff
@@ -93,3 +93,99 @@ fn count_one_cycle_without_sync(mut counter Counter, mut group sync.WaitGroup) {
 	}
 	group.done()
 }
+
+fn test_atomic_vals() {
+	mut v_bool := stdatomic.new_atomic(bool(false))
```
Delta456, line 98: No need to explicitly cast `bool` here as `false` is already a boolean value.

PR #24563, opened 2025-05-24
math.big: rewrite subtract_in_place() to fix Karatsuba, add tests

tankf33der:
`Karatsuba` can not pass the tests without the fix.
Values found via multi-million loop vs. `gmp`.

`Highlight` of the patch:

- add `inline` like in `add_in_place()` function
- to avoid introducing new bugs, the code for the `subtract_in_place()` function is borrowed from the `subtract_digit_array()` function from the `array_ops.v` file
- now `Karatsuba` passed 20M loop vs. `gmp`


spytheman:
Excellent work.

PR #24566, opened 2025-05-24
math.big: optimize divide_array_by_digit()

tankf33der:
Capacity for `quotient` array already defined in `integer.v` like this:

```
...
       // Division for positive integers
        mut q := []u32{cap: int_max(1, dividend.digits.len - divisor.digits.len + 1)}
        mut r := []u32{cap: dividend.digits.len}
        divide_digit_array(dividend.digits, divisor.digits, mut q, mut r)
...
```

so just use it instead of full `operand_a`.
One of the important clients of this function is `str()` function.
Everyone will feel the difference.

Performance improvements:

- `"./v test vlib/math/big/"` works `~3+` times faster
- `str()` on ~180k digits works `~6` times faster

### DIFF
--- a/vlib/math/big/array_ops.v
+++ b/vlib/math/big/array_ops.v
@@ -228,19 +228,18 @@ fn divide_array_by_digit(operand_a []u32, divisor u32, mut quotient []u32, mut r
 	}
 	// Dividend has more digits
 	mut rem := u64(0)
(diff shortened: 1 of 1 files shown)


enghitalo:
Good catch!

spytheman:
Excellent.

PR #24573, opened 2025-05-25
sync.stdatomic: new_atomic[T]() turn panic into compile_error

kbkpbot:
This PR help find out `new_atomic[T]()` type error at compile time, not in run time.

### DIFF
--- a/vlib/v/checker/tests/sync_stdatomic_compile_err.out
+++ b/vlib/v/checker/tests/sync_stdatomic_compile_err.out
@@ -1,0 +1,7 @@
+vlib/sync/stdatomic/atomic.c.v:83:3: error: atomic: only support number and bool types
+   81 |         }
+   82 |     } $else {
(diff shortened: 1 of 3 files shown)


vlib/v/checker/tests/sync_stdatomic_compile_err.out
Delta456, line 1: This shouldn't be shown as the code doesn't match the file code.

kbkpbot: This is because `$compile_error()` work like this. It can only locate in file where it is called.

Delta456: Let's make a separate issue to track this. 

We somehow need to show the code which causes the compile error.

kbkpbot: post at issue #24575

spytheman (approved):
Good work.

PR #24574, opened 2025-05-25
sync: add condition support

kbkpbot:
Just like `go.Cond`.

### DIFF
--- a/vlib/sync/cond.v
+++ b/vlib/sync/cond.v
@@ -27,6 +27,7 @@ pub fn new_cond(m &Mutex) &Cond {
 //     cond.wait()
 // }
 // mutex.unlock()
+@[direct_array_access]
 pub fn (mut c Cond) wait() {
 	// Create a channel for this waiting operation with capacity 1
 	ch := chan bool{cap: 1}
@@ -56,6 +57,7 @@ pub fn (mut c Cond) wait() {
 }
 
 // signal wakes one waiting thread
+@[direct_array_access]
 pub fn (mut c Cond) signal() {
 	c.inner_mutex.lock()
 	defer { c.inner_mutex.unlock() }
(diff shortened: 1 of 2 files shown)


Delta456 (approved):

vlib/sync/cond_test.v
```diff
@@ -0,0 +1,204 @@
...
+		}
+
+		for !done.load() {
+			cond.wait() // Wait for signal
+			wake_count.add(1)
+		}
+	}()
+
+	// Ensure waiter has entered waiting state
+	time.sleep(100 * time.millisecond)
```
spytheman, line 25: This seems a bit brittle (the CI could be slow sometimes).
Could you instead, pass another channel to the spawned function, write to it inside it, and block on reading from it here?

vlib/sync/cond.v
```diff
@@ -0,0 +1,127 @@
...
+	return &Cond{
+		mutex:       m
+		inner_mutex: new_mutex()
+		waiters:     []chan bool{}
+	}
+}
+
+// wait_for waits for the condition to be signaled with a `timeout`,
+// returning true if notified before the timeout expires, or false otherwise
+pub fn (mut c Cond) wait_for(timeout time.Duration) bool {
```
spytheman, line 27: Go's `sync.Cond` does not expose a `wait_for` method (looking at https://pkg.go.dev/sync#Cond ).

Looking at the implementation and the test, it seems complex 🤔.
Is it really needed?

vlib/sync/cond_test.v
kbkpbot: OK, I will fix it

vlib/sync/cond.v
kbkpbot: OK, I will remove it

Issue #24172, opened 2025-04-09
native: builtin module tasks


Eliyaan:
This is the parent issue to keep track of what needs to be done for native. (tasks taken from #17782 )
- [ ] library linking support for MacOS
- [ ] global variables
- [ ] rebuild codegen of `ast.AssignStmt` to accept all expressions
- [ ] missing expressions #24170 
- [ ] replace current inline implementations `println()`, `assert()`, `exit()`, `C.syscall()`, etc. by their `builtin` counterparts
- [ ] support floats (asserts, ...)
- [ ] missing statements #24171

PR #24578, opened 2025-05-25
native: use builtin exit function

Eliyaan:
- allow the exit function from builtin (instead of using the native one)

changes need to be made to issues:
#24172 cross out exit() from the list
    and assert() as it is not longer a function from builtin (now specific to each backend)

### DIFF
--- a/vlib/v/gen/native/amd64.v
+++ b/vlib/v/gen/native/amd64.v
@@ -1758,7 +1758,7 @@ pub fn (mut c Amd64) call_fn(node ast.CallExpr) {
 	name := node.name
 	mut n := name
 	if !n.contains('.') && n !in c.g.fn_addr.keys() { // if the name is in keys, it is a function from builtin
-		n = 'main.${n}' 
+		n = 'main.${n}'
 	}
 	if node.is_method {
 		n = '${c.g.table.get_type_name(node.receiver_type)}.${node.name}'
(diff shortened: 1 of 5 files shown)


vlib/v/gen/native/tests/builtin.vv
```diff
@@ -0,0 +1,46 @@
...
+*/
+
+gc_is_enabled() // not yet enabled so only calling it 
+
+println(int_max(32, -32))
+println(int_max(-32, 32))
+println(int_min(32, -32))
+println(int_min(-32, 32))
+
+// print_character(`a`) enable when C vars will work
```
JalonSolov, line 27:
```suggestion
// print_character(`a`) enable when runes work
```
C characters are 8-bit chars.  `a` is a rune, or UTF-8 char in V.

Eliyaan: (runes seem to be already working)

vlib/v/gen/native/amd64.v
```diff
@@ -1757,7 +1757,7 @@ fn (mut c Amd64) sar8(r Amd64Register, val u8) {
 pub fn (mut c Amd64) call_fn(node ast.CallExpr) {
 	name := node.name
 	mut n := name
-	if !n.contains('.') {
+	if !n.contains('.') && n !in c.g.fn_addr.keys() { // if the name is in keys, it is a function from builtin
```
spytheman, line 1760: imho, cache `c.g.fn_addr.keys()` in `c`, since calling .keys() for each call can become expensive.

PR #24580, opened 2025-05-26
markused: fix for thread waiter usage

### LINKED ISSUE #24577: Spawning anonymous functions that have no input arguments  (example from book: Getting Started with V Programming) fails
<details>
<summary>V version: V 0.4.10 52c7130, press to see full `v doctor` output</summary>

|V full version      |V 0.4.10 22d2d1a.52c7130
|OS                  |linux, "Manjaro Linux"
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 9 7940HS w/ Radeon 780M Graphics
|Memory              |13.68GB/30.57GB
|                    |
|V executable        |/home/pepe/Dev/v/v
|V last modified time|2025-05-25 10:55:16
|                    |
|V home dir          |OK, value: /home/pepe/Dev/v
|VMODULES            |OK, value: /home/pepe/.vmodules
|Current working dir |OK, value: /home/pepe/_mysrc/test/vlang
|                    |
|V git status        |weekly.2025.21-26-g52c7130a
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.1.1 20250425
|gcc version         |gcc (GCC) 15.1.1 20250425
|tcc version         |tcc version 0.9.28rc 2024-02-05 HEAD@105d70f7 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 a0799a5b
|emcc version        |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  concurrency06.v && concurrency06`
```v
module main

fn main() {
	t := go fn () string {
		return 'hi'
	}()
	x := t.wait()
	println(typeof(x).name) // string
	println(x) // hi
}

```

**What did you see?**
```
Compilation with unknown C compiler
cc: /tmp/v_1000/concurrency06.01JW4FBQ8TPYRJTYPD78066AW9.tmp.c:1290: warning: implicit declaration of function '_v_free'
cc: tcc: error: undefined symbol '_v_free'
builder error: 
==================
```

'hi'

felipensp:
Fix #24577

### DIFF
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -32,6 +32,7 @@ pub mut:
 	map_update       bool            // {...foo}
 	interpolation    bool            // '${foo} ${bar}'
 	option_or_result bool            // has panic call
+	waiter           bool            // has thread waiter
 	print_types      map[int]bool    // print() idx types
 	used_fns         map[string]bool // filled in by markused
 	used_consts      map[string]bool // filled in by markused
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -2726,6 +2726,9 @@ fn (mut c Checker) spawn_expr(mut node ast.SpawnExpr) ast.Type {
 		c.error('option handling cannot be done in `spawn` call. Do it when calling `.wait()`',
 			node.call_expr.or_block.pos)
 	}
+	if node.is_expr {
+		c.table.used_features.waiter = true
+	}
 	// Make sure there are no mutable arguments
 	for arg in node.call_expr.args {
 		if arg.is_mut && !arg.typ.is_ptr() {
@@ -2760,6 +2763,9 @@ fn (mut c Checker) go_expr(mut node ast.GoExpr) ast.Type {
 	if node.call_expr.or_block.kind != .absent {
 		c.error('option handling cannot be done in `go` call. Do it when calling `.wait()`',
 			node.call_expr.or_block.pos)
+	}
+	if node.is_expr {
+		c.table.used_features.waiter = true
 	}
 	// Make sure there are no mutable arguments
 	for arg in node.call_expr.args {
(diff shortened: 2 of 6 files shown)


ctkjose:
The `go` statement is a **coroutine**. In the issue #24577 the `go` statement became a `spawn` because the option `-use-coroutines` was not used.

The problem is that spawn does not use v code that the builder could use to detect dependencies.

If modifying `markused.v` you could just do:

```v
if table.gostmts > 0 {
	all_fn_root_names << 'free'
}
```

Using `table.gostmts` should be an acceptable way to assume a `spawn` without the need to modify checker or the table.

A **better solution** is to set the dependency explicitly in `spawn_expr()` with `p.register_auto_import('builtin.free')`. This option for me is the more elegant and isolates the changes to only one place without guessing.

**PS** In `vlib/v/gen/c/cheaders.v` we declare `void v_free(voidptr ptr);`. This is hardwired, so was it meant to be a base function always included? It was committed in [88810a7](https://github.com/vlang/v/commit/8810af76df00943c5c8bef881a0116690fc6a57d) for `-gc boehm` which added the `C.GC_*` functions.  Wonder @spytheman could give us more clues.

spytheman (approved):
Good work.

ctkjose:
> I agree, that using `if table.gostmts > 0 {` would be simpler, but it will also have a lower precision. Consider a `go f()` call, with `fn f() { println('hi') }` for example, that does not return anything, and that will not generate a `_v_free()` call in the V gowrapper function.

If I recall correctly the fix in this PR will also inject free regardless, it also adds more noise to the table. 

The issue is that the spawn_and_go.v does not use actual v code for it to properly detect dependencies. Maybe an alternate version of `parser.register_auto_import()` that allows to register functions would avoid having to hack situations like this.

Issue #24584, opened 2025-05-26
checker: wrapping mutable array in parenthesis circumvents checker


4hnme:
### Describe the bug

when taking a pointer to an element of a mutable array, adding extra parenthesis sidesteps safety check

### Reproduction Steps

```v
mut arr := [1, 2, 3]
mut ptr1 := &(arr[1]) // <- this should error, but it doesn't
mut ptr2 := &arr[1]   // <- this errors, unless put inside unsafe block
```

### Expected Behavior

the checker should catch this

### Current Behavior

the checker ignores it

### Additional Information/Context

there's already [a check](https://github.com/vlang/v/blob/master/vlib/v/checker/checker.v?plain=#L4832-L4837) in place for the exact situation, but it is sidestepped with added parenthesis. my attempts at dump debugging show that in this case `typ_sym` is treated like an immutable value rather than a mutable array

### V version

V 0.4.10 52c7130

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 52c7130a2f870afff1242c476aa6db11941abdb9
|:-------------------|:-------------------
|OS                  |linux, "Void Linux"
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i5-2500 CPU @ 3.30GHz
|Memory              |1.11GB/7.74GB
|                    |
|V executable        |/home/username/thirdparty/v/v
|V last modified time|2025-05-26 13:18:58
|                    |
|V home dir          |OK, value: /home/username/thirdparty/v
|VMODULES            |OK, value: /home/username/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/username/thirdparty/v
|                    |
|Git version         |git version 2.49.0
|V git status        |weekly.2025.08-533-g52c7130a
|.git/config present |true
|                    |
|cc version          |cc (GCC) 14.2.1 20250405
|gcc version         |gcc (GCC) 14.2.1 20250405
|clang version       |clang version 19.1.4
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.41
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

[...]

PR #24588, opened 2025-05-27
checker: cycle through all `ast.ParExpr` first in `prefix_expr` 

Delta456:
Fixes https://github.com/vlang/v/issues/24584


StunxFS:
Possibly out of scope for this PR, but an `Expr.remove_paren()` method would be great for removing the `()`, so it can be used anywhere you need to `expr is SomeExpr`

Delta456:
> Possibly out of scope for this PR, but an `Expr.remove_paren()` method would be great for removing the `()`, so it can be used anywhere you need to `expr is SomeExpr`

It's a good idea @StunxFS. I am planning to make this method after this PR gets merged. It will help in all other occurrences.

spytheman (approved):
Excellent work.

Issue #24595, opened 2025-05-28
Upgrade thirdparty/zstd to version 1.5.7


lcheylus:
Currently, `thirdparty/zstd` (Zstandard - Fast real-time compression algorithm from Meta/Facebook) **version = 1.5.6**
https://github.com/vlang/v/blob/52c7130a2f870afff1242c476aa6db11941abdb9/thirdparty/zstd/zstd.h#L106-L109

**Zstd latest version = 1.5.7**, released Feb 19 2025 => https://github.com/facebook/zstd/releases/tag/v1.5.7

Proposal to upgrade `thirdparty/zstd` to the latest version => combine files in 2 files `zstd.{c,h}` with scripts from https://github.com/facebook/zstd/tree/dev/build/single_file_libs

lcheylus:
`thirdparty/zstd` upgraded to version 1.5.7

PR #24597, opened 2025-05-28
thirdparty.zstd: upgrade to 1.5.8

kbkpbot:
require by issue #24595


jorgeluismireles:
I think your title has a typo 1.5.8 as the version is 1.5.7

lcheylus:
@kbkpbot Thanks for your work. 

But why do you use current `zstd` GitHub source instead of **the latest official release (version 1.5.7)** as requested in my issue ? => https://github.com/facebook/zstd/releases/tag/v1.5.7

**For thirdparty, it's better to use release/tag for update** => simpler to follow and with current sources from GH, there may be regressions and bugs.

PR #24611, opened 2025-05-30
thirdparty.zstd: upgrade to v1.5.7 release

kbkpbot:
Required by #24595

A update for PR #24597, with a release version of `zstd`.

Tested on Ubuntu/OpenBSD/FreeBSD with `tcc`.

### DIFF
--- a/thirdparty/zstd/zstd.c
+++ b/thirdparty/zstd/zstd.c
@@ -108,9 +108,9 @@
  */
 #if defined(__linux) || defined(__linux__) || defined(linux) || defined(__gnu_linux__) || \
     defined(__CYGWIN__) || defined(__MSYS__)
-# if !defined(_GNU_SOURCE) && !defined(__ANDROID__) /* NDK doesn't ship qsort_r(). */
-#   define _GNU_SOURCE
-# endif
+#if !defined(_GNU_SOURCE) && !defined(__ANDROID__) /* NDK doesn't ship qsort_r(). */
+#define _GNU_SOURCE
+#endif
 #endif
 
 #include <limits.h>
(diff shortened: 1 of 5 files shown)


lcheylus:
**Tests are not OK for Ubuntu/MUSL** in Docker CI (https://github.com/vlang/v/actions/runs/15339431994/job/43162690935?pr=24611) => errors with tests for zstd
```
To reproduce just failure 1 run:    '/__w/v/v/v' -silent '/__w/v/v/vlib/compress/zstd/read_zstd_files_test.v'
To reproduce just failure 2 run:    '/__w/v/v/v' -silent '/__w/v/v/vlib/compress/zstd/zstd_test.v'
```

lcheylus:
> Yes, I saw it. And I have no idea how to detect system is using `MUSL` or not. :(

TIL there is no `__MUSL__` macro to detect it :( https://wiki.musl-libc.org/faq.html#Q:_why_is_there_no_MUSL_macro_.3F

spytheman:
V does support `$if musl ? {`, which works based on the presence of `/etc/alpine-release` or if the `v` executable itself was built with it (i.e. environment/runtime detection).

Having a file with the `_d_musl.c.v` platform postfix, should also work (it can do `#define ZSTD_USE_C90_QSORT` in it) if it is needed.

thirdparty/zstd/zstd.c
```diff
@@ -108,9 +108,9 @@
  */
 #if defined(__linux) || defined(__linux__) || defined(linux) || defined(__gnu_linux__) || \
     defined(__CYGWIN__) || defined(__MSYS__)
-#if !defined(_GNU_SOURCE) && !defined(__ANDROID__) /* NDK doesn't ship qsort_r(). */
-#define _GNU_SOURCE
-#endif
+# if !defined(_GNU_SOURCE) && !defined(__ANDROID__) /* NDK doesn't ship qsort_r(). */
+#   define _GNU_SOURCE
+# endif
```
JalonSolov, line 113: Why the extra spaces?

kbkpbot: just copy from https://github.com/facebook/zstd/pull/4312/files
I will remove them.

PR #24627, opened 2025-05-31
native: nested structs, more support for right expr of IndexExpr

Eliyaan:
- support for nested structs
- support more expressions for the right expr of IndexExpr
- make the StringLiterals more robust (explicit lea instead of relying on previous state)

### DIFF
--- a/vlib/v/gen/native/tests/struct.vv
+++ b/vlib/v/gen/native/tests/struct.vv
@@ -208,6 +208,15 @@ struct Nest3 {
 	a Nest2
 }
 
+struct Int {
+        a int
+}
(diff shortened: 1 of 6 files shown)


vlib/v/gen/native/tests/struct.vv
```diff
@@ -196,9 +196,45 @@ fn assign_fields() {
 	assert f.a == 2
 }
 
+struct Nest {
+        a string
+}
```
spytheman, line 201: Can you add a `before string` and `after string` here too, to make sure that the offset of the inner field is correct?

PR #24628, opened 2025-05-31
type_resolver: support `$if T is $pointer {` and `$if T is $voidptr {`, to make it easier to implement a pure V dump(), without relying on specific backends

spytheman:


enghitalo:
I love you! I've been waiting for this for a looooong time!

enghitalo:
Related to #20398

Issue #24646, opened 2025-06-03
Add error code for free_memory function


lcheylus:
In `vlib/runtime`, `free_memory` function returns free physical memory found on the system. Implemented only for on Darwin, FreeBSD, Linux, OpenBSD and Windows ; otherwise, returns 1.

On Darwin, FreeBSD, OpenBSD, implementation uses unsafe C code (see `vlib/runtime/free_memory_impl_{darwin,freebsd,openbsd}.c.v`) but errors are not checked. In some case, this unsafe code can fail.

I propose to **return 0 as error code for `free_memory`** function to handle this case.

1. Check return value for unsafe C code on Darwin, FreeBSD and OpenBSD
2. Return 0 if error

By the end, `free_memory` function returns:

- free physical memory found on the system
- 1 if not implemented
- 0 if error.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

JalonSolov:
What if the amount of free memory is 0, or 1 byte?  :-\

Sounds like `free_memory` should've had a Result type return value.

spytheman:
I agree, if it can fail, it should return an error, that is the V convention.

spytheman:
in case it is not implemented, it can `return error('not implemented')`

PR #24651, opened 2025-06-04
runtime: free_memory() total_memory() return Result types

kbkpbot:
Fix issue #24646

Fix for `macos`:  free host port by using `mach_port_deallocate()`

### DIFF
--- a/vlib/runtime/runtime_test.v
+++ b/vlib/runtime/runtime_test.v
@@ -1,11 +1,17 @@
 import runtime
 
 fn test_physical_memory() {
-	total := runtime.total_memory() or { 1 }
-	free := runtime.free_memory() or { 1 }
-	println('total memory: ${total}')
-	println('free  memory: ${free}')
-	assert total > 0 && free > 0
+	$if windows || linux || darwin || freebsd || openbsd {
+		total := runtime.total_memory()!
+		free := runtime.free_memory()!
+		println('total memory: ${total}')
+		println('free  memory: ${free}')
+		assert total > 0 && free > 0
+	} $else {
+		total := runtime.total_memory()!
+		_ := runtime.free_memory() or { assert err.msg().contains('not implemented') }
+		assert total > 0
+	}
 }
 
 fn test_nr_cpus() {


vlib/runtime/free_memory_impl_linux.c.v
```diff
@@ -1,15 +1,20 @@
...
 	$if cross ? {
-		return 1
+		return error('free_memory: not implemented')
 	}
 	$if !cross ? {
 		$if linux {
 			page_size := usize(C.sysconf(C._SC_PAGESIZE))
+			c_errno_1 := C.errno
 			av_phys_pages := usize(C.sysconf(C._SC_AVPHYS_PAGES))
+			c_errno_2 := C.errno
```
Delta456: Are `c_errno_1` and `c_errno_2` the error codes of `page_size` and `av_phys_pages`? If so, can you add a comment for that?

vlib/runtime/free_memory_impl_linux.c.v
```diff
@@ -1,15 +1,23 @@
...
-		return 1
+		return error('free_memory: not implemented')
 	}
 	$if !cross ? {
 		$if linux {
 			page_size := usize(C.sysconf(C._SC_PAGESIZE))
+			c_errno_1 := C.errno
+			if page_size == usize(-1) {
+				return error('free_memory: `C.sysconf(C._SC_PAGESIZE)` return error code = ${c_errno_1}')
+			}
```
lcheylus, line 13: `c_errno_1` is unnecessary => you could print error code directly with `return error code = ${C.errno}`

vlib/runtime/free_memory_impl_linux.c.v
```diff
@@ -1,15 +1,23 @@
...
 			av_phys_pages := usize(C.sysconf(C._SC_AVPHYS_PAGES))
+			c_errno_2 := C.errno
+			if av_phys_pages == usize(-1) {
+				return error('free_memory: `C.sysconf(C._SC_AVPHYS_PAGES)` return error code = ${c_errno_2}')
+			}
```
lcheylus, line 18: `c_errno_2` is unnecessary => you could print error code directly with `return error code = ${C.errno}`

vlib/runtime/free_memory_impl_freebsd.c.v
```diff
@@ -2,21 +2,37 @@ module runtime
...
-		return 1
+		return error('free_memory: not implemented')
 	}
 	$if !cross ? {
 		$if freebsd {
 			page_size := usize(C.sysconf(C._SC_PAGESIZE))
+			c_errno_1 := C.errno
+			if page_size == usize(-1) {
+				return error('free_memory: `C.sysconf()` return error code = ${c_errno_1}')
+			}
```
lcheylus, line 15: `c_errno_1` is unnecessary => you could print error code directly with `return error code = ${C.errno}`

vlib/runtime/free_memory_impl_freebsd.c.v
```diff
@@ -2,21 +2,37 @@ module runtime
...
 			mut mib := [4]int{}
 			mut len := usize(4)
-			unsafe { C.sysctlnametomib(c'vm.stats.vm.v_free_count', &mib[0], &len) }
+			retval_2 := unsafe {
+				C.sysctlnametomib(c'vm.stats.vm.v_free_count', &mib[0], &len)
+			}
+			c_errno_2 := C.errno
+			if retval_2 == -1 {
+				return error('free_memory: `C.sysctlnametomib()` return error code = ${c_errno_2}')
```
lcheylus, line 23: Idem, print directly `${C.errno}`

vlib/runtime/free_memory_impl_freebsd.c.v
```diff
@@ -2,21 +2,37 @@ module runtime
...
+			}
 			mut free_pages := int(0)
 			bufsize := usize(4)
-			unsafe { C.sysctl(&mib[0], mib.len, &free_pages, &bufsize, 0, 0) }
+			retval_3 := unsafe {
+				C.sysctl(&mib[0], mib.len, &free_pages, &bufsize, 0, 0)
+			}
+			c_errno_3 := C.errno
+			if retval_3 == -1 {
+				return error('free_memory: `C.sysctl()` return error code = ${c_errno_3}')
```
lcheylus, line 32: Idem, print directly `${C.errno}`

vlib/runtime/runtime_nix.c.v
```diff
@@ -8,15 +8,23 @@ pub fn nr_cpus() int {
 }
 
 // total_memory returns total physical memory found on the system.
-pub fn total_memory() usize {
+pub fn total_memory() !usize {
 	page_size := usize(C.sysconf(C._SC_PAGESIZE))
+	c_errno_1 := C.errno
+	if page_size == usize(-1) {
+		return error('total_memory: `C.sysconf(C._SC_PAGESIZE)` return error code = ${c_errno_1}')
```
lcheylus, line 15: Print directly `${C.errno}`

vlib/runtime/runtime_nix.c.v
```diff
@@ -8,15 +8,23 @@ pub fn nr_cpus() int {
...
+	}
 	phys_pages := usize(C.sysconf(C._SC_PHYS_PAGES))
+	c_errno_2 := C.errno
+	if phys_pages == usize(-1) {
+		return error('total_memory: `C.sysconf(C._SC_PHYS_PAGES)` return error code = ${c_errno_2}')
```
lcheylus, line 20: Idem, print directly `${C.errno}`

vlib/runtime/runtime_test.v
```diff
@@ -1,8 +1,8 @@
 import runtime
 
 fn test_physical_memory() {
-	total := runtime.total_memory()
-	free := runtime.free_memory()
+	total := runtime.total_memory() or { 1 }
+	free := runtime.free_memory() or { 1 }
```
lcheylus: Why return 1 if error here ? If there is an error when calling `total_memory`/`free_memory`, this test must fail => return `or { 0 }` instead.

vlib/runtime/free_memory_impl_darwin.c.v
```diff
@@ -17,21 +18,35 @@ pub struct C.vm_statistics64_data_t {
...
 
-fn free_memory_impl() usize {
+fn free_memory_impl() !usize {
 	$if macos {
 		mut hs := C.vm_statistics64_data_t{}
 		mut vmsz := u32(C.HOST_VM_INFO64_COUNT)
 		mut hps := u32(0)
 		mut host := C.mach_host_self()
+		defer {
+			_ := C.mach_port_deallocate(C.mach_task_self(), host)
```
spytheman, line 48: Is the need for that documented somewhere?

vlib/runtime/runtime_nix.c.v
```diff
@@ -8,15 +8,23 @@ pub fn nr_cpus() int {
...
+	}
 	return page_size * phys_pages
 }
 
 // free_memory returns free physical memory found on the system.
 // Note: implementation available only on Darwin, FreeBSD, Linux, OpenBSD and
 // Windows. Otherwise, returns 1.
```
lcheylus: This note for `free_memory` must be updated : 

// free_memory returns free physical memory found on the system or error. 
// Note: implementation available only on Darwin, FreeBSD, Linux, OpenBSD and
// Windows. Otherwise, returns 'free_memory: not implemented'

vlib/runtime/free_memory_impl_freebsd.c.v
spytheman: No, it is better to assign it first, 
right after the C.sysconf() call, that could potentially fail; string interpolation may also call libc functions, that may reset C.errno .

vlib/runtime/free_memory_impl_freebsd.c.v
spytheman: no, it is better as it is.

vlib/runtime/free_memory_impl_linux.c.v
kbkpbot: I do think this is the correct way catch the real `errno` create by the function call `C.sysconf(C._SC_PAGESIZE)`. 
Because following string interpolation may overwrite the `errno`. 
Just like under Windows,  in `os.ptr_win_get_error_msg()`, it will call some functions such as `C.FormatMessageW` which will overwrite global error code.

vlib/runtime/free_memory_impl_freebsd.c.v
lcheylus: OK, thanks for the explanation. 

@kbkpbot You can mark as resolved all my remarks about `C.errno`.

vlib/runtime/runtime_test.v
spytheman: in order to have the next prints() show, and the asserts after that to work

lcheylus: I don't understand. If there is an error in `total_memory` or `free_memory` in this test, value = 1 will be returned and the following assert `assert total > 0 && free > 0` does not fail whereas in this case it should.

kbkpbot: This is a workaround for systems which do not support `free_memory` `total_memory`.
Or we have to write  complex system detection logic.

lcheylus:
To have the same prototype on every system, `total_memory` and `free_memory` functions have to return  Result type (`!usize`) on Windows too => `vlib/runtime/runtime_windows.c.v`.

vlib/runtime/runtime_test.v
lcheylus: OK but on systems where `free_memory` is supported (FreeBSD, OpenBSD, Linux, Darwin and Windows), this test will never fail even if there is an error.

spytheman:
Excellent work.

PR #24652, opened 2025-06-04
parser, ast, fmt: convert `"hello".str` => `c"hello"` (fix #24635)

### LINKED ISSUE #24635: vfmt: format `"c string literal".str` to `c"c string literal"`
### Describe the feature

I've seen many people writing `"some string".str` instead of `c"some string"` to create a native string, which seems unnecessary to me. I suggest that `vfmt` formats that piece of code as needed: `"my string".str` becomes `c"my string"` (`js"my string"` if the backend is javascript).

### Use Case

```v
$ cat main.v
fn main() {
	x := 'hello'.str
	y := c'hello'
	_, _ = x, y
}
$ v -printfn main__main main.v
	u8* x = _SLIT("hello").str;
	u8* y = "hello";
	{u8* _ = x;}
	;
	{u8* _ = y;}
	;
}
```

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### Version used

V latest

any

> [!NOTE]
>

StunxFS:
The compiler will now treat `"hello".str` as if it were written `c"hello"` (or `js"hello"`), which avoids creating an unnecessary `string` object. The formatter will also correct that piece of code by adding the necessary prefix. 

~~In addition to that, this PR depends on another PR (#24653) that will implement JS string literals (I thought V already supported them), now it will be possible to create native JS strings using a simple prefix (just like you would do with C).~~
The JS backend is ignored.

Fix #24635.

### DIFF
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -3039,7 +3039,8 @@ pub fn (mut f Fmt) select_expr(node ast.SelectExpr) {
 }
 
 pub fn (mut f Fmt) selector_expr(node ast.SelectorExpr) {
-	if node.expr is ast.StringLiteral && node.field_name == 'str' && f.file.language != .js {
+	if node.expr is ast.StringLiteral && node.field_name == 'str' && !f.pref.backend.is_js()
+		&& f.file.language != .js {
 		if f.file.language != .v {
 			f.write(f.file.language.str())
 		} else {
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -3169,6 +3169,8 @@ pub fn (mut f Fmt) string_literal(node ast.StringLiteral) {
 		f.write('r')
 	} else if node.language == ast.Language.c {
 		f.write('c')
+	} else if node.language == ast.Language.js {
+		f.write('js')
 	}
 	if node.is_raw {
 		f.write('${quote}${node.val}${quote}')
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -3039,7 +3039,7 @@ pub fn (mut f Fmt) select_expr(node ast.SelectExpr) {
 }
 
 pub fn (mut f Fmt) selector_expr(node ast.SelectorExpr) {
-	if node.expr is ast.StringLiteral && node.field_name == 'str' {
+	if node.expr is ast.StringLiteral && node.field_name == 'str' && f.file.language != .js {
 		if f.file.language != .v {
 			f.write(f.file.language.str())
 		} else {
--- a/vlib/v/fmt/fmt.v
+++ b/vlib/v/fmt/fmt.v
@@ -3042,8 +3042,8 @@ pub fn (mut f Fmt) select_expr(node ast.SelectExpr) {
 pub fn (mut f Fmt) selector_expr(node ast.SelectorExpr) {
 	// TODO(StunxFS): Even though we ignored the JS backend, the `v/gen/js/tests/js.v`
 	// file was still formatted/transformed, so it is specifically ignored here. Fix this.
-	if node.expr is ast.StringLiteral && node.field_name == 'str' && !f.pref.backend.is_js()
-		&& f.file.language != .js
+	if f.file.language != .js && node.expr is ast.StringLiteral && node.field_name == 'str'
+		&& !f.pref.backend.is_js()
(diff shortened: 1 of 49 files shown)


vlib/v/fmt/fmt.v
```diff
@@ -3039,6 +3040,15 @@ pub fn (mut f Fmt) select_expr(node ast.SelectExpr) {
 }
 
 pub fn (mut f Fmt) selector_expr(node ast.SelectorExpr) {
+	// TODO(StunxFS): Even though we ignored the JS backend, the `v/gen/js/tests/js.v`
+	// file was still formatted/transformed, so it is specifically ignored here. Fix this.
+	if node.expr is ast.StringLiteral && node.field_name == 'str' && !f.pref.backend.is_js()
+		&& f.file.language != .js
```
spytheman: put that condition first, since it is cheaper to check, compared to the string comparisons and the function calls

vlib/v/transformer/transformer.v
```diff
@@ -673,11 +674,21 @@ pub fn (mut t Transformer) expr(mut node ast.Expr) ast.Expr {
...
-						pos: node.pos
+			if mut node.expr is ast.StringLiteral {
+				// TODO(StunxFS): Even though we ignored the JS backend, the `v/gen/js/tests/js.v`
+				// file was still formatted/transformed, so it is specifically ignored here. Fix this.
+				if node.field_name == 'str' && !t.pref.backend.is_js() && t.file.language != .js
+					&& !t.file.path.ends_with(os.join_path('v', 'gen', 'js', 'tests', 'js.v')) {
+					return ast.StringLiteral{
+						...node.expr
+						language: .c
+					}
```
spytheman: The fmt conversion is enough. This edge case in the transformer, is something that can be avoided.

StunxFS: Okay, I removed that from the transformer.

vlib/v/fmt/fmt.v
```diff
@@ -3039,8 +3040,11 @@ pub fn (mut f Fmt) select_expr(node ast.SelectExpr) {
 }
 
 pub fn (mut f Fmt) selector_expr(node ast.SelectorExpr) {
+	// TODO(StunxFS): Even though we ignored the JS backend, the `v/gen/js/tests/js.v`
+	// file was still formatted/transformed, so it is specifically ignored here. Fix this.
 	if node.expr is ast.StringLiteral && node.field_name == 'str' && !f.pref.backend.is_js()
-		&& f.file.language != .js {
+		&& f.file.language != .js
+		&& !f.file.path.ends_with(os.join_path('v', 'gen', 'js', 'tests', 'js.v')) {
```
JalonSolov, line 3047:
```suggestion
		&& !f.file.path.ends_with(os.join_path('.v', '.gen', '.js', '.tests', '.js.v')) {
```
If you don't include the `.`, you could run into an issue where the base name includes those characters.

vlib/v/transformer/transformer.v
```diff
@@ -674,7 +675,10 @@ pub fn (mut t Transformer) expr(mut node ast.Expr) ast.Expr {
 		ast.SelectorExpr {
 			node.expr = t.expr(mut node.expr)
 			if mut node.expr is ast.StringLiteral {
-				if node.field_name == 'str' && !t.pref.backend.is_js() && t.file.language != .js {
+				// TODO(StunxFS): Even though we ignored the JS backend, the `v/gen/js/tests/js.v`
+				// file was still formatted/transformed, so it is specifically ignored here. Fix this.
+				if node.field_name == 'str' && !t.pref.backend.is_js() && t.file.language != .js
+					&& !t.file.path.ends_with(os.join_path('v', 'gen', 'js', 'tests', 'js.v')) {
```
JalonSolov: Same as above.

vlib/v/fmt/fmt.v
StunxFS: Why? If I add `.`, the result returned by `join_path` will be `.v/.gen/.js/.tests/.js.v`

JalonSolov: Hmm... ok, I was obviously looking at it wrong.  I saw the `js.v` and was thinking that was an extension.  Is it actually a subdir?  Seems... wrong...

StunxFS: Yes, `js.v` is a test file

spytheman (approved):
Excellent work.

PR #24653, opened 2025-06-04
parser, fmt, gen: support `js"string literal"`

StunxFS:
This PR aims to add native JS string support to the compiler. It will now be possible to write `js"Hello"` just like you would write `c"Hello"`.

This test ran successfully locally (using JS Browser):

```v
fn test_js_string() {
	s := js'hello V'
	assert s.charAt(0) == js'h'
	assert s.charAt(6) == js'V'
	assert s.charCodeAt(0) == JS.Number(104)
	assert s.toUpperCase() == js'HELLO V'
	assert s.toLowerCase() == js'hello v'
	assert s.concat(js' from JS') == js'hello V from JS'
	assert s.includes(js' ') == JS.Boolean(true)
	assert s.startsWith(js'hello') == JS.Boolean(true)
	assert s.endsWith(js'V') == JS.Boolean(true)
}
```

### DIFF
--- a/vlib/builtin/js/string_test.js.v
+++ b/vlib/builtin/js/string_test.js.v
@@ -985,11 +985,9 @@ fn test_index_any() {
 
 fn test_js_string() {
 	s := js'hello V'
-	assert s.charAt(0) == js'h'
-	assert s.charAt(6) == js'V'
-	// FIXME: This line of code causes a bug where V claims that `voidptr` was used to
-	// initialize a `JS.Number` value.
-	// assert s.charCodeAt(0) == JS.Number(104)
+	assert s.charAt(JS.Number(0)) == js'h'
+	assert s.charAt(JS.Number(6)) == js'V'
+	assert s.charCodeAt(JS.Number(0)) == JS.Number(104)
 	assert s.toUpperCase() == js'HELLO V'
 	assert s.toLowerCase() == js'hello v'
 	assert s.concat(js' from JS') == js'hello V from JS'


enghitalo:
What is a JS string?

JalonSolov:
JavaScript has 2 types of strings - primitive strings and String objects.  Primitive strings do not have any methods or properties, while String objects have multiples of both.

Basically, it is the same as the difference between C strings and V strings.

vlib/v/checker/checker.v
```diff
@@ -148,6 +148,8 @@ mut:
 
 	v_current_commit_hash string // same as old C.V_CURRENT_COMMIT_HASH
 	assign_stmt_attr      string // for `x := [1,2,3] @[freed]`
+
+	js_string ?ast.Type // will receive a value when `js"hello"` is evaluated
```
Delta456: Doesn't this need to be optional?

vlib/v/checker/checker.v
```diff
@@ -3278,6 +3280,15 @@ pub fn (mut c Checker) expr(mut node ast.Expr) ast.Type {
 				// string literal starts with "c": `C.printf(c'hello')`
 				return ast.u8_type.set_nr_muls(1)
 			}
+			if node.language == .js {
+				// string literal starts with "js": `JS.console.log(js'hello')`
+				if c.js_string != none {
```
Delta456: I am sure the check could be `!c.js_string.is_void()`

vlib/builtin/js/string_test.js.v
```diff
@@ -982,3 +982,19 @@ fn test_index_any() {
 	assert x.index_any('ef') == 4
 	assert x.index_any('fe') == 4
 }
+
+fn test_js_string() {
+	/*
+	TODO(StunxFS): Uncomment when the PR is merged
```
spytheman: Why uncomment it later, and not now? 
V is bootstrapped in the PR, and the test should be runnable imho as it is?

spytheman (approved):
Good work.
@medvednikov what do you think about it (adding syntax support for `js'strings'`) ?

vlib/v/checker/checker.v
```diff
@@ -148,6 +148,8 @@ mut:
 
 	v_current_commit_hash string // same as old C.V_CURRENT_COMMIT_HASH
 	assign_stmt_attr      string // for `x := [1,2,3] @[freed]`
+
+	js_string ast.Type = ast.void_type // when `js"string literal"` is used, `js_string` will be equal to `JS.String`
```
Delta456, line 152: There's no need to have the default value `ast.void_type`

StunxFS:
For some strange reason, `test_js_string` fails with `v test`. The error it throws is the following:
```v
>> compilation failed:
jsgen error: could not generate string method voidptr_str for type 'voidptr'
```

If I generate .js from a normal file with the same content, everything compiles and runs perfectly:

```v
$ cat main.v
fn main() {
        s := js'hello V'
        assert s.charAt(0) == js'h'
        assert s.charAt(6) == js'V'
        assert s.charCodeAt(0) == JS.Number(104)
        assert s.toUpperCase() == js'HELLO V'
        assert s.toLowerCase() == js'hello v'
        assert s.concat(js' from JS') == js'hello V from JS'
        assert s.includes(js' ') == JS.Boolean(true)
        assert s.startsWith(js'hello') == JS.Boolean(true)
        assert s.endsWith(js'V') == JS.Boolean(true)
}
$ v -b js main.v
$ node main.js
$ 
```

JalonSolov:
First impression... it's related to line 152 in checker.v, where you explicitly set the type to `ast.void_type`.

StunxFS:
@JalonSolov 
> First impression... it's related to line 152 in checker.v, where you explicitly set the type to ast.void_type.

If I remove the default value from the field, V panics:

```
V panic: table.sym: invalid type (typ=ast.Type(0x0 = 0) idx=0). Compiler bug. This should never happen. Please report the bug using `v bug file.v`.
```

medvednikov:
Looks good to me. A simple change.

StunxFS:
After some testing, I discovered that the source of the error with `v test` is this line of code:

`assert s.charCodeAt(0) == JS.Number(104)`

V thinks that some `voidptr` value was used to initialize a `JS.Number` instance and so fails to create the corresponding `.str` method. Now I'm not sure if it would be appropriate for the JS backend to support generating `.str()` for `voidptr` or not.

westtrade:
Looks cool, but for what purposes is this necessary?

PR #24658, opened 2025-06-05
checker: fix missing type bounding to match expr on or-expr

### LINKED ISSUE #24656: Match in `or` or block returns unexpected value.
### Describe the bug

Match in `or` or block returns unexpected value.

### Reproduction Steps

```v
struct Foo {
	option ?int = none
}

fn main() {
	test := true
	foo := Foo{}
	result := foo.option or {
		match test {
			true { 1 }
			else { 2 }
		}
	}

	assert result == 1
}
```

### Expected Behavior

assert should not fail

### Current Behavior

```
./main.v:15: FAIL: fn main.main: assert result == 1
  left value: result = 0
  right value: 1
V panic: Assertion failed...
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.10 262a72e

|V full version      |V 0.4.10 da3112e.262a72e
|Memory              |0.16GB/8GB
|                    |
|V last modified time|2025-06-03 18:43:40
|                    |
|Current working dir |OK, value: /Users/mike/Documents/github/bug
|                    |
|env VFLAGS          |"-path /Users/mike/Documents/github|@vlib|@vmodules -message-limit 5 -no-parallel"
|V git status        |weekly.2025.17-211-g262a72ea
|.git/config present |true
|                    |
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

felipensp:
Fix #24656

### DIFF
--- a/vlib/v/tests/match_or_expr_test.v
+++ b/vlib/v/tests/match_or_expr_test.v
@@ -1,0 +1,16 @@
+struct Foo {
+	option ?int = none
+}
+
+fn test_main() {
+	test := true
+	foo := Foo{}
+	result := foo.option or {
+		match test {
+			true { 1 }
+			else { 2 }
+		}
+	}
+
+	assert result == 1
+}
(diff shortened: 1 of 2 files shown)


vlib/v/tests/match_or_expr_test.v
```diff
@@ -0,0 +1,16 @@
+struct Foo {
+	option ?int = none
```
spytheman, line 2: that gives a warning, and afaik is not needed for the PR

spytheman (approved):
Good work.

PR #24660, opened 2025-06-05
native: c constants

Eliyaan:
- is able to link with c constants (important prerequisite for enabling builtin)
-  += -= assigns

for enums, it will calculate the expression multiple times, but it is simpler for now.


vlib/v/gen/native/amd64.v
```diff
@@ -311,6 +311,9 @@ fn (mut c Amd64) cmp_var_reg(var Var, reg Register, config VarConfig) {
 					// TODO
 					// g.cmp()
 				}
+				else {
+					c.g.n_error('${@LOCATION} unsupported var type ${var_object}')
+				}
```
spytheman: Having no `else{}` forced the match to be exhaustive, and thus the compiler would help you discover the places where you need to add support for new variants (you just need to follow the compilation errors, after adding a new variant to the sumtype).

With the `else{}`, you will need to do this manually or get runtime errors instead.

Imho keep the match exhaustive and add a separate branch for the new `ExternVar` .

vlib/v/gen/native/amd64.v
```diff
@@ -330,6 +333,9 @@ fn (mut c Amd64) cmp_var_reg(var Var, reg Register, config VarConfig) {
 		GlobalVar {
 			// TODO
 		}
+		else {
+			c.g.n_error('${@LOCATION} unsupported var type ${var}')
+		}
```
spytheman: same - exhaustive matches are better for long term maintenance

vlib/v/gen/native/expr.v
```diff
@@ -174,6 +184,16 @@ fn (mut g Gen) local_var_ident(ident ast.Ident, var LocalVar) {
 	}
 }
 
+fn (mut g Gen) extern_var_ident(var ExternVar) {
+	if g.pref.os == .linux {
+		g.extern_vars[g.pos()] = var.name
+		g.code_gen.mov64(Amd64Register.rax, Number(i64(0)))
```
spytheman:
```suggestion
		g.code_gen.mov64(g.main_reg(), Number(i64(0)))
```

using amd64 names in the non platform specific expr.v file, will make it harder to support other CPUs

vlib/v/gen/native/blacklist.v
```diff
@@ -57,6 +57,36 @@ const whitelist = {
...
+	'string.indent_width':  false
+	'string.index_u8':      false
+	'string.last_index':    true
+	'string.last_index_u8': false
+	'string.contains_u8':   false
+}
+
 fn (g &Gen) is_blacklisted(name string, is_builtin bool) bool {
-	return whitelist[name] or { is_builtin }
+	$if windows {
```
spytheman: Please use `if g.pref.os == .windows {` instead, so that the .exe generation, can be tested more easily on Linux too (through wine; the generation of an executable, can be done with just: `VNATIVE_PE_DLL_PATH=~/Downloads/system32/ ./v -b native -os windows -o x.exe examples/hello_world.v` for example).

Using `$if windows {` in contrast, requires that the backend is compiled as a windows executable as well, and should be avoided inside the compiler, except for blocks, that call platform dependent C. functions (i.e. for code, that can not compile on other platforms).

Issue #23995, opened 2025-03-21
database:  Add connection pool support（go：database/sql）


Avey777:
### Describe the feature

This issue proposes adding a connection pool mechanism to the database module to optimize database interactions. 
﻿
Configurable pool parameters: pool_size, max_connections, idle_timeout
Thread-safe connection reuse with automatic recycling
Health checks (e.g., ping() validation for stale connections)
Extend Connection interface with release() and reset() methods
Add pool lifecycle management (start/stop during database initialization)

refer to :[go:database/sql]( https://github.com/golang/go/blob/master/src/database/sql/sql.go)

### Use Case

![Image](https://github.com/user-attachments/assets/3051c78e-fd79-4a72-b6f2-98b6a81258e5)
[gorm](https://gorm.io/docs/connecting_to_the_database.html)

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

v4.10.0

### Environment details (OS name and version, etc.)

ALL
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Avey777:
@medvednI think this and go language database / sql library, do a unified digital library interface, and provide database connection pool function!
This can be done with database / sql.

Avey777:
@sp
Please do your own thing, and don't comment on any of my statements if you don't have anything to say.
You are not welcome, and the issue is often reopened without being resolved. When people say that the feature tag should not be added, they never think about whether it restricts the function.

Avey777:
```
module main

import db.mysql
import db.pg

type DBconfig = mysql.Config | pg.Config

struct ConnectionPool[T] {
mut:
	connections chan T   // mysql.DB, pg.DB
	config      DBconfig // mysql.Config, pg.Config
}

pub fn new_db_pool2[T](conf DBconfig, size int) !&ConnectionPool[T] {
	for _ in 0 .. size {
		if conf is mysql.Config {
			mysql_conf := conf as mysql.Config
			mut pool_mysql := &ConnectionPool[mysql.DB]{
				connections: chan mysql.DB{cap: size}
				config:      mysql_conf
			}
			for _ in 0 .. size {
				conn := mysql.connect(mysql_conf)!
				pool_mysql.connections <- conn
			}
			return pool_mysql
[...]

PR #24661, opened 2025-06-06
vlib: add pool module

kbkpbot:
Feature request by #23995

Note: This a generic connection pool implementation for database connection, network connection....
To support database pool, it need further enhance in `db.mysql` or other database, to support the `ConnectionPoolable` interface.

```v
// ConnectionPoolable defines the interface for connection objects
pub interface ConnectionPoolable {
mut:
	// validate_conn checks if the connection is still usable
	validate_conn() !bool
	// close_conn terminates the physical connection
	close_conn() !
	// reset_conn returns the connection to initial state for reuse
	reset_conn() !
}
```

This module provides a robust connection pooling implementation for 
managing reusable resources like database connections. It handles 
connection lifecycle, validation, and efficient resource allocation 
with minimal overhead.

### DIFF
--- a/vlib/pool/connection.v
+++ b/vlib/pool/connection.v
@@ -1,0 +1,741 @@
+module pool
+
+import sync
+import sync.stdatomic { new_atomic }
+import time
+
+// Eviction channel capacity
+const eviction_ch_cap = 1000
+
+// ConnectionPoolable defines the interface for connection objects
+pub interface ConnectionPoolable {
+mut:
+	// validate_conn checks if the connection is still usable
+	validate_conn() !bool
+	// close_conn terminates the physical connection
+	close_conn() !
+	// reset_conn returns the connection to initial state for reuse
+	reset_conn() !
+}
+
+// ConnectionPoolConfig holds configuration settings for the connection pool
+@[params]
+pub struct ConnectionPoolConfig {
+pub mut:
+	max_conns          int           = 20               // Maximum allowed connections
+	min_idle_conns     int           = 5                // Minimum idle connections to maintain
+	max_lifetime       time.Duration = time.hour        // Max lifetime of a connection
+	idle_timeout       time.Duration = 30 * time.minute // Time before idle connections are cleaned up
+	get_timeout        time.Duration = 5 * time.second  // Max time to wait for a connection
+	retry_base_delay   time.Duration = 1 * time.second  // Base delay for retry backoff
+	max_retry_delay    time.Duration = 30 * time.second // Maximum delay for retry backoff
+	max_retry_attempts int           = 5                // Maximum retry attempts
+}
+
+// ConnectionWrapper contains metadata about a pooled connection
+struct ConnectionWrapper {
+mut:
+	conn          &ConnectionPoolable // The actual connection object
+	created_at    time.Time           // When connection was created
+	last_used_at  time.Time           // Last time connection was used
+	last_valid_at time.Time           // Last time connection was validated
+	usage_count   int                 // How many times this connection has been used
+}
+
+// EvictionPriority indicates urgency of connection cleanup
+pub enum EvictionPriority {
+	low    // Routine cleanup (connection return)
+	medium // Connection get failure
+	high   // Configuration change
+	urgent // Database/Server recovery
+}
+
+// ConnectionPool manages a pool of reusable connections
+pub struct ConnectionPool {
+mut:
+	config ConnectionPoolConfig
+	// Lock order:
+	// config_mutex > create_mutex > idle_pool_mutex > all_conns_mutex > wait_queue_mutex
+	config_mutex     &sync.RwMutex @[required] // Guards configuration changes
+	create_mutex     &sync.Mutex // Serializes connection creation
+	idle_pool_mutex  &sync.RwMutex              @[required] // Protects idle connections
+	all_conns_mutex  &sync.RwMutex              @[required] // Protects all connections map
+	wait_queue_mutex &sync.RwMutex              @[required] // Protects wait queue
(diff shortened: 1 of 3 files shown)


gechandesu:
I suggest rename the ConnectionPoolable methods, IMO the struct name (that implements interface) should make it clear that the action is performed on the connection. Using the `implements` keyword will make this even more explicit.
```
validate_conn() !bool  -> validate() !bool
close_conn() !         -> close() !
reset_conn() !         -> reset() !
```

einar-hjortdal:
That ConnectionPool looks so complex to me

vlib/pool/connection.v
```diff
@@ -0,0 +1,739 @@
...
+		// Check connection lifetime
+		age := time.since(wrapper.created_at)
+		if age > max_lifetime {
+			// Close expired connection
+			p.all_conns_mutex.lock()
+			p.all_conns.delete(wrapper.conn)
+			p.all_conns_mutex.unlock()
+			wrapper.conn.close() or {}
+			continue
+		}
```
spytheman: That seems cheaper to check, compared to calling `wrapper.conn.validate()`. Should not it be moved above the validation check?

vlib/pool/connection.v
```diff
@@ -0,0 +1,739 @@
...
+		p.idle_pool << wrapper
+		p.all_conns_mutex.lock()
+		p.all_conns[conn] = wrapper
+		p.all_conns_mutex.unlock()
+	}
+
+	// Start background maintenance thread
+	p.cleanup_thread = spawn p.background_maintenance()
+	// Initial connection pruning
+	p.prune_connections()
```
spytheman, line 132: why not do that at the start of .background_maintenance() ?

enghitalo:
Do we have any way to test race conditions?

vlib/pool/connection_test.v
```diff
@@ -0,0 +1,270 @@
...
+	// Modify configuration
+	new_config := pool.ConnectionPoolConfig{
+		max_conns:      5
+		min_idle_conns: 3
+	}
+	p.update_config(new_config)!
+
+	// Trigger idle connection replenishment
+	time.sleep(100 * time.millisecond)
+	assert p.stats().idle_conns >= 3
```
spytheman, line 259: please add an assertion for idle_conns before the update_config call too.

vlib/pool/connection_test.v
```diff
@@ -0,0 +1,270 @@
...
+	defer {
+		p.close()
+	}
+	mut conn := p.get()! as MockConn
+	conn.close_flag = true
+	p.put(conn) or { assert err.msg().contains('simulated close error') }
+
+	// Test reset error handling
+	conn.reset_flag = true
+	p.put(conn) or { assert err.msg().contains('reset') }
```
spytheman, line 288: please add assertions about the pool stats here too (idle, total etc)

vlib/pool/connection_test.v
```diff
@@ -0,0 +1,270 @@
...
+	factory := create_mock_factory(mut test_conns, true, 0)
+
+	mut p := pool.new_connection_pool(factory, pool.ConnectionPoolConfig{})!
+	c := p.get()!
+	p.put(c)!
+
+	p.close()
+
+	// Attempt to acquire connection after close
+	p.get() or { assert err.msg().contains('closed') }
```
spytheman, line 229: please add assertions about the pool stats here as well

enghitalo:
Please, use time.utc instead of time.now, for performance reasons. Is that ok?

vlib/pool/connection.v
kbkpbot: This will make sure when return from `new_connection_pool()`, it has all idle connections ready. If put `p.prune_connections()` at beginning of `background_maintenance()`, we can't make sure this.

vlib/pool/connection.v
kbkpbot: Sure, I will fix this

enghitalo:
> > Do we have any way to test race conditions?
> 
> 1. I ask AI to check for race condition..... (Indeed it can solve some bugs...)
> 2. In the test file, we can loop for many times to ensure some function work as expected.
> 

Usually, valgrid is used to do these tests. I saw that there are some tests using valgrid in V. Maybe @spytheman can help us with that.

spytheman:
Should this be rebased on master, now that a9f4a94 is merged, or are there other things that will need to be fixed before that?

spytheman:
> MSVC generated code has bug...
> 
> ```v
> mut conn := p.get()! as MockConn
> ```

try simplifying it to:
```v
mut tconn := p.get()!
mut conn  := tconn as MockConn
```
to see if it will trigger a non buggy codegen in msvc

spytheman:
Oh, you already did.

spytheman:
The CI passed now for msvc too after the second retry.
The first try was https://github.com/vlang/v/actions/runs/15674428897/job/44151427768 where `connection_test.v` timed out after ~1 hour of waiting.
Perhaps there is a deadlock somewhere?

spytheman:
Yes, it can be skipped with `// vtest build: !msvc` .

spytheman:
It timed out for a tcc-windows job on master too:
https://github.com/vlang/v/actions/runs/15737512081/job/44354089208 .

I'll see how it will behave in the coming days, it may need a !windows build tag 🤔 for now.

spytheman:
It fails sporadically for gcc-windows too: https://github.com/vlang/v/actions/runs/15751959355/attempts/1 and https://github.com/vlang/v/actions/runs/15751959355/attempts/2 . It also failed on the PR that had the WaitGroup fix https://github.com/vlang/v/actions/runs/15752112204/job/44399224811 the first time.

I'll change the build tag to !windows to prevent further false positives for unrelated PRs, until it is fixed for good.

PR #24669, opened 2025-06-07
compiler: implement `expr.remove_para`

Delta456:
A method for `Expr` to get the innermost Expr if there are nested Parentheses. This also decreases the lines of code with the same functionality.


StunxFS:
If I'm not mistaken, this pattern is also used: `if mut expr is ast.ParExpr`

Delta456:
> `if mut expr is ast.ParExpr`

It is used in `parser/parser.v` but we don't need that here

```v
if mut expr is ast.ParExpr {
			expr = expr.expr
		} else {
			p.error('asm in/output must be enclosed in brackets')
			return []
		}
```

spytheman (approved):
Excellent work.

Issue #24671, opened 2025-06-08
jsgen: cannot create maps with enum keys


gchumillas:
### Describe the bug

Cannot create maps with enum keys.

### Reproduction Steps

Write a small `main.v` app:
```v
enum Colors {
	red
	green
	blue
}

fn main() {
	_ := {
		Colors.red: 'red'
		.green: 'green'
		.blue: 'blue'
	}
}
```

Transpile to JS and execute it:
```bash
v -b js main.v
node main.js
```

It outputs the following tautology:
```
undefined is not a function (amen)
```

### Expected Behavior

It should work.

### Current Behavior

It doesn't work.

### V version

V 0.4.10 a5f400e

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 566d22ab66b14655f1b1f2f4aa8396264ebf0c26.a5f400e
|:-------------------|:-------------------
|OS                  |macos, macOS, 15.5, 24F74
|Processor           |8 cpus, 64bit, little endian, Apple M1 Pro
|Memory              |0.22GB/16GB
|                    |
|V executable        |/Users/gonzalo/Projects/Personal/v/vlang/v
|V last modified time|2025-06-08 09:23:28
|                    |
|V home dir          |OK, value: /Users/gonzalo/Projects/Personal/v/vlang
|VMODULES            |OK, value: /Users/gonzalo/.vmodules
|VTMP                |OK, value: /tmp/v_501
|Current working dir |OK, value: /Users/gonzalo/Projects/Personal/v/vlang
|                    |
|Git version         |git version 2.39.1
|V git status        |weekly.2025.23-14-ga5f400ee
|.git/config present |true
|                    |
|cc version          |Apple clang version 17.0.0 (clang-1700.0.13.5)
|gcc version         |Apple clang version 17.0.0 (clang-1700.0.13.5)
|clang version       |Apple clang version 17.0.0 (clang-1700.0.13.5)
|tcc version         |tcc version 0.9.28rc 2024-02-05 HEAD@105d70f7 (AArch64 Darwin)
|tcc git status      |thirdparty-macos-arm64 e447816c
|emcc version        |N/A
|glibc version       |N/A

PR #24673, opened 2025-06-08
jsgen: fix maps are always constructed using string keys (fix #24607 #24671)

### LINKED ISSUE #24607: jsgen: maps are always constructed using string keys
### Describe the bug

Maps are always constructed using string keys, regardless of the original key type (which could be a number).

### Reproduction Steps

Write a small `main.v` app:

```v
fn map_keys[S, T](items map[S]T) []S {
	mut keys := []S{}
	for key, _ in items {
		keys << key
	}
	return keys
}

fn main() {
	items := {
		1: 'one'
		2: 'two'
		3: 'three'
	}

	JS.console.log(items)
	// As you can see, maps are constructed using the strings '1', '2', '3',
	// instead of the corresponding numbers.
	// {
	// 		map: { '1': { str: 'one' }, '2': { str: 'two' }, '3': { str: 'three' } },
	// 		length: 0
	// }

	// fails because maps are constructed using string keys, not number keys
	keys := map_keys(items)
	assert keys.contains(2)

}
```

Compile and execute it:
```bash
v -b js main.v
node main.js
```

Output:
```
{
  map: { '1': { str: 'one' }, '2': { str: 'two' }, '3': { str: 'three' } },
  length: 0
}
main.v:26: FAIL: fn main.main(): assert keys.contains(2)
```

### Expected Behavior

The assertion should not fail.

### Current Behavior

The assertion fails.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.10 d6a2a5e

|V full version      |V 0.4.10 52c7130a2f870afff1242c476aa6db11941abdb9.d6a2a5e
|Memory              |0.56GB/16GB
|                    |
|V executable        |/Users/gonzalo/Projects/Personal/vlang/v
|V last modified time|2025-05-29 16:43:24
|                    |
|V home dir          |OK, value: /Users/gonzalo/Projects/Personal/vlang
|Current working dir |OK, value: /Users/gonzalo/Projects/Personal/vlang
|                    |
|V git status        |weekly.2025.22-13-gd6a2a5e9
|.git/config present |true
|                    |
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

gchumillas:
Fixes #24607 and #24671

The original problem is that map keys do not preserve the prototype methods of the original key values. The trick is to store the original key values, along with their prototype, so the following V code:

```v
m := {
	1: 'one'
	2: 'two'
	3: 'three'
}
```

Is translated to this JS code:

```js
const map = {
	"map": {
		// the `key` property preserves the original key value, along with its prototype
		"1": { "val": { "str": "one" }, "key": { "val": 1 } },
		"2": { "val": { "str": "two" }, "key": { "val": 2 } },
		"3": { "val": { "str": "three" }, "key": { "val": 3 } }
	},
	"length": 3
};
```

Later, we can then use the "key values" for other purposes, such as returning the keys from the map.

### DIFF
--- a/vlib/v/gen/js/tests/map.v
+++ b/vlib/v/gen/js/tests/map.v
@@ -1,6 +1,15 @@
+type T0 = int | string
+type T1 = T0 | rune
+
 struct Point {
 	x f64
 	y f64
+}
+
+enum Colors {
+	red = 1
+	green
+	blue
 }
 
 fn generic_map[T](items map[string]T) []T {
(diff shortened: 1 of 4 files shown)


vlib/v/gen/js/tests/map.v
spytheman, line 1: Please add new test functions, instead of modifying the existing ones, unless there is a bug. That makes reviewing a lot easier.

PR #24676, opened 2025-06-08
jsgen: fix slightly incorrect JS (breaks esbuild) (fix #23711)

### LINKED ISSUE #23711: Slightly incorrect JS (breaks esbuild)
### Describe the bug

Hello!

Basically:

```
PS C:\Users\Ingwie Phoenix\work\v-test> v -skip-unused -backend js -prod .\js_hello_world.v -o hello.js
PS C:\Users\Ingwie Phoenix\work\v-test> esbuild --minify --platform=browser --bundle hello.js --outfile=hello.min.js
✘ [ERROR] Cannot assign to "len" because it is a constant

    hello.js:3010:2:
      3010 │     len = arr.length
           ╵     ~~~

  The symbol "len" was declared a constant here:

    hello.js:3009:8:
      3009 │     const len = new int(new int(0));
           ╵           ~~~

✘ [ERROR] Could not resolve "os"

    hello.js:7:20:
      7 │ const $os = require("os");
        ╵                     ~~~~

  The package "os" wasn't found on the file system but is built into node. Are you trying to bundle
  for node? You can use "--platform=node" to do that, which will remove this error.

2 errors
```

Versions:

```
> v version
V 0.4.9 e9c2314
> esbuild --version
0.25.0
```

Wanted to try and use V to write my frontend instead of TypeScript and just threw some of my favorite tools at the output to see what would happen.

### Reproduction Steps

1. Take `examples/js_hello_world.v` verbatim.
2. Compile using the `js` (also same for `js_browser`) backend to `hello.js`
3. Throw `esbuild` at it.

Commands are above in the snippet.

### Expected Behavior

The browser has neither `require()` nor an `os` module. It just shouldn't be emitted, or a polyfill included.

### Current Behavior

A call to `require()` is made although there shouldn't be one.

Also, as an aside, it attempts to modify a `const`, which it shouldn't. Make it `let` or `var` instead (probably the former)

### Possible Solution

There should be some polyfills that you could use if you needed `os` support. I haven't looked at what exactly the issue is yet, though, nor which APIs are needed.

Got none; just tried stuff out...

### V version

V 0.4.9 e9c2314

Windows 10: `Microsoft Windows [Version 10.0.19045.5440]`

> [!NOTE]
>

gchumillas:
Fixes #23711

How to test in Linux (macos):

```bash
# install esbuild (if not already installed)
npm -g install esbuild

# compile and minimize the code
cd examples
v -skip-unused -backend js_browser -prod ./js_hello_world.v -o hello.js
esbuild --minify --platform=browser --bundle hello.js --outfile=hello.min.js

# run it
node hello.min.js
```

Output:

```
Hello from V.js (0)
Hello from V.js (1)
Hello from V.js (2)
```

### DIFF
--- a/vlib/builtin/js/jsfns.js.v
+++ b/vlib/builtin/js/jsfns.js.v
@@ -122,7 +122,7 @@ fn (v JS.Map) toString() JS.String
 // Hack for "`[]JS.String` is not a struct" when returning arr.length or arr.len
 // TODO: Fix []JS.String not a struct error
 fn native_str_arr_len(arr []JS.String) int {
-	len := 0
+	mut len := 0
 	#len = arr.length
 
 	return len


spytheman:
`-skip-unused` should be on by default now.

spytheman:
I've just checked, and it works on my machine, with latest esbuild as well, while failing on master:
image

spytheman (approved):

Issue #22126, opened 2024-08-28
Test failed on OpenBSD for vlib/net/http/server_test.v


lcheylus:
### Describe the bug

With the latest version of V builded on OpenBSD current/amd64, test fail for `vlib/net/http/server_test.v`: `./v test vlib/net/http/server_test.v`

### Reproduction Steps

- Clone sources via Git on OpenBSD current/amd64
- Build V with clang version 16
- Run test: `./v test vlib/net/http/server_test.v`

### Expected Behavior

Run tests with no error.

### Current Behavior

Test failed for `vlib/net/http/server_test.v`
```bash
$ ./v test vlib/net/http/server_test.v
---- Testing... ----
 FAIL  1436.753 ms vlib/net/http/server_test.v
         retry: 1
      comp_cmd: '/home/fox/dev/vlang.git/v' -skip-running   -o '/tmp/v_1000/tsession_9953d4b96f8_01J6BZ0SKPFMCMSZKJBBN6WWRX/0_0/server_test' '/home/fox/dev/vlang.git/vlib/net/http/server_test.v'
       run_cmd: '/tmp/v_1000/tsession_9953d4b96f8_01J6BZ0SKPFMCMSZKJBBN6WWRX/0_0/server_test'
failure code: 1; foutput.len: 2264; failure output:
---------------------------------------------------------------------------------------------------- retry: 0 ; max_retry: 0 ; r.exit_code: 1 ; trimmed_output.len: 2095
2024-08-28 09:43:25.771097 [INFO ] testsuite_begin
2024-08-28 09:43:25.771641 [WARN ] test_server_stop started
Server handler not set, using debug handler
Listening on http://0.0.0.0:9009/
2024-08-28 09:43:26.306881 [WARN ] test_server_stop finished
2024-08-28 09:43:26.306911 [WARN ] test_server_close started
accept() failed, reason: accept failed; skipping
2024-08-28 09:43:26.366875 [WARN ] test_server_close finished
2024-08-28 09:43:26.366915 [WARN ] test_server_custom_listener started
Server handler not set, using debug handler
accept() failed, reason: accept failed; skipping
2024-08-28 09:43:26.510787 [WARN ] test_server_custom_listener finished
2024-08-28 09:43:26.510814 [WARN ] test_server_custom_handler started
Listening on http://0.0.0.0:18197/
[...]

PR #24698, opened 2025-06-11
net.http: set IP address for HTTP server during tests

lcheylus:
On OpenBSD, TCP connect fails (return EINVAL) if attempted with 0.0.0.0 address.

In `vlib/net/http/server_test.v`, for tests `test_server_custom_handler` and `test_host_header_sent_to_server`, use 127.0.0.1 as IP addr for the HTTP server.

Fix https://github.com/vlang/v/issues/22126

**Tests OK** on OpenBSD/amd64 and Linux/amd64 with tcc, clang and gcc

```bash
$ ./v test vlib/net/http/server_test.v 
 OK    1996.242 ms vlib/net/http/server_test.v
Summary for all V _test.v files: 1 passed, 1 total. Elapsed time: 3590 ms, on 1 job. Comptime: 1585 ms. Runtime: 1996 ms.

$ ./v -cc clang test vlib/net/http/server_test.v 
 OK    2112.866 ms vlib/net/http/server_test.v
Summary for all V _test.v files: 1 passed, 1 total. Elapsed time: 5325 ms, on 1 job. Comptime: 3206 ms. Runtime: 2112 ms.

$ ./v -cc egcc test vlib/net/http/server_test.v 
 OK    2044.330 ms vlib/net/http/server_test.v
Summary for all V _test.v files: 1 passed, 1 total. Elapsed time: 6186 ms, on 1 job. Comptime: 4136 ms. Runtime: 2044 ms.
```

### DIFF
--- a/vlib/net/http/server_test.v
+++ b/vlib/net/http/server_test.v
@@ -121,7 +121,7 @@ fn test_server_custom_handler() {
 	mut server := &http.Server{
 		accept_timeout: atimeout
 		handler:        handler
-		addr:           ':18197'
+		addr:           '127.0.0.1:18197'
 	}
 	t := spawn server.listen_and_serve()
 	server.wait_till_running()!
@@ -261,12 +261,13 @@ fn (mut handler MyCustomHttpHostHandler) handle(req http.Request) http.Response 
 }
 
 fn test_host_header_sent_to_server() {
+	ip := '127.0.0.1'
 	port := 54671
 	log.warn('${@FN} started')
 	defer { log.warn('${@FN} finished') }
 	mut server := &http.Server{
 		handler: MyCustomHttpHostHandler{}
-		addr:    ':${port}'
+		addr:    '${ip}:${port}'
 	}
 	t := spawn server.listen_and_serve()
 	server.wait_till_running()!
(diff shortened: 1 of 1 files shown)


JalonSolov:
Believe it or not, some versions of Windows do not set 127.0.0.1 in their hosts file, so this could break on them.

This change will also be a problem on systems with IP6 as the default/only scheme.

lcheylus:
> Believe it or not, some versions of Windows do not set 127.0.0.1 in their hosts file, so this could break on them.

No issue in Windows CI with my fix :)

> This change will also be a problem on systems with IP6 as the default/only scheme.

This fix is only for tests of HTTP server. I doubt any serious developer would use a system with only an IPv6 stack.

So I think that my fix is not a problem. And it allows that every test runned with `VTEST_JUST_ESSENTIAL` are OK on OpenBSD (proposal to add a specific CI for OpenBSD soon...).

spytheman:
We can make the ip setting conditional on the platform, if there are problems in the future.
Imho it is a good change, since it will allow more checks (an OpenBSD CI), with a small chance of breaking the test for non CI users.

spytheman:
`use_net_and_net_unix_together_test.v` seems to be stuck on the gcc-windows job, but that is unrelated to the PR.

PR #24700, opened 2025-06-11
examples: vanilla web server

enghitalo:


JalonSolov:
Can we get rid of the `src` subdirs?  V will stop supporting that as anything special, so better to not add it to new examples.

enghitalo:
I'm going to close this PR for now, as I won't have time to fix the CI and move the files from the src folder.

PR #23107, opened 2024-12-09
checker: fix veb route method param with non ctx name (fix #23105)

yuyi98:
This PR fix veb route method param with non ctx name (fix #23105).

- Fix veb route method param with non ctx name.
- Add test.

```v
import veb

pub struct User {
pub mut:
	name string
	id   int
}

// Our context struct must embed `veb.Context`!
pub struct Context {
	veb.Context
pub mut:
	// In the context struct we store data that could be different
	// for each request. Like a User struct or a session id
	user       User
	session_id string
}

pub struct App {
pub:
	// In the app struct we store data that should be accessible by all endpoints.
	// For example, a database or configuration values.
	secret_key string
}

// This is how endpoints are defined in veb. This is the index route
pub fn (app &App) index(mut context Context) veb.Result {
	return context.text('Hello V! The secret key is "${app.secret_key}"')
}

fn main() {
	mut app := &App{
		secret_key: 'secret'
	}
	// Pass the App and context type and start the web server on port 8080
	veb.run[App, Context](mut app, 8080)
}

PS D:\Test\v\tt1> v run .
[veb] Running app on http://localhost:8080/
```

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzU2OTQ5MzFlMDYzMmRkMDhlMjcwOTIiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.J5N1UNvvJLlLx4fP2bh61n6zxkssrHUK2JE1ALaQLlk">Huly&reg;: <b>V_0.6-21544</b></a></sub>

PR #24703, opened 2025-06-12
Fix handling of default CorsOptions.allowed_headers

einar-hjortdal:
veb dev forgot about checking for default case

### DIFF
--- a/vlib/veb/middleware.v
+++ b/vlib/veb/middleware.v
@@ -220,12 +220,14 @@ pub fn (options &CorsOptions) set_headers(mut ctx Context) {
 		ctx.set_header(.access_control_allow_credentials, 'true')
 	}
 
-	if options.allowed_headers.len > 0 {
+	if options.allowed_headers.len > 0 && options.allowed_headers != ['*'] { // user-provided list
 		ctx.set_header(.access_control_allow_headers, options.allowed_headers.join(','))
 	} else if _ := ctx.req.header.get(.access_control_request_headers) {
 		// a server must respond with `Access-Control-Allow-Headers` if
 		// `Access-Control-Request-Headers` is present in a preflight request
 		ctx.set_header(.access_control_allow_headers, cors_safelisted_response_headers.join(','))
+	} else if options.allowed_headers == ['*'] { // default
+		ctx.set_header(.access_control_allow_headers, '*')
 	}
 
 	if options.allowed_methods.len > 0 {


einar-hjortdal:
Accidentally fixed another bug while I was at it

spytheman:
Please, add a test, to prevent future silent regressions.

einar-hjortdal:
> Please, add a test, to prevent future silent regressions.

I am on it.
Do you have any idea why `cors_test.v` and `cors_2_test.v` are seemingly identical?

JalonSolov:
Weird... they _are_ identical, except for `ctx` vs `context` in a couple of places... neither would affect the tests.

spytheman:
vlib/veb/tests/cors_2_test.v was added in #23107 , which made sure, that if the `context` is passed explicitly, veb will still work, even though Alex added an edge case for implicitly passing `context` as an experiment (breaking compatibility with older veb apps). 

TLDR: do not mind `cors_2_test.v`, it is unrelated, but needed to prevent regressions.
When you add your test, add it as a new file, or in `cors_test.v`, which uses `mut ctx Context`.

einar-hjortdal:
So, the test is a WIP: I recycled some code from the other test. It looks like the app instances created with a test do not exit fast enough for the next test. However, when running them individually, these tests do pass. I don't really know how to make it work, I would like suggestions or help.

Regarding changes:
I just removed the default allowed_headers `['*']`. The wildcard does not work when requests have credentials, and checking for the default needlessly complicated the logic. If a user really wants the wildcard he can just set it, to me it seems a much more obvious choice than setting a wildcard as default and expecting users to provide an empty array if they do not want a wildcard instead.

Instead of calculating the `cors_safelisted_response_headers.join(',')` at every request, I made it a const. No other code used that.

vlib/veb/tests/cors_regression_test.v
```diff
@@ -0,0 +1,150 @@
+import veb
+import net.http
+import time
+import os
+
+const port = 13012
```
spytheman: Change the port to some other value, that is not used by other potentially concurrent tests, for example:
```suggestion
const port = 53012
```

einar-hjortdal: > Change the port to some other value, that is not used by other potentially concurrent tests, for example:

I can do that, but it doesn't resolve the issue about app instances not exiting fast enough, I don't know how to solve other than give each test a different port, which isn't ideal

einar-hjortdal:
I'm using v compiled with these changes to develop an actual application and it works as expected.

PR #24724, opened 2025-06-15
native: preprocessed vars like EOF

Eliyaan:
Add support for variables like EOF


vlib/v/gen/native/tests/linux.vv
```diff
@@ -25,4 +25,5 @@ fn main() {
 	test_c_extern_vars()
 	flush_stderr()
 	flush_stdout()
+	print_character(`a`)
```
spytheman, line 28: imho move the call before the flush ones

vlib/v/gen/native/gen.v
```diff
@@ -15,6 +15,10 @@ import v.eval
 import term
 import strconv
 
+const c_preprocessed = {
+	'C.EOF': -1
```
spytheman, line 19: hardcoding this does not seem scalable to me, is it just temporary?

Eliyaan: Until we are able to parse macros I do not see another way

Eliyaan: the map will be easily replaceable once we are able to get the macros

JalonSolov: Yes, hard-coding (whether done manually, or with c2v) is the only way to handle this without adding a C preprocessor to V.

Eliyaan: So yes it is temporary

vlib/v/gen/native/tests/linux.vv
Eliyaan: the flush calls are here to test if the calls to builtin works, they are not needed by the prints (but it will move as you wanted in the next pr)

vlib/v/gen/native/gen.v
Delta456: Can you add a TODO note for this? It would help in future.

Issue #24074, opened 2025-03-29
db: Support redis


Avey777:
Redis, as the de facto cached reference library standard, has every reason to be included in the v standard library

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

medvednikov:
I agree.

JalonSolov:
Well, valkey perhaps, with the changes/monetization of redis.

spytheman:
Do you mean a connection wrapper for it, or?

Avey777:
> Do you mean a connection wrapper for it, or?

Yes, same as connecting mysql

Delta456:
We can have `vlang/redis`, aka an official library which can be installed.

Avey777:
> We can have `vlang/redis`, aka an official library which can be installed.

Yes, you can do that.

Jengro777:
https://github.com/vlang/v/discussions/24714

PR #24730, opened 2025-06-16
db: add redis

kbkpbot:
Redis Client for V

### DIFF
--- a/vlib/db/redis/redis_test.v
+++ b/vlib/db/redis/redis_test.v
@@ -1,4 +1,4 @@
-///// vtest build: present_redis?
+// vtest build: present_redis?
 import db.redis
 import time
 
--- a/vlib/db/redis/redis.v
+++ b/vlib/db/redis/redis.v
@@ -364,7 +364,7 @@ fn (mut db DB) read_response_bulk_string() !RedisValue {
 		return error('`read_response_bulk_string()`: invalid data terminator')
 	}
 
-	return data_buf[0..data_length]
+	return data_buf[0..data_length].clone()
 }
 
 // read_response_i64 handles Redis integer responses (format: :<number>\r\n)


vlib/db/redis/redis_test.v
```diff
@@ -0,0 +1,126 @@
+///// vtest build: present_redis?
```
spytheman:
```suggestion
// vtest build: present_redis?
```

The check in `cmd/tools/modules/testing/common.v:835` is just `if line.starts_with('// vtest build:') {`

Avey777:
https://github.com/vlang/v/discussions/24714

einar-hjortdal:
This is a very naive implementation that requires complete rewrite in order to support cluster mode.
My concern is the quality of vlib libraries and the user experience they bring. I don't understand why you'd write this instead of contributing or forking what is already existing and being used.
I also don't understand why this would be merged despite #24074
@spytheman

spytheman:
> This is a very naive implementation

Simple implementations tend to be easier to support and maintain.
Simple and tested implementations even more so.

This happens to be both. I think it is a good addition to vlib, and I do think, that it will be easier to improve it, compared to some of the pre-existing external modules, which were not submitted as PRs here.

> My concern is the quality of vlib libraries and the user experience they bring. 

Mine too, as well as how easy they are to support and extend. I do think that the quality of the code here is high, and that your concerns (cluster mode) should be implemented on a higher level (a connection pool), and that even without it, the client here is usable (Redis is often used without any clustering, which is a relatively advanced feature, and these days a cheap single server can have the resources of several, just 5-10 years ago).

I hope that clarifies my position about why I merged it, and did not wait more, for a hypothetical thirdparty redis library to mature (which had 4-5 years at this point; I contributed to 2 of them, and I've implemented redis clients 3 times outside of V, in other languages) and to be submitted for inclusion in vlib.

spytheman:
As for #24074, you write there:
> Redis is relatively young

Redis started in 2009, which is 16 years ago. It is perhaps young relative to postgresql and mysql, but it had a lot of time to mature, and its protocol (the line based one) and basic usage is simple and not going to change. That is in fact part of why it was used so much - everyone could implement a client for it, and many people did, including me.

Creating a client for mysql/mariadb or postgresql/RDS/yugabyte/CockroachDB/etc is a lot more involved, and imho that is the reason why many choose to not do it from scratch, but instead to link to a library that implements it, and call its APIs instead.

einar-hjortdal:
I understand your position now.
I disagree that this library is usable as it is now, it seems to suffer from the same issues that patrickpissurno/vredis was suffering from.
I wish I could rely on this library instead of having to use mine, because I do not want to maintain my own. I hope this matures fast.

Delta456:
How about you contribute to the library to make it as usable as this one? This will speed up the development.

einar-hjortdal:
> How about you contribute to the library to make it as usable as this one? This will speed up the development.

I have other projects with higher priority. My redict library currently works well. I don't have time to work on yet another library, that's why I would like vlib's library to replace mine eventually.

PR #24734, opened 2025-06-16
cleanup if smartcasts

felipensp:
Cleanup

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -1115,6 +1115,9 @@ fn (mut c Checker) fail_if_immutable(mut expr ast.Expr) (string, token.Pos) {
 			}
 			return '', expr.pos
 		}
+		ast.AsCast {
+			to_lock, pos = c.fail_if_immutable(mut expr.expr)
+		}
 		else {
 			if !expr.is_pure_literal() {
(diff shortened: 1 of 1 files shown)


vlib/v/ast/ast.v
```diff
@@ -2594,10 +2594,8 @@ pub fn (mut lx IndexExpr) recursive_arraymap_set_is_setter() {
 	lx.is_setter = true
 	if mut lx.left is IndexExpr {
 		lx.left.recursive_arraymap_set_is_setter()
-	} else if mut lx.left is SelectorExpr {
-		if mut lx.left.expr is IndexExpr {
-			lx.left.expr.recursive_arraymap_set_is_setter()
-		}
+	} else if mut lx.left is SelectorExpr && lx.left.expr is IndexExpr {
```
spytheman, line 2597: why is `mut` needed for the first, but not the second condition?

felipensp: Probably because `lx.left` is already take as immutable one on second condition (since it internally is taken as `(lx.left as ast.SelectorExpr).expr`.

felipensp: I'll push a fix for when it is used though, currently it crashes the checker if used on second cond.

PR #24737, opened 2025-06-16
tools: make  `v doc -f md module` output useful by default

gechandesu:
This PR makes autogenerated markdown docs useful.

Before this PR `v doc -f md module` without `-comments` produces useless docs without function signatures, struct bodies, etc and actual docstrings. There is no `-comments` flag needed anymore for getting normal docs.

There is example [before](https://gist.github.com/gechandesu/5daeb38eb0fc6b0caefc2ae801869bb3#file-vdoc_net-smtp-md) and [after](https://gist.github.com/gechandesu/5daeb38eb0fc6b0caefc2ae801869bb3#file-vdoc_fixed_net-smtp-md) output.

### DIFF
--- a/cmd/tools/vdoc/markdown.v
+++ b/cmd/tools/vdoc/markdown.v
@@ -28,16 +28,16 @@ fn (vd &VDoc) gen_markdown(d doc.Doc, with_toc bool) string {
 
 fn (vd &VDoc) write_markdown_content(contents []doc.DocNode, mut cw strings.Builder, mut hw strings.Builder,
 	indent int, with_toc bool) {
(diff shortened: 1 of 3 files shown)


spytheman (approved):
Excellent work.

spytheman:
I wonder if it should include the readme as well?
Currently passing `-readme` forces an error:
```
vdoc: Including README.md for doc generation is supported on HTML output, or when running directly in the terminal.
```
... but perhaps it may be possible to just include the content of the readme in the output too, without any further conversion (like the HTML backend does).

PR #24744, opened 2025-06-17
orm: add table attrs; add table/field comment support for mysql and pg

kbkpbot:
1. Add orm table attrs support;
2. Add table/field comment support for mysql & pg.

### DIFF
--- a/vlib/orm/orm.v
+++ b/vlib/orm/orm.v
@@ -91,6 +91,8 @@ pub enum OrderType {
 
 pub enum SQLDialect {
 	default
+	mysql
+	pg
 	sqlite
 }
 
@@ -153,6 +155,12 @@ pub:
 	right    Primitive
 }
 
+pub struct Table {
+pub mut:
+	name  string
+	attrs []VAttribute
+}
+
 pub struct TableField {
 pub mut:
 	name        string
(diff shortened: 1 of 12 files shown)


vlib/orm/orm.v
```diff
@@ -200,11 +208,11 @@ pub mut:
...
-	insert(table string, data QueryData) !
-	update(table string, data QueryData, where QueryData) !
-	delete(table string, where QueryData) !
-	create(table string, fields []TableField) !
-	drop(table string) !
+	insert(table Table, data QueryData) !
+	update(table Table, data QueryData, where QueryData) !
+	delete(table Table, where QueryData) !
+	create(table Table, fields []TableField) !
+	drop(table Table) !
```
spytheman, line 215: That is a breaking change to all clients and drivers of the orm, including the ones that are outside the V repo, like vsql.
Is it needed?

kbkpbot: Yes, I know it. 
But original design did not pass the attributes to `orm`, only pass the struct/table name string.  So `orm` can't get the struct custom attributes except `table name`.
By changing the table name string into a struct `Table`, we can get custom attributes defined on the struct, and this will help future design.

PR #24753, opened 2025-06-19
pool: fix initialization of WaitGroup in test_concurrent_access()

kimshrier:
Creating a new empty WaitGroup structure does not initialize the semaphore structure member.  Calling sync.new_waitgroup() makes sure the proper initialization happens.


spytheman (approved):
Thank you 🙇🏻.

I should have spotted it in the review of the PR 🤦🏻‍♂️.

PR #24754, opened 2025-06-19
thirdparty: add NDEBUG check before glGetErrorCode if-statement in sokol_gfx.h

pisaiah:
Fixes a performance drop caused by https://github.com/vlang/v/commit/37e5ead75d18e0c4d1261b7858f3a6a63c642c9d

When built for release, assert(x) is no-op. Commit 37e5ead, had introduced an if-block & fprintf statement to print the glGetError() code to make diagnostic easier. However, this change will execute even when assert(x) is void.

This commit provides a fix by having an ifdef check for NDEBUG, and if NDEBUG is defined then _SG_GL_CHECK_ERROR will be defined as nothing, as the result of SOKOL_ASSERT/ASSERT, would also be nothing (glibc/assert.h).

Edit:

Video of performance drop (left is after this pull, right is before):

https://github.com/user-attachments/assets/1b71a5d0-dd93-4df5-82ea-49480af40bb5

### DIFF
--- a/thirdparty/sokol/sokol_gfx.h
+++ b/thirdparty/sokol/sokol_gfx.h
@@ -4878,7 +4878,11 @@ inline int sg_append_buffer(sg_buffer buf_id, const sg_range& data) { return sg_
     #endif
     #ifndef _SG_GL_CHECK_ERROR
 	// __v_ start
+	#ifdef NDEBUG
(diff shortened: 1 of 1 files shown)


spytheman:
Sorry for not being careful and testing with -prod too.

PR #24755, opened 2025-06-19
arrays: add reverse_iterator/1 + tests

spytheman:


spytheman:

PR #24769, opened 2025-06-21
builtin: add a rune iterator method to strings, allowing `for for i, r in s.runes_iterator() {` without first allocating an array for all the runes

spytheman:
- **builtin: add a s.runes_iterator() method, allowing for `for i, r in s.runes_iterator() {` with no allocations**
- **fix if_else_return.c.must_have**


Delta456:
When to use `.runes_iterator()` and `.runes()`?

spytheman:
> When to use `.runes_iterator()` and `.runes()`?

If you need to use runes themselves *in an array*, use `.runes()` .

The iterator is useful, when you just need to iterate over all runes in a string, applying some computation on each rune in turn, without allocating an array in advance.

PR #23320, opened 2024-12-30
cgen: fix type_default for option type when the default expr is `none`

felipensp:
Fix #23318

<sub><a href="https://huly.app/guest/vlang-66f40c4d-a476b54c67-771fdd?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJsaW5rSWQiOiI2NzcyZDcyMWYxNDRmNTg1YWU1MTcxY2YiLCJndWVzdCI6InRydWUiLCJlbWFpbCI6IiNndWVzdEBoYy5lbmdpbmVlcmluZyIsIndvcmtzcGFjZSI6InctYWxleGFuZGVyLXZsYW5nLTY2ZjQwYzRkLWE0NzZiNTRjNjctNzcxZmRjIn0.Sjgbkgei8ZEgXo03zz2Onn1IojO5CJVK_5CJjmtiMY8">Huly&reg;: <b>V_0.6-21752</b></a></sub>

PR #24773, opened 2025-06-22
json: remove unnecessary default value for string in tests

lcheylus:
Tests OK for `vlib/json` with `-W` flag
```bash
$ ./v -W test vlib/json/tests/
(...)
Summary for all V _test.v files: 43 passed, 43 total. Elapsed time: 5457 ms, on 7 parallel jobs. Comptime: 35674 ms. Runtime: 368 ms.
```

### DIFF
--- a/vlib/json/tests/json_encode_default_option_none_test.v
+++ b/vlib/json/tests/json_encode_default_option_none_test.v
@@ -3,7 +3,7 @@ module main
 import json
 
 struct Test {
-	id ?string = none
+	id ?string
 }
 
 fn test_main() {


spytheman:
I am not sure what is best here - the test was added as an explicit regression check, for that default value (which previously triggered a cgen error, see #23320).

Deleting the `none` makes the test pointless.
On the other hand, making it compile without a warning is also nice, but serves a smaller goal (clean test suite with -W).

I think that it is best to not remove it for now, until we have better support for custom compilation flags for all tests first, so that we can pass `-w` in tests like this one, and then modify it.

spytheman:
An alternative is to move the test file to `vlib/v/tests/` for now, since it is more of a compiler/cgen related test.

spytheman:
I've implemented support for `// vtest vflags: -w` and I am waiting for my local tests to pass.
It will obsolete the need for this PR.

spytheman:
Done in 91a3844 .

lcheylus:
@spytheman Thanks for the explanations about this test and your commit to add `// vtest vflags: -w` to bypass tests with warnings.

PR #24781, opened 2025-06-23
builtin: add string.expand_tabs()

kbkpbot:
Add string.expand_tabs() replaces tab characters (\t) in the input string with spaces to achieve proper column alignment.
Example: assert 'AB\tHello!'.expand_tabs(4) == 'AB  Hello!'

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -604,11 +604,11 @@ pub fn (s string) normalize_tabs(tab_len int) string {
 // Example: assert 'AB\tHello!'.expand_tabs(4) == 'AB  Hello!'
 pub fn (s string) expand_tabs(tab_len int) string {
 	if tab_len <= 0 {
-		return s // Handle invalid tab length
+		return s.clone() // Handle invalid tab length
 	}
 	mut output := strings.new_builder(s.len)
 	mut column := 0
-	for r in s.runes() {
+	for r in s.runes_iterator() {
 		match r {
 			`\t` {
 				spaces := tab_len - (column % tab_len)
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -600,6 +600,34 @@ pub fn (s string) normalize_tabs(tab_len int) string {
 	return s.replace_char(`\t`, ` `, tab_len)
 }
 
+// expand_tabs replaces tab characters (\t) in the input string with spaces to achieve proper column alignment
+// Example: assert 'AB\tHello!'.expand_tabs(4) == 'AB  Hello!'
+pub fn (s string) expand_tabs(tab_len int) string {
+	if tab_len <= 0 {
+		return s // Handle invalid tab length
+	}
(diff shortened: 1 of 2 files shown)


enghitalo:
I think that some method `replace` already does that

JalonSolov:
Yep.  The `.replace*` methods will do exactly the same thing if all you're doing is replacing TAB with a fixed number of spaces.

`expand_tabs()` does something different, in that it follows TAB stops.  That's why the example only shows a single space in the output... that's all that was needed to get to the next tab stop of every 4 character positions.

vlib/builtin/string.v
```diff
@@ -600,6 +600,34 @@ pub fn (s string) normalize_tabs(tab_len int) string {
...
 
+// expand_tabs replaces tab characters (\t) in the input string with spaces to achieve proper column alignment
+// Example: assert 'AB\tHello!'.expand_tabs(4) == 'AB  Hello!'
+pub fn (s string) expand_tabs(tab_len int) string {
+	if tab_len <= 0 {
+		return s // Handle invalid tab length
+	}
+	mut output := strings.new_builder(s.len)
+	mut column := 0
+	for r in s.runes() {
```
spytheman:
```suggestion
	for r in s.runes_iterator() {
```

vlib/builtin/string.v
```diff
@@ -600,6 +600,34 @@ pub fn (s string) normalize_tabs(tab_len int) string {
 	return s.replace_char(`\t`, ` `, tab_len)
 }
 
+// expand_tabs replaces tab characters (\t) in the input string with spaces to achieve proper column alignment
+// Example: assert 'AB\tHello!'.expand_tabs(4) == 'AB  Hello!'
+pub fn (s string) expand_tabs(tab_len int) string {
+	if tab_len <= 0 {
+		return s // Handle invalid tab length
```
spytheman:
```suggestion
		return s.clone() // Handle invalid tab length
```

vlib/builtin/string_test.v
```diff
@@ -632,7 +632,7 @@ fn main() {
         println(2)
 }'
 	code_trim_right := '
-
+
```
spytheman:
```suggestion
\t
```
relying on invisible quoted tabs/spaces is brittle (some editors strip whitespace at the end of lines, when saving)

vlib/builtin/string_test.v
```diff
@@ -622,7 +622,7 @@ fn test_trim_space() {
 	assert a.trim_space_right() == ' a'
 
 	code := '
-
+
```
spytheman:
```suggestion
\t
```

vlib/builtin/string.v
spytheman: The goal is to have the return string independent from the input/receiver one, which works better with -autofree.

spytheman (approved):
Excellent work.

PR #24788, opened 2025-06-25
sync: add SpinLock

kbkpbot:
SpinLock is a mutual exclusion lock that busy-waits (spins) when locked.
When one thread holds the lock, any other thread attempting to acquire it
will loop repeatedly until the lock becomes available.
[Uses 1 byte of memory for lock state]

### DIFF
--- a/vlib/sync/spinlock_test.v
+++ b/vlib/sync/spinlock_test.v
@@ -1,0 +1,22 @@
+import sync
+
+fn test_spinlock() {
+	mut counter := 0
+	mut s := sync.new_spin_lock()
+	num_threads := 10
+	mut wg := sync.new_waitgroup()
+	wg.add(num_threads)
+
+	for _ in 0 .. num_threads {
+		spawn fn (mut wg sync.WaitGroup, s &sync.SpinLock, counter_ref &int) {
+			defer {
+				s.unlock()
+				wg.done()
+			}
+			s.lock()
+			(*counter_ref)++
+		}(mut wg, s, &counter)
+	}
+	wg.wait()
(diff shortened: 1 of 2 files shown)


spytheman:
The msvc-windows job failure seems unrelated. I'll restart it after the rest are finished.

spytheman:
I am not sure why `v-apps-compile` failed. https://github.com/felipensp/vtcc/tree/stable has not been updated in 3 months, and the changes here should not affect its tests, but apparently, they do somehow 🤔 .

shove70:
Suggested add an "exponential backoff" strategy to reduce meaningless spins in high contention scenarios and reduce contention for bus/memory bandwidth:

```v
@[inline]
pub fn (s &SpinLock) lock() {
	mut expected := u8(0)
	mut spin_count := 0
	max_spins := 100
	base_delay := 100  // nanosecond
	max_delay := 10000 // nanosecond

	for {
		if C.atomic_compare_exchange_weak_byte(&s.locked, &expected, 1) {
			C.atomic_thread_fence(C.memory_order_acquire)
			return
		}
		
		spin_count++
		if spin_count > max_spins {
			delay := min(base_delay * (1 << min(spin_count/max_spins, 10)), max_delay)
			time.sleep(delay * time.nanosecond)
		} else {
			C.cpu_relax()
		}
		
		expected = C.atomic_load_byte(&s.locked)
	}
}
```

felipensp:
> I submit a PR to fix `vtcc` bug at [felipensp/vtcc#5](https://github.com/felipensp/vtcc/pull/5)

Merged, thanks.

PR #24789, opened 2025-06-25
docs: add section for Lambda functions

### LINKED ISSUE #23223: Lamba not documented?
### Describe the issue

I couldn't find lamba usage (See https://github.com/vlang/v/issues/23221) in documentation.

### Links

https://docs.vlang.io/introduction.html

lcheylus:
Fix #23223

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -120,6 +120,7 @@ by using any of the following commands in a terminal:
     * [Mutable arguments](#mutable-arguments)
     * [Variable number of arguments](#variable-number-of-arguments)
     * [Anonymous & higher-order functions](#anonymous--higher-order-functions)
+    * [Lambda expressions](#lambda-expressions)
     * [Closures](#closures)
     * [Parameter evaluation order](#parameter-evaluation-order)
 * [References](#references)
@@ -3021,17 +3022,17 @@ fn main() {
 }
 ```
 
-### Lambda
+### Lambda expressions
 
-V supports lambdas, small anonymous functions, with `||` syntax.
+Lambda expressions in V are small anonymous functions, defined using
+the `|variables| expression` syntax. Note: this syntax is valid only inside calls to higher
+order functions.
 
-```v oksyntax
-// Sort list defining comparator with lambda function
-a.sort(|x, y| x > y)
-
-a := [1, 2, 3]
-b := a.map(|x| x * 10)
-println(b) // prints [10, 20, 30]
+Here are some examples:
+```v
+mut a := [1, 2, 3]
+a.sort(|x, y| x > y) // sorts the array, defining the comparator with a lambda expression
+println(a.map(|x| x * 10)) // prints [30, 20, 10]
 ```
 
 ```v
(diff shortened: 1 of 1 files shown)


doc/docs.md
```diff
@@ -3021,6 +3021,29 @@ fn main() {
...
+
+```v oksyntax
+// Multiply by 10 each element of an array
+a.map(|x| x * 10)
+
+// Sort list defining comparator with lambda function
+a.sort(|x, y| x > y)
+```
+
+```v oksyntax
```
spytheman:
```suggestion
```v
```
the example is self sufficient, and compiles fine

doc/docs.md
```diff
@@ -3021,6 +3021,29 @@ fn main() {
 }
 ```
 
+### Lambda
+
+V supports lambdas, small anonymous functions, with `||` syntax.
+
+```v oksyntax
+// Multiply by 10 each element of an array
+a.map(|x| x * 10)
```
spytheman: if you add `mut a := [1,2,3]` at the start, and `println(a)` at the end, the example would be self sufficient, at the cost of 2 lines more

spytheman: although that can distract from the syntax 🤔

doc/docs.md
```diff
@@ -3021,6 +3021,29 @@ fn main() {
 }
 ```
 
+### Lambda
+
+V supports lambdas, small anonymous functions, with `||` syntax.
```
Delta456:
```suggestion
V supports lambdas, small anonymous functions, with `|captured variable|` syntax.
```

lcheylus: The number of variables in `||` syntax could be > 1. I prefer not to add it here.

doc/docs.md
lcheylus: OK, example modified to be self sufficient.

doc/docs.md
Delta456: I meant `variables`.`||` looks odd to me.

doc/docs.md
```diff
@@ -3021,6 +3021,30 @@ fn main() {
...
+
+```v oksyntax
+// Sort list defining comparator with lambda function
+a.sort(|x, y| x > y)
```
spytheman: this is now out of order, and if you remove the `oksyntax` tag, `v check-md` will fail (or if the user copy/pastes it in the playground/editor).

spytheman: One of the major benefits of self sufficient examples, is that with them you get this:
image
on https://docs.vlang.io/hello-world.html .

spytheman: And then, it is trivial to play with them, which makes learning V easier for newbies.

doc/docs.md
spytheman: I agree with @Delta456 here - V also has a `||` logical operator, and so it can be a bit confusing at first. The example below, will clear the confusion, but imho it is better to not introduce it in the first place.

spytheman:
@lcheylus thank you for taking the initiative to document that aspect of V 🙇🏻‍♂️ .

lcheylus:
@spytheman Thanks to fix my commits about lambda expressions and for your explanations about V examples in docs.

PR #24793, opened 2025-06-25
cgen: add an explicit call to `pthread_exit(ret_ptr);` in the implementation of `res := spawn fn() {}.wait()`

spytheman:


spytheman:
The goal of this is to make `vlib/sync/thread_test.v` pass on recent development versions of macos, see the discussion started at https://discord.com/channels/592103645835821068/592114487759470596/1387456096314331166 . 

Unfortunately I have no access to that environment currently, so this may not be a fix. It may still be better to be explicit, and not rely on the implicit behavior of the pthreads implementation.

spytheman:
The change, makes:
`v -cc clang -cflags -fno-omit-frame-pointer -cflags -fsanitize=address vlib/v/tests/concurrency/thread_wait_ptr_test.v`
very likely to fail (~1 time out of 5).

I am closing the PR for now, until there is a way to verify what the problem is locally.

PR #24798, opened 2025-06-27
sync: fix spin lock, add destroy() and try_lock()

kbkpbot:
1. Bug fix for spin lock expected value reset to unlock before retry.
2. Update test.
3. Add `destroy()`.
4. Add `try_lock()`.
5. Add `valgrind` `ANNOTATE` support  to avoid false error report with spin lock detection. When using with `valgrind`, please add `-d valgrind` to enable this.

Original code(`expected = C.atomic_load_byte(&s.locked)`) will cause the updated test fail sometimes.

### DIFF
--- a/vlib/sync/spinlock_test.v
+++ b/vlib/sync/spinlock_test.v
@@ -28,5 +28,6 @@ fn test_spinlock() {
 		}(mut wg, s, &counter, iterations)
 	}
 	wg.wait()
+	s.destroy()
 	assert counter == num_threads * iterations
 }
--- a/vlib/sync/spinlock_test.v
+++ b/vlib/sync/spinlock_test.v
@@ -1,22 +1,32 @@
 import sync
+import rand
+import time
 
 fn test_spinlock() {
 	mut counter := 0
 	mut s := sync.new_spin_lock()
 	num_threads := 10
+	iterations := 10
 	mut wg := sync.new_waitgroup()
 	wg.add(num_threads)
 
 	for _ in 0 .. num_threads {
-		spawn fn (mut wg sync.WaitGroup, s &sync.SpinLock, counter_ref &int) {
-			defer {
+		spawn fn (mut wg sync.WaitGroup, s &sync.SpinLock, counter_ref &int, iterations int) {
+			for _ in 0 .. iterations {
+				s.lock()
+
+				unsafe {
+					tmp := *counter_ref
+					randval := rand.intn(100) or { 1 }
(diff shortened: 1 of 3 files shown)


vlib/sync/sync.c.v
```diff
@@ -23,24 +23,35 @@ fn should_be_zero(res int) {
 // will loop repeatedly until the lock becomes available.
 pub struct SpinLock {
 mut:
-	locked  u8     // Lock state: 0 = unlocked, 1 = locked
-	padding [63]u8 // Cache line padding (fills to 64 bytes total)
+	locked      u8     // Lock state: 0 = unlocked, 1 = locked
+	initialized bool   // use for Valgrind tracking
```
spytheman: a `bool` can be 4 bytes for some environments; using `u8` is safer, if you want to guarantee that the whole thing will be 64 bytes

spytheman: I think that it would be a good idea for the checks and panics to happen when an uninitialized instance is detected, even if valgrind is not used.

vlib/sync/sync.c.v
```diff
@@ -23,24 +23,35 @@ fn should_be_zero(res int) {
 // will loop repeatedly until the lock becomes available.
 pub struct SpinLock {
```
spytheman, line 25:
```suggestion
@[noinit]
pub struct SpinLock {
```

spytheman: The attribute will prevent from other modules doing `x := sync.Spinlock{}` at compile time.

vlib/sync/sync.c.v
spytheman: Do you think it will impact the performance if the checks for `if !s.initialized {` are left in by default?

vlib/sync/spinlock_test.v
```diff
@@ -1,22 +1,42 @@
...
-		spawn fn (mut wg sync.WaitGroup, s &sync.SpinLock, counter_ref &int) {
-			defer {
+		spawn fn (mut wg sync.WaitGroup, s &sync.SpinLock, counter_ref &int, iterations int) {
+			for _ in 0 .. iterations {
+				s.lock()
+
+				unsafe {
+					tmp := *counter_ref
+					randval := rand.intn(100) or { 1 }
+					time.sleep(randval * time.nanosecond)
```
spytheman, line 24: afaik the granularity of sleep is ~1-2ms on windows

spytheman (approved):
Excellent work.

PR #24805, opened 2025-06-28
scanner: fix multi-level string interpolation in if/match branch

Krchi:
fix multi-level string interpolation in if/match branch, and "fix" match expr type in string interpolation, there still has some codegen error for `assert '${match true { true { '${x}' } else { '${x + 1}' } }}' == '0'`

### DIFF
--- a/vlib/v/scanner/scanner.v
+++ b/vlib/v/scanner/scanner.v
@@ -35,6 +35,8 @@ pub mut:
 	is_inter_end                bool
 	is_enclosed_inter           bool
 	is_nested_enclosed_inter    bool
+	string_count                int
+	is_str_dollor_need_rcbr     []bool = []
 	line_comment                string
 	last_lt                     int = -1 // position of latest <
 	is_print_line_on_error      bool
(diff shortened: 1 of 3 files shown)


vlib/v/scanner/scanner.v
```diff
@@ -35,6 +35,8 @@ pub mut:
 	is_inter_end                bool
 	is_enclosed_inter           bool
 	is_nested_enclosed_inter    bool
+	string_count                int
+	is_str_dollor_need_rcbr     []bool = []
```
spytheman:
```suggestion
	str_dollar_needs_rcbr     []bool = []
```

spytheman: `is_` is usually used as a prefix for bools, not stacks of them, in the V codebase, and the expectation is that you can `if x.is_something {` later

vlib/v/tests/builtin_strings_and_interpolation/string_interpolation_multi_level_test.v
```diff
@@ -0,0 +1,42 @@
+fn foo() ?string {
+	return none
+}
+
+fn test_string_interpolation_multi_level() {
+	x := 0
+
+	// vfmt off
```
spytheman: why?

spytheman: Does vfmt change the code in a way that fails to compiler later, or?

Krchi: > Does vfmt change the code in a way that fails to compiler later, or?

sorry, its my fault, prepare some error test

spytheman:
Excellent work.

PR #24809, opened 2025-06-28
builtin: add min(...), max(...)

kbkpbot:
builtin : add min(...) and max(...)
they are frequently used functions.


JalonSolov:
Only one problem.  @medvednikov has stated "No generics in builtin."  So unless he has changed his mind, this likely won't get merged.

PR #24817, opened 2025-06-30
cgen: fix codegen for multi return assignment with option type

### LINKED ISSUE #24812: Cgen error after optional type cast and tuple result
### Describe the bug

It's not very clear why this code fails, but seems related to a mut ?Type(none) inference type and assignment from a tuple returning function.
Just check the code below

### Reproduction Steps

```v
module main

type Cmd = fn () string

fn cplx() (bool, ?Cmd) {
	return true, stringer
}

fn stringer() string {
	return 'a string'
}

fn main() {
	mut cmd := ?Cmd(none)
	mut truth := false
	truth, cmd = cplx()
	if cmd != none {
		println(cmd())
	}
}
```

### Expected Behavior

Expected console ouput:
```
a string
```

### Current Behavior

cc: /usr/tmp/v_10230/main.01JYXTD0CTC46HN4SARER7EBHQ.tmp.c:5566:26: error: assigning to 'main__Cmd' (aka 'struct string (*)()') from incompatible type '_option_anon_fn___string' (aka 'struct _option_anon_fn___string')
cc:  5566 |         (*(main__Cmd*)cmd.data) = mr_201.arg1;if ((cmd).state != 2) {
cc:       |                                 ^ ~~~~~~~~~~~
cc: 1 error generated.

### Possible Solution

_No response_

This issue arose after trying to mix assigning an existing and new variable from a function returning a tuple via `:=`
Eg:
```v
fn my_tuple() (int, int) {
    return 1, 2
}
mut a := 3
a, b := my_tuple()
```

This would work in go, but since v doesn't allow mixing undeclared and declared variables in a assignment with `:=` either both have to be declared or neither should be.

Thus, in my quest to fix it I came up with declaring using a `?Type(none)` value, which led to the cgen error

### V version

V 0.4.11 bd465b5

Errors on both my phone and laptop. These are the phone v doctor results:
```txt
|V full version      |V 0.4.11 bd465b5254c10ac2ef2d214bc7845e5cebfef946
|OS                  |termux, 4.19.191-g6f453d21fb7f-dirty, #1 SMP PREEMPT Fri Feb 23 18:16:46 CST 2024
|Processor           |8 cpus, 64bit, little endian
|Memory              |N/A
|                    |
|V executable        |/data/data/com.termux/files/home/tmp/v/v
|V last modified time|2025-06-28 12:46:57
|                    |
|V home dir          |OK, value: /data/data/com.termux/files/home/tmp/v
|VMODULES            |OK, value: /data/data/com.termux/files/home/.vmodules
|VTMP                |OK, value: /data/data/com.termux/files/usr/tmp/v_10230
|Current working dir |OK, value: /data/data/com.termux/files/home
|                    |
|Git version         |git version 2.50.0
|V git status        |bd465b52
|.git/config present |true
|                    |
|cc version          |clang version 20.1.7
|gcc version         |clang version 20.1.7
|clang version       |clang version 20.1.7
|tcc version         |N/A
|tcc git status      |thirdparty-unknown-unknown de82a130
|emcc version        |N/A
|glibc version       |N/A
```

felipensp:
Fix #24812

### DIFF
--- a/vlib/v/gen/c/assign.v
+++ b/vlib/v/gen/c/assign.v
@@ -1083,8 +1083,11 @@ fn (mut g Gen) gen_multi_return_assign(node &ast.AssignStmt, return_type ast.Typ
 				'${mr_var_name}.arg${i}'
 			}
 			if mr_types[i].has_flag(.option) {
+				old_left_is_opt := g.left_is_opt
+				g.left_is_opt = true
 				g.expr(lx)
-				g.write(' = ${tmp_var};')
+				g.writeln(' = ${tmp_var};')
+				g.left_is_opt = old_left_is_opt
 			} else {
 				g.write('_option_ok(&(${base_typ}[]) { ${tmp_var} }, (${option_name}*)(&')
 				tmp_left_is_opt := g.left_is_opt
--- a/vlib/v/tests/options/option_fn_multi_ret_test.v
+++ b/vlib/v/tests/options/option_fn_multi_ret_test.v
@@ -1,0 +1,23 @@
+module main
+
+type Cmd = fn () string
+
+fn cplx() (bool, ?Cmd) {
+	return true, stringer
+}
+
+fn stringer() string {
+	return 'a string'
+}
+
+fn test_main() {
+	mut cmd := ?Cmd(none)
+	mut truth := false
+	truth, cmd = cplx()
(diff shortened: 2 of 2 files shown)


spytheman (approved):
Excellent work.

PR #24830, opened 2025-07-02
Implement next() function in veb/context.v

Avey777:


Jengro777:
Please do not merge

spytheman:
Please make experiments in your own fork, without making a PR before you are ready.

Jengro777:
There is no problem with the optimization here, mainly because I still want to add other support.
It doesn't matter if it's closed here, let's talk about it later!
veb We have to rebuild multi-thread support anyway.

Issue #24829, opened 2025-07-02
support sizeof in $if expr


kbkpbot:
### Describe the feature

sometime it need to know the sizeof(T) at compile time.

### Use Case

currently:

```v
pub fn (mut c Counter[T]) increment_by(delta T) T {
	$if T is int {
		if sizeof(int) == 4 {
			return int(C.atomic_fetch_add_u32(&c.value, u32(delta)))
		} else {
			return int(C.atomic_fetch_add_u64(&c.value, u64(delta)))
		}
	}
...
}
```

If we can use `sizeof` at compile time, it can be written as , and there will be no runtime cost:
```v
pub fn (mut c Counter[T]) increment_by(delta T) T {
	$if sizeof(T) == 1 {
		return T(C.atomic_fetch_add_byte(&c.value, delta))
	} $else $if sizeof(T) == 2 {
		return T(C.atomic_fetch_add_u16(&c.value, delta))
	} $else $if sizeof(T) == 4 {
		return T(C.atomic_fetch_add_u32(&c.value, delta))
	} $else $if sizeof(T) == 8 {
		return T(C.atomic_fetch_add_u64(&c.value, delta))
	} 
...
}

```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.11 2c063ea19141163d7ea6a167abc355e7b4341ee4

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 2c063ea19141163d7ea6a167abc355e7b4341ee4
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.2 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |10.35GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-07-01 13:36:27
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/lfrb
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-334-g2c063ea1-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
[...]

PR #24831, opened 2025-07-02
cgen,checker: add  $if sizeof(T) support

kbkpbot:
Feature request by #24829

support:

```v
$if sizeof(T) == 4 {
}

$if sizeof(MyStruct) == 16 {
}
```

todo:

```v
// TODO: support struct.fieldname
x := MyStruct{}

$if sizeof(x.a) == 2 {
}
```

### DIFF
--- a/vlib/v/checker/if.v
+++ b/vlib/v/checker/if.v
@@ -350,52 +350,58 @@ fn (mut c Checker) if_expr(mut node ast.IfExpr) ast.Type {
 							else {}
 						}
 					} else if left is ast.SizeOf && right is ast.IntegerLiteral {
-						s, _ := c.table.type_size(c.unwrap_generic(left.typ))
-						skip_state = match branch.cond.op {
-							.gt {
-								if s > right.val.i64() {
-									ComptimeBranchSkipState.eval
-								} else {
+						// TODO: support struct.fieldname
+						typ := c.unwrap_generic(left.typ)
+						if typ == 0 {
+							c.error('invalid `\$if` condition: expected a type', branch.cond.left.pos())
+						} else {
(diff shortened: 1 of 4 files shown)


vlib/v/gen/c/testdata/comptime_if_sizeof.vv
spytheman, line 1: please add checks for <= and >= too

spytheman: also add a function, where the `assert true` statements are in the else branches (to check the evaluation)

spytheman: I am curious, why is not this a _test.v file, instead of a .vv/.out pair?
To me it seems that it does ordinary asserts, without doing anything special, that would make testing harder?

kbkpbot: > I am curious, why is not this a _test.v file, instead of a .vv/.out pair? To me it seems that it does ordinary asserts, without doing anything special, that would make testing harder?

This may cause compile error on old compiler....

spytheman: How so? The compiler will be bootstrapped for all _test.v files .
If it is able to compile the .vv file, it should be able to also run a _test.v file.

kbkpbot: OK, I changed it to a _test.v file.

spytheman:
In my opinion, file a separate issue for `$if sizeof(x.a) == 2 {` , after this PR is merged.

PR #24837, opened 2025-07-02
math.unsigned: new uint256_new() function, add tests

tankf33der:
While creating module testing, I discovered that this function is very much missing.
Lets create it.

`uint128_new()` already in the code base.


spytheman (approved):
Excellent work.

PR #24839, opened 2025-07-03
datatypes: add lockfree version of counter and ringbuffer

kbkpbot:
Thread-safe high-performance datatypes implemented using atomic operations, currently supporting `counter` and `ringbuffer`.

TBD:

1. Stack
2. Double-Ended Queue (Deque)
3. Concurrent Map
4. ...

### DIFF
--- a/vlib/datatypes/lockfree/README.md
+++ b/vlib/datatypes/lockfree/README.md
@@ -1,0 +1,159 @@
+# Lockfree Library for V
+
+A high-performance, thread-safe collection of lock-free data structures for 
+the V programming language.Designed for concurrent applications requiring 
+low-latency and high-throughput data processing.
+
+## Features
+
+- **Truly Lock-Free**: No mutexes or spinlocks
+- **Cross-Platform**: Works on Windows, Linux, macOS
+- **Memory Safe**: Built-in hazard pointers and safe memory reclamation
+- **Configurable**: Tune parameters for specific workloads
+- **High Performance**: Optimized for modern multi-core processors
+
+## Data Structures
+
+### 1. Atomic Counter
+
+A thread-safe counter with atomic operations.
+
+```v
+import datatypes.lockfree
+
+mut counter := lockfree.new_counter[int](0)
+counter.increment()
+counter.increment_by(5)
+value := counter.get() // 6
+counter.decrement()
+counter.clear()
+```
+
+**Features**:
+- Atomic increment/decrement
+- Batch operations
+- Get current value
+- Reset functionality
+
+### 2. Lock-Free Stack (LIFO)(TBD)
+
+A Last-In-First-Out stack with push/pop operations.
+
+```v ignore
+import datatypes.lockfree
+
+mut stack := lockfree.new_stack[int]()
+stack.push(42)
+stack.push(100)
+top := stack.pop() // 100
+size := stack.size()
+```
+
+**Features**:
+- Push/pop operations
+- Batch push/pop
+- Size approximation
+- Empty check
+- Thread-safe iteration
+
+### 3. Ring Buffer
+
+A circular buffer for producer-consumer scenarios.
+
+```v
+import datatypes.lockfree
+
+mut rb := lockfree.new_ringbuffer[int](1024)
+rb.push(10)
+rb.push(20)
+item := rb.pop() // 10
+free := rb.remaining()
+```
+
+**Features**:
+- Single/Multi producer/consumer modes
+- Blocking/non-blocking operations
+- Batch operations
+- Configurable size
+- Memory efficient
+
+### 4. Double-Ended Queue (Deque)(TBD)
+
+A thread-safe double-ended queue.
+
+```v ignore
+import datatypes.lockfree
+
+mut deque := lockfree.new_deque[string]()
+deque.push_front("first")
+deque.push_back("last")
+front := deque.pop_front() // "first"
+back := deque.pop_back() // "last"
+```
+
+**Features**:
+- Push/pop from both ends
+- Size tracking
+- Empty check
+- Batch operations
+- Iterators
+
+### 5. Concurrent Map(TBD)
+
+A thread-safe key-value store.
+
+```v ignore
+import datatypes.lockfree
+
+mut cmap := lockfree.new_map[string, int]()
+cmap.set("answer", 42)
+value := cmap.get("answer") // 42
+exists := cmap.exists("answer")
+cmap.delete("answer")
+```
+
+**Features**:
+- Get/set/delete operations
+- Key existence check
+- Size approximation
+- Iterators
+- Atomic updates
+
+## Best Practices
+
+1. **Choose the right mode**:
+   - SPSC: Single Producer Single Consumer
+   - MPSC: Multiple Producer Single Consumer
+   - MPMC: Multiple Producer Multiple Consumer
+
+2. **Batch operations**:
+   ```v ignore
+   // More efficient than single pushes
+   stack.push_batch([1, 2, 3, 4, 5])
+   ```
+
+3. **Monitor contention**:
+   ```v ignore
+   stats := stack.stats()
+   if stats.avg_retries > 5.0 {
+       // Consider tuning configuration
+   }
+   ```
+
+4. **Use appropriate backoff**:
+   - Low contention: Minimal backoff
+   - High contention: Exponential backoff
+
+## Acknowledgements
+
+This library incorporates research and design principles from:
+- Intel Threading Building Blocks (TBB)
+- Facebook Folly
+- Java Concurrent Package
+- Dmitry Vyukov's lock-free algorithms
+- DPDK rte ring
+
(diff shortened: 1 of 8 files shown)


vlib/datatypes/lockfree/README.md
```diff
@@ -0,0 +1,159 @@
+# Lockfree Library for V
+
+A high-performance, thread-safe collection of lock-free data structures for 
+the V programming language.Designed for concurrent applications requiring
```
spytheman:
```suggestion
the V programming language. Designed for concurrent applications requiring 
```

vlib/datatypes/lockfree/ringbuffer.v
```diff
@@ -0,0 +1,478 @@
+module lockfree
+
+// copy from vlib/sync/stdatomic/1.declarations.c.v, as we don't import `sync`
```
spytheman: you already did `import sync.stdatomic` in counter.v, so importing it again is  free so to speak in the same module, and will save you some C fn redeclarations

vlib/datatypes/lockfree/README.md
```diff
@@ -0,0 +1,159 @@
...
+   stats := stack.stats()
+   if stats.avg_retries > 5.0 {
+       // Consider tuning configuration
+   }
+   ```
+
+4. **Use appropriate backoff**:
+   - Low contention: Minimal backoff
+   - High contention: Exponential backoff
+
```
spytheman: imho remove that, and add it back with the implementations, when they are done,
to prevent false positives when searching and using AIs

vlib/datatypes/lockfree/ringbuffer.v
kbkpbot: This creates some inconvenience. If I only want to use the `ringbuffer` without the `counter`, directly importing `sync` will trigger warnings about unused imports.

kbkpbot:
```v
import datatypes.lockfree

x := lockfree.new_ringbuf[int](1024)
...
```

kbkpbot: Or we have to do some trick here, something like this:

```v
module lockfree
import sync {Mutex}
...
pub fn new_ringbuffer[T](size u32, param RingBufferParam) &RingBuffer[T] {
   _ := Mutex{}               // This is just use for avoid warning of import but not use `sync`
   ...
}
```

spytheman: I am not talking about `import sync` but about `import sync.stdatomic`, which is smaller, and you already import it.

As for the warning, you can `import sync.stdatomic as _` : 
image

spytheman:
For me, the bench program seems to block at `Testing scenario: MPSC (8P1C)`, tested on Ubuntu 20.04 with both clang-18 and tcc . It uses all my cores (i3 with 2 physical, 4 logical ones), but does not seem to progress. I'll have to test on macos and windows as well.

Edit: it blocked with gcc as well, on the same step, I've waited for ~2 minutes with no progress.
https://github.com/user-attachments/assets/2a6f7863-1ab9-45cb-95ea-23dd7bda8e27

spytheman:
Slightly related: `v -cg vlib/datatypes/lockfree/bench/bench_ringbuffer.v` currently produces a cgen error. I'll try to minimize it and file a separate issue.

spytheman:
I am not sure, if it is a good idea, but this seems to have helped on my machine:
https://github.com/spytheman/v/commit/1cdb493a48e96b0e70e323061bd100f02c6a167b

With it: https://pastebin.com/HqeGnLJQ for tcc and `-gc none` (in order to make it a bit more reliable/easy to debug), TLDR:
```
Performance Results
====================================================================
Scenario                Throughput (M ops/s)    Avg Latency (ns)        CPU Usage (%)
--------------------------------------------------------------------
                SPSC       17.45                          286.57                          0.0
         MPSC (2P1C)       17.28                          289.39                          0.0
         MPSC (4P1C)       16.13                          310.07                          0.0
         MPSC (8P1C)       20.17                          247.86                          0.0
         SPMC (1P2C)       22.18                          225.44                          0.0
         SPMC (1P4C)       10.45                          478.36                          0.0
         MPMC (2P2C)       26.06                          191.84                          0.0
         MPMC (4P4C)       23.74                          210.64                          0.0
====================================================================

Best performance: MPMC (2P2C) with 26.06 M ops/s
```


The negative is that it made the SPSC scenario a lot worse, it was:
```
Testing scenario: SPSC
Completed 1000000 items in 28ms
Completed 1000000 items in 25ms
Completed 1000000 items in 25ms
Completed 1000000 items in 25ms
Completed 1000000 items in 25ms
```
before, and with the patch, it became:
```
Testing scenario: SPSC
Completed 1000000 items in 326ms
Completed 1000000 items in 229ms
Completed 1000000 items in 335ms
Completed 1000000 items in 397ms
Completed 1000000 items in 143ms
```

spytheman:
What does the channels implementation do in those cases? Does it call time.sleep or just busy wait?

JalonSolov:
Can you run the same timing tests against the code you're replacing?  It's cool to see numbers that say how fast the new stuff is, but it would also be good to see the difference between the old & new.

vlib/datatypes/lockfree/ringbuffer.v
kbkpbot: OK, thanks

spytheman:
> And also the max threads is limited to nr_cpus() in benchmark.

That is avoiding the problem, not solving it.

If it can not be solved/mitigated with sleeps as I showed, imho at the very least, it should be documented, so that users can know about it, and act accordingly.

Otherwise there will be situations, where they will hard code some limits, because it is easier, and expect that their product, may not run as optimal on their client systems, if they happen to have less cores (or are running in a virtual environment/hypervisor, that restricts them), but will still run in a reasonable time, and not block and use 100% CPU (which is a reasonable expectation for a module, that claims to be lock free, and thus suitable in more situations than a solution that has manual locks/mutexes around a simpler datastructure to protect it against concurrent access by multiple threads).

I've changed the runtime.nr_cpus() to runtime.nr_jobs(), and I saw personally, that the observed behavior of blocking still exists for me (i3, that has 4 logical cores and 2 physical ones, ran with VJOBS=8), when running `bench_ringbuffer.v` .

Running `bench_channel.v` with VJOBS=8, in contrast, does not exhibit that pathological case. It becomes slower, but it does not stop: `Completed 8000000 items in 11533ms` .

spytheman:
> I think this is solved by introducing a max_waiting_prod_cons , which will limit max allowed waiting producers/consumers before rejecting operations(push/pop).

Excellent.

spytheman:
> I ran some tests and all passed.

I am testing as well, so far it seems good.

spytheman:
I've tested on several machines so far, running macos, linux, windows and freebsd, with tcc, gcc, clang and msvc (only on windows), with VJOBS in 1, 2, 3, 4, 8, 16 (after changing nr_cpus() to nr_jobs(), which I still think should be done, to make such tests easier).

There was not complete blocking, and the performance was nearly always significantly in favor of the lockfree ringbuffer. The only abnormality was on a freebsd vps, with a single virtual CPU, where the channel version of the SPSC was a lot better (`Completed 1000000 items in 385ms`, vs `Completed 1000000 items in 78621ms`). It still progressed though, unlike before.

spytheman (approved):
Excellent work.

spytheman:
@kbkpbot there was one failure for the gcc-windows job, triggered by a timeout in the new ringbuffer test:
https://github.com/vlang/v/actions/runs/16120341173/job/45484257089 .

spytheman:
Another one is https://github.com/vlang/v/actions/runs/16145967313/job/45564784286 , again for gcc-windows .

spytheman:
And https://github.com/vlang/v/actions/runs/16146723811/job/45567429390?pr=24867 , same job gcc-windows, the ringbuffer test lasted for ~90 minutes, and the job timed out.

spytheman:
https://github.com/vlang/v/actions/runs/16246693189/job/45870554088 failed as well, it is also a gcc-windows job.

spytheman:
Another similar CI failure is https://github.com/vlang/v/actions/runs/16268092257/job/45928446968 .

spytheman:
I'll mark it to be skipped for `gcc-windows` for now, to reduce the CI failure rate for independent PRs.

spytheman:
Done in 9db3881 .

spytheman:
It failed for msvc-windows too - https://github.com/vlang/v/actions/runs/16435051920/job/46443413077 . It also failed once yesterday for tcc-windows , but I forgot to copy the URL of the job.

I will add both jobs as skip conditions, to avoid the flakyness affecting other unrelated PRs.

spytheman:
Done in https://github.com/vlang/v/commit/9f0094b0b40c576bd38648c9408344ea0f60d97a .

PR #24840, opened 2025-07-03
math.unsigned: fix lsh() for Uint128, add tests

tankf33der:
The `math.unsigned` module is also raw, untested and therefore buggy.
This patch allowed me to pass a multi-billion test against `gmplib`.
This is the first fix from a future series of corrections.


spytheman:
> The math.unsigned module is also raw, untested and therefore buggy.

I disagree with that qualification, but I welcome the new tests.
```
#0 10:25:24 ^ master ~/v>wc  vlib/math/unsigned/*_test.v
  176   347  4319 vlib/math/unsigned/uint128_test.v
  282   524 10086 vlib/math/unsigned/uint256_test.v
  458   871 14405 total
#0 10:26:09 ^ master ~/v>
```

spytheman (approved):
Excellent work.

PR #24841, opened 2025-07-03
math.unsigned: fix rsh() for Uint128, add tests

tankf33der:
Lets fix rsh() for Uint128.
Multi-billion loop against `libgmp` passed.


spytheman (approved):
Excellent work.

PR #24847, opened 2025-07-05
cgen: fix array data for option array/fixed array(?[]u8/?[3]u8)

Krchi:
1. fix push data in array of option array
2. fix push data in array of option fixed array
3. fix str() for option fixed array


vlib/v/tests/builtin_arrays/array_of_option_fixed_array_test.v
```diff
@@ -0,0 +1,32 @@
+module builtin_arrays
```
spytheman, line 1: there is no need for this module declaration

Delta456 (approved):

PR #24849, opened 2025-07-05
sync: add thread local storage support

kbkpbot:
Add Thread Local Storage(TLS)  support.

### DIFF
--- a/vlib/sync/tls.v
+++ b/vlib/sync/tls.v
@@ -65,7 +65,7 @@ fn convert_t_to_voidptr[T](value T) !voidptr {
 	} $else $if T is rune {
 		f.f_rune = value
 	} $else {
-		return error('Unsupported data type `${T}` to voidptr')
+		return error('Unsupported data type `${T.name}` to voidptr')
 	}
 	return unsafe { f.f_voidptr }
 }
@@ -107,5 +107,5 @@ fn convert_voidptr_to_t[T](value voidptr) !T {
 	} $else $if T is rune {
 		return unsafe { f.f_rune }
 	}
-	return error('Unsupported data type `${T}` from voidptr')
+	return error('Unsupported data type `${T.name}` from voidptr')
 }
--- a/vlib/sync/tls_default.c.v
+++ b/vlib/sync/tls_default.c.v
@@ -10,7 +10,7 @@ fn C.pthread_getspecific(key u64) voidptr
 pub fn new_tls[T](value T) !ThreadLocalStorage[T] {
 	$if T !in [i8, i16, i32, i64, u8, u16, u32, u64, isize, usize, f32, f64, rune, int, voidptr,
 		$pointer] {
-		return error('invalid type ${T}')
+		return error('new_tls: invalid type ${T.name}')
 	}
 	mut key := u64(0)
 	// Validate key allocation
--- a/vlib/sync/tls_test.v
+++ b/vlib/sync/tls_test.v
@@ -5,6 +5,9 @@ import time
 fn test_basic_tls() {
 	println('--- Testing basic datatypes TLS ---')
 
+	// invalid type
+	sync.new_tls[string]('hello world') or { assert err.msg().contains('invalid type') }
+
 	mut tls_i8 := sync.new_tls[i8](-3)!
 	assert tls_i8.get()! == -3
 	tls_i8.set(-4)!
--- a/vlib/sync/tls_windows.c.v
+++ b/vlib/sync/tls_windows.c.v
@@ -10,7 +10,7 @@ fn C.TlsFree(key u32) bool
 pub fn new_tls[T](value T) !ThreadLocalStorage[T] {
 	$if T !in [i8, i16, i32, i64, u8, u16, u32, u64, isize, usize, f32, f64, rune, int, voidptr,
 		$pointer] {
-		return error('error type')
+		return error('new_tls: invalid type ${T.name}')
 	}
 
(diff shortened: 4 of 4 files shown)


JalonSolov:
Heh.  I still wonder why people bother adding `@[inline]`.  It is only a hint to the C compiler that you would like something to be inlined, but it is completely free to ignore it.

It is also completely free to inline things that weren't marked.

So adding the attribute seems to me to be a complete waste of time and space in the file.

spytheman:
> Due to there being too many checks, can we move `CI macOS` to the first check in the CI? This may help us identify macOS bugs quickly.

Afaik there is not an easy way to do that; the ci scripts are independent and run in parallel, and we do not control the github actions scheduler too.
Your best bet afaik if you really want to focus only on it, would be to close the PR (so that it is not using the common V repo job queue), and then in your private fork, turn off manually every other CI job, and leave only the macos one.
(through the control panel here: https://github.com/kbkpbot/v/actions )

vlib/sync/thread_default.c.v
```diff
@@ -13,3 +13,70 @@ fn C.pthread_self() usize
...
 }
+
+// Pthread TLS API functions (via C interface)
+fn C.pthread_key_create(key voidptr, voidptr) int
+fn C.pthread_key_delete(key u64) int
+fn C.pthread_setspecific(key u64, const_ptr voidptr) int
+fn C.pthread_getspecific(key u64) voidptr
+
+// new_tls creates new TLS storage initialized with the given `value`
+pub fn new_tls[T](value T) !&ThreadLocalStorage[T] {
```
spytheman: Why does it return `!&ThreadLocalStorage[T]` instead of `!ThreadLocalStorage[T]` ?
Is there a need for the indirection?

vlib/sync/thread_test.v
```diff
@@ -20,3 +20,55 @@ fn test_sync_thread_id() {
...
 }
+
+fn test_sync_thread_local_storage() {
+	// test create/destroy
+	mut tls_1 := sync.new_tls(1.1)!
+	mut tls_2 := sync.new_tls(333)!
+	mut tls_3 := sync.new_tls(`d`)!
+	dump(tls_1)
+	dump(tls_2)
+	dump(tls_3)
```
spytheman: imho move those new tests into their own file `tls_test.v` .

New functionality like this tends to be flaky at first, and I would prefer to have it separate for now, to monitor more easily their failures.

vlib/sync/thread_test.v
```diff
@@ -20,3 +20,55 @@ fn test_sync_thread_id() {
...
+	p = tls_ptr.get()!
+	assert p == &b
+	tls_ptr.destroy()!
+	if _ := tls_ptr.get() {
+		assert false
+	} else {
+		assert err.msg() == 'get: TLS storage is already destroyed'
+	}
+	tls_ptr.set(&a) or { assert err.msg() == 'set: TLS storage is already destroyed' }
+}
```
spytheman: I did not see an example/test for using the same TLS variable in 2 different threads, where it potentially could be set to different values in each thread, without those values messing the value in the other thread.

Imho that is the primary use/purpose of TLS variables.

spytheman: Can you for example do: `__global tls_var = sync.new_tls(123)`
then in one thread do: `tls_var.set(456)!; assert tls_var.get()! == 456`,
and in another thread do: `tls_var.set(789); assert tls_var.get()! == 789`,
and after the threads are finished, in the main thread: `assert tls_var.get()! == 123` ?

vlib/sync/tls_default.c.v
```diff
@@ -0,0 +1,62 @@
...
+fn C.pthread_key_create(key voidptr, voidptr) int
+fn C.pthread_key_delete(key u64) int
+fn C.pthread_setspecific(key u64, const_ptr voidptr) int
+fn C.pthread_getspecific(key u64) voidptr
+
+// new_tls creates new TLS storage initialized with the given `value`
+pub fn new_tls[T](value T) !ThreadLocalStorage[T] {
+	$if T !in [i8, i16, i32, i64, u8, u16, u32, u64, isize, usize, f32, f64, rune, int, voidptr,
+		$pointer] {
+		return error('error type')
```
Delta456:
```suggestion
		return error('invalid type ${T}')
```

vlib/sync/tls.v
```diff
@@ -0,0 +1,111 @@
...
+	} $else $if T is int {
+		f.f_int = value
+	} $else $if T is f32 {
+		f.f_f32 = value
+	} $else $if T is f64 {
+		f.f_f64 = value
+	} $else $if T is rune {
+		f.f_rune = value
+	} $else {
+		return error('Unsupport data type to voidptr')
```
Delta456:
```suggestion
		return error('Unsupported data type `${T}` to voidptr')
```

vlib/sync/tls.v
```diff
@@ -0,0 +1,111 @@
...
+	} $else $if T is int {
+		return unsafe { f.f_int }
+	} $else $if T is f32 {
+		return unsafe { f.f_f32 }
+	} $else $if T is f64 {
+		return unsafe { f.f_f64 }
+	} $else $if T is rune {
+		return unsafe { f.f_rune }
+	}
+	return error('Unsupport data type from voidptr')
```
Delta456:
```suggestion
	return error('Unsupported data type `${T}` from voidptr')
```

PR #24857, opened 2025-07-06
tests: add union_implementing_interface_test.v

spytheman:
Prevent silent regression of currently not tested functionality discovered in https://discord.com/channels/592103645835821068/784039809459027979/1391493199402504293 , i.e.

Unions can have methods, and they can also use `implements` similar to structs.


larpon (approved):
Nice discovery

PR #24859, opened 2025-07-07
math.unsigned: fix quo_rem(), add tests

tankf33der:
Current `quo_rem()` is broken.
In order not to create new problems, I took the implementation from the [zig](https://github.com/ziglang/zig/blob/master/lib/compiler_rt/udivmod.zig#L92). This is a line by line translation.
Tested with many tests and many ranges against gmplib.


spytheman:
Excellent work.

PR #24865, opened 2025-07-08
math.unsigned: fix rsh() for uint256, add tests

tankf33der:
`rsh()` for uint256 is broken.
The same mechanism as used for `uint128` is used for the fix.
Now passed big loop against `gmplib`.


vlib/math/unsigned/uint256_test.v
```diff
@@ -280,3 +280,14 @@ fn test_separators() {
...
+
+fn test_rsh() {
+	a := unsigned.uint256_from_dec_str('115792089237316195423570985008687907853269984665640564039457584007913129639935')!
+	assert a.str() == a.rsh(0).str()
+	assert '57896044618658097711785492504343953926634992332820282019728792003956564819967' == a.rsh(1).str()
+	assert '6277101735386680763835789423207666416102355444464034512895' == a.rsh(64).str()
+	assert '91343852333181432387730302044767688728495783935' == a.rsh(100).str()
+	assert '340282366920938463463374607431768211455' == a.rsh(128).str()
+	assert '170141183460469231731687303715884105727' == a.rsh(129).str()
+	assert unsigned.uint256_zero == a.rsh(256)
```
spytheman, line 292: A small nitpick: the usual convention in other V tests, is to put the known/expected value on the right, i.e. `assert actual == expected` .

spytheman: It is not a big deal, though - it just changes the focus of the reader on what is tested (the thing after the assert keyword), vs what is already known about it.

tankf33der: i will. thanks.

PR #24867, opened 2025-07-08
math.unsigned: fix lsh() for uint256, add tests

tankf33der:
`lsh()` for uint256 is broken.
The same mechanism as used for uint128 is used for the fix.
Now passed big loop against gmplib.


spytheman:
The CI job failure for gcc-windows is unrelated (it is due to a flaky new test).

PR #24869, opened 2025-07-09
math.unsigned: fix div_128() for uint128, add tests

tankf33der:
Current `div_128()` is broken and cannot even pass this new minimal test.
Took the algorithm that is already added in `quo_rem()` and adapted and extended it for these needed types.
Now passed multi-`B` loop against `gmplib`.
New test run several seconds.


spytheman (approved):
Excellent work.

PR #24872, opened 2025-07-09
math.unsigned: fix rotate_left() for uint256, add test

tankf33der:
Lets fix rotate_left() too.
Without this patch, the current `rotate_left()` cannot pass this test.

Passed `1B` loop against loop `gmplib`.


spytheman (approved):
Excellent work.

spytheman:
Thank you for taking the time to go over the methods in `math.unsigned` methodically, and verify and fix them 🙇🏻‍♂️ @tankf33der .

PR #24875, opened 2025-07-10
fix: Update veb README.md error words

whiter001:
Correcting typos in the document


JalonSolov:
Those `v`s you took out are there for visual markers to illustrate what the text is talking about.  They belong where they are.

If you don't like them there, you'll need to come up with a viable replacement, not just remove them.

PR #24885, opened 2025-07-12
markused: `_option_none` missing

### LINKED ISSUE #24862: C error: json encode with option field without default value
<details>
<summary>V version: V 0.4.11 00e0eae, press to see full `v doctor` output</summary>

|V full version      |V 0.4.11 7dc3889f19e7bdc6fb803ad4d1d65fd9d50c6e0c.00e0eae
|OS                  |linux, Ubuntu 22.04.5 LTS
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz
|Memory              |0.62GB/3.7GB
|                    |
|V executable        |/home/jorge/v/v
|V last modified time|2025-07-07 22:40:19
|                    |
|V home dir          |OK, value: /home/jorge/v
|VMODULES            |OK, value: /home/jorge/.vmodules
|Current working dir |OK, value: /home/jorge/bugs
|                    |
|Git version         |git version 2.34.1
|V git status        |weekly.2025.16-407-g00e0eae9
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
|gcc version         |gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
|clang version       |Ubuntu clang version 14.0.0-1ubuntu1.1
|emcc version        |emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.5 ()
|glibc version       |ldd (Ubuntu GLIBC 2.35-0ubuntu3.10) 2.35

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  250707.v && 250707`
```v
import json

pub struct Struct {
	version ?int// = none
}

fn main() {
	s := Struct{}
	println('${json.encode(s)}')
}
```

**What did you see?**
```
cc: /tmp/v_1000/250707.01JZKGSHCYC0V91WMMQKQFB4EX.tmp.c:2047: warning: implicit declaration of function '_option_none'
cc: /home/jorge/v/thirdparty/tcc/lib/libgc.a: error: 'GC_noop1_ptr' defined twice
cc: tcc: error: undefined symbol '_option_none'
builder error: 
==================
```

Encoded string `{}`

**Workaround**

```v
pub struct Struct {
	version ?int = none
}

```
though we get next warning:
```
250707.v:4:17: warning: unnecessary default value of `none`: struct fields are zeroed by default
    2 | 
    3 | pub struct Struct {
    4 |     version ?int = none
      |                    ~~~~
    5 | }
    6 |
{}

```
> [!NOTE]
>

felipensp:
Fix #24862


vlib/v/tests/skip_unused/struct_field_none.vv
```diff
@@ -0,0 +1,10 @@
+import json
+
+pub struct Struct {
+	version ?int // = none
+}
+
+fn main() {
+	s := Struct{}
+	println('${json.encode(s)}')
```
spytheman, line 9: hm, that is weird ... json.encode() is used here, but in the generated code, the error is inside `json__decode__option_int` which is not used by json.encode, but by json.decode, which is not used by the main program.

spytheman: To me, it seems like the fix is in the wrong place - it will ensure that the `builtin` `_option_none` is always available to cgen, no matter if it is used or not.

felipensp: But we already do the same for `_option_ok` and `_option_clone`.

spytheman: Yes, we do the same for `_option_ok` and `_option_clone` too, and they are not used by most short programs either.

I think, that we should work on improving markused, so that the use of options is tracked, and add those only in cases, when they are used as well.

spytheman: i.e. I can merge this as it is, at the cost of a small increase in cgen output for small programs, but if we tracked the use of options, we could avoid it.

felipensp: Right. 💯

Issue #24882, opened 2025-07-12
[aligned:64] is not work as expected.


kbkpbot:
### Describe the bug

setting a attr `@[aligned:64]` on a struct, it does not create a struct align at 64 address.

### Reproduction Steps

t.v

```v
module main

@[aligned: 64]
struct MyStruct {
        a int
}

fn main() {

        x := u64(voidptr(&MyStruct {}))
        if x % 64 == 0 {
                println('aligned to 64')
        }
        println('x = ${x:016X}')
}

```

### Expected Behavior

output `aligned to 64` and a address

### Current Behavior

only output address, no `aligned to 64`
```
$ ./t
x = 0000763C17AC3FA0
```

### Possible Solution

`memdup` does not honer `aligned` attr, fix `memdup`.

### V version

V 0.4.11 55b1236421e6bc642e8e3bb2d2f5045155d68f7f.b7b1c2e

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 55b1236421e6bc642e8e3bb2d2f5045155d68f7f.b7b1c2e
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.2 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |4.32GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-07-12 12:10:06
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/align
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-371-gb7b1c2e2
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
[...]

PR #24886, opened 2025-07-12
builtin,cgen,markused: add struct [aligned] support

kbkpbot:
Fix issue #24882

This make following example always aligned at 64,

```v
@[aligned: 64]
struct MyStruct {
        a int
}

fn test_struct_aligned() {
        x := u64(voidptr(&MyStruct{}))
        assert x % 64 == 0

        y := MyStruct{}
        ptr := u64(voidptr(&y))
        assert ptr % 64 == 0
}
```

This PR modifies the built-in `malloc()` family of functions. On Windows, all `malloc()` calls now utilize `_aligned_malloc()`, meaning the returned pointers must be freed using `_aligned_free()`.

Therefore, in your codebase:

- **Avoid** directly calling `C.malloc()`, `C.realloc()`, `C.calloc()` or `C.free()`
- **Instead use** the built-in memory management functions: `malloc()`, `v_realloc()`, `vcalloc()`, and `free()`

### DIFF
--- a/vlib/builtin/prealloc.c.v
+++ b/vlib/builtin/prealloc.c.v
@@ -69,7 +69,7 @@ fn vmemory_block_new(prev &VMemoryBlock, align isize, at_least isize) &VMemoryBl
 	}
 
 	fixed_align := if align <= 1 { 1 } else { align }
-	$if msvc {
+	$if windows {
 		v.start = unsafe { C._aligned_malloc(block_size, fixed_align) }
 	} $else {
 		if fixed_align == 1 {
@@ -189,8 +189,8 @@ fn prealloc_vcleanup() {
 	}
 	unsafe {
 		for g_memory_block != 0 {
-			$if msvc {
-				// Warning! On msvc, we always use _aligned_free to free memory.
+			$if windows {
+				// Warning! On windows, we always use _aligned_free to free memory.
 				C._aligned_free(g_memory_block.start)
 			} $else {
 				C.free(g_memory_block.start)
--- a/vlib/builtin/prealloc.c.v
+++ b/vlib/builtin/prealloc.c.v
@@ -37,7 +37,7 @@ fn vmemory_abort_on_nil(p voidptr, bytes isize) {
 }
 
 @[unsafe]
-fn vmemory_block_new(prev &VMemoryBlock, align isize, t_least isize) &VMemoryBlock {
+fn vmemory_block_new(prev &VMemoryBlock, align isize, at_least isize) &VMemoryBlock {
 	vmem_block_size := sizeof(VMemoryBlock)
 	mut v := unsafe { &VMemoryBlock(C.calloc(1, vmem_block_size)) }
 	vmemory_abort_on_nil(v, vmem_block_size)
--- a/vlib/builtin/prealloc.c.v
+++ b/vlib/builtin/prealloc.c.v
@@ -39,7 +39,7 @@ fn vmemory_abort_on_nil(p voidptr, bytes isize) {
 @[unsafe]
 fn vmemory_block_new(prev &VMemoryBlock, align isize, at_least isize) &VMemoryBlock {
 	vmem_block_size := sizeof(VMemoryBlock)
-	mut v := unsafe { &VMemoryBlock(C.calloc(1, vmem_block_size)) }
+	mut v := unsafe { &VMemoryBlock(vcalloc(vmem_block_size)) }
 	vmemory_abort_on_nil(v, vmem_block_size)
 	if unsafe { prev != 0 } {
 		v.id = prev.id + 1
--- a/vlib/builtin/prealloc.c.v
+++ b/vlib/builtin/prealloc.c.v
@@ -69,7 +69,7 @@ fn vmemory_block_new(prev &VMemoryBlock, align isize, t_least isize) &VMemoryBlo
 	}
 
 	fixed_align := if align <= 1 { 1 } else { align }
-	$if windows {
+	$if msvc {
 		v.start = unsafe { C._aligned_malloc(block_size, fixed_align) }
 	} $else {
 		if fixed_align == 1 {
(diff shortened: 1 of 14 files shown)


vlib/v/markused/markused.v
```diff
@@ -61,6 +61,7 @@ pub fn mark_used(mut table ast.Table, mut pref_ pref.Preferences, ast_files []&a
 			'main.main',
 			'init_global_allocator', // needed for linux_bare and wasm_bare
 			'memdup',
+			'memdup_align',
```
spytheman: Can you track, if an aligned struct is used anywhere, and only whitelist `memdup_align`, if such a struct is used?

spytheman: The overhead for simple programs like hello world is currently not very big:
image

vlib/v/tests/structs/struct_aligned_test.v
```diff
@@ -0,0 +1,13 @@
+@[aligned: 512]
+struct MyStruct {
+	a int
+}
+
+fn test_struct_aligned() {
+	x := u64(voidptr(&MyStruct{}))
```
spytheman, line 7: What happens for stack allocated structs that have that attribute, i.e. for `x := MyStruct{}` ? Will their addresses be aligned as well?

spytheman: Or will you get a compile time error?

vlib/builtin/prealloc.c.v
```diff
@@ -37,9 +37,9 @@ fn vmemory_abort_on_nil(p voidptr, bytes isize) {
 }
 
 @[unsafe]
-fn vmemory_block_new(prev &VMemoryBlock, at_least isize) &VMemoryBlock {
+fn vmemory_block_new(prev &VMemoryBlock, align isize, at_least isize) &VMemoryBlock {
 	vmem_block_size := sizeof(VMemoryBlock)
-	mut v := unsafe { &VMemoryBlock(C.calloc(1, vmem_block_size)) }
+	mut v := unsafe { &VMemoryBlock(vcalloc(vmem_block_size)) }
```
spytheman: This should not happen. `vmemory_block_new` is supposed to be a lower level implementation of a bump allocator, so that later multiple calls to malloc() and vcalloc() for smaller blocks, will be amortized to a single call to vmemory_block_new.

spytheman: See the implementation of vcalloc in builtin.c.v, that itself calls prealloc_calloc, which in turn will call vmemory_block_malloc, and that will call potentially vmemory_block_malloc -> you will get unlimited recursion.

vlib/builtin/prealloc.c.v
```diff
@@ -170,28 +189,53 @@ fn prealloc_vcleanup() {
...
+	return unsafe { vmemory_block_malloc(align, n) }
+}
+
+@[unsafe]
+fn prealloc_realloc_align(align isize, old_data &u8, old_size isize, new_size isize) &u8 {
+	new_ptr := unsafe { vmemory_block_malloc(align, new_size) }
+	min_size := if old_size < new_size { old_size } else { new_size }
+	unsafe { C.memcpy(new_ptr, old_data, min_size) }
+	return new_ptr
+}
```
spytheman: this seems unused, you can remove it

vlib/builtin/prealloc.c.v
```diff
@@ -170,28 +189,53 @@ fn prealloc_vcleanup() {
...
+
+@[unsafe]
+fn prealloc_calloc_align(align isize, n isize) &u8 {
```
spytheman: same, this is not used anywhere, remove it

vlib/builtin/prealloc.c.v
```diff
@@ -37,9 +37,9 @@ fn vmemory_abort_on_nil(p voidptr, bytes isize) {
 }
 
 @[unsafe]
-fn vmemory_block_new(prev &VMemoryBlock, at_least isize) &VMemoryBlock {
+fn vmemory_block_new(prev &VMemoryBlock, align isize, at_least isize) &VMemoryBlock {
```
spytheman: why is the new parameter passed in the middle of the older ones, instead of at the end?

vlib/builtin/prealloc.c.v
```diff
@@ -69,15 +88,15 @@ fn vmemory_block_new(prev &VMemoryBlock, at_least isize) &VMemoryBlock {
 }
 
 @[unsafe]
-fn vmemory_block_malloc(n isize) &u8 {
+fn vmemory_block_malloc(align isize, n isize) &u8 {
```
spytheman: Is there a reason to pass it as first parameter, instead of as a last one?
To me, logically, the desired size of the allocation, is more important, compared to the desired alignment 🤔 .

vlib/v/tests/structs/struct_aligned_test.v
kbkpbot: yes, in the `struct_aligned_test.v`, as `y` test this

```v
fn test_struct_aligned() {
        x := u64(voidptr(&MyStruct{}))
        assert x % 512 == 0

        y := MyStruct{}
        ptr := u64(voidptr(&y))
        assert ptr % 512 == 0
}

```

vlib/v/markused/markused.v
kbkpbot: sure, I will add it.

vlib/builtin/prealloc.c.v
kbkpbot: I will remove it

vlib/builtin/prealloc.c.v
kbkpbot: I will fix it

vlib/builtin/prealloc.c.v
kbkpbot: Yes, this some kind of style for linux align memory functions:
```
void *aligned_alloc(size_t alignment, size_t size);
 [[deprecated]] void *memalign(size_t alignment, size_t size);
```

but it can be changed, for Windows it use reverse order:
```
_aligned_malloc(size_t size,size_t alignment);
```

I will change it and place it at last.

vlib/builtin/prealloc.c.v
kbkpbot: ok

vlib/v/markused/markused.v
spytheman: thanks

Issue #21479, opened 2024-05-09
Bignum operations should be optimized


Redysz:
[Issue moved from "discussion" [here](https://github.com/vlang/v/discussions/19291)]
Hi, I'm working with some big numbers operations. In other cases v is extremely faster than Python and close enough for me to C. However, with those big number operations Python is faster than v and it is crushing difference. First - powers:
```v
import time
import math.big

fn main() {
	start := time.now()
	a := big.integer_from_int(9)
	result := a.pow(33420489)
	stop := time.now()
	println("time elapsed: ${(stop-start).seconds()}")
}
```
On my computer it took `227.2s`. Python's equivalent:
```python
from time import perf_counter

def main():
    start = perf_counter()
    result = 9**33420489
    stop = perf_counter()
    print(f"time elapsed: {(stop-start):.2f}")
```
this took only `39.75s`.
I've checked `v`'s `pow` implementation for bigints and I couldn't find place for optimalization. It works as it should (using (x^2)^(n/2) "trick"). So... I checked multiplication big numbers:

```v
import time
import math.big

fn main() {
    a := big.integer_from_int(9)
	first := a.pow(99999)
	second := a.pow(88888)
	start := time.now()
	for _ in 0..1000 {
		result := first * second
	}
	stop := time.now()
	println("time elapsed: ${(stop-start).seconds()}")
}
```
it took `57.8`s, and equivalent for Python:
```python
from time import perf_counter

def main():
    first = 9**99999
    second = 9**88888
    start = perf_counter()
    for _ in range(1000):
        result = first * second
    stop = perf_counter()
    print(f"time elapsed: {(stop-start):.2f}")
```
took only `12`s. Doing the same for `+` operation I can conclude that bigints in Python are just "somehow" implemented with better performance.
I would like to make some "big computation" in `v`.

Relevant implementation for Python: [here](https://github.com/python/cpython/blob/main/Objects/longobject.c) - one can see that multiplication is implemented with Karatsuba algorithm.
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

blackshirt:
[...]

PR #24888, opened 2025-07-12
math.big: fix toom-cook 3-way multiplication

tankf33der:
**Description:**
Karatsuba is already fixed, now it's time for `Toom-Cook-3`.
The current Toom-Cook-3 is broken and although Toom-Cook-3 was correctly implemented and worked correctly if not raising two to large powers for example.

**New thresholds:**
Karatsuba+Toom-Cook-3 are twin brothers with similar characteristics and the threshold values at which they come into operation should be close. If the threshold for Toom-Cook-3 is now lowered below 250, even the current tests will break. The values for the two thresholds were chosen programmatically by a script and now 9 to the power of 33420489 works ~7 times faster and approaches and/or exceeds cpython.

**Toom-Cook-3 fix:**
If the some operand contains all zeros except for the last limb, then when dividing the limb array into three parts, should divide only if not all zeros and leave the default zero value that was already set above.

**New test:**
The range for the test was selected programmatically and works and will work only with the new or lower threshold. A perfect one-line test. Without the patch it will fail.

**Note to followers:** Sounds like a small revolution of multiplication in math.big. Now that a new member has appeared in the multiplication stack, I recommend checking correctness with your own tests.

### DIFF
--- a/vlib/math/big/big_test.v
+++ b/vlib/math/big/big_test.v
@@ -948,3 +948,9 @@ fn test_integer_from_bytes_ignores_potential_leading_zero_bytes() {
 		}
 	}
 }
+
+fn test_pow2_is_power_of_2() {
+	for n in 22000 .. 22010 {
+		assert big.two_int.pow(n).is_power_of_2(), 'pow2: ${n}'
+	}
+}
--- a/vlib/math/big/array_ops.v
+++ b/vlib/math/big/array_ops.v
@@ -108,9 +108,9 @@ fn subtract_digit_array(operand_a []u32, operand_b []u32, mut storage []u32) {
 	shrink_tail_zeros(mut storage)
 }
 
-const karatsuba_multiplication_limit = 240
+const karatsuba_multiplication_limit = 70
 
-const toom3_multiplication_limit = 10_000
+const toom3_multiplication_limit = 360
 
 @[inline]
 fn multiply_digit_array(operand_a []u32, operand_b []u32, mut storage []u32) {
--- a/vlib/math/big/special_array_ops.v
+++ b/vlib/math/big/special_array_ops.v
@@ -164,12 +164,20 @@ fn toom3_multiply_digit_array(operand_a []u32, operand_b []u32, mut storage []u3
 
 	// Slices of a and b
 	a0 := Integer{
-		digits: unsafe { operand_a[0..k] }
-		signum: 1
+		digits: unsafe { operand_a[..k] }
+		signum: if operand_a[..k].all(it == 0) {
+			0
+		} else {
+			1
+		}
 	}
 	a1 := Integer{
 		digits: unsafe { operand_a[k..k2] }
-		signum: 1
+		signum: if operand_a[k..k2].all(it == 0) {
+			0
+		} else {
+			1
+		}
 	}
 	a2 := Integer{
 		digits: unsafe { operand_a[k2..] }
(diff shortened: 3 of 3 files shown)


vlib/math/big/big_test.v
```diff
@@ -948,3 +948,9 @@ fn test_integer_from_bytes_ignores_potential_leading_zero_bytes() {
 		}
 	}
 }
+
+fn test_pow2_is_power_of_2() {
+	for n in 22000 .. 22010 {
+		assert big.two_int.pow(n).is_power_of_2(), 'pow2: ${n}'
+	}
+}
```
spytheman, line 956: that specific test, passes on master too 🤔

spytheman: image

spytheman: It does so on linux as well:
image

spytheman: On the branch of the PR, it is significantly faster:
image

Is this the effect that you wanted to achieve/demonstrate, or are there values for which on master, there are failures, but here, everything passes?

tankf33der: You're right, it's easy to get confused. :) 
I tried to describe everything in the PR, but let me repeat it once more.
Follow the white rabbit. Have fun.

How to break multiplication:
1. master
2. set toom3_multiplication_limit to 250
3. v test .
4. BOOM, fail.
5. add test to big_test.v:
```v
fn test_pow2_is_power_of_2() {
	for n in 22000 .. 22010 {
		assert big.two_int.pow(n).is_power_of_2(), 'pow2: ${n}'
	}
}
```
6. v test .
7. BOOM two times
8. set toom3_multiplication_limit to 360
9. v test .
10. only new fails again
11. add my patch to toom-cook multiplication
12. v test .
13. all passed

**regarding '7 times faster' I take back those words. I was wrong and believed someone else's numbers without verification. Lesson learned.**

Performance steps:
1. lets copy vlang and python code from #21479
2. cat 21479.v
```v
import time
import math.big

fn main() {
        start := time.now()
        a := big.integer_from_int(9)
        _ := a.pow(33420489)
        stop := time.now()
        println("time elapsed: ${(stop-start).seconds()}")
}
```
3. run it
```
pulsar:~# v -prod 21479.v && ./21479 
time elapsed: 33.10089567
pulsar:~# ./21479
time elapsed: 30.323453564
```
4. run it on v0.4.9:
```
pulsar:~# ~/049/v-0.4.9/v -prod 21479.v && ./21479 
time elapsed: 45.513068917
```
5. run it on v0.4.10:
```
pulsar:~# ~/0410/v-0.4.10/v -prod 21479.v && ./21479 
time elapsed: 45.063047533
```
6. python code
```
pulsar:~# python 21479.py 
time elapsed: 33.27
```

During the week I was playing around with graphs for multiplication. Generated this kind of picture. The graph shows a lot about the performance of individual algorithms, not sure if anyone is interested, but if you zoom in, you can see on the blue line when toom-cook multiplication kicks in around 360 limbs.

mul_comparison

spytheman: Yes, I can confirm, that the performance is improved, compared to before:
image

spytheman: The slow_mul_1.py example is currently slower than the version in this PR, but faster than the version on master too:
image

spytheman:
Thanks for working on this. The speedup is significant.

medvednikov:
Indeed. Amazing job.

Issue #24890, opened 2025-07-13
Type Alias with Generics and datatypes causes cgen error


mike-ward:
### Describe the bug

```v
import datatypes

struct Data {
    id string
}

type DataCollection = datatypes.DoublyLinkedList[Data]

fn main() {
    mut data := DataCollection{}
    data.push_back(Data{ id: 'a' })
    for d in data {
        println(d.id)
    }
}
```

### Reproduction Steps

`v run` the program above

### Expected Behavior

Should compile

### Current Behavior

cgen error: `next` method not found

### V version

V 0.4.11 b7b1c2e

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 da3112e.b7b1c2e
|:-------------------|:-------------------
|OS                  |macos, macOS, 15.5, 24F74
|Processor           |8 cpus, 64bit, little endian, Apple M2
|Memory              |0.13GB/8GB
|                    |
|V executable        |/Users/mike/Documents/github/v/v
|V last modified time|2025-07-12 14:37:24
|                    |
|V home dir          |OK, value: /Users/mike/Documents/github/v
|VMODULES            |OK, value: /Users/mike/.vmodules
|VTMP                |OK, value: /tmp/v_501
|Current working dir |OK, value: /Users/mike/Documents/github/bug
|                    |
|env VFLAGS          |"-path @vlib|@vmodules|/Users/mike/Documents/github -message-limit 5 -no-parallel"
|Git version         |git version 2.50.0
|V git status        |weekly.2025.25-112-gb7b1c2e2
|.git/config present |true
|                    |
|cc version          |Apple clang version 17.0.0 (clang-1700.0.13.5)
|gcc version         |Apple clang version 17.0.0 (clang-1700.0.13.5)
|clang version       |Apple clang version 17.0.0 (clang-1700.0.13.5)
|tcc version         |tcc version 0.9.28rc 2024-02-05 HEAD@105d70f7 (AArch64 Darwin)
|tcc git status      |thirdparty-macos-arm64 c8df4e27
|emcc version        |N/A
|glibc version       |N/A

PR #24894, opened 2025-07-13
cgen: allow alias types to be iterated if parent type has next method 

Delta456:
Fixes https://github.com/vlang/v/issues/24890


spytheman (approved):
Excellent work.

spytheman:
What happens if the alias also has its own `fn (mut a AliasTypeIterator) next() ?AliasType {` method defined?

Delta456:
I assume it would use the alias one instead.

Delta456:
> What happens if the alias also has its own `fn (mut a AliasTypeIterator) next() ?AliasType {` method defined?

Haven't checked 😕 

It should probably use the alias-ed method

Delta456:
> What happens if the alias also has its own `fn (mut a AliasTypeIterator) next() ?AliasType {` method defined?

It doesn't work. It uses the parent type `next` method. I will fix this in a while

PR #24897, opened 2025-07-14
runtime: add used_memory()

kbkpbot:
// used_memory retrieves the current physical memory usage of the process.
pub fn used_memory() !u64

### DIFF
--- a/vlib/runtime/used_memory_test.v
+++ b/vlib/runtime/used_memory_test.v
@@ -1,0 +1,19 @@
+import runtime
+
+fn test_used_memory() {
+	used1 := runtime.used_memory()!
+	println('used memory : ${used1}')
+	assert used1 > 0
+
+	_ := []int{len: 100000}
+
+	used2 := runtime.used_memory()!
+	println('used memory : ${used2}')
+	assert used2 > used1
+
+	_ := []int{len: 400000}
+
+	used3 := runtime.used_memory()!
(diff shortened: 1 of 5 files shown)


spytheman (approved):
Excellent work.

enghitalo:
This new feature will be very useful! 👏🏼

vlib/runtime/used_memory_test.v
```diff
@@ -0,0 +1,21 @@
+// vtest build: !musl?
+// test fail for Docker CI / docker-ubuntu-musl
+import runtime
+
+fn test_used_memory() {
+	used1 := runtime.used_memory()!
+	println('used memory 1 : ${used1}')
+
+	_ := unsafe { malloc(4096 * 1024) }
```
spytheman: imho fill the memory block with vmemset, to ensure that it is not just mapped in the address space, but actually used

vlib/runtime/used_memory_test.v
```diff
@@ -0,0 +1,21 @@
...
+
+	used2 := runtime.used_memory()!
+	println('used memory 2 : ${used2}')
+
+	_ := unsafe { malloc(8192 * 1024) }
```
spytheman: same, using vmemset can ensure that the results are more consistent across the different libc implementations/allocators

kbkpbot: ok

PR #24899, opened 2025-07-14
os: fix windows get_error_msg() leak

kbkpbot:
```v
C.FormatMessageW(format_message_allocate_buffer,...)
```

The function allocates a buffer large enough to hold the formatted message, and it must be freed by `C.LocalFree()`.


spytheman (approved):
Excellent work.

PR #24901, opened 2025-07-14
runtime: eliminate allocations done in runtime.used_memory/0 on linux

spytheman:


spytheman:
Tested manually by running:
```v
import os
import time
import runtime

fn main() {
	bsize := 1024 * 1024
	mut blocks := unsafe { [64]voidptr{} }
	for n := 0; true; n++ {
		if n > 0 && n % blocks.len == 0 {
			for i in 0 .. blocks.len {
				unsafe { free(blocks[i]) }
				blocks[i] = unsafe { nil }
			}
			eprintln('done freeing')
		}
		blocks[n % blocks.len] = unsafe { malloc(bsize) }
		unsafe { vmemset(blocks[n % blocks.len], 123, bsize) }
		um := runtime.used_memory()!
		eprintln('> pid: ${os.getpid():9} | n: ${n:6} | um: ${um / 1024:9} KB')
		time.sleep(50 * time.millisecond)
		gc_collect()
	}
}
```

PR #24909, opened 2025-07-15
runtime: add used_memory() implementation for FreeBSD

kimshrier:
Adding a FreeBSD implementation of used_memory() so that the vlib/runtime/used_memory_test.v passes.

### DIFF
--- a/vlib/runtime/used_memory_freebsd.c.v
+++ b/vlib/runtime/used_memory_freebsd.c.v
@@ -1,0 +1,37 @@
+module runtime
+
+import os
+
+#flag -lprocstat
+
+#include <sys/user.h>
+#include <libprocstat.h>
+
+struct C.procstat {}
+
+struct C.kinfo_proc {
+	ki_rssize u64
+}
+
+fn C.procstat_open_sysctl() &C.procstat
+fn C.procstat_close(&C.procstat)
+fn C.procstat_getprocs(&C.procstat, int, int, &u32) &C.kinfo_proc
+
+// used_memory retrieves the current physical memory usage of the process.
+pub fn used_memory() !u64 {
+	mut proc_status := C.procstat_open_sysctl()
+	defer {
+		C.procstat_close(proc_status)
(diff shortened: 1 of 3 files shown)


spytheman:
I propose using `getrusage/2` and the ru_maxrss field, for tcc on freebsd.

It is not ideal (since it does not shrink, when memory is freed), but it is still better than returning 0 or error, and will probably compile fine.

vlib/runtime/used_memory_freebsd.c.v
```diff
@@ -0,0 +1,37 @@
...
+		C.procstat_close(proc_status)
+	}
+
+	mut count := u32(0)
+
+	kip := C.procstat_getprocs(proc_status, C.KERN_PROC_PID | C.KERN_PROC_INC_THREAD,
+		os.getpid(), &count)
+
+	if kip != 0 {
+		return u64(kip.ki_rssize)
```
spytheman: ki_rssize is in pages, the expected returned value is in bytes

kimshrier: I'll correct that.

kimshrier:
I initially used getrusage but it didn't look like it was working properly.  I'll give it another try.

spytheman:
I have a version that does produce this:

image

getrusage behaves weirdly indeed - it sometimes returns 0 in ru_maxrss, and sometimes the reported size does not grow, even after allocation and writing was done 🤔 


edit: forgot the link to the change - https://github.com/spytheman/v/commit/43e1fb97676d58942fd7d95688249df852c0a975

kimshrier:
It's rather late and I have a ton of work to do tomorrow at my day job.  I'll look into what is going on with getrusage after I finish work tomorrow.

spytheman:
I can push the tcc fix and merge the PR, to make the test pass on freebsd if you want, for now?

kimshrier:
Go ahead and apply your fix.

PR #24912, opened 2025-07-16
cgen: move closure code to vlib

kbkpbot:
This PR:

1. move some closure codes from `cgen` to `vlib/builtin`.
2. add a benchmark of closure in `vlib/bench/bench_closure.v`

TODO:
freestanding can't work now because lack of `pthread_mutex`


vlib/builtin/closure.v
spytheman, line 1: I am against this being directly in `builtin`, since it will affect even programs that do not use closures at all.

It could be in a `builtin.closure` module instead, that gets imported selectively, similar to how `sync` is imported for channels.

spytheman: It will also solve the problem of more exotic environments breaking (like freestanding etc).

spytheman:
Interesting ... why is there a diff for the `Small Closure Call` case?

vlib/builtin/closure.v
kbkpbot: By moving it in `builtin.closure`, I encounter some errors:

```
 FAIL  [  35/2449] C:  1512.3 ms, R:     0.000 ms vlib/builtin/js/array_test.js.v
>> compilation failed:
builder error: error: import cycle detected between the following modules:
 * builtin -> builtin.closure -> builtin
 * main -> builtin -> builtin.closure -> builtin
 * builtin.closure -> builtin -> builtin.closure
 * os -> builtin -> builtin.closure -> builtin
 * strings.textscanner -> builtin -> builtin.closure -> builtin

 FAIL  [  36/2449] C:  1420.1 ms, R:     0.000 ms vlib/builtin/js/string_test.js.v
>> compilation failed:
builder error: error: import cycle detected between the following modules:
 * builtin -> builtin.closure -> builtin
 * main -> builtin -> builtin.closure -> builtin
 * builtin.closure -> builtin -> builtin.closure
 * os -> builtin -> builtin.closure -> builtin
 * strings.textscanner -> builtin -> builtin.closure -> builtin

 OK    [  37/2449] C:   214.7 ms, R:     1.441 ms vlib/builtin/wchar/wchar_test.v
 FAIL  [  38/2449] C:  1542.9 ms, R:     0.000 ms vlib/builtin/js/int_test.js.v
>> compilation failed:
builder error: error: import cycle detected between the following modules:
 * builtin -> builtin.closure -> builtin
 * main -> builtin -> builtin.closure -> builtin
 * builtin.closure -> builtin -> builtin.closure
 * os -> builtin -> builtin.closure -> builtin
 * strings.textscanner -> builtin -> builtin.closure -> builtin

 OK    [  39/2449] C:   244.9 ms, R:     1.559 ms vlib/cli/command_test.v

```

debugging.......

spytheman: try adding `builtin.closure` in `builtin_module_parts` in `vlib/v/util/util.v:17`

spytheman: although that may need adding that change separately in a single PR, so that the compiler can bootstrap later with it on master

vlib/builtin/closure/closure.v
spytheman, line 1: is the file needed if it is empty?

kbkpbot: It seems there is some bug of js builder, if remove `closure.v`, then error :
```
 FAIL  [  35/2461] C:  1502.9 ms, R:     0.000 ms vlib/builtin/js/array_test.js.v
>> compilation failed:
vlib/builtin/js/string.js.v:72:36: builder error: cannot import module "builtin.closure" (no .v files in "/media/HD/github/kbkpbot/v/vlib/builtin/closure")
   70 |
   71 | pub fn (s string) split(dot string) []string {
   72 |     tmparr := s.str.split(dot.str).map(fn (it JS.Any) JS.Any {
      |                                       ^
   73 |         res := ''
   74 |         #res.str = it

 FAIL  [  36/2461] C:  1517.1 ms, R:     0.000 ms vlib/builtin/js/int_test.js.v
>> compilation failed:
vlib/builtin/js/string.js.v:72:36: builder error: cannot import module "builtin.closure" (no .v files in "/media/HD/github/kbkpbot/v/vlib/builtin/closure")
   70 |
   71 | pub fn (s string) split(dot string) []string {
   72 |     tmparr := s.str.split(dot.str).map(fn (it JS.Any) JS.Any {
      |                                       ^
   73 |         res := ''
   74 |         #res.str = it

 FAIL  [  37/2461] C:  1384.5 ms, R:     0.000 ms vlib/builtin/js/string_test.js.v

```

kbkpbot: or it need some check, when language is not `.c`, then will not import builtin.closure.

kbkpbot: js gen need at least one .v under the module dir...

spytheman: Oof, that is definitely a problem. Thanks for checking 🙇🏻‍♂️ .

JalonSolov: Maybe name it `closure.js.v`, so only the js builds will even look at it?

vlib/builtin/closure/closure.c.v
```diff
@@ -0,0 +1,241 @@
...
+	read_write
+}
+
+// refer to https://godbolt.org/z/r7P3EYv6c for a complete assembly
+// vfmt off
+pub const closure_thunk = $if amd64 {
+    [
+    u8(0xF3), 0x44, 0x0F, 0x7E, 0x3D, 0xF7, 0xBF, 0xFF, 0xFF,  // movq  xmm15, QWORD PTR [rip - userdata]
+        0xFF, 0x25, 0xF9, 0xBF, 0xFF, 0xFF                     // jmp  QWORD PTR [rip - fn]
+    ]
```
spytheman, line 32: would a fixed array like [...]! work here? 

edit: probably not, because the sizes for the different architectures are different 🤔

vlib/v/parser/expr.v
```diff
@@ -476,6 +476,9 @@ fn (mut p Parser) check_expr(precedence int) !ast.Expr {
 			} else {
 				// Anonymous function
 				node = p.anon_fn()
+				if p.file_backend_mode == .v || p.file_backend_mode == .c {
+					p.register_auto_import('builtin.closure')
+				}
```
spytheman, line 481: not all anonymous functions are closures ... many of them do not capture anything

spytheman: we should probably make this more specific in the future 🤔

vlib/v/parser/parser.v
```diff
@@ -2007,6 +2007,9 @@ fn (mut p Parser) dot_expr(left ast.Expr) ast.Expr {
 	}
 	is_filter := field_name in ['filter', 'map', 'any', 'all', 'count']
 	if is_filter || field_name == 'sort' || field_name == 'sorted' {
+		if p.file_backend_mode == .v || p.file_backend_mode == .c {
+			p.register_auto_import('builtin.closure')
+		}
```
spytheman, line 2012: I am not sure about this either ... the parameter may be a simple anonymous function, that does not capture anything and does not need allocating thunks etc.

spytheman (approved):

vlib/v/parser/struct.v
```diff
@@ -284,6 +284,22 @@ fn (mut p Parser) struct_decl(is_anon bool) ast.StructDecl {
 					// error is set in parse_type
 					return ast.StructDecl{}
 				}
+
+				// for field_name []fn, cgen will generate closure, so detect here
```
spytheman, line 288: why would cgen generate a closure for a field init?
afaik it only does it for stuff like `field: instance.method` , not for normal functions, whether anonymous or not

spytheman (approved):
Excellent work.

PR #24913, opened 2025-07-16
runtime: add note for the availability of the used_memory implementation

lcheylus:
`used_memory` function available only on Darwin, Linux and Windows. Otherwise returns `used_memory: not implemented`.

### DIFF
--- a/vlib/runtime/used_memory_default.c.v
+++ b/vlib/runtime/used_memory_default.c.v
@@ -1,7 +1,7 @@
 module runtime
 
 // used_memory retrieves the current physical memory usage of the process.
-// Note: implementation available only on Darwin, Linux and Windows. Otherwise,
+// Note: implementation available only on macOS, Linux and Windows. Otherwise,
 // returns 'used_memory: not implemented'.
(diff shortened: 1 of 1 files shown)


JalonSolov:
FYI - #24909 is adding support for FreeBSD.

vlib/runtime/used_memory_default.c.v
```diff
@@ -1,6 +1,8 @@
 module runtime
 
 // used_memory retrieves the current physical memory usage of the process.
+// Note: implementation available only on Darwin, Linux and Windows. Otherwise,
```
spytheman:
```suggestion
// Note: implementation available only on macOS, Linux and Windows. Otherwise,
```

spytheman: The name of the exact kernel is only slightly related; the popular name for the OS/platform is more important for the callers imho.

spytheman:
> FYI - #24909 is adding support for FreeBSD.

We can update the doc comment before merging the other PR. This one reflects the current state.

PR #24937, opened 2025-07-20
math.big: add shrink zeros to karatsuba and toom-cook

tankf33der:
Next stage of fighting with zeros in Karatsuba and Toom-Cook.
Zeros can be not only at the end, but also in the middle of the array.
Before the limb arrays are passed further down the computation stack, they must be (!) normalized.

The test will not pass now on `master` without the patch and the values were obtained programmatically.


spytheman (approved):
Thank you 🙇🏻 .
Excellent work.

PR #24944, opened 2025-07-22
builtin: extract cfns_windows.c.v from cfns.c.v

spytheman:


spytheman:
Constructs such as this one defeat the purpose of the extraction:
```v
    res := $if windows {
        C.closesocket(handle)
    } $else {
        C.close(handle)
    }
```

(since *both* branches are checked, and that can not be done properly, if `C.closesocket` is not already known, and it will not be, because the cfns_windows.c.v file will not be parsed on non windows hosts)

PR #24954, opened 2025-07-23
slow_tests: fix crash on clang for asm_test.amd64.v

tankf33der:
Currently, this test crashes on `clang`.
This needs to be fixed by proactively specifying that the `eax` register is modified.
It worked on `gcc` as is, but clang requires an explicit specification.


spytheman (approved):
Thank you.
I wonder why was not it tested/found to fail on the existing CI jobs.

PR #24955, opened 2025-07-23
eval: reset return value after const eval

kbkpbot:
eval: reset return value after const eval.
This PR will reset the return values after eval a const statement.

### DIFF
--- a/vlib/v/eval/expr.c.v
+++ b/vlib/v/eval/expr.c.v
@@ -150,7 +150,12 @@ pub fn (mut e Eval) expr(expr ast.Expr, expecting ast.Type) Object {
 						e.returning = true
 						e.return_values = []
 						e.stmts(branch.stmts)
-						return e.return_values[0]
+						val := if e.return_values.len > 0 {
+							e.return_values[e.return_values.len - 1]
+						} else {
+							empty
+						}
+						return val
 					}
 				}
 				return empty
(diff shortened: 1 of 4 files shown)


vlib/v/eval/tests/const_test.v
spytheman, line 1: This test passes on master too.

spytheman (approved):

vlib/v/eval/tests/const_test.v
kbkpbot: ye, I updated the if-expr also, so it should support both comptime-if-expr and if-expr now.

vlib/v/eval/expr.c.v
```diff
@@ -145,22 +145,31 @@ pub fn (mut e Eval) expr(expr ast.Expr, expecting ast.Type) Object {
...
 			for i, branch in expr.branches {
-				result := if expr.is_comptime {
-					e.comptime_cond(branch.cond)
-				} else {
-					e.expr(branch.cond, ast.bool_type_idx) as bool
+				mut result := false
+				if expr.is_comptime {
+					result = e.comptime_cond(branch.cond)
+				} else if expr.branches.len != i + 1 {
+					result = e.expr(branch.cond, ast.bool_type_idx) as bool
```
JalonSolov: How about this, so you don't need to make `result` mutable?

```suggestion
				result := if expr.is_comptime {
					result = e.comptime_cond(branch.cond)
				} else if expr.branches.len != i + 1 {
					result = e.expr(branch.cond, ast.bool_type_idx) as bool
				} else {
					false
```

spytheman:
(the msvc bootstrap errors are fixed on master and are not related to this PR)

PR #24959, opened 2025-07-23
markused: some cleanups and improvements

felipensp:


vlib/v/markused/markused.v
```diff
@@ -63,11 +63,13 @@ pub fn mark_used(mut table ast.Table, mut pref_ pref.Preferences, ast_files []&a
...
 			'tos2',
 			'error',
-			'builtin_init',
 			'fast_string_eq',
 			'println',
 			'ptr_str',
 		]
+		if ast_files[ast_files.len - 1].imports.len > 0 {
+			core_fns << 'builtin_init'
+		}
```
spytheman, line 72: That heuristic seems a bit hacky - you could have a freestanding program that imports its own modules, that do not use `builtin` too.

Doing a loop over ast_files and checking if at least one .mod == 'builtin' may be more reliable.

spytheman (approved):
Excellent work.
Thanks.

PR #24960, opened 2025-07-23
regex: fix OOB bug discovered through `./v -g -force-bounds-checking test vlib/regex/`

spytheman:
- **regex: fix OOB bug discovered through `./v -g -force-bounds-checking vlib/regex/`**


penguindark:
OK, I had a look at this, your solution can work but to avoid further problems I suggest to replace `re.group_stack.clear()` at line 356 with:
```v
// restore initial state of the stack
for mut x in re.group_stack {
    x = -1
}
```

PR #24963, opened 2025-07-24
markused: add `@[markused]` for interface

felipensp:


vlib/v/tests/skip_unused/interface_struct_unused.vv
spytheman, line 1: That compiles on master too.
Imho add a .out/.c.must_have pair in `vlib/v/gen/c/testdata/` , that can be tested with `vlib/v/gen/c/coutput_test.v` .

PR #24965, opened 2025-07-24
eval:add more infix op support; fix early func return

kbkpbot:
This PR:

1. add support for `&&`, `||`, `>=`, `<=`,`!`;
2. comptime infix expr support;
3. early function return.

### DIFF
--- a/vlib/v/eval/eval.v
+++ b/vlib/v/eval/eval.v
@@ -64,6 +64,7 @@ pub mut:
 	scope_idx              int      // this is increased when e.open_scope() is called, decreased when e.close_scope() (and all variables with that scope level deleted)
 	returning              bool
 	return_values          []Object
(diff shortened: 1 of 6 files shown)


vlib/v/eval/eval.v
```diff
@@ -64,6 +64,7 @@ pub mut:
 	scope_idx              int      // this is increased when e.open_scope() is called, decreased when e.close_scope() (and all variables with that scope level deleted)
 	returning              bool
 	return_values          []Object
+	execute_return_stmt    bool // already execute a retrun stmt in func or not
```
JalonSolov:
```suggestion
	execute_return_stmt    bool // already execute a return stmt in func or not
```
:-)

spytheman (approved):
Excellent work.
Thank you 🙇🏻.

PR #24972, opened 2025-07-25
eval: fix if-else; add infix op; fix func calls

kbkpbot:
This PR:

1. fix if-else bug;
2. add more infix op support : `&`, `|`, `^`, `%`,`~`;
3. fix func call return, only return require results.

### DIFF
--- a/vlib/v/eval/stmt.v
+++ b/vlib/v/eval/stmt.v
@@ -72,6 +72,10 @@ pub fn (mut e Eval) stmt(stmt ast.Stmt) {
 			for i, expr in stmt.exprs {
 				e.return_values << e.expr(expr, stmt.types[i])
 			}
+			if e.return_values.len > stmt.exprs.len {
+				// keep only results
+				e.return_values.drop(e.return_values.len - stmt.exprs.len)
+			}
 			e.returning = old_returning
 		}
 		ast.ForInStmt {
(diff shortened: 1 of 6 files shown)


vlib/v/eval/stmt.v
```diff
@@ -72,6 +72,10 @@ pub fn (mut e Eval) stmt(stmt ast.Stmt) {
 			for i, expr in stmt.exprs {
 				e.return_values << e.expr(expr, stmt.types[i])
 			}
+			if e.return_values.len > stmt.exprs.len {
+				// keep only results
+				e.return_values.drop(e.return_values.len - stmt.exprs.len)
```
spytheman, line 77: hm, should not that be an error?

spytheman (approved):
Excellent work.

vlib/v/eval/stmt.v
kbkpbot: For `func_call_test.v`,  if has no this workaround, then the final `ret` would contains two results.

```v
import v.eval

fn test_func_return() {
        mut e := eval.create()

        ret := e.run('
        fn sub(a int) int {
                if a > 105 {
                        return 101
                }
                return 5151
        }

        fn key(b int) int {
                println(b)
                if b > 100 {
                        return 3 + sub(sub(sub(b)))
                }
                return 7171
        }

        key(110)')!

        dump(ret)
        assert ret[0].int_val() == 104
}

```

```
[func_call_test.v:24] ret: [eval.Object(101), eval.Object(eval.Int{
    val: 104
    size: 64
})]
```

We don't need the first result : `101`.

spytheman: Where is the first 101 coming from?

spytheman: I would have to study the code in more details, but on the surface, it looks like this is masking a deeper issue.

PR #24981, opened 2025-07-26
v: support `-div-by-zero-is-zero`, which allows for `x / 0 == 0, x % 0 == x`, avoiding division by zero traps/panics

spytheman:
- **v: support -div-by-zero-is-zero, which allows for `x / 0 == 0`, avoiding division by zero traps/panics**
- **add a description for the new -div-by-zero-is-zero option to vlib/v/help/build/build-c.txt**


vlib/v/help/build/build-c.txt
```diff
@@ -361,3 +361,11 @@ see also `v help build`.
           https://learn.microsoft.com/en-us/cpp/build/reference/fp-specify-floating-point-behavior?view=msvc-170&redirectedfrom=MSDN
           https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffast-math
           https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-ffast-math
+
+   -div-by-zero-is-zero
+      Avoids division by zero errors, by treating `x / 0` as being equal to `0`.
+      It also treats `x % 0` as being equal to `x`, because `x%y`==`x-y*(x/y)`,
+      which in turn eliminates another source of division by zero errors.
+      It is mainly useful when prototyping games, and other experimental code with
+      lots of arithmetic expressions, where you do not want to check for the 0 divisor
```
Delta456: `zero` word sounds better than `0`

```suggestion
      lots of arithmetic expressions, where you do not want to check for the zero divisor
```

spytheman: I do not want to trigger all CI jobs just for it.

It can be changed later in another PR (with less checks, since it will affect a single .txt file, unlike here, where there are .v files in the compiler changed too).

vlib/v/gen/c/testdata/div_by_zero_is_zero.c.must_have
```diff
@@ -1 +1 @@
-int c = (int)(VSAFE_DIV_int(a,z));
+int c = (int)(VSAFE_DIV_int(a , z));
```
JalonSolov, line 1: Really a space both before & after?

spytheman: yes, it is needed for the other tests too

spytheman: (and for cases like this: `x--5`, which with the spaces is `x - -5`)

JalonSolov: Weird... but if that's what's needed, eh.

Issue #24977, opened 2025-07-25
time.custom_format() panics if time struct is uninitialized


mike-ward:
### Describe the bug

`time.custom_format()` panics if time struct is uninitialized

### Reproduction Steps

```v
import time

fn main() {
	bad_date := time.Time{}
	println(bad_date.custom_format('MMMM YYYY'))
}
```

### Expected Behavior

return an empty string?

### Current Behavior

```bash
v run .
V panic: array.get: index out of range (i,a.len):-1, 12
v hash: 4037765
0   bug                                 0x0000000104a30c24 array_get_with_check + 0
1   bug                                 0x0000000104a29604 array_get + 112
2   bug                                 0x0000000104a3c39c time__Time_custom_format + 1204
3   bug                                 0x0000000104a3ec74 main__main + 100
4   bug                                 0x0000000104a3f004 main + 84
5   dyld                                0x0000000183d3ab98 start + 6076
```

### Possible Solution

Suggest `time.custom_format` return `!string` instead of `string`.

### V version

V 0.4.11 4037765

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 da3112e.4037765
|:-------------------|:-------------------
|OS                  |macos, macOS, 15.5, 24F74
|Processor           |8 cpus, 64bit, little endian, Apple M2
|Memory              |0.14GB/8GB
|                    |
|V executable        |/Users/mike/Documents/github/v/v
|V last modified time|2025-07-23 13:51:06
|                    |
|V home dir          |OK, value: /Users/mike/Documents/github/v
|VMODULES            |OK, value: /Users/mike/.vmodules
|VTMP                |OK, value: /tmp/v_501
|Current working dir |OK, value: /Users/mike/Documents/github/bug
|                    |
|env VFLAGS          |"-path @vlib|@vmodules|/Users/mike/Documents/github -message-limit 5 -no-parallel"
|env LDFLAGS         |"-L/opt/homebrew/opt/ruby/lib"
|Git version         |git version 2.50.1
|V git status        |weekly.2025.25-171-g40377658
|.git/config present |true
|                    |
|cc version          |Apple clang version 17.0.0 (clang-1700.0.13.5)
|gcc version         |Apple clang version 17.0.0 (clang-1700.0.13.5)
|clang version       |Apple clang version 17.0.0 (clang-1700.0.13.5)
|tcc version         |tcc version 0.9.28rc 2024-02-05 HEAD@105d70f7 (AArch64 Darwin)
[...]

PR #24985, opened 2025-07-27
time: add default value for Time

kbkpbot:
Fix issue #24977


spytheman:
I think it would be better, if the panic-ing function is changed to not panic anymore for the 0 value, instead of changing the default (which then breaks all kinds of assumptions).

spytheman:
image
^ this is a much less disruptive change - it eliminates the panic, without affecting the rest of the tests.

PR #24990, opened 2025-07-27
log: level string shouldn't contain white space

ktunprasert:
Minor adjustment to the logging module, the fixed width log level string is removed let 4 characters be themselves without any empty space suffix


spytheman (commented):
No.
They are deliberately 5 character long, to match up with the DEBUG, ERROR and FATAL ones.

Delta456:
> No. They are deliberately 5 character long, to match up with the DEBUG, ERROR and FATAL ones.

We should add comments to describe that.

PR #24991, opened 2025-07-27
cgen: add endianness detection

kbkpbot:
For detect endianness at comptime.
First PR for bootstrap work.

### DIFF
--- a/vlib/v/gen/c/cheaders.v
+++ b/vlib/v/gen/c/cheaders.v
@@ -114,6 +114,30 @@ const c_common_macros = '
 	#define __V_architecture 9
 #endif
 
+#if defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
+    #define __V_little_endian 1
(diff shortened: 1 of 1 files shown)


vlib/v/gen/c/cheaders.v
```diff
@@ -114,6 +114,30 @@ const c_common_macros = '
...
+        #define __V_little_endian 1
+    #elif defined(__ARMEB__) || defined(__BIG_ENDIAN__)
+        #define __V_little_endian 0
+    #else
+        #error "Cannot determine endianness for ARM32. Please please raise an issue on GitHub: https://github.com/vlang/v/issues/new/choose."
+    #endif
+#else
+    #error "Cannot determine endianness. Please please raise an issue on GitHub: https://github.com/vlang/v/issues/new/choose."
+#endif
+
```
spytheman, line 140: We already have `TARGET_ORDER_IS_LITTLE` and `TARGET_ORDER_IS_BIG` macros
which are used with `$if little_endian {` and `$if big_endian {` .

Are not they enough?

kbkpbot: sorry, i search for the wrong word. 
:(

PR #24992, opened 2025-07-28
comptime: support string interpolation for `compile warn/error`

### LINKED ISSUE #24856: can't resolve T.name at compile time.
### Describe the bug

in $compile_error() or $compile_warn(), can't resolve ${T.name}

### Reproduction Steps

cc.v
```v
module main

fn t[T]() {
        $compile_warn('invalid type ${T.name}')
}

fn main() {
        t[f64]()
}
```

```
v cc.v
``

### Expected Behavior

```
invalid type f64
```

### Current Behavior

```
$ v cc.v
cc.v:4:30: error: unexpected token `$2`, expecting `)`
    2 |
    3 | fn t[T]() {
    4 |     $compile_warn('invalid type ${T.name}')
      |                                 ^
    5 | }
    6 |

```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.11 2c063ea19141163d7ea6a167abc355e7b4341ee4

|V full version      |V 0.4.11 2c063ea19141163d7ea6a167abc355e7b4341ee4
|OS                  |linux, Ubuntu 24.04.2 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |7.79GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-07-06 11:33:44
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|Current working dir |OK, value: /home/mars/v/bug/ptr
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-363-gc19e5ad1-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.4) 2.39

### PR DESCRIPTION
**Pull request by Krchi**
fix #24856, support `T.name`, `T.idx`, `T`
modify vfmt

### DIFF
--- a/vlib/v/checker/comptime.v
+++ b/vlib/v/checker/comptime.v
@@ -16,10 +16,10 @@ fn (mut c Checker) comptime_call(mut node ast.ComptimeCall) ast.Type {
 		node.left_type = c.expr(mut node.left)
 	}
 	if node.method_name == 'compile_error' {
-		c.error(c.comptime_call_msg(node) or { '' }, node.pos)
+		c.error(c.comptime_call_msg(node), node.pos)
 		return ast.void_type
 	} else if node.method_name == 'compile_warn' {
-		c.warn(c.comptime_call_msg(node) or { '' }, node.pos)
+		c.warn(c.comptime_call_msg(node), node.pos)
 		return ast.void_type
 	}
 	if node.is_env {
@@ -204,16 +204,13 @@ fn (mut c Checker) comptime_call(mut node ast.ComptimeCall) ast.Type {
 	return f.return_type
 }
 
-fn (mut c Checker) comptime_call_msg(node ast.ComptimeCall) ?string {
-	if node.method_name !in ['compile_error', 'compile_warn'] {
-		return none
-	}
+fn (mut c Checker) comptime_call_msg(node ast.ComptimeCall) string {
 	return if node.args_var.len > 0 {
 		node.args_var
 	} else if value := c.eval_comptime_const_expr(node.args[0].expr, -1) {
-		value.string()
+		value.string() or { '' }
 	} else {
-		none
+		''
 	}
 }
 

### CODE REVIEW

File: vlib/v/checker/comptime.v [hunk 1]
```diff
@@ -9,16 +9,33 @@ import v.token
...
 	if node.method_name == 'compile_error' {
-		c.error(node.args_var, node.pos)
+		msg := if node.args_var.len > 0 {
+			node.args_var
+		} else {
+			value := c.eval_comptime_const_expr(node.args[0].expr, -1) or {
+				ast.ComptTimeConstValue('')
+			}
+			value.string() or { '' }
+		}
```
spytheman: this can be extracted in its own method, and called for both the error and warning cases

vlib/v/tests/comptime/comptime_compile_warn_with_string_interp_test.v
```diff
@@ -0,0 +1,40 @@
+const msg = 'invalid type'
+
+fn t[T]() int {
+	$if T is i8 {
+		assert typeof[T]().name == typeof[i8]().name
+		$compile_warn('invalid type ${T.name} ${T.idx}')
```
spytheman, line 6: what happens if you make a syntax error here, like `${typo.name}` ?

vlib/v/checker/comptime.v
Krchi: > this can be extracted in its own method, and called for both the error and warning cases

how about with this
```v
if node.method_name in ['compile_error', 'compile_warn'] {
		func := if node.method_name == 'compile_error' { c.error } else { c.warn }
		if node.args_var.len > 0 {
			func(node.args_var, node.pos)
		} else {
			value := c.eval_comptime_const_expr(node.args[0].expr, -1) or {
				ast.ComptTimeConstValue('')
			}
			func(value.string() or { '' }, node.pos)
		}
		return ast.void_type
	}
```

vlib/v/tests/comptime/comptime_compile_warn_with_string_interp_test.v
Krchi: > what happens if you make a syntax error here, like `${typo.name}` ?

current is skip with error expr or unsupprt expr, i saw that not all conditions have been taken into account.
 i want use const in the case, but it seems more work to do
```v
ast.StringInterLiteral {
	if nlevel < 0 {
// other place use eval_comptime_const_expr for judge
		mut sb := strings.new_builder(20)
		for i, val in expr.vals {
			sb.write_string(val)
			if e := expr.exprs[i] {
				if value := c.eval_comptime_const_expr(e, nlevel + 1) {
					sb.write_string(value.string() or { '' })
				} else {
// error with error expr or unsupport expr
					c.error('unsupport expr `${e.str()}`', e.pos())
				}
			}
		}
		return sb.str()
	}
}

```

vlib/v/checker/comptime.v
spytheman: No, I do not want closures in the compiler code (they make code harder to port and bootstrap on new platforms).

Lines 19-26 and lines 30-37 are identical, and do not require much context beside passing the node.

They can be easily extracted into a method that just returns the `msg` string, and called in both places.

vlib/v/checker/comptime.v
```diff
@@ -220,6 +204,19 @@ fn (mut c Checker) comptime_call(mut node ast.ComptimeCall) ast.Type {
 	return f.return_type
 }
 
+fn (mut c Checker) comptime_call_msg(node ast.ComptimeCall) ?string {
+	if node.method_name !in ['compile_error', 'compile_warn'] {
+		return none
+	}
```
spytheman: that is already checked in the callsites, no need to check it here too

vlib/v/checker/comptime.v
```diff
@@ -220,6 +204,19 @@ fn (mut c Checker) comptime_call(mut node ast.ComptimeCall) ast.Type {
...
+	return if node.args_var.len > 0 {
+		node.args_var
+	} else if value := c.eval_comptime_const_expr(node.args[0].expr, -1) {
+		value.string()
+	} else {
+		none
```
spytheman: just `''`

spytheman: There is no need for all the callsites to do `or { '' }` . Just make this return `string` instead of `?string`.

vlib/v/checker/comptime.v
spytheman: it also makes the method less useful for similar situations in the future, for example if we decide to add `$compile_note(message)` construct, we would have to update this list too

PR #24999, opened 2025-07-29
markused: clean up features tracking from checker - remove 1 all_fns loop

felipensp:
clean up

### DIFF
--- a/vlib/v/markused/walker.v
+++ b/vlib/v/markused/walker.v
@@ -61,6 +61,7 @@ mut:
 	uses_memdup        bool // sumtype cast and &Struct{}
 	uses_arr_void      bool // auto arr methods
 	uses_index         bool
+	uses_append        bool // var << item
 }
 
 pub fn Walker.new(params Walker) &Walker {
@@ -585,6 +586,9 @@ fn (mut w Walker) expr(node_ ast.Expr) {
 			}
 			if !w.uses_eq && node.op in [.eq, .ne] {
 				w.uses_eq = true
+			}
+			if !w.uses_append && node.op == .left_shift {
+				w.uses_append = true
 			}
 		}
 		ast.IfGuardExpr {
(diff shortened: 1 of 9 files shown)


vlib/v/markused/walker.v
```diff
@@ -508,10 +512,11 @@ fn (mut w Walker) expr(node_ ast.Expr) {
 			w.expr(node.call_expr)
 		}
 		ast.IndexExpr {
+			w.uses_index = true
```
spytheman: You may need to consider functions tagged with `@[direct_array_access]` and `-force-bounds-checking` + `-no-bounds-checking` .

spytheman: cgen does it around here: 
`vlib/v/gen/c/fn.v:81:g.is_direct_array_access = node.is_direct_arr || g.pref.no_bounds_checking`

spytheman: i.e. `g.is_direct_array_access` is true inside such a function, and then a[idx] used inside it, should not generate array_get and array_set calls, same with s[idx], which should not generate string_at calls.

felipensp: Yeah. Good catch.

vlib/v/markused/walker.v
```diff
@@ -959,28 +949,37 @@ pub fn (mut w Walker) call_expr(mut node ast.CallExpr) {
 	}
 }
 
-pub fn (mut w Walker) fn_by_name(fn_name string) {
-	if w.used_fns[fn_name] {
-		return
-	}
-	stmt := w.all_fns[fn_name] or { return }
+pub fn (mut w Walker) fn_stmts(stmt ast.FnDecl) {
```
spytheman: fn_decl already exists 🤔

felipensp: hmm, good point.

felipensp: but it does not works calling w.fn_decl() where we do w.stmts(stmt.stmts) for some reason...  needs futher debugging.

vlib/v/markused/walker.v
```diff
@@ -814,18 +829,25 @@ pub fn (mut w Walker) fn_decl(mut node ast.FnDecl) {
...
+	w.is_direct_array_access = node.is_direct_arr || w.pref.no_bounds_checking
+	defer {
+		w.is_direct_array_access = last_is_direct_array_access
+	}
 	if w.trace_enabled {
 		w.level++
-		defer { w.level-- }
+		defer {
+			w.level--
+		}
```
spytheman, line 844: hm, interesting ... vfmt decided to expand the oneliner 🤔

spytheman:
@felipensp sorry for the hazzle, but I have reverted it, since it was worse for small programs.
I should have tested more thoroughly before merging in the first place 🤦🏻‍♂️.

PR #25007, opened 2025-07-31
markused: improve array resources tracking + remove all_fns loop for orm

felipensp:


spytheman:
Excellent work.
It is indeed much better now for `-cc gcc` and short programs.
It is also consistently faster than v on master.
The only remaining obstacle before merging is the vlang/adventofcode CI failure.

vlib/v/markused/walker.v
```diff
@@ -628,7 +628,8 @@ fn (mut w Walker) expr(node_ ast.Expr) {
 			w.expr(node.expr)
 			w.mark_by_type(node.expr_type)
 			w.uses_guard = true
-			if !w.uses_str_index_check && w.table.final_sym(node.expr_type).kind in [.u8, .string] {
+			if node.expr_type != 0 && !w.uses_str_index_check
+				&& w.table.final_sym(node.expr_type).kind in [.u8, .string] {
```
spytheman, line 632: Can it check using just `node.expr_type`, without having to call final_sym for each expression?

felipensp: but what about alias ?

felipensp: We can improve checking by `node.expr is ast.IndexExpr` though.

spytheman: > but what about alias ?

true ... I have not thought about that

PR #25020, opened 2025-08-01
math.bits: add asm implementations for some 64 bit ops

kbkpbot:
Add asm implementation for `mul_64`, `div_64`; add new function `mul_add_64`.


tankf33der:
1. Verified that the performance is the same, it didn't get worse. 
2. Added `mul_add_64()` to my test suite for testing the `math.bits` module. Passed.

spytheman:
The other PR is merged now.
@kbkpbot please rebase this one over current master, so that the CI can check it again.

spytheman (approved):
Excellent work.

PR #25021, opened 2025-08-01
decoder2: add custom decoders

Larsimusrex:
Adds custom decoders for strings/numbers/booleans/null. Also implements them for time.Time, math.big and json2's null.

example usage:

```v
import x.json2.decoder2
import x.json2
import math.big

struct MyString implements decoder2.StringDecoder, decoder2.NumberDecoder, decoder2.BooleanDecoder, decoder2.NullDecoder {
mut:
	data string
}

pub fn (mut ms MyString) from_json_string(raw_string string) ! {
	ms.data = raw_string
}

pub fn (mut ms MyString) from_json_number(raw_number string) ! {
	mut first := true

	for digit in raw_number {
		if first {
			first = false
		} else {
			ms.data += '-'
		}

		ms.data += match digit {
			`-` { 'minus' }
			`.` { 'dot' }
			`e`, `E` { 'e' }
			`0` { 'zero' }
			`1` { 'one' }
			`2` { 'two' }
			`3` { 'three' }
			`4` { 'four' }
			`5` { 'five' }
			`6` { 'six' }
			`7` { 'seven' }
			`8` { 'eight' }
			`9` { 'nine' }
			else { 'none' }
		}
	}
}

pub fn (mut ms MyString) from_json_boolean(boolean_value bool) {
	ms.data = if boolean_value {'yes'} else {'no'}
}

pub fn (mut ms MyString) from_json_null() {
	ms.data = 'default value'
}

fn main() {
	println(decoder2.decode[[]MyString]('["hi", -9.8e7, true, null]')!)
	
	println(decoder2.decode[json2.Any]('{"hi": 90, "bye": ["lol", -1, null]}')!)
	
	println(decoder2.decode[big.Integer]('-12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890')!)
}

// output

// [MyString{
//     data: 'hi'
// }, MyString{
//     data: 'minus-nine-dot-eight-e-seven'
// }, MyString{
//     data: 'yes'
// }, MyString{
//     data: 'default value'
// }]
// {"hi":90,"bye":["lol",-1,null]}
// -12345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890

```


enghitalo:
I need to think about the concept of this new feature a bit more before revising it. It bothers me a bit that other libraries have to care about what JSON does, plus a few other things. I think it's worth checking if other languages like go have something similar. What do you think @JalonSolov @medvednikov @spytheman ?

Larsimusrex:
I don't really like the way how go does custom decoders. You often have to call json.Unmarshal again, like they do in their docs https://pkg.go.dev/encoding/json#example-package-CustomMarshalJSON. We already know the json type and we know it's correct, because we checked it. So there is no reason to call e.g. `decode[string](data)` again.
As for the other libraries knowing about json2, the only alternative I can think of is making type aliases for these Types in json2 so that the from_json_* methods can be written there. But I don't like that either.
Always open for suggestions though.

JalonSolov:
Perhaps table the idea of custom decoders, for now.

Emphasize json2 as a complete replacement for json, so we can finally get rid of cJSON, etc.

Then once that is working, revisit custom decoders.

I do agree that whatever is done should handle things in a single pass... custom decoders should be used in place of the regular decoders, rather than making extra calls.

JalonSolov:
To put it another way... I'd rather have a plain json decoder that handled the json spec as-is, preferably reasonably fast.  I'd want that before I would worry about adding extra features.

Otherwise, you spend so much time on the features that the basic functionality is never finished.

Larsimusrex:
Well this was mostly so that things like json2.Null and time.Time can be decoded without hardcoding them into the decoder, like they are right now. see https://github.com/vlang/v/pull/24949#issuecomment-3104636935. With this PR json2.Any would be fully supported so json2's decode could be replaced solving a major headache.

vlib/x/json2/decoder2/custom.v
```diff
@@ -0,0 +1,50 @@
...
+	// called with raw string of number e.g. '-1.234e23'
+	from_json_number(raw_number string) !
+}
+
+// implements decoding json true/false
+pub interface BooleanDecoder {
+mut:
+	// called with converted bool
+	// already checked so no error needed
+	from_json_boolean(boolean_value bool)
```
spytheman, line 22: why this gets a `bool`, but the other methods expect a string parameter?

edit: I saw the call site, it gets a bool, because the comparison is done there, instead of implementers having to do `if value == 'true' {` themselves...
I still wonder if the potential optimization of having it being done once, instead of twice is worth the consistency break with the other interfaces.

spytheman: is there an useful case, that this interface covers?

spytheman:
> It bothers me a bit that other libraries have to care about what JSON does

The nice thing is that they *do not have to*, if they do not want their types to have custom decodes, but if they do, they can just implement a few methods with a specific signature (note that in V, as in Go, interfaces can be implemented implicitly, without having to import anything).

The nicer thing is that the json module *also* does not have to care about custom types too, but only that they implemented a specific set of methods, to be able to support them.

vlib/x/json2/decoder2/decode.v
spytheman, line 1: The `import time` on top of `vlib/x/json2/decoder2/decode.v`, is unused due to the changes here.
Please remove the unused import, because otherwise programs that `import x.json2.decoder` will not be able to be compiled with -prod (it turns out warnings into errors).

spytheman:
> I'd rather have a plain json decoder that handled the json spec as-is, preferably reasonably fast. I'd want that before I would worry about adding extra features.

This PR helps to achieve it.

Consider this small example, that passes here, but at runtime breaks on master:
```v
import x.json2
import x.json2.decoder2
  
x := decoder2.decode[json2.Any]('["hi", -9.8e7, true, null]')!
dump(x)
```
producing:
```
[a.v:5] x: ["hi",-98000000,true,null]
```


On master:
```
V panic: result not set (
1:8: Invalid json: Syntax: expected digit got -
["hi", -9.8e
       ^)
```


The alternative (which works on master), using the current decoder in json2 is:
```v
import x.json2
x := json2.decode[json2.Any]('["hi", -9.8e7, true, null]')!
dump(x)
```

-> the decoder2 here is improved, and closer to replacing the one in x.json2, while also being faster, and easier to customize for types like time.Time etc.

spytheman:
@Larsimusrex to pass the full CI, you need to also document the new `pub` symbols:
```
vlib/math/big/json_decode.v:3:0:pub fn (mut result Integer) from_json_number(raw_number string) ! 
vlib/time/json_decode.c.v:3:0:pub fn (mut t Time) from_json_string(raw_string string) ! 
vlib/x/json2/types.v:44:0:pub fn (mut n Null) from_json_null() 
```
(this is from https://github.com/vlang/v/actions/runs/16682503917/job/47224275305?pr=25021)

vlib/x/json2/decoder2/custom.v
Larsimusrex: I would think when you have json data that for example also contains 0 and 1 for booleans you could create a custom decoder for that, but still support normal true/false by implementing this. 
Mostly for completeness though, would feel wrong to have decoders for numbers/strings/null then leave out booleans. It's a compile time check so it wont make a performance difference anyways.

spytheman:
> vlib modules CI / build-module-docs (push)  failing

This will need a rebase/merge with master too.
(it is unrelated to this PR, so you can also just ignore it)

spytheman (approved):
Excellent work.
Thank you @Larsimusrex 🙇🏻‍♂️ .

PR #25030, opened 2025-08-02
net.http: remove debug print statement from post_multipart_form

D4MI4NX:
The post_multipart_form function prints the configured headers when called. I suppose this is not intended. This PR just removes this debug print statement.


Delta456 (approved):
Good catch!

spytheman:
Thank you @D4MI4NX .

PR #25034, opened 2025-08-03
parser,checker,cgen: remove unused `err` declaration or `or { }` blocks

felipensp:


spytheman (approved):
Excellent work.

vlib/v/checker/fn.v
```diff
@@ -3819,10 +3822,11 @@ fn scope_register_a_b(mut s ast.Scope, pos token.Pos, typ ast.Type) {
 
 fn scope_register_var_name(mut s ast.Scope, pos token.Pos, typ ast.Type, name string) {
```
spytheman, line 3823: What do you think of renaming that to `scope_register_special_var_name` ?

felipensp: Great.

Issue #25032, opened 2025-08-03
Please add `like` to Function-Call-API


keesse:
### Describe the feature

When trying to do a `qb.where('$name_type like ?', '$name%')!`
I got the following error:

```
V panic: result not set (parse_conditions(): unsupported operator: `like`
        lastname like ?
```

Please add `like` to Function-Call-API

### Use Case

```v
module main

import db.sqlite
import orm

@[table: 'names']
struct Name {
    id int @[primary; serial]
    firstname string
    lastname string
}

fn main () {
    db := sqlite.connect('disney.db')!
    mut qb := orm.new_query[Name](db)

    name_type := 'lastname'

    //qb.where('lastname = ?', 'Mouse')!   // works
    qb.where('$name_type like ?', 'Mou%')! // unsupported operator: `like`
    ...
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

Current V version: V 0.4.11 d249bfb, timestamp: 2025-08-03 10:27:43 +0300

### Environment details (OS name and version, etc.)

Manjaro Linux
Linux lati3 6.15.7-1-MANJARO #1 SMP PREEMPT_DYNAMIC Thu, 17 Jul 2025 20:47:04 +0000 x86_64 GNU/Linux

PR #25035, opened 2025-08-03
orm: support lowercase operators `like`, `in`, `not in` etc 

Delta456:
solves https://github.com/vlang/v/issues/25032


JalonSolov:
I would be somewhat simpler to create an UPPERCASE copy of the input, and stepping though that, instead of adding all the lowercase versions of things.

Delta456:
> I would be somewhat simpler to create an UPPERCASE copy of the input, and stepping though that, instead of adding all the lowercase versions of things.

I did think about `ss.input[ss.pos..].to_upper()` but that might be overkill for small comparison.

JalonSolov:
No, I meant
```v
tmp_ss := ss.input.to_upper()
```
then
```v
if tmp_ss[ss.pos..].starts_with('IS NOT NULL')
```
later in the code.

In other words, keep `ss.input` as-is, do the `.to_upper()` call once, and only do comparisons against `tmp_ss`.

JalonSolov:
Of course, if you copy anything out of `ss.input`, always copy out of the original, not `tmp_ss`.

Delta456:
> Of course, if you copy anything out of `ss.input`, always copy out of the original, not `tmp_ss`.

Good idea. I just pushed the changes. Do review

JalonSolov (approved):

spytheman (approved):
Excellent work.

PR #25062, opened 2025-08-07
gg:improve drawing effect and logic of the `draw_rounded_rect_empty`

tzSharing:
fix local pixel misalignment and simplify the drawing logic of the `gg`/`draw_rounded_rect_empty()`

BaiduShurufa_2025-8-7_15-15-3

BaiduShurufa_2025-8-7_16-4-37

### DIFF
--- a/vlib/gg/draw.c.v
+++ b/vlib/gg/draw.c.v
@@ -290,89 +290,84 @@ pub fn (ctx &Context) draw_rounded_rect_empty(x f32, y f32, w f32, h f32, radius
 	sgl.c4b(c.r, c.g, c.b, c.a)
 
 	mut new_radius := radius
+	if radius < 1 {
+		new_radius = 0
+	}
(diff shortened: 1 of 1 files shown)


vlib/gg/draw.c.v
```diff
@@ -290,89 +290,84 @@ pub fn (ctx &Context) draw_rounded_rect_empty(x f32, y f32, w f32, h f32, radius
...
+
 	r := new_radius * ctx.scale
 	sx := x * ctx.scale // start point x
 	sy := y * ctx.scale
 	width := w * ctx.scale
 	height := h * ctx.scale
+
+	align_pixel := fn (v f32) f32 {
+		return math.floorf(v) + 0.5
+	}
```
spytheman, line 310: just add a normal named function, outside of draw_rounded_rect_empty

PR #25067, opened 2025-08-08
cgen: fix asm stmt separators

kbkpbot:
```
"mul %[lo], %[x], %[y];"
"umulh %[hi], %[x], %[y];"
```

This violates GCC/Clang's inline assembly syntax requirements. And this PR replaced semicolons with proper `\n\t` separators.

```
"mul %[lo], %[x], %[y]\n\t"
"umulh %[hi], %[x], %[y]\n\t"
```

### DIFF
--- a/vlib/v/slow_tests/assembly/stmt_separator_test.amd64.v
+++ b/vlib/v/slow_tests/assembly/stmt_separator_test.amd64.v
@@ -1,0 +1,48 @@
+import os
+import time
+
+const asm_file_content = '
+fn asm_fn() {
+	mut x := u64(0)
+	mut y := u64(0)
+	mut z := u64(0)
+	mut hi := u64(0)
+	mut lo := u64(0)
+	asm amd64 {
+		mulq rdx
+		addq rax, z
+		adcq rdx, 0
+		; =a (lo)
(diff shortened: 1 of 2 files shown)


spytheman:
please add a test too

vlib/v/slow_tests/assembly/stmt_separator_test.amd64.v
```diff
@@ -0,0 +1,48 @@
...
+	eprintln('> compile_cmd: ${compile_cmd}')
+	time.sleep(1000 * time.millisecond) // improve chances of working on windows
+	compile_res := os.system(compile_cmd)
+	assert compile_res == 0
+
+	content := os.read_file(c_file)!
+	os.rm(v_file)!
+	os.rm(genexe_file)!
+	os.rm(c_file)!
+	assert content.contains('addq %[z], %%rax\\n\\t')
```
JalonSolov: Perhaps a raw string would be better?  Wouldn't look as messy, anyway...
```suggestion
	assert content.contains(r'addq %[z], %%rax\n\t')
```

spytheman (approved):
Excellent work.

Issue #25070, opened 2025-08-08
can't parse some asm constraint in inline assembly


kbkpbot:
### Describe the bug

The V parser does not support some asm constraint in inline assembly.
for example :

```
                ; =&r (lo)
                  =&r (hi)

```

### Reproduction Steps

t.v
```v
module main

fn mul_64(x u64, y u64) (u64, u64) {
        mut hi := u64(0)
        mut lo := u64(0)
        asm amd64 {
                mov rax, x
                mulq y
                mov lo, rax
                mov hi, rdx
                ; =&r (lo)
                  =&r (hi)
                ; r (x)
                  r (y)
                ; rax
                  rdx
                  cc
        }
        return hi, lo
}

fn main() {
        i, j := mul_64(100, 200)

        println(i)
        println(j)
}
```

### Expected Behavior

compile OK

### Current Behavior

```
$ $ v t.v
t.v:11:6: error: unexpected token `&`, expecting name
    9 |         mov lo, rax
   10 |         mov hi, rdx
   11 |         ; =&r (lo)
      |            ^
   12 |           =&r (hi)
   13 |         ; r (x)
```

### Possible Solution

Fix parser

### V version

V 0.4.11 3b561d590d5281745018b084ba62d312741aa409.0dcd3b1

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 3b561d590d5281745018b084ba62d312741aa409.0dcd3b1
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |8.2GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-08-08 14:10:09
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/asm
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-534-g0dcd3b11-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
[...]

PR #25072, opened 2025-08-09
parser: fix asm modifier parse

kbkpbot:
Fix issue #25070

Output constrains: `=r`, `=&r`, `+r`, `+&r`,`=m`,`=rm`,`=@ccl`, `=*r`
Input constrains: `r`, `0`, `rm`, `i`, `%r`, `*m`

### DIFF
--- a/vlib/v/slow_tests/assembly/asm_test.amd64.v
+++ b/vlib/v/slow_tests/assembly/asm_test.amd64.v
@@ -110,6 +110,26 @@ fn test_inline_asm() {
 		; ; eax
 	}
 	assert o.str()[0].is_capital()
+
+	x := u64(100)
+	y := u64(200)
+	mut hi := u64(0)
+	mut lo := u64(0)
+	asm amd64 {
+		mov rax, x
+		mulq y
+		mov lo, rax
+		mov hi, rdx
+		; =&r (lo)
+		  =&r (hi)
+		; r (x)
+		  r (y)
+		; rax
+		  rdx
+		  cc
+	}
+	assert hi == 0
+	assert lo == 20000
 }
 
 @[packed]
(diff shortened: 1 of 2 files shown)


vlib/v/slow_tests/assembly/asm_test.amd64.v
```diff
@@ -110,6 +110,26 @@ fn test_inline_asm() {
...
+	x := u64(100)
+	y := u64(200)
+	mut hi := u64(0)
+	mut lo := u64(0)
+	asm amd64 {
+		mov rax, x
+		mulq y
+		mov lo, rax
+		mov hi, rdx
+		; =*m (lo)
```
spytheman, line 123: hm, that does not work with my local tcc 🤔:
`./v -cc tcc -no-retry-compilation vlib/v/slow_tests/assembly/asm_test.amd64.v`

spytheman:
```
=== C compilation error (from /home/delian/v/thirdparty/tcc/tcc.exe): 
cc: /tmp/v_1000/asm_test.amd64.01K26VFD0GKTC5S097ADB5HCSA.tmp.c:8644: 
  error: unknown constraint '*'
=====================================================================

spytheman (approved):
Excellent work.

vlib/v/slow_tests/assembly/asm_test.amd64.v
spytheman: If you intend to add asm blocks for commonly used modules like `math.bits` or `strconv`, please use syntax that tcc can also process, otherwise it will make bootstrapping more difficult, and compilation for many programs a lot slower, due to the retrying and falling back to slower compilers like clang/gcc.

PR #25073, opened 2025-08-09
checker: fix array init with interface

### LINKED ISSUE #24255: C error when testing an array of empty interfaces
The error was encountered when running [v run src/statement_test.v](https://github.com/einar-hjortdal/firebird/blob/541f5c7b745532030db44699f349fda217ded69d/src/statement_test.v#L382)

It can be reproduced by cloning the repo and checking out at the hash of the link.

<details>
<summary>V version: V 0.4.10 ccb3d5c, press to see full `v doctor` output</summary>

|V full version      |V 0.4.10 537605a.ccb3d5c
|OS                  |linux, "openmamba release 2024.6 for x86_64 (rolling)"
|Memory              |13.52GB/31.16GB
|                    |
|V last modified time|2025-04-17 08:10:26
|                    |
|Current working dir |OK, value: /home/einar/Documents/projects/vlang/active/firebird
|                    |
|V git status        |weekly.2025.16-11-gccb3d5cf
|.git/config present |true
|                    |
|cc version          |cc (GCC) 14.2.1 20240909
|gcc version         |gcc (GCC) 14.2.1 20240909
|clang version       |N/A
|emcc version        |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  src/statement_test.v && src/statement_test`
See attached file `src/statement_test.v`

**What did you see?**
```
cc: /tmp/v_1000/statement_test.01JS1GYQSY01NY8645NM7XCTC9.tmp.c:24658: warning: cast between pointer and integer of different size
cc: /home/einar/.local/lib64/v/thirdparty/tcc/lib/libgc.a: error: 'GC_noop1_ptr' defined twice
cc: tcc: error: undefined symbol 'I_int_to_Interface_firebird__Value'
...
cc: tcc: error: undefined symbol 'I_int_to_Interface_firebird__Value'
(note: the original output was 5 lines long; it was truncated to its first 3 lines + the last line)
builder error: 
==================
```

No error

> [!NOTE]
>

Krchi:
fix #24255


vlib/v/tests/builtin_arrays/array_init_with_interface_test.v
spytheman, line 1: this test passes on master too

spytheman: image

spytheman: The example that @felipensp gave however, fails both here and on master:
```v
interface Value {}

fn main() {
        mut a := []Value{}
        a << [Value(i32(1)), 2]
        println('ok')
}
```

Imho it should be tested with a `.vv`/`.run.out` pair in 
`vlib/v/tests/skip_unused/`, since otherwise the test infrastructure (which also uses interfaces and printing) interferes with what needs to be checked.

Something like this in the `check_interface_push.vv` file:
```v
interface Value {}

fn main() {
	mut a := []Value{}
	a << [Value(i32(1)), 2]
	println(a.len)
}
```
and in the corresponding `check_interface_push.run.out` file next to it:
```
2
```

can be then checked with: `VTEST_ONLY=check_interface_push ./v vlib/v/compiler_errors_test.v`

(to decrease the iteration time locally; the CI runs the full suite)

spytheman: note: it passes with `-no-skip-unused`, so it will probably also need modifications in `markused` as well.

Krchi: origin test is `assert a == [Value(i32(1)), 2.0, 3]`, test is success in my pc, I didn't realize `Value(2.0)` and `Value(3)` will complete the type for interface
I'm testing it in another branch

Krchi: > note: it passes with `-no-skip-unused`, so it will probably also need modifications in `markused` as well.

`-no-skip-unused` is ok with tcc(cgen is ok, there is a builder error with libgc in mac), but not work in gcc and clang

spytheman: My point is that it should not be tested with this _test.v file at all.
This specific _test.v file passes on master too - it is not effective in checking that the changes here work and fix the original problem.
In fact, they do not, as evidenced by running the minimal example that @felipensp gave, which does fail both here and on master, with a cgen error.

This can be tested with a .vv/.out pair, to avoid the interference of the testing framework.

spytheman: i.e. if I merge this PR as it is, the issue will be closed, but the problem will remain.

Krchi: > i.e. if I merge this PR as it is, the issue will be closed, but the problem will remain.

you are right, i get what error now

spytheman (approved):
Excellent work.

spytheman:
Thank you for persevering, and fixing it completely @Krchi 🙇🏻‍♂️ .

Issue #25071, opened 2025-08-08
use zstd example outputs nothing


l1t1:
I add a main() function in the file vlib/compress/zstd/read_zstd_files_test.v
```v
import os
import compress.zstd

const samples_folder = os.join_path(os.dir(@FILE), 'samples')

fn main() {test_reading_and_decoding_a_known_zstded_file()
	}

fn s(fname string) string {
	return os.join_path(samples_folder, fname)
}

fn read_and_decode_file(fpath string) !([]u8, string) {
	compressed := os.read_bytes(fpath)!
	decoded := zstd.decompress(compressed)!
	content := decoded.bytestr()
	return compressed, content
}

fn test_reading_and_decoding_a_known_zstded_file() {
	compressed, content := read_and_decode_file(s('known.zst'))!
	assert compressed#[0..3] == [u8(40), 181, 47]
	assert compressed#[-5..] == [u8(10), 78, 32, 170, 44]
	assert content.contains('## Description')
	assert content.contains('## Examples:')
	assert content.ends_with('```')
}

fn test_decoding_all_samples_files() {
	for zstd_file in os.walk_ext(samples_folder, '.zst') {
		_, content := read_and_decode_file(zstd_file)!
		assert content.len > 0, 'decoded content should not be empty: `${content}`'
	}
}

fn test_reading_zstd_files_compressed_with_different_compress_level() {
	_, content1 := read_and_decode_file(s('readme_level_1.zst'))!
	_, content5 := read_and_decode_file(s('readme_level_5.zst'))!
	_, content9 := read_and_decode_file(s('readme_level_9.zst'))!
	_, content19 := read_and_decode_file(s('readme_level_19.zst'))!
	assert content19 == content9
	assert content9 == content5
	assert content5 == content1
}
```
I tried following command, they all output nothing
```bash
root@6ae32a5ffcde:/par/v/vlib/compress/zstd# ls
README.md  read_zstd_files_test.v  samples  zstd.c.v  zstd_d_musl.c.v  zstd_test.v
root@6ae32a5ffcde:/par/v/vlib/compress/zstd# v read_zstd_files_test.v
root@6ae32a5ffcde:/par/v/vlib/compress/zstd# ./read_zstd_files_test
bash: ./read_zstd_files_test: No such file or directory
root@6ae32a5ffcde:/par/v/vlib/compress/zstd# ls
README.md  read_zstd_files_test.v  samples  zstd.c.v  zstd_d_musl.c.v  zstd_test.v
[...]

PR #25078, opened 2025-08-09
zstd: add buffer length overflow detect

kbkpbot:
May fix issue #25071

Add a error message when buffer length is bigger than `int_max`.


spytheman:
no, I do not want to merge large compressed binaries for a specific test

Issue #25079, opened 2025-08-09
v gen wrong code for translated v code, ptr operation is wrong


kbkpbot:
### Describe the bug

When use V compile a translated v code `vtcc`, it generated wrong c code, which will cause `SIGSEGV`

### Reproduction Steps

a test file, p.v, this case is extracted from `vtcc/src/i386-asm.v`, function `asm_compute_constraints`

```v
@[translated]
module main

struct ASMOperand {
        constraint [16]i8
}

fn main() {
        op := &ASMOperand{}
        str := &char(0)
        str = op.constraint

        dump(str)
}
```

``` 
v p.v
```

### Expected Behavior

compile & run OK.

### Current Behavior

```
$ ./p
0x7a8202d88a6c: at ???: RUNTIME ERROR: invalid memory access
/tmp/v_1000/p.01K27AADJKBY9B4WR2J8XB4DP4.tmp.c:2187: by main
```

### Additional Information/Context

in generated C file:
```c
VV_LOC void main__main(void) {
        main__ASMOperand* op = ((main__ASMOperand*)memdup(&(main__ASMOperand){.constraint = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0},}, sizeof(main__ASMOperand)));
        char* str = ((char*)(0));
        memcpy(str, op->constraint, sizeof(Array_fixed_i8_16));
         _v_dump_expr_char__ptr(_S("p.v"), 13, _S("str"), str);
}
```

It use `memcpy` instead of a pointer assignment here.

### V version

V 0.4.11 3b561d590d5281745018b084ba62d312741aa409.acbb12f

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 3b561d590d5281745018b084ba62d312741aa409.acbb12f
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |1.95GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-08-09 08:56:05
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/ptr
[...]

PR #25080, opened 2025-08-09
cgen: fix translated file fixed-array assignment

kbkpbot:
Fix issue #25079 

generate a pointer assignment instead of `memcpy`.


spytheman:
imho only `g.file.is_translated` should remain ; `c2v` afaik already does use `@[translated]` tags for the generated .v files, and it provides finer grained control, compared to a global CLI option

spytheman (approved):

spytheman:
Good work on fixing cgen.


I am not fully convinced, that `p = i.fixed` should implicitly take the address of `i.fixed` 🤔 .

From one viewpoint, the previous behavior, at least could be explained as a logical consequence of the automatic dereferencing.

From another, in C, a fixed array can be treated as the adress of its first element, and I guess, that is fine in `@[translated]` code, since it will minimize the logic in c2v 🤔 .

PR #25083, opened 2025-08-09
cgen: fix codegen for passing int from selector to voidptr expect arg

### LINKED ISSUE #25081: incompatible integer to pointer conversion
```
struct Mike {
mut:
        value int
}

fn main() {
        shared m := Mike{}
        ch := chan int{}
        ch.try_push(rlock m { m.value })
}
```

```
================== C compilation error (from clang-22): ==============
cc: /tmp/v_0/a.01K27J0CD1QTFXEZYS9AHWSAXA.tmp.c:9260:44: error: incompatible integer to pointer conversion initializing 'voidptr' (aka 'void *') with an expression of type 'int' [-Wint-conversion]
cc:  9260 |                 sync__Channel_try_push(ch, ADDR(voidptr, _t1));
cc:       |                                                          ^~~
cc: 1 error generated.
builder error: 
==================
```

```
gcc 14.2.0 - ok

clang 18,20,22 - error
```

> [!NOTE]
>

felipensp:
Fix #25081

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -2707,6 +2707,8 @@ fn (mut g Gen) ref_or_deref_arg(arg ast.CallArg, expected_type ast.Type, lang as
 						}
 						if arg.expr.is_as_cast() {
 							g.inside_smartcast = true
+						} else if arg_typ_sym.is_int() {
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.

PR #25094, opened 2025-08-12
CI: migrate workflows to checkout v5

rejected-l:
Node 24 compatibility: switch all jobs to actions/checkout@v5. Requires runner v2.327.1+. Pure maintenance, behavior unchanged.

Ref: https://github.com/actions/checkout/releases/tag/v5.0.0


spytheman:
There is already a dependabot PR for this exact change.

PR #25096, opened 2025-08-12
cgen: allow alias sumtype smartcasting

Delta456:
Fixes https://github.com/vlang/v/issues/25085


vlib/v/checker/if.v
```diff
@@ -707,14 +707,15 @@ fn (mut c Checker) smartcast_if_conds(mut node ast.Expr, mut scope ast.Scope, co
 				right_sym := c.table.sym(right_type)
 				mut expr_type := c.unwrap_generic(node.left_type)
 				left_sym := c.table.sym(expr_type)
+				left_final_sym := c.table.final_sym(expr_type)
```
spytheman, line 710: If we need to get `left_final_sym`, do we still need the `left_sym` too?
Can we use `left_final_sym` in all places in this branch instead?

Issue #25108, opened 2025-08-14
mut sumtype value can't be changed


kbkpbot:
### Describe the bug

When a function try to change a `mut` sumtype, it will fail.

### Reproduction Steps

mut.v
```v
module main

struct CondA {
pub mut:
        val int
}

struct CondB {
pub mut:
        val int
}

type Cond = CondA | CondB

fn func(mut cond Cond) {
        if mut cond is CondA {
                cond.val = 1234
        }
}

fn main() {
        mut x := CondA{}
        x.val = 1

        func(mut x)
        assert x.val == 1234
        dump(x)
}
```

```
v mut.v
```

run it

### Expected Behavior

run without  assert fail

### Current Behavior

```
$ ./mut
mut.v:26: FAIL: fn main.main: assert x.val == 1234
   left value: x.val = 1
  right value: 1234
V panic: Assertion failed...
v hash: 7d67e22
/tmp/v_1000/mut.01K2MD1A9PM22ZM47G85ADYHQB.tmp.c:3691: at _v_panic: Backtrace
/tmp/v_1000/mut.01K2MD1A9PM22ZM47G85ADYHQB.tmp.c:5352: by main__main
/tmp/v_1000/mut.01K2MD1A9PM22ZM47G85ADYHQB.tmp.c:5381: by main

```

### V version

V 0.4.11 d5458b58ebd87877de2477b7cdcbfcfabe2747de.7d67e22

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 d5458b58ebd87877de2477b7cdcbfcfabe2747de.7d67e22
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |1.24GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-08-14 13:31:04
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/mut
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-547-g7d67e22b-dirty
|.git/config present |true
|                    |
[...]

PR #25111, opened 2025-08-15
cgen: fix mutable sumtype

kbkpbot:
Fix issue #25108
remove `$d('mutable_sumtype', false)`, and this will make `sumtype` is mutable.


spytheman (approved):
Excellent work!

Thank you @kbkpbot 🙇🏻‍♂️.

PR #25117, opened 2025-08-16
checker: fix missing map float key duplicated checking

### LINKED ISSUE #25098: Map can't detect repeated `float` keys
### Describe the bug

Map initialized with repeated float keys is accepted though finally replaces repeated pairs with the last one.

### Reproduction Steps

```v
fn main() {
	mi := {
		1: 'one'
		// 1: 'one' duplicated key "1" in map literal
	}
	dump(mi)
	m64 := {
		0.000001: 'small-1'
		0.000001: 'small-2'
		1.e-06:   'small-3'
	}
	dump(m64)
}
```

### Expected Behavior

Error like:

```
0.000001: 'small-2' duplicated key "0.000001" in map literal
```

### Current Behavior

[code.v:6] mi: {1: 'one'}
[code.v:12] m64: {1.e-06: 'small-3'}

### Possible Solution

_No response_

_No response_

### V version

V 0.4.11 e5100a8

https://play.vlang.io/p/90c282e33e

> [!NOTE]
>

felipensp:
Fix #25098

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -5293,6 +5293,15 @@ fn (mut c Checker) check_dup_keys(node &ast.MapInit, i int) {
 				}
 			}
 		}
+	} else if key_i is ast.FloatLiteral {
+		for j in 0 .. i {
+			key_j := node.keys[j]
+			if key_j is ast.FloatLiteral {
+				if key_i.val.f64() == key_j.val.f64() {
+					c.error('duplicate key "${key_i.val}" in map literal', key_i.pos)
+				}
+			}
+		}
 	}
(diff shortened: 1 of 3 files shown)


Delta456:
Is there a better way to check for duplicate keys? Right now we are iterating for every node and comparing.

Can we use a sum type here and compare implicitly? It should help to reduce number of loops.

spytheman:
> Can we use a sum type here and compare implicitly? It should help to reduce number of loops.

I am not sure I understand the idea.
Can you please make another PR with it?

spytheman (approved):
Excellent work.

PR #25118, opened 2025-08-16
checker: fix missing check for empty array to generic param

### LINKED ISSUE #25056: use of undeclared identifier 'T'
```
import datatypes

fn main() {
        mut a := datatypes.Set[int]{}
        a.add_all([])
}
```

```
================== C compilation error (from clang-22): ==============
cc: /tmp/v_1000/intersection.01K202DAX1XFPP40A1EQGGP4HR.tmp.c:3558:79: error: use of undeclared identifier 'T'
cc:  3558 |         datatypes__Set_T_int_add_all_T_int(&a, __new_array_with_default(0, 0, sizeof(T), 0));
cc:       |                                                                                      ^
cc: 1 error generated.
builder error: 
==================
```

> [!NOTE]
>

felipensp:
Fix #25056

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -2576,6 +2576,9 @@ fn (mut c Checker) method_call(mut node ast.CallExpr, mut continue_check &bool) 
 			}
 		}
 		if exp_arg_typ.has_flag(.generic) {
+			if arg.expr is ast.ArrayInit && arg.expr.exprs.len == 0 {
+				c.error('cannot use empty array as generic argument', arg.pos)
+			}
 			method_concrete_types := if method_generic_names_len == rec_concrete_types.len {
 				rec_concrete_types
 			} else {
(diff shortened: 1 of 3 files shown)


vlib/v/checker/fn.v
```diff
@@ -1545,6 +1545,9 @@ fn (mut c Checker) fn_call(mut node ast.CallExpr, mut continue_check &bool) ast.
 		arg_typ_sym := c.table.sym(arg_typ)
 		if arg_typ_sym.kind == .none && param.typ.has_flag(.generic) && !param.typ.has_flag(.option) {
 			c.error('cannot use `none` as generic argument', call_arg.pos)
+		} else if node.raw_concrete_types.len == 0 && param.typ.has_flag(.generic)
+			&& call_arg.expr is ast.ArrayInit && call_arg.expr.typ == ast.void_type {
+			c.error('cannot use empty array as generic argument', call_arg.pos)
```
spytheman: imho extract the check and error, in its own method and call it in both places

felipensp: Ok, done.

spytheman (approved):
Excellent work.

Issue #25099, opened 2025-08-13
$if in $for attribute, does not issue error when attr field not exisit


kbkpbot:
### Describe the bug

When traversing properties with `$for`, if an `$if` appears, it does not check whether the field exists.

### Reproduction Steps

t.v
```v
@[foo: 123]
@[bar]
fn abc() {
}

fn main() {
        mut c := 0
        $for f in abc.attributes {
                $if f.has_xdargfdsafsdfx {
                        c++
                        println('has_arg')
                }
        }
        assert c == 0
}
```

```
v t.v
```

### Expected Behavior

Should not compile, and issue a error msg, about field `has_xdargfdsafsdfx` does not exit.

### Current Behavior

compile, but  run fail

```
$ v aa.v
mars@mars-QiTianM428-A688:~/v/bug/comptime$ ./aa
has_arg
has_arg
aa.v:14: FAIL: fn main.main: assert c == 0
   left value: c = 2
  right value: 0
V panic: Assertion failed...
v hash: 7d67e22
/tmp/v_1000/aa.01K2GDNH040VVHMSJHQ8N50W0V.tmp.c:3178: at _v_panic: Backtrace
/tmp/v_1000/aa.01K2GDNH040VVHMSJHQ8N50W0V.tmp.c:4821: by main__main
/tmp/v_1000/aa.01K2GDNH040VVHMSJHQ8N50W0V.tmp.c:4848: by main
```

### Possible Solution

fix checker

### V version

V 0.4.11 d5458b58ebd87877de2477b7cdcbfcfabe2747de.7d67e22

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 d5458b58ebd87877de2477b7cdcbfcfabe2747de.7d67e22
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |1.35GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-08-13 00:31:01
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/comptime
|                    |
|Git version         |git version 2.43.0
[...]

Issue #24938, opened 2025-07-20
`$if` with (same) test ||'ed seems to be silently ignored


mid1221213:
### Describe the bug

Inside a `$for` loop on a struct fields, a `$if` test with two same conditions OR'ed together seems to make the compilation behave like there was no `$if` at all.

Actually anything ||'ed in the condition, even if different tests, seems to produce the bug.

### Reproduction Steps

Create the file `tst.v` with this content:

```v
struct Struct {
    s string
    i int
}

fn main() {
    tst := Struct{
        s: 'tst-s'
        i: 42
    }

    $for field in Struct.fields {
        $if field.typ is string || field.typ is string {
            if tst.$(field.name) == '' {
                panic('fail')
            }
        }
    }
}
```

then, `v run tst.v`

### Expected Behavior

```
$ v run tst.v
$
```

(no error)

### Current Behavior

```
$ v run tst.v
tst.v:14:11: error: infix expr: cannot use `string` (right expression) as `int`
   12 |     $for field in Struct.fields {
   13 |         $if field.typ is string || field.typ is string {
   14 |             if tst.$(field.name) == '' {
      |                    ~~~~~~~~~~~~~~~~~~~
   15 |                 panic('fail')
   16 |             }
```

### V version

V 0.4.11 987c211

### Environment details (OS name and version, etc.)

Debian but the bug is reproducible in the playground

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

kbkpbot:
This is because V compiler checker does not eval comptime condition at some case.
I think it may need to rewrite the checker's comptime condition logic.
Basically, I think all comptime condition should be eval at checker, return `.eval` or `.skip`, no `.unknown`.

PR #25122, opened 2025-08-17
checker: rewrite comptime_if_cond() to support comptime if cond evaluate

kbkpbot:
This PR:

1. Rewrite `comptime_if_cond()` in `checker`, fully support comptime if conditon evaluate to `true`/`false`.
2. Add comptime iter $for support for enum/method/method params/sumtype variants/attribute.
3. Fix issue #25099, #24938.

for test `vlib/v/gen/c/testdata/comp_if_unknown.vv`

```v
fn main() {
        mut x := 1
        $if glibc {
                x = 2
        } $else {
                x = 3
        }
        println('done')
}
```

I think the output should contain :

```c
#if defined(__GLIBC__)
x = 2;
#else
#endif
```

instead of :

```c
#if defined(__GLIBC__)
x = 2;
#else
x = 3;
#endif
```

### DIFF
--- a/vlib/v/checker/comptime.v
+++ b/vlib/v/checker/comptime.v
@@ -5,7 +5,6 @@ module checker
 import os
 import v.ast
 import v.pref
-import v.token
 import v.util
 import v.pkgconfig
 import v.type_resolver
@@ -284,9 +283,9 @@ fn (mut c Checker) comptime_for(mut node ast.ComptimeFor) {
 			}
 			has_different_types := fields.len > 1
 				&& !fields.all(c.check_basic(it.typ, fields[0].typ))
+			c.push_new_comptime_info()
 			for field in fields {
 				prev_inside_x_matches_type := c.inside_x_matches_type
-				c.push_new_comptime_info()
 				c.comptime.inside_comptime_for = true
 				if c.field_data_type == 0 {
 					c.field_data_type = c.table.find_type('FieldData')
@@ -310,9 +309,9 @@ fn (mut c Checker) comptime_for(mut node ast.ComptimeFor) {
 							info.size_expr, true)
 					}
 				}
-				c.pop_comptime_info()
 				c.inside_x_matches_type = prev_inside_x_matches_type
 			}
+			c.pop_comptime_info()
 		} else if node.typ != ast.void_type && c.table.generic_type_names(node.typ).len == 0
 			&& sym.kind != .placeholder {
 			c.error('iterating over .fields is supported only for structs and interfaces, and ${sym.name} is neither',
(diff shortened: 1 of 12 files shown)


vlib/v/tests/comptime/comptime_if_glibc_test.v
```diff
@@ -0,0 +1,21 @@
...
+	$if glibc {
+		result += '+glibc'
+		x = 2
+	} $else {
+		result += '-glibc'
+		x = 3
+	}
+
+	$if !glibc {
+		result += '!glibc'
```
spytheman:
```suggestion
		result += '+!glibc'
```
would be clearer imho

vlib/v/tests/comptime/comptime_if_glibc_test.v
```diff
@@ -0,0 +1,21 @@
...
+	} $else {
+		result += '!!glibc'
```
spytheman:
```suggestion
		result += '-!glibc'
```

vlib/v/tests/comptime/comptime_if_glibc_test.v
```diff
@@ -0,0 +1,21 @@
...
+	}
+
+	println(result)
+	assert result == '+glibc!!glibc' || result == '-glibc!glibc'
```
spytheman:
```suggestion
	assert result == '+glibc-!glibc' || result == '-glibc+!glibc'
```

vlib/v/checker/comptime.v
```diff
@@ -284,9 +283,9 @@ fn (mut c Checker) comptime_for(mut node ast.ComptimeFor) {
 			}
 			has_different_types := fields.len > 1
 				&& !fields.all(c.check_basic(it.typ, fields[0].typ))
+			c.push_new_comptime_info()
 			for field in fields {
 				prev_inside_x_matches_type := c.inside_x_matches_type
-				c.push_new_comptime_info()
```
spytheman: that saves a push/pop per iteration, but also has the potential to influence subsequent $for iterations, where the checker settings for each field/attribute etc, could be potentially different from each other.

Imho restore the push/pop as it was, one pair per iteration, not just one per loop.

vlib/v/checker/comptime.v
```diff
@@ -321,31 +320,42 @@ fn (mut c Checker) comptime_for(mut node ast.ComptimeFor) {
...
 	} else if node.kind == .values {
 		if sym.kind == .enum {
-			c.push_new_comptime_info()
-			c.comptime.inside_comptime_for = true
-			if c.enum_data_type == 0 {
-				c.enum_data_type = c.table.find_type('EnumData')
+			if sym.info is ast.Enum {
+				if sym.info.vals.len > 0 {
+					c.push_new_comptime_info()
+					for _ in sym.info.vals {
```
spytheman: same, imho it is safer/less error prone, to do the push/pop once per iteration, not outside the loop

vlib/v/checker/comptime.v
```diff
@@ -321,31 +320,42 @@ fn (mut c Checker) comptime_for(mut node ast.ComptimeFor) {
...
 		} else {
 			c.error('iterating over .values is supported only for enums, and ${sym.name} is not an enum',
 				node.typ_pos)
 			return
 		}
 	} else if node.kind == .methods {
 		methods := sym.get_methods()
-		for method in methods {
+		if methods.len > 0 {
 			c.push_new_comptime_info()
```
spytheman: if you move the push/pop inside the loop, there will be no need for the if too

vlib/v/checker/comptime.v
```diff
@@ -365,26 +365,22 @@ fn (mut c Checker) comptime_for(mut node ast.ComptimeFor) {
 			return
 		}
 		method := c.comptime.comptime_for_method
-		if method.params.len > 0 {
+		for param in method.params[1..] {
```
JalonSolov, line 367: If you don't check the length first, could you get an out-of-bounds error here?

kbkpbot: // example: fn (mut d MyStruct) add(x int, y int) string
		// `d` is params[0], `x` is params[1], `y` is params[2]
		// so we at least has one param (`d`) for method

spytheman (approved):
Excellent work 🙇🏻‍♂️ @kbkpbot .
Thank you very much for working on this.

spytheman:
This PR also has a positive effect on compiling small programs too:
image

PR #25126, opened 2025-08-17
math.big: fix bytes() if bit_len is divisible by 60, add tests

tankf33der:
I came up with an idea to cross `bytes()` and `integer_from_bytes()` to test the operation of both functions. Tests showed that when the length of the input number in bits is divisible by `60`, the assert breaks.

This patch is an attempt to fix this - without the patch, `master` won't pass this test.
Now the loop of 5M iterations executes successfully.


tankf33der:
@kbkpbot - take a look.

PR #25127, opened 2025-08-17
v.pref: fix termux detection

kbkpbot:
Issue found by PR #25122
https://github.com/vlang/v/actions/runs/17017667938/job/48242410975?pr=25122

When cross compile to `termux`, we have no env define for `TERMUX_VERSION`, so the original `termux` detection, by `os.getenv('TERMUX_VERSION') != ''` is not working.

BTW: does V has separate OS definition, that is, `host_os` and `target_os` ?

### DIFF
--- a/vlib/v/pref/os.v
+++ b/vlib/v/pref/os.v
@@ -205,7 +205,7 @@ pub fn (o OS) str() string {
 }
 
 pub fn get_host_os() OS {
-	if os.getenv('TERMUX_VERSION') != '' {
+	$if termux {
 		return .termux
 	}
 	$if android {


spytheman:
This is backwards imho. Doing a comptime check in this place will make it harder to bootstrap V on an actual termux installation.

If `TERMUX_VERSION` is not defined on the CI, but is on a real Termux installation (just checked, and it is), we should make the .yml that emulates the Termux installation define the env variable, so that it is closer to the actual situation where it will be used.

On my termux installation, the value of TERMUX_VERSION is `0.118.3` . I'll update the .yml file.

spytheman:
> BTW: does V has separate OS definition, that is, host_os and target_os ?

afaik, currently it does not, there is only the target_os, which is autodetected based on the host, if `-os` is not explicitly set.

spytheman:
I've merged https://github.com/vlang/v/pull/25128 which sets the env variable.

Issue #24591, opened 2025-05-27
array should has a method pop_left


kbkpbot:
### Describe the feature

`array` should has a method `pop_left`, this is similar with `pop()`, but remove from the first element.

### Use Case

This help building an array as FIFO.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.10 ed7a8a34f1815816984aaa1d2982d2a5fb9a97f2.52c7130

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 ed7a8a34f1815816984aaa1d2982d2a5fb9a97f2.52c7130
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.2 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |6.62GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-05-25 10:51:48
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /media/HD/github/kbkpbot/v/vlib/db/pool
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-165-g0cd14a29
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.4) 2.39

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

kbkpbot:
> We already have https://github.com/vlang/v/blob/master/vlib/datatypes/queue.v

[...]

PR #25133, opened 2025-08-18
builtin: add arr.pop_left() func

kbkpbot:
Feature require by #24591

Just like `array.pop()`,  `array.pop_left()` act like `ret=array.first() array.delete(0)`

### DIFF
--- a/vlib/builtin/array.v
+++ b/vlib/builtin/array.v
@@ -511,7 +511,7 @@ pub fn (a array) last() voidptr {
 	}
 }
 
-// shift returns the first element of the array and removes it by advancing the data pointer.
+// pop_left returns the first element of the array and removes it by advancing the data pointer.
 // If the `array` is empty, this will panic.
 // NOTE: This function:
 //   - Reduces both length and capacity by 1
@@ -524,7 +524,7 @@ pub fn (a array) last() voidptr {
 // ```v
 // mut a := [1, 2, 3, 4, 5]
 // b := unsafe { a[..5] } // full slice view
-// first := a.shift()
+// first := a.pop_left()
 //
 // // Array now starts from second element
 // dump(a) // a: [2, 3, 4, 5]
@@ -541,9 +541,9 @@ pub fn (a array) last() voidptr {
 // a[0] = 99
 // assert b[1] == 99  // changed in both
 // ```
-pub fn (mut a array) shift() voidptr {
+pub fn (mut a array) pop_left() voidptr {
 	if a.len == 0 {
-		panic('array.shift: array is empty')
+		panic('array.pop_left: array is empty')
 	}
 	first_elem := a.data
 	unsafe {
--- a/vlib/builtin/array_d_gcboehm_opt.v
+++ b/vlib/builtin/array_d_gcboehm_opt.v
@@ -216,10 +216,10 @@ fn (mut a array) prepend_many_noscan(val voidptr, size int) {
 	unsafe { a.insert_many_noscan(0, val, size) }
 }
 
-// shift returns the first element of the array and removes it by advancing the data pointer.
-fn (mut a array) shift_noscan() voidptr {
+// pop_left returns the first element of the array and removes it by advancing the data pointer.
+fn (mut a array) pop_left_noscan() voidptr {
 	if a.len == 0 {
-		panic('array.shift: array is empty')
+		panic('array.pop_left: array is empty')
 	}
 	first_elem := a.data
 	unsafe {
--- a/vlib/builtin/array_test.v
+++ b/vlib/builtin/array_test.v
@@ -1197,36 +1197,36 @@ fn test_reverse_in_place() {
 	assert c == [[5, 6], [3, 4], [1, 2]]
 }
 
-fn test_array_int_shift() {
+fn test_array_int_pop_left() {
 	mut a := [1, 2, 3, 4, 5]
 	b := unsafe { a[..5] } // full slice view
 	assert a.len == 5
 	first := a[0]
-	x := a.shift()
+	x := a.pop_left()
 	assert first == x
 	assert a.len == 4
 	assert a.cap == 4
-	y := a.shift()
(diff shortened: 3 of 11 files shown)


JalonSolov:
So... same as `arr.delete(0)`, correct?

JalonSolov:
Ah, alright.

vlib/builtin/array.v
```diff
@@ -511,6 +511,49 @@ pub fn (a array) last() voidptr {
...
+pub fn (mut a array) shift() voidptr {
+	if a.len == 0 {
+		panic('array.shift: array is empty')
+	}
+	first_elem := a.data
+	unsafe {
+		a.data = &u8(a.data) + u64(a.element_size)
+	}
+	a.len--
+	a.cap--
```
spytheman, line 554: I think it may also need to update `a.offset = a.offset + a.element_size` here, otherwise .free() would not get the same pointer, after `a.data` is increased on line 550.

vlib/builtin/array_d_gcboehm_opt.v
```diff
@@ -216,6 +216,20 @@ fn (mut a array) prepend_many_noscan(val voidptr, size int) {
...
+fn (mut a array) shift_noscan() voidptr {
+	if a.len == 0 {
+		panic('array.shift: array is empty')
+	}
+	first_elem := a.data
+	unsafe {
+		a.data = &u8(a.data) + u64(a.element_size)
+	}
+	a.len--
+	a.cap--
```
spytheman, line 230: same, please consider updating .offset here too

spytheman:
I am curious, why is it called .shift() here, and not .pop_left() as in the issue?

vlib/builtin/array.v
spytheman: you can test the need for this with:
```v
@[manualfree]
fn main() {
        mut a := [10, 20, 30, 40]
        println(a.data)
        println(a.offset)
        println(a)
        b := a.shift()
        println(a)
        println(b)
        println(a.data)
        println(a.offset)
        unsafe { a.free() }
}
```
ran with `v -gc none run a.v`, it will:
```
free(): invalid pointer
0x7fe15bc9900b: at ???: RUNTIME ERROR: abort() called
0x7fe15bceb2fc: by ???
```
if .offset is not updated.

spytheman: while with the update, the result will be something like this:
image

spytheman:
I have no opinion on the name. If there is an existing precedent for .shift(), fine.

vlib/builtin/array_d_gcboehm_opt.v
```diff
@@ -125,6 +125,7 @@ fn (mut a array) ensure_cap_noscan(required int) {
 		unsafe { vmemcpy(new_data, a.data, u64(a.len) * u64(a.element_size)) }
 		// TODO: the old data may be leaked when no GC is used (ref-counting?)
 	}
+	a.offset += a.element_size
```
spytheman: no, that is not a good place, the fix should be in the shift method, not here in ensure_cap_noscan

spytheman:
No one updates .offset except the .slice method (for creating a new slice view) and now the .shift one (which in essence also creates a new slice + returning the first element of the old).
-> the free method is correct.

spytheman:
(i.e. the purpose of the .offset field, is that in combination with .data, you can always get back to the original memory block start, even when creating multiple slices/views to the same elements)

JalonSolov:
So...

<table>
<tr><th>Lang</th><th>Function</th></tr>
<tr><td>JavaScript</td><td>shift()</td></tr>
<tr><td>C++</td><td>erase(offset)</td></tr>
<tr><td>Java</td><td>remove(offset)</td></tr>
<tr><td>Python</td><td>del arr[offset]</td></tr>
<tr><td>C#</td><td>RemoveAt(offset)</td></tr>
</table>

Basically, there is no "standard" way of doing this.

blackshirt:
`shift` sounds like the brother of `left_shift` and `right_shift` for me

spytheman:
I think that pop_front or pop_left may be more descriptive and less confusing (but I am fine with shift too).
The erase/deletes that use an offset tend to not return the element, so they are not direct analogs/models for this.

JalonSolov:
To me, `shift`, `left_shift`, and `right_shift` don't make sense.  You're not shifting anything.  An array makes more sense as a stack (think of a stack of plates), so you would `push` things onto the stack, and `pop` them off.

Of course, then you run into... `push` where?  Onto the top of the stack or the bottom?  Etc., etc.

At any rate, as long as it's documented, it doesn't matter that much.  It's more a cognitive thing... what would make the most sense if you didn't know what was there, and you were searching for it?

Last... it seems like it wouldn't hurt anything to just make `arr.delete(offset)` return the value that was deleted.  No new routines to be added, and no breaking changes to the code.

spytheman:
no, delete has a different semantic (and implementation) compared to this method

spytheman:
> You're not shifting anything.

yes, you are shifting the start of the array, `mut a:=[10,20,30]; a.shift(); println(a)` will print `[20,30]` after this method is called.

spytheman:
>  An array makes more sense as a stack

The intention is that this method can be used in the implementation of a queue (avoiding the copying/movement that delete(x) does).

spytheman:
> At any rate, as long as it's documented, it doesn't matter that much. It's more a cognitive thing... what would make the most sense if you didn't know what was there, and you were searching for it?

I agree that documentation can help with clearing the meaning and behavior. The name decision at this point boils down to a balance between mnemonics and clarity.

spytheman:
I tend to prefer `.pop_left()` the more I think about it, because we already do have `.pop()`, and we already do have several other `_left()` suffixed methods in builtin.


i.e. my preference is currently in this order:
1. pop_left() or shift()
3. pop_front()

blackshirt:
`pop_left()` feel the good names ...its describes well on its own

Issue #24893, opened 2025-07-13
Enhance "Identifier not found" when identifier exists but is missing {}


mike-ward:
### Describe the feature

The error message when an identifier is exists but is not used correctly should be enhanced.

### Use Case

```v
struct Data {
	id string
}

type DataCollection = Data

fn main() {
	mut data := DataCollection
}

```

V will return an error message saying the identifier `DataCollection` not found

### Proposed Solution

Since the identifier clearly exists, a better message include:

"Did you mean DataCollection{}?"

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.11 efd36b8

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 da3112e.efd36b8
|:-------------------|:-------------------
|OS                  |macos, macOS, 15.5, 24F74
|Processor           |8 cpus, 64bit, little endian, Apple M2
|Memory              |0.18GB/8GB
|                    |
|V executable        |/Users/mike/Documents/github/v/v
|V last modified time|2025-07-13 18:06:18
|                    |
|V home dir          |OK, value: /Users/mike/Documents/github/v
|VMODULES            |OK, value: /Users/mike/.vmodules
|VTMP                |OK, value: /tmp/v_501
|Current working dir |OK, value: /Users/mike/Documents/github/bug
|                    |
|env VFLAGS          |"-path @vlib|@vmodules|/Users/mike/Documents/github -message-limit 5 -no-parallel"
|Git version         |git version 2.50.0
|V git status        |weekly.2025.25-118-gefd36b89
|.git/config present |true
|                    |
|cc version          |Apple clang version 17.0.0 (clang-1700.0.13.5)
|gcc version         |Apple clang version 17.0.0 (clang-1700.0.13.5)
|clang version       |Apple clang version 17.0.0 (clang-1700.0.13.5)
|tcc version         |tcc version 0.9.28rc 2024-02-05 HEAD@105d70f7 (AArch64 Darwin)
|tcc git status      |thirdparty-macos-arm64 c8df4e27
|emcc version        |N/A
|glibc version       |N/A

JalonSolov:
Actually, the message should be `DataCollection must be initialized`.  There's nothing "missing" except the `{}` to initialize it.

PR #25134, opened 2025-08-18
checker: add error for struct not init

Krchi:
#24893 add error for struct not init
modify some tests

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -4377,7 +4377,7 @@ fn (mut c Checker) ident(mut node ast.Ident) ast.Type {
 						}
 					}
 				}
-				if !c.check_is_struct_name(node) {
+				c.check_known_struct_name(node) or {
 					c.error(util.new_suggestion(node.name, const_names_in_mod).say('undefined ident: `${node.name}`'),
 						node.pos)
 				}
@@ -4400,7 +4400,7 @@ fn (mut c Checker) ident(mut node ast.Ident) ast.Type {
 					}
 				}
 
-				if !c.check_is_struct_name(node) {
+				c.check_known_struct_name(node) or {
 					c.error('undefined ident: `${node.name}`', node.pos)
 				}
 			}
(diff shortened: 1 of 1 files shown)


vlib/v/checker/checker.v
```diff
@@ -4398,13 +4400,32 @@ fn (mut c Checker) ident(mut node ast.Ident) ast.Type {
...
+				}
 			}
 		}
 	}
 	return ast.void_type
 }
 
+fn (mut c Checker) check_is_struct_name(ident ast.Ident) bool {
+	mut is_struct := false
+	if ident.mod == 'builtin' || ident.mod == 'main' {
```
spytheman, line 4414: why is there a special case for main here?
would not the else clause handle it?

vlib/v/checker/checker.v
```diff
@@ -4398,13 +4400,32 @@ fn (mut c Checker) ident(mut node ast.Ident) ast.Type {
...
+				if !c.check_is_struct_name(node) {
+					c.error('undefined ident: `${node.name}`', node.pos)
+				}
 			}
 		}
 	}
 	return ast.void_type
 }
 
+fn (mut c Checker) check_is_struct_name(ident ast.Ident) bool {
```
spytheman:
```suggestion
fn (mut c Checker) check_known_struct_name(ident ast.Ident) ? {
```
then just `return none` at the end, instead of false, and the callers can become:
`c.check_known_struct_name() or { c.error() }`, without `if !f() {`

vlib/v/checker/checker.v
Krchi: > why is there a special case for main here? would not the else clause handle it?
if not, this will happen
examples/issues/i24893/main.v:14:11: error: undefined ident: `v.ast.Type`

Krchi: > why is there a special case for main here? would not the else clause handle it?
if not, this will happen
examples/issues/i24893/main.v:14:11: error: undefined ident: `v.ast.Type`

spytheman: thank you

spytheman (approved):
Excellent work, thank you.

PR #25138, opened 2025-08-19
math.big: add_digit_array() refactoring

tankf33der:
The first of two patches for addition.

It all started when I found that there was no fast exit from the function when the input array was empty. Currently there are three places in the library that use this function - `subtract`, `addition` and `Karatsuba`.

Simple arithmetic has protection and doesn't let empty arrays pass through, but `Karatsuba` does.

The patch contains three blocks of changes:

1. just protection for the future if someone uses the function incorrectly
2. `Karatsuba` can pass an empty array, it is necessary to do `shrink`, without `shrink` tests will fail
3. Here the point is to remove the unnecessary `if` and `delete_last()`

Multi-million loop passed against 'gmplib`.


spytheman (approved):
Thank you 🙇🏻‍♂️.

PR #25139, opened 2025-08-19
math.big: optimization of add_digit_array()

tankf33der:
Last part of refactoring.

The patch contains two parts: in the first part, I remove unnecessary `if` statements in `imin` and `imax` because I already have everything needed for range knowledge; in the second part, I remove unnecessary `u64()` conversions.

Patch passed multi-million loop against `gmplib`.

### DIFF
--- a/vlib/math/big/array_ops.v
+++ b/vlib/math/big/array_ops.v
@@ -45,28 +45,25 @@ fn add_digit_array(operand_a []u64, operand_b []u64, mut sum []u64) {
 		return
 	}
 
-	// First pass intersects with both operands
-	smaller_limit := imin(operand_a.len, operand_b.len)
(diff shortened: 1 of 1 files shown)


spytheman:
Is there a speedup?

spytheman (approved):
Excellent work.

tankf33der:
> Is there a speedup?

```v
import math.big
import rand
import time

fn main() {
        for n := 1; n < 40960; n += 18 {
                a_str := rand.string_from_set('0123456789', n)
                a := big.integer_from_string(a_str)!
                t1 := time.now()
                for _ in 0 .. 512 {
                        _ := a + a
                }
                println('${n} ${time.since(t1).seconds()}')
        }
}
```

image

PR #25146, opened 2025-08-20
checker: ensure the defer behavior matches that of cgen

kbkpbot:
This PR is try to ensure the defer behavior matches that of `cgen`.

In `cgen`, defer stmts may be copied into a `$for` iteration. This is different with `checker`.
For example:
```v
module main

struct Defer {
        a int
        b int
}

fn main() {
        x := Defer{}
        mut inside_for_flag := false
        defer {
                // g.comptime.inside_comptime_for = true here
                // but c.comptime.inside_comptime_for = false here
                println('inside_for_flag = ${inside_for_flag}')
        }

        $for f in x.fields {
                // g.comptime.inside_comptime_for == true here
                return
        }
        inside_for_flag = false
}
```

Note that in the `defer` block, `inside_comptime_for` in `checker` and `cgen` is different.
I can't provide a test now, maybe in another PR.

### DIFF
--- a/vlib/v/checker/return.v
+++ b/vlib/v/checker/return.v
@@ -33,6 +33,12 @@ fn (mut c Checker) return_stmt(mut node ast.Return) {
 	defer {
 		c.inside_return = prev_inside_return
 	}
+
+	// check `defer_stmts` in return, to ensure the same behavior with `cgen`
+	for i := c.table.cur_fn.defer_stmts.len - 1; i >= 0; i-- {
+		c.stmts(mut c.table.cur_fn.defer_stmts[i].stmts)
+	}
+
 	c.expected_type = c.table.cur_fn.return_type
 	mut expected_type := c.unwrap_generic(c.expected_type)
 	if expected_type != 0 && c.table.sym(expected_type).kind == .alias {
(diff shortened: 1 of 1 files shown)


spytheman (commented):
That will check the defer statements many times, if there are several returns, in addition to the check that is done for the defer {} block itself.

Is there a specific context where that is beneficial/needed?

spytheman:
note that the checker does not care (it can not) about the actual runtime values of the statements in the defer block, neither does cgen, and the defer block in your example is outside the $for loop, so I do not understand how checking it several times (one per iteration in the $for body), will change things

spytheman:
do you want to use a comptime value from the $for block somehow in the defered statements?

spytheman:
Very good explanations. I have only a few remarks:

>    Identical defer logic gets replicated at every return point, causes ​exponential growth​ in generated code size.

It is not exponential, it is linear in the number of returns per function.

>   So, is is possible replace return-point duplication with a ​unified exit pathway?

It may be, but I have not thought much about how it can affect the planed change for the defers to become block scoped, instead of function scoped. I guess we can just change the placement for the goto targets to the end of the corresponding scope, instead of the end of the entire function 🤔 .

spytheman:
`for i in 0..3 {` already defines a scope, imho there is no need for the nested one

spytheman (approved):
Excellent work.

Issue #25123, opened 2025-08-17
comptime if does not work as expected


kbkpbot:
### Describe the bug

It seems `cgen` has some bug when generate code for some comptime `$if`.

### Reproduction Steps

some code extract from `vlib/v/tests/comptime/comptime_if_expr_test.v`

version.v
```v
const version = 123
const disable_opt_features = true

// Note: the `unknown_fn()` calls are here on purpose, to make sure that anything
// that doesn't match a compile-time condition is not even parsed.
fn ct_expressions() {
        foo := version
        bar := foo
        $if bar == 123 {
                println('bar == 123')
                assert true
        }
        $if bar != 123 {
                unknown_fn()
        } $else $if bar != 124 {
                println('bar != 124')
                assert true
        } $else {
                unknown_fn()
        }
}

fn main() {
        ct_expressions()
}
```

```
v run version.v
```

### Expected Behavior

output
```
bar == 123
bar != 124
```

### Current Behavior

output
```
bar == 123
```

### V version

V 0.4.11 460b48a33e3842adcee4a247bcf67ab79f272b8c.4f8a3da

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 460b48a33e3842adcee4a247bcf67ab79f272b8c.4f8a3da
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |3.38GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-08-17 01:46:38
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/comptime
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-589-g206fdf4a
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
[...]

PR #25150, opened 2025-08-21
checker,cgen: evaluate comptime if results in checker only

kbkpbot:
This PR propagate comptime if computation results from `checker` to `cgen`, then we can maintain only one copy of  `comptime_if_cond()`.

Fix a bug in `vlib/v/tests/comptime/comptime_selector_member_test.v`
Fix issue #25123, #25156

NOTE: Currently `hash_stmt()` only support top level  `$if` branch without `$else` or `$else $if` branch.

### DIFF
--- a/vlib/v/gen/c/comptime.v
+++ b/vlib/v/gen/c/comptime.v
@@ -430,7 +430,10 @@ fn (mut g Gen) comptime_if(node ast.IfExpr) {
 			}
 			// directly use `checker` evaluate results
 			// for `cgen`, we can use `is_true.c_str` or `is_true.value` here
-			g.writeln('${is_true.c_str}\t/* ${node.branches[i].cond} | generic=[${comptime_branch_context_str}] */')
+			g.writeln('${is_true.c_str}')
+			$if debug_comptime_branch_context ? {
+				g.writeln('/* ${node.branches[i].cond} | generic=[${comptime_branch_context_str}] */')
+			}
 		} else {
 			g.writeln('#else')
 		}
--- a/vlib/v/tests/comptime/comptime_test_ident_test.v
+++ b/vlib/v/tests/comptime/comptime_test_ident_test.v
@@ -30,6 +30,5 @@ fn test_test_ident() {
 		result += '0'
 	}
 
-	// TODO: correct result should be '13689', it may fix for vtest or builder
-	assert result == '3689'
+	assert result == '23689'
 }
--- a/vlib/v/checker/comptime.v
+++ b/vlib/v/checker/comptime.v
@@ -784,7 +784,7 @@ fn (mut c Checker) evaluate_once_comptime_if_attribute(mut node ast.Attr) bool {
 	return node.ct_skip
 }
 
-fn (mut c Checker) comptime_if_to_ifdef(name string, is_comptime_option bool) !string {
+fn (mut c Checker) comptime_if_to_ifdef(name string) !string {
 	match name {
 		// platforms/os-es:
 		'windows' {
@@ -964,15 +964,9 @@ fn (mut c Checker) comptime_if_to_ifdef(name string, is_comptime_option bool) !s
 			// turned on by: `-cflags -ffast-math`
 			return '__FAST_MATH__'
 		}
-		else {
-			if is_comptime_option
-				|| (c.pref.compile_defines_all.len > 0 && name in c.pref.compile_defines_all) {
-				return 'CUSTOM_DEFINE_${name}'
-			}
-			return error('bad os ifdef name "${name}"') // should never happen, caught in the checker
-		}
+		else {}
 	}
-	return error('none')
+	return error('bad os ifdef name "${name}"')
 }
 
 fn (mut c Checker) get_expr_type(cond ast.Expr) ast.Type {
@@ -1105,11 +1099,11 @@ fn (mut c Checker) comptime_if_cond(mut cond ast.Expr, mut sb strings.Builder) (
 			should_record_ident = true
 			is_user_ident = true
 			ident_name = cname
-			ifdef := c.comptime_if_to_ifdef(cname, true) or {
-				c.error(err.msg(), cond.pos)
-				return false, false
-			}
-			sb.write_string('defined(${ifdef})')
+			// ifdef := c.comptime_if_to_ifdef(cname, true) or {
+			//	c.error(err.msg(), cond.pos)
+			//	return false, false
+			//}
+			sb.write_string('defined(CUSTOM_DEFINE_${cname})')
 			is_true = cname in c.pref.compile_defines
 			return is_true, false
 		}
@@ -1655,7 +1649,7 @@ fn (mut c Checker) comptime_if_cond(mut cond ast.Expr, mut sb strings.Builder) (
 				c.error('invalid \$if condition: unknown indent `${cname}`', cond.pos)
 				return false, false
 			}
-			if ifdef := c.comptime_if_to_ifdef(cname, false) {
+			if ifdef := c.comptime_if_to_ifdef(cname) {
 				sb.write_string('defined(${ifdef})')
 			} else {
 				sb.write_string('${is_true}')
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -1029,16 +1029,16 @@ pub fn (mut g Gen) init() {
 	if g.pref.gc_mode in [.boehm_full, .boehm_incr, .boehm_full_opt, .boehm_incr_opt, .boehm_leak] {
 		g.comptime_definitions.writeln('#define _VGCBOEHM (1)')
 	}
-	if g.pref.is_debug || 'debug' in g.pref.compile_defines {
+	if g.pref.is_debug {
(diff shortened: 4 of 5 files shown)


JalonSolov:
The problem with moving things to cgen is that now that only works with the C backend.

Things that are in cgen now should be moving OUT (like the auto-magic `str()` method creation, etc.), nothing else should be moving IN (unless it is extremely specific to C).

vlib/v/tests/comptime/comptime_test_ident_test.v
```diff
@@ -0,0 +1,35 @@
...
+		result += '8'
+	}
+
+	$if $d('test', true) {
+		result += '9'
+	} $else {
+		result += '0'
+	}
+
+	// TODO: correct result should be '13689', it may fix for vtest or builder
```
spytheman: No.

The correct result should be `23689`. `test?` is different than `test`.

`test?` becomes true if `-d test` is passed.
`test` becomes true, when a _test.v file is compiled.
i.e. it is not passed through a -d flag by the test runner, it is defined by V itself,
and you can check with:
```v
fn test_abc() {
   $if test ? {
          println('ok test ?')
   }
   $if test {
          println('ok test')
   }
   println('done')
}
```

with `v a_test.v`, should show:
```
ok test
done
```

The bug is that the $else branch for `$if test ? {` was not evaluated, but should have.

kbkpbot: So , there must be some mix understanding of system defines and user custom defines.
for example, in `vlib/v/gen/c/cgen.v`, in `pub fn (mut g Gen) init()` function:
```v
        if g.pref.gc_mode in [.boehm_full, .boehm_incr, .boehm_full_opt, .boehm_incr_opt, .boehm_leak] {
                g.comptime_definitions.writeln('#define _VGCBOEHM (1)')
        }
        if g.pref.is_debug || 'debug' in g.pref.compile_defines {
                g.comptime_definitions.writeln('#define _VDEBUG (1)')
        }
        if g.pref.is_prod || 'prod' in g.pref.compile_defines {
                g.comptime_definitions.writeln('#define _VPROD (1)')
        }
        if g.pref.is_test || 'test' in g.pref.compile_defines {
                g.comptime_definitions.writeln('#define _VTEST (1)')
        }
        if g.pref.is_prof || 'profile' in g.pref.compile_defines {
                g.comptime_definitions.writeln('#define _VPROFILE (1)')
        }
        if g.pref.autofree {
                g.comptime_definitions.writeln('#define _VAUTOFREE (1)')
        }
        if g.pref.prealloc {
                g.comptime_definitions.writeln('#define _VPREALLOC (1)')
        }
        if g.pref.use_cache {
                g.comptime_definitions.writeln('#define _VUSECACHE (1)')
        }
        if g.pref.build_mode == .build_module {
                g.comptime_definitions.writeln('#define _VBUILDMODULE (1)')
        }
        if g.pref.is_livemain || g.pref.is_liveshared {
                g.generate_hotcode_reloading_declarations()
        }
...
```

and in `.vmodules/vsl/plot/show.v`, in function `pub fn (p Plot) show(config PlotConfig) ! {`
```v
        $if test ? {
                println('Ignoring plot.show() because we are running in test mode')
        } $else {
                mut handler := PlotlyHandler{
                        use_cdn: true
                        plot:    p
                }
....
```

spytheman: Yes, there is mixed understanding, ignore vsl , it should be fixed there, currently it is not showing anything in both branches, because of the bug.

spytheman: `-d test` can be used to check the test mode without moving code in a _test.v file.

Compiling/running a `_test.v` file, sets `pref.is_test`, and makes `$if test {` pass.

vlib/v/gen/c/comptime.v
```diff
@@ -346,32 +400,39 @@ fn (mut g Gen) comptime_if(node ast.IfExpr) {
...
-			comptime_if_stmts_skip, comptime_may_skip_else = g.comptime_if_cond(branch.cond,
-				branch.pkg_exist)
-			if !comptime_if_stmts_skip && comptime_may_skip_else {
-				comptime_may_skip_else = false // if the cond is false, not skip else branch
-			}
-			comptime_if_stmts_skip = !comptime_if_stmts_skip
-			g.writeln('')
+			// directly use `checker` evaluate results
+			// for `cgen`, we can use `is_true.c_str` or `is_true.value` here
+			g.writeln('${is_true.c_str}\t/* ${node.branches[i].cond} | generic=[${comptime_branch_context_str}] */')
```
spytheman:
```suggestion
			g.writeln('${is_true.c_str}')
			$if debug_comptime_branch_context ? {
				g.writeln('/* ${node.branches[i].cond} | generic=[${comptime_branch_context_str}] */')
			}
```

spytheman: The debug comments are relatively big, compared to small programs (most tests).
For the hello world example:

Image

spytheman: Also:

Image

vlib/v/tests/comptime/comptime_test_ident_test.v
kbkpbot: fixed and waiting vsl fix `test ?` or the CI can't pass

vlib/v/gen/c/comptime.v
kbkpbot: fixed

spytheman:
> Things that are in cgen now should be moving OUT

Yes, that is why I like this PR a lot, it moved complexity out of cgen.

spytheman:
@kbkpbot I've reverted the change in 2124452, since it breaks bootstrapping (the generated vc/v.c can not be compiled at its second stage on macos, freebsd etc).

You can check what needs to be fixed with the following example:
```v
import os
dump(os.user_os())
```

Compile it with:
```
./v -cross -printfn _v_malloc run x.v
```

It should print something like this:
```c
#0 01:50:49 ^ master ~/v>./v -cross -printfn _v_malloc run x.v         
vlib/os/os.c.v:530:16: warning: method `u8.free` must be called from an `unsafe` block
  528 |             initial_size := 256 * wide_char_size
  529 |             mut buf := malloc_noscan(initial_size)
  530 |             defer { buf.free() }
      |                         ~~~~~~
  531 |             mut capacity := initial_size
  532 |             mut offset := 0
u8* _v_malloc(isize n) {
        #if defined(CUSTOM_DEFINE_trace_malloc)
        {
                total_m += n;
                fprintf(stderr, "_v_malloc %6d total %10d\n", n, total_m);
        }
        #endif
        if (n < 0) {
                _memory_panic(_S("malloc"), n);
                VUNREACHABLE();
        } else if (n == 0) {
                return ((u8*)(((void*)0)));
        }
        u8* res = ((u8*)(((void*)0)));
        #if defined(_VPREALLOC)
        {
                return prealloc_malloc(n);
        }
        #elif defined(_VGCBOEHM)
        {
                { // Unsafe block
                        res = GC_MALLOC(n);
                }
        }
        #elif defined(_VFREESTANDING)
        {
        }
        #else
        {
                #if defined(_WIN32)
                {
                        res = _aligned_malloc(n, 1);
                }
                #else
                {
                        res = malloc(n);
                }
                #endif
        }
        #endif
        if (res == 0) {
                _memory_panic(_S("malloc"), n);
                VUNREACHABLE();
        }
        #if defined(CUSTOM_DEFINE_debug_malloc)
        {
                memset(res, 0x4D, n);
        }
        #endif
        return res;
}

[x.v:2] os.user_os(): linux
```

Note that with `-cross` *all ifdef branches for the OS checks* should be present.


Before the revert, it produced the much trimmed:
```c
#0 01:53:38 ^ (HEAD detached at cee72f71a5) ~/v>./v -w -cross -printfn _v_malloc x.v
u8* _v_malloc(isize n) {
        if (n < 0) {
                _memory_panic(_S("malloc"), n);
                VUNREACHABLE();
        } else if (n == 0) {
                return ((u8*)(((void*)0)));
        }
        u8* res = ((u8*)(((void*)0)));
        #if defined(_VPREALLOC)
        {
        }
        #elif defined(CUSTOM_DEFINE_gcboehm)
        {
        }
        #elif defined(_VFREESTANDING)
        {
        }
        #else
        {
                #if defined(_WIN32)
                {
                }
                #else
                {
                        res = malloc(n);
                }
                #endif
        }
        #endif
        if (res == 0) {
                _memory_panic(_S("malloc"), n);
                VUNREACHABLE();
        }
        return res;
}
```

spytheman:
Note: I've tried `mut keep_stmts := c.pref.output_cross_c` in `vlib/v/checker/comptime.v` and later `return is_true, keep_stmts` in the ast.Ident branch in `comptime_if_cond`, but that only restored the ifdef's for the missing branches, not their bodies.

The bodies are presumably removed somewhere else, that I could not find right away.

spytheman:
> It seems this PR has been closed, should I create another PR?

Yes, please make another PR.

Github closes them after merging and I could not find a way to re-open it after the revert.

PR #25154, opened 2025-08-22
math.big: restore and refactor str() and integer_from_string()

tankf33der:
To transition to a 60-bit base for `math.big`, it is necessary to return and rewrite the code for `str()` and `integer_from_string()`. This is also a good opportunity to additionally perform refactoring. 

Tests show that for multi-thousand digit numbers, `str()` will now be `~14x` faster, and `integer_from_string()` `~11x` faster.

The repo already contains the necessary tests for all `radix`.

Patch passed tests against `gmplib`.

### DIFF
--- a/vlib/math/big/integer.v
+++ b/vlib/math/big/integer.v
@@ -1,8 +1,19 @@
 module big
 
+import math
 import math.bits
+import strings
 
 const digit_array = '0123456789abcdefghijklmnopqrstuvwxyz'.bytes()
+// vfmt off
+const radix_options = {
+	2: 59, 3: 37, 4: 29, 5: 25, 6: 23, 7: 21, 8: 19, 9: 18, 10: 18,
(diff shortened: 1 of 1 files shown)


vlib/math/big/integer.v
```diff
@@ -1,8 +1,19 @@
 module big
 
+import math
 import math.bits
+import strings
 
 const digit_array = '0123456789abcdefghijklmnopqrstuvwxyz'.bytes()
+// vfmt off
+const radix_options = {
```
spytheman, line 9: The keys seem to have very low cardinality, would not an array instead of a map be better? It can eliminate one hash per key lookup, and the array data will have better cache locality.

spytheman: Although `radix_options` does not seem to be used in loops, so it may not matter much.

spytheman (approved):
Excellent. 
Thank you 🙇🏻‍♂️.

Issue #25156, opened 2025-08-23
v test generate wrong result


kbkpbot:
### Describe the bug

when source code including `$if test ?` , it will generate wrong C code.

### Reproduction Steps

comptime_test_ident_test.v
```v
fn test_test_ident() {
        mut result := ''
        $if test ? {
                result += '1'
        } $else {
                result += '2'
        }

        $if test {
                result += '3'
        } $else {
                result += '4'
        }

        $if !test {
                result += '5'
        } $else {
                result += '6'
        }

        $if $d('test', false) {
                result += '7'
        } $else {
                result += '8'
        }

        $if $d('test', true) {
                result += '9'
        } $else {
                result += '0'
        }

        assert result == '23689'
}
```

```
v test comptime_test_ident_test.v
```

### Expected Behavior

should pass assertion

### Current Behavior

```
$ v test comptime_test_ident_test.v
---- Testing... ---------------------------------------------------------------------------------------------------------------
 FAIL     1.142 ms /media/HD/github/kbkpbot/v/vlib/v/tests/comptime/comptime_test_ident_test.v
/media/HD/github/kbkpbot/v/vlib/v/tests/comptime/comptime_test_ident_test.v:33: fn test_test_ident
   > assert result == '23689'
     Left value (len: 4): `3689`
    Right value (len: 5): `23689`
-------------------------------------------------------------------------------------------------------------------------------
To reproduce just failure 1 run:    '/media/HD/github/kbkpbot/v/v'  '/media/HD/github/kbkpbot/v/vlib/v/tests/comptime/comptime_test_ident_test.v'
Summary for all V _test.v files: 1 failed, 1 total. Elapsed time: 207 ms, on 1 job. Comptime: 205 ms. Runtime: 1 ms.

```

### Possible Solution

Fix pref?

### V version

V 0.4.11 9a93e822390c79f324648b2c7555200025d2212f.2e23f2a

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 9a93e822390c79f324648b2c7555200025d2212f.2e23f2a
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |5.94GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-08-23 00:39:37
|                    |
[...]

PR #25162, opened 2025-08-24
checker,cgen: evaluate comptime if results in checker only

kbkpbot:
Fix for PR #25150, more discussion can be found there.

This PR propagate comptime if computation results from `checker` to `cgen`, then we can maintain only one copy of `comptime_if_cond()`.

Fix a bug in vlib/v/tests/comptime/comptime_selector_member_test.v
Fix issue https://github.com/vlang/v/issues/25123, https://github.com/vlang/v/issues/25156

NOTE: Currently hash_stmt() only support top level $if branch without $else or $else $if branch.


spytheman:
> Add test for -cross mode generated C file.

Thank you.

Issue #23629, opened 2025-02-02
The first line after `// vfmt off` is replaced with `// vfmt on` if the first line ends with a comment


revosw:
### Describe the bug

The first line after `// vfmt off` is replaced with `// vfmt on` if the first line ends with a comment.

### Reproduction Steps

Code: https://play.vlang.io/p/472fc92506

```v
fn main() {
	// vfmt off
	num := 4 // some comment
	// vfmt on
}
```

Format the code using the format button in the playground link, or run `v fmt -w` on the file

### Expected Behavior

The code is left untouched:

```v
fn main() {
	// vfmt off
	num := 4 // some comment
	// vfmt on
}
```

### Current Behavior

Output:

```v
fn main() {
	// vfmt off
	// vfmt on
	// vfmt on
}
```

### V version

V 0.4.9 b91bbad

### Environment details (OS name and version, etc.)

```
|V full version      |V 0.4.9 b91bbad
|:-------------------|:-------------------
|OS                  |linux, Debian GNU/Linux 12 (bookworm) (VM)
|Processor           |2 cpus, 64bit, little endian, Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz
|Memory              |1.24GB/2.02GB
|                    |
|V executable        |/home/admin/v/v
|V last modified time|2025-01-31 21:25:26
|                    |
|V home dir          |OK, value: /home/admin/v
|VMODULES            |OK, value: .vmodules
|VTMP                |OK, value: /tmp/v_0
|Current working dir |OK, value: /home/admin/playground
|                    |
|Git version         |git version 2.39.5
|V git status        |Error: fatal: detected dubious ownership in repository at '/home/admin/v'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v
|.git/config present |true
|                    |
|cc version          |cc (Debian 12.2.0-14) 12.2.0
|gcc version         |gcc (Debian 12.2.0-14) 12.2.0
|clang version       |Debian clang version 14.0.6
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
|tcc git status      |Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v/thirdparty/tcc
 Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
[...]

PR #25163, opened 2025-08-24
fmt: fix comment line number in file with crlf line separator

### LINKED ISSUE #23524: vfmt inconsistency between CRLF and LF end of line
### Describe the bug

If u "disable" fmt with `// vfmt off` and after reenable it with `// vfmt on` on a block on a file with CRLF as end of line `v fmt` delete the line under  `// vfmt off` and add a new `// vfmt on` after the already existing `// vfmt on` breaking the code.
On a file with LF as eol this problem doesn't subsist.

### Reproduction Steps

Create a simple file like:

```
module main

fn main() {
	// vfmt off
	a := [
		1, 2, 3,
		4, 5, 6,
	]
	// vfmt on

	println('Hello World!')
}
```

And set the eol as CRLF

### Expected Behavior

Running `v fmt .` nothing change as all is already formatted and/or you want to keep the style as is.
`v fmt` should not act differently based of the eol.

### Current Behavior

With CRLF eol.
Running `v fmt .` (from ide or terminal) break the code returning this:

```
fn main() {
        // vfmt off
                1, 2, 3,
                4, 5, 6,
        ]
        // vfmt on
        // vfmt on
        println('Hello World!')
}
```

### Possible Solution

_No response_

Setting the file to LF eol resolve the issue.
I dont think that the problem could be the terminal used (Win Terminal) because using `v fmt -w .` write the wrong output on the file.

### V version

V 0.4.9 2aa5651.fda0af4

|V full version      |V 0.4.9 2aa5651.fda0af4
|OS                  |windows, Microsoft Windows 11 Pro 22621 64-bit
|Processor           |24 cpus, 64bit, little endian, 13th Gen Intel(R) Core(TM) i7-13700K
|Memory              |22.48GB/31.78GB
|                    |
|V executable        |E:\Coding\v\v\v.exe
|V last modified time|2025-01-19 13:39:47
|                    |
|V home dir          |OK, value: E:\Coding\v\v
|VMODULES            |OK, value: C:\Users\Alpha\\.vmodules
|VTMP                |OK, value: C:\Users\Alpha\AppData\Local\Temp\v_0
|Current working dir |OK, value: E:\Coding\v\testfmt
|                    |
|Git version         |git version 2.47.1.windows.1
|V git status        |weekly.2025.1-84-gfda0af48
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |N/A
|clang version       |N/A
|msvc version        |N/A
|tcc git status      |N/A
|emcc version        |N/A
|glibc version       |N/A

Krchi:
fix #23524
file with crlf line separator will replace with lf after format

### DIFF
--- a/vlib/v/scanner/scanner.v
+++ b/vlib/v/scanner/scanner.v
@@ -1068,15 +1068,11 @@ pub fn (mut s Scanner) text_scan() token.Token {
 					mut comment_line_end := s.pos
 					if s.text[s.pos - 1] == b_cr {
 						comment_line_end--
-						if s.text[s.pos] == b_lf {
-							s.pos = s.pos - 2
-							s.line_nr--
-						}
-					} else {
-						// fix line_nr, \n was read; the comment is marked on the next line
 						s.pos--
-						s.line_nr--
 					}
+					// fix line_nr, \n was read; the comment is marked on the next line
+					s.pos--
+					s.line_nr--
 					if s.should_parse_comment() {
 						s.line_comment = s.text[start + 1..comment_line_end]
 						mut comment := s.line_comment
--- a/vlib/v/scanner/scanner.v
+++ b/vlib/v/scanner/scanner.v
@@ -1068,6 +1068,10 @@ pub fn (mut s Scanner) text_scan() token.Token {
 					mut comment_line_end := s.pos
 					if s.text[s.pos - 1] == b_cr {
 						comment_line_end--
+						if s.text[s.pos] == b_lf {
+							s.pos = s.pos - 2
+							s.line_nr--
+						}
 					} else {
 						// fix line_nr, \n was read; the comment is marked on the next line
 						s.pos--


spytheman:
Please add a test case.

spytheman:
The test can be a pair of `xxx_input.vv/xxx_expected.vv` files, where the input version has the CRLF line endings (you can check with `xxd`).

spytheman:
You will probably also need to add an `vlib/v/fmt/tests/.gitattributes` file, similar to `./vlib/v/checker/tests/.gitattributes`, which contains:
```
*_crlf_*  binary
*_lf_*    binary
```
so that git will then know to not auto convert the CRLFs in the input to just LFs

spytheman:
> #23629 it could be same problem

It could be, although I think that it is playground specific.
If I remember correctly it had to do with how the JS code for the frontend there treated the server responses, but I could be wrong.

spytheman:
hm, sorry; I will add the test, since it mostly works locally

Krchi:
> hm, sorry; I will add the test, since it mostly works locally

You are right, this test case is hard for me

spytheman:
It was indeed tricky to add, since I also had a git hook to always run vfmt on the changed .vv files :-) too; I had to `git commit --no-verify` to avoid it explicitly...

vlib/v/scanner/scanner.v
```diff
@@ -1068,11 +1068,11 @@ pub fn (mut s Scanner) text_scan() token.Token {
...
 					if s.text[s.pos - 1] == b_cr {
 						comment_line_end--
-					} else {
-						// fix line_nr, \n was read; the comment is marked on the next line
 						s.pos--
-						s.line_nr--
 					}
+					// fix line_nr, \n was read; the comment is marked on the next line
+					s.pos--
+					s.line_nr--
```
spytheman, line 1075: I am not sure if this `s.line_nr--` would not mess up the line numbers for errors in the non fmt case, but the error tests do seem to pass 🤔.

spytheman: manual testing also seems to be fine:

Image

spytheman (approved):
Excellent work!

Issue #25100, opened 2025-08-13
comptime should have $match


JalonSolov:
### Describe the feature

comptime `$match` for the same reason there is a regular `match` - to avoid long chains of `$if`, `$if $else`, `$else` if there are multiple things to handle.

### Use Case

Any time there are multiple different branches to be handled, a `$match` would be cleaner than a `$if/$else` tree.

### Proposed Solution

Add `$match` to comptime.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

Not that it matters, but `V 0.4.11 e5100a8`

### Environment details (OS name and version, etc.)

Also doesn't matter...
```
|V full version      |V 0.4.11 f581bb7.e5100a8
|:-------------------|:-------------------
|OS                  |linux, "Manjaro Linux"
|Processor           |32 cpus, 64bit, little endian, AMD Ryzen 9 7950X 16-Core Processor
|Memory              |7.25GB/30.49GB
|                    |
|V executable        |/home/jalon/git/v/v
|V last modified time|2025-08-12 16:56:21
|                    |
|V home dir          |OK, value: /home/jalon/git/v
|VMODULES            |OK, value: /home/jalon/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/jalon
|                    |
|Git version         |git version 2.50.1
|V git status        |weekly.2025.33-9-ge5100a8f
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.1.1 20250729
|gcc version         |gcc (GCC) 15.1.1 20250729
|clang version       |clang version 20.1.8
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 0134e9b9
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.42
```

kbkpbot:
should we add some comptime var to V ? 
for example, `$os` indicate current os, in one of the valid os [windows, linux, macos, darwin, ios, bsd, freebsd, openbsd, netbsd, android, mach, dragonfly, gnu, hpux, haiku, qnx, solaris, termux, serenity, vinix, plan9].

also, we need `$compiler`, `$platform`..

then we can write some code like this:

```v
$match $os {
    windows {
    }
    linux {
    }
    macos {
    }
    $else {
    }
}
```

[...]

PR #25165, opened 2025-08-25
checker,cgen: add comptime match support

kbkpbot:
Feature require by #25100

This PR support two kinds of `$match`, a type `$match` and a value `$match`.

ex:

```v
$match T {
   int { }
   string {}
   $else{}    // `$else` is optional
}
```

```v
$match x {
   1 {}
   2 {}
   3 {}
}
```

also, support assignment :

```v
   x := $match val {
            1 { '1' }
            2 { '2' }
            $else { 'unknown' }
   }
```

### DIFF
--- a/vlib/v/checker/comptime.v
+++ b/vlib/v/checker/comptime.v
@@ -969,6 +969,15 @@ fn (mut c Checker) comptime_if_to_ifdef(name string) !string {
 	return error('bad os ifdef name "${name}"')
 }
 
+// check if `ident` is a function generic, such as `T`
+fn (mut c Checker) is_generic_ident(ident string) bool {
+	if !isnil(c.table.cur_fn) && ident in c.table.cur_fn.generic_names
+		&& c.table.cur_fn.generic_names.len == c.table.cur_concrete_types.len {
+		return true
+	}
+	return false
+}
+
 fn (mut c Checker) get_expr_type(cond ast.Expr) ast.Type {
 	match cond {
 		ast.Ident {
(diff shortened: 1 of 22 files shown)


vlib/v/checker/comptime.v
```diff
@@ -969,6 +969,15 @@ fn (mut c Checker) comptime_if_to_ifdef(name string) !string {
 	return error('bad os ifdef name "${name}"')
 }
 
+// check if `ident` is a function generic, such as `T`
+fn (mut c Checker) is_generic_ident(ident string) bool {
+	if !isnil(c.table.cur_fn) && ident in c.table.cur_fn.generic_names
+		&& c.table.cur_fn.generic_names.len == c.table.cur_concrete_types.len {
```
spytheman, line 975: Using `c.table.cur_fn.generic_names.len == c.table.cur_concrete_types.len` before the `x in y` can be a bit faster.

vlib/v/tests/comptime/comptime_match_assign_test.v
```diff
@@ -0,0 +1,62 @@
+fn test_comptime_match_assign() {
+	os := 'windows'
+	x := $match os {
+		'linux' { 'linux' }
+		'windows' { 'windows' }
+		$else { 'unknown' }
+	}
```
spytheman, line 7: what happens for `-cross` ?

spytheman: will it generate all branches?

spytheman: oh, it matches based on the literal, not the actual os

PR #25169, opened 2025-08-26
net.http_proxy : Use a url when using a proxy instead of only the path

KyleFontenot:
In the `net.http_proxy.http_do` method, have it construct a url for the request instead of just the path. It should only happen when using a proxy
```
fn (pr &HttpProxy) http_do(host urllib.URL, method Method, path string, req &Request) !Response {
	host_name, port := net.split_address(host.hostname())!

	full_url := if host.scheme == 'http' || host.scheme == 'https' {
		port_part := if port == 80 || port == 0 { '' } else { ':${port}' }
		'${host.scheme}://${host_name}${port_part}${path}'
	} else {
		path
	}

	s := req.build_request_headers(req.method, host_name, port, full_url)

```
This change fixed being able to use HTTP_PROXY and HTTPS_PROXY environment variables with the current DX. Will probably make sense to fill in more proxy handling in other modules, but this for now works with the current http_proxy API.

My reproducing cade: 
```
  proxy:       proxy
  max_retries: 2
}) or { panic('failed') }

result := testrequest.do()!
println(result.body)
```
Will try to build the test (I'm not used to testing yet) in the next commit.

### DIFF
--- a/vlib/net/http/http_proxy.v
+++ b/vlib/net/http/http_proxy.v
@@ -94,7 +94,7 @@ fn (pr &HttpProxy) http_do(host urllib.URL, method Method, path string, req &Req
 		path
 	}
 
-	s := req.build_request_headers(req.method, host_name, port, path)
+	s := req.build_request_headers(req.method, host_name, port, full_url)
 	if host.scheme == 'https' {
 		mut client := pr.ssl_dial('${host.host}:443')!
 
--- a/vlib/net/http/http_proxy.v
+++ b/vlib/net/http/http_proxy.v
@@ -87,6 +87,13 @@ fn (pr &HttpProxy) build_proxy_headers(host string) string {
 fn (pr &HttpProxy) http_do(host urllib.URL, method Method, path string, req &Request) !Response {
 	host_name, port := net.split_address(host.hostname())!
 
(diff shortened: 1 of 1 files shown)


KyleFontenot:
Wrote a basic test that certainly can be improved; would need some guidance with it.

spytheman:
Is there a way to do an end to end test, with a supported http proxy, instead of doing unit tests with mocks, that mimic the actual logic of the implementation, and are thus unnecessarily brittle to new changes?

KyleFontenot:
That is a good question, and I thought about that during tests, but assumed a real proxy request attempt in testing would be problematic. From what I know, free public proxy servers change a lot so theyre not reliable. Having anyone reproducing the test that pings some random proxy server sounds a bit iffy. So a more internal solution makes more sense.

Only thing I can think of is if maybe vlang.io had an endpoint serving as a proxy and have it protected with an authentication key of sorts. If the user doesnt have the auth key detected in a compile-time check, skip the test. Without authentication, I'd worry about scrapers. 

We could test via running a proxy server service within the test, but that would need a user's custom port open and it'd treat it as localhost which defeats a bit of the purpose. What do you think?

PR #25174, opened 2025-08-26
v: add support for `@OS`, `@CCOMPILER` `@BACKEND` and `@PLATFORM`

kbkpbot:
add `@OS`  `@COMPILER`  `@BACKEND` `@PLATFORM` comptime support.


spytheman:
I am not sure if the `@COMPILER` name is clear enough.
V has other backends too, which makes me think that `@CCOMPILER` may be a better name.

spytheman:
Another possibility is to allow `@COMPILER` to be one of `wasm`, `js`, `go`, `native` too 🤔 .

spytheman:
I've asked @medvednikov on Discord, and he said that the current names of these sound good.

Issue #25063, opened 2025-08-07
C error using `rune_bytes` as variable name


jorgeluismireles:
<details>
<summary>V version: V 0.4.11 8a7949d, press to see full `v doctor` output</summary>

|V full version      |V 0.4.11 7dc3889f19e7bdc6fb803ad4d1d65fd9d50c6e0c.8a7949d
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 22.04.5 LTS
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz
|Memory              |0.53GB/3.7GB
|                    |
|V executable        |/home/jorge/v/v
|V last modified time|2025-08-07 16:58:05
|                    |
|V home dir          |OK, value: /home/jorge/v
|VMODULES            |OK, value: /home/jorge/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/jorge/bugs
|                    |
|Git version         |git version 2.34.1
|V git status        |weekly.2025.16-576-g8a7949d9
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
|gcc version         |gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
|clang version       |Ubuntu clang version 14.0.0-1ubuntu1.1
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.5 ()
|glibc version       |ldd (Ubuntu GLIBC 2.35-0ubuntu3.10) 2.35

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  250807.v && 250807`
```v
fn main() {
	mut r := rune(0)
	rune_bytes := r.bytes()
	_ = rune_bytes
}

```

**What did you see?**
```
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/250807.01K22QJEFY9KRGE9JN1S37WR90.tmp.c:1538: error: function pointer expected
=================================================================
(You can pass `-cg`, or `-show-c-output` as well, to print all the C error messages).
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
[...]

PR #25178, opened 2025-08-27
cgen: ensure variable names do not conflict with builtin methods

dy-tea:
See #25063.

This will probably cover _most_ methods. Might be better to have a map of method names or something? Not sure how significant the performance hit of doing these checks is.


spytheman:
> Not sure how significant the performance hit of doing these checks is.

Run `./v run .github/workflows/compare_pr_to_master.v` to find out.

dy-tea:
Looks like at worst about 30% slower. Not exactly amazing. I will think about a more efficient way to do these checks.

gechandesu:
Just as an idea. Can we prefix the builtin methods to avoid conflicts instead of doing checks? It shouldn't have such a performance impact, but the side effects are unknown

dy-tea:
Technically you could take the approach of prefixing built-in methods with something and output an error if the user tries naming a variable with that prefix. 
That would prevent name collisions entirely with little performance hit.

dy-tea:
New impl is at worst around 4% slower than master, at lot better than before.

Delta456:
You would need to cover a lot more cases than you think 🥲

dy-tea:
Indeed there are a lot more cases of this than I expected (let me know if I am missing any obvious ones). Notably I added a lot more string, array and map methods. The second check I added doesn't appear to significantly affect performance over the one I had previously but there might be a better way to combine the checks?
Will look at it more tomorrow if I have time.

dy-tea:
My newest implementation seems to have basically no performance impact. I'm covering a lot more cases (although likely still missing a few).

dy-tea:
Noticed that I broke autofree so I added handling for that too. Let me know if I'm missing testcases, otherwise the functionality is there.

spytheman:
The code from the issue #25063 , still leads to a cgen error 🤔.
```
#1 05:40:19 ^ master ~/v>git switch -
Switched to branch 'pr_25178'
#0 05:40:23 ^ pr_25178 ~/v>./v self
V self compiling ...
#0 05:40:27 ^ pr_25178 ~/v>./v vlib/v/tests/clash_var_fn_name_test.v
#0 05:40:37 ^ pr_25178 ~/v>cat /home/delian/code/misc/2025_09_03__05/xx.v
fn main() {
        mut r := rune(0)
        rune_bytes := r.bytes()
        _ = rune_bytes
}

#0 05:40:44 ^ pr_25178 ~/v>./v /home/delian/code/misc/2025_09_03__05/xx.v
cc: /tmp/v_1000/xx.01K46Q5KDGB27TBYVVGWBQ5RKA.tmp.c:1544: error: '{' expected (got ";")
#1 05:40:51 ^ pr_25178 ~/v>./v -cc clang-18 /home/delian/code/misc/2025_09_03__05/xx.v
================== C compilation error (from clang-18): ==============
cc: /tmp/v_1000/xx.01K46Q5V6Z5PE53QK72MZC7690.tmp.c:1965:12: error: initializing 'Array_u8' (aka 'struct array') with an expression of incompatible type 'Array_u8 (rune)' (aka 'struct array (unsigned int)')
cc:  1965 |         {Array_u8 _ = rune_bytes;}
cc:       |                   ^   ~~~~~~~~~~
cc: 1 error generated.
#1 05:40:59 ^ pr_25178 ~/v>
```

V generated:
```c
VV_LOC void main__main(void) {
    rune r = ((rune)(0));
    Array_u8 _var_rune_bytes = rune_bytes(r);
    {Array_u8 _ = rune_bytes;}
    ;
}
```

spytheman:
On master, the generated code was:
```c
VV_LOC void main__main(void) {
    rune r = ((rune)(0));
    Array_u8 rune_bytes = rune_bytes(r);
    {Array_u8 _ = rune_bytes;}
    ;
}
```

dy-tea:
Can't believe I missed that 🥲. I will add some extra tests.

Edit: yes the fix was quite simple, I'm also adding a test for assignment.

dy-tea:
CI seems to be timing out on installing dependencies for some checks. Should I rerun?

spytheman:
> CI seems to be timing out on installing dependencies for some checks. Should I rerun?

Yes. Github Actions runners are flaky in that way unfortunately :-| .

spytheman (approved):
Excellent work.

spytheman:
Sorry, but after merging it, the performance is much worse:
image

I've reverted the change for now.

Issue #25185, opened 2025-08-29
parser should not allow same symbol names import from different module


kbkpbot:
### Describe the bug

parser should not allow same symbol names import from different module

### Reproduction Steps

a/a.v:
```v
module a

pub struct Top {
        a int
}

```

b/b.v:
```v
module b

pub struct Top {
        b int
}
```

fmt.v
```v
module main

import a { Top }
import b { Top }

fn main() {
        x := Top{}
        dump(x)
}
```

### Expected Behavior

parser should issue an error msg, that `Top` already registered by another module.

### Current Behavior

```
$ v fmt.v
fmt.v:3:8: warning: module 'a' is imported but never used
    1 | module main
    2 |
    3 | import a { Top }
      |        ^
    4 | import b { Top }
    5 |
```

### Possible Solution

Fix parser

### V version

V 0.4.11 9e1273ae713331889407246bfe53320943a85af9

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 9e1273ae713331889407246bfe53320943a85af9
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |10.7GB/15.51GB
|                    |
|V executable        |/media/HD/github/lang/v/v
|V last modified time|2025-08-29 03:37:07
|                    |
|V home dir          |OK, value: /media/HD/github/lang/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/comptime/top
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.35-9-g9e1273ae
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.5) 2.39

> [!NOTE]
[...]

PR #25187, opened 2025-08-29
parser: add duplicate import symbol detect

kbkpbot:
Fix issue #25185

remove `vlib/v/fmt/tests/import_duplicate_input.vv` as it is treated as a parser error now.


spytheman (approved):
Excellent work.

Issue #25166, opened 2025-08-25
`flag.to_struct`   parser handles empty (tail?) bool args wrong


larpon:
### Describe the bug

`flag.to_struct`   parser handles empty (tail?) bool args wrong. See below.

### Reproduction Steps

```v
module main

import flag

struct CliOptions {
	show_help bool @[long: 'help'; short: h]
}

fn main() {
	cli_options, unmatched := flag.to_struct[CliOptions](arguments(),
		skip:  1
		style: .v
		mode:  .relaxed
	)!

	if unmatched.len > 0 {
		eprintln('Unmatched arguments: ${unmatched}')
		eprintln(flag.to_doc[CliOptions]()!)
		exit(1)
	}
	if cli_options.show_help {
		eprintln(flag.to_doc[CliOptions]()!)
		exit(0)
	}
}
```

### Expected Behavior

Output from `v run t.v -h` should be:
```

Options:
  -h, --help
```

### Current Behavior

```
v run t.v -h
V panic: result not set (can not assign `` to bool field `show_help`)
v hash: 3b25f1e
/dev/shm/v_1000/t.01K3GTKF1XF60JNWW1N1YY2FXB.tmp.c:4421: at _v_panic: Backtrace
/dev/shm/v_1000/t.01K3GTKF1XF60JNWW1N1YY2FXB.tmp.c:4385: by panic_result_not_set
/dev/shm/v_1000/t.01K3GTKF1XF60JNWW1N1YY2FXB.tmp.c:9363: by main__main
/dev/shm/v_1000/t.01K3GTKF1XF60JNWW1N1YY2FXB.tmp.c:9416: by main
```

### Possible Solution

I think changing line 935 in `flag_to.v` to:
```v
				} $else $if field.typ is bool {
					if arg := f.arg {
						if arg != '' {
							return error('can not assign `${arg}` to bool field `${field.name}`')
						}
					}
					result.$(field.name) = !the_default.$(field.name)
				} $else $if field.typ is string {
````
fixes it but requires some testing...

### V version

V 0.4.11 3b25f1e

### Environment details (OS name and version, etc.)

EndeavourOS (Arch-based)
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #25189, opened 2025-08-29
flag: better covering fix for #25166 tail/single bool flag error in `flag_to.v`

larpon:
Turns out the previous fix did not cover cases where there were multiple flags involved. This improved fix should cover cases where the bool flag is not only the tail/last in the argument list. New tests included for the new case

### DIFF
--- a/vlib/flag/flag_to_tail_bool_test.v
+++ b/vlib/flag/flag_to_tail_bool_test.v
@@ -1,14 +1,22 @@
 import flag
 
-const args_bool_short = ['some.exe', '-h']
-const args_bool_long = ['some.exe', '-help']
+const args_bool_short_tail = ['some.exe', '-h']
+const args_bool_long_tail = ['some.exe', '-help']
+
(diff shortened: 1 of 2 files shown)


vlib/flag/flag_to_tail_bool_test.v
```diff
@@ -1,14 +1,56 @@
 import flag
 
-const args_bool_short = ['some.exe', '-h']
-const args_bool_long = ['some.exe', '-help']
+const args_bool_short_tail = ['some.exe', '-h']
+const args_bool_long_tail = ['some.exe', '-help']
+
+const args_bool_short_mixed = ['some.exe', '-h', '-long', 'val']
+const args_bool_long_mixed = ['some.exe', '-help', '-long', 'val']
```
spytheman, line 7: why define consts, that will be used in just one place (per each const)?

The tests are already named appropriately, imho there is no need for the indirection.

larpon: You're right. I've removed them in this PR: #25192

Issue #25115, opened 2025-08-15
json2 can't encode runes of more than two bytes


jorgeluismireles:
<details>
<summary>V version: V 0.4.11 bbb61ab, press to see full `v doctor` output</summary>

|V full version      |V 0.4.11 7dc3889f19e7bdc6fb803ad4d1d65fd9d50c6e0c.bbb61ab
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 22.04.5 LTS
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz
|Memory              |0.48GB/3.7GB
|                    |
|V executable        |/home/jorge/v/v
|V last modified time|2025-08-14 17:07:08
|                    |
|V home dir          |OK, value: /home/jorge/v
|VMODULES            |OK, value: /home/jorge/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/jorge/bugs
|                    |
|Git version         |git version 2.34.1
|V git status        |weekly.2025.16-626-gbbb61ab3
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
|gcc version         |gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
|clang version       |Ubuntu clang version 14.0.0-1ubuntu1.1
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.5 ()
|glibc version       |ldd (Ubuntu GLIBC 2.35-0ubuntu3.10) 2.35

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  json_emoji/json_emoji.v && json_emoji/json_emoji`
```v
module main

import x.json2

struct Struct {
	a []string
}

fn main() {
	s := Struct{
		a: ['\0', '\t', 'a', 'ñ', '♥', '𝄞']
	}
	input := r'{"a":["\u0000","\t","a","\u00f1","\u2665","\ud834\udd1e"]}'

	// custom string encode
	mut bytes := []u8{}
	bytes << '{"a":['.bytes()
	for j in 0 .. s.a.len {
		if j > 0 {
[...]

PR #25193, opened 2025-08-29
decoder2: Add support for surrogates

Larsimusrex:
Will now decode utf-16 surrogates, used by some encoders for characters outside the bilingual plane.

`println(decoder2.decode[string](r'"\ud83d\ude00"')!) // '😀'`

### DIFF
--- a/vlib/x/json2/decoder2/tests/decode_escaped_string_test.v
+++ b/vlib/x/json2/decoder2/tests/decode_escaped_string_test.v
@@ -10,3 +10,50 @@ fn test_decode_escaped_string() {
 
 	assert escaped_strings == decoded_strings
 }
+
+fn test_surrogate() {
+	assert decoder2.decode[string](r'"\ud83d\ude00"')! == '😀'
+	assert decoder2.decode[string](r'"\ud83d\ude00 text"')! == '😀 text'
+}
+
+fn test_invalid_surrogate() {
+	if x := decoder2.decode[string](r'"\ud83d"') {
+		assert false
+	} else {
+		if err is decoder2.JsonDecodeError {
+			assert err.line == 1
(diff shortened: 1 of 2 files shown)


vlib/x/json2/decoder2/tests/decode_escaped_string_test.v
```diff
@@ -10,3 +10,50 @@ fn test_decode_escaped_string() {
 
 	assert escaped_strings == decoded_strings
 }
+
+fn test_surrogate() {
+	assert decoder2.decode[string](r'"\ud83d\ude00"')! == '😀'
+	assert decoder2.decode[string](r'"\ud83d\ude00 text"')! == '😀 text'
```
spytheman, line 16: What will be the result, if you later JSON encode the decoded string?
Can you please add a test for that too?

spytheman:
Excellent work @Larsimusrex.

I am curious, what are some of the JSON encoders, that produce such output?
If they are easy to install on the CI, we can add round trip tests with them.

Larsimusrex:
Python definitely does it by default. I think java and c# too.

vlib/x/json2/decoder2/tests/decode_escaped_string_test.v
Larsimusrex: the json2 encoder currently handles these characters incorrectly see #25115. In my new implementation it outputs utf-8 by default unless specified otherwise

spytheman:
Python is already preinstalled on the CI, so we can add a V test, that invokes a python program, that generates a json encoded value with surrogates on stdout, then decodes it and asserts on the output.

We also support `// vtest build: present_python?` so that the V test could be skipped on environments that do not have Python.

PR #25195, opened 2025-08-29
veb: fix a spurious error with Chromium, for POST requests, split into 2 parts, 1st with only http headers, without body (fix #25191)

### LINKED ISSUE #25191: veb fails to receive raw binary upload
### Describe the bug

veb fails to parse binary requests with `[veb] error parsing request: unsupported version`

Because of this issue plus #24975 veb is unable to receive file uploads.

This stops my work. It is a serious issue.

### Reproduction Steps

See [here](https://github.com/einar-hjortdal/vlang-issue-25191)

### Expected Behavior

veb receives binary payload.

### Current Behavior

veb fails.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.11 7831fb0

```
|V full version      |V 0.4.11 214628a939d744768e5bba9b49699ea81a8d3f83.7831fb0
|OS                  |linux, "openmamba release 2025.8.1 for x86_64 (rolling)"
|Memory              |0.72GB/31.16GB
|                    |
|V last modified time|2025-08-29 16:36:27
|                    |
|Current working dir |OK, value: /home/einar/Documents/projects/vlang/bin_payload
|                    |
|V git status        |weekly.2025.28-286-g7831fb0b
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20250818
|gcc version         |gcc (GCC) 15.2.1 20250818
|clang version       |N/A
|emcc version        |N/A
```

> [!NOTE]
>

spytheman:
- **net: improve tracing output for `-d trace_tcp`; move `error_code()` calls right after the corresponding C APIs (errno is a global that should be read immediately to be valid)**
- **veb: fix a spurious error with Chromium based browsers, when an http POST request is split into several parts, where the first one has only http headers, without a body (fix #25191)**

### DIFF
--- a/vlib/veb/tests/handle_split_requests_in_chrome_test.v
+++ b/vlib/veb/tests/handle_split_requests_in_chrome_test.v
@@ -1,0 +1,95 @@
+import log
+import veb
+import time
+import net
+import x.json2
+import net.http
+
+const port = 31228
+
+pub struct Context {
+	veb.Context
+}
+
+pub struct App {
+pub mut:
+	started chan bool
+}
+
+pub fn (mut app App) before_accept_loop() {
+	app.started <- true
+}
+
+@['/data/:filename'; post]
+fn (mut app App) data(mut ctx Context, filename string) veb.Result {
+	content_type := ctx.get_header(http.CommonHeader.content_type) or { return ctx.no_content() }
+	f := http.FileData{
+		filename:     filename
+		content_type: content_type
+		data:         ctx.req.data
+	}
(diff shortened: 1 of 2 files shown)


spytheman:
The problem was manifested in situations, where the request to the veb server was done in 2 parts, the first containing *only* the http request start line + headers, and the other parts, containing a pure body.

Chromium based browsers tend to produce such requests on keep-alive connections, when the POST requests follow quickly one after the other (quicker than ~200ms for me). Firefox does not split the requests in that way - it tries to send some of the body with the headers instead, which did not trigger the veb bug.

Issue #25204, opened 2025-08-30
v checker panic if incorrect number of parameters supplied to method


dy-tea:
### Describe the bug

Checker panics when running included code snippet. Found while writing test cases for #25178. Haven't looked too deeply but it's probably somewhat related to that issue.

### Reproduction Steps

```v
fn main() {
	mut m := map[string]int
	m['test'] = 10
	map_delete := m.delete()
}
```

### Expected Behavior

Map delete returns void so maybe we need an error here? 

### Current Behavior

```
V panic: array.get: index out of range (i,a.len):0, 0
v hash: cb2756e
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:19677: at _v_panic: Backtrace
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:19725: by panic_n2
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:19231: by array_get
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:644: by v__checker__Checker_map_builtin_method_call
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:65026: by v__checker__Checker_method_call
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:63608: by v__checker__Checker_call_expr
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:54528: by v__checker__Checker_expr
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:49130: by v__checker__Checker_assign_stmt
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:53661: by v__checker__Checker_stmt
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:54285: by v__checker__Checker_stmts_ending_with_expression
/tmp/v_1000/v2.01K3Y9KY3GYMXJ5EQMTV5QMH44.tmp.c:54262: by v__checker__Checker_stmts
[...]

PR #25205, opened 2025-08-30
checker: add more checks for map.delete 

Delta456:
Fixes https://github.com/vlang/v/issues/25204


spytheman:
Excellent work.
Thank you @Delta456 🙇🏻.

PR #25210, opened 2025-08-31
markused: cleanup c code generated

felipensp:
cleanup


vlib/v/ast/table.v
```diff
@@ -38,6 +38,10 @@ pub mut:
 	// json             bool            // json is imported
 	comptime_calls map[string]bool // resolved name to call on comptime
 	comptime_syms  map[Type]bool   // resolved syms (generic)
+	//
+	uses_attr_noreturn bool // @[noreturn]
+	uses_attr_hidden   bool // @[hidden]
+	uses_attr_weak     bool // @[weak]
```
spytheman: why `uses_`, when all the others were `used_` ?

vlib/v/gen/c/cheaders.v
```diff
@@ -451,9 +453,6 @@ void v_free(voidptr ptr);
 	#endif
 #endif
 
-// g_live_info is used by live.info()
-static void* g_live_info = NULL;
-
```
spytheman, line 456: Oops, that seems indeed unused anymore. Good find.

vlib/v/ast/table.v
felipensp: Fixed.

spytheman (approved):
Excellent work.

Issue #25211, opened 2025-08-31
$for fails for values of aliases to enums


Larsimusrex:
### Describe the bug

iterating over the values of a alias to an enum results in a c error.

### Reproduction Steps

```v
enum Stuff {
	a
	b
}

type StuffAlias = Stuff

fn main() {
	$for value in StuffAlias.values {
		
	}	
}
```

### Expected Behavior

nothing

### Current Behavior

```
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/test.01K40SVD1PVYZ4T5EG71Y1PEVY.tmp.c:1519: error: 'main__StuffAlias__a' undeclared
=================================================================
(You can pass `-cg`, or `-show-c-output` as well, to print all the C error messages).
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```

### V version

V 0.4.11

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 7831fb0b82263ce0f2ac6795c5e7b12bd5ea3835.d31aaec
|:-------------------|:-------------------
|OS                  |linux, "Void Linux"
|Processor           |12 cpus, 64bit, little endian, AMD Ryzen 5 5600G with Radeon Graphics
|Memory              |4.07GB/15.52GB
|                    |
|V executable        |/home/lars/v/v
|V last modified time|2025-08-31 09:43:30
|                    |
|V home dir          |OK, value: /home/lars/v
|VMODULES            |OK, value: /home/lars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/lars
|                    |
|Git version         |git version 2.50.1
|V git status        |weekly.2025.06-1111-gd31aaecc
|.git/config present |true
|                    |
|cc version          |cc (GCC) 14.2.1 20250405
|gcc version         |gcc (GCC) 14.2.1 20250405
|clang version       |clang version 17.0.6
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.3 (a9651ff57165f5710bb09a5fe52590fd6ddb72df)
[...]

PR #25212, opened 2025-08-31
cgen: fix alias enum comptime for 

Delta456:
Fixes https://github.com/vlang/v/issues/25211#issue-3370612553


vlib/v/tests/comptime/comptime_enum_values_test.v
```diff
@@ -33,3 +35,14 @@ fn test_main() {
...
+fn test_alias_enum() {
+	mut values := []AnotherCharGroup{}
+	for entry in AnotherCharGroup.values() {
+		println('Value: ${entry}  ${entry.value()}')
+		values << entry
+	}
+
+	assert values == [AnotherCharGroup.chars, AnotherCharGroup.alphanumerics,
+		AnotherCharGroup.numeric, AnotherCharGroup.special]
+}
```
spytheman: the test passes on master too:

Image

spytheman: Given the fix in cgen, imho the new test function should be using a `$for`, not `for`.

Delta456: should be done now

Issue #23836, opened 2025-03-01
allow $if work everywhere


kbkpbot:
### Describe the feature

$if conditional compile can work only on some code regions. 

### Use Case

some use case :

```v
module main

$if x64 { 
	pub type Digits = u64
} 
$else {
	pub	type Digits = u32
}


fn main() {
	y := []Digits{}
	dump(y)
}

```

### Proposed Solution

Generate `#ifdef` at `$if`.

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.9 ab2eb00.2cd1f6e

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.9 ab2eb00.2cd1f6e
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.2 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |11.79GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-03-01 11:56:45
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /media/HD/github/kbkpbot/v
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.09-24-g2cd1f6e9
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2024-07-31 HEAD@1cee0908 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 0134e9b9-dirty
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.4) 2.39

larpon:
Hear. We need top level support for comptime logic, also for `import` statements

JalonSolov:
... and `const`s.  :-)

medvednikov:
Yeah I'd like that very much

enghitalo:
[...]

PR #25216, opened 2025-09-01
parser,fmt,markused: add top level comptime if support

kbkpbot:
This PR enable top level `$if` in code.
Partially support the requirements of Issue #23836

It is done by totally skip the code in a false branch in `parser`.

TODO: support `$if` in between struct declare fields, maybe in another PR.


JalonSolov:
Why not?  Just as useful as `$if`.  In fact, one could likely make a case for `$for`, as well.

vlib/v/fmt/fmt.v
```diff
@@ -2354,6 +2361,7 @@ pub fn (mut f Fmt) if_expr(node ast.IfExpr) {
 	start_len := f.line_len
 	for {
 		for i, branch in node.branches {
+			f.branch_processed_imports.clear()
```
spytheman, line 2364: Nice, constructs like this one are well formatted:
```v
$if linux {
    import common_linux as common
    $if amd64 {
        import abc_amd64 as imp
    }
} $else $if windows {
    import common_windows as common
    $if amd64 {
        import windows_abc_amd64 as imp
    }
}
common.f()
imp.g()
```

vlib/v/parser/if_match.v
```diff
@@ -570,3 +605,295 @@ fn (mut p Parser) select_expr() ast.SelectExpr {
 		has_exception: has_else || has_timeout
 	}
 }
+
+fn (mut p Parser) comptime_if_cond(mut cond ast.Expr) bool {
```
spytheman, line 608: The code here seems like a simplified duplicate of the code in 
`fn (mut c Checker) comptime_if_cond(mut cond ast.Expr, mut sb strings.Builder) (bool, bool) {`

Imho that should be extracted in a common module, imported and reused in both,
otherwise in the future, we will have subtle bugs, when the two places diverge more as they evolve.

vlib/v/parser/parser.v
```diff
@@ -677,25 +696,52 @@ fn (mut p Parser) top_stmt() ast.Stmt {
...
+							expr: expr
+							pos:  pos
+						}
+					}
+					.name {
+						// handles $dbg directly without registering token
+						if p.peek_tok.lit == 'dbg' {
+							return p.dbg_stmt()
+						} else {
+							mut pos := p.tok.pos()
```
spytheman, line 733: I do not understand, why does `$dbg` need special handling here?

vlib/v/parser/parser.v
```diff
@@ -111,6 +111,7 @@ mut:
 	generic_type_level       int  // to avoid infinite recursion segfaults due to compiler bugs in ensure_type_exists
 	main_already_defined     bool // TODO move to checker
 	is_vls                   bool
+	is_begining_of_file      bool
```
spytheman:
```suggestion
	is_begining_of_file      bool = true
```

spytheman: hm, no ... that would not work ... the actual meaning of the field is `start of statements after module`, it is not literal beginning of a file

spytheman: perhaps `inside_import_section` would be a clearer name, since it starts as false, and becomes true after the module statement, and remains true, until the last import/top level $if had been processed

vlib/v/parser/parser.v
```diff
@@ -451,6 +453,12 @@ fn (mut p Parser) parse_block() []ast.Stmt {
 	return stmts
 }
 
+fn (mut p Parser) is_in_top_level_comptime(inside_assign_rhs bool) bool {
+	// TODO: find out a better way detect we are in top level.
+	return p.cur_fn_name == '' && p.inside_ct_if_expr && !inside_assign_rhs && !p.script_mode
+		&& p.tok.kind != .name
```
spytheman: would not `p.opened_scopes == 0` be suitable ?

spytheman: if you do need to keep the string check, please put it at the end of the && chain, so that the cheaper checks can shortcircuit it

vlib/v/tests/comptime/comptime_if_top_test.v
```diff
@@ -0,0 +1,97 @@
...
+		enum1_a
+		enum1_b
+	}
+
+} $else $if new_c_2 ? {
+	pub enum Enum1 {
+		enum1_c
+		enum1_d
+	}
+
```
spytheman: why is there an empty line here after the enum declaration?

spytheman: the other top level declarations do not have empty lines 🤔

vlib/v/tests/comptime/comptime_if_top_test.v
spytheman, line 1: please also duplicate this file, but with a vflags line, that ensures that the *second* choice is taken, not the first. That will test the logic evaluation more thoroughly.

spytheman:
> Do we need a $match at top level?

Please add support for it in another PR, to keep this one more focused.

enghitalo:
I think it would be good to include some description about this new feature in the README.md

JalonSolov:
The only thing "new" about it is that now it can be used at the top level, where it couldn't be used before.  More of a bug fix than a new feature.

However, it wouldn't hurt to have an example about imports, or some such added to what's already in the docs.

vlib/v/parser/if_match.v
kbkpbot: Yes, placing it in a common module would be a better approach. However, as seen in `comptime_if_cond()`, it relies on constructs like `c.expr()`, which complicates the extraction process.

vlib/v/parser/parser.v
kbkpbot: I am not sure what is `$dbg` indeed. I did a search and found it seems not being use by other code.
```
./fmt/fmt.v:    f.writeln('\$dbg;')
./parser/parser.v:                                              // handles $dbg directly without registering token
./parser/parser.v:                                      // handles $dbg directly without registering token
./gen/c/cgen.v: has_debugger         bool   // $dbg has been used in the code

```

```
./gen/c/cgen.v: has_debugger         bool   // $dbg has been used in the code
./gen/c/cgen.v:         has_debugger:         'v.debug' in table.modules
./gen/c/cgen.v:         has_debugger:          'v.debug' in global_g.table.modules

```

vlib/v/parser/parser.v
kbkpbot: ok

vlib/v/parser/parser.v
kbkpbot: I will run some tests on this.

vlib/v/tests/comptime/comptime_if_top_test.v
kbkpbot: It seems to be a `vfmt` bug here..... :)

vlib/v/tests/comptime/comptime_if_top_test.v
kbkpbot: ok

vlib/v/parser/parser.v
JalonSolov: It isn't in any of the regular V code, as it halts the program for debugging.  It is there for Users to be able to debug their code without needing an external debugger such as gdb, etc.

And yes, I can see it being used at the top level, especially if you wanted to ensure the top level things were being set correctly before the main code started running.

vlib/v/tests/comptime/comptime_if_top_test.v
kbkpbot: fix

vlib/v/parser/parser.v
kbkpbot: fixed

vlib/v/parser/parser.v
kbkpbot: fixed

vlib/v/parser/if_match.v
kbkpbot: Unfortunately, it seems that the `comptime_if_cond()` in the `checker` cannot be merged with the `comptime_if_cond()` in the `parser`. This is because the `checker` supports more operators, such as `is`, and requires additional type checking, while the `parser` is relatively simpler.

```v
$if T is u32 {
```
which is not possible in top level.

vlib/v/parser/parser.v
kbkpbot: Should we remove this code?

spytheman: > And yes, I can see it being used at the top level, especially if you wanted to ensure the top level things were being set correctly before the main code started running.

I do not, since it is entirely a runtime thing. You can not meaningfully put a `$dbg` statement between for example 2 struct declarations.

It injects a small repl call between others, in running functions, and can only appear at the top level because of script mode. I was puzzled about why does it need special treatment here, because imho it should be handled just like all the other fn calls, assignments etc, that also can be top level statements, just because of script mode.

spytheman:
I think this is excellent work, and will enable a lot of future cleanups. With it, we can get rid of (or at least reduce the use of) less flexible implementation hacks/features, like the platform suffix files.

I am on the fence, whether it is a good idea to document it right away, since it may still have some rough edges, but it probably is - we will gather more feedback from more people faster, if it is documented.

spytheman:
Given the CI load though, the documentation may be faster to be done in a separate PR, after this one is merged, that will only touch doc/docs.md, so that iterations on it, would not trigger all the sanitized jobs etc.

larpon:
@kbkpbot, on a non-technical note; This PR is awesome! Thanks for your fantastic work on V and for making this happen 🔥🔥🔥

jorgeluismireles:
I tested 
```v
$if is_coroutine ? {
	import coroutines
}
```
is very helpful since no trace of photon remains in the C code when not using `-use-coroutines`

Issue #25219, opened 2025-09-02
V REPL after input Chinese character there is a misalignment in cursor positioning


kbkpbot:
### Describe the bug

V's REPL, after inputting Chinese characters, when moving the cursor, there is a misalignment in cursor positioning.

Linux Only, Windows has no this bug.

### Reproduction Steps

Start V REPL, input some Chinese characters, moving the cursor left or right.

### Expected Behavior

Correct cursor positioning

### Current Behavior

Can't move cursor and stuck at middle.

### V version

V 0.4.11 19d31f221fb72987c8e25b6889d746da2da57630

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 19d31f221fb72987c8e25b6889d746da2da57630
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |6.78GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-09-02 11:01:38
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /media/HD/github/kbkpbot/v
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-659-g19d31f22
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.5) 2.39

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #25220, opened 2025-09-02
readline: fix wide char display width

kbkpbot:
Fix issue #25219

Use `east_asian.display_width()` calculate the display width of a string.


spytheman:
Excellent work, thank you 🙇🏻‍♂️.
I just tested with `a := 'a你好wesome'` , and it works well.

PR #25222, opened 2025-09-02
parser,checker: add top level comptime match support

kbkpbot:
This PR enable top level `$match`.

Add @token support in `$if` and `$match`.
So we can write code such as :
```v
$match $OS {
    'linux' {
        import mymodule_linux
    }
    'windows' {
        import mymodule_windows
    }
    'macos' {
        import mymodule_macos
    }
    $else {
        import mymodule_default
    }
}
```

Top level comptime `$if` or `$match` only support @token `@MOD` `@OS` `@CCOMPILER` `@BACKEND` or `@PLATFORM` right now.

I am not sure other @token is useful or not at top level: `@VROOT`, `@VMODROOT`, `@VEXEROOT`, `@FN`, `@METHOD`, `@STRUCT`,
	`@VEXE`, `@FILE`, `@DIR`, `@LINE`, `@COLUMN`, `@VHASH`, `@VCURRENTHASH`, `@VMOD_FILE`,
	`@VMODHASH`, `@FILE_LINE`, `@LOCATION`, `@BUILD_DATE`, `@BUILD_TIME`, `@BUILD_TIMESTAMP`

### DIFF
--- a/vlib/v/checker/comptime.v
+++ b/vlib/v/checker/comptime.v
@@ -1018,6 +1018,30 @@ fn (mut c Checker) comptime_if_cond(mut cond ast.Expr, mut sb strings.Builder) (
 				}
 				.eq, .ne, .gt, .lt, .ge, .le {
 					match mut cond.left {
+						ast.AtExpr {
+							// @OS == 'linux'
+							left_type := c.expr(mut cond.left)
+							right_type := c.expr(mut cond.right)
+							if !c.check_types(right_type, left_type) {
+								left_name := c.table.type_to_str(left_type)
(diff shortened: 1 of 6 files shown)


spytheman:
> Top level comptime `$if` or `$match` only support `@MOD`, `@OS`, `@CCOMPILER`, `@BACKEND` or `@PLATFORM` right now.

I think that is fine. The rest have values, that can not be compared easily against known ones (except perhaps the build ones, but I am not sure, if it is a good idea to do compile time logic based on them anyway), without also making builds a lot more brittle.

Issue #25223, opened 2025-09-02
$compile_error() always triggers in compile time $match


RubyCascade:
### Describe the bug

When using `$compile_error()` in a compile time `$match` statement `$compile_error()` always triggers even when it should not.

### Reproduction Steps

```v
module main

fn test_1[T]() bool {
	$match T {
		u8, u16 { return true }
		$else {
			// return false
			$compile_error('fail')
		}
	}
}

fn test_2[T]() bool {
	$match T {
		i32 { return true }
		i64 { return true }
		$else {
			// return false
			$compile_error('fail')
		}
	}
}

fn main() {
	println(test_1[u8]())  // should pass ('true')
	println(test_1[u16]()) // should pass ('true')
	// println(test_1[u32]()) // should fail ('false')

	// println(test_2[u8]())  // should fail ('false')
	println(test_2[i32]()) // should pass ('true')
	println(test_2[i64]()) // should pass ('true')
}
```

### Expected Behavior

The program to successfully compile and run.

### Current Behavior

`$compile_error()` triggers even though it should not.

### Possible Solution

There was a similar issue with `$compile_error()` and compile time `$if` statements. The solution for this may or may not be similar to the one for that.

### V version

V 0.4.11 f6b60e4

### Environment details (OS name and version, etc.)

macOS Sequoia 15.6

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #25225, opened 2025-09-03
checker: comptime match only eval true branch

kbkpbot:
Fix issue #25223


vlib/v/tests/comptime/comptime_match_eval_only_true_branch_test.v
spytheman, line 1: I think that there should be a test .vv/.out pair in `vlib/v/checker/tests/`, similar to this test, but that forces the branch for $compile_error too, and thus ensures that the error message is shown for that situation.

spytheman (approved):
Excellent work.

PR #25226, opened 2025-09-03
net.http: Open up fetch and http verb function parameters to accept Request struct.

KyleFontenot:
The objective is to address proxy settings for http requests. Currently, when using a HttpProxy, one would have to do something like this:
```
  proxy:       proxy
}) or { panic('failed') }
result := testrequest.do()!
```
It requires the `http_do` method on an HttpProxy, so its handled very differently from the rest of the http verb functions because they aren't used. `proxy` is already a key of Request that successfully does do proxy-specific behavior, so if we simply allowed a `HttpProxy` struct into the `proxy` key to be passed in with the fetch/http verb function parameters, I only see benefits.  It streamlines `fetch()` , though it changes a few things downstream within the same `net` stdlib module. 

Instead, if we had the fetch function parameters as `fn fetch(url string, req ?Request)`, we can make the API simpler for using proxies, but also just allow any other Request config keys to be configured with each function. Currently the second parameter is just the `.data` field of Request. So we would be able to do:
```
new_proxy := http.new_http_proxy(os.environ()['HTTP_PROXY']) or { panic('failed to create proxy') }
get("http://something.com", Request{ proxy:  new_proxy })
```

I noticed since the `FetchConfig` struct was nearly identical to a `Request`, I merged and consolidated them. 
`FetchConfig` has the `params` key which is more akin to what the more url-specific `urllib.URL` struct should hold. What FetchConfig didn't have that Request does is `read_timeout` and `write_timeout`, nothing detrimental or warranting a separate struct for a different DX. Because of this and because all functions are already doing `urllib.parse()` in the `prepare()` function, I don't see a problem with changing Request.url to the more descriptive `urllib.URL` that already exists instead of a string because it already has a `str()` method and fetch requests already parse urls. So I moved what `build_url_from_fetch()` did to inside the `urllib.parse()` function using the existing `urllib.parse_query()` method. 

Because of that, some functions became useless like `build_url_from_fetch()` and `new_request()` since there wasn't anything useful that other functions weren't already doing.  

I plan on making a test for it when I figure out a good way of testing actual proxy info similar to: https://github.com/vlang/v/pull/25169


KyleFontenot:
I will separate objective scopes into two different PRs. Became too messy

Issue #25229, opened 2025-09-03
v repl execute a command with line comment, will output nothing


kbkpbot:
### Describe the bug

v repl execute a command with line comment, will output nothing

### Reproduction Steps

```
mars@mars-QiTianM428-A688:~/github/kbkpbot/v$ v
 ____    ____
 \   \  /   /  |  Welcome to the V REPL (for help with V itself, type  exit , then run  v help ).
  \   \/   /   |  Note: the REPL is highly experimental. For best V experience, use a text editor,
   \      /    |  save your code in a  main.v  file and execute:  v run main.v
    \    /     |  V 0.4.11 f6b60e4 . Use  list  to see the accumulated program so far.
     \__/      |  Use Ctrl-C or  exit  to exit, or  help  to see other available commands.

>>> time.now()
2025-09-03 21:26:40
>>> time.now() // haha, this will output nothing
>>>

```

### Expected Behavior

Function call ending with line comment, should output function call result.

### Current Behavior

Output nothing.

### Possible Solution

fix repl

### V version

V 0.4.11 f6b60e4d9f4213f5fa08651da7af7b4ef2806ce1

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 f6b60e4d9f4213f5fa08651da7af7b4ef2806ce1
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |5.35GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-09-03 13:26:33
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /media/HD/github/kbkpbot/v
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-662-gf6b60e4d-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
[...]

PR #25230, opened 2025-09-03
repl: remove line comment

kbkpbot:
Fix issue #25229

### DIFF
--- a/cmd/tools/vrepl.v
+++ b/cmd/tools/vrepl.v
@@ -109,6 +109,40 @@ fn endline_if_missed(line string) string {
 	return line + '\n'
 }
 
+fn remove_comment(s string) string {
+	mut result := []rune{}
+	mut in_double_quote := false
+	mut in_single_quote := false
+	mut escaped := false
+
+	for i := 0; i < s.len; {
+		c := s[i]
+
+		if c == `\\` && (in_double_quote || in_single_quote) {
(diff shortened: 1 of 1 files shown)


cmd/tools/vrepl.v
```diff
@@ -441,7 +476,7 @@ fn run_repl(workdir string, vrepl_prefix string) int {
 			prompt = '... '
 		}
 		oline := r.get_one_line(prompt) or { break }
-		line := oline.trim_space()
+		line := remove_comment(oline).trim_space()
```
spytheman, line 479: Would not just `line := oline.all_before('//').trim_space()` work?

spytheman: image

It seems to solve the immediate issue.

It would not track // in quotes, but that can get really complex if it has to be done reliably since V also allows for mixing nested interpolations with different quotes and // in them ...

spytheman:
`v vlib/v/slow_tests/repl/repl_test.v` also fails with the incomplete parser solution, while it is fine with the .all_before('//') one:
image

spytheman:
(on a side note, the CI should be changed so that the REPL tests are run when vrepl.v is changed, currently they are skipped, so you would have to run them manually locally)

spytheman:
I added the simpler solution + test in 8868696 for now.

PR #25231, opened 2025-09-03
fix: gen_struct_equality_fn null pointer crash

get200:


get200:
```v
// D:\\v\\v\\vlib\\json\\tests\\json_option_struct_test.v
import json

pub struct SomeStruct {
pub mut:
	test ?string
}

pub struct MyStruct {
pub mut:
	result ?SomeStruct
	id     string
}

fn test_main() {
	a := MyStruct{
		id:     'some id'
		result: SomeStruct{}
	}
	encoded_string := json.encode(a)
	assert encoded_string == '{"result":{},"id":"some id"}'
	test := json.decode(MyStruct, encoded_string)!
	dump(test)
	dump(a)
	dump(test == a)
	assert test == a
}


```

```console
$ v run D:\v\v\vlib\json\tests\json_option_struct_test.v               
[D:\\v\\v\\vlib\\json\\tests\\json_option_struct_test.v:22] test: MyStruct{
    result: Option(none)
    id: 'some id'
}
[D:\\v\\v\\vlib\\json\\tests\\json_option_struct_test.v:23] a: MyStruct{
    result: Option(SomeStruct{
        test: Option(none)
    })
    id: 'some id'
}
[D:\\v\\v\\vlib\\json\\tests\\json_option_struct_test.v:24] test == a: true
```

@spytheman may be this is a bug?

spytheman:
@get200, I do not understand, what you are asking, since what you show, is buggy on master, but passes here on this PR, due to the fixed comparison.

Why do you point it out again, with a separate code example?
Is there some other bug, different than the one that this PR solves, that is exemplified here?

If there is (even though I could not see it), imho you should file a separate issue for it, and explain what you expected and what you observed.

Delta456 (approved):

PR #25233, opened 2025-09-04
time: always return utc() timezone for Time.unix/0

### LINKED ISSUE #17784: epoch generation is based on timezone
### Describe the bug

The epoch generation is considering timezone into account. This is inconsistent with the standard epoch implementation. 
In linux, you can generate an epoch by `date +%s` command, and that generates an epoch that's independent of timezone. 
Same with python's `print(time.time())` method. 

```v
import time
println(time.now().unix_time())
```

### Expected Behavior

`time.now().unix_time()` should generate the same value as Linux's `date +%s` command

### Current Behavior

On a system that's on a different timezone than UTC, generate the epoch timestamp in V, and generate the unix timestamp in Linux. When you subtract one value from the other, you get a value that's the timezone difference in seconds. 

Example: 
Run this script on a machine that's in a different timezone than GMT

```v
import time
import os

fn main() {
	linux_epoch := os.execute("date +%s").output.trim(r'\s').i64()
	v_epoch := time.now().unix_time()
	diff := linux_epoch - v_epoch
	println("linux_poch = ${linux_epoch} | v_epoch = ${v_epoch} | diff = ${diff}")
}
```

And you will get a value that is equal to the timezone difference in seconds

### Reproduction Steps

Run this script on a machine that's in a different timezone than GMT

```v
import time
import os

fn main() {
	linux_epoch := os.execute("date +%s").output.trim(r'\s').i64()
	v_epoch := time.now().unix_time()
	diff := linux_epoch - v_epoch
	println("linux_poch = ${linux_epoch} | v_epoch = ${v_epoch} | diff = ${diff}")
}
```

And you will get a value that is equal to the timezone difference in seconds

### Possible Solution

_No response_

_No response_

### V version

V 0.3.3 713c95f

MacOs Ventura M1

Le0Developer:
Fixes #17784

We definitely need to clean time up, supporting only two timezones is quite bad and we diverged quite a bit from Go.

It's possible that I forgot to dereference one of the calls in the cgen for json. Not all the branches are covered by tests.

*Copilot summary*

This pull request introduces significant improvements to time handling in the V standard library, focusing on more accurate and explicit management of local and UTC time representations. The changes ensure that comparisons, arithmetic, and formatting involving `time.Time` objects are correct and consistent, especially with respect to time zones and internal representation. Additionally, the pull request updates serialization logic and cleans up obsolete or incorrect tests.

**Time Representation and Comparison Improvements:**

* Added new method `local_unix()` to `time.Time` to return the local UNIX timestamp, and updated the existing `unix()` method to always return the UTC UNIX timestamp for clarity and correctness. (`[vlib/time/time.vR109-R114](diffhunk://#diff-31c344f2cfe6fecdafedef24cc42ca595e29556e07d4efc67c5ab8115e199c88R109-R114)`)
* Updated the equality (`==`) and less-than (`<`) operators for `time.Time` to use both the `is_local` flag and the appropriate local UNIX timestamp, ensuring comparisons are accurate for both local and UTC times. (`[vlib/time/operator.vL6-R15](diffhunk://#diff-a3d381c0223e3b7cdaa9f0d8991a5ca77e3560884795429fa8403504ba4027acL6-R15)`)

**Time Arithmetic and Formatting Adjustments:**

* Changed time addition logic in `add()` to use local UNIX time and handle recalculation of the UNIX timestamp for local times, preventing infinite recursion and ensuring correct results. (`[vlib/time/time.vL138-R163](diffhunk://#diff-31c344f2cfe6fecdafedef24cc42ca595e29556e07d4efc67c5ab8115e199c88L138-R163)`)
* Updated formatting and relative time calculation functions to use the new `unix()` method, guaranteeing consistent UTC-based time calculations. (`[[1]](diffhunk://#diff-31c344f2cfe6fecdafedef24cc42ca595e29556e07d4efc67c5ab8115e199c88L180-R198)`, `[[2]](diffhunk://#diff-31c344f2cfe6fecdafedef24cc42ca595e29556e07d4efc67c5ab8115e199c88L242-R260)`, `[[3]](diffhunk://#diff-bfb6f36bf9173246fa5841608044c7e7830996f0bb08297b8c2d333907efaad4L15-R15)`)

**Serialization and Debugging Updates:**

* Modified JSON encoding logic for `time.Time` in the C backend to use the correct `time__Time_unix()` function, ensuring that the serialized value is always the UTC UNIX timestamp. (`[[1]](diffhunk://#diff-1067da5849e2cc623b0efeb705a1dfb35a5beb58d9e8972cde1dd2945bdc5d48L420-R420)`, `[[2]](diffhunk://#diff-1067da5849e2cc623b0efeb705a1dfb35a5beb58d9e8972cde1dd2945bdc5d48L445-R445)`, `[[3]](diffhunk://#diff-1067da5849e2cc623b0efeb705a1dfb35a5beb58d9e8972cde1dd2945bdc5d48L957-R959)`)
* Enhanced the `debug()` output for `time.Time` to include the `is_local` flag, making it easier to distinguish between local and UTC time objects during debugging. (`[vlib/time/time.vL367-R387](diffhunk://#diff-31c344f2cfe6fecdafedef24cc42ca595e29556e07d4efc67c5ab8115e199c88L367-R387)`)

**Test Suite Cleanup and Corrections:**

* Removed obsolete or incorrect tests that depended on old time behavior, and updated existing tests to reflect the new time handling logic, including correct usage of local and UTC UNIX timestamps and formatting. (`[[1]](diffhunk://#diff-6c972a3c6164cafde5e7324da5b2d8f065262bf1ebb61b8008b846401a59c51dL1-L14)`, `[[2]](diffhunk://#diff-2fd7b5446296ac4dcb322e862c1a68b67fa3afb7c5febb4570925d9d6dc963f3L18-R18)`, `[[3]](diffhunk://#diff-b56c81f693db672f625845374a4bd67074d4d60ba2180b658f6d5b63fc0106b7L93-L103)`, `[[4]](diffhunk://#diff-b56c81f693db672f625845374a4bd67074d4d60ba2180b658f6d5b63fc0106b7L112-L128)`, `[[5]](diffhunk://#diff-0a2aaf850dfca20c675e71c8062ef7823f57fcd9e2a4d248c1ed06e951eb7ed6L396-R398)`)

---

These changes collectively improve the reliability and clarity of time operations in V, especially when dealing with local versus UTC time, and ensure that serialization and comparisons behave as expected across different platforms and time zones.

### DIFF
--- a/vlib/time/time.v
+++ b/vlib/time/time.v
@@ -109,7 +109,7 @@ pub fn (t Time) unix() i64 {
 	return time_with_unix(t.local_to_utc()).unix
 }
 
-// unix returns the UNIX local time with second resolution.
+// local_unix returns the UNIX local time with second resolution.
 @[inline]
 pub fn (t Time) local_unix() i64 {
 	return time_with_unix(t).unix


vlib/time/operator.v
```diff
@@ -3,13 +3,16 @@ module time
 // operator `==` returns true if provided time is equal to time
 @[inline]
 pub fn (t1 Time) == (t2 Time) bool {
-	return t1.unix() == t2.unix() && t1.nanosecond == t2.nanosecond
+	return t1.is_local == t2.is_local && t1.local_unix() == t2.local_unix()
+		&& t1.nanosecond == t2.nanosecond
```
spytheman, line 7: Put the cheaper comparisons (the direct ones for the nanosecond field, which is more likely to fail) first, before the more complex ones (that call functions). That allows for more optimal checks, due to the short circuiting behavior of the && expressions.

vlib/time/time.v
```diff
@@ -135,14 +141,26 @@ pub fn (t Time) add(duration_in_nanosecond Duration) Time {
...
 	if increased_time_nanosecond < 0 {
 		increased_time_second--
 		increased_time_nanosecond += second
 	}
 	res := unix_nanosecond(increased_time_second, int(increased_time_nanosecond))
-	return if t.is_local { res.as_local() } else { res }
+
+	if t.is_local {
+		// we need to reset unix to 0, because we don't know the offset
+		// and we can't calculate it without it without causing infinite recursion
```
spytheman, line 154:
```suggestion
		// and we can't calculate it without it causing infinite recursion
```

spytheman: I am not sure, that producing a completely invalid `.unix` field is good idea.

The user started with some `t` that had a valid .unix field, and decided to add for example 1 second to it.

Why would his code then need to handle a case, where the returned time has .unix == 0 ?

Le0Developer: Unix being 0 is valid and handled correctly inside time.
Not sure if we should even expose the raw Unix…
I agree this isn’t the fun solution, but I don’t know a better one atm.

spytheman: I've checked manually with this program, and the output results seem fine so far:
```v
import time

fn check(t1 time.Time) {
    println('='.repeat(80))
	dump(t1)
	dump(t1.debug())
	dump(t1.is_local) 
	t2 := t1.add(5*time.second)
	dump(t2)
	dump(t2.debug())
	dump(t2.is_local)
	t3 := t2.add(5*time.second)
	dump(t3)
	dump(t3.debug())
	dump(t3.is_local)
}

t1 := time.new(year: 2025, month: 1, day: 1)
check(t1)
check(time.utc())
check(time.now())
```

spytheman: > Not sure if we should even expose the raw Unix…

Hiding it at this point would be a breaking change, and imho is out of scope for this PR.

Le0Developer: Remove it and you’ll see the tests fail. I don’t remember of the top of my head which edge case it was

spytheman: I do not want to remove it.

I've checked manually that what I was afraid of is not true, with the above program, so imho it is fine as it is.

Issue #25206, opened 2025-08-30
`int` should change size based on target system


JalonSolov:
In `Go`, `int` is 32-bit on 32-bit targets, 64-bit on 64-bit targets.  V's `int` should follow these same rules.

Alex has stated (on Discord) that this is an intended change.

I would like to see this change made sooner, rather than later, as it affects other tools such as `go2v`, which currently changes `int` to `isize` in the V code, to force the types to work the same between the languages.

Once this change it made, that forced type change will no longer be required.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Delta456:
`-d new_int` allows `int` size to be changed based on the target system.

medvednikov:
Fixed in V2.

PR #25236, opened 2025-09-04
cgen,builtin: support for 64bit int 1

kbkpbot:
Feature required by #25206

This is the first PR to enable 64bit `int` on an 64bit system.

It can be use with `-d new_int` to enable this feature,

```
v self -d new_int
```

This will build a new V compiler with this feature enabled.

NOTE: Currently the V compiler itself is working, but some works needed to make `vlib` pass all tests.

When enable `-d new_int`, internal `int` will be replaced by a `vint_t` in the compiler.


JalonSolov:
Question:  Would it be simpler to force `int` to be an alias to either `i32` or `i64`?  This could be done early, and inserted into the AST.  You wouldn't need a `max_int`, as it would resolve to either `max_i32` or `max_i64`, etc. - or simply force that as an alias, as well.

JalonSolov:
1. It would be the same as `type MySum = string | string`, just with a different error message.
```
error: sum type MySum cannot hold the type `string` more than once
    1 | type MySum = string | string
      |                       ~~~~~~
```
So something like
```
error: sum type MySum cannot hold the type `i64` more than once (`int` is an alias for `i64`)
    1 | type MySum = int | i64
      |                    ~~~
```

2. Those methods should be on `i64` and `i32` already.

3. That's the tougher one...

vlib/builtin/int.v
```diff
@@ -42,8 +42,8 @@ pub const max_i16 = i16(32767)
 pub const min_i32 = i32(-2147483648)
 pub const max_i32 = i32(2147483647)
 
-pub const min_int = $if new_int ? { min_i64 } $else { min_i32 }
-pub const max_int = $if new_int ? { max_i64 } $else { max_i32 }
+pub const min_int = $if new_int ? { int(min_i64) } $else { int(min_i32) }
```
spytheman: imho move those below min_i64 and max_i64 declarations on the next lines; it can reduce the compilation quirks a bit

spytheman: another option is to just hard code/duplicate the exact numbers, instead of relying that the const comptime evaluation and casting is flawless, something like:
```v
pub const min_int = $if new_int ? { int(-9223372036854775807 - 1) } $else { int(-2147483648) }
```

spytheman:
> For example, does fn C.memset(str voidptr, c int, n usize) voidptr need to be changed to fn C.memset(str voidptr, c i32, n usize) voidptr? 

I think that preferably yes.

`int` in the C side does not change its size between 32 and 64 bit machines, unlike in Go where it does.

I think that using `i32` instead, will be less ambiguous in the C declarations.

spytheman:

JalonSolov:
That might be a separate PR.

Not only do the fn defs need to be changed, but we should have a checker error to not allow `int` in `C.` defs from now on.

spytheman:
> That might be a separate PR.

Yes, definitely.

spytheman:
> but we should have a checker error to not allow int in C. defs from now on.

perhaps a notice first; otherwise it will break a lot of projects

spytheman:
`v fmt` could also just change `int` into `i32` just inside `fn C.` 🤔 .
I do not see downsides to that so far.

JalonSolov:
I'd say a warning rather than a notice.

The code will likely fail in unexpected ways if `int` is used.  At least warning up front will let people know of the problem, and the warning turning to an error will block the "weird" failures in production code.

spytheman:
The usual escalation for such things is notice->warning->error with at least 1 week before each transition, preferably 1 month.

spytheman:
> The code will likely fail in unexpected ways if int is used.

I do not see how, `-d new_int` is still not the default, and while that is not changed, a notice seems completely fine to me - it will give people time to upgrade.

JalonSolov:
Hmm... notice if `new_int` isn't defined, warning/error if it is?

Because if it's on, using `int` is fine as long as the `int` size matches what the C. def needs, but it can badly mess things up if it doesn't match.  For example, using `i64` where `i32` was the requirement (which might just give bad answers), or using `i32` where `i64` was required (which can overwite stack), etc.

vlib/v/gen/c/cgen.v
```diff
@@ -8547,8 +8556,10 @@ fn (mut g Gen) check_noscan(elem_typ ast.Type) string {
 
 // vint2int rename `_vint_t` to `int`
 fn vint2int(name string) string {
-	if name == ast.int_type_name {
-		return 'int'
+	$if new_int ? && (arm64 || amd64 || rv64 || s390x || ppc64le || loongarch64) {
+		if name == ast.int_type_name {
+			return 'int'
+		}
```
JalonSolov, line 8563: Wouldn't it still be correct to always return `int`, whether you enabled this change or not?

If you were returning `i64` vs `i32`, then I could see needing a new block.

kbkpbot: When `new_int` is not enabled, then we get a faster return, only `return name`.
`name` could be something other than `vint_t`, such as `Array_u8`...

spytheman:
> I believe that most of the conversions for 64-bit int inside cgen are already working.
> TODO:

Please do not expand the scope of the PR. It was already good as it is, and given its nature, it *has* to be done in stages.

Adding more stuff here, just complicates things a lot, both for you, for me, and for the users of `c2v` :-|

PR #25239, opened 2025-09-05
builtin: fix C prefix for proc_pidpath()

kbkpbot:
It may issue warning

### DIFF
--- a/vlib/builtin/cfns.c.v
+++ b/vlib/builtin/cfns.c.v
@@ -54,7 +54,7 @@ fn C.isdigit(c int) bool
 fn C.popen(c &char, t &char) voidptr
 
 // <libproc.h>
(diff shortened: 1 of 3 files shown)


JalonSolov:
Looks like a bootstrapping problem.  May need to have both definitions in a merge, then 2nd merge to remove the one without `C.`.

spytheman:
Using a C declaration, does not add the prototype (it expects that the C headers are included, and this one is was in <libproc.h> on macos).

The not C declaration worked because of a quirk in `builtin` - the generated C function names for V functions in there are not prefixed, and functions without bodies, just add the same prototype 2 times ... So in effect declaring a non C body less function, was a way to just add a prototype, so later calls would succeed, unless the function was not present while linking, but it was since it is in macos's libc.

PR #25242, opened 2025-09-05
math.big: replace division with Knuth

tankf33der:
I am pleased and relieved to present a patch for transitioning to `Knuth's` division.

The patch consists of two parts: the first is a reorganization of division processing to reduce memory reallocation, and the second is Knuth's division itself.

Testing of the new division was conducted on random numbers of random length with over 20 billion runs, and if to-from string conversion is included, the number is even higher. Even special edge-case tests were created for dividing numbers that are multiples of 60 bits.

Division performance will significantly improve. Now, dividing a number with 366k digits by one half its length is over 100+ times faster. 

The overall comparison chart of the two division methods is presented below:
division

To followers: If you have code that uses the math.big module, you have a great opportunity to test it before the merge.


spytheman (approved):
Excellent work.
Thank you @tankf33der for improving `math.big` 🙇🏻 .

PR #25245, opened 2025-09-05
Update README.md

449:
ORM now returns an ID when inserting a record so the readme should be updated to reflect this.

### DIFF
--- a/vlib/veb/auth/README.md
+++ b/vlib/veb/auth/README.md
@@ -36,6 +36,7 @@ struct User {
 }
 
 fn main() {
+
 	mut app := &App{
 		db: pg.connect(host: 'localhost', user: 'postgres', password: '', dbname: 'postgres')!
 	}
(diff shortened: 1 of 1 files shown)


vlib/veb/auth/README.md
```diff
@@ -36,6 +36,7 @@ struct User {
 }
 
 fn main() {
+
```
JalonSolov: There is no need to add extraneous blank lines.

449: First official PR, sorry.

JalonSolov: No worries.  Just need to make the example compile properly, and it'll be done.

You can run checks locally with `v check-md docs.md`

449: Ok sorry I think I've made a mess of this. Feel free to delete it

JalonSolov: Naw, looks like it'll work now.

449: Nice. I'm getting the hang of it then, do I need to do anything else?

JalonSolov: Nope.  Only a couple of people have permissions to merge, so you'll need to wait for one of them to find this.  :-)

PR #25258, opened 2025-09-08
ci: use `cache-apt-packages-action` in more CI jobs

spytheman:
- **ci: use the concrete name `ubuntu-24.04` instead of `ubuntu-latest` in the yml files (the *-latest variants cause redirection/more waiting, and change)**
- **ci: use awalsh128/cache-apt-pkgs-action@v1.5.3 in linux_ci.yml too**
- **ci: extract .github/actions/cache-apt-packages-action/action.yml, and use it in linux_ci.yml**
- **remove inputs: and outputs: keys from .github/actions/cache-apt-packages-action/action.yml**
- **fix the freaking tabs in linux_ci.yml**
- **add the libsdl2 packages too; let puzzle_vibes_ci.yml reuse the common cache-apt-packages-action**
- **cleanup, use .github/actions/cache-apt-packages-action/action.yml in more places**

### DIFF
--- a/.github/actions/cache-apt-packages-action/action.yml
+++ b/.github/actions/cache-apt-packages-action/action.yml
@@ -18,6 +18,7 @@ runs:
         version: 1.0
         packages: expect binutils postgresql sqlite3 clang valgrind \
           imagemagick openimageio-tools xvfb xsel xclip \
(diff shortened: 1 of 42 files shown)


medvednikov:
I implemented this in https://github.com/vlang/v/commit/a51ed8416224e15d17196584165f9fae71ec613f

PR #25259, opened 2025-09-08
v: allow decomposing array of interface to comptimecall (just like []string)

### LINKED ISSUE #24870: checker error
### Describe the bug

[this function](https://github.com/einar-hjortdal/peony/blob/21d031d4ca56107944767dec342e23ef4bd3a6f1/conduit_products.v#L6) causes a cgen error `error: cannot convert 'struct string' to 'struct main__RetrieveProductParams'`

### Reproduction Steps

clone, try to compile, cgen error.
Comment out function and it will compile.

### Expected Behavior

no cgen error

### Current Behavior

cgen error

### Possible Solution

_No response_

_No response_

### V version

V 0.4.11 499283e

```
|V full version      |V 0.4.11 47145ea6441e9a2fc6553b1a510fa2fd9df7bdf3.499283e
|OS                  |linux, "openmamba release 2025.5 for x86_64 (rolling)"
|Memory              |7.89GB/31.16GB
|                    |
|V last modified time|2025-07-09 15:01:35
|                    |
|Current working dir |OK, value: /home/einar/Documents/projects/vlang/active/peony
|                    |
|Git version         |git version 2.50.1
|V git status        |weekly.2025.28-8-g499283ed
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.1.1 20250606
|gcc version         |gcc (GCC) 15.1.1 20250606
|clang version       |N/A
|emcc version        |N/A
```

> [!NOTE]
>

felipensp:
Fix #24870

### DIFF
--- a/vlib/v/gen/c/comptime.v
+++ b/vlib/v/gen/c/comptime.v
@@ -162,8 +162,8 @@ fn (mut g Gen) comptime_call(mut node ast.ComptimeCall) {
 			sym_arg := g.table.final_sym(arg.typ)
 
 			arg.expr in [ast.IndexExpr, ast.Ident] && (g.table.type_to_str(arg.typ) == '[]string'
-				|| sym_arg.info is ast.Array
-				&& g.table.final_sym(sym_arg.info.elem_type).kind == .interface)
+				|| (sym_arg.info is ast.Array
+				&& g.table.final_sym(sym_arg.info.elem_type).kind == .interface))
 				&& g.table.type_to_str(param.typ) != '[]string'
 		} else {
 			false


spytheman:
I do not get the fix 🤔 .

Before, cgen generated (through the comptime code in `veb`, at `vlib/veb/veb.v:730`), for 

`app.$method(mut user_context, args)` : 
```c
if (method.args.len > 1 && can_have_data_args) {
    Array_string args = __new_array_with_default(0, (int)(method.args.len + 1), sizeof(string), 0);
...
    main__App_conduit_products_get(app, user_context, ((string*)args.data) [0] );
} else {
...
}
```

which did not compile, since `conduit_products_get` is:
`fn (mut app App) conduit_products_get(mut ctx Context, p RetrieveProductParams) veb.Result {`

and so it expected `p` to be `RetrieveProductParams`, which is:
```v
struct RetrieveProductParams {
    id               ZeroArrayString
    handle           ZeroString
    ...
}
```

while it was passed an array of strings (i.e. `[]string`) by `args`.

Now, it compiles, by changing the cgen output of the `app.$method(mut user_context, args)` call to:
```c
main__App_conduit_products_get(app, user_context, ((main__RetrieveProductParams*)args.data) [0] );
```
... but that code, will pass bullshit data to `main__App_conduit_products_get`, turning a compile time error into a harder to diagnose runtime one 🤔 .

spytheman:
In my opinion, either the checker should be stricter and check in advance that all router methods in veb (the ones returning veb.Result), do not have non string arguments, or the comptime code in `veb` needs to somehow filter those in the route checks, so it does not compile `app.$method(mut user_context, args)` for them.

spytheman:
A simpler self contained case, to test/show the problem, without the peony repo, is:
```v
import veb

pub struct App {}

pub struct Context {
        veb.Context
}

struct MyStruct {
        id   int
        name string
}

        println('> app: ${voidptr(app)}')
        return ctx.text('Hello World')
}

pub fn (app &App) another_method(mut ctx Context, p MyStruct) veb.Result {
        println('> app: ${voidptr(app)}')
        println('> p: ${p}')
        return ctx.text('some response')
}

fn main() {
        mut app := &App{}
        veb.run_at[App, Context](mut app, port: 48091)!
}
```

spytheman:
It does not compile on master, but does here, and produces:
```
#0 11:10:41 ^ fix_comptime_args_decomposing ~/v>v run /home/delian/code/misc/2025_09_10__11/x.v
[veb] Running app on http://localhost:48091/
> app: 7f7ad650efe0
> app: 7f7ad650efe0
> app: 7f7ad650efe0
> p: MyStruct{
    id: 10583436
    name: ''
}
> app: 7f7ad650efe0
> p: MyStruct{
    id: 10583436
    name: ''
}
```
for these requests, done in another shell:
```sh
curl http://localhost:48091/
curl http://localhost:48091/
curl http://localhost:48091/another_method
curl http://localhost:48091/another_method
```

spytheman:
@einar-hjortdal can you please clarify what is the intention of this method ?
```
fn (mut app App) conduit_products_get(mut ctx Context, p RetrieveProductParams) veb.Result {
```

is it intended to be used as a helper, that is called by other methods in the `peony` app, or is it intended to be called by the routing mechanism in `veb` itself?

spytheman:
I wonder if we can make `veb` itself, skip all non `pub` methods 🤔 too.

spytheman:
Here is an isolated case, without `veb`, just with pure V:
```v
struct MyResult {
	value int
}

struct MyStruct {
	id   int
	name string
}

struct App {}

fn (app &App) unrelated1() int {
	println('> ${@LOCATION} | app: ${voidptr(app)}')
	return 123
}

fn (app &App) unrelated2() string {
	println('> ${@LOCATION} | app: ${voidptr(app)}')
	return 'abc'
}

fn (app &App) m1() MyResult {
	println('> ${@LOCATION} | app: ${voidptr(app)}')
	return MyResult{1}
}

fn (app &App) helper(p MyStruct) MyResult {
	println('> ${@LOCATION} | app: ${voidptr(app)}')
	println('> p: ${p}')
	return MyResult{2}
}

fn main() {
	mut app := &App{}
	$for method in App.methods {
		$if method.return_type is MyResult {
			dump(method.name)
			dump(method.args)
			if method.args.len > 0 {
				args := ['p1', 'p2', 'p3']
				dump(args)
				app.$method(args)
			} else {
				app.$method()
			}
		}
	}
}
```

producing:
```
#0 11:32:09 ^ fix_comptime_args_decomposing ~/v>v -g run y.v
[y.v:37] method.name: m1
[y.v:38] method.args: []
> y.v:23, &main.App{}.m1 | app: 7f9bdb3f7fe0
[y.v:37] method.name: helper
[y.v:38] method.args: [MethodParam{
    typ: 109
    name: 'p'
}]
[y.v:41] args: ['p1', 'p2', 'p3']
> y.v:28, &main.App{}.helper | app: 7f9bdb3f7fe0
0x7f9bdb26ed58: at ???: RUNTIME ERROR: invalid memory access
/home/delian/v/vlib/builtin/array.v:786: by array_push_many
/home/delian/v/vlib/strings/builder.c.v:133: by strings__Builder_write_string
/home/delian/v/vlib/builtin/string_interpolation.v:227: by StrIntpData_process_str_intp_data
/home/delian/v/vlib/builtin/string_interpolation.v:706: by str_intp
/tmp/v_1000/y.01K4SC1WSJZ24A71R0QTX9CSW8.tmp.c:1396: by indent_main__MyStruct_str
/tmp/v_1000/y.01K4SC1WSJZ24A71R0QTX9CSW8.tmp.c:1369: by main__MyStruct_str
/home/delian/v/y.v:29: by main__App_helper
/home/delian/v/y.v:42: by main__main
/tmp/v_1000/y.01K4SC1WSJZ24A71R0QTX9CSW8.tmp.c:11227: by main
```

because of the generated C, which implicitly casts the passed []string value, to a MyStruct one:
```c
Array_string args = new_array_from_c_array(3, 3, sizeof(string), _MOV((string[3]){_S("p1"), _S("p2"), _S("p3")}));
_v_dump_expr_Array_string(_S("y.v"), 41, _S("args"), args);
main__App_helper(app, ((main__MyStruct*)args.data) [0] );
```

spytheman:
Hm, I am also not sure how are interfaces related, since in the simpler case, I used only structs and comptime features 🤔 .

spytheman:
The difference in the generated C code is only for the `app.$method(args)` call in the example:
image


(after running `git pull --rebase origin master`)

spytheman:
Thinking more about it, I think that `app.$method(args)` should just check the method call for compatibility with the arguments 🤔 . In each `$for method in App.methods {` iteration, it should know exactly what the method signature is, and what the `args` are, and whether they are compatible or not.

spytheman:
We have `vlib/v/gen/c/comptime.v:191: // TODO: check argument types` ... implementing that TODO, is an option too, since we already do skip methods with more arguments for `app.$method()` calls .

The positive for that, over the checker one, is that it will allow the current flexibility for framework like code (similar to `veb`), while not producing potentially confusing checker errors in the generic framework code, triggered by user source, that the framework was not prepared to handle.

einar-hjortdal:
> @einar-hjortdal can you please clarify what is the intention of this method ?
> 
> ```
> fn (mut app App) conduit_products_get(mut ctx Context, p RetrieveProductParams) veb.Result {
> ```
> 
> is it intended to be used as a helper, that is called by other methods in the `peony` app, or is it intended to be called by the routing mechanism in `veb` itself?

It was intended to be called by veb route handlers exclusively. It was a helper to split the concerns of route handlers. veb should not have known it existed.
I have since rewritten all `(mut app App)counduit_(mut ctx Context, ...)` functions to `counduit_(mut app App, mut ctx Context, ...)`

felipensp:
I think we should make fix this magic []string decomposing stuff. And keep only the explicit `...args` decomposing and the multi return one (e.g. f(g()) where g() returns multi values).

felipensp:
@spytheman about this issue, lets just make the checker stricter so ?

PR #25270, opened 2025-09-09
builtin: fixed the thread naming issue by using RaiseException (0x406D1388) on windows.

get200:
On Windows, SetThreadName traditionally raises exception 0x406D1388 to set a thread name in the debugger. This exception is harmless and serves only as a debugging aid.Certain libraries might raise this exception and crash the program; it should be ignored or filtered.


spytheman:
See also https://learn.microsoft.com/en-us/archive/blogs/stevejs/naming-threads-in-win32-and-net .

spytheman (approved):
Thank you 🙇🏻‍♂️.

PR #25276, opened 2025-09-10
eval: add more complex comptime-cond support

kbkpbot:
As the `eval` need to execute some `builtin`, that need to enhance the `eval` comptime-cond support.


spytheman (approved):
Excellent work.

PR #25278, opened 2025-09-10
x.crypto.ascon: improves the core of Ascon permutation routine

blackshirt:
This PR improves the core of `x.crypto.ascon` permutation routine in the mean of internal routine `ascon_pnr` function.
After some profiling, the most widely called routines was 3 function of the core permutation routines, ie, `ascon_pnr`, `ascon_perm` and utility `ascon_rotate_right` used in permutation.
The number before patch using `sum256` as a test case:

```bash
== Before patch ==
Benchmarking ascon.sum256 before patch ...
Average time ascon.sum256 before patch: 11 µs
```

And here the number after patch

```bash
== After patch ==
Benchmarking ascon.sum256 after patch ...
Average time ascon.sum256 after patch: 6 µs
```

Nearly 50% improvement in performance was a good addition. The numbers i talk on [Discord](https://discord.com/channels/592103645835821068/592106336838352923/1415203902441263105) confirm this approach.

This patch fundamentally unifies `ascon_pnr` and `ascon_perm` into single one routine. Its also embed directly bits rotate right logic onto this single routine and improves all performances.
This patch also contains some small cleans up, updates readme file. and other small thing.

### DIFF
--- a/vlib/builtin/linux_bare/libc_impl.v
+++ b/vlib/builtin/linux_bare/libc_impl.v
@@ -4,6 +4,7 @@ import dlmalloc
 
 __global global_allocator dlmalloc.Dlmalloc
 
+@[export: 'memcpy']
 @[unsafe]
 pub fn memcpy(dest voidptr, src voidptr, n usize) voidptr {
 	dest_ := unsafe { &u8(dest) }
@@ -22,6 +23,7 @@ fn __malloc(n usize) voidptr {
 	return unsafe { global_allocator.malloc(n) }
 }
 
+@[export: 'strlen']
 @[unsafe]
 fn strlen(_s voidptr) usize {
 	s := unsafe { &u8(_s) }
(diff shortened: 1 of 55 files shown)


vlib/x/crypto/ascon/aead128_test.v
```diff
@@ -2,8 +2,9 @@
 // Use of this source code is governed by an MIT license
 // that can be found in the LICENSE file.
 //
+module ascon
+
 import encoding.hex
-import x.crypto.ascon
```
spytheman: no, please use `import x.crypto.ascon`

blackshirt: I think its my mistake when reverting my git, i would change it

spytheman:
You can add the code that you run for the benches in the `bench/crypto` folder or in `vlib/x/crypto/ascon/bench/` . That way it can be run by everyone, interested in changing the module in the future.

blackshirt:
> You can add the code that you run for the benches in the `bench/crypto` folder or in `vlib/x/crypto/ascon/bench/` . That way it can be run by everyone, interested in changing the module in the future.

Okey .. i would add it in next time of cycles.Its on plan

spytheman:
> Okey .. i would add it in next time of cycles.Its on plan

Thank you 🙇🏻 .

PR #25264, opened 2025-09-08
cgen: prefix builtin methods with builtin__

dy-tea:
We would like to prevent user-defined variable names from being able to conflict with c functions generated by the compiler. 

Most simple thing in most cases is prefixing with _ but we need to specifically check for builtin types in some places as we don't want to add this to user-defined functions which will then reintroduce the issue (maybe we want a separate prefix for user-defined variables as well so they don't conflict with c functions?).

I have some commented out lines in the test at the moment for array and map initialization. Wasn't sure how to handle this.

I handle `int_literal` and `float_literal` separately but there might be a more elegant way to do this?

There are some places where I only prefix an underscore and was wondering if it's more efficient to use the string plus operator or string formatting, I currently mostly use string formatting but it can be changed.

This does not handle name collisions with included c functions like `printf`. Maybe have a per-module c_reserved array with the included function names? Not sure about the best solution for this.

Since this adds quite a few `is_builtin()` checks it likely will have some performance impact. It didn't seem too bad but I didn't do any extensive tests.

PR #25280, opened 2025-09-10
checker: prevent usage of imported module name prefix as identifier n…

dy-tea:
Given the below program:
```v
import time

fn builtin__string_str() {
}

fn time__utc() {
}

fn os__getwd() {
}

fn main() {
	builtin__string_str := 'Hello V!'.str()
	time__now := time.now()
	os__log := 'Hello V!'
	println(builtin__string_str)
	println(time__now)
	println(os__log)
}
```
Currently will fail to compile on the C backend, as the identifiers `builtin__string_str`, `time__utc` and `time__now` conflict with the function definitions for the respective methods.

This should be a checker error, but the identifiers `os__getwd` and `os__log` should not produce an error because `os` is never imported and as such there is no name conflict.

This is an implementation of the fix suggested by @JalonSolov in #25264.


dy-tea:
Well currently if we do:
```v
import time

fn main() {
	time__now := time.now()
	println(time__now)
}
```
This will generate the C code:
```c
VV_LOC void main__main(void) {
	time__Time time__now = time__now();
	builtin__println(time__Time_str(time__now));
}
```
```
cc: /tmp/v_1000/main.01K4TMSYGE7C39NK336P7GEGB0.tmp.c:6773: error: function pointer expected
```
Since `time__now` is the translated form of `time.now()`, having a variable with the same name will cause compilation to fail.

There are a few ways to prevent this, such as:
- Disallowing the prefix as I do it here
- prefixing all method calls with `_` when generating c code so users can't collide variable names
- adding a `_var_` prefix or similar to all variables so they cannot collide with function names

Delta456:
I don't like the first option. We should go with either 2nd or 3rd.

dy-tea:
I believe option 2 will have the better performance and option 3 will generate the smallest code. I will try implementing 2 and see how it goes.

vlib/v/checker/assign.v
```diff
@@ -444,6 +444,24 @@ fn (mut c Checker) assign_stmt(mut node ast.AssignStmt) {
...
+									import_sym.alias
+								} else {
+									import_sym.mod.all_after_last('.')
+								}
+								if prefix == module_name {
+									c.error('identifier cannot use prefix `${prefix}__` of imported module `${prefix}`',
+										left.pos)
+								}
+							}
+						}
```
spytheman: please extract a method, and call it in both places

vlib/v/checker/assign.v
```diff
@@ -444,6 +444,24 @@ fn (mut c Checker) assign_stmt(mut node ast.AssignStmt) {
...
 							c.error('duplicate of an import symbol `${left.name}`', left.pos)
 						}
+						if left.name.starts_with('builtin__') {
+							c.error('identifier cannot use reserved `builtin__` prefix',
+								left.pos)
+						}
+						if left.name.contains('__') {
+							prefix := left.name.all_before('__')
+							for import_sym in c.file.imports {
+								module_name := if !import_sym.alias.is_blank() {
```
spytheman: why call .is_blank(), instead of just comparing to '' ?

spytheman: Imho this whole loop should be done just once in `change_current_file`, and the result stored in a Checker field, then here, it can be just `if prefix in c.short_module_names {` .

dy-tea:
Another CI timeout

JalonSolov:
This time, it timed out trying to download the wasmer installer.  <sigh>  I restarted the job, and it succeeded.

Delta456:
I hope this solution is temporary and we don't have to ban such usages.

PR #25281, opened 2025-09-10
native: structs multi_assign and multi_return

Eliyaan:
- add support multiple assignment and multiple return for structs

### DIFF
--- a/vlib/v/gen/native/amd64.v
+++ b/vlib/v/gen/native/amd64.v
@@ -2813,7 +2813,8 @@ fn (mut c Amd64) return_stmt(node ast.Return) {
 								c.add(Amd64Register.rax, size % 8)
 								c.add(Amd64Register.rdx, size % 8)
 								c.mov_deref(Amd64Register.rcx, Amd64Register.rax, ast.i64_type_idx)
-								// Doesn't it write too far as the size is not 64bits?
+								// TODO: check if it does not write too far as the size of
+								// the remaining data is not 64bits
 								c.mov_store(.rdx, .rcx, ._64)
 							}
 							c.mov_var_to_reg(c.main_reg(), LocalVar{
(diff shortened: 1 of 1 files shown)


vlib/v/gen/native/amd64.v
```diff
@@ -2812,6 +2813,7 @@ fn (mut c Amd64) return_stmt(node ast.Return) {
 								c.add(Amd64Register.rax, size % 8)
 								c.add(Amd64Register.rdx, size % 8)
 								c.mov_deref(Amd64Register.rcx, Amd64Register.rax, ast.i64_type_idx)
+								// Doesn't it write too far as the size is not 64bits?
```
spytheman: Please, reformulate it as a statement, like:
`// TODO: check if it does not write too far etc`
or
`// NOTE: This writes too far, but it is ok, because etc`
or just delete it, if the answer is known.

Try to think as the reader, when writing code comments - asking a question, stops the flow of reading, until the answer is known, or until the reader decides that he/she does not care. If the reader is trying to fix a bug, he has to then understand several contexts - the code itself, the question, the intent of the original writer of the code/comment, and the potential answer, which is undefined and needs thinking.

Formulating it as a statement of fact, or a TODO, instead of a question, makes it clearer, and more actionable.

vlib/v/gen/native/amd64.v
```diff
@@ -2841,7 +2843,38 @@ fn (mut c Amd64) return_stmt(node ast.Return) {
...
+					c.mov_store(.rdx, .rcx, match data_size {
+						1 { ._8 }
+						2 { ._16 }
+						4 { ._32 }
+						else { ._64 }
+					})
+					s_size -= data_size
+					c.add(Amd64Register.rax, data_size)
+					c.add(Amd64Register.rdx, data_size)
+				}
```
spytheman: Can you please comment more on the intention of this block?

For example, why is there a second for loop, instead of a single if ?
If there is a remainder of `s_size % 8` of say 5 bytes, would it not be handled with just 1 iteration?

vlib/v/gen/native/amd64.v
```diff
@@ -3001,7 +3035,31 @@ fn (mut c Amd64) multi_assign_stmt(node ast.AssignStmt) {
...
+				c.mov_store(.rax, .rcx, match data_size {
+					1 { ._8 }
+					2 { ._16 }
+					4 { ._32 }
+					else { ._64 }
+				})
+				s_size -= data_size
+				c.add(Amd64Register.rax, data_size)
+				c.add(Amd64Register.rdx, data_size)
+			}
```
spytheman: This looks nearly identical to the first block above; imho extract it into a method of its own.

PR #25282, opened 2025-09-11
x.crypto.ascon: improves single-shot functions of ascon hasing variant

blackshirt:
This small patch is the series of previous merged [25278](https://github.com/vlang/v/pull/25278) PR.
In this PR was focus on improving ascon hashing routines that works in single-shot way, ie, in the mean of `sum256`, `xof128` and `cxof128`.
The current approach was using internal `Digest` opaque thats working in streaming-way. For this single-shot function, this is
not really needed, and adds small overhead on it.
So, this patch changes the situation by working on generic and raw function for the single-shot function. All the rest still work as previous.

The result was included in the new benchmark directory.

This PR also adds benhmark into the `bench` directory. Its contains 3 of benchs,

- Benchmarked single-shot function patched with this PR
- Benchmarked ascon `sum256` and `xof128` compared to builtin `crypto.sha256` and `sha3.shake256` for xof output.
- Benchmarked aead of Ascon-AEAD128 `encrypt` and `decrypt` compared to already stocked `x.crypto.chacha20poly1305` encrypt and decrypt.

The results was very promising, based on the results number

- Its competitive enought with `crypto.sha256` even outperforms it for small sized of messages
- Its outperforms the`sha3.shake256` in all size of messages for `xof128` benchmark.
- For AEAD benchmark, its also outperforms `x.crypto.chacha20poly1305` in encryption and decryption

TODO: benchmarking it with others lang implementation

Cheers,

### DIFF
--- a/vlib/x/crypto/ascon/README.md
+++ b/vlib/x/crypto/ascon/README.md
@@ -7,7 +7,7 @@ thats provides Authenticated Encryption, Hash, and Extendable Output Functions.
 See the [NIST.SP.800-232 Standard](https://doi.org/10.6028/NIST.SP.800-232) for more detail.
 
 This module mostly implements all the features availables on the document.
-Its currently implements: 
+It currently implements:
 - `Ascon-Hash256`, Ascon-based hashing implementation that produces 256-bits output.
 - `Ascon-XOF128`, Ascon-based eXtendable Output Function (XOF) where the output size of 
 the hash of the message can be selected by the user.


vlib/x/crypto/ascon/README.md
```diff
@@ -7,7 +7,7 @@ thats provides Authenticated Encryption, Hash, and Extendable Output Functions.
 See the [NIST.SP.800-232 Standard](https://doi.org/10.6028/NIST.SP.800-232) for more detail.
 
 This module mostly implements all the features availables on the document.
-Its currently implements:
+Its currently implements:
```
spytheman:
```suggestion
It currently implements:
```

PR #25288, opened 2025-09-12
x.crypto.ascon: small changes of forgotten bits

blackshirt:
This small patch improves `ascon_generic_hash`, even not too much, but its valuable to sent. Its also reduces `ascon_pnr` alloc to single u64 temporary variable.


vlib/x/crypto/ascon/ascon.v
```diff
@@ -59,22 +58,15 @@ fn ascon_pnr(mut s State, nr int) {
...
-		s.e3 = s.e3 ^ (~s.e4 & x0) // t4, change s.e0 to x0
-		s.e4 = y0
+		// Set temp var to value
+		x0 = s.e0 & ~s.e4
+
+		s.e0 ^= s.e2 & ~s.e1
+		s.e2 ^= s.e4 & ~s.e3
+		s.e4 ^= s.e1 & ~s.e0
+		s.e1 ^= s.e3 & ~s.e2
+		s.e3 ^= x0
```
spytheman: That transformation seems a bit weird to me (because `s.e1 ^= s.e3 & ~s.e2` uses `s.e2` which was already changed 2 lines above).

Are you sure that the order should be changed, instead of using the one before?

blackshirt: Yeah, its Its look weird to me too ...the order is matter ..
Actually, i'm prefer for the previous one in the readibilty manner ..
The improvement not too much gain ..even its a worth when in heavy calculated round .. 
The current state was mostly competitive enough with other builtins.  
I can revert this one,

blackshirt: So, also i don't think the check for `nr` in `ascon_prnd` are really needed..this module only use 8 and 12 as permutation round number throught defined constant, `ascon_prnd_8` and `ascon_prnd_12` respectively ...i think we can omit this check ? What do you think ?

spytheman: I agree, you can replace the check in `ascon_pnr` with just `assert nr > 0 && nr < 17` (asserts are eliminated with `-prod`), or just remove the check completely since the API is not exposed.

Another option is to:
```v
enum PrndEnum {
   ascon_prnd_6 = 6
   ascon_prnd_8 = 8
   ascon_prnd_12 = 12
}
```
and then have `fn ascon_pnr(mut s State, nr PrndEnum) {`,
with no more additional runtime checks - the compiler will then check things at compile time.

blackshirt: Okey..i would come with that approach

PR #25292, opened 2025-09-12
cgen: fix option generic passing to `T` param

### LINKED ISSUE #24466: cgen, generics: C gen error when using optional element in array

### LINKED ISSUE #25227: error: cannot convert 'struct main__Address' to 'struct _option_main__Address'
### Describe the bug

This error happens when using an optional inner struct (?Address in this case):
```
cc: /tmp/v_1000/temp.01K47PKK8Y9305R7FV44D58E15.tmp.c:9842: error: cannot convert 'struct main__Address' to 'struct _option_main__Address'
builder error: 
==================
```
### Reproduction Steps

v temp.v
```v
import x.json2

struct Person {
mut:
	name     string
	address  ?Address
}

struct Address {
	street string
	city   string
	state  string
	zip    string
}

fn main() {
	resp := '{"name": "Bob", "address": {"street": "123 Main St", "city": "Anytown", "state": "CA", "zip": "12345"}}'
	person := json2.decode[Person](resp)!
}
```
### Expected Behavior

Should not error

### Current Behavior

Above error is displayed on compile

### Possible Solution

_No response_

_No response_

### V version

V 0.4.11 f6b60e4

|V full version      |V 0.4.11 d31aaecc42286a63a7a2e95b24d48a2e0bfbfda4.f6b60e4
|OS                  |linux, Solus 4.7 Endurance
|Processor           |8 cpus, 64bit, little endian, AMD Ryzen Embedded V1605B with Radeon Vega Gfx
|Memory              |7.99GB/30.31GB
|                    |
|Git version         |git version 2.49.1
|V git status        |f6b60e4d
|.git/config present |true
|                    |
|cc version          |cc (Solus) 15.1.1 20250508
|gcc version         |gcc (Solus) 15.1.1 20250508
|clang version       |N/A
|emcc version        |N/A
|glibc version       |ldd (Solus) 2.41

> [!NOTE]
>

### LINKED ISSUE #25267: V error in `json2.encode` with array of sumtype `int | ?int`
### Describe the bug

`json` encodes an array of sum type `int | ?int` "correctly" while `json2` prints a V error

### Reproduction Steps

```v
import json
import x.json2

type Elem = int | ?int

const empty = ?int(none)
const array = [ Elem(1), Elem(empty), 3 ]

fn main() {
	dump(array)
	e := json.encode(array)
	dump(e) // '[1,{},3]'
	dump(json2.encode(array)) // V error in json2
}
```

### Expected Behavior

For V array `[Elem(1), Elem(Option(none)), Elem(3)]` we expect `[1,null,3]` or `[1,{},3]`

### Current Behavior

- json encodes "correctly" mapping `none` into empty map `{ }`
```
[1,{},3]
```

- json2 V error:
```
$ v run 250908-f.v 
/home/jorge/v/vlib/x/json2/encoder.v:165:31: error: cannot use `?int` as `int`, it must be unwrapped first in argument 1 to `x.json2.Encoder.encode_value_with_level`
  163 |         $for v in val.variants {
  164 |             if val is v {
  165 |                 e.encode_value_with_level(val, level, mut buf)!
      |                                           ~~~
  166 |             }
  167 |         }


```

### Possible Solution

Check how `json` parses sum type `int | ?int` to be incorporated in `json2`.

_No response_

### V version

V 0.4.11 847c15e

Ubuntu

felipensp:
Fix #25267
Fix #24466
~Fix #25227~ - it turns the cgen error into a checker one, but ideally the code from the issue should compile and work, so it needs more work in another PR to x.json2 .

### DIFF
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -1739,7 +1739,7 @@ pub fn (mut t Table) convert_generic_type(generic_type Type, generic_names []str
 		} else {
 			rtyp = rtyp.clear_flag(.generic)
 		}
-		if typ.has_flag(.option) {
+		if !generic_type.has_flag(.result) && typ.has_flag(.option) {
 			rtyp = rtyp.set_flag(.option)
 			if generic_type.is_ptr() {
 				rtyp = rtyp.set_flag(.option_mut_param_t)


vlib/v/gen/c/fn.v
```diff
@@ -2809,10 +2808,10 @@ fn (mut g Gen) ref_or_deref_arg(arg ast.CallArg, expected_type ast.Type, lang as
...
 		&& !expected_type.has_flag(.option) {
 		// allow to pass val.$(filed.name) where T is expected, doing automatic unwrap in this case
-		styp := g.base_type(arg_typ)
-		g.write('*(${styp}*)')
-		g.expr_with_cast(arg.expr, arg_typ, expected_type)
-		g.write('.data')
+		// styp := g.base_type(arg_typ)
+		// g.write('*(${styp}*)')
+		g.expr_with_opt(arg.expr, arg_typ, expected_type.set_flag(.option))
+		// g.write('.data')
```
spytheman: Please delete the commented code.
Keeping it, has no utility imho, since unlike debugging code, it can not be used to quickly check/trace stuff later.

vlib/v/gen/native/tests/assign.vv
spytheman, line 1: this file has only whitespace changes in it, imho revert it

vlib/v/gen/native/tests/multi_assign.vv
spytheman, line 1: imho revert this too, it is just whitespace changes, that are not relevant to the PR

vlib/v/gen/native/tests/struct.vv
spytheman, line 1: imho revert this too

vlib/x/json2/decoder2/decode.v
```diff
@@ -830,7 +830,7 @@ fn (mut decoder Decoder) decode_map[K, V](mut val map[K]V) ! {
 }
 
 fn create_value_from_optional[T](val ?T) ?T {
-	return T{}
+	return ?T{}
```
spytheman: Would not `return T{}` mean the same?

vlib/x/json2/encode.v
```diff
@@ -400,6 +400,13 @@ fn (mut encoder Encoder) encode_struct[T](val T) {
 			if field_info.is_omitempty {
 				$if value is $option {
 					write_field = check_not_empty(value)
+					$if value is ?string {
+						write_field = value ? != ''
```
spytheman: Where would the `?` propagate?

The function signature is `fn (mut encoder Encoder) encode_struct[T](val T) {`
it does not return `?`.

vlib/x/json2/encode.v
```diff
@@ -444,7 +451,7 @@ fn (mut encoder Encoder) encode_struct[T](val T) {
 				if val.$(field.name) == none {
 					unsafe { encoder.output.push_many(null_string.str, null_string.len) }
 				} else {
-					encoder.encode_value(val.$(field.name))
+					encoder.encode_value(val.$(field.name) ?)
```
spytheman: Here, the `?` should not be needed (and it should not work anyway, since the function does not return `?`), because of the smartcast.
`val.$(field.name)` should be already unwrapped.

vlib/v/gen/c/fn.v
felipensp: Sure.

vlib/v/gen/native/tests/struct.vv
felipensp: oh, it was the pre-hook one on merged files.

spytheman (approved):
Excellent work.

Issue #25291, opened 2025-09-12
msvc fixed array init bug


kbkpbot:
### Describe the bug

`vlib/v/tests/modules/sub/global_fixed_array_test.v` has bug with `msvc` .

### Reproduction Steps

a modified version of `vlib/v/tests/modules/sub/global_fixed_array_test.v`

bug.v
```v
@[has_globals]
module main

import sub.foo

struct DummyStruct {
	dummy_item int    // change `i32` to `int` will fire the bug
}

// vfmt off
__global (
	d [foo.num_elements]DummyStruct
)
// vfmt on

const f = [foo.num_elements]DummyStruct{}

fn test_main() {
	assert dump(foo.num_elements) == 2
	assert dump(f) == [foo.num_elements]DummyStruct{}
	assert dump(d) == [foo.num_elements]DummyStruct{}
}
```

```
v -cc msvc bug.v
```

### Expected Behavior

compile 

### Current Behavior

```
D:\v\v\v>v vlib/v/tests/modules/sub/global_fixed_array_test.v -showcc -keepc -cc msvc
> C compiler cmd: "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.44.35207\bin\Hostx64\x64\cl.exe" "@C:\Users\DDT\AppData\Local\Temp\v_0\global_fixed_array_test.tmp.c.rsp"
> C compiler response file "C:\Users\DDT\AppData\Local\Temp\v_0\global_fixed_array_test.tmp.c.rsp":
[...]

PR #25293, opened 2025-09-12
cgen: fix fixed-array const initializer

kbkpbot:
Fix issue #25291

In `msvc`,  initializer for the global array uses compound literals (which are not constant expressions in C) and also uses a cast-like syntax that is not allowed in constant expressions.

And this PR also enable `i32` in the check.

```c
Array_fixed_main__DummyStruct_2 _const_main__f = {(main__DummyStruct){.dummy_item = 0,}, (main__DummyStruct){.dummy_item = 0,}}; // fixed array const
```

=>

```c
Array_fixed_main__DummyStruct_2 _const_main__f = {{.dummy_item = 0,}, {.dummy_item = 0,}}; // fixed array const
```


spytheman (approved):
Excellent work.
Thank you @kbkpbot 🙇🏻.

PR #25298, opened 2025-09-13
vfmt: support for 64bit int 2

kbkpbot:
This is second PR to support 64bit/32bit int as PR #25236 is the first one.

This PR introduce a new switch `-new_int` to `vfmt` and it let `vfmt` translates all `int` to `i32`, if

1. The whole module is `translated`, that is, a attribute `@[translated]` is at the beginning of the file, or
2. A C function declaration, such as `C.abc(a int) int`.

example:
a.v

```v
@[translated]
module aa

fn C.abc(a int, b int)

fn abc(x int, y int) {
        s := int(123)
        dump(s)
}
```

```
v fmt -new_int a.v
```

=>

```v
@[translated]
module aa

fn C.abc(a i32, b i32)

fn abc(x i32, y i32) {
        s := i32(123)
        dump(s)
}
```

another example:
b.v

```v
module main

fn C.abc(a int, b int, c charptr) int

fn abc(a int, b int, c charptr) int
```

```
v fmt -new_int b.v
```

=>

```v
module main

fn C.abc(a i32, b i32, c charptr) i32

fn abc(a int, b int, c charptr) int
```

### DIFF
--- a/cmd/tools/check_os_api_parity.v
+++ b/cmd/tools/check_os_api_parity.v
@@ -99,7 +99,7 @@ fn gen_api_for_module_in_os(mod_name string, os_ pref.OS) string {
 				fn_mod := s.modname()
 				if fn_mod == mod_name {
 					fn_signature := b.table.stringify_fn_decl(&s, mod_name, map[string]string{},
-						false, false)
+						false)
 					fline := '${fn_mod}: ${fn_signature}'
 					res << fline
 				}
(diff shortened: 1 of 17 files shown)


spytheman:
Add a new field on `ast.Table`, then you will not have to modify all the intermediate function signatures. You can set it to true in `cmd/tools/vfmt.v`.

spytheman (approved):
Excellent work.

PR #25299, opened 2025-09-13
native: beginning of array support

Eliyaan:
- generate arrays  in assignments `a := []Type{len: x, cap: y}`
- add support for indexing arrays (read/write)


vlib/builtin/builtin.c.v
```diff
@@ -486,7 +486,9 @@ pub fn malloc_noscan(n isize) &u8 {
 		_memory_panic(@FN, n)
 	}
 	mut res := &u8(unsafe { nil })
-	$if prealloc {
+	$if native {
+		res = unsafe { C.malloc(n) }
+	} $else $if prealloc {
```
spytheman, line 491: Why does it need this special case?
That is the default on line 511.

vlib/v/gen/native/gen.v
```diff
@@ -929,6 +931,22 @@ fn (mut g Gen) get_type_size(raw_type ast.Type) i32 {
...
+				ast.int_type] {
+				f_size := g.get_type_size(f_typ)
+				f_align := g.get_type_align(f_typ)
+				padding := (f_align - size % f_align) % f_align
+				strc.offsets << size + padding
+				size += f_size + padding
+				if f_align > align {
+					align = f_align
+				}
+			}
```
spytheman, line 946: When will this loop end 🤔 ?
`f_typ` is not modified in the body, and there are no `break` statements too...

vlib/builtin/builtin.c.v
Eliyaan: I do not know why but it was not working without this

vlib/v/gen/native/gen.v
Eliyaan: When all the elements of the arrays are iterated over. (I discovered a few days ago that there was a "while version: while f_typ is in array" : `for (ftyp in array)` with the parenthesis)

vlib/builtin/builtin.c.v
Eliyaan: No code was being generated

vlib/v/gen/native/gen.v
spytheman: you are right - it is not checking if `f_typ` is in the array, but iterating over the array indeed

vlib/builtin/builtin.c.v
spytheman: That seems like a bug that has to be fixed 🤔. 
I may be wrong, but I think that the compile time branching is already done in the parser and checker, before the backends.

spytheman (approved):
Excellent work.
Thank you @Eliyaan.

Issue #25301, opened 2025-09-14
V should not allow export same const name twice


kbkpbot:
### Describe the bug

V should not allow export same const name twice

### Reproduction Steps

cc.v
```v
module main

@[export: 'asm_instrs']
const asm_instrs_a = [0]!

@[export: 'asm_instrs']
const asm_instrs_b = [0]!

fn main() {
        dump(asm_instrs_a)
        dump(asm_instrs_b)
}
```

```
v cc.v
```

### Expected Behavior

should issue a error msg, about that the const name `asm_instrs` already be used/exported.

### Current Behavior

```
$ v cc.v
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/cc.01K52RBMCAJ0SKVQRPY4T6NCMW.tmp.c:1237: error: redefinition of 'asm_instrs'
=================================================================
(You can pass `-cg`, or `-show-c-output` as well, to print all the C error messages).
builder error:
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .

```

### Possible Solution

move `@[export]` check to checker, not in `cgen`.

### V version

V 0.4.11 50d03eb7b93bcdcfa90a43fc6cfa68916f38d28e.c221b32

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 50d03eb7b93bcdcfa90a43fc6cfa68916f38d28e.c221b32
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |12.24GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-09-13 23:47:19
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /media/HD/github/kbkpbot/v/vlib/v/checker
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.36-65-gc221b322-dirty
|.git/config present |true
|                    |
[...]

PR #25302, opened 2025-09-14
checker: redefine export const global func names

kbkpbot:
Fix issue #25301

1. disallow export same const name twice;
2. disallow export const name which has been defined as a global name, `__global`;
3. disallow export same function name twice.


JalonSolov:
I would just keep track of all exported names, and not allow duplicates.  It should not matter what type of thing it is, only whether or not the exported name is the same.

spytheman (approved):

Issue #25300, opened 2025-09-13
export const cause error: '_const_main__asm_instrs' undeclared


kbkpbot:
### Describe the bug

when a const declare with `@[export]` and set the value to a fixed array, it will cause v compile fail.

### Reproduction Steps

c.v
```v
module main

@[export: 'asm_instrs']
const asm_instrs = [0]!

fn main() {
        dump(asm_instrs)
}
```

```
v c.v
```

### Expected Behavior

compile

### Current Behavior

```
$ v c.v
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/c.01K51QD6SPKHEX2PY4MM4QBM5V.tmp.c:5270: error: '_const_main__asm_instrs' undeclared
=================================================================
(You can pass `-cg`, or `-show-c-output` as well, to print all the C error messages).
builder error:
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```

### Possible Solution

It seems `@[export]` does not work as expected.

### Additional Information/Context

set a var length array, it can compile, but the generated C file has no the exported symbol name : `asm_instrs`, only `_const_main__asm_instrs`.

```v
const asm_instrs = [0]     // remove the `!`, it can pass compile, but missing `asm_instrs`
```

### V version

V 0.4.11 50d03eb7b93bcdcfa90a43fc6cfa68916f38d28e.1bdf70f

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 50d03eb7b93bcdcfa90a43fc6cfa68916f38d28e.1bdf70f
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |7.56GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-09-13 10:51:09
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/const
|                    |
|Git version         |git version 2.43.0
[...]

PR #25303, opened 2025-09-14
cgen,parser,markused: add export const support

kbkpbot:
Fix issue #25300

This PR enable `@[export]` const in V.


spytheman (approved):
Excellent work.

PR #25305, opened 2025-09-14
scanner: uncomment working rune test 

Delta456:
Uncomment working tests in `scanner`


spytheman:
Good find.
Thank you @Delta456.

Issue #25295, opened 2025-09-13
comptime if does not work when using js backend


kbkpbot:
### Describe the bug

When using comptime if `$if` in different files, it will not work as expected when using js backend.

### Reproduction Steps

modify `vlib/builtin/js/int.js.v` add a single line:

```v
pub const val = $if amd64 { 1 } $else { 0 }
```

then create a new user file:
j.js.v
```v
module main

fn main() {
        println(val)
}
```

```
v -b js j.js.v
```

### Expected Behavior

compile and run ok

### Current Behavior

compile error:
```
$ v -b js j.js.v
V panic: checker error: cond result idx string not found => [|j.js.v|token.Pos{
    len: 5
    line_nr: 34
    pos: 832
    col: 20
    last_line: 34
}|]
v hash: a8d200a
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:20100: at builtin___v_panic: Backtrace
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:45980: by v__gen__js__JsGen_comptime_if
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:50063: by v__gen__js__JsGen_gen_if_expr
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:48358: by v__gen__js__JsGen_expr
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:47728: by v__gen__js__gen
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:30681: by v__builder__jsbuilder__gen_js
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:30658: by v__builder__jsbuilder__build_js
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:30653: by v__builder__jsbuilder__compile_js
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:30576: by v__builder__Builder_rebuild
/tmp/v_1000/js_builder.01K507CV0JFZ3BG4F9XG45BEAN.tmp.c:29806: by v__builder__compile
[...]

PR #25308, opened 2025-09-14
checker,gen: introduce if match branch's id

kbkpbot:
The original comptime branch identification mechanism for `$if` `$match` used a file name + location approach for identification. However, in some cases, it is difficult to obtain the current processing file name. This leads to identification errors. To avoid this, an ID mechanism has been introduced, which enables unified global management of the branch IDs for comptime `$if` `$match`. The ID is allocated in the `checker` and used in the `gen`.

Related to issue #25295

### DIFF
--- a/vlib/builtin/js/int.js.v
+++ b/vlib/builtin/js/int.js.v
@@ -17,16 +17,8 @@ pub const max_i32 = i32(2147483647)
 pub const min_i64 = i64(-9223372036854775807 - 1)
 pub const max_i64 = i64(9223372036854775807)
 
-pub const min_int = $if new_int ? && (arm64 || amd64 || rv64 || s390x || ppc64le || loongarch64) {
-	int(min_i64)
-} $else {
-	int(min_i32)
-}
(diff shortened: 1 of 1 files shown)


vlib/v/checker/if.v
```diff
@@ -107,9 +107,14 @@ fn (mut c Checker) if_expr(mut node ast.IfExpr) ast.Type {
...
 				// The first part represents the current context of the branch statement, `comptime_branch_context_str`, formatted like `T=int,X=string,method.name=json`
-				// The second part indicates the branch's location in the source file.
+				// The second part is the branch's id.
 				// This format must match what is in `cgen`.
-				idx_str := comptime_branch_context_str + '|${c.file.path}|${branch.cond.pos()}|'
+				if branch.id == 0 {
+					// this is a new branch, alloc a new id for it
+					c.cur_ct_id++
+					branch.id = c.cur_ct_id
+				}
```
spytheman, line 116: extract a method

PR #25313, opened 2025-09-15
js: simplify comptime if

kbkpbot:
Fix issue #25295

And now js backend support comptime if-expr:

```v
const max_int = $if amd64 { int(max_i64) } $else { int(max_i32) }
```

And remove unused code from comptime if in js backend.


vlib/builtin/js/int.js.v
```diff
@@ -17,8 +17,16 @@ pub const max_i32 = i32(2147483647)
...
+pub const min_int = $if new_int ? && (arm64 || amd64 || rv64 || s390x || ppc64le || loongarch64) {
+	int(min_i64)
+} $else {
+	int(min_i32)
+}
+pub const max_int = $if new_int ? && (arm64 || amd64 || rv64 || s390x || ppc64le || loongarch64) {
+	int(max_i64)
+} $else {
+	int(max_i32)
+}
```
JalonSolov, line 29: Why are these duplicated in the js builtins?  Isn't there common builtin code shared across the systems?  If not, there definitely should be...

kbkpbot: If comment out these lines, js can't find `max_int`, seems js backend has it's own `int.js.v` ? not sure about this.

```
$ v -b js vlib/builtin/js/int_test.js.v
vlib/builtin/js/int_test.js.v:249:7: error: undefined ident: `max_int`
  247 |
  248 | fn test_max_int() {
  249 |     dump(max_int)
      |          ~~~~~~~
  250 | }
  251 |

```

JalonSolov: Yes, that appears to be the case now, and you shouldn't change anything else for this PR (meaning leave what you have here).

However, it is something that we definitely need to look into correcting in the future.  All backends should share as much code as possible, so we don't have to make individual duplicate changes like this.

PR #25315, opened 2025-09-15
cgen,autofree: fix method chaining generating invalid c code

### LINKED ISSUE #9094: String corruption when using autofree
**V version:** V 0.2.2 822d20d
**OS:** macOS 10.15.7

**What did you do?**

```v
$ cat bad.v
pub struct Test {
mut:
	tags        []string
}
pub fn (mut t Test) set_tags(tags string) {
	t.tags = tags.to_lower().replace(',', ' ').trim_space().split(' ')
}

fn main() {
	mut test := Test{}
	test.set_tags('')
}
```

and then bad things happen:

```
$ v -autofree /tmp/bad.v
==================
                               ^
/var/folders/12/517hk27n0pxbnbv9yxlh4jxm0000gn/T/v/bad.6528083534070178736.tmp.c:9446:35: error: use of undeclared identifier 'assignment2'
string _arg_expr_split_0_148 = re-assignment2
                                  ^
2 errors generated.
...
==================

builder error:
==================

If you were not working with C interop, this is a compiler bug, please raise an issue on GitHub:

```

A working compiled program with valid C code, and not a half overwritten string in the middle of the generated C code.

By searching the 're-assignment' string in the v compiler code its clear that the resulting C code is not possible to be generated, so the bug seems to be related to how strings are managed in V, and specially when using the autofree mode.

Building V with ASAN doesnt spot any issue either. So its not a classic memory corruption bug but probably related to overwritten stack variables
 
**What did you see instead?**

invalid C code

dy-tea:
Fixes #9094.
The generated C code still looks a bit wonky but it does work.


dy-tea:
Internal server error, haven't seen that one before.

JalonSolov:
Just your normal flaky internet connections.  :-\

I restarted the job.

vlib/v/gen/c/fn.v
```diff
@@ -2345,7 +2345,20 @@ fn (mut g Gen) autofree_call_pregen(node ast.CallExpr) {
 			})
 			s = 'string ${t} = '
 		}
-		s += g.expr_string(arg.expr)
+		g.is_autofree_tmp = true
```
spytheman, line 2348: You may need to save/restore the previous state of `g.is_autofree_tmp` too.
`g.expr()` can recurse through here and mess it up otherwise.

spytheman (approved):
Excellent work.
Thank you.

Issue #25272, opened 2025-09-09
time format rfc3339 parsing errors


jorgeluismireles:
### Describe the bug

Check some strings in `time.parse_rfc3339` function. `t` and `z` should be valid as `T` and `Z`. Space ` ` could replace `T` and some other details.

### Reproduction Steps

```v
import time

const times = [
	'1979-05-27T07:32:00Z' // 0
	'1979-05-27T07:32:00-01:00' // 1
	'1979-05-27T07:32:00.999999-01:00' // 2

	'1979-05-27t07:32:00z' // 3
	
	'1979-05-27 07:32:00Z' // 4
	'1979-05-27X07:32:00Z' // 5
	
	'1979-05-27T07:32:00' // 6
	'1979-05-27' // 7
	'07:32:00'   // 8
	
	'1979-05-27T00:32:00.999999' // 9
]

fn main() {
	for pos, s in times {
		t := time.parse_rfc3339(s) or {
			println('${pos:2} ${err}')
			continue
		}
		println('${pos:2} ${t}')
	}
}
```

### Expected Behavior

- All valid strings should be converted to complete Time object
- All invalid strings should print messages without panic.

### Current Behavior

Program above results:
```
 0 1979-05-27 07:32:00
 1 1979-05-27 08:32:00
 2 1979-05-27 08:32:00
 3 1979-05-27 00:00:00
 4 1979-05-27 00:00:00
 5 1979-05-27 00:00:00
 6 timezone error: datetime string is too short
 7 1979-05-27 00:00:00
 8 missing date part of RFC 3339
V panic: string index out of range(idx,s.len):26, 26
v hash: bae7684
/tmp/v_1000/250909-a.01K4R6Y96JSE57TJX58S16HT84.tmp.c:4048: at _v_panic: Backtrace
/tmp/v_1000/250909-a.01K4R6Y96JSE57TJX58S16HT84.tmp.c:4079: by panic_n2
/tmp/v_1000/250909-a.01K4R6Y96JSE57TJX58S16HT84.tmp.c:5816: by string_at
/tmp/v_1000/250909-a.01K4R6Y96JSE57TJX58S16HT84.tmp.c:6880: by time__parse_rfc3339
/tmp/v_1000/250909-a.01K4R6Y96JSE57TJX58S16HT84.tmp.c:7124: by main__main
/tmp/v_1000/250909-a.01K4R6Y96JSE57TJX58S16HT84.tmp.c:7174: by main
```
Issues
| cases |  | |
[...]

PR #25316, opened 2025-09-16
time: make parse_rfc3339 accept `t`, and space, in addition to `T`, reject incomplete date-times and such with no timezone

jorgeluismireles:
Fixes issue https://github.com/vlang/v/issues/25272 for parsing dates in format rfc3339.

- Accepts `t` and ` ` (space) as alternatives to default date time separator `T`.
- Continue accepting `z` as alternative of default `Z` for timezone zero.
- Reject (as golang does) not complete date-time strings or without timezone.
- Fix panic in fractions of second without `Z`.

Revised function `time.parse_rfc3339(string) !time.Time` can be used as simple and unique format for `json2.Any.to_time()` of type `string`.

### DIFF
--- a/vlib/time/parse.c.v
+++ b/vlib/time/parse.c.v
@@ -160,7 +160,7 @@ pub fn parse_rfc3339(s string) !Time {
 		year, month, day = check_and_extract_date(s)!
 	}
 	if s.len <= date_format_buffer.len {
-		return error('date too short to parse')
+		return error('date-time too short to parse')
 	}
 	if s[10] !in [u8(`T`), `t`, ` `] {
 		return error('invalid date-time separator:${s[10].ascii_str()}')
--- a/vlib/time/parse.c.v
+++ b/vlib/time/parse.c.v
@@ -158,18 +158,16 @@ pub fn parse_rfc3339(s string) !Time {
 
 	if is_date {
 		year, month, day = check_and_extract_date(s)!
-		if s.len == date_format_buffer.len {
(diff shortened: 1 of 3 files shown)


vlib/time/parse.c.v
```diff
@@ -158,18 +158,16 @@ pub fn parse_rfc3339(s string) !Time {
...
-			return new(Time{
-				year:     year
-				month:    month
-				day:      day
-				is_local: false
-			})
-		}
+	}
+	if s.len <= date_format_buffer.len {
+		return error('date too short to parse')
```
spytheman:
```suggestion
		return error('date-time too short to parse')
```
otherwise you will get:
```
assert invalid_rfc3339('1979-05-27') == 'date too short to parse'
```

which is confusing, since `1979-05-27` is a valid date, it is not too short, it is just that you expect also the T and the time part too now

vlib/time/parse.c.v
```diff
@@ -189,7 +187,7 @@ pub fn parse_rfc3339(s string) !Time {
 				timezone_start_position++
 				if timezone_start_position == s.len {
 					return error('timezone error: expected "Z" or "z" or "+" or "-" in position ${timezone_start_position}, not "${[
-						s[timezone_start_position],
+						s[s.len - 1],
```
spytheman, line 190: good find

spytheman:
Excellent work.
Thank you @jorgeluismireles 🙇🏻.

PR #25317, opened 2025-09-16
checker: allow int i32 to be used interchangeably

kbkpbot:
This is 3rd PR to support 64bit/32bit int as PR https://github.com/vlang/v/pull/25236 is the first one and #25298 is the second one.

Before apply `v fmt -new_int -w .` to the whole v codebase, it is need to ease some type check first.
This PR allow `i32` <=> `bool` and also `int` <=> `i32` type check.

### DIFF
--- a/vlib/v/checker/check_types.v
+++ b/vlib/v/checker/check_types.v
@@ -19,6 +19,15 @@ fn (mut c Checker) check_types(got ast.Type, expected ast.Type) bool {
 	got_is_ptr := got.is_ptr()
 	exp_is_ptr := expected.is_ptr()
 
+	$if !new_int ? {
+		// allow int & i32 to be used interchangeably for C functions
+		// mut flags := C.fcntl(handle, C.F_GETFL, 0)
+		// flags &= ~C.O_NONBLOCK
+		if (got in [ast.int_type_idx, ast.int_literal_type_idx] && expected == ast.i32_type_idx)
+			|| (expected in [ast.int_type_idx, ast.int_literal_type_idx] && got == ast.i32_type_idx) {
+			return true
+		}
+	}
 	// allow int literals where any kind of real integers are expected:
 	if (got == ast.int_literal_type && expected.is_pure_int())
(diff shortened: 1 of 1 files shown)


spytheman:
Why is `i32 <=> bool` needed?

JalonSolov:
In a lot of C compilers `bool` is `typedef int bool`.  On a 32-bit system, an `int` is `i32`, so `bool` will also be `i32`, and I've seen code where it was simply treated as an `int` rather than a separate type.

It used to be quite common to see
```c
typedef int bool;
#define true 1
#define false 0
```
Then your code do
```c
if (mybool == true) // or more often, `if (mybool)`
{
 ...
}
```
or whatever, which is all `i32`s.

JalonSolov:
That would be the better thing all around, really, as you cannot guarantee that `int` == `i32` in all C compilers.  The only guarantee on sizes is the `sizeof(char) == 1`.  Everything else is "implementation defined".

spytheman:
Please do focus on the current PR, and avoid discussing things unrelated to it here.

There is https://github.com/vlang/v/discussions and Discord for chatting/discussing other topics.

spytheman:
> 		// unknown C types are set to int, allow int to be used for types like `&C.FILE`
>		// eg. `C.fflush(C.stderr)` - error: cannot use `int` as `&C.FILE` in argument 1 to `C.fflush`

This should now just apply to `i32` .
The C side is not going to change, just because V decided to be more like Go on a whim.

spytheman:
> // allow bool & int to be used interchangeably for C functions

Similarly, this should apply to `i32` now, and only for C functions, to allow for the same easy interoperability as before.

On the V side, now `bool` != `int`, and in the future it will be still != `int` *and* != `i32`.

spytheman:
> Or should we introduce something redecalre the C.xx's type in V?
> This will help integrate C code.

That can be a separate proposal/PR, for C globals/consts .

spytheman (approved):
Excellent work.
Thank you very much @kbkpbot 🙇🏻.

PR #25321, opened 2025-09-16
x.crypto.chacha20: adds some missing bits into stream reset

blackshirt:
This patch some missing bits of resetting fields into Stream.reset

### DIFF
--- a/vlib/x/crypto/chacha20/stream.v
+++ b/vlib/x/crypto/chacha20/stream.v
@@ -9,7 +9,7 @@ import encoding.binary
 
 // max_64bit_counter is a 64-bit maximum internal counter of original ChaCha20 variant.
 const max_64bit_counter = max_u64
-// max_64bit_counter is a 32-bit maximum internal counter of standard IETF ChaCha20 variant.
+// max_32bit_counter is a 32-bit maximum internal counter of standard IETF ChaCha20 variant.
 const max_32bit_counter = u64(max_u32)
 
 // default chacha20 quarter round number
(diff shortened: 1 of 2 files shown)


vlib/x/crypto/chacha20/stream.v
```diff
@@ -107,11 +107,16 @@ fn new_stream(key []u8, nonce []u8) !Stream {
...
+	s.mode = .standard
 	s.extended = false
 	unsafe {
 		_ := vmemset(&s.key, 0, 32)
 		_ := vmemset(&s.nonce, 0, 16)
 	}
+	s.precomp = false
+	s.p1, s.p5, s.p9, s.p13 = u32(0), u32(0), u32(0), u32(0)
+	s.p2, s.p6, s.p10, s.p14 = u32(0), u32(0), u32(0), u32(0)
+	s.p3, s.p7, s.p11, s.p15 = u32(0), u32(0), u32(0), u32(0)
```
spytheman, line 119: what about p4, p8, p12 ?

edit: there are no such fields

blackshirt: Literally speaking, its a small optimization technique adapted from the golang version, its for standard variant of the stream. It for storing  the counter-independent results of the first round after they are computed the first time. 
p0, p4, p8, and p12 was on the same column on the initialization state (matrix) where the counter lives on s[12] of the state s.
Technically, it can be adapted to 64-bit counter too with 2 column of independent of counter.
I;m not truely happy on readibility of this section, its hard to grasp or to understand by other people who want to learn.
This technique also can be completely removed and do full quarter round instead.

PR #25322, opened 2025-09-16
v: implement comptime regular func params traversing

felipensp:
```v
fn demo(a int, b string) {
}

fn g [T](cb T) {
	mut params := []FunctionParam{}
     $if T is $function {
        $for param in T.params {
			params << param
        }
    }
	assert params.len == 2
	assert params[0].name == 'a'
	assert params[0].typ == 8
	assert params[1].name == 'b'
	assert params[1].typ == 21
}

fn main() {
    g(demo)

	mut params := []FunctionParam{}
	$for param in demo.params {
		params << param
	}
	assert params.len == 2
	assert params[0].name == 'a'
	assert params[0].typ == 8
	assert params[1].name == 'b'
	assert params[1].typ == 21
}
```

### DIFF
--- a/vlib/builtin/builtin.v
+++ b/vlib/builtin/builtin.v
@@ -88,8 +88,8 @@ fn __print_assert_failure(i &VAssertMetaInfo) {
 	}
 }
 
-// MethodParam holds type information for function and/or method arguments.
-pub struct MethodParam {
+// FunctionParam holds type information for function and/or method arguments.
+pub struct FunctionParam {
 pub:
 	typ  int
 	name string
@@ -100,7 +100,7 @@ pub struct FunctionData {
 pub:
 	name        string
 	attrs       []string
-	args        []MethodParam
+	args        []FunctionParam
 	return_type int
 	typ         int
 }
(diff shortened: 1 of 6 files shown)


spytheman (approved):
Excellent work.

vlib/builtin/builtin.v
```diff
@@ -88,8 +88,8 @@ fn __print_assert_failure(i &VAssertMetaInfo) {
 	}
 }
 
-// MethodParam holds type information for function and/or method arguments.
-pub struct MethodParam {
+// FunctionParam holds type information for function and/or method arguments.
+pub struct FunctionParam {
```
spytheman, line 92: That rename can be a bit breaking, but afaik the name is not used much outside of the V repo, and even here, only indirectly 🤔 .

felipensp: maybe should it be `VFunctionParam` instead ?

Delta456 (approved):

vlib/builtin/builtin.v
spytheman: I think it is fine as you named it.

I've searched `builtin` for other similar structs, and it seems they do not follow a convention - some use `V` as a prefix, some do not.

At a future time, we should probably do a cleanup and use a common naming scheme, but that is out of scope for this PR.

PR #25323, opened 2025-09-16
x.crypto.chacha20: removes some ridiculous and bogus check on set counter

blackshirt:
This patch removes ridiculous and bogus check on `Strem.set_counter`. See the issues at [here](https://discord.com/channels/592103645835821068/592320321995014154/1417519455193399337). This patch also reorganizes `Stream.keystream_with_blocksize` to be on par with go one


blackshirt:
I think there are something bad with two revert commits between the pr , it should not be there ...

PR #25324, opened 2025-09-17
vlib,tests: fix int i32 for 64bit int

kbkpbot:
This PR fix `vlib` for future apply 64bit int.
As `C.func` will only return `i32` not `int` any more, to avoid the error message:
```
error: array len needs to be an int
```
And also fix `vlib/v/gen/c/testdata/translated/` files to workaround `int` in source code and generated code.


spytheman:
imho change the check to:
```v
    if sym.kind !in [.int, .i32, .int_literal] {
        c.error('array ${para} needs to be an int', pos)
    }
```

sizeof(i32) is always <= sizeof(int) even when `int` becomes 64 bit, so passing an `i32` value to that field should be allowed

spytheman (approved):
Excellent work.
Thank you @kbkpbot 🙇🏻‍♂️.

PR #25336, opened 2025-09-17
autofree,flag: fix double free in FlagParser free

### LINKED ISSUE #9815: Compiling string flag with -autofree ends in heap-corruption
**V version:** 0.2.2 4a1f75c
**OS:** all

**What did you do?**
```v
import flag
import os

fn main() {
	mut fp := flag.new_flag_parser(os.args)
	test := fp.string('test', 0, '', 'this is a test')

	_ := fp.finalize() or {
		println(fp.usage())
		return
	}
	println(test)
}
```
Built it with: `v -autofree -prod -cc [gcc, msvc, tcc] test.v`
Executed with: `test.exe --test sometext`

sometext
 
**What did you see instead?**
some garbage and after repeating several times (on Windows): 
```
Unhandled Exception 0xC0000374
11: RtlIsZeroMemory            ?? : address = 0x61c4dce108
10: RtlpNtSetValueKey          ?? : address = 0x61c4dce108
9 : RtlpNtSetValueKey          ?? : address = 0x61c4dce108
8 : RtlpNtSetValueKey          ?? : address = 0x61c4dce108
7 : RtlGetCurrentServiceSessionId  ?? : address = 0x61c4dce108
6 : RtlFreeHeap                ?? : address = 0x61c4dce108
5 : free_base                  ?? : address = 0x61c4dce108
4 : flag__FlagParser_usage     ?? : address = 0x61c4dce108
3 : wmain                      ?? : address = 0x61c4dce108
2 : __scrt_common_main_seh     d:\agent\_work\63\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl:288
1 : BaseThreadInitThunk        ?? : address = 0x61c4dce108
0 : RtlUserThreadStart         ?? : address = 0x61c4dce108
```

I think the error is related to os.args and -autofree

dy-tea:
Fixes #9815.

It seems to me like calling these frees on `f.args` frees `os.args` only on `-autofree`? Not sure why that would happen since `os.args` is always cloned.

I set it to `@[manualfree]` here but it might actually end up leaking some memory since the `new_flag_parser` function creates some clones, but it was preferable to freeing the variables before (and after) returning them?

A few flag tests still fail with autofree but not in a way related to this issue.

### DIFF
--- a/vlib/flag/flag.v
+++ b/vlib/flag/flag.v
@@ -107,10 +107,6 @@ pub mut:
 @[unsafe]
 fn (mut f FlagParser) free() {
 	unsafe {
-		for a in f.args {
-			a.free()
-		}
-		f.args.free()
 		//
 		for flag in f.flags {
 			flag.free()
@@ -130,6 +126,7 @@ pub const underline = '-----------------------------------------------'
 pub const max_args_number = 4048
 
 // new_flag_parser - create a new flag parser for the given args.
+@[manualfree]
 pub fn new_flag_parser(args []string) &FlagParser {
 	original_args := args.clone()
 	idx_dashdash := args.index('--')
--- a/vlib/flag/flag_autofree_test.v
+++ b/vlib/flag/flag_autofree_test.v
@@ -1,0 +1,10 @@
+// vtest vflags: -autofree
+import flag
+import os
+
+fn test_main() {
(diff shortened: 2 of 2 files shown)


spytheman (approved):
Good work, however I think that the problem, may be in the `args.clone()` call, for []string , which does not clone the individual elements/strings of the array.

Issue #25333, opened 2025-09-17
Multi-dimensional array type error


kbkpbot:
### Describe the bug

When declare a multi-dimensional array with wrong syntax, V compiler does not issue a error message.

### Reproduction Steps

y.v
```v
module main

fn main() {
        a2 := [][][]f32[][]
        dump(typeof(a2).name)
}
```

### Expected Behavior

Should issue a error message about the wrong syntax of array declare, or output a correct array type.

### Current Behavior

```
$ v run y.v
y.v:4:8: warning: use `x := []Type{}` instead of `x := []Type`
    2 |
    3 | fn main() {
    4 |     a2 := [][][]f32[][]
      |           ~~~~~~~~~~~~~
    5 |     dump(typeof(a2).name)
    6 | }
[y.v:5] typeof(a2).name: [][][][][]f32
```

Note that the type name of `a2` is wrong.

### Possible Solution

Fix parser?

### V version

V 0.4.11 14ef7651511b33887c9a2d6d042314b401160897

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 14ef7651511b33887c9a2d6d042314b401160897
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |9.5GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-09-17 14:07:53
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/arr
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.36-84-g14ef7651-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
[...]

PR #25338, opened 2025-09-17
parser: detect wrong multi-dim arr declare

kbkpbot:
Fix issue #25333
Terminate the array type scan after parser got an element type, do not scan further `[]` pairs.

I think this is a code snippet left behind from the early stages for detecting multidimensional arrays with legacy syntax.


spytheman:
> I think this is a code snippet left behind from the early stages for detecting multidimensional arrays with legacy syntax.

Yes, it is done for compatibility with syntax from before 3ab8dc0 2020-04-26 .
I think at this point it can safely become an error, not just a warning.

PR #25341, opened 2025-09-18
gg: allow mut Rect, pub image_cache

Eliyaan:
implements #25332

### DIFF
--- a/vlib/gg/gg.c.v
+++ b/vlib/gg/gg.c.v
@@ -184,12 +184,12 @@ mut:
 	render_text bool = true
 	// a cache with all images created by the user. used for sokol image init and to save space
 	// (so that the user can store image ids, not entire Image objects)
-	image_cache   []Image
 	needs_refresh bool = true
 	ticks         int // for ui mode only
 pub:
 	native_rendering bool
 pub mut:
+	image_cache   []Image
 	scale       f32 = 1.0 // will get set to 2.0 for retina, will remain 1.0 for normal
 	width       int
 	height      int
--- a/vlib/gg/image.v
+++ b/vlib/gg/image.v
@@ -27,7 +27,7 @@ pub enum ImageEffect {
 
 // Rect represents a rectangular shape in `gg`.
 pub struct Rect {
-pub:
+pub mut:
 	x      f32
 	y      f32
 	width  f32


vlib/gg/gg.c.v
```diff
@@ -184,12 +184,12 @@ mut:
 	render_text bool = true
 	// a cache with all images created by the user. used for sokol image init and to save space
 	// (so that the user can store image ids, not entire Image objects)
-	image_cache   []Image
 	needs_refresh bool = true
 	ticks         int // for ui mode only
 pub:
 	native_rendering bool
 pub mut:
+	image_cache []Image
```
spytheman: I am not sure that exposing the cache to users is a good idea, to me it seems like an implementation detail, and handling it can be cumbersome.

Are you sure it will be useful? What use cases do you have in mind, that require user apps to control the cache directly?

vlib/gg/image.v
```diff
@@ -27,7 +27,7 @@ pub enum ImageEffect {
 
 // Rect represents a rectangular shape in `gg`.
 pub struct Rect {
-pub:
+pub mut:
```
spytheman, line 30: That is useful indeed.
If the PR contained only this, I would have merged it right away.

vlib/gg/gg.c.v
Eliyaan: Right now, the user cannot handle images, so if they want to create an image with `gg.create_image` and then use the resulting image in a custom draw image function (instead of using the default `gg.draw_image_with_config`) to have better performance while still using gg for example they cannot:
The image the user gets from `gg.create_image` is a "false" one, only the id it contains is used (the get the actual image in the `ctx.image_cache`), and so the user is stuck with `gg.draw_image_with_config` and needs to move away completly from `gg` if the actual system is not performant enough/suited to their needs. (I'm not in this case but I imagine that if they want to draw an image in a funny way (for example: like a trapezoid instead of a square), they would need to move away from gg as they cannot access the image cache)

spytheman: How would exposing the cache field help in that situation?
Note, that there are already these APIs:
`pub fn (mut ctx Context) cache_image(img Image) int {`
`pub fn (mut ctx Context) get_cached_image_by_idx(image_idx int) &Image {`
`pub fn (mut ctx Context) remove_cached_image_by_idx(image_idx int) {`

spytheman: > The image the user gets from gg.create_image is a "false" one

That is not the case.
There are 2 cases - images loaded before sokol's initialization and those loaded after that.
The cache exists in part to allow for deferred loading, which may be the reason you think that the returned image is "false".

Eliyaan: > How would exposing the cache field help in that situation? Note, that there are already these APIs: `pub fn (mut ctx Context) cache_image(img Image) int {` `pub fn (mut ctx Context) get_cached_image_by_idx(image_idx int) &Image {` `pub fn (mut ctx Context) remove_cached_image_by_idx(image_idx int) {`

Sorry, I should have checked this. This should work well enough. I'll fix the PR with only mut Rect if it works.

Eliyaan: > > The image the user gets from gg.create_image is a "false" one
> 
> That is not the case. There are 2 cases - images loaded before sokol's initialization and those loaded after that. The cache exists in part to allow for deferred loading, which may be the reason you think that the returned image is "false".

Oh okay, I assumed that `Image.simg_ok` was used to differentiate image from the cache and image given to the user. Thanks

Eliyaan:
I've reverted the image_cache modification

PR #25343, opened 2025-09-18
v: cleanup, replace complex 64bit CPU comptime ident with `&& x64`

kbkpbot:
Replace `(arm64 || amd64 || rv64 || s390x || ppc64le || loongarch64)` with `x64`.

In `cgen`, `x64` comptime ident will generate following code:
```c
#if INTPTR_MAX == INT32_MAX
        #define TARGET_IS_32BIT 1
#elif INTPTR_MAX == INT64_MAX
        #define TARGET_IS_64BIT 1
#else
        #error "The environment is not 32 or 64-bit."
#endif

...

        #if defined(TARGET_IS_64BIT)
        {
                println(_S("x64"));
        }
        #endif
```

This should be a better way detect the compile environment  is 64-bit or not.

And this will help compile 32bit `int` on an 64bit environment.

And also, it is much shorter. :)

### DIFF
--- a/vlib/builtin/int.v
+++ b/vlib/builtin/int.v
@@ -48,16 +48,8 @@ pub const max_i32 = i32(2147483647)
 pub const min_i64 = i64(-9223372036854775807 - 1)
 pub const max_i64 = i64(9223372036854775807)
 
-pub const min_int = $if new_int ? && (arm64 || amd64 || rv64 || s390x || ppc64le || loongarch64) {
-	int(min_i64)
-} $else {
-	int(min_i32)
-}
-pub const max_int = $if new_int ? && (arm64 || amd64 || rv64 || s390x || ppc64le || loongarch64) {
-	int(max_i64)
-} $else {
-	int(max_i32)
-}
+pub const min_int = $if new_int ? && x64 { int(min_i64) } $else { int(min_i32) }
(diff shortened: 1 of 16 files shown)


JalonSolov:
Is `INTPTR_MAX` defined for native, WASM, Go, JS, etc. backends?

spytheman:
The `vlib/flag/flag_autofree_test.v` sanitized job failures are not related, and are already skipped on master.

spytheman:
> Is INTPTR_MAX defined for native, WASM, Go, JS, etc. backends?

That is not related to the PR. It affects only the C backend.
`INTPTR_MAX` is a C macro.

spytheman (approved):
Excellent work.
It is my bad, sorry - I should have spotted it in the review before.
Thanks for the cleanup @kbkpbot .

JalonSolov:
Yes, I know what `INTPTR_MAX` is.  My point is that we need something that can be used by all backends, especially native.

spytheman:
`INTPTR_MAX` is not used in the V code.

The expression on the V side is `$if ... && x64 { ... } $else { ... }` .

spytheman:
i.e. each backend is free to implement it like any other comptime expression, it will not have to deal with `INTPTR_MAX` etc, which is used only by the code generated by cgen.

PR #25349, opened 2025-09-19
gg: fix color use i32 type instead of int type

kbkpbot:
As `int` will be a 64bit/32bit depend on OS/Compiler, replace `int` in `gg.color` with `i32`.

### DIFF
--- a/vlib/builtin/js/string.js.v
+++ b/vlib/builtin/js/string.js.v
@@ -287,6 +287,11 @@ pub fn (s string) i16() i16 {
 	return i16(JS.parseInt(s.str))
 }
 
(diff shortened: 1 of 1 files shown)


vlib/gg/color.v
```diff
@@ -298,7 +298,7 @@ const string_colors = {
 pub fn color_from_string(s string) Color {
 	if s.starts_with('#') {
 		mut hex_str := '0x' + s[1..]
-		return hex(hex_str.int())
+		return hex(i32(hex_str.int()))
```
spytheman:
```suggestion
		return hex(hex_str.i32())
```

spytheman (approved):
Good work.

Delta456 (approved):

PR #25351, opened 2025-09-19
vlib: fix new_int for vlib

kbkpbot:
Fix `new_int` for `vlib` working.


vlib/builtin/int_test.v
```diff
@@ -17,7 +17,11 @@ fn test_str_methods() {
 	assert int(1).str() == '1'
 	assert int(-1).str() == '-1'
 	assert int(2147483647).str() == '2147483647'
-	assert int(u32(2147483648)).str() == '-2147483648'
+	assert int(u32(2147483648)).str() == $if new_int ? && x64 {
+		'2147483648'
+	} $else {
+		'-2147483648'
+	}
```
spytheman, line 24: I think, it would have been clearer, if it was:

```suggestion
	$if new_int ? && x64 {
		assert int(u32(2147483648)).str() == '2147483648'
	} $else {
		assert int(u32(2147483648)).str() == '-2147483648'
	}
```
because then the potentially failing assertion will be simpler and easier to diagnose from just the output.

For comparison:

Image

spytheman: As you can see, in the second version, you can immediately see that the (potentially) failing assertion is the one on line 23, and the expected value right away, while in the first case, you use the full condition + the comptime logic and have to interpret it, before understanding which case it was, and what was the actual expected value.

vlib/v/tests/assign/assign_array_fixed_from_union_test.v
```diff
@@ -13,9 +13,17 @@ fn test_main() {
 	assert a == [1, 2, 3, 4, 5]!
 	assert p != 0
 	$if little_endian {
-		assert c == [u8(1), 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0, 4, 0, 0, 0, 5, 0, 0, 0]!
+		$if new_int ? && x64 {
+			assert c == [u8(1), 0, 0, 0, 0, 0, 0, 0, 2, 0, 0, 0, 0, 0, 0, 0, 3, 0, 0, 0]!
+		} $else {
+			assert c == [u8(1), 0, 0, 0, 2, 0, 0, 0, 3, 0, 0, 0, 4, 0, 0, 0, 5, 0, 0, 0]!
+		}
```
spytheman, line 20: nice, this is much better!

vlib/v/tests/comptime/comptime_aggregate_var_test.v
```diff
@@ -17,9 +17,17 @@ fn d(val Value) string {
...
+	}
 	assert d(Value(i64(1))) == 'Value is number or byte array, size=8 1'
 	assert d(Value(u64(2))) == 'Value is number or byte array, size=8 2'
-	assert d(Value([u8(1), 2])) == 'Value is number or byte array, size=32 [1, 2]'
+	assert d(Value([u8(1), 2])) == $if new_int ? && x64 {
+		'Value is number or byte array, size=48 [1, 2]'
+	} $else {
+		'Value is number or byte array, size=32 [1, 2]'
+	}
 	assert d(Value('')) == 'Value is string: '
```
spytheman, line 32: imho just change `int` to `i32` in the sumtype here, because that way the diff will be smaller, and the test is otherwise unrelated to the `int` -> isize change.

vlib/v/tests/fns/fixed_array_on_voidptr_test.v
```diff
@@ -2,7 +2,7 @@ fn test_main() {
 	mut b := [2]int{}
 	b[0] = 1
 	b[1] = 2
-	mut a := unsafe { memdup(b, 8) }
+	mut a := unsafe { memdup(b, $if new_int ? && x64 { 16 } $else { 8 }) }
```
spytheman, line 5: same, imho just change `int` to `i32` here - the test does not need to be concerned with the new_int change

vlib/v/tests/generics/generics_fn_variable_2_test.v
```diff
@@ -10,12 +10,20 @@ fn f[T]() u32 {
 fn test_generic_fn_variable() {
 	r1 := f[int]()
 	println(r1)
-	assert r1 == 4
```
spytheman, line 13: same, changing `int` to `i32` will produce a simpler diff, that will not have to be edited out again, when `new_int` becomes the default, and the test is not directly related to the sizes of the integers; they are used here just as a proxy that the comptime generic worked

vlib/v/tests/structs/struct_generic_sizeof_test.v
```diff
@@ -8,10 +8,17 @@ mut:
 
 fn test_main() {
 	x := PaddedSlot[int]{}
-	assert '${x}' == 'PaddedSlot[int]{
+	$if new_int ? && x64 {
+		assert '${x}' == 'PaddedSlot[int]{
```
spytheman, line 12: excellent

vlib/v/tests/c_array_test.c.v
```diff
@@ -17,7 +17,7 @@ fn test_carray_to_varray() {
 	}
 
 	c_int_array := C.gen_c_int_array(size)
-	v_int_array := unsafe { arrays.carray_to_varray[int](c_int_array, size) }
+	v_int_array := unsafe { arrays.carray_to_varray[i32](c_int_array, size) }
```
spytheman, line 20: excellent

spytheman (approved):
Very good work.

PR #25352, opened 2025-09-19
gx: fix int to i32

kbkpbot:
Additional fix for PR #25349


spytheman (commented):
`gx` is deprecated

JalonSolov:
... but not gone, yet.

Issue #25347, opened 2025-09-18
json2: examples\news_fetcher.v: `decoder2.decode` fails when is called within a `sync.pool` callback


jorgeluismireles:
<details>
<summary>V version: V 0.4.11 b93dfa1, press to see full `v doctor` output</summary>

|V full version      |V 0.4.11 e96355aca59809ed3f642e82d3136a1f1b3fd4c9.b93dfa1
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 22.04.5 LTS
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz
|Memory              |0.51GB/3.7GB
|                    |
|V executable        |/home/jorge/v/v
|V last modified time|2025-09-18 22:13:56
|                    |
|V home dir          |OK, value: /home/jorge/v
|VMODULES            |OK, value: /home/jorge/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/jorge/bugs
|                    |
|Git version         |git version 2.34.1
|V git status        |weekly.2025.16-797-gb93dfa1a-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
|gcc version         |gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
|clang version       |Ubuntu clang version 14.0.0-1ubuntu1.1
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.5 ()
|glibc version       |ldd (Ubuntu GLIBC 2.35-0ubuntu3.10) 2.35

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  news_fetcher.v && news_fetcher`

Was playing with [examples/news_fetcher.v](https://github.com/vlang/v/blob/master/examples/news_fetcher.v) and make changes to incorporate `json2.decoder2` (temporarily removed license and not related stuff):

```v
import net.http
import x.json2
import x.json2.decoder2
import sync.pool

struct Story {
	title string
	url   string
}

fn worker_fetch(mut p pool.PoolProcessor, cursor int, worker_id int) voidptr {
[...]

PR #25353, opened 2025-09-19
json2: make examples/news_fetcher.v to work with decoder2

jorgeluismireles:
Fixes issue https://github.com/vlang/v/issues/25347
- `decode2.v` seems to had a typo I don't understand totally but seems fixes the issue.
- examples/news_fetcher.v working with decode2.


spytheman (approved):
Thank you 🙇🏻 .

PR #25355, opened 2025-09-20
fix: print block the program when no stdout attached.

get200:
When compiled as a non-console program, such as a GUI application, using the print function will cause the program to hang because there is no stdout.

```v
// eg
#flag -mwindows

```

### DIFF
--- a/vlib/v/gen/c/cmain.v
+++ b/vlib/v/gen/c/cmain.v
@@ -133,6 +133,12 @@ fn (mut g Gen) gen_c_main_function_only_header() {
 
 fn (mut g Gen) gen_c_main_function_header() {
 	g.gen_c_main_function_only_header()
+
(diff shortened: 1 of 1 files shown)


spytheman:
Can you please add a test, and if the problem is platform specific, also check for the platform?

get200:
> Can you please add a test, and if the problem is platform specific, also check for the platform?

I don’t know how to add GUI test code, because running it in the console works fine, but it doesn’t work when double-clicked in Explorer, for example, in the following code.

```v
$if msvc {
	#flag -lgdi32
} $else $if gcc {
	#flag -mwindows
}

fn main() {
	println('print unblock!')
}
```

spytheman:
At the very least, add it inside `gen_c_main_function_only_header`, in the part, that checks for gui programs on windows. 

As it is, it is not testable, and it breaks V programs everywhere else :-| .

PR #25358, opened 2025-09-20
comptime: fix $compile_error on $match when used on return stmt

### LINKED ISSUE #25348: `$compile_error()` always triggers when using compile time `$match` to pass a value
### Describe the bug

When using `$compile_error()` in a compile time `$match` statement to pass/provide a variable through the `$match` block rather than setting it inside the `$match` block `$compile_error()` will always trigger even when it should not.

This is probably related to #25223

### Reproduction Steps

```v
module main

fn test[T]() bool {
	return $match T {
		i8  { true }
		i16 { true }
		$else { $compile_error('unsupported type') }
	}
}

fn main() {
	println(test[i8]())
}
```

`v run file.v`

### Expected Behavior

The program to successfully compile and run.

### Current Behavior

```v
file.v:7:11: error: unsupported type
    5 |         i8  { true }
    6 |         i16 { true }
    7 |         $else { $compile_error('unsupported type') }
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    8 |     }
    9 | }
file.v:7:11: error: return type mismatch, it should be `bool`, but it is instead `void`
    5 |         i8  { true }
    6 |         i16 { true }
    7 |         $else { $compile_error('unsupported type') }
      |                 ^
    8 |     }
    9 | }
```

### Possible Solution

The fix will probably be similar to the one for #25223 in pull https://github.com/vlang/v/pull/25225

The following arrangement causes the issue to not occur as this arrangement has been patched by pull https://github.com/vlang/v/pull/25225

```v
fn test[T]() bool {
	$match T {
		i8  { return true }
		i16 { return true }
		$else { $compile_error('unsupported type') }
	}
}
```

### V version

V 0.4.11 b93dfa1

macOS Sequoia 15.6

> [!NOTE]
>

felipensp:
Fix #25348

### DIFF
--- a/vlib/v/gen/c/comptime.v
+++ b/vlib/v/gen/c/comptime.v
@@ -1004,49 +1004,57 @@ fn (mut g Gen) comptime_match(node ast.MatchExpr) {
 		} else {
 			g.writeln('#else')
 		}
-
-		if node.is_expr {
+		if node.is_expr && !branch.is_comptime_err {
 			len := branch.stmts.len
 			if len > 0 {
-				last := branch.stmts.last() as ast.ExprStmt
-				if len > 1 {
-					g.indent++
-					g.writeln('{')
-					g.stmts(branch.stmts[..len - 1])
-					g.set_current_pos_as_last_stmt_pos()
-					prev_skip_stmt_pos := g.skip_stmt_pos
-					g.skip_stmt_pos = true
-					if is_opt_or_result {
-						tmp_var2 := g.new_tmp_var()
-						g.write('{ ${g.base_type(node.return_type)} ${tmp_var2} = ')
-						g.stmt(last)
-						g.writeln('builtin___result_ok(&(${g.base_type(node.return_type)}[]) { ${tmp_var2} }, (_result*)(&${tmp_var}), sizeof(${g.base_type(node.return_type)}));')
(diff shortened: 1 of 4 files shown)


vlib/v/gen/c/comptime.v
```diff
@@ -1004,49 +1004,57 @@ fn (mut g Gen) comptime_match(node ast.MatchExpr) {
...
+						prev_skip_stmt_pos := g.skip_stmt_pos
+						g.skip_stmt_pos = true
+						if is_opt_or_result {
+							tmp_var2 := g.new_tmp_var()
+							g.write('{ ${g.base_type(node.return_type)} ${tmp_var2} = ')
+							g.stmt(last)
+							g.writeln('builtin___result_ok(&(${g.base_type(node.return_type)}[]) { ${tmp_var2} }, (_result*)(&${tmp_var}), sizeof(${g.base_type(node.return_type)}));')
+							g.writeln('}')
+						} else {
+							println(last)
```
spytheman: this may need to be removed

spytheman:
```suggestion
```

spytheman (approved):
Excellent.
Thank you.

PR #25334, opened 2025-09-17
x.crypto.chacha20: fix internal counter handling

blackshirt:
This small patch fix internal counter handling after rewrites the internal of `x.crypto.chacha20`. The counter increments phase not be adjusted to align with the changes and leads to the bug in [25318](https://github.com/vlang/v/issues/25318).
In my test for `firebird` by @einar-hjortdal, it pass successfully,

```bash
---- Testing... ------------------------------------------------------------------------------------------------------
 OK    [1/5] C:   645.2 ms, R:     2.047 ms /workspaces/firebird/protocol_test.v
 OK    [2/5] C:   613.6 ms, R:     2.126 ms /workspaces/firebird/protocol_utils_test.v
 OK    [3/5] C:   606.7 ms, R:    69.396 ms /workspaces/firebird/secure_remote_password_test.v
 OK    [4/5] C:   746.3 ms, R:   603.540 ms /workspaces/firebird/statement_test.v
 OK    [5/5] C:   595.4 ms, R:     2.202 ms /workspaces/firebird/time_test.v
```

Besides of counter adjusment, this pr also contains some bits of clean up. It also maintain relatives stable performance with previous one, see the benchmark at [here](https://discord.com/channels/592103645835821068/592320321995014154/1417074261096661002)

cc @einar-hjortdal please give this pr a try, i hope this fix the bug

PR #25363, opened 2025-09-21
x.crypto.chacha20: makes 64-bit counter cipher works as an expected 

blackshirt:
This PR was motivated by discord disccussion [here](https://discord.com/channels/592103645835821068/592114487759470596/1417903638705082489), especially, for firebird client testing by @einar-hjortdal with 64-bit counter chacha20 cipher that does not works as an expected previously. This patch was inspired from [nakagami chacha20](https://github.com/nakagami/chacha20) used by golang firebird client in [golang firebirdsql](https://github.com/nakagami/firebirdsql). Its adapted into existing `x.crypto.chacha20` implementation.

This PR contains:

- Rewrites of `xor_key_stream` building block with the new one. Its splitted out keystream generation and xoring bytes into separated logic. Previously, its falls on the same `keystream_with_blocksize`  routine. With this patch, key stream generation separated into `Stream.keystream`, focussing on keystream generation and updating internal counter, where xor-ing messages happens on `xor_key_stream` and or `keystream_full` helper.
- After this applied, the benchmark for `xor_key_stream` show a nice improvement, from ±53.430ms down into ±43.668ms. The benchmark also added into `bench/bench.v` file. It also make it more simpler and readable.
- Some bits of clean up: split out increasing internal counter into own routine in `Stream.inc_ctr`
- Marked a `Cipher.reset` method for deprecation. It intended for internal use, See some possibles culprits of this misbehaved calls tested by @tankf33der at [here](https://github.com/vlang/v/pull/25334#issuecomment-3311553263). Thanks to @tankf33der 👍
- Removes unnecessary codes

After this patch applied, in my testing for [v firebird client](https://github.com/einar-hjortdal/firebird.git) with 64-bit counter activated, the tests was passed successfully. @einar-hjortdal: can you give it a try ?

Cheers,
thanks

### DIFF
--- a/vlib/x/crypto/chacha20/bench/bench.v
+++ b/vlib/x/crypto/chacha20/bench/bench.v
@@ -19,6 +19,13 @@
 // -----------
 // Iterations:      10000          Total Duration:   48.242ms      ns/op:       4824       B/op:      3    allocs/op:      4
 //
+// Chacha20 old xor_key_stream_backup
+// -----------
+// Iterations:      10000          Total Duration:   53.430ms      ns/op:       5343       B/op:     11    allocs/op:     12
+// ChaCha20 new xor_key_stream
+// -----------
+// Iterations:      10000          Total Duration:   43.668ms      ns/op:       4366       B/op:      0    allocs/op:      1
+//
 import x.benchmark
 import encoding.hex
 import x.crypto.chacha20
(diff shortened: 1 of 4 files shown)


einar-hjortdal:
It seems to work well, good work!

blackshirt:
> It seems to work well, good work!

Glad to know its working for you. Thank you for testing it

PR #25373, opened 2025-09-22
v.ast: add Table.update_sym_by_idx/2

kbkpbot:
Allow update existing symbol via call `update_sym()`.

a use case:
To support the new array syntax, we need to update the fixed-arrays size after scan the array lit. For example, 
```v
[..][..]int[[1 2][3 4]]
```
When parse this, we don't known the size of the fixed-array first, but register a `[-1][-1]int` symbol in the system.
 And then after scan the array lit, `[[1 2][3 4]]`, we can update the existing `[-1][-1]int` symbol to `[2][2]int`.

### DIFF
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -840,6 +840,7 @@ fn (mut t Table) rewrite_already_registered_symbol(typ TypeSymbol, existing_idx 
 	$if trace_rewrite_already_registered_symbol ? {
 		eprintln('>> rewrite_already_registered_symbol sym: ${typ.name} | existing_idx: ${existing_idx} | existing_symbol: ${existing_symbol.name}')
 	}
+	t.delete_cached_type_to_str(idx_to_type(existing_idx), map[string]string{})
 	if existing_symbol.kind == .placeholder {
 		// override placeholder
 		t.type_symbols[existing_idx] = &TypeSymbol{
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -835,12 +835,21 @@ pub fn (t &Table) unaliased_type(typ Type) Type {
 	return typ
 }
 
-// update_sym updates the `existing_idx` sym with new `typ`
-pub fn (mut t Table) update_sym(typ TypeSymbol, existing_idx int) {
+// update_sym_by_idx replaces the symbol on the `existing_idx`, with the new `sym`
+pub fn (mut t Table) update_sym_by_idx(existing_idx int, sym &TypeSymbol) {
 	t.delete_cached_type_to_str(idx_to_type(existing_idx), 0)
 	t.type_symbols[existing_idx] = &TypeSymbol{
-		...typ
-		idx: existing_idx
+		...sym
+		idx:   existing_idx
+		size:  -1 // enforce recalculation of the size, for future t.type_size(idx) calls
+		align: -1
+	}
+	for mut esym in t.type_symbols {
+		if esym.size != -1 && esym.info is Alias && esym.info.parent_type == existing_idx {
+			// make sure to force recalculation, if t.type_size(idx) on an already existing alias is called again:
+			esym.size = -1
+			esym.align = -1
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


vlib/v/ast/types.v
```diff
@@ -1459,6 +1459,15 @@ fn strip_extra_struct_types(name string) string {
 	}
 }
 
+// delete_cached_type_to_str remove a `type_to_str` from the cache
+pub fn (t &Table) delete_cached_type_to_str(typ Type, import_aliases map[string]string) {
+	cache_key := (u64(import_aliases.len) << 32) | u64(typ)
```
spytheman:
```suggestion
pub fn (t &Table) delete_cached_type_to_str(typ Type, import_aliases_len int) {
	cache_key := (u64(import_aliases_len) << 32) | u64(typ)
```

spytheman:
Can you please also add a test case?

vlib/v/ast/table.v
```diff
@@ -835,6 +835,15 @@ pub fn (t &Table) unaliased_type(typ Type) Type {
 	return typ
 }
 
+// update_sym updates the `existing_idx` sym with new `typ`
+pub fn (mut t Table) update_sym(typ TypeSymbol, existing_idx int) {
```
spytheman:
```suggestion
// update_sym_by_idx updates the symbol on the `existing_idx`, with the new `sym`
pub fn (mut t Table) update_sym_by_idx(existing_idx int, sym &TypeSymbol) {
```

spytheman: Using `typ` for `TypeSymbol` is confusing (given that name is used for types/integers pretty much everywhere else).

spytheman (approved):

vlib/v/ast/types.v
```diff
@@ -1459,6 +1459,15 @@ fn strip_extra_struct_types(name string) string {
 	}
 }
 
+// delete_cached_type_to_str remove a `type_to_str` from the cache
+pub fn (t &Table) delete_cached_type_to_str(typ Type, import_aliases_len int) {
+	cache_key := (u64(import_aliases_len) << 32) | u64(typ)
+	mut mt := unsafe { &Table(t) }
```
spytheman, line 1465: Why not just `mut t Table` for the receiver?

It is only called by update_sym_by_idx, which already has a mutable receiver 🤔 .

spytheman: I think this new method, will need to loop over all cached_type_to_str entries and delete the ones that have `typ` in the lower bits of the key.

Otherwise the strings for the same type, could be cached in different entries, only differing by their high parts of the key (which is depending on the declared aliases in each .v file).

Previously that was fine, since the type symbols were immutable, but if we allow updates, that cache can go stale, and lead to bugs, and we need to be careful to update or invalidate it properly.

spytheman: I.e. the new signature should be something like this:
```v
pub fn (mut t Table) delete_cached_type_to_str(typ Type) {
```

spytheman:
Thinking about it more, would not it be easier, to just register a placeholder symbol, while parsing the `[..]int[10,20]`, then after the parser knows the complete information about the size, register a completely *new* type and symbol, based on the complete information, without updating the old placeholder one?

Then we would not have to invalidate any caches, and cgen could just skip the placeholder symbols completely.

spytheman:
> Think about it, dealing with array declarations like [5][..][][const][..]int feels quite tricky.

You may be right, but I do not understand why?

In one case, you will have to update an existing symbol, and deal with the cache invalidations/updates (because currently the assumption is that the type/symbol information does not change much once registered).

In the other, you will be copying an existing placeholder symbol on each recursion level, changing it in the process, and storing a new version of it (under a new type id), leaving the old one as it is - the cached entries in the table will continue to work for the placeholder types/symbols, and after the parsing is done, you would have a valid and complete type for the full array init. The negative could be using more memory for the new symbols, and potentially some logic in cgen to skip the incomplete ones.

spytheman:
Anyway, given that you have a clear picture of how to do it with your approach, while I have not considered all implications of mine, I do not want to block it - I will merge this PR as it is, and we will see how it will pan out. If there are too many bugs found due to the cache maintenance, we could then reconsider etc.

PR #25374, opened 2025-09-22
redis: add TLS support

davlgd:
This PR adds TLS support to Redis module, by adding a `tls` parameter to `Config` and handlers to manage commands with a TCP or a TLS connexion

### DIFF
--- a/vlib/db/redis/redis.v
+++ b/vlib/db/redis/redis.v
@@ -3,6 +3,7 @@
 module redis
 
 import net
+import net.ssl
 import strings
 
 // RedisValue represents all possible RESP (Redis Serialization Protocol) data types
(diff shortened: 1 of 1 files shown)


vlib/db/redis/redis.v
```diff
@@ -58,7 +70,39 @@ pub fn connect(config Config) !DB {
...
+		return db.conn.read(mut buf)
+	}
+}
+
+fn (mut db DB) read_ptr_data(ptr &u8, len int) !int {
+	if db.use_tls {
+		mut buf := []u8{len: len}
+		bytes_read := db.ssl_conn.read(mut buf)!
+		unsafe { C.memcpy(ptr, buf.data, bytes_read) }
+		return bytes_read
```
spytheman:
```suggestion
		return db.ssl_conn.socket_read_into_ptr(buf.data, buf.len)!
```

spytheman: We should probably also add a `read_ptr` method to ssl.SSLConn , but that is out of scope for this PR.

spytheman (changes requested):

PR #25377, opened 2025-09-23
x.crypto.chacha20: expands xchacha20 construction to support for 64-bit counter

blackshirt:
This patch expands eXtended ChaCha20 construction to support for 64-bit counter, completes support for the existing 32-bit counter construct.
@spytheman, @tankf33der: can you look into it ?

Cheers,
thanks

### DIFF
--- a/vlib/x/crypto/chacha20/chacha.v
+++ b/vlib/x/crypto/chacha20/chacha.v
@@ -32,12 +32,20 @@ enum CipherMode {
 	original
 }
 
+// Configuration options
+@[params]
+pub struct Options {
+pub mut:
+	// currently, used for XChaCha20 construct
+	with_64bit_ctr bool
+}
+
 // encrypt encrypts plaintext bytes with ChaCha20 cipher instance with provided key and nonce.
 // It was a thin wrapper around two supported nonce size, ChaCha20 with 96 bits
 // and XChaCha20 with 192 bits nonce. Internally, encrypt start with 0's counter value.
 // If you want more control, use Cipher instance and setup the counter by your self.
-pub fn encrypt(key []u8, nonce []u8, plaintext []u8) ![]u8 {
-	mut stream := new_stream(key, nonce)!
+pub fn encrypt(key []u8, nonce []u8, plaintext []u8, opt Options) ![]u8 {
+	mut stream := new_stream_with_options(key, nonce, opt)!
 	mut dst := []u8{len: plaintext.len}
 	stream.keystream_full(mut dst, plaintext)!
 	unsafe { stream.reset() }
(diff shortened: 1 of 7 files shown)


tankf33der:
Extensively tested. All passed.

vlib/x/crypto/chacha20/chacha.v
```diff
@@ -32,12 +32,20 @@ enum CipherMode {
 	original
 }
 
+// Configuration options
+@[params]
+pub struct Options {
+pub mut:
+	// currently, used for XChaCha20 construct
+	with_64bit_ctr bool
```
spytheman:
```suggestion
	use_64bit_counter bool
```

spytheman: @JalonSolov what do you think?

JalonSolov: Your suggestion is better.  `with` is more for telling something that's already done or being done.  `use` means you are telling it you want it to be done that way.

vlib/x/crypto/chacha20/chacha.v
```diff
@@ -71,8 +79,28 @@ mut:
...
+// mut c := chacha20.new_cipher(key, nonce)!
+// // Do what you want with c
+//
+// // 2. creates an eXtended ChaCha20 with 64-bit counter
+// key2 := rand.bytes(32)!
+// nonce2 := rand.bytes(24)!
+//
+// mut c2 := chacha20.new_cipher(key2, nonce2, with_64bit_ctr: true)!
+//
+// ```
```
spytheman: Imho this example will be clearer if it is extracted as a code block in the Readme file, where it will be easier to maintain/check/copy .
The reason for that, is because it is longer than 2 lines, and it has comments on its own.

blackshirt: There is code block on the readme file. I Will update it

vlib/x/crypto/chacha20/chacha.v
blackshirt: Ok..will be updated

vlib/x/crypto/chacha20/chacha.v
blackshirt: Updated

vlib/x/crypto/chacha20/chacha.v
blackshirt: Updated

blackshirt:
> Thank you @blackshirt .

Thank you @spytheman for suggestion and reviews. There are patch in queue related to apply precomputation technique for 64-bit counter variant, like we do on 32-bit counter variant, but, i dont know, is this should be sent or not?. In my test, it improves a small of bit, but not too much.

PR #25382, opened 2025-09-24
x.crypto.ascon: fix decryption

tankf33der:
Currently, decryption is broken due to a trivial typo.

Here is the line in reference [implementation](https://github.com/ascon/ascon-c/blob/main/crypto_aead/asconaead128/ref/aead.c#L208) that can be easily compared to spot the typo.

Test will be added in the next PR when encryption will be fixed too.


tankf33der:
@blackshirt - take a look.

blackshirt:
Looks good  for me. Thank you @tankf33der for  correcting this and sent a pr for the fix

spytheman (approved):
Thank you @tankf33der 🙇🏻 .

PR #25385, opened 2025-09-25
x.crypto.ascon: fix encryption, add test

tankf33der:
Similarly to decryption, encryption is broken if the remainder of dividing the message length by 16 is greater than or equal to 8.

I refactored this code block to resemble the reference implementation for easier [comparison.](https://github.com/ascon/ascon-c/blob/main/crypto_aead/asconaead128/ref/aead.c#L84)

This patch passed `~1B` iteration vs. reference implementation library.
Test also implemented.

### DIFF
--- a/vlib/x/crypto/ascon/aead128.v
+++ b/vlib/x/crypto/ascon/aead128.v
@@ -370,18 +370,12 @@ fn aead128_process_msg(mut out []u8, mut s State, msg []u8) int {
 	}
 	// process partial block if it exists
 	if mlen >= 8 {
-		mut first_block := unsafe { msg[midx..] }
-		s.e0 ^= load_bytes(first_block[0..8], 8)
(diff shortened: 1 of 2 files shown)


JalonSolov:
Has the performance of the routine changed?  Since it is supposed to be for IoT level, it should be as fast as possible.

tankf33der:
@blackshirt take a look.

tankf33der:
> Has the performance of the routine changed? Since it is supposed to be for IoT level, it should be as fast as possible.

@JalonSolov 

ascon

spytheman (approved):
Excellent work.
Thank you @tankf33der 🙇🏻 .

blackshirt:
> @blackshirt take a look.

Sorry for late respond.Thank for spotting the issue and sent fix for it 👍

PR #25392, opened 2025-09-26
os: propagate POSIX error code from os.ls/1 and os.rmdir_all/1 on UNIX-like systems

gechandesu:
Return the error code from `os.ls/1` and `os.mkdir_all/1` to enable detailed error handling e.g. you can now differ ENOENT (no such file or directory) and other possible I/O errors:

```
>>> os.ls('/nonexistent') or { println(err.code()) }
2
>>> os.rmdir_all('/nonexistent') or { println(err.code()) }
2
```

### DIFF
--- a/vlib/os/os_nix.c.v
+++ b/vlib/os/os_nix.c.v
@@ -282,11 +282,9 @@ pub fn ls(path string) ![]string {
 	mut res := []string{cap: 50}
 	dir := unsafe { C.opendir(&char(path.str)) }
 	if isnil(dir) {
-		// return error_with_code('ls() couldnt open dir "${path}"', C.errno)
 		return error_posix(msg: 'ls() couldnt open dir "${path}"')
 	}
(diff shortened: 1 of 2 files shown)


vlib/os/os_nix.c.v
```diff
@@ -282,10 +282,9 @@ pub fn ls(path string) ![]string {
 	mut res := []string{cap: 50}
 	dir := unsafe { C.opendir(&char(path.str)) }
 	if isnil(dir) {
-		return error('ls() couldnt open dir "${path}"')
+		return error_posix(msg: 'ls() couldnt open dir "${path}"')
```
spytheman, line 285: This is not safe, since the C.errno could be changed during the string interpolation in the argument.

spytheman: you can call `e := last_error()` separately, if you need to include details about the exact reason, and then access its e.code() and e.msg() to compose the final error

PR #25396, opened 2025-09-27
cgen: reduce empty lines in default C prelude stored in cheaders.v

spytheman:


spytheman:
@Avey777, `huly-for-github` is a bot. What exactly is there to like in its comment?

Issue #24650, opened 2025-06-04
urllib.URL has a nil pointer field


einar-hjortdal:
### Describe the bug

URL.user should be an option type, not a nil pointer.
As it is now, it forces users to use unsafe blocks instead of the idiomatic unwrapping

```v
	url := parse_url(opts.url)!
	opts.address = get_address(url.host)
	unsafe { // why the hell is url.user a nil pointer?
		if url.user != nil {
			opts.username = url.user.username
			opts.password = url.user.password
		}
	}
	opts.db = url.path.trim('/').int()
```

### Reproduction Steps

none

### Expected Behavior

URL.user should be an option type

### Current Behavior

URL.user is a pointer, nil by default

### V version

V 0.4.10 061da6a

### Environment details (OS name and version, etc.)

```
|V full version      |V 0.4.10 464e2973511a9090c6e70a0f53ec5b74aabbffcb.061da6a
|:-------------------|:-------------------
|OS                  |linux, "openmamba release 2025.5 for x86_64 (rolling)"
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz
|Memory              |17.1GB/31.16GB
|                    |
|V executable        |/home/einar/.local/lib64/v/v
|V last modified time|2025-06-04 08:11:21
|                    |
|V home dir          |OK, value: /home/einar/.local/lib64/v
|VMODULES            |OK, value: /home/einar/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/einar/Documents/projects/vlang/active/redict
|                    |
|Git version         |git version 2.49.0
|V git status        |weekly.2025.22-53-g061da6aa
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.1.0
|gcc version         |gcc (GCC) 15.1.0
|clang version       |N/A
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.41
```

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

jorgeluismireles:
Next doesn't need check against `nil` but for no user return empty fields:
[...]

PR #25401, opened 2025-09-27
net.urllib: use option instead of ptr for `Userinfo` field 

Delta456:
Fixes https://github.com/vlang/v/issues/24650


vlib/net/urllib/urllib.v
```diff
@@ -714,12 +714,13 @@ pub fn (u URL) str() string {
 	if u.opaque != '' {
 		buf.write_string(u.opaque)
 	} else {
-		if u.scheme != '' || u.host != '' || !u.user.empty() {
-			if u.host != '' || u.path != '' || !u.user.empty() {
+		user := u.user or { Userinfo{} }
```
spytheman, line 717:
```suggestion
		if user := u.user {
```

vlib/net/urllib/urllib_test.v
```diff
@@ -88,7 +88,7 @@ fn test_parse_empty_query_two() {
 	assert query_str == query_encode
 }
 
-fn test_parse() {
+fn test_parse() ? {
```
spytheman: there is no need for this change - test functions are similar to main

vlib/net/urllib/urllib.v
Delta456: If we do this, the tests will fail

```
 FAIL     1.974 ms vlib/net/urllib/urllib_test.v
/home/delta/Documents/git/v_dev/vlib/net/urllib/urllib_test.v:45: fn test_parse_missing_host
   > assert url.str() == 'http://///'
     Left value (len: 8): `http:///`
    Right value (len: 10): `http://///`
 ```

vlib/net/urllib/urllib_test.v
Delta456: Right

spytheman (approved):
Excellent work.

Issue #25404, opened 2025-09-28
complex const expr not at first dim, will cause v parser fail


kbkpbot:
### Describe the bug

If declare an array type which has a complex expr not at first dim, it will cause V parser fail.

### Reproduction Steps

```v
module main

const k = 2

fn main() {
        a := [(k-1)*2][2]f64{}        // This is OK
        dump(a)
        b := [2][(k-1)*2]f64{}   // fail
        dump(b)
}
```

```
v m.v
```

### Expected Behavior

compile ok && run ok

### Current Behavior

```
$ v m.v
$ v m.v
m.v:8:11: error: unexpected token `(`, expecting `]`
    6 |     a := [(k-1)*2][2]f64{}
    7 |     dump(a)
    8 |     b := [2][(k-1)*2]f64{}
      |              ^
    9 |     dump(b)
   10 | }
```

### V version

V 0.4.12 1dd172e0809aad749b6e3e701220ce12ffcfa55c

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 1dd172e0809aad749b6e3e701220ce12ffcfa55c
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |11.63GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-09-28 00:35:11
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/arr
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.36-131-g1dd172e0
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
[...]

PR #25409, opened 2025-09-28
parser: fix eval fixed array size

kbkpbot:
Fix issue #25404

BTW, I think we should remove `eval_array_fixed_sizes()` from `parser`. Because in `parser`, it has no full information, for example, `[2][const_a - sizeof(T)]int{}`, `sizeof(T)` is ready in `checker`. And `checker` also has it's own `eval_array_fixed_sizes()`.

### DIFF
--- a/vlib/v/parser/parse_type.v
+++ b/vlib/v/parser/parse_type.v
@@ -11,74 +11,108 @@ import v.token
 const maximum_inline_sum_type_variants = 3
 const generic_type_level_cutoff_limit = 10 // it is very rarely deeper than 4
 
-fn (mut p Parser) parse_array_type(expecting token.Kind, is_option bool) ast.Type {
-	p.check(expecting)
-	// fixed array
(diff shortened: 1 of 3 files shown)


spytheman:
> BTW, I think we should remove eval_array_fixed_sizes() from parser. 

I agree completely - the checker is more suitable to calculate it.

vlib/v/parser/parse_type.v
```diff
@@ -11,75 +11,109 @@ import v.token
...
-					}
+				else {
+					size, unresolved := p.eval_array_fixed_sizes(mut size_expr.expr)
+					return int(size), unresolved
 				}
-				ast.InfixExpr {
+			}
+		}
+		ast.Ident {
+			if mut const_field := p.table.global_scope.find_const(size_expr.full_name().all_after('builtin.')) {
```
spytheman, line 67: why is the `.all_after('builtin.')` needed ?

PR #25419, opened 2025-09-29
`net.urllib`: `Userinfo` -> `UserInfo`

Delta456:


spytheman:
Go uses Userinfo, see https://pkg.go.dev/net/url#Userinfo . I do not see a good reason to change to UserInfo (but I do not mind it either).

medvednikov:
I agree it should be the same as Go

Delta456:
Ok, I will close the PR

PR #25421, opened 2025-09-30
x.crypto.chacha20poly1305: enables support for (x)chacha20 with 64-bit counter construct

blackshirt:
This patch enables supports for underlying variant already supported by `x.crypto.chacha20` module, like a supports for 64-bit counter with 8-bytes nonce, and experimental features of extended construct with 64-bit counter.
This patch contains:

- Enables support for original 64-bit counter with 8-bytes nonce.
- Enables a `chacha20.Options` in the constructor and encryption routines to pass an option into underlying chacha20 constructor.
- Unifies generic `encryption` and `decryption` routines into single one, its already shares a common thing.
- Add a test material for `XChaCha20Poly1305` construct, adapted from `pycryptodome` vector tests
- Some bits of cleanup

@tankf33der: can you take a look into it?

NB: This is my latest bench run of `vlib/x/crypto/ascon/bench/aead.v`

```bash
Encryption..
-----------
Iterations:      10000          Total Duration:   24.708ms      ns/op:       2470       B/op:     16    allocs/op:     17
Iterations:      10000          Total Duration:  133.934ms      ns/op:      13393       B/op:     16    allocs/op:     16

Decryption..
-----------
Iterations:      10000          Total Duration:   27.657ms      ns/op:       2765       B/op:      6    allocs/op:      8
Iterations:      10000          Total Duration:  147.583ms      ns/op:      14758       B/op:      6    allocs/op:      9
```

So, after several patch, for `x.crypto.chacha20poly1305` aead operations, the number  tells the improvement from previous  number: For aead encryption, the total duration:  `158.865ms` reduced into `133.934ms` and for aead decryption, the total duration:  `158.373ms` drop into `147.583ms`. 

Thanks
Cheers.

### DIFF
--- a/vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
+++ b/vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
@@ -174,24 +174,24 @@ fn (c Chacha20Poly1305) generic_crypt(msg []u8, nonce []u8, ad []u8, mode Mode) 
 	s.encrypt(mut otkey, otkey)!
 
 	// destination buffer, with overhead spaces for generated tag without reallocating
-	mut dst := []u8{len: src.len + c.overhead()}
+	mut dst := []u8{len: src.len, cap: src.len + c.overhead()}
 
 	// Next, the ChaCha20 encryption function is called to encrypt (decrypt) message input,
 	// using the same key and nonce, and with the initial counter set to 1.
 	s.set_counter(1)
-	s.encrypt(mut dst[..src.len], src)!
+	s.encrypt(mut dst, src)!
 
 	// Finally, the Poly1305 function is called with the Poly1305 key calculated above
 	// to build message authentication code (tag).
 	// length of constructed message
 	cm_length := if mode == .encrypt {
-		length_constructed_msg(ad, dst[..src.len])
+		length_constructed_msg(ad, dst)
 	} else {
 		length_constructed_msg(ad, src)
 	}
 	mut constructed_msg := []u8{len: cm_length}
 	if mode == .encrypt {
-		construct_msg(mut constructed_msg, ad, dst[..src.len])
+		construct_msg(mut constructed_msg, ad, dst)
 	} else {
 		construct_msg(mut constructed_msg, ad, src)
 	}
@@ -215,11 +215,11 @@ fn (c Chacha20Poly1305) generic_crypt(msg []u8, nonce []u8, ad []u8, mode Mode) 
 			return error('chacha20poly1305: unmatching tag')
 		} else {
 			// return the decrypted message (plaintext) when the tag was matching
-			return dst[..src.len]
+			return dst
 		}
 	}
-	// In the encryption mode, copies the tag into end of destination buffer
-	_ := copy(mut dst[src.len..], tag)
+	// In the encryption mode, appends the tag into end of destination buffer
+	dst << tag
 	return dst
 }
 
--- a/vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
+++ b/vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
@@ -5,7 +5,7 @@
 // AEAD_CHACHA20_POLY1305 is an authenticated encryption with additional data algorithm.
 // The inputs to AEAD_CHACHA20_POLY1305 are:
 //   A 256-bit key
-//   A 96-bit nonce (or bigger 192 bit nonce) -- different for each invocation with the same key
+//   A 64-bit nonce, 96-bit (or bigger 192 bit nonce) -- different for each invocation with the same key
 //   An arbitrary length plaintext
 //   Arbitrary length additional authenticated data (AAD)
 module chacha20poly1305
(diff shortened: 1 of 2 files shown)


vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -170,69 +145,99 @@ pub fn (c Chacha20Poly1305) decrypt(ciphertext []u8, nonce []u8, ad []u8) ![]u8
...
+		} else {
+			// return the decrypted message (plaintext) when the tag was matching
+			return dst
 		}
-		return error('chacha20poly1305: unmatching tag')
 	}
-
-	return plaintext
+	// In the encryption mode, appends the tag into ciphertext buffer
+	dst << tag
```
tankf33der: I think that using `<<` here and in `construct_msg()` is not optimal. Can the same be implemented differently?

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -170,69 +145,99 @@ pub fn (c Chacha20Poly1305) decrypt(ciphertext []u8, nonce []u8, ad []u8) ![]u8
...
+	// with the supplied mac, otherwise return error on fails and free allocated resources.
+	if mode == .decrypt {
+		if subtle.constant_time_compare(mac, tag) != 1 {
+			// free allocated resource
+			unsafe {
+				s.free()
+				tag.free()
+				otkey.free()
+				dst.free()
+				constructed_msg.free()
```
tankf33der, line 213: Can you add `po.free()` too?

tankf33der:
Old two `nonces` and new one passed `100+M` loop vs. `Monocypher` library.
Good sign.

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -170,69 +145,99 @@ pub fn (c Chacha20Poly1305) decrypt(ciphertext []u8, nonce []u8, ad []u8) ![]u8
...
-	encrypted := ciphertext[0..ciphertext.len - c.overhead()]
-	mac := ciphertext[ciphertext.len - c.overhead()..]
+	// generates 32-bytes of one-time key for later poly1305 operation
+	mut otkey := []u8{len: key_size}
+	mut s := chacha20.new_cipher(c.key, nonce, c.opt)!
+	s.encrypt(mut otkey, otkey)!
 
-	mut plaintext := []u8{len: encrypted.len}
+	// destination buffer
+	mut dst := []u8{len: src.len}
```
JalonSolov:
```suggestion
	mut dst := []u8{cap: src.len}
```
By setting `cap` instead of `len`, the array space is allocated, but not initialized.

If you use `len` here, then later use `dst << tag`, you are guaranteed to reallocate the array, since it is already max size.

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: Something like builtin `copy`?
Or what do you think forbetter alternatives. ?

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: I think its should be on different patch. `po.finish` already done `.reset` internally

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
JalonSolov: See my comment about `len` vs `cap` on `dst` init.

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: I Will take a look into it

blackshirt: > By setting `cap` instead of `len`, the array space is allocated, but not initialized.
> 
> If you use `len` here, then later use `dst << tag`, you are guaranteed to reallocate the array, since it is already max size.

@JalonSolov : without properly correctly initialized `dst` length, `chacha20.encrypt` would fail with error `chacha20: dst buffer is to small`

JalonSolov: Don't check `dst.len`.  Check `dst.cap`.

JalonSolov: Or just check `src.len`, since that's the value used to init the array.

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: @tankf33der. @JalonSolov i've updated it based on the suggestion

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -173,19 +173,25 @@ fn (c Chacha20Poly1305) generic_crypt(msg []u8, nonce []u8, ad []u8, mode Mode)
 	mut s := chacha20.new_cipher(c.key, nonce, c.opt)!
 	s.encrypt(mut otkey, otkey)!
 
-	// destination buffer
-	mut dst := []u8{len: src.len}
+	// destination buffer, with overhead spaces for generated tag without reallocating
+	mut dst := []u8{len: src.len + c.overhead()}
```
JalonSolov:
```suggestion
	mut dst := []u8{len: src.len, cap: src.len + c.overhead()}
```

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -173,19 +173,25 @@ fn (c Chacha20Poly1305) generic_crypt(msg []u8, nonce []u8, ad []u8, mode Mode)
...
 
 	// Next, the ChaCha20 encryption function is called to encrypt (decrypt) message input,
 	// using the same key and nonce, and with the initial counter set to 1.
 	s.set_counter(1)
-	s.encrypt(mut dst, src)!
+	s.encrypt(mut dst[..src.len], src)!
```
JalonSolov: Use the previous line.

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -173,19 +173,25 @@ fn (c Chacha20Poly1305) generic_crypt(msg []u8, nonce []u8, ad []u8, mode Mode)
...
+	// length of constructed message
+	cm_length := if mode == .encrypt {
+		length_constructed_msg(ad, dst[..src.len])
+	} else {
+		length_constructed_msg(ad, src)
+	}
+	mut constructed_msg := []u8{len: cm_length}
 	if mode == .encrypt {
-		construct_msg(mut constructed_msg, ad, dst)
+		construct_msg(mut constructed_msg, ad, dst[..src.len])
```
JalonSolov: Use the previous line.

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -209,11 +215,11 @@ fn (c Chacha20Poly1305) generic_crypt(msg []u8, nonce []u8, ad []u8, mode Mode)
 			return error('chacha20poly1305: unmatching tag')
 		} else {
 			// return the decrypted message (plaintext) when the tag was matching
-			return dst
+			return dst[..src.len]
```
JalonSolov: Use the previous line.

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
```diff
@@ -209,11 +215,11 @@ fn (c Chacha20Poly1305) generic_crypt(msg []u8, nonce []u8, ad []u8, mode Mode)
...
 		}
 	}
-	// In the encryption mode, appends the tag into ciphertext buffer
-	dst << tag
+	// In the encryption mode, copies the tag into end of destination buffer
+	_ := copy(mut dst[src.len..], tag)
```
JalonSolov: Use the previous 2 lines.

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: Applied

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: Applied

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: Applied

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: Applied

vlib/x/crypto/chacha20poly1305/chacha20poly1305.v
blackshirt: Applied

tankf33der:
Latest commit(s) passed.

tankf33der:
This is fastest copy i know:
```
./sm4.v:                unsafe { vmemcpy(&input_16, &input[idx], 16) } // convert to fixed size array
```

blackshirt:
> This is fastest copy i know:
> 
> ```
> ./sm4.v:                unsafe { vmemcpy(&input_16, &input[idx], 16) } // convert to fixed size array
> ```

its can be applied directly. but, I think c-gen backend can optimize this typical of copy operation.
I really enjoy writing code in pure v, and try to reduce some `unsafe` style code in my code, except some small parts of it.

blackshirt:
> Latest commit(s) passed.

Its sounds a good thing

PR #25427, opened 2025-10-01
thirdparty, vlib: add thirdparty/stb_vorbis; add a thin wrapper module for it too + test in `encoding.vorbis`

spytheman:


Delta456:
Why do we need this in `vlib`? Can't it be a separate installable package?

spytheman:
> Why do we need this in vlib? Can't it be a separate installable package?

For the same reason, that we wrap `thirdparty/stb_image/` - just instead of popular compressed image formats, it can load/decode a popular compressed audio format (vorbis/ogg) . An mp3 decoder would be also nice to have, but I could not find a small one without dependencies.

It allows for small games/demos that use `gg`, to play simple music and sound effects, with very minimal overhead (the size of the decoder is ~60KB with -prod and gcc), and without relying on external libs like raylib, sdl3, or miniaudio.

larpon:
Very nice addition!

PR #25428, opened 2025-10-02
checker: always use ... prefix when passing array to vararg

### LINKED ISSUE #25424: checker miss leads to cgen error when using sum type in var arg as array in static method
<details>
<summary>V version: V 0.4.12, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 9772649250016a8558fe9bf07ba94f2c812fa9ef
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 8845HS w/ Radeon 780M Graphics
|Memory              |16.11GB/30.64GB
|                    |
|V executable        |/home/lmp/Projects/v/v
|V last modified time|2025-09-30 07:15:01
|                    |
|V home dir          |OK, value: /home/lmp/Projects/v
|VMODULES            |OK, value: /home/lmp/.vmodules
|VTMP                |OK, value: /dev/shm
|Current working dir |OK, value: /home/lmp/Projects/v
|                    |
|V git status        |weekly.2025.40-2-g97726492
|.git/config present |true
|                    |
|clang version       |clang version 20.1.8
|emcc version        |shared:INFO: old sanity: 4.0.12-git|/opt/emscripten-llvm/bin

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  bugs/t.v && /home/lmp/Projects/v/bugs/t`
```v
module main

type Nil = u8

const c_nil = Nil(`\0`)

pub type Value = Nil | int | string | []Value | Map

pub fn (v Value) string() string {
	return '${v.str()}'
}

pub struct Map {
mut:
	m map[string]Value
}

pub fn Map.create(values ...Value) Map {
	assert values.len > 1
	assert values.len % 2 == 0
	mut m := Map{}
	for i := 0; i < values.len; i += 2 {
		m.set(values[i], values[i + 1])
	}
	return m
}

pub fn (mut m Map) set(key Value, value Value) {
	m.m[key.string()] = value
}

pub fn build_map(index_from int, index_to int) Map {
	mut kv_pairs := []Value{len: index_to - index_from, init: Value(c_nil)}
	for i := index_from; i < index_to; i++ {
		kv_pairs[i - index_from] = Value(i)
	}
	assert kv_pairs.len % 2 == 0
	// NOTE: this alternative works:
	// mut m := Map{}
	// for i := 0; i < kv_pairs.len; i += 2 {
	// 	m.set(kv_pairs[i], kv_pairs[i + 1])
	// }
	// return m
	// NOTE: this fails
	return Map.create(kv_pairs)
}

fn main() {
	m := build_map(2, 4)
	dump(m)
}

```

**What did you see?**
```
cc: /dev/shm/t.01K6DDH13DJ9Z0CYK7DB8C337T.tmp.c:6645: error: cannot convert 'struct main__Value' to 'struct array'
builder error: 
==================
```

Something else than the C error, probably the `dump(m)`

> [!NOTE]
>

dy-tea:
Fixes #25424.

When passing an array of sumtype to a function taking a vararg, the `...` prefix was not required which caused a cgen error. 
This makes sure the `...` prefix is required for arrays.

The check is a bit long so let me know if there's a simpler way.

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -1528,20 +1528,39 @@ fn (mut c Checker) fn_call(mut node ast.CallExpr, mut continue_check &bool) ast.
 				c.expr(mut call_arg.expr)
 			}
 			if i == args_len - 1 {
-				if c.table.sym(typ).kind == .array && call_arg.expr !is ast.ArrayDecompose
-					&& c.table.sym(expected_type).kind !in [.sum_type, .interface]
-					&& !param.typ.has_flag(.generic) && expected_type != typ {
-					styp := c.table.type_to_str(typ)
-					elem_styp := c.table.type_to_str(expected_type)
+				kind := c.table.sym(typ).kind
+				is_decompose := call_arg.expr is ast.ArrayDecompose
+				mut allow_variadic_pass := false
+				if call_arg.expr is ast.Ident && !func.is_method {
+					ident := call_arg.expr as ast.Ident
+					if ident.obj is ast.Var {
+						var_obj := ident.obj as ast.Var
+						if var_obj.is_arg && c.table.cur_fn != unsafe { nil }
+							&& c.table.cur_fn.is_variadic && c.table.cur_fn.params.len > 0
+							&& ident.name == c.table.cur_fn.params.last().name && func.is_variadic {
+							cur_fn_param_sym := c.table.sym(c.table.cur_fn.params.last().typ)
+							if cur_fn_param_sym.info is ast.Array {
+								allow_variadic_pass = cur_fn_param_sym.info.elem_type == expected_type
+							}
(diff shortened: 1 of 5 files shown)


dy-tea:
Small refactor, should be shorter now.

spytheman (approved):
Sorry for not reviewing it sooner.
Excellent work.
Thanks you.

koplenov:
unexpected breaking change(
изображение

dy-tea:
The first one should indeed be an error but the second one shouldn't be, can you make a repro?

koplenov:
https://github.com/koplenov/treevee before last fix commit

dy-tea:
I found a repro, will open an issue.
See #25504.

PR #25429, opened 2025-10-02
token,scanner,parser,checker,cgen: add file_idx to token.Pos

kbkpbot:
The original `token.Pos` does not include file name information.
This PR adds a `file_idx` field to `token.Pos`, allowing us to identify which source file each token belongs to.

### DIFF
--- a/vlib/v/token/pos.v
+++ b/vlib/v/token/pos.v
@@ -5,11 +5,11 @@ module token
 
 pub struct Pos {
 pub:
-	file_idx int = -1 // file idx in the global table `filelist`
 	len      int // length of the literal in the source
 	line_nr  int // the line number in the source where the token occurred
 	pos      int // the position of the token in scanner text
 	col      int // the column in the source where the token occurred
+	file_idx int = -1 // file idx in the global table `filelist`
 pub mut:
 	last_line int // the line number where the ast object ends (used by vfmt)
 }


vlib/v/token/pos.v
```diff
@@ -5,10 +5,11 @@ module token
 
 pub struct Pos {
 pub:
-	len     int // length of the literal in the source
-	line_nr int // the line number in the source where the token occurred
-	pos     int // the position of the token in scanner text
-	col     int // the column in the source where the token occurred
+	file_idx int = -1 // file idx in the global table `filelist`
```
spytheman: That field will be very rarely accessed, please add it at the end.

spytheman:
I've reverted it, because it caused a performance regression, without a clear benefit for correctness or usability:
image

spytheman:
> Can you share the script you use to compare performance?

The screenshot is from the #fast channel in the V discord server, but you can run something locally, that shows a similar result with:
`./v run .github/workflows/compare_pr_to_master.v -prod`

(executed on the branch that has the changes)

spytheman:
I suspect, that the slowdown is caused mainly by the increased size of the Pos struct, since it is used a lot in practically every AST node.

If the new file index is encoded in the high 16 bits of the `col` field, the total size of ast.Pos will be the same, and perhaps the impact will be much lower.
Although I am not sure if just changing it to `col u16` and `idx i16` may not be enough. It will still allow for 65536 different columns, and 32767 different file indexes.

PR #25439, opened 2025-10-04
transformer: add test

kbkpbot:
Add test for `transformer`, test `infix_expr` right now.

### DIFF
--- a/vlib/v/transformer/tests/const_infix_expr_test.v
+++ b/vlib/v/transformer/tests/const_infix_expr_test.v
@@ -1,0 +1,33 @@
+import v.ast
+import v.pref
+import v.parser
+import v.checker
+import v.transformer
+
+fn test_const_infix_expr() {
+	println(@LOCATION)
+	source_text := '
+const k = 2
+fn main() {
+	x := [(k-1)*2]int{}
+}
+'
+	mut table := ast.new_table()
+	vpref := &pref.Preferences{}
+	mut prog := parser.parse_text(source_text, '', mut table, .skip_comments, vpref)
+	mut checker_ := checker.new_checker(table, vpref)
+	checker_.check(mut prog)
+	mut t := transformer.new_transformer_with_table(table, vpref)
+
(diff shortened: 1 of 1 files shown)


vlib/v/transformer/tests/const_infix_expr_test.v
```diff
@@ -0,0 +1,33 @@
...
+'
+	mut table := ast.new_table()
+	vpref := &pref.Preferences{}
+	mut prog := parser.parse_text(source_text, '', mut table, .skip_comments, vpref)
+	mut checker_ := checker.new_checker(table, vpref)
+	checker_.check(mut prog)
+	mut t := transformer.new_transformer_with_table(table, vpref)
+
+	// get the `InfixExpr`(`(k-1)*2`) from table
+	main_fn := unsafe { table.cur_fn[0] }
```
spytheman, line 23:
```suggestion
	main_fn := unsafe { table.cur_fn }
```

vlib/v/transformer/tests/const_infix_expr_test.v
```diff
@@ -0,0 +1,33 @@
...
+	assign_stmt := main_fn.stmts[0] as ast.AssignStmt
+	array_init_expr := assign_stmt.right[0] as ast.ArrayInit
+	mut dim_expr := array_init_expr.exprs[0] as ast.InfixExpr
+	dump(dim_expr)
+
+	// verify `infix_expr` work as expected
+	folded_expr := t.infix_expr(mut dim_expr)
+	dump(folded_expr)
+	assert '${folded_expr}' == '(1) * 2'
```
spytheman, line 32: Should not that reduce to just `2` ?

spytheman: what happens if you call `t.infix_expr(folded_expr)`, will it produce `1*2`, removing the () ?

kbkpbot: Yes, because it was not reduced to 2, it caused calculation issues when using as eval const expr.

PR #25440, opened 2025-10-04
transformer: eval par expr for literal and ident

kbkpbot:
Remove par when the inner expr is literal or ident.
This will folder const expr as expected.

I'm not sure if doing this has side effects? For example, in terms of syntax, parentheses might have new uses, such as `(k)` representing some special meaning. If we remove the parentheses, it might change the semantics.


spytheman:
Afaik removing the () around a literal/ident, should not change the semantics of the expression. I think it will be ok.

Good work.

Issue #25379, opened 2025-09-23
Program hangs `for i := from; i <= to; i++` where `u8` and `to = 255`


jorgeluismireles:
<details>
<summary>V version: V 0.4.12 0df4435, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 e96355aca59809ed3f642e82d3136a1f1b3fd4c9.0df4435
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 22.04.5 LTS
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz
|Memory              |0.2GB/3.7GB
|                    |
|V executable        |/home/jorge/v/v
|V last modified time|2025-09-23 19:58:14
|                    |
|V home dir          |OK, value: /home/jorge/v
|VMODULES            |OK, value: /home/jorge/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/jorge/bugs
|                    |
|Git version         |git version 2.34.1
|V git status        |weekly.2025.16-822-g0df44356-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
|gcc version         |gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
|clang version       |Ubuntu clang version 14.0.0-1ubuntu1.1
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.5 ()
|glibc version       |ldd (Ubuntu GLIBC 2.35-0ubuntu3.11) 2.35

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  250923.v && 250923`
```v
fn main() {
	start(1, 255)
}

fn start(from u8, to u8) {
    mut ips := []string{}
    subnet := '10.8.1'
    for i := from; i <= to; i++ {
        ips << '${subnet}.${i}'
    }
    println('ips: ${ips.len}')
}
```

**What did you see?**
```
Killed
```

**What did you expect to see?**

ips: 255

**Works when `to = 254` or ` i < to` or not `u8`.
> [!NOTE]
[...]

PR #25442, opened 2025-10-04
perf,cgen: add integer overflow check

kbkpbot:
Feature request by #25379

support integer overflow detect for :

- PostExpr: `i++`, `i--`
- Assignment : `i += 1`, ` i -= 1`, `i *= 1`
- InfixExpr: `i = i + 1`, ` i = i - 1`, `i = i * 1`

usage :

```
v -check-overflow my.v
```

It use compiler builtin for detect integer overflow when perform add/sub/mutl ops.
When it detect an integer overflow, it will panic.

```
$ ./o
   module: builtin
 function: panic_result_not_set()
  message: result not set (attempt to add with overflow)
     file: /media/HD/github/kbkpbot/v/vlib/builtin/builtin.c.v:117
   v hash: 396ef5c
/tmp/v_1000/o.tmp.c:3392:                               | 0x41698d | ./o(builtin__vcurrent_hash+0)
/tmp/v_1000/o.tmp.c:1180:                               | 0x40d104 | ./o()
/tmp/v_1000/o.tmp.c:5364: (d. 3):                       | 0x41d47b | ./o(main__start+0x15d)
/tmp/v_1000/o.tmp.c:5360:                               | 0x41d31b | ./o(main__main+0x17)
/tmp/v_1000/o.tmp.c:5397:                               | 0x41d690 | ./o(main+0x4c)
                                                        | 0x7384baa2a1ca | /lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca)
                                                        | 0x7384baa2a28b | /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b)
                                                        | 0x40bda5 | ./o(_start+0x25)
```

Note: currently, only `gcc`, `clang` provide compiler builtins for integer overflow check, `tcc` and `msvc` use workaround.

### DIFF
--- a/vlib/builtin/builtin_nix.c.v
+++ b/vlib/builtin/builtin_nix.c.v
@@ -21,8 +21,6 @@ fn break_if_debugger_attached() {
 
 @[noreturn]
 pub fn panic_lasterr(base string) {
-	cerrno := C.errno
-	se := unsafe { tos_clone(&u8(C.strerror(cerrno))) }
 	// TODO: use strerror_r and errno
-	panic(base + se)
+	panic(base + ' unknown')
 }


spytheman:
> I don't known how to write a test file for detect panic when running:

Add a .vv/.out pair in `vlib/v/gen/c/testdata/` and test with `v vlib/v/gen/c/coutput_test.v` .

The .out file should have `================ V panic ================` in it, since
`vlib/v/gen/c/coutput_test.v` has a special case for it, that will remove the actual stacktrace/line numbers (which can vary between the compilers/platforms) that follow the panic message.

You can see `vlib/v/slow_tests/inout/panic_on_fixed_array_slice.out` for an example of an .out file with that format:
```
   module: builtin
 function: slice()
  message: array.slice: slice bounds out of range (3 >= 2)
     file: vlib/builtin/array.v
```

spytheman:
> wait for PR #25456 merged, then we can fix `cheaders.v`.

merged

vlib/builtin/overflow.v
spytheman, line 1: Please move it out of `builtin` to a subfolder, for example `vlib/builtin/overflow` - all the vlib/builtin/*.v files are parsed for pretty much every V program, and that feature will be rarely used, so imho it is better to not have to pay the cost for its processing, when it is not.

vlib/builtin/overflow.v
```diff
@@ -0,0 +1,494 @@
...
+	mut res := i8(0)
+	mut is_overflow := false
+	$if gcc || clang {
+		is_overflow = C.__builtin_add_overflow(x, y, &res)
+	} $else {
+		res = x + y
+		is_overflow = (x > 0 && y > 0 && res < 0) || (x < 0 && y < 0 && res > 0)
+	}
+	if is_overflow {
+		panic_result_not_set('attempt to add with overflow(i8(${x}) + i8(${y}))')
```
spytheman: imho just `panic()`

vlib/builtin/overflow.v
spytheman: (and when you move the file, you can also drop the `_overflow_` from the names of the functions in it)

vlib/builtin/overflow.v
```diff
@@ -0,0 +1,494 @@
...
+// mul_overflow_u32 computes `x` * `y` for u32 values, and panic if overflow occurs
+@[inline]
+fn mul_overflow_u32(x u32, y u32) u32 {
+	mut res := u32(0)
+	mut is_overflow := false
+	$if gcc || clang {
+		is_overflow = C.__builtin_mul_overflow(x, y, &res)
+	} $else {
+		res = x * y
+		is_overflow = y != 0 && x > max_u32 / y
```
spytheman: imho for the types with size < 64 bit, you can just do the calculation in 64 bits, then compare the result to the limits for the corresponding type; that way it will be symmetrical for the operands.

spytheman: `is_overflow = y != 0 && x > max_u32 / y` is not symmetrical - it depends on whether y or x was 0 etc

vlib/v/gen/c/assign.v
```diff
@@ -308,11 +308,11 @@ fn (mut g Gen) assign_stmt(node_ ast.AssignStmt) {
...
 		mut cur_indexexpr := -1
 		is_safe_add_assign := node.op == .plus_assign && g.pref.is_check_overflow
-			&& var_type.is_int()
+			&& var_type.is_int() && g.not_in_builtin_overflow_fns()
 		is_safe_sub_assign := node.op == .minus_assign && g.pref.is_check_overflow
-			&& var_type.is_int()
+			&& var_type.is_int() && g.not_in_builtin_overflow_fns()
 		is_safe_mul_assign := node.op == .mult_assign && g.pref.is_check_overflow
-			&& var_type.is_int()
+			&& var_type.is_int() && g.not_in_builtin_overflow_fns()
```
spytheman: this will do the not_in_builtin_overflow_fns() call several times

vlib/v/gen/c/cgen.v
```diff
@@ -8715,3 +8639,45 @@ fn determine_integer_literal_type(node ast.IntegerLiteral) ast.Type {
 		}
 	}
 }
+
+fn (g Gen) not_in_builtin_overflow_fns() bool {
+	if !isnil(g.cur_fn) && g.file.mod.name == 'builtin' && g.cur_fn.name in builtin_overflow_fns {
+		return false
+	}
+	return true
+}
```
spytheman: imho, that can be cached in a `g.` field, and only recalculated once per function in `vlib/`v/gen/c/fn.v:236` .

vlib/builtin/overflow.v
kbkpbot: It seems need a benchmark to eval the performance.

vlib/builtin/overflow.v
kbkpbot: I tried to move it to `vlib/overflow/overflow.v`, and create a new module `module overflow`, and I soon found that if I enable `-check-overflow`, it need an auto import `builtin.overflow` for almost every module, even the `builtin` module itself.
This will create some import cycles:
```
 * builtin -> strings -> builtin.overflow -> builtin
 * main -> builtin -> strings -> builtin.overflow -> builtin
 * strings -> builtin.overflow -> builtin -> strings
 * builtin.closure -> builtin.overflow -> builtin -> strings -> builtin.overflow
 * strconv -> builtin.overflow -> builtin -> strings -> builtin.overflow
 * builtin.overflow -> builtin -> strings -> builtin.overflow
 * math.bits -> builtin.overflow -> builtin -> strings -> builtin.overflow
```

kbkpbot: > I tried to move it to `vlib/overflow/overflow.v`, and create a new module `module overflow`, and I soon found that if I enable `-check-overflow`, it need an auto import `builtin.overflow` for almost every module, even the `builtin` module itself. This will create some import cycles:
> 
> ```
>  * builtin -> strings -> builtin.overflow -> builtin
>  * main -> builtin -> strings -> builtin.overflow -> builtin
>  * strings -> builtin.overflow -> builtin -> strings
>  * builtin.closure -> builtin.overflow -> builtin -> strings -> builtin.overflow
>  * strconv -> builtin.overflow -> builtin -> strings -> builtin.overflow
>  * builtin.overflow -> builtin -> strings -> builtin.overflow
>  * math.bits -> builtin.overflow -> builtin -> strings -> builtin.overflow
> ```

Solved.

StunxFS:
Will this flag be enabled by default in the future?
It might work for debug builds, and could be disabled for `-prod` builds

JalonSolov:
Perhaps enabled by default, with a separate switch to disable it?

Some may want the extra protection in their `-prod` builds, too.

spytheman:
I am inclined to not want it as a default for -prod builds, and probably not want it for debug builds too. It can impact performance of computational hot loops *a lot*, since it essentially wraps every arithmetic operation with function calls that do additional checks.

vlib/v/gen/c/infix.v
```diff
@@ -1306,7 +1322,7 @@ fn (mut g Gen) gen_plain_infix_expr(node ast.InfixExpr) {
 	if is_ctemp_fixed_ret {
 		g.write(', sizeof(${g.styp(node.right_type)}))')
 	}
-	if is_safe_div || is_safe_mod {
+	if true in [is_safe_add, is_safe_sub, is_safe_mul, is_safe_div, is_safe_mod] {
```
spytheman:
```suggestion
	if is_safe_add || is_safe_sub || is_safe_mul || is_safe_div || is_safe_mod {
```

vlib/v/gen/c/infix.v
```diff
@@ -1275,13 +1281,23 @@ fn (mut g Gen) gen_plain_infix_expr(node ast.InfixExpr) {
 		g.write('memcmp(')
 	}
 	mut opstr := node.op.str()
-	if is_safe_div || is_safe_mod {
-		vsafe_fn_name := if is_safe_div { 'VSAFE_DIV_${typ_str}' } else { 'VSAFE_MOD_${typ_str}' }
+	if true in [is_safe_add, is_safe_sub, is_safe_mul, is_safe_div, is_safe_mod] {
```
spytheman:
```suggestion
	if is_safe_add || is_safe_sub || is_safe_mul || is_safe_div || is_safe_mod {
```

JalonSolov:
Well... on by default plays into the V claim... "Simple, fast, **safe**, compiled."

vlib/builtin/overflow/overflow_test.v
spytheman, line 1: Given that is a slightly modified copy of vlib/v/gen/c/coutput_test.v,
there are special cases here that will never happen for the generated .vv/.out files.
It also supports general options that will always be true (the options inside the .vv files). 

I think it can be simplified a lot.

vlib/v/markused/markused.v
```diff
@@ -142,6 +142,32 @@ pub fn mark_used(mut table ast.Table, mut pref_ pref.Preferences, ast_files []&a
 		if pref_.should_use_segfault_handler() {
 			core_fns << 'v_segmentation_fault_handler'
 		}
+		if pref_.is_check_overflow {
+			core_fns << 'builtin.overflow.add_i8'
```
spytheman: would not 3 nested loops + interpolation be shorter and less error prone?

spytheman:
> Well... on by default plays into the V claim... "Simple, fast, safe, compiled."

I want to be able to iterate fast on prototypes. To do that, I want a compiler, that can produce code that is not pessimized by default. I tend to relatively often have out of bounds errors in that mode, but nearly never overflow related ones. -> For me, a default that does OOB checks, but not overflow ones makes perfect sense.

For others, that balance may well be different 🤷🏻‍♂️.
For all I care, they can go use Rust or some other language, *or* they can just turn on the options that make V produce more checks (and slower code) in their VFLAGS.

spytheman:
The argument is similar to why we have `-div-by-zero-is-zero` not on by default.

spytheman:
> Or we can set `-check-overflow` for `v test`, and this will help identify vlib's overflow bug.

That is a very good idea, but imho should be done in another PR, after the feature is merged.

medvednikov:
Great job!

I also think that no check should be the default, check with -check-overflow or a module/function attribute.

PR #25449, opened 2025-10-05
x.crypto.chacha20: mark `State` as public to make it a reusable thing

blackshirt:
While have a working prototype implementation of `ChaCha20-Poly1305-PSIV` construct, its used an internal of `chacha20` by rewriting it as a new routine, makes it a duplicated in a lot places. I think its should be a public to make it reusable thing. See the working implementation of this construct at [here](https://github.com/blackshirt/v/blob/psiv/vlib/x/crypto/chacha20poly1305/psiv.v).
By making it accessible outside the module, its make possible to expand by new construction based on this state, like above psiv construct or possibility of [this](https://arxiv.org/pdf/2507.18157v11). This also mean reduces code duplication.

This patch was related to previously merged PR [#25438](https://github.com/vlang/v/pull/25438), but comes with the changes on the `chacha20` sides..
Its exposes some cores of the `chacha20` as public in the means of :

- exposes a `chacha20.State` into user, mainly as a new building-block tool
- exposes minimal api for this state, currently only `.clone`, `.reset` and fundamental quarter round `.qround`  routines.
- adjustment on the State-related thing.

There are construct availables outside that relies on `hchacha20`, but its not the concerns of this PR.

Thanks
Cheers

### DIFF
--- a/vlib/x/crypto/chacha20/chacha_test.v
+++ b/vlib/x/crypto/chacha20/chacha_test.v
@@ -2,6 +2,39 @@ import rand
 import encoding.hex
 import x.crypto.chacha20
 
+// test for chacha20.State exposed api. The test material was taken from
+// Test Vector for the ChaCha20 Block Function ch 2.3.2. from RFC 8439 doc.
+// See https://datatracker.ietf.org/doc/html/rfc8439#section-2.3.2
+fn test_chacha20_state() ! {
+	// ChaCha state with the key setup.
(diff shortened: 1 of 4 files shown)


blackshirt:
> If you expose new APIs, please also add tests for them.
> 
> Otherwise people can start to depend on them, and we would not know if/when they regress.

Okey..i would add it 

NB; add it in [#8bca962](https://github.com/vlang/v/pull/25449/commits/8bca962b8f9e905863879c5aebbc3d679cf0d420)

spytheman (approved):
Thank you @blackshirt 🙇🏻 .

Issue #25455, opened 2025-10-07
output a muti-line string ending with slash


kbkpbot:
### Describe the bug

How can I output a muti-line string which ending with a `\`?

For example :
```
line1 \
line2 \
line3
```

### Reproduction Steps

```v
str := r'
line1 \
line2 \
line3
'
dump(str)
```

### Expected Behavior

output 
```
line1 \
line2 \
line3
```

### Current Behavior

```
$ ./m
[m.v:6] str:
line1 line2 line3


```

### V version

V 0.4.12 3176e7a0458bb2f522696f0f0bd5fdc14339c40e.1aa2dc3

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 3176e7a0458bb2f522696f0f0bd5fdc14339c40e.1aa2dc3
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |12.32GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-10-06 14:20:33
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/overflow
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.40-24-gfcff7a17
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.6) 2.39

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

[...]

PR #25456, opened 2025-10-07
scanner: fix raw string ending withs backslash

kbkpbot:
Fix issue #25455 

In raw string, should not remove the endings backslash.


spytheman (approved):
Excellent work.
Thank you.

PR #25472, opened 2025-10-10
x.json2: replace x.json2.decode() with x.json2.decoder2.decode()

dy-tea:
Original pr with `@[deprecated_after]` moved to #25480

### DIFF
--- a/vlib/x/json2/decoder2/stub.v
+++ b/vlib/x/json2/decoder2/stub.v
@@ -1,0 +1,15 @@
+module decoder2
+
+import x.json2
+
+@[deprecated: '`decode` has been moved to `x.json2`, use `decode` from `x.json2` instead']
+@[deprecated_after: '2025-10-12']
+fn decode[T](val string) !T {
+	return json2.decode[T](val)
+}
+
+@[deprecated: '`decode` has been moved to `x.json2`, use `decode` from `x.json2` instead']
+@[deprecated_after: '2025-03-18']
+fn decode_array[T](src string) !T {
+	return json2.decode[T](src)
+}
--- a/vlib/x/json2/decoder_deprecated.v
+++ b/vlib/x/json2/decoder_deprecated.v
@@ -1,5 +1,6 @@
 module json2
 
+// Decodes a JSON string into an `Any` type. Returns an option.
 @[deprecated: 'use `decode[json2.Any]` instead']
 @[deprecated_after: '2025-10-10']
 pub fn raw_decode(src string) !Any {
--- a/vlib/x/json2/tests/decode_option_field_test.v
+++ b/vlib/x/json2/tests/decode_option_field_test.v
@@ -11,7 +11,7 @@ mut:
 
 fn test_main() {
 	resp := '{"name": "Bob", "age": 20, "birthday": "2025-10-12 10:14:52"}'
-	person := decoder2.decode[Person](resp)!
+	person := json2.decode[Person](resp)!
 	assert '${person}' == "Person{
     name: 'Bob'
     age: Option(20)


dy-tea:
Ah that's an issue, now that the deprecation is working it's giving errors in all those cases.

JalonSolov:
Rip 'em out.  @spytheman does it manually, occasionally - we don't have automation to remove completely deprecated things.

dy-tea:
I will rip it out and just add it to this pr then.

dy-tea:
`vlib/x/json2/tests/decode_map_of_map_test.v` appears to have a runtime error when using decoder2.
```
/tmp/v_1000/tsession_7ff32091a740_01K77ZGP4W5NGGYPW048KY0JGP/decode_struct_test.01K77ZM2EBP6GNEWXZC:19137: at x__json2__decoder2__Decoder_decode_string_T_string: RUNTIME ERROR: invalid memory access
/tmp/v_1000/tsession_7ff32091a740_01K77ZGP4W5NGGYPW048KY0JGP/decode_struct_test.01K77ZM2EBP6GNEWXZC:16981: by x__json2__decoder2__Decoder_decode_value_T_string
/tmp/v_1000/tsession_7ff32091a740_01K77ZGP4W5NGGYPW048KY0JGP/decode_struct_test.01K77ZM2EBP6GNEWXZC:18839: by x__json2__decoder2__Decoder_decode_value_T_main__StructTypeSub
/tmp/v_1000/tsession_7ff32091a740_01K77ZGP4W5NGGYPW048KY0JGP/decode_struct_test.01K77ZM2EBP6GNEWXZC:12439: by x__json2__decoder2__Decoder_decode_value_T_main__StructType_T_main__StructTypeSub
/tmp/v_1000/tsession_7ff32091a740_01K77ZGP4W5NGGYPW048KY0JGP/decode_struct_test.01K77ZM2EBP6GNEWXZC:10796: by x__json2__decoder2__decode_T_main__StructType_T_main__StructTypeSub
/tmp/v_1000/tsession_7ff32091a740_01K77ZGP4W5NGGYPW048KY0JGP/decode_struct_test.01K77ZM2EBP6GNEWXZC:21331: by main__test_types
/tmp/v_1000/tsession_7ff32091a740_01K77ZGP4W5NGGYPW048KY0JGP/decode_struct_test.01K77ZM2EBP6GNEWXZC:22646: by main
```

dy-tea:
Potentially could move decoder2 up a directory to json2 (since the tooling also depends on it).

JalonSolov:
Separate PR.  Only change what's needed for the intent of the PR each time.  Keeps them smaller, as well as making them quicker/easier to review.

Moving those up a level will definitely deserve a separate PR.  You have to move the current files up, create "stub" files that give a deprecation message, and import the files that were moved, etc., etc.

spytheman:
Imho it should always require the deprecated tag, because of the message; using just `deprecated_after` does not provide a way to add one

spytheman:
> Didn't add a test as this would change depending on the date which would be tricky.

use a far future deprecation date or a date in the past, depending on what you need to test

spytheman:
There is a script `./cmd/tools/show_ancient_deprecations.v` that can be used to show what needs to be cleaned up, which I run ~3-4 times/year.

spytheman:
I think that this PR should be split - the parser change and a test for it should be 1, and the json changes should be an entirely separate PR (potentially first to be merged).

spytheman:
> Only change what's needed for the intent of the PR each time. Keeps them smaller, as well as making them quicker/easier to review.

yes, 100%

spytheman:
@Larsimusrex what do you think about the x.json2 changes?

Larsimusrex:
So it just removes `json2.decode`? It doesn't really matter to me as I'm working on decoder2 anyways, but it might be confusing/inconvenient for the people who use json2. It would be better to just replace it with decoder2 outright, I just hadn't had time to do that yet. Also for the json2 tests most of them a redundant, as they were already rewritten for decoder2.

dy-tea:
> I think that this PR should be split - the parser change and a test for it should be 1, and the json changes should be an entirely separate PR (potentially first to be merged).

Yes I can do the json changes in a seperate pr but they will have to be merged first (as the json decode function is a hard error).

> So it just removes `json2.decode`? It doesn't really matter to me as I'm working on decoder2 anyways, but it might be confusing/inconvenient for the people who use json2. It would be better to just replace it with decoder2 outright, I just hadn't had time to do that yet. Also for the json2 tests most of them a redundant, as they were already rewritten for decoder2.

I think the best thing would be to replace `json2.decode` outright, some tests use `json2.raw_decode` and there is one that creates a Parser so those need to be changed (or removed). 

What I will do for now is remove the `@[deprecated_after]` commit from this pr and open another one (although it is only a single line change it makes it clearer). Then we can discuss about whether or not the `@[deprecated]` tag should be a hard requirement for `@[deprecated_after]` then.

Edit: opened a seperate pr for `@[deprecated_after]` here #25480.

dy-tea:
Docs CI is failing because of a bug in decoder2:
https://play.vlang.io/p/165805379c

Will open an issue.

enghitalo:
IMHO, in this case, just replacing it is much better

dy-tea:
> IMHO, in this case, just replacing it is much better

Yeah that's what I'm going ahead and doing right now as it seemed the most sensible option.

vlib/x/json2/tests/decode_map_of_map_test.v
```diff
@@ -5,9 +5,6 @@ struct CrossVerifyResult {
 }
 
 fn test_main() {
-	x := json2.decode[CrossVerifyResult]('') or {
-		assert err.msg().contains('invalid token')
-		return
-	}
+	x := json2.decode[CrossVerifyResult]('') or { return }
```
spytheman: is the err message here the same as before, or different, but does not matter now?

dy-tea: It's different, I could probably update it with the new one but removing it was simply faster. 
Might update it with the new message when I have time.

examples/get_weather/get_weather.v
```diff
@@ -2,7 +2,7 @@ import net.http
 import os
 import rand
 import x.json2
-import x.json2.decoder2 as json
+import x.json2 as json
```
JalonSolov, line 4: 2 separate imports for json2?

JalonSolov:
The stub module isn't in the spirit of deprecations... a (possibly?) better way would be to make that module import json2 itself, then call the functions from there, instead of returning an error.

Just returning an error breaks compiles as badly as removing routines directly.

Also, it might be good to choose a new deprecation date (as in today), since nobody has been notified about the previous date before, and it is already past the date where those deprecations would change from warnings to errors - another hard break in compiles.

vlib/x/json2/decoder_deprecated.v
```diff
@@ -1,104 +1,14 @@
...
-// code returns the error code of the UnknownTokenError
-@[deprecated_after: '2025-03-18']
-pub fn (err UnknownTokenError) code() int {
-	return 1
-}
-
-// msg returns the error message of the UnknownTokenError
+// decode_array is a generic function that decodes a JSON string into the array target type.
+@[deprecated: 'use `decode` instead']
 @[deprecated_after: '2025-03-18']
```
JalonSolov: Date change, perhaps?

examples/get_weather/get_weather.v
dy-tea: Ah well-spotted. It must have been left over from decoder2, will change it.

vlib/x/json2/decoder_deprecated.v
dy-tea: Check the bottom of the file, I didn't change the deprecation date. I could potentially remove it even.

dy-tea: Removed it completely now, the reason I am keeping `raw_decode` is because people would not get the deprecation warning at all due to the original bug.

dy-tea:
Will wait for #25489 as this should fix the docs CI and then I can update the test in there to use the updated decoder2.

dy-tea:
Rebased, CI should pass now.

spytheman:
@dy-tea thank you 🙇🏻

vlib/x/json2/decoder2/stub.v
```diff
@@ -0,0 +1,15 @@
+module decoder2
+
+import x.json2
+
+@[deprecated: '`decode` has been moved to `x.json2`, use `decode` from `x.json2` instead']
+@[deprecated_after: '2025-10-12']
+fn decode[T](val string) !T {
```
spytheman, line 7: there is no need to deprecate a non `pub` function (since people were not able to depend on it)

vlib/x/json2/decoder2/stub.v
```diff
@@ -0,0 +1,15 @@
...
+	return json2.decode[T](val)
+}
+
+@[deprecated: '`decode` has been moved to `x.json2`, use `decode` from `x.json2` instead']
+@[deprecated_after: '2025-03-18']
```
spytheman, line 12: should the date here be `2025-03-18` or the current one?

vlib/x/json2/decoder2/stub.v
dy-tea: Ah that's right.

vlib/x/json2/decoder2/stub.v
dy-tea: I believe it was deprecated earlier, I just moved it to stub.v

PR #25479, opened 2025-10-11
v: add array fixed `filter()` method

felipensp:


vlib/v/tests/array_fixed_filter_test.v
```diff
@@ -0,0 +1,4 @@
+fn test_main() {
+	a := [1, 2, 3]!
+	assert a.filter(it > 2).len == 1
```
Delta456, line 3: Does this return a fixed array or dynamic array?

spytheman: it is a dynamic one, this passes:

```suggestion
	assert a.filter(it > 2).len == 1
	assert typeof(a.filter(it > 2)).name == '[]int'
```

PR #25483, opened 2025-10-11
native: unify registers, refactor

Eliyaan:
- replace Register = Amd64Regs | Arm64Regs by an enum : reg0, reg1...
- change the names of the function in the CodeGen interface to start with `cg_` to have a visible separation between target dependant and target independant functions

This improves/will improve support for other targets than amd64. (a lot of code was written with Amd64 registers and it will now be possible to generalize it for multiple targets). 
A future PR will move out some function from the CodeGen interface for different targets to be able to share the same functions (some functions in the interface are too general). It is not in this PR because it would add a lot of complexity to the already huge PR.

### DIFF
--- a/vlib/v/gen/native/gen.v
+++ b/vlib/v/gen/native/gen.v
@@ -119,14 +119,14 @@ mut:
 	cg_gen_asm_stmt(asm_node ast.AsmStmt)
 	cg_gen_cast_expr(expr ast.CastExpr)
 	cg_gen_exit(expr ast.Expr)
-	cg_gen_index_expr(ast.IndexExpr)     
+	cg_gen_index_expr(ast.IndexExpr)
 	cg_gen_match_expr(expr ast.MatchExpr)
 	cg_gen_print_reg(r Register, n i32, fd i32)
 	cg_gen_print(s string, fd i32)
 	cg_gen_syscall(node ast.CallExpr)
 	cg_inc_var(var Var, config VarConfig)
 	cg_infix_expr(node ast.InfixExpr)
-	cg_infloop()                     
+	cg_infloop()
 	cg_init_struct(var Var, init ast.StructInit)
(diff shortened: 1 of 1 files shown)


vlib/v/gen/native/gen.v
```diff
@@ -147,14 +147,14 @@ mut:
...
 	cg_patch_relative_jmp(pos i32, addr i64)
 	cg_pop(r Register)
-	cg_prefix_expr(node ast.PrefixExpr) // TODO: make platform-independant
+	cg_prefix_expr(node ast.PrefixExpr)
 	cg_push(r Register)
 	cg_ret()
-	cg_return_stmt(node ast.Return) // TODO: make platform-independant
-	cg_reverse_string(r Register)   // TODO: remove
+	cg_return_stmt(node ast.Return)
+	cg_reverse_string(r Register)
```
JalonSolov:
```suggestion
	cg_reverse_string(r Register)
```

JalonSolov:
Very nice!  Looks a lot better this way, with more "generic" handling in the code, instead of using the names of x86 registers.

vlib/v/gen/native/builtins.v
```diff
@@ -29,21 +29,21 @@ pub fn (mut g Gen) init_builtins() {
 		.int_to_string:  BuiltinFn{
 			// 32-bit signed integer to string conversion
 			body:     fn (builtin BuiltinFn, mut g Gen) {
-				g.code_gen.convert_int_to_string(builtin.arg_regs[0], builtin.arg_regs[1])
+				g.cg.cg_convert_int_to_string(builtin.arg_regs[0], builtin.arg_regs[1])
```
spytheman, line 32: I do not like the .cg.cg_ repetition, but I guess it will be cleaned up in later PRs.

PR #25490, opened 2025-10-12
token,scanner,parser,checker,cgen: add file_idx to token.Pos

kbkpbot:
This is a fix for PR #25429
change the `file_idx` to `i16` and `col` to `u16` to keep the same size  of `token.Pos`.
I ran a local `v run .github/workflows/compare_pr_to_master.v`, and see almost no performance impact.

BTW, integrate the `file_idx` into `token.Pos` is help for identify which file this token is come from. In `vls`, the `goto definition` function need the filename when it try to jump to the location.

### DIFF
--- a/vlib/v/scanner/scanner.v
+++ b/vlib/v/scanner/scanner.v
@@ -949,13 +949,11 @@ pub fn (mut s Scanner) text_scan() token.Token {
 					// treat shebang line (#!) as a comment
 					comment := s.text[start - 1..s.pos].trim_space()
 					if s.line_nr != 1 {
-						col := s.current_column() - comment.len
-						u16_col := if col < 0 { u16(1) } else { u16(col) }
 						comment_pos := token.Pos{
 							line_nr:  s.line_nr - 1
 							len:      comment.len
 							pos:      start
-							col:      u16_col
+							col:      u16_col(s.current_column() - comment.len)
 							file_idx: s.file_idx
 						}
 						s.error_with_pos('a shebang is only valid at the top of the file',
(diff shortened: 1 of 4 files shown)


vlib/v/scanner/scanner.v
```diff
@@ -1511,9 +1519,10 @@ fn trim_slash_line_break(s string) string {
 ///   escaped utf8 runes in octal like `\342\230\205` => (★)
 pub fn (mut s Scanner) ident_char() string {
 	lspos := token.Pos{
-		line_nr: s.line_nr
-		pos:     s.pos
-		col:     s.pos - s.last_nl_pos - 1
+		line_nr:  s.line_nr
+		pos:      s.pos
+		col:      u16(s.pos - s.last_nl_pos - 1)
```
spytheman, line 1524: why not `u16_col`?

vlib/v/scanner/scanner.v
```diff
@@ -1193,9 +1200,10 @@ pub fn (mut s Scanner) ident_string() string {
 		s.is_nested_string = false
 	}
 	lspos := token.Pos{
-		line_nr: s.line_nr
-		pos:     s.pos
-		col:     s.pos - s.last_nl_pos - 1
+		line_nr:  s.line_nr
+		pos:      s.pos
+		col:      u16(s.pos - s.last_nl_pos - 1)
```
spytheman, line 1205:
```suggestion
		col:      u16_col(s.pos - s.last_nl_pos - 1)
```

vlib/v/scanner/scanner.v
kbkpbot: `col` sometimes can be a negative value....

spytheman: hm, would not a negative `col` be a symptom of a bug?

spytheman: afaik `u16_col()` just like `u16()` will also always return a positive value, the difference is that `u16_col()` will return 0 for all the potentially negative inputs, while `u16()` will turn them into high positive values.

PR #25492, opened 2025-10-13
parser,ast: add vls info for support vls goto definition

kbkpbot:
This PR will enable access all kind of declares in `checker`.
This is useful for `vls` goto definition where we need to known the location of declaration, also get the comment as help information.

### DIFF
--- a/vlib/v/parser/parser.v
+++ b/vlib/v/parser/parser.v
@@ -2897,8 +2897,8 @@ fn (mut p Parser) type_decl() ast.TypeDecl {
 			name_pos:      name_pos
 			is_markused:   attrs.contains('markused')
 		}
+		p.table.register_sumtype(node)
 		if p.pref.is_vls {
-			p.table.register_sumtype(node)
 			key := 'sumtype_${p.prepend_mod(name)}'
 			val := ast.VLSInfo{
 				pos: node.pos


vlib/v/parser/v_parser_test.v
```diff
@@ -314,3 +314,54 @@ fn test_parse_with_stdout() {
...
+	mut prog := parse_text(source_text, '', mut table, .skip_comments, vpref)
+	mut checker_ := checker.new_checker(table, vpref)
+	checker_.check(mut prog)
+
+	assert 'main.my_const' in table.const_decls
+	assert 'main.MyS' in table.struct_decls
+	assert 'main.MyInterface' in table.interface_decls
+	assert 'main.MyS.add' in table.fn_decls // MyS method
+	assert 'main.foo' in table.fn_decls
+	assert 'my_global' in table.global_decls
```
spytheman: Why is the global not prefixed by `main.` ?
The const and everything else is.

spytheman:
@kbkpbot if you store just token.Pos values in the maps, it will be cheaper (in terms of RAM usage) and you can still use them for vls's purposes

vlib/v/parser/v_parser_test.v
kbkpbot: I suppose that all `__global` in a project should has same name space.
Is that correct?

spytheman: No. V globals are namespaced by their module name, just like constants are.

(edit: currently they *are* in a single namespace, but they should not remain that way - the current state is the buggy one, since it negates the benefit that modules give)

spytheman: @medvednikov should globals have a common namespace, or should they be prefixed by their module name in the generated code?

JalonSolov: People have asked more than once for "module specific globals", or, basically not-really-globals-but-global-to-the-module.

Or, worded another way, module level variables, namespaced variable, or whatever else you want to call them.

vlib/v/parser/enum.v
```diff
@@ -253,6 +253,12 @@ fn (mut p Parser) enum_decl() ast.EnumDecl {
 
 	if !already_exists {
 		p.table.register_enum_decl(enum_decl)
+		key := 'enum_${name}'
```
StunxFS: Why instead of adding the `enum_` prefix to the name, you add one more field to the struct `VLSInfo` that is of type TypeKind (I don't remember the exact name)?

So just use `kind: .enum`

spytheman: it is part of the key, because it will be used for fast lookup later in the checker (in the part that interfaces with the queries that vls makes)

vlib/v/ast/table.v
```diff
@@ -2765,3 +2773,8 @@ pub fn (mut t Table) get_veb_result_type_idx() int {
 	t.veb_res_idx_cache = t.find_type('veb.Result')
 	return t.veb_res_idx_cache
 }
+
+@[inline]
+pub fn (mut t Table) register_vls_decl(key string, val VLSInfo) {
+	t.vls_decls[key] = val
```
spytheman: Can this be a nop in normal compilations (without -vls-mode) ?

kbkpbot: In `table`, we can't access the `pref`...
Instead, I removed the calls to `register_vls_decl` when not in `vls` in `parser`.

vlib/v/parser/v_parser_test.v
kbkpbot: OK, I add both : module specific global and a real `global`.

vlib/v/parser/fn.v
```diff
@@ -753,6 +753,19 @@ run them via `v file.v` instead',
 		p.table.register_fn_generic_types(fn_decl.fkey())
 	}
 	p.label_names = []
+	type_str := if (is_method || is_static_type_method) && rec.typ != ast.no_type {
+		p.table.sym(rec.typ.idx_type()).name.all_after_last('.')
+	} else {
+		''
+	}
```
spytheman: that should be in the `if p.pref.is_vls {` block too

PR #25497, opened 2025-10-14
cgen: cleanup assign_stmt

felipensp:


spytheman:
Thank you @felipensp .

PR #25503, opened 2025-10-14
parser: add doc in VLSInfo

kbkpbot:
Add documentation in `VLSInfo`, this may help `vls` get help information about fn/struct/..., etc


spytheman (approved):
Excellent work.
Do you have a way to test it on the vls side too @kbkpbot ?

PR #25510, opened 2025-10-15
cgen, type_resolver: Fix resolution type Map[string]T  when T is AliasToMap

### LINKED ISSUE #25494: json2.encode with type alias parameter cause compile error
<details>
<summary>V version: V 0.4.12 1d9c583, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 9772649250016a8558fe9bf07ba94f2c812fa9ef.1d9c583
|OS                  |macos, macOS, 15.6.1, 24G90
|Processor           |6 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8500 CPU @ 3.00GHz
|Memory              |0.6GB/16GB
|                    |
|V executable        |/Users/11048490/lang/v/v
|V last modified time|2025-10-10 06:22:58
|                    |
|V home dir          |OK, value: /Users/11048490/lang/v
|VMODULES            |OK, value: /Users/11048490/.vmodules
|VTMP                |OK, value: /tmp/v_1922266506
|Current working dir |contains these non ASCII characters: [`代`, `码`], value: /Users/11048490/代码/v_test
|                    |
|Git version         |git version 2.50.1 (Apple Git-155)
|V git status        |0.4.12-82-g1d9c5833
|.git/config present |true
|                    |
|cc version          |Apple clang version 17.0.0 (clang-1700.3.19.1)
|gcc version         |Apple clang version 17.0.0 (clang-1700.3.19.1)
|clang version       |Apple clang version 17.0.0 (clang-1700.3.19.1)
|tcc version         |tcc version 0.9.27 (x86_64 Darwin)
|tcc git status      |thirdparty-macos-amd64 975f1ad8
|emcc version        |N/A
|glibc version       |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  main.v && main`
```v
import x.json2

pub struct AppJsonValue {
pub:
	file  string
	md5   string
	creat i64
}

pub type AppJson = map[string]AppJsonValue

fn main() {
	// no problem
	app_json1 := map[string]AppJsonValue{}
	res1 := json2.encode(app_json1)
	println('res is ${res1}')

	// compile error
	app_json2 := AppJson{}
	res2 := json2.encode(app_json2)
	println('res is ${res2}')
}

```

**What did you see?**
```
/Users/11048490/lang/v/vlib/x/json2/encode.v:73:22: error: cannot use `AppJson` as `map[string]AppJson` in argument 1 to `x.json2.Encoder.encode_map`
   71 |         encoder.encode_array(val)
   72 |     } $else $if T.unaliased_typ is $map {
   73 |         encoder.encode_map(val)
      |                            ~~~
   74 |     } $else $if T.unaliased_typ is $enum {
   75 |         encoder.encode_enum(val)
```

> [!NOTE]
>

felipensp:
Fix #25494

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -2672,10 +2672,6 @@ fn (mut c Checker) method_call(mut node ast.CallExpr, mut continue_check &bool) 
 				&& c.check_basic(c.table.value_type(got_arg_typ).clear_flag(.option), c.table.value_type(exp_arg_typ)) {
 				continue
 			}
-			if c.table.type_to_str(c.table.unaliased_type(got_arg_typ)) == c.table.type_to_str(exp_arg_typ) {
-				continue
-			}
-			println('>>> ${c.table.type_to_str(c.table.unaliased_type(got_arg_typ))} ${c.table.type_to_str(exp_arg_typ)}')
 			c.error('${err.msg()} in argument ${i + 1} to `${left_sym.name}.${method_name}`',
 				arg.pos)
 		}


vlib/v/gen/c/for.v
spytheman, line 1: I am surprised that this was needed; the test itself does not have for loops, but I suppose that they are in the generic implementation in x.json2 🤔 .

Issue #25515, opened 2025-10-16
parser: parse_files() with .parse_comments arg will cause parse vlib fail with errors


kbkpbot:
### Describe the bug

The `parse_files()` function in `parser`, can't accept a `.parse_comments`(currently it is hard code to `.skip_comments`), if set, then parse the vlib will generate errors.

### Reproduction Steps

in `vlib/v/parser/parser.v`
https://github.com/vlang/v/blob/e5a5f027c108ac2c2f689898d49970fc2cdddfd1/vlib/v/parser/parser.v#L385-L400

modify 
```v
files << parse_file(path, mut table, .skip_comments, pref_)
```
to
```v
files << parse_file(path, mut table, .parse_comments, pref_)
```

then compile v self 
```
v self -g
```

run a test on vlib
```
v test vlib
```

### Expected Behavior

v test pass with success

### Current Behavior

```
$ v test vlib
cannot compile `/media/HD/github/kbkpbot/v/cmd/tools/vtest.v`: 1
vlib/strconv/atoi.v:264:1: error: unreachable code
  262 |             }
  263 |             underscored = true
  264 |             continue // Skip underscore
      | ~~~~~~~~~~~~~~~~~~
  265 |         } else {
  266 |             if c > 9 {
vlib/strconv/atoi.v:321:1: error: unreachable code
  319 |             }
  320 |             underscored = true
  321 |             continue // Skip underscore
      | ~~~~~~~~~~~~~~~~~~
  322 |         } else {
  323 |             if c > 9 {
vlib/strconv/atou.v:51:1: error: unreachable code
   49 |             }
   50 |             underscored = true
   51 |             continue // Skip underscore
      | ~~~~~~~~~~~~~~~~~~
   52 |         } else {
   53 |             if c > 9 {
vlib/builtin/builtin.c.v:704:1: error: unreachable code
  702 |         }
  703 |     }
  704 |     return &u8(unsafe { nil }) // not reached, TODO: remove when V's checker is improved
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  705 | }
  706 |
vlib/builtin/builtin.c.v:731:1: error: unreachable code
  729 |         return unsafe { vcalloc(n) }
  730 |     }
  731 |     return &u8(unsafe { nil }) // not reached, TODO: remove when V's checker is improved
[...]

PR #25524, opened 2025-10-17
parser,checker: autocomplete cleanup

kbkpbot:
Move some autocomplete logic from `vls` to `v`, this may simplify `vls` server design.(https://github.com/vlang/vls/pull/467)
And,
1. code cleanup, generate the json result in `v`
2. fix `p.is_vls` to `p.pref.is_vls`, as `p.is_vls` is only valid when process the edited file, not other files. we need `vlib` comments.
https://github.com/vlang/v/blob/e5a5f027c108ac2c2f689898d49970fc2cdddfd1/vlib/v/parser/parser.v#L266-L272
3. add test

Due to issue #25515, currently `vlib` is processed with `.skip_comments`, so comment is not work now for `vls`.

### DIFF
--- a/vlib/v/tests/vls/autocomplete_module_test.v
+++ b/vlib/v/tests/vls/autocomplete_module_test.v
@@ -44,16 +44,21 @@ fn test_main() {
 			println('fail execute ${t.cmd}')
 			panic(res.output)
 		}
-		if t.output != res.output {
+		res_output := $if windows {
+			res.output.replace('\r\n', '\n')
+		} $else {
+			res.output
+		}
+		if t.output != res_output {
 			println('${term.red('FAIL')} ${t.cmd}')
-			if diff_ := diff.compare_text(t.output, res.output) {
+			if diff_ := diff.compare_text(t.output, res_output) {
 				println(term.header('difference:', '-'))
 				println(diff_)
 			} else {
 				println(term.header('expected:', '-'))
 				println(t.output)
 				println(term.header('found:', '-'))
-				println(res.output)
+				println(res_output)
 			}
 			println(term.h_divider('-'))
 			total_errors++


vlib/v/parser/enum.v
```diff
@@ -48,7 +48,7 @@ fn (mut p Parser) enum_decl() ast.EnumDecl {
 	}
 	p.check(.key_enum)
 	end_pos := p.tok.pos()
-	mut comments_before_key_enum := if p.is_vls {
+	mut comments_before_key_enum := if p.pref.is_vls {
```
spytheman, line 51: Why go back to `p.pref.is_vls` ?

spytheman: ~~imho revert all the changes in `vlib/v/parser/` - they are not substantial for this PR.~~

spytheman: > fix p.is_vls to p.pref.is_vls, as p.is_vls is only valid when process the edited file, not other files. we need vlib comments.

I think that having `p.is_vls` and pref.is_vls being named the same, but meaning slightly different things is very confusing and thus creates the conditions for future bugs.

I think that `is_vls` in Parser should be named something like `is_edited_vls_file`, to make it clear, that it is not dependent solely on the command line flag.

vlib/v/tests/vls/autocomplete_module_test.v
```diff
@@ -0,0 +1,65 @@
...
+}
+
+struct TestData {
+	cmd    string
+	output string
+}
+
+const test_data = [
+	TestData{
+		cmd:    'v -check -json-errors -nocolor -vls-mode -line-info "${text_file}:18:3" ${text_file}'
```
spytheman:
```suggestion
		cmd:    'v -check -json-errors -nocolor -vls-mode -line-info "${text_file}:18:3" ${os.quoted_path(text_file)}'
```

vlib/v/tests/vls/autocomplete_module_test.v
```diff
@@ -0,0 +1,65 @@
+import os
+import term
+import v.util.diff
+
+const vroot = os.real_path(@VMODROOT)
+
+const text_file_orig = os.join_path(vroot, 'vlib', 'v', 'tests', 'vls', 'sample_text.vvv')
```
spytheman: why .vvv instead of .vv ?

vlib/v/parser/enum.v
kbkpbot: `p.is_vls` is only valid when process the edited file, not other files. we need `vlib` comments.

```v
 mut p := Parser{ 
 	scanner: scanner.new_scanner_file(path, file_idx, comments_mode, pref_) or { panic(err) } 
 	table:   table 
 	pref:    pref_ 
 	// Only set vls mode if it's the file the user requested via `v -vls-mode file.v` 
 	// Otherwise we'd be parsing entire stdlib in vls mode 
 	is_vls:   pref_.is_vls && path == pref_.path 
```

spytheman:
Why use `|` instead of `:` ?
That is a symbol that can be confused for piping in most shells, and unlike `:` is not used by any editor for specifying an exact location...

spytheman:
> Maybe a better solution is just pass a string to vls server, and then vls just put it it the middle of json response ?

What will be the content of the string?
Why invent a bespoke serialization format/protocol, if vls has to use a json parser anyway, because of the need to handle jsonrpc for communication with the LS client?

spytheman:
just use `all_before_last(':')` instead of split ... there is no need to change the separator

spytheman:
I do not want Windows idiosyncrasies to dictate what the parameters will be, and `path:line:col` is an established convention that should not be broken without a really good reason - it is very helpful for jumping to locations in all kinds of editors/IDEs.

spytheman:
you can also `os.file_name(path)` , and then split the result on `:` to get the line and column, in order to avoid the start, that can have a `c:\` prefix on windows

JalonSolov:
https://modules.vlang.io/builtin.html#string.trim_space_right

If you need to output the lines, add whatever EOL is needed for the system.

spytheman (approved):
Excellent work.

Thank you @kbkpbot 🙇🏻.

PR #25529, opened 2025-10-18
semver: fix parsing on `major.minor.patch.pre-release` standard

### LINKED ISSUE #25528: git version parsing error
### Describe the bug

When I did "v upgrade" or "v list", I got the following error:
> error: Invalid version format for input "2.51.1.dirty"

### Reproduction Steps

 I am on CachyOS (ArchLinux). I have the latest git package installed (as of this day).
> $ git --version
> git version 2.51.1.dirty

### Expected Behavior

"v upgrade" and "v list" commands, should work, as usual.

### Current Behavior

> $ v upgrade 
> error: Invalid version format for input "2.51.1.dirty"

### Possible Solution

I needed to downgrade my git package to a previous version. Now it reads as follows
> $ git --version
> git version 2.51.0

After the git package downgrade, v started working.

Now, the "v upgrade" and "v list" commands, work as intended.

### V version

V 0.4.12 1a4cf3b

 OS: CachyOS Linux (ArchLinux variant: rolling, latest package updates)

> [!NOTE]
>

felipensp:
Fix #25528

semver EBNF establish
image
but git version on ArchLinux uses `2.51.1.dirty` version name.


larpon:
Can we not just ignore everything after (and including) a third dot?

JalonSolov:
The simplest way to handle it would be to ignore everything after the 3rd set of digits, regardless of what's there.

git's version (and far, far too many other programs out there) is not a valid semver format, so it will never work to check it against semver.

The official rules on https://semver.org state:

Given a version number MAJOR.MINOR.PATCH, increment the:

    MAJOR version when you make incompatible API changes
    MINOR version when you add functionality in a backward compatible manner
    PATCH version when you make backward compatible bug fixes

Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

In fact, there shouldn't be a third dot at all.  Anything after the PATCH version should be separated by either a `+` or `-`.

**Is there a suggested regular expression (RegEx) to check a SemVer string?**

There are two. One with named groups for those systems that support them (PCRE [Perl Compatible Regular Expressions, i.e. Perl, PHP and R], Python and Go).

See: https://regex101.com/r/Ly7O1x/3/

^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$

And one with numbered capture groups instead (so cg1 = major, cg2 = minor, cg3 = patch, cg4 = prerelease and cg5 = buildmetadata) that is compatible with ECMA Script (JavaScript), PCRE (Perl Compatible Regular Expressions, i.e. Perl, PHP and R), Python and Go.

See: https://regex101.com/r/vkijKf/1/

^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$

spytheman:
However I also agree that we should stick with the formal spec, and reject a version like `2.51.1.dirty` in the `semver` parser (btw, it currently can not report parsing errors, which we should probably fix too).

spytheman:
I think that we should modify vpm instead, to strip the `.dirty` suffix before passing it further.

spytheman:
Fixed in 0db62d5 .

JalonSolov:
Perhaps semver parsing needs a `strict` flag?

PR #25531, opened 2025-10-18
parser,checker: vls skip unrelated files

kbkpbot:
This is discussed at #25519

This PR skip all other files except the user file itself.
Hope this can speedup the `vls` process.


JalonSolov:
That would make sense.  The latest C standard did, indeed, introduce BOOL as a real datatype, not just a #define.

medvednikov:
Great job!

PR #25532, opened 2025-10-18
cgen: ensure option none/ok state is also compared

### LINKED ISSUE #25530: match true error
<details>
<summary>V version: V 0.4.12 f6e5118, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 c1d1efb2091e989dbb09e38b308c82c2cade1aa8.f6e5118
|OS                  |windows, Microsoft Windows 11 Pro 22635 64-bit
|Processor           |8 cpus, 64bit, little endian, 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz  
|Memory              |42.22GB/63.67GB
|                    |
|V executable        |D:\software\v\v.exe
|V last modified time|2025-10-06 00:04:56
|                    |
|V home dir          |OK, value: D:\software\v
|VMODULES            |OK, value: C:\Users\Sammy\.vmodules
|VTMP                |OK, value: C:\Users\Sammy\AppData\Local\Temp\v_0
|Current working dir |OK, value: D:\Code\v\unnamed_compiler\v_bugs
|                    |
|Git version         |git version 2.51.0.windows.2
|V git status        |weekly.2025.16-866-gf6e51188
|.git/config present |true
|                    |
|cc version          |cc (Rev2, Built by MSYS2 project) 13.2.0
|gcc version         |gcc (Rev2, Built by MSYS2 project) 13.2.0
|clang version       |clang version 21.1.3
|msvc version        |N/A
|emcc version        |emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 4.0.3 (a9651ff57165f5710bb09a5fe52590fd6ddb72df)
|glibc version       |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  .\match_true_else.v && D:\Code\v\unnamed_compiler\v_bugs\match_true_else.exe`
```v
enum A_ {
	b
	c
	d
}

fn main() {
	mut a := ?A_(none)
	a = .b
	b := ?A_(none)
	match true {
		a == ?A_(.b) && a != b {
			println('b')
		}
		a == ?A_(.c) && a != b {
			println('c')
		}
		else {
			println('else')
		}
	}
}

```

**What did you see?**
```
else
```

"b"

dy-tea:
Fixes #25530.

When comparing options, we can't just compare data. State needs to be compared. When state is 2 for both left and right, data should not be compared, but if they have the same non-2 state then data should be compared.

### DIFF
--- a/vlib/v/gen/c/infix.v
+++ b/vlib/v/gen/c/infix.v
@@ -403,13 +403,25 @@ fn (mut g Gen) infix_expr_eq_op(node ast.InfixExpr) {
 		old_inside_opt_or_res := g.inside_opt_or_res
 		g.inside_opt_or_res = true
 		if node.op == .eq {
-			g.write('!')
+			g.write('(')
+		} else {
+			g.write('!(')
 		}
-		g.write('memcmp(')
+		g.write('(')
 		g.expr(node.left)
-		g.write('.data, ')
+		g.write('.state == 2 && ')
 		g.expr(node.right)
-		g.write('.data, sizeof(${g.base_type(left_type)}))')
+		g.write('.state == 2) || (')
+		g.expr(node.left)
+		g.write('.state == ')
+		g.expr(node.right)
+		g.write('.state && ')
+		g.expr(node.left)
+		g.write('.state != 2 && !memcmp(&')
+		g.expr(node.left)
+		g.write('.data, &')
+		g.expr(node.right)
+		g.write('.data, sizeof(${g.base_type(left_type)}))))')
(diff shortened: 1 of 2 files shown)


dy-tea:
Is windows CI failing because it expects `<stdbool.h>`?

JalonSolov:
That's what it appears to be.  Wonder why it isn't finding it with gcc?

spytheman:
> That's what it appears to be. Wonder why it isn't finding it with gcc?

because only the latest windows runner is updated for now

PR #25534, opened 2025-10-19
ci: fix gcc-windows -std=c99

kbkpbot:
Fix CI issue https://github.com/vlang/v/pull/25531#issuecomment-3419100363


spytheman:
Thank you.

PR #25536, opened 2025-10-19
checker: fix mutability in `for in` loops with pointer values (fix #25520)

### LINKED ISSUE #25520: Unpredictable memory accesses when iterating values "mutably"
<details>
<summary>V version: V 0.4.12, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 7f6db12a461a5388b1b9765acec74138ce2817bc
|OS                  |linux, Linux version 6.16.11-200.fc42.x86_64 (mockbuild@2a5103c3cf1c4acbaad71d7113e31c7b) (gcc (GCC) 15.2.1 20250808 (Red Hat 15.2.1-1), GNU ld version 2.44-6.fc42) #1 SMP PREEMPT_DYNAMIC Mon Oct  6 20:00:39 UTC 2025
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 7800X3D 8-Core Processor
|Memory              |2.05GB/30.94GB
|                    |
|V executable        |/home/yann/Apps/v/v
|V last modified time|2025-10-05 13:36:24
|                    |
|V home dir          |OK, value: /home/yann/Apps/v
|VMODULES            |OK, value: /home/yann/.vmodules
|Current working dir |OK, value: /home/yann/devel/perso/v/bug_report2
|                    |
|V git status        |N/A
|.git/config present |false
|                    |
|cc version          |cc (GCC) 15.2.1 20250808 (Red Hat 15.2.1-1)
|gcc version         |gcc (GCC) 15.2.1 20250808 (Red Hat 15.2.1-1)
|clang version       |clang version 20.1.8 (Fedora 20.1.8-4.fc42)
|emcc version        |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  dereference_mut.v && dereference_mut`
```v
module main

struct MyStruct {
mut:
	v int
}

fn inc(mut s MyStruct) {
	s.v++
}

fn main() {
	mut ex1 := [&MyStruct{v: 1}]

	for mut c in ex1 { // c is **MyStruct whereas I would expect it to be *MyStruct
		inc(mut *c)	// Need to de-reference c
	}
	assert ex1[0].v == 2

	mut ex2 := [&MyStruct{v: 1}]

	for mut c in ex2 {
		inc(mut c) // We can see that we're not incrementing the correct memory location as we're incrementing the reference to c :/
	}
	assert ex2[0].v == 2
}

```

**What did you see?**
```
dereference_mut.v:25: FAIL: fn main.main: assert ex2[0].v == 2
   left value: ex2[0].v = 0
  right value: 2
V panic: Assertion failed...
v hash: 
/tmp/v_1000/dereference_mut.01K7Q089MZ050Z5GARNCPXPJMM.tmp.c:3447: at builtin___v_panic: Backtrace
/tmp/v_1000/dereference_mut.01K7Q089MZ050Z5GARNCPXPJMM.tmp.c:5925: by main__main
/tmp/v_1000/dereference_mut.01K7Q089MZ050Z5GARNCPXPJMM.tmp.c:5953: by main
```

The assert in place to pass

**Summay**

* the mut/non mut behavior is confusing in loops
* although if we're aware of the behavior we can act accordingly, there's no proper check of the argument passed to the called function that may cause unpredictable result: in such a case I'd expect to have a compiling error saying that &&MyStruct can't be passed as &MyStruct.
> [!NOTE]
>

StunxFS:
Currently, the compiler generates an unnecessary pointer when using `mut` in `for in` with pointer values:

```v
$ cat x.v
struct Xyz {}

fn main() {
        mut stack := [&Xyz{}]

        for mut x in stack {
                _ = x
        }
        for x in stack {
                _ = x
        }
}
```

Generates:

```c
for (int _t1 = 0; _t1 < stack.len; ++_t1) {
        main__Xyz** x = ((main__Xyz**)stack.data) + _t1; // <-- Here, it shouldn't be `main__Xyz**` but simply `main__Xyz*`
        {main__Xyz** _ = x;}
        ;
}
for (int _t2 = 0; _t2 < stack.len; ++_t2) {
        main__Xyz* x = ((main__Xyz**)stack.data)[_t2];
        {main__Xyz* _ = x;}
        ;
}
```

With this PR, this behavior is corrected, and now an extra pointer is not generated, but rather the same value that is already a pointer type is used, which avoids the need to make indirections when they are not necessary.

Fix #25520.

### DIFF
--- a/vlib/v/gen/c/for.v
+++ b/vlib/v/gen/c/for.v
@@ -442,11 +442,11 @@ fn (mut g Gen) for_in_stmt(node_ ast.ForInStmt) {
 				g.writeln('memcpy(*(${val_styp}*)${c_name(node.val_var)}, (byte*)builtin__DenseArray_value(&${cond_var}${dot_or_ptr}key_values, ${idx}), sizeof(${val_styp}));')
 			} else {
 				val_styp := g.styp(node.val_type)
-				if node.val_type.is_ptr() {
-					if node.val_is_mut || node.val_is_ref {
-						g.write('${val_styp} ${c_name(node.val_var)} = &(*(${val_styp})')
-					} else {
+				if node.val_is_mut || node.val_is_ref {
+					if g.table.value_type(node.cond_type).is_ptr() {
 						g.write('${val_styp} ${c_name(node.val_var)} = (*(${val_styp}*)')
+					} else {
+						g.write('${val_styp} ${c_name(node.val_var)} = ((${val_styp})')
 					}
 				} else {
 					g.write('${val_styp} ${c_name(node.val_var)} = (*(${val_styp}*)')


medvednikov:
Great job!

PR #25538, opened 2025-10-19
encoding.base58: fix encoding, add test

tankf33der:
I love one-line patches.
Due to this typo, `encoding.base58` couldn't pass testing against an external library.
Now everything is fine, passed `300+ M` loop on random input of random length.
I found copy-paste source library in Internet, [no typo here](https://github.com/mr-tron/base58/blob/master/base58.go#L54).
Now `2x` faster for performance.
I decided not to set `@[direct_array_access]` for `decode()`, tested it has no effect.
Now `master` can not pass test without patch.

These performances requires verification and further investigation.
Need to check if anything can be improved.
base
p.s. `base32` and `base64` are fine, tested too.

### DIFF
--- a/vlib/encoding/base58/base58.v
+++ b/vlib/encoding/base58/base58.v
@@ -50,6 +50,7 @@ pub fn encode_walpha(input string, alphabet Alphabet) string {
 }
 
 // encode_walpha encodes the input array to base58 with a custom aplhabet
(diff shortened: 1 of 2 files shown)


spytheman:
Thank you @tankf33der 🙇🏻.
Excellent work.

PR #25539, opened 2025-10-19
arrays: fix rotate_left() and rotate_right(), add test

tankf33der:
One line patch, typo.
In the current implementation, if a bug occurs, some elements are lost during `rotate_left()` or `rotate_right()`, while the number of elements remains the same.
`master` can not pass the test.


spytheman:
Excellent. Thank you.

PR #25569, opened 2025-10-23
x.crypto.chacha20poly1305: small clean up of internal mac generator and encryption routine for psiv

blackshirt:
This small PR basically improves twos internal routines used in psiv aead operation, in the mean of `psiv_gen_tag` and `psiv_encrypt_internal` cleansup. Its contains changes:
- Removed out the use of `arrays.chunk` machinery in the `psiv_encrypt_internal`. I think It has an overhead by creating a clone on arrays being chunked. It changed to use bytes directly. It also removes not really necessary splitting tag routine, and work with bytes slicing directly.
- Improves `psiv_gen_tag` routine by reusing previously allocated buffer to work on serialization, so it reduces some allocation internally.

Thanks,
Cheers

### DIFF
--- a/vlib/x/crypto/chacha20poly1305/psiv.v
+++ b/vlib/x/crypto/chacha20poly1305/psiv.v
@@ -9,7 +9,6 @@
 // See the detail on the [A Robust Variant of ChaCha20-Poly1305](https://eprint.iacr.org/2025/222).
 module chacha20poly1305
 
-import arrays
 import encoding.binary
 import crypto.internal.subtle
 import x.crypto.chacha20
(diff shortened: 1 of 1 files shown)


tankf33der:
The PR patch is similar to mine, which I already showed and stalled without making a PR.
(I need to change my approach to development and make PRs right away, rather than waiting for a wave of inspiration.)

I have tests with full broad coverage for 99.9% of all `crypto` and `x.crypto` modules.
I've got your back.

All passed.

blackshirt:
> The PR patch is similar to mine, which I already showed and stalled without making a PR. (I need to change my approach to development and make PRs right away, rather than waiting for a wave of inspiration.)
> 
> I have tests with full broad coverage for 99.9% of all `crypto` and `x.crypto` modules. I've got your back.
> 
> All passed.

Yes, actually it based on yours..i'm waiting for your patch to be sent. But, i think it not ready yet.. @spytheman suggested to sent a patch in smaller pr so it can be easily reviewed. So, gradually it Will be updated. Don't worry to share your thought..

blackshirt:
@tankf33der : i hope you can sent it too ...thank for working together ..

spytheman:
Good work @blackshirt.
Thanks @tankf33der.

I do think that multiple smaller/more focused PRs are a lot better compared to 1 bigger one - they are easier to create (smaller uninterrupted time blocks are needed to create them), review, test, merge without conflicts (in case multiple people work on the same area in parallel), and revert if there are regressions.

spytheman (approved):

PR #25572, opened 2025-10-23
parser,checker: fix vls fn signature

kbkpbot:
This PR enable fn signature support for struct method also.
And support fn signature in incomplete code.

work with https://github.com/vlang/vls/pull/469

image


spytheman (approved):
Excellent work.

PR #25574, opened 2025-10-24
x.crypto.chacha20poly1305: move up responsibility for allocs into higher caller

blackshirt:
This patch was related with previous merged PR in [#25569](https://github.com/vlang/v/pull/25569). This PR fundamentally does not changes the core functionality. The reason for this changes was to move up buffer allocation in `psiv_encrypt_internal` and `psiv_gen_tag` into the caller responsible to do that allocs, especially in `.encrypt` and `.decrypt`. It also acts as preliminary step to move into fixed arrays.

This patch contains changes in the means of :
- The main one was a changed in the signature of `psiv_encrypt_internal` and `psiv_gen_tag` to accept output buffer as the first params and adapts in another places that call both of them.
- Updates comments on the code
- Removes out left unused remaining bits, ie `split_tag` helper
- Adjust the tests

Thanks, 
Cheers


tankf33der:
All passed. OK.

blackshirt:
> All passed. OK.

Thank you

PR #25584, opened 2025-10-24
crypto.cipher: fix decryption in CBC mode, add test

tankf33der:
The test against `golang` failed and a trivial audit showed that a [typo](https://cs.opensource.google/go/go/+/refs/tags/go1.25.3:src/crypto/cipher/cbc.go;l=199) occurred again.
Worked for a long time on the test to properly trigger the decryption error.
This new test cannot pass on `master`.
Please review my work and approve.

### DIFF
--- a/vlib/crypto/cipher/aes_cbc_test.v
+++ b/vlib/crypto/cipher/aes_cbc_test.v
@@ -4,6 +4,30 @@
 import crypto.aes
 import crypto.cipher
 
+fn test_aes_cbc_double() {
+	orig1 := []u8{len: 64, init: index}
+	orig2 := []u8{len: 16, init: index}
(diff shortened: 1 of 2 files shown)


vlib/crypto/cipher/aes_cbc_test.v
```diff
@@ -4,6 +4,30 @@
 import crypto.aes
 import crypto.cipher
 
+fn test_aes_cbc_double() {
+	orig1 := []u8{len: 64, init: index}
+	orig2 := []u8{len: 16, init: index}
```
spytheman, line 9: I am not sure if it may have been better to use a more varied content, something like `init: index*10` .

PR #25588, opened 2025-10-25
orm: fix f32 retrieval data for sqlite

### LINKED ISSUE #25583: ORM: fail field with type `f32`
### Describe the bug

Code: https://play.vlang.io/p/b820c03b72

```v
import db.sqlite

struct Test {
	value f32
	valu2 f64
}

conn := sqlite.connect(':memory:')!

data := Test{32.32, 64.64}

sql conn {
	create table Test
	insert data into Test
}!

s := sql conn {
	select from Test
}!

dump(s)
```

### Reproduction Steps

https://play.vlang.io/p/b820c03b72

### Expected Behavior

[code.v:21] s: [Test{
    value: 32.32
    valu2: 64.64
}]

### Current Behavior

Output:

```
[code.v:21] s: [Test{
    value: -2.0
    valu2: 64.64
}]
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.12 b68897c6d7b13254a41f827e3c8b91b48705459c.bfd36f8

```
|V full version      |V 0.4.12 b68897c6d7b13254a41f827e3c8b91b48705459c.bfd36f8
|Memory              |0.82GB/2.02GB
|                    |
|V last modified time|2025-10-24 11:25:40
|                    |
|                    |
|V git status        |N/A
|.git/config present |true
|                    |

|emcc version        |N/A
```
> [!NOTE]
>

felipensp:
Fix #25583


vlib/v/gen/c/orm.v
```diff
@@ -940,6 +940,8 @@ fn (mut g Gen) write_orm_select(node ast.SqlExpr, connection_var_name string, re
 			fields << field
 		}
 	}
+	conn_type := g.get_db_expr_type(node.db_expr) or { ast.void_type }
+	is_sqlite := g.table.type_to_str(conn_type) == 'sqlite.DB'
```
spytheman: hm, why do it only for sqlite? i.e. why special case it?

vlib/v/gen/c/orm.v
```diff
@@ -1239,6 +1241,9 @@ fn (mut g Gen) write_orm_select(node ast.SqlExpr, connection_var_name string, re
 				mut typ := sym.cname
 				g.writeln('${tmp}.${c_name(field.name)} = (${typ}) (*(${array_get_call_code}._i64));')
 				fields_idx++
+			} else if is_sqlite && sym.kind == .f32 {
+				g.writeln('${field_var} = *(${array_get_call_code}._f64);')
+				fields_idx++
```
spytheman: would that work without checking for sqlite explicitly?

vlib/v/gen/c/orm.v
felipensp: because when the data is inserted on sqlite driver it casts to f64. MySQL and Postgres handles f32 and f64 separately.

vlib/v/gen/c/orm.v
felipensp: I haven't tried it on other drivers.

spytheman: imho the `orm` integration in cgen should not care about the details, they should be hidden behind the interface, i.e. there should not be DB specific code here

spytheman: otherwise you will have divergent bugs, that work/not work depending on the connection :-|

felipensp: Yes. I do agree. But looks like we do have a very specific code to handle here.

felipensp: I've changed to an universal fix way.

vlib/v/gen/c/orm.v
spytheman: that should be handled by a conversion function in the orm driver

spytheman: i.e. somewhere in `vlib/db/sqlite/orm.v`, not in the general cgen integration

vlib/v/gen/c/orm.v
felipensp: hmmm, it breaks mysql f32 reading.

spytheman: I do think that it should be fixed in `vlib/db/sqlite/orm.v`, not here.

spytheman: image

this makes it pass on master, without changing cgen

spytheman: i.e. changing just `vlib/db/sqlite/orm.v:167: fn (stmt Stmt) sqlite_select_column(idx int, typ int) !orm.Primitive {` to do a cast for the f32 case

spytheman: The rest of the tests also passed (the impact of the change is minimal):
image

felipensp: hmmm, good catch 👍🏻

Issue #25590, opened 2025-10-25
gg.Context.get_cached_image_by_idx(image_idx) faults if image_idx not found


mike-ward:
### Describe the bug

`get_cached_image_by_idx(image_idx)` faults if image_idx not found. 

### Reproduction Steps

```
v
>>> import gg
>>> mut ctx := gg.new_context(width: 1)
>>> ctx.get_cached_image_by_idx(-1)
V panic: array.get: index out of range (i,a.len):-1, 0
v hash: b0198f2
0   .noprefix.01K8DRKJ4JRNAJ6GDQT0X02B3 0x00000001042e7cd4 builtin__array_get_unsafe + 0
1   .noprefix.01K8DRKJ4JRNAJ6GDQT0X02B3 0x00000001042e7e40 builtin__array_get + 112
2   .noprefix.01K8DRKJ4JRNAJ6GDQT0X02B3 0x00000001042fab9c gg__Context_get_cached_image_by_idx + 52
3   .noprefix.01K8DRKJ4JRNAJ6GDQT0X02B3 0x00000001042fadc8 main__main + 456
4   .noprefix.01K8DRKJ4JRNAJ6GDQT0X02B3 0x00000001042fb28c main + 84
5   dyld                                0x0000000193745d54 start + 7184
>>>
```

### Expected Behavior

Should return `Image{}`

### Current Behavior

Faults or returns the wrong image if the index in out of range or items have been deleted from the cache.

### Possible Solution

- `get_cached_image_by_idx` needs to do a bounds check.
-  `remove_cached_image_by_idx` needs to replace the indexed image with an empty image (Image{})

Callers can test the `Image.ok` member for image validity.

Suggested implementation:

```v
// get_cached_image_by_idx returns a cached `Image` identified by `image_idx`.
//
// See also: cache_image
// See also: remove_cached_image_by_idx
pub fn (mut ctx Context) get_cached_image_by_idx(image_idx int) &Image {
    if image_idx < 0 or image_idx > ctx.image_cache.len - 1 {
        return &Image{}
    }
    return &ctx.image_cache[image_idx]
}

// remove_cached_image_by_idx removes an `Image` identified by `image_idx` from the
// image cache.
//
// See also: cache_image
// See also: get_cached_image_by_idx
pub fn (mut ctx Context) remove_cached_image_by_idx(image_idx int) {
    if image_idx < 0 || image_idx > ctx.image_cache.len - 1 {
        return
    }
[...]

PR #25591, opened 2025-10-25
gg: `Add bounds checking for image caching methods in gg`

mike-ward:
- Implement bounds checks in `get_cached_image_by_idx` and `remove_cached_image_by_idx` to handle invalid indices.
- Add unit tests to verify behavior of caching functions.
- fixes bug #25590

### DIFF
--- a/vlib/gg/image.v
+++ b/vlib/gg/image.v
@@ -46,6 +46,8 @@ pub fn (mut ctx Context) cache_image(img Image) int {
 	return image_idx
 }
 
+const missing_image = Image{}
+
 // get_cached_image_by_idx returns a cached `Image` identified by `image_idx`.
 // If image not found, returns `Image{ok: false}`
 //
@@ -53,7 +55,7 @@ pub fn (mut ctx Context) cache_image(img Image) int {
 // See also: remove_cached_image_by_idx
 pub fn (mut ctx Context) get_cached_image_by_idx(image_idx int) &Image {
 	if image_idx < 0 || image_idx > ctx.image_cache.len - 1 {
-		return &Image{}
+		return &missing_image
 	}
 	return &ctx.image_cache[image_idx]
 }
@@ -67,7 +69,7 @@ pub fn (mut ctx Context) remove_cached_image_by_idx(image_idx int) {
 	if image_idx < 0 || image_idx > ctx.image_cache.len - 1 {
 		return
 	}
-	ctx.image_cache[image_idx] = &Image{}
+	ctx.image_cache[image_idx] = &missing_image
 }
 
 // Draw part of an image using uv coordinates


vlib/gg/image.v
```diff
@@ -47,10 +47,14 @@ pub fn (mut ctx Context) cache_image(img Image) int {
 }
 
 // get_cached_image_by_idx returns a cached `Image` identified by `image_idx`.
+// If image not found, returns `Image{ok: false}`
 //
 // See also: cache_image
 // See also: remove_cached_image_by_idx
 pub fn (mut ctx Context) get_cached_image_by_idx(image_idx int) &Image {
+	if image_idx < 0 || image_idx > ctx.image_cache.len - 1 {
+		return &Image{}
```
spytheman: that will allocate a new image for each cache miss

spytheman: what do you think of creating a `const missing_image = &Image{}`, and returning it here, and also setting the slots to it in `remove_cached_image_by_idx` ?

mike-ward: Don't see a way around it without changing the function signature. A malicious caller could modify the reference,

mike-ward: > what do you think of creating a `const missing_image = &Image{}`, and returning it here, and also setting the slots to it in `remove_cached_image_by_idx` ?

A malicious caller could modify the reference. On the flip side, your suggestion is much more memory-friendly.

mike-ward: If it is allocated in a const segment and the OS prevents it from being modified, then it would be safe.

spytheman: Malicious in what way? The callers will be made by the app author 🤔, that just uses `gg`.

mike-ward: In other words, I'm fine with spy's change and can update the PR if that's what is decided.

mike-ward: > Malicious in what way? The callers will be made by the app author 🤔, that just uses `gg`.

The caller could modify the contents of the returned Image. That would affect future calls.

JalonSolov: It doesn't have to be malicious, either.  Could be accidental, or an honest mistake, or just a bug in the user's code.

If what the const is pointed to is modified, it will affect everything that has a pointer to it.

mike-ward: Don't OS's block changes to memory in constant segments? Or is that something from my distant past OS/2 days? If that's the case, spy's change is safe. I probably would make a slight change to `const missing_image = Image{}`. Otherwise, the compiler might allocate the reference const, but the Image{} on the heap? Maybe I'm overthinking this. :)

mike-ward: A quick test on my Mac proves it can be modified. 
```
v
>>> import gg
>>> const img := gg.Image{}
>>> fn get() &gg.Image { return &img }
>>> mut pt := get()
>>> println(pt.ok)
false
>>> pt.ok = true
>>> println(pt.ok)
true
>>> println(get().ok)
true
>>>
```

JalonSolov: The only thing constant in the suggested change is the pointer to the image struct, not the struct itself.  It means you can change the struct contents, just not the pointer to it.

In this case, creating a new struct on every cache miss is a good thing.  It won't likely happen often, and the struct is fairly small when it is empty.  It also means only a single instance might get modified.

mike-ward:
```
================== C compilation error (from musl-gcc): ==============
cc: In file included from /tmp/v_0/tsession_7fedc9d68d08_01K8E0M4JGYBSPFZ2C3TT7HZY7/image_test.01K8E0QCMGF9T1M6DKA3E6G8Q5.tmp.c:889:
Error: cc: /__w/v/v/thirdparty/sokol/sokol_app.h:2024:18: fatal error: GL/gl.h: No such file or directory
cc:  2024 |         #include <GL/gl.h>
cc:       |                  ^~~~~~~~~
```
Is this something I can fix, or is it a CI thing?

JalonSolov:
Make sure you have the opengl headers installed.  opengl-dev or similar package.  Or in this case, that'll have to be done in that CI step.

vlib/gg/image.v
spytheman: I disagree - the allocation cost is high, and the context matters; in gui apps sometimes drawing images can happen hundreds/thousands of times per frame, and the modification (see the test) will be used to later check a single bit -> very high overhead for a very common operation.

The allocated size of a new gg.Image is also not very small (for something that could potentially be called thousands of times):
```v
import gg
x := sizeof(gg.Image)
dump(x)
```
for me shows 80 bytes on Ubuntu.

The argument for safety in this context also does not make much sense to me - it will be protecting the developer against himself, for usage situations that will not happen in practice. The images are not used in contexts where they will be modified, but drawn (read from). Given the test, they will be also checked if they are ok before usage -> the chance of someone modifying by mistake the const value used for marking the empty slots are slim.

spytheman: Thinking about it more, storing/returning voidptr(0) will be even cheaper, at the cost of causing panics for later operations, if they use the return value without checking.

Edit: nah ... returning a nil pointer would be bad for prototyping games

spytheman:
> Is this something I can fix, or is it a CI thing?

It is a CI thing; `docker-ubuntu-musl` compiles things with `musl-gcc` on Ubuntu, and that environment is very restrictive and suitable for compiling only a limited set of programs (missing headers/broken library paths, most packages on Ubuntu are compiled against glibc, not musl etc).

I just marked the test so that it will be skipped on that specific job.

vlib/gg/image.v
mike-ward: As I said earlier, I'm fine with spy's changes. Can we resolve it, and I'll update the PR?

spytheman:
@mike-ward thank you 🙇🏻.

PR #25594, opened 2025-10-25
time: handle offset correctly in parse_rfc3339(), add tests

tankf33der:
The offset calculation in `parse_rfc3399()` got confused between minutes and hours.
Added tests for different combinations.


tankf33der:
my typo in commit message - should be parse_rfc3339()

spytheman:
do not worry, the one in the merged one will get its title from the PR title, which is now fixed

spytheman (approved):
Excellent find.
Thank you @tankf33der 🙇🏻 .

spytheman:
hm, I was wrong - the merged commit did not get the fixed title from the PR 🤔

Issue #25600, opened 2025-10-27
comptime need a way detect a field is shared or not


kbkpbot:
### Describe the feature

When use generic/comptime , it need some way to detect a struct field is `shared` or not.

### Use Case

a sample code, share.v :

```v
import encoding.binary

struct MyS {
	a int
	cache shared map[u64]string
}

fn main() {
	mut s := MyS{}
	
	b := binary.encode_binary(s)!
	dump(b)
}
```

```
v share.v
```

compile this will just issue a error message :
```
$ v share.v
/media/HD/github/kbkpbot/v/vlib/encoding/binary/serialize.v:69:29: error: `value` is `shared` and must be `rlock`ed or `lock`ed to be passed as non-mut argument
   67 |                 encode_map(mut s, value)!
   68 |             } $else {
   69 |                 encode_primitive(mut s, value)!
      |                                         ~~~~~
   70 |             }
   71 |         }
```

### Proposed Solution

introduce something like 
```v
$if field is shared {
} $else {
}
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.12 ac2e7d7f29ac7c5a914f9a51eb91b3bd28948d89.4b5f337

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 ac2e7d7f29ac7c5a914f9a51eb91b3bd28948d89.4b5f337
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |11.73GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-10-26 13:12:30
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/table
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.42-63-gac2e7d7f-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
[...]

PR #25613, opened 2025-10-28
encoding.binary: fix serialize skip struct shared fields

kbkpbot:
Currently, we can't handle `shared` fields in struct, so skip them.

This is related to issue #25600


spytheman (commented):
why not lock the shared field to get the value?

PR #25615, opened 2025-10-28
gg: Destroy unused GPU resources before replacing cached images in `gg`

mike-ward:
- Ensure proper cleanup of GPU resources (`simg`, `ssmp`) before replacing images in the cache.
- Prevent potential resource leaks by invoking `gfx.destroy_image` and `gfx.destroy_sampler` as needed.

### DIFF
--- a/vlib/gg/image.js.v
+++ b/vlib/gg/image.js.v
@@ -18,5 +18,5 @@ pub fn (ctx &Context) draw_image_with_config(config DrawImageConfig) {
 }
 
 // destroy GPU resources associated with the image
-pub fn (img &Image) destroy() {
+fn (img &Image) destroy() {
 }


mike-ward:
Are these check failures related to my change? Appears to be something related to `.js` files.

spytheman:
> Are these check failures related to my change? Appears to be something related to `.js` files.

The JS struct Image from `vlib/gg/image.js.v` does not have those fields, that were used in the backend independent `image.v` file.

I've pushed a fix, that extracted the operations into its own method, that is implemented in 2 ways - one in image.c.v and one in image.js.v .

spytheman:
Thank you @mike-ward 🙇🏻 .

PR #25617, opened 2025-10-28
rand: fix randomness in shuffle()

tankf33der:
The current implementation incorrectly iterates through the array for shuffling and doesn't touch the `last two` elements, whereas it should only leave the `last one` untouched.

Due to this, some combinations don't occur at all, here's a histogram with 100k attempts.
master

My proposed patch increases randomness and in the future `all` combinations will occur.
patch

Here is text data for histogram above:
```
4086 4113 4116 4134 4231 4305 4153 4178
4268 4163 4137 4129 4264 4179 4189 4197 
4157 4190 4150 4131 4173 4125 4122 4110
```
Now any combination will occur approximately equally.


spytheman:
Excellent find.
Thanks @tankf33der!

PR #25624, opened 2025-10-29
strings: make strings end with zero in repeat() and repeat_string()

tankf33der:
In these two functions the created strings are end with zero and not with NULL.
Demo:
```v
import strings

fn main() {
        s := strings.repeat(48, 3)
        for i in 0 .. 10 {
                unsafe { C.printf(c'%d\n', s.str[i]) }
        }
}
```
p.s. for another module decided to create a separate PR for the sake of cleanliness


spytheman (approved):
Thank you @tankf33der 🙇🏻 .

PR #25625, opened 2025-10-29
builtin:  make string end with null in repeat()

tankf33der:
Now `repeat()` generate 0 as last byte of string not null.
Lets fix this too.

Demo:
```v
fn main() {
        s := u8(57).repeat(3)
        for i in 0 .. 10 {
                unsafe { C.printf(c'%d\n', s.str[i]) }
        }
}
```

### DIFF
--- a/vlib/builtin/int.v
+++ b/vlib/builtin/int.v
@@ -588,7 +588,7 @@ pub fn (b u8) repeat(count int) string {
 	mut bytes := unsafe { malloc_noscan(count + 1) }
 	unsafe {
 		vmemset(bytes, b, count)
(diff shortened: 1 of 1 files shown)


JalonSolov:
Ah... it wasn't NULL at the end, as that would be correct.  Instead it was a `0` (not the NUL value, which is ASCII code 0x00, instead the character `0`, which is ASCII code 0x30).

spytheman:
It was actual 0, before #22100 , which changed it to the ASCII `0` digit.
It is my bad that I did not caught it in the review.
Excellent find @tankf33der. Thank you 🙇🏻‍♂️ .

PR #25631, opened 2025-10-31
x.crypto.chacha20poly1305: small cleans up and improves safety check

blackshirt:
After some while for looking it back, i think there are some deficiency that can be improved on. Two's main changes on this pr was
- Changes `precomp` flag to `done` flag. precomputation phase for deriving keys was already done and stored on the `mac_key`, `enc_key` and `poly1305` instance set on psiv creation. By changing to `done` flag to mark this psiv construct as unusable anymore and effectively preventing to call `.encrypt` or `.decrypt`. so we can just returns error if it happens, instead of panic.
- Removes out `key` field. Its not used on any other places. Its already stored on derived keys, so we can safely remove this out.
- Small cleans up by changing hard coded length  on `vmemset` call 
- Add safety check on `.encrypt` and `.decrypt` call to aware of `done` flag

Thanks,
Cheers.

### DIFF
--- a/vlib/x/crypto/chacha20poly1305/psiv.v
+++ b/vlib/x/crypto/chacha20poly1305/psiv.v
@@ -38,8 +38,6 @@ pub fn new_psiv(key []u8) !&Chacha20Poly1305RE {
 
 	// set the values
 	c := &Chacha20Poly1305RE{
-		key:     key.clone()
-		precomp: true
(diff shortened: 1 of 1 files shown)


tankf33der:
All passed.

spytheman:
Thank you @blackshirt 🙇🏻 .

PR #25639, opened 2025-10-31
all: implement scoped defer (part 1)

### LINKED ISSUE #14701: defer not working correctly from loops
OS: linux, Devuan GNU/Linux 5 (daedalus/ceres)
Processor: 8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
CC version: cc (Debian 11.3.0-3) 11.3.0

**What did you do?**

```
fn f(k int){
	eprintln(k)
}

fn main() {
	mut i := 0
	for i<10 {
		i++
		defer{f(i)}
	}
}
```

output of 10 lines
 
**What did you see instead?**

only one line, the defer is executed once only.

this is the similar code in go:

```
package main
import "log"

func main() {
	i := 0
	for i<10{
		defer log.Printf("close %d", i)
		i += 1
	}
	log.Printf("done")
}
```

which works as expected.

### LINKED ISSUE #22445: RUNTIME ERROR: invalid memory access
### Describe the bug

RUNTIME ERROR: invalid memory access

### Reproduction Steps

```
fn main() {
    m := { "1":"1", "2":"2" }
    if b := m["1"] {
        defer { println(b)}
        println("ok")
    }
}
```
Result:
```
ufm@ufmpc:~/src/vtest$ ./vt
ok

ufm@ufmpc:~/src/vtest$ ./vt
ok

ufm@ufmpc:~/src/vtest$ ./vt
ok

ufm@ufmpc:~/src/vtest$ ./vt
ok

ufm@ufmpc:~/src/vtest$ ./vt
ok
0x774c4afa07cd: at ???: RUNTIME ERROR: invalid memory access
0x01f69500: by ???
Segmentation fault (core dumped)

```

### Expected Behavior

Get the value of "b" or get a compiler error saying that this is not allowed.

### Current Behavior

0x7178a25a07cd: at ???: RUNTIME ERROR: invalid memory access
0x02892500: by ???
Segmentation fault (core dumped)

### Possible Solution

_No response_

_No response_

### V version

0.4.8 e1b842a

Linux

> [!NOTE]
>

### LINKED ISSUE #24663: Undefined behavior if variable is out of scope when defer is called
### Describe the bug

if we define a defer block inside the loop with a variable that is in local scope we get undefined behavior when defer is being called.

### Reproduction Steps

```v
fn main() {
    for i in 1..4 {
	defer {
	    println('Deferred execution for $i')
	}
	println('Hello, World! $i')
    }
}
```

### Expected Behavior

compile error?? (I don't know)

### Current Behavior

output:
```
Hello, World! 1
Hello, World! 2
Hello, World! 3
Deferred execution for 49598448
```

### Possible Solution

_No response_

_No response_

### V version

0.4.10 a91e2def83051b5ad177f89e058f4a09e79433ee

|V full version      |V 0.4.10 a91e2def83051b5ad177f89e058f4a09e79433ee
|Processor           |8 cpus, 64bit, little endian, Apple M1
|Memory              |0.3GB/16GB
|                    |
|V executable        |/Users/user/dev/v/v
|V last modified time|2025-06-05 15:14:14
|                    |
|V home dir          |OK, value: /Users/user/dev/v
|VMODULES            |OK, value: /Users/user/.vmodules
|Current working dir |OK, value: /Users/user
|                    |
|V git status        |weekly.2025.14-315-ga91e2def
|.git/config present |true
|                    |
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

StunxFS:
Based on https://github.com/vlang/v/issues/22445#issuecomment-2405137334.

This PR modifies the behavior of `defer` so that it only executes at the end of the scope where it was declared. Currently, `defer` only executes at the end of each function.

To test this new behavior, you can pass the `-scoped-defer` flag to the compiler.

For example, this piece of code:
```v
fn main() {
	for i in 1 .. 4 {
		defer {
			println('Deferred execution for ${i}')
		}
		println('Hello, World! ${i}')
	}
}
```

If we compile it with the V compiler using the current behavior, we have:
```sh
$ v run x.v
Hello, World! 1
Hello, World! 2
Hello, World! 3
Deferred execution for -561879496
```

But with this PR we have:
```sh
$ v -scoped-defer run x.v
Hello, World! 1
Deferred execution for 1
Hello, World! 2
Deferred execution for 2
Hello, World! 3
Deferred execution for 3
```

The old behavior can still be obtained by using `defer(fn) {`.

Fix #14701.
Fix #22445.
Fix #24663.

### DIFF
--- a/vlib/v/tests/options/option_free_method_test.v
+++ b/vlib/v/tests/options/option_free_method_test.v
@@ -30,7 +30,9 @@ fn test_main() {
 
 	defer {
 		unsafe {
-			t2?.free()
+			if t2 != none {
+				t2.free()
+			}
 		}
 	}
 }


spytheman:
imho it should be the default, without needing a flag

StunxFS:
@spytheman 
> imho it should be the default, without needing a flag

That's exactly the idea. But vlib has code that uses `defer` that needs to be updated to the new behavior.

vlib/v/gen/c/defer.v
```diff
@@ -0,0 +1,63 @@
...
+import v.ast
+
+fn (g &Gen) defer_flag_var(stmt &ast.DeferStmt) string {
+	return '${g.last_fn_c_name}_defer_${stmt.idx_in_fn}'
+}
+
+@[inline]
+fn is_same_scope(a &ast.Scope, b &ast.Scope) bool {
+	return a.start_pos == b.start_pos && a.end_pos == b.end_pos
+}
```
spytheman: this is better to be a method on ast.Scope; it is useful for other parts of the compiler too

vlib/v/pref/pref.v
```diff
@@ -762,6 +764,12 @@ pub fn parse_args_and_show_errors(known_external_commands []string, args []strin
 			'-experimental' {
 				res.experimental = true
 			}
+			'-scoped-defer' {
+				// experimental, remove once defer is scoped by default
+				res.scoped_defer = true
+				res.compile_defines << 'scoped_defer'
+				res.compile_defines_all << 'scoped_defer'
```
spytheman: why? it should not be available at comptime.

vlib/v/tests/defer/defer_test.v
```diff
@@ -187,3 +187,26 @@ fn test_defer_inside_comptime_if_else() {
 	}
 	assert true
 }
+
+fn test_scoped_defer() {
```
spytheman: extract a separate _test.v file; it will make it easier to selectively compile it with the new behavior

vlib/v/pref/pref.v
spytheman: is it needed only during the transition?

StunxFS: I originally tried to make the new behavior the default, but although V compiled successfully, there are certain parts of the code that don't fit with this new behavior.

For example: `pub fn (mut w Walker) fn_decl(mut node ast.FnDecl)` (`vlib/v/markused/walker.v:918-924`)
```v
	if w.level == 0 {
		last_is_builtin_mod := w.is_builtin_mod
		w.is_builtin_mod = node.mod in ['builtin', 'os', 'strconv', 'builtin.closure']
		defer {
			// Here the change will be made immediately upon exiting `if` scope; this needs to be changed.
			w.is_builtin_mod = last_is_builtin_mod
		}
	}
```

My idea was to do something like:
```v
	if w.level == 0 {
		last_is_builtin_mod := w.is_builtin_mod
		w.is_builtin_mod = node.mod in ['builtin', 'os', 'strconv', 'builtin.closure']
		$if !scoped_defer ?
			defer {
				w.is_builtin_mod = last_is_builtin_mod
			}
		}
	}
```

To be able to make a smooth transition-

StunxFS: > is it needed only during the transition?

@spytheman Yes, once the transition is complete, I'll remove that.

vlib/v/ast/scope.v
```diff
@@ -251,6 +251,11 @@ pub fn (s &Scope) contains(pos int) bool {
 	return pos >= s.start_pos && pos <= s.end_pos
 }
 
+@[inline]
+pub fn (s &Scope) is_same(o &Scope) bool {
```
Delta456: I can sure you can use an overloaded `==` here

vlib/v/pref/pref.v
spytheman: 1) we could support also `defer fn { }` for the current behavior, and `defer {}` for the scoped one (behind the flag)
2) once V is bootstrapped (without the new behavior being on), we can gradually mark stuff with `defer fn {` where it is needed.
3) once we do that, and check that tests pass with `-scoped-defer`, we can toggle the default, and code will continue to work (thankfully, there is not much code that depends on using defer in loops now, since it would have been bugged for  obvious reasons).

I do not see the benefit of supporting `$if !scoped_defer ? {` which we will need to then remove at some point 🤔 .

edit: `defer {` -> `defer fn {` could also be done automatically by a modified version of vfmt

StunxFS: @spytheman Okay, I like the idea. In fact, I was thinking that after I finish implementing scoped `defer`, I could implement `defer` modes, that is, `defer` blocks that would only execute under certain circumstances. The modes would be:

- `defer {`: This would be the traditional way, it would be executed at the end of the scope where it was declared.
- `defer(error) {`: This mode would be the same as the previous one, only it would be executed when an error is returned.
- `defer(success) {`: Same as the previous one, but it is only executed when no error was returned.
- `defer(fn) {`: This mode would only be executed at the end of each function (following the old behavior).

What do you think about that?

vlib/v/ast/scope.v
StunxFS: Hmm, maybe

vlib/v/pref/pref.v
spytheman: `defer(fn) {` may be less ambiguous to parse/require less lookahead, compared to `defer fn {` .

I do not have opinion on the rest - can you please show some usage examples for the error/success cases, and their potential interactions with control structures like loops/if conditions?

spytheman:
I am adding this example, since there was confusion on Discord about the order of execution with the scoped defers:
```v
for i in 1 .. 4 {
	defer { println('Deferred execution for ${i}. Defer 1.') }
	defer { println('Deferred execution for ${i}. Defer 2.') }
	defer { println('Deferred execution for ${i}. Defer 3.') }
	println('Hello, World! ${i}')
}
```
produces:
```
#0 22:08:32 ^ feat/scoped-defer ~/v> ./v self && ./v -scoped-defer run xx.v
Hello, World! 1
Deferred execution for 1. Defer 3.
Deferred execution for 1. Defer 2.
Deferred execution for 1. Defer 1.
Hello, World! 2
Deferred execution for 2. Defer 3.
Deferred execution for 2. Defer 2.
Deferred execution for 2. Defer 1.
Hello, World! 3
Deferred execution for 3. Defer 3.
Deferred execution for 3. Defer 2.
Deferred execution for 3. Defer 1.
#0 22:08:34 ^ feat/scoped-defer ~/v>
```
i.e. with `-scoped-defer`, the defers in the same scope, are still executed in LIFO order, but just at the end of each iteration, not just once at the end of the function .

vlib/v/pref/pref.v
StunxFS: `defer(error)` is a `defer` that would only be executed if a function that returns an error fails, for example:

```v
import os

fn main() {
	file := os.open_file('file.txt', 'r')!
	defer(error) { file.close() }
	defer(success) { println('hi, i am defer(success)') }

	mut buf := []u8{}
	// In case an error is returned, `defer(error)` will be executed, and `file` will be closed.
	file.read_bytes_into(0, mut buf)!

	// More operations with `file`...

	// `defer(success)` will be executed if no error is returned by `read_bytes_into`
}
```

`defer(error)`/`defer(success)` will only be executed when returning with `return` or the function reaches its end.

spytheman: should not the file be closed in all cases, no matter if there is an error or not?

StunxFS: Yes, that's true. Here's another example:

```v
fn update_data(db DB) ! {
	// Start the transaction
	tx := db.begin_transaction()!

	// Defer for the error case (Rollback)
	defer(error) {
		tx.rollback()
		// Additional fault handling logic, such as logging.
		// Optionally: log the error 'err' if available in the defer(error) block.
	}

	// Defer for the success case (Commit)
	defer(success) {
		tx.commit()!
	}

	// --- Main logic ---
	db.update_user(tx, 1)!
	db.insert_log(tx, "User updated")!
	// --- End of main logic ---

	// The implicit/explicit return at the end (if there were no errors) triggers defer(success).
	// An error propagation (`!`) in the logic triggers defer(error) before exiting.
}
```

spytheman: The error/success variants are function scoped?

StunxFS: Yes, exactly

spytheman: will they also work, in functions, that return options?

StunxFS: I was thinking of implementing it only for `!T`, but it could also be for options

spytheman: compile time errors for them for the other functions?

StunxFS: Yes, exactly. `defer(error)`/`defer(success)` would only be valid for functions that return `!T`.

Although now that I think about it, `defer(success)` could be valid for `?T` when `none` is not returned.
```v
fn option() ? {
    defer(success) {
        // Do something if `none` is not returned ...
    }

    if some_operation() {
        return 
    }

    if some_operation2() {
        return
    }

    return none
}
```

gechandesu:
Why is there a special syntax for `defer` in functions (old behavior)? Isn't a function just another scope?

I would prefer to use the same syntax for both cases, so that no changes to existing code would be required. I think there is no need to clarify that the `defer` block will be executed on the function exit if it is declared in the function body not in a nested scope.

```v
fn foo() {
    defer {
        println('Deferred execution for fn scope (old behavior)')
    }
 
    for i in 1 .. 4 {
        defer {
            println('Deferred execution for ${i}')
        }
        println('Hello, World! ${i}')
    }

    println('Hello, World!')
}
```

StunxFS:
@gechandesu Since `defer` will change its default behavior from function-scoped to block-scoped, `defer(fn)` is used to make `defer` execute at the end of the function and not at the exit of the block where it was declared.

```v
$ cat x.v
fn main() {
        {
                defer { println('defer inside main().scope.1') }
                println('hi from main().scope.1')
        }
        println('hi from main().scope.0')
}
$ ./v -scoped-defer run x.v
hi from main().scope.1
defer inside main().scope.1
hi from main().scope.0
```

With `defer(fn)`:
```v
$ cat x.v
fn main() {
        {
                defer(fn) { println('defer inside main().scope.1') }
                println('hi from main().scope.1')
        }
        println('hi from main().scope.0')
}
$ ./v -scoped-defer run x.v
hi from main().scope.1
hi from main().scope.0
defer inside main().scope.1
```

StunxFS:
I consider this first part of the scoped-defer implementation complete. V can also parse, generate, and format the `defer(fn) {` syntax.

To do:
- [x] Format the necessary code in vlib to use `defer(fn) {` instead of `defer {` (#25653).
- [ ] Make `defer` scoped by default, remove flag `-scoped-defer` (#25665).

spytheman:
Excellent work @StunxFS .

Thank you very much for working consistently on this important feature 🙇🏻‍♂️. 
It will enable a lot of future improvements and cleanups.

I worked on another implementation of it last year in <https://github.com/spytheman/v/tree/scoped_defer>, but I could not progress it as much as you did here.

vlib/v/tests/defer/scoped_defer_test.v
```diff
@@ -0,0 +1,22 @@
+fn test_scoped_defer() {
+	mut res := 0
+
+	defer(fn) {
+		res++
+		assert res == 5
+	}
+	{
+		res++
```
spytheman, line 9: We should probably have asserts in the future, on the forward direction (for the normal control flow) too for clarity.

spytheman:
@gechandesu you are right, that `defer{}` in the top fn scope, is 1:1 equivalent to `defer(fn){}`, but there are some cases in the existing code, where `defer{}` is used deeply nested in other scopes, introduced by conditions/matches, and those places may have been written, with the current semantic in mind, and may or may not work with the scoped one, and we need a way to mark/re-work them.

(edit: my expectation is that at least for the V repo, in say 90% of the cases, they will work regardless, since they modify things for which the order does not matter, as long as they are executed once the function finish, but I may be wrong ... the biggest difference between the two is for defers in loops, and it was not used like that before)

spytheman:
@medvednikov I think that this PR is an excellent improvement.
What do you think? Do you agree with the new `defer(fn){` syntax, for explicitly opting for the old behaviour?

medvednikov:
Great job @StunxFS 

I agree with the new default behavior.

I'd use `defer fn {`, but this is also fine.

PR #25642, opened 2025-11-01
cgen: minimize the amount of code generated per each assert (generate the common parts for the pass/fail case once)

spytheman:


spytheman:
(rebased over latest master, to get the sumtype ref fix)

PR #25651, opened 2025-11-03
cgen: fix `defer` stmts with `if` exprs and `or` blocks that return a value

StunxFS:
- Fixed the generation of `defer` stmts with `or` blocks that return some value (`x() or { return 123 }`).
- Fixed the generation of `defer` within `if` expressions with a condition that requires a temporary variable (the `defer` was generated after the `goto`).


spytheman:
Please add test cases too.

PR #25653, opened 2025-11-03
vlib: use `defer(fn) {` syntax where appropriate

StunxFS:
Continuation of PR #25639.
~Merge after PR #25651.~

This PR makes use of the syntax `defer(fn) {` where appropriate.

`v -scoped-defer -o v cmd/v` works!

### DIFF
--- a/examples/readline/readline.v
+++ b/examples/readline/readline.v
@@ -14,7 +14,9 @@ fn run() ! {
 		// Explicit comptime block for other OSes than Windows is required to not break compilation on Windows.
 		mut r := readline.Readline{}
 		r.enable_raw_mode_nosig()
-		defer { r.disable_raw_mode() }
+		defer {
+			r.disable_raw_mode()
+		}
 
 		for {
 			entered := r.read_char()!
--- a/vlib/net/websocket/websocket_client.v
+++ b/vlib/net/websocket/websocket_client.v
@@ -182,7 +182,7 @@ pub fn (mut ws Client) listen() ! {
 			}
 			.close {
 				ws.debug_log('read: close')
-				defer(fn) {
+				defer {
 					ws.manage_clean_close()
 				}
 				if msg.payload.len > 0 {
--- a/vlib/os/asset/asset.v
+++ b/vlib/os/asset/asset.v
@@ -12,9 +12,8 @@ pub fn get_path(base_folder string, relative_path string) string {
 		return relative_path.clone()
 	} $else {
 		fpath := os.join_path_single(base_folder, relative_path)
-		res := os.resource_abs_path(fpath)
-		unsafe { fpath.free() }
-		return res
+		defer { unsafe { fpath.free() } }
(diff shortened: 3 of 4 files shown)


vlib/net/mbedtls/ssl_connection.c.v
```diff
@@ -522,7 +522,7 @@ pub fn (s &SSLConn) peer_addr() !net.Addr {
 pub fn (mut s SSLConn) socket_read_into_ptr(buf_ptr &u8, len int) !int {
 	mut res := 0
 	$if trace_ssl ? {
-		defer {
+		defer(fn) {
```
spytheman, line 525: Why does this need to be changed?
Because it is inside a the comptime $if scope, or for some other reason?

examples/readline/readline.v
```diff
@@ -14,9 +14,7 @@ fn run() ! {
 		// Explicit comptime block for other OSes than Windows is required to not break compilation on Windows.
 		mut r := readline.Readline{}
 		r.enable_raw_mode_nosig()
-		defer {
-			r.disable_raw_mode()
-		}
+		defer { r.disable_raw_mode() }
```
spytheman: this seems unrelated 🤔, imho it should not be in the PR

vlib/net/websocket/websocket_client.v
```diff
@@ -182,7 +182,7 @@ pub fn (mut ws Client) listen() ! {
 			}
 			.close {
 				ws.debug_log('read: close')
-				defer {
+				defer(fn) {
```
spytheman: the block ends with `return`, imho it can be left as just `defer {`

vlib/os/asset/asset.v
```diff
@@ -12,8 +12,9 @@ pub fn get_path(base_folder string, relative_path string) string {
 		return relative_path.clone()
 	} $else {
 		fpath := os.join_path_single(base_folder, relative_path)
-		defer { unsafe { fpath.free() } }
-		return os.resource_abs_path(fpath)
+		res := os.resource_abs_path(fpath)
+		unsafe { fpath.free() }
+		return res
```
spytheman: why?

vlib/net/mbedtls/ssl_connection.c.v
StunxFS: Exactly, `defer` uses the comptime `$if` scope, in a future PR I will make `defer` use the parent scope instead of that one.

examples/readline/readline.v
StunxFS: Corrected

vlib/os/asset/asset.v
StunxFS: Corrected

vlib/net/websocket/websocket_client.v
StunxFS: Corrected

PR #25664, opened 2025-11-04
cgen: fix passing sumtype child to generic function taking sumtype pa…

### LINKED ISSUE #25660: C error while passing sumtype to generic function
### Describe the bug

C error while passing sumtype to generic function.

### Reproduction Steps

```v
struct EventA { a u32 }
struct EventB { b u32 }

type Event = EventA | EventB

struct Queue[T] {}

fn Queue.new[T](size u64) Queue[T] {
	return Queue[T]{}
}

fn (mut self Queue[T]) push(val T) {
	_ := val
}

fn main() {
	mut queue := Queue.new[Event](10)
	queue.push(EventA{a: 10})
}
```

### Expected Behavior

Should be able to compile without c error.

### Current Behavior

```
error: passing 'main__EventA'
      (aka 'struct main__EventA') to parameter of incompatible type 'main__Event'
      (aka 'struct main__Event')
 6859 |         main__Queue_T_main__Event_push_T_main__Event(&queue, ((main__EventA){.a = 10,}));
      |                                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~
build/blob.c:1827:103: note: passing argument to parameter 'val'
      here
 1827 |   ...self, main__Event val);
      |                        ^
1 error generated.
```

### Possible Solution

_No response_

_No response_

### V version

V (dy-tea's sumtype-voidptr branch #25652)

V 0.4.12 2ad83399239548c11fb3f61c2b6309f7d101f6f5
> [!NOTE]
>

dy-tea:
Fixes #25660.

It seems like some info was being lost in call_args which meant that sumtype members (children) were not being casted to the parent type in generic fn call args.

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -2411,6 +2411,40 @@ fn (mut g Gen) call_args(node ast.CallExpr) {
 			}
 		}
 	}
+	if node.is_method {
+		left_sym := g.table.sym(node.left_type)
+		if left_sym.info is ast.Struct && left_sym.info.generic_types.len > 0
+			&& left_sym.info.concrete_types.len > 0 {
+			base_type_idx := g.table.find_type_idx(left_sym.name.all_before('['))
+			mut func := ast.Fn{}
+			mut found := false
+			for {
+				if base_type_idx > 0 {
+					base_sym := g.table.sym(ast.idx_to_type(base_type_idx))
+					if base_func := g.table.find_method(base_sym, node.name) {
+						func = base_func
+						found = true
+						break
+					}
+				}
+				if base_func := g.table.find_method(left_sym, node.name) {
+					func = base_func
+					found = true
+				}
+				break
+			}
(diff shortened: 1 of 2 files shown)


vlib/v/gen/c/fn.v
```diff
@@ -2411,6 +2411,40 @@ fn (mut g Gen) call_args(node ast.CallExpr) {
 			}
 		}
 	}
+	if node.is_method {
```
spytheman, line 2414: this code seems like something that the checker or the transformer stage should have done 🤔, i.e. on its own, it does not emit any C, but just updates `expected_types` ?

spytheman: on the other hand, it fits with the code above, that also should have been better in the checker, but is not; I guess we will just need to clean up and move logic at some point from cgen to the previous stages 🤔

dy-tea: Yeah you are probably right.

Issue #25329, opened 2025-09-17
False positive of ``if` expression requires an expression as the last statement of every branch`


xliiv:
### Describe the bug

Nested `if` wrongly expects expression value.

### Reproduction Steps

Run code below

```v
fn some_fn(text string) {
    mut some_map := map[string]string
    dump(some_map)

    key := ''
    found := some_map[key] or {
        key2 := ''

        // THIS WORKS
        // if key2 in some_map {
        //     _ := if some_map[key2] == key {
        //         0
        //     } else {
        //         panic('err')
        //     }
        // } else {
        //     some_map[key2] = key
        //     0
        // }

        // WHILE THIS IS BROKEN
        if key2 in some_map {
            if some_map[key2] == key {
            } else {
                panic('err')
            }
        } else {
            some_map[key2] = key
        }

        key2
    }
    dump(found)
}

fn main() {
    some_fn('abc abc')
}
```

### Expected Behavior

code compiles and runs

### Current Behavior

```
x.v:23:4: error: `if` expression requires an expression as the last statement of every branch
   21 |         // WHILE THIS IS BROKEN
   22 |         if key2 in some_map {
   23 |             if some_map[key2] == key {
      |             ~~~~~~~~~~~~~~~~~~~~~~~~
   24 |             } else {
   25 |                 panic('err')
x.v:23:4: error: the final expression in `if` or `match`, must have a value of a non-void type
   21 |         // WHILE THIS IS BROKEN
   22 |         if key2 in some_map {
   23 |             if some_map[key2] == key {
      |             ~~
   24 |             } else {
   25 |                 panic('err')
x.v:28:19: error: `if` expression requires an expression as the last statement of every branch
   26 |             }
   27 |         } else {
   28 |             some_map[key2] = key
      |                            ^
   29 |         }
   30 |
```

### V version

V 0.4.11

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.11 487feb9b0e330a6e3d9720adb0322f55b9daf361
|V last modified time|2025-09-01 17:45:42

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #25667, opened 2025-11-05
checker: fix or block endwiths expr

kbkpbot:
fix issue #25329


spytheman (approved):
Excellent work.
Thank you @kbkpbot .

Issue #23404, opened 2025-01-08
Warn about dead code within `match` and `if`/`else`


jorgeluismireles:
### Describe the feature

A programmer mistake can produce dead code if uses within `if`/`else` branches comparisons like `a == a` and within `match` like `match a { a { ... } else { ... } }`. 

The compiler could look for these types of comparisons and issue *warnings* about the possibility of dead code.

### Use Case

Dead code with `if` and `else`:
```v
fn f1(a int) string {
	if a == a {
		return 'a' // always
	} else if a == a {
		return 'a again' // never happens
	} else {
		return 'other' // never happens
	}
}
```
Dead code within `match`:
```v
fn f2(a int) string {
	return match a {
		a { 'a' } // always
		1 { '1' } // never
		else { 'else' } // never
	}
}
```
`match` is more robust than `if`/`else` since already prevents repeated constants and exhaust the search with mandatory `else`. But with variables or post-calculated constants also can produce dead code. Some examples are in issue https://github.com/vlang/v/issues/23403

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.9 9fc8352 

### Environment details (OS name and version, etc.)

Playground

PR #25674, opened 2025-11-05
checker,transformer: add always true false branch detect

kbkpbot:
feature request by #23404
This will warn on always true cond, such as `a == a`, and always false cond, such as ` 1==2`.

Fix `transformer` support `ast.CharLiteral` expr, support general var such as `struct.f != struct.f`

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -152,8 +152,8 @@ mut:
 	v_current_commit_hash string // same as old C.V_CURRENT_COMMIT_HASH
 	assign_stmt_attr      string // for `x := [1,2,3] @[freed]`
 
-	js_string           ast.Type = ast.void_type // when `js"string literal"` is used, `js_string` will be equal to `JS.String`
-	checker_transformer &transformer.Transformer
+	js_string           ast.Type                 = ast.void_type // when `js"string literal"` is used, `js_string` will be equal to `JS.String`
+	checker_transformer &transformer.Transformer = unsafe { nil }
 }
 
 pub fn new_checker(table &ast.Table, pref_ &pref.Preferences) &Checker {
--- a/vlib/v/checker/tests/always_true_false_branch.out
+++ b/vlib/v/checker/tests/always_true_false_branch.out
@@ -194,7 +194,7 @@ vlib/v/checker/tests/always_true_false_branch.vv:61:1: notice: condition is alwa
       | ^
    62 |         1
    63 |     } else {
-vlib/v/checker/tests/always_true_false_branch.vv:69:24: notice: condition is always true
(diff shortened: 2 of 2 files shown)


vlib/v/checker/if.v
```diff
@@ -624,3 +633,124 @@ fn (mut c Checker) check_non_expr_branch_last_stmt(stmts []ast.Stmt) {
 		c.error('expression evaluated but not used', last_stmt.pos)
 	}
 }
+
+// check for always true branch cond, e.g., `a == a`, `true`
+fn (mut c Checker) is_always_true_cond(mut cond ast.Expr) (bool, bool) {
+	mut result := false
+	mut resolved := false
+	match mut cond {
```
StunxFS: Wouldn't it be a good idea to use `transformer` to resolve the expressions? 🤔

kbkpbot: Your suggestion reminded me that we could make `is_always_true_cond()` a general AST function. This way, it can be called within the `checker` to issue warnings and within the `transformer` to optimize the code.

kbkpbot: or move the code to `transformer`.

Delta456: Let's move this code to `transformer`

vlib/v/checker/if.v
```diff
@@ -156,6 +158,16 @@ fn (mut c Checker) if_expr(mut node ast.IfExpr) ast.Type {
...
 						branch.cond.pos())
 				}
+
+				mut check_expr := branch.cond
+				t_expr := t.expr(mut check_expr)
+				if t_expr is ast.BoolLiteral {
+					if t_expr.val {
+						c.warn('condition is always true', branch.cond.pos())
+					} else {
+						c.warn('condition is always false', branch.cond.pos())
```
spytheman: Use notices. Blocks like `if true { return }` otherwise produce `unreachable code` errors with -prod.

vlib/v/checker/match.v
```diff
@@ -235,6 +237,23 @@ fn (mut c Checker) match_expr(mut node ast.MatchExpr) ast.Type {
...
+				op:    .eq
+				left:  node.cond
+				right: expr
+			}
+			t_expr := t.expr(mut check_expr)
+			if t_expr is ast.BoolLiteral {
+				if t_expr.val {
+					c.warn('match is always true', expr.pos())
+				} else {
+					c.warn('match is always false', expr.pos())
```
spytheman: same - use a notice here

spytheman:
What is the performance impact of the additional checks?
Can you please run `v run .github/workflows/compare_pr_to_master.v` ?

spytheman:
> Is this acceptable ?

Yes, 1-2% hit is a lot more palatable.
Thank you.

Issue #25641, opened 2025-11-01
[json2] Wrong int decode.


K1ndWha1e:
### Describe the bug

Hey! After encoding and decoding from struct to json and back doubles int value.
https://play.vlang.io/p/8197dec6ca

### Reproduction Steps

```v
import x.json2

struct Person {
	name string = "Frodo"
	surname string = "Baggins"
	age u16 = 25
}

fn main() {
	p := Person{}
	
    recv_data := json2.encode(p)
	decoded_data := json2.decode[Person](recv_data.str()) or { panic(err) }
	println(decoded_data)
}
```

### Expected Behavior

```text
Person{
    name: 'Frodo'
    surname: 'Baggins'
    age: 25
}
```

### Current Behavior

```text
Person{
    name: 'Frodo'
    surname: 'Baggins'
    age: 2525
}
```

### V version

V 0.4.12 7801f5b

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 c1d1efb2091e989dbb09e38b308c82c2cade1aa8.7801f5b
|:-------------------|:-------------------
|OS                  |windows, Њ ©Єа®б®дв Windows 10 Pro 19045 64-а §ап¤­ п
|Processor           |12 cpus, 64bit, little endian, Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz  
|Memory              |47.18GB/63.92GB
|                    |
|V executable        |C:\Users\nikit\v\v.exe
|V last modified time|2025-10-29 16:51:56
|                    |
|Git version         |git version 2.47.1.windows.2
|V git status        |weekly.2025.40-134-g7801f5b1
|.git/config present |true
|                    |
|cc version          |cc (Rev6, Built by MSYS2 project) 13.1.0
|gcc version         |gcc (Rev6, Built by MSYS2 project) 13.1.0
|tcc version         |tcc version 0.9.27 (x86_64 Windows)

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

felipensp:
<img width="545" height="189" alt="Image" src="https://github.com/user-attachments/assets/bb3d7867-4d20-44ad-b6fb-0f3201982ebb" />

makes it wrong.

chethanyadav456:
@felipensp @K1ndWha1e fixed it in #25675 

[...]

PR #25675, opened 2025-11-05
x.json2: fix decode_number not initializing value to zero

chethanyadav456:
Fixes issue #25641 where u16 and other integer types were incorrectly decoded from JSON. The bug occurred because decode_number accumulated digits without first initializing the target value to zero.

When decoding struct fields, if the struct had a pre-existing value, the decoder would append digits to that value instead of replacing it. For example, decoding 25 into a field that already contained 25 would result in 2525.

The fix adds explicit zero initialization at the start of both positive and negative number decoding paths before accumulating digits.

Changes:
- Initialize *val = 0 before processing positive integers
- Initialize *val = 0 before processing negative integers
- Add test case to verify the fix and prevent regression

Screenshot 2025-11-05 at 9 09 55 PM

### DIFF
--- a/vlib/x/json2/decode.v
+++ b/vlib/x/json2/decode.v
@@ -946,65 +946,78 @@ fn (mut decoder Decoder) decode_number[T](val &T) ! {
 		}
 
 		// doing it like this means the minimum of signed numbers does not overflow before being inverted
+		*val = 0 // initialize to zero before accumulating digits
 		if !is_negative {
-			*val = 0 // Initialize to zero before accumulating digits
 			digit_amount := get_number_digits(*val)
 
 			if number_info.length > digit_amount {
 				decoder.decode_error('overflows ${typeof(val).name}')!
 			}
 
-			for index < number_info.length {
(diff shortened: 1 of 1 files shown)


chethanyadav456:
Fixed the formatting issue. The test file has been reformatted with `v fmt` and force-pushed. The CI should pass now.

vlib/x/json2/decode.v
```diff
@@ -947,76 +947,64 @@ fn (mut decoder Decoder) decode_number[T](val &T) ! {
 
 		// doing it like this means the minimum of signed numbers does not overflow before being inverted
 		if !is_negative {
+			*val = 0 // Initialize to zero before accumulating digits
```
spytheman:
```suggestion
*val = 0 // Initialize to zero before accumulating digits
if !is_negative {
```

spytheman: This is the only relevant line to the issue, the rest is not related and should not be in the PR.

spytheman:
Thank you @chethanyadav456 🙇🏻 .

PR #25677, opened 2025-11-06
cgen: fix `defer` generation within comptime if/match/for

StunxFS:
I had forgotten to generate the `defer` for `$if`, `$match` and `$for` 😞 .


Delta456:
FYI: We also have `$match` now 🙂

spytheman (approved):
Excellent work.
Thanks @StunxFS.

PR #25680, opened 2025-11-07
builder: fix json-errors windows path separator

kbkpbot:
This PR fix `json-errors` message which will pass to `vls`. 

This bug will cause json decode fail when the `path` is a windows path separator.

```
js=[
{
"path":"C:\Users\DDT\AppData\Local\Temp\atof.v",
"message":"unknown field `no_extra_char` in struct literal of type `strconv.AtoF64Param`.\nDid you mean `allow_extra_chars`?",
"line_nr":11,
"col":27,
"len":0
}
]

failed to parse json failed to decode JSON string: {
"path":"C:\Users\DDT\AppDat
run_v_check errors:[]
```

### DIFF
--- a/vlib/v/tests/vls/autocomplete_module_test.v
+++ b/vlib/v/tests/vls/autocomplete_module_test.v
@@ -307,7 +307,7 @@ fn test_main() {
 			println('${term.green('OK  ')} ${t.cmd}')
 		}
 
-		// Try to deocde the response message and verify
+		// Try to decode the response message and verify
 		// TODO: remove `unresolved type, maybe`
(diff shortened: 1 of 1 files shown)


vlib/v/tests/vls/autocomplete_module_test.v
```diff
@@ -236,6 +306,90 @@ fn test_main() {
 		} else {
 			println('${term.green('OK  ')} ${t.cmd}')
 		}
+
+		// Try to deocde the response message and verify
```
spytheman:
```suggestion
		// Try to decode the response message and verify
```

spytheman (approved):
Thanks @kbkpbot .

PR #25681, opened 2025-11-07
checker: error/warn when using `defer(fn)` inside function-scope and `lock` stmts

StunxFS:
This PR now also allows the use of `defer {` inside `lock` stmts.


vlib/v/checker/tests/defer_fn_inside_fn_scope_err.out
```diff
@@ -0,0 +1,6 @@
+vlib/v/checker/tests/defer_fn_inside_fn_scope_err.vv:3:2: error: `defer` is already in function scope; just use `defer {` instead
```
spytheman: this should be a notice (or at most a warning), but not an error

spytheman: refactoring/prototyping becomes needlessly harder with hard errors like this one, where the semantics is ultimately the same

PR #25682, opened 2025-11-07
io.string_reader: fix fill_buffer(), add tests

tankf33der:
This module is proving difficult, been quite a pain.
I wanted to make one big patch, but don't have the brainpower to grasp the ungraspable all at once. Will have to split it into chunks. Better to focus on improved testing.
Can't really break much anyway, because nothing was working correctly anyway. This is the first patch in the series.

Currently the `fill_buffer()` function doesn't work correctly with any parameter, sometimes it fills too little, sometimes too much. Tests cannot pass on the `master` branch.

In the future, I will gradually make the necessary patches for all functions and there will be tests with full coverage for the entire module.

### DIFF
--- a/vlib/io/string_reader/string_reader_test.v
+++ b/vlib/io/string_reader/string_reader_test.v
@@ -1,7 +1,6 @@
 module string_reader
 
 import io
-import math
 
 struct Buf {
 pub:


vlib/io/string_reader/string_reader_test.v
```diff
@@ -18,6 +25,18 @@ fn (mut b Buf) read(mut buf []u8) !int {
 	return n
 }
 
+fn (mut r TwoByteReader) read(mut buf []u8) !int {
+	if r.pos >= r.data.len {
+		return io.Eof{}
+	}
+	min := math.min(r.data.len - r.pos, 2)
```
spytheman:
```suggestion
	min := int_min(r.data.len - r.pos, 2)
```

spytheman: there is no need to `import math` just for the min/max functions (it adds a dependency that makes compiling the test a bit slower)

vlib/io/string_reader/string_reader_test.v
```diff
@@ -1,6 +1,7 @@
 module string_reader
 
 import io
+import math
```
spytheman:
```suggestion
```

PR #25685, opened 2025-11-07
io.string_reader: fix fill_buffer_until(), add tests

tankf33der:
Saga continues.
The next function to fix is `fill_buffer_until()`.
Various tests were made on different combinations to verify the function's operation.

### DIFF
--- a/vlib/io/string_reader/string_reader.v
+++ b/vlib/io/string_reader/string_reader.v
@@ -114,8 +114,12 @@ pub fn (mut r StringReader) fill_buffer_until(n int) !int {
 	}
 
 	mut end := start
+	defer {
+		// shrink the length of the buffer to the total of bytes read
+		r.builder.go_back(r.builder.len - end)
+	}
 	for {
-		read := reader.read(mut r.builder[start..]) or {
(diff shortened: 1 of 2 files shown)


vlib/io/string_reader/string_reader.v
```diff
@@ -124,10 +128,11 @@ pub fn (mut r StringReader) fill_buffer_until(n int) !int {
 		if read == 0 || end - start == n {
 			break
 		} else if r.builder.len == end {
-			if n - end > io.read_all_grow_len {
+			remaining := n - (end - start)
+			if remaining > io.read_all_grow_len {
 				unsafe { r.builder.grow_len(io.read_all_grow_len) }
 			} else {
-				unsafe { r.builder.grow_len(n - end) }
+				unsafe { r.builder.grow_len(remaining) }
```
spytheman, line 135: using int_min can simplify this too

tankf33der: Thanks for doing the merge without this. I wasn't ignoring you, just struggling with my reluctance to add optimization to this PR. The progress is going really tough.

PR #25693, opened 2025-11-08
parser, checker: only collect inner vars when `defer` is function-scoped

StunxFS:
For this piece of code:

```v
fn main() {
	mut x := 0
	{
		a := 1
		defer(fn) {
			x = a
		}
	}
}
```

The compiler will extract the declaration of `a` and place it in the same scope as `x`, so that it is accessible from the function's scope.

Currently, the compiler does this regardless of whether the `defer` is block-scoped or function-scoped.
This PR ensures that this only happens when the `defer` is function-scoped.

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -2435,24 +2435,24 @@ fn (mut c Checker) defer_stmt(mut node ast.DeferStmt) {
 		if c.locked_names.len != 0 || c.rlocked_names.len != 0 {
 			c.error('`defer(fn)`s are not allowed in lock statements', node.pos)
 		}
-	}
-	for i, ident in node.defer_vars {
-		mut id := ident
-		if mut id.info is ast.IdentVar {
-			if id.comptime
-				&& (id.tok_kind == .question || id.name in ast.valid_comptime_not_user_defined) {
-				node.defer_vars[i] = ast.Ident{
-					scope: unsafe { nil }
-					name:  ''
+		for i, ident in node.defer_vars {
+			mut id := ident
+			if mut id.info is ast.IdentVar {
+				if id.comptime
+					&& (id.tok_kind == .question || id.name in ast.valid_comptime_not_user_defined) {
+					node.defer_vars[i] = ast.Ident{
+						scope: unsafe { nil }
+						name:  ''
+					}
+					continue
 				}
-				continue
+				typ := c.ident(mut id)
+				if typ == ast.error_type_idx {
+					continue
+				}
+				id.info.typ = typ
+				node.defer_vars[i] = id
 			}
-			typ := c.ident(mut id)
-			if typ == ast.error_type_idx {
-				continue
-			}
-			id.info.typ = typ
-			node.defer_vars[i] = id
 		}
 	}
 	c.stmts(mut node.stmts)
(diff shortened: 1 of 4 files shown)


spytheman:
I thought, that we would limit/add more checks for the `defer(fn) {}` construct, so that it could use only variables, that are in the fn scope directly 🤔 (i.e. reduce/eliminate the implicit hoisting).

The implicit hoisting can have some surprising effects, for example:
```v
fn main() {
	mut x := 0
	{
		for a := 1; a < 3; a++ {
			defer {
				println('scope deferred x: ${x}, a: ${a}')
			}
			defer(fn) {
				x = a
				println('fn deferred x: ${x}, a: ${a}')
			}
		}
	}
}
```
prints:
```
scope deferred x: 0, a: 1
scope deferred x: 0, a: 2
fn deferred x: 3, a: 3
```
i.e. even though `a` was never 3 in the loop body, it did became 3, at the end of the last iteration, before the loop check exited the loop .

Are there cases, where that is still desirable?

StunxFS:
@spytheman Yes, in fact, there is code in cgen that uses this implicit hoisting (but not within loops).

My thought was to only ban the use of variables declared within `for` loops, thus avoiding problems like the one described in your example.

I have a branch in my fork where I tried to prevent the use of variables that are not in the function's scope; this is the output I got:

```v
vlib/v/markused/walker.v:926:23: error: cannot use block-scoped variable `last_is_builtin_mod` inside `defer(fn) {`
  924 |         w.is_builtin_mod = node.mod in ['builtin', 'os', 'strconv', 'builtin.closure']
  925 |         defer(fn) {
  926 |             w.is_builtin_mod = last_is_builtin_mod
      |                                ~~~~~~~~~~~~~~~~~~~
  927 |         }
  928 |     }
vlib/v/gen/c/cgen.v:2933:27: error: cannot use block-scoped variable `old_inside_smartcast` inside `defer(fn) {`
 2931 |                 g.inside_smartcast = true
 2932 |                 defer(fn) {
 2933 |                     g.inside_smartcast = old_inside_smartcast
      |                                          ~~~~~~~~~~~~~~~~~~~~
 2934 |                 }
 2935 |             } else {
vlib/v/gen/c/fn.v:2317:46: error: cannot use block-scoped variable `t` inside `defer(fn) {`
 2315 |                 t := '_arg_expr_${arg.expr.name.replace('.', '_')}_${arg.expr.pos.pos}'
 2316 |                 defer(fn) {
 2317 |                     g.writeln(';\n\tbuiltin__string_free(&${t});')
      |                                                             ^
 2318 |                 }
 2319 |             }
vlib/v/gen/c/infix.v:1294:31: error: cannot use block-scoped variable `inside_interface_deref_old` inside `defer(fn) {`
 1292 |         g.inside_interface_deref = true
 1293 |         defer(fn) {
 1294 |             g.inside_interface_deref = inside_interface_deref_old
      |                                        ~~~~~~~~~~~~~~~~~~~~~~~~~~
 1295 |         }
 1296 |     }
```

spytheman:
@StunxFS thank you for checking.

Very well then - those do seem like good use cases to me.

spytheman (approved):
Excellent work.
I am very glad, that you are continuing to refine it @StunxFS 🙇🏻 .

PR #25695, opened 2025-11-08
checker: in vls mode, make ensure_type_exists always return true

kbkpbot:
In `vls` mode, make `ensure_type_exists()` always return true, this will enable continue check the remaining stmts in `vls` mode, and let autocomplete or goto  definition work.

### DIFF
--- a/cmd/tools/modules/testing/common.v
+++ b/cmd/tools/modules/testing/common.v
@@ -490,7 +490,8 @@ fn worker_trunner(mut p pool.PoolProcessor, idx int, thread_id int) voidptr {
 		skip_running = ''
 	}
 	reproduce_cmd := '${os.quoted_path(ts.vexe)} ${reproduce_options.join(' ')} ${os.quoted_path(file)}'
-	cmd := '${os.quoted_path(ts.vexe)} ${skip_running} ${cmd_options.join(' ')} ${os.quoted_path(file)}'
+	compile_options := cmd_options.filter(it != '-silent')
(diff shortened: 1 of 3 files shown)


vlib/v/checker/checker.v
```diff
@@ -5576,7 +5576,7 @@ fn (mut c Checker) ensure_generic_type_specify_type_names(typ ast.Type, pos toke
 fn (mut c Checker) ensure_type_exists(typ ast.Type, pos token.Pos) bool {
 	if typ == 0 {
 		c.error('unknown type', pos)
-		return false
+		return if c.pref.is_vls { true } else { false }
```
spytheman:
```suggestion
		return c.pref.is_vls
```

vlib/v/checker/checker.v
```diff
@@ -5576,7 +5576,7 @@ fn (mut c Checker) ensure_generic_type_specify_type_names(typ ast.Type, pos toke
 fn (mut c Checker) ensure_type_exists(typ ast.Type, pos token.Pos) bool {
 	if typ == 0 {
```
spytheman, line 5577: Would it not work, if you put a single `if c.pref.is_vls { return true }` here at the start of the method?

kbkpbot: No, if put a `if c.pref.is_vls { return true }` at the beginning of `ensure_type_exists()`, we will not able to get an error message.
I want to get the error message, and also let `v` continue the check.

spytheman (approved):
Excellent work 🙇🏻 .

PR #25699, opened 2025-11-09
ast,checker: better type checking for sumtypes with generics

### LINKED ISSUE #25690: Can't infer subtype for generic sumtype in constructor
### Describe the bug

When using a generic sumtype, the compiler fails to infer the subtype for one of its variants during construction.

The non-generic version compiles fine, but the generic version requires an explicit cast like `Chain[f64](Empty{})` to work.

### Reproduction Steps

The regular version works:

```v
struct Empty {}

struct Node {
    value f64
    next Chain
}

type Chain = Empty | Node

fn get(chain Chain) f64 {
    return match chain {
        Empty { 0 }
        Node { chain.value }
    }
}

fn main() {
    chain := Node{0.2, Empty{}}
    println(get(chain)) // output: 0.2
}
```

While the generic version doesn't:

```v
struct Empty {}

struct Node[T] {
    value T
    next Chain[T]
}

type Chain[T] = Empty | Node[T]

fn get[T](chain Chain[T]) T {
    return match chain {
        Empty { 0 }
        Node[T] { chain.value }
    }
}

fn main() {
    chain := Node{0.2, Empty{}}
    println(get(chain))
}
```

Compile error:

```
main.v:41:24: error: cannot assign to field `next`: expected `Chain[f64]`, not `Empty`
   39 |
   40 | fn main() {
   41 |     chain := Node{0.2, Empty{}}
      |                        ~~~~~~~
   42 |     println(get(chain))
   43 | }
main.v:42:17: error: cannot use `Node[f64]` as `Chain[f64]` in argument 1 to `get`
   40 | fn main() {
   41 |     chain := Node{0.2, Empty{}}
   42 |     println(get(chain))
      |                 ~~~~~
   43 | }
```

However, `chain := Node{0.2, Chain[f64](Empty{})}` does compile.

### Expected Behavior

The generic version can compile.

### Current Behavior

The compiler raises a type mismatch error, expecting `Chain[f64]` instead of `Empty`, even though `Empty` is a valid variant of `Chain[T]`.

### Possible Solution

_No response_

_No response_

### V version

V 0.4.12 f6336fc

V 0.4.12 f6336fc0ecbd4ce3e91c24a67c01f207d16674d5
> [!NOTE]
>

dy-tea:
Fixes #25690.

Had to modify ast a bit as there were some issues with generated var names.
Indentation level in `check_types` in a bit high, maybe better to extract it to a method?

### DIFF
--- a/vlib/v/checker/struct.v
+++ b/vlib/v/checker/struct.v
@@ -969,33 +969,9 @@ or use an explicit `unsafe{ a[..] }`, if you do not want a copy of the slice.',
 		&& c.table.cur_concrete_types.len == 0 {
 		if struct_sym.info.concrete_types.len == 0 {
 			concrete_types := c.infer_struct_generic_types(node.typ, node)
-			if concrete_types.len > 0 && concrete_types.len == struct_sym.info.generic_types.len {
-				mut bs_name := struct_sym.name + '['
-				mut bs_cname := struct_sym.cname + '_T_'
-				for i, ct in concrete_types {
-					ct_sym := c.table.sym(ct)
-					bs_name += ct_sym.name
-					bs_cname += ct_sym.cname
-					if i < concrete_types.len - 1 {
-						bs_name += ', '
-						bs_cname += '_'
-					}
-				}
-				bs_name += ']'
-				if existing_idx := c.table.type_idxs[bs_name] {
-					node.typ = ast.new_type(existing_idx)
-				} else {
-					idx := c.table.register_sym(ast.TypeSymbol{
-						kind:   .generic_inst
-						name:   bs_name
-						cname:  bs_cname
-						ngname: struct_sym.ngname
-						mod:    struct_sym.mod
-						info:   ast.GenericInst{
-							parent_idx:     node.typ.idx()
-							concrete_types: concrete_types
-						}
-					})
+			if concrete_types.len > 0 {
+				idx := c.table.find_or_register_generic_inst(node.typ, concrete_types)
+				if idx > 0 {
 					node.typ = ast.new_type(idx)
 					c.table.generic_insts_to_concrete()
 				}
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -935,9 +935,11 @@ pub fn (mut t Table) register_sym(sym TypeSymbol) int {
 	idx = t.type_symbols.len
 	t.type_symbols << &TypeSymbol{
 		...sym
-		ngname: strip_generic_params(sym.name)
 	}
 	t.type_symbols[idx].idx = idx
+	if t.type_symbols[idx].ngname == '' {
+		t.type_symbols[idx].ngname = strip_generic_params(sym.name)
+	}
 	t.type_idxs[sym_name] = idx
 	return idx
 }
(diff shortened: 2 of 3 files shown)


spytheman:
> Indentation level in check_types in a bit high, maybe better to extract it to a method?

Sure, as long as it will not need a lot of state to be passed as parameters.

vlib/v/checker/check_types.v
```diff
@@ -460,6 +460,24 @@ fn (mut c Checker) check_basic(got ast.Type, expected ast.Type) bool {
...
+	if exp_sym.kind == .placeholder && c.expected_type != ast.void_type {
+		base_type_name := exp_sym.name.all_before('[')
+		base_type := c.table.find_type(base_type_name)
+		if base_type != 0 {
+			base_sym := c.table.sym(base_type)
+			if base_sym.kind == .sum_type && base_sym.info is ast.SumType {
+				base_info := base_sym.info as ast.SumType
+				for variant in base_info.variants {
+					variant_sym := c.table.sym(variant)
+					got_base_name := got_sym.name.all_before('[')
```
spytheman: extract that before the loop

vlib/v/checker/fn.v
```diff
@@ -1569,6 +1569,35 @@ fn (mut c Checker) fn_call(mut node ast.CallExpr, mut continue_check &bool) ast.
...
+			base_type_name := param_type_sym.name.all_before('[')
+			base_type := c.table.find_type(base_type_name)
+			if base_type != 0 {
+				base_sym := c.table.sym(base_type)
+				if base_sym.kind == .sum_type && base_sym.info is ast.SumType {
+					base_info := base_sym.info as ast.SumType
+					arg_typ_sym := c.table.sym(arg_typ)
+					for variant in base_info.variants {
+						variant_sym := c.table.sym(variant)
+						arg_base_name := arg_typ_sym.name.all_before('[')
```
spytheman: extract this before the loop (it will not change in it)

vlib/v/checker/struct.v
```diff
@@ -783,9 +791,35 @@ or use an explicit `unsafe{ a[..] }`, if you do not want a copy of the slice.',
...
-							init_field.pos)
+					mut needs_sum_type_cast := false
+					if exp_type_sym.kind == .placeholder {
+						base_type := c.table.find_type(exp_type_sym.name.all_before('['))
+						if base_type != 0 {
+							base_sym := c.table.sym(base_type)
+							if base_sym.kind == .sum_type && base_sym.info is ast.SumType {
+								base_info := base_sym.info as ast.SumType
+								for variant in base_info.variants {
+									if c.table.sym(variant).name.all_before('[') == got_type_sym.name.all_before('[') {
```
spytheman: the second name will not change inside the loop

vlib/v/checker/struct.v
```diff
@@ -935,9 +969,35 @@ or use an explicit `unsafe{ a[..] }`, if you do not want a copy of the slice.',
...
+						cname: bs_cname
+						mod:   struct_sym.mod
+						info:  ast.GenericInst{
+							parent_idx:     node.typ.idx()
+							concrete_types: concrete_types
+						}
+					})
+					node.typ = ast.new_type(idx)
+					c.table.generic_insts_to_concrete()
+				}
```
spytheman: This seems to me, like code that should be in a Table method in `v.ast`, not in the checker (it knows about `_T_` and cares about `, ` and `_`, it registers symbols etc).

spytheman:
What do you think of caching `sym.name.all_before('[')` in a `non_generic_name` field of `TypeSymbol` @dy-tea?

dy-tea:
> What do you think of caching `sym.name.all_before('[')` in a `non_generic_name` field of `TypeSymbol` @dy-tea?

Yeah it would probably be better to do this.

vlib/v/checker/struct.v
dy-tea: Yeah this code is a bit ugly, I will look and see about putting it i n v.ast

dy-tea:
Just added the `ngname` field to `TypeSymbol` that contains the non generic name, will look into moving that other chunk of code to v.ast

dy-tea:
Moved the code into `v.ast`, also noticed there were some `TypeSymbol` creation in `v.parser` that was missing `ngname` init so added that too.

PR #25703, opened 2025-11-09
builder: vls-mode can compile all src dir files

kbkpbot:
Fix https://github.com/vlang/vls/issues/473

work with PR https://github.com/vlang/vls/pull/475

### DIFF
--- a/vlib/v/builder/compile.v
+++ b/vlib/v/builder/compile.v
@@ -373,27 +373,36 @@ pub fn (v &Builder) get_user_files() []string {
 	if !does_exist {
 		verror("${dir} doesn't exist")
 	}
-	is_real_file := does_exist && !os.is_dir(dir)
-	resolved_link := if is_real_file && os.is_link(dir) { os.real_path(dir) } else { dir }
-	if is_real_file && (dir.ends_with('.v') || resolved_link.ends_with('.vsh')
-		|| v.pref.raw_vsh_tmp_prefix != '' || dir.ends_with('.vv')) {
-		single_v_file := if resolved_link.ends_with('.vsh') { resolved_link } else { dir }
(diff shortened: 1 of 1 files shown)


spytheman:
Why add a special support for this, and pass a different `-path`, when you can `v /full/path` instead, which will also get all the `.v` files in the folder as part of the same program?

JalonSolov:
I'm not sure I like the idea of having vls automatically read all the files in a dir anyway.  Perhaps it could be an option?

I've had dirs with multiple completely unrelated V "script" files, which would show a _lot_ of errors if they were all parsed together (for instance, all of them having `fn main()`, etc.)

spytheman:
The VLS mode should work as similar as possible with V itself, i.e. if V supports `v folder/`, then the vls mode should be able to `v -some -vls -specific -options folder/` too as a way of specifying what the full project/context is.

JalonSolov:
V can already compile a directory of files, by either `v directory/` or `v .` if the directory is the current one.

Having it compile a specific list of files... possible, of course, but it sounds like it would be complicated to get right, and to document clearly.

spytheman:
@kbkpbot I do not understand, why not just use `v folder/`, which *already* works?

Why do you insist on implementing more options, that will do that exact thing?

spytheman:
> Currently, vls only supports the analysis of individual files and does not incorporate concepts such as a "workspace" or "project." I think it may be necessary to first modify the code on the vls side so that when a directory is opened, it automatically creates a workspace or project. Then, when using vls, the entire path can be passed to the v compiler. This way, v can compile all the .v files in the entire directory, including any other user-defined modules located within it.

That is exactly what `v folder/` does. VLS can take advantage of that, and just pass its query option to tell V in what info (file/location) it is interested in.

PR #25704, opened 2025-11-09
io.string_reader: fix read_line(), add tests

tankf33der:
The `read_line()` function works quite well, but managed to break the newline handler and create a working test on how to reproduce this.
Only the `test_read_line_8()` fails on the `master` branch.


spytheman (approved):
Excellent work.
Thank you 🙇🏻 .

spytheman:
@tankf33der what do you think of making `get_string_part` return Eof for negative `n` ?

tankf33der:
> @tankf33der what do you think of making `get_string_part` return Eof for negative `n` ?

I will very soon get to these functions and consider this idea.

Issue #25452, opened 2025-10-05
Bad error when declaring that a field is a vec3[f64] instead of a Vec3


Linklancien:
### Describe the bug

When we declare a struct with a field type vec3 the compiler doesn't report that it should be a Vec3 instead

### Reproduction Steps

import math.vec {vec3, Vec3}

struct Struct_veca {
	field vec3[f64]
}

fn main(){
	f := vec3[f32](2_000, 0 ,0)
}

### Expected Behavior

The deboger should indique that vec3[f64] is not a type and instead suggest that it may be a Vec3

### Current Behavior

The code return an error in the vec3 file

```/v/vlib/math/vec/vec3.v:384:19: error: unexpected token `,`, expecting `)`
  382 |     m := v.magnitude()
  383 |     if m == 0 {
  384 |         return vec3[T](0, 0, 0)
      |                         ^
  385 |     }
  386 |     return Vec3[T]{```

### V version

V 0.4.12 f6e5118

### Environment details (OS name and version, etc.)

linux, Linux version 6.11.7-300.fc41.x86_64 (mockbuild@6c49ea81e72a4908b02fd96c22f92edf) (gcc (GCC) 14.2.1 20240912 (Red Hat 14.2.1-3), GNU ld version 2.43.1-2.fc41) #1 SMP PREEMPT_DYNAMIC Fri Nov  8 19:23:10 UTC 2024
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

**Comment by jorgeluismireles** [comment]
`vec3` is a function and `Vec3` is a struct. Setting `field` as `Vec3[f64]` compiles:
```V
import math.vec {vec3, Vec3}

struct Struct_veca {
	field Vec3[f64]
}

fn main(){
	f := vec3[f32](2_000, 0 ,0)
}
```
though seems a bug

jorgeluismireles:
Making a typo like this is catched:
```v
import math.vec

struct Struct_veca {
	field vec.vec3[f64]
}
```
```
code.v:4:12: error: imported types must start with a capital letter
    2 | 
    3 | struct Struct_veca {
    4 |     field vec.vec3[f64]
      |               ~~~~
    5 | }
```

But your's doesn't:
```v
import math.vec { vec3 }

struct Struct_veca {
	field vec3[f64]
}
```

[...]

PR #25705, opened 2025-11-09
parser: disallow using generic non struct as type inside struct decl 

Delta456:
Fixes https://github.com/vlang/v/issues/25452


spytheman:
Thank you @Delta456 🙇🏻 .

PR #25712, opened 2025-11-11
parser: warn on `@[deprecated_after]` used without a corresponding `@[deprecated]` attribute

spytheman:
- **parser: error when `@[deprecated_after]` is used without `@[deprecated]`**
- **fix vlib/v/util/diff/diff.v**
- **fix vlib/x/crypto/chacha20/chacha.v**


spytheman:
Deprecates #25480 , by requiring the use of `@[deprecated]` for each `@[deprecated_after]`.

PR #25713, opened 2025-11-11
io.string_reader: fix write(), add test

tankf33der:
The turn has come to the next function `write()` which behaves incomprehensibly to me.
Currently on the `master` branch after calling `offset` counter doesn't update and some functions break.
```v
...
reader.read_bytes(1)!
reader.write('a'.bytes())!
reader.read_bytes(1)!      <- XXX, byte ignored
...
```
My idea of PR: The `offset` counter is the number of bytes in the `builder` and must be updated after writing additional bytes.

### DIFF
--- a/vlib/io/string_reader/string_reader_test.v
+++ b/vlib/io/string_reader/string_reader_test.v
@@ -386,3 +386,16 @@ fn test_write() {
 	reader.read_bytes(1)!
 	assert reader.get_data() == [u8(49), 97, 50]
 }
+
+fn test_write_2() {
+	mut two := TwoByteReader{
+		data: '1234567890'
+	}
+	mut reader := StringReader.new(reader: two)
+	reader.write('<'.bytes())!
+	reader.read_all(false)!
+	reader.write(':'.bytes())!
+	assert reader.read_string(3)! == '345'
+	reader.write('>'.bytes())!
+	assert reader.get_string() == '<12:345>'
+}


vlib/io/string_reader/string_reader.v
```diff
@@ -240,6 +240,7 @@ pub fn (mut r StringReader) read_line(config io.BufferedReadLineConfig) !string
 
 // write implements the Writer interface
 pub fn (mut r StringReader) write(buf []u8) !int {
+	r.offset += buf.len
```
spytheman, line 243: I do not get it - `offset` is documented as:
`offset int // current offset in the buffer`

By setting it here, essentially you make it so the written data will be always ignored when reading (since the checks will show that the buffer needs refilling).

In the test later, you do not test, that what was written can be read too, but just the current content of the buffer.

spytheman: Please do some asserts, to show that what was written, can be read later as is by one of the `.read` methods, preferably `.read_string`, since that is easier to compare visually with the rest of the data in the tests.

spytheman: I am also not sure that a `StringReader` needs a .write() method at all, especially if it is not tested.
Can it be just deleted/deprecated instead?

spytheman: As it is on master, the .write() method seems to allow to inject data into the buffer, which does not seem particularly useful, but that it could be still read later.
On this PR, the .write() method will just inject data, that is later ignored, which is even less useful.

tankf33der: @spytheman 
Having a `write()` function is strategically useful. It's easy to imagine cases when you'll be accumulating in the `buffer` data that isn't in the `reader`, for example, a `separator` that you'll use to separate raw data and/or fields from the reader. Moreover, `the goal` of this PR is to fix something and make it work.

It would be more useful to prohibit adding untested code to the codebase, in which you then easily find bug(s) upon first contact. A lot of energy gets spent on this afterwards.

Returning to the PR, I vote for keeping the `write()` function in this iteration, I'll add another test with `read_string()`.

spytheman: Perhaps it may be clearer, to see this:
```v
import io
import io.string_reader

struct TwoByteReader {
mut:
        data string
        pos  int
}

fn (mut r TwoByteReader) read(mut buf []u8) !int {
        if r.pos >= r.data.len {
                return io.Eof{}
        }
        min := int_min(int_min(r.data.len - r.pos, 2), buf.len)
        for i in 0 .. min {
                buf[i] = r.data[r.pos]
                r.pos++
        }
        return min
}

fn test_write_method() {
        mut two := TwoByteReader{
                data: [u8(48), 49, 50, 51, 52, 53, 54, 55, 56, 57].bytestr()
        }
        mut reader := string_reader.StringReader.new(reader: two)
        dump(reader.write([u8(70), 70, 70])!)
        dump(reader.read_bytes(1)!) // [70] on master
        dump(reader.read_bytes(1)!) // [70] on master
        dump(reader.read_bytes(1)!) // [70] on master
        dump(reader.write([u8(80), 80, 80])!) // 80 is `P`
        dump(reader.read_line()!) // `PPP0123456789` on master
        // for { dump(reader.read_bytes(1) or { break }) }
}
```

spytheman: Producing:
```
#0 09:46:58 ^ master ~/v>./v xx_test.v
[xx_test.v:27] reader.write([u8(70), 70, 70])!: 3
[xx_test.v:28] reader.read_bytes(1)!: [70]
[xx_test.v:29] reader.read_bytes(1)!: [70]
[xx_test.v:30] reader.read_bytes(1)!: [70]
[xx_test.v:31] reader.write([u8(80), 80, 80])!: 3
[xx_test.v:32] reader.read_line(io.BufferedReadLineConfig{....})!: PPP0123456789
#0 09:47:00 ^ master ~/v>
#0 09:47:01 ^ master ~/v>
#0 09:47:01 ^ master ~/v>git switch -
Switched to branch 'write'
#0 09:47:02 ^ write ~/v>
#0 09:47:03 ^ write ~/v>
#0 09:47:03 ^ write ~/v>./v xx_test.v
[xx_test.v:27] reader.write([u8(70), 70, 70])!: 3
[xx_test.v:28] reader.read_bytes(1)!: [48]
[xx_test.v:29] reader.read_bytes(1)!: [49]
[xx_test.v:30] reader.read_bytes(1)!: [50]
[xx_test.v:31] reader.write([u8(80), 80, 80])!: 3
[xx_test.v:32] reader.read_line(io.BufferedReadLineConfig{....})!: 3456789
```

spytheman: @Casper64 can you please clarify the intent for the module/method, and what is the desired behavior? Unfortunately it is not imported by anything in vlib or the examples, so I can not infer the usecases :-| .

tankf33der:
@spytheman 

although I accidentally deleted the repo along with the branch, not only this bug but other bugs remained. `Casper64` won't come back and we need to make a difficult decision about what to do with `io.string_reader`.

spytheman:
I thought a lot about it too, and I think, that we should deprecate all the APIs in the existing `io.string_reader` module, and delete the whole module after 1-2 months.
That will give time to people that still use the old version (although I doubt there are many, if any), and at the same time, free the namespace for a future one, hopefully more intuitive/well tested one.

JalonSolov:
Or... perhaps create `io.streams`, and have string reader, file reader, etc.

spytheman:
https://github.com/vlang/v/pull/26274 adds the deprecation that I mentioned above. Given the general deprecation policy, we would have to wait till 2026/07 to remove it:
```
[21:26:49]@m1: (deprecate_io_string_reader) /opt/v $ v /Users/delyanangelov/code/misc/2026_01_06__21/a.v
/Users/delyanangelov/code/misc/2026_01_06__21/a.v:1:1: warning: module `io.string_reader` has been deprecated since 2026-01-06, it will be an error after 2026-07-05; If you use it, create a local module named `string_reader`, and copy https://github.com/vlang/v/blob/f24d49259db5d2d7211e79bcf6b6f507d22847f3/vlib/io/string_reader/string_reader.v there.
    1 | import io.string_reader as _
      | ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    2 |
0[21:27:00]@m1: (deprecate_io_string_reader) /opt/v $
```

spytheman:
> Or... perhaps create `io.streams`, and have string reader, file reader, etc.

what would just `import io.streams` provide, beside the namespace?

JalonSolov:
Depends on what's added to it.

It would be nice to have more of the Go style routines, where (mostly) every type can be read/written as a stream, but that would involve a lot more changes to other areas.

Could be as simple as things like
`streams.new_string_reader()`
`streams.new_string_writer()`
`streams.new_file_reader()`
`streams.new_file_writer()`
`streams.new_socket_reader()`
`streams.new_socket_writer()`

Or whatever else makes sense.

Issue #25714, opened 2025-11-12
Add support for capture stdout stderr of current process


kbkpbot:
### Describe the feature

This API allows for capturing the `stdout` and `stderr` streams of the current process. Unlike some methods, it does not launch a separate process.

```v
os.capture_stdout_stderr_start()
x() // Invoke a function that produces output
stdout_str, stderr_str := os.capture_stdout_stderr_stop()
```

### Use Case

The primary use case is when a process invokes a third-party library that generates its own `stdout`/`stderr` output, and the user wishes to suppress this output to avoid cluttering the console.

This is particularly useful in scenarios such as developing a new tool like `vls`, where you need to call `checker.check()` but must capture its error messages for custom handling, without allowing them to print directly to the standard error.

### Proposed Solution

Use pipe implement these functions.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.12 33e129b19e1124bb0cb6b057ee04324cc25409b4

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 33e129b19e1124bb0cb6b057ee04324cc25409b4
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 20.04.6 LTS
|Processor           |4 cpus, 64bit, little endian, aarch64
|Memory              |1.57GB/3.78GB
|                    |
|V executable        |/home/vpn/github/lang/v/v
|V last modified time|2025-11-10 02:27:13
|                    |
|V home dir          |OK, value: /home/vpn/github/lang/v
|VMODULES            |OK, value: /home/vpn/.vmodules
|VTMP                |OK, value: /tmp/v_1001
|Current working dir |OK, value: /home/vpn
|                    |
|Git version         |git version 2.25.1
|V git status        |weekly.2025.45-35-g33e129b1
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
|gcc version         |gcc (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
|clang version       |clang version 10.0.0-4ubuntu1
|tcc version         |tcc version 0.9.28rc 2025-03-11 mob@8c4e6738 (AArch64 Linux)
|tcc git status      |thirdparty-linux-aarch64 ecc2f67a-dirty
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.31-0ubuntu9.18) 2.31

[...]

PR #25716, opened 2025-11-12
os: add pipe support

kbkpbot:
Feature request by #25714
Add pipe support.

### DIFF
--- a/vlib/os/pipe.c.v
+++ b/vlib/os/pipe.c.v
@@ -100,6 +100,23 @@ mut:
 }
 
 // stdio_capture starts capturing stdout and stderr by redirecting them to pipes
+// example:
+// ```v
+//    	mut cap := os.stdio_capture()!
+//    	println('hello println')
+//    	eprintln('hello eprintln')
+//    	cap.stop()
+//    	sout := cap.stdout.slurp()
+//    	serr := cap.stderr.slurp()
+//    	cap.close()
+// ```
+// or
+// ```v
+//    	mut cap := os.stdio_capture()!
+//    	println('hello println')
+//    	eprintln('hello eprintln')
+//    	sout, serr := cap.finish()
+// ```
 pub fn stdio_capture() !IOCapture {
 	mut c := IOCapture{}
 	mut pipe_stdout := pipe()!


vlib/os/pipe.c.v
```diff
@@ -0,0 +1,201 @@
...
+			pipe_stderr.close()
+			return error('Failed to redirect stderr')
+		}
+		// Close original write ends (duplicated by dup2)
+		C._close(pipe_stdout.write_fd)
+		C._close(pipe_stderr.write_fd)
+	} $else {
+		c.original_stdout_fd = C.dup(1) // Save stdout
+		c.original_stderr_fd = C.dup(2) // Save stderr
+		if C.dup2(pipe_stdout.write_fd, 1) == -1 {
```
spytheman:
```suggestion
		if C.dup2(pipe_stdout.write_fd, C.STDOUT_FILENO) == -1 {
```

vlib/os/pipe.c.v
```diff
@@ -0,0 +1,201 @@
...
+			pipe_stdout.close()
+			pipe_stderr.close()
+			return error('Failed to redirect stdout')
+		}
+		if C.dup2(pipe_stderr.write_fd, 2) == -1 {
```
spytheman:
```suggestion
		if C.dup2(pipe_stderr.write_fd, C.STDERR_FILENO) == -1 {
```

vlib/os/pipe.c.v
```diff
@@ -0,0 +1,201 @@
...
+			C.dup2(c.original_stdout_fd, 1) // Restore stdout
+			pipe_stdout.close()
+			pipe_stderr.close()
+			return error('Failed to redirect stderr')
+		}
+		// Close original write ends (duplicated by dup2)
+		C.close(pipe_stdout.write_fd)
+		C.close(pipe_stderr.write_fd)
+	}
```
spytheman: this can be simplified/unified, if you have a private wrapper for C.dup2 and C._dup2, with the comptime check in it

vlib/os/pipe.c.v
```diff
@@ -0,0 +1,201 @@
...
+			C.close(c.original_stdout_fd)
+			c.original_stdout_fd = -1
+		}
+
+		if c.original_stderr_fd != -1 {
+			C.dup2(c.original_stderr_fd, 2)
+			C.close(c.original_stderr_fd)
+			c.original_stderr_fd = -1
+		}
+	}
```
spytheman: same - if you have a wrapper for the dup2, this can be unified

vlib/os/pipe.c.v
```diff
@@ -0,0 +1,201 @@
...
+		for {
+			n := c.stdout_pipe.read(mut buffer) or { break }
+			if n == 0 {
+				break
+			}
+			stdout_result.write(buffer[0..n])!
+		}
+		c.stdout_pipe.close()
+		c.stdout_pipe = unsafe { nil }
+	}
```
spytheman: if you extract a helper method on the Pipe, that reads all the content and returns a string builder or a string, you can just call it 2 times here:
```v
stdout_result := c.stdout_pipe.slurp()
stderr_result := c.stderr_pipe.slurp()
```

vlib/os/pipe.c.v
```diff
@@ -0,0 +1,201 @@
...
+	result := C.write(p.write_fd, buffer.data, buffer.len)
+	if result == -1 {
+		return error('Write failed')
+	}
+	return result
+}
+
+// Capture manages redirection of standard output and error streams
+@[noinit]
+pub struct Capture {
```
spytheman: `os.Capture` does not seem very clear to me 🤔 ... 
What do you think of `os.IOCapture` ?

vlib/os/pipe.c.v
```diff
@@ -0,0 +1,201 @@
...
+	original_stderr_fd int   = -1
+}
+
+// Capture.new creates a new `Capture`
+pub fn Capture.new() Capture {
+	return Capture{}
+}
+
+// stdout_stderr_capture_start starts capturing stdout and stderr by redirecting them to pipes
+pub fn (mut c Capture) stdout_stderr_capture_start() ! {
```
spytheman: How about just:
```suggestion
pub fn (mut c Capture) start() ! {
```

vlib/os/pipe.c.v
```diff
@@ -0,0 +1,201 @@
...
+	pipe_stdout.write_fd = -1
+	pipe_stderr.write_fd = -1
+
+	// Store pipes for later reading
+	c.stdout_pipe = &pipe_stdout
+	c.stderr_pipe = &pipe_stderr
+}
+
+// stdout_stderr_capture_stop stops capturing and returns the captured stdout and stderr content
+pub fn (mut c Capture) stdout_stderr_capture_stop() !(string, string) {
```
spytheman: How about just:
```suggestion
pub fn (mut c Capture) stop() !(string, string) {
```

vlib/os/pipe.c.v
```diff
@@ -0,0 +1,201 @@
...
+pub struct Capture {
+mut:
+	stdout_pipe        &Pipe = unsafe { nil }
+	stderr_pipe        &Pipe = unsafe { nil }
+	original_stdout_fd int   = -1
+	original_stderr_fd int   = -1
+}
+
+// Capture.new creates a new `Capture`
+pub fn Capture.new() Capture {
```
spytheman: `c := os.stdio_capture()` looks more natural to me, compared to:
`c := os.Capture.new()`

@JalonSolov what do you think?

spytheman: It is a bit similar to `os.pipe()` (`Pipe` also does not have a static new() method).

spytheman: Also, if the factory function calls .start() or has its code directly, the usage sites will be simplified, i.e. the user will have to call just 1 thing before and 1 thing after the block that produces output, instead of 2 things before, and 1 thing after.

Is there an usecase for just constructing a new Capture, without then immediately calling .start() on it?

vlib/os/pipe.c.v
spytheman: afaik STDOUT_FILENO and STDERR_FILENO are macros that can expand to function calls, not just plain integer literals on some systems

vlib/os/pipe.c.v
JalonSolov: > `c := os.stdio_capture()` looks more natural to me, compared to: `c := os.Capture.new()`
> 
> @JalonSolov what do you think?

Agreed.  `Capture.new` has no context for what might be captured.

spytheman:
> An example usage:
> 
> ```v
> 	mut cap := os.stdio_capture()!
> 	println('hello println')
> 	eprintln('hello eprintln')
> 	cap.stop()
> 	stdout_str := cap.stdout.slurp()
> 	stderr_str := cap.stderr.slurp()
> 	cap.close()
> ```

I like it 👍🏻 .

spytheman:
Can it also have a utility method .finish() that returns both strings, and closes the pipes?
i.e. allowing for:
```v
mut cap := os.stdio_capture()!
println('hello println')
eprintln('hello eprintln')
sout, serr := cap.finish()
```

spytheman:
Thank you @kbkpbot 🙇🏻 .

PR #25718, opened 2025-11-12
math.unsigned: fix put_bytes()

tankf33der:
The two-line patch speaks for itself and test can not pass on `master`.

p.s. I'm not sure if I have the right(s) to include and rely on the `rand` module for random numbers, but it can always be redone.


spytheman:
> p.s. I'm not sure if I have the right(s) to include and rely on the rand module for random numbers, but it can always be redone.

There is no problem - you can import whatever you need in the tests; it does not affect other users of the module (each _test.v file is compiled as a separate program, independent from the rest).

PR #25737, opened 2025-11-14
db.pg: Support returning `Result` type with column names

omerrob:
Currently, `db.pg` queries return `[]Row`, which is just 
```
pub struct Row {
pub mut:
	vals []?string
}
```
Then the developer needs to map the values based on ordinal places.

I understand that this was built with the ORM in mind, and the ORM maintains a mapping of the ordinal places with the struct members.

Sometimes we don't want to use the ORM or can't because we use a lot of `join`s and CTEs (WITH clauses). In that case, it is convenient to have the column names in order to map the query.

This PR adds the `Result` struct:
```
pub struct Result {
pub:
	cols map[string]int
	rows []Row
}
```
 and adds matching `fn *_result()` functions to the currently implemented functions.

It also adds a quality-of-life mapping method to `Result`:
```
fn (res Result) as_structs[T](mapper fn (Result, Row) !T) ![]T
```
that accepts a mapper function to return a typed `struct`.

### DIFF
--- a/.github/workflows/windows-install-sqlite.bat
+++ b/.github/workflows/windows-install-sqlite.bat
@@ -1,11 +1,11 @@
 @echo off
 
-curl -L https://www.sqlite.org/2024/sqlite-amalgamation-3450300.zip -o sqlite-amalgamation-3450300.zip
+curl -L https://sqlite.org/2025/sqlite-amalgamation-3510000.zip -o sqlite-amalgamation-3510000.zip
 
-unzip sqlite-amalgamation-3450300.zip -d thirdparty\
+unzip sqlite-amalgamation-3510000.zip -d thirdparty\
 
-del thirdparty\sqlite-amalgamation-3450300\shell.c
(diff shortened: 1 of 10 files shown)


spytheman:
https://github.com/vlang/v/actions/runs/19352895759/job/55368240782?pr=25737 shows that the PR needs the new `pub` APIs to be documented.
Please add a small comment on top of each, starting with the function/method name.

omerrob:
Added function documentation.

JalonSolov:
None of those "Unhandled Exception" codes are valid codes.

omerrob:
Sorry, not sure what you mean by "Unhandled Exception" codes.

JalonSolov:
Not you - in the CI failure logs.

spytheman:
The CI is a bit flaky unfortunately in the last few days :-| .

JalonSolov:
Merge master again.  Should take care of the Windows failures, at least.

spytheman:
Thank you @omerrob 🙇🏻 .

Issue #25732, opened 2025-11-13
Unexpected value when nesting `or { }` blocks


gergelyk:
### Describe the bug

Unwrapping option in a nested `or { }` blocks seem to return unexpected value. This behavior changed with weekly.2024.47. It used to work well in the previous releases.

### Reproduction Steps

```
module main

fn x1() ?int {
    return none
}
fn x2() ?int {
    return none
}
fn x3() ?int {
    return none
}

fn def() int {
    return 123
}

fn main() {
    y := x1() or { x2() or { x3() or { def() } }  }
    println('y=${y}')
}
```

### Expected Behavior

`y=123` printed to stdout

### Current Behavior

`y=0` printed to stdout

### V version

0.4.12

### Environment details (OS name and version, etc.)

linux-x86_64

PR #25739, opened 2025-11-14
parser: fix last stmt is fn call in or_expr

kbkpbot:
Fix issue #25732

In `mark_last_call_return_as_used()`, if the expr has a `or_expr`, we need to check and set the last expr in the `or_expr` as used.


spytheman (approved):
Excellent work.

Issue #25740, opened 2025-11-14
c error


gw-dev101:
### Describe the bug

got a c error while trying to compile a v program 

### Reproduction Steps
```v
a := unsafe{*nil}
```
### Expected Behavior

No c error
probably a segfault

### Current Behavior

Running code...
Can't run code. The server returned an error:
```
code.v:1:1: warning: unused variable: `a`
    1 | a := unsafe{*nil}
      | ^
/box/code.v:1: error: cannot convert 'void' to 'void *'
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
Exited with error status 1
Please try again.
```
### V version

V 0.4.12 b39a585

### Environment details (OS name and version, etc.)

(playground)
|V full version      |V 0.4.12 b39a585b0763f214a697bda40f6965e7c8fb08ea
|:-------------------|:-------------------
|OS                  |linux, Debian GNU/Linux 12 (bookworm) (VM)
|Processor           |2 cpus, 64bit, little endian, Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz
|Memory              |0.74GB/2.02GB
|                    |
|V executable        |/home/admin/v/v
|V last modified time|2025-11-14 13:25:35
|                    |
|V home dir          |NOT writable, value: /home/admin/v
|VMODULES            |OK, value: /box/.vmodules
|VTMP                |OK, value: /tmp/v_60000
|Current working dir |OK, value: /box
|                    |
|Git version         |git version 2.39.5
|V git status        |N/A
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |gcc (Debian 12.2.0-14+deb12u1) 12.2.0
|clang version       |N/A
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v/thirdparty/tcc
 Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v/thirdparty/tcc

|emcc version        |N/A
[...]

PR #25746, opened 2025-11-16
checker: disallow deference nil

kbkpbot:
Fix issue #25740


spytheman (approved):
Excellent work.
Thank you @kbkpbot .

PR #25748, opened 2025-11-16
encoding.utf8.validate: fix validation, add test

tankf33der:
Module testing showed incorrect operation.
When iterating through all combinations of 4-byte code points, it produces an error `186k` (!) times. The test I added that cannot pass correctly on `master` is a code point that vlang incorrectly considers valid.

I propose to replace the massive implementation with `Björn Höhrmann's` table-based [method](http://bjoern.hoehrmann.de/utf-8/decoder/dfa/) which has stood the test of time.

I ported the method and tested it against `3` (!) different libraries for all combinations of 1-4 bytes.
List of libraries: `glib`, [c-utf8-valid](https://github.com/chansen/c-utf8-valid), [is_utf8](https://github.com/simdutf/is_utf8)

### DIFF
--- a/vlib/encoding/utf8/validate/validate_utf8.v
+++ b/vlib/encoding/utf8/validate/validate_utf8.v
@@ -24,10 +24,12 @@ const utf8d = [
 	12, 12, 12 ]!
 // vfmt on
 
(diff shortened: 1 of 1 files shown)


spytheman (commented):
Excellent!
Thank you @tankf33der 🙇🏻 .

Issue #25707, opened 2025-11-10
Add `-file_list` compiler option for flexible multi-file compilation


kbkpbot:
### Describe the feature

Currently, the V compiler supports two main compilation modes:

- `v dir` - compiles all `.v` files in a directory

- `v a.v`- compiles a single file

I propose adding a new `-file_list` option to provide more flexible control over which files are included in compilation, particularly useful for projects with multiple entry points or specific file combinations.

```
v a.v -file_list "b.v,c.v,dir"
```

Where:

`a.v` is the entry point/main file

`-file_list` accepts a comma-separated list containing:

- Individual `.v` files (e.g., `b.v`, `c.v`)

- Directories (e.g., `dir`) - all `.v` files within the directory will be included

### Use Case

1. Multiple entry points: Compile different main files with shared modules without including unrelated files
2. Selective compilation: Include specific files from different directories
3. Build flexibility: Better control over compilation scope for complex projects
4. IDE/tool integration: Enable language servers to compile exact file subsets

```
# Compile main1.v with specific helpers and a utils directory
v main1.v -file_list "helper.v,utils/"

# Compile test runner with test files from multiple directories  
v test_runner.v -file_list "tests/unit/,tests/integration/,test_helpers.v"

# Complex project with modules from different locations
v app.v -file_list "lib/core/,lib/ui/,config.v,third_party/"
```

Some discussion at #25703

### Other Information

Alternatives Considered

`-files` (shorter but less descriptive)
`-extra_files` (misleading since first file is the main entry point)

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.12 af62364a29312a09f76aa9ad3f734790e534c162.982681c

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 af62364a29312a09f76aa9ad3f734790e534c162.982681c
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |12.26GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-11-09 10:47:54
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
[...]

PR #25749, opened 2025-11-16
builder,pref: add  `-file-list` support

kbkpbot:
Feature request by #25707


spytheman:
Why not use `,` as a separator, instead of `;` ?

PR #25752, opened 2025-11-16
atof.c: handle subnormal numbers

### LINKED ISSUE #25751: Float Underflow with string.f64()
### Describe the bug

V's `string.f64()` method has a bug that causes underflow for subnormal (denormalized) floats in scientific notation, incorrectly returning `0.0` instead of the actual value.

### Affected Range

Subnormal (denormalized) floats near the lower limit of f64:
- **f64 min normal**: `2.2250738585072014e-308`
- **f64 min subnormal**: `5e-324`
- **Affected values**: Approximately between `1e-308` and `2.2e-308`

### Reproduction Steps

```v
fn main() {
    val1 := '1.23e-308'.f64()
    val2 := 1.23e-308

    println('string.f64(): ${val1}')  // Prints: 0.0 ❌
    println('Literal:      ${val2}')  // Prints: 1.23e-308 ✅

    println('Are they equal? ${val1 == val2}')  // false
}
```

### Expected Behavior

```
string.f64(): 1.23e-308
Literal:      1.23e-308
Are they equal? true
```

### Current Behavior

```
string.f64(): 0.0
Literal:      1.23e-308
Are they equal? false
```

### Possible Solution

My current workaround is to use `C.strtod()`, which correctly handles subnormal numbers:

```v
fn C.strtod(&char, &&char) f64

fn main() {
    s :='1.23e-308'
    val1 := C.strtod(&char(s.str), unsafe { nil })
    val2 := 1.23e-308

    println('string.f64(): ${val1}')  // Prints: 1.23e-308 ✅
    println('Literal:      ${val2}')  // Prints: 1.23e-308 ✅

    println('Are they equal? ${val1 == val2}')  // true
}
```

_No response_

### V version

v 0.4.12

macOS 25.0.0 Darwin
> [!NOTE]
>

davlgd:
This PR fixes #25751 and add tests for subnormal numbers conversion

### DIFF
--- a/vlib/strconv/atof.c.v
+++ b/vlib/strconv/atof.c.v
@@ -298,6 +298,30 @@ fn converter(mut pn PrepNumber) u64 {
 			s0 = q0
 		}
 	}
+
+	// Handle subnormal (denormalized) numbers - very small numbers near zero
+	//
+	// Normal floats have an implicit leading 1 bit in their mantissa (like 1.xxxxx).
+	// When numbers get too small (binexp < -1022), we can't represent them normally.
+	// Instead, we use subnormals: set exponent to 0 and shift the mantissa right,
+	// losing precision gradually. This prevents abrupt underflow to zero.
+	//
+	// Example: 1.23e-308 is smaller than the minimum normal float, so we:
+	// 1. Keep the normalized mantissa from s2 and s1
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Excellent work.
Thank you @davlgd 🙇🏻‍♂️ .

penguindark:
Good job! 👍 
Thank you 🙇.

Issue #25729, opened 2025-11-13
checker: variable from return value of `array.sorted(a < b)` cannot be named `a` or `b`


dy-tea:
### Describe the bug
```v
arr := [4 2 1 3]
a := arr.sorted(a < b)
println(a)
```

### Reproduction Steps

Included code.

### Expected Behavior

Should compile.

### Current Behavior

Output:

```
code.v:2:17: error: undefined variable: `a`
    1 | arr := [4 2 1 3]
    2 | a := arr.sorted(a < b)
      |                 ^
    3 | println(a)
Exited with error status 1
```

### V version

V 0.4.12 4f773fc9fdbfc02a692f6fd635b61645732a3335

### Environment details (OS name and version, etc.)

```
|V full version      |V 0.4.12 4f773fc9fdbfc02a692f6fd635b61645732a3335
|:-------------------|:-------------------
|OS                  |linux, Debian GNU/Linux 12 (bookworm) (VM)
|Processor           |2 cpus, 64bit, little endian, Intel(R) Xeon(R) CPU E5-2680 v4 @ 2.40GHz
|Memory              |0.76GB/2.02GB
|                    |
|V executable        |/home/admin/v/v
|V last modified time|2025-11-13 17:25:31
|                    |
|V home dir          |OK, value: /home/admin/v
|VMODULES            |OK, value: .vmodules
|VTMP                |OK, value: /tmp/v_0
|Current working dir |OK, value: /home/admin/playground
|                    |
|Git version         |git version 2.39.5
|V git status        |N/A
|.git/config present |true
|                    |
|cc version          |cc (Debian 12.2.0-14+deb12u1) 12.2.0
|gcc version         |gcc (Debian 12.2.0-14+deb12u1) 12.2.0
|clang version       |Debian clang version 14.0.6
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v/thirdparty/tcc
 Error: fatal: detected dubious ownership in repository at '/home/admin/v/thirdparty/tcc'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/admin/v/thirdparty/tcc

|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.36-9+deb12u10) 2.36
```
[...]

PR #25755, opened 2025-11-17
parser: allow use `a` `b` `it` as var name when using array builtin methods

kbkpbot:
Fix issue #25729

I think further check, for example, left expr is an array/fixed array, should be done in `checker` and not in `parser`.

### DIFF
--- a/vlib/crypto/blake2b/blake2b_block_generic.v
+++ b/vlib/crypto/blake2b/blake2b_block_generic.v
@@ -11,7 +11,7 @@ module blake2b
 import math.bits
 
 // mixing function g
-@[inline]
+@[direct_array_access; inline]
 fn g(mut v []u64, a u8, b u8, c u8, d u8, x u64, y u64) {
 	v[a] = v[a] + v[b] + x
 	v[d] = bits.rotate_left_64((v[d] ^ v[a]), nr1)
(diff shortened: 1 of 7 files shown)


spytheman:
I think that the parser should do very little amount of checks of this kind (semantic ones). Ideally, it should be only concerned with syntax errors.

vlib/v/parser/assign.v
```diff
@@ -49,6 +49,10 @@ fn (mut p Parser) check_undefined_variables(names []string, val ast.Expr) ! {
 		}
 		ast.CallExpr {
 			p.check_undefined_variables(names, val.left)!
+			// arr.sort(a < b) , arr.sorted(a < b)
+			if val.name in ['sort', 'sorted'] && val.args.len == 1 {
+				return
+			}
```
spytheman: Can we handle this the same way?
```v
it := [2,3,4].map(it*2)
dump(it)
```
which currently produces:
```
a.v:1:19: error: undefined variable: `it`
    1 | it := [2,3,4].map(it*2)
      |                   ~~
    2 | dump(it)
```

spytheman: Also what is the performance impact of this additional check?
(measured by `./v run .github/workflows/compare_pr_to_master.v`)

spytheman: I suspect that it will be faster, since it avoids a recursive call, but on the other hand, it does make 2 additional string comparisons.

vlib/v/parser/assign.v
```diff
@@ -49,6 +49,10 @@ fn (mut p Parser) check_undefined_variables(names []string, val ast.Expr) ! {
 		}
 		ast.CallExpr {
 			p.check_undefined_variables(names, val.left)!
+			// arr.sort(a < b) , arr.sorted(a < b)
+			if val.name in ['sort', 'sorted'] && val.args.len == 1 {
```
spytheman:
```suggestion
			if val.args.len == 1 && val.name in ['sort', 'sorted'] {
```
the integer check is cheaper, and is better to be first

PR #25758, opened 2025-11-17
veb: gzip support for StaticHandler

### LINKED ISSUE #20859: vweb: add compression support for static files
### Describe the feature

It could be great to be able to declare a compression method ('gzip' or 'brotli') to serve static files. 

### Use Case

It's now default for web server to serve compressed files, especially for static files. As it seems not to be possible through vweb, it's a downside for this great lib. 

### Proposed Solution

_No response_

### Other Information

I tried to use the `encode_gzip` middleware but I got an error: 

```
/libexec/vlib/x/vweb/middleware.v:128:11: error: field `Context.return_type` is not public
  126 |         handler: fn [T](mut ctx T) bool {
  127 |             // TODO: compress file in streaming manner, or precompress them?
  128 |             if ctx.return_type == .file {
      |                    ~~~~~~~~~~~
  129 |                 return true
  130 |             }
```

And it seems it's not for static files, as this is mentionned in the `middleware.v`: 

```
// encode_gzip adds gzip encoding to the HTTP Response body.
// This middleware does not encode files, if you return `ctx.file()`.
```

### Acknowledgements

### Version used

V 0.4.4 c256013

macOS (ARM64) 14.3.1

davlgd:
This PR removes the "skip gzip if static file" behavior by adding two features:
- If content was precompressed, serves the `.gz` variant
- If gzip compression is enabled and file under a thresold, lazy compression
  - It's compressed one
  - After that it's served as precompressed files

I've added tests and a documentation example for this. Feel free to tell me if you need some adjusments or a different behavior

Fix #20859

### DIFF
--- a/vlib/veb/context.v
+++ b/vlib/veb/context.v
@@ -37,9 +37,7 @@ mut:
 	static_gzip_max_size int
 	// if true the response should not be sent and the connection should be closed
 	// manually.
-	takeover bool
-	// how the http response should be handled by veb's backend
-	return_type ContextReturnType = .normal
+	takeover    bool
 	return_file string
 	// already_compressed indicates that the response body is already gzip-compressed
 	// and the encode_gzip middleware should skip it
@@ -50,6 +48,8 @@ pub:
 	// You can use it to determine how much time is spent on your request.
 	page_gen_start i64
 pub mut:
+	// how the http response should be handled by veb's backend
+	return_type       ContextReturnType = .normal
 	req               http.Request
 	custom_mime_types map[string]string
 	// TCP connection to client. Only for advanced usage!
--- a/.github/workflows/windows-install-sqlite.bat
+++ b/.github/workflows/windows-install-sqlite.bat
@@ -1,11 +1,11 @@
 @echo off
 
-curl -L https://www.sqlite.org/2024/sqlite-amalgamation-3450300.zip -o sqlite-amalgamation-3450300.zip
+curl -L https://sqlite.org/2025/sqlite-amalgamation-3510000.zip -o sqlite-amalgamation-3510000.zip
 
-unzip sqlite-amalgamation-3450300.zip -d thirdparty\
+unzip sqlite-amalgamation-3510000.zip -d thirdparty\
 
-del thirdparty\sqlite-amalgamation-3450300\shell.c
+del thirdparty\sqlite-amalgamation-3510000\shell.c
 
-move /y thirdparty\sqlite-amalgamation-3450300 thirdparty\sqlite
+move /y thirdparty\sqlite-amalgamation-3510000 thirdparty\sqlite
 
 dir thirdparty\sqlite
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -921,16 +921,16 @@ f2 := 456e+2 // 45600
 
 An array is a collection of data elements of the same type. An array literal is a
 list of expressions surrounded by square brackets. An individual element can be
-accessed using an *index* expression. Indexes start from `0`:
+accessed using an *index* expression. Indexing starts from `0`.
 
 ```v
-mut nums := [1, 2, 3]
-println(nums) // `[1, 2, 3]`
-println(nums[0]) // `1`
-println(nums[1]) // `2`
+mut nums := [10, 20, 30]
+println(nums) // `[10, 20, 30]`
+println(nums[0]) // `10`
+println(nums[1]) // `20`
 
 nums[1] = 5
-println(nums) // `[1, 5, 3]`
+println(nums) // `[10, 5, 30]`
 ```
 
 <a id='array-operations'></a>
@@ -4214,8 +4214,8 @@ fn pass_time(w World) {
 
 ### Option/Result types and error handling
 
-Option types are for types which may represent `none`. Result types may
-represent an error returned from a function.
+Option types can represent a value or `none`. Result types may
+represent a value, or an error returned from a function.
 
 `Option` types are declared by prepending `?` to the type name: `?Type`.
 `Result` types use `!`: `!Type`.
@@ -4467,9 +4467,9 @@ post := posts_repo.find_by_id(1)? // find_by_id[Post]
 ```
 
 Currently generic function definitions must declare their type parameters, but in
-future V will infer generic type parameters from single-letter type names in
-runtime parameter types. This is why `find_by_id` can omit `[T]`, because the
-receiver argument `r` uses a generic type `T`.
+future versions, V will infer generic type parameters from single-letter type names in
+runtime parameter types. This is why the `find_by_id(1)` calls above can omit `[T]`,
+because the receiver argument `r` in the method declaration, uses a generic type `T`.
 
 Another example:
 
@@ -5135,10 +5135,11 @@ Remaining small percentage of objects is freed via GC. The developer doesn't nee
 anything in their code. "It just works", like in Python, Go, or Java, except there's no
 heavy GC tracing everything or expensive RC for each object.
 
-For developers willing to have more low level control, memory can be managed manually with
+For developers willing to have more low-level control, memory can be managed manually with
 `-gc none`.
 
-Arena allocation is available via v `-prealloc`.
+Arena allocation is available via a `-prealloc` flag. Note: currently this mode is only
+suitable to speed up short lived, single-threaded, batch-like programs (like compilers).
 
 ### Control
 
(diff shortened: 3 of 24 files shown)


vlib/veb/tests/static_gzip_test.v
```diff
@@ -0,0 +1,155 @@
+import veb
+import net.http
+import os
+import time
+import compress.gzip
+
+const port = 13010
```
spytheman:
```suggestion
const port = 14013
```
Port 13010 can be used by `vlib/x/sessions/tests/session_app_test.v`, which can be a problem, if it is run at the same time (`v test .` runs tests in parallel, which is non deterministic). Afaik 14013 is currently unused.

vlib/veb/tests/static_gzip_test.v
```diff
@@ -0,0 +1,155 @@
...
+
+	app.handle_static('testdata_gzip', true) or { panic(err) }
+
+	// Add gzip middleware
+	app.use(veb.encode_gzip[Context]())
+
+	spawn veb.run_at[App, Context](mut app, port: port, timeout_in_seconds: 25, family: .ip)
+	_ := <-app.started
+	// Give server time to fully start
+	time.sleep(500 * time.millisecond)
```
spytheman: This should not be needed, since `_ := <-app.started` will block, until the `before_accept_loop` method is called, and at that point, the server thread is guaranteed to be already listening.

spytheman:
```suggestion
```

vlib/veb/tests/static_gzip_test.v
```diff
@@ -0,0 +1,155 @@
...
+}
+
+fn test_gz_file_cache_creation() {
+	// First request creates .gz cache file
+	mut req := http.new_request(.get, '${localserver}/test.txt', '')
+	req.add_header(.accept_encoding, 'gzip')
+	_ := req.do()!
+
+	// Give it a moment to write the file
+	time.sleep(100 * time.millisecond)
```
spytheman: Why does it need to wait here?
The request should be finished after the .do() line 🤔 .

vlib/veb/static_handler.v
```diff
@@ -15,6 +17,18 @@ pub mut:
...
+	// 1. Serve existing .gz files in zero-copy streaming mode (manual pre-compression)
+	// 2. Auto-generate .gz files for files < static_gzip_max_size (lazy compression cache)
+	// 3. Validate .gz freshness (regenerate if source file is newer)
+	// Files larger than the threshold are served uncompressed in streaming mode.
+	// Default: false (for backward compatibility)
+	enable_static_gzip bool
+	// static_gzip_max_size sets the maximum file size (in KB) for auto-compression.
+	// Files larger than this threshold will not be auto-compressed (but manual .gz files are still served).
+	// Default: 1024 (1MB). Set to 0 to use the default value.
+	static_gzip_max_size int = 1024
```
spytheman: imho, if that is in bytes, instead of KB, it will be more intuitive (no need for clarifications about the unit size) and it would avoid a runtime multiplication later.

spytheman:
```suggestion
	static_gzip_max_size int = 1024*1024
```

vlib/veb/README.md
```diff
@@ -419,6 +419,110 @@ app.static_mime_types['.what'] = 'txt/plain'
...
+pub fn (mut app App) index(mut ctx Context) veb.Result {
+	return ctx.html('<h1>Gzip compression demo</h1><p>Visit <a href="/app.js">/app.js</a> or <a href="/style.css">/style.css</a></p>')
+}
+
+fn main() {
+	mut app := &App{}
+
+	// Enable static file gzip compression (disabled by default)
+	app.enable_static_gzip = true
+	app.static_gzip_max_size = 1024 // Maximum file size in KB for auto-compression (default: 1024)
```
spytheman:
```suggestion
	app.static_gzip_max_size = 1024*1024 // Maximum file size in bytes for auto-compression (default: 1MB)
```

vlib/veb/context.v
```diff
@@ -200,22 +207,93 @@ fn (mut ctx Context) send_file(content_type string, file_path string) Result {
 	}
 	file.close()
 
+	// Check if client accepts gzip encoding
+	accept_encoding := ctx.req.header.get(.accept_encoding) or { '' }
+	client_accepts_gzip := accept_encoding.contains('gzip')
+
+	max_size_bytes := ctx.static_gzip_max_size * 1024 // Convert KB to bytes
```
spytheman:
```suggestion
	max_size_bytes := ctx.static_gzip_max_size
```

vlib/veb/context.v
```diff
@@ -200,22 +207,93 @@ fn (mut ctx Context) send_file(content_type string, file_path string) Result {
...
+				ctx.send_response_to_client(content_type, '')
+
+				// Get .gz file size for Content-Length header
+				mut gz_file := os.open(gz_path) or {
+					eprintln('[veb] error opening .gz file: ${err.msg()}')
+					return ctx.server_error('could not read compressed resource')
+				}
+				gz_file.seek(0, .end) or { 0 }
+				gz_size := gz_file.tell() or { 0 }
+				gz_file.close()
```
spytheman: Why opening/seeking to the end, instead of `os.file_size(gz_path)` or `os.stat(gz_path)!.size` ?

vlib/veb/context.v
```diff
@@ -200,22 +207,93 @@ fn (mut ctx Context) send_file(content_type string, file_path string) Result {
...
+				// Fallback: serve uncompressed in streaming mode
+				ctx.return_type = .file
+				ctx.return_file = file_path
+				ctx.send_response_to_client(content_type, '')
+				ctx.res.header.set(.content_length, file_size.str())
+				return Result{}
+			}
+
+			// Try to save compressed version for future requests (ignore errors)
+			os.write_file(gz_path, compressed.bytestr()) or {
```
spytheman, line 261: This can be a problem for some readonly filesystems 🤔, although for such cases, I guess the app can just set the `static_gzip_max_size` to -1 to disable the feature.

vlib/veb/context.v
```diff
@@ -200,22 +207,93 @@ fn (mut ctx Context) send_file(content_type string, file_path string) Result {
...
-		ctx.send_response_to_client(content_type, '')
-		ctx.res.header.set(.content_length, file_size.str())
-		return Result{}
 	}
+
+	// Default: serve uncompressed file in streaming mode (zero-copy sendfile)
+	ctx.return_type = .file
+	ctx.return_file = file_path
+	ctx.send_response_to_client(content_type, '')
+	ctx.res.header.set(.content_length, file_size.str())
```
spytheman: I think the order here should be reversed (to send the headers first), but I have not tested 🤔 . All the existing usages do something like `return ctx.send_response_to_client()` i.e. send the body back as the last thing done in the handlers.

davlgd:
@spytheman Thanks for the review. I've implemented fixes in dedicated commits to make it easier for you to see changes, I'll squash when it will be ok for you. 
- 1 commits for minor changes
- 1 commit for read-only management

Let me know if you see any better way to test read-only for windows (I don't have a Windows system here to check)

vlib/veb/context.v
```diff
@@ -200,22 +207,98 @@ fn (mut ctx Context) send_file(content_type string, file_path string) Result {
...
+				// Serve existing .gz file in streaming mode (zero-copy)
+				ctx.return_type = .file
+				ctx.return_file = gz_path
+				ctx.res.header.set(.content_encoding, 'gzip')
+				ctx.res.header.set(.vary, 'Accept-Encoding')
+				ctx.send_response_to_client(content_type, '')
+
+				// Get .gz file size for Content-Length header
+				gz_size := os.file_size(gz_path)
+				ctx.res.header.set(.content_length, gz_size.str())
```
spytheman: Setting the header should be before the `ctx.send_response_to_client/2` call.

vlib/veb/context.v
```diff
@@ -200,22 +207,98 @@ fn (mut ctx Context) send_file(content_type string, file_path string) Result {
...
+				eprintln('[veb] error while trying to read file: ${err.msg()}')
+				return ctx.server_error('could not read resource')
+			}
+
+			compressed := gzip.compress(data.bytes()) or {
+				// Fallback: serve uncompressed in streaming mode
+				ctx.return_type = .file
+				ctx.return_file = file_path
+				ctx.send_response_to_client(content_type, '')
+				ctx.res.header.set(.content_length, file_size.str())
```
spytheman: setting the header should be first

spytheman:
@davlgd thank you 🙇🏻 .
> I'll squash when it will be ok for you.

There is no need to squash them manually, the repo is set to squash changes automatically before they are merged in master.

spytheman:
> Let me know if you see any better way to test read-only for windows (I don't have a Windows system here to check)

Skipping the readonly test on windows (as you did) is perfectly fine and acceptable - most people will not run web servers on that OS, and will probably not have a readonly filesystem either.

PR #25766, opened 2025-11-18
db.sqlite: update windows version

JalonSolov:
Windows sqlite package from 2024 appears to no longer be available.  Only the newer 2025 version.


spytheman:
We should update https://github.com/vlang/sqlite too 🤔 .

JalonSolov:
Umm... yeah.  The README in that repo mentions an even older 2022 version.

... but the scripts reference the 2024 version.

All updated in https://github.com/vlang/sqlite/pull/1

PR #25773, opened 2025-11-18
cgen: cast member variables to correct types in comptime $for

### LINKED ISSUE #25771: comptime: struct option field wrong **VALUE** after unwrap. Continuation of solved issue #25761
<details>
<summary>V version: V 0.4.12 ca7bca8, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 c1d1efb2091e989dbb09e38b308c82c2cade1aa8.ca7bca8
|OS                  |linux, Ubuntu 22.04.5 LTS
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i3-4150 CPU @ 3.50GHz
|Memory              |1.72GB/3.7GB
|                    |
|V executable        |/home/jorge/v/v
|V last modified time|2025-11-18 17:43:44
|                    |
|V home dir          |OK, value: /home/jorge/v
|VMODULES            |OK, value: /home/jorge/.vmodules
|Current working dir |OK, value: /home/jorge/bugs
|                    |
|Git version         |git version 2.34.1
|V git status        |weekly.2025.16-1086-gca7bca80-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
|gcc version         |gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
|clang version       |Ubuntu clang version 14.0.0-1ubuntu1.1
|emcc version        |emcc (Emscripten gcc/clang-like replacement + linker emulating GNU ld) 3.1.5 ()
|glibc version       |ldd (Ubuntu GLIBC 2.35-0ubuntu3.11) 2.35

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  251117a.v && 251117a`
```v
struct Options {
	a ?string
	b ?int
	c ?f64
}

fn unwrap_not_none_field_types[T](t T) []string {
	mut arr := []string{}
	$for f in T.fields {
		v := t.$(f.name)
		$if f is $option {
			if v != none {
				arr << '${typeof(v).name}:${f.name}=`${v}`'
			}
		}
	}
	return arr
}

fn main() {
	arr := unwrap_not_none_field_types(Options{
		a: 'x'
		b: 1
		c: 2.3
	})
	assert arr.join(' ') == 'string:a=`x` int:b=`1` f64:c=`2.3`'
}

```

**What did you see?**
```
251117a.v:26: FAIL: fn main.main: assert arr.join(' ') == 'string:a=`x` int:b=`1` f64:c=`2.3`'
   left value: arr.join(' ') = string:a=`x` int:b=`` f64:c=``
  right value: 'string:a=`x` int:b=`1` f64:c=`2.3`' = string:a=`x` int:b=`1` f64:c=`2.3`
V panic: Assertion failed...
v hash: ca7bca8
/tmp/v_1000/251117a.01KAC41S99V0F2R8TWQRF1VX8A.tmp.c:3449: at builtin___v_panic: Backtrace
/tmp/v_1000/251117a.01KAC41S99V0F2R8TWQRF1VX8A.tmp.c:5320: by main__main
/tmp/v_1000/251117a.01KAC41S99V0F2R8TWQRF1VX8A.tmp.c:5349: by main
```

correct value of unwrapped field (type now is correct thanks to commit ca7bca8)

**Details**

PR #25763 solved the type of the unwrapped field, though now the **value** seems should be "cleared" too since is like first field value (string) is reported for the rest of fields regardless their types. If first type is int instead, second field value (int) is ok but third (f64) reports "noise".

Yes 🤦, I should have tested also the value (apart type) to hit two birds with the same stone.

Thanks in advance! 

@dy-tea 

> [!NOTE]
>

dy-tea:
Fixes #25771.

Updated the test to include the value checking from the issue.
This affected `vlib/v/tests/comptime/comptime_var_assignment_test.v` and it seems more correct now? Let me know if it looks incorrect but I think the updated behavior is more correct.

### DIFF
--- a/vlib/v/gen/c/str_intp.v
+++ b/vlib/v/gen/c/str_intp.v
@@ -220,7 +220,8 @@ fn (mut g Gen) str_val(node ast.StringInterLiteral, i int, fmts []u8) {
 		g.write2('${dot}_object', ')')
 	} else if fmt == `s` || typ.has_flag(.variadic) {
 		mut exp_typ := typ
-		if expr is ast.Ident {
+		is_comptime_for_var := expr is ast.Ident && g.is_comptime_for_var(expr)
+		if !is_comptime_for_var && expr is ast.Ident {
 			if g.comptime.get_ct_type_var(expr) == .smartcast {
 				exp_typ = g.type_resolver.get_type(expr)
 			} else if expr.obj is ast.Var {
@@ -236,7 +237,17 @@ fn (mut g Gen) str_val(node ast.StringInterLiteral, i int, fmts []u8) {
 				}
 			}
 		}
-		g.gen_expr_to_string(expr, exp_typ)
+		if exp_typ.has_flag(.option) && expr is ast.Ident && g.is_comptime_for_var(expr) {
+			str_fn_name := g.get_str_fn(exp_typ.clear_flag(.option))
+			g.write('${str_fn_name}(*(${g.base_type(exp_typ)}*)(')
+			old_inside_opt_or_res := g.inside_opt_or_res
+			g.inside_opt_or_res = true
+			g.expr(expr)
+			g.inside_opt_or_res = old_inside_opt_or_res
+			g.write('.data))')
+		} else {
+			g.gen_expr_to_string(expr, exp_typ)
+		}
 	} else if typ.is_number() || typ.is_pointer() || fmt == `d` {
 		if typ.is_signed() && fmt in [`x`, `X`, `o`] {
 			// convert to unsigned first befors C's integer propagation strikes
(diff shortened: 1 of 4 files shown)


dy-tea:
Github is having issues with cloning repos right now, will rerun the failing CIs tomorrow.

vlib/v/gen/c/str_intp.v
```diff
@@ -220,7 +220,8 @@ fn (mut g Gen) str_val(node ast.StringInterLiteral, i int, fmts []u8) {
 		g.write2('${dot}_object', ')')
 	} else if fmt == `s` || typ.has_flag(.variadic) {
 		mut exp_typ := typ
-		if expr is ast.Ident {
+		is_comptime_for_var := expr is ast.Ident && g.is_comptime_for_var(expr)
+		if !is_comptime_for_var && expr is ast.Ident {
```
spytheman, line 224: why check `expr is ast.Ident` 2 times?

vlib/v/gen/c/str_intp.v
```diff
@@ -283,26 +294,26 @@ fn (mut g Gen) string_inter_literal(node ast.StringInterLiteral) {
...
-							type_to_use = obj.typ
+							field_typ = obj.typ
 						}
 					}
 				}
 			}
 		}
-		if g.comptime.is_comptime(expr) {
-			ctyp := g.type_resolver.get_type_or_default(expr, node_.expr_types[i])
+		if g.comptime.is_comptime(expr) || (g.comptime.inside_comptime_for && expr is ast.Ident) {
```
spytheman, line 315: the second expression in the `||`, is cheaper to evaluate, imho put it first

vlib/v/gen/c/str_intp.v
dy-tea: To cast expr to an `ast.Ident` so it can be passed to `g.is_comptime_var`.

vlib/v/gen/c/str_intp.v
dy-tea: Yeah probably better.

Issue #25778, opened 2025-11-19
[arm64] vlib/os/notify/notify_test.c.v test fail


kbkpbot:
### Describe the bug

`vlib/os/notify/notify_test.c.v` test fail

### Reproduction Steps

```
v test vlib/os/notify/notify_test.c.v
```

### Expected Behavior

pass test

### Current Behavior

```
$ v test vlib/os/notify/notify_test.c.v
---- Testing... ---------------------------------------------------------------------------------------------------------------
                 retrying 1/3 of vlib/os/notify/notify_test.c.v ; known flaky: true ...
                 retrying 2/3 of vlib/os/notify/notify_test.c.v ; known flaky: true ...
                 retrying 3/3 of vlib/os/notify/notify_test.c.v ; known flaky: true ...
>>> flaky failures so far:
>>>>>> -------------------------------------------------------------------------------------------------------------------------------
>>>>>>  retry: 0 ; max_retry: 3 ; r.exit_code: 1 ; trimmed_output.len: 1309
>>>>>> /home/mars/github/kbkpbot/v/vlib/os/notify/notify_test.c.v:181: ✗ fn check_read_event
>>>>>>    > assert events[0].fd == reader_fd
>>>>>>      Left value (len: 1): `0`
>>>>>>     Right value (len: 1): `4`
>>>>>>
>>>>>> /home/mars/github/kbkpbot/v/vlib/os/notify/notify_test.c.v:181: ✗ fn check_read_event
>>>>>>    > assert events[0].fd == reader_fd
>>>>>>      Left value (len: 1): `0`
>>>>>>     Right value (len: 1): `7`
>>>>>>
>>>>>> /home/mars/github/kbkpbot/v/vlib/os/notify/notify_test.c.v:181: ✗ fn check_read_event
>>>>>>    > assert events[0].fd == reader_fd
>>>>>>      Left value (len: 1): `0`
>>>>>>     Right value (len: 2): `10`
>>>>>>
>>>>>> /home/mars/github/kbkpbot/v/vlib/os/notify/notify_test.c.v:130: ✗ fn test_hangup
>>>>>>    > assert events[0].fd == reader
>>>>>>      Left value (len: 1): `0`
>>>>>>     Right value (len: 2): `13`
>>>>>>
>>>>>> /home/mars/github/kbkpbot/v/vlib/os/notify/notify_test.c.v:151: ✗ fn test_write
>>>>>>    > assert events[0].fd == writer
[...]

PR #25779, opened 2025-11-19
os.notify: fix struct epoll_event alignment

kbkpbot:
Original code force  `struct epoll_event` size = 12 under ARM64/Linux, this is not true, as it's size = 16.
Fix issue #25778

### DIFF
--- a/vlib/os/notify/epoll.h
+++ b/vlib/os/notify/epoll.h
@@ -1,6 +1,12 @@
 // NOTE: tcc does not support yet __attribute__ ((__packed__)) properly, 
 // so the __EPOLL_PACKED macro that /usr/include/bits/epoll.h uses does not work :-| .
 // However, it *does support* the older `#pragma pack(push, 1)`
+#if defined(__TINYC__) && defined(__V_amd64)
 #pragma pack(push, 1)
+#endif
+
 #include <sys/epoll.h>
+
+#if defined(__TINYC__) && defined(__V_amd64)
 #pragma pack(pop)
+#endif


spytheman:
@kbkpbot this essentially reverts 6fcab01 2022-09-29, which solved the following problem with tcc:
```
#0 05:56:24 ^ master ~/v>jed a.v
#0 05:56:46 ^ master ~/v>v -cc tcc run a.v
[a.v:3] sizeof(C.epoll_event): 12
#0 05:56:48 ^ master ~/v>v -cc clang-18 run a.v
[a.v:3] sizeof(C.epoll_event): 12
#0 05:56:54 ^ master ~/v>v -cc gcc-11 run a.v
[a.v:3] sizeof(C.epoll_event): 12
#0 05:56:59 ^ master ~/v>
#0 05:57:00 ^ master ~/v>
#0 05:57:00 ^ master ~/v>git switch -
Switched to branch 'fix-struct-epoll-event-align'
#0 05:57:05 ^ fix-struct-epoll-event-align ~/v>
#0 05:57:06 ^ fix-struct-epoll-event-align ~/v>
#0 05:57:06 ^ fix-struct-epoll-event-align ~/v>v -cc gcc-11 run a.v
[a.v:3] sizeof(C.epoll_event): 12
#0 05:57:10 ^ fix-struct-epoll-event-align ~/v>v -cc clang-18 run a.v
[a.v:3] sizeof(C.epoll_event): 12
#0 05:57:13 ^ fix-struct-epoll-event-align ~/v>v -cc tcc run a.v
[a.v:3] sizeof(C.epoll_event): 16
#0 05:57:16 ^ fix-struct-epoll-event-align ~/v>
#0 05:57:17 ^ fix-struct-epoll-event-align ~/v>
#0 05:57:17 ^ fix-struct-epoll-event-align ~/v>cat a.v
import os.notify as _

dump(sizeof(C.epoll_event))
#0 05:57:23 ^ fix-struct-epoll-event-align ~/v>
```

Note, how here, with tcc, the size of the same structure, is reported to be *16* with tcc, while it is 12 with everything else...

spytheman:
I would like a solution, where the reported size is still the same with all compilers, so imho keep the custom .h file, and add some conditional logic there, to only do the packing for tcc/non arm?

spytheman:
Can we just do something like this? :
```c
#if defined(__TINYC__) && defined(__V_amd64)
#pragma pack(push, 1)
#endif

#include <sys/epoll.h>

#if defined(__TINYC__) && defined(__V_amd64)
#pragma pack(pop)
#endif
```

spytheman:
> So it depend on the define of __EPOLL_PACKED :(

that is indeed unfortunate - we can not check that before including the header, and after that it is already too late, because the C compiler would have registered the struct definitions

spytheman:
Ideally `tcc` should be extended to support `__attribute__ ((__packed__))`, but that is out of scope of this PR.

Issue #25765, opened 2025-11-18
ASAN complains about the test comptime_map_it_test.v


tankf33der:
How to repeat:
```
cd v
v -cc clang -cflags -fsanitize=address -gc none test vlib/v/tests/comptime/comptime_map_it_test.v
```

```
---- Testing... ----------------------------------------------------------------------------------------------------------------------------
 FAIL    23.463 ms vlib/v/tests/comptime/comptime_map_it_test.v
=================================================================
==20133==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x5020000000d4 at pc 0x564b85c66f3f bp 0x7ffed3ee9990 sp 0x7ffed3ee9150
READ of size 16 at 0x5020000000d4 thread T0
    #0 0x564b85c66f3e in __asan_memcpy (/tmp/v_0/tsession_7f30b313f740_01KABPGX9VQ1FKDGF1NM9G91ZV/0_0/comptime_map_it_test+0xcff3e) (BuildId: ed8f24c6a37d435634a74aa9b2463bf07c790661)
    #1 0x564b85cd8523 in main__encode_T_main__Arr (/tmp/v_0/tsession_7f30b313f740_01KABPGX9VQ1FKDGF1NM9G91ZV/0_0/comptime_map_it_test+0x141523) (BuildId: ed8f24c6a37d435634a74aa9b2463bf07c790661)
    #2 0x564b85cd925c in main__test_main (/tmp/v_0/tsession_7f30b313f740_01KABPGX9VQ1FKDGF1NM9G91ZV/0_0/comptime_map_it_test+0x14225c) (BuildId: ed8f24c6a37d435634a74aa9b2463bf07c790661)
    #3 0x564b85cdab88 in main (/tmp/v_0/tsession_7f30b313f740_01KABPGX9VQ1FKDGF1NM9G91ZV/0_0/comptime_map_it_test+0x143b88) (BuildId: ed8f24c6a37d435634a74aa9b2463bf07c790661)
    #4 0x7fcb025f0ca7  (/lib/x86_64-linux-gnu/libc.so.6+0x29ca7) (BuildId: def5460e3cee00bfee25b429c97bcc4853e5b3a8)
    #5 0x7fcb025f0d64 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29d64) (BuildId: def5460e3cee00bfee25b429c97bcc4853e5b3a8)
[...]

Issue #25757, opened 2025-11-17
Fil-C complains the test map_generic_call_test.v


tankf33der:
```
~/v/v -cc clang-20 -d no_backtrace -gc none test v/vlib/v/tests/builtin_maps/map_generic_call_test.v 
---- Testing... ----------------------------------------------------------------------------------------------------------------------------
 FAIL     4.417 ms /root/v/vlib/v/tests/builtin_maps/map_generic_call_test.v
filc safety error: cannot read pointer with ptr >= upper.
    pointer: 0x7f97e8507be8,0x7f97e8507b70,0x7f97e8507b80
    expected 4 bytes.
semantic origin:
    <somewhere>: builtin__map_meta_less
check scheduled at:
    <somewhere>: builtin__map_meta_less
    <somewhere>: builtin__map_delete
    <somewhere>: main__Foo_T_string_int_f_T_string_int
    <somewhere>: main__test_main
    <somewhere>: main
    src/env/__libc_start_main.c:79:7: __libc_start_main
    <runtime>: start_program
[9513] filc panic: thwarted a futile attempt to violate memory safety.
Trace/breakpoint trap
--------------------------------------------------------------------------------------------------------------------------------------------
To reproduce just failure 1 run:    '/root/v/v' -cc clang-20 -d no_backtrace -gc none '/root/v/vlib/v/tests/builtin_maps/map_generic_call_test.v'
Summary for all V _test.v files: 1 failed, 1 total. Elapsed time: 2818 ms, on 1 job. Comptime: 2812 ms. Runtime: 4 ms.
```

the test can pass if and only if the lines are commented out:
```
...
// y.delete(p.k)
...
// assert y.clone().len == 0
...
```

all tests in `vlib/builtint/` and `vlib/v/tests/builtin_maps/` passed successfully, this is the only hit
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

tankf33der:
Found one more combination to pass Fil-C, you could comment just one line:
```
// y.reserve(6)
```

tankf33der:
Found one more interesting combination and see another type of error:
```
root@lambda:~# ~/v/v -cc clang-20 -d no_backtrace -gc none test v/vlib/v/tests/builtin_maps/map_generic_call_test.v 
[...]

PR #25780, opened 2025-11-19
ci: new job sanitize-address-clang-without-gc

tankf33der:
While testing a new patch for `builtint/int.v`, I discovered that `ASAN` doesn't catch out-of-bounds access to allocated memory that I deliberately reduced, and it dawned on me that `boehm-gc` intercepts `malloc` calls by default (of course) and `ASAN` cannot place its [guard(s)](https://en.wikipedia.org/wiki/Warrant_canary#Etymology) to detect errors.

For example, this [bug](https://github.com/vlang/v/issues/25765) was discovered after manually running through all the source files.

I'm adding a minimal version of this job to reduce the run time. I'm sure this is useful and it won't make things worse. From this moment on, the `job` becomes critically important and real triggers will be more frequent because `-gc none` is set.

### DIFF
--- a/vlib/v/tests/builtin_maps/map_generic_call_test.v
+++ b/vlib/v/tests/builtin_maps/map_generic_call_test.v
@@ -1,3 +1,4 @@
+// vtest build: !sanitize-address-clang-without-gc
 struct FooParams[K, V] {
 	k  K
 	v  V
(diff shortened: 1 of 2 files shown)


tankf33der:
```
To reproduce just failure 1 run:    '/home/runner/work/v/v/v' -cflags -fsanitize=address -silent '/home/runner/work/v/v/vlib/v/tests/builtin_maps/map_generic_call_test.v'
To reproduce just failure 2 run:    '/home/runner/work/v/v/v' -cflags -fsanitize=address -silent '/home/runner/work/v/v/vlib/v/tests/comptime/comptime_map_it_test.v'
VFLAGS was: "-cc clang -gc none -cflags -fno-omit-frame-pointer"

Error: failed 2 times.
```

Both issues already known - #25757, #25765

medvednikov:
Nice, thanks!

PR #25782, opened 2025-11-19
veb: support markdown content negociation

davlgd:
This PR adds `text/markdown` as a supported MIME Type and a `enable_markdown_negotiation` parameter to Veb Static Handler. If set to true, during a request with `Accept: text/markdown` header, the web server looks for a Markdown file at one of these locations:
- path.md
- path.html.md
- path/index.html.md

If available it serves the first detected. If not, it serves the original asked path. 

This automates and comply to [llms.txt](https://llmstxt.org/) which is widely used to serve markdown files to LLMS and other AI services more efficiently. More and more web server are implementing such a feature, it could be great to have this included natively in Veb. 

I've added tests and documentation about it.

### DIFF
--- a/vlib/veb/tests/static_handler_test.v
+++ b/vlib/veb/tests/static_handler_test.v
@@ -61,6 +61,9 @@ fn run_app_test() {
 
 	app.handle_static('testdata', true) or { panic(err) }
 
+	// Enable markdown content negotiation for testing
+	app.enable_markdown_negotiation = true
+
 	if _ := app.mount_static_folder_at('testdata', 'static') {
 		assert true == false, 'should throw invalid mount path error'
 	} else {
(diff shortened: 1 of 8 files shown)


jorgeluismireles:
In order the browser display the md someone needs to convert md to html tags (as repos do with README.md). Veb needs to call a corresponding markdown library too?

davlgd:
> In order the browser display the md someone needs to convert md to html tags (as repos do with README.md). Veb needs to call a corresponding markdown library too?

I'm not sure to understand the question, as there is no conversion stuff here:
- If there are only HTML files, only HTML files will be served
- If there are also corresponding Markdown files: 
  - If a user request a Markdown file directly, it's served as Markdown
  - If a user request an HTML file without specific header, it's served as HTML
  - If a user request an HTML file with `Accept: text/markdown`, it's served as Markdown

It's only for pre-rendered/static stuff (and static site generator such as Hugo start to generates HTML + Markdown files for `llms.txt` natively for example). When an AI service (for example) request a page, it often start to ask for the Markdown version if available through the Accept header. The goal here is to reply accordingly if a Markdown file is already there. As we do for `gzip`.

We could imagine that Veb could generate Markdown from HTML if a pre-rendered file is not available, but it can be part of another PR if needed.

jorgeluismireles:
Like [vlang/markdown](https://github.com/vlang/markdown)

JalonSolov:
If veb serves it to the browser, with a type of `text/markdown`, it is up to the browser how it gets handled.

spytheman:
> We could imagine that Veb could generate Markdown from HTML if a pre-rendered file is not available, but it can be part of another PR if needed.

imho it is not; the current PR is good

spytheman:
> In order the browser display the md someone needs to convert md to html tags (as repos do with README.md). Veb needs to call a corresponding markdown library too?

Browsers already do display .md files as text. There is no need to convert them to html at all. In fact that would be counterproductive to their use by LLMs (which prefer clear text as input).

vlib/veb/tests/static_handler_test.v
```diff
@@ -126,3 +129,110 @@ fn test_upper_case_mime_type() {
...
+	}
+	x := http.fetch(config)!
+
+	assert x.status() == .ok
+	assert x.header.get(.content_type)! == 'text/markdown'
+	assert x.body.contains('# Index HTML Markdown')
+}
+
+fn test_markdown_fallback_priority() {
+	// Verifies priority order: url.md, url.html.md, url/index.html.md
```
spytheman: should not it do requests to all 3 explicit urls too, to compare the results?

spytheman: i.e. have also static `about.html.md`, and `about/index.html.md` files, with different content, and check that the returned result is the content of `about.md` and not one of the other.

vlib/veb/tests/static_handler_test.v
```diff
@@ -126,3 +129,110 @@ fn test_upper_case_mime_type() {
...
+		url:    '${localserver}/page.html.md'
+		header: http.new_header(key: .accept, value: 'text/markdown')
+	}
+	x_md := http.fetch(config_md)!
+	assert x_md.status() == .ok
+	assert x_md.header.get(.content_type)! == 'text/markdown'
+	assert x_md.body.contains('# Page HTML Markdown')
+
+	// With Accept: text/html or no specific accept, should serve page.html
+	x_html := http.get('${localserver}/page.html')!
```
spytheman: the url here is different than on line 210 🤔

vlib/veb/README.md
```diff
@@ -527,6 +527,84 @@ curl -H "Accept-Encoding: gzip" -i http://localhost:8080/style.css
   to disable auto-compression completely. For optimal performance on read-only systems,
   pre-compress all files with `gzip -k`.
 
+### Markdown content negotiation
+
+veb provides automatic content negotiation for markdown files, allowing you to serve
```
spytheman:
```suggestion
veb can provide automatic content negotiation for markdown files, allowing you to serve
```

vlib/veb/tests/static_handler_test.v
davlgd: I let an old test name here. But regarding your other comments I've reorganized tests.

Issue #25788, opened 2025-11-20
Incomplete definition of C.SYSTEM_INFO


qtc-de:
### Describe the bug

Hi there 👋 

inside the file `vlib/builtin/cfns.c.v` there is an definition for `C.SYSTEM_INFO`, but it only contains two of the twelve struct members that are actually available in this struct.

```v
// C.SYSTEM_INFO contains information about the current computer system. This includes the architecture and type of the processor, the number of processors in the system, the page size, and other such information.
@[typedef]
pub struct C.SYSTEM_INFO {
	dwNumberOfProcessors u32
	dwPageSize           u32
}
```

Why this is a problem? My code for example needs to access the `wProcessorArchitecture` member of this struct. Since it is already defined, I'm not able to redefine it and tell v that this member exists. Therefore, I'm forced to create a new struct basically mirroring `C.SYSTEM_INFO`, although it is completely unnecessary.

### Reproduction Steps

Minimal example code:

```v
@[typedef]
pub struct C.SYSTEM_INFO {
	dwOemId                     u32
	wProcessorArchitecture      u16
	wReserved                   u16
	dwPageSize                  u32
	lpMinimumApplicationAddress voidptr
	lpMaximumApplicationAddress voidptr
	dwActiveProcessorMask       &u32 = unsafe { nil }
	dwNumberOfProcessors        u32
	dwProcessorType             u32
	dwAllocationGranularity     u32
	wProcessorLevel             u16
	wProcessorRevision          u16
}

fn main()
{
	system_info := C.SYSTEM_INFO{}
	C.GetSystemInfo(&system_info)

	println(system_info.wProcessorArchitecture)
}
```

### Expected Behavior

Should compile fine.

### Current Behavior

```
test.v:22:22: error: type `C.SYSTEM_INFO` has no field named `wProcessorArchitecture`.
2 possibilities: `dwNumberOfProcessors`, `dwPageSize`.
   20 |     C.GetSystemInfo(&system_info)
   21 |
   22 |     println(system_info.wProcessorArchitecture)
      |                         ~~~~~~~~~~~~~~~~~~~~~~
   23 | }
```

### Possible Solution

From my point of view there are two options:

1. Make sure that any C struct you declare within *vlib* is complete
2. Prioritize user definitions of C structs before *vlib* definitions

### V version

V 0.4.12 2a12104

### Environment details (OS name and version, etc.)

Not required
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
[...]

PR #25789, opened 2025-11-20
parser: fix C struct nested  anon struct language

kbkpbot:
Fix issue #25807
force anon struct/union in C.struct keep the same language of outside.

To fix issue #25788 in another PR due to bootstrap problem.

### DIFF
--- a/vlib/v/tests/c_structs/cstruct_nested_anon_test.c.v
+++ b/vlib/v/tests/c_structs/cstruct_nested_anon_test.c.v
@@ -1,0 +1,23 @@
+module main
+
+#insert "@VMODROOT/deep.h"
+
+struct C.DeepStruct {
+	A1 int
+	S1 struct {
+		A2 int
+		S2 struct {
+			A3 int
+			S3 struct {
+				A4 int
+			}
+		}
+	}
+}
+
+fn test_cstruct_nested_anon() {
+	x := C.DeepStruct{}
+	y := C.DeepStruct{}
+	dump(x)
+	assert x == y
+}
--- a/vlib/v/tests/c_structs/deep.h
+++ b/vlib/v/tests/c_structs/deep.h
@@ -1,0 +1,16 @@
+#ifndef __DEEP_H__
+#define __DEEP_H__
+
+struct DeepStruct{
+	int A1;
+	struct {
+		int A2;
+		struct {
+		       int A3;
+		       struct {
+				int A4;
+		       } S3;
+                } S2;
+        } S1;
+};
+#endif


vlib/builtin/cfns.c.v
```diff
@@ -542,8 +542,23 @@ fn C.sysconf(name int) int
 // C.SYSTEM_INFO contains information about the current computer system. This includes the architecture and type of the processor, the number of processors in the system, the page size, and other such information.
 @[typedef]
 pub struct C.SYSTEM_INFO {
-	dwNumberOfProcessors u32
-	dwPageSize           u32
+	DUMMYUNIONNAME              union {
```
StunxFS: This wasn't necessary. Simply defining the fields without using anonymous unions/structs is sufficient.

kbkpbot: When access the field, using `system_info.dwOemId` instead of `system_info.DUMMYUNIONNAME.dwOemId` ?

StunxFS: Sorry, my bad, I hadn't looked at the C definition of that struct properly. In this case, it is necessary.

vlib/v/gen/c/auto_str_methods.v
```diff
@@ -1188,7 +1188,11 @@ fn struct_auto_str_func(sym &ast.TypeSymbol, lang ast.Language, _field_type ast.
 		if sym.kind == .struct {
 			if sym.info is ast.Struct && sym.info.is_anon && !_field_type.has_flag(.option)
 				&& !_field_type.has_flag(.shared_f) {
-				typed_obj := '*(${sym.cname}*)&(${obj})'
+				typed_obj := if lang == .c {
+					'(${sym.cname}*)&(${obj})'
+				} else {
+					'*(${sym.cname}*)&(${obj})'
+				}
```
spytheman, line 1195: Why is this different for a C one?

kbkpbot: The V compiler handles C structs differently than V structs.

```v
// V auto str functions:
static string SYSTEM_INFO_2_str(SYSTEM_INFO_2* it) { return indent_SYSTEM_INFO_2_str(it, 0);}
static string C___VAnonStruct1_str(C___VAnonStruct1* it) { return indent_C___VAnonStruct1_str(it, 0);}
static string C___VAnonStruct2_str(C___VAnonStruct2* it) { return indent_C___VAnonStruct2_str(it, 0);}
```

```v
// V auto str functions:
static string main__DeepStruct_str(main__DeepStruct it) { return indent_main__DeepStruct_str(it, 0);}
static string main___VAnonStruct1_str(main___VAnonStruct1 it) { return indent_main___VAnonStruct1_str(it, 0);}
static string main___VAnonStruct2_str(main___VAnonStruct2 it) { return indent_main___VAnonStruct2_str(it, 0);}
```

As shown above, when the compiler encounters a C struct, it accesses the value via a pointer. In contrast, for a V struct, it passes the struct by value (making a copy).

To ensure auto_str works correctly with a V anonymous struct, a corresponding `C___VAnonStruct1` instance must be created for it.

kbkpbot: We can't direct access a C anon struct

spytheman: It may be better in the long term, to pass the V struct for the generated auto str methods by a reference too, to keep things consistent 🤔.

spytheman (approved):
Excellent work.
Thanks @kbkpbot .

PR #25793, opened 2025-11-20
builtin,cgen: ensure array of string is not cloned with depth 0

### LINKED ISSUE #25783: Assiging os.args to a variable with autofree enabled causes a Double Free.
### Describe the bug

If you try to assign a variable to `os.args` with the `-autofree `compilation flag, a Double Free occurs, this seemingly happens at the end of a program.

### Reproduction Steps

```v
import os

fn main() {
	args := os.args
	println(args) // this line is not technically needed
}
```
Compile and run this sample with `v -autofree run .` to cause the crash.

### Expected Behavior

The program should run and print the program arguments, as it does when the `-autofree` flag is not provided.

### Current Behavior

Output:

```
['./osargs-bug']
free(): double free detected in tcache 2
0x7fcf78c9890c: at ???: RUNTIME ERROR: abort() called
0x7fcf78c3e3a0: by ???
0x7fcf78c2557a: by ???
0x7fcf78c26613: by ???
0x7fcf78ca2d25: by ???
0x7fcf78ca82a8: by ???
/tmp/v_1000/osargs-bug.01KAF0Y1TG4BC4B2WP9WA3NBEK.tmp.c:2392: by builtin___v_free
/tmp/v_1000/osargs-bug.01KAF0Y1TG4BC4B2WP9WA3NBEK.tmp.c:3576: by builtin__string_free
/tmp/v_1000/osargs-bug.01KAF0Y1TG4BC4B2WP9WA3NBEK.tmp.c:1858: by Array_string_free
/tmp/v_1000/osargs-bug.01KAF0Y1TG4BC4B2WP9WA3NBEK.tmp.c:3714: by _vcleanup
/tmp/v_1000/osargs-bug.01KAF0Y1TG4BC4B2WP9WA3NBEK.tmp.c:3738: by main
```

### Possible Solution

_No response_

I managed to reproduce the bug on my machine running ArchLinux after running `v up`.

### V version

V 0.4.12 2a121047627b93badb74be8332ca2ea63a0bdb5d

```
|V full version      |V 0.4.12 2a121047627b93badb74be8332ca2ea63a0bdb5d
|Memory              |0.85GB/2.02GB
|                    |
|V last modified time|2025-11-19 20:25:33
|                    |
|                    |
|V git status        |N/A
|.git/config present |true
|                    |

|emcc version        |N/A
```
> [!NOTE]
>

dy-tea:
Fixes #25783.

`os.args` would clone with depth 0 when assigning it, which would cause any variable set to it to free a pointer to it (double free). This solution just uses depth 1 instead which means it can free both pointers normally. I looked at having a `@[manualfree]` for constants but that doesn't work with more than one assignment of `os.args`.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -8532,6 +8532,8 @@ pub fn (mut g Gen) get_array_depth(el_typ ast.Type) int {
 	if sym.kind == .array {
 		info := sym.info as ast.Array
 		return 1 + g.get_array_depth(info.elem_type)
+	} else if sym.kind in [.string, .map] {
+		return 1
 	} else {
 		return 0
 	}
--- a/vlib/builtin/array.v
+++ b/vlib/builtin/array.v
@@ -724,6 +724,13 @@ pub fn (a &array) clone_to_depth(depth int) array {
 			unsafe { arr.set_unsafe(i, &ar_clone) }
 		}
 		return arr
+	} else if depth > 0 && a.element_size == sizeof(string) && a.len >= 0 && a.cap >= a.len {
+		for i in 0 .. a.len {
+			str_ptr := unsafe { &string(a.get_unsafe(i)) }
+			str_clone := (*str_ptr).clone()
+			unsafe { arr.set_unsafe(i, &str_clone) }
+		}
+		return arr
 	} else {
 		if a.data != 0 && source_capacity_in_bytes > 0 {
 			unsafe { vmemcpy(&u8(arr.data), a.data, source_capacity_in_bytes) }
--- a/vlib/os/os_args_autofree_test.v
+++ b/vlib/os/os_args_autofree_test.v
@@ -1,0 +1,14 @@
+// vtest build: !sanitize-memory-gcc && !sanitize-address-gcc && !sanitize-address-clang
+// vtest vflags: -autofree
+import os
+
+fn test_os_args_no_double_free() {
+	args := os.args
+	assert args.len > 0
+}
+
+fn test_os_args_clone() {
+	a1 := os.args
+	a2 := os.args
+	assert a1 == a2
+}


spytheman:
Would not `vlib/os/os.c.v:17:pub const args = arguments().clone()` be simpler?

dy-tea:
Maybe it would be. Are there any cases where you think this would be beneficial?


Something like this:
```v
// test/test.v
module test

pub fn get_args() []string {
	return ['hello', 'world']
}

pub const args = get_args()
```
```v
// main.v
module main

import test

fn main() {
	a := test.args
	b := test.args
	assert a == b
}
```
~~If a user makes something like this there will still be an unexpected double free which does not happen without `-autofree`.
Maybe alternatively the `const` values could implicitly clone?~~

Edit: never mind seems to not happen for some reason.

dy-tea:
~~I think adding `.clone()` is fine as a solution.~~
Edit: Seems to not work though? Still have a double free.

vlib/v/gen/c/cgen.v
```diff
@@ -3537,7 +3537,12 @@ fn (mut g Gen) gen_clone_assignment(var_type ast.Type, val ast.Expr, typ ast.Typ
 				g.write('->val')
 			}
 			elem_type := (right_sym.info as ast.Array).elem_type
-			array_depth := g.get_array_depth(elem_type)
+			elem_sym := g.table.final_sym(elem_type)
+			array_depth := if elem_sym.kind in [.string, .map, .array] {
+				1 + g.get_array_depth(elem_type)
+			} else {
+				g.get_array_depth(elem_type)
+			}
```
spytheman: why correct this here, instead of in `get_array_depth` ?

dy-tea: Couldn't get it to work inside `get_array_depth`, I'm trying it right now.

dy-tea:
Should hopefully be cleaner now. Was busy for a while and couldn't push it.

spytheman (approved):
Thanks @dy-tea 🙇🏻 .
Excellent work.

PR #25795, opened 2025-11-20
docs: Changed abc/def into abc/xyz in the module hierarchy section

Zeviraty:
This creates clarity because `def` could be read as define, and could be interpreted as a special subfolder instead of just abcdef

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -1606,19 +1606,19 @@ Modules names in .v files, must match the name of their directory.
 A .v file `./abc/source.v` must start with `module abc`. All .v files in this directory
 belong to the same module `abc`. They should also start with `module abc`.
 
-If you have `abc/def/`, and .v files in both folders, you can `import abc`, but you will have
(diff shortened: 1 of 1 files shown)


spytheman:
There is nothing wrong with `def` as a name, just like there is nothing wrong with `xyz`.
If this is merged, however, tomorrow, someone can dislike `xyz` too, because it is a common abbreviation for 3D axis and change it to `zyx` etc, which does not improve things at all :-| .

Zeviraty:
maybe but def is used commonly in programming languages (and also as a special folder)
xyz on the other hand is not commonly used in programming languages and also not as a folder

JalonSolov:
The simplest way to "fix" this, then, would be to change it completely.  Instead of using `abc/def`, use `dir/subdir` or `parent/child` or `dir1/dir2` or ... whatever.

spytheman:
Those are good points, more meaningful longer names will be better indeed.

PR #25808, opened 2025-11-22
builtin: add `v_getpid()` and `v_gettid()`

spytheman:
- **markused: cleanup no longer used vgettid name**
- **builtin: add declarations for C.GetCurrentThreadId(), C.gettid() and C.pthread_self()**
- **builtin: implement cross platform v_getpid() and v_gettid(); add tests**

### DIFF
--- a/vlib/builtin/builtin.c.v
+++ b/vlib/builtin/builtin.c.v
@@ -971,16 +971,8 @@ pub fn v_getpid() u32 {
 pub fn v_gettid() u32 {
 	$if windows {
 		return u32(C.GetCurrentThreadId())
-	} $else $if linux {
-		$if musl ? {
-			$if threads {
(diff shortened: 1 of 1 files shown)


Delta456:
Why do we need these functions inside `builtin`?

JalonSolov:
Fairly basic routines... where else would they go?

spytheman:
> Why do we need these functions inside `builtin`?

Because I want to use them ultimately inside the panic reports, and they are generated in `builtin`, and I do not want to import either `os` or `sync` for normal programs that do not need everything in them.

spytheman:
Obsoleted by https://github.com/vlang/v/pull/25809 (which has more tests running, due to markused actually noticing that the new fns are called/used)

PR #25809, opened 2025-11-22
builtin: add pid and tid fields to the panic reports (part 2 of #25808)

spytheman:
- **markused: cleanup no longer used vgettid name**
- **builtin: add declarations for C.GetCurrentThreadId(), C.gettid() and C.pthread_self()**
- **builtin: implement cross platform v_getpid() and v_gettid(); add tests**
- **fix `-b native` failure**
- **fix musl failures; use pthread_self() as a fallback for it too**
- **simplify comptime logic in v_gettid**
- **builtin: add tid and pid fields to the panic reports**

### DIFF
--- a/vlib/builtin/builtin.c.v
+++ b/vlib/builtin/builtin.c.v
@@ -72,7 +72,7 @@ fn panic_debug(line_no int, file string, mod string, fn_name string, s string) {
 		eprint('     file: '); eprint(file); eprint(':');
 	    C.fprintf(C.stderr, c'%d\n', line_no)
 		eprint('   v hash: '); eprintln(vcurrent_hash())
-		$if !vinix {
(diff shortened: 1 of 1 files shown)


squidink7:
This broke a program I'm making that targets WASM (using the C backend + wasi-sdk not V's WASM backend) which errors with `undefined symbol: getpid`. Perhaps there should be another check for platforms that don't support getting the process ID?

spytheman:
How about a `-d panic_no_pid_and_tid` flag?

I do not want to cripple the default panic report, but I also do not want to make it harder to people that use V in more limited contexts.

spytheman:
Latest V 64f1162 has now `-d no_getpid` and `-d no_gettid` support.

Issue #25811, opened 2025-11-22
project() from math.vec seem to be incorrect


Linklancien:
### Describe the bug

The ``project()`` function from  math.vec for Vec2 - Vec3 - Vec4 seems to behave strangely

### Reproduction Steps

percent := v.dot(u) / u.dot(v) always equal 1 and divide by 0 when u or v equal 0

For:
fn main(){
	u := vec.Vec2{
		x: 1
	}
	v := vec.Vec2{
		x: 2
		y: 1
	}
	vu := v.project(u)
	uv := u.project(v)
	println(vu)
	println(uv)
}

### Expected Behavior

vu should be :{
   x: 2*a
   y: a
}
uv should be :{
   x: 2
   y: 0
}
Or something like that

### Current Behavior

vu{
    x: 1
    y: 0
}
uv{
    x: 2
    y: 1
}

### Possible Solution

```
pub fn (v Vec2[T]) project(u Vec2[T]) Vec2[T] {
    denom := u.dot(u)
    // if u is zero, this avoids division by zero
    if denom == 0 {
        return Vec2[T]{0, 0}
    }
    percent := v.dot(u) / denom
    return u.mul_scalar(percent)
}```
Depending on the vector you want to project


### Additional Information/Context

Currently working on a fix.

### V version

V 0.4.12 defae4d

### Environment details (OS name and version, etc.)

linux, Linux version 6.11.7-300.fc41.x86_64 (mockbuild@6c49ea81e72a4908b02fd96c22f92edf) (gcc (GCC) 14.2.1 20240912 (Red Hat 14.2.1-3), GNU ld version 2.43.1-2.fc41)

tankf33der:
I will try to implement test suite for this module.

Linklancien:
> I will try to implement test suite for this module.
Thanks, there is a patch incomming https://github.com/vlang/v/pull/25813

Linklancien:
I don't know how to specify in ``Development`` that this PR is linked to this issue

PR #25813, opened 2025-11-22
math: fix vec2,3,4 project not using the right formulas 

gw-dev101:
addresses #25811 by replacing some of the formulas with the correct ones 
(the old one only worked when both vectors had the same magnitude)

i think there is still some more work to be done on the module before it is satifactory

### DIFF
--- a/vlib/math/vec/vec2_test.v
+++ b/vlib/math/vec/vec2_test.v
@@ -232,24 +232,24 @@ fn test_vec2_rotate_around_ccw_2() {
 // Test for Vec2 projection
 //
 fn test_vec2_project_onto_basic() {
-	v1 := vec.vec2(3.0, 4.0) // magnitude 5 vector
-	v2 := vec.vec2(5.0, 6.0) // magnitude ~7.81 vector
+	u := vec.vec2(3.0, 4.0) // magnitude 5 vector
+	v := vec.vec2(5.0, 6.0) // magnitude ~7.81 vector
 	// hand-computed:
 	// u·v = 5*3 + 6*4 = 39
 	// |v|^2 = 3^2 + 4^2 = 25
 	// scale = 39/25 = 1.56
 	// proj = scale * v = (1.56*3, 1.56*4) = (4.68, 6.24)
-	proj := v1.project(v2)
+	proj := u.project(v)
 	assert tolerance(proj.x, 4.68, vec.vec_epsilon)
 	assert tolerance(proj.y, 6.24, vec.vec_epsilon)
 }
 
 // Test for Vec2 projection onto zero vector
-// project v2 into the null vector
+// project v into the null vector
 fn test_vec2_project_onto_zero() {
-	v1 := vec.vec2(0.0, 0.0)
-	v2 := vec.vec2(5.0, 6.0)
-	proj := v1.project(v2)
+	u := vec.vec2(0.0, 0.0)
+	v := vec.vec2(5.0, 6.0)
+	proj := u.project(v)
 	// must be nan
 	assert is_nan(proj.x)
 	assert is_nan(proj.y)
@@ -258,9 +258,9 @@ fn test_vec2_project_onto_zero() {
 // Test for Vec2 projection of zero vector
 // project a null vector
 fn test_vec2_project_zero_vector() {
-	v1 := vec.vec2(3.0, 4.0)
-	v2 := vec.vec2(0.0, 0.0)
-	proj := v1.project(v2)
+	u := vec.vec2(3.0, 4.0)
+	v := vec.vec2(0.0, 0.0)
+	proj := u.project(v)
 	assert proj.x == 0.0
 	assert proj.y == 0.0
 }
@@ -268,18 +268,18 @@ fn test_vec2_project_zero_vector() {
 // Test for Vec2 projection onto itself
 //
 fn test_vec2_project_onto_self() {
-	v1 := vec.vec2(3.0, 4.0)
-	proj := v1.project(v1)
-	assert veryclose(proj.x, v1.x)
-	assert veryclose(proj.y, v1.y)
+	u := vec.vec2(3.0, 4.0)
+	proj := u.project(u)
+	assert veryclose(proj.x, u.x)
+	assert veryclose(proj.y, u.y)
 }
 
 // Test for Vec2 projection onto orthogonal vector
 //
 fn test_vec2_project_onto_orthogonal() {
-	v1 := vec.vec2(1.0, 0.0)
-	v2 := vec.vec2(0.0, 1.0)
-	proj := v1.project(v2)
+	u := vec.vec2(1.0, 0.0)
+	v := vec.vec2(0.0, 1.0)
+	proj := u.project(v)
 	// more sensitive to floating point errors so i think close is better here
 	assert close(proj.x, 0.0)
 	assert close(proj.y, 0.0)
@@ -288,14 +288,14 @@ fn test_vec2_project_onto_orthogonal() {
 // Test for Vec2 projection with negative components
 //
 fn test_vec2_project_negative_components() {
-	v1 := vec.vec2(-3.0, 4.0)
-	v2 := vec.vec2(5.0, -6.0)
+	u := vec.vec2(-3.0, 4.0)
+	v := vec.vec2(5.0, -6.0)
 	// hand-computed:
 	// u·v = 5*-3 + -6*4 = -15 - 24 = -39
 	// |v|^2 = -3^2 + 4^2 = 9 + 16 = 25
 	// scale = -39/25 = -1.56
 	// proj = scale * v = (-1.56*-3, -1.56*4) = (4.68, -6.24)
-	proj := v1.project(v2)
+	proj := u.project(v)
 	assert tolerance(proj.x, 4.68, vec.vec_epsilon)
 	assert tolerance(proj.y, -6.24, vec.vec_epsilon)
 }
--- a/vlib/math/vec/vec3_test.v
+++ b/vlib/math/vec/vec3_test.v
@@ -92,14 +92,14 @@ fn test_vec3_f64_utils_2() {
 
 // sample tests for vec3 projection
 fn test_vec3_project_onto_basic() {
-	v1 := vec.vec3(3.0, 4.0, 0.0) // magnitude 5 vector
-	v2 := vec.vec3(5.0, 6.0, 0.0) // magnitude ~7.81 vector
+	u := vec.vec3(3.0, 4.0, 0.0) // magnitude 5 vector
+	v := vec.vec3(5.0, 6.0, 0.0) // magnitude ~7.81 vector
 	// hand-computed:
 	// u·v = 5*3 + 6*4 + 0*0 = 39
 	// |v|^2 = 3^2 + 4^2 +0^2 = 25
 	// scale = 39/25 = 1.56
 	// proj = scale * v = (1.56*3, 1.56*4, 1.56*0) = (4.68, 6.24, 0)
-	proj := v1.project(v2)
+	proj := u.project(v)
 	assert veryclose(proj.x, 4.68)
 	assert veryclose(proj.y, 6.24)
 	assert veryclose(proj.z, 0.0)
@@ -108,9 +108,9 @@ fn test_vec3_project_onto_basic() {
 // Test for Vec3 projection onto zero vector
 //
 fn test_vec3_project_onto_zero() {
-	v1 := vec.vec3(3.0, 4.0, 0.0)
-	v2 := vec.vec3(0.0, 0.0, 0.0)
-	proj := v1.project(v2)
+	u := vec.vec3(3.0, 4.0, 0.0)
+	v := vec.vec3(0.0, 0.0, 0.0)
+	proj := u.project(v)
 	assert proj.x == 0.0
 	assert proj.y == 0.0
 	assert proj.z == 0.0
@@ -119,14 +119,14 @@ fn test_vec3_project_onto_zero() {
 // Test for vec3 projection at an angle
 //
 fn test_vec3_project_onto_angle() {
-	v1 := vec.vec3(1.0, 0.0, 0.0) // magnitude 1 vector
-	v2 := vec.vec3(1.0, 1.0, 0.0) // magnitude sqrt(2) vector
+	u := vec.vec3(1.0, 0.0, 0.0) // magnitude 1 vector
+	v := vec.vec3(1.0, 1.0, 0.0) // magnitude sqrt(2) vector
 	// hand-computed:
 	// u·v = 1*1 + 0*1 + 0*0 = 1
 	// |v|^2 = 1^2 + 0^2 +0^2 = 1
 	// scale = 1/1 = 1
 	// proj = scale * v = (1*1, 1*0, 1*0) = (1, 0, 0)
-	proj := v1.project(v2)
+	proj := u.project(v)
 	assert veryclose(proj.x, 1.0)
 	assert veryclose(proj.y, 0.0)
 	assert veryclose(proj.z, 0.0)
--- a/vlib/math/vec/vec4_test.v
+++ b/vlib/math/vec/vec4_test.v
@@ -100,12 +100,12 @@ fn test_vec4_f64_utils_2() {
 
 // sample tests for vec4 projection
 fn test_vec4_project_onto_basic() {
-	v1 := vec.vec4(3.0, 4.0, 0.0, 0.0) // magnitude 5 vector
-	v2 := vec.vec4(5.0, 6.0, 0.0, 0.0) // magnitude ~7.81 vector
+	u := vec.vec4(3.0, 4.0, 0.0, 0.0) // magnitude 5 vector
+	v := vec.vec4(5.0, 6.0, 0.0, 0.0) // magnitude ~7.81 vector
 	// hand-computed:
 	// u·v = 5*3 + 6*4 + 0*0 + 0*0 = 39
 	// |v|^2 = 3^2 + 4^2 +0^2 +0^2 = 25
-	proj := v1.project(v2)
+	proj := u.project(v)
 	assert proj.x == 3.0
 	assert proj.y == 4.0
 	assert proj.z == 0.0
@@ -115,9 +115,9 @@ fn test_vec4_project_onto_basic() {
 // Test for Vec4 projection onto zero vector
 //
 fn test_vec4_project_onto_zero() {
-	v1 := vec.vec4(3.0, 4.0, 0.0, 0.0)
-	v2 := vec.vec4(0.0, 0.0, 0.0, 0.0)
-	proj := v1.project(v2)
+	u := vec.vec4(3.0, 4.0, 0.0, 0.0)
+	v := vec.vec4(0.0, 0.0, 0.0, 0.0)
+	proj := u.project(v)
 	assert proj.x == 0.0
 	assert proj.y == 0.0
 	assert proj.z == 0.0


spytheman (changes requested):
Please add tests for the changed .project methods, to prevent future regressions.

spytheman:
You have to also run `v fmt -w vlib/math` (or mention the specific changed files), in order to make sure that everything is kept formatted properly.
Then commit the result and push.

tankf33der:
Lets wait for my test(s).

tankf33der:
My research.

@spytheman Yes, the current implementation is incorrect. I've checked everything and I'm not approving the merge. I have several sources to prove that his implementation is not correct.

@gw-dev101 Thanks for working on this. I appreciate it. You got confused in the details and parameters, fix the code, comments, tests and come back with new commit(s).

gw-dev101:
That's the main issue with trying to make a pr at 2am 
I was awake enough to notice the old one was definitely broken 
But not enough to make a clean and working fix 
Hopefully my second pr will go better 
My bad and sorry for wasting your time

JalonSolov:
Not a waste of time - it brought an issue up that not many were even thinking about.

You also don't need to make a second PR - just push modifications to this one.

Linklancien:
Is anyone working on this issue currently ?

I can try to fix some check the weren't successful if nobody is on it already

Linklancien:
I do not understand why, but it seems to be a issue
May be because of the generics ?

```v
import math.vec

fn main() {
	v1 := vec.vec2(3.0, 4.0) // magnitude 5 vector
	v2 := vec.vec2(5.0, 6.0) // magnitude ~7.81 vector
	proj := v1.project(v2)
	solution := vec.vec2(4.68, 6.24)
	denom := v1.dot(v1)
	
	scale := v2.dot(v1) / denom
	proj2 := vec.Vec2[f64]{v1.x * scale, v1.y * scale}
	println('x: $proj.x $proj2.x $solution.x')
	println('y: $proj.y $proj2.y $solution.y')
	println('Vec: $proj $proj2 $solution')
}
```

Returns 

```v 
x: proj: 5.0 proj2: 4.68 solution: 4.68
y: proj: 6.0 proj2: 6.24 solution: 6.24
Vec: proj: vec.Vec2[f64]{
    x: 5.0
    y: 6.0
} proj2: vec.Vec2[f64]{
    x: 4.68
    y: 6.24
} solution: vec.Vec2[f64]{
    x: 4.68
    y: 6.24
}
```

when using 


```v
pub fn (v Vec2[T]) project(u Vec2[T]) Vec2[T] {
	denom := T(v.dot(v))
	if denom <= vec_epsilon {
		return Vec2[T]{T(0.0), T(0.0)}
	}
	scale := u.dot(v) / denom
	return Vec2[T]{v.x * scale, v.y * scale}
}
```
While it should have been the same result with the function and with the equivalent code used in the main

Linklancien:
The generic may not be involved, by defining a function directly in the file:
```v
fn project_intern[T](v vec.Vec2[T], u vec.Vec2[T]) vec.Vec2[T] {
	denom := v.dot(v)
	if denom <= vec.vec_epsilon {
		return vec.Vec2[T]{T(0.0), T(0.0)}
	}
	scale := u.dot(v) / denom
	return vec.Vec2[T]{v.x * scale, v.y * scale}
}
```

It works as entended:

```v
proj3 := project_intern(v1, v2)
```
returns 
```v
proj3: vec.Vec2[f64]{
    x: 4.68
    y: 6.24
}
```

tankf33der:
@Linklancien 

@gw-dev101 pushed wrong commits and we are waiting for him.

Linklancien:
> The generic may not be involved, by defining a function directly in the file:
> ```v
> fn project_intern[T](v vec.Vec2[T], u vec.Vec2[T]) vec.Vec2[T] {
> 	denom := v.dot(v)
> 	if denom <= vec.vec_epsilon {
> 		return vec.Vec2[T]{T(0.0), T(0.0)}
> 	}
> 	scale := u.dot(v) / denom
> 	return vec.Vec2[T]{v.x * scale, v.y * scale}
> }
> ```
> 
> It works as entended:
> 
> ```v
> proj3 := project_intern(v1, v2)
> ```
> returns 
> ```v
> proj3: vec.Vec2[f64]{
>     x: 4.68
>     y: 6.24
> }
> ```

It isn't an issue, I was juste using the wrong compilation of V, so it didn't uses the version of math.vec I was working on

tankf33der:
There is a fix for `vec3` only. Rest must have the same.
This patch passed `1M` loop against `cglm` library + zero vectors too.

This is a [candidate](https://github.com/recp/cglm/blob/master/include/cglm/vec3.h#L839) for comparison.
```diff
commit 28a0a03663fa359da954816d0affbaa3fc0163b8 (HEAD -> vec_project)
Author: Mike <tankf33der@disroot.org>
Date:   Tue Nov 25 20:39:48 2025 +0200

    math.vec: fix

diff --git a/vlib/math/vec/vec3.v b/vlib/math/vec/vec3.v
index 6f82d231e..5e72616f1 100644
--- a/vlib/math/vec/vec3.v
+++ b/vlib/math/vec/vec3.v
@@ -261,7 +261,7 @@ pub fn (v Vec3[T]) perpendicular(u Vec3[T]) Vec3[T] {
 
 // project returns the projected vector.
 pub fn (v Vec3[T]) project(u Vec3[T]) Vec3[T] {
-       percent := v.dot(u) / u.dot(v)
+       percent := v.dot(u) / u.dot(u)
        return u.mul_scalar(percent)
 }
```

Linklancien:
> ``` 
>  // project returns the projected vector.
>  pub fn (v Vec3[T]) project(u Vec3[T]) Vec3[T] {
> -       percent := v.dot(u) / u.dot(v)
> +       percent := v.dot(u) / u.dot(u)
>         return u.mul_scalar(percent)
>  }
> ```
This doesn't work for u : (0; 0; 0)
Does it ?

If you project any vector on the null vector, it must return a null vector or a nan vector ?

tankf33der:
> > ```
> >  // project returns the projected vector.
> >  pub fn (v Vec3[T]) project(u Vec3[T]) Vec3[T] {
> > -       percent := v.dot(u) / u.dot(v)
> > +       percent := v.dot(u) / u.dot(u)
> >         return u.mul_scalar(percent)
> >  }
> > ```
> 
> This doesn't work for u : (0; 0; 0) Does it ?
> 
> If you project any vector on the null vector, it must return a null vector or a nan vector ?

i see `nan` for my patch and `cglm`.

Linklancien:
@tankf33der 

Si it is intended that
(1;0;0) Project on (0;0;0) returns 
(nan;nan;nan)
And not
(0;0;0)

It feel strange because, if you project into a vector that doesn't existe, the line passing by this vectors also doesn't exist
And so it must return a (0;0;0) vector

I am sorry if I am not clear enought

gw-dev101:
that was my assumption too

tankf33der:
> that was my assumption too

@Linklancien @gw-dev101 - `cglm` returns `NaN` and I'm convinced that we should follow this machinery. Do you have other libraries that we can compare with and check what they return?

Linklancien:
> > that was my assumption too
> 
> @Linklancien @gw-dev101 - `cglm` returns `NaN` and I'm convinced that we should follow this machinery. Do you have other libraries that we can compare with and check what they return?

Makes sense, I will change accordigly

tankf33der:
@Linklancien @gw-dev101 python's `numpy` also returns `NaN`.

gw-dev101:
unreal engine one does return the zero vector 
i didn't even manage to find the numpy one

tankf33der:
> unreal engine one does return the zero vector i didn't even manage to find the numpy one

```python
import numpy as np

v = np.array([1.0, 0.0, 0.0])
u = np.array([0.0, 0.0, 0.0])

projection = u * (np.dot(v, u) / np.dot(u, u))
print(projection)
```
which generates a warning and `NaN`.
Is it ok? :/

Linklancien:
> > unreal engine one does return the zero vector i didn't even manage to find the numpy one
> 
> ```python
> import numpy as np
> 
> v = np.array([1.0, 0.0, 0.0])
> u = np.array([0.0, 0.0, 0.0])
> 
> projection = u * (np.dot(v, u) / np.dot(u, u))
> print(projection)
> ```
> 
> which generates a warning and `NaN`. Is it ok? :/

I am ok with returning a nan
I just do not know how to declare that a nan is expected for the test
```v
fn test_vec2_project_zero_vector() {
	v1 := vec.vec2(0.0, 0.0)
	v2 := vec.vec2(5.0, 6.0)
	proj := v1.project(v2)
	println(proj)
	assert proj.x == nan
	assert proj.y == nan
}
```
When using this code it doesn't work, do you know a way ?

tankf33der:
> > > unreal engine one does return the zero vector i didn't even manage to find the numpy one
> > 
> > 
> > ```python
> > import numpy as np
> > 
> > v = np.array([1.0, 0.0, 0.0])
> > u = np.array([0.0, 0.0, 0.0])
> > 
> > projection = u * (np.dot(v, u) / np.dot(u, u))
> > print(projection)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which generates a warning and `NaN`. Is it ok? :/
> 
> I am ok with returning a nan I just do not know how to declare that a nan is expected for the test
> 
> ```v
> fn test_vec2_project_zero_vector() {
> 	v1 := vec.vec2(0.0, 0.0)
> 	v2 := vec.vec2(5.0, 6.0)
> 	proj := v1.project(v2)
> 	println(proj)
> 	assert proj.x == nan
> 	assert proj.y == nan
> }
> ```
> 
> When using this code it doesn't work, do you know a way ?

i am using math.is_nan()

gw-dev101:
NaNs are weird in that by definition they are
not equal to themselves or anything else
```v
fn test_vec2_project_zero_vector() {
	v1 := vec.vec2(0.0, 0.0)
	v2 := vec.vec2(5.0, 6.0)
	proj := v1.project(v2)
	println(proj)
	assert proj.x != proj.x
	assert proj.y != proj.y
}
```
this should work

Linklancien:
> this should work

This indeed work, maybe math.is_nan() is more explicit ?

gw-dev101:
a quick check tell me that math.isnan() is doing exactly that so yeah
`could it be worth making isnan() inline somewhere in the future ?`

JalonSolov:
Remember:  Marking something `inline` does not guarantee that it will be inlined - the C compiler is free to ignore that directive.

gw-dev101:
i sure hope marking something inline never make it `not inline` if the compiler was going to inline it to begin with

JalonSolov:
No, the compiler can completely ignore it.  It will automatically inline things that are worthwhile to inline, and refuse to inline things that aren't.

Adding that marker just tells the compiler you would _like_ for it to be inlined, but as I said, it can ignore it.

medvednikov:
We can probably fix it on the V side by inlining it ourselves :)

gw-dev101:
on the v side isnan(x) is basically 
```v
return x!=x
```
so it should inline pretty nicely

Linklancien:
All the test files must passed with this push, it least it does on my PC

vlib/math/vec/vec2.v
```diff
@@ -221,6 +221,17 @@ pub fn (v Vec2[T]) magnitude_y() T {
 }
 
 // dot returns the dot product of `v` and `u`.
+// The dot product is a scalar value that represents the magnitude of one vector
```
spytheman, line 224: Doc comments should start with the name of the function/method that follow them.
Reformulate the comment, so that it is so (i.e. the previous one started that way on purpose).

vlib/math/vec/vec2.v
```diff
@@ -221,6 +221,17 @@ pub fn (v Vec2[T]) magnitude_y() T {
...
+// projected onto another vector.
+// It is calculated by multiplying the corresponding components of the vectors
+// and summing the results.
+// example:
+// ```
```
spytheman:
```suggestion
// ```v
```

vlib/math/vec/vec2.v
```diff
@@ -221,6 +221,17 @@ pub fn (v Vec2[T]) magnitude_y() T {
...
+// v := vec2[f32](3, 4) //magnitude = 5
+// u := vec2[f32](5, 6) //magnitude = 7.81
+// dot := v.dot(u) // 3*5 + 4*6 = 15 + 24 = 39
+// 	(dot) // Output: 39
```
spytheman, line 233:
```suggestion
// println(dot) // Output: 39
```

spytheman: Examples that are using ```v blocks, will get extracted and verified with:
 `v doc -unsafe-run-examples -f none vlib/math/vec`
The .github/workflows/module_docs_ci.yml CI will run more or less the above, and check that the examples still work/compile.

spytheman: Another improvement here is to also add `// assert dot == 30`, so that the example will fail at runtime, in addition to printing the value of `dot`, if the tested functionality regresses in the future.

vlib/math/vec/vec2.v
```diff
@@ -253,9 +264,22 @@ pub fn (v Vec2[T]) perpendicular(u Vec2[T]) Vec2[T] {
 }
 
 // project returns the projected vector.
+// The projection of vector `u` onto vector `v` is the orthogonal projection
```
spytheman, line 267: Same - start with `project` here, so that `v doc` can pick it up later.

vlib/math/vec/vec2.v
```diff
@@ -253,9 +264,22 @@ pub fn (v Vec2[T]) perpendicular(u Vec2[T]) Vec2[T] {
...
+// of `u` onto a straight line parallel to `v` that passes through the origin.
+// This is equivalent to the vector projection of `u` onto the unit vector in the direction of `v`.
+// and is given by the formula: proj_v(u) = (u · v / |v|^2) * v
+// where "·" denotes the dot product and |v| is the magnitude of vector `v`.
+// If `u` is a zero vector, the result will also be a zero vector.
+// example:
+// ```
```
spytheman:
```suggestion
// ```v
```

vlib/math/vec/vec2.v
```diff
@@ -253,9 +264,22 @@ pub fn (v Vec2[T]) perpendicular(u Vec2[T]) Vec2[T] {
...
+// v := vec2[f32](3, 4)
+// u := vec2[f32](5, 6)
+// proj := v.project(u)
+// println(proj) // Output: vec2[f32](3.61, 4.81)
```
spytheman, line 278: adding an assertion here would be nice too

vlib/math/vec/vec2.v
```diff
@@ -351,6 +375,13 @@ pub fn (p1 Vec2[T]) angle_towards(p2 Vec2[T]) T {
 }
 
 // angle returns the angle in radians of the vector.
+// example:
+// ```
```
spytheman:
```suggestion
// ```v
```

vlib/math/vec/vec2.v
```diff
@@ -351,6 +375,13 @@ pub fn (p1 Vec2[T]) angle_towards(p2 Vec2[T]) T {
...
+// v := vec2[f32](3.0, 4.0)
+// a := v.angle() // a == 0.64 (approximate value in radians)
+// w := vec2[f32](0.0, 1.0)
+// b := w.angle() // b == 1.57 (approximate value in radians)
```
spytheman:
```suggestion
// a := v.angle() 
// assert a == 0.64 (approximate value in radians)
// w := vec2[f32](0.0, 1.0)
// b := w.angle() 
// assert b == 1.57 (approximate value in radians)
```

vlib/math/vec/vec2.v
```diff
@@ -392,6 +419,14 @@ pub fn (mut v Vec2[T]) clean_tolerance[U](tolerance U) {
 }
 
 // inv returns the inverse, or reciprocal, of the vector.
+// If a field is zero, its inverse is also set to zero to avoid division by zero.
```
spytheman, line 424: start the comment with `inv`

vlib/math/vec/vec2.v
```diff
@@ -392,6 +419,14 @@ pub fn (mut v Vec2[T]) clean_tolerance[U](tolerance U) {
...
+// the direction the vector points is generally not preserved, but
+// the magnitude of each field is inverted.
+// example:
+// ```
```
spytheman:
```suggestion
// ```v
```

vlib/math/vec/vec2.v
```diff
@@ -400,6 +435,13 @@ pub fn (v Vec2[T]) inv() Vec2[T] {
 }
 
 // normalize normalizes the vector.
+// A normalized vector has the same direction as the original vector but a magnitude of 1.
+// If the vector has a magnitude of 0, a zero vector is returned since we cannot find the direction of a zero-length vector.
+// example:
+// ```
```
spytheman:
```suggestion
// ```v
```

vlib/math/vec/vec2.v
```diff
@@ -400,6 +435,13 @@ pub fn (v Vec2[T]) inv() Vec2[T] {
 }
 
 // normalize normalizes the vector.
+// A normalized vector has the same direction as the original vector but a magnitude of 1.
```
spytheman, line 440: start with `normalize`

vlib/math/vec/vec2.v
```diff
@@ -400,6 +435,13 @@ pub fn (v Vec2[T]) inv() Vec2[T] {
...
+// If the vector has a magnitude of 0, a zero vector is returned since we cannot find the direction of a zero-length vector.
+// example:
+// ```
+// v := vec2[f32](3.0, 4.0)//magnitude = 5.0
+// n := v.normalize() // n == vec2[f32](0.6, 0.8) // magnitude = 1.0
```
spytheman, line 445: turn the comment at the end of the line, into a commented assertion on a separate line

vlib/math/vec/vec2.v
```diff
@@ -412,6 +454,11 @@ pub fn (v Vec2[T]) normalize() Vec2[T] {
 }
 
 // sum returns a sum of all the fields.
+// example:
+// ```
+// v := vec2[f32](3.0, 4.0)
+// s := v.sum() // s == 7.0
+// ```
```
spytheman:
```suggestion
// ```v
// v := vec2[f32](3.0, 4.0)
// s := v.sum()
// assert s == 7.0
// ```
```

vlib/math/vec/vec2_test.v
```diff
@@ -228,3 +228,74 @@ fn test_vec2_rotate_around_ccw_2() {
 	assert close(v.x, -1.0)
 	assert close(v.y, -1.0)
 }
+
+// Test for Vec2 projection
+//
+fn test_vec2_project_onto_basic() {
+	v1 := vec.vec2(3.0, 4.0) // magnitude 5 vector
+	v2 := vec.vec2(5.0, 6.0) // magnitude ~7.81 vector
```
spytheman: use `u` and `v` instead of `v1` and `v2` - it is easier to follow the logic that way

vlib/math/vec/vec2.v
gw-dev101: you would have to add a tolerance since floating point arithmetic isn't perfect 
and that would make it equivalent to one of the tests

spytheman: tests are not shown in the documentation, but ok

vlib/math/vec/vec2_test.v
Linklancien: That's done, I don't find where to put this change as done

vlib/math/vec/vec2.v
Linklancien: Does each line of the comment need to start with the name of the function / struct ?

For instance here, line 266 already start with project, does line 267 also need to start with project ?

spytheman: No, only the first line has to start like that.

spytheman:
Thank you @gw-dev101, @tankf33der and @Linklancien 🙇🏻 .

tankf33der:
I didn't hesitate to go back and check my test after the merge and they're fatally failing for `vec3.project()` against `cglm` library. I'll make a new PR if necessary.

PR #25814, opened 2025-11-23
gg: new documentation for the FN** types

Linklancien:
This PR aims to improve a little the gg documentation
Focusing on the descrution of the FN** type used in gg

It is my first PR so I keep it simple to assure that it is clear.
I look forward to any advice to improve my future contributions.

### DIFF
--- a/vlib/gg/gg.v
+++ b/vlib/gg/gg.v
@@ -11,13 +11,13 @@ pub type FNEvent = fn (e &Event, data voidptr)
 // FNEvent2 same as FNEvent with inverted arguments TODO: deprecate this, in favor of event_fn
 pub type FNEvent2 = fn (data voidptr, e &Event)
 
-// FNFail sdefines the type of a function that will be called when there is a fail
+// FNFail defines the type of a function that will be called when there is a fail
 pub type FNFail = fn (msg string, data voidptr)
 
 // FNKeyDown defines the type of a function that will be called for every key pressed down
 pub type FNKeyDown = fn (c KeyCode, m Modifier, data voidptr)
 
-// FNKeyUp defines the type of a function that will be called for every key up to it original position
+// FNKeyUp defines the type of a function that will be called for every release of a pressed key
 pub type FNKeyUp = fn (c KeyCode, m Modifier, data voidptr)
 
 // FNMove defines the type of a function that will be called for every mouse move on the screen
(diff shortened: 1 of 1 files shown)


vlib/gg/gg.v
```diff
@@ -5,20 +5,26 @@ module gg
 
 pub type FNCb = fn (data voidptr)
 
+// Define the type of function that can be given to a Config and that will be called on every events
```
spytheman: Documentation comments should start with the name of the entity, `FNEvent` in this case. It is how `v doc` later can extract them and show them properly.

spytheman: i.e. use something like `// FNEvent defines the type of a function that will be called for every event`

Linklancien: > i.e. use something like `// FNEvent defines the type of a function that will be called for every event`

Okay I will apply those change

vlib/gg/gg.v
```diff
@@ -5,22 +5,31 @@ module gg
 
 pub type FNCb = fn (data voidptr)
 
+// FNEvent defines the type of a function that will be called for every event
 pub type FNEvent = fn (e &Event, data voidptr)
 
+// FNEvent2 same as FNEvent with inverted arguments TODO: deprecate this, in favor of event_fn
 pub type FNEvent2 = fn (data voidptr, e &Event)
 
+// FNFail sdefines the type of a function that will be called when there is a fail
```
spytheman:
```suggestion
// FNFail defines the type of a function that will be called when there is a fail
```

vlib/gg/gg.v
```diff
@@ -5,22 +5,31 @@ module gg
...
+// FNKeyUp defines the type of a function that will be called for every key up to it original position
 pub type FNKeyUp = fn (c KeyCode, m Modifier, data voidptr)
 
+// FNMove defines the type of a function that will be called for every mouse move on the screen
 pub type FNMove = fn (x f32, y f32, data voidptr)
 
+// FNClick defines the type of a function that will be called for every mouse click
 pub type FNClick = fn (x f32, y f32, button MouseButton, data voidptr)
 
+// FNUnClick defines the type of a function that will be called for every the mouse released
```
spytheman:
```suggestion
// FNUnClick defines the type of a function that will be called every time a mouse button is released
```

vlib/gg/gg.v
```diff
@@ -5,22 +5,31 @@ module gg
...
+// FNEvent2 same as FNEvent with inverted arguments TODO: deprecate this, in favor of event_fn
 pub type FNEvent2 = fn (data voidptr, e &Event)
 
+// FNFail sdefines the type of a function that will be called when there is a fail
 pub type FNFail = fn (msg string, data voidptr)
 
+// FNKeyDown defines the type of a function that will be called for every key pressed down
 pub type FNKeyDown = fn (c KeyCode, m Modifier, data voidptr)
 
+// FNKeyUp defines the type of a function that will be called for every key up to it original position
```
spytheman:
```suggestion
// FNKeyUp defines the type of a function that will be called for every release of a pressed key
```

PR #25816, opened 2025-11-23
veb: add zstd compression support

davlgd:
This PR adds `zstd` compression to veb for dynamic/static routes. I've implemented it in a more extensible way (if we want to add brotli for example). `veb` can now serve `zstd` or `gzip` depending on request `Accept-Encoding` headers. 

Priority rules are: 
- zstd over gzip
- pre-compressed files over on-demand compression

User can decide if he lets this automatic mode, or if he wants to provide only gzip/zstd compression. It's a way to keep support for gzip previous configuration values. We could also only keep the automatic way, but it would be a breaking change.

I've renamed `static_gzip_max_size` to a more generic `static_compression_max_size`. I don't consider it as breaking, as it was only introduced some days ago and never "released". 

I've added doc and tests. Let me know if you want me to change anything

### DIFF
--- a/vlib/veb/README.md
+++ b/vlib/veb/README.md
@@ -421,18 +421,20 @@ app.handle_static('static', true)!
 
 ### Compression for static files (zstd/gzip)
 
-veb provides automatic compression (zstd and gzip) for static files with smart caching. When enabled,
-veb will serve compressed versions of your static files to clients that support compression,
-reducing bandwidth usage and improving load times. Zstd is preferred over gzip when the client supports both.
+veb provides automatic compression (zstd and gzip) for static files with smart caching.
+When enabled, veb will serve compressed versions of your static files to clients that
+support compression, reducing bandwidth usage and improving load times. Zstd is preferred
+over gzip when the client supports both.
 
 **How it works:**
 
-1. **Manual pre-compression**: If you create `.zst` or `.gz` files manually, veb will serve them in
-   zero-copy streaming mode for maximum performance.
+1. **Manual pre-compression**: If you create `.zst` or `.gz` files manually, veb will serve
+   them in zero-copy streaming mode for maximum performance.
 2. **Lazy compression cache**: Files smaller than the threshold are automatically compressed
-   on first request and cached as `.zst` or `.gz` files on disk (zstd preferred when client supports it).
-3. **Cache validation**: If the original file is modified, the compressed cache is automatically
-   regenerated on the next request.
+   on first request and cached as `.zst` or `.gz` files on disk (zstd preferred when client
+   supports it).
+3. **Cache validation**: If the original file is modified, the compressed cache is
+   automatically regenerated on the next request.
 4. **Streaming for large files**: Files larger than the threshold are served uncompressed in
    streaming mode (unless a manual `.zst` or `.gz` file exists).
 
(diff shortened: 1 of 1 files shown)


JalonSolov:
You can run `v check-md <md file>` locally to ensure you've fixed everything.

vlib/veb/context.v
```diff
@@ -207,79 +210,64 @@ fn (mut ctx Context) send_file(content_type string, file_path string) Result {
...
+	max_size_bytes := ctx.static_compression_max_size
 
-	// Try to serve pre-compressed .gz file if static gzip is enabled and client accepts it
-	if ctx.enable_static_gzip && client_accepts_gzip {
-		gz_path := '${file_path}.gz'
+	// Determine which compression modes are enabled
+	use_zstd := (ctx.enable_static_zstd && client_accepts_zstd)
+		|| (ctx.enable_static_compression && client_accepts_zstd)
+	use_gzip := (ctx.enable_static_gzip && client_accepts_gzip)
+		|| (ctx.enable_static_compression && client_accepts_gzip)
```
spytheman, line 224:
```suggestion
	use_zstd := client_accepts_zstd && (ctx.enable_static_zstd || ctx.enable_static_compression)
	use_gzip := client_accepts_gzip && (ctx.enable_static_gzip || ctx.enable_static_compression)
```

spytheman: Although thinking about it more, I am not convinced that the whole `enable_static_compression` is needed, since it complicates the logic, documentation and examples, while if it is missing, people can just set the individual flags for each compression kind, and if we introduce more compression formats, their old code will continue to work without surprises, until they explicitly update it to enable the new ones too.

vlib/veb/context.v
```diff
@@ -207,79 +210,64 @@ fn (mut ctx Context) send_file(content_type string, file_path string) Result {
...
-				ctx.res.header.set(.vary, 'Accept-Encoding')
-				ctx.send_response_to_client(content_type, compressed.bytestr())
-				ctx.already_compressed = true
-			}
+			// Compression failed: serve uncompressed in streaming mode
+			ctx.return_type = .file
+			ctx.return_file = file_path
+			ctx.res.header.set(.content_length, file_size.str())
+			ctx.send_response_to_client(content_type, '')
 			return Result{}
```
spytheman:
```suggestion
			return ctx.send_response_to_client(content_type, '')
```
(that way makes it deliberately a bit harder to put more code like setting headers etc after send_response_to_client in the future)

vlib/veb/context.v
```diff
@@ -301,6 +289,71 @@ fn (mut ctx Context) send_file(content_type string, file_path string) Result {
...
+	ctx.already_compressed = true
+	return true
+}
+
+// serve_compressed_static compresses data and serves it, optionally caching to disk.
+// Returns Result on success, none on compression failure.
+fn (mut ctx Context) serve_compressed_static(content_type string, file_path string, data string, encoding ContentEncoding) ?Result {
+	compressed, ext, encoding_name := match encoding {
+		.zstd {
+			c := zstd.compress(data.bytes()) or { return none }
```
spytheman, line 319: We should add a function in `zstd`, that accepts strings directly too, to avoid the conversion here (`C.ZSTD_compress2` ultimately accepts just a pointer and a length) 🤔 .

spytheman: An alternative is to:
`zstd.compress(unsafe{ reuse_string_as_data(data) })`

vlib/veb/context.v
davlgd: The point on this was: if it's enabled, we manage any kind of compression, what's done depends on client support. But I agree on the complexity it adds

PR #25819, opened 2025-11-24
rand: Add i8 function to rand

squidink7:
The rand module was missing an i8 function, despite it being a method on PRNG. Probably just an oversight, however this PR adds one in.


JalonSolov:
May need to merge main into this PR to make it succeed.  The file it's complaining about has been formatted in main.

spytheman (approved):
Good find, thank you 🙇🏻 .

PR #25840, opened 2025-11-27
math: add tests for vec2,3,4 perpendicular

Linklancien:
This PR fix an issue cause because of the change of the project function

It also add some test to assure that won't happend again

### DIFF
--- a/vlib/math/vec/vec2.v
+++ b/vlib/math/vec/vec2.v
@@ -258,7 +258,7 @@ pub fn (v Vec2[T]) perp_ccw() Vec2[T] {
 	return Vec2[T]{-v.y, v.x}
 }
 
-// perpendicular returns the `u` projected perpendicular vector to 'v' vector.
+// perpendicular returns the `v` projected perpendicular vector to the 'u' vector.
 pub fn (v Vec2[T]) perpendicular(u Vec2[T]) Vec2[T] {
 	return v - v.project(u)
 }
--- a/vlib/math/vec/vec3.v
+++ b/vlib/math/vec/vec3.v
@@ -253,7 +253,7 @@ pub fn (v Vec3[T]) unit() Vec3[T] {
 	return Vec3[T]{v.x * invm, v.y * invm, v.z * invm}
 }
 
-// perpendicular returns the `u` projected perpendicular vector to the 'v' vector.
+// perpendicular returns the `v` projected perpendicular vector to the 'u' vector.
 @[inline]
 pub fn (v Vec3[T]) perpendicular(u Vec3[T]) Vec3[T] {
 	return v - v.project(u)
--- a/vlib/math/vec/vec4.v
+++ b/vlib/math/vec/vec4.v
@@ -270,7 +270,7 @@ pub fn (v Vec4[T]) unit() Vec4[T] {
 	}
 }
 
-// perpendicular returns the `u` projected perpendicular vector to the 'v' vector.
+// perpendicular returns the `v` projected perpendicular vector to the 'u' vector.
 pub fn (v Vec4[T]) perpendicular(u Vec4[T]) Vec4[T] {
 	return v - v.project(u)
 }


Linklancien:
It may conflict with #25841 because of the inversion of u and v

tankf33der:
I am firmly certain that this PR is suspicious.

@Linklancien 
project() is incorrect itself for vec2,3,4.

Linklancien:
The main issue I wanted to fix was to introduce a tests for the the perpendicular function to avoid regressions

tankf33der:
@Linklancien 

Our `perpendicular()` is correct. Proof is below:
```c++
namespace glm
{
	template<typename genType>
	GLM_FUNC_QUALIFIER genType perp(genType const& x, genType const& Normal)
	{
		return x - proj(x, Normal);
	}
}//namespace glm
```

Linklancien:
> @Linklancien
> 
> Our `perpendicular()` is correct. Proof is below:
> 
> ```c++
> namespace glm
> {
> 	template<typename genType>
> 	GLM_FUNC_QUALIFIER genType perp(genType const& x, genType const& Normal)
> 	{
> 		return x - proj(x, Normal);
> 	}
> }//namespace glm
> ```

Yes it is
It just lack tests

spytheman:
This PR does not pass its tests either before or after rebasing on #25841 (which is now merged).

Imho there is a bug somewhere.

vlib/math/vec/vec3_test.v
```diff
@@ -131,3 +131,26 @@ fn test_vec3_project_onto_angle() {
...
+	assert tolerance(per.y, v.y, vec.vec_epsilon)
+	assert tolerance(per.z, v.z, vec.vec_epsilon)
+}
+
+// 'u' and 'v' are orthogonal so it must return the null vector
+fn test_vec3_orthogonal_angle() {
+	u := vec.vec3(1.0, 0.0, 0.0)
+	v := vec.vec3(0.0, 3.0, 2.0)
+
+	per := u.perpendicular(v)
```
spytheman: perhaps the inputs here are wrong, since they are the same as lines 138 and 139, thus the result will also be the same, which will fail the asserts.

spytheman: making `v` the same as `u`, makes it pass

Linklancien: I did a wrong copy past, for vec3 and 4

Linklancien:
Thoses push should be good and only add test for perpendicular

JalonSolov:
You need to `v fmt` your tests.

Linklancien:
Oh no

Linklancien:
I forgot to fmt

vlib/math/vec/vec3_test.v
```diff
@@ -131,3 +131,26 @@ fn test_vec3_project_onto_angle() {
...
+	per := v.perpendicular(u)
+	assert tolerance(per.x, v.x, vec.vec_epsilon)
+	assert tolerance(per.y, v.y, vec.vec_epsilon)
+	assert tolerance(per.z, v.z, vec.vec_epsilon)
+}
+
+// 'u' and 'v' are orthogonal so it must return the null vector
+fn test_vec3_orthogonal_angle() {
+	u := vec.vec3(1.0, 0.0, 0.0)
+	v := vec.vec3(3.0, 0.0, 0.0)
```
spytheman: `u` and `v` here are not orthogonal - `v` is a `u` just magnified by 3, but pointing in the same direction

vlib/math/vec/vec4_test.v
```diff
@@ -123,3 +124,28 @@ fn test_vec4_project_onto_zero() {
...
+	assert tolerance(per.x, v.x, vec.vec_epsilon)
+	assert tolerance(per.y, v.y, vec.vec_epsilon)
+	assert tolerance(per.z, v.z, vec.vec_epsilon)
+	assert tolerance(per.w, v.w, vec.vec_epsilon)
+}
+
+// 'u' and 'v' are orthogonal so it must return the null vector
+fn test_vec4_orthogonal_angle() {
+	u := vec.vec4(1.0, 0.0, 0.0, 0.0)
+	v := vec.vec4(3.0, 0.0, 0.0, 0.0)
```
spytheman: same: imho the name for the test should be changed, to something that does not have `orthogonal` in it

vlib/math/vec/vec3_test.v
Linklancien: Oh yes the name is wrong
It is collinear

vlib/math/vec/vec4_test.v
spytheman: perhaps `collinear` will be better?

JalonSolov:
Something useful if you're working on V itself...

From `CONTRIBUTING.md`:

>    3.1 (optional) Run these commands, which ensure that all your code will be
>    automatically formatted, before committing:
>    ```
>    cp cmd/tools/git_pre_commit_hook.vsh .git/hooks/pre-commit
>    chmod 755 .git/hooks/pre-commit
>    ```

JalonSolov:
Side note:  Perhaps that pre-commit trigger should be set by default in the V repo.  Would solve this particular "headache" for new people, and perhaps help some not-so-new.

spytheman:
> Side note: Perhaps that pre-commit trigger should be set by default in the V repo. Would solve this particular "headache" for new people, and perhaps help some not-so-new.

how? should it be put in the makefiles?

Linklancien:
> Something useful if you're working on V itself...
> 
> From `CONTRIBUTING.md`:
> 
> > 3.1 (optional) Run these commands, which ensure that all your code will be
> > automatically formatted, before committing:
> > ```
> > cp cmd/tools/git_pre_commit_hook.vsh .git/hooks/pre-commit
> > chmod 755 .git/hooks/pre-commit
> > ```

I did read that but didn't understand where to execute the command even if it seemed really usefull

spytheman:
I am using the git hook for several years at this point daily, so I guess it is stable enough (I only occasionally have to pass `--no-verify` to `git commit` to skip it, when debugging issues in vfmt itself, but most people will not have to debug vfmt itself) 🤔 .

spytheman:
> I did read that but didn't understand where to execute the command even if it seemed really usefull


It has to be done in a shell where you have done `cd /root/of/your/V_repo` once.
I.e. you have to be in the top level folder, that contains the `.git` folder for the V repository.

JalonSolov:
> > Side note: Perhaps that pre-commit trigger should be set by default in the V repo. Would solve this particular "headache" for new people, and perhaps help some not-so-new.
> 
> how? should it be put in the makefiles?

That would be one way, for certain.

vlib/math/vec/vec2_test.v
```diff
@@ -299,3 +299,24 @@ fn test_vec2_project_negative_components() {
...
+fn test_vec2_perpendicularity_angle() {
+	u := vec.vec2(1.0, 0.0)
+	v := vec.vec2(0.0, 3.0)
+
+	per := v.perpendicular(u)
+	assert tolerance(per.x, v.x, vec.vec_epsilon)
+	assert tolerance(per.y, v.y, vec.vec_epsilon)
+}
+
+// 'u' and 'v' are orthogonal so it must return the null vector
```
spytheman:
```suggestion
// 'u' and 'v' are collinear so it must return the null vector
```

vlib/math/vec/vec4_test.v
```diff
@@ -123,3 +124,28 @@ fn test_vec4_project_onto_zero() {
...
+	v := vec.vec4(0.0, 3.0, 2.0, 0.0)
+
+	per := v.perpendicular(u)
+	assert tolerance(per.x, v.x, vec.vec_epsilon)
+	assert tolerance(per.y, v.y, vec.vec_epsilon)
+	assert tolerance(per.z, v.z, vec.vec_epsilon)
+	assert tolerance(per.w, v.w, vec.vec_epsilon)
+}
+
+// 'u' and 'v' are orthogonal so it must return the null vector
```
spytheman:
```suggestion
// 'u' and 'v' are collinear so it must return the null vector
```

vlib/math/vec/vec3_test.v
```diff
@@ -131,3 +131,26 @@ fn test_vec3_project_onto_angle() {
...
+	u := vec.vec3(1.0, 0.0, 0.0)
+	v := vec.vec3(0.0, 3.0, 2.0)
+
+	per := v.perpendicular(u)
+	assert tolerance(per.x, v.x, vec.vec_epsilon)
+	assert tolerance(per.y, v.y, vec.vec_epsilon)
+	assert tolerance(per.z, v.z, vec.vec_epsilon)
+}
+
+// 'u' and 'v' are orthogonal so it must return the null vector
```
spytheman:
```suggestion
// 'u' and 'v' are collinear so the result must be the null vector
```

vlib/math/vec/vec2.v
```diff
@@ -258,7 +258,7 @@ pub fn (v Vec2[T]) perp_ccw() Vec2[T] {
 	return Vec2[T]{-v.y, v.x}
 }
 
-// perpendicular returns the `u` projected perpendicular vector to this vector.
+// perpendicular returns the `u` projected perpendicular vector to 'v' vector.
```
tankf33der: @Linklancien @spytheman i think this must be in reverse mode for all three functions. Rest are ok.:
```
// perpendicular returns the `v` projected perpendicular vector to 'u' vector.
```

Linklancien: Oh yes I change that

Linklancien: It must be corrected

tankf33der (approved):
The main thing is that our `perpendicular()` function is correct and new tests never hurt.

tankf33der:
@spytheman lets merge.

spytheman (approved):

spytheman:
Thank you @Linklancien 🙇🏻 .
Thank you @tankf33der 🙇🏻 .

PR #25841, opened 2025-11-27
math.vec: second attempt to fix project() for vec2,3,4

tankf33der:
The previous authors did a lot of work, but still got confused, sigh.
The changes turned out to be too large. The most frustrating thing is that I posted the correct solution to PR, but it was completely ignored. I decided to fix it with a single patch as well :). 

The function, comments, tests, line order, and the order of using variables `v` and `u` have been changed. The `vec3` version was tested against the `cglm` library, and then all the others were based on it.

### DIFF
--- a/vlib/math/vec/vec2.v
+++ b/vlib/math/vec/vec2.v
@@ -264,22 +264,22 @@ pub fn (v Vec2[T]) perpendicular(u Vec2[T]) Vec2[T] {
 }
 
 // project returns the projected vector.
-// The projection of vector `u` onto vector `v` is the orthogonal projection
-// of `u` onto a straight line parallel to `v` that passes through the origin.
-// This is equivalent to the vector projection of `u` onto the unit vector in the direction of `v`.
-// and is given by the formula: proj_v(u) = (u · v / |v|^2) * v
-// where "·" denotes the dot product and |v| is the magnitude of vector `v`.
-// If `u` is a zero vector, the result will also be a zero vector.
+// The projection of vector `v` onto vector `u` is the orthogonal projection
+// of `v` onto a straight line parallel to `u` that passes through the origin.
+// This is equivalent to the vector projection of `v` onto the unit vector in the direction of `u`.
+// and is given by the formula: proj_v(u) = (v · u / |u|^2) * u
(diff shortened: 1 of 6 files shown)


tankf33der:
> Just to be clear, what you change was the order of u and v in the project function and it's test ?
> 
> Thanks for your hard work If you want, I can draft my PR waiting for yours to be merged and then PR for the perpidencular test

@Linklancien 
why do you think perpedencular() is broken? How to prove?

Linklancien:
> > Just to be clear, what you change was the order of u and v in the project function and it's test ?
> > 
> > Thanks for your hard work If you want, I can draft my PR waiting for yours to be merged and then PR for the perpidencular test
> 
> @Linklancien 
> why do you think perpedencular() is broken? How to prove?

With your second fix of project() I think it won't be broken anymore
So I juste want to add tests that will guarantee it

Linklancien:
So it isn't about fixing the function now, juste making test

I hope I am clear 😅

tankf33der:
This is my prove my patch from this PR is correct.

```v
import math.vec
import rand


#pkgconfig cglm

type V3 = [3]f32
fn C.glmc_vec3_proj(V3, V3, V3)

fn main() {
	a1 := rand.f32_in_range(-10.0, 10.0)!
	b1 := rand.f32_in_range(-10.0, 10.0)!
	c1 := rand.f32_in_range(-10.0, 10.0)!
	a2 := rand.f32_in_range(-10.0, 10.0)!
	b2 := rand.f32_in_range(-10.0, 10.0)!
	c2 := rand.f32_in_range(-10.0, 10.0)!

	mut v := V3{}
	mut u := V3{}
	mut x := V3{}
	v = V3([a1, b1, c1]!)
	u = V3([a2, b2, c2]!)
	C.glmc_vec3_proj(v, u, x)

	vlang_v := vec.vec3[f32](a1, b1, c1)
	vlang_u := vec.vec3[f32](a2, b2, c2)
	vlang_x := vlang_v.project(vlang_u)

	println(x)
	println(vlang_x)
}
```

spytheman:
> The most frustrating thing is that I posted the correct solution to PR, but it was completely ignored. I decided to fix it with a single patch as well :).

You can adjust PRs too @tankf33der , and I do trust your judgement.

PR #25843, opened 2025-11-27
cgen,checker: handle array decompose for dynamic arrays and array init

### LINKED ISSUE #25838: Passing array with variadic `...` to function with several args seems bug but maybe not.
### Describe the bug

Originally case mentioned here https://github.com/vlang/v/issues/25827#issuecomment-3576696815 seems to be a bug but further analysis seems a *correct* way to pass some elements of an array to a function with several arguments. Like the inverse of passing from `...args` to `array`.

This issue is open to discuss whether bug or not.

### Reproduction Steps

Next code show *first elements* of an array can be passed to functions having several arguments.

```v
fn print_1(s string) {
	dump(s)
}

fn print_2(s string, t string) {
	dump('${s} ${t}')
}

fn print_3(s string, t string, u string) {
	dump('${s} ${t} ${u}')
}

fn print_4(s string, t string, u string, v string) {
	dump('${s} ${t} ${u} ${v}')
}

array := ['a','b','c']
print_1(array[0]) // prints `a`
print_1(...array) // prints `a` unexpectedly (originally)
print_2(...array) // prints `a` `b`
print_3(...array) // prints `a` `b` `c`
// print_4(...array) // PANIC
```
As far the array is not larger than the number of function arguments the array elements are passed as individual arguments otherwise panic.

Now seems logical that `...array` passes 1, 2 or more elements for functions with 1, 2 or more arguments.

So now `print_3(...array)` works the same that longer call `print_3(array[0], array[1], array[2])`

### Expected Behavior

Originally though that `call(...array)` should not be accepted for function `fn call(s string)` and be marked as V error but since also works for functions with several arguments seems to be *inverse* of `...args` to `array`.

### Current Behavior

```
[code.v:2] s: a
[code.v:2] s: a
[code.v:6] '$s $t': a b
[code.v:10] '$s $t $u': a b c
```

### Possible Solution

Just document this possibility and limits.

_No response_

### V version

V 0.4.12 64f1162 

https://play.vlang.io/p/5be2d7a401
> [!NOTE]
>

dy-tea:
Fixes #25838.

Doing an array decompose on an array init can be verified by the checker, but this is not the case for dynamic arrays. Here I instead am adding a panic message that should be more useful.

Fixed array decomposition isn't allowed so no handling is needed for that.

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -1524,6 +1524,20 @@ fn (mut c Checker) fn_call(mut node ast.CallExpr, mut continue_check &bool) ast.
 			}
 		}
 		has_decompose = call_arg.expr is ast.ArrayDecompose
+		if !func.is_variadic && call_arg.expr is ast.ArrayDecompose {
+			if mut call_arg.expr is ast.ArrayDecompose {
+				if call_arg.expr.expr is ast.ArrayInit {
+					extra_params := func.params.len - i
+					array_init := call_arg.expr.expr as ast.ArrayInit
+					if array_init.exprs.len < extra_params {
+						elem_word := if array_init.exprs.len == 1 { 'element' } else { 'elements' }
+						verb_word := if extra_params == 1 { 'is' } else { 'are' }
+						c.error('array decompose has ${array_init.exprs.len} ${elem_word} but ${extra_params} ${verb_word} needed for `${func.name}`',
+							call_arg.pos)
+					}
+				}
+			}
+		}
 		already_checked := node.language != .js && call_arg.expr is ast.CallExpr
 		if func.is_variadic && param_i >= func.params.len - 1 {
 			param_sym := c.table.sym(param.typ)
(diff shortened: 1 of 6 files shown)


spytheman:
@dy-tea Excellent work. Thank you.


Array decomposition initially was only intended to be used for passing dynamic arrays as a parameter to functions, that accepted a variadic argument:
```v
fn f(args ...int) {
    println(args)
}
f(...[1,2,3]) // using a decomposition
f(10,20,30) // using a direct variadic call
```

It was not intended to be used for passing args to a normal function, that just happens to have arguments of the same type:
```v
fn g(a1 int, a2 int, a3 int) {
    println('a1: ${a1} | a2: ${a2} | a3: ${a3}')
}
g(...[1,2])
```

At this point however, it may be worth reconsidering, if V should stick to that, and introduce a checker error to limit decomposition to its original `f()` usecase, or add more complexity to cgen to support better the `g()` case.

PR #25844, opened 2025-11-27
crypto.cipher: fix xor_key_stream() for OFB mode, add test

tankf33der:
The tests showed that byte-by-byte operation does not work correctly, while block-by-block operation works properly in OFB mode.

Trivial typo in the code that was not tested, proof [#1](https://github.com/vlang/v/blob/master/vlib/crypto/cipher/ctr.v#L71) and [#2](https://github.com/vlang/v/blob/master/vlib/crypto/cipher/cfb.v#L89).

The test vector was created using `openssl`.

### DIFF
--- a/vlib/crypto/cipher/ofb.v
+++ b/vlib/crypto/cipher/ofb.v
@@ -56,7 +56,7 @@ pub fn (mut x Ofb) xor_key_stream(mut dst []u8, src []u8) {
 
 			copy(mut x.next, x.out)
 
-			n := xor_bytes(mut local_dst, local_src, x.out)
+			n := xor_bytes(mut local_dst, local_src, x.out[x.out_used..])
 			local_dst = local_dst[n..]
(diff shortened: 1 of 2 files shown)


tankf33der:
@blackshirt, @einar-hjortdal - please take a look.

tankf33der:
Only this patch allows run test successfully against `openssl`.

einar-hjortdal:
Seems good to me, I would maybe add in the comments how the test array was obtained (openssl version etc.). But that's minor and may be unnecessary.
@blackshirt is the expert, I am just a user

blackshirt:
> Seems good to me, I would maybe add in the comments how the test array was obtained (openssl version etc.). But that's minor and may be unnecessary. @blackshirt is the expert, I am just a user

Not really true for the expert thing ..it share the same thiing with the others, just some user..

PR #25848, opened 2025-11-28
builder: switch to c99 for linux

### LINKED ISSUE #25833: Name "linux" breaks gcc builds
### Describe the bug

Using "linux" as a var name or struct field name causes gcc builds to fail. This is apparently because gcc has a predefined macro named "linux".

### Reproduction Steps

```v
// test.v
fn main() {
	linux := 'hello world'
	println(linux)
}
```
```
v -cc gcc test.v
```

### Expected Behavior

I would expect V to handle this gracefully by choosing a different name to use in the generated C code.

### Current Behavior

```
cc: /tmp/v_1000/test.01KAYYXPNRVCHV3HV6DVENV5BS.tmp.c: In function ‘main__main’:
cc: /tmp/v_1000/test.01KAYYXPNRVCHV3HV6DVENV5BS.tmp.c:5207:16: error: expected identifier or ‘(’ before numeric constant
cc:  5207 |         string linux = _S("hello world");
cc:       |                ^~~~~
cc: /tmp/v_1000/test.01KAYYXPNRVCHV3HV6DVENV5BS.tmp.c:144:77: error: expected statement before ‘)’ token
cc:   144 | #define _S(s) ((string){.str=(byteptr)("" s), .len=(sizeof(s)-1), .is_lit=1})
cc:       |                                                                             ^
cc: /tmp/v_1000/test.01KAYYXPNRVCHV3HV6DVENV5BS.tmp.c:5207:24: note: in expansion of macro ‘_S’
cc:  5207 |         string linux = _S("hello world");
cc:       |                        ^~
cc: /tmp/v_1000/test.01KAYYXPNRVCHV3HV6DVENV5BS.tmp.c:5208:26: error: incompatible type for argument 1 of ‘builtin__println’
cc:  5208 |         builtin__println(linux);
...
cc:       |                       ~~~~~~~^
(note: the original output was 18 lines long; it was truncated to its first 12 lines + the last line)
builder error: 
==================
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.12 85ff886

|V full version      |V 0.4.12 ce3a1611a16cbeecbe4b46f407de84e03eb08668.85ff886
|OS                  |linux, "Arch Linux"
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8350U CPU @ 1.70GHz
|Memory              |1.37GB/23.22GB
|                    |
|V executable        |/home/scott/builds/v/v
|V last modified time|2025-11-26 02:25:02
|                    |
|V home dir          |OK, value: /home/scott/builds/v
|VMODULES            |OK, value: /home/scott/.vmodules
|Current working dir |OK, value: /home/scott
|                    |
|V git status        |weekly.2025.13-1240-g85ff8864
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20251112
|gcc version         |gcc (GCC) 15.2.1 20251112
|clang version       |clang version 21.1.6
|emcc version        |N/A

> [!NOTE]
>

dy-tea:
Fixes #25833.

`linux` gets defined by `#define linux 1` on gcc and clang for `-std=gnu99` if you are running on linux.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -111,7 +111,7 @@ ifdef LEGACY
 	rm -rf $(TMPLEGACY)
 	$(eval override LDFLAGS+=-L$(realpath $(LEGACYLIBS))/lib -lMacportsLegacySupport)
 endif
-	$(CC) $(CFLAGS) -std=gnu99 -w -o v1$(EXE_EXT) $(VC)/$(VCFILE) -lm -lpthread $(LDFLAGS) || cmd/tools/cc_compilation_failed_non_windows.sh
+	$(CC) $(CFLAGS) -std=c99 -w -o v1$(EXE_EXT) $(VC)/$(VCFILE) -lm -lpthread $(LDFLAGS) || cmd/tools/cc_compilation_failed_non_windows.sh
 ifdef NETBSD
 	paxctl +m v1$(EXE_EXT)
 endif
@@ -229,4 +229,3 @@ etags:
 
 ctags:
 	./v$(EXE_EXT) -print-v-files cmd/v | grep -v :parse_text| ctags -L -
-
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -11,9 +11,7 @@ import v.vcache
 import term
 
 const c_std = 'c99'
-const c_std_gnu = 'gnu99'
 const cpp_std = 'c++17'
-const cpp_std_gnu = 'gnu++17'
 
 const c_verror_message_marker = 'VERROR_MESSAGE '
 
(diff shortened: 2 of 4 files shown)


spytheman:
I do not like this :-| because `linux` is not a reserved word in C, and other compilers can have different predefined macros too, i.e. ~~it creates the precedent of adding a never ending stream of special cases to the blacklist~~.

edit: the precedent was already created by `unix` :-| .

spytheman:
Can we explore the `-std=c99` approach instead?

dy-tea:
`unix` is defined in this array as well which is not a reserved keyword but a macro.

spytheman:
> `unix` is defined in this array as well which is not a reserved keyword but a macro.

And I do not like that too :-( .

dy-tea:
Yeah the `c99` approach may be better, we can remove both of them then.

spytheman:
The change was initially done in https://github.com/vlang/v/pull/12236 , not sure if it is still relevant (I guess one way to find out is to `v test vlib/sync` with a changed compiler).

dy-tea:
Appears to work. Let me try pushing so I can test the full CI.

spytheman:
So far, so good - it seems that only the sanitized jobs remain, which will probably pass too (afaik that change does not affect code generation in anyway, it is purely a frontend thing for the C compilers)

spytheman:
btw, the reason the CI does ~2x as many checks for the PR, is because it is using a branch of the main V repo, not one on a fork, so it triggered both the `push` and `pull_request` handlers in the .yml files

spytheman:
All CI jobs, passed on one trigger or another (the remaining one will duplicate one of the already passed ones).

Excellent work @dy-tea , thank you 🙇🏻 .

PR #25863, opened 2025-11-30
math.complex: fix log(), replace tests

tankf33der:
Incorrect handling of `WolframAlpha` parameters and mathematical notation led to the error not being detected during development and testing.

Some systems have the base as the first parameter of the function.

`vlang` has it second: 
image

Great `mpmath` has it second too and with correct formula:
image

But `WolframAlpha` has it first:
image

Had to remove all current tests and replace them with one that is consistent with `cmath` from the Python distribution and the `mpmath` library.

p.s. I wrote tests for the entire `complex` module against `complex.h` file for the functions that match, and other functions are awaiting their turn for manual testing.

### DIFF
--- a/vlib/math/complex/complex_test.v
+++ b/vlib/math/complex/complex_test.v
@@ -295,8 +295,8 @@ fn test_complex_log() {
 	a := cmplx.complex(11.22, 33.44)
 	b := cmplx.complex(55.66, 77.88)
 	c := a.log(b)
-	assert c.re == 0.8032210844549097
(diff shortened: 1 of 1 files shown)


tankf33der:
I will handle this errors.

tankf33der:
The test showed that comparing `f64` using `==` is a bad idea, which is natural and would depend on the system. Therefore, [this](https://github.com/vlang/v/blob/master/vlib/math/complex/complex.v#L375) also needs to be replaced in the future.

PR #25867, opened 2025-11-30
encoding.binary: use unions for small speed increase and readability

JalonSolov:
Easier to read code (verified on discord) plus small performance increase.


Delta456 (approved):

spytheman:
The code here is longer in all cases, but if it is more performant, fine.

PR #25868, opened 2025-11-30
arrays: add subtract/2

gechandesu:
Added a generic function `subtract` for subtracting array elements and a test.

```v
assert arrays.subtract([1, 2, 3, 4, 5, 6, 7], [3, 5, 6]) == [1, 2, 4, 7]
```

This is a convenient way to solve quite often task to find the difference between sets of elements. `arrays.diff` module is not suitable for this.

### DIFF
--- a/vlib/arrays/arrays.v
+++ b/vlib/arrays/arrays.v
@@ -782,14 +782,14 @@ pub fn each_indexed[T](a []T, cb fn (i int, e T)) {
 	}
 }
 
-// substract returns a new array consisting of the elements of array `a` that are not
+// subtract returns a new array consisting of the elements of array `a` that are not
 // among the elements of array `b`. Performance of this implementation tends to O(n+m),
 // where n and m are the lengths of input arrays.
 //
 // Note: The array element type must be valid type for V `map` keys.
 //
-// Example: arrays.substract([1, 2, 3, 4, 5, 6, 7], [3, 5, 6]) // -> [1, 2, 4, 7]
-pub fn substract[T](a []T, b []T) []T {
+// Example: arrays.subtract([1, 2, 3, 4, 5, 6, 7], [3, 5, 6]) // -> [1, 2, 4, 7]
+pub fn subtract[T](a []T, b []T) []T {
 	mut result := []T{cap: a.len}
 	mut bset := map[T]u8{}
 	for elem in b {
--- a/vlib/arrays/arrays_test.v
+++ b/vlib/arrays/arrays_test.v
@@ -605,9 +605,9 @@ fn test_each_indexed() {
 	assert control_sum == sum
 }
 
-fn test_substract() {
-	assert substract([1, 2, 3, 4, 5, 6, 7], [3, 5, 6]) == [1, 2, 4, 7]
-	assert substract([1, 2, 3], [1, 2, 3]) == []
-	assert substract([]int{}, [1, 2, 3]) == []
-	assert substract([]int{}, []int{}) == []
+fn test_subtract() {
+	assert subtract([1, 2, 3, 4, 5, 6, 7], [3, 5, 6]) == [1, 2, 4, 7]
+	assert subtract([1, 2, 3], [1, 2, 3]) == []
+	assert subtract([]int{}, [1, 2, 3]) == []
+	assert subtract([]int{}, []int{}) == []
 }


tankf33der:
@gechandesu what if `subtract`? Check spelling.

gechandesu:
> @gechandesu what if `subtract`? Check spelling.

You're right, my english is bad...

tankf33der:
My tests show this is an equivalent function. What do you think?

```v
fn mike[T](a []T, b []T) []T {
	mut result := []T{cap: a.len}
	for elem in a {
		if elem !in b {
			result << elem
		}
	}
	return result
}
```

gechandesu:
> My tests show this is an equivalent function. What do you think?
> 
> ```v
> fn mike[T](a []T, b []T) []T {
> 	mut result := []T{cap: a.len}
> 	for elem in a {
> 		if elem !in b {
> 			result << elem
> 		}
> 	}
> 	return result
> }
> ```

This is **much** slower on large arrays. I've tested it `elem !in b` actually is a nested loop over `b` array.

In the worst case (`O(n*m)` vs `O(n+m)`):
```
/tmp $ v run arrays.v
 SPENT    10.685 ms in subtract
 SPENT  8108.553 ms in mike
```

<details><summary>Benchmark</summary>
<p>

```v
import benchmark

fn mike[T](a []T, b []T) []T {
	mut result := []T{cap: a.len}
	for elem in a {
		if elem !in b {
			result << elem
		}
	}
	return result
}

fn subtract[T](a []T, b []T) []T {
	mut result := []T{cap: a.len}
	mut b_set := map[T]bool{}
	for elem in b {
		b_set[elem] = false
	}
	for elem in a {
		if elem !in b_set {
			result << elem
		}
	}
	return result
}

fn main() {
	iters := 1
	len := 100_000
	a := []int{len: len, init: 1}
	b := []int{len: len}
	mut bench := benchmark.start()
	for _ in 0 .. iters {
		_ := subtract(a, b)
	}
	bench.measure('subtract')
	for _ in 0 .. iters {
		_ := mike(a, b)
	}
	bench.measure('mike')
}
```

</p>
</details> 

I'd like to avoid map, as it limits the input types, but the price is very low performance. Of course, I could add comptime type check and use a slower algorithm for anything that isn't supported as a `map` key, but for now I've decided to leave it as is.

tankf33der:
For science, i would like to show where `mike()` is faster too.
```v
import benchmark
import rand

fn mike[T](a []T, b []T) []T {
        mut result := []T{cap: a.len}
        for elem in a {
                if elem !in b {
                        result << elem
                }
        }
        return result
}

fn subtract[T](a []T, b []T) []T {
        mut result := []T{cap: a.len}
        mut b_set := map[T]bool{}
        for elem in b {
                b_set[elem] = false
        }
        for elem in a {
                if elem !in b_set {
                        result << elem
                }
        }
        return result
}

fn main() {
        iters := 1_000_000
        mut bench := benchmark.start()
        for _ in 0 .. iters {
                a := rand.bytes(rand.u8())!
                b := rand.bytes(rand.u8())!
                _ := subtract(a, b)
        }
        bench.measure('subtract')
        for _ in 0 .. iters {
                a := rand.bytes(rand.u8())!
                b := rand.bytes(rand.u8())!
                _ := mike(a, b)
        }
        bench.measure('mike')
}
```

gechandesu:
Yes, map allocations may slow down on small arrays.
```
/tmp $ v run arrays.v
 SPENT 20909.702 ms in subtract
 SPENT 12172.886 ms in mike
/tmp $ v run arrays.v
 SPENT 20919.367 ms in subtract
 SPENT 12194.749 ms in mike
/tmp $ v run arrays.v
 SPENT 20893.922 ms in subtract
 SPENT 12174.510 ms in mike
```
I'll try to find input data on which the implementation without `map` starts to work slower than the implementation with `map` to write a combined version that will select the version of the algorithm based on the length of the arrays. It is okay?

jorgeluismireles:
Or the word **difference** could be used as more appropiate for sets: https://en.wikipedia.org/wiki/Set_(mathematics)#Set_difference
Venn0110

gechandesu:
True. I thought about it some more. I close this PR since the same thing is already implemented in `datatypes.Set`, I missed that. It's better than a new ugly function with inconsistent behavior.

Issue #25860, opened 2025-11-30
Checker disallows `t as Type` Option unwrapping inside `if` expression


gechandesu:
<details>
<summary>V version: V 0.4.12 c3e0637, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 ebf629dc4124a6207b18028cf89dc6910d25eaf3.c3e0637
|:-------------------|:-------------------
|OS                  |linux, Linux version 6.17.6-arch1-1 (linux@archlinux) (gcc (GCC) 15.2.1 20250813, GNU ld (GNU Binutils) 2.45.0) #1 SMP PREEMPT_DYNAMIC Wed, 29 Oct 2025 22:23:47 +0000
|Processor           |16 cpus, 64bit, little endian, 12th Gen Intel(R) Core(TM) i5-1240P
|Memory              |1.66GB/15.34GB
|                    |
|V executable        |/home/ge/.vlang/v
|V last modified time|2025-11-29 15:34:32
|                    |
|V home dir          |OK, value: /home/ge/.vlang
|VMODULES            |OK, value: /home/ge/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /tmp
|                    |
|env VFLAGS          |"-Wimpure-v"
|Git version         |git version 2.51.2
|V git status        |c3e06373
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20250813
|gcc version         |gcc (GCC) 15.2.1 20250813
|clang version       |clang version 21.1.4
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.42

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  opt_unwrap.v && opt_unwrap`
```v
fn foo(s ?string) int {
	if s as string != '' {
		return 1
	}
	return 0
}

fn main() {
	assert foo(none) == 0
	assert foo('') == 0
	assert foo('x') == 1
}

```

**What did you see?**
```
opt_unwrap.v:2:5: error: variable `s` is an Option, it must be unwrapped first
    1 | fn foo(s ?string) int {
    2 |     if s as string != '' {
      |        ^
    3 |         return 1
    4 |     }
```

**What did you expect to see?**

`s` unwrapped

[...]

PR #25869, opened 2025-11-30
checker, markused: allow `if s as ?string`

Delta456:
Fixes https://github.com/vlang/v/issues/25860

### DIFF
--- a/vlib/v/tests/options/option_unwrap_as_cast_test.v
+++ b/vlib/v/tests/options/option_unwrap_as_cast_test.v
@@ -12,6 +12,13 @@ fn get_opt_int(a int) ?int {
 	return 12
 }
 
+fn foo(s ?string) int {
+	if s as string != '' {
+		return 1
+	}
+	return 0
+}
+
 fn test_option_unwrap_as_cast() {
 	x := get_opt(1)
 	d := get_opt_int(12)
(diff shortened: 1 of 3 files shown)


vlib/v/tests/options/option_unwrap_as_cast_test.v
```diff
@@ -12,6 +12,13 @@ fn get_opt_int(a int) ?int {
 	return 12
 }
 
+fn foo(s ?string) int {
+	if s as string != '' {
```
spytheman, line 16: `s as string` is not well defined when `s` is `?string`

spytheman: when the value of `s` is `none`, it has no valid string representation

spytheman: Consider for example also a modifier version like:
```v
struct Struct {
        x int = 45
}

fn foo(s ?Struct) int {
        if s as Struct != Struct{} {
                return 1
        }
        return 0
}

dump( foo(Struct{}) )
dump( foo(Struct{123}) )
dump( foo(none) )
```

Delta456: We already allow `s as string` where `s` is `?string` in general

spytheman: > We already allow `s as string` where `s` is `?string` in general

where is that used?

medvednikov:
`?string as string` is not valid V, it should result in a V compiler error.

`as` can only be used with interfaces and sum types (and imported modules).

PR #25870, opened 2025-12-01
pkgconfig: support multiple packages per line

### LINKED ISSUE #12944: Pkgconfig: error without description
**V version:** V 0.2.4 db1175e.a2468e1
**OS:** Pop Os 21.10

**What did you do?**
I ran the vui webview example

Code with error:
#pkconfig gtk+-3.0 webkit2gtk-4.0

An error with description 
**What did you see instead?**
An error without description

Trying this with the pkgconfig cli gives as output "none" which is probably unexspected and throws an error but there is no string for the error.

I think the problem is that there are 2 "pkgs" defined in one statement which seems to not work(no information about this in the docs), but there is no check for that and it just errors with no description.

It should also be documentet that a "statement"(#pkgconfig pkg1, #pgconfig pk2...) per "pkg" is needed.

dy-tea:
Fixes #12944.

Current behavior:
image

### DIFF
--- a/vlib/v/pkgconfig/main.v
+++ b/vlib/v/pkgconfig/main.v
@@ -94,7 +94,7 @@ pub fn (mut m Main) run() !string {
 			res += pcdep.description
 		}
 		if unsafe { pc != 0 } {
-			pc.extend(pcdep)!
+			pc.extend(pcdep)
 		} else {
 			pc = pcdep
 		}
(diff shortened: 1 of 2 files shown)


dy-tea:
Added some logic so it also checks for flags.

spytheman:
The documentation states explicitly that `Multiple modules can be passed` - <https://docs.vlang.io/v-and-c.html##pkgconfig> .

dy-tea:
Maybe I need to change the docs as well then. Because currently you can't. 

Or allow multiple per line.

dy-tea:
Fixed so it instead supports multiple packages per line.

dy-tea:
Yeah wasn't sure about it either. Would likely have to check for a package already on the CI.

PR #25875, opened 2025-12-01
cgen: fix as cast on mutable receiver

### LINKED ISSUE #25861: `t as Type` Option unwrapping causes C error with `mut` struct fields
<details>
<summary>V version: V 0.4.12 c3e0637, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 ebf629dc4124a6207b18028cf89dc6910d25eaf3.c3e0637
|OS                  |linux, Linux version 6.17.6-arch1-1 (linux@archlinux) (gcc (GCC) 15.2.1 20250813, GNU ld (GNU Binutils) 2.45.0) #1 SMP PREEMPT_DYNAMIC Wed, 29 Oct 2025 22:23:47 +0000
|Processor           |16 cpus, 64bit, little endian, 12th Gen Intel(R) Core(TM) i5-1240P
|Memory              |1.29GB/15.34GB
|                    |
|V executable        |/home/ge/.vlang/v
|V last modified time|2025-11-29 15:34:32
|                    |
|V home dir          |OK, value: /home/ge/.vlang
|VMODULES            |OK, value: /home/ge/.vmodules
|Current working dir |OK, value: /tmp
|                    |
|env VFLAGS          |"-Wimpure-v"
|Git version         |git version 2.51.2
|V git status        |c3e06373
|.git/config present |true
|                    |
|clang version       |clang version 21.1.4
|emcc version        |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  opt_unwrap_struct_field.v && opt_unwrap_struct_field`
```v
struct Foo {
mut:
	state ?string
}

fn (mut c Foo) foo() int {
	if c.state as string != '' {
		return 1
	}
	return 0
}

fn main() {
	mut f1 := Foo{}
	mut f2 := Foo{''}
	mut f3 := Foo{'foo'}
	assert f1.foo() == 0
	assert f2.foo() == 0
	assert f3.foo() == 1
}

```

**What did you see?**
```
cc: /tmp/v_1000/opt_unwrap_struct_field.01KB96CK7RE4EDW1CP4D30KKWF.tmp.c:4929: error: struct or union expected
builder error: 
==================
```

Successful compilation

**Additional info**

Unwrapping works until receiver argument is not `mut`.

So this code works as expected:
```v
struct Foo {
mut:
	state ?string
}

fn (c Foo) foo() int {
	if c.state as string != '' {
		return 1
	}
	return 0
}

fn main() {
	f1 := Foo{}
	f2 := Foo{''}
	f3 := Foo{'foo'}
	assert f1.foo() == 0
	assert f2.foo() == 0
	assert f3.foo() == 1
}

```

felipensp:
Fix #25861

### DIFF
--- a/vlib/v/tests/options/option_as_cast_test.v
+++ b/vlib/v/tests/options/option_as_cast_test.v
@@ -1,0 +1,20 @@
+struct Foo {
+mut:
+	state ?string
+}
+
+fn (mut c Foo) foo() int {
+	if c.state as string != '' {
+		return 1
+	}
+	return 0
+}
+
+fn test_main() {
+	mut f1 := Foo{}
+	mut f2 := Foo{''}
+	mut f3 := Foo{'foo'}
+	assert f1.foo() == 0
+	assert f2.foo() == 0
+	assert f3.foo() == 1
+}
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -7992,7 +7992,12 @@ fn (mut g Gen) as_cast(node ast.AsCast) {
 		} else if node.expr is ast.SelectorExpr {
 			if node.expr.expr is ast.Ident && node.expr.typ.has_flag(.option)
 				&& !unwrapped_node_typ.has_flag(.option) {
-				g.unwrap_option_type(node.expr.typ, '${node.expr.expr.name}.${node.expr.field_name}',
+				dot_or_ptr := if node.expr.expr.is_mut() {
+					'->'
+				} else {
+					g.dot_or_ptr(node.expr.typ)
+				}
+				g.unwrap_option_type(node.expr.typ, '${node.expr.expr.name}${dot_or_ptr}${node.expr.field_name}',
 					node.expr.expr.is_auto_heap())
 				is_optional_ident_var = true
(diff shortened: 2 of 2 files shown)


vlib/v/tests/options/option_as_cast_test.v
```diff
@@ -0,0 +1,20 @@
+struct Foo {
+mut:
+	state ?string
+}
+
+fn (mut c Foo) foo() int {
+	if c.state as string != '' {
```
spytheman, line 7: I am not convinced that `c.state as string` should work, since .state is `?string`, and has to be unwrapped first, and once that is done, it is `string`, in which case the `as string` is a NOP.

I am open to an argument why would it need to work, but keep also in mind non trivial cases, like changed field default values, and more complex types like structs that also can have custom default values etc, in which case the direct casting may be bogus.

spytheman: It is very similar to #25869 , and perhaps @medvednikov should decide for both.

medvednikov:
`?string as string` is not valid V, it should result in a V compiler error.

`as` can only be used with interfaces and sum types (and imported modules).

PR #25881, opened 2025-12-02
cgen: allow @[cinit] for const variables

wenxuanjun:
This PR adds support for the `@[cinit]` attribute on `const` declarations, allowing them to be initialized statically in C instead of at runtime inside `_vinit`.

### Motivation
Currently, V initializes complex constants (like structs or arrays) at runtime within the generated `_vinit` function. While safe, this is problematic for:
1.  **Bare-metal/Embedded development**: Where the runtime environment (and `_vinit`) might not be available, or where hardware descriptors need to be present in memory immediately upon boot.
2.  **Startup optimization**: Reducing the work done at program startup.

`__global` variables already support `@[cinit]` for this purpose. This PR extends that support to `const`.

### Implementation
The change detects if the `cinit` attribute is present on a const field or the const block. If detected, it bypasses the standard "init later" logic and generates a direct C static initializer (e.g., `Type name = { ... };`).

#### V Code
```v
struct Driver {
	name  &char
	probe fn ()
}

fn my_probe() {}

// Without this PR: Initialized in _vinit
// With this PR: Initialized statically in C
@[cinit]
const my_driver = Driver{
	name: c'HID'
	probe: my_probe
}
```

#### Generated C Code (Before)
```c
main__Driver _const_main__my_driver; // inited later

void _vinit(int ___argc, voidptr ___argv) {
    _const_main__my_driver = ((main__Driver){.name = "HID",.probe = (voidptr)main__my_probe,});
}
```

#### Generated C Code (After)
```c
main__Driver _const_main__my_driver = ((main__Driver){.name = "HID",.probe = (voidptr)main__my_probe,});
```

### Notes

Using `@[cinit]` assumes the programmer provides a valid C static constant expression. If the expression contains a function call (other than macros that expand to constants), the C compiler will emit an error, which is the expected behavior for `cinit`.

### DIFF
--- a/vlib/v/gen/c/testdata/const_init_cinit.c.must_have
+++ b/vlib/v/gen/c/testdata/const_init_cinit.c.must_have
@@ -1,1 +1,1 @@
-static main__Config _const_main__conf = ((main__Config){.id = 123,.name = "test_config",.valid = true,});
+main__Config _const_main__conf = ((main__Config){.id = 123,.name = "test_config",.valid = true,});


vlib/v/gen/c/testdata/const_init_cinit.c.must_have
```diff
@@ -0,0 +1 @@
+{.id = 123,.name = "test_config",.valid = true,}
```
spytheman: that will match on master too

Image

Imho change it to the full line:

```suggestion
static main__Config _const_main__conf = ((main__Config){.id = 123,.name = "test_config",.valid = true,});
```

wenxuanjun: It seems the CI environment runs with `-parallel-cc` which causes `g.static_non_parallel` to be empty, omitting the `static` keyword

spytheman (approved):
Excellent work.
Thank you @wenxuanjun 🙇🏻 .

PR #25893, opened 2025-12-04
os: enhance os.cp() to mimic Python shutil.copy2()

quaesitor-scientiam:
Mimics Pythons' `shutil.copy2()` function to support copying a file to a target directory. Also surfaced optional parameter to control overwriter or failure on copy to existing file in the destination directory

Enhanced the existing `os.cp()` test to test the copy to directory name

### DIFF
--- a/vlib/v/markused/walker.v
+++ b/vlib/v/markused/walker.v
@@ -454,7 +454,9 @@ fn (mut w Walker) expr(node_ ast.Expr) {
 		}
 		ast.CallExpr {
 			w.call_expr(mut node)
-			if node.name == 'json.decode' {
+			if node.is_fn_a_const {
+				w.mark_const_as_used(node.name)
+			} else if node.name == 'json.decode' {
 				w.mark_by_type((node.args[0].expr as ast.TypeNode).typ)
 			} else if node.name == 'json.encode' && node.args[0].typ != 0 {
 				sym := w.table.final_sym(node.args[0].typ)
--- a/vlib/v/tests/consts/const_markused_test.v
+++ b/vlib/v/tests/consts/const_markused_test.v
@@ -1,0 +1,6 @@
+import my_utils
+
+fn test_main() {
+	v := f64(1.0 / 8)
+	assert my_utils.my_round(v, 2) == '0.13'
+}
--- a/vlib/v/tests/consts/my_utils/my_utils.v
+++ b/vlib/v/tests/consts/my_utils/my_utils.v
@@ -1,0 +1,5 @@
+module my_utils
+
+import strconv
+
+pub const my_round = strconv.f64_to_str_lnd1
--- a/vlib/v/tests/generics/generics_str_intp_test.v
+++ b/vlib/v/tests/generics/generics_str_intp_test.v
@@ -1,0 +1,27 @@
+module main
+
+import strings
+
+pub struct MyStruct[T] {
+pub mut:
+	result T
+}
+
+pub fn (it MyStruct[T]) indent_str[T]() string {
+	mut res := strings.new_builder(32)
+	res.write_string('${it.result}')
+	return res.str()
+}
+
+fn test_generics_str_intp() {
+	x := MyStruct[int]{
+		result: 100
+	}
+
+	y := MyStruct[string]{
+		result: 'hello'
+	}
+
+	assert x.indent_str() == '100'
+	assert y.indent_str() == 'hello'
+}
--- a/vlib/v/transformer/transformer.v
+++ b/vlib/v/transformer/transformer.v
@@ -1268,6 +1268,11 @@ pub fn (mut t Transformer) simplify_nested_interpolation_in_sb(mut onode ast.Stm
 		return false
 	}
 	original := nexpr.args[0].expr as ast.StringInterLiteral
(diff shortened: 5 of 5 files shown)


vlib/os/os.v
```diff
@@ -82,7 +82,7 @@ pub fn cp_all(src string, dst string, overwrite bool) ! {
 				return error('Destination file path already exist')
 			}
 		}
-		cp(source_path, adjusted_path)!
+		cp(source_path, adjusted_path, fail_if_exists: false)!
```
spytheman:
```suggestion
		cp(source_path, adjusted_path)!
```

spytheman: the default should not change, so this call should stay as it is

vlib/os/os.c.v
```diff
@@ -231,12 +231,22 @@ pub fn rename(src string, dst string) ! {
 	}
 }
 
-// cp copies files or folders from `src` to `dst`.
-pub fn cp(src string, dst string) ! {
+@[params]
+pub struct FailIfExists {
```
spytheman:
```suggestion
pub struct CopyParams {
```

vlib/os/os_js.js.v
```diff
@@ -109,7 +109,12 @@ pub fn rm(path string) ! {
 	}
 }
 
-pub fn cp(src string, dst string) ! {
+@[params]
+pub struct FailIfExists {
```
spytheman:
```suggestion
pub struct CopyParams {
```

vlib/os/os.c.v
```diff
@@ -231,12 +231,22 @@ pub fn rename(src string, dst string) ! {
...
+	fail_if_exists bool
+}
+
+// cp Copies the file src to the file or directory dst. If dst specifies a directory, the file will be copied into dst
+//    using the base filename from src. If dst specifies a file that already exists, it will be replaced by
+//    default. Can be overridden to fail by setting fail_if_exists: true
```
spytheman:
```suggestion
// cp copies the file src to the file or directory dst. If dst specifies a directory, the file will be copied into dst
// using the base filename from src. If dst specifies a file that already exists, it will be replaced by
// default. Can be overridden to fail by setting fail_if_exists: true
```

spytheman: To me, it is not clear what will happen in the `cp('folder1', 'folder2')` case.

vlib/os/os.v
quaesitor-scientiam: Ok

vlib/os/os_test.c.v
```diff
@@ -330,11 +330,22 @@ fn test_cp() {
 	new_file_name := 'cp_new_example.txt'
 	os.write_file(old_file_name, 'Test data 1 2 3, V is awesome #$%^[]!~⭐') or { panic(err) }
 	os.cp(old_file_name, new_file_name) or { panic(err) }
-	old_file := os.read_file(old_file_name) or { panic(err) }
-	new_file := os.read_file(new_file_name) or { panic(err) }
+	mut old_file := os.read_file(old_file_name) or { panic(err) }
+	mut new_file := os.read_file(new_file_name) or { panic(err) }
```
spytheman: Please do not use `mut` here.

If you add a new test, use new variable names - the tests are better if they are more concrete and have no logic in them - that way, when they fail, you know right away where the problem is, and do not have to spend time to first debug/track the logic and mutability in the test itself.

vlib/os/os.c.v
quaesitor-scientiam: The trailing struct literal boolean is initialized to zero. Making the parameter `fail_if_exists` false.  Which was the previously defaulted value. False value allows the destination file to be overridden. So no change in functionality.

When True the OS wile return an error.

spytheman: That does not answer the question, because there is another change too, not just the parametrized one:
```
		if is_dir(dst) {
			w_dst = join_path_single(w_dst, file_name(src))
		}
```
in case the target is a folder, this will change the destination, compared to master, and imho it is good to explicitly describe what it will do in the documentation comment too, just like there is an explicit description about what will happen if the source is a file, there can be an explicit description if there will be a `target/src/` with the files from src/ in it at the end, or just `target/` with the files from `src/` in it at the end.

vlib/os/os_test.c.v
quaesitor-scientiam: Made a separate test will that work?

vlib/os/os.c.v
quaesitor-scientiam: > To me, it is not clear what will happen in the `cp('folder1', 'folder2')` case.

Will fail the source cannot be a folder should probably be a check to verify its a file

vlib/os/os_test.c.v
quaesitor-scientiam: I added separate tests

quaesitor-scientiam (commented):
Thanks for the cleanup.  I was wondering how to assert the fail_if_exists test that was very useful.

spytheman:
There is no need to merge master after every commit to it, usually that just wastes CI time.

spytheman (approved):
Thank you @quaesitor-scientiam 🙇🏻 .

Issue #25896, opened 2025-12-04
cgen error: cannot convert 'struct string' to 'long'


kbkpbot:
### Describe the bug

generic T function may fail with `strings.write_string` but not `strings.writeln`.

found by PR #25857

### Reproduction Steps

g.v
```v
module main

import strings

pub struct MyStruct[T] {
pub mut:
        result T
        id     string
}

fn main() {
        x := MyStruct[int]{}
        y := MyStruct[string]{}

        dump(x.indent_str())
        dump(y.indent_str())
}

pub fn (it MyStruct[T]) indent_str[T]() string {
        mut res := strings.new_builder(128)
        res.write_string('${it.result}')
        // following is OK
        // res.writeln('${it.result}')
        return res.str()
}
```

```
v g.v
```

### Expected Behavior

compile ok

### Current Behavior

```
$ v g.v
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/g.01KBMWV9MNPR8A6ZW47790YW53.tmp.c:5033: error: cannot convert 'struct string' to 'long'
=================================================================
Try passing `-g` when compiling, to see a .v file:line information, that correlates more with the C error.
(Alternatively, pass `-show-c-output`, to print the full C error message).
builder error:
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```

### Possible Solution

use `writeln` instead of `write_string`, it will pass compile

### V version

V 0.4.12 dafa6fd0aaaa329a72c724297e895d6097303265

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 dafa6fd0aaaa329a72c724297e895d6097303265
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |6.54GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-12-04 14:35:47
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/autostr
|                    |
[...]

PR #25897, opened 2025-12-05
transformer: disable generic str_intp opt

kbkpbot:
Fix issue #25896 

`transformer` can't handle generic string intp opt right now, so skip it.


spytheman (approved):
Good find. Thank you @kbkpbot 🙇🏻 .

PR #25904, opened 2025-12-05
io.util: use correct TempDirOptions in temp_dir()

tankf33der:
I suspect this is an unfortunate typo.
Please confirm and merge.


JalonSolov:
Since the 2 structs are identical (including field names & types, `@[params]` tag, etc.), there is no functional change, but it is better in case we decide a difference is required in the future.

PR #25905, opened 2025-12-05
example (Vanilla HTTP Server): fix and improve

enghitalo:
```sh
Running 1m test @ http://127.0.0.1:3001/
  16 threads and 512 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.21ms    1.49ms  27.24ms   84.63%
    Req/Sec    31.27k     3.87k   54.16k    76.33%
  29888678 requests in 1.00m, 2.62GB read
Requests/sec: 497314.66
Transfer/sec:     44.58MB
```


spytheman:
Thank you @enghitalo .

PR #25914, opened 2025-12-06
cgen: fix fntype unwrap on ifguard

### LINKED ISSUE #25911: Call to unwrapped `?&Fn` fails in if guard and iterator loops
### Describe the bug

Calling a function pointer obtained from a `?&Fn` (optional pointer to function) generates invalid C code when unwrapped via if guards or iterator loops `for x in iter`.

The C generator creates a double pointer `void (**)(void)` but attempts a direct call `p()`, causing a function pointer expected error. But unwrapping via `or { return }` works correctly.

### Reproduction Steps

The if guard version compared to `or { return }` version:
```v
type MyFn = fn()

fn hello() {}

fn get_ptr() ?&MyFn {
	return &hello
}

fn main() {
	if p := get_ptr() {
		p() // error: function pointer expected
	}

	p2 := get_ptr() or { return } // works
	p2() // OK
}
```

The iterator fails either:
```v
type MyFn = fn()

struct Iter {
	f MyFn = fn(){}
mut:
	done bool
}

fn (mut it Iter) next() ?&MyFn {
	if it.done { return none }
	it.done = true
	return &it.f
}

fn main() {
	for p in Iter{} {
		p() // error: function pointer expected
	}
}
```

### Expected Behavior

The if guard and iterator versions should compile and run.

### Current Behavior

C compilation fails: `error: function pointer expected`

### Possible Solution

_No response_

_No response_

### V version

V 0.4.12 ecd0018

V 0.4.12 ecd0018
> [!NOTE]
>

felipensp:
Fix #25911

### DIFF
--- a/vlib/v/tests/options/option_fn_guard_test.v
+++ b/vlib/v/tests/options/option_fn_guard_test.v
@@ -1,6 +1,8 @@
 type MyFn = fn ()
 
-fn hello() {}
+fn hello() {
+	println(@FN)
+}
 
 fn get_ptr() ?&MyFn {
 	return &hello
@@ -8,10 +10,12 @@ fn get_ptr() ?&MyFn {
 
 fn test_main() {
 	if p := get_ptr() {
+		assert voidptr(p) == voidptr(hello)
 		dump(p)
 		p()
 	}
 
 	p2 := get_ptr() or { return }
 	p2()
+	assert voidptr(p2) == voidptr(hello)
 }


vlib/v/tests/options/option_fn_guard_test.v
```diff
@@ -0,0 +1,17 @@
...
+
+fn test_main() {
+	if p := get_ptr() {
+		dump(p)
+		p()
+	}
+
+	p2 := get_ptr() or { return }
+	p2()
+}
```
spytheman, line 21:
```suggestion
	assert voidptr(p2) == voidptr(hello)
}
```

PR #25915, opened 2025-12-06
cgen: fix cast type changed before last stmt in `stmts_with_tmp_var`

Krchi:
like nested match clear `expected_cast_type`


spytheman (approved):
Excellent work.
Thank you @Krchi 🙇🏻‍♂️ .

Issue #25917, opened 2025-12-07
Add indent() method to string type


kbkpbot:
### Describe the feature

Add a `indent()` method to the V string type that formats strings containing curly braces `{`, `}` with proper indentation. The method should automatically:

- Insert line breaks after opening braces `{` and before closing braces `}`
- Apply configurable indentation levels
- Maintain proper nesting for multiple brace levels
- Preserve content outside braces

May related to PR #25857

### Use Case

Currently, V's string type lacks built-in support for brace-based formatting. This method would be particularly useful for:

- Generating properly formatted code
- Creating readable JSON/XML-like outputs
- Formatting configuration files
- Pretty-printing nested data structures
- Use for `auto_str()`, then we will not need `.indent_str()` anymore

### Proposed Solution

```v
// As a method on the string type
fn (s string) indent(indent_spaces int) string
```

or 

```v
fn (s string) indent(indent int, options ...FormatOption) string
```

### Other Information

a `dump` can be called as:
```v
dump(output_string.indent(4))
```

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.12 ff6413acd3c24c8d28170ed0b3494efaf5ea637b.dd066bb

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 ff6413acd3c24c8d28170ed0b3494efaf5ea637b.dd066bb
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |12.37GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-12-07 00:07:42
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /media/HD/github/kbkpbot/v
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.46-151-gdd066bbd-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
[...]

PR #25921, opened 2025-12-08
strings: add indent()

kbkpbot:
Feature request by #25917

### DIFF
--- a/vlib/strings/builder.c.v
+++ b/vlib/strings/builder.c.v
@@ -366,12 +366,13 @@ enum IndentState {
 // 6. Maintains existing line breaks from the input.
 //
 // Example:
-//   import strings
-//   input := 'User{name:"John" settings:{theme:"dark"}}'
-//   mut b := strings.new_builder(64)
-//   b.indent(input, indent_count: 2)
-//   println(b.str()) // Formatted output: 'User{\n  name:"John" settings:{\n    theme:"dark"\n  }\n}'
-//
+// ```v
+// import strings
+// input := 'User{name:"John" settings:{theme:"dark"}}'
+// mut b := strings.new_builder(64)
+// b.indent(input, indent_count: 2)
+// println(b.str()) // Formatted output: 'User{\n  name:"John" settings:{\n    theme:"dark"\n  }\n}'
+// ```
 @[direct_array_access]
 pub fn (mut b Builder) indent(s string, param IndentParam) {
 	if s.len == 0 {


vlib/builtin/string.v
```diff
@@ -3049,3 +3049,112 @@ pub fn (mut ri RunesIterator) next() ?rune {
 	}
 	return rune(impl_utf8_to_utf32(start, len))
 }
+
+@[params]
+pub struct IndentParam {
+pub mut:
+	open_char    rune = `{` // Character that triggers increased indentation
+	close_char   rune = `}` // Character that triggers decreased indentation
```
JalonSolov: Would make more sense (to me, at least) to name these differently...

```suggestion
	block_start	rune = `{` // Character that starts a new block (+ indent)
	block_end	rune = `}` // Character that ends a new block (- indent)
```

vlib/builtin/string_test.v
```diff
@@ -1702,3 +1702,157 @@ fn test_hex() {
 		assert c.ascii_str().hex() == [c].hex()
 	}
 }
+
+struct IndentTest {
+	input  string
+	output string
+}
+
+// it seems vfmt has some bug
```
spytheman: That is not helpful :-|, and can confuse people in the future.
Either describe the bug, or remove the comment.

vlib/builtin/string.v
```diff
@@ -3049,3 +3049,150 @@ pub fn (mut ri RunesIterator) next() ?rune {
...
+				}
+			}
+			.in_string {
+				// Inside a string literal: preserve all characters as-is
+				sb.write_rune(c)
+
+				// Check for string termination
+				// The character must match the opening quote and not be escaped
+				if c == string_char {
+					if i == 0 || s[i - 1] != `\\` {
```
spytheman: How can the `i==0` become true?
The function starts with `mut state := IndentState.normal` and `mut i := 0`, and can only transition to .in_string in the first match branch, where also `i++`.

vlib/builtin/string.v
```diff
@@ -3049,3 +3049,150 @@ pub fn (mut ri RunesIterator) next() ?rune {
...
+		match state {
+			// Normal state: process characters outside of string literals
+			.normal {
+				match c {
+					`"`, `'` { // make it display correctly in editor "
+						state = .in_string
+						string_char = c
+						// Add indentation if at the start of a line
+						if at_line_start {
+							sb.write_string(param.indent_char.repeat(indent_level * param.indent_count))
```
spytheman: That will allocate a temporary string needlessly.
Use this instead:
```v
for _ in 0..indent_level * param.indent_count { 
    sb.write_rune(param.indent_char) 
}
```

vlib/builtin/string.v
```diff
@@ -3049,3 +3049,150 @@ pub fn (mut ri RunesIterator) next() ?rune {
...
+						}
+						// Write the opening quote
+						sb.write_rune(c)
+						i++
+					}
+					param.block_start {
+						// Start of a new block
+						// Add indentation if at the start of a line
+						if at_line_start {
+							sb.write_string(param.indent_char.repeat(indent_level * param.indent_count))
```
spytheman: Same - use a for loop, or if you prefer, extract a strings builder method `.write_repeated_rune(r rune, count int)` to do it, and use that here.

vlib/builtin/string.v
```diff
@@ -3049,3 +3049,150 @@ pub fn (mut ri RunesIterator) next() ?rune {
...
+							indent_level--
+						}
+
+						// If not at the start of a line, add a newline
+						if !at_line_start {
+							sb.write_rune(`\n`)
+						}
+
+						// Add indentation for the block end
+						sb.write_string(param.indent_char.repeat(indent_level * param.indent_count))
```
spytheman: same - use a loop, instead of creating an entirely new string with repeats in it, only to write it right away in the builder instance

vlib/builtin/string.v
```diff
@@ -3049,3 +3049,150 @@ pub fn (mut ri RunesIterator) next() ?rune {
...
+						if c == `\n` {
+							at_line_start = true
+						}
+						i++
+					}
+					else {
+						// Any other character
+						// Add indentation if at the start of a line
+						if at_line_start {
+							sb.write_string(param.indent_char.repeat(indent_level * param.indent_count))
```
spytheman: use a loop

vlib/builtin/string_test.v
```diff
@@ -1702,3 +1702,157 @@ fn test_hex() {
...
+	},
+]
+
+// vfmt on
+
+fn test_indent() {
+	for t in indent_test_data {
+		assert t.input.indent() == t.output
+	}
+}
```
spytheman: imho also add tests for each of the parameters

spytheman: (one per parameter is fine, as long as they are all mentioned at least once, to guarantee that all of them do something, and that user code can rely on their names not changing in the future)

vlib/builtin/string.v
```diff
@@ -3049,3 +3049,150 @@ pub fn (mut ri RunesIterator) next() ?rune {
...
 }
+
+// IndentParam holds configuration parameters for the indent() function
+@[params]
+pub struct IndentParam {
+pub mut:
+	block_start  rune = `{` // Character that starts a new block (+ indent)
+	block_end    rune = `}` // Character that ends a new block (- indent)
+	indent_char  rune = ` ` // Character used for indentation (space or tab)
+	indent_count int  = 4   // Number of indent_char per indentation level
```
spytheman: You may also want a starting level too, if the intent is to use this as a substitute for the cgen generated methods.

The reason, is that you can have nested structs, and if you support a starting level, then you can reuse the same code for different indentation levels, without having to re-interpolate it.

Another nice to have, is a version, that accepts a mutable string builder to be passed in (instead of created inside and stringified at the end).

Then you can have a string builder created just once per high level .str() operation of say a struct, then passed to each field's generated str implementation with starting_level + 1, then stringified just once.

spytheman: (on the other hand, that may be just over-complicating it)

vlib/builtin/string.v
```diff
@@ -3049,3 +3049,150 @@ pub fn (mut ri RunesIterator) next() ?rune {
...
+// 5. Adds newlines after `block_start` and before `block_end` (except for empty blocks)
+// 6. Maintains line breaks in the original input
+// Example: assert 'User{name:"John" settings:{theme:"dark" language:"en"}}'.indent() == 'User{\n    name:"John" settings:{\n        theme:"dark" language:"en"\n    }\n}'
+@[direct_array_access]
+pub fn (s string) indent(param IndentParam) string {
+	if s.len == 0 {
+		return ''
+	}
+
+	mut sb := strings.new_builder(s.len * 2)
```
spytheman: why `* 2` instead of just s.len, or s.len + 10 ?

spytheman: what will be the most common input to be indented in your estimation?

spytheman: imho the reasoning should be put in a comment after the line for clarity

vlib/builtin/string.v
kbkpbot: > You may also want a starting level too, if the intent is to use this as a substitute for the cgen generated methods.
> 
> The reason, is that you can have nested structs, and if you support a starting level, then you can reuse the same code for different indentation levels, without having to re-interpolate it.
> 
> Another nice to have, is a version, that accepts a mutable string builder to be passed in (instead of created inside and stringified at the end).
> 
> Then you can have a string builder created just once per high level .str() operation of say a struct, then passed to each field's generated str implementation with starting_level + 1, then stringified just once.

Is it a better choice that put this function in `strings/builder.c.v` instead of `builtin/string.v`?

spytheman: > Is it a better choice that put this function in strings/builder.c.v instead of builtin/string.v?

It could be, yes - `builtin` has many restrictions that `strings` does not.

vlib/builtin/string.v
```diff
@@ -3090,7 +3090,7 @@ pub fn (s string) indent(param IndentParam) string {
 	// 1. Prevents frequent reallocations for unformatted inputs
 	// 2. Avoids excessive memory waste for well-formatted inputs
 	// 3. Handles most real-world formatting scenarios efficiently
-	mut sb := strings.new_builder(int(s.len * 1.5))
+	mut sb := strings.new_builder(int(f32(s.len) * 1.5))
```
spytheman: `s.len * 3 / 2` achieves the same, but without using floats

spytheman: I still do not understand why not just use `s.len` or `* 2` though. The builder will do the reallocation if needed internally. Why insist on doing it beforehand, and with that specific coefficient? Was that number the result of experiments or something else? The explanations are ok (if a bit redundant, for someone that understands the purpose of pre-allocating), but they do not explain how it was choosen, and why another coefficient was not a reasonable choice instead.

spytheman: i.e. I do not want an AI essay, just a short comment about why the simplest `s.len` was not reasonable.

vlib/strings/builder.c.v
```diff
@@ -334,3 +334,157 @@ pub fn (mut b Builder) write_repeated_rune(r rune, count int) {
...
+		return
+	}
+
+	mut state := IndentState.normal
+	mut indent_level := param.starting_level
+	mut string_char := `\0`
+	mut at_line_start := true
+	mut i := 0
+
+	for i < s.len {
```
spytheman:
```suggestion
	for i := 0; i<s.len; i++ {
```
then all the `i++` lines below can be removed, and the `i += 2` can become `i++`

PR #25925, opened 2025-12-09
math: fix asinh(), add tests

tankf33der:
The same type of pseudo typo as in #25923.
Only this patch allows running multi-million tests against `mpfr` and `mpmath` libraries.
Tests cannot pass on `master`.


spytheman (approved):
Excellent 👓 🕶️ .
Thank you very much @tankf33der 🙇🏻 .

PR #25933, opened 2025-12-10
strings: add write_repeated_rune()

kbkpbot:
// write_repeated_rune appends multiple copies of the same rune to the accumulated buffer
```v
pub fn (mut b Builder) write_repeated_rune(r rune, count int)
```


spytheman (approved):
Excellent work.

I would have just used a loop with a .write_rune(r) call in it, but the more verbose version can be more performant for more runes, since it pays all the overhead just once.

PR #25857, opened 2025-11-29
builder,checker,ast: move auto str() before backend gen

kbkpbot:
Feature request by #25826

This PR aims to decouple the `auto_str()` function from individual backend generators and move it to the frontend. This way, each backend can remove its own `auto_str()` implementation. 
The general approach is to introduce a new phase, `gen_auto_fn()` in the `builder`—after the `checker` and before code generation. This phase leverages the results from the `checker` and `markused` to identify which types require the `auto_str()` function. It then manually creates these virtual functions, processes them using `parser.parse_text()`, and performs checks via `check()`. The output is subsequently passed to the respective backends for code generation.

Currently, only the generation of `auto_str()` for structs is supported. Future extensions may include support for other `str()` functions, with the ultimate goal of fully migrating this functionality to the frontend.

PR #25938, opened 2025-12-11
markused: add new auto_str() used funcs for bootstrap

kbkpbot:
As PR #25857, new `auto_str()` used some functions in `strings`, so add them to make bootstrap working.


spytheman (approved):
Excellent.
Thank you @kbkpbot .

PR #25956, opened 2025-12-13
math.easing: fix bounce family, add tests

tankf33der:
Whole `bounce` family is broken.

Two bugs - an extra `if` in `in_out_bounce()` and completely incorrect `out_bounce()` due to `n1` coming into play in the wrong order. The entire bounce family doesn't work correctly because the other two functions use the broken `out_bounce()`.

Last commit in chain, now the entire module will work correctly.


vlib/math/easing/easing.v
```diff
@@ -244,26 +244,19 @@ pub fn in_bounce(x f64) f64 {
...
+	} else if x < 2.0 / d1 {
+		n1 * pow(x - 1.5 / d1, 2.0) + 0.75
+	} else if x < 2.5 / d1 {
+		n1 * pow(x - 2.25 / d1, 2.0) + 0.9375
 	} else {
-		pow((x - 2.625 / d1), 2.0) + 0.984375
-	})
+		n1 * pow(x - 2.625 / d1, 2.0) + 0.984375
+	}
 }
```
spytheman, line 257: @tankf33der thank you a lot for fixing it - my mistake, that I could not discover before, was in extracting the multiplication by `n1`, because I was looking at just the start of the expressions, while overlooking *the addition* of the different coefficients after the multiplication in each branch.

I do think that `xd1 := x * d1` optimization is still valuable though, i.e. having something like:

```suggestion
pub fn out_bounce(x f64) f64 {
	xd1 := x * d1
	return if xd1 < 1.0 {
		n1 * x * x
	} else if xd1 < 2.0 {
		n1 * pow(x - 1.5 / d1, 2.0) + 0.75
	} else if xd1 < 2.5 {
		n1 * pow(x - 2.25 / d1, 2.0) + 0.9375
	} else {
		n1 * pow(x - 2.625 / d1, 2.0) + 0.984375
	}
}
```

does 1 multiplication and later just comparisons, until the proper branch is found, instead of several divisions and comparisons, until the proper branch is found.

I just tested with it, and both `v test vlib/math/easing/` and the comparison with the .out file work with it.

What do you think?
(edited to fix the spaces->tabs)

spytheman:
About this CI failure (on macos): https://github.com/vlang/v/actions/runs/20190893269/job/57968163915?pr=25956 
image
... we can use `:8.2f` in the `vlib/v/slow_tests/inout/math_easing_tables.vv` file, and regenerate the `.out` file again 🤔 for now.

(i.e. I do not think that the rounding error is significant enough for further investigation for the easing functions - they are used for visual effects mostly, where the 3rd digit after the point will not be perceivable in most cases, and I prefer to have a stable CI, that would not trigger false positives for slight changes of the compiler/math functions in libc, for unrelated functionality)

vlib/math/easing/easing.v
tankf33der: My main idea was to make the code as close to the original as possible so that someone could easily audit (!) it in the future. I've encountered the situation more than once where code is ported from somewhere, for example `golang`, and they change too much in it, the result is not an audit, but a re-porting in your head with new bugs.

Lets save your optimization here!

tankf33der:
i've pushed optimization and switch to `8.2f`.

tankf33der:
@spytheman - `v run examples/gg/easing_animation.v` works better, right?

spytheman:
> @spytheman - `v run examples/gg/easing_animation.v` works better, right?

Absolutely - the animations there look very smooth and nice, without any weird sudden jumps 🥳 .

tankf33der:
`macos` failed again, let me bootstrap vlang on wife's mac and play around, it will take some time. will be back asap.

spytheman:
It worked on macos in the opposite direction (:8.4f), which is weird to me 🤔, but it also happens to work on linux too:
image

tankf33der:
> It worked on macos in the opposite direction (:8.4f), which is weird to me 🤔, but it also happens to work on linux too: image

FYI, `8.2f` works with `gcc-15` from brew.

spytheman:
> FYI, 8.2f works with gcc-15 from brew.

I think the diff is due to some combination of the used libc and default optimization levels for the C compiler (which is the Apple's version of clang on macos by default).

PR #25957, opened 2025-12-13
ci: re-enable job with gcc on OpenBSD 7.8

lcheylus:
- Fix installation of gcc version 11 on OpenBSD 7.8 used in CI (`pkg_add gcc%11`)
- gcc binary = egcc on OpenBSD (`export VFLAGS='-cc egcc'`)

**Tests OK with tcc/clang/gcc** for OpenBSD CI using relase 7.8 => see this run in my personal repo https://github.com/lcheylus/v/actions/runs/20190921203

### DIFF
--- a/.github/workflows/openbsd_ci.yml
+++ b/.github/workflows/openbsd_ci.yml
@@ -74,3 +74,29 @@ jobs:
             sudo ./v symlink
             export VFLAGS='-cc clang'
             ./v run ci/openbsd_ci.vsh all
+
+  gcc-openbsd:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v6
+      - name: Tests on OpenBSD with gcc
+        id: tests-openbsd-clang
+        uses: cross-platform-actions/action@v0.30.0
+        with:
+          operating_system: openbsd
+          version: '7.8'
+          memory: 4G
+          shell: sh
+          sync_files: runner-to-vm
+          run: |
+            sudo pkg_add git sqlite3 gmake boehm-gc libiconv gcc%11
+            # Mandatory: hostname not set in VM => some tests fail
+            sudo hostname -s openbsd-ci
+            echo "### OS infos"
+            uname -a
+            git config --global --add safe.directory .
+            gmake
+            sudo ./v symlink
+            export VFLAGS='-cc egcc'
+            ./v run ci/openbsd_ci.vsh all


spytheman:
Sorry @lcheylus, but I am against merging this - the decision of OpenBSD's package maintainers to not have just `pkg_add install gcc` work, without a specific version, and to later not have just `gcc`, but instead `egcc` is actively stupid (as it achieves nothing useful, but harms others), and managed to waste several hours already (for me, and for several other people that also had opened PRs with failing CIs here).

I would like to avoid this happening in the future (for OpenBSD 7.9), if the OpenBSD gcc maintainers do not intend to fix it (which is likely, since it seems to be like that for months/years).

If they can be convinced, that it is better to have the `gcc` package, behave the same as `git`, `gmake` etc, i.e. have an executable named `gcc` by default, and having just `pkg_add install gcc` install the latest supported version, instead of relying on specifying an exact one each time, then I am willing to reconsider this decision.

lcheylus:
@spytheman I'm sad to read this. Your decision is based on a lack on knowledge of OpenBSD OS and the history of its development: 

- for most of the 2010s, the OpenBSD base system has been stuck with GCC 4.2.1, for licensing reasons ; FreeBSD did the same, and Mac OS X as well.
- Clang is the default compiler on amd64, arm64, armv7, i386, loongson, macppc, octeon, powerpc64, and riscv64.
GCC 4.2.1 is still the default compiler on alpha, hppa, landisk, and sparc64.
- More recent versions of GCC are available via OpenBSD ports: **gcc version 8 and 11 for amd64 architecture in 7.8 release** and it's why **gcc binary is renamed egcc** for these versions, avoiding confusion with the old version GCC 4 installed by the base system (binary = `/usr/bin/gcc`).
- More recent GCC version 15 will be available in future OpenBSD 7.9 release.

That's why to install a specific version of gcc from ports, `pkg_add install gcc` does not work => need to select the most recent version 11 available in ports via `pkg_add install gcc%11`.

If you use `pkg_add install gcc` command, the choice for version is ambiguous:
```sh
$ doas pkg_add -n gcc
quirks-7.163 signed on 2025-12-10T15:39:38Z
Ambiguous: choose package for gcc
a       0: <None>
        1: gcc-8.4.0p28
        2: gcc-11.2.0p19
```

For a detailed explanation about "C toolchains in OpenBSD", see https://www.cambus.net/the-state-of-toolchains-in-openbsd/

spytheman:
I saw all that unfortunately for myself, while researching during the hours I lost today trying to fix the CI. 

I find it personally both tragic, ridiculous and absurd, that all of FreeBSD, Linux and macOS moved on, and now in 2025, you can install `gcc` on all of them (even on Windows), and get an executable that you can invoke by typing `gcc`, while OpenBSD seems to be stuck in a situation that requires it to name the modern version `egcc`, due to causes, that are now decades in the past.

As tragic as it is for them, I do not want to relearn all that again in a few months. I prefer to invest my time and energy in more productive endeavors where possible, so until the gcc package maintainers for OpenBSD decide to move on, and fix their broken naming scheme (which again, they seem to follow for everything else), there will be no gcc job for OpenBSD on this CI, but only clang and tcc jobs.

PR #25963, opened 2025-12-14
cgen: correct T{} init for []Type aliases

### LINKED ISSUE #25962: Invalid generated C code when initializing aliased type
### Describe the bug

This is a weird one. I ran into a case where the compiler generates invalid C code. I am creating a type system for a toy interpreter and I have a module that is defining some value types using type aliases that all implement a simple `Value` interface.

### Reproduction Steps

I distilled it down into a simple module with a test:
```
module value

pub interface Value {
	str() string
}

pub type List = []Value
pub fn (x List) str() string { return x.str() }
```

test:
```
module value

fn test_list() {
	mut list := make_list()!
	assert (list as List).len == 0
}

fn make_list() !Value {
	mut list := List{}
	return list
}
```

### Expected Behavior

Test should pass, `list` should be initialized with zero length

### Current Behavior

This results in a compilation error in the C compiler:

```
cc: C:/Users/Me/AppData/Local/Temp/v_0/tsession_2298_01KCD3T3RC207TSYGBGQ72R8R0/base_test.exe.01KCD3T508YKSKB84NC4K8XYJM.tmp.c:7993: error: ';' expected (got "}")
```

### Possible Solution

Seems like the `HEAP` macro(?) is misbehaving here?

The generated C code in question looks like this:

```
VV_LOC _result_value__Value value__make_list(void) {
	value__List *list = HEAP(value__List, ({0})); // <-- CRASHES HERE
	_result_value__Value _t1 = {0};
	builtin___result_ok(&(value__Value[]) { I_value__List_to_Interface_value__Value(&(*(list))) }, (_result*)(&_t1), sizeof(value__Value));
	return _t1;
}
```

In order to repro this I have to wrap the `List` initialization in a separate function that returns the `Value` interface. If I return `List` directly, as below, the code compiles and works.
```
module value

fn test_list() {
	mut list := make_list()
	assert list.len == 0
}

fn make_list() List {
	mut list := List{}
	return list
}
```

### V version

0.4.12

```
|V full version      |V 0.4.12 55436caa9ae9c179639a4b238fc358ee7d768645.c15523b
|OS                  |windows, Microsoft Windows 11 Pro 26200 64-bit
|Processor           |28 cpus, 64bit, little endian, 13th Gen Intel(R) Core(TM) i7-13850HX
|Memory              |15.31GB/31.61GB
|                    |
|V executable        |C:\Program Files\v_windows\v\v.exe
|V last modified time|2025-12-14 00:22:30
|                    |
|V home dir          |contains spaces, value: C:\Program Files\v_windows\v
|VMODULES            |contains spaces, value: C:\Users\Casey Duncan\.vmodules
|VTMP                |contains spaces, value: C:\Users\Casey Duncan\AppData\Local\Temp\v_0
|Current working dir |contains spaces, value: C:\Users\Casey Duncan\projects\vebol
|                    |
|Git version         |git version 2.52.0.windows.1
|V git status        |weekly.2025.50-36-gc15523b2
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |N/A
|clang version       |N/A
|msvc version        |N/A
|tcc git status      |N/A
|emcc version        |N/A
|glibc version       |ldd (cygwin) 3.6.5
```
> [!NOTE]
>

dy-tea:
Fixes #25962.

This fixes the cgen portion of this. The test kind of confuses me though, `make_list` should return `List` or `[]Value` but right now it returns `!Value` without issue. Secondly the `str()` method here calls itself, although that is less of a problem.

### DIFF
--- a/vlib/v/tests/aliases/modules/value/value_test.v
+++ b/vlib/v/tests/aliases/modules/value/value_test.v
@@ -1,0 +1,11 @@
+module value
+
+fn test_list() {
+	mut list := make_list()!
+	assert (list as List).len == 0
+}
+
+fn make_list() !Value {
+	mut list := List{}
+	return list
+}
--- a/vlib/v/gen/c/struct.v
+++ b/vlib/v/gen/c/struct.v
@@ -52,7 +52,7 @@ fn (mut g Gen) struct_init(node ast.StructInit) {
 			g.write(g.type_default_sumtype(unwrapped_typ, sym))
 		}
 		return
-	} else if sym.kind == .map {
+	} else if sym.kind == .map || (sym.kind == .array && node.init_fields.len == 0) {
 		g.write(g.type_default(unwrapped_typ))
 		return
 	}
--- a/vlib/v/tests/aliases/modules/value/value.v
+++ b/vlib/v/tests/aliases/modules/value/value.v
@@ -1,0 +1,11 @@
+module value
+
+pub interface Value {
+	str() string
+}
+
+pub type List = []Value
+
+pub fn (x List) str() string {
+	return x.str()
+}


caseman:
Thanks for this! Yeah the `str()` method is broken as is, but for some reason it won't recognize the built-in `str()` to satisfy the interface. If you omit it, you get:

```
error: `value.List` doesn't implement method `str` of interface `value.Value`
```

Also if you return `[]Value` it also doesn't satisfy the `Value` interface 😭 which is what I need. What a world.

vlib/v/tests/aliases/modules/value/value_test.v
spytheman, line 1: Does it need to be in a separate file and in a module?
If it does not, please merge both in one single _test.v file, directly under `‎vlib/v/tests/aliases`.

spytheman:
> Also if you return []Value it also doesn't satisfy the Value interface 😭 which is what I need. What a world.

@caseman - you can try: `pub fn (x []Value) str() string { return 'some_string' }`

vlib/v/tests/aliases/modules/value/value_test.v
spytheman: Hm, interesting ... having it in a single file, leads to a checker error:
```v
pub interface Value {
	str() string
}

pub type List = []Value

pub fn (x List) str() string {
	return x.str()
}

fn test_list() {
	mut list := make_list()!
	assert (list as List).len == 0
}

fn make_list() !Value {
	mut list := List{}
	return list
}
```
produces:
```
xx_test.v:18:2: error: `[]Value` doesn't implement method `str` of interface `Value`
   16 | fn make_list() !Value {
   17 |     mut list := List{}
   18 |     return list
      |     ~~~~~~~~~~~
```

... and adding `pub fn (x []Value) str() string { return 'some string' }` leads to:
```
#0 13:21:13 ^ master ~/v>./v -g xx_test.v
   module: builtin
 function: __as_cast()
  message: as cast: cannot cast `[]main.Value` to `main.List`
     file: vlib/builtin/builtin.v:34
vlib/builtin/builtin.c.v:93: at builtin__panic_debug: Backtrace
vlib/builtin/builtin.v:34: by builtin____as_cast
xx_test.v:25: by main__test_list
/tmp/v_1000/xx_test.01KCE9GCEB5K848PVMQQY5MBR0.tmp.c:17287: by main
```

but `xx_test.v` does not have 25 lines, it has only 21 🤔 .

Anyway, that weirdness is not directly related to the PR.

caseman:
Yeah in reality the interface will be more fleshed out. I was using aliases because I want a predictable name for the type as a simple string. hence `List` for `[]Value`. But it is nice that you can treat `[]Value` as a bone-a-fide type 😄

dy-tea:
Yeah the checker error here when I combine the files is strange but I left it like that for now.

PR #25964, opened 2025-12-14
checker,cgen: fix variadic call on sumtype with array of itself

### LINKED ISSUE #25898: array pass to variadic fn C error
### Describe the bug

Code: https://play.vlang.io/p/425ebf8d52

```v
import x.json2

fn foo(a int, args ...json2.Any) {
    println(args)
}

foo(1, [json2.Any(42)])
```

### Reproduction Steps

click run 

### Expected Behavior

compile ok

### Current Behavior

Output:

```
/box/code.v:7: error: cannot convert 'struct x__json2__Any' to 'struct array'
builder error: 
==================
Exited with error status 1
```

### Possible Solution

_No response_

_No response_

### V version

V 0.4.12 dafa6fd0aaaa329a72c724297e895d6097303265.b01273e

```
|V full version      |V 0.4.12 dafa6fd0aaaa329a72c724297e895d6097303265.b01273e
|Memory              |0.62GB/2.02GB
|                    |
|V last modified time|2025-12-05 04:25:30
|                    |
|                    |
|V git status        |N/A
|.git/config present |true
|                    |

|emcc version        |N/A
```
> [!NOTE]
>

Krchi:
fix #25898

### DIFF
--- a/vlib/v/checker/tests/sumtype_array_to_variadic.vv
+++ b/vlib/v/checker/tests/sumtype_array_to_variadic.vv
@@ -35,5 +35,4 @@ pub fn build_map(index_from int, index_to int) Map {
 fn main() {
 	m := build_map(2, 4)
 	_ = m
-	_ := Map.create(1, [Value(1)])
 }
--- a/vlib/v/tests/fns/call_args_variadic_sumtype_test.v
+++ b/vlib/v/tests/fns/call_args_variadic_sumtype_test.v
@@ -14,26 +14,6 @@ fn print_names(animals ...Animal) {
 	}
 }
 
-type Any = []Any | int
-
-fn print_any(arr ...Any) {
-	if any := arr[0] {
-		match any {
-			int {
-				assert any == 1
-			}
-			[]Any {
-				assert any == [Any(1)]
-			}
-		}
-	}
-	assert true
-}
-
-fn print_int_any(_ int, arr ...Any) {
-	print_any(...arr)
-}
-
 fn test_main() {
 	cat := Cat{
 		name: 'Kitty'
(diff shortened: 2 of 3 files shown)


vlib/v/checker/tests/sumtype_array_to_variadic.vv
```diff
@@ -2,7 +2,7 @@ type Nil = u8
 
 const c_nil = Nil(`\0`)
 
-pub type Value = Nil | int | string | []Value | Map
+pub type Value = Nil | int | string | Map
```
spytheman, line 5: Why does it need to change an existing test?
Is that on purpose, or is it a regression?

vlib/v/tests/fns/call_args_variadic_sumtype_test.v
spytheman, line 1: In general, it is better to have several isolated _test.v files, each focusing on a single aspect/bug fix, than append new unrelated tests to existing ones, which in time makes them harder and harder to understand.

Please, split the new cases into a separate file.

vlib/v/checker/tests/sumtype_array_to_variadic.vv
Krchi: []Value can be treated as Value, so i think it can be passed to Map.create, with `[]Value`, the file could be compiled

spytheman: i.e. it is on purpose and the PR requires it?

Krchi: > i.e. it is on purpose and the PR requires it?

Not just that. `Map.create([]Value{})` is equals to `Map.create(Value([]Value{}))` that has no problem, I think `[]Value` should be removed

spytheman (approved):
Excellent work.
Thank you @Krchi 🙇🏻 .

PR #25967, opened 2025-12-14
transformer: fix `-new-transformer` (ast.ArrayInit) (part 2)

Eliyaan:
The `-new-transformer` flag works for native but not for the c backend.
This is an unfinished attempt to fix it for the C backend, it works better than before but not entirely yet.
I create this PR in case someone wants to try to fix it as I will not have much time (and I am struggling a lot).

The last thing that I tried was applying the transformer to the new node (see `final_expr` in transformer/array.v), but that causes segfaults in the compiler while compiling programs that did not fail before.


dy-tea:
I tried something that seems to maybe work?

vlib/v/transformer/array.v
```diff
@@ -73,13 +80,13 @@ pub fn (mut t Transformer) array_init(mut node ast.ArrayInit) ast.Expr {
 	mut call_expr := ast.CallExpr{
 		name:        fn_name
 		mod:         'builtin'
+		scope:       unsafe { nil }
 		args:        [len_arg, len_arg, sizeof_arg, fixed_array_arg] //, sizeof(voidptr), _MOV((voidptr[${len}]){')
 		return_type: node.typ
 	}
 	// println('call expr')
 	// println(call_expr)
-	final_expr := t.expr(mut call_expr)
```
Eliyaan: Shouldn't the transformer apply to the new expression (and its arguments)?

dy-tea: It probably should but it was crashing. I'll try a few things.

dy-tea: It looks like the transformer might not need to apply to the new expression, it seems to already be in the right state.

Eliyaan: Okay great.
Couldn't it need to be change later if the transformer changes?

dy-tea: Possibly, at the moment it just causes issues though.

Eliyaan:
I'm running `v -new-transformer test-self ` on the last commit to see.
I can compile `v -new-transformer self` with previously compiled `v -new-transformer self` that's already better! 
Thank you

Eliyaan:
this is the result of `v -new-transformer test-self` (with `v` being compiled with `v -new-transformer self`)
```
Summary for testing: cmd, vlib: 95 failed, 2481 passed, 43 skipped, 2619 total.
```
with attached the commands to reproduce the failures
[fail.txt](https://github.com/user-attachments/files/24154674/fail.txt)
Great progress! Thanks again

dy-tea:
Good that it compiles for you at least now. Will have to look into the test failures.

Eliyaan:
@spytheman 
Do we merge the PR as it is an improvement for `-new-transformer`? (before it would not compile `v -new-transformer self` and now the generated executable can recompile `v -new-transformer self`) (before not even the first tests of `v -new-transformer test-self` would pass).
Or do we wait until we fixed all the tests of `v -new-transformer test-self`?

spytheman:
If you think it is an improvement, it is better to merge it asap, to minimize the chances of conflicts with other PRs. Just mark it as non draft, then Alex or I will review it and merge it.

Afaik it is still gated behind `-new-transformer` so it will not affect normal runs for now?

vlib/v/transformer/transformer.v
```diff
@@ -1347,3 +1350,52 @@ pub fn (mut t Transformer) simplify_nested_interpolation_in_sb(mut onode ast.Stm
 	}
 	return true
 }
+
+fn expr_type(e ast.Expr) ast.Type {
```
StunxFS: I believe this function should be a method of `ast.Expr`

spytheman (approved):
Thank you @Eliyaan 🙇🏻‍♂️ .

enghitalo:
Any test? Or is it not possible to test?

spytheman:
> Any test? Or is it not possible to test?

it is still WIP; I've tested it manually, and it works so far

spytheman:
I'll add a test in `v test-all` soon after I finish watching and fixing stuff pointed out by https://youtu.be/IuE6Bo1klK0 .

Eliyaan:
Thank you @dy-tea !

Issue #25935, opened 2025-12-10
veb and vweb fail listening to IPv4 and IPv6 by default on FreeBSD 15


kimshrier:
### Describe the bug

With the release of FreeBSD 15, the setting of sysctl net.inet.ip.connect_inaddr_wild is now defaulted to 0.  This sysctl was introduced in FreeBSD 14 with a default value of 1.  The value of 1 in FreeBSD 14 made it compatible with earlier releases prior to the introduction of this sysctl.

To quote the release notes for FreeBSD 15:

> Making a connection to INADDR_ANY, i.e., using it as an alias for localhost, is now disabled by default. This functionality can be re-enabled by setting the net.inet.ip.connect_inaddr_wild sysctl to 1. 

If the desired behavior is to have veb.run (and vweb.run) open both an IPv4 and IPv6 socket when this sysctl is set to 0, then we would need to modify the code to open 2 sockets.

If on the other hand, we only want both IPv4 and IPv6 sockets opened when this sysctl is 1, we should modify the unit test(s) to assert that only IPv4 is being listened to when it is 0 and both IPv4 and IPv6 are being listened to when it is 1.

I'm happy to make the modifications.  I just need guidance on how to address this.

### Reproduction Steps

On FreeBSD 14 or later, set net.inet.ip.connect_inaddr_wild to 0 and then run vlib/veb/tests/veb_should_listen_on_both_ipv4_and_ipv6_by_default_test.v or vlib/vweb/tests/vweb_should_listen_on_both_ipv4_and_ipv6_by_default_test.v.

```
# sysctl net.inet.ip.connect_inaddr_wild=0
net.inet.ip.connect_inaddr_wild: 1 -> 0
```

```
$ v test vlib/veb/tests/veb_should_listen_on_both_ipv4_and_ipv6_by_default_test.v
---- Testing... ----------------------------------------------------------------------------------------------------------------------------
 FAIL    29.476 ms vlib/veb/tests/veb_should_listen_on_both_ipv4_and_ipv6_by_default_test.v
2025-12-10T19:00:43.449899Z [DEBUG] testsuite_begin
2025-12-10T19:00:43.449935Z [INFO ] working curl_executable found at: /usr/local/bin/curl
2025-12-10T19:00:43.449945Z [DEBUG] starting watchdog thread to ensure the test will always exit one way or another...
2025-12-10T19:00:43.450003Z [DEBUG] starting webserver...
[veb] Running app on http://localhost:48872/
[...]

PR #25972, opened 2025-12-14
veb, vweb: update unit tests for FreeBSD

kimshrier:
update unit tests to deal with net.inet.ip.connect_inaddr_wild on FreeBSD

This is a potential fix for issue [25935](https://github.com/vlang/v/issues/25935).

It checks for the existence of the net.inet.ip.connect_inaddr_wild sysctl and if it exists and if it is zero, it skips checking for the IPv6 socket listening for connections.

Prior to FreeBSD 14, there was no sysctl named net.inet.ip.connect_inaddr_wild and opening a socket on localhost would open both an IPv4 and IPv6 socket.  In FreeBSD 14, this sysctl was introduced with a default value or 1.  This retained the behavior of FreeBSD 13 and earlier.  With the release of FreeBSD 15, this value now defaults to 0 which means that opening a socket on localhost only opens the IPv4 socket.  There are more details in the related issue.

### DIFF
--- a/vlib/veb/tests/veb_should_listen_on_both_ipv4_and_ipv6_by_default_test.v
+++ b/vlib/veb/tests/veb_should_listen_on_both_ipv4_and_ipv6_by_default_test.v
@@ -82,6 +82,16 @@ fn test_net_http_connecting_through_ipv6_works() {
 		log.warn('skipping test ${@FN} on windows for now')
 		return
 	}
+	$if freebsd {
+		// if the sysctl net.inet.ip.connect_inaddr_wild doesn't exist or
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Thank you for investigating, and keeping these tests pass on the latest FreeBSD (even if skipped).

What do you think of the idea, of detecting it also at startup, in the implementation code for vweb and veb, and then showing a message about it (searching for `params.show_startup_message` can show where)?

kimshrier:
Adding a message at startup time sounds like a good idea.  I'll look at it if you like.

Issue #25970, opened 2025-12-14
Veb silently fails to bind a handler when the result type is not veb.Result


bradcypert:
### Describe the bug

Hi, thanks for the work on V and Veb! I noticed some behavior with Veb that felt off and wanted to report it. If the return type of a Veb handler is not veb.Result (for example, it is `!veb.Result`), veb silently fails to bind the handler (the compilation succeeds, but when running Veb and hitting that URL, it 404s).

I know that `!veb.Result` is a bit of a weird example, but it would be nice if compiler indicated that we're doing something wrong instead of letting Veb silently fail to bind the handler. If we couldn't address it at compile time, I think runtime would be an okay solution, too.

### Reproduction Steps

```v
module main

import veb

const port = 8082
pub struct App {}

struct Context {
	veb.Context
}

@['/tasks/'; get]
//pub fn (app &App) get_task(mut ctx Context) !veb.Result {
pub fn (app &App) get_task(mut ctx Context) veb.Result {
	return ctx.text("Hello")
}

fn main() {
	mut app := &App{}
	veb.run_at[App, Context](mut app, port: port, timeout_in_seconds: 2) or {
		panic(err)
	}
}
```

In the above code example, there is a single handler function. This one works as expected, however, if you comment out the handler declaration and uncomment the one above it (or you could simply change the return type from `veb.Result` to `!veb.Result`), the code will compile and run, but the handler will not be registered and Veb will return a 404 when you try to access that route.

### Expected Behavior

I would expect for V to give me a compiler error indicating that I am trying to use an invalid function type when registering a route handler which is invalid.

### Current Behavior

Currently, V compiles the program successfully and it is runnable. There is no indication that Veb failed to bind that handler to that route, and when you make a request to that route, Veb returns a 404, never calling the handler.

### Possible Solution

At compile-time, ensure that Veb handlers have the correction function signature and fail compilation if not. If this is not possible due to existing architecture, at runtime, we could log that a handler did not match the expected format and therefore it was not registered in veb.

### V version

V 0.4.12 dc0b0f5

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 90fa8bc56fd4c3870a9041e39fc156bfacdcd1b1.dc0b0f5
|:-------------------|:-------------------
[...]

PR #25973, opened 2025-12-15
veb:  fix veb silently fails to bind a handler when result type is not veb.Result

bradcypert:
Potential Fix for #25970.

Fixes an issue where Veb would silently fail to bind a handler when the return type of a handler function is not veb.Result (and is, for example, `!veb.Result`). I've also added tests to cover this. I'm happy to take a different approach if needed, too. This is my first contribution to V and am still learning the ropes.

Changes:
- Added a function to check for route attributes which is used in other items below
- When generating routes, adds a branch to handle the case where a method does not have a veb.Result return type, but does have route attributes, in which case, error to warn the user of unexpected behavior.

### DIFF
--- a/vlib/veb/parse.v
+++ b/vlib/veb/parse.v
@@ -95,3 +95,23 @@ fn parse_form_from_request(request http.Request) !(map[string]string, map[string
 	}
 	return http.parse_form(request.data), map[string][]http.FileData{}
 }
+
+// has_route_attributes checks if a method has attributes that indicate it should be a route handler
+fn has_route_attributes(attrs []string) bool {
+	if attrs.len == 0 {
+		return false
+	}
+	for attr in attrs {
+		attru := attr.to_upper()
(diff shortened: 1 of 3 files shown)


spytheman:
I played a bit with it, and discovered, that we are missing a source location information in the metadata for the method, but it would have been useful for situations like this 🤔 . 

For example, I got:
```
V panic: method `get_task` has route attributes but invalid return type. Handler methods must return `veb.Result`, not `!veb.Result` or other types
```

If we had a location information, then the error message could become:
```
V panic: method `vvv.v:13:18:get_task` has route attributes but invalid return type. Handler methods must return `veb.Result`, not `!veb.Result` or other types
```

I'll see what can be done about it asap.

spytheman:
With latest V 22fafd6, the error is now:
```
method `get_task` at `vlib/veb/tests/invalid_return_type_test.v:25:18` has route attributes but invalid return type. Handler methods must return `veb.Result`, not `!veb.Result` or other types
```

PR #25978, opened 2025-12-15
cgen: fix iterator on alias to fn ptr

felipensp:
Complement fix for #25911

### DIFF
--- a/vlib/v/tests/fns/iter_alias_fn_test.v
+++ b/vlib/v/tests/fns/iter_alias_fn_test.v
@@ -15,8 +15,11 @@ fn (mut it Iter) next() ?&MyFn {
 }
 
 fn test_main() {
+	mut c := 0
 	for p in Iter{} {
 		println(p)
 		p()
+		c += 1
 	}
+	assert c == 1
 }


vlib/v/tests/fns/iter_alias_fn_test.v
```diff
@@ -0,0 +1,22 @@
...
+fn (mut it Iter) next() ?&MyFn {
+	if it.done {
+		return none
+	}
+	it.done = true
+	return &it.f
+}
+
+fn test_main() {
+	for p in Iter{} {
```
spytheman, line 19: please also add a counter here, and an assertion for it, to make sure that the loop had been executed at least once

spytheman (approved):
Excellent work. 
Thank you @felipensp 🙇🏻 .

vlib/v/gen/c/for.v
```diff
@@ -548,19 +548,25 @@ fn (mut g Gen) for_in_stmt(node_ ast.ForInStmt) {
 		g.writeln('\tif (${t_var}.state != 0) break;')
 		val := if node.val_var in ['', '_'] { g.new_tmp_var() } else { node.val_var }
 		val_styp := g.styp(ret_typ.clear_option_and_result())
+		ret_sym := g.table.final_sym(ret_typ)
```
spytheman, line 551: `ret_sym` is not used in the `if node.val_is_mut {` branch, but it is cheap to calculate, and the overhead is probably very small

Issue #25986, opened 2025-12-16
can't assign value to a shared int in a struct


kbkpbot:
### Describe the bug

even lock the field, the checker issue an error message.

### Reproduction Steps

s.v
```v
module main

struct Foo {
mut:
        a shared int
}

fn main() {
        mut x := Foo{}
        lock x.a {
                x.a = 100
        }
}
```

```
v s.v
```

### Expected Behavior

compile ok

### Current Behavior

```
$ v s.v
s.v:11:9: error: cannot assign to `x.a`: expected `shared int`, not `int literal`
    9 |     mut x := Foo{}
   10 |     lock x.a {
   11 |         x.a = 100
      |               ~~~
   12 |     }
   13 | }
```

### Possible Solution

fix checker.

### V version

V 0.4.12 967a4a2d0a9e2e2e5da714c17fe5a8536fae53a8

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 967a4a2d0a9e2e2e5da714c17fe5a8536fae53a8
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |9.91GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-12-16 13:09:58
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/autostr
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.46-218-ga595c85d-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.6) 2.39

[...]

PR #25988, opened 2025-12-16
checker: allow assign to a shared int var

kbkpbot:
Fix issue #25986


spytheman (approved):
Very good work.
Thank you @kbkpbot 🙇🏻 .

Issue #25984, opened 2025-12-16
str intp can't work with shared int in a struct


kbkpbot:
### Describe the bug

If a struct contains a field with shared int type, it can't be used in a string intp correctly.

May related to PR #25857

### Reproduction Steps

shared.v
```v
module main

struct Foo {
        baz shared int
}

fn main() {
        mut t := Foo{}
        rlock t.baz {
                k := '${t.baz}'
                dump(k)
        }
}
```

```
v run shared.v
```

### Expected Behavior

```
[shared.v:11] k: 0
```

### Current Behavior

```
$ v run shared.v
[shared.v:11] k: 7d9711c1efa0
```

a pointer value.

### V version

V 0.4.12 967a4a2d0a9e2e2e5da714c17fe5a8536fae53a8

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 967a4a2d0a9e2e2e5da714c17fe5a8536fae53a8
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |10.17GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-12-16 11:04:05
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/autostr
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.46-218-ga595c85d-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.6) 2.39

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
[...]

PR #25989, opened 2025-12-16
cgen: fix shared int string intp

kbkpbot:
Fix issue #25984

### DIFF
--- a/vlib/v/tests/shared_str_inp_test.v
+++ b/vlib/v/tests/shared_str_inp_test.v
@@ -1,0 +1,34 @@
+module main
+
+struct Foo {
+mut:
+	int shared int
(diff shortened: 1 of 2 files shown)


vlib/v/tests/shared_str_inp_test.v
```diff
@@ -0,0 +1,34 @@
+module main
+
+struct Foo {
+mut:
+	int shared int
+	i32 shared i32
```
spytheman, line 6: i16?

spytheman: although that 8 and 16 bit numbers may not work now

penguindark:
Great work!
Congrats. 😄

PR #25993, opened 2025-12-17
Io/multiplexing/epoll

enghitalo:
This will help with consistency and code reduction across all these files.
```sh
grep -rl 'C.epoll_' .
```
```sh
./vlib/fasthttp/fasthttp_linux.v
./vlib/os/notify/backend_linux.c.v
./vlib/picoev/loop_linux.c.v
./vlib/picoev/loop_termux.c.v
```

### DIFF
--- a/vlib/io/multiplexing/epoll/epoll_linux.c.v
+++ b/vlib/io/multiplexing/epoll/epoll_linux.c.v
@@ -16,8 +16,22 @@ union C.epoll_data {
 	u64 u64
 }
 
+// C.epoll_event represents an event structure used by epoll.
+//
+// The 'events' field is a bitmask of event flags that epoll will monitor for the file descriptor.
+// Common event flags include:
+//   EPOLLIN      (0x001):  The associated file is available for read operations.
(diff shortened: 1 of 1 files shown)


spytheman:
epoll is linux specific, how do you intend to deal with that?

enghitalo:
I would prefer to throw an error message saying the library doesn't exist rather than creating a struct to handle it. It seems more sensible and more compatible with reality and with the C library. What do you think?

Furthermore, `$if linux { import epoll}` is already a V feature

spytheman:
Currently, there are no platform specific modules in `vlib`, where that can be helped (even though for some functionality, there are shims and empty implementations, that just keep code compile-able, even if it does not work). Adding something like this would create a precedent, and may need to be thought about more carefully - for example, it will delegate the decision on how to handle the different platforms to each using app/module, instead of encapsulating it in a single place.

enghitalo:
In fact, perhaps it's not the right time for that yet. I'm going to close this PR.

Issue #25959, opened 2025-12-13
compiler invalid runtime memory access


dy-tea:
### Describe the bug

Scanner issue in my project. Compiles fine before latest commit (of my project).
See https://github.com/dy-tea/mrpenishot commit 70bf04e30a9c6fdf3484e2f6a8436960ce5b07e3

### Reproduction Steps

On linux (you will need wayland-client, wayland-scanner, wayland-protocols, pixman-1 and pkg-config for build to work):
```
git clone https://github.com/dy-tea/mrpenishot
cd mrpenishot
git checkout 70bf04e30a9c6fdf3484e2f6a8436960ce5b07e3
v run build.vsh
v .
```

### Expected Behavior

Should build.

### Current Behavior

```
❯ v run .
/home/dylan/Repos/v/vlib/v/token/keywords_matcher_trie.v:173: at v__token__TrieNode_find: RUNTIME ERROR: invalid memory access
/home/dylan/Repos/v/vlib/v/token/keywords_matcher_trie.v:48: by v__token__KeywordsMatcherTrie_find
/home/dylan/Repos/v/vlib/v/token/keywords_matcher_trie.v:54: by v__token__KeywordsMatcherTrie_matches
/home/dylan/Repos/v/vlib/v/scanner/scanner.v:994: by v__scanner__Scanner_text_scan
/home/dylan/Repos/v/vlib/v/scanner/scanner.v:592: by v__scanner__Scanner_scan_remaining_text
/home/dylan/Repos/v/vlib/v/scanner/scanner.v:580: by v__scanner__Scanner_scan_all_tokens_in_buffer
/home/dylan/Repos/v/vlib/v/scanner/scanner.v:127: by v__scanner__new_scanner_file
/home/dylan/Repos/v/vlib/v/parser/parser.v:281: by v__parser__parse_file
/home/dylan/Repos/v/vlib/v/parser/parser.v:416: by v__parser__parse_files
/home/dylan/Repos/v/vlib/v/builder/builder.v:237: by v__builder__Builder_parse_imports
/home/dylan/Repos/v/vlib/v/builder/builder.v:115: by v__builder__Builder_front_stages
/home/dylan/Repos/v/vlib/v/builder/builder.v:170: by v__builder__Builder_front_and_middle_stages
/home/dylan/Repos/v/vlib/v/builder/cbuilder/cbuilder.v:72: by v__builder__cbuilder__gen_c
[...]

PR #25997, opened 2025-12-17
scanner: remove old generic check logic

kbkpbot:
Fix issue #25959

TODO in another PR maybe:
I am not sure why there is a need for `import v.ast as _`, remove this will cause `vlib/v/scanner/scanner_test.v` test fail. It seems to relate to `markused`?

### DIFF
--- a/cmd/tools/modules/testing/common.v
+++ b/cmd/tools/modules/testing/common.v
@@ -912,8 +912,8 @@ fn check_openssl_present() bool {
 		return false
 	}
 	$if openbsd {
-		return os.execute('eopenssl34 --version').exit_code == 0
-			&& os.execute('pkg-config eopenssl34 --libs').exit_code == 0
+		return os.execute('eopenssl35 --version').exit_code == 0
+			&& os.execute('pkg-config eopenssl35 --libs').exit_code == 0
 	} $else {
 		return os.execute('openssl --version').exit_code == 0
 			&& os.execute('pkg-config openssl --libs').exit_code == 0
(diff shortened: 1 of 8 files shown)


JalonSolov:
`import v.ast as _` is a way to have the initialization of `v.ast` run, without actually needing anything else in it.

spytheman:
@kbkpbot it seems that this fixes the need for `import v.ast as _` .
It is indeed a markused problem - it could not handle the const initialisation through a generic declared in another module, that is used in v.scanner, but moving it inside scanner.v works:
image

spytheman:
`scanner_matcher` is not used by other parts of the compiler, so it can be moved here for now more or less safely

spytheman:
hm, no - I am wrong; it is a deeper problem 🤔

spytheman:
This can be used as a workaround for now:
```
#0 17:57:54 ^ remove-old-generic-detect-logic ~/v>git diff

─────────────────────────────────────────────┐
• vlib/v/scanner/scanner.v:9: import v.token │
─────────────────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────
 9| 9| import v.pref
10|10| import v.util
11|11| import v.errors
12|  |-import v.ast as _
  |12|+
  |13|+@[markused]
  |14|+const workaround_markused_bug = map[string]int{}
13|15|
14|16| const single_quote = `'`
15|17| const double_quote = `"`
#0 17:57:56 ^ remove-old-generic-detect-logic ~/v>./v  cmd/tools/measure/scanner_speed.v
#0 17:58:09 ^ remove-old-generic-detect-logic ~/v>./v  -o x cmd/v
#0 17:58:17 ^ remove-old-generic-detect-logic ~/v>./x  -o x cmd/v
#0 17:58:22 ^ remove-old-generic-detect-logic ~/v>
```

Issue #25852, opened 2025-11-28
[generics] vlang gets wrong about type(s)


tankf33der:
`Short story first`: changing unrelated type affects the type of another variable in the future.

I spent 4 hours just to catch this strange behavior, now I'll have to spend a long time describing it, every detail is important otherwise the bug doesn't manifest!

Create dirs:
```
mkdir -p ~/.vmodules/tankf33der/mike
mkdir -p ~/gene
```
Put files:
```~/.vmodules/tankf33der/mike/vec2.v```
```v
module mike

pub struct Vec2[T] {
pub mut:
        x T
        y T
}

pub fn vec2[T](x T, y T, z T) Vec2[T] {
        return Vec2[T]{
                x: x
                y: y
        }
}
```
```~/.vmodules/tankf33der/mike/vec3.v```
```v
module mike

import math

pub struct Vec3[T] {
pub mut:
        x T
        y T
        z T
}

pub fn vec3[T](x T, y T, z T) Vec3[T] {
        return Vec3[T]{
                x: x
                y: y
                z: z
        }
}

pub fn (v Vec3[T]) dot(u Vec3[T]) T {
        return T((v.x * u.x) + (v.y * u.y) + (v.z * u.z))
}

pub fn (v Vec3[T]) angle_between(u Vec3[T]) T {
        n_v := T(math.sqrt(v.dot(v)))
        println('${typeof(n_v).name}: ${n_v}')
        n_u := T(math.sqrt(u.dot(u)))
        println('${typeof(n_u).name}: ${n_u}: n_u')
        norm := T(T(1.0) / T(n_v * n_u))
        println('${typeof(norm).name}: ${norm}: norm')
        dot := v.dot(u) * norm
        println('${typeof(dot).name}: ${dot}: dot')
        return T(math.acos(dot))
}
```
```~/gene/cglm.v```
```v
module kuku

import tankf33der.mike

fn eqeqeqeq22(a mike.Vec2[f32]) bool {
        return false
}

fn eq33eqeqeq(a mike.Vec3[f32]) bool {
        return false
}
```
```~/gene/gene_test.v```
```v
module kuku

import tankf33der.mike

fn test_main() {
        a1 := f64(11.1)
        b1 := f64(22.2)
        c1 := f64(33.3)
        a2 := f64(10.0)
        b2 := f64(11.0)
        c2 := f64(10.0)
        v := mike.vec3[f64](a1, b1, c1)
        u := mike.vec3[f64](a2, b2, c2)
[...]

PR #26008, opened 2025-12-17
checker: partial fix for generic type pollution bug

bradcypert:
Potential fix for #25852

In generic functions with multiple type instantiations, typeof() expressions show incorrect types. Type information from one instantiation (e.g., f32) pollutes another instantiation (e.g., f64).

AST nodes are shared across all generic instantiations. When checking multiple instantiations, type information gets stored in shared AST nodes and pollutes subsequent instantiations.

Added infrastructure to cache types per instantiation:
- generic_type_cache map in Table
- Scope cloning for each instantiation
- Type caching during checking
- Type retrieval during generation

The fix works for the reported bug but ultimately seems to cause more harm than good.

At this point, I'm looking for guidance from maintainers who are more familiar with V's inner-workings than I am :)

Some possible paths forward:
1. Enhance type resolution map to handle all edge cases
2. Targeted typeof() fix only (wouldn't really address root cause)
3. Something else?

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -2952,3 +2952,14 @@ pub fn validate_type_string_is_pure_literal(typ Type, str string) ! {
 		return error('expected pure literal, found "${str}"')
 	}
 }
+
+// clone creates a shallow copy of the scope for generic function instantiation
+// Variables will be re-created during checking, so we just need a fresh container
+pub fn (s &Scope) clone_shallow() &Scope {
+	return &Scope{
+		parent:               s.parent
+		detached_from_parent: s.detached_from_parent // Keep same detachment as original
+		start_pos:            s.start_pos
+		end_pos:              s.end_pos
+	}
+}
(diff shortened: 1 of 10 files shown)


bradcypert:
I think we need to implement full AST cloning for generic function instantiations because the current architecture shares AST nodes across all instantiations, causing type information from one instantiation (e.g., Vec3[f32]) to pollute subsequent instantiations (e.g., Vec3[f64]), and I think that any patch that doesn't address this fundamental shared-state problem will continue to cause regressions (we currently have many test failures despite fixing the original bug).

spytheman:
@kbkpbot, @felipensp what do you think?

JalonSolov:
I would much prefer a small amount of memory overhead during the compile, in order to have more correct/robust handling of the code.

spytheman:
Sorry, but it seems to me, that solving this issue, will probably need to wait, till a separate stage resolves it properly.

PR #26010, opened 2025-12-18
wasm: fix const reference another const (fix #25722)

### LINKED ISSUE #25722: unexpected key and value in map foreach
### Describe the bug

v/gen/wasm/mem.v:815 pub fn (mut g Gen) make_vinit() 
```v
		for idx, gv in g.global_vars {
			println('`$idx` $gv')
			if init := gv.init {
				g.set_with_expr(init, gv.v)
			}
		}
```
I add `println` before use it, and print out the text below, `g.global_vars` does not have an empty str key
```
...
...
...
`main.t_mask1` wasm.Global{
    init: Option(value)
    v: wasm.Var{
        name: 'main.t_mask1'
        typ: ast.Type(0x8 = 8)
        idx:         wasm.LocalIndex(0)
        is_address: false
        is_global: true
        g_idx:         wasm.GlobalIndex(5)
        offset: 0
    }
}
`` wasm.Global{
    init: Option([unhandled expr type unknown v.ast.Expr])
    v: wasm.Var{
        name: ''
        typ: ast.Type(0x0 = 0)
        idx:         wasm.LocalIndex(0)
        is_address: false
        is_global: false
        g_idx:         wasm.GlobalIndex(0)
        offset: 0
    }
}
```
first, unexpected key and value should not happen, and then, the wrong value `gv.init` seem like `none`, cannot set to `init`

### Reproduction Steps

1. code
```v
module main

const value = 53
const t_mask1 = value
const t_mask2 = value
const t_mask3 = value
const t_mask4 = value
const t_mask5 = value

fn main() {
	println(t_mask1)
	println(t_mask2)
	println(t_mask3)
	println(t_mask4)
	println(t_mask5)
}
```
2. command
```sh
v -b wasm -g main.v
```

### Expected Behavior

compile the code

### Current Behavior

DESKTOP-D5EJCFL% v -b wasm -g main.v
wasm error: wasm.expr(): unhandled node: unknown v.ast.Expr

### Possible Solution

_No response_

_No response_

### V version

V 0.4.12

windows and wsl both have the problem
> [!NOTE]
>

bradcypert:
Should fix: #25722

With code like this:
```v
const value = 53
const t_mask1 = value  // t_mask1 references another const
```

The WASM backend needs to initialize `t_mask1` with the value of value. The expression value is represented as an `ast.Ident` (an identifier node in the AST).

When it encountered `ast.Ident` (a reference to another constant), it returned none, which caused the initialization to be deferred to the _vinit function. But the deferred initialization couldn't handle the unresolved identifier, resulting in the error:  `wasm error: wasm.expr(): unhandled node: unknown v.ast.Expr`

   The Solution

   Our fix adds handling for ast.Ident by:

- Looking up the identifier - Checking if it's a ConstField or GlobalField
- Recursively resolving it - Calling literal_to_constant_expression() again with the referenced constant's expression (obj.expr)
- Returning the resolved value - This allows const-to-const references to resolve at compile time

This means when the compiler sees const t_mask1 = value, it now:
- Recognizes value is an identifier
- Looks up what value refers to (the constant with value 53)
- Recursively resolves 53 to a constant expression
- Uses that constant expression to initialize t_mask1

The recursive approach naturally handles chained references like:
```v
const a = 42
const b = a
const c = b  // c -> b -> a -> 42
```
   Each level of the recursion peels back one layer of the reference until it reaches an actual literal value.

### DIFF
--- a/vlib/v/gen/wasm/tests/misc.vv
+++ b/vlib/v/gen/wasm/tests/misc.vv
@@ -38,8 +38,8 @@ const runtime_init = constant()
 // Test constants referencing other constants
 const base_value = 53
 const ref_const1 = base_value
-const ref_const2 = base_value
-const ref_const3 = base_value
+const ref_const2 = ref_const1
+const ref_const3 = ref_const2
 
 struct EE {
 	a int


vlib/v/gen/wasm/tests/misc.vv
```diff
@@ -35,6 +35,12 @@ const integer = 888
 
 const runtime_init = constant()
 
+// Test constants referencing other constants
+const base_value = 53
+const ref_const1 = base_value
+const ref_const2 = base_value
+const ref_const3 = base_value
```
spytheman: Why have 3 consts here, that all follow the same pattern for the right side of the assignment?

After reading the explanation, I thought there will be some using of `ref_const1` by `ref_const2` and `ref_const3` using `ref_const2` 🤔 .

spytheman (commented):
Excellent work.
Thank you @bradcypert .

PR #26022, opened 2025-12-18
checker: ensure fn pointer cannot be used as non-fn argument

### LINKED ISSUE #26017: error: 'string_int' undeclared
<details>
<summary>V version: V 0.4.12 9381714, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 9381714b5897072238cd3ccbf9822c6e2ba61fde
|OS                  |linux, Ubuntu 24.04.2 LTS
|Processor           |8 cpus, 64bit, little endian, 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
|Memory              |0.62GB/7.56GB
|                    |
|V executable        |/opt/v/v
|V last modified time|2025-12-18 10:25:35
|                    |
|V home dir          |OK, value: /opt/v
|VMODULES            |OK, value: /home/arghyas/.vmodules
|Current working dir |OK, value: /home/arghyas/Programs/v_playground
|                    |
|Git version         |git version 2.43.0
|V git status        |N/A
|.git/config present |false
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |N/A
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.6) 2.39

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  calc.v && calc`
```v
import os


fn main() {
	if os.args.len < 2 {
		usage()
		panic("Invalid Invocation")
	}

	option := os.args[1]
	arg1 := os.args[2]
	arg2 := os.args[3]

	result := match option {
		'-a', '--add' { add(arg1.int, arg2.int) }
		'-s', '--sub' { sub(arg1.int, arg2.int) }
		else { panic("Invalid Option") }
	}

	println(result)
}

fn usage() {
	println("BASIC CALCULATOR")
	println("Usage:")
	println("\tcalc --add 2 3")
	println("\tcalc --sub 5 2")
}

fn add(x int, y int) int {
	return x+y
}

fn sub(x int, y int) int {
	return x-y
}

```

**What did you see?**
```
cc: /tmp/v_1000/calc.01KCRJ6FM9CT1E575CYX1WNFNH.tmp.c:6097: error: 'string_int' undeclared
builder error: 
==================
```

Compiling and running this program without passing any arguments(`v run calc.v`) the output should have been "BASIC CALCULATOR"

> [!NOTE]
>

dy-tea:
Fixes #26017.

Test is simplified from the original issue.

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -1900,8 +1900,10 @@ fn (mut c Checker) selector_expr(mut node ast.SelectorExpr) ast.Type {
 	if mut method := c.table.sym(c.unwrap_generic(typ)).find_method_with_generic_parent(field_name) {
 		c.markused_comptime_call(typ.has_flag(.generic), '${int(method.params[0].typ)}.${field_name}')
 		if c.expected_type != 0 && c.expected_type != ast.none_type {
-			fn_type := ast.new_type(c.table.find_or_register_fn_type(method, false, true))
-			// if the expected type includes the receiver, don't hide it behind a closure
+			mut method_copy := method
+			method_copy.name = ''
+			fn_type := ast.new_type(c.table.find_or_register_fn_type(method_copy, false,
+				true))
 			if c.check_types(fn_type, c.expected_type) {
 				c.table.used_features.anon_fn = true
 				return fn_type
--- a/vlib/v/checker/tests/method_ptr_used_as_argument_err.out
+++ b/vlib/v/checker/tests/method_ptr_used_as_argument_err.out
@@ -1,0 +1,14 @@
+vlib/v/checker/tests/method_ptr_used_as_argument_err.vv:8:16: error: cannot use `fn () int` as `int` in argument 1 to `add`
+    6 |     arg1 := '10'
+    7 |     arg2 := '20'
+    8 |     result := add(arg1.int, arg2.int)
+      |                   ~~~~~~~~
+    9 |     println(result)
+   10 | }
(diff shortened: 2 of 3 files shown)


vlib/v/checker/checker.v
```diff
@@ -1900,8 +1900,10 @@ fn (mut c Checker) selector_expr(mut node ast.SelectorExpr) ast.Type {
 	if mut method := c.table.sym(c.unwrap_generic(typ)).find_method_with_generic_parent(field_name) {
 		c.markused_comptime_call(typ.has_flag(.generic), '${int(method.params[0].typ)}.${field_name}')
 		if c.expected_type != 0 && c.expected_type != ast.none_type {
-			fn_type := ast.new_type(c.table.find_or_register_fn_type(method, false, true))
-			// if the expected type includes the receiver, don't hide it behind a closure
+			mut method_copy := method
+			method_copy.name = ''
```
spytheman, line 1904: hm, can you please clarify why does setting the name to '' work here?

dy-tea: It forces it to generate a function signature instead of using `int` as the type.

dy-tea: Essentially it is using `int` as the type while it should only be the method name.

spytheman: I still do not understand it, but ok - the test does work.

PR #26026, opened 2025-12-19
vet: use new annotation syntax ([] -> @[])

Le0Developer:
Vet was still using the old syntax for parsing function annotations.

Nobody noticed because it's collecting the annotations but not doing anything with them 🤷


spytheman:
I am a bit uncomfortable to merge changes without corresponding tests, but it is good work, since the older `[attrib]` syntax is deprecated for an year now, in favor of the `@[attrib]` one.

Issue #26028, opened 2025-12-19
bool(10) != true


tankf33der:
Please comment or fix.

```
$ cat bool.v
fn mike_1() bool {
        ret := unsafe { bool(1) }
        dump(ret)
        return ret
}

fn mike_10() bool {
        ret := unsafe { bool(10) }
        dump(ret)
        return ret
}

fn main() {
        v1 := mike_1()
        dump(v1)[bool.v:3] ret: true
        println(v1 == true)

        v10 := mike_10()
        dump(v10)
        println(v10 == true)
}
$ v run bool.v
[bool.v:3] ret: true
[bool.v:15] v1: true
true
[bool.v:9] ret: true
[bool.v:19] v10: true
false
$
```

Although for tests with `10`, `true` is always written at all stages, the last comparison writes `false`.

It was discovered while testing i386 assembly tests.
Will have to depend and wait.
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

tankf33der:
@spytheman @kbkpbot 

I would like to reopen this issue to show code that still doesn't work for me.
Note that `dump(ret)` before `return ret` shows `true`, but then again it cannot be compared with `true`. What you think?
```
fn generic_asm[T](var &T) T {
	dump(var)
	mut ret := unsafe { T(14) }
	unsafe {
		asm amd64 {
			add var, ret
			; +m (var[0]) as var
			  +r (ret) as ret
			; ; memory
		}
	}
	dump(ret)
	return ret
}

fn main() {
	b := unsafe { bool(123) }
	dump(b)
	println(generic_asm(b) == true)
}
```

spytheman:
imho make a new issue - the new code uses generics, and likely needs a different solution too

PR #26031, opened 2025-12-19
cgen: fix bool cast from value

kbkpbot:
Fix issue #26028 

because in `cgen`, `true` is defined as `1`.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5687,13 +5687,14 @@ fn (mut g Gen) cast_expr(node ast.CastExpr) {
 		if node_typ_is_option {
 			g.expr_with_opt(node.expr, expr_type, node.typ)
 		} else {
-			if g.pref.translated || g.file.is_translated {
+			if (g.pref.translated || g.file.is_translated) && g.inside_global_decl {
 				styp := g.styp(node.typ)
 				g.write('(${styp})')
 				g.expr(node.expr)
 			} else if node.typ == ast.bool_type && expr_type == ast.bool_type {
 				g.expr(node.expr)
 			} else {
+				// due to tcc(0.9.27) bug, can't use `(cond)?1:0` here
 				styp := g.styp(node.typ)
 				g.write('(${styp}[]){(')
 				g.expr(node.expr)


vlib/v/gen/c/cgen.v
```diff
@@ -5688,6 +5688,10 @@ fn (mut g Gen) cast_expr(node ast.CastExpr) {
 		g.write('(${styp}[]){(')
 		g.expr(node.expr)
 		g.write(')?1:0}[0]')
+	} else if node.typ == ast.bool_type {
+		g.write('((')
+		g.expr(node.expr)
+		g.write(' == 0)?0:1)')
```
spytheman: can we merge those 2 cases?

spytheman: it seems that the above one was for `windows-tcc` 🤔

spytheman: and for `int(boolean)`, so no

kbkpbot: Yes

vlib/v/tests/casts/cast_bool_to_int_test.v
```diff
@@ -6,9 +6,18 @@ fn test_cast_bool_to_int() {
...
-	assert a[int(i)] == 2
+		println(a[x])
+		assert a[x] == 2
+	} $else {
+		println(a[int(!i)])
+		assert a[int(!i)] == 1
+
+		println(a[int(i)])
+		assert a[int(i)] == 2
+	}
```
spytheman: Adding environment specific workarounds, for something that worked before, is not a good idea for such a PR - it will break existing user code in the wild.

Imho just use the previous `g.write('(${styp}[]){(')` based solution, but also add the separate case too for the reverse cast - initially I thought that both can be combined, but it is now clear that leads to more hacks and workarounds.

spytheman: I apologize for the confusion (that my comment from yesterday created).

kbkpbot: This is due to a bug of old version(0.9.27) TCC/windows, after upgrade to newer version(0.9.28rc) of tcc/windows, this bug has been fixed.
 Never mind, I will revert to original workaround code.

spytheman:
Afaik the failures now only happen for code with `g.pref.translated`.
What do you think of using the simpler expression if that is true?

As an alternative, if you want, we can try updating the prebuilt tcc windows version, and then use the simpler static cast without the [] workaround, but afaik, last time there were other issues with it 🤔 .

PR #26032, opened 2025-12-19
toml, tests: update `burntsushi` -> `toml-lang` and `alexcrichton` ->`toml-rs`

larpon:
Due to the new names/repositories using `-` I chose to keep things on the V side using `_` since it makes it easier to handle in regards to directory names and, more importantly, variable names. So e.g. `toml-lang` is `toml_lang` *except* in some special-cases like in some comment URLs (which I have checked is still valid).

### DIFF
--- a/.github/workflows/download_full_toml_test_suites.sh
+++ b/.github/workflows/download_full_toml_test_suites.sh
@@ -2,15 +2,15 @@
 
 set -ex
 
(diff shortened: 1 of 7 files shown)


spytheman (approved):
Thank you @larpon 🙇🏻 .
Excellent work.

Issue #26027, opened 2025-12-19
x.json2 fails to decode correct json it generated. `Invalid json: Data: expected integer but got real number 	"g": 58,`


medvednikov:
### Describe the bug

```v
import gg
import x.json2
import os

const min_text_size = 18

enum Cursor {
	block
	beam
	variable
}


struct Config {
mut:
	// settings        toml.Doc
	dark_mode       bool
	cursor_style    Cursor
	text_size       int = min_text_size
	line_height     int = 20
	char_width      int = 8
	tab_size        int = 4
	tab             int = int(`\t`) // TODO read from config file?
	backspace_go_up bool
	vcolor          gg.Color // v selection background color
	split_color     gg.Color
	bgcolor         gg.Color // base00
	errorbgcolor    gg.Color // base08
	title_color     gg.Color // base04
	cursor_color    gg.Color // base05
	string_color    gg.Color // base0B
	string_cfg      gg.TextCfg
	key_color       gg.Color // base0E
	key_cfg         gg.TextCfg
	lit_color       gg.Color // base0E
	lit_cfg         gg.TextCfg
	text_color      gg.Color // base05
	txt_cfg         gg.TextCfg
	comment_color   gg.Color // base03
	comment_cfg     gg.TextCfg
	file_name_color gg.Color
	file_name_cfg   gg.TextCfg
	plus_color      gg.Color
	plus_cfg        gg.TextCfg
	minus_color     gg.Color
	minus_cfg       gg.TextCfg
	line_nr_color   gg.Color // base01
	line_nr_cfg     gg.TextCfg
	green_color     gg.Color // base0B
	green_cfg       gg.TextCfg
	red_color       gg.Color // base08
	red_cfg         gg.TextCfg
	disable_mouse   bool = true
	show_file_tree  bool
	// Config.json
	disable_fmt bool
}

text := os.read_file('config.json')!

x := json2.decode[Config](text) or { panic(err)}
println(x)

```

config.json:
```
{
    "dark_mode": false,
    "cursor_style": "block",
    "text_size": 20,
    "line_height": 22,
    "char_width": 9,
    "tab_size": 4,
    "tab": 9,
    "backspace_go_up": false,
    "vcolor": {
        "r": 226,
        "g": 233,
        "b": 241,
        "a": 255
    },
    "split_color": {
        "r": 223,
        "g": 223,
        "b": 223,
        "a": 255
    },
    "bgcolor": {
        "r": 245,
        "g": 245,
        "b": 245,
[...]

PR #26033, opened 2025-12-19
x.json2: fix decode_number for u8

kbkpbot:
Fix issue #26027 

two digits u8 decode may fail with origin code.
Use `strconv.atoxx` instead.


enghitalo:
Some performance impact?

spytheman:
> Some performance impact?

correctness is more important

PR #26036, opened 2025-12-19
Examples: Add Lorem Ipsum Generator.

mike-ward:
Adds a, "Lorem Ipsum" generator to the examples folder.

Features:
- All parameters configurable via flags
- Reproducible output via --seed
- No invalid punctuation sequences
- Em-dashes never near sentence edges

### DIFF
--- a/examples/lorem.v
+++ b/examples/lorem.v
@@ -16,18 +16,22 @@ pub struct LoremCfg {
 	words_per_sentence      int = 10
 	sentences_per_paragraph int = 5
 	commas_per_sentence     int = 2
+	seed                    int
 }
 
 // ---------------- Generator --------------------
 
 pub fn generate_lorem(cfg LoremCfg) string {
-	lorem_words := (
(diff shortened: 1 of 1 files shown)


examples/lorem.v
```diff
@@ -0,0 +1,165 @@
...
+		eprintln(err)
+		println(fp.usage())
+		return
+	}
+
+	// Deterministic and portable seeding
+	if seed != 0 {
+		rand.seed([u32(seed), u32(seed ^ 0x9e3779b9)])
+	} else {
+		rand.seed([u32(0x12345678), 0x9abcdef0])
```
spytheman: Why fix it here?
If a specific seed is not given, it can generate random output every time, that matches the other constraints, like `pwgen` does.

mike-ward: Yeah, I'm having one of those senior moment days. Will fix.

spytheman (approved):
Excellent work.
Thank you @mike-ward 🙇🏻 .

PR #26039, opened 2025-12-19
parser: change error message prefix to 'parser error:'

enghitalo:


spytheman:
why?

JalonSolov:
It does help to know where the error came from, rather than having to grep the source.  This won't tell you exactly which file, but it does make the phase explicit... parser rather than checker, cgen, etc.

spytheman:
From the point of view of a normal user, that does not help much - an error is an error, no matter what stage produced it.

From the point of view of a compiler developer, it does not help either - you still need to search for the exact error, if you want to see where it occurred, to debug/fix it, in case it is wrong or inappropriate.

You can also use `v -check-syntax` or `v -check` to localize it to a specific stage.

Issue #25630, opened 2025-10-30
example examples/wasm/functions.v fails


Manuelbaun:
### Describe the bug

when trying this example, 
the resulting wasm file, does not export pub functions.

I have uploaded the wasm file here https://wa2.dev and showed the wat format, which did not include 
(export "gcd" (func 0))        
(export "powi" (func 1))      

```log
(base) ➜  wasm wasmer run functions.wasm -i main.gcd 20 10
⠁ Compiling to WebAssembly                                                                                                                                                    error: The module doesn't export a function named "main.gcd"
╰─▶ 1: Missing export main.gcd
```

### Reproduction Steps

// v -b wasm -no-builtin functions.v
// wasmer functions.wasm -i main.powi <a> <b>
// wasmer functions.wasm -i main.gcd <a> <b>

### Expected Behavior

functions should be exported

### Current Behavior

functions are not exported

### V version

V 0.4.12 f001561

### Environment details (OS name and version, etc.)

macos 26
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #26041, opened 2025-12-19
wasm: fix export pub fns in main module when targeting wasi

bradcypert:
Fixes: #25630

The WASM backend was only exporting pub functions when the target OS was `.browser`. When compiling with -b wasm (which defaults to .wasi target), pub functions were not exported. This prevented tools like wasmer from invoking the functions.

I've updated the logic to determine public exports in the main module to also include the `.wasi` target. Not entirely sure what a test for this would look like without adding a dependency on something like wasmer (or parsing the generated output file, maybe?)


spytheman:
Excellent work, thank you.
I'll add a test, do not worry about it.

spytheman:
The test is in https://github.com/vlang/v/pull/26048 .

Issue #26045, opened 2025-12-20
create valid float string in scientific notation


kbkpbot:
### Describe the bug

In scientific notation of float, it should not omit the `0` between `.` and `e`.
e.g.
`1.0e+10`, but not `1.e+10`

Although in `C` and `Go`, `1.e+10` is a valid format.
 
This is related to issue #26043

### Reproduction Steps

f.v
```v
module main

fn main() {
        val1 := 1.e+10
        val2 := 1.0e+10

        str1 := '${val1}'
        str2 := '${val2}'

        dump(str1)
        dump(str2)
}
```

```
v run f.v
```

### Expected Behavior

```
$ v run f.v
[f.v:10] str1: 1e+10
[f.v:11] str2: 1e+10
```

### Current Behavior

```
$ v run f.v
[f.v:10] str1: 1.e+10
[f.v:11] str2: 1.e+10
```

### V version

V 0.4.12 1cdb0f5755595ddb1dbca2b5bc29da8080d643ad.8b423a4

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 1cdb0f5755595ddb1dbca2b5bc29da8080d643ad.8b423a4
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |11.02GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-12-20 01:26:49
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/json
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.46-247-g97445433
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
[...]

PR #26046, opened 2025-12-20
strconv: fix float scientific notation

kbkpbot:
Fix issue #26045 

`1.e+10` => `1e+10`


penguindark:
I've doubt because C/C++ support the float format `1.e+1` and it is valid and used.
What is the rationale for V not supporting it? 🤔 

```c
float x = 1.e+2; // C Valid, converts double 100.0 to float 100.0
```

This code yield the same output:
```c
#include <stdio.h>
int main()
{
    float a = 1.e+2;
    float b = 1e+2;
    printf("%f %f",a,b);
    return 0;
}
```

### C/C++ float/double Summary Table
| Literal | Valid? | Type | Value |
| :--- | :--- | :--- | :--- |
| `1.e+2` | **Yes** | `double` | 100.0 |
| `1.` | **Yes** | `double` | 1.0 |
| `.5` | **Yes** | `double` | 0.5 |
| `1.e+2f` | **Yes** | `float` | 100.0 |
| `1e2` | **Yes** | `double` | 100.0 | (No dot needed if `e` is present) |

spytheman:
Thank you @penguindark , thank you @kbkpbot 🙇🏻 .

PR #26048, opened 2025-12-20
wasm: add vlib/v/gen/wasm/tests_decompile/must_have_test.v to allow more extensive output testing for different compilation options

spytheman:


bradcypert:
Nice to see how'd you'd test that change in #26041, thank you!

PR #26050, opened 2025-12-20
cgen,slow_tests: add riscv64 inline asm support, add tests

tankf33der:
I am conducting an audit of all inline assembler support for each architecture and tests, and will create PRs as I find things that need to be fixed.

Lets add `riscv64` inline asm support.
Patch and tests created on real riscv64 environment (ubuntu 24.04.3) under qemu.
Tests works under gcc and clang.

One could nitpick about the test file name `asm_test.rv64.v`, but I chose the name based on the actual architecture name inside sources, all other numerous variations are aliases.


spytheman:
Please also add `./v test vlib/v/slow_tests/assembly` to the end of `.github/workflows/riscv64_linux_ci.yml` too, if you add such tests.

PR #26052, opened 2025-12-20
toml, tests: update `toml-rs` test suite to last available commit before files where removed

larpon:
For the sake of traceability and brief digital archeology/history:

This PR accounts for a few changes to the `toml-rs` test suite we've tested V's `toml` module against;
* [7f5472c](https://github.com/toml-rs/toml/commit/7f5472c) the `test-suite` directory moves to the `crates/` sub-directory
* [8461f7c](https://github.com/toml-rs/toml/commit/8461f7cfd963f116eb66b6c1558faed0532e453d#diff-ab1b28a84f7396fedb32848a984de4e1197f7b7ea34b0dbc6c7c898b5f68d1f1) *a lot* of test files are removed (in *hope*?) that they are covered by the compliance test suite (that I assume to be BurntSushi/toml-test, later `toml-lang/toml-test`)

V is now tested against commit [9bd454c](https://github.com/toml-rs/toml/commit/9bd454c) which, my detective work tells me, is the last known good commit that contains the test data we've been using so far.

Commits after `9bd454c` does seem to contain some mostly invalid `.toml` files but they are scattered around the project so would require some work to write tests for.

I assume that `toml-rs` is also using the now official `toml-lang/toml-test` test suite (I haven't dug deep to try and figure it out though).
So... despite some confusion I think we can concentrate on testing against https://github.com/toml-lang/toml-test going forward. For now I do not think it hurts much to still test against the stale ones (iarna/toml-rs remains).

### DIFF
--- a/.github/workflows/download_full_toml_test_suites.sh
+++ b/.github/workflows/download_full_toml_test_suites.sh
@@ -12,5 +12,9 @@ git -C vlib/toml/tests/testdata/iarna checkout 1880b1a
 ./v retry -- git clone -n https://github.com/toml-lang/toml-test.git vlib/toml/tests/testdata/toml_lang
 git -C vlib/toml/tests/testdata/toml_lang checkout f30c716
 
+# A few history notes of toml-rs (previously alexcrichton):
+#  commit 7f5472c the test-suite dir moves to the crates/ sub-directory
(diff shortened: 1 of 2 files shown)


larpon:
I'm happy to be contributing whenever I can!

PR #26054, opened 2025-12-20
C builtin function: memchr

enghitalo:
libc memchr is vector-accelerated via glibc IFUNC

use case example
```v
fn find_byte(buf &u8, len int, c u8) int {
	unsafe {
		p := C.memchr(buf, c, len)
		if p == voidptr(nil) {
			return -1
		}
		return int(p - buf)
	}
}
```

### DIFF
--- a/vlib/builtin/cfns.c.v
+++ b/vlib/builtin/cfns.c.v
@@ -12,7 +12,7 @@ fn C.memmove(dest voidptr, const_src voidptr, n usize) voidptr
 
 fn C.memset(str voidptr, c int, n usize) voidptr
 
-fn C.memchr(buf &u8, char int, len usize) &u8
+fn C.memchr(str voidptr, c int, n usize) voidptr
 
 @[trusted]
 fn C.calloc(int, int) &u8


enghitalo:
## TCC
```sh
 SPENT  5031.665 ms in find_byte_by_memchr
 SPENT 13530.281 ms in find_byte_by_index
 SPENT 27789.837 ms in find_byte_by_loop
```

## GCC
```sh
 SPENT     0.000 ms in find_byte_by_memchr
 SPENT     0.000 ms in find_byte_by_index
 SPENT   123.458 ms in find_byte_by_loop
```
## Benchmark
```v
import benchmark

const interactions = 1_000_000_000

fn C.memchr(buf &u8, char int, len usize) &u8

@[inline]
fn find_byte_by_memchr(buf &u8, len int, c u8) int {
	unsafe {
		p := C.memchr(buf, c, len)
		if p == voidptr(nil) {
			return -1
		}
		return int(p - buf)
	}
}

fn find_byte_by_index(buf []u8, c u8) int {
	return buf.index(c)
}

fn find_byte_by_loop(buf []u8, c u8) int {
	for i := 0; i < buf.len; i++ {
		if buf[i] == c {
			return i
		}
	}
	return -1
}

fn main() {
	data := 'GET /path/resource.html HTTP/1.1\r\nHost: example.com\r\nUser-Agent: VlangTestClient/1.0\r\nAccept: */*\r\n\r\n'
	buf := data.bytes()

	mut b := benchmark.start()
	for _ in 0 .. interactions {
		_ = find_byte_by_memchr(buf.data, buf.len, u8(` `))
	}
	b.measure('find_byte_by_memchr')


	for _ in 0 .. interactions {
		_ = find_byte_by_index(buf, u8(` `))
	}
	b.measure('find_byte_by_index')

	for _ in 0 .. interactions {
		_ = find_byte_by_loop(buf, u8(` `))
	}
	b.measure('find_byte_by_loop')

	b.stop()
}
```

vlib/builtin/cfns.c.v
```diff
@@ -12,6 +12,8 @@ fn C.memmove(dest voidptr, const_src voidptr, n usize) voidptr
 
 fn C.memset(str voidptr, c int, n usize) voidptr
 
+fn C.memchr(buf &u8, char int, len usize) &u8
```
spytheman:
```suggestion
fn C.memchr(str voidptr, c int, n usize) voidptr
```

spytheman: The exact types should be as close as possible to the C ones, to get more help from the compiler when fixign `-cstrict` errors.

PR #26067, opened 2025-12-21
toml: add compliance up until official toml-lang/toml-test@c6a78f1

larpon:
This PR adds compliance with the official TOML test suite up until and including toml-lang/toml-test@c6a78f1 (TOML v1.0)

I had to introduce a new parser helper array (`immutable`) to help detect immutability issues.

I'm starting to think the TOML parser might have been better off keeping track of scopes and types as well since some of the fixes here feels more like shoehorning to me but let's see what the future holds.

I'm hoping to fix some "timezone overflow" checks (added in commits after toml-lang/toml-test@c6a78f1) in a follow up PR which should bring us closer to where TOML v1.1 support was pushed to the test suite.

### DIFF
--- a/vlib/toml/parser/parser.v
+++ b/vlib/toml/parser/parser.v
@@ -59,6 +59,8 @@ mut:
 	// The root map (map is called table in TOML world)
 	root_map                          map[string]ast.Value
 	root_map_key                      DottedKey
+	value_is_immutable                bool
+	immutable                         []DottedKey
 	explicit_declared                 []DottedKey
 	explicit_declared_array_of_tables []DottedKey
 	implicit_declared                 []DottedKey
@@ -274,6 +276,14 @@ fn (p &Parser) build_abs_dotted_key(key DottedKey) DottedKey {
 // TODO: remove.
 fn todo_msvc_astring2dkey(s []string) DottedKey {
 	return s
+}
+
+// check_immutable returns an error if `key` has been declared as immutable.
+fn (p &Parser) check_immutable(key DottedKey) ! {
+	if p.immutable.len > 0 && p.immutable.has(key) {
+		return error(@MOD + '.' + @STRUCT + '.' + @FN +
+			' key `${key.str()}` is immutable. Unexpected mutation at "${p.tok.kind}" "${p.tok.lit}" in this (excerpt): "...${p.excerpt()}..."')
+	}
 }
 
 // check_explicitly_declared returns an error if `key` has been explicitly declared.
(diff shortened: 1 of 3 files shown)


vlib/toml/parser/parser.v
```diff
@@ -276,6 +278,14 @@ fn todo_msvc_astring2dkey(s []string) DottedKey {
 	return s
 }
 
+// check_immutable returns an error if `key` has been declared as immutable.
+fn (p &Parser) check_immutable(key DottedKey) ! {
+	if p.immutable.len > 0 && p.immutable.has(key) {
+		return error(@MOD + '.' + @STRUCT + '.' + @FN +
+			' key `${key.str()}` is immutable. Unexpected mutation at "${p.tok.kind}" "${p.tok.lit}" in this (excerpt): "...${p.excerpt()}..."')
```
spytheman, line 285: What test case triggers this error?

vlib/toml/parser/parser.v
```diff
@@ -59,6 +59,8 @@ mut:
 	// The root map (map is called table in TOML world)
 	root_map                          map[string]ast.Value
 	root_map_key                      DottedKey
+	value_is_immutable                bool
+	immutable                         []DottedKey
```
spytheman, line 63: How many members is that field expected to have on average?

vlib/toml/parser/parser.v
larpon: > What test case triggers this error?

Assuming you have `toml-lang/toml-test` checked out into `vlib/toml/tests/testdata/toml_lang/` at `8bdabb2` or above (e.g. `c6a78f1` as in this PR).

Then the following test triggers the error:
`vlib/toml/tests/testdata/toml_lang/tests/invalid/inline-table/overwrite-02.toml`

There was a bunch of `invalid/inline-table/overwrite-*` added in `8bdabb2` so this PR covers those cases.

I could not make it work with the existing `explicit_declared` setup 🙁 
I assume it is because there's duplicate keys involved in some of the tests that it is hard to look up/deal with in `explicit_declared` without some kind of scope info or something 🤷🏻

vlib/toml/parser/parser.v
larpon: Not that many I think.

It is currently used for the inline_table parsing only, which is `map[string]Value` (but only in some cases) in the `Value` type:
```v
pub type Value = Bool
	| Date
	| DateTime
	| Null
	| Number
	| Quoted
	| Time
	| []Value
	| map[string]Value // <- some fraction of these occurrences
```
So it is some fraction of the total values appearing in the document 🤷🏻

vlib/toml/parser/parser.v
spytheman: Thank you.

PR #26069, opened 2025-12-21
builtin: fix some C fn signatures

tankf33der:
This patch affects 6 functions - `calloc()`, `malloc()`, `realloc()`, `mmap()`, `mprotect()`, `aligned_alloc()`. 

I manually compared and verified the signatures of these functions.
I spent 90% of the time selecting the correct type for `offset` from the `mmap()` function.

Please check my PR.

p.s. 
```
$ cat vlib/builtin/allocation.c.v | grep "pub fn malloc" | head -1
pub fn malloc(n isize) &u8 {
```
to be honest, I don't fully understand how this affects this and how `cfns.c.v` and `allocation.c.v` work with each other.


spytheman:
` C.malloc` != V's `malloc()`

PR #26072, opened 2025-12-21
fasthttp: improve minimal server example

kimshrier:
While experimenting with fasthttp on darwin and FreeBSD, the minimal server example, which does function, does not comply with the HTTP protocol exactly.  When using curl to hit the server, it complains about a response using HTTP 0.9.  Adding a status header to the response gets past that hurdle.  However, it then complains about other missing headers and no reply content.

The changes I am proposing make the response compatible with HTTP/1.1 and has enough headers and content to keep curl happy.  This might make it more obvious to people using this module that they are responsible for crafting legitimate responses which includes the headers.  I tried to keep the code minimal while satisfying the protocol constraints.

I do have another question.

I was able to copy fasthttp_darwin.v to fasthttp_freebsd.v and everything works with no changes.  I suspect that the same would work for OpenBSD, NetBSD, and DragonflyBSD.  However, I think having 5 identical copies of the kqueue implementation under os specific file names would not be ideal.

Unfortunately, I have to work with many different languages in my day job and I don't have the proper mindset to really write v code in proper idiomatic v.  I only get to write v code in my spare time which is quite scarce at the moment.

My first idea is to create a fasthttp_default.v file with the kqueue implementation enclosed in:
```
$if darwin || freebsd || openbsd || netbsd || dragonflybsd {
...
}
```
and remove the fasthttp_darwin.v file.  Somehow, that doesn't feel right.

I am open to any suggestions.

### DIFF
--- a/vlib/fasthttp/README.md
+++ b/vlib/fasthttp/README.md
@@ -29,11 +29,26 @@ import fasthttp
 fn handle_request(req fasthttp.HttpRequest) ![]u8 {
 	path := req.buffer[req.path.start..req.path.start + req.path.len].bytestr()
 
+	mut body := ''
+	mut status_line := ''
+
 	if path == '/' {
-		return 'Hello, World!'.bytes()
+		body = 'Hello, World!\n'
+		status_line = 'HTTP/1.1 200 OK'
+	} else {
+		body = '${path} not found\n'
+		status_line = 'HTTP/1.1 404 Not Found'
 	}
 
-	return '404 Not Found'.bytes()
+	headers := [
+		status_line,
+		'Content-Type: text/plain',
+		'Content-Length: ${body.len}',
+		'Connection: close',
+	]
+	header_string := headers.join('\r\n')
+
+	return '${header_string}\r\n\r\n${body}'.bytes()
 }
 
 fn main() {


JalonSolov:
You can make one named `fasthttp_bsd.v`, and that should apply to all the BSD flavors.  If it doesn't, it should, just as there is a `_nix.v` to apply to all unix flavors, etc.

kimshrier:
Thanks for the pointer.  I was unaware of that capability.  I knew about `_nix.v` but I didn't know what `_bsd.v` would get me.

enghitalo:
About the second: If the kqueue can handle it, I don't see why not.

enghitalo:
Regarding the documentation and examples, I think it's a great idea to change that to be closer to the standards. I would only use string builder to avoid low-performance concatenation and other things.

vlib/fasthttp/README.md
```diff
@@ -29,11 +29,26 @@ import fasthttp
...
+	} else {
+		body = '${path} not found\n'
+		status_line = 'HTTP/1.1 404 Not Found'
 	}
 
-	return '404 Not Found'.bytes()
+	headers := [
+		status_line,
+		'Content-Type: text/plain',
+		'Content-Length: ${body.len}',
```
enghitalo, line 46: Use string builder

enghitalo (changes requested):
Use string builder instead

spytheman:
I am merging it as it is, since correctness comes first before performance. It is a definite improvement on that front. The examples should be as clear as possible too.

spytheman:
(it is also Christmas, and a major release is coming soon)

PR #26073, opened 2025-12-22
fasthttp: move kqueue implementation from _darwin.v to _bsd.v

kimshrier:
It looks like the kqueue implementation for fasthttp will work on all BSD variants.  I have tested this on darwin and FreeBSD.  The other BSD variants should work, as well.

### DIFF
--- a/vlib/v/pref/os.v
+++ b/vlib/v/pref/os.v
@@ -33,6 +33,7 @@ pub enum OS {
 	wasi    // -b wasm -os wasi
 	raw
 	all
+	bsd // macos freebsd openbsd netbsd dragonfly
 }
 
 // Helper function to convert string names to OS enum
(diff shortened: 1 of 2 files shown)


kimshrier:
I am looking at `vlib/v/pref/should_compile.v` and it seems that the function `is_target_of` on line 294 is not taking account of the `bsd` target.

Also, the `fname_without_platform_postfix` function on line 122 doesn't consider 'bsd' either.

Is this where I should be looking to fix the CI test failures?

JalonSolov:
Unfortunately, it appears that `_bsd` is not supported, though it should be.  There are definitely common things amongst all the different bsd flavors, just as there are common things between all the linux flavors (which is why `_nix` is supported).

kimshrier:
Looks like my latest change made things worse.

kimshrier:
I won't have time to look at this for several days.  I'll close this for now and look into it when I can.

Issue #26049, opened 2025-12-20
Incorrect function reference captured in closure leads to crash


Yv-vY:
This bug follows on https://github.com/vlang/v/issues/25940

<details>
<summary>V version: V 0.4.12 60f7c81, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 caf7d413ad1a68aafc20e2a2786d08c30a3964d7.60f7c81
|:-------------------|:-------------------
|OS                  |linux, Linux version 6.17.12-300.fc43.x86_64 (mockbuild@008798550d8d49b8909ac02baf9be783) (gcc (GCC) 15.2.1 20251111 (Red Hat 15.2.1-4), GNU ld version 2.45.1-1.fc43) #1 SMP PREEMPT_DYNAMIC Sat Dec 13 05:06:24 UTC 2025
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 7800X3D 8-Core Processor
|Memory              |25.77GB/30.94GB
|                    |
|V executable        |/home/yann/Apps/v/v
|V last modified time|2025-12-20 09:31:02
|                    |
|V home dir          |OK, value: /home/yann/Apps/v
|VMODULES            |OK, value: /home/yann/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/yann/devel/perso/v/bug_function_ref_closure
|                    |
|Git version         |git version 2.52.0
|V git status        |weekly.2025.43-352-g60f7c815
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20251211 (Red Hat 15.2.1-5)
|gcc version         |gcc (GCC) 15.2.1 20251211 (Red Hat 15.2.1-5)
|clang version       |clang version 21.1.7 (Fedora 21.1.7-1.fc43)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.42

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  main.v && main`
```v
module main

fn f_a() {
	println('a')
}

fn main() {
	f := f_a
	ref := &f
	println('ref=0x${ptr_str(ref)}')

[...]

PR #26075, opened 2025-12-22
cgen: fix fn_var_signature() support nr_muls

kbkpbot:
Fix issue #26049

original code only support gen fn sig :
```c
void (*var_name) (int, string)
```
and now we can support fn sig such as :
```c
void (*****var_name) (int, string)
```


JalonSolov:
Crash and burn waiting to happen...

PR #26078, opened 2025-12-22
slow_tests: update inline asm tests for i386

tankf33der:
Cross-checked, verified, and brought the tests for `i386 inline asm` to a more realistic form.
Now all ok, works under gcc and clang.

### DIFF
--- a/vlib/v/slow_tests/assembly/asm_test.i386.v
+++ b/vlib/v/slow_tests/assembly/asm_test.i386.v
@@ -1,5 +1,4 @@
 // vtest build: !msvc
-import v.tests.assembly.util
 
 fn test_inline_asm() {
 	a, mut b := 10, 0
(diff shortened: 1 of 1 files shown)


spytheman:
The test passed for me too:
```
#0 13:04:00 ^ i686_asm ~/v>./v -gc none -m32 -cc clang vlib/v/slow_tests/assembly/asm_test.i386.v
#0 13:04:07 ^ i686_asm ~/v>
#0 13:04:09 ^ i686_asm ~/v>./v -gc none -m32 -cc gcc vlib/v/slow_tests/assembly/asm_test.i386.v
#0 13:04:15 ^ i686_asm ~/v>
```

The CI will skip it for all CI jobs afaik; I'll have to think of some solution for it, to guarantee it is tested on at least one.

PR #26081, opened 2025-12-22
toml: reduce exceptions in v1.0.0 tests, check February leap validity, check `CRLF` validity

larpon:
I'm trying to prepare a bit for a change/addition in the official test suite where tests are split per version (1.0.0 and 1.1.0) and stumbled across some things in our exception lists that we do not need anymore, have fixed, or has been moved/renamed in the test suite. We should probably switch to test against the suites [list files](https://github.com/toml-lang/toml-test/tree/main?tab=readme-ov-file#usage-without-toml-test-binary) to better know if we are testing against v1.0.0 or v1.1.0:

For v1.0.0 we should use:
https://github.com/toml-lang/toml-test/blob/main/tests/files-toml-1.0.0

... and for v.1.1.0:
https://github.com/toml-lang/toml-test/blob/main/tests/files-toml-1.1.0

I'm still pondering/working on how to change our tests in the least intrusive way - so this PR helps by reducing some "noise" 😅 

**EDIT**
Fun fact, there's a NET zero in line change but more tests are passing:
image


spytheman:
Thank you @larpon 🙇🏻 .

PR #26086, opened 2025-12-22
transformer: fixes for `-new-transformer`

Eliyaan:
Continuing to fix `-new-transformer`.
More tests from `v -new-transformer self` ` v -new-transformer test-self` are passing. 
(43 failing and 27 skipped on my machine, for more than 90 failing before)

### DIFF
--- a/vlib/v/transformer/transformer.v
+++ b/vlib/v/transformer/transformer.v
@@ -10,10 +10,10 @@ import v.util
 pub struct Transformer {
 	pref &pref.Preferences
 pub mut:
-	index   &IndexState
-	table   &ast.Table = unsafe { nil }
-	file    &ast.File  = unsafe { nil }
-	checker bool // is the checker transformer, set by the checker
+	index                &IndexState
+	table                &ast.Table = unsafe { nil }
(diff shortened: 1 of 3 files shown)


vlib/v/transformer/transformer.v
```diff
@@ -10,9 +10,10 @@ import v.util
 pub struct Transformer {
 	pref &pref.Preferences
 pub mut:
-	index &IndexState
-	table &ast.Table = unsafe { nil }
-	file  &ast.File  = unsafe { nil }
+	index   &IndexState
+	table   &ast.Table = unsafe { nil }
+	file    &ast.File  = unsafe { nil }
+	checker bool // is the checker transformer, set by the checker
```
spytheman: imho name this after the operation it is doing, and not the place where it is set; i.e. `skip_array_transform` instead of `checker` - it would be clearer why is it used, and if it can be used in other places that want (or not) that behavior too

Eliyaan:
For the moment it does not seem worth it as it is not used and it not planned to be used alone (before becoming the default).

Issue #25969, opened 2025-12-14
Issue building with -autofree when using sqlite


bradcypert:
<details>
<summary>V version: V 0.4.12 dc0b0f5, press to see full `v doctor` output</summary>

|V full version      |V 0.4.12 90fa8bc56fd4c3870a9041e39fc156bfacdcd1b1.dc0b0f5
|:-------------------|:-------------------
|OS                  |linux, Linux version 6.17.9-arch1-1 (linux@archlinux) (gcc (GCC) 15.2.1 20251112, GNU ld (GNU Binutils) 2.45.1) #1 SMP PREEMPT_DYNAMIC Mon, 24 Nov 2025 15:21:09 +0000
|Processor           |24 cpus, 64bit, little endian, AMD Ryzen AI 9 HX 370 w/ Radeon 890M
|Memory              |43.42GB/62.08GB
|                    |
|V executable        |/home/brad/Projects/v/v
|V last modified time|2025-12-14 15:50:46
|                    |
|V home dir          |OK, value: /home/brad/Projects/v
|VMODULES            |OK, value: /home/brad/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/brad/Projects/vtask
|                    |
|Git version         |git version 2.52.0
|V git status        |dc0b0f50
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20251112
|gcc version         |gcc (GCC) 15.2.1 20251112
|clang version       |clang version 21.1.6
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.42

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg -autofree main.v && /home/brad/Projects/vtask/main`
```v
module main

import veb
import db.sqlite

const port = 8082

struct State {
mut:
	db sqlite.DB
}

pub struct App {
	state shared State
}

@[table: 'tasks']
struct Task {
pub:
	name string @[sql: 'name'; sql_type: 'TEXT']
	description string @[sql: 'description'; sql_type: 'TEXT']
	done bool @[default: false; sql_type: 'BOOLEAN']
mut:
	id int @[immutable; primary; sql: 'id'; sql_type: 'TEXT'; unique]
}

struct Context {
[...]

PR #26089, opened 2025-12-22
autofree: fix detection for is_or with block style

bradcypert:
Fix (#25969) autofree mis-detection so any assignment whose RHS contains a block-style or {} such as sql, selectors, prefix expressions, etc. is marked `is_or`, preventing premature frees, and add a regression test covering ORM selects under -autofree.

Additionally, I noticed an issue with double frees after getting the first step resolved, so I went ahead and fixed that, as well, so there's also a fix that will prevent double-frees during table creation by clearing unique_fields after transferring ownership into fs in orm_table_gen, and extend the regression test schema with a @[unique] field to exercise that path. (just trying to get the example to run, but unfortunately, there are still more issues, though it does actually boot the server in the original issue after my fixes at least -- and the fix here is simpler so it captures working code after these fixes).

Verified by rebuilding v, rerunning the ORM regression program (with and without -autofree), and running the new test file directly.

I'm still a bit new to some of the intricacies of V, so if this isn't tested appropriately or is missing the mark, please let me know.

### DIFF
--- a/vlib/v/gen/c/coutput_test.v
+++ b/vlib/v/gen/c/coutput_test.v
@@ -245,7 +245,13 @@ pub fn get_file_options(file string) FileOptions {
 	return res
 }
 
+const github_job = os.getenv('GITHUB_JOB')
+
 fn should_skip(relpath string) bool {
+	if github_job == 'docker-ubuntu-musl' && relpath.ends_with('autofree_sql_or_block.vv') {
+		eprintln('> skipping ${relpath} on docker-ubuntu-musl, since it uses db.sqlite, and its headers are not available to the C compiler in that environment')
+		return true
+	}
 	if user_os == 'windows' {
 		if relpath.contains('_nix.vv') {
 			eprintln('> skipping ${relpath} on windows')


vlib/v/parser/assign.v
```diff
@@ -239,16 +239,14 @@ fn (mut p Parser) partial_assign_stmt(left []ast.Expr) ast.Stmt {
...
+					if p.pref.autofree && right.len > 0 {
+						expr_for_or := if v.expr !is ast.EmptyExpr {
+							v.expr
+						} else {
+							right[0]
+						}
+						if expr_has_block_or(expr_for_or) {
+							v.is_or = true
 						}
 					}
```
spytheman, line 251: that whole if does not seem right to be in the parser, it may be better to move it in the checker instead

spytheman: however it is out of scope of the PR too 🤔

spytheman: @medvednikov what do you think?

bradcypert:
@spytheman delayed response? I only opened the PR less than a day ago haha. No need to apologize, you are very prompt at reviewing submissions! 😄

PR #26092, opened 2025-12-23
toml: remove special cases for `jq` normalization

larpon:
Not sure if these will trigger the CI anymore (they do not trigger anything locally with `jq` v1.8.1), let's see 🤷🏻


spytheman:
Thank you @larpon 🙇🏻 .

PR #26094, opened 2025-12-23
toml: switch to test against entries in `files-toml-1.0.0` for `toml-lang` tests

larpon:
This PR change the `toml_lang_test.v` to use the official test file list described here: https://github.com/toml-lang/toml-test/tree/main?tab=readme-ov-file#usage-without-toml-test-binary

This reliefs us from making distinctions between TOML v1.0.0 and v1.1.0 tests when testing our parser.

### DIFF
--- a/vlib/toml/tests/toml_lang_test.v
+++ b/vlib/toml/tests/toml_lang_test.v
@@ -8,8 +8,7 @@ import toml
 import toml.ast
 import x.json2
 
-const test_root = os.join_path(@VROOT, 'vlib', 'toml', 'tests', 'testdata', 'toml_lang',
-	'tests')
+const test_root = os.join_path(@VROOT, 'vlib/toml/tests/testdata/toml_lang/tests')
 const test_files_file = os.join_path(test_root, 'files-toml-1.0.0')
 
 const hide_oks = os.getenv('VTEST_HIDE_OK') == '1'


vlib/toml/tests/toml_lang_test.v
```diff
@@ -8,6 +8,10 @@ import toml
 import toml.ast
 import x.json2
 
+const test_root = os.join_path(@VROOT, 'vlib', 'toml', 'tests', 'testdata', 'toml_lang',
+	'tests')
```
spytheman:
```suggestion
const test_root = os.join_path(@VROOT, 'vlib/toml/tests/testdata/toml_lang/tests')
```

spytheman: (it makes it a bit easier to search/copy/paste later)

larpon: Indeed it does. What happens on Windows when paths are mixed `\\` and `/`?
Maybe we only test `toml` on Unix based so it does not matter currently 🙂

spytheman: the `\\` -> `/` replacements should take care of that for the comparisons/reports; functions like os.walk_ext do not care about the mix (windows itself also usually does not care)

spytheman: (msvc sometimes does, but afaik that is not directly related to the `toml` module and its tests)

Issue #25968, opened 2025-12-14
function returning both result and fixed array syntax `!` collision


jorgeluismireles:
### Describe the bug

A function returning both result and fixed array can't use  as last return `!` since fixed array values won't be correct. Omitting `!` is a workaround working by now, though is confusing and is hard to debug lines like this.

### Reproduction Steps

```v
import strconv

pub fn valid_triangle(a u8, b u8, c u8) ![3]u8 {
	if int(a) + int(b) <= int(c) {
		return error('Invalid: a + b <= c')
	}
	if int(b) + int(c) <= int(a) {
		return error('Invalid: b + c <= a')
	}
	if int(c) + int(a) <= int(b) {
		return error('Invalid: c + a <= b')
	}
	return [a, b, c]!
}

pub fn triangle_from_string(sides string) ![3]u8 {
	s := sides.split(',')
	if s.len != 3 {
		return error('Invalid: number of sides')
	}
	a := strconv.atou8(s[0])!
	b := strconv.atou8(s[1])!
	c := strconv.atou8(s[2])!
	// warning: don't append `!` to next call 
	// will confuse with fixed and return `garbage`.
	return valid_triangle(a, b, c)!
}

fn main() {
	dump(triangle_from_string('3,4,5')!)
}
```

### Expected Behavior

```
[code.v:30] triangle_from_string('3,4,5')!: [3,4,5]
```

### Current Behavior

```
[code.v:30] triangle_from_string('3,4,5')!: [192, 0, 0]
```
where 192 is random

### Additional Information/Context

A workaround is omit `!` in last line of method `triangle_from_string`
```diff
-	return valid_triangle(a, b, c)!
+	return valid_triangle(a, b, c)
```
### V version

V 0.4.12 dc0b0f5

### Environment details (OS name and version, etc.)

https://play.vlang.io/p/a9ae34e637
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #26095, opened 2025-12-23
cgen: shorten option fn return non-option in `return stmt`

Krchi:
Reduce type handling in return_stmt.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -2435,25 +2435,25 @@ fn (mut g Gen) expr_with_tmp_var(expr ast.Expr, expr_typ ast.Type, ret_typ ast.T
 					&& (expr.right as ast.StructInit).init_fields.len == 0 {
 					g.write('builtin___option_none(&(${styp}[]) { ')
 				} else if final_expr_sym.kind == .array_fixed {
+					assign_op := g.assign_op
+					defer(fn) {
+						g.assign_op = assign_op
+					}
+					g.assign_op = .unknown
 					expr_is_fixed_array_var = true
 					info := final_expr_sym.array_fixed_info()
 					mut no_cast := false
 					if expr in [ast.CastExpr, ast.CallExpr, ast.Ident, ast.SelectorExpr] {
 						no_cast = true
 					}
-					assign_op := g.assign_op
-					defer(fn) {
-						g.assign_op = assign_op
-					}
-					g.assign_op = .unknown
 					elem_sym := g.table.sym(info.elem_type)
 					if elem_sym.kind == .struct {
 						expr_is_fixed_array_var = false
 						g.write('builtin___option_ok(&(${styp}[]) { ')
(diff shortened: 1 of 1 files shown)


vlib/v/gen/c/cgen.v
```diff
@@ -2434,10 +2434,27 @@ fn (mut g Gen) expr_with_tmp_var(expr ast.Expr, expr_typ ast.Type, ret_typ ast.T
 					&& expr.right is ast.StructInit
 					&& (expr.right as ast.StructInit).init_fields.len == 0 {
 					g.write('builtin___option_none(&(${styp}[]) { ')
-				} else if expr in [ast.Ident, ast.SelectorExpr]
-					&& final_expr_sym.kind == .array_fixed {
+				} else if final_expr_sym.kind == .array_fixed {
+					assign_op := g.assign_op
+					defer(fn) {
+						g.assign_op = assign_op
+					}
```
spytheman, line 2441: Why is this `defer(fn) {` and not just `defer {` ?
The variable that saves the previous state, is valid only in this scope.

spytheman:
Please add a test case too, or describe in more details the why of your PR, and link issues if possible.

One possibility for the test (given that cgen is most affected), is to add a pair of .vv/.c.must_have files in 
`vlib/v/gen/c/testdata/` , see the https://github.com/vlang/v/blob/master/TESTS.md#v-vlibvgenccoutput_testv .

Krchi:
> Please add a test case too, or describe in more details the why of your PR, and link issues if possible.
> 
> One possibility for the test (given that cgen is most affected), is to add a pair of .vv/.c.must_have files in `vlib/v/gen/c/testdata/` , see the https://github.com/vlang/v/blob/master/TESTS.md#v-vlibvgenccoutput_testv .

No issues at the moment, it’s just for consistency, and I can’t guarantee coverage of all cases yet, but it passed checks now, if there are have new problem with fn return option fixed array, we can do some changes in `expr_with_tmp_var`, as `expr_with_tmp_var`'s doc says: expr_with_tmp_var is used in assign expr to option or result type, applicable to situations where the expr_typ does not have option and result. I'm just complete the functionality that should have been there, I want to change the fn return result fixed array in return_stmt too, may relate to #25968

vlib/v/gen/c/cgen.v
Krchi: > Why is this `defer(fn) {` and not just `defer {` ? The variable that saves the previous state, is valid only in this scope.

I want change assign_op only in fixed array cond, but assign_op must restore after expr，cant resotre in this scope

spytheman: I've tried to use the normal scoped defer, and that leads to a failure for `vlib/v/tests/structs/struct_field_init_with_fixed_array_opt_test.v` .

spytheman (approved):
Excellent work.
Thank you @Krchi 🙇🏻 .

PR #26100, opened 2025-12-24
db.sqlite: support ability to process multiple sql calls with parameters of data

quaesitor-scientiam:
sqlite.c.v: Added sqlite reset() call and allows exec_param_many() multiple arrays of data and logic to process that.


spytheman:
I think adding a separate new API, that accepts just `[][]string`, leaving the old one to accept just a `[]string` parameter will be clearer, and less breaking, than changing the old one to accept a sum type, just to immediately do a type switch inside it.

quaesitor-scientiam:
> I think adding a separate new API, that accepts just `[][]string`, leaving the old one to accept just a `[]string` parameter will be clearer, and less breaking, than changing the old one to accept a sum type, just to immediately do a type switch inside it.

`exec_param_many2()` perhaps?   I thought about using a generic syntax but thought that was a breaking change as well.

quaesitor-scientiam:
> I think adding a separate new API, that accepts just `[][]string`, leaving the old one to accept just a `[]string` parameter will be clearer, and less breaking, than changing the old one to accept a sum type, just to immediately do a type switch inside it.

Ran sqlite tests with no changes no breaks found 
```
PS S:\repo\vlang2> v test vlib/db/sqlite
 OK    [1/6] C:   776.6 ms, R:   379.223 ms vlib/db/sqlite/sqlite_f32_test.v
 OK    [2/6] C:   776.7 ms, R:   379.221 ms vlib/db/sqlite/sqlite_comptime_field_test.v
 OK    [3/6] C:   777.5 ms, R:   378.881 ms vlib/db/sqlite/parent_child_test.v
 OK    [4/6] C:   772.2 ms, R:   401.230 ms vlib/db/sqlite/sqlite_orm_test.v
 OK    [5/6] C:   772.3 ms, R:   400.798 ms vlib/db/sqlite/sqlite_test.v
 OK    [6/6] C:   771.9 ms, R:   401.556 ms vlib/db/sqlite/sqlite_vfs_lowlevel_test.v
Summary for all V _test.v files: 6 passed, 6 total. Elapsed time: 1338 ms, on 6 parallel jobs. Comptime: 4647 ms. Runtime: 2340 ms.
```

spytheman (approved):
Thank you @quaesitor-scientiam 🙇🏻 .

cmd/tools/modules/testing/common.v
```diff
@@ -60,8 +60,7 @@ pub const is_ruby_present = os.execute('ruby --version').exit_code == 0
 pub const is_python_present = os.execute('python --version').exit_code == 0
 	&& os.execute('pkg-config python3 --libs').exit_code == 0
 
-pub const is_sqlite3_present = os.execute('sqlite3 --version').exit_code == 0
-	&& os.execute('pkg-config sqlite3 --libs').exit_code == 0
+pub const is_sqlite3_present = os.exists(@VEXEROOT + '/thirdparty/sqlite/sqlite3.c')
```
spytheman, line 63: that works only on windows :-| ... I'll edit it later

quaesitor-scientiam: I thought the new `.vsh` script applied to all platforms.  My mistake

PR #26101, opened 2025-12-24
toml, tests: support v2.0.0 of the test-suite at toml-lang/toml-test@229ce2e

larpon:
This PR "fast forwards" to handle toml-lang/toml-test@229ce2e which is the latest commit as of today.

A version 2.0 of the suite was released last week so hopefully the setup should hold up for a while.

This comes at the small "expense" of having to add a few new entries to the `invalid_exceptions` list (I need more time/motivation before fixing those) where a few entries was also renamed to use `0` (zero) prefixes for numbers in the file names.

I've also included a small fix for detecting invalid parsed time, because... well... it's Christmas time and I felt joy doing it 🙂🎄🎅🏻 (fixes concrete test: `vlib/toml/tests/testdata/toml_lang/tests/invalid/datetime/second-trailing-dot.toml`)

Merry Christmas to everyone on the V team, that celebrates Christmas.


spytheman (approved):
Thank you @larpon .
Excellent work.

Merry Christmas to you and your family.

Issue #26091, opened 2025-12-23
[new_veb]  net.http.HeaderKeyError when parsing JSON request body


Avey777:
### Describe the bug

[veb] Failed to parse request: net.http.HeaderKeyError: Invalid header key: '{"username"'

### Reproduction Steps

```v
#!/usr/bin/env -S v run

import veb
import x.json2 as json

struct Context {
	veb.Context
}

struct App {
	veb.Middleware[Context]
}

@["/index";get;post]
fn (mut app App) index(mut ctx Context) veb.Result {
  req := json.decode[UserReq](ctx.req.data) or {
		return ctx.json('json failed')
	}

	return ctx.json('index succcess ${req}')
}

fn main() {
	port := 9008
	mut app := &App{}
	veb.run[App, Context](mut app, port)
}


pub struct UserReq {
	username     string    @[json: 'username']
}

```
`v -d new_veb run .`

`curl -X POST http://localhost:9008/index   -H "Content-Type: application/json"   -d '{"username": "test"}'   --trace-ascii /dev/stdout`

### Expected Behavior

no failed

### Current Behavior

```
[veb] Running multi-threaded app on http://localhost:9008/
listening on http://localhost:9008/
[veb] Failed to parse request: net.http.HeaderKeyError: Invalid header key: '{"username"'
s=
POST /index HTTP/1.1
Host: localhost:9008
User-Agent: curl/8.11.0
Accept: */*
Content-Type: application/json
Content-Length: 20

{"username": "test"}
==============

```

### V version

V 0.4.12 5d6f1a9

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 5d6f1a99b6ebe6e09d27be3620092035bfcc8e32
|:-------------------|:-------------------
|OS                  |linux, Deepin 25
|Processor           |6 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
|Memory              |0.27GB/7.6GB
|                    |
|V executable        |/home/Jengro/.vmr/versions/v_versions/v_latest/v
|V last modified time|2025-12-23 03:35:25
|                    |
|V home dir          |OK, value: /home/Jengro/.vmr/versions/v_versions/v_latest
|VMODULES            |OK, value: /home/Jengro/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/Jengro
|                    |
|Git version         |git version 2.50.1
[...]

PR #26104, opened 2025-12-24
fasthttp: expand http request parser

enghitalo:
support `header_fields` and `body` on HTTP Parser
related to #26091 part1

### DIFF
--- a/vlib/fasthttp/fasthttp.v
+++ b/vlib/fasthttp/fasthttp.v
@@ -48,6 +48,7 @@ pub:
 	len   int
 }
 
+// HttpRequest represents an HTTP request.
 // TODO make fields immutable
 pub struct HttpRequest {
 pub mut:
--- a/vlib/fasthttp/request_parser.v
+++ b/vlib/fasthttp/request_parser.v
@@ -16,6 +16,7 @@ fn find_byte(buf &u8, len int, c u8) int {
 	}
 }
 
+// parse_http1_request_line parses the request line of an HTTP/1.1 request.
 // spec: https://datatracker.ietf.org/doc/rfc9112/
 // request-line is the start-line for for requests
 // According to RFC 9112, the request line is structured as:
@@ -26,6 +27,7 @@ fn find_byte(buf &u8, len int, c u8) int {
 // REQUEST-TARGET is the path or resource being requested
 // HTTP-VERSION is the version of HTTP being used (e.g., HTTP/1.1)
 // CRLF is a carriage return followed by a line feed
+// returns the position after the CRLF on success
 @[direct_array_access]
 pub fn parse_http1_request_line(mut req HttpRequest) !int {
 	unsafe {


vlib/fasthttp/request_parser.v
```diff
@@ -0,0 +1,145 @@
...
+	unsafe {
+		p := C.memchr(buf, c, len)
+		if p == voidptr(nil) {
+			return -1
+		}
+		return int(&u8(p) - buf)
+	}
+}
+
+// spec: https://datatracker.ietf.org/doc/rfc9112/
```
spytheman, line 20: doc comments for public APIs should start with the name of the documented function

vlib/fasthttp/fasthttp.v
```diff
@@ -48,12 +48,15 @@ pub:
 	len   int
 }
 
+// TODO make fields immutable
```
spytheman, line 52: Please add a proper doc comment, starting with the name of the entity.

vlib/fasthttp/request_parser.v
```diff
@@ -0,0 +1,145 @@
...
+		if len < 12 {
+			return error('Too short')
+		}
+
+		// METHOD
+		pos1 := find_byte(buf, len, empty_space)
+		if pos1 <= 0 {
+			return error('Invalid method')
+		}
+		req.method = Slice{0, pos1}
```
spytheman, line 46: where do you check that the http method name is not some XYZD instead of the subset that you support (presumably one of GET, HEAD, POST, PUT, DELETE, CONNECT, OPTIONS, TRACE)?

vlib/fasthttp/request_parser.v
```diff
@@ -0,0 +1,145 @@
...
+// According to RFC 9112, the request line is structured as:
+// `request-line   = method SP request-target SP HTTP-version`
+// where:
+// METHOD is the HTTP method (e.g., GET, POST)
+// SP is a single space character
+// REQUEST-TARGET is the path or resource being requested
+// HTTP-VERSION is the version of HTTP being used (e.g., HTTP/1.1)
+// CRLF is a carriage return followed by a line feed
+@[direct_array_access]
+pub fn parse_http1_request_line(mut req HttpRequest) !int {
```
spytheman, line 32: Document also what is the return value here; the single `int` is not intuitive for a fn named `parse_` .

vlib/fasthttp/request_parser.v
enghitalo: This is outside the scope of [RFC9112](https://datatracker.ietf.org/doc/rfc9112/)

method and some others HTTP elements are designed specifically to be extensible.

I recommend that this type of check be done outside of FastHTTP.

spytheman: ok, although that seems a bit unexpected ... can we at least document that the function accepts arbitrary method names? 

Afaik for several decades, there were no new methods added for normal web browsers/servers...

spytheman (approved):
Thank you @enghitalo 🙇🏻 .

PR #26107, opened 2025-12-24
string: speed up is_pure_ascii

enghitalo:


enghitalo:
@spytheman I don't managed to reproduce the sanitize fail in my PC neither gcc nor clang

spytheman:
Well, it is clearly reproducible repeatedly on the CI 🤷🏻‍♂️ .

Please experiment on your **private fork** though, if you decide to pursue it further.

The CI on the main repo is clogged, and I prefer it, to stay available as much as possible for actual improvements and fixes, instead of potential "it could be cool for microbenchmarks, but it is not really working yet" experiments, that hinder others in the process :-| .

PR #26115, opened 2025-12-25
checker: add gotodef for more node types

dy-tea:
This is in an effort to make vls go to definition better (currently it only supports functions).

I tested these changes with vls on zed (on vls latest git) which worked quite well. 

Here I add go to definition for structs, struct members, enums, enum members and sumtypes. See the `goto_def_test.v` file for all supported cases.


spytheman (approved):
Thank you @dy-tea .

PR #26123, opened 2025-12-25
builtin: make `string.index_` public, to simplify code using a redundant `x := s.index(sub) or { -1 }` pattern

spytheman:


Delta456:
Can we come up with a better name?

spytheman:
There is already a precedent - `string.index_after_`. Imho there is no need to rename it.

JalonSolov:
Why does `string.index_after_` have a trailing `_`?  And why would `string.index_` need the trailing `_`?

spytheman:
because there are shorter versions, that return option values instead

JalonSolov:
I would rather see `string.index_after_no_option` and `string.index_no_option` or some similar explicit name.  Having just a trailing `_` would make me have to look up both to see what the difference was... just like I asked here.

spytheman:
I would rather just merge it as it is, right after the full CI passes, sticking to the existing convention, and postpone the renaming till later.

JalonSolov:
Renaming is always harder "later", but... <shrug>

spytheman:
I understand that. If the proposed naming scheme is accepted, a rename will be inevitable.
If it is not, this PR is enough.
In both cases, the current state on master is worse (there is no way to get the index without using an option, and the overhead for that is important in some situations).

-> The master branch will be improved by merging this as it is, given that the CI passed.

PR #26125, opened 2025-12-25
toml: fix remaining valid `key/*` exceptions

larpon:


spytheman (approved):
Excellent, thank you @larpon 🙇🏻 .

PR #26126, opened 2025-12-25
parser: obsolete generic cleanup 1

kbkpbot:
cleanup obsolete generic code in `parser`

### DIFF
--- a/vlib/v/parser/checks.v
+++ b/vlib/v/parser/checks.v
@@ -82,7 +82,7 @@ fn (p &Parser) is_array_type() bool {
 }
 
 fn (mut p Parser) is_following_concrete_types() bool {
-	if !(p.tok.kind in [.lt, .lsbr] && p.tok.is_next_to(p.prev_tok)) {
+	if !(p.tok.kind == .lsbr && p.tok.is_next_to(p.prev_tok)) {
 		return false
 	}
 	mut i := 1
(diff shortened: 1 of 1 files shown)


vlib/v/parser/checks.v
```diff
@@ -150,22 +146,22 @@ fn (p &Parser) is_typename(t token.Token) bool {
...
-//	   otherwise it is not generic because it may be multi-value (e.g. `return f < foo, 0`).
-// 6. `f<mod.Foo>` is same as case 3
-// 7. `f<mod.Foo,` is same as case 5
+// heuristics to detect `func[T]()` from `var [ expr`
+// 1. `f[[]` is generic(e.g. `f[[]int]`) because `var [ []` is invalid
+// 2. `f[map[` is generic(e.g. `f[map[string]string])
+// 3. `f[foo]` is generic because `v1 [ foo ] v2` is invalid syntax
+// 4. `f[foo[bar` is generic when bar is not generic T (f[foo[T](), in contrast, is not generic!)
+// 5. `f[Foo,` is generic when Foo is typename.
+//	   otherwise it is not generic because it may be multi-value (e.g. `return f [ foo, 0`).
```
spytheman, line 155: that line does not make sense now

vlib/v/parser/checks.v
```diff
@@ -150,22 +146,22 @@ fn (p &Parser) is_typename(t token.Token) bool {
...
-// heuristics to detect `func<T>()` from `var < expr`
-// 1. `f<[]` is generic(e.g. `f<[]int>`) because `var < []` is invalid
-// 2. `f<map[` is generic(e.g. `f<map[string]string>)
-// 3. `f<foo>` is generic because `v1 < foo > v2` is invalid syntax
-// 4. `f<foo<bar` is generic when bar is not generic T (f<foo<T>(), in contrast, is not generic!)
-// 5. `f<Foo,` is generic when Foo is typename.
-//	   otherwise it is not generic because it may be multi-value (e.g. `return f < foo, 0`).
-// 6. `f<mod.Foo>` is same as case 3
-// 7. `f<mod.Foo,` is same as case 5
+// heuristics to detect `func[T]()` from `var [ expr`
```
spytheman, line 149: this line may need to be removed/edited, because `var [ expr` is not valid V syntax

spytheman:
@kbkpbot thank you for the cleanup 🙇🏻 . 

It is long due - the <> -> [] change was started in 2022-11-26, so I think the vast majority of V codebases are already updated to the new syntax.

vlib/v/parser/checks.v
spytheman: hm, actually it is - var could be an array value, and `var[expr]` is valid, so it should stay.

vlib/v/parser/checks.v
spytheman: it should probably stay, because of `return var[ expr ], 0` .

JalonSolov, line 155: Unbalanced... looks like it should either be

```suggestion
//	   otherwise it is not generic because it may be multi-value (e.g. `return f ( foo, 0`).
```
or

```suggestion
//	   otherwise it is not generic because it may be multi-value (e.g. `return f [ foo, 0`].
```
or possibly

```suggestion
//	   otherwise it is not generic because it may be multi-value (e.g. `return f ([ foo ], 0`).
```

spytheman: they are partial syntax (prefix) examples of the different lookups the parser has to do to determine if it is dealing with a generic function call or some other kind of expression

PR #26130, opened 2025-12-25
docs: add note about implicit cloning of slices

gechandesu:


doc/docs.md
```diff
@@ -1347,6 +1347,10 @@ println(a) // [0, 1, 2, 3, 4, 5]
 println(b) // [7, 3]
 ```
 
+Note that V by default makes implicit clone of the slice and displays a notice about this.
+So without the .clone() call result of the code above will be the same. Make slice in
+`unsafe {}` block if you want to reuse memory, othervise prefer to use explicit cloning.
```
JalonSolov:
```suggestion
Note that, by default, V makes an implicit clone of the slice and displays a notice about this.
So without the .clone() call the result of the code above will be the same. Make the slice in an
`unsafe {}` block if you want to reuse memory, otherwise use explicit cloning.
```

doc/docs.md
```diff
@@ -1347,6 +1347,10 @@ println(a) // [0, 1, 2, 3, 4, 5]
 println(b) // [7, 3]
 ```
 
+Note that, by default, V makes an implicit clone of the slice and displays a notice about this.
+So without the .clone() call the result of the code above will be the same. Make the slice in an
```
Delta456:
```suggestion
So without the `.clone()` call the result of the code above will be the same. Make the slice in an
```

spytheman (approved):
Thank you @gechandesu 🙇🏻 .

PR #26135, opened 2025-12-26
transformer: add line no for fn with `expand_simple_interpolation`

Krchi:
current cgen lost line no, when debug with it, it will jump to line 1


spytheman:
Thank you @Krchi .

spytheman (approved):

PR #26137, opened 2025-12-26
builtin: add a StringBuilder implementation (part 1, just a copy of strings.Builder, with no aliases and deprecations)

spytheman:


spytheman:
closing for now, until some discovered bugs are fixed

PR #26144, opened 2025-12-26
toml: allow `true` and `false` as keys when parsing root table

larpon:
... which also happens to fix `valid/comment/after-literal-no-ws.toml` 🙂


larpon:
> People these days have no respect for reserved keywords, and want to use them everywhere ... V, now TOML ...



😄

larpon:
I agree though. It is just making everything more complex behind the scenes 🙁

PR #26152, opened 2025-12-26
strings: remove write method of builder

enghitalo:
`write` doesn't return a error


spytheman:
It is there intentionally, for compatibility with the `io.Writer` interface.

enghitalo:
@spytheman Is it ok if I write a `write_buffer` without result?

spytheman:
you can just `<<` to it; it is an alias to `[]u8` so you can do everything you can with that

PR #26153, opened 2025-12-26
transformer: early return to avoid transform getting applied twice

Eliyaan:
Without the return, the tranform gets applied twice.


spytheman (approved):
Thank you @Eliyaan 🙇🏻 .

PR #26155, opened 2025-12-26
time: implement faster and simpler `push_http_header`

enghitalo:


enghitalo:
### Benchmark Source
```v
module main


import time
import benchmark

const time_to_test = time.Time{
	year:   1980
	month:  7
	day:    11
	hour:   21
	minute: 23
	second: 42
}

fn main() {
	println('Testing time formatting performance...')

	mut http1_1_buffer1 := 'HTTP/1.1 200 OK\r\nDate: '.bytes()
	mut http1_1_buffer2 := 'HTTP/1.1 200 OK\r\nDate: '.bytes()

	mut b := benchmark.start()

	for _ in 0 .. 1_000_000 {
		a := time_to_test.http_header_string()
		unsafe {
			http1_1_buffer1.push_many(a.str, a.len)
		}
	}
	b.measure( 'http_header_string 1_000_000 times')

	for _ in 0 .. 1_000_000 {
		time_to_test.push_to_http_header(mut http1_1_buffer2)
	}
	b.measure( 'push_to_http_header 1_000_000 times')
}
```

enghitalo:
@spytheman I know that the changes to `pub fn (a array) push_many(vals voidptr, size int) {` are unrelated, but I only implemented the minimum to prevent CI from failing. I hope you understand. I can open an issue later reporting that `push_many` needs to be completed for JS since I don't have the expertise to do that.

spytheman:
@enghitalo it is not a problem - I understand why it is there.

JalonSolov:
If the target is ES6 or later, something as simple as
```javascript
arr.push(...vals)
```
should work fine.

That would be ignoring the `size` parameter, though.

If the `size` is the number that's supposed to be pushed (even if the array is larger), you could do something like
```javascript
arr.push(...vals.slice(0, size))
```

(that's what I read on stack overflow, at least...)

vlib/builtin/js/array.js.v
```diff
@@ -414,6 +414,12 @@ pub fn (a array) to_number_array() JS.Array {
 	return tmp
 }
 
+// push_many - appends multiple values to the end of the array.
+pub fn (a array) push_many(vals voidptr, size int) {
+	// TODO: implement variadic push_many in JS
+	#a.val.arr.len.val += size
```
spytheman: this should either call panic('TODO'), or
`a.insert_many(a.len, vals, size)` which is already implemented above

enghitalo: ok

PR #26157, opened 2025-12-26
checker: gotodef for struct init and sumtype rhs

dy-tea:
I split up the test file into functions so hopefully less positions need to be updated on each gotodef init. There are probably a few more cases I am missing (haven't tested `interface` or channels for example). The things I use most should at least have a gotodef implementation.


martobtw:
thanks mrpenis

Issue #26068, opened 2025-12-21
Full AST cloning for generics and comptime constructs


Eliyaan:
A lot of bugs are related to the complicated handling of generics. 
A proposed solution that came up repeatedly is to fully clone the AST nodes for each concrete type trading a bit of memory overhead for a simpler logic. Everything related to generics would be handled at the AST cloning and none after that.
See the discussion here [26008](https://github.com/vlang/v/pull/26008#issuecomment-3673409151)

Same for comptime constructs (as suggested below)

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

kbkpbot:
I'm also considering whether it's necessary to clone the AST nodes for compile-time constructs like `$if`, `$match`, and `$for`.

Eliyaan:
if they were solved at one early stage too, it would be simpler

kbkpbot:
A possible solution is to introduce a new stage in the builder that transforms all generic and compile-time constructs (such as `$if`, `$match`, `$for`, etc.) into AST nodes with concrete types. This way, subsequent processing stages no longer need to handle generics or compile-time logic.

Eliyaan:
I should have asked the question here instead of opening a discussion (#26175 ).  I did not think about this 😅

kbkpbot:
I have a dream that someday, in backend designs like `cgen`, we won't have to worry about complex issues such as generics, comptime, and the like... simply translating into C code is all that's needed.

Eliyaan:
It would be great. I'll try to add the new stage to solve generics (in a few days)

JalonSolov:
All _any_ of the backends should do is translate the AST into whatever their output type may be.

medvednikov:
This has been implemented in v2, which will be the default by this summer.

PR #26161, opened 2025-12-26
v: add a new v.comptime stage `COMPTIME`, that runs after the checker (`CHECK`), and before the transformer (`TRANSFORM`), to simplify the work that later stages do #26068

Eliyaan:
I'm still very new to the compiler so I'm not very confident in these changes. (feel free to push changes)
See #26068 for the issue talking about this change

The idea behind this is to move all the comptime logic in a new stage so no later stages need to manage this. 
I tried to keep his PR minimal, by introducing the new stage and moving only 1 small comptime  resolution from cgen to the new comptime stage. 

The most of the code in the new comptime stage is a simplified version of the transformer. (without this code the PR is small)

@spytheman @kbkpbot and other contributors what do you think?

### DIFF
--- a/vlib/v/comptime/comptime.v
+++ b/vlib/v/comptime/comptime.v
@@ -137,7 +137,7 @@ type StmtOrExpr = ast.Expr | ast.Stmt
 pub fn (mut c Comptime) expr_stmt(mut node ast.Expr) StmtOrExpr {
 	match mut node {
 		ast.IfExpr {
-			if node.is_comptime { // TODO, move that in expr_stmt
+			if node.is_comptime {
 				if !node.is_expr && !node.has_else && node.branches.len == 1 {
 					if node.branches[0].stmts.len == 0 {
 						// empty ifdef; result of target OS != conditional => skip


JalonSolov:
It's tough to separate these stages, but the compiler will be better for it in the end.

For example:  It would almost be better to check at least some of the comptime stuff before the type resolver, as there is no reason to resolve something that won't be included...

If the parser took care of comptime `$if`, such as `$if windows`, then it would know whether or not to even bother parsing that section depending on the target OS.

However, that starts getting very complex very quickly.  Later stages simply need to know what to keep and what to throw out.

comptime section before checker does make sense, as the checker wouldn't be able to check stuff that's already been removed... same argument as for the parser - if it didn't bother to parse things in the `$else` blocks, they wouldn't be in the AST for the rest of the code to deal with.

It all comes down to which tradeoffs to make in memory vs performance.

Eliyaan:
@kbkpbot 
I agree that your stage order seems better for performance. 
But one thing is bothering me, the code that is not used (either eliminated through comptime or markused) is not getting checked by the checker. So errors could go undetected.

Eliyaan:
the CI failures seem unrelated as it does fail in this commit too https://github.com/vlang/v/actions/runs/20527758186/job/58974092824

spytheman:
`markused` should stay as close to the end (before the backends) as possible imho, because otherwise it would have to predict the work that the next stages do, which is tough and error prone (just like most of the complexity in it, currently is there, because it tries to predict what cgen would do and generate calls for, instead of just walk an AST that is completely processed)

spytheman:
> the code that is not used (either eliminated through comptime or markused) is not getting checked by the checker. So errors could go undetected

There is `-os cross`, which makes all comptime branches to be checked afaik, beside other things, so you should be able to `v -os cross -check .` and be reasonably sure that you get all errors (except perhaps from constructs like $when_first_existing(), $pkgconfig(), $d() and $env(), which depend on the local filesystem/env variables).

spytheman:
> the CI failures seem unrelated as it does fail in this commit too https://github.com/vlang/v/actions/runs/20527758186/job/58974092824

yes, that is fixed on master, but please do not rebase, since the CI is a bit clogged now.

vlib/v/gen/c/cgen.v
spytheman, line 1: Why are there changes here 🤔 ?
imho they are unrelated to the PR?

vlib/v/gen/c/comptime.v
spytheman, line 1: Can it work without this change, so that the PR contains only the introduction of the new stage?
(i.e. if the cgen change is done in a separate cleanup PR)?

vlib/v/gen/c/cgen.v
Eliyaan: The IfExpr that gets solved gets replaced by a ast.Block, but when IfExpr was outside of function (top level declarations) the C compiler did not like the generated brackets.
maybe an example will be clearer than my explaination:
before the change in cgen:
```
{ // those where making the compiler fail
#ifdef ....
#endif
}

int main() {
 ...
}
```

vlib/v/gen/c/comptime.v
Eliyaan: If you think it is better, yes.

spytheman: I am not sure if it is better or not yet, I'll have to test a bit.

PR #26167, opened 2025-12-27
checker,parser: multifile gotodef support for -line-info

dy-tea:
Jumping to a definition in another file of the same module would not always work. Here I add a fix so `-line-info` properly reports this. Small test is also included.

### DIFF
--- a/vlib/v/tests/vls/multifile_gotodef_test.v
+++ b/vlib/v/tests/vls/multifile_gotodef_test.v
@@ -5,6 +5,7 @@ const vroot = @VMODROOT
 const test_dir = os.join_path(vroot, 'vlib', 'v', 'tests', 'vls', 'multifile_gotodef')
 const main_file = os.join_path(test_dir, 'main.v')
 const types_file = os.join_path(test_dir, 'types.v')
+const types_expected = os.join_path('multifile_gotodef', 'types.v')
 
 struct TestCase {
 	name        string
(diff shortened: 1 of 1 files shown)


vlib/v/checker/checker.v
```diff
@@ -389,12 +389,25 @@ pub fn (mut c Checker) check_files(ast_files []&ast.File) {
...
 		mut files_from_main_module := []&ast.File{}
+		// Determine the project directory when using -line-info
+		mut project_dir := ''
+		if c.pref.is_vls && c.pref.line_info != '' {
+			project_dir = if os.is_dir(c.pref.path) {
+				os.real_path(c.pref.path)
+			} else {
+				os.real_path(os.dir(c.pref.linfo.path))
+			}
+		}
```
spytheman: this does not need to be inside the unsafe block

vlib/v/tests/vls/multifile_gotodef_test.v
```diff
@@ -0,0 +1,105 @@
+import os
+import term
+
+const vroot = os.real_path(@VMODROOT)
```
spytheman: afaik `@VMODROOT` is already full real path 🤔

dy-tea: Ah yeah this is redundant.

spytheman:
phew, 2 of the windows jobs finally passed 🥳

PR #26168, opened 2025-12-27
arrays: add compare()

kbkpbot:
compare two arrays
```v
pub fn compare[T](a []T, b []T) bool
```

### DIFF
--- a/vlib/arrays/arrays.v
+++ b/vlib/arrays/arrays.v
@@ -789,10 +789,7 @@ pub fn compare[T](a []T, b []T) bool {
 	if a.len != b.len {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


vlib/arrays/arrays.v
```diff
@@ -781,3 +782,14 @@ pub fn each_indexed[T](a []T, cb fn (i int, e T)) {
 		cb(idx, item)
 	}
 }
+
+// compare compares two arrays.
```
spytheman, line 786: please also describe how it is different than just `a == b`

kbkpbot: sorry, I forgot this :(

PR #26172, opened 2025-12-27
builtin.closure: improove ppc64le closure if PIE

tankf33der:
The lifetime of the ported `ppc64le` architecture in `vlang` is ~250 days.
All this time I remembered that closure crashes on `debian13` (magic?).
When auditing inline assembler support, I ran into this crash again.
`Debian13` - crashes, `Almalinux9` - ok.

The issue turned out to be that Debian13 enabled `PIE` by default.
The compiled code relies on special register `R12`.

Picture below is PIE code. Right from the first lines the code relies on R12 and all calculations are based on it. 
gdb-pie

No R12 register if -no-pie.
gdb-no-pie

I got access to such real hardware `ppc64le` with such Linux distributions:
| CPU | Distro |
|--------|--------|
| Power10 | AlmaLinux 9.7 |
| Power9 | Debian 13.2 |
| Power9 | AlmaLinux 8.10 |
| Power8 | Centos 7 | 

Only `Debian13` do `PIE` by default in this list.

Now `VTEST_ONLY=closure ./v test vlib` works everywhere in my tests.
After the `merge` I will check all servers from scratch again because right now there's a mess there and everything is torn apart.

This issue cost me 40 hours.


spytheman (approved):
Thank you a lot @tankf33der 🙇🏻 .

I am sorry about the time you lost debugging this.

On the bright side, you can think of it as saved time for everyone else that wanted to use V on that CPU, and for most people it would have been a lot longer, or even infinity.

spytheman:
(the CI failure is unrelated and fixed on master)

tankf33der:
`VTEST_ONLY=closure ./v test vlib/`

check report after `merge`:
| Hardware | CPU | Distro |
|--------|--------|--------|
| qemu | Power10 | AlmaLinux 10.1 |
| real | Power10 | AlmaLinux 9.7 |
| real | Power9 | Debian 13.2 |
| real | Power9 | AlmaLinux 8.10 |
| real | Power8 | Centos 7 | 

All ok.

PR #26174, opened 2025-12-27
ast,parser: obsolete generic cleanup 2

kbkpbot:
followup of PR #26126
remove obsolete generic support.


spytheman:
well ... that is a lot less exciting, but still fine - at least the parser will be somewhat smaller

PR #26177, opened 2025-12-27
toml: fix invalid exception by erroring on duplicate keys in inline table

larpon:


spytheman:
The CI failures are unrelated.

PR #26189, opened 2025-12-28
toml: fix remaining invalid exceptions for `table/*`

larpon:
I had to change one of our own test cases in `vlib/toml/tests/nested_test.v`.

Judging from the newly introduced checks it indeed does look like the `[firewall.rules]` should not be re-declared/re-defined since it already gets implicitly defined as a table via the `[firewall.rules.limit]` key above it. The test itself seems to be about testing value retrieval so I think it makes sense to move the definition of `firewall.rules` up (keeping it's indentation which the test also implies to test)

### DIFF
--- a/vlib/toml/parser/parser.v
+++ b/vlib/toml/parser/parser.v
@@ -399,21 +399,24 @@ pub fn (mut p Parser) find_in_table(mut table map[string]ast.Value, key DottedKe
 
 // is_all_tables returns `true` if *all* entries in `dotted_key` (`a.b.c`) are tables (`map[string]ast.Value`), `false` otherwise.
 fn is_all_tables(table map[string]ast.Value, dotted_key DottedKey) bool {
-	util.printdbg(@MOD + '.' + @STRUCT + '.' + @FN, 'locating "${dotted_key}" in map ${ptr_str(table)}')
 	if dotted_key.len == 0 {
 		return false
 	}
-	mut key := dotted_key.clone()
-	k := key.pop_left()
-	if val := table[k] {
-		if val is map[string]ast.Value {
-			if key.len == 0 {
-				return true
+	unsafe {
+		mut t := &table
+		for key in dotted_key {
+			if val := t[key] {
+				if val is map[string]ast.Value {
+					t = &val
+				} else {
+					return false
+				}
+			} else {
+				return false
 			}
-			return is_all_tables(val, key)
 		}
 	}
-	return false
+	return true
 }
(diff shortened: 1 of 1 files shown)


vlib/toml/parser/parser.v
```diff
@@ -397,6 +397,25 @@ pub fn (mut p Parser) find_in_table(mut table map[string]ast.Value, key DottedKe
...
+	mut key := dotted_key.clone()
+	k := key.pop_left()
+	if val := table[k] {
+		if val is map[string]ast.Value {
+			if key.len == 0 {
+				return true
+			}
+			return is_all_tables(val, key)
+		}
+	}
```
spytheman: why not just iterate over the parts of the key, instead of recurse?

vlib/toml/tests/toml_lang_test.v
```diff
@@ -26,10 +26,6 @@ const invalid_exceptions = [
 	'do_not_remove',
 	'string/multiline-escape-space-02.toml',
 	'string/missing-quotes-array.toml',
-	'table/append-with-dotted-keys-05.toml',
-	'table/duplicate-key-03.toml',
-	'table/duplicate-key-10.toml',
-	'table/redefine-02.toml',
```
spytheman, line 32: wow, nice - 4 exceptions eliminated at once!

vlib/toml/parser/parser.v
larpon: ... to me it was the natural path / easier to implement for some reason - I actually started out by copying the implementation of `find_in_table` (which is iterative but `mut`) but ended up with this 🤷🏻 
I can try doing an iterative version

vlib/toml/tests/toml_lang_test.v
larpon: Yup... we're close to ~100% compliance 🙂

vlib/toml/parser/parser.v
larpon: I have pushed [4fef82e](https://github.com/vlang/v/pull/26189/commits/4fef82ef2e2798362779f34ac98e1772a00a7baf) that uses iteration instead

spytheman: I was just curious, since most keys would have probably been short and it would not have mattered one way or another, but thank you 🙇🏻 , since an iterative solution can often scale a bit better (for pathological cases).

larpon:
My pleasure *tips hat*

Issue #18858, opened 2023-07-13
FTP get() doesn't work


lv37:
### Describe the bug

FTP get() breaks the FTP instance

### Expected Behavior

return the content of the files

### Current Behavior

V panic: result not set (Cannot stablish data connection)
v hash: a9a94cf
/tmp/v_1002/main.15824724610745333533.tmp.c:9715: at _v_panic: Backtrace
/tmp/v_1002/main.15824724610745333533.tmp.c:9685: by panic_result_not_set
/tmp/v_1002/main.15824724610745333533.tmp.c:25290: by main__main
/tmp/v_1002/main.15824724610745333533.tmp.c:25754: by main

### Reproduction Steps

```v
module main

import net.ftp

fn main() {
	mut r := ftp.new()
	r.connect(ip)!
	r.login(user, pass)!

	println(r.get('/path/to/any/file')!) // works
	println(r.pwd()!) // should return '/', returns '226 Transfer complete.'
	println(r.get('/path/to/any/file')!) // returns error
}
```

### V version

V 0.4.0 a9a94cf

### Environment details (OS name and version, etc.)

V full version: V 0.4.0 52a055b.a9a94cf
OS: linux, Linux version 6.3.11-200.fc38.x86_64 (mockbuild@375e694195ec4df6b93501f3da7da879) (gcc (GCC) 13.1.1 20230614 (Red Hat 13.1.1-4), GNU ld version 2.39-9.fc38) #1 SMP PREEMPT_DYNAMIC Sun Jul  2 13:17:31 UTC 2023
Processor: 4 cpus, 64bit, little endian, Intel(R) Core(TM) i5-4310U CPU @ 2.00GHz

getwd: /home/user/Projects/v-testing/ftp
vexe: /home/user/.local/lib/v/v
vexe mtime: 2023-07-13 16:47:46

vroot: OK, value: /home/user/.local/lib/v
VMODULES: OK, value: /home/user/.vmodules
VTMP: OK, value: /tmp/v_1002

Git version: git version 2.41.0
Git vroot status: weekly.2023.27-63-ga9a94cfd
.git/config present: true

CC version: cc (GCC) 13.1.1 20230614 (Red Hat 13.1.1-4)
thirdparty/tcc status: thirdparty-linux-amd64 12f392c3

JalonSolov:
Can you give a more complete example?  For instance, what is `ip`?

JalonSolov:
[...]

PR #26194, opened 2025-12-28
net.ftp: fix get() command conversation, add test

tankf33der:
Fixes the #18858 

Currently, after receiving bytes, the `FTP` module immediately closes the connection and does not read the `final operation status`. This is easy to fix.

Important note:
I formally added a test, but it still doesn't execute because it's logically disabled due to the fact that it connects to external resources over the network.

If enable and run test it failed on `master`.

### DIFF
--- a/vlib/net/ftp/ftp.v
+++ b/vlib/net/ftp/ftp.v
@@ -249,6 +249,10 @@ pub fn (mut zftp FTP) get(file string) ![]u8 {
 		return error('Data connection not ready')
 	}
 	blob := dtp.read()!
+	result, _ := zftp.read()!
+	if result != complete {
+		return error('`RETR` not ok')
+	}
 	dtp.close()
 	return blob
 }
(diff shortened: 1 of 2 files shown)


vlib/net/ftp/ftp.v
```diff
@@ -249,6 +249,10 @@ pub fn (mut zftp FTP) get(file string) ![]u8 {
 		return error('Data connection not ready')
 	}
 	blob := dtp.read()!
+	result, _ := zftp.read()!
+	if result != complete {
+		return error('`RETR` not ok')
```
spytheman, line 254: it may be a good idea to first call `dtp.close()` here too, before the return

spytheman: or to use a defer for it, right after the first line of the method

spytheman: anyway, that is unrelated to the PR

spytheman:
(I've tested manually and it works well, after commenting the ftp.redhat.com case, which does not, since the server is probably restricted for the public)

PR #26199, opened 2025-12-29
x.sessions: Verify HMAC signatures when extracting sessions IDs from cookies

### LINKED ISSUE #26198: x.sessions: get_session_id() does not verify HMAC signature, allowing session forgery
### Describe the bug

The `x.sessions` module creates HMAC-signed session cookies (format: `session_id.base64_signature`) using the provided secret, but **fails to verify the signature** when retrieving sessions.

The `get_session_id()` function extracts the session ID by simply splitting the cookie on `.` and returning the first part, completely **ignoring the signature**. This unverified session ID is then used by `get()`, `save()`, `destroy()`, and other functions.

While `verify_session_id()` and `validate_session()` functions exist and correctly implement HMAC verification, they are **never called** internally by the module.

As a result, the secret field is **effectively unused**, for session validation changing the secret _does not_ invalidate existing sessions, and attackers who know or guess a valid session ID can **forge session** cookies _without_ a valid signature.

### Reproduction Steps

``` v
import x.sessions

// Simple in-memory store for demo
struct DemoStore {
mut:
	data map[string]string
}

fn (mut s DemoStore) get(sid string, max_age int) !string {
	if sid in s.data {
		return s.data[sid]
	}
	return error('not found')
}

fn (mut s DemoStore) set(sid string, data string) ! {
	s.data[sid] = data
}

fn (mut s DemoStore) destroy(sid string) ! {
	s.data.delete(sid)
}

fn main() {
	secret1 := 'original-secret-key'.bytes()
	secret2 := 'completely-different-secret'.bytes()

	// Create a signed session ID with secret1
	sid, signed_cookie := sessions.new_session_id(secret1)

	println('=== Session Creation ===')
	println('Secret 1: original-secret-key')
	println('Session ID: ${sid}')
	println('Signed cookie: ${signed_cookie}')

	// Verify with correct secret - should pass
	verified_sid1, valid1 := sessions.verify_session_id(signed_cookie, secret1)
	println('\n=== Verification with CORRECT secret ===')
	println('Valid: ${valid1} (expected: true)')
	println('Session ID: ${verified_sid1}')

	// Verify with wrong secret - should fail
	verified_sid2, valid2 := sessions.verify_session_id(signed_cookie, secret2)
	println('\n=== Verification with WRONG secret ===')
	println('Valid: ${valid2} (expected: false)')
	println('Session ID: ${verified_sid2}')

	// But here's the bug: get_session_id() just splits on '.' and returns first part
	// It NEVER calls verify_session_id()
	println('\n=== The Bug ===')
	println("get_session_id() extracts ID by splitting on '.' without verification:")
	parts := signed_cookie.split('.')
	extracted_sid := parts[0]
	println('Extracted session ID: ${extracted_sid}')
	println('Matches original: ${extracted_sid == sid}')

	println('\n=== Security Impact ===')
	println('1. Attacker can use session ID without valid signature')
	println('2. Changing SECRET does not invalidate existing sessions')
	println('3. Session forgery possible if session ID is known/guessed')

	// Prove it works with wrong signature
	println('\n=== Proof: Forged cookie with invalid signature ===')
	forged_cookie := '${sid}.INVALID_SIGNATURE_HERE'
	forged_parts := forged_cookie.split('.')
	println('Forged cookie: ${forged_cookie}')
	println('Extracted ID (what get_session_id does): ${forged_parts[0]}')
	println('Would access session: ${forged_parts[0] == sid}')
}

```

### Expected Behavior

Would access session: false

### Current Behavior

Would access session: true

### Possible Solution

The `get_session_id()` function at line should call `verify_session_id(cookie, s.secret)` and return `none` if validation fails, instead of blindly splitting the cookie on `.` and returning the first part. This ensures all dependent functions (`get()`, `save()`, `destroy()`) reject forged or tampered cookies.

_No response_

### V version

V 0.4.12 0b0f380

```
V full version: V 0.4.12 9a73766f5afd8f4808964c9b48280abf06c4b73b.0b0f380
OS: Linux
```
> [!NOTE]
>

fleximus:
Now verifies the HMAC when extracting session IDs from cookies.

New tests in vlib/x/sessions/tests/session_test.v:
  - test_forged_signature_rejected - verifies forged signatures are rejected
  - test_wrong_secret_rejected     - verifies wrong secrets are rejected
  - test_malformed_cookie_rejected - verifies malformed cookies are rejected

Fixes #26198

### DIFF
--- a/vlib/x/sessions/tests/session_test.v
+++ b/vlib/x/sessions/tests/session_test.v
@@ -19,7 +19,7 @@ fn test_session_id() {
 	verified_sid, valid := sessions.verify_session_id(sid_with_hmac, secret)
 
 	assert unverified_sid == verified_sid
-	assert valid == true
+	assert valid
 }
 
 fn test_forged_signature_rejected() {
@@ -31,7 +31,7 @@ fn test_forged_signature_rejected() {
 	verified_sid, valid := sessions.verify_session_id(forged_cookie, secret)
 
 	// Forged signature must be rejected
-	assert valid == false
+	assert !valid
 	assert verified_sid == sid
 }
 
@@ -44,19 +44,19 @@ fn test_wrong_secret_rejected() {
 	_, valid := sessions.verify_session_id(signed_cookie, wrong_secret)
 
 	// Must be rejected when using wrong secret
-	assert valid == false
+	assert !valid
 }
 
 fn test_malformed_cookie_rejected() {
 	// Cookie without signature separator
 	_, valid1 := sessions.verify_session_id('just_a_session_id', secret)
-	assert valid1 == false
+	assert !valid1
 
 	// Empty cookie
 	_, valid2 := sessions.verify_session_id('', secret)
-	assert valid2 == false
+	assert !valid2
 
 	// Cookie with empty parts
 	_, valid3 := sessions.verify_session_id('.', secret)
-	assert valid3 == false
+	assert !valid3
 }


vlib/x/sessions/tests/session_test.v
```diff
@@ -21,3 +21,42 @@ fn test_session_id() {
...
+fn test_forged_signature_rejected() {
+	// Create a valid session ID
+	sid, _ := sessions.new_session_id(secret)
+
+	// Forge a cookie with valid session ID but invalid signature
+	forged_cookie := '${sid}.INVALID_SIGNATURE'
+	verified_sid, valid := sessions.verify_session_id(forged_cookie, secret)
+
+	// Forged signature must be rejected
+	assert valid == false
```
spytheman:
```suggestion
	assert !valid
```

vlib/x/sessions/tests/session_test.v
```diff
@@ -21,3 +21,42 @@ fn test_session_id() {
...
+fn test_wrong_secret_rejected() {
+	// Create a session with one secret
+	_, signed_cookie := sessions.new_session_id(secret)
+
+	// Try to verify with a different secret
+	wrong_secret := 'wrong_secret'.bytes()
+	_, valid := sessions.verify_session_id(signed_cookie, wrong_secret)
+
+	// Must be rejected when using wrong secret
+	assert valid == false
```
spytheman:
```suggestion
	assert !valid
```

vlib/x/sessions/tests/session_test.v
```diff
@@ -21,3 +21,42 @@ fn test_session_id() {
...
+}
+
+fn test_malformed_cookie_rejected() {
+	// Cookie without signature separator
+	_, valid1 := sessions.verify_session_id('just_a_session_id', secret)
+	assert valid1 == false
```
spytheman:
```suggestion
	assert !valid1
```

vlib/x/sessions/tests/session_test.v
```diff
@@ -21,3 +21,42 @@ fn test_session_id() {
...
+
+	// Empty cookie
+	_, valid2 := sessions.verify_session_id('', secret)
+	assert valid2 == false
```
spytheman:
```suggestion
	assert !valid2
```

vlib/x/sessions/tests/session_test.v
```diff
@@ -21,3 +21,42 @@ fn test_session_id() {
...
+
+	// Cookie with empty parts
+	_, valid3 := sessions.verify_session_id('.', secret)
+	assert valid3 == false
```
spytheman:
```suggestion
	assert !valid3
```

vlib/x/sessions/tests/session_test.v
```diff
@@ -21,3 +21,42 @@ fn test_session_id() {
 	assert unverified_sid == verified_sid
 	assert valid == true
```
spytheman:
```suggestion
	assert valid
```

spytheman (approved):
Thank you @fleximus 🙇🏻 .
Excellent work.

Issue #26197, opened 2025-12-28
`v fmt` does not change `it` to `index`


JalonSolov:
### Describe the bug

It should be a simple thing for `v fmt` to tell when `it` is used where `index` should be.  Instead, it just says "Already formatted".

### Reproduction Steps

x.v
```v
println([]int{len: 10, init: it + 2})
```
try to format it...
```
$ v fmt -w x.v
Already formatted file: /home/jalon/x.v
$
```
then try to run it...
```
$ v run x.v
x.v:1:36: warning: variable `it` in array initialization will soon be replaced with `index`
    1 | println([]int{len: 10, init: it + 2})
      |                                    ^
[2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
$
```

### Expected Behavior

`it` is changed to `index` when formatting.

### Current Behavior

Does not change `it` to `index`, so warning/error happens.

### Possible Solution

Change `it` to `index` while formatting.

### V version

`V 0.4.12 0b0f380`

### Environment details (OS name and version, etc.)

```
|V full version      |V 0.4.12 cfc3dd02a369a7b91bd617beebad2ccc979cfe81.0b0f380
|:-------------------|:-------------------
|OS                  |linux, "Manjaro Linux"
|Processor           |32 cpus, 64bit, little endian, AMD Ryzen 9 7950X 16-Core Processor
|Memory              |17.64GB/30.49GB
|                    |
|V executable        |/home/jalon/git/v/v
|V last modified time|2025-12-28 22:57:51
|                    |
|V home dir          |OK, value: /home/jalon/git/v
|VMODULES            |OK, value: /home/jalon/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/jalon
|                    |
|Git version         |git version 2.52.0
|V git status        |weekly.2025.52-79-g0b0f380e
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20251112
|gcc version         |gcc (GCC) 15.2.1 20251112
|clang version       |clang version 21.1.6
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.42
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
[...]

PR #26201, opened 2025-12-29
fmt: for array init, it=>index

kbkpbot:
Fix issue #26197


spytheman (approved):
Excellent work.
Thanks @kbkpbot 🙇🏻 .

PR #26202, opened 2025-12-29
checker: clear all ref from autostr receiver_type

kbkpbot:
This is a feature needed by PR #25857, which can make sure always generate auto str function like this:
```v
pub fn (it main.Example) str() string 
```
instead of something like this :
```v
pub fn (it &main.Example) str() string 
pub fn (it &&main.Example) str() string 
```

This should has no impact on current auto str mechanism.


spytheman:
Is there a performance impact?

spytheman (approved):
Thank you @kbkpbot 🙇🏻 .

PR #26206, opened 2025-12-29
docs: typo fix

warpfork:
The tiniest typo fix :))

Insufficient linebreaks left a note paragraph hiding on the end of a bullet list item about a specific variable that they weren't actually relevant to.


warpfork:
42 minutes is also a new record for fastest merge I've ever seen, without someone being down the hall from me, so, thank you for that incredible energy :D

PR #26207, opened 2025-12-29
sokol: add IME support for macOS (NSTextInputClient)

tuannx:
This PR implements the `NSTextInputClient` protocol in [sokol_app.h]
to support Input Method Editors (IME) on macOS.

## Why
Currently, Sokol applications on macOS (like `ved`) directly read `event.characters`, which bypasses the system IME. This prevents users from typing languages that require composition, such as Vietnamese (Telex/VNI), Chinese, Japanese, and Korean. The feature matrix in [sokol_app.h] previously listed IME support as TODO for macOS.

## What
- Implemented `NSTextInputClient` protocol for `_sapp_macos_view`.
- Added `markedText`, `markedRange`, and `selectedRange` to `_sapp_macos_t` to track composition state.
- Updates `keyDown:` to use `interpretKeyEvents:` when `SOKOL_NO_IME` is NOT defined, allowing the system to handle character composition.
- Wraps the new logic in `#if !defined(SOKOL_NO_IME)` to allow opting out and to ensure backward compatibility/safety.

This enables proper text input for CJK and other languages in V applications using Sokol/GG on macOS.

## Testing
Tested manually with `ved` editor on macOS. Vietnamese Telex input works correctly (e.g., typing `a` + `a` produces `â`).

### DIFF
--- a/thirdparty/sokol/sokol_app.h
+++ b/thirdparty/sokol/sokol_app.h
@@ -2440,9 +2440,15 @@ MyView2* g_view;
 @end
 #if defined(SOKOL_METAL)
     @interface _sapp_macos_view : MTKView
+    #if !defined(SOKOL_NO_IME)
+    <NSTextInputClient>
+    #endif
     @end
 #elif defined(SOKOL_GLCORE)
     @interface _sapp_macos_view : NSOpenGLView
+    #if !defined(SOKOL_NO_IME)
+    <NSTextInputClient>
+    #endif
     - (void)timerFired:(id)sender;
     @end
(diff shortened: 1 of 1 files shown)


spytheman:
Is that implemented in the upstream sokol library?
If not, please make a PR there.

I would prefer, if our patches stay small, and not implement new features that could be incompatible with future upstream changes, and make the upgrade process even worse than it currently is.

We plan to upgrade to the latest Sokol in March, so if you can manage to upstream it till then, it will go into V too, and if it is already there, you will just have to wait till March, with no further effort on your part.

PR #26210, opened 2025-12-30
x.json2: remove workaround_cast(), add isize usize decode support

kbkpbot:
This PR remove the  `workaround_cast()`, as alias cast is supported in PR #26079.
Also add support to decode `isize` and `usize`.


spytheman (approved):
Excellent work.
Thank you @kbkpbot 🙇🏻 .

PR #26185, opened 2025-12-28
thirdparty: fix tcc __atomic_thread_fence

kbkpbot:
Fix issue #25856
Fix issue #26158

This bug is caused by on `aarch64`, `tcc` remove the prefix `__` from `__atomic_thread_fence`, so a workaround needed here.
Maybe a `tcc` fix is needed for upstream.

ping @spytheman would you please help me verify that the test file is written in a correct format ? I just want to check it can be compiled under `tcc/arm64`. Do it need to set a `VFLAGS` = `-no-retry-compilation`?

PR #26211, opened 2025-12-30
v.builder: add compile asm file support

kbkpbot:
Then we can compile `.S` file to `.o` file.
Feature request by PR #26185


spytheman (approved):
Excellent work.
Thank you 🙇🏻 .

enghitalo:
That's going to be a game changer ❤️

JalonSolov:
Yep... it's going to complicate every other backend...  :-\

spytheman:
> Yep... it's going to complicate every other backend... :-\

other backends do not use `#flag` directives, so no

PR #26218, opened 2025-12-30
net.http: http keep-alive connections in server

### LINKED ISSUE #26136: std net lib http.Server not support Connection: keep-alive
### Describe the bug

every request has new connect, it cause request very slowly.

### Reproduction Steps

The response to multiple requests is slow

### Expected Behavior

when response header has Connection: keep-alive, the handle_conn method should not conn.close()

### Current Behavior

The response to multiple requests is slow

### Possible Solution

when response header has Connection: keep-alive, the handle_conn method should not conn.close()

_No response_

### V version

V 0.4.12 f6336fc

|V full version      |V 0.4.12 f6336fc0ecbd4ce3e91c24a67c01f207d16674d5
|OS                  |windows, Microsoft Windows 10 Education 19045 64-bit
|Processor           |4 cpus, 64bit, little endian, Intel(R) Core(TM) i5-7400 CPU @ 3.00GHz  
|Memory              |2.35GB/15.91GB
|                    |
|V executable        |F:\v\v.exe
|V last modified time|2025-11-05 03:37:48
|                    |
|V home dir          |OK, value: F:\v
|VMODULES            |OK, value: C:\Users\admin\.vmodules
|VTMP                |OK, value: C:\Users\admin\AppData\Local\Temp\v_0
|Current working dir |OK, value: I:\workspace-locale\v-hono
|                    |
|Git version         |git version 2.34.1.windows.1
|V git status        |weekly.2025.45-11-gf6336fc0
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.1.0
|gcc version         |gcc (GCC) 15.1.0
|clang version       |N/A
|msvc version        |N/A
|tcc git status      |N/A
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

fleximus:
**Implementation**
  1. Added `max_keep_alive_requests` setting - Limits requests per connection to prevent resource exhaustion. Default: `100` (`0` = unlimited)
  2. Updated `handle_conn`
    - Added request loop to handle multiple requests per connection
    - Track request count per connection
    - Determine `keep-alive` based on:
        - Max requests limit reached
      - Response Connection header
      - Request Connection header
      - HTTP version default (`1.1` -> `keep-alive`, `1.0` -> `close`)
    - Automatically set `Connection` header in response
    - Close connection when `keep-alive` should end
  3. Updated HandlerWorker - Added `max_keep_alive_requests` field and propagated it through worker creation

**Test cases**

 `test_server_keep_alive` test case:
  - Creates a TCP connection to the server
  - Sends 3 requests over the same connection with `Connection: keep-alive`
  - Verifies the response counter increments (proving same server handled all requests)
  - Verifies the Connection header is correctly echoed in responses
  - Sends final request with `Connection: close` to terminate

`test_server_max_keep_alive_requests`
  - Tests that connection closes after limit is reached

Fixes #26136

### DIFF
--- a/vlib/net/http/server_test.v
+++ b/vlib/net/http/server_test.v
@@ -364,6 +364,63 @@ fn test_server_keep_alive() {
 	assert handler.request_count == 3
 }
 
+fn test_server_max_keep_alive_requests() {
+	log.warn('${@FN} started')
+	defer { log.warn('${@FN} finished') }
+	mut handler := KeepAliveHandler{}
+	mut server := &http.Server{
+		accept_timeout:          atimeout
+		handler:                 handler
+		addr:                    '127.0.0.1:18200'
+		show_startup_message:    false
+		max_keep_alive_requests: 3 // Limit to 3 requests per connection
+	}
+	t := spawn server.listen_and_serve()
+	server.wait_till_running() or {
+		estr := err.str()
+		if estr == 'maximum retries reached' {
+			log.error('>>>> Skipping test ${@FN} since its server could not start, err: ${err}')
+			return
+		}
+		log.fatal(estr)
+	}
+
+	// Test that connection closes after max_keep_alive_requests
+	mut conn := net.dial_tcp('127.0.0.1:18200')!
+	defer { conn.close() or {} }
+	conn.set_read_timeout(5 * time.second)
+	conn.set_write_timeout(5 * time.second)
+
+	// Send first request
+	request1 := 'GET /test1 HTTP/1.1\r\nHost: 127.0.0.1:18200\r\nConnection: keep-alive\r\n\r\n'
+	conn.write(request1.bytes())!
+	mut resp1 := read_http_response(mut conn)!
+	log.info('Response 1: ${resp1}')
+	assert resp1.contains('request #1')
+	assert resp1.to_lower().contains('connection: keep-alive')
+
+	// Send second request
+	request2 := 'GET /test2 HTTP/1.1\r\nHost: 127.0.0.1:18200\r\nConnection: keep-alive\r\n\r\n'
+	conn.write(request2.bytes())!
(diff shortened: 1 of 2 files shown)


vlib/net/http/server_test.v
```diff
@@ -285,3 +285,180 @@ fn test_host_header_sent_to_server() {
...
+	handler.request_count++
+	mut r := http.Response{
+		body: 'request #${handler.request_count}'
+	}
+	r.set_status(.ok)
+	r.set_version(req.version)
+	// Echo back the Connection header from the request if present
+	if conn := req.header.get(.connection) {
+		r.header.set(.connection, conn)
+	}
```
spytheman, line 306: What do you think of doing these inside the server implementation after each custom handle() callback returns?

spytheman: Are there cases, where they should be avoided?

vlib/net/http/server_test.v
```diff
@@ -285,3 +285,180 @@ fn test_host_header_sent_to_server() {
...
+
+	// Test keep-alive by sending multiple requests over a single TCP connection
+	mut conn := net.dial_tcp('127.0.0.1:18198')!
+	defer { conn.close() or {} }
+	conn.set_read_timeout(5 * time.second)
+	conn.set_write_timeout(5 * time.second)
+
+	// Send first request with Connection: keep-alive
+	request1 := 'GET /test1 HTTP/1.1\r\nHost: 127.0.0.1:18198\r\nConnection: keep-alive\r\n\r\n'
+	conn.write(request1.bytes())!
```
spytheman, line 338:
```suggestion
	conn.write_string(request1)!
```

spytheman (approved):
Excellent work.

Issue #23424, opened 2025-01-10
json decode silently discards quoted numeric types


louis77:
### Describe the bug

The `json` parser silently discards numeric values which are quoted (which is sadly often encountered using external APIs). `x.json2` properly casts these values into the desired numeric type.

This behaviour was already reported in https://github.com/vlang/v/issues/13996 but the issue was closed without a solution.

### Reproduction Steps

```v test.v
module main

import json
import x.json2

struct Rating {
	name          string
	score         f64
	sources_count int
	reviews       int
}

fn main() {
	jsonstr := '{"name": "V Language", "score": "4.9", "sources_count": 6, "reviews": "9999"}'

	rating := json.decode(Rating, jsonstr)!

	println(rating)

	// Prints:
	//
	// Rating{
	// 	name: 'V Language'
	// 	score: 0.0
	// 	sources_count: 6
	//  reviews: 0
	// }

	rating2 := json2.decode[Rating](jsonstr)!
	println(rating2)

	// Prints:
	//
	// Rating{
	// 	name: 'V Language'
	// 	score: 4.9
	// 	sources_count: 6
	// 	reviews: 9999
	// }
}
```

### Expected Behavior

I expect `json.decode()` to either fail with an error describing which field couldn't be parsed, or try casting the values like `json2.decode()` does.

### Current Behavior

`json` silently ignores the fields, which can lead to uncaught bugs in applications, i.e. when external APIs change their schema.

### V version

0.4.9 3953445

### Environment details (OS name and version, etc.)

```
V full version: V 0.4.9 3953445
OS: macos, macOS, 15.2, 24C101
Processor: 8 cpus, 64bit, little endian, Apple M2

vexe: /opt/homebrew/Cellar/vlang/0.4.9/libexec/v
vexe mtime: 2024-12-22 00:48:41

vroot: OK, value: /opt/homebrew/Cellar/vlang/0.4.9/libexec
VMODULES: OK, value: /Users/user/.vmodules
VTMP: OK, value: /tmp/v_501

Git version: git version 2.39.5 (Apple Git-154)
Git vroot status: 4.4.15 (62319 commit(s) behind V master)
.git/config present: false

CC version: Apple clang version 16.0.0 (clang-1600.0.26.4)
emcc version: N/A
thirdparty/tcc: N/A
```

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
[...]

PR #26220, opened 2025-12-31
x.json2: reject string-to-number type casting during decoding

### LINKED ISSUE #26082: x.json: string to number cast, not following JSON specfications
### Describe the bug

Tests like `vlib/x/json2/tests/decode_budget_number_test.v` are supporting string cast

### Reproduction Steps

```v
import x.json2 as json

fn test_budget_number() {
	assert json.decode[int]('"0"')! == 0
	assert json.decode[int]('"100"')! == 100
	assert json.decode[f64]('"-23.6e1"')! == -236.0

	assert json.decode[[]int]('["100", 99, "98", 97]')! == [100, 99, 98, 97]
}

fn test_budget_number_malformed() {
	json.decode[int]('"+100"') or {
		if err is json.JsonDecodeError {
			assert err.line == 1
			assert err.character == 2
			assert err.message == 'Syntax: expected digit got +'
		}

		return
	}

	assert false
}

```

### Expected Behavior

json check error

### Current Behavior

asset pass

### Possible Solution

_No response_

_No response_

### V version

V 0.4.12 07331bd

Linux
> [!NOTE]
>

fleximus:
Honor the JSON specification which requires numeric values to be unquoted.

This fix lets `json.decode[int]('"100"')` correctly return a `JsonDecodeError` with message `"Expected number, but got string"`, while unquoted numbers like `json.decode[int]('100')` continue to work as expected.

Fixes #26082

### DIFF
--- a/vlib/x/json2/decode.v
+++ b/vlib/x/json2/decode.v
@@ -52,8 +52,8 @@ pub:
 	strict bool
 }
 
-// DecodeContext is the internal decoding state.
-struct DecodeContext {
+// Decoder is the internal decoding state.
+struct Decoder {
 	json   string // json is the JSON data to be decoded.
 	strict bool   // strict mode rejects quoted strings as numbers
 mut:
@@ -153,7 +153,7 @@ fn (e JsonDecodeError) msg() string {
 }
 
 // checker_error generates a checker error message showing the position in the json string
-fn (mut checker DecodeContext) checker_error(message string) ! {
+fn (mut checker Decoder) checker_error(message string) ! {
 	position := checker.checker_idx
 
 	mut line_number := 0
@@ -214,7 +214,7 @@ fn (mut checker DecodeContext) checker_error(message string) ! {
 }
 
 // decode_error generates a decoding error from the decoding stage
-fn (mut decoder DecodeContext) decode_error(message string) ! {
+fn (mut decoder Decoder) decode_error(message string) ! {
 	mut error_info := ValueInfo{}
 	if decoder.current_node != unsafe { nil } {
 		error_info = decoder.current_node.value
@@ -293,18 +293,18 @@ pub fn decode[T](val string, params DecoderOptions) !T {
 			character: 1
 		}
 	}
-	mut ctx := DecodeContext{
+	mut decoder := Decoder{
 		json:   val
 		strict: params.strict
 	}
 
-	ctx.check_json_format()!
+	decoder.check_json_format()!
 
 	mut result := T{}
-	ctx.current_node = ctx.values_info.head
-	ctx.decode_value(mut result)!
+	decoder.current_node = decoder.values_info.head
+	decoder.decode_value(mut result)!
 	unsafe {
-		ctx.values_info.free()
+		decoder.values_info.free()
 	}
 	return result
 }
@@ -315,7 +315,7 @@ fn get_dynamic_from_element[T](t T) []T {
 
 // decode_value decodes a value from the JSON nodes.
 @[manualfree]
-fn (mut decoder DecodeContext) decode_value[T](mut val T) ! {
+fn (mut decoder Decoder) decode_value[T](mut val T) ! {
 	// Custom Decoders
 	$if val is StringDecoder {
 		struct_info := decoder.current_node.value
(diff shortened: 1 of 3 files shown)


Delta456 (approved):

spytheman:
@Larsimusrex , @dy-tea what do you think?

Larsimusrex:
#23424 originally requested this feature. I don't think it is really a problem because a decoder that "does more than it should" doesn't create incompatible json files. It just increases compatibility with other poorly written encoders. Another example would be surrogate decoding which is non-standart but if removed would break compatibility with pythons json encoder.

Also the json is only defines the syntax, not which datatypes should/shouldn't be allowed.

spytheman:
What do you all think of @JalonSolov's idea, of adding a `strict` decoding mode, that will error in such cases, while keeping the default lenient (i.e. accepting them, to keep the compatibility with existing encoders in the wild)?

dy-tea:
I think having this as an option is good but agree it doesn't need to be the default.

fleximus:
Alright, good feedback guys. I've added a strict mode flag for JSON decoding with two modes:

  | Mode    | Behavior                                                                |
  |---------|-------------------------------------------------------------------------|
  | Default | Accepts quoted strings as numbers for compatibility (e.g., "123" → 123) |
  | Strict  | Rejects quoted strings as numbers per JSON spec compliance              |

  Usage looks like this:
``` v
  import x.json2 as json

  // Default mode - accepts quoted strings as numbers
  value := json.decode[int]('"123"')!        // returns 123
  arr := json.decode[[]int]('["1", 2, "3"]')! // returns [1, 2, 3]

  // Strict mode - rejects quoted strings as numbers
  value := json.decode2[int]('123', strict: true)!  // returns 123
  json.decode2[int]('"123"', strict: true) or {
      // Error: "Expected number, but got string"
  }
```

vlib/x/json2/decode.v
```diff
@@ -271,8 +282,17 @@ fn (mut decoder Decoder) decode_error(message string) ! {
...
+// Use `decode2` with `strict: true` for strict JSON spec compliance.
 pub fn decode[T](val string) !T {
+	return decode2[T](val, DecodeParams{})
+}
+
+// decode2 decodes a JSON string into a specified type with configurable parameters.
+// In strict mode, quoted strings are not accepted as numbers (JSON spec compliant).
+// In default mode, quoted strings like '"123"' are accepted as numbers for compatibility.
+@[manualfree]
+pub fn decode2[T](val string, params DecodeParams) !T {
```
gechandesu: Why introduce `decode2()`? You can extend the current `decode()` without breaking backward compatibility. I would rename the `DecodeParams` struct to `DecoderOptions`, since we already have `EncoderOptions`.

enghitalo:
Hmmm.... May I suggest that the strict flag not be included in the function parameters? A public function to create a decoder with configuration would be nice, more or less like how it works in the encoder.

enghitalo:
@spytheman Do you prefer strict to be the default, or the other way around?

vlib/x/json2/decode.v
fleximus: You are correct on both points, let me correct that.

fleximus:
@enghitalo I agree on all your points.

For the question what should be the default, I think a tolerant parsing is the better choice and have the strict one as opt-in. It's pragmatic and follows the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) also known as Postel's law. This pattern is also seen in browsers and chooses practicality over pedantry.

spytheman:
@enghitalo I also prefer the lenient mode (the one that is present right now) to be the default. I do not see a reason for the strict mode to be the default what so ever - it will break the backwards compatibility, and will make the overall system more brittle, since the decoder will be normally used to parse data coming from external sources.

spytheman:
@fleximus please change it to:
```v
pub fn decode[T](val string, params DecoderOptions) !T {
```
No deprecations are needed. Old code will still be able to call it just with:
```v ignore
res := decode[Type]('....')!
```
while new code can also do:
```v ignore
res := decode[Type]('....', strict: true)!
```

spytheman:
That will also remove the need for the `pub struct Decoder {` that currently does nothing but just stores the passed parameters.

spytheman (approved):
Thank you @fleximus 🙇🏻 .
Thanks to all reviewers too 🙇🏻‍♂️ .

PR #26223, opened 2025-12-31
checker: error for unwrapped option/result types used with in operator

### LINKED ISSUE #26208: veb: using optional type with `in` operator in template causes C compilation error
### Describe the bug

Using an optional type (`?string`) with the `in` operator in a veb template causes a C compilation error: `cannot convert 'struct _option_string' to 'unsigned char *'`

### Reproduction Steps

templates/test.html:
``` html
<!DOCTYPE html>
<html>
<body>
@for item in items
    @if item.owner_id in owner_map
    <p>Owner: @{owner_map[item.owner_id]}</p>
    @else
    <p>No owner</p>
    @end
@end
</body>
</html>

```
main.v:
``` v
module main

import veb

struct Item {
	id       int
	name     string
	owner_id ?string  // Optional field - THIS CAUSES THE BUG
}

struct App {
	veb.StaticHandler
}

struct Context {
	veb.Context
}

@['/test'; get]
pub fn (mut app App) test(mut ctx Context) veb.Result {
	items := [Item{id: 1, name: "test", owner_id: "user1"}]
	owner_map := {"user1": "John Doe"}
	
	return $veb.html('test.html')
}

fn main() {
	mut app := &App{}
}
```

### Expected Behavior

``` v
error: unwrapped Option cannot be used with `in`
```

### Current Behavior

```
cc: /tmp/v_1000/vbug.01KDNXBKW6W7170EHVVJMVP4YA.tmp.c:15751: warning: assignment from incompatible pointer type
cc: /tmp/v_1000/vbug.01KDNXBKW6W7170EHVVJMVP4YA.tmp.c:21639: error: cannot convert 'struct _option_string' to 'unsigned char *'
...
cc: /tmp/v_1000/vbug.01KDNXBKW6W7170EHVVJMVP4YA.tmp.c:21639: error: cannot convert 'struct _option_string' to 'unsigned char *'
(note: the original output was 5 lines long; it was truncated to its first 2 lines + the last line)
builder error: 
==================
```

### Possible Solution

The compiler should either:
 ~1. Auto-unwrap the optional for the `in` check or~
  2. Provide a clear compile-time error like `error: unwrapped Option cannot be used with 'in'`

_No response_

### V version

V 0.4.12 8f8a491

V full version: V 0.4.12 9a73766f5afd8f4808964c9b48280abf06c4b73b.8f8a491
OS: Linux

> [!NOTE]
>

fleximus:
Implementation:
  - Prevents a C compilation error and gives the developer the feedback to unrwap first
  - Matches V's philosophy of requiring explicit handling of `Optional` types rather than implicit conversions

Test:
  - Optional struct field (`item.owner_id ?string`) used with map `in` operator
  - Optional function return (`get_opt() ?int`) used with array `in` operator

Fixes #26208

### DIFF
--- a/vlib/v/checker/tests/unwrapped_option_result_in_operation_err.out
+++ b/vlib/v/checker/tests/unwrapped_option_result_in_operation_err.out
@@ -1,25 +1,25 @@
-vlib/v/checker/tests/unwrapped_option_result_in_operation_err.vv:3:6: error: unwrapped result cannot be used with `!in`
+vlib/v/checker/tests/unwrapped_option_result_in_operation_err.vv:3:6: error: unwrapped Result cannot be used with `!in`
     1 | fn main() {
     2 |     list := ['string']
     3 |     _ = return_string_or_error() !in list
       |         ~~~~~~~~~~~~~~~~~~~~~~~~
     4 |     _ = return_string_or_error() in list
     5 |     _ = return_string_or_none() in list
-vlib/v/checker/tests/unwrapped_option_result_in_operation_err.vv:4:6: error: unwrapped result cannot be used with `in`
+vlib/v/checker/tests/unwrapped_option_result_in_operation_err.vv:4:6: error: unwrapped Result cannot be used with `in`
     2 |     list := ['string']
     3 |     _ = return_string_or_error() !in list
     4 |     _ = return_string_or_error() in list
       |         ~~~~~~~~~~~~~~~~~~~~~~~~
     5 |     _ = return_string_or_none() in list
     6 |     _ = return_string_or_error() !in list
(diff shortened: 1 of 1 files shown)


spytheman:
@fleximus you can use `VAUTOFIX=1 ./v vlib/v/compiler_errors_test.v` 2 times in a row, to fix the previous .out files that recorded the old error message.
The first run will fail, but also overwrite the .out files with the new results, and the second run should then complete fine.

spytheman (approved):
Excellent work.
Thanks @fleximus .

fleximus:
Excellent mentoring in practice.
Thanks @spytheman.

PR #26229, opened 2025-12-31
rand.xoroshiro128pp: prevent generation of extra zeros in u16()

tankf33der:
Audit of `rand` module continues.

if you look at the patch from top to bottom, it consists of parts for `u8()` and for `u16()`.

u8:
just removed an extra variable `value` whose purpose was to be used as the function result.

u16:
and here it's more interesting - a logical error leads to generation of extra zeros every 4th call, breaks the work logic and statistics.

 demo code:
```v
import rand.xoroshiro128pp

fn main() {
	mut xo := xoroshiro128pp.XOROS128PPRNG{}
	for _ in 0 .. 32 {
		println(xo.u16())
	}
}
```


tankf33der:
> Are you planning on reviewing the other `rand` generators too?

almost done.

Issue #26233, opened 2025-12-31
undefined symbol 'builtin__array_push_noscan'


tankf33der:
```v
fn main() {
    mut a := ['a', 'b', 'c']
    f := fn (e u8) u8 { return if e.is_letter() { e } else { ` ` } }
    a.sort_with_compare(fn [f] (x &string, y &string) int {
        return compare_strings(x.bytes().map(f).bytestr(), y.bytes().map(f).bytestr())
    })
}
```

```
$ v run m.v
lambda:~$ v run m.v 
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/m.01KDTJ0PNPEXW7HHBM8RJRM775.tmp.c:744: warning: implicit declaration of function 'builtin__array_push_noscan'
cc: tcc: error: undefined symbol 'builtin__array_push_noscan'
=================================================================
Try passing `-g` when compiling, to see a .v file:line information, that correlates more with the C error.
(Alternatively, pass `-show-c-output`, to print the full C error message).
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

tankf33der:
another variation:
```v
fn main() {
    x := u8(1)
    f := fn [x] (e u8) u8 { return e + x }
    [u8(5)].map(f)
}
```

jorgeluismireles:
Same error with this reduction too:
```v
fn main() {
    f := fn (e u8) u8 { return e }
    [u8(5)].map(f)
}
```

felipensp:
Possible temporary fix: -no-skip-unused

JalonSolov:
Yes.  Adding `-no-skip-unused` works.

PR #26238, opened 2026-01-01
markused: add array.map() and .filter() support

kbkpbot:
Fix issue #26233


spytheman (approved):
Excellent work.
Thank you @kbkpbot 🙇🏻 .

PR #26245, opened 2026-01-02
ast: refer to Table.type_symbols instead of Table.types in comment

revosw:
A comment above `struct TypeSymbol` refers to a field `Table.types` that doesn't exist. It should refer to `Table.type_symbols`


spytheman (approved):

spytheman:
Good spot. Thanks.

PR #26246, opened 2026-01-02
wasm: Basic match keyword handling

SheatNoisette:
Implement the WASM codegen for the "match" keyword with ranges, based on the "if" implementation with their associated tests.

Only numbers are handled (And by extensions enums), as string comparisons are not implemented and should be another PR.


SheatNoisette:
Still working on it.

I found a bug when passing a struct value in a match, maybe the inner match expression returns a value to an outer match's else branch, and so the generated WASM doesn't seems properly preserve the return value. First, I though just adding a flush would work or set a proper exit code for Wasmer.

TLDR:
```vlang
struct TestStruct {
	value bool
}

fn function(other TestStruct) {
	var := 1
	println(match var {
		-12...0 {
			'Wrong'
		}
		else {
			match other.value {
				true {
					// Adding a \n doesn't change anything
					'Right'
				}
				else {
					'False'
				}
			}
		}
	})
}

fn main() {
	ts := TestStruct{
		value: true
	}
	function(ts)

	// Adding this works and force to print "right", but I'm sure this is a side effect of the problem
	// println("")
}
```

Delta456 (commented):
Can we add tests for boolean types?

SheatNoisette:
Uh, maybe I found a bug in the WASM backend and this may not be related of my changes:
```vlang
struct TestStruct {
	value bool
}
fn function(other TestStruct) {
	println('Test')
}
fn main() {
	ts := TestStruct{
		value: true
	}
	function(ts)
}
```
This prints nothing.

But this works:
```vlang
fn function() {
	println('Test')
}
fn main() {
	function()
}
```

SheatNoisette:
@Delta456 Bool tests added

vlib/v/gen/wasm/gen.v
```diff
@@ -553,6 +553,90 @@ pub fn (mut g Gen) if_expr(ifexpr ast.IfExpr, expected ast.Type, existing_rvars
...
+		g.func.eq(wasm_type)
+	}
+
+	blk := g.func.c_if([], unpacked_params)
+	{
+		if branch.stmts.len > 0 {
+			g.rvar_expr_stmts(branch.stmts, expected, existing_rvars)
+		}
+	}
+	{
```
spytheman, line 626: cute

PR #26248, opened 2026-01-02
wasm: Basic for in expression support

SheatNoisette:
Basic numeric for in ranges support in WASM, such as:
```vlang
for i in 0 .. 5 {
    print(i)
}
```

Missing maps and list support in the backend restrict me to go even further for now.

### DIFF
--- a/vlib/v/gen/wasm/gen.v
+++ b/vlib/v/gen/wasm/gen.v
@@ -955,6 +955,49 @@ pub fn (mut g Gen) expr(node ast.Expr, expected ast.Type) {
 	}
 }
 
+pub fn (mut g Gen) for_in_stmt(node ast.ForInStmt) {
+	loop_var_type := unpack_literal_int(node.val_type)
+	block := g.func.c_block([], [])
+	{
+		mut loop_var := Var{}
+		loop_var = g.new_local(node.val_var, loop_var_type)
+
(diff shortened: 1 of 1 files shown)


vlib/v/gen/wasm/gen.v
```diff
@@ -1034,6 +1034,51 @@ pub fn (mut g Gen) expr_stmt(node ast.Stmt, expected ast.Type) {
 			}
 			g.func.c_end(block)
 		}
+		ast.ForInStmt {
```
medvednikov, line 1080: I think it's best to move this to a separate method.

medvednikov: otherwise the expr_stmt will get too big

SheatNoisette: Sure, dunno where to put it...

I think gen.v should be splitted into different files.

SheatNoisette: Done

JalonSolov: I'd say since you're the one actually working on it now, if you think gen.v should be split, then split it.

As long as the split makes sense, I don't think anyone will complain.

I would do that in a separate PR, just to avoid "muddying the waters" between that and other changes.

PR #26250, opened 2026-01-02
wasm: Fix infinite loop using continue in C-like for loops

SheatNoisette:
When a continue statement was executed, it jumped to the loop label, which was at the start of the loop. This skipped the increment step (for example, i += 2), causing the loop variable to never increment and creating an infinite loop.

Using the following example from the [official VLang C - Like loops](https://docs.vlang.io/statements-&-expressions.html#c-for):
```vlang
for i := 0; i < 10; i += 2 {
    // Don't print 6
    if i == 6 {
        continue
    }
    println(i)
}
```
Before this patch, this prints the following and goes into an infinite loop:
```
0
2
4
```

C Backend output:
```
0
2
4
8
```

### DIFF
--- a/vlib/v/gen/wasm/tests/labeled_for_inc.vv
+++ b/vlib/v/gen/wasm/tests/labeled_for_inc.vv
@@ -1,0 +1,78 @@
+fn labeled_break_continue_with_inc() int {
+	// Should not infinite loop
+	outer: for i := 4; true; i++ {
+		for {
+			if i < 7 {
+				continue outer
+			} else {
+				break outer
+			}
+		}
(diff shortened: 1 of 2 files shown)


SheatNoisette:
By more testing, this also resolved the infinite loop with the documentation example of labelled break & continue:
```
outer: for i := 4; true; i++ {
    println(i)
    for {
        if i < 7 {
            continue outer
        } else {
            break outer
        }
    }
}
```

spytheman (approved):
Excellent work.
Thank you @SheatNoisette 🙇🏻 .

Issue #26247, opened 2026-01-02
add array  method `last_index()`


kbkpbot:
### Describe the feature

Just like a string, I think we need a `last_index()` for array too, currently, array has a method `index()`

This is a feature request by fixing issue #26243

### Use Case

For string builder work for searching a buffer from the end.
```v
mut sb := strings.new_builder(100)
....
if sb.last_index('=') {
 ....
}
```

### Proposed Solution

add a method `last_index()` for array type.

### Acknowledgements

- [x] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.5.0 49bade567852d01003fa48c777259ec36dd7d2b6

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.0 49bade567852d01003fa48c777259ec36dd7d2b6
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |10.52GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2026-01-02 14:37:03
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/arr
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.52-109-g49bade56-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Ubuntu GLIBC 2.39-0ubuntu8.6) 2.39

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #26252, opened 2026-01-03
builtin: add array.last_index()

kbkpbot:
Feature request by #26247

generated C code would like this:
```c
// V auto functions:
static int Array_int_index(Array_int a, int v) {
        int* pelem = a.data;
        for (int i = 0; i < a.len; ++i, ++pelem) {
                if (*pelem == v) {
                        return i;
                }
        }
        return -1;
}

static int Array_int_last_index(Array_int a, int v) {
        if (a.len == 0) return -1;
        static int Array_int_last_index(Array_int a, int v) {
        if (a.len == 0) return -1;
        int* pelem = (byte*)a.data + (a.len-1)*a.element_size;
        for (int i = a.len-1; i >= 0; --i, --pelem) {
                if (*pelem == v) {
                        return i;
                }
        }
        return -1;
}
```

### DIFF
--- a/vlib/v/gen/js/array.v
+++ b/vlib/v/gen/js/array.v
@@ -28,12 +28,11 @@ fn (mut g JsGen) gen_array_index_method(left_type ast.Type, is_last_index bool) 
 
 		mut fn_builder := strings.new_builder(512)
 		fn_builder.writeln('function ${fn_name}(a, v) {')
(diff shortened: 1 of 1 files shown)


spytheman (approved):
Excellent work.
Thank you @kbkpbot .

PR #26254, opened 2026-01-03
bench: Add benchmark for string dedup

kbkpbot:
Benchmark comparison of four string deduplication methods in V: basic array, pre-allocated array, map, and set


spytheman (approved):
I am not sure I get the point of this, but if you find it useful, ok.

PR #26255, opened 2026-01-03
cgen: fix array.contains() when right is ref

kbkpbot:
Fix issue #26253


vlib/v/tests/arrays_reverse_ref_test.v
```diff
@@ -0,0 +1,12 @@
+module main
+
+import arrays
+
+fn test_main() {
+	a := ['abc', 'def']
+
+	for str in arrays.reverse_iterator(a) {
+		println(str in a)
+		assert (str in a) == true
```
spytheman, line 10:
```suggestion
		assert (str in a)
```

Issue #26243, opened 2026-01-02
complex array define for option type cause cgen error


kbkpbot:
### Describe the bug

see code
may related to PR #25365

### Reproduction Steps

```v
_ := [2][3]?int{}
```

### Expected Behavior

compile ok

### Current Behavior

```
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/option2.01KDY7C9X6QWEZ0K1W6VJZZKGQ.tmp.c:563: error: ';' expected (got "IError")
=================================================================
Try passing `-g` when compiling, to see a .v file:line information, that correlates more with the C error.
(Alternatively, pass `-show-c-output`, to print the full C error message).
builder error:
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .

```

### V version

V 0.5.0 49bade567852d01003fa48c777259ec36dd7d2b6

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.0 49bade567852d01003fa48c777259ec36dd7d2b6
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |11.74GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2026-01-02 01:52:09
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug/arr
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.52-109-g49bade56-dirty
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
[...]

PR #26258, opened 2026-01-04
cgen: fix multi dim fixed array with opt elem

kbkpbot:
Fix issue #26243


spytheman (approved):
Excellent work.
Thank you.

PR #26260, opened 2026-01-04
wasm: Basic string operations

SheatNoisette:
Attempting to add string operations to the WASM backend.

This adds:
- `+` : Concatenation 
- `==`
- `<`
- `>`
- match
- if

This required a slight refactor of "ops.v" to avoid an if/else/match hell.


JalonSolov:
Looks like the only thing needed is formatting the file.

For peace of mind in the future, run `v git-fmt-hook install`, and your code will be auto-formatted before every commit (if needed).

JalonSolov:
Good ol' Windows... <sigh>

SheatNoisette:
Just checked with a `make clean`, `make all` and `v -stats test vlib/v/gen/wasm/`. I don't have any explanation for this failure... (Fedora 43 x86_64)

SheatNoisette:
This should be enough to get started.

However, I'm not entirely sure about the name of the builtin functions. Should I rename them to "string_cmp" and "string_less_than" ? I hoped some overloading would work, but I'm calling these function myself manually at code generation.

Still no idea of the Windows Failure.

Delta456:
> Should I rename them to "string_cmp" and "string_less_than" ?

Yes, let's go with that.

SheatNoisette:
@Delta456 Done!

The "+" function cannot be changed as is. This would cause issues with += handling when the backend does function patching.

Krchi:
```
C:\Applications\vlang\cmd\tools\builders\wasm_builder.exe C:\Applications\vlang\vlib\v\gen\wasm\tests\arith.vv
wasm error: get_wasm_type: unreachable type '[]u8' ast.TypeInfo(ast.Array{
    nr_dims: 1
    elem_type: ast.Type(0xb = 11)
})
```

SheatNoisette:
Sure, but what I don't understand is why this branch is taken on Windows and not on Linux. I'll try to make a VM later.

Krchi:
> Sure, but what I don't understand is why this branch is taken on Windows and not on Linux. I'll try to make a VM later.

you can update v code, it may be path separator problem in windows, wasmgen need escape stringliteral's val, it pass all test with your code

SheatNoisette:
Hmmm, you have an example ?

Krchi:
> Hmmm, you have an example ?

#26264 i found wasm tests is not passed under windows in many commit befre b9e9192b, tests all passed  with your code in my local

SheatNoisette:
I see. Well, I can't rebase at the moment sadly.

SheatNoisette:
By thinking about it, I should move the string functions into string.v built in and switch again to something similar as the real string.v until we have enough features.

SheatNoisette:
Done

PR #26262, opened 2026-01-04
bitfield: fix pop_count(), add test

tankf33der:
I have created a new testsuite, now for `bitfield` module vs. some random github [repo](https://github.com/ciubotaru/bitfield), 20 tests.

The `pop_count()` function and indirectly `hamming()` work incorrectly. If the number of bits is a multiple of 32, then the last word is not processed at all.

I propose a naive patch. Passed 1+M tests for bit sequences of random length.


JalonSolov:
Looks like someone forgot the end of a range is exclusive.  Nice, simple fix.

PR #26264, opened 2026-01-05
wasm: fix path separator in windows

Krchi:
repalce backslash with slash for file path, fix tests in windows


spytheman:
Thank you @Krchi 🙇🏻 .

PR #26266, opened 2026-01-05
rand.mt19937: relax data for -autofree, update test

### LINKED ISSUE #13398: Abort in mt19937 using autofree engine
**V version:** 0.2.4 4ef7d26
**OS:** linux, "Arch Linux"

**What did you do?**
`v -autofree test vlib/rand/mt19937/`

No `double free or corruption` message
 
**What did you see instead?**
```
---- Testing... --------------------------------------------------------------------------------------------------------
 FAIL   435.583 ms vlib/rand/mt19937/mt19937_test.v
double free or corruption (top)
7f1c68ebfd22 : at ???: RUNTIME ERROR: abort() called
```

tankf33der:
Fixes #13398

When operating in `-autofree` mode, memory errors that should not occur are detected.
Other generators are not affected by this.

1. `rng.state = calculate_state(seed_data, mut rng.state)` - if one is already being modified, then there's no need to return it.
2. `mut rnd := &rand.PRNG(&mt19937.MT19937RNG{})` - the usage principle is copied from another generator's test

### DIFF
--- a/vlib/rand/mt19937/mt19937.v
+++ b/vlib/rand/mt19937/mt19937.v
@@ -73,14 +73,13 @@ fn get_first_state(seed_data []u32) []u64 {
 
 // calculate_state returns a random state array calculated from the `seed_data`.
 @[ignore_overflow]
-fn calculate_state(seed_data []u32, mut state []u64) []u64 {
+fn calculate_state(seed_data []u32, mut state []u64) {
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Thank you @tankf33der 🙇🏻 .

Issue #26221, opened 2025-12-31
`v -d veb_livereload watch -o app run . `  This command caused the hot load to not be triggered


Avey777:
### Describe the bug

`-o app `  ：Ensure that veb hot loading is not triggered

https://zed.dev/

### Reproduction Steps

`v -d veb_livereload watch -o app run .`

### Expected Behavior

Saving modified files and hot loading can trigger normally

### Current Behavior

After saving the file , hot loading did not trigger.

### V version

V 0.5.0 b9e9192

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.0 8f8a491723cafe9ec8bf352d1756a3d67c729656.b9e9192
|:-------------------|:-------------------
|OS                  |linux, Deepin 25
|Processor           |6 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
|Memory              |0.7GB/7.6GB
|                    |
|V executable        |/home/Jengro/.vmr/versions/v_versions/v_latest/v
|V last modified time|2026-01-04 08:30:32
|                    |
|V home dir          |OK, value: /home/Jengro/.vmr/versions/v_versions/v_latest
|VMODULES            |OK, value: /home/Jengro/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/Jengro
|                    |
|Git version         |git version 2.51.0
|V git status        |weekly.2025.52-127-gb9e9192b
|.git/config present |true
|                    |
|cc version          |cc (Deepin 12.3.0-17deepin15) 12.3.0
|gcc version         |gcc (Deepin 12.3.0-17deepin15) 12.3.0
|clang version       |Deepin clang version 17.0.6 (5deepin6)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.38-6deepin17) 2.38
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Jengro777:
@felipensp 
The issue here is confirmed to be caused by the use of the `- o app` parameter

`v -d veb_livereload watch  run . ` ： This command is normal

[...]

PR #26269, opened 2026-01-05
vwatch: fix process handle leak in `kill_pgroup()`

fleximus:
**Problem**
Process handles not being released when `vwatch` kills and restarts child processes.

**Change**
Add missing `close()` call after `wait()` to properly release OS resources.
Without this, each restart cycle leaks file descriptors and process
handles, causing resource exhaustion during long watch sessions
(e.g., veb livereload development).

**Why**
`wait()` only waits for the process to exit and retrieves its `status. close()` is required to actually free the OS resources (file descriptors, process table entries). Without it, resources accumulate with each restart cycle.

### DIFF
--- a/cmd/tools/vwatch.v
+++ b/cmd/tools/vwatch.v
@@ -282,6 +282,7 @@ fn (mut context Context) kill_pgroup() {
 		context.child_process.signal_pgkill()
 	}
 	context.child_process.wait()
+	context.child_process.close()
 }
 
 fn (mut context Context) run_before_cmd() {


Jengro777:
Great! I’ve been struggling with this for a long time.

fleximus:
It eats ressources. I also suggest we increase `livereload_poll_interval_ms` in `vlib/veb/context.v ` to 1000 - every client would poll 4 times a second this tiny server and once per second is fine enough. That said with this fix it gets calmer but both together are more powerful.

Jengro777:
https://github.com/vlang/v/issues/26221

vwatch also has this issue.

PR #26270, opened 2026-01-06
veb: fix double-send error when after-middleware sends response

fleximus:
When a global after-middleware sends a response, route-specific
  after-middleware was still executed, triggering the error
  "a response cannot be sent twice over one connection".

  Skip route-specific after-middleware if a response was already
  sent by global after-middleware.


spytheman:
Please, add a small test to ensure it stays that way in the future too.

spytheman:
(the CI failures are not related)

PR #26274, opened 2026-01-06
io: deprecate the io.string_reader module (since it is not used in vlib, and has confusing and under tested/documented API - see https://github.com/vlang/v/pull/25713 )

spytheman:


vlib/io/string_reader/string_reader.v
```diff
@@ -1,3 +1,5 @@
+@[deprecated: 'If you use it, create a local module named `string_reader`, and copy https://github.com/vlang/v/blob/f24d49259db5d2d7211e79bcf6b6f507d22847f3/vlib/io/string_reader/string_reader.v there.']
```
Delta456, line 1:
```suggestion
@[deprecated: 'If you want to use it, create a local module named `string_reader`, and copy the code from https://github.com/vlang/v/blob/f24d49259db5d2d7211e79bcf6b6f507d22847f3/vlib/io/string_reader/string_reader.v there.']
```

spytheman: No, the message is intentionally as it is - it should not be used in new projects, and it is targeting potential existing users (which may not exist even, but there is no way to know that for sure :-| ).

PR #26275, opened 2026-01-06
examples (tcp_notify_echo_server): Improve platform checks and buffer size in echo server

enghitalo:
Update error message for Windows compatibility and increase buffer size for reading data.

### DIFF
--- a/examples/tcp_notify_echo_server.v
+++ b/examples/tcp_notify_echo_server.v
@@ -9,7 +9,7 @@ import time
 
 fn main() {
 	$if windows {
-		eprintln('This example doe not work on Windows yet.')
+		eprintln("This example doesn't work on Windows yet.")
 		exit(1)
 	}
 


examples/tcp_notify_echo_server.v
```diff
@@ -8,8 +8,8 @@ import time
 // netcat in separate shells, for example: `nc localhost 9001`
 
 fn main() {
-	$if !linux {
-		eprintln('This example only works on Linux')
+	$if windows {
+		eprintln('This example doe not work on Windows yet.')
```
enghitalo:
```suggestion
		eprintln('This example doesn't work on Windows yet.')
```

spytheman:
I've tested on Linux, macos, and FreeBSD and it seems that the limitation should stay as it is on master:

Linux:
image

Macos:
image

FreeBSD:
image

PR #26276, opened 2026-01-06
crypto.blake2b: add `sum()` and `blake2b_long()` for variable-length output

fleximus:
Add two new public functions to support flexible hash output lengths:

- `sum(data, outlen)`: produces Blake2b hash with configurable output length (1-64 bytes), useful when standard sizes don't fit the use case

- `blake2b_long(data, outlen)`: implements the Blake2b extension mechanism for arbitrary-length output (>64 bytes) as specified in the Argon2 RFC, enabling future Argon2 password hashing implementation

Both functions include comprehensive regression tests covering output length validation, determinism, consistency with existing functions, and the specific 1024-byte output case required by Argon2.

### DIFF
--- a/vlib/crypto/blake2b/blake2b.v
+++ b/vlib/crypto/blake2b/blake2b.v
@@ -375,3 +375,79 @@ pub fn pmac160(data []u8, key []u8) []u8 {
 	d.write(data) or { panic(err) }
 	return d.checksum_internal() or { panic(err) }
 }
+
+// sum returns the Blake2b checksum of the data with the specified output length (1-64 bytes).
+pub fn sum(data []u8, outlen int) []u8 {
+	if outlen < 1 || outlen > size512 {
+		panic('blake2b: invalid output length ${outlen}, must be between 1 and ${size512}')
+	}
+	mut d := new_digest(u8(outlen), []u8{}) or { panic(err) }
+	d.write(data) or { panic(err) }
+	return d.checksum_internal() or { panic(err) }
+}
+
+// blake2b_long produces a hash of arbitrary length using the Blake2b extension mechanism.
+// This is used by Argon2 for variable-length output greater than 64 bytes.
+// For outputs <= 64 bytes, it's equivalent to sum().
+// For outputs > 64 bytes, it uses iterative hashing as specified in the Argon2 RFC.
+pub fn blake2b_long(data []u8, outlen int) []u8 {
+	if outlen < 1 {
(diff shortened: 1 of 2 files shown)


spytheman:
@kimshrier please take a look when you have time 🙇🏻

tankf33der:
Since one PR contains so many changes, I won't do the same and will split my response into two parts.

`sum()` part:
I implemented my tests for `sum(data, outlen)` and they passed. Let's say we can do the merge and this slightly improves the functionality. But I'm pointing out that users may possibly need to get a non-incremental hash when a key is needed too. Currently there is only incremental functionality. There's still my feeling that the blake2[b,s] family is still incomplete.

blackshirt:
> Add two new public functions to support flexible hash output lengths:
> 
> * `sum(data, outlen)`: produces Blake2b hash with configurable output length (1-64 bytes), useful when standard sizes don't fit the use case
> * `blake2b_long(data, outlen)`: implements the Blake2b extension mechanism for arbitrary-length output (>64 bytes) as specified in the Argon2 RFC, enabling future Argon2 password hashing implementation
> 
> Both functions include comprehensive regression tests covering output length validation, determinism, consistency with existing functions, and the specific 1024-byte output case required by Argon2.

Is it possible to make this `blake2b` fully implements `interface Hash` from `hash` module ?. I think its also good to adapt the go version of `XOF` interface into v, See [XOF](https://cs.opensource.google/go/go/+/go1.25.5:src/hash/hash.go;l=75). Its added in go1.25.0. We have several extendable output function (xof) functionalities on the stock module.

tankf33der:
With the second part it's more complicated. Firstly, for argon2 you can get with default length, and if you're going to implement such functionality then it's called [blake2x](https://www.blake2.net/blake2x.pdf). I vote against merge.

tankf33der:
@fleximus how is doing? My test vectors for `blake2x` are ready and waiting for your great implementation.

JalonSolov:
@fleximus any further plans for this PR?

JalonSolov:
@tankf33der still have reservations about merging this PR as-is?

tankf33der:
@JalonSolov This PR should not be merged, I vote to simply close it.

tankf33der:
BTW, i have working tested prototype for `blake2x`.
DM, if somebody want to play with.

JalonSolov:
Since it has been so long with no response, closing this PR.

PR #26278, opened 2026-01-06
fmt: fix parser failing on duplicate methods in $if/$else blocks

### LINKED ISSUE #26271: v fmt fails on mutually exclusive $if blocks with duplicate method definitions
### Describe the bug

v fmt reports `duplicate method` error when the same method is defined in mutually exclusive `$if/$else` or `$if flag ?/$if !flag ?` blocks. The code compiles and runs correctly, but cannot be formatted.

### Reproduction Steps

``` v
module main

$if foo ? {

pub fn (x int) str() string {
    return 'foo mode'
}

} $else {

pub fn (x int) str() string {
    return 'normal mode'
}

}

fn main() {
    println(42.str())
}
```

Steps to reproduce:
``` bash
v run test.v        # works
v -d foo run test.v # works
v fmt test.v        # fails
```

### Expected Behavior

v fmt should recognize mutually exclusive $if blocks and not report duplicate methods

### Current Behavior

``` v
  test.v:10:16: error: duplicate method `str`
     8 | } $else {
     9 |
    10 | pub fn (x int) str() string {
       |                ~~~
    11 |     return 'normal mode'
    12 | }
```

### Possible Solution

_No response_

**Impact:** Cannot format files which use this pattern for conditional compilation.

### V version

V 0.5.0 e3616f2

V full version: V 0.5.0 9a73766f5afd8f4808964c9b48280abf06c4b73b.e3616f2
OS: Linux
> [!NOTE]
>

fleximus:
Fixes #26271

### DIFF
--- a/vlib/v/parser/fn.v
+++ b/vlib/v/parser/fn.v
@@ -660,7 +660,7 @@ fn (mut p Parser) fn_decl() ast.FnDecl {
 			}
 			// when formatting, methods in mutually exclusive $if/$else branches
 			// may appear as duplicates since all branches are parsed
-			if is_duplicate && !(p.pref.is_fmt && p.inside_ct_if_expr) {
+			if is_duplicate && !p.pref.is_fmt {
 				if type_sym.kind == .enum
 					&& name in ['is_empty', 'has', 'all', 'set', 'set_all', 'clear', 'clear_all', 'toggle', 'zero', 'from'] {
 					if enum_fn := type_sym.find_method(name) {


vlib/v/parser/fn.v
```diff
@@ -658,7 +658,9 @@ fn (mut p Parser) fn_decl() ast.FnDecl {
 					is_duplicate = !type_sym.info.has_method(name)
 				}
 			}
-			if is_duplicate {
+			// when formatting, methods in mutually exclusive $if/$else branches
+			// may appear as duplicates since all branches are parsed
+			if is_duplicate && !(p.pref.is_fmt && p.inside_ct_if_expr) {
```
spytheman: imho just this is enough:

```suggestion
			if is_duplicate && !p.pref.is_fmt {
```

spytheman: since it is not `vfmt`'s responsibility to report these errors; they are not syntax ones, and parse fine;

ideally, those checks should be done in the checker, not the parser

PR #26285, opened 2026-01-07
cgen: allow non generic sumtypes initialiations to be with HEAP

Eliyaan:


spytheman (approved):
Thank you @Eliyaan 🙇🏻 .

PR #26287, opened 2026-01-07
wasm: fix string plus assign with call expr

Krchi:
fix string plus assign with call expr


spytheman:
(rebased over current master after resolving a merge conflict)

spytheman (approved):
Thank you @Krchi 🙇🏻 .
Excellent work.

PR #26290, opened 2026-01-07
wasm: Handle for in Strings and FixedArrays

SheatNoisette:
Handle the following `for _ in val` statement in the WASM codegen:
```
a := "My insane string"
b := ["a", "b", "c"]!

for v in a {
    print(v)
}

for v in b {
    print(v)
}
```

Only Strings and FixedArrays are supported for now.


spytheman:
Can it be merged as it is, or do you plan more work for the PR?

SheatNoisette:
It depends of your priorities, if your want to move fast and refactor later or not

SheatNoisette:
It has some duplication though, so refactor later and add the feature now or try to improve it now and merge it later. You decide.

spytheman:
Given that there were multiple people that made changes to the same area (the wasm backend) in the last 2 days, I prefer to merge it as early as possible, to minimize the deltas/conflicts that would have to be solved in each case.

PR #26296, opened 2026-01-09
cgen: fix interface pass string ref arg

### LINKED ISSUE #26293: Runtime memory access error I can't explain
### Describe the bug

When running [this test](https://github.com/einar-hjortdal/peony/blob/c8db48b55eae20f3c9308ea70dc42127cba45780/peony_test.v#L152), [this line](https://github.com/einar-hjortdal/peony/blob/c8db48b55eae20f3c9308ea70dc42127cba45780/model_user.v#L105) causes no issue, but [this line](https://github.com/einar-hjortdal/peony/blob/c8db48b55eae20f3c9308ea70dc42127cba45780/model_user.v#L112) causes a runtime memory access error. I believe this is a compiler bug.

### Reproduction Steps

1) clone
2) git checkout c8db48b55eae20f3c9308ea70dc42127cba45780
3) v run peony_test.v

### Expected Behavior

Both println should output the same string

### Current Behavior

second println causes runtime memory access error

### Possible Solution

_No response_

Looks like this bug causes the database to receive bad parameters too: instead of returning the expected result of 1, it returns 0. That is because it doesn't receive the string provided but something else, however I can't output what that something else is, because of the memory bounds violation.

### V version

V 0.5.0 cab281f

|V full version      |V 0.5.0 e3b915dac183902833f2358a6fb76c46d07a52c3.cab281f
|OS                  |linux, "openmamba release 2025.12.1 for x86_64 (rolling)"
|Memory              |10.07GB/31.16GB
|                    |
|V last modified time|2026-01-08 14:54:14
|                    |
|Current working dir |OK, value: /home/einar/Documents/projects/vlang/active/peony
|                    |
|V git status        |weekly.2025.38-682-gcab281ff
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20251019
|gcc version         |gcc (GCC) 15.2.1 20251019
|clang version       |N/A
|emcc version        |N/A
> [!NOTE]
>

kbkpbot:
Closes #26293 

When converting a value-type function argument (e.g., `string`) to an interface type (e.g., `Value`), the compiler incorrectly used direct address-of (&) operation instead of heap allocation. This caused a dangling pointer issue because the function argument is a local variable that gets destroyed when the function returns, but the interface still held a pointer to the invalidated memory, leading to runtime memory access errors.

### DIFF
--- a/vlib/v/tests/interface_string_ref_arg_test.v
+++ b/vlib/v/tests/interface_string_ref_arg_test.v
@@ -7,11 +7,10 @@ interface Value {}
 fn generate_params(s string) []Value {
 	mut params := []Value{}
 	params = arrays.concat(params, s)
-	println(params)
 	return params
 }
 
 fn test_interface_string_ref_arg() {
-	params := generate_params('info@peony.com')
-	println(params)
+	params := generate_params('any_string')
+	assert params == [Value('any_string')]
 }


vlib/v/tests/interface_string_ref_arg_test.v
```diff
@@ -0,0 +1,17 @@
...
+
+fn generate_params(s string) []Value {
+	mut params := []Value{}
+	params = arrays.concat(params, s)
+	println(params)
+	return params
+}
+
+fn test_interface_string_ref_arg() {
+	params := generate_params('info@peony.com')
```
einar-hjortdal: I don't think using an email address (to a domain we do not control) is a good idea
Maybe switch the string to `'kbkpbot'`

vlib/v/tests/interface_string_ref_arg_test.v
spytheman, line 1: Please add an assertion too. Without one, it only checks that the test compiles, but it is not clear what does it do, and what feature is tested.

PR #26303, opened 2026-01-09
Time/remove int_to_byte_array_no_pad

enghitalo:
- No breaking changes
- Include missing test file
- remove int_to_byte_array_no_pad to keep it one single-way
- remove manualfree


enghitalo:
Close until https://github.com/vlang/v/issues/26304 be solved

PR #26307, opened 2026-01-09
cgen: remove no longer required words from c_reserved

dy-tea:
I am keeping `stdout`, `stdin` and `stderr` for now as they are file handles. The "keywords" in `rpcndr.h` which are simply defines should be safe to remove but I am not on windows so these might still fail. 
I am unsure why `link` is a keyword in here, as I haven't found anything that defines it. 
~~The `unix` word is a preprocessor define so it might be better to actually keep it? Not sure about this.~~
Edit: unix is still needed

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -17,16 +17,14 @@ import v.type_resolver
 import sync.pool
 
 // Note: some of the words in c_reserved, are not reserved in C, but are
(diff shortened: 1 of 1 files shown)


dy-tea:
Hmm linker error on windows.

spytheman:
it could be a temporary fluke (sqlite is downloaded by the CI on windows, and thus can fail for unrelated reasons)

PR #26309, opened 2026-01-09
gg: adding rotate ellipse

Linklancien:
Add two function to draw ellipse rotate by some angle
- draw_ellipse_filled_rotate
- draw_ellipse_empty_rotate

 I do not know what test can be added or how to implement them

Also the functions are implement with a clockwise rotation
- may add an anti-clockwise 
- may add another argument, a boolean for witch direction it is

### DIFF
--- a/vlib/gg/testdata/draw_elipses.vv
+++ b/vlib/gg/testdata/draw_elipses.vv
@@ -1,26 +1,18 @@
-module main
-
 import gg
-import math
 
-fn main() {
-	mut context := gg.new_context(
-		width:        600
-		height:       200
-		window_title: 'Ellipses'
-		frame_fn:     frame
-	)
-	context.run()
-}
-
-fn frame(mut ctx gg.Context) {
-	ctx.begin()
-	drot := f32(ctx.frame) / 60
-	ctx.draw_ellipse_filled(100, 100, 100, 50, gg.red)
-	ctx.draw_ellipse_empty(100, 100, 50, 25, gg.black)
-	ctx.draw_ellipse_filled_rotate(300, 100, 100, 50, math.pi / 4 - drot, gg.blue)
-	ctx.draw_ellipse_empty_rotate(300, 100, 50, 25, math.pi / 4 - drot, gg.black)
-	ctx.draw_ellipse_filled_rotate(500, 100, 100, 50, -math.pi / 4 + drot, gg.yellow)
-	ctx.draw_ellipse_empty_rotate(500, 100, 50, 25, -math.pi / 4 + drot, gg.black)
-	ctx.end()
-}
+gg.start(
+	window_title: 'Ellipses'
+	width:        600
+	height:       200
+	frame_fn:     fn (mut ctx gg.Context) {
+		ctx.begin()
+		drot := f32(ctx.frame) / 60
+		ctx.draw_ellipse_filled(100, 100, 100, 50, gg.red)
+		ctx.draw_ellipse_empty(100, 100, 50, 25, gg.black)
+		ctx.draw_ellipse_filled_rotate(300, 100, 100, 50, -drot, gg.blue)
+		ctx.draw_ellipse_empty_rotate(300, 100, 50, 25, -drot, gg.black)
+		ctx.draw_ellipse_filled_rotate(500, 100, 100, 50, drot, gg.yellow)
+		ctx.draw_ellipse_empty_rotate(500, 100, 50, 25, drot, gg.black)
+		ctx.end()
+	}
+)


vlib/gg/draw.c.v
```diff
@@ -1030,6 +1030,61 @@ pub fn (ctx &Context) draw_ellipse_filled(x f32, y f32, rw f32, rh f32, c Color)
...
+	sgl.c4b(c.r, c.g, c.b, c.a)
+
+	sgl.begin_line_strip()
+	for i := 0; i < 360; i += 10 {
+		x_current := math.sinf(f32(math.radians(i))) * rw
+		x_next := math.sinf(f32(math.radians(i + 10))) * rw
+		y_current := math.cosf(f32(math.radians(i))) * rh
+		y_next := math.cosf(f32(math.radians(i + 10))) * rh
+		cos_rot := math.cosf(rota)
+		sin_rot := math.sinf(rota)
```
spytheman: move those before the loop

vlib/gg/draw.c.v
```diff
@@ -1030,6 +1030,61 @@ pub fn (ctx &Context) draw_ellipse_filled(x f32, y f32, rw f32, rh f32, c Color)
...
+
+	sgl.begin_triangle_strip()
+	for i := 0; i < 360; i += 10 {
+		sgl.v2f(x, y)
+		x_current := math.sinf(f32(math.radians(i))) * rw
+		x_next := math.sinf(f32(math.radians(i + 10))) * rw
+		y_current := math.cosf(f32(math.radians(i))) * rh
+		y_next := math.cosf(f32(math.radians(i + 10))) * rh
+		cos_rot := math.cosf(rota)
+		sin_rot := math.sinf(rota)
```
spytheman: move those before the loop - they only depend on `rota` which will not change while the function runs

spytheman:
> I do not know what test can be added or how to implement them

Please draw several ellipses, with different rotations in `vlib/gg/testdata/draw_elipses.vv` .
You can expand the window, and make the new ones non overlapping if you want too.

The .vv files in that folder are compiled by `vlib/gg/draw_fns_api_test.v`, and ensure that there is an easy way to check the output before/after a PR that changes the gg module.

Linklancien:
> > I do not know what test can be added or how to implement them
> 
> Please draw several ellipses, with different rotations in `vlib/gg/testdata/draw_elipses.vv` .
> You can expand the window, and make the new ones non overlapping if you want too.
> 
> The .vv files in that folder are compiled by `vlib/gg/draw_fns_api_test.v`, and ensure that there is an easy way to check the output before/after a PR that changes the gg module.

Okay

vlib/gg/draw.c.v
```diff
@@ -1030,6 +1030,63 @@ pub fn (ctx &Context) draw_ellipse_filled(x f32, y f32, rw f32, rh f32, c Color)
...
 }
 
+// draw_ellipse_empty_rotate draws the outline of an ellipse.
+// `x`,`y` defines the center of the ellipse.
+// `rw` defines the *width* radius of the ellipse.
+// `rh` defines the *height* radius of the ellipse.
+// `rota`  define the *rotation* of the ellipse.
+// `is_clockwise` defines in witch direction the ellipse is rotate.
+// `c` is the color of the outline.
+pub fn (ctx &Context) draw_ellipse_empty_rotate(x f32, y f32, rw f32, rh f32, rota f32, is_clockwise bool, c Color) {
```
spytheman: there is no need for the `is_clockwise` parameter; the user can already pass a negative `rota` angle, to achieve the same

spytheman: i.e. `sin(-angle) == -sin(angle)`

vlib/gg/draw.c.v
```diff
@@ -1030,6 +1030,63 @@ pub fn (ctx &Context) draw_ellipse_filled(x f32, y f32, rw f32, rh f32, c Color)
 	sgl.end()
 }
 
+// draw_ellipse_empty_rotate draws the outline of an ellipse.
+// `x`,`y` defines the center of the ellipse.
+// `rw` defines the *width* radius of the ellipse.
+// `rh` defines the *height* radius of the ellipse.
+// `rota`  define the *rotation* of the ellipse.
+// `is_clockwise` defines in witch direction the ellipse is rotate.
```
spytheman:
```suggestion
```

vlib/gg/draw.c.v
```diff
@@ -1030,6 +1030,63 @@ pub fn (ctx &Context) draw_ellipse_filled(x f32, y f32, rw f32, rh f32, c Color)
 	sgl.end()
 }
 
+// draw_ellipse_empty_rotate draws the outline of an ellipse.
+// `x`,`y` defines the center of the ellipse.
+// `rw` defines the *width* radius of the ellipse.
+// `rh` defines the *height* radius of the ellipse.
+// `rota`  define the *rotation* of the ellipse.
```
spytheman:
```suggestion
// `rota` defines the *rotation* angle of the ellipse, in radians.
```

vlib/gg/draw.c.v
```diff
@@ -1030,6 +1030,63 @@ pub fn (ctx &Context) draw_ellipse_filled(x f32, y f32, rw f32, rh f32, c Color)
...
+	}
+	sgl.end()
+}
+
+// draw_ellipse_filled draws an opaque ellipse.
+// `x`,`y` defines the center of the ellipse.
+// `rw` defines the *width* radius of the ellipse.
+// `rh` defines the *height* radius of the ellipse.
+// `rota`  define the *rotation* of the ellipse.
+// `is_clockwise` defines in witch direction the ellipse is rotate.
```
spytheman:
```suggestion
// `rota` defines the *rotation* angle of the ellipse, in radians.
```

Linklancien:
I just think about something, if rot is define in radians, shouldn't the rotation follows the trigonometric direction
This would juste change ``sin_rot := math.sinf(rota)`` by ``sin_rot := -math.sinf(rota)``
Let me know what do you think

PR #26313, opened 2026-01-10
transformer: Resolve wrong cast fixed ArrayInit to voidptr

SheatNoisette:
This PR attempt to correct the wrong cast done in fixed Arrays using the newer "transformer". This fixes the tests in `vlib/crypto/ed25519/internal/`.

### DIFF
--- a/vlib/v/transformer/array.v
+++ b/vlib/v/transformer/array.v
@@ -53,10 +53,17 @@ pub fn (mut t Transformer) array_init(mut node ast.ArrayInit) ast.Expr {
 		}
 		typ:  ast.int_type
 	}
+	elem_sym := t.table.sym(t.table.unwrap(node.elem_type))
+	// For function types, use voidptr size instead of function type size
+	sizeof_type := if elem_sym.kind == .function {
+		ast.voidptr_type
+	} else {
+		node.elem_type
+	}
 	sizeof_arg := ast.CallArg{
 		expr: ast.SizeOf{
 			is_type: true
-			typ:     node.elem_type
+			typ:     sizeof_type
 		}
 		typ:  ast.int_type
 	}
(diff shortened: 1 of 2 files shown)


vlib/v/transformer/array.v
```diff
@@ -53,10 +53,17 @@ pub fn (mut t Transformer) array_init(mut node ast.ArrayInit) ast.Expr {
 		}
 		typ:  ast.int_type
 	}
+	elem_sym := t.table.sym(t.table.unwrap(node.elem_type))
+	// For function types, use voidptr size instead of function type size
```
spytheman, line 57: hm, why should be this special cased?
sizeof(fntype) should be the same as sizeof(voidptr) in all cases afaik already

spytheman:
```
0[09:13:48]@m1: (master) /opt/v $ cat s.v
type FnType = fn ()
dump(sizeof[FnType]())
dump(sizeof(voidptr))
0[09:13:49]@m1: (master) /opt/v $ ./v -gc none -cc tcc run s.v
[s.v:2] sizeof(fn ()): 8
[s.v:3] sizeof(voidptr): 8
0[09:13:53]@m1: (master) /opt/v $ ./v -gc none -cc clang run s.v
[s.v:2] sizeof(fn ()): 8
[s.v:3] sizeof(voidptr): 8
0[09:13:56]@m1: (master) /opt/v $ ./v -gc none -cc gcc run s.v
[s.v:2] sizeof(fn ()): 8
[s.v:3] sizeof(voidptr): 8
0[09:13:58]@m1: (master) /opt/v $
```

SheatNoisette: I attempted to match the C output of the old transformer. There not much logic whatsoever unfortunately.

spytheman:
hm, why is this closed?

SheatNoisette:
I don't believe that this is a good patch.

PR #26328, opened 2026-01-12
cgen: fix call with fixed array of builtin struct

Krchi:
Fix call with fixed array of builtin struct


vlib/v/gen/c/cgen.v
```diff
@@ -1763,7 +1763,7 @@ pub fn (mut g Gen) write_typedef_types() {
 			.array_fixed {
 				info := sym.info as ast.ArrayFixed
 				elem_sym := g.table.sym(info.elem_type)
-				if elem_sym.is_builtin() {
+				if elem_sym.is_builtin() && elem_sym.kind != .struct {
```
spytheman: please swap the operands; the != check is a bit cheaper than a fn call

Issue #25706, opened 2025-11-10
Implement JSON-RPC Module in V Standard Library


kbkpbot:
### Describe the feature

Propose adding a comprehensive JSON-RPC 2.0 implementation to V's standard library as a standalone module at `vlib/jsonrpc`.

### Use Case

`JSON-RPC` is a lightweight remote procedure call protocol that is widely used in:
- Language Server Protocol (LSP) implementations
- Blockchain and cryptocurrency APIs (Ethereum, Bitcoin)
- Microservices communication
- IoT device management
- AI/ML tooling (Model Context Protocol, OpenAI APIs)

Currently, V lacks a standardized way to handle `JSON-RPC`, forcing developers to implement protocol logic repeatedly. A standard implementation would:
- Enable better tooling (language servers, debug adapters)
- Facilitate integration with existing JSON-RPC-based services
- Provide foundation for higher-level protocols (like MCP)
- Reduce boilerplate in networked applications

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.4.12 33e129b19e1124bb0cb6b057ee04324cc25409b4

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.12 33e129b19e1124bb0cb6b057ee04324cc25409b4
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 20.04.6 LTS
|Processor           |4 cpus, 64bit, little endian, aarch64
|Memory              |1.65GB/3.78GB
|                    |
|V executable        |/home/vpn/github/lang/v/v
|V last modified time|2025-11-10 02:28:38
|                    |
|V home dir          |OK, value: /home/vpn/github/lang/v
|VMODULES            |OK, value: /home/vpn/.vmodules
|VTMP                |OK, value: /tmp/v_1001
|Current working dir |OK, value: /home/vpn/github/lang/v/vlib
|                    |
|Git version         |git version 2.25.1
|V git status        |weekly.2025.45-35-g33e129b1
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
|gcc version         |gcc (Ubuntu 10.5.0-1ubuntu1~20.04) 10.5.0
|clang version       |clang version 10.0.0-4ubuntu1
|tcc version         |tcc version 0.9.27 mob:da11cf6 (AArch64 Linux)
[...]

PR #26330, opened 2026-01-12
net: add jsonrpc module

Te4nick:
Adding JSON-RPC 2.0 implementation

## Limitaions
- Request/Response use only string id
- JSON-RPC 1.0 incompatible

## Features
- Request/Response single/batch json encoding/decoding
- Server to work with any io.ReaderWriter
- Server automatically manages batch Requests and builds batch Response
- Client to work with any io.ReaderWriter
- Interceptors for custom events, raw Request, Request, Response, raw Response

### DIFF
--- a/.github/workflows/freebsd_ci.yml
+++ b/.github/workflows/freebsd_ci.yml
@@ -39,6 +39,10 @@ jobs:
           sync_files: runner-to-vm
           run: |
             sudo pkg install -y git sqlite3 gmake boehm-gc-threaded libiconv
+            # Install DB packages needed to build examples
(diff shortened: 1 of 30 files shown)


Jengro777:
https://github.com/vlang/v/issues/25706

medvednikov:
Thanks a lot!

medvednikov:
Turns out we did have this btw:

https://github.com/vlang/v-analyzer/tree/main/src/jsonrpc

medvednikov:
Maybe some logic can be taken from there.

PR #26334, opened 2026-01-12
db.mysql: improve documentation for support on FreeBSD

lcheylus:
- `vlib/db/mysql/README.md`: add doc to install FreeBSD package needed for MariaDB/MySQL lib and includes
- `vlib/db/mysql/_cdefs_nix.c.v`: add FreeBSD case to display error message for include `mysql.h`

### DIFF
--- a/vlib/db/mysql/README.md
+++ b/vlib/db/mysql/README.md
@@ -24,6 +24,8 @@ Use `docker container rm some-mysql` to remove it completely, after it is stoppe
 
 For Linux, you need to install `MySQL development` package and `pkg-config`.
 
+For FreeBSD, you need to install the `mariadb118-client` package.
+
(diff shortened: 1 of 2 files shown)


vlib/db/mysql/README.md
```diff
@@ -24,6 +24,8 @@ Use `docker container rm some-mysql` to remove it completely, after it is stoppe
 
 For Linux, you need to install `MySQL development` package and `pkg-config`.
 
+For FreeBSD, you need to install the `mariadb118-client` package.
```
spytheman, line 27: Given this name, should not the message for the missing mysql.h header change too?

lcheylus: OK, modified error message in `vlib/db/mysql/_cdefs_nix.c.v` for `freebsd` case.

PR #26341, opened 2026-01-13
transformer,markused,cgen: fix some problem for array init when use `-new-transformer`

Krchi:
1. fix cast for fixed array of fn, type is fn name, which sizeof(c_function_name) = 1 in c
2. disable in js backend, js do not have fn new_array_from_c_array
3. fix fixed array undefined in comptimecall, traverse remained ast node in markused
4. fix cast twice for fixed array, set inside_cast to false
5. fix array translated in or_expr, call_expr need set is_return_used
6. fix with array and shared array, think of &call() is not allow, now `shared arr := [1, 2, 3]` and `shared arr1 := &[1, 2, 3]` generate same c code, which different with array

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5864,13 +5864,16 @@ fn (mut g Gen) cast_expr(node ast.CastExpr) {
 						g.write('&'.repeat(ptr_cnt))
 					}
 				}
+				old_inside_cast := g.inside_cast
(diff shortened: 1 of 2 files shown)


spytheman:
Thank you @Krchi . What are the stats now?

Krchi:
> Thank you @Krchi . What are the stats now?

9 failed in windows, 18 failed in git codespaces, 9 failed in mac

spytheman:
Excellent.

Issue #24575, opened 2025-05-25
$compile_error should locate the actual location of the error


kbkpbot:
### Describe the bug

Currently , `$compile_error()` will point to where it is called, not the real location of error.

### Reproduction Steps

compile_error.v
```v
module main

fn son[T](val T) {
        $if T !is int {
                $compile_error('son only taken int as input')
        }
}

fn main() {
        son(false)
}
```

```
v compile_error.v
```

### Expected Behavior

something like this:
```
$ v compile_error.v
compile_error.v:5:3: error: son only taken int as input
    9 | fn main() {
   10 |        son(false)
        |~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   11 |}
```

### Current Behavior

```
$ v compile_error.v
compile_error.v:5:3: error: son only taken int as input
    3 | fn son[T](val T) {
    4 |     $if T !is int {
    5 |         $compile_error('son only taken int as input')
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    6 |     }
    7 | }
```

### Additional Information/Context

Found be PR #24573

### V version

V 0.4.10 ed7a8a34f1815816984aaa1d2982d2a5fb9a97f2.52c7130

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 ed7a8a34f1815816984aaa1d2982d2a5fb9a97f2.52c7130
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.2 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |6.98GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
|V last modified time|2025-05-25 10:51:48
|                    |
|V home dir          |OK, value: /media/HD/github/kbkpbot/v
|VMODULES            |OK, value: /home/mars/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/mars/v/bug
|                    |
|Git version         |git version 2.43.0
|V git status        |weekly.2025.17-162-g0f4e12e7
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
[...]

Issue #16127, opened 2022-10-20
'$tmpl' error message improvement


kahsa:
**V version: 0.3.1 f8a28b5** 
**OS: ArchLinux** 

**What did you do?**
```v
fn foo(){
	name := 'abc'
	$tmpl('template.txt')
}

fn bar(){
	$tmpl('template.txt')
}

fn main(){
}
```

**What did you expect to see?**
Line number in V file and template file.
 
**What did you see instead?**
Only template file line number.
```
> v run .
template.txt:9:32: error: undefined ident: `name`
```

bogen85:
There are other confusing error message for `$tmpl`. See #16154

esquerbatua:
With the last version of v gets:
```
template.txt:9:32: error: undefined ident: `name` (veb action: main__bar)
template.txt:9:32: error: expression does not return a value (veb action: main__bar)
make: *** [Makefile:11: run] Error 1
```

PR #26356, opened 2026-01-14
checker: add error message call stack support

kbkpbot:
Feature request by #16127, #24575

Currently, support `$compile_error()` and `$tmpl` call stack showing in error message:
```
vlib/v/checker/tests/compile_error_call_position.vv:5:3: error: son only taken int as input
    3 | fn son[T](val T) {
    4 |     $if T !is int {
    5 |         $compile_error('son only taken int as input')
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    6 |     }
    7 | }
called from vlib/v/checker/tests/compile_error_call_position.vv:10:2
    8 |
    9 | fn main() {
   10 |     son(false)
      |     ~~~~~~~~~~
   11 | }
```

```
template_call_position_test.txt:11:2: error: undefined ident: `unknown_var` (veb action: main__main)
called from vlib/v/checker/tests/template_call_position.vv:4:2
    2 |
    3 | fn main() {
    4 |     $tmpl('template_call_position_test.txt')
      |     ^
    5 | }
template_call_position_test.txt:11:2: error: expression does not return a value (veb action: main__main)
called from vlib/v/checker/tests/template_call_position.vv:4:2
    2 |
    3 | fn main() {
    4 |     $tmpl('template_call_position_test.txt')
      |     ^
    5 | }
```

### DIFF
--- a/vlib/v/checker/tests/compile_error_explicit_type.out
+++ b/vlib/v/checker/tests/compile_error_explicit_type.out
@@ -1,13 +1,14 @@
-vlib/v/checker/tests/compile_error_explicit_type.vv:5:3: error: son only taken int as input
+vlib/v/checker/tests/compile_error_explicit_type.vv:5:3: error: son only take `int` as input
     3 | fn son[T](val T) {
     4 |     $if T !is int {
-    5 |         $compile_error('son only taken int as input')
-      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    5 |         $compile_error('son only take `int` as input')
+      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     6 |     }
     7 | }
-called from vlib/v/checker/tests/compile_error_explicit_type.vv:10:2
-    8 | 
+called from vlib/v/checker/tests/compile_error_explicit_type.vv:11:2
     9 | fn main() {
-   10 |     son[bool](false)
+   10 |     son[int](123)
+   11 |     son[bool](false)
       |     ~~~~~~~~~~~~~~~~
-   11 | }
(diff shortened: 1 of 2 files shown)


vlib/v/checker/check_types.v
```diff
@@ -1282,6 +1282,11 @@ fn (mut c Checker) infer_fn_generic_types(func &ast.Fn, mut node ast.CallExpr) {
 	if c.table.register_fn_concrete_types(func.fkey(), inferred_types) {
 		c.need_recheck_generic_fns = true
 	}
+	// Record the call position for better error reporting in `$compile_error()`
+	if c.generic_call_positions.len == 0 {
+		c.generic_call_positions = map[string]token.Pos{}
+	}
```
spytheman: that seems to be a nop to me 🤔

vlib/v/checker/tests/comptime_else_compile_error_no_return.out
```diff
@@ -4,4 +4,10 @@ vlib/v/checker/tests/comptime_else_compile_error_no_return.vv:5:3: error: not an
     5 |         $compile_error('not an int')
       |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     6 |     }
-    7 | }
```
kbkpbot, line 7: why format is incorrect here? :(

spytheman:
Excellent work.

Can this be extended in the future, to show the callstack for errors, that happen in nested generic calls like in `x.json2` ?

spytheman (approved):

spytheman:
For example, I tried to update https://github.com/herudi/cdv to work with latest V a few days ago, but got:
image

with no easy way to know what and where was the original call that triggered each x.json2 error ...

PR #26268, opened 2026-01-05
wasm: fix type_size and precompute symbol sizes, alignment and offset of fields in structs

SurmanPP:
Previously type_size would incorrectly compute sizes for pointers to structs in struct fields as if they were structs themselves and not pointers to structs leading to an infinite loop for structs containing pointers to themselves besides not handling pointers correctly. Offsets of fields in structs are needed to allocate memory and be able to access these fields these are precomputed now together with symbol sizes and their alignment like it is in the native backend. The code for the function type_size is adapted from the almost equivalent function in the ast module.

Handling structs which contain references to themself no longer segfaults the compiler.

```v
struct Foo {
   i_crash_the_compile &Foo
   field int
}

fn (foo &Foo) print_field() {
   println(foo.field)
}

PR #26357, opened 2026-01-14
wasm: fix pointers in structs

SurmanPP:
A redo of the pr #26268

This pr does effectively the same with less changes(for better or for worse). Unrelated changes are excluded too(if those are any good they should be in separate prs).

Pointers in structs were not handled correctly in the size and offset calculations on the wasm backend.

I added a test to v/gen/wasm/tests/misc.vv
Accessing fields from pointers to structs(have not tested) or fields from a pointer to a struct in a struct(have tested) does not seem to work correctly.
Thats why the misc.vv test has some weird casts. Actually I think that it does work for the wrong reason that is that the field "a" is at the same address as the struct "same". My assumption is that "same" for some reason it is not dereferenced and the field accessed because doing something like println(unsafe { int(voidptr(s.same.msg)) }) prints the same address as it would without these casts(if one were to read the value at that address one would see "1234"). 

The problems with vdoc and the previous pr do not appear with these changes(vdoc still has a bug).

### DIFF
--- a/vlib/v/gen/wasm/serialise/serialise.v
+++ b/vlib/v/gen/wasm/serialise/serialise.v
@@ -58,9 +58,9 @@ pub fn (mut p Pool) type_struct_info(typ ast.Type) ?StructInfo {
 }
 
 pub fn (mut p Pool) type_size(typ ast.Type) (int, int) {
-  if typ.nr_muls() > 0 {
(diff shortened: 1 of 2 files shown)


spytheman (approved):
Very elegant work.
Thank you @SurmanPP 🙇🏻 .

PR #26360, opened 2026-01-15
math: fix acosh() for 0.0

tankf33der:
Among other my repos with tests for different modules, there is a repo with tests for `math` against the `mpfr` library. 
I was able to integrate tests for 37 functions.
Everything was already fixed, it came down to edge case tests.
I concentrated on tests with zero input parameter. 
In the test suite, only `acosh()` showed up.

For `numpy` it would be like this:
```
$ python
Python 3.12.12 (main, Oct 11 2025, 01:16:26) [GCC 15.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> numpy.acosh(0.0)
np.float64(nan)
```

### DIFF
--- a/vlib/math/math_test.v
+++ b/vlib/math/math_test.v
@@ -231,8 +231,8 @@ fn test_acosh() {
 		f := acosh(a)
 		assert veryclose(acosh_[i], f)
 	}
-	vfacosh_sc_ := [inf(-1), 0.5, 1, inf(1), nan()]
-	acosh_sc_ := [nan(), nan(), 0, inf(1), nan()]
+	vfacosh_sc_ := [inf(-1), 0.5, 0.0, 1, inf(1), nan()]
+	acosh_sc_ := [nan(), nan(), nan(), 0, inf(1), nan()]
 	for i := 0; i < vfacosh_sc_.len; i++ {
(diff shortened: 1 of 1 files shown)


vlib/math/invhyp.v
```diff
@@ -5,7 +5,7 @@ import math.internal
 // acosh returns the non-negative area hyperbolic cosine of x
 pub fn acosh(x f64) f64 {
 	if x == 0.0 {
-		return 0.0
+		return nan()
 	} else if x > 1.0 / internal.sqrt_f64_epsilon {
```
spytheman: would it not be simpler, to just remove the special case here, and let it fall to the default at the end, that just does `return nan()` ?

tankf33der: Of course you're right. If you noticed it too, then it definitely needs to be fixed. Sometimes I'm too gentle with other people's code.
I committed a fix.

Delta456 (commented):
Can you add a test for this?

tankf33der:
@Delta456 

> Can you add a test for this?

Sure. Committed.

tankf33der:
Now I have implemented full edge case coverage for 37 functions and my tests caught 5 more functions with errors. I will make some noise.

PR #26362, opened 2026-01-15
cgen: add rv32 inline assembly support

matthiaaas:
This change builds upon #26050, which adds identical support but for RV32.


vlib/v/gen/c/cgen.v
```diff
@@ -3449,7 +3447,7 @@ fn (mut g Gen) asm_arg(arg ast.AsmArg, stmt ast.AsmStmt) {
 			g.write('\$${arg.val.str()}')
 		}
 		ast.AsmRegister {
-			if stmt.arch == .rv64 {
+			if stmt.arch == .rv64 || stmt.arch == .rv32 {
```
JalonSolov:
```suggestion
			if stmt.arch in [.rv64, .rv32] {
```
...for consistency with your other changes.

PR #26371, opened 2026-01-16
vtimer: Self-Contained Lightweight timer, thread safe

quaesitor-scientiam:
Surfaces public functions for ad-hoc timings, supports multiple platforms, provides formatted output

Tests added via `vtimer_test.v`, works within the the existing `vlib/time` module

### DIFF
--- a/vlib/time/vtimer.v
+++ b/vlib/time/vtimer.v
@@ -152,6 +152,7 @@ pub fn (mut t Timer) start() {
 	t.running = true
 }
 
+// stop - Atomically set end_t to current counter
 pub fn (mut t Timer) stop() {
 	now := get_monotonic_time()
 	stdatomic.store_i64(&t.end_t, now)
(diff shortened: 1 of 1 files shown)


lcheylus:
1/ Please rename your "module" as `vtimer` => every V module in stdlib uses **lowercase name**.
2/ To add new functions in V stdlib, please provide inline documentation for each one. Your first commit for `VTimer.v` **has no documentation** :(

quaesitor-scientiam:
> 1/ Please rename your "module" as `vtimer` => every V module in stdlib uses **lowercase name**. 2/ To add new functions in V stdlib, please provide inline documentation for each one. Your first commit for `VTimer.v` **has no documentation** :(

Pub functions documented

spytheman:
@quaesitor-scientiam sorry, but I think this is more suited for a separate repo, outside of vlib.

I would like to keep core modules like `builtin`, `strings` and `strconv` as light as possible, since they affect the bootstrapping of the compiler itself on existing and new platforms in non trivial ways. That restriction does not apply for other modules, or for external modules that are not part of vlib.

Also with the current approach in the PR, it will be hard to make it work properly in all cases (with the .js and .wasm backends for example, as illustrated by the above error) - which will not be a problem, if you are not concerned with them (and modules outside of vlib do not need to be). It also introduces an external dependency, on a module that is otherwise not needed anywhere else in time (needlessly imho - measuring time in my experience is usually best done within threads, not across them). It also duplicates some of the functionality of the existing time.StopWatch .

PR #26376, opened 2026-01-17
new generic stage: more tests passing for vlib/v/tests/generics

Eliyaan:
- remove wrong line (gkind_field)
- solve T.name
- check if type implements interface
- do not modify method name with struct's generic type

Fixes more tests from `v -new-generic-solver test vlib/v/tests/generics/`, status : `69 failed, 192 passed, 261 total` compared to `80 failed, 181 passed, 261 total` for the previous PR

### DIFF
--- a/vlib/v/generics/generics.v
+++ b/vlib/v/generics/generics.v
@@ -16,7 +16,6 @@ pub mut:
 	styp_cache          map[ast.Type]string
 	cur_fn              &ast.FnDecl = unsafe { nil }
 	cur_concrete_types  []ast.Type
-	cur_generic_names   []string
 	inside_struct_init  bool
 	cur_struct_init_typ ast.Type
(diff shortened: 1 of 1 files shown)


vlib/v/gen/c/cgen.v
```diff
@@ -8302,7 +8302,7 @@ return ${cast_shared_struct_str};
 			method_names := methods.map(it.name)
 			match st_sym.info {
 				ast.Struct, ast.Interface, ast.SumType {
-					if st_sym.info.parent_type.has_flag(.generic) {
+					if st_sym.info.parent_type.has_flag(.generic) && !g.pref.new_generic_solver {
```
spytheman, line 8305:
```suggestion
					if !g.pref.new_generic_solver && st_sym.info.parent_type.has_flag(.generic) {
```

PR #26381, opened 2026-01-17
math: rewrite round(), update test

tankf33der:
Now `round()` can not pass multiple tests vs. `mpfr`.
The simplest thing was to update and rewrite the implementation from [golang](https://github.com/golang/go/blob/master/src/math/floor.go#L90).
Now `round()` and `round_to_even()` are similar and both translated from golang.

I added one of the values that doesn't match `mpfr` to the tests, the test value were found manually.

Special note:This is the last PR from my testing of the `math` module vs. `mpfr`. In one of the previous PRs I wrote that the number of PRs would be greater, but it turned out that the tests don't pass not because of an error, but because these are special cases and it's documented.

### DIFF
--- a/vlib/math/floor.v
+++ b/vlib/math/floor.v
@@ -77,25 +77,25 @@ pub fn trunc(x f64) f64 {
 // round(±inf) = ±inf
 // round(nan) = nan
 pub fn round(x f64) f64 {
-	if x == 0 || is_nan(x) || is_inf(x, 0) {
-		return x
-	}
-	// Largest integer <= x
-	mut y := floor(x) // Fractional part
-	mut r := x - y // Round up to nearest.
-	if r > 0.5 {
-		y += 1.0
-		return y
-	}
-	// Round to even
-	if r == 0.5 {
-		r = y - 2.0 * floor(0.5 * y)
-		if r == 1.0 {
-			y += 1.0
+	mut bits := f64_bits(x)
+	mut e_ := (bits >> shift) & mask
+	if e_ < bias {
+		// Round abs(x) < 1 including denormals.
+		bits &= sign_mask // +-0
+		if e_ == bias - 1 {
+			bits |= uvone // +-1
 		}
+	} else if e_ < bias + shift {
+		// Round any abs(x) >= 1 containing a fractional component [0,1).
+		//
+		// Numbers with larger exponents are returned unchanged since they
+		// must be either an integer, infinity, or NaN.
+		half := u64(1) << (shift - 1)
+		e_ -= bias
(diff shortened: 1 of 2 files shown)


vlib/math/floor.v
```diff
@@ -77,25 +77,25 @@ pub fn trunc(x f64) f64 {
...
 		}
+	} else if e_ < bias + shift {
+		// Round any abs(x) >= 1 containing a fractional component [0,1).
+		//
+		// Numbers with larger exponents are returned unchanged since they
+		// must be either an integer, infinity, or NaN.
+		half := u64(1) << (shift - 1)
+		e_ -= bias
+		bits += half >> e_
+		bits &= ~(frac_mask >> e_)
```
spytheman, line 96: Would not that have the same problem as `round_to_even` did, if `e_` becomes > 63 ?

spytheman:
image

On master, V and Python agree, while here, the result for `math.round(-0.5)` is different, and `math.round(-0.49999)` is `-0.0` ...

spytheman:
It is consistent with Go though on this PR:
image

spytheman:
Can we keep the old version just renamed to for example `round_half_to_even` ?

I am looking at https://en.wikipedia.org/wiki/Rounding#Round_half_to_even and it seems that Python may be using that since version 3 🤔 .

spytheman:
It is definitely more consistent on this PR (when it comes to the sign of the result):
image

tested with:
```v
import math

//
//
//
//
//
//
//
//

dump(math.round(-0.50001))
dump(math.round(-0.50000))
dump(math.round(-0.49999))
dump(math.round(-0.10000))
dump(math.round(-0.00000))

dump(math.round(0.00000))
dump(math.round(0.10000))
dump(math.round(0.49999))
dump(math.round(0.50000))
dump(math.round(0.50001))
```

spytheman:
python3 produces the following for this sequence:
```py
print(round(-0.50001))
print(round(-0.50000))
print(round(-0.49999))
print(round(-0.10000))
print(round(-0.00000))

print(round(0.00000))
print(round(0.10000))
print(round(0.49999))
print(round(0.50000))
print(round(0.50001))
```
```
-1
0
0
0
0
0
0
0
0
1
```

vlib/math/floor.v
tankf33der: I already checked and don't see large values of `e_`.

tankf33der:
`mpfr` returns this:
```
-1.0
-1.0
-0.0
-0.0
-0.0
0.0
0.0
0.0
1.0
1.0
```

PR #26384, opened 2026-01-18
net.http: fix HTTPS proxy connection and add Gemini streaming example

whiter001:
Description:
   - Fix: Corrected the CONNECT request in http_proxy.v to use the relative path instead of the full URL, resolving HTTP/0.9
     errors on some proxies.
   - Improvement: Enhanced buffering in ssl_dial to read byte-by-byte until the headers end, preventing data loss during the SSL
     handshake.
   - Fix: Added a guard to on_progress_body to ensure it's only called after the body start has been identified.
   - Example: Added examples/gemini/gemini.v demonstrating a high-performance streaming AI client using native V.


spytheman:
That is AI slop, with no clear purpose and goal.
Please be more thorough and follow CONTRIBUTING.md .

PR #26386, opened 2026-01-18
wasm: only compile .wasm.v, .v.v and .v files

SurmanPP:
Compiling with '-b wasm' only compiles the aforementioned type of files with this pr. Previously .c.v or .js.v files were included in the compilation process.

### DIFF
--- a/vlib/v/pref/should_compile.v
+++ b/vlib/v/pref/should_compile.v
@@ -292,7 +292,7 @@ pub fn (prefs &Preferences) should_compile_js(file string) bool {
 
 pub fn (prefs &Preferences) should_compile_wasm(file string) bool {
 	if !file.ends_with('.wasm.v') && file.count('.') >= 2 {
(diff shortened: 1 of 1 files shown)


vlib/v/pref/should_compile.v
```diff
@@ -287,6 +290,14 @@ pub fn (prefs &Preferences) should_compile_js(file string) bool {
 	return true
 }
 
+pub fn (prefs &Preferences) should_compile_wasm(file string) bool {
+	if !file.ends_with('.wasm.v') && !file.ends_with('.v.v') && file.count('.') >= 2 {
```
spytheman: There is no separate support for a `.v.v` suffix currently - that is just `.v` - such files get used by *all* backends .

SurmanPP: Whoopsie I confused some thing.

spytheman (approved):
Thank you @SurmanPP 🙇🏻 .
Excellent work.

PR #26387, opened 2026-01-18
strings: add `lorum` Markov chain text generator 

mike-ward:
This PR integrates `lorum` into `vlib/strings`, refining the implementation to be suitable for the standard library.

**Changes:**
- **Removes `rand` dependency:** Implements a simple internal LCG (Linear Congruential Generator) to handle pseudo-randomness. This prevents circular dependencies (e.g., `strings` -> `rand` -> `time`), which is critical for `vlib` core modules.
- **Deterministic by default:** The generator now defaults to a fixed seed if none is provided, ensuring reproducible output for tests and simple calls.
- **Renames API:** Renames `lorem_generate()` to `lorum()` for consistency with the file name and conciseness.
- **Updates Example:** Refactors `examples/lorem.v` to use the new `strings.lorum()` API. The example now handles random seeding explicitly using `rand` and `time` to maintain the expected CLI behavior (random output by default).

**Verification:**
- `v test vlib/strings/lorum_test.v` passes.
- `v run examples/lorem.v` runs successfully and produces varied output.

### DIFF
--- a/examples/lorem.v
+++ b/examples/lorem.v
@@ -14,7 +14,7 @@ Example:
 
     ./lorem -order 2 -words 12 -sentences 4 -paragraphs 3 -corpus poe
 */
-import strings
+import strings.lorem
 import flag
 import os
 import rand
@@ -45,7 +45,7 @@ fn main() {
 		rng_seed = rand.int()
 	}
 
-	cfg := strings.LoremCfg{
+	text := lorem.generate(
 		markov_order:            order
 		words_per_sentence:      words_per_sentence
 		sentences_per_paragraph: sentences_per_paragraph
@@ -53,8 +53,7 @@ fn main() {
 		corpus_name:             corpus_name
 		seed_text:               seed_text
 		rng_seed:                rng_seed
-	}
+	)
 
-	text := strings.lorum(cfg)
 	println(text)
 }
--- a/vlib/strings/lorem/lorem.v
+++ b/vlib/strings/lorem/lorem.v
@@ -1,4 +1,4 @@
-module strings
+module lorem
 
 const lorem_corpora = {
 	'lorem':  lorem_corpus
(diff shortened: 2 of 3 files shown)


vlib/strings/lorem/lorem.v
spytheman, line 1: `strings`, and `strconv` are imported by `builtin` and become part of the compilation of pretty much every V program (that is later trimmed down by the compiler's markused stage, but still, it is better for those modules to be as slim as possible, to simplify bootstrapping, and speed up compilation of small programs, that will not use that new functionality).

That said, there is no problem for there to exist a `strings.lorem` module, i.e. this file can become `vlib/strings/lorem/lorem.v` and then it can also import `rand` without a a problem too, since there will be no cycle involved...

I would also suggest a rename of the API to `
pub fn generate(cfg LoremCfg) string {` so that users can then:
`import strings.lorem`
`println( lorem.generate() )`

mike-ward: Good suggestion. I'll update PR accordingly.

JalonSolov:
Just curious... why change lor**e**m to lor**u**m ?

mike-ward:
> Just curious... why change lor**e**m to lor**u**m ?

That was a goof. Fixed it up in the most recent commit.

examples/lorem.v
```diff
@@ -106,17 +30,23 @@ fn main() {
...
 		eprintln(err)
 		return
 	}
 
-	text := generate_text(
-		order:                   order
+	if rng_seed == 0 {
+		t := time.now().unix_milli()
+		rand.seed([u32(t), u32(t >> 32)])
+		rng_seed = rand.int()
```
spytheman, line 45:
```suggestion
		rng_seed = rand.int()
```

spytheman: the default random number generator is already time seeded by default

vlib/strings/lorem/lorem_test.v
```diff
@@ -0,0 +1,66 @@
+module lorem
```
spytheman, line 1:
```suggestion
import strings.lorem
```

spytheman, line 1:
```suggestion
import strings.lorem
```

Turning this into a normal (non-internal) test, has the benefit that all of the tests also become usage examples, and as a user, you can also more easily see, what the public (and well tested) APIs are.

spytheman: afaik only `lorem_vary` and `lorem_tokenize` are tested here, and are private?

mike-ward:
Thanks guys. I thought I had done my due diligence to make this easy to merge. I'll do better next time...

PR #26391, opened 2026-01-19
new generic stage: more fixes

Eliyaan:
- fix match branch cloning
- better support for aliased types
- support more comptime type conditions
- fix some failures related to markused

Fixes more tests from `v -new-generic-solver test vlib/v/tests/generics/`, status : `53 failed, 208 passed, 261 total` compared to `69 failed, 192 passed, 261 total` for the previous PR

### DIFF
--- a/vlib/v/generics/generics.v
+++ b/vlib/v/generics/generics.v
@@ -452,11 +452,14 @@ fn (mut g Generics) cc_type(typ ast.Type, is_prefix_struct bool) string {
 	return styp
 }
 
-pub fn (mut g Generics) method_concrete_name(old_name string, concrete_types []ast.Type, receiver_type ast.Type) string {
+pub fn (mut g Generics) method_concrete_name(old_name string, concrete_types []ast.Type, _receiver_type ast.Type) string {
 	mut name := old_name
-	if receiver_type != 0 {
-		info := g.table.sym(g.unwrap_generic(receiver_type)).info
-		if info is ast.Struct {
+	if _receiver_type != 0 {
+		mut info := g.table.sym(g.unwrap_generic(_receiver_type)).info
+		if mut info is ast.Alias {
+			info = g.table.sym(g.table.unaliased_type(g.unwrap_generic(_receiver_type))).info
+		}
+		if mut info is ast.Struct {
 			fn_conc_types := concrete_types#[info.generic_types.len..] // concrete types without the generic types of the struct
 
 			if fn_conc_types.len > 0 {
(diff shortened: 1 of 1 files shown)


vlib/v/gen/c/auto_str_methods.v
```diff
@@ -78,7 +78,7 @@ fn (mut g Gen) get_str_fn(typ ast.Type) string {
 	if sym.is_builtin() && !str_fn_name.starts_with('builtin__') {
 		str_fn_name = 'builtin__${str_fn_name}'
 	}
-	if sym.has_method_with_generic_parent('str') {
+	if sym.has_method_with_generic_parent('str') && !g.pref.new_generic_solver {
```
spytheman, line 81:
```suggestion
	if !g.pref.new_generic_solver && sym.has_method_with_generic_parent('str') {
```

vlib/v/generics/generics.v
```diff
@@ -351,6 +365,62 @@ pub fn (mut g Generics) styp(_t ast.Type) string {
...
+// incomplete implementation: TODO
+fn (mut g Generics) register_result(t ast.Type) string {
+	styp, _ := g.result_type_name(t)
+	return styp
+}
+
+// TODO: this really shouldn't be separate from typ
+// but I(emily) would rather have this generation
+// all unified in one place so that it doesn't break
+// if one location changes
```
spytheman, line 383:
```suggestion
```
you are not emily, and the comment is not really useful anymore imho

vlib/v/generics/generics.v
```diff
@@ -382,12 +452,15 @@ fn (mut g Generics) cc_type(typ ast.Type, is_prefix_struct bool) string {
 	return styp
 }
 
-pub fn (mut g Generics) method_concrete_name(old_name string, concrete_types []ast.Type, receiver_type ast.Type) string {
+pub fn (mut g Generics) method_concrete_name(old_name string, concrete_types []ast.Type, _receiver_type ast.Type) string {
```
spytheman, line 455: why use a `_name` here?
`original_receiver_type` may be clearer imho

vlib/v/gen/c/auto_str_methods.v
Eliyaan: I have put it like that to avoid checking for the new solver even if  the condition is not valid, as the new solver is not the default path for now. But if you find this clearer, I can change it

spytheman: Checking a boolean is faster than doing a fn call, and `&&` does not evaluate its right operand, if the first is false ->  it is a good habbit to put the cheap operation on the left side, even if it may not matter much in this case...

Eliyaan: Oh true I did not consider the cost of the fn call. Thank you

PR #26395, opened 2026-01-19
ci: add task to build V with prealloc on OpenBSD

lcheylus:
As with Linux, macOS and FreeBSD CI, add task in "CI OpenBSD" to build V with prealloc (`tcc` and `clang` compiler)

✅ Run of "CI OpenBSD" with build of V with prealloc https://github.com/lcheylus/v/actions/runs/21133699275


medvednikov:
Thank you so much for improving openbsd ci!

PR #26396, opened 2026-01-19
vcomplete: add git-fmt-hook command

lcheylus:
Add shell completion to find `v git-fmt-hook` command.

**Tests OK with Bash shell** on OpenBSD and Linux.


spytheman (approved):
Thank you @lcheylus .

PR #26397, opened 2026-01-19
misc: add micro editor syntax link for V

marc-dantas:
I appreciate this language a lot but I use [Micro](https://micro-editor.github.io/) as my main editor for a lot of my stuff.

So I decided to make my own syntax file for Micro. I did my best to search the docs and find everything that could be highlighted properly. Also searched in other syntax plugins/files for other editors and it seems like mine is proper.


spytheman:
Thank you for the contribution 🙇🏻 .

Micro already has builtin support for the V syntax, see: <https://github.com/zyedidia/micro/blob/master/runtime/syntax/v.yaml> .

You can submit the link to your project to https://github.com/vlang/awesome-v instead though.

PR #26399, opened 2026-01-19
gg: opti draw_rounded_rect_filled

Linklancien:
Another opti for ``draw_rounded_rect_filled``

There is only 2 cases, either there is no radius then you only need to draw a rectangle
Or there is a radius and you can juste draw the to rounded part and then the bottom rounded part with an order that create the inner rectangle automaticaly

Now it only use one ``sgl.begin_`` in both cases


spytheman:
It looks good to me, on macos (before/after renders are the same):
image

PR #26400, opened 2026-01-20
orm: add explicit JOIN support (INNER, LEFT, RIGHT, FULL OUTER)

### LINKED ISSUE #21635: Add explicit preload/join feature to ORM
### Describe the feature

Allow the user to explicitly load fkey rows, instead of the ORM doing it automatically.

In my opinion, child relationships should only be loaded manually, never automatically, as there could be millions of connections.

### Use Case

If we have a database structure like `Grandparent -> Parent -> Child`, when we select from `Grandparent`, it seems to automatically preload `Parents` associated with it.

```
sql db {
	select from Grandparent
}!
// Grandparent{name:'Anna', parents:[
//     Parent{name:'Claude', children:[]},
//     Parent{name:'Gloria', children:[]}
// }]
```

From what I can tell, it does try to load the children as well, but there is a logic error somewhere as mentioned in #21598.
That PR fixes the compile-time panic and causes the program to behave predictably, but I'm not sure if it's behaving correctly.

### Proposed Solution

The best solution *I can think of* is introducing a `load` / `preload` feature to the ORM, which would optionally accept a condition similar to SQL `JOIN`.

```
sql db {
	select from Grandparent
	preload Grandparent.parents p on p.name == 'Claude'
	preload c.children
}
// Grandparent{name:'Anna', parents:[
//     Parent{name:'Claude', children:[
//         Child{name:'Ralph'},
//         Child{name:'Michael'}
//     ]}
//     /* NOTE Gloria from above isn't loaded */
// }]
```

I don't particularly like this exact solution but I can't think of one at the moment that wouldn't stray too far from SQL behavior while comfortably fitting V.

### Other Information

_No response_

### Acknowledgements

### Version used

V 0.4.6 6d2c3a9

V full version: V 0.4.6 ed7f1ff.6d2c3a9
OS: macos, macOS, 14.5, 23F79
Processor: 11 cpus, 64bit, little endian, Apple M3 Pro

vexe mtime: 2024-06-03 06:55:49

VTMP: OK, value: /tmp/v_501

Git vroot status: weekly.2024.22-32-g6d2c3a9c
.git/config present: true

thirdparty/tcc status: thirdparty-macos-arm64 5c1d002f

watzon:
Implements explicit JOIN operations for V's ORM, addressing issue #21635.

New syntax:
  sql db {
      select from User
      join Department on User.department_id == Department.dept_id
      where active == true
  }!

Supported join types:
- join / inner join - INNER JOIN
- left join - LEFT JOIN
- right join - RIGHT JOIN
- full outer join - FULL OUTER JOIN

Changes:
- ast: Add JoinKind enum and JoinClause struct to SqlExpr
- parser: Parse JOIN clauses between FROM and WHERE
- checker: Validate joined tables and ON expressions
- orm: Add JoinConfig to SelectConfig, generate JOIN SQL
- gen/c: Generate C code for JOIN configurations

Closes #21635

### DIFF
--- a/vlib/orm/orm_join_test.v
+++ b/vlib/orm/orm_join_test.v
@@ -1,5 +1,5 @@
 // vtest retry: 3
-// vtest build: present_sqlite3? && !windows
+// vtest build: present_sqlite3? && !windows && !sanitize-memory-clang
 import db.sqlite
 
 // Department table for testing JOINs - using dept_id to avoid column name conflicts


vlib/orm/orm_join_test.v
spytheman, line 1: The test is not testing the new feature. Was it written by an AI?

spytheman: It also passes on master too :-| ...

watzon: `v fmt` :/

watzon: Didn't realize I had to fix `v fmt` so that it wouldn't remove the new JOIN clauses, so when I ran `v fmt` locally it stripped them and made the test pointless

medvednikov:
Looks amazing! Great job. A very important feature.

PR #26404, opened 2026-01-20
gg: add draw_text in gg/testdata

Linklancien:
This PR aims to add a new file in gg/testdata that draw text using ``draw_text`` and by variating some of the fields in the ``gg.TextCfg``

However, I am not sure if the ``bold`` and ``italic`` field have any effect or if I juste don't see it

### DIFF
--- a/vlib/gg/testdata/draw_text.vv
+++ b/vlib/gg/testdata/draw_text.vv
@@ -3,7 +3,7 @@ module main
 import gg
 
 gg.start(
-	width:        300
+	width:        640
 	height:       300
 	window_title: 'Circles'
 	frame_fn:     fn (mut ctx gg.Context) {


Linklancien:
> However, I am not sure if the ``bold`` and ``italic`` field have any effect or if I juste don't see it

I was told that ``bold`` and ``italic`` only works when you define a typeface before

vlib/gg/testdata/draw_text.vv
```diff
@@ -0,0 +1,45 @@
+module main
+
+import gg
+
+gg.start(
+	width:        300
```
spytheman:
```suggestion
	width:        640
```
this will make the window large enough, so that all letters are visible by default, without having to resize it.

Linklancien:
Do you think there is a need of something else ?
@spytheman

spytheman:
No, it is good to merge.
Thank you @Linklancien 🙇🏻 .

PR #26418, opened 2026-01-22
math: rewrite hypot(), update tests

tankf33der:
Reworked my tests, added more candidates to the edge cases list and `hypot()` got caught in the net. 
Didn't find anything better than rewriting the function from [golang](https://github.com/golang/go/blob/master/src/math/hypot.go).
Added one test that panics on `master`.

The patch passes all my tests against `mpfr`.

### DIFF
--- a/vlib/math/hypot.v
+++ b/vlib/math/hypot.v
@@ -1,25 +1,27 @@
 module math
 
-// hypot returns the hypotenuse of the triangle give two sides
+// hypot returns the hypotenuse of the triangle give two sides.
+//
+// special cases are:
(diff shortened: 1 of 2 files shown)


vlib/math/hypot.v
```diff
@@ -1,25 +1,27 @@
 module math
 
-// hypot returns the hypotenuse of the triangle give two sides
+// hypot returns the hypotenuse of the triangle give two sides.
```
spytheman, line 3:
```suggestion
// hypot returns the hypotenuse of a triangle given two of its sides.
```

PR #26424, opened 2026-01-23
Optimize/http2 http3 performance 20260123

jupilhwang:
http/2, http/3 - Client / Server Lib for vlang

HTTP/2, HTTP/3 Performance
- Frame encoding: 0.34 μs (87% faster)
- Throughput: 3,051 MB/s (20,965% faster)
- HPACK encoding: 1.64 μs (93% faster)
- Headers/second: 609,347 (2,244% faster)

HTTP/3
- Using OpenSSL, ngtcp2


medvednikov:
Wow, that'd be great to have.

Will you try again later?

Issue #26437, opened 2026-01-25
x.json2 does not respect `@[json: name]`attribute for enums


gechandesu:
### Describe the bug

The subject. See example below.

### Reproduction Steps

```v
import x.json2

enum Foo {
        bar  @[json: '_BAR']
        baz  @[json: '_BAZ']
}

struct FooBar {
        barbaz Foo
}

fn main() {
        f := FooBar{Foo.bar}
        e2 := json2.encode[FooBar](f)
        dump(e2)
        d2 := json2.decode[FooBar](e2)!
        dump(d2)

        assert d2 == f
}
// [enumjson2.v:15] e2: {"barbaz":"bar"}
// [enumjson2.v:17] d2: FooBar{
//     barbaz: bar
// }
```

### Expected Behavior

Output JSON:
```json
{"barbaz":"_BAR"}
```

### Current Behavior

Output JSON:
```json
{"barbaz":"bar"}
```

### Additional Information/Context

cJSON works as expected:

```v
import json

enum Foo {
	bar  @[json: '_BAR']
	baz  @[json: '_BAZ']
}

struct FooBar {
	barbaz Foo
}

fn main() {
	f := FooBar{Foo.bar}
	e := json.encode(f)
	dump(e)
	d := json.decode(FooBar, e)!
	dump(d)
	assert d == f
}
// [enumjson.v:15] e: {"barbaz":"_BAR"}
// [enumjson.v:17] d: FooBar{
//     barbaz: bar
// }
```

### V version

V 0.5.0 c3b924c

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.0 19e79c4b468d4e82484fe8ace5b6380cf7949245.c3b924c
|:-------------------|:-------------------
|OS                  |linux, Linux version 6.18.6-arch1-1 (linux@archlinux) (gcc (GCC) 15.2.1 20260103, GNU ld (GNU Binutils) 2.45.1) #1 SMP PREEMPT_DYNAMIC Sun, 18 Jan 2026 00:34:07 +0000
|Processor           |16 cpus, 64bit, little endian, 12th Gen Intel(R) Core(TM) i5-1240P
|Memory              |0.96GB/15.34GB
|                    |
|V executable        |/home/ge/.vlang/v
|V last modified time|2026-01-25 16:23:07
|                    |
|V home dir          |OK, value: /home/ge/.vlang
|VMODULES            |OK, value: /home/ge/.vmodules
[...]

PR #26441, opened 2026-01-25
x.json2: add `@[json: name]` attr support for enum values

gechandesu:
Fix https://github.com/vlang/v/issues/26437

This feature is already supported in `json` module, now in `x.json2` too.

### DIFF
--- a/vlib/x/json2/decode.v
+++ b/vlib/x/json2/decode.v
@@ -854,7 +854,7 @@ fn (mut decoder Decoder) decode_enum[T](mut val T) ! {
 
 		$for value in T.values {
 			for attr in value.attrs {
-				if attr.all_after('json: ') == result {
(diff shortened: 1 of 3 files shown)


gechandesu:
I just noticed that I made a bug in encode.v, which caused attributes that didn't start with `json:` to be used as values ​​as is. No change was required in decode.v. Upd: fixed

spytheman:
> I just noticed that I made a bug in encode.v, which caused attributes that didn't start with `json:` to be used as values ​​as is. No change was required in decode.v. Upd: fixed

Can you please also modify the test, so that it has another tag on a field, to make sure that it will not regress in the future (I also could not spot the bug initially, so imho the regression test can save us time in detecting similar situations)?

gechandesu:
Thanks, I added tests and also discovered that the error was still in decode.v... Now it's fixed.

PR #26443, opened 2026-01-25
cgen: fix or-block with fixed array constants with GCC 15.2

### LINKED ISSUE #26442: cgen: invalid C cast to array type when or-block returns fixed array constant
## Describe the bug

When an or-block returns a fixed array constant or variable, the C code generator produces invalid C code that attempts to cast to an array type. This causes compilation failure when using GCC 15.2 as the C backend.

## Reproduction Steps

```v
const default_arr = [f32(1.0), 2.0, 3.0, 4.0]!

fn get_arr(key string) ?[4]f32 {
	if key == 'valid' {
		return [f32(0.1), 0.2, 0.3, 0.4]!
	}
	return none
}

fn main() {
	result := get_arr('invalid') or { default_arr }
	println(result)
}
```

Compile with GCC:
```
v -cc gcc test.v
```

## Expected Behavior

The code should compile successfully with all supported C compilers.

## Current Behavior

GCC compilation fails with:

```
error: cast specifies array type
 8489 |                 memcpy(_t1.data, (Array_fixed_f32_4)_const_main__default_arr, sizeof(Array_fixed_f32_4));
      |                                  ^
```

The generated C code incorrectly casts a constant/variable to an array type, which is invalid C. The cast `(Array_fixed_f32_4)` is only valid for compound literals (brace-enclosed initializers), not for existing variables or constants.

Note: TCC accepts this cast, so the bug only manifests when using GCC. Per C99 6.5.4, cast operators shall not specify array types.

## Possible Solution

In `gen_or_block_stmts()`, only apply the compound literal cast for expressions that generate brace-enclosed initializers (`ArrayInit`, `StructInit`, or `CastExpr` wrapping `ArrayInit`). For constants and variables, omit the cast and pass them directly to `memcpy`.

## Additional Information/Context

The fix needs to distinguish between:
- `or { [1, 2, 3]! }` 
    - generates `{1, 2, 3}` which needs cast to become compound literal
- `or { some_const }`
    - generates `_const_some_const` which should not be cast

## Environment details (OS name and version, etc.)

|V full version      |V 0.5.0 c3b924c7bd292340b2fc9a2dc0ea91a3fedbdcee.0b0bc20
|OS                  |linux, "Arch Linux" (VM)
|Processor           |12 cpus, 64bit, little endian, AMD Ryzen Threadripper 7960X 24-Cores
|Memory              |5.1GB/47GB
|                    |
|V executable        |/home/jesusa/code/v/v/v
|V last modified time|2026-01-25 19:00:59
|                    |
|V home dir          |OK, value: /home/jesusa/code/v/v
|VMODULES            |OK, value: /home/jesusa/.vmodules
|Current working dir |OK, value: /home/jesusa/code/v/vrasta
|                    |
|V git status        |weekly.2026.03-49-g0b0bc205 (26 commit(s) behind V master)
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20260103
|gcc version         |gcc (GCC) 15.2.1 20260103
|clang version       |clang version 21.1.6
|emcc version        |N/A

> [!NOTE]
>

demizer:
Fix invalid C cast when or-block returns a fixed array constant/variable with GCC 15.2.                                                                                                                                                                   
                                                                                                                                                                                                                                                
  Fixes #26442                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                
  Changes                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                
  - Only use compound literal cast for brace-enclosed initializers (ArrayInit, StructInit), not for constants/variables                                                                                                                         
  - Add test case for or-block with fixed array constant                                                                                                                                                                                        
                                                                                                                                                                                                                                                
  Test plan                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                
  - v -cc gcc vlib/v/tests/options/option_fixed_arr_const_test.v passes                                                                                                                                                                         
  - Verified test fails on master, passes with fix

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -7404,7 +7404,19 @@ fn (mut g Gen) gen_or_block_stmts(cvar_name string, cast_typ string, stmts []ast
 						g.write('${cvar_name} = ')
 					}
 					if is_array_fixed {
-						g.write('memcpy(${cvar_name}${tmp_op}data, (${cast_typ})')
+						// For fixed arrays, we need memcpy. Use compound literal cast only for
+						// expressions that generate brace-enclosed initializers (which need
+						// to become compound literals), not for constants/variables where
+						// casting to array type is invalid C.
+						is_array_init := expr_stmt.expr is ast.ArrayInit
+							|| expr_stmt.expr is ast.StructInit
+							|| (expr_stmt.expr is ast.CastExpr
+							&& (expr_stmt.expr as ast.CastExpr).expr is ast.ArrayInit)
+						if is_array_init {
+							g.write('memcpy(${cvar_name}${tmp_op}data, (${cast_typ})')
+						} else {
+							g.write('memcpy(${cvar_name}${tmp_op}data, ')
+						}
 					}
 					// return expr or { fn_returns_option() }
 					if is_option && g.inside_return && expr_stmt.expr is ast.CallExpr
--- a/vlib/v/tests/options/option_fixed_arr_const_test.v
+++ b/vlib/v/tests/options/option_fixed_arr_const_test.v
@@ -1,0 +1,20 @@
+// Test for or-block returning a fixed array constant.
+// Regression test for issue where cgen generated invalid C cast to array type.
+const default_arr = [f32(1.0), 2.0, 3.0, 4.0]!
+
+fn get_arr(key string) ?[4]f32 {
+	if key == 'valid' {
+		return [f32(0.1), 0.2, 0.3, 0.4]!
+	}
+	return none
+}
+
+fn test_or_block_with_fixed_array_const() {
+	result := get_arr('invalid') or { default_arr }
+	assert result == default_arr
+}
+
+fn test_or_block_with_valid_key() {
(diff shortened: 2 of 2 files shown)


vlib/v/gen/c/cgen.v
```diff
@@ -7396,8 +7396,20 @@ fn (mut g Gen) gen_or_block_stmts(cvar_name string, cast_typ string, stmts []ast
...
 						g.write('${cvar_name} = ')
 					}
+					// For fixed arrays, we need memcpy. Use compound literal cast only for
+					// expressions that generate brace-enclosed initializers (which need
+					// to become compound literals), not for constants/variables where
+					// casting to array type is invalid C.
+					is_array_init := expr_stmt.expr is ast.ArrayInit
+						|| expr_stmt.expr is ast.StructInit
+						|| (expr_stmt.expr is ast.CastExpr
+						&& (expr_stmt.expr as ast.CastExpr).expr is ast.ArrayInit)
```
spytheman: please move this inside the next if (i.e. closer to the place where it is used)

Issue #26433, opened 2026-01-25
Accessible private field in generic struct


iperov:
### Describe the bug

Can access private field in generic struct

### Reproduction Steps

```V
module cont

pub struct Foo[T] {
    i int
}

pub struct Bar {
    i int
}
```

```V
fn main() 
{  
   q := cont.Foo[int]{i:3} // no error
    
   w := cont.Bar{i:3} // error: cannot access private field `i` on `cont.Bar`    
}
```

### Expected Behavior

expected error

### Current Behavior

no error

### V version

V 0.5.0 37577eb

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.0 37577eb4aa94eb17b2f54db7819142efd4454a9a
|:-------------------|:-------------------
|OS                  |windows, Microsoft Windows 10 Pro 19045 64-bit
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz  
|Memory              |45GB/63.94GB
|                    |
|V executable        |C:\PORTABLE\DEVELOP\V\VProject\_internal\v\v.exe
|V last modified time|2026-01-21 15:55:10
|                    |
|V home dir          |OK, value: C:\PORTABLE\DEVELOP\V\VProject\_internal\v
|VMODULES            |OK, value: C:\PORTABLE\DEVELOP\V\VProject\_internal\_z\u\.vmodules
|VTMP                |OK, value: C:\PORTABLE\DEVELOP\V\VProject\_internal\_z\t\v_0
|Current working dir |OK, value: C:\PORTABLE\DEVELOP\V\VProject\master
|                    |
|Git version         |git version 2.50.1.windows.1
|V git status        |N/A
|.git/config present |false
|                    |
|cc version          |N/A
|gcc version         |gcc (MinGW-W64 x86_64-ucrt-posix-seh, built by Brecht Sanders, r4) 15.2.0
|clang version       |N/A
|msvc version        |N/A
|tcc version         |tcc version 0.9.27 (x86_64 Windows)
|tcc git status      |N/A
|emcc version        |N/A
|glibc version       |N/A

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #26450, opened 2026-01-26
parser, checker: check generic struct fields and init 

Delta456:
Fixes https://github.com/vlang/v/issues/26433 https://github.com/vlang/v/issues/26436

### DIFF
--- a/vlib/v/parser/parse_type.v
+++ b/vlib/v/parser/parse_type.v
@@ -1026,12 +1026,15 @@ fn (mut p Parser) parse_generic_inst_type(name string, name_pos token.Pos) ast.T
 				}
 			}
 		}
+		// mod.Foo[int] -> mod
+		// mod.submod.Foo[int] -> mod.submod
+		mod := name.all_before_last('.')
 
 		idx := p.table.register_sym(ast.TypeSymbol{
 			kind:  .generic_inst
 			name:  bs_name
 			cname: util.no_dots(bs_cname)
-			mod:   name.all_before_last('.')
+			mod:   mod
 			info:  ast.GenericInst{
 				parent_idx:     parent_idx
 				concrete_types: concrete_types
--- a/vlib/builtin/js/promise.js.v
+++ b/vlib/builtin/js/promise.js.v
@@ -14,7 +14,7 @@ pub fn JS.Promise.race(JS.Array) JS.Promise
 
 // Promise represents the eventual completion (or failure) of an asynchronous operation and its resulting value.
 pub struct Promise[T] {
-mut:
+pub mut:
 	promise JS.Promise @[noinit]
 }
 


Delta456:
CI Linux / tcc-linux (pull_request) failure is not related to my PR and fails in the main repo.

vlib/v/parser/parse_type.v
```diff
@@ -1031,7 +1031,7 @@ fn (mut p Parser) parse_generic_inst_type(name string, name_pos token.Pos) ast.T
 			kind:  .generic_inst
 			name:  bs_name
 			cname: util.no_dots(bs_cname)
-			mod:   p.mod
```
Delta456, line 1034: `p.mod` had always been `main`, so I had to manually extract the module name.

vlib/v/checker/struct.v
```diff
@@ -1078,7 +1077,8 @@ fn (mut c Checker) check_uninitialized_struct_fields_and_embeds(node ast.StructI
 							} else {
 								parts.last()
 							}
-							if !c.inside_unsafe {
+							if !c.inside_unsafe && !(c.pref.backend.is_js()
```
spytheman: please call c.pref.backend.is_js() just once before the loop, or even better, cache it on field of the Checker struct; it will be called *a lot* if it is inside this loop

vlib/v/parser/fn.v
```diff
@@ -842,7 +842,7 @@ run them via `v file.v` instead',
 			is_non_local = elem_type_sym.mod.len > 0 && elem_type_sym.mod != p.mod
 				&& elem_type_sym.language == .v
 		}
-		if is_non_local {
+		if is_non_local && !type_sym.mod.starts_with('Promise') {
```
spytheman: Why special case Promise here, instead of marking the offending struct fields just once with pub: in their source file?

vlib/v/parser/parse_type.v
spytheman: hm, that seems weird - afaik p.mod changes once per file, by the `module xyz` statements, when a .v file has one

spytheman:
> CI Linux / tcc-linux (pull_request) failure is not related to my PR and fails in the main repo.

Yes, it is my fault, sorry.
It should be fine now on master.

vlib/v/parser/parse_type.v
Delta456: Yes, it does set, but when calling this function, it is always `main`

vlib/v/checker/struct.v
Delta456: That's a good idea

vlib/v/parser/fn.v
Delta456: Here `p.mod` is `builtin`, which doesn't match in the source file. This is a special case.

vlib/v/checker/struct.v
spytheman: set a field named `is_js_backend` just once inside Checker's new_checker method, then here it can be just `if !c.inside_unsafe && !(c.is_js_backend`

vlib/v/parser/fn.v
spytheman: Here, what I mean, is that there should be at least a check for `p.file_backend_mode == .js` too, to avoid checking the `!type_sym.mod.starts_with('Promise')` for all v code; it is something that is relevant only inside .js.v files.

spytheman: i.e. something like this:
```suggestion
		if is_non_local && !(p.file_backend_mode == .js && type_sym.mod.starts_with('Promise')) {
```

spytheman: alternatively also add: `pub:` to the `pub struct Promise[T] {` fields, but this afaik would not affect the error

vlib/v/checker/struct.v
Delta456: done

PR #26451, opened 2026-01-26
v.builder: add -stats for top level statements and non-vlib top level statements

spytheman:
Some more stats for the examples and V itself (on macos):
image


spytheman:
Screenshot showing the main module top level statements too:
image

PR #26455, opened 2026-01-27
x.json2: rename Any.arr() to Any.as_array() for consistence with as_map(), as_map_of_string(), deprecate Any.arr()

gechandesu:
A bit more letters, but a consistent and obvious function name.

### DIFF
--- a/vlib/x/json2/README.md
+++ b/vlib/x/json2/README.md
@@ -153,7 +153,7 @@ fn (mut p Person) from_json(f json2.Any) {
 `x.json2` provides methods for turning `Any` types into usable types.
 The following list shows the possible outputs when casting a value to an incompatible type.
 
-1. Casting non-array values as array (`as_array()`) will return an array with the value as the content.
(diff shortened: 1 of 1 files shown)


vlib/x/json2/README.md
```diff
@@ -153,7 +153,7 @@ fn (mut p Person) from_json(f json2.Any) {
 `x.json2` provides methods for turning `Any` types into usable types.
 The following list shows the possible outputs when casting a value to an incompatible type.
 
-1. Casting non-array values as array (`arr()`) will return an array with the value as the content.
+1. Casting non-array values as array (`as_array()`) will return an array with the value as the content.
```
spytheman:
```suggestion
1. Casting non-array values with `as_array()` will return an array with the value as the content.
```

spytheman: (one of the CI jobs enforces a line length limit to the readme markdown lines)

Issue #26449, opened 2026-01-26
Void Linux with musl error cc: tcc: error: undefined symbol '__atomic_thread_fence'


robotamer:
### Describe the bug

I have upgraded V after a few months of inactivity.
Now I am getting this error
```
      cc: tcc: error: undefined symbol '__atomic_thread_fence'
```
I can compile simple code such as hello world etc.
But anything more complex such as vpm, ved ... gives me this error.
```
musl-dev is installed !!!

    [*] musl-devel-1.2.5_2 Musl C library - development files
```
I am not sure if this was working before the upgrade. I was only learning V, hence doing simple stuff.
I am now ready to deep dive in to V, but need this to work first.

### Reproduction Steps
```
cd ved
v .
```
### Expected Behavior

should_compile: true

### Current Behavior
```
$ v -v outdated                                                                                  
launch_tool vexe        : /home/tamer/code/v/v/v
launch_tool vroot       : /home/tamer/code/v/v
launch_tool tool_source : /home/tamer/code/v/v/cmd/tools/vpm
launch_tool tool_exe    : /home/tamer/code/v/v/cmd/tools/vpm/vpm
launch_tool tool_args   : '-v' 'outdated'
launch_tool should_compile: true
Compiling vpm with: "'/home/tamer/code/v/v/v' '/home/tamer/code/v/v/cmd/tools/vpm'"
cannot compile `/home/tamer/code/v/v/cmd/tools/vpm`: 1
================== C compilation error (from tcc): ==============
cc: tcc: error: undefined symbol '__atomic_thread_fence'
=================================================================
Try passing `-g` when compiling, to see a .v file:line information, that correlates more with the C error.
(Alternatively, pass `-show-c-output`, to print the full C error message).
builder error: 
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```      

### V version

V 0.5.0 f479e98

### Environment details (OS name and version, etc.)
```
    "os": {
      "name": "Void Linux",
      "vendor": "void"
    },
    "kernel": {
      "release": "6.12.63_1",
      "version": "#1 SMP PREEMPT_DYNAMIC Thu Dec 18 18:36:00 UTC 2025",
      "architecture": "x86_64"
    },
  "bios": {
      "vendor": "AMI",
      "version": "F.64",
[...]

PR #26456, opened 2026-01-27
Void Linux with musl error cc: tcc: error: undefined symbol '__atomic_thread_fence' #26449

robotamer:
Detect musl on any nix system, not just alpine

https://github.com/vlang/v/issues/26449


JalonSolov:
Apologies.  The new code should be
```v
fn detect_musl(mut res pref.Preferences) {
        res.is_glibc = true
        res.is_musl = false
        musl := os.execute('ldd --version').output.contains('musl')
        if musl {
                res.is_musl = true
                res.is_glibc = false
        }
}
```

PR #26469, opened 2026-01-30
Change the port type of PG from int to u32

Jengro777:

### DIFF
--- a/vlib/db/pg/pg.c.v
+++ b/vlib/db/pg/pg.c.v
@@ -91,7 +91,7 @@ pub:
 pub struct Config {
 pub:
 	host     string = 'localhost'
-	port     int    = 5432
+	port     u32    = 5432
 	user     string
 	password string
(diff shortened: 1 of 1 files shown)


JalonSolov:
Should it be `u32` or `i32`?  `int` is currently `i32`.

Jengro777:
> Should it be `u32` or `i32`?  `int` is currently `i32`.

This should be consistent with MySQL.

Jengro777:
The port number should be a positive integer. Using `u32` can help prevent issues earlier in the process.

spytheman:
Imho that is a breaking change (as evidenced by the CI), for no good reason.

@medvednikov do you want it?

medvednikov:
No, postgres integer types are all signed.

medvednikov:
And also for the same reason array.len is an int. Discussed several times.

PR #26477, opened 2026-01-31
breaking,x.ttf: extract vlib/x/ttf/render_sokol_cpu.v in a separate module x.ttf.render_sokol, to enable headless programs, that do not use gg, to not depend on OpenGL, DX12, Metal etc, through sokol

spytheman:


spytheman:
(rebased over latest master, to pass the CI, after `v fmt -w vlib/v2/pref/` was done there)

spytheman:

this PR is breaking `x.ttf` but I think the change is worth it in the long term, since it will allow the module to be used in more situations.

spytheman:
Alex told me that it looks good to him in a DM. Merging.

PR #26478, opened 2026-01-31
fasthttp: multi threading fix reads in the socket in linux

esquerbatua:
**fasthttp:**
- Allow to the user to select the family and max_request_buffer_size of the connection.
- Fix read all of the data in the socket in linux.

**veb:**
- Allows to run veb in multithreading with the flag: " -d new_veb"

### DIFF
--- a/vlib/fasthttp/fasthttp_linux.v
+++ b/vlib/fasthttp/fasthttp_linux.v
@@ -94,7 +94,7 @@ fn close_socket(fd int) bool {
 	return true
 }
 
-fn create_server_socket(server &Server) int {
+fn create_server_socket(server Server) int {
 	// Create a socket with non-blocking mode
 	server_fd := C.socket(server.family, net.SocketType.tcp, 0)
 	if server_fd < 0 {
@@ -229,7 +229,7 @@ fn process_events(mut server Server, epoll_fd int, listen_fd int) {
 			if events[i].events & u32(C.EPOLLIN) != 0 {
 				// Read all available data from the socket
 				mut total_bytes_read := 0
-				mut all_data := []u8{cap: server.max_request_buffer_size}
+				mut readed_request_buffer := []u8{len: server.max_request_buffer_size, cap: server.max_request_buffer_size}
 
 				for {
 					bytes_read := C.recv(client_fd, unsafe { &request_buffer[0] }, server.max_request_buffer_size - 1,
@@ -253,16 +253,16 @@ fn process_events(mut server Server, epoll_fd int, listen_fd int) {
 						break
 					}
 					unsafe {
-						all_data.push_many(&request_buffer[0], bytes_read)
+						readed_request_buffer.push_many(&request_buffer[0], bytes_read)
 					}
 					total_bytes_read += bytes_read
 
 					// Check if we've received the complete HTTP request (look for \r\n\r\n)
 					if total_bytes_read >= 4 {
-						if all_data[total_bytes_read - 4] == `\r`
-							&& all_data[total_bytes_read - 3] == `\n`
-							&& all_data[total_bytes_read - 2] == `\r`
-							&& all_data[total_bytes_read - 1] == `\n` {
+						if readed_request_buffer[total_bytes_read - 4] == `\r`
+							&& readed_request_buffer[total_bytes_read - 3] == `\n`
+							&& readed_request_buffer[total_bytes_read - 2] == `\r`
+							&& readed_request_buffer[total_bytes_read - 1] == `\n` {
 							break
 						}
 					}
@@ -276,7 +276,7 @@ fn process_events(mut server Server, epoll_fd int, listen_fd int) {
 						handle_client_closure(epoll_fd, client_fd)
 						continue
 					}
-					mut decoded_http_request := decode_http_request(all_data) or {
+					mut decoded_http_request := decode_http_request(readed_request_buffer) or {
 						eprintln('Error decoding request ${err}')
 						C.send(client_fd, tiny_bad_request_response.data, tiny_bad_request_response.len,
 							C.MSG_NOSIGNAL)


spytheman:
@esquerbatua try `v fmt -w vlib/fasthttp` to fix the CI

spytheman:
also `v git-fmt-hook install` can make it so that each commit also runs vfmt automatically
(has to be run while your working folder is inside the v repo)

esquerbatua:
Sorry I formatted the veb folder instead fasthttp, it's formatted now

enghitalo:
I recommend not changing the logic behind closing the client connection for now.
Or you could do it in a separate PR, so you'd have to handle things like the HTTP version and request header to make it more appropriate. In fact, the perfect solution for me would be for `veb` to have the power to close the connection whenever it wants, but I haven't found a clever way to do this without creating a bit of a mess. Although, regardless of how it's done, it would still be much better than dealing with request interpretation within FastHTTP

vlib/fasthttp/fasthttp_linux.v
```diff
@@ -1,6 +1,8 @@
 module fasthttp
 
 import net
+import net.http
+import io
```
enghitalo: what is `io` used for?

vlib/fasthttp/fasthttp_linux.v
```diff
@@ -225,21 +229,56 @@ fn process_events(mut server Server, epoll_fd int, listen_fd int) {
 				continue
 			}
 			if events[i].events & u32(C.EPOLLIN) != 0 {
-				bytes_read := C.recv(client_fd, unsafe { &request_buffer[0] }, server.max_request_buffer_size - 1,
-					0)
-				if bytes_read > 0 {
+				// Read all available data from the socket
+				mut total_bytes_read := 0
+				mut all_data := []u8{cap: server.max_request_buffer_size}
```
enghitalo: I think it should be a fixed array with a error (HTTP response) if it exceed `max_request_buffer_size` limit and renaming `all_data` to `readed_request_buffer`.

esquerbatua:
If I do not close the connection the client keeps waiting until timeout, tested.

enghitalo:
> If I do not close the connection the client keeps waiting until timeout, tested.

This is the default expected behavior for http1.1

enghitalo:
@claude How connection should bahave in which http version? Show me the rfc reference.

esquerbatua:
• HTTP/0.9 (very old, single-request): one TCP connection, one request, server returns raw response then closes. (No formal modern RFC; historical notes only.)  • No persistent connections, no headers, no request/response framing.

• HTTP/1.0 (RFC 1945): by default the server closes the TCP connection after the response; clients and servers may use the Connection: keep-alive extension to request persistence, but it is optional and not reliably supported by all implementations.  • RFC: RFC 1945 (HTTP/1.0).

• HTTP/1.1 (message framing, persistent by default): connections are persistent by default (multiple requests/responses can be sent sequentially on the same TCP connection). Pipelining (sending multiple requests without waiting for responses) is defined but problematic in practice; implementations more commonly use sequential requests or upgrade to multiplexing (HTTP/2). The Connection header and hop-by-hop semantics are defined here.  • Key RFCs:    • RFC 7230 (HTTP/1.1 message syntax and routing) — see section on connection management and persistent connections (RFC 7230 §6).
    • RFC 7231 (semantics) and the rest of the RFC723x family for HTTP/1.1.


• HTTP/2 (binary framing, multiplexed streams over one TCP connection): a single TCP connection carries many independent, interleaved request/response streams concurrently (no head-of-line blocking at the HTTP message level, though TCP-level HOL remains). Uses HPACK for header compression and has explicit stream and connection lifecycle rules.  • Key RFCs:    • RFC 7540 (Hypertext Transfer Protocol Version 2) — see multiplexing, streams, connection flow control.
    • RFC 7541 (HPACK — header compression).


• HTTP/3 (HTTP semantics over QUIC): moves from TCP to QUIC (UDP-based transport), giving multiplexing without TCP head-of-line blocking and integrated stream- and connection-level flow/congestion control; uses QPACK for header compression with HTTP/3-specific considerations.  • Key RFCs:    • RFC 9114 (HTTP/3)
    • RFC 9000 (QUIC: the transport)
    • RFC 9204 (QPACK — header compression for HTTP/3)


Pointers / exact references to read

• HTTP/1.0: RFC 1945 — https://datatracker.ietf.org/doc/html/rfc1945
• HTTP/1.1 message framing & persistent connections: RFC 7230 — https://datatracker.ietf.org/doc/html/rfc7230 (see §6 on connection management/persistence)
• HTTP/1.1 semantics: RFC 7231 — https://datatracker.ietf.org/doc/html/rfc7231
• HTTP/2: RFC 7540 — https://datatracker.ietf.org/doc/html/rfc7540 ; HPACK: RFC 7541 — https://datatracker.ietf.org/doc/html/rfc7541
• QUIC (transport used by HTTP/3): RFC 9000 — https://datatracker.ietf.org/doc/html/rfc9000
• HTTP/3: RFC 9114 — https://datatracker.ietf.org/doc/html/rfc9114 ; QPACK: RFC 9204 — https://datatracker.ietf.org/doc/html/rfc9204

esquerbatua:
The point it's like if I use vweb or veb without flags, the service works as expected, if I use fast http, the connection keeps open

spytheman:
Can there be a separate thread, that loops over all current opened connections, every say 10 seconds, and closes them, if there were no activity on them for say 1 minute?

enghitalo:
@esquerbatua Can you share the code which you're facing issues?

esquerbatua:
> @esquerbatua Can you share the code which you're facing issues?

The simplest veb server should give timeout in all of the request. I didn't have time to test it changing the handing of the connection close.

```v
module main

import veb { Result }

pub struct Context {
	veb.Context
}

struct App {
}

fn main() {
	mut app := App{}

	veb.run_at[App, Context](
		mut app,
		host: '0.0.0.0',
		port: 8080,
		family: .ip,
	) or { panic(err) }
}

@['/']
pub fn (app &App) index(mut ctx Context) Result {
	return ctx.ok('')
}
```

esquerbatua:
> Can there be a separate thread, that loops over all current opened connections, every say 10 seconds, and closes them, if there were no activity on them for say 1 minute?

Sorry if I'm wrong:

If we do that, always we are going to give extra time to all of the request, making veb or fasthttp so slow, make them useless with big load time and extra handle of the connection.
We should check if the connection have X request, process it in the same order that arrived. But that logic needs to be implemented.
Changing only the way that we close the connection we make fasthttp unusable.

enghitalo:
Why not just close the connection at client side, as usual or handle `Connection: close` header at server `veb` side?

enghitalo:
> The point it's like if I use vweb or veb without flags, the service works as expected, if I use fast http, the connection keeps open

I'm sorry, but I'm still struggling to understand how the connection remaining active after sending the reply could be a problem

enghitalo:
@esquerbatua 
> If I do not close the connection the client keeps waiting until timeout, tested.

The client doesn't have to wait for the connection timeout; it simply receives all the responses, check `content-lenght` and then closes automatically. Unless there's a problem with your client or the response you're sending.

esquerbatua:
I reverted the changes in the clousure as they are, shouldn't modify the old behaviour @enghitalo

enghitalo:
Thanks

vlib/fasthttp/fasthttp_linux.v
```diff
@@ -92,9 +94,9 @@ fn close_socket(fd int) bool {
 	return true
 }
 
-fn create_server_socket(port int) int {
+fn create_server_socket(server &Server) int {
```
enghitalo: Why is it a pointer here?

spytheman (commented):
Is it all ready now?

esquerbatua:
> Is it all ready now?

yes

vlib/fasthttp/fasthttp_linux.v
esquerbatua: this it's outdated, removed.

vlib/fasthttp/fasthttp_linux.v
esquerbatua: removed pointer

vlib/fasthttp/fasthttp_linux.v
```diff
@@ -229,7 +229,7 @@ fn process_events(mut server Server, epoll_fd int, listen_fd int) {
 			if events[i].events & u32(C.EPOLLIN) != 0 {
 				// Read all available data from the socket
 				mut total_bytes_read := 0
-				mut all_data := []u8{cap: server.max_request_buffer_size}
+				mut readed_request_buffer := []u8{len: server.max_request_buffer_size, cap: server.max_request_buffer_size}
```
spytheman, line 232: `[]u8{len: server.max_request_buffer_size}`

esquerbatua: this it's not changed

Issue #26382, opened 2026-01-17
'$str' results in string interpolation/compilation error


Delta456:
### Describe the bug

`$str` still results in string interpolation when it has been with `${str}`. 

### Reproduction Steps

The following code results an error:

```v
text :="a$b"
println(text)
```

output:

```
code.v:1:11: error: undefined ident: `b`
    1 | text :="a$b"
      |           ^
    2 | println(text)
code.v:1:11: error: expression does not return a value
    1 | text :="a$b"
      |           ^
    2 | println(text)
Exited with error status 1
```

this results to string interpolation when it shouldn't

```v
b :=10
text :="a$b"
println(text)
```

output:
```
a10
```

### Expected Behavior

This ultimately makes `$` to be used in strings impossible

### Current Behavior

Error and unexpected behavior

### Possible Solution

`$` should be useful in strings

### V version

latest

### Environment details (OS name and version, etc.)

Latest
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

revosw:
Just curious, is there a stance on removing the `$var` syntax entirely in favor of only `${var}`?

JalonSolov:
Yes.  The stance is that `$var` was deprecated several years ago.  Only `${var}` should work now.  `$var` should just be the string `$var`.

So in the example above, the output should have been `a$b`, not `a10`.

spytheman:
Yes, we should remove the special support for the old syntax, enough time has passed already.

esquerbatua:
temporally you can use:
```v
text :="a\$b"
println(text)
```
output: `a$b`

PR #26487, opened 2026-02-01
bench: refactor old string interpolation to new string interpolation for bench_euclid.v

revosw:
Part of https://github.com/vlang/v/issues/26382

The regex I have used to search for strings is
```
(?<!\\)\$(?!if\b|for\b|else\b|match\b|pointer\b|voidptr\b|tmpl\b|array\b|array_dynamic\b|array_fixed\b|function\b|shared\b|interface\b|enum\b|int\b|string\b|struct\b|map\b|option\b|float\b|sumtype\b|alias\b|pointer\b|voidptr\b)(\w+\.?\w{0,})
```
(Dollar sign _not_ preceded by `\`, then _not_ succeeded by comptime keywords. Select any word character after the dollar sign)

### DIFF
--- a/vlib/v/tests/bench/math_big_gcd/bench_euclid.v
+++ b/vlib/v/tests/bench/math_big_gcd/bench_euclid.v
@@ -93,7 +93,7 @@ fn main() {
 		bench_euclid_vs_binary(prime_cfg, false, predicate_fn, mut clocks)
 
 		// just-to-be-sure, but makes no difference in this test.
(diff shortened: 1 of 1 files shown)


spytheman:
Why do separate PRs for the same change?
Reviewing does not get easier from it, you do more work, and the CI gets clogged ...

revosw:
Okay, I'll make a single pr with all the changes

spytheman:
> Okay, I'll make a single pr with all the changes

Thank you 🙇🏻 .

Issue #26491, opened 2026-02-01
checker: type narrowing in none removes mut


Le0Developer:
<details>
<summary>V version: V 0.5.0 75fc8be, press to see full `v doctor` output</summary>

|V full version      |V 0.5.0 75fc8bed1876a3dccbdea5005125473ac9fbf8cb
|:-------------------|:-------------------
|OS                  |macos, macOS, 26.2, 25C56
|Processor           |10 cpus, 64bit, little endian, Apple M4
|Memory              |0.23GB/16GB
|                    |
|V executable        |/Users/leodev/p/v/v/v
|V last modified time|2026-01-31 10:45:14
|                    |
|V home dir          |OK, value: /Users/leodev/p/v/v
|VMODULES            |OK, value: /Users/leodev/.vmodules
|VTMP                |OK, value: /tmp/v_501
|Current working dir |OK, value: /Users/leodev/p/v/test
|                    |
|Git version         |git version 2.52.0
|V git status        |weekly.2026.05-80-g75fc8bed
|.git/config present |true
|                    |
|cc version          |Apple clang version 17.0.0 (clang-1700.0.13.3)
|gcc version         |Apple clang version 17.0.0 (clang-1700.0.13.3)
|clang version       |Apple clang version 17.0.0 (clang-1700.0.13.3)
|tcc version         |tcc version 0.9.28rc 2026-01-10 HEAD@5ec0e6f8 (AArch64 Darwin)
|tcc git status      |thirdparty-macos-arm64 f995efa3
|emcc version        |N/A
|glibc version       |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  repro.v && repro`
```v
fn main() {
	mut x := ?int(none)

	if x == none {
		x = 10
	} else {
		x = 5
	}
}

```

**What did you see?**
```
repro.v:7:3: error: `x` is immutable, declare it with `mut` to make it mutable
    5 |         x = 10
    6 |     } else {
    7 |         x = 5
      |         ^
    8 |     }
    9 | }
```

**What did you expect to see?**

x is mutable, code should be compiled

PR #26495, opened 2026-02-01
checker: preserve mut in other non-none if branch

Delta456:
Fixes https://github.com/vlang/v/issues/26491


spytheman (approved):
Thank you @Delta456 🙇🏻 .
Excellent work.

PR #26522, opened 2026-02-05
all: replace C.int with i32

kbkpbot:
```
v fmt -new_int -w .
```
This will convert all `C.xx(xx int) int` to `C.xx(xx i32) i32`
As `int` will be platform depend, `i32` or `i64`.

related to PR #25298, #25236

BTW: I think we can fix checker to issue a notice about use `int` in C function declaration.


JalonSolov:
Not sure this is a good idea.  According to C specs, the size of int and all other integer types are implementation defined.  The only guarantee is that `int` will be _at least_ large enough to handle `i16` range of values.

On some C compilers (mostly older ones), `int` == `i16`.  On most modern compilers, `int` == `i32`, but on some, `int` == `i32` for 32-bit targets, and `int` == `i64` on 64-bit targets.

spytheman:
> BTW: I think we can fix checker to issue a notice about use int in C function declaration.

A notice is fine, yes.

spytheman:
>  On most modern compilers, int == i32, but on some, int == i32 for 32-bit targets, and int == i64 on 64-bit targets.

On which C ones `int` is not yet fixed?

JalonSolov:
zigcc, for one example, and it is one several people have been trying to use as V's backend (with varying degrees of success).

"In Zig, the size of an `int` can vary depending on the platform, and it's recommended to use explicit types like `i32`, `i64`, etc., for precise size requirements."

spytheman:
Zig the language != zigcc; the `zig cc` subsystem, relies on clang/llvm afaik, and in clang, `int` is 32 bit on 64bit platforms

spytheman:
i.e. Zig is in the position, that V wants to migrate to, and it recommends exactly to avoid `int` in *Zig* code, to avoid problems when interfacing with C.

spytheman:
It also has <https://ziglang.org/documentation/master/#Primitive-Types> i.e. types that start with a `c_` prefix, like `c_int`, `c_uint` etc too.

JalonSolov:
Yep, understood, and if we want to go with it (for now, at least), that's fine.

The C standard itself specifically says that the size is "implementation defined", so it is something to be aware of.

Even if _most_ C compilers these days have settled on `i32` size, it's not guaranteed, and we could run into issues... yes, we can run into issues if we leave it as `int`, too.

spytheman:
> The C standard itself specifically says that the size is "implementation defined", so it is something to be aware of.

> Even if most C compilers these days have settled on i32 size, it's not guaranteed, and we could run into issues... yes, we can run into issues if we leave it as int, too.

Yes. The way Zig handles it - separate types, intentionally prefixed with `c_` is cleaner.

That way, their definition can be updated at compile time, depending on what the target platform expects, mimicking whatever crazy convention is used by whatever C compiler.

On *64bit platforms*, however I do think that `int` is 32 bit, not 16bit and not 64bit, for *ALL* C compilers, used in practice (your counter example about `zig cc` failed, and you did not produce another).

Issue #26474, opened 2026-01-30
Native atomic operations without C dependency


RbPyer:
### Introduction

I have implemented native atomic operations for V using inline assembly. The implementation is currently available as a proof-of-concept module at https://github.com/RbPyer/v-atomics with comprehensive benchmarks available at https://github.com/RbPyer/v-atomics/tree/master/benchmarks

The module provides atomic primitives implemented directly in V without relying on C FFI or libatomic. Currently supported architectures are amd64 and i386. For i386, I deliberately chose to use MMX registers for 64-bit load/store operations despite MMX not being available on all i386 processors. This decision was made because MMX provides significant performance benefits for atomic load and store operations on i386. For processors without MMX support, a fallback to cmpxchg8b could be added in the future, but this would come with a performance penalty.

The implementation includes the following atomic operations for i32, i64, u32, and u64 types: atomic load, atomic store, atomic add (returning the new value after addition), atomic swap (returning the old value), and compare-and-swap. All operations currently provide sequentially consistent semantics with full memory barriers.

The module has been tested extensively with unit tests covering all operations and includes proper alignment checking. However, due to some complexities when building with the -prod flag, likely related to O2/O3 optimizations and current compiler specifics, the behavior on unaligned memory access differs between build modes: with -prod flag, execution immediately hits ud2 instruction (invalid opcode trap), while without -prod it properly calls panic. This is an area that needs further investigation and improvement in the compiler.

Benchmarks demonstrate that the native implementation performs comparably to C-based atomics while eliminating the dependency on libatomic. Performance measurements show that native V atomics achieve equivalent or better performance compared to the current C FFI implementation across different operations and contention scenarios.

### Motivation and Benefits

Currently V implements atomic operations through calls to C functions, which requires linking against libatomic and maintaining FFI bindings. This creates several problems. First, it adds a mandatory dependency on the C toolchain and platform-specific C headers. Second, it makes the actual implementation opaque to V developers and harder to reason about. Third, it complicates cross-compilation because you need the appropriate C toolchain for each target platform.

Native atomic operations would solve these issues by making atomics a first-class feature of V itself. This is particularly important for systems programming use cases where developers want to avoid C dependencies entirely. It would also benefit scenarios like building lock-free data structures, implementing concurrent algorithms without mutexes, and writing high-performance multithreaded applications where understanding the exact memory ordering guarantees is critical.

Having atomics implemented in V also serves an educational purpose by making low-level concurrency primitives visible and understandable to V programmers.

[...]

PR #26529, opened 2026-02-06
vlib: add x.atomics (implement native_x86-64_atomics in V and assembly, without depending on an external C library)

RbPyer:
https://github.com/vlang/v/issues/26474

### DIFF
--- a/vlib/x/atomics/atomics.amd64.v
+++ b/vlib/x/atomics/atomics.amd64.v
@@ -5,49 +5,25 @@ module atomics
 // Panics if dest is not 4-byte aligned.
 pub fn add_i32(dest &i32, delta i32) i32 {
 	mut result := i32(0)
-
-	$if prod {
-		asm volatile amd64 {
-			mov rdx, dest
-			test rdx, 3
-			jz '1f'
-			ud2
-			jmp '2f'
-			1:
-			mov eax, delta
-			lock xadd [rdx], eax
-			add eax, delta
-			mov result, eax
-			2:
-			; =r (result)
-			; r (dest)
-			  r (delta)
-			; eax
-			  rdx
-			  memory
-		}
-	} $else {
-		asm volatile amd64 {
-			mov rdx, dest
-			test rdx, 3
-			jz '1f'
-			call panicUnaligned
-			jmp '2f'
-			1:
-			mov eax, delta
-			lock xadd [rdx], eax
-			add eax, delta
-			mov result, eax
-			2:
-			; =r (result)
-			; r (dest)
-			  r (delta)
-			; eax
-			  rdx
-			  memory
-		}
+	asm volatile amd64 {
+		mov rdx, dest
+		test rdx, 3
+		jz '1f'
+		call panicUnaligned
+		jmp '2f'
+		1:
+		mov eax, delta
+		lock xadd [rdx], eax
+		add eax, delta
+		mov result, eax
+		2:
+		; =r (result)
+		; r (dest)
+		  r (delta)
+		; eax
+		  rdx
+		  memory
 	}
-
 	return result
 }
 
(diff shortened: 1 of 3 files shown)


RbPyer:
I fixed the code, there was a problem with the "al" clobber in compilation with tcc, I removed it as it was unnecessary. And msvc, judging by the jobs, doesn't support inline assembly

vlib/x/atomics/README.md
```diff
@@ -0,0 +1,100 @@
...
+---
+
+## Examples
+
+See the [examples](examples/) directory for complete runnable examples.
+
+### Basic Usage
+
+```v
+import atomics
```
spytheman:
```suggestion
import x.atomics
```

vlib/x/atomics/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+	// Atomic swap: returns the old value
+	old := atomics.swap_i32(&value, 100)
+}
+```
+
+### Compare-and-Swap (CAS)
+
+```v
+import atomics
```
spytheman:
```suggestion
import x.atomics
```

vlib/x/atomics/benchmarks/atomic_benchmark.v
```diff
@@ -0,0 +1,376 @@
...
+fn C.atomic_exchange_u64(voidptr, u64) u64
+
+const iterations = 100_000_000
+
+fn keepalive_u64(x u64) {
+	asm volatile amd64 {
+		nop
+		; ; r (x)
+	}
+}
```
spytheman, line 31: nice

vlib/x/atomics/benchmarks/atomic_benchmark.v
```diff
@@ -0,0 +1,376 @@
+module main
+
+import atomics
```
spytheman:
```suggestion
import x.atomics
```

vlib/x/atomics/examples/basic.v
```diff
@@ -0,0 +1,23 @@
+module main
+
+import atomics
```
spytheman:
```suggestion
import x.atomics
```

vlib/x/atomics/examples/counter.v
```diff
@@ -0,0 +1,40 @@
+module main
+
+import atomics
```
spytheman:
```suggestion
import x.atomics
```

vlib/x/atomics/examples/spinlock.v
```diff
@@ -0,0 +1,59 @@
+module main
+
+import atomics
```
spytheman:
```suggestion
import x.atomics
```

vlib/x/atomics/examples/basic.v
spytheman: It is better to use the full vlib import name, since that does not depend on where the code will be later used/copied (it works locally here, because it is inside a subfolder, but users could want to test it in arbitrary locations by copying in their project, where it would break with just `import atomics`) .

vlib/x/atomics/atomics.amd64.v
```diff
@@ -0,0 +1,999 @@
+module atomics
+
+// add_i32 atomically adds delta to the value at dest and returns the new value.
+// The operation is performed with sequential consistency.
+// Panics if dest is not 4-byte aligned.
+pub fn add_i32(dest &i32, delta i32) i32 {
+	mut result := i32(0)
+	// vfmt off
+	$if prod {
```
spytheman: is there still a need for the `// vfmt off` tags here?

vlib/x/atomics/atomics.amd64.v
```diff
@@ -0,0 +1,999 @@
...
+		asm volatile amd64 {
+			mov rdx, dest
+			test rdx, 3
+			jz '1f'
+			ud2
+			jmp '2f'
```
spytheman: Why not just call `panicUnaligned` here too (instead of the `ud2`)? Then there will be no need for the `$if prod {` blocks and all implementations could shrink in 2 🤔

spytheman:
(rebased over master to get ahead of the other changes that can break the CI indirectly)

spytheman:
@RbPyer Sorry for the late review.

Thank you, excellent work.

vlib/x/atomics/atomics.amd64.v
RbPyer: I think I mentioned in my issue that when I compiled the code with the flags -prod -gc none, I got an error calling panic_Unaligned(), as if this function didn't exist. That's why I made this decision, perhaps not entirely correct

https://github.com/vlang/v/issues/26474#:~:text=with%20%2Dprod%20flag%2C%20execution%20immediately%20hits%20ud2%20instruction%20(invalid%20opcode%20trap)%2C%20while%20without%20%2Dprod%20it%20properly%20calls%20panic.%20This%20is%20an%20area%20that%20needs%20further%20investigation%20and%20improvement%20in%20the%20compiler.

spytheman: Thanks.

I think that can be solved by marking the function declaration of panic_Unaligned with `@[markused]`, but I have to test.

spytheman: Fixed in 7193b06 through the use of:
```v
$if prod && (gcc || clang) {
    #flag -Wl,--undefined=panicUnaligned
}
```
to make sure that LTO will not eliminate the panicUnaligned symbol (it is only called from within the assembly blocks, so without it the C compiler in LTO mode will consider it unused, which later will fail linking).

TODO (in another PR): support a new V attribute to emit `__attribute__((used))`, instead of using a `#flag` .

ylluminate:
@RbPyer have you looked at arm64 support yet?

RbPyer:
> @RbPyer have you looked at arm64 support yet?

At the moment, I'm approaching the point of finalizing a stable API for x86-64, after which I will add support for arm64

Issue #26520, opened 2026-02-05
Invalid generic return type syntax is not caught and causes side effects


erkode:
### Describe the bug

The compiler fails to catch an incorrect syntax for generic return types. Using `[T]` instead of `T` as a return type results in a `cgen error` and incorrect code rewriting when using `v fmt`.

### Reproduction Steps

Create a file named `main.v` and then run `v run main.v`:
```v
module main

fn foo[T](v T) [T] { // <- wrong return type syntax
    return v
}

fn main() {
    println(foo[string]('Hello World'))
}
```

### Expected Behavior

A syntax error should be emitted.

### Current Behavior

1. **Compilation error:** Running `v run main.v` produces a low-level C generation error instead of a syntax error:
   ```text
   cgen error: could not generate string method `builtin__void_str` for type `void`
   ```
2. **v fmt corruption:** Running `v fmt -w main.v` transforms the function signature into invalid code:
   ```v
   fn foo[T](v T) [T]void {
       return v
   }
   ```
   The formatter appends `void` to the brackets.

### V version

V 0.5.0

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.0 50a7d536c1ee60f4a9aad8445eb4fc40798ff530.11bf3dd
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS (WSL 2)
|Processor           |6 cpus, 64bit, little endian, AMD Ryzen 5 3500 6-Core Processor
|Memory              |4.51GB/7.72GB
|                    |
|V executable        |/v/v
|V last modified time|2026-02-05 09:24:30
|                    |
|V home dir          |OK, value: /v
|VMODULES            |OK, value: /root/.vmodules
|VTMP                |OK, value: /tmp/v_0
|Current working dir |OK, value: /workspace
|                    |
|Git version         |git version 2.43.0
|V git status        |11bf3dd7
|.git/config present |true
|                    |
|cc version          |cc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|gcc version         |gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
|clang version       |Ubuntu clang version 18.1.3 (1ubuntu1)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
[...]

PR #26530, opened 2026-02-06
parser: disallow `[T]` as return type 

Delta456:
Fixes https://github.com/vlang/v/issues/26520


spytheman (approved):
Excellent work.
Thank you @Delta456 🙇🏻 .

PR #26533, opened 2026-02-06
checker: resolve @VROOT in #flag directives

### LINKED ISSUE #26532: not resolved: resolve_pseudo_variables missing @VROOT handling
## Description

 in  directives is not resolved to an actual path, causing thirdparty object compilation to fail with MSVC. The literal string  is passed through to the compiler command, producing paths like .

## Reproduction

Any vmodule using  will fail. For example, the  vmodule (Scanning `markdown`...
Updating module `markdown` in `%USERPROFILE%\.vmodules\markdown`...
Updated module `markdown`.
Error 0x80090330 returned by DecryptMessage) has this in :

Running  (which imports ) produces:

## Root cause

In , the  /  case (line ~2947) correctly handles  by converting it to  and resolving:

However, the  case delegates to  (line ~3046), which handles , , and  — but **not **. The deprecated  token passes through unresolved.

## Suggested fix

Add  handling to , mirroring the  logic:

## Environment

- V 0.5.0
- Windows, MSVC compiler
> [!NOTE]
>

quaesitor-scientiam:
## Summary
- resolve_pseudo_variables() was missing @VROOT handling for #flag directives, causing literal @VROOT paths in compiler commands
- Added @VROOT to @VMODROOT conversion, matching the existing #include logic

## Test plan
- [x] V self-compiles successfully
- [x] v doc no longer fails with msvc thirdparty object build error
- [ ] Verify with any vmodule using #flag @VROOT/...

Fixes #26532

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -3067,6 +3067,13 @@ fn (mut c Checker) hash_stmt(mut node ast.HashStmt) {
 
 fn (mut c Checker) resolve_pseudo_variables(oflag string, pos token.Pos) ?string {
 	mut flag := oflag
+	if flag.contains('@VROOT') {
+		// @VROOT is deprecated, treat it as @VMODROOT
+		flag = util.resolve_vmodroot(flag.replace('@VROOT', '@VMODROOT'), c.file.path) or {
+			c.error(err.msg(), pos)
+			return none
+		}
+	}
 	if flag.contains('@VEXEROOT') {
 		// expand `@VEXEROOT` to its absolute path
 		flag = flag.replace('@VEXEROOT', c.pref.vroot)


spytheman:
@quaesitor-scientiam `@VROOT` was deprecated a long time ago (several years at this point), and support for it was deliberately removed in e3fda7d 2026-01-11 in the main repo.
The vlang/markdown was also updated to not use it, in its latest update 6ecbf4c 2026-01-11 in its own repo.

spytheman:
If you still have an older markdown module, use `v update markdown` to get its latest changes.

quaesitor-scientiam:
> If you still have an older markdown module, use `v update markdown` to get its latest changes.

```
Updating module `markdown` in `%USERPROFILE%\.vmodules\markdown`...
Skipped module `markdown`. Already up to date.
```

So that can't be the problem

spytheman:
> > If you still have an older markdown module, use `v update markdown` to get its latest changes.
> 
> ```
> Updating module `markdown` in `%USERPROFILE%\.vmodules\markdown`...
> Skipped module `markdown`. Already up to date.
> ```
> 
> So that can't be the problem

yes, it can - you are on windows, and on windows many things can and fail :-|, for example file modification checks ...

imho try removing the folder ~/.vmodules/markdown manually, and then `v install markdown` again. After that verify that ~/.vmodules/markdown does not contain `@VROOT` anymore anywhere in its files.

spytheman:
or alternatively, check that ~/.vmodules/markdown is at commit 6ecbf4c

quaesitor-scientiam:
> or alternatively, check that ~/.vmodules/markdown is at commit 6ecbf4c

It is 6ecbf4c519de2ca4cb26432bb2653c9cb9f17309

PR #26535, opened 2026-02-07
db.redis: fix bug in RESP2, add RESP3, and more tests

JalonSolov:
Bug in previous code would sometimes send CRLF out of sequence.

Code now tries to use RESP3 by default, but falls back to RESP2 if that's what the server supports.

More redis types supported, more error checking added.


vlib/db/redis/redis_test.v
```diff
@@ -1,129 +1,541 @@
...
+	mut db := redis.connect()!
+	defer { db.close() or { eprintln('test_redis_basic: close failed: ${err}') } }
+
+	// Basic health check
 	assert db.ping()! == 'PONG'
 
-	// delete all keys first
-	db.cmd('FLUSHALL')!
+	// Keep DB small and local for fast execution
+	db.cmd('FLUSHALL') or { eprintln('FLUSHALL failed: ${err}') }
```
spytheman, line 59: why not just `!` here, as it was before?
this will ensure that the `FLUSHALL` command also worked; with an eprintln, it could fail, and the CI will still work :-|

vlib/db/redis/redis.v
```diff
@@ -38,9 +82,9 @@ pub struct Config {
 pub mut:
 	host     string = '127.0.0.1' // Redis server host
 	port     u16    = 6379        // Redis server port
-	password string // Redis server password (optional)
-	tls      bool   // Enable TLS/SSL connection
-	version  int = 2 // RESP protocol version (default: v2)
+	password string = os.getenv('VREDIS_PASSWORD') // Redis server password (optional)
```
spytheman, line 85: I am not sure that is a good idea to hard code that in the db.redis module itself.

To me it seems *the app*/*the test* should be responsible for loading configuration metadata from whatever it decides - be it an .env file, an ENV variable, a command line or some other source. The decisions that different apps will make, are different.


edit: the more I think about it, the less I like it; none of the other db modules hardcode such decisions/use os.getenv in their configurations too.
Please remove that new default value.

vlib/db/redis/redis.v
```diff
@@ -358,8 +612,15 @@ pub fn (mut db DB) expire(key string, seconds int) !bool {
 		db.write_data(db.cmd_buf)!
 
 		// read resp
-		resp := db.read_response()! as i64
-		return resp != 0
+		rv := db.read_response()!
+
+		// normalize to boolean result as before
+		match rv {
+			i64 { return (rv as i64) != 0 }
```
spytheman, line 619: hm, would not `rv` be already `i64` here?

vlib/db/redis/redis_test.v
JalonSolov: Testing slop... removed on next commit.

vlib/db/redis/redis.v
JalonSolov: Agreed.  More testing slop, removed on next commit.

vlib/db/redis/redis.v
JalonSolov: Yes, it is.  Fixed.  Along with all the other places...

PR #26540, opened 2026-02-07
regex: fixed missing OR operation on anchor flag

penguindark:
**What's inside**

- Fixed missing `|` while setting `re.flag |= f_ms` and `re.flag |= f_me`


penguindark:
@spytheman Why is this PR failing?
It seems not related to regex, can we merge it anyway?

PR #26545, opened 2026-02-07
regex: new PCRE compliant regex submodule

penguindark:
**What's inside**
A new PCRE regex submodule located in `vlib/regex/pcre`

The `regex.pcre` module provides a **Virtual Machine (VM)** based regular expression engine with 
UTF-8 support.
Unlike recursive engines, this implementation uses an explicit heap stack, 
making it safe for complex patterns and long strings without risking stack overflows.

It supports compilation of patterns, searching, full matching, global replacement, named groups,
and iterative searching.

## Supported Syntax

| Feature | Syntax | Description |
| :--- | :--- | :--- |
| **Literals** | `abc` | Matches exact characters. |
| **Wildcard** | `.` | Matches any character (excluding `\n` unless `(?s)` flag is used). |
| **Alternation** | `\|` | Matches the left OR right expression (e.g., `cat|dog`). |
| **Quantifiers** | `*` | Matches 0 or more times. |
| | `+` | Matches 1 or more times. |
| | `?` | Matches 0 or 1 time. |
| | `{m}` | Matches exactly `m` times. |
| | `{m,n}` | Matches between `m` and `n` times. |
| **Groups** | `(...)` | Capturing group. |
| | `(?:...)` | Non-capturing group. |
| | `(?P<name>...)` | Named capturing group. |
| **Anchors** | `^` | Matches start of string (or line start with `(?m)`). |
| | `$` | Matches end of string (or line end with `(?m)`). |
| | `\b` | Matches a word boundary (start/end of word). |
| | `\B` | Matches a non-word boundary. |
| **Classes** | `[abc]` | Matches any character in the set. |
| | `[^abc]` | Matches any character NOT in the set. |
| | `[a-z]` | Matches a range of characters. |
| | `\w`, `\W` | Word / Non-word character (`[a-zA-Z0-9_]`). |
| | `\d`, `\D` | Digit / Non-digit. |
| | `\s`, `\S` | Whitespace / Non-whitespace. |
| | `\a` | Lowercase character (`[a-z]`). |
| | `\A` | Uppercase character (`[A-Z]`). |
| **Escapes** | `\xHH` | Matches 1-byte hex value. |
| | `\XHHHH` | Matches 2-byte hex value. |
| **Flags** | `(?i)` | Case-insensitive matching. |
| | `(?m)` | Multiline mode (`^` and `$` match start/end of lines). |
| | `(?s)` | Dot-all mode (`.` matches `\n`). |


vlib/regex/pcre/regex_test.v
```diff
@@ -0,0 +1,522 @@
...
+	test_non_capturing_groups()
+
+	// New features tests
+	test_find_all()
+	test_find_from()
+	test_replace()
+	// test_stress_vm()
+
+	println('\nAll tests passed!')
+}
```
spytheman, line 29: There is no need for a main function in a _test.v file - all test_ functions will be run by the builtin test framework.

PR #26546, opened 2026-02-07
db.redis: fix bug in RESP2, add RESP3, and more tests

JalonSolov:
Bug in previous code would sometimes send CRLF out of sequence.

Code now tries to use RESP3 by default, but falls back to RESP2 if that's what the server supports.

More redis types supported, more error checking added.

### DIFF
--- a/vlib/v2/transformer/transformer_test.v
+++ b/vlib/v2/transformer/transformer_test.v
@@ -337,7 +337,7 @@ fn test_transform_map_init_expr_non_empty_lowers_to_runtime_ctor() {
 	assert result is ast.CallExpr, 'expected CallExpr, got ${result.type_name()}'
 	call := result as ast.CallExpr
 	assert call.lhs is ast.Ident
-	assert (call.lhs as ast.Ident).name == 'builtin__new_map_init_noscan_value'
+	assert (call.lhs as ast.Ident).name == 'new_map_init_noscan_value'
 	assert call.args.len == 9, 'expected 9 args for map constructor, got ${call.args.len}'
 	assert call.args[7] is ast.ArrayInitExpr, 'expected key array arg'
 	assert call.args[8] is ast.ArrayInitExpr, 'expected value array arg'
@@ -497,7 +497,7 @@ fn test_expand_single_or_expr_with_return_in_or_block() {
 
 	// The if-block body should contain only the return statement (err not used, so no err assign)
 	if_stmt := (prefix_stmts[1] as ast.ExprStmt).expr as ast.IfExpr
-	assert if_stmt.stmts.len == 1, 'expected 1 stmt in if body (return only, err not used), got ${if_stmt.stmts.len}'
+	assert if_stmt.stmts.len == 2, 'expected 2 stmt in if body (return only, err not used), got ${if_stmt.stmts.len}'
 
 	// base_type is unknown => defaults to 'int' => not void => returns data access
 	assert result is ast.SelectorExpr, 'expected SelectorExpr (.data access) for default base type'


JalonSolov:
Note: I hate git.

Previous changes got so out of whack, I wound up deleting the branch the previous PR was based on, so it got closed.

vlib/db/redis/redis_test.v
```diff
@@ -77,23 +145,402 @@ fn test_redis_pipeline() {
...
+
+	// store and retrieve
+	assert db.set('hugebin', big.clone())! == 'OK'
+	got := db.get[[]u8]('hugebin')!
+	assert got.len == big.len
+	// spot checks
+	assert got[0] == big[0]
+	assert got[1023] == big[1023]
+	assert got[1024] == big[1024]
+	assert got[size - 1] == big[size - 1]
```
spytheman: why not just do the full comparison with: `assert got == big` ?
I think it will still be much faster than the rest (sending/receiving from redis will dominate the time for this test, so adding a few ms more will not matter much, while checking the whole arrays can detect subtle data corruptions).

vlib/db/redis/redis_test.v
```diff
@@ -77,23 +145,402 @@ fn test_redis_pipeline() {
...
+	}
+	results := db.pipeline_execute()!
+	// Expect 2 * count responses
+	assert results.len == count * 2
+	for i in 0 .. count {
+		set_idx := i * 2
+		get_idx := i * 2 + 1
+		// SET should return OK
+		match results[set_idx] {
+			string { assert (results[set_idx] as string) == 'OK' }
```
spytheman, line 374:
```suggestion
			string { assert results[set_idx] == 'OK' }
```

vlib/db/redis/redis_test.v
```diff
@@ -77,23 +145,402 @@ fn test_redis_pipeline() {
...
+			else { assert false }
+		}
+		// GET should return the value as []u8
+		expected := ('v${i}').bytes()
+		match results[get_idx] {
+			[]u8 { assert (results[get_idx] as []u8) == expected }
```
spytheman, line 380:
```suggestion
			[]u8 { assert results[get_idx] == expected }
```

vlib/db/redis/redis_test.v
```diff
@@ -77,23 +145,402 @@ fn test_redis_pipeline() {
...
+	mut x := seed
+
+	count := 50
+	for iter := 0; iter < count; iter++ {
+		// size up to ~200 KiB
+		size := int((x % 200_000) + 1)
+		mut data := []u8{len: size}
+		for i in 0 .. size {
+			x = x * 6364136223846793005 + 1442695040888963407
+			data[i] = u8((x >> 16) & 0xFF)
```
spytheman:
```suggestion
			data[i] = rand.u8()
```

vlib/db/redis/redis_test.v
JalonSolov: Doesn't work.  Apparently the `assert` has to "re-resolve" the sum type.

If I make the change you suggested, it gives an error.

JalonSolov: I plan to enter a separate issue about that...

vlib/db/redis/redis_test.v
JalonSolov: Same as above.

vlib/db/redis/redis_test.v
JalonSolov: Hmm...

JalonSolov:
Heh.  CI failures for missing docs in files that aren't part of this PR.

vlib/db/redis/redis_test.v
JalonSolov: No noticeable change in speed, but it is a lot cleaner looking.

It was mainly that way to match the individual assignments above, but... all it needs to do is prove it read back what it wrote, so a straight `==` works.

vlib/db/redis/redis_test.v
JalonSolov: Yep.  Makes sense, and it was even simpler... just added `, init: rand.u8()` to the `data` declaration, and removed the whole `for` loop.

vlib/db/redis/redis_test.v
JalonSolov: See #26550

vlib/db/redis/redis_test.v
```diff
@@ -77,23 +146,361 @@ fn test_redis_pipeline() {
...
+	db.cmd('FLUSHALL')!
+
+	// LCG PRNG seed using process id
+	mut x := u64(os.getpid())
+
+	count := 50
+	for iter in 0 .. count {
+		// size up to ~200 KiB
+		size := int((x % 200_000) + 1)
+		mut data := []u8{len: size, init: rand.u8()}
```
spytheman: this is a bit tricky, since just `init: rand.u8()` will make it call `rand.u8()` just once, and initialize all elements with that value:
```
>>> []u8{len: 10, init: rand.u8()}
[140, 140, 140, 140, 140, 140, 140, 140, 140, 140]
>>> []u8{len: 10, init: rand.u8()}
6, 46, 46, 46, 46, 46, 46, 46]
>>> []u8{len: 10, init: rand.u8()}
[147, 147, 147, 147, 147, 147, 147, 147, 147, 147]
```

in order to make it not cache the expression, it has to use `index` currently:
```
>>> []u8{len: 10, init: rand.u8() + 0*index}
[212, 49, 161, 33, 89, 6, 182, 27, 80, 222]
```

spytheman: I think, it may be a good idea to implement something like an explicit `@[impure]` tag, and then mark all `rand` functions with it, and then make the array initialization not cache expressions that have calls to impure functions in it, even if they do not involve the `index` .

spytheman: otherwise it will continue to be a common gotcha - it has bit me, I've seen others on Discord hit it too, and now you as well

spytheman: or `@[side_effect]` ?

spytheman: or `@[stateful]` ?

spytheman:
(rebased over latest master, to avoid an unrelated docs job failure)

vlib/db/redis/redis.v
```diff
@@ -60,9 +103,42 @@ pub fn connect(config Config) !DB {
...
+
+	// send HELLO and attempt to read response. If any step fails, fallback to RESP2 + AUTH.
+	db.write_data(db.cmd_buf) or {
+		// write failed (connection error?) — fallback to RESP2 and perform AUTH if needed
+		db.version = 2
+		if config.password.len > 0 {
+			db.auth(config.password)!
+		}
+		return db
+	}
```
spytheman: it may be worth extracting this in a method, called here and a few lines below too

vlib/db/redis/redis.v
```diff
@@ -393,21 +648,88 @@ fn (mut db DB) read_response_bulk_string() !RedisValue {
...
+	if data_length == -1 {
+		return RedisNull{}
+	}
+
+	// If zero-length payload, read exactly the 2-byte terminator CRLF reliably
 	if data_length == 0 {
-		mut terminator := []u8{len: 2}
-		db.read_data(mut terminator)!
-		if terminator[0] != `\r` || terminator[1] != `\n` {
+		mut term := []u8{len: 2}
```
spytheman:
```suggestion
		mut term := [2]u8{}
```
will avoid the allocation - the array seems to be used as just a fixed buffer anyway

vlib/db/redis/redis.v
```diff
@@ -393,21 +648,88 @@ fn (mut db DB) read_response_bulk_string() !RedisValue {
...
+			n := db.read_ptr_data(ptr, 2 - total_term)! // read remaining terminator bytes
+			if n == 0 && total_term < 2 {
+				return error('`read_response_bulk_string()`: incomplete terminator for empty string')
+			}
+			total_term += n
+		}
+		if term[0] != `\r` || term[1] != `\n` {
 			return error('invalid terminator for empty string')
 		}
 		return []u8{}
```
spytheman, line 674:
```suggestion
		return []
```

vlib/db/redis/redis.v
```diff
@@ -393,21 +648,88 @@ fn (mut db DB) read_response_bulk_string() !RedisValue {
...
+		if n == 0 && term_read < 2 {
+			return error('`read_response_bulk_string()`: incomplete terminator after payload')
+		}
+		term_read += n
+	}
+	if term[0] != `\r` || term[1] != `\n` {
+		return error('`read_response_bulk_string()`: invalid data terminator')
+	}
+
+	return data_buf.clone()
```
spytheman: why clone it?
it is already allocated, and imho can be just returned as is?

vlib/db/redis/redis.v
```diff
@@ -417,16 +739,196 @@ fn (mut db DB) read_response_bulk_string() !RedisValue {
...
+fn (mut db DB) read_resp3_set_payload() !RedisSet {
+	header := db.read_header()!
+	count := header.i64()
+	if count == -1 {
+		return RedisSet{
+			elements: []RedisValue{}
+		}
 	}
+	mut elems := []RedisValue{cap: int(count)}
+	for _ in 0 .. int(count) {
```
spytheman: hm, why cast here? count is i64, and the `for x in 0..count {` should work

spytheman: an alternative is also `count := header.int()`, which will also eliminate the cast from the `cap: int(count)` above too

vlib/db/redis/redis.v
JalonSolov: This change (after changes above changed line numbers) gives
```
redis.v:672:10: error: array_init: no type specified (maybe: `[]Type{}` instead of `[]`)
  670 |             return error('invalid terminator for empty string')
  671 |         }
  672 |         return []
      |                ~~
  673 |     }
  674 |
redis.v:672:3: error: `[]` used as value
  670 |             return error('invalid terminator for empty string')
  671 |         }
  672 |         return []
      |         ~~~~~~~~~
  673 |     }
  674 |
```

vlib/db/redis/redis.v
JalonSolov: Yes... whole section changed in next push.

vlib/db/redis/redis.v
JalonSolov: Done.  Slightly cleaner looking, but only saved 3 lines of code, total.

And it now hides setting the version to `2` over the default `3`, unless you go and look at that function.

vlib/db/redis/redis.v
JalonSolov: True, but only because of the current "magic" that all dynamic arrays are allocated on the heap.  Otherwise, it's only declared/allocated locally.

vlib/db/redis/redis.v
JalonSolov: Cast is not needed (except on the `cap:` bit), but safer to keep `.i64()` and the cast, as we have no idea how large the data will be.

If it's > 2G, `int` won't work... at least, not until `int` adjusts to target system size, but then it won't work for > 2G on 32-bit systems.

JalonSolov:
I'm about to give up on this whole thing.  I now _despise_ git.

It has gotten me into some sort of weird merge situation, where I would have been pushing 37 files when I only changed 2.  That is unacceptable.

I've blown away my local branch, and will be re-doing the most recent changes, and that will _hopefully_ get me back on track.  <sigh>

vlib/db/redis/redis_test.v
JalonSolov: I agree we need something.  My expectation with the change was that `rand.u8()` would be called every time through to init loop, not just once.  Not sure the best way to do so, though.

PR #26547, opened 2026-02-07
crypto.ripemd160: improve block(), update tests

tankf33der:
Current `ripemd160` implementation translated from [Go](https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.47.0:ripemd160/), but it's not that simple.
Let's go through the patch from top to bottom:
1. `free()` - added second array to free
2. typo
3. `hexhash()` - remove second init(), already done in `new()`
4. test file - lets test `hexhash()` too
5. switch arrays from u32 to u8, less memory
6. big boss `block()` - remove unnecessary `clone()` and rest trivial optimization(s).

My standard 10MB bench code and cheap laptop:
```
import crypto.ripemd160
import time

fn main() {
	a := []u8{len: 10_000_000}
	t1 := time.now()
	println(ripemd160.hexhash(a.bytestr()))
	println(time.since(t1))
}
```
`master`
```
$ v run ripem.v && v -prod ripem.v && xtime ./ripem
b52e79a8919bb82acb98ce46527f2b822fc9f3ee
2:47.822
b52e79a8919bb82acb98ce46527f2b822fc9f3ee
2:54.161
```

`patch`
```
$ v run ripem.v && v -prod ripem.v && xtime ./ripem
b52e79a8919bb82acb98ce46527f2b822fc9f3ee
217.005ms
b52e79a8919bb82acb98ce46527f2b822fc9f3ee
36.332ms
```

`tcc` will work `700+` times faster
`gcc` will work `4800+` times faster

### DIFF
--- a/vlib/crypto/ripemd160/ripemd160block.v
+++ b/vlib/crypto/ripemd160/ripemd160block.v
@@ -7,7 +7,7 @@ import math.bits
 
 // vfmt off
 const n__ = [
-	u32(0), 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+	u8(0), 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
 	7, 4, 13, 1, 10, 6, 15, 3, 12, 0, 9, 5, 2, 14, 11, 8,
 	3, 10, 14, 4, 9, 15, 8, 1, 2, 7, 0, 6, 13, 11, 5, 12,
 	1, 9, 11, 10, 0, 8, 12, 4, 13, 3, 7, 15, 14, 5, 6, 2,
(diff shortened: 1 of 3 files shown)


spytheman (approved):
That is a ***super impressive*** speedup!
Excellent work @tankf33der 🙇🏻 .
Thank you.

vlib/crypto/ripemd160/ripemd160block.v
```diff
@@ -7,66 +7,62 @@ import math.bits
 
 // vfmt off
 const n__ = [
-	u32(0), 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
+	u8(0), 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15,
 	7, 4, 13, 1, 10, 6, 15, 3, 12, 0, 9, 5, 2, 14, 11, 8,
 	3, 10, 14, 4, 9, 15, 8, 1, 2, 7, 0, 6, 13, 11, 5, 12,
 	1, 9, 11, 10, 0, 8, 12, 4, 13, 3, 7, 15, 14, 5, 6, 2,
 	4, 0, 5, 9, 7, 12, 2, 10, 14, 1, 3, 8, 11, 6, 15, 13,
 ]
```
spytheman, line 15: consider also using `const ... = [ ... ]!` here

spytheman: (it can lead to less indirections at runtime)

spytheman:
(the CI failure is completely unrelated)

vlib/crypto/ripemd160/ripemd160block.v
spytheman: The effect with -prod is not much (it is within the noise of my system); without -prod, using fixed array consts, leads to a 2-3% speedup:
image

PR #26557, opened 2026-02-08
vlib: add `os.mmap` module for memory mapping

StunxFS:
This PR adds a new module, `os.mmap`, based on the library created by user @jdonnerstag (which hasn't been updated in 5 years).

The library repository is located here: https://github.com/jdonnerstag/vlang-mmap.

### DIFF
--- a/vlib/os/mmap/mmap_test.v
+++ b/vlib/os/mmap/mmap_test.v
@@ -176,7 +176,7 @@ fn test_mlock_unlock() {
 fn test_mmap_protection_flags() {
 	size := usize(4096)
 
-	// Test read-only mapping
+	// Test read-only mapping (supported on all platforms)
 	read_only_addr := mmap.mmap(
 		len:    size
 		prot:   mmap.prot_read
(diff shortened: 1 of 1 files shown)


vlib/os/mmap/mman_nix.c.v
```diff
@@ -0,0 +1,135 @@
...
+
+// mmap map files or devices into memory
+pub fn mmap(args MmapOptions) !voidptr {
+	addr := C.mmap(args.addr, args.len, args.prot, args.flags, args.fd, args.offset)
+	if addr != C.MAP_FAILED {
+		return addr
+	}
+
+	match C.errno {
+		C.EACCES {
```
spytheman: why not just `return error_posix()` instead?

vlib/os/mmap/mman_windows.c.v
```diff
@@ -0,0 +1,239 @@
...
+#include <windows.h>
+#include <errno.h>
+#include <io.h>
+#include <sys/types.h>
+
+// ----------------------------------------------------------------------------
+// Constants: please make sure, the same constants exist in *_nix.c.v
+// ----------------------------------------------------------------------------
+
+// TODO I would prefer the different constants to be an enum. Should be more type-safe.
```
spytheman: why not?

vlib/os/mmap/mman_windows.c.v
```diff
@@ -0,0 +1,239 @@
...
+
+fn C.FormatMessageA(dwFlags int, lpSource voidptr, dwMessageId int, dwLanguageId int, lpBuffer voidptr, nSize int, arguments voidptr) int
+
+// ----------------------------------------------------------------------------
+// (Windows) Implementations
+// ----------------------------------------------------------------------------
+
+// get_last_error_as_string  Retrieve the (windows) last-error-ID and
+// determine the associated error message
+fn get_last_error_as_string() string {
```
spytheman:
```
code := int(C.GetLastError())
return code, os.get_error_msg(code)
```

spytheman: later in the callers: `code, msg := get_last_error()` and `return error_with_code(msg, code)`.
Alternatively, you can also just:
`return error(os.get_error_msg(int(C.GetLastError())))`
without an additional function here

vlib/os/mmap/mman_windows.c.v
spytheman: You can assign values to each enum label without a problem now, V allows it.
You can even assign with `label = C.name` inside an enum if needed.

vlib/os/mmap/mman_test.v
```diff
@@ -0,0 +1,39 @@
+module mmap
```
spytheman: It is preferable to have normal tests, if they only test public API, so imho just remove the `module map` here, and do `import os.mmap` instead, and adjust the calls.

It has the benefit that later people and AIs can understand it easier, and also use the test as an additional documentation of the behavior of the module from the user's perspective, and copy code from it etc.

spytheman: Can you please also show example usages/tests for the other public APIs too?

spytheman:
@StunxFS thank you for working on this 🙇🏻 .

Can you also modify `examples/1brc/solution/main.v` to use that new module in the same PR, and add a performance comparison before/after for its functionality?

vlib/os/mmap/mman_windows.c.v
StunxFS: The problem is that in `_windows.v` bit operations are used, so I would have to mark those enums as `@[flag]`, but then the compiler won't let me assign values ​​to the fields:

```v
x.v:3:2: error: when an enum is used as a bit field, you can not assign custom values
    1 | @[flag]
    2 | enum BitFlag {
    3 |     abc = 2004
      |     ~~~
    4 | }
```
I prefer to leave those values ​​as constants.

vlib/os/mmap/mman_windows.c.v
StunxFS: Thanks!

vlib/os/mmap/mman_windows.c.v
```diff
@@ -175,8 +144,7 @@ pub fn mmap(args MmapOptions) !voidptr {
 
 	fm := C.CreateFileMapping(h, C.NULL, protect, dw_max_size_high, dw_max_size_low, C.NULL)
 	if fm == C.NULL {
-		msg := get_last_error_as_string()
-		return error('CreateFileMapping() failed: ${msg}')
+		return os.error_win32()
```
spytheman: afaik that will panic on windows, when called with no parameters:

```v
@[params]
pub struct SystemError {
pub:
    msg  string
    code int = error_code_not_set
}
```
and later:

```v
@[inline; manualfree]
pub fn error_win32(e SystemError) IError {
    $if windows {
        if e.code == error_code_not_set {
            panic('before calling `error_win32`, you must set `e.code` first.')
        }
...
```

spytheman:
```v
@[inline; manualfree]
pub fn error_posix(e SystemError) IError {
    code := if e.code == error_code_not_set { C.errno } else { e.code }
    message := if e.msg == '' { posix_get_error_msg(code) } else { e.msg }
    return error_with_code(message, code)
}
```

I think that `error_win32` should be changed to be more like error_posix for the default case, and get the code from `C.GetLastError()` instead of panic-ing

Issue #26551, opened 2026-02-08
Type alias to interface doesn't trigger implicit cast in function calls


kbkpbot:
### Describe the bug

When a type alias points to an interface, V compiler fails to generate the necessary implicit cast code when passing a concrete struct to a function that accepts the aliased interface type. This results in a C compilation error.

### Reproduction Steps

test_v_bug.v
```v
module main

// Define a simple interface
pub interface Node {
        kind() int
        name() string
}

// Implement the interface
pub struct MyNode {
pub:
        value int
        text  string
}

pub fn (n MyNode) kind() int {
        return 1
}

pub fn (n MyNode) name() string {
        return n.text
}

// Define a type alias to the interface
pub type Expr = Node

// Function using the type alias
pub fn process_node(expr Expr) string {
        return expr.name()
}

fn main() {
        node := MyNode{
                value: 42
                text:  'test'
        }
        result := process_node(node) // <-- Error occurs here
        println('Result: ${result}')
}
```

### Expected Behavior

The code should compile successfully. V should implicitly cast MyNode to Node interface type when passing it to process_node, similar to how it works when the parameter type is directly Node.

### Current Behavior

C compilation fails with:

```
================== C compilation error (from tcc): ==============
cc: /tmp/v_1000/test_v_bug.01KGY2TPHZXTKY6JB3Y258YKA2.tmp.c:5060: error: cannot convert 'struct main__MyNode' to 'struct main__Node'
=================================================================
Try passing `-g` when compiling, to see a .v file:line information, that correlates more with the C error.
(Alternatively, pass `-show-c-output`, to print the full C error message).
builder error:
==================
C error found. It should never happen, when compiling pure V code.
This is a V compiler bug, please report it using `v bug file.v`,
or goto https://github.com/vlang/v/issues/new/choose .
You can also use #help on Discord: https://discord.gg/vlang .
```

### V version

V 0.5.0 d19074abb635e9b475b2b39a8c2e8be157535d0b

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.0 d19074abb635e9b475b2b39a8c2e8be157535d0b
|:-------------------|:-------------------
|OS                  |linux, Ubuntu 24.04.3 LTS
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i7-9700 CPU @ 3.00GHz
|Memory              |10.54GB/15.51GB
|                    |
|V executable        |/media/HD/github/kbkpbot/v/v
[...]

PR #26565, opened 2026-02-09
checker,cgen: fix alias for interface and option types

kbkpbot:
Fix issue #26551

Fix type aliases to interface and option types were not properly resolved during type checking and code generation.


vlib/v/tests/aliases/alias_option_test.v
```diff
@@ -0,0 +1,15 @@
...
+
+pub fn is_some(val MaybeInt) bool {
+	return val != none
+}
+
+fn test_alias_option() {
+	v := ?int(42)
+	result := is_some(v)
+	println('Has value: ${result}')
+	assert result == true
```
spytheman:
```suggestion
	assert result
```

PR #26570, opened 2026-02-09
builtin: document flush_stdout and flush_stderr

dy-tea:
Tried out `v quest` today, pointed me to this.


vlib/builtin/printing.c.v
```diff
@@ -55,6 +55,7 @@ pub fn eprint(s string) {
 	}
 }
 
+// flush_stdout flushes the stdout buffer, ensuring all remaining data is written.
```
spytheman, line 58:
```suggestion
// flush_stdout flushes the stdout buffer, ensuring all remaining data is written.
// See also unbuffer_stdout() .
```

PR #26572, opened 2026-02-09
cgen: fix hanging when printing to console in a `-subsystem windows` executable

### LINKED ISSUE #26481: Process hangs when using -subsystem windows if you print to screen.
### Describe the bug

If you try printing to the screen with this code.

```vlang
fn main() {
	println("hello, world!")
	exit(0)
}
```

and these compiler options

```
v -subsystem windows -prod -cc gcc poc.v
```

and then you double click the exe in windows, the process will hang and be present forever until you kill it with system informer/task manager

### Reproduction Steps

```vlang
fn main() {
	println("hello, world!")
	exit(0)
}
```

and these compiler options

```
v -subsystem windows -prod -cc gcc poc.v
```

### Expected Behavior

The exe would exit.

### Current Behavior

The exe hangs and does not exit.

### Possible Solution

_No response_

_No response_

### V version

V 0.5.0 e2f5d6c

Windows 11 Pro
Version	10.0.26100 Build 26100

> [!NOTE]
>

revosw:
Resolves #26481

When you double click an executable from explorer, there is no parent console to attach to. So when trying to println("hello, world!") from an exe that is subsystem windows instead of subsystem console, it just hangs.

Ran this test program, and it successfully made a file `test` after the println
```
module main
import os

fn main() {
  println("hello, world!")
  os.create("test")!
  exit(0)
}
```

### DIFF
--- a/vlib/v/gen/c/cmain.v
+++ b/vlib/v/gen/c/cmain.v
@@ -115,16 +115,16 @@ fn (mut g Gen) gen_c_main_function_only_header() {
 			} else {
 				g.writeln('\tcon_valid = AttachConsole(ATTACH_PARENT_PROCESS);')
 			}
-			g.writeln('\tFILE* res_fp = 0')
-			g.writeln('\terrno_t err')
+			g.writeln('\tFILE* res_fp = 0;')
+			g.writeln('\terrno_t err;')
 			g.writeln('\tif (con_valid) {')
-			g.writeln('\t\terr = freopen_s(&res_fp, "CON", "w", stdout)')
-			g.writeln('\t\terr = freopen_s(&res_fp, "CON", "w", stderr)')
+			g.writeln('\t\terr = freopen_s(&res_fp, "CON", "w", stdout);')
+			g.writeln('\t\terr = freopen_s(&res_fp, "CON", "w", stderr);')
 			g.writeln('\t} else {')
-			g.writeln('\t\terr = freopen_s(&res_fp, "NUL", "w", stdout)')
-			g.writeln('\t\terr = freopen_s(&res_fp, "NUL", "w", stderr)')
+			g.writeln('\t\terr = freopen_s(&res_fp, "NUL", "w", stdout);')
+			g.writeln('\t\terr = freopen_s(&res_fp, "NUL", "w", stderr);')
 			g.writeln('\t}')
-			g.writeln('\t(void)err')
+			g.writeln('\t(void)err;')
 
 			return
 		}


vlib/v/gen/c/cmain.v
```diff
@@ -121,7 +121,12 @@ fn (mut g Gen) gen_c_main_function_only_header() {
 			g.writeln('\terr = freopen_s(&res_fp, "CON", "w", stdout);')
 			g.writeln('\terr = freopen_s(&res_fp, "CON", "w", stderr);')
 			g.writeln('\t(void)err;')
+			g.writeln('} else {')
+			g.writeln('\tFILE* res_fp = 0;')
+			g.writeln('\tfreopen_s(&res_fp, "NUL", "w", stdout);')
+			g.writeln('\tfreopen_s(&res_fp, "NUL", "w", stderr);')
 			g.writeln('}')
+
```
revosw, line 128: Unnecessarily added a newline, you could edit it out

JalonSolov: Personally, I like the occasional blank line, to separate logical blocks of code... `return` is definitely different enough from `g.writeln` that the blank line shows a distinct end of the writeln calls before the return.

vlib/v/gen/c/cmain.v
```diff
@@ -121,7 +121,12 @@ fn (mut g Gen) gen_c_main_function_only_header() {
 			g.writeln('\terr = freopen_s(&res_fp, "CON", "w", stdout);')
 			g.writeln('\terr = freopen_s(&res_fp, "CON", "w", stderr);')
 			g.writeln('\t(void)err;')
+			g.writeln('} else {')
+			g.writeln('\tFILE* res_fp = 0;')
+			g.writeln('\tfreopen_s(&res_fp, "NUL", "w", stdout);')
+			g.writeln('\tfreopen_s(&res_fp, "NUL", "w", stderr);')
 			g.writeln('}')
```
spytheman: imho move the `g.writeln('\tFILE* res_fp = 0;')` and `g.writeln('\terrno_t err;')` statements before the `g.writeln('if (con_valid) {')` and the `g.writeln('\t(void)err;')` after the closing `}`, 
and use `g.writeln('\terr = ` consistently in both branches.

Afaik it is written that way to satisfy `-cstrict` checks.

vlib/v/gen/c/cmain.v
spytheman: @JalonSolov  is right - an empty line is not a problem, and does make the code more readable.

vlib/v/gen/c/cmain.v
revosw: I also fixed the indentation of the code around in my latest commit, so now when running `vnew -subsystem windows -o main.c .` in any project, I get:
```c
int WINAPI wWinMain(HINSTANCE instance, HINSTANCE prev_instance, LPWSTR cmd_line, int show_cmd){
	LPWSTR full_cmd_line = GetCommandLineW(); // Note: do not use cmd_line
	typedef LPWSTR*(WINAPI *cmd_line_to_argv)(LPCWSTR, int*);
	HMODULE shell32_module = LoadLibrary(L"shell32.dll");
	cmd_line_to_argv CommandLineToArgvW = (cmd_line_to_argv)GetProcAddress(shell32_module, "CommandLineToArgvW");
	int ___argc;
	wchar_t** ___argv = CommandLineToArgvW(full_cmd_line, &___argc);
	BOOL con_valid = FALSE;
	con_valid = AttachConsole(ATTACH_PARENT_PROCESS);
	FILE* res_fp = 0;
	errno_t err;
	if (con_valid) {
		err = freopen_s(&res_fp, "CON", "w", stdout);
		err = freopen_s(&res_fp, "CON", "w", stderr);
	} else {
		err = freopen_s(&res_fp, "NUL", "w", stdout);
		err = freopen_s(&res_fp, "NUL", "w", stderr);
	}
	(void)err;
	g_main_argc = ___argc;
	g_main_argv = ___argv;
#if defined(_VGCBOEHM)
	GC_set_pages_executable(0);
	GC_INIT();
#endif
	_vinit(___argc, (voidptr)___argv);
	main__main();
	_vcleanup();
	return 0;
}
```

PR #26581, opened 2026-02-11
parser: fix pub not being detected in -vls-mode

dy-tea:
Fixes https://github.com/vlang/vls/issues/483.

For some reason there was only detection for `mut` here.


spytheman:
Sorry, I will do it - it is working and I have it locally.

PR #26582, opened 2026-02-11
vpcre: Added 'Non-greedy quantifiers' fix #26579

### LINKED ISSUE #26579: regex.pcre cannot compile pattern r"\$t\((.*?)\)
### Describe the bug

I want to use regex.pcre to find and replace matched string with pattern r"\$t\((.*?)\), but failed.

I have reported this in Telegram, I am not sure if it should be reported here and also if this is a bug

### Reproduction Steps

[Playground](https://play.vlang.io/?query=caccece0f6)

### Expected Behavior

Compile successfully and find multiple results
In the playground, it is expected to find groups "common.hello" and "common.name"

### Current Behavior

compile function returns error "Quantifier must follow a token"

### Possible Solution

_No response_

_No response_

### V version

V 0.5.0 aac7792

Manjaro Linux
DISTRIB_ID="ManjaroLinux"
DISTRIB_RELEASE="26.0.2"
DISTRIB_CODENAME="Anh-Linh"
DISTRIB_DESCRIPTION="Manjaro Linux"
> [!NOTE]
>

penguindark:
**What's in**
- added support for non greedy quantifiers `*?`, `+?`, `??`
- tests for non greedy quantifiers
- modified examples

Note: this new feature require more tests in future
This PR should solve the issue #26579

A.I. support: `Gemini 3 Pro Preview`


spytheman (changes requested):
I want to keep the backwards compatibility with the vlang/pcre module.

I do not see why were the shims for that removed? 

They were added in 5b2db04, and are:
`pub fn (m Match) get(idx int) ?string {`
`pub fn (m Match) get_all() []string {`
`pub fn new_regex(pattern string, _ int) !Regex {`
`pub fn (r Regex) match_str(text string, start_index int, _ int) ?Match {`

Also why was the example edited at all, given that it does not exercise the non greedy qualifiers feature?

Imho it and the shims should be restored, and if needed, another example can be added, that shows the new feature too. I am also perfectly fine for it to be just tested, without an example - a regex engine is a complex thing, and not all of its features can be shown in an example, they just need to be documented/tested.

penguindark:
@spytheman 👍 ok I'm not a fan of it but it can be easily implemented, I'll do asap as I'll have few spare time.

spytheman:
> @spytheman 👍 ok I'm not a fan of it but it can be easily implemented, I'll do asap as I'll have few spare time.

Thank you 🙇🏻 .

penguindark:
> Thank you 🙇🏻 .

Done 🙇

spytheman (approved):

spytheman:
Excellent work.
Thanks.

PR #26588, opened 2026-02-12
Add autofree vtest flags to test math.min with time.Time

esquerbatua:
Add vtest flags for autofree in the test file, to check the autofree error with generics


vlib/v/tests/autofree_generic_math_test.v
```diff
@@ -1,3 +1,5 @@
+// vtest vflags: -autofree
+
```
spytheman:
```suggestion
// vtest vflags: -autofree
// vtest build: !sanitize-address-gcc && !sanitize-address-clang
```

esquerbatua: ty!

PR #26590, opened 2026-02-12
tests: support `v -silent` in the remaining test runners as well; note it in AGENTS.md

spytheman:
Goal
============
This PR adds support for a -silent option, to easily reduce the amount of produced output of most test runners. The output, affects directly the amount of context that AI models have to process, so in many cases, reducing it (without losing important details), may allow smaller models to work better, and to avoid hitting the quota limits of larger ones.

Motivation
=============
I've noticed that before, `./v vlib/v/compiler_errors_test.v` consumed ~1% of my weekly quota with codex, due to it producing slightly above 132KB because of the progress messages printed for each processed .vv/.out pair, even if the corresponding test passed.

For a human, the progress messages are helpful, and the amount of them is easy to ignore, since they are printed in place (overwriting the last line). An AI agent monitoring the process, however will consume *all of them*...

Result
=============
Now `./v -silent vlib/v/compiler_errors_test.v` when everything is OK, produces:
```
Summary for -skip-unused tests: 51 passed, 51 total. Elapsed time: 17761 ms.
Summary for comptime define tests: 8 passed, 8 total. Elapsed time: 2368 ms.
Summary for all tests: 1456 passed, 4 skipped, 1460 total. Elapsed time: 128964 ms.
```

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -42,7 +42,7 @@ Get operational from the repo root in three steps:
    * Debug-friendly (recommended): `./v -g -keepc -o ./vnew cmd/v`
 3. Use `./vnew` for everything:
    * Run a file: `./vnew run examples/hello_world.v`
-   * Run tests: `./vnew test vlib/v/`
+   * Run tests: `./vnew -silent test vlib/v/`
    * Format: `./vnew fmt -w path/to/file.v`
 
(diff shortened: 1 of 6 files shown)


esquerbatua:
Great job!

PR #26591, opened 2026-02-12
cgen: fix extra dereference when indexing array of fns passed by pointer (fix #26556)

### LINKED ISSUE #26556: cgen error
`^~~~~~~~` marker points to extra asterisk.

Fails with `gcc` and `clang`.

```
lambda:~$ v -prod -cflags -fno-strict-aliasing -cc clang -keepc test v/vlib/v/tests/builtin_maps/map_and_array_with_fns_test.v
 FAIL     0.000 ms /home/mpech/v/vlib/v/tests/builtin_maps/map_and_array_with_fns_test.v
>> compilation failed:
================== C compilation error (from clang): ==============
cc:       |                                               ^~~~~~~~
cc: /tmp/v_1000/tsession_7f6439c96b28_01KGYVWHCZQX33T3D820BY5JG5/map_and_array_with_fns_test.tmp.c:9128:40: error: operand of type 'void' where arithmetic or pointer type is required
cc:  9128 |         return ((int (*) (string ))((voidptr*)*a->data)[0])(_S("hello"));
cc:       |                                               ^~~~~~~~
cc: 1 warning and 1 error generated.
...
cc: 1 warning and 1 error generated.
(note: the original output was 7 lines long; it was truncated to its first 5 lines + the last line)
===================================================================
builder error: 
==================

To reproduce just failure 1 run:    '/home/mpech/v/v' -prod -cflags -fno-strict-aliasing -cc clang -keepc '/home/mpech/v/vlib/v/tests/builtin_maps/map_and_array_with_fns_test.v'
Summary for all V _test.v files: 1 failed, 1 total. Elapsed time: 1766 ms, on 1 job. Comptime: 1670 ms. Runtime: 0 ms.
```
> [!NOTE]
>

kbkpbot:
…ter (fix #26556)

   When an array of functions is passed as a mutable pointer argument and
   indexed with a direct array access, the generated C code incorrectly
   added an extra `*` dereference before `->data`.

   Before: `((voidptr*)*a->data)[0]`
   After:  `((voidptr*)a->data)[0]`

   The `->` operator already handles the pointer dereference, so the
   explicit `*` was redundant and caused a compilation error.

### DIFF
--- a/vlib/v/gen/c/index.v
+++ b/vlib/v/gen/c/index.v
@@ -301,7 +301,7 @@ fn (mut g Gen) index_of_array(node ast.IndexExpr, sym ast.TypeSymbol) {
 						g.write(')(*(${elem_type_str}*)builtin__array_get(')
 					}
 				}
-				if left_is_ptr && !left_is_shared {
+				if left_is_ptr && !left_is_shared && !is_direct_array_access {
 					g.write('*')
 				}
 			} else if is_direct_array_access {


spytheman (approved):
Thank you @kbkpbot 🙇🏻 .

PR #26592, opened 2026-02-12
autofree: remove duplicate free calls after `return`, triggered by error string interpolation

esquerbatua:
## Problem

Autofree generated duplicate free statements after `return`/`break`/`continue` in or blocks with string interpolation. The duplicates were unreachable but bloated generated C code.

```v
_ := gzip.compress(data) or {
    msg := 'error'
    println('Error: ${msg}')
    return  // frees correctly placed here
}
// duplicate frees generated here (unreachable)
```

Generated C (before fix):
```c
builtin__string_free(&msg);
return;
    builtin__string_free(&msg);  // duplicate
}
```

### Core Fix (`vlib/v/gen/c/cgen.v`)
- Modified `stmts_with_tmp_var` to track `BranchStmt` in addition to `Return`
- Skip `autofree_scope_vars` call when `last_stmt_was_return` is true
- Root cause: autofree cleanup was running after control had already exited the scope

### Test (`vlib/v/gen/c/testdata/`)
- Added `autofree_or_block_string_interp.vv` with reproduction case
- Added `autofree_or_block_string_interp.c.must_have` to verify correct C generation
- Uses existing coutput test framework (runtime tests can't detect unreachable code)

## Result

Generated C (after fix):
```c
builtin__string_free(&msg);
return;
}  // no unreachable code
```


spytheman (approved):
Excellent work.
Thank you @esquerbatua 🙇🏻 .

PR #26599, opened 2026-02-13
autofree: fix nested if expression

esquerbatua:
The autofree fix for match expressions with nested if statements was too broad, causing the compiler to fail bootstrapping. While `g.outer_tmp_var` correctly prevented `_t = if (...)` patterns in match branches, it was being set for all if expressions, breaking nested contexts in the compiler's own codegen.

## Changes

- **Scope `outer_tmp_var` to autofree contexts only**: Added conditions to only set the flag when autofree is enabled AND the if expression requires tmp var handling AND not already in option/result handling paths
- **Preserve existing codegen paths**: Non-autofree code and nested if expressions in other contexts (ternaries, option/result handlers) remain unaffected

## Example

Before: Setting `outer_tmp_var` for all if expressions broke this pattern in `if.v`:
```v
resolved_typ := if condition {
    unwrapped := g.unwrap_generic(node_typ)
    if unwrapped == node_typ && g.cur_fn.return_type.has_flag(.generic) {
        // Inner if expression incorrectly affected by outer_tmp_var
        ...
    }
}
```

After: Inner if expressions only affected when in autofree match contexts, allowing normal codegen to proceed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -173,4 +173,3 @@ issue*
 # ignore cmd2/v stuff
 cmd/v2/*.txt
 cmd/v2/*.json
-vautofree
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -2363,15 +2363,10 @@ fn (mut g Gen) stmts_with_tmp_var(stmts []ast.Stmt, tmp_var string) bool {
 						is_array_fixed_init = true
 						ret_type = stmt.expr.typ
 					}
-					// Check if this is an if expression
-					// Pass the tmp_var to it so it can use it, avoiding invalid C patterns like "_t = if"
-					if stmt.expr is ast.IfExpr {
-						// Only apply this fix if autofree is enabled
-						// Don't apply it if we're already inside an if expression context
-						if g.is_autofree && !g.inside_if_option && !g.inside_if_result {
-							is_if_expr_with_tmp = true
-							g.outer_tmp_var = tmp_var
-						}
+					if stmt.expr is ast.IfExpr && g.is_autofree && !g.inside_if_option
+						&& !g.inside_if_result {
+						is_if_expr_with_tmp = true
+						g.outer_tmp_var = tmp_var
 					}
 				}
 				if !is_noreturn && !is_if_expr_with_tmp {
--- a/vlib/v/gen/c/if.v
+++ b/vlib/v/gen/c/if.v
@@ -181,10 +181,8 @@ fn (mut g Gen) needs_conds_order(node ast.IfExpr) bool {
 }
 
 fn (mut g Gen) if_expr(node ast.IfExpr) {
-	// Check if there's an outer_tmp_var set for this if expression
 	use_outer_tmp := g.outer_tmp_var != ''
 	saved_outer_tmp_var := g.outer_tmp_var
-	// Clear it immediately so nested contexts don't see it
 	if use_outer_tmp {
 		g.outer_tmp_var = ''
 	}


JalonSolov:
Run `v git-fmt-hook install` to avoid pushing code that needs to be formatted.  Or tell the AI to make sure it's formatted for you.

vlib/v/gen/c/cgen.v
```diff
@@ -2364,15 +2363,29 @@ fn (mut g Gen) stmts_with_tmp_var(stmts []ast.Stmt, tmp_var string) bool {
 						is_array_fixed_init = true
 						ret_type = stmt.expr.typ
 					}
+					// Check if this is an if expression
+					// Pass the tmp_var to it so it can use it, avoiding invalid C patterns like "_t = if"
+					if stmt.expr is ast.IfExpr {
+						// Only apply this fix if autofree is enabled
+						// Don't apply it if we're already inside an if expression context
+						if g.is_autofree && !g.inside_if_option && !g.inside_if_result {
```
spytheman: the comments are duplicating the code itself 1:1, which makes it harder to read; please remove them

vlib/v/gen/c/if.v
```diff
@@ -181,15 +181,26 @@ fn (mut g Gen) needs_conds_order(node ast.IfExpr) bool {
 }
 
 fn (mut g Gen) if_expr(node ast.IfExpr) {
+	// Check if there's an outer_tmp_var set for this if expression
+	use_outer_tmp := g.outer_tmp_var != ''
+	saved_outer_tmp_var := g.outer_tmp_var
+	// Clear it immediately so nested contexts don't see it
```
spytheman: Same - the comments here are less clear, than the code itself and should be removed here; it is a common pattern for saving the state, there is no need to read several lines more, and think while reading, why was it needed to be commented on, and how is it different from everywhere else :-| .

vlib/v/gen/c/if.v
```diff
@@ -181,15 +181,26 @@ fn (mut g Gen) needs_conds_order(node ast.IfExpr) bool {
...
 	// For if expressions with multiple statements or another if expression inside, it's much
 	// easier to use a temp var, than do C tricks with commas, introduce special vars etc
 	// (as it used to be done).
 	// Always use this in -autofree, since ?: can have tmp expressions that have to be freed.
-	needs_tmp_var := g.inside_if_option || g.need_tmp_var_in_if(node)
+	needs_tmp_var := g.inside_if_option || g.need_tmp_var_in_if(node) || use_outer_tmp
 	needs_conds_order := g.needs_conds_order(node)
-	tmp := if g.inside_if_option || (node.typ != ast.void_type && needs_tmp_var) {
+	tmp := if use_outer_tmp {
+		// Use the tmp var from outer context (e.g. from stmts_with_tmp_var)
```
spytheman, line 199: ok

vlib/v/gen/c/if.v
```diff
@@ -267,7 +278,8 @@ fn (mut g Gen) if_expr(node ast.IfExpr) {
 		}
 		cur_line = g.go_before_last_stmt()
 		g.empty_line = true
-		if tmp != '' {
+		if tmp != '' && !use_outer_tmp {
+			// Only declare the tmp var if it's not from outer context
```
spytheman, line 280: ok

.gitignore
```diff
@@ -173,3 +173,4 @@ issue*
 # ignore cmd2/v stuff
 cmd/v2/*.txt
 cmd/v2/*.json
+vautofree
```
spytheman: why?

spytheman (approved):
Excellent work.
Thank you @esquerbatua 🙇🏻 .

PR #26606, opened 2026-02-14
pg: update Windows installation instructions for postgres

revosw:
I have favored downloading the postgres binaries as opposed to building postgres from source, since building from source is unnecessarily complex when you achieve the same thing downloading the binaries.

I checked the DLL requirements for libpq.dll using `ldd.exe` that was available via my installation of git, `C:/Program Files/Git/usr/bin/ldd.exe`


spytheman (approved):
Thank you @revosw 🙇🏻 .

Issue #26458, opened 2026-01-27
V panics on `v -check .` in module with missing import in file


gechandesu:
### Describe the bug

V panics on running `v -check .` in the module with multiple files.

### Reproduction Steps

Create the `foo` directory with files:
```
./
├── connection.v
├── read.v
└── v.mod
```
```v
// connection.v
module foo

import io

struct Connection {
	buf_size int = 4096
mut:
	stream io.ReaderWriter
}
```
```v
// read.v
module foo

// Note the missing `import io` and `if err is io.Eof` check in read() fn

// No issue if uncomment this import
// import io

pub fn (mut c Connection) read() !string {
	mut res := []u8{}
	mut nn := 0
	for {
		mut buf := []u8{len: c.buf_size, cap: c.buf_size}
		n := c.stream.read(mut buf) or {
			if err is io.Eof {
				break
			} else {
				return err
			}
		}
		nn += n
		if n == 0 {
			break
		}
		res << buf[..n]
		if res[nn - 1] == `\n` {
			break
		}
	}
	return res#[..-1].bytestr()
}
```
```v
// v.mod
Module {
	name: 'foo'
	description: ''
	version: '0.0.0'
	license: 'MIT'
	dependencies: []
}
```
Run in `foo/` dir:
```
v -check .
```

### Expected Behavior

Some normal V error.

### Current Behavior

```
V panic: table.sym: invalid type (typ=ast.Type(0x0 = 0) idx=0). Compiler bug. This should never happen. Please report the bug using `v bug file.v`.

 v hash: 9335a84
    pid: 0x12ce5a
    tid: 0x12ce5a
/tmp/v_1000/v2.01KG0H1W5FN6DDC86HDR5FKA8Z.tmp.c:22311: at builtin___v_panic: Backtrace
/tmp/v_1000/v2.01KG0H1W5FN6DDC86HDR5FKA8Z.tmp.c:38705: by v__ast__default_table_panic_handler
/tmp/v_1000/v2.01KG0H1W5FN6DDC86HDR5FKA8Z.tmp.c:38711: by v__ast__Table_panic
/tmp/v_1000/v2.01KG0H1W5FN6DDC86HDR5FKA8Z.tmp.c:39428: by v__ast__Table_sym
/tmp/v_1000/v2.01KG0H1W5FN6DDC86HDR5FKA8Z.tmp.c:7233: by v__checker__Checker_infix_expr
[...]

PR #26607, opened 2026-02-14
v.checker: avoid panic on unresolved infix operand types in -check (fix: https://github.com/vlang/v/issues/26458 )

hedgeg0d:


medvednikov:
Thanks!

PR #26608, opened 2026-02-14
Refactor variant type inference to use checker type information

medvednikov:
## Summary
This PR refactors the sum type variant matching logic to rely on the checker's type information instead of heuristic-based inference. The changes improve type safety and reduce code complexity by centralizing type resolution.

## Key Changes

- **Replaced `infer_variant_type()` with checker-based approach**: The old function used pattern matching on AST nodes to guess variant types. Now uses `get_expr_type()` to get accurate type information from the checker, then converts to C names for matching.

- **Introduced `match_variant()` helper**: New function that matches a C-style type name against variant names, supporting both exact matches and short name matching (e.g., matching `ast__InfixExpr` to `InfixExpr`).

- **Simplified `wrap_sumtype_value()` and `wrap_sumtype_value_transformed()`**: Both functions now use the same type resolution pattern: get type from checker → convert to C name → match against variants.

- **Removed complex inference functions**: Deleted `infer_variant_type()`, `infer_method_array_elem_type()`, and `infer_struct_type()` which contained fragile heuristics. Simplified `infer_enum_type()` and `infer_array_type()` to use `get_expr_type()` directly.

- **Simplified `infer_expr_type()`**: Now delegates to `get_expr_type()` instead of duplicating scope lookup logic.

- **Fixed VLA initialization in C code generation**: Added proper handling for variable-length arrays in fixed array initialization - literal sizes use `= {0}` syntax while VLAs use `memset()` instead.

- **Updated C keywords list**: Added `unix` and `linux` to the reserved C keywords to prevent naming conflicts.

- **Updated comments**: Removed references to deleted `infer_variant_type()` function.

## Implementation Details

The refactoring centralizes type resolution around the checker's type information, making the code more maintainable and less error-prone. The new `match_variant()` function handles the complexity of matching qualified type names (with module prefixes) to variant names, supporting both exact and short-name matching patterns.

https://claude.ai/code/session_012YVBheSYmhRYXqGDFW6XL4

### DIFF
--- a/vlib/v2/gen/cleanc/cleancnew.v
+++ b/vlib/v2/gen/cleanc/cleancnew.v
@@ -3786,8 +3786,8 @@ fn (mut g Gen) gen_stmt_inline(node ast.Stmt) {
 const c_keywords = ['auto', 'break', 'case', 'char', 'const', 'continue', 'default', 'do', 'double',
 	'else', 'enum', 'extern', 'float', 'for', 'goto', 'if', 'inline', 'int', 'long', 'register',
 	'restrict', 'return', 'short', 'signed', 'sizeof', 'static', 'struct', 'switch', 'typedef',
-	'union', 'unsigned', 'void', 'volatile', 'while', '_Bool', '_Complex', '_Imaginary',
-	'unix', 'linux']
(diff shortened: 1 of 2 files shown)


medvednikov:
This was implemented in a recent commit of mine.

PR #26614, opened 2026-02-16
GNUmakefile: Include Macport legacy headers in include path

programmingkidx:
The Macports header files include code that adds newer constants and functions to older versions of Mac OS X. A lot of the code in these header files are needed to build V on Mac OS 10.4. This patch adds the path of the Macport legacy headers to the CFLAGS variable in the GNUmakefile.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -41,6 +41,7 @@ MAC := 1
 TCCOS := macos
 ifeq ($(shell expr $(shell uname -r | cut -d. -f1) \<= 15), 1)
 LEGACY := 1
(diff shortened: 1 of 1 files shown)


spytheman:
What do you think of moving line 113:
```
$(eval override LDFLAGS+=-L$(realpath $(LEGACYLIBS))/lib -lMacportsLegacySupport)
```

to the place where this PR appends to CFLAGS as well?

programmingkidx:
That does sound like a good idea.
Thank you.

PR #26615, opened 2026-02-16
autofree: fix C codegen for complex expressions as function arguments

esquerbatua:
The `-autofree` flag generated invalid C code when function arguments contained complex expressions (if-expressions, match-expressions, or result types with unwrapping). This broke compilation of `examples/fizz_buzz.v` and `examples/random_ips.v`.

## Problem

The `autofree_call_pregen()` function used `cut_to()` to extract generated expression code, but complex expressions insert declarations via `go_before_last_stmt()` before the saved position, resulting in malformed C:

```c
// Generated invalid code:
string _arg_expr_println_1_21 = bool _t1 = true;
((_t1 == ...) ? "FizzBuzz" : ...);  // Syntax error
```

## Solution

Generate expressions to a separate builder to capture all code, then parse based on expression type:

- **If-expressions**: Extract temp variable declarations, filter final temp reference, assign to autofree temp
- **Match-expressions**: Separate bool declaration from ternary, assign ternary result to autofree temp  
- **Result/Option types**: Emit error-checking statements, extract final value, assign to autofree temp

**Modified**: `vlib/v/gen/c/fn.v` (`autofree_call_pregen()`)

**Added tests**: 3 valgrind tests in `vlib/v/slow_tests/valgrind/` for if-expressions, match-expressions, and result unwrapping

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -2435,9 +2435,6 @@ fn (mut g Gen) autofree_call_pregen(node ast.CallExpr) {
 			// Try to find the result variable from the generated code
 			mut result_var := ''
 
-			// Pattern 1: if-expr with "TYPE _tN; /* if prepend */"
-			// Pattern 2: match-expr with "bool _t1 = true;"
-			// Pattern 3: result-expr with "_result_T _tN = ..."
 			for line in lines {
 				trimmed := line.trim_space()
 				if trimmed.contains('/* if prepend */') {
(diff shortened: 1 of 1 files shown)


JalonSolov:
@tankf33der will your changes take care of this as well?

tankf33der:
@Delta456 - Thank you for your vigilance.
@JalonSolov - This can't be merged, at least because some tests works on master. A very suspiciously patch. 
@esquerbatua - By a very-very big coincidence, I'm preparing a patch(es) for examples/*.v. I'll be releasing it gradually and slowly. You'll have to wait. Nice try. 

I won't call it slop, but I vote against it.

PR #26632, opened 2026-02-20
move up AEAD interface from experimental namespace to `cipher.AEAD`

blackshirt:
This PR moves up `interface AEAD` from `vlib/x/crypto/chacha20poly1305.AEAD` space into `cipher.AEAD` space.
I think its should live at that correct space instead the current one. It also follow the `go` version of the same interface.
Its previously has been discussed at [here](https://discord.com/channels/592103645835821068/592320321995014154/1206029352412778577) lengthy.

Cheers.


medvednikov:
Thanks!

PR #26634, opened 2026-02-20
builtin: fix `camel_to_snake` underscore placement in uppercase runs

meopedevts:
`camel_to_snake()` currently places the underscore *after* a run of uppercase
letters when followed by lowercase, which produces wrong results for most
real-world identifiers:

```
'HTMLParser'.camel_to_snake()   == 'htmlp_arser'    // expected: html_parser
'XMLToJSON'.camel_to_snake()    == 'xmlt_o_json'    // expected: xml_to_json
'getHTTPSUrl'.camel_to_snake()  == 'get_httpsu_rl'  // expected: get_https_url
'LastNName'.camel_to_snake()    == 'last_nn_ame'    // expected: last_n_name
```

The problem is in the transition from an uppercase run to a lowercase letter.
When we hit the lowercase char, the underscore should go **before** the last
uppercase (since it starts the new word), not after it:

```
HTMLParser → HTML|Parser → html_parser    (P starts "Parser")
XMLToJSON  → XML|To|JSON → xml_to_json   (T starts "To")
```

This matches the convention used by pretty much every other language/stdlib out
there (Ruby's `ActiveSupport#underscore`, Python's `inflection`, Go community
libs, etc.).

### The fix

When the branch `!c_is_upper && prev_is_upper` is hit (uppercase run ending),
instead of appending `_` at the current position, we shift the last written
uppercase one position forward and insert the `_` before it:

```v
if !c_is_upper && prev_is_upper {
    b[pos] = b[pos - 1]
    b[pos - 1] = `_`
} else {
    b[pos] = `_`
}
```

### Test changes

Some existing test expectations for synthetic/artificial inputs change to match
the standard convention:

| Input      | Before (wrong) | After (correct) |
|------------|----------------|-----------------|
| `AAbb`     | `aa_bb`        | `a_abb`         |
| `aaBBcc`   | `aa_bb_cc`     | `aa_b_bcc`      |
| `AAbbCC`   | `aa_bb_cc`     | `a_abb_cc`      |
| `_ISspace` | `_is_space`    | `_i_sspace`     |
| `_AbCDe_`  | `_ab_cd_e_`    | `_ab_c_de_`     |

These are all artificial cases — `_ISspace` for instance isn't valid camelCase
(should be `IsSpace`). Meanwhile, every real-world identifier I tested is fixed:
`HTMLParser`, `XMLToJSON`, `getHTTPSUrl`, `LastNName`, etc.

No changes to cases that were already correct (`aBcd`, `aaBB`, `aaBbCcDD`,
`aa_BB`, `JVM_PUBLIC_ACC`, etc.).

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -2845,6 +2845,7 @@ pub fn (s string) is_identifier() bool {
 // Example: assert 'aaBB'.camel_to_snake() == 'aa_bb'
 // Example: assert 'BBaa'.camel_to_snake() == 'bb_aa'
 // Example: assert 'HTTPServer'.camel_to_snake() == 'http_server'
+// Example: assert 'HTTP2Server'.camel_to_snake() == 'http2_server'
 // Example: assert 'XML2JSON'.camel_to_snake() == 'xml_2_json'
 @[direct_array_access]
 pub fn (s string) camel_to_snake() string {
@@ -2889,7 +2890,11 @@ pub fn (s string) camel_to_snake() string {
 		mut has_boundary_before_upper := false
 		c := s[i]
 		c_is_upper := c.is_capital()
+		c_is_number := c.is_digit()
 		next_is_lower := i + 1 < s.len && s[i + 1].is_letter() && !s[i + 1].is_capital()
+		next2_is_lower := i + 2 < s.len && s[i + 2].is_letter() && !s[i + 2].is_capital()
+		// Cases: `XML2JSON == xml_2_json` || `HTTP2Server == http2_server`
+		skip_digit := c_is_number && prev_is_upper && !next_is_lower && next2_is_lower
 		// Cases: `HTTPServer == http_server` || `getHTTPSUrl == get_https_url`
 		if c_is_upper && prev_is_upper && i >= 2 && s[i - 2].is_capital() && next_is_lower
 			&& c != `_` {
@@ -2905,7 +2910,7 @@ pub fn (s string) camel_to_snake() string {
 		// TODO: remove this workaround for v2's parser
 		// vfmt off
 		if ((c_is_upper && !prev_is_upper) ||
-			(!c_is_upper && prev_is_upper && s[i - 2].is_capital() && !prev_inserted_boundary)) &&
+			(!c_is_upper && prev_is_upper && s[i - 2].is_capital() && !prev_inserted_boundary && !skip_digit)) &&
 			c != `_` {
 			unsafe {
 				if b[pos - 1] != `_` {
--- a/vlib/builtin/string_test.v
+++ b/vlib/builtin/string_test.v
@@ -1679,7 +1679,7 @@ fn test_camel_to_snake() {
 	assert '_a_Bcd'.camel_to_snake() == '_a_bcd'
 	assert '_AbCDe_'.camel_to_snake() == '_ab_cd_e_'
 	assert 'HTTPServer'.camel_to_snake() == 'http_server'
-	assert 'HTTP2Server'.camel_to_snake() == 'http_2_server'
+	assert 'HTTP2Server'.camel_to_snake() == 'http2_server'
 	assert 'XML2JSON'.camel_to_snake() == 'xml_2_json'
 }
 


StunxFS:
Hey! Thanks for your contribution 🙇

Just one small detail (which the Codex review points out):

```v
$ cat x.v
println("HTTPServer".camel_to_snake())
println("HTTP2Server".camel_to_snake())
$ v run x.v
http_server
htt_p2_server
```

`HTTP2Server` becomes `htt_p2_server`, shouldn't it be `http2_server`? 🤔

medvednikov:
A good pr, but yeah that has to be fixed

omartelo:
Thanks for the feedback!!

I actually missed this case in my tests.

The current implementation produces a different result than `http2_server`:

```vlang
$ cat x.v
println("HTTP2Server".camel_to_snake())
$ v run x.v
http_2_server

```

I think this result is acceptable, although `http2_server` makes more sense. I believe this largely depends on the context.

`HTTP2Server` would read better as `http2_server`.  
`XML2JSON` would probably be better as `xml_2_json`?

By checking whether the current character is a number, we can preserve the current behavior in the new implementation:

```vlang
if b[pos - 1] != `_` {
    if !c_is_number && !c_is_upper && prev_is_upper {
        // Shift the last uppercase one slot forward and insert '_' before it.
        // e.g.: HTTPServer -> http_server, not https_erver
        b[pos] = b[pos - 1]
        b[pos - 1] = `_`
    } else {
        b[pos] = `_`
    }
    pos++
}

```

What do you think?

omartelo:
Refactored `camel_to_snake` to fix a `c2v` regression where identifiers like `_ISspace` were incorrectly generated as `_i_sspace`.

This new implementation uses a look-ahead (`next_is_lower`) to properly handle acronym boundaries. It ensures both the `HTTPServer -> http_server` fix and compatibility with the current codebase (e.g., `_ISspace -> _is_space`), while removing previous character-shifting workarounds.

**Updated test cases:**

-   `_ISspace` -> `_is_space` (Fixed c2v regression)
    
-   `HTTPServer` -> `http_server` (Fix preserved)
    
-   `AAbb` -> `aa_bb`

medvednikov:
> HTTP2Server would read better as http2_server.
> XML2JSON would probably be better as xml_2_json?

Yeah it's always been tricky. I'm not sure how to handle this.

JalonSolov:
Yeah, it hard because _we_ know what we want, but it's not easy to tell the computer how to make the changes correctly.

In those 2 cases, though... since `http2` is correct, and recognized throughout the computer world, maybe add that as a special case?  Then `HTTP2Server` -> `http2_server`, but `XML2JSON` -> `xml_2_json`.

omartelo:
Does this make sense? I handled the following edge case:

If the current char is a digit, preceded by uppercase and followed by an upper -> lower sequence, we bypass the logic and just insert the digit.

It's a bit of a workaround, but it handles the edge case effectively.

JalonSolov:
Indeed.  Nice - no exception list, this way.

JalonSolov:
Here are a few others to try (Camel vs expected snake)... they came up in discussions on discord a while ago...

BeginMode3D -> begin_mode_3d
Http2Mode -> http2_mode
Plus7Days -> plus_7_days
The48thParallel -> the_48th_parallel

omartelo:
Great test cases!

`BeginMode3D -> begin_mode_3d` and `Http2Mode -> http2_mode` can be easily handled by refining the `skip_digit` logic — basically checking if the previous character is lower and the next one is upper.

`Plus7Days` and `The48thParallel` are interesting edge cases. I'll need to give them more thought to ensure we don't break the existing rules while supporting them.

omartelo:
Ok, this is more complex than I initially thought…

`Http2Mode -> http2_mode` is already valid at the moment.

But in my validations, the condition that turns `BeginMode3D -> begin_mode_3d` ends up producing `http_2mode`.

The same applies to `Plus7Days`: generating `plus_7_days` instead of `plus7_days` is still not clear in my mind.

JalonSolov:
`plus_7_days` is correct for `Plus7Days`, so no confusion on that one.  :-)

omartelo:
Yeah, but right now it’s being generated as `plus7_days`. It’s possible to fix this by simply checking the character before and after the number, but that would break `Http2Mode -> http2_mode`.

It’s a double-edged sword: we fix one side and break the other.

JalonSolov:
Yup.  That's why I suggested some sort of list in the first place.  Your workaround seemed decent... until we came up with instances where it isn't.

However, a simple list wouldn't be enough.  It would contain things like common things like `http2`, common suffixes for numbers (like the `th` on `48th`, `st` on `1st`, ...), etc.

The problem with this, of course, is that it would need to be kept up.

A suggestion a while back was to make `camel_to_snake` take an optional list of things, and that could be the list I'm talking about... use a trailing params struct so it can be left out altogether (possibly with a default list like we've talked about), but if people wanted things converted differently, they could pass in their own list.  Regardless, it would have to be a heckuva list... or rather struct, or some such.

All that said - it doesn't have to be done now.  You could take out the workaround so that `HTTP2Server` comes out as `http_2_server` as it was with your original fix, and we could modify it again later.  What you had before is still better than what's in V now.

omartelo:
I get your point. That said, the current implementation already brings some improvements and fixes a few previous conversions that didn’t really make sense — and it does so without breaking backward compatibility. Do you think it makes sense to go ahead and merge this?

About the list-based approach: it makes sense to me, but keeping it up to date could be tricky. Maybe this could be left to the developer’s responsibility? Something like: “We did our best with the defaults — if you need different/custom conversions, provide your own list (or map?) with the expected matches.”

We can definitely mature this idea further — I’m happy to help iterate on it.

JalonSolov:
Yes, I think what you have it good enough to merge, as it does fix most of the "silliness" of the previous code.

I also think it will take some design work to figure out what will make sense for future changes.

omartelo:
Thank you for the feedback. 😁

PR #26639, opened 2026-02-20
sync: remove unused header files from sync_darwin.c.v

programmingkidx:
The file vlib/sync/sync_darwin.c.v includes headers it does not use. These headers prevent building of vlib on certain versions of Mac OS X. Remove these header files to fix building of v on Mac OS X.


medvednikov:
Thanks!

PR #26640, opened 2026-02-21
[fasthttp] Remove zero-byte initialization in read buffer

sibkod:
Fix HTTP request parsing by removing zero-byte buffer initialization

The read buffer was previously created with a fixed length, causing it to be 
filled with zero bytes at creation. When actual request data was appended,
these leading zeros corrupted HTTP headers parsing (e.g., "GET" became "\0\0\0GET").

Now the buffer is created with capacity only, containing only the actual
read bytes without any pre-initialized zeros.


medvednikov:
Nice! Thanks

medvednikov:
Hm ideally it should have a simple test.

PR #26653, opened 2026-02-25
checker: fix panic on wrong arg count with or-block containing if-expr

medvednikov:
When `check_expected_arg_count` detected wrong argument count and
returned early from `fn_call`, `node.return_type` was never set
(staying as 0). Back in `call_expr`, the or-block was still processed
using this zero type, and when `if_expr` inside the or-block called
`table.type_kind(0)` -> `table.sym(0)`, the compiler panicked with
"table.sym: invalid type (typ=ast.Type(0x0 = 0) idx=0)".

Fix: set `node.return_type = func.return_type` before the early return
so the or-block processes correctly even after an argument count error.

https://claude.ai/code/session_01RB9tJf3t9GEuuqcw1inq8k

### DIFF
--- a/vlib/v/checker/fn.v
+++ b/vlib/v/checker/fn.v
@@ -1517,7 +1517,10 @@ fn (mut c Checker) fn_call(mut node ast.CallExpr, mut continue_check &bool) ast.
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


medvednikov:
this was fixed in 28d61f4dfab52f4687236234039a217e612be9f7

PR #26655, opened 2026-02-25
v2: fix transformer deadlock on Linux caused by nested lock on shared scopes

medvednikov:
Four functions in struct.v acquired `lock t.env.scopes` and then called
functions (get_field_type_name, type_to_name, type_to_c_name, is_sum_type)
that also try to acquire the same lock, causing a deadlock on Linux where
mutexes are non-reentrant.

Fix: extract the needed type data while holding the lock, then call the
potentially-locking functions outside the lock block.

Affected functions:
- get_struct_field_type_name
- resolve_struct_field_type
- get_field_array_elem_sumtype_name
- get_field_array_elem_c_name

Also renames the transformer integration test back from _darwin_ to _nix_
since the deadlock that required restricting it to macOS is now fixed.

https://claude.ai/code/session_01RQNT6SFstGf6SNVn8vT5W8


JalonSolov:
In general, it is always better to keep things locked for the briefest possible time, whether locks are re-entrant or not.

vlib/v2/transformer/struct.v
```diff
@@ -1303,51 +1312,59 @@ fn (t &Transformer) resolve_struct_field_type(struct_name string, field_name str
...
 	}
-
+	// Extract the type under lock, call get_field_type_name outside to avoid deadlock.
+	mut found_type := types.Type(types.void_)
+	mut found := false
 	lock t.env.scopes {
 		scope_names := t.env.scopes.keys()
 		for scope_name in scope_names {
 			scope := t.env.scopes[scope_name] or { continue }
-			// Try the lookup name directly in the appropriate module scope
```
StunxFS: I think these comments shouldn't be removed, as they explain what's happening 🤔

Jengro777:
How is this going? I'm looking forward to the improvement of v2

vlib/v2/transformer/struct.v
kbtale, line 1312: The comment in L1313 is not necessary. However, I agree that the L1311 one would be useful. At least it was useful for me when I was reviewing the code.

medvednikov:
Fixed via t.cached_scopes.

Issue #26658, opened 2026-02-26
@[minify] corrupts enum field values when V is built from source with Clang; v fmt fails on every file


quaesitor-scientiam:
### Describe the bug

When V is compiled from source using Clang (`-cc clang`), the resulting binary silently corrupts enum-typed fields in `@[minify]`-annotated structs. All keyword tokens appear as `unknown` kind, causing `v fmt` — and likely all compiler operations — to fail on every V file.

The prebuilt V binary (TCC-compiled) is not affected. The bug is introduced by the Clang rebuild.

The `@[minify]` C backend emits enum bit fields without `unsigned`:

```c
v__token__Kind kind : 7;   // no 'unsigned' — signedness is implementation-defined in C
```

Since V enums are always non-negative, `unsigned int` should be used to make behavior well-defined.

---

### Reproduction Steps

```sh
# On a clean V repo (tested on weekly.2026.08, Windows x64):
cd /path/to/v

# Baseline — prebuilt binary works fine:
./v fmt test.v       # outputs formatted source correctly

# Rebuild with Clang:
./v -cc clang -o vnew cmd/v

# Clang-built binary fails:
./vnew fmt test.v
```

`test.v`:
```v
module main
fn main() { println('hello') }
```

To inspect the generated C bit fields:
```sh
./v -keepc -cc clang -o out test.v
# Search for 'kind :' in the generated .c file
```

Generated C shows:
```c
struct v__token__Token {
    v__token__Kind kind : 7;   // plain enum type, no 'unsigned'
};
struct v__ast__TypeSymbol {
    v__ast__Kind kind : 6;     // plain enum type, no 'unsigned'
};
```

---

### Expected Behavior

`./vnew fmt test.v` outputs the formatted source, regardless of whether V was built with TCC or Clang.

---

### Current Behavior

`./vnew fmt test.v` (with Clang-built binary) fails:

```
invalid expression: unexpected keyword `module`
```

All keyword tokens read as `unknown`. Diagnostic output confirmed the corruption:

```
tok[0]: kind=unknown (-37)  lit=module    // key_module=91; observed -37
tok[2]: kind=unknown (-47)  lit=fn        // key_fn=81;     observed -47
tok[4]: kind=( (52)         lit=          // lpar=52; stored correctly
```

Cascading failures before fix (removing `@[minify]`):

| Struct | Bit field | Affected values | Symptom |
|--------|-----------|----------------|---------|
[...]

PR #26661, opened 2026-02-27
enum struct fields: token, ast: remove @[minify] from structs with enum fields

quaesitor-scientiam:
When V is built from source with Clang, @[minify] generates C bit fields for enum-typed struct fields without 'unsigned'. Clang's handling of enum bit field signedness causes keyword token values (key_module=91 etc.) to be stored as negative numbers, corrupting the AST and breaking v fmt on every file.

Fix: remove @[minify] from Token, TypeSymbol, CallExpr, InfixExpr, AssignStmt, and all other structs in token.v, ast/types.v, ast/ast.v, and ast/attr.v that contain enum-typed fields.

Fixes https://github.com/vlang/v/issues/26658

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -250,7 +250,6 @@ pub:
 	pos token.Pos
 }
 
-@[minify]
 pub struct StringLiteral {
 pub:
 	val      string
@@ -302,7 +301,6 @@ pub enum GenericKindField {
 }
 
 // `foo.bar`
-@[minify]
 pub struct SelectorExpr {
 pub:
 	pos        token.Pos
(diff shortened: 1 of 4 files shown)


quaesitor-scientiam:
The single CI failure (`vlib/datatypes/lockfree/ringbuffer_test.v`, exit code 2) is a pre-existing, unrelated flaky test. It uses an internal watchdog thread that calls `exit(2)` after 10 s to catch deadlocks in lock-free operations on Windows. The test was added 2026-02-16 and already carries a `// vtest retry: 2` annotation. It has no dependency on any of the files changed in this PR (`vlib/v/token/token.v`, `vlib/v/ast/ast.v`, `vlib/v/ast/types.v`, `vlib/v/ast/attr.v`). All 2639 other tests passed.

PR #26678, opened 2026-03-02
cgen: use unsigned int for enum bit fields in @[minify] structs

### LINKED ISSUE #26658: @[minify] corrupts enum field values when V is built from source with Clang; v fmt fails on every file

quaesitor-scientiam:
## Summary

- In C, the signedness of a bit field declared with an enum type is implementation-defined. Clang on Windows x64 treats it as **signed**, so enum values ≥ 64 stored in a 7-bit field become negative — corrupting token kinds, type kinds, operators, etc. TCC and GCC happen to default to unsigned, masking the bug.
- Fix: emit `unsigned int` instead of the C enum type alias when generating bit fields for enum fields in `@[minify]` structs. Three lines changed in `vlib/v/gen/c/struct.v`.
- Adds a regression test with a 65-value enum (forces a 7-bit field) confirming values ≥ 64 are not corrupted under Clang.
- Updates `minify_option_field_not_msvc_windows.c.must_have` to match the new output.

Fixes #26658

## Test plan

- [x] `./vnew -silent vlib/v/gen/c/coutput_test.v` — all pass
- [x] `./vnew_clang -cc clang -cflags "--target=x86_64-pc-windows-gnu" run vlib/v/gen/c/testdata/minify_enum_unsigned_not_msvc_windows.vv` — outputs `true` / `64` (would output `false` / `-64` before the fix)

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -250,6 +250,7 @@ pub:
 	pos token.Pos
 }
 
+@[minify]
 pub struct StringLiteral {
 pub:
 	val      string
@@ -301,6 +302,7 @@ pub enum GenericKindField {
 }
 
 // `foo.bar`
+@[minify]
 pub struct SelectorExpr {
 pub:
 	pos        token.Pos
@@ -351,6 +353,7 @@ pub:
 	pos token.Pos
 }
 
+@[minify]
 pub struct StructField {
 pub:
 	pos              token.Pos
@@ -413,6 +416,7 @@ pub mut:
 }
 
 // const declaration
+@[minify]
 pub struct ConstDecl {
 pub:
 	is_pub bool
@@ -424,6 +428,7 @@ pub mut:
 	is_block     bool         // const() block
 }
 
+@[minify]
 pub struct StructDecl {
 pub:
 	pos           token.Pos
@@ -467,6 +472,7 @@ pub:
 	comments []Comment
 }
 
+@[minify]
 pub struct InterfaceDecl {
 pub:
 	name          string
@@ -510,6 +516,7 @@ pub mut:
 //    ...a
 //    field1: 'hello'
 // }`
+@[minify]
 pub struct StructInit {
 pub:
 	pos             token.Pos
@@ -574,6 +581,7 @@ pub mut:
 }
 
 // function or method declaration
+@[minify]
 pub struct FnDecl {
 pub:
 	name                  string // 'math.bits.normalize'
@@ -667,6 +675,7 @@ pub fn (f &FnDecl) new_method_with_receiver_type(new_type_ Type) FnDecl {
 	}
 }
 
+@[minify]
 pub struct FnTrace {
 pub mut:
 	name string
@@ -678,6 +687,7 @@ pub:
 	is_fn_var   bool
 }
 
+@[minify]
 pub struct Fn {
 pub:
 	is_variadic           bool
@@ -729,6 +739,7 @@ fn (f &Fn) method_equals(o &Fn) bool {
 		&& f.name == o.name
 }
 
+@[minify]
 pub struct Param {
 pub:
 	pos        token.Pos
@@ -794,6 +805,7 @@ fn (p []Param) equals(o []Param) bool {
 }
 
 // break, continue
+@[minify]
 pub struct BranchStmt {
 pub:
 	kind  token.Kind
@@ -869,6 +881,7 @@ pub enum CallKind {
 }
 
 // function or method call expr
+@[minify]
 pub struct CallExpr {
 pub:
 	pos      token.Pos
@@ -926,6 +939,7 @@ pub struct AutofreeArgVar {
 */
 
 // function call argument: `f(callarg)`
+@[minify]
 pub struct CallArg {
 pub:
 	is_mut   bool
@@ -963,6 +977,7 @@ pub enum ComptimeVarKind {
 	aggregate     // aggregate var
 }
 
+@[minify]
 pub struct Var {
 pub:
 	name            string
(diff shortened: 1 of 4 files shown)


vlib/v/gen/c/struct.v
```diff
@@ -734,7 +736,8 @@ fn (mut g Gen) struct_decl(s ast.Struct, name string, is_anon bool, is_option bo
 				}
 			}
 			if !field_is_anon {
-				g.type_definitions.writeln('\t${volatile_prefix}${type_name} ${field_name}${size_suffix};')
+				actual_type := if is_enum_bitfield { 'unsigned int' } else { type_name }
```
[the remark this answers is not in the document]
quaesitor-scientiam: Resolved with commit 4fd491c

vlib/v/gen/c/struct.v
```diff
@@ -734,7 +736,17 @@ fn (mut g Gen) struct_decl(s ast.Struct, name string, is_anon bool, is_option bo
 				}
 			}
 			if !field_is_anon {
-				g.type_definitions.writeln('\t${volatile_prefix}${type_name} ${field_name}${size_suffix};')
+				actual_type := if is_enum_bitfield {
+					enum_info := field_sym.info as ast.Enum
+					if enum_info.typ != 0 && enum_info.typ.is_unsigned() {
+						type_name
+					} else {
+						'unsigned int'
```
JalonSolov, line 745: This doesn't seem correct.  If you had `enum as i8`, it would come out as `unsigned int`.

quaesitor-scientiam: Fixing it now

PR #26682, opened 2026-03-03
thirdparty,sokol: add wayland support

dy-tea:
Adds wayland support to sokol, translates some sokol code to V (such as sfons), also updates the sokol version.

Can be enabled by passing the `-d sokol_wayland` flag during compilation.

The relevant wayland protocol files are bundled for simpler compilation.


JalonSolov:
The .h files are fine, but should the .a files be there?

dy-tea:
Perhaps not, although currently they are expected to be compiled there somehow.
Might be able to remove this step.

dy-tea:
Yeah not sure why I decided to do it that way before, just including the files instead now.

JalonSolov:
@medvednikov what do you think about this?  Do we want this much patching on our copy of sokol?

Wajinn:
Just so people will be aware, there is a fork of X11, called [XLibre](https://github.com/X11Libre/xserver). It has been adopted by a few distributions already. It's an updated version of X11, that will compete with Wayland. Have no idea how it will turn out in the battle between them, and it is good to have both covered.

dy-tea:
x11 is already supported by sokol. Should work fine on XLibre too.

medvednikov:
That's indeed a lot of code.

And I've been recently moving away from sokol. For example no sokol in the game engine, and I have a sokol free gg branch.

Perhaps in our official `sokol` module? `v install sokol`

medvednikov:
Since we're not going to update to another sokol version, perhaps it's fine to merge.

PR #26696, opened 2026-03-06
pref: Give user the ability to specify the default c compiler to use

programmingkidx:
Currently v picks the default compiler to use. For Mac OS it uses /usr/bin/cc. This compiler might be good. Then again it could be old and not meeting v's requirements. The user might also want to compare performance differences between different c compilers. To be able to do this the user needs to be able to specify the default c compiler used by v. This pull request gives the user to do this by setting the use_c_compiler flag to the name of the desired compiler. Then add that flag to the VFLAGS variable.

Here is an example of how to build v using this flag: 
VFLAGS="-d use_c_compiler=clang-mp-3.8" make

### DIFF
--- a/vlib/v/pref/default.v
+++ b/vlib/v/pref/default.v
@@ -284,7 +284,7 @@ pub fn default_tcc_compiler() string {
 pub fn (mut p Preferences) default_c_compiler() {
 	// TODO: fix $if after 'string'
 	$if windows {
-		p.ccompiler = 'gcc'
+		p.ccompiler = $d('use_c_compiler', 'gcc')
 		return
 	}
 	if p.os == .ios {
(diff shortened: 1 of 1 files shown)


JalonSolov:
Rather than modify V itself, (since the `-cc` option works fine), why not modify the GNUMakefile so you can just do `make -clang-mp-3.8` or perhaps better, `make -cc clang-mp-3.8` instead of having to modify VFLAGS?

programmingkidx:
Well on Mac OS 10.4 V uses /usr/bin/cc when using the REPL. This is a link to gcc-4.0. This compiler is pretty old and does not implement the C11 standard that v requires. So without these changes running ./v would always display an error message. Having v use a newer compiler makes the REPL work on Mac OS 10.4.

programmingkidx:
Would also like to note that trying to use the REPL like this still produces an error: ./v -cc clang-mp-3.8

JalonSolov:

JalonSolov:
You should also be able to do
```
VFLAGS="-cc clang-mp-3.8" make
```
without modifying V.  And if you want the REPL to use the same, you can either do
```
VFLAGS="-cc clang-mp-3.8" v
```
or
```
export VFLAGS="-cc clang-mp-3.8"
```
if you don't care about it always being slower than it would be with tcc.

programmingkidx:
Trying 'VFLAGS="-cc clang-mp-3.8" ./v' did work. But there is the convenience factor to consider. The user only has to specify the compiler once if using the way this pull request implements. Doing it the way you mention is not very well known or documented yet.

medvednikov:
We have a one way policy.

VFLAGS works and is documented in docs.md:

image

This new way would have to be documented too, and we'd have 2 ways to do the same thing.

PR #26705, opened 2026-03-08
sokol: fix `v run examples/2048/` on macos (regressed after 5d6de17)

spytheman:


tankf33der:
I confirm that the patch helps, now works.

Issue #26683, opened 2026-03-04
[crypto] Add ML-DSA algorithm


Avey777:
### Describe the feature

The ML-DSA algorithm is more secure in the quantum age, and using this algorithm in JWT can greatly reduce the risk of being cracked

https://www.nist.gov/publications/module-lattice-based-digital-signature-standard

### Use Case

https://github.com/golang/go/tree/master/src/crypto/internal/fips140/mldsa

### Acknowledgements

- [ ] I may be able to implement this feature request
- [ ] This feature might incur a breaking change

### Version used

V 0.5.0 0b465b4

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.0 9a9f3a515d6ef49d3eba07d92996341bba2db9f8.0b465b4
|:-------------------|:-------------------
|OS                  |linux, Deepin 25
|Processor           |6 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
|Memory              |0.24GB/7.6GB
|                    |
|V executable        |/home/Jengro/opt/v/v
|V last modified time|2026-03-02 08:47:03
|                    |
|V home dir          |OK, value: /home/Jengro/opt/v
|VMODULES            |OK, value: /home/Jengro/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/Jengro
|                    |
|Git version         |git version 2.51.0
|V git status        |weekly.2026.06-176-g0b465b47
|.git/config present |true
|                    |
|cc version          |cc (Deepin 12.3.0-17deepin15) 12.3.0
|gcc version         |gcc (Deepin 12.3.0-17deepin15) 12.3.0
|clang version       |Deepin clang version 17.0.6 (5deepin6)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.38-6deepin17) 2.38

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

blackshirt:
[...]

PR #26708, opened 2026-03-09
x.crypto.mldsa: add ML-DSA signature algorithm

cestef:
This adds the ML-DSA signature algorithm, as per [FIPS 204](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf). The code added was ported from [golang's own implementation](https://github.com/golang/go/tree/master/src/crypto/internal/fips140/mldsa) of the algorithm, thus why I added their BSD license to the code.

_They are also in the process of discussing whether to add this impl to the public go `crypto` API: https://github.com/golang/go/issues/77626_

ML-DSA is NIST's postquantum digital signature standard, which is believed to be resistant against quantum computers. It leverages the [shortest vector problem](https://en.wikipedia.org/wiki/Lattice_problem#Shortest_vector_problem_(SVP)) as the one-way function, more specifically the [short integer problem](https://en.wikipedia.org/wiki/Short_integer_solution_problem).

We implement the three parameters sets: `ML-DSA-44`, `ML-DSA-65` and `ML-DSA-87` (specified in FIPS 204, sec. 4).

Even though optional, Pre-Hash ML-DSA is also implemented in `prehash.v` (we basically sign the hash of the message instead of the message itself), and is disabled by default (configurable on `.sign(...)` via `prehash: <hash-alg>`).

The code lives at `vlib/x/crypto/mldsa`, and [XOF](https://en.wikipedia.org/wiki/Extendable-output_function) was added to the `vlib/crypto/sha3` module in `xof.v`. I can split this into a separate PR if you think this is too much.

```v
import crypto.sha3

// one-shot (existing)
digest := sha3.shake256(message, 32)

// streaming (new)
mut xof := sha3.new_shake256()
xof.write(rho)        
xof.write([u8(row)])   
xof.write([u8(col)])   
coefs := xof.read(256) // squeeze out 256 bytes

// we can read again from the same state
more := xof.read(64)

// reset and reuse
xof.reset()
xof.write(seed)
hash := xof.read(64)
```

XOF was needed because ML-DSA makes extensive use of it: we already exposed `sha3.shakeXXX()` oneshot functions, but here we need to stream input in multiple writes before reading output. See [FIPS 202](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.202.pdf) if you're curious about it, I also have a short post with visual explainations of XOFs: https://blog.cstef.dev/posts/sponge

Code is mainly tested by using [NIST's ACVP test vectors](https://pages.nist.gov/ACVP/draft-celi-acvp-ml-dsa.html) (615 total):

- [ML-DSA-keyGen-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-keyGen-FIPS204): `nist_keygen_test.v` (75 vecs)
- [ML-DSA-sigVer-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigVer-FIPS204): `nist_sigver_test.v` (180 vecs)
- [ML-DSA-sigGen-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigGen-FIPS204): `nist_siggen_test.v` (360 vecs)

## Example
```v
import x.crypto.mldsa

fn main() {
        // generate a new ML-DSA-65 key pair
        sk := mldsa.PrivateKey.generate(.ml_dsa_65)!
        pk := sk.public_key()

        // sign a message (with an optional context string)
        msg := 'Hello ML-DSA'.bytes()
        sig := sk.sign(msg, context: 'not-a-drill')!

        // verify the signature with the same context
        verified := pk.verify(msg, sig, context: 'not-a-drill')!
        assert verified // true

        // deterministic signing is also available
        sig2 := sk.sign(msg, context: 'not-a-drill', deterministic: true)!
        verified2 := pk.verify(msg, sig2, context: 'not-a-drill')!
        assert verified2 // true
}
```

<details>
  <summary>Benchmarks (Intel Core Ultra 5 135U)</summary>
<p></p>

`mldsa_bench_test.v` vs `BenchmarkMLDSA` @ [`crypto/internal/fips140test`](https://github.com/golang/go/tree/b7d1c5887561ee5502b6c04baaa27ead13b63971/src/crypto/internal/fips140test)

  | Operation | Param | Go | V | V/Go |
  |-----------|-------|----|---|---------|
  | **KeyGen** | ML-DSA-44 | 165 µs | 120 µs | **1.38** |
  | | ML-DSA-65 | 250 µs | 194 µs | **1.29** |
  | | ML-DSA-87 | 328 µs | 316 µs | **1.04** |
  | **Sign** | ML-DSA-44 | 275 µs | 261 µs | **1.06** |
  | | ML-DSA-65 | 453 µs | 434 µs | **1.04** |
  | | ML-DSA-87 | 498 µs | 469 µs | **1.06** |
  | **Verify** | ML-DSA-44 | 127 µs | 35 µs | **3.66** |
  | | ML-DSA-65 | 186 µs | 47 µs | **3.94** |
  | | ML-DSA-87 | 273 µs | 67 µs | **4.10** |

  </details>

Closes https://github.com/vlang/v/issues/26683

### DIFF
--- a/vlib/crypto/sha3/xof.v
+++ b/vlib/crypto/sha3/xof.v
@@ -16,18 +16,22 @@ mut:
 	squeeze_buf  []u8
 }
 
+// new_shake128 returns a new Shake instance for SHAKE-128 extended output function.
 pub fn new_shake128() &Shake {
 	return &Shake{
 		rate: xof_rate_128
 	}
 }
 
+// new_shake256 returns a new Shake instance for SHAKE-256 extended output function.
 pub fn new_shake256() &Shake {
 	return &Shake{
 		rate: xof_rate_256
 	}
 }
 
+// write absorbs more data into the sponge state.
+// Panics if called after `read`.
 @[direct_array_access]
 pub fn (mut s Shake) write(data []u8) {
 	if s.finalized {
@@ -94,6 +98,8 @@ fn (mut s Shake) finalize() {
 	s.input_buffer = []u8{}
 }
 
+// read squeezes `out_len` bytes from the sponge state.
+// Finalizes the sponge on first call; further calls to `write` will panic.
 @[direct_array_access]
 pub fn (mut s Shake) read(out_len int) []u8 {
 	if !s.finalized {
@@ -119,6 +125,7 @@ pub fn (mut s Shake) read(out_len int) []u8 {
 	return result
 }
 
+// reset clears the sponge state, allowing the Shake instance to be reused.
 pub fn (mut s Shake) reset() {
 	s.s = State{}
 	s.input_buffer = []u8{}


tankf33der:
Too slow tests, they must be faster, move another to https://github.com/vlang/slower_tests

tankf33der:
do `fmt -w .`

cestef:
> Too slow tests, they must be faster, move another to https://github.com/vlang/slower_tests

Are you talking about the bench test ? or all of the tests ?

tankf33der:
> > Too slow tests, they must be faster, move another to https://github.com/vlang/slower_tests
> 
> Are you talking about the bench test ? or all of the tests ?

@cestef 
How I would do it — set aside all test vectors for now, create minimal vectors against the golang implementation, but with full (widest) coverage (!), which should run in no more than one second, and then after the merge, if it happens, add the official vectors to the `slower_tests` repo.

tankf33der:
@blackshirt make some noise. Thank you.

cestef:
> > > Too slow tests, they must be faster, move another to https://github.com/vlang/slower_tests
> > 
> > 
> > Are you talking about the bench test ? or all of the tests ?
> 
> @cestef How I would do it — set aside all test vectors for now, create minimal vectors against the golang implementation, but with full (widest) coverage (!), which should run in no more than one second, and then after the merge, if it happens, add the official vectors to the `slower_tests` repo.

@tankf33der 
so that would imply writing a small go helper to generate those vectors ?

tankf33der:
@cestef 
Make sure this works too and all tests pass before pushing. I returned your tests temporarily and see a bunch of errors, deal with that too.
```
v -cc gcc -cflags -fsanitize=address,undefined test vlib/x/crypto/mldsa/
```

tankf33der:
> > > > Too slow tests, they must be faster, move another to https://github.com/vlang/slower_tests
> > > 
> > > 
> > > Are you talking about the bench test ? or all of the tests ?
> > 
> > 
> > @cestef How I would do it — set aside all test vectors for now, create minimal vectors against the golang implementation, but with full (widest) coverage (!), which should run in no more than one second, and then after the merge, if it happens, add the official vectors to the `slower_tests` repo.
> 
> @tankf33der so that would imply writing a small go helper to generate those vectors ?

@cestef 
If you are capable and smart enough to port an algorithm from Golang, then you have enough brains to make your own test vectors in Golang. Ready to discuss the details privately on Discord if needed.

cestef:
> @cestef Make sure this works too and all tests pass before pushing. I returned your tests temporarily and see a bunch of errors, deal with that too.
> 
> ```
> v -cc gcc -cflags -fsanitize=address,undefined test vlib/x/crypto/mldsa/
> ```

tests run fine without Asan or UBsan. the errors I am getting look more like Boehm GC compatibility issues. see the [bdwgc mailing list](https://bdwgc.opendylan.narkive.com/zptxsigk/gc-address-sanitizer-reports-stack-underflow), https://github.com/ivmai/bdwgc/issues/442 and https://github.com/ivmai/bdwgc/issues/321

Disabling the GC and skipping leaks in Asan runs fine as well:
```bash
ASAN_OPTIONS=detect_leaks=0 v -gc none -cc gcc -cflags -fsanitize=address,undefined test vlib/x/crypto/mldsa/
```

JalonSolov:
Since gc is default, this will need to work with gc on as well.

cestef:
UBsan alone runs fine as well:

```bash
v -cc gcc -cflags '-fsanitize=undefined' test vlib/x/crypto/mldsa
```

cestef:
@tankf33der @JalonSolov

FYI, x/crypto/chacha20, crypto/ed25519, and x/crypto/poly1305 all fail `-fsanitize` with similar errors:
```bash
v -cc gcc -cflags '-fsanitize=address,undefined' test vlib/x/crypto/chacha20/xchacha_test.v vlib/x/crypto/poly1305/poly1305_test.v vlib/crypto/ed25519/internal/edwards25519/point_test.v vlib/crypto/ed25519/internal/edwards25519/element_test.v
```
imo this is because of asan + boehm gc incompatibility

cestef:
I also tried recompiling `thirdparty/libgc/gc.c` with `-fsanitize=address` (so that the `GC_ATTR_NO_SANITIZE_ADDR` annotations are active), but the crashes persist, moving from scan to alloc.

cestef:
https://github.com/bdwgc/bdwgc/issues/789 mentions setting `detect_stack_use_after_return` to false, which seems to work here:

```
ASAN_OPTIONS="detect_stack_use_after_return=0" ./v -cc gcc -cflags '-fsanitize=address,undefined' test vlib/x/crypto/mldsa/
```

JalonSolov:
Which is fine, but the average user of this lib routine will have to know to add that option... better to find another way.

cestef:
@JalonSolov found the proper fix for this. V's bundled bdwgc predates https://github.com/ivmai/bdwgc/commit/5bc30920d which adds `detect_stack_use_after_return=0` to `__asan_default_options()`, the upstream workaround for https://github.com/bdwgc/bdwgc/issues/789.

fix would be to re-amalgamate `thirdparty/libgc/gc.c` from current bdwgc master and have V compile gc.c from source (`-d use_bundled_libgc` maybe?) when `-fsanitize=address` is in `cflags` (since `__asan_default_options` is only compiled when `__SANITIZE_ADDRESS__` is defined). I can open a separate PR for this if needed

PS: this also solves the problem for other failing modules (i.e. chacha20, ed25519, and poly1305)

tankf33der:
@JalonSolov @cestef 

If you read issue 789 more carefully, you'll find a mention of `tankf33der` who proposed this as a workaround. :) I managed to dig into what happens with libgc+ASAN and that's why this is currently the only solution to the problem.

`ASAN_OPTIONS=detect_stack_use_after_return=0`

cestef:
sorry about the force-pushes, forgot to update my global gitignore :sweat_smile:

JalonSolov:
🤷🏼‍♂️ - what can I say... I've only started on this PR merging journey, so I probably missed quite a bit in the past.

I have no problem with the idea of updating V's gc code, if someone wanted to make a PR for it.

PR #26709, opened 2026-03-09
cgen,autofree: don't clone array fn args in struct field init, add test

tankf33der:
Slightly changing the strategy — it will be more beneficial for everyone if `vlib/` works in `-autofree` mode, rather than `examples/`; examples/ will then work on its own if everything is done correctly and in full.

Right now all arrays in -autofree mode are cloned when placed into a `struct`; this is my attempt to implement an exception so that `arrays.reverse_iterator()` does not clone and it is possible to modify data in-place inside the loop.

`v -autofree test vlib/v/tests/structs/` tests continue to pass successfully, just as before the patch.

From now `v -autofree test vlib/arrays/reverse_iterator_test.v` will work in -autofree mode too.


JalonSolov:
Close/re-open to run CI with latest V.

PR #26711, opened 2026-03-09
x.crypto.mldsa: add ML-DSA signature algorithm

cestef:
This adds the ML-DSA signature algorithm, as per [FIPS 204](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.204.pdf). The code added was ported from [golang's own implementation](https://github.com/golang/go/tree/master/src/crypto/internal/fips140/mldsa) of the algorithm, thus why I added their BSD license to the code.

_They are also in the process of discussing whether to add this impl to the public go `crypto` API: https://github.com/golang/go/issues/77626_

ML-DSA is NIST's postquantum digital signature standard, which is believed to be resistant against quantum computers. It leverages the [shortest vector problem](https://en.wikipedia.org/wiki/Lattice_problem#Shortest_vector_problem_(SVP)) as the one-way function, more specifically the [short integer problem](https://en.wikipedia.org/wiki/Short_integer_solution_problem).

We implement the three parameters sets: `ML-DSA-44`, `ML-DSA-65` and `ML-DSA-87` (specified in FIPS 204, sec. 4).

Even though optional, Pre-Hash ML-DSA is also implemented in `prehash.v` (we basically sign the hash of the message instead of the message itself), and is disabled by default (configurable on `.sign(...)` via `prehash: <hash-alg>`).

The code lives at `vlib/x/crypto/mldsa`, and [XOF](https://en.wikipedia.org/wiki/Extendable-output_function) was added to the `vlib/crypto/sha3` module in `xof.v`. I can split this into a separate PR if you think this is too much.

```v
import crypto.sha3

// one-shot (existing)
digest := sha3.shake256(message, 32)

// streaming (new)
mut xof := sha3.new_shake256()
xof.write(rho)        
xof.write([u8(row)])   
xof.write([u8(col)])   
coefs := xof.read(256) // squeeze out 256 bytes

// we can read again from the same state
more := xof.read(64)

// reset and reuse
xof.reset()
xof.write(seed)
hash := xof.read(64)
```

XOF was needed because ML-DSA makes extensive use of it: we already exposed `sha3.shakeXXX()` oneshot functions, but here we need to stream input in multiple writes before reading output. See [FIPS 202](https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.202.pdf) if you're curious about it, I also have a short post with visual explainations of XOFs: https://blog.cstef.dev/posts/sponge

Code is mainly tested by using [NIST's ACVP test vectors](https://pages.nist.gov/ACVP/draft-celi-acvp-ml-dsa.html) (615 total):

- [ML-DSA-keyGen-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-keyGen-FIPS204): `nist_keygen_test.v` (75 vecs)
- [ML-DSA-sigVer-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigVer-FIPS204): `nist_sigver_test.v` (180 vecs)
- [ML-DSA-sigGen-FIPS204](https://github.com/usnistgov/ACVP-Server/tree/master/gen-val/json-files/ML-DSA-sigGen-FIPS204): `nist_siggen_test.v` (360 vecs)

## Example
```v
import x.crypto.mldsa

fn main() {
        // generate a new ML-DSA-65 key pair
        sk := mldsa.PrivateKey.generate(.ml_dsa_65)!
        pk := sk.public_key()

        // sign a message (with an optional context string)
        msg := 'Hello ML-DSA'.bytes()
        sig := sk.sign(msg, context: 'not-a-drill')!

        // verify the signature with the same context
        verified := pk.verify(msg, sig, context: 'not-a-drill')!
        assert verified // true

        // deterministic signing is also available
        sig2 := sk.sign(msg, context: 'not-a-drill', deterministic: true)!
        verified2 := pk.verify(msg, sig2, context: 'not-a-drill')!
        assert verified2 // true
}
```

<details>
  <summary>Benchmarks (Intel Core Ultra 5 135U)</summary>
<p></p>

`mldsa_bench_test.v` vs `BenchmarkMLDSA` @ [`crypto/internal/fips140test`](https://github.com/golang/go/tree/b7d1c5887561ee5502b6c04baaa27ead13b63971/src/crypto/internal/fips140test)

  | Operation | Param | Go | V | V/Go |
  |-----------|-------|----|---|---------|
  | **KeyGen** | ML-DSA-44 | 165 µs | 120 µs | **1.38** |
  | | ML-DSA-65 | 250 µs | 194 µs | **1.29** |
  | | ML-DSA-87 | 328 µs | 316 µs | **1.04** |
  | **Sign** | ML-DSA-44 | 275 µs | 261 µs | **1.06** |
  | | ML-DSA-65 | 453 µs | 434 µs | **1.04** |
  | | ML-DSA-87 | 498 µs | 469 µs | **1.06** |
  | **Verify** | ML-DSA-44 | 127 µs | 35 µs | **3.66** |
  | | ML-DSA-65 | 186 µs | 47 µs | **3.94** |
  | | ML-DSA-87 | 273 µs | 67 µs | **4.10** |

  </details>

Closes https://github.com/vlang/v/issues/26683

---

_This replaces #26708, which was accidentally closed during a history rewrite._

### DIFF
--- a/vlib/x/crypto/mldsa/mldsa.v
+++ b/vlib/x/crypto/mldsa/mldsa.v
@@ -118,8 +118,25 @@ pub fn (sk &PrivateKey) sign(msg []u8, opts SignerOpts) ![]u8 {
 	return sign_internal(sk, mu, slice_to_32(rand.read(32)!))
 }
 
+// sign_mu signs a precomputed mu value with explicit randomness.
+// mu must be 64 bytes. rnd must be 32 bytes (use all zeros for deterministic signing).
+pub fn (sk &PrivateKey) sign_mu(mu []u8, rnd []u8) ![]u8 {
+	if mu.len != 64 {
+		return error('mu must be 64 bytes')
+	}
+	if rnd.len != 32 {
+		return error('rnd must be 32 bytes')
+	}
+	return sign_internal(sk, slice_to_64(mu), slice_to_32(rnd))
+}
+
 pub fn (pk &PublicKey) bytes() []u8 {
 	return pk.raw.clone()
+}
+
+// tr returns the 64-byte transcript hash (H(pk)) used in mu computation.
+pub fn (pk &PublicKey) tr() []u8 {
+	return pk.tr[..]
 }
 
 pub fn (pk &PublicKey) equal(other &PublicKey) bool {
@@ -292,7 +309,8 @@ fn new_public_key(raw []u8, p Params) !PublicKey {
 }
 
 // algo. 2, lines 10-11: M' = 0x00 || |ctx| || ctx || M; mu = H(tr || M', 64)
-fn compute_mu(tr []u8, msg []u8, context string) [64]u8 {
+// compute_mu computes mu = H(tr || M', 64) where M' = 0x00 || |ctx| || ctx || msg.
+pub fn compute_mu(tr []u8, msg []u8, context string) [64]u8 {
 	mut h := sha3.new_shake256()
 	h.write(tr)
 	h.write([u8(0)]) // pure mode domain sep
--- a/vlib/x/crypto/mldsa/mldsa_test.v
+++ b/vlib/x/crypto/mldsa/mldsa_test.v
@@ -145,6 +145,72 @@ fn test_prehash_sign_verify() {
 	}
 }
 
+fn test_field_to_montgomery_roundtrip() {
+	for val in [u32(0), 1, 2, 100, 1000, q - 1] {
+		m := field_to_montgomery(val) or { panic(err) }
+		back := field_from_montgomery(m)
+		assert back == val, 'roundtrip failed for ${val}: got ${back}'
+	}
+}
+
+fn test_field_add_sub() {
+	a := field_to_montgomery(100) or { panic(err) }
+	b := field_to_montgomery(200) or { panic(err) }
+	sum := field_add(a, b)
+	assert field_from_montgomery(sum) == 300
+
+	diff := field_sub(sum, b)
+	assert field_from_montgomery(diff) == 100
+}
+
+fn test_field_mul() {
+	a := field_to_montgomery(1000) or { panic(err) }
+	b := field_to_montgomery(2000) or { panic(err) }
+	prod := field_montgomery_mul(a, b)
+	assert field_from_montgomery(prod) == (1000 * 2000) % q
+}
+
+fn test_ntt_inverse_ntt_roundtrip() {
+	mut f := RingElement{}
+	for i in 0 .. n {
+		f[i] = field_to_montgomery(u32(i % 100)) or { panic(err) }
+	}
+	ntt_f := ntt(f)
+	back := inverse_ntt(ntt_f)
+	for i in 0 .. n {
+		assert field_from_montgomery(back[i]) == field_from_montgomery(f[i]), 'NTT roundtrip failed at index ${i}'
+	}
+}
+
+fn test_ntt_mul_is_polynomial_product() {
+	// (1 + x)^2 ?= x^2 + 2x + 1
+	mut a := RingElement{}
+	a[0] = field_to_montgomery(1) or { panic(err) }
+	a[1] = field_to_montgomery(1) or { panic(err) }
+
+	a_ntt := ntt(a)
+	prod_ntt := ntt_mul(a_ntt, a_ntt)
+	prod := inverse_ntt(prod_ntt)
+
+	assert field_from_montgomery(prod[0]) == 1, 'expected x^2'
+	assert field_from_montgomery(prod[1]) == 2, 'expected 2x'
+	assert field_from_montgomery(prod[2]) == 1, 'expected 1'
+
+	for i in 3 .. n {
+		assert field_from_montgomery(prod[i]) == 0, 'expected 0 at index ${i}, got ${field_from_montgomery(prod[i])}'
+	}
+}
+
+fn test_power2_round() {
+	for val in [u32(0), 1, 100, 1000, q / 2, q - 1] {
+		r := field_to_montgomery(val) or { panic(err) }
+		hi, lo := power2_round(r)
+		reconstructed := field_add(field_to_montgomery(u32(hi) << d) or { panic(err) },
+			lo)
+		assert field_from_montgomery(reconstructed) == val, 'power2_round failed for ${val}'
+	}
+}
+
 fn test_private_key_roundtrip() {
 	for kind in [Kind.ml_dsa_44, .ml_dsa_65, .ml_dsa_87] {
 		seed := []u8{len: 32, init: 0x04}
--- a/vlib/x/crypto/mldsa/prehash.v
+++ b/vlib/x/crypto/mldsa/prehash.v
@@ -52,7 +52,8 @@ fn prehash_message(msg []u8, ph PreHash) []u8 {
 
 // algo. 4, line 23: M' = 0x01 || |ctx| || ctx || OID || PH(M)
 // algo. 7, line 6: mu = H(tr || M')
-fn compute_mu_prehash(tr []u8, msg []u8, context string, ph PreHash) [64]u8 {
+// compute_mu_prehash computes mu for prehash mode: H(tr || 0x01 || |ctx| || ctx || OID || PH(msg), 64).
+pub fn compute_mu_prehash(tr []u8, msg []u8, context string, ph PreHash) [64]u8 {
 	mut h := sha3.new_shake256()
 	h.write(tr)
 	h.write([u8(0x01)]) // domain sep


JalonSolov:
Tsk.  The Markdown stuff is from @medvednikov changelog edits.  Don't worry about those.

JalonSolov:
Closing/re-opening to re-run CI with latest V.

JalonSolov:
You have one example in the readme... how about some tests, to guard against regressions?

vlib/crypto/sha3/xof.v
```diff
@@ -0,0 +1,134 @@
+// Copyright (c) 2023 Kim Shrier. All rights reserved.
+// Use of this source code is governed by an MIT license
+// that can be found in the LICENSE file.
+
+// streaming shake-128/256 xof per FIPS 202
+// https://nvlpubs.nist.gov/nistpubs/fips/nist.fips.202.pdf
+
+module sha3
+
+pub struct Shake {
```
blackshirt, line 11: Is this should be marked with `@[noinit]` tag ? there are constructor for this and i think the outside user should be prohibited to create this opaque directly with `sha3.Shake{}`

vlib/crypto/sha3/xof.v
```diff
@@ -0,0 +1,134 @@
...
+pub fn new_shake256() &Shake {
+	return &Shake{
+		rate: xof_rate_256
+	}
+}
+
+// write absorbs more data into the sponge state.
+// Panics if called after `read`.
+@[direct_array_access]
+pub fn (mut s Shake) write(data []u8) {
```
blackshirt, line 37: I think its should follow the `io.Writer`  and `hash.Hash.write` semantics, so aligns with the rest

vlib/crypto/sha3/xof.v
```diff
@@ -0,0 +1,134 @@
...
+
+	state_bytes := s.s.to_bytes()
+	s.squeeze_buf = state_bytes[..s.rate].clone()
+	s.input_buffer = []u8{}
+}
+
+// read squeezes `out_len` bytes from the sponge state.
+// Finalizes the sponge on first call; further calls to `write` will panic.
+@[direct_array_access]
+pub fn (mut s Shake) read(out_len int) []u8 {
```
blackshirt, line 105: Even we have no `interface XOF`, i think  this also should follow the `io.Reader` semantic

vlib/x/crypto/mldsa/mldsa_test.v
blackshirt: I think this tests must stay here and also other basic tests, ands moves publicly intensive tests into `slower_tests` repo

vlib/x/crypto/mldsa/field.v
```diff
@@ -16,9 +16,9 @@ const mont_minus_one = u32(4186625) // (q-1)*R mod q
 const n = 256
 const d = 13
 
-type FieldElement = u32
-type RingElement = [256]FieldElement
-type NttElement = [256]FieldElement
+pub type FieldElement = u32
```
blackshirt: Is this really used outside the module ? or internal implementation detail

vlib/x/crypto/mldsa/field.v
```diff
@@ -16,9 +16,9 @@ const mont_minus_one = u32(4186625) // (q-1)*R mod q
...
+pub type RingElement = [256]FieldElement
```
blackshirt: What's usecases or benefit export them to the outside ?
Mostly, field element was tightly coupled with internal specific detail of they operates on

cestef:
> You have one example in the readme... how about some tests, to guard against regressions?

@JalonSolov

I was told they should be moved to vlang/slower_tests, see https://github.com/vlang/slower_tests/pull/5

I will be adding a few basic test vectors back here as well, maybe validating against the go impl as @tankf33der mentioned

tankf33der:
> > You have one example in the readme... how about some tests, to guard against regressions?
> 
> @JalonSolov
> 
> I was told they should be moved to vlang/slower_tests, see [vlang/slower_tests#5](https://github.com/vlang/slower_tests/pull/5)
> 
> I will be adding a few basic test vectors back here as well, maybe validating against the go impl as @tankf33der mentioned

@cestef - I remind you that they should have as much coverage as possible, whatever that means — only you can do this, it is your responsibility. Thank you.

vlib/x/crypto/mldsa/mldsa.v
```diff
@@ -0,0 +1,423 @@
...
+pub struct PrivateKey {
+	seed [32]u8
+	pk   PublicKey
+	s1   []NttElement // len = l
+	s2   []NttElement // len = k
+	t0   []NttElement // len = k
+	k    [32]u8
+}
+
+pub struct PublicKey {
```
blackshirt: Its also should have `@[noinit]`

vlib/crypto/sha3/xof.v
cestef: the `Shake` XOF API intentionally differs from `io.Writer`/`io.Reader` (I initially implemented it like you mentioned), absorption never fails (so `!int` is _kindof_ misleading) and squeezing produces exact byte counts (so `read(n) []u8` makes more sense than preallocating a buffer imo?). Go's `ShakeHash` also never returns errors from write/read, the signatures exist only for interface conformance. No harm in implementing a io.Reader/Writer wrapper though, if that's you meant!

vlib/crypto/sha3/xof.v
cestef: done in [664dde3](https://github.com/vlang/v/pull/26711/commits/664dde38646177e0224e7eb4d318383bab160832)

cestef:
I added a `gen.vsh` script that downloads the golang source at master and adds `gen.go` into it. we need to do this kind of hack because go hasn't yet put their mldsa in their public crypto api.

we currently test against 12 test vecs (4 per kind, 44/65/87), to regenerate test vectors:

```bash
v run vlib/x/crypto/mldsa/testdata/gen.vsh
```

this will pull https://github.com/golang/go on master to `/tmp/go`, build the toolchain, add our gen.go and run it.
test vecs will be at `vlib/x/crypto/mldsa/testdata/vectors.json`, to be then used in `mldsa_test.v`

is this what you had in mind? @tankf33der

JalonSolov:
It doesn't need to be tested against the Go code, it just needs to be tested that it gets the "correct" output.

In other words, hash something, then check that against the _known_ hash value.  Things like that.

Simple unit tests that make sure the code isn't completely broken.

Small tests like that can go in the main repo.

Larger tests that take longer should go to the slow tests repo.

cestef:
> It doesn't need to be tested against the Go code, it just needs to be tested that it gets the "correct" output.

@JalonSolov 
right, we are simply using go's implementation (supposed to be a correct one) to generate those values (once in a while, not in the CI) and compare against them. we also have few roundtrip tests (pk roundtrip, sk roundtrip) and that's it for the unit tests in the main repo.

do you think this is too overkill?

JalonSolov:
Not as long as they don't slow things down too much.  And generating the hashes with Go is completely disconnected.

Technically, Go shouldn't be involved at all, unless we find a problem with the hashes already generated - those should be hard-coded in the tests.

cestef:
CI having a rough day huh

tankf33der:
@cestef I played around with the code and got a lot of emotions out of it. I managed to make the code hang on the signature. If you make the private key fields public and modifiable and simply zero out the first value, the code will hang. This is not a problem, right?

```v
import x.crypto.mldsa

fn main() {
	mut pr := mldsa.PrivateKey.generate(mldsa.Kind.ml_dsa_44)!
	pr.t0[0][0] = 0
	_ := pr.sign([]u8{len: 10_000, init: index})!
}
```

cestef:
@tankf33der yeah that's expected, ML-DSA signing uses rejection sampling. so if you corrupt the key internals it can loop forever. I'll add a max iteration count as a safeguard though, good catch

tankf33der:
@cestef - add tests from `wycheproof` project for mldsa to `slower_tests` too.

cestef:
> @cestef - add tests from `wycheproof` project for mldsa to `slower_tests` too.

see [a0f6035](https://github.com/vlang/slower_tests/pull/5/commits/a0f60357ae7dc6245abd5f90eea0eedaea983366)

blackshirt:
> @cestef I played around with the code and got a lot of emotions out of it. I managed to make the code hang on the signature. If you make the private key fields public and modifiable and simply zero out the first value, the code will hang. This is not a problem, right?
> 
> ```v
> import x.crypto.mldsa
> 
> fn main() {
> 	mut pr := mldsa.PrivateKey.generate(mldsa.Kind.ml_dsa_44)!
> 	pr.t0[0][0] = 0
> 	_ := pr.sign([]u8{len: 10_000, init: index})!
> }
> ```

I think, allowing modifies directly in that fashion was not a good thing

cestef:
good to go on my end. let me know if anything else needs changes

JalonSolov:
Close/re-open again.

tankf33der:
@cestef lets add XOF test vectors for sha3 too to `slower_tests` repo

https://csrc.nist.gov/CSRC/media/Projects/Cryptographic-Algorithm-Validation-Program/documents/sha3/shakebytetestvectors.zip contains variable output vectors.

cestef:
@tankf33der see https://github.com/vlang/slower_tests/pull/5/commits/8ab7e608891408e42858e3f519655ce7742cf079

PR #26714, opened 2026-03-10
checker: resolve @VROOT in #flag pseudo-variable expansion

### LINKED ISSUE #26698: vdoc fails to compile on macOS Sequoia
### Describe the bug

I'm trying to do `v doc` and I'm getting this error
```
cannot compile `/Users/USER/v/cmd/tools/vdoc`: 1
> File not found: @VROOT/thirdparty/md4c-lib{.c,.cpp,.S}
builder error: build_thirdparty_obj_file only support .c, .cpp, and .S source file.
```

### Reproduction Steps

Do `v doc` in Terminal on a system running macOS Sequoia. (?)

note: it apparently works on Tahoe

### Expected Behavior

`vdoc` compiles correctly and begins generating the docs.

### Current Behavior

After hitting enter several seconds pass before the error appears.

### Possible Solution

_No response_

I initially [brought this up in the Vlang discord](https://discord.com/channels/592103645835821068/592114487759470596/1478563320310272027).

### V version

V 0.5.0 5d6de17

macOS Sequoia 15.7.3
> [!NOTE]
>

watzon:
Fix `v doc` failing against stale external tool dependencies.

The original failure came from `vdoc` recompiling against an outdated local
`markdown` module checkout that still used deprecated `@VROOT` paths for its
`md4c` third-party objects. The `markdown` module has already fixed that
upstream, so this patch does not restore `@VROOT` handling in the compiler.

Instead, it makes tool recompilation refresh external module dependencies before
building the tool binary. That allows `v doc` to pick up the current
`vlang/markdown` sources, which already use non-deprecated paths.

Fixes #26698.

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -3155,6 +3155,12 @@ fn (mut c Checker) hash_stmt(mut node ast.HashStmt) {
 
 fn (mut c Checker) resolve_pseudo_variables(oflag string, pos token.Pos) ?string {
 	mut flag := oflag
+	if flag.contains('@VROOT') {
+		flag = util.resolve_vmodroot(flag.replace('@VROOT', '@VMODROOT'), c.file.path) or {
+			c.error(err.msg(), pos)
+			return none
+		}
+	}
 	if flag.contains('@VEXEROOT') {
 		// expand `@VEXEROOT` to its absolute path
 		flag = flag.replace('@VEXEROOT', c.pref.vroot)
--- a/vlib/v/tests/cflags/vmodroot_and_vroot_test.c.v
+++ b/vlib/v/tests/cflags/vmodroot_and_vroot_test.c.v
@@ -11,8 +11,20 @@ fn C.add(i32, i32) i32
 #flag -I @VEXEROOT/thirdparty/stb_image
 #include "stb_image.h"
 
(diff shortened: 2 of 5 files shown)


JalonSolov:
Bouncing to force CI to re-run.

vlib/v/checker/checker.v
```diff
@@ -3155,6 +3155,12 @@ fn (mut c Checker) hash_stmt(mut node ast.HashStmt) {
 
 fn (mut c Checker) resolve_pseudo_variables(oflag string, pos token.Pos) ?string {
 	mut flag := oflag
+	if flag.contains('@VROOT') {
+		flag = util.resolve_vmodroot(flag.replace('@VROOT', '@VMODROOT'), c.file.path) or {
+			c.error(err.msg(), pos)
+			return none
+		}
+	}
```
JalonSolov, line 3163: While this is a good idea, it doesn't solve the problem of source having `@VROOT`, as it only replaces after the source has been read.

Instead, the compiler should give an error message if `@VROOT` is used, and suggest running `v fmt` to fix the problem.

Of course, this also means that `vfmt.v` will need to be modified to make that substitution, if it doesn't already do so.

StunxFS:
Closing this PR. 1e1a0223ac7dd035337b66519957bbf9deda3213.

PR #26734, opened 2026-03-15
v2: parallel pipeline, register allocator, and arm64 self-host fixes

medvednikov:
## Summary
- Parallel transformer, SSA builder, and ARM64 codegen via pthreads with worker clone/merge pattern
- Block-local register allocator (x19-x28 callee-saved pool, 18% binary size reduction)
- ARM64 self-host chain fixes: v2→v3→v4→v5→v6 converged (identical binaries)
- Type layout caches (arrays instead of maps), struct field offset cache, phi copy emission for -O0 mode
- Fix shared map iteration in Environment snapshot functions to avoid ARM64 chained-access bug

## Test plan
- [x] Self-host chain v2→v3→v4→v5→v6 converges (v5=v6 identical at 21,025,988 bytes)
- [x] 13 test suites passing (398+ tests): string, array, map, rune, hex, sorting, math, etc.
- [x] v4-compiled binary successfully compiles hello.v

### DIFF
--- a/vlib/builtin/string.v
+++ b/vlib/builtin/string.v
@@ -84,7 +84,8 @@ pub fn (s string) runes() []rune {
 // It will panic, if the pointer `s` is 0.
 @[unsafe]
 pub fn cstring_to_vstring(const_s &char) string {
-	return unsafe { tos2(byteptr(const_s)) }.clone()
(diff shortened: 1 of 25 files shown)


medvednikov:
This logic is already on master.

PR #26737, opened 2026-03-16
builtin: call the C function wrappers instead of the C functions directly

kimshrier:
builtin/array.v and builtin/map.v were calling C functions directly from a .v file which is now not allowed.

This converts those calls to using the wrapper functions in builtin/cfns_wrapper.c.v.


tankf33der:
@kimshrier - this code works on `master`. What happened and what didn't work that made you come up with creating this PR?

```v
fn main() {
	mut a := []int{len: 8,init: index}
	a[2] = 100

	b := a.sorted_with_compare(|x, y| *x - *y)
	println(b)
}
```

JalonSolov:
V has a special check for `C.` routines in plain `.v` files.  Those files either have to be renamed to `.c.v`, or V equivalent routines/wrappers need to replace the `C.` routines.

The plan has been to make including `C.` stuff in plain `.v` files an error.

Issue #24270, opened 2025-04-19
`v new --web` creates `vweb` projects, not `veb`


JalonSolov:
### Describe the bug

Since `vweb` is deprecated, `v new --web` should not be creating `vweb` projects, only `veb` projects.

### Reproduction Steps

`v new --web` then respond to the prompts.

### Expected Behavior

A `veb` project is created.

### Current Behavior

A `vweb` project is created.

### Possible Solution

Someone who understands the differences between `vweb` and `veb` will need to update the templates `v new` uses.  It is far from a simple matter of changing all `import vweb` to `import veb`... if you do that, the code won't compile at all.

### V version

V 0.4.10 11e25bc

### Environment details (OS name and version, etc.)

Doesn't matter.  This is about the output of `v new --web`.

qibinlin:
I encountered the same problem.

emerikaji:
Is this assigned to anyone yet? I can work on migrating the web model to veb, unless there are specifics currently making this impossible.

JalonSolov:
All PRs are welcome.  I've assigned you @emerikaji - please feel free to make the necessary changes.

emerikaji:
@JalonSolov 

Things are going relatively smooth so far, but there is one catch.

In the original `vweb` project, the view calls on the controller externally, like so:

```v
// product_view
@['/products'; get]
pub fn (app &App) products(mut ctx Context) veb.Result {
    ...
	user := get_user(token)
    ...
}

// user_view_api
pub fn get_user(token string) !User {
    ...
	resp := http.fetch(url: 'http://localhost:8082/controller/user', header: header)!
    ...
}
```

This doesn't seem like an ideal way to do it, but I could have understood from the standpoint of showcasing how a view could access data from an external API.

However, this functionality doesn't work at all now that it's ported to `veb`. The controller routes work as expected when queried from the terminal, for instance; But it seems the `veb` server calling one of the routes it serves itself just blocks out functionality entirely. I can't get an error from the `http.fetch` either, as it never returns.

As I'm not familiar enough yet with the full code of `veb`, my current guess is that it's waiting for the current request (GET /products) to be done before executing the next request (GET /controller/user).

I'm perfectly fine rewriting the code structure to not feature such a call, but I want to check in with you in case this API design serves a specific purpose.

JalonSolov:
@spytheman what do you think?

spytheman:
imho just create a shim, that returns a fake user struct, without doing an additional http request.

[...]

PR #26745, opened 2026-03-17
vcreate: migrate vweb model to veb

emerikaji:
See issue #24270 for context.

This PR migrates the example project used as a demo for the `vweb` library into the `veb` library.

The structure has been slightly altered to follow more consistent web development standards, and align more closely with the code shown in the `vdoc`. However, functionality has been kept identical.

As a side note, it would perhaps be useful in the future to have a better system for the creation of such example projects, rather than pasting the full code into one file. For simplicity, this is fine for now.

### DIFF
--- a/cmd/tools/vcreate/project_model_web.v
+++ b/cmd/tools/vcreate/project_model_web.v
@@ -8,11 +8,12 @@ fn (mut c Create) set_web_project_files() {
 	// v source code
 
 	c.files << ProjectFiles{
-		path: os.join_path(base, "main.v")
+		path:    os.join_path(base, 'main.v')
 		content: "module main
 
 import veb
 import db.sqlite
+import os
 
 struct Context {
 	veb.Context
@@ -29,28 +30,30 @@ pub fn (app &App) index(mut ctx Context) veb.Result {
 }
 
 fn main() {
-	mut db := sqlite.connect('app.db')!
+	mut db := sqlite.connect(os.resource_abs_path('app.db'))!
 	sql db {
 		create table User
 		create table Product
 	}!
 	defer { db.close() or { panic(err) } }
 
-	mut app := &App{ db: db }
-	app.handle_static('static', true)!
+	mut app := &App{
+		db: db
+	}
+	app.handle_static(os.resource_abs_path('static'), true)!
 	veb.run[App, Context](mut app, 8082)
 }
 "
 	}
 	c.files << ProjectFiles{
-		path: os.join_path(base, "user_model.v")
+		path:    os.join_path(base, 'user_model.v')
 		content: "module main
 
 @[table: 'users']
 pub struct User {
-	mut:
-	id       int       @[primary; sql: serial]
-	username string    @[unique]
+mut:
+	id       int    @[primary; sql: serial]
+	username string @[unique]
 	password string
 	active   bool
 	products []Product @[fkey: 'user_id']
@@ -58,7 +61,7 @@ pub struct User {
 "
 	}
 	c.files << ProjectFiles{
-		path: os.join_path(base, "user_view.v")
+		path:    os.join_path(base, 'user_view.v')
 		content: "module main
 
 import veb
@@ -113,7 +116,7 @@ pub fn (app &App) controller_create_user(mut ctx Context) veb.Result {
 "
 	}
 	c.files << ProjectFiles{
-		path: os.join_path(base, "user_controller.v")
+		path:    os.join_path(base, 'user_controller.v')
 		content: "module main
 
 import crypto.bcrypt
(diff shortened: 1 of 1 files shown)


JalonSolov:
Close/re-open for CI.

JalonSolov:
Need to run `v fmt -w` on the files.

emerikaji:
> Need to run `v fmt -w` on the files.

`v fmt` seems to have some weirdly inconsistent behavior in this case.

It will sometimes pick single quotes, sometimes double quotes in the `content` field of the `ProjectFiles`, and swap all escaped quotation marks in the process.
It doesn't seem to be a matter of length (for instance, `products.css` gets changed despite its short length) or type of content, but maybe there's something I'm missing. This behavior was already visible on the original version.

Here's an example, this one is changed:
```vlang
c.files << ProjectFiles{
                path:    os.join_path(base, 'auth_controller.v')
                content: 'module main

import crypto.bcrypt
import crypto.hmac
import encoding.base64
import crypto.sha256
import json

fn (app &App) do_auth(username string, password string) !string {
        users := sql app.db {
                select from User where username == username
        }!
        if users.len == 0 {
                return error(\'user not found\')
        }
        user := users.first()
        if !user.active {
                return error(\'user is not active\')
        }
        bcrypt.compare_hash_and_password(password.bytes(), user.password.bytes()) or {
                return error(\'Failed to auth user, \${err}\')
        }
        return make_token(user)
}

fn (app &App) verify_auth(token string) bool {
        if token == \'\' {
                return false
        }
        secret := \'SECRET_KEY\' // os.getenv(\'SECRET_KEY\')
        token_split := token.split(\'.\')
        signature_mirror := hmac.new(secret.bytes(), \'\${token_split[0]}.\${token_split[1]}\'.bytes(),
                sha256.sum, sha256.block_size).bytestr().bytes()
        signature_from_token := base64.url_decode(token_split[2])
        return hmac.equal(signature_from_token, signature_mirror)
}

fn (app &App) user_auth(token string) !int {
        if !app.verify_auth(token) {
                return error("Invalid token")
        }
        jwt_payload_stringify := base64.url_decode_str(token.split(\'.\')[1])
        jwt_payload := json.decode(JwtPayload, jwt_payload_stringify) or {
                return err
        }
        return jwt_payload.sub.int()
}
'
```

While this one is not:

```vlang
c.files << ProjectFiles{
                path:    os.join_path(base, 'user_controller.v')
                content: "module main

import crypto.bcrypt

fn (app &App) add_user(username string, password string) ! {
        hashed_password := bcrypt.generate_from_password(password.bytes(), bcrypt.min_cost)!
        user_model := User{
                username: username
                password: hashed_password
                active: true
        }
        sql app.db {
                insert user_model into User
        }!
}

fn (app &App) get_users() ![]User {
        return sql app.db {
                select from User
        }!
}

fn (app &App) get_user(id int) !User {
        results := sql app.db {
                select from User where id == id
        }!
        if results.len == 0 {
                return error('no results')
        }
        return results[0]
}
"
```

JalonSolov:
That first example does look odd.

`v fmt` is supposed to examine strings, and if they have `"` in them, change the outer set to `'`, or if they have `'` in them, change the outer set to `"`.

It should not create strings with `'` as the outer set, then `\'` inside the string.  The `\$` bits are correct regardless of outer quotes.

Can you manually modify the first example to have `"` as the outer set, and remove all `\` inside the string unless it is before `$`?  And if you do, does it still change things?

emerikaji:
That's actually what I did prior to running `v fmt` again and sending the examples in question. I ran it once using `v fmt -w`, noticed the quotation mark swaps, corrected them, and ran it again without overwrite this time to check if it behaved in the same way, which it did.

My current assessment is that the clause dictating, if there are `"` in the string then use `'` as the outer, takes priority over the other clause. I believe this is the case as `v fmt` defaults to `'` when no quotation marks are used in the string. This isn't bad behavior, but there should definitely be a way for `v fmt` to check if there are more `'` or `"` in a given string and choose the outer accordingly to avoid such cases.

I will be pushing the changes soon. This was my fault as this first example had a string with double quotes whereas `v fmt` employs single quotes.

As a side note, would it be of help if I developed a simple `v` app/script to convert a `v` project into a model file like this one?

JalonSolov:
That's basically what `v create` does... build a project from model files.  Those files just happen to be strings in the code.

One change you could make is to extract those strings into separate files (each template project as a separate subdir, perhaps), and use `$embed_file()` to make them part of the final executable.

That way, you could format them once, then not have to worry about them again, unless the template files themselves change.

JalonSolov:
Merging what's here... can always do a new PR for changes.

PR #26776, opened 2026-03-26
feat: add HTTP/2, HTTP/3, and QUIC protocol support

jupilhwang:
Implement complete HTTP/2 (RFC 7540/7541), HTTP/3 (RFC 9114/9204), and QUIC (RFC 9000/9001) protocol support for the V standard library.

HTTP/2 (net.http.v2):
- HPACK header compression with Huffman coding and O(1) static table
- All 10 frame types with padding, CONTINUATION flood protection (CVE-2024-27316)
- Stream multiplexing, flow control (bidirectional), stream state machine
- TLS (h2) and plain TCP (h2c) server modes with h2c upgrade mechanism
- Connection pooling, CONNECT tunneling, GREASE, cookie compression
- Request/response validation per RFC 7540 Section 8

HTTP/3 (net.http.v3):
- QPACK header compression with ring buffer dynamic table and blocked stream queueing
- 17 H3 error codes, control/encoder/decoder unidirectional streams
- 2-phase GOAWAY graceful shutdown, background control stream reader
- Alt-Svc discovery and caching, GREASE support
- Request validation, header lowercase enforcement per RFC 9114

QUIC (net.quic):
- ngtcp2 C bindings with TLS 1.3 crypto (AES-128-GCM, HKDF, header protection)
- Connection migration with PATH_CHALLENGE/RESPONSE and NAT rebinding
- 0-RTT session resumption with anti-replay cache and ticket extraction
- CONNECTION_CLOSE frames, idle timeout monitoring
- CID-based packet matching, flow control exposure

Integration (net.http):
- Version negotiation with automatic HTTP/2/3 selection
- ALPN get_alpn_selected() added to both mbedtls and OpenSSL backends
- Alt-Svc header parsing and HTTP/3 endpoint discovery
- 421 Misdirected Request handling

Security:
- CONTINUATION flood protection, max header/body size limits
- Connection count limits, forbidden cipher blacklist
- Thread-safe flow control, pools, caches with sync.Mutex
- Never-indexed HPACK encoding for sensitive headers
- Single-allocation AEAD encryption (zero-copy)

Tests: 37 test files, all passing (19 HTTP/2 + 12 HTTP/3 + 5 QUIC + 1 Alt-Svc)

External dependencies: ngtcp2, ngtcp2_crypto_ossl, OpenSSL 3.x


changrui:
Maybe module `net.http.http2`, module `net.http.http3` and module `net.http.quic`.

JalonSolov:
Personally, I'd rather keep the single `net.http`, but the code negotiates with the other end of the connection as to the best style to use.

That way I don't write something that does `import net.http.quic` and then fails to work with a lot of sites.  I also don't need 4 different imports to make a connection work.

jupilhwang:
I am currently modifying it so that http, http2, and http3 can all be used with a single net.http import.

Jengro777:
How's that going?

jupilhwang:
> How's that going?

I already committed the fix for single import.

Jengro777:
@JalonSolov

JalonSolov:
Look at the list of conflicts that need to be resolved.  Can't merge until those are fixed.

Jengro777:
> Look at the list of conflicts that need to be resolved.  Can't merge until those are fixed.


Recently, a large number of submissions have indeed put some pressure on merging new PRs.

medvednikov:
1. P0: net.http does not compile
     /tmp/v-pr-26776/vlib/net/http/request.v:314 calls req.ssl_do(nport, method, host_name,
     path) with 4 args, but ssl_do now requires 6: (port, method, host, path, data, header).
     This breaks any program importing net.http, plus request_test.v, alt_svc_test.v,
     examples, and CI vpm builds.
  2. P1: import net.http now requires QUIC/ngtcp2 native libraries
     /tmp/v-pr-26776/vlib/net/http/request_version.v:7 imports net.http.v3, which imports
     net.quic; /tmp/v-pr-26776/vlib/net/quic/ngtcp2.c.v:11 unconditionally links -lngtcp2,
     -lngtcp2_crypto_ossl, -lssl, -lcrypto.
     That makes HTTP/1.1-only users depend on external HTTP/3 native deps. This is a broad
     stdlib regression.
  3. P1: HTTP/3/QUIC receive path cannot deliver response/request bodies correctly
     /tmp/v-pr-26776/vlib/net/quic/quic_stubs.c:120 receives stream data but explicitly
     ignores data and datalen; /tmp/v-pr-26776/vlib/net/quic/connection_io.v:163 then returns
     stream.data, which is never populated from the peer.
     Separately, /tmp/v-pr-26776/vlib/net/http/v3/server_packet.v:51 decrypts raw UDP packets
     and parses them as HTTP/3 frames, but HTTP/3 frames live inside QUIC stream frames. This
     likely passes unit tests while failing real interop.
  4. P1: http.Server.handler is a public API break
     /tmp/v-pr-26776/vlib/net/http/server.v:23 changes Server.handler from the old Handler
     interface to ServerHandler. The old adapter exists in /tmp/v-pr-26776/vlib/net/http/
     server_adapters.v:20, but existing code using http.Server{ handler: MyHandler{} } no
     longer compiles unless manually wrapped.
  5. P2: request-line parsing regresses double-slash paths
     /tmp/v-pr-26776/vlib/net/http/request.v:981 changed request-target parsing from
     parse_request_uri to parse. For GET //another.html HTTP/1.1, this now treats
     another.html as the host instead of preserving path //another.html. The PR also removes
     the tests that covered this case.
  6. P2: header override logic uses the wrong header source
     /tmp/v-pr-26776/vlib/net/http/request.v:331 accepts a per-attempt header argument, but
     default suppression checks still read req.header. Redirect handling mutates local
     method, data, and header, so this can produce duplicate or stale Host, User-Agent, and
     Content-Length.
  7. P2: a new HTTP/2 test file is stale and does not compile
     /tmp/v-pr-26776/vlib/net/http/v2/integration_full_test.v:57 initializes
     ServerResponse{ headers: ... }, but the struct field is header. ./vnew -gc none vlib/
     net/http/v2/integration_full_test.v fails with four unknown field headers errors.
  8. P2: TLS verification/mTLS settings are not forwarded into the HTTP/2 path
     /tmp/v-pr-26776/vlib/net/http/request_version.v:64 builds a fresh v2 client from the
     address only. /tmp/v-pr-26776/vlib/net/http/v2/client.v:23 creates TLS with only ALPN.
     Existing http.Request settings like verify, cert, cert_key, and in-memory verification
     are ignored.

medvednikov:
Thanks. Conflicts now.

medvednikov:
Code is not vfmt'ed

JalonSolov:
Another conflict to merge, now.

Also, if you run `v git-fmt-hook install` in your repo, any V files you modify will be formatted when you commit, so you will never run into that again.

medvednikov:
Still not vfmt'ed

Jengro777:
v git-fmt-hook install

medvednikov:
**Won't compile.** In `vlib/net/http/request.v` the new `method_and_url_to_response` removed `data`/`header` params, but the proxy branch (line 160 of the patch) still references `data` and `header`, and `build_request_headers` calls `build_request_cookies_header_with_header(header)` with no such param in scope. `ssl_do` arity is wrong as the maintainer noted.

**Protocol layering is wrong.** `v3/server_packet.v` decrypts UDP packets and runs `decode_varint` straight over the plaintext as HTTP/3 frames. HTTP/3 frames live inside QUIC STREAM frames; you can't parse them out of a raw packet payload. This would never interoperate with a real HTTP/3 peer — it likely only passes its own unit tests.

**Forces ngtcp2 + OpenSSL on every `net.http` user.** Even an HTTP/1.1-only program now drags in `net.quic` via `request_version.v` → `net.http.v3` → `#flag -lngtcp2 -lngtcp2_crypto_ossl -lssl -lcrypto`. The `_d_use_ngtcp2` files suggest a half-finished attempt to gate this; verify the import graph is actually conditional.

**Breaks public API.** `http.Server.handler`'s type changed; existing user code won't compile without manual wrapping.

**Scope is unreviewable.** 28,411 additions / 171 files / a from-scratch HPACK + QPACK + Huffman + QUIC binding + AEAD + 0-RTT + migration + ALPN + Alt-Svc — all in one PR. Even if the bugs were fixed, this should be 5–10 PRs (frame parser, HPACK, h2 client, h2 server, ALPN/Alt-Svc, QUIC bindings, HTTP/3 client, HTTP/3 server). It's not vfmt'ed, has stale tests (`ServerResponse{ headers: ... }` — field is `header`), and has been collecting merge conflicts since it sits idle.

The author clearly put serious effort in — RFC references are real, security threats (CONTINUATION flood / CVE-2024-27316) are considered, tests exist. But it has correctness bugs at the protocol layer, doesn't compile as-is, breaks the stdlib for everyone, and is too large to land safely. It needs to be split up and rewritten with the HTTP/3 frame dispatch fixed before it's worth re-reviewing.

medvednikov:
Let's split this up into several smaller PRs. Will be easier to review, test, and merge. 30k loc is too much for a single PR.

PR #26790, opened 2026-03-30
db.sqlite: add cols field to Row for column name access

quaesitor-scientiam:
## Summary
- Adds `cols []string` to `pub struct Row` in `vlib/db/sqlite/sqlite.c.v`
- Populates `cols` via the already-bound `C.sqlite3_column_name` in `exec()` and `exec_param_many()`
- Backwards-compatible — existing code reading `.vals` is unaffected
- Enables callers to access column names alongside values without separate queries or SQL parsing

## Motivation
`Row` currently only exposes `vals []string`, so callers needing column names must parse SQL or run `PRAGMA table_info` separately. The C API already provides column names at the statement level via `sqlite3_column_name` (already declared in `sqlite.c.v` line 115 and used in `exec_map`). This change makes the same information available through the standard `exec()` path.

## Test plan
- [x] New `test_exec_returns_column_names` covers `exec`, `exec_one`, and `exec_param`
- [x] All existing `sqlite_test.v` tests pass unchanged
- [x] `./vnew -d present_sqlite3 test vlib/db/sqlite/sqlite_test.v` — 1 passed, 0 failed
- [x] `./vnew fmt -verify` clean on both touched files

### DIFF
--- a/vlib/db/sqlite/sqlite.c.v
+++ b/vlib/db/sqlite/sqlite.c.v
@@ -88,6 +88,7 @@ pub fn (db &DB) str() string {
 
 pub struct Row {
 pub mut:
+	cols []string
 	vals []string
 }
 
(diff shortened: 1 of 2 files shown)


quaesitor-scientiam:
Closing — column names are already available via exec_map(). Not needed.

PR #26805, opened 2026-04-02
repl: mkdir temp folder before chdir to avoid startup failure

### LINKED ISSUE #26804: repl: mkdir temp folder before chdir to avoid startup failure
## Description

The V REPL fails on startup when the temp directory does not exist yet, printing:

```
V repl error: Could not change the REPL working folder to `<TEMP>/v_0/repl`: No such file or directory; code: 2
```

## Root Cause

In `cmd/tools/vrepl.v`, `run_repl` calls `os.chdir(workdir)` before `new_repl(workdir)`. The `os.mkdir_all` that creates the folder lives inside `new_repl`, so it is never reached when `chdir` fails first.

```v
fn run_repl(workdir string, vrepl_prefix string) int {
    os.chdir(workdir) or { // fails — folder doesn't exist yet
        rerror('Could not change the REPL working folder ...')
        return 1
    }
    ...
    mut r := new_repl(workdir) // os.mkdir_all is here — never reached
```

## Fix

Add `os.mkdir_all(workdir) or {}` immediately before `os.chdir` in `run_repl`.

## Steps to Reproduce

1. Set `TEMP`/`TMP` to a path whose subdirectory `v_0/repl` does not exist (e.g. a freshly cleared temp folder).
2. Run `v` (the REPL).

## Expected Behavior

The REPL creates the temp folder and starts normally.

## Actual Behavior

```
V repl error: Could not change the REPL working folder to `S:\temp\v_0\repl`: No such file or directory; code: 2
```

> [!NOTE]
>

quaesitor-scientiam:
## Summary

- `run_repl` called `os.chdir(workdir)` before `new_repl(workdir)`, so the `os.mkdir_all` inside `new_repl` was never reached when the temp directory did not exist
- Fix: add `os.mkdir_all(workdir) or {}` before `os.chdir` in `run_repl`
- Reproduces whenever `$TEMP/v_0/repl` does not exist (e.g. freshly cleared temp folder)

Fixes #26804

## Test plan

- [ ] Run `v` with a cleared temp folder — REPL starts without error
- [ ] `./vnew -silent vlib/v/slow_tests/repl/repl_test.v`

### DIFF
--- a/cmd/tools/vrepl.v
+++ b/cmd/tools/vrepl.v
@@ -458,6 +458,7 @@ fn remove_comment(oline string) string {
 }
 
 fn run_repl(workdir string, vrepl_prefix string) int {
+	os.mkdir_all(workdir) or {}
 	os.chdir(workdir) or {
 		rerror('Could not change the REPL working folder to `${workdir}`: ${err}')
 		return 1
--- a/cmd/tools/vrepl.v
+++ b/cmd/tools/vrepl.v
@@ -458,7 +458,10 @@ fn remove_comment(oline string) string {
 }
 
 fn run_repl(workdir string, vrepl_prefix string) int {
-	os.mkdir_all(workdir) or {}
+	os.mkdir_all(workdir) or {
+		rerror('Could not create the REPL working folder `${workdir}`: ${err}')
+		return 1
+	}
 	os.chdir(workdir) or {
 		rerror('Could not change the REPL working folder to `${workdir}`: ${err}')
 		return 1


cmd/tools/vrepl.v
```diff
@@ -458,6 +458,7 @@ fn remove_comment(oline string) string {
 }
 
 fn run_repl(workdir string, vrepl_prefix string) int {
+	os.mkdir_all(workdir) or {}
```
JalonSolov: If it can't make a dir it needs, it should error out, rather than ignoring it and moving on to a later failure.

Perhaps add an `eprintln` in the `or` block, then `exit(1)`

quaesitor-scientiam (commented):
Updated per feedback — replaced  with  +  to match the existing error-handling style in .

quaesitor-scientiam:
@JalonSolov  is something missing?

JalonSolov:
Nope.  Just waiting for CI.  The remaining problems will be dealt with soon, though.

PR #26809, opened 2026-04-02
parser: fix static generic calls with reference types

wenxuanjun:
Treat `Type.method[&T]()` as a generic static call instead of misparsing it as selector/index syntax.

This fixes cases where reference concrete types are used in static generic calls.


JalonSolov:
Conflicts to resolve.

PR #26817, opened 2026-04-04
sqlite: use bundled amalgamation on Linux/macOS when pkg-config is unavailable

### LINKED ISSUE #26830: sqlite: cannot compile on Linux/macOS without system libsqlite3 when pkg-config is unavailable
## Description

When `pkg-config sqlite3` fails (e.g. `libsqlite3-dev` is not installed), the SQLite module falls back to `#flag -lsqlite3`, which also requires a system development package. This means users who have run `vlib/db/sqlite/install_thirdparty_sqlite.vsh` to download the bundled amalgamation still cannot compile `db.sqlite` without a system package.

## Relationship to #26724

PR #26724 (merged) made a partial fix to this area:
- Moved `-I@VEXEROOT/thirdparty/sqlite` outside the `$if windows` block so all platforms can find `sqlite3.h` from the thirdparty directory
- Added a `$pkgconfig('sqlite3')` check as the preferred path
- But **deliberately kept** `-lsqlite3` as the non-Windows fallback

This means after #26724, the header is found correctly — but **linking still fails** on systems where only `libsqlite3.so.0` is present (no dev package, no `.so` symlink for the linker).

This issue tracks the remaining linker half of the problem.

## Steps to reproduce

1. On a Linux system without `libsqlite3-dev` installed (only the runtime `.so.0` present, no `.so` symlink)
2. Run `vlib/db/sqlite/install_thirdparty_sqlite.vsh` to download the bundled amalgamation
3. Compile any program using `db.sqlite`

**Result:**
```
cc: tcc: error: library 'sqlite3' not found
```

## Expected behavior

The bundled amalgamation (`thirdparty/sqlite/sqlite3.c`) downloaded by `install_thirdparty_sqlite.vsh` should be compiled directly when pkg-config is unavailable, as it already is on Windows. Both the header and link steps should work without any system package.

## Fix

PR #26817 completes what #26724 started — it replaces `#flag -lsqlite3` with `#flag @VEXEROOT/thirdparty/sqlite/sqlite3.c` in the non-Windows, no-pkg-config fallback branch, mirroring the Windows behaviour.

Verified on WSL2 (Ubuntu, no `libsqlite3-dev`): fails on master, passes with the fix applied.

quaesitor-scientiam:
## Summary

- When `pkg-config sqlite3` fails on Linux/macOS, the module previously fell back to `-lsqlite3`, which requires a system development package to be installed
- This PR replaces that fallback with `#flag @VEXEROOT/thirdparty/sqlite/sqlite3.c`, compiling the bundled amalgamation directly — consistent with what already works on Windows
- Users without a system SQLite package can run `vlib/db/sqlite/install_thirdparty_sqlite.vsh` to obtain the amalgamation

## Test plan

- [ ] Run `vlib/db/sqlite/install_thirdparty_sqlite.vsh` on a Linux system without `libsqlite3-dev` installed
- [ ] Verify `v run vlib/db/sqlite/sqlite_test.v` passes

Fixes #26830

### DIFF
--- a/vlib/db/sqlite/sqlite.c.v
+++ b/vlib/db/sqlite/sqlite.c.v
@@ -11,7 +11,7 @@ $if $pkgconfig('sqlite3') {
 	#flag windows -L@VEXEROOT/thirdparty/sqlite
 	#flag windows @VEXEROOT/thirdparty/sqlite/sqlite3.o
 } $else {
-	#flag -lsqlite3
+	#flag @VEXEROOT/thirdparty/sqlite/sqlite3.c
 }
 #include "sqlite3.h" # The SQLite header file is missing. Please run vlib/db/sqlite/install_thirdparty_sqlite.vsh to download an SQLite amalgamation, or install its development package.
 


quaesitor-scientiam:
All the failures are **unrelated to this change**. The common failure across all Linux CI jobs is:

```
native error: vlib/v/gen/native/gen.v:1218, &native.Gen{}.patch_calls fn addr of `__at_least_one` = 0
```

This is a pre-existing bug in the native backend (`-b native`) affecting `vlib/v/gen/native/tests/native_test.v`. The CI installs `libsqlite3-dev` itself, so it hits the `pkg-config` branch and this change is never even exercised.

The **mssql** and **mysql** failures are also unrelated — the mssql log shows the SQL Server container initializing (a timeout/infrastructure issue).

**The docker-alpine-musl** jobs are likely also the same native backend issue. This change is clean — the failures are pre-existing on master.

PR #26890, opened 2026-04-18
veb: fix static handler

squidink7:
Fixes static file handling in veb. Currently static files are incorrectly handled and return 404 when accessed, this PR fixes that.
Note: this only fixes static file handling when compiled with tcc. When using GCC there is a segfault in the parallel fasthttp backend and clang doesn't compile at all, giving a C error related to mismatched pointer types. I believe these are other issues however, and the bugs are likely in the V compiler rather than veb.


Jengro777:
#https://github.com/vlang/v/commit/8aca2fb26fcee0aedd620ae9bf38c7eca23adacd

PR #26925, opened 2026-04-20
x.markdown: add pure V markdown module

JalonSolov:


SheatNoisette:
Hmm, shouldn't be moved to https://github.com/vlang/markdown ?

JalonSolov:
It's used by an internal tool (`v doc`), so seems like a good fit for internal module rather than external.  🤷🏼‍♂️

PR #26926, opened 2026-04-20
x.bzip2: add pure V bzip module

JalonSolov:


squidink7:
Now V has optional parameters, should the *_with_params functions just use an optional flag?
e.g.
```
fn decompress(src []u8, verify_crc ?bool)
```
It depends if you think there may be more parameters to add in the future though, but with just one then it may make more sense as an option.

JalonSolov:
Actually, yes, it would make sense to only have `compress` and `decompress` since the params are optional... trailing literal struct syntax.

JalonSolov:
The README examples have also been adjusted.  There's no need for `verify_crc ?bool` as the `DecompressParams` struct takes care of it, in a more V way.

vlib/x/compress/bzip2/bzip2.v
```diff
@@ -0,0 +1,1079 @@
...
+	}
+	return w.out
+}
+
+struct BitReader {
+	data []u8
+mut:
+	byte_pos  int
+	bit_pos   int
+	last_bits int
```
SheatNoisette: This variable is only assigned once and never read (line 1039)

vlib/x/compress/bzip2/bzip2.v
```diff
@@ -0,0 +1,1079 @@
...
+fn rle1_encode(src []u8) []u8 {
+	if src.len == 0 {
+		return []u8{}
+	}
+	mut out := []u8{cap: src.len + (src.len / 4) + 8}
+	mut i := 0
+	for i < src.len {
+		b := src[i]
+		mut run := 1
+		for i + run < src.len && src[i + run] == b && run < 0x7fffffff {
```
SheatNoisette: The magic number 0x7ffffff for max 32bits int should be made into a const

vlib/x/compress/bzip2/bzip2.v
```diff
@@ -0,0 +1,1079 @@
...
+	selector_mtf := encode_selector_mtf(selectors, n_groups)
+	freq := symbol_freq(mtf_symbols, alpha_size)
+	lens := make_huffman_lengths(freq, bzip2_max_code_len)
+	codes := build_huffman_codes(lens)
+	mut code_lengths := [][]int{len: n_groups}
+	mut code_words := [][]u32{len: n_groups}
+	for i in 0 .. n_groups {
+		code_lengths[i] = lens.clone()
+		code_words[i] = codes.clone()
+	}
```
SheatNoisette, line 388: Real bzip2 partitions the symbol stream into groups of 50 symbols each, clusters groups with similar frequency distributions together and builds per-group Huffman tables.

vlib/x/compress/bzip2/bzip2.v
```diff
@@ -0,0 +1,1079 @@
...
+		for _ in 0 .. 8 {
+			if (crc & u32(0x80000000)) != 0 {
+				crc = (crc << 1) ^ bzip2_crc_poly
+			} else {
+				crc <<= 1
+			}
+		}
+	}
+	return ~crc
+}
```
SheatNoisette, line 1099: whining: Using a CRC table will make the function dramatically faster for big archives

vlib/x/compress/bzip2/bzip2.v
```diff
@@ -0,0 +1,1079 @@
...
+	for s in symbols {
+		if s >= 0 && s < alpha_size {
+			freq[s]++
+		}
+	}
+	for i in 0 .. alpha_size {
+		if freq[i] == 0 {
+			freq[i] = 1
+		}
+	}
```
SheatNoisette, line 450: This forces Huffman codes for symbols that never appear, wasting code space and bits. From what I read bzip2 only assigns codes to symbols present in each group.

SheatNoisette (commented):
Pretty good first implementation of bzip2. Largely good enough.

I would have preferred more comments on the implementation with any reference used and, if I didn't missed it, a test with a real bzip archive from disk with something in it. I skipped most of the optimization possible.

JalonSolov:
> Pretty good first implementation of bzip2. Largely good enough.
> 
> I would have preferred more comments on the implementation with any reference used and, if I didn't missed it, a test with a real bzip archive from disk with something in it. I skipped most of the optimization possible.

The code was tested against both the Linux command line `bzip2` and the Python `bz2` package, going both ways... having this code create a file and the others decode it, and having the others create files and this code decode it, at various levels from 1 to 9.  Wound up being a matrix of 90 tests - all passed.

However, including a test like that might be good in the CI, but won't work for everyone unless they have `bzip2` and Python's `bz2` installed locally.

I suppose I could still include the test, but as something like a `.vsh` file in a subdir... don't want `v test` to run it automatically.

vlib/x/compress/bzip2/bzip2.v
JalonSolov: Didn't seem to matter when comparing against command line bzip2 or Python bz2 package.

Wouldn't hurt to see if there's a performance difference, though.

vlib/x/compress/bzip2/bzip2.v
JalonSolov: True.  For big files (as bzip2 is a compressor, not an archive type :-) ).

JalonSolov:
... and I was following the "Make it work, first, _then_ optimize." rule.  No reason some optimizations can't be added up front, though.

vlib/x/compress/bzip2/bzip2.v
JalonSolov: I have not been able to find anything about partitioning the symbol stream.  If you have a link, I'd like to see it.

vlib/x/compress/bzip2/bzip2.v
JalonSolov: Debugging leftover...

vlib/x/compress/bzip2/bzip2.v
JalonSolov: Basically, it's for safety.  I asked the AI what purpose those lines serve (since it writes better than I), and here's what it said:

---

After counting how often each symbol appears in the stream, any symbol with a count of 0 is bumped to 1. This ensures every symbol in the alphabet gets at least a minimal weight when the Huffman tree is built.

Without this guard:

* A symbol with `freq = 0` would get no Huffman code assigned.
* If the decoder later encounters that symbol (e.g., due to a selector assigning it), decoding would fail or produce invalid results.
* The Huffman length builder could also produce degenerate trees if some symbols are entirely absent.

By setting unseen symbols to `1`, the compressor guarantees the generated code table is complete across the full alphabet (`alpha_size` entries), keeping the encoder/decoder tables consistent.

---

It may be a tiny performance hit to keep it, but better to avoid degenerate trees, etc.

JalonSolov:
I added an `interop` subdir with a `README` and a `.vsh` file.  It runs this code against `bzip2` command and Python `bz2`, at compression levels 1, 6, and 9.  A matrix of 135 tests, which can _only_ succeed if results are byte-for-byte identical.

The code currently passes these tests.

JalonSolov:
$%#@$%^ git... I'll try this again after Alex's commits settle.

PR #26983, opened 2026-04-25
support comptime dollar if bsd test

kimshrier:
I saw with commit 10e413a34cae26c7f6c53e6fcdaf7ef5c1511b1c that there was bsd environment specific file support.

This update adds the capability to do `$if bsd` tests in the same way one could test for a specific os.


medvednikov:
Thanks!

PR #26990, opened 2026-04-25
eval: support string concatenation with the + operator

kimshrier:
This fix addresses `invalid operands to +: string and string` errors
that I see when running `v test-all`.

The real fix is to fix the generator for `vlib/v/eval/infix.v`.  If I understand
things right, I saw that fixes for `&&` and `||` had been applied after 
`vlib/v/eval/infix.v` had been generated.  This fix follows that pattern.

I will look at doing the generator fix.


medvednikov:
I've just pushed
`all: remove v.eval (replaced by more complete v2.eval)`

so this is no longer needed

PR #27008, opened 2026-04-27
net.s3, net.http: add an S3-compatible client and a scheme-handler dispatch in http.fetch

davlgd:
S3 is one of the most widely deployed cloud APIs across the major providers and self-hosted object stores. Having a first-class client in V, with no third-party dependencies and no need to shell out to aws-cli, makes the language a credible choice for cloud workloads.

I am proposing this module for native integration under `vlib/net/s3` because of how common the stack is, but please tell me if you would rather see it land under `vlib/x/` first. The module is self-contained either way.

`vlib/net/s3` is an S3-compatible client with no third-party dependencies. Signing is built on `crypto.hmac` and `crypto.sha256` only, and is verified against the AWS-published Signature V4 reference vectors.

It exposes single-shot ops (`put`, `get`, `get_string`, `stat`, `exists`, `size`, `delete`, `presign`, `list`), bucket ops (`create_bucket`, `delete_bucket`, `bucket_exists` with client-side validation of S3 bucket-naming rules), and multipart upload. The multipart layer covers `upload_file` (auto single/multipart), `upload_file_multipart`, `upload_bytes_multipart`, plus a streaming `MultipartUploader` for chunks generated on the fly, with up to `Client.queue_size` parallel part uploads with per-part retry and exponential back-off.

A small File handle (`Client.file(key)` returning a reference type that forwards to the client via `f.read()`, `f.write()`, `f.presign()`, and friends) is provided for ergonomic call sites. `s3.fetch('s3://bucket/key', ...)` acts as a direct entry point, plus an init() registers the `s3://` scheme with net.http so `http.fetch(url: 's3://...')` works once the module is imported.

`Credentials.from_env()` resolves credentials with first-non-empty resolution across S3_*, AWS_*, CELLAR_ADDON_*, SCW_*, B2_*, R2_*, SPACES_* so the same client targets the major hosted and self-hosted S3 providers with no reconfiguration.

A short usage example:

```v
import net.s3
c := s3.new_client(s3.Credentials.from_env())
c.put('hello.txt', 'Hi from V!'.bytes())!
text := c.get_string('hello.txt')!
url := c.presign('hello.txt', expires_in: 3600)!
```

`vlib/net/http/http.v` adds a tiny scheme-handler registry to let modules like `net.s3` plug into `http.fetch` for non-HTTP schemes. `register_scheme(scheme, handler)` and `unregister_scheme(scheme)` let modules claim a scheme from their own `init()`. `fetch()` looks up a handler for any non-empty, non-HTTP/HTTPS scheme and delegates if one is registered. Otherwise the call path is byte-for-byte identical to master, with no behaviour change for plain HTTP calls. The new `__global scheme_handlers` map is opt-in via `@[has_globals]` on the module, per V's standard convention.

On the architecture side, `client.v` holds the Client struct, single-shot CRUD and the `do_http core`. `multipart.v` holds the multipart upload, the parallel worker pool and the MultipartUploader. `bucket.v` handles bucket lifecycle and bucket-name validation. `signer.v` is pure SigV4 (no HTTP coupling) and exposes `sign_request` and `presign_url`. `credentials.v` holds the Credentials struct, env merging and region resolution. `encoding.v` has URI percent-encoding, hex helpers and CRLF guards. `errors.v` defines S3Error and parses the XML <Error> envelope. `fetch.v` implements the `s3.fetch('s3://...')` URL helper. `http_bridge.v` registers the `s3://` scheme with `net.http`. `file.v` exposes the File handle bound to a Client and a key. `list.v` parses ListObjectsV2 XML. `types.v` gathers the shared option and response structs (Acl, StorageClass and friends).

The signer is intentionally decoupled from `net.http`: it is a pure function that takes credentials and a SignRequest and returns the headers. This keeps it testable against the published SigV4 vectors and reusable outside the HTTP client.

On the security side, every user-supplied value going into a signed header or signed query param is screened for CR/LF before signing (InvalidHeader, InvalidQueryParam, InvalidCredentials). The signer rejects anything outside `GET`, `PUT`, `POST`, `DELETE` and `HEAD`. `s3.fetch` only accepts `s3://` URLs, `https://`, `http://`, `file://` and friends are rejected up-front to prevent accidental cross-protocol calls with S3 credentials. The default scheme is HTTPS whenever endpoint does not carry an explicit `http://` prefix and `insecure_http` is not set. When compiled with `-d s3_debug`, request headers are dumped but Authorization and `X-Amz-Security-Token` values are masked. `redact_url` strips query strings before logging so presigned URLs never leak signature material into error text.

The unit suite covers 8 files and runs offline by default. `v test vlib/net/s3/` reports 8 passed, 8 total. The tests cover signer vectors (AWS reference, canonical request, HMAC chain cross-checked with openssl), encoding, error parsing, credentials env merging, bucket-name validation, list-response parsing, fetch URL parsing, redaction and CRLF rejection.

Integration tests live in `integration_test.v`, gated on `S3_INTEGRATION=1`. They exercise put/get round-trip, range read, presign GET, list with prefix, exists/delete, `s3.fetch` and the `http.fetch('s3://...')` bridge, multipart (parallel, with out-of-order completion), bucket lifecycle and an invalid-credentials cleanup-error path. Example invocation:

```shell
S3_INTEGRATION=1 \
S3_HOST=https://s3.example.com \
S3_KEY_ID=... S3_KEY_SECRET=... \
S3_BUCKET=v-s3-tests \
v test vlib/net/s3/integration_test.v
```

The module has been verified against AWS S3 SigV4 reference vectors (byte-exact Authorization header matches in signer_test.v), against a local self-hosted S3-compatible server in debug and -prod builds (full integration suite green), and against a hosted S3-compatible production endpoint (full integration suite green).

This work builds on top of #27000 (HEAD, 1xx, 204 and 304 responses on Content-Length-bearing connections), which had to land first for stat, exists and range responses to behave correctly under V's HTTP client.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,9 @@
 !vlib/mcp/server.v
 !vlib/mcp/server_test.v
 
+# unignore vlib/net/s3
+!vlib/net/s3/**
+
 !.github/show_manual_release_cmd_test.v
 !*/
 # Do not add !*.* here; it overrides local excludes from .git/info/exclude.
--- a/vlib/net/http/http.v
+++ b/vlib/net/http/http.v
@@ -1,6 +1,7 @@
 // Copyright (c) 2019-2024 Alexander Medvednikov. All rights reserved.
 // Use of this source code is governed by an MIT license
 // that can be found in the LICENSE file.
+@[has_globals]
 module http
 
 import net.urllib
@@ -197,10 +198,49 @@ pub fn prepare(config FetchConfig) !Request {
 	return req
 }
 
+// SchemeHandlerFn dispatches a `fetch()` call for a non-HTTP scheme. Used by
+// out-of-tree-friendly modules like `net.s3` to register themselves at init
+// time without forcing `net.http` to know about them statically.
+pub type SchemeHandlerFn = fn (config FetchConfig) !Response
+
+__global scheme_handlers = map[string]SchemeHandlerFn{}
+
+// register_scheme attaches `handler` as the dispatcher for URLs with the
+// given `scheme` (e.g. `'s3'`). Handlers are looked up by `fetch()` before
+// the native HTTP path runs. Modules typically call this from `init()`.
+pub fn register_scheme(scheme string, handler SchemeHandlerFn) {
+	scheme_handlers[scheme] = handler
+}
+
+// unregister_scheme removes a previously-registered scheme handler. Mostly
+// useful in tests that install a temporary handler.
+pub fn unregister_scheme(scheme string) {
+	scheme_handlers.delete(scheme)
+}
+
+fn scheme_of(url string) string {
+	colon := url.index(':') or { return '' }
+	if colon == 0 {
+		return ''
+	}
+	return url[..colon]
+}
+
 // TODO: @[noinline] attribute is used for temporary fix the 'get_text()' intermittent segfault / nil value when compiling with GCC 13.2.x and -prod option ( Issue #20506 )
 // fetch sends an HTTP request to the `url` with the given method and configuration.
+// When `config.url` uses a scheme registered via `register_scheme` (e.g.
+// `s3://`), the call is delegated to that handler instead of the native
+// HTTP path.
 @[noinline]
 pub fn fetch(config FetchConfig) !Response {
+	if scheme_handlers.len > 0 {
+		scheme := scheme_of(config.url)
+		if scheme != '' && scheme != 'http' && scheme != 'https' {
+			if h := scheme_handlers[scheme] {
+				return h(config)
+			}
+		}
+	}
 	req := prepare(config)!
 	return req.do()!
 }
--- a/vlib/net/s3/README.md
+++ b/vlib/net/s3/README.md
@@ -1,0 +1,109 @@
+`net.s3` is an S3-compatible client written in pure V.
+
+It speaks the AWS Signature Version 4 protocol on top of `crypto.hmac` and
+`crypto.sha256`, with no third-party dependencies. The same client targets any
+S3-compatible endpoint (hosted or self-hosted) by configuring the right
+`endpoint`, `region` and credentials.
+
+## Quick start
+
+```v ignore
+import net.s3
+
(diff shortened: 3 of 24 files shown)


davlgd:
Pushed `23501bfe8` addressing the four points.

P1 — `build_object_path` no longer calls `strip_slashes` on the key. S3 keys are opaque identifiers, so `folder/`, `/x` and `a//b` now reach the wire byte-exact (only percent-encoded). The bucket continues to be slash-stripped since it isn't a key. Covered by a new `test_build_object_path_preserves_leading_and_trailing_slashes` in `fetch_test.v`.

P1 — `guess_region` now returns `us-east-1` instead of `auto` for any `*.amazonaws.com` endpoint that doesn't carry an explicit region in the host. Covers `s3.amazonaws.com` and the legacy `s3-external-1.amazonaws.com` forms. Three assertions added to `test_guess_region_paths` in `credentials_test.v`. Non-AWS endpoints still resolve to `auto`.

P2 — In `upload_bytes_multipart`, the `TooManyParts` check moved before the enqueue (was: enqueue → increment → check). Previously an upload that used exactly `max_parts` parts was aborted right after queuing the legitimate final part; now `max_parts` is allowed inclusive, matching the loose `MultipartUploader.upload` path (line 85) which already had the check in the right place.

P2 — `s3.fetch` no longer aliases `.post` to PUT. The `.post` arm is removed from the match and falls through to the `else` branch returning `InvalidMethod`. Rationale: `s3.fetch` is a "URL → bytes" helper, and S3 POST endpoints (`CreateMultipartUpload`, `CompleteMultipartUpload`, browser form upload) all take a structured body that `fetch` doesn't generate — silently rewriting to PUT was misleading. Callers needing POST should use the multipart API directly. Covered by `test_fetch_rejects_post_method`.

`v fmt`, `v vet` and the 8 unit tests are green in both default and `-prod` builds.

medvednikov:
Thanks!

> Having a first-class client in V, with no third-party dependencies and no need to shell out to aws-cli, makes the language a credible choice for cloud workloads.

I agree.

Issue #24590, opened 2025-05-27
gitignore: remove !*.* rule that interferes with personal exclusions


gchumillas:
### Describe the bug

This gitignore rule (`!*.*`) prevents us from ignoring local files:
https://github.com/vlang/v/blob/master/.gitignore#L10

For example, I usually ignore all "test.txt" files, as I often use them to add private notes. Therefore, I add the following rule to the `.git/info/exclude` file:

```
test.*
```

But the `!*.*` rule is more relevant than the previous rules. Therefore, the previous rule has no effect.

### Reproduction Steps

Add a rule to the `.git/info/exclude` file. For example:
```
test.*
```

Include a `test.txt` file into the project directory and execute the `git status` command:
```bash
git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        test.txt
```

As you can see, the `test.txt` file is listed in the "Untracked files" section.

### Expected Behavior

The .gitignore file should not override local ignored files.

### Current Behavior

The .gitignore file overrides local ignored files.

### V version

V 0.4.10 52c7130

### Environment details (OS name and version, etc.)

|V full version      |V 0.4.10 52c7130a2f870afff1242c476aa6db11941abdb9
|:-------------------|:-------------------
|OS                  |macos, macOS, 15.5, 24F74
|Processor           |8 cpus, 64bit, little endian, Apple M1 Pro
|Memory              |0.96GB/16GB
|                    |
|V executable        |/Users/gonzalo/Projects/Personal/vlang/v
|V last modified time|2025-05-26 22:06:31
|                    |
|V home dir          |OK, value: /Users/gonzalo/Projects/Personal/vlang
|VMODULES            |OK, value: /Users/gonzalo/.vmodules
|VTMP                |OK, value: /tmp/v_501
|Current working dir |OK, value: /Users/gonzalo/Projects/Personal/vlang
|                    |
|Git version         |git version 2.39.1
|V git status        |52c7130a
|.git/config present |true
|                    |
|cc version          |Apple clang version 17.0.0 (clang-1700.0.13.3)
|gcc version         |Apple clang version 17.0.0 (clang-1700.0.13.3)
|clang version       |Apple clang version 17.0.0 (clang-1700.0.13.3)
|tcc version         |tcc version 0.9.28rc 2024-02-05 HEAD@105d70f7 (AArch64 Darwin)
|tcc git status      |thirdparty-macos-arm64 e447816c
[...]

PR #27010, opened 2026-04-27
all: remove `*/**/*` gitignore rule and its !path allowlist

Maokaman1:
Simplify `.gitignore` by removing the `*/**/*` ignore-everything rule and the `!path` allowlist that grew on top of it.

Until #24590 (commit `257611f09c`) the broad rule was effectively neutralized by `!*.*`, which re-included any file with an extension. Once `!*.*` was removed, `*/**/*` started actually hiding files with extensions under subdirectories, with two consequences:

1. **Lost editor / search visibility.** Tools that consult `.gitignore` to scope their work - Zed editor search, VS Code search, ripgrep, fzf integrations, language servers, GitHub's web code search - silently skipped large portions of `vlib/` and the tests/testdata trees. Tracked source files still existed in git but became invisible to everyday discovery and indexing.

2. **Maintenance tax on contributors.** Every PR that added a new test fixture or source file under a deep path also had to add a matching `!path` entry. The pattern is visible in recent merges.

Removing the broad rule eliminates both costs. The accumulated `!path` allowlist becomes unnecessary and is dropped along with `!*/` (whose only job was to keep git recursing into directories that `*/**/*` would otherwise have ignored).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,44 +1,5 @@
-# ignore sub-level untracked files and the v binary
-*/**/*
-
-# unignore checker test files
-!vlib/v/checker/tests/*.vv
-!vlib/v/checker/tests/*.out
-!vlib/v/gen/c/testdata/nested_unary_minus_regression.vv
-!vlib/v/gen/c/testdata/nested_unary_minus_regression.out
-!vlib/v/tests/interfaces/interface_pointer_method_value_issue_22237_test.v
-!vlib/v/tests/net_http_request_proxy_interpolation_test.v
-!vlib/v/tests/project_issue_26873/**
-!vlib/v/tests/project_issue_26873_check_test.v
-!vlib/v/checker/tests/modules/c_type_cast_imported_from_module/**
-!vlib/v/checker/tests/modules/c_type_cast_imported_from_module.out
-!vlib/veb/tests/catchall_route_order_regression_test.v
-!vlib/net/http/head_with_content_length_test.v
-!vlib/net/mbedtls/mbedtls_head_with_content_length_test.v
-!vlib/v/checker/tests/modules/imported_struct_field_with_any_type_err/**
-!vlib/v/checker/tests/modules/imported_struct_field_with_any_type_err.out
-
-# unignore c codegen testdata files
-!vlib/v/gen/c/testdata/*.vv
-!vlib/v/gen/c/testdata/*.out
-!vlib/v/gen/c/testdata/*.c
-!vlib/v/gen/c/testdata/*.c.must_have
-
-# unignore targeted js backend regression testdata files
-!vlib/v/gen/js/tests/testdata/comptime_if_expr_issue_24784.v
-!vlib/v/gen/js/tests/testdata/comptime_if_expr_issue_24784.out
-
-# unignore vlib/x/markdown
-!vlib/x/markdown/**
-!vlib/mcp/server.v
-!vlib/mcp/server_test.v
-
-# unignore vlib/net/s3
-!vlib/net/s3/**
-
-!.github/show_manual_release_cmd_test.v
-!*/
-# Do not add !*.* here; it overrides local excludes from .git/info/exclude.
+# ignore generated binaries and object files
+# Do not add broad unignore rules like !*.*; they override local excludes.
 *.exe
 *.o
 *.so
@@ -78,6 +39,7 @@ fns.txt
 temp/
 tmp/
 cache/
+.cache/
 
 # unignore special files without extension
 !.github/PULL_REQUEST_TEMPLATE
@@ -161,6 +123,7 @@ vls.log
 # ignore v2go tmperror files
 *.tmperr
 
+.vtmp_cov_*/
 */**/tmp.*
 
 # ignore Intellij files


.gitignore
```diff
@@ -1,44 +1,5 @@
...
-!vlib/mcp/server_test.v
-
-# unignore vlib/net/s3
-!vlib/net/s3/**
-
-!.github/show_manual_release_cmd_test.v
-!*/
-# Do not add !*.* here; it overrides local excludes from .git/info/exclude.
+# ignore generated binaries and object files
+# Do not add broad unignore rules like !*.*; they override local excludes.
```
JalonSolov, line 2: I would also add `*.dll`.

Shouldn't these be
```
**/*.dll
**/*.exe
**/*.o
**/*.so
```
to ensure blocking them from all directories?

There are many other things which should be blocked by default, as well... I'm fairly sure that's why the original "ignore everything" rule was there.

However, I agree this was getting a bit out of hand...

Maokaman1: `*.dll` is already listed.

In this case, `**/*.dll` and `*.dll` mean the same thing.

Maokaman1: About blocking other things: for normal changes, unexpected files should be caught during PR review, with specific `.gitignore` rules added when needed.
If there is automation that pushes directly, like `vlang/vc`, I guess we should review it to make sure it cannot push unwanted files.

JalonSolov: > `*.dll` is already listed.
> 
> In this case, `**/*.dll` and `*.dll` mean the same thing.

Actually, according to .gitignore rules:

In `.gitignore`, `*.exe` ignores all `.exe` files in the current directory, while `**/*.exe` ignores all `.exe` files in the current directory and all its subdirectories.

Maokaman1: https://git-scm.com/docs/gitignore

- A leading `**` followed by a slash means match in all directories. For example, `**/foo` matches file or directory `foo` anywhere, **the same as pattern** `foo`.

- The pattern `hello.*` matches any file or directory whose name begins with `hello.`. If one wants to restrict this only to the directory and not in its subdirectories, one can prepend the pattern with a slash, i.e. `/hello.*`; the pattern now matches `hello.txt`, `hello.c` but not `a/hello.java`.

JalonSolov: We shall see, I suppose.

Here, on the other hand, are some of those things you left out...  after a fresh `v up`, I see this:
```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        cmd/tools/vcheck-md
        cmd/tools/vfmt
        cmd/tools/vgit-fmt-hook
        cmd/tools/vself
        cmd/tools/vup

nothing added to commit but untracked files present (use "git add" to track)
$
```
Those were previously blocked by the "block everything" rule, but now they'll be committed if people aren't careful.  They are Linux executables, so there is no extension to use to block them.

JalonSolov: As it turns out, you are correct about `*.exe` and the others, and git itself will tell you if something will be ignored or not, and which line causes it to be ignored.  Doesn't matter what any web sites may say, or how they say it, as the tool itself is what actually decides:
```
$ git check-ignore -v x.exe
.gitignore:3:*.exe      x.exe
$ git check-ignore -v vlib/x.exe
.gitignore:3:*.exe      vlib/x.exe
$ git check-ignore -v vlib//x/crypto/x.exe
.gitignore:3:*.exe      vlib//x/crypto/x.exe
$
```

Maokaman1: > Here, on the other hand, are some of those things you left out... after a fresh `v up`
> ...
> Those were previously blocked by the "block everything" rule, but now they'll be committed if people aren't careful. They are Linux executables, so there is no extension to use to block them.

Good catch. This is broader than I first framed it.

Reverting to the pre-`257611f09c` state may be an option.
I think the question is which cost is worse:
- pre-`257611f09c`: extensionless build outputs are hidden, and files with extensions remain visible, but local excludes can be overridden by `!*.*`.
- current state: source visibility and local excludes are better, but Unix extensionless binaries from direct compiles need more explicit ignore coverage.
- compromise: restore the broad extensionless-output ignore, but replace `!*.*` with narrower allow rules like `!*.v`, `!*.vv`, `!*.vsh`, `!*.md`, `!*.out`, etc. That would reduce the `.git/info/exclude` problem compared to `!*.*`, but it still keeps the repo in allowlist mode and would need ongoing updates for legitimate file types.

As an immediate mitigation for the current state, we could add explicit ignores for the common `v up` / maintenance tool binaries:

```gitignore
# ignore compiled cmd/tools binaries (extensionless on Unix)
/cmd/tools/vcheck-md
/cmd/tools/vdoctor
/cmd/tools/vdoc/vdoc
/cmd/tools/vfmt
/cmd/tools/vgit-fmt-hook
/cmd/tools/vpm/vpm
/cmd/tools/vself
/cmd/tools/vsymlink/vsymlink
/cmd/tools/vtest
/cmd/tools/vtimeout
/cmd/tools/vup
/cmd/tools/vvet/vvet
/cmd/tools/vwipe-cache
```

JalonSolov: I think your `compromise` it likely the best bet.

Since V supports so many different OSes, and Windows is the only one I can think of where extensions are common for executables, having a general "exclude everything" rule with exceptions for V extensions is the better idea.

However, if that's done, how do you ignore entire directories (such as `temp/` and `tmp/` from current rules) including .v files under them?

Maokaman1: Gitignore is order-sensitive within the same ignore source: the last matching rule wins.
So one possible compromise layout would be:

```gitignore
# ignore extensionless/generated files below repo root
*/**/*

# allow traversal into subdirectories
!*/

# allow source/docs/fixtures with known extensions
!*.v
!*.vv
!*.vsh
!*.md
!*.out
!*.c
!*.h
!*.json
!*.toml

# hard-ignore generated/temp directories after the allowlist
temp/
tmp/
cache/
.cache/
```

With that order, `vlib/foo.v` is visible, but `tmp/foo.v` is still ignored because `tmp/` comes later.

The caveat is that `$GIT_DIR/info/exclude` has lower priority than repository `.gitignore` files. So if the repo `.gitignore` contains `!*.v`, a personal exclude like this:
```
bug.v
```
in `.git/info/exclude` will still not ignore `bug.v`. That is the same class of problem as `!*.*`, just with a narrower scope.

JalonSolov: I think that sounds like a much better plan, to avoid adding extensionless executables to the repo.

It will still need a thorough check to ensure there are no other files that would be excluded that shouldn't be, such as test .zip files (if there are any), etc., etc.

Maokaman1: After looking closer, I think the extension allowlist has the same problem as `!*.*`, just split by extension.

The repo has many tracked fixture/doc formats, including `.txt`, `.json`, `.toml`, `.out`, `.must_have`, `.tar`, `.gz`, `.bin`, `.xz`, etc. To avoid hiding legitimate files, the allowlist would need to grow pretty wide. But every added rule like `!*.txt` can override a developer's `.git/info/exclude` rule for that extension, which is the same class of problem as #24590.

Given that, restoring the old broad behavior may be the better short-term tradeoff: it keeps extensionless Unix build outputs out of `git status`, while files with extensions remain visible through `!*.*`.

Longer term, the cleaner fix is probably to change build/test workflows so generated executables go into a predictable ignored output directory instead of being written next to sources.

Maokaman1: One practical convention we already support is using `tmp.*` for temporary files that should stay ignored.

The repo `.gitignore` already contains:
```gitignore
*/**/tmp.*
```

So files like these stay ignored even with `!*.*` restored:
```
tmp.test.v
tmp.repro.txt
vlib/foo/tmp.issue_123.v
examples/bar/tmp.notes.md
```

JalonSolov: Wouldn't `**/tmp.*` work as well as `*/**/tmp.*`?

Maokaman1: `*/**/tmp.*` works for first-level directories and deeper, while another existing rule, `/tmp*`, covers root-level `tmp.*` files.
So yes, `**/tmp.*` or just `tmp.*` can replace both old rules.

JalonSolov: ... but only if you also add a `tmp/` rule, to exclude subdirs named `tmp` (which is already there...).

<sigh>

Nothing's easy but this also gets confusing and cumbersome to catch all the edge cases.

PR #27011, opened 2026-04-27
dtm2: introduce parsed runtime template manager

GGRei:
This PR introduces `x.templating.dtm2`, a modernized runtime template manager for V.

The existing `x.templating.dtm` module was originally my older contribution. I wanted to revisit it now with a cleaner architecture, better performance characteristics, and a simpler migration path for existing users.

DTM2 keeps the same core idea: templates remain normal files on disk and can be changed without recompiling the application. The main architectural change is that DTM2 caches parsed template trees instead of rendered HTML responses, removing the old async rendered-cache server from the main rendering path.

The legacy `x.templating.dtm` API remains available for compatibility and now delegates rendering internally to the DTM2 engine, so existing users can migrate progressively.

Main points:

- Adds `x.templating.dtm2`
- Keeps `x.templating.dtm` source-compatible
- Adds parsed-template caching
- Supports HTML/XML-like templates and text templates
- Supports JSON-based extension configuration
- Keeps placeholder values escaped by default
- Restricts templates and includes to the configured `template_dir`
- Adds tests and a DTM2 benchmark runner

Validation run locally:

- `./v fmt -w` on touched V files
- `./v vet vlib/x/templating/dtm2`
- `./v test vlib/x/templating/dtm2`
- `./v test vlib/x/templating/dtm`
- `./v -prod test vlib/x/templating/dtm2`
- `./v -prod test vlib/x/templating/dtm`
- DTM2 benchmark wrapper in `dev`, `prod`, and `prod_o2` modes

### DIFF
--- a/vlib/x/templating/dtm/README.md
+++ b/vlib/x/templating/dtm/README.md
@@ -3,6 +3,15 @@
 A simple template manager integrated into the V project, designed to combine the power of V
 templates with Veb, without the need to recompile the application with every change.
 
+> **Status:** `x.templating.dtm` is the legacy version of Dynamic Template
+> Manager. New projects should use `x.templating.dtm2`, which provides the
+> modern runtime renderer and parsed-template cache.
+>
+> Existing `dtm` users can keep their current API while migrating progressively:
+> the v1 facade now delegates rendering to the DTM2 engine internally, preserving
+> source compatibility while removing the old rendered-cache server from the
+> main rendering path.
+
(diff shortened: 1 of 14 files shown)


GGRei:
I reworked the points raised by Codex.

If there is anything else that needs to be changed, let me know. Otherwise, from my side this is ready for review/integration.

medvednikov:
Thanks!

PR #27029, opened 2026-04-29
x.async: add structured concurrency helpers

GGRei:
This PR introduces `x.async`, a small structured concurrency helper module built only on existing V primitives.

I originally wrote this kind of tooling for my larger V projects, where I needed a simple and explicit way to coordinate tasks, cancellation, timeouts, worker limits, and error propagation. The module now feels mature enough to share as an experimental `x.` module, while keeping the design intentionally modest and easy to review.

What it provides:

- `Group` for running related jobs with first-error propagation and cooperative cancellation.
- `Task[T]` for one value-returning background computation with explicit `wait()`.
- `Pool` for bounded concurrent work with non-blocking `try_submit()` backpressure.
- `every()` for blocking periodic work controlled by context cancellation.
- `with_timeout()` helpers for bounding cooperative jobs.
- Examples, integration tests, benchmarks, and validation scripts.

What it deliberately does not provide:

- no new runtime;
- no scheduler changes;
- no event loop;
- no async/await syntax;
- no hidden background task system;
- no panic recovery;
- no promise chaining;
- no multi-consumer futures;
- no detached ticker object in this first version.

The current API is deliberately conservative:

- cancellation is cooperative;
- `Task.wait()` is one-shot;
- `Group` keeps first-error behavior simple;
- `Pool.try_submit()` is explicit and non-blocking;
- periodic work is blocking and lifecycle-driven by context.

The implementation is only a documented layer over existing V tools like `spawn`, `sync.WaitGroup`, `sync.Mutex`, channels, `context`, and `time`.

I also added examples and integration-style tests for selected V modules such as `net.http`, `net.websocket`, `veb`, and `mcp`, while keeping them local and synthetic so they do not depend on external services, ports, or fragile server shutdown behavior.

Future additions could be discussed separately if maintainers think the direction is useful, for example:

- blocking or timeout-based pool submission;
- detached periodic handles;
- higher-level helpers for server-style code;
- optional error collection helpers;
- more examples/tests around existing V modules that already use concurrency internally;
- possible future adaptation of some current v modules to use `x.async`, if maintainers decide that would be valuable.

The goal is not to replace V’s concurrency model, but to provide a small, predictable, well-tested helper layer for common coordination patterns.

### DIFF
--- a/vlib/x/async/README.md
+++ b/vlib/x/async/README.md
@@ -333,9 +333,10 @@ Error behavior:
 
 - if the job returns first, the job error is returned unchanged;
 - if the timeout expires first, the public error is `async: timeout`;
-- if the parent context is canceled first, the parent context error is returned;
-- a job that observes the timeout by returning `context deadline exceeded` is
-  normalized to `async: timeout`.
+- if the parent context is canceled first, including when the parent's own
+  deadline expires first, the parent context error is returned;
+- a job that observes the local `x.async` timeout by returning
+  `context deadline exceeded` is normalized to `async: timeout`.
 
 The result channel used internally is buffered so the spawned job can finish and
 publish its result even if the caller has already returned on timeout.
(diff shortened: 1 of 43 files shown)


GGRei:
Ready for review / integration. Codex p2 fixed.

PR #27033, opened 2026-04-29
Fix TCC compiler integration for parallel compilation

medvednikov:
## Summary
This PR fixes several issues with TCC (Tiny C Compiler) support, particularly in the context of parallel C compilation. The changes address path resolution problems, duplicate symbol detection, and header inclusion guards.

## Key Changes

- **Parallel CC working directory handling**: Changed TCC compilation to execute from the V root directory, since TCC resolves include and library search paths relative to the invocation directory. This ensures `thirdparty/tcc/lib/tcc/...` lookups work correctly for standard headers and runtime objects.

- **TCC duplicate symbol detection**: Added detection for TCC's specific behavior where duplicate symbol errors are reported via stderr but the linker still exits with code 0. The build now properly fails when `defined twice` appears in TCC's output.

- **Simplified header inclusion guards**: Removed TCC-specific `__TINYC__` preprocessor checks from `get_guarded_include_text()` and `get_inttypes_or_stdint_include_text()` functions, relying instead on the more portable `__has_include` feature that TCC supports.

- **Parallel compilation symbol visibility**: Modified the `VV_LOC` macro definition to be empty (instead of `static`) when `_VPARALLELCC` is defined, allowing proper symbol visibility across separately compiled object files in parallel compilation mode.

## Implementation Details

- Uses `os.getwd()` and `os.chdir()` with defer to safely manage working directory changes during TCC compilation
- Maintains backward compatibility by only applying TCC-specific behavior when `b.ccoptions.cc == .tcc`
- The `_VPARALLELCC` flag controls whether symbols should be static or globally visible across compilation units

https://claude.ai/code/session_012PDdNjq22fSCxGV3PiTBMT

### DIFF
--- a/vlib/v/builder/cbuilder/parallel_cc.v
+++ b/vlib/v/builder/cbuilder/parallel_cc.v
@@ -77,24 +77,26 @@ fn parallel_cc(mut b builder.Builder, result c.GenOutput) ! {
 	}
 
 	cc := b.quote_compiler_name(parallel_cc_compiler_path(b))
-	compile_args := b.get_compile_args()
-	linker_args := b.get_linker_args()
+	mut compile_args := b.get_compile_args()
+	mut linker_args := b.get_linker_args()
+	if b.ccoptions.cc == .tcc {
+		// vlang/tcc has its system headers under `${vroot}/thirdparty/tcc/lib/tcc/include/`
+		// and its runtime objects (libtcc1.a, bt-*.o) under `${vroot}/thirdparty/tcc/lib/tcc/`.
+		// `-B` controls tcc's include search (`${B}/include`) and `-L` adds a library search path,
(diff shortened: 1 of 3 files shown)


JalonSolov:
I thought we were going to drop tcc once v2 was done.  There would no longer be a need for it.

Even bootstrapping a fresh install could be accomplished other ways.

medvednikov:
This is a V1 fix, not V2. Have to make all tests pass.

Jengro777:
> This is a V1 fix, not V2. Have to make all tests pass.



This will consume a lot of time, and it would be better if v2 could pass all tests as soon as possible if possible.

JalonSolov:
All the tests would have to be converted over, all the CI jobs would have to be updated, etc.  If the goal is to pass all tests with v1, then switch to v2, that's alright.

PR #27043, opened 2026-04-30
gzip: fix decompression of a zero length payload

kimshrier:
I found an edge case where decompressing a 0 length payload did not
return a 0 length result.

I have added some additional test cases around this.


kimshrier:
actually, it looks like zlib, deflate, and zstd also have an issue with decompressing a 0 length payload.

I'll figure out a different approach.

PR #27048, opened 2026-05-01
gen: fix comptime zero/new type expression resolution

### LINKED ISSUE #26995: `$new(T.pointee_type)` allocates wrong type when generic function is instantiated for multiple pointer types
## Describe the bug

When a generic function uses `$new(T.pointee_type)` and is instantiated with two or more different pointer types in the same program, every instantiation generates C code that allocates the *same* (last-seen) pointee type rather than the per-instantiation type. The result is silently wrong code: the returned pointer's static V type is correct but the heap object behind it is of a different type, leading to memory corruption, wrong values, or runtime errors.

Introduced by 8b5f74e6a7b52c1100cb52ef2832fd0c090fb407.

## Reproduction Steps

```v
module main

struct Foo {
    name string
}

fn make_pointer[T]() T {
    $if T.indirections == 1 {
        mut p := $new(T.pointee_type)
        return p
    } $else {
        return $zero(T)
    }
}

fn main() {
    int_ptr := make_pointer[&int]()
    foo_ptr := make_pointer[&Foo]()
    dump(*int_ptr)
    dump(*foo_ptr)
}
```

## Expected Behavior

```
[file.v:18] *int_ptr: 0
[file.v:19] *foo_ptr: Foo{
    name: ''
}
```

## Current Behavior

```
[file.v:18] *int_ptr: 8636372
[file.v:19] *foo_ptr: Foo{
    name: ''
}
```

`*int_ptr` reads garbage. Inspecting the generated C with `v -keepc` shows the `make_pointer[&int]` instantiation allocates a `Foo`, not an `int`:

```c
VV_LOC int* main__make_pointer_T___ptr__int(void) {
    main__Foo* p = HEAP(main__Foo, ((main__Foo){.name = (string){.str=(byteptr)\"\", .is_lit=1},}));
    return p;
}
```

Both `_T___ptr__int` and `_T___ptr__main__Foo` instantiations emit `HEAP(main__Foo, ...)`. The "last seen" pointee type wins for all instantiations.

This blocks refactoring `vlib/x/json2/decode.v` to use `$new(T.pointee_type)` instead of the manual `create_decoded_ptr` helper.

## Possible Solution

Resolve `T.pointee_type` per-instantiation in the C generator instead of caching/sharing across instantiations.

## V version

V 0.5.1 6dd9033de29819f9ada4011f783b1d8b9d47d7e8.eb1d47b

## Environment details

linux, Ubuntu 24.04 LTS, gcc 14.2.0, tcc 0.9.28rc
> [!NOTE]
>

### LINKED ISSUE #26996: `$zero(field.typ.payload_type)` inside `$for field in T.fields` uses the same payload type for every option field
## Describe the bug

Inside `$for field in T.fields`, when a struct has multiple `?T` fields with *different* non-pointer payload types, `$zero(field.typ.payload_type)` resolves to the same payload type for every iteration (the last one) instead of the current field's payload type. This causes a C compilation error.

Introduced by 8b5f74e6a7b52c1100cb52ef2832fd0c090fb407.

Related: #26995 — same family of bug for `T.pointee_type`.

## Reproduction Steps

```v
module main

struct Foo {
mut:
    a ?int
    b ?string
}

fn fill[T](mut x T) {
    $for field in T.fields {
        $if field.typ is $option {
            mut v := $zero(field.typ.payload_type)
            x.$(field.name) = v
        }
    }
}

fn main() {
    mut f := Foo{}
    fill(mut f)
    dump(f)
}
```

## Expected Behavior

Compile cleanly. `f.a` becomes `Some(0)` and `f.b` becomes `Some('')`.

## Current Behavior

```
cc: /tmp/v_1000/<...>.tmp.c:8530: error: cannot convert 'struct string' to 'int'
```

With gcc:

```
error: incompatible types when initializing type 'int' using type 'string'
 8530 |    builtin___option_ok(&(int[]) { v }, (_option*)(&_t2), sizeof(int));
```

Generated C for the `a: ?int` iteration declares `v` as `string` (the payload type of `b: ?string`, the last option field) and then tries to wrap it in `_option_int`, hence the type mismatch.

## Possible Solution

Resolve `field.typ.payload_type` per-iteration of `$for field in T.fields` instead of using a single resolved type for the whole expansion.

## V version

V 0.5.1 6dd9033de29819f9ada4011f783b1d8b9d47d7e8.eb1d47b

## Environment details

linux, Ubuntu 24.04 LTS, gcc 14.2.0, tcc 0.9.28rc
> [!NOTE]
>

### LINKED ISSUE #26997: `$new(field.typ.pointee_type)` inside `$for field in T.fields` allocates the wrong type when generic is instantiated for multiple structs
## Describe the bug

When a generic function using `$new(field.typ.pointee_type)` is instantiated for multiple struct types whose pointer fields have different pointee types, every instantiation generates C code that allocates the *same* (last-seen) pointee type. Statically the return type of `$new` is correct, but the underlying heap object is of the wrong type — producing silently wrong reads, type confusion, and runtime crashes when the wrong-typed object is dereferenced.

Introduced by 8b5f74e6a7b52c1100cb52ef2832fd0c090fb407.

Related: #26995 (same family of bug for `T.pointee_type` directly), #26996 (`$zero(field.typ.payload_type)`).

## Reproduction Steps

```v
module main

struct Inner {
    value int
}

struct WrapperA {
mut:
    a &int = unsafe { nil }
}

struct WrapperB {
mut:
    b &Inner = unsafe { nil }
}

fn fill[T](mut x T) {
    $for field in T.fields {
        $if field.indirections == 1 {
            mut p := $new(field.typ.pointee_type)
            x.$(field.name) = p
        }
    }
}

fn main() {
    mut a := WrapperA{}
    mut b := WrapperB{}
    fill(mut a)
    fill(mut b)
    dump(*a.a)
    dump(b.b.value)
}
```

## Expected Behavior

`fill[WrapperA]` allocates an `int` for `a.a`; `fill[WrapperB]` allocates an `Inner` for `b.b`.

## Current Behavior

The program runs and prints `0` / `0`, but inspecting `v -keepc` shows that *both* `fill[WrapperA]` and `fill[WrapperB]` emit `main__Inner* p = HEAP(main__Inner, ...)`:

```c
VV_LOC void main__fill_T_main__WrapperA(main__WrapperA* x) {
    /* field 0 : a — field.typ = 8 (&int) */
    {
        main__Inner* p = HEAP(main__Inner, ((main__Inner){.value = 0,}));
        x->a = p;   // assigning Inner* to int* (silent in C)
    }
}
```

`x->a` is declared `int*` but receives an `Inner*` from the heap. In this minimal repro it appears to "work" because zeroed memory reads as `0`, but in real code (e.g. `vlib/x/json2/decode.v` decoding multiple struct types each containing pointer fields) it leads to runtime panics during downstream decode, since the heap object is the wrong type and shape.

## Possible Solution

Resolve `field.typ.pointee_type` per-field-per-instantiation instead of sharing one resolved type across all `$for` expansions.

## V version

V 0.5.1 6dd9033de29819f9ada4011f783b1d8b9d47d7e8.eb1d47b

## Environment details

linux, Ubuntu 24.04 LTS, gcc 14.2.0, tcc 0.9.28rc
> [!NOTE]
>

GGRei:
fix #26995 
fix #26996 
fix #26997 

Resolving `$zero(TypeExpr)` and `$new(TypeExpr)` in the current C generation context instead of relying on a stale type stored on the shared AST node. The patch covers generic specializations like `T.pointee_type`, per-field comptime metadata like `field.typ.payload_type` / `field.typ.pointee_type`, and adds a C-output regression for the case where runtime could hide the wrong allocation. During tests, I also found and covered a related generic `typeof(x).payload_type` case for both `$zero(...)` and `$new(...)`.

### DIFF
--- a/vlib/v/gen/c/assign.v
+++ b/vlib/v/gen/c/assign.v
@@ -924,8 +924,12 @@ fn (mut g Gen) assign_stmt(node_ ast.AssignStmt) {
 							}
 						}
 					} else if val is ast.ComptimeCall {
-						key_str := '${val.method_name}.return_type'
-						var_type = g.type_resolver.get_ct_type_or_default(key_str, var_type)
+						var_type = if val.kind in [.zero, .new] {
+							g.comptime_zero_new_result_type(val, var_type)
+						} else {
+							key_str := '${val.method_name}.return_type'
+							g.type_resolver.get_ct_type_or_default(key_str, var_type)
+						}
 						val_type = var_type
 						left.obj.typ = var_type
 						g.type_resolver.update_ct_type(left.name, var_type)
@@ -1140,6 +1144,9 @@ fn (mut g Gen) assign_stmt(node_ ast.AssignStmt) {
 				ast.ComptimeSelector {
 					should_recompute_decl_type = true
 				}
+				ast.ComptimeCall {
+					should_recompute_decl_type = val.kind in [.zero, .new]
+				}
 				ast.CastExpr {
 					should_recompute_decl_type = val.typ.has_flag(.generic)
 						|| g.type_has_unresolved_generic_parts(val.typ)
@@ -1166,8 +1173,10 @@ fn (mut g Gen) assign_stmt(node_ ast.AssignStmt) {
 					|| val.receiver_type.has_flag(.generic)
 					|| g.type_has_unresolved_generic_parts(val.receiver_type)))
 			}
-			mut resolved_val_type := if val is ast.Ident && val.or_expr.kind != .absent
-				&& g.cur_fn != unsafe { nil } && g.cur_concrete_types.len > 0 {
+			mut resolved_val_type := if val is ast.ComptimeCall && val.kind in [.zero, .new] {
+				g.comptime_zero_new_result_type(val, val_type)
+			} else if val is ast.Ident && val.or_expr.kind != .absent && g.cur_fn != unsafe { nil }
+				&& g.cur_concrete_types.len > 0 {
 				or_value_type := g.resolved_or_block_value_type(val.or_expr)
 				if or_value_type != 0 {
 					or_value_type
--- a/vlib/v/gen/c/comptime.v
+++ b/vlib/v/gen/c/comptime.v
@@ -63,6 +63,197 @@ fn (mut g Gen) comptime_zero_value(typ ast.Type) string {
 	return default_value
 }
 
+fn (mut g Gen) comptime_type_expr_type(expr ast.Expr, fallback_type ast.Type) ast.Type {
+	match expr {
+		ast.ParExpr {
+			return g.comptime_type_expr_type(expr.expr, fallback_type)
+		}
+		ast.TypeNode {
+			return g.unwrap_generic(g.recheck_concrete_type(expr.typ))
+		}
+		ast.TypeOf {
+			if expr.is_type {
+				return g.unwrap_generic(g.recheck_concrete_type(expr.typ))
+			}
+			mut default_type := g.get_type(expr.typ)
+			default_type = g.resolve_typeof_expr_type(expr.expr, default_type)
+			if expr.expr is ast.Ident && expr.expr.obj is ast.Var {
+				resolved := g.comptime_typeof_generic_ident_type(expr.expr)
+				if resolved != 0 {
+					return resolved
+				}
+			}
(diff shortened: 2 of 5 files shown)


GGRei:
Looks like the GitCodex review is happy. Ready for final review!

PR #27058, opened 2026-05-02
cgen: resolve generic map types in concrete contexts

guweigang:
## Summary

Fix C generation for generic map key/value types when a generic struct method is instantiated with multiple concrete types.

The bug showed up when a generic struct owns a map whose value type depends on `T`, for example `map[string]StoredValue[T]`. The C backend could reuse stale map key/value types from a previous concrete instantiation, producing mixed C types for different `T`s.

## Changes

- Resolve map key/value types from the current concrete map type before generating map literals and map index access.
- Reuse the same map type resolution path for map index expressions.
- Resolve `if value := map[key]` guard variables from the concrete map value type.
- Use the expected struct field map type when generating map literals inside struct initialization.
- Add a regression test covering:
  - `map[string]Wrapped[T]`
  - `map[string]T`
  - `map[int]Wrapped[T]`
  - `map[int]T`
  - empty and seeded map literals
  - map set/get and if-guard access

## Validation

- `./v test vlib/v/tests/generics`
- `./v test vlib/v/tests/options/option_map_struct_field_init_test.v`
- `./v test vlib/v/tests/concurrency/shared_elem_test.v`
- `./v vlib/v/generics/new_generics_regression_test.v`
- `git diff --check`


guweigang:
@codex review

guweigang:
@codex review

guweigang:
Is there anything I can do for this PR?

medvednikov:
Sorry for the delay, can be merged now that master is green.

Please fix the conflict, and I'll merge it.

PR #27065, opened 2026-05-02
ci: more diagnostics for tcc-windows SEH crash

medvednikov:


medvednikov:
Closing — the 256 MiB stack-reserve bump propagated correctly to v.exe (PE diagnostic confirmed `SizeOfStackReserve = 0x10000000`), but the install_thirdparty_sqlite.vsh crash persists with the exact same signature (`0xc00000fd` at `e56950`). `v -cc gcc -o v_gcc.exe cmd/v` also crashes at the same address.

This is a longstanding tcc-windows failure on master (60+ consecutive failed runs going back days), not something this PR introduced. My current read is that TCC doesn't emit `__chkstk` stack probes on Windows, so a function with a single large stack frame jumps past the guard page regardless of `SizeOfStackReserve` — bumping reserve has no effect because the guard sits just above *committed* stack, not reserved.

Replacing the `v vsh-script` sqlite install with a direct PowerShell download in a separate PR to sidestep the bug and unblock the rest of CI.

PR #27075, opened 2026-05-03
cgen: guard coverage counter file creation

guweigang:
## Summary

Guard coverage counter file creation before writing coverage rows.

## Root cause

The generated `vprint_coverage_stats()` opened the counter CSV with `fopen(...)` and immediately passed the result to `fprintf(...)`. If the coverage directory was not writable or was otherwise invalid, `fopen` could return `NULL`, causing a segfault during `_vcleanup()`.

This also fixes a typo in the generated non-Windows timestamp assignment, where `ts.tv_nsec` was assigned to `secs` instead of `nsecs`.

## Changes

- Return early from generated `vprint_coverage_stats()` when the counter file cannot be opened.
- Generate `nsecs = ts.tv_nsec;` for the monotonic timestamp nanoseconds field.
- Add a C output regression test for the generated guard and timestamp assignment.

## Validation

- `./v -o ./v4 ./cmd/v`
- `./v4 -cov /private/tmp/vcov_as_file vlib/v/tests/fns/closure_struct_init_cov_regression_test.v`
- `./v4 -cov /private/tmp/vcov_ok vlib/v/tests/fns/closure_struct_init_cov_regression_test.v`
- `./v4 test vlib/v/gen/c/coutput_test.v`
- `git diff --check`


medvednikov:
Sorry for the delay, PRs get merged instantly now that master is not chaotic anymore :)

PR #27077, opened 2026-05-03
checker: support comptime fn type signature checks

guweigang:
## Summary

This PR teaches the parser and checker to handle anonymous function types in generic/comptime type contexts:

- recognizes generic calls like `foo[fn (int) int]()`
- makes `$if T is fn (int) int` compare function type signatures, including aliases and anonymous function types
- keeps pointer/function-type modifiers distinct, so `&fn (int) int` does not match `fn (int) int`

## Root Cause

`foo[fn (int) int]()` was parsed through the generic-call heuristic before `parse_type()` could see the anonymous function type. The heuristic did not recognize `fn` as a concrete type start in that position.

For `$if T is fn (...)`, the checker compared resolved type ids. That works for many named types, but not for equivalent function signatures where one side is an alias or an anonymous function type.

## Validation

- `vlib/v/tests/comptime/comptime_if_fn_type_signature_test.v`
- `vlib/v/tests/comptime`
- `vlib/v/parser vlib/v/checker`
- `vlib/v/tests/generics`
- `git diff --check`

### DIFF
--- a/vlib/v/ast/table.v
+++ b/vlib/v/ast/table.v
@@ -2557,6 +2557,12 @@ pub fn (mut t Table) convert_generic_type(generic_type Type, generic_names []str
 				rtyp = rtyp.set_flag(.option_mut_param_t)
 			}
 		}
+		if typ.has_flag(.shared_f) {
+			rtyp = rtyp.set_flag(.shared_f)
+		}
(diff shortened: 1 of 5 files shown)


guweigang:
Closing this because the changes were included via #27085, which has been merged into master.

PR #27084, opened 2026-05-03
ci: fix windows tcc stack

Maokaman1:
Fixes the Windows TCC CI crash by matching `makev.bat`’s stack reserve when compiling `v.c` directly with TCC.

The direct TCC-built compiler was using the default Windows stack and could crash while compiling `net.http`-heavy scripts such as the SQLite installer. The workflow now passes `-Wl,-stack=33554432`, removes the temporary diagnostic probe, and restores the SQLite install step to use the existing V `.vsh` script instead of the PowerShell workaround.


medvednikov:
thanks!

PR #27114, opened 2026-05-08
fix orm dynamic query-data guard folding

### LINKED ISSUE #27112: ORM dynamic query-data: optional `if guard` inside `where/set` block causes cgen error
### Describe the bug

Summary

Using an optional `if guard` inside an ORM dynamic query-data block causes cgen to fail with:

```text
cgen error: ORM: dynamic query-data leaf must be an infix expression
```

This appears related to ORM dynamic query-data parsing/codegen. A normal condition like:

```v
if req.ids.len > 0 { id in req.ids }
```

works, but an optional guard like:

```v
if username := req.username { username == username }
```

is accepted syntactically, then fails during cgen.

This was found while building a project with:

```bash
v -prod -o app ./main

### Reproduction Steps
```

```
import db.sqlite

@[table: 'users']
struct User {
mut:
	id       string @[primary]
	username string
	nickname string
}

struct ListReq {
	username ?string
	nickname ?string
	ids      []string
}

fn main() {
	defer {
		db.close() or { panic(err) }
	}

	sql db {
		create table User
	}!

	alice := User{
		id:       'u1'
		username: 'alice'
		nickname: 'Alice'
	}
	bob := User{
		id:       'u2'
		username: 'bob'
		nickname: 'Bob'
	}

	sql db {
		insert alice into User
		insert bob into User
	}!

	req := ListReq{
		username: 'alice'
		ids:      ['u1', 'u2']
	}

	where_expr := {
		if username := req.username { username == username },
		if nickname := req.nickname { nickname == nickname },
		if req.ids.len > 0 { id in req.ids }
	}

	rows := sql db {
		dynamic select from User where where_expr
	}!

	assert rows.len == 1
	println(rows.len)
}

```

v -prod  orm_dynamic_optional_guard_bug.v

### Expected Behavior

The optional guard should be handled as a dynamic query-data `if` branch, equivalent to:

```v
if username := req.username {
	username == username
}
```

and should generate valid ORM query data.

### Current Behavior

cgen error: ORM: dynamic query-data leaf must be an infix expression

### Possible Solution

_No response_

This seems different from #26957 / commit `22795fdc9018355f8a7a40745c0437d99f08491a`, which added support for `ast.CastExpr` in dynamic update blocks.

The failure here appears to happen because the cgen path for ORM dynamic query-data expects each leaf to be an `ast.InfixExpr`:

```v
fn (mut g Gen) emit_sql_query_data_leaf(...) {
	expr_ := expr.remove_par()
	if expr_ is ast.InfixExpr {
		...
	} else {
		verror('ORM: dynamic query-data leaf must be an infix expression')
	}
}
```

But `if username := req.username { username == username }` appears to reach cgen as a non-infix leaf instead of being represented/emitted as `SqlQueryDataIf`.

### Workaround

Move optional unwrapping outside the ORM dynamic block:

```v
mut username := ''
mut has_username := false
if v := req.username {
	username = v
	has_username = true
}

where_expr := {
	if has_username { username == username }
}
```
### V version

V 0.5.1 9f76fd0

|V full version      |V 0.5.1 d45d40e98dcc1e9a2deb8d2b09ed0ced2c4fc5f7.9f76fd0
|Memory              |4.67GB/31.03GB
|                    |
|V last modified time|2026-05-08 06:12:46
|                    |
|                    |
|V git status        |0.5.1-1504-g9f76fd04
|.git/config present |true
|                    |
|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.38-6deepin21) 2.38

> [!NOTE]
>

Jengro777:
Fixes #27112.

This prevents the transformer from folding dynamic ORM query-data leaf
expressions like `name == name` to `true` in `-prod` builds.

In dynamic ORM query-data blocks, the left side of a comparison is a table
field, while the right side can be a V variable with the same name. When the
query-data block is stored in an alias and then used by `dynamic select` or
`dynamic update`, `-prod` optimization could treat it as a normal self
comparison and fold it, causing cgen to fail with:

```text
cgen error: ORM: dynamic query-data leaf must be an infix expression
```

The fix reuses the existing SQL-context guard while transforming
`SqlQueryDataLeaf` expressions.

Tests added for:
- dynamic `where` query-data alias with an optional guard
- dynamic `set` query-data alias with an optional guard

### DIFF
--- a/vlib/orm/orm_dynamic_test.v
+++ b/vlib/orm/orm_dynamic_test.v
@@ -11,6 +11,10 @@ mut:
 	status string
 }
 
+struct DynamicMemberFilter {
+	name ?string
+}
+
 @[table: 'dynamic_cast_members']
 struct DynamicCastMember {
 mut:
@@ -76,6 +80,21 @@ fn test_dynamic_select_with_inline_where_block() {
 	assert rows[0].name == 'Alice'
 	assert rows[0].email == 'alice@example.com'
 	assert rows[0].age == 31
+
+	filter := DynamicMemberFilter{
+		name: 'Alice'
+	}
+	where_expr := {
+		if name := filter.name {
+			name == name
+		}
+	}
+
+	guard_rows := sql db {
+		dynamic select from DynamicMember where where_expr
+	}!
+
+	assert guard_rows.len == 2
 }
 
 fn test_dynamic_update_with_alias_set_block() {
@@ -149,8 +168,13 @@ fn test_dynamic_update_with_alias_set_block_cast_expr() {
 	id := db.last_id()
 	next_name := 'Alicia'
 	next_required := true
+	filter := DynamicMemberFilter{
+		name: next_name
+	}
 	update_expr := {
-		name == next_name,
+		if name := filter.name {
+			name == name
+		},
 		is_required == u8(if next_required { 1 } else { 0 })
 	}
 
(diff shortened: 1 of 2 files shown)


Jengro777:
@codex review

PR #27162, opened 2026-05-15
compress.zlib: changed to pure V RFC-compliant code

JalonSolov:
Also added `interop` subdir with tests for full V<->C<->Python cross-validation.


JalonSolov:
Bounce to re-run CI.

vlib/compress/zlib/zlib_test.v
```diff
@@ -1,8 +1,67 @@
...
+	decompress(data) or {
+		assert err.msg() == reason
+		return
+	}
+	return error('did not error')
+}
+
+fn test_zlib_roundtrip_text() {
+	data := 'Hello world!'.bytes()
+	compressed := compress(data)!
```
spytheman, line 19:
```suggestion
	compressed := compress(data)!
	assert compressed != data
```

spytheman (commented):
It may be a good idea, to verify that compress does not just return its input.

vlib/compress/zlib/zlib_test.v
spytheman: It may be a good idea, to verify that compress does not just return its input.

JalonSolov: I suppose it couldn't hurt, but it should never be necessary.  `zlib` will always add it's header to the output, so even if the original data is unchanged, the output will always be different due to that header.

PR #27189, opened 2026-05-18
compress.zlib: changed to pure V RFC-compliant code

JalonSolov:
Also has `interop` subdir with code for full V<->C<->Python cross-validation.


JalonSolov:
The last commit fixed a bug in embed_file that's been there a long time, but miniz allowed it to work.  It was caught because this zlib code is stricter than miniz.

It didn't fail unless you used `-prod` to compile your code.

**Why it only failed in `-prod`**

In `-prod`, embedded files are loaded from the generated static blob table, and compressed embeds use the compressed path.

Outside `-prod`, the same compressed-embed runtime path wasn't followed, so the bug didn’t show up the same way.

**Why the new zlib code exposes it**

The previous miniz-backed path was more forgiving and could ignore junk after the DEFLATE stream.

The new code is stricter (per the RFC) and now rejects malformed zlib containers where extra bytes appear between the end of the DEFLATE payload and the final Adler-32 footer

So the stricter decoder exposed a bug that was already present in the caller.

**Fix**

The runtime needed to remember the compressed length separately from the uncompressed length.
In practice:
* added a compressed_len field to EmbedFileData
* set it from compressed_bytes.len in gen/c/embed.v
* used that compressed length in ed.compressed.vbytes(...)

Before this fix, the code only stored one len, and that value was being used for the wrong purpose.

JalonSolov:
Note that the `v -silent vlib/v/gen/c/coutput_test.v` that's failing in CI, runs fine on Linux.

I don't have a Windows system to test on.  No BSD systems, either.  I'll look into a docker container, qemu, etc. unless someone else can take a look...

medvednikov:
```
>>>>>>  retry: 0 ; max_retry: 3 ; r.exit_code: 1 ; trimmed_output.len: 587
>>>>>> V panic: net.mbedtls SSLListener.accept, mbedtls_ssl_handshake failed 1; handshake ret: -80
>>>>>>  v hash: 552d61d
>>>>>>     pid: 0x13a2a
>>>>>>     tid: 0x13a2f
>>>>>> /tmp/v_1001/tsession_7f7c60d6f740_01KS6T2VHWNM0JTGKKC956X0FD/mbedtls_concurrent_large_response_test:1022: at v_stable_sort: Backtrace
>>>>>> /tmp/v_1001/tsession_7f7c60d6f740_01KS6T2VHWNM0JTGKKC956X0FD/mbedtls_concurrent_large_response_test:1022: by v_stable_sort
>>>>>> /tmp/v_1001/tsession_7f7c60d6f740_01KS6T2VHWNM0JTGKKC956X0FD/mbedtls_concurrent_large_response_test:1022: by v_stable_sort
>>>>>> 0x007d720f: by ???
>>>>>> 0x7a502dabb4110dfd: by ???
>>>>>> 0x00000000: by ???
```

JalonSolov:
How would that be affected by the code in this PR?

PR #27194, opened 2026-05-19
gg: Draw lines as lines instead of line strips

SurmanPP:
The benefit is that sokol can batch lines together while it cant batch together line strips which provide no obvious benefit over lines.

I cannot make out any drawback from this change. I traced this line back to this PR: #10534


SurmanPP:
Whoopsie gonna undo that... git's complicated...

medvednikov:
Good to see green PRs again 😅

PR #27196, opened 2026-05-19
gg: make create_image* take a config...

SurmanPP:
...including options for texture_filter and nr_mipmaps

Configuring mipmaps is important depending on the use case and while at it I simplified the code and deprecated the `*_with_filter` functions. With increasing minification the texture quality degrades increasingly without mipmaps.

When the `nr_mipmaps` is set to a value below 0 the maximum number of mipmap levels is used. The default is still 0. This code is fully backwards compatible.

I also removed the `num_slices` option from the `StreamingImageConfig` as it is unused and the slice functionality of sokol is not exposed through any API on the V side.

P.S.: Also removed the `num_mipmaps` option from `StreamingImageConfig` as currently the current API provided by gg is not sufficient to make use of this.

### DIFF
--- a/vlib/gg/image.c.v
+++ b/vlib/gg/image.c.v
@@ -18,7 +18,7 @@ pub mut:
 	width          int
 	height         int
 	nr_channels    int
-  nr_mipmaps     int
+	nr_mipmaps     int
 	ok             bool
 	data           voidptr
 	ext            string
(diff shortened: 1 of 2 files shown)


SurmanPP:
There was a CI failure related to my changes. I ran the `image_test.v` locally and it passed. Should hopefully fix any failures related to my changes. I'll write some additional tests for my additions.

SurmanPP:
Okay. I'm going to have to revise it once again. I had the misconception sokol would generate mipmaps and how StreamingImages worked. It makes sense to expose the nr_mipmaps and nr_slices on the StreamingImageConfig. They could be used through `gfx.Image.update_image` directly but at the same time it would be also worth a consideration exposing an API from gg.

SurmanPP:
So okay i think a cleanup of image.v is needed. I think it would be best separating the mipmap generating from image creation.

PR #27198, opened 2026-05-19
datatypes: BSTree to_left and to_right fix

ZillaZ:
There's a bug when calling the BSTree method to_left and to_right on the leftmost non root node and rightmost non root node.
Example:
```v
module main

import datatypes { BSTree }

fn main() {
	mut tree := BSTree[int]{}
	tree.insert(20)
	tree.insert(10)
	tree.insert(30)
	mut min := tree.min()!
	left := tree.to_left(min) or { -1 }
	println(left)
}
```
Will give this error:
```
/tmp/v_0/sample.01KS0R4G5XD5DEAKN3CMK264FN.tmp.c:837: at v_stable_sort: RUNTIME ERROR: invalid memory access
/tmp/v_0/sample.01KS0R4G5XD5DEAKN3CMK264FN.tmp.c:837: by v_stable_sort
/tmp/v_0/sample.01KS0R4G5XD5DEAKN3CMK264FN.tmp.c:837: by v_stable_sort
0x78ec4d5836c1: by ???
0x78ec4d5837f9: by ???
0x006041b5: by ???
```
This happens because the new_node method on `vlib/datatypes/bstree.v` does not initialize the left and right nodes:
```v
// new_node creates a new bst node with a parent reference.
fn new_node[T](parent &BSTreeNode[T], value T) &BSTreeNode[T] {
	return &BSTreeNode[T]{
		is_init: true
		value:   value
		parent:  parent
	}
}
```

Initializing the nodes is another possible solution, but that would cause the `to_left` and `to_right` methods to return a "zeroed" instance of the value instead of nothing.

### DIFF
--- a/vlib/datatypes/bstree.v
+++ b/vlib/datatypes/bstree.v
@@ -249,13 +249,7 @@ fn (bst &BSTree[T]) get_node(node &BSTreeNode[T], value T) &BSTreeNode[T] {
 	return bst.get_node(node.left, value)
 }
 
-// to_left returns the value of the node to the left of the node with `value` specified if it exists,
-// otherwise the a false value is returned.
-//
-// An example of usage can be the following one
-//```v
-// left_value, exist := bst.to_left(10)
-//```
+// to_left returns the value of the node to the left of the node with `value` specified if it exists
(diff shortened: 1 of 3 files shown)


medvednikov:
Thanks!

PR #27200, opened 2026-05-19
cgen: emit GC_allow_register_threads() constructor for shared libs

### LINKED ISSUE #27178: -shared: auto-call GC_allow_register_threads() from a library constructor
> Reproduced on V 0.5.1 (commit 99f141f74 in upstream master);
> verified no relevant commits land between that and current
> master b6dfae560. OS: macOS 26.4.1 ARM64.

### Problem

A V `-shared` library built with the default Boehm GC is unusable from
host-spawned threads (Rust cargo workers, C# tasks, Java JNI threads)
because libgc aborts with `"Collecting from unknown thread"` whenever a
collection is triggered from a thread libgc has not been told about.

To register host threads, the binding must:

1. Call `GC_allow_register_threads()` **once** at module load (process-
   level).
2. Call `GC_register_my_thread(&stack_base)` **per thread** before that
   thread enters any V code.

Step 2 is inherently per-binding (the binding knows when it spawns a
thread). Step 1 is process-level and could be done by V itself.

### Proposed fix

When emitting a `-shared` library and the default GC is Boehm, emit a
constructor that runs at `dlopen` time:

    __attribute__((constructor))
    static void _v_shared_lib_init(void) {
        GC_allow_register_threads();
    }

That is the entire change. After it, host bindings only need the
per-thread call, which they need anyway.

### Downstream evidence

CX currently exports its own `cx_init` symbol whose entire body is

    void cx_gc_allow_register_threads(void) {
        GC_allow_register_threads();
    }

Every binding (Rust, C#, Java, …) is required to call `cx_init()` at
module load, and every binding's onboarding doc has a "remember to
call cx_init" footgun. With this V change, the `cx_init` export goes
away.

### Why not GC-side?

bdwgc deliberately requires the host to opt in to thread-registration,
because for single-threaded users the cost of the lookup tables is
real. V already knows when it's building a multi-threaded `-shared`
library; emitting the constructor is the right layer.

> [!NOTE]
>

eptx:
Closes #27178.

When a V `-shared` library is dlopen'd by a host runtime that spawns
its own OS threads (Rust cargo workers, C# tasks, Java JNI threads,
etc.), the host threads that enter V code must be registered with
libgc — otherwise libgc aborts with `"Collecting from unknown thread"`
on the first collection triggered from an unregistered thread.

Per-thread registration via `GC_register_my_thread()` is inherently
caller-side, but the process-level enable `GC_allow_register_threads()`
can be done by V itself at library load time. This PR emits a
`__attribute__((constructor))` that calls it, so bindings no longer
have to remember.

The constructor also calls `GC_set_pages_executable(0)` before
`GC_INIT()` to disable libgc's trampoline allocator (which requests
rwx pages via mprotect — denied under macOS hardened runtime). The
explicit `GC_INIT()` makes the timing deterministic so subsequent
allocations don't lazy-init under different settings.

All three calls are:
- Gated on `gc_mode in [boehm_full, boehm_incr, boehm_full_opt, boehm_incr_opt, boehm_leak]`
- Skipped on Windows (which uses a different shared-lib init path via gen_windows_shared_library_boehm_init)
- Guarded by `#if defined(GC_THREADS)` for the thread-registration call

Tested locally on macOS 26.4 ARM64 against a real-world V `-shared`
consumer (the cx-home/cx project, a structured-data library with 5
host-language bindings). Before this PR, every Rust/Go/C#/Java/TS
binding had to hand-roll a `cx_init` wrapper that called
`GC_allow_register_threads()` at module load. After: bindings can
just dlopen the library and start using it.

V's own `make test vlib/builtin` passes 35/35 with this patch.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -9739,6 +9739,36 @@ fn (mut g Gen) write_init_function() {
 		// provide a constructor/destructor pair, ensuring that all constants
 		// are initialized just once, and that they will be freed too.
 		// Note: os.args in this case will be [].
+		//
+		// Fix for vlang/v#27178: when the GC mode is one of the Boehm
+		// variants, host-spawned threads (Rust cargo workers, C# tasks,
+		// JNI threads, etc.) that enter V code must be registered with
+		// libgc via GC_register_my_thread() to avoid the libgc
+		// "Collecting from unknown thread" abort. The per-thread step
+		// is inherently caller-side, but `GC_allow_register_threads()`
+		// is process-level and can be done by V itself at library load
+		// time. Emit a constructor for it so the caller doesn't have
+		// to remember.
+		if g.pref.os != .windows && g.pref.gc_mode in [.boehm_full, .boehm_incr,
+			.boehm_full_opt, .boehm_incr_opt, .boehm_leak] {
+			g.writeln('__attribute__ ((constructor))')
+			g.writeln('static void _v_shared_lib_gc_init(void) {')
+			// Fix for macOS hardened-runtime: libgc's trampoline allocator
+			// requests rwx pages via mprotect(PROT_EXEC), which macOS
+			// hardened runtime refuses by default. Disabling executable
+			// pages before GC_INIT() opts libgc out of trampolines (it
+			// falls back to a non-JIT mark stack). Required for V `-prod`
+			// + shared lib + macOS, and harmless on Linux.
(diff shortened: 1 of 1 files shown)


SurmanPP:
It already is emitted but into the main function. It would make more sense to just move it into the `_vinit()` function.

medvednikov:
@SurmanPP is correct.

I fixed it in https://github.com/vlang/v/commit/fa090e26a8d3225d6b2f3adc9012b1a2077dd30b

PR #27201, opened 2026-05-19
thirdparty/libgc/gc.h: use GC_word instead of u64 so header is valid standalone C

### LINKED ISSUE #27179: vlib/gc wrapper's emitted gc.h references V-only types, breaks C shims
> Reproduced on V 0.5.1 (commit 99f141f74 in upstream master);
> verified no relevant commits land between that and current
> master b6dfae560. OS: macOS 26.4.1 ARM64.

### Problem

A C shim that needs to call libgc directly (e.g.
`GC_register_my_thread`, `GC_get_stack_base`) cannot just write:

    #include "gc.h"

because V's `vlib/gc` module ships a wrapper `gc.h` that uses V-only
types (`u64`, etc.) and is not legal standalone C. Including it from a
plain `.c` file produces a cascade of unknown-type errors.

The workaround is to bypass the V wrapper by including the bundled
libgc header directly:

    #include <gc/gc.h>   /* subdirectory path skips V's gc.h */

This works but is non-obvious and fragile (depends on libgc's own
include layout, which has shifted historically).

### Reproducer

`shim.c`:

    #include "gc.h"   /* compiled in the context of a V module that imports gc */
    int dummy(void) { return 0; }

Compilation fails with "use of undeclared identifier `u64`" (or
similar) because V's wrapper assumes V's prelude.

### Proposed fix

Either:

1. Make `vlib/gc`'s emitted `gc.h` legal standalone C (replace V-only
   types with the underlying C primitives — `u64` → `uint64_t`, etc.,
   guarded by an `#ifdef __V__` so V's cgen still works).
2. Or emit it under a V-specific name (`v_gc.h`) and leave the system
   path `gc.h` to resolve to libgc's real header.

### Downstream evidence

CX's `vcx/cx/gc_thread_shim.c` has a comment block explaining the
workaround:

    /* Use the subdirectory header to bypass V's `<gc.h>` wrapper
     * (which references V-only types like `u64` and isn't legal C on
     * its own). */
    #include <gc/gc.h>

> [!NOTE]
>

eptx:
Closes #27179.

The backward-compat \`thirdparty/libgc/include/gc.h\` shipped with V's
bundled libgc cast its pointer argument to V's \`u64\` type when calling
\`GC_noop1\`. This makes the header illegal to include from standalone C
— a binding's C shim doing \`#include <gc.h>\` via V's
\`-I@VEXEROOT/thirdparty/libgc/include\` path would fail with:

    error: unknown type 'u64'

\`GC_word\` is the libgc-defined type that \`GC_noop1\` actually takes —
\`typedef\`'d to \`uintptr_t\` on most platforms in \`gc/gc.h\`. Using it
makes this header valid both:

- from V code (V knows \`GC_word\` via the
  \`fn C.GC_noop1(GC_word)\` declaration in
  \`builtin_d_gcboehm.c.v\`), AND
- from standalone C shims (which include \`gc/gc.h\` transitively and
  get the typedef).

This unblocks bindings that need to call libgc directly for thread-
base stack detection or per-thread registration in their FFI layer —
they can now use the canonical \`#include <gc.h>\` instead of the
non-obvious workaround \`#include <gc/gc.h>\` that bypasses the wrapper.

Tested by compiling a C shim with the patched header on macOS 26.4
ARM64 (clang 17):

    cat > shim.c <<'EOF'
    #include <gc.h>
    int main(void) {
        struct GC_stack_base sb;
        GC_get_stack_base(&sb);
        return 0;
    }
    EOF
    cc -I\$VEXEROOT/thirdparty/libgc/include shim.c

Compiles clean. Without the patch: \"error: unknown type 'u64'\".


medvednikov:
Ci fails

medvednikov:
`cc: /home/runner/work/v/v/thirdparty/libgc/include/gc.h:4: error: declaration expected
`

Issue #27125, opened 2026-05-10
conditional struct field


kobyzev:
### Describe the feature

```vlang
struct MainContainer {
   view MainView
$if !no_log ? {
   logger Logger
}
   dev_menu DevMenu @[if !no_dev_menu ?]
}
```
V has a good feature of conditional compilation, but I see no way to make it inside `struct` declaration.
Please, give me existing workaround if any.
I know I can wrap by $if whole struct now, but it is painful if I have multiple conditions and vls-like apps going crazy

gechandesu:
You can use `$if` on top-level for now:

```v
$if enable_bar ? {
	struct FooBar {
		foo int
		bar int
	}
} $else {
	struct FooBar {
		foo int
	}
}

fn main() {
	println(FooBar{})
}
```

```console
$ v run foobar.v
FooBar{
    foo: 0
}
$ v -d enable_bar run foobar.v
FooBar{
    foo: 0
    bar: 0
}
```

kobyzev:
I know, but in my example with MainContainer I will need more `$if` statements - not an acceptable solution

medvednikov:
Done in V2: https://github.com/vlang/v/pull/27249

PR #27249, opened 2026-05-25
v2: parser — conditional struct fields via `$if` blocks and `@[if cond ?]`

medvednikov:
## Summary

Implements #27125 for v2 (`vlib/v2`). Struct bodies now accept:

- `$if cond { ... } $else $if cond2 { ... } $else { ... }` blocks grouping fields under a compile-time condition.
- Per-field `@[if cond ?]` attribute that drops the field when the condition is false.

Example:

```v
struct MainContainer {
    view MainView
$if !no_log ? {
    logger Logger
} $else $if alt_logger ? {
    logger AltLogger
}
    dev_menu DevMenu @[if !no_dev_menu ?]
}
```

Conditions are evaluated at parse time so `StructDecl.fields` stays flat — no new AST nodes. The evaluator mirrors the transformer's `eval_comptime_cond` and supports: `Ident`, `!`, `&&`, `||`, postfix `?`, and `ParenExpr`. Recognized flags: os/arch/endian, `debug`, `native`, `prealloc`, plus any `-d` user define. `$else $if` chains route through `force_skip` so at most one branch's fields contribute.

## Test plan

- [x] `vlib/v2/transformer/transformer_test.v` — 6 new tests cover: default-branch selection, `-d` selecting the `$if` branch, omission when flag unset, `$else $if` first-match, attribute-based elision (false and true).
- [x] Existing parser/transformer test suites still pass.
- [x] Manual end-to-end: built a `MainContainer` with `$if`/`$else`/`@[if ...]` fields, confirmed correct branch and attribute behavior under cleanc and eval backends.


medvednikov:
@codex review

Issue #26967, opened 2026-04-24
[x.json2] Josn2 test error set


Jengro777:
### Describe the bug

Here is a summary of the latest version v and the bugs that appeared in testing for JSON 2

### Reproduction Steps

[json2_test_new_err.zip](https://github.com/user-attachments/files/27034051/json2_test_new_err.zip)

### Expected Behavior

success

### Current Behavior

error

### V version

V 0.5.1 b469eb9

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.1 8ef799c99cb0141a119e9e06f6b629c9e98486bf.b469eb9
|:-------------------|:-------------------
|OS                  |linux, Deepin 25
|Processor           |20 cpus, 64bit, little endian, 13th Gen Intel(R) Core(TM) i5-13500
|Memory              |13.32GB/31.03GB
|                    |
|V executable        |/home/Jengro/opt/v/v
|V last modified time|2026-04-24 00:49:09
|                    |
|V home dir          |OK, value: /home/Jengro/opt/v
|VMODULES            |OK, value: /home/Jengro/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/Jengro
|                    |
|Git version         |git version 2.51.0
|V git status        |0.5.1-1330-gb469eb9f
|.git/config present |true
|                    |
|cc version          |cc (Deepin 12.3.0-17deepin17) 12.3.0
|gcc version         |gcc (Deepin 12.3.0-17deepin17) 12.3.0
|clang version       |Deepin clang version 17.0.6 (5deepin7)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.38-6deepin21) 2.38

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Jengro777:
After these testing bugs are resolved, it can be considered that complete coverage of cjson testing in v has been achieved

PR #27252, opened 2026-05-25
x.json2: fix decode/encode regressions reported in #26967

medvednikov:
## Summary

Addresses several real bugs from the bundle of failing tests in #26967:

- **encode**: reorder `check_not_empty[T]` branches so option-aliased types (e.g. `?MyAlias`) are checked before the unaliased_typ comparisons. Previously a `?Alias` field hit the int/float branch and produced a `?Alias` vs int-literal compile error.
- **encode**: emit a `_type` discriminator on each element of array variants in user-defined sumtypes such as `type Prices = Price | []Price`. A new helper `encode_array_of_sumtype_variants` is hooked in from `encode_sumtype[T]` so the variant is no longer routed through the generic `encode_value` path that loses tagging.
- **decode**: in both struct-field code paths (with and without embedded structs), accept JSON object/array into a plain `string` field by capturing the raw JSON, and accept `null` into a non-required `string` field as an empty string.
- **decode**: accept `null` for non-option array/map fields, leaving the destination empty instead of erroring out.
- **decode_sumtype**: extract `get_array_element_type` result into a local to work around a generic codegen wrapping bug (`string_to_sumtype_Any(...)` was being emitted around the call).

Remaining failures in the issue bundle are pre-existing differences between legacy `json` and `x.json2` (error-message formatting, no Unix-epoch encoding for `time.Time`, no `\uXXXX` escapes) plus one separate V compiler bug for `?[N]T` generic struct fields. Those are intentionally not in scope here.

## Test plan
- [x] `./v test vlib/x/json2/` — 56 passed, 1 pre-existing skip, 0 failed
- [x] `checker_test.v` still errors on top-level `decode[string]('{...}')` (no regression to top-level non-string-into-string path)
- [x] New tests pass: `json_option_alias_test.v`, `json_decode_with_sumtype_test.v`, `json_sumtype_test.v`, `json_prim_type_validation_test.v::test_decode_object_into_string_field`, `4_json_test.v` non-required-null path

### DIFF
--- a/vlib/x/json2/decode.v
+++ b/vlib/x/json2/decode.v
@@ -566,10 +566,21 @@ fn decode_struct_key[T](mut decoder Decoder, val T, key_info ValueInfo, prefix s
 							}
 						}
 					} $else $if field.unaliased_typ is $array_dynamic {
-						new_val.$(field.name).clear()
-						decoder.decode_array(mut new_val.$(field.name))!
+						if decoder.current_node.value.value_kind == .null
+							&& !field_info.is_required {
+							new_val.$(field.name).clear()
+							decoder.skip_current_value()
+						} else {
+							new_val.$(field.name).clear()
+							decoder.decode_array(mut new_val.$(field.name))!
+						}
 					} $else $if field.unaliased_typ is $map {
-						decoder.decode_map(mut new_val.$(field.name))!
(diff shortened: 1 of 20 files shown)


Jengro777:
We should add testing

medvednikov:
Done :)

PR #27262, opened 2026-05-25
v2: link minimal Windows x64 PE executables hello world!

GGRei:
## Summary

This PR extends the V2 x64 backend with minimal Windows PE/COFF executable support.

It builds on the existing Linux ELF and macOS Mach-O object format work, and adds the Windows side of the same native x64 backend path.

The current scope is intentionally small: generate COFF objects, apply the Windows x64 ABI, and link a minimal PE executable internally.

## What Works

This PR adds:

- COFF object emission for the V2 x64 backend;
- Windows x64 ABI lowering/codegen support;
- a minimal internal PE32+ linker for AMD64;
- kernel32 import table / IAT generation;
- support for linking:
  - `fn main() {}`
  - `fn main() { println('x') }`

The `println` path avoids depending on CRT stdio globals for the minimal PE backend. It writes through a small kernel32-based path, preserves the console Unicode path via `WriteConsoleW`, and falls back to `WriteFile` for redirected output while handling partial writes.

## Scope

This is not a full Windows runtime implementation yet.

The PE linker currently stays deliberately minimal:

- kernel32 imports only;
- fixed-base image;
- no `.reloc` yet;
- no `.pdata` / `.xdata` unwind metadata yet;
- no broad CRT/UCRT/MSVCRT import layer.

The goal is to establish the first working Windows x64 executable path for the V2 backend, equivalent in spirit to the existing minimal Linux/macOS native backend milestones.

## Non-Regression Guards

The patch includes tests to ensure:

- Windows x64 selects COFF + Windows ABI;
- Linux x64 remains ELF + SysV;
- macOS/darwin x64 remains Mach-O + SysV;
- target OS file filtering does not accidentally import Windows files on Linux/macOS;
- conditional imports respect the requested target OS;
- imported module `init()` and runtime const initialization are preserved for direct, conditional, and dotted imports;
- the Windows minimal runtime path is enabled only for Windows x64;
- the minimal Windows `println` path does not retain broad runtime modules such as `os`, `time`, `io`, `dl`, `sha256`, or `binary`;
- redirected Windows output handles partial `WriteFile` writes without truncating output.

## Validation

Run locally on Linux:

- `./v test vlib/v2/builder/native_test.v vlib/v2/builder/target_os_test.v`
- `./v test vlib/v2/builder`
- `./v test vlib/v2/gen/x64`
- `./v test vlib/v2/pref/comptime_test.v`
- `./v test vlib/v2/transformer`
- `./v fmt -verify ...`
- `git diff --check`

All targeted checks pass locally.

A real Windows smoke test is intentionally left to GitHub/CI or a Windows runner: compile and run a minimal `println("Hello World!")` program with the V2 x64 backend and verify exit code/output.


medvednikov:
You're on fire today! :)

PR #27266, opened 2026-05-25
v2: add module-scoped global variables

### LINKED ISSUE #22714: Module Variables
### Describe the feature

Weeks ago, on V's Telegram channel, I had proposed the idea of ​​introducing module variables.

The objective of this idea is to remove `__global` from the compiler completely, since module variables could replace its function (in a safer and more idiomatic way).

Global variables in V require a flag to work (`-enable-globals`) and can be accessed from anywhere in the code, plus they are always mutable, making them somewhat unsafe to use.

Additionally, with the implementation of module variables, you could make `const` stop receiving values ​​that are computed in runtime, because `static` would fulfill that function perfectly. Then `const` would receive pure values ​​that can be computed in comptime.

My suggestion is to use the `static` keyword (already used to declare static variables inside functions) to declare module variables, which can be declared as `pub` and `mut` as needed:

```v
// report/report.v
module report

pub static mut errors = int(0)
```

```v
// main.v
module main

import report

fn main() {
    report.errors += 1
}
```

### Use Case

In addition to what has already been said, the idea is also to make a distinction for constants, which currently receive all types of values ​​(both runtime and comptime), which could confuse new users who are learning V, but come from other languages ​​where `const` receives pure values ​​that are computed in comptime.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

- [X] This feature might incur a breaking change

### Version used

latest

anywhere

GGRei:
## Summary

This PR implements V2 module-scoped global variables using the existing `__global` spelling.

Fix #22714 

The goal is to turn `__global` into a safer module-owned storage declaration for V2, instead of treating it like the old unrestricted global-variable feature. A `__global` declaration now belongs to its declaring module, is private by default, and can only be accessed from another module when explicitly marked `pub`.

## What This Enables

This allows modules to keep internal mutable state without exposing unrestricted process-wide globals.

Example:

    module state

    __global mut counter = 0

    pub fn next() int {
        counter++
        return counter
    }

    pub __global mut public_counter = 10

From another module:

    import state

    fn main() {
        println(state.next())

        state.public_counter++
        println(state.public_counter)
    }

Private module storage remains private:

    import state

    fn main() {
        println(state.counter) // rejected: counter is private
        counter++              // rejected: imported module storage is not visible bare
    }

## Scope

This PR is intentionally limited to the V2 compiler path.

It does not change V1 behavior, and it does not rework regular `const` declarations. The implementation focuses on module-owned storage semantics for `__global`.

The spelling is also kept isolated in the parser/checker path so it can be changed later if the project decides that another keyword, such as `static`, is a better long-term surface syntax.

## Behavior Rules

This PR implements the following rules:

- `__global name = value` declares module-owned storage.
- `__global mut name = value` is accepted and records explicit mutability.
- `pub __global name = value` exposes module storage to qualified external access.
- `pub __global mut name = value` exposes mutable module storage explicitly.
- module storage is private by default.
- external access must be qualified through the declaring module.
- bare access to imported module storage is rejected.
- generated C names are module-mangled, so two modules can declare storage with the same source name without colliding.
- C extern globals such as `C.errno` and stdio globals keep their existing behavior.
- existing C attributes such as `@[c_extern]`, `export`, `weak`, `hidden`, and `markused` are preserved.

## Compatibility Note

The current patch does not make `mut` mandatory yet for mutation of legacy `__global name` declarations.

That is deliberate: the PR supports and propagates explicit `__global mut`, but does not immediately turn older `__global name` code into hard errors. This keeps the first V2 step focused on module ownership and visibility. The parser/checker now carries the mutability metadata, so stricter enforcement can be enabled later without redesigning the feature.

In V2, `-enable-globals` is still accepted only for command-line compatibility with existing scripts and test runners. It does not enable the old unrestricted global-variable behavior. The new `__global` semantics remain the same with or without that flag: module-owned storage, private by default, and externally accessible only through qualified `pub` module access.

## Implementation Notes

The patch updates the V2 pipeline end to end:

- AST storage metadata;
- parser support for `pub __global`, `__global mut`, and `pub __global mut`;
- checker visibility rules for module-owned storage;
- C name mangling for module-scoped storage;
- CleanC generation;
- V generator output;
- SSA propagation;
- live transformer propagation;
- AST dump support;
- builder/cache header handling;
- regression tests and a small example fixture.

## Tests

Added coverage for:

- private module-scoped storage;
- public module-scoped storage;
- qualified external access;
- rejection of private qualified external access;
- rejection of bare imported module storage access;
- same-name storage across multiple modules;
- explicit `__global mut` parsing/propagation;
- C extern global preservation;
- generated C module name mangling;
- SSA storage propagation;
- cache/header handling;
- final usage example fixture under `vlib/v2/tests/module_storage_example`.

Validation run locally after syncing with latest `master`:

- `/usr/local/bin/v -o ./v cmd/v`
- `./v test vlib/v2/types/module_storage_test.v vlib/v2/ssa/module_storage_test.v vlib/v2/gen/cleanc/module_storage_codegen_test.v vlib/v2/builder/module_storage_cache_test.v`
- `./v test vlib/v2/gen/cleanc`
- `./v test vlib/v2/ssa`
- `./v test vlib/v2/transformer`
- `./v test vlib/v2/builder`
- `./v fmt -verify` on modified V files
- `git diff --check`

All targeted checks pass locally.

I also ran the broader informative check `./v test vlib/v2/types`. It currently still fails in `checker_escape_test.v` and `checker_ownership_test.v`, but the same failures reproduce on a clean `origin/master` worktree at the same commit, without this patch applied. So those failures appear to be pre-existing V2 issues, not caused by this PR.

### DIFF
--- a/vlib/v2/ast/ast.v
+++ b/vlib/v2/ast/ast.v
@@ -405,6 +405,8 @@ pub:
 	typ        Expr = empty_expr // can be empty as used for const (unless we use something else)
 	value      Expr = empty_expr
 	attributes []Attribute
+	is_public  bool
+	is_mut     bool
 }
 
 pub struct FieldInit {
@@ -908,6 +910,7 @@ pub struct GlobalDecl {
 pub:
 	attributes []Attribute
 	fields     []FieldDecl
+	is_public  bool
 }
 
 pub struct ImportStmt {
--- a/vlib/v2/ast_dump/ast_dump.v
+++ b/vlib/v2/ast_dump/ast_dump.v
@@ -621,6 +621,9 @@ fn (mut jb JsonBuilder) write_global_decl(stmt ast.GlobalDecl) {
 	jb.sb.write_string(',\n')
 
 	jb.write_indent()
+	jb.sb.write_string('"is_public": ${stmt.is_public},\n')
+
+	jb.write_indent()
 	jb.sb.write_string('"fields": ')
 	jb.write_field_decls(stmt.fields)
 	jb.sb.write_string('\n')
@@ -1279,6 +1282,13 @@ fn (mut jb JsonBuilder) write_field_decl(field ast.FieldDecl) {
 	jb.write_indent()
 	jb.sb.write_string('"attributes": ')
 	jb.write_attributes(field.attributes)
+	jb.sb.write_string(',\n')
+
+	jb.write_indent()
+	jb.sb.write_string('"is_public": ${field.is_public},\n')
+
+	jb.write_indent()
+	jb.sb.write_string('"is_mut": ${field.is_mut}')
 	jb.sb.write_string('\n')
 
 	jb.indent--
--- a/vlib/v2/builder/cache_headers.v
+++ b/vlib/v2/builder/cache_headers.v
@@ -1540,12 +1540,15 @@ fn (b &Builder) build_header_ast_for_files(source_files []ast.File, module_name 
 							name:       field.name
 							typ:        global_typ
 							attributes: field.attributes
+							is_public:  field.is_public
+							is_mut:     field.is_mut
 						}
 					}
 					if gfields.len > 0 {
 						decl_stmts << ast.Stmt(ast.GlobalDecl{
 							attributes: stmt.attributes
 							fields:     gfields
+							is_public:  stmt.is_public
 						})
 					}
 				}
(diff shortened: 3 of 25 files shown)


GGRei:
Well. It's ready now. ^^

medvednikov:
Thanks!

Wajinn:
I find this a bit unsettling, though it looks to be a well done PR. Kind of feels like a backdoor to excessive usage of global variables. It can be like an addiction. Hopefully the flag stays off for normal usage.

PR #27268, opened 2026-05-26
v2: add module_mut struct fields

### LINKED ISSUE #9894: Access modifier pub mut allows modification outside parent module
V full version: V 0.2.2 bca9409.3877522
OS: windows, Microsoft Windows 7 Professional v7601 64-bit

**What did you do?**
Created a file foo.v
~~~v
module main
import bar

fn main() {
	mut x := bar.Bar{}
	x.i += 1
	println(x.i)
	x.test()
	println(x.i)
}
~~~

in addition a directory bar with a file bar.v

~~~v
module bar

pub struct Bar {
pub mut:
	i int
}

pub fn (mut b Bar) test() {
	b.i++
}
~~~

Compiler error reporting that field "i" is immutable
 
**What did you see instead?**
No error, "i" could be changed.

GGRei:
## Summary

This PR adds support for module-mutable struct fields through the `@[module_mut] pub:` access block.

Fix #9894 

The goal is to address the case where a field should be publicly readable, but only mutable from inside the module that declares the struct.

Example:

```v
module state

pub struct Counter {
@[module_mut]
pub:
	value int
}

pub fn (mut c Counter) inc() {
	c.value++
}
```

From another module:

```v
import state

fn main() {
	mut c := state.Counter{}
	println(c.value) // allowed
	c.inc()          // allowed, mutation happens inside module `state`
	c.value++        // rejected
}
```

## Behavior

`@[module_mut] pub:` means:

- the field is public for reading;
- the field may be mutated inside the declaring module;
- external modules cannot mutate it directly;
- `pub mut:` keeps its existing behavior and remains publicly mutable.

This is implemented only in V2. V1 is not changed.

## Invalid Uses

The parser rejects invalid uses of `@[module_mut]` explicitly.

Rejected cases include:

- use outside struct fields, such as on `fn`, `struct`, `enum`, `interface`, `const`, `__global`, or file/module level;
- use on private fields or `mut:`;
- use together with `pub mut:`;
- use on embedded fields;
- use in unions or C structs;
- arguments such as `@[module_mut()]` or `@[module_mut: true]`;
- duplicate `@[module_mut]`.

## Mutation Checks

External mutation is rejected for:

- direct assignment;
- compound assignment;
- postfix `++` / `--`;
- array/map-style mutation through the field;
- nested mutation through the field;
- passing the field as `mut`;
- taking a mutable reference to the field;
- struct initialization from another module;
- assoc/struct update from another module;
- calling a `mut` receiver method on a module-mutable field from another module.

## Implementation Notes

The patch adds field metadata to V2 AST/type information so the checker can preserve:

- field visibility;
- mutability;
- module-mutability;
- declaring module ownership.

It also preserves this metadata through the relevant V2 paths, including cache headers, gen/v output, transformer copies, and generic field copies.

The checker uses a central field access helper to resolve the field and its owner before deciding whether an external mutation should be rejected.

## Tests

A dedicated V2 regression test was added:

- `vlib/v2/types/module_mut_field_test.v`

It covers:

- allowed external reads;
- allowed internal mutation through module-owned methods;
- unchanged `pub mut:` behavior;
- parser metadata for `mut:`, `pub:`, `@[module_mut] pub:`, and `pub mut:`;
- all supported external mutation rejection paths;
- invalid attribute usages.


GGRei:
Ready ^

medvednikov:
how about `pub module_mut:` ?

JalonSolov:
Shorter, cleaner.

The attribute could be for individual fields...
```v
pub struct Counter {
pub mut:
	x int @[module_mut]
	y int
	z int
}
```
That would say that `y` and `z` were modifiable outside the module, but `x` was only modifiable inside.  Without the attribute, it would be longer...
```v
pub struct Counter {
pub module_mut:
	x int
pub mut:
	y int
	z int
}
```

GGRei:
Ok, but then which direction should I implement?

`pub module_mut:` would make this no longer an attribute-based feature, but it changes the design a bit.

From your last two comments I see two possible directions:

1. Use only `pub module_mut:` as the official syntax.
2. Support `pub module_mut:` for blocks, and keep `@[module_mut]` only for individual fields inside other sections like `pub mut:`.

Both make sense, but they are not exactly the same design. I can adapt the PR either way.

What should I do?

medvednikov:
It's longer, but the grouping is nice, it's consistent, and it's one way.

So i'd only have this:


```
pub struct Counter {
pub module_mut:
	x int
pub mut:
	y int
	z int
}
```

medvednikov:
having `@[module_mut]` is also inconsistent with the current design, because in V immutable is default, here we have to explicitly say it's immutable outside the module.

medvednikov:
But it does look very ugly to be honest :(

```
pub struct Counter {
pub module_mut:
	x int
pub mut:
	y int
	z int
}
```

This entire grouping thing is the only thing I don't like in V's syntax, and I don't know how to solve it.

medvednikov:
In some cases the order of the fields is important, and we can't change them because of the grouping.

GGRei:
No worries, :) take your time to think about the syntax/design.

I’ll put this PR on standby for now, since the implementation depends on the final direction here.

In the meantime I’ll continue working on autofree and the x64 backend for V2.

Wajinn:
> But it does look very ugly to be honest :(

```
pub struct Counter {
pub module_mut:
	x int
pub mut:
	y int
	z int
}
```
Alternative:

```
pub struct Counter {
pub mut_mod:
	x int
pub mut:
	y int
	z int
}
```
As a user coming across the language, think it would look better to be shorter. Maybe `mut_mod` or `mod_mut`. But think `mut_mod`, as easier to remember as `mutable` plus `something else`. Kind of goes better with `mut` below it.

gechandesu:
Rust has a `pub(super)` syntax, which changes the normal `pub` meaning to "visible only to the parent module." V already has something similar for the `defer` keyword — the `defer(fn) {}` block will be executed on function exit, not in the current scope as `defer {}`.

A similar syntax could be implemented for `module_mut`. In my opinion, it doesn't look ugly and clearly indicates some different behavior for this `mut`.

```
pub struct Counter {
pub mut(module):
    x int
pub mut:
    y int
    z int
}
```

JalonSolov:
Kicking the CI tires...

medvednikov:
Ok, I made the decision.

```
pub struct Counter {
pub module_mut:
	x int
pub mut:
	y int
	z int
}
```

mod_mut or mut_mod are not as readable, because both are 3 letter words staring with m

Wajinn:
Looks too late, but maybe reconsidered in the future; liked gechandesu's suggestion (`mut(module)`). Something about having mut in the front, seems to flow well.

GGRei:
## Summary

The V2 module-mutable field proposal to use the approved syntax:

    pub struct Counter {
    pub module_mut:
        value int
    pub mut:
        label string
    }

pub module_mut: means the field is publicly readable, but mutable only from inside the module that declares the struct.

This replaces the previous attribute-based proposal. 

## Behavior

This PR keeps the intended semantics:

- pub: remains publicly readable and externally immutable.
- pub mut: remains publicly readable and publicly mutable.
- pub module_mut: is publicly readable, but mutable only inside the declaring module.
- Public methods in the declaring module can still mutate the field internally.

Example:

    module state

    pub struct Counter {
    pub module_mut:
        value int
    pub mut:
        label string
    }

    pub fn (mut c Counter) inc() int {
        c.value++
        return c.value
    }

From another module:

    mut c := state.Counter{}
    println(c.value) // allowed
    c.inc()          // allowed
    c.value++        // rejected
    c.label = 'x'    // allowed, because label is pub mut

## Implementation Notes

The patch:

- parses pub module_mut: as a contextual struct-field access block;
- does not add module_mut as a global token or keyword;
- removes the previous attribute-specific handling;
- preserves field metadata through AST/type/checker paths;
- preserves module ownership metadata for mutation checks;
- updates gen/v to emit pub module_mut:;
- adds cache/header coverage to ensure pub module_mut: survives generated headers.

## Mutation Checks

External mutation of pub module_mut: fields is rejected for:

- direct assignment;
- compound assignment;
- postfix ++ / --;
- array/map-style mutation through the field;
- nested mutation through the field;
- passing the field as mut;
- taking a mutable reference to the field;
- struct initialization from another module;
- assoc/struct update from another module;
- calling a mut receiver method on a module-mutable field;
- mutable for-in iteration through a module-mutable collection field.

## Tests

This PR updates and extends the V2 regression coverage in:

- vlib/v2/types/module_mut_field_test.v
- vlib/v2/builder/module_storage_cache_test.v

Coverage includes:

- parser metadata for pub module_mut:;
- if/else and else-if access-state propagation;
- external read allowed;
- internal mutation allowed;
- pub mut: behavior unchanged;
- import alias handling;
- contextual use of module_mut as a normal identifier outside access blocks;
- gen/v round-trip output;
- cache/header preservation;
- all main external mutation rejection paths.

A small readable V2 example was also added:

- vlib/v2/tests/module_mut_example/

GGRei:
Ready.

PR #27295, opened 2026-05-30
cgen: clone comptime selector assignments

### LINKED ISSUE #27292: autofree: missing array/map clone for ComptimeSelector inside $for/$if blocks - use-after-free
When a local variable is assigned from a struct field inside a `$for`/`$if` compile-time reflection block, V's autofree mode generates `Array_string_free(&arr)` without first cloning the array. Since `arr` is a shallow reference to the struct field's data, this frees memory the struct still references — a use-after-free.

Outside `$for`/`$if` blocks, V correctly inserts `builtin__array_clone_static_to_depth(val.tags, 1)` before the assignment. Inside `$for`/`$if`, the clone is skipped because `gen_clone_assignment()` in `vlib/v/gen/c/cgen.v` only handles `ast.Ident` and `ast.SelectorExpr`, but inside comptime blocks the field access becomes an `ast.ComptimeSelector` — which falls through the check.

### Root cause

[`vlib/v/gen/c/cgen.v`](https://github.com/vlang/v/blob/master/vlib/v/gen/c/cgen.v), line 3710:

```v
fn (mut g Gen) gen_clone_assignment(var_type ast.Type, val ast.Expr, typ ast.Type, add_eq bool) bool {
    if val !in [ast.Ident, ast.SelectorExpr] {
        return false  // ComptimeSelector not handled, clone skipped
    }
    right_sym := g.table.sym(typ)
    if g.is_autofree {
        // ... clone logic ...
    }
    return true
}
```

Inside `$for`/`$if` blocks, `val.$(field.name)` is represented as `ast.ComptimeSelector`, which is not in the `[ast.Ident, ast.SelectorExpr]` set, so `gen_clone_assignment` returns `false` and no clone is inserted.

### Possible fix

In `vlib/v/gen/c/cgen.v`, line 3710, change:

```v
if val !in [ast.Ident, ast.SelectorExpr] {
```

to:

```v
if val !in [ast.Ident, ast.SelectorExpr, ast.ComptimeSelector] {
```

`ComptimeSelector` generates the same C output as `SelectorExpr` (a field access like `val.tags`), so `g.expr(val)` already produces the correct code. The function just needs to accept it at the gate.

There probably also needs to be handling for `ComptimeSelector` in the clone body — `g.expr(val)` on a `ComptimeSelector` emits the field access correctly, so no additional changes may be needed in the clone body itself.

### Verification

```bash
v run repro2.v          # GC mode — works fine
v -autofree run repro2.v # autofree — segfault (before fix)
# After fix:
v -autofree run repro2.v # prints correct output, no crash
```

### Reproducer (`repro2.v`)

```v
module main

struct Item {
    name string
    tags []string
}

fn process[T](val T) string {
    mut result := '['
    $for field in T.fields {
        $if field.unaliased_typ is []string {
            arr := val.$(field.name)
            for i, tag in arr {
                if i > 0 {
                    result += ','
                }
                result += '"${tag}"'
            }
        }
    }
    result += ']'
    return result
}

fn main() {
    item := Item{name: 'test', tags: ['a', 'b', 'c']}
    println('before: ${item.tags}')
    result := process(item)
    println('result: ${result}')
    println('after:  ${item.tags}')
}
```

### Generated C comparison

Without fix (`$for`/`$if` block — **no clone**):
```c
Array_string arr = val.tags;  // shallow reference!
// ... loop body ...
Array_string_free(&arr); // autofreed var — frees data val.tags still references!
```

With fix (same as outside `$for`/`$if` — **clone inserted**):
```c
Array_string arr = builtin__array_clone_static_to_depth(val.tags, 1);  // deep copy
// ... loop body ...
Array_string_free(&arr); // safe — arr owns its own data
```

Outside `$for`/`$if` blocks (already works correctly):
```c
Array_string tags = builtin__array_clone_static_to_depth(val.tags, 1);
```

medvednikov:
Fixes #27292.

## What changed

- Allow `ast.ComptimeSelector` expressions through `gen_clone_assignment`, so autofree emits clones for comptime-selected string and array fields before assigning them to locals.
- Apply the normal map-copy checker diagnostic to comptime-selected map fields instead of letting them shallow-copy and crash later under autofree.
- Add C output and checker regression coverage for both cases.

## Validation

- `VTEST_ONLY=comptime_selector_map_copy_err ./vnew -silent vlib/v/compiler_errors_test.v`
- `VTEST_ONLY=autofree_comptime_selector_clone ./vnew -silent vlib/v/gen/c/coutput_test.v`
- `./vnew -silent vlib/v/compiler_errors_test.v` (`1572 passed, 5 skipped`)

Note: a later broad `./vnew -silent test vlib/v/` attempt in the dirty local checkout was stopped after unrelated failures in untracked eval/interpreter work and local bundled `tcc` header resolution.


medvednikov:
@codex review

PR #27297, opened 2026-05-30
v2: flat ast and cleanc fixes

medvednikov:
## Summary

- Adds flat AST storage/cursor support and wires v2 parser, transformer, markused, checker, and SSA builder paths to consume flat files where available.
- Ports the SSA builder through flat-native function body, statement, and expression paths, with parity tests for the new cursor lowering helpers.
- Fixes cleanc generic/option/result/array/comptime metadata cases needed for the full v2 test suite and the Gitly build.

## Remaining Flat AST Migration Work

- SSA still has several cursor helpers that rehydrate nested operands with `decode_expr` where legacy lowerers require full AST pattern matching.
- Some SSA registration and markused top-level declaration paths still decode selected declarations instead of consuming all fields directly from cursors.
- Cleanc and transformer have flat-aware paths, but complete migration would mean removing the remaining legacy AST fallback paths after coverage and parity tests are stronger.

## Validation

- `./v -o cmd/v2/v2 cmd/v2/v2.v`
- Focused transformer and cleanc regression tests
- `./vnew run cmd/v2/test_all.vsh`


medvednikov:
Closed in favor of #27299, which is based on current master and keeps the clean five-commit stack.

PR #27314, opened 2026-06-01
v2: implement native x64 backend, part 1b

GGRei:
## Summary

This PR implements part 1b of the V2 native x64 backend work: reducing the binary footprint for small native programs while keeping normal fallback paths intact.

It adds tiny output paths for the currently supported V2 x64 examples:

- examples/hello_world.v
- examples/fizz_buzz.v

The optimization is automatic when the program is eligible, and falls back to the normal native path when it is not.

## What changed

- Reduced Linux x64 executable size with a tiny ELF path for eligible minimal programs.
- Reduced macOS x64 executable size with an automatic tiny Mach-O object path for eligible programs.
- Reduced Windows x64 executable size by trimming unused globals and keeping PE imports focused on the required Kernel32 symbols.
- Kept documented platform behavior:
  - Linux uses direct syscalls where practical.
  - macOS still links through the normal system linker/libSystem path.
  - Windows uses documented Win32 imports instead of unsupported syscall tricks.
- Added `-no-mos-tiny` as a macOS-only opt-out for comparing against the normal Mach-O/linker path and debugging dead-strip related behavior. (`ld -dead_strip`, documented by Apple).
- Added regression and runtime smoke coverage for Linux, macOS x64, and Windows x64 paths.

## Validation

Locally tested:

- git diff --check
- v fmt -verify on touched V files
- V2 x64 targeted tests
- V2 x64 runtime smoke test

Also validated through the isolated CI footprint workflow on Linux, macOS Intel, and Windows.

### DIFF
--- a/cmd/v/vvm_test.v
+++ b/cmd/v/vvm_test.v
@@ -199,6 +199,36 @@ fn test_v2_delegation_forwards_target_specific_flags() {
 	assert marker.contains('main.v'), marker
 	assert !marker.contains('-v2'), marker
 
+	res_macos := run_v_command_with_exe(wrapper_exe, project_dir, [
+		'-v2',
+		'-b',
+		'x64',
+		'-os',
+		'macos',
+		'-no-mos-tiny',
+		'-o',
+		'out',
+		'main.v',
+	], envs)
+	assert res_macos.exit_code == 0, res_macos.output
+	macos_marker := os.read_file(marker_file) or { panic(err) }
+	assert macos_marker.contains('-b x64'), macos_marker
+	assert macos_marker.contains('-os macos'), macos_marker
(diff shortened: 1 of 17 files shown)


GGRei:
@medvednikov  I also ran the requested command:

    cd cmd/v2
    sh test_all.sh

On my comp it currently stops at step 1/15: ARM64 self-host hello world.

The failure happens because the test produces a Mach-O 64-bit arm64 object, then the local Linux linker rejects it:

    main.o: file not recognized: file format not recognized

GGRei:
@codex review

GGRei:
Ready

medvednikov:
@GGRei I see. It should be trivial to update the test_all.vsh V program so that it accepts an arch parameter and uses that to test.

PR #27320, opened 2026-06-01
compress.gzip: use deflate for RFC compliance

JalonSolov:
Deprecated a bunch of unused stuff, including the entire `compress.c.v` file.


medvednikov:
@JalonSolov the goal is to remove all top level files in `compress/`, right?

So that it's like in Go:

image

by the way Go doesn't have deflate for some reason

JalonSolov:
Go has `flate` rather than `deflate`.  No idea why they didn't use the "standard" name, but it is the same general code.

And yes, there is only one top-level file under `compress` right now (other than the README.md in there), and it is a C wrapper that's no longer needed after this code is merged.

I left it there with all the deprecates so it doesn't disappear on people, but it does need to go away eventually.

JalonSolov:
Bouncy-bouncy...

PR #27324, opened 2026-06-02
orm: add DataScope support for per-instance request-level filtering

### LINKED ISSUE #27154: [Feature]  ORM DataScope: Request-Level Data Scope as a Replacement for Global `@[tenant_filter]`
## Problem

The current `@[tenant_filter]` ORM feature stores tenant state in a mutable
`__global` variable. This causes several concrete problems:

1. **Race conditions**: concurrent requests/goroutines write to the same global
   state without locks or TLS.
2. **No request isolation**: one request's tenant id can leak into another
   request in web servers.
3. **Single-purpose design**: it only supports one tenant field, while real
   applications often need multiple automatic filters, such as:
   - `tenant_id = ?`
   - `org_id = ?`
   - `shop_id = ?`
   - `deleted_at IS NULL`
4. **Backend coupling**: every database backend has to know about tenant
   filtering.

## Proposal: `orm.DB` + `DataScope`

Introduce `orm.DB`, a small wrapper around any `orm.Connection`.

`orm.DB` carries a `DataScope`, and automatically applies its filters to ORM
queries before delegating to the wrapped connection.

```v
mut db := orm.new_db(pool, orm.DataScope{
	filters: [
		orm.QueryFilter{
			field: 'tenant_id'
			value: orm.Primitive(5)
		},
	]
})
```

Then normal `sql` blocks work unchanged:

```v
users := sql db {
	select from User
}!
```

The query is automatically scoped:

```sql
SELECT ... FROM user WHERE tenant_id = ?
```

## Core API

```v
pub enum QueryFilterMode {
	static
	dynamic
}

pub struct QueryFilter {
pub:
	field    string
	value    Primitive
	operator OperationKind = .eq
	mode     QueryFilterMode = .static
}

pub struct DataScope {
pub:
	enabled bool = true
	filters []QueryFilter
}

pub struct DB {
mut:
	conn Connection
pub:
	scope           DataScope
	skip_all_scopes bool
	skip_fields     []string
}

pub fn new_db(conn Connection, scope DataScope) DB

pub fn (db DB) unscoped(unscoped_fields ...string) DB
```

## Static By Default, Dynamic When Needed

Most scope filters have a fixed SQL shape. For example, `tenant_id = ?` should
be known from the scope definition and can be treated as static.

```v
orm.QueryFilter{
	field: 'tenant_id'
	value: orm.Primitive(tenant_id)
}
```

The value can still be runtime data. The static part is the filter shape:
field + operator.

For filters that need runtime behavior, use `mode: .dynamic`:

```v
mut db := orm.new_db(pool, orm.DataScope{
	filters: [
		orm.QueryFilter{
			field: 'tenant_id'
			value: orm.Primitive(tenant_id)
		},
		orm.QueryFilter{
			field: 'shop_id'
			value: orm.Primitive(shop_id)
			mode: .dynamic
		},
	]
})
```

This allows mixed scopes:

- `tenant_id` can be handled as a stable/default filter.
- `shop_id` can be applied dynamically based on request role, shop access, or
  middleware logic.

## Bypassing Scope Filters

`db.unscoped()` returns a new `orm.DB` value. It does not mutate the original DB.

Skip all filters:

```v
unscoped_db := db.unscoped()
```

Skip one filter:

```v
unscoped_db := db.unscoped('tenant_id')
```

Skip multiple filters:

```v
unscoped_db := db.unscoped('tenant_id', 'shop_id')
```

This works for `select`, `insert`, `update`, and `delete`.

## Table-Level Opt-Out

A table can opt out of all automatic scope filters:

```v
@[unscoped]
struct Config {
	key string
	value string
}
```

Queries for this table ignore `DataScope`.

## Multi-Level Tenancy Example

```v
mut db := orm.new_db(pool, orm.DataScope{
	filters: [
		orm.QueryFilter{
			field: 'tenant_id'
			value: orm.Primitive(5)
		},
		orm.QueryFilter{
			field: 'shop_id'
			value: orm.Primitive(shop_id)
			mode: .dynamic
		},
		orm.QueryFilter{
			field: 'deleted_at'
			value: orm.Primitive(orm.Null{})
			operator: .is_null
		},
	]
})
```

Generated behavior:

```sql
SELECT ... WHERE tenant_id = ? AND shop_id = ? AND deleted_at IS NULL
UPDATE ... WHERE user_condition AND tenant_id = ? AND shop_id = ? AND deleted_at IS NULL
DELETE ... WHERE user_condition AND tenant_id = ? AND shop_id = ? AND deleted_at IS NULL
```

For inserts, scope fields are auto-filled when they are not explicitly provided.

## Transaction Support

`orm.DB` also implements `orm.TransactionalConnection` when the wrapped
connection supports transactions:

```v
db.orm_begin()!
sql db {
	insert User{name: 'new'}
}!
db.orm_commit()!
```

Scope filters continue to apply inside transactions.

### No global state

Scope data lives on the `orm.DB` value, so each request can carry its own scoped
connection.

### No changes to the SQL DSL

`sql db { ... }` already works with any value implementing `orm.Connection`.
`orm.DB` implements the same interface, so the compiler-facing DSL does not need
a new syntax.

### Supports more than tenant filtering

The same mechanism supports tenancy, soft delete, row-level access control, and
other automatic filters.

### Explicit escape hatch

`db.unscoped()` gives a controlled way to bypass all or selected filters without
temporarily mutating global state.

## Expected Benefits

- Thread-safe request-level scope isolation
- Multiple filters instead of one hardcoded tenant field
- Works with existing `sql` blocks
- Keeps raw database connections backward-compatible
- Allows static default filters and explicit dynamic filters
- Removes tenant-filter-specific logic from individual DB backends

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### Version used

V 0.5.1 e02966e

|V full version      |V 0.5.1 d45d40e98dcc1e9a2deb8d2b09ed0ced2c4fc5f7.e02966e
|Memory              |3.29GB/31.03GB
|                    |
|V last modified time|2026-05-12 07:12:05
|                    |
|                    |
|V git status        |0.5.1-1548-ge02966e4
|.git/config present |true
|                    |
|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.38-6deepin21) 2.38
> [!NOTE]
>

Jengro777:
fix https://github.com/vlang/v/issues/27154

Replaces the global `configure_tenant_filter`/`set_current_tenant_id` mechanism with a per-instance `DataScope` decorator pattern. The new `orm.DB` wraps any `orm.Connection` and injects scope filters into all CRUD operations, eliminating global state and enabling thread-safe multi-tenant isolation. Works with both `sql` block syntax and `orm_func` (QueryBuilder).

---

### Key Changes

**New file: `vlib/orm/orm_scope.v`**
- `QueryFilterMode` enum — `.unset` (zero value, must be explicit), `.static` (reserved), `.dynamic` (runtime-applied)
- `DataScope` — holds a set of `QueryFilter` conditions
- `QueryFilter` — individual filter (field, value, operator, mode)
- `DB` — decorator implementing `orm.Connection` + `TransactionalConnection`, injects scope filters before delegating to underlying connection
- `apply_scope_filters` / `apply_scope_insert_filters` — WHERE and INSERT scope injection with batch support
- `unscoped()` — skip all or selected scope filters
- `gen_qualified_field` — table-qualified column names for JOIN disambiguation

**`vlib/orm/orm.v`**
- Restored deprecated global tenant filter APIs for backward compatibility (original layout preserved)
- `gen_where_clause` refactored to use `data_idx` counter, fixing unary operator (`IS NULL`) + array (`IN`) ordering
- Table types (`Table`, `SelectConfig`) extended with `fields`/`columns` metadata

**`vlib/orm/orm_func.v`**
- `table_from_struct` now populates `Table.fields` and `Table.columns`, enabling `QueryBuilder` (orm_func) to fully support DataScope

**`vlib/v/gen/c/orm.v`**
- `orm_table_field_names` / `orm_table_column_names` — recursive struct field collection at compile time
- `write_orm_table_struct` — generates `Table.fields`/`Table.columns` C initializers

**`vlib/orm/orm_scope_test.v`** — 1922 lines of tests covering:
- select / insert / update / delete with scope
- batch insert (overwrite, append, multi-field)
- `unscoped()` (selective skip, skip all)
- `@[unscoped]` table attribute
- JOIN + scope (table-qualified columns)
- transactions / savepoints with scope
- middleware pattern (per-request DB configuration)

**`examples/orm/orm_scope_middleware.v`** — complete middleware example showing role-based scope (admin/manager/normal)

**`vlib/orm/README.md`** — DataScope documentation

---

### Usage

```v
// Create scoped connection
db := orm.new_db(raw_conn, orm.DataScope{
    filters: [
        orm.QueryFilter{
            field: 'tenant_id'
            value: orm.Primitive(ctx.user.tenant_id)
            mode:  .dynamic
        },
        orm.QueryFilter{
            field:    'deleted_at'
            operator: .is_null
            mode:     .dynamic
        },
    ]
})

// sql block syntax
users := sql db { select from User }!

// orm_func (QueryBuilder)
users := orm.new_query[User](db).query()!

// Skip scope filters by role
admin_db := db.unscoped()           // skip all
manager_db := db.unscoped('org_id') // skip specific field
```

### Migration (global → per-instance)

```v
// Old (thread-unsafe, deprecated)
orm.configure_tenant_filter(field_name: 'tenant_id')
orm.set_current_tenant_id(1)
users := sql raw_db { select from User }

// New (thread-safe)
db := orm.new_db(raw_db, orm.DataScope{
    filters: [orm.QueryFilter{
        field: 'tenant_id', value: orm.Primitive(1), mode: .dynamic
    }]
})
users := sql db { select from User }

### DIFF
--- a/examples/orm/orm_scope_middleware.v
+++ b/examples/orm/orm_scope_middleware.v
@@ -1,0 +1,269 @@
+// orm_scope_middleware.v - ORM middleware pattern demo
+//
+// Shows how middleware manages DataScope (multi-tenant) transparently.
+// Business code only changes acquire() -> pool_acquire_scoped().
+//
+// Run: v run orm_scope_middleware.v
+
+module main
+
+import db.sqlite
+import orm
+
+// =============================================================================
+// Model
+// =============================================================================
+
+@[table: 'sys_users']
+struct SysUser {
+	id        int @[primary; sql: serial]
+	name      string
+	tenant_id int
+	org_id    int
+}
+
+// =============================================================================
+// Request context
+// =============================================================================
+
+struct UserInfo {
+mut:
+	role      string
+	tenant_id int
+}
+
+pub struct Context {
+pub mut:
+	user   UserInfo
+	dbpool &DatabasePool
+}
+
+// =============================================================================
+// Connection pool (SQLite version)
+// In real projects, use adapter.dbpool.DatabasePoolable
+// =============================================================================
+
+pub struct DatabasePool {
+	conn &sqlite.DB
+}
+
+pub fn new_pool() &DatabasePool {
+	mut db := sqlite.connect(':memory:') or { panic(err) }
+	return &DatabasePool{
+		conn: &db
+	}
+}
+
+// acquire gets a connection from pool
+pub fn (mut p DatabasePool) acquire() !&sqlite.DB {
+	return p.conn
+}
+
+// release returns a connection to pool
+pub fn (mut p DatabasePool) release(conn &sqlite.DB) ! {
+	// no-op for single-connection SQLite
+	// real pool: p.inner.put(conn)!
+}
+
+// =============================================================================
+// Middleware layer -- pool_acquire_scoped
+//
+// Responsibilities:
+//   1. Acquire connection from pool
+//   2. Create orm.DB with DataScope injected
+//   3. Skip scope filters based on user role
+//   4. Return configured orm.DB
+//
+// Business code just replaces acquire() -> pool_acquire_scoped()
+// =============================================================================
+
+pub fn pool_acquire_scoped(mut ctx Context) !(orm.DB, &sqlite.DB) {
+	// 1. acquire raw connection
+	raw_conn := ctx.dbpool.acquire()!
+
+	// 2. create base orm.DB with tenant scope
+	base_db := orm.new_db(raw_conn, orm.DataScope{
+		filters: [
+			orm.QueryFilter{
+				field: 'tenant_id'
+				value: orm.Primitive(ctx.user.tenant_id)
+				mode:  .dynamic
+			},
+		]
+	})
+
+	// 3. skip scope filters by role
+	mut scoped_db := base_db
+	match ctx.user.role {
+		'admin' {
+			scoped_db = scoped_db.unscoped()
+		}
+		'manager' {
+			scoped_db = scoped_db.unscoped('org_id')
+		}
+		'normal' {}
+		else {}
+	}
+
+	// 4. return scoped DB
+	return scoped_db, raw_conn
+}
+
+// =============================================================================
+// Business layer -- Repository
+//
+// No awareness of DataScope at all.
+// Only change: acquire() -> pool_acquire_scoped()
+// =============================================================================
+
+fn get_users(mut ctx Context) ![]SysUser {
+	db, conn := pool_acquire_scoped(mut ctx)!
+	defer {
+		ctx.dbpool.release(conn) or {}
+	}
+
+	return sql db {
+		select from SysUser
+	}!
+}
+
+fn get_user_by_name(mut ctx Context, name string) ![]SysUser {
+	db, conn := pool_acquire_scoped(mut ctx)!
+	defer {
+		ctx.dbpool.release(conn) or {}
+	}
+
+	return sql db {
+		select from SysUser where name == name
+	}!
+}
+
+fn count_users(mut ctx Context) !int {
+	db, conn := pool_acquire_scoped(mut ctx)!
+	defer {
+		ctx.dbpool.release(conn) or {}
+	}
+
+	return sql db {
+		select count from SysUser
+	}!
+}
+
+// =============================================================================
+// Demo
+// =============================================================================
+
+fn main() {
+	// setup
+	mut pool := new_pool()
+	db := pool.conn
+
+	sql db {
+		create table SysUser
+	}!
+
+	users := [
+		SysUser{
+			name:      'Alice'
+			tenant_id: 1
+			org_id:    10
+		},
+		SysUser{
+			name:      'Bob'
+			tenant_id: 1
+			org_id:    20
+		},
+		SysUser{
+			name:      'Charlie'
+			tenant_id: 2
+			org_id:    10
+		},
+		SysUser{
+			name:      'Diana'
+			tenant_id: 2
+			org_id:    20
+		},
+	]
+	for u in users {
+		sql db {
+			insert u into SysUser
+		}!
+	}
+
+	println('=== test data ===')
+	println('Alice  : tenant=1, org=10')
+	println('Bob    : tenant=1, org=20')
+	println('Charlie: tenant=2, org=10')
+	println('Diana  : tenant=2, org=20')
+
+	mut ctx := Context{
+		user:   UserInfo{
+			role:      'normal'
+			tenant_id: 1
+		}
+		dbpool: pool
+	}
+
+	// scenario 1: normal user, tenant=1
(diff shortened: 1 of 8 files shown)


Jengro777:
@codex review

Jengro777:
@codex review

Jengro777:
@JalonSolov
@GGRei 

Can it be review in this new PR?

GGRei:
@Jengro777  I can give you my opinion, but I am not the final reviewer. Medvednikov will be the one deciding whether this PR is mergeable or not.

My personal feeling is that this PR is much better than the previous ones, no doubt about that. I did not review every line of the code in depth, but there are two things I think may be worth double-checking:

- QueryBuilder / `orm_func`: maybe check that a scoped query through `orm.new_query[T](db)` really applies the filter when the scoped field exists. If embedded fields are supposed to be supported there too, it may also be worth adding a regression test with a scope on something like `Embedded.field`, just to make sure the QueryBuilder metadata path behaves like the normal `sql {}` path.

- `QueryFilter.mode: .static`: since `.static` is public but reserved for now, maybe verify that using it cannot accidentally look like an active scope while producing an unscoped query. A small regression test or a clear runtime error, if `.static` is not supported yet, could make that behavior safer.

Apart from that, the direction looks much better to me, especially because the old global tenant-filter path is kept for compatibility while the new per-instance `orm.DB` path is introduced.

Again, this is only my personal opinion, based on the PR explanation and a partial look at the code. I did not review every line in detail.

Jengro777:
> @Jengro777 I can give you my opinion, but I am not the final reviewer. Medvednikov will be the one deciding whether this PR is mergeable or not.
> 
> My personal feeling is that this PR is much better than the previous ones, no doubt about that. I did not review every line of the code in depth, but there are two things I think may be worth double-checking:
> 
> * QueryBuilder / `orm_func`: maybe check that a scoped query through `orm.new_query[T](db)` really applies the filter when the scoped field exists. If embedded fields are supposed to be supported there too, it may also be worth adding a regression test with a scope on something like `Embedded.field`, just to make sure the QueryBuilder metadata path behaves like the normal `sql {}` path.
> * `QueryFilter.mode: .static`: since `.static` is public but reserved for now, maybe verify that using it cannot accidentally look like an active scope while producing an unscoped query. A small regression test or a clear runtime error, if `.static` is not supported yet, could make that behavior safer.
> 
> Apart from that, the direction looks much better to me, especially because the old global tenant-filter path is kept for compatibility while the new per-instance `orm.DB` path is introduced.
> 
> Again, this is only my personal opinion, based on the PR explanation and a partial look at the code. I did not review every line in detail.

thank you

medvednikov:
Thanks and sorry for the delay!

PR #27348, opened 2026-06-05
cgen: fix fixed array return via local

### LINKED ISSUE #27345: cgen: fixed array return via local emits `.ret_arr` on raw array
## Description

A fixed-array return value assigned to an intermediate local and then returned generates C that treats the local raw fixed array as if it were the fixed-array return wrapper struct.

Plain `return hud_rgba()` compiles. The trigger is assigning the return value to a local first and returning that local.

## Reproduction

```v
fn hud_rgba() [4]u8 {
	return [u8(1), 2, 3, 4]!
}

fn shop_category_color() [4]u8 {
	col := hud_rgba()
	return col
}

fn main() {
	println(shop_category_color())
}
```

## Actual Result

C compilation fails with generated code that accesses `.ret_arr` on a raw C array:

```text
error: member reference base type 'Array_fixed_u8_4' (aka 'unsigned char[4]') is not a structure or union
  memcpy(_t1.ret_arr, col.ret_arr, sizeof(_v_Array_fixed_u8_4));
```

## Expected Result

The program should compile and print the returned fixed array.

## Repro Status

Reproduced on `origin/master` at `a72a784ee`, after #27342 was merged.

Original bugs.vlang.io source cluster: `doka_build` used the same pattern through a `match` over an enum.

> [!NOTE]
>
> [!NOTE]
>

medvednikov:
## Summary

Fixes #27345.

- Copy fixed-array identifier returns from the raw local array instead of reading `.ret_arr` from it.
- Add a regression test for returning a fixed array through an intermediate local initialized from another fixed-array-returning function.

## Root Cause

Fixed-array function returns use a wrapper struct with a `.ret_arr` field in generated C, but ordinary local fixed-array variables are raw C arrays. The return path was using the fixed-array return-wrapper type metadata to decide that an identifier return should read `local.ret_arr`, which is invalid for raw local arrays.

## Validation

- issue repro with `./vnew run /tmp/.../repro.v`
- `./vnew -silent test vlib/v/tests/return_fixed_array_test.v`


medvednikov:
Superseded by #27350, which contains only the issue #27345 fixed-array return change. Closing this PR because this branch also included unrelated commits/files.

PR #27363, opened 2026-06-06
v2: skip whole-program decode when transforming generic-free programs

medvednikov:
> **Draft — does NOT fix the `-gc none` flat self-host OOM. See the correction below.** Kept open for a disposition decision (land as a minor improvement, or close in favor of flat-native transform).

## What this does

`needs_full_files_for_transform()` returned `true` unconditionally, so the flat transform pipeline always decoded the whole program to legacy `[]ast.File` (`flat.to_files()`) and ran the monomorphize fixpoint. This makes it return a new `program_needs_generics` field (computed by a pure cursor scan, `program_needs_generics_from_flat`); generic-free programs then take the existing bounded per-file dispatch (`to_files_range`, one file at a time). The bounded path also now streams the three non-generic whole-program collects it needs (`declared_method_fns`, `struct_default_decl_infos`, `concrete_embedded_owner_names`) via `collect_non_generic_decl_infos_streaming`, reusing the exact per-file collectors so the maps are identical to the full-file path.

## Honest result (corrected)

I initially reported an "8.5× memory win". That was **resident set size**, and against the *wrong baseline*. Measuring the real OOM trigger (`peak memory footprint`) on **merged master**, full `v3→v4` self-host, `-gc none`:

| | RSS | peak footprint | result |
|---|---|---|---|
| baseline (full path) | 7.9 GB | 97.4 GB | OOM-killed |
| this PR (bounded path) | 4.3 GB | 94.0 GB | OOM-killed |

- The `-gc none` flat self-host was **already** OOM-killed on master — this is **not a regression**, and this PR is marginally better on both metrics.
- But it **does not fix the OOM**: under `-gc none` the flat→legacy re-decode is never freed and accumulates regardless of all-at-once vs streamed. Streaming lowers RSS, not total footprint.
- The real fix is **flat-native transform** (transform reads the flat directly, never re-decoding to legacy → nothing to accumulate), which will supersede this bounded path entirely.

## Validation

- All 19 `vlib/v2/transformer/` tests green, including a new `transform_files`-vs-`transform_files_from_flat` parity fixture covering methods + struct defaults + an embedded struct (guards the streamed collects == full-file collects).
- `program_needs_generics_test.v` pins the predicate (edge offsets + lifetime filter).
- No regression vs merged-master baseline (table above).

### DIFF
--- a/vlib/v2/transformer/monomorphize.v
+++ b/vlib/v2/transformer/monomorphize.v
@@ -210,10 +210,18 @@ fn (t &Transformer) generic_types_spec_count() int {
 fn (mut t Transformer) collect_declared_method_fns(files []ast.File) {
 	t.declared_method_fns = map[string]bool{}
 	for file in files {
-		for stmt in file.stmts {
-			if stmt is ast.FnDecl && stmt.is_method {
-				t.register_declared_method_fn(stmt, file.mod)
-			}
+		t.collect_declared_method_fns_in_file(file)
+	}
+}
+
+// collect_declared_method_fns_in_file is the per-file body of
+// collect_declared_method_fns, factored out so the bounded (streaming)
+// transform path can accumulate the same whole-program map one decoded
+// file at a time without holding the full file set.
+fn (mut t Transformer) collect_declared_method_fns_in_file(file ast.File) {
+	for stmt in file.stmts {
+		if stmt is ast.FnDecl && stmt.is_method {
+			t.register_declared_method_fn(stmt, file.mod)
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


medvednikov:
## ⚠️ Correction — converting to draft

The full `v3_flat → v4` self-host fixpoint (the canonical acceptance test: `-gc none`, arm64, generic-free) **crashes in the transform phase** — clean Scan & Parse + Type Check (132s), then a signal-termination at ~7.75 GB peak (so *not* an OOM).

What this means:
- The **memory win is real** — no ~45 GB blowup; the whole-program `to_files()` is gone.
- But the bounded path skips `prepare_files_for_transform`'s whole-program collects (`declared_method_fns` / struct-default-decl-infos / concrete-embedded-owner-names). At least one of these is needed by `transform_file` for some construct that only the **full compiler** exercises — the parity fixtures and small programs don't hit it, so they pass while the full self-host crashes.

The fix is to port the needed collect(s) cursor-native into `pre_pass_from_flat` so the bounded path is *complete*, not just to flip the gate. Marking this draft until that lands and the full fixpoint converges.

medvednikov:
Closing in favor of the real fix. The investigation here established that the `-gc none` flat self-host OOM is fundamental: under `-gc none` the transformer's flat→legacy re-decode (`to_files`/`to_files_range`) is never freed, so it accumulates regardless of all-at-once vs streamed (baseline full path: 97 GB peak footprint; this bounded path: 94 GB — both OOM-killed). The bounded-streaming approach lowers RSS but not the footprint that triggers the kill, so it can't fix the goal and would be superseded anyway.

The real fix is **flat-native transform** (the transformer reads the flat directly and writes a FlatBuilder, never materializing legacy AST → nothing to accumulate). Scoping that now. The branch `claude/flat-migration-cont` is kept for reference (correct, non-regressing, with the predicate + collect helpers + parity fixture that the flat-native work can reuse).

PR #27377, opened 2026-06-07
cgen: add deref for aliased reference types using mut captures

### LINKED ISSUE #27355: cgen, closures: `mut` capture of a pointer-alias type emits a one-deref-short member access
### Description

When a closure captures a variable by `mut` whose type is an alias of a pointer (`type AppRef = &App`), cgen stores the capture in the closure context struct as a **double** pointer (`AppRef* app;`, i.e. `App**`), but the closure body emits only a single dereference (`_V_closure_ctx->app->field`). The C compiler then rejects the member access because `_V_closure_ctx->app` is itself a pointer.

A non-`mut` capture of the same alias compiles correctly, which isolates the extra (unaccounted-for) pointer level to the `mut` capture path.

**V version:** V 0.5.1, current `master` (`5e36217f8`)
**OS/Backend:** reproduced on macOS arm64; originally reported on Linux (a `_V_closure_ctx->app` member-reference error from a captured pointer).

### Reproducer

```v
struct App {
mut:
	score int
}

type AppRef = &App

fn make_handler(mut app AppRef) fn () {
	return fn [mut app] () {
		app.score++
	}
}

fn main() {
	mut a := App{}
	h := make_handler(mut &a)
	h()
}
```

Compile with:

```
v file.v
```

### Current behavior

```
error: member reference base type 'main__AppRef' (aka 'struct main__App *') is not a structure or union
  _V_closure_ctx->app->score++;
```

The generated context struct field is `main__AppRef* app;` (a `struct main__App **`), but the body accesses it as `_V_closure_ctx->app->score` (one deref short). It should be either `(*_V_closure_ctx->app)->score`, or the field should be stored as `main__AppRef` (single pointer).

### Expected behavior

The captured-pointer member access in the closure body should match the pointer level of the context-struct field, so the program compiles.

### Notes

- Reported via the bugs.vlang.io crash reporter (a `'_V_closure_ctx->app' is a pointer; did you mean to use '->'?` variant from a captured pointer in a closure). This reproducer is the same closure-captured-pointer member-access family; the operator/pointer-level mismatch direction differs.

> [!NOTE]
>

dy-tea:
Fixes #27355

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -7110,7 +7110,9 @@ fn (mut g Gen) ref_or_deref_arg_ex(arg ast.CallArg, expected_type_ ast.Type, lan
 		&& g.table.unaliased_type(expected_type) == arg_typ {
 		if arg.expr is ast.PrefixExpr && arg.expr.op == .amp {
 			g.write('&(${exp_sym.cname}[]){')
+			g.arg_no_auto_deref = true
 			g.expr(ast.Expr(arg.expr))
+			g.arg_no_auto_deref = false
 			g.write('}[0]')
 		} else {
 			g.write('&/*mut alias*/')
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -1874,7 +1874,8 @@ fn (mut g Gen) gen_anon_fn(mut node ast.AnonFn) {
 				if obj := node.decl.scope.parent.find(var.name) {
 					if obj is ast.Var {
 						is_auto_heap = !obj.is_stack_obj && obj.is_auto_heap
-						is_auto_deref_capture = obj.is_auto_deref && !var.is_mut
+						is_auto_deref_capture = obj.is_auto_deref
+							&& (!var.is_mut || g.table.sym(obj.typ).kind == .alias)
 							&& (is_ptr || obj.typ.is_ptr())
 						if obj.smartcasts.len > 0 {
 							if g.table.type_kind(obj.typ) == .sum_type {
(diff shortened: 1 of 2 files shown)


medvednikov:
@codex review

medvednikov:
## Verdict: Yes, it's a good PR — correct, well-scoped, and idiomatic. ✅

### What I verified
- **Bug reproduces on master**: the reproducer fails with `error: struct or union expected` / `member reference base type 'main__AppRef'... is not a structure or union` (one-deref-short access on the captured pointer).
- **Fix works**: with the PR applied and `v` rebuilt, the reproducer compiles and prints `ok`.
- **No regressions**: full `vlib/v/tests/aliases/` suite passes (66 passed, 1 skipped, incl. the new test), plus targeted closure tests (`closure_test.v`, `closure_of_method_defined_on_alias_test.v`, `mut_closure_fixed_array_thread_test.v`, `type_alias_of_fn_with_mut_args_test.v`, `alias_to_ptr_arg_test.v`).
- **Edge cases I added all behave correctly**: pointer-alias to a non-struct (`type IntRef = &int`), read-only alias capture, plain `mut` struct capture, and passing an alias-typed variable (the `else` branch). Importantly, mutation **propagates through the pointer** (score = 2 after two calls), so the alias-as-pointer semantics are preserved, not silently copied.
- **Formatting**: `v fmt -verify` is clean on both files.

### Why the approach is sound
The fix is three coordinated changes, and each one mirrors an existing pattern right next to it:
1. `gen_anon_fn` (struct init) and `gen_anon_fn_decl` (struct field type) — symmetrically extend the existing `is_auto_deref && !var.is_mut` deref so it *also* fires for `mut` captures when the type is an `.alias`. This makes the context-struct field a single pointer (`AppRef` = `App*`), so the body's `->score` matches.
2. `ref_or_deref_arg_ex` — adds a branch parallel to the existing struct `expected_type == arg_typ.ref()` branch, emitting the extra `&` indirection for the `mut` alias arg. The new branch is tightly guarded (`exp_sym.kind == .alias && unaliased_type(expected) == arg_typ`) and is disjoint from the struct branch below it, so no shadowing.

### Minor nitpicks (non-blocking)
- **The test is weak**: `test_main()` only checks that it compiles and runs — there's no `assert a.score == 1`. I confirmed the mutation does propagate, so adding an assertion would pass and would lock in the *semantics*, not just "doesn't crash." Worth requesting.
- **Small asymmetry**: the new `PrefixExpr` path does `g.expr(...)` without setting `g.arg_no_auto_deref = true`, whereas the analogous struct branch immediately below does. It's harmless for the `&a` case, but a reviewer might flag the inconsistency.
- Some structural duplication with the struct branch — acceptable for readability.

medvednikov:
Thanks!

PR #27379, opened 2026-06-07
cgen, coroutines: fix `go expr` handle referencing undeclared `thread__tN`

### LINKED ISSUE #27338: cgen, coroutines: `go expr` handle references undeclared `thread__tN` C variable
### Description

With `-use-coroutines`, a `go expr` whose handle is stored/used makes cgen reference a C variable `thread__tN` that it never declares, so the generated C fails to compile.

**V version:** V 0.5.1, current `master` (`27f394507`)
**OS/Backend:** reproduced on macOS arm64; originally reported on Linux/cc. Requires `-use-coroutines`.

### Reproducer

```v
struct App {
mut:
	worker thread
}

fn (a App) loop() {}

fn main() {
	mut a := App{}
	a.worker = go a.loop()
	a.worker.wait()
}
```

Compile with:

```
v -use-coroutines file.v
```

### Current behavior

C compilation fails — V references an undeclared `thread__tN`:

```
error: use of undeclared identifier 'thread__t2'
 4112 | thread__t2;
```

The generated C for the coroutine handle is:

```c
a.worker = /*go (coroutine) */ thread__t2;
```

The coroutine branch emits the `photon_thread_create_and_migrate_to_work_pool(...)` call but never declares the `thread__tN` handle variable it then reads.

### Expected behavior

Compiles and runs (as the non-coroutine `spawn` path does).

### Notes

- Reported via the bugs.vlang.io crash reporter (original symbol `thread__t4`, from `db.flush_thread = go db.flush_loop()`).
- Likely in `vlib/v/gen/c/spawn_and_go.v`: the `is_go` (coroutine) branch sets the handle expression to `thread_${tmp}` without emitting its declaration, unlike the regular `spawn` path.

medvednikov:
## Description

Fixes #27338.

With `-use-coroutines`, a `go expr` whose handle is stored/used (e.g. `a.worker = go a.loop()`) made cgen reference a C variable `thread__tN` that it never declared:

```
error: use of undeclared identifier 'thread__t2'
 a.worker = /*go (coroutine) */ thread__t2;
```

The coroutine branch in `spawn_and_go_expr` only emitted the `photon_thread_create*` call (which returns `void` and yields no joinable handle), yet the handle `thread_<tmp>` was still written back as the expression's value.

## Fix

The Photon wrapper exposes no waitable handle, so a `go` whose handle is **actually used** (`node.is_expr`) is now lowered to the regular `spawn` (pthread) path, which declares and assigns `thread_<tmp>`. The generated C now contains:

```c
pthread_t thread__t2;
...
int _t2_thr_res = pthread_create(&thread__t2, ..., main__App_loop_thread_wrapper, arg__t2);
a.worker = thread__t2;
```

so it compiles **and** the handle is joinable via `.wait()`, matching the non-coroutine `spawn` path. Plain statement-form `go f()` (no handle used) is unchanged and keeps using the Photon work-pool path.

A companion change in `markused` is required: a `go` handle that is used must mark the same thread-handle type and pthread error helpers (`panic_error_number`, `c_error_number_str`, …) that `spawn` needs — otherwise `-skip-unused` (the default) prunes them and the rerouted spawn path fails to link with `undefined symbol 'builtin__panic_error_number'`.

## Verification

- The reproducer now generates a properly-declared `pthread_t` handle (was: undeclared identifier).
- `panic_error_number` / `c_error_number_str` are emitted (no longer pruned).
- The compiler self-compiles cleanly, and existing spawn/thread tests pass (`array_of_threads_wait_test.v`, `spawn_with_cond_fncall_test.v`, `generic_spawn_test.v`).
- Added `examples/coroutines/go_handle.v` documenting the now-working pattern.

> Note: an automated CI test isn't included because `-use-coroutines` requires the Photon `.so` (downloaded at build time, not available on CI) — consistent with the existing coroutines examples being marked `vtest build: false`.

### DIFF
--- a/examples/coroutines/go_handle.v
+++ b/examples/coroutines/go_handle.v
@@ -1,0 +1,24 @@
+// vtest build: false // This should be built with: `v -use-coroutines go_handle.v`
+// Note: the Photon wrapper is not yet trivial enough to build/install on the CI.
+//
+// Demonstrates storing the handle returned by a `go` expression and waiting on
+// it. Because the Photon coroutine wrapper exposes no joinable handle, a `go`
+// whose handle is used is lowered to the regular `spawn` (pthread) path, so the
+// handle can be waited on just like `spawn`.
+module main
+
+struct App {
+mut:
+	worker thread
+}
+
+fn (a App) loop() {
+	println('hello from the worker coroutine')
+}
+
+fn main() {
+	mut a := App{}
+	a.worker = go a.loop()
+	a.worker.wait()
+	println('done')
+}
--- a/vlib/v/gen/c/spawn_and_go.v
+++ b/vlib/v/gen/c/spawn_and_go.v
@@ -15,8 +15,19 @@ fn (mut g Gen) spawn_and_go_expr(node ast.SpawnExpr, mode SpawnGoMode) {
 	if node.call_expr.should_be_skipped {
 		return
 	}
-	is_spawn := mode == .spawn_
-	is_go := mode == .go_
+	mut is_spawn := mode == .spawn_
+	mut is_go := mode == .go_
+	if is_go && node.is_expr {
+		// A `go expr` whose handle is used as a value (e.g. `h := go f()` or
+		// `obj.field = go f()`) needs a real, joinable thread handle. The photon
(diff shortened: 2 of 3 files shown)


medvednikov:
@codex review

PR #27380, opened 2026-06-07
markused: mark mut-receiver interface implementors (fix `-skip-unused` dropping anon fn)

### LINKED ISSUE #27354: cgen, skip-unused: anon fn passed to a C callback (via interface-dispatched method) is not marked used
### Description

An anonymous function passed as a C callback argument is referenced via `closure_create` in the generated C, but its definition is omitted by `-skip-unused` (the default), so the C compiler fails with `use of undeclared identifier 'anon_fn_...'`. The anon fn is reached only through a C function-pointer call inside a method that is itself dispatched via an interface, and markused never marks it used.

Compiling the same program with `-no-skip-unused` succeeds, which pins the fault to the markused / skip-unused pass.

This is the root cause of several recurring auto-submitted reports (e.g. `vlib/net/mbedtls/ssl_connection.c.v` `SSLListener.init_sni` passing a capturing closure to `C.mbedtls_ssl_conf_sni`, and `at_exit(fn () {...})`).

**V version:** V 0.5.1, current `master` (`5e36217f8`)
**OS/Backend:** reproduced on macOS arm64; originally reported on Linux/FreeBSD.

### Reproducer

```v
module main

fn C.atexit(cb voidptr) int

interface Handler {
	setup()
}

struct Impl {
mut:
	n int
}

fn (mut i Impl) setup() {
	C.atexit(fn () {})
}

fn run(mut h Handler) {
	h.setup()
}

fn main() {
	mut impl := Impl{n: 99}
	run(mut impl)
	println('done')
}
```

Compile with:

```
v file.v
```

### Current behavior

```
error: use of undeclared identifier 'anon_fn_..._145'
  atexit(builtin__closure__closure_create(anon_fn_..._145, ...))
```

The generated C emits the closure context struct and the `closure_create(...)` call site, but neither a forward declaration nor a body for `anon_fn_..._145`.

Building with `-no-skip-unused` compiles and runs cleanly, confirming the anon fn is dropped by the used-functions analysis.

### Expected behavior

The anon fn reached through the C callback should be marked used and emitted, so the program compiles under the default `-skip-unused`.

### Notes

- Reported via the bugs.vlang.io crash reporter. Real-world instances: mbedtls SNI callback (`anon_fn_..._mbedtls_ssl_context_..._8883`) and `at_exit` in `v.dotgraph` (`anon_fn_..._257__81`).
- The combination that triggers it: the closure is passed to a `C.` function pointer, and the enclosing method is reached only via interface dispatch (so markused must follow the interface method to find the anon fn).

> [!NOTE]
>

medvednikov:
## Description

Fixes #27354.

An anonymous function passed as a C callback argument inside a method that is reached **only via interface dispatch on a `mut` receiver** was dropped by `-skip-unused` (the default), so the C compiler failed:

```
error: use of undeclared identifier 'anon_fn_..._145'
  atexit(anon_fn_..._145);
```

Building with `-no-skip-unused` worked, pinning the fault to the used-functions analysis.

## Root cause

`markused`'s interface walk (`mark_by_sym`, the `ast.Interface` arm) iterated only `isym.info.types` to find implementing types and walk their interface-method bodies. But the checker stores implementors that require a **mutable receiver** (e.g. `fn (mut i Impl) setup()`) in `isym.info.mut_types`, not `types`. Those method bodies were therefore never walked, and the symbols they reference (here, the anon fn argument) were pruned, even though cgen still emits the method via the interface vtable adapter.

## Fix

Walk `isym.info.implementor_types(true)` instead, the existing helper that merges `types` with `mut_types` (deduplicated by idx). This only ever **adds** implementors to the walk, so it cannot introduce new undeclared-symbol errors — worst case is marginally less aggressive pruning for some mut-receiver interface methods that were incorrectly skipped.

## Verification

- The reproducer now compiles and runs under default `-skip-unused`, printing `done`.
- Added regression test `vlib/v/tests/skip_unused/interface_mut_receiver_anon_fn.{vv,run.out}` (run by `compiler_errors_test.v`).
- Compiler self-compiles cleanly; all 96 `vlib/v/tests/interfaces/` tests pass, plus `generic_fn_instantiation_pruning_test.v` and `unused_fn_fixed_array_ret_test.v`.


medvednikov:
@codex review

PR #27382, opened 2026-06-07
net.http: add server-side HTTP/2 (ALPN + h2 frame demux)

quaesitor-scientiam:
## What

Lets `net.http.Server` speak **HTTP/2**, building on the TLS-termination PR
(#27373). When a client negotiates `h2` via ALPN, the server serves the request
through a new HTTP/2 driver; otherwise the existing HTTP/1.1 path is unchanged.

This completes a full client **and** server HTTP/2 stack in pure V (framing,
HPACK, ALPN, TLS, client, server) with no external dependencies.

```v
mut srv := &http.Server{
    cert:                   cert_pem
    cert_key:               key_pem
    in_memory_verification: true
    enable_http2:           true     // advertise ALPN h2, http/1.1
    handler:                MyHandler{}
}
srv.listen_and_serve()
```

## Changes

| File | Change |
|------|--------|
| `h2_server.v` | `H2ServerConn`: preface, SETTINGS, frame loop, HEADERS→Request, Response→HEADERS/DATA |
| `server.v` | new opt-in `enable_http2` field |
| `server_tls_notd_use_openssl.v` | advertise ALPN when enabled; dispatch to the h2 driver when the handshake selects `h2` |
| `h2_server_test.v`, `server_tls_test.v` | tests |

## Design

- **Opt-in, additive.** With `enable_http2` unset (or for non-TLS servers),
  behaviour is exactly as before — no ALPN advertised, no new code on that path.
- **The `Handler` interface is untouched.** `Handler.handle(Request) Response`
  is invoked once per request stream. `req.url` is the request-target (the
  `:path` pseudo-header) and `Host` comes from `:authority`, so existing
  HTTP/1.1 handlers work as-is on the new transport.
- **Serial for v1.** The driver advertises `SETTINGS_MAX_CONCURRENT_STREAMS=1`,
  so the peer sends one request at a time. Stream multiplexing with a
  background reader is a planned follow-up.
- **Robustness:** HEADERS+CONTINUATION assembly with the §6.10 ordering rule,
  flow-control replenishment on DATA, hop-by-hop response headers dropped
  (§8.1.2.2), an 8 MiB request-body cap with `RST_STREAM(REFUSED_STREAM)`, and a
  best-effort GOAWAY on protocol error.
- **mbedtls backend.** Like the TLS PR, the h2 server path is on the default
  backend; `-d use_openssl` keeps the TLS-server stub. An OpenSSL server
  listener is a follow-up.

## Tests

- `h2_server_test.v` drives the server through an in-memory **blocking pipe**
  with the existing HTTP/2 **client** — GET, POST with a body, and a non-200
  status all round-trip, exercising the real framing + HPACK on both sides.
- `server_tls_test.v` adds a **TLS + ALPN end-to-end** test: a real HTTPS
  listener with `enable_http2: true` serves `h2` to `http.fetch(enable_http2:
  true)` and `http/1.1` to a plain `http.fetch` on the same port.

```
./vnew test vlib/net/http/h2_server_test.v vlib/net/http/server_tls_test.v   # pass
./vnew -W -cstrict -cc clang test vlib/net/http/h2_server_test.v             # pass
./vnew -silent test vlib/net/http/                                          # 20 passed, 1 skipped (Windows)
./vnew -d use_openssl -silent test vlib/net/http/                           # 20 passed, 1 skipped
```

## Follow-ups

- Stream multiplexing + a background reader (concurrent streams per connection).
- OpenSSL server listener (so `-d use_openssl` can serve TLS / h2 too).

### DIFF
--- a/vlib/net/http/h2_server.v
+++ b/vlib/net/http/h2_server.v
@@ -163,12 +163,30 @@ fn (mut c H2ServerConn) dispatch_frame(frame H2Frame, mut handler Handler) ! {
 fn (mut c H2ServerConn) apply_settings(settings []H2Setting) {
 	for s in settings {
 		match s.id {
-			h2_settings_header_table_size { c.peer.header_table_size = s.value }
-			h2_settings_enable_push { c.peer.enable_push = s.value != 0 }
-			h2_settings_max_concurrent_streams { c.peer.max_concurrent_streams = s.value }
-			h2_settings_initial_window_size { c.peer.initial_window_size = s.value }
-			h2_settings_max_frame_size { c.peer.max_frame_size = s.value }
-			h2_settings_max_header_list_size { c.peer.max_header_list_size = s.value }
+			h2_settings_header_table_size {
+				c.peer.header_table_size = s.value
+			}
+			h2_settings_enable_push {
+				c.peer.enable_push = s.value != 0
+			}
+			h2_settings_max_concurrent_streams {
+				c.peer.max_concurrent_streams = s.value
+			}
+			h2_settings_initial_window_size {
+				// RFC 7540 Section 6.9.2: a change to the initial window size
+				// adjusts the send window of every active stream by the delta.
+				delta := i64(s.value) - i64(c.peer.initial_window_size)
+				c.peer.initial_window_size = s.value
+				for _, mut st in c.streams {
+					st.send_window += delta
+				}
+			}
+			h2_settings_max_frame_size {
+				c.peer.max_frame_size = s.value
+			}
+			h2_settings_max_header_list_size {
+				c.peer.max_header_list_size = s.value
+			}
 			else {}
 		}
 	}
@@ -365,9 +383,23 @@ fn (mut c H2ServerConn) send_body(stream_id u32, body []u8) ! {
 	max := int(c.peer.max_frame_size)
 	mut off := 0
 	for off < body.len {
+		// Respect both the connection and per-stream send windows
+		// (RFC 7540 Section 6.9). When either is exhausted, read frames until
+		// the peer grows a window with WINDOW_UPDATE.
+		for c.send_window <= 0 || c.stream_send_window(stream_id) <= 0 {
+			c.pump_for_window(stream_id)!
+		}
+		avail := if c.send_window < c.stream_send_window(stream_id) {
+			c.send_window
+		} else {
+			c.stream_send_window(stream_id)
+		}
 		mut chunk := body.len - off
 		if chunk > max {
 			chunk = max
+		}
+		if i64(chunk) > avail {
+			chunk = int(avail)
 		}
 		next := off + chunk
 		c.send_frame(H2DataFrame{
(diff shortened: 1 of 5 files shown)


quaesitor-scientiam:
> Codex usage limits have been reached for code reviews. Please check with the admins of this repo to increase the limits by adding credits.

Running local codex review given usage uses are exceeded

• Findings

  - High: Windows default clients never exercise the new HTTP/2 TLS path. On Windows, ssl_do() immediately routes to SChannel at /abs?. In
    this repo: vlib/net/http/backend.c.v:9 calls vlib/net/http/backend_vschannel_windows.c.v:21, which builds and sends an HTTP/1.1 request
    and has no ALPN/enable_http2 handling. The new test at vlib/net/http/server_tls_test.v:110 fails in the default Windows configuration:
    resp.version() is HTTP/1.1, not HTTP/2.0. Either SChannel needs ALPN/H2 support, or this test/feature path needs to be gated to -d
    no_vschannel.

  - Medium: Server response DATA ignores HTTP/2 outbound flow control. vlib/net/http/h2_server.v:128 tracks WINDOW_UPDATE into connection and
    stream send windows, but vlib/net/http/h2_server.v:364 sends the whole body chunked only by max frame size, without waiting for/
    decrementing either window. Large responses or clients that lower SETTINGS_INITIAL_WINDOW_SIZE can receive more DATA than permitted and
    reset with FLOW_CONTROL_ERROR.

  Tests Run

  - .\vnew.exe test vlib\net\http\h2_server_test.v passed.
  - .\vnew.exe test vlib\net\http\server_tls_test.v failed on Windows default SChannel path: expected HTTP/2.0, got HTTP/1.1.
  - .\vnew.exe -d no_vschannel test vlib\net\http\server_tls_test.v passed.

quaesitor-scientiam:
Thanks for running the local review — both addressed in ed346468f:

**High (Windows ALPN):** `test_server_tls_h2_negotiation` now skips on the default Windows/SChannel configuration (`$if windows && !no_vschannel ?`), since the SChannel client can't advertise ALPN and so can't negotiate HTTP/2. The path stays covered under `-d no_vschannel` (mbedtls client), matching how the rest of the suite handles the SChannel limitation. Full SChannel ALPN is a separate follow-up. The server side itself is unaffected (it uses the mbedtls listener regardless).

**Medium (server send flow control):** `send_body` now bounds each DATA frame by `min(connection, stream)` window, decrements both, and waits for `WINDOW_UPDATE` (servicing SETTINGS/PING/WINDOW_UPDATE and RST_STREAM for the active stream) when a window is exhausted. `apply_settings` also adjusts active streams' send windows by the delta on `SETTINGS_INITIAL_WINDOW_SIZE` change (§6.9.2). Added `test_h2_server_respects_send_window`: a raw client advertises `INITIAL_WINDOW_SIZE=10`, and the test asserts the first DATA frame is ≤ 10 bytes and the full 100-byte body arrives after a `WINDOW_UPDATE`.

Verified: `-d no_vschannel test server_tls_test.v` passes; full `vlib/net/http` suite green on mbedtls and `-d use_openssl`; `-W -cstrict -cc clang` clean.

quaesitor-scientiam:
Filed #27383 to track full SChannel ALPN / HTTP/2 support — removing the `$if windows && !no_vschannel ?` skip added here is listed as one of its acceptance criteria.

Issue #27383, opened 2026-06-07
net.http: add ALPN (and HTTP/2) support to the Windows SChannel TLS backend


quaesitor-scientiam:
## Summary

The Windows SChannel (vschannel) TLS backend does not advertise or negotiate
ALPN, so on a default Windows build the HTTP client cannot negotiate HTTP/2 even
when `enable_http2: true` is set — it silently falls back to HTTP/1.1. The
mbedtls and OpenSSL backends already support ALPN (#27343); SChannel is the gap.

## Background

ALPN support landed for mbedtls and OpenSSL in #27343, which explicitly deferred
the SChannel backend. Since then the gap has surfaced in several places:

- `http.fetch(enable_http2: true)` on default Windows reaches
  `vlib/net/http/backend_vschannel_windows.c.v`, which builds an HTTP/1.1
  request and has no ALPN / `enable_http2` handling, so it never speaks HTTP/2.
- The server-side HTTP/2 PR (#27382) had to gate its ALPN end-to-end test with
  `$if windows && !no_vschannel ?` because the default Windows client can't
  negotiate `h2`; the path is only covered with `-d no_vschannel` (mbedtls
  client).

## What's needed

Add ALPN to the SChannel client (and, ideally, listener) so Windows reaches
parity with the mbedtls/OpenSSL backends:

- Advertise the ALPN protocol list during the handshake via
  `SecApplicationProtocolNegotiationExt_ALPN` in the `SCH_CREDENTIALS` /
  `SecBuffer` passed to `InitializeSecurityContext`.
- Expose the negotiated protocol (the SChannel equivalent of
  `negotiated_alpn()`), via `SECPKG_ATTR_APPLICATION_PROTOCOL` /
  `QueryContextAttributes`.
- Wire the existing `enable_http2` path in `backend_vschannel_windows.c.v` to
  advertise `h2, http/1.1` and dispatch to `H2Conn` when `h2` is selected,
  mirroring `net_ssl_do` in `backend.c.v`.

## Acceptance criteria

- A `negotiated_alpn()`-equivalent works on the SChannel backend.
- `http.fetch(url, enable_http2: true)` negotiates HTTP/2 on default Windows
  against an h2 server.
- The `$if windows && !no_vschannel ?` skip in
  `vlib/net/http/server_tls_test.v::test_server_tls_h2_negotiation` can be
  removed.

## References

- #27343 (ALPN for mbedtls/OpenSSL; deferred SChannel)
[...]

PR #27384, opened 2026-06-07
net.http: enable HTTP/2 by default for https requests

quaesitor-scientiam:
## What

Flips the `enable_http2` default (on `Request` and `FetchConfig`) from `false`
to `true`, so `http.fetch` / `http.get` to an `https://` URL **negotiate HTTP/2
when the server offers it**, and fall back to HTTP/1.1 otherwise. To force
HTTP/1.1, set `enable_http2: false`.

```v
resp := http.get('https://example.com/')        // HTTP/2 if the server supports it
resp.version() // .v2_0 on most modern servers, .v1_1 otherwise

forced := http.fetch(url: 'https://example.com/', enable_http2: false)! // always HTTP/1.1
```

This is the opt-out counterpart to the opt-in support added in #27362, now that
the HTTP/2 client also handles redirects, streaming callbacks (#27369), and flow
control, and has been exercised against real servers.

## Scope

The change is three lines (the two field defaults + the doc) plus a test update.

| File | Change |
|------|--------|
| `request.v`, `http.v` | `enable_http2 bool` → `enable_http2 bool = true`, with updated docs |
| `h2_client_test.v` | the gated real-server test now asserts a default `get()` negotiates HTTP/2 and that `enable_http2: false` forces HTTP/1.1 |

## Behaviour / compatibility

- **Unaffected:** plain `http://` (no ALPN, never HTTP/2), and the Windows
  SChannel backend, which has no ALPN yet and stays on HTTP/1.1 (tracked in
  \#27383).
- **No latency regression:** the client sends the connection preface, its
  SETTINGS, and the request together without waiting for the server's SETTINGS,
  so there's no added round trip; HPACK also shrinks request headers.
- **Transparent fallback:** advertising `h2, http/1.1` is safe — servers that
  don't understand ALPN simply ignore it and the client uses HTTP/1.1.

## Honest tradeoff for the maintainer to weigh

Each `fetch` is still **one connection / one request** (no pooling or
multiplexing yet), so HTTP/2's headline win — multiplexing many requests over
one connection — doesn't yet materialize. This PR is purely about defaulting to
the newer protocol; pooling/multiplexing is a planned follow-up. If you'd rather
keep HTTP/2 opt-in until pooling lands, this PR can simply wait — the opt-in
path (#27362) already works.

## Tests

- Full `vlib/net/http` suite green (mbedtls; 1 Windows-only skip).
- The existing **real-network HTTPS test suite** (`http_httpbin_test.v`:
  GET/POST/PUT/PATCH/DELETE/HEAD against a live server) passes through the new
  default HTTP/2 path unchanged.
- `-d network` test confirms a default `get('https://www.google.com/')`
  returns `HTTP/2.0`, and `enable_http2: false` returns `HTTP/1.1`.
- `-W -cstrict -cc clang` clean.

### DIFF
--- a/vlib/net/http/backend.c.v
+++ b/vlib/net/http/backend.c.v
@@ -21,10 +21,12 @@ fn net_ssl_do(req &Request, port int, method Method, host_name string, path stri
 		eprint(req_headers)
 		eprintln('')
 	}
-	// Advertise ALPN `h2` (with an `http/1.1` fallback) only when HTTP/2 is
-	// requested, so existing callers see no change on the wire. The HTTP/2 read
-	// path now feeds the same streaming callbacks and honors the stop limits,
-	// so they no longer force HTTP/1.1.
+	// Advertise ALPN `h2` (with an `http/1.1` fallback) when HTTP/2 is enabled.
+	// This is the default for https requests, so ordinary get()/fetch() calls
+	// advertise ALPN and use HTTP/2 when the server selects it; callers can opt
+	// out with `enable_http2: false`. The HTTP/2 read path feeds the same
+	// streaming callbacks and honors the stop limits, so they do not force
+	// HTTP/1.1.
 	alpn := if req.enable_http2 { ['h2', 'http/1.1'] } else { []string{} }
 	for {
 		mut ssl_conn := ssl.new_ssl_conn(
--- a/vlib/net/http/h2_client_test.v
+++ b/vlib/net/http/h2_client_test.v
@@ -79,18 +79,19 @@ fn test_h2_authority_omits_default_port() {
 	assert h2_authority('example.com', 8443) == 'example.com:8443'
 }
 
-// Opt-in end-to-end test against a real HTTP/2 server. Run with `-d network`,
+// End-to-end test against a real HTTP/2 server. Run with `-d network`,
 // e.g. `v -d network test vlib/net/http/h2_client_test.v`.
 fn test_http2_fetch_real_server() {
 	$if !network ? {
 		return
 	}
-	resp := fetch(url: 'https://www.google.com/', enable_http2: true)!
+	// HTTP/2 is negotiated by default for https requests.
+	resp := get('https://www.google.com/')!
 	assert resp.version() == .v2_0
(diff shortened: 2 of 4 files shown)


quaesitor-scientiam:
**• Codex Review Findings**

  - Medium: The updated -d network test will fail on default Windows builds. get('https://www.google.com/') at vlib/net/http/
    h2_client_test.v:89 now expects .v2_0, but Windows still routes HTTPS through SChannel at vlib/net/http/backend.c.v:10. That backend
    builds a plain HTTP/1.1 request and has no ALPN/HTTP/2 path at vlib/net/http/backend_vschannel_windows.c.v:21. The test needs a Windows/
    no_vschannel gate or the SChannel path needs HTTP/2 support.

  - Low: The comment in vlib/net/http/backend.c.v:24 is now stale: it says ALPN is advertised “only when HTTP/2 is requested” so existing
    callers see no wire change. With enable_http2 defaulting to true, ordinary get()/fetch() calls now advertise ALPN on non-SChannel HTTPS.

  Tests Run

  - .\vnew.exe test vlib\net\http\h2_client_test.v passed; the -d network block was skipped.
  - .\vnew.exe fmt -verify vlib\net\http\http.v vlib\net\http\request.v vlib\net\http\h2_client_test.v passed.
  - git diff --check master...HEAD passed.

quaesitor-scientiam:
Both addressed in fa3d8c17c:

**Medium (Windows network test):** `test_http2_fetch_real_server` now skips on the default Windows/SChannel configuration (`$if windows && !no_vschannel ?`), since SChannel has no ALPN yet (#27383) so a default `get()` there stays on HTTP/1.1. Still covered with `-d no_vschannel` (mbedtls client).

**Low (stale comment):** Updated the comment in `backend.c.v` — it no longer claims "no change on the wire"; it now notes that with `enable_http2` defaulting to true, ordinary `get()`/`fetch()` advertise ALPN and use HTTP/2 when the server selects it, with opt-out via `enable_http2: false`.

Verified: full `vlib/net/http` suite green; `-d network` test passes (non-Windows); `fmt -verify` clean.

PR #27385, opened 2026-06-07
v2: fix x64 tiny CI regressions

GGRei:
## Summary

This fixes V2 x64 CI regressions exposed by the runtime smoke tests.

The Linux hanoi regression appeared after the recent AST / FlatAst call-lowering changes: the legacy SSA path already avoided `string__str(x)` when `x` was already a `string`, but the FlatAst path did not mirror that guard yet.

As a result, the real `examples/hanoi.v` path was still lowering string parameters through `builtin__string__str`, which made the tiny linker reach `builtin__malloc_noscan -> malloc`. Strict tiny correctly rejected the build, and auto tiny fell back to a hosted binary.

This PR mirrors the existing legacy SSA guard in the FlatAst call path, so `string__str(x)` / `builtin__string__str(x)` now returns `x` directly when `x` is already a string.

It also fixes the related Windows x64 PE smoke failure where some examples now reach `malloc`. The PE runtime now provides a small `malloc` thunk backed by `GetProcessHeap` / `HeapAlloc`, without importing the CRT, so the Windows x64 path can keep linking those smoke examples correctly.

This should restore the failing V2 x64 smoke CI checks on Linux and Windows.

## Validation

- `./v fmt -verify vlib/v2/ssa/builder.v vlib/v2/gen/x64/x64_runtime_smoke_test.v vlib/v2/gen/x64/pe.v vlib/v2/gen/x64/x64_pe_linker_test.v`
- `git diff --check`
- `./v test -run-only test_x64_linux_hanoi_example_strict_tiny_no_libc vlib/v2/gen/x64/x64_runtime_smoke_test.v`
- `./v test -run-only test_x64_linux_hanoi_example_auto_tiny_no_libc vlib/v2/gen/x64/x64_runtime_smoke_test.v`
- `./v test -run-only test_x64_linux_string_plus_two_runtime_strict_tiny_no_libc vlib/v2/gen/x64/x64_runtime_smoke_test.v`
- `./v test vlib/v2/gen/x64/x64_pe_linker_test.v`
- `./v test vlib/v2/ssa/tree_sumtype_lowering_test.v`

### DIFF
--- a/vlib/v2/gen/x64/pe.v
+++ b/vlib/v2/gen/x64/pe.v
@@ -390,6 +390,10 @@ fn (l PeLinker) build_runtime_text() PeRuntimeText {
 		rt.symbols['_errno'] = runtime_base + u32(rt.bytes.len)
 		pe_emit_runtime_errno(mut rt)
 	}
+	if l.uses_undefined_symbol('malloc') {
+		rt.symbols['malloc'] = runtime_base + u32(rt.bytes.len)
+		pe_emit_runtime_malloc(mut rt)
+	}
 	if l.uses_undefined_symbol('calloc') {
 		rt.symbols['calloc'] = runtime_base + u32(rt.bytes.len)
 		pe_emit_runtime_calloc(mut rt)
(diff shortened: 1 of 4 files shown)


GGRei:
The V2 x64 backend failures are fixed now ( confirmed in CI checks ). It's ready.

Issue #27358, opened 2026-06-05
Shared canonical Huffman builder (hash.huffman) for compress.deflate and net.http HPACK


quaesitor-scientiam:
## Summary

`compress.deflate` and the new HTTP/2 HPACK code (#27353) both implement
canonical Huffman coding. Per discussion on #27353 (with @JalonSolov), the
shared piece should be factored into a common builder — tentatively
`hash.huffman` — that both modules use, rather than duplicating the logic.

This issue tracks the design and the follow-up PR. PR #27353 lands as-is; this
is deliberately separate so the cross-cutting refactor of `compress.deflate`
(which is stable and used by zlib/gzip) gets its own review.

## What is actually shared

The canonical code-assignment step: given an array of per-symbol bit lengths,
produce the per-symbol codes (and a decode structure). Verified that HPACK's
RFC 7541 Appendix B table is canonical — all 257 codes rebuild exactly from the
lengths alone using the same `bl_count`/`next_code` algorithm `compress.deflate`
already uses. So HPACK could ship only the lengths and derive its codes via the
shared builder.

## What differs (and must be parameters, not assumptions)

- **Max code length:** DEFLATE caps at 15 bits and decodes via a single flat
  `2^max_bits` table (~32K entries); HPACK codes go up to 30 bits, where a flat
  `2^30` table is infeasible (needs a tree / multi-level table / bit-at-a-time
  decode). So the builder must choose its decode representation based on the
  bit-width.
- **Bit order:** DEFLATE is LSB-first (it `bit_reverse`s every code); HPACK is
  MSB-first.
- Not shared: the actual bit I/O loops, and codec-specific semantics
  (HPACK EOS/all-ones padding; DEFLATE end-of-block, extra-bits, distance
  alphabet, length-code RLE).

## Proposed shape (per @JalonSolov)

A `hash.huffman` builder parameterized by:
- number of bits (max code length), and
- bit order (LSB vs MSB).

Open design question raised on the PR: whether those params should have
defaults. Current lean is **no defaults**, with the config marked `required` /
`@[noinit]` so callers must specify them explicitly. Decide as part of the PR.

## Acceptance criteria

- A shared canonical-Huffman builder in `hash.huffman` (or agreed location).
- `compress.deflate` migrated to it with **no regression** to its flat-table
  decode performance (benchmark before/after).
- `net.http` HPACK migrated to build its codes from the lengths array via the
  shared builder.
[...]

PR #27392, opened 2026-06-08
hash.huffman: add shared canonical Huffman builder; migrate HPACK to it

quaesitor-scientiam:
## What

Adds `hash.huffman`, a small module that performs the **canonical Huffman code-assignment step** (RFC 1951 §3.2.2 `bl_count` / `next_code`) shared by `compress.deflate` and the HTTP/2 HPACK code in `net.http`, and migrates HPACK onto it.

This is the first half of #27358.

## Why

`compress.deflate` and HPACK (#27353) both implement canonical Huffman coding. Per the discussion on #27353 (with @JalonSolov), the shared piece — assigning per-symbol codes from per-symbol bit lengths — should live in one place rather than being duplicated.

## The module

`huffman.build(cfg)` assigns canonical codes from a lengths array. It is parameterized by the two things the callers genuinely differ on, with **no defaults** (`@[required]` config fields, as agreed on the issue — defaults would silently fit only one caller):

- **`max_bits`** — DEFLATE caps at 15 bits and decodes via a flat `2^max_bits` table; HPACK goes to 30 bits, where a flat table is infeasible and decoding is a map / bit-at-a-time.
- **`bit_order`** — DEFLATE is LSB-first (bit-reverses each code); HPACK is MSB-first.

API:
- `build(cfg) !Table` — returns the per-symbol `codes` + metadata (for callers needing the codes, e.g. HPACK encoding, and/or `decode_map()`).
- `flat_table(cfg) ![]u32` — builds a flat `2^max_bits` decode table directly from the lengths (the DEFLATE strategy), for both bit orders. Allocates no intermediate codes array, does not copy the lengths, and **skips the invalid pre-fill for complete codes** (the shared `next_codes` reports completeness from the Kraft residual for free) — so it does strictly less work than a hand-rolled fill on the common path. This is what the deflate follow-up consumes.
- `(Table) decode_map() !map` — packed `(len,code)→symbol` map for codecs too wide for a flat table (HPACK's 30 bits).

Validation (shared `next_codes`) rejects over-subscribed codes via the Kraft inequality while permitting incomplete codes (both DEFLATE and HPACK rely on that).

## HPACK migration

The RFC 7541 Appendix B table is canonical, so the generated `h2_huffman_codes` array (258 lines) is **dropped**; the codes are now rebuilt at startup from the per-symbol lengths via `huffman.build(...)`. A regression test pins a few known Appendix B codes (NUL, '0', '1', 'a', EOS) so a bad rebuild is caught.

## Follow-up (separate PR)

Migrating `compress.deflate` onto `flat_table()` is the second half of #27358 and lands separately, since deflate is stable and used by zlib/gzip. That migration is **already implemented and benchmarked**: decode-dominated inflate throughput is unchanged, and build-dominated workloads (many small dynamic-block streams) run **~5–11% faster** than today, because the complete-code fast path skips the invalid table pre-fill the current code always pays. It will be opened once this merges.

## Tests

- New `vlib/hash/huffman/huffman_test.v`: 14 tests incl. the RFC 1951 §3.2.2 worked example, LSB/MSB equivalence, flat-table round-trips for **both** bit orders, the incomplete-code gap path, `decode_map`, and the error paths (over-subscription, length > max_bits, etc.).
- `net.http` HPACK + Huffman suites stay green.

Verified: huffman module + `net.http` HPACK pass on **mbedtls + openssl** and under **`-cstrict -cc clang`**; `v fmt -verify` clean.


JalonSolov:
Bouncing to pick up small change in master.

quaesitor-scientiam:
## Benchmark: deflate inflate, before/after the `hash.huffman` migration

Posting numbers for review since the follow-up deflate PR depends on `flat_table()` here. The migration is **performance-neutral on decode** and **faster on build-dominated workloads**, because `flat_table()` skips the invalid table pre-fill for *complete* codes (the common case) — work the current `build_huff_tree` always does.

### Method
- Build: `v -prod`, same machine, no other load.
- **baseline** = current master `build_huff_tree` (hand-rolled `bl_count`/`next_code` + flat-table fill).
- **candidate** = `compress.deflate` migrated to `huffman.flat_table()` (with the complete-code fast path).
- Each number is the **median of 7 runs**; ranges are min–max across those runs.
- Decode is timed by inflating a pre-compressed buffer N times (compress excluded from timing).

### Corpora
| name | source | block type | stresses |
|---|---|---|---|
| `text-8MiB` | V compressor, English-ish text | fixed (btype 1) | decode loop |
| `varied-4MiB` | V compressor, xorshift bytes | fixed | decode loop (low-compressibility) |
| `dyn-2MiB` | **zlib** level 9 (`wbits=-15`) | dynamic (btype 2) | decode loop + per-block tree build |
| `small-256B` | V compressor, 2000 × 256 B msgs | fixed | **`build_huff_tree` (build-dominated)** |
| `dyn-small400` | **zlib**, 2000 × ~400 B msgs | dynamic | **per-block cl/ll/dist tree build (build-dominated)** |

(V's own compressor only emits fixed-Huffman blocks, so the dynamic-block paths are fed by zlib output to exercise `build_huff_tree`'s btype-2 path.)

### Results (median of 7; ranges in brackets)
| corpus | baseline | candidate | Δ | ranges overlap? |
|---|---|---|---|---|
| `text-8MiB` (decode) | 627 MB/s | 636 MB/s | +1.5% | yes — noise |
| `varied-4MiB` (decode) | 162 MB/s | 160 MB/s | −1.4% | yes — noise |
| `dyn-2MiB` (decode) | 345 MB/s | 346 MB/s | +0.3% | yes — noise |
| `small-256B` (**build**) | 201,247/s `[198.8–202.8k]` | 222,798/s `[211.5–224.9k]` | **+10.7%** | **no** |
| `dyn-small400` (**build**) | 108,847/s `[108.5–110.5k]` | 114,385/s `[112.9–115.7k]` | **+5.1%** | **no** |

The two build-dominated rows have **non-overlapping** min–max ranges → the speedup is real and repeatable, not run-to-run jitter. Decode-dominated rows are at parity (the `huff_decode` loop is byte-for-byte unchanged; a one-time table build is amortized over a large stream).

### Why build-dominated wins
A *complete* code (Kraft equality) tiles every index of the `2^max_bits` table, so the `0xFFFF_FFFF` pre-fill the baseline always runs is dead work. `next_codes` reports completeness from the Kraft residual for free, and `flat_table()` then allocates zeroed (`vcalloc`) and skips the fill — falling back to the fill only for incomplete codes, where gaps must read back as `flat_invalid_entry`. V's fixed lit/len (512=2⁹) and dist (32=2⁵) trees are complete, as are most zlib dynamic trees.

### Reproduce
The harness + a Python data generator (zlib dynamic-block corpora) are available on request; happy to drop them in `cmd/tools/` or a gist if the team wants the numbers re-run on CI hardware. Correctness is covered by the existing `compress.deflate` / gzip / zlib suites (all green) plus the new `hash.huffman` tests including the incomplete-code gap path.

PR #27394, opened 2026-06-08
compress.deflate: build Huffman trees via the shared hash.huffman builder

### LINKED ISSUE #27358: Shared canonical Huffman builder (hash.huffman) for compress.deflate and net.http HPACK

quaesitor-scientiam:
## What

Migrates `compress.deflate` onto `hash.huffman` (merged in #27392), removing its two in-tree copies of the canonical Huffman code-assignment (RFC 1951 §3.2.2 `bl_count` / `next_code`).

This is the **second half of #27358** (the first half — the shared module + the HPACK migration — landed in #27392).

## Changes

- **Decode** — `build_huff_tree()` now fills its flat LSB-first lookup table via `huffman.flat_table()`. The table format (`entry = (symbol << 5) | length`, `0xFFFF_FFFF` for invalid) and the `huff_decode` loop are **unchanged**. `build_huff_tree()` now returns `!HuffTree` so an over-subscribed (malformed) code surfaces as a clean error instead of a silently corrupt table; all call sites propagate with `!`.
- **Encode** — `fixed_litlen_encode()` takes its reversed codes straight from `huffman.build(..., bit_order: .lsb_first).codes`.

No behavior change for any valid DEFLATE stream (incomplete codes are still accepted).

## Performance

Benchmarked before/after with `v -prod` (full method + corpora in the [#27392 benchmark comment](https://github.com/vlang/v/pull/27392#issuecomment-4650916973)):

| corpus | baseline | this PR | Δ |
|---|---|---|---|
| text-8MiB (fixed, **decode**) | ~627 MB/s | ~636 MB/s | parity (noise) |
| dyn-2MiB (dynamic, **decode**) | ~345 MB/s | ~346 MB/s | parity (noise) |
| small-256B (fixed, **build**) | ~204k inflates/s | ~227k inflates/s | **+10.8%** |
| dyn-small400 (dynamic, **build**) | ~110k inflates/s | ~118k inflates/s | **+7.5%** |

Decode-dominated throughput is unchanged (the decode loop is byte-for-byte identical); build-dominated workloads (many small streams) are faster, because `flat_table()` skips the invalid table pre-fill for *complete* codes — which V's fixed trees and most dynamic trees are. The build-dominated ranges are non-overlapping across runs, so the speedup is real, not jitter.

## Tests

`compress.deflate` / gzip / zlib / snappy / zstd suites stay green; `-cstrict -cc clang` clean; `v fmt -verify` clean.

Closes #27358.

### DIFF
--- a/vlib/compress/deflate/deflate_compress.v
+++ b/vlib/compress/deflate/deflate_compress.v
@@ -1,5 +1,7 @@
 module deflate
 
+import hash.huffman
+
 const deflate_hash_bits = 15
 const deflate_hash_size = 1 << deflate_hash_bits
 const deflate_max_chain = 64
@@ -8,36 +10,12 @@ const deflate_max_match = 258
 const deflate_window = 32768
 
 // fixed_litlen_encode returns (reversed_codes, code_lengths) for fixed Huffman lit/len.
+// The LSB-first (bit-reversed) codes come straight from the shared canonical
+// builder, since the encoder writes bits LSB-first.
 fn fixed_litlen_encode() ([]u32, []int) {
 	lens := fixed_litlen_lengths()
-	mut max_bits := 0
-	for l in lens {
-		if l > max_bits {
-			max_bits = l
-		}
-	}
-	mut bl_count := []int{len: max_bits + 1}
-	for l in lens {
-		if l > 0 {
-			bl_count[l]++
-		}
-	}
-	mut next_code := []u32{len: max_bits + 1}
-	mut c := u32(0)
-	for bits in 1 .. max_bits + 1 {
-		c = (c + u32(bl_count[bits - 1])) << 1
-		next_code[bits] = c
-	}
-	mut codes := []u32{len: 288}
-	for sym in 0 .. 288 {
-		l := lens[sym]
-		if l == 0 {
-			continue
-		}
-		codes[sym] = bit_reverse(next_code[l], l)
-		next_code[l]++
-	}
-	return codes, lens
+	t := huffman.build(lengths: lens, max_bits: 9, bit_order: .lsb_first) or { panic(err) }
+	return t.codes, lens
 }
 
 // fixed_dist_encode returns (reversed_codes, code_lengths) for fixed Huffman distance.
--- a/vlib/compress/deflate/deflate_compress.v
+++ b/vlib/compress/deflate/deflate_compress.v
@@ -12,9 +12,9 @@ const deflate_window = 32768
 // fixed_litlen_encode returns (reversed_codes, code_lengths) for fixed Huffman lit/len.
 // The LSB-first (bit-reversed) codes come straight from the shared canonical
 // builder, since the encoder writes bits LSB-first.
-fn fixed_litlen_encode() ([]u32, []int) {
+fn fixed_litlen_encode() !([]u32, []int) {
 	lens := fixed_litlen_lengths()
-	t := huffman.build(lengths: lens, max_bits: 9, bit_order: .lsb_first) or { panic(err) }
+	t := huffman.build(lengths: lens, max_bits: 9, bit_order: .lsb_first)!
 	return t.codes, lens
 }
 
@@ -113,8 +113,8 @@ fn (mut w BitWriter) flush() {
 
 // deflate_compress_fixed compresses data to RFC 1951 DEFLATE using fixed Huffman codes.
 @[direct_array_access]
-fn deflate_compress_fixed(data []u8) []u8 {
-	ll_codes, ll_lens := fixed_litlen_encode()
+fn deflate_compress_fixed(data []u8) ![]u8 {
+	ll_codes, ll_lens := fixed_litlen_encode()!
 	d_codes, d_lens := fixed_dist_encode()
 	mut w := BitWriter{}
 	// BFINAL=1, BTYPE=01 (fixed Huffman)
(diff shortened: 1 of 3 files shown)


vlib/compress/deflate/deflate_compress.v
```diff
@@ -8,36 +10,12 @@ const deflate_max_match = 258
...
-	for sym in 0 .. 288 {
-		l := lens[sym]
-		if l == 0 {
-			continue
-		}
-		codes[sym] = bit_reverse(next_code[l], l)
-		next_code[l]++
-	}
-	return codes, lens
+	t := huffman.build(lengths: lens, max_bits: 9, bit_order: .lsb_first) or { panic(err) }
```
JalonSolov: Always better to return an error than to panic.  panic is immediate exit with no cleanup possible.

vlib/compress/deflate/deflate_compress.v
```diff
@@ -8,36 +10,12 @@ const deflate_max_match = 258
 const deflate_window = 32768
 
 // fixed_litlen_encode returns (reversed_codes, code_lengths) for fixed Huffman lit/len.
+// The LSB-first (bit-reversed) codes come straight from the shared canonical
+// builder, since the encoder writes bits LSB-first.
 fn fixed_litlen_encode() ([]u32, []int) {
```
JalonSolov: Why return 2 arrays?  Also, prefer `i32` to `int` if it is supposed to be 32 bits, since `int` will change size in the future to be 64-bit on 64-bit machines.

vlib/compress/deflate/deflate_compress.v
quaesitor-scientiam: Good call — done in 7c79ac384. `fixed_litlen_encode()` now returns `!([]u32, []int)` and `deflate_compress_fixed()` returns `![]u8`, propagating the error instead of panicking. The public `compress*` APIs are already `![]u8`, so nothing changes for callers. This also makes the encode side consistent with the decode side, where `build_huff_tree()` already returns `!` in this PR.

(In practice the error is unreachable here — the fixed lit/len lengths are a constant valid code — but propagating it is the right convention and costs nothing.)

vlib/compress/deflate/deflate_compress.v
quaesitor-scientiam: **Two arrays:** both are consumed together by the bit writer — it needs the code *and* its bit length for every symbol (`w.write_bits(ll_codes[sym], ll_lens[sym])` in `deflate_compress_fixed`). The pair-of-arrays shape predates this PR (`fixed_dist_encode` returns the same), and this PR only changed how the codes are computed. Happy to bundle them into a small struct (or return the `huffman.Table` directly, which already carries `codes` + `lengths`) if you'd prefer that shape — just let me know and I'll do it consistently for both `fixed_*_encode` functions.

**`i32` vs `int`:** agreed in principle, but the `[]int` here is load-bearing: it's the type of `hash.huffman.build`'s `Config.lengths` parameter (merged in #27392), and `int` is used for lengths/positions throughout `compress.deflate`. Switching just this function to `[]i32` would mismatch the builder's signature. I think the right move is a separate, module-wide `int → i32` pass (covering the builder's API and deflate together) rather than a one-off here — want me to file an issue to track that?

PR #27399, opened 2026-06-09
markused: fix generic fn instantiation dedup key collision for identical module names

### LINKED ISSUE #27398: V Compiler: Generic function instantiations silently dropped when multiple modules share the same name
### Describe the bug

When a V project contains **two or more modules with the same module name** (declaring identical `module <name>`) located in **different directory paths**, and both modules:

1. Define **struct types with identical short names** (e.g. both define `LoginByAccountResp`)
2. Call **the same generic functions** with those types (e.g. `json.decode[T]`, `json_success_200[T]`, `ctx.json[T]`)

...the V compiler may **fail to generate C code** for some of the generic function instantiations. The C compilation then fails because the generated C code calls functions that were never emitted.

The bug is **non‑deterministic** — it triggers with higher probability when more duplicate‑name module pairs exist (threshold ≈ 6+ pairs), but can occasionally trigger with as few as 5 pairs.

### Reproduction Steps

A self‑contained reproduction is available at `repro/`:

[generic_bug.zip](https://github.com/user-attachments/files/28744045/generic_bug.zip)

```
repro/
├── v.mod
├── main.v
├── common/api/          (generic struct + generic helper functions)
├── structs/             (Context embedding veb.Context)
├── route/               (24 route files, each importing one module)
├── m0a/common/ ~ m11b/common/
│                        (12 module pairs = 24 modules)
│                         all named `common`, all with same 5 type names
```

Compile:

```bash
cd repro/ && v .
```

Expected output: `ok` (clean build)

Actual output (100% reproducible with 12+ pairs):

```
cc: ... warning: implicit declaration of function
     'common__api__json_success_200_T_m6a__common__Type_cResp'
cc: ... error: cannot convert 'int' to
     'struct common__api__ApiSuccessResponse_T_m6a__common__Type_cResp'
```

The exact function that goes missing varies between runs (`json_success_200<...>`, `json.decode<...>`, `ctx.json<...>` for different type parameters).

### Expected Behavior

success

### Current Behavior

```
cc: /tmp/v_1000/repro.01KTNKD8B9FG731TRV2N6QT47H.tmp.c:37783: warning: implicit declaration of function 'x__json2__decode_T_m7a__common__Type_dReq'
cc: /tmp/v_1000/repro.01KTNKD8B9FG731TRV2N6QT47H.tmp.c:37783: error: '{' expected (got ";")
...
cc: /tmp/v_1000/repro.01KTNKD8B9FG731TRV2N6QT47H.tmp.c:37783: error: '{' expected (got ";")
(note: the original output was 21 lines long; it was truncated to its first 2 lines + the last line)
C compiler bug report was not sent to https://bugs.vlang.io/bug-report: v bug-report-send: net.mbedtls SSLConn.dial, failed to connect to host
builder error: 
==================
C error found while compiling generated C code.
This can be caused by invalid C interop code, C compiler flags, or a V compiler bug.
If your code is pure V and this still happens, please report it using `v bug file.v`,
```

### Possible Solution

The V compiler uses a **type‑name‑keyed map** internally to track which generic function instantiations need to be generated in C code.

When two modules in different directories both declare `module common`, and both define a type called `Type_aResp`, the compiler's internal type‑tracking map uses the **short type name** (`Type_aResp`) as the key, rather than the fully‑qualified name (`m0a.common.Type_aResp` vs `m0b.common.Type_aResp`).

As a result, the second module processed **overwrites** the first module's type entry in the map. The C code generator then only emits ONE C function for `json_success_200<Type_aResp>` (for the "winning" module), while both modules' generated C code call this function. The losing module's call site references a function that was never generated, causing the C compiler to report an implicit declaration error.

This explains all observed behaviour:

| Observation | Explanation |
|---|---|
| Non‑deterministic | File processing order varies between runs |
| Some instantiations work, some don't | Which module "wins" per type depends on processing order |
| Different generic functions affected (`json_success_200`, `json.decode`, `ctx.json`) | All use the same internal type‑tracking map |
| Error only in C compilation, not V compilation | V sees both calls as valid; C is missing the actual function body |
| More module pairs = higher failure rate | More collisions increase the chance of at least one being wrong |

### Workaround

Avoid sharing short type names across modules that have the same module name. Use unique struct type names for each module, or avoid having two modules with identical names.

In the specific project that triggered this bug, the workaround applied was:

1. **Replace `api.json_success_200(result)` calls with direct `api.ApiSuccessResponse[T]{...}` struct literals** — this avoids calling the generic function, so the compiler only needs to instantiate the generic struct (which it does correctly).
2. **Add `const` instantiation‑forcing for `json.decode[T]`** calls that cannot be eliminated.

A cleaner fix would be for the V compiler to key generic instantiations by **fully‑qualified type name** (module path + type name) rather than by the short type name alone.

_No response_

### V version

V 0.5.1 a063235

|V full version      |V 0.5.1 3767be3e1f7b0cbbfe0d938c69d944e687f2e601.a063235
|Memory              |8.68GB/31.02GB
|                    |
|V last modified time|2026-06-09 05:52:31
|                    |
|                    |
|V git status        |0.5.1-1874-ga0632356
|.git/config present |true
|                    |
|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.38-6deepin24) 2.38

> [!NOTE]
>

Jengro777:
When a V project has two or more modules that share the same name (e.g., both declare `module common` in different directory paths like `m0a/common/` and `m0b/common/`), and both define struct types with identical short names, the compiler silently drops some generic function instantiations. The C compilation then fails with "implicit declaration" / "undefined symbol" errors because the generated C code calls functions that were never emitted.

This occurs because the markused walker's `fn_generic_types_key` function uses `type_to_str(typ)` (the V type name) to build deduplication keys. When two types from different same-named modules share the same short name, `type_to_str` can return identical strings, causing the walker to treat subsequent generic instantiations as duplicates and skip them.

## Root Cause

In `vlib/v/markused/walker.v`, the `fn_generic_types_key` function:

```v
fn (w &Walker) fn_generic_types_key(fkey string, concrete_types []ast.Type) string {
    mut parts := []string{cap: concrete_types.len}
    for typ in concrete_types {
        parts << w.table.type_to_str(typ)  // ← uses V type name
    }
    return '${fkey}:${parts.join('|')}'
}
```

`type_to_str` returns the V-level type name string. Types from modules with the same declared name but different import paths could produce the same key string, causing the dedup check `used_fn_generic_type_keys[key]` to incorrectly skip needed instantiations.

## Fix

Use `typ.str()` — the underlying Type ID integer — instead of `type_to_str`:

```v
-   parts << w.table.type_to_str(typ)
+   parts << typ.str()
```

Type IDs are unique across all type symbols, guaranteeing that every concrete type produces a distinct key regardless of overlapping module paths or short names.

## Reproduction

A test with 6 module pairs (12 modules total, all named `common`, each defining the same 3 struct type names, each calling the same generic API function) consistently fails before this fix:

| Before | After |
|--------|-------|
| 30/36 C functions generated | 36/36 C functions generated |
| 6 "undefined symbol" link errors | Clean compile, outputs "ok" |

## Tests

- Reproduction case compiles and runs successfully
- `examples/hello_world.v` compiles and runs
- Relevant compiler test suites pass

## Changed Files

- `vlib/v/markused/walker.v` — 1 line change

Fixes #27398.


Jengro777:
This can solve the problem, but it may make the generated c unreadable

GGRei:
About the generated C readability concern and from what I can see, this key is an internal `markused` tracking/dedup key for generic instantiations. It does not feed into the generated C symbol text. The C names are still produced by the normal Cgen naming path, so this change should not make generated C names unreadable. ^

For the fix itself, maybe a slightly clearer version would be to use `u32(typ).str()` instead of `typ.str()`, since the goal here is to use the packed `ast.Type` identity, not the debug string formatting. :)

Also, if possible, adding a small regression test?

Nice catch for the bug/fix.

Jengro777:
> About the generated C readability concern and from what I can see, this key is an internal `markused` tracking/dedup key for generic instantiations. It does not feed into the generated C symbol text. The C names are still produced by the normal Cgen naming path, so this change should not make generated C names unreadable. ^
> 
> 
> 
> For the fix itself, maybe a slightly clearer version would be to use `u32(typ).str()` instead of `typ.str()`, since the goal here is to use the packed `ast.Type` identity, not the debug string formatting. :)
> 
> 
> 
> Also, if possible, adding a small regression test?
> 
> 
> 
> Nice catch for the bug/fix.


thankyou
I'll fix it, but adding tests doesn't seem appropriate to me because this is a hash collision issue and it behaves differently on different machines

GGRei:
Yes but the failure may vary between machines in terms of which specialization is missing but the expected behavior is not machine-dependent. No required generic specialization should be silently dropped.

So the test does not need to check the exact missing symbol. It only needs to build a reduced project with the same pattern and assert that it compiles/runs.

Of course, this is only my personal opinion, so feel free to handle it however you think is best :)

Jengro777:
> Yes but the failure may vary between machines in terms of which specialization is missing but the expected behavior is not machine-dependent. No required generic specialization should be silently dropped.
> 
> 
> 
> So the test does not need to check the exact missing symbol. It only needs to build a reduced project with the same pattern and assert that it compiles/runs.
> 
> 
> 
> Of course, this is only my personal opinion, so feel free to handle it however you think is best :)




thank you
I need to think about this, because even on my machine, I have to run it multiple times to consistently reproduce the bug. Maybe we could create a bigger example for testing, But it's not suitable to put it in the test file

Jengro777:
In order to reproduce the bug on every machine as much as possible, the test files are deliberately written here



Please review

GGRei:
The test fixture feels a bit large. 

Since `projects_that_should_compile_test.v` already creates temporary projects in other tests, maybe this case could generate the duplicate-module project under `os.vtmp_dir()` with a deterministic loop and a fixed pair count, for example 20/24/32 pairs, then run it with `-skip-unused run` flag and assert `ok`.

That would keep the compile/run regression strong, while avoiding ~5k LOC of mechanical fixture files. 

If CI proves the pair count is too low, it can be raised by changing one constant instead of adding more files?

24 pairs seems like a good compromise, I tested it locally.

Jengro777:
> The test fixture feels a bit large.
> 
> Since `projects_that_should_compile_test.v` already creates temporary projects in other tests, maybe this case could generate the duplicate-module project under `os.vtmp_dir()` with a deterministic loop and a fixed pair count, for example 20/24/32 pairs, then run it with `-skip-unused run` flag and assert `ok`.
> 
> That would keep the compile/run regression strong, while avoiding ~5k LOC of mechanical fixture files.
> 
> If CI proves the pair count is too low, it can be raised by changing one constant instead of adding more files?
> 
> 24 pairs seems like a good compromise, I tested it locally.



Thank you for reviewing
I have now switched to a different testing method
The implementation method of V2 is completely different, so this problem should not occur

GGRei:
@codex review

GGRei:
@Jengro777 can you try launching codex review please?  :)

Jengro777:
> @Jengro777 can you try launching codex review please?  :)


I used Codex Review locally and it worked fine

Jengro777:
https://github.com/Jengro777/v/tasks/727da2a5-de44-433d-9718-555574646b21

Jengro777:
@codex review

Jengro777:
@codex review

medvednikov:
@GGRei I've sent you an invitation via github, please accept.

GGRei:
@medvednikov Accepted, thank you for the invitation. I’m honored. :)

medvednikov:
avoiding ~5k LOC was a great call you made here :)

PR #27410, opened 2026-06-11
math.bits: cleaned up comments, added inline attrs, extra checks for 0, and tests

JalonSolov:
Also added casts to ensure number ranges, and changed one panic message to be correct.


medvednikov:
@codex review

PR #27442, opened 2026-06-13
checker: register nested generic str deps

### LINKED ISSUE #27441: cgen, generics: nested generic `str()` (a generic field whose type param is itself generic) is not emitted
### Description

When a generic struct's auto-generated `str()` method needs to stringify a **nested generic type** (a generic struct field whose type parameter is itself a generic instantiation), the `str()` method for that nested type is referenced but never generated, so the C compiler fails with an undeclared-function error.

**V version:** V 0.5.1, current `master` (`ddb021b98`)
**OS/Backend:** reproduced on macOS arm64; originally reported on macOS/cc.

### Reproducer

```v
import datatypes

struct MyNode[T] {
	val T
}

struct MyContainer[T] {
	lst datatypes.LinkedList[MyNode[T]]
}

fn main() {
	c := MyContainer[string]{}
	println(c)
}
```

Compile with:

```
v file.v
```

### Current behavior

```
error: call to undeclared function 'datatypes__LinkedList_T_main__MyNode_T_string_str_T_main__MyNode_T_string'
  string _t1 = datatypes__LinkedList_T_main__MyNode_T_string_str_T_main__MyNode_T_string(it.lst);
```

The auto-`str` for `MyContainer[string]` calls the `str()` of its `LinkedList[MyNode[string]]` field, but that nested generic `str()` instantiation is never emitted.

### Expected behavior

The `str()` method for the nested generic type (`LinkedList[MyNode[string]]`) should be generated, so printing the container compiles and works.

### Notes

- Reported via the bugs.vlang.io crash reporter (3 reports). Original file name: `generic_struct_parameter_err`.

> [!NOTE]
>

medvednikov:
Fixes #27441.

## Summary
- register generic `str()` dependencies discovered while marking auto-str usage
- recursively inspect nested field/container types so generated auto-str can call generated generic `str()` methods
- add regression coverage for `LinkedList[MyNode[T]]` inside a generic struct

## Testing
- Not run during PR creation, per request


medvednikov:
@codex review

PR #13, opened 2019-03-10
Fix generic docs after pull #10

ntrel:
Resolve some issues discussed in pull #10.

Issue #27116, opened 2026-05-09
Restore CI step: Ensure V2 can be compiled with -autofree


medvednikov:
The CI step `Ensure V can be compiled with -autofree` in `.github/workflows/other_ci.yml` (`./v -autofree cmd/v`) has been temporarily commented out because `-autofree` produces a compiler that cannot reliably self-compile or compile non-trivial programs.

## Current status

`v_af` (a `cmd/v` build with `-autofree`) can compile a trivial `hello world` end-to-end, but aborts with double-free / use-after-free errors on programs that exercise more parser paths (e.g. struct + method + array).

Several distinct autofree-codegen bugs surface in sequence:

1. **Shared sub-buffers via shallow array push** — e.g. `params << params_t` and `for attr in attrs { p.attrs << attr }`. The destination's elements share string buffers with the source; freeing both at scope end double-frees.
2. **Non-owning pointer locals** — e.g. `for sc := s; true; sc = sc.parent { ... }`. `sc` is a non-owning `&Scope`, but autofree generates `Scope_free(sc)` because `Scope` has a user-defined `free()` method.
3. **Sumtype payload double-free** — auto-generated sumtype `_free` calls `_v_free` on the heap-boxed variant pointer, but the same pointer is shared across copies of the sumtype value (e.g. `empty_scope_object` referenced by every default-constructed `Ident.obj`).
4. **Pointer struct fields treated as owned** — `gen_free_for_struct` recursively frees fields like `&Scope`, but pointer fields in V are conventionally non-owning, so this corrupts shared state.
5. **Locals leaked into globals/struct fields** — e.g. `lines_cache.lines[path] = lines`, `p.table.module_attrs[p.mod] = module_attrs`, `&ast.File{stmts: stmts.clone()}` — autofree of the local frees buffers still referenced through the leak target.

## What's needed

Each fix here is partial — every bug fixed exposes a deeper one. A more principled approach is needed than the current shallow-detection patches:

- A real ownership model in `-autofree` (escape analysis for locals stored into external storage)
- Idempotent `_free` for shared types (Scope, sumtype payloads, etc.)
- Don't auto-emit `_free` for pointer struct fields by default
- Don't `_v_free` the variant box inside auto-generated sumtype `_free`

## Reproducer

```sh
./v -o v_af -autofree cmd/v
echo 'fn main() { println("hi") }' > /tmp/h.v
./v_af -o /tmp/h /tmp/h.v   # works
cat > /tmp/t.v << 'V'
[...]

PR #27458, opened 2026-06-15
Architecture E (POC, opt-in `-gc e`): Perceus reuse-in-place front line + precise STW tracing GC backstop for the C backend

### LINKED ISSUE #2: Can you release the closed-source compiler right now for us to play?
I am not crazy about open-source unlike others so would be okay with trying your compiler- I want to see the GUI story.

### LINKED ISSUE #6: .v file extension is Verilog
Hi, this looks like a cool language.  However using .v is not a wise choice, as this has been used for the Verilog language for the past 25 years.

https://fileinfo.com/extension/v

eptx:
# Architecture E: a Perceus front-line + precise stop-the-world tracing backstop for V's C backend

## What this is (please read first)

This is a **proof-of-concept, developed with Claude (Anthropic's coding agent)**, shared
because **the results look very positive for V and we'd like the community to verify
them**. We (the CX project — a tree-walking language interpreter written in V) set out to
test whether V could meet our memory-management and multi-core needs *instead of*
switching to Rust, while staying aligned with V's stated direction (autofree /
reuse-in-place). It worked well enough to be worth contributing — at minimum a baseline
POC, possibly a real contribution.

**Direct about the claims:** every performance number below was measured **on our machines
and our workloads only** — no broad independent benchmark suite, no third-party review.
Treat them as *claims to verify*, not facts. The correctness work is firmer (TSan, a
deterministic white-box self-check, and a churn reproducer — all included) but also wants
independent eyes. We'd value the community pressure-testing both.

Context: follows up on the Perceus discussion **#27166** (and a Discord exchange where
@JalonSolov suggested a PR so Alex could look it over). **Open question for maintainers
up front: target v1 (current master, where it's built + tested) or plan for v2?** If v2
reworks the backend/codegen we're happy to advise on a port — better to know before deep
review.

All changes are **provider-neutral V-runtime / codegen** work: CX was the workload that
surfaced the bugs and motivated the optimizations, but nothing here is specific to it
(source scrubbed of downstream-specific naming).

## Summary (TL;DR)

This adds a new opt-in memory-management mode for the C backend, **`-gc e`**, that pairs
a **Perceus-style reference-counting front line** (compiler-emitted, in-place reuse) with
a **precise, from-scratch stop-the-world tracing collector** (`vgc`) as the backstop —
plus the bug fixes and allocator optimizations that made the combination sound and fast
under heavy multi-threaded allocation. The front line reclaims the common, uniquely-owned
case with zero tracing; the backstop reclaims arbitrary aliased/cyclic graphs that RC
cannot. `vgc` alone is also usable (`-gc vgc`).

**Why**: Boehm (V's default conservative GC) anti-scales on alloc-heavy multicore
workloads (its parallel marker and alloc lock serialize mutators) and over-retains
(conservative). E targets *both* single-thread throughput (reuse-in-place avoids
allocation) and multicore scaling (per-thread allocation + accounting, no shared
alloc-path lock in the steady state).

**Status / honesty**: developed and gated against one large real consumer + a battery of
provider-neutral micro-benchmarks (included). The performance numbers below are
**measured on those workloads and should be independently verified** before any claim is
relied on. STW is the default collection strategy; concurrent mark is behind a separate
`-d vgc_concurrent` and is *not* proposed for default. The verification tooling
(mark-closure verifier, root-finder) is compiled out unless `-d vgc_verify`.

---

## Architecture & rationale

V today offers Boehm (conservative, default), `-autofree` (compiler-managed scope frees,
single-ownership assumption), and `-gc none`. None gives both *low allocation* and
*linear multicore scaling* for an allocation-heavy program with aliased/graph-shaped data.

**Architecture E = front line + backstop, decoupled from `-autofree`:**

1. **Perceus front line** (`vlib/v/gen/c/perceus.v`, new). A compile-time ownership/share
   analysis emits in-place **reuse** and **drop** for values it can prove uniquely owned,
   reclaiming the common case without touching the collector. Crucially it is **decoupled
   from `-autofree`**: `-autofree` restructures codegen assuming sole ownership and is
   incompatible with a backing collector (corrupts under *any* GC — demonstrated). E runs
   the drop analysis off its own `perceus` define, so Perceus drops are the sole frees and
   the analysis stays sound (it pins assignment-aliases, call-result aliases, and any
   value whose heap field is exposed → those fall through to the backstop).

2. **Precise STW tracing backstop** (`vlib/builtin/vgc_*.c.v`, new). A from-scratch
   mark/sweep collector with a **Go-`mcache`-style segregated allocator** (per-thread span
   caches, per-size-class central lists, arena-backed spans). It reclaims what RC can't
   (cycles, aliased graphs) and runs *rarely* because the front line absorbs most frees.
   Precise (type-driven) marking where sound; conservative stack/register scanning for
   roots. Mutators are stopped via OS-level suspend (mach / signal).

3. **The hybrid is the point.** RC alone leaks cycles; tracing alone pays full mark cost
   on every cycle. Perceus handles the dominant uniquely-owned case in-place; the tracing
   backstop is the correctness net for the rest. This mirrors Koka/Lean's Perceus + a
   collector, adapted to V (which lacks a uniform per-object header, so the backstop owns
   arbitrary-graph reclamation rather than a global RC header scheme).

**Isolation-for-scaling doctrine**: linear multicore scaling comes from *per-thread
isolation of the allocation path* (per-thread span caches + per-thread heap accounting),
not from a faster shared collector. The collector is the rare backstop; the steady-state
alloc/free fast path touches no shared cacheline or lock.

---

## Bugs fixed (correctness)

Each is provider-neutral and was reproduced under heavy concurrent alloc/free (a
multi-reactor HTTP server + churn micro-benchmarks). The commit hashes below are the
**originating development commits**; each commit on this PR branch carries a
`(cherry picked from commit …)` trailer, and the PR's *Commits* tab is the authoritative
per-change view.

| # | Fix | Commit | Notes |
|---|-----|--------|-------|
| 1 | Collector self-scan anchored at the **real SP**, not the frame pointer | `50fde691` | `setjmp` spills callee-saved regs *below* the FP; an FP-anchored scan missed a live root held only in a spilled reg → reclaimed-while-live. |
| 2 | Advance `gc_cycle` + GC trigger **under STW**, before resuming the world | `8baa8db0` | A resumed mutator stamping a fresh span's `sweep_gen` with the old cycle let the next sweep recycle a still-in-flight span (UAF). TSan: 30→0 races. |
| 3 | Publish `narenas` with **release/acquire** | `c39ce23f` | Lock-free `vgc_find_span` read `narenas` while `span_alloc` wrote it under lock — publication race on the arena it gates. TSan-pinpointed. |
| 4 | Publish `page_span` slots with **release/acquire** | `46d2ae5a` | Spans carved from an existing arena don't bump `narenas`, so the page-map writes weren't published to the lock-free `find_span` reader → stale span. |
| 5 | mcache bitmap **atomic** `fetch_or`/`fetch_and` + atomic count | `46d2ae5a` | Unlocked RMW on `alloc_bits` (alloc fast path) raced a cross-thread `free`'s RMW under the central lock → one slot handed out twice. |
| 6 | `vgc_span_alloc_obj` two-pass scan **start-byte** coverage | `871dceda` | The free-index offset was applied in both passes, leaving `[0,start_bit)` of the start byte scanned in neither → a span with a free low slot reported "full" → `vgc_malloc` NULL → caller null-deref. |
| 7 | Don't reclaim **mcache-resident** spans in sweep | `871dceda` | A cached span that momentarily empties was recycled while still referenced by an mcache slot / a suspended owner's local (span descriptors live outside the GC arena, so the root scan can't protect them). Fix: stamp registered threads' cached spans' `sweep_gen` under STW. |
| 8 | **Conservative** backstop mark; drop the unsound per-span `ptrmap` | `004b02f2` | `ptrmap` was a per-*span* property set by the first typed alloc, but a size class packs many types → objects whose layout differed had live child pointers skipped → reclaimed-while-reachable. Conservative scanning over-retains, never under-retains. |
| 9 | Sound eager-drop for **aliased call results** + `?&T` free-method codegen + `vgc_free` central lock | `38607b2d` | (a) a heap value bound from a call may alias the callee's traversed sub-objects → pin it (backstop, don't deep-drop); (b) option-of-pointer free emitted a struct member-access on the `_option_*` wrapper (C compile error); (c) `vgc_free` now takes the per-class central lock (was a real MP soundness gap). |
| 10 | Option-aware free methods for `?SumType` / `?[]T` fields | `26ac2bbe` | `gen_free_for_sumtype`/`_array` emitted `it->_typ`/`it->len` on the `_option_*` wrapper → C error for any program freeing such a field under autofree/Perceus. |
| 11 | `contains_ptr` treats `?T` / `!T` as pointer-bearing | `d112d5c8` | `[]?int` was flagged noscan (the option strips to `.int`), but `_option_int` carries an `IError` pointer → a pointer-bearing object marked noscan. |
| 12 | Four `-gc e` correctness fixes: map tiny-free, Perceus drop, HEAP_vgc arity, overflow-thread panic | `a95aff916b` | Incl. the >`vgc_max_threads` case that indexed `caches[-1]` and recursed through malloc in the panic path. The HEAP_vgc-arity fix alone cleared 22 of 34 of V's own `-gc e` test failures. |
| 13 | Drop extraneous `)` when freeing an **option-pointer local** (`b := &?Foo{}`) | `3bcf843fb9` | The option branch closed the free call's paren and the shared tail closed it again → `free((Foo**)b.data));` C error. Fixes `option_init_ptr_test` under `-gc e`; boehm/none unaffected. |
| 14 | Generate the **option-element free for `[]?T`** | `3d537762` | An array of `?string` referenced `_option_string_free`, a wrapper no path generated (the unwrapped sym has a user `free` → string-construct branch). Now inline the option-element payload free. Fixes `option_ifguard_array_of_option_test`. |
| 15 | **Atomic `live_threads`** in register/unregister | `c69fd59b` | `vgc_maybe_gc` reads `live_threads` lock-free for per-thread GC pacing; the plain `++/--` raced that atomic read (TSan-flagged). |

---

## Optimizations (performance)

| # | Optimization | Commit | Claimed effect (verify) |
|---|--------------|--------|-------------------------|
| A | **Per-thread heap accounting** (Go per-P style): alloc/free bump thread-private `live_delta`/`alloc_delta`, flush to the global atomics only every ~1 MB | `677770dd` / `38607b2d` | Removes global-atomic cacheline contention on the accounting path. |
| A2 | **Lock-free free fast path** for mcache-resident / dropped spans (`on_central == 0`): `vgc_free` skips the per-class `central[].lock` (kept only for spans actually on a central list); bitmap+count stay atomic, the `fetch_and` prior value gates the decrement (double-free-safe) | `c69fd59b` | The residual-#4 fix had added that lock to *every* non-tiny free → a same-class free storm (`bench_scalar`: 8 threads alloc+drop one 32 B class) serialized N-way and anti-scaled (35→7 Mops/s T1→T8, below Boehm). With the skip it is **near-linear again: 45→326 Mops/s T1→T8 (7.2×, ~5.5× Boehm at T8)**; `bench_mp` T1 5.9→76. Verified residual-#4-safe (white-box selftest + container churn 15 rounds niltrace=0 + TSan 0). |
| B | **Perceus deep-free** of nested heap fields of a dropped `&Foo`, gated by a sound deep-drop analysis | `677770dd` | Nested-object MP T1→T8 ~7.5× (near-linear); removes the GC pressure that compounded MP contention. |
| C | **In-place reuse**: direct indexed stores for reused map slots | `d7e9f5a1` | Avoids re-hashing on reuse-in-place. |
| D | **Dynamic span registry** (mmap-backed, lazily committed) + env-gated GC pacing | `72edb9e5` | Removes a fixed 262k-span cap; lets the trigger scale without a hard abort. |
| E | **Concurrent tri-color mark** behind `-d vgc_concurrent` (opt-in, STW stays default) | `d22ae0ee` | ~1.2–1.4× on a parallel alloc-heavy fold vs STW; *not* proposed for default (needs a sound GC-assist first). |
| F | **Alloc-path lock-contention removal** (B18): span-descriptor **bump slab** (drops a per-carve `mmap` from under the heap lock; ~3× lower RSS) + **drop full spans** instead of returning them to the central full-list (the never-reused per-fill central-lock traffic) + **per-thread GC pacing on by default** (adaptive — only when >1 mutator) | `82e39343` | Parallel alloc-heavy workload recovered from anti-scaling (≈parity) to ~3.8× its serial at a high trigger; ~3× lower RSS. |

Profile evidence: under parallel churn the alloc fast path was ~98% spin on two global
locks (`vgc_heap.lock` for span carving — whose hold included an `mmap` syscall — and the
per-class central lock for span return); 8 *separate processes* scaled but 8 in-process
workers did not, isolating the cost to in-process shared allocator state (not bandwidth).

---

## Soundness evidence

- **TSan** (Linux, clang `-fsanitize=thread`) found and confirmed fixes #2, #3, #5
  (race count → 0 after each).
- **Deterministic white-box self-check** for fixes #6/#7: `vlib/builtin/vgc_selftest_d_vgc.c.v`
  (driven by `bench/parallel-alloc/vgc_residual4_test.v`) — reverting either fix fails it.
- **Churn batteries** (provider-neutral, in `bench/parallel-alloc/`): `g_churn`
  (alloc/free/realloc storms, multi-thread), `bench_mp`/`bench_scalar` (MP alloc scaling),
  `par_live` (large concurrent live sets), `cm_stress` (concurrent-mark hazards),
  `cm_barrier_proto.c` (the write-barrier model with deterministic teeth).
- **Differential**: program output byte-identical across `-gc none` / `boehm` / `vgc` / `e`.
- Long multi-reactor HTTP soundness run: tens of millions of requests, crash-free.

## Scope / what to review carefully

- New GC backend is large; suggest reviewing `vgc_d_vgc.c.v` (allocator) and
  `vgc_gc_d_vgc.c.v` (collector) first, then `perceus.v` (analysis) and the codegen
  touch-points (`assign.v`, `auto_free_methods.v`, `autofree.v`, `cgen.v`, `fn.v`).
- **Not for default upstream**: `-d vgc_concurrent` (needs a sound GC-assist),
  `vgc_verify` tooling (debug-gated), and the experimental `cx_region.c.v` /
  transport-layer patches (consumer-specific; excluded from this proposal).
- Based on `a83aabb10f`; a rebase onto current master is required.
- Known follow-ups: **full deferred cross-thread free** (the targeted lock-free path #A2
  already covers owner-frees of mcache-resident spans — the dominant case; a complete
  mimalloc-style per-span atomic thread-free list would also make cross-thread frees of
  central-listed spans lock-free); sound
  concurrent-mark GC-assist (cooperative safepoints); generational
  option; and V's own `-gc e` codegen edge cases — **~10 of 2146** `vlib/v/tests` programs
  (all `-gc e`-specific, pass under `none`/`boehm`), characterized as three families:
  (1) **option-wrapper / generic / sub-module `_free` not generated** — e.g. an array of
  `?string` references `builtin___option_string_free` but the value-option wrapper free is
  never emitted (free-method generation vs `-skip-unused` DCE; the unwrapped element sym
  has a user `free`, so the option-wrapper free path is skipped);
  (2) **reflection metadata reclaimed** (4 reflection / generic-anon-fn tests segfault);
  (3) **Perceus string early-drop** (3 tmpl/comptime/interface-str tests produce
  truncated/aliased strings). These touch shared autofree/option/Perceus codegen
  (boehm-regression-sensitive) and runtime mark soundness — each warrants a dedicated pass,
  not bundled here. Fix #13 above cleared one (`option_init_ptr`).

## Test environment (so the numbers mean something — and what's NOT covered)

Everything below was measured on a **single machine**. This is a real limitation: we have
not tested other CPUs, x86, or native (non-virtualized) Linux. Please reproduce on your
own hardware.

- **Dev + all macOS benchmarks:** Apple **M2 Max**, 12 cores (8 performance + 4
  efficiency), 64 GB RAM, macOS 26.4.1 (build 25E253), Apple clang 21.0.0. `-prod`
  builds via `-cc cc`.
- **Linux correctness/concurrency testing:** a **Docker container (Ubuntu 24.04.4,
  clang 18.1.3, wrk 4.1.0), `aarch64`** — i.e. Linux 6.12 (linuxkit) running in Docker's
  VM **on that same M2 Max**, not a separate native or x86 host. TSan + the concurrent-
  HTTP churn reproducer ran here. So: **arm64 only; x86, native Linux, and other core
  counts are unverified.** The collector's conservative stack/register scan and the
  OS-suspend STW path are platform-sensitive — independent runs on x86/native Linux are
  exactly the verification we're asking for.
- Numbers are best-of-3 (compute benches) wall-clock; `-gc boehm` is the baseline.

## How to verify

```
# build the dev compiler, then for any program:
v -gc e   prog.v     # Perceus front line + vgc backstop
v -gc vgc prog.v     # backstop only
v -gc e -d vgc_concurrent prog.v   # opt-in concurrent mark
# benches (provider-neutral):
v -enable-globals -gc e bench/parallel-alloc/poc/bench_scalar.v   # MP alloc scaling vs boehm
v -enable-globals -gc e bench/parallel-alloc/poc/bench_mp.v
v -gc e test bench/parallel-alloc/vgc_residual4_test.v        # white-box fix self-check
```


eptx:
Thanks for approving the CI run — that surfaced exactly what we needed.

I triaged the 38 failures; they collapse to **four root causes, all on our side and all low-risk**, now addressed on the branch:

1. **`-os cross` / bootstrap break (real regression, the important one).** The concurrent-mark write barrier `vgc_wb_store` is defined only under `-d vgc` (`-gc e`), but its call sites in `array.v`/`map.v` sit in `$if vgc_concurrent ?` blocks. `v -os cross` emits *every* comptime branch into `v.c`, and the checker walks inactive branches too — so in a default (boehm) build the symbol is unresolved and `bootstrap-v`, `build-vc`, `cross-*` and the BSD cross-compiles fail with `unknown function: vgc_wb_store`. Fixed by adding a no-op `vgc_wb_store` fallback in a `_notd_vgc.c.v` sibling (mutually exclusive with the real definition via file suffix; zero behavioral change — the barrier is only ever live under `-gc e -d vgc_concurrent`).

2. **`vfmt`.** `vlib/v/gen/c/perceus.v` and `cgen.v` weren't `vfmt`-clean. Since `test-cleancode` runs early in most jobs, this alone reddened the linux/macos/windows compilers, the docker images, and the sanitize-* and tcc-* jobs (they abort at the fmt gate before reaching their namesake work). Reformatted.

3. **`unused parameter: typ`.** `write_heap_alloc[_close]` no longer use their `typ` arg (the precise-pointer-map `HEAP_vgc` variant is intentionally dead — see the note there). V prints a notice to stderr even on a successful build, and the `tools-*` harness treats any stderr from a tool compile as failure, so the whole `tools-*` matrix (and the riscv64 build) went red. Marked the param `_`.

4. **Missing doc comments** on `pub fn vgc_init` / `vgc_set_watch`. Added.

Worth flagging: this CI matrix builds V's **default** configuration, so the `-gc e` code paths (C11 atomics, `@[thread_local]`, the conservative scanner) aren't actually compiled or exercised by these jobs — the failures above are all default-build lint/cross issues, not anything specific to the new backend. If it'd be useful, I'm happy to add a small opt-in `-gc e` lane so the new collector gets real coverage; the two known limitations there are (a) tcc can't compile the C11 atomics the barrier/STW code uses, so a `-gc e` lane would need a non-tcc compiler, and (b) the conservative mark scan will trip ASan/MSan/UBSan without suppressions. I'd rather take your steer on whether you even want that lane than guess.

A couple of things I'd like direction on before going further:

- **Fallback approach for (1):** the no-op `_notd_vgc` stub is the minimal fix, but if you'd prefer the barrier machinery never reference a vgc-only symbol from `builtin` at all (e.g. gating the call sites differently), I'll restructure to match your conventions.
- **v1 vs v2:** this POC targets the current C backend (`vlib/v/gen/c`). Is that the right place for an experiment like this, or would you want it oriented toward `v2`?

This is still a POC / "needs verification" — happy to adjust scope, split it, or hold any of it pending your read on the architecture.

JalonSolov:
So... I decided to run `-gc vgc` vs `-gc boehm` (I know it's the default... I made it explicit...) vs `-gc e`,  I compiled V itself after checking out the PR, and ran the test with hyperfine... here are my results for bench_scalar:
```
Summary
  ./ve -enable-globals -gc boehm run bench/parallel-alloc/poc/bench_scalar.v ran
    1.31 ± 0.07 times faster than ./ve -enable-globals -gc vgc run bench/parallel-alloc/poc/bench_scalar.v
    1.39 ± 0.08 times faster than ./ve -enable-globals -gc e run bench/parallel-alloc/poc/bench_scalar.v
```
and for bench_mp:
```
Summary
  ./ve -enable-globals -gc vgc run bench/parallel-alloc/poc/bench_mp.v ran
    1.02 ± 0.04 times faster than ./ve -enable-globals -gc boehm run bench/parallel-alloc/poc/bench_mp.v
    1.07 ± 0.02 times faster than ./ve -enable-globals -gc e run bench/parallel-alloc/poc/bench_mp.v
```
Which means on my system, this new way is slower than both the others.  😕 

Latest V, CachyOS, on 7950X CPU, 32G RAM, and NVME disks.

JalonSolov:
Alex will have to make final decisions, but my opinion:

* never reference a vgc-only symbol from builtin at all, unless `-gc vgc` is used
* implement for both backends

v2 is still not quite done, yet, so we can test earier/more stably on v1.  However, checking again v2 _before_ it is finalized could be very useful, as well.

GGRei:
I think the core idea is interesting, Perceus-style reuse, ownership-driven drops/frees, and a tracing backstop are all directions worth studying.

My concern is more about integration and timing. I have been doing some work on the V2 ownership/autofree direction based on the ownership problems listed in #27116.

In the current V2 work, the scope is deliberately staged, it collects ownership/autofree facts from the V2 post-transform FlatAst and type information, classify transfers and release/cleanup eligibility in the V2 type/checker layer, then emit bounded cleanup first through the V2 CleanC backend and fixture tests. The goal, however, should be a backend-neutral V2 ownership/free contract. Once the model is stable, the same ownership decisions will need clear lowering rules for the native backends too, including x64, arm64, and any future backend. That model needs to cover shallow copies, escaping locals, non-owning pointers, sumtype payloads, pointer fields, globals/struct storage, etc.

That makes me wary of adding another memory-management path before the V2 model is implemented, reviewed, and stabilized.

For the current C backend side, I also think this is still too experimental for `master` as-is. It is true that V already has several memory-management modes there, such as Boehm, none, and autofree, but autofree is already a sensitive area with known issues in that path.

This PR is not just adding a small isolated switch, it also changes the current C backend code generator, autofree/free-method generation, builtins, VGC runtime behavior, platform STW/root scanning, allocator behavior, and Boehm/libgc-related code. Even if `-gc e` is opt-in, that maintenance surface is not really isolated, so I am cautious about adding a new experimental hybrid mode.

So in my opinion, the safest path would be to keep this as a dedicated experimental/RFC branch for now, possibly as a dedicated branch under the V org if maintainers want to explore it, and extract the useful parts separately like benchmarks, reproducers, design notes, and small isolated fixes with tests. If the experimental branch later proves strong results across supported OSes, with correctness tests and stable performance data, then moving parts of it toward the current C backend / master path would make much more sense.

This is only my personal opinion, of course. ^^

Just to be clear, I still think this PR is interesting and definitely worth exploring further.

eptx:
Really appreciate the careful look — and thank you for taking the time to actually benchmark it, @JalonSolov.

**On performance:** that's a fair and important data point. Our wins were measured on M2 Max (arm64/macOS); your 7950X / CachyOS numbers — `-gc e` behind *both* boehm and vgc on bench_scalar and bench_mp — tell me the result is workload- and platform-dependent, not a universal win, and I won't claim otherwise. It would need broad cross-OS / cross-arch data before being a serious `master` proposal, and it doesn't have that yet.

**On integration & timing, @GGRei:** I think you're right, and I'd rather follow your lead than push a ~5k-line cross-cutting change at `master`:
- Happy to keep the full hybrid as a dedicated **experimental / RFC branch** (under the V org if you'd like a home for it; otherwise my fork), explicitly *not* targeting `master`.
- I'd prefer to peel off the independently-useful pieces as **small, isolated, tested PRs** — the benchmarks + reproducers, the design notes, and the few genuinely isolated fixes we hit along the way. Would those be welcome, and do you want them as separate PRs?
- I also don't want this competing with the **v2 ownership/autofree** direction in #27116 — that's clearly the strategic path. I'll exercise the ideas against v2 early (per @JalonSolov) so anything useful feeds *into* that model rather than around it.

**On the builtin reference (@JalonSolov):** agreed — I'll rework so the barrier never references a vgc-only symbol outside `-gc vgc` builds (gating the call sites, rather than the builtin no-op fallback I pushed just to clear CI), and look at covering both `vgc` and `e`.

I'll re-frame the PR accordingly. Thanks again — this is exactly the steer I was hoping for.

GGRei:
Of course, please keep in mind that I am nobody special in the project, just a modest contributor giving my personal opinion. 

Medvednikov is the real final decision-maker here, with JalonSolov's help of course.

medvednikov:
@GGRei you are correct. We should at least wait until the full transition to V2. But definitely something to explore.

eptx:
### Progress update — a multi-mutator soundness (and macOS perf) fix in the `-gc e` STW

Following the close (agreed: this stays experimental and waits for V2 — I'm not reopening the merge question), I wanted to put the latest work on the record in case it's useful when you do explore this, since it turned out to be a concrete, isolated correctness finding in the `-gc e`/`vgc` stop-the-world that's relevant to V's MP story.

**TL;DR:** the precise STW collector had a **macOS-only soundness bug** under ≥2 concurrent mutators + frequent collection — a peer thread's register-resident live pointer was intermittently missed at suspend, then swept while live (UAF). Linux was always sound. Fixed by giving macOS the same signal-based suspend the Linux path already uses. One file, isolated, with a reproducer.

#### The bug (in V terms)
On darwin the STW stops peers with the **asynchronous** `thread_suspend()` and reads their SP/registers from an external `thread_get_state()`, gated only by a heuristic spin-until-(SP,PC)-stable settle (`thirdparty/vgc/vgc_platform.h`, the `vgc_suspend_thread` at the `thread_suspend((thread_act_t)t)` call). That settle is a **weaker stop guarantee than a kernel-delivered signal point**: with N≥2 allocating mutators and frequent GC, a peer can be captured at an instant where a live pointer lives only in a register the scan then misses → the object is reclaimed while reachable → sweep-while-live UAF.

Two things worth flagging because they rule out the obvious explanations:
- **Not register coverage.** macOS already captures *more* than Linux (x0–x28 + fp + lr + all 64 NEON lanes) and still failed. Adding registers was never the issue.
- **Linux was sound the whole time.** The Linux signal-suspend path (handler captures its own `ucontext` at a kernel-delivered boundary, ACKs, parks) showed **zero** events at ~20× the macOS exposure. The defect was specific to the async external-capture model.

#### The fix
Switch the darwin STW peer-suspend to a **signal-based mechanism mirroring the existing Linux path**: the collector `pthread_kill`s each peer (`SIGURG`, resolved from the mach port via `pthread_from_mach_thread_np`); the peer's handler captures its own register file + SP from the signal `ucontext` at a clean instruction boundary, ACKs, and parks until released; the collector spins on the ACK before trusting the captured state. The async `mach_suspend` path is kept behind `-DVGC_MACH_SUSPEND` for A/B. The Go runtime uses `SIGURG` for the same purpose, and it doesn't collide with the net path.

- Isolated to one file: `thirdparty/vgc/vgc_platform.h` (+151/−19), no V-layer change (the three C entry points keep the same signatures, keyed by the existing `mach_port`).
- Public branch/commit if useful to diff: `cx-home/v@7b15dbcec4` (branch `cx-home/v-cx-patches`).

#### Evidence (paired, same build, same workload: ≥2 reactors cloning string-keyed maps under churn-cadence GC)
A masking-proof oracle at the use site (a freed map-key buffer read in `map_clone_string`) + crash count:

| target | sweep-while-live events | crashes | throughput |
|---|---|---|---|
| macOS arm64 — async mach (before) | 4–8 / run | 2–3 | ~16–19 rps |
| **macOS arm64 — signal (after)** | **0** | **0** | **~370 rps** |
| **Linux arm64 (Docker)** | **0** | **0** | **~400 rps** |

Soundness was achieved at **~23× higher** macOS throughput — i.e. the async `mach_suspend` was *also* a per-GC bottleneck, not just a correctness gap; the fix brings macOS to Linux-class throughput. (This is a separate axis from @JalonSolov's single-process `bench_scalar`/`bench_mp` numbers on Linux/7950X, which use the already-sound signal path and are unaffected by this — I'm not claiming it changes that result.)

#### Why this may be useful for V's mem-mgmt + MP direction
- It's a **parallel-correctness** property of a precise tracing collector: the STW root scan must capture each suspended mutator at a point where its register-resident roots are recoverable. The **signal-suspend + ACK model is the safe canonical form on both platforms**; relying on async external `thread_get_state` capture is not. That contract is equally relevant to a future V2 backend's STW/root-scan lowering, whatever the ownership model on top.
- The reproducer methodology (a masking-proof *use-site* oracle, since a freed-buffer read is caught deterministically regardless of timing; plus a native-arm64 Docker gate) may be reusable for validating any precise/parallel GC work in V.

#### Honest scope
Validated on macOS-arm64 and Linux-Docker-arm64 only (not x86-64 macOS or other OSes). This does **not** revisit the "too cross-cutting for master / wait for V2" decision — it's offered purely as one of the isolated, tested pieces you suggested peeling off (a real bug + fix + reproducer in the `-gc e` STW), for whenever the exploration resumes. Happy to package it as a standalone PR with a minimal multi-threaded test if that'd be useful.

eptx:
Follow-up with a root cause + fix that may interest anyone building a
signal-suspend STW collector for V (or embedding one): we found and fixed a
use-after-free race family in our Go-style mark/sweep GC for V that took a
long forensic hunt precisely because the missed "root" wasn't a root at all.

**The race:** an async-signal stop-the-world (SIGURG + pthread_kill, the same
shape Linux debuggers/profilers use) freezes mutator threads at *arbitrary*
instruction boundaries — including inside the allocator's lock-free fast
paths. Two windows bite:

1. **Tiny-allocator cursor invalidation.** Our mcache has a Go-style tiny
   allocator (`cache.tiny` + offset packing sub-16B objects into one slot).
   The collector defensively zeroed every thread's tiny cursor under STW
   (post-sweep). A mutator frozen *between* its `cache.tiny != 0` check and
   the `cache.tiny + off` carve resumes and computes `ptr = 0 + off` — a
   near-NULL pointer handed out as a live buffer. We caught this live in
   lldb: a string-clone writing through `res.str = 0x8`. The zeroing was also
   unnecessary in our design (the caches array lives in a conservatively
   scanned global, so the tiny block is rooted while cursor'd).

2. **Mid-claim sweep theft.** The allocation fast path claims a slot with an
   atomic OR on the alloc bitmap, and only *afterwards* does the object's
   address exist anywhere a conservative scan can see (it is still implicit
   `span + index` register state). A thread frozen in that window leaves the
   slot `alloc=1, mark=0` — indistinguishable from garbage — so the sweep
   clears the bit under the in-flight claim, and the slot is later handed out
   a second time: two live owners, torn structs, use-after-free. Note
   Go-style allocate-black can't close this window: any phase check at the
   claim runs *before* the freeze, and the mark-bit wipe at cycle start
   erases pre-set marks.

**The fix** (small): never invalidate allocator state a frozen thread can be
mid-use of. Keep the tiny cursor (it's rooted); and skip bit-sweeping for the
span set a mutator can currently be mid-claim in — for us that's exactly the
spans already generation-stamped as "acquired/cached this cycle" (mcache-
resident + freshly acquired + the tiny-cursor owner). Their garbage reclaims
one cycle late, and the delay is self-limiting (unswept slots keep their
alloc bits, the span fills, leaves the cache, and the next cycle sweeps it).

Verification: our worker-churn reproducer went from 46 detector catches + 2
SIGSEGVs per 12 rounds (under deliberate CPU-load scheduling perturbation —
the bug needs a loaded machine to fire) to 0/0/0; full soundness gate green
at every thread/reactor count on macOS arm64 and Linux arm64; no measurable
throughput or RSS cost.

Moral for anyone pairing Boehm-alternative GCs with signal-based STW: the
allocator fast path is a GC-atomicity domain. Either make the collector
never mutate/invalidate per-thread allocator state (our choice), or keep
suspended-PC awareness (restartable-sequence style) so threads frozen inside
the allocator are handled specially.

eptx:
### Branch updated — the complete, fixed collector is now on `pr/mem-mgmt-poc`

Following up on the two notes above: rather than leave the branch at its
close-date state (which still contains the STW/allocator races we later
root-caused), I've updated `cx-home/v@pr/mem-mgmt-poc` so that whenever this
is picked up again (post-V2/V3), the full corrected picture is in one place.
Five commits on top of the reviewed series:

1. **`vgc: wait for an in-flight collection in the vgc_alloc_large retry path`** —
   N concurrent large allocators hitting arena exhaustion could burn all
   retries while the one winning collector was mid-sweep, then spuriously
   OOM-panic.
2. **`vgc: publish vgc_addr_map entries with release/acquire u16 atomics`** —
   the last unsynchronized lock-free reader/writer pair in the allocator
   (TSan-flagged under concurrent HTTP load).
3. **`vgc: scan suspended threads' NEON/SIMD registers + add a Pinner API`** —
   clang lowers string/map memcpy through SIMD; a pointer live only in a NEON
   register at suspend time was invisible to the root scan. Plus a
   `runtime.Pinner` analog for FFI-held pointers.
4. **`vgc: signal-based stop-the-world suspend on macOS + park/ack hardening`** —
   the async mach `thread_suspend`+`thread_get_state` capture stops a peer at
   a weaker point than a kernel-delivered signal (details in my previous
   comment). Darwin now mirrors the proven Linux signal-suspend shape
   (SIGURG + self-capture from the ucontext + ACK + park-in-handler); the
   mach path is retained behind `-DVGC_MACH_SUSPEND` for A/B. Ack waits are
   unbounded for live targets (ESRCH-only skip), parked threads block on a
   semaphore/futex instead of yield-spinning, and every park carries a
   stop-cycle generation stamp.
5. **`vgc: collector/allocator soundness — cooperative STW, allocator-vs-STW
   races, pacing`** — the bundle: cooperative-safepoint STW as default,
   spawn-root lock (a dying thread's swap-remove raced the collector's
   unlocked spawn-root scan), conservative tracing of untyped spawn-arg
   structs, the tiny-cursor & mid-claim sweep-theft allocator races (the
   root cause from my previous comment), atomic sweep bitmap write-back,
   unconditional alloc-black, additive pacing + soft heap limit
   (`VGC_PACE_MB` / `VGC_MEMLIMIT_MB`, a GOMEMLIMIT analog), `VGC_GCTRACE=1`
   per-cycle tracing, and full-arena span pooling. The `-d vgc_verify`
   mark-closure verifier + root-finder these were proven with is included.

Validation on this exact branch (macOS arm64, 12 cores):
- `v test vlib/builtin/` — 35 passed, 1 skipped, 0 failed; `v fmt -verify`
  and `v vet` clean on the changed files.
- A pure-V reproducer (waves of short-lived spawned threads cloning maps
  with small string keys, allocating main thread, `VGC_NEXT_GC_MB=1`
  cadence, 12 CPU spinners): pre-update branch tip crashes
  (`string_clone` SIGSEGV / thread-wrapper panics) in 10/10 rounds; updated
  branch 0/10 clean.
- Linux arm64 (gcc 16, container): the same reproducer built from the
  updated tree runs 8/8 rounds clean under equivalent load and cadence.

No behavioral change for any default build: everything here is inside
`-gc e` / `-gc vgc`, still opt-in.

PR #27466, opened 2026-06-15
tests: add fat closure acceptance targets

MartenH:
This is a small draft/test-target PR for the fat-closure direction discussed around #27446/#27445.

It intentionally does not carry the frame-epoch reclamation runtime. Instead it adds opt-in acceptance tests for the desired representation boundary:

- under `-d fat_closures`, pure V stored closures should be GC-visible values and should not emit `builtin.closure` trampoline creation/init code;
- explicit closure-to-raw-function-pointer coercions still have a tested boundary where a C-callable shim is expected.

Normal test runs pass because the checks are gated behind `fat_closures`. To exercise the target today:

```sh
v -d fat_closures -gc none test vlib/v/tests/fns/fat_closure_acceptance_test.v
```

That currently fails on the pure-V closure assertion, as intended: it is the executable acceptance target for a future fat-closure implementation.

### DIFF
--- a/vlib/v/tests/fns/fat_closure_acceptance_test.v
+++ b/vlib/v/tests/fns/fat_closure_acceptance_test.v
@@ -1,0 +1,113 @@
+import os
+import rand
+
+const vexe = @VEXE
+
+fn write_file(path string, lines []string) {
+	os.write_file(path, lines.join('\n')) or { panic(err) }
+}
+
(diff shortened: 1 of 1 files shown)


MartenH:
After doing more research this is quite and undertaking and consider all the changes in memory management I decided to close at this tine

PR #27529, opened 2026-06-21
wasm: add table & element sections, call_indirect, and table instructions

enghitalo:
## What & why

The hand-written wasm encoder (`vlib/wasm`) declared the `table_section` (§4) and `element_section` (§9) ids in its `Section` enum but had **no encoder** for either, **no `call_indirect`**, and **no table instructions**. This means the library cannot emit funcref/externref tables or call functions indirectly — the foundation needed for function values, callbacks/events, interface dispatch, and externref reference slots.

This PR adds those primitives. It is **purely additive**: the new sections are emitted only when populated, so existing modules compile byte-identically.

## Changes

**`module.v`**
- `Table` / `Element` / `ElementMode` structs and the `tables` / `elements` module fields.
- `assign_table(name, export, reftype, min, max) TableIndex` — declare a funcref/externref table (pass `max` as `none` for a growable table).
- `new_active_element(tableidx, offset, funcs) ElementIndex` — populate a table at instantiation.
- `new_declarative_element(funcs) ElementIndex` — declare functions so that `ref.func` to a non-exported function validates. Per the spec, `ref.func N` only validates if `N` is declared via an element segment, export, global, or table; today `ref_func` emits the index with no declaration, so it validates only by accident when the target happens to be exported.
- `new_functype` is now `pub` and returns a `TypeIndex`, so callers can intern a signature for `call_indirect`. New `TableIndex` / `TypeIndex` / `ElementIndex` aliases.

**`instructions.v`**
- `call_indirect(typeidx, tableidx)` (`0x11`).
- `table.get` / `table.set` (`0x25` / `0x26`), `table.grow` / `table.size` / `table.fill` (`0xFC 0x0F`/`0x10`/`0x11`).

**`encoding.v`**
- Emit the **table section** (after the function section) and **element section** (after the start section, supporting active table-0, active explicit-tableidx, and declarative funcref segments), and allow **exporting tables**.
- Element-segment function-index resolution goes through a new `get_local_func_idx` helper, also reused by `get_function_idx`.

## Tests

New `vlib/wasm/tests/table_test.v` (same programmatic-build + `wasm-validate` pattern as `call_test.v`):
1. A funcref table populated by an active element segment, dispatched via `call_indirect`.
2. A growable externref table exercised with `table.size`/`grow`/`set`/`get`.
3. A declarative element segment making `ref.func` to a non-exported function validate.

All `vlib/wasm` tests pass, and the unchanged native wasm backend (`vlib/v/gen/wasm`) stays green. The `call_indirect` dispatch and externref `table.grow` were additionally confirmed to **execute** correctly under wasmer (5.0.4 and 7.1.0), and the declarative-segment requirement was confirmed by observing that, without it, `wasm-validate` fails with *"function 0 is not declared in any elem sections"*.

### DIFF
--- a/vlib/wasm/encoding.v
+++ b/vlib/wasm/encoding.v
@@ -63,15 +63,17 @@ fn push_f64(mut buf []u8, v f64) {
 	buf << u8(rv >> u32(56))
 }
 
+fn (mod &Module) get_local_func_idx(name string) int {
+	ftt := mod.functions[name] or { panic('function ${name} does not exist') }
+	return ftt.idx + mod.fn_imports.len
+}
+
 fn (mod &Module) get_function_idx(patch CallPatch) int {
 	mut idx := -1
 
 	match patch {
 		FunctionCallPatch {
-			ftt := mod.functions[patch.name] or {
-				panic('called function ${patch.name} does not exist')
-			}
-			idx = ftt.idx + mod.fn_imports.len
+			idx = mod.get_local_func_idx(patch.name)
 		}
 		ImportCallPatch {
 			for fnidx, c in mod.fn_imports {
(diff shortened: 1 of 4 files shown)


enghitalo:
Consolidated into #27532, which now contains both the encoder primitives and the codegen that uses them in a single PR.

PR #27532, opened 2026-06-21
wasm: add table & element sections, call_indirect, and function values

enghitalo:
## What & why

Adds funcref/externref **tables** and **indirect calls** to V's hand-written wasm encoder, and wires the native `wasm` codegen to use them so a function can be used as a **value** (an `i32` index into a funcref table) and called via `call_indirect`. This is the keystone for closure-free reactive code (a top-level `fn (ctx voidptr)` invoked through a stored handle), and the prerequisite for events, interface dispatch, and externref DOM slots.

Previously the `Section` enum declared `table_section`/`element_section` ids with no encoder, there was no `call_indirect` or table instruction, and both `ast.Ident`→fn and `ast.AnonFn` fell through `expr()` to an error.

## `vlib/wasm` (encoder)

- **Table section (§4)** and **element section (§9)** writers — active (table 0 and explicit-tableidx) and **declarative** funcref segments. A declarative segment makes `ref.func` to a non-exported function validate (per spec, `ref.func N` only validates if `N` is declared).
- Builders `assign_table`, `new_active_element`, `new_declarative_element`; table export support; `new_functype` is now `pub` so callers can intern a type for `call_indirect`.
- Instructions `call_indirect` (`0x11`) and `table.get`/`set` (`0x25`/`0x26`), `table.grow`/`size`/`fill` (`0xFC 0x0F`/`0x10`/`0x11`).
- Purely additive: sections are emitted only when populated, so existing modules are byte-identical.

## `vlib/v/gen/wasm` (codegen)

- A `fn` type lowers to `i32` (a table index) in `get_wasm_type` / `is_pure_type`.
- Referencing a top-level fn by name, or a non-capturing `ast.AnonFn`, pushes its table index; a single `fn name → slot` map is the sole owner of the indirect-call table (reused by later phases). Non-capturing anon fns are compiled after the toplevel statements, when per-function state is clean.
- Calling a function value emits `call_indirect`: fn-typed locals/params (`f(x)`), fn-typed **struct fields** which the checker models as method calls (`b.op(x)`, detected via `find_field` + `FnType`), and `(expr)(x)`. The indirect-call type index is built from the value's fn signature with the same lowering as `fn_decl`.
- `housekeeping` declares `__indirect_function_table` and emits one active element segment populating it.
- **Capturing closures** (`fn [x] () { ... }`) are rejected with a clear, source-located error — they cannot be lowered to a bare funcref on this backend yet. Multi-return / option-result function-value signatures are likewise rejected for now.

## Tests

- `vlib/wasm/tests/table_test.v`: hand-built funcref table dispatched via `call_indirect`, a growable externref table (`table.size`/`grow`/`set`/`get`), and a declarative element segment — each validated with `wasm-validate`.
- `vlib/v/gen/wasm/tests/fn_value.vv` (+ `.out`): runs every codegen path (local fn value, fn-typed struct field, fn-typed parameter, anonymous fn, reassignment/dynamic dispatch) and checks concrete output.
- `vlib/v/gen/wasm/tests_decompile/fn_value.vv` (+ `.wasm.must_have`): asserts `call_indirect` and a funcref table are emitted.

`v self` builds; all `vlib/wasm` and `vlib/v/gen/wasm` suites pass; runtime output verified under wasmer 7.1.0; `call_indirect` dispatch and externref `table.grow` confirmed to execute; a capturing closure confirmed to fail compilation with a clear located error.


medvednikov:
@codex review

PR #27533, opened 2026-06-21
wasm: monomorphize generic functions

enghitalo:
## What & why

Generic functions previously **ICEd** on the native wasm backend: `fn_decl` ignored `table.fn_generic_types`, so a generic type parameter `T` reached `get_wasm_type` and errored ("unreachable type"). This adds **monomorphization of free generic functions**, mirroring the C backend's pattern.

## Changes (`vlib/v/gen/wasm`)

- **`fn_decl`**: when it sees a generic template (`generic_names.len > 0`), it emits **one specialized function per concrete instantiation** the checker recorded in `table.fn_generic_types[fkey]`, re-entering `fn_decl` with a specialization context set. Each monomorph is committed under a mangled name (`generic_fn_name`, e.g. `max_T_int`).
- **`get_wasm_type` / `is_pure_type`**: resolve a generic type through the current context (`unwrap_generic` → `table.convert_generic_type`). An unresolved generic now yields a clear error instead of the opaque "unreachable type".
- **`call_expr`**: mangles the callee to the matching monomorph, and sets the specialization context (from the callee's `generic_names` and the call's `concrete_types`) so the call's still-generic `expected_arg_types` resolve while arguments are lowered.
- New `Gen` context fields `cur_concrete_types` / `cur_fn_generic_names` (+ a dedup set), restored after each monomorph / call.

## Scope

Free generic functions with scalar/pointer type parameters (`max[int]`, `max[i64]`, `identity[T]`, …). **Out of scope:** generic **methods on generic receivers** (e.g. `Signal[T].set`), where the checker leaves the type parameters on the receiver rather than the method — handling them needs the receiver-generic merging the C backend does (`effective_fn_generic_names` + receiver instantiations). Confirmed by a spike (`a.set(5)` on `Signal[int]` currently reports the monomorph as missing); that is the immediate follow-up. Generic structs are already concrete type-syms at codegen, so they need no work here.

## Tests

`tests/generic.vv` (+ `.out`), run via wasmer: `max[int]`, `max[i64]`, `identity[int]`, `identity[i64]` — asserts each monomorph returns the right value, proving distinct specializations are emitted and dispatched. `v self` builds; all `vlib/v/gen/wasm` suites pass; verified at runtime under wasmer 7.1.0. (A decompile fixture was considered but dropped: non-exported monomorphs are anonymized by `wasm-decompile` unless `-g` is passed, so the runnable test is the more robust proof.)


medvednikov:
fyi old wasm backend will be removed soon and replaced by v3/wasm

medvednikov:
monomorphization is done for all backends in v3 once

Issue #26928, opened 2026-04-21
[markused]Segmentation fault


Jengro777:
### Describe the bug

https://github.com/vlang/v/discussions/26922

Code compilation section error
Reproducing code is the smallest stable set of errors that I can achieve in reproducing segments

### Reproduction Steps

[segv_min.zip](https://github.com/user-attachments/files/26916100/segv_min.zip)

### Expected Behavior

No paragraph errors

### Current Behavior

Segmentation fault

### V version

V 0.5.1 16c1df2

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.1 aba5907d66296985d6b5a3616c9e2f2b1026c5d8.16c1df2
|:-------------------|:-------------------
|OS                  |linux, Deepin 25
|Processor           |6 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
|Memory              |1.26GB/7.6GB
|                    |
|V executable        |/home/Jengro/opt/v/v
|V last modified time|2026-04-21 01:37:55
|                    |
|V home dir          |OK, value: /home/Jengro/opt/v
|VMODULES            |OK, value: /home/Jengro/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/Jengro/Documents/Dev/RuoQi-DoDo/segv_min
|                    |
|Git version         |git version 2.51.0
|V git status        |weekly.2026.06-1255-g16c1df25
|.git/config present |true
|                    |
|cc version          |cc (Deepin 12.3.0-17deepin15) 12.3.0
|gcc version         |gcc (Deepin 12.3.0-17deepin15) 12.3.0
|clang version       |Deepin clang version 17.0.6 (5deepin7)
|tcc version         |tcc version 0.9.28rc 2025-02-13 HEAD@f8bd136d (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 696c1d84
|emcc version        |N/A
|glibc version       |ldd (Debian GLIBC 2.38-6deepin19) 2.38
Jengro@Jengro-PC:~/Documents/Dev/RuoQi-DoDo/segv_min$ 

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Jengro777:
[...]

PR #27537, opened 2026-06-22
tests: harden skip-unused json2 test against helper refactors

enghitalo:
## Background

On #27535, the `chatgpt-codex-connector` review [flagged](https://github.com/vlang/v/pull/27535#discussion_r3449280531) that `vlib/v/tests/skip_unused/generic_fn_instantiation_pruning_test.v` would start failing. This PR fixes the underlying brittleness so the test no longer hard-codes a json2 implementation detail.

## The bug

`test_skip_unused_keeps_json2_embedded_struct_decode_helpers` (regression for #26928) asserts that `-skip-unused` keeps generic json2 helpers reachable only through comptime `$for field` codegen. One assertion hard-coded the generated symbol of a specific helper:

```v
assert res.output.contains('x__json2__create_value_from_optional_T_time__Time')
```

`create_value_from_optional` is a json2 implementation detail. #27535 removes it (the option payload is now built inline with `$zero(field.typ.payload_type)`), so the symbol disappears and this assertion fails.

### Reproduction

Compile the test program to C and grep the symbols (this is exactly what the test does):

```v
// issue_26928.v
module main
import time
import x.json2
struct Meta { created_at ?time.Time }
struct Req  { Meta name string }
fn main() { _ := json2.decode[Req]('{"name":"x"}') or { panic(err) } }
```

```
v -w -o - issue_26928.v | grep create_value_from_optional
```

- on `master`: `x__json2__create_value_from_optional_T_time__Time` is present → test passes.
- on the #27535 branch: the symbol is gone → `test_skip_unused_keeps_json2_embedded_struct_decode_helpers` fails at that assertion.

## The fix

Assert the symbol that actually decodes the `?time.Time` payload of the embedded `Meta` — `x__json2__Decoder_decode_value_T_time__Time` — instead of a specific helper name. In this program `time.Time` only appears as that optional embedded field, so the instantiation is reachable **only** through comptime `$for field` codegen — exactly the skip-unused guarantee #26928 is about. Crucially, it is emitted **whether or not** the json2 helpers exist, so the test no longer breaks when they are refactored.

## Testing

`v -w test vlib/v/tests/skip_unused/generic_fn_instantiation_pruning_test.v` → **passes** both:
- on `master` (json2 unchanged), and
- against the #27535 branch (helper removed).

This is independent of #27535 and green on `master` today; once it lands, #27535 just needs a rebase to pick it up.

### DIFF
--- a/vlib/v/tests/skip_unused/generic_fn_instantiation_pruning_test.v
+++ b/vlib/v/tests/skip_unused/generic_fn_instantiation_pruning_test.v
@@ -155,7 +155,11 @@ fn test_skip_unused_keeps_json2_embedded_struct_decode_helpers() {
 	}
 	assert res.output.contains('x__json2__decode_struct_key_T_main__Req')
 	assert res.output.contains('x__json2__check_required_struct_fields_T_main__Req')
-	assert res.output.contains('x__json2__create_value_from_optional_T_time__Time')
+	// the `?time.Time` payload of the embedded `Meta` is decoded through a generic
+	// instantiation reachable only via comptime `$for field` codegen; skip-unused
+	// must keep it. Assert the payload decoder itself rather than a specific json2
+	// helper name, so the test does not break when those helpers are refactored.
+	assert res.output.contains('x__json2__Decoder_decode_value_T_time__Time')
 }
 
 fn test_skip_unused_marks_dependencies_inside_generic_anon_fns() {


enghitalo:
Folding this into #27535 instead — the test fix now travels with the change that removes the helper. Closing in favor of that single PR.

PR #27539, opened 2026-06-22
v3: add test_all self-host coverage

medvednikov:
## Summary
- add `vlib/v3/test_all.vsh` with V3 unit tests, C backend hello, C self-host convergence, language fixture coverage, and an ARM64 one-generation self-host smoke on arm64 hosts
- fix V3 SSA call resolution for dotted selector calls in the current module, which unblocks emitting an ARM64 V3 compiler
- keep interface dispatch methods and related runtime helpers reachable, and gate unsupported generic diagnostics behind `-selfhost`
- update V3 fixtures/tests for the current V3 backend behavior

## Notes
- The ARM64 coverage added here is a one-generation self-host smoke: C-built V3 emits an ARM64 V3 compiler, and that generated compiler emits/runs ARM64 hello.
- A deeper ARM64 `v4 -> v5` self-host chain was tested manually and still exits after `annotate types`, so this PR does not wire that failing chain into `test_all.vsh`.

## Tests
- `./vnew fmt -w vlib/v3/test_all.vsh vlib/v3/ssa/builder.v vlib/v3/tests/markused_test.v`
- `./vnew -silent test vlib/v3/ssa/`
- `./vnew -silent vlib/v3/tests/markused_test.v`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

### DIFF
--- a/vlib/v3/flat/flat.v
+++ b/vlib/v3/flat/flat.v
@@ -136,9 +136,10 @@ pub enum Op {
 
 pub struct Node {
 pub mut:
-	value   string
-	typ     string
-	kind_id int
+	value          string
+	typ            string
+	generic_params []string
+	kind_id        int
 pub:
 	pos            token.Pos
 	children_start i32
--- a/vlib/v3/gen/c/fn.v
+++ b/vlib/v3/gen/c/fn.v
@@ -46,7 +46,14 @@ fn (mut g FlatGen) should_emit_fn_node_in_module(node flat.Node, node_index int,
 	if cfn in ['Array_string__join', 'array_string_join'] {
 		return false
 	}
+	if g.has_used_fn_filter() && is_generated_fn_after_markused(node.value) {
+		return !g.has_generic_params(node)
+	}
 	if module_name == 'main' {
+		if g.has_used_fn_filter() && !g.used_fn_contains(node.value) && !g.used_fn_contains(dfn)
+			&& !g.used_fn_contains(cfn) && !g.used_fn_contains(qfn) {
+			return false
+		}
 		return !g.has_generic_params(node)
 	}
 	if g.has_used_fn_filter() && !g.used_fn_contains(node.value) && !g.used_fn_contains(dfn)
@@ -57,6 +64,10 @@ fn (mut g FlatGen) should_emit_fn_node_in_module(node flat.Node, node_index int,
 		return false
 	}
 	return true
+}
+
+fn is_generated_fn_after_markused(name string) bool {
+	return name.starts_with('__anon_fn_')
 }
 
 fn (g &FlatGen) used_fn_contains(name string) bool {
--- a/vlib/v3/gen/c/interface.v
+++ b/vlib/v3/gen/c/interface.v
@@ -262,12 +262,44 @@ fn (mut g FlatGen) interface_method_stubs() {
 	for iface_name, methods in g.interfaces {
 		cn := c_name(iface_name)
 		for method in methods {
+			if !g.should_emit_interface_dispatch(iface_name, method) {
+				continue
+			}
 			g.gen_interface_dispatch(iface_name, cn, method)
 		}
 	}
 	if g.interfaces.len > 0 {
 		g.writeln('')
 	}
+}
+
+fn (g &FlatGen) should_emit_interface_dispatch(iface_name string, method string) bool {
+	if !g.has_used_fn_filter() {
+		return true
+	}
+	name := '${iface_name}.${method}'
+	if g.used_fn_contains(name) || g.used_fn_contains(c_name(name)) {
+		return true
+	}
+	short_name := '${iface_name.all_after_last('.')}.${method}'
+	return short_name != name && g.interface_dispatch_short_name_is_unambiguous(short_name, method)
+		&& (g.used_fn_contains(short_name) || g.used_fn_contains(c_name(short_name)))
+}
+
+fn (g &FlatGen) interface_dispatch_short_name_is_unambiguous(short_name string, method string) bool {
+	mut matches := 0
+	for iface_name, methods in g.interfaces {
+		if method !in methods {
+			continue
+		}
+		if '${iface_name.all_after_last('.')}.${method}' == short_name {
+			matches++
+			if matches > 1 {
+				return false
+			}
+		}
+	}
+	return matches == 1
 }
 
 fn (mut g FlatGen) gen_interface_dispatch(iface_name string, cn string, method string) {
--- a/vlib/v3/markused/markused.v
+++ b/vlib/v3/markused/markused.v
@@ -414,7 +414,7 @@ fn enqueue_auto_roots(fn_decls map[string]FnDeclInfo, mut used map[string]bool, 
 
 fn enqueue_main_module_roots(fn_decls map[string]FnDeclInfo, mut used map[string]bool, mut queue []string) {
 	for name, info in fn_decls {
-		if info.module != 'main' || name.contains('.') {
+		if info.module != 'main' || name != 'main' {
 			continue
 		}
 		enqueue(name, mut used, mut queue)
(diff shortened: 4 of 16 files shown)


medvednikov:
Addressed both review comments in `b7f9ac211`.

- Moved the generic-placeholder fallback until after real type resolution, with a `struct A` argument-check regression.
- Removed eager rooting of every interface implementer method. Markused now keeps implementers through reachable interface dispatch, including selector calls on module globals, and C generation skips unused interface dispatch stubs under the used-function filter.

Retested:
- `./vnew -silent vlib/v3/tests/markused_test.v`
- `./vnew -silent vlib/v3/tests/type_checker_errors_test.v`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

medvednikov:
Addressed the latest review in `5e7442e6c`.

- Step 5 now compares the generated compiler C outputs (`v5_chain.c` vs `v6_chain.c`) byte-for-byte. The linked Mach-O binaries have equal sizes but different bytes due linker output, while the generated C is deterministic and directly catches codegen drift.
- Step 6 now compares the fixture stdout exactly against checked-in `vlib/v3/tests/test_all_lang_features.out`. `run_output` captures stdout only so the fixture's intentional `eprint/eprintln` coverage does not pollute the expected stdout stream.

Retested:
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

medvednikov:
Addressed the smoke-script review in `db843c30d`.

- `vlib/v3/tests/test_v3.vsh` now captures fixture stdout separately and compares it exactly against `vlib/v3/tests/test_all_lang_features.out` before reporting `v3 C: OK`.
- Stderr remains visible during the binary run, matching the `test_all.vsh` handling for the fixture's intentional stderr coverage.

Retested:
- `./vnew vlib/v3/tests/test_v3.vsh`
- `git diff --check`

medvednikov:
Addressed the latest review in `c14d682c8`.

- `-selfhost` now rejects marker-only generic struct declarations, with a regression for `struct Phantom[T] { value int }`. Normal union struct declaration markers remain exempt from generic type-string checking unless they include `generic`.
- The self-host compiler builds in `vlib/v3/test_all.vsh` now pass `-selfhost`; I also enabled it for the ARM64 self-host compiler build in step 4.

Retested:
- `./vnew -silent vlib/v3/tests/generics_test.v`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

medvednikov:
Addressed the interface dispatch review in `ab92089c5`.

- `should_emit_interface_dispatch` now still honors exact fully qualified interface method names first, but only accepts a short-name hit when that short interface method is unambiguous across the generator's interface table.
- Added a regression with reachable `Reader.read` in `main` plus an unused imported `moda.Reader.read` with a different signature; the generated C keeps `Reader__read` and does not emit `moda__Reader__read`.

Retested:
- `./vnew -silent vlib/v3/tests/markused_test.v`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

medvednikov:
Addressed the imported-source self-host generic review in `fc2f06cd3`.

- Self-host builds now keep ordinary diagnostics limited to the initial input files, but add a generic-only diagnostic whitelist for parsed files under the input project root. For `vlib/v3/v3.v`, that covers imported compiler modules under `vlib/v3/**` without surfacing unrelated stdlib diagnostics that V3 still intentionally suppresses.
- Unsupported generic diagnostics now use a dedicated `record_unsupported_generic` path, so imported project-root files can report only the self-host generic guard.
- Added a regression where `main.v` imports `badmod`, and `badmod/badmod.v` declares `struct Phantom[T] { value int }`; `-selfhost` now reports it.

Retested:
- `./vnew -silent vlib/v3/tests/generics_test.v`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed the latest review in `12f784bda`.

- Main-module methods are now pruned under the used-function filter when markused did not keep them, matching interface dispatch stub pruning. Added a regression for an unused `main` method that calls through an interface; generated C no longer emits either the unused method or its dispatch stub.
- Bare single-letter unknown types now fall through to `unknown type` outside `-selfhost`. The parser carries declaration generic parameters on flat nodes, so real generic declarations like `fn id[T](x T) T` still pass in non-selfhost mode while `fn f(x Z) Z` reports `unknown type `Z``.

Retested:
- `./vnew -silent vlib/v3/tests/generics_test.v`
- `./vnew -silent vlib/v3/tests/type_checker_errors_test.v`
- `./vnew -silent vlib/v3/tests/markused_test.v`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

medvednikov:
Addressed the undeclared generic parameter review in `e37030474`.

- `infer_decl_generic_params` now accepts only explicit `node.generic_params` plus type parameters introduced by a real generic method receiver. It no longer scans the entire declaration and promotes every unknown single-letter type seen under generic context.
- The receiver path also checks that the first parameter type matches the dotted method owner before treating it as a receiver, so static methods do not accidentally introduce params.
- Added regressions for `fn f[T](x T, y Z) {}` and `struct Box[T] { other U }`; both now report `unknown type` for the undeclared type parameter.

Retested:
- `./vnew -silent vlib/v3/tests/generics_test.v`
- `./vnew -silent vlib/v3/tests/type_checker_errors_test.v`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

medvednikov:
Addressed the latest checker review in `fcb4368c5`.

- `-selfhost` now rejects nonempty `node.generic_params` on non-struct declarations too, so marker-only declarations like `fn unused[T]() {}`, `interface I[T] {}`, and `type Alias[T] = int` are rejected even when no child type string mentions the generic parameter. Generic structs keep their existing `unsupported generic struct` diagnostic.
- Non-selfhost generic applications now validate the base type before returning, so `fn f(x Missing[int]) {}` reports `unknown type `Missing`` instead of silently deferring failure.
- Added regressions for marker-only generic function/interface/type alias declarations and for an unknown generic application base.

Retested:
- `./vnew -silent vlib/v3/tests/generics_test.v`
- `./vnew -silent vlib/v3/tests/type_checker_errors_test.v`
- `./vnew vlib/v3/test_all.vsh` (first run hit a transient parallel unit-test link failure in an existing case; direct `type_checker_errors_test.v` passed, and the rerun of `test_all.vsh` passed end to end)
- `git diff --check`

medvednikov:
Addressed the latest checker review in `aa4faa65f`.

- Non-selfhost generic applications now recursively validate bracket arguments after checking the base type, so `Box[Missing]` reports an unknown type diagnostic for `Missing` instead of collapsing to `Box` silently.
- Selfhost semantic checks now include `.c_fn_decl` no-body declarations, so `fn helper[T]()` is rejected by the same generic declaration guard.
- Added regressions for `Box[Missing]` and a selfhost generic no-body declaration.

Retested:
- `./vnew -silent vlib/v3/tests/generics_test.v`
- `./vnew -silent vlib/v3/tests/type_checker_errors_test.v`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed the ARM64 host-gating review in `2bd94d003`.

- `vlib/v3/test_all.vsh` now records `os.user_os()` and runs the ARM64 self-host smoke only on `macos/arm64` hosts.
- Non-macOS ARM64 hosts now skip the smoke with an explicit Mach-O-only message, avoiding exec-format failures on Linux/aarch64.
- Updated the final coverage note to call out the macOS ARM64 smoke.

Retested:
- `./vnew fmt -w vlib/v3/test_all.vsh`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

medvednikov:
Addressed the latest cgen/markused review in `0c6a04838`.

- The used-function filter now applies to all main-module functions, so unreachable top-level helpers are pruned consistently with methods.
- `markused` no longer roots every top-level `module main` function, but still keeps real auto roots. Generated fn literals (`__anon_fn_*`) and the `flag_default_value` lowering helper remain emitted because they are introduced or referenced after markused.
- Updated the parallel-cgen fixture so it forces parallel generation with reachable helpers instead of relying on dead-code emission.
- Added regressions for an unreachable main helper calling a pruned method, reachable lifted fn literals, and `flag_default_value` lowering.

Retested:
- `./vnew -silent vlib/v3/tests/markused_test.v`
- `./vnew -silent vlib/v3/tests/parallel_cgen_test.v`
- `./vnew vlib/v3/test_all.vsh`
- `git diff --check`

PR #27544, opened 2026-06-23
builtin, cgen, checker: fix spawn-in-if/match-expr, zero-alloc error sentinel, cheaper slice marking, scalable GC

### LINKED ISSUE #27485: C codegen error: assigning an `if`/`match` expression whose branches are `spawn` calls
**Describe the bug**

Assigning the result of an `if` (or `match`) **expression** to a variable, where the branches are `spawn` calls — so the expression yields a `thread T` — produces invalid C and fails to compile. Assigning a `spawn` directly works fine, so the value is well-typed; only the if/match-expression form mis-generates.

**Reproduction Steps**

```v
fn f() int {
	return 1
}

fn main() {
	t := if true { spawn f() } else { spawn f() }
	println(t.wait())
}
```

`v run bug.v`

The `match`-expression form fails identically:

```v
fn f() int {
	return 1
}

fn main() {
	x := 1
	t := match x {
		1 { spawn f() }
		else { spawn f() }
	}
	println(t.wait())
}
```

**Expected Behavior**

Compiles and prints `1`. The direct form works:

```v
fn f() int { return 1 }
fn main() {
	t := spawn f()        // OK, prints 1
	println(t.wait())
}
```

**Current Behavior**

C compilation error (same with tcc and gcc):

```
cc: /tmp/v_1000/spawnmin.<hash>.tmp.c:4207: error: ')' expected (got "arg__t1")
```

The generated C for the `spawn` inside the if/match-expression branch emits a malformed call (the `arg__tN` thread-argument wrapper is referenced where the expression result is materialized).

**Possible Solution**

`spawn`/`go` expressions used as the value of an `if`/`match` expression need the same thread-result handling as the statement form `t := spawn f()`. Likely in `vlib/v/gen/c` where the spawn (go_expr) result is assigned within an expression context.

**Possible workaround**

Use the statement form:

```v
mut t := spawn f()
if cond { t = spawn g() }
```

**V version:** V 0.5.1 498584c
**Environment:** Ubuntu 24.04 LTS, Linux 6.8.0-41-generic, x86_64

> [!NOTE]
>

### LINKED ISSUE #27486: Concurrent allocation does not scale: default (Boehm) GC serializes GC_malloc across threads
**Describe the bug** (performance / scalability)

On a multi-core machine, allocation-heavy work does not scale across threads under V's default (Boehm) GC. N threads each doing a fixed number of small heap allocations take ~N× longer in **wall time** (i.e. they run serially), while identical pure-compute work scales perfectly across cores. This caps thread-per-core servers (a common V use case): request handling that allocates per request cannot use more than ~one core's worth of allocation throughput.

**Reproduction Steps**

```v
import time
import os

fn compute_work(iters int) u64 {
	mut acc := u64(1)
	for i in 0 .. iters {
		acc = acc * 6364136223846793005 + u64(i)
		acc ^= acc >> 13
	}
	return acc
}

fn alloc_work(iters int) u64 {
	mut acc := u64(0)
	for i in 0 .. iters {
		mut b := []u8{len: 256} // one small heap allocation per iteration
		b[0] = u8(i)
		acc += u64(b[0])
	}
	return acc
}

fn main() {
	mode := if os.args.len > 1 { os.args[1] } else { 'alloc' }
	nthreads := if os.args.len > 2 { os.args[2].int() } else { 1 }
	per := 3_000_000
	sw := time.new_stopwatch()
	mut threads := []thread u64{}
	for _ in 0 .. nthreads {
		if mode == 'compute' {
			threads << spawn compute_work(per)
		} else {
			threads << spawn alloc_work(per)
		}
	}
	mut total := u64(0)
	for t in threads {
		total += t.wait()
	}
	println('${mode} threads=${nthreads} wall=${sw.elapsed().milliseconds()}ms')
}
```

```
v -prod -o gcs bug.v
./gcs compute 1 ; ./gcs compute 16    # control: pure compute, no allocation
./gcs alloc   1 ; ./gcs alloc   16    # one small heap alloc per iteration
```

**Current Behavior** (16-core AMD Ryzen 7 5800H, Ubuntu 24.04)

```
compute threads=1   wall=4ms
compute threads=16  wall=5ms       # 16x the work, ~same wall time → scales linearly
alloc   threads=1   wall=260ms
alloc   threads=16  wall=4226ms    # 16x the work, 16.3x wall time → fully serialized
```

Each thread does identical, independent work. The only difference between the two modes is the per-iteration `[]u8{len:256}` allocation. Compute parallelizes across all cores; allocation does not.

`strace -fc ./gcs alloc 16` shows the time is in stop-the-world coordination, not the allocation itself:

```
% time     calls    syscall
 45.21    192192    clock_nanosleep
 45.00    220599    rt_sigsuspend
  4.05     31467    futex
  2.49    912906    rt_sigprocmask
  ...      441198    rt_sigreturn / getpid / tgkill
```

i.e. concurrent `GC_malloc` serializes on Boehm's global allocation lock / stop-the-world.

**Expected Behavior**

Concurrent small-object allocation should scale with cores (much closer to the compute control), so multi-threaded / thread-per-core programs can allocate in parallel.

**Possible Solution**

The bundled Boehm GC appears not to use thread-local allocation. Building libgc with `-DTHREAD_LOCAL_ALLOC` (Boehm's per-thread free lists, used by `GC_malloc` for small objects) lets concurrent small allocations proceed without taking the global lock on the fast path, and should restore most of the scaling. If that is impractical, documenting the limitation plus the `-gc none` + arena/manual-memory path for hot multi-threaded code would help server authors.

**V version:** V 0.5.1 498584c
**Environment:** Ubuntu 24.04 LTS, Linux 6.8.0-41-generic, x86_64

> [!NOTE]
>

### LINKED ISSUE #27488: Allocation throughput does not scale across cores (default GC: 16 cores ≈ 1 core aggregate)
### Context

For thread-per-core workloads (e.g. HTTP servers where each worker is pinned to a core and allocates independently), allocation throughput should scale roughly with core count. In practice V's allocation does **not** scale — and with the default GC it gets *worse* with more cores.

This is a discussion/RFC, not a crash bug. The data below is reproducible on master.

### Benchmark

Each of N threads performs 500k `[]u8{len: 64}` allocations in a tight loop (memory bounded by count — N × 500k × 64 B). We measure aggregate allocations/sec. If allocation scaled, aggregate would grow ~linearly with N and per-thread would stay roughly flat.

```v
import time
import os

const alloc_size = 64

fn worker(count int, sink &u64) {
	mut acc := u64(0)
	for _ in 0 .. count {
		b := []u8{len: alloc_size}
		unsafe {
			b[0] = u8(acc) // touch so the alloc isn't optimised away
			acc += u64(b.data)
		}
	}
	unsafe { *sink = acc }
}

fn main() {
	nthreads := os.getenv_opt('NT') or { '1' }.int()
	count := os.getenv_opt('COUNT') or { '500000' }.int()
	mut sinks := []u64{len: nthreads}
	sw := time.new_stopwatch()
	mut threads := []thread{}
	for i in 0 .. nthreads {
		threads << spawn worker(count, unsafe { &sinks[i] })
	}
	threads.wait()
	secs := f64(sw.elapsed().nanoseconds()) / 1e9
	total := i64(nthreads) * i64(count)
	println('NT=${nthreads} aggregate=${f64(total) / secs:.0f}/s per_thread=${f64(total) / secs / f64(nthreads):.0f}/s time=${secs:.3f}s')
}
```

Run (pin to N cores):
```sh
v -prod -gc none -o as_none bench.v   # libc malloc
v -prod        -o as_boehm bench.v    # default GC (Boehm)
for nt in 1 2 4 8 16; do NT=$nt taskset -c 0-$((nt-1)) ./as_none;  done
for nt in 1 2 4 8 16; do NT=$nt taskset -c 0-$((nt-1)) ./as_boehm; done
```

### Results (AMD Ryzen 7 5800H, 8 cores / 16 threads; 3× re-runs, stable)

| threads | `-gc none` aggregate/s | per-thread | default GC aggregate/s | per-thread |
|--:|--:|--:|--:|--:|
| 1  | 13.0 M | 13.0 M | 18.5 M | 18.5 M |
| 2  | 20.4 M | 10.2 M | 21.7 M | 10.8 M |
| 4  | 33.3 M |  8.3 M | 18.1 M |  4.5 M |
| 8  | 33.1 M |  4.1 M | 16.6 M |  2.1 M |
| 16 | 32.0 M |  2.0 M | 18.0 M |  1.15 M |

### Observations

- **Default GC (Boehm):** aggregate throughput is essentially **flat at ~18 M/s regardless of core count** — 16 cores deliver the same *total* allocation throughput as a single core, and per-thread throughput collapses ~16×. Allocation is effectively serialized.
- **`-gc none` (libc malloc):** scales ~2.5× from 1→4 threads, then **plateaus** past 4 cores (well below the 8 physical cores) — glibc per-arena lock / arena contention.

### Discussion / questions

1. Is this scaling profile expected for the default GC?
2. Would thread-local allocation caches (or a per-thread arena in front of the GC) be in scope?
3. Is an optional scalable allocator backend (e.g. mimalloc / jemalloc) selectable via `-gc` something the project would consider?
4. Failing the above, would docs guidance for thread-per-core designs ("minimise per-request allocation; the allocator does not scale across cores") be welcome?

For the server this came from, the practical answer was to drive the hot path to zero allocations — but the underlying allocator scalability seemed worth surfacing.

### V version
V 0.5.1 98bbdd7 (master)
### OS
Linux x86_64 (AMD Ryzen 7 5800H, 8C/16T)

> [!NOTE]
>

### LINKED ISSUE #27507: Performance: array.slice() marks the source buffer (mark_buffer_has_slices) on every slice — ~13% of a hot loop for transient read-only views
## Describe the bug

`array.slice()` (i.e. every `a[start..end]`) unconditionally calls `mark_buffer_has_slices()` on the **source** array — which computes the array's malloc data-header pointer (`data_header()`) and writes `header.has_slices = true` — plus a `uses_noscan_data()` flag check and the result-struct build, on **every** slice.

For the extremely common pattern of a **transient, read-only sub-view** (parsers, byte scanners, splitting a buffer), this bookkeeping is pure overhead: nothing ever consults `has_slices`, because the parent is not `delete()`d/shrunk while the view is alive. In a hot loop it dominates.

## Reproduction

```v
module main

@[direct_array_access]
fn sum_via_slice(buf []u8, iters int) u64 {
	mut acc := u64(0)
	for _ in 0 .. iters {
		s := buf[8..buf.len] // read-only window, dropped immediately
		for b in s {
			acc += b
		}
	}
	return acc
}

fn main() {
	mut buf := []u8{len: 0, cap: 1024}
	for i in 0 .. 256 {
		buf << u8(i)
	}
	println('acc=${sum_via_slice(buf, 2_000_000)}')
}
```

```
v -prod -g -gc none -cc gcc -o slice_repro slice_repro.v
valgrind --tool=callgrind ./slice_repro && callgrind_annotate callgrind.out.*
```

**Result (V 0.5.1 c0624b2, Linux x86-64, gcc):**

```
830,199,252 (100.0%)  PROGRAM TOTALS
108,000,000 (13.01%)  vlib/builtin/array.v:builtin__array_slice   (2,000,000x → 54 instr/slice)
```

`array_slice` alone is **13%** of all instructions — for a slice that is only read and then discarded. In a real HTTP/1.1 server hot path (re-slicing the read buffer twice per request) it measured **11% of `-prod` instructions**; replacing those two slices with a hand-rolled non-marking view cut total instructions/request by ~10%.

## What the codegen shows

`-prod` emits (abridged):

```c
VV_LOC array builtin__array_slice(array a, int start, int _end) {
	... bounds checks (still present in -prod) ...
	builtin__array_mark_buffer_has_slices(&a);     // <-- unconditional, every call
	... uses_noscan_data check ...
	array res = (array){ .data=..., .len=l, .cap=l, .flags=ArrayFlags__is_slice, ... };
	return res;
}

inline VV_LOC void builtin__array_mark_buffer_has_slices(array* a) {
	if (!ArrayFlags_has(&a->flags, managed) || a->data == 0) return;
	ArrayDataHeader* header = builtin__array_data_header(*a);   // header pointer round-trip
	if (header) header->has_slices = true;
}
```

## Why this is avoidable

`has_slices` exists so `delete()`/shrink can preserve correctness when a live slice still references the buffer. But the overwhelmingly common case — a slice that is read and dropped before any mutation of the parent — never benefits from the flag. Marking the parent on every transient view is wasted work, and it is especially pointless under `-gc none` / `-manualfree`, where the buffer is managed by hand and never `delete()`d with a live slice.

## Proposals (any one would help)

1. **A zero-cost read-only view primitive**, e.g. `a.window(start, len)` or an explicit `unsafe`/non-marking slice form, returning a `[]T` that shares the backing but does **not** mark the parent. (Today this requires hand-fabricating the `array` struct in `unsafe`.)
2. **Skip `mark_buffer_has_slices()`** when the source is not autofree-managed (e.g. `-gc none`/`-manualfree`), where `has_slices` is dead.
3. **Make the marking cheaper** — avoid the `data_header()` pointer round-trip per slice (e.g. fold `has_slices` into the array header that's already in registers, or set it lazily only when a `delete`/shrink actually runs and a slice exists).

## Environment

- V 0.5.1 c0624b2, Linux x86-64, gcc, `-prod -gc none`.

Happy to provide the full HTTP-server profile or test any patch.

> [!NOTE]
>

### LINKED ISSUE #27508: error() heap-allocates a MessageError on every call; provide a zero-alloc sentinel error like `none` for hot not-found paths
### Describe the feature

`error('msg')` heap-allocates a `MessageError` on **every** call and boxes it into
the `IError` interface. This happens on the hot path of any `!T` ("result") function
that returns an error, **even when the caller discards the result with `or {}`** —
the allocation happens inside the callee, before it returns, regardless of what the
caller does with the value.

Current implementation, `vlib/builtin/chan_option_result.v:128-134`:

```v
@[inline]
pub fn error(message string) IError {
	trace_error(message)
	return &MessageError{
		msg: message
	}
}
```

The `&` heap-allocates a `MessageError` (GC-managed). The compiler lowers `error(...)`
to `builtin___v_error(...)`. Generated C (`v -printfn builtin___v_error file.v`):

```c
inline IError builtin___v_error(string message) {
	;
	GC_reachable_here(&message);
	return I_MessageError_to_Interface_IError((HEAP(MessageError, ((MessageError){.msg = message,.code = 0,}))));
}
```

`HEAP(type, expr)` expands to `((type*)builtin__memdup((void*)&((type[]){expr}[0]), sizeof(type)))`
— a real allocation (`builtin__memdup`) plus interface boxing, **per call**.

This contrasts with `none`, which already has a zero-allocation cached sentinel
(`vlib/builtin/chan_option_result.v:203-211`):

```v
const none__ = IError(&None__{})

struct None__ {
	Error
}
```

### Use Case

A very common pattern is a "lookup / not-found" function on a hot path:

```v
fn find(x int) !int {
	if x < 0 {
		return error('not found')
	}
	return x
}

fn main() {
	mut n := 0
	for i := -1_000_000; i < 0; i++ {
		find(i) or { n++ } // result discarded, yet each error() above still allocated
	}
	println(n)
}
```

Every miss allocates a `MessageError` on the heap, even though the error object is
immediately discarded. For tight loops, caches, parsers, and any "fast not-found path",
this is a per-call allocation that cannot currently be avoided while still using the
idiomatic `!T` / `error()` mechanism.

Generated `main__find` (identical under default **and** `-prod`):

```c
VV_LOC _result_int main__find(int x) {
	if (x < 0) {
		return (_result_int){ .is_error=true, .err=builtin___v_error(_S("not found")), .data={E_STRUCT} };
	}
	...
}
```

The callee always builds the boxed error (`builtin___v_error` → `HEAP(...)` →
`builtin__memdup`) before returning, so `or {}` at the call site cannot avoid it.

V already proves a zero-alloc error return is possible: `?T` functions that
`return none` compile to a cached const, no allocation:

```c
return (_option_int){ .state=2, .err=_const_none__, .data={E_STRUCT} };
```

There is no equivalent for the very common "return an error with no extra state" case.

### Proposed Solution

Provide a non-allocating, cached/sentinel error path for the common stateless-error
case, mirroring how `none` works. Possible approaches (any one would address this):

1. **A cached/static sentinel error** for the stateless case (e.g. a public
   `error_none` constant, or reusing a single static `IError`), so a "not found" fast
   path can be expressed allocation-free.
2. **A non-allocating `error()` for compile-time-constant messages.** String literals
   are already static, so the `MessageError` could be emitted as a static instance
   instead of `HEAP(...)` when the message is a constant.
3. **A compiler optimization**: when the result of `error(...)` is provably discarded
   by the caller's `or {}` (message unused), skip the `HEAP` allocation and return a
   cached sentinel.

The `none` machinery in `vlib/builtin/chan_option_result.v:203-211` is the reference
implementation pattern.

Exact code involved: `vlib/builtin/chan_option_result.v:129-134` (`error`), its lowering
to `builtin___v_error`, and the matching `HEAP(MessageError, ...)` expansion above.

### Other Information

Reproduction verified with `v run`, `v -printfn ...`, and `v -prod` on the version below.
Swapping `!int` → `?int` and `return error('not found')` → `return none` removes the
allocation (`.err=_const_none__`), demonstrating both the asymmetry and that a zero-alloc
path is feasible.

No existing duplicate was found via `gh search issues --repo vlang/v` for terms like
"error allocation", "MessageError", "zero allocation error", "cached error",
"IError boxing".

### Acknowledgements

### Version used

V 0.5.1 6dd9033de29819f9ada4011f783b1d8b9d47d7e8.eb1d47b

linux, Ubuntu 24.04 LTS; cc (GCC) 14.2.0

> [!NOTE]
>

medvednikov:
Fixes five reported issues in a single PR.

Fixes #27485, fixes #27508, fixes #27507, fixes #27486, fixes #27488.

### #27485 — C codegen error: `if`/`match` expression whose branches are `spawn` calls
Assigning the result of an `if`/`match` *expression* whose branches are `spawn` calls produced invalid C: the spawn's statement-level setup (arg-struct alloc, `pthread_create`, ...) was emitted inside a C ternary operand.

- `gen/c`: `need_tmp_var_in_expr` now returns `true` for `SpawnExpr`/`GoExpr`, forcing the tmp-var (non-ternary) form.
- `checker`: a trailing `spawn`/`go` in a value-producing `if`/`match` branch is marked `is_expr`, so it materializes a real, joinable thread handle (waiter, no `pthread_detach`) exactly like the assignment form `t := spawn f()`.

```v
fn f() int { return 1 }
fn main() {
	t := if true { spawn f() } else { spawn f() } // now compiles & joins
	println(t.wait())
}
```

### #27508 — `error()` heap-allocates a `MessageError` on every call
Adds `error_sentinel`, a reusable, allocation-free `IError` (a cached const, mirroring how `none` works for `?T`). `return error_sentinel` from a `!T` function avoids the per-call `MessageError` allocation that `error('...')` incurs.

```v
fn find(x int) !int {
	if x < 0 { return error_sentinel } // zero-alloc
	return x
}
```
Verified in the generated C: `return error_sentinel` emits `.err=_const_error_sentinel` (no per-call `HEAP`/`memdup`; a single init-time allocation, like `_const_none__`). Mirrored in the JS builtin (`vlib/builtin/js`) so `-b js` builds resolve it too, and exempted from `free()` like `none__` so `-autofree` cleanup does not double-free the shared singleton (incl. copies/re-exports such as `const my_err = error_sentinel`).

### #27507 — `array.slice()` marks the source buffer on every slice
Makes `mark_buffer_has_slices()` cheaper: the data-header lookup is inlined (dropping the redundant `managed`/`nil` re-check and the now-dead `header != nil` test inside `data_header()`), and the flag is stored only when not already set, so repeated transient slices of the same buffer no longer dirty its header cache line on every call. Behavior (copy-on-grow correctness for live slices) is unchanged.

### #27486 + #27488 — Concurrent allocation does not scale (Boehm GC serializes `GC_malloc`)
Builds the bundled Boehm GC with `-DTHREAD_LOCAL_ALLOC=1`, so each registered thread gets its own small-object free lists and concurrent `GC_malloc` no longer serializes on the global allocator lock. V's spawned threads are registered automatically via the `pthread_create` → `GC_pthread_create` redirect. The flag is gated on `GC_THREADS` (which TLA requires).

This closes the one path that lacked TLA: the from-source amalgamation `gc.c`, generated with `--enable-thread-local-alloc=no`. The prebuilt `thirdparty/tcc/lib/libgc.a`/`.dylib` used by the default fast path is already built with TLA (verified: it exports `GC_init_thread_local`); the `*_bdwgc.sh` scripts now pass `--enable-thread-local-alloc` explicitly so that can't silently regress. A system libgc V merely links against is unaffected. No single-threaded regression (NT=1 throughput unchanged).

### Tests
Adds regression tests:
- `vlib/v/tests/concurrency/spawn_in_if_match_expr_test.v`
- `vlib/v/tests/error_sentinel_test.v`
- `vlib/v/tests/autofree_error_sentinel_test.v` (`-autofree -gc none`; double-free guard)
- `vlib/v/tests/array_slice_buffer_mark_test.v`

Validated locally: full self-host, plus `vlib/sync` (37), `vlib/v/tests/concurrency` (68), if/match-expr (25), option/result (27) and builtin array suites all green; the `-autofree` double-free was reproduced/fixed under AddressSanitizer; `v fmt` clean.

### DIFF
--- a/thirdparty/build_scripts/thirdparty-freebsd-amd64_bdwgc.sh
+++ b/thirdparty/build_scripts/thirdparty-freebsd-amd64_bdwgc.sh
@@ -41,6 +41,7 @@ CC=$CC CFLAGS='-Os -mtune=generic -fPIC' LDFLAGS='-Os -fPIC' ./configure \
 	--enable-parallel-mark \
 	--enable-single-obj-compilation \
 	--enable-gc-debug \
+	--enable-thread-local-alloc \
 	--with-libatomic-ops=yes \
 	--enable-sigrt-signals \
 	--enable-mmap
--- a/thirdparty/build_scripts/thirdparty-linux-amd64_bdwgc.sh
+++ b/thirdparty/build_scripts/thirdparty-linux-amd64_bdwgc.sh
@@ -41,6 +41,7 @@ CC=$CC CFLAGS='-Os -mtune=generic -fPIC' LDFLAGS='-Os -fPIC' ./configure \
 	--enable-parallel-mark \
 	--enable-single-obj-compilation \
 	--enable-gc-debug \
+	--enable-thread-local-alloc \
 	--with-libatomic-ops=yes \
 	--enable-sigrt-signals
 
--- a/thirdparty/build_scripts/thirdparty-linux-armv7_bdwgc.sh
+++ b/thirdparty/build_scripts/thirdparty-linux-armv7_bdwgc.sh
@@ -44,6 +44,7 @@ CC=$CC CFLAGS='-fPIC' LDFLAGS='-fPIC' ./configure \
 	--enable-parallel-mark \
 	--enable-single-obj-compilation \
 	--enable-gc-debug \
+	--enable-thread-local-alloc \
 	--with-libatomic-ops=yes \
 	--enable-sigrt-signals
 
--- a/thirdparty/build_scripts/thirdparty-macos-arm64_bdwgc.sh
+++ b/thirdparty/build_scripts/thirdparty-macos-arm64_bdwgc.sh
@@ -42,6 +42,7 @@ CC=$CC CFLAGS="-Os -mtune=generic -fPIC" LDFLAGS="-Os -fPIC" ./configure \
 	--enable-shared=yes \
 	--enable-single-obj-compilation \
 	--enable-gc-debug \
+	--enable-thread-local-alloc \
 	--enable-large-config \
 	--enable-cplusplus \
 	--with-libatomic-ops=check \
--- a/vlib/builtin/allocation.c.v
+++ b/vlib/builtin/allocation.c.v
@@ -447,6 +447,15 @@ pub fn free(ptr voidptr) {
 	if ptr == none_err._object {
 		return
 	}
+	// `error_sentinel` is likewise a cached, non-owning singleton IError (see
+	// chan_option_result.v). Under `-autofree`, const cleanup frees every IError
+	// const's boxed object, and the sentinel can also be copied/re-exported
+	// (`const my_err = error_sentinel`), so guard its shared object here to avoid
+	// a double free / dangling singleton, exactly as for `none__` above.
+	sentinel_err := &C.IError(&error_sentinel)
+	if ptr == sentinel_err._object {
+		return
+	}
 	$if prealloc {
 		return
 	} $else $if vgc ? {
--- a/vlib/builtin/array.v
+++ b/vlib/builtin/array.v
@@ -118,9 +118,15 @@ fn (mut a array) mark_buffer_has_slices() {
 	if !a.flags.has(.managed) || a.data == unsafe { nil } {
 		return
 	}
+	// Compute the data-header pointer inline. `data_header()` would redundantly
+	// re-check `.managed`/`nil` (already guaranteed above), so skip the call and
+	// the now-dead `header != nil` test. Only store when the flag is not already
+	// set, to avoid dirtying the buffer's header cache line on every transient
+	// slice of the same buffer in a hot loop. See issue #27507.
 	unsafe {
-		header := a.data_header()
-		if header != nil {
+		base_data := &u8(a.data) - u64(a.offset)
+		header := &ArrayDataHeader(base_data - array_data_header_size())
+		if !header.has_slices {
 			header.has_slices = true
 		}
 	}
--- a/vlib/builtin/builtin_d_gcboehm.c.v
+++ b/vlib/builtin/builtin_d_gcboehm.c.v
@@ -7,9 +7,16 @@ $if !no_gc_threads ? {
 	// serializes on the global allocator lock. Without it, allocation does not scale
 	// across cores (16 cores ≈ 1 core aggregate). V's spawned threads are registered
 	// via the `pthread_create` -> `GC_pthread_create` redirect, so their free lists
-	// are set up automatically. TLA requires GC_THREADS (defined just above); it only
-	// affects the bundled gc.c build and is inert for a system libgc that V merely
-	// links against. See issues #27486 and #27488.
+	// are set up automatically. TLA requires GC_THREADS (defined just above).
+	//
+	// This flag only matters for the bundled `gc.c` that V compiles from source (the
+	// `-prod`/no-prebuilt-archive branches below): its amalgamation was generated with
+	// `--enable-thread-local-alloc=no` (see thirdparty/libgc/amalgamation.txt), so the
+	// flag turns TLA back on. The prebuilt `thirdparty/tcc/lib/libgc.a`/`.dylib` used by
+	// the default fast path is already built with TLA (bdwgc enables it by default with
+	// `--enable-threads=pthreads`; see thirdparty/build_scripts/*_bdwgc.sh), and a system
+	// libgc that V merely links against is unaffected. So both bundled GC paths end up
+	// thread-local-alloc enabled. See issues #27486 and #27488.
 	#flag -DTHREAD_LOCAL_ALLOC=1
 }
 
--- a/vlib/builtin/builtin_d_gcboehm.c.v
+++ b/vlib/builtin/builtin_d_gcboehm.c.v
@@ -2,6 +2,15 @@ module builtin
 
 $if !no_gc_threads ? {
 	#flag -DGC_THREADS=1
+	// Enable Boehm's thread-local allocation: each registered thread gets its own
+	// small-object free lists, so concurrent `GC_malloc` on the fast path no longer
+	// serializes on the global allocator lock. Without it, allocation does not scale
+	// across cores (16 cores ≈ 1 core aggregate). V's spawned threads are registered
+	// via the `pthread_create` -> `GC_pthread_create` redirect, so their free lists
+	// are set up automatically. TLA requires GC_THREADS (defined just above); it only
+	// affects the bundled gc.c build and is inert for a system libgc that V merely
+	// links against. See issues #27486 and #27488.
+	#flag -DTHREAD_LOCAL_ALLOC=1
 }
 
 $if use_bundled_libgc ? {
--- a/vlib/builtin/chan_option_result.v
+++ b/vlib/builtin/chan_option_result.v
@@ -144,6 +144,17 @@ pub fn error_with_code(message string, code int) IError {
 	}
 }
 
+// error_sentinel is a reusable, allocation-free `IError` for hot "not found"/stateless
+// error paths. Unlike `error('...')`, which heap-allocates a fresh `MessageError` on every
+// call, `error_sentinel` is a single cached const (mirroring how `none` works for `?T`),
+// so `return error_sentinel` from a `!T` function avoids the per-call allocation.
+// Use it when the caller discards the error (e.g. `x() or { ... }`) and the specific
+// message is not needed; its `.msg()` is just `'error'`.
+// Example: fn find(x int) !int { if x < 0 { return error_sentinel } return x }
+pub const error_sentinel = IError(&MessageError{
+	msg: 'error'
+})
+
 // Option is the base of V's internal option return system.
 struct Option {
 	state u8 // 0 - ok; 2 - none; 1 - ?
--- a/vlib/builtin/js/builtin.v
+++ b/vlib/builtin/js/builtin.v
@@ -149,6 +149,17 @@ pub fn error_with_code(message string, code int) IError {
 	}
 }
 
+// error_sentinel is a reusable, allocation-free `IError` for hot "not found"/stateless
+// error paths. Unlike `error('...')`, which allocates a fresh `MessageError` on every
+// call, `error_sentinel` is a single cached const (mirroring how `none` works for `?T`),
+// so `return error_sentinel` from a `!T` function avoids the per-call allocation.
+// Use it when the caller discards the error (e.g. `x() or { ... }`) and the specific
+// message is not needed; its `.msg()` is just `'error'`.
+// Example: fn find(x int) !int { if x < 0 { return error_sentinel } return x }
+pub const error_sentinel = IError(&MessageError{
+	msg: 'error'
+})
+
 // free allows for manually freeing memory allocated at the address `ptr`.
 // However, this is a no-op on JS backend
 @[unsafe]
--- a/vlib/v/checker/if.v
+++ b/vlib/v/checker/if.v
@@ -395,6 +395,14 @@ fn (mut c Checker) if_expr(mut node ast.IfExpr) ast.Type {
 			if branch.stmts.len > 0 {
 				mut stmt := branch.stmts.last()
 				if mut stmt is ast.ExprStmt {
+					// A trailing `spawn`/`go` is the value of this if-expression branch,
+					// so it must produce a usable thread handle (waiter, no detach),
+					// like the assignment form `t := spawn f()`. See issue #27485.
+					if mut stmt.expr is ast.SpawnExpr {
+						stmt.expr.is_expr = true
+					} else if mut stmt.expr is ast.GoExpr {
+						stmt.expr.is_expr = true
+					}
 					if mut stmt.expr is ast.ConcatExpr {
 						for mut val in stmt.expr.vals {
 							c.check_expr_option_or_result_call(val, c.expr(mut val))
--- a/vlib/v/checker/match.v
+++ b/vlib/v/checker/match.v
@@ -317,6 +317,14 @@ fn (mut c Checker) match_expr(mut node ast.MatchExpr) ast.Type {
 			&& (!node.is_comptime || (node.is_comptime && comptime_match_branch_result)) {
 			mut stmt := branch.stmts.last()
 			if mut stmt is ast.ExprStmt {
+				// A trailing `spawn`/`go` is the value of this match-expression branch,
+				// so it must produce a usable thread handle (waiter, no detach),
+				// like the assignment form `t := spawn f()`. See issue #27485.
+				if mut stmt.expr is ast.SpawnExpr {
+					stmt.expr.is_expr = true
+				} else if mut stmt.expr is ast.GoExpr {
+					stmt.expr.is_expr = true
+				}
 				c.expected_type = if c.expected_expr_type != ast.void_type {
 					c.expected_expr_type
 				} else {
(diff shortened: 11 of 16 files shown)


medvednikov:
Thanks for the review — both points addressed in 126d9947c.

**1. `error_sentinel` missing in the JS builtin** — Fixed. Mirrored the const in `vlib/builtin/js/builtin.v` (the `-b js` backend loads `vlib/builtin/js`). Verified a `-b js` program using `return error_sentinel` compiles and behaves identically to `error()`.

**2. TLA not applied to the prebuilt archive on the default path** — The premise turned out not to hold: the prebuilt `thirdparty/tcc/lib/libgc.a`/`.dylib` (used by the default fast path) is **already** built with thread-local allocation — bdwgc enables it by default with threads, and the `*_bdwgc.sh` scripts never disabled it:

```
$ nm thirdparty/tcc/lib/libgc.dylib | grep -c 'GC_init_thread_local\|GC_mark_thread_local_fls'
3
```

The only path that lacked TLA was the **from-source amalgamation** `gc.c`, which `thirdparty/libgc/amalgamation.txt` shows was generated with `--enable-thread-local-alloc=no` — and that is exactly what the `-DTHREAD_LOCAL_ALLOC=1` flag re-enables (confirmed: the compiled `gc.o` exports `GC_init_thread_local`). So **both** bundled GC paths are now TLA-enabled.

To make the prebuilt archive's TLA explicit (rather than relying on bdwgc's implicit default, so it can't silently regress) I also added `--enable-thread-local-alloc` to all four `thirdparty/build_scripts/*_bdwgc.sh`, and reworded the comment in `builtin_d_gcboehm.c.v` to spell out which collector each build path uses.

medvednikov:
Good catch — fixed in 3ee418e28.

`error_sentinel` is now exempted from `free()` the same way `none__` is: `free(ptr)` returns early when `ptr` matches the sentinel's boxed `_object`. Because the guard is on the shared object pointer (not the const name), it also covers copies / re-exports like `const my_err = error_sentinel` — every cleanup of that object becomes a no-op.

Verified the scenario and the fix with AddressSanitizer under `-gc none -autofree` (a program that uses `error_sentinel` and a re-exporting const, both of which the generated `_vcleanup` frees):

- **Without** the guard:
  ```
  AddressSanitizer: attempting double-free ... in builtin__IError_free
  SUMMARY: AddressSanitizer: double-free ... in builtin___v_free
  ```
- **With** the guard: clean exit, no error.

Added `vlib/v/tests/autofree_error_sentinel_test.v` (`// vtest vflags: -autofree -gc none`) exercising the sentinel and a re-exporting const so CI guards against a regression.

PR #27545, opened 2026-06-23
v3: add WebAssembly backend

medvednikov:
## Summary

Adds a `-b wasm` backend to the v3 compiler. It walks the flat AST directly (like the C backend) — WASM's structured control flow (`block`/`loop`/`if`/`br`) maps cleanly from the tree, so no SSA or relooping is needed — and emits a self-contained `.wasm` module through a minimal built-in binary encoder (LEB128 + section assembly), mirroring how the ARM64 backend ships its own `asm`/`macho`/`linker`. v3 stays self-contained.

### New files — `vlib/v3/gen/wasm/`
- **`encode.v`** — WASM binary encoder (LEB128, type/import/func/memory/export/code/data sections).
- **`code.v`** — instruction-emission helpers (one method per opcode).
- **`gen.v`** — the flat-AST tree-walking code generator.

### Scope (basic; generics ignored for now)
- Functions with numeric/bool params and locals, arithmetic, comparison, logical (short-circuit), bitwise/shift, casts, signed/unsigned correctness, `if`/`else`/`else if`, all `for` forms with `break`/`continue`, direct calls, recursion.
- `print`/`println`/`eprint`/`eprintln` of string literals, integers, and booleans via WASI `fd_write` + a built-in `itoa`.
- Emits a WASI command (`_start` → `main`) and exports every compiled function for direct testing.
- Out of scope: strings-as-values, structs, arrays, maps, generics — such functions are skipped gracefully without producing invalid wasm.

### Shared transformer fix
`stringify_expr` now falls back to `reliable_stringify_type` for compound arguments (the same pattern string interpolation already uses), so `println(a + b)` for ints is stringified for **all** backends (C/arm64/wasm), not just bare idents.

## Test plan
- New `vlib/v3/tests/wasm_codegen_test.v` validates the emitted module structure always, and — when `node` is present — runs it via `node:wasi` / the WebAssembly API to check stdout and exported-function results.
- Verified **byte-identical output** against the C backend on programs covering arithmetic, control flow, unsigned/`u64`, shifts, and bools.
- Canonical `hello.v` and `if_for.v` run correctly under WASI.
- All v3 unit tests pass (`v test vlib/v3/`).

### DIFF
--- a/vlib/v3/README.md
+++ b/vlib/v3/README.md
@@ -3,10 +3,11 @@
 Clean rewrite of the V compiler. Reuses v2's scanner, uses a flat AST parser
 with Pratt parsing, a structured type system with sum-type variants, lexical
 scoping, a transformer for AST simplification, a shared type-checking phase, a
-markused pass for dead-code elimination, recursive import resolution, and two
-backends: a direct flat-AST-to-C backend and a native ARM64 backend via SSA IR
-with a built-in linker. With `-prod`, the ARM64 backend runs SSA optimization,
-MIR lowering, and instruction selection.
+markused pass for dead-code elimination, recursive import resolution, and three
+backends: a direct flat-AST-to-C backend, a native ARM64 backend via SSA IR with
+a built-in linker, and a direct flat-AST-to-WebAssembly backend. With `-prod`,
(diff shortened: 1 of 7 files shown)


medvednikov:
@codex review

Issue #27550, opened 2026-06-24
cgen, generics: generic struct as default interface-field value in a generic wrapper isn't monomorphized (`Text_T_T`)


medvednikov:
### Description

Using a **generic struct as the default value of an interface-typed field** inside a generic wrapper struct generates broken C: the generic struct's type parameter is left unresolved (`Text_T_T` instead of `Text_T_int`), and the interface variant type-index constant for it (`_<Iface>_<Variant>_index`) is never emitted.

**V version:** V 0.5.1, current `master` (`7a89e581d`)
**OS/Backend:** reproduced on macOS arm64; originally reported on Linux/cc.

### Reproducer

`template_refactor/lib.v`:

```v
module template_refactor

pub interface Tag {
	tag() string
}
```

`main.v` (with a `v.mod` so the sub-module resolves):

```v
module main

import template_refactor

pub struct Text[T] {
	val T
}

fn (t Text[T]) tag() string {
	return 'text'
}

pub struct Wrapper[T] {
	tag template_refactor.Tag = Text[T]{}
}

fn main() {
	println(Wrapper[int]{})
}
```

Compile with:

```
v .
```

### Current behavior

```
error: unknown type name 'main__Text_T_T'; did you mean 'main__Text_T_int'?
```

The default interface-field value `Text[T]{}` in `Wrapper[T]` keeps `T` unresolved (`main__Text_T_T`) instead of monomorphizing to `Text[int]` for `Wrapper[int]`. In variants where the interface value is stringified, the related error surfaces as `_template_refactor__Tag_main__Text_index undeclared` (the interface variant type-index constant is not emitted).

### Expected behavior

The default field value should monomorphize with the wrapper's type argument (`Text[int]`), and the interface variant type-index for it should be emitted.

### Notes

- Reported via the bugs.vlang.io crash reporter (11 reports), original error: `'_template_refactor__Tag_main__Text_index' undeclared; did you mean 'template_refactor__Tag_T_main__TextParams'?`.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Issue #27549, opened 2026-06-24
cgen: option-of-pointer field (`?&T`) unwrap via `== nil` ternary emits one too many dereferences


medvednikov:
### Description

Unwrapping an **option-of-pointer struct field** (`?&T`) inside a `== nil` ternary and accessing a member emits one too many dereferences: cgen produces `(**(T**)(field.data))->member`, which dereferences down to a `T` value and then applies `->` to a non-pointer.

**V version:** V 0.5.1, current `master` (`7a89e581d`)
**OS/Backend:** reproduced on macOS arm64; originally reported on Linux/cc.

### Reproducer

```v
module main

struct SessionNode {
	id     string
	parent ?&SessionNode
}

fn node_to_json(node &SessionNode) string {
	return unsafe {
		if node.parent == nil { 'root' } else { node.parent.id }
	}
}

fn main() {
	println(node_to_json(&SessionNode{id: 'x'}))
}
```

Compile with:

```
v file.v
```

### Current behavior

```
error: member reference type 'main__SessionNode' (aka 'struct main__SessionNode') is not a pointer; did you mean to use '.'?
  string pid = ((node->parent).state == 2 ? (_S("root")) : ((**(main__SessionNode**)(node->parent.data))->id));
```

(GCC phrases it as `invalid type argument of '->' (have 'session__SessionNode')`.) Since `parent` is `?&SessionNode`, the option's `.data` holds a single `&SessionNode`, so the correct emission is `(*(main__SessionNode**)(node->parent.data))->id` — one deref, not two.

### Isolation

- The smartcast form `if p := node.parent { p.id } else { 'root' }` compiles cleanly.
- Only the `== nil` ternary form (`if node.parent == nil { ... } else { node.parent.id }`) triggers the double deref.

### Expected behavior

The `== nil` unwrap should emit the same (correct) deref count as the smartcast form.

### Notes

- Reported via the bugs.vlang.io crash reporter (7 reports). Original instance: `ai-hub engine/session/jsonl_repo.v`.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Issue #27548, opened 2026-06-24
cgen: casting a string rvalue (slice/concat) into a sumtype takes its address (`cannot take the address of an rvalue`)


medvednikov:
### Description

Casting a **string rvalue** (e.g. a slice `s[a..b]` or a concatenation) into a sumtype generates C that takes the address of the rvalue, which the C compiler rejects.

**V version:** V 0.5.1, current `master` (`7a89e581d`)
**OS/Backend:** reproduced on macOS arm64; originally reported on macOS/cc.

### Reproducer

```v
type Value = int | string

fn slice(s string) Value {
	return Value(s[1..3])
}

fn main() {
	println(slice('hello'))
}
```

Compile with:

```
v file.v
```

### Current behavior

```
error: cannot take the address of an rvalue of type 'string' (aka 'struct string')
  string_to_sumtype_main__Value(&builtin__string_substr(s, 1, 3), false)
```

The sumtype constructor takes `&<rvalue>` where the operand (`s[1..3]`, returning a `string` value) has no address.

### Expected behavior

The rvalue should be materialized into a temporary before its address is taken, so the cast compiles.

### Notes

- Reported via the bugs.vlang.io crash reporter (4 reports). Original instance: `v3/eval/eval.v`, `return Value(container[start..end])`.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #27551, opened 2026-06-24
cgen: fix sumtype string-rvalue cast, ?&T == nil unwrap, and generic default interface field

medvednikov:
Fixes three cgen bugs reported via the bugs.vlang.io crash reporter.

### #27548 — casting a string rvalue (slice/concat) into a sumtype takes its address

`Value(s[1..3])` generated `string_to_sumtype_main__Value(&builtin__string_substr(s, 1, 3), false)`, taking the address of an rvalue, which strict C compilers reject (`cannot take the address of an rvalue`). A slice `s[a..b]` is reported as an lvalue by `is_lvalue()`, so the sumtype cast emitted a bare `&` instead of materializing the value. The fix treats a slice as an rvalue in `call_cfn_for_casting_expr`, so the cast goes through the `ADDR` macro.

### #27549 — option-of-pointer field (`?&T`) unwrap via `== nil` emits one too many dereferences

`if node.parent == nil { ... } else { node.parent.id }` (with `parent ?&SessionNode`) generated `(**(main__SessionNode**)(node->parent.data))->id` — a double deref that dereferences down to a value and then applies `->`. The single-deref path in `selector_expr` was gated on `g.left_is_opt`, which is only set in the smartcast (`if p := ... {}`) form, not the `== nil`/`== none` form. The option's `.data` buffer holds a single `&T`, so one deref of the `(T**)` cast is always correct for `?&T`, independent of that flag. The `== nil` form now matches the smartcast form.

### #27550 — generic struct as default interface-field value in a generic wrapper isn't monomorphized

```v
struct Wrapper[T] {
    tag Tag = Text[T]{}
}
```

`println(Wrapper[int]{})` generated broken C referencing `main__Text_T_T` (unresolved `T`) and an undeclared interface variant index. The struct init itself already monomorphizes `Text[T]{}` -> `Text[int]{}` correctly; the breakage came from a leftover unresolved-generic variant (`Text[T]`) in the interface's `types` list being emitted by the interface auto-`str()` dispatch. `gen_str_for_interface` now skips unresolved-generic struct variants, mirroring the existing skip in `interface_table()`. Only their concrete instantiations are real runtime variants.

### Tests

Added regression tests for all three (each fails on `master` with the exact C error and passes with this change):
- `vlib/v/tests/casts/cast_string_rvalue_to_sumtype_test.v`
- `vlib/v/tests/options/option_ptr_field_nil_check_unwrap_test.v`
- `vlib/v/tests/generics/generic_struct_default_interface_field_test.v`

Verified no regressions across `vlib/v/tests/{interfaces,generics,casts,options,builtin_strings_and_interpolation}` (clang and gcc).

### DIFF
--- a/vlib/v/gen/c/auto_str_methods.v
+++ b/vlib/v/gen/c/auto_str_methods.v
@@ -439,6 +439,15 @@ fn (mut g Gen) gen_str_for_interface(info ast.Interface, styp string, typ_str st
 	fn_builder.writeln('${g.static_non_parallel}string indent_${str_fn_name}(${styp} x, ${ast.int_type_name} indent_count) { /* gen_str_for_interface */')
 	fn_builder.writeln('\tif (x._typ == 0 && x._object == NULL) return _S("nil");')
 	for typ in info.types {
+		// Skip unresolved generic struct variants (e.g. a leftover `Text[T]`
+		// registered when a generic struct is used as the default value of an
+		// interface-typed field in a generic wrapper). Only their concrete
+		// instantiations (`Text[int]`) are real runtime variants. This mirrors
+		// the same skip in interface_table().
+		type_sym := g.table.sym(typ)
+		if type_sym.info is ast.Struct && type_sym.info.is_unresolved_generic() {
+			continue
+		}
 		sub_sym := g.table.sym(ast.mktyp(typ))
 		if g.pref.skip_unused && sub_sym.idx !in g.table.used_features.used_syms {
 			continue
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -5153,13 +5153,19 @@ fn (mut g Gen) fn_ptr_cast_typ(func ast.FnType) string {
 	return g.fn_ptr_decl_str(func, ptr_name).replace_once(ptr_name, '')
 }
 
-// expr_is_range_index reports whether expr is a slice (`s[a..b]`), possibly
-// wrapped in parentheses. A slice yields a fresh rvalue with no stable address,
-// so it must be materialized via ADDR rather than `&` in a sumtype cast.
-fn expr_is_range_index(expr ast.Expr) bool {
+// expr_is_range_index_rvalue reports whether expr is rooted in a slice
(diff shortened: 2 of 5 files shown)


medvednikov:
Reviewed PR `vlang/v#27551`. I found one actionable issue.

### Finding

**[P2] Parenthesized string slices can still hit the `&<rvalue>` sumtype-cast bug.**

The new `expr_is_lvalue` logic only treats a *top-level* `ast.IndexExpr` with a `RangeExpr` as an rvalue:

```v
expr.is_lvalue() && !(expr is ast.IndexExpr && expr.index is ast.RangeExpr)
```

That fixes `Value(s[1..3])`, but `ast.Expr.is_lvalue()` explicitly recurses through `ParExpr`, so `Value((s[1..3]))` will still report true from `expr.is_lvalue()` while bypassing the new top-level `IndexExpr` check. In that case cgen can still choose the bare `&` path instead of `ADDR(...)`, reproducing the same “address of rvalue” failure for a trivially parenthesized expression.

Suggested fix: use a small recursive helper for “slice result has no stable address,” at least unwrapping `ParExpr`:

```v
fn expr_is_range_index(expr ast.Expr) bool {
	return match expr {
		ast.IndexExpr { expr.index is ast.RangeExpr }
		ast.ParExpr { expr_is_range_index(expr.expr) }
		else { false }
	}
}
```

Then use:

```v
expr.is_lvalue() && !expr_is_range_index(expr)
```

I’d also add a regression test next to the new cast test:

```v
fn paren_slice(s string) Value {
	return Value((s[1..3]))
}

fn test_string_paren_slice_cast_to_sumtype() {
	v := paren_slice('hello')
	assert v is string
	assert (v as string) == 'el'
}
```

### Other notes

The `?&T == nil` change looks consistent with the stated representation: the updated branch now uses `typ.nr_muls()` for option-of-pointer unwrapping instead of adding one extra dereference, which matches the issue’s expected `(*(T**)(data))->field` shape.

The generic-interface `str()` skip also looks reasonable and localized: it skips unresolved generic struct variants before generating interface auto-str dispatch entries.

CI had started but was still queued when I checked, so I did not treat CI as a signal either way.

PR #27565, opened 2026-06-26
checker: reject multi-return elements that disagree on option-ness (fix #27562)

### LINKED ISSUE #27562: cgen/checker: multi-return with a mismatched optional element emits incompatible `multi_return_*` C structs
### Description

When a multi-return value's element is supplied as an option (`?string(none)`) in one place but the destination slot is a plain `string`, V generates two mismatched multi-return C structs — `multi_return_string_string` (element typed `string`) and `multi_return_string_option_string` (element typed `_option_string`) — and then either writes an `_option_string` none literal into a `string`-typed slot, or assigns one struct type to the other. The checker does not catch the `(string, string)` vs `(string, ?string)` mismatch; only the C compiler does.

**V version:** V 0.5.1, current `master` (`cc39afe7d`)
**OS/Backend:** reproduced on macOS arm64; originally reported on Linux/cc.

### Reproducer

**Pattern A — `or`-default of `split_once` mixes `string` and `?string`:**

```v
fn parse(s string) ?(string, string) {
	subtype, parameter := s.split_once(';') or { s, ?string(none) }
	return subtype, parameter
}

fn main() {
	a, b := parse('foo;bar') or { return }
	println('${a} ${b}')
}
```

```
error: initializing 'u8 *' (aka 'unsigned char *') with an expression of incompatible type '_option_string'
  *(multi_return_string_string*) _t1.data = (multi_return_string_string){.arg0=s, .arg1=(_option_string){ .state=2, ... }};
```

**Pattern B — `if`-expr branches disagree on the 2nd element's optional-ness:**

```v
fn extract(s string) (string, ?string) {
	rawurl, description := if s.contains(' ') {
		s[..1], s[1..]
	} else {
		s, ?string(none)
	}
	return rawurl, description
}

fn main() {
	a, b := extract('hello world')
	println(a)
	println(b)
}
```

```
error: assigning to 'multi_return_string_string' from incompatible type 'multi_return_string_option_string'
  _t1 = (multi_return_string_option_string){.arg0=link_str, .arg1=(_option_string){ .state=2, ... }};
```

Compile with:

```
v file.v
```

### Expected behavior

V should reconcile the optional-ness of multi-return elements (coerce a plain `string` into `?string`, or report a clear V-level type error) instead of generating mismatched multi-return C structs.

### Notes

- Reported via the bugs.vlang.io crash reporter (~7 reports). Original instance: `gemini.v` (`split_once` `or`-default and an `if`-expr building `(string, ?string)`).

> [!NOTE]
>

medvednikov:
## What

Fixes #27562.

When a multi-return value's element is supplied as an option (`?string(none)`) in one place but the destination slot is a plain `string`, V silently generated two mismatched multi-return C structs — `multi_return_string_string` (element typed `string`) and `multi_return_string_option_string` (element typed `_option_string`) — and only the C compiler caught the resulting type mismatch.

## Why

`Checker.check_basic` compares two multi-return types element-by-element with a recursive `check_types` call. That call matches by type **index** and ignores the `.option`/`.result` flags, so `(string, ?string)` was treated as compatible with `(string, string)`. But each multi-return element lowers to a distinct C struct field, and `?T` is a `_option_T` struct that is **not** layout-compatible with a plain `T` — hence the bad C.

## Fix

In the multi-return branch of `check_basic`, require each element's option/result-ness to match exactly before falling through to the index-based comparison.

This turns both reported reproducers into clear V-level errors:

- **Pattern A** — `or`-block default mixing `string` and `?string`:
  ```
  error: wrong return type `(string, ?string)` in the `or {}` block, expected `(string, string)`
  ```
- **Pattern B** — `if`/`match` branches disagreeing on an element's optional-ness:
  ```
  error: mismatched types `(string, string)` and `(string, ?string)`
  ```

Legitimate element coercion in `return` (e.g. returning `(string, string)` where the function declares `(string, ?string)`) is unaffected — it goes through the dedicated `can_use_expected_multi_return_*` path that builds the correct struct, not this comparison.

## Tests

- New: `vlib/v/checker/tests/multi_return_option_element_mismatch_err.{vv,out}` covering both patterns.
- Updated: `multi_return_option_result_return_expr_err.out` — the stricter check now also flags the `!int`-in-`match` branch precisely at the offending element (in addition to the existing return-argument error).
- Full `vlib/v/compiler_errors_test.v` passes (1580 ok), and the multi-return / option / or-block runtime suites in `vlib/v/tests` pass. The compiler self-builds through the modified checker.

### DIFF
--- a/vlib/v/checker/check_types.v
+++ b/vlib/v/checker/check_types.v
@@ -577,6 +577,16 @@ fn (mut c Checker) check_basic(got ast.Type, expected ast.Type) bool {
 			return false
 		}
 		for i in 0 .. exp_types.len {
+			// Each multi-return element lowers to a distinct C struct field, so an
+			// optional/result element (`?T`/`!T`, a `_option_T`/`_result_T` struct) is
+			// not layout-compatible with a plain `T`. Comparing via `check_types` below
+			// ignores these flags (it matches by type index), which would let
+			// `(string, ?string)` masquerade as `(string, string)` and emit mismatched
+			// `multi_return_*` structs. Require the option/result-ness to match exactly.
+			if got_types[i].has_flag(.option) != exp_types[i].has_flag(.option)
+				|| got_types[i].has_flag(.result) != exp_types[i].has_flag(.result) {
+				return false
+			}
 			if !c.check_types(got_types[i], exp_types[i]) {
 				return false
 			}
--- a/vlib/v/checker/check_types.v
+++ b/vlib/v/checker/check_types.v
@@ -579,10 +579,12 @@ fn (mut c Checker) check_basic(got ast.Type, expected ast.Type) bool {
 		for i in 0 .. exp_types.len {
 			// Each multi-return element lowers to a distinct C struct field, so an
 			// optional/result element (`?T`/`!T`, a `_option_T`/`_result_T` struct) is
-			// not layout-compatible with a plain `T`. Comparing via `check_types` below
-			// ignores these flags (it matches by type index), which would let
-			// `(string, ?string)` masquerade as `(string, string)` and emit mismatched
-			// `multi_return_*` structs. Require the option/result-ness to match exactly.
+			// not layout-compatible with a plain `T`. `check_types` below matches by type
+			// index and ignores these flags, which would let `(string, ?string)`
+			// masquerade as `(string, string)` and emit mismatched `multi_return_*`
+			// structs. Keep this a strict layout check (both directions): the safe
+			// `T` -> `?T` element promotion is opted into only where codegen wraps it (the
+			// `return` and `or {}` paths, via `can_use_expected_multi_return_*`).
 			if got_types[i].has_flag(.option) != exp_types[i].has_flag(.option)
 				|| got_types[i].has_flag(.result) != exp_types[i].has_flag(.result) {
 				return false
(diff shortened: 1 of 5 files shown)


medvednikov:
Good catch — fixed in 204ced407f.

The strict element check was rejecting the safe `T` -> `?T` promotion too, breaking valid or-block defaults like `foo(-1) or { 2, 2 }` (where `foo` returns `!(int, ?int)`).

Resolution: keep `check_types` a strict multi-return layout check in **both** directions, but opt the promotion back in only where codegen actually wraps the element. `concat_expr`'s `inside_or_block`/`inside_return` paths already wrap a plain element into `_option_T`, so `check_or_last_stmt` now also accepts the or-block default via `can_use_expected_multi_return_expr_type` — which allows the widening `T` -> `?T` direction and still rejects narrowing `?T` -> `T` (the original bug).

Net effect:
- `foo(-1) or { 2, 2 }` (plain -> option promotion) — compiles again ✅ (`option_concatexpr_test.v` passes)
- or-block default that narrows `?string` -> `string` — clear V error ✅
- `if`/`match` branches that disagree on an element's option-ness — clear V error in **both** branch orders ✅ (the reverse order previously slipped through and emitted bad C; added to the regression test)

`vlib/v/tests/options/` (215 tests) and the multi-return/or-block suites all pass.

PR #27571, opened 2026-06-27
v3: fix ARM64 self-host (test_all.vsh step 4)

medvednikov:
## What

Makes `v3 -selfhost -b arm64` produce a working arm64 compiler that can in turn compile programs with the native ARM64 backend (test_all.vsh step 4). Before this, the arm64-self-hosted binary crashed immediately (parser stack overflow), then progressively later as each bug was fixed.

## ARM64 self-host now works end to end

```
v -o v3 vlib/v3/v3.v                         # build v3 with the bootstrap compiler
./v3 --no-parallel -selfhost -b arm64 -o v4_arm vlib/v3/v3.v   # v3 compiles itself via the native ARM64 backend
./v4_arm --no-parallel -b arm64 -o hello vlib/v3/tests/hello.v # the arm64-self-hosted compiler compiles a program
./hello                                       # -> hello world
```

## Bugs fixed

All in the ARM64-only SSA path (`ssa/builder.v`, `gen/arm64/gen.v`) except one shared transformer fix:

- **`array_push_many_ptr`** — registered as its own synthetic (the C-macro name the transformer emits for `arr << ptr` / fixed-array appends; was an unknown SSA function).
- **`<Enum>__autostr`** — synthesize enum-to-string helpers (the C backend emits these; the SSA path had none).
- **Duplicated enum member built as `0`** — `x = .member` / `return .member` whose member name is duplicated across enums (e.g. `number`) is now resolved against the assignment/return target type instead of defaulting to 0.
- **`(&Struct(rawptr)).field = x` wrote to NULL** — a pointer-cast selector base returned a null lvalue address.
- **Static type-method passed a phantom receiver** — a cross-module `Type.method(args)` (e.g. `token.Token.from_string_tinyv`) was mis-classified as a method call, shifting its argument into the wrong registers; now re-classified from the resolved signature's parameter count.
- **String literals had `is_lit = 0`** — the SSA `string` struct lacked an `is_lit` field and `materialize_string` didn't set it, so freeing a string literal aborted.
- **Slice clone / `&slice`** — `arr.clone()` lowers to a header-less synthetic `array__clone`, and inline slice rvalues (`a[lo..hi]`) are materialized to a temp when their address is taken.
- Stub the `v3.bench.*` rss helpers, matching the full import-path qualification that fn names now carry.

## Verification

- ARM64 self-host (test_all.vsh step 4): passes — `v4_arm` compiles `hello.v` to a working binary that prints `hello world`.
- C-backend self-host chain still converges byte-for-byte (`v5.c == v6.c`).
- No new unit-test failures attributable to this change.

### DIFF
--- a/vlib/v3/gen/arm64/gen.v
+++ b/vlib/v3/gen/arm64/gen.v
@@ -1411,7 +1411,13 @@ fn (mut g Gen) materialize_string(val_id int, reg int) {
 	g.macho.add_reloc(g.macho.text_data.len, str_sym_idx, arm64_reloc_pageoff12, false)
 	g.emit32(asm_add_pageoff(Reg(reg)))
 
-	g.emit_mov_imm(10, i64(str_len))
+	// x10 holds the string struct's second 8-byte word: `len` in the low 32 bits and
+	// `is_lit` in the high 32 bits (matching `string{ str, len int, is_lit int }`).
+	// Every caller stores/moves x10 as that whole word, so set is_lit=1 here: string
+	// literals point at static data and must NOT be passed to free() (`string.free`
+	// returns early only when is_lit==1). Without this, freeing a literal (e.g. the
+	// `mut res := ''` in os.real_path) aborts with a libmalloc "pointer not allocated".
+	g.emit_mov_imm(10, i64(str_len) | (i64(1) << 32))
 }
 
 // ==================== Global access ====================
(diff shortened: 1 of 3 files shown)


medvednikov:
@codex review

medvednikov:
@GGRei ARM64 self-host works now! Good to do x64 next! (Let's rename it to amd64 as was requested earlier)

GGRei:
@medvednikov GJ!

A follow-up PR should be ready later today to extend the V3 C baseline coverage further, with the goal of supporting all examples currently handled by the V2 x64 backend.

Once that baseline is done, migrating the V2 x64 backend to V3 will be the next objective. And yes, I’ll rename it to amd64 if that’s the preferred name. :)

PR #27477, opened 2026-06-17
refactor(fasthttp): make both Linux (epoll) and BSD (kqueue) backends asynchronous

guweigang:
### Summary
This PR refactors both the Linux (`epoll`) and BSD (`kqueue`) backends of the `fasthttp` module to be fully asynchronous using the `LoopCommand` channel pattern and `spawn` coroutines. 

It resolves the upstream review comments regarding the BSD kqueue async PR (#27477) by:
1. **P1 (Linux block/hang on loopback request)**: The Linux backend previously processed requests synchronously inline within the main epoll event loop thread, blocking the loop and causing timeouts for nested loopback requests (e.g. `test_handler_can_make_loopback_request_to_same_server`). We refactored `vlib/fasthttp/fasthttp_linux.v` to decouple request handling into async coroutines, allowing loopback requests to process without blocking.
2. **P2 (Connection lifetime race under `.manual` takeover)**: For `.manual` takeover mode, we deferred the `server.end_request()` call in both backends until the event loop has successfully deregistered the fd and cleaned tracking maps, avoiding early socket closures during server shutdown.
3. **Keep-alive epoll re-registration**: Fixed a bug where keep-alive connections could lose epoll registration after their first request if the write completed instantly (i.e. `epollout_was_armed` was false). The socket is now unconditionally re-added to the epoll listen mask on keep-alive completion.

### Key Changes
- **`vlib/fasthttp/fasthttp_linux.v`**: 
  - Added `LoopCommandKind` and `LoopCommand` structs.
  - Decoupled request processing into `process_request_async`.
  - Implemented `drain_loop_commands` to safely process commands and update connection/write maps in the main thread.
  - Unconditionally re-registers sockets to epoll for keep-alive connections on write completion.
- **`vlib/fasthttp/fasthttp_bsd.c.v`**: 
  - Deferred `server.end_request()` under `.manual_takeover` to the loop command handler.
- **`vlib/fasthttp/fasthttp_test.v`**: 
  - Re-enabled the loopback nested request test for Linux (`$if linux || bsd`).

### Verification
- Tested locally on macOS (BSD kqueue backend): `2 passed, 2 total`.
- Tested in a Linux (Debian aarch64) container (Linux epoll backend): `2 passed, 2 total` (including nested loopback test).

PR #27585, opened 2026-06-28
fix(fasthttp): rearm keep-alive poll events

guweigang:
This is the small standalone bug-fix split from #27477, following upstream feedback. It intentionally does not include the async-by-default/per-request spawn refactor.

Changes:
- Linux /epoll: after a keep-alive response completes, always DEL+ADD the fd back with EPOLLIN|EPOLLET so unread pipelined bytes get a fresh edge even when EPOLLOUT was never armed.
- BSD /kqueue: after a keep-alive response completes, re-add EVFILT_READ for the same reason.
- Manual takeover path: document and preserve the lifetime order by untracking/removing poll events before ending the active request.
- Add focused Linux and BSD regression tests for keep-alive read re-registration after the current edge was consumed.

Validation:
- macOS: ./vnew -silent test vlib/fasthttp/ -> 3 passed, 1 Linux-only skipped
- Ubuntu Orb VM: export VFLAGS='-gc none -cc gcc'; ./v -g -keepc -o ./vnew cmd/v -> passed
- Ubuntu Orb VM: ./vnew -silent test vlib/fasthttp/ -> 3 passed, 1 BSD-only skipped

The loopback self-call/deadlock scenario from #27477 is deliberately left for a separate design, since fixing that without hurting the fasthttp hot path likely requires an opt-in worker-pool/coroutine design rather than per-request OS threads.

### DIFF
--- a/vlib/fasthttp/fasthttp_bsd.c.v
+++ b/vlib/fasthttp/fasthttp_bsd.c.v
@@ -307,6 +307,7 @@ fn complete_response(server &Server, kq int, c_ptr voidptr, mut clients map[int]
 	}
 	c.read_start = 0
 	c.should_close = false
+	add_event(kq, u64(c.fd), i16(C.EVFILT_READ), u16(C.EV_ADD | C.EV_ENABLE | C.EV_CLEAR), c)
 }
 
 // process_request handles a complete HTTP request: decodes, calls the handler,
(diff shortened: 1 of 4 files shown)


JalonSolov:
@codex review

JalonSolov:
The CI failures appear to be unrelated to the changes in this PR... and most, if not all of them have been fixed in master.

PR #27587, opened 2026-06-28
cgen: cover heap struct literals in ternaries

### LINKED ISSUE #27329: cgen: corrupted C output for heap struct literal (`&T{}`) init — split identifier emitted at file scope
## Description

A real-world program produced **syntactically corrupted C** for a heap struct literal initialization. The generated C splits a type identifier in half (`main__Object` → `main__Obj` + `ect`) and splices a different element's initialization statements into the middle of an unfinished expression. The corrupted fragment is emitted **outside any function** (the C compiler reports "not in a function").

The struct is initialized via the all-fields `&T{...}` heap path (`direct_heap_struct_init` in `vlib/v/gen/c/struct.v`), which allocates with `builtin___v_malloc(...)` and then assigns each field individually. When several such heap initializations are reordered/hoisted (array element or nested element ordering), the statement-splitting bookkeeping (`go_before_last_stmt` / `go_before_ternary`) appears to cut at the wrong buffer position.

## Generated C (from the bug report)

```c
VV_LOC bool main__truthy(main__Object* obj) {
	return ((obj->kind == (_const_main__kind_always_true))? (true) : /* ... long ternary ... */ (true));
}
	main__Object* _t1 = (main__Obj	main__Object* _t2 = (main__Object*)builtin___v_malloc(sizeof(main__Object) == 0 ? 1 : sizeof(main__Object));
	_t2->kind = _const_main__kind_always_false;
	_t2->ival = 0;
	_t2->len_val = 0;
	ect*)builtin___v_malloc(sizeof(main__Object) == 0 ? 1 : sizeof(main__Ob	main__Object* _t3 = (main__Object*)builtin___v_malloc(sizeof(main__Object) == 0 ? 1 : sizeof(main__Object));
	_t3->kind = _const_main__kind_length_based;
```

The `_t1` initializer `(main__Object*)builtin___v_malloc(... sizeof(main__Object))` is broken in two: it begins with `(main__Obj`, the full `_t2` block is inserted, and the tail `ect*)builtin___v_malloc(... sizeof(main__Ob` reappears afterwards.

## C compiler error

```
error: 'main__Obj' undeclared here (not in a function); did you mean 'main__Object'?
```

## Triggering conditions (identified)

- Several `&T{...}` heap struct literals with **all fields specified** (this selects `direct_heap_struct_init`, the per-field `malloc`+assign form, instead of the `HEAP(T, (T){...})` compound-literal form), and
- those initializations are **reordered/hoisted** (array elements and/or nested elements), and
- the result lands at/near file scope.

The struct in the report: `kind`/`ival`/`len_val` integer fields, with `kind` values being module-level `const`s (`_const_main__kind_*`), plus a `truthy()` method (the long ternary).

## Repro status

Not yet reproduced on master (`98ea23b`). Targeted attempts — array of all-fields `&Object{...}`, nested `child &Object` (2–3 levels), and a top-level `const` array of `&Object{...}` — all produced **correct, well-ordered** C on current master (each nested allocation is fully hoisted before use). The exact trigger still needs to be isolated; the original source was not captured by the bug reporter (only the generated-C window above).

## Environment

- V 0.5.1
- Original report: Linux (cc)

> [!NOTE]
>

medvednikov:
## Summary
- extend the #27329 heap struct init fixture with all-field `&T{...}` literals in `if` assignment, returned `if`, and `match` expression branches
- add C-output checks that those ternary cases keep complete hoisted heap allocations before the branch expression

## Notes
#27349 fixed the stale statement-position bug for direct heap struct init hoists. The later issue comment isolated a ternary-shaped repro, so this adds coverage for that exact shape.

Fixes #27329

## Validation
- `./vnew run vlib/v/gen/c/testdata/heap_struct_init_order.vv`
- `VTEST_ONLY=heap_struct_init_order ./vnew -silent vlib/v/gen/c/coutput_test.v`

### DIFF
--- a/vlib/v/gen/c/coutput_test.v
+++ b/vlib/v/gen/c/coutput_test.v
@@ -286,6 +286,52 @@ fn test_comptime_for_empty_attrs_does_not_emit_new_array_calls() {
 	assert !compilation.output.contains('.args = builtin____new_array_with_default(0, 0, sizeof(FunctionParam), 0)')
 }
 
+fn assert_function_body_contains(output string, signature string, expected_lines []string) {
+	assert output.contains(signature), 'missing generated function: ${signature}'
+	body := output.all_after(signature).all_before('\n}')
+	for expected_line in expected_lines {
+		assert body.contains(expected_line), 'missing expected C line in ${signature}: ${expected_line}'
+	}
+}
+
+fn test_heap_struct_init_order_ternary_helpers_have_branch_allocs() {
+	os.chdir(vroot) or {}
+	path := os.join_path(testdata_folder, 'heap_struct_init_order.vv')
+	relpath := vroot_relative(path)
+	if should_skip(relpath) {
+		return
+	}
+	file_options := get_file_options(path)
+	cmd := '${os.quoted_path(vexe)} -o - ${file_options.vflags} ${os.quoted_path(path)}'
+	compilation := os.execute(cmd)
+	ensure_compilation_succeeded(compilation, cmd)
+	alloc := '(main__Issue27329Object*)builtin___v_malloc(sizeof(main__Issue27329Object) == 0 ? 1 : sizeof(main__Issue27329Object));'
+	assert_function_body_contains(compilation.output,
+		'VV_LOC main__Issue27329Object* main__issue_27329_if_expr_assignment(bool flag) {', [
+		'main__Issue27329Object* _t1 = ${alloc}',
+		'_t1->ival = 4;',
+		'main__Issue27329Object* _t2 = ${alloc}',
+		'_t2->ival = 5;',
+		'main__Issue27329Object* object = (flag ? ( _t1) : ( _t2));',
+	])
+	assert_function_body_contains(compilation.output,
+		'VV_LOC main__Issue27329Object* main__issue_27329_if_expr_return(bool flag) {', [
+		'main__Issue27329Object* _t2 = ${alloc}',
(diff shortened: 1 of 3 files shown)


medvednikov:
Addressed the review by strengthening the ternary heap-literal coverage:

- added branch-specific malloc/temp initializer patterns to `heap_struct_init_order.c.must_have`
- added body-scoped assertions in `coutput_test.v` for the three new helper functions, so the checks are tied to the generated function bodies rather than only global substrings

Validation rerun:
- `./vnew run vlib/v/gen/c/testdata/heap_struct_init_order.vv`
- `VTEST_ONLY=heap_struct_init_order ./vnew -silent vlib/v/gen/c/coutput_test.v`

medvednikov:
Addressed the latest review:

- the body-scoped `heap_struct_init_order` C-output check now calls `should_skip(relpath)` before compiling directly
- it uses `get_file_options(path).vflags`, so the fixture's `// vtest vflags: -gc boehm` is honored like the `.out`/`.c.must_have` harnesses

Validation rerun:
- `./vnew run vlib/v/gen/c/testdata/heap_struct_init_order.vv`
- `VTEST_ONLY=heap_struct_init_order ./vnew -silent vlib/v/gen/c/coutput_test.v`
- `GITHUB_JOB=musl VTEST_ONLY=heap_struct_init_order ./vnew -silent vlib/v/gen/c/coutput_test.v`

PR #27605, opened 2026-06-30
gg: add draw_rounded_rect_border function

tzSharing:
Added the `draw_rounded_rect_border` function to the `gg` library and created `testData`. Now, you can directly draw rounded rectangles with borders using `draw_rounded_rect_border`. Both the border width and color are fully customizable. Additionally, by adjusting the parameters, you can also easily achieve hollow (outline-only) and solid filled rounded rectangle effects.
draw_rounded_rect_border


tzSharing:
@codex review

PR #27609, opened 2026-06-30
[codex] speed up v3 compiler builds

medvednikov:
## What changed

This speeds up v3 compiler build paths while preserving correctness for both v3 self-builds and `cmd/v` builds.

- Avoid treating `cmd/v` as the `-building-v` generic-skip path, because v1 compiler sources do require generic helper monomorphization.
- Keep the v3 self-build fast path by using cheap generic-template erasure and skipping annotate/generic transform work where safe.
- Skip post-transform type annotation for `cmd/v` builds, while still running full monomorphization.
- Cache monomorphize node/module mapping and generic receiver method lookup to avoid repeated whole-AST scans.
- Reuse collected C function generation items and fixed-array typedef discovery in cgen.
- Fix headerless macOS `_dyld_get_image_name` declaration handling.

## Why

The previous generic-skip path could look fast for `cmd/v`, but produced invalid C. This keeps correctness, removes avoidable repeated work, and makes v3 self-build template erasure cheap again.

## Validation

- `./vnew -gc none -prod -o vlib/v3/v3 vlib/v3/v3.v`
- `cd vlib/v3 && ./v3 -o v1 ../../cmd/v`
- `cd vlib/v3 && ./v3 -building-v -o v4 v3.v`
- `./vnew -silent test vlib/v3/tests/generics_test.v vlib/v3/tests/prod_flag_test.v vlib/v3/tests/export_attr_codegen_test.v vlib/v3/tests/test_file_harness_codegen_test.v vlib/v3/tests/selfhost_backend_prune_test.v`
- `git diff --check -- vlib/v3/gen/c/cleanc.v vlib/v3/gen/c/fn.v vlib/v3/gen/c/fn_d_parallel.v vlib/v3/markused/markused.v vlib/v3/transform/expr.v vlib/v3/transform/monomorphize.v vlib/v3/transform/transform.v vlib/v3/v3.v`

## Notes

On this machine, `cmd/v` now compiles successfully through clang after tcc reports `ARM asm not implemented`, which is the existing fallback behavior on this path.

### DIFF
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -29,6 +29,7 @@ mut:
 	a                            &flat.FlatAst = unsafe { nil }
 	used_fns                     map[string]bool
 	used_fn_names                []string
+	fn_gen_items                 []FlatFnGenItem
 	test_files                   map[string]bool
 	str_lits                     []string
 	str_lit_ids                  map[string]int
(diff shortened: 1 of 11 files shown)


medvednikov:
@codex review

PR #27625, opened 2026-07-02
v3: make 10 more tests pass

medvednikov:
## Summary
- fix several V3 parser/checker/codegen gaps so 10 more listed tests pass
- handle local type declarations, keyword field/enum cases, @FILE_LINE, tuple-like multi-return if/match, boolean exhaustive match returns, and ...voidptr variadic rvalues
- move the 10 passing tests from v3_test_failures.txt to v3_test_passes.txt

## Validation
- ./vnew -d parallel -g -keepc -o /tmp/v3_local vlib/v3/v3.v
- 10 promoted tests passed with /tmp/v3_local
- ./vnew -silent vlib/v3/tests/parallel_cgen_test.v
- ./vnew -silent vlib/v3/tests/selfhost_backend_prune_test.v
- git diff --check

Note: I also ran ./vnew -silent test vlib/v3/ before the final self-host cleanup; it reported 73 passed / 8 failed. The two self-host-related failures from that run were rerun and now pass.

### DIFF
--- a/vlib/v3/gen/c/array.v
+++ b/vlib/v3/gen/c/array.v
@@ -149,7 +149,7 @@ fn (mut g FlatGen) gen_fixed_array_data_arg(id flat.NodeId, arr types.ArrayFixed
 	}
 	// A fixed-array value (e.g. `[4]u8` color) is sometimes represented as a dynamic
 	// `Array`; a C fixed-array parameter decays to `elem*`, so pass the data pointer.
-	if g.tc.resolve_type(id) is types.Array {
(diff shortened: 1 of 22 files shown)


medvednikov:
@codex review

Your latest review is already covered in the current working tree.

The fix is in parser.v: local types now synthesize as `Row@local@...`, which user source cannot declare as a normal top-level type. C names preserve that marker via [naming.v], so it does not collapse back to the old underscore-only spelling.

Regression is already in [type_checker_errors_test.v]

Issue #27638, opened 2026-07-03
time: allocation-free fixed-buffer variant of push_to_http_header (cached HTTP Date headers)


enghitalo:
## Describe the feature

HTTP servers cache the `Date:` header line and refresh it once per second (nginx model). The natural vlib API for that is `time.push_to_http_header`, but its current shape makes an allocation-free cached-Date rebuild impossible:

1. It can only **append to a dynamic `[]u8`** — there is no way to write the 29 bytes into caller-owned storage (a fixed array, or a preallocated region of a dynamic array) at an offset.
2. Internally it calls `weekday_str()` and `smonth()`, which each do a substr (`long_days[i][0..3]`, `months_string[i*3..(i+1)*3]`) — **two heap allocations per call**.

### Measurements (current master, `-prod`, 1M iterations, x86-64 Linux)

| approach | per call |
|---|---|
| `line.clear()` + appends + `time.utc().push_to_http_header(mut line)` | **~1.25 µs** |
| `time.utc()` alone | ~1.2 µs |
| `C.time(0)` | ~2.5 ns |
| fixed `[37]u8` cache, incremental digit update (reference impl below) | **~9.2 ns (135x)** |

## Proposed Solution — minimal diff, validated (PR #27639)

The body of `push_to_http_header` already builds the 29 bytes in a stack template via the pointer-based `int_to_ptr_byte_array_no_pad` — it is 90% of the needed API. The change moves that body into `@[unsafe] write_http_header(dst &u8, dst_len int) !` — pointer-based so one API serves fixed arrays, dynamic arrays and arenas at any offset; **bounds-checked** (refuses `dst_len < time.http_header_len`, writing nothing) so a caller cannot silently overrun; `@[unsafe]`-tagged so the raw-pointer contract is explicit at the call site (like `vbytes`). `push_to_http_header` becomes a thin wrapper with unchanged behavior — all 17 `vlib/time` tests pass, output verified byte-identical to `http_header_string()` across second/minute/hour/day/leap-year transitions on both storage shapes.

```diff
diff --git a/vlib/time/format.v b/vlib/time/format.v
index b914ea0e8..beae94955 100644
--- a/vlib/time/format.v
+++ b/vlib/time/format.v
@@ -736,24 +736,42 @@ pub fn (t Time) http_header_string() string {
 	return buf.bytestr()
 }
 
-// push_to_http_header returns a date string in the format used in HTTP headers, as defined in RFC 2616.
[...]

PR #27642, opened 2026-07-03
time: declare the Linux timerfd C API (timerfd_create/settime/gettime)

enghitalo:
## Summary

Declare the Linux `timerfd` API once, in the `time` module (`timerfd_linux.c.v`): `fn C.timerfd_create/C.timerfd_settime/C.timerfd_gettime` + `pub struct C.itimerspec` (reusing the module's existing `C.timespec`) + the `#include <sys/timerfd.h>`.

C declarations are visible to importing modules and the include lands in the generated C globally — so `import time` is enough for event-loop code to use the kernel API directly, instead of every project copying the same three declarations and a hand-built `itimerspec`:

```v ignore
import time

$if linux {
	tfd := C.timerfd_create(C.CLOCK_MONOTONIC, C.TFD_CLOEXEC | C.TFD_NONBLOCK)
	spec := C.itimerspec{
		it_value:    C.timespec{1, 0}
		it_interval: C.timespec{1, 0} // kernel-paced 1 s tick, no drift
	}
	C.timerfd_settime(tfd, 0, &spec, unsafe { nil })
	// register tfd with epoll / os.notify; on readiness:
	mut expirations := u64(0)
	C.read(tfd, &expirations, 8) // re-levels the fd; > 1 reports missed ticks
}
\`\`\`

## Why declarations-only (reworked from the first version of this PR)

The first iteration was a full `os.timerfd` wrapper module — but V values cross-platform APIs, and timerfd is a Linux kernel interface with no honest portable equivalent (kqueue's `EVFILT_TIMER` is not a readable fd you can hand to *another* multiplexer, so a portable façade would leak). Plain declarations in a `_linux.c.v` file are the honest shape: zero abstraction, zero overhead, `$if linux` at call sites, and nothing pretending to be portable.

## Tests

`vlib/time/timerfd_test.c.v` (self-skips off Linux; the module and importers still compile everywhere — verified `v -check -os windows|macos`): periodic 5 ms timer via the declared API, EAGAIN before expiry on a non-blocking fd, missed-tick counting after a 23 ms sleep (`read` returns the u64 expiration count), disarm via zero `it_value`, `timerfd_gettime` readback.

Use case: a per-worker cached `Date:` header refreshed by a 1 s timerfd inside the worker's own epoll loop (no extra thread, no locks) — https://github.com/enghitalo/vanilla/blob/main/examples/async_date_timerfd/main.v — related to #27638 / #27639 / #27641.

### DIFF
--- a/vlib/time/timerfd_linux.c.v
+++ b/vlib/time/timerfd_linux.c.v
@@ -1,0 +1,34 @@
+module time
+
+// Declarations for the Linux timerfd API (timerfd_create(2)): kernel timers
+// delivered as FD READABILITY. Import `time` and call them directly — create,
+// arm with C.timerfd_settime + C.itimerspec, then read 8 bytes (a u64 count
+// of expirations since the last read) from the fd, or register the fd with
+// epoll / os.notify and treat expiry as just another readiness event.
+//
+// This file is only compiled on Linux; guard call sites with `$if linux {`.
+//
+// Example (a 1 s periodic tick inside an event loop):
+// ```v ignore
+// tfd := C.timerfd_create(C.CLOCK_MONOTONIC, C.TFD_CLOEXEC | C.TFD_NONBLOCK)
+// spec := C.itimerspec{
+// 	it_value:    C.timespec{1, 0}
+// 	it_interval: C.timespec{1, 0}
+// }
+// C.timerfd_settime(tfd, 0, &spec, unsafe { nil })
+// // ... register tfd for readability; on the event:
+// mut expirations := u64(0)
+// C.read(tfd, &expirations, 8) // re-levels the fd; > 1 reports missed ticks
+// ```
+
+#include <sys/timerfd.h>
+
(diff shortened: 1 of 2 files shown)


enghitalo:
Note for reviewers: the three inline comments above were made against the first revision of this PR, which was a full `os.timerfd` wrapper module. That approach was dropped — V favours cross-platform APIs, and timerfd has no honest portable equivalent (kqueue's `EVFILT_TIMER` is not a readable fd), so the PR is now **declarations only** in `vlib/time/timerfd_linux.c.v` (`fn C.timerfd_create/settime/gettime` + `C.itimerspec`). The flagged code — `Timerfd{}` closing stdin, EINTR handling in a wrapper `read`, the test racing the next tick — no longer exists; call sites use the raw C API under `$if linux` and own their own fd lifetime and EINTR policy.

medvednikov:
@codex review

PR #27650, opened 2026-07-04
v3: fix another batch of tracked test failures

medvednikov:
## Summary
- Fix another batch of V3 markused, transform, and C codegen issues exposed by tracked text failures.
- Move 50 validated V3 test entries from `v3_test_failures.txt` to `v3_test_passes.txt`.
- Add regressions for receiver/import homonyms, interface dispatch, C type storage, generic cross-module calls, and related post-merge review cases.

## Validation
- `./vnew -gc none -path "$(pwd)/vlib|@vlib|@vmodules" -o /tmp/v3_batch vlib/v3/v3.v`
- `./vnew -silent test vlib/v3/tests/markused_test.v`
- `./vnew -silent test vlib/v3/tests/mini_calculator_markused_codegen_test.v`
- `/tmp/v3_batch vlib/compress/lz/lz_test.v -o /tmp/v3_lz_test_bin`
- `/tmp/v3_batch vlib/v/tests/generics/generic_array_ret_test.v -o /tmp/v3_generic_array_ret_bin`
- `/tmp/v3_batch vlib/v/util/util_test.v -o /tmp/v3_util_test_bin`
- `/tmp/v3_batch vlib/math/stats/stats_test.v -o /tmp/v3_stats_test_bin`
- 50-file V3 candidate sweep: 50 passed, 0 failed
- Manifest check: 0 pass/fail overlap, 0 candidates still in failures, 50 candidates in passes

### DIFF
--- a/vlib/v/tests/enums/enum_flag_from_test.v
+++ b/vlib/v/tests/enums/enum_flag_from_test.v
@@ -19,3 +19,20 @@ fn test_fail() {
 	a := Flag.from(0b110101110) or { Flag(Flag.flag0) }
 	assert a == Flag.flag0
 }
+
+struct FlagFromCounter {
+mut:
+	calls int
+}
+
+fn (mut counter FlagFromCounter) next() int {
+	counter.calls++
+	return 0b110
+}
+
+fn test_from_evaluates_argument_once() {
+	mut counter := FlagFromCounter{}
+	a := Flag.from(counter.next()) or { panic(err) }
+	assert counter.calls == 1
+	assert a == Flag.flag1 | .flag2
+}
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -2800,6 +2800,12 @@ fn (mut g FlatGen) gen_expr_with_expected_type(id flat.NodeId, expected types.Ty
 		g.expected_enum = expected.name
 	}
 	node := g.a.nodes[int(id)]
+	if g.is_ierror_type_name(expected.name()) && node.kind == .none_expr {
+		g.write(g.ierror_none_literal_string())
+		g.expected_expr_type = old_expected
+		g.expected_enum = old_expected_enum
+		return
+	}
 	if g.is_ierror_type_name(expected.name()) && g.expr_is_error_call(id) {
 		g.gen_ierror_from_error_call(node)
 		g.expected_expr_type = old_expected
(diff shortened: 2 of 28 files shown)


medvednikov:
Reviewed PR #27650 from the GitHub diff and CI logs. I would **request changes** before merging.

### Blocking findings

**1. CI is already red, and one failure points directly at this PR’s C extern parameter change.**
The PR’s V3 CI run completed with `conclusion: failure`, and the failed job is the `Test v3` step. The logs show `vlib/v3/tests/c_anonymous_param_codegen_test.v` failing because it still expects the typedef-shaped prototype `int XLookupString(AnonNative* event);`.

The new `c_extern_param_c_type()` now emits `struct ${base.name[2..]}*` for every non-preamble `&C.X` struct, which ignores `@[typedef] struct C.AnonNative` cases. The existing test defines `@[typedef] struct C.AnonNative` and explicitly asserts both `AnonNative*` output and no `struct AnonNative*` output.

Suggested fix: preserve typedef C struct spelling. For example, only emit `struct X*` for non-typedef C structs, or ask the type/AST metadata whether the `C.` struct came from an `@[typedef]` declaration before overriding `g.tc.c_type(pt)`.

**2. `Flag.from(expr)` evaluates `expr` twice.**
`gen_flag_enum_from_call()` converts the argument to a string once, but then interpolates that same expression in both `.ok` and `.value`:
`... (${arg}) ... .value = (${value_ct})(${arg})`. That means `Flag.from(next())` or `Flag.from(arr.pop())` would run the side effect twice. The added/covered `enum_flag_from_test.v` cases only use integer literals, so this regression is not covered.

Suggested fix: emit a statement-expression/temp so the argument is evaluated once, then use the temp in both the mask check and value assignment.

**3. The flattened generic receiver “short variant” helper appears to be a no-op.**
`cgen_flattened_generic_receiver_short_variants()` first checks that the input contains `__`, but then does `leaf.split('_')` and checks each part for `part.contains('__')`. After splitting on single underscores, the parts should no longer contain `__`, so `changed` never becomes true and the function returns `[]`. This undermines the new lookup paths that rely on this helper for flattened generic receiver aliases.

Suggested fix: parse `__` before splitting on `_`, or transform module-qualified fragments with a regex/state-machine over the original string. A small unit test for something like `foo__Bar_baz__Qux` would catch this.

### Other notes

The PR is large for a “batch” fix: 25 files changed, 2,335 additions, and 278 deletions, with multiple unrelated compiler areas touched in one PR. That makes the red CI harder to triage and raises regression risk.

I’d split or at least fix the C typedef regression and `Flag.from` double-evaluation before re-running the full V3 suite.

PR #27653, opened 2026-07-04
v3: fix transform regression followups

medvednikov:
Summary:
- allow v3 array literals to accept newline/semicolon separated entries with line comments
- preserve mut fn literal captures and pointer-value rvalue derefs while avoiding selector-base over-deref
- handle optional/result void or-block lowering and correct truncate C fd signatures

Tests:
- ./vnew -silent test -run-only test_const_array_allows_newline_separators_with_line_comments vlib/v3/tests/review_transform_regressions_test.v
- ./vnew -silent test vlib/os/os_test.c.v


medvednikov:
Reviewed the new head `19bcf27e607c740363a4ad2bf43dab8d50381de2`, ignoring CI.

I’d still request changes, but only for one blocker.

### Finding

**P1 — `p := &v` for a heap-moved local now lowers to the value, not the pointer.**

The heap-escaping path explicitly says that when `v` has already been moved to the heap, the `p := &v` alias should be “simply that pointer,” and the implementation returns `t.transform_expr(amp_child)`. This PR now marks the heap-moved source local in `pointer_value_rvalues`, and `transform_ident_expr()` turns such rvalue idents from `v` into `*v`. So the special alias path that should produce `p := v` now produces `p := *v`, while the declaration path still sets `p`’s inferred type to the source pointer type. That can create a `&T` declaration initialized from a `T` value and breaks the exact stale-mutation escape case this code is trying to preserve.

A regression to add:

```v
struct S {
	n int
}

fn leak() &S {
	mut s := S{
		n: 1
	}
	p := &s
	s = S{
		n: 2
	}
	return p
}

fn main() {
	p := leak()
	println(int_str(p.n))
}
```

Expected output should be `2`. With the current change, the `p := &s` lowering can use the new rvalue auto-deref and materialize `*s` instead of the heap pointer `s`.

A narrow fix would be to avoid the normal rvalue transform in this special branch, for example by returning a plain ident for the heap pointer path and preserving its `&T` type, or by temporarily suppressing `pointer_value_rvalues` only while lowering that alias.

PR #27657, opened 2026-07-04
checker: reject address of bound method values

### LINKED ISSUE #27656: V Compiler Bug (Taking address of conversion expression generates invalid C)
<details>
<summary>V version: V 0.5.1, press to see full `v doctor` output</summary>

|V full version      |V 0.5.1 0c3183c55b39534f9bb0d2f796bb575d39c9d229
|Processor           |8 cpus, 64bit, little endian, Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
|Memory              |4.31GB/19.29GB
|                    |
|V executable        |/usr/lib/vlang/v
|V last modified time|2026-07-04 08:19:04
|                    |
|V home dir          |NOT writable, value: /usr/lib/vlang
|VMODULES            |OK, value: /home/specter/.vmodules
|Current working dir |OK, value: /home/specter/Templates/vtest
|                    |
|V git status        |N/A
|.git/config present |false
|                    |
|cc version          |cc (GCC) 16.1.1 20260625
|gcc version         |gcc (GCC) 16.1.1 20260625
|clang version       |clang version 22.1.6
|tcc git status      |Error: fatal: detected dubious ownership in repository at '/usr/lib/vlang/thirdparty/tcc'

	git config --global --add safe.directory /usr/lib/vlang/thirdparty/tcc
 Error: fatal: detected dubious ownership in repository at '/usr/lib/vlang/thirdparty/tcc'

	git config --global --add safe.directory /usr/lib/vlang/thirdparty/tcc

|emcc version        |N/A

</details>

**What did you do?**
`./v -g -o vdbg cmd/v && ./vdbg  main.v && main`
```v
module main

import os

fn main() {
	x := os.input('> ')

	y := &x.f64

	println(y)
}

```

**What did you see?**
```
cc: /tmp/v_1000/main.01KWQDE14FE4FPWK5TEN4ZYFTB.tmp.c:6780: error: lvalue expected
builder error: 
==================
```

produce a V compiler error explaining that taking the address of a temporary value is not allowed.

> [!NOTE]
>

medvednikov:
Fixes #27656.

## Summary
- Reject taking the address of selectors resolved as bound method values, such as `&x.f64`.
- Add a regression case to the prefix error fixture.

## Testing
- `VTEST_ONLY=prefix_err.vv ./vnew -silent vlib/v/compiler_errors_test.v`
- `./vnew -silent test vlib/v/checker/`
- `./vnew -silent test vlib/v/` (fails in `vlib/v/tests/loops/for_in_mut_array_with_mul_test.v`; rerun alone fails with generated C `*i = i * i`)


medvednikov:
Reviewed PR **vlang/v#27657**.

**Verdict:** I don’t see any blocking correctness issues. The change looks targeted and appropriate for the linked bug, and I’d be comfortable with it once the normal CI finishes.

The linked issue is caused by `&x.f64`, where `x.f64` is resolved as a selector-style conversion/bound method value and cgen later emits invalid C with `lvalue expected` instead of the checker rejecting it.  The PR adds the check in `prefix_expr` after the RHS has been checked and parentheses removed, specifically inside the existing `&selector` validation path. That means cases like `&s.f64` and `&(s.f64)` should be caught before reaching codegen.

The implementation uses `ast.SelectorExpr.has_hidden_receiver`, which is already part of the selector AST node, so the fix is not guessing from the selector name or type string.  The new error is also placed alongside related addressability checks for selectors, such as rejecting addresses of literal-backed selectors, struct-init member access, minified fields, option fields, and map values outside `unsafe`, which is a reasonable location.

The regression test is minimal and directly covers the failure mode by adding:

```v
s := '1.2'
_ := &s.f64
```

and expecting `cannot take the address of s.f64`.  That is a good fit for the existing `prefix_err` fixture.

One small non-blocking suggestion: since the PR title says “bound method values,” it may be worth adding a second regression for a user-defined bound method value if V supports that syntax in the same way. The current test covers the reported builtin conversion case, which is enough for the linked issue, but a user-defined method-value case would make the intent of `has_hidden_receiver` clearer.

CI was still mostly queued or in progress when I checked; several jobs had already succeeded, and the only completed failure shown was the automated “Claude Code Review” job, not one of the main compiler/test jobs.

PR #27658, opened 2026-07-04
cgen,builder,markused,vcache: make -usecache correct for multi-module builds (fix #27592)

### LINKED ISSUE #27592: `-usecache` link failure on the clang/gcc backend (macOS arm64): duplicate `__global` symbols (g_autostr_*_stack) from cached builtin.o
### Describe the bug

`-usecache` produces a binary that fails to **link** when the C backend is `clang` (or any compiler whose linker enforces ODR for tentative definitions, i.e. everything except `tcc`). On macOS the default system `cc` *is* clang, so `v -cc cc -usecache ...` is broken for every program — including a bare `println('hello')`.

The link fails with **duplicate symbols** for the `__global` cycle-guard variables declared in `vlib/builtin/autostr.v`:

```
duplicate symbol '_g_autostr_addr_stack' in:
    /var/folders/.../T/hello-XXXX.o
    /Users/me/.vmodules/.cache/68/6808fcb...module.builtin.o
duplicate symbol '_g_autostr_type_stack' in:
    /var/folders/.../T/hello-XXXX.o
    /Users/me/.vmodules/.cache/68/6808fcb...module.builtin.o
ld: 2 duplicate symbols
```

The reason this has gone unnoticed: the only `-usecache` coverage in `cmd/tools/vtest-all.v` runs against the **default cc**, which on the CI hosts is `tcc`. `tcc` treats these as tentative/common definitions and silently merges them, so the clang/gcc-linker path is never exercised.

### Reproduction Steps

On macOS arm64 (any program reproduces; hello-world is enough):

```sh
echo "fn main() { println('hello') }" > /tmp/h.v
v wipe-cache
v -cc clang -usecache run /tmp/h.v     # or: v -cc cc -usecache run /tmp/h.v
```

For contrast, the tcc path succeeds:

```sh
v wipe-cache
v -usecache run /tmp/h.v               # default cc = tcc -> prints "hello"
```

### Expected Behavior

`v -cc clang -usecache run /tmp/h.v` links and prints `hello`, matching both the non-cached build (`v -cc clang run /tmp/h.v`) and the tcc-cached build.

### Current Behavior

Link fails with duplicate `_g_autostr_addr_stack` / `_g_autostr_type_stack` symbols (see above), exit code 1.

### Root cause

`g_autostr_type_stack` / `g_autostr_addr_stack` (and the `_len` companions) are `__global` declarations in `vlib/builtin/autostr.v`:

```v
__global g_autostr_type_stack     = [autostr_type_stack_max_depth]int{}
__global g_autostr_type_stack_len = 0
__global g_autostr_addr_stack     = [autostr_type_stack_max_depth]voidptr{}
```

The C backend emits a **definition** (not an `extern` declaration) for each `__global` into every translation unit that sees it. Without `-usecache` everything is one TU, so there is a single definition. With `-usecache`, `builtin` is compiled into a standalone `module.builtin.o` that defines them, and the main program's generated `.c` *also* defines them — two definitions of the same symbol at link time.

`tcc` accepts this because it treats uninitialised/`__global` definitions as tentative (common-block) symbols and merges them; `ld64` (and GNU `ld` without `-fcommon`) reject duplicate strong definitions, which is why the failure only shows up off the tcc path.

The fix is presumably to emit `__global` definitions in exactly one TU under `-usecache` (the owning module's object) and `extern` declarations everywhere else — the same shape that already works for ordinary cached module functions.

### Environment

- `v version`: V 0.5.1, commit `b156327` (master, 2026-06-28)
- OS: macOS 26.4.1 (build 25E253), arm64
- cc: Apple clang 21.0.0 (`clang-2100.1.1.101`)
- Also reproduces with `-cc gcc` (any non-tcc linker).

> [!NOTE]
>

eptx:
Fixes the family of `-usecache` correctness bugs root-caused in #27592 (duplicate `__global` symbols there are one instance of a general pattern), and makes cache invalidation sound. Includes a multi-module regression test.

## Root cause

Under `-usecache` the program TU is compiled with `skip_unused` ON while each cached module (`v build-module`) is compiled with `skip_unused` OFF, so the two disagree about what is defined vs referenced (and about type-index numbering). This PR consistently routes every cross-TU definition/initialization through a single owner — the program TU — with cached modules referencing `extern`.

## Layers fixed

- **`__global`s (#27592):** extern globals emit decl-only in the program TU instead of a duplicate definition; `g_main_argc`/`g_main_argv` are defined by the program TU (they otherwise have no definition at all under `-usecache`).
- **fixed-array globals:** the `_vinit`-time init form `dst = *(T*)&((T[]){..}[0])` is illegal C for arrays (not assignable) — emit memcpy, mirroring the existing struct path.
- **sumtype `is`:** the fallthrough compared a module-local `int(right_type)` literal while the cast wrappers and `match` use the unified `_v_type_idx_<T>()` value — so `x is T` silently disagrees with how the value was tagged across cached objects (misdispatch, no diagnostic). Now `type_sidx()` (a no-op for plain builds).
- **consts:** cached objects carry their consts as per-TU `static`s, but only the program TU has `_vinit` const-init code — a cached module's consts stay zeroed (e.g. `max_int == 0` inside `builtin.o` makes `array.push` panic on any growth; `strconv`'s tables read as zeros). Each cached object now emits an idempotent `<mod>__init_consts()`, and the program TU calls them all *before* its own inline const inits (so const-generator functions that run in cached objects, e.g. ed25519's `d_const`, see live values).
- **closures:** (a) `closure_init()` was only called when the program TU itself generated a closure (`nr_closures > 0`), so a closure created inside a cached module ran on an uninitialized trampoline pool → segfault; now gated on the whole-program `used_features.anon_fn`. (b) Bundled modules (`util.bundle_modules`: `builtin.closure`, `builtin.overflow`) are skipped in *every* build-module compile, yet the program TU also emitted them decl-only for non-test builds — their bodies existed in no object: `undefined symbol: builtin__closure__closure_create_with_data` for any non-test `-usecache` build that uses closures.
- **definition-`#include`s** (C amalgamations): emitted only in the owning TU, so the same C definitions don't land in several cached objects.
- **test builds:** `handle_usecache` cached `builtin` a second time via its absolute path (missing the import-loop guards on the test-module loop) → both objects linked, ~9000 duplicate symbols. Also, `-usecache` test builds re-emitted every used cached-module body in the program TU (blanket `!is_test`); now only fns defined in `_test.v` files get bodies there.
- **cache invalidation:** source hashes were saved *before* invalidated modules were rebuilt, and `v_build_module`'s exit status was ignored — one failed rebuild (racing edit, transient cc error) leaves a stale object that every later build silently links until a manual `v wipe-cache`. And cache keys ignored which compiler built an object, so a rebuilt `v` (after `v self` from uncommitted changes) reuses objects produced by an older cgen. Failed rebuilds are now loud + remove the stale object before hashes are recorded, and the cache namespace is salted with the running v executable's size+mtime.

## Regression test

`vlib/v/tests/usecache_multi_module_test.v` + `testdata/usecache_multi_module/`: a small multi-module project exercising, across the program-TU/cached-module boundary: sumtype `is`, interface dispatch, closures created in the cached module, module consts / array growth (`max_int`), and float formatting (strconv's const tables in the cached builtin object). Compiled as a normal program and as a test build, cold and warm cache each, with an isolated `VCACHE`/`VTMP`. On vanilla master the program build fails to link and the test build has ~9000 duplicate symbols; with this PR everything passes.

## Validation

(macOS arm64, `-cc clang`; every `-usecache` item below fails on vanilla master)

- `v wipe-cache && v -cc clang -usecache run hello.v` — the #27592 repro — links and runs, cold and warm.
- The new regression test: passes on this branch, fails on vanilla master.
- `v -cc clang -usecache -o v2 cmd/v` — the compiler itself builds under `-usecache` with clang and the resulting binary works (the scenario the CI task disabled on macOS in #23145).
- `v test vlib/v/tests/usecache_interface_index_symbol_test.v` (the #27381 regression test), `v test vlib/v/vcache/` — pass; `v test vlib/builtin/` — identical results to master; `v fmt -verify` clean on all touched files; `v self` fine.

Happy to split this if preferred — the builder/vcache invalidation fixes and the cgen cross-TU fixes are cleanly separable.

Fixes #27592

### DIFF
--- a/vlib/v/ast/ast.v
+++ b/vlib/v/ast/ast.v
@@ -1123,6 +1123,7 @@ pub mut:
 pub struct EmbeddedFile {
 pub:
 	compression_type string
+	source_file      string // path of the .v file containing the $embed_file call; the anchor for relative-path resolution (a const may be checked from an importing file's context, so Checker.file is not reliable here)
 pub mut:
 	rpath string // used in the source code, as an ID/key to the embed
 	apath string // absolute path during compilation to the resource
@@ -1165,6 +1166,7 @@ pub mut:
 	used_imports          []string
 	implied_imports       []string                  // ​imports that the user's code uses but omitted to import explicitly, used by `vfmt`
 	embedded_files        []EmbeddedFile            // list of files to embed in the binary
+	embedded_apaths       []string                  // absolute paths of $embed_file targets (deduped, checker-resolved); drives -usecache invalidation of embedded assets
 	imported_symbols      map[string]string         // used for `import {symbol}`, it maps symbol => module.symbol
 	imported_symbols_trie token.KeywordsMatcherTrie // constructed from imported_symbols, to accelerate presense checks
 	imported_symbols_used map[string]bool
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -1321,6 +1321,11 @@ fn (mut v Builder) setup_output_name() {
 			get_dsc_content('PREF.PATH: ${v.pref.path}\nVOPTS: ${v.pref.cache_manager.vopts}\n')) or {
 			panic(err)
 		}
+		// Record the content hashes of every $embed_file target that went into this
+		// cached object, so rebuild_cached_module can invalidate it when an embedded
+		// asset changes (embedded assets are invisible to the .v source hashes).
+		v.pref.cache_manager.mod_save(v.pref.path, '.embeds.txt', v.pref.path,
+			embeds_manifest(v.parsed_files)) or { panic(err) }
 	}
 	if os.is_dir(v.pref.out_name) {
 		verror('${os.quoted_path(v.pref.out_name)} is a directory')
--- a/vlib/v/builder/rebuilding.v
+++ b/vlib/v/builder/rebuilding.v
@@ -237,9 +237,25 @@ fn (mut b Builder) v_build_module(vexe string, imp_path string) int {
 	$if trace_v_build_module ? {
 		eprintln('> Builder.v_build_module: ${rebuild_cmd}')
 	}
+	// The child must reconstruct EXACTLY the parent's build configuration from
+	// the relayed b.pref.build_options — the parent already folded VFLAGS into
+	// those while parsing its own args. Letting the child re-apply VFLAGS on
+	// top duplicates its values (e.g. `-cflags x` becomes `x x`), so the child
+	// derives a DIFFERENT vcache key than the one the parent looks up, and the
+	// freshly built module object is never found ('could not rebuild cache
+	// module ... does not exist yet' — hit by any -usecache build running
+	// under a VFLAGS environment, e.g. the sanitized CI jobs).
+	old_vflags := os.getenv('VFLAGS')
+	if old_vflags != '' {
+		os.setenv('VFLAGS', '', true)
+	}
 	// The exit status is the caller's problem now: ignoring it while
 	// the hashes were saved eagerly permanently poisoned the module cache.
-	return os.system(rebuild_cmd)
+	rc := os.system(rebuild_cmd)
+	if old_vflags != '' {
+		os.setenv('VFLAGS', old_vflags, true)
+	}
+	return rc
 }
 
 // embed_file_content_hash returns the invalidation hash for one embedded-asset
--- a/vlib/v/builder/rebuilding.v
+++ b/vlib/v/builder/rebuilding.v
@@ -227,7 +227,15 @@ fn (mut b Builder) handle_usecache(vexe string) {
 	builtin_obj_path := b.rebuild_cached_module(vexe, 'vlib/builtin')
 	libs << builtin_obj_path
 	for ast_file in b.parsed_files {
-		if b.pref.is_test && ast_file.mod.name != 'main' {
+		// Cache the test's own non-main modules. Apply the same guards as the import
+		// loop below: builtin (and its parts: strconv/strings/math.bits/...) are already
+		// inside builtin.o, and a module already queued must not be built again — without
+		// these guards a test build caches builtin a SECOND time via its absolute path
+		// (a086..module.builtin.o AND 7c16..module.<abspath>.builtin.o), linking both and
+		// duplicating every builtin symbol (~9000 "duplicate symbol" errors at link).
+		if b.pref.is_test && ast_file.mod.name != 'main' && ast_file.mod.name != 'help'
+			&& !util.module_is_builtin(ast_file.mod.name) && ast_file.mod.name !in built_modules
+			&& !util.should_bundle_module(ast_file.mod.name) {
 			imp_path := b.find_module_path(ast_file.mod.name, ast_file.path) or {
 				verror('cannot import module "${ast_file.mod.name}" (not found)')
 				break
--- a/vlib/v/builder/rebuilding.v
+++ b/vlib/v/builder/rebuilding.v
@@ -27,19 +27,54 @@ pub fn (mut b Builder) rebuild_modules() {
 	$if trace_invalidations ? {
 		eprintln('> rebuild_modules all_files: ${all_files}')
 	}
-	invalidations := b.find_invalidated_modules_by_files(all_files)
+	invalidations, snew_hashes := b.find_invalidated_modules_by_files(all_files)
 	$if trace_invalidations ? {
 		eprintln('> rebuild_modules invalidations: ${invalidations}')
 	}
+	mut stale_object_might_survive := false
 	if invalidations.len > 0 {
 		vexe := pref.vexe_path()
 		for imp in invalidations {
-			b.v_build_module(vexe, imp)
+			rc := b.v_build_module(vexe, imp)
+			if rc != 0 {
+				// CACHE-POISONING GUARD: a failed module rebuild used to be
+				// silently ignored while the new source hashes were ALREADY saved —
+				// every later build then saw "hashes unchanged", skipped invalidation,
+				// and linked the STALE cached object forever (until a manual cache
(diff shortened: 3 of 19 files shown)


medvednikov:
V1 is going to be removed this month, so it's better to implement such things in V3.

medvednikov:
Well it'll still stay in oldv/ for 1 year, and will be available with -oldv. But it's no longer going to be developed.

Jengro777:
> V1 is going to be removed this month, so it's better to implement such things in V3.

How do I use v3 now

eptx:
Added one commit (`8184f8cf52`): **cgen: no cross-TU tentative definitions under -usecache/build_module**.

While validating downstream we hit another instance of the same class this PR addresses, worth fixing here too: cached module objects emit *tentative* definitions (common symbols) for interface `*_name_table`s, bundled-module globals (`g_closure`), and `g_main_argc`/`g_main_argv` — relying on the linker merging commons across objects. Current Apple ld and lld reject that (`N duplicate symbols` instead of merging), which can break every `-usecache` link on macOS depending on how many cached objects carry the tables. Cached objects now emit `extern` declarations; the program TU keeps the single initialized definition (and runs bundled-module global initializers under `-usecache`).

Validation on the branch: the multi-module regression test from this PR, the #27381 test, and `vlib/v/vcache/` all pass; `v fmt -verify` clean. `nm` on cached objects shows `U` where it previously showed `(__DATA,__common)`.

JalonSolov:
> > V1 is going to be removed this month, so it's better to implement such things in V3.
> 
> How do I use v3 now

`v vlib/v3`
`vlib/v3/v3 <rest of normal V command line>`

eptx:
Pushed one more `-usecache` correctness commit to this branch: 2ad1a2a9f1 — deterministic `$embed_file` resolution + embedded-asset cache invalidation.

**Symptom class:** with `-usecache`, a cached module object can silently carry *different embedded content* than the program TU, and editing only the embedded asset (no `.v` change) keeps serving the stale cached object forever.

**Root causes (two, related):**
1. The checker resolved a relative `$embed_file` path against the process CWD *first*, using the documented source-relative location ("Paths can be absolute or relative to the source file", `doc/docs.md`) only as a fallback. The `v build-module` child runs `chdir`-ed to VROOT (`Builder.v_build_module`), so when a CWD-relative twin of the asset exists under VROOT's parent tree, the child bakes that twin into the cached `.o` while the main TU embeds the intended file. Additionally, the fallback's "source file" anchor was `Checker.file`, which is the *importing* file's context when a module const initializer is checked from a use site — not the declaring file. The parser now records the declaring file on `ast.EmbeddedFile.source_file` (same pattern as `HashStmt.source_file`) and the checker resolves source-relative first against it, keeping CWD-relative resolution as a backwards-compatibility fallback. Resolution is now CWD-independent, so parent and build-module child always agree.
2. Embedded assets are invisible to the `.v` source hashes, so nothing ever invalidated a cached module object when only the asset changed. Each build-module object now records a `.embeds.txt` manifest (content hash + absolute path per resolved `$embed_file` target, via the new `ast.File.embedded_apaths`), and `Builder.rebuild_cached_module` verifies it before serving a cache hit; a failed rebuild removes the stale object instead of serving it.

**Regression test:** `vlib/v/tests/usecache_embed_file_test.v` — builds a module whose embed has a CWD-relative decoy twin while the compiler runs from a foreign CWD (resolution), then edits only the asset under a warm cache (invalidation). Verified: fails on the branch without this commit (built at 8184f8cf52), passes with it; `vlib/v/tests/usecache_multi_module_test.v`, `vlib/v/embed_file/` (6/6) and `vlib/v/vcache/` stay green (macOS arm64, clang).

Found in the wild: a git-worktree checkout sharing one V binary with its main checkout — `$embed_file('../../asset')` in the worktree's module cached the *main checkout's* copy of the asset, because the build-module child's VROOT-relative CWD probe hit the other checkout's file.

eptx:
Pushed `6bbd24692c` fixing the three CI regressions this branch introduced/exposed (triaged every red job):

1. **`link_generated_c_files_test`** (all jobs) — real regression from `8184f8cf52`: the program-owned `g_main_argc`/`g_main_argv` definition leaked into `-is_o` outputs. Each `-is_o` output is a standalone hostable TU and several are expected to be linked together, so these now keep their object-local (`static`) linkage there, exactly as on master.

2. **`v_parser_test`** (tcc-linux) — the bundled-module `should_init` clause in `8184f8cf52` was dangling: comment lines between `||` continuations terminated the expression, so the final `|| (...)` parsed as a discarded statement (and the clause was semantically inert). Joined properly.

3. **`usecache_multi_module_test` / `usecache_embed_file_test`** (sanitize jobs only) — these fail on **vanilla master** too, the new tests just exposed it: with a `VFLAGS` environment carrying `-cflags …` (as the sanitized jobs set), the spawned `v build-module` child re-applies VFLAGS on top of the relayed `build_options`, duplicating the cflags value — the child then derives a *different* vcache key than the parent looks up, and every `-usecache` build panics `could not rebuild cache module … does not exist yet`. Repro on master: `VFLAGS='-cc clang -cflags -fno-omit-frame-pointer' v -usecache run hello.v`. Fixed by clearing VFLAGS around the child spawn (the parent has already folded VFLAGS into the relayed options).

The remaining red I could find is master-side, not from this branch: `db/pg/pg_config_test`, `x/executor/admission_test`, the mbedtls concurrent-response flake, and `sokol windows_backend_flags_test` (already fixed on master by the recent "skip nested cross-compile" commit).

Validated locally: `link_generated_c_files_test`, `v_parser_test`, both usecache tests (plain and under the sanitize jobs' VFLAGS), `vcache`, `embed_file` 6/6, `fmt -verify` + `vet` clean.

medvednikov:
@codex review

PR #27659, opened 2026-07-04
cgen: fix heap struct init in if expression

### LINKED ISSUE #27601: cgen: heap struct literal (`&T{}`) in an `if`-expr condition is hoisted to file scope (`initializer element is not constant`)
### Description

A heap struct literal (`&T{...}`) used inside the **condition of an `if`-expression** is hoisted into a **file-scope/global** temporary whose initializer is a `v_malloc(...)` call. A C global initializer must be a compile-time constant, so the generated C fails to compile.

**V version:** V 0.5.1, current `master` (`05bea2aaa`)
**OS/Backend:** reproduced on macOS arm64; originally reported on Linux/cc.

### Reproducer

```v
struct Obj {
	a int
}

fn main() {
	x := if (&Obj{a: 1}).a > 0 { 10 } else { 20 }
	println(x)
}
```

Compile with:

```
v file.v
```

### Current behavior

```
error: initializer element is not a compile-time constant
  main__Obj* _t1 = (main__Obj*)builtin___v_malloc(sizeof(main__Obj) == 0 ? 1 : sizeof(main__Obj));
```

The `&Obj{...}` temp (`main__Obj* _t1 = ...v_malloc(...)`) is emitted **outside any function** (at file scope), so its non-constant `malloc` initializer is rejected. With more complex programs the same mis-hoisting also corrupts the generated `main` (surfacing as `'main' undeclared here (not in a function)`).

### Expected behavior

The temporary for the `&T{}` literal in the `if`-condition should be emitted **inside** the enclosing function (here `main`), not at file scope.

### Notes

- Reported via the bugs.vlang.io crash reporter (a fuzzer, `scan_nested_ifexpr_cond_branch`); this single root accounts for ~73 reports across `initializer element is not constant` (57) and `'main' undeclared here` (16).
- Related to the previously-fixed #27329 (`&T{}` heap-literal init emitted at file scope) — the `if`-condition trigger appears not to be covered by that fix.

> [!NOTE]
>

medvednikov:
## Summary
- force if expressions that contain address-of struct literals out of the C ternary path
- keep direct heap struct-init setup statements inside the current function scope
- add a cgen regression for #27601

Fixes #27601.

## Testing
- ./vnew -e 'struct Obj { a int } fn main() { x := if (&Obj{a: 1}).a > 0 { 10 } else { 20 } println(x) }'
- VTEST_ONLY=heap_struct_init_if_expr_cond ./vnew -silent vlib/v/gen/c/coutput_test.v
- ./vnew -silent vlib/v/compiler_errors_test.v


medvednikov:
Found one code issue, ignoring CI as requested.

## Finding

**P1: `expr_contains_addr_of_struct_init()` is not a complete AST walk, so the same hoist bug can still survive in nested/less common expression shapes.**

The new helper is used to decide whether an `if` expression must avoid the C ternary path. That is the right direction, but the traversal misses child expressions that can still contain `&Struct{}`. The most suspicious case is `ast.IfExpr`: the helper only checks `expr.left`, but `IfExpr.left` is the assignment-left side field, not the condition/body content; the actual conditions and branch statements are in `branches`.

This can matter when an inner `if` expression is embedded under an expression kind whose existing `need_tmp_var_in_expr()` does not recursively inspect all children. For example, `ConcatExpr` in `need_tmp_var_in_expr()` only checks option/result-returning calls, while the new detector walks concat values but then misses nested `IfExpr` branches because of the `ast.IfExpr { return ...expr.left }` case.

A likely reproducer shape:

```v
struct Obj {
	a int
}

fn main() {
	s := if true {
		'a' + (if (&Obj{ a: 1 }).a > 0 { 'b' } else { 'c' })
	} else {
		'd'
	}
	println(s)
}
```

I would change the `ast.IfExpr` case in `expr_contains_addr_of_struct_init()` to walk every branch condition and branch expression statement, or better, centralize this as a generic “expression emits pre-statements” predicate so it is not a second partial AST traversal.

There are a few other child fields that look omitted and worth auditing at the same time: `CastExpr.arg`, `ArrayInit.update_expr`, `MapInit.update_expr`, `ChanInit.cap_expr`, `RangeExpr.low/high`, `Likely.expr`, `DumpExpr.expr`, and `UnsafeExpr.expr`. Several of these are explicit expression children in the AST but are not currently visited by the new detector.

## Summary

The main fix concept is sound, and the added regression covers the direct issue from #27601. I would not merge as-is until the new detector is made exhaustive enough, because this class of bug is exactly about hidden pre-statement-emitting expressions inside ternary candidates.

PR #27660, opened 2026-07-05
markused: fix undeclared C identifier for alias array type

### LINKED ISSUE #27648: C code gen: undeclared alias type for a dynamic array when using a reference to it
### Describe the bug

When referencing a dynamic array instantiated via a type alias, V compiler generates invalid C code.

### Reproduction Steps

```v
type Array = []int

fn main() {
	a := Array{}
	n := &a
}
```

### Expected Behavior

success

### Current Behavior

```c
warning: tcc compilation failed, falling back to cc (this is much slower)
cc: /tmp/v_501/cgen_undeclared.01KWMK7Q88Y1RNVMBEAGJEA459.tmp.c:4112: error: 'main__Array' undeclared
================== C compiler bug report ==============
Generated C lines sent from /tmp/v_501/cgen_undeclared.01KWMK7Q88Y1RNVMBEAGJEA459.tmp.c:6910:
    6905 | 
    6906 | #line 3 "/cgen_undeclared.v"
    6907 | VV_LOC void main__main(void) {
    6908 | 
    6909 | #line 4 "/cgen_undeclared.v"
>   6910 |      main__Array *a = HEAP(main__Array, (builtin____new_array_with_default_noscan(0, 0, sizeof(int), 0)));
    6911 | 
    6912 | #line 5 "/cgen_undeclared.v"
    6913 |      int* n = &(*(int*)builtin__array_get((*(a)), 0));
    6914 | }
    6915 | 
Corresponding V lines sent from /cgen_undeclared.v:4:
       1 | type Array = []int
       2 | 
       3 | fn main() {
>      4 |      a := Array{}
       5 |      n := &a[0]
       6 | }
=======================================================
```

### Possible Solution

_No response_

_No response_

### V version

V 0.5.1 9dce771

not relevant
> [!NOTE]
>

rilaaax:
_NB: all credits to Claude_

## What

Fixes #27648

C compilation error ("undeclared identifier") when an array type alias is used to declare a variable whose address (or an element's address) is taken later, and the alias isn't referenced anywhere else in the program.

## Why

`-skip-unused` is enabled by default, so cgen only emits a C typedef for an alias type if `markused` recorded it as used. For `a := Array{}`, the RHS `ast.ArrayInit` node's `.typ` is already resolved to the underlying `[]int` type by the checker, not to the `Array` alias — so `markused` never sees the alias and its typedef gets stripped by dead-code elimination. cgen then emits the auto-heap declaration using the *declared* type (the alias), which was never typedef'd, hence the undeclared identifier.

## Fix

`ast.AssignStmt.left_types` holds the declared type(s) of the LHS. Marking those as used keeps `markused` and cgen in sync.

## Test

Regression test added. Note: the alias must not appear anywhere else in the test file (a stray `==` comparison, `<<`, etc. would mark it used through a different path and silently defeat the test).

### DIFF
--- a/vlib/v/markused/walker.v
+++ b/vlib/v/markused/walker.v
@@ -726,6 +726,9 @@ pub fn (mut w Walker) stmt(node_ ast.Stmt) {
 			}
 		}
 		ast.AssignStmt {
+			for t in node.left_types {
+				w.mark_by_type(t)
+			}
 			w.exprs(node.left)
 			w.exprs(node.right)
 		}
--- a/vlib/v/tests/aliases/alias_array_reference_test.v
+++ b/vlib/v/tests/aliases/alias_array_reference_test.v
@@ -1,0 +1,8 @@
+type Arr = []int
+
+fn test_alias_array_reference() {
+	mut a := Arr{len: 1}
+	a[0] = 69
+	p := &a[0]
+	assert *p == 69
+}


medvednikov:
Reviewed vlang/v#27660. Verdict: LGTM / approve — no blocking findings.

The change is small and targeted: markused now marks every AssignStmt.left_types entry before walking the assignment expressions. That is the right place to fix this, because mark_by_type() records the exact type and, for aliases, also walks through to the parent type, so the alias typedef is retained instead of only the underlying array type.

The root cause also checks out. Alias array literals are represented with the concrete parent array as typ and the alias separately as alias_type; the checker returns alias_type as the array-init result when present, and assignment checking stores the computed LHS type in left_types. The old markused ArrayInit path marked elem_type and node.typ, but not alias_type, so it could miss the alias even though cgen later used the assignment type to emit the declaration.

That matches the failing codegen path: cgen takes node.left_types[i] as var_type, derives styp, and for auto-heap declarations emits HEAP(${styp}, ...); if styp is the alias C name and the typedef was skipped, the generated C fails exactly as reported.

The regression test is well chosen: the alias appears only in Arr{len: 1}, then &a[0] forces the auto-heap/address-taking path while the assertion verifies the generated program still behaves correctly.

PR #27661, opened 2026-07-05
checker: reject `&` reference of fixed array alias (fix #27646)

### LINKED ISSUE #27646: C code gen: invalid zero-init syntax ({0}) when a fixed array alias is forced to the heap
### Describe the bug

When referencing a fixed array instantiated via a type alias, V compiler generates invalid C code for the heap initialization.

### Reproduction Steps

```v
type Array = [8]int

fn main() {
	a := Array{}
	n := &a[0]
}
```

### Expected Behavior

success

### Current Behavior

```c
warning: tcc compilation failed, falling back to cc (this is much slower)
cc: /tmp/v_501/cgen_invalid_heap_init.01KWMJV853PHB3C85N69Z3J0M6.tmp.c:4117: error: ';' expected (got '}')
================== C compiler bug report ==============
Generated C lines sent from /tmp/v_501/cgen_invalid_heap_init.01KWMJV853PHB3C85N69Z3J0M6.tmp.c:6915:
    6910 | 
    6911 | #line 3 "/cgen_invalid_heap_init.v"
    6912 | VV_LOC void main__main(void) {
    6913 | 
    6914 | #line 4 "/cgen_invalid_heap_init.v"
>   6915 |      Array_fixed_int_8 *a = HEAP(Array_fixed_int_8, ({0}));
    6916 | 
    6917 | #line 5 "/cgen_invalid_heap_init.v"
    6918 |      Array_fixed_int_8 n = &(*(a));
    6919 | }
    6920 | 
Corresponding V lines sent from /cgen_invalid_heap_init.v:4:
       1 | type Array = [8]int
       2 | 
       3 | fn main() {
>      4 |      a := Array{}
       5 |      n := &a
       6 | }
=======================================================
```

### Possible Solution

_No response_

_No response_

### V version

V 0.5.1 9dce771

not relevant
> [!NOTE]
>

medvednikov:
### Problem

```v
type Array = [8]int

fn main() {
	a := Array{}
	n := &a[0]
}
```

generated invalid C and crashed the C compiler:

```c
Array_fixed_int_8 *a = HEAP(Array_fixed_int_8, ({0}));
//                                              ^^^^^ error: ';' expected
```

A plain `[8]int` already errors with *"cannot reference fixed array `a` outside `unsafe` blocks as it is supposed to be stored on stack"*, because fixed arrays live on the stack and cannot be referenced/heap-promoted. The alias form slipped past that guard.

### Cause

`Checker.mark_as_referenced` detected the fixed array via the *direct* type symbol kind. For a `type Arr = [N]T` alias the direct symbol kind is `.alias`, not `.array_fixed`, so the guard was skipped and the variable was wrongly marked `is_auto_heap`. cgen then emitted `HEAP(Arr, ({0}))` — but a braced initializer cannot be parenthesized, and the `HEAP` macro (`(type[]){expr}[0]`) fundamentally cannot allocate array types.

### Fix

Resolve the alias to its final symbol so a fixed array alias (including nested aliases) hits the same guard and produces the same error as a plain fixed array. Referencing inside `unsafe {}` still works, and dynamic array aliases (`type Arr = []int`) are unaffected.

### Tests

Added `vlib/v/checker/tests/reference_fixed_array_alias_err.{vv,out}`.

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -7753,6 +7753,11 @@ fn (mut c Checker) mark_as_referenced(mut node ast.Expr, as_interface bool) {
 					return
 				}
 				type_sym := c.table.sym(obj.typ.set_nr_muls(0))
+				// Resolve aliases so that a `type Arr = [N]T` is treated like its
+				// underlying fixed array: fixed arrays live on the stack, so they
+				// cannot be referenced/heap-promoted. Without this, the alias slips
+				// past the guard below and cgen emits invalid `HEAP(Arr, ({0}))`.
+				final_kind := c.table.final_sym(obj.typ.set_nr_muls(0)).kind
 				if obj.is_stack_obj && !type_sym.is_heap() && !type_sym.is_int()
 					&& !c.pref.translated && !c.file.is_translated {
 					suggestion := if type_sym.kind == .struct {
@@ -7763,7 +7768,7 @@ fn (mut c Checker) mark_as_referenced(mut node ast.Expr, as_interface bool) {
 					mischief := if as_interface { 'used as interface object' } else { 'referenced' }
 					c.error('`${node.name}` cannot be ${mischief} outside `unsafe` blocks as it might be stored on stack. Consider ${suggestion}.',
 						node.pos)
-				} else if type_sym.kind == .array_fixed {
+				} else if type_sym.kind == .array_fixed || final_kind == .array_fixed {
 					c.error('cannot reference fixed array `${node.name}` outside `unsafe` blocks as it is supposed to be stored on stack',
 						node.pos)
 				} else {
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -7756,17 +7756,19 @@ fn (mut c Checker) mark_as_referenced(mut node ast.Expr, as_interface bool) {
 				// Resolve aliases so that a `type Arr = [N]T` is treated like its
 				// underlying fixed array: fixed arrays live on the stack, so they
 				// cannot be referenced/heap-promoted. Without this, the alias slips
-				// past the guard below and cgen emits invalid `HEAP(Arr, ({0}))`.
-				// Use `fully_unaliased_type` (not `final_sym`, which discards
-				// `nr_muls`/flags) and require no pointer indirection, so a
-				// genuine pointer alias like `type Ptr = &Arr` (the pointer level
-				// lives on the alias parent, which `fully_unaliased_type`
-				// preserves) is not misread as a bare fixed array value.
-				// `set_nr_muls(0)` intentionally strips a *top-level* pointer
-				// first: for a fixed array the only way a local ends up with a
-				// top-level `&Arr` type is the auto-deref form `p := &arr`, where
-				// that pointer is transparent and `p` behaves as the array value,
-				// so it must still be rejected.
+				// past the guard below and cgen emits invalid `HEAP(Arr, ({0}))`
+				// (#27646). Use `fully_unaliased_type` (not `final_sym`, which
+				// discards `nr_muls`/flags).
+				//
+				// The only usable pointer to a fixed array is a *named* pointer
+				// alias (`type Ptr = &Arr`): V represents it as a real pointer, so
+				// `&p` is valid there. A raw top-level `&Arr` (from `&arr`, or a
+				// `fn () &Arr` return) is not a usable pointer - cgen decays it to
+				// the fixed array value - so it must be rejected like the array
+				// itself. `set_nr_muls(0)` strips the raw top-level pointer (so a
+				// raw `&Arr` is treated as `Arr` and rejected), while a pointer
+				// carried on the alias parent survives `fully_unaliased_type` and
+				// keeps `nr_muls() > 0` (so a named `Ptr` is accepted).
 				unaliased_typ := c.table.fully_unaliased_type(obj.typ.set_nr_muls(0))
 				is_fixed_array := unaliased_typ.nr_muls() == 0
 					&& c.table.sym(unaliased_typ).kind == .array_fixed
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -7759,8 +7759,14 @@ fn (mut c Checker) mark_as_referenced(mut node ast.Expr, as_interface bool) {
 				// past the guard below and cgen emits invalid `HEAP(Arr, ({0}))`.
 				// Use `fully_unaliased_type` (not `final_sym`, which discards
 				// `nr_muls`/flags) and require no pointer indirection, so a
-				// pointer alias like `type Ptr = &Arr` is not misread as a bare
-				// fixed array value.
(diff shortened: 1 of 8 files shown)


medvednikov:
Findings:

* **Potential false positive for aliases whose parent type is a pointer/flagged fixed array.** The new check uses `c.table.final_sym(obj.typ.set_nr_muls(0)).kind` and rejects when that final kind is `.array_fixed`. `final_sym()` follows alias parents by type index only, so it does not preserve `nr_muls` or flags stored on an alias’s parent type. For example, an alias like `type P = &[8]int` can resolve to the fixed-array symbol even though the variable itself is a pointer alias, not a fixed array value. That would make address-taking of a `P` variable hit the new “cannot reference fixed array” error. The table already has `fully_unaliased_type()`, which preserves parent pointer indirection by adding the alias parent’s `nr_muls` to the alias type before returning. A safer guard would likely check the fully unaliased type and only reject when the unaliased type has no pointer indirection and its symbol is `.array_fixed`.

Everything else in the PR looks aligned with the stated fix: the changed branch prevents the alias from falling through to the auto-heap promotion path, which is where `obj.is_auto_heap` would otherwise be set for alias kinds.

medvednikov:
Addressed the review finding about pointer/flag aliases.

The first cut used `final_sym(obj.typ).kind`, which follows the alias chain by type index only and drops `nr_muls`/flags. A pointer alias like `type Ptr = &[8]int` (or a chain `type PtrFA = &FA` where `type FA = [8]int`) therefore resolved to the fixed-array symbol, so `&p` for a `Ptr`/`PtrFA` variable was wrongly rejected with *"cannot reference fixed array"*.

Fixed by resolving with `fully_unaliased_type` (which preserves the parent pointer indirection and flags) and only treating the variable as a fixed array when the unaliased type has `nr_muls() == 0`:

```v
unaliased_typ := c.table.fully_unaliased_type(obj.typ.set_nr_muls(0))
is_fixed_array := unaliased_typ.nr_muls() == 0
    && c.table.sym(unaliased_typ).kind == .array_fixed
```

This subsumes the plain fixed-array case as well, so the special-casing of `type_sym.kind == .array_fixed` is no longer needed.

Added `vlib/v/tests/aliases/reference_fixed_array_pointer_alias_test.v` as a regression guard (it fails with the old `final_sym`-based check and passes now). Verified the original error test still matches and the full `compiler_errors_test.v` suite passes.

medvednikov:
Thanks — I dug into the direct `&FA` path and it turns out `set_nr_muls(0)` is actually the correct operation here, so I kept it (with a clearer comment) and added a regression test for that path.

The key point is that `p := unsafe { &arr }` for a **fixed array** does not give `p` a plain `&FA` pointer — V makes it an **auto-deref var**, so `p` behaves as the array value. You can see this in the generated C, and it's already broken on its own, independent of `&p`:

```v
type FA = [8]int
fn main() {
    mut arr := FA{}
    p := unsafe { &arr }
    unsafe { println(p[0]) }
}
```
```c
// cgen output — invalid regardless of taking &p:
Array_fixed_int_8 p = &arr;          // error: array initializer must be an initializer list
builtin__println(builtin__int_str((*p)[0]));
```

So the auto-deref case can never produce a working program. Resolving the guard from `obj.typ` (without `set_nr_muls(0)`) makes `&p` fall through to the auto-heap path and emit `HEAP(Array_fixed_int_8, (&arr))` — i.e. it replaces the clean *"cannot reference fixed array"* checker error with a C compiler error, and the suggested `assert q != unsafe { nil }` test would not actually pass.

The distinction that matters is auto-deref vs. genuine pointer, and `set_nr_muls(0)` captures it precisely:

- **Auto-deref local** (`p := &arr`): the pointer is a transparent top-level level → `set_nr_muls(0)` strips it → treated as the fixed array → **rejected** (clean error). ✔
- **Genuine pointer alias** (`type Ptr = &FA`): the pointer lives on the alias *parent*, which `fully_unaliased_type` preserves (`set_nr_muls(0)` on the alias itself is a no-op) → `nr_muls() == 1` → **accepted**. ✔

Added `vlib/v/checker/tests/reference_fixed_array_alias_autoderef_err.{vv,out}` to cover the direct `&FA` path you flagged (it asserts the clean checker error). The pointer-alias positive test and the full `compiler_errors_test.v` + `vlib/v/tests/aliases/` suites still pass.

Separately, the `Array_fixed_int_8 p = &arr;` codegen for `p := &fixed_array` is a genuine pre-existing cgen bug (auto-deref of a pointer-to-fixed-array), but it's orthogonal to #27646 — happy to file it as its own issue.

medvednikov:
Good catch that the `fn () &FA` return form hits the same branch — I instrumented the checker to get ground truth, and it turns out the two forms are indistinguishable and both are already broken independent of `&p`, so the clean rejection is the right call.

**The two `&FA` sources are byte-identical at the reference site.** Instrumenting `mark_as_referenced`:

```
p := unsafe { &arr }   ->  obj.typ = &main.FA, nr_muls=1, is_auto_deref=false
p := get_ptr() (&FA)   ->  obj.typ = &main.FA, nr_muls=1, is_auto_deref=false
```

(My earlier "auto-deref" wording was wrong — `is_auto_deref` is false; I've corrected the comment.) Because both present as the same `&main.FA` type, no check can accept one and reject the other. So the reviewer's suggested "use `fully_unaliased_type(obj.typ)` + a separate check to reject only `p := unsafe { &arr }`" can't actually separate them — it would either reject both (what we already do) or accept both.

**And accepting them means shipping broken code, not a working pointer.** `fn () &FA` doesn't produce a usable pointer — V decays the `&FA` return to a fixed-array value wrapper (`_v_Array_fixed_int_8`), so it fails to compile *without any `&p`*:

```v
type FA = [8]int
fn get_ptr() &FA { mut a := FA{}; return unsafe { &a } }
fn main() {
    p := get_ptr()
    unsafe { println(p[0]) }   // cgen: (*p)[0] -> "subscripted value is not an array"
}
```

Same for `p := unsafe { &arr }` (`Array_fixed_int_8 p = &arr;`). Both are broken on master too — this PR doesn't touch that cgen path — so rejecting `&p` on them is not a regression; it just turns a confusing C error into a clean checker error.

**The one form that genuinely works is the named pointer alias**, which stays accepted (pointer lives on the alias parent, preserved by `fully_unaliased_type`):

```v
type FA = [8]int
type PtrFA = &FA
fn get_ref(p PtrFA) bool { q := &p; return q != unsafe { nil } }  // works, returns true
```

So `set_nr_muls(0)` is doing exactly the right split: raw top-level `&FA` (not a usable pointer) → treated as the array → rejected; pointer carried on a named alias parent → accepted.

Changes in `ae69019e`:
- Corrected the comment to describe the real rule (named-alias pointer vs. raw `&FA`), not auto-deref.
- Replaced the auto-deref-only test with `reference_fixed_array_alias_raw_pointer_err.{vv,out}` covering **both** raw `&FA` sources you raised (`&arr` local and `fn () &FA` return).

Full `compiler_errors_test.v` and `vlib/v/tests/aliases/` suites pass. (Separately, the `&FA`-return/`&arr`-local cgen decay is a genuine pre-existing bug I can file on its own.)

PR #27663, opened 2026-07-05
transform: fix O(n²) late-used-fn-bodies pass (7-10x faster transform)

medvednikov:
## Problem

Profiling the v3 transform phase showed it dominated by a single quadratic pass, `transform_late_used_fn_bodies` in `vlib/v3/transform/transform.v`. Its worklist loop rescanned the **entire** node array `[0, limit)` on every round:

```v
for pending.len > 0 {
    _ := pending.pop()
    transform_pending_late_used_fn_bodies(limit, ...)  // walks ALL nodes from 0 each round
}
```

Every round re-classified each node (millions of `node_kind_id` calls) and re-ran the expensive `fn_decl_has_unresolved_generics` check (→ `type_text_has_generic_placeholder`) on every non-matching `fn_decl`. With ~pending rounds over ~N nodes this is `O(pending × nodes)`.

Profiled call counts on a single input (`markused.v`):

| function | calls |
|---|---|
| `node_kind_id` | 25.7M |
| `type_text_has_generic_placeholder` | 1.79M |
| `fn_decl_has_unresolved_generics` | 373k |

## Fix

Build the `fn_decl` candidate list **once**. The scan range `[0, limit)` is fixed — `transform_fn_body` only appends nodes beyond `limit`, and those generated nodes are already handled inline — so the set of candidate fn_decls, their `(file, module)` context, and their generic-ness never change across rounds. Each round now iterates a compact, pre-filtered list and does only the cheap `late_used_fn_matches` map check that genuinely depends on the growing `late` set.

The round structure and processing order are preserved exactly, so the pass is behavior-preserving.

## Results

| Input | transform before → after | total before → after |
|---|---|---|
| `markused.v` | 2320 ms → **294 ms** (7.9×) | 3450 ms → **843 ms** (4.1×) |
| `v3.v` self-compile | 55033 ms → **5564 ms** (9.9×) | 58015 ms → **7579 ms** (7.7×) |

The win scales with input size, as expected for an O(N²)→O(N) fix.

## Correctness

- **Byte-identical generated C output** on `markused.v` (552 KB) and on the full `v3.v` self-compile (5.67 MB — exercises generics and every module).
- **v3 test suite: no regressions.** The failing tests in this tree fail identically with and without this change (verified by stashing the change, rebuilding, and re-running the same set); they are pre-existing and unrelated.

### DIFF
--- a/vlib/v3/transform/transform.v
+++ b/vlib/v3/transform/transform.v
@@ -1331,12 +1331,14 @@ fn (mut t Transformer) transform_late_used_fn_bodies(names []string, node_limit 
 	t.cur_file = old_file
 	limit := if node_limit < t.a.nodes.len { node_limit } else { t.a.nodes.len }
 	// Build the fn_decl candidate list once. The scan range [0, limit) is fixed:
-	// transform_fn_body only appends nodes beyond `limit` (those generated nodes
-	// are handled inline below), so the set of candidate fn_decls, their
-	// (file, module) context, and their generic-ness never change across rounds.
-	// Re-deriving all of this every round previously made this pass O(pending *
-	// nodes): a full node walk (millions of node_kind_id calls) plus a repeated
-	// fn_decl_has_unresolved_generics check on every non-matching fn_decl.
+	// transform_fn_body rewrites the current fn_decl in place and appends new
+	// nodes beyond `limit` (those generated nodes are handled inline below); it
+	// never inserts or moves existing top-level nodes below `limit`. So the set
+	// of candidate fn_decls, their (file, module) context, and their generic-ness
+	// never change across rounds. Re-deriving all of this every round previously
+	// made this pass O(pending * nodes): a full node walk (millions of
+	// node_kind_id calls) plus a repeated fn_decl_has_unresolved_generics check
+	// on every non-matching fn_decl.
 	mut candidates := []LateFnCandidate{}
 	mut scan_file := ''
 	mut scan_module := ''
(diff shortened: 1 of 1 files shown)


medvednikov:
## Findings

No blocking findings.

The change looks behavior-preserving from the code paths I inspected. The old implementation re-scanned `[0, limit)` once per pending work item, recomputing file/module context and `fn_decl_has_unresolved_generics` each round before checking `late_used_fn_matches`. The new implementation precomputes only the static candidate set and keeps the dynamic `late_used_fn_matches(late, node, module)` check inside each worklist scan, so newly discovered late-used names are still evaluated against every unprocessed candidate.

The fixed `limit` assumption also matches the prior behavior: the old code never scanned appended nodes beyond `limit` either; it only handled generated functions through the explicit `node_count_before .. t.a.nodes.len` loop after transforming a matched function. The new code preserves that generated-node handling.

Caching candidate indices appears safe because `transform_fn_body` rewrites the current function node in place and appends new children/nodes, rather than inserting or moving existing top-level nodes before `limit`. The new late pass also reloads `node := t.a.nodes[idx]` before matching, so it is not holding stale `flat.Node` copies for unprocessed candidates.

Caching the “non-generic” predicate looks safe as well. `fn_decl_has_unresolved_generics` is based on the declaration’s generic params, value/type text, param type text, and known type maps; it does not appear to depend on the evolving `late` or `pending` sets.

## Non-blocking note

The comment says `transform_fn_body only appends nodes beyond limit`; technically it also overwrites the transformed function node in place. That does not affect correctness here, but wording like “does not insert or move nodes below `limit`; it rewrites the current fn_decl in place and appends new nodes” would be a little more precise.

Overall: looks good to merge from a code-review perspective.

PR #27664, opened 2026-07-05
v3: fix another batch of test failures

medvednikov:
## Summary

Moves another batch of tracked v3 test failures to passing and fixes the follow-on self-host issues uncovered by `test_all.vsh`. This bundles the three most recent unpushed v3 test-failure fix batches.

Highlights from the latest batch:
- C backend array / fixed-array equality pruning
- TCC-safe atomic pointer loads
- ARM64 self-host memory behavior
- Temp-path isolation in a long generic test

Changes are confined to `vlib/v3` (`gen/c`, `gen/arm64`, `ssa`, `transform`, `types`, `tests`, `test_all.vsh`, `v3.v`, and the `v3_test_{failures,passes}.txt` tracking files).

## Validation

- `./vnew -gc none -path "$(pwd)/vlib" vlib/v3/test_all.vsh` passed all 8 phases.
- Phase 1: `120/120` v3 unit tests passed.
- Phase 5: `33/35` C oracle examples passed (`2048` and `tetris` skipped — no display).
- Phase 7: self-host chain converged, `v5.c == v6.c`.
- Phase 8: language feature parity passed.


vlib/v3/gen/c/cleanc.v
```diff
@@ -1986,17 +1993,29 @@ fn (mut g FlatGen) emit_c_directives() {
...
 	mut emitted := false
+	mut emitted_includes := map[string]bool{}
 	directives := g.ordered_c_directives()
 	for i, directive in directives {
 		if !c_contains_preserved_system_include_directive(directive) {
 			continue
 		}
+		clean := directive.trim_space()
+		if c_is_preserved_system_include_directive(clean) {
+			if emitted_includes[clean] {
```
medvednikov: This dedupes preserved system includes by the raw include line before considering the lifted context. If the same header appears under different guards (for example one `#ifdef __linux__` block and one `#ifdef __APPLE__` block), the second include is skipped even though it needs a different context. The existing top-level include dedupe tracks directive depth, but this path applies to all preserved includes. Could the dedupe key include the lifted prefix/context, or only dedupe when the context is identical?

vlib/v3/v3.v
```diff
@@ -264,7 +301,11 @@ fn main() {
 		} else if args[i].starts_with('-') {
 			i++
 		} else {
-			input_file = args[i]
+			if should_run && input_file.len > 0 {
+				run_args << args[i]
```
medvednikov: With this parser, `run` only captures non-dash arguments after the input file. A command like `v3 run app.v --help` hits the earlier `args[i].starts_with('-')` branch and silently drops `--help` instead of forwarding it to the program. Once `should_run && input_file.len > 0` is true, this should probably append all remaining args (or honor a `--` separator) before interpreting more compiler flags.

medvednikov (commented):
Reviewed with CI intentionally skipped. I found two behavior issues that look PR-introduced and worth checking before merge.

vlib/v3/gen/c/cleanc.v
medvednikov: Fixed in c90eaa1fd. The dedupe now keys on the lifted guard context together with the include line (`prefix.join('\n') + '\x00' + clean`), so it only dedupes when the context is identical. The same header under different guards (e.g. `#ifdef __linux__` vs `#ifdef __APPLE__`) now emits both, each with its own context.

Repro before the fix showed the second, differently-guarded include was dropped *and* the first guard's `#endif` was left dangling; after the fix both blocks are emitted and balanced. Added a regression test (`test_same_system_include_under_different_guards_is_kept` in `posix_wait_header_codegen_test.v`).

vlib/v3/v3.v
medvednikov: Fixed in c90eaa1fd. Added a guard at the top of the arg loop: once `run` has captured its input file, every remaining argument is forwarded verbatim to `run_args` (including `-`-prefixed flags), instead of being interpreted as compiler flags. So `v3 run app.v --help` now forwards `--help` to the program.

Compiler flags placed before the input file still parse as before. Added a regression test (`test_run_forwards_all_args_after_input_file` in `test_file_cli_run_test.v`) covering `v3 run app.v --help pos --flag=1`.

PR #27669, opened 2026-07-06
cgen: fix generic method &Interface param extra & indirection (#27612)

### LINKED ISSUE #27612: Add test    `generic methods with &Interface parameters`
Suggest adding this test

```
// Test: calling concrete generic methods with &Interface parameters
// should not incorrectly add an extra & indirection to the argument.
// Regression test for the V compiler codegen bug in ref_or_deref_arg_ex.
//
// The pattern matches the original dbpool issue:
//   - A generic function returns (T, &Interface)
//   - Caller destructures, then passes &Interface var to a generic method
//   - Compiler must NOT add extra & to the already-&Interface arg

interface ICalculator {
	calculate(x int) int
}

struct Doubler {}

fn (d Doubler) calculate(x int) int {
	return x * 2
}

struct ComputeBox[T] {
	val int
}

fn (b ComputeBox[T]) compute(calc &ICalculator) int {
	return calc.calculate(b.val)
}

// Returns &Interface from a generic function, like DatabasePool.acquire_raw
fn make_calculator[T](val T) (T, &ICalculator) {
	calc := &ICalculator(Doubler{})
	return val, calc
}

fn test_generic_method_ref_interface_param() {
	box := ComputeBox[Doubler]{
		val: 21
	}
	_, ref := make_calculator[Doubler](Doubler{})
	result := box.compute(ref)
	assert result == 42
}

```
> [!NOTE]
>

Jengro777:
fixed: https://github.com/vlang/v/issues/27612

Generic method calls on concrete generic structs could incorrectly add an extra & to &Interface arguments. The checker auto-dereferences &Interface variables to their value type when typing call arguments, and the cgen's expected_type resolution left the .generic flag set. This caused ref_or_deref_arg_ex to miss that the argument was already a pointer and emit an extra &, resulting in &Interface** being passed instead of &Interface*.

Fix: in ref_or_deref_arg_ex, when about to emit & for an Ident referring to a Var, compare the variable's actual type (clearing .generic flag) against the expected type (also clearing .generic flag). If they match, emit the expression directly.

Add regression test using the existing Wrapper[T].extract() infrastructure.

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -7235,6 +7235,16 @@ fn (mut g Gen) ref_or_deref_arg_ex(arg ast.CallArg, expected_type_ ast.Type, lan
 				}
 			} else if effective_arg_sym.kind == expected_ref_inner_sym.kind
 				&& effective_arg_typ.idx() == expected_ref_inner_type.idx() {
+				// Fix: when a &Interface variable is auto-dereferenced by the checker to its
+				// value type, arg_typ loses the pointer flag. Check the actual variable type
+				// to avoid adding an extra & on an already-pointer variable.
+				if effective_arg_expr is ast.Ident && effective_arg_expr.obj is ast.Var {
+					var_typ := (effective_arg_expr.obj as ast.Var).typ
+					if var_typ.clear_flag(.generic) == expected_type.clear_flag(.generic) {
+						g.expr(effective_arg_expr)
+						return
+					}
+				}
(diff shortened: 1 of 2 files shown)


Jengro777:
Is it still necessary to fix it in V? Should we wait for v3 to default and fix it on v3

medvednikov:
It's ok, the test is added, so it will be implemented for V3.

Jengro777:
> It's ok, the test is added, so it will be implemented for V3.

Can I turn this off here

PR #27670, opened 2026-07-06
bitfield: add way to convert back into byte array

acook:
This allows a `bitfield` to be converted back into a `[]u8` losslessly.

Before this I wrote a version in my own code which steps through bit-wise, as that was the only way to do so with public functions. But this, accessing `BitField.field` directly and operating on 32-bits at a time, should be far more performant.

### DIFF
--- a/vlib/bitfield/bitfield.v
+++ b/vlib/bitfield/bitfield.v
@@ -53,6 +53,21 @@ pub fn (bf BitField) str() string {
 	return output.bytestr()
 }
 
+// bytes converts the bit array into an array of bytes.
+pub fn (bf BitField) bytes() []u8 {
+	mut output := []u8{}
+	mut i := 0
+	for f in bf.field {
+		for j in 1..5 {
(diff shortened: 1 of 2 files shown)


JalonSolov:
Need to run `v fmt -w` on the files.

medvednikov:
@codex review

medvednikov:
@acook 

The test coverage is too thin for a low-level bit-ordering API. One ASCII string verifies the happy path, but it does not cover edge cases like []u8{}, 0x00, 0xff, 0x01, 0x80, 0x0f, 0xf0, lengths crossing 4-byte slot boundaries, or non-byte-aligned bitfields created with from_str() / set_bit().
The implementation is correct-looking but unnecessarily fragile. The comment says “if slot_size changes this may break,” and bits.reverse_32(f) is recomputed four times per slot. That is not a correctness blocker today, but it is a smell in library code. A simpler version could precompute s := bits.reverse_32(f) once per slot, or extract each stored byte and apply bits.reverse_8().

acook:
Yep, the reversing should only happen once per field. Fixed.

PR #27674, opened 2026-07-06
v3: fix 58 more tracked tests: sum-type deep equality, option smartcast, missing-return analysis

medvednikov:
Moves 58 tests from `v3_test_failures.txt` to `v3_test_passes.txt` (12 encoding/xml, 17 options, 11 sumtype/match, 18 across arrays/generics/structs/interfaces/sync).

### Sum-type deep equality (the big one)
`a == b` on a sum type compiled to `tag == tag && memcmp(&a, &b, sizeof(Sum))`, which compares the boxed payload **pointers** — always false for distinct allocations. Sum equality is now lowered to a call to a synthesized `__v3_sum_eq_<Sum>(a, b)` helper that unboxes the active variant and compares it by value through the usual membership-eq machinery, so recursive types (an XML node tree) compare correctly at any depth. Call sites just record the sum type during the (possibly parallel) transform; the helpers are synthesized serially afterwards, under the module/file context of the requesting call site (unqualified type resolution needs it — building them under leftover context diverged only in v4, via map-iteration order). New Transformer maps are cloned in `fork_worker` and merged in `merge_worker_used_fns` (a plain struct copy shares map backing across worker threads).

### Sum-type str() now matches V1
`Sum(Foo{})` instead of `Sum(Foo(Foo{}))`; string payloads quoted, runes backticked, alias variants keep their `Ints([1, 2])` wrapper. Also fixed a temp-decl placement bug where the base temp of a non-trivial sum interpolation was emitted inside the first branch of its own if-chain (invalid C).

### Option smartcast
- `if x != none { ... }` unwraps `x` inside the then-branch — locals, struct fields, and chains; both reads and writes lower through `.value`.
- `if x == none { ... } else { ... }` (and `== nil` on `?&T`) unwraps in the else branch.
- Interface structs with `?&Iface` fields now emit properly typed `Optional_*` C fields (previously the generic `Optional`, which broke assignments).

### Missing-return analysis
`stmt_definitely_returns` now recognizes exhaustive sum-type matches, `lock`/`rlock` bodies that always return, infinite `for {}` loops without `break`, and calls to `@[noreturn]` functions — the parser now records that attribute into the flat AST (previously only a hardcoded `panic`/`exit` list counted).

### cgen
- Fully qualified enum values (`token.Kind.lsbr`) resolve to their integer value in expressions and const-array initializers (gated on capitalization to stay off hot selector paths).
- `chan.cap` / `chan.len` field access.

### Verification
- Self-host: `v -gc none -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` works; the chain v4→v5→v6 converges byte-identically.
- Stage timings and peak RSS at parity with master (interleaved min-of-N A/B against a clean-HEAD prod build on the same machine; run-to-run noise is ±30%, mins: transform 103ms vs 107ms, cgen 181ms vs 184ms, RSS 1454MB vs 1441MB).
- All 1558 previously-passing tests were differentially checked old-vs-new; the two real regressions this surfaced (alias formatting in sum str) were fixed before finalizing.
- The 58 flips were re-verified with the final `-prod` binary.


medvednikov:
Review findings, CI ignored:

1. **[P1] Sum-equality helper synthesis can introduce used operator functions after the last late-transform pass.**
   In the used-function-filtered path, `should_transform_fn` only transforms functions that are already considered used, and `transform_late_used_fn_bodies(...)` runs before `synthesize_sum_eq_helpers()`. The generated sum helper then builds payload comparisons through `make_membership_eq_expr_with_seen`, which can mark a payload struct’s overloaded `==` method as used. If that operator is only reachable through the new sum equality helper, it is marked after the late pass and can miss body transformation/enqueueing. Move helper synthesis before the late-used scan, or run another late-used transform over names added during helper synthesis.

2. **[P2] Exhaustive sum-type match detection over-covers variants with the same short name.**
   `match_without_else_exhaustive_sumtype_returns` marks a variant covered when `short_type_name(variant) == pattern_short`. For a sum like `type S = a.Foo | b.Foo`, a branch matching `a.Foo` has short name `Foo` and will also mark `b.Foo` as covered. That can make missing-return analysis accept a `match` without `else` even though one variant can still fall through. The short-name fallback should only apply after resolving the pattern uniquely, or it should be removed in favor of exact/resolved variant matching.

3. **[P2] Interface option/result fields can emit `Optional_T` before `T` is complete.**
   `emit_interface_struct` now emits typed option/result fields and calls `emit_struct_option_typedefs` for interface fields, but the interface dependency check still uses `g.tc.c_type(field.typ)`. For `?Foo`, that check sees the generic option type rather than the `Foo` payload dependency, so the interface may be emitted before `Foo`; `emit_optional_typedef` then writes `Foo value;`, which requires `Foo` to be fully defined. The struct path already unwraps `OptionType`/`ResultType` for dependency ordering, so the interface path should mirror that logic.

PR #27678, opened 2026-07-06
v3: parallel parser behind the v3_no_parallel define, like transform/cgen

medvednikov:
Makes the v3 parse phase run across worker threads, gated exactly like the parallel transformer and cgen: on by default, compiled out with `-d v3_no_parallel`, disabled at runtime with `--no-parallel` (Windows always falls back to serial, like the other phases).

### Design

- The input file list is split into contiguous, size-balanced chunks (by source bytes). Each worker parses its chunk into a private `Parser` + `FlatAst` on its own pthread (64 MB stacks); the master parses chunk 0 directly into `p.a` while the helpers run, then joins and merges workers in input order, shifting child node ids and `children_start` offsets by the merge position.
- A file's nodes never reference another file's nodes and chunks are contiguous slices merged in input order, so the merged nodes/children arrays are **byte-identical to a serial parse** — downstream phases see the same AST either way. No AST clones are needed (unlike the transform workers), so the only extra memory is the workers' own parse output.
- `@[export]` registrations are replayed through per-worker records and revalidated against `disabled_fns` accumulated from earlier chunks, reproducing the serial parse's order-dependent check; `disabled_fns`/`noreturn_fns` are unioned in chunk order.
- `resolve_imports` now collects breadth-first waves of newly discovered modules and parses each wave through the dispatcher instead of parsing one module inline at a time. The old loop also appended modules at the end of the node array in discovery order, so the layout is unchanged — verified by identical generated C vs an unmodified build with both running serially.

### Verification

- New `vlib/v3/tests/parallel_parser_test.v`: asserts the threaded parse produces a node-for-node, child-for-child identical flat AST (plus identical side tables and per-file region starts) vs serial; end-to-end multi-module compile reporting `parse (parallel)`; `--no-parallel` opt-out stays serial.
- Self-host fixpoint holds with parallel parse active at every stage (v3 → v4 → v5 → v6, `v5.c == v6.c`), for both dev and `-prod` builds; output is run-to-run deterministic at fixed `VJOBS`.
- A `--no-parallel -building-v` self-host build contains zero parser thread symbols in the generated C; a `-d v3_no_parallel` build of v3 works and reports the phase as serial.
- Full `vlib/v3/tests/` suite: 112/117 pass; the 5 failures reproduce with identical failing subtests on unmodified master (pre-existing, unrelated).

### Performance (self-hosting v3 with `-prod`, Apple Silicon)

```
before:  parse                  123.34 ms      total 845.08 ms
after:   parse (parallel)        49.51 ms      total 770.13 ms
```

Post-parse RSS goes from ~147 MB to ~185 MB under `-gc none` (worker arrays are never freed) — the same tradeoff the parallel transform makes.

### DIFF
--- a/vlib/v3/parser/parser.v
+++ b/vlib/v3/parser/parser.v
@@ -52,9 +52,20 @@ mut:
 	in_for_container  bool
 	local_type_names  map[string]string
 	local_type_scopes []string
+	export_records    []ExportRecord
 pub mut:
 	a              &flat.FlatAst = unsafe { nil }
 	parsed_v_files int
+}
+
+// ExportRecord captures one accepted `@[export: name]` registration in file
+// order. The parallel-parse merge replays worker exports through these records
+// so they can be revalidated against disabled fns from earlier chunks (see
+// merge_parsed_worker), reproducing the serial parse's order-dependent check.
+struct ExportRecord {
+	name  string
+	qname string
+	value string
 }
 
 // new creates a Parser value for parser.
@@ -86,11 +97,25 @@ pub fn (mut p Parser) parse_files(paths []string) &flat.FlatAst {
 	return p.a
 }
 
+// parse_files_with_starts parses paths in order like parse_files, recording the
+// first node id of each file's region. Import resolution uses the starts to
+// bound per-file post-processing (module-name canonicalization) when files are
+// parsed in batches instead of one at a time.
+pub fn (mut p Parser) parse_files_with_starts(paths []string) []int {
+	mut starts := []int{cap: paths.len}
+	for path in paths {
+		starts << p.a.nodes.len
+		p.parse_into(path)
+	}
+	return starts
+}
+
 // parse_into reads parse into input for parser.
 pub fn (mut p Parser) parse_into(path string) {
 	p.cur_file = path
 	p.cur_module = ''
 	p.cur_fn = ''
+	p.has_peek = false
 	p.pending_flag = false
 	p.pending_params = false
 	p.pending_export = ''
@@ -1007,6 +1032,11 @@ fn (mut p Parser) register_pending_export(name string) {
 		return
 	}
 	p.a.export_fn_names[qname] = p.pending_export
+	p.export_records << ExportRecord{
+		name:  name
+		qname: qname
+		value: p.pending_export
+	}
 	p.pending_export = ''
 }
 
--- a/vlib/v3/parser/parser_parallel_d_v3_no_parallel.v
+++ b/vlib/v3/parser/parser_parallel_d_v3_no_parallel.v
@@ -1,0 +1,8 @@
+module parser
+
+// parse_files_dispatch falls back to the serial parse when v3 is built with
+// the internal `v3_no_parallel` define. Returns each file's first node id in
+// p.a and whether worker threads were used.
+pub fn (mut p Parser) parse_files_dispatch(paths []string, _ bool) ([]int, bool) {
+	return p.parse_files_with_starts(paths), false
+}
--- a/vlib/v3/parser/parser_parallel_notd_v3_no_parallel.v
+++ b/vlib/v3/parser/parser_parallel_notd_v3_no_parallel.v
@@ -1,0 +1,249 @@
+module parser
+
+import os
+import runtime
+import v3.flat
+
+// Parallel file parsing. The input file list is split into contiguous,
+// size-balanced chunks; each worker parses its chunk into a private Parser +
+// FlatAst on its own thread, and the master folds the results back in chunk
+// order, shifting node ids and children offsets by the merge position. A
+// file's nodes only ever reference nodes of the same file, and chunks are
+// contiguous slices of the input list merged in input order, so the merged
+// nodes/children arrays are laid out exactly as a serial parse of the same
+// list — the phases downstream see an identical AST either way.
+
+const min_parallel_parse_files = 4
+const min_parallel_parse_bytes = 131072
+const max_parallel_parse_jobs = 8
+
+$if !windows {
+	// ParseChunkArgs is the payload handed to each worker thread.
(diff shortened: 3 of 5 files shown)


medvednikov (commented):
I reviewed this ignoring CI. The core parser chunk merge looks consistent with the current flat AST layout, but I found one behavioral concern in the import-resolution batching:

In the old resolver, implicit imports were seeded immediately after each parsed module (`seed_implicit_sync_import` / `seed_implicit_embed_file_import` ran inside the per-module parse block). In the new resolver, they are seeded once after the whole wave is parsed. These synthetic import nodes do not carry their own source-file context; the next resolve pass infers `importing_file` from the last preceding `.file` node. Because `Preferences.get_module_path()` first tries a path relative to that importing file before falling back to vlib, moving the synthetic import from “after the module that required it” to “after the last module in the wave” can resolve `sync`/`v.embed_file` under a different directory, and it also changes the AST/import-node order compared with serial resolution.

I’d keep the implicit import insertion tied to the originating module/file range, or otherwise attach enough context to these synthetic imports so their resolution does not depend on where the batch appends them.

medvednikov:
Good catch on the implicit-import seeding — fixed in 5c35673fe.

Two things were off relative to serial resolution:
1. **Trigger boundary**: the seeds ran once after the whole wave, so the need scan saw every module in the wave at once instead of re-checking after each module.
2. **Resolution context**: the synthetic import node lands after the wave, so the scan inferred `importing_file` from the last wave file's `.file` marker instead of the module that triggered the seed — and `get_module_path()` resolves project-relative before falling back to vlib.

The fix follows your second suggestion (explicit context rather than moving the node):
- New `seed_implicit_*_upto(mut a, end_node)` variants bound the need scan; the wave loop re-checks at **each module boundary in parse order**, using the region bounds already known from the per-file starts. This fires the seed for exactly the module the serial after-every-module check fired it for (the already-imported dedup still scans the whole array so earlier synthetics are never duplicated).
- `resolve_imports` records the triggering module's last file in a `synthetic_import_ctx` map keyed by node id, and the scan prefers that over the `.file`-marker-inferred `cur_file` when resolving the synthetic import — so `sync`/`v.embed_file` resolve with the same importing-file context as serial, regardless of where the batch appends the node.

Added a regression test (`test_parallel_parser_seeds_implicit_sync_import_mid_wave`): a project where only a mid-wave imported module needs sync (shared field + lock, no explicit import). The parsed sync module symbol set in the generated C is identical to a serial master build. Also re-verified that a `--no-parallel` self-host build of the fixed code produces byte-identical C to the previous commit (the seeding refinement is inert when no seed fires).

One pre-existing issue this surfaced, unrelated to the PR: linking that shared+lock project fails on master too (`undefined symbol '_sync__RwMutex__init'`) — markused prunes RwMutex methods that only cgen-synthesized code calls. The new test therefore checks the generated C rather than the linked binary.

medvednikov (commented):
Thanks for the fix — the explicit context map addresses the case where a synthetic import node is actually appended. I think there is still one related edge case though:

`seed_implicit_*_upto()` bounds the *need* scan, but the duplicate check still calls `ast_has_import()` over the whole AST. Because the entire import wave has already been parsed before the per-module boundary loop runs, imports from modules later in the same wave are now visible to `ast_has_import()` and can suppress a synthetic import that the serial loop would have appended before those later modules existed.

For example: the initial file imports `locker` and `uses_local_sync`; `locker` needs the implicit `sync`, while `uses_local_sync` contains an explicit `import sync` that resolves via its own local `uses_local_sync/sync` directory. Serial resolution seeds `sync` immediately after parsing `locker`, using `locker`'s context, before `uses_local_sync` has been parsed. With the batched wave, `uses_local_sync`'s explicit import is already in `a.nodes`, so `ast_has_import(a, 'sync')` can prevent the locker-triggered synthetic node from being created at all; the new `synthetic_import_ctx` map then never gets a chance to apply.

I think the duplicate check needs to be bounded to the same serial boundary plus already-created synthetic imports, or tracked separately with explicit “already seeded” state, rather than scanning the whole post-wave AST.

medvednikov (commented):
The bounded duplicate scan fixes the “later explicit import suppresses the seed” part, but I think there is still a serial-order mismatch because the synthetic import is still physically appended after the whole wave.

After a wave is parsed, all wave module files sit before the synthetic import nodes that are appended by the per-module boundary pass. On the next resolver pass, scanning resumes at the start of the wave’s parsed files, so imports inside later modules in the same wave are visited before the synthetic import that serial resolution would have appended immediately after the earlier module.

Concretely: main imports `implock` first and `explsync` second. `implock` needs implicit `sync`; `explsync` explicitly imports `sync`, and has a local `explsync/sync` module. Serial resolution parses `implock`, appends synthetic `sync`, then parses `explsync` after that, so the synthetic `sync` is the first `sync` import to be resolved. With the batched layout, `explsync` is parsed in the same wave before the synthetic node is appended, so the next scan sees `explsync`’s explicit `import sync` first, records `parsed_modules['sync']`, and the synthetic node later gets skipped before its `synthetic_import_ctx` can apply.

So I think the fix also needs to preserve the scan order of the synthetic imports relative to later modules in the same wave — for example by inserting/queuing the synthetic import work at the module boundary before later wave-module nodes are scanned, or by resolving/marking those synthetic imports before resuming at the start of the wave.

medvednikov:
@codex review

medvednikov (commented):
Re-reviewed the latest head ignoring CI. The synthetic import handling now preserves the serial module-boundary behavior: seed checks are bounded to the serial boundary, pending synthetic imports are spliced back into the flat AST at the triggering module boundary rather than appended at the wave tail, and the children table is remapped for shifted node ids. I don’t see remaining blockers in the parser/import-resolution path.

medvednikov:
@codex review

PR #27679, opened 2026-07-06
checker,gen: fix generic type inference from a callback with Result return type

### LINKED ISSUE #27108: C error on returning generic anon fn result
### Describe the bug

Running `./v -g -o vdbg cmd/v && ./vdbg 'generic_fn_return.v' && 'generic_fn_return'` on `generic_fn_return.v` produced a compiler error.

### Reproduction Steps

```sh
./v -g -o vdbg cmd/v && ./vdbg 'generic_fn_return.v' && 'generic_fn_return'
```

```v
fn foo[T](do fn () !T) !T {
	mut r := T{}
	r = do()!
	return r
}

fn main() {
	docb := fn () !int {
		return -1
	}
	r := foo(docb)!
	assert r == -1
}
```

### Expected Behavior

Successful compilation and passed assert.

### Current Behavior

```
cc: /tmp/v_1000/generic_fn_return.01KR1TN1DGENRFD9V8GM7MN9P3.tmp.c:8553: error: '{' expected (got ";")
builder error: 
==================
C error found while compiling generated C code.
This can be caused by invalid C interop code, C compiler flags, or a V compiler bug.
If your code is pure V and this still happens, please report it using `v bug file.v`,
```

### Possible Solution

_No response_

Generated by `v bug` from `generic_fn_return.v`.

### V version

V 0.5.1 701d7f3

```
|V full version      |V 0.5.1 2f3c0f1f397c08bfc5ef73e7ff62794471b2b811.701d7f3
|OS                  |linux, Linux version 7.0.3-arch1-1 (linux@archlinux) (gcc (GCC) 15.2.1 20260209, GNU ld (GNU Binutils) 2.46) #1 SMP PREEMPT_DYNAMIC Thu, 30 Apr 2026 18:41:12 +0000
|Processor           |16 cpus, 64bit, little endian, 12th Gen Intel(R) Core(TM) i5-1240P
|Memory              |4.13GB/15.34GB
|                    |
|V executable        |/home/ge/.vlang/v
|V last modified time|2026-05-07 18:19:14
|                    |
|V home dir          |OK, value: /home/ge/.vlang
|VMODULES            |OK, value: /home/ge/.vmodules
|Current working dir |OK, value: /tmp
|                    |
|Git version         |git version 2.54.0
|V git status        |weekly.2026.03-1869-g701d7f37
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20260209
|gcc version         |gcc (GCC) 15.2.1 20260209
|clang version       |clang version 22.1.3
|emcc version        |N/A
```

rilaaax:
# Fixes #27108

## Problem

```v
fn run[T](cb fn () !T) !T {
	mut r := T{}
	r = cb()!
	return r
}

fn main() {
	cb := fn () !int {
		return -1
	}
	run(cb)!
}
```

fails with a C compilation error:

```c
error: initializing '_result_int' (aka 'struct _result_int') with an expression of incompatible type 'int'
```

The issue is specific to the combination of: a generic parameter that is itself a callback (`fn () !T`), where the callback's *return type* carries the generic marker inside a Result (`!T`).
> A generic function with a plain (non-callback) parameter and a `!T` return already works fine.
> A callback parameter with a plain (non-Result) return already works fine too.

Note: the wrong inference happens unconditionally whenever a generic function takes a callback parameter typed `fn () !T` — it isn't specific to using `T{}` in the body. It only surfaces as a visible compile error when the body does something that needs T's plain (unwrapped) C representation, like zero-initializing `T{}`; without that, the mismatch can go unnoticed since the code still happens to work despite T being wrong internally.

## Cause

V infers `T` here by structurally comparing the parameter's function type against the argument's actual function type. Two separate places do this comparison (once in the checker, once in a duplicate implementation in cgen), and in both, the branch that extracts the concrete type from the return position took the argument's return type as-is, without stripping its `.result` flag first. So for a closure returning `!int`, `T` got inferred as `!int` (the Result-wrapped type) instead of `int`. This produced two inconsistent monomorphized versions of the generic function — one correctly named/typed for `T=int`, and a second, wrongly named for `T=!int`, which is the one actually invoked at the call site, leading to invalid C.

## Fix

Strip the option/result flags before binding the inferred type, in both places, using the existing `clear_option_and_result()` helper (already used a few lines above for the analogous case where the parameter itself, rather than just its return type, is Result-wrapped).

## Test

Regression test added.

_NB: credits to Claude_

### DIFF
--- a/vlib/v/checker/check_types.v
+++ b/vlib/v/checker/check_types.v
@@ -1819,7 +1819,11 @@ fn (mut c Checker) infer_fn_generic_types(func &ast.Fn, mut node ast.CallExpr) {
 						}
 						if param_sym.info.func.return_type.has_flag(.generic)
 							&& c.table.sym(param_sym.info.func.return_type).name == gt_name {
-							typ = arg_sym.info.func.return_type.clear_option_and_result()
+							typ = if param_sym.info.func.return_type.has_option_or_result() {
+								arg_sym.info.func.return_type.clear_option_and_result()
+							} else {
+								arg_sym.info.func.return_type
+							}
 							if param_sym.info.func.return_type.nr_muls() > 0 && typ.nr_muls() > 0 {
 								typ = typ.set_nr_muls(0)
 							}
--- a/vlib/v/checker/check_types.v
+++ b/vlib/v/checker/check_types.v
@@ -1819,7 +1819,7 @@ fn (mut c Checker) infer_fn_generic_types(func &ast.Fn, mut node ast.CallExpr) {
 						}
 						if param_sym.info.func.return_type.has_flag(.generic)
 							&& c.table.sym(param_sym.info.func.return_type).name == gt_name {
-							typ = arg_sym.info.func.return_type
+							typ = arg_sym.info.func.return_type.clear_option_and_result()
 							if param_sym.info.func.return_type.nr_muls() > 0 && typ.nr_muls() > 0 {
 								typ = typ.set_nr_muls(0)
 							}
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -3890,7 +3890,11 @@ fn (mut g Gen) update_generic_call_concrete_types_from_fn_types(generic_names []
 	}
 	if param_fn.func.return_type.has_flag(.generic) {
 		gt_name := g.table.sym(param_fn.func.return_type).name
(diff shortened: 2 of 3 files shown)


GGRei:
@codex review

GGRei:
@codex review

medvednikov:
@codex review

medvednikov:
Not an issue in V3.

PR #27686, opened 2026-07-07
v3: comptime `$for field`/`$for value` reflection + &Struct/alias string formatting

medvednikov:
Fixes 42 tracked v3 tests across two areas. Each commit is self-hosted (v5.c == v6.c byte-identical) with no perf/RAM regression and zero regressions across the tracked pass set + comptime suite.

## String formatting (commit 1, +21)
- `&Struct`/`&SumType` interpolation prints the pointee's custom `.str()` (no `&`), else `&nil` for null or `&Type{...}` — instead of a raw pointer.
- `(&Struct).str()` keeps the `&` prefix.
- Alias auto-str: `type Block = [8]u8` → `Block([...])`; aliases of map/struct/sumtype/enum wrap as `Name(...)`; primitive aliases stay bare (matches V1).
- Array string/rune elements quote through alias resolution (`type Literal = string`).
- `@STRUCT` compile-time pseudo-variable (was emitted as undeclared `_v_STRUCT`).

## Comptime `$for` reflection (commits 2-4, +21)
- **`$for field in T.fields`** (previously the parser discarded the body): unrolled once the type is concrete, exposing the full `FieldData` — `name`, `typ`, `unaliased_typ`, `attrs`, `is_option`/`embed`/`array`/`map`/`chan`/`struct`/`enum`/`alias`/`shared`/`atomic`, `indirections`. Supports `receiver.$(field.name)` selectors and per-field `$if field.typ is X` / `$else $if`.
- **`$for value in Enum.values`**: unrolls over enum members (`value.name`/`value.value`, bare `value` → `EnumData`).
- Comptime condition operators (`==` `!=` `<` `>` `<=` `>=` `in` `!in`) were being dropped from the parsed condition string; now emitted, so `$if field.name == 'x'` etc. work.
- Added `$shared`/`$pointer`/`$voidptr` comptime type kinds.
- Enum-alias member access (`type Col = Color` → `Col.member`), resolved in checker + cgen.

### Design notes
- The generic clone no longer specializes calls nested inside a `$for` body (they reference not-yet-resolved loop members), which previously produced uncompilable, uncalled instances.
- Constructs not yet modelled (bodies calling generic functions, `match field.typ`, `typeof(...)`, ORM, bare `FieldData` passed to a generic fn) conservatively fall back to the prior skip behavior, so nothing regresses. `$new`/`$zero`/`.payload_type`, `$for methods/variants`, and recursive generic comptime remain follow-ups (json2 decode needs those).


medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
Pushed ownership/ripgrep_v follow-up commit `eb6ef2b5889e528667dd18965bb07d505eb066b4` to this PR branch.

Summary:
- Preserve checker-selected channel alias `close()` methods instead of lowering them to the builtin runtime close.
- Emit `sync__Channel__close` with the required `IError` array argument, including pointer-to-channel receiver paths.
- Fix generic static associated functions with dotted names from being treated as receiver methods, and broaden generic emitted-name suffix lookup.
- Confirmed the recent comptime ownership branch fixes are already represented on this branch.

Validation run locally:
- `./v -ownership vlib/v3/tests/channel_close_codegen_test.v`
- `./v -ownership vlib/v3/tests/generic_cross_module_arg_codegen_test.v`
- `/Users/alexander/code/v/v -ownership main.v` from `/Users/alexander/code/ripgrep_v`

Note: the broad `vlib/v3/tests/post_merge_review_fixes_test.v` currently stops earlier at the existing `assert item.n == 2` for-mut receiver regression, so I used the focused channel-close regression instead.

PR #27694, opened 2026-07-08
checker: reject unresolved types used in interface `match`/`is` branches

### LINKED ISSUE #26972: cgen error: could not generate string method `redict__Nil_str` for type `redict__Nil`
### Describe the bug

cgen error

### Reproduction Steps

```
git clone https://github.com/einar-hjortdal/redict.git
cd redict
git checkout 832078873655ec1cc5b59986a50cc79041496a88
v install
v -n run command_test.v 
```

### Expected Behavior

no cgen error

### Current Behavior

```
cgen error: could not generate string method `redict__Nil_str` for type `redict__Nil`
```

### Possible Solution

_No response_

Note that I don't get the usual message about cgen errors should never happen etc.

### V version

V 0.5.1 1b7486a

|V full version      |V 0.5.1 58f18573557247e9c743f7d60e5b3ab783f74479.1b7486a
|OS                  |linux, "openmamba release 2026.4 for x86_64 (rolling)"
|Memory              |6.26GB/31.17GB
|                    |
|V last modified time|2026-04-24 08:01:03
|                    |
|Current working dir |OK, value: /home/einar/Documents/projects/vlang/active/redict
|                    |
|Git version         |git version 2.54.0
|V git status        |weekly.2026.03-1709-g1b7486a1
|.git/config present |true
|                    |
|cc version          |cc (GCC) 15.2.1 20251019
|gcc version         |gcc (GCC) 15.2.1 20251019
|clang version       |N/A
|emcc version        |N/A
> [!NOTE]
>

rilaaax:
# Fixes #26972

## Problem

Matching on an interface value with a branch type that was never actually declared (e.g. a typo, or a leftover/renamed type) compiles without any V error, but crashes the underlying C compiler with a confusing error instead:

```c
cc: ...: error: unknown type name 'main__Nil'
```

Minimal repro:

    interface Value {}

    fn (v Value) get_str() string {
        return match v {
            Nil { '' }      // `Nil` is never declared anywhere
            else { '' }
        }
    }

## Cause

The parser registers an unresolved type name like `Nil` as a `.placeholder` type symbol (the normal mechanism used to support forward references). The checker is supposed to reject any `.placeholder` that's still unresolved by the time it's actually used, but `type_implements_with_mut_receiver()` in `checker.v` never checked for this. It validates a type against an interface by iterating over the interface's required methods/fields; for an interface with few or none of those (e.g. an empty `interface Value {}`), the loop does nothing and the function returns `true` by default — regardless of whether the type it was given actually exists. The bogus placeholder type then flows unchecked into cgen, which generates casts and variant-index machinery for a struct that was never defined, and only `cc` catches it.

## Fix

- `checker.v`: `type_implements_with_mut_receiver()` now returns `false` immediately when given a `.placeholder` type, without erroring itself (some callers already have their own dedicated handling for this case).
- `match.v`: the interface branch of `match_exprs()` now reports `unknown type '<name>'` when the branch type doesn't implement the interface *and* is unresolved, mirroring the existing sum type variant error.
- `infix.v`: the `is`/`as` interface check is skipped when the type is a placeholder, since that case is already reported a few lines above, avoiding a duplicate error message.

## Test

Regression test added.

### DIFF
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -2177,7 +2177,7 @@ fn (mut c Checker) type_implements_with_mut_receiver(typ ast.Type, interface_typ
 	utyp := c.unwrap_generic(typ)
 	styp := c.table.type_to_str(utyp)
 	typ_sym := c.table.sym(utyp)
-	if typ_sym.kind == .placeholder {
+	if typ_sym.kind == .placeholder && typ_sym.language != .c {
 		// `typ` is unresolved (undeclared), so never implements anything.
 		// Callers report the error themselves to avoid duplicate messages.
 		return false
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -2177,6 +2177,11 @@ fn (mut c Checker) type_implements_with_mut_receiver(typ ast.Type, interface_typ
 	utyp := c.unwrap_generic(typ)
 	styp := c.table.type_to_str(utyp)
 	typ_sym := c.table.sym(utyp)
+	if typ_sym.kind == .placeholder {
+		// `typ` is unresolved (undeclared), so never implements anything.
+		// Callers report the error themselves to avoid duplicate messages.
+		return false
+	}
 	mut inter_sym := c.table.final_sym(resolved_interface_type)
 	if !inter_sym.is_pub && inter_sym.mod !in [typ_sym.mod, c.mod] && typ_sym.mod != 'builtin' {
 		c.error('`${styp}` cannot implement private interface `${inter_sym.name}` of other module',
--- a/vlib/v/checker/checker.v
+++ b/vlib/v/checker/checker.v
@@ -2158,21 +2158,30 @@ fn (c &Checker) generic_names_for_type_parent(typ_sym &ast.TypeSymbol) []string 
 	return []string{}
 }
 
-// sym_has_unresolved_placeholder reports whether `sym` is
-// itself an unresolved `.placeholder` type symbol (e.g. an undeclared type)
-// or a generic instantiation whose parent type or any of its concrete types is
-// (possibly transitively) an unresolved placeholder.
-fn (c &Checker) sym_has_unresolved_placeholder(sym &ast.TypeSymbol) bool {
+// find_unresolved_placeholder_name recursively looks for an unresolved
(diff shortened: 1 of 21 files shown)


medvednikov:
@codex review

GGRei:
@codex review

GGRei:
The direct unresolved type case is fixed. To address the second Codex review, the checker still needs to reject unresolved placeholders inside generic instantiations, such as `Box[Missing]`, with a regression test covering that case.

Thanks!

GGRei:
@codex review

GGRei:
One Codex point is still only partially fixed. I reproduced it on a clean checkout: `Box[fn (Missing) int]` still reaches C generation and fails with an unknown type error. Please extend the unresolved-type traversal to `ast.FnType` parameters and return types, and add a regression test before closing “Recurse into container type arguments.”

Thanks!

medvednikov:
@codex review

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. One unresolved-type traversal gap remains.

vlib/v/checker/checker.v
```diff
@@ -2158,6 +2158,128 @@ fn (c &Checker) generic_names_for_type_parent(typ_sym &ast.TypeSymbol) []string
...
+	// nested inside one is not missed.
+	nested_types := match sym.info {
+		ast.Array { [sym.info.elem_type] }
+		ast.ArrayFixed { [sym.info.elem_type] }
+		ast.Chan { [sym.info.elem_type] }
+		ast.Map { [sym.info.key_type, sym.info.value_type] }
+		ast.MultiReturn { sym.info.types }
+		ast.Alias { [sym.info.parent_type] }
+		ast.FnType { [sym.info.func.return_type] }
+		else { []ast.Type{} }
```
medvednikov, line 2217: `ast.Thread` is still absent from this recursive container traversal. That means an unresolved type nested in a thread return type, e.g. `Box[thread Missing]` or `Box[thread (int, Missing)]`, can still make `find_unresolved_placeholder_name()` return none and reach the interface `is`/`match` machinery as if fully resolved. Please recurse through `ast.Thread.return_type` here and add a regression for both single and multi-return thread payloads.

PR #27697, opened 2026-07-08
picoev: survive peer RST without changing SIGPIPE disposition

eptx:
A picoev server writing to a connection the client already reset receives SIGPIPE, whose default action terminates the process silently. Long-held file descriptors such as SSE or streaming responses make this routine rather than rare.

This uses local guards without replacing an embedding application’s process-wide signal handler:

- `pico_http_parser.Response.end()` passes `MSG_NOSIGNAL` on non-Windows targets that lack `SO_NOSIGPIPE`.
- `picoev.setup_sock()` enables `SO_NOSIGPIPE` on macOS, iOS, FreeBSD, NetBSD, and DragonFly sockets.

Writes after a peer reset therefore return an error so picoev can drop the connection, including when a caller changes the SIGPIPE disposition or constructs the server on a worker thread. Regression coverage exercises the per-send behavior, verifies the BSD socket option, and ensures failed construction preserves the existing signal disposition.

### DIFF
--- a/vlib/picoev/picoev.v
+++ b/vlib/picoev/picoev.v
@@ -1,6 +1,7 @@
 module picoev
 
 import net
+import os
 import pico_http_parser
 import time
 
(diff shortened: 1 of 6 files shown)


Jengro777:
Is it really necessary to keep the picoev library in the standard library?

medvednikov:
Yes, it's fine in vlib.

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

PR #27701, opened 2026-07-08
parser: fix @@ escape before complex at-expressions in templates

### LINKED ISSUE #27700: Template: `@@` escape broken before complex at-expression (`@@get('/x')` renders `@{get('/x')}`)
### Describe the bug

In compile-time templates (`$tmpl` / `$veb.html()`), the documented `@@`
escape for a literal `@` is broken whenever it directly precedes a **complex
at-expression** (`ident(...)`, `ident[...]`, `ident.field`). Instead of
emitting the literal `@`, the second `@` is re-parsed as the start of a
complex at-expression and wrapped in `@{...}`.

This matters in practice for [Datastar](https://data-star.dev/) hypermedia
apps built with `veb`, where HTML attributes carry action verbs like
`data-on:click="@get('/increment')"`. These must be escaped as `@@get(...)`
in a V template, and currently render broken.

### Reproduction Steps

`main.v`:

```v
fn main() {
	println($tmpl('t.txt'))
}
```

`t.txt` (single line):

```
@@get('/x')
```

```
$ v run .
@{get('/x')}
```

### Expected Behavior

`@@get('/x')` should render as the literal `@get('/x')` (the documented `@@`
escape, see `vlib/v/TEMPLATES.md`, "Escaping").

### Current Behavior

`@@get('/x')` renders as `@{get('/x')}`. The bare-identifier case (`@@world`)
already works correctly; only the complex-at-expression variant is affected.

### Possible Solution

In `rewrite_complex_template_at_expressions` (`vlib/v/parser/tmpl.v`), the
`@@` branch writes a single `@` and advances the index by only **1**:

```v
if next == `@` || next == `{` {
    b.write_u8(`@`)
    i++          // consumes only ONE `@`
    continue
}
```

The second `@` is left in place and re-parsed on the next iteration; when
followed by `get('/x')`, `find_tmpl_complex_at_expr_end` finds the `(...)`
suffix and wraps it into `@{get('/x')}`.

Fix: write **both** `@` characters and advance the index by **2**, leaving the
final `@@` → `@` collapse to `insert_template_code` (which already performs
it). A regression test is included in the linked PR.

The bare-ident case (`@@world`) already works because
`find_tmpl_complex_at_expr_end` returns `-1` for a bare identifier, which is
why the existing escape test (`vlib/v/tests/tmpl/escape.txt`) never exercised
this. Related to #22299 (the bare-ident `@@` case, fixed) — this is its
residual: the complex-at-expression variant was never covered by a regression
test.

### V version

```
V 0.5.1 5bc44d3
```

Reproduced on master at commit `5bc44d3f`. (Did not run `v up` to avoid
clobbering a local working tree; the bug is present in the listed commit.)

```
OS                  | linux, "Fedora Linux 43 (KDE Plasma Desktop Edition)"
Processor           | 12 cpus, 64bit, AMD Ryzen 5 7600
cc version          | gcc (GCC) 15.2.1 20260123 (Red Hat 15.2.1-7)
clang version       | clang version 21.1.8 (Fedora 21.1.8-4.fc43)
tcc version         | tcc version 0.9.28rc 2025-02-13
glibc version       | 2.42
```

(`v doctor` summary; full output available on request.)

EvgenyGuguev:
## What & why

Fixes #27700.

In compile-time templates (`$tmpl` / `$veb.html()`), the documented `@@` escape
for a literal `@` rendered incorrectly whenever it preceded a complex
at-expression (`ident(...)`, `ident[...]`, `ident.field`):

- `@@get('/x')` rendered as `@{get('/x')}` instead of the literal `@get('/x')`.

This is a real-world pain point for Datastar hypermedia apps written with
`veb`, where HTML attributes carry action verbs
(`data-on:click="@get('/increment')") that must be escaped as `@@get(...)` in
a V template.

## Root cause

In `rewrite_complex_template_at_expressions` (`vlib/v/parser/tmpl.v`), the
`@@` branch wrote a single `@` and advanced the index by only 1, leaving the
second `@` to be re-parsed as the start of a complex at-expression (wrapped
into `@{...}`). The bare-ident case (`@@world`) already worked because
`find_tmpl_complex_at_expr_end` returns `-1` for a bare identifier, so the
existing escape test never exercised the complex-suffix case.

## The fix

Write both `@` characters and advance the index by 2, deferring the
`@@` → `@` collapse to `insert_template_code` (which already handles it).

## Tests

- Added a focused regression test:
  `test_escape_double_at_before_complex_at_expression` covering
  `@@get('/x')` and `@@post('/save')`.
- The existing `@@world` / `@{expr}` / `@url` behaviors are unchanged.
- `v test vlib/v/tests/` → 2175 passed, 6 skipped, 0 failures (full template
  suite green, including the new regression). GitHub CI runs the complete
  suite on this PR.


medvednikov:
@codex review

medvednikov:
Thanks!

PR #27703, opened 2026-07-08
ci: stabilize x.async and x.executor Windows tests

GGRei:
## Summary

This PR stabilizes the `x.async` and `x.executor` tests on Windows CI.

The failures exposed both timing-sensitive test assumptions and a Windows semaphore issue. `Semaphore.timed_wait()` could consume an available token through its slow path while still reporting a timeout.

The runtime fix is limited to the Windows semaphore implementation. The public API remains unchanged.

## Changes

- Fix Windows `Semaphore.timed_wait()` so successful acquisition is reported correctly
- Add a deterministic regression test for the affected slow path
- Make `x.async` pool close/drain tests track released jobs explicitly instead of relying on shared anonymous release signals
- Make `x.executor` lifecycle and `run_sync` tests use owner-thread barriers before calling `stop()`
- Keep executor stress coverage strong on Linux/macOS while using platform-appropriate workloads on Windows
- Use Windows-specific wait budgets where CI scheduling is slower

## Validation

- Full `vlib/sync` tests in development and production modes
- Full `x.async` and `x.executor` tests in development and production modes
- Linux: GCC, Clang and TCC
- macOS: Clang
- Windows: GCC, MSVC and TCC
- Dedicated Windows semaphore regression test
- Smoke CI: https://github.com/GGRei/v/actions/runs/29150716206

No CI workflow files are changed by this PR.

### DIFF
--- a/vlib/sync/semaphore_windows_test.v
+++ b/vlib/sync/semaphore_windows_test.v
@@ -1,0 +1,13 @@
+module sync
+
+import time
+
+fn test_semaphore_timed_wait_reports_acquisition_after_fast_path() {
+	mut sem := new_semaphore_init(1)
+	did_acquire := sem.wait_for_available_count(time.second)
+	was_consumed := !sem.try_wait()
+	sem.destroy()
+
+	assert did_acquire
+	assert was_consumed
+}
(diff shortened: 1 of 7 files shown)


GGRei:
Ready

PR #27709, opened 2026-07-09
v3: fix 50 more tests

medvednikov:
## Summary
- fix V3 lowering/codegen gaps for shared values, lock expressions, optional/result multi-return values, enum backing types, fixed-array optional returns, generic option calls, and map/array/string `or` lowering
- preserve user-defined `free()` methods while keeping default generated free calls no-op where appropriate
- preserve backed `[flag] enum ... as u64` values through backing-width C symbols and autostr/mask generation
- preserve wrapped `+`/`-` infix continuations after scanner-inserted semicolons
- keep option/result wrappers around normalized fixed-array type names like `?int[4]` and `!Foo[2]`
- move 101 confirmed tests from `v3_test_failures.txt` to `v3_test_passes.txt`
- keep self-host working through `v3 -> v4`

## Verification
- focused V3 regression set for the latest 50 moved tests: `PASS 50/50`
- `vlib/v/tests/enums/enum_bitfield_64bit_test.v`: pass
- `vlib/v/tests/enums/maps_with_enum_flag_keys_test.v`: pass
- `./v test vlib/v3/tests/parser_regression_test.v`: pass
- focused wrapped plus/minus continuation snippet with V3: pass (`7`, `7`)
- focused option/result fixed-array unwrap snippet with V3: pass (`10`, `26`)
- manifest check: no pass/fail overlap; `enum_bitfield_64bit_test.v` is in passes and not in failures
- `git diff --check`
- `vlib/v3 $ ../../v -gc none -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`
- latest self-host run after parser/option fixed-array review fixes: parse `80.84 ms / 186 MB`, check `56.99 ms / 399 MB`, markused `50.88 ms / 483 MB`, transform `185.99 ms / 2028 MB`, monomorphize `4.02 ms / 2030 MB`, cgen `152.75 ms / 2365 MB`, cc `96.10 ms / 2365 MB`, total `627.72 ms`; outer `/usr/bin/time -l` wall `31.83 s`, max RSS `2479849472` bytes


medvednikov:
Addressed the review in 0b1bb8318. The compiler default `.free()` no-op path now only applies to typed receiver expressions with no known receiver `free` method, so module-qualified `free` functions and user-defined receiver `free()` methods dispatch normally.\n\nVerified:\n- `../../v test tests/review_cgen_regressions_test.v`\n- `vlib/v3 $ ../../v -gc none -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`

medvednikov:
Pushed `f476dd920` with 50 more confirmed V3 fail-to-pass tests.

Verification:
- focused compile+run verifier: `PASS 50/50`
- latest batch manifest check: `0` remain in `v3_test_failures.txt`, all `50` are in `v3_test_passes.txt`
- `git diff --check`: clean
- self-host passed: `vlib/v3 $ ../../v -gc none -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`
- latest self-host: parse `71.00 ms / 186 MB`, check `55.57 ms / 398 MB`, markused `50.31 ms / 481 MB`, transform `163.55 ms / 2024 MB`, monomorphize `3.93 ms / 2026 MB`, cgen `131.14 ms / 2359 MB`, cc `87.92 ms / 2359 MB`, total `563.59 ms`; `/usr/bin/time -l` max RSS `2473689088` bytes

Perf/RSS comparison against a clean `HEAD` temp-worktree baseline: all reported stage times were lower and outer max RSS was lower by `1032192` bytes. The cumulative RSS labels at transform/annotate/monomorphize measured `+4 MB`; cgen/cc measured `-1 MB`.

medvednikov:
Addressed the backed flag enum review in `4724f9506`.

What changed:
- backed `[flag] enum ... as <int>` fields no longer use the `map[string]int` enum-value cache for emitted values; they are stored/emitted through backing-width C symbols such as `PawnsBoard__h1`
- flag enum autostr now tests bits using the backing storage type, so `${wb}` preserves bits above 31
- flag masks for backed enums are built from the backed symbols instead of an `int` mask
- added zero-padded base formatting support for `064b`, needed by `enum_bitfield_64bit_test.v`
- moved `vlib/v/tests/enums/enum_bitfield_64bit_test.v` from V3 failures to passes

Verification:
- `vlib/v/tests/enums/enum_bitfield_64bit_test.v`: pass
- `vlib/v/tests/enums/maps_with_enum_flag_keys_test.v`: pass
- `git diff --check`: clean
- self-host passed: `vlib/v3 $ ../../v -gc none -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`
- latest self-host run: total `611.14 ms`, max RSS `2479931392` bytes

medvednikov:
Addressed the two new review comments in `48e1bdff1`.

- `vlib/v3/parser/parser.v`: wrapped `+`/`-` continuations now consume the scanner-inserted semicolon like other infix continuations.
- `vlib/v3/types/checker.v`: normalized `?T[N]` / `!T[N]` type names now parse as option/result wrappers around fixed arrays, not fixed arrays of wrappers.

Verification:
- `./v test vlib/v3/tests/parser_regression_test.v`: pass
- focused V3 wrapped plus/minus snippet: pass (`7`, `7`)
- focused V3 option/result fixed-array unwrap snippet: pass (`10`, `26`)
- `git diff --check`: pass
- `vlib/v3 $ ../../v -gc none -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`: pass; latest warmed self-host run `31.83s` wall, max RSS `2479849472` bytes, stage RSS unchanged at the same ~2.365 GB cgen plateau.

I also reran `./v test vlib/v3/tests/review_transform_regressions_test.v`; it still has the existing unrelated `if_expr_selector_decl_smartcast_local` checker failure (`Animal` as `Cat`).

medvednikov:
Addressed the two new `vlib/v3/transform/or.v` review comments in `dd650c12d`:

- channel receive `or` lowering now stores the transformed channel source in a `&sync.Channel` temp and reuses it for both `sync__Channel__pop` and `sync__Channel__closed_error`
- array index `or` lowering for optional/result elements now binds `err` from the loaded optional/result element on in-bounds element failure, while preserving the synthetic `none` error for bounds failure

Verification:

- `./v fmt -verify vlib/v3/transform/or.v vlib/v3/tests/or_expr_transform_review_test.v`
- `git diff --cached --check` before commit
- `./v test vlib/v3/tests/or_expr_transform_review_test.v` passed
- `cd vlib/v3 && ../../v -gc none -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` passed twice

Latest self-host metrics:

```text
parse        73.99 ms / 187 MB
check        58.01 ms / 399 MB
markused     51.19 ms / 482 MB
transform   177.98 ms / 2032 MB
monomorphize  3.74 ms / 2034 MB
cgen        155.59 ms / 2367 MB
cc          111.87 ms / 2367 MB
total       632.51 ms
max RSS 2482323456
```

medvednikov:
Addressed the remaining unresolved review threads in `5456f5ce5`:

- preserved user-defined `.free()` methods while keeping compiler-default missing free calls as `((void)0)`
- aligned backed enum map key storage size with the selected hash/equality callbacks
- required actual `<-` prefix receives before channel receive `or` lowering
- preserved loaded map optional/result element errors through `or` lowering

Verification:

- `./v fmt -verify vlib/v3/gen/c/fn.v vlib/v3/gen/c/struct.v vlib/v3/transform/or.v vlib/v3/transform/map.v vlib/v3/tests/or_expr_transform_review_test.v`
- `git diff --check`
- `./v test vlib/v3/tests/or_expr_transform_review_test.v` passed
- `cd vlib/v3 && ../../v -gc none -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` passed 3x

Best stable self-host sample:

```text
parse        79.81 ms / 187 MB
check        55.20 ms / 399 MB
markused     50.49 ms / 483 MB
transform   188.89 ms / 2036 MB
monomorphize  3.84 ms / 2038 MB
cgen        156.53 ms / 2303 MB
cc          103.88 ms / 2303 MB
total       638.78 ms
max RSS 2414919680
```

Peak RSS is below the previous exact run; stage timings are within the observed run-to-run spread from the repeated self-host passes.

medvednikov:
Addressed the two latest review threads in `942f15eae`:

- backed enum map keys now use the enum backing storage type for map-facing temporaries and lowered map operations, so 8-byte callbacks do not read from 4-byte enum temps.
- channel `.try_push()` / `.try_pop()` now dereference pointer channel receivers before calling the runtime helpers.

Verification already run before the later “skip tests” scope change:
- `./v test vlib/v3/tests/or_expr_transform_review_test.v`
- `vlib/v3 $ ../../v -gc none -prod -o v3 v3.v`
- `vlib/v3 $ ./v3 -building-v -o v4 v3.v`

medvednikov:
Addressed the latest thread-wait review in `7e5ef1abc`.

`[]thread ?T` / `[]thread !T` wait wrappers now store the first failed optional/result error, continue joining every thread handle, and return that saved error only after the join loop completes.

Per the latest scope request, I skipped test runs for this follow-up and only ran `git diff --check`.

medvednikov:
Addressed the two latest channel review issues in `4ad3b3c09`:

- `chan.close(err...)` C-gen now preserves explicit close error arguments by emitting a populated `[]IError` instead of always passing an empty array.
- Channel receive `or` lowering now tracks pointer channel sources and dereferences them before storing the `&sync.Channel` used by `sync__Channel__pop`/`closed_error`.

Verification: `git diff --check -- vlib/v3/gen/c/fn.v vlib/v3/transform/or.v`.

Per request, I did not run tests.

medvednikov:
Addressed the backed flag enum `from()` review in `c576a4ef9`:

- `Enum.from()` now stores the input in the enum storage C type before validation, so backed flag enums keep high bits during the mask check.
- Optional payload naming now uses the value C type, so backed enum optionals use a payload wide enough for the enum typedef instead of falling back through `int`.

Verification: `git diff --check -- vlib/v3/gen/c/fn.v vlib/v3/gen/c/types.v`.

Per request, I did not run tests.

medvednikov:
@codex review

medvednikov:
Addressed the optional/result thread-array wait helper review in `08a00eb80`:

- `ensure_thread_optional_arr_wait_fn` is now cached by the emitted optional ABI shape (`ret_ct`) instead of by source spelling (`?T` vs `!T`).
- Removed the now-unused `ret_name` parameter so the helper cache and emitted symbol naming stay aligned.

Verification: `git diff --check -- vlib/v3/gen/c/array.v`.

Per request, I did not run tests.

medvednikov:
Addressed the generic backed-enum map key review in `85e10c546`:

- Generic-cloned `new_map` calls now retarget `sizeof(key)` and map callbacks through `map_key_storage_type()` after type substitution, so `K = Wide as u64` uses `u64` storage/hash/eq.
- Lowered map key temp identifiers/declarations from generic clones are retargeted to the concrete backing storage type for internal map operations.

Verification: `git diff --check -- vlib/v3/transform/monomorphize.v`.

Per request, I did not run tests.

medvednikov:
Addressed the unsupported fixed-array method review in `f35792998`:

- `fixed_array_dynamic_receiver_call_info` now returns only for methods in `fixed_array_lowered_methods`.
- Existing dedicated `pointers` handling remains separate, so fixed-array `.pointers()` still follows its addressability checks and intrinsic path.

Verification: `git diff --check -- vlib/v3/types/checker.v`.

Per request, I did not run tests.

medvednikov:
Addressed the sign-preserving zero padding review in `eb92adb96`.

`v3_string_zpad` now keeps a leading `-` at the front and inserts zero padding after the sign, so base-formatted negative values like `${a:08x}` preserve the existing interpolation semantics.

Verification: `git diff --check -- vlib/v3/gen/c/cleanc.v`.

Tests skipped per request.

medvednikov:
Addressed the latest `vlib/v3/gen/c/fn.v` review comments in `9f15d77cd`.

Changes:
- `.try_push()` now uses `g.value_c_type(channel_type.elem_type)` for the payload compound literal, preserving concrete optional/result storage like `Optional_string`.
- flag `Enum.from()` now validates the original argument in a `u64` temp against the mask before narrowing into the enum backing storage, so narrow backed flags reject wrapped values like `Small.from(256)`.

Verification: `git diff --check -- vlib/v3/gen/c/fn.v`.

Tests skipped per request.

medvednikov:
Addressed the backed flag `set_all()` review in `b1a3988e9`.

For backed flag enums, transform now leaves `set_all()` as a flag enum method call instead of lowering it through the transform `int` mask helper. That defers emission to C generation, where `flag_enum_mask_expr(...)` uses the backed enum constants and preserves high bits such as `u64(1) << 63`.

Verification: `git diff --check -- vlib/v3/transform/fn.v`.

Tests skipped per request.

medvednikov:
Addressed the latest shared-parameter and backed-enum review comments in `86b7f8d3c`.

Changes:
- The checker now preserves shared-parameter metadata in `CallInfo`, tracks shared local bindings by scope owner, and rejects non-shared arguments before C generation can emit a value for a `__shared__T*` parameter.
- Backed non-flag enums now emit `typedef <backing> EnumName;` plus typed field constants, and enum value lookup/stringification uses those backed constants instead of a C `typedef enum` ABI.

Verification:
- `git diff --check -- vlib/v3/types/checker.v vlib/v3/gen/c/types.v vlib/v3/gen/c/cleanc.v`
- `./v -check vlib/v3/types` and `./v -check vlib/v3/gen/c` parsed the touched modules and only reported the expected no-main-module message.

Tests skipped per request.

medvednikov:
Addressed the shared-field shared-parameter review in `ed153c9f5`.

Changes:
- Cache direct `shared` struct fields from raw field declarations during V3 type collection.
- Allow `expr_is_shared_arg` to accept selector expressions whose selected field is declared `shared`.
- Emit shared field selector arguments through the existing shared storage path, including parenthesized/prefixed forms and shared receivers.

Verification:
- `git diff --check -- vlib/v3/types/checker.v vlib/v3/gen/c/fn.v vlib/v3/gen/c/struct.v`
- `./v -check vlib/v3/types` (expected no-main boundary)
- `./v -check vlib/v3/gen/c` (expected no-main boundary)
- `./v -check vlib/v3/v3.v` (passed with existing notices only)

Tests skipped per request.

medvednikov:
Addressed the latest two review comments in `df38e12ed`.

Changes:
- Skip unresolved generic shared parameter type texts when collecting global shared wrapper typedefs, so `fn f[T](shared x T)` no longer emits `__shared__T` before specialization.
- Validate scalar variadic extra arguments in `check_call_arg_types`, which makes `chan.close(error('a'), 123)` reject the second argument before cgen.

Verification:
- `git diff --check -- vlib/v3/gen/c/struct.v vlib/v3/types/checker.v`
- `./v -check vlib/v3/v3.v` (passed with existing notices only)
- `./v -check vlib/v3/gen/c` (expected no-main boundary)
- `./v -check vlib/v3/types` (expected no-main boundary)

Tests skipped per request.

medvednikov:
Addressed the enum backing alias review in `0cbae0edd`.

Change:
- Backed enum typedef/autostr emission now uses `enum_backing_info(name).storage_c_type` when available, preserving aliases resolved in the enum's defining module. Raw backing reparsing remains only as a fallback/registration path.

Verification:
- `git diff --check -- vlib/v3/gen/c/types.v`
- `./v -check vlib/v3/v3.v` (passed with existing notices only)
- `./v -check vlib/v3/gen/c` (expected no-main boundary)

Tests skipped per request.

medvednikov:
Addressed the zero-argument function-value spawn review in `34abe710a`.

Change:
- Function-value spawn paths now match on `fn_type.params.len == call_arg_count` even when the count is zero, so `spawn cb()` packages the function pointer and calls through `p->f()` instead of emitting a named no-arg wrapper that references local `cb` out of scope.

Verification:
- `git diff --check -- vlib/v3/gen/c/fn.v`
- `./v -check vlib/v3/v3.v` (passed with existing notices only)
- `./v -check vlib/v3/gen/c` (expected no-main boundary)

Tests skipped per request.

medvednikov:
Fixed the latest two review issues in `4a8af39b4`:

- `Color.from_string(...)` now falls through to user-defined static methods when the checker resolved a real method, so the synthetic enum field-name lowering only handles the builtin enum path.
- spawn wrapper argument packing now captures shared storage expressions, so shared fields like `s.a` are packed as the shared wrapper pointer for `shared` parameters.

Verification run, no full tests per request:
- `git diff --check -- vlib/v3/transform/or.v vlib/v3/gen/c/fn.v`
- `./v -check vlib/v3/v3.v`
- package checks reached the expected no-main diagnostic for `vlib/v3/gen/c` and `vlib/v3/transform` after parsing

medvednikov (commented):
Fixed the pointer-channel statement receive issue in `899931eb8`.

Statement-position `<-ch` now unwraps pointer-channel sources, emits `sync__Channel__pop(*(ch), ...)`, and still uses the existing direct-channel pop path for non-pointer channels.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/gen/c/cleanc.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the shared generic signature issue in `320d2062f`.

Generic signature substitution now recurses through `shared ` and preserves the shared prefix after parsing/normalization, so clones of `fn f[T](shared x T)` keep concrete shared parameter text such as `shared int` for wrapper collection/codegen instead of registering `__shared__T`.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/transform/monomorphize.v vlib/v3/types/checker.v`
- `./v -check vlib/v3/v3.v`
- package checks for `vlib/v3/transform` and `vlib/v3/types` reached the expected no-main diagnostic after parsing

medvednikov (commented):
Fixed the latest two review issues in `278bc05e5`.

- Fixed-array index guards now share the array-index lowering metadata and use compile-time fixed-array lengths for guard/or bounds checks, so `if x := fixed[i] { ... }` is lowered instead of falling through to optional-style C.
- Fixed-array receiver recovery now only accepts bracket text that is a real fixed-array length expression/const, so generic receiver names such as `Box[int]` are not misidentified as fixed arrays before method lookup.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/types/checker.v vlib/v3/transform/or.v vlib/v3/transform/if.v`
- `./v -check vlib/v3/v3.v`
- package checks for `vlib/v3/transform` and `vlib/v3/types` reached expected no-main diagnostics after parsing

medvednikov (commented):
Fixed the enum backing alias map-lowering review issue in `48debb15f`.

`enum_backing_types` now stores the backing storage type after resolving aliases in the enum declaration module, and map lowering uses that stored storage type directly instead of normalizing raw alias text in the importing/current module. This avoids emitting out-of-scope alias names such as `Bits` for `map[board.Square]T`.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/transform/map.v vlib/v3/transform/transform.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the latest two review issues in `651dc5285`.

- Shared receiver methods now validate the receiver expression with `expr_is_shared_arg` when param 0 is marked shared, matching the explicit shared-argument path.
- Backed enum declarations now get a dedicated semantic validation path: the backing text is parsed in the enum declaration context, aliases are unwrapped, unknown types are rejected, and only integer storage types are accepted.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/types/checker.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the latest two review issues in `1d3a26dbb`.

- `Enum.from_string` lowering now resolves selector import aliases in the transformer before looking up enum metadata, matching the checker-accepted `import mod as alias; alias.Enum.from_string(...)` form.
- Backed enum declarations now validate folded explicit and auto-incremented field values against the backing storage before C emission. For flag enums this validates the bit index against the backing width, preserving valid wide backed flags such as `u64` bit 63 while rejecting out-of-range narrow storage.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/types/checker.v vlib/v3/transform/or.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the unsigned backed enum `.str()` fallback issue in `f4aba8a2b`.

Backed non-flag enum autostr now detects unsigned storage C types (`u8`, `u16`, `u32`, `u64`, `size_t`) and emits `strconv__format_uint((u64)(storage)it, 10)` for fallback formatting. Signed storage still uses the existing `strconv__format_int` path.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/gen/c/types.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the backed enum field-reference emission issue in d619694fb.

Backed non-flag enum constants now fold initializer expressions with enum-declaration context before emission, so bare or qualified references to enum fields resolve to computed values instead of raw C identifiers like `a`. If an initializer cannot be folded, cgen still preserves the old expression fallback while rewriting enum-field references to the generated constants.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/gen/c/types.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the fixed-array optional ABI issue in 2f12984e0.

Fixed-array C names and emitted element types now use the same concrete optional/result ABI for elements such as `[2]?string`: typedef names, direct fixed-array parameter signatures, return wrappers, declarations, data casts, and fixed-array slices now agree on `Optional_string` instead of mixing `Optional_string[N]` with `Optional[N]`. The fixed-array typedef emitter also emits optional element typedef dependencies before using them.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/types/checker.v vlib/v3/gen/c/types.v vlib/v3/gen/c/cleanc.v vlib/v3/gen/c/fn.v vlib/v3/gen/c/array.v vlib/v3/gen/c/struct.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the backed enum unfolded-initializer validation issue in a944f8dcf.

The backed-enum range checker now records an error for explicit initializers that cannot be folded instead of validating the implicit `next_val`. The enum initializer folder also handles integer `cast_expr` nodes, so cases like `u16(300)` are folded to `300` and checked against the backing range.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/types/checker.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the shared-argument review issues in `5ad3de571`.

Shared argument validation/codegen now only unwraps parentheses, not arbitrary prefix expressions, so computed values like `-n`/`!flag` no longer satisfy `shared` parameters or get stripped during emission. Method-call explicit arguments now mirror normal call shared-param emission, so `fn (f Foo) use(shared x int)` receives the shared wrapper pointer for later shared parameters.

Verification, no full tests per request:
- `git diff --check -- vlib/v3/types/checker.v vlib/v3/gen/c/fn.v vlib/v3/gen/c/struct.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the explicit shared argument review issue in `4c4db3724`.

`expr_is_shared_arg` now unwraps only the parser's explicit `shared` prefix before validating the child as shared storage, while other computed prefix expressions remain rejected. The shared codegen helpers also understand that explicit prefix so normal calls, method calls, and spawn packing use the shared wrapper/storage expression consistently.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/types/checker.v vlib/v3/gen/c/fn.v vlib/v3/gen/c/struct.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the shared-prefix and backed flag `from_string` review issues in `12450de16`.

Explicit `shared` call prefixes are now rejected by the checker for non-shared parameters, and the generic prefix emitter falls back to the child value instead of blindly emitting the wrapper pointer. For backed flag enum `from_string` lowering, matched members now assign through enum value nodes so cgen uses the backing-width enum constants rather than the `int` metadata value.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/types/checker.v vlib/v3/gen/c/cleanc.v vlib/v3/transform/or.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the function-value spawn optional ABI issue in `493d7fe20`.

Spawn argument packing now derives heap argument field types from `value_c_type`, so option/result parameters with concrete payloads use the same C layout as normal argument emission, e.g. `Optional_string` rather than erased `Optional`. I applied the same helper to the adjacent named-call spawn packing path so both spawn wrappers agree with emitted parameter ABI.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/gen/c/fn.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the fixed-array if-guard payload issue in `c8d9821bc`.

Single optional/result if-guard bindings now mirror the existing multi-return fixed-array handling: declare the destination fixed array and copy `tmp.value` with `memmove` instead of trying to initialize one C array object from another.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/gen/c/if.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the fixed-array function-value spawn wrapper issue in `11fd1c832`.

Spawn-packed arguments now know when they hold fixed arrays. The spawn emitters initialize those heap fields with `memmove` and keep wrapper calls passing `p->aN`, so the fixed array decays to the callee’s expected element pointer instead of trying to assign one C array object to another.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/gen/c/fn.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the non-addressable `try_pop` destination issue in `1e63616fe`.

`chan.try_pop` now has a checker-side destination guard: pointer expressions are accepted, and non-pointer destinations must be addressable mutable storage. This rejects cases like `ch.try_pop(1)` and `ch.try_pop(make_value())` before cgen can emit invalid `&expr` C.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/types/checker.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the three latest review issues in `6b7a036f0`.

Changes:
- Pointer-like selector calls resolved to `free`/`builtin.free` now emit `free(receiver)` instead of taking the compiler-default no-op path.
- Optional/result thread-array waits now copy fixed-array payloads with `memmove` instead of assigning C array typedef values.
- Map `or` lowering no longer treats arbitrary `Optional_*` source names as option/result values; it uses actual optional/result type syntax after alias normalization.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/gen/c/fn.v vlib/v3/gen/c/array.v vlib/v3/transform/map.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the two shared-storage review issues in `cf334cb5c`.

Changes:
- Shared locals initialized from fixed-array values now allocate the wrapper first and copy into `->val` with `memmove`, avoiding C array field initialization from another array.
- `gen_shared_storage_expr` now unwraps parenthesized/explicit-shared local wrappers, so lock targets like `lock (n)` address `n->mtx` rather than `n->val->mtx`.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/gen/c/stmt.v vlib/v3/gen/c/struct.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the two latest `fn.v` review issues in `760cd5276`.

Changes:
- `chan.try_push` now materializes fixed-array element payloads in a temporary and copies them with `memmove` before passing them to `sync__Channel__try_push`.
- Flag-enum method lowering now only intercepts actual flag enums, so normal enums with user-defined `toggle`/`set_all`/`clear_all` methods fall through to normal method dispatch.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/gen/c/fn.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the deferred optional multi-return `none` review issue in `2d1326252`.

Change:
- `return_expr_string` now mirrors the direct return path and treats all-`none` multi-return children as the outer optional/result `none` before building a successful tuple value.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/gen/c/stmt.v`
- `./v -check vlib/v3/v3.v`

medvednikov (commented):
Fixed the generic spawn return type review issue in `26ed4ebf5`.

Change:
- Spawn expression typing now falls back to the resolved call expression type when the direct callee lookup returns `Unknown`, preserving specialized generic return payloads like `thread int` for `spawn work[int](...)`.

Verification, no full tests per prior request:
- `git diff --check -- vlib/v3/types/checker.v`
- `./v -check vlib/v3/v3.v`

PR #27715, opened 2026-07-09
markused,gen: fix direct `str()`call on a smart-casted interface variable

### LINKED ISSUE #27713: cgen: `builtin__str_intp` is not imported when calling`.str()` directly on a smart-casted interface variable
### Describe the bug

Calling `.str()` directly on an interface variable that's been smart-cast to a primitive type inside a `match` branch (without rebinding it to a new variable first) crashes the C compiler.

### Reproduction Steps

```v
interface Value {}

fn get_str(v Value) string {
	return match v {
		int {
			v.str()
		}
		else {
			''
		}
	}
}

fn main() {
	get_str(42)
}
```

### Expected Behavior

success

### Current Behavior

```c
cc: error: call to undeclared function 'builtin__str_intp'
cc: error: use of undeclared identifier 'StrIntpData'
```

### Possible Solution

_No response_

_No response_

### V version

V 0.5.1 48f5856

not relevant
> [!NOTE]
>

rilaaax:
# Fixes #27713

## Problem

Calling `.str()` directly on an interface variable that's been smart-cast to a primitive type inside a `match` branch (without rebinding it to a new variable first) had two separate bugs:

1. It crashed the C compiler instead of producing a working binary:
```
cc: error: call to undeclared function 'builtin__str_intp'
cc: error: use of undeclared identifier 'StrIntpData'
```

2. Once that crash was fixed, the printed result was wrong: `&42` instead of `42`

Minimal repro:
```v
interface Value {}

fn get_str(v Value) string {
	return match v {
		int { v.str() }
		else { '' }
	}
}

fn main() {
	println(get_str(42))
}
```

Note: rebinding first (`x := v; x.str()`) avoids both issues, which was the
first clue pointing at cgen/dead-code-elimination rather than the checker.

## Cause

Interfaces box non-pointer-kind variants (like `int`) behind an allocated pointer in C (`_int` is stored as `int*`). For a direct `.str()` call on a smart-cast identifier, cgen resolves the receiver using the *raw* internal smartcast type, which still carries that pointer flag — unlike the "exposed" V-level type (plain `int`) that the checker and `node.left_type` use.

This raw, pointer-flagged type causes two independent problems:

1. `markused/walker.v` decides whether to keep `str_intp` (the string interpolation runtime) based on `node.left_type` (plain `int`, no pointer flag), so it never learns that cgen's pointer-flagged code path will actually need `str_intp`, and strips it from the final C output.
2. `gen_expr_to_string()` in `str.v` treats any pointer-flagged type as "this is a real reference the user wants printed with `&`" — correct for genuine pointers/structs stored by reference, but not for a primitive value that only *looks* like a pointer because of interface boxing.

## Fix

- `markused/walker.v`: when registering `.str()` usage, also check whether the receiver is a smart-cast identifier whose raw smartcast type is a pointer, and register that type too, so `str_intp`/`StrIntpData` aren't stripped when cgen will actually need them.
- `str.v`: detect the same case (an interface variable smart-cast to a primitive/non struct-like type) and use an empty ref prefix instead of `&`, leaving all other pointer/struct/reference printing behavior unchanged.

## Test

Regression test added.

### DIFF
--- a/vlib/v/gen/c/coutput_test.v
+++ b/vlib/v/gen/c/coutput_test.v
@@ -335,6 +335,24 @@ fn test_windows_sharedlive_string_interpolation_in_ternary_does_not_emit_inline_
 	assert compilation.output.contains('builtin__str_intp')
 }
 
+fn test_interface_match_smartcast_direct_str_call() {
+	os.chdir(vroot) or {}
+	test_source := os.join_path(os.vtmp_dir(), 'coutput_interface_match_smartcast_str_intp.vv')
+	test_exe := os.join_path(os.vtmp_dir(), 'coutput_interface_match_smartcast_str_intp')
+	os.write_file(test_source,
+		"module main\n\ninterface Value {}\n\nfn get_str(v Value) string {\n\treturn match v {\n\t\tint {\n\t\t\tv.str()\n\t\t}\n\t\telse {\n\t\t\t''\n\t\t}\n\t}\n}\n\nfn main() {\n\tprintln(get_str(42))\n}\n")!
+	defer {
+		os.rm(test_source) or {}
+		os.rm(test_exe) or {}
+	}
+	build_cmd := '${os.quoted_path(vexe)} -o ${os.quoted_path(test_exe)} ${os.quoted_path(test_source)}'
+	build_result := os.execute(build_cmd)
+	ensure_compilation_succeeded(build_result, build_cmd)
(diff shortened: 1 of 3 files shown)


GGRei:
@codex review

rilaaax:
Closing since this was fixed in 29a44b92713840a10f8ce519d10a0adf1fe436a9

PR #27723, opened 2026-07-10
net: avoid BSD fasthttp loopback deadlock

guweigang:
## Summary

This fixes a BSD/macOS fasthttp deadlock where a handler can make an HTTP
request back to the same server while the single kqueue loop is blocked inside
that handler.

The BSD backend now starts a fixed set of kqueue event-loop workers. Each worker
has its own client tracking map, while the listening socket remains shared
between the workers. Accepted client sockets are owned by the worker that
accepted them.

## Design notes

This is intentionally smaller than the earlier broader refactor attempt in
https://github.com/vlang/v/pull/27477. That approach moved too much of the
request lifecycle around and was too risky for fasthttp's hot path.

I also tried aligning BSD with the Linux backend more directly by creating one
listener socket per worker with `SO_REUSEPORT`. That model is cleaner for fd
ownership, but on macOS it does not reliably solve the loopback case: the inner
loopback connection can be distributed back to the same worker that is currently
blocked inside the outer handler.

So this PR keeps the shared listener approach as a narrow BSD-specific fix:

- fixed worker count, not per-request spawning;
- each worker has its own kqueue and client map;
- handlers still run inline on the worker that owns the client fd;
- listener readiness is level-triggered, so accept batching does not lose queued
  connection readiness;
- each accept pass is capped to avoid one worker draining the whole backlog;
- client fd registration failures now close/free the accepted connection;
- fatal non-shutdown `kevent` errors clean up that worker's clients before
  exiting.

This keeps the loopback fix while avoiding the heavier request-lifecycle changes
from the previous attempt.

## Tests

On macOS:

```text
./vnew test vlib/fasthttp/fasthttp_bsd_loopback_test.c.v
./vnew -silent test vlib/fasthttp/
```

I also repeated the loopback test three additional times locally to check for
obvious flakiness.


medvednikov:
@codex review

GGRei:
@codex review

GGRei:
@codex review

GGRei:
@medvednikov / @JalonSolov From my perspective, this PR is ready to merge. Thanks!

PR #27731, opened 2026-07-10
ci: smoke test the veb example (fasthttp) http server on Windows

medvednikov:
## What

Adds an end-to-end CI smoke test that compiles `examples/veb/veb_example.v` (veb uses the fasthttp backend), runs the produced server, and performs real HTTP GET requests to verify it accepts connections and returns the expected HTML/text.

## Why

Follow-up to #27727 (which fixed the pointer-sized Windows `SOCKET` truncation in `accept()`). That was a codegen/type-width fix with only a compile-time regression test — nothing in CI actually *ran* a veb/fasthttp server on Windows and hit it over HTTP, which is the exact path that crashed:

```
Unhandled Exception 0x1e94fa30 at 6577c7
... by builtin__unhandled_exception_handler
```

This adds that missing coverage so the crash can't silently come back.

## How

New helper `.github/workflows/veb_example_windows_smoke.v` (placed alongside the other `.v` CI helper scripts, so `test-self` never double-runs it):

1. Compiles `examples/veb/veb_example.v`.
2. Starts the produced server as a background process.
3. Polls until it accepts connections, **failing fast** if the process dies early (the crash signature) instead of waiting out the timeout.
4. `GET /` → verifies the rendered `index.html` template (title, included header, `Hello world from veb, request number:`, footer, …).
5. `GET /show_text` → verifies the plain-text route returns `Hello world from veb`.
6. Always shuts the server down afterwards; uses short (5s) HTTP timeouts so a half-crashed server can't stall the job.

Wired into all three Windows workflows (`windows_ci_tcc.yml`, `windows_ci_gcc.yml`, `windows_ci_msvc.yml`) right after `Build examples`:

```yaml
- name: Compile and run the veb example (fasthttp) and check its HTML
  run: v run .github/workflows/veb_example_windows_smoke.v
```

The logic is cross-platform, so the same helper can be reused on other CI targets later.

## Testing

The fasthttp backend also runs on macOS/Linux, so the helper was validated locally:
- Happy path: compiles, serves, HTML/text assertions pass, clean shutdown, exit `0`.
- Failure path (port pre-occupied so the server can't bind): detected immediately via `is_alive()`, exit `1`, no hang.
- `v fmt -verify` clean.

### DIFF
--- a/.github/workflows/veb_example_windows_smoke.v
+++ b/.github/workflows/veb_example_windows_smoke.v
@@ -1,0 +1,136 @@
+// veb_example_windows_smoke.v is a CI smoke test for the fasthttp based `veb`
+// backend. It compiles `examples/veb/veb_example.v`, runs the produced server,
+// and performs real HTTP GET requests to verify that the server actually
+// accepts connections and returns the expected HTML.
+//
+// It is primarily a regression test for the Windows crash that happened when
+// `accept()` truncated pointer-sized `SOCKET` handles (see
+// vlib/fasthttp/fasthttp_windows.c.v), which made every veb/fasthttp server
+// crash on the very first incoming request on Windows. The logic itself is
+// cross platform, so the same program can be reused on other CI targets.
+//
+// Usage: v run .github/workflows/veb_example_windows_smoke.v
+module main
+
+import os
+import time
+import net.http
+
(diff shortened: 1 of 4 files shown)


medvednikov:
@codex review

PR #27744, opened 2026-07-11
v3: fix 39 more tests

medvednikov:
Continues the v3 backend work. Moves 39 tests from `v3_test_failures.txt` to `v3_test_passes.txt` (each verified twice; excludes everything already handled by #27730).

## Compiler fixes

**json2 / generics**
- Removed the early-dev stubs that replaced `json2.decode`/`encode` calls with `none`/default values; `json2`/`x.json2` now monomorphize normally (the old C-magic `json` module keeps its cgen shortcut, matching #27730).
- `$if T.unaliased_typ is X` conditions: `T` is now substituted in metadata selectors and the `.typ`/`.unaliased_typ` suffix is resolved, so specialized branches prune correctly.
- `$if` guards decidable after generic substitution are evaluated during cloning, so dead branches no longer leak uncompilable specializations (`decode_string[SomeStruct]`) into the monomorphize scan.
- Generic type-arg inference no longer picks up stale same-named locals from previously transformed functions (a `decode_value[T]` call inferred `T=ParserState` from an unrelated `result` local), and the internal `&T` mut-param annotation no longer leaks into inferred args.
- Cloned bodies register their params/locals, so nested generic call retargeting sees correct types.

**Checker**
- No more false errors for `x is T` in generic templates, calls with placeholder generic args (`p.read_element[T]()`), and method calls on fields of generic struct instances.
- Locals declared in comptime `$for` bodies that reference the loop var now enter the scope (`mut fo := ...(field.attrs)` followed by `fo.install_default(...)`).

**Codegen correctness**
- Struct `!=` negated only the first field comparison (`!a && b` instead of `!(a && b)`); prefix operators now parenthesize binary operands.
- `>>>` / `>>>=` were emitted as plain signed shifts; now logical shifts at the operand's width with oversized-count clamping.
- Int-literal shifts >= 31 were performed at C `int` width — `u64(1 << 51)` wrapped and poisoned all of ed25519's field math.
- u8/u16 arithmetic is truncated back to its V width in comparisons and string conversion (`u8(255) + u8(1) == 0`).
- Single-char `c'x'` literals used as byte values are dereferenced (`*"x"`); multi-char `c'...'` stay C strings.
- Recursive sum types (`Any = []Any | map[string]Any`) crashed the compiler with a stack overflow in shared-field lookup; added a visited set.
- Import-alias qualified types (`json.Any` under `import x.json2 as json`) now resolve in const/global initializers and map-literal lowering (cgen tracks the declaring file).
- `typeof(x).name` prints `[][3]int` instead of the internal `[]int[3]` form.
- `(*f._string).len` parenthesization for smartcast variant bases.
- `#insert`ed headers keep `<dlfcn.h>` / `<limits.h>` (per-OS macro values can't be inlined) — fixes `dlopen` and MbedTLS "8-bit chars" errors.
- `IError == IError` with unknown concrete types compiled to invalid C; now compares `msg()` + `code()`.
- `$if test {}` is true when compiling `_test.v` files.

## Self-host

`v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` verified, plus v4 → v5:

| stage | before | after |
|---|---|---|
| host `-prod` build of v3 | 35.8 s / 1.48 GB | 34.5 s / 1.47 GB |
| `./v3 -building-v` self-compile | 1.27 s / 2.34 GB | 0.94 s / 2.19 GB |

v3 unit tests (`v test vlib/v3`) pass.

Review follow-ups are in the second commit: comptime `$if` pointer-var semantics, `c'...'` literal contexts, `>>>=` single-evaluation, `ptrdiff_t`/`size_t` shift widths, IError `==` operand spilling, non-generic `is T` validation, and the implicit `test` define ordering. `generic_selector_type_test.v` was dropped from the moved set after turning out to be flaky under the parallel checker (method resolution through an alias of a generic instance fails on some runs with identical sources).

## Remaining failure clusters (root-caused, for follow-ups)

- ~124 `vlib/v2_toberemoved/*` tests fail with v1 too (`import v2.x` dangles since the rename) — not v3 bugs.
- Unimplemented subsystems: `$for m in T.methods/attributes` (~16), `$tmpl()` (~11), closure capture (18+), MbedTLS header inlining (~25), ORM (stubbed), `v.reflection`, interface `str()` dispatch.
- json2 sum-type decode/encode additionally needs `$for v in Sum.variants` from #27730.

### DIFF
--- a/vlib/v3/gen/c/array.v
+++ b/vlib/v3/gen/c/array.v
@@ -908,7 +908,11 @@ fn (mut g FlatGen) gen_index_assign(node flat.Node) {
 			g.write('; int _i${tmp} = ')
 			g.gen_expr(g.a.child(&lhs, 1))
 			g.write('; array__set(_a${tmp}, _i${tmp}, &(${c_elem}[]){')
-			if op := compound_assign_to_infix_op(node.op) {
+			if node.op == .right_shift_unsigned_assign {
+				lhs_text := '*(${c_elem}*)array_get(*_a${tmp}, _i${tmp})'
+				g.gen_unsigned_right_shift_from_text(lhs_text, g.a.child(&node, 1),
+					arr_type.elem_type)
+			} else if op := compound_assign_to_infix_op(node.op) {
 				if arr_type.elem_type is types.String && op == .plus {
 					g.write('string__plus(')
 					g.write('*(string*)array_get(*_a${tmp}, _i${tmp})')
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -56,9 +56,11 @@ mut:
 	interfaces                     map[string][]string
 	const_vals                     map[string]flat.NodeId
 	const_modules                  map[string]string
+	const_files                    map[string]string // const name -> declaring file (for import-alias type resolution)
 	const_init_order               []string
 	fixed_storage_consts           map[string]bool
 	global_modules                 map[string]string
+	global_files                   map[string]string      // qualified global name -> declaring file (for import-alias type resolution)
 	global_inits                   map[string]flat.NodeId // qualified global name -> initializer value node
 	global_init_order              []string               // qualified global names, in declaration order
 	enum_backing_infos             map[string]EnumBackingInfo
@@ -379,9 +381,11 @@ pub fn FlatGen.new() FlatGen {
 		interfaces:                     map[string][]string{}
 		const_vals:                     map[string]flat.NodeId{}
 		const_modules:                  map[string]string{}
+		const_files:                    map[string]string{}
 		const_init_order:               []string{}
 		fixed_storage_consts:           map[string]bool{}
 		global_modules:                 map[string]string{}
+		global_files:                   map[string]string{}
 		global_inits:                   map[string]flat.NodeId{}
 		global_init_order:              []string{}
 		enum_backing_infos:             map[string]EnumBackingInfo{}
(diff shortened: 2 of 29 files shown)


medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

PR #27756, opened 2026-07-13
v3: fix 62 more tests

medvednikov:
Moves 62 tests from `v3_test_failures.txt` to `v3_test_passes.txt` (1385 → 1323 failing, 1915 → 1977 passing). Verified with a full sweep of both lists; the previously-passing list shows no regressions (every remaining harness failure fails identically with a baseline v3 built from master).

Self-host still converges (`v5.c` ≡ `v6.c`) at baseline speed and memory: ~600ms total, 1.62–1.66s user, RSS in the usual 2.3–2.5GB range.

### Main fixes

**Headerless C preamble gaps** (compress/zstd, gzip, szip, bzip2, veb infra)
- Inlined third-party C sources (`zstd.c`, miniz) lost their system includes; the preamble now declares the libc names they need early (malloc/calloc/realloc/free, pthread create/join/detach/cond_*, stdio/fopen family, mktime/localtime/utime/stat, fprintf/fflush/qsort_r, clock), plus `clock_t`/`CLOCKS_PER_SEC`, `mode_t`, `struct utimbuf`, `S_IRWX*`/`S_IS*`/`EEXIST`, an `<assert.h>` replacement, and keeps `<arm_neon.h>` in place (guarded by `__ARM_NEON`, so tcc skips it and cc gets real intrinsics).
- `typedef struct X X;` is now supplied for inlined C structs that have a tag but no typedef (`C.zip_t`).
- stdint replacement guards `_INTMAX_T`/`_UINTMAX_T` and uses `long` on Apple LP64 to avoid SDK typedef conflicts.

**toml cluster (15 tests)**
- `val is map[string]ast.Value` cross-module resolved the pattern via `all_after_last('.')`, collapsing the container to `Value` and matching the sum type itself (tag index 0 → every `is` check false). Container patterns are now excluded from the short-name resolution, sum names resolve through import aliases, and cgen's `sum_type_index` matches module-stripped container spellings.

**Language/codegen fixes**
- Match branches can now produce tuple values for multi-assign: `a, b, c := match e { .x { v1, v2, v3 } }` (unblocked `veb/context.v`).
- Gated slices `a#[..]` lower to `slice_ni`/`substr_ni` (clamped, negative-index aware) instead of the panicking `slice`; the scanner previously dropped the `#` marker entirely.
- `[..]u32[...]` inferred-size fixed array literals parse (bzip2, crypto tables).
- Variadic fns called with named args across modules (`http.new_header(key: .content_type, value: ...)`) resolve the variadic flag through alias/C-name keys and wrap the struct literal into a one-element array.
- Const initializers with `.map()` chains (veb's mime tables) emit their lowered statements as a braced runtime init; block values in expression position no longer drop a bare trailing expression.
- `&char` accepted where `&u8` is expected (C interop, `tos_clone(C.ZSTD_versionString())`).
- Three `mut` value-param miscompiles: `&param` typed `&&T` emitting `map** t = table` + `*t = ...` (clobbered the caller's map in toml's `find_in_table`), sum assignment through a mut param taking a spurious `&`, and `T(v)` casts of `$for v in T.variants` loop vars assigning a raw `VariantData` literal instead of wrapping the variant's zero value.

### Flipped tests by area
toml (15), compress: zstd/gzip/szip/bzip2 (5), json2/jsonrpc (2), veb (2), net.s3 (3), comptime `$match`/`$if` (8), match/sumtype (7), and 20 more individual `vlib/v/tests` cases.


medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

PR #27757, opened 2026-07-13
fasthttp: keep Windows IOCP operations reachable

Maokaman1:
Windows retains only the raw `OVERLAPPED` pointer while asynchronous IOCP work is pending, which is invisible to V’s GC.
This can allow `IocpOperation` to be collected before completion and cause a use-after-free crash.

Keep the pending operation reachable through its registered connection until completion or submission failure.
Add Windows regression coverage for pending reads/writes, forced GC, and large responses.


medvednikov:
@codex review

PR #27760, opened 2026-07-13
cgen: fix option reference string interpolation

### LINKED ISSUE #27752: cgen: `${&val}` on an option value emits `&&val` and fails to C-compile
**V version:** `V 0.5.1 c7c032c` (also reproduced on fresh `origin/master`, `ec8aa142c9`)
**OS:** macOS (cc/clang), also seen on Linux

### Describe the bug

Taking the address of an **option-typed** value inside a string interpolation, e.g. `'${&val}'`, makes cgen emit a double ampersand `&&val` in the generated C. The C compiler parses `&&` as the GNU computed-goto *label-address* operator (`&&label`), so the build fails.

### Reproduction

```v
fn f(val ?int) {
	println('${&val}')
}

fn main() {
	f(?int(42))
}
```

```
v run repro.v
```

### Expected behavior

Compiles and prints the option's address representation (e.g. `Option(&...)`), the same way `'${&val}'` works for a non-option `val`.

### Current behavior

The C compiler fails:

```
error: expression cannot be followed by a postfix '.' operator; add parentheses
error: member reference base type 'void *' is not a structure or union
error: use of undeclared label 'val'
```

The last error is the tell: `&&val` is being read as the address of a label named `val`.

### Generated C

The interpolation of `&val` produces `&&val` (the user's `&` plus the option str-interp helper's own `&`):

```c
builtin__isnil(*(int*)&&val.data) ? _S("Option(&nil)")
                                  : builtin___option_int_str(*(_option_int*)&&val)
```

### Notes

- The alias is irrelevant — a plain `?int` parameter reproduces it. A `type MaybeInt = ?int` alias behaves the same.
- Dropping the `&` (`'${val}'`), or using a non-option value with `&`, compiles fine — so the extra `&` is added specifically on the option-address path.

> [!NOTE]
>

cyphercodes:
Fixes #27752

This fixes C generation for string interpolation of references to option values, such as `${&val}` where `val` is `?int`.

The interpolation stringification path now materializes the existing option-payload-reference lowering into a temporary before adding the generated nil check/string call, avoiding invalid generated C such as `&&val.data`.

A regression test covers both a payload option and `none`.

## Tests

```sh
V_C_ERROR_BUG_REPORT_DISABLED=1 ./v test vlib/v/tests/builtin_strings_and_interpolation/string_interpolation_test.v
V_C_ERROR_BUG_REPORT_DISABLED=1 ./v test vlib/v/tests/options/option_reference_of_option_test.v
git diff --check
```


GGRei:
This looks good for the direct ?int case, but one case mentioned in #27752 is still missing, an alias such as `type MaybeInt = ?int` still generates `&&val.data` and fails during C compilation.

Could you also handle the aliased option type and add a regression case for it? After that, the issue should be fully covered.

medvednikov:
@codex review

medvednikov:
lots of ci errors

cyphercodes:
Updated this to cover the aliased option case from the review.

What changed:
- detect option aliases when deciding whether `&option` string interpolation needs an option-payload temporary
- generate the temporary as `?&T` after unaliasing, so aliased `?T` avoids the bad `&&val.data` C output
- added regression coverage for `type MaybeInt = ?int`

Verification run locally:
- `./v self`
- `./v -showcc -cg -keepc -o /tmp/alias_option_ref /tmp/alias_option_ref.v` (prints `&Option(42)`)
- `./v -showcc -cg -keepc -o /tmp/string_interp_test vlib/v/tests/builtin_strings_and_interpolation/string_interpolation_test.v`
- `/tmp/string_interp_test`
- `git diff --check`

GGRei:
@codex review

GGRei:
The red CI jobs appear to share the same root cause, the chained `&& match right_expr { ... }` expression in `vlib/v/gen/c/str.v` generates invalid C around a `goto`. The build then stops early, causing the remaining CI failures.

Can you fix it plz?

cyphercodes:
Updated to avoid the `&& match right_expr { ... }` condition in `vlib/v/gen/c/str.v` that was generating invalid C around a `goto`.

Verification run locally:
- `make -j2`
- `./v self`
- `./v -cc gcc -showcc -cg -keepc -o /tmp/alias_option_ref /tmp/alias_option_ref.v` + `/tmp/alias_option_ref` (prints `&Option(42)`)
- `./v -cc gcc -showcc -cg -keepc -o /tmp/string_interp_test vlib/v/tests/builtin_strings_and_interpolation/string_interpolation_test.v` + `/tmp/string_interp_test`
- `git diff --check`

GGRei:
@codex review

medvednikov:
Addressed all Codex findings on `878448e175` in `3e026b0f94`:

- preserve payload aliases (including aliases nested behind an option alias) when materializing `?&T` interpolation temporaries
- perform option-reference lowering before enum-specific stringification
- resolve concrete generic interpolation/right-hand types and correctly handle referenced heap arguments across multiple generic instantiations
- strengthen regression coverage with exact output checks for direct options, option aliases, custom-string aliases, enums, and generics

Validation:
- targeted string interpolation test
- targeted test with `-cstrict -cc clang`
- compiler errors: 1589 passed, 1 skipped
- C output: 96 output cases and 125 must-have cases passed (platform skips only)
- diagnostic in/out: 81 passed, 9 expected panics, 1 skipped

medvednikov:
@codex review

medvednikov:
@codex review

cyphercodes:
Merged current `master` into this branch and resolved the string interpolation conflict by preserving the option-reference temporary path alongside the newer auto-struct root guard / generic interpolation type resolution.

Verification run locally:
- `./v -nocache -d v1_fallback -g -keepc -o ./vnew cmd/v`
- `./vnew -cc gcc -showcc -cg -keepc -o /tmp/alias_option_ref /tmp/alias_option_ref.v` + `/tmp/alias_option_ref` (prints `&Option(42)`)
- `./vnew -cc gcc -showcc -cg -keepc -o /tmp/string_interp_test vlib/v/tests/builtin_strings_and_interpolation/string_interpolation_test.v` + `/tmp/string_interp_test`
- `./vnew -silent test vlib/v/gen/c/coutput_test.v`
- `git diff --check`

PR #27798, opened 2026-07-15
v3: fix group 6 test failures

medvednikov:
## Summary
- fixes the V3 failures tracked in group 6
- preserves shared-local raw type metadata so shared array append/postfix autolock works without the expensive declaration scan
- adds the grouped V3 failure list files and README

## Validation
- vlib/v3/v3_test_failure_groups/06_v_tests_fns_and_control_flow.txt: total=110 pass=110 fail=0
- cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v
- /Users/alexander/code/v6/vnew test vlib/v3/transform/shared_postfix_test.v
- git diff --check


medvednikov:
Addressed the review points in 331b72f48e.

- Unsigned inclusive `for` lowering now transforms the RHS bound once per attempted iteration into a temp and reuses that same value for both the loop condition and the tail guard. Added a side-effecting bound regression.
- I did not reject fixed-slot spreads globally, because V3 already expands that supported shape in transform before C generation (`transform_spread_arg_over_fixed_variadic_tail`). Added a focused codegen regression for `take(...xs)` where the spread fills a fixed param plus the variadic tail.

Validation:
- `./vnew test vlib/v3/tests/for_multi_init_codegen_test.v`
- `./vnew test vlib/v3/tests/variadic_call_codegen_test.v`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`
- group 6 list: 110/110 pass

medvednikov:
Addressed the latest review points in 95c23d804b.

- Fixed-slot spreads with trailing args now append the remaining explicit args onto a fresh variadic tail array, so `take(...xs, 9)` keeps `9` instead of dropping it.
- For-post multi-assign now uses the regular multi-return assignment expansion before falling back to plain temp-based lowering, so `a, b = step()` updates both LHSs.
- For-post arity mismatches now preserve/report a `for post assignment mismatch` marker; `i = bump(1), bump(2)` is rejected before codegen instead of losing the extra RHS silently.

Validation:
- `./vnew test vlib/v3/tests/variadic_call_codegen_test.v`
- `./vnew test vlib/v3/tests/for_multi_init_codegen_test.v`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`
- group 6 list: 110/110 pass

medvednikov:
@codex review

medvednikov:
Addressed the latest review pass in commit 3295fff828.

Changes:
- for-post continue rewriting now descends into match branches and similar branch containers while still stopping at nested loops.
- assignment checking now invalidates written smartcasts after RHS checking, with preservation for assignments that still produce the active smartcasted value.
- captured fn literal aliases are tracked and rejected when appended/stored in callback containers, covering direct and aliased container storage without regressing existing group-6 closure coverage.

Validation:
- ./vnew test vlib/v3/tests/checker_assignment_review_test.v
- ./vnew test vlib/v3/tests/for_multi_init_codegen_test.v
- ./vnew test vlib/v3/tests/capturing_callback_thread_local_codegen_test.v
- v3 group 6 sweep: 110/110
- cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v
- git diff --check

Note: I also attempted the full vlib/v3/tests/post_merge_review_fixes_test.v sweep, but stopped it after it was still running past 5 minutes; the directly relevant multi-assignment smartcast snippet was run with the rebuilt v3 and passed.

medvednikov:
Addressed the latest review findings in the amended commit 5ae7e671e0:\n\n- lowered for-post continuations now structurally descend through select branches and other transparent child containers while still stopping at nested loops/function literals\n- shared-array selector indexing no longer treats pointer receivers as pointer array expressions\n- shared-array push markers now preserve the field owner module for imported shared array fields\n\nValidation run:\n- ./vnew test vlib/v3/tests/for_multi_init_codegen_test.v\n- ./vnew test vlib/v3/tests/shared_array_codegen_test.v\n- cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v\n- group 6 sweep: 110/110\n- git diff --check

medvednikov:
Addressed the latest review findings in amended commit 758ccbf0b4:

- labeled `continue` nodes are no longer wrapped with the local lowered for-post body; labeled loop lowering already routes them through the target loop's continue label
- for-init/for-post multi-assign nodes now preserve LHS count, and scalar single RHSs for multi-LHS assignments are rejected unless the RHS resolves as a multi-return value

Validation run:
- `./vnew test vlib/v3/tests/for_multi_init_codegen_test.v`
- `./vnew test vlib/v3/tests/checker_assignment_review_test.v`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`
- group 6 sweep: `110/110`
- `git diff --check`

medvednikov:
Addressed the latest review findings in amended commit b01a62e8e5:

- `for k, mut v in m` with a delete-triggered map snapshot now binds mutable values through the original map, with a snapshot fallback if that key was deleted before its turn
- capturing fn literal local markers now track binding owners, so an inner shadowing declaration does not inherit the outer captured marker, while the outer marker still applies after the shadowing block

Validation run:
- `./vnew test vlib/v3/tests/checker_assignment_review_test.v`
- `./vnew test vlib/v3/tests/map_for_mut_snapshot_codegen_test.v`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`
- group 6 sweep: `110/110`
- `git diff --check`

medvednikov:
Addressed the latest review findings in amended commit eb32f34331:

- Preserved string-alias operator overloads by making alias-operator detection recover explicit cast targets and allowing alias equality overload lookup to mark the current infix as the use.
- Fixed mixed duplicate lock lowering so write mode propagates backward through duplicate runs before acquisition skips later duplicates.
- Added loop-control copybacks for lowered mutable map value writebacks so fixed-array map values are copied back before `continue`/`break` paths that leave or advance the map loop.

Validation:
- `./vnew test vlib/v3/tests/operator_result_method_checker_test.v`
- `./vnew test vlib/v3/tests/map_for_mut_snapshot_codegen_test.v`
- `./vnew test vlib/v3/tests/mixed_lock_codegen_test.v`
- `./vnew test vlib/v3/tests/lock_codegen_test.v`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` (second-stage total 1229.52 ms, peak 2304 MB resident RAM)
- Group 6 sweep: `GROUP6_PASS=110/110`
- `git diff --check`

medvednikov:
Addressed latest review findings in amended commit 133c15669c:

- markused now collects the checker-resolved sum receiver method before falling back to variant helper methods for non-local receiver expressions.
- variadic fn-value local markers are tracked by binding owner, so shadowing locals no longer inherit variadic call behavior.
- option-context `if` branches reject dynamic `error()`/`IError` branches, while literal-condition branches only check the selected branch.
- fixed result/error cgen fallout so optional/result failure literals render `err` fields as raw `IError`, not nested option wrappers.

Validation:
- `./vnew test vlib/v3/tests/fn_value_decl_type_test.v`
- `./vnew test vlib/v3/tests/markused_test.v`
- `./vnew test vlib/v3/gen/c/`
- Direct dynamic option/error rejection check for `fn f(ok bool) ?int { return if ok { error('bad') } else { 1 } }`
- Direct `vlib/v/tests/conditions/ifs/if_expr_of_option_test.v` check
- Group 6 sweep: `GROUP6_PASS=110/110`
- Self-host: `../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` passed; final run total 1194.93 ms, peak 2332 MB resident RAM.

Note: full `vlib/v3/tests/type_checker_errors_test.v` was also attempted, but it still fails on unrelated existing generic/interface cases (`global_amp_initializers`, generic literal diagnostics, fixed-array interface dispatch, etc.); the new option/error case was verified directly.

medvednikov:
Addressed the latest review findings in f60de580c3.\n\n- for-post checking now isolates condition smartcasts so the post clause is not accepted under stale condition-only narrowing.\n- top-level markused selector collection now prefers checker-selected/exact typed receiver methods before sum variant helpers.\n\nValidation:\n- ./vnew test vlib/v3/tests/for_multi_init_codegen_test.v\n- ./vnew test vlib/v3/tests/markused_test.v\n- group 6 sweep: 110/110 passed\n- cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v

medvednikov:
Addressed the latest parser/ORM review findings in 0baaadc12d.

- Parser: newline before `(` is only treated as a multiline call continuation for ident/selector lhs forms when the `(` line is indented beyond the lhs line indentation. This preserves the existing `new_address\n\t(...)` style while keeping `(foo())` as a separate parenthesized statement after `println('a')`.
- ORM cgen: NULL TEXT fallback now uses `g.intern_string('')` instead of hard-coded `_str_27`.

Validation:
- `./vnew test vlib/v3/tests/parser_regression_test.v`
- `./vnew test vlib/v3/tests/orm_null_text_codegen_test.v`
- `./vlib/v3/v3 -silent test vlib/v/tests/fns/fn_call_with_newline_opening_brace_test.v`
- group 6 sweep: 110/110 passed
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`

Self-host timing/RSS check: final run total 1148.80 ms; peak printed RSS 2316 MB. I also ran the same self-host command against baseline f60de580c3 in a detached worktree; it printed essentially the same peak RSS (2308 MB), so the RSS level is not introduced by these fixes.

medvednikov:
@codex review

medvednikov:
Addressed the P1 match parser review in 9b11864836.

- `match` now always parses through `match_stmt()`, so parenthesized and array-literal subjects like `match (maybe() or { ... }) {}` and `match []int{} {}` are preserved.
- Added a parser regression covering both subject forms.

Validation:
- `./vnew test vlib/v3/tests/parser_regression_test.v`
- `./vnew -run-only test_eval_or_block_return_propagates_from_match_subject vlib/v3/eval/eval_test.v`
- `./vnew -run-only test_eval_match_expression_propagates_branch_return vlib/v3/eval/eval_test.v`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`
- group 6 sweep: 110/110 passed

Note: the full `vlib/v3/eval/eval_test.v` still hits an unrelated pre-existing eval panic in `test_eval_index_assignment_lhs_index_propagates_return`, so I ran the cited match test directly with `-run-only`.

medvednikov:
Addressed the nested optional infix `or` review. Nested optional leaves now lower into a single whole-expression result temp guarded by an `ok` flag, so a failed left leaf assigns the outer fallback and skips later operands; when the right side contains the optional, the left side is stabilized first to preserve effects before the failure.

Validation:
- `./vnew test vlib/v3/tests/or_expr_transform_review_test.v`
- group 6 sweep: 110/110 passed
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` passed
- RSS sanity check against pre-fix HEAD: baseline peak 2318 MB, patched peak 2341 MB in this worktree, so no material stage memory regression from this fix.

medvednikov:
Addressed the latest checker reviews.

Changes:
- Plain pointer truthiness is no longer accepted for normal `if` / condition-only `for` conditions.
- C-style `for init; ptr; post` pointer conditions remain accepted to match current V behavior and keep group 6 passing.
- Optional maps remain rejected in for-in before codegen.
- Optional arrays are now lowered with an `ok` guard, so `none` skips the loop instead of reading `.value` unconditionally.

Validation:
- `./vnew test vlib/v3/tests/for_in_review_test.v`
- direct minimal snippets for pointer `if`, pointer `for`, optional map rejection, optional array `none` skip, and C-style pointer condition
- group 6 sweep: 110/110 passed
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` passed, peak RSS 2331 MB

medvednikov:
Addressed the ORM wide integer SQLite review.

- `i64`, `u64`, `isize`, and `usize` ORM integer fields now bind with `sqlite3_bind_int64` instead of `sqlite3_bind_int`.
- Reads for those fields now use `sqlite3_column_int64` and cast back to the field C type.
- Narrow integer fields still use the existing 32-bit SQLite APIs.

Validation:
- `./vnew test vlib/v3/tests/orm_null_text_codegen_test.v`
- group 6 sweep: `GROUP6 pass=110 fail=0 total=110`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` (passed, peak RSS 2371 MB)

medvednikov:
Addressed the spawned capturing literal review.

The wrapper still restores captures before calling the lifted literal, but those `__anon_fn_*` capture slots are emitted by `global_decls()` as per-thread lvalues (`_Thread_local` for normal C compilers, pthread-key backed under TinyCC). I added an explicit comment at the restore site and a regression that spawns many mutable capturing literals from the same source site, checks the generated TLS/thread-wrapper shape, and verifies each thread observes its own capture.

Validation:
- `./vnew test vlib/v3/tests/capturing_callback_thread_local_codegen_test.v`
- group 6 sweep: `GROUP6 pass=110 fail=0 total=110`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` (passed, peak RSS 2371 MB)

medvednikov:
Addressed the latest generic-call and unsigned-loop reviews.

Changes:
- Explicit generic calls now always emit runtime arguments starting from call child `1`; type arguments stay on the `.index` callee, so `f[int](1, 2)` no longer drops `1`.
- Unsigned inclusive loop overflow guards now skip the cached-bound lowering when the loop body writes a name used by the RHS bound. Immutable/effect-only bounds still use the per-iteration cached value for the overflow guard, preserving the existing group-6 max-bound behavior.

Validation:
- `./vnew test vlib/v3/tests/generics_test.v`
- `./vnew test vlib/v3/tests/for_multi_init_codegen_test.v`
- group 6 sweep: `GROUP6 pass=110 fail=0 total=110`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` (passed, peak RSS 2342 MB)

medvednikov:
Addressed the ORM `u32` SQLite integer review.

- `u32` ORM fields now use the int64 SQLite path for both bind and read, avoiding signed `int` conversions for values above `INT_MAX`.
- Signed `int`/`i32` remains on the existing `sqlite3_bind_int` / `sqlite3_column_int` path.
- Extended the ORM codegen regression to cover `u32(1) << 31` alongside the existing 64-bit and pointer-sized integer cases.

Validation:
- `./vnew test vlib/v3/tests/orm_null_text_codegen_test.v`
- group 6 sweep: `GROUP6 pass=110 fail=0 total=110`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` (passed, peak RSS 1102 MB)

medvednikov:
Addressed the ORM transform-gating reviews.

Changes:
- SELECT is now routed to V3 C lowering only for the shapes the C generator actually supports: `select from T` and `select from T limit N`. `where` / `order by` forms stay on the fallback path instead of emitting a stripped `SELECT ... FROM` query.
- INSERT is now routed to C lowering only when the inserted value has a known scalar target struct type. Bulk values like `[]User` stay on the fallback path until bulk lowering exists.
- Added codegen regressions for unsupported SELECT clauses and bulk INSERT fallback, while the existing scalar insert ORM regression still verifies the supported path.

Validation:
- `./vnew test vlib/v3/tests/orm_null_text_codegen_test.v`
- group 6 sweep: `GROUP6 pass=110 fail=0 total=110`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` (passed, peak RSS 2326 MB)

medvednikov:
Addressed the latest review comments in amended commit a3928ef65c.

Changes:
- Preserve failures from lambdas coerced to `?void` / `!void` callbacks by directly returning compatible optional/result-void bodies, and returning explicit `error(...)` bodies for `!void`.
- Route all checker-resolved user `new_map(...)` calls to `main__new_map(...)`, regardless of arity.
- Added module-keyed fn return-type lookup so builtin runtime `new_map(...)` and user `main__new_map(...)` keep separate return types.
- Added regressions for result/option-void callback lambdas and `new_map(x)` call emission.

Validation:
- Manual regression snippets: `new_map(41)` -> `42`, generated C has `map new_map(...)` and `int main__new_map(int)`, lambda `!void` failure -> `fail`, lambda `?void` none -> `none`.
- `git diff --check`
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` passed, peak second-stage RSS: 1448 MB.
- Group 6 sweep: `GROUP6 pass=110 fail=0 total=110`.

medvednikov:
Addressed the latest review comment in amended commit `59aa7ebf08`.

Changes:
- Removed the broad `IError` / error-payload acceptance from `return_type_compatible` for `?T` returns, so direct `return error(...)` into an option return is rejected.
- Kept the existing constant-if option behavior used by group 6, where only the selected compile-time branch is considered, by limiting the error-tail allowance to constant-if selected branches.
- Added a regression for `fn f() ?int { return error('bad') }`.

Validation:
- Manual bad snippet now rejects before C generation with `cannot return `IError` as `?int``.
- Manual result-return sanity check still compiles for `fn f() !int { return error('bad') }`.
- `vlib/v/tests/conditions/ifs/if_expr_of_option_test.v` passes with rebuilt v3.
- `git diff --check`.
- `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v` passed, peak second-stage RSS: 1449 MB.
- Group 6 sweep: `GROUP6 pass=110 fail=0 total=110`.

Note: I started the full `vlib/v3/tests/type_checker_errors_test.v` run, but interrupted it after several minutes with no result and validated the new regression directly with a freshly built v3 instead.

medvednikov:
Addressed the ORM review comment in amended commit `e4d6132649`.

Changes:
- `transform_sql_expr_is_basic_sqlite` now validates table fields before keeping ORM SQL on the C fast path.
- `create table`, scalar `insert`, and simple `select` only use C lowering when every table field is supported by the current C ORM emitter (`string` or integer fields).
- Unsupported tables such as `time.Duration` alias fields stay on the fallback path instead of reaching `gen_sql_expr()` and panicking.
- Added a regression for `TestDurationAlias { duration time.Duration }` create-table fallback.

Validation:
- `./vnew test vlib/v3/tests/orm_null_text_codegen_test.v`
- `git diff --check`
- Exact self-host command passed before the push: `cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v`, peak second-stage RSS: 1784 MB.

Skipped broad/group test runs per latest request.

medvednikov:
Addressed the shared receiver/shared arg review in amended commit `460364689e`.

Changes:
- shared receiver diagnostics now run independently of receiver type compatibility
- shared locals/params are tracked by exact scope binding owner list, so shadowed locals no longer satisfy shared args while ordinary shared bindings still do
- parallel checker per-function state now resets/restores the shared owner list
- added focused regressions for the plain-value shared receiver, shadowed shared arg, and valid shared arg/receiver call

Validation:
- rebuilt temp V3: `./vnew -gc none -path "vlib|@vlib|@vmodules" -o /tmp/v3_shared_review vlib/v3/v3.v`
- manual snippets: `bad_shared_receiver_plain_value`, `bad_shared_arg_shadowed_local`, `good_shared_arg_receiver`
- `./vnew fmt -w vlib/v3/tests/review_checker_regressions_test.v vlib/v3/types/checker.v vlib/v3/types/checker_parallel.v`
- `git diff --check`

Skipped broad/group test runs per latest request.

medvednikov:
Addressed the unsigned inclusive for-post and ORM runtime-limit review in amended commit `880e77657c`.

Changes:
- unsigned inclusive loop post guards now check whether the original post would overflow/underflow the unsigned loop variable, instead of breaking whenever the loop variable equals the bound
- added regressions for body writes to the inclusive bound before `i++`/`j--`
- `select from T limit X` now stays on the SQLite C fast path only when `X` is a numeric literal or resolvable integer const; runtime limits fall back
- added a runtime-limit ORM fallback regression

Validation:
- rebuilt temp V3: `./vnew -gc none -path "vlib|@vlib|@vmodules" -o /tmp/v3_review_for_orm vlib/v3/v3.v`
- manual loop snippets: body writes to inclusive bound, plus unsigned max-bound guard
- manual ORM C-generation snippets: runtime `limit n` fallback, numeric/const limits still fast-path
- `./vnew fmt -w vlib/v3/transform/for.v vlib/v3/transform/orm.v vlib/v3/tests/for_multi_init_codegen_test.v vlib/v3/tests/orm_null_text_codegen_test.v`
- `git diff --check`

Skipped broad/group test runs per latest request.

medvednikov:
Addressed the SQLite ORM gate, constant-if option/error, and C-style pointer condition review in amended commit `4d9a428b84`.

Changes:
- SQLite ORM C lowering now requires the SQL receiver type to be `sqlite.DB`/`&sqlite.DB`; non-SQLite DB receivers stay on the fallback path
- removed the constant-if shortcut that allowed selected `error()` tails in `?T` contexts
- removed the remaining C-style `for` pointer-condition exemption for ordinary V source
- added focused regressions for non-SQLite ORM fallback, option-valued constant-if `error()`, and C-style pointer conditions

Validation:
- rebuilt temp V3: `./vnew -gc none -path "vlib|@vlib|@vmodules" -o /tmp/v3_review_sql_condition vlib/v3/v3.v`
- manual snippets: non-SQLite ORM fallback, SQLite ORM still fast-path, constant-if option/error rejection, C-style pointer-condition rejection
- `./vnew fmt -w vlib/v3/transform/orm.v vlib/v3/types/checker.v vlib/v3/tests/for_in_review_test.v vlib/v3/tests/type_checker_errors_test.v vlib/v3/tests/orm_null_text_codegen_test.v`
- `git diff --check`

Skipped broad/group test runs per latest request.

medvednikov:
Fixed the spawned fn-value closure capture review. Fn-value spawn wrappers now carry the lifted fn literal capture slots in the thread argument packet and restore them in the spawned thread before calling the function pointer; the wrapper cache/name includes the capture set so same-signature closures do not share the wrong restore body.

Validation:
- Built temporary V3: `./vnew -gc none -path "/Users/alexander/code/v6-v3-group6-current/vlib|@vlib|@vmodules" -o /tmp/v3_spawn_fn_value_review vlib/v3/v3.v`
- Manual runtime check for `spawn (cb)()` capturing fn value printed the captured value.
- `./vnew test vlib/v3/tests/spawn_args_test.v`

medvednikov:
Fixed the two new review items:

- Fixed-array data-arg expansion now binds the magic `index` per emitted slot when an `init:` expression reaches cgen.
- Omitted `&[]shared T` struct fields now stay pointer-zero-initialized instead of emitting an `Array` value initializer; the skip preserves comma/default state for later fields.

Validation:
- Built temporary V3: `./vnew -gc none -path "/Users/alexander/code/v6-v3-group6-current/vlib|@vlib|@vmodules" -o /tmp/v3_pr_review_array_struct vlib/v3/v3.v`
- Manual fixed-array arg init check printed `14`.
- Manual `&[]shared` default check printed `nil,0` and generated C omitted `.items = array_new(...)`.
- `./vnew test vlib/v3/tests/shared_array_codegen_test.v`

medvednikov:
Addressed the latest capturing-fn-literal review comments.

Changes:
- `check_return` now rejects direct capturing fn literals and local aliases before they escape through `return`.
- struct init field values now reject direct capturing callbacks and aliases before they can be stored in fields.
- kept the existing container-specific diagnostic path separate so array/map/append cases still report `capturing fn literal cannot be stored in a container`.
- added focused regression coverage for alias returns plus direct/aliased struct-field callbacks.

Focused validation only, per request to skip big runs:
- built temp V3: `./vnew -gc none -path "$(pwd)/vlib|@vlib|@vmodules" -o /tmp/v3_pr_review_capturing_escape vlib/v3/v3.v`
- manually verified rejection for direct return, alias return, direct struct field, alias struct field
- manually verified container storage still uses the container diagnostic
- `./vnew fmt -w vlib/v3/types/checker.v vlib/v3/tests/review_checker_regressions_test.v`
- `git diff --check`

medvednikov:
Addressed the latest map fixed-array return-copyback review.

Changes:
- return cleanup now emits all pending loop copybacks, so `for _, mut v in map[string][N]T` writes the local fixed-array copy back before returning from the function.
- returns with a value and pending loop copybacks are routed through the existing temp-return path, so the return expression is evaluated before copybacks/defers/lock cleanup.
- fixed the fixed-array map-value binding metadata so `v[0] = ...` is generated against the local fixed-array copy instead of treating `v` as pointer storage.
- added regression coverage for both `return` and `return value` after `v[0] = ...` in mutable map fixed-array iteration.

Focused validation only:
- `./vnew test vlib/v3/tests/map_for_mut_snapshot_codegen_test.v`
- manual reduced repro with `v[0] = ...; return` and `v[0] = ...; return 9`
- `./vnew fmt -w vlib/v3/gen/c/for.v vlib/v3/gen/c/stmt.v vlib/v3/tests/map_for_mut_snapshot_codegen_test.v`
- `git diff --check`

medvednikov:
Addressed the latest shared-argument and ORM select review comments.

Changes:
- `shared x` now recurses into `x` for shared-argument validation, so an explicit prefix cannot make a plain local satisfy a shared parameter.
- kept explicit `shared` on real shared bindings accepted.
- ORM table lookup now carries the resolved struct name, so `select from User` inside module `model` emits `sizeof(model__User)` / `model__User row` while keeping the SQL table name as `User`.
- added focused regression coverage for explicit shared plain locals and module-local ORM select C type qualification.

Focused validation only:
- built temp V3 with `./vnew -gc none -path "$(pwd)/vlib|@vlib|@vmodules" -o /tmp/v3_pr_review_shared_orm vlib/v3/v3.v`
- manual bad check: `take(shared s)` with plain `s := St{}` is rejected
- manual good check: `take(shared s)` with `shared s := St{}` is accepted and runs
- manual ORM project check: module-local `select from User` generated `sizeof(model__User)` and no `sizeof(User)`
- `./vnew fmt -w vlib/v3/types/checker.v vlib/v3/gen/c/orm.v vlib/v3/tests/review_checker_regressions_test.v vlib/v3/tests/orm_null_text_codegen_test.v`
- `git diff --check`

medvednikov:
Addressed the latest interface variadic forwarding review.

Changes:
- when the sole argument for an interface variadic slot is already the variadic array type, the transform now keeps the direct array path instead of repacking it as one interface element.
- retained the boxing path for other single array arguments to interface variadics, where the array is intended as one interface value.
- added a regression for `fn forward(items ...Displayable) { sink(items) }`, verifying the forwarded `[]Displayable` reaches the sink as two elements.

Focused validation only:
- `./vnew test vlib/v3/tests/variadic_call_codegen_test.v`
- `./vnew fmt -w vlib/v3/transform/fn.v vlib/v3/tests/variadic_call_codegen_test.v`
- `git diff --check`

medvednikov:
Addressed the selector-named `error` review.

Changes:
- `expr_is_error_call()` now requires a checker-resolved builtin `error` / `error_with_code` call for selectors; unresolved fallback remains only for bare builtin-style identifiers.
- return emission and return-expression stringification now use that stricter predicate too, avoiding the same selector-name trap in result/optional return paths.
- added a regression where `Stmt.error(int) IError` is passed to an `IError`-expected argument, returned as `IError`, and returned as a result error. The method is called and its message is preserved.

Focused validation only:
- `./vnew test vlib/v3/tests/option_arg_codegen_test.v`
- `./vnew fmt -w vlib/v3/gen/c/stmt.v vlib/v3/tests/option_arg_codegen_test.v vlib/v3/tests/ierror_payload_compat_codegen_test.v`
- `git diff --check`

Note: I also tried `./vnew test vlib/v3/tests/ierror_payload_compat_codegen_test.v`; it failed before the new regression on the existing `custom_error() !int` payload case with generated C assigning `CustomError` to an `int` result value. I moved the new regression to `option_arg_codegen_test.v`, which passes.

medvednikov:
Addressed the ORM insert type review in `vlib/v3/transform/orm.v`.

The SQLite fast-path gate now resolves the target table to its qualified type and requires the inserted scalar value type to match that exact resolved type, so `other.User` no longer passes `insert user into User` for a local `User` table. Added `test_orm_insert_requires_qualified_value_type_match` to cover the fallback path.

Targeted validation:
- `./vnew test vlib/v3/tests/orm_null_text_codegen_test.v`

medvednikov:
Addressed the capturing callback assignment review in `vlib/v3/types/checker.v`.

For ordinary assignment paths, non-local LHS targets now reject capturing fn literals and aliases instead of falling through to local capture tracking. Field/global-style assignment uses the stored/returned capture diagnostic, and indexed assignment uses the existing container-storage diagnostic.

Targeted validation:
- `./vnew test vlib/v3/tests/checker_assignment_review_test.v`

medvednikov:
Addressed the nested optional logical short-circuit review.

The nested optional `or {}` lowering now treats logical infix nodes specially: it stabilizes/evaluates the left side first and only lowers/evaluates the RHS optional path under `!lhs` for `||` or `lhs` for `&&`. It also unwraps propagation operands like `ok()?` to their optional source when they are being handled by the outer fallback.

Targeted validation:
- `./vnew test vlib/v3/tests/or_expr_transform_review_test.v`

medvednikov:
Addressed the lowered for-post/lock cleanup review.

Lowered C-style post bodies now use a loop continue label instead of injecting the post before each unlabeled `continue`. Unlabeled continues target that label, so existing branch cleanup releases active locks before the appended post runs. I also kept the unsigned-inclusive guard from hoisting side-effecting bounds by transforming the condition through the normal per-iteration condition path.

Targeted validation:
- `./vnew test vlib/v3/tests/for_multi_init_codegen_test.v`

medvednikov:
Addressed the map-mut fixed-array return ordering review.

Pending loop copybacks are now flushed before materializing non-void return values in `gen_return_with_defers()`, so a return expression that reads the backing map sees mutations made through the fixed-array loop copy. The normal return cleanup still runs afterward, preserving mutations performed inside the return expression itself.

Targeted validation:
- `./vnew test vlib/v3/tests/map_for_mut_snapshot_codegen_test.v`

medvednikov:
Addressed the by-value parameter address review.

The checker now only treats explicit `&T` arguments as compatible with by-value `T` parameters for the same syntactic forms that cgen rewrites: addressed struct literals, casts, calls, and range indexes. Lvalue addresses such as `take(&n)` are rejected instead of reaching C as an `int*` for an `int` parameter.

Targeted validation:
- `./vnew test vlib/v3/tests/review_checker_regressions_test.v`

PR #27804, opened 2026-07-15
fmt: migrate deprecated `json` module usage to `json2`

medvednikov:
The `json` module is deprecated in favour of the pure V `json2` module. This makes `v fmt` migrate its usage automatically, the same way vfmt already rewrites other deprecated constructs.

## Mapping

| Before | After |
|--------|-------|
| `import json` | `import json2` |
| `json.decode(T, s)` | `json2.decode[T](s)` |
| `json.encode(x)` | `json2.encode(x)` |
| `json.encode_pretty(x)` | `json2.encode(x, prettify: true)` |

Notes:
- The key transform is `decode`, where the target type moves from the first positional argument into the generic bracket `[T]`.
- `encode_pretty` maps to `encode(..., prettify: true)` because json2's own `encode_pretty` is itself deprecated.
- Aliased imports (`import json as j`) have the alias dropped and all call sites rewritten to `json2`; or-blocks and comments on the call are preserved.
- Selective imports (`import json { ... }`) are left untouched, since their call sites use bare, un-prefixed names that are not migrated.
- If a file already imports `json2`, the migrated `import json` collapses into it via the existing dedup.

## Implementation

- `call_expr` early-branches on the existing builtin call kinds (`.json_decode`, `.json_encode`, `.json_encode_pretty`) into a new `json2_migrate_call` helper. The decode branch guards against malformed (too-few-args) input, since vfmt runs on unchecked ASTs.
- `imp_stmt_str` rewrites `import json` (and `import json as x`) to `import json2`.

## Tests

- Adds `json_migrate_to_json2_input.vv` / `_expected.vv` covering decode, array-typed decode, encode, encode_pretty, and the propagation error path. This replaces the old `json_decode_fmt_keep.vv`, which is no longer idempotent.
- Migrates the incidental `json` usage in existing fmt/slow_test fixtures (`array_option_anon_struct_keep.vv`, the shebang import-ordering expected file, and the `issue_24521` project fixtures) to json2, so the keep / vfmt-everything checks stay green.

`fmt_test.v` and `fmt_keep_test.v` pass, and the migrated output compiles and runs correctly.


medvednikov:
Good catch — fixed in d0d8910006.

A selective import (`import json { decode }`) does resolve the bare `decode(...)` call to the `.json_decode` kind (parser rewrites the imported symbol to `json.decode` at fn.v:131-133), so `call_expr` was already rewriting it to `json2.decode[...]` while `imp_stmt_str` left the import as `import json { decode }` — producing code that references `json2` without importing it.

I chose to migrate the selective import too: every `import json` form (plain, aliased, and selective) now becomes a plain `import json2`. Dropping the alias/symbols is safe because json's whole public API (decode/encode/encode_pretty) is rewritten to qualified `json2.x` calls, so nothing needs to be in scope unqualified. Added a `json_migrate_selective_import` input/expected test; the migrated output compiles and runs.

medvednikov:
Fixed in 31c72b62d4.

Confirmed the bug: with `import json` + `import json2 { Any }`, vfmt emitted both `import json2` and `import json2 { Any }`, and the checker rejects that ("module `json2` was already imported").

Fix: the formatter now tracks whether the file already imports `json2` (in any form), and when migrating `import json` it drops the migrated import entirely instead of emitting a second `import json2`. The rewritten call sites use the qualified `json2.x` form, which a selective `import json2 { ... }` still provides (verified `import json2 { Any }` alone resolves `json2.encode(...)`), so nothing is lost. Works regardless of import order. Added a `json_migrate_merge_json2_import` input/expected test; the merged output compiles and runs.

medvednikov:
Fixed in 0b3d1d581e.

Confirmed both cases produced non-compiling output: with `import json` + `import json2 as j2`, vfmt dropped the migrated import but rewrote calls to `json2.encode` (undefined — only `j2` was in scope); same for a branch-local `$if { import json2 }`, where `json2` isn't available at the top-level call site.

Fix: only a **top-level** `json2` import now suppresses the migrated `import json` (branch-local ones are nested inside comptime-if statements, not direct `file.stmts`, so they're scanned out and a top-level `import json2` is emitted instead). And the migrated call sites are rewritten with that import's **local name** (`f.json2_prefix`) — its alias for `import json2 as j2` (→ `j2.encode(...)`), else `json2`. So the aliased case reuses `j2` with no duplicate import, and everything stays in scope and idempotent.

Verified both cases compile, run, and re-format to themselves; earlier cases (plain, selective, `json2 { Any }` merge) still hold. Added `json_migrate_aliased_json2_import` and `json_migrate_branch_local_json2_import` tests.

medvednikov:
Fixed in 7ef59d59db.

Confirmed: cJSON's pretty printer indents each depth level with a tab (`thirdparty/cJSON/cJSON.c:1831`), while `json2.EncoderOptions.indent_string` defaults to four spaces (`vlib/json2/encode.v:10`). The migration now emits `json2.encode(x, prettify: true, indent_string: '\t')` so the indentation shape is preserved.

Side-by-side on a nested struct confirms the indentation now matches (tab per level). One residual difference I can't close via `EncoderOptions`: cJSON writes `"key":\t value` (tab after the colon) whereas json2 writes `"key": value` (single space), and there's no option to control the colon separator — but the indentation, which is what the review flagged, is preserved. Updated the `json_migrate_to_json2` expected fixture accordingly.

medvednikov:
Fixed in 8514fe8953 — went with the first option (don't migrate `encode_pretty`).

You're right that `indent_string: '\t'` only closed the indentation gap, not the colon separator: cJSON writes `"key":\tvalue` (cJSON.c:1785-1788) while json2 hard-codes a single space (encode.v:496-498), and json2 has no option to change it. Adding public API to json2 just to serve the formatter felt out of scope, so I took the safe route.

vfmt now detects `json.encode_pretty` usage (a cheap AST walk via `walker.inspect`, only for files that import `json`) and leaves that file's json usage **entirely unmigrated** — its `encode_pretty`, `encode`, and `decode` calls and its `import json` all stay byte-identical. I skip the whole file (rather than just the encode_pretty call) so I never end up needing both `import json` and `import json2`; mixing migrated and unmigrated calls in one file is what would force that. Files without `encode_pretty` migrate to json2 exactly as before.

Verified: a file with `encode_pretty` (alone or alongside encode/decode) formats to itself byte-for-byte and still emits the original tab-after-colon output; encode/decode-only files still migrate. Dropped encode_pretty from the migration fixture and added a `json_encode_pretty_not_migrated` keep test; both fmt suites pass.

medvednikov:
Both fixed in 447aae4146.

**Branch-local json scan (fmt.v:190):** confirmed — `$if linux { import json }` + `json.encode_pretty` got the branch import rewritten to `json2` while the call stayed `json.encode_pretty`. The `imports_json` guard only looked at top-level `file.stmts`. Fixed by detecting the json import via `file.imports` (which does include branch-local imports), so the unmigratable-call scan runs for them too. The top-level-only `json2` detection used for import merging stays on `file.stmts`, as that distinction still matters. Verified the reviewer's exact `$if { import json; fn f(){...encode_pretty...} }` shape now stays unchanged.

**decode type-arg comments (fmt.v:2381):** confirmed the comment was dropped. I first tried carrying it into the bracket, but found that `json2.decode[/* c */ T]` — while it compiles — cannot be re-parsed by vfmt's own `.parse_comments` mode (`unexpected name, expecting ]`), so that output isn't formattable. So I took the "skip migrating that call" route: a file whose `json.decode` has comments on its type argument is left entirely unmigrated (same mechanism as encode_pretty), keeping the call and its comment intact. Verified the comment is preserved and the file still compiles/runs.

Generalized the "leave unmigrated" detection to `file_has_unmigratable_json_call` (encode_pretty OR decode-with-type-arg-comments). Added a branch-local encode_pretty keep test and a decode-type-comment input/expected test; both fmt suites pass.

medvednikov:
Fixed in 7f3903f5ec.

Confirmed the bug: `import json; import json2 as _; json.encode(x)` produced `import json2 as _` + `_.encode(x)`, which doesn't compile.

I checked what valid output is even possible:
- `import json2` + `import json2 as _` → `A module `json2` was already imported`, so I can't emit a plain `import json2` alongside the `_` one.
- `import json2 as _` alone → `undefined ident: json2`, confirming `_` provides no usable qualifier.

Since there's no way to produce valid `json2.x` calls here, I took the skip route (as you suggested): a top-level `import json2 as _` is treated as providing no usable qualifier, and the file's json usage is left entirely unmigrated via the existing keep_json_unmigrated path — `import json`, `import json2 as _`, and the `json.*` calls all stay as-is. Verified it's now unchanged, idempotent, and compiles; usable aliases (`json2 as j2`) and plain json still migrate as before. Added a `json_migrate_blank_json2_alias` keep test.

medvednikov:
Fixed in 66cc9d1213.

Confirmed the difference: legacy `json.encode('😀')` == `"😀"` (escaped), while `json2.encode` defaults `escape_unicode: false` and emits raw unicode.

The migration now emits `json2.encode(x, escape_unicode: true)`. I verified this is **byte-identical** to legacy `json.encode` across ascii, accented (`é`), CJK (`한`), emoji/surrogate-pairs (`😀`), control escapes, and structs with unicode fields — all IDENTICAL. Updated the migration expected fixtures; existing json2 calls in the sources (e.g. `j2.encode(42)`) are correctly left untouched, and both fmt suites pass.

(Went with the `escape_unicode: true` option rather than skipping, since unlike encode_pretty this one *can* reproduce the legacy bytes exactly.)

medvednikov:
Both addressed in 3763cc5d26.

These two are different in kind from the earlier reviews: the divergence depends on the payload's runtime type (`time.Time` → unix vs RFC3339; `@[raw]` → normalized vs raw slice), and json2 has **no encoder/decoder option** to restore the legacy bytes for either — so unlike escape_unicode, they can't be preserved, only skipped. And vfmt runs **without the checker** (both calls are commented out in vfmt.v), so it has no type for an `encode(x)` payload.

What it *can* see reliably is struct field declarations. So I extended the existing "leave unmigrated" scan: if the file declares a struct with a `time.Time` field or a `@[raw]` field, the whole file keeps `import json` and all `json.*` calls byte-identical. Verified both cases now stay unmigrated and preserve the legacy output — `"reg_date":1784151346` (unix) and `{"Y":123}` (normalized) — while plain structs still migrate.

Honest limitation: this only sees **file-local** struct declarations, since vfmt doesn't parse imported modules. A payload type defined in another module (or built via an untyped expression) can't be detected without type resolution — closing that fully would mean running the checker in vfmt, which is a bigger call. Flagging it in case you'd prefer that route or want to narrow the feature further. Added `json_migrate_time_field` and `json_migrate_raw_field` keep tests; both fmt suites pass.

medvednikov:
**vfmt-off (fmt.v:498)** — fixed in cb49a3f85c. Confirmed the bug: `import json` + a `// vfmt off` block calling `json.encode` formatted to `import json2` while the verbatim off-block kept `json.encode`, leaving no json import. Since a `// vfmt off` comment can sit on either side of the rewrite, the file's json usage is now left unmigrated whenever the source contains a `// vfmt off` region. Added a keep test; verified it stays byte-identical and compiles.

**Cross-file payload types (fmt.v:245)** — you're right that the struct-field guard only sees file-local declarations, and I want to be straight about why: vfmt runs **without the checker** and does **not parse imported modules**, so for `json.decode(model.User, s)` the target is an uninspectable placeholder, and for `json.encode(u)` the payload type is never known at all. json2 also has no option to reproduce the legacy time.Time/@[raw] bytes, so these can only be skipped, not preserved.

That leaves three real options, and it's a judgement call about this feature rather than something I can quietly patch:
1. **Best-effort (current):** file-local guard catches the common cases; imported/unknown payloads are a documented limitation.
2. **Strict:** only migrate provably-safe payloads — which, with no type info, means *not migrating encode at all* and skipping every imported decode target. That removes most of the feature's value.
3. **Type resolution in vfmt:** run the checker / resolve imports so payloads are inspected precisely — a larger change, and vfmt must still format code that doesn't type-check.

I've kept it on (1) since the intent is a migration aid for a deprecated module and (2) guts it, but I'm happy to implement (2) or (3) if that's the preferred direction — it's really a maintainer call on how much fidelity vs. reach this should have.

medvednikov:
Fixed in 98128518d6.

Confirmed: `fn f(json2 Foo, u User) string { return json.encode(u) }` migrated to `json2.encode(...)`, which binds to the `json2` parameter and doesn't compile.

I went with the skip route: the "leave unmigrated" scan now also flags a **declaration** of the qualifier name (`f.json2_prefix` — `json2`, or an existing alias like `j2`) — a function parameter or receiver, a local `:=` declaration, a for-in loop variable, or a const. Checking declarations rather than uses avoids false-positives on legitimate existing `json2.*` calls (whose qualifier isn't a standalone ident). Verified the reviewer's exact case now stays `import json` + `json.encode` (idempotent, compiles), the alias variant (`import json2 as j2` + param `j2`) is caught too, and non-shadowed files still migrate. Added a `json_migrate_shadowed_qualifier` keep test; both fmt suites pass.

medvednikov:
All three fixed in 5d5497e011.

**Custom JSON hooks (fmt.v:258):** the json2 hook methods are `to_json`/`json_str` (encode; custom.v) and `from_json_string/number/boolean/null` (decode). The scan now flags a file-local method declaration named `to_json`, `json_str`, or `from_json_*` and leaves the file unmigrated. Verified `json.encode(Custom{...})` with a `to_json` method stays on json (and still emits the field-based `{"a":1}`, not the hook's `"custom"`).

**If-guard shadowing (fmt.v:273):** good catch — and the walker made it non-obvious: `IfBranch.children()` yields only the branch stmts, not the `cond`, so an `IfGuardExpr` visitor never fires. I instead inspect the branch conditions from the enclosing `IfExpr` and match `IfGuardExpr.vars` against the qualifier. Verified `if json2 := opt() { ... json.encode(u) }` now stays unmigrated.

**Branch-local selective json2 (fmt.v:170):** right — a plain branch `import json2` dedups against the added top-level one, but a selective/aliased one doesn't and trips the checker's already-imported check. I detect a branch-local json2 import (in file.imports, not a top-level file.stmt, matched by pos) that is selective or aliased, and skip migration then; a plain branch import still migrates as before. Merging wasn't safe since it could drop selective symbols the branch uses.

Added three keep tests; both fmt suites pass.

medvednikov:
Both fixed in b5262b4ca9.

**C-style for initializer (fmt.v:317):** right — same walker gap as the if-guard case: `ForCStmt.children()` yields only the body, not `init`, so `for json2 := 0; ...` was never seen. I now inspect `ForCStmt.init` directly (factored the `:=`-declares-name check into a small helper shared with the plain assignment case). Verified `for json2 := 0; json2 < 1; json2++ { return json.encode(u) }` now stays unmigrated.

**Aliased time.Time (fmt.v:277):** good catch — the guard compared the raw field symbol name, so `type MyTime = time.Time; reg_date MyTime` reported `MyTime` and slipped through. Switched to `final_sym(field.typ)`, which follows the alias chain to `time.Time`. Verified the aliased-field file now stays on json (preserving the legacy timestamp bytes) while plain non-time structs still migrate.

Added `json_migrate_forc_shadow` and `json_migrate_aliased_time_field` keep tests; both fmt suites pass. (final_sym only resolves file-local aliases — an alias imported from another module remains part of the documented cross-module limitation.)

medvednikov:
All three fixed in 78cdc612ee.

**Branch-local json2 ordering (fmt.v:202):** right — my `is_plain` exemption was unsafe. A plain branch `import json2` emitted *before* the top-level `import json` goes to branch_processed_imports, so the later migrated `import json2` isn't deduped (the check only sees global_processed_imports) → two json2 imports. Dropped the exemption: any branch-local json2 import now leaves the file unmigrated, regardless of symbols or order. (The old plain-branch input/expected test becomes a keep test, plus a new before-import keep test.)

**Unmigratable call in an if condition (fmt.v:325):** confirmed — `if json.encode_pretty(u) == '' {}` was missed because the walker doesn't descend into `IfBranch.cond`. The IfExpr case now sub-walks each branch condition (reusing the same visitor via `walker.inspect(branch.cond, data, ...)`), which reaches nested calls and the if-guard vars. Verified if/else with empty else-cond doesn't crash, and formatting real vlib files (checker.v, os.v) is clean.

**Lambda param shadowing (fmt.v:303):** confirmed — `ast.LambdaExpr.params` (a `[]Ident`) wasn't checked. Added a LambdaExpr case to the scan. Verified `|json2| json.encode(u)` now stays unmigrated.

Added three keep tests; both fmt suites pass.

medvednikov:
Both fixed in 85575fa669.

**json2 alias collision (fmt.v:621):** confirmed — `import rand as json2` + `import json` would migrate to a second `import json2` sharing the `json2` local name (the checker's duplicate-import check compares aliases, per checker.v:531-539) and resolve calls to the wrong module. I now scan for an import whose alias equals `f.json2_prefix` but whose source isn't `json2`, and skip migration then. This also covers a submodule imported as `json2` (e.g. `import x.json2`). Verified `import rand as json2` + `json.encode` stays unmigrated and compiles.

**Comptime-for shadowing (fmt.v:355):** confirmed — I handled runtime `ForInStmt` but not `ast.ComptimeFor.val_var`. Added a ComptimeFor case to the scan. Verified `$for json2 in User.fields { ... }` around a `json.encode` now stays unmigrated.

Added two keep tests; both fmt suites pass. (The alias-collision check keys off the resolved prefix, so it works for an aliased existing json2 too.)

medvednikov:
Both fixed in 20b7f3128b.

**Loop headers (fmt.v:395):** confirmed — same walker gap as the if-condition case: For*.children() yields only the body. The scan now sub-walks the header exprs — `ForStmt.cond`, `ForInStmt`'s iterable (`cond`) and range `high`, and `ForCStmt`'s `init`/`cond`/`inc` (guarded by has_init/has_cond/has_inc). The init sub-walk also subsumes the earlier C-for loop-variable shadowing check via the AssignStmt case. Verified `for json.encode_pretty(u) != '' {}`, the C-style cond, and a for-in iterable all leave the file unmigrated; normal loops still migrate, and formatting real files (parser.v, array.v, os.v) is clean.

**Merged-import comment (fmt.v:603):** confirmed — the merge drop path only re-emits next_comments, so `import json // legacy` + an existing `import json2` lost the comment. I went with 'avoid merging': when the merged json import carries same-line comments, the file is left unmigrated, so `import json // legacy` stays intact. (The non-merge path already preserves the comment by emitting it after `import json2`, so only the merge case needed this.)

Added two keep tests; both fmt suites pass.

medvednikov:
Fixed in 74eeac34ff.

Confirmed — same walker gap as the loop-header/if-condition cases: `StructField.default_expr` isn't part of `Node.children()`, so `payload string = json.encode_pretty(Default{})` was never scanned. The StructDecl branch now sub-walks each field's `default_expr` (when `has_default_expr`), reusing the same visitor. Verified the field-default file now keeps `import json` and its `json.encode_pretty` call, while structs with plain defaults still migrate. Added a `json_migrate_field_default` keep test.

(Side note: `json.encode_pretty` in a field default hits a separate pre-existing codegen limitation — the fixture is a formatting-only `.vv`, which the fmt tests never compile, so that's orthogonal to this scan fix.)

medvednikov:
Both fixed in e8bf818ed9.

**Implied json imports (fmt.v:245):** confirmed the mismatch. Investigating it surfaced a wrinkle: when `json` is omitted, whether `json.encode` even gets the `.json_encode` call kind depends on whether `json` resolved as a module in the parse context — so the call may or may not rewrite, and either way pairing it with an implied-`json`-emitted-verbatim import breaks. I took the 'suppress the rewrite' route: an implied `json` import leaves the file's json usage as-is. Each pass stays compilable — vfmt first materializes the implied `import json` (keeping json), and once it's an explicit import a later pass migrates it normally. (It's inherently two-pass, so there's no single-pass fixture; verified manually that every pass compiles.)

**module json2 (fmt.v:215):** confirmed — `import json2` into `module json2` is rejected (checker.v:503-505, cannot import a module into a module with the same name). Added `cur_module_is_json2 := file.mod.name == 'json2'` to the skip conditions. Verified a `module json2` file that imports legacy json now stays unmigrated; added a keep test.

Both fmt suites pass.

medvednikov:
Both fixed in 5f9702f98e.

**spawn/go calls (fmt.v:313):** confirmed — `SpawnExpr.call_expr`/`GoExpr.call_expr` aren't in `Node.children()`. Added SpawnExpr/GoExpr cases that sub-walk the spawned call. One subtlety I had to debug: passing the bare `CallExpr` to `walker.inspect` silently walked nothing, because it converts only one level to `ast.Node` (my other sub-walks pass `ast.Expr`/`ast.Stmt`); wrapping it as `ast.Expr(call_expr)` fixed it (confirmed with a walker probe). Verified `spawn json.encode_pretty(u)`, a nested `spawn worker(json.encode_pretty(u))`, and the `go` form all stay unmigrated.

**time.Time in containers (fmt.v:328):** the literal examples (`[]time.Time`, `map[string]time.Time`) were already caught by the substring match, but the real gap was an *aliased* element — `tagged []MyTime` where `type MyTime = time.Time` — since final_sym only unaliases the top level. Replaced the check with a recursive `type_mentions_time` that follows aliases and descends into array/fixed-array/map/chan element/key/value types, matching `time.Time` exactly at each level. Verified `[]MyTime`, `map[string]MyTime`, and the direct container forms all keep the file on json; plain structs still migrate.

Added two keep tests; both fmt suites pass.

medvednikov:
Fixed in 80b28922b9.

Confirmed — same walker-gap family as the loop/if/spawn cases: `AssertStmt.children()` exposes only the asserted expression, not `AssertStmt.extra` (the message). The scan now sub-walks `extra`. Verified `assert ok, json.encode_pretty(u)` now keeps `import json` and its calls, while asserts with a plain message still migrate. Added a `json_migrate_assert_message` keep test; both fmt suites pass.

medvednikov:
Fixed in 64f60b96ac.

Confirmed — same walker-gap family: `ArrayInit.children()` exposes only element expressions, not `len_expr`/`cap_expr`/`init_expr`. The scan now sub-walks those (guarded by has_len/has_cap/has_init). Verified `[]string{len: 1, init: json.encode_pretty(u)}` and a `cap:` variant keep `import json`; plain array inits (e.g. `init: json.encode(u)`) still migrate. Added a `json_migrate_array_init_field` keep test; both fmt suites pass.

medvednikov:
Both fixed in ac3bef74ba.

**Generic wrappers (fmt.v:310):** confirmed — `box Box[time.Time]`'s final symbol is the GenericInst `main.Box[time.Time]` (kind generic_inst), and time.Time lives only in its concrete type args. `type_mentions_time` now recurses into `GenericInst.concrete_types`. Verified `Box[time.Time]` and `[]Box[time.Time]` keep the file on json, while `Box[int]` still migrates. (This is the conservative reading — even a `Box[T]` that doesn't use `T` in a field is kept, which is safe.)

**dump() (fmt.v:493):** confirmed — same walker gap, `DumpExpr.expr` isn't in `Node.children()`. The scan now sub-walks it. Verified `dump(json.encode_pretty(u))` keeps `import json`.

Added two keep tests; both fmt suites pass.

medvednikov:
All three fixed in 2ff008479f.

**Top-level fn json2 (fmt.v:376):** confirmed — `fn json2()` + `import json` migrated to `import json2` and collided. The shadowing scan now also treats `!decl.is_method && decl.short_name == s.json2_prefix` as a collision (methods are namespaced by receiver, so only top-level fns count). Verified it stays unmigrated.

**Branch json import dedup (fmt.v:751):** confirmed — a branch-local json import migrated into branch_processed_imports isn't deduped against a top-level migrated `import json2` in the global list (the duplicate shows up when the branch import is emitted first). Rather than reworking the shared dedup, I left the file unmigrated whenever a branch-local json import exists — symmetric with the branch-local json2 handling already there. Verified `import json as old` + `$if { import json ... }` stays on json.

**SQL DSL (fmt.v:505):** confirmed — `SqlExpr`/`SqlStmt` have no `Node.children()`. The scan now sub-walks SqlExpr's db/where/order/limit/offset exprs plus joined sub-queries (recursively), and SqlStmt's db expr and per-line where/update exprs. Verified `sql db { select from User where name == json.encode_pretty(u) }` keeps `import json`.

Added three keep tests; both fmt suites pass.

medvednikov:
Fixed in 4267a0db25.

Confirmed — a root `time.Time` payload has no struct field for the scan to see, so it migrated and changed the contract. The scan now checks the call's payload type directly:

- **decode:** the target is a known TypeNode, so I run `type_mentions_time` on it — covers `json.decode(time.Time, s)`, `[]time.Time`, `map[string]time.Time`, aliases, generic wrappers, etc.
- **encode:** a struct-literal or cast argument carries a type at fmt time, so `json.encode(time.Time{})` / `json.encode(time.Time(x))` are caught too.

Verified all of those now keep `import json`, while `decode(User, s)` still migrates.

The one residual is the reviewer's `json.encode(time.utc())` — a runtime value whose type vfmt can't resolve without the checker. That's the same cross-module/root-value type-resolution limit I flagged earlier (encode of an untyped expression); noted it in the test comment. Added a `json_migrate_root_time_type` keep test; both fmt suites pass.

medvednikov:
Both fixed in 082f304174.

Both are payload-type serialization differences, so I generalized the existing time.Time payload guard (`type_mentions_time` → `type_needs_legacy_json`) to also match `rune`, `f32`, and `f64` — recursively through array/map/chan and generic wrappers, following aliases — and it already feeds the struct-field, decode-target, and encode-literal checks.

- **rune (encode_rune):** string vs u32 number — a type-level difference, always applies. `struct Thing { ch `é` }` now keeps `import json`.
- **floats (json_float_to_raw_string):** `null` for NaN/Inf vs `nan`/`inf`. Only non-finite *values* differ, but vfmt can't see runtime values, so I skip any f32/f64 payload — the reviewer's `Stats{value: math.nan()}` is a struct field, so it's caught. (The other option, fixing json2 to emit `null` for NaN/Inf — which would also make its output valid JSON — is a json2 change; I kept this vfmt-only and conservative, consistent with the rest of the PR, but happy to do the json2 fix instead if you'd prefer to keep float payloads migratable.)

Added rune/float keep tests; both fmt suites pass. Verified int/string/bool payloads still migrate.

medvednikov:
Fixed in 6421d48d21.

Confirmed — my earlier ArrayInit case sub-walked len/cap/init but not `update_expr` (the `...a` spread), which is likewise outside `Node.children()`. Added `if ai.has_update_expr { walker.inspect(ai.update_expr, ...) }`. Verified `return [...parts(json.encode_pretty(u)), 'x']` keeps `import json`; a plain spread (`[...base, json.encode(u)]`) still migrates. Added a `json_migrate_array_spread` keep test; both fmt suites pass.

medvednikov:
@codex review

medvednikov:
Fixed in 7aeac60eb3.

You're right about the risk. Probing the parser directly, an aliased `old.encode` can come back as `kind=unknown` (not `json.encode`) — the tagging depends on the alias being resolved to the canonical module, which isn't guaranteed in every parse context. Since `imp_stmt_str` unconditionally rewrites `import json as old` to a plain `import json2` (dropping the alias), a missed call tag would leave `old.encode(...)` with no `json` import.

Went with the 'keep unmigrated' option: the scan now detects a json import whose alias isn't the default `json` and leaves the file unmigrated. Verified `import json as old` + `old.encode`/`old.decode` now stays byte-identical, while plain `import json` and unaliased selective `import json { encode }` still migrate. Added a `json_migrate_aliased_json_import` keep test; both fmt suites pass.

medvednikov:
Both fixed in f727bba637.

**Non-call or-blocks** — confirmed `Node.children()` exposes an or-block only for `CallExpr` (`children << Expr(call_expr.or_block)`); on identifier / index / selector / prefix / infix / comptime-call/selector expressions the `or_expr`/`or_block` field is never reached. Added a `walk_or_block` helper that sub-walks the block's statements (guarding `kind != .absent`) and wired it into cases for each of those expr types, plus `SqlExpr`/`SqlStmt`'s `or_expr`. Verified `opt or { json.encode_pretty(u) }` and `m[k] or { json.encode_pretty(u) }` now stay on `import json`, while a migratable `opt or { json.encode(u) }` still becomes `json2.encode(u, escape_unicode: true)` inside the preserved or-block.

**Sum type variants** — `type_needs_legacy_json` now recurses into `ast.SumType.variants`, alongside the existing Array/Map/Chan/GenericInst recursion. Verified `type Val = int | rune` as an encode payload field and `type Val = f64 | string` as a decode target both keep the file unmigrated.

Added `json_migrate_or_block` and `json_migrate_sumtype_variant` keep tests; both fmt suites pass.

medvednikov:
Both fixed in 464c76bdf7.

**Recursive sum types** — `type_needs_legacy_json` now threads a visited `[]int` of type indexes and bails when it re-sees one. Split into a 2-arg entry point plus a `_rec` helper so the four call sites are unchanged. The visited guard runs *after* the `time.Time`/`rune`/`f32`/`f64` name check, so a repeated index can never hide a legacy-sensitive leaf — migration decisions for non-recursive types are identical. Verified `type Val = []Val | rune` (decode target) no longer hangs and stays on `json`, while `type Node = []Node | int` terminates and still migrates to `json2.decode[Node]`.

**Comptime-call args** — the `ComptimeCall` case now sub-walks each `args[i].expr` in addition to the or-block, since `comptime_call` prints them via `call_arg_spread_str` but `Node.children()` doesn't expose them. Verified `$for method in App.methods { a.$method(json.encode_pretty(u)) }` keeps the file on `import json`.

Added `json_migrate_recursive_sumtype` and `json_migrate_comptime_call_arg` keep tests; both fmt suites pass.

medvednikov:
Both fixed in 155c519088.

**Option sum variants** — `type_needs_legacy_json` now checks `typ.has_flag(.option)` before `final_sym` (which strips the flag), so a `?int` variant no longer descends as plain `int`. This also covers plain option struct fields. Verified `type Elem = int | ?int` in a payload keeps the file on `import json`.

**Typed container literals** — confirmed the reviewer's point that without the checker the container's own `typ`/`elem_type`/`value_type` are frequently unresolved at fmt time: `json.encode([time.Time{}, ..])` and `json.encode({'x': rune(65)})` both migrated before this fix. But the element/value **literals and casts** do carry types. Replaced the flat StructInit/CastExpr chain with a recursive `payload_expr_needs_legacy_json` that walks StructInit / CastExpr / ArrayInit / MapInit and their element/key/value sub-expressions. Verified all three forms now stay on legacy json:
- `json.encode([]time.Time{})` (element type on ArrayInit.typ)
- `json.encode([time.Time{}, time.Time{}])` (only on the element struct-literals)
- `json.encode({'x': rune(65)})` (only on the value cast)

…while `json.encode([]int{})` and `json.encode(User{})` still migrate. Added `json_migrate_option_variant` and `json_migrate_container_literal_encode` keep tests; both fmt suites pass.

medvednikov:
All three fixed in 7b8790ba26.

**Struct/map spread-update exprs** — `StructInit.update_expr`/`MapInit.update_expr` are outside `Node.children()` (only fields/keys/values are), but `struct_init`/`map_init` print them unchanged. Added visitor cases that sub-walk `update_expr`. Verified `User{ ...base(json.encode_pretty(u)), name: 'x' }` and `{ ...parts(json.encode_pretty(u)), 'x': 'y' }` stay on `import json`, while a migratable `...base(json.encode(u))` still migrates.

**SQL query-data items** — `SqlQueryDataExpr` (the `{ if cond { name == .. } }` dynamic-ORM form, parsed at `.lcbr`) has no `Node.children()` case. Added a recursive `walk_sql_query_data_items` (leaf `.expr`, branch `.cond`, nested `branch.items`) and a visitor case. Verified a `json.encode_pretty` leaf inside `where { if .. { name == json.encode_pretty(u) } }` keeps the file unmigrated.

**`__global json2` shadowing** — added a `GlobalDecl` case mirroring the existing `ConstDecl` name check, so a `__global json2 ...` under `-enable-globals` leaves the file unmigrated (otherwise `json2.encode` would bind to the global value, not the module). Verified it stays on `import json`.

Added `json_migrate_struct_map_spread`, `json_migrate_sql_query_data`, and `json_migrate_global_json2` keep tests; both fmt suites pass.

medvednikov:
Fixed in ea094be3a5.

Confirmed `SqlExpr.joins` (`[]JoinClause`) was the one query part the scan skipped — it walked db/where/order/limit/offset and sub_structs but not the joins, while `sql_expr` prints each `join.on_expr` via `f.expr`. The `SqlExpr` case now iterates `sqlexpr.joins` and sub-walks each `join.on_expr`.

Verified `join Department on Department.name == json.encode_pretty(u)` keeps the file on `import json`, while the same predicate with a migratable `json.encode(u)` still migrates to `json2.encode(u, escape_unicode: true)`. Added a `json_migrate_sql_join` keep test; both fmt suites pass.

medvednikov:
Both fixed in cde4f18e1b.

**Struct payload recursion** — `type_needs_legacy_json` now has an `ast.Struct` case that recurses into `sym.info.fields`, so a struct payload with no file-local StructDecl is caught. The visited-index guard added earlier for recursive types also protects this (`struct Node { next &Node }`). Verified `json.encode(struct { t time.Time }{ .. })` keeps the file on `import json`, while a nested all-safe struct still migrates.

**omitempty composites** — confirmed the encoder asymmetry: legacy json drops an `@[omitempty]` field equal to the type default (gen/c/json.v:1025-1047), but json2's `check_not_empty` only guards string/number/array/map/option/pointer and returns `true` (always emit) for bool / struct / sum type / enum (encode.v:582-622). Added `struct_field_attr_needs_legacy_json` covering `@[raw]` and `@[omitempty]`-on-`.bool`/`.struct`/`.sum_type`/`.enum`, used by both the struct-decl scan and the new struct recursion. Verified:
- `@[omitempty]` on bool / struct / sum type → kept
- `@[omitempty]` on string / int / array → still migrates (json2 guards those identically)

Added `json_migrate_anon_struct_payload` and `json_migrate_omitempty` keep tests; both fmt suites pass.

medvednikov:
I dug into this one and don't think it's currently reachable — sharing the evidence in case I'm missing something.

**The inline `is X as name` form does not parse in the current compiler.** Every variant I tried errors at parse time (so before any migration runs):

```
if a is Dog as d { ... }        // error: unexpected keyword `as`, expecting `{`
if w.a is Dog as d { ... }      // same
if get() is Dog as d { ... }    // same
match a { Dog as x { ... } }    // same
```

The only reference to the syntax in the tree is an aspirational parser comment (`expr.v:780`, `// return early \`if bar is SumType as b {\``); the early-return there leaves the `as` token unconsumed, and nothing downstream reads it, so the if-branch parser rejects it. There are zero functional uses of `is <Type> as <name>` anywhere in vlib, and `doc/docs.md` doesn't document it. Because such a file fails `v fmt` outright, it can never reach the migration path — worst case is a parse error, not a silent mis-migration.

**The functional equivalent is already covered.** When the smartcast alias is written as a statement, it's a normal `:=` declaration:

```v
if a is Dog {
    json2 := a as Dog   // AssignStmt declaring `json2`
    return json.encode(u)
}
```

The existing `AssignStmt` shadowing check (`assign_declares_name`) already catches that — verified the file stays on `import json` and `json.encode` is not rewritten.

Given the inline form is unparseable and its functional counterpart is already guarded, I held off adding a guard, since it would be dead code targeting an AST shape that can't currently exist (and whose eventual representation is unknown). Happy to add one the moment the syntax lands (or if you can point me at a snippet that actually parses).

medvednikov:
@codex review

medvednikov:
Good catch — fixed in 145e403c71.

Root cause confirmed: `call_expr` gated migration on `!keep_json_unmigrated`, which defaults false and is only *computed* when the file imports json. So a file with no legacy `import json` fell straight through and migrated any call the parser tagged `.json_decode`/`.json_encode` — and since the parser keys that tag on the literal qualifier name, `import json2 as json` calls carry it too. `json2_migrate_call` then read the decode type from the legacy first argument (the string `s`), producing `json.decode[s]()` from `json.decode[User](s)`.

Replaced the negative gate with a positive `migrate_json_calls` flag set to `imports_json && !keep_json_unmigrated`, so calls are rewritten only when the file actually imports the legacy `json` module being migrated. (`imp_stmt_str` was already correctly gated on `source_name == 'json'`, so imports were never the problem — only calls.)

Verified:
- `import json2 as json` → `json.decode[User](s)` / `json.encode(u)` stay byte-identical (no corruption).
- `import rand as json` → `json.seed(..)` untouched.
- Real `import json` → still migrates to `json2.decode[User](s)` / `json2.encode(u, escape_unicode: true)`.

Added a `json_migrate_json2_as_json` keep test; both fmt suites pass.

medvednikov:
Both fixed in 32ca168fb9.

**asm operands** — added an `AsmStmt` scan case that sub-walks each input/output `AsmIO.expr`. Worth flagging: while confirming the fix I found the problem is a bit wider than encode_pretty. `asm_ios` prints operands verbatim via `${io.expr}` (the AST str()), *not* through `call_expr`, so the migration rewrite never reaches them at all — a plain migratable `json.encode(u)` in an operand would have its import migrated to `json2` while the operand stayed `json.encode(u)`. So the scan there uses a dedicated `json_asm_operand_scan_visit` that keeps the file unmigrated on *any* `json.encode`/`json.decode`/`json.encode_pretty` in an operand, not just the losslessly-unrewritable ones. Verified both `encode_pretty` and plain `encode` in an operand keep the file on `import json`, while an asm block with no json still lets the file migrate its other json calls.

**fixed-array omitempty** — added `.array_fixed` to the omitempty kind list. json2's `$array` emptiness check is a `.len` test that doesn't match a fixed array, while legacy json compares `[2]int` to its default and omits it when equal — so `codes [2]int @[omitempty]` now keeps the file unmigrated, while a dynamic `[]int @[omitempty]` still migrates (json2 guards it the same way).

Added `json_migrate_asm_operand` and `json_migrate_fixed_array_omitempty` keep tests; both fmt suites pass.

medvednikov:
Both fixed in f36b1a60ca.

**Selective-import decode targets** — confirmed: `import json { decode }` + `decode(time.Time, s)` was migrating to `json2.decode[time.Time](s)`, changing the legacy root contract. Since the parser leaves the target as a name expression (not a TypeNode) for selective imports, added `decode_target_expr_needs_legacy_json`: it builds the name from the ident/selector (`time.Time`), resolves it via `table.find_type` (with a `${cur_mod}.${name}` fallback for a bare local type), and runs `type_needs_legacy_json`; an unresolved name keeps the file unmigrated. Verified `decode(time.Time, s)` now stays on `import json { decode }`, while the existing `decode(User, s)` selective test still migrates to `json2.decode[User](s)` (resolution works at fmt time).

**Generic parent omitempty** — the `GenericInst` check only walked concrete type args, so `Box[bool]{}` with `struct Box[T] { val T @[omitempty] }` migrated. Now it also recurses the base struct's fields; a generic-parameter field resolves to kind `.any`, which I added to the omitempty-sensitive set (its concrete type is unknown at fmt time, so conservatively keep). Verified `json.encode(Box[bool]{})` stays on legacy json, `Box[time.Time]{}` still keeps, and a plain `Box[string]{}` without omitempty still migrates.

Added `json_migrate_selective_decode_time` and `json_migrate_generic_omitempty` keep tests; both fmt suites pass.

medvednikov:
Fixed in e0608559bd.

Confirmed by instrumenting the helper: for `import models` + `json.encode(models.Event{})`, the payload resolves to `ast.Type(32)` with `name=models.Event kind=placeholder mod=models nfields=-1` — i.e. a bare placeholder, not an `ast.Struct`, so the field recursion never ran and it fell through to `false` and migrated.

Added a `sym.kind == .placeholder` check right after the time.Time/rune/float name check in `type_needs_legacy_json_rec`: an unresolved/imported payload is treated as legacy-sensitive and kept on `json`. This propagates to every caller — encode payloads, decode targets, and local structs that embed an imported-type field. A fully resolved local struct still migrates (its `kind` is `.struct` with visible fields).

Verified:
- `json.encode(models.Event{})` (imported, hidden time.Time) → stays on `import json`.
- `json.decode(models.Event, s)` → kept too.
- `json.encode(User{})` for a local struct → still migrates.

Added a `json_migrate_imported_payload` keep test (uses `http.Cookie`, a real imported type with a hidden `time.Time` field); both fmt suites pass.

medvednikov:
Fixed in 155dfac7b2.

You're right that the in-file hook scan can't see a hook whose method lives in a sibling file. I went with the 'inspect module files' option rather than the blanket 'keep all local payloads', since the latter would disable migration for almost every real json file.

Added `module_has_sibling_json_hook`: for a file that imports the legacy `json`, it scans the other `.v` sources in the same directory for a `to_json`/`json_str`/`from_json_*` method declaration (a lightweight line-level check — `fn (recv) hook(...)` — no parsing needed) and keeps the file unmigrated if one is found. This mirrors the existing coarse in-file hook check, just at module scope. It runs last in the keep OR-chain, so the directory I/O only happens for files that would otherwise migrate, and it reads only `.v` files so the `.vv` fmt fixtures never trip it.

Verified with a two-file module: `User` declared + `json.encode(u)` in main.v, `fn (u User) to_json()` in a sibling hooks.v → stays on `import json`; delete hooks.v → migrates to json2. Added a `json_sibling_hook` vmodules test that isolates this path (the payload struct is declared in the input file, so it's a resolved local struct, not the placeholder case — only the sibling hook keeps it). Both fmt suites pass.

One note on scope: this is a coarse module-level signal (a hook anywhere in the module keeps the file), matching how the in-file check already works. Precise per-payload-type matching would need to parse sibling receivers; happy to tighten it if you'd prefer that.

medvednikov:
Both fixed in 93ead09107.

**Comptime-call args** — right, this is the same shape as the asm-operand case: `comptime_call`'s `else` branch builds `node.args.map(call_arg_spread_str).join(', ')`, i.e. `arg.str()`, never `call_expr`. So a migratable `json.encode(u)` in `a.$method(json.encode(u))` couldn't be rewritten, yet the general scan didn't flag it and the import migrated — stranding the call. I renamed the asm-operand visitor to `json_verbatim_arg_scan_visit` (it flags any json.encode/decode/encode_pretty) and now route both asm operands and comptime-call args through it. Verified `a.$method(json.encode(u))` keeps the file on `import json`, while a comptime call with no json still lets the file migrate its other calls.

**Parenthesized payloads** — added a `ParExpr` arm to `payload_expr_needs_legacy_json` that unwraps and recurses before matching the inner literal. Verified `json.encode((rune(233)))` and `json.encode((struct { t time.Time }{...}))` stay on legacy json, while `json.encode((int(5)))` / `json.encode((User{}))` still migrate.

Updated the `json_migrate_comptime_call_arg` keep test to use a migratable `json.encode` (so it actually exercises this fix rather than the always-unmigratable encode_pretty) and added a `json_migrate_paren_payload` keep test. Both fmt suites pass.

(Aside: while testing I hit a false-keep from running vfmt on files sitting in `/tmp` — the new sibling-hook scan picked up unrelated `.v` files there with `to_json` methods. Not a real-project concern since a module dir only holds its own files, but it's a concrete reminder of the coarseness we discussed on the previous commit.)

medvednikov:
Reworked the migration trigger and addressed both review points in d7f6c55e8f.

**Design change:** `v fmt` now migrates `json`->`json2` **by default**, but skips `_test.v` files and `.vv` fixtures (they exercise the legacy module on purpose — the json module's own suite, the json-vs-json2 benches, etc.). `-no-migrate-json2` opts out; the detection lives in the vfmt tool (`should_migrate_json2`), while the engine (`FmtOptions.migrate_json2`) is unchanged. Because tests are skipped and the few in-scope non-test json files are already kept by the safe-migration guard, `v fmt -verify` stays green with no bulk codebase diff.

**P2 — document the flag:** `vlib/v/help/common/fmt.txt` now describes the default migration behavior and the `-no-migrate-json2` flag, so `v fmt -help` shows it.

**P2 — generated C artifacts:** these were swept in by a too-broad `git add vlib/v/tests/`. I force-pushed to drop that commit entirely; the branch now adds **0** `.c`/`.v3cc` files (`git diff master --name-only | grep -c '\.c$'` -> 0). The 8 test-file migrations from that commit are reverted too, since `_test.v` is now skipped.

Verified: default migrates a regular `.v`, skips `_test.v`/`.vv`, `-no-migrate-json2` disables; json module tests + option/fixed-array diff tests are all stable under `v fmt -verify`; both fmt unit suites pass.

medvednikov:
Both fixed in b01ab184d8.

**Unresolved encode identifiers** — added `encode_ident_needs_legacy_json`. When a `json.encode(x)` payload is a plain identifier, it resolves the declared type via `ident.obj`/`scope.find_var` — function parameters carry a type at parse time even without the checker. Verified:
- `fn f(c http.Cookie) string { return json.encode(c) }` → kept (http.Cookie is a placeholder here with a hidden time.Time).
- `fn f(u User) string { return json.encode(u) }` with a local safe `User` → still migrates (param resolves to a non-sensitive type).
- `u := User{}; json.encode(u)` → kept, since an inferred local's type is unresolved at fmt time and can't be proven safe.

**Attributed sibling hooks** — confirmed `@[manualfree] fn (u User) to_json()` on one line is valid V, and the scan skipped it because the trimmed line started with `@[`. It now strips leading attributes before matching the method name; verified a same-line-attributed sibling hook keeps the file unmigrated.

Added a `json_migrate_encode_ident_param` keep test. For the sibling-hook fix I added a `@[manualfree]` attribute to the existing `json_sibling_hook` fixture — note the committed form must be on its own line (a same-line attribute isn't vfmt-clean, so it can't live in a verify-checked `.v` file); the exact same-line path is covered by a local multi-file check. Both fmt suites pass.

medvednikov:
Fixed in ba8d472581.

You're right — the encode guard only reasoned about literal/cast/container/identifier payloads and treated everything else as safe, so `json.encode(time.now())` / `json.encode(make_event())` migrated and changed the contract.

Inverted the default: replaced `payload_expr_needs_legacy_json` with `encode_payload_is_safe_to_migrate`, which migrates only when the payload is *provably* safe — a typed non-sensitive literal/cast, a safe container whose elements are each provably safe, a resolved-safe identifier (parameter), or a plain scalar literal. Any call, selector/index expression, unresolved local, or imported/placeholder type now keeps the file on legacy json.

Verified:
- `json.encode(time.now())`, `json.encode(make_event())`, `json.encode(b.name)` → kept.
- `json.encode(User{})`, `json.encode(safeParam)`, `json.encode(5)` → still migrate.

Added a `json_migrate_encode_call_payload` keep test; both fmt suites pass, fmt.v is vfmt-clean, and the in-scope non-test json files remain stable.

medvednikov:
Fixed in 8d3af0b623.

Right — the sum-type case only kept when a *variant* was legacy-sensitive, missing the zero-value difference. legacy json omits an unset sum value (`json.encode(Box{})` -> `{}`) while json2 emits a synthetic wrapper for the unknown value (per encode_struct_test.v), and vfmt can't know the runtime value.

Changed the `ast.SumType` branch in `type_needs_legacy_json` to return true unconditionally, so any sum-type payload/field keeps the file on legacy json (a sensitive variant would have kept it anyway). Verified `struct Box { val Val }` with `type Val = int | string` and a direct `json.encode(v Val)` both stay on legacy json, while a plain struct still migrates. Added a `json_migrate_safe_sumtype_field` keep test (all-safe variants); both fmt suites pass.

medvednikov:
@codex review

medvednikov:
Fixed in 2f880f3714.

Confirmed: `json.encode(u // note)` migrated to `json2.encode(u // note\n\t , escape_unicode: true)` — the appended option landed after the comment as a stray leading comma.

Since the synthetic `escape_unicode: true` isn't a real AST argument, it can't cleanly participate in call_args' comment/newline layout. So I leave any `json.encode` whose argument carries a comment unmigrated (`call.args[0].comments.len > 0`), matching the existing decode commented-type-arg handling. Verified `json.encode(u // note)` now stays on `import json` untouched, while a comment-free `json.encode(u)` still migrates. Added a `json_migrate_encode_arg_comment` keep test; both fmt suites pass and fmt.v is vfmt-clean.

medvednikov:
Fixed in 2be13ba977.

Confirmed `json.encode([...times])` with `times []time.Time` migrated — the spread lives on `ArrayInit.update_expr` and there are no explicit `exprs`, so the safety check fell through.

The `ast.ArrayInit` arm of `encode_payload_is_safe_to_migrate` now also checks `update_expr` when `has_update_expr` is set, recursing through the same provably-safe logic. Verified:
- `json.encode([...times])` / `json.encode([...times, time.now()])` (times `[]time.Time`) → kept.
- `json.encode([...nums])` with `nums []int` → still migrates (safe element type).

Added a `json_migrate_array_spread_payload` keep test; both fmt suites pass and fmt.v is vfmt-clean.

medvednikov:
Fixed in 78a4306a88, mirroring the array-spread fix.

Confirmed `json.encode({ ...times })` with `times map[string]time.Time` migrated — the spread lives on `MapInit.update_expr`, there are no explicit `vals`, and the literal's typ/key/value are zero at fmt time.

The `ast.MapInit` arm of `encode_payload_is_safe_to_migrate` now inspects `update_expr` when `has_update_expr` is set, recursing through the same provably-safe logic. Verified `json.encode({ ...times })` (map[string]time.Time) → kept, while `json.encode({ ...nums })` with `map[string]int` → still migrates. Added a `json_migrate_map_spread_payload` keep test; both fmt suites pass and fmt.v is vfmt-clean.

medvednikov:
Fixed in 28c760e07f.

Confirmed the decode divergence: `json.decode(Settings, s)` for `struct Settings { name string = 'anon' @[omitempty] }` migrated, but on `{"name":""}` json2 skips the present empty value and keeps the default `'anon'`, while legacy json assigns `''` — different bytes in.

`struct_field_attr_needs_legacy_json` now also keeps any `@[omitempty]` field with a default value (`field.has_default_expr`), in addition to the composite kinds. Verified:
- decode and encode of a struct with `name string = 'anon' @[omitempty]` → kept.
- a plain `name string @[omitempty]` (no default) → still migrates.
- a defaulted `name string = 'anon'` without `@[omitempty]` → still migrates.

Added a `json_migrate_defaulted_omitempty` keep test; both fmt suites pass and fmt.v is vfmt-clean.

medvednikov:
Fixed in 6186b2bea8.

Confirmed `handlers[json.encode('x')]()` migrated to `import json2` while the callee still read `json.encode('x')` — call_expr prints a non-call callee verbatim via `node.left.str()` (the `mod == '' && name == ''` branch), so the nested call is never rewritten.

The scan now sub-walks such a callee — `node.left` when it is not a CallExpr — with `json_verbatim_arg_scan_visit` (the same helper used for asm operands and comptime-call args), keeping the file on legacy json if it holds any json call. A callee that *is* a call (`f(json.encode(u))()`) goes through `f.expr` and migrates normally, so it's excluded. Verified both. Added a `json_migrate_fn_value_callee` keep test; both fmt suites pass and fmt.v is vfmt-clean.

medvednikov:
All four addressed in 97f73251f9.

**json2 as an identifier** — added `file_blocks_json_migration`, a small scan that keeps the file on legacy json when it declares a parameter/local/function/const/global named `json2` (adding `import json2` would duplicate the symbol and the checker would reject it). Verified `fn f(json2 int, u User) string { return json.encode(u) }` and a local `json2 :=` both stay on `import json`.

**Decode type-arg comments** — the same scan keeps a file whose `json.decode` has a comment on its type argument, since migration formats only the type expression inside `[]` and would drop the comment. Verified `json.decode(/* target */ User, s)` stays unmigrated.

**Pretty encoding** — `json.encode_pretty(x)` now migrates to `json2.encode(x, prettify: true)` (the non-deprecated form) instead of `json2.encode_pretty`, so it won't re-introduce a deprecation warning.

**Benchmark** — restored `bench_json_vs_json2.v` to its legacy `import json` + `json.*` calls and excluded it from the vfmt migration in `vtest-cleancode.v` (it has `fn main()` and is run via `v crun`, so it can't be an `_test.v`). It keeps comparing the two modules.

Added `json_migrate_json2_identifier` and `json_migrate_decode_type_comment` keep tests plus a `json_migrate_encode_pretty` migrate pair; both fmt suites pass and the sources are vfmt-clean.

medvednikov:
Fixed in 2fcfd592f2.

Right — the collision scan only handled `:=`, missing binders stored on the loop node. Extended `file_blocks_json_migration` to cover:
- `for json2 in users` / `for json2, v in m` — ForInStmt key_var/val_var
- `$for json2 in T.fields` — ComptimeFor val_var
- `for json2 := 0; ...` — ForCStmt init (a loop-header field outside children())
- `if json2 := opt() {}` — IfGuardExpr vars (sub-walking each branch condition, since IfBranch.children() omits the condition)
- `|json2| ...` — LambdaExpr params (anon-fn params were already covered via the FnDecl case)

Verified all of the above keep the file on `import json`, while a non-shadowing `for u in users` still migrates. Factored the `:=` check into `assign_decl_shadows` and made the visitor a named function so it can sub-walk. Added a `json_migrate_loop_binder` keep test; both fmt suites pass and fmt.v is vfmt-clean.

medvednikov:
Fixed in ff6c7e6325.

Confirmed `for json.decode(/* target */ Config, raw)!.enabled {}` migrated and dropped the comment — `Node.children()` yields only the body for ForStmt/ForInStmt/ForCStmt, so the header expressions were outside the scan.

`file_blocks_json_migration`'s visitor now sub-walks the loop-header expressions (mirroring the if-condition sub-walk already there):
- ForStmt condition
- ForInStmt iterable + range `high`
- ForCStmt init / cond / inc

Verified a commented decode in each of those headers keeps the file on `import json`, while a plain `for json.decode(Config, raw)!.enabled {}` still migrates (the header call is rewritten normally via `f.expr`). Added a `json_migrate_loop_header_decode` keep test; both fmt suites pass and fmt.v is vfmt-clean.

medvednikov:
@codex review

medvednikov:
Both fixed in dc7c287d7f.

**P1 — function-value references** — confirmed `[1, 2].map(json.encode)` migrated the import to json2 while leaving `json.encode` (unresolved → uncompilable). A legacy json function used as a value is a plain qualified `Ident` (`json.encode`), not a CallExpr, so it never hit the call migration. `file_blocks_json_migration` now keeps a file that references `json.encode`/`json.decode`/`json.encode_pretty` as a bare identifier (or `json.x` selector). Critically verified that a normal `json.encode(u)` / `json.decode(User, s)` call — a CallExpr with a resolved kind, not a bare Ident — still migrates, so this doesn't disable real-call migration.

**P2 — pretty unicode escaping** — `json.encode_pretty(x)` now migrates to `json2.encode(x, prettify: true, escape_unicode: true)`, matching the plain encode branch and legacy json's `\uXXXX` escaping (json2 defaults escape_unicode to false).

Added a `json_migrate_fn_value_ref` keep test, updated the encode_pretty expected; both fmt suites pass and fmt.v is vfmt-clean.

medvednikov:
Handled in 4dcbede594.

Added `file_has_embedded_field_collision`: for each struct declared in the file it uses `table.struct_fields` (which flattens embeds recursively) and flags a duplicate JSON key. Legacy json flattens embedded fields under their original names (duplicate `a`/`c`) while json2 disambiguates as `Foo.a` / `Foo.Baz.c`, so such a payload is kept on legacy json. `json_encoded_field_name` honours `@[json: 'x']` / `@[json: '-']` / `@[skip]`, so a rename that removes the collision still migrates.

Verified:
- `struct Bar { Foo; Baz; a int }` (Foo/Baz sharing `c`, colliding `a`) → kept.
- an embed with no name collision → still migrates.
- a collision resolved via `@[json: 'foo_a']` → still migrates.

Added a `json_migrate_embedded_collision` keep test; both fmt suites pass and fmt.v is vfmt-clean.

PR #27824, opened 2026-07-17
v3: compile Volt and parallelize monomorphization

medvednikov:
## Summary

- make v3 compile and run Volt, including its generic/comptime, JSON, C interop, and code-generation patterns
- keep ownership-only checking and runtime support behind `-d ownership`
- cache generic metadata and resolution views, and parallelize generic specialization over shared AST worker regions
- retain detailed monomorphization profiling behind `-v` while leaving normal output concise

## Why

Volt exposed missing v3 compatibility paths and made generic monomorphization spend minutes in repeated full-AST scans, type-view construction, and serial specialization. The new caches avoid repeated discovery work, nested specializations are discovered in the active pass, and independent specializations are emitted in parallel.

## Impact

Volt now builds and launches with v3, loads its cached chats/messages, and its monomorphization stage is below the one-second target (779.97 ms in the validated full native build).

## Validation

- `./v -gc none -prealloc -o /tmp/v3_pr_rebased vlib/v3/v3.v`
- normal and `-d v3_no_parallel` v3 compiler builds
- `vlib/v3/tests/review_transform_regressions_test.v`
- `vlib/v3/tests/c_variadic_call_codegen_test.v`
- `vlib/v3/gen/c/target_test.v`
- `vlib/v3/types/checker_scoped_transform_test.v`
- full native Volt build and launch; cached chats/messages loaded without a crash
- remaining broad suites were skipped at the requester's direction


medvednikov:
@codex review

PR #27873, opened 2026-07-19
builtin: document the rest of int.v and fix some typos in the docs

tkhuu01:
# Summary

Documents the rest of int.v public functions and fixes a few typos in the docs.

### DIFF
--- a/vlib/builtin/int.v
+++ b/vlib/builtin/int.v
@@ -140,10 +140,15 @@ pub fn (n u16) str() string {
 	return int(n).str_l(6)
 }
 
+// str returns the value of the `i32` as a `string`.
+// Example: assert i32(-32769).str() == '-32769'
 pub fn (n i32) str() string {
 	return int(n).str_l(11)
 }
 
+// hex_full returns the value of the `int` as a *full* 8-digit hexadecimal `string`.
+// Example: assert int(20).hex_full() == '00000014'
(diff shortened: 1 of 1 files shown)


vlib/builtin/int.v
```diff
@@ -441,55 +446,74 @@ pub fn (nn i64) hex() string {
...
+// Note that the output after '0x' is ***not*** zero padded.
 pub fn (nn byteptr) str() string {
 	return '0x' + u64(nn).hex()
 }
 
+// str returns the value of the `charptr` as a '0x'-prefixed hexadecimal `string`.
+// Note that the output after '0x' is ***not*** zero padded.
 pub fn (nn charptr) str() string {
 	return '0x' + u64(nn).hex()
 }
```
tkhuu01, line 470: I just noticed looking back at the PR that the functions are named `str()` instead of `hex()`, is this intentional?

gechandesu: The `str()` method returns a string representation of an object and is implicitly called when you print an object or use it in string interpolation. For all types, `str()` is generated automatically by the compiler, but a custom implementation is used here. A separate `hex()` method seems to have been unnecessary so far, so it remains as is.

medvednikov:
@codex review

medvednikov:
Thanks!

Issue #27675, opened 2026-07-06
HTTP/3 (QUIC) support for net.http


quaesitor-scientiam:
## Summary

Add HTTP/3 support to `net.http`, built on a new QUIC transport implementation
(`vlib/net/quic/`). This tracks a large, multi-phase effort matching the existing
HTTP/1.1 + HTTP/2 quality bar (connection pooling, conformance testing) already
in the repo, targeting Linux, Windows, and macOS.

## Why a new TLS 1.3 handshake instead of using vendored mbedTLS

QUIC (RFC 9001) needs per-encryption-level secret export (Initial/Handshake/
1-RTT) and a way to carry TLS handshake bytes over QUIC CRYPTO frames instead of
a normal TLS record layer/socket.

The vendored mbedTLS (3.6.6) has **no QUIC support at all** — confirmed by
reading the vendored source (only the generic `mbedtls_ssl_set_export_keys_cb`
exists; there is no record-layer bypass and no way to carry the
`quic_transport_parameters` extension). This isn't specific to our vendored
version: it's a long-standing, unresolved upstream limitation
([Mbed-TLS/mbedtls#6178](https://github.com/Mbed-TLS/mbedtls/issues/6178),
[#4731](https://github.com/Mbed-TLS/mbedtls/issues/4731)). For comparison,
`ngtcp2` (the reference QUIC implementation) supports OpenSSL, LibreSSL,
GnuTLS, BoringSSL, AWS-LC, picotls, and wolfSSL as TLS backends — mbedTLS is
absent from that list for exactly this reason.

Patching mbedTLS ourselves to add the missing hooks was estimated at
~1150–2200 LOC inside the library's most actively-maintained, most complex
files (`ssl_tls13_client.c`/`ssl_tls13_server.c`), with real ongoing rebase risk
against every upstream mbedTLS release.

**Decision: implement a QUIC-scoped TLS 1.3 handshake (RFC 8446) from scratch in
pure V**, reusing already-existing, already-tested V crypto (X25519 ECDH,
Ed25519, HKDF, AES-GCM, SHA-256), and delegating only X.509 parsing/chain
validation to mbedTLS's already-bound C functions — the same thing HTTP/1.1 and
HTTP/2 already do today, with no mbedTLS source patch required. This lands at
comparable total LOC (~1500–3000) to patching mbedTLS, but as fully-owned V code
with no upstream-rebase risk.

## Scope for v1 (open to discussion)

[...]

PR #27884, opened 2026-07-21
HTTP/3 (QUIC) foundation: Phase 7 (Loss detection and NewReno congestion control)

quaesitor-scientiam:
## Summary

Phase 7 of HTTP/3/QUIC support (#27675), continuing from the completed
Phase 6 work in #27882 (stream layer and flow control). This branch
starts from #27882's tip, since Phase 7 (loss detection & NewReno
congestion control) needs the packet number spaces from Phase 5 and the
frame/stream layer from Phase 6 to have anything meaningful to time out or
retransmit.

All ancestor PRs (#27680, #27877, #27880, #27881, #27882) have now merged
into `master`; this branch has been rebased onto current `master` (cut at
#27882's own merge commit) and contains only this phase's own work.

See [`vlib/net/quic/PROGRESS.md`](https://github.com/quaesitor-scientiam/v/blob/http3-quic-loss-detection/vlib/net/quic/PROGRESS.md)

## Scope

- `rtt.v` — RFC 9002 §5.3 RTT estimation: first-sample vs. subsequent-sample
  formulas differ (first sample seeds `smoothed_rtt`/`rttvar` directly, no
  EWMA); ACK Delay adjustment clamped to never drive the adjusted RTT below
  `min_rtt`, and skipped entirely for Initial/Handshake spaces.
- `loss_detection.v` — packet-threshold (`kPacketThreshold = 3`) OR
  time-threshold (`9/8 * max(latest_rtt, smoothed_rtt) + kGranularity`) loss
  detection, either condition alone sufficient. A single connection-wide PTO
  timer sourced from the earliest-expiring of the three packet number
  spaces (not three independent per-space timers — packet number spaces are
  independent for numbering purposes, but the PTO timer is not). Exponential
  PTO backoff, reset on ack-eliciting progress.
- `congestion_control.v` — NewReno per RFC 9002 Appendix B: slow start,
  congestion avoidance, ssthresh halving on ordinary loss, a distinct
  persistent-congestion collapse path (harsher reset than ordinary loss),
  and recovery-period tracking so a single loss episode isn't double-counted.

## Test plan

- [x] First-sample vs. subsequent-sample RTT formula, both paths exercised
      (`test_rtt_estimator_first_sample_seeds_directly`,
      `test_rtt_estimator_subsequent_sample_uses_ewma`, both against
      hand-derived exact values).
- [x] ACK Delay clamped so the adjusted RTT sample never drops below
      `min_rtt`, and is skipped for Initial/Handshake spaces
      (`test_rtt_estimator_ack_delay_never_pushes_below_min_rtt`,
      `test_rtt_estimator_ack_delay_ignored_for_initial_and_handshake`,
      `test_rtt_estimator_max_ack_delay_clamp_only_after_confirmation`).
- [x] Packet-threshold-only loss detection (time threshold not yet met) —
      `test_packet_threshold_only_loss_detection`.
- [x] Time-threshold-only loss detection (packet threshold not yet met) —
      `test_time_threshold_only_loss_detection`.
- [x] Single PTO timer correctly sourced from the earliest-expiring space
      across all three independent packet number spaces, including
      application_data excluded pre-confirmation and included post-
      confirmation (`test_pto_time_and_space_single_timer_sourced_from_earliest_expiring_space`,
      `test_pto_time_and_space_excludes_application_data_before_handshake_confirmed`,
      `test_pto_time_and_space_includes_application_data_once_handshake_confirmed`).
- [x] Persistent-congestion collapse as a dedicated test, distinct from
      ordinary-loss ssthresh halving
      (`test_on_packets_lost_persistent_congestion_collapses_to_minimum_window`
      vs. `test_on_packets_lost_ordinary_loss_halves_window_via_ssthresh`).
- [x] Single reaction per recovery episode (no double-counting repeated
      losses within one recovery period) —
      `test_single_reaction_per_recovery_episode`, the plan's own explicitly
      named test.
- [x] All new/modified code passing under `./vnew test <path>` — 31/31
      files in `vlib/net/quic/`.

`/vreview` (full A-G pass) found and fixed two issues before the initial
commit: `NewRenoCongestionControl.is_in_slow_start()` originally shortcut to
`ssthresh == none` instead of RFC 9002's actual `congestion_window <
ssthresh` — these diverge after a persistent-congestion collapse (which
resets `congestion_window` to `kMinimumWindow` while leaving the larger,
just-computed `ssthresh` untouched), so cwnd can legitimately fall back
below an already-set ssthresh and must re-enter slow start rather than stay
in congestion avoidance. And a real DoS: `on_ack_received`'s newly-acked
extraction originally iterated each ACK range's own `[smallest, largest]`
span directly, but `largest_acknowledged`/`first_ack_range` are independent
wire varints (RFC 9000 §19.3) — a tiny, well-formed ACK frame can legally
claim a range spanning up to 2^62-1 packet numbers with no relationship to
the frame's own wire size, and the parser's existing `ack_range_count`
bound (checked against remaining buffer length) does nothing to limit an
individual range's span. Fixed by iterating `sent_packets` (bounded by how
many packets *we* actually have outstanding) and testing range membership
instead of iterating the peer-supplied span; regression test constructs a
`largest_acknowledged = 2^62` range and confirms the call still completes
and resolves correctly.

### Rebase + full RFC 9002 audit (2026-08-11)

Rebased onto current `master` after #27882 merged (`git rebase --onto
upstream/master 92b4fda8d6 HEAD`, same technique used for the prior two
phases in this stack). Two stale CI workflow commits (OpenSSL install path
fixes, superseded by upstream's own CI fixes) were dropped during the
rebase; no code conflicts.

While bringing the branch up to date, did a full requirements-driven audit
of `is_persistent_congestion` and its two callers against the actual RFC
9002 §6.2/§7.6/Appendix A.9 text (fetched directly, not from memory) and
found three gaps, all fixed:

- The persistent-congestion duration formula (§7.6.1) was missing its
  `max_ack_delay` term entirely — persistent congestion could trigger
  earlier than the spec permits for the `application_data` space.
- §7.6.2's third condition — "a prior RTT sample existed when these two
  packets were sent" — was never checked at all.
- `on_loss_detection_timeout`'s timer-triggered loss branch (Appendix
  A.9's own `OnPacketsLost` call) never determined `persistent_congestion`
  for that path at all; only the ACK-triggered path did.

5 new regression tests, each Phase-R'd (confirmed failing with just its own
fix ablated, then confirmed passing restored). Also closed 4 missing
doc-comment gaps found by `missdoc` during this pass. Full
`vlib/net/quic/` suite: 31/31.


JalonSolov:
Local AI Review findings:

1. **Persistent-congestion duration incorrectly omits `max_ack_delay` for Initial/Handshake** (`vlib/net/quic/loss_detection.v`, around lines 234-236 and 420-423). RFC 9002 §7.6.1 explicitly says the duration includes `max_ack_delay` irrespective of packet-number space; the per-space zeroing rule applies to PTO computation in §6.2.1, not persistent-congestion duration. As written, losses in Initial/Handshake can collapse the congestion window after only `3 * PTO` instead of `3 * (PTO + max_ack_delay)`. Please pass the peer `max_ack_delay` through unchanged when evaluating persistent congestion, or keep the space-specific rule out of this calculation.

2. **Persistent congestion is reported from the timer-triggered loss path** (`loss_detection.v`, around lines 414-428). RFC 9002 §7.6.2 says a sender establishes persistent congestion *after the receipt of an acknowledgment*; a time-threshold loss notification from `on_loss_detection_timeout` is not an acknowledgment and must not by itself trigger the persistent-congestion collapse. This new branch can reduce `congestion_window` to `kMinimumWindow` during a timeout without any ACK evidence. The `test_on_loss_detection_timeout_reports_persistent_congestion_for_timer_triggered_loss` test appears to encode the opposite behavior and should be revised.

quaesitor-scientiam:
Thanks — both confirmed and fixed in 6b3e5deec2.

1. **max_ack_delay per-space zeroing**: confirmed. RFC 9002 §7.6.1 states explicitly: "Unlike the PTO computation in Section 6.2, this duration includes the max_ack_delay irrespective of the packet number spaces in which losses are established." I'd copied `pto_time_and_space`'s per-space zeroing onto the persistent-congestion formula by analogy without checking whether §7.6.1 actually carries the same caveat as §6.2.1 — it doesn't. `on_ack_received` now passes `max_ack_delay` through to `is_persistent_congestion` unchanged for every space. New regression test exercises `on_ack_received` directly (not just `is_persistent_congestion`) with the Initial space and a nonzero `max_ack_delay`, since the bug was in the caller, not the formula itself.

2. **Timer-triggered persistent congestion**: confirmed. RFC 9002 §7.6.2 opens with "A sender establishes persistent congestion after the receipt of an acknowledgment" — a time-threshold loss batch from `on_loss_detection_timeout`, with no ACK involved, must never itself trigger the collapse. Reverted: `LossTimeoutResult` no longer carries a `persistent_congestion` field at all; only `on_ack_received`'s `AckProcessingResult` does. The old test you flagged is removed rather than inverted, since there's nothing left to assert once the field doesn't exist.

Full `vlib/net/quic/` suite: 31/31.

quaesitor-scientiam:
CI failures are not related to this PR

PR #27885, opened 2026-07-21
HTTP/3 (QUIC) foundation: Phase 8 (Connection lifecycle)

quaesitor-scientiam:
## Summary

PR tracking Phase 8 of HTTP/3/QUIC support (#27675), continuing from the
completed Phase 7 work in #27884 (loss detection and NewReno congestion
control). Idle-timeout reset timing and the draining-state ping-pong guard
both need to reason about ack-eliciting sends/receives, a concept Phase 7's
loss detection already tracks.

**Update:** #27680, #27877, #27880, #27881, #27882, and #27884 have all
now merged. This branch has been cleanly rebased onto current `master`
(cherry-picked as a single self-contained commit) and contains only
Phase 8's own work — 2 commits total. Ready for review; Phase 9+ continues
on later branches in the stack.

See [`vlib/net/quic/PROGRESS.md`](https://github.com/quaesitor-scientiam/v/blob/http3-quic-connection-lifecycle/vlib/net/quic/PROGRESS.md)

## Scope

- `idle_timeout.v` — effective idle timeout is the min of the two peers'
  non-zero `max_idle_timeout` transport parameters, where 0 means "no
  timeout" (not literally zero) — all 4 combinations (both zero, one zero,
  neither zero) need distinct handling. Reset on an ack-eliciting packet
  RECEIVED or ANY packet SENT — a deliberate asymmetry, not "any packet
  either direction".
- `connection_close.v` — CONNECTION_CLOSE with a transport error code vs.
  an application error code are distinct wire variants (RFC 9000 §19.19);
  closing/draining state to prevent an infinite CONNECTION_CLOSE
  ping-pong (RFC 9000 §10.2's rate-limited-retransmission-on-receipt
  vs. draining's fully silent no-response).
- `stateless_reset.v` — recognized only as a fallback once normal AEAD
  decryption has already failed, never as a first-choice interpretation;
  immediate ungraceful teardown, no draining period. Scoped-down
  NEW_CONNECTION_ID handling: record stateless-reset tokens for matching
  purposes without implementing full connection ID rotation/migration.
- `ecn.v` — no OS-level ECN socket option exists in V today, so ECN
  validation always resolves to failed/unsupported (a spec-legal fallback,
  RFC 9000 §13.4.2) rather than being ignored outright; received ACK
  frames' ECN counts are still parsed without erroring, just never trigger
  an ECN-based congestion response.
- `pmtu.v` — pinned to the 1200-byte RFC 9000 §14.1 safe minimum, no
  active DPLPMTUD probing in v1 (a legitimate, spec-sanctioned
  conservative choice). Connection migration (RFC 9000 §9) stays
  explicitly out of scope.

## Test plan

- [x] Idle-timeout min-of-non-zero-or-infinite across all 4 combinations
      (both zero, local-only zero, peer-only zero, neither zero) —
      `test_effective_idle_timeout_all_four_zero_nonzero_combinations`.
- [x] Idle-timeout reset asymmetry: an ack-eliciting RECEIVE resets it, a
      non-ack-eliciting receive does not, but ANY send does —
      `test_idle_timeout_state_reset_asymmetry`.
- [x] CONNECTION_CLOSE closing-state rate-limited retransmission on
      receipt vs. draining-state fully silent behavior —
      `test_connection_close_tracker_closing_rate_limited_retransmission`,
      `test_connection_close_tracker_draining_is_fully_silent`.
- [x] Stateless reset matched only via a token previously recorded for
      the exact connection ID queried, never fabricated
      (`test_is_stateless_reset_recognized_only_via_a_matching_token`).
      The "only after a decrypt failure" ordering itself is a caller
      contract documented on `is_stateless_reset` — there's no real
      packet-processing pipeline yet to integration-test that ordering
      against (arrives with Phase 9's `QuicConn`).
- [x] ECN counts parsed from a received ACK frame without erroring, and
      `is_validated()` confirmed to always report false regardless of
      what was recorded (including a non-zero ECN-CE count) —
      `test_ecn_state_parses_counts_without_erroring_and_never_validates`.
- [x] Runtime assertion that `fits_within_pmtu` stays pinned at exactly
      1200 bytes with no active probing path —
      `test_fits_within_pmtu_runtime_assertion_at_1200_bytes`.
- [x] All new/modified code passing under `./vnew test <path>` — 36/36
      files in `vlib/net/quic/`.

`/vreview` (full A-G pass) found and fixed one issue before the original
commit: a real integer-overflow bug. `max_idle_timeout` is a peer-supplied
transport-parameter varint that `transport_parameters.v` accepts with NO
upper bound (up to the full 2^62-1 varint range) — `effective_idle_timeout`
originally multiplied it directly by `time.millisecond`, which overflows
the `i64` backing `time.Duration` for any peer value above ~9.2 trillion
ms, silently producing a nonsensical (possibly negative) timeout: a
hostile or buggy peer could self-inflict a near-immediate teardown, or
effectively disable the idle timeout altogether. Fixed by clamping to the
exact mathematically-necessary bound (derived from `time.infinite`, i.e.
`i64::MAX` nanoseconds, not an arbitrary policy number) before scaling;
regression test feeds the maximum possible QUIC varint (2^62-1) through
both the one-sided and both-sided paths and confirms a large-but-finite,
strictly positive `Duration` comes back.

While rebasing onto `master`, a second `/vreview` pass found and fixed two
more issues: `stateless_reset.v` imported a module (`crypto.internal.subtle`)
that doesn't exist upstream — fixed to the real `crypto.subtle`, whose
`constant_time_compare` has the exact API already called and was verified
genuinely constant-time by reading its implementation. Also added 5
missing `pub fn`/`pub struct` doc comments this repo's `report-missing-fn-doc`
CI check would otherwise have flagged.


quaesitor-scientiam:
CI failure analysis shows none are related to this PR

PR #27898, opened 2026-07-22
v3: restore fast cached Volt rebuilds

medvednikov:
## Summary

- restore reliable Volt module-cache reuse after the initial v3 caching merge
- validate cached source, headers, native inputs, ABI-sensitive runtime data, and exact cache dependencies without repeating expensive filesystem work
- report cached-header parsing separately and speed cached parse setup/import resolution
- keep ordinary main-function body edits on the incremental path while falling back safely for new reachability
- materialize new function-only generic specializations incrementally, including nested generic calls
- rebuild the cached prefix when a generic edit introduces a new concrete named-type layout
- keep production monomorphization unchanged
- precompute interface requirement indexes so incremental C generation does not repeatedly rescan interface metadata

## Why

This is the follow-up to #27883. Cached Volt builds had regressed to full compilation, and after restoring cache reuse, small body edits still restarted the full checker/transform/monomorph/cgen pipeline.

The incremental path now reuses cached generic specializations and can append newly required function specializations to the replaceable program body. New generic structs, aliases, sum types, or interfaces still trigger a full rebuild because their C layout declarations belong in the cached prefix. This avoids introducing a second runtime-generics ABI while preserving the existing production monomorphizer.

An isolated warm Volt build using this branch completed in 608 ms. The resulting binary launched successfully and continued through normal initialization.

## Validation

- `../../v -gc none -prealloc -prod -o v3 v3.v`
- focused incremental module-cache tests covering new and nested generic function specializations
- fallback coverage for new generic named-type layouts
- incremental tests covering real main logic, newly reachable functions, C support typedefs, sum equality helpers, compile-signature artifacts, and interface type IDs
- `./vnew vlib/v3/types/interface_impl_test.v`
- `./vnew vlib/v3/gen/c/target_test.v`
- clean-cache and warm-cache builds of `~/code/volt`
- launched the warm-built Volt binary and verified normal startup

### DIFF
--- a/vlib/os/execute_capture_nix.h
+++ b/vlib/os/execute_capture_nix.h
@@ -6,7 +6,7 @@
 // empty (seen on macOS arm64 in CI). Setting CLOEXEC closes the fd
 // automatically across exec, which fixes the race for both fork/execvp and
 // posix_spawn paths.
-static void v_os_execute_set_cloexec(int fd) {
+static inline void v_os_execute_set_cloexec(int fd) {
 	int flags = fcntl(fd, F_GETFD, 0);
 	if (flags >= 0) {
 		fcntl(fd, F_SETFD, flags | FD_CLOEXEC);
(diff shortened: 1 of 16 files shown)


medvednikov:
@codex review

PR #27905, opened 2026-07-22
v3: const/map/static-local codegen fixes, or-block precedence, atomics skip

medvednikov:
## v3: batch of C-backend/parser fixes + failure-list cleanup

Incremental fixes to the v3 compiler. Each is verified independently: a
differential compile check over the passing set (a test that compiled before
must still compile after), a self-host build, byte-identical generated C for
`v3.v` where the change is purely additive, and the v3 unit tests
(parser/gen-c/types/transform).

### Compiler fixes
- **Fixed-array-alias const initializers** — `type UUID = [16]u8` const emitted
  as `const Array_fixed_u8_16 x = (u8[16]){...}`; C cannot initialize an array
  object from a compound-literal array rvalue, so the cast is stripped to a bare
  brace list.
- **Address-taken consts** — a scalar const emitted as `#define main__foo (u32(1))`
  could not have its address taken (`&foo`, e.g. `const keys = [&foo, &bar]!`);
  such consts now get addressable `static const` storage.
- **or-block precedence** — `!f() or { false }` parsed as `(!f()) or { false }`;
  now `!(f() or { false })`, matching V (the or-block binds to the operand).
- **`map == map`** — C cannot compare `struct map` with `==`; map operands now
  route through the existing `v3_map_map_eq` runtime helper.
- **Non-constant `static` locals** — `mut static x := []int{}` emitted
  `static Array x = array_new(...)` (rejected: initializer not compile-time
  constant); now zeroed static storage plus a one-shot guarded init on first
  entry, preserving cross-call persistence.

### Test-list maintenance
- **vlib/x/atomics → skipped** — implemented only via amd64/i386 inline assembly,
  no arm64 backend, so its tests cannot compile on arm64 with any V compiler
  (V 0.5.2 fails identically); they were miscategorised as v3 failures.
- Moved now-passing tests from `v3_test_failures.txt` to `v3_test_passes.txt`
  and updated the `v3_test_failure_groups/` counts.

`vlib/v/tests/consts` is now fully green; several interface / static-var tests
compile and run. Combined with the method-name-mangling fix already on master,
tracked v3 failures drop from 248 to 193.

Not a complete sweep — the remaining failures are a long tail of distinct
codegen / type-checker / comptime / generics bugs (each greening ~1 test) plus
the stored/returned closure-capture feature (which blocks ~30 tests, incl. all
of net/http h2). Those are left for follow-ups.

### DIFF
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -11299,13 +11299,18 @@ fn (mut g FlatGen) gen_string_infix_fallback(node flat.Node, lhs_id flat.NodeId,
 
 // gen_map_infix_eq lowers `map == map` / `map != map` to the runtime map
 // equality helper. C cannot compare `struct map` values with `==`, so without
-// this the generated comparison fails to compile.
+// this the generated comparison fails to compile. Pointer operands are left
+// alone (like the array-equality path): `&m1 == &m2` must keep comparing the
+// pointer addresses, not the pointed-to map contents.
 fn (mut g FlatGen) gen_map_infix_eq(node flat.Node, lhs_id flat.NodeId, rhs_id flat.NodeId, lhs_type types.Type, rhs_type types.Type) bool {
 	if node.op !in [.eq, .ne] {
 		return false
 	}
-	lhs_is_map := map_str_clean_type(types.unwrap_pointer(lhs_type)) is types.Map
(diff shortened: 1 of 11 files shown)


medvednikov:
@codex review

medvednikov:
@codex review

PR #27911, opened 2026-07-22
v3: compile and link all veb examples

medvednikov:
## What

Makes every example under `examples/veb/` compile **and link** with the v3 backend (previously 0/11 linked → now 11/11), along with the net/web fixes those examples depend on.

## Compiler changes

- **Template engine** — port the `$veb.html()` / `$tmpl()` comptime template engine (`parser/tmpl.v`, a `veb_template` flat node, statement-level expansion in the parser). Generates plain string-concat builders, verified byte-identical to v1 on standalone `$tmpl` programs.
- **markused** — follow generic `.index` callees so the `veb.run → run_at → … → AtomicVal` chain stays reachable (this was the universal atomic link gate blocking *all* veb examples), and collect omitted `@[params]` struct field-default calls (`zstd.default_c_level`).
- **`main.Context` vs `veb.Context` collision** — a user `struct Context { veb.Context }` collides by short name with the imported `veb.Context`. Resolved across struct-init codegen, embedded-field access, and generic closure params. The final blocker was a `file_modules` corruption in `checker.enter_module`: a foreign `module_decl` visited while `cur_file` still pointed at the main file rebound that file's module (any anonymous fn in the main file triggered it, e.g. websocket's `on_message` closures). `enter_module` no longer overwrites a file's already-known module.
- **Implicit veb `ctx`** registered as a mutable parameter.
- **Pointer arithmetic** — infer result types for `byteptr`/`charptr`/`voidptr`.
- **C extern prototypes** — skip those that clash with system headers.

## Test changes

- Pass `-gc none` when a test builds `v3.v` as a subprocess. `v3.v` now `$compile_error`s when built with a GC, which was breaking those sub-builds on GC-default toolchains (~20 tests affected).

## Verification

- All 11 `examples/veb/` compile and link to valid binaries.
- Full self-host builds a valid binary; `types/` + `parser/` 8/8, `transform/` + `markused/` + `gen/` 19/19, top-level examples 67/4 (unchanged pre-existing failures).

## Known pre-existing failures (unrelated to this work)

Three of the `-gc none`-fixed tests now build `v3.v` successfully but fail at separate **pre-existing** bugs (confirmed reproducible with a build that omits the `enter_module` fix):
- `selfhost_backend_prune_test` — a `-selfhost` backend-pruning codegen bug (`c__unknown` return type).
- `shared_flag_test` — module-object cache-count assertions from the caching feature.
- `type_checker_errors_test` — error-message-text assertions + a generic link error.

These are noted for a follow-up; the `-gc none` build-flag fix itself is correct.


medvednikov:
@codex review

medvednikov:
@codex review

PR #27919, opened 2026-07-23
parser: report a clear error for the `if if cond {` typo (fix #27907)

### LINKED ISSUE #27907: Foolish parser, Dumb error messages! (sorry for harsh words)
## V is the best language, but it has the dumbest parser!

Image

## Dont get me wrong, I LOVE V, (even more then c), BUT BUT BUT, It has a dumbass parser(or maybe lexer?).
1. use Bison and Flex (If you dont want to write a parser yourself).
2. Pratt parsing (please!)

### Reproduction Steps

Image

```sh
_*_ Running: `v run /home/dzebra/Work/probe/Programming/hsys25/probe_opt/polymorphile/piper/piper.v` _*_


piper.v:69:3: error: unexpected token `}`, expecting `{`
   67 |       }
   68 |     }
   69 |   }
      |   ^
   70 |   return commands
   71 | }

Process completed with exit code: 1
```

### Expected Behavior

Just tell me that I added anathor `if` keyword.

### Current Behavior

Destroys the feel of debug-ing process!

### Possible Solution

Pratt parsing?

I LOVE V-LANG!

### V version

V 0.5.1 1bd23ec

Debian 13
> [!NOTE]
>

medvednikov:
## What

Fixes #27907.

Accidentally writing `if` twice (e.g. `} else if if x == 2 {`) produced a confusing, misplaced error, because the second `if` was parsed as an *if-expression used as the condition*, and the failure only surfaced later when the parser could not find the body `{`:

```
piper.v:69:3: error: unexpected token `}`, expecting `{`
```

Now the parser detects this typo directly and points at the offending `if`:

```
x.v:6:12: error: the condition of an `if` should be a boolean expression, not another `if` statement; did you write `if` twice by mistake?
    4 |     if x == 1 {
    5 |         commands << 'a'
    6 |     } else if if x == 2 {
      |               ~~
    7 |         commands << 'b'
    8 |     }
```

## How

In `if_expr`, after parsing the condition, if the next token is not the body `{` and the condition itself parsed into an `ast.IfExpr` with no `else` branch, we emit the targeted error pointing at the inner `if`.

This is precise and does not regress legitimate code: a real if-expression used as a condition (e.g. `if if x == 1 { true } else { false } { ... }`) has an `else` branch and is followed by the body `{`, so neither guard fires.

## Test

Added `vlib/v/parser/tests/if_double_if_cond_err.{vv,out}`. The full `vlib/v/compiler_errors_test.v` suite passes, and the compiler still self-compiles.

### DIFF
--- a/vlib/v/parser/if_match.v
+++ b/vlib/v/parser/if_match.v
@@ -184,6 +184,17 @@ fn (mut p Parser) if_expr(is_comptime bool, is_expr bool) ast.IfExpr {
 			p.comptime_if_cond = false
 		}
 		comments << p.eat_comments()
+		// catch the common typo `if if cond {` / `} else if if cond {`, where a
+		// second `if` was written by mistake. The inner `if` gets parsed as the
+		// condition expression, so bail out with a helpful message instead of the
+		// confusing `expecting {` error that appears further down the file.
+		if !is_comptime && p.tok.kind != .lcbr {
+			if cond is ast.IfExpr && !cond.has_else {
+				p.error_with_pos('the condition of an `if` should be a boolean expression, not another `if` statement; did you write `if` twice by mistake?',
+					cond.pos)
+				return ast.IfExpr{}
+			}
+		}
 		end_pos := p.prev_tok.pos()
 		body_pos := p.tok.pos()
 		p.inside_if = false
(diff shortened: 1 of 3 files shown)


GGRei:
@codex review

PR #27925, opened 2026-07-23
v3: fix systemic parser/checker/transform/markused/gen issues; +75 passing tests

medvednikov:
## Summary

Rechecked all 156 previously-failing entries end-to-end and moved 75 now-passing tests into [`v3_test_passes.txt`](../blob/agent/v3-fix-75-tests/vlib/v3/v3_test_passes.txt).

- Pass list increased from **2,941 to 3,016**.
- [`v3_test_failures.txt`](../blob/agent/v3-fix-75-tests/vlib/v3/v3_test_failures.txt) now contains **81 reproducible failures**.
- Both lists are sorted, unique, and disjoint.

## What changed

Fixed systemic issues across V3's parser, checker, transform, mark-used, and C generator, including:

- generics / generic-interfaces
- optional / ABI handling
- closures
- variadic short structs
- embedded methods
- iterator dispatch
- preallocated self-host metadata lifetime

## Validation

- All directly modified unit tests pass.
- Key option, HTTP, array, generic-interface, iterator, and embedded-function regressions pass.
- The self-host command succeeds:
  ```
  ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -building-v -o v4 v3.v
  ```
  and `v4` successfully compiled and ran `examples/hello_world.v`.
- Stage-1 performance essentially unchanged: **61.07s / 2.39 GB** max RSS vs. ~61.2s / 2.36 GB baseline.
- V3 self-host stage: **3.15s / 1.49 GB** max RSS.
- All touched V files are formatted; `git diff --check` passes.

## Notes

The retained 81 failures are genuine unsupported V3 cases plus platform/SDK-dependent Windows, Linux, BSD, MySQL, MSSQL, and Redis tests. No unrelated tracked changes were modified.


vlib/v3/gen/c/cleanc.v
```diff
@@ -5854,7 +5855,11 @@ fn c_include_arg_for_target(raw string, vroot string, source_file string, target
 
 fn c_flag_args(raw string, vroot string, source_file string, target pref.Target) []string {
 	target_arg := c_directive_arg_for_target(raw.trim_space(), target) or { return []string{} }
-	clean := c_expand_existing_path_macros(target_arg, vroot, source_file) or { return []string{} }
+	defaults_expanded := c_expand_default_define_macros(target_arg) or { return []string{} }
```
medvednikov: [P1] Preserve `-d` overrides when expanding `$d` in C flags

This always selects the fallback arm, so an explicit `-d N=value` is silently ignored for `#flag ...$d('N', 'default')`. `$d` semantics are to prefer `Preferences.compile_values` and use the default only when the key is absent (see `vlib/v/util/d_value.v`). Please resolve against the compile-values map before argument splitting and add an override test; the new tests cover only fallback expansion.

vlib/v3/transform/fn.v
```diff
@@ -8719,7 +8746,39 @@ fn (mut t Transformer) lift_fn_literal(_id flat.NodeId, node flat.Node) flat.Nod
...
+	context_init := t.a.add_node(flat.Node{
+		kind:           .struct_init
+		value:          context_type
+		typ:            context_type
+		children_start: context_start
+		children_count: flat.child_count(context_fields.len)
+	})
+	context_ptr := t.make_prefix(.amp, context_init)
+	t.set_node_typ(int(context_ptr), '&${context_type}')
+	create := t.make_call_typed('closure.closure_create_with_data', [
```
medvednikov, line 9264: [P1] Reclaim non-escaping runtime closures

Every capturing literal now calls `closure_create_with_data`, which claims an executable closure slot and owns the heap context, but V3 does not emit `closure_try_destroy` or establish a `closure.Lifetime` frame for ordinary non-escaping locals. A capturing closure created in a loop therefore retains one context and executable slot per iteration, eventually causing `closure_alloc()` to keep mapping executable pages. Please integrate escape-aware scope cleanup on all exits, or use lifetime frames, and add a hot-loop regression that verifies slot reuse.

vlib/v3/gen/c/cleanc.v
```diff
@@ -11610,6 +11753,23 @@ fn (mut g FlatGen) gen_string_infix_fallback(node flat.Node, lhs_id flat.NodeId,
...
+		|| g.tc.c_type(lhs_type) != '__v_thread' || g.tc.c_type(rhs_type) != '__v_thread' {
+		return false
+	}
+	lhs_name := g.tmp_name()
+	rhs_name := g.tmp_name()
+	g.write('({ __v_thread ${lhs_name} = ')
+	g.gen_expr(lhs_id)
+	g.write('; __v_thread ${rhs_name} = ')
+	g.gen_expr(rhs_id)
+	g.write('; memcmp(&${lhs_name}, &${rhs_name}, sizeof(__v_thread)) ')
```
medvednikov: [P2] Compare POSIX thread IDs with `pthread_equal`

`pthread_t` is opaque, so bytewise `memcmp` is not a portable equality operation. On POSIX, unwrap `__v_thread.handle` and call `pthread_equal` (inverting the result for `!=`), with a separate Windows implementation. The copied-handle regression can pass on common libc representations while still leaving this incorrect on valid POSIX targets.

medvednikov (commented):
Code review only; CI was intentionally excluded. GitHub does not permit the PR author to submit `REQUEST_CHANGES` on their own pull request, so I am leaving these three blocking correctness/resource-management findings as a comment review. These findings are based on static inspection of the diff and the surrounding compiler/runtime code.

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

PR #27933, opened 2026-07-25
builder: upload a self-contained reproducer for C error bug reports

medvednikov:
### Background

The C error bug reporter already maps the failing C line back to the exact V line (via the generated `#line` directives), but the uploaded `v_source` was only a fixed **±40-line window** around that line. For anything but a tiny single-file program that window is a mid-function fragment that does not compile standalone — so in practice the large majority of reports could not be reproduced without hand-reconstructing the program. (In a sweep of a recent batch, ~90% of `v_source`-bearing reports were unusable truncated fragments.)

Knowing the failing *line* is not enough; a reproducer needs the failing line's **dependency closure** — the enclosing declaration plus the types/consts/globals/functions/imports it transitively uses.

### What this does

At report time the builder still has the parsed AST (`v.parsed_files`) available (the `-g`/vlines path already uses it), so this builds a self-contained reproducer from it:

- Index every **user** top-level declaration (fn/struct/enum/interface/type/const/global) by the names it defines, with its exact source span; collect each file's imports and the failing file's C-interop `#` directives.
- Find the declaration the failing line maps into, and compute the transitive closure of the declarations it references. The reference graph is approximated by matching identifier tokens against the set of top-level names (bounded by a declaration count + byte budget).
- Seed the closure from **both** the failing declaration **and** `main`, so the reproducer stays runnable and the failing code stays reachable (not dropped by `skip_unused`). The failing declaration is seeded first, so its own dependencies win the byte budget.
- Emit a single `module main` file: referenced imports, the failing file's `#`-directives, then the inlined declarations. Only user code is inlined; vlib/installed modules are referenced via `import`. Paths are compared with `real_path` (so a `/tmp` vs `/private/tmp` symlink mismatch does not hide the failing declaration).
- Anything unexpected (no mapping, no user declaration found, nothing to inline) falls back to the previous source window.

### Verified end to end

Compiling a program whose failing `[]u8(ptr)` cast lives in a **helper** (with an unrelated function and an import only that function uses) produces:

```v
module main

const three = 3

fn to_slice(p voidptr) []u8 {
	return unsafe { ([]u8(&u8(p)))[0..three] }
}

fn main() {
	s := 'hello'
	a := to_slice(voidptr(s.str))
	println(a.len)
}
```

i.e. `main` + the helper + the `const` it uses — with the unrelated function **and** its now-unused `import os` correctly dropped — and this file compiles standalone and reproduces the original C error.

### Tests

`c_error_reproducer_test.v` covers the pure logic (identifier extraction, decl-source slicing, closure following references / bounded / seeded-from-main, assembly with import filtering and `#` prelude). The AST-extraction layer is defensive (bounds-checked, `or {}` on I/O) and never panics; on any failure it returns `''` and the caller keeps the old behaviour.

### DIFF
--- a/vlib/v/builder/c_error_report.v
+++ b/vlib/v/builder/c_error_report.v
@@ -109,7 +109,16 @@ fn (mut v Builder) new_c_error_bug_report(ccompiler string, c_output string) CEr
 	// `v_context` shows the lines of whatever file the C error maps to (which can be an
 	// included header, not V source).
 	mapped_source := if v_file != '' { os.read_file(v_file) or { '' } } else { '' }
-	v_chunk := selected_v_source(v_file, mapped_source.split_into_lines(), v_line)
+	// Prefer a self-contained reproducer (the failing declaration plus the closure of the user
+	// declarations it references). It already keeps itself within the byte budget, returning ''
+	// when it cannot, so it is uploaded verbatim; otherwise fall back to a plain source window.
+	repro := v.v_source_reproducer(v_file, v_line, c_error_bug_report_max_v_source_bytes)
+	v_source := if repro != '' {
+		repro
+	} else {
+		v_chunk := selected_v_source(v_file, mapped_source.split_into_lines(), v_line)
+		bounded_v_source(v_chunk.text, c_error_bug_report_max_v_source_bytes, v_chunk.focus)
+	}
 	return CErrorBugReport{
 		kind:           'v-c-compiler-error'
 		v_version:      version.full_v_version(true)
@@ -126,8 +135,7 @@ fn (mut v Builder) new_c_error_bug_report(ccompiler string, c_output string) CEr
 		v_line:         v_line
 		v_context:      numbered_context_lines(mapped_source.split_into_lines(), v_line,
 			c_error_context_radius)
-		v_source:       bounded_v_source(v_chunk.text, c_error_bug_report_max_v_source_bytes,
-			v_chunk.focus)
+		v_source:       v_source
 	}
 }
 
--- a/vlib/v/builder/c_error_reproducer.v
+++ b/vlib/v/builder/c_error_reproducer.v
@@ -84,6 +84,7 @@ fn (v &Builder) v_source_reproducer(v_file string, v_line int, max_bytes int) st
 	mut main_id := -1
 	mut extra_seeds := []int{} // skip_unused roots (init/cleanup/@[markused]/@[export])
 	mut file_headers := map[int]string{} // per-file `module` header (attributes are per-file)
+	mut start_to_decl := map[string]int{} // '<file_id>:<start line>' -> decl id (solver clone dedup)
 	mut file_id := -1
 	for pf in v.parsed_files {
 		file := pf
@@ -164,14 +165,44 @@ fn (v &Builder) v_source_reproducer(v_file string, v_line int, max_bytes int) st
 			if names.len == 0 {
 				continue
 			}
-			id := decls.len
-			decls << ReproDecl{
-				names:   names
-				source:  source
-				file_id: file_id
+			// `-new-generic-solver` (generics.solve_files) removes a generic `fn foo[T]`
+			// declaration and appends concrete clones named `foo_T_int` that keep the original
+			// position. The copied source and its call sites still say `foo`, so also index the
+			// name the source text itself declares.
+			if stmt is ast.FnDecl {
+				src_name := repro_source_fn_name(source)
+				if src_name != '' && src_name !in names {
+					names << src_name
+				}
 			}
-			for n in names {
-				name_to_decl[n] << id
+			// ... and collapse clones sharing one source range into a single declaration:
+			// emitting the same copied source once per concrete instantiation would not compile
+			start_key := '${file_id}:${start}'
+			mut id := -1
+			if prev := start_to_decl[start_key] {
+				id = prev
+				mut merged_names := decls[id].names.clone()
+				for n in names {
+					if n !in merged_names {
+						merged_names << n
+						name_to_decl[n] << id
+					}
+				}
+				decls[id] = ReproDecl{
+					...decls[id]
+					names: merged_names
+				}
+			} else {
+				id = decls.len
+				start_to_decl[start_key] = id
+				decls << ReproDecl{
+					names:   names
+					source:  source
+					file_id: file_id
+				}
+				for n in names {
+					name_to_decl[n] << id
+				}
 			}
 			// index instance methods under a synthetic `<ReceiverType>#methods` key too, so a
 			// declaration that reflects over `<Type>.methods` (`$for m in Foo.methods`) can pull in
@@ -451,6 +482,15 @@ fn repro_closure(decls []ReproDecl, name_to_decl map[string][]int, seeds []int) 
 		// overloaded operators are indexed under their punctuation method name (`+`, `[]`, ...),
 		// which is not an identifier token, so add the operator methods a declaration may use
 		refs << repro_operator_refs(decls[id].source)
+		// string interpolation and the print/dump/assert family invoke custom `str` methods
+		// implicitly: the token scan sees `Foo` in `'\${Foo{}}'` but never the method name, and
+		// replaying with the default `skip_unused` would then drop `Foo.str` (and anything only
+		// it reaches), making the C error vanish. Retain every custom `str` method instead —
+		// a safe over-approximation, mirroring the mark-used walker, which roots string methods
+		// for interpolation.
+		if repro_source_stringifies(decls[id].source, refs) {
+			refs << 'str'
+		}
 		// `$for m in Foo.methods` reflection needs Foo's methods, which are indexed under the
 		// synthetic `Foo#methods` key; add those refs so the whole method set is retained
 		for t in repro_reflection_method_targets(decls[id].source) {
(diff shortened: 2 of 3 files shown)


medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
Addressed both open threads in b034e74ded:

- **Solved generics** (the two `:158` threads): declarations are now also indexed under the name parsed from their own copied source text, so a caller containing `foo(1)` finds `fn foo[T](...)` even after `generics.solve_files` has renamed the AST clones to `foo_T_int`. Clones sharing one source range are collapsed into a single declaration — emitting the same copied source once per concrete instantiation would not have compiled either. This keys off the source text rather than the `-new-generic-solver` pref, so any pass that rewrites declaration names in place is covered.
- **Implicit string methods** (`:450`): a declaration that interpolates (`'${Foo{}}'`) or calls the print/dump/assert family now also references every custom `str` method, mirroring how the mark-used walker roots string methods for interpolation — so `Foo.str` and helpers only it reaches survive the replay's default `skip_unused`. This over-approximates safely: unneeded `str` methods only cost bytes, and an over-budget closure already falls back to the source window.

Unit tests added for the source-level name parser, the stringification detector, and a closure case where a helper is reachable only through an interpolation-invoked `Foo.str`.

medvednikov:
Both addressed in decf6883e7:

- **free under -autofree**: every custom `free` method is now seeded into the closure when the build uses `-autofree`, mirroring how the mark-used walker roots them — so a helper reachable only from `Foo.free()` survives the replay (whose recorded build options include `-autofree`). Same safe over-approximation shape as the `str` retention.
- **import/local collisions**: an import-bound name used in a retained declaration from another contributing file that does not itself import it now triggers the source-window fallback. Token-level scanning cannot distinguish a parameter/local binding from a reference, so any such cross-file use is treated as a collision; uses in files carrying an equivalent import remain ordinary references and do not fall back. Unit test added for the collision predicate.

medvednikov:
Both addressed in ce593d46f5:

- **$env**: now counted as a machine-local resource alongside `$embed_file`/`$tmpl`/`$res` — a retained declaration using it triggers the source-window fallback, since the report records no environment values and the receiver would bake different ones into the C.
- **before_request**: the root predicate now seeds any function named `before_request`, mirroring the mark-used pass's `ends_with` root, so helpers reachable only from a veb hook survive the replayed `skip_unused`. Unit tests added for both predicates.

medvednikov:
All three addressed in fc16247e04:

- **spaced compile-time calls**: detection is token-wise now (`$embed_file ('x')`, `$env\t('HOME')` fall back; `$reserve(` no longer matches `$res`).
- **machine-dependent pseudos**: `@FILE`/`@DIR`/`@LINE`/`@FILE_LINE`/`@LOCATION`, the root/exe pseudos, the hash pseudos and the build-time pseudos now trigger the source-window fallback (flattening changes path/line values; replaying changes root/hash/build values). `@FN`/`@METHOD`/`@MOD`/`@STRUCT`/`@COLUMN` survive the verbatim flatten and `@OS`/`@PLATFORM`/`@BACKEND`/`@CCOMPILER` are reproducible from the recorded build options, so those remain allowed. Matched with identifier boundaries so `bob@FILEserver` in a string stays inert.
- **brackets in attribute strings**: the per-line balance now skips string literals and trailing `//` comments. Since an attribute *string spanning source lines* defeats any per-line balance, the walk additionally verifies the nearest `@[` line with a forward scan of the joined text that requires one uninterrupted group — an unrelated attribute above ordinary code fails it because its own `]` returns the depth to zero early. Unit tests cover the review example, the multi-line-string-with-bracket combination, and the const-array non-swallowing guards.

Also carries the `before_request` suffix-match refinement (mirroring markused's `ends_with` exactly).

medvednikov:
Both addressed in 952663ba1c:

- **veb actions**: `repro_is_markused_root` now receives the table's cached `veb.Result` type index and seeds every function returning it, mirroring `mark_markused_fns` — a helper reached only from a route survives the replayed `skip_unused`. When the program uses no veb the cache index is unset and roots nothing.
- **exported globals/consts**: the non-function root check accepts `@[export]` alongside `@[markused]`, matching `mark_markused_consts`/`mark_markused_globals` (`is_markused || is_exported`). Tests extended for both (including the no-veb negative case).

medvednikov:
Both addressed in 833db843c9:

- **public consts in `-shared` builds**: the non-function root branch now also seeds `pub const` declarations when `v.pref.is_shared` is set, mirroring markused's `is_shared && con.is_pub` retention — a private helper called only from such an initializer survives the replayed `skip_unused`.
- **multiline attributes before generic names**: `repro_source_fn_name` no longer walks lines; it scans the copied source as one string and skips attribute groups with string-aware bracket depth, so a multiline attribute value is consumed whole and the `fn` behind it is found. Tests cover the review example, a bracket inside the attribute string, and stacked single-line attributes.

medvednikov:
Addressed in 4d6b7d1173: a declaration containing a `for ... in ...` loop now also references every custom `next` method, mirroring the mark-used walker's struct-iteration retention — so `Iterator.next` (and helpers only it reaches) survive the replay. The detection is keyword-boundary matched inside the `for` header (`$for` metadata loops, C-style/condition loops, and identifiers like `informal`/`xfor` do not count), with the same safe over-approximation shape as the `str`/`free` retentions. Unit tests cover the detector and an iterator-only-reachable helper closure case.

medvednikov:
Both addressed in 6d78cbf2f4:

- **JSON hooks**: a declaration referencing `encode` now also retains every custom `to_json` and `json_str` method, mirroring the walker's JsonEncoder/Encodable handling — so a helper reachable only through a hook survives the replay instead of json2 silently selecting ordinary field encoding. Same over-approximation shape as the `str`/`free`/`next` retentions.
- **translated `@[c]` roots**: `repro_is_markused_root` now receives the `translated` pref and seeds every function carrying a `c` attribute, matching `markused.mark_used`. Tests cover the hook closure case (helper reachable only through `to_json`) and both translated/non-translated attribute cases.

medvednikov:
All three addressed in 6acbb5ae2d:

- **`$veb.html()`**: counted as a local compile-time resource (both the no-argument and pathed forms, with the same tokenized-spacing tolerance) — a retained action using it falls back to the source window.
- **embedded hash directives in comptime blocks**: retained `$if`/`$match` sources are scanned for `#` lines; a project-local one triggers the fallback, while system headers/libraries embedded in a block stay uploadable verbatim (ordinary top-level hash statements are separate declarations and never appear inside another declaration's source, so the scan is precise).
- **hash-only module files**: the contributing-file restriction on hash directives is gone — the original build applies `#flag`/`#include` from every parsed module file, so all module hashes are now kept (deduplicated) and a local one anywhere in the module forces the fallback. Tests cover the veb template forms, local/system embedded directives, and the existing hash-locality cases still pass.

medvednikov:
All four addressed in 1dacecf284:

- **lifecycle methods**: the `!f.is_method` restriction is gone — mark-used roots any key ending in `.init`/`.cleanup`, so `Foo.init()`/`Foo.cleanup()` are seeded (test flipped accordingly).
- **rooted operator methods**: `+ - * % / < ==` methods are additionally indexed under a synthetic `<Recv>#op` key, and the closure references every retained name's `#op` set — mirroring the walker finalizer's receiver-type-used rooting, so `Vec.+` survives even when no `+` appears in retained source.
- **hash-only conditional blocks**: an active `$if`/`$match` block containing only directives is kept verbatim as an always-included declaration (synthetic `#`-prefixed name, unreferencable by tokens), preserving its module-wide `#flag`/`#include`; the embedded local-path scan still forces the fallback for project-local paths inside it. This also handles unsolved runtime-decided blocks correctly, since the block uploads verbatim with its condition.
- **`$pkgconfig`**: the comptime call form now triggers the source-window fallback like `$env` (package availability is build-machine state); the `#pkgconfig` directive form remains allowed since it resolves on the receiver like a system library. Tests cover each case.

medvednikov:
Both addressed in 0a7986f923:

- **duplicate module imports**: retained imports now merge per module — exact duplicates collapse, and `import x as _` is subsumed by any real import of the module (importing at all runs its initializers). Genuinely different forms of one module (`import log` + `import log as l`) cannot be expressed in one flattened file and trigger the source-window fallback. The merge is a standalone helper with unit tests for all four cases.
- **`-d used_fns` roots**: functions selected by the define are now seeded, matching `markused.mark_used` exactly — literal entries compare against the qualified fn key, `*` patterns glob-match it — so a helper whose only caller is the selected entry survives the replay with the recorded define.

medvednikov:
All three addressed in 04552fc2e8:

- **nested solved blocks**: the opener search now recurses into nested positionless blocks and reports the nesting depth, then walks up past that many inner `$if`/`$match` openers to the one belonging to the outer block — the first naive fix stopped at the inner opener and produced a truncated source range, which the unit test caught.
- **trivia-separated attributes**: the upward walk skips blank and `//` comment lines before looking for the group's closing `]`, matching the parser's trivia handling; the copied declaration keeps the intervening trivia verbatim. The const-array non-swallowing guard still holds (covered by a test).
- **include delimiter**: only the first non-whitespace character after `#include`/`#insert`/`#preinclude` decides system-vs-local, so `#include "private.h" // see <API>` correctly falls back while `#include <stdio.h> // has "quotes"` stays allowed.

medvednikov:
Both addressed in c5669c2306:

- **decode hooks**: a declaration referencing `decode`/`raw_decode` now retains `from_json_string`/`from_json_number`/`from_json_boolean`/`from_json_null`, completing the encode-side hook retention — a helper reachable only through a decoding hook survives, and the replay keeps recognizing the decoder interface.
- **bare relative flag files**: `repro_flag_has_bare_file` classifies a non-option `#flag` field containing a `.` as a filename (platform prefixes carry no dot; option fields start with `-`), and `-include` is treated as local outright — so `#flag helper.o` and `#flag -include config.h` fall back instead of uploading a directive the receiver cannot satisfy, while `-lssl`/`-DVERSION=1.2`/`windows -lgdi32` stay allowed. Tests cover the hook closure case and the flag classification matrix.

PR #27935, opened 2026-07-25
fix(tccbin_tests): surface compile errors instead of swallowing them

quaesitor-scientiam:
`run.sh`/`run.ps1` redirected the compile step's stderr to `/dev/null`
(or `Out-Null`) and just reported `compile error` on failure, with no
way to see why. This made diagnosing a genuine failure on a new
platform (e.g. a real tcc/libgc linking gap, discovered while wiring
up freebsd-amd64's CI) require re-running the compile manually outside
the harness every time.

Both now capture and include the actual compiler output in the FAIL
line, matching how runtime failures already report exit code + output.

Verified: a deliberately incomplete GC flag set now shows the real
`undefined symbol`/`unresolved reference` text instead of a bare
`compile error`, on both a real Linux `tcc.exe` (WSL) and a real
Windows `tcc.exe`, with no regression in the full correctly-flagged
4/4 pass.

### DIFF
--- a/thirdparty/tccbin_tests/run.ps1
+++ b/thirdparty/tccbin_tests/run.ps1
@@ -52,9 +52,9 @@ foreach ($dir in (Get-TestDirs)) {
         $expectSubstr = if ($lines.Count -gt 1) { $lines[1] } else { "" }
 
         $exe = Join-Path $work "$name.exe"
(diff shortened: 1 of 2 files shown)


GGRei:
@codex review

PR #27943, opened 2026-07-26
builtin: prevent prealloc recycle cache bootstrap race

medvednikov:
Older bootstrap compilers only know that `g_memory_block` must be emitted as thread-local storage. When the recycle cache was added as a separate `g_prealloc_block_cache` global, those compilers emitted it as process-global state. Parallel v3 parser/checker workers could then recycle the same scoped blocks concurrently, causing a SIGBUS or SIGSEGV immediately after parsing during a self-host build.

Store the recycle-cache pointer on `VMemoryBlock` instead. Every block inherits it from the existing TLS arena chain, so the cache stays thread-local even when the source is compiled by an older bootstrap compiler. Remove the no-longer-needed cache-name special cases from both C generators and add a concurrent scoped-recycling regression test.

Validation:

- reproduced the crash with an older bootstrap compiler and confirmed `g_prealloc_block_cache` was emitted in `__DATA,__bss`
- rebuilt with that same compiler after this change; the standalone cache symbol is gone and nine consecutive v3 self-builds passed
- `../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -nocache -building-v -o v4 v3.v`
- `./vnew -prealloc test vlib/builtin/prealloc_scope_ownership_test.v`
- `./vnew -silent test vlib/v/` (2306 passed, 22 skipped)

`test-all` was not run.

### DIFF
--- a/vlib/builtin/prealloc.c.v
+++ b/vlib/builtin/prealloc.c.v
@@ -40,7 +40,6 @@ const prealloc_scope_block_size = 256 * 1024
 const prealloc_default_align = sizeof(voidptr) * 2
 
 __global g_memory_block &VMemoryBlock
-__global g_prealloc_block_cache &VPreallocBlockCache
 __global g_prealloc_allocation_count i64
 __global g_prealloc_allocated_bytes i64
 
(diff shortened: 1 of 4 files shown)


medvednikov:
@codex review

PR #27949, opened 2026-07-26
arrays: handles arrays.window invalid attributes

tkhuu01:
# Summary
Makes `arrays.window` return an empty array when its attributes cannot produce a valid window:
- `step == 0` previously caused an infinite loop when `size` fit the input.                                                                                                                                                                                                                                                                                                                         
- `step < 0` previously moved the index backward and caused an out-of-range slice panic.
- `size == 0` previously returned `array.len + 1` empty windows.
- `size < 0` previously caused an invalid-slice panic.
- `size > array.len` previously requested a negative initial capacity.                                                                                                                                                                                               

# Tests
*  ./vnew vlib/arrays/arrays_test.v

### DIFF
--- a/vlib/arrays/arrays.v
+++ b/vlib/arrays/arrays.v
@@ -208,20 +208,21 @@ pub fn chunk_while[T](a []T, predicate fn (before T, after T) bool) [][]T {
 	return chunks
 }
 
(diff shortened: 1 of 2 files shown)


medvednikov:
@codex review

PR #27955, opened 2026-07-27
cgen: fix invalid C for type-alias'd fixed array struct-init literals

### LINKED ISSUE #27894: cgen: declaring a type alias as a fixed array of empty structs generates invalid C
### Describe the bug

If we declare a fixed array of **empty** structs as a type alias, C compilation fails.

- no issue with structs containing fields
- no issue with dynamic arrays
- no issue with empty interfaces or maps

### Reproduction Steps

```v
type Arr = [4]Box

struct Box {}

fn main() {
	_ := Arr{}
}
```

### Expected Behavior

success

### Current Behavior

```c
cc:  error: initializer for aggregate with no elements requires explicit braces
cc:  4443 |         {main__Arr _ = {0};}
cc:       |                         ^
cc: 1 error generated.
```

### Possible Solution

_No response_

_No response_

### V version

0.5.2

not relevant
> [!NOTE]
>

rilaaax:
Fixes #27894

## Problem

Struct-init literal syntax (`Arr{}`) on a type alias for a fixed array (`type Arr = [N]T`) generates invalid C in two situations:

1. When `T` is an empty struct, declaring, taking the address of, or discarding such a value emits `{0}` as initializer. This is invalid on gcc/clang, since the underlying C array has zero size. Also applies recursively to nested arrays (`[2][3]Box`).
2. Reassigning (not declaring) such an alias, e.g. `c = Arr{}` after `mut c := Arr{}`, emits a bare brace-init list outside of a declaration context (`c = {0};`), which is invalid regardless of element type.

## Cause

1. `struct_init()` (`struct.v`) and `fixed_array_init()` (`array.v`) each fall back to writing a literal `0` to zero-init an aggregate, without checking if the element type is itself an empty struct. The `E_STRUCT` macro already exists for this (expands to nothing on gcc/clang, `0` + padding field on msvc/tcc) and is already used by `type_default_impl()`, but not by these two call sites. `is_empty_struct_array()` also only checked one level deep.
2. `Arr{}` parses as `ast.StructInit`, not `ast.ArrayInit`, even though its type is a fixed array. `assign_stmt()` (`assign.v`) only routes `ast.ArrayInit` or specific lvalue-like expressions through the memcpy-based fixed-array assignment path; a `StructInit` here fell through to the generic `left = <expr>;` codegen.

## Fix

1. `struct_init()` and `fixed_array_init()` now check `sym.is_empty_struct_array()` (made recursive to cover nested arrays) and emit `E_STRUCT` instead of `0` when the element type is an empty struct.
2. `assign_stmt()` now also routes a `StructInit` fixed-array alias through the memcpy path on reassignment, materializing it into a temp var first (like the existing `ArrayInit` case). Declarations and blank assignments (`_ = Arr{}`, handled correctly by a separate path) are excluded.

## Test

Tests added.

### DIFF
--- a/vlib/v/ast/types.v
+++ b/vlib/v/ast/types.v
@@ -1177,6 +1177,9 @@ pub fn (t &TypeSymbol) is_empty_struct_array() bool {
 		if elem_sym.info is Struct {
 			return elem_sym.info.is_empty_struct()
 		}
+		if elem_sym.info is ArrayFixed {
+			return elem_sym.is_empty_struct_array()
+		}
 	}
 	return false
 }
--- a/vlib/v/gen/c/array.v
+++ b/vlib/v/gen/c/array.v
@@ -554,7 +554,18 @@ fn (mut g Gen) fixed_array_init(node ast.ArrayInit, array_type Type, var_name st
 			g.add_commas_and_prevent_long_lines(i, info.size)
 		}
 	} else if is_amp {
-		g.write('0')
+		is_empty := if elem_sym.info is ast.Struct {
+			elem_sym.info.is_empty_struct()
+		} else {
+			// covers nested fixed arrays of empty structs
+			elem_sym.is_empty_struct_array()
+		}
+		if is_empty {
+			// A fixed array of empty structs has no elements to zero-init in C
+			g.write('E_STRUCT')
+		} else {
+			g.write('0')
+		}
 	} else {
 		if elem_sym.kind == .map {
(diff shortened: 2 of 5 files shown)


medvednikov:
Sorry, I fixed it before seeing your PR :(

rilaaax:
> Sorry, I fixed it before seeing your PR :(

No problem, I fixed some other cases that were not mentioned in the issue, will open another one ;)

PR #27968, opened 2026-07-28
v3: pass all checker fixtures

medvednikov:
## What changed

- run v3 diagnostic fixture suites with exact `.vv`/`.out` output and expected-exit-status comparison
- align v3 scanner, parser, checker, formatter, template, and compile-time diagnostics with the established checker fixture expectations
- add recursive value formatting used by compatible diagnostics
- suppress secondary diagnostics only when they are genuine cascades, while preserving independent errors
- keep fixture execution serial so large checker sweeps have a bounded memory footprint

## Why

The v3 compiler could compile many checker fixtures, but its diagnostic text, ordering, source ranges, parser recovery, and cascade behavior differed from v1. The v3 test command also needed to compare compiler output and exit status against adjacent `.out` files. These differences prevented the complete `vlib/v/checker/tests` fixture suite from being used as an exact compatibility gate.

## Impact

`v3 test vlib/v/checker/tests` now validates every paired checker fixture and all 1,264 fixtures pass without changing their expected outputs. The runner executes one compiler child at a time to avoid the previous multi-process memory spikes.

## Validation

- built v3 with `../../v -gc none -prealloc -o v3 v3.v`
- `./v3 -silent -no-parallel -nocache test ../v/checker/tests`
  - `checker fixtures: 1264/1264 passed, 0 failed, 0 abnormal exits`
- formatting verified for all touched V files with `./vnew fmt -verify ...`
- `git diff --check`
- no `.out` files changed


medvednikov:
@codex review

PR #27973, opened 2026-07-29
v3: fix self-host OOM and five checker/cgen regressions blocking -building-v

medvednikov:
\`./v3 -nocache -building-v -o v4 -no-memory-limit v3.v\` was OOM-killed by macOS right after the parse phase (the process allocated at ~20 GB/s in the parallel checker until jetsam sent SIGKILL — \`-no-memory-limit\` only disables the internal watchdog, not the OS kill). Fixing the OOM then exposed five more self-host blockers. With this PR the full bootstrap chain is green again: v1 → v3(prod) → v4 → v5, ~1.8 s total, ~740 MB steady footprint.

### Memory explosion (checker)
- \`source_enclosing_fn_has_generic_param\` copied the whole source file twice per call (\`source[..offset]\`, \`source[fn_start..]\` — ~5 MB per call on checker.v) and was reached for every identifier in \`resolve_type\`/\`check_ident\` because the cheap \`is_bare_generic_param\` guard was applied after the expensive \`||\` chain. Now index-scans without copies, and both call sites short-circuit on the guard first (behavior-preserving: the chain's result was only ever consumed under that same guard).
- \`closest_identifier_span\` copied the file suffix once per occurrence of the searched name (O(occurrences × file size) bytes). Now uses \`index_after_\` and stops after the first word match past the anchor (later matches are strictly farther).

### Checker fixes
- \`s[a..n] or { ... }\` (gated range slice) was typed as the element type \`u8\` instead of the slice type; \`or_expr_payload_type\` now recognizes \`value == 'range'\` index sources (broke \`v/util/version.v\`).
- \`name.str\` on a string in infix position resolved to the \`str()\` method value (\`fn () string\`) instead of the \`&u8\` field; declared builtin-struct fields now take priority over builtin method values in \`selector_type\`.

### Transform fixes
- \`a, b := match e { ... else { return } }\`: the multi-value match lowering bailed out on branches whose tail exits (return/break/continue/noreturn), falling into a broken \`_ifexpr\` fallback. Exiting branches now keep their body unchanged with no assignment.
- Value-producing if-guards with multi-value destructuring (\`x := if a, b, _ := f() { ... }\`) bound the whole tuple to the first name; \`build_if_value_guard_chain\` now emits per-name \`.value.argN\` bindings like the statement path.

### Cgen fixes (-building-v)
- The fixed-storage-const optimization miscompiled \`in const_array\`: the lowered \`array_get(const, i)\` shares its const AST node with a \`.len\` selector base, which the safety scan had whitelisted, so the const was wrongly promoted to a C fixed array. Const references in plain call arguments now always count as dynamic uses.
- \`v.token\` newly entered the v3.v closure (checker.v imports \`v.util\`), and \`-building-v\` template erasure left \`scanner_matcher = new_keywords_matcher_trie(...)\` calling a deleted function (the keep-list preserves the template AST, but cgen never emits open generic templates). Const initializers that call erased templates are now dropped with their storage; the compiler never reads them, and any new use fails loudly at C compile time instead.

Each fix was reduced to a minimal repro first (match-tuple exit branches, guard destructuring, slice-or, \`.str\` infix, \`import v.token\` under \`-building-v\`), and all repros compile and run correctly after the fix.

### DIFF
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -9218,6 +9218,23 @@ fn (mut g FlatGen) collect_fixed_storage_consts() {
 						}
 					}
 				}
+				// A const passed as a plain call argument (e.g. the lowered
+				// `array_get(const, i)` from a containment loop) needs dynamic
+				// array representation even when another use site aliases the
+				// same node as an index or `.len` base and marked it safe.
+				for ai in 1 .. node.children_count {
+					arg_id := g.a.child(node, ai)
+					arg_node := g.a.node(arg_id)
+					if g.const_ref_node_may_match_fixed_candidate(arg_node,
+						fixed_candidate_idents, fixed_candidate_shorts)
+					{
+						call_base_items << FixedStorageConstRefItem{
+							id:     arg_id
+							file:   cur_file
+							module: cur_module
+						}
+					}
+				}
 			}
 			.ident, .paren {
 				if g.const_ref_node_may_match_fixed_candidate(node, fixed_candidate_idents,
--- a/vlib/v3/transform/if.v
+++ b/vlib/v3/transform/if.v
@@ -973,15 +973,39 @@ fn (mut t Transformer) build_if_value_guard_chain(if_node flat.Node, target_name
 	t.drain_pending(mut result)
 	result << t.make_decl_assign_typed(tmp_name, rhs_expr, rhs_type)
 	ok_cond := t.make_selector(t.make_ident(tmp_name), 'ok', 'bool')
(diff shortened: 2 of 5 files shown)


medvednikov:
Rebased on master. Note: the rebase surfaced a separate master regression from 79e0c1c663 — permanently arming `qualify_name_cache` on the main checker segfaults every `-prealloc` build (scoped driver stages grow the cache map inside disposable scope arenas; `freeze_interface_impl_names` later reads freed buckets in `map_get_check`). Fixed here as a second commit by restoring the nil-unless-armed discipline; parallel cgen forks keep their private armed instances. Verified: v1 → v3 → v4 → v5 bootstrap chain green with `-gc none -prealloc`.

PR #27979, opened 2026-07-29
term.ui: fix CJK/IME input on Windows (fake control key events, lost console mode flags)

maaamahAhh:
This fixes two related bugs in `term.ui` on Windows that made
CJK (Chinese/Japanese/Korean) input via an IME (input method editor)
unusable:

1. In `init()`, the second `SetConsoleMode` call completely replaced the
   mode set by the first one, so `ENABLE_EXTENDED_FLAGS` (0x80) was lost
   immediately after being set. `SetConsoleMode` replaces the whole input
   mode instead of extending it, so all flags are now set in a single
   call. The `0x80` literal was replaced with the equivalent
   `C.ENABLE_EXTENDED_FLAGS` constant, matching the other flags in the
   same expression (which also made the comment explaining the magic
   number unnecessary).

2. In `parse_events()`, the fallback branch of the key code match mapped
   the low byte of the character (`uChar.AsciiChar`) to a `KeyCode`. For
   non-ASCII characters (e.g. from CJK IME input), this produced bogus
   `KeyCode` values, including fake control key events: for example `创`
   (U+521B) or `唛` (U+551B) have a low byte of `0x1B`, so typing them
   generated a spurious `.escape` key_down event. Such characters now
   report `KeyCode.null`, while the character itself is still delivered
   correctly through the `utf8` field. ASCII input is unaffected.

Tested on Windows 10 (22H2) with a Chinese IME, using a small `term.ui`
echo program: IME composition works, and characters with a `0x1B` low
byte like `创` (U+521B) now produce `code=null` events instead of fake
`.escape` events, which previously made TUI applications quit.
`v fmt -verify` passes and `v test vlib/term/ui` is green, including a
new `input_windows_test.v` that covers the `KeyCode` mapping fix.

### DIFF
--- a/vlib/term/ui/input_windows.c.v
+++ b/vlib/term/ui/input_windows.c.v
@@ -52,15 +52,13 @@ pub fn init(cfg Config) &Context {
 	}
 
 	// enable extended input flags (see https://stackoverflow.com/a/46802726)
-	// 0x80 == C.ENABLE_EXTENDED_FLAGS
-	if !C.SetConsoleMode(stdin_handle, 0x80) {
-		panic('could not set raw input mode')
-	}
-	mut input_mode := u32(C.ENABLE_WINDOW_INPUT)
+	mut input_mode := u32(C.ENABLE_EXTENDED_FLAGS) | u32(C.ENABLE_WINDOW_INPUT)
 	if ctx.cfg.mouse_enabled {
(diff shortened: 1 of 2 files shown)


medvednikov:
@codex review

medvednikov:
Thanks!

PR #28004, opened 2026-07-31
parser, v3: show where the conflicting signature is defined in C fn redeclaration errors

medvednikov:
## Motivation

When a `C.` function is declared with incompatible signatures in two different modules, the compiler reports:

```
/root/v/vlib/os/os.c.v:29:6: error: C function `C.open` was already declared with a different signature
```

The error only points at the *second* declaration, leaving no hint about *where the conflicting signature* was defined. That makes these conflicts (often between two third-party modules) hard to track down.

## Change

Both the mainline compiler (`vlib/v`) and `vlib/v3` now include the module and location of the previously-seen declaration:

**Before**
```
modb/modb.v:3:6: error: C function `C.getpid` was already declared with a different signature
```

**After**
```
modb/modb.v:3:6: error: C function `C.getpid` was already declared with a different signature in module `moda` at moda/moda.v:3:6
```

- `vlib/v/parser/fn.v`: the existing `ast.Fn` already carries `file`/`name_pos`/`mod`, so the location is built from those.
- `vlib/v3/types/checker.v`: `check_c_fn_redeclarations` now tracks the position string of the declaration whose signature is stored, and references it when reporting the conflict.

## Tests

Extended the existing redeclaration tests in both `vlib/v/checker/c_fn_redeclaration_test.v` and `vlib/v3/tests/c_fn_redeclaration_and_empty_struct_test.v` to assert the referenced location/module now appears in the output. Both pass.

### DIFF
--- a/vlib/v/checker/c_fn_redeclaration_test.v
+++ b/vlib/v/checker/c_fn_redeclaration_test.v
@@ -26,8 +26,12 @@ fn test_conflicting_c_fn_redeclarations_across_modules_are_reported_at_declarati
 	}
 	result := os.execute('${c_fn_redeclaration_vexe} -check ${os.quoted_path(root)}')
 	assert result.exit_code != 0, result.output
-	assert result.output.contains('moda/moda.v:3:') || result.output.contains('modb/modb.v:3:')
 	assert result.output.contains('C function `C.getpid` was already declared with a different signature')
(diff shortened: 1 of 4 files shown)


medvednikov:
@codex review

medvednikov:
@codex review

PR #28006, opened 2026-08-01
v3: fix self-hosting and test suite regressions

medvednikov:
## Summary

- enable supported macOS `cmd/v` builds through the in-process V3 path while preserving `-old-compiler` as the compatibility path
- fix V3 driver, module-cache, parallel-phase, checker, transformer, markused, and C-generation regressions exposed by self-hosting and broad test coverage
- make `vlib/v3/test_all.vsh` deterministic and storage-bounded by running sorted unit-test batches with a suite-owned cache, and move non-test fixtures under `tests/testdata`
- add focused regression coverage for imported sum-type map assignments, generic specialization collisions, diagnostic ownership, no-GC C output, and the other corrected language/codegen cases

## Root cause

V3 self-hosting exercised paths that the previous `cmd/v` dispatch excluded and exposed several missing semantic and code-generation cases. Cache plans and parallel stages also retained or shared state too broadly. In the full unit suite, one unbounded shared `V3CACHE` accumulated hundreds of generated C fragments until the test host ran out of disk space, while fixture files in the test directory could be mistaken for test inputs.

## Impact

Unflagged supported `v cmd/v` builds can use V3 on macOS, while `v -old-compiler cmd/v` remains available. V3 cache reuse and parallel phases are safer, diagnostics are scoped to the correct function, and the V3 test suite has bounded temporary storage.

## Validation

- built `cmd/v` through both V3 and `-old-compiler`
- all 239 discovered V3 unit-test files passed across staged, isolated batch runs
- a repeated integrated `./vnew vlib/v3/test_all.vsh` run passed batches 1-6 and the first 5 tests in batch 7 before being intentionally stopped at the user's request
- focused monomorphization, generic collision, C-output/no-GC, post-merge diagnostic, preallocation, and veb implicit-context regressions passed
- pre-commit formatter confirmed all touched V/VSH files were already formatted
- `git diff --check` passed


medvednikov:
@codex review

PR #28009, opened 2026-08-01
fix(parser): avoid panic after asm lock prefix

ShiroKSH:
## Summary

- prevent the parser from indexing an empty token after an x86 `lock` prefix
- add a parser-error regression fixture for a truncated inline-assembly block

## Root cause

After parsing `lock`, the parser advances to EOF for truncated input and indexed the empty token literal to check an instruction suffix. This crashed the compiler instead of reporting a syntax error.

## Validation

- `./vnew -silent test vlib/v/parser/`
- `./vnew -silent test vlib/encoding/base58/`
- `./vnew -silent test vlib/encoding/binary/`

### DIFF
--- a/vlib/v/parser/asm.v
+++ b/vlib/v/parser/asm.v
@@ -90,7 +90,8 @@ fn (mut p Parser) asm_stmt(is_top_level bool) ast.AsmStmt {
 			if p.tok.kind == .key_lock && arch in [.i386, .amd64] {
 				p.next()
 
-				has_suffix := p.tok.lit[p.tok.lit.len - 1] in [`b`, `w`, `l`, `q`]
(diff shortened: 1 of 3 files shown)


medvednikov:
@codex review

ShiroKSH:
Rebased onto current `master`. The previous failing jobs were unrelated to this parser change: the Ubuntu musl runner lacked `sys/cdefs.h`, and the OpenBSD tools suite saw an unrelated `vtest_test` exit-code mismatch. Verified the updated parser change locally with:

`./vnew.exe -silent test vlib/v/parser`

Result: 9 parser tests passed.

medvednikov:
Sorry for the delay.

PR #28016, opened 2026-08-03
v3: speed up self-host compiler stages

medvednikov:
## Summary

- reduce repeated AST scans, string allocations, cache cloning, and redundant map work in V3 checking
- share frozen transform context across workers and merge only fresh worker cache ranges
- batch C generation work and cheaply filter fixed-storage candidates before expensive name lookups
- preserve qualified function ownership during mark-used pruning and add a regression test

## Why

V3 self-host compilation had regressed to roughly 1.99 seconds across check, transform, and C generation. The hot paths repeatedly scanned large AST regions, allocated short/owner name strings, rebuilt worker-local indexes, and merged cache entries that could not overlap.

These changes keep the existing semantics while reusing read-only indexes, reducing allocations under `-prealloc`, and increasing useful work per parallel dispatch.

## Impact

Using:

```sh
cd vlib/v3
../../v -gc none -prealloc -prod -o v3 v3.v
./v3 -nocache -building-v -o v4 v3.v
```

A representative run measured:

- check: 265.81 ms
- transform: 269.65 ms
- cgen: 240.97 ms
- combined: 776.43 ms

After rebasing the work onto merged `master`, three warm runs measured 769.35, 789.25, and 805.96 ms (789.25 ms median). Wall-clock timing is sensitive to concurrent workstation load.

## Validation

- `./vnew -silent vlib/v3/tests/markused_test.v`
- `./vnew -silent vlib/v3/tests/fixed_array_const_materialization_codegen_test.v`
- `./vnew -silent vlib/v3/types/checker_parallel_test.v`
- `./vnew -silent vlib/v3/transform/transform_parallel_test.v`
- `./vnew -silent vlib/v3/tests/parallel_determinism_test.v`
- `git diff --check`

All focused tests pass on the merged `master` base.

### DIFF
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -28,6 +28,17 @@ const c_common_c_attributes = ['alias', 'aligned', 'always_inline', 'cold', 'con
 const c_has_attribute_predicate = '__has_attribute'
 const c_has_attribute_override_key = '@function:__has_attribute'
 
+// c_short_name_view returns the suffix after the final dot without allocating.
+@[direct_array_access; inline]
+fn c_short_name_view(name string) string {
+	for i := name.len - 1; i >= 0; i-- {
+		if name[i] == `.` {
+			return unsafe { name.substr_unsafe(i + 1, name.len) }
+		}
+	}
+	return name
+}
+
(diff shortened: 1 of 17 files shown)


medvednikov:
@codex review

PR #28021, opened 2026-08-04
arrays: add merge_desc for descending sorted arrays

ruedigerlenz:
## What does this PR do?

Add `arrays.merge_desc`, which merges two descending-sorted arrays while
preserving their order.

## Tests

All tests passed.


tankf33der:
Tested manually.

PR #28024, opened 2026-08-04
vself: do not default to tcc on macOS arm64 when it would just fail

### LINKED ISSUE #28023: mac: `v self` not working with `tcc`
### Describe the bug

`v self` fails to build with `tcc` (which is the default)

### Reproduction Steps

`v self`

### Expected Behavior

success

### Current Behavior

```
V self compiling (-cc tcc -gc none -o v2)...
warning: tcc compilation failed, falling back to cc
```

### Possible Solution

_No response_

_No response_

### V version

0.5.2 a5a89d5

|V full version      |V 0.5.2 76f77e39ed0451ae813e24d8e85d52b569b073c6.a5a89d5
|OS                  |macos, macOS, 27.0, 26A5388g
|                    |
|V git status        |0.5.2-210-ga5a89d56
|                    |
|gcc version         |Apple clang version 21.0.0 (clang-2100.1.1.101)
|clang version       |Apple clang version 21.0.0 (clang-2100.1.1.101)
|tcc version         |tcc version 0.9.28rc 2026-07-24 HEAD@85ba3ae8 (AArch64 Darwin)
|tcc git status      |thirdparty-macos-arm64 274abd24
|emcc version        |N/A
|glibc version       |N/A
> [!NOTE]
>

quaesitor-scientiam:
## Summary
- A plain `v self` on Apple Silicon macOS always fails to compile with the bundled `tcc` and silently falls back to `cc`, as reported in #28023.
- Root cause: `Preferences.fill_with_defaults()` forces `-prealloc` on for macOS builds of V itself when no real GC is selected (exactly what `v self`'s implicit `-gc none` does). `-prealloc` declares its allocator global `g_memory_block` as `_Thread_local`, a C11 keyword the bundled tcc does not implement.
- The builder's own `try_to_use_tcc_by_default()` already avoids defaulting to tcc when `-prealloc` is set, but `cmd/tools/vself.v` bypasses that guard by unconditionally injecting an explicit `-cc tcc` for Apple Silicon before the guard ever runs.
- Fix: only add `-cc tcc` in `vself.v` when the caller explicitly requested a real GC via `-gc` (the one case where `-prealloc` isn't forced and tcc remains compatible). Plain `v self` (and `v self -gc none`) now goes straight to `cc` with no failing attempt; `v self -gc boehm` still gets the fast tcc path.

## Test plan
- [x] Reproduced the failure on macOS/arm64 (`v self` → `warning: tcc compilation failed, falling back to cc`), traced it to the `_Thread_local g_memory_block` declaration via `-keepc`.
- [x] Verified `v self`, `v self -gc none`, and `v self -gc boehm` all build cleanly after the fix, with correct compiler selection in each case.
- [x] `v test vlib/v/builder/cc_tcc_retry_test.v` passes.
- [x] `v fmt -verify cmd/tools/vself.v` passes.

Fixes #28023

### DIFF
--- a/cmd/tools/vself.v
+++ b/cmd/tools/vself.v
@@ -32,10 +32,14 @@ fn main() {
 		// compiling by default, i.e. `v self`:
 		uos := os.user_os()
 		uname := os.uname()
-		if uos == 'macos' && uname.machine == 'arm64' {
+		if uos == 'macos' && uname.machine == 'arm64' && gc_mode_from_args(args) !in ['', 'none'] {
 			// Apple silicon, like m1, m2 etc
 			// Use tcc by default for V, since tinycc is much faster and also
-			// it already supports compiling many programs like V itself, that do not depend on inlined objective-C code
+			// it already supports compiling many programs like V itself, that do not depend on inlined objective-C code.
+			// Note: a plain `v self` (no explicit -gc) forces Preferences.prealloc on
+			// for macOS builds of V itself, which declares its allocator state
+			// `_Thread_local` - a keyword the bundled tcc cannot parse. Only default
+			// to tcc here when an explicit non-`none` `-gc` rules that out.
 			args << '-cc tcc'
 		} else if uos == 'linux' && uname.machine in ['arm64', 'aarch64'] {
 			// Bundled TCC can hang while bootstrapping V on Linux ARM64, so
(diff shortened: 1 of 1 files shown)


medvednikov:
Not the right fix. Need to use tcc by default.

quaesitor-scientiam:
> Not the right fix. Need to use tcc by default.

fixes the warning which is false because it currently bypasses the guard the tcc issue i found is at https://github.com/vlang/tccbin/issues/86

JalonSolov:
It may fix the warning, but the _right_ fix will be to make tcc work as expected, rather than just blocking it.

quaesitor-scientiam:
> It may fix the warning, but the _right_ fix will be to make tcc work as expected, rather than just blocking it.

Requires https://github.com/vlang/tccbin/issues/86 to be completed

quaesitor-scientiam:
See comments https://github.com/vlang/v/issues/28023#issuecomment-5181766311

PR #28031, opened 2026-08-05
cgen: fix undeclared result temp for `or {}` block containing a comptime `$if/$else`

### LINKED ISSUE #28022: cgen: `or {}` block containing a comptime `$if/$else` emits an undeclared result temp (`_t3`)
**V version:** `V 0.5.2` (master)
**OS:** Linux (reported via bugs.vlang.io)

### Description

An `or {}` block whose body is a compile-time `$if / $else` emits an **undeclared result temp**: the branch bodies assign to and reference `_t3`, but `_t3` is never declared, so the generated C fails with `use of undeclared identifier '_t3'`.

### Reproduction

```v
module main

fn g(i int) !int {
	return error('!')
}

x := g(1) or {
	$if !windows {
		panic(err)
	} $else {
		u32(0)
	}
}
```

### Actual (generated C fails to compile)

```
error: use of undeclared identifier '_t3'
    _t3 = builtin___v_panic(builtin__IError_str(err));
    ...
    _t3;
```

The `or {}` result temp `_t3` is used in the `$if`/`$else` branch bodies but never declared.

### Expected

Compiles; the `or {}` block's result temp is declared for the comptime-`$if` body.

_Reported via bugs.vlang.io (also seen inside a larger real program)_

> [!NOTE]
>

medvednikov:
## Fixes #28022

An `or {}` block whose body is a compile-time `$if / $else` could emit an **undeclared result temp**, producing C that fails to compile with `use of undeclared identifier '_t3'`.

### Reproduction

```v
fn g(i int) !int {
	return error('!')
}

x := g(1) or {
	$if !windows {
		panic(err)
	} $else {
		u32(0)
	}
}
```

### Root cause

In `comptime_if()` (`vlib/v/gen/c/comptime.v`), when a `$if` is used as an expression whose overall type is `void`, the result type is inferred by scanning each branch's last statement. In this scenario:

- The **live** branch (`$if !windows`) ends in `panic(err)`, whose type is `void`.
- The **value-producing** `$else` branch (`u32(0)`) is a *dead* comptime branch and is pruned to zero statements before codegen, so its type is never seen.

Inference therefore yields `void`, and the result-temp declaration is skipped — yet the code still generated `<tmp> = ...` per-branch assignments and a trailing `<tmp>` reference, leaving the temp undeclared.

### Fix

Introduce `gen_as_expr := node.is_expr && inferred_typ != ast.void_type` and gate the three expression-style emissions (the temp declaration, the per-branch `<tmp> = ...` assignment, and the trailing `<tmp>` reference) on it. When the inferred type is `void`, codegen falls back to the plain statement path, emitting the branch body (e.g. `panic(err)`) as a bare statement with no result temp. This matches the runtime semantics: the only live branch never yields a value.

### Testing

- Added `vlib/v/tests/comptime/comptime_if_in_or_block_test.v`, which fails on the unfixed compiler with the exact `_t3 undeclared` C error and passes with the fix. The panicking branch is kept unreachable at runtime.
- `vlib/v/tests/comptime/` — 171/171 pass.
- `vlib/v/tests/` — the only failures are pre-existing `vls/` language-server tests, which fail identically on the unfixed compiler and are unrelated to this change.

---
_Generated by [Claude Code](https://claude.ai/code/session_0187bMp9ApbDVDJqkNVjPppf)_

### DIFF
--- a/vlib/v/gen/c/comptime.v
+++ b/vlib/v/gen/c/comptime.v
@@ -848,7 +848,14 @@ fn (mut g Gen) comptime_if(node ast.IfExpr) {
 	}
 	is_opt_or_result := inferred_typ.has_option_or_result()
 	is_array_fixed := g.table.final_sym(inferred_typ).kind == .array_fixed
-	line := if node.is_expr && inferred_typ != ast.void_type {
+	// When the if is used as an expression but no branch yields a concrete
+	// (non-void) type - e.g. an `or {}` body whose only live comptime branch
+	// ends in `panic(err)` while the value-producing branch was pruned as dead
+	// code - there is no result temp to declare. Generating `${tmp_var} = ...`
+	// and a trailing `${tmp_var}` reference would then emit an undeclared temp
+	// (see issue #28022). Fall back to plain statement codegen in that case.
+	gen_as_expr := node.is_expr && inferred_typ != ast.void_type
+	line := if gen_as_expr {
 		stmt_str := g.go_before_last_stmt()
 		g.write(util.tabs(g.indent))
 		styp := g.styp(inferred_typ)
@@ -914,7 +921,7 @@ fn (mut g Gen) comptime_if(node ast.IfExpr) {
 			}
 			g.defer_ifdef += expr_str
 		}
-		if node.is_expr {
+		if gen_as_expr {
 			if is_true.val {
 				g.bind_comptime_if_generic_types(branch.cond)
 				len := branch.stmts.len
@@ -994,7 +1001,7 @@ fn (mut g Gen) comptime_if(node ast.IfExpr) {
 	}
 	g.defer_ifdef = ''
 	g.writeln('#endif')
-	if node.is_expr {
+	if gen_as_expr {
 		g.write('${line}${tmp_var}')
 	}
 }
--- a/vlib/v/gen/c/comptime.v
+++ b/vlib/v/gen/c/comptime.v
@@ -848,12 +848,17 @@ fn (mut g Gen) comptime_if(node ast.IfExpr) {
 	}
 	is_opt_or_result := inferred_typ.has_option_or_result()
 	is_array_fixed := g.table.final_sym(inferred_typ).kind == .array_fixed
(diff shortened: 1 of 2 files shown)


medvednikov:
Thanks for the thorough review — both points addressed in the follow-up commit.

**1 (P2) — per-branch decision.** The global `gen_as_expr` predicate now only controls the result *temp* (whether it's declared and referenced at the end, based on whether *any* branch yields a value). Whether an individual branch *assigns* to that temp is decided per branch, from that branch's own final statement:

```v
mut branch_produces_value := false
if gen_as_expr && branch.stmts.len > 0 {
    branch_last := branch.stmts.last()
    if branch_last is ast.ExprStmt {
        branch_typ := g.type_resolver.get_type_or_default(branch_last.expr, branch_last.typ)
        branch_produces_value = branch_typ != ast.void_type && !branch_typ.has_flag(.generic)
    }
}
```

A void/noreturn branch (`panic(err)`) now falls to the statement path even when a different retained branch supplies the value type, and value branches are emitted under cross mode too. Verified with `-os cross` (using `int(0)` in the `$else`, since the report's literal `u32(0)` is rejected earlier by the checker as `wrong return type u32 ... expected int`):

```c
int _t3;
#if !defined(_WIN32)
{
	builtin___v_panic(builtin__IError_str(err));
	VUNREACHABLE();
}
#else
	_t3 = ((int)(0));
#endif
*(int*) _t3_data = _t3;
```

The invalid `_t3 = builtin___v_panic(...)` is gone, and the value branch is no longer dropped.

**2 (P3) — platform-independent regression.** The runtime test now uses `$if true { panic(err) } $else { int(0) }`, so the void branch is selected on every target. I also added an `output_cross_c` codegen regression that runs `v -os cross` and asserts the void branch is *not* assigned to the result temp (`assert !generated.contains('= builtin___v_panic')`) — this fails on the pre-fix compiler and passes now.

`vlib/v/tests/comptime/` (171), the or-block tests, and `vlib/v/gen/c/coutput_test.v` all pass.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #28032, opened 2026-08-05
cmd/v: restore macOS V3 helpers for cross-generated VC

medvednikov:
`v -cross cmd/v` excludes `macos_v3_darwin.c.v`, but the generic ownership-dispatch code in `cmd/v/v.v` retains its macOS branches in the generated cross-platform C source. Those branches reference `macos_v3_has_v1_only_leading_option` and `macos_v3_forwarded_args`, so `VC gen / build-vc` stopped with unknown-function errors while regenerating `vc/v.c`.

Move the platform-independent constants and argument helpers into `cmd/v/macos_v3_args.c.v`, which is compiled alongside both the native Darwin dispatcher and the default implementation selected during cross generation. This restores the missing cross-generated symbols without maintaining a second helper implementation; the existing macOS argument tests now exercise the same code used by generated VC.

### Testing

- [x] `VC gen / build-vc` — regenerated both `v.c` and `v_win.c`
- [ ] `Bootstrapping CI / bootstrap-v (ubuntu-latest)` — in progress; build passed and cross bootstrap is running
- [x] `Bootstrapping CI / bootstrap-v (macos-14)`
- [x] `Cross CI / cross-windows`
- [x] `Cross CI / cross-linux`
- [x] `Cross CI / cross-macos`

### DIFF
--- a/cmd/v/macos_v3_args.c.v
+++ b/cmd/v/macos_v3_args.c.v
@@ -1,0 +1,80 @@
+module main
+
+import v.pref
+
+const macos_v3_compat_c99_flag = '-macos-v3-compat-c99'
+const macos_v3_internal_quiet_flag = '-macos-v3-internal-quiet'
+
+// These helpers are shared by the native Darwin dispatcher and the default
+// implementation selected while generating cross-platform VC sources.
+fn macos_v3_has_v1_only_leading_option(args []string, command string) bool {
+	mut i := 0
+	for i < args.len {
+		arg := args[i]
+		if arg == '--' {
+			return false
+		}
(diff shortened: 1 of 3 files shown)


medvednikov:
Addressed the non-blocking drift concern: the two constants and three argument helpers now live in `cmd/v/macos_v3_args.c.v`, which is compiled by both the native Darwin dispatcher and the default/cross-generated path. The existing macOS helper tests therefore exercise the same implementation used by generated VC. Current-head VC generation, bootstrapping, and cross-CI reruns are queued.

PR #28048, opened 2026-08-07
cmd/v: guard macOS V3 helpers with $if macos to fix unused-declaration notices

quaesitor-scientiam:
## Summary
- `cmd/v/macos_v3_args.c.v` keeps its helpers/consts platform-neutral (no `_darwin.c.v` suffix) so `v -os cross` still codegens the macOS branch when producing VC sources on non-macOS hosts, but that left the consts/functions unguarded at file scope, triggering unused-declaration notices on non-macOS hosts since their only callers are themselves macOS-gated (`cmd/v/v.v`'s `autofree_args_require_standard_compiler` / `v3_ownership_forwarded_args`). Wraps the import, consts, and function definitions in a single top-level `$if macos {}` block.
- `cmd/v/macos_v3_test.v` had one call to `is_macos_v3_relevant_command` (defined in the macOS-only `macos_v3_darwin.c.v`) missing the `$if macos` guard every sibling assertion in the file already uses, which failed compilation on non-macOS hosts (`unknown function: is_macos_v3_relevant_command`).

## Test plan
- [x] Native Windows: `v self` clean, no notices
- [x] `-os cross`: exits 0; diffed generated `v.c` against baseline — only change is the three helper definitions gaining a matched `#if defined(__APPLE__)`/`#endif`, call sites unaffected
- [x] `v test cmd/v/` on Windows: 2 passed, 2 total (was 1 passed / 1 failed at HEAD)
- [x] Native Linux (WSL): `v self` clean; `v test cmd/v/macos_v3_test.v cmd/v/vvm_test.v`: 2 passed, 2 total
- [x] `v fmt -w` on both touched files: no changes needed

### DIFF
--- a/cmd/v/macos_v3_args.c.v
+++ b/cmd/v/macos_v3_args.c.v
@@ -1,80 +1,90 @@
 module main
 
-import v.pref
-
-const macos_v3_compat_c99_flag = '-macos-v3-compat-c99'
-const macos_v3_internal_quiet_flag = '-macos-v3-internal-quiet'
-
 // These helpers are shared by the native Darwin dispatcher and the default
-// implementation selected while generating cross-platform VC sources.
-fn macos_v3_has_v1_only_leading_option(args []string, command string) bool {
-	mut i := 0
-	for i < args.len {
-		arg := args[i]
-		if arg == '--' {
-			return false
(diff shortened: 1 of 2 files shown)


quaesitor-scientiam:
All CI failures are  not related to this PR

PR #28051, opened 2026-08-08
v3: fix assert truncating a negated wide integer literal to 32 bits

### LINKED ISSUE #28050: v3: assert captures a negated out-of-range integer literal into an `int` temp, truncating the comparison
### Describe the bug

On macOS arm64, a bare (uncast) integer literal that does not fit in 32 bits is truncated to its low 32 bits when it is compared against an `i64` **inside `assert`**. The comparison yields the wrong answer, and `assert`'s own diagnostic prints the literal as a different number than the one written in the source:

```
assert get_plain() == -123456789012345
   left value: get_plain() = -123456789012345
  right value: -123456789012345 = 2045911175
```

`2045911175` is exactly `-123456789012345 & 0xFFFFFFFF`.

The same comparison written in a plain `if` evaluates **correctly** on the same platform, so this appears specific to the code `assert` generates rather than to comparison codegen in general.

The same file passes on `ubuntu-latest` (x86_64), so identical source produces different results per architecture.

Possibly related: V's checker *rejects* the very same literal when it is bound rather than compared —

```v
const c = -123456789012345  // error: overflow in implicit type `int`, use explicit type casting instead
y := -123456789012345       // error: overflow in implicit type `int`, use explicit type casting instead
```

so the literal is being typed as `int` in those positions and correctly diagnosed, while the `assert` comparison position silently accepts it and then truncates. Promoting the literal to the other operand's type (or raising the same overflow error) in that position would presumably fix it.

### Reproduction Steps

`probe/literal_test.v`:

```v
fn get() !i64 {
	return i64(-123456789012345)
}

fn get_plain() i64 {
	return i64(-123456789012345)
}

// FAILS on macOS arm64
fn test_unwrap_eq_bare_literal() ! {
	assert get()! == -123456789012345
}

// FAILS on macOS arm64
fn test_plain_eq_bare_literal() {
	assert get_plain() == -123456789012345
}

// PASSES — same comparison outside of assert
fn test_if_comparison() {
	mut hit := false
	if get_plain() == -123456789012345 {
		hit = true
	}
	assert hit, 'if-comparison against bare literal did not match'
}

// PASSES — explicit cast on both sides (the workaround)
fn test_unwrap_eq_cast_literal() ! {
	assert get()! == i64(-123456789012345)
}
```

```sh
v test probe/
```

Runnable branch, with a CI workflow that runs it on macos-latest and ubuntu-latest side by side:
https://github.com/we-be/protobuf.v/blob/probe-i64-literal/probe/literal_test.v

### Expected Behavior

All four tests pass. `-123456789012345` is representable as `i64`, the left operand is `i64`, and the same literal compares correctly in a plain `if` on the same machine — so the `assert` comparison should agree.

(Alternatively, if a bare literal in this position is genuinely meant to be `int`, it should raise the same `overflow in implicit type int` error the checker already raises for `const c = -123456789012345`, rather than silently truncating.)

### Current Behavior

macOS arm64:

```
FAIL    21.020 ms probe/literal_test.v
V panic: Assertion failed...
probe/literal_test.v:16: fn test_unwrap_eq_bare_literal
assert get()! == -123456789012345
  right value: -123456789012345 = 2045911175
V panic: Assertion failed...
probe/literal_test.v:21: fn test_plain_eq_bare_literal
assert get_plain() == -123456789012345
   left value: get_plain() = -123456789012345
  right value: -123456789012345 = 2045911175
Summary for all V _test.v files: 1 failed, 1 total.
```

`test_if_comparison` and `test_unwrap_eq_cast_literal` pass.

ubuntu-latest x86_64: all four pass.

CI run showing both platforms on the same commit:
https://github.com/we-be/protobuf.v/actions/runs/31235511455

### Possible Solution

In the `assert` comparison position, infer the untyped integer literal from the other operand's type (here `i64`) instead of defaulting it to `int`. If defaulting to `int` is intended, apply the existing `overflow in implicit type int` check there too so it fails loudly at compile time instead of silently truncating at runtime.

- Found while building a protobuf library for V, where test vectors compare decoded `i64`/`sint64` values against literals. Real impact is limited by the workaround (cast both sides), but it is easy to hit and fails only on Apple Silicon, so it presents as a mysterious platform-specific test failure.
- Reproduced on three separate CI runs over three days (2026-08-05, 2026-08-08 ×2), each against a freshly downloaded `refs/heads/master` tarball.
- Not reproducible on x86_64 Linux with either gcc 16.1.1 or clang 22.1.4.

### V version

V 0.5.2, built from `https://codeload.github.com/vlang/v/legacy.tar.gz/refs/heads/master` on 2026-08-08 (installed by `vlang/setup-v@v1`; the tarball has no `.git`, so `v version` reports a bare `V 0.5.2` with no commit hash).

Locally on Linux (where it does **not** reproduce): `V 0.5.2 e65ec29e8738bfc1875d722161b9babae372eb86`.

Failing platform — GitHub Actions `macos-latest`:

```
Image: macos-26-arm64 / 20260728.0273
Arch:  arm64 (Apple Silicon)
V:     V 0.5.2 (master tarball, 2026-08-08)
```

Passing platform — GitHub Actions `ubuntu-latest`:

```
Image: ubuntu24 / 20260720.247
Arch:  x86_64
V:     V 0.5.2 (master tarball, 2026-08-08)
```

Local machine where it also does not reproduce (`v doctor`):

```
V full version      | V 0.5.2 e65ec29e8738bfc1875d722161b9babae372eb86
OS                  | linux, "Bazzite" (Fedora Atomic)
Processor           | 12 cpus, 64bit, little endian, AMD Ryzen 5 3600 6-Core Processor
cc version          | cc (GCC) 16.1.1 20260515 (Red Hat 16.1.1-2)
clang version       | Homebrew clang version 22.1.4
glibc version       | ldd (GNU libc) 2.43
```

> [!NOTE]
>

hunterjsb:
Fixes #28050.

`assert` captures numeric operands into temps for its failure diagnostic, exempting literals. A signed literal parses as a `.prefix` over the literal, so it missed the exemption, resolved to the default `int`, and truncated:

```c
i64 _t1 = (i64)(get_plain());
int _t2 = (int)(-123456789012345);   // 2045911175
if (!(_t1 == _t2)) {
```

The comparison reads the temp, so the assert **evaluated** wrong, not just reported wrong. Only macOS delegates to v3 by default, which is why it looked platform-specific; driving v3 directly on Linux reproduces it.

**Fix:** exempt side-effect-free constant operands from capture: literals, sign/`~` prefixes, parens, and constant infix — `-9223372036854775807 - 1` (`math_test.v:1145`) is an infix, so a literal-only exemption missed it. The classic backend never captures these shapes (`assert_subexpression_to_ctemp` allowlists only side-effecting exprs), so this restores parity: assert evaluates constants exactly like a plain `if`. Non-constant operands still capture, preserving single evaluation of side effects.

**Diagnostics:** per review, display suppression is a separate, narrower predicate — only a bare or sign-prefixed literal drops its `= value` (the label already spells it out). `~5`, `-(-1)`, `(5)`, `2 * 3` keep their evaluated output; captured-then-truncated constants previously printed the truncated value and now print the true one.

**Affected in vlib today:** `time_test.v:505,512,518`, `enum_explicit_size_big_and_small_test.v:78-80`, `math_test.v:1145`.

**Tests:** `vlib/v3/tests/assert_negated_literal_codegen_test.v` inspects the emitted `if (!(` conditions for a negated literal and a constant product; each fails without its part of the fix. It asserts on emitted C rather than running a binary — v3's prelude omits `<stdlib.h>` and gcc 16 rejects the resulting implicit `qsort`, which also fails `assert_stderr_shadow_codegen_test.v` on master. The three `v3_assert_*` inout fixtures emit byte-identical C with and without this patch.

### DIFF
--- a/vlib/v3/gen/c/stmt.v
+++ b/vlib/v3/gen/c/stmt.v
@@ -2943,7 +2943,9 @@ fn (mut g FlatGen) gen_assert_capture_numeric_operands(condition_id flat.NodeId)
 	for operand_index in 0 .. 2 {
 		operand_id := g.a.child(condition, operand_index)
 		node := g.a.node(operand_id)
-		if node.kind in [.int_literal, .float_literal, .char_literal] || g.arg_is_const_ident(node) {
+		// exempt operands whose value a capture temp would truncate: constant
+		// numeric literals/exprs (ours) and const identifiers (upstream)
+		if g.is_constant_numeric_literal(operand_id) || g.arg_is_const_ident(node) {
 			continue
 		}
 		operand_type := g.usable_expr_type(operand_id)
@@ -2965,13 +2967,36 @@ fn (mut g FlatGen) gen_assert_capture_numeric_operands(condition_id flat.NodeId)
 	return captured_ids
 }
 
-fn (mut g FlatGen) gen_assert_numeric_value(prefix string, id flat.NodeId) {
+// A signed literal parses as a prefix over the literal; capturing it would type
+// it as the default `int` and truncate anything wider than 32 bits.
+fn (g &FlatGen) is_constant_numeric_literal(id flat.NodeId) bool {
 	node := g.a.node(id)
+	if node.kind in [.int_literal, .float_literal, .char_literal] {
+		return true
+	}
+	if node.children_count == 0 {
+		return false
+	}
+	return match node.kind {
+		.paren {
+			g.is_constant_numeric_literal(g.a.child(node, 0))
+		}
+		.prefix {
+			node.op in [.plus, .minus, .bit_not]
+				&& g.is_constant_numeric_literal(g.a.child(node, 0))
+		}
+		else {
+			false
+		}
+	}
+}
+
+fn (mut g FlatGen) gen_assert_numeric_value(prefix string, id flat.NodeId) {
 	label := g.assert_source_text(id)
 	if label.len == 0 {
 		return
 	}
-	if node.kind in [.int_literal, .float_literal, .char_literal] {
+	if g.is_constant_numeric_literal(id) {
 		g.writeln('v3_eprint_lit("${c_escape(prefix)}: ${c_escape(label)}\\n");')
 		return
 	}
--- a/vlib/v3/gen/c/stmt.v
+++ b/vlib/v3/gen/c/stmt.v
@@ -2967,8 +2967,9 @@ fn (mut g FlatGen) gen_assert_capture_numeric_operands(condition_id flat.NodeId)
 	return captured_ids
 }
 
-// A signed literal parses as a prefix over the literal; capturing it would type
-// it as the default `int` and truncate anything wider than 32 bits.
+// Capture eligibility: these are side-effect free, so skipping the temp is safe.
+// A signed literal parses as a prefix over the literal, and capturing it would
+// type it as the default `int` and truncate anything wider than 32 bits.
 fn (g &FlatGen) is_constant_numeric_literal(id flat.NodeId) bool {
 	node := g.a.node(id)
 	if node.kind in [.int_literal, .float_literal, .char_literal] {
@@ -2991,12 +2992,27 @@ fn (g &FlatGen) is_constant_numeric_literal(id flat.NodeId) bool {
 	}
 }
 
+// Display suppression: narrower than capture eligibility, because the label only
+// spells out the value for a bare or sign-prefixed literal. `~5` and `-(-1)`
+// still need evaluating.
+fn (g &FlatGen) is_signed_numeric_literal(id flat.NodeId) bool {
+	node := g.a.node(id)
+	if node.kind in [.int_literal, .float_literal, .char_literal] {
+		return true
+	}
+	if node.kind != .prefix || node.op !in [.plus, .minus] || node.children_count == 0 {
+		return false
+	}
+	operand := g.a.node(g.a.child(node, 0))
+	return operand.kind in [.int_literal, .float_literal, .char_literal]
+}
+
 fn (mut g FlatGen) gen_assert_numeric_value(prefix string, id flat.NodeId) {
 	label := g.assert_source_text(id)
 	if label.len == 0 {
 		return
 	}
-	if g.is_constant_numeric_literal(id) {
+	if g.is_signed_numeric_literal(id) {
 		g.writeln('v3_eprint_lit("${c_escape(prefix)}: ${c_escape(label)}\\n");')
 		return
 	}
--- a/vlib/v3/gen/c/stmt.v
+++ b/vlib/v3/gen/c/stmt.v
@@ -2942,8 +2942,7 @@ fn (mut g FlatGen) gen_assert_capture_numeric_operands(condition_id flat.NodeId)
 	mut captured_ids := []int{cap: 2}
 	for operand_index in 0 .. 2 {
 		operand_id := g.a.child(condition, operand_index)
-		node := g.a.node(operand_id)
-		if node.kind in [.int_literal, .float_literal, .char_literal] {
+		if g.is_constant_numeric_literal(operand_id) {
 			continue
 		}
 		typ := g.value_unalias_type(g.tc.resolve_type(operand_id))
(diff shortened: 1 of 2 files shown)


hunterjsb:
Two things I should disclose after reviewing this more adversarially.

**1. A diagnostic change beyond the reported bug.** Exempting `~` means a failing `assert f == ~5` now prints the operand's source text rather than its computed value:

```
before:   right value: ~5 = -6
after:    right value: ~5
```

For a sign-prefixed literal this is lossless (`-1` prints as `-1` either way), and it matches how a bare `.int_literal` already prints. For `~` it does lose the folded value. I included `.bit_not` because `~<wide literal>` truncates through the identical path, but it is the one case where the exemption costs information. There are currently **zero** `~`-literal asserts in vlib/cmd/examples, so this is theoretical — happy to narrow the patch to `.plus`/`.minus`/`.paren` only if you would rather keep the scope tight and leave `~<wide>` for a separate change.

More generally, a failing assert against any signed literal now prints `right value: -1` instead of `right value: -1 = -1`. No `.out` fixture in the repo depends on that format, and it makes signed literals consistent with bare ones.

**2. Scope is narrower than the issue title suggested.** Positive wide literals were never affected — they are bare `.int_literal` nodes and already hit the exemption. Only sign-prefixed (and parenthesized) literals reach the capture path. `assert factoriali(20) == 2432902008176640000` is correct on both sides of this patch.

The bug does reach vlib's own suite, though — these all exceed 32 bits and truncate under v3 today:

- `vlib/time/time_test.v:505,512` — `== -62167132800`
- `vlib/time/time_test.v:518` — `== -62135596800`
- `vlib/v/tests/enums/enum_explicit_size_big_and_small_test.v:78-80` — `== -999999999999` and friends
- `vlib/math/math_test.v:1145` — `== -9223372036854775807 - 1`

**Verification of no downstream effect:** I compiled the three v3 assert fixtures (`v3_assert_operand_once`, `v3_assert_percent_label`, `v3_assert_unsigned_value`) with and without the patch — the emitted C is byte-identical apart from the embedded version string. Operand capture for non-constant expressions is untouched, so single-evaluation of side-effecting operands (`values.pop()`) is preserved, and `-x` where `x` is a variable still captures normally.

medvednikov:
## Review result: **request changes**

CI ignored as requested. I reviewed the current PR head, `5889122c72e1ce0837dee1bc40c5d3c833de99c8`.

### 1. [P2] Separate capture eligibility from diagnostic suppression

**File:** `vlib/v3/gen/c/stmt.v:2995`

`is_constant_numeric_literal()` is appropriate for deciding whether an operand must be captured: it limits the exemption to numeric literals wrapped in parentheses or `+`, `-`, and `~`, so calls and other side-effecting expressions remain captured.

However, the same predicate is now used by `gen_assert_numeric_value()` to decide whether to omit the evaluated value. Because it recursively accepts `.bit_not` and nested prefixes, assertion diagnostics regress:

```text
right value: ~5 = -6
```

becomes:

```text
right value: ~5
```

Likewise, `-(-1)` loses its useful `= 1` output. The PR description acknowledges the `~5` behavior change, and the implementation takes the source-only branch for all expressions accepted by the helper.

The capture and display decisions should use separate predicates. Keep the broad helper in `gen_assert_capture_numeric_operands()`, but make the display predicate narrower—at minimum, do not suppress evaluation for `.bit_not` or arbitrary nested unary expressions.

### 2. [P2] Make the regression test inspect the generated condition

**File:** `vlib/v3/tests/assert_negated_literal_codegen_test.v:23-24`

This check scans the entire generated C file:

```v
assert code.contains('== -123456789012345')
```

The generator also embeds the original assertion source in the failure-detail string:

```c
v3_eprint_lit("... assert get() == -123456789012345 ...");
```

Therefore, the positive assertion succeeds even when the actual generated condition still compares against a temporary.

The preceding negative check only rejects the exact spelling:

```v
(int)(-123456789012345)
```

A regression emitting this would pass both checks:

```c
int _t2 = -123456789012345;
if (!(_t1 == _t2)) {
    // diagnostic string still contains == -123456789012345
}
```

Extract the generated `if (!(` condition line and assert that the literal occurs inline there. It would also be useful to assert that no `int` temporary is initialized from the wide literal.

## Overall

The capture-side change directly addresses the reported truncation and remains narrowly limited to side-effect-free constant expressions. I would hold approval for the diagnostic regression and the false-positive-prone regression test. This was a static source review; no CI results were considered.

hunterjsb:
Both addressed in ac6e506

**1. Split the predicates.** `is_constant_numeric_literal()` stays as capture eligibility. Display now uses `is_signed_numeric_literal()`, which is non-recursive and accepts only a bare literal or a single `.plus`/`.minus` prefix over one — the cases where the label already spells out the value. `.bit_not`, nested prefixes and parens all fall through to evaluation:

```c
fprintf(..., "  right value", "~5", (long long)(~5));        // = -6
fprintf(..., "  right value", "-(-1)", (long long)(-(-1)));  // = 1
fprintf(..., "  right value", "(5)", (long long)((5)));      // = 5
v3_eprint_lit("  right value: -123456789012345\n");          // label is the value
```

Conditions still compare inline in all four cases, so nothing regains a truncating temp.

**2. The test was false-positive-prone exactly as described.** It now extracts the generated condition rather than scanning the file:

```v
conditions := lines.filter(it.trim_space().starts_with('if (!(') && it.contains('123456789012345'))
assert conditions.len == 1
assert conditions[0].contains('== -123456789012345')
assert !lines.any(it.trim_space().starts_with('int _t') && it.contains('123456789012345'))
```

Against pristine master it now fails on `conditions.len == 1` (`got 0`) rather than on the spelling of the cast, so your hypothetical `int _t2 = -123456789012345;` regression is caught by both the condition check and the temp check.

hunterjsb:
Pushed 2f3d857: re-checking adjacent shapes turned up a constant **infix** operand that was still captured and truncated — `assert minof[i64]() == -9223372036854775807 - 1` (`math_test.v:1145`) emitted `int _t2 = (int)(-9223372036854775807 - 1)`, i.e. compared against 0. The capture exemption now also covers infix over constant operands (predicate renamed to `is_constant_numeric_expr`); the display predicate is untouched, so `2 * 3` etc. still print their evaluated value. This matches the classic backend, which never captures constant operands (`assert_subexpression_to_ctemp` allowlists only side-effecting exprs). The test now covers both shapes and the `v3_assert_*` fixture C stays byte-identical.

hunterjsb:
Superseded by 6186cae0 ("ci: fix current cross-platform failures"), which folds the negated-wide-literal assert fix into the v3 cross-platform CI sweep and adds a `v3_assert_large_negative_literal` regression test. The fix is on master now, so closing this — thanks for picking it up!

PR #28054, opened 2026-08-08
Fix -usecache producing broken binaries in v1

kestred:
This fixes the behavior of `-usecache` in the V1 compiler.

The flag reuses module objects compiled earlier, but the program and those objects disagreed about which one defines a symbol. Values filled in at startup were declared everywhere and defined nowhere, so printing a number crashed, and some generated helpers were defined twice. Modules that pull in C files with `#flag` could not be built at all, and a failed module build neither reported which module failed nor stopped the stale object from being linked. A second commit splits the `-stats` timing into the V part and the C compiler part.

Programs that use C interop, interfaces and globals now build and run with `-usecache`, including V itself, so repeated builds and `v test` can reuse the cached objects instead of recompiling the dependency tree each time.

### DIFF
--- a/vlib/v/builder/builder.v
+++ b/vlib/v/builder/builder.v
@@ -1,6 +1,7 @@
 module builder
 
 import os
+import time
 import v.token
 import v.pref
 import v.errors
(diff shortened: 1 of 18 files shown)


medvednikov:
@codex review

PR #28059, opened 2026-08-09
v3: reduce Cgen RSS after parallel stages

medvednikov:
## What changed

- promote and release the scoped transform arena by default for no-cache self-host C builds
- close the persistent compiler worker pool immediately after its final Cgen work
- retain the launched worker count so utilization metrics remain meaningful after pool teardown
- cover statistics queried after pool closure

The former behavior kept transform scratch, worker root arenas, recycled scoped blocks, and touched worker stacks resident until the entire compiler process exited. None of that state is needed after C generation.

The old transform-arena retention remains available as an explicit diagnostic/performance experiment with `V3_RETAIN_TRANSFORM_SCOPE=1`.

## Impact

For the no-cache V3 self-host benchmark on macOS arm64:

```text
before: cgen (parallel)  224.32 ms  1350 MB RSS
 after: cgen (parallel)  247.41 ms   869 MB RSS
```

This reduces reported Cgen RSS by 481 MB (35.6%). The complete benchmark remained approximately 1.10 seconds, and two independent C-only generations with a fixed `SOURCE_DATE_EPOCH` were byte-identical.

## Validation

- `./vnew -silent test vlib/v3/workers/`
- `./vnew -silent test vlib/v3/driver/`
- `../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -nocache -building-v -o v4 v3.v`
- deterministic two-pass C output comparison with `SOURCE_DATE_EPOCH=1700000000`

A full V3 suite run was intentionally skipped. Rebuilding the debug `vnew` compiler was attempted but encountered an unrelated existing V1 C-generation failure in `vlib/v/ast/cflags.v`; the existing working `vnew` was used for the focused tests.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -7640,7 +7640,7 @@ pub fn run(args []string) {
 			prealloc_scope_leave_for_v3(transform_scope)
 			retain_transform_scope := building_v && backend == 'c' && !cache_state.manager.enabled
 				&& retained_transform_regions.len == 0
-				&& os.getenv('V3_NO_RETAIN_TRANSFORM_SCOPE') == ''
+				&& os.getenv('V3_RETAIN_TRANSFORM_SCOPE') == '1'
 			if retain_transform_scope {
 				// Cgen is the only remaining semantic consumer in this no-cache self-host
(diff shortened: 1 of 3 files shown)


medvednikov:
@codex review

Issue #28063, opened 2026-08-10
net.http HTTP/2 server: no API to send response trailers


hunterjsb:
V's HTTP/2 server (`vlib/net/http/h2_server.v`) has no way to **send** response trailers.

The only public server surface is `pub interface Handler { handle(Request) Response }` (`server.v:21`) and `pub struct Response` (`response.v:14`), which has no `trailers` field. `send_response` (`h2_server.v:924`) closes the stream by setting `END_STREAM` on the final HEADERS or DATA frame, so no trailing HEADERS block can follow the body.

The server already *receives* and validates request trailers (the §5.1/§8.1 trailing-HEADERS handling in `h2_server.v`), so the framing + HPACK machinery is in place — only a send path and a public field are missing.

**Why it matters:** a successful gRPC response must carry `grpc-status`/`grpc-message` in HTTP/2 trailers (RFC 9113 §8.1), so a conformant gRPC *server* can't be built on the current API — this is the one remaining gap between V and native gRPC (cf. #5017). Motivating use case: [we-be/grpc.v](https://github.com/we-be/grpc.v), whose client already interops with grpc-go but whose server has to fall back to Connect for exactly this reason.

**Proposed:** add a `trailers Header` field to `Response`, and have `send_response` emit a trailing HEADERS block with `END_STREAM` after the body when `trailers` is non-empty. Happy to send a PR if the shape sounds right.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

hunterjsb:
closing — @quaesitor-scientiam (V's h2 author) is already on this; design + discussion consolidated in we-be/grpc.v#4. thanks!

PR #28066, opened 2026-08-10
net.http: add server-side HTTP/2 response trailers

quaesitor-scientiam:
## Summary

This PR ended up shipping two related changes together (branching mistake on my part — the second was meant to be #28067, a separate PR — see note at the bottom):

**1. Server-side HTTP/2 response trailers** (`response.v`, `h2_server.v`, `h2_server_test.v`)
- `http.Response` gains a `trailers Header` field; the HTTP/2 server emits it as a trailing HEADERS block after the body, or folds it into the single response HEADERS when the body is empty (RFC 9113 §8.1's Trailers-Only shape — also the common gRPC "fails immediately" case).
- Outbound trailer fields go through the same RFC 9113 §8.2.2 hop-by-hop / pseudo-header / forbidden-octet filtering already applied to inbound fields elsewhere in this file.
- `HTTP/1.1` write path leaves `trailers` untouched (documented as HTTP/2-only; H1 chunked-trailer emission is a separate, unimplemented feature).
- Closes the last gap for a conformant gRPC server on V's `net.http` h2 stack: a successful gRPC response must carry `grpc-status`/`grpc-message` in HTTP/2 trailers, and the server previously had no send path for them — it could only receive/validate them.

**2. Prior-knowledge cleartext HTTP/2 (h2c) on the plain listener** (`server.v`, `server_h2c.v`, `server_h2c_test.v`)
- `Server.enable_http2` now also opts the plain (non-TLS) listener into h2c per RFC 9113 §3.4: a connection whose first bytes are the HTTP/2 client preface is served by the existing h2 driver; everything else is parsed as HTTP/1.1 unchanged.
- Detection uses `io.BufferedReader.peek()`, which never consumes bytes on a miss.
- Only prior-knowledge h2c is implemented — RFC 9113 dropped the old HTTP/1.1 `Upgrade: h2c` mechanism entirely, and this matches what real gRPC clients use for cleartext.

Related to vlang/v#28063 and we-be/grpc.v#4.

## Test plan
- [x] `./vnew -silent test vlib/net/http/h2_server_test.v` — trailers tests: trailers-after-body, the Trailers-Only single-frame optimization, hop-by-hop field filtering, and a regression test for a stream reset mid-body
- [x] `./vnew -silent test vlib/net/http/server_h2c_test.v` — h2c tests: a real TCP connection driven with raw h2 frames against a plain `enable_http2: true` listener, and a regression check that ordinary HTTP/1.1 traffic on the same listener is untouched
- [x] `./vnew -silent test vlib/net/http/` — full package suite green (27 passed, 2 pre-existing skips, 0 failures)

## Note on #28067

#28067 was meant to carry the h2c change as its own PR, and was opened, reviewed, and fixed independently. It ended up redundant with what's here purely due to a branching mistake on my end (the h2c commit never got removed from this branch's history after being copied over to #28067's branch) — not a duplicate submission. #28067 does contain two small review-prompted fixes not present in this PR's version of the h2c code (debug-only error logging to match the TLS/ALPN listener's existing pattern, and a doc-comment clarification); those are minor and can be picked up separately if wanted.

### DIFF
--- a/vlib/net/http/h2_server.v
+++ b/vlib/net/http/h2_server.v
@@ -936,11 +936,62 @@ fn (mut c H2ServerConn) send_response(stream_id u32, resp Response, mut handler 
 	}
 	body := resp.body.bytes()
 	has_body := body.len > 0
+	trailer_fields := c.h2_outbound_trailer_fields(resp.trailers)
+	has_trailers := trailer_fields.len > 0
+
+	if !has_body && has_trailers {
+		// Trailers-Only: no body separates the response from the trailers, so
+		// both go in one HEADERS block (RFC 9113 §8.1 permits a single HEADERS
+		// frame to carry response and trailer fields together when there is no
+		// DATA in between). This is also the common gRPC "fails immediately"
+		// shape and saves a frame over always splitting.
+		fields << trailer_fields
+		block := c.encoder.encode(fields)
+		c.send_header_block(stream_id, block, true)!
+		return
+	}
+
 	block := c.encoder.encode(fields)
 	c.send_header_block(stream_id, block, !has_body)!
 	if has_body {
-		c.send_body(stream_id, body, mut handler)!
+		c.send_body(stream_id, body, !has_trailers, mut handler)!
 	}
+	// send_body may return early (stream_id no longer in c.streams) if the peer
+	// RST_STREAM'd this stream while we were blocked on flow control. Do not
+	// encode or send trailers in that case: the stream is already torn down on
+	// both ends, and encoding a block that is never transmitted would still
+	// mutate c.encoder's dynamic table, desyncing it from the client's decoder
+	// for every later response on this connection.
+	if has_trailers && stream_id in c.streams {
+		trailer_block := c.encoder.encode(trailer_fields)
+		c.send_header_block(stream_id, trailer_block, true)!
+	}
+}
+
+// h2_outbound_trailer_fields converts handler-authored trailers into wire
+// fields, applying the same RFC 9113 §8.2.2 hop-by-hop filter as the main
+// response headers plus a pseudo-header guard — trailers (like headers) must
+// never carry a pseudo-header field (§8.1) — and the same §8.2.1 forbidden-
+// octet check applied to every inbound field elsewhere in this file
+// (h2_request_field_error, finalize_trailers): a handler-supplied value is
+// wire-bound data too, and Header.add_custom only validates the field NAME,
+// not the value, so nothing upstream of this already rejects an embedded
+// NUL/CR/LF.
+fn (c &H2ServerConn) h2_outbound_trailer_fields(trailers Header) []H2HeaderField {
+	mut fields := []H2HeaderField{}
+	for key in trailers.keys() {
+		lkey := key.to_lower()
+		if lkey.starts_with(':') || lkey in h2_conn_specific_headers {
+			continue
+		}
+		for val in trailers.custom_values(key) {
+			if h2_field_value_has_forbidden_octet(val) {
+				continue
+			}
+			fields << H2HeaderField{lkey, val}
+		}
+	}
+	return fields
 }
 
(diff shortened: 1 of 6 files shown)


quaesitor-scientiam:
All 4 failing checks on #28066 are pre-existing/environmental - PR is ready to merge

JalonSolov:
A local AI review found the following:

Issues:

1. **`send_body` signature change is a breaking change** to an internal function — acceptable since it is not `pub`, but the function is now called with 5 arguments from `send_response` while the old call was 4. The PR correctly adds the `close_stream bool` parameter, but the commit message should call this out explicitly.

2. **RST_STREAM guard is correct but subtly fragile.** After `send_body` returns (it may return early after a reset), the check `stream_id in c.streams` protects the HPACK encoder. The test (`test_h2_server_skips_trailers_after_stream_reset_during_body`) is exactly the right regression test here — good.

3. **Trailers-Only optimization merges trailer fields into `fields` in-place** (`fields << trailer_fields`). Since `fields` is a local `mut []H2HeaderField`, this is safe; but the RFC 9113 §8.1 note in the PR description says this is valid — the `:status` pseudo-header comes first and trailer fields follow in the same HEADERS block. This is permitted but note that **gRPC requires pseudo-headers to precede regular header fields** (HTTP/2 §8.1.2.1). Since `:status` was prepended at the start of fields and trailer fields are appended after, the ordering is compliant.

4. **`has_body = true` but `has_trailers = false` path:** when the response has a body but no trailers, `close_stream = !has_trailers = true`, so `end_stream` is set on the last DATA frame — correct, same as before.

5. **No `h2_field_value_has_forbidden_octet` check on the field name** (only on the value). This mirrors how the same issue exists for response headers in the existing code, so it is consistent — but it is worth noting as a known gap.

Two minor issues:

1. **`test_h2_server_skips_trailers_after_stream_reset_during_body` uses `size: 100` with a window of 10**, which causes the server to block after 10 bytes. The test works today, but if `max_frame_size` or default window mechanics change, `100 / 10 = 10 chunks` may or may not be enough to reliably trigger the RST_STREAM before the body completes. A comment explaining the intended constraint (body >> initial window, so RST arrives before send_body unblocks) would help.

2. **`or {}` on `trailers.add_custom(...)` silently swallows errors** — this is fine in test handlers where the keys are statically known valid strings, but is a V convention worth flagging for new contributors.

quaesitor-scientiam:
Thanks for the review — went through each point:

**#2, #3, #4** — confirmed correct as described, no changes needed.

**#1** (`send_body` now takes 4 params, commit message doesn't call it out) — fair, but it's a private function with one call site and the diff is self-explanatory; not worth a rebase of an already-CI'd commit to fix in the message alone.

**#5** (no forbidden-octet check on trailer *names*, only values) — checked this against `Header`'s actual encapsulation: `data` is `pub mut`, but `cur_pos` (the field every read method — `keys()`, `custom_values()` — bounds its iteration by) is private, and `add_custom()` is the only way to advance it. `add_custom` already calls `is_valid(key)!`, which enforces strict RFC 7230 token characters — NUL/CR/LF included. A malformed name genuinely can't reach `h2_outbound_trailer_fields` through the public API, so there's no live gap here. The *value* check was necessary specifically because `add_custom` validates only the key, not the value — that asymmetry was real and is what's fixed. This mirrors the pre-existing `resp.header` path exactly, nothing new introduced.

**Minor #2** (`or {}` swallowing errors) — agreed it's fine, just a V-idiom note.

**Minor #1** (test's `size:100`/window `10` could use a comment) — good suggestion, added in `af479caa00`. One correction on the mechanics: the server blocks after exactly *one* 10-byte DATA frame, not 10 chunks — the window is fully exhausted by the first chunk since 10 < max_frame_size, so `pump_for_window` reads the already-buffered `RST_STREAM` on the very next iteration regardless of how many chunks the remaining body would eventually take. The comment now spells this out.

PR #28067, opened 2026-08-11
net.http: add prior-knowledge cleartext HTTP/2 (h2c) to the plain listener

quaesitor-scientiam:
## Summary
- `Server.enable_http2` now also opts the **plain (non-TLS) listener** into prior-knowledge cleartext HTTP/2 (h2c, RFC 9113 §3.4): a connection whose first bytes are the HTTP/2 client preface is handed off to the existing h2 server driver; everything else is parsed as HTTP/1.1 exactly as before.
- Detection uses `io.BufferedReader.peek()`, which never consumes bytes on a miss — the HTTP/1.1 path is unaffected byte-for-byte when the peek doesn't match.
- Only **prior-knowledge** h2c is implemented. RFC 9113 dropped the old HTTP/1.1 `Upgrade: h2c` mechanism (RFC 7540 §3.2) entirely — it's not supported here, matching the spec and matching what real gRPC clients use for cleartext (e.g. `grpc.Dial` with insecure credentials).
- New `H2cTransport` adapts the buffered connection reader to the existing `H2Transport` interface, so this plugs into the unmodified h2 driver with zero changes to `h2_server.v`.

This is a standalone follow-up to #28066 (server-side response trailers) — same motivation, independent change. Together they're the two items originally scoped in #28063.

Related to vlang/v#28063 and we-be/grpc.v#4 (native gRPC server tracking issue).

## Test plan
- [x] `./vnew -silent test vlib/net/http/server_h2c_test.v` — new end-to-end tests: a real TCP connection driven with raw h2 frames against a plain `enable_http2: true` listener, and a regression check that ordinary HTTP/1.1 traffic on the same listener is untouched
- [x] `./vnew -silent test vlib/net/http/` — full package suite green (27 passed, 2 pre-existing skips, 0 failures), run in complete isolation from #28066 (this branch has exactly one commit ahead of `upstream/master`)
- [x] `./vnew fmt -w` on all touched/new files

### DIFF
--- a/vlib/net/http/server.v
+++ b/vlib/net/http/server.v
@@ -52,7 +52,7 @@ pub mut:
 	cert                   string
 	cert_key               string
 	in_memory_verification bool
-	enable_http2           bool // opt in to HTTP/2 on the TLS listener: advertises ALPN `h2, http/1.1`. Clients that select `h2` are served by the HTTP/2 driver; clients that select `http/1.1` (or send no ALPN) keep the existing HTTP/1.1 path.
+	enable_http2           bool // opt in to HTTP/2. On the TLS listener, advertises ALPN `h2, http/1.1`; clients that select `h2` are served by the HTTP/2 driver, others keep the existing HTTP/1.1 path. On the plain listener, opts into prior-knowledge cleartext h2c (RFC 9113 §3.4): a connection whose first bytes are the HTTP/2 client preface is served by the HTTP/2 driver, others are parsed as HTTP/1.1 as before. A `Server` only ever runs one listener (see listen_and_serve), so one flag unambiguously covers both.
 
 	on_running fn (mut s Server) = unsafe { nil } // Blocking cb. If set, ran by the web server on transitions to its .running state.
 	on_stopped fn (mut s Server) = unsafe { nil } // Blocking cb. If set, ran by the web server on transitions to its .stopped state.
@@ -99,7 +99,7 @@ pub fn (mut s Server) listen_and_serve() {
 	// Create workers
 	mut ws := []thread{cap: s.worker_num}
 	for wid in 0 .. s.worker_num {
-		ws << new_handler_worker(wid, ch, s.handler, s.max_keep_alive_requests)
+		ws << new_handler_worker(wid, ch, s.handler, s.max_keep_alive_requests, s.enable_http2)
 	}
 
 	if s.show_startup_message {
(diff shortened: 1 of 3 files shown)


quaesitor-scientiam:
Same 4 checks, same failure signatures as #28066. This is the same underlying master-wide breakage.  PR is ready to merge.

JalonSolov:
Local AI review says:

* **Minor (should fix):** `serve_h2_conn ... or {}` — add `$if debug { eprintln(...) }` to match the TLS path's error handling.
* **Nit:** Add a short inline comment to `h2c_preface_prefix` noting the literal value being matched (`'PRI '`).

quaesitor-scientiam:
Both addressed in `4421dd2160`:

- `serve_h2_conn(...) or {}` now logs via `$if debug { eprintln('h2c server error: ${err}') }`, matching the TLS/ALPN path's exact pattern in `server_tls_notd_use_openssl.v`.
- `h2c_preface_prefix`'s doc comment now spells out the literal it matches (`"PRI "`).

Full `net/http` suite still green (27 passed, 2 pre-existing skips) after both changes.

JalonSolov:
Conflict likely due to changes in #28066

quaesitor-scientiam:
Closing this — the h2c change here is already in master. My mistake, explaining fully:

This PR was meant to carry the h2c listener as its own standalone change, split off from #28066 (trailers) specifically so the two could be reviewed and merged independently. I did split it off correctly onto this branch. But afterward, when I went back to address review feedback on #28066, I checked out that PR's branch to add a fix commit — and the original h2c commit was still sitting on it locally (I'd copied it over here via cherry-pick, but never removed it from the source branch). Pushing that fix commit silently carried the leftover h2c commit along with it, and #28066 ended up merging both changes together without me noticing until after the fact.

Net result: `vlib/net/http/server_h2c.v` and the h2c changes to `server.v` are already in `upstream/master`, delivered via #28066 instead of this PR. I've corrected #28066's description to reflect that. This PR is now a pure duplicate of already-merged code, which is why GitHub is showing it as conflicting.

One thing genuinely not yet in master: two small fixes from @JalonSolov's review here (`$if debug` error logging to match the TLS/ALPN listener, and a doc-comment clarification on `h2c_preface_prefix`) — both in `4421dd2160`. Happy to bring those over in a fresh small PR if wanted; they're non-functional polish either way.

Apologies for the noise on this one.

PR #28073, opened 2026-08-12
compiler: fix V3 captured callbacks and V1 #insert fallback

medvednikov:
On macOS, this failure was a two-stage compiler cascade:

1. V3 is the default, but its prefix parameter scan treated a leading closure capture as the end of a function literal parameter region. A captured callback such as `fn [state] (int, string)` was therefore typed as `fn ()`, causing V3 to request the compatibility fallback.
2. V1 then mishandled `#insert` in a module that links C libraries. It emitted synthesized `extern` declarations after inserted static definitions, which exposed the reported linkage/qualifier conflicts.

This change fixes both paths:

- V3 now skips explicit capture identifiers before scanning the function literal parameter prefix, with coverage for a captured two-parameter callback stored in a variable.
- V1 now classifies `#insert` like `#include` and `#preinclude` when deciding whether inserted C already supplies declarations, for both same-file and module-wide extern fallback decisions.

V3 already handles the inserted Office VBA header correctly; compiling that focused module directly does not emit conflicting externs.

Focused checks:
- `./vnew -old-compiler test vlib/v3/tests/review_transform_regressions_test.v -run-only test_capturing_callback_variable_keeps_declared_parameters`
- `VFLAGS=-old-compiler ./vnew -old-compiler -silent vlib/v/compiler_errors_test.v`
- default macOS V3 Office build completed without a V3-to-V1 retry: `/Users/alex/code/v/vnew -no-memory-limit -o /tmp/office_excel_v3_default cmd/excel`

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -469,7 +469,7 @@ pub fn gen(files []&ast.File, mut table ast.Table, pref_ &pref.Preferences) GenO
 				if stmt.kind == 'flag' && stmt.main.contains('-l') {
 					global_g.mods_with_c_libs[file.mod.name] = true
 				}
-				if stmt.kind in ['include', 'preinclude'] {
+				if stmt.kind in ['include', 'preinclude', 'insert'] {
 					global_g.mods_with_c_includes[file.mod.name] = true
 				}
 			}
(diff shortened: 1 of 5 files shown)


medvednikov:
@codex review

PR #28081, opened 2026-08-13
db.pg: expose PostgreSQL result field metadata

guweigang:
## Summary

- expose libpq column metadata through `pg.Result.fields`
- preserve type OID, type modifier, size, format, and source table information before `PQclear()`
- document the new result metadata API
- cover direct, parameterized, prepared, empty-result, mixed-type, and table-origin queries

## Why

`pg.Result` previously retained only column names and rows. Column type information from libpq was
discarded when the underlying `PGresult` was cleared, forcing callers to treat every result column as
text.

## Validation

- `./vnew fmt -w vlib/db/pg/pg_result_test.v`
- `git diff --check`
- `./vnew -d network -silent vlib/db/pg/pg_result_test.v`
- `./vnew -silent test vlib/db/pg/` (2 passed, 4 network-gated files skipped)
- `./vnew -d network -silent test vlib/db/pg/` (5 passed; existing
  `test_listen_notify` failed because the received channel was empty)

### DIFF
--- a/vlib/db/pg/README.md
+++ b/vlib/db/pg/README.md
@@ -188,6 +188,28 @@ defer { conn.close() or {} }
 rows := conn.exec('select 1')!
 ```
 
+## Result Column Metadata
+
+Queries made with `exec_result()`, `exec_param_many_result()`, or
+`exec_prepared_result()` return a `pg.Result` whose `fields` array contains the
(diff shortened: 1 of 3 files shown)


medvednikov:
@codex review

PR #28083, opened 2026-08-13
HTTP/3 (QUIC) foundation: Phase 9 (QuicConn top-level struct and event loop)

quaesitor-scientiam:
## Summary

Phase 9 of HTTP/3/QUIC support (#27675), continuing from the Phase 8 work in
#27885 (connection lifecycle) — idle timeout, connection close, stateless
reset, and ECN bookkeeping are all wired directly into the connection driver
this phase adds.

**Update:** #27885 has merged. This branch has been rebased cleanly onto
`master` (`git rebase --onto upstream/master <old-base> HEAD`, no
conflicts) and no longer carries any of its commits.

9a and 9b are both implemented, formatted, passing the full `net.quic` suite
(37/37), and have been through a full `/vreview` pass (see Test plan below).

See [`vlib/net/quic/PROGRESS.md`](https://github.com/quaesitor-scientiam/v/blob/http3-quic-quicconn/vlib/net/quic/PROGRESS.md)

## Scope

Unlike every prior phase, Phase 9 had no detailed design written down
before starting — every earlier phase's `PROGRESS.md` entry cites specific
RFC sections and prior art; Phase 9's was one line. `QuicConn`
(`vlib/net/quic/conn.v`, new file) is the top-level struct that wires
together everything Phases 0-8 built as independent, unit-tested pieces —
no existing file composes any two of them yet, by design (each phase's own
doc comments defer that wiring to "a future `QuicConn`").

Design recovered from the tracking issue's own wording (struct name, file
name, the `poll()`/`process_timeouts()` event-loop contract, the `role`
field, "single-threaded caller-driven... mirroring how quiche/ngtcp2
expose themselves as libraries") plus ~15 scattered "deferred to Phase 9"
doc comments across the existing 31 files that name exactly which piece of
connection-level state (`scids`, current/original DCID tracking, "at most
one Retry accepted," MAX_STREAMS enforcement, transport-parameter default
application, etc.) belongs here.

Sub-phased like Phase 2, landing as one PR:

- **9a — Connection establishment**: `dial()`, Retry/VN handling,
  Initial/Handshake packet processing, driving the TLS handshake state
  machine to `handshake_confirmed`, key derivation/discard, CRYPTO
  reassembly, ACK generation tied into loss detection, idle timeout.
- **9b — Steady state**: 1-RTT packet processing (STREAM/RESET_STREAM/
  STOP_SENDING/MAX_DATA/MAX_STREAM_DATA/MAX_STREAMS dispatch),
  `open_stream`/`write_stream`/`read_stream`/`close` application API,
  connection- and stream-level flow control, stateless-reset detection,
  ECN recording, graceful/immediate close with RFC 9000 §10.2
  closing/draining retransmission + timeout.

Connection ID rotation/migration stays explicitly out of scope for v1 (not
deferred — several existing files independently document this), matching
0-RTT (a separate, later phase). Client-initiated 1-RTT key update
rotation is also out of v1 scope — an existing decision from Phase 8's own
`key_update.v` (only responding to a peer-initiated update is implemented),
not something Phase 9 needed to add.

`/vreview` found and fixed two bugs before the final commit, both
reproduced with a failing test first:

- `write_stream`/`read_stream` never called `ensure_stream_windows` for a
  stream that only exists because RFC 9000 §2.1 auto-created it as a
  lower-numbered sibling of a peer-referenced higher stream ID — such a
  stream had no flow-control window, so a local write to it queued forever
  with no error, never reaching the wire.
- `handle_peer_connection_close` and the stateless-reset branch of
  `note_one_rtt_processing_failed` transitioned to `.draining` without
  setting `closing_deadline` (RFC 9000 §10.2.2 requires the draining
  period to be bounded, same as closing) — a peer-initiated close, the
  most common real-world close path, left the connection draining forever
  instead of eventually reaching `.closed`. Fixed with a shared
  `enter_draining(now)` helper mirroring `close_with_error`'s existing
  deadline formula.

## Test plan

- [x] 9a: hand-built ServerHello→Finished fixtures (reusing Phase 2's RFC
      8448/quiche vectors — no live server) driving `dial()` + `poll()` to
      `handshake_confirmed`; key transitions/discards verified at each
      level.
- [x] 9b: STREAM data round-trip through `poll()`/`open_stream`/
      `write_stream`/`read_stream`; MAX_STREAMS blocking + STREAMS_BLOCKED
      + unblock; application-initiated `close()`; peer-initiated
      CONNECTION_CLOSE → draining (including the RFC 9000 §10.2.2 bounded-
      draining regression test).
- [x] All new/modified code passing under `./vnew test vlib/net/quic/`
      (37/37).
- [x] Full `/vreview` A-G pass before commit — found and fixed the two
      bugs described above.

### DIFF
--- a/vlib/net/quic/PROGRESS.md
+++ b/vlib/net/quic/PROGRESS.md
@@ -864,14 +864,14 @@ The largest, highest-risk phase. Sub-phases, in build order:
       one-sided and both-sided paths and confirms a large-but-finite,
       strictly positive `Duration` comes back.
 
-## Phase 9 — QuicConn top-level struct and event loop (IN PROGRESS)
+## Phase 9 — QuicConn top-level struct and event loop
 
 No prior phase composes any two of the pieces built so far — every phase
 built independently unit-tested state and deferred wiring to "a future
 QuicConn." `vlib/net/quic/conn.v` is that wiring. Sub-phased like Phase 2,
 landing as one PR:
 
-- [ ] **9a — Connection establishment** (`conn.v`): `dial()` picks
+- [x] **9a — Connection establishment** (`conn.v`): `dial()` picks
       `scid`/`original_dcid`, derives Initial secrets, builds ClientHello.
       `poll()`/`process_timeouts()` handle Retry/VN detection (RFC 9000
       §17.2.5.2's at-most-one-Retry + VN/Retry anti-spoof state, both
@@ -882,7 +882,7 @@ landing as one PR:
       `CryptoStreamReassembler`, ACK generation/processing for
       Initial/Handshake tied into `loss_detection.v`, idle timeout, and key
       discard on handshake confirmation (RFC 9001 §4.9).
-- [ ] **9b — Steady state**: 1-RTT packet processing (STREAM/ACK/
+- [x] **9b — Steady state**: 1-RTT packet processing (STREAM/ACK/
       CONNECTION_CLOSE dispatch into `QuicStreamSet` + flow control both
       directions), `open_stream`/`write_stream`/`read_stream`, full
       loss-detection↔congestion-control wiring for 1-RTT sends,
@@ -890,10 +890,24 @@ landing as one PR:
       current limit, stateless-reset detection on decrypt failure, ECN
       count recording, graceful/immediate close, 1-RTT key update rotation
       (`app_read_keys`/`app_write_keys`).
-- [ ] Tests: `conn_test.v`, hand-built ServerHello→Finished fixtures
+- [x] Tests: `conn_test.v`, hand-built ServerHello→Finished fixtures
       (reusing Phase 2's RFC 8448/quiche vectors, no live server) driving
       `dial()`+`poll()` to `handshake_confirmed`; STREAM data round-trip
       through `poll()`; CONNECTION_CLOSE handling; key update.
+- [x] `/vreview` found and fixed two real bugs, both reproduced with a
+      failing test before the fix landed (Phase R): (1) `write_stream`/
+      `read_stream` never called `ensure_stream_windows` for a stream that
+      only exists because RFC 9000 §2.1 auto-created it as a lower-numbered
+      sibling of a peer-referenced higher stream ID — such a stream had no
+      flow-control window, so a local write to it queued forever with no
+      error, never reaching the wire; (2) `handle_peer_connection_close`
+      and the stateless-reset branch of `note_one_rtt_processing_failed`
+      transitioned to `.draining` without setting `closing_deadline` (RFC
+      9000 §10.2.2 requires the draining period to be bounded, same as
+      closing) — a peer-initiated close, the most common real-world close
+      path, left the connection draining forever instead of eventually
+      reaching `.closed`. Fixed with a shared `enter_draining(now)` helper
+      mirroring `close_with_error`'s existing deadline formula.
 
 Connection ID rotation/migration is explicitly OUT of scope (not deferred
 to a later phase — `stateless_reset.v`/`pmtu.v` both say so independently):
--- a/vlib/net/quic/PROGRESS.md
+++ b/vlib/net/quic/PROGRESS.md
@@ -864,12 +864,46 @@ The largest, highest-risk phase. Sub-phases, in build order:
       one-sided and both-sided paths and confirms a large-but-finite,
       strictly positive `Duration` comes back.
 
-## Phases 9-14 (NOT STARTED)
+## Phase 9 — QuicConn top-level struct and event loop (IN PROGRESS)
+
+No prior phase composes any two of the pieces built so far — every phase
+built independently unit-tested state and deferred wiring to "a future
+QuicConn." `vlib/net/quic/conn.v` is that wiring. Sub-phased like Phase 2,
+landing as one PR:
+
+- [ ] **9a — Connection establishment** (`conn.v`): `dial()` picks
+      `scid`/`original_dcid`, derives Initial secrets, builds ClientHello.
+      `poll()`/`process_timeouts()` handle Retry/VN detection (RFC 9000
+      §17.2.5.2's at-most-one-Retry + VN/Retry anti-spoof state, both
+      explicitly documented in `retry.v`/`version_negotiation.v` as this
+      phase's job), Initial/Handshake packet demux→unprotect→frame-parse,
+      driving `Tls13ClientHandshake` through to `process_finished`, key
+      derivation/promotion per level, CRYPTO-frame reassembly via
+      `CryptoStreamReassembler`, ACK generation/processing for
+      Initial/Handshake tied into `loss_detection.v`, idle timeout, and key
+      discard on handshake confirmation (RFC 9001 §4.9).
+- [ ] **9b — Steady state**: 1-RTT packet processing (STREAM/ACK/
+      CONNECTION_CLOSE dispatch into `QuicStreamSet` + flow control both
+      directions), `open_stream`/`write_stream`/`read_stream`, full
+      loss-detection↔congestion-control wiring for 1-RTT sends,
+      MAX_STREAMS enforcement on locally-opened streams against the peer's
+      current limit, stateless-reset detection on decrypt failure, ECN
+      count recording, graceful/immediate close, 1-RTT key update rotation
+      (`app_read_keys`/`app_write_keys`).
+- [ ] Tests: `conn_test.v`, hand-built ServerHello→Finished fixtures
+      (reusing Phase 2's RFC 8448/quiche vectors, no live server) driving
+      `dial()`+`poll()` to `handshake_confirmed`; STREAM data round-trip
+      through `poll()`; CONNECTION_CLOSE handling; key update.
+
+Connection ID rotation/migration is explicitly OUT of scope (not deferred
+to a later phase — `stateless_reset.v`/`pmtu.v` both say so independently):
+`QuicConn` holds exactly one local `scid` and tracks the peer's current
+`dcid`, no active CID set, no NEW_CONNECTION_ID/RETIRE_CONNECTION_ID.
+
+## Phases 10-14 (NOT STARTED)
 
 See the tracking issue for full detail on each. In order:
 
-9. `QuicConn` — top-level struct, `poll()`/`process_timeouts()` event-loop
-   contract.
 10. HTTP/3 framing (RFC 9114) — incremental/resumable parsing (structurally
     different from HTTP/2's single-shot framing).
 11. QPACK (RFC 9204) — absolute indexing, encoder/decoder streams, blocked-
--- a/vlib/net/quic/conn.v
+++ b/vlib/net/quic/conn.v
@@ -1,0 +1,1486 @@
+module quic
+
+import time
+import rand
+
+// RFC 9000/9001 — QuicConn is the top-level connection struct wiring
+// together every independently-built piece from Phases 0-8: packet/header
+// protection, the TLS 1.3 handshake, packet number spaces, CRYPTO stream
+// reassembly, loss detection, NewReno congestion control, and
+// connection-lifecycle bookkeeping (idle timeout, CONNECTION_CLOSE,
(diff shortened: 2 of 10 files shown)


quaesitor-scientiam:
CI Failures none are  related to this PR code

JalonSolov:
Local AI findings:

**Severity: High**
**Issue:** The connection close deadline is enforced in `process_timeouts()`, but it is never scheduled by `compute_next_timeout()`.
**Why this matters:** after a peer-triggered close or stateless reset, the code sets `closing_deadline` and transitions to `.draining`, but the next timeout returned to the caller is still computed from `loss-detection` and `idle-timeout` only. If neither of those is active, `next_timeout` becomes none, so the caller never calls `process_timeouts()` at the required closing/draining deadline. The connection can then remain stuck in `.draining` forever even though the code explicitly says the deadline is required to reach `.closed`.

This is the core mismatch:

`process_timeouts()` checks `c.closing_deadline`
`compute_next_timeout()` does not include `c.closing_deadline`
therefore the close deadline is not actually wired into the caller-driven event loop
That is a real RFC 9000 §10.2 lifecycle bug, not just a cosmetic issue.

quaesitor-scientiam:
Good catch, and confirmed as a real bug. Fixed in commit 94343fcfea.

`compute_next_timeout()` was only merging `loss_detection.next_timeout()` and `idle_timeout_deadline()` -- it never looked at `closing_deadline` at all. So a connection that entered `.closing`/`.draining` with no active PTO timer and no idle timeout configured would return `next_timeout = none`, and a real caller-driven event loop would never know to call `process_timeouts()` again. Exactly the mismatch you described.

Fixed by merging `closing_deadline` into `compute_next_timeout()` with the same min-merge pattern already used for `idle_deadline`.

One note on reproduction: the existing integration test (`test_peer_connection_close_enters_draining`) happened to still pass even with the bug present, because that fixture never constructs ACK frames back from the fake server, so the client's own sent packets are never acked and `loss_detection` always has a stale PTO armed -- which incidentally supplied a non-`none` `next_timeout` and masked the gap. Added a same-module white-box test (`test_compute_next_timeout_includes_closing_deadline`) that force-zeroes `bytes_in_flight` to isolate the exact "nothing else armed" scenario you described, confirmed it fails against the pre-fix code, then confirmed it passes post-fix. Also added a `next_timeout != none` assertion right after entering draining in the existing integration test, so this class of gap can't silently regress there either.

Full `net.quic` suite: 37/37.

quaesitor-scientiam:
CI Failures(same ones) none are related to this PR code

JalonSolov:
Local AI Findings (with a different LLM):

**P1:** Incoming ack-eliciting packets never reset the idle timer. conn.v processes received packets but never calls `note_packet_received`; only outgoing packets update the timer. An otherwise active connection that receives data without sending will therefore close at the original idle deadline. Record whether each successfully decrypted packet is ack-eliciting and call `idle_timeout.note_packet_received(...)`.

**P1:** `HANDSHAKE_DONE` is accepted from a client peer. conn.v marks the handshake as done without checking the endpoint role, although RFC 9000 permits this frame only from a server. Since this PR only constructs clients, receipt should produce `PROTOCOL_VIOLATION` rather than advance handshake state.

quaesitor-scientiam:
> Local AI Findings (with a different LLM):
> 
> **P1:** Incoming ack-eliciting packets never reset the idle timer. conn.v processes received packets but never calls `note_packet_received`; only outgoing packets update the timer. An otherwise active connection that receives data without sending will therefore close at the original idle deadline. Record whether each successfully decrypted packet is ack-eliciting and call `idle_timeout.note_packet_received(...)`.
> 
> **P1:** `HANDSHAKE_DONE` is accepted from a client peer. conn.v marks the handshake as done without checking the endpoint role, although RFC 9000 permits this frame only from a server. Since this PR only constructs clients, receipt should produce `PROTOCOL_VIOLATION` rather than advance handshake state.

What LLM are you using?

JalonSolov:
It varies.  Latest findings were from GPT-5.6 Luna.

quaesitor-scientiam:
Both confirmed. Fixed in commit fda4f83268.

**P1 (idle timer never reset on receive):** confirmed exactly as described -- `conn.v` never called `IdleTimeoutState.note_packet_received` at all, only `note_packet_sent`. Wired it in from `process_initial_or_handshake` and `process_one_rtt_packet`, right where `processed_first_server_packet` is already set on a successful decrypt.

Verifying this turned up a second, deeper bug in the same function: `note_packet_received` itself had RFC 9000 §10.1 backwards -- it gated the RECEIVE-side restart on `is_ack_eliciting`, when the RFC's ack-eliciting condition applies only to the SEND side ("An endpoint restarts its idle timer when a packet from its peer is received and processed successfully" has no ack-eliciting qualifier at all; that only appears in the next sentence, about sending). So even wiring in the call with the old signature would still have left non-ack-eliciting receives (e.g. a lone ACK) failing to reset the timer. Fixed by dropping the now-unnecessary `is_ack_eliciting` parameter -- receive now restarts unconditionally, matching the RFC.

**P1 (HANDSHAKE_DONE role check):** confirmed and fixed, though worth being precise about scope: this codebase only ever constructs clients in v1 (`dial()` hardcodes `role: .client`, no server constructor exists), so the `c.role == .server` branch is currently unreachable through any real call path. Added it anyway rather than deferring to Phase 13 -- it's a one-line, zero-risk guard matching an explicit RFC MUST (§19.20), and `stream.v`'s own `QuicRole` doc comment plus `PROGRESS.md` both state the role field exists specifically so this dispatch code needs no rework once server support lands.

**One more thing, for transparency:** while fixing the idle-timeout gap I initially "corrected" `idle_timeout_deadline()` in a way that was actually wrong -- routing its already-real, nanosecond-scale `time.Duration` through `.milliseconds()` before adding it to `now`, which made the deadline arrive almost immediately instead of after the real configured timeout. `now` is a `time.sys_mono_now()`-sourced nanosecond instant throughout this module (confirmed independently via `loss_detection_test.v`'s own RTT-sample assertions), and the ORIGINAL code was correct all along. Caught this via an empirical test run rather than by inspection alone, and while tracing it found the real, adjacent bug: `closing_or_draining_deadline` (extracted into a shared helper earlier in this PR's history) had the identical mistake in the same direction -- shrinking `pto_period()` to a tiny millisecond count before combining it with `now`, so the closing/draining deadline was landing almost immediately instead of the intended ~3xPTO out. Fixed both, and sibling-searched the whole module for any other `.milliseconds()` call site combining a `Duration` with `now` -- none found.

Full `net.quic` suite: 37/37. Full `/vreview` pass on the diff, no further findings.

quaesitor-scientiam:
3 CI failings are  not related to this PR

JalonSolov:
GPT-5.6 Luna findings:

**High: discarded packet-number spaces are still processed**
In `vlib/net/quic/conn.v:577`, Initial and Handshake packets are dispatched without checking `initial_keys_discarded` or `handshake_keys_discarded`. The discard functions reset bookkeeping but leave the keys usable, so stale packets can still be decrypted and mutate handshake state after RFC-required key discard. Reject packets in discarded spaces before decryption.

**High: destination Connection IDs are never validated**
`vlib/net/quic/conn.v:641` accepts any authenticated server packet and replaces `dcid`/`peer_scid` with its source CID without checking header`.dcid == c.scid`. The short-header path similarly ignores the parsed CID at `vlib/net/quic/conn.v:695`. Initial packets are forgeable by observers because Initial keys are derivable, allowing an attacker to inject a packet and redirect the client’s subsequent packets.

quaesitor-scientiam:
Both confirmed. Fixed in commit baff78bc80.

**Discarded packet-number spaces:** confirmed exactly as described — `discard_initial_keys`/`discard_handshake_keys` only reset per-space bookkeeping (`pn_spaces`, `loss_detection`, `received_pns`), the key material itself is never cleared, so `process_initial_or_handshake` kept decrypting and dispatching post-discard. Fixed with an explicit `c.initial_keys_discarded`/`c.handshake_keys_discarded` check at the top of the function, before any key material is touched.

**Destination CID validation:** also confirmed on both paths. Added `header.dcid != c.scid` (long header) and `short_header.dcid != c.scid` (short header) checks. The short-header path routes a mismatch through the existing `note_one_rtt_processing_failed` rather than a bare drop, since a genuine stateless reset (RFC 9000 §10.3) is deliberately shaped like an ordinary short-header packet with unpredictable bytes where a real DCID would be — a bare drop would have broken stateless-reset detection for the exact packets it's meant to catch.

While fixing these I found three more real gaps in the same area, none previously reported, all Phase-R'd (failing test against the unfixed code, confirmed passing after) and included in the same commit:

- RFC 9000 §7.2 also requires discarding a later packet whose *source* CID differs from the one first established — `process_initial_or_handshake` only ever wrote `peer_scid` on the first packet, never checked a later one against it. Same forgery angle as the destination-CID gap (Initial keys are derivable from the DCID alone).
- RFC 9000 §10.2.3: `close()` always marked its request application-level with the caller's real reason string, with no way to know which packet protection level would end up carrying it. Calling `close()` before 1-RTT keys exist sent the real reason string at Initial/Handshake protection, which the RFC requires downgrading to transport-type with the reason cleared.
- RFC 9000 §10.2.2: `enter_draining` recomputed the closing/draining deadline from the current time on every closing→draining transition instead of preserving the original end time — a peer CONNECTION_CLOSE arriving while this endpoint was already `.closing` extended the period instead of the RFC-required "same end time."

Full `net.quic` suite green (37/37). Also used this round to build out `quic_conformance_matrix.md`'s coverage of `conn.v` against RFC 9000 §5/§7/§10 and RFC 9001 §4.9, which is what actually surfaced the three self-found gaps above — the matrix had a gap for exactly this file that let earlier rounds' misses happen in the first place.

quaesitor-scientiam:
Follow-up after the fixes above: the user directly challenged whether Phase 9 was actually RFC-complete, given how many rounds of external findings this PR has already been through. Rather than answer from confidence, I built out `quic_conformance_matrix.md` sections for the parts of `conn.v` no prior round (including my own) had a standing RFC checklist for: flow control admission, ACK generation timing, loss-detection/congestion-control wiring, and key-update wiring. That audit found 3 more real, previously-unreported gaps, all fixed in commit `a2eb054732`, all with regression tests:

**RFC 9000 §13.2.1** — "An endpoint MUST NOT send a non-ack-eliciting packet in response to a non-ack-eliciting packet." Every processed packet's number was queued for the next auto-ACK regardless of whether it was itself ack-eliciting, so a peer's own ACK-only packet triggered an ACK-only reply back — the exact feedback loop this MUST exists to prevent. Fixed with a frame-type classifier gating the auto-ACK.

**RFC 9002 §5.3** — `handle_ack_frame` used the hardcoded default `ack_delay_exponent` (3) instead of the peer's actually-advertised value (legal up to 20) when scaling an ACK frame's raw ACK Delay field, corrupting RTT/PTO computation for any peer advertising a non-default value. Provable directly from `loss_detection.v`'s own doc comment, which already documents the contract this call was violating.

**RFC 9001 §6.2 (the significant one)** — "The endpoint MUST update its send keys to the corresponding key phase in response" to a peer-initiated key update. The read side was already correct and tested, but this client's own send keys never advanced past generation 0 — every packet sent after a server-initiated rotation, including the ACK for the packet that triggered it, stayed on the old generation. §6.2's own final paragraph gives a compliant server explicit license to treat that as a fatal `KEY_UPDATE_ERROR`. Added write-side generation tracking (`app_write_secret`/`app_write_generation`) and a `sync_write_keys_to_peer_update()` call wired into the same code path that processes an incoming 1-RTT packet, so the send side catches up before anything (including the auto-ACK) goes out.

One further gap was found and left *open* rather than silently fixed: RFC 9002 §6.4's bytes-in-flight credit-back when Initial/Handshake keys are discarded. It's not new — `discard_initial_keys`/`discard_handshake_keys` already document it as a deferred follow-up — so it's flagged in the matrix as still-open rather than re-claimed as resolved.

Full `net.quic` suite green (37/37) throughout.

quaesitor-scientiam:
Analysis of CI failures are not related to this PR

PR #28087, opened 2026-08-14
v3: fix for-in-by-ref copies and @[heap] struct allocation

### LINKED ISSUE #28084: fixed array "for in" compilation error
https://docs.vlang.io/statements-&-expressions.html#array-for
fixed array "for in" compilation error

```vlang
module main

struct Item {
	a [100000]f32
}

@[heap]
struct Test {
	buf [10]Item
}

fn main() {
	t := Test{}
	for i in &t.buf {
		println(i.a[0])
	}
}
```
```
error: initializing 'main__Item *' (aka 'struct main__Item *') with an expression
of incompatible type 'main__Item' (aka 'struct main__Item'); take the address with &
cc:                 main__Item* i = (*_t2)[_t3];
```
> [!NOTE]
>

quaesitor-scientiam:
## Summary

Fixes two related bugs in the `v3` compiler pipeline (the default on macOS for `v run`/`v build`) reported in #28084:

- **`for i in &arr` copied elements by value instead of by reference.** The for-in node's `.amp` op is only set for `for mut i in arr`, unrelated to a `&` on the container expression. The transformer stripped the container's own leading `&` off its type without tracking that fact anywhere downstream, so `lower_indexed_for_in` always bound the element by value regardless of the explicit `&`. Now tracked directly from the container's AST shape (a genuine `.prefix .amp` container expression), independent of `node.op`.

- **`@[heap]`-tagged structs were never actually heap-allocated for plain `t := Test{}` declarations** — only `&Test{}` worked. The type checker already treats `@[heap]` as an unconditional heap-allocation promise for borrow-safety checks (see `checker.v`/`checker_tail.v`), but nothing in transform/codegen enforced it, so the promise was unsound. This hooks the existing escape-analysis heap-promotion machinery (`heap_escaping_source_decl`, already used for locals whose address is later found to escape) so `@[heap]` structs get the same treatment *unconditionally* at declaration, including a real deep copy (not a pointer alias) on `y := t`, matching classic V's value-copy-with-heap-storage semantics exactly.

Fixing the second bug surfaced a real latent codegen bug: `stringify_expr`/`string_interp_needs_value_read` assumed any local whose tracked type is `&T` is a genuine nilable pointer, and produced invalid C (a pointer/value type mismatch) for the new heap-promoted case — `println(t)` for a `@[heap]` struct. Fixed alongside.

## Test plan

- Both fixes verified against the exact repro in #28084, and cross-checked output/generated-C against the classic (`-old-compiler`) backend for parity across: methods (value + pointer receiver), by-value function args, `mut` params, struct auto-str/`println`, return-value capture, nested scopes, per-iteration loop locals, self-referential structs, arrays/maps of heap structs — all byte-identical to classic V.
- Ran the full `vlib/v3/tests/` suite (206 files) before and after: the 30 failures seen post-patch are pre-existing (28 fail identically on the unpatched baseline) or environmental flakiness under 10-way parallel test execution (2 tests flip pass/fail across repeated runs on *both* the baseline and patched binaries) — no reproducible new regressions.
- `./vnew fmt -diff` clean on all touched files.

Fixes #28084

### DIFF
--- a/vlib/v3/transform/fn.v
+++ b/vlib/v3/transform/fn.v
@@ -3920,6 +3920,18 @@ fn (mut t Transformer) stringify_expr(expr_id flat.NodeId) flat.NodeId {
 	if raw_alias_type.len > 0 {
 		typ = raw_alias_type
 	}
+	// `transform_expr` above already auto-dereferenced a `pointer_value_rvalues`
+	// ident (mutable for-in bindings, `@[heap]`-promoted locals, ...) to its value,
+	// but `typ` was resolved independently and still carries the storage `&T`.
+	// Strip it so `wrap_string_conversion` sees a type matching what `expr` (now a
+	// plain value read) actually evaluates to, instead of treating it as a
+	// genuine nilable pointer.
+	if int(expr_id) >= 0 && typ.starts_with('&') {
+		raw_node := t.a.nodes[int(expr_id)]
+		if raw_node.kind == .ident && t.pointer_value_rvalues[raw_node.value] {
+			typ = typ[1..]
+		}
+	}
 	return t.wrap_string_conversion(expr, typ)
 }
 
--- a/vlib/v3/transform/for.v
+++ b/vlib/v3/transform/for.v
@@ -931,6 +931,11 @@ fn (mut t Transformer) lower_indexed_for_in(id flat.NodeId, node flat.Node, key_
 	} else {
 		iter_type
 	}
+	// `for x in &arr` requests by-reference elements via a `&` on the container
+	// expression itself, distinct from `node.op == .amp` (`for mut x in arr`). A
+	// `mut` parameter is stored as a pointer by the C ABI even without an explicit
+	// `&`, so only a genuine address-of/reference expression counts here.
+	container_is_explicit_reference := container_node.kind == .prefix && container_node.op == .amp
 	source_is_owned_temporary := !source_container_type.starts_with('&')
 		&& !t.expr_can_take_address(container_id)
 	direct_map_index_container := node.op == .amp && container_node.kind == .index
@@ -990,7 +995,8 @@ fn (mut t Transformer) lower_indexed_for_in(id flat.NodeId, node flat.Node, key_
 		elem_name = val.value
 	}
 	t.set_var_type(idx_name, 'int')
-	elem_is_mut := node.op == .amp && actual_iter_type != 'string'
+	elem_is_mut := (node.op == .amp || container_is_explicit_reference)
+		&& actual_iter_type != 'string'
 	elem_needs_ref := elem_is_mut
 	elem_var_type := if elem_needs_ref { '&${elem_type}' } else { elem_type }
 	t.set_var_type(elem_name, elem_var_type)
--- a/vlib/v3/transform/transform.v
+++ b/vlib/v3/transform/transform.v
@@ -5279,6 +5279,14 @@ fn (t &Transformer) string_interp_needs_value_read(name string, typ string) bool
 	if t.mut_param_values[name] {
 		return true
 	}
+	// A name whose own tracked type is already `&T` is genuinely pointer-backed
+	// storage (mutable for-in bindings, `@[heap]`-promoted locals, ...). Those
+	// still read as a plain value everywhere else via `pointer_value_rvalues`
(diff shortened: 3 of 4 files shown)


quaesitor-scientiam:
The \`v3\` CI check is failing, but it's unrelated to this change: the failure is in \`vlib/v3/tests/aoc_compat_codegen_test.v\`, which doesn't touch for-in iteration or \`@[heap]\` structs at all. It's a closure-argument-count bug for captured-variable closures passed to \`.any()\`/\`.all()\`/\`.count()\`:

\`\`\`
error: function needs exactly 1 argument
    println(values.any(fn [limit] (value int) bool {
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        return value > limit
    }))
\`\`\`

Confirmed pre-existing by running that exact test with \`-old-compiler\` (routes entirely through the classic \`vlib/v/\` compiler — a code path this PR doesn't touch at all, since every change here is under \`vlib/v3/\`): it fails identically there too, so it can't be caused by this diff.

No existing tracking issue for it that I could find — happy to file one separately if useful.

JalonSolov:
Since this is v3, @medvednikov should decide on merging.

Local AI check said:  No concrete correctness issues found

medvednikov:
This would complicate the language. Just use for with an index.

quaesitor-scientiam:
> This would complicate the language. Just use for with an index.

Can't compile my code in V3 then

medvednikov:
my bad, it's already in the language

> Array elements are taken by value by default; apply & to the array to take them by reference.

medvednikov:
Why #28087 itself should not be merged as-is

The PR combines three concerns:

Reference iteration over fixed arrays.
Actual heap promotion for plain @[heap] struct literals.
Stringification behavior for compiler-promoted pointer-backed values.

Although the diff is only 45 additions across four compiler files, those are distinct semantic changes. There are also no committed regression-test files in the patch; the extensive test work is described in the PR body but is not preserved as tests that future compiler changes would run.

I would submit two separate changes:

1. Narrow reference-iteration fix

Only fix:

for item in &fixed_array {
}

The compiler’s typed representation should carry an explicit element-binding mode—conceptually value, ref, or mut_ref—rather than discovering reference semantics late in lowering by inspecting whether the original container AST happened to be a prefix-& node. That would keep parentheses, aliases, pointer variables, fixed arrays, and dynamic arrays consistent.

Tests should cover at least:

A local fixed array.
A fixed array stored in a struct field.
An ordinary dynamic array, to prevent regression.
Read-only reference iteration versus for mut item in array.
The exact reproduction from #28084.
2. Separate @[heap] fix

Heap promotion changes storage representation, assignment semantics, escaping references, copying, method calls, and string conversion. That deserves a dedicated PR and dedicated tests rather than being coupled to a for lowering correction.

JalonSolov:
Although it would still be cleaner to use for with an index.  :-\

quaesitor-scientiam:
> Why #28087 itself should not be merged as-is
> 
> The PR combines three concerns:
> 
> Reference iteration over fixed arrays. Actual heap promotion for plain @[heap] struct literals. Stringification behavior for compiler-promoted pointer-backed values.
> 
> Although the diff is only 45 additions across four compiler files, those are distinct semantic changes. There are also no committed regression-test files in the patch; the extensive test work is described in the PR body but is not preserved as tests that future compiler changes would run.
> 
> I would submit two separate changes:
> 
> 1. Narrow reference-iteration fix
> 
> Only fix:
> 
> for item in &fixed_array { }
> 
> The compiler’s typed representation should carry an explicit element-binding mode—conceptually value, ref, or mut_ref—rather than discovering reference semantics late in lowering by inspecting whether the original container AST happened to be a prefix-& node. That would keep parentheses, aliases, pointer variables, fixed arrays, and dynamic arrays consistent.
> 
> Tests should cover at least:
> 
> A local fixed array. A fixed array stored in a struct field. An ordinary dynamic array, to prevent regression. Read-only reference iteration versus for mut item in array. The exact reproduction from #28084. 2. Separate @[heap] fix
> 
> Heap promotion changes storage representation, assignment semantics, escaping references, copying, method calls, and string conversion. That deserves a dedicated PR and dedicated tests rather than being coupled to a for lowering correction.

I will split it

PR #28094, opened 2026-08-15
HTTP/3 (QUIC) foundation: Phase 10 (HTTP/3 framing, RFC 9114)

quaesitor-scientiam:
## Summary

Phase 10 of the HTTP/3/QUIC implementation (tracking issue #27675): the HTTP/3
framing layer (RFC 9114 §6.2, §7, §8.1, §9, §10.8). All prior phases (0-9) are
merged to master; this branches directly off master, no stacking needed.

Unlike QUIC's own frames (parsed once from an already-fully-buffered,
already-decrypted packet payload), HTTP/3 frames ride inside QUIC STREAM byte
streams and "can span multiple packets" (§7) — this phase's whole scope
centers on `H3FrameDecoder`, an incremental/resumable reader, not a
single-shot parser.

New files (`vlib/net/quic/`), each with a paired `_test.v`:
- `h3_reserved.v` — the shared `0x1f*N+0x21` grease-codepoint formula, used
  identically across frame types, stream types, SETTINGS identifiers, and
  error codes (verified byte-for-byte identical wording in all four RFC
  citations before writing one shared helper instead of four copies).
- `h3_error.v` — `H3ErrorCode`, all 17 values (§8.1/Table 4).
- `h3_stream_type.v` — the unidirectional Stream Type header (§6.2/6.2.1/
  6.2.2/6.2.3): control, push (+ push ID), reserved, unknown. Incremental —
  returns `none`, not an error, on a short buffer.
- `h3_frame.v` — the frame Type/Length envelope (§7.1) + all 7 defined frame
  types' payloads (§7.2.1-7.2.7) + the 4 HTTP/2-carryover reserved frame
  types (§7.2.8, rejected outright as a connection error) vs. grease/
  genuinely-unknown types (§9, preserved and silently ignored, never
  rejected). SETTINGS payload parsing rejects duplicate identifiers and the
  5 reserved HTTP/2-carryover identifiers, while correctly NOT rejecting
  QPACK's own identifiers (RFC 9204, not reserved by 9114 itself) so Phase 11
  can add them without touching this reserved set.
- `h3_message_state.v` — the one piece of §4's message-framing rules that
  needs only a stream's ROLE, not full request/response context: Table 1's
  per-role frame-type legality, and the control-stream
  SETTINGS-must-be-first-and-only-once discipline.

A full requirements matrix (`.claude/skills/code-review/
quic_conformance_matrix.md`, "HTTP/3 framing layer" section) was built
**before** writing any code, listing every requirement in the sections
audited — including explicit N/A-by-role rows (this is a v1 client-only
implementation) and rows deliberately deferred to Phase 12 (cross-frame
state — advertised max push ID, push IDs already seen, control-stream
uniqueness — needs a real connection/request registry this framing-only
phase doesn't have). Nothing is a silent gap.

Self-review caught and fixed a real bug before the first compile: a
declared frame Length is an attacker-controlled u64 up to 2^62-1; casting it
straight to `int` before checking how many bytes were actually buffered
could truncate/wrap on a 32-bit `int`. Fixed by doing the readiness
comparison entirely in u64 space. Regression test:
`test_h3_frame_decoder_huge_declared_length_with_insufficient_bytes_does_not_panic`.

## Test plan

- [x] Full `net.quic` suite green (41/41)
- [x] Full `/vreview` pass (clean beyond the overflow fix and one doc-comment
      correction, both applied during authoring, before this push)
- [x] RFC 9114 conformance matrix built and cross-checked against every row
      before considering the phase complete

### DIFF
--- a/vlib/net/quic/PROGRESS.md
+++ b/vlib/net/quic/PROGRESS.md
@@ -914,15 +914,79 @@ to a later phase — `stateless_reset.v`/`pmtu.v` both say so independently):
 `QuicConn` holds exactly one local `scid` and tracks the peer's current
 `dcid`, no active CID set, no NEW_CONNECTION_ID/RETIRE_CONNECTION_ID.
 
-## Phases 10-14 (NOT STARTED)
+## Phase 10: HTTP/3 framing (RFC 9114) — CODE COMPLETE, not yet wired to conn.v
+
+Full section-by-section requirements matrix built BEFORE writing any code
+(`.claude/skills/code-review/quic_conformance_matrix.md`, "HTTP/3 framing
+layer" section) — the direct structural response to Phase 9's own
+postmortem ("build the RFC checklist before/during implementation, not
+reactively"). See that section for the complete requirement-by-requirement
+breakdown; summary here.
+
+New files, all in `vlib/net/quic/`:
+
+- [x] `h3_reserved.v` — the single 0x1f*N+0x21 grease-codepoint formula
+      shared identically across frame types (§7.2.8), stream types
+      (§6.2.3), SETTINGS identifiers (§7.2.4.1), and error codes (§8.1) —
+      confirmed byte-for-byte identical wording in all 4 RFC citations
+      before writing one shared helper instead of four copies.
+- [x] `h3_error.v` — `H3ErrorCode` enum, all 17 values (§8.1/Table 4).
+- [x] `h3_stream_type.v` — unidirectional Stream Type header (§6.2/6.2.1/
+      6.2.2/6.2.3): control (0x00), push (0x01 + push ID), reserved,
+      unknown. Incremental (returns `none`, not an error, on a short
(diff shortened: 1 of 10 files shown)


JalonSolov:
GPT-5.6 Luna said:  No actionable findings.

quaesitor-scientiam:
> GPT-5.6 Luna said: No actionable findings.

Wow, maybe my explicit instructions are to do this correctly this time worked

quaesitor-scientiam:
All CI failures on PR #28094's latest commit (515750e30f) are diagnosed and confirmed pre-existing/infra — none caused by this PR. The docker-ubuntu-musl fix from last round worked (h3 files no longer implicated; it's now red for an unrelated libpq-fe.h gap in db.pg), and the previously-unseen termux-build/build-vc failures turned out to be a dead apt mirror and the known fork-token 403 on the diagnose step, respectively.  So this PR is ready to merge.

PR #28103, opened 2026-08-16
HTTP/3 (QUIC) foundation: Phase 11 (QPACK field compression, RFC 9204)

quaesitor-scientiam:
## Summary

Implements Phase 11 of the HTTP/3/QUIC client (tracking issue #27675): QPACK field compression (RFC 9204).

- **Primitives** (`qpack_primitives.v`): prefixed integer + string literal codec, generalized to QPACK's variable prefix widths, verified against the already-shipped `h2_hpack.v` HPACK codec before writing.
- **Huffman** (`qpack_huffman_table.v` / `qpack_huffman.v`): a byte-verified verbatim copy of `h2_hpack_huffman_table.v`/`h2_hpack_huffman.v` — RFC 9204 §4.1.2 mandates identical reuse of RFC 7541 Appendix B's table.
- **Tables** (`qpack_static_table.v`, `qpack_dynamic_table.v`): all 99 static entries; dynamic table with insert/evict/duplicate/capacity, 4 distinct absolute/relative/post-Base indexing resolution functions, reference-counted eviction protection.
- **Instruction codecs** (`qpack_encoder_instructions.v`, `qpack_decoder_instructions.v`): all 4 encoder-stream + 3 decoder-stream instruction types.
- **Field line codec** (`qpack_field_line.v`): Required Insert Count wraparound math + Base sign/delta math transcribed directly from the RFC's own pseudocode, all 6 field line representation types.
- **Drivers** (`qpack_encoder.v`, `qpack_decoder.v`): full `QpackEncoder`/`QpackDecoder` state machines — Known Received Count tracking, blocked-field-section detection, invalid-reference rejection.
- Plumbing: `qpack_error.v`, `qpack_stream_type.v` (mirrors `H3ControlStreamState`'s shape), `qpack_settings.v`.

Built the RFC 9204 requirements matrix (`quic_conformance_matrix.md`) from a full read of the spec (all of §1-§8, Appendix A, Appendix B, Appendix C) before writing any code, mirroring Phase 10's methodology.

**Scope boundary** (mirrors Phase 10's own precedent): the tables, wire codecs, and encoder/decoder state machines are fully self-contained — Phase 12 only needs to feed them bytes from real QUIC streams and wire in peer SETTINGS values.

## Self-review findings (both fixed pre-push)

1. A doc-comment/code mismatch in `encode_field_section`, resolved by sibling comparison against `h2_hpack.v`'s `encode_field` — the original code (gating a static exact-match behind `never_index`) was correct; the doc comment claiming "always used" was wrong.
2. A real gap: `qpack_is_sensitive` existed but was never wired into the encoder, unlike its HPACK sibling which applies the equivalent check unconditionally — a caller forgetting to flag a Cookie/Authorization field got no protection. Fixed to match HPACK's defense-in-depth behavior.

## Test plan

- [x] Full `net.quic` suite green (52/52 test files), zero regressions in Phases 0-10
- [x] End-to-end verification against all 5 of RFC 9204 Appendix B's byte-exact worked examples
- [x] Per-file edge-case tests: prefixed-int/string boundary values and overflow rejection, dynamic-table eviction/evictability/indexing-context boundaries, RIC/Base wraparound math, all 7 instruction types round-tripped, blocked-field-section and invalid-reference detection
- [x] `/vreview` high-effort pass (2 findings, both fixed)


quaesitor-scientiam:
Every failing job on this PR maps 1:1 to master's own current failure set except build-vc, which is the already-diagnosed fork-PR infra limitation. Nothing here traces back to Phase 11's code

JalonSolov:
GPT-5.6 Luna Findings

**P1:** Dynamic name references are not tracked in qpack_encoder.v. When insertion is unavailable, the encoder emits a field line referencing an acknowledged dynamic name at qpack_encoder.v, but does not add that entry to `referenced` or update `required_insert_count`. The resulting field section advertises RIC `0` while referencing a dynamic entry, so the decoder rejects it as `QPACK_DECOMPRESSION_FAILED`. It also permits eviction of the referenced entry before acknowledgment.

**P1:** Malformed encoder-stream instructions are treated as incomplete input. qpack_encoder_instructions.v converts all decode errors, including invalid Huffman data and oversized integers, into `none`; qpack_decoder.v consequently returns `applied: false` instead of raising `QPACK_ENCODER_STREAM_ERROR`. A peer can therefore leave the decoder waiting indefinitely rather than triggering the required connection error.

quaesitor-scientiam:
Fixed 2 P1 findings from GPT-5.6 Luna's review in `5c01b8205b`:

**F1 — Dynamic name references not tracked in `encode_field_section`.** When
insertion was unavailable, the encoder emitted a field line referencing an
already-acknowledged dynamic entry by name, but never added that entry to
`referenced` or folded it into `required_insert_count`. The resulting field
section advertised RIC 0 while pointing at a dynamic entry, which the decoder
correctly rejects as `QPACK_DECOMPRESSION_FAILED` (RFC 9204 §4.5.1.1); it also
left the referenced entry evictable before the peer acknowledged the section.
Fixed by adding the same `add_ref`/`referenced`/RIC bookkeeping this branch's
sibling (the name+value dynamic match) already did.

**F2 — Malformed encoder-stream instructions treated as incomplete input.**
`qpack_encoder_instructions.v` collapsed every decode failure — genuine
truncation *and* malformed data (invalid Huffman codes, oversized integers
past the implementation limit) — into a bare `none`, so `qpack_decoder.v`
returned `applied: false` instead of raising
`QPACK_ENCODER_STREAM_ERROR` (RFC 9204 §4.5, §8.3). A peer sending malformed
bytes could leave the decoder waiting indefinitely rather than tripping the
required connection error. Fixed by distinguishing the two cases at the
primitive layer (`decode_prefixed_int`/`decode_prefixed_string` now tag
genuine truncation with a private sentinel error code) and threading that
distinction through `decode_qpack_encoder_instruction`, which now returns
`!QpackDecodedEncoderInstruction` instead of an `Option` so malformed input
propagates as a real error. Applied the identical fix to the decoder-stream
instruction decoder (`qpack_decoder_instructions.v`) even though it has no
production caller yet, since it had the same latent bug.

Both reproduced via a Phase-R test against the unfixed code before the fix
landed (confirmed each failed with the exact predicted symptom), plus two
sibling sweeps afterward: a full audit of every remaining `or {}` swallow in
the QPACK files for the same truncated-vs-malformed conflation, and a from-
scratch re-derivation of `encode_field_section`'s complete contract to check
for other RIC/eviction-timing gaps. No further instances found. Full
`net.quic` suite green (52/52) after the fix.

quaesitor-scientiam:
No action needed on this PR for any of the CI failures— they're vlang/v-wide, not QPACK-specific.. Nothing here traces back to Phase 11's code.  If no further AI checks can this be merged @JalonSolov

JalonSolov:
With all the github problems today, I was unable to do a final check.  Perhaps tomorrow.

quaesitor-scientiam:
> With all the github problems today, I was unable to do a final check. Perhaps tomorrow.

Github issues are resolved as of yesterday

JalonSolov:
Gpt-5.6 Luna Findings

**P1:** `QpackEncoder.set_capacity` can evict entries with outstanding references. After encoding a field section referencing a dynamic entry, reducing capacity can remove that entry before the decoder acknowledges the section, causing decompression failure. Capacity reduction must preserve referenced entries or be rejected. `vlib/net/quic/qpack_encoder.v:72`

**P2:** `encode_ric` divides by zero when `max_table_capacity < 64` and `req_insert_count > 0`. The decoder handles this case, but the public encoder does not. `vlib/net/quic/qpack_field_line.v:48`

quaesitor-scientiam:
Fixed 2 more P1 findings from GPT-5.6 Luna's review, plus one self-caught during this round's `/vreview`, in `9ef5aedb5d`:

**P1 — `QpackEncoder.set_capacity` could evict entries with outstanding
references.** It called `dynamic_table.set_capacity()` unconditionally,
which evicts oldest-first purely on `cur_size > capacity` — unlike `insert`,
which goes through `can_insert()` first specifically to simulate eviction
safety (RFC 9204 §2.1.1: an entry with any outstanding reference, or not yet
acknowledged, must not be evicted) before committing to it. Fixed by adding
`QpackDynamicTable.can_set_capacity`, mirroring `can_insert`'s existing
pattern, and gating the encoder-level `set_capacity` on it before applying a
reduction. The lower-level `QpackDynamicTable.set_capacity` stays
unconditional by design — the decoder calls it directly on every
encoder-supplied capacity with no reject option, which is correct since only
the encoder's own reference bookkeeping can know what's safe to evict.

**P2 — `encode_ric` divided by zero for small `max_table_capacity`.** The
actual threshold is `< 32`, not `< 64` as originally cited —
`qpack_max_entries = capacity/32`, so `full_range` only reaches 0 below 32.
`decode_ric` already guarded this exact case (`full_range == 0`) two lines
below; `encode_ric` never got the equivalent guard, crashing the whole
process on `% full_range` instead of returning an error. Fixed identically,
rippling `encode_ric`/`encode_field_section_prefix`/`encode_field_section`
to Result-returning signatures.

**Self-caught — the fix for P1 had its own bug.** `/vreview`'s mandatory
fresh-contract + sibling-parity check (comparing the new `can_set_capacity`
against `can_insert` clause-by-clause) found its tail case returned a bare
`true` instead of `freed >= over` — wrong for any negative `new_capacity`,
though unreachable via the only current caller today since it's a `pub`
function. Fixed to match `can_insert`'s own tail exactly.

All three Phase-R reproduced against the pre-fix code before fixing. Updated
`quic_conformance_matrix.md`: row 557 was a false checkmark (claimed
`set_capacity` respected the evictable-only constraint when it didn't) and
row 574 now notes `encode_ric`'s guard. Full `net.quic` suite green (52/52)
throughout.

JalonSolov:
Local AI says:  No blocking issues found.

PR #28109, opened 2026-08-17
arrays: add merge_desc for descending sorted arrays

ruedigerlenz:
# What does this PR do?

Add arrays.merge_desc, which merges two descending-sorted arrays while
preserving their order.

# Tests

All tests passed.

### DIFF
--- a/vlib/arrays/arrays.v
+++ b/vlib/arrays/arrays.v
@@ -85,9 +85,41 @@ pub fn merge[T](a []T, b []T) []T {
 	mut ia := 0
 	mut ib := 0
 	mut j := 0
-	// TODO: efficient approach to merge_desc where: a[ia] >= b[ib]
 	for ia < a.len && ib < b.len {
 		if a[ia] <= b[ib] {
+			m[j] = a[ia]
+			ia++
+		} else {
+			m[j] = b[ib]
+			ib++
+		}
+		j++
+	}
+	// a leftovers
+	for ia < a.len {
+		m[j] = a[ia]
+		ia++
+		j++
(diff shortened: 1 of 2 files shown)


antono3 (approved):
Replayed cleanly on current master and reviewed the two-pointer merge; I found no issue.

Independent Actions coverage passed formatting and the full `arrays` module suite. I also added review-only probes for descending strings and a custom type defining `<`; equal keys retain left-array precedence, matching `arrays.merge` stability.

Actions: https://github.com/antono3/v/actions/runs/34110099595
Additional test commit: https://github.com/antono3/v/commit/5606950f426d3a63295b19eee46e054063cdd9c7

medvednikov:
@codex review

medvednikov:
I think it’s a reasonable stdlib addition, with one API-design caveat.

The case for it is fairly strong. arrays.merge already exists specifically for merging two ascending-sorted arrays in O(n+m), and the source even contains an explicit TODO for an efficient descending equivalent. This PR essentially completes that symmetry with the expected two-pointer implementation. It preserves equal-element stability by taking from the left array first, and an independent reviewer tested strings and a custom ordered type in addition to the submitted integer tests.

It also solves a real enough primitive operation. The alternatives are worse:

concatenate + sorted(a > b) → O((n+m) log(n+m))
reverse both arrays, merge, then reverse → still linear, but more passes/copies and much less direct
handwritten two-pointer loop at every call site → exactly the sort of boring algorithm a utility module can sensibly own

medvednikov:
Thanks!

PR #28112, opened 2026-08-17
orm: deprecate orm_fn (Function Call API), removal deadline 2027-08-17

Jengro777:
Declare that the Function Call API (orm_fn, `orm.new_query[T]` / QueryBuilder) will be moved out of the standard library one year after this deprecation, to be maintained in a separate repository.

- orm_func.v: mark `new_query` (the API entry point) with @[deprecated] and @[deprecated_after]
- README.md / doc/docs.md: document the deprecation

First step towards #27001.


Jengro777:
@Copilot  review

vlib/orm/orm_func.v
```diff
@@ -26,6 +26,8 @@ pub mut:
 }
 
 // new_query create a new query object for struct `T`
+@[deprecated: 'the Function Call API (orm_fn) will be moved out of the standard library after 2027-08-17; see https://github.com/vlang/v/issues/27001']
+@[deprecated_after: '2027-08-17']
```
medvednikov: `@[deprecated_after]` is the date on which V starts deprecating the symbol, not a removal deadline. The checker emits only a note before this date, a warning from this date, and a hard error 180 days later. With `2027-08-17` here, callers receive no deprecation warning during the promised one-year migration window, while the docs say the API is deprecated now and will be removed after that date. Please remove this attribute and keep `@[deprecated]` for immediate warnings; enforce the actual removal separately on the planned date.

vlib/orm/README.md
```diff
@@ -6,6 +6,15 @@ regardless of the DB driver you decide to use.
 Driver authors using the shared SQL generators can target SQLite, PostgreSQL, MySQL, and
 H2-backed connections with the built-in ORM dialect helpers.
 
+## Deprecation notice: orm_fn
```
medvednikov, line 9: Non-blocking: please repeat this warning directly under `## Function Call API` (or rewrite that section's introduction as legacy/deprecated). The README links readers straight to `#function-call-api`, where the current text still positively recommends the API, so deep-linked readers can miss this notice.

medvednikov (changes requested):
Requesting changes for one semantic blocker in the deprecation timeline. The compiler's `deprecated_after` attribute schedules when deprecation begins; it does not represent a removal deadline. CI/check status was intentionally not considered.

Non-blocking documentation note: readers following the existing `#function-call-api` anchor can bypass the new top-level warning, so the legacy section should also identify itself as deprecated.

medvednikov:
## Review submitted: **Changes requested**

I submitted a `CHANGES_REQUESTED` review on **vlang/v#28112**, anchored to head commit `42d00b11ef2ff54406078a296d9b579dbe7e2eec`. CI and check results were excluded as requested.

### Blocker — `deprecated_after` encodes the wrong date

The PR adds:

```v
@[deprecated: '... moved out ... after 2027-08-17 ...']
@[deprecated_after: '2027-08-17']
pub fn new_query[T](conn Connection) &QueryBuilder[T]
```

V interprets `deprecated_after` as **the date deprecation begins**, not as a removal deadline:

* Before August 17, 2027: emit a note saying the function *will be deprecated*.
* Starting August 17, 2027: emit a deprecation warning.
* After another 180 days—February 13, 2028—make usage a compiler error.

That behavior comes directly from `Checker.deprecate`.

This contradicts the documentation, which says the API is deprecated now and will be removed after August 17, 2027. It also eliminates the intended one-year warning period: warnings would begin only when removal is supposed to happen.

**Recommended correction:**

```v
@[deprecated: 'the Function Call API (orm_fn) will be moved out of the standard library after 2027-08-17; see https://github.com/vlang/v/issues/27001']
pub fn new_query[T](conn Connection) &QueryBuilder[T] {
```

In other words, remove `@[deprecated_after]`. Plain `@[deprecated]` produces immediate warnings; removal can be performed separately at the planned deadline.

### Non-blocking documentation issue

The new warning appears near the top of `vlib/orm/README.md`, but the README links directly to `#function-call-api`. Readers following that anchor land on a section that still says “You can utilize the Function Call API” and describes it positively, without a nearby warning.

I recommended repeating the warning directly under `## Function Call API`, or rewriting that section’s opening as legacy/deprecated documentation.

The earlier Copilot date-message comment has already been addressed and resolved. No other blockers were found in the three-file diff.

medvednikov:
@codex review

medvednikov:
Addressed the current Codex finding while preserving the documented removal date:

- scheduled `orm.new_query` deprecation warnings for 2027-02-18, exactly 180 days before calls become compiler errors on 2027-08-17
- routed ORM-internal users through a private helper so recommended ORM APIs do not emit migration notices for their own implementation
- documented the notice/warning/error timeline

Validation: all 39 `vlib/orm/` tests pass; `check-md` completed for both changed Markdown files with no errors or formatting errors; V formatting and `git diff --check` pass.

@codex review

medvednikov:
@codex review

PR #28120, opened 2026-08-18
veb, fasthttp: bind to the configured host instead of the wildcard address

### LINKED ISSUE #28119: veb: run_at(host:) is ignored — server always binds wildcard 0.0.0.0/:: (fasthttp backend)
### Describe the bug

`veb.run_at` (and `veb.run`) accept a `host:` parameter, and the startup log prints it (e.g. `Running multi-threaded app on http://127.0.0.1:5016/`), but the actual listening socket **ignores it and always binds the wildcard address** (`0.0.0.0` / `::`). A service explicitly configured with `host: '127.0.0.1'` is still reachable from every network interface.

### Reproduction Steps

```v
module main

import veb

struct App {
	veb.StaticHandler
}

struct Context {
	veb.Context
}

fn main() {
	mut app := &App{}
	veb.run_at[App, Context](mut app, host: '127.0.0.1', port: 5599) or { panic(err) }
}

	return ctx.text('hello')
}
```

Run it, then from another terminal:

```sh
lsof -nP -iTCP:5599 -sTCP:LISTEN
```

### Expected Behavior

Socket bound to `127.0.0.1:5599` only (loopback), matching the `host:` parameter and the startup message.

### Current Behavior

```
COMMAND  PID  USER  FD   TYPE DEVICE SIZE/OFF NODE NAME
main    7358  byf   11u  IPv6 ...      0t0  TCP *:5599 (LISTEN)
```

The socket listens on `*` (all interfaces, both IPv4 and IPv6 via the dual-stack `::` wildcard), reachable from the LAN/public internet despite `host: '127.0.0.1'`.

Root cause in vlib (commit 4a8793cbfc):

1. `vlib/veb/veb_fasthttp.v:98-104` — `veb.run_at` builds `fasthttp.ServerConfig{family: ..., port: ...}`; the `host` parameter is never passed down. `vlib/fasthttp/fasthttp.v:167` (`pub struct ServerConfig`) has **no `host` field at all** — only `family` and `port`.
2. `vlib/fasthttp/fasthttp_bsd.c.v:693-698` — `create_server_socket` hardcodes the wildcard address:
   ```v
   addr := if server.family == .ip6 {
       net.new_ip6(u16(server.port), [u8(0), 0, 0, ... /* all zero = :: */ ]!)
   } else {
       net.new_ip(u16(server.port), [u8(0), 0, 0, 0]!)  // 0.0.0.0
   }
   ```
3. `vlib/fasthttp/fasthttp_bsd.c.v:842` even prints `listening on http://0.0.0.0:${port}/`, while `vlib/veb/veb_fasthttp.v:111` prints the *configured* host — the two startup messages can disagree, which hides the misconfiguration.

So `host:` currently only affects the cosmetic startup log, not the bind address. This is a security-relevant footgun: developers routinely rely on `host: '127.0.0.1'` to keep an admin/debug HTTP service local-only behind a reverse proxy.

### Possible Solution

- Add `host string` to `fasthttp.ServerConfig` (default `''` = wildcard for backward compatibility).
- In `create_server_socket`, when `host` is set, resolve/parse it (at minimum dotted IPv4 and IPv6 literals, e.g. via `net.parse_ip`/`net.resolve_addrs`) and bind to that address instead of the wildcard.
- Pass `params.host` through from `veb.run_at` / `veb_fasthttp.v` into `ServerConfig`.

Found while hardening a small veb admin service: even after setting `host: '127.0.0.1'`, the port stayed open on all interfaces; only application-level auth prevented anonymous writes. Happy to test any patch on macOS (kqueue path).

### V version

V 0.5.2 4a8793cbfc

macOS 26 (arm64, Apple Silicon), kqueue backend (`fasthttp_bsd.c.v`). The Linux (io_uring/epoll) backend likely needs the same fix.

> [!NOTE]

> [!NOTE]
>

medvednikov:
## Fixes #28119

`veb.run_at` / `veb.run` accept a `host:` parameter and the startup log prints it
(e.g. `Running multi-threaded app on http://127.0.0.1:5599/`), but the parallel
**fasthttp** backend ignored it: `create_server_socket` always bound the wildcard
address (`0.0.0.0` / `::`). A service configured with `host: '127.0.0.1'` was still
reachable on every network interface — a security-relevant footgun for admin/debug
services meant to stay loopback-only behind a reverse proxy.

### Root cause
- `fasthttp.ServerConfig` had no `host` field, and `veb`'s fasthttp backend never
  passed `params.host` down.
- `create_server_socket` (BSD/kqueue, Linux/epoll, Windows/IOCP) hardcoded the
  wildcard bind address.
- The backend even printed `listening on http://0.0.0.0:${port}/`, which could
  disagree with veb's own startup line and hide the misconfiguration.

### Fix
- Add a `host` field to `fasthttp.ServerConfig` and to the per-platform `Server`
  structs; wire `params.host` through from `veb.run_new`.
- Add a shared `resolve_bind_addr(host, family, port)` helper:
  - empty `host` → historical wildcard bind (`0.0.0.0` for `.ip`, dual-stack `::`
    for `.ip6`), so **existing servers are unaffected**;
  - non-empty `host` → resolved via `net.resolve_addrs`/getaddrinfo (numeric
    IPv4/IPv6 literals and host names) and bound explicitly.
- The **resolved address's family wins**, so `host: '127.0.0.1'` binds loopback
  IPv4 even with the default `family: .ip6` (the exact config in the issue). All
  three backends now create the listening socket with `addr.family()`.
- The `listening on …` startup line reflects the configured host.

### Verification
Repro from the issue (`host: '127.0.0.1'`, default `family: .ip6`):

Before → `TCP *:5599 (LISTEN)` (all interfaces, IPv6 dual-stack).
After:
```
$ lsof -nP -iTCP:15701 -sTCP:LISTEN
COMMAND   PID USER   FD   TYPE ... NODE NAME
srv     56340 alex    3u  IPv4 ... TCP 127.0.0.1:15701 (LISTEN)
```
The default (no `host`) path still binds the wildcard `*:port`, unchanged.

### Tests
- `fasthttp_bind_addr_test.v` (cross-platform): unit tests for `resolve_bind_addr`
  — wildcard for empty host, IPv4 literal overriding the default `.ip6` family,
  matching-family, IPv6 literal, and unresolvable-host error.
- `fasthttp_bsd_host_bind_test.c.v` (kqueue): end-to-end — a server with
  `host: '127.0.0.1'` and default `family: .ip6` binds loopback-only (verified via
  `getsockname`) and still serves requests.

### DIFF
--- a/vlib/fasthttp/fasthttp.v
+++ b/vlib/fasthttp/fasthttp.v
@@ -163,13 +163,66 @@ fn (mut resp HttpResponse) take_request_arena() voidptr {
 	return request_arena
 }
 
+// resolve_bind_addr computes the socket address a fasthttp listener binds to.
+//
+// An empty `host` keeps the historical wildcard bind — 0.0.0.0 for .ip and the
+// dual-stack :: for .ip6 — so existing servers are unaffected. A non-empty
+// `host` is resolved (numeric IPv4/IPv6 literals and host names, via
+// getaddrinfo) and bound explicitly; that is what makes `host: '127.0.0.1'`
+// reachable only on loopback instead of on every interface.
+//
+// The resolved address's family can differ from the requested `family` (e.g. an
+// IPv4 literal while `family` is still the default .ip6), so callers must create
+// the listening socket with `addr.family()`, not the originally requested one.
+fn resolve_bind_addr(host string, family net.AddrFamily, port int) !net.Addr {
+	if host == '' {
+		if family == .ip6 {
+			return net.new_ip6(u16(port), [u8(0), 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]!)
+		}
+		return net.new_ip(u16(port), [u8(0), 0, 0, 0]!)
+	}
+	// A bare IPv6 literal must be bracketed so split_address keeps the whole
+	// address instead of treating the last group as a port.
+	saddr := if host.contains(':') && !host.starts_with('[') {
+		'[${host}]:${port}'
+	} else {
+		'${host}:${port}'
+	}
+	// Prefer the requested family, but fall back to any family, so an IPv4/IPv6
+	// literal that does not match the default family still binds as intended.
+	if addrs := net.resolve_addrs(saddr, family, .tcp) {
+		if addrs.len > 0 {
+			return addrs[0]
+		}
+	}
+	addrs := net.resolve_addrs(saddr, .unspec, .tcp) or {
+		return error('could not resolve listen host `${host}`: ${err}')
+	}
+	if addrs.len == 0 {
+		return error('could not resolve listen host `${host}`')
+	}
+	return addrs[0]
+}
+
+// listen_host_display renders the configured host for the "listening on ..."
+// startup line; an empty host is shown as the wildcard 0.0.0.0.
+fn listen_host_display(host string) string {
+	return if host == '' { '0.0.0.0' } else { host }
+}
+
 // ServerConfig bundles the parameters needed to start a fasthttp server.
 pub struct ServerConfig {
 pub:
-	family                  net.AddrFamily = .ip6
-	port                    int            = 3000
-	max_request_buffer_size int            = 8192
-	timeout_in_seconds      int            = 30
+	family net.AddrFamily = .ip6
+	// host is the interface address the listening socket binds to. Empty (the
+	// default) keeps the historical wildcard bind — 0.0.0.0 for .ip and the
+	// dual-stack :: for .ip6 — reachable on every interface. Set it to a
+	// numeric address or host name (e.g. '127.0.0.1') to bind only that
+	// interface. See resolve_bind_addr for the resolution rules.
+	host                    string
+	port                    int = 3000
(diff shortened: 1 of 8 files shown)


medvednikov:
@codex review

medvednikov:
## Re-review: no findings

Reviewed the updated head `e03ac774d1d2882791a703ef458539e21694e44b`. **Verdict: approve.**

Both previous findings are resolved:

* **Linux startup failures now propagate correctly.** The bind address is resolved once before worker creation, `create_server_socket` returns `!int`, and socket, bind, listen, and epoll setup failures return errors instead of silently succeeding.
* **IPv6 listener URLs are now valid and accurate.** Wildcards reflect the configured family, bare IPv6 addresses are bracketed, and the veb startup line applies equivalent formatting.
* Regression coverage now includes the Linux `run()` error path, IPv4 overriding the default IPv6 family, wildcard behavior, and IPv6 display formatting.

I found no new blocking issues in the BSD, Linux, Windows, or veb changes. CI was ignored as requested. This was a source/diff review; I could not execute the V tests locally because the environment did not have the repository or V toolchain available.

PR #28121, opened 2026-08-18
vdoc, missdoc, cgen, doc: fix packaging/tooling issues (#28115, #28114, #28108)

### LINKED ISSUE #28108: 0.5.2: `v -prod build-tools` runs out of memory; `vbug-report.v` fails to compile on glibc (L_tmpnam)
Two build regressions in 0.5.2 (0.5.1 built cleanly), found while packaging V for Homebrew.

Env: Linux x86_64 (glibc 2.35, gcc 11.4) and macOS 26 arm64 (Apple clang 21).

### 1. `v -prod build-tools` exhausts memory
`doc/packaging_v_for_distributions.md` recommends `./v -prod build-tools`. On 0.5.2 the `-prod` C compile (LTO) consumes runaway memory on the largest tools — `cmd/tools/builders/c_builder.v` and `cmd/tools/check_os_api_parity.v` each spend ~5-6 min in the C compiler and then fail (killed, no C diagnostic). Reproduced on several machines including a 48 GB one. Dropping `-prod` compiles the same tools in seconds.

```
./v -prod -d dynamic_boehm build-tools
# -> Summary for building cmd/tools: 2 failed  (c_builder.v, check_os_api_parity.v)
```

### 2. `vbug-report.v` generated C fails to compile on glibc (Linux)
Independent of `-prod`, `cmd/tools/vbug-report.v` fails on Linux:

```
/usr/include/stdio.h:205:27: error: 'L_tmpnam' undeclared here (not in a function)
```

`L_tmpnam` is defined unconditionally in `bits/stdio_lim.h`, and only `L_tmpnam` is affected (not `FOPEN_MAX`, `TMP_MAX`, ...), so the generated C appears to `#undef L_tmpnam` (or otherwise drop it) before `<stdio.h>` is processed. macOS is unaffected.

```
./v -d dynamic_boehm -cc gcc cmd/tools/vbug-report.v
```

### Downstream
Worked around in Homebrew by building `build-tools` without `-prod` and skipping `vbug-report.v` on Linux: Homebrew/homebrew-core#292840

> [!NOTE]
>

### LINKED ISSUE #28114: v missdoc -e 'string' is ignored, v missdoc --exclude 'string' works fine
`v missdoc -e 'string` or `v missdoc --exclude 'string'` will exclude all files with a path which include ' string'  from checking on missing documentations.
While `v missdoc --exclude 'main'` works fine, is the given string 'main' on `v missdoc -e 'main'` just ignored:
```
> v missdoc --exclude 'main'
---
> v missdoc -r -e 'main'
b/main.v:7:0:pub fn inline() 
b/main.v:9:0:pub fn ohne_attribute() 
b/main.v:20:0:pub fn (t Tee) methode() 
a/main.v:7:0:pub fn inline() 
a/main.v:21:0:pub fn (t Tee) methode() 
main.v:7:0:pub fn inline() 
main.v:9:0:pub fn ohne_attribute() 
main.v:20:0:pub fn (t Tee) methode() 
```

I believe `-e` should work as `--exclude` is working....
> [!NOTE]
>

### LINKED ISSUE #28115: v doc -theme-dir <directory> works - but only, if arguments are given in defined series
While using the wrong series of the arguments the option `-theme-dir` of the command `v doc` is not working:
```
> v doc -f html -theme-dir theme ./main.v -o html_theme
vdoc: No valid V files were found. Skipping folder: /home/matthias/Dokumente/v/trials/doc/theme.
vdoc: No documentation found for /home/matthias/Dokumente/v/trials/doc/theme
```
Switching '-theme-dir theme' and './main.v' however works:
```
[matthias@infinitybook doc]$ v doc -f html ./l225.v -o html_theme -theme-dir theme
Copying    18032 bytes from `theme/doc.css` to `/home/matthias/Dokumente/v/trials/doc/html_theme/doc.css` ...
Copying     2091 bytes from `theme/normalize.css` to `/home/matthias/Dokumente/v/trials/doc/html_theme/normalize.css` ...
Copying    10895 bytes from `theme/doc.js` to `/home/matthias/Dokumente/v/trials/doc/html_theme/doc.js` ...
Copying      133 bytes from `theme/dark-mode.js` to `/home/matthias/Dokumente/v/trials/doc/html_theme/dark-mode.js` ...
Generating     4016 bytes of html in `/home/matthias/Dokumente/v/trials/doc/html_theme/main.html` ...
Creating search index...
Generating search_index.js of      158 bytes in `/home/matthias/Dokumente/v/trials/doc/html_theme/search_index.js ...
Copying favicons...
Copied 14 icons to `/home/matthias/Dokumente/v/trials/doc/html_theme` .
```
Acccording to claude.ai the fix is trivial:
Inside the parser cmd/tools/vdoc/main.v there is a `i++` missing:

```
'-theme-dir' {
    cfg.theme_dir = cmdline.option(current_args, '-theme-dir', default_theme)
    // missing i++ !
}
```

Did not checked, if that's correct...

> [!NOTE]
>

medvednikov:
Fixes three separately reported issues.

### `v doc -theme-dir <dir>` order-dependent (fixes #28115)
The `-theme-dir` case in `cmd/tools/vdoc/main.v` did not advance the argument index past its value, so the directory name was consumed as the input path/symbol filter. `v doc -f html -theme-dir theme main.v -o out` therefore failed with *"No valid V files were found"*, while the same options in a different order worked. Added the missing `i++`.

### `v missdoc -e 'x'` silently ignored (fixes #28114)
`v missdoc -e main` ignored the exclude pattern, while `v missdoc --exclude main` worked. The cause was **not** in `missdoc` or the `flag` module (both handle `-e` correctly) — V's own argument parser interpreted the `-e` *after* the `missdoc` command as its eval-argument flag (`v -e 'code'`) and rewrote the whole invocation into a temp-`.vsh` run before the tool ever saw it.

Fix: once a known external command has been recognized, stop interpreting the following arguments as V compiler flags — the external tool owns its argument space, and `launch_tool` already forwards the original `os.args` verbatim. Added a regression test in `vlib/v/pref/pref_test.v`.

### `vbug-report.v` fails to compile on glibc: `'L_tmpnam' undeclared` (fixes #28108)
Since 0.5.2, V declares the stdio functions manually and no longer includes `<stdio.h>` in its C prelude. On glibc, the stdio limit macros (`L_tmpnam`, `FILENAME_MAX`, `TMP_MAX`, …) are only defined *while `<stdio.h>` is being processed*. When a module header (`sqlite3.h`, `gc.h`, …) later pulls `<stdio.h>` in, it can be entered with those macros no longer defined, breaking glibc's own `tmpnam` prototype:

```
/usr/include/stdio.h:205:27: error: 'L_tmpnam' undeclared here (not in a function)
```

0.5.1 built fine because it always included `<stdio.h>`. This restores that include on the glibc/GNU-libc path only, up front, while keeping the manual prototypes (which redeclare the stdio functions compatibly — exactly as already happens today when a module header includes `<stdio.h>`). Other platforms are unaffected.

### `v -prod build-tools` runs out of memory (addresses #28108)
`build-tools` forwards its leading flags to every tool compile, so `-prod` triggers a per-tool LTO C compile that exhausts memory on the largest generated tools (e.g. `c_builder.v`). Updated `doc/packaging_v_for_distributions.md` to build the tools without `-prod`, matching what downstream packagers already do.

### Verification
- `#28115`: both argument orders now generate docs correctly.
- `#28114`: `-e main` now excludes and `--exclude` still works; `vlib/v/pref` tests pass.
- `#28108` (glibc): the generated Linux C now `#include`s `<stdio.h>` on glibc; the `L_tmpnam` prototype has its macro defined. Manual-decls / shadowed-C-fn tests pass and the macOS build is unchanged (the new include is a no-op off glibc). **The glibc compile itself is left for Linux CI to confirm.**

### DIFF
--- a/cmd/tools/vdoc/main.v
+++ b/cmd/tools/vdoc/main.v
@@ -116,6 +116,7 @@ fn parse_arguments(args []string) Config {
 			}
 			'-theme-dir' {
 				cfg.theme_dir = cmdline.option(current_args, '-theme-dir', default_theme)
+				i++
 			}
 			'-l' {
 				cfg.show_loc = true
--- a/cmd/v/v.v
+++ b/cmd/v/v.v
@@ -115,14 +115,14 @@ fn main() {
 			if os.is_atty(0) == 0 {
 				mut args_and_flags := util.join_env_vflags_and_os_args()[1..].clone()
 				args_and_flags << ['run', '-']
-				pref.parse_args_and_show_errors(external_tools, args_and_flags, true)
+				pref.parse_args_for_launcher(external_tools, args_and_flags, true)
 			}
 		}
 		util.launch_tool(false, 'vrepl', os.args[1..])
 		return
 	}
 	mut args_and_flags := util.join_env_vflags_and_os_args()[1..]
-	prefs, command := pref.parse_args_and_show_errors(external_tools, args_and_flags, true)
+	prefs, command := pref.parse_args_for_launcher(external_tools, args_and_flags, true)
 	maybe_delegate_to_vvmrc(command, prefs)
 	maybe_delegate_to_ownership(command, prefs, args_and_flags)
 	macos_v3_c_error_report := maybe_delegate_to_macos_v3(command, prefs)
--- a/doc/packaging_v_for_distributions.md
+++ b/doc/packaging_v_for_distributions.md
@@ -106,7 +106,7 @@ echo "Alternatively, if you do want a more recent V version, just clone V from s
 echo "then follow the instructions here: https://github.com/vlang/v#installing-v-from-source')" >> cmd/tools/vself.v
 
 v -prod -o v cmd/v                            ## build V itself with -prod
-./v -prod build-tools                         ## build all tools with -prod too
+./v build-tools                               ## build all tools (NB: do *not* pass -prod here; some large tools, like c_builder.v, can exhaust memory during the -prod LTO C compile)
 touch ./cmd/tools/.disable_autorecompilation  ## tell V to not try to recompile any tool anymore
 
 ### Cleanup folders that would not be needed inside a package,
--- a/vlib/v/gen/c/cheaders.v
+++ b/vlib/v/gen/c/cheaders.v
@@ -554,6 +554,17 @@ typedef struct _IO_FILE FILE;
 extern FILE* stdin;
 extern FILE* stdout;
 extern FILE* stderr;
+#if defined(__GLIBC__) || defined(__GNU_LIBRARY__)
+// On glibc, the stdio limit macros (L_tmpnam, FILENAME_MAX, TMP_MAX, ...) are
+// only defined while stdio.h itself is being processed. V declares the stdio
+// functions manually and does not include stdio.h here, so a later include of
+// it from a module header (sqlite3.h, gc.h, ...) can be entered with those
+// macros no longer defined, which breaks the glibc tmpnam prototype (an error
+// about L_tmpnam being undeclared, see vlang/v#28108). Pull stdio.h in up front
+// here, so those macros are defined; the manual prototypes below still
+// redeclare the stdio functions compatibly.
+#include <stdio.h>
+#endif
 	#endif
 typedef __builtin_va_list va_list;
 #ifndef va_start
--- a/vlib/v/pref/pref.v
+++ b/vlib/v/pref/pref.v
@@ -423,6 +423,19 @@ fn optional_arg_value(args []string, idx int, command string, known_external_com
 }
 
 pub fn parse_args_and_show_errors(known_external_commands []string, args []string, show_output bool) (&Preferences, string) {
+	return parse_args_impl(known_external_commands, args, show_output, false)
+}
+
+// parse_args_for_launcher works like parse_args_and_show_errors, but once a known external command
+// (tool) is recognized, every argument after it is left for that tool, instead of being interpreted
+// as a V compiler option here. It is meant for the top level `v` launcher in cmd/v, which forwards
+// the original os.args on to the tool verbatim. Do NOT use it when you actually need the V
+// preferences that follow the command name, for example `v fmt -translated file.v`; see vlang/v#28114.
+pub fn parse_args_for_launcher(known_external_commands []string, args []string, show_output bool) (&Preferences, string) {
+	return parse_args_impl(known_external_commands, args, show_output, true)
+}
+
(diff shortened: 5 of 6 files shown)


medvednikov:
@codex review

medvednikov:
Addressed the P2 review comment (scope external-command passthrough to the top-level launcher).

The stop-after-external-command behavior is now an **explicit parser mode** rather than being inferred from `known_external_commands`:

- Split `parse_args_and_show_errors` into a private `parse_args_impl(..., pass_external_command_args bool)`.
- `parse_args_and_show_errors` / `parse_args` keep the **old** behavior (`pass_external_command_args = false`) — so `vfmt`, `cbuilder`, tests, etc. that recognize a command name but still need the general V flags after it are unchanged.
- New `parse_args_for_launcher(...)` (`= true`) is used **only** by `cmd/v` (the two launcher call sites).

So `v fmt -translated generated.v`, `v fmt -os linux source.v`, etc. again reach `vfmt`'s `Preferences`, while `v missdoc -e main` still leaves `-e` for the tool.

Added regression tests in `vlib/v/pref/pref_test.v`:
- `test_launcher_leaves_flags_after_a_known_external_command_for_the_tool` — launcher mode still leaves `-e` for the tool (`!prefs.is_eval_argument`).
- `test_non_launcher_parse_keeps_v_flags_after_the_command` — asserts `-translated` (→ `prefs.translated`) and `-os linux` (→ `prefs.os == .linux`) after `fmt` are still parsed in the default (non-launcher) mode.

Verified locally: `vlib/v/pref` tests pass, `v fmt -translated file.v` formats and honors the flag, and `v missdoc -e main` still excludes.

Thanks for catching this — the original guard would indeed have dropped those flags for `vfmt`.

medvednikov:
## Re-review: no findings

Reviewed the updated head **`8d7488ddc0b9a10e971de5861c1d10739f387446`**. CI was ignored as requested.

The previous P2 issue is resolved:

* Passthrough behavior now lives in a dedicated `parse_args_for_launcher()` path.
* Only the top-level `cmd/v` launcher uses that path.
* Existing callers such as `vfmt` continue using `parse_args_and_show_errors()`, so V flags after `fmt` are still parsed.
* The new tests cover both `v missdoc -e ...` passthrough and preservation of `-translated`/`-os` after `fmt`.

I found no further correctness issues in the six changed files. I also rechecked the glibc prelude pattern locally with GCC 14 and Clang 17 under fortified, warning-as-error builds; both accepted it.

**Verdict: approve.** No GitHub review was submitted.
[](url)

Issue #28107, opened 2026-08-16
ORM PostgreSQL: Table name mismatch with module prefix


Gako91:
When using PostgreSQL ORM with structures in a module, the ORM automatically adds the module name as a table prefix (e.g., `models.sendsmsrequest`), but the table is created without this prefix (e.g., `sendsmsrequest`), causing an error during insert operations.

**Reproduction code:**

`main.v`:
```v
module main
import orm
import db.pg
import models

fn main() {
	mut conn := pg.connect(pg.Config{host: "localhost", user: "user", port: 5439, password: "pass", dbname: "db"}, pg.PoolConfig{}) or { panic(err) }
	mut db := orm.new_db(conn, orm.DataScope{})
	sql db { create table models.SendSMSRequest } or { panic(err) }
	mut sms_send := models.SendSMSRequest{msisdn: "123456", msg: "Test", sender: "sender"}
	mut qb := orm.new_query[models.SendSMSRequest](db)
	qb.insert(sms_send) or { panic(err) }
}
```

`models/models.v`:
```v
module models
pub struct SendSMSRequest {
	pub mut:
		id int @[primary; sql: serial]
		msisdn string
		msg string
		sender string
}
```

**Expected behavior:** The ORM should use the same table name for creation and insert operations.

**Actual behavior:**
- Table created: `sendsmsrequest`
- Insert attempted in: `models.sendsmsrequest` → Error "relation does not exist"

**Environment:**
- V version: 0.5.2
- OS: Linux
- Database: PostgreSQL
> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

medvednikov:
please confirm it's been fixed for you

PR #28123, opened 2026-08-19
orm: strip module prefix from query table names so it matches `create table` (fixes #28106, #28107)

### LINKED ISSUE #28106: ORM PostgreSQL: Table name mismatch with module prefix
When using PostgreSQL ORM with structures in a module, the ORM automatically adds the module name as a table prefix (e.g., `models.sendsmsrequest`), but the table is created without this prefix (e.g., `sendsmsrequest`), causing an error during insert operations.

**Reproduction code:**

`main.v`:
```v
module main
import orm
import db.pg
import models

fn main() {
	mut conn := pg.connect(pg.Config{host: "localhost", user: "user", port: 5439, password: "pass", dbname: "db"}, pg.PoolConfig{}) or { panic(err) }
	mut db := orm.new_db(conn, orm.DataScope{})
	sql db { create table models.SendSMSRequest } or { panic(err) }
	mut sms_send := models.SendSMSRequest{msisdn: "123456", msg: "Test", sender: "sender"}
	mut qb := orm.new_query[models.SendSMSRequest](db)
	qb.insert(sms_send) or { panic(err) }
}
```

`models/models.v`:
```v
module models
pub struct SendSMSRequest {
	pub mut:
		msisdn string
		msg string
		sender string
}
```

**Expected behavior:** The ORM should use the same table name for creation and insert operations.

**Actual behavior:**
- Table created: `sendsmsrequest`
- Insert attempted in: `models.sendsmsrequest` → Error "relation does not exist"

**Environment:**
- V version: 0.5.2
- OS: Linux
- Database: PostgreSQL
> [!NOTE]
>

medvednikov:
Fixes #28106 and #28107 (duplicate reports).

### The bug
When an ORM struct is defined in a module, `create table` and the runtime query builder disagree on the table name:

- `sql db { create table models.SendSMSRequest }` creates **`sendsmsrequest`** — the compiler strips the module prefix via `util.strip_mod_name`.
- `orm.new_query[models.SendSMSRequest](db).insert(...)` targets **`models.sendsmsrequest`** — the runtime kept the module prefix, only lowercasing `T.name`.

So inserts/selects fail with `relation "models.sendsmsrequest" does not exist` on PostgreSQL (`no such table: models.sendsmsrequest` on sqlite). Keeping the `.` is also wrong for SQL, where `models.sendsmsrequest` reads as *schema.table*.

### The fix
`vlib/orm/orm_func.v` — in `table_from_struct[T]`, strip the module prefix (`all_after_last('.')`) before lowercasing, so the runtime table name matches what the compiler emits for `create table`:

```v
if !has_custom_table_name {
    table_name = table_name.all_after_last('.')  // models.SendSMSRequest -> SendSMSRequest
    table_name = table_name.to_lower()
}
```

`@[table: '...']` custom names are untouched, and structs in `main` (unqualified `T.name`) are unaffected.

### Test
Added `vlib/orm/orm_module_table_prefix/` — a struct in a `models` submodule, exercised through the exact bug-report flow (compiler `create table` + runtime `new_query.insert` + `select`) on sqlite. It reproduces the mismatch: **without the fix it fails** with `no such table: models.sendsmsrequest`; with the fix it passes. Full `vlib/orm` suite (39 files) stays green.

### DIFF
--- a/vlib/orm/orm_func.v
+++ b/vlib/orm/orm_func.v
@@ -786,6 +786,11 @@ fn table_from_struct[T](meta []TableField) Table {
 		}
 	}
 	if !has_custom_table_name {
+		// Strip the module prefix, e.g. `models.SendSMSRequest` -> `SendSMSRequest`, so the table
+		// name matches the one the compiler generates for `sql db { create table ... }` (which uses
+		// util.strip_mod_name). Otherwise inserts/selects would target `models.sendsmsrequest`, while
+		// the table was created as `sendsmsrequest` (see vlang/v#28106, vlang/v#28107).
+		table_name = table_name.all_after_last('.')
 		// Keep default ORM table names aligned with unquoted SQL identifiers across DB drivers.
 		table_name = table_name.to_lower()
 	}
--- a/vlib/orm/orm_module_table_prefix/models/models.v
+++ b/vlib/orm/orm_module_table_prefix/models/models.v
@@ -1,0 +1,9 @@
+module models
+
+pub struct SendSMSRequest {
+pub mut:
+	id     int @[primary; sql: serial]
+	msisdn string
+	msg    string
+	sender string
+}
--- a/vlib/orm/orm_module_table_prefix/orm_module_table_prefix_test.v
+++ b/vlib/orm/orm_module_table_prefix/orm_module_table_prefix_test.v
@@ -1,0 +1,36 @@
+import db.sqlite
+import orm
+import models
+
+// Regression test for https://github.com/vlang/v/issues/28106 and
+// https://github.com/vlang/v/issues/28107 : an ORM struct defined in a module must resolve to the
+// same table name for `create table` (the compiler path) and for the runtime query builder used by
+// inserts/selects. Previously the query builder kept the module prefix (`models.sendsmsrequest`),
+// while `create table` stripped it (`sendsmsrequest`), so inserts failed with a missing relation.
+fn test_module_prefixed_struct_table_name_is_consistent() {
+	mut db := sqlite.connect(':memory:')!
+	defer {
+		db.close() or {}
+	}
+
+	// created as `sendsmsrequest` - the compiler strips the `models.` module prefix
+	sql db {
+		create table models.SendSMSRequest
+	}!
+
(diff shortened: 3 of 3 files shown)


medvednikov:
## Review result

**No substantive findings. Verdict: approve.**

Reviewed PR #28123 at head `e2142ae962cdde0b94c7194bcf384fda02d3df14`: one commit, three changed files, 50 additions.

### Assessment

The runtime fix in `vlib/orm/orm_func.v:786` is correct. `table_from_struct[T]` already removes generic parameters and resolves `@[table: '…']`; the PR now strips the module prefix only for default names, before lowercasing. Custom table names—including explicitly schema-qualified names—remain unchanged.

This matches the compiler’s naming algorithm:

1. Remove generic parameters.
2. Strip the module prefix.
3. Use the explicit `@[table]` value when present; otherwise lowercase the default name.

The compiler implements this through `get_table_name_by_struct_type` and `util.strip_mod_name`, whose non-generic behavior is likewise `all_after_last('.')`.

The regression test is meaningful rather than merely inspecting an internal value: it creates the table through compiler-generated ORM SQL, inserts through `orm.new_query`, and selects the inserted row through compiler-generated SQL. Before the patch, the insert targets `models.sendsmsrequest`; afterward, both paths target `sendsmsrequest`.

I also traced the affected runtime paths. `new_query[T]` centralizes table construction through `table_from_struct[T]`, and selections, qualified fields, ordering, joins, inserts, aggregates, `save`, bulk inserts, create, and drop all consume that resulting `Table.name`. I found no secondary runtime path that would continue retaining the module prefix.

### Non-blocking compatibility note

Code that deliberately relied on an imported V module name becoming an implicit PostgreSQL schema will change behavior. Such usage can preserve schema qualification explicitly with `@[table: 'schema.table']`. Since compiler-generated ORM operations never followed the implicit-module-as-schema behavior, this is not a reason to block the fix.

The added regression runs against SQLite rather than PostgreSQL, but the corrected logic produces the table-name string before dispatching to a database driver, so the test covers the underlying defect. I reviewed the change statically and did not execute tests locally; CI status was not used in this verdict.

PR #28124, opened 2026-08-19
v3: add vfmt + `-new-compiler` (in-process V3), fix markused helper pruning

medvednikov:
Three related v3 changes (three commits).

### 1. vfmt for v3 (`vlib/v3/gen/v/`)
A V source formatter for v3, ported from the old v2 `gen/v` to v3's flat node AST. It walks a parsed `flat.FlatAst` and reconstructs V source. Validated to re-parse with zero diagnostics and format idempotently across the whole v3 compiler codebase and most of vlib; unit tests in `gen/v/gen_test.v`.

### 2. `-new-compiler` flag — runs vlib/v3 in-process
Symmetric to `-old-compiler`: `-new-compiler` runs the embedded V3 driver (`vlib/v3`) **in the same process** — never a separate `v3` executable — exactly like the default macOS path. It forces V3 for `run`/`build` targets and disables the automatic V1 fallback, so a V3 failure is reported instead of silently retried with V1. The V3 driver is now linked into `cmd/v` on every normal build (not only macOS), so it works cross-platform; the portable cross-VC bootstrap keeps its stub. `-old-compiler` takes precedence when both are given. pref + macos_v3 unit tests included.

### 3. Fix markused pruning of lowering-introduced builtin helpers
#28101 gated markused's body-call collection behind `detect_reachable_generics`. But the body collector is the only source of implicit helpers introduced by lowering (`array__get`, `v_ni_index`, array bounds checks, …); the checker's dependency edges cover only source-level calls. So for non-generic programs those helpers were pruned and left undeclared — which modern clang rejects (the V1 fallback and gcc's warnings masked it, so CI stayed green). This runs body-call collection on every path while keeping the generics bookkeeping gated, so #28101's self-host speedup (where `detect_reachable_generics` is already true) is intact. It also fixes the default macOS V3 path, which was silently falling back to V1 for every program.

### Testing
- `v -new-compiler run examples/hello_world.v` compiles with V3 in-process and produces a working binary (no fallback); same for generics/arrays/maps/strings and fibonacci.
- Default macOS V3 path no longer falls back to V1.
- markused test passes; vfmt/pref/macos_v3 unit tests included.
- `cmd/v` cross-compiles to Linux C (the embed + dispatch compile cross-platform).

Note: a separate, pre-existing issue remains — V3 self-host in C99 mode on modern clang fails on a generic-type forward-decl (`sync__ThreadLocalStorage_T`); it does not affect user programs or `-new-compiler`.

### DIFF
--- a/cmd/v/macos_v3_default.c.v
+++ b/cmd/v/macos_v3_default.c.v
@@ -1,9 +1,42 @@
 module main
 
+import os
 import v.pref
+import v.util
 
+// On platforms where V3 is not the default compiler, `v` still runs the V1
+// compiler normally. `-new-compiler` opts into the embedded V3 driver and runs
+// it in THIS process, the same way V3 runs by default on macOS.
 fn maybe_delegate_to_macos_v3(command string, prefs &pref.Preferences) ?MacosV3CErrorReport {
+	if prefs.new_compiler {
+		run_new_compiler_in_process(command, prefs)
+	}
 	_ = command
-	_ = prefs
 	return none
+}
+
+// run_new_compiler_in_process compiles with the embedded V3 driver (vlib/v3) in
+// the current process. It never launches a separate `v3` executable and never
+// falls back to V1, so a V3 failure is reported as-is.
+fn run_new_compiler_in_process(command string, prefs &pref.Preferences) {
+	if !macos_v3_driver_is_available() {
+		eprintln('`-new-compiler` requires a build that embeds the V3 compiler, which this one does not.')
+		exit(1)
+	}
+	raw_args := util.join_env_vflags_and_os_args()[1..]
+	if macos_v3_has_v1_only_leading_option(raw_args, command) || v3_has_v1_only_preferences(prefs)
+		|| (prefs.gc_set_by_flag && prefs.gc_mode != .no_gc) || (prefs.autofree && prefs.is_run) {
+		eprintln('`-new-compiler` cannot be combined with a V1-only option or mode; drop `-new-compiler` or the option.')
+		exit(1)
+	}
+	vexe := pref.vexe_path()
+	util.set_vroot_folder(os.dir(vexe))
+	os.setenv('VCHILD', 'true', true)
+	os.setenv('VEXE', os.real_path(vexe), true)
+	forwarded := macos_v3_forwarded_args(prefs, raw_args)
+	if prefs.is_verbose {
+		println('Running V3 compiler in process: ${util.args_quote_paths(forwarded)}')
+	}
+	macos_v3_driver_run(forwarded)
+	exit(0)
 }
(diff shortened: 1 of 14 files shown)


medvednikov:
@codex review

cmd/v/macos_v3_default.c.v
```diff
@@ -1,9 +1,42 @@
...
 
+import os
 import v.pref
+import v.util
 
+// On platforms where V3 is not the default compiler, `v` still runs the V1
+// compiler normally. `-new-compiler` opts into the embedded V3 driver and runs
+// it in THIS process, the same way V3 runs by default on macOS.
 fn maybe_delegate_to_macos_v3(command string, prefs &pref.Preferences) ?MacosV3CErrorReport {
+	if prefs.new_compiler {
```
medvednikov: **Gate this to compile invocations and honor `-old-compiler`.** This hook runs before external-tool and normal command dispatch in `cmd/v/v.v`, so on non-macOS `prefs.new_compiler` currently diverts every command into the V3 driver. For example, `v -new-compiler fmt ...` and `version` are treated as compiler inputs, `test` bypasses vtest, and `-old-compiler -new-compiler` still enters V3. The Darwin path already guards these cases. Please mirror its old-compiler precedence and relevant-command/unsupported-mode gating before calling `run_new_compiler_in_process`, and cover the non-macOS path with tests.

vlib/v3/gen/v/gen.v
```diff
@@ -0,0 +1,1744 @@
...
+// select_branch_header renders a select branch's guard: a receive binding
+// (`x := <-ch` / `x = <-ch`) or a plain send/expression condition.
+fn (mut g Gen) select_branch_header(value string, conds []flat.NodeId) {
+	if conds.len == 2 && (value == 'recv' || value == 'recv_assign'
+		|| value.starts_with('recv_compound')) {
+		g.expr(conds[0])
+		if value == 'recv' {
+			g.write(' := ')
+		} else {
+			g.write(' = ')
```
medvednikov, line 1175: **Preserve the compound receive operator.** The parser stores compound receive guards as `recv_compound:<op>` (for example, `recv_compound:+=`), but this branch emits `=` for every non-declaration receive. Formatting `select { x += <-ch { ... } }` therefore changes it to `x = <-ch`, altering program semantics. Extract and emit the suffix after `recv_compound:` and add a round-trip test for at least one compound receive.

medvednikov (commented):
I found two correctness issues in the changed code. CI status was intentionally ignored: this review is based on the diff and surrounding launcher/parser behavior. Please add focused regression coverage for the non-macOS dispatch path and compound `select` receives.

medvednikov (commented):
Re-reviewed head `1fa6be4` with CI intentionally ignored. The remaining routing issue is fixed: `crun` and `build-module` are rejected before the path-based compile heuristics, the Darwin and non-macOS gates share the same exclusion list, and focused regression assertions cover both commands. The compound `select` receive fix from the prior update remains correct. I found no further blocking issues in this re-review. GitHub does not allow approving one's own pull request, so this is submitted as a comment review.

Issue #27495, opened 2026-06-19
cgen, -g: `#line` directive emitted with no newline before an if/match-expr array-literal branch (`expected expression`)


medvednikov:
### Description

When compiling with `-g` (debug line info), cgen can emit a `#line` directive **glued directly to the following expression with no newline**, so the C preprocessor swallows the expression as trailing tokens of the `#line` directive. The assignment then has no value, and the C compiler fails.

It happens for an `if`/`match`-**expression** branch that returns an **array literal** (with call elements), under the default GC codegen path (which lowers the branch to statement form with `volatile` temps).

**V version:** V 0.5.1, current `master` (`415659729`)
**OS/Backend:** reproduced on macOS arm64 (clang) and Linux (gcc). Requires `-g`.

### Reproducer

```v
import os

fn pick() []string {
	home := os.home_dir()
	return match os.user_os() {
		'windows' {
			[os.join_path(os.getenv('LOCALAPPDATA'), 'a'), os.join_path(home, 'b')]
		}
		else {
			[]string{}
		}
	}
}

fn main() {
	println(pick())
}
```

Compile with:

```
v -g file.v
```

### Current behavior

The generated C contains (note no newline after the filename string):

```c
	_t3 =
#line 7 "/path/file.v"builtin__new_array_from_c_array(2, 2, sizeof(string), _MOV((string[2]){_t4, _t5}));
```

so the C compiler reports:

```
warning: extra tokens at end of #line directive [-Wextra-tokens]
error: expected expression
```

(GCC phrases it as `expected expression before '}' token`.)

### Expected behavior

The `#line` directive should be terminated by a newline before the expression, so the assignment's value is preserved.

### Notes

- Reported via the bugs.vlang.io crash reporter (~25 reports across `expected expression before '}'`/`'{'`). Original instance: `vab/android/sdk/sdk.v`.
- This is the same `#line`-emission family as the earlier `#line2520` (a `#line` directive emitted without a space) seen in other reports.
- Detect in generated C with: `v -g -o out.c -b c file.v` then `grep -nE '#line [0-9]+ "[^"]*"[^ ]' out.c`.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

PR #28129, opened 2026-08-19
HTTP/3 (QUIC) foundation: Phase 12 (HTTP/3 client wiring)

### LINKED ISSUE #2: Can you release the closed-source compiler right now for us to play?

quaesitor-scientiam:
## Summary

Phase 12 of the HTTP/3 (QUIC) implementation tracked at #27675 — the final
v1-scope phase. Phases 0-11 (merged: #27680, #27877, #27880, #27881,
#27882, #27884, #27885, #27980-ish quicconn, #28XXX h3-framing, #28103
QPACK) each built a self-contained, unit-tested building block, but
nothing composes any of them together and nothing connects them to the
existing `net.http` module. This phase does that wiring: `Request`/
`Response`/`Transport` gain an `enable_http3`-gated HTTP/3 client path.

Sub-scoped into 4 sequential commits (mirroring Phase 2's 2a/2b/2c and
Phase 9's 9a/9b precedent), each a hard dependency of the next, plus a
5th commit fixing everything a subsequent adversarial-review pass found —
**all implemented, tested, and reviewed**:

- **12a** — surgical additions to already-merged Phase 9 code
  (`conn.v`/`tls13_handshake.v`): a negotiated-ALPN accessor, a
  `peer_stream_opened` discovery event (careful about a `get_or_create`
  sibling-auto-creation trap under UDP reordering), a stream terminal-
  status query. No new files.
- **12b** — HTTP/3 + QPACK connection wiring, pure `module quic`: `H3Conn`
  driving the control stream, both QPACK streams (previously entirely
  unwired), the request-stream message-framing state machine (RFC 9114
  §4.1), and a blocked-HEADERS retry/re-queue mechanism.
- **12c** — UDP transport + `H3MuxConn` threading, `module http`: first
  UDP socket code and first background-thread-drives-a-`poll()`-state-
  machine code in the repo. `H3MuxConn` needs only ONE lock (unlike
  `H2MuxConn`'s wmu/fmu/smu split) since exactly one thread — its own
  driver — ever touches `H3Conn`/the transport; request threads queue and
  block on their own condition variable. `H3UdpTransport` decouples the
  driver from a concrete `net.UdpConn`, letting tests drive a real driver
  thread against an in-memory fake transport.
- **12d** — `Transport`/`Request`/`Response` integration: `enable_http3`
  opt-in (default `false`, no automatic h2/h1 fallback), a 3rd connection
  pool alongside `h1_idle`/`h2_conns`, `h3_client.v`/`transport_h3.v`
  mirroring the h2 conversion/dial machinery. `Version` gained a `v3_0`
  case so `resp.version()` reports something meaningful for an h3
  response.
- **Phase A fixup** (new) — a multi-agent adversarial-verification
  Workflow (5 independent finder lenses, each candidate finding checked
  by 2 independent skeptics) reviewed the combined 12a-12d diff and
  confirmed 7 real, cross-sub-phase bugs no single sub-phase's own
  `/vreview` pass could see; a same-session follow-up review found 2 more
  in the same code. All 9 fixed, tested, and re-reviewed — see the Test
  plan below and the commit message for the full list. Highlights: the
  very first HTTP/3 request against a freshly dialed connection could
  fail outright with no retry (making `enable_http3` effectively
  non-functional against real servers); RFC 9114 §5.2 GOAWAY draining was
  never implemented; RFC 9110 §15.2 1xx informational responses had no
  handling at all (a `103`-then-`200` sequence delivered the 103 as final
  status and misdelivered the real response as trailers).

Two scope decisions (documented in `PROGRESS.md`, open to reconsideration
during review): server push is permanently disabled for v1 (client never
sends MAX_PUSH_ID, so no push is ever authorized per §7.2.7 and any
PUSH_PROMISE/CANCEL_PUSH received is rejected `H3_ID_ERROR`); `enable_http3`
has no automatic h1/h2 fallback (UDP has no fast-fail signal, so racing
every `https://` request against h3 by default would regress the common
case — a real happy-eyeballs-style fallback is a separate follow-up).

**Known v1 limitation, found and documented during 12d's review**:
`net.quic`'s TLS 1.3 stack has no OS/default trust-store fallback at all
(unlike the h1/h2 `ssl.SSLConn` path) — `req.verify` is therefore
effectively **required** for HTTP/3 today; leaving it unset means every
h3 request fails certificate verification against every real server.
Documented prominently on `enable_http3`'s own doc comment, not left as a
silent surprise. `req.validate` (skip-verification) and `req.cert`/
`req.cert_key` (mutual TLS) are also not honorable on the h3 path in v1,
both documented there too.

## Test plan

- [x] 12a: `conn_test.v` fixture-handshake extensions (ALPN accessor,
      reordering regression test for stream discovery, terminal-status
      query) — `net.quic` suite green
- [x] 12b: `h3_conn_test.v`/`h3_control_stream.v`/
      `h3_qpack_streams.v`/`h3_request_stream.v` — fixture-driven,
      no socket — `net.quic` suite green (54/54, zero regressions)
- [x] 12c: `h3_mux_conn_test.v`/`h3_udp_dial_test.v` — a real driver
      thread against an in-memory `FakeH3UdpTransport` (not a real
      socket/reactive fake server — see the file's own doc comment for
      why: `conn_test.v`'s fixture-handshake bytes are private,
      test-file-only helpers with no cross-module export; a manual run
      against a real `quiche`/`ngtcp2` container is the recommended
      pre-merge check for full wire-level behavior). Directly regression-
      tests the sub-phase's top risk (a request queued by `do()` between
      driver-loop iterations must not be stranded if the connection dies
      in that window) with real threads/mutexes/cv — `net.http` suite
      green
- [x] 12d: `h3_client_test.v` (pure conversion tests)/
      `transport_h3_test.v` (pool-key folding, mTLS fast-fail, 3-way idle
      eviction with h3 entries) — no live h3 server exists in this repo
      (Phase 13 is out of scope), so the singleflight dial machinery
      itself and a full end-to-end `fetch(enable_http3: true)` round trip
      are not exercised here, matching 12c's own documented scope-down —
      `net.http` suite green (31 passed, 2 pre-existing skips, 0
      regressions)
- [x] Full `/vreview` pass before each commit (repeat-offender contract
      re-derivation for 12a's touches to already-merged `conn.v`/
      `tls13_handshake.v`, and for 12d's re-touch of `transport.v`/
      `transport_h2.v` from 12c) — run for 12a, 12b, 12c, and 12d. Found
      and fixed 2 concurrency/lifecycle bugs in 12c and 2 documentation/
      efficiency issues in 12d (see commit messages for detail)
- [x] Phase A adversarial-workflow verification pass before declaring the
      phase complete, particular attention to 12c's threading model —
      **done**: found 7 confirmed bugs (fresh-dial first-request race;
      GOAWAY draining absent; H3_REQUEST_REJECTED not retryable; orphaned-
      connection leak; self-terminated connection never leaving the pool;
      a peer-controlled error code colliding with the internal retryable
      sentinel; unbounded per-connection map growth), plus 2 more found in
      a same-session follow-up review (RFC 9110 §15.2 1xx informational
      responses unhandled; incomplete `:status` validation). All 9 fixed
      with new regression tests, `/vreview`'d, `net.quic` 54/54 and
      `net.http` 31/33 (2 pre-existing skips) green, zero regressions
- [x] Post-review fix: maintainer review (JalonSolov, 2026-08-21) flagged
      `dead_request_streams` (`h3_conn.v`) as unbounded on a long-lived
      pooled connection -- a 10th instance of the same growth-bug class as
      the Phase A finding above, on the field right next to the two it had
      already fixed. Phase R'd (regression test fails on the pre-fix
      commit, passes after): added `prune_terminal_dead_streams()`,
      dropping an entry once the underlying QUIC stream's receive side
      reaches its own terminal state (reset or fin received) -- RFC 9000
      guarantees no further frames for that ID can ever arrive at that
      point. `net.quic` 54/54 green. Commit `97dfbe75c3`
- [x] Post-review fix #2: JalonSolov's tool flagged `conn.v:707`'s
      `or { return }` as invalid V syntax -- turned out to be a real
      cgen bug (confirmed via CI check-run annotations across 4 jobs,
      not eyeballing), same `#line`-emission family as vlang/v#27495:
      an if/match-expression branch valued by an `or {}` block emits a
      malformed `#line` directive that swallows the following C
      expression. First exposed by this PR since it's the first to
      make `net.http` import `net.quic` at all -- the affected code
      (`conn.v`, `flow_control.v`) predates this PR by months and is
      unrelated to its own diff. Filed upstream: vlang/v#28163.
      Workaround: if/else statement instead of if-expression at both
      sites, semantics-preserving. `net.quic` 54/54 green. Commit
      `e7c5c9d7ed`
- [x] `./vnew fmt -w` + `./vnew check-md` on all touched files
- [x] Zero regressions in Phases 0-11's existing tests


quaesitor-scientiam:
Analysis of CI errors are not related to this PR @JalonSolov

JalonSolov:
Local AI Findings

**P1: Unbounded growth in dead request tracking**
In vlib/net/quic/h3_conn.v, `fail_request_stream()` records every failed stream in `dead_request_streams` and never removes it again. That set is consulted later in vlib/net/quic/h3_conn.v to skip already-failed streams.

This is the same lifecycle bug the PR is explicitly fixing for `request_streams` and `request_decoders`, but the fix was only partial: `dead_request_streams` is still a permanent “all time” history of every failed stream. Since QUIC stream IDs grow monotonically for the life of the connection, a long-lived pooled HTTP/3 connection can steadily accumulate an unbounded map of dead IDs, which is a real memory-growth issue and can turn into a performance problem under sustained load.

quaesitor-scientiam:
Confirmed and fixed in 97dfbe75c3: `prune_terminal_dead_streams()` drops a `dead_request_streams` entry once the underlying QUIC stream's receive side reaches a terminal state (reset or fin received) -- bounded to failed-streams-not-yet-drained instead of the connection's whole lifetime. New regression test added. Thanks for catching this.

JalonSolov:
On line 707 of `vlib/net/quic/conn.v`, your `or { return }` is causing problems, because the code is trying to assign a value to a variable, and `return` is not a valid value...

quaesitor-scientiam:
Thanks for flagging this! I took a close look and I'm not seeing the issue, though — line 707 itself is just `c.dcid = header.scid.clone()`, not an `or {}` block. The nearby ones are at lines 686 and 694 (`c.handshake_keys_server or { return }` and `unprotect_packet(...) or { return }`), and those look fine to me: the enclosing function (`process_initial_or_handshake`) returns bare `!` with no success value, so a bare `return` inside the `or {}` is valid V there -- it's the same pattern used throughout this file and it compiles/passes CI as-is.

Your message also got cut off mid-sentence ("...is not a valid value...") -- could you paste the rest, or point me at the exact line/commit your tool flagged? Want to make sure I'm not missing something it actually meant.

JalonSolov:
Starting on line 704, you have:
```v
	keys := if space == .initial {
		c.initial_keys_server
	} else {
		c.handshake_keys_server or { return }
	}
```
The problem is the `keys :=` assignment.  V expects all possible contortions of the `if` body to create a value which can be stored in `keys`.  Since `return` is a statement, not an expression, that's why the errors are "expression expected before `}`".

You can see this by clicking on the `Files changed` tab, looking for `conn.v` on the left... it has a red X in a circle to the right of if, then scroll down 'til you see all the boxes with errors reported by the CI... all of them on line 707, which is
```v
		c.handshake_keys_server or { return }
```

quaesitor-scientiam:
You were completely right, and I'm sorry for the runaround on my last reply — I looked at the wrong thing and dismissed this too fast. Once I actually pulled the CI check-run annotations (not just eyeballing the diff), the errors were right there, on all 4 jobs, exactly as you described.

Turns out it's a real V compiler bug, not a mistake in this PR's code: cgen emits a `#line` directive glued directly to the following line with no newline, for an if/match-expression branch whose value comes from an `or {}` block — the C preprocessor then swallows the expression as trailing tokens of the directive, and gcc fails with exactly the "expected expression before..." error you quoted. Same `#line`-emission bug family as vlang/v#27495 (which fixed a different triggering shape — array-literal branches — but not this one).

This is also why it only shows up now: this PR is the first one to make `net.http` import `net.quic` at all, so it's the first time `conn.v`/`flow_control.v` get compiled as part of the app-building jobs (gitly-windows, v-apps-compile, etc.) — the affected code itself predates this PR by months and has nothing to do with its own diff.

Filed upstream: vlang/v#28163. Worked around both sites in `e7c5c9d7ed` (an if/else statement instead of an if-expression — same behavior, sidesteps the buggy codegen path). Thanks for pushing back and not letting me wave it off.

quaesitor-scientiam:
Reviewed CI errors none caused by this PR

PR #28130, opened 2026-08-19
v3: fuse inlined C header typedef scans into a single pass

medvednikov:
## What

Two small, self-contained optimizations to the inlined-C-header declaration
scan in the v3 C backend (`gen/c/cleanc.v`):

- **Fuse the three per-kind typedef scans.** `collect_inlined_c_structs` ran
  three separate full-text passes — `c_typedef_struct_aliases`,
  `c_typedef_union_aliases`, `c_typedef_enum_aliases` — each scanning the whole
  header for `typedef struct` / `typedef union` / `typedef enum`. The new
  `c_typedef_all_aggregate_aliases` does one scan over `typedef ` positions,
  dispatches on the following keyword, and shares the identical tail logic. The
  output is an unordered set, so finding all three kinds in text order yields
  exactly the same alias names. (The individual helpers stay — they're still
  used by `c_source_defines_used_c_type`.)
- **Run-copy in `c_strip_comments`.** It copied non-comment bytes one at a time
  with `write_u8`; it now copies runs with `write_string`. Output is identical,
  including the preserved block-comment newlines that keep line counts stable
  for multi-line declaration scanning.

## Why

For large inlined system headers this scan is a measurable slice of the serial
cgen setup. On a large program it cuts the inlined-header scan from ~17 ms to
~13 ms.

## Correctness

- Output is byte-identical: verified by building a large real program and
  diffing its search output across many modes, plus a generics/map/closure
  program — no differences.
- New `gen/c/inlined_c_header_scan_test.v` locks the invariant directly: the
  fused scan equals the union of the three original scans over a matrix of
  typedef shapes (tagged, bodyless `typedef struct tag Alias;`, braced body,
  comma-separated declarators, `structFoo` near-misses), plus `c_text_matches_at`
  and the comment-stripping behavior.
- Full `v test vlib/v3/gen/c/` passes (15/15).

### DIFF
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -7519,23 +7519,42 @@ fn c_text_matches_at(text string, pos int, word string) bool {
 }
 
 // c_typedef_all_aggregate_aliases collects `typedef struct|union|enum` alias
-// names in a single scan. It replaces three separate full-text passes (one per
-// kind); every kind shares the identical tail logic below, and the result is a
-// set, so finding all three in text order yields the same names.
+// names in a single scan, replacing three separate full-text passes (one per
+// kind). The result is a set, so finding all three in text order yields the same
+// names as the three passes.
+//
+// It reproduces those passes exactly, including their failure behavior: each
+// per-kind scan `break`ed at its own first malformed candidate (an unbalanced
+// brace or missing `;`, e.g. from a `{` inside a comment in the raw header
+// text). So each kind here stops independently at its first malformed candidate
+// and is skipped thereafter. This keeps one kind's failure from suppressing the
+// others, and — crucially — never resumes *inside* a malformed candidate to pick
+// up a later same-kind typedef (such as a complete one inside the same comment)
+// that the original scan would already have stopped before.
 fn c_typedef_all_aggregate_aliases(text string) []string {
 	mut aliases := []string{}
 	mut start := 0
+	// kind codes: 0 = struct, 1 = union, 2 = enum.
+	mut stopped := [false, false, false]
 	for start < text.len {
 		idx := text.index_after('typedef ', start) or { break }
 		kw := idx + 'typedef '.len
 		mut prefix_len := 0
+		mut kind := 0
 		if c_text_matches_at(text, kw, 'struct') {
 			prefix_len = 'typedef struct'.len
+			kind = 0
 		} else if c_text_matches_at(text, kw, 'union') {
 			prefix_len = 'typedef union'.len
+			kind = 1
 		} else if c_text_matches_at(text, kw, 'enum') {
 			prefix_len = 'typedef enum'.len
+			kind = 2
 		} else {
+			start = kw
+			continue
+		}
+		if stopped[kind] {
 			start = kw
 			continue
 		}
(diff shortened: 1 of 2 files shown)


medvednikov:
@codex review

vlib/v3/gen/c/cleanc.v
```diff
@@ -7265,20 +7262,29 @@ fn c_strip_comments(text string) string {
...
 					i++
 				}
+				run_start = i
 				continue
 			}
 		}
-		sb.write_u8(c)
 		i++
 	}
+	if i > run_start {
```
medvednikov: The EOF flush must not run while `in_block` is still true. At block-comment entry the preceding run is written, but `run_start` is not advanced until a closing `*/` is found. Thus `c_strip_comments('int x; /* unfinished')` now returns `int x; int x; /* unfinished`, whereas the previous implementation returned `int x; `. This reintroduces comment text into declaration scanning. Please guard this flush with `!in_block` (or otherwise preserve the old EOF behavior) and add an unterminated-block regression test.

vlib/v3/gen/c/cleanc.v
```diff
@@ -7493,6 +7499,93 @@ fn c_typedef_plain_aliases(text string) []string {
...
+					start = semi_idx + 1
+					continue
+				}
+			}
+			start = pos + 1
+			continue
+		}
+		close_idx := c_matching_brace_end(text, pos)
+		if close_idx < 0 {
+			break
```
medvednikov: This `break` now terminates scanning for all three aggregate kinds, whereas previously a bad `struct` candidate only stopped the struct pass and the union/enum passes still ran. Because this scans raw header text, a valid header containing `/* example: typedef struct Broken { */` before `typedef union U { int x; } U;` makes the unmatched `{` in the comment drive `c_matching_brace_end` to `-1`, so `U` is no longer collected. That can trigger incorrect fallback/duplicate declarations. Please strip comments for this scan or advance past the failed candidate and continue, with a cross-kind regression test.

medvednikov (commented):
Changes required: two correctness regressions block approval. Both are independent of CI: the run-copy path changes EOF behavior for unterminated block comments, and the fused aggregate scan lets one false/malformed match suppress later declarations of every aggregate kind.

medvednikov:
Thanks for the careful review — both are real regressions from the fusion, fixed in 4fcc069daf.

**1 (High) — fused scan terminating on a bad candidate.** On an unbalanced brace (e.g. a `{` inside a comment in the raw header) or a missing terminating `;`, `c_typedef_all_aggregate_aliases` now skips the candidate and keeps scanning instead of `break`ing. This restores the behavior of the three independent scans, where a malformed `struct` candidate can't suppress later `union`/`enum` aliases. Verified on your exact example — `U` is now collected, and the fused result equals the union of the three separate scans for it.

**2 (Medium) — unterminated block comment re-appended.** The final run flush is now guarded with `!in_block`, so a comment that runs to EOF is discarded rather than flushed back (`run_start` only advances past a closing `*/`). `c_strip_comments('int x; /* unfinished')` again returns `int x; `.

Added regression tests for both (a valid union after a comment brace; `c_strip_comments` on an unterminated comment). Locally: full `v test vlib/v3/gen/c/` passes 15/15, and the large-program build output is still byte-identical. CI will run on the push.

medvednikov (commented):
One correctness blocker remains (CI not considered).

The unterminated-block-comment fix is correct, and the new unbalanced-brace test covers the original `close_idx < 0` failure. However, a false bodyless match can still consume a later aggregate of another kind:

```c
/* typedef struct Broken */
typedef union UTag { int x; } UAlias;
```

The fused scan first matches the comment's `typedef struct Broken`. Because the token after `Broken` is `*/`, it enters the bodyless path and searches for the next `;`. That semicolon is the union member's `int x;`; the scanner records `x` and advances `start` past it. The real `typedef union` began before that new start, so `UAlias` is never scanned.

The old independent scans return `x` from the struct pass plus `UAlias` from the union pass; the fused scan returns only `x`. This is a valid header, and `collect_inlined_c_structs` only records the tag `UTag` in its line scan, so the alias `UAlias` can be lost entirely.

Please avoid advancing past a later `typedef` candidate of another kind (or strip comments before aggregate scanning), and add this case to the fused-vs-separate regression test.

medvednikov:
@codex review

vlib/v3/gen/c/cleanc.v
```diff
@@ -7493,6 +7503,100 @@ fn c_typedef_plain_aliases(text string) []string {
...
+			continue
+		}
+		close_idx := c_matching_brace_end(text, pos)
+		if close_idx < 0 {
+			// Unbalanced brace — e.g. a `{` inside a comment in the raw header
+			// text. Skip this candidate and keep scanning for the other
+			// aggregates instead of terminating the whole fused pass: the three
+			// separate scans this replaces were independent, so a malformed
+			// struct candidate must not suppress later union/enum aliases.
+			start = pos + 1
```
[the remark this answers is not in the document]
medvednikov: Fixed in 3b98c35ce3 — good catch, resuming at `pos+1` could re-enter the malformed candidate.

It now reproduces the three independent per-kind scans exactly: each kind stops at its own first malformed candidate (unbalanced brace / missing `;`) and is skipped thereafter, advancing only past the current `typedef ` keyword so the other kinds keep scanning. So a bad `struct` still can't suppress later `union`/`enum` aliases, but for your example the struct scan stops at `Broken` and never reaches `Hidden`, so `Hidden` is not collected.

Added a regression test for a complete `typedef struct Hidden {...} Hidden;` nested inside an unbalanced-brace comment. Full `v test vlib/v3/gen/c/` passes 15/15 and the large-program build output is unchanged.

PR #28150, opened 2026-08-21
thirdparty: fix tccbin contract bootstrap

GGRei:
## Summary

This follow-up fixes the tccbin automation contract CI introduced by #28137.

The contract worker now checks out a known-good immutable `vlang/tccbin` Linux bundle before running `make local=1`, and verifies that the checkout is exact, detached, clean, and contains a working `tcc.exe` and `libgc.a`.

It also removes the five repeated V compiler notices emitted by the tccbin automation modules.

## Why

`make local=1` correctly preserved the pinned VC snapshot, but it also disabled normal TCC provisioning. On a clean runner, the contract job therefore failed before the test suite because `thirdparty/tcc/lib/libgc.a` was missing.


GGRei:
Ready Alex. Thanks!

PR #28158, opened 2026-08-22
v3: fix -autofree on literal-only programs

kestred:
`v -autofree` fails to build any program whose body is only literal print calls, including `examples/hello_world.v`, with `use of undeclared identifier 'MessageError__free'`.
markused's trivial-literal fast path skipped the ownership destructor seeds while cgen still emitted the `_result` destructor that calls them, so this roots the IError implementers' destructors beside the IError sentinels that are already seeded unconditionally.


kestred:
Superseded by #28159, which is based on the feature branch rather than my fork's main.

PR #28159, opened 2026-08-22
v3: fix -autofree codegen

kestred:
Fixes `v -autofree`, which could not build most programs. The generated C referred to variables that were out of scope, called interface methods by a name autofree never defines, and emitted destructors for functions that had already been removed.

One of those fixes also corrects two places that skipped an ownership scope drop slot for every block rather than only the synthesised ones, which could free a value twice or run the wrong destructor.


medvednikov:
@codex review

medvednikov:
Thanks! :)

PR #28184, opened 2026-08-26
v3: add regression tests for for-in-by-ref element binding

### LINKED ISSUE #28084: fixed array "for in" compilation error

quaesitor-scientiam:
## Summary

Follow-up to #28087 (which fixed #28084's for-in-by-ref half: `for x in
&arr` was binding elements by value instead of by reference). #28087 merged
without committed regression tests, per @medvednikov's review comment on
that PR. This adds them, with no code changes - just test coverage for the
already-shipped fix.

Covers the specific cases requested in review:
- a local fixed array (`&` rejected outside `unsafe`, same as any other `&`
  of a local fixed array - a borrow-safety check, not something this fix
  touches)
- a fixed array reached through a struct field (the exact shape of the
  #28084 report)
- an ordinary dynamic array, as a regression guard against the by-value
  bug reappearing
- `for mut item in arr` as a binding mode distinct from `for item in &arr`
- a `mut` array parameter's ABI pointer-shape not being mistaken for an
  explicit `&` in the for-in line (the false-positive risk the fix had to
  avoid)
- the two-variable indexed form (`for idx, item in &arr`) binding the
  value variable, not the index, by reference
- a generic function/method instantiation

## Test plan

- [x] `./vnew test vlib/v3/tests/for_in_ref_binding_test.v` - all 7 tests
  pass, run 3x for consistency
- [x] `./vnew fmt -diff` clean

### DIFF
--- a/vlib/v3/tests/for_in_ref_binding_test.v
+++ b/vlib/v3/tests/for_in_ref_binding_test.v
@@ -1,0 +1,136 @@
+import os
+import rand
+
+// Regression coverage for #28084: `for x in &arr` must bind elements by
+// reference, not by value. The checker's `for_in_iterable_yields_ref` used to
+// be a dead stub that always returned false, so `annotate_for_in` never
+// recorded the loop variable as pointer-typed for this case and mutations
+// through it were silently lost.
+
+const for_in_ref_binding_vexe = @VEXE
+const for_in_ref_binding_tests_dir = os.dir(@FILE)
+const for_in_ref_binding_v3_dir = os.dir(for_in_ref_binding_tests_dir)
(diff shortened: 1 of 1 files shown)


medvednikov:
@codex review

PR #28196, opened 2026-08-26
v3: use target width for pointer map callbacks

medvednikov:
## Summary

- select V3 map hash/equality/clone callbacks for pointer-sized keys from the target pointer width
- propagate the target into parallel C generator workers
- cover pointer-sized callback selection on 32-bit and 64-bit targets

This fixes the s390x V3 failure in `vlib/builtin/map_test.v:test_voidptr_keys`, where a 64-bit `voidptr` key incorrectly used 4-byte callbacks.

## Tests

- `./vnew -silent test vlib/v3/gen/c/map_test.v vlib/v3/gen/c/parallel_worker_test.v`
- `./vnew -silent test vlib/v3/gen/c/`
- built standalone V3 and ran `vlib/builtin/map_test.v`
- generated x86 and s390x C and verified all `voidptr` map callbacks use the target width


medvednikov:
Superseded by the consolidated CI repair batch in #28197 so all currently failing V3 gates can validate together.

Issue #28193, opened 2026-08-26
v3: project importing toml fails to build with -ownership


plasticgaming99:
### Describe the bug

Trying to build project that using toml library, cc reported error.

### Reproduction Steps

```v
module main

import toml

struct CommonStruct {
	field string
}

fn main() {
	d := toml.parse_text('field = "some"')!
	t := d.decode[CommonStruct]()!

	println(t)
}
```

### Expected Behavior

No errors

### Current Behavior

```sh
v . -ownership -no-memory-limit -cc clang
  [ v3 benchmark lines --- omitted ]
C compilation failed:
src.c:14387:15: error: assigning to 'token__Token' (aka 'struct token__Token') from incompatible type 'int'
 14387 |                 p->peek_tok = 0;
       |                             ^ ~
src.c:14403:15: error: assigning to 'token__Token' (aka 'struct token__Token') from incompatible type 'int'
 14403 |                 p->peek_tok = 0;
       |                             ^ ~
2 errors generated.
```

### V version

V 0.5.2 711c3f2

### Environment details (OS name and version, etc.)

|V full version      |V 0.5.2 8ef799c99cb0141a119e9e06f6b629c9e98486bf.711c3f2
|:-------------------|:-------------------
|OS                  |linux, "CachyOS"
|Processor           |16 cpus, 64bit, little endian, AMD Ryzen 7 3800X 8-Core Processor
|Memory              |1.5GB/31.26GB
|                    |
|V executable        |/home/plastek/.v/v
|V last modified time|2026-08-26 14:04:38
|                    |
|V home dir          |OK, value: /home/plastek/.v
|VMODULES            |OK, value: /home/plastek/.vmodules
|VTMP                |OK, value: /tmp/v_1000
|Current working dir |OK, value: /home/plastek/plas/vtest
|                    |
|Git version         |git version 2.55.0
|V git status        |weekly.2025.51-3177-g711c3f28
|.git/config present |true
|                    |
|cc version          |cc (GCC) 16.2.1 20260810
|gcc version         |gcc (GCC) 16.2.1 20260810
|clang version       |clang version 22.1.8
|tcc version         |tcc version 0.9.28rc 2026-07-24 HEAD@85ba3ae8 (x86_64 Linux)
|tcc git status      |thirdparty-linux-amd64 d6e7ac1b
|emcc version        |N/A
|glibc version       |ldd (GNU libc) 2.44
> [!NOTE]
[...]

PR #28224, opened 2026-08-28
toml: fix -ownership build failure in parser next()

medvednikov:
### What

Fixes the C compilation failure reported in #28193 when building a project that imports `toml` with `-ownership`:

```
error: assigning to 'token__Token' (aka 'struct token__Token') from incompatible type 'int'
 14387 |   p->peek_tok = 0;
```

### Why it happens

`toml.parser.Parser.next()` advances the token stream with:

```v
p.peek_tok = p.tokens.first()
p.tokens.delete(0)
```

Under `-ownership`, `token.Token` is a *move-only* struct — it owns its `lit` string and does not implement `IClone` — so `[]token.Token.first()` cannot produce an independent copy of the element. For code in the **top** module the ownership checker rejects this (`cannot return an independent array element: token.Token requires ownership destruction but has no clone() method`), but for an **imported** module (like `toml`) the check does not fire, and cgen falls through to the "invalid program" placeholder, emitting `p->peek_tok = 0;` — an `int` assigned to a struct, which the C compiler rejects.

### The fix

`first()` immediately followed by `delete(0)` is a *pop-front*: read the element, then drop the slot. Indexing `p.tokens[0]` expresses exactly that, and in ownership mode indexing **moves** the element out (no clone required), so `token.Token` no longer needs `IClone`. In the default backend `p.tokens[0]` is an ordinary copy, identical in effect to `first()`, so behaviour is unchanged there.

- `vlib/toml/tests/toml_test.v` (the full toml-lang/toml-test suite) passes.
- `vlib/toml/tests/reflect_test.v` (exercised by the v3 suite) passes.

### Note — a separate, deeper `-ownership` issue remains

This PR only removes the **compile-time** error. While diagnosing it I found that `toml`'s parser still mis-runs under `-ownership` because of a distinct, more fundamental ownership-analysis bug: reading a heap-owning struct field into a local **aliases** the field instead of copying it, so the value dangles once the field is later reassigned/moved. Minimal repro (prints garbage under `-ownership`, correct otherwise):

```v
struct Holder {
mut:
	name string
}

fn build(a string, b string) string {
	mut s := a
	s += b
	return s
}

fn run() string {
	mut h := Holder{ name: build('XX', 'YY') }
	n := h.name          // n aliases h.name instead of copying it
	h.name = build('AA', 'BB') // frees the old 'XXYY' that n still points at
	return n             // -> use-after-free ("reas"/garbage) under -ownership
}

fn main() {
	println(run()) // expected: XXYY
}
```

In `toml` this surfaces via the token rotation in `next()` freeing the strings that the parsed keys/values still reference, so `parse_text('field = "some"')` currently decodes to an empty value under `-ownership`. That is a compiler-side fix (clone/keep-alive on field read) and is out of scope for this build-error fix; happy to file it as a separate issue.

Related: #28193

### DIFF
--- a/vlib/toml/parser/parser.v
+++ b/vlib/toml/parser/parser.v
@@ -149,12 +149,15 @@ fn (mut p Parser) next() ! {
 	p.prev_tok = p.tok
 	p.tok = p.peek_tok
 	if p.tokens.len > 0 {
-		p.peek_tok = p.tokens.first()
+		// Take the front token out and drop the emptied slot. Indexing (instead of
+		// `first()`) moves the element in `-ownership` mode, so `token.Token` needs no
+		// clone; in the default backend it is an ordinary copy followed by a delete.
+		p.peek_tok = p.tokens[0]
 		p.tokens.delete(0)
 		p.peek(1)!
 	} else {
 		p.peek(1)!
-		p.peek_tok = p.tokens.first()
+		p.peek_tok = p.tokens[0]
 		p.tokens.delete(0)
 	}
 }


medvednikov:
@codex review

medvednikov:
Superseded by #28244, which fixes this entirely in the v3 compiler (value-copy of owned reads under `-ownership`) with **no `toml` change** — the structural-clone relaxation makes the original `p.tokens.first()` build and run correctly, so the parser workaround here is no longer needed. Closing in favor of #28244.

Issue #28226, opened 2026-08-28
v fmt silently strips 'mut' from 'if mut x := map[key] { ... }'


quaesitor-scientiam:
## Describe the bug

`v fmt -w` (and `v fmt -diff`) silently rewrites `if mut x := map[key] { ... }`
to `if x := map[key] { ... }` — dropping the `mut` qualifier. If the block body
then mutates a field through `x` (e.g. `x.field += n`), the reformatted file
fails to compile ("`x` is immutable, declare it with `mut` to make it
mutable"), or — worse, if the body doesn't happen to mutate anything directly
visible to the checker in a way that trips an error — could silently change
behavior.

In my case this surfaced while formatting an existing, working, compiling
file: `v fmt -w` on `vlib/net/http/h2_server.v` (from a fresh `vlang/v`
`master` checkout, commit `51bda9911f`) stripped `mut` from three separate
`if mut s := c.streams[...]`/`if mut existing := c.streams[...]` sites, and
one of them (`if mut existing := c.streams[frame.stream_id] { existing.trailer_block = []; ... }`)
immediately failed to compile afterward.

The same `fmt` pass also reordered/misplaced comments that sat directly above
an `... or { ... }` block, moving them to before the whole statement instead
of staying attached to the `or {}` handler — e.g. turning

```v
frame := c.read_frame() or {
    // Treat a clean transport close as end of session.
    return
}
```

into

```v
frame := 
// Treat a clean transport close as end of session.
c.read_frame() or {
    return
}
```

which is at best confusing (the comment no longer reads as being about the
`or` branch) though not itself a compile error.

## Reproduction

Minimal, no unrelated context — a struct holding a `map[u32]&Item` and a
method that bumps a field through a `mut`-qualified map-lookup pattern:

```v
module main

struct Item {
mut:
	send_window i64
}

struct Holder {
mut:
	items map[u32]&Item
}

fn (mut h Holder) bump(id u32, inc i64) {
	if mut it := h.items[id] {
		it.send_window += inc
	}
}

fn main() {
	mut h := Holder{}
	h.bump(1, 10)
}
```

Run:

```
v fmt -diff repro.v
```

Output:

```diff
 fn (mut h Holder) bump(id u32, inc i64) {
-	if mut it := h.items[id] {
+	if it := h.items[id] {
 		it.send_window += inc
 	}
 }
```

The `mut` is dropped. In this minimal example the body still compiles (V
apparently doesn't always flag the resulting immutability at this call
[...]

PR #28227, opened 2026-08-28
net.http: h2_server.v CONNECT pseudo-header validation (RFC 9113 §8.5)

quaesitor-scientiam:
## Summary

`h2_validate_request_pseudo` (`h2_server.v`) required `:method`/`:path`/`:scheme`
unconditionally, so every conforming CONNECT request — RFC 9113 §8.5 requires a
CONNECT request to **omit** `:scheme` and `:path`, and instead carry `:authority`
— was rejected as malformed (400) before ever reaching the `Handler`.

This is the same bug independently found and fixed in `h3_server.v`'s
`h3_validate_request_pseudo` (Codex review, [PR #28164](https://github.com/vlang/v/pull/28164),
[pullrequestreview-5044139767](https://github.com/vlang/v/pull/28164#pullrequestreview-5044139767)).
That fix's own commit message noted `h2_server.v` had the identical gap, left as
an explicitly out-of-scope follow-up — this PR closes it.

## Fix

Mirrors the `h3_server.v` reference exactly: track `:method` during validation;
when it's `CONNECT`, require `:authority` (non-empty) and forbid `:scheme`/`:path`;
otherwise keep the existing `:method`/`:path`/`:scheme` requirement. Extended
CONNECT (RFC 8441's `:protocol`, WebSockets-over-HTTP/2) stays out of scope, same
as the H3 side — this only stops a conforming plain CONNECT from being rejected,
it doesn't implement CONNECT tunneling itself.

**A second, related gap found and fixed while reviewing this same diff
full-file:** the `:authority` match arm only ever checked for duplication, never
emptiness — unlike `:method`/`:path`/`:scheme`, which all reject an empty value
inline. A CONNECT request with a present-but-empty `:authority` would have
slipped past the new check, violating RFC 9113 §8.5's "MUST be provided...
contains the host and port to connect to". Fixed by capturing the value and
checking it specifically in the CONNECT branch — the ordinary-request arm is
left alone, since RFC 9113 doesn't mandate a non-empty `:authority` there.

## Test plan

- [x] Regression tests mirroring `h3_server_pseudo_test.v`'s shape (new file,
      `h2_server_pseudo_test.v`): accepts a conforming CONNECT, rejects
      CONNECT+`:scheme`, rejects CONNECT+`:path`, rejects CONNECT missing
      `:authority`, rejects CONNECT with an empty `:authority`, and an
      end-to-end `build_request` test asserting `.connect` method with an
      empty `url` (there is no `:path` to populate it from)
- [x] Both fixes Phase-R'd: each regression test confirmed failing against the
      pre-fix code, then passing after
- [x] `./vnew -no-memory-limit -silent test vlib/net/http/` green (30 passed,
      4 pre-existing platform skips, 0 failed)
- [x] `/vreview high-risk` — full-file read of `h2_server.v` (1192 lines); 1
      finding (the empty-`:authority` gap above), fixed

## Note: an unrelated `v fmt` bug hit along the way

While formatting this change I found `v fmt -w` silently strips `mut` from
`if mut x := map[...] { ... }` and reorders comments around `or {}` blocks —
filed separately as [#28226](https://github.com/vlang/v/issues/28226), with a
minimal repro. I reverted its output on this file and hand-applied only this
PR's intended change instead of trusting `fmt -w` on it.

## Not done here, flagged rather than silently skipped

The H2 conformance matrix (a maintained local reviewer artifact, not part of
this repo) should get a CONFORMS row for RFC 9113 §8.5 here, mirroring the
QUIC matrix's own update for the H3 fix — not done in this PR.

### DIFF
--- a/vlib/net/http/h2_server.v
+++ b/vlib/net/http/h2_server.v
@@ -88,15 +88,34 @@ fn h2_request_field_error(name string, value string) string {
 
 // h2_validate_request_pseudo enforces the RFC 9113 §8.1.2/§8.3 rules a request
 // header list must satisfy: only the request pseudo-headers, each at most once,
-// all appearing before any regular field, with :method, :path and :scheme
-// present; every regular field must pass h2_request_field_error. A non-ok return
-// means the request is malformed (a stream error at the call site).
+// all appearing before any regular field; every regular field must pass
+// h2_request_field_error. A non-ok return means the request is malformed (a
+// stream error at the call site).
+//
+// Mandatory-pseudo-header shape depends on :method (RFC 9113 §8.5, "The
+// CONNECT Method"): an ORDINARY request needs :method/:path/:scheme, but a
+// CONNECT request "MUST omit" :scheme and :path entirely and instead needs
+// :authority ("contains the host and port to connect to"). "A CONNECT
+// request that does not conform to these restrictions is malformed."
+// Extended CONNECT (RFC 8441's `:protocol`, e.g. WebSockets-over-HTTP/2)
+// is NOT handled here -- a documented v1 scope limit, not an oversight:
+// this function only avoids REJECTING a conforming plain CONNECT's
+// pseudo-header shape, it doesn't implement CONNECT tunneling itself
+// (h2_server.v has no Handler-visible CONNECT semantics yet either).
+// Missed in an earlier version: EVERY request, CONNECT included, was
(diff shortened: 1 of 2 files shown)


quaesitor-scientiam:
Reviewed CI errors none are related to this PR. @JalonSolov if no further code reviews can we please merge this.

JalonSolov:
Local AI said:  No actionable findings in PR #28227.

PR #28228, opened 2026-08-28
cgen: pop cur_indexexpr after assigning a function literal to a map/array index

kbkpbot:
Assigning an anonymous function to a map/array index (`m[k] = fn (...) {...}`) pushed an entry onto g.cur_indexexpr in index.v, but the ast.AnonFn RHS branch in assign_stmt ended with `continue`, bypassing the pop that the other RHS kinds perform at the end of the statement. The leaked entry made every later non-scalar map/array assignment in the same compilation unit skip its `= ` operator, producing invalid C like
`(*(T*)builtin__map_get_and_set(...))(value)` instead of `... = value`.

Pop the entry (and recompute is_arraymap_set) in that branch, matching the pop logic used for the other RHS kinds.

Add a regression test (vlib/v/tests/anon_fn_map_assign_test.v) that fails to compile on the unfixed backend.

Fixes a bug that broke compilation of modules (e.g. [vxui](https://github.com/kbkpbot/vxui)) which assign function literals to maps and then do struct-value map assignments.


medvednikov:
@codex review

PR #28230, opened 2026-08-28
db.pg: fix garbled connect error message by deep-copying libpq error before PQfinish (#28225)

### LINKED ISSUE #28225: Garbled PostgreSQL Error Message
### Describe the bug

If PostgreSQL isn't running when the program starts it produces an error, which is great. But for some reason parts of the error string are garbled. This affects both the v2 and v3 compilers.

### Reproduction Steps

Attempt to connect to any PostgreSQL instance which does not exist. In this case I was using the same code as #28210 

### Expected Behavior

A well-formed error message.

### Current Behavior

```
V panic: result not set (Connection to a PG database failed: �5]]^�D�&Urver at "localhost" (::1), port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
connection to server at "localhost" (127.0.0.1), port 5432 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
```

### Possible Solution

There's probably some pointer math gone wrong here as the error message prints binary garbage and then repeats part of the message twice. I'm betting the error is coming from `libpq` and something about the response from `libpq` is being misinterpreted in this particular edge case.

In case it is relevant here is information about the `libpq` version I'm on:

```
postgresql-libpq, version: 18.4, release: 66
```

### V version

V 0.5.2 51bda99

|V full version      |V 0.5.2 915f2e7556e670425038df557071fb80ac01f1a9.51bda99
|OS                  |linux, Solus 4.9 Serenity
|Processor           |32 cpus, 64bit, little endian, AMD Ryzen 9 5950X 16-Core Processor
|Memory              |14.37GB/125.7GB
|                    |
|V executable        |/opt/vlang/v
|V last modified time|2026-08-28 02:26:39
|                    |
|V home dir          |OK, value: /opt/vlang/v
|VMODULES            |OK, value: /home/user/.vmodules
|Current working dir |OK, value: /home/user/source/repotest
|                    |
|V git status        |0.5.2-397-g51bda991
|.git/config present |true
|                    |
|cc version          |cc (Solus) 15.2.1 20260604
|gcc version         |gcc (Solus) 15.2.1 20260604
|clang version       |Homebrew clang version 22.1.8
|emcc version        |N/A
|glibc version       |ldd (Solus) 2.43
> [!NOTE]
>

waterWang:
Fixes #28225

## Root cause

`PQerrorMessage` returns a pointer to **libpq's internal error buffer**, and V's `&char.vstring()` only does a *zero-copy wrap* (`.str` points at the C buffer, no ownership transfer). In `connect_slot`, the message was interpolated **after** `C.PQfinish(conn)` had already freed that buffer, so the error string became a dangling pointer - the reported garbled / null-byte output.

Note that `error_msg := '${c_error_msg}'` is **not** a deep copy - the compiler optimizes an interpolation of a single string variable into a plain assignment, so the original code's intent ("force the construction of a new string") was defeated.

## Fix

Take an explicit `.clone()` of the wrapped string **before** `PQfinish` destroys the handle:

```v
c_error_msg := unsafe { C.PQerrorMessage(conn).vstring().clone() }
C.PQfinish(conn)
return error('Connection to a PG database failed: ${c_error_msg}')
```

## Test

Added `pg_error_test.v` - connects to a non-running PostgreSQL port (no live DB required) and asserts the error message contains readable text and **no** embedded NUL bytes. Verified:
- Without the fix: test FAILS with the garbled message (exactly #28225's symptom).
- With the fix: all 7 `vlib/db/pg` tests pass.

### DIFF
--- a/vlib/db/pg/pg.c.v
+++ b/vlib/db/pg/pg.c.v
@@ -386,12 +386,16 @@ fn connect_slot(conninfo string) !IdleSlot {
 	}
 	status := unsafe { ConnStatusType(C.PQstatus(conn)) }
 	if status != .ok {
-		// We force the construction of a new string as the
-		// error message will be freed by the next `PQfinish` call
-		c_error_msg := unsafe { C.PQerrorMessage(conn).vstring() }
-		error_msg := '${c_error_msg}'
+		// `PQerrorMessage` returns a pointer to libpq's internal buffer, and
+		// `vstring()` only does a zero-copy wrap (no ownership transfer). The
+		// buffer is freed by the `PQfinish` call below, so we must take a
+		// deep copy now, before the handle is destroyed, otherwise the error
+		// message is a dangling pointer (garbled / null bytes in the output).
+		// Note: a plain `'${msg}'` interpolation is optimized by the compiler
+		// to a no-op assignment, so an explicit `.clone()` is required.
+		c_error_msg := unsafe { C.PQerrorMessage(conn).vstring().clone() }
 		C.PQfinish(conn)
-		return error('Connection to a PG database failed: ${error_msg}')
+		return error('Connection to a PG database failed: ${c_error_msg}')
 	}
 	return IdleSlot{
 		handle:     conn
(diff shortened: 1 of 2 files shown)


JalonSolov:
https://modules.vlang.io/builtin.html#cstring_to_vstring

So the fix would be simpler as
```v
c_error_msg := unsafe {cstring_to_vstring(C.PQerrorMessage(conn))}
C.PQfinish(conn)
return error('Connection to a PG database failed: ${c_error_msg}')
```

medvednikov:
@codex review

PR #28237, opened 2026-08-28
v3: bound native-input scan RSS so cold builds don't hit the memory ceiling

quaesitor-scientiam:
## Problem

On a **cold** V3 module cache, the `check` stage's native-input scan drives peak RSS to ~9 GB when building a project with a large combined native surface (OpenSSL via `crypto.ecdsa`, Cocoa/AppKit, and `sokol`/`stb`/`mbedtls` single-header C libs together). That trips v3's 10 GiB memory ceiling and forces a silent fallback to the V1 compiler. Warm builds are unaffected (~750 MB), which is why it's intermittent.

Root cause is two transient allocations in the native-input path, both retained for the whole stage because v3 `-prealloc` builds run `-gc none` — nothing is freed until the enclosing stage scope exits:

### 1. `c_typedef_is_function_pointer` (dominant, ~8 GB)

```v
relative := source[offset..].index(name) or { return false }
```

`string[offset..]` allocates a fresh copy of the entire remaining tail of `source` on **every loop iteration**. This runs once per typedef name per native header, so large headers (`sokol_gfx.h`, `stb_truetype.h`, mbedtls, openssl) × hundreds of typedef names accumulate gigabytes of tail copies. Replaced with `source.index_after(name, offset)`, which scans in place — same first-match semantics, no allocation.

### 2. `cache_native_inputs_language` (~2.2 GB)

For every header `#include` in the AST it recursively inlined the header's entire transitive `#include` closure into a builder — once per include site, no cross-site dedup, no size cap — purely to decide whether the native inputs need Objective-C. Now: memoize the result per include argument (`<...>` keyed on the argument, quoted keyed on including-file + argument), stop once Objective-C is known, and run each remaining inline inside a disposable prealloc scope so its closure is freed as soon as the boolean is read.

### 3. `c_collect_external_input_tree` (minor, ~200 MB)

It built a second full-size line-buffer copy of each file and held it across the nested-include recursion. Now it collects cheap string refs during the scan and does the one byte-copy after the recursion returns.

## Result

Cold `check` stage peak on the affected project: **~9058 MB → ~980 MB**; the whole cold build now peaks at cgen (~1.9 GB), well under the ceiling. Warm builds unchanged.

**No behavior change:** the scan produces the same native-input set, the same C-language classification, and the same static-storage flags — only less transient memory.

## Tests

- `vlib/v3/gen/c/target_test.v` and `vlib/v3/modulecache/` pass.
- Added `test_cache_native_inputs_language_header_branch_and_memo_key` covering the previously-untested header-include branch, including a regression guard for the quoted-include memo key (two same-named headers in different module dirs with different Objective-C-ness must not share a cache entry — verified this assertion fails if the key's including-file component is dropped).

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -2681,8 +2681,14 @@ fn register_headerless_c_types(mut tc types.TypeChecker) {
 fn c_typedef_is_function_pointer(source string, name string) bool {
 	mut offset := 0
 	for offset < source.len {
-		relative := source[offset..].index(name) or { return false }
-		start := offset + relative
+		// index_after scans in place from `offset`. `source[offset..].index(name)`
+		// allocated a fresh copy of the whole remaining tail of `source` on every
+		// iteration; this function runs once per typedef name per native header, so
+		// under -prealloc (allocations in a stage scope are not freed until the
+		// scope ends) those tail copies accumulated to multiple GB of transient
+		// RSS on a build pulling large native headers (sokol, stb, mbedtls,
+		// openssl) — enough to trip v3's memory ceiling and fall back to V1.
+		start := source.index_after(name, offset) or { return false }
 		end := start + name.len
 		if (start == 0 || (!source[start - 1].is_alnum() && source[start - 1] != `_`))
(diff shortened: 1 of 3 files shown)


medvednikov:
@codex review

PR #28250, opened 2026-08-29
json2: reduce generated struct field code

medvednikov:
## Summary

- move runtime JSON struct-field matching out of the comptime field expansion
- share encoder key collision and selection logic across generated struct fields
- teach V3 C generation to preserve `@[noinline]`, including generic specializations and parallel workers

The previous `$for field in T.fields` paths emitted the same runtime key matching and collision logic once per field for every concrete struct type. Large structs consequently produced very large decoder and encoder functions.

## Size impact

Measured with an arm64 macOS FuseCode production build using `-prod -gc none`:

- executable: 3,915,096 -> 3,715,400 bytes (-199,696, 5.10%)
- Mach-O `__text`: 3,224,188 -> 3,025,320 bytes (-198,868, 6.17%)
- generated `JsonTheme` decoder: 56,548 -> 26,492 bytes
- generated `JsonTheme` encoder: 21,152 -> 8,564 bytes

A controlled V3 fixture containing a FuseCode-shaped JSON theme struct went from 173,456 to 141,072 bytes, while `__text` went from 110,336 to 80,400 bytes.

## Testing

- rebuilt V with `./v -g -keepc -o ./vnew cmd/v`
- JSON2 test suite: 74 passed
- V3 C generator unit suite: 17 passed
- compiler error/parser suite: 1,619 passed, 1 skipped
- C output regression suite passed
- verified generated generic C functions contain `__attribute__((noinline))`
- built and inspected the FuseCode production binary with `nm`, `size`, `file`, and `otool`

### DIFF
--- a/vlib/json2/decode.v
+++ b/vlib/json2/decode.v
@@ -417,6 +417,25 @@ fn mark_struct_field_decoded(decoded_mask u64, mut decoded_fields []bool, field_
 	return decoded_mask
 }
 
+// find_struct_field centralizes the runtime part of struct key matching. Keeping
+// this loop outside the comptime field loop avoids emitting the same skip,
+// omitempty, length, and memory-comparison checks once for every struct field.
+@[noinline]
+fn (decoder &Decoder) find_struct_field(field_infos []StructFieldInfo, key_ptr voidptr, key_len int) int {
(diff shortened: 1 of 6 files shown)


medvednikov:
@codex review

PR #28255, opened 2026-08-29
v3: address FastC review feedback

medvednikov:
Follow-up to #28254 for review findings posted after that PR merged.

This change:
- rejects function literals and closures instead of emitting behavior-changing stubs
- preserves nested generic call rewrites in emitted concrete instances
- specializes generic parameters through composite return and option/multi-return types
- infers implicit generic arguments through composite parameters such as `[]T` and `map[string]T`
- scopes the json2 voidptr workaround to the exact internal decoder helper

Focused validation:
- rebuilt `./vnew` with `./v -g -keepc -o ./vnew cmd/v`
- passed seven focused FastC generic/closure regressions in `fastc_test.v`
- verified gitly now rejects its first reached closure explicitly instead of publishing a binary with discarded callback behavior

Large suites were intentionally skipped.

### DIFF
--- a/vlib/v3/gen/fastc/anonfn.v
+++ b/vlib/v3/gen/fastc/anonfn.v
@@ -1,62 +1,8 @@
 module fastc
 
-// read_anonymous_function_stub parses a function literal / closure
-// (`fn [captures] (params) ret { body }`) in expression position and lowers it to a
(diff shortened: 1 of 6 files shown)


medvednikov:
@codex review

medvednikov:
@codex review

PR #28265, opened 2026-08-30
math.big, crypto.rand: add primality testing and prime generation

Chocapikk:
## What

Adds `Integer.is_probably_prime` and `jacobi` to `math.big`, and `prime` / `safe_prime` to `crypto.rand`.

`math.big` already has `mod_pow`, `mod_inverse` and `gcd`, but there is no way to decide primality, so RSA, DSA and Diffie-Hellman parameter generation cannot be written in V today. `vsl` has an `is_prime`, but it is trial division on `int`, so it does not apply to `big.Integer`, and `crypto` cannot depend on a third-party module anyway.

## How

`is_probably_prime` is a Baillie-PSW test:

1. trial division by the primes below 60
2. Miller-Rabin
3. an almost extra strong Lucas test, Baillie-OEIS method C

Method C fixes `Q = 1`, so only the V sequence is needed and the Lucas step reduces to repeated squaring with no modular halving. `U(s)` is recovered at the end through Crandall and Pomerance eq. 3.13 without ever computing the U sequence.

The first 13 Miller-Rabin bases are the fixed set that is provably sufficient below `3317044064679887385961981`; later rounds draw random bases, since fixed bases past that point add nothing against a crafted input.

Two things keep it fast:

- Trial division reduces the input **once per precomputed product** instead of once per prime, via a `mod_small` helper that walks the digit array without allocating. Rejecting a composite costs about 140 ns, which matters because key generation discards hundreds of candidates per prime found.
- Every Miller-Rabin round **shares one montgomery context** rather than deriving a modular inverse per round. This is why `mont_odd` is split into `mont_odd_with_ctx`, which takes a caller-supplied context; `mont_odd` keeps its old signature and behaviour, and is the only change to existing code.

Prime generation lives in `crypto.rand` rather than `math.big`, both because `crypto.rand` already imports `math.big` (the other direction would be an import cycle) and because key material should come from the OS entropy source. This also matches where Go puts it. `prime` sets the two highest bits, so the product of two same-sized primes has exactly twice as many bits, as RSA expects.

## Testing

`vlib/math/big/primality_test.v` and `vlib/crypto/rand/prime_test.v` cover:

- Carmichael numbers, which defeat a Fermat test
- strong pseudoprimes through the first 13 prime bases, including `3317044064679887385961981`
- Mersenne primes and Mersenne composites
- perfect squares, where no valid Lucas `D` exists
- semiprimes, negatives, zero, and the `rounds <= 0` default
- generated primes: exact bit length, oddness, primality, distinctness, and that `(p - 1) / 2` is prime for `safe_prime`

Separately, cross-checked against an independent sieve over **every odd number below 200000**, with no false positives and no false negatives.

## Numbers

`-prod`, averaged over repeated runs since generation timings vary with the candidate stream:

| operation | time |
|---|---|
| reject a 512-bit composite | 137 ns |
| test a 512-bit prime, 40 rounds | 29.5 ms |
| generate a 512-bit prime | 76 ms |
| generate a 1024-bit prime | 623 ms |
| generate a 2048-bit prime | ~6.3 s |

Time is dominated by the modular exponentiation in `mont_mul`, not by anything added here: a round using a randomly drawn base costs about the same as one using a fixed base. Reducing the allocations in `mont_mul` and `from_mont` would help the whole module, but that is a change to existing arithmetic and belongs in its own PR.

## Notes

- `v fmt -verify` and `v vet` are clean on all touched files.
- `v test vlib/math/big/` and `v test vlib/crypto/rand/` pass in full, so the `mont_odd` split does not regress existing exponentiation behaviour.
- The doc comment states plainly that a `true` result is a probability, not a certainty, and that this should not be used on its own to validate an adversary-supplied modulus.

### DIFF
--- a/vlib/crypto/README.md
+++ b/vlib/crypto/README.md
@@ -20,6 +20,25 @@ system's cryptographically secure random source and can return an error. The sep
 
 ## Examples
 
+### Prime generation
+
+Use `crypto.rand.prime(bits)` to generate an odd prime with exactly `bits` bits. Its two
+highest bits are set, so multiplying two primes of the same size produces a `2 * bits`-bit
+RSA modulus. `crypto.rand.safe_prime(bits)` additionally requires `(p - 1) / 2` to be prime
+and is considerably slower. Both functions use the operating system's cryptographically
+secure random source and can return an error.
+
+```v
+import crypto.rand
+
+fn main() {
+	p := rand.prime(256)!
+	safe := rand.safe_prime(256)!
+	assert p.bit_len() == 256
+	assert safe.bit_len() == 256
+}
+```
+
 ### Constant-time comparisons
 
 Use `crypto.subtle` for low-level helpers whose running time does not depend on secret data:
@@ -99,21 +118,27 @@ fn main() {
 }
 
 fn make_token(secret string) string {
-	header :=
-		base64.url_encode(json2.encode(JwtHeader{'HS256', 'JWT'}, escape_unicode: true).bytes())
+	header := base64.url_encode(
+		json2.encode(JwtHeader{'HS256', 'JWT'}, escape_unicode: true).bytes(),
+	)
 	payload := base64.url_encode(json2.encode(JwtPayload{'1234567890', 'John Doe', 1516239022},
 		escape_unicode: true
 	).bytes())
-	signature := base64.url_encode(hmac.new(secret.bytes(), '${header}.${payload}'.bytes(),
-		sha256.sum, sha256.block_size))
+	signature := base64.url_encode(
+		hmac.new(secret.bytes(), '${header}.${payload}'.bytes(), sha256.sum, sha256.block_size),
+	)
 	jwt := '${header}.${payload}.${signature}'
 	return jwt
 }
 
 fn auth_verify(secret string, token string) bool {
 	token_split := token.split('.')
-	signature_mirror := hmac.new(secret.bytes(), '${token_split[0]}.${token_split[1]}'.bytes(),
-		sha256.sum, sha256.block_size)
+	signature_mirror := hmac.new(
+		secret.bytes(),
+		'${token_split[0]}.${token_split[1]}'.bytes(),
+		sha256.sum,
+		sha256.block_size,
+	)
 	signature_from_token := base64.url_decode(token_split[2])
 	return hmac.equal(signature_from_token, signature_mirror)
 }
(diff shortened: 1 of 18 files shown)


Chocapikk:
@codex review

tankf33der:
testing.

medvednikov:
@codex review

medvednikov:
Addressed all five findings in 02edb35d93:

- run the full deterministic Miller–Rabin prelude, then every requested random round above the proven bound
- handle d == 1 without entering the Montgomery exponent path
- treat Lucas parameter-search exhaustion as inconclusive after Miller–Rabin acceptance
- make prime(2) consistently return 3 so the documented bit/oddness guarantees hold
- document the new math.big primality/Jacobi and crypto.rand prime APIs

Added regressions for 257, 65537, and the two-bit prime guarantee. Validation passed in release and debug modes for both primality_test.v and prime_test.v; check-md passed for both touched READMEs.

medvednikov:
@codex review

medvednikov:
Addressed the second review in the new head:

- random Miller–Rabin bases now use rejection sampling over [0, x - 3), producing a uniform base in [2, x - 2] without modulo bias
- safe_prime handles the unique four- and five-bit results directly, with regressions for 3-, 4-, and 5-bit calls

Both primality_test.v and crypto/rand/prime_test.v pass in release and debug modes.

medvednikov:
@codex review

medvednikov:
Addressed the global-PRNG finding in the new head:

- extracted the existing OS entropy reader to the dependency-neutral crypto.rand.internal module
- kept crypto.rand.bytes/read as API-compatible wrappers
- math.big now samples Miller–Rabin witnesses directly from that private OS source, so rand.set_rng cannot bias or stall primality checks
- added a regression that installs a seeded global generator and verifies primality testing does not consume it

Validation passed: primality test in release/debug, all 3 crypto.rand test files, prime test in release/debug, and C generation for Linux, Windows, FreeBSD, OpenBSD, and Solaris.

medvednikov:
@codex review

medvednikov:
Addressed both fourth-review findings in the new head:

- added is_probably_prime_checked, which propagates witness-entropy failures; the existing bool API now fails closed instead of panicking
- prime and safe_prime use the checked form, so their Result APIs propagate OS entropy failures
- added a Web Crypto implementation for JS and an explicit fallible WASM fallback
- the JS reader batches getRandomValues calls at the Web Crypto 65,536-byte limit

Validation passed: primality release/debug, prime release/debug, all crypto.rand tests, check-md, direct Node/Web Crypto execution with 70,000 bytes, and WASM frontend/code generation. Direct full math.big JS compilation remains blocked by the branch baseline duplicate JS.BigInt declaration, independent of this change; the previously missing entropy read now resolves and runs in isolation.

medvednikov:
@codex review

PR #28266, opened 2026-08-30
math.big: interleave montgomery multiplication with its reduction

Chocapikk:
## What

Rewrites `mont_mul` so that multiplication and montgomery reduction happen in a single pass, instead of building a double width product and reducing it afterwards. `big_mod_pow` and `mod_pow` get roughly 2-3x faster.

This is independent of #28265 and touches only `math.big`.

## Why

The current path allocates repeatedly for every montgomery multiplication:

```v
fn (a Integer) mont_mul(b Integer, ctx MontgomeryContext) Integer {
	t := a * b                  // full 2s digit product
	return t.from_mont(ctx)
}

fn (a Integer) from_mont(ctx MontgomeryContext) Integer {
	log2n := u32(ctx.n.bit_len())
	r := (a + ((a.mask_bits(log2n) * ctx.ni).mask_bits(log2n) * ctx.n)).right_shift(log2n)
	...
}
```

That is one `s x s` multiply for the product, two more inside `from_mont`, plus a mask, an add and a shift, each allocating. A windowed exponentiation calls this once per squaring, so a 512 bit modexp runs it several hundred times.

## How

The new `mont_mul` keeps one `s + 2` digit accumulator and, for each digit of `b`, adds `a * b[i]` into it and immediately cancels the low digit before shifting down. The product never exists as a value, and the work drops from three `s x s` multiplies to two.

Two supporting changes:

- **The radix becomes `base^s` instead of `2^bit_len(n)`**, so reduction is a digit shift rather than a bit shift. It is still coprime to an odd modulus and still larger than it, so montgomery form is unaffected.
- **The context stores a single digit inverse instead of a full one.** `n0inv` is `-n[0]^-1 mod 2^digit_bits`, found with five Newton steps, replacing the `mod_inv` call that `montgomery()` used to make. Building a context is now cheap, which matters when the caller creates one per exponentiation.

`mont_odd`, `to_mont`, `from_mont` and the sliding window loop are otherwise untouched, and nothing outside `exponentiation.v` refers to any of this.

## Numbers

`-prod`, same machine, master vs this branch:

| operation | before | after |
|---|---|---|
| 512-bit `big_mod_pow`, x200 | 166 ms | 54 ms |
| 2048-bit `big_mod_pow`, x10 | 256 ms | 125 ms |

## Testing

Adds `vlib/math/big/montgomery_test.v`, which pins the invariants the rewrite depends on:

- the digit inverse satisfies `d * mod_inv_digit(d) == 1 (mod base)`
- `n * n0inv == -1 (mod base)`, which is what cancels the low digit
- `rr == R^2 mod n`
- `to_mont` produces `x * R mod n`, and `from_mont` inverts it
- `mont_mul` really is multiplication in montgomery space
- zero, one and `n - 1` as operands
- Fermat's little theorem on several primes, which fails loudly on any reduction bug

Beyond that, cross-checked against an independent implementation over **25000 random `big_mod_pow` cases**, moduli from 64 to 800 bits, sizes chosen to straddle 60-bit digit boundaries. No disagreement.

That fuzzing was worth doing: it caught a real bug in an earlier version of this patch. When the modulus fills its top digit, twice the modulus no longer fits in `s` digits, so the accumulator can carry into digit `s`; dropping that carry produced wrong results for about half of all moduli with the high bit set, and no hand-written test case had exposed it.

## Notes

- `v fmt -verify` and `v vet` are clean.
- `v test vlib/math/big/` passes in full, including the existing exponentiation tests.

### DIFF
--- a/vlib/math/big/exponentiation.v
+++ b/vlib/math/big/exponentiation.v
@@ -7,9 +7,9 @@ are based on refer to https://github.com/vlang/v/pull/18461
 
 // internal struct to make passing montgomery values simpler
 struct MontgomeryContext {
-	n  Integer // |modulus|
-	ni Integer // n^(-1)
-	rr Integer // for conversions
+	n     Integer // |modulus|
+	rr    Integer // R^2 mod n, for conversions into montgomery form
+	n0inv u64 // -n[0]^-1 mod 2^digit_bits, the reduction multiplier
 }
 
 // montgomery calculates a montgomery context for reductions to montgomery space based on
@@ -21,16 +21,16 @@ fn (m Integer) montgomery() MontgomeryContext {
 	}
 
 	n := m.abs()
-	b := u32(n.bit_len())
+	// R is the smallest power of the base above n, so that reduction is a
+	// digit-wise shift rather than a bit-wise one.
+	r_bits := u32(n.digits.len * digit_bits)
 
 	return MontgomeryContext{
 		n: n
-		// r = 2^(log_2(n))
-		// ri := multiplicative inverse of r in the ring Z/nZ
-		// ri * r == 1 (mod n)
-		// ni = ((ri * 2^(log_2(n))) - 1) / n
-		ni: (one_int.left_shift(b).mod_inv(n).left_shift(b) - one_int) / n
-		rr: one_int.left_shift(b * 2) % n
+		rr: one_int.left_shift(r_bits * 2) % n
+		// n * n0inv == -1 (mod base), which is what makes the low digit of
+		// t + m * n vanish during reduction.
+		n0inv: (-mod_inv_digit(n.digits[0])) & max_digit
 	}
 }
 
@@ -185,12 +185,15 @@ fn (a Integer) mont_even(x Integer, m Integer) Integer {
 	t1 := x1.mask_bits(m2n)
 	t2 := x2.mask_bits(m2n)
 
-	t := (if t2.abs_cmp(t1) >= 0 {
-		(t2 - t1).mask_bits(m2n)
-	} else {
-		// (x2 - x1) % m2 = 1 + ((~((x2 % m2) - (x1 % m2))) % m2)
-		(t1 - t2).abs().bitwise_not().mask_bits(m2n) + one_int
-	} * m1i).mask_bits(m2n)
+	t := (
+		if t2.abs_cmp(t1) >= 0 {
+			(t2 - t1).mask_bits(m2n)
+		} else {
+			// (x2 - x1) % m2 = 1 + ((~((x2 % m2) - (x1 % m2))) % m2)
+			(t1 - t2).abs().bitwise_not().mask_bits(m2n) + one_int
+		} * m1i
+	)
+	.mask_bits(m2n)
 
 	return x1 + m1 * t
 }
@@ -295,18 +298,6 @@ fn get_window_size(n u32) int {
 	}
 }
 
-// mont_mul performs multiplication of two variables in montgomery
-// space and reduces the result to montgomery space
-fn (a Integer) mont_mul(b Integer, ctx MontgomeryContext) Integer {
-	if (a.digits.len + b.digits.len) > 2 * ctx.n.digits.len {
-		return zero_int
-	}
-
-	t := a * b
-
-	return t.from_mont(ctx)
-}
-
 fn (a Integer) to_mont(ctx MontgomeryContext) Integer {
 	return a.mont_mul(ctx.rr, ctx)
 }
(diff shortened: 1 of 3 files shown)


tankf33der:
testing

medvednikov:
## Review verdict

**Request changes.** Ignoring CI as requested, I found no defect in the new Montgomery arithmetic itself, but the rewrite introduces an asymptotic performance regression for large operands. There is also one misleading test that never executes the changed implementation.

### [P2] Preserve a subquadratic path for large moduli

The new `mont_mul` unconditionally performs two nested `s × s` loops: one for multiplication and one for reduction. Its runtime is therefore always \(O(s^2)\).

The previous implementation performed three `Integer` multiplications, but those multiplications go through `multiply_digit_array`, which switches to Karatsuba at 70 limbs and Toom-3 at 360 limbs. Consequently, the old implementation is subquadratic for sufficiently large operands, whereas the replacement never is.

With 60-bit limbs, the Karatsuba threshold is approximately 4,200 bits. The PR only reports 512- and 2,048-bit results, both below that threshold, so those benchmarks cannot reveal the regression.

I would retain the CIOS implementation for small and medium operands but keep or reconstruct the previous multiplication-based reduction above a measured crossover. At minimum, benchmark 4,096, 8,192, 16,384, and 32,768-bit moduli before removing the subquadratic path. The precise wall-clock crossover needs measurement, but the worse asymptotic behavior is definite.

### [P3] `test_big_mod_pow_matches_mod_pow` does not test Montgomery multiplication

Every exponent in this test is below \(2^{60}\):

```v
for e in [u64(2), 3, 17, 65537, 4294967311] {
    assert b.big_mod_pow(integer_from_u64(e), m)! == b.mod_pow(e, m)
}
```

They therefore occupy one `Integer` digit. `big_mod_pow` explicitly routes one-digit exponents directly to `mod_pow`, meaning both sides of the assertion execute `mod_pow`; none of the new Montgomery code is exercised.

Use an exponent that still fits `u64` but spans two 60-bit digits, for example:

```v
e := (u64(1) << digit_bits) + 65537
assert b.big_mod_pow(integer_from_u64(e), m)! == b.mod_pow(e, m)
```

That preserves `mod_pow` as an independent oracle while forcing `big_mod_pow` through `mont_odd`. The `m2` known-value assertion with exponent `1000` also bypasses Montgomery, although the larger known exponent and Fermat tests do exercise it.

## Arithmetic validation

The core rewrite otherwise looks sound:

* Five Newton iterations are sufficient for a 60-bit inverse, starting from the three correct low bits.
* The carry entering `mul_add_digit` is bounded so that `t[j] + carry < 2^61`; it cannot overflow `u64`.
* The product plus incoming value is below \(2^{120}\), so `bits.mul_add_64` has sufficient width.
* Preserving digit `s` correctly handles moduli whose most-significant limb has its high bit set.
* The final accumulator is below \(2n\), so one conditional subtraction is sufficient.

I also translated the limb algorithm into an independent exact-integer model. It matched the Montgomery formula for **90,000 random multiplications** across one through six limbs, boundary operands such as `0`, `1`, and `n−1`, and **648 complete sliding-window exponentiations**, including window sizes 4, 5, and 6.

So the arithmetic change appears correct; the blocking concern is retaining acceptable behavior once operands cross the existing Karatsuba and Toom-3 thresholds.

tankf33der:
Tested. No issues.
Confirm performance boost.

Tested two big branches of tests against GMP:

`20M` input random numbers of random bit length for both 3 arguments.
and
```
base is random
exponent's limb length >= 2
modulus is odd,
```

@Chocapikk - Let me know if the tests are too dumb and pointless.

medvednikov:
Addressed both review requests in `0706f5423c`:

- retain allocation-light CIOS below the existing 70-limb Karatsuba threshold
- switch to the multiplication-based Montgomery reduction at/above that threshold, preserving Karatsuba and Toom-3 scaling
- add a two-limb `u64` exponent so `big_mod_pow` actually exercises Montgomery code against `mod_pow`
- add threshold coverage comparing the dispatcher result directly with CIOS

Measured per-operation crossover (Apple arm64, repeated Montgomery multiplies):
- 4096 bits / 69 limbs: CIOS 117.8 µs, multiplication reduction 141.6 µs
- 8192 bits: CIOS 468.2 µs, multiplication reduction 378.1 µs
- 16384 bits: CIOS 1852.3 µs, multiplication reduction 1040.3 µs
- 32768 bits: CIOS 7269.2 µs, multiplication reduction 2608.8 µs

At 4320 bits, where Karatsuba is active, the multiplication reduction was already faster (126.8 µs vs 88.6 µs), supporting the existing 70-limb threshold.

Validation:
- rebuilt `./vnew` with a contemporaneous bootstrap compiler
- formatted all three touched V files
- all 8 `vlib/math/big/*_test.v` files passed individually
- `montgomery_test.v` passed after the final threshold equality assertion
- `git diff --check`

The module test-runner wrapper on this stale branch initially failed before running tests because a current bootstrap emitted `autostr` calls absent from the branch builtin; rebuilding with the workspace’s Aug 28 bootstrap resolved that tooling mismatch.

medvednikov:
@codex review

PR #28268, opened 2026-08-30
encoding.punycode: add RFC 3492 punycode encoding

Chocapikk:
## What

Adds `encoding.punycode`, with `encode` and `decode`. Punycode is the ASCII-compatible encoding that internationalised domain names are built on, specified in RFC 3492.

Independent of #28265 and #28266.

## Why

Nothing in the ecosystem provides it: not vlib, not vpm, not awesome-v. The only hit on GitHub is a private helper inside an unrelated project.

There is also a concrete consumer in the standard library today. `net.urllib.parse_host` keeps a non-ASCII host verbatim, so:

```v
u := urllib.parse('https://münchen.de/x')!
println(u.hostname()) // münchen.de, not xn--mnchen-3ya.de

http.get('https://münchen.de/')! // 400, the Host header carries raw UTF-8
```

Punycode is the missing piece for an IDNA layer that would fix this. Whether `urllib` should convert automatically is a separate decision, so this PR only adds the encoding and changes no existing behaviour.

It also fits the neighbourhood: `encoding/` already holds base32, base58, base64, hex and leb128, and punycode is the same kind of thing, a small self-contained standard encoding.

## Scope

Single labels only, and deliberately no IDNA processing: the module neither adds nor strips the `xn--` prefix, does not case fold, and does not normalise. Those are IDNA decisions and belong to the caller.

Deltas that walk past U+10FFFF or into the surrogate range are reported as errors rather than producing an invalid rune. The encoder emits lowercase digits; the decoder accepts either case, as RFC 3492 section 5 requires.

## Testing

`punycode_test.v` covers:

- **all 19 sample strings from RFC 3492 section 7.1**, in both directions, including the pure-ASCII example (S) that breaks host name rules
- the labels behind several real `xn--` domains
- round trips, empty input, pure ASCII, mixed-case digits
- the malformed cases the RFC calls out: an invalid digit, a delta that never terminates, a non-basic byte in the literal part, and a code point past U+10FFFF

Beyond the RFC vectors, cross-checked against an independent implementation over **12000 random strings**: 6000 encoded here and verified elsewhere, 6000 encoded elsewhere and decoded here, with code points up to U+2FFFF. No disagreement in either direction.

That cross-check earned its keep: it surfaced the missing U+10FFFF bound, which the RFC vectors alone never reach.

## Notes

- `v fmt -verify` and `v vet` are clean.
- 228 lines of implementation, 202 of tests, no dependencies beyond builtin.

### DIFF
--- a/vlib/encoding/punycode/punycode.v
+++ b/vlib/encoding/punycode/punycode.v
@@ -1,0 +1,228 @@
+// Package punycode implements the Punycode encoding described in RFC 3492,
+// the ASCII-compatible encoding used by internationalised domain names.
+//
+// Punycode is an instance of the more general Bootstring encoding, tuned for
+// Unicode code points and host name labels. It rewrites a string of arbitrary
+// code points as a string of basic ASCII code points, so that names outside
+// the ASCII range survive protocols that only accept ASCII.
+//
+// This module encodes and decodes single labels. It performs no IDNA
+// processing: it does not add or strip the `xn--` prefix, does not case fold
+// and does not normalise. Callers doing IDNA are expected to handle those
+// steps themselves.
+module punycode
+
+// Bootstring parameters for Punycode, from RFC 3492 section 5. Only `base`,
+// `tmin`, `tmax` and the delimiter are fixed by the format; the rest affect
+// efficiency alone.
+const base = u32(36)
+const tmin = u32(1)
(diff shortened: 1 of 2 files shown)


medvednikov:
@codex review

PR #28269, opened 2026-08-30
net.imap: add an IMAP4rev1 client

Chocapikk:
`net.smtp` sends mail but nothing in vlib reads it, so a V program that wants to look at a mailbox has to shell out or reimplement IMAP. This adds `net.imap`, a client for IMAP4rev1 (RFC 3501).

I checked vpm and awesome-v first: there is no IMAP module in either.

```v
mut c := imap.new_client(
    server:   'imap.example.com'
    username: 'someone@example.com'
    password: 'hunter2'
    ssl:      true
)!
defer { c.close() or {} }

inbox := c.select_mailbox('INBOX')!
println('${inbox.exists} messages, ${inbox.recent} recent')

unread := c.search('UNSEEN')!
for msg in c.fetch(unread, '(UID ENVELOPE)')! {
    if envelope := msg.envelope {
        println('${envelope.from[0].addr()}: ${envelope.subject}')
    }
}
```

### Responses are tokenised, not searched

The module reads the grammar of RFC 3501 section 9 rather than looking for keywords in a response line. That is not tidiness, it is correctness: a message whose subject is `UID 999` answers a search for `UID ` exactly as well as the real field does, and a body containing `)` ends a response that a bracket-counting parser would still be reading.

`wire.v` is a decoder over the four shapes the protocol is built from: atoms, quoted strings, literals and parenthesised lists. It reads the connection directly with one octet of lookahead, so a literal is taken by count and its content is never offered back to the parser. An item this module does not model is stepped over whole by `skip_value`, so a server extension cannot desynchronise the rest of the response.

Two grammar details worth naming, because both are easy to get wrong and both are covered by tests:

- `env-cc = "(" 1*address ")"`. The addresses in an envelope are concatenated with nothing between them, not separated by spaces like every other list in the protocol.
- `[` is a valid `ATOM-CHAR`; only `]` is excluded. Reading a fetch item name with a plain atom reader therefore swallows the opening bracket of `BODY[]` and leaves the rest unreadable.

### Mailbox names

Names are ordinary UTF-8 in the API and travel as the modified UTF-7 of section 5.1.3, encoded on the way out and decoded on the way in. Without it, `list_mailboxes` hands back `&AMk-l&AOk-ments` where the user's mailbox is called `Éléments`. The encoding is checked against the two examples printed in the RFC and cross-checked against an independent implementation.

### Sequence sets

`SeqSet` keeps message numbers as merged ranges, so a search matching fifty thousand messages is sent as `FETCH 1:50000 ...`. A comma separated list of every number would be a command line no server is obliged to accept.

### What is implemented

CAPABILITY, NOOP, LOGOUT, STARTTLS, AUTHENTICATE PLAIN, LOGIN, SELECT, EXAMINE, CREATE, DELETE, RENAME, SUBSCRIBE, UNSUBSCRIBE, LIST, LSUB, STATUS, APPEND, CHECK, CLOSE, UNSELECT, EXPUNGE, SEARCH, FETCH, STORE, COPY, MOVE, and the UID forms of the last five.

FETCH parses UID, FLAGS, RFC822.SIZE, INTERNALDATE, ENVELOPE, BODYSTRUCTURE and any number of `BODY[...]` sections, keyed by the specification the server echoed back.

Some things that took deliberate handling:

- An argument a quoted string cannot hold, an eight bit password or a message body, is sent as a literal: the introducer goes out, the client waits for the `+` continuation, then the octets follow. Section 2.2.1 says the server may send untagged data or reject the command instead of continuing, so the wait handles both rather than assuming the next line is the `+`.
- A mailbox changes under a client at any moment, and the server says so by slipping EXISTS, EXPUNGE and FETCH responses into whatever command happens to be in flight. Those are collected rather than discarded, and `Client.exists` and `Client.recent` follow them.
- A completion carrying a tag other than the outstanding one is an error. It means the client and the server disagree about which command is running, and nothing later recovers from that.
- A literal announces its own length, so a server can ask a client to allocate as much memory as it likes. There is a cap, checked before the allocation.
- An untagged BYE closes the session, and a greeting that turns the connection away closes the socket rather than leaving the server to time it out.

`ESEARCH` (RFC 4731) and `MOVE` (RFC 6851) are supported; `supports()` is there to guard the optional ones.

### Tests

Four files, 513 assertions, no external server needed.

- `utf7_test.v` covers the RFC's own examples, a round trip over Japanese, Russian, Chinese, French and an emoji, the astral plane surrogate pair, and six malformed inputs that must be refused.
- `seqset_test.v` covers merging, ordering, `*`, parsing and the fifty thousand message case.
- `response_test.v` runs the response reader over transcripts taken from the RFC, including its envelope and body structure examples, plus the two traps above: a body and an envelope subject that both read like protocol.
- `imap_test.v` drives the client over a real socket against a scripted server, checking the exact bytes each command puts on the wire, the literal handshake for APPEND and for an eight bit password, and that an unsolicited EXISTS reaches the client.

`v fmt -verify` and `v vet -W` are clean on all nine files, and none of them show up in `v test-cleancode`.

### DIFF
--- a/vlib/net/imap/imap.v
+++ b/vlib/net/imap/imap.v
@@ -1,0 +1,671 @@
+// Package imap implements a client for the Internet Message Access Protocol,
+// version 4rev1, as described in RFC 3501.
+//
+// IMAP is the reading half of the mail pair whose writing half is `net.smtp`:
+// where SMTP hands a message to a server, IMAP browses what a server already
+// holds. Messages stay on the server, so a client selects a mailbox, searches
+// it, and fetches only the parts it wants.
+//
+// Mailbox names are ordinary UTF-8 strings here. They are encoded and decoded
+// on the way in and out, so the modified UTF-7 the protocol carries them in
+// never reaches a caller.
+//
+// A minimal session looks like this:
+//
+// ```v ignore
+// mut c := imap.new_client(
+//     server:   'imap.example.com'
+//     username: 'someone@example.com'
+//     password: 'hunter2'
+//     ssl:      true
+// )!
+// defer { c.close() or {} }
+//
+// inbox := c.select_mailbox('INBOX')!
+// println('${inbox.exists} messages, ${inbox.recent} recent')
+//
+// unread := c.search('UNSEEN')!
+// for msg in c.fetch(unread, '(UID ENVELOPE)')! {
+//     if envelope := msg.envelope {
+//         println('${envelope.subject}')
+//     }
+// }
+// ```
+module imap
+
+import io
+import net
+import net.ssl
+import encoding.base64
+import time
+
+// The port a server listens on depends on whether TLS is negotiated up front.
+pub const default_port = 143
+pub const default_ssl_port = 993
+
+// Every command carries a tag that its completion repeats back, which is what
+// lets a client tell a reply to its command from the mailbox updates a server
+// may interleave with it.
+const tag_prefix = 'a'
+
+// Config holds the settings used to open a session.
+//
+// Leave `port` unset to take 993 when `ssl` is true and 143 otherwise. Set
+// `ssl` for a connection encrypted from the first byte, or `starttls` to
+// upgrade a plain connection once it is open. The two are mutually exclusive.
+pub struct Config {
+pub:
+	server   string
+	port     int
+	username string
+	password string
+	ssl      bool
+	starttls bool
+	timeout  time.Duration
+}
+
+// Client is a connection to an IMAP server.
+pub struct Client {
+	Config
+mut:
+	conn     net.TcpConn
+	ssl_conn &ssl.SSLConn = unsafe { nil }
+	dec      ?&Decoder
+	tag_seq  int
+pub mut:
+	is_open   bool
+	encrypted bool
+	// selected names the mailbox the session is working in, and is empty when
+	// none has been selected.
+	selected string
+	// exists and recent track the counts the server reports, which it may do
+	// during any command as messages arrive or are removed by another client.
+	exists u32
+	recent u32
+}
+
+// new_client opens a session, greets the server, upgrades the connection if
+// asked to, and logs in.
+pub fn new_client(config Config) !&Client {
+	if config.ssl && config.starttls {
+		return error('imap: cannot use both implicit SSL and STARTTLS')
+	}
+	mut c := &Client{
+		Config: config
+	}
+	c.connect()!
+	c.login()!
+	return c
+}
+
+// connect opens the transport and reads the server greeting, without logging
+// in. `new_client` calls it; call it directly only to drive a session by hand.
+pub fn (mut c Client) connect() ! {
+	c.conn = net.dial_tcp('${c.server}:${c.effective_port()}')!
+	if c.timeout != 0 {
+		c.conn.set_read_timeout(c.timeout)
+		c.conn.set_write_timeout(c.timeout)
+	}
+	c.dec = &Decoder{
+		reader: io.new_buffered_reader(reader: c.conn)
+	}
+	c.is_open = true
+
+	if c.ssl {
+		c.upgrade_to_tls()!
+	}
+	// A greeting that turns the connection away leaves nothing to say goodbye
+	// to, so the socket goes rather than being left for the server to time
+	// out.
+	c.read_greeting() or {
+		c.shutdown()
+		return err
+	}
+	if c.starttls {
+		c.run('STARTTLS')!
+		c.upgrade_to_tls()!
+	}
+}
+
+// login authenticates with the LOGIN command, and does nothing when no
+// username was configured.
+pub fn (mut c Client) login() ! {
+	if c.username == '' {
+		return
+	}
+	text, literals := build_args('LOGIN', [c.username, c.password])
+	c.send(text, literals)!
+}
+
+// login_plain authenticates with the SASL PLAIN mechanism instead, which some
+// servers require and others prefer.
+pub fn (mut c Client) login_plain() ! {
+	tag := c.next_tag()
+	c.write_line('${tag} AUTHENTICATE PLAIN')!
+	c.await_continuation(tag)!
+	// RFC 4616: an authorisation identity, an authentication identity and a
+	// password, joined by NUL bytes.
+	c.write_line(base64.encode_str('\0${c.username}\0${c.password}'))!
+	c.read_response(tag)!
+}
+
+// capability returns the extensions the server advertises.
+pub fn (mut c Client) capability() ![]string {
+	return c.run('CAPABILITY')!.capabilities
+}
+
+// supports reports whether the server advertises `name`, which is how an
+// optional command should be guarded before it is sent.
+pub fn (mut c Client) supports(name string) !bool {
+	for cap in c.capability()! {
+		if cap.to_upper() == name.to_upper() {
+			return true
+		}
+	}
+	return false
+}
+
+// noop asks the server for nothing. It keeps the connection alive, and gives
+// the server the chance to report anything that changed in the mailbox.
+pub fn (mut c Client) noop() ! {
+	c.run('NOOP')!
+}
+
+// list_mailboxes returns the mailboxes matching `pattern` below `reference`.
+//
+// The usual call is `list_mailboxes('', '*')`, which lists everything the
+// account can see. `%` matches within one level of the hierarchy where `*`
+// crosses levels.
+pub fn (mut c Client) list_mailboxes(reference string, pattern string) ![]MailboxInfo {
+	return c.run_list('LIST', reference, pattern)
+}
+
+// list_subscribed returns the subscribed mailboxes matching `pattern`, which
+// is the set a mail client would show by default.
+pub fn (mut c Client) list_subscribed(reference string, pattern string) ![]MailboxInfo {
+	return c.run_list('LSUB', reference, pattern)
+}
+
+// select_mailbox opens a mailbox for reading and writing, and returns its
+// state. Later fetches, searches and stores act on it.
+pub fn (mut c Client) select_mailbox(name string) !Mailbox {
+	return c.open_mailbox('SELECT', name)
+}
+
+// examine_mailbox opens a mailbox read-only. It behaves like
+// `select_mailbox`, except that nothing the session does marks messages as
+// seen or otherwise alters the mailbox.
+pub fn (mut c Client) examine_mailbox(name string) !Mailbox {
+	return c.open_mailbox('EXAMINE', name)
+}
+
+// status reports on a mailbox without selecting it, which is how a client
+// polls for new mail while working in another mailbox.
+//
+// `items` names what to ask for: MESSAGES, RECENT, UIDNEXT, UIDVALIDITY and
+// UNSEEN.
+pub fn (mut c Client) status(name string, items []string) !MailboxStatus {
+	if items.len == 0 {
+		return error('imap: a STATUS command must ask for at least one item')
+	}
+	text, literals := build_args('STATUS', [utf7_encode(name)])
+	last := text.len - 1
+	mut tail := text.clone()
+	tail[last] = '${text[last]} (${items.join(' ')})'
+	res := c.send(tail, literals)!
+	if res.statuses.len == 0 {
+		return error('imap: the server reported no status for `${name}`')
+	}
+	return res.statuses[0]
+}
+
+// create_mailbox creates a mailbox.
+pub fn (mut c Client) create_mailbox(name string) ! {
+	c.run_with_mailbox('CREATE', name)!
+}
+
+// delete_mailbox deletes a mailbox and everything in it.
+pub fn (mut c Client) delete_mailbox(name string) ! {
+	c.run_with_mailbox('DELETE', name)!
+}
+
+// subscribe adds a mailbox to the set a client shows by default.
+pub fn (mut c Client) subscribe(name string) ! {
+	c.run_with_mailbox('SUBSCRIBE', name)!
+}
+
+// unsubscribe removes it from that set, without touching the mailbox itself.
+pub fn (mut c Client) unsubscribe(name string) ! {
+	c.run_with_mailbox('UNSUBSCRIBE', name)!
+}
+
+// rename_mailbox renames a mailbox, along with everything below it in the
+// hierarchy.
+pub fn (mut c Client) rename_mailbox(from string, to string) ! {
+	text, literals := build_args('RENAME', [utf7_encode(from), utf7_encode(to)])
+	c.send(text, literals)!
+}
+
+// append adds a message to a mailbox, which is how a sent message is filed
+// into a Sent folder or a draft is saved.
+//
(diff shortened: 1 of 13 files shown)


Chocapikk:
Ran a full session against a real Dovecot 2.3.21, since a mock I wrote myself cannot catch a misreading I baked into both sides of it.

It found a bug, fixed in 225072e. Dovecot answers `EXPUNGE` and `CLOSE` with the removals alone and no fresh `EXISTS` behind them, so the client went on reporting three messages after expunging one of three. Each `EXPUNGE` now decrements the count, and an `EXISTS` the server does send still wins. The scripted server had been sending one and hiding it; both cases are covered now.

The rest held up:

```
created `Travail/Été 2026`
  name=`Travail/Été 2026` delim=`/` attrs=['\HasNoChildren']
unicode name round-tripped through the server: true

INBOX: exists=3 recent=0 uidnext=4 uidvalidity=1788105033 readonly=false
search UNSEEN -> 2:3, SEEN -> 1

  seq=1 uid=1 size=306 flags=['\Seen'] internal=1996-07-17 02:23:25
  seq=1 subject=`UID 999 RFC822.SIZE 5`
  parsed uid=1 size=306 (must not be 999 and 5)

  seq=3 multipart/alternative
      part text/plain enc=7bit size=13
      part text/html enc=quoted-printable size=19

  text of message 2 (eight bit, went out as a literal): `Déjà vu, à demain. 日本語`
  two sections in one fetch, keys: ['BODY[HEADER]', 'BODY[1]']

  uid search ALL -> 1:3
  expunged: [1], client now sees exists=2
```

Worth pulling out: the mailbox is created and listed by its UTF-8 name with the modified UTF-7 never surfacing; the message whose subject is `UID 999 RFC822.SIZE 5` parses as `uid=1 size=306`, which is the reason responses are tokenised rather than searched; the eight bit message went out as a literal behind a `+` continuation and came back byte for byte; `INTERNALDATE` round-tripped exactly; searches come back as ranges rather than expanded lists.

Setup was `dovecot/dovecot:2.3.21`, static passdb, no TLS. Happy to add this as an integration test behind a build tag, though it needs a server so it cannot run in CI unguarded.

medvednikov (commented):
Thanks for the very thorough parser work and test coverage. I found several correctness issues that should be addressed before merge:

1. **Greeting/authentication state is lost (`imap.v`).** `read_greeting()` accepts `PREAUTH`, but `new_client()` still calls `LOGIN` whenever `username` is set. A `PREAUTH` greeting already puts the connection in authenticated state, so both `LOGIN` and `STARTTLS` are invalid there. The greeting's `CAPABILITY` response code is also parsed and discarded, so a server can advertise `LOGINDISABLED` and the client will still send `LOGIN` (including the password). Please retain the greeting state/capabilities and gate STARTTLS/authentication accordingly. Regression tests for `PREAUTH` with configured credentials and for greeting-time `LOGINDISABLED` would catch both cases.

2. **A failed `SELECT`/`EXAMINE` leaves stale selected-mailbox state.** IMAP returns to authenticated state after a failed selection. `open_mailbox()` only assigns `c.selected` on success, so selecting a missing mailbox while another mailbox is selected leaves `c.selected` pointing at the old mailbox. Clear the selected state (and any selected-mailbox counters) when a new selection is attempted/fails, and add a selected -> failed-select test.

3. **Several post-dial error paths can leak the transport.** Failures during implicit TLS, STARTTLS, or the automatic login escape without `shutdown()`. Separately, an untagged `BYE` sets `is_open = false`, so a later `close()` returns before closing the local socket. Please make every error after a successful dial, plus BYE/logout handling, tear the transport down exactly once.

4. **Malformed server `INTERNALDATE` data can panic the process.** `parse_internal_date()` feeds unchecked `.int()` results into `time.new()`, whose range validation panics. A response containing an out-of-range day/hour/etc. should return an IMAP parse error rather than terminate the client. The zone digits/ranges should be validated too.

5. **Unknown untagged responses containing literals still desynchronise the decoder.** The fallback in `read_untagged()` calls `d.text()`, which stops at the CRLF after a literal marker such as `{3}` and leaves the literal payload to be parsed as the next response. A transcript like `* XDATA {3}\r\nabc\r\na1 OK done\r\n` demonstrates it. The unknown-response skip needs to be literal-aware to uphold the stated extension-safety guarantee.

6. **`APPEND` can mislabel local time as UTC.** `format_internal_date()` always emits `+0000` using the supplied struct's raw fields, while `time.now()` is local time. Convert with `local_to_utc()` before formatting (or emit the actual offset), otherwise the stored internal date is shifted by the caller's timezone.

Smaller edge case: `seq_max = u32(0xffffffff)` collides with the largest valid IMAP number. For example, `parse_seq_set('4294967295')!.str()` becomes `*`; using a wider internal bound would keep the numeric value distinct from the star sentinel.

medvednikov:
Addressed the existing review at e686d693bf:

- retain greeting authentication/capability state, skip LOGIN after PREAUTH, reject greeting-time LOGINDISABLED, and prevent PREAUTH from bypassing requested STARTTLS
- clear selected mailbox state and counters before each SELECT/EXAMINE attempt
- centralize post-dial cleanup for implicit TLS, STARTTLS, and automatic-login failures; make BYE/logout close the local transport idempotently
- validate every INTERNALDATE component before constructing time.Time
- skip unknown untagged responses across literal payloads
- convert local APPEND timestamps to UTC before emitting +0000
- preserve 4294967295 as a numeric sequence value distinct from *

Focused formatting and all four net.imap test files pass with tcc. The directory test runner itself cannot be built by this stale PR-head compiler because its generated C omits builtin autostr declarations; the individual test executables avoid that unrelated harness failure.

medvednikov:
@codex review

medvednikov:
Fixed both Codex findings at 640d9992b4:

- TLS peer validation is now enabled by default, with CA bundle, in-memory PEM, and client-certificate settings exposed through imap.Config and forwarded to both SSL backends before any credentials are sent
- unknown FETCH extension attributes now consume attached section suffixes such as BINARY[1] before skipping their values

All four focused net.imap test executables pass again, including the new default-validation and BINARY[1] regressions.

medvednikov:
@codex review

medvednikov:
Fixed the three follow-up findings at HEAD:

- retain the number of EXPUNGE events after the last EXISTS so mailbox changes are applied in wire order
- read NO/BAD completions as data first, absorb preceding unsolicited updates, then return the command error
- recognize RFC 3516 literal8 (`~{n}`) values while skipping extension FETCH attributes

Added socket-level regressions for ordered updates and updates preceding a refused command, plus a BINARY[1] literal8 transcript. All four focused net.imap test executables pass.

medvednikov:
@codex review

medvednikov:
Fixed all three findings on the current head:

- require an explicit CA bundle whenever TLS certificate validation is enabled, so validation cannot silently run without trust roots
- verify the peer certificate hostname after OpenSSL completes chain validation
- add a construction-time `auth_method` choice and authenticate with SASL PLAIN when selected, including `LOGINDISABLED` greetings

Added regressions for TLS trust-root preflight and SASL PLAIN construction. Focused IMAP tests pass with both the default SSL backend and `-d use_openssl`.

medvednikov:
@codex review

medvednikov:
Fixed both follow-up findings at the current head:

- OpenSSL certificate identity checks now dispatch valid textual IP literals through `X509_check_ip_asc`, while DNS names continue through `X509_check_host`
- failed `connect` paths now release initialized TLS backend state in both OpenSSL and mbedTLS; OpenSSL hostname rejection is part of `connect`, so it shares the same cleanup path

The focused IMAP socket tests pass with both the default backend and `-d use_openssl`.

medvednikov:
@codex review

medvednikov:
Fixed both follow-up findings at the current head:

- public `connect()` now rejects an already-open client before replacing its owned transport, preserving the original live session and state
- the generic FETCH extension skipper consumes backslash-prefixed flags, including `\Inbox` and `\*`, inside nested values

Added regressions proving a rejected reconnect leaves the first session usable and that X-GM-LABELS with a system flag does not derail later FETCH fields. All focused IMAP tests pass on the default backend and the socket suite also passes with `-d use_openssl`.

medvednikov:
@codex review

medvednikov:
Fixed both follow-up findings at the current head:

- fatal response-reader failures now shut down the desynchronized session, while successfully parsed NO/BAD completions still apply unsolicited state and remain reusable
- hostname and IP certificate identity checks now call version-guarded compatibility wrappers; pre-1.0.2 OpenSSL builds remain linkable and fail identity validation closed when those APIs are unavailable

Added a socket regression proving that a mismatched completion tag closes the transport. Focused IMAP response/socket tests pass on the default backend and the socket suite passes with `-d use_openssl`.

medvednikov:
@codex review

medvednikov:
Fixed all five findings at the current head:

- the TLS example now supplies the explicit CA bundle required by the secure default
- every command/authentication write failure invalidates and closes the session
- continuation parsing distinguishes a fully parsed command refusal from fatal malformed/wrong-tag responses, retaining reuse only for the former
- IMAP debug output redacts LOGIN credentials and SASL PLAIN payloads
- multipart BODYSTRUCTURE parsing retains content-type parameters such as inner and outer boundaries

Added socket regressions for fatal and refused literal continuations, direct redaction checks, and multipart boundary assertions. All four focused default-backend tests pass; the socket suite also passes with `-d use_openssl` and `-d imap_debug`, with no test credentials in debug output.

medvednikov:
@codex review

medvednikov:
Fixed the follow-up finding at the current head: `BodyStructure` now exposes the embedded envelope and nested MIME tree of `message/rfc822` parts instead of discarding both after parsing. Added a forwarded-message BODYSTRUCTURE regression that verifies its subject, message ID, nested content type, charset, size, and line counts. Focused response/socket tests pass on the default backend and the socket suite passes with `-d use_openssl`.

medvednikov:
@codex review

medvednikov:
Addressed the latest review findings in b0c545afb3:

- enable `SSL_VERIFY_PEER` whenever OpenSSL certificate validation is enabled, before checking the certificate result and hostname
- distinguish unknown-extension free text ending in `{n}` from a literal payload, including a regression where consuming the following tagged reply would desynchronize the session

Focused validation passed:
- `./vnew fmt -verify vlib/net/imap/wire.v vlib/net/imap/response_test.v`
- `./vnew -cc tcc vlib/net/imap/response_test.v`
- `./vnew -cc tcc vlib/net/imap/imap_test.v`
- `./vnew -cc tcc vlib/net/imap/seqset_test.v`
- `./vnew -cc tcc vlib/net/imap/utf7_test.v`
- `./vnew -d use_openssl -cc tcc vlib/net/imap/imap_test.v`

medvednikov:
@codex review

medvednikov:
Addressed the latest METADATA literal finding in 398d228267. Unknown response parsing now tracks quote and parenthesized-list context across physical lines, so free-form top-level text ending in `{n}` remains text while literals after extension arguments (including multiple METADATA values) are consumed correctly.

Validation passed:
- `./vnew fmt -w vlib/net/imap/wire.v vlib/net/imap/response_test.v`
- `./vnew -cc tcc vlib/net/imap/response_test.v`
- `./vnew -cc tcc vlib/net/imap/imap_test.v`
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed the plaintext-credential finding in 3cb1138669. Configured credentials now require SSL/STARTTLS by default; intentional plaintext authentication requires the explicit `allow_insecure_auth: true` opt-in. The guard applies both before `new_client` dials and in the public LOGIN/SASL PLAIN methods. Existing loopback tests now opt in explicitly, and a regression verifies default plaintext credentials are rejected before dialing.

Validation passed:
- `./vnew fmt -w vlib/net/imap/imap.v vlib/net/imap/imap_test.v`
- `./vnew -cc tcc vlib/net/imap/imap_test.v`
- `./vnew -cc tcc vlib/net/imap/response_test.v`
- `./vnew -cc tcc vlib/net/imap/seqset_test.v`
- `./vnew -cc tcc vlib/net/imap/utf7_test.v`
- `./vnew -d use_openssl -cc tcc vlib/net/imap/imap_test.v`
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed all four findings from the latest review in 3017c1acb0:

- OpenSSL validation loads default verify paths when no explicit CA bundle is configured, preserving platform-root behavior for existing `net.ssl` callers
- unknown extension parsing consumes subsequent top-level literals once a literal has established a structured response (including the ACL example)
- LIST/LSUB parsing consumes RFC 5258 LIST-EXTENDED data after the mailbox name
- bracketed IPv6 server names are normalized before TLS SNI/identity verification while retaining the original dial address

Validation passed:
- `./vnew fmt -w` on all touched IMAP `.v` files
- `./vnew -cc tcc vlib/net/imap/response_test.v`
- `./vnew -cc tcc vlib/net/imap/imap_test.v`
- `./vnew -cc tcc vlib/net/imap/seqset_test.v`
- `./vnew -cc tcc vlib/net/imap/utf7_test.v`
- `./vnew -d use_openssl -cc tcc vlib/net/imap/imap_test.v`
- `./vnew -d use_openssl -cc tcc -e 'import net.ssl; mut c := ssl.new_ssl_conn(validate: true)!; c.shutdown()!'`\n- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed all four findings from the latest review in 1fa60c0f7a:

- SEARCH/SORT number parsing now consumes RFC 7162 `(MODSEQ n)` metadata instead of treating `(` as a number
- STATUS parses known base fields as 32-bit values but generically skips extension values such as RFC 8438 `SIZE 5000000000`
- RFC 4314 ACL/LISTRIGHTS/MYRIGHTS responses allow top-level literals after atom arguments while the free-text ambiguity safeguard remains for other unknown responses
- OpenSSL versions before 1.0.2 reject `validate: true` explicitly during construction, before allocation or connection, instead of failing every identity check after a handshake
- included `<openssl/x509v3.h>` so hostname/IP identity shims compile cleanly under strict C

Validation passed:
- `./vnew fmt -w vlib/net/imap/response.v vlib/net/imap/response_test.v vlib/net/imap/wire.v`
- `./vnew -cc tcc vlib/net/imap/response_test.v`
- `./vnew -cc tcc vlib/net/imap/imap_test.v`
- `./vnew -cc tcc vlib/net/imap/seqset_test.v`
- `./vnew -cc tcc vlib/net/imap/utf7_test.v`
- `./vnew -d use_openssl -cc tcc vlib/net/imap/imap_test.v`
- `./vnew -d use_openssl -cc tcc -e 'import net.ssl; mut c := ssl.new_ssl_conn(validate: true)!; c.shutdown()!'`\n- `./vnew -d use_openssl -cstrict -cc clang -e 'import net.ssl; mut c := ssl.new_ssl_conn(validate: true)!; c.shutdown()!'`\n- `git diff --check`

medvednikov:
@codex review

PR #28285, opened 2026-09-01
builder: drop the "source inputs changed" V3 fallback note

medvednikov:
### What

Removes the user-facing note

```
note: source inputs changed before the stable compiler retry completed; the V3 fallback report was not submitted.
```

### Why

When V3 fails and the stable compiler succeeds from a slightly different source snapshot, V discards the unverified V3 fallback report. Previously it also printed the note above, which showed up on ordinary builds (e.g. a plain `v run .` of a hello world program) and was confusing noise for end users.

### How

The staged fallback report is already dropped by the `at_exit` cleanup when it is left unconsumed in the `.changed` branch, so no user-facing note is needed. This removes the `eprintln` (and the now-empty `discard_unverified_v3_fallback_report` helper) and updates `test_macos_v3_discards_fallback_report_when_native_input_changes` to assert the note is no longer emitted (the report is still silently discarded and the build still succeeds).

### DIFF
--- a/cmd/v/macos_v3_test.v
+++ b/cmd/v/macos_v3_test.v
@@ -1737,7 +1737,8 @@ exec "\$REAL_CC" "\$@"
 		exit_code := process.code
 		process.close()
 		assert exit_code == 0, compiler_output
-		assert compiler_output.contains('source inputs changed'), compiler_output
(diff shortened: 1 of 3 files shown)


medvednikov:
@codex review

PR #28286, opened 2026-09-01
net.http: make HTTP/3 opt-in behind `-d http3` to avoid requiring OpenSSL

medvednikov:
## Problem

Building `veb` (or anything using `net.http`) on Linux without the OpenSSL development headers fails:

```
==> warning: tcc compilation failed, falling back to cc
builder error: Header file <openssl/ecdsa.h>, needed for module crypto.ecdsa was not found.
Please install the corresponding development headers.
```

The HTTP/3 client wired `net.quic` into `net.http` via **unconditional** `import net.quic` statements. Because `net.quic`'s TLS 1.3 stack pulls in `crypto.ecdsa`, which hard-requires `<openssl/ecdsa.h>`, the dependency chain

```
veb → net.http → net.quic → crypto.ecdsa → <openssl/ecdsa.h>
```

forced **every** `net.http` user (veb, the http client, and everything built on them) to require OpenSSL at build time — even for plain HTTP that never touches QUIC.

## Fix

Make HTTP/3 opt-in behind a `-d http3` build flag, using V's conditional file compilation (the same `_d_`/`_notd_` pattern veb already uses for SSL):

- Rename the 4 `net.quic`-importing files to `*_d_http3.v` (compiled only with `-d http3`).
- Add `transport_h3_notd_http3.v` — a stub (compiled when the flag is off) providing the minimal surface `transport.v` references: empty `H3MuxConn`/`H3DialCall` pool types and an `h3_round_trip` that returns a clear `not compiled in; rebuild with \`-d http3\`` error instead of dragging OpenSSL into every build.
- Gate the 4 h3 test files with `// vtest build: present_openssl?` and `// vtest vflags: -d http3`.
- Document the `-d http3` requirement on `Request.enable_http3`.

No h3 logic changes — the source files are pure renames, so HTTP/3 behavior is unchanged when the flag is set.

## Verification

- `net.http` and `veb` compile **without** `-d http3`; generated C has zero quic/ecdsa/openssl references.
- An `enable_http3: true` request without the flag fails fast with the clear error message.
- **With** `-d http3`, the whole module (quic + openssl) compiles and all four h3 tests type-check.


medvednikov:
@codex review

PR #28289, opened 2026-09-01
v3: speed up FastC self-host generation with work-stealing

medvednikov:
## What

Replace the static byte-weighted **chunk-per-worker** split in the parallel per-file generator (`fastc_generate_file_outputs`) with a lock-free **work-stealing queue**. Every worker — including the main thread — atomically claims the next file from a largest-first order until the queue drains, so faster cores process more files instead of each worker draining a fixed chunk whose slowest core sets the wall time.

## Why

On a heterogeneous CPU (e.g. Apple Silicon's 6 performance + 12 efficiency cores) the static split left the slowest efficiency core — handed an average-size chunk — setting the wall time while performance cores idled; per-file generation scaled only ~8.4× on 18 threads. Work-stealing lets every core keep pulling files until the queue is empty, so they all finish together.

## Results

Measured on the `v3.v` self-host generation (173 files / 65.6k lines), best-of-100, interleaved, `-prod -prealloc`:

| | Generation | Throughput |
|---|---|---|
| before | ~24.7 ms | ~2.66M LOC/s |
| after | ~23.2 ms | ~2.83M LOC/s |

**~6–7% faster, byte-identical generated C** (outputs are restored to file order, so the scheduler change is not observable in the output).

The shared counter uses the `__atomic_fetch_add` compiler builtin (no header or library), so it remains available in both the gen/c host build and the FastC self-host output — verified that FastC emits `__atomic_fetch_add(&queue->next, 1, 0)` for its own source, and that 5 consecutive self-host generations are byte-identical (race-free).

## Validation

- Byte-identical C on `vlib/v3/v3.v` vs the previous scheduler ✓
- 5 consecutive self-host generations byte-identical (race check) ✓
- `examples/hello_world.v` compiles and runs in both parallel and `VJOBS=1` modes ✓
- Serial `-d v3_no_parallel` build compiles and runs ✓
- `fastc_test.v` regression subset (incl. the new `test_fastc_file_generation_order_is_largest_first`) passes ✓

## Also

Harden three `fastc_test.v` cases that passed a reference-typed local by address (`prefs: &prefs`, where `prefs` is already `&Preferences`). The redundant `&` makes the gen/c host heap-box the local and emit a member access with one-too-few dereferences (a codegen-layout-sensitive miscompile that any change to the module can trip). Passing the pointer directly (`prefs: prefs`) is both correct and robust.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -8191,10 +8191,11 @@ pub fn run(args []string) {
 		// allocation-heavy phases) — so compiler builds default to it.
 		// -no-prealloc opts out (also restores tcc linking: tcc has no
 		// thread-local storage support, so prealloc builds link with cc).
-		// FastC output is always compiled by bundled TinyCC, where the arena
-		// pointer cannot be thread-local; a FastC compiler generation spawns
-		// worker threads, so it must use plain thread-safe malloc instead.
-		if !no_prealloc && 'prealloc' !in user_defines && backend != 'fastc' {
+		// The FastC backend honors it too: it emits the arena root
+		// `g_memory_block` as per-thread storage (a pthread key under bundled
+		// TinyCC, which lacks thread-local storage), so its worker-thread
+		// generations bump-allocate safely (see fastc_write_prealloc_tls_global).
+		if !no_prealloc && 'prealloc' !in user_defines {
 			user_defines << 'prealloc'
 		}
 	}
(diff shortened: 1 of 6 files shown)


medvednikov:
@codex review

medvednikov:
**Manual review** (automated reviewer is off).

Reviewed the work-stealing scheduler: verified atomic-counter uniqueness/bounds (no OOB, no off-by-one on `order[claimed]`), no data races (workers mutate only the shared atomic counter + thread-local `outputs`; the shared array is written on the main thread after joins), correct worker count/termination, and matching per-platform atomics (nix `__atomic_fetch_add` relaxed / Windows `InterlockedExchangeAdd`, both returning the previous value). The `prefs: &&Preferences` → `prefs: prefs` test fix is correct.

No correctness issues found. ✅

PR #28291, opened 2026-09-02
v3: speed up FastC self-host generation ~1.8x (TCC) and ~1.3x (-prod)

medvednikov:
Speeds up FastC self-host generation of `vlib/v3/v3.v` (173 files, ~66K lines) on an M5 Max (6 fast + 12 slower cores):

| build | master | this PR |
|---|---|---|
| `-prod -gc none -prealloc` standalone driver (`fastc parse+gen`) | 26.5-27.9 ms (~2.5M loc/s) | 20.1-21.4 ms (~3.1-3.3M loc/s) |
| `-prod -gc none -prealloc` `cmd/v` (`-show-timings`) | 28.6-30.9 ms | 21.4-22.2 ms |
| TCC-built self-host generation (`FASTC_BENCH=1`) | 133 ms (~490K loc/s) | 74 ms (~895K loc/s) |

### Parallelism
- Parallel constant-initializer pass with an exactness rule: a file whose constant dependencies are all declared in that same file is accepted from the parallel pass, any other file is re-parsed in order.
- Fragment cut points are scanned in parallel.
- Per-file generation, constant parsing and the declaration/signature collection passes claim work from the shared atomic counter (`FastcGenQueue`, largest files first); partials are merged in file order, so the output is unchanged.

### Less per-file work
- Per-file memos for name, type-key and function-key lookups.
- `@[direct_array_access]` is honored for string and array indexing (scanner hot loops).
- One file record per scanner pass instead of a `FileSet`; shared module dependency order; pre-sized stitch buffers; incremental position tracking when ordering composite definitions.
- The first parallel pass records which declaration keywords (`interface`, `$if`, type keywords, generic `fn` syntax) each file mentions, and later collection scans skip files that cannot contain what they look for. Discovery waves now only scan imports.
- Allocation-free target-suffix checks in `pref`; faster decimal/hex digit loop in the scanner.

### Self-host runtime
- Map keys are hashed one 64-bit word at a time (the preamble hash was one byte per call, and TinyCC does not inline).
- `?`/`!` payload boxes are bump-allocated from a per-thread chunk instead of `malloc` (they were never freed anyway).
- `MultiReturn` slots are no longer zero-filled, and **components larger than 32 bytes are boxed**: a returned `map` (~120 B) used to overflow the 64-byte slot, which the zero-fill had masked. Repeating generation in-process (`FASTC_BENCH_LOOP`) crashed before this fix.

### Other
- The V3 C backend no longer emits prototypes for compiler builtins (`__atomic_*`, `__builtin_*`, `__sync_*`); clang rejects redeclaring `__atomic_fetch_add` (from the work-stealing generator's `fn C.__atomic_fetch_add`). Note that on master a V3 build of `fastc_test.v` on macOS still fails afterwards with unrelated `strconv__unknown` / `__order_snapshot_1` C errors; the test results below were obtained with the pre-#28277 compiler binary, which builds it.
- `FASTC_BENCH_PHASES=1`, `FASTC_BENCH_FILES=1` and `FASTC_BENCH_LOOP=N` diagnostics (documented in the v3 README).

### Verification
- Generation 2 and 3 compilers emit byte-identical C; serial (`V3_FASTC_NO_PARALLEL=1`) and parallel output are identical; the new generator's output matches the old generator's on the same input at every step.
- `fastc_test.v` (built with the pre-#28277 compiler binary, see above): no assertion failures, only the pre-existing panic in `enum_literal_map_lowering.v` that master shows too; scanner, token, pref, fastcdriver and gen/c preamble tests pass.

### DIFF
--- a/vlib/builtin/map.c.v
+++ b/vlib/builtin/map.c.v
@@ -18,7 +18,7 @@ fn fast_string_eq(a string, b string) bool {
 }
 
 fn map_hash_string(pkey voidptr) u64 {
-	key := *unsafe { &string(pkey) }
+	key := unsafe { &string(pkey) }
 	// XTODO remove voidptr cast once virtual C.consts can be declared
 	return C.wyhash(key.str, u64(key.len), 0, &u64(voidptr(C._wyp)))
 }
--- a/vlib/v3/README.md
+++ b/vlib/v3/README.md
@@ -154,6 +154,25 @@ compiler uses the small `v3.fastcdriver` entry point and can build further FastC
 the flat AST or conventional C backend. Set `V_MACOS_V3_NO_FALLBACK=1` while validating a chain to
 turn any attempted compatibility fallback into a hard failure.
 
+Set `FASTC_BENCH=1` when running a FastC self-host compiler to print the generation time and
+`loc/s` for its input (`FASTC_BENCH_REPEAT=N` reports the best of N child runs). `FASTC_BENCH_PHASES=1`
+prints the time of every generation phase, `FASTC_BENCH_FILES=1` the generation time of every source
+file, and `FASTC_BENCH=1 FASTC_BENCH_LOOP=N` repeats generation N times in-process so an external
+sampler can profile it. The compiler's own C preamble and runtime (hashing, option boxing, tuple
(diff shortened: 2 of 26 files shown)


medvednikov:
@codex review

PR #28293, opened 2026-09-02
v3: make `int` 64-bit on 64-bit targets (i64), 32-bit on 32-bit (i32)

medvednikov:
## What

Makes v3's `int` a platform-width type: it lowers to **`i64` on 64-bit targets** and **`i32` on 32-bit**, chosen from `target.pointer_bits`, across both the **C** and **FastC** backends. Previously `int` was always C `int` (32-bit).

## Design

`int` stays a *distinct* type from `i64` (same source name, its own methods) — only the emitted **C spelling** changes. This keeps `int` and `i64` from collapsing their `str`/`hex`/… method namespaces. The change activates V's existing `$if new_int ?` builtin guards on 64-bit, so `max_int`/`min_int` become i64 bounds and `int.str` formats via i64.

## Changes

- **types**: emission-only lowering in `c_type`; literal-overflow and integer-range checks use the platform width; `platform_int_c_type` + `set_platform_int_bits`.
- **gen/c**: `FlatGen.int_ct` drives the hand-written `[]int` runtime helpers (index/last_index/contains/sort/fixed-array-contains) and the `int__str` / `array__push_many` / `v3_int_zpad` prototypes and the IError `code()` dispatch. **C-extern declarations** (`fn C.foo(x int) int`) keep C `int` for ABI (V converts at the call boundary). `[sizeof(int)]u8` fixed-array typedef names are kept consistent with their references.
- **gen/fastc**: parallel emission-only lowering at the C-write sites.
- **driver / pref**: pins the width from the target before checking and codegen.
- **tests**: codegen expectations updated to `i64` where appropriate; genuine C `int` (libc / pthread / `fn C.*` externs / `int main`) is left unchanged.

## Validation

- **C backend runtime-correct**: `sizeof(int) == 8`, `2e9 + 2e9 == 4e9` (no 32-bit overflow), `int.str`/arrays/`sort`/`Optional`/generics/interfaces all work; a 32-bit target emits `i32`.
- **FastC self-host reaches a gen2 == gen3 fixpoint** and runs.

### DIFF
--- a/examples/sokol/sounds/melody.v
+++ b/examples/sokol/sounds/melody.v
@@ -9,14 +9,16 @@ mut:
 	gg      &gg.Context = unsafe { nil } // used for drawing
 }
 
-fn my_audio_stream_callback(mut soundbuffer &f32, num_frames int, num_channels int, mut acontext AppState) {
-	for frame := 0; frame < num_frames; frame++ {
+fn my_audio_stream_callback(mut soundbuffer &f32, num_frames i32, num_channels i32, mut acontext AppState) {
+	frame_count := int(num_frames)
+	channel_count := int(num_channels)
+	for frame := 0; frame < frame_count; frame++ {
 		t := int(f32(acontext.frame_0 + frame) * 0.245)
 		// "Techno" by Gabriel Miceli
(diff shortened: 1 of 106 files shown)


medvednikov:
@codex review

PR #28295, opened 2026-09-02
v3: fix self-host segfault in transform call-operand re-dispatch

medvednikov:
## Problem

```
$ ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -nocache -building-v -o v4 v3.v
  markused                37.16 ms     1256 MB RSS ...
zsh: segmentation fault  ./v3 -nocache -building-v -o v4 v3.v
```

The crash is a stack overflow from an unbounded `transform_call_expr` self-recursion (~6300 frames):

```
frame #1: v3__types__TypeChecker_resolve_type_uncached
...
frame #9:  v3__transform__Transformer_transform_call_expr + 2848
frame #10: v3__transform__Transformer_transform_call_expr + 3168
frame #11: v3__transform__Transformer_transform_call_expr + 3168   <- x6276
```

## Cause

A call with a value `match`/`if` operand is rebuilt over the materialized operands and re-dispatched over the rebuilt node. `transform_value_operand` only materializes a branch when it has a usable value type; otherwise it falls back to plain `transform_expr`, which rebuilds the branch with the same shape under a fresh id. The re-dispatch then saw a *changed but still a branch* operand, rebuilt the call again, and recursed forever.

The self-host trigger is in `vlib/v3/transform/sum.v`:

```v
field_sel := t.make_selector_op(new_expr, field, field_typ, if expr_type.starts_with('&') {
    .arrow
} else {
    .dot
})
```

Inside a clone the checker has no recorded type for the enum-shorthand `if`, so it never materializes.

## Fix

`materialize_value_branch_operand` makes the re-dispatch protocol require progress: when the lowering leaves a branch behind, it drops the prelude the failed attempt queued and returns the operand unchanged, so the caller falls back to its ordinary operand lowering. Applied at every rebuild-and-re-dispatch site — call receiver, callee, arguments, infix operands and cast operand.

Two pre-existing bugs sat behind the crash and also blocked the self-host build (they are only reachable once transform completes):

- **Module qualifier snapshotted as a value.** A module/type-qualified callee (`os.abs_path(...)`, `flat.node_payload(...)`) is a selector whose base names a compile-time namespace, not a value. When a later argument hoisted a value branch, the source-order guards snapshotted that base into a temp and emitted `unknown __order_snapshot_0 = os;`.
- **`for k, v in <&map>` bound the value by reference.** Only `for k, mut v in m` binds by reference, and C generation declares the binding by reference on that same condition. A container that is merely a map reference (`m &map[string]bool`) still binds a plain value copy, so typing it `&V` made every use emit a dereference of a non-pointer local (`if (!*is_used)`).

## Verification

With all three fixed:

- `v3 -nocache -building-v -o v4 v3.v` completes, and the following `v4 -> v5` generation completes too.
- `v3 -nocache -selfhost -o v4 v3.v` completes (on master it segfaults in transform in this mode as well).

`vlib/v3/tests/selfhost_transform_regression_test.v` adds a minimal repro for each of the three, all of which segfault or fail C compilation on master:

- an untyped enum-shorthand `if` argument in a generic clone (segfault),
- `os.abs_path(pick(name) or { if ... })` (`use of undeclared identifier 'os'`),
- `for name, is_used in used` over `used &map[string]bool` (`indirection requires pointer operand ('int' invalid)`).

## Known remaining, not addressed here

`-selfhost` with the module cache enabled (i.e. without `-nocache`) still crashes later, in cgen: `parse_type_cache_put` -> `map_set` writes through a corrupt map, so `pre_tc.type_cache` is dangling on that path. It is pre-existing and unreachable on master (transform crashes first), and independent of these changes — `-nocache` builds are unaffected.

### DIFF
--- a/vlib/v3/tests/selfhost_transform_regression_test.v
+++ b/vlib/v3/tests/selfhost_transform_regression_test.v
@@ -43,7 +43,38 @@ fn selfhost_regression_run(name string, source string) string {
 // `transform_expr`, which rebuilds the branch unchanged in shape. The re-dispatch used to see a
 // changed-but-still-a-branch operand and recurse forever, overflowing the stack.
 fn test_untyped_enum_branch_argument_in_generic_clone_terminates() {
-	out := selfhost_regression_run('untyped_enum_branch_arg', '\nenum Op {\n\tdot\n\tarrow\n}\n\nstruct Sel {\nmut:\n\tn int\n}\n\nfn (mut s Sel) make_selector_op(base int, field string, typ string, op Op) string {\n\ts.n++\n\treturn "${base}-${field}-${typ}-${op}"\n}\n\nfn (mut s Sel) wrap[U](v U, expr_type string) string {\n\tbase := s.n + 1\n\tfield := "${v}"\n\tfield_typ := "x${field}"\n\treturn s.make_selector_op(base, field, field_typ, if expr_type.starts_with("&") {\n\t\t.arrow\n\t} else {\n\t\t.dot\n\t})\n}\n\nfn main() {\n\tmut s := Sel{}\n\tprintln(s.wrap(1, "&int"))\n\tprintln(s.wrap("s", "int"))\n}\n')
+	out := selfhost_regression_run('untyped_enum_branch_arg', 'enum Op {
+	dot
+	arrow
+}
+
+struct Sel {
+mut:
+	n int
+}
+
+fn (mut s Sel) make_selector_op(base int, field string, typ string, op Op) string {
+	s.n++
+	return "\${base}-\${field}-\${typ}-\${op}"
+}
+
+fn (mut s Sel) wrap[U](v U, expr_type string) string {
+	base := s.n + 1
+	field := "\${v}"
+	field_typ := "x\${field}"
+	return s.make_selector_op(base, field, field_typ, if expr_type.starts_with("&") {
+		.arrow
+	} else {
+		.dot
+	})
+}
+
+fn main() {
+	mut s := Sel{}
+	println(s.wrap(1, "&int"))
+	println(s.wrap("s", "int"))
+}
+')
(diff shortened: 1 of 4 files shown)


medvednikov:
@codex review

medvednikov:
Fixed in 9c00e83d29 — the finding was correct and reproducible.

`call_selector_base_is_namespace` returned early for every non-`.ident` base, so the `.selector` branch of `static_assoc_fn_name` (which is what recognizes `mod.Type.fn(...)`) was never consulted. Confirmed with a `mod.Type.make(<value branch>)` call:

```v
fn build(n int) shapes.Box {
    return shapes.Box.make(pick(n) or {
        if n == -1 { 7 } else { 9 }
    })
}
```

which generated exactly the invalid snapshot you predicted:

```c
void __order_snapshot_0 = shapes__Box;
...
void__make(__order_snapshot_0, ...)
```

```
src.c:5899:7: error: variable has incomplete type 'void'
src.c:5899:28: error: unexpected type name 'shapes__Box': expected expression
```

Applied your suggested shape: the `C`/import checks stay `.ident`-only, and `static_assoc_fn_name` is attempted for identifier and selector bases alike. Added `test_module_qualified_static_assoc_call_with_branch_argument` as a two-module fixture covering it — it fails on master and passes here.

While adding it I found the three regression tests in the first commit never compiled: their embedded V sources used unescaped `${...}`, which the outer test file interpolated (`undefined ident: base`). Fixed too; the file now runs clean:

```
OK [2/5] test_untyped_enum_branch_argument_in_generic_clone_terminates()
OK [3/5] test_module_qualified_call_with_branch_argument()
OK [4/5] test_for_in_over_map_reference_binds_value()
OK [5/5] test_module_qualified_static_assoc_call_with_branch_argument()
Summary: 17 passed, 17 total.
```

Re-verified after the change that `v3 -nocache -building-v -o v4 v3.v` and the following `v4 -> v5` generation both still complete.

PR #28297, opened 2026-09-02
v3: fix vfmt reformatting valid code (#28292)

### LINKED ISSUE #28292: New vfmt questions
### Describe the bug

I just started using a new `v fmt` (0.5.2?) and already `v fmt` codebase just started to be mofified via `v fmt`.

Below i'll be posting (via update of the message content) what i found

1. additional braces
  ```diff
    shorter, longer := if var1.len <= var2.len {
  -		var1.len, var2.len
  +		{
  +			var1.len
  +			var2.len
  +		}
    } else {
  -		var2.len, var1.len
  +		{
  +			var2.len
  +			var1.len
  +		}
    }
  ```
1. blank line
  ```diff
   // ..
  -
   module name
  ```
  1. array thing
  ```diff
      SomeStruct{
  -			expected: [
  -				SomeStruct2({
  -				}),
  -			]
  +			expected: [SomeStruct2({
  +			})]
      },
  ```
1. long line break
  ```diff
  -		expected :=
  -			case.expected.map(a lot of args..).sorted_with_compare(fn)
  +		expected := case.expected.map(a lot of args ..).sorted_with_compare(fn)
  ```
  1. `pub` shuffle
  ```diff
  -pub:
  +
    // docs...
  +pub:
    field_name   StructName
  ```
1. long line break 2
  ```diff
  -		assert found.some_long_field_name_lalalala1 == actual.some_long_field_name_lalalala1
  -			&& found.some_long_field_name_lalalala2 == actual.some_long_field_name_lalalala2
  +		assert found.some_long_field_name_lalalala1 == actual.some_long_field_name_lalalala1 && found.some_long_field_name_lalalala2 == actual.some_long_field_name_lalalala2
  ```
  1. multi-line query break

  ```diff
  -		query_sql := '
  -			SELECT
  -				*
  -			FROM
  -				${table}
  -			;
  -		'
  +		query_sql := '\n\t\t\tSELECT\n\t\t\t\t*\n\t\t\tFROM\n\t\t\t\t${table}\n\t\t\t;\n\t\t'
  ```

### Reproduction Steps

simply run
```
v fmt dir
```

### Expected Behavior

code keeps the same

### Current Behavior

changes appears

### Possible Solution

_No response_

_No response_

### V version

V 0.5.2 0d40805

Linux

medvednikov:
Fixes every item reported in #28292, where `v fmt` rewrote already valid code.

All of these are in the v3 formatter (`vlib/v3/gen/v/gen.v`), which is what `v fmt` runs today.

### 1. `a, b` branch values got wrapped in a `{}` block

The parser groups a comma expression statement into a synthetic `comma_exprs` block. The formatter printed that block's braces:

```v
shorter, longer := if var1.len <= var2.len {
	var1.len, var2.len          // formatted to `{ var1.len \n var2.len }`
} else {
	var2.len, var1.len
}
```

The result does not compile. The same happened for `or { '', '' }`, for `match` branch values and for `rlock a, b, c { x.a, y.a, z.a }`. The block is now printed as the comma list it was parsed from.

### 2. Multiline interpolated strings were escaped onto one line

```v
query_sql := '
	SELECT *
	FROM ${table}
'
// became: query_sql := '\n\tSELECT *\n\tFROM ${table}\n'
```

Plain multiline strings were already kept verbatim; interpolated ones now are too. A line break that only wraps an embedded `${...}` expression is still reflowed, so `println('x ${f(a,\n\tb)}')` still joins.

### 3. Blank line after a leading comment was dropped

```diff
 // Copyright header
-
 module name
```

### 4. Doc comments were hoisted above `pub:` / `mut:`

```diff
 pub struct Foo {
-
-	// docs...
-pub:
+pub:
+	// docs...
 	field_name string
 }
```

Comments that really do sit before the specifier still stay before it.

### 5. An expanded array collapsed after a one-line sibling

```diff
     expected: [
-        SomeStruct2({
-        }),
-    ]
+    expected: [SomeStruct2({
+    })]
```

`array_breaks` caches one layout decision per nesting depth and is only cleared when the depth returns to 0, so every later array at that depth inherited the decision of the first one. The cache is deliberate for the width heuristic — it keeps a row of small nested arrays such as `[[1, 2], [3, 4], [5, 6]]` uniform rather than breaking at whichever element happens to cross the column limit — so only the source-driven half of the decision is now made per literal.

### 6. Line breaks before `&&` / `||` and before an assignment value were joined

```diff
-	assert found.some_long_field_name_1 == actual.some_long_field_name_1
-		&& found.some_long_field_name_2 == actual.some_long_field_name_2
+	assert found.some_long_field_name_1 == actual.some_long_field_name_1 && found.some_long_field_name_2 == actual.some_long_field_name_2
```

The break the source had is now kept, for `&&`/`||` continuations and for an assignment value written on its own line. Breaks are not added where the source had none, so single-line conditions stay single-line.

Finding the operand boundaries needs `leftmost_source_start`/`rightmost_source_end` rather than the node span, because a few nodes (`!in`, `!is`) are synthesized around their operand and carry the position of a later token.

### Validation

Formatting all 6249 `.v` files under `vlib`, `cmd`, `examples` and `tools` with both the `master` and the patched `vfmt`, and comparing against the committed source:

|                                                    | before | after |
| -------------------------------------------------- | -----: | ----: |
| files whose formatting differs from the source      |   3334 |  2959 |
| files where a second `vfmt` pass changes the first  |    362 |   345 |

No file regressed in either column. Diffing old `vfmt` output against new, every difference is one of the fixes above (plus a handful of array literals that no longer get expanded with a trailing comma, because the line now fits once the `&&` break is kept).

`vlib/v3/gen/v/gen_test.v` gets 14 new regression tests, 11 of which fail on `master`. Formatter timing on `vlib/v/gen/c/cgen.v` is unchanged (0.15s before and after).

The two `vlib/v3/gen/fastc` test failures on this branch also fail on `master` and are unrelated.

Fixes #28292

### DIFF
--- a/vlib/v3/gen/v/gen.v
+++ b/vlib/v3/gen/v/gen.v
@@ -1213,13 +1213,19 @@ fn (mut g Gen) array_literal(id flat.NodeId) {
 		return
 	}
 	g.array_depth++
+	first := g.a.node(children[0])
+	// An array whose source starts on the line below `[` keeps that layout even
+	// when an earlier array at this nesting depth was written on one line.
+	source_break := g.source_line(first.pos.offset) > g.source_line(n.pos.offset)
 	if g.array_depth > g.array_breaks.len {
-		first := g.a.node(children[0])
+		// The width heuristic is decided once per nesting depth, so a row of
+		// small nested arrays stays uniform instead of breaking at whichever
+		// element happens to cross the limit.
 		first_width := g.array_expr_width(children[0])
-		g.array_breaks << g.source_line(first.pos.offset) > g.source_line(n.pos.offset)
+		g.array_breaks << source_break
 			|| (first_width > 0 && g.output_line_len() + first_width > formatter_array_first_break)
 	}
-	line_break := g.array_breaks[g.array_depth - 1]
+	line_break := source_break || g.array_breaks[g.array_depth - 1]
 	mut indented := false
 	for i, child in children {
 		if i == 0 {
--- a/vlib/v3/gen/v/gen.v
+++ b/vlib/v3/gen/v/gen.v
@@ -611,10 +611,47 @@ fn (mut g Gen) emit_blank_line_between(previous flat.NodeId, current flat.NodeId
 	}
 	prev := g.a.node(previous)
 	cur := g.a.node(current)
-	if !prev.pos.is_valid() || !cur.pos.is_valid()
-		|| !g.source_has_blank_line_between(prev.pos.end, cur.pos.offset) {
+	if !prev.pos.is_valid() || !cur.pos.is_valid() {
 		return
 	}
+	mut limit := cur.pos.offset
+	if g.comment_i < g.comments.len {
+		// a comment in the gap keeps the blank lines that follow it, so only the
+		// part of the gap before that comment decides the separator here
+		comment_start := g.comments[g.comment_i].pos.offset
+		if comment_start >= prev.pos.end && comment_start < limit {
+			limit = comment_start
+		}
+	}
+	if !g.source_has_blank_line_between(prev.pos.end, limit) {
+		return
+	}
+	g.write_blank_line()
+}
+
+// emit_blank_line_after_comment writes a blank line when the source kept one
+// between the comment that was just emitted and the code starting at `limit`.
+fn (mut g Gen) emit_blank_line_after_comment(limit int) {
+	gap := g.source_span(g.source_end, limit) or { return }
+	// Only the blank run directly after the comment counts: `limit` is a node
+	// position, which for some statements starts past their first token.
+	mut newlines := 0
+	for c in gap {
+		if c == `\n` {
+			newlines++
+		} else if c !in [` `, `\t`, `\r`] {
+			break
+		}
+	}
+	if newlines < 2 {
+		return
+	}
+	g.write_blank_line()
+}
+
+// write_blank_line ends the current line and adds an empty one, unless the
+// output already ends with a blank line.
+fn (mut g Gen) write_blank_line() {
 	if !g.on_newline {
 		g.writeln('')
 	}
@@ -640,7 +677,13 @@ fn (mut g Gen) stmt(id flat.NodeId) {
 		eprintln('stmt ${n.kind} | pos: ${n.pos.offset}')
 	}
 	stmt_start, stmt_end := g.stmt_source_span(n)
+	comment_i := g.comment_i
 	g.emit_comments_before(stmt_start)
+	if g.comment_i > comment_i {
+		// keep a blank line the source had between the last comment and this
+		// statement (e.g. a file header comment above `module`)
+		g.emit_blank_line_after_comment(stmt_start)
+	}
 	g.source_end = int_max(g.source_end, stmt_start)
 	match n.kind {
 		.module_decl {
@@ -833,9 +876,7 @@ fn (mut g Gen) expr(id flat.NodeId) {
 			g.write('.${n.value}')
 		}
 		.infix {
-			g.expr(g.a.child(n, 0))
-			g.write(' ${op_str(n.op)} ')
-			g.expr(g.a.child(n, 1))
+			g.infix_expr(id)
 		}
 		.prefix {
 			g.prefix_expr(id)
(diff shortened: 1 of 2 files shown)


medvednikov:
@codex review

medvednikov:
Thanks — both findings reproduced and are fixed in 62c7ad4.

**1. Comment duplication in the raw-source path.** Confirmed. Reduced case:

```v
fn f(a int, b int) string {
	return 'line1
line2 ${a + /* keep */ b}
line3'
}
```

emitted `/* keep */` a second time after the string. `string_interp` now calls `g.skip_comments_before(n.pos.end)` before returning, matching the `struct_init` raw-source path.

One note on the suggested `n.pos.end + 1`: `pos.end` is already exclusive (it is `p.prev_tok_end`, just past the closing quote), so `n.pos.end` covers every comment inside the literal, while `+ 1` would additionally swallow a comment starting immediately after the closing quote. I used `n.pos.end` and added `test_formatter_keeps_comment_after_multiline_interpolation` to pin that down.

**2. Escaped `\n` misclassified as a physical newline.** Confirmed:

```v
fn f(a int, b int) string {
	return 'a\nb ${add(a,
		b)}'
}
```

was copied verbatim instead of reflowing the wrapped `${...}`.

I could not detect this from the literal children's source spans as suggested: those nodes are built with `p.add_val_id()` (parser.v:455), which falls through to `add_node`'s `p.current_pos()` default, so a segment carries the position of whatever token is current when it is created, not its own text. Instead the check now scans the interpolation's own source span, skipping escapes and everything between `${` and its matching `}` (stepping over nested string/rune literals so their braces do not throw off the depth count). That is the same classification you asked for — physical source rather than decoded value — and it also replaces the old two-step `value.contains('\n')` + whole-span check with a single precise one.

**Tests:** 3 added (`comments_inside_multiline_interpolation_once`, `comment_after_multiline_interpolation`, `reflows_interpolation_wrap_when_text_only_has_escaped_newline`). The first and third fail without the fix.

**Re-validation.** I also re-ran the whole-tree sweep, and in doing so found my original baseline had been measured over a slightly different file list than the patched run, so the numbers in the PR body were not directly comparable. Both sides are now measured over the identical 6249-file list; the corrected table is in the description:

| | before | after |
| --- | ---: | ---: |
| files whose formatting differs from the source | 3334 | 2959 |
| files where a second `vfmt` pass changes the first | 362 | 345 |

Still zero regressions in either column. Only 2 files in the tree change output as a result of this round of fixes, both instances of finding 2.

medvednikov:
Item 3 of the report ("array thing") turned out to be reproducible after all — I had been building the wrong shape. Fixed in 7a41f7f, so the PR now covers all of #28292.

The trigger is a *sibling* array, not the element:

```v
x := [
	Outer{
		list: [Elem{
			a: 1
		}]
	},
	Outer{
		list: [
			Elem{
				a: 2
			},
		]
	},
]
```

The second `list` collapses to `[Elem{\n\ta: 2\n}]`, exactly the shape in the issue. `array_breaks` caches one layout decision per nesting depth and is only cleared when the depth returns to 0, so every later array at that depth inherits the first one's decision.

Worth noting for review: making the whole decision per-literal is *not* the right fix. I tried that first and it regressed `vlib/math/math_test.v`, `vlib/builtin/array_test.v` and `vlib/crypto/hmac/hmac_test.v`, because a row of small nested arrays would then be measured at whatever `output_line_len()` happens to be deep into a wrapped line:

```v
[-pi, 0], [-pi, inf(1)], [
	-pi,
	nan(),
], [f64(-0.0), inf(-1)], ...
```

The cache is load-bearing for that heuristic. So only the source-driven half of the decision is now per literal; the width heuristic stays shared per depth.

Net effect across the tree: the whole-tree counts are unchanged (2959 differing / 345 non-idempotent, still zero regressions against `master`), and two files move closer to their committed source — `vlib/crypto/pem/pem_test.v` drops from 88 to 12 differing lines, and `vlib/v3/transform/array.v` now honours the expanded `[` in `make_call_typed('drop_owned', [ ... ])`. Two more tests, one of which fails without the fix; the other pins the `[[1, 2], [3, 4], [5, 6]]` case that the first attempt broke.

PR #28304, opened 2026-09-03
v3: never spill a namespace base into an ordering-guard temp

medvednikov:
A Linux user reported uncompilable C generated for `strconv`:

```
strconv__unknown __order_snapshot_1 = strings;
Array res = strconv__unknown__new_builder(__order_snapshot_1);
```

`strconv.format_es_old` calls `strings.new_builder(if p.len0 > fs.len { p.len0 } else { fs.len })`. The value `if` argument makes the call-operand ordering guards stabilize the preceding operand — here the callee base `strings`, which names an import, not a value. Spilling it declares a temp at whatever the checker recorded for a module identifier (`unknown`), so the generated C names both an undefined type and an undefined variable.

## Why a second layer

`call_selector_base_is_namespace` already keeps namespace bases out of the ordering guards, and its `is_import_alias_ident` arm is what covers this shape. I confirmed that by disabling only that arm — the identical error comes back. Notably the lexical module-call check does not match `strings.new_builder`, nor even `os.abs_path`.

So that one classifier is the only thing standing between a namespace base and invalid C, and it has now been extended twice after the fact (`unknown ... = os;` in #28295, `void ... = shapes__Box;` in the same PR).

This adds a backstop where the temp is actually declared: the ordering-snapshot helpers refuse a type of `unknown` or `void` — the two spellings a namespace base can carry — and leave the operand inline instead. That only forfeits a source-order guarantee for something that has no runtime storage, so it degrades to the pre-guard behaviour rather than breaking the build. With the classifier arm disabled, the backstop alone fixes both the `strings` and the `os` case.

An empty type stays allowed: the temp is then declared without one and C generation infers it from the initializer.

## It is a no-op on existing code

An instrumented build never takes the new branch while compiling `cmd/v`, `vlib/v3/v3.v`, or any `examples/*.v`, and the generated C for the reported program is byte-identical to before.

## Testing

- New regression test pins the reported program. It reproduces the exact error above when the classifier arm is removed, and passes with the fix.
- Full `vlib/v3/tests` A/B (264 files, V3 as the compiler, patched vs unpatched): same 30 pre-existing failures on both sides, no new ones.
- Verified on macOS/arm64, Debian 13 arm64 and Debian 13 amd64 (the reporter's arch).
- `v fmt -verify` clean.

Note for the reporter: the specific `strings.new_builder` failure was already fixed on master by #28295; their V predates it, so updating V resolves it without this PR. This change removes the way that class of bug reaches C generation at all.

### DIFF
--- a/vlib/v3/transform/ordering_guard_test.v
+++ b/vlib/v3/transform/ordering_guard_test.v
@@ -68,7 +68,12 @@ fn test_untyped_callee_base_is_not_spilled_into_an_ordering_temp() {
 		t.cur_module = 'main'
 
 		call, node := build_namespace_call(mut t, 'nsbase', base_type, 'abs_path')
-		assert !t.call_selector_base_is_namespace(t.a.child(&node, 0), 'abs_path', node.value), 'the classifier must miss this base, otherwise the backstop is not what is under test'
+		// `call_selector_base_is_namespace` takes the *base* of the callee selector, so unwrap the
+		// callee first — passing the callee itself would assert about the wrong node and could
+		// read false even for a base the classifier does recognize.
+		callee := t.a.nodes[int(t.a.child(&node, 0))]
+		assert callee.kind == .selector
+		assert !t.call_selector_base_is_namespace(t.a.child(&callee, 0), 'abs_path', node.value), 'the classifier must miss this base, otherwise the backstop is not what is under test'
 
 		t.transform_call_expr(call, node)
 		assert ordering_snapshot_decl_count(&t) == 0, 'a `${base_type}` callee base must not be spilled into an ordering temp'
--- a/vlib/v3/transform/ordering_guard_test.v
+++ b/vlib/v3/transform/ordering_guard_test.v
@@ -1,0 +1,145 @@
+module transform
+
+import v3.flat
+import v3.types
+
+// build_namespace_call assembles `<base>.<method>(if cond { 1 } else { 2 })`, where `base` is an
+// identifier the checker typed at `base_type`. The value `if` argument is what makes the
+// call-operand ordering guards look at the preceding operand — here the callee base.
+fn build_namespace_call(mut t Transformer, base_name string, base_type string, method string) (flat.NodeId, flat.Node) {
+	base := t.a.add_val(.ident, base_name)
+	t.set_node_typ(int(base), base_type)
+	callee := t.make_selector_op(base, method, 'string', .dot)
+
+	cond := t.a.add_val(.bool_literal, 'true')
+	then_val := t.a.add_val(.int_literal, '1')
+	else_val := t.a.add_val(.int_literal, '2')
+	branch_start := t.a.children.len
+	t.a.children << cond
+	t.a.children << then_val
+	t.a.children << else_val
+	arg := t.a.add_node(flat.Node{
+		kind: .if_expr
+		children_start: branch_start
+		children_count: 3
+		typ: 'int'
+	})
+
+	call_start := t.a.children.len
+	t.a.children << callee
+	t.a.children << arg
+	call := t.a.add_node(flat.Node{
+		kind: .call
+		children_start: call_start
+		children_count: 2
+		value: '${base_name}.${method}'
+	})
+	return call, t.a.nodes[int(call)]
+}
+
+fn ordering_snapshot_decl_count(t &Transformer) int {
+	mut n := 0
+	for stmt in t.pending_stmts {
+		node := t.a.nodes[int(stmt)]
+		if node.kind != .decl_assign || node.children_count == 0 {
+			continue
+		}
+		lhs := t.a.nodes[int(t.a.child(&node, 0))]
+		if lhs.kind == .ident && lhs.value.starts_with('__order_snapshot') {
+			n++
+		}
+	}
+	return n
+}
+
+// Direct coverage of the backstop. call_selector_base_is_namespace recognizes every namespace
+// spelling that reaches it in practice, so the base here is deliberately absent from `tc.imports`
+// and from the file-import table: the classifier genuinely misses it (asserted below), leaving
+// only the backstop to keep the spill out of the output. Removing the check in
+// transform_call_expr fails this test.
+//
+// The guards must leave such a base inline rather than declare a temp at a name no C declaration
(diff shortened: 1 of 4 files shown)


medvednikov:
@codex review

medvednikov (commented):
I found two issues worth addressing before I’d approve this:

1. **The backstop is broader than the namespace bug.** `ordering_snapshot_type_is_usable()` is applied inside both generic ordering-snapshot helpers, including `snapshot_transformed_expr_for_reuse()`. The known call sites of that helper are ordinary runtime values (membership operands, struct field values, array len/cap). Returning an expression inline when its type text is `unknown`/`void` drops the source-order guarantee these helpers exist to provide. `unknown` is evidence that type information is missing; it is not, by itself, proof that the expression is a compile-time namespace. If a legitimate runtime value is temporarily unresolved, a later hoisted value-branch prelude can mutate what that earlier operand observes. Please scope this fallback to an expression that is actually proven to be a namespace/callee base (or otherwise recover/prove the type) rather than keying generic snapshot behavior only off the type spelling. In particular, I don’t see a namespace path that needs the check in `snapshot_transformed_expr_for_reuse()`.

2. **The new regression test does not exercise the new fallback.** On current master, `call_selector_base_is_namespace()` / `is_import_alias_ident()` already classifies the `strings.new_builder(...)` callee base as a namespace, so the snapshot helper is never reached for `strings`. The added test therefore passes even if the new `ordering_snapshot_type_is_usable()` checks are removed. Since this PR’s behavioral change is explicitly a defensive second layer, please add direct coverage that fails without that layer (or refactor the backstop into something that can be exercised directly without disabling production classifier code).

I ignored CI/status checks as requested and reviewed the code paths only.

medvednikov:
Both points were right — thanks. Pushed `3a81091b8e` (branch also rebased onto master so it builds again after #28307).

**1. Scope.** Reverted both generic helpers to their original form; the check now sits at the one operand a namespace can occupy, the selector callee base in `transform_call_expr`.

Your framing turned out to understate it: `node_type` returns `unknown` for *anything* it cannot resolve, so there was no distinction at all between "typed unknown" and "no type recorded" — my original comment claiming an absent type stayed snapshotted was simply wrong. Keying off the spelling was therefore weak even at the callee base. It now tests positively:

- `void` is conclusive on its own, since no value carries it.
- `unknown` counts only for a bare identifier that names nothing in scope. A runtime receiver is always a binding, and a generic clone still records one where it has lost the expression type — so exactly the case you raised keeps its snapshot. That is asserted in `test_untyped_but_bound_receiver_is_still_spilled_into_an_ordering_temp` and in the classification test.

And you were right that no namespace path needed it in `snapshot_transformed_expr_for_reuse()` — with the check only at the callee base, the `strings` and `os` cases are both still fixed with `is_import_alias_ident` disabled.

**2. Coverage.** Added `vlib/v3/transform/ordering_guard_test.v`, which drives `transform_call_expr` with a base the classifier genuinely misses — the test asserts `!call_selector_base_is_namespace(...)` first, so it cannot silently stop testing the backstop — and requires that no ordering temp is declared. Removing the check in `transform_call_expr` fails it:

```
> assert ordering_snapshot_decl_count(&t) == 0
  Message: a `unknown` callee base must not be spilled into an ordering temp
```

No production classifier code is disabled. Companion cases pin that a typed receiver, an untyped-but-bound receiver, and a non-identifier base all keep their snapshots.

The selfhost test stays, with its comment corrected: it pins the reported program against a regression in the classifier (the path that actually resolves this shape), not against the backstop.

**Re-verified:** still a no-op on real code — an instrumented build never takes the branch while compiling `cmd/v`, `vlib/v3/v3.v` or any `examples/*.v`. `transform_parallel_test.v` fails on pristine master too, so it is pre-existing.

vlib/v3/transform/ordering_guard_test.v
```diff
@@ -0,0 +1,145 @@
...
+// can use (`unknown __order_snapshot_0 = os;`, `void __order_snapshot_0 = shapes__Box;`).
+fn test_untyped_callee_base_is_not_spilled_into_an_ordering_temp() {
+	for base_type in ['unknown', 'void'] {
+		mut a := flat.FlatAst.new()
+		mut tc := types.TypeChecker.new(&a)
+		mut t := new_transformer(mut a, &tc, map[string]bool{})
+		t.cur_module = 'main'
+
+		call, node := build_namespace_call(mut t, 'nsbase', base_type, 'abs_path')
+		assert !t.call_selector_base_is_namespace(t.a.child(&node, 0), 'abs_path', node.value), 'the classifier must miss this base, otherwise the backstop is not what is under test'
```
medvednikov: Nit: `call_selector_base_is_namespace` expects the selector **base** id, but `t.a.child(&node, 0)` is the call's callee selector. So this assertion can stay false even if the actual base would be classified as a namespace, and it does not quite prove the backstop is what the test exercises. Consider extracting the callee first, then passing `t.a.child(&callee, 0)`. The current `nsbase` setup still exercises the backstop, so I don't think this is blocking.

medvednikov (commented):
Re-review: the previous correctness concern is addressed. The backstop is now scoped to selector callee bases rather than the generic snapshot helpers, so unresolved runtime operands elsewhere keep their ordering guarantees. The added unit coverage also directly exercises the backstop and preserves snapshots for runtime receivers.

No blocking findings. One non-blocking test nit inline.

vlib/v3/transform/ordering_guard_test.v
medvednikov: Good catch — fixed in `e84fe6d941`. The test now unwraps the callee and passes `t.a.child(&callee, 0)`.

You were right that it was asserting about the wrong node. I checked how much it mattered: the old assertion happened to be false for a reason unrelated to the base — `call_selector_base_is_namespace` fell through to `static_assoc_fn_name`, which read the selector as the type identifier `nsbase.abs_path` and found no such type. So it was passing by accident and would have kept passing regardless of the base.

To confirm the corrected assertion actually has teeth, I registered `nsbase` in `tc.imports` — it now trips:

```
Message: the classifier must miss this base, otherwise the backstop is not what is under test
```

With the previous `t.a.child(&node, 0)` it stayed false in that same setup. So the guard against this test quietly ceasing to cover the backstop is real now, rather than nominal. Also re-confirmed the test still fails when the check in `transform_call_expr` is removed.

PR #28306, opened 2026-09-03
vfmt: fix non-idempotent blank line after a commented const (fix #28305)

### LINKED ISSUE #28305: v fmt -w is not idempotent: inserts an extra blank line after a commented const on each pass
## Describe the bug

`v fmt -w` is not idempotent for a top-level `const` declaration carrying a
trailing `//` comment, immediately followed by another `const` (no blank
line between them) in the original source. Each `fmt -w` pass inserts an
*additional* blank line between the two, instead of leaving an already-
inserted blank line alone — so the file keeps changing for 2 passes before
settling on a stable (but almost certainly unintended) result of **two**
consecutive blank lines.

## Reproduction

```v
module quic

const frame_type_crypto = u64(0x06)
const frame_type_stream_base = u64(0x08) // 0x08-0x0f, OFF/LEN/FIN bits in the low 3 bits
const frame_type_max_data = u64(0x10)
const frame_type_max_stream_data = u64(0x11)
```

Save as `repro.v`, then:

```
v fmt -w repro.v   # pass 1: reports "Reformatted file"
v fmt -w repro.v   # pass 2: reports "Reformatted file" AGAIN
v fmt -w repro.v   # pass 3: reports "Already formatted file" -- now stable
```

**After pass 1** (a single blank line is inserted after the commented
`const`):
```v
module quic

const frame_type_crypto = u64(0x06)
const frame_type_stream_base = u64(0x08) // 0x08-0x0f, OFF/LEN/FIN bits in the low 3 bits

const frame_type_max_data = u64(0x10)
const frame_type_max_stream_data = u64(0x11)
```

**After pass 2** (a *second* blank line is inserted -- this is the bug):
```v
module quic

const frame_type_crypto = u64(0x06)
const frame_type_stream_base = u64(0x08) // 0x08-0x0f, OFF/LEN/FIN bits in the low 3 bits


const frame_type_max_data = u64(0x10)
const frame_type_max_stream_data = u64(0x11)
```

Pass 3 (and every pass after) reports "Already formatted file" and makes
no further change -- `fmt` does converge, but to a state with two
consecutive blank lines between top-level `const`s, which doesn't match
this codebase's own convention elsewhere (single blank line, or none, is
the norm; I didn't find any other double-blank-line separator between
top-level consts in `vlib`).

## Expected behavior

`v fmt -w` should be idempotent: a single pass should reach the same
fixed point a second pass would reach, and that fixed point shouldn't
accumulate an extra blank line each time the trigger pattern (a
`//`-commented `const` immediately followed by another `const`) is
re-encountered.

## Real-world context

Found while running `v fmt -w` on `vlib/net/quic/` in
[vlang/v#28303](https://github.com/vlang/v/pull/28303) -- several files in
that directory needed 2 `fmt -w` passes to reach a stable result. This
minimal repro isolates the exact pattern from
`vlib/net/quic/frame.v:237-238` at the time of that PR.

## V version

```
$ v version
V 0.5.2
```
Built from `vlang/v` around commit `25f5ced70b3` (upstream/master,
2026-09-03).

## Environment

- OS: Windows 11
- Compiler backend: gcc (MinGW-w64)

> [!NOTE]
>

medvednikov:
Fixes #28305.

## Problem

A top-level `const` carrying a trailing `//` comment gained a blank line after it on every `v fmt -w` pass, so the file kept changing for two passes before settling on two consecutive blank lines:

```v
const frame_type_crypto = u64(0x06)
const frame_type_stream_base = u64(0x08) // 0x08-0x0f, OFF/LEN/FIN bits in the low 3 bits
const frame_type_max_data = u64(0x10)
```

## Cause

`expr()` ends with `emit_trailing_comments(n.pos.end)`, which flushes a same-line `//` comment and terminates the line (`write_comment` uses `writeln`). `const_decl` then wrote an unconditional `g.writeln('')` on top of that, producing a blank line.

That blank line then feeds back into `top_level`: it only keeps two consecutive `const`s adjacent when the *source* has no blank line between them, so once pass 1 wrote one, pass 2 added the separator *and* the stray newline — two blank lines.

## Fix

Guard the newline with `g.on_newline`, the same way `global_field` and `named_init_field` already do. The duplicated single-field / loop bodies of `const_decl` are folded into one `const_field` so the guard lives in a single place.

`vlib/v3/gen/arm64/linker.v` had the doubled blank lines committed (it was formatted by the buggy vfmt), so it is reformatted here.

## Validation

- New regression test in `vlib/v3/gen/v/gen_test.v`; it fails on `master` and passes here. `v test vlib/v3/gen/v/gen_test.v` and `v test cmd/tools/vfmt_test.v` pass.
- Whole-tree A/B sweep, two `v fmt` passes over all 6307 `.v` files in the repo with a `master` build and this build:
  - **24** files that were not idempotent under `master` are now stable (296 → 272), **0** newly unstable.
  - **18** more files now match their committed source, **1** newly differs — `vlib/v3/gen/arm64/linker.v`, whose committed content *is* the bug's output; it is reformatted in this PR.
  - 91 files change output, and in every one the only change is a removed blank line.
- The two `vlib/v3/gen/fastc` failures and `vlib/v/fmt/fmt_bin2v_test.v` fail identically on `master`.

### DIFF
--- a/vlib/v3/gen/arm64/linker.v
+++ b/vlib/v3/gen/arm64/linker.v
@@ -43,22 +43,17 @@ const csslot_requirements = u32(2)
 const csslot_cms_signature = u32(0x10000)
 const cs_adhoc = u32(0x2) // Ad-hoc signing flag
 
-
 const cs_hashtype_sha256 = u8(2)
 const cs_hash_size = 32 // SHA256 = 32 bytes
 
-
 const cs_page_size_arm64 = 16384 // Code signing page size for ARM64 macOS
 
-
 const cs_page_shift_arm64 = 14 // log2(16384)
-
 
 const o_wronly_creat_trunc = $if linux { 0x241 } $else { 0x601 }
 
 // ARM64 page size on macOS
 const page_size = 0x4000 // 16KB
-
 
 // Base address for executables
 const base_addr = u64(0x100000000)
--- a/vlib/v3/gen/v/gen.v
+++ b/vlib/v3/gen/v/gen.v
@@ -3265,34 +3265,28 @@ fn (mut g Gen) const_decl(id flat.NodeId) {
 	n := g.a.node(id)
 	g.emit_attrs(id)
 	pub_prefix := if n.op == .arrow { 'pub ' } else { '' }
-	fields := g.a.children_of(n)
-	if fields.len == 1 {
-		f := g.a.node(fields[0])
-		g.emit_comments_before(f.pos.offset)
-		g.write('${pub_prefix}const ${f.value}')
-		if f.children_count > 0 {
-			g.write(' = ')
-			g.expr(g.a.child(f, 0))
-		} else if f.typ.len > 0 {
-			g.write(' ${g.type_text(f.typ)}')
-		}
-		g.writeln('')
-		g.source_end = int_max(g.source_end, f.pos.end)
-		return
+	for fid in g.a.children_of(n) {
+		g.const_field(fid, pub_prefix)
 	}
-	for fid in fields {
-		f := g.a.node(fid)
-		g.emit_comments_before(f.pos.offset)
-		g.write('${pub_prefix}const ${f.value}')
-		if f.children_count > 0 {
-			g.write(' = ')
-			g.expr(g.a.child(f, 0))
-		} else if f.typ.len > 0 {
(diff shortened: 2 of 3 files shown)


medvednikov:
@codex review

PR #28308, opened 2026-09-03
v3: header-free FastC self-host C, reachability prune, concurrent TinyCC units, in-process signing

medvednikov:
Reduces what TinyCC has to compile for a FastC self-host build, then runs TinyCC concurrently and takes Apple's `codesign` out of the loop. Everything is verified with the self-host chain (`v4 → v5 → v6`, `v5.c == v6.c`, also with `VJOBS=1` and `V3_FASTC_NO_PARALLEL=1`, every output passing `codesign --verify --strict`) and the FastC unit tests (only the 4 known `test_selfhost_*` failures and the known panic remain).

## No `#include` in the generated C
- `gen/fastc/c_abi.v` holds, per target (64-bit macOS, glibc Linux x86_64/aarch64), the C library declarations the emitted code uses: integer/size typedefs, `struct stat`/`dirent`/`timespec`/`timeval`/`tm` layouts, pthread type sizes, `fd_set` with `FD_*` macros, the `O_*`/`S_*`/errno/`SEEK_*`/`PROT_*`/`MAP_*`/... values, `stdin`/`stdout`/`stderr`/`errno`/`environ`, and exact prototypes for ~100 functions (with `__asm__` labels for `realpath$DARWIN_EXTSN` and the x86_64 macOS `$INODE64` family).
- A self-host build replaces the preamble's header block with that prelude, leaves the `#include` lines of V sources out of the output, and inlines V's own C helper headers (`vlib/os/execute_capture_nix.h`, `vlib/sync/thread_helper.h`, ...) after the prelude with their include lines stripped (`thread_helper.h` gained a `V_FASTC_NO_HEADERS` guard). Other targets keep the headers.
- The build passes `-Werror=implicit-function-declaration` and no longer passes `-w` (which silenced it in either order): a C function missing from the table fails the build instead of being called through an implicit `int` prototype that truncates pointer results (that is how a missing `fdopen` crashed a self-host in `flockfile` during this work).
- `c_abi_test.v` compiles the prelude (with a `v_abi_` prefix) next to the real headers with the bundled TinyCC and checks every type size, struct field offset, macro value and prototype (`_Static_assert` + `__builtin_types_compatible_p`), then runs the `PTHREAD_ONCE_INIT`/`FD_SET`/`errno`/stdio/`environ` checks. It runs on whichever host executes the tests, so the Linux tables are checked by the Linux CI jobs.

TinyCC parsed 109,893 preprocessed lines before (64K of them from the headers); it now parses the 42,497 lines of the file itself.

## Reachability prune
The source-level reachability keeps every function that shares a used name, so the self-host C carried ~5K lines of functions nothing calls (`strconv` converters, `map_free`, `wtf8_*`, enum print helpers, ...). Now each generation worker records, per function, its C span, its prototype span and the indexed C names it mentions (an identifier scan for `__`/`v_fastc_` names looked up in a C-name table built from the signature index before generation); the stitch walks from `main`, the lifecycle hooks and every non-body piece (runtime, constants, globals, dispatch tables, spawn helpers, generic instances) and cuts the unreachable functions, their prototypes and the unused enum `str`/print helpers (now per-enum pieces) out of the emitted ranges, with no copy of the kept text. Bare main-module names are not indexed (always kept), and a program without `main` keeps everything. The prune costs about 0.6 ms in the stitch.

## Concurrent TinyCC units
- `fastc_write_c_units` cuts the pieces into up to 8 translation units (`VJOBS` caps it; `VJOBS=1` keeps the single-file build): the head (typedefs, prototypes, runtime) in every unit, the interface dispatch tables, lifecycle functions and `main` only in the first, the file bodies grouped by size; the constants and globals are definitions in the first unit and `extern` declarations in the others. The units are written by concurrent threads (one buffered write each).
- `fastc_compile_c_units` compiles them with concurrent `tcc -c` processes started through `posix_spawn` (`v_os_exec_capture_start`; forking the 220 MB driver with `os.Process` cost 4 ms more), drains their output and waits; one `tcc` call links the objects, also through `posix_spawn` instead of `cmdexec`'s fork-and-poll loop. `-keepc` still writes the whole program as one file, and that file is identical to the single-unit build's.
- `unit_compile_nix.c.v` / `unit_compile_windows.c.v` hold the platform variants (Windows keeps `os.Process`).

## No Apple `codesign`
TinyCC signs the executables it links on macOS by running `codesign -f -s - <file>`, which costs ~50 ms per build. The drivers put a no-op `codesign` (a link to `/usr/bin/true`) first on TinyCC's PATH and ad-hoc sign the Mach-O themselves: `gen/fastc/macho_sign.v` reuses or adds the `LC_CODE_SIGNATURE` load command, patches the `__LINKEDIT` sizes, and appends a SuperBlob with a v0x20400 CodeDirectory (SHA-256 4 KiB page hashes through CommonCrypto on macOS, in about a millisecond), empty requirements and empty CMS wrapper, patched into the file in place. `macho_sign_test.v` links a program with TinyCC without any `codesign`, signs it, runs it, verifies it with `codesign --verify --strict`, re-signs it, and replaces a larger signature.

## Also
- The macOS SDK path is taken from `SDKROOT` or the known SDK directories before `xcrun --show-sdk-path` is run; that call cost 7-9 ms inside every build (master pays it too).
- `fastc_platform_int_c_type` stays a global (master fixed the bootstrap build itself in #28307).
- Both drivers print `tcc.*` phase marks under `FASTC_BENCH_PHASES=1` (units written/started/compiled, linked, signed).

## Numbers (`vlib/v3/v3.v`, 177 files, 70.4K lines; clang `-O2 -gc none -prealloc` drivers, interleaved, quiet machine)
| | master | this PR |
|---|---|---|
| self-host C | 47,106 lines, 38 `#include` | 42.5K lines, 0 `#include`, 8 units |
| driver `fastc parse+gen` row | 14-15.5 ms | 15.5-17.5 ms |
| driver `tcc` row | 99-104 ms | 29-32 ms |
| driver `total` | 115-118 ms | 47-50 ms |
| TinyCC-built self-host, whole `v5 → v6` build | 0.16-0.18 s | 0.10 s |

The `tcc` row is now ~0.8 ms writing the units, ~13-14 ms of concurrent compiles (8 processes on 6 performance cores), ~14 ms of TinyCC linking (7 ms of it fixed even for a tiny object: the `sh -c codesign` call, the stdlib and the output), ~1 ms signing and cleanup. Those floors, plus the ~16 ms of generation, put the whole build at 1.4-1.5M lines/s; the 2.5M lines/s (28 ms) target is not reachable with TinyCC as an external compile-and-link step.

### DIFF
--- a/vlib/sync/thread_helper.h
+++ b/vlib/sync/thread_helper.h
@@ -18,8 +18,10 @@ static inline int v_sync_thread_create_detached(void *start, void *arg) {
 	return 0;
 }
 #else
+#ifndef V_FASTC_NO_HEADERS
 #include <pthread.h>
 #include <stdlib.h>
+#endif
 
 static inline int v_sync_thread_create_detached(void *start, void *arg) {
 	pthread_attr_t attr;
--- a/vlib/v3/README.md
+++ b/vlib/v3/README.md
@@ -202,6 +202,20 @@ constant and global declarations: a large file's constants are parsed as separat
 candidates (merged in source order), and the global phase parses only the recorded global
 declarations instead of whole files.
 
+A self-host build (`-selfhost`) for 64-bit macOS or glibc Linux emits C with no `#include`
+at all: `gen/fastc/c_abi.v` holds, per target, the C library types, struct layouts, macros,
+globals and function prototypes the emitted code uses, and that prelude replaces the header block
+of the preamble; V's own C helper headers (`vlib/os/execute_capture_nix.h`, ...) are inlined
+after it, and `#include` lines from V sources are left out. TinyCC then parses about 42K lines
+instead of the 110K that the system headers expanded to. The build passes
+`-Werror=implicit-function-declaration`, so a C function missing from the table fails the build
+instead of being called through an implicit `int` prototype; `c_abi_test.v` compiles the table
+against the host headers and checks every layout, size, value and prototype. The stitch also
+drops the indexed functions (and enum `str`/print helpers) that nothing reachable from `main`,
+the lifecycle hooks or the non-body pieces refers to, which the source-level name-grouped
+reachability keeps: each worker records its definitions and the mangled names they mention, and
+the assembly cuts the unreachable spans out of the pieces (about 9% of the self-host C).
+
 The standalone compiler supports `self` directly and defaults that command to FastC. For example,
 `./v self x5` replaces the compiler through five descendant FastC generations, with each installed
 generation compiling the next one. `-b fastc`, `-gc none`, `-cc tinyc|tcc`, `-keepc`, `-silent`,
--- a/vlib/v3/README.md
+++ b/vlib/v3/README.md
@@ -208,13 +208,26 @@ globals and function prototypes the emitted code uses, and that prelude replaces
 of the preamble; V's own C helper headers (`vlib/os/execute_capture_nix.h`, ...) are inlined
 after it, and `#include` lines from V sources are left out. TinyCC then parses about 42K lines
 instead of the 110K that the system headers expanded to. The build passes
-`-Werror=implicit-function-declaration`, so a C function missing from the table fails the build
-instead of being called through an implicit `int` prototype; `c_abi_test.v` compiles the table
+`-Werror=implicit-function-declaration` (and no `-w`, which would silence it), so a C function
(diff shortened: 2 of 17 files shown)


medvednikov:
@codex review

PR #28309, opened 2026-09-03
vfmt: stop rewriting a NUL escape into an octal escape

medvednikov:
`v fmt -w` silently changes string data.

The V3 formatter stores string literals decoded and re-encodes them, and `escape_string` wrote a NUL byte as `\0`. V's octal escape absorbs the digits that follow it, so whenever a NUL is followed by an octal digit the re-encoded literal parses as a *different* string:

```v
a := 'x\x0041y'   // x, NUL, `4`, `1`, y  ->  [120, 0, 52, 49, 121]
```

after `v fmt -w` becomes

```v
a := 'x\041y'     // x, `!`, y            ->  [120, 33, 121]
```

Five bytes become three, with no diagnostic. Reproduced end to end by running the program before and after formatting.

## Fix

Drop the special case so NUL falls through to the `\x` branch. `u8.hex()` always emits two digits, so `\x00` cannot absorb a following character and the literal round-trips. That branch already handles every other control byte, so this removes a special case rather than adding one. It also covers the same hazard for `` `\0` `` char literals.

## Testing

- New regression test in `vlib/v3/gen/v/gen_test.v`, covering the octal-absorbing case, a plain `\x00`, a char literal, and idempotency. Confirmed it fails without the fix (`assert out.contains("'x\x0041y'")`) and passes with it.
- `v fmt -verify` over the tree accepts exactly the same files as before: 6214 pass, and the same 49 fail — I diffed the failing sets and they are identical. Those 49 fail for unrelated pre-existing reasons (files that match neither the V3 nor the legacy formatter, e.g. `FnWorkItem` in `vlib/v3/transform/transform.v`, where fields were added without re-running vfmt).
- `cmd/tools/vfmt_test.v` fails identically on pristine master (`test_fmt_checks_accept_legacy_formatted_source`), so it is pre-existing and unrelated.

## Not in this PR

While tracking this down I confirmed two further V3-formatter deviations from the legacy formatter, both style rather than correctness, so I left them out to keep this focused:

- `pub type F = fn (a int)` is reformatted to `fn(a int)` — the space is dropped.
- `$if !flag ? {` is reformatted to `$if !flag? {`.

Happy to do either as a follow-up. Note also that `v fmt -w` rewriting legacy-formatted files wholesale is by design — `verify_file` deliberately accepts both formatters during the transition — so that churn is expected, not a bug.

### DIFF
--- a/vlib/v3/gen/v/gen.v
+++ b/vlib/v3/gen/v/gen.v
@@ -4185,10 +4185,11 @@ fn escape_string(s string, quote u8) string {
 			12 {
 				b.write_string('\\f')
 			}
-			0 {
-				b.write_string('\\0')
-			}
 			else {
+				// NUL is written as `\\x00`, never `\\0`: `u8.hex()` always emits two digits, so
+				// the escape cannot absorb a following character, whereas `\\0` before an octal
+				// digit re-parses as one octal escape and silently changes the string's bytes
+				// (`'x\\x0041y'` would come back as `'x\\041y'`, i.e. `x!y`).
 				if c < 32 || c == 127 {
 					b.write_string('\\x${c.hex()}')
 				} else if c == quote {
--- a/vlib/v3/gen/v/gen_test.v
+++ b/vlib/v3/gen/v/gen_test.v
@@ -1591,3 +1591,18 @@ fn test_formatter_keeps_rows_of_small_nested_arrays_on_one_line() {
 	out := vfmt('rows_of_small_nested_arrays', source)
 	assert out == source, out
 }
+
+// A NUL byte must be written as `\x00`, not `\0`. V's octal escape absorbs the digits that
+// follow, so `\0` before an octal digit re-parses as a single escape and the formatter would
+// silently rewrite the string's bytes: `'x\x0041y'` (x, NUL, `4`, `1`, y) came back as
+// `'x\041y'`, which is `x!y`.
+fn test_formatter_keeps_nul_escape_unambiguous() {
+	source := "fn main() {\n\ta := 'x\\x0041y'\n\tb := 'p\\x00q'\n\tc := `\\x00`\n\tprintln(a + b + c.str())\n}\n"
+	out := vfmt('nul_escape', source)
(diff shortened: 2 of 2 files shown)


medvednikov:
Note on the red CI — none of it comes from this PR. All three causes are pre-existing on master:

**1. `vlib/v/tests/builtin_arrays/sorting_compare_fn_with_mut_reference_test.v:74: unexpected token `@`, expecting name`** (`clang-linux`, `gcc-linux`, `tools-*`)

The V1 parser rejects `@[required]` on its own line before a struct field. The file is untouched by this PR, and master fails identically — reproduced locally by building both and running the same command:

```
$ ./v -old-compiler -o /tmp/x vlib/v/tests/builtin_arrays/sorting_compare_fn_with_mut_reference_test.v
<my branch> exit=1  ...:74:2: error: unexpected token `@`, expecting name
<master>    exit=1  ...:74:2: error: unexpected token `@`, expecting name
```

Formatting the file moves the attribute inline and clears that error, but V1 then rejects it for a second, unrelated reason (`sort_with_compare callback function parameter 'a' with type '&Thing' should be '&&Thing'`), so the test is V3-only either way. Not something formatting can fix.

**2. `cannot assign to field 'x11_keycodes': expected '[256]int', not '[256]i32'`** (`Linux (gcc)`)

In `vlib/x/multiwindow/native_operation_proof_test.v`. Fallout from the `int` → 64-bit change (#28293).

**3. `incompatible pointer types initializing 'v3__flat__FlatAst *' with 'v3__flat__FlatAst **'`** (`clang-macos` ×3)

A v3 C-generation issue. `clang-macos (macos-14/15/26)` are also failing on master's own commit.

Worth noting master's check list is not a usable baseline right now — most of its jobs are still `queued`, and several of these workflows have path filters that skip them for a vlib/v3-only commit, so they show no result there rather than a pass.

For what this PR does change, the check that matters is `v fmt -verify` over the tree: it accepts exactly the same files before and after — 6214 pass, the same 49 fail — across `vlib/`, `cmd/` and `examples/`, with the failing sets diffed rather than counted.

PR #28310, opened 2026-09-03
vfmt: keep the space in `fn (` types and before a `$if` flag marker

medvednikov:
Two places where the V3 formatter rewrites source into a spelling the tree does not use, so running `v fmt` on an untouched file produces a diff.

## `fn (int) int` → `fn(int) int`

The type system spells a function type without the space internally, and the formatter renders types straight from that name. Every function type in a file was rewritten — declarations, struct fields, parameters and return types alike:

```v
pub type Cb = fn (a int, b string) !   ->  fn(a int, b string) !
	cb  fn (int) int = unsafe { nil }  ->  fn(int) int
	m   map[string]fn (int) int        ->  map[string]fn(int) int
```

The tree writes `fn (` in 139 files and `fn(` in one; the legacy formatter emits `fn (`. Fixed by restoring the space when rendering a type, guarded so an ordinary name ending in `fn` (`myfn(`) is untouched.

## `$if flag ? {` → `$if flag? {`

The condition is stored without that space so flag lookups can match on it. `parse_attribute_comptime_cond` already makes an `is_fmt` exception for `@[if flag ?]`; `parse_comptime_cond` now makes the same one, so only formatting is affected and flag evaluation is unchanged. The tree writes `? {` in 9 files and `?{` in none.

## Files updated to match the tree

Three files carried the formatter's own spelling and would otherwise start failing `v fmt -verify`:

- `vlib/v/tests/anon_fn_map_assign_test.v` — one `type ... = fn(int) int`, whose own body already used `fn (x int) int`.
- `vlib/v3/fastcdriver/fastcdriver.v` and `vlib/v3/gen/fastc/arm64_test.v` — 34 `$if arm64? {`. I checked the diff is 34 bare `$if` statements and no string literal was touched (`arm64_test.v` embeds V source in a literal that already used the spaced form).

## Testing

- Two new regression tests in `vlib/v3/gen/v/gen_test.v`, both confirmed to fail without their fix, covering every function-type position and idempotency.
- One existing test encoded `fn(` in its expected output; it is about anon-fn body expansion and the spelling was incidental, so the expectation is updated.
- `v fmt -verify` over the tree accepts exactly the same files as before — 6214 pass, the same 49 fail. I diffed the failing sets: identical. (Those 49 fail for unrelated pre-existing reasons; they match neither formatter, mostly from edits that skipped vfmt.)
- `vlib/v3/gen/v/` and `vlib/v3/parser/` suites pass (6/6); `vlib/v3/gen/fastc/arm64_test.v` passes; the compiler still builds itself after the `$if` edits.

## Ordering note

These matter before any tree-wide formatting sweep: a sweep run today would bake `fn(` and `flag?` into every file it touched.

Independent of #28309 (the NUL-escape fix) — different code, no conflict.

### DIFF
--- a/vlib/v/tests/anon_fn_map_assign_test.v
+++ b/vlib/v/tests/anon_fn_map_assign_test.v
@@ -2,7 +2,7 @@
 // leak `cur_indexexpr`, so that any later non-scalar map/array assignment in the
 // same compilation unit was generated without the `= ` operator (broken C).
 // See https://github.com/vlang/v/issues (map assignment of fn literal leaks cgen state)
-type AnonFnMapAssignCb = fn(int) int
+type AnonFnMapAssignCb = fn (int) int
 
 struct AnonFnMapAssignPoint {
 	x int
--- a/vlib/v3/fastcdriver/fastcdriver.v
+++ b/vlib/v3/fastcdriver/fastcdriver.v
@@ -393,7 +393,7 @@ pub fn run(args []string) {
 	prefs.ccompiler = 'tinyc'
 	prefs.building_v = real_input.ends_with('/vlib/v3/v3.v')
 	prefs.selfhost = prefs.building_v
-	$if arm64? {
+	$if arm64 ? {
 		prefs.target = pref.Target{
 			os: 'macos'
 			arch: 'arm64'
--- a/vlib/v3/gen/fastc/arm64_test.v
+++ b/vlib/v3/gen/fastc/arm64_test.v
@@ -4,7 +4,7 @@ import os
 import v3.pref
 
 fn test_fastc_parser_emits_arm64_without_c() {
-	$if arm64? {
+	$if arm64 ? {
 		test_dir := os.join_path(os.temp_dir(), 'fastc_arm64_${os.getpid()}')
 		os.rmdir_all(test_dir) or {}
 		os.mkdir_all(test_dir) or { panic(err) }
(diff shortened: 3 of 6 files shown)


medvednikov:
Note on the red CI — none of it comes from this PR. All three causes are pre-existing on master:

**1. `vlib/v/tests/builtin_arrays/sorting_compare_fn_with_mut_reference_test.v:74: unexpected token `@`, expecting name`** (`clang-linux`, `gcc-linux`, `tools-*`)

The V1 parser rejects `@[required]` on its own line before a struct field. The file is untouched by this PR, and master fails identically — reproduced locally by building both and running the same command:

```
$ ./v -old-compiler -o /tmp/x vlib/v/tests/builtin_arrays/sorting_compare_fn_with_mut_reference_test.v
<my branch> exit=1  ...:74:2: error: unexpected token `@`, expecting name
<master>    exit=1  ...:74:2: error: unexpected token `@`, expecting name
```

Formatting the file moves the attribute inline and clears that error, but V1 then rejects it for a second, unrelated reason (`sort_with_compare callback function parameter 'a' with type '&Thing' should be '&&Thing'`), so the test is V3-only either way. Not something formatting can fix.

**2. `cannot assign to field 'x11_keycodes': expected '[256]int', not '[256]i32'`** (`Linux (gcc)`)

In `vlib/x/multiwindow/native_operation_proof_test.v`. Fallout from the `int` → 64-bit change (#28293).

**3. `incompatible pointer types initializing 'v3__flat__FlatAst *' with 'v3__flat__FlatAst **'`** (`clang-macos` ×3)

A v3 C-generation issue. `clang-macos (macos-14/15/26)` are also failing on master's own commit.

Worth noting master's check list is not a usable baseline right now — most of its jobs are still `queued`, and several of these workflows have path filters that skip them for a vlib/v3-only commit, so they show no result there rather than a pass.

For what this PR does change, the check that matters is `v fmt -verify` over the tree: it accepts exactly the same files before and after — 6214 pass, the same 49 fail — across `vlib/`, `cmd/` and `examples/`, with the failing sets diffed rather than counted.

PR #28314, opened 2026-09-03
vfmt: keep embed state out of field attributes

medvednikov:
Follow-up to #28311 after it was merged.

The embed marker introduced there shared the unrestricted user-attribute namespace. A field such as thread thread with the __v3_embedded_field attribute could therefore be misclassified as an embed and rewritten.

This passes embed state explicitly to apply_field_meta, preserves arbitrary user attributes, and extends the formatter regression to cover the colliding attribute name.

Tests:
- ./vnew -silent test vlib/v3/gen/v/gen_test.v
- ./vnew -silent test vlib/v3/parser/


medvednikov:
@codex review

PR #28316, opened 2026-09-03
vfmt: preserve gated slice markers

### LINKED ISSUE #28315: v fmt -w silently strips the # from a gated slice (body#[a..b]), changing clamped access into a panic
## Describe the bug

`v fmt -w` silently strips the `#` prefix from a gated/clamping slice
expression (`body#[a..b]`), rewriting it to a plain slice (`body[a..b]`).
This is not a cosmetic change: V's `#[...]` operator returns a
clamped/empty result for an out-of-range slice, while a plain `[...]`
slice panics. So `v fmt -w` can silently turn a defensive, panic-safe
expression into one that crashes on exactly the inputs the `#[` operator
existed to handle safely -- with no warning, on a file that already
compiled and ran correctly before formatting.

## Reproduction

```v
fn main() {
	body := 'hello world'
	start := 3
	end := 100
	x := body#[start..end]
	println(x)
}
```

```
v run repro.v      # prints "lo world" -- clamped, no panic
v fmt -w repro.v    # "Reformatted file"
v run repro.v       # V panic: substr(3, 100) out of bounds (len=11) s=hello world
```

The formatted file reads `x := body[start..end]` -- the `#` is gone.

## Expected behavior

`v fmt -w` should preserve the `#[...]` operator exactly as written; it
must never rewrite it to a plain `[...]` slice, since the two have
different runtime behavior on out-of-range indices.

## Real-world impact

Found via `vlib/net/http/request.v`'s `parse_multipart_form()`, which
uses `body#[line_segments[N].start..line_segments[N].end]` specifically
to avoid panicking on a malformed/adversarial multipart body. A `v fmt -w`
pass over that file silently converted both occurrences to plain slices in
[vlang/v#28303](https://github.com/vlang/v/pull/28303) (an unrelated PR),
discovered only because an adversarial-review pass flagged it -- the
commit's own author did not notice, since the diff still compiled and all
existing tests still passed (the surrounding loop's own invariants happen
to keep every segment in-bounds today, so the lost safety net was not yet
exercised).

## V version

```
$ v version
V 0.5.2
```
Reproduced on `vlang/v` master at commit `c6674198d5` (2026-09-03), i.e.
AFTER the recent #28309/#28310/#28311 vfmt fixes -- this is a separate,
still-open bug, not covered by any of those.

> [!NOTE]
>

medvednikov:
## Summary
- emit the gated `#` marker when V3 formats range slices
- cover bounded, open-ended, and fully open gated slices
- verify the formatter remains a fixed point

## Testing
- `./vnew -silent test vlib/v3/gen/v/gen_test.v`
- `./vnew fmt vlib/net/http/request.v` preserves both affected `body#[...]` expressions
- `./vnew -silent test cmd/tools/vfmt_test.v` reaches an unrelated existing failure in `test_fmt_checks_accept_legacy_formatted_source`: current V3 output preserves a blank line after the header comment

Closes #28315

### DIFF
--- a/vlib/v3/gen/v/gen.v
+++ b/vlib/v3/gen/v/gen.v
@@ -1549,8 +1549,9 @@ fn (mut g Gen) index_expr(id flat.NodeId) {
 		return
 	}
 	g.expr(children[0])
+	gate := if n.op == .gated_index { '#' } else { '' }
 	if n.value == 'range' {
-		g.write('[')
+		g.write('${gate}[')
 		if children.len > 1 && int(children[1]) >= 0 {
 			g.expr(children[1])
 		}
@@ -1561,7 +1562,6 @@ fn (mut g Gen) index_expr(id flat.NodeId) {
 		g.write(']')
 		return
 	}
-	gate := if n.op == .gated_index { '#' } else { '' }
 	g.write('${gate}[')
 	g.expr_list(children[1..], ', ')
 	g.write(']')
--- a/vlib/v3/gen/v/gen_test.v
+++ b/vlib/v3/gen/v/gen_test.v
@@ -87,6 +87,13 @@ fn test_formatter_preserves_blank_lines_between_statements() {
 	assert vfmt('statement_blank_lines_twice', out) == out
 }
 
+fn test_formatter_preserves_gated_slices() {
+	source := 'fn slices(body string, items []int, start int, end int) {\n\t_ := body#[start..end]\n\t_ := items#[..end]\n\t_ := items#[start..]\n\t_ := items#[..]\n}\n'
+	out := vfmt('gated_slices', source)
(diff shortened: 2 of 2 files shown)


medvednikov:
@codex review

medvednikov:
No blocking findings. I’d approve it.

The fix is narrowly scoped and correct: gate is now computed before the range branch, so .gated_index emits #[...] for range slices just as it already did for ordinary indexes. This matches the existing formatter semantics, which emits # based on gatedness before formatting the index contents.

The regression test covers bounded, missing-low, missing-high, and fully open gated slices, and verifies formatter idempotence. That directly addresses the reported semantic regression where body#[a..b] was being changed to body[a..b].

Verdict: LGTM / approve.

PR #28318, opened 2026-09-03
v3: scope C function call metadata by module

### LINKED ISSUE #28251: checker: variadic vs fixed-arity `fn C.fcntl` across modules is a fatal error (breaks third-party code that imports `net`)
## Describe the bug

Since the cross-module C-function redeclaration check was added in #27964 / #27791 (parser/fn.v `c_fn_declarations_are_compatible`), a module can no longer declare `C.fcntl` with the classic fixed-argument prototype if `vlib/net` (or `vlib/fasthttp`) is anywhere in the build.

`vlib/net/aasocket.c.v:110`, `vlib/fasthttp/*` and `vlib/builtin/cfns.c.v:574` all declare:

```v
fn C.fcntl(fd i32, cmd i32, arg ...voidptr) i32
```

Third-party code (and lots of existing code in the wild) commonly declares the same libc function non-variadically, e.g.:

```v
fn C.fcntl(fd int, cmd int, arg int) int
```

Both are legitimate bindings of `int fcntl(int, int, ...)`. `c_fn_declarations_are_compatible_at_depth` rejects the pair outright on `left.is_variadic != right.is_variadic`, so the build fails with a hard error instead of picking a winner or warning.

`builtin` is exempted from the check (`existing.mod != 'builtin' && new_fn.mod != 'builtin'`), so the clash only surfaces against `net`/`fasthttp`.

This broke a real project: `main.v` imports the third-party `vxui` module, which imports `net.websocket`. `vxui/display_macos.v` declares `fn C.fcntl(fd int, cmd int, arg int) int`, and the build now dies:

```
vlib/net/aasocket.c.v:110:6: error: C function `C.fcntl` was already declared with a different signature in module `vxui` at .vmodules/vxui/display_macos.v:18:7
```

## Reproduction

`repro.v`:

```v
import net

fn C.fcntl(fd int, cmd int, arg int) int

fn main() {
	println(net.default_tcp_socket)
}
```

```
$ v run repro.v
vlib/net/aasocket.c.v:110:6: error: C function `C.fcntl` was already declared with a different signature in module `main` at repro.v:3:6
  108 | fn C.ioctlsocket(s i32, cmd i32, argp &u32) i32
  109 |
  110 | fn C.fcntl(fd i32, cmd i32, arg ...voidptr) i32
      |      ~~~~~
  111 |
  112 | fn C.select(ndfs i32, readfds &C.fd_set, writefds &C.fd_set, exceptfds &C.fd_set, timeout &C.timeval) i32
```

## Expected Behavior

A non-variadic binding of a C variadic function should not hard-fail the build against vlib's variadic binding. Options:

- Treat a variadic C declaration as compatible with a fixed-arity one whose fixed params match the variadic's fixed params (C calling convention makes them ABI-compatible for the arg counts actually used).
- Or downgrade the cross-module mismatch to a warning and keep the first-registered declaration, which is what #27791 originally asked for ("the winner poisons the other module's call sites" — a warning + deterministic winner, not a fatal error).

## Current Behavior

Fatal `error:` from `p.error_with_pos_no_advance` in `vlib/v/parser/fn.v`, with no way for downstream code to build unless it patches the third-party module (or vlib) by hand.

## Possible Solution

In `Table.c_fn_declarations_are_compatible_at_depth` (`vlib/v/ast/table.v`), when exactly one side `is_variadic` and its `left_fixed_params == right fixed params` and those params are component-compatible, return `true` instead of failing on the `is_variadic` flag difference.

## Environment details

- V full version: `V 0.5.2 9fd548810f09e5b49f0b06e500ef5d6845cc8142.19a2674`
- OS: macOS 26.6.2 (Darwin 25.6.0), arm64
- C compiler: Apple clang 21.0.0

### Regression

- `vlib/net` `C.fcntl` became variadic in `a87a4d73b9a` (2026-02-07).
- The cross-module C redeclaration check that turns the mismatch fatal landed in `dff6356bde` / #27964 (2026-07-28), following #27791.

Before #27964 the same code compiled.

> [!NOTE]
>

medvednikov:
## Summary

- resolve V3 C-call parameter and variadic metadata from the declaration in the current module
- allow ABI-safe fixed declarations to coexist with variadic declarations across modules
- keep fixed/variadic declarations incompatible when C default argument promotions change the ABI
- cover the `fcntl` conflict from #28251 and both directions of module-local CGen metadata lookup

Closes #28251

## Testing

- `VFLAGS="-gc none" ./vnew -old-compiler -gc none -silent test vlib/v3/gen/c/` (18 passed)
- `VFLAGS="-gc none" ./vnew -old-compiler -gc none vlib/v3/tests/c_fn_redeclaration_and_empty_struct_test.v`
- `VFLAGS="-gc none" ./vnew -old-compiler -gc none vlib/v3/tests/c_variadic_call_codegen_test.v`
- `VFLAGS="-gc none" ./vnew -silent vlib/v/compiler_errors_test.v` (1696 passed, 1 skipped)

`vlib/v/gen/c/coutput_test.v` was also attempted, but this checkout does not have the Boehm `gc` library required by its explicit `-gc boehm` fixture.

### DIFF
--- a/vlib/v3/types/checker.v
+++ b/vlib/v3/types/checker.v
@@ -690,6 +690,7 @@ pub mut:
 	c_fn_module_ret_types            map[string]Type
 	c_fn_module_param_types          map[string][]Type
 	c_fn_module_variadic             map[string]bool
+	c_fn_abi_variadic_prefixes       map[string]int
 	fn_shared_params                 map[string][]bool
 	mut_receiver_methods             map[string]bool
 	source_no_body_fns               map[string]bool
@@ -1008,6 +1009,7 @@ pub fn TypeChecker.new(a &flat.FlatAst) TypeChecker {
 		c_fn_module_ret_types:                 map[string]Type{}
 		c_fn_module_param_types:               map[string][]Type{}
 		c_fn_module_variadic:                  map[string]bool{}
+		c_fn_abi_variadic_prefixes:            map[string]int{}
 		fn_shared_params:                      map[string][]bool{}
 		mut_receiver_methods:                  map[string]bool{}
 		source_no_body_fns:                    map[string]bool{}
@@ -1169,6 +1171,7 @@ fn (tc &TypeChecker) fork_program_view(ast &flat.FlatAst, direct_dependencies_by
 		c_fn_module_ret_types:              tc.c_fn_module_ret_types
 		c_fn_module_param_types:            tc.c_fn_module_param_types
 		c_fn_module_variadic:               tc.c_fn_module_variadic
+		c_fn_abi_variadic_prefixes:         tc.c_fn_abi_variadic_prefixes
 		fn_shared_params:                   tc.fn_shared_params
 		mut_receiver_methods:               tc.mut_receiver_methods
 		source_no_body_fns:                 tc.source_no_body_fns
@@ -3581,6 +3584,11 @@ fn (mut tc TypeChecker) collect_after_index(a &flat.FlatAst) {
 						false)
 				}
 				if is_variadic {
+					tc.c_fn_abi_variadic_prefixes[c_name] = if ptypes.len > 0 {
+						ptypes.len - 1
+					} else {
+						0
+					}
 					tc.register_c_variadic_fn(node.value)
 				}
 				if !node.value.starts_with('C.') {
--- a/vlib/v3/types/checker.v
+++ b/vlib/v3/types/checker.v
@@ -4048,10 +4048,30 @@ fn c_fn_decl_signatures_compatible(a CFnDeclSignature, b CFnDeclSignature) bool 
 	if !c_fn_decl_abi_types_compatible(a.return_type, b.return_type) {
 		return false
 	}
-	if a.is_variadic != b.is_variadic || a.params.len != b.params.len {
+	if a.is_variadic != b.is_variadic {
+		variadic := if a.is_variadic { a } else { b }
+		fixed := if a.is_variadic { b } else { a }
+		if fixed.params.len < variadic.params.len
+			|| !c_fn_decl_param_prefix_compatible(fixed.params, variadic.params, variadic.params.len) {
+			return false
+		}
+		for i in variadic.params.len .. fixed.params.len {
+			if !c_fn_fixed_param_is_compatible_with_variadic(fixed.params[i]) {
+				return false
+			}
+		}
+		return true
+	}
+	if a.params.len != b.params.len {
 		return false
 	}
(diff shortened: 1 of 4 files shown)


medvednikov:
@codex review

vlib/v3/types/checker.v
```diff
@@ -4048,12 +4048,32 @@ fn c_fn_decl_signatures_compatible(a CFnDeclSignature, b CFnDeclSignature) bool
...
+	if a.params.len != b.params.len {
 		return false
 	}
 	return c_fn_decl_param_prefix_compatible(a.params, b.params, a.params.len)
 }
 
+fn c_fn_fixed_param_is_compatible_with_variadic(typ string) bool {
+	// These types are changed by C's default argument promotions. A fixed
+	// declaration would therefore not describe the same argument ABI.
+	return typ !in ['i8', 'i16', 'u8', 'u16', 'float', 'char', 'bool']
```
medvednikov, line 4082: This accepts a fixed trailing V `int` as ABI-compatible with a C variadic tail, but the generator does not lower that call as variadic in the fixed module anymore. On 64-bit targets a V `int` value is emitted as i64; `c_call_arg_cabi_cast` deliberately relies on a *fixed* C prototype to convert that to `int`, and only performs the explicit `(int)` cast for variadic-tail values. With this PR, `c_fn_is_variadic()` returns the module-local `false`, so `arg := some_runtime_int; C.fcntl(fd, cmd, arg)` is emitted as an i64 vararg even though `<fcntl.h>` declares `fcntl(int, int, ...)`. The new test passes literal `0`, which is already a C `int` and therefore hides this.

Please keep the module-local parameter metadata, but preserve a separate ABI-variadic fact/prefix for calls when another compatible declaration is variadic (or otherwise force vararg lowering for the fixed declaration's extra params), and add a non-literal `int` regression.

medvednikov (commented):
The module-scoped metadata direction looks right, but there is still a C-varargs ABI/codegen hole in the fixed-vs-variadic compatibility case. The inline comment has a concrete `int` case that is masked by the current literal-only `fcntl` regression test.

vlib/v3/types/checker.v
medvednikov: Fixed in 7ec0d2f3b4. V3 now records a symbol-level C ABI variadic prefix separately from each module’s semantic signature. CGen keeps the fixed module parameter metadata, but arguments after that ABI prefix use variadic lowering. The regression now passes a runtime V int and asserts that the generated call contains `(int)(arg)`. The focused regression, all 18 V3 CGen tests, the existing C variadic integration test, and the 1,697-case compiler error suite pass.

PR #28325, opened 2026-09-03
v3: apply FastC link options before output setup

medvednikov:
## Summary

Follow-up to #28321.

- pass final linker arguments into libtcc preparation so all state-affecting options are parsed before `tcc_set_output_type`
- add explicit library and file operands only after output state initialization, matching TinyCC's required API ordering
- add a regression test proving `-Wl,-nostdlib` cannot resolve `libtcc1` through TinyCC's default library path
- preserve the original optimization by keeping preparation concurrent with translation-unit compilation

## Testing

- `./vnew fmt -verify vlib/v3/gen/fastc/link.v vlib/v3/gen/fastc/link_darwin.c.v vlib/v3/driver/driver.v vlib/v3/fastcdriver/fastcdriver.v vlib/v3/gen/fastc/macho_sign_test.v`
- `./vnew vlib/v3/gen/fastc/macho_sign_test.v`
- `./vnew -no-memory-limit vlib/v3/fastcdriver/fastcdriver_test.v`
- optimized-host end-to-end V3 FastC build, strict signature verification, and Hello World compile/run on the original PR branch before transplanting the identical patch

Broad test suites were skipped as requested.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -7454,6 +7454,12 @@ fn compile_v3_fastc_source(pieces []string, units fastc.FastcUnitLayout, bin_fil
 	if is_debug {
 		cc_args << '-g'
 	}
+	mut final_args := user_c_flags.clone()
+	if uses_threads {
+		final_args << '-lpthread'
+	}
+	final_args << '-lm'
+	final_args << environment_ld_flags
 	mut shim_dir := fastc.FastcCodesignShim{}
 	defer {
 		fastc.fastc_remove_codesign_shim_dir(shim_dir)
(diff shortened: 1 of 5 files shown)


medvednikov:
Code review finding (CI ignored): I think the new option pre-pass breaks operand-scoped linker state.

`fastc_libtcc_option_args(final_args)` removes literal link inputs, then applies *all* remaining options before `tcc_set_output_type()`. `fastc_finish_libtcc_link()` later adds the archives/files without replaying the options around them.

For example:

```text
-Wl,--whole-archive /tmp/libfoo.a -Wl,--no-whole-archive
```

TinyCC implements `--whole-archive` by toggling `s->filetype`, and `tcc_add_file()` reads that state when the archive is actually added. With this PR, the pre-pass applies both toggles first (ending with whole-archive off), filters `/tmp/libfoo.a` out of that pass, and then `fastc_finish_libtcc_link()` adds the archive after the state has already been cleared. The previous finish loop preserved this ordering by flushing pending options immediately before each recognized link input.

This is a real linker-ordering pattern (including pkg-config/static archive flows). Please preserve option/input ordering while still applying only the options that must precede `tcc_set_output_type()` (such as `-nostdlib`) early, and add a regression test with a literal `.a` wrapped in balanced `--whole-archive` / `--no-whole-archive`.

medvednikov:
Re-reviewed current head `3a5551866cf72680db12b682da41eb36725ca377` (CI ignored). The ordering issue I raised is fixed.

The preparation pass now extracts only the options that need to affect TinyCC before `tcc_set_output_type()` (notably `-nostdlib`/`-Wl,-nostdlib` and library/toolchain search paths), while `fastc_finish_libtcc_link()` again flushes the original pending options immediately before each `-l...` or literal link input. That preserves operand-scoped state such as balanced `--whole-archive` / `--no-whole-archive` around an archive.

The new regression test also exercises exactly that case and verifies the forced archive member is present in the output. I don't see another blocker in this PR.

PR #28329, opened 2026-09-03
net.http: accept HTTP status line without a reason phrase

### LINKED ISSUE #28182: net.http: a response with no reason phrase (`HTTP/1.1 200`) is rejected, and the request is retried 5x
### Describe the bug

`parse_status_line` in `vlib/net/http/response.v` requires exactly three space-separated tokens, so a status line that omits the reason phrase — `HTTP/1.1 200` — fails with `expected at least 3 tokens, but found: 2`.

Every mainstream HTTP client accepts that line. V is the only one I found that does not, which makes `net.http` unable to talk to servers that browsers, curl and Python reach without comment.

There is a second half. `parse_status_line` returns a plain `error()` with code 0, and `is_no_need_retry_error` only exempts a short list of socket error codes — so a *parse* failure is classified as retryable and the whole request is re-sent up to `max_retries` (5) times. The retry loop in `method_and_url_to_response` wraps `http_do`/`ssl_do` for every method, so a `POST` whose response cannot be parsed is delivered to the server five times.

### Reproduction Steps

Self-contained, no external server needed — `v run repro.v`:

```v
import net
import net.http
import time

struct Server {
mut:
	l    net.TcpListener
	line string
	hits int
}

// Answers every request with `line` as the status line, counting how many arrive.
fn (mut s Server) run() {
	for {
		mut c := s.l.accept() or { return }
		s.hits++
		mut buf := []u8{len: 4096}
		c.read(mut buf) or {}
		c.write_string('${s.line}\r\nContent-Length: 2\r\nConnection: close\r\n\r\nok') or {}
		c.close() or {}
	}
}

fn try(line string) ! {
	mut s := &Server{
		l:    net.listen_tcp(.ip, ':0')!
		line: line
	}
	port := s.l.addr()!.port()!
	spawn s.run()
	time.sleep(200 * time.millisecond)
	r := http.get('http://127.0.0.1:${port}/') or {
		println('"${line}" -> ERROR after ${s.hits} request(s): ${err}')
		return
	}
	println('"${line}" -> ${r.status_code} after ${s.hits} request(s)')
}

fn main() {
	try('HTTP/1.1 200 OK')! // reason phrase present
	try('HTTP/1.1 200 ')! // SP present, phrase empty
	try('HTTP/1.1 200')! // no SP, no phrase -- what some servers send
}
```

### Expected Behavior

```
"HTTP/1.1 200 OK" -> 200 after 1 request(s)
"HTTP/1.1 200 " -> 200 after 1 request(s)
"HTTP/1.1 200" -> 200 after 1 request(s)
```

### Current Behavior

```
"HTTP/1.1 200 OK" -> 200 after 1 request(s)
"HTTP/1.1 200 " -> 200 after 1 request(s)
"HTTP/1.1 200" -> ERROR after 5 request(s): expected at least 3 tokens, but found: 2
```

The third line is wrong twice over: the request failed, and it was sent five times.

There is also a case where the real error never reaches the caller at all. Because the parse error is treated as retryable, what surfaces is whatever the *last* attempt produced. If the server does not answer the retries (a one-shot server, or one that closes after an error), the caller gets

```
net: op timed out; code: 9
```

after a 30-second stall — naming neither HTTP nor the status line. The response had in fact arrived, complete, on the first read. I checked this with a traced build of the module: the read loop frames the response and breaks correctly (`has_cl=true exp=2 body=2` → `break`), so the stall is entirely the retry of a request whose answer was already in hand.

### Possible Solution

In `vlib/net/http/response.v`:

```diff
 	data := line.split_nth(' ', 3)
-	if data.len != 3 {
-		return error('expected at least 3 tokens, but found: ${data.len}')
+	if data.len < 2 {
+		return error('expected at least 2 tokens, but found: ${data.len}')
 	}
```

and at the return:

```diff
-	return version, strconv.atoi(data[1])!, data[2]
+	// RFC 9112 §4: the reason-phrase is optional, and servers that omit it
+	// sometimes omit the SP before it too.
+	reason := if data.len == 3 { data[2] } else { '' }
+	return version, strconv.atoi(data[1])!, reason
```

The `line.len < 5 || line[..5].to_lower() != 'http/'` guard just above already rejects anything that is not an HTTP status line, so this does not loosen the parser against garbage — a 0- or 1-token line is still an error.

I applied exactly this to a local copy of `vlib/net/http` and the reproduction above prints `200 after 1 request(s)` for all three status lines. I have not run the full `vlib/net/http` test suite against the patch, so that is worth checking before merging.

Separately, and I am happy to split it into its own issue: `is_no_need_retry_error` is an *exclusion* list of socket error codes, so every error that is not a socket error — including every response-parsing failure — is retried. Retrying a request because its response was malformed is not useful for any method, and for non-idempotent methods it is harmful. An inclusion list of genuinely retryable socket conditions would be the safer shape.

RFC 9112 §4 gives `status-line = HTTP-version SP status-code SP [ reason-phrase ]`, so a server that omits the trailing SP is a byte short of the grammar. But the same section says the reason-phrase carries no semantics and that clients SHOULD ignore it, and every implementation I checked treats the space as optional too:

| client | `HTTP/1.1 200`, no trailing space |
|---|---|
| Go `net/http` | accepted — `strings.Cut(line, " ")`, everything after the first space is the status |
| Python `http.client` | accepted — `except ValueError:` falls back to a 2-way split with `reason = ""` |
| curl, browsers | accepted |
| V `net.http` | error, then 5 retries |

V is also inconsistent with itself here: `net.websocket` validates its handshake with `starts_with('HTTP/1.1 101')` and reads such a server happily, so the same response is acceptable on one path and fatal on the other.

This is not hypothetical. I hit it against the REST API of a commercial USB CAN interface, which answers `HTTP/1.1 200\r\nTransfer-Encoding: chunked\r\n\r\n` (captured off the wire, not deduced). The device is reachable from curl, PowerShell and the vendor's own Python tooling, and unreachable from V — so my program has to ship a hand-written HTTP client whose only reason to exist is this one missing byte.

`parse_status_line` on `master` today is unchanged from the code quoted above.

### V version

V 0.5.1 7a7cf783a49229063175710d18385cb357b19b92.7134f48

```
OS: linux, Ubuntu 24.04.3 LTS (WSL 2)
Processor: 28 cpus, 64bit, little endian, 13th Gen Intel(R) Core(TM) i7-13850HX
cc version: gcc (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0
```

> [!NOTE]
>

HuntedByTheIRS:
Fixes #28182

## Problem

`parse_status_line` in `vlib/net/http/response.v` required exactly three
space-separated tokens, so a status line that omits the reason phrase —
`HTTP/1.1 200` — failed with `expected at least 3 tokens, but found: 2`.

Per RFC 9112 §4, the grammar is:

```
status-line = HTTP-version SP status-code SP [ reason-phrase ]
```

The reason-phrase is optional, and carries no semantics. Every mainstream
HTTP client accepts a bare `HTTP/1.1 200`; V was the only one that rejected it,
which made `net.http` unable to talk to servers that browsers, curl, Go and
Python all reach without comment.

## Fix

Relax `parse_status_line` to accept a two-token status line, defaulting the
reason phrase to `''`. The existing `http/` prefix guard still rejects
garbage, so a 0- or 1-token line remains an error.

## Verification

Added regression tests for both `HTTP/1.1 200` (no trailing SP) and
`HTTP/1.1 200 ` (empty reason phrase). The issue's reproduction now prints
`200 after 1 request(s)` for all three status-line forms:

```
$ ./vnew run repro.v
"HTTP/1.1 200 OK" -> 200 after 1 request(s)
"HTTP/1.1 200 "   -> 200 after 1 request(s)
"HTTP/1.1 200"    -> 200 after 1 request(s)
```

Tests run: `./vnew -silent test vlib/net/http/` — 30 passed, 4 skipped (Windows-only).

## Not addressed

The issue also notes that a *parse* failure is classified as retryable, so a
malformed response is re-sent up to `max_retries` (5) times. That is a
separate, broader concern about `is_no_need_retry_error`'s exclusion-list
shape and is out of scope for this PR (the author of the issue offered to
split it out too).

### DIFF
--- a/vlib/net/http/response.v
+++ b/vlib/net/http/response.v
@@ -153,8 +153,8 @@ fn parse_status_line(line string) !(string, int, string) {
 		return error('response does not start with HTTP/, line: `${line}`')
 	}
 	data := line.split_nth(' ', 3)
-	if data.len != 3 {
-		return error('expected at least 3 tokens, but found: ${data.len}')
+	if data.len < 2 {
+		return error('expected at least 2 tokens, but found: ${data.len}')
 	}
 	version := data[0].substr(5, data[0].len)
 	// validate version is 1*DIGIT "." 1*DIGIT
@@ -167,7 +167,10 @@ fn parse_status_line(line string) !(string, int, string) {
 			return error('HTTP version must contain only integers, found: `${digit}`')
 		}
 	}
-	return version, strconv.atoi(data[1])!, data[2]
+	// RFC 9112 §4: the reason-phrase is optional, and servers that omit it
+	// sometimes omit the SP before it too.
+	reason := if data.len == 3 { data[2] } else { '' }
+	return version, strconv.atoi(data[1])!, reason
 }
 
 // cookies parses the Set-Cookie headers into Cookie objects
--- a/vlib/net/http/response_test.v
+++ b/vlib/net/http/response_test.v
@@ -49,6 +49,31 @@ fn test_parse_response() {
 	assert x.body == 'Foo'
 }
 
+fn test_parse_response_without_reason_phrase() {
+	// RFC 9112 §4: the reason-phrase is optional, and some servers omit the
+	// trailing SP before it as well (`HTTP/1.1 200`).
+	content := 'HTTP/1.1 200\r\nContent-Length: 3\r\n\r\nFoo'
+	x := parse_response(content)!
+	assert x.http_version == '1.1'
+	assert x.status_code == 200
+	assert x.status_msg == ''
+	assert x.header.contains(.content_length)
+	assert x.header.get(.content_length)? == '3'
+	assert x.body == 'Foo'
+}
+
+fn test_parse_response_with_empty_reason_phrase() {
+	// trailing SP present, but the reason phrase itself is empty
+	content := 'HTTP/1.1 200 \r\nContent-Length: 3\r\n\r\nFoo'
+	x := parse_response(content)!
+	assert x.http_version == '1.1'
+	assert x.status_code == 200
+	assert x.status_msg == ''
+	assert x.header.contains(.content_length)
+	assert x.header.get(.content_length)? == '3'
+	assert x.body == 'Foo'
+}
+
 fn test_parse_response_with_cookies() {
 	cookie_id := 'v_is_best'
 	content := 'HTTP/1.1 200 OK\r\nSet-Cookie: id=${cookie_id}\r\nContent-Length: 3\r\n\r\nFoo'


medvednikov:
LGTM. The parser change is narrowly scoped: it keeps the existing HTTP/version validation, accepts the two-token interoperability case by defaulting the reason phrase to `''`, and still relies on `strconv.atoi` to reject a malformed status-code token. The regression tests cover both `HTTP/1.1 200 ` (empty reason phrase) and the intentionally permissive `HTTP/1.1 200` form.

Non-blocking standards note: RFC 9112's grammar still includes the SP after the status code even when the optional reason phrase is empty, so the no-trailing-SP form is technically an interoperability extension rather than strictly RFC-conforming. That matches the real-world behavior described in #28182 and seems appropriate here.

medvednikov:
thanks!

PR #28330, opened 2026-09-03
builder: rewrite gcc windows paths to short 8.3 names, like tcc (fixes #28126)

### LINKED ISSUE #28126: build： `v -cc gcc` fails in non-ASCII directories, but `v -cc tcc` and direct GCC work
### Describe the bug

 `v -cc gcc run main.v` fails when the source file/current directory path contains non-ASCII characters, for example Chinese characters. The same program works fine with: `v -cc tcc run main.v`

I also tested compiling and running a plain C file directly with GCC in the same non-ASCII directory, and it works correctly. This suggests that the issue is likely related to how V invokes GCC and passes paths on Windows, rather than GCC itself being unable to handle non-ASCII paths.

### Reproduction Steps

main.v
``` v
fn main() {
	println('hello')
}
```
```cmd
D:\工作目录>v -cc tcc run main.v
hello

D:\工作目录>v -cc gcc run main.v
cc: D:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot open output file D:\???????\main.exe: Invalid argument
cc: collect2.exe: error: ld returned 1 exit status
```
test.c
```c
#include <stdio.h>

int main(void) {
    printf("hello from gcc\n");
    return 0;
}
```
```cmd
D:\工作目录>gcc test.c -o test.exe

D:\工作目录>test.exe
hello from gcc

D:\工作目录>
```

### Expected Behavior

When the project directory contains non-ASCII characters, it should still be able to be compiled and run properly.or at least, V should provide clear diagnostic messages.

### Current Behavior

```cmd
cc: D:/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/14.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot open output file D:\???????\main.exe: Invalid argument
cc: collect2.exe: error: ld returned 1 exit status
```

### Possible Solution

_No response_

_No response_

### V version

V 0.5.2 1d0d86d

|V full version      |V 0.5.2 1d0d86dd855d639539602f604e3d8f534a057707
|OS                  |windows, Microsoft Windows 10 专业版 19045 64 位
|Processor           |4 cpus, 64bit, little endian, AMD Athlon(tm) X4 760K Quad Core Processor
|Memory              |6.99GB/11.94GB
|                    |
|V executable        |D:\v\v.exe
|V last modified time|2026-08-15 02:19:10
|                    |
|V home dir          |OK, value: D:\v
|VMODULES            |OK, value: C:\Users\TZ\.vmodules
|VTMP                |OK, value: C:\Users\TZ\AppData\Local\Temp\v_0
|Current working dir |contains these non ASCII characters: [`工`, `作`, `目`, `录`], value: D:\工作目录
|                    |
|Git version         |git version 2.45.1.windows.1
|V git status        |weekly.2025.33-3716-g1d0d86dd
|.git/config present |true
|                    |
|cc version          |N/A
|gcc version         |gcc (MinGW-W64 x86_64-ucrt-posix-seh, built by Brecht Sanders, r3) 14.2.0
|clang version       |(built by Brecht Sanders, r3) clang version 19.1.7
|msvc version        |N/A
|tcc version         |tcc version 0.9.28rc 2026-07-14_HEAD@d9d02c56* (x86_64 Windows)
|tcc git status      |thirdparty-windows-amd64 f7c7199b-dirty
|emcc version        |N/A
|glibc version       |N/A

> [!NOTE]
>

HuntedByTheIRS:
Fixes https://github.com/vlang/v/issues/28126

## Problem

On Windows, `v -cc gcc run main.v` fails when the project is in a directory whose path contains non-ASCII characters (e.g. `D:\工作目录`), while `v -cc tcc` and a direct `gcc test.c -o test.exe` both work:

```
ld.exe: cannot open output file D:\???????\main.exe: Invalid argument
```

## Root cause

V passes the C compiler arguments through a `.rsp` response file, whose contents are written using the ANSI code page encoding (`write_response_file`). The MinGW gcc toolchain reads response-file contents with the ANSI C runtime. When the code page cannot represent the non-ASCII characters (or the compiler build expects UTF-8), the paths are mangled into `?` before they reach `ld`.

tcc never hits this, because V already rewrites **all** Windows path arguments to pure ASCII 8.3 short names (`os.short_path`) when `cc == .tcc`. ASCII short paths are encoding-agnostic, so they survive the response file intact for both ANSI-CRT and UTF-8-native toolchains.

## Fix

Extend that existing mechanism to gcc: new `cc_uses_short_windows_paths()` helper gates the rewrite for `.gcc` and `.tcc`. Since `tcc_windows_path`/`tcc_windows_path_arg`/`tcc_quoted_path` already feed the `-o` output path, source paths, thirdparty args and the parallel-compile rsp path, no other call site needs changes.

clang, msvc, icc, emcc are intentionally excluded: msvc already keeps Unicode on the direct wide `CreateProcessW` command line (`msvc_should_use_rsp`), and the LLVM toolchain handles Unicode paths itself.

## Tests

- Added `test_cc_uses_short_windows_paths()` (Linux-runnable) in `vlib/v/builder/cc_tcc_retry_test.v`.
- Existing `rewrite_windows_path_arg` unit tests already cover the rewriting mechanics.
- `./vnew -silent test vlib/v/builder/` → 15 passed, 1 skipped (Windows-only).
- `./vnew -silent test vlib/v/builder/cc_tcc_retry_test.v` → 1 passed.
- `compiler_errors_test.v`: 52/53 pass; the 1 failure (`skip_unused/auto_str.vv`) also fails on clean master in this environment (V1/V3 default drift), unrelated to this change.

## Caveat

The fix relies on 8.3 short-name support on the target volume (Windows default). On volumes with 8.3 name creation disabled, gcc on non-ASCII paths will still fail — the same limitation tcc already has.

### DIFF
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -1551,6 +1551,16 @@ fn rewrite_windows_path_arg(arg string, resolver WindowsPathResolver) string {
 	return arg
 }
 
+// cc_uses_short_windows_paths reports whether the given C compiler needs the
+// Windows paths passed to it to be rewritten to their ASCII 8.3 short forms
+// first. Both tcc and the MinGW gcc toolchain read response file contents
+// with the ANSI C runtime: non-ASCII characters in paths get mangled on
+// Windows setups, whose active code page can not represent them (or whose
+// compiler expects UTF-8), so V passes them only as pure ASCII short paths.
+fn cc_uses_short_windows_paths(cc CC) bool {
+	return cc in [.gcc, .tcc]
+}
+
 fn short_windows_path(path string) string {
 	$if windows {
 		return os.short_path(path)
@@ -1560,7 +1570,7 @@ fn short_windows_path(path string) string {
 
 fn (v &Builder) tcc_windows_path(p string) string {
 	$if windows {
-		if v.ccoptions.cc == .tcc {
+		if cc_uses_short_windows_paths(v.ccoptions.cc) {
 			return short_windows_path(p)
 		}
 	}
@@ -1569,7 +1579,7 @@ fn (v &Builder) tcc_windows_path(p string) string {
 
 fn (v &Builder) tcc_windows_path_arg(arg string) string {
 	$if windows {
-		if v.ccoptions.cc == .tcc {
+		if cc_uses_short_windows_paths(v.ccoptions.cc) {
 			return rewrite_windows_path_arg(arg, short_windows_path)
 		}
 	}
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -1523,18 +1523,15 @@ fn looks_like_windows_path(value string) bool {
 	return value.contains('\\') || value.contains('/') || (value.len > 1 && value[1] == `:`)
 }
 
-fn rewrite_windows_path_arg(arg string, resolver WindowsPathResolver) string {
+// rewrite_windows_path_operand_arg rewrites only the path value of compiler
+// arguments that take a filesystem operand: the path bearing options `-I`,
+// `-L`, `-B`, `-o ` and `-c `, and bare source/object/archive file paths.
+// Option values that merely contain a path looking substring are left
+// untouched, so that e.g. `-DROOT="C:\Program Files\SDK"` keeps its exact
+// macro value after the rewrite.
+fn rewrite_windows_path_operand_arg(arg string, resolver WindowsPathResolver) string {
 	if arg == '' {
 		return ''
-	}
-	if start := arg.index('"') {
-		end := arg.last_index('"') or { -1 }
-		if end > start {
-			path := arg[start + 1..end]
-			if looks_like_windows_path(path) {
-				return arg[..start] + '"${resolver(path)}"' + arg[end + 1..]
-			}
-		}
 	}
 	for prefix in ['-I', '-L', '-B', '-o ', '-c '] {
 		if arg.starts_with(prefix) {
@@ -1549,6 +1546,27 @@ fn rewrite_windows_path_arg(arg string, resolver WindowsPathResolver) string {
 		return '"${resolver(trimmed)}"'
 	}
 	return arg
+}
+
+// rewrite_windows_path_arg rewrites any quoted path substring of a compiler
+// argument, in addition to the whole path operands handled by
+// rewrite_windows_path_operand_arg. tcc keeps this broader, pre-existing
+// rewrite, so its whole argument vector stays ASCII on Windows; the narrower
+// operand only rewrite is used for gcc, see tcc_windows_path_arg.
+fn rewrite_windows_path_arg(arg string, resolver WindowsPathResolver) string {
+	if arg == '' {
+		return ''
+	}
+	if start := arg.index('"') {
+		end := arg.last_index('"') or { -1 }
+		if end > start {
+			path := arg[start + 1..end]
+			if looks_like_windows_path(path) {
+				return arg[..start] + '"${resolver(path)}"' + arg[end + 1..]
+			}
+		}
+	}
+	return rewrite_windows_path_operand_arg(arg, resolver)
 }
 
 // cc_uses_short_windows_paths reports whether the given C compiler needs the
@@ -1577,10 +1595,20 @@ fn (v &Builder) tcc_windows_path(p string) string {
 	return p
 }
 
+// tcc_windows_path_arg rewrites the Windows path arguments of tcc and the
+// MinGW gcc toolchain to ASCII 8.3 short paths, so that non-ASCII project
+// paths survive the ANSI response file encoding (see issue #28126). tcc keeps
+// the broader rewrite_windows_path_arg, while gcc gets only its real
+// filesystem operands rewritten: gcc argument vectors can contain user
+// `CFLAGS` with path looking values (e.g. `-DROOT="C:\Program Files\SDK"`),
+// which must not be altered.
 fn (v &Builder) tcc_windows_path_arg(arg string) string {
 	$if windows {
-		if cc_uses_short_windows_paths(v.ccoptions.cc) {
+		if v.ccoptions.cc == .tcc {
 			return rewrite_windows_path_arg(arg, short_windows_path)
+		}
+		if v.ccoptions.cc == .gcc {
+			return rewrite_windows_path_operand_arg(arg, short_windows_path)
 		}
 	}
 	return arg
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -1874,8 +1874,32 @@ fn resolved_windows_ccompiler_path(ccompiler string) string {
 	return name
 }
 
+fn execute_windows_batch_ccompiler(cmd string) os.Result {
+	$if windows {
+		// `os.execute` expands percent-delimited environment variables before starting
+		// cmd.exe. Expand this indirection once inside cmd.exe instead, so percent signs
+		// carried by compiler arguments are not scanned again.
+		command_env_name := 'V_CCOMPILER_BATCH_COMMAND_${os.getpid()}'
+		old_command := os.getenv_opt(command_env_name)
+		os.setenv(command_env_name, cmd, true)
+		defer {
+			if previous := old_command {
+				os.setenv(command_env_name, previous, true)
+			} else {
+				os.unsetenv(command_env_name)
+			}
+		}
+		command_interpreter := os.getenv_opt('COMSPEC') or { 'cmd.exe' }
+		return os.exec([command_interpreter, '/d', '/v:off', '/s', '/c', '%${command_env_name}%'])
+	}
+	return os.execute(cmd)
+}
+
 fn (v &Builder) execute_ccompiler(ccompiler string, cmd string, exec_args []string) os.Result {
-	if exec_args.len > 0 && !ccompiler_is_windows_batch_file(ccompiler) {
+	if exec_args.len > 0 {
+		if ccompiler_is_windows_batch_file(ccompiler) {
+			return execute_windows_batch_ccompiler(cmd)
+		}
 		return os.exec(exec_args)
 	}
 	return os.execute(cmd)
--- a/vlib/v/builder/cc.v
+++ b/vlib/v/builder/cc.v
@@ -1529,21 +1529,37 @@ fn looks_like_windows_path(value string) bool {
 // Option values that merely contain a path looking substring are left
 // untouched, so that e.g. `-DROOT="C:\Program Files\SDK"` keeps its exact
 // macro value after the rewrite.
+fn single_windows_path_operand(value string) ?string {
+	trimmed := value.trim_space()
+	if trimmed.len >= 2 && trimmed[0] == `"` {
+		if trimmed[trimmed.len - 1] != `"` {
+			return none
+		}
+		return trimmed[1..trimmed.len - 1]
+	}
+	if trimmed.bytes().any(it.is_space()) {
+		return none
+	}
+	return trimmed
+}
+
 fn rewrite_windows_path_operand_arg(arg string, resolver WindowsPathResolver) string {
 	if arg == '' {
 		return ''
 	}
 	for prefix in ['-I', '-L', '-B', '-o ', '-c '] {
 		if arg.starts_with(prefix) {
-			path := arg[prefix.len..].trim_space().trim('"')
+			path := single_windows_path_operand(arg[prefix.len..]) or { return arg }
 			if looks_like_windows_path(path) {
 				return prefix + '"${resolver(path)}"'
 			}
 		}
 	}
-	trimmed := arg.trim_space().trim('"')
-	if !arg.starts_with('-') && looks_like_windows_path(trimmed) {
-		return '"${resolver(trimmed)}"'
+	if !arg.starts_with('-') {
+		path := single_windows_path_operand(arg) or { return arg }
+		if looks_like_windows_path(path) {
+			return '"${resolver(path)}"'
+		}
 	}
 	return arg
 }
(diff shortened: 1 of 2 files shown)


medvednikov:
Reviewed the code changes only; ignored CI as requested.

**1 finding — I’d request changes.**

**[P2] Limit GCC short-path rewriting to actual filesystem operands.** Once `.gcc` is added to `cc_uses_short_windows_paths`, the entire compiler argument vector is passed through `rewrite_windows_path_arg`.  That helper rewrites *any* quoted substring that merely looks like a Windows path, before checking whether the argument is actually a path-bearing compiler option.  Since compile args include user `CFLAGS`, this can mutate preprocessor values as well.

For example, on Windows:

```text
-DROOT="C:\Program Files\SDK"
```

can become something like:

```text
-DROOT="C:\PROGRA~1\SDK"
```

So this PR can change the program’s compiled macro value, even though that argument is data rather than a compiler filesystem operand. It also happens before deciding whether an `.rsp` file is used, so `-no-rsp` doesn’t avoid the new behavior.

I’d restrict the GCC rewrite to known path operands (`-o`, `-I`, `-L`, `-B`, source/object paths, etc.) rather than the generic first/last-quote branch, and add a regression test proving path-looking `-D...="..."` values remain unchanged.

Aside from that, the intended `.rsp` path is wired through the new GCC short-path logic, including the response-file filename itself, and the non-existent output-file case is handled by `os.short_path` shortening the existing parent directory.

HuntedByTheIRS:
Thanks for the careful review — the `-DROOT=\"C:\\Program Files\\SDK\"` case is a real data-corruption risk.

Fixed in commit `f38d8da2b3`: gcc now uses a new `rewrite_windows_path_operand_arg`, which rewrites **only** real filesystem operands:

- path-bearing options `-I`, `-L`, `-B`, `-o `, `-c `
- bare source/object/archive file paths

and leaves any argument that merely *contains* a path-looking quoted substring untouched. So `-DROOT=\"C:\\Program Files\\SDK\"` (and `-D_DEFAULT_SOURCE`, `-std=c99`, `-bt25`) pass through byte-for-byte unchanged.

tcc keeps the broader, pre-existing `rewrite_windows_path_arg` rewrite unchanged, so there is zero behavior change on the tcc path.

Added regression tests covering:
- `rewrite_windows_path_operand_arg` rewriting `-I`/`-L`/`-B`/`-o `/bare-object operands
- `rewrite_windows_path_operand_arg` leaving `-DROOT=\"...\"` path-like values alone (the case you flagged)
- a guard test documenting that the broader tcc rewrite still rewrites such quoted substrings (so the two functions cannot silently converge)

Verified: `vlib/v/builder/` 15 passed/1 skipped (Windows-only), `compiler_errors_test.v` 52/53 with the sole failure being the pre-existing environmental `skip_unused/auto_str.vv` case unrelated to this change.

antono3 (changes requested):
The follow-up fixes the `-DROOT=...` mutation and its helper tests pass, but a real Windows GCC reproduction still fails, so this does not yet resolve #28126 on a standard `windows-2025` runner.

I built V with GCC, created `D:\a\_temp\工作目录\main.v`, then ran:

```
v.exe -old-compiler -showcc -cc gcc -o main.exe main.v
```

The emitted response file still contained:

```
-o "D:\a\_temp\工作目录\main.exe"
```

and MinGW failed with:

```
ld.exe: cannot open output file D:\a\_temp\????\main.exe: Invalid argument
```

This is the `os.short_path()` fallback described in the caveat: when the existing parent directory has no 8.3 alias, it returns the original Unicode path. Because the issue itself concerns a `D:` path, relying on 8.3 availability leaves the reported failure intact on a common Windows environment.

Could the response-file encoding/path transport be fixed without requiring 8.3 aliases, or could there be another fallback for GCC when `short_path == original`? An end-to-end Windows regression using a non-ASCII directory would catch this; the current resolver-based unit tests cannot.

Failing Actions reproduction (the builder unit tests pass first): https://github.com/antono3/v/actions/runs/34107441679

medvednikov:
Addressed the requested no-8.3 fallback in 52462d0302.

After GCC path operands are selectively shortened, Windows now skips the ANSI response file if any argument is still non-ASCII. That is precisely the os.short_path fallback case from the review: the same arguments go through the existing direct CreateProcessW command line instead. ASCII and successfully shortened argument sets keep using response files, so long normal builds retain the current transport, and path-like macro data remains byte-for-byte unchanged.

Coverage now includes both the ASCII decision helper and a real Windows GCC build/run whose source and output live under a non-ASCII directory. Local checks passed: cc_tcc_retry_test.v; all vlib/v/builder tests (15 passed, 1 Windows-only skip on macOS); compiler_errors_test.v (1696 passed, 1 skipped).

medvednikov:
@codex review

medvednikov:
Addressed the compound CFLAGS/LDFLAGS finding in c662c98707. The selective GCC path rewrite now first proves that the suffix is exactly one operand: either one fully quoted path or one whitespace-free path. Compound environment entries such as -IC:\sdk -DFOO=1, -I"C:\Program Files\SDK" -DFOO=1, linker flags, and quoted bare paths followed by another flag are returned byte-for-byte unchanged. If one of those compound entries contains Unicode, the prior fallback still detects it and uses the direct wide command line.

Added regression cases for unquoted and quoted include flags, linker flags, and bare object paths followed by another flag. Rebuilt vnew from this worktree and the full cc_tcc_retry_test.v passes.

medvednikov:
@codex review

medvednikov:
Addressed the GNU C++ driver finding in d3fea2e6f1. On Windows, the existing cplusplus compiler classification now participates in the same selective 8.3 rewrite and non-ASCII direct-command fallback as gcc. This covers both g++ and the generic c++ driver without changing clang++, which is already identified as clang before the cplusplus fallback.

The Windows end-to-end regression now builds and runs from a non-ASCII directory with every available driver in gcc, g++, and c++. The routing assertions also cover the cplusplus/unknown-CC combination called out in the review. Rebuilt vnew from this worktree; the complete cc_tcc_retry_test.v passes locally.

medvednikov:
@codex review

medvednikov:
Addressed the single-quoted GCC operand finding in 6007336c97. The single-path parser now accepts either balanced double or single quotes, strips the matching pair before short-path resolution, and continues to normalize rewritten response-file operands to double quotes. Mismatched and compound values still remain untouched.

Added regression coverage for both a single-quoted -I operand and a single-quoted bare object path. Rebuilt vnew from this worktree; the focused regression, complete cc_tcc_retry_test.v, and compiler_errors_test.v (1696 passed, 1 skipped) all pass; touched files were formatted and git diff --check is clean.

medvednikov:
@codex review

medvednikov:
Addressed the cmd.exe length-limit finding in 7498f26793. When Windows GCC/GNU C++ still has non-ASCII arguments after 8.3 rewriting, the compiler is now launched with os.exec and an exact argument vector, which goes directly through CreateProcessW and avoids cmd.exe’s 8191-character ceiling. The existing response-file path remains unchanged for ASCII/shortened arguments, and other compiler launches retain the prior os.execute behavior.

The Windows end-to-end regression now deliberately generates more than 8191 bytes of C flags while building/running from a non-ASCII directory, covering the no-8.3 direct-launch path on affected systems. A pure regression also checks UTF-8 path preservation and splitting of quoted/compound option fragments. Rebuilt vnew from this worktree; focused tests, cc_tcc_retry_test.v, all builder tests (15 passed, 1 Windows-only skip locally), and compiler_errors_test.v (1696 passed, 1 skipped) pass; touched files are formatted and git diff --check is clean.

medvednikov:
@codex review

medvednikov:
Fixed the two latest review findings in 2200a390d645d81aa115184843d1a075f4fa51e9:

- doubled trailing backslashes before the generated closing quote so paths ending in `\` remain valid Windows command-line arguments
- parse backslash/quote runs using Windows argv rules and serialize exact GCC response-file arguments without dropping escaped literal quotes
- added focused regressions for trailing-separator path operands and `-DNAME=\"foo\"`

Validation passed:
- `VFLAGS="-gc none" ./vnew -silent test -run-only "test_rewrite_windows_path_operand_arg_rewrites_path_operands,test_ccompiler_exec_args_split_shell_formatted_options,test_gcc_response_file_content_quotes_exact_arguments" vlib/v/builder/cc_tcc_retry_test.v`
- `VFLAGS="-gc none" ./vnew -silent vlib/v/builder/cc_tcc_retry_test.v`
- `VFLAGS="-gc none" ./vnew -silent test vlib/v/builder/` (15 passed, 1 Windows-only skip)
- `VFLAGS="-gc none" ./vnew -silent vlib/v/compiler_errors_test.v` (1696 passed, 1 skip)
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Fixed the latest batch-wrapper review finding in c7ee8adb39a80a61668d17fa8765713e96841a8d:

- `.bat` and `.cmd` GNU compiler wrappers now retain the command-interpreter execution path instead of being passed to `os.exec`/CreateProcessW directly
- added case-insensitive, quoted-path regression coverage for both wrapper suffixes

Validation passed:
- rebuilt `./vnew`
- `VFLAGS="-gc none" ./vnew -silent vlib/v/builder/cc_tcc_retry_test.v`
- `VFLAGS="-gc none" ./vnew -silent test vlib/v/builder/` (15 passed, 1 Windows-only skip)
- `VFLAGS="-gc none" ./vnew -silent vlib/v/compiler_errors_test.v` (1696 passed, 1 skip)
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Fixed the latest command-length review finding in 44e1f02cc3dba0e70db4d47bc92b82f9ab6c7e45:

- split the exact GCC argument vector into ordered transport segments, keeping Unicode arguments on the wide direct command line while moving maximal ASCII runs into response files
- preserve argument order and use the same bounded transport for `.bat`/`.cmd` compiler wrappers
- add a regression with more than 32,767 bytes of ASCII flags around Unicode output/source paths, verifying the remaining command line stays below the batch limit

Validation passed:
- rebuilt `./vnew`
- focused large-command response-plan regression
- `./vnew -silent vlib/v/builder/cc_tcc_retry_test.v`
- `./vnew -silent test vlib/v/builder/` (15 passed, 1 Windows-only skip)
- `./vnew -silent vlib/v/compiler_errors_test.v` (1696 passed, 1 skip)
- `./vnew fmt -verify vlib/v/builder/cc.v`
- `git diff --check`

The current formatter also proposes unrelated pre-existing rewrites earlier in `cc_tcc_retry_test.v`; the newly added regression produces no formatter diff.

medvednikov:
@codex review

medvednikov:
Fixed the repeated-Unicode command-length finding in 21f1dcc2c9cd36b798d66f473c8439cc31cff609:

- account conservatively for Windows quoting/separators in the remaining direct transport
- retain ordered ANSI response segments plus direct Unicode arguments while they fit safely
- when repeated Unicode arguments would exceed the applicable CreateProcessW or cmd.exe budget, collapse the exact argument vector into one UTF-8 response file so no operand remains expanded on the command line
- add a regression with 1,200 Unicode cached-object paths totaling more than 32,767 bytes

Validation passed so far:
- rebuilt `./vnew`
- focused large-ASCII and repeated-Unicode response-plan regressions
- complete `vlib/v/builder/cc_tcc_retry_test.v`
- `./vnew fmt -verify vlib/v/builder/cc.v`
- `git diff --check`

Full builder and compiler-error checks continue during review.

medvednikov:
@codex review

medvednikov:
Addressed the latest review findings in b765af9812a62d4c82d25a61c6d23772d637691f:

- keep the active ANSI response encoding only when a round trip proves it lossless
- report an explicit error when an oversized Unicode command has no lossless GCC transport
- resolve extensionless compiler names through PATH before deciding whether `.cmd`/`.bat` requires `cmd.exe`
- added focused planner and Windows batch-resolution regressions

Validation: rebuilt `./vnew`; full `cc_tcc_retry_test.v`; `git diff --check`.

medvednikov:
@codex review

medvednikov:
Correction: the exact pushed head is `b765af9812c520b84a606edaf57532452732b147`.

medvednikov:
Addressed the latest review findings in 0ba6b95eeecd85afbbacc74b1b8e7bcb340df819:

- render batch-wrapper arguments with Windows backslash/quote escaping so embedded quotes survive `%*` forwarding
- size direct Unicode transport after the same exact Windows argument escaping used by `os.exec`
- add regressions for embedded quotes, trailing backslashes, and the escape-expanded command limit

Validation: rebuilt `./vnew`; full `cc_tcc_retry_test.v`; `git diff --check`.

medvednikov:
@codex review

medvednikov:
Fixed the exact-head review finding in e80d188ab9: batch compiler commands that require the Unicode-safe path now launch through cmd.exe with one non-recursive environment indirection, avoiding both ExpandEnvironmentStringsW and a second percent scan of literal arguments. Added a Windows-only .cmd regression with a defined KEEP variable and a Unicode `%KEEP%` path. Validation: rebuilt vnew; cc_tcc_retry_test passed; vlib/v/builder passed (15 passed, 1 skipped); Windows cross-target check passed; compiler_errors passed (1696 passed, 1 skipped); diff check clean.

medvednikov:
@codex review

medvednikov:
Fixed the exact-head response-encoding finding in ef6fb0ea37. Oversized Unicode commands now use a two-phase plan: split/direct plans do not probe; only an overflowing plan probes the selected GCC/C++ driver end-to-end through its real executable or batch wrapper and transformed @rsp path. UTF-8 is preferred only after a UTF-8 response opens a Unicode-named source; ANSI is selected only after the full content round-trips through the active code page and an ANSI probe succeeds. Probe/I/O failures no longer silently mean ANSI. Added decision-matrix, split/full, and byte-writer regressions. Validation: rebuilt vnew; full cc_tcc_retry_test and Windows cross-target check passed; diff check clean.

medvednikov:
@codex review

PR #28331, opened 2026-09-03
v3: speed up large program compilation

medvednikov:
## What this does

Speeds up large V3 builds across scanning, parsing, checking, monomorphization, and C generation.

The main changes are:

- tune production C flags for generated units of at least 8 MiB: `-O2 -flto`, plus a lower Clang inline threshold
- compile cached native support objects without LTO so they are not re-optimized during every final link
- cap parallel monomorphization for very large ASTs to reduce memory pressure and contention
- bound generic-clone scratch capacity instead of reserving space for the whole AST
- remove redundant whole-AST canonicalization passes
- reduce allocations and repeated work in scanner, parser, checker, transformer, name mangling, string interpolation, and C header processing
- preserve large VROOT Sokol/STB implementation headers as includes
- raise the ordinary V3 build memory guard to 4032 MiB while keeping the self-host limit at 3840 MiB

## Fusecode benchmark

Equivalent cache-disabled production builds used the same flags; only `-old-compiler` changed. Runs were alternated to reduce ordering bias.

| Backend | Runs | Average |
| --- | --- | --- |
| New/default | 27.99 s, 25.31 s | 26.65 s |
| Old (`-old-compiler`) | 30.27 s, 31.11 s | 30.69 s |

The optimized V3 compiler is 4.04 seconds, or 13.2%, faster on these runs. The host was under sustained background load, so the relative comparison is more useful than the absolute times.

Fusecode produces a 9,524,350-byte C unit and selects `-O2 -flto -mllvm -inline-threshold=75` with Clang.

## Validation

- production self-host build with `-gc none -d skip_fastc`
- `vlib/v/compiler_errors_test.v`: 1696 passed, 1 skipped
- `vlib/v3/gen/c/names_test.v`
- `vlib/v3/driver/environment_test.v`
- `vlib/v3/driver/c_compiler_flags_test.v`
- focused large-AST monomorph worker-limit test
- `check-md vlib/v3/README.md`
- `git diff --check`

The full `test-all` suite was not run.

### DIFF
--- a/vlib/v3/README.md
+++ b/vlib/v3/README.md
@@ -88,8 +88,11 @@ currently supported collector mode. Directory builds read `subdirs` through the 
 Native C compilation uses `-fwrapv` on supported targets so signed integer overflow retains V's
 two's-complement semantics. On macOS, `-cg` links executables with exported symbols for symbolic
 backtraces while plain `-g` retains its V-source debug behavior.
-The driver monitors compiler memory throughout the build and exits when it reaches 3840 MiB,
-leaving sampling headroom below a 4 GiB process ceiling.
+For `-prod`, generated C units of at least 8 MiB use `-O2 -flto`; Clang also uses a bounded inline
+threshold for those units. Smaller production units retain `-O3 -flto`.
+The driver monitors compiler memory throughout the build. Ordinary builds stop at 4032 MiB;
+compiler-tree and self-host builds stop at 3840 MiB, leaving extra sampling headroom below a
+4 GiB process ceiling.
 On macOS it uses physical footprint, matching Activity Monitor more closely; elsewhere it uses
 current RSS. Pass `-no-memory-limit`/`--no-memory-limit` to disable this safety limit.
 On macOS and Linux, `make` and the default `v self` build the compiler with `-prealloc`, enabling
--- a/vlib/v3/bench/bench.v
+++ b/vlib/v3/bench/bench.v
@@ -5,9 +5,9 @@ import runtime
 import sync
 import time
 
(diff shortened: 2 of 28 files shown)


medvednikov:
@codex review

medvednikov:
I found one blocking lifetime issue in the new unused-diagnostic prealloc scope.

In `vlib/v3/driver/driver.v`, `diagnose_unused_private_declarations()` can grow `pre_tc.notices` while `unused_diag_scope` is the current allocator. The code then calls `pre_tc.notices.clear()` and frees that scope. `clear()` only resets the array length; it retains the backing `data`/`cap`, so `pre_tc.notices` can be left pointing into freed scoped storage. The same checker is reused later by monomorphization/concrete semantic checking, and a later notice append can reuse that retained capacity and write into freed memory.

Please reset/rebind `pre_tc.notices` to an empty parent-owned array after leaving the child scope and before freeing it (or otherwise promote its backing storage), rather than only calling `.clear()`.

medvednikov:
Addressed in `1f429aae98`. The diagnostic teardown now leaves the child arena, clears and rebinds the notices array to zero capacity under the parent allocator, and only then frees the child scope. A prealloc regression test forces child-scope growth and verifies the reset plus a subsequent parent-owned append. I also formatted the PR-touched V3 sources flagged by CI.

medvednikov:
Re-reviewed the fix in `1f429aae98`. The blocker is resolved: `release_unused_diagnostic_scope` leaves the child scope, clears and rebinds `notices` to a fresh zero-cap array under the parent allocator, then frees the scoped arena. The new regression test also forces scoped growth and verifies a subsequent append is safe. No further blocking issues from my review. LGTM.

PR #28340, opened 2026-09-04
ci: fix current master failures

medvednikov:
Fixes the non-FastC failures visible on current master:

- emit V3 optional wrappers for result multi-return payloads, fixing the macOS multiwindow C build
- fall back to the checker shared-field index when parallel codegen has a stale short type name, fixing serialize_test on s390x and riscv64
- keep JSON-RPC Router method-value receivers heap-backed, fixing the Linux TinyCC runtime failure
- disable the V3 safety memory limit for the intentionally huge multiwindow proof tests
- apply current vfmt output to the two V3 transform files rejected by Linux clean-code checks

Validation (all local commands used VJOBS=1):

- ./vnew -silent vlib/v/compiler_errors_test.v (1696 passed, 1 skipped)
- ./vnew -new-compiler -cc tcc -d network -silent vlib/net/jsonrpc/server_test.v
- ./vnew -silent vlib/v3/tests/option_arg_codegen_test.v
- ./vnew -silent vlib/v3/gen/c/types_test.v
- exact macOS multiwindow V3 compilation with clang
- s390x C generation for vlib/encoding/binary/serialize_test.v; verified shared map stringification uses ->val
- vfmt verification for all touched V files
- YAML parse of linux_ci.yml

FastC failures are intentionally out of scope.

### DIFF
--- a/.github/workflows/linux_ci.yml
+++ b/.github/workflows/linux_ci.yml
@@ -521,7 +521,7 @@ jobs:
           for test in "${tests[@]}"; do
             "$RUNNER_TEMP/multiwindow_ci_watchdog" 300 - \
               env -u DISPLAY -u WAYLAND_DISPLAY -u XDG_SESSION_TYPE \
-              ./v -cc gcc -d gg_multiwindow test "$test"
+              ./v -no-memory-limit -cc gcc -d gg_multiwindow test "$test"
           done
 
           ./v -cc gcc -d gg_multiwindow \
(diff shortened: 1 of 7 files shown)


medvednikov:
Reviewed the non-CI changes only. Since the PR is already merged, this is a post-merge review.

**Verdict: LGTM — no blocking correctness issues found.**

* **JSON-RPC `Router` lifetime:** adding `@[heap]` and changing `handle_jsonrpc` to an `&Router` receiver is consistent with the router method escaping as the server callback. I don’t see a problematic ownership/lifetime consequence.
* **V3 result/multi-return codegen:** exempting generated `multi_return_*` payloads from the stale-qualified-struct rejection is appropriately narrow; these are generated tuple structs, not stale source structs.
* The accompanying regression exercises both execution and the generated C typedef for `!(int, Ticket)`, which covers the failure being fixed.
* **Shared-field fallback:** falling back to `tc.struct_field_is_shared()` when the generator's shared-field metadata misses looks reasonable. The value-selector path also has a second attempt using the type annotation retained on the AST node, which makes the stale-short-name handling less brittle.
* The large changes in the transform files appear predominantly formatter/string-literal normalization rather than new behavior.

**One non-blocking test gap:** I’d add a dedicated regression for the new shared-field fallback, ideally two modules with the same short struct name/field name where only one field is `shared`. That specifically guards against the dangerous failure mode here: resolving a stale short name to the wrong struct and emitting the wrong `->val` selector. I didn’t find evidence that this currently happens, so I would **not** block on it.

PR #28350, opened 2026-09-04
v3: preserve raw C header semantics

medvednikov:
Follow-up to #28348.

## Summary

- keep `@[typedef] C.*` declarations and C structs from header-backed plain `.v` files owned by native headers, using the V1 source-file heuristic without scanning headers
- reset top-level include deduplication after macro changes and conditional boundaries so repeated single-header implementation includes are preserved
- select Objective-C for raw include, insert, preinclude, and postinclude headers on macOS/iOS without opening them, while preserving the V1 TinyCC exemption
- use the same implicit native-input language for module-cache macro probes and the final generated translation unit
- restore focused compile/run regressions for header-owned structs, repeated includes across macro changes, and header-only Objective-C under direct `-new-compiler`

## Tests

- `./vnew -silent vlib/v3/gen/c/target_test.v`
- `./vnew -silent vlib/v3/tests/header_owned_c_struct_codegen_test.v`
- `./vnew -silent test vlib/v3/gen/c/` (18 passed)
- `./vnew -silent test vlib/v3/driver/` (6 passed)
- `./vnew -silent vlib/v3/driver/cache_prune_test.v`
- `./vnew -silent vlib/v/compiler_errors_test.v` (1696 passed, 1 skipped)

## Existing test issues

- `./vnew -silent vlib/v3/tests/c_directive_order_codegen_test.v` still fails on the merged #28348 base because that suite expects local `#insert` headers to be recursively inlined and removed from generated C.
- `./vnew -silent vlib/v/gen/c/coutput_test.v` reaches one stale V3 assertion in `test_c_fallback_decl_uses_c_helper_submodule_includes`, which still expects a declaration recovered by the header scanner removed in #28348.

This follow-up leaves both unrelated stale assertions unchanged.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -2564,7 +2564,7 @@ fn prepare_v3_cache_external_inputs(mut state V3ModuleCacheState, a &flat.FlatAs
 		cache_input_modules[module_name] = true
 	}
 	cache_input_modules['main'] = true
-	native_inputs_language := cgen.cache_native_inputs_language(a, prefs.vroot, user_c_flags, prefs.c99, prefs.target)
+	native_inputs_language := cgen.cache_native_inputs_language(a, prefs.vroot, user_c_flags, prefs.c99, prefs.ccompiler, prefs.target)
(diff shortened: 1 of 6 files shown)


medvednikov:
@codex review

PR #28351, opened 2026-09-04
v3: exceed 2M lines per second with uncached FastC

medvednikov:
This speeds up uncached V3 FastC self-host builds while preserving the existing compiler and linker semantics.

- compact self-host-only generated C names and avoid redundant type/declaration work
- reuse unchanged resolver memo stamps and tune parallel generation fragments
- prestart bundled TinyCC while FastC is still generating C
- render 12 balanced translation units directly into TinyCC stdin
- overlap completed object loading with compilation and keep ordered linker operands intact
- use native parallel hashing and mapped writes for macOS ad-hoc signing
- keep `-no-parallel` and `V3_FASTC_NO_PARALLEL` behavior intact

An optimized host compiling the unoptimized 79,877-line V3 target with `-nocache` completed FastC generation plus real TCC compilation, linking, and signing in 36.790 ms (2.171 MLOC/s).

Validation:

- `./vnew -silent vlib/v3/gen/fastc/macho_sign_test.v`
- uncached FastC self-host build, followed by compiling and running Hello World
- strict `codesign --verify` on the generated V3 and Hello World binaries
- `git diff --check`

Large test suites were skipped as requested. The broad `fastc_test.v` currently stops at the unchanged `test_selfhost_multi_return_interface_component_is_macro_safe` unnamed-receiver case that is also present on the base branch.

### DIFF
--- a/vlib/os/execute_capture_nix.h
+++ b/vlib/os/execute_capture_nix.h
@@ -176,3 +176,78 @@ static inline int v_os_exec_capture_start(char *const argv[], int *child_pid, in
 	return 0;
 }
 #endif
+
+#if defined(__APPLE__)
+static inline void v_os_fd_write_all(int fd, const char *data, size_t len) {
+	while (len > 0) {
+		ssize_t written = write(fd, data, len);
+		if (written <= 0) {
+			return;
+		}
+		data += written;
+		len -= (size_t)written;
+	}
+}
+
+// v_os_exec_capture_input_start is the argv-based capture helper with an
+// additional pipe connected to the child's stdin. F_SETNOSIGPIPE makes a
+// writer observe EPIPE instead of terminating the parent when a child exits
+// after reporting an early input error.
+static inline int v_os_exec_capture_input_start(char *const argv[], int *child_pid, int *read_fd, int *write_fd) {
+	if (argv == NULL || argv[0] == NULL) {
+		return -1;
+	}
+	int output_pipe[2];
+	int input_pipe[2];
+	if (pipe(output_pipe) != 0) {
+		return -1;
+	}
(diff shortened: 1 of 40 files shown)


medvednikov (commented):
I found a regression in the advertised single-job path on macOS. `fastc_tcc_job_count()` returns 1 for `-no-parallel`, `VJOBS=1`, or `V3_FASTC_NO_PARALLEL`, but uncached/debug self-host builds still prestart one TinyCC process. If that prestart matches the final args, `compile_v3_fastc_source()` calls `fastc_begin_render_prestarted_c_units()`. That calls `fastc_c_unit_plan()`, which intentionally returns an empty plan when `jobs < 2`; the prestarted renderer then rejects the layout (`plan.paths.len == 0` vs `prestarted.unit_count == 1`) and returns an error instead of falling back to the existing single-file compile path.

So on macOS a command such as an uncached self-host with `-no-parallel` (and likewise `V3_FASTC_NO_PARALLEL`, or `VJOBS=1`) can fail specifically because the new prestart path is active. Please skip/discard prestarting when `jobs < 2`, or treat the empty split plan as a normal fallback to the single-file path.

medvednikov (commented):
Re-reviewed the follow-up fix, ignoring CI as requested. The single-job macOS FastC regression is fixed: prestarting is now gated on `prestart_jobs >= 2`, so `-no-parallel`, `VJOBS=1`, and `V3_FASTC_NO_PARALLEL` skip the split/prestarted path and fall back to the existing single-file TinyCC flow. The follow-up commit only changes this guard/use of `prestart_jobs`; I don't see any new issue from the fix.

PR #28352, opened 2026-09-04
v3: speed up and stabilize large cached builds

medvednikov:
## What this does

- makes large V3 builds more reliable by bounding retained cache data and avoiding duplicate work
- overlaps native-input discovery with the V frontend for large builds
- skips generic placeholder parsing for already-concrete types and caches repeated type lookups
- keeps HTTP/3 and QUIC opt-in behind `-d http3`, preserving the smaller and faster HTTP default
- preserves ABI ownership for known included headers after V3 stopped recursively scanning C headers
- fixes option emission for qualified interface payloads

Together these changes allow V3 to compile the Brew.v project without the previous bus error or
conflicting C declarations.

## Performance

On Brew.v, using production compilers and `-nocache -o .c`:

- V1 median: approximately 0.45 s
- V3 median: approximately 1.04 s

The branch is approximately 34% faster than the current upstream V3 compiler in the matched debug
frontend benchmark (3.59 s versus 5.43 s). Native-input overlap saves a further approximately 3.4%
on the large-build path.

## Validation

- rebuilt `vnew` from the stable compiler
- compiled Brew.v with `-new-compiler -no-memory-limit`; resulting executable runs successfully
- `vlib/v/compiler_errors_test.v`: 1696 passed, 1 skipped
- V3 driver, type, transform, C-generation, and HTTP/3 gate tests pass
- C output suite: 98 output tests passed, 8 skipped; 133 must-have tests passed, 9 skipped
- full `vlib/v/` suite remains at the existing V3 baseline: 97 failed, 2251 passed, 26 skipped


medvednikov:
@codex review

medvednikov (commented):
One correctness gap in the new array callback hoist exception: `array_dsl_callback` only recognizes dynamic-array receivers via `callback_base_type.starts_with('[]')`. Fixed arrays (`[N]T`) support the same compiler-provided `filter`, `map`, `any`, `all`, and `count` `it` expressions, so an `if`/`match` callback on a fixed array still goes through the generic value-branch materialization path and can be hoisted before the generated loop, where `it` is not bound. Please include fixed-array receiver types here as well (for example with `t.is_fixed_array_type(callback_base_type)`) and add a fixed-array `if`/`match` callback regression test.

medvednikov (commented):
The fixed-array callback issue still appears unresolved on `c944749`. `array_dsl_callback` is still gated only by `callback_base_type.starts_with('[]')`, so `[N]T` receivers still miss the exception and can hoist an `if`/`match` callback before the generated loop where `it` is bound. I also don't see a fixed-array regression test in the follow-up commits. Please include fixed arrays here as well (e.g. `callback_base_type.starts_with('[]') || t.is_fixed_array_type(callback_base_type)`) and cover at least one fixed-array `if`/`match` callback case.

PR #28353, opened 2026-09-04
v3: dedupe monomorph signature type logging

medvednikov:
Follow-up to #28349.

## Summary

- record monomorph signature types once per specialization key
- discard private parallel-worker signature logs because the master registers every emitted specialization
- add a regression test for queue-time plus emission-time registration

## Tests

- `./vnew -silent vlib/v3/transform/monomorphize_test.v`
- `./vnew -d v3_no_parallel vlib/v3/transform/monomorphize_test.v`
- V3 transform directory: 8/9 test files pass; `transform_parallel_test.v` has four pre-existing assertion failures reproduced unchanged on an untouched `origin/master` archive with the same toolchain
- rebuilt `./vnew` from the updated compiler sources
- `git diff --check`


medvednikov:
@codex review

PR #28354, opened 2026-09-04
v3: speed up uncached self-host transforms

medvednikov:
Self-host transform's expansion estimator routes ordinary compiler functions with struct
initializers into serial lowering. In the measured build, 2,629 functions were deferred.
Use growable worker ASTs for self-host transforms, copy their bases concurrently, and allow
up to 12 workers when disposable scratch arenas are available. This removes the expensive
capacity scan while preserving the ability to grow for metadata-driven expansion.

Uncached builds of the v3 compiler entry also use the normal import resolver and retain the
transform arena for C generation, avoiding duplicate import discovery and metadata promotion.
Adds a regression comparing parallel and serial execution with 300 functions containing
large nested struct defaults.

Measured on an Apple M5 Max with 18 CPU cores, using a `-gc none -prealloc -prod` bootstrap.
From `vlib/v3`, the measured command was:

```sh
./v3 -nocache -building-v -o v4 v3.v
```

| Measurement | Before | Final run 1 | Final run 2 | Final run 3 |
| --- | ---: | ---: | ---: | ---: |
| Transform | 889.61 ms | 226.88 ms | 242.88 ms | 240.73 ms |
| Total minus CC | 1479.04 ms | 739.26 ms | 775.11 ms | 769.47 ms |

All three final runs meet the 400 ms transform and 800 ms non-CC targets. One earlier
workspace run reached 821 ms non-CC; timings vary with system load. Peak RSS was about
3,700 MB, compared with about 3,350 MB before.

Validation:

- Passed `./vnew -old-compiler -gc none vlib/v3/tests/selfhost_parallel_expansion_test.v`.
- Passed `./vnew -old-compiler -gc none vlib/v3/tests/selfhost_transform_regression_test.v`.
- Passed `./vnew -old-compiler -gc none vlib/v3/tests/transformer_parity_test.v`.
- Passed `./vnew -old-compiler -gc none vlib/v3/driver/eager_import_scan_test.v`.
- Built v4 through uncached self-hosting; v4 compiled and ran Hello World successfully.
- Ran `./vnew -old-compiler -gc none -cflags '-Wl,-stack_size,0x4000000' -silent test vlib/v3/transform/`:
  8 test files passed; `transform_parallel_test.v` retains the existing
  `test_pointer_formatted_interp_still_accounts_for_temp_hoisting` assertion failure.
  Reproduced that failure on unmodified base commit `e9bb2f0d85`.
- Ran formatting on the touched V files and `git diff --check`.

No language or public API changes. The full `test-all` suite was not run; validation focused
on the affected transform, import, and self-host paths.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -10034,7 +10034,8 @@ pub fn run(args []string) {
 			prealloc_scope_leave_for_v3(transform_scope)
 			retain_transform_scope := building_v && current_parallel_transform && backend == 'c'
 				&& !cache_state.manager.enabled && retained_transform_regions.len == 0
-				&& os.getenv('V3_RETAIN_TRANSFORM_SCOPE') != ''
+				&& (input_is_v3_compiler_entry(input_file)
+					|| os.getenv('V3_RETAIN_TRANSFORM_SCOPE') != '')
 			if retain_transform_scope {
 				// Cgen is the only remaining semantic consumer in this no-cache self-host
 				// path. Keep the typed transform arena alive through it instead of cloning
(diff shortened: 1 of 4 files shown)


medvednikov:
@codex review

PR #28355, opened 2026-09-04
parser: support unnamed method receivers

### LINKED ISSUE #28339: parser: importing a C-backed module (#flag/#include) breaks method declarations in the importing file: 'expecting type declaration'
### Describe the bug

When a file imports a module that contains C interop directives (`#flag` / `#include`, typically in `.c.v` files), **any** method declaration on a local type in the importing file fails to parse with `expecting type declaration` at the method receiver. Pure-V imports (`os`, `strings`, `math`, ...) do not trigger this.

The same bogus error also masks unresolved imports: `import nonexistent_module_xyz` plus a method declaration produces `expecting type declaration` at the receiver instead of a "cannot find module" error.

### Reproduction Steps

`repro.v`:

```v
import crypto.ecdsa

struct Foo {}

fn (Foo) work(x int) int {
	return x
}

fn main() {
	println(Foo{}.work(1))
}
```

Run: `v run repro.v`

Also reproducible with:
- an empty module in `~/.vmodules/fake` containing only `module fake` + `#flag -lm` + one `pub fn`;
- interface method implementations (`interface Worker { work(x int) int }` + `fn (Foo) work(...)`) instead of a plain method;
- value-receiver methods on empty structs referenced by a sum type.

### Expected Behavior

Compiles and prints `1`.

### Current Behavior

```
repro.v:5:8: error: expecting type declaration
    3 | struct Foo {}
    4 | 
    5 | fn (Foo) work(x int) int {
      |        ^
    6 |     return x
```

Removing the import, or importing a pure-V module instead, makes the same file compile and run. Deleting the V module cache (`~/.cache/v`) does not change the result; the failure is deterministic on this machine, including with `-old-compiler` and with `-no-memory-limit` (so it is not specific to the experimental V3 frontend, and not a side effect of the V3 memory limit).

### Possible Solution

The error comes from `vlib/v/parser/parse_type.v` (`name == ''` branch, ~line 798): when the program set includes a C-backed module, the receiver type (`Foo`) is apparently not found in the type table while parsing the method declaration. The same misreport for unresolvable imports hints that type registration silently degrades when import resolution is incomplete or involves C-directive modules.

Discovered while building a project against a local C-bindings module (V bindings for mlx-c). The practical impact is significant: a project importing any C-backed module cannot declare methods/interfaces/sum-type members in the importing files at all.

Environment details (`v doctor`):

```
|V full version      |V 0.5.2 5ad2d1fc334ae419d1fb86883d5762637b27a420.1e30cf3
|OS                  |macos, macOS, 26.6.2, 25G83
|Processor           |10 cpus, 64bit, little endian, Apple M5
|Memory              |1.87GB/24GB
|V executable        |/Users/aki/code/v/v
|V home dir          |OK, value: /Users/aki/code/v
|VMODULES            |OK, value: /Users/aki/.vmodules
|V git status        |1e30cf3b
```

(I searched existing issues for `expecting type declaration`; #25330 has a different trigger (`thread` field ordering) and is already closed.)

> [!NOTE]
>

medvednikov:
Fixes #28339

Bring the stable parser in line with V3 by accepting type-only method receivers such as `fn (Foo) work()`. The receiver is represented by an ignored `_` binding, so method dispatch and code generation still receive argument 0.

Adds a regression with a local C-backed imported module, matching the fallback path from the issue.

Tests:
- `./vnew -silent vlib/v/tests/unnamed_receiver_c_import_test.v`
- exact `crypto.ecdsa` reproduction and unresolved-import diagnostic
- `./vnew -silent test vlib/v/parser/`
- `./vnew -silent test vlib/v/` (attempted; stopped after unrelated existing failures in anon-function/fixed-array tests)

### DIFF
--- a/vlib/v/tests/unnamed_receiver_c_import_test.v
+++ b/vlib/v/tests/unnamed_receiver_c_import_test.v
@@ -2,7 +2,7 @@ import v.tests.testdata.unnamed_receiver_c_module
 
 struct Foo {}
 
-fn (_ Foo) work(value int) int {
+fn (Foo) work(value int) int {
 	return value
 }
 
--- a/vlib/v/tests/unnamed_receiver_c_import_test.v
+++ b/vlib/v/tests/unnamed_receiver_c_import_test.v
@@ -1,0 +1,11 @@
+import v.tests.testdata.unnamed_receiver_c_module
+
+struct Foo {}
+
+fn (_ Foo) work(value int) int {
+	return value
+}
+
+fn test_unnamed_receiver_with_c_backed_import() {
+	assert Foo{}.work(unnamed_receiver_c_module.identity(1)) == 1
+}
--- a/vlib/v/ast/str.v
+++ b/vlib/v/ast/str.v
@@ -126,7 +126,11 @@ pub fn (t &Table) stringify_fn_decl(node &FnDecl, cur_mod string, m2a map[string
 			f.write_string(node.receiver.typ.share().str() + ' ')
 			styp = styp[1..] // remove &
 		}
-		f.write_string(node.receiver.name + ' ')
+		is_type_only_receiver := node.receiver.name == '_' && node.receiver.type_pos.len > 0
+			&& node.receiver.pos.pos == node.receiver.type_pos.pos
+		if !is_type_only_receiver {
+			f.write_string(node.receiver.name + ' ')
+		}
 		styp = util.no_cur_mod(styp, cur_mod)
 		if t.new_int_fmt_fix && styp == 'int' {
 			styp = 'i32'
--- a/vlib/v/fmt/tests/unnamed_receiver_keep.vv
+++ b/vlib/v/fmt/tests/unnamed_receiver_keep.vv
@@ -1,0 +1,5 @@
+struct Receiver {}
+
+fn (Receiver) type_only() {}
+
+fn (_ Receiver) explicitly_ignored() {}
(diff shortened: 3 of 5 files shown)


medvednikov:
@codex review

medvednikov (commented):
The parser change looks plausible, but the regression test currently does not exercise the new syntax.

vlib/v/tests/unnamed_receiver_c_import_test.v
```diff
@@ -0,0 +1,11 @@
+import v.tests.testdata.unnamed_receiver_c_module
+
+struct Foo {}
+
+fn (_ Foo) work(value int) int {
```
medvednikov: This test uses `fn (_ Foo)`, which takes the existing named-receiver path. The new `is_type_only_receiver` branch only runs for the issue's syntax, `fn (Foo) work(...)`. As written, this test can pass even if the parser change is removed. Please make the regression use `fn (Foo) work(...)` (ideally the exact #28339 repro) so it fails on master and covers the fix.

medvednikov: Fixed in e8cacc37b4. The regression now uses the exact `fn (Foo)` syntax. This exposed that vfmt was rewriting it to `fn (_ Foo)`, so the formatter now preserves type-only receivers while continuing to preserve explicitly ignored receivers; a fmt keep fixture covers both forms.

medvednikov (commented):
Re-reviewed e8cacc37b4. The regression now uses the actual type-only receiver syntax (`fn (Foo)`), so it exercises the new parser branch, and the follow-up also preserves that spelling through vfmt while distinguishing it from an explicitly ignored receiver (`fn (_ Foo)`). That addresses my previous review point. I don't see a remaining blocker in the fix; CI is still pending/queued.

PR #28358, opened 2026-09-04
asm: add raw templates and Intel syntax

### LINKED ISSUE #28298: Inline asm{}: raw-template form, Intel-syntax modifier, conformance corpus (C backend)
### Describe the feature

Extend the C backend's inline `asm{}` support with:

1. A **raw‑template form** — a verbatim GNU inline‑asm string where V only parses and
   binds the operand list, leaving instruction text untouched.
2. An **Intel‑syntax block modifier** (`.intel_syntax noprefix` / `.att_syntax` wrap),
   which also removes the need for the current operand‑reorder hack.
3. A **conformance test corpus** for `asm{}` run under gcc / clang / tcc.
4. **Better parser diagnostics** for mistyped registers and unknown clobbers.

Instruction encoding, register allocation and scheduling stay with GCC/Clang — this is a
front‑end/codegen change, not a new assembler.

### Use Case

`vlib/crypto` is pure V but follows Go's *generic* (non‑assembly) code paths, so
symmetric throughput (AES‑GCM, ChaCha20‑Poly1305, SHA) and asymmetric latency (RSA,
P‑256) are multiples slower than OpenSSL / Mbed TLS. The accelerated instructions that
close that gap — `AESENC`, `PCLMULQDQ`, `SHA256RNDS2`, `ADCX`/`ADOX`, `VPSHUFB` — have no
high‑level expression and are only reachable via intrinsics or inline asm.

`vlib/math/bits/bits.amd64.v` already shows the intended pattern
(`asm amd64 { mulq … }` + `$if msvc { C._umul128 }` + portable fallback), but anything
more than a single instruction runs into the DSL's limits:

- The template parser (`vlib/v/parser/asm.v:142-206`) rejects any token outside
  `.name/.number/.string/.chartoken/.colon/.lsbr/.rcbr/.semicolon/.comma` with
  `p.error('invalid token in assembly block')`. `{` is a hard parse error, so **AVX‑512
  `{k1}{z}` mask and `{1to8}` broadcast operands cannot be written at all**.
- `vlib/v/gen/c/cgen.v:6321-6325` unconditionally moves the last operand to the front
  for x86 (assuming "last = destination"), which is wrong for `imul r,r/m,imm`,
  `vpternlogd`, `vgf2p8affineqb`, compares, `cmpxchg` → silently incorrect codegen.
- There is no regression corpus for `asm{}` beyond the small per‑arch `asm_test.*.v`.

This blocks writing hand‑tuned crypto kernels (AES‑GCM, GHASH, ChaCha20, Montgomery
multiply, constant‑time barriers) in‑tree, and it blocks the follow‑on work of adding
per‑arch `crypto/*_amd64.v` accelerated primitives with runtime‑dispatched fallbacks.

#### Pipeline (for reference; checkout `b338390253` / `c89f6d3101`)

| Stage | Location |
|---|---|
| Parse | `vlib/v/parser/asm.v:10` `asm_stmt`, `:307` `reg_or_alias`, `:420` `asm_addressing`, `:579` `asm_ios` |
| AST | `vlib/v/ast/ast.v:1868` `AsmStmt`, `:1886` `AsmTemplate`, `:1940` `AddressingMode`, `:1958` `AsmIO`, `:1998` `x86_with_number_register_list`, `:3191` `all_registers` |
| Check | `vlib/v/checker/checker.v:4276` `asm_stmt`, `:4341` `asm_arg`, `:4362` `asm_ios` |
| Cgen | `vlib/v/gen/c/cgen.v:6291` `asm_stmt`, `:6380` `asm_arg`, `:6512` `gen_asm_ios` |
| Siblings | `vlib/v/fmt/asm.v:8`, `cmd/tools/vast/vast.v:2047` |

Already adequate, leave alone: the `xmm/ymm/zmm/mm` register model (`ast.v:1998`);
constraint parsing for `= + & * %`, numbered `0-9`, `@cc` (`asm.v:592-641`), passed
through verbatim by `gen_asm_ios` (`cgen.v:6517`); IO alias binding and
`fail_if_immutable` on outputs (`checker.v:4362`).

#### 1. Raw‑template form (core of this request)

Spelling to be settled in review (`raw` keyword vs `@[raw]` attribute):

```v
asm amd64 raw {
    "vaesenc %[rk], %[state], %[state]\n\t"
    "vaesenclast %[rk2], %[state], %[state]\n\t"
    ; [state] "+x" (state)
    ; [rk] "x" (rk0)
      [rk2] "x" (rk1)
    ; cc
}
```

- **Parser:** when the block body is one or more string literals, capture them verbatim
  into a new `AsmTemplate.raw string` field (or a dedicated `AsmRawStmt` node); do not
  tokenize instructions. Still parse `; out ; in ; clobber` via the existing `asm_ios`.
- **Cgen (`asm_stmt`):** emit the raw strings directly inside
  `__asm__ [volatile] (...)`, then `: <outputs> : <inputs> : <clobbers>` from the parsed
  IO. Skip the operand swap and `asm_arg` re‑stringification for raw blocks.
- **Checker:** for raw blocks run `asm_ios` (type‑check + immutability + alias scope) but
  skip `asm_arg` / directive / operand‑count validation.
- **fmt / vast:** preserve raw strings unchanged; format only the IO section; dump the
  new field.
- **Effect:** instruction coverage becomes whatever the target's `gas` / LLVM‑MC accept
  — full AES/SHA/VAES/GFNI/AVX‑512 — with no V‑side instruction tables.

#### 2. Intel‑syntax block modifier

```v
asm amd64 intel { ... }
```

Emit `".intel_syntax noprefix\n\t"` before and `".att_syntax prefix\n\t"` after the
block (GCC and Clang both support this). When set, do not apply the operand swap. Also
fix the structured non‑intel path: limit the swap to a known set of 2‑operand mnemonics
rather than applying it unconditionally.

#### 3. Conformance corpus

Add to `vlib/v/slow_tests/assembly/`: ~100 snippets exercising raw + structured + intel
forms (AES round, GHASH Karatsuba reduce, ChaCha20 quarter‑round, a Montgomery multiply
step, a constant‑time `value_barrier`, one `{k}{z}` AVX‑512 op in raw form, each
addressing mode). Build and run under gcc, clang, and tcc (skipping forms tcc can't
assemble); assert on computed output, not just "it compiles".

#### 4. Diagnostics

- `asm.v:204` `else` branch and `reg_or_alias` (`asm.v:307`): when an unknown name is
  within edit distance 1–2 of a register in `scope.objects`, emit
  `unknown register \`xmm01\`; did you mean \`xmm1\`?` at the token position instead of
  silently producing an `AsmAlias` (`asm.v:322`).
- `checker.v` `asm_stmt`: validate each `clobbered` register name against
  `ast.all_registers(...)` for the block's arch.

#### Task breakdown

- [ ] T1 — AST: `AsmTemplate.raw` (or `AsmRawStmt`) + `AsmStmt.is_intel` / `is_raw`.
- [ ] T2 — Parser: `raw` / `intel` modifiers; verbatim capture; keep `asm_ios`.
- [ ] T3 — Checker: raw‑aware `asm_stmt`; clobber‑name validation.
- [ ] T4 — Cgen: emit raw verbatim; `.intel_syntax` wrap; suppress/narrow the operand
      swap.
- [ ] T5 — fmt + vast support.
- [ ] T6 — Parser diagnostics (near‑miss register hints).
- [ ] T7 — Conformance corpus + runner wiring.
- [ ] T8 — Docs: `asm{}` section — `raw`, `intel`, when to use which.
- [ ] T9 — Proof: rewrite a `bits.amd64.v` routine (or add
      `crypto.subtle.value_barrier`) using the raw form; confirm identical
      codegen / behaviour.

#### Acceptance criteria

- A raw `asm amd64 raw { … }` block using `vaesenc` / `pclmulqdq` / `vpshufb` compiles
  and runs correctly under `-cc gcc` and `-cc clang` on amd64.
- `asm amd64 intel { … }` compiles an Intel‑ordered body with matching results;
  `asm_test.amd64.v` still passes.
- The conformance corpus passes under gcc, clang and (where applicable) tcc in CI.
- A mistyped register name errors at the correct source position, not as a C‑compiler
  error on generated code.
- No behavioural change for existing `asm{}` blocks that don't use the new modifiers
  (`vlib/builtin/linux_bare/`, `vlib/math/bits/`, `vlib/v/slow_tests/assembly/`).

#### Estimate

~3–5 focused weeks for T1–T7; low blast radius. T8–T9 small.

### Other Information

**Non‑goals:**

- The native backend byte encoder (`vlib/v/gen/native/`) — reaching `gas` parity is
  person‑years; external `.S` files (a separate follow‑on) cover that need.
- MSVC‑proper (`-cc msvc`) — has no GNU inline asm on x64; served separately by
  external‑assembler / `ml64.exe` integration, or by clang‑cl.
- Structured (non‑raw) AVX‑512 mask/broadcast operands (`k0-k7` registers,
  `{k}{z}{1to8}` decorators) — deferred; the raw form covers the use case.
- FastC (`vlib/v3/gen/fastc/`) — no AST and no `asm{}` support today
  (`fastc/stmt.v:75` has no `.key_asm` branch; `fastc/source.v:522` excludes `.amd64.v`
  files). If asm support is added there later, the raw form is the only one it could
  support (verbatim token copy).

**Follow‑on issues (not this one):**

- External‑assembler build integration — `foo.gas.S` / `foo.masm.asm` /
  `foo.armasm64.asm` file convention + `ml64` / `armasm64` invocation in the builder,
  extending the suffix logic in `vlib/v/pref/should_compile.v:349`.
- `vlib/cpu` runtime CPU‑feature detection + a `value_barrier[T]` intrinsic.
- `vlib/crypto` accelerated primitives — ChaCha20‑Poly1305 (AVX2) first, then AES‑GCM
  (AES‑NI + CLMUL), SHA‑256 (SHA‑NI), each a `*_amd64.v` / `*_arm64.v` sibling with a
  runtime‑dispatched portable fallback.
- Bignum core — `mul64x64_128` / `addc64` abstraction (`__uint128_t` / `_umul128`),
  constant‑time Montgomery, P‑256, RSA; consider adopting fiat‑crypto for the field
  layer.

I can implement this feature request.

### Version used

V 0.5.2 (vlib checkout `b338390253` / `c89f6d3101`, 2026‑09‑02/03)

Affects all C‑backend targets that use GNU inline asm (gcc / clang / clang‑cl) on any
OS; developed/tested primarily on Windows 11 x86‑64 and Linux x86‑64. Not applicable to
`-cc msvc`, `-cc tcc`, the native backend, or the wasm backend.

> [!NOTE]
>

medvednikov:
Fixes #28298.

This extends C-backend inline assembly with:

- `asm <arch> raw` templates passed through to GNU inline asm while retaining typed input/output checking
- `asm i386|amd64 intel` blocks wrapped in `.intel_syntax noprefix` / `.att_syntax prefix`
- correct destination-first to AT&T operand reversal for structured multi-operand x86 instructions
- GNU named operand syntax (`[name] "constraint" (expression)`)
- architecture-aware unknown-register and clobber diagnostics with near-match suggestions
- formatter and `vast` support, documentation, and regression coverage including AVX-512 mask syntax

Tests:
- compiler errors: 1698 passed, 1 skipped
- parser suite: 9 passed, 1 skipped
- checker suite: 4 passed
- compiler diagnostic fixtures: 95 files, 0 errors
- C output suite: 240 fixtures checked, 0 errors
- vfmt suite: 90 passed
- raw/Intel x86 regression built and ran with GCC and Clang
- existing amd64 assembly regressions built and ran with Clang
- `vast` compiled and serialized the new AST fields
- `./vnew check-md doc/docs.md`
- `./vnew -silent test vlib/v/` attempted; stopped after the unrelated existing `vlib/v/tests/anon_fn_map_assign_test.v` failure reproduced

Bundled TCC cannot perform an x86 cross-assembly run on this ARM64 macOS host (`movl` is reported as an unsupported ARM64 instruction); V’s normal compiler fallback succeeds. Native TCC coverage remains available on x86 CI.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -6305,8 +6305,15 @@ fn (mut g Gen) asm_stmt(stmt ast.AsmStmt) {
 	if stmt.templates.len == 0 {
 		g.writeln('""')
 	}
+	if stmt.is_intel {
+		g.writeln('".intel_syntax noprefix\\n\\t"')
+	}
 	for template_tmp in stmt.templates {
 		mut template := template_tmp
+		if stmt.is_raw {
+			g.writeln('"${template.raw_template}"')
+			continue
+		}
 		g.write('"')
 		if template.is_directive {
 			g.write('.')
@@ -6317,11 +6324,11 @@ fn (mut g Gen) asm_stmt(stmt ast.AsmStmt) {
 		} else {
 			g.write(' ')
 		}
-		// swap destination and operands for att syntax, not for arm64
-		if template.args.len != 0 && !template.is_directive
+		// V's structured x86 assembly uses destination-first operand order. AT&T uses
+		// the reverse order; Intel blocks already have the order expected by the assembler.
+		if template.args.len > 1 && !template.is_directive && !stmt.is_intel
 			&& stmt.arch !in [.arm64, .s390x, .ppc64le, .loongarch64, .rv64, .rv32] {
-			template.args.prepend(template.args.last())
-			template.args.delete(template.args.len - 1)
+			template.args.reverse_in_place()
 		}
 
 		for i, arg in template.args {
@@ -6340,6 +6347,9 @@ fn (mut g Gen) asm_stmt(stmt ast.AsmStmt) {
 			g.write('\\n\\t')
 		}
 		g.writeln('"')
+	}
+	if stmt.is_intel {
+		g.writeln('".att_syntax prefix\\n\\t"')
 	}
 
 	if stmt.output.len != 0 || stmt.input.len != 0 || stmt.clobbered.len != 0 || stmt.is_goto {
@@ -6394,7 +6404,9 @@ fn (mut g Gen) asm_arg(arg ast.AsmArg, stmt ast.AsmStmt) {
 			g.write("'${arg.val}'")
 		}
 		ast.IntegerLiteral {
-			if stmt.arch == .arm64 {
+			if stmt.is_intel {
+				g.write(arg.val)
+			} else if stmt.arch == .arm64 {
 				g.write('#${arg.val}')
 			} else if stmt.arch in [.s390x, .ppc64le, .loongarch64, .rv64, .rv32] {
 				g.write('${arg.val}')
@@ -6403,17 +6415,25 @@ fn (mut g Gen) asm_arg(arg ast.AsmArg, stmt ast.AsmStmt) {
 			}
 		}
 		ast.FloatLiteral {
-			if g.pref.nofloat {
+			if stmt.is_intel {
+				g.write(if g.pref.nofloat { arg.val.int().str() } else { arg.val })
+			} else if g.pref.nofloat {
 				g.write('\$${arg.val.int()}')
 			} else {
 				g.write('\$${arg.val}')
 			}
 		}
 		ast.BoolLiteral {
-			g.write('\$${arg.val.str()}')
+			if stmt.is_intel {
+				g.write(arg.val.str())
+			} else {
+				g.write('\$${arg.val.str()}')
+			}
 		}
 		ast.AsmRegister {
-			if stmt.arch in [.rv64, .rv32] {
+			if stmt.is_intel {
+				g.write(arg.name)
+			} else if stmt.arch in [.rv64, .rv32] {
 				g.write('${arg.name}')
 			} else if stmt.arch == .loongarch64 {
 				g.write('$${arg.name}')
@@ -6425,6 +6445,10 @@ fn (mut g Gen) asm_arg(arg ast.AsmArg, stmt ast.AsmStmt) {
 			}
 		}
 		ast.AsmAddressing {
+			if stmt.is_intel {
+				g.asm_addressing_intel(arg, stmt)
+				return
+			}
 			if arg.segment != '' {
 				g.write('%%${arg.segment}:')
 			}
@@ -6507,6 +6531,49 @@ fn (mut g Gen) asm_arg(arg ast.AsmArg, stmt ast.AsmStmt) {
 			g.write(arg)
 		}
 	}
+}
+
+fn (mut g Gen) asm_addressing_intel(arg ast.AsmAddressing, stmt ast.AsmStmt) {
+	if arg.segment != '' {
+		g.write('${arg.segment}:')
+	}
+	g.write('[')
+	match arg.mode {
+		.base {
+			g.asm_arg(arg.base, stmt)
+		}
+		.displacement {
+			g.asm_arg(arg.displacement, stmt)
+		}
+		.base_plus_displacement, .rip_plus_displacement {
+			g.asm_arg(arg.base, stmt)
+			g.write(' + ')
+			g.asm_arg(arg.displacement, stmt)
+		}
+		.index_times_scale_plus_displacement {
+			g.asm_arg(arg.index, stmt)
+			g.write(' * ${arg.scale} + ')
+			g.asm_arg(arg.displacement, stmt)
+		}
+		.base_plus_index_plus_displacement {
+			g.asm_arg(arg.base, stmt)
+			g.write(' + ')
+			g.asm_arg(arg.index, stmt)
+			g.write(' + ')
+			g.asm_arg(arg.displacement, stmt)
+		}
+		.base_plus_index_times_scale_plus_displacement {
+			g.asm_arg(arg.base, stmt)
+			g.write(' + ')
+			g.asm_arg(arg.index, stmt)
+			g.write(' * ${arg.scale} + ')
+			g.asm_arg(arg.displacement, stmt)
+		}
+		.invalid {
+			g.error('invalid addressing mode', arg.pos)
+		}
+	}
+	g.write(']')
 }
 
 fn (mut g Gen) gen_asm_ios(ios []ast.AsmIO) {
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -6396,6 +6396,10 @@ fn (mut g Gen) asm_arg(arg ast.AsmArg, stmt ast.AsmStmt) {
 				|| (name !in stmt.input.map(it.alias) && name !in stmt.output.map(it.alias)) {
 				asm_formatted_name := if name in stmt.global_labels { '%l[${name}]' } else { name }
 				g.write(asm_formatted_name)
+			} else if stmt.is_intel {
+				// `%V` asks GCC and Clang to substitute an x86 register without the `%`
+				// prefix required by AT&T syntax.
+				g.write('%V[${name}]')
 			} else {
 				g.write('%[${name}]')
 			}
--- a/cmd/tools/vast/vast.v
+++ b/cmd/tools/vast/vast.v
@@ -2051,6 +2051,8 @@ fn (t Tree) asm_stmt(node ast.AsmStmt) &Node {
 	obj.add_terse('is_basic', t.bool_node(node.is_basic))
 	obj.add_terse('is_volatile', t.bool_node(node.is_volatile))
 	obj.add_terse('is_goto', t.bool_node(node.is_goto))
+	obj.add_terse('is_raw', t.bool_node(node.is_raw))
+	obj.add_terse('is_intel', t.bool_node(node.is_intel))
 	obj.add('scope', t.scope(node.scope))
 	// obj.add('scope', t.number_node(int(node.scope)))
 	obj.add('pos', t.pos(node.pos))
@@ -2069,6 +2071,7 @@ fn (t Tree) asm_register(node ast.AsmRegister) &Node {
 	obj.add_terse('name', t.string_node(node.name))
 	obj.add_terse('typ', t.type_node(node.typ))
 	obj.add_terse('size', t.number_node(node.size))
+	obj.add('pos', t.pos(node.pos))
 	return obj
 }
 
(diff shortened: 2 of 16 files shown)


vlib/v/gen/c/cgen.v
```diff
@@ -6305,8 +6305,15 @@ fn (mut g Gen) asm_stmt(stmt ast.AsmStmt) {
 	if stmt.templates.len == 0 {
 		g.writeln('""')
 	}
+	if stmt.is_intel {
+		g.writeln('".intel_syntax noprefix\\n\\t"')
```
medvednikov, line 6309: `.intel_syntax noprefix` changes the assembler parser, but it does not change how GCC/Clang format extended-asm operand placeholders. Structured operands still emit `%[name]`; under Clang 17 a `+r`/`r` example is instantiated as `add %eax, %ecx` and the integrated assembler rejects `%eax` in `noprefix` mode. GCC happens to accept register prefixes here, but an `"m"` constraint is instantiated as AT&T memory syntax (for example `-4(%rbp)`) and fails under Intel syntax too. So `asm amd64 intel` is not safe for extended operands as implemented. Please either constrain/reject unsupported operand forms (and document it), or make operand rendering use a dialect-compatible mechanism; a Clang regression plus an `m`-constraint case would catch this.

medvednikov (commented):
Found one correctness/portability issue in the new Intel mode.

vlib/v/gen/c/cgen.v
medvednikov: Fixed in 07791ab342. Structured Intel operands now use the GNU x86 `%V[name]` modifier, which emits bare register names for both Clang and GCC. The checker restricts structured Intel operands to register-only `r` constraints and reports `m`/other constraints, since those can still expand with AT&T addressing; raw Intel remains available for explicit compiler-specific modifiers. I added a forced-Clang amd64 regression and a checker fixture for the memory-constraint rejection.

medvednikov (commented):
Re-reviewed 07791ab342. This addresses my Intel-mode portability concern: structured Intel aliases now use `%V[name]` for bare GPR names, non-`r` extended operands are rejected conservatively, and there is both a forced-Clang regression and an `m`-constraint checker fixture. The docs also spell out the restriction and the raw-Intel escape hatch. I don't see a remaining blocker in this fix; CI is still pending/queued.

quaesitor-scientiam:
**Post-merge test note — OpenSSL-style scalar kernel ports cleanly on x86-64**

Ported OpenSSL's scalar `poly1305_blocks` inner sequence (36 instructions: `mul`/`imul` with implicit `rdx:rax`, multi-limb carry chains) verbatim into `asm amd64 raw { }`, verified against the RFC 8439 §2.5.2 test vector.

Against master `c0e47bf`:

| Backend | ABI | Result |
|---|---|---|
| gcc (Windows) | Win64 | ✅ correct tag, `-prod` and not |
| clang / llvm-mingw (Windows) | Win64 | ✅ |
| tcc (Windows) | Win64 | ✅ |
| gcc 16 (Linux) | SysV | ✅ |

Zero friction:
- Verbatim AT&T paste — `%[name]` substitution, hardcoded `%%rax`, `$imm`.
- 9 `"r"` operands + `rax`/`rdx` + 3 hard scratch regs listed in the clobber section — every backend, tcc included, found a valid allocation at that pressure.
- Callee-saved `r12` correctly saved/restored around the block by all three compilers.
- Scratch registers work best as plain clobbered hard regs (`%%r10`/`%%r11`/`%%r12`) referenced in the template — no need to thread throwaway `+r` operands, which also makes the port a mechanical substitution of perlasm's `$d1/$d2/$d3` aliases.

Not yet exercised (likely the next questions for crypto use):
- Loops / local labels inside a `raw` block (kept the per-16-byte loop in V here).
- Whole-register-file SIMD kernels (AES-GCM scale) and `call` to subroutines from inside a block.

The asm block:

```v
asm amd64 raw {
    "add  %[m0], %[h0]\n\t"
    "adc  %[m1], %[h1]\n\t"
    "adc  %[hi], %[h2]\n\t"
    "mov  %[r1], %%rax\n\t"
    "mul  %[h0]\n\t"
    "mov  %%rax, %%r11\n\t"        // d2
    "mov  %[r0], %%rax\n\t"
    "mov  %%rdx, %%r12\n\t"        // d3
    "mul  %[h0]\n\t"
    "mov  %%rax, %[h0]\n\t"
    "mov  %[r0], %%rax\n\t"
    "mov  %%rdx, %%r10\n\t"        // d1
    "mul  %[h1]\n\t"
    "add  %%rax, %%r11\n\t"
    "mov  %[s1], %%rax\n\t"
    "adc  %%rdx, %%r12\n\t"
    "mul  %[h1]\n\t"
    "mov  %[h2], %[h1]\n\t"
    "add  %%rax, %[h0]\n\t"
    "adc  %%rdx, %%r10\n\t"
    "imul %[s1], %[h1]\n\t"
    "add  %[h1], %%r11\n\t"
    "mov  %%r10, %[h1]\n\t"
    "adc  $0, %%r12\n\t"
    "imul %[r0], %[h2]\n\t"
    "add  %%r11, %[h1]\n\t"
    "mov  $-4, %%rax\n\t"
    "adc  %[h2], %%r12\n\t"
    "and  %%r12, %%rax\n\t"
    "mov  %%r12, %[h2]\n\t"
    "shr  $2, %%r12\n\t"
    "and  $3, %[h2]\n\t"
    "add  %%r12, %%rax\n\t"
    "add  %%rax, %[h0]\n\t"
    "adc  $0, %[h1]\n\t"
    "adc  $0, %[h2]\n\t"
    ; [h0] "+r" (h0)
      [h1] "+r" (h1)
      [h2] "+r" (h2)
    ; [r0] "r" (r0)
      [r1] "r" (r1)
      [s1] "r" (s1)
      [m0] "r" (m0)
      [m1] "r" (m1)
      [hi] "r" (hibit)
    ; rax
      rdx
      r10
      r11
      r12
      cc
}
```

Thanks for landing this — the `raw` escape hatch is already enough for scalar bignum crypto (Poly1305, X25519 field arithmetic, `bn_mul_mont` core) on x86-64.

PR #28359, opened 2026-09-05
fasthttp: harden framing and connection state

medvednikov:
## Summary

- make the pure HTTP/1.1 framer authoritative for decoding and all three server backends
- reject conflicting Content-Length values, Transfer-Encoding plus Content-Length, non-final/unsupported transfer codings, empty chunk sizes, invalid chunk CRLF, and offset arithmetic overflow
- preserve and serially dispatch pipelined request suffixes on kqueue and IOCP
- distinguish blocked, drained, and closed Linux writes so a later response cannot cross a pending file response
- preserve complete requests and queued responses across read-side EOF
- pause reads while responses are blocked, add BSD write deadlines, and defer kqueue connection frees until the current event batch is consumed
- wire per-worker state through IOCP and make fields shared with housekeeping atomic
- add an independent 64 MiB default body limit, return 431 for header-limit failures, and cap retained Linux pool buffers
- cover overflow/framing cases, Linux file backpressure and half-close behavior, and BSD pipelining plus half-close behavior

## Validation

- `./vnew -silent test vlib/fasthttp/` (7 passed, 2 platform-specific tests skipped on macOS)
- `./vnew -cc clang -cflags '-fsanitize=undefined' -ldflags '-fsanitize=undefined' -silent vlib/fasthttp/framing_test.v`
- `./vnew -silent vlib/veb/tests/large_payload_test.v`
- `./vnew -silent vlib/veb/tests/persistent_connection_test.v`
- `./vnew check-md vlib/fasthttp/README.md`
- generated C successfully for the Linux and Windows fasthttp regression tests

Linux and Windows runtime tests were not available on the macOS host; their platform-specific regression programs were cross-compiled through C generation.

### DIFF
--- a/vlib/fasthttp/README.md
+++ b/vlib/fasthttp/README.md
@@ -164,8 +164,9 @@ fn handle(req fasthttp.HttpRequest, mut out []u8, ws voidptr, mut ctl fasthttp.R
 ## Configuration
 
 `ServerConfig` fields: `family` (`.ip` / `.ip6`), `port`,
-`max_request_buffer_size` (bounds the request head; an oversized head gets `413`),
-`timeout_in_seconds` (read/write deadlines; a stalled request gets `408`),
+`max_request_buffer_size` (bounds the request head; an oversized head gets `431`),
+`max_request_body_size` (defaults to 64 MiB; set to `0` only to allow unlimited buffered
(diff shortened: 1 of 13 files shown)


medvednikov:
@codex review

PR #28360, opened 2026-09-05
orm: emit backtick delimited `default:` values as quoted SQL literals (fix #27987)

### LINKED ISSUE #27987: ## Bug: V ORM `default:` attribute strips string quotes, generating invalid SQL
### Minimal Reproduction

**Model** (`model.v`):
```v
module main
import time

@[table: 'demo_default']
pub struct DemoDefault {
    id        string    @[primary; sql_type: 'CHAR(36)']
    home_path string    @[default: '/dashboard'; sql_type: 'VARCHAR(255)']
    tags      string    @[default: '["all"]'; sql_type: 'VARCHAR(1000)']
    name      string    @[default: '"我的团队"'; sql_type: 'VARCHAR(100)']
    method    string    @[default: 'POST'; sql_type: 'VARCHAR(32)']
    created_at time.Time @[sql_type: 'TIMESTAMP']
}
```

**Init** (`main.v`):
```v
module main
import db.pg

fn main() {
    mut db := pg.connect(host: 'localhost', user: 'test', password: 'test', dbname: 'test')!

    sql db {
        create table DemoDefault
    }!
    println('OK')
}
```

**Generated DDL (from `-d trace_orm`):**
```sql
CREATE TABLE IF NOT EXISTS "demo_default" (
    "id" CHAR(36) NOT NULL,
    "home_path" VARCHAR(255) DEFAULT /dashboard NOT NULL,   -- ← unquoted, syntax error
    "tags" VARCHAR(1000) DEFAULT ["all"] NOT NULL,           -- ← unquoted, syntax error
    "name" VARCHAR(100) DEFAULT "我的团队" NOT NULL,          -- ← unquoted, syntax error
    "method" VARCHAR(32) DEFAULT POST NOT NULL,              -- ← treated as column ref
    "created_at" TIMESTAMP NOT NULL,
    PRIMARY KEY("id")
);
```

**Error (PostgreSQL):**
```
ERROR:  syntax error at or near "/"
LINE 1: ... VARCHAR(255) DEFAULT /dashboard NOT NULL...
```

### Root Cause

V ORM's `default:` attribute renders the V expression value directly into SQL without SQL string escaping:

- `default: '/dashboard'` → `DEFAULT /dashboard` (missing quotes around string)
- `default: 'POST'` → `DEFAULT POST` (treated as column reference in PG)
- `default: 0` → `DEFAULT 0` (works because numbers don't need quoting)

The `default:` value is a **V expression** (V string literal `'/dashboard'`), but it needs to be rendered as a **SQL literal** (`'/dashboard'` with quotes) in the DDL.

### Expected Behavior

V string defaults should be properly quoted in generated SQL:
```sql
"home_path" VARCHAR(255) DEFAULT '/dashboard' NOT NULL
"tags" VARCHAR(1000) DEFAULT '["all"]' NOT NULL
```

Or alternatively, support `default: sql('/dashboard')` for raw SQL expressions.

### Workaround

Move the default into `sql_type`:
```v
home_path string @[sql_type: "VARCHAR(255) DEFAULT '/dashboard'"]
```
This works because `sql_type` is rendered as raw SQL — but it's fragile and mixes type definition with default value.

> [!NOTE]
>

medvednikov:
Fixes #27987.

### Problem

`[default: 'raw_sql']` is documented as inserting `raw_sql` **verbatim** into the `DEFAULT` clause — that is what makes `CURRENT_TIME`, `CURRENT_TIMESTAMP` and `gen_random_uuid()` work, and `vlib/db/pg`, `vlib/db/mysql` and `vlib/db/sqlite` all have tests relying on it.

The README additionally documents that *plain string* defaults should be surrounded with backticks — but that was never implemented. So a string default reached the DDL unquoted:

```sql
CREATE TABLE IF NOT EXISTS "demo_default" (
    "home_path" VARCHAR(255) DEFAULT /dashboard NOT NULL,   -- syntax error
    "method"    VARCHAR(32)  DEFAULT POST       NOT NULL,   -- parsed as a column reference
    ...
);
```

PostgreSQL rejects the first with `syntax error at or near "/"`. The only workaround was to smuggle quotes in by hand (`@[default: '"yes"']`), which produces a double-quoted *identifier* in PostgreSQL rather than a string literal.

### Fix

Implement the documented backtick form. A `default:` value surrounded by backticks is a plain string and is emitted as a properly quoted SQL string literal, with embedded single quotes doubled. Everything else keeps its verbatim raw-SQL meaning, so no existing struct changes behaviour.

```v ignore
struct DemoDefault {
    home_path  string @[default: '`/dashboard`']    // DEFAULT '/dashboard'
    tags       string @[default: '`["all"]`']       // DEFAULT '["all"]'
    quoted     string @[default: "`o'brien`"]       // DEFAULT 'o''brien'
    created_at string @[default: 'CURRENT_TIMESTAMP'] // DEFAULT CURRENT_TIMESTAMP  (unchanged)
    amount     int    @[default: 42]                  // DEFAULT 42                  (unchanged)
}
```

### Tests

- `vlib/orm/orm_fn_test.v`: new `test_orm_table_gen_string_defaults()` unit test over `orm_table_gen()`, covering the literal, escaping, empty and verbatim cases.
- `vlib/orm/orm_default_string_literal_test.v`: end-to-end test that creates the table in an in-memory SQLite DB, asserts the stored DDL in `sqlite_master`, and round-trips the default values back through a `select`.

Both fail without the change and pass with it. All 40 `vlib/orm` tests pass.

### DIFF
--- a/vlib/orm/orm.v
+++ b/vlib/orm/orm.v
@@ -1362,6 +1362,20 @@ fn parse_table_attr_fields(table Table, attr VAttribute, valid_sql_field_names [
 	return attr_fields
 }
 
+// sql_string_literal renders `value` as a quoted SQL string literal.
+// Doubling single quotes is standard SQL. MySQL additionally treats backslash as
+// an escape character inside string literals, unless the server runs with
+// NO_BACKSLASH_ESCAPES, so a literal backslash has to be doubled there too -
+// otherwise a value like `C:\tmp\new` would be stored with a tab and a newline
+// in it.
+fn sql_string_literal(sql_dialect SQLDialect, value string) string {
+	mut escaped := value
+	if sql_dialect == .mysql {
+		escaped = escaped.replace('\\', '\\\\')
+	}
+	return "'" + escaped.replace("'", "''") + "'"
+}
+
 pub fn orm_table_gen(sql_dialect SQLDialect, table Table, q string, defaults bool, def_unique_len int, fields []TableField, sql_from_v fn (int) !string,
 	alternative bool) !string {
 	mut str := 'CREATE TABLE IF NOT EXISTS ${q}${table.name}${q} ('
@@ -1535,7 +1549,7 @@ pub fn orm_table_gen(sql_dialect SQLDialect, table Table, q string, defaults boo
 		stmt = '${q}${field_name}${q} ${col_typ}'
 		if defaults && has_default {
 			if default_is_literal {
-				stmt += " DEFAULT '${default_val.replace("'", "''")}'"
+				stmt += ' DEFAULT ${sql_string_literal(sql_dialect, default_val)}'
 			} else if default_val != '' {
 				stmt += ' DEFAULT ${default_val}'
 			} else {
--- a/vlib/orm/orm.v
+++ b/vlib/orm/orm.v
@@ -1363,17 +1363,19 @@ fn parse_table_attr_fields(table Table, attr VAttribute, valid_sql_field_names [
 }
 
 // sql_string_literal renders `value` as a quoted SQL string literal.
-// Doubling single quotes is standard SQL. MySQL additionally treats backslash as
-// an escape character inside string literals, unless the server runs with
-// NO_BACKSLASH_ESCAPES, so a literal backslash has to be doubled there too -
-// otherwise a value like `C:\tmp\new` would be stored with a tab and a newline
-// in it.
-fn sql_string_literal(sql_dialect SQLDialect, value string) string {
-	mut escaped := value
-	if sql_dialect == .mysql {
-		escaped = escaped.replace('\\', '\\\\')
+// Doubling single quotes is standard SQL, and is understood by every dialect.
+//
+// A backslash has no single MySQL spelling: with the default sql_mode it escapes
+// the next character, so `C:\tmp` has to be written `C:\\tmp`, while under
+// NO_BACKSLASH_ESCAPES that same text is two literal backslashes. orm_table_gen()
+// generates DDL without a connection and cannot know the server's mode, so rather
+// than silently storing one or two backslashes depending on it, a backslash in a
+// MySQL string literal default is rejected.
+fn sql_string_literal(sql_dialect SQLDialect, value string) !string {
+	if sql_dialect == .mysql && value.contains('\\') {
+		return error('orm: a backtick delimited `default:` value cannot contain a backslash on MySQL, because its meaning depends on the NO_BACKSLASH_ESCAPES sql_mode of the server; use `sql_type` with an explicit DEFAULT clause instead')
 	}
-	return "'" + escaped.replace("'", "''") + "'"
+	return "'" + value.replace("'", "''") + "'"
 }
 
 pub fn orm_table_gen(sql_dialect SQLDialect, table Table, q string, defaults bool, def_unique_len int, fields []TableField, sql_from_v fn (int) !string,
@@ -1549,7 +1551,7 @@ pub fn orm_table_gen(sql_dialect SQLDialect, table Table, q string, defaults boo
 		stmt = '${q}${field_name}${q} ${col_typ}'
 		if defaults && has_default {
 			if default_is_literal {
-				stmt += ' DEFAULT ${sql_string_literal(sql_dialect, default_val)}'
+				stmt += ' DEFAULT ${sql_string_literal(sql_dialect, default_val)!}'
 			} else if default_val != '' {
 				stmt += ' DEFAULT ${default_val}'
 			} else {
--- a/vlib/orm/orm.v
+++ b/vlib/orm/orm.v
@@ -1417,6 +1417,7 @@ pub fn orm_table_gen(sql_dialect SQLDialect, table Table, q string, defaults boo
 		}
 		mut default_val := field.default_val
 		mut has_default := default_val != ''
+		mut default_is_literal := false
 		mut nullable := field.nullable
 		mut is_unique := false
 		mut is_skip := false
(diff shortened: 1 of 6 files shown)


vlib/orm/orm.v
```diff
@@ -1524,7 +1534,9 @@ pub fn orm_table_gen(sql_dialect SQLDialect, table Table, q string, defaults boo
 		}
 		stmt = '${q}${field_name}${q} ${col_typ}'
 		if defaults && has_default {
-			if default_val != '' {
+			if default_is_literal {
+				stmt += " DEFAULT '${default_val.replace("'", "''")}'"
```
medvednikov: This escaping is SQL-standard for quotes, but `orm_table_gen` is also used with `.mysql`, where backslashes are escape characters by default. A literal such as a Windows path containing `\t`, `\n`, `\0`, etc. can therefore be stored with a different value even though this path is documented as a plain string literal. The current runtime regression is SQLite-only, so it won't catch this. Please make literal rendering dialect-aware (or otherwise handle/document MySQL's backslash semantics) and add a MySQL-oriented rendering regression.

medvednikov (commented):
The documented backtick form is the right distinction from verbatim SQL, and the SQLite/comptime-path coverage is useful. I found one cross-dialect escaping issue in the literal renderer.

vlib/orm/orm.v
medvednikov: Good catch — fixed in b769e17434.

Literal rendering is now dialect-aware, via `sql_string_literal(sql_dialect, value)`:

```v
fn sql_string_literal(sql_dialect SQLDialect, value string) string {
	mut escaped := value
	if sql_dialect == .mysql {
		escaped = escaped.replace('\\', '\\\\')
	}
	return "'" + escaped.replace("'", "''") + "'"
}
```

Single quotes are doubled everywhere (standard SQL); backslashes are doubled for `.mysql` only, so `@[default: '`C:\tmp\new`']` keeps its backslashes there instead of becoming a tab and a newline. `.pg`, `.sqlite`, `.h2` and `.default` are unchanged — Postgres has `standard_conforming_strings` on by default, and the others treat backslash literally.

Added `vlib/orm/orm_default_literal_dialect_test.v`, a pure rendering test (no server needed), asserting both directions for `C:\tmp\new` and `o'brien\x`:

- `.mysql` → `DEFAULT 'C:\\tmp\\new'`, `DEFAULT 'o''brien\\x'`
- `.default` / `.pg` / `.sqlite` / `.h2` → `DEFAULT 'C:\tmp\new'`, `DEFAULT 'o''brien\x'`

It fails against the previous quote-only escaping and passes now, under `-old-compiler` and under a strict (no-fallback) V3 run. All 42 `vlib/orm` tests and the 8 `vlib/db/sqlite` tests still pass. The README now documents the MySQL backslash behaviour too.

vlib/orm/orm.v
```diff
@@ -1362,6 +1362,20 @@ fn parse_table_attr_fields(table Table, attr VAttribute, valid_sql_field_names [
...
+// sql_string_literal renders `value` as a quoted SQL string literal.
+// Doubling single quotes is standard SQL. MySQL additionally treats backslash as
+// an escape character inside string literals, unless the server runs with
+// NO_BACKSLASH_ESCAPES, so a literal backslash has to be doubled there too -
+// otherwise a value like `C:\tmp\new` would be stored with a tab and a newline
+// in it.
+fn sql_string_literal(sql_dialect SQLDialect, value string) string {
+	mut escaped := value
+	if sql_dialect == .mysql {
+		escaped = escaped.replace('\\', '\\\\')
```
medvednikov: This fixes MySQL's default backslash-escape mode, but it reverses the bug when `NO_BACKSLASH_ESCAPES` is enabled: in that mode `\\` is two literal backslashes, so a source value containing one backslash is stored with two. The comment notes that mode, while the README now says the value keeps its backslashes on every dialect. Since `orm_table_gen` has no SQL-mode input, please either use a MySQL representation that is mode-independent for backslashes, or explicitly document/reject the `NO_BACKSLASH_ESCAPES` limitation rather than promising round-trip behavior there. A regression should cover the chosen contract.

medvednikov (commented):
The default-mode MySQL corruption is fixed, but there is one remaining SQL-mode edge case in the new escaping path.

vlib/orm/orm.v
medvednikov: You're right, and the README was over-promising. Fixed in 896b0a5e56 by changing the contract rather than the escaping.

There is no single-quoted spelling of a backslash that is correct in both MySQL modes, and `orm_table_gen()` builds DDL without a connection, so it cannot know which mode the server runs in. Rather than pick a mode and silently corrupt the other, a backslash in a backtick-delimited `default:` is now **rejected on MySQL only**:

```v
fn sql_string_literal(sql_dialect SQLDialect, value string) !string {
	if sql_dialect == .mysql && value.contains('\\') {
		return error('orm: a backtick delimited `default:` value cannot contain a backslash on MySQL, because its meaning depends on the NO_BACKSLASH_ESCAPES sql_mode of the server; use `sql_type` with an explicit DEFAULT clause instead')
	}
	return "'" + value.replace("'", "''") + "'"
}
```

The contract is now:

| | quotes | backslash |
|---|---|---|
| `.default` / `.pg` / `.sqlite` / `.h2` | doubled | verbatim |
| `.mysql` | doubled | error, pointing at `sql_type` |

I considered a mode-independent MySQL representation — a hex literal (`DEFAULT X'...'`) is the only one that avoids backslash processing in both modes — but it makes the DDL unreadable, is charset-sensitive for non-binary columns, and I have no MySQL server here to confirm it is accepted as a column default, so I did not want to ship it untested.

`orm_default_literal_dialect_test.v` now covers the whole contract in three tests: quotes doubled across all five dialects, backslashes verbatim on the four standard ones, and the MySQL rejection (asserting the error mentions both `backslash` and `NO_BACKSLASH_ESCAPES`). The README no longer claims round-trip on every dialect and documents the `sql_type` workaround. All 42 `vlib/orm` tests pass, under `-old-compiler` and under a strict (no-fallback) V3 run.

PR #28362, opened 2026-09-05
cgen, v3: fix fixed array sumtype variants on both backends (fix #28127)

### LINKED ISSUE #28127: cgen: fixed-array variant of a sumtype emits invalid C (`initializing 'f64 *' with 'double'`)
**V version:** `V 0.5.2` (master)

### Description

A fixed-array variant of a sumtype (`[3]f64` in `type Val = f64 | [3]f64 | []Val`) generates invalid C when the value flows through a `map`/`match`: `initializing 'f64 *' … with an expression of incompatible type 'double'`.

### Reproduction

```v
module main

type Val = f64 | [3]f64 | []Val

fn main() {
	mut m := map[string]Val{}
	m['p'] = [0.0, 0.0, 0.0]!
	v := Val([1.0, 2.0, 3.0]!)
	match v {
		[3]f64 { println('vec ${v}') }
		[]Val { println('list') }
		else {}
	}
	_ = m
}
```

### Actual

```
error: initializing 'f64 *' (aka 'double *') with an expression of incompatible type 'double'
```

### Expected

Compiles; the `[3]f64` fixed-array variant is stored/read from the sumtype correctly.

_Reported via bugs.vlang.io_

> [!NOTE]
>

medvednikov:
Fixes #28127, on both backends.

### V1 (`vlib/v/gen/c`)

```v
type Val = f64 | [3]f64 | []Val

fn main() {
	mut m := map[string]Val{}
	m['p'] = [0.0, 0.0, 0.0]!
}
```

```
error: initializing 'f64 *' (aka 'double *') with an expression of incompatible type 'double'
```

`assign_stmt()` routed this through the fixed-array assignment path, which `memcpy()`s the raw fixed-array bytes over the destination using the *destination's* C type. With a sumtype on the left that produced a temporary declared as the sumtype, and a `memcpy()` spliced into a half-written `map_set()` call:

```c
main__Val _t1 = {0.0, 0.0, 0.0};
memcpy(builtin__map_set(&m, &(string[]){_S("p")}, &(main__Val[]) {, _t1, sizeof(main__Val));
```

That path is only correct when the left side is a fixed array too. When it is a sumtype or an interface, the value has to be boxed into the variant (`Array_fixed_f64_3_to_sumtype_main__Val(...)`) by the regular assignment path — which is what `v := Val([1.0, 2.0, 3.0]!)` already did correctly. The new `is_boxed_fixed_array_val` flag disables both the fixed-array assignment branch and `is_fixed_array_var` for exactly that case.

### V3 (`vlib/v3`)

The same issue family reproduces on V3 — the issue's own reproducer fails there too, under a strict run (`V_MACOS_V3_NO_FALLBACK=1`, which disables the silent `-old-compiler` retry that normally hides it). Two defects:

**Parser.** `match_branch_cond()` only recognised `[]T` as an array type pattern, so a fixed array variant was parsed as an array *literal*:

```v
match v {
	[3]f64 { println('vec ${v}') }   // parsed as the value `[3]f64`
	f64 { println('num') }           // then: `f64` evaluated but not used
}
```

That turned the whole match into a value match, and lowered the branch into a sum type equality call handed a raw `double[3]`. A `!` right after the closing `]` still marks a fixed array *literal* (`[1, 2]!`), so those value patterns keep parsing as values — covered by its own test.

**Transform.** A comparison whose operand is smartcast through an index or selector kept the operand's *declared* sum type, so

```v
if a[0] is [3]f64 {
	assert a[0] == [1.0, 1.5, 2.0]!
}
```

fell past the fixed-array comparison in `transform_infix_array_ops()` and reached the sum type's equality helper, which C could not compile. It now prefers the checker's narrowed type, via a new `TypeChecker.smartcast_type_name()`.

### Covered shapes

`vlib/v/tests/sumtypes/sumtype_fixed_array_variant_assign_test.v` covers a fixed array literal *and* a fixed array variable assigned to:

- a map value of sumtype type (the reported case)
- a plain sumtype variable
- an array element of sumtype type
- a struct field of sumtype type

plus `match` branches on a fixed array variant, `match` branches on fixed array literal *values*, and a guard that ordinary fixed-array assignment (including into a `map[string][3]f64`) still uses the `memcpy()` path. It fails on `master` under `-old-compiler` and under a strict V3 run, and passes here on both.

### Tests

`vlib/v/tests/` — 2212 passed, 4 failed, 7 skipped. The 4 failures are all in `vlib/v/tests/vls/` and reproduce identically on `master` without this change.

For V3, the repro programs and the regression test were checked against a compiler built from clean `origin/master` (all fail there) and against this branch (all pass), plus a targeted `vlib/v3/tests` subset — its only failure, `fixed_array_const_materialization_codegen_test.v`, fails identically on `master`. The full `vlib/v3/test_all.vsh` runs here via the V3 CI workflow.

### DIFF
--- a/vlib/v/gen/c/assign.v
+++ b/vlib/v/gen/c/assign.v
@@ -1388,7 +1388,13 @@ fn (mut g Gen) assign_stmt(node_ ast.AssignStmt) {
 		right_sym := g.table.sym(unwrapped_val_type)
 		unaliased_right_sym := g.table.final_sym(unwrapped_val_type)
 		unaliased_left_sym := g.table.final_sym(g.unwrap_generic(var_type))
-		is_fixed_array_var := unaliased_right_sym.kind == .array_fixed && val !is ast.ArrayInit && (val in [
+		// A fixed array assigned to a sumtype/interface has to be boxed into the variant
+		// (`Array_fixed_f64_3_to_sumtype_Val(...)`), not memcpy()-ed over the receiving
+		// sumtype/interface struct.
+		is_boxed_fixed_array_val := unaliased_right_sym.kind == .array_fixed
+			&& unaliased_left_sym.kind in [.sum_type, .interface] && !var_type.has_flag(.option)
+		is_fixed_array_var := unaliased_right_sym.kind == .array_fixed && !is_boxed_fixed_array_val
+			&& val !is ast.ArrayInit && (val in [
 			ast.Ident,
 			ast.IndexExpr,
 			ast.CallExpr,
@@ -1493,7 +1499,8 @@ fn (mut g Gen) assign_stmt(node_ ast.AssignStmt) {
 				}
 				g.writeln(';}')
 			}
-		} else if node.op == .assign && (is_fixed_array_init || is_fixed_array_var || (unaliased_right_sym.kind == .array_fixed && val is ast.CastExpr)) {
+		} else if node.op == .assign && !is_boxed_fixed_array_val
+			&& (is_fixed_array_init || is_fixed_array_var || (unaliased_right_sym.kind == .array_fixed && val is ast.CastExpr)) {
 			// Fixed arrays
 			if unaliased_left_sym.kind != .array_fixed && unaliased_right_sym.kind == .array_fixed && (g.pref.translated || g.file.is_translated) {
 				// translated:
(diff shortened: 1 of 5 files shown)


medvednikov (commented):
Reviewed both backend fixes and the regression coverage. The V1 change correctly keeps fixed-array values out of the raw `memcpy` assignment path when the destination needs sumtype/interface boxing. On V3, treating `[N]T` as a match type pattern and then honoring the checker's narrowed type for fixed-array comparisons addresses the two distinct failures without changing ordinary fixed-array assignments or `[a, b]!` value patterns. The added tests cover the important destination shapes and both match forms. I don't see a blocking issue in this PR.

PR #28363, opened 2026-09-05
cgen: always terminate `#line` directives with a newline (fix #28163)

### LINKED ISSUE #27495: cgen, -g: `#line` directive emitted with no newline before an if/match-expr array-literal branch (`expected expression`)

### LINKED ISSUE #28163: cgen, -g: #line directive glued to following line for an if/match-expression branch valued by `or {}` (expected expression)
### Description

Same `#line`-emission family as #27495 (closed as completed for the array-literal-branch case), but for a *different* triggering shape that isn't covered by that fix: an `if`/`match`-**expression** branch whose value comes from an `or {}` block, rather than an array literal.

`#line` gets emitted glued directly to the following expression with no newline, so the C preprocessor swallows the expression as trailing tokens of the `#line` directive — the assignment/return then has no value and the C compiler fails with `expected expression before '}' token` (or `before 'goto'`, when the `or {}` block's un-taken branch lowers to a jump).

### Where it was hit

vlang/v#28129 (an HTTP/3 client PR), which is the first PR to make `net.http` transitively import `net.quic` — so this is the first time `vlib/net/quic/conn.v`/`flow_control.v` get compiled as part of any downstream-app-building CI job (`gitly-windows`, `v-apps-compile`, `v-compiles-os-android`, `vab-compiles-v-examples`). The source itself is unrelated to that PR's own diff — both affected functions predate it by several months (`flow_control.v`, merged via #27882; `conn.v`, merged via #28083) — this is a **pre-existing, latent compiler bug newly exposed**, not something introduced by that PR.

### Reproducer (both real instances, confirmed via CI annotations)

`vlib/net/quic/conn.v` (`process_initial_or_handshake`):
```v
keys := if space == .initial {
    c.initial_keys_server
} else {
    c.handshake_keys_server or { return }
}
```

`vlib/net/quic/flow_control.v` (`initial_send_limit_for_stream`/`initial_receive_limit_for_stream`):
```v
return if locally_initiated {
    peer_params.initial_max_stream_data_bidi_remote or { 0 }
} else {
    peer_params.initial_max_stream_data_bidi_local or { 0 }
}
```

### Actual CI errors (verbatim, via GitHub check-run annotations on commit `78df3a7c69`)

```
vlib/net/quic/conn.v:710:63: warning: extra tokens at end of #line directive
vlib/net/quic/conn.v:707:16: error: expected expression before '}' token
        c.handshake_keys_server or { return }

vlib/net/quic/flow_control.v:171:71: warning: extra tokens at end of #line directive
vlib/net/quic/flow_control.v:168:17: error: expected expression before 'goto'
        peer_params.initial_max_stream_data_bidi_remote or { 0 }

vlib/net/quic/flow_control.v:170:16: error: expected expression before '}' token
        peer_params.initial_max_stream_data_bidi_local or { 0 }

vlib/net/quic/flow_control.v:189:71: warning: extra tokens at end of #line directive
vlib/net/quic/flow_control.v:186:17: error: expected expression before 'goto'
        own_params.initial_max_stream_data_bidi_local or { 0 }

vlib/net/quic/flow_control.v:188:16: error: expected expression before '}' token
        own_params.initial_max_stream_data_bidi_remote or { 0 }
```

CI runs where this is visible (GitHub check-run annotations, `vlang/v` PR #28129, commit `78df3a7c69`):
- https://github.com/vlang/v/actions/runs/32634022138/job/97189947946 (gitly-windows)
- https://github.com/vlang/v/actions/runs/32634022149/job/97189947919 (v-apps-compile ubuntu-latest — full gcc output in the log)
- https://github.com/vlang/v/actions/runs/32634022149/job/97189946608 (v-compiles-os-android)
- https://github.com/vlang/v/actions/runs/32634022149/job/97189946470 (vab-compiles-v-examples)

Note the `flow_control.v` failures happen with `or { 0 }` (a plain value, not a diverging `return`) — so this isn't specific to divergent `or {}` blocks either; any `or {}` used as an if/match-expression branch value seems to trigger it, matching #27495's own broader "same `#line`-emission family" framing.

### Notes

- I could not reproduce this locally with a plain `v -g -o out.c -b c` on macOS/clang (0 matches for `#line [0-9]+ "[^"]*"[^ ]`, both with a `make`-bootstrapped `./v` and a pre-existing `/usr/local/bin/v`) — the CI failures are on Linux/gcc and Windows. #27495 notes it reproduced on both clang/macOS and gcc/Linux for the array-literal shape, so this may be narrower (compiler-version- or platform-specific) for the `or {}` shape specifically.
- Workaround applied in vlang/v#28129 for both sites: restructure the if/match-expression-with-`or{}`-branch into a plain `mut` variable + if/else *statement* (each branch a plain assignment) instead of an if-expression — this compiles cleanly and preserves the original lazy/short-circuit evaluation of each branch.

> [!NOTE]
>

medvednikov:
Fixes #28163.

### Problem

With `-g`, cgen routinely cuts an unfinished statement out of the output with `go_before_last_stmt()` / `go_before_ternary()`, trims it, and writes it back after generating the nested code. When that chunk ends with a `#line N "file"` directive, whatever is written next lands on the same line, and the C preprocessor swallows it as trailing tokens of the directive:

```c
_t2 =
#line 10 "vlib/net/quic/flow_control.v" (*(u64*)_t4->data);
```

```
warning: extra tokens at end of #line directive [-Wextra-tokens]
error: expected expression before '}' token
error: expected expression before 'goto'
```

The shape reported in #28163 is an `or {}` block used as the value of an `if`/`match` **expression** branch, which broke `vlib/net/quic` in every downstream-app-building CI job.

### Fix

#27495 patched two individual write-back sites (array-literal branches). Rather than adding a third — cgen has ~100 `go_before_last_stmt()` write-back sites, any of which can end with a directive — this closes the invariant centrally: an unterminated `#line` directive is tracked as it is written, and terminated before the next write.

Concretely, `write()` / `write_decimal()` / `writeln()` first flush a pending directive, and `write()` records whether the chunk it just wrote ends with an open one. `go_back()`, `go_back_to()`, `go_before_last_stmt()`, `go_before_ternary()`, `insert_before_stmt()` and `insert_at()` clear the flag, since they rewind the output.

### No effect when `-g` is off

`vlines_pending_nl` can only ever be set while `pref.is_vlines` is true, and `terminate_open_line_directive()` is a single already-false bool check otherwise. `write2()` / `writeln2()` keep their fused fast path and only delegate to `write()` / `writeln()` in vlines mode.

Verified empirically: `v -o out.c cmd/tools/vdoc` produces byte-identical C before and after (7.7 MB, only the embedded compiler path differs). With `-g`, the only difference is the newline being added — `master` emits 690 glued `#line` directives for that same program, this branch emits 0.

### Tests

`vlib/v/tests/or_block_in_if_expr_branch_with_g_test.v` compiles both reduced forms from the issue with `-g` and asserts that no `#line` directive in the generated C has anything following the file name. It fails on `master` (`found #line directive glued to an expression: #line 8 "…" (*(u64*)_t4->data);`) and passes here. The existing `match_expr_array_literal_with_g_test.v` from #27495 keeps passing.

`vlib/v/tests/` — 2212 passed, 4 failed, 7 skipped; the 4 failures are all in `vlib/v/tests/vls/` and reproduce identically on `master`. `v -g -o v cmd/v` also builds and runs.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -115,6 +115,7 @@ mut:
 	is_sql                               bool // Inside `sql db{}` statement, generating sql instead of C (e.g. `and` instead of `&&` etc)
 	is_shared                            bool // for initialization of hidden mutex in `[rw]shared` literals
 	is_vlines_enabled                    bool // is it safe to generate #line directives when -g is passed
+	vlines_pending_nl                    bool // the output ends with an unterminated `#line` directive
 	is_autofree                          bool // false, inside the bodies of fns marked with [manualfree], otherwise === g.pref.autofree
 	is_autofree_tmp                      bool // when generating autofree temporary variables
 	is_builtin_mod                       bool
--- a/vlib/v/gen/c/text_manipulation.v
+++ b/vlib/v/gen/c/text_manipulation.v
@@ -7,6 +7,9 @@ import v.util
 
 const trace_gen_wanted_value = $d('trace_gen_wanted_value', '')
 
+// line_directive is the start of a `#line N "file"` C preprocessor directive.
+const line_directive = '#line '
+
 @[if trace_gen_wanted ?]
 fn (mut g Gen) trace_gen_wanted_context(last_character_len int, s string) {
 	last_n := g.out.last_n(last_character_len)
@@ -38,18 +41,69 @@ fn (mut g Gen) trace_gen(reason string, s string) {
 	}
 }
 
+// ends_with_open_line_directive reports whether the last line of `s` is a `#line`
+// directive that was not terminated by a newline.
+// With `-g`, cgen routinely cuts an unfinished statement out of the output with
+// go_before_last_stmt()/go_before_ternary(), trims it, and writes it back later.
+// When that chunk ends with a `#line N "file"` directive, whatever is written
+// next lands on the same line, and the C preprocessor swallows it as trailing
+// tokens of the directive - `warning: extra tokens at end of #line directive`,
+// followed by `error: expected expression` for the statement that just lost its
+// value. Tracking it here terminates such a directive before the next write,
+// instead of patching every one of the ~100 write back sites.
+@[direct_array_access]
+fn (g &Gen) ends_with_open_line_directive(s string) bool {
+	if !g.pref.is_vlines || s.len == 0 || s[s.len - 1] == `\n` {
+		return false
+	}
+	mut i := s.len - 1
+	for i > 0 && s[i - 1] != `\n` {
+		i--
+	}
+	for i < s.len && s[i] in [` `, `\t`, `\r`] {
+		i++
+	}
+	if s.len - i < line_directive.len {
+		return false
+	}
+	for j in 0 .. line_directive.len {
+		if s[i + j] != line_directive[j] {
+			return false
+		}
+	}
+	return true
+}
+
+// terminate_open_line_directive closes a pending `#line` directive, so that the
+// next generated code starts on its own line.
+@[inline]
+fn (mut g Gen) terminate_open_line_directive() {
+	if g.vlines_pending_nl {
+		g.vlines_pending_nl = false
+		g.out.writeln('')
+		g.empty_line = true
+	}
+}
+
 @[expand_simple_interpolation]
 fn (mut g Gen) write(s string) {
 	g.trace_gen_wanted(s)
 	g.trace_gen('write', s)
+	g.terminate_open_line_directive()
 	if g.indent > 0 && g.empty_line {
 		g.out.write_string(util.tabs(g.indent))
 	}
 	g.out.write_string(s)
 	g.empty_line = false
+	g.vlines_pending_nl = g.ends_with_open_line_directive(s)
 }
 
 fn (mut g Gen) write2(s1 string, s2 string) {
+	if g.pref.is_vlines {
+		g.write(s1)
+		g.write(s2)
+		return
+	}
 	g.trace_gen_wanted2(s1, s2)
 	g.trace_gen('write2 s1', s1)
 	if g.indent > 0 && g.empty_line {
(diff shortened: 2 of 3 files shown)


medvednikov (commented):
Reviewed the centralized `#line` termination invariant and the regression. Moving this into the write helpers is preferable to adding more call-site-specific fixes: it covers extracted/rewritten statement fragments while keeping the non-`-g` fast paths intact. The rewind helpers also clear the pending state before rebuilding output. I don't see a blocking correctness issue in this change.

PR #28369, opened 2026-09-05
checker: reject narrow hard registers next to named intel asm operands (fix #28364)

### LINKED ISSUE #28364: intel asm{} blocks: %V-substituted named operands break when combined with explicit sized registers (regression from #28358)
### What did you do?

Compiled the canonical `intel` example from #28298 (the issue this PR closed), unmodified:

```v
fn intel_sub(a int, b int) int {
	mut res := 0
	asm amd64 intel {
		mov eax, a
		sub eax, b
		mov res, eax
		; +r (res)
		; r (a)
		  r (b)
	}
	return res
}

fn main() {
	println(intel_sub(50, 8))
}
```

Run with either `-cc gcc` or `-cc clang`.

### What did you expect to see?

`42`, compiling and running cleanly. This is the same "load into eax, operate, store from eax"
idiom already used for the pre-existing AT&T `asm amd64 { }` example directly above the new
`intel` section in `doc/docs.md` (`mov eax, a` / `mov c, eax`), just with `intel` added.

### What did you see instead?

A C compilation error, under both compilers:

```
error: invalid operand for instruction
<inline asm>:2:2: note: instantiated into assembly here
        mov eax, rcx
             ^
```

(GNU as, invoked via `-cc gcc`, reports the equivalent `operand type mismatch for 'mov'`.)

### Root cause

`intel` blocks substitute a named operand via the `%V[name]` GNU/LLVM operand modifier
(`vlib/v/gen/c/cgen.v`, added in #28358). `%V` always prints the operand's **full native-width**
register name (e.g. `rcx` on amd64), regardless of the V variable's actual declared type - here
`a`/`b`/`res` are all plain `int` (32-bit), but `%V[a]` renders as `rcx`, not `ecx`.

The moment a named operand shares an instruction with an explicitly-sized hard register -
`mov eax, a` above, or any similarly natural intel-syntax code - the two operands have mismatched
widths and the assembler rejects the instruction outright. This isn't a corner case: it's the
natural way anyone familiar with x86 assembly would write an intel-syntax structured block, and
it's the exact idiom the existing AT&T example already demonstrates. The PR's own `intel` doc
example (`add value, increment`) happens to avoid the issue only because *both* operands are
named aliases, so both get the same (64-bit) `%V` treatment consistently - which masks the bug
rather than avoiding it, since the operation is then silently happening at 64-bit width, not the
variables' actual 32-bit `int` width.

Confirmed reproducible on latest master (`ac7b9980ac`) under both `-cc gcc` and `-cc clang`, on
Windows/amd64.

> [!NOTE]
>

medvednikov:
Fixes #28364.

### Problem

The canonical `intel` example fails to build under both `-cc gcc` and `-cc clang`:

```v ignore
asm amd64 intel {
	mov eax, a
	sub eax, b
	mov res, eax
	; +r (res)
	; r (a)
	  r (b)
}
```

```
error: invalid operand for instruction
<inline asm>:2:2: note: instantiated into assembly here
        mov eax, rcx
```

### Why it cannot be fixed by picking a better modifier

`.intel_syntax noprefix` requires register names without `%`, and `%V` is the **only** GNU x86 operand modifier that omits the prefix — but it always prints the 64-bit register regardless of the operand's type. I probed every documented modifier under `clang --target=x86_64` with a 32-bit `int` operand:

| modifier | emitted |
|---|---|
| `%k` | `%ecx` |
| `%w` | `%cx` |
| `%b` | `%cl` |
| `%q` | `%rcx` |
| `%V` | `rcx` |

Every width-aware modifier keeps the `%`. `.intel_syntax prefix` is not an escape hatch either — clang's integrated assembler rejects it outright (*"registers must not have a '%' prefix in .intel_syntax"*).

So within a structured `intel` block a named operand is necessarily 64-bit, and mixing it with a narrower hard register cannot be made to work.

### Fix

Reject it in the checker, so the user gets a diagnostic against their own source instead of against generated assembly:

```
error: a named operand in a structured `intel` block is always substituted as a 64-bit
register, so it cannot share an instruction with the 32-bit register `eax`; use the
64-bit register instead, or a `raw intel` block with explicit operand modifiers
   4 |         mov eax, a
     |         ~~~~~~~~~~
```

The check is skipped when the operand constraints were already rejected, so an unsupported `m` constraint doesn't also emit a width error (this keeps `asm_intel_memory_constraint.out` unchanged). The docs now state that named operands are always 64-bit, and point at `raw intel` for width control.

The 64-bit form is accepted and produces the C I verified assembles cleanly under `clang --target=x86_64`:

```c
".intel_syntax noprefix\n\t"
"mov rax, %V[a]\n\t"
"sub rax, %V[b]\n\t"
"mov %V[res], rax\n\t"
".att_syntax prefix\n\t"
```

### Tests

`vlib/v/checker/tests/asm_intel_operand_width_err.{vv,out}`. The full `vlib/v/compiler_errors_test.v` suite passes, and the existing `intel` slow test (`add intel_result, rhs`, no hard registers) is unaffected.


medvednikov (commented):
I found a correctness issue with the width check as implemented. `%V` is target-native width, not universally 64-bit: on i386 it names the 32-bit register, so this checker would reject valid combinations such as `mov eax, a`. More importantly, on amd64 this only diagnoses a narrow *hard* register; a named `int` operand is still rendered through `%V` as a 64-bit register even when there is no hard register in the instruction. That can silently change semantics for width-sensitive operations (for example signed `cmp`/branches or `sar` on a 32-bit negative value), rather than merely producing an assembler error. I think the restriction needs to be based on the target/native width and the named operand's actual V type (or structured Intel operands need another width-aware mechanism), with i386 and a width-sensitive amd64 regression. As written, the new diagnostic closes the reported `mov eax, a` symptom but leaves silent width mismatches accepted.

medvednikov (commented):
There is another correctness problem beyond the existing native-width concern: this check rejects *any* structured Intel instruction that contains both a named operand and a hard register narrower than 64 bits, but valid x86 instructions can intentionally mix widths. For example, on amd64 `mov cl, 1` followed by `shl value, cl` (with `value` as a named `+r` operand) emits a valid `shl r64, cl`; `movzx out, al` is another valid mixed-width form. `check_asm_intel_operand_widths` rejects both simply because it sees `cl`/`al` in the same instruction. So compatibility cannot be inferred from "named operand + any narrow hard register" alone; the check needs to be operand/instruction-specific (and should also address the existing i386/native-width issue) rather than blanket-rejecting mixed-width instructions.

medvednikov:
Addressed both width-review concerns at `bb8104d476`:

- validate each named structured-Intel operand against the target-native register width using its checked V type, including aliases inside addressing expressions
- keep valid mixed-width hard-register instructions such as `shl r64, cl` and `movzx r64, al` accepted
- accept 32-bit named operands for i386, while diagnosing named-only 32-bit operations such as `sar int, 1` on amd64

Validation: V1 compiler rebuild; compiler errors (1700 passed, 1 skipped); checker suite (4/4); amd64 Intel/Clang runtime regressions under an x86_64 macOS target; i386 cross-check; formatting and diff checks. The broader inout run reached 82 passing cases plus 9 expected panics and 1 skip, with 3 unrelated stale-branch V3 assertion-output mismatches.

medvednikov:
@codex review

medvednikov:
Addressed both follow-up findings at `ba884a3619`. Integer-literal width classification now lives on the AST literal and is shared by C emission and Intel-assembly operand validation, so literal inputs follow the exact C type actually emitted. The docs now describe `%V` as target-native (64-bit on amd64, 32-bit on i386), and the regression coverage exercises both amd64 rejection and i386 acceptance.\n\nValidation: compiler rebuilt; compiler-errors suite passed (1700, 1 skip); checker suite passed (4/4); amd64 Clang and i386 cross-compile assembly checks passed; C-output suite passed (98/106 outputs, 8 platform skips; 133/142 must-have fixtures, 9 platform skips); `check-md` passed with zero errors; formatting and diff checks passed.

medvednikov:
@codex review

medvednikov:
Addressed the cross-target `%V` finding at `1dd0e8037e`. When the compilation target is x86, structured Intel operands are now checked against that actual target rather than the architecture label on the assembly block; this catches an `asm i386` block compiled for amd64 while preserving true i386 builds. The documentation now makes that distinction explicit.\n\nValidation: compiler rebuilt; the i386-target check passes; the same i386-labelled block is correctly rejected for an amd64 target; compiler-errors passed (1700, 1 skip); checker passed (4/4); `check-md` completed with zero errors; formatting and diff checks passed.

medvednikov:
@codex review

medvednikov:
Addressed the `-m32` machine-width finding. `%V` validation now derives native register width from `pref.m64`, matching the generated C compiler mode even when an explicit `-arch` value is retained. The diagnostic and docs now describe the current compilation target rather than assuming the architecture enum determines width.\n\nValidation: compiler rebuilt; `-arch amd64 -m32` accepts the 32-bit regression; 64-bit target rejection remains covered; compiler-errors passed (1700, 1 skip); checker passed (4/4); `check-md` completed with zero errors; formatting and diff checks passed.

medvednikov:
@codex review

medvednikov:
Addressed both follow-up findings at `8a6178a642`. Generic operand types are unwrapped before width validation and unresolved placeholders are deferred to the concrete generic recheck. Same-width Intel integer instructions now also validate explicit hard-register widths against native-width named operands, while intentional mixed-width forms such as `shl r64, cl` and `movzx r64, al` remain accepted.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); amd64 Intel/Clang generic and mixed-width runtime regressions passed; `-arch amd64 -m32` native-width cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed the latest two findings at `042841e2e3`. Hard-register validation now normalizes instruction mnemonics before matching, so uppercase forms such as `MOV eax, value` cannot bypass the check. Valid segment-selector source forms remain accepted only in the applicable `mov native_alias, segment_register` operand position.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); uppercase negative fixture passed; amd64 Intel/Clang generic, mixed-width, and segment-register runtime regressions passed; `-arch amd64 -m32` cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed the remaining segment-register finding at `ba3e09f79c`. The same-width check now permits segment-register MOVs in both assembler-accepted directions while still rejecting CS as a destination. The Clang regression performs a safe DS read/write round trip through a native-width named operand.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); amd64 Intel/Clang runtime coverage passed, including the DS round trip, generic operands, and valid mixed-width forms; `-arch amd64 -m32` cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed the bit-test family finding at `717c4a0744`. `bt`, `btc`, `btr`, and `bts` now receive the same hard-register compatibility validation as the other same-width integer operations. Mnemonic normalization also strips the parser-preserved `lock ` prefix before family matching. Added an amd64 negative regression for `bt eax, i64_named`.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); amd64 Intel/Clang runtime coverage passed; `-arch amd64 -m32` cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed both latest findings at `71c505b485`. `shld`/`shrd` now validate their same-width data operands while deliberately allowing the immediate/8-bit CL count position. On 32-bit targets, CR/DR MOV forms use the target-native 32-bit width despite the shared register table metadata. Regressions cover rejected `shld eax, i64_named`, accepted `shld i64_named, i64_named, cl`, and accepted i386 control-register MOVs.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); amd64 Intel/Clang runtime coverage passed; `-arch amd64 -m32` control-register cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed the BMI-family finding at `130e386bf0`. The hard-register compatibility table now covers the complete BMI1/BMI2 same-width families, along with related ADX and MOVBE operations, while retaining operand-position exceptions for instructions that intentionally mix widths. Added an amd64 negative regression for `andn eax, i64_left, i64_right`.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); amd64 Intel/Clang runtime coverage passed; `-arch amd64 -m32` cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed the enum-storage finding at `9332165440`. Intel operand width checks now follow aliases and use an explicitly sized enum’s declared backing type, matching C emission, while preserving pointer and option/result sizing. Regressions cover accepting `enum IntelWide as u64` in amd64 structured Intel assembly and diagnosing `enum IntelNarrow as u8` as 8-bit.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); amd64 Intel/Clang runtime coverage passed, including the wide enum; `-arch amd64 -m32` cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed the encoding-prefix finding at `b7aaec5d15`. Mnemonic normalization now peels the parser-supported `rex`, `vex`, and `xop` prefixes (including dotted modifiers) and `lock` in sequence before applying instruction-family width rules. Added an exact amd64 negative regression for `rex add eax, i64_named`.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); amd64 Intel/Clang runtime coverage passed; `-arch amd64 -m32` cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed the decimal INT_MIN finding at `c3b8bf542b`. Integer-literal concrete typing now classifies the unsigned magnitude using C candidate-type boundaries and treats a leading minus as a unary operator. Thus decimal `-2147483648` is native 64-bit on amd64, while non-decimal literals retain the applicable unsigned 32-bit candidate. Added a Clang runtime regression for the exact boundary.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); amd64 Intel/Clang runtime coverage passed; `-arch amd64 -m32` cross-check passed; touched V sources pass `fmt -verify`; `git diff --check` passed. C-output must-have fixtures passed (133, 9 platform skips); the output pass had one transient unrelated `if_define_check_not_set` miss, and its focused rerun passed.

medvednikov:
@codex review

medvednikov:
Addressed the size-suffix finding at `f52a0a1c87`. Same-width instruction matching now recognizes the parser-supported `b`/`w`/`l`/`q` suffixes only when removing the suffix yields a known instruction, so forms such as `addq eax, i64_named` receive the same width validation without misclassifying unsuffixed mnemonics. Added the exact amd64 negative regression.\n\nValidation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); amd64 Intel/Clang runtime coverage passed; `-arch amd64 -m32` cross-check passed; touched standalone V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed the MOVQ vector-transfer finding at `111f0545a4`. Size-suffixed scalar validation now recognizes native-width `movq` transfers involving MMX/XMM registers as intentional mixed-container-width forms, while retaining the width check for scalar GPR forms. Added a Clang runtime round-trip through `xmm0`.\n\nValidation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); explicit x86_64 Clang Intel-assembly regression passed with `-gc none`; `-arch amd64 -m32` cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed. The initial host-default Clang invocation targeted ARM64 and rejected the existing x86 clobbers; the architecture-correct rerun passed.

medvednikov:
@codex review

medvednikov:
Addressed the negative non-decimal literal regression at `d4a4effb00`. Negative integer magnitudes that require more than 31 bits now remain signed through C emission when they fit `i64`, preventing an unsigned suffix from turning `i64(-0xffffffff)` into `1`. Positive non-decimal candidates retain their C-compatible unsigned classification. Added an exact C-output runtime fixture expecting `-4294967295`.\n\nValidation: compiler rebuilt; exact runtime regression passed; C-output suite passed (99/107 runnable output fixtures with 8 platform skips; 133/142 must-have fixtures with 9 platform skips); compiler-errors passed (1700, 1 skip); checker passed (4/4); x86_64 Clang Intel-assembly suite passed with `-gc none`; `-arch amd64 -m32` cross-check passed; touched V files pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed the alias-only overreach at `dc538cad2c`. Structured Intel width validation now runs only when a same-width instruction combines a named register operand with an explicitly conflicting hard register; all-named narrow V operands remain accepted. Restored the original alias-list plumbing and C integer-literal emitter, removing the now-obsolete AST/type-width machinery and its incidental Cgen changes. Updated the docs and expected diagnostics accordingly.\n\nValidation: compiler rebuilt; existing all-named `int` Intel regression passed under x86_64 Clang; explicit hard-register conflict fixture passed; compiler-errors passed (1700, 1 skip); checker passed (4/4); x86_64 Clang Intel-assembly suite passed with `-gc none`; `-arch amd64 -m32` cross-check passed; `i64(-0xffffffff)` again evaluates to `-4294967295` through the restored emitter; markdown check completed with 302 OK examples, 1 pre-existing warning, and 0 errors; formatting and diff checks passed.

medvednikov:
@codex review

medvednikov:
Addressed both latest findings at `fbdd2c5e31`. Recognized `b`/`w`/`l`/`q` suffixes now retain their explicit operand width, so a non-native suffix applied to a `%V` named operand is diagnosed even when the hard register itself is native-width. CRC32 now validates a hard destination against a named native-width source while preserving its valid mixed-width form with a named destination and narrow explicit source. Regressions cover rejected `addl rax, i64_named`, rejected `crc32 eax, i64_named`, and compiling `crc32 native_alias, al`.\n\nValidation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); x86_64 Clang all-named/Intel assembly regressions passed with `-gc none`; `-arch amd64 -m32` cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed both operand-role findings at `27c744540e`. Address expressions are now checked independently: when a base/index is a named `%V` alias, an explicit register in that same address must use the same target address width, without conflating address registers with top-level data operands. CRC32 now permits narrower explicit sources next to a native named destination only for the architecture-valid combinations: 8/16-bit sources with a 32-bit destination, and only the byte form with a 64-bit destination. Regressions cover rejected `[native_alias + eax + 0]`, accepted `[native_alias + rax + 0]`, rejected `crc32 native_alias, eax`, and accepted byte-source CRC32.\n\nValidation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); x86_64 Clang all-named/Intel assembly regressions passed with `-gc none`; `-arch amd64 -m32` cross-check passed; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Addressed both exact-head findings and the adjacent audited cases at `fe3cf6ef66`. Address alias discovery and hard-register checks now include `AsmAddressing.displacement`, covering both `[eax * 4 + native_alias]` and `[native_alias * 4 + eax]`. CRC32 on 32-bit targets now exempts only legal 8-/16-bit narrow sources, while a 64-bit hard source is rejected; persisted i386 coverage accepts the byte/word forms. SHLD/SHRD now exempts only `cl` as a register count, and CMOV condition-plus-size forms such as `cmovzl` retain their explicit suffix width without misreading `cmovl` (condition less).\n\nValidation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker passed (4/4); x86_64 Clang all-named/Intel assembly regressions passed with `-gc none`; i386 byte/word CRC forms pass under `-arch amd64 -m32`; an isolated m32 `crc32 named, rax` check produces the intended 64-vs-32 diagnostic; touched V fixtures pass `fmt -verify`; `git diff --check` passed.

medvednikov:
@codex review

medvednikov:
Fixed the label-shadowing finding in d96c53a7ce: width validation now mirrors Cgen label precedence by excluding local/global labels that collide with named I/O aliases. Added checker regressions for both local and global collisions. Validation passed: compiler rebuild; 1,700 compiler-error tests (1 skipped); checker suite (4/4); Clang amd64 Intel assembly; i386 native-width check; formatting and diff checks.

medvednikov:
@codex review

medvednikov:
Addressed both exact-head findings at `34c9a8fc75`.

- structured Intel shift/rotate destinations now retain operand types and reject narrow destinations whose semantics would otherwise change under `%V` native-register substitution (including the `sar int` signed case)
- existing valid all-named narrow `add` behavior remains accepted
- width checks now exclude aliases shadowed by file-wide global asm labels, matching C generation precedence
- added exact diagnostics and file-global-label collision regressions, and documented the width-sensitive exception

Validation completed: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker suite passed (4/4); focused regression passed; amd64 Clang cross-build passed; existing all-alias `add` cross-build passed; `-m32` i386 width cross-build passed; `check-md` passed (302 examples, 0 warnings/errors/format errors); diff check passed. The broader `vlib/v/` suite is still running in parallel.

medvednikov:
@codex review

medvednikov:
Addressed the RIP-relative follow-up at `d13d4d9c41`. Structured Intel validation now rejects a named operand specifically in the displacement of `[rip + ...]`, preventing GCC rejection and Clang silently dropping RIP. Added the exact regression.

Validation: compiler rebuilt; compiler-errors passed (1700, 1 skip); checker suite passed (4/4); amd64 Clang cross-build passed; fixture formatting and diff checks passed.

medvednikov:
@codex review

medvednikov:
Addressed the two exact-head Codex findings in `23ffc9e40d`:

- preserve vector indexes for recognized VSIB gather/scatter instructions while retaining scalar address-width validation elsewhere
- reject narrow signed aliases used by `cmp`, whose `%V` substitution changes signed comparison semantics on amd64
- add focused error coverage plus a successful Clang codegen regression for the VSIB form

Validation:
- `./vnew -silent vlib/v/compiler_errors_test.v` (1700 passed, 1 skipped)
- `./vnew -silent test vlib/v/checker/` (4 passed)
- focused `asm_intel_operand_width_err` regression (passed)
- Clang amd64 cross-build of `asm_intel_clang_test.amd64.v` (passed)
- `./vnew check-md doc/docs.md` (302 OK, 0 warnings/errors/fmt errors)
- `git diff --check` (passed)

medvednikov:
@codex review

medvednikov:
Addressed the exact-head Codex findings from review 5136679274 in `d8d62a0a3b`:

- reject narrow signed aliases for `test` as well as `cmp`, preventing incorrect sign flags after `%V` zero-extension
- reject named count operands for legacy shifts/rotates and `shld`/`shrd`, directing callers to `cl` or explicit raw modifiers
- add exact diagnostic regressions and document both restrictions

Validation:
- rebuilt `./vnew`
- `./vnew -silent vlib/v/compiler_errors_test.v` (1700 passed, 1 skipped)
- `./vnew -silent test vlib/v/checker/` (4 passed)
- focused `asm_intel_operand_width_err` regression (passed)
- Clang amd64 cross-build of `asm_intel_clang_test.amd64.v` (passed)
- `./vnew check-md doc/docs.md` (302 OK, 0 warnings/errors/fmt errors)
- `git diff --check` (passed)

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
Addressed all three exact-head findings from review 5136801745 in `1b8d71652b` and audited adjacent operand roles:

- reject named `movsx`/`movsxd`/`movzx` sources while retaining valid byte/word hard sources and 32-bit `movsxd` sources
- validate CRC32 memory suffixes against the native named destination, while retaining valid byte/native memory forms
- reject narrow aliases in width-dependent data roles for legacy/double/BMI shifts and rotates, leading/trailing-zero counts, byte swaps, bit tests, CRC32, and MULX
- reject named immediate-only `rorx` counts in addition to legacy `cl`-only shift counts
- add focused diagnostics, positive Clang codegen coverage, and documentation

Validation:
- rebuilt `./vnew`
- `./vnew -silent vlib/v/compiler_errors_test.v` (1700 passed, 1 skipped)
- `./vnew -silent test vlib/v/checker/` (4 passed)
- focused `asm_intel_operand_width_err` regression (passed)
- Clang amd64 cross-build of `asm_intel_clang_test.amd64.v` (passed)
- `./vnew check-md doc/docs.md` (302 OK, 0 warnings/errors/fmt errors)
- `git diff --check` (passed)

medvednikov:
@codex review

medvednikov:
Addressed the exact-head review findings in `2ce92c5d31`:

- reject narrow named operands for `div`, `idiv`, `mul`, and one-operand `imul`
- reject ambiguous addressed sources for structured `movsx`/`movzx`, while retaining valid `movsxd` memory sources
- recognize signed enum backing types in narrow `cmp`/`test` validation
- add focused error regressions, a positive Clang MOVSXD memory test, and documentation

Validation:
- rebuilt `./vnew`
- exact operand-width fixture: 1/1 passed
- compiler errors: 1700 passed, 1 skipped
- checker: 4/4 passed
- amd64 Clang cross-build and execution: passed
- `check-md doc/docs.md`: 302 OK, 0 warnings/errors/fmt errors
- broad `vlib/v/` suite reproduced only the same two unrelated baseline failures in `anon_fn_map_assign_test.v` and `assign_option_of_fixed_array_test.v`, then was stopped
- `git diff --check`: passed

medvednikov:
@codex review

medvednikov:
Addressed the exact `2ce92c5d31` review in `fc3d45475a`:

- reject unsized addressed operands for `div`, `idiv`, `mul`, and one-operand `imul`, while allowing explicit-width forms
- reject register-valued third address components while preserving valid two-register forms
- reject signed narrow address aliases that would be substituted without sign extension
- track later flag consumers and reject narrow arithmetic whose native-width flags would be observed; preserve narrow low-result arithmetic when flags are unused
- broaden flag-sensitive coverage to related arithmetic/BMI forms
- add focused error regressions and positive Clang tests for explicit-width memory arithmetic, unused narrow-add flags, unsigned narrow indexes, and MOVSXD memory

Validation:
- rebuilt `./vnew`
- exact operand-width fixture: 1/1 passed
- compiler errors: 1700 passed, 1 skipped
- checker: 4/4 passed
- amd64 Clang cross-build and execution: passed
- `check-md doc/docs.md`: 302 OK, 0 warnings/errors/fmt errors
- `git diff --check`: passed

The broader `vlib/v/` run on the preceding exact patch reproduced the same two unrelated baseline failures (`anon_fn_map_assign_test.v` and `assign_option_of_fixed_array_test.v`) and was stopped after confirming them.

medvednikov:
@codex review

medvednikov:
Addressed the exact `fc3d45475a` review in `9c5c470197`:

- treat an unconditional jump after narrow arithmetic conservatively, so lexically unreachable flag overwrites cannot hide a reachable later flag consumer
- add the jump/overwrite/label/`seto` regression
- remove GNU-only mnemonic suffixes from the Clang-forced test and move those cases to Linux/GCC-only coverage
- replace the Clang test asm negative literal with parser-valid instructions

Validation:
- rebuilt `./vnew`
- exact operand-width fixture: 1/1 passed
- compiler errors: 1700 passed, 1 skipped
- checker: 4/4 passed
- actual amd64 Clang test harness (copied without the host-architecture filename skip): 1/1 compiled, ran, and passed
- Linux/GCC suffix fixture formatted and checker/C-generation validated locally; genuine GNU GCC is unavailable on this macOS host
- prior `check-md doc/docs.md`: 302 OK, 0 warnings/errors/fmt errors (docs unchanged in this commit)
- `git diff --check`: passed

medvednikov:
@codex review

medvednikov:
Addressed the exact-head review findings in 9fc4c64086:

- track individual status flags so INC/DEC overwrite OF/ZF/SF/PF/AF while preserving CF
- recognize GNU movzb*/movzw*/movsb*/movsw*/movslq extension forms and validate explicit source/destination widths
- add negative source/destination and CF-preservation regressions plus positive Clang/GCC coverage

Validation: compiler errors (1700 passed, 1 skipped), checker tests (4/4), inout compiler tests (95 fixtures, 0 errors), copied amd64 Clang harness (1/1), GNU suffix fixture checker pass, and git diff --check. The broad vlib/v run was attempted but stopped after unrelated existing anonymous-function/fixed-array failures.

medvednikov:
@codex review

medvednikov:
Addressed the exact 9fc4c64086 review findings in 980e3a7fcc:

- classify CMC as both a CF reader and writer so carry dependencies are retained
- recognize statically nonzero shift/rotate counts as overwriting their affected status flags, including count-one OF
- add CMC carry-flow rejection plus positive SHL/ROL overwrite regressions

Validation: rebuild; compiler errors (1700 passed, 1 skipped); checker tests (4/4); inout compiler tests (95 fixtures, 0 errors); copied amd64 Clang harness (1/1); git diff --check.

medvednikov:
@codex review

medvednikov:
Addressed the exact 980e3a7fcc review finding in 7e9f6a2ac2:

- treat all j*, ljmp, and loop* instructions as conservative control-flow boundaries during structured Intel flag-liveness scanning
- add the exact narrow ADD / JRCXZ / lexical XOR / SETC regression

Validation: rebuild; compiler errors (1700 passed, 1 skipped); checker tests (4/4); inout compiler tests (95 fixtures, 0 errors); copied amd64 Clang harness (1/1); git diff --check.

medvednikov:
@codex review

medvednikov:
Addressed the exact 7e9f6a2ac2 review finding in 514053ae64:

- reject unsuffixed CRC32 addressed sources because their byte/qword source width is ambiguous
- retain the existing explicitly suffixed CRC32 memory-source validation
- add the exact `crc32 result, [base]` diagnostic regression

Validation: rebuild; compiler errors (1700 passed, 1 skipped); checker tests (4/4); inout compiler tests (95 fixtures, 0 errors); copied amd64 Clang harness (1/1); git diff --check.

medvednikov:
@codex review

medvednikov:
Addressed the exact 514053ae64 review findings in 898a910bbe:

- reject signed narrow named sources when a same-width instruction consumes them as a wider destination-width value without sign extension
- mask static shift/rotate counts according to the actual operand width (6 bits for 64-bit operands, 5 bits otherwise) before deciding whether flags are overwritten
- add MOV/ADD signed-source errors, a count-33 positive regression, and a count-64 flag-preservation regression

Validation: rebuild; compiler errors (1700 passed, 1 skipped); checker tests (4/4); inout compiler tests (95 fixtures, 0 errors); copied amd64 Clang harness (1/1); git diff --check.

medvednikov:
@codex review

PR #28371, opened 2026-09-05
i18n: read translations from .json files as well (fix #27766)

### LINKED ISSUE #27766: [feat] Support JSON as an alternative format for i18n translation files
### Proposal

I would like to propose adding optional support for **JSON format** as an alternative to the current `.tr` files for internationalization (i18n), while keeping `.tr` as the default and primary format.

### Motivation

The `.tr` format is simple and works well for the compiler. However, as projects grow and more developers with different backgrounds join, the lack of tooling and flexibility around `.tr` files becomes a pain point. Allowing JSON as an **opt-in alternative** could:

1. **Improve developer experience (DX)** – JSON is a universal, well-known format with built-in syntax highlighting, linting, auto-completion, and formatting support in every modern code editor. This reduces friction for new contributors.

2. **Better modularization** – JSON allows structuring translations by module/feature (e.g., `zh/common.json`, `zh/dashboard.json`), which is easier to manage in large codebases. Currently, `.tr` files are typically monolithic per language.

3. **Lower barrier to entry** – Many developers are already familiar with JSON-based i18n workflows (e.g., i18next, Vue I18n). Adopting JSON would make V language more approachable to web developers and newcomers.

4. **Leverage existing translation tools** – The rich ecosystem of JSON translation platforms (e.g., POEditor, Lokalise, Crowdin) can directly generate or consume JSON files, streamlining the localization workflow.

### Why this doesn't conflict with V's design philosophy

- **Compile-time embedding still works** – JSON files can be embedded at compile time (via `include` or similar mechanism), just like `.tr` files. Performance and single-binary distribution are preserved.
- **`.tr` remains the default** – JSON support would be optional and opt-in. The core compilation pipeline, which prioritizes speed and simplicity, would not be affected if the project doesn't use JSON.
- **Parser overhead is minimal** – A lightweight JSON parser (or using a well-established C/C++ library like `cJSON`) can be added to the compiler. Since this would only be activated when JSON files are present, the default build speed for `.tr`-only projects stays unchanged.

### Implementation sketch (for discussion)

A possible implementation approach:

1.  **Detect file extensions** in the `i18n/` directory – support both `.tr` and `.json`.
2.  **Parse JSON files** into the same internal key-value representation as `.tr` files.
3.  **Merge translations** from both formats (with conflict resolution, e.g., `.tr` takes precedence).
4.  **Embed merged data** at compile time exactly as is done today.

### Open questions

- Should we support **nested keys** in JSON (`{"common": {"hello": "Hello"}}`)? If so, how to map them to the flat key lookup in `tr()`?
- How to handle **comments**? (maybe a `.jsonc` variant, or ignore unknown fields?)
- Should we define a **schema** or convention for the JSON structure?

### Alternatives considered

- **Third-party tool** – A pre-build tool could convert JSON to `.tr` format. While this is workable, it adds an extra step to the build process and doesn't provide native IDE integration for the compiled `.tr` files.

### Conclusion

Adding optional JSON support would significantly improve the i18n experience in V without compromising its core values of performance and simplicity. It would make V more accessible to a wider audience while keeping the existing `.tr` workflow untouched.

I'm happy to help with an initial implementation if the community is open to this direction. Looking forward to your thoughts!

---

**Labels:** `proposal`, `i18n`, `enhancement`

---

medvednikov:
Fixes #27766.

`.tr` stays the primary format and is completely unchanged. `.json` files in the same translations directory are now loaded as well, so translations can come from tooling that already speaks JSON (POEditor, Crowdin, i18next-style workflows).

### The issue's open questions, answered

**Nested keys** — flattened with `.`, so the flat lookup in `tr()` is untouched:

```json
{ "msg_hello": "Hello", "menu": { "file": "File", "edit": { "undo": "Undo" } } }
```

defines `msg_hello`, `menu.file` and `menu.edit.undo`.

**Structure / modularization** — `translations/en.json` names its language exactly like `en.tr` does. A file inside a *language subdirectory* is namespaced by its own file name, which gives the per-feature split the issue asked for:

```
translations/zh/dashboard.json   ->  lang `zh`, keys `dashboard.title`, `dashboard.widgets.clock`
```

**Conflicts** — a `.tr` entry wins when both formats define the same key, as proposed.

Two smaller decisions: non-string scalars are stored via their JSON representation (`"answer": 42` → `"42"`), and arrays are skipped rather than given invented index keys. A malformed file now reports its path and is skipped, instead of discarding every other translation.

### Tests

`vlib/i18n/i18n_test.v` gains three tests over new testdata — a `<lang>.json` beside the existing `.tr` files, nested flattening, scalar values, the `.tr`-wins rule, and the `zh/dashboard.json` namespaced layout. The pre-existing tests still pass unchanged, under `-old-compiler` and under a strict (no-fallback) V3 run.

Also adds `vlib/i18n/README.md`, since the module had no docs at all.


medvednikov (commented):
The JSON loading/flattening design looks coherent, and loading JSON before `.tr` makes the documented `.tr` precedence straightforward. One scalar edge needs tightening: `json2.Any` includes `json2.Null`, but `flatten_tr_json` falls through to `value.str()` for it. `Null` has JSON encoding support, but generic V stringification is not the JSON literal `null`, so a JSON translation like `{"missing": null}` will not follow the stated “non-string scalars use their JSON representation” rule. Please special-case null (either store `null` or explicitly skip it) and add a regression. Otherwise I don’t see a blocking issue in the main format/namespace handling.

PR #28372, opened 2026-09-05
db.mysql: add a `local_infile` Config option (fix #27525)

### LINKED ISSUE #27525: db.mysql: Add local_infile option to Config for LOAD DATA LOCAL INFILE support.
### Describe the feature

Summary
The db.mysql module currently does not support LOAD DATA LOCAL INFILE, which is the fastest way to bulk-load data into MySQL (typically 10-20x faster than multi-value INSERT statements). This is because MYSQL_OPT_LOCAL_INFILE must be set via mysql_options() before mysql_real_connect(), but DB.conn is a private field, making it impossible for users to call C.mysql_options() from outside the module.

Current Problem
When bulk-loading a 570K-row, 61-column CSV file:

Multi-value INSERT (current approach): ~33 seconds
LOAD DATA LOCAL INFILE (if available): estimated ~2-3 seconds
The performance gap is significant for any data ingestion use case.

Proposed Solution
Add a local_infile field to mysql.Config:

```v
 
pub struct Config {
pub:
    host       string
    port       u32    = 3306
    user       string
    password   string
    dbname     string
    local_infile bool = false  // <-- new field
}```
In the connect() function, before calling C.mysql_real_connect():


```c
 
if config.local_infile {
    C.mysql_options(conn, C.MYSQL_OPT_LOCAL_INFILE, &true_val)
}```
This is a minimal, backward-compatible change — local_infile defaults to false, so existing code is unaffected.

### Use Case

Use Case Example

```v
 
import db.mysql

mut db := mysql.connect(
    host:        '127.0.0.1'
    port:        3306
    user:        'root'
    password:    ''
    dbname:      'mydb'
    local_infile: true   // <-- enable LOAD DATA LOCAL INFILE
)!```

// Now LOAD DATA LOCAL INFILE works
db.exec_none("LOAD DATA LOCAL INFILE '/path/to/data.csv' INTO TABLE mytable FIELDS TERMINATED BY ',' ENCLOSED BY '\"' LINES TERMINATED BY '\\n' IGNORE 1 LINES")
Alternative Considered
Using StmtHandle.execute() with prepared statements for batch inserts, but StmtHandle reallocates bind arrays on every call (see stmt.c.v:844), making it slower than multi-value INSERT for large batches.

### Proposed Solution

_No response_

### Other Information

_No response_

### Acknowledgements

### Version used

V lastest

V version: latest
MySQL: 8.0+
OS: Linux (WSL)
> [!NOTE]
>

### PR DESCRIPTION
**Pull request by medvednikov**
Fixes #27525.

### Problem

`LOAD DATA LOCAL INFILE` is the fastest way to bulk-load a table, but libmysqlclient 8.x disables it client-side unless `MYSQL_OPT_LOCAL_INFILE` is set **before** `mysql_real_connect()`. `DB.conn` is private, so this could not be done from outside the module — the statement was simply unavailable to `db.mysql` users.

### Change

`Config.local_infile`, following the existing `ssl_mode` pattern:

```v ignore
config := mysql.Config{
	host:         '127.0.0.1'
	username:     'root'
	dbname:       'mysql'
	local_infile: true
}
```

When set, `connect()` sets `MYSQL_OPT_LOCAL_INFILE` *and* adds the `.client_local_files` capability flag. The issue only asked for the option, but the statement needs both, and `client_local_files` was already in `ConnectionFlag` — requiring users to remember to set it separately would make the boolean a trap. It defaults to `false`, so existing code is unaffected. The server must still allow it (`local_infile=ON`), which the README now documents.

### Verification

I installed `mariadb-connector-c` locally to actually compile this rather than eyeball it. A program using `local_infile: true` builds, links and runs, and the emitted C contains the call in the right place:

```c
db__mysql__DB_set_option(&db, MYSQL_OPT_LOCAL_INFILE, &enabled);
```

`vlib/db/mysql/mysql_test.v` gains a server-free test asserting the option defaults to off, that `.client_local_files` is not set by default, and that it can be enabled. The file is `// vtest build: started_mysqld?`-guarded so it is skipped without a server; I verified it passes by running a copy with that guard removed.

Note: `v build-module vlib/db/mysql` fails on current master for an unrelated reason — the `_v_type_idx_orm__[]orm__Primitive___...` invalid C identifier that open PR #27891 fixes — so I verified via a normal build instead.

### DIFF
--- a/vlib/db/mysql/README.md
+++ b/vlib/db/mysql/README.md
@@ -75,6 +75,29 @@ Sharing one `mysql.DB` across threads now serializes connection-level queries sa
 For concurrent servers, prefer `mysql.new_connection_pool(...)` so requests do not share the same
 session and transaction state on one connection.
 
+## Bulk loading with `LOAD DATA LOCAL INFILE`
+
+libmysqlclient 8.x disables client side `LOAD DATA LOCAL INFILE` by default. Set
+`local_infile` so `connect()` enables it before the handshake - it also turns on the
+`.client_local_files` capability, which the statement needs:
+
+```v oksyntax
+import db.mysql
+
+config := mysql.Config{
+	host:         '127.0.0.1'
+	username:     'root'
+	password:     '12345678'
+	dbname:       'mysql'
+	local_infile: true
+}
+
+mut db := mysql.connect(config)!
+db.exec_none("load data local infile '/tmp/users.csv' into table users fields terminated by ','")
+```
+
+The server has to allow it as well (`local_infile=ON`).
+
 ## Transaction
 
 ```v oksyntax
--- a/vlib/db/mysql/local_infile_options_test.c.v
+++ b/vlib/db/mysql/local_infile_options_test.c.v
@@ -1,3 +1,4 @@
+// vtest build: started_mysqld?
 module mysql
 
 fn test_local_infile_connect_options() {
--- a/vlib/db/mysql/local_infile_options_test.c.v
+++ b/vlib/db/mysql/local_infile_options_test.c.v
@@ -1,0 +1,22 @@
+module mysql
+
+fn test_local_infile_connect_options() {
+	mut db := DB{
+		conn: C.mysql_init(0)
+	}
+	assert !isnil(db.conn)
+	defer {
+		C.mysql_close(db.conn)
+	}
+
+	flags := db.apply_local_infile(Config{
+		flag: .client_found_rows
+		local_infile: true
+	})
+	assert flags.has(.client_found_rows)
+	assert flags.has(.client_local_files)
+
+	mut enabled := u32(0)
+	assert C.mysql_get_option(db.conn, C.MYSQL_OPT_LOCAL_INFILE, &enabled) == 0
+	assert enabled == 1
+}
--- a/vlib/db/mysql/mysql.c.v
+++ b/vlib/db/mysql/mysql.c.v
@@ -96,6 +96,12 @@ pub mut:
 	// `MYSQL_OPT_SSL_MODE` call is made and the libmysqlclient default applies.
 	ssl_mode SslMode
 
+	// local_infile enables `LOAD DATA LOCAL INFILE`, which libmysqlclient 8.x
+	// disables by default. It sets `MYSQL_OPT_LOCAL_INFILE` before connecting and
+	// adds the `.client_local_files` capability flag, since the statement needs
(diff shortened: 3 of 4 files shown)


medvednikov (commented):
The production change is in the right place: `MYSQL_OPT_LOCAL_INFILE` is set before `mysql_real_connect()`, and deriving the effective connection flags from `config.flag` preserves the caller’s other capabilities. I don’t see a blocking code issue. The regression test does overstate what it verifies, though: it only checks the public `Config` fields and would still pass if the `set_option()` call and `connection_flag.set(.client_local_files)` were removed. Since those two lines are the feature, it would be better to make the effective-flag logic testable (or add a client-option/integration assertion where the MySQL CI is available) so this can’t silently regress.

medvednikov (commented):
Re-reviewed the current head. I still don't see a blocking production-code issue: the option is set before `mysql_real_connect()`, and the effective connection flags preserve the caller's existing flags while adding `client_local_files`. The existing test-coverage note remains the only concern: the new test only checks `Config` fields and would still pass if the `set_option()` call or effective-flag mutation regressed. No additional finding beyond that.

medvednikov:
Addressed the review coverage gap in `68ccf6e4ee`:

- extracted the exact pre-connect local-infile setup used by `connect()` so it can be exercised without a server
- added a same-module C-backend regression that verifies `MYSQL_OPT_LOCAL_INFILE` is actually enabled
- verifies `.client_local_files` is added while preserving caller-supplied flags
- retained the public configuration default/enablement check and corrected its scope comment

Validation:
- `./vnew -silent vlib/db/mysql/local_infile_options_test.c.v`
- `./vnew -silent test vlib/db/mysql/` (1 passed, 4 network-gated skips)
- `./vnew check-md vlib/db/mysql/README.md`
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed the Codex optional-dependency finding in `4db4c24045`:

- added the module-standard `// vtest build: started_mysqld?` guard to the C-backed regression
- dependency-free directory runs now skip the test before resolving MySQL headers
- the regression remains directly runnable where the client development package is available

Validation:
- `./vnew -silent vlib/db/mysql/local_infile_options_test.c.v` (passed)
- `./vnew -silent test vlib/db/mysql/` (all 5 correctly dependency-gated/skipped)
- formatting and `git diff --check`

medvednikov:
@codex review

PR #28376, opened 2026-09-05
fasthttp, v3: fix the two `-cstrict` build errors

medvednikov:
Unbreaks the `Test tools (-cstrict)` step, which is what fails `tools-linux (gcc)` (and the `tools-linux (clang)`/`(tcc)` siblings) on master. Two unrelated errors.

### 1. `vlib/fasthttp/request_parser.v` — pointer sign

```
error: pointer targets in passing argument 2 of 'fasthttp__ascii_ci_eq'
differ in signedness [-Werror=pointer-sign]
```

`c'chunked'` is a `char*`; `ascii_ci_eq(a &u8, b &u8, len int)` takes `&u8`. The other call site already passes a `string.str`, so only this one needed a cast.

Reproduced locally — clang gives the same error with `-cstrict`:

```
error: passing 'char[8]' to parameter of type 'u8 *' (aka 'unsigned char *') converts
between pointers to integer types where one is of the unique plain 'char' type and the
other is not [-Werror,-Wpointer-sign]
```

### 2. `vlib/v3/gen/c/cleanc.v` — duplicated branches

```
error: this condition has identical branches [-Werror=duplicated-branches]
return ((node.kind == array_literal)? (false) : (node.kind == array_init)? (true) : (true));
```

`is_safe_global_init` returned a match whose `.array_init` and `else` arms were both `true`. The whole match means "false iff `.array_literal`", so it now says that directly — 18 lines to 4, and the redundant branch is gone.

`-Wduplicated-branches` is GCC-only, so I could not reproduce it with clang here. I located it from the exact generated line in the CI log, then confirmed the shape disappears from the generated C: the `NodeKind__array_literal))? (false)` ternary goes from 1 occurrence to 0.

### Tests

- `vlib/fasthttp/` under `-cstrict`: 7 passed, 2 skipped (was a hard C error before)
- v3 const/global codegen tests pass. `fixed_array_global_for_in_codegen_test.v` fails, but it is one of the failures I previously confirmed also fails on a compiler built from clean `origin/master`.

### DIFF
--- a/vlib/fasthttp/request_parser.v
+++ b/vlib/fasthttp/request_parser.v
@@ -526,7 +526,10 @@ fn parse_transfer_encoding(buf []u8, value Slice, mut state TransferEncodingStat
 		if pos == token_start {
 			return false
 		}
-		is_chunked := pos - token_start == 7 && ascii_ci_eq(&buf[token_start], c'chunked', 7)
+		// `c'chunked'` is a `char*`; `ascii_ci_eq` takes `&u8`, and passing one for the
+		// other is an error under `-cstrict` (`-Werror=pointer-sign`).
+		is_chunked := pos - token_start == 7
+			&& ascii_ci_eq(&buf[token_start], &u8(c'chunked'), 7)
 		state.seen = true
 		state.final_chunked = is_chunked
 		state.chunked_seen = is_chunked
--- a/vlib/fasthttp/request_parser.v
+++ b/vlib/fasthttp/request_parser.v
@@ -526,10 +526,7 @@ fn parse_transfer_encoding(buf []u8, value Slice, mut state TransferEncodingStat
 		if pos == token_start {
 			return false
 		}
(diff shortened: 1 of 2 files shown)


medvednikov (commented):
Code review only; I intentionally ignored CI.

The two changes are mechanically sound, but this branch is stale enough that the `fasthttp` half has already landed on `master` as commit `11302b5019fe` (`fasthttp: cast the chunked token to &u8`). Please rebase and drop that hunk so this PR only carries the remaining V3 change; otherwise it is carrying an already-landed fix.

The `is_safe_global_init` simplification itself looks equivalent to the old match: it remains false only for `.array_literal`, with `.array_init` and every other kind returning true.

medvednikov:
Addressed the stale fasthttp delta in a focused follow-up commit (a5e3b3bf54) without rewriting history: the net PR diff now contains only the remaining V3 change. A synthetic merge against current `origin/master` via `git merge-tree --write-tree` succeeds and retains master's existing fasthttp cast while applying the V3 simplification.

@codex review

PR #28378, opened 2026-09-05
vself: make macOS and Linux self-builds V3-only

medvednikov:
## Summary

- remove the V1 compiler from `cmd/v` on macOS and Linux and route direct C builds through the embedded V3 compiler without a V1 fallback
- fix file-scoped import alias lowering so V3 can self-host `cmd/v`
- make macOS production self-builds use parallel C compilation, preallocation, and one production compile instead of the three-pass PGO flow
- emit parallel-C inserted definitions in exactly one translation unit and fix the Darwin libtcc `system` interposer
- preserve the compiler V root after moving a self-built binary and support fatal compiler errors through the V3 CLI
- stop active stage-memory monitors before preallocation cleanup during process exit

## Performance on macOS

The reported `v -prod self` regression was 9–11 minutes, with 48.3 MB / 667,012 lines of generated C and an approximately 5 GB linker footprint.

With this change, `/usr/bin/time -l ./vnew -prod self -o /tmp/v3-only-prod-self-final2` completed in **3:29**. The generated C was **22.4 MB / 411,652 lines**. The build performs one parallel production clang compile instead of three full PGO compiles.

## Validation

Passed:

- two consecutive V3 generations of `cmd/v`, followed by compiling hello world with the moved result
- `./vnew -prod -parallel-cc -o ./vnew cmd/v`
- `./vnew cmd/v/macos_v3_test.v`
- `./vnew cmd/tools/vself_test.v`
- `./vnew vlib/v/builder/parallel_cc_failure_test.v`
- `./vnew vlib/v3/bench/bench_test.v`
- V3 type cache, C type generation, driver environment, and Darwin Mach-O signing tests
- 20 repeated `-Wfatal-errors` compiler-error exits under preallocation, with no memory-monitor shutdown crash
- `./vnew check-md README.md`
- `./vnew check-md vlib/v3/README.md`

## Known V3 cutover gaps

This deliberately removes the V1 fallback, so existing V3 compatibility gaps become visible in broad V1-oriented suites:

- `vlib/v/compiler_errors_test.v`: 819 of 1700 cases reported output mismatches/failures before the run was stopped near completion
- `./vnew -silent test vlib/v/`: 19 of the first 29 tests failed before the run was stopped
- `vlib/v/gen/c/coutput_test.v`: an output mismatch was followed by an autofree panic
- `./vnew check-md doc/docs.md`: 286 examples passed, 16 had compile errors, and 4 had formatting errors

These are disclosed as remaining V3 compatibility work rather than being hidden by automatic fallback to V1.


medvednikov:
@codex review

Issue #26961, opened 2026-04-23
[Degradation] V compiler has become significantly slower


tankf33der:
Alarm.
```
lambda:~$ time v -o mike.c sha3.v 

real	0m0.430s
user	0m0.740s
sys     0m0.346s
lambda:~$ time ~/Downloads/v-0.5.1/v -o mike.c sha3.v 

real	0m0.278s
user	0m0.226s
sys     0m0.056s
```

`sha3.v` is tiny.
```
lambda:~$ wc -l sha3.v 
9 sha3.v
```

```v
import crypto.sha3
import time

fn main() {
	a := []u8{len: 10_000_000}
	t1 := time.now()
	_ := sha3.sum512(a)
	println(time.since(t1))
}
```

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

medvednikov:
yeah, will be brought back to old numbers

tankf33der:
Not helped.

medvednikov:
@tankf33der try again please.

tankf33der:
@medvednikov 

Numbers don't lie.
You can see the slowdown even with the naked eye. 
Am I really the only one who feels it this way?
```
lambda:~$ time v -o mike.c sha3.v 

real	0m0.449s
user	0m0.755s
sys     0m0.395s
lambda:~$ time ~/Downloads/v-0.5.1/v -o mike.c sha3.v 

real	0m0.216s
user	0m0.239s
sys     0m0.050s
```

Another way.
Compilation was always 500-600ms, now 2x slower on my cheapest laptop.
```
lambda:~/v$ v test vlib/crypto/
---- Testing... ----------------------------------------------------------------------------------------------------------------------------
 OK    [ 1/67] C:  1173.5 ms, R:     0.673 ms vlib/crypto/blake2b/blake2b_block_test.v
 OK    [ 2/67] C:  1211.8 ms, R:     6.273 ms vlib/crypto/aes/block_generic_test.v
 OK    [ 3/67] C:  1282.3 ms, R:    31.580 ms vlib/crypto/blake2s/blake2s_test.v
 OK    [ 4/67] C:  1355.4 ms, R:    35.350 ms vlib/crypto/blake2b/blake2b_test.v
[...]

PR #28380, opened 2026-09-05
vself: speed up Linux production builds

medvednikov:
## Summary

- make Linux production self-builds use the bounded single-build path already used on macOS
- add parallel C compilation and remove the monolithic LTO plus three-pass PGO cost
- disable the self-host memory guard for production generation, which can exceed 10 GiB
- honor parallel-cc supplied through VFLAGS when deciding whether to enable PGO
- document and test the Linux behavior

This follows PR #28378 and addresses the Linux production timing reported in #26961.

## Validation

- macOS: ./vnew -no-memory-limit cmd/tools/vself_test.v
- Debian x86_64: ./vtesthost -old-compiler -gc none cmd/tools/vself_test.v
- Debian x86_64: ./vtesthost -old-compiler fmt -verify cmd/tools/vself.v cmd/tools/vself_test.v
- git diff --check

README check-md reaches a pre-existing formatting error at README.md:388, outside this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,7 +301,8 @@ hello world
 ```
 
 `v self` defaults to `-gc none`. Pass `-gc <mode>` if you need a different GC mode.
-On macOS, `v -prod self` uses a single production build instead of the three-pass PGO cycle.
+On macOS and Linux, `v -prod self` uses a single production build instead of the three-pass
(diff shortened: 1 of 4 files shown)


medvednikov:
@codex review

PR #28386, opened 2026-09-05
macos: fix V3 module caching for UI builds

medvednikov:
## Summary

- add typed Objective-C message helpers for macOS UI code
- keep Objective-C declarations compatible with cached V3 module headers
- resolve return statements without emitting resolve_type warnings
- preserve tuple payload types while specializing option/result expressions
- add regression coverage for the macOS bridge and V3 type transformations

V3 module caching remains enabled by default. A cold UI build populates the cache, and the next build loads 18 .vh files while parsing only the application .v file.

## Testing

- ./vnew -silent test vlib/macos/macos_test.v
- ./vnew -silent test vlib/v3/types/resolve_type_test.v
- ./vnew -silent test vlib/v3/transform/fn_test.v
- VTEST_ONLY_FN=test_print_v_files_includes_warm_cached_module_sources ./vnew -silent test vlib/v3/tests/module_cache_test.v
- cold/warm V3 builds of ui2 examples/flight_booker
- cold/warm V3 builds of ui2 examples/dropdown

### DIFF
--- a/vlib/macos/README.md
+++ b/vlib/macos/README.md
@@ -1,0 +1,24 @@
+# macos
+
+The `macos` module provides typed access to the Objective-C runtime on macOS. It includes
+Objective-C object, selector, rectangle, point, and range types together with typed message
+senders.
+
+Choose the message sender whose return and argument types exactly match the Objective-C method.
+For example:
+
+```v
+import macos
+
+value := macos.msg_id_range(
+	macos.get_class('NSValue'),
+	'valueWithRange:',
+	macos.range(2, 5),
+)
+result := macos.msg_range(value, 'rangeValue')
+println('${result.location}, ${result.length}')
+```
+
+The typed senders hide the platform-specific `objc_msgSend` casts. Framework-specific methods
+still require the corresponding `#flag darwin -framework ...` declaration in the importing
+program.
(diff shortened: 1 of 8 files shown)


medvednikov:
@codex review

medvednikov:
Reviewed PR #28386 code-only, ignoring CI as requested. I found **no blocking correctness issues** and would approve it.

The Objective-C bridge changes look sound. The `const_` parameter renames are functional rather than cosmetic: V’s C generator emits `const` for non-mut pointer parameters whose names start with `const_`, matching the `const char *` / `const void *` signatures in `objc_bridge.h`.   The new `Point`, `Range`, and typed `objc_msgSend` casts also line up with their V-facing wrappers.

The V3 changes are coherent too. Tuple payloads are now recursively decomposed before specialization/qualification, consistent with the existing tuple handling in the normalization path.   Treating a `return_stmt` itself as `Void` also fits the existing generator contract: `if_expr_tail_has_no_value` explicitly uses `resolve_type(id) is Void` to distinguish control-flow tails from value-producing tails.

My only **non-blocking note** is test coverage around the Objective-C ABI helpers. This PR adds quite a few distinct `objc_msgSend` signatures, while the regression tests mainly exercise point/range plus a small subset of object/bool calls. A wrong cast can compile successfully and only fail at runtime, so representative coverage for one mixed scalar/object sender and one SEL-based sender would be useful. I wouldn’t hold the PR for that.

**Verdict: approve.**

PR #28387, opened 2026-09-05
v3: preserve parallel worker metadata

### LINKED ISSUE #28382: v3: compiler segfaults on json2.decode[T] (memmove crash, unwalkable stack)
### V version
`V 0.5.2 edf8242` (built from `edf824295b`, current `master` HEAD `98ee3c1c`)

### Environment
macOS (arm64), self-hosted V3-only build (`v` compiled after #28378)

### What did you do?
Compiled a minimal program that calls `json2.decode[T]` on a struct with a string field:

```v
import x.json2

struct Config {
	vlang_repo string
}

fn main() {
	content := '{"vlang_repo":"a"}'
	config := json2.decode[Config](content) or {
		println('err')
		return
	}
	println(config.vlang_repo)
}
```

```
v run min2.v
```

### What did you expect to see?
Either successful compilation and `a` printed, or a normal compiler error.

### What did you see instead?
The compiler itself segfaults (exit code 139) — this crashes during compilation,
before any program runs:

```
$ v run min2.v
Segmentation fault: 11
```

Confirmed with lldb: `EXC_BAD_ACCESS` inside `_platform_memmove`, with a stack that
cannot be unwound past frame 0 (consistent with heap/stack corruption rather than a
clean null-deref).

Reproduced identically with:
- The installed self-built `v` (`edf8242`)
- A fresh rebuild of `cmd/v` from current `master` HEAD (`98ee3c1c`)

So this is not a stale-binary artifact — it reproduces on current master.

### Additional context
This was originally hit via a real `.vsh` script (a small daily-scheduled tool) that
calls `json2.decode[Config](...)` to parse a JSON config file — i.e. this is an
ordinary, non-exotic use of `json2.decode`.

Separately: after #28378 ("vself: make macOS and Linux self-builds V3-only"), macOS/
Linux `v` builds no longer fall back to the established compiler on a V3 failure —
`-old-compiler` now reports `this V executable contains only the V3 compiler`. That
change is presumably intentional, but it means a V3 crash like this one now reaches
end users directly as a raw segfault instead of being silently masked by the old
fallback-and-auto-report path. Flagging in case that's relevant to how this should be
triaged.

> [!NOTE]
>

medvednikov:
## Summary

- copy parallel interface implementation indexes into the parent arena before worker storage is released
- deep-copy monomorph cache argument strings before monomorph worker arenas are released
- add cold-cache regressions for both the four-interface trigger and `json2.decode[T]`

The V3 preallocated build was publishing slices containing strings allocated in disposable worker arenas. Later cache preparation read those strings after the workers had shut down, causing the nondeterministic `memmove`/`wyhash` crashes reported on macOS arm64.

Fixes #28382

## Testing

- `./v -no-memory-limit -g -keepc -o ./vnew cmd/v`
- fresh-cache four-interface reproducer: prints `hi`
- fresh-cache issue JSON2 reproducer: prints `a`
- `VFLAGS='-no-memory-limit' VTEST_ONLY_FN='test_cold_module_cache_preserves_parallel_*' ./vnew -silent vlib/v3/tests/module_cache_test.v`
- `./vnew -silent vlib/v3/types/checker_parallel_test.v`
- `./vnew -silent vlib/v3/types/interface_impl_test.v`
- `./vnew -silent vlib/v3/transform/monomorphize_test.v`

The legacy `vlib/v/compiler_errors_test.v` corpus is not green under current V3-only master because of existing V3 diagnostic/output differences and three malformed-input hangs. `./vnew -silent test vlib/v/` also currently fails while building `vtest` because V3 cannot resolve its `testing` import. No expected outputs were changed.

### DIFF
--- a/vlib/v3/tests/module_cache_test.v
+++ b/vlib/v3/tests/module_cache_test.v
@@ -4485,6 +4485,84 @@ fn main() {
 	assert run_module_cache_binary(second_output) == '7'
 }
 
+fn test_cold_module_cache_preserves_parallel_interface_indexes() {
+	v3_bin := build_module_cache_v3()
+	root := os.join_path(os.temp_dir(), 'v3_module_cache_parallel_interfaces_${os.getpid()}')
+	os.rmdir_all(root) or {}
+	os.mkdir_all(root) or { panic(err) }
+	defer {
+		os.rmdir_all(root) or {}
+	}
+	write_module_cache_file(root, 'contract/contract.v', 'module contract
+
+pub interface JsonEncoder {
+	to_json() string
+}
+
+pub interface A {
+	a() int
+}
+
+pub interface B {
+	b() int
+}
+
+pub interface C {
+	c() int
+}
+
+pub struct Null {}
+
+pub fn (null Null) to_json() string {
+	_ = null
+	return "null"
+}
+')
+	main_file := os.join_path(root, 'main.v')
+	write_module_cache_file(root, 'main.v', 'module main
+
+import contract
+
+fn main() {
+	println("hi")
+}
+')
+	cache_dir := os.join_path(root, 'cache')
+	output := os.join_path(root, 'app')
+	compile_module_cache_project(v3_bin, cache_dir, main_file, output)
+	assert run_module_cache_binary(output) == 'hi'
+}
+
+fn test_cold_module_cache_preserves_parallel_monomorph_specs() {
+	v3_bin := build_module_cache_v3()
+	root := os.join_path(os.temp_dir(), 'v3_module_cache_parallel_monomorph_${os.getpid()}')
+	os.rmdir_all(root) or {}
+	os.mkdir_all(root) or { panic(err) }
+	defer {
+		os.rmdir_all(root) or {}
+	}
+	main_file := os.join_path(root, 'main.v')
+	write_module_cache_file(root, 'main.v', 'module main
+
+import x.json2
+
+struct Result {
+	success bool
+	values  []string
+}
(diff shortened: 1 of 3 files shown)


medvednikov:
@codex review

medvednikov:
Reviewed PR #28387, ignoring CI as requested.

**Verdict: approve — no blocking correctness findings.**

* The interface-index fix looks correct. The escaping `InterfaceImplIndex` contains only `names` and `ids`; the patch deep-copies every worker-owned name and then rebuilds the ID map from those owned strings. That preserves the documented deterministic ID assignment and avoids leaving map keys/string payloads in disposable worker arenas.
* The monomorph fix closes the specific shallow-copy hole. `record_monomorph_cache_spec` stores `args.clone()`, which copies the array but not the string payloads. `sorted_monomorph_cache_specs()` runs while worker arenas are still live and now clones each argument string before `release_monomorph_worker_scopes()`. The driver subsequently deep-copies the returned specs again before freeing the outer monomorph scope. The lifetime chain is sound.
* The two regressions target the reported failure modes directly: the four-interface structural-implementation case and a cold-cache `json2.decode[T]` specialization.  The original issue describes both the dangling-string-style `memmove`/`wyhash` crashes and the four-interface trigger.
* One **non-blocking cleanup**: the worker still computes `stable_interface_type_ids(impls)`, but that map is now discarded and recomputed on the parent. Removing the worker-side ID calculation would avoid redundant hashing/allocation.

I did **not** submit a review/approval on GitHub; this was the code review only.

PR #28392, opened 2026-09-05
cgen, net.mbedtls, vpm: fix the build with glibc 2.42 and gcc 15

medvednikov:
Reported by a user whose `v install --git https://github.com/vlang/ui2.git` failed with `cannot compile cmd/tools/vpm`, on an up to date Linux distro. Two unrelated C errors, both hit while building `cmd/tools/vpm`.

### 1. `_Generic` clash on `memchr` / `strchr` / `strrchr` / `strstr`

```
error: expected identifier or '(' before '_Generic'
  450 | V_CRT_LINKAGE void * V_CRT_CALL memchr(const void *str, int c, size_t n);
```

glibc 2.42 implements C23's type-generic `memchr`/`strchr`/`strrchr`/`strstr` as function-like macros over `_Generic`, so that a const-qualified argument yields a const-qualified return type, and gcc 15 defaults to `-std=gnu23`, so they are active. `cheaders.v` declares those four manually instead of including `<string.h>`, but by then an earlier include (`mbedtls/net_sockets.h`, `netdb.h`, `dirent.h`, ...) has usually pulled `<string.h>` in already, so the declaration expands into the middle of a `_Generic` expression.

Guarded the four declarations with `#ifndef`. A defined macro also means `<string.h>` has already declared the real function, so skipping V's declaration loses nothing. The reverse order keeps working as before: a `<string.h>` pulled in later by a module header defines the macro *after* these declarations, which C permits.

### 2. `conflicting types for 'v_mbedtls_pk_ec_group_id'`

```
error: conflicting types for 'v_mbedtls_pk_ec_group_id'; have 'int(mbedtls_pk_context *)'
note: previous definition with type 'mbedtls_ecp_group_id(const mbedtls_pk_context *)'
```

The helper was `mbedtls_ecp_group_id f(const mbedtls_pk_context *)` in C, but `fn C.v_mbedtls_pk_ec_group_id(pk &C.mbedtls_pk_context) int` in V. V emits its own prototype for that function into translation units that also include `mbedtls_helpers.h`, and a prototype differing in return type or in a parameter's const qualifier is a hard error, not a warning. V cannot spell `const T*` or a C enum return type in an `fn C.` declaration, so the C side is the one that has to match.

The other four `v_mbedtls_*` helpers were checked against their V declarations and already agree.

### Also

The two diagnostics `vpm` printed on the way there: an `abs` variable shadowing a function declaration (`parse.v`), and a deprecated implicit reference to voidptr in `set_shared_context` (`update.v`).

### Verification

Error 2 was reproduced directly and is clean after the fix. Error 1 needs glibc 2.42, so it was reproduced by injecting the equivalent `_Generic` macros into generated C: the old codegen fails with exactly the reported error, the new codegen compiles clean. `vlib/net/mbedtls` (9/9) and `vlib/net/quic/tls13_certificate_chain_test.v` (which covers the changed helper, both the P-256 and the non-EC case) pass, `cheaders_manual_stdlib_decls_test.v` passes, V self-builds, and `vpm` now builds with no notice or warning.

### DIFF
--- a/cmd/tools/vpm/parse.v
+++ b/cmd/tools/vpm/parse.v
@@ -265,9 +265,9 @@ fn is_local_repository(query string) bool {
 			// local repository. This keeps `v install vsl@<tag>` working when
 			// cwd happens to be the vmodules directory (the test setup for
 			// versioned installs does exactly this).
-			abs := os.real_path(path)
+			abs_path := os.real_path(path)
 			vmodules_real := os.real_path(settings.vmodules_path)
-			if abs.starts_with(vmodules_real + os.path_separator) || abs == vmodules_real {
+			if abs_path.starts_with(vmodules_real + os.path_separator) || abs_path == vmodules_real {
 				continue
 			}
 			return true
--- a/cmd/tools/vpm/update.v
+++ b/cmd/tools/vpm/update.v
@@ -20,7 +20,7 @@ fn vpm_update(query []string) {
 	idents := if query.len == 0 { get_installed_modules() } else { query.clone() }
 	mut pp := pool.new_pool_processor(callback: update_module)
 	ctx := UpdateSession{idents}
-	pp.set_shared_context(ctx)
+	pp.set_shared_context(&ctx)
 	pp.work_on_items(idents)
 	mut errors := 0
 	for res in pp.get_results[UpdateResult]() {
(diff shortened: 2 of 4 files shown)


medvednikov:
@codex review

PR #28394, opened 2026-09-05
v3: bring fast arm64 self-hosting to FastC parity

medvednikov:
## Summary

- make the V3 direct ARM64 backend self-host `vlib/v3/v3.v` without the macOS C fallback
- add the FastC parity fixes required by native V3 descendants
- optimize direct parsing, SSA construction, ARM64 emission, Mach-O linking, and signing
- reuse FastC source resolution, declaration, and signature collection in the typed ARM64 path

## Performance

Measured using optimized V compiler binaries on macOS arm64:

| backend | median | best | included work |
| --- | ---: | ---: | --- |
| FastC + TinyCC | 1.972 MLOC/s | 2.046 MLOC/s | C generation, TinyCC compilation, linking |
| Fast + ARM64 | 1.958 MLOC/s | 2.012 MLOC/s | typed parsing, ARM64 generation, Mach-O linking and signing |

Fast + ARM64 is within about 0.7% of FastC + TinyCC and is 2.73x faster than its previous 0.718 MLOC/s baseline.

## Validation

With `V_MACOS_V3_NO_FALLBACK=1`:

```text
optimized C-hosted V3
  -> native ARM64 child
  -> native ARM64 grandchild
  -> native ARM64 great-grandchild
```

Every generated compiler was an executable arm64 Mach-O and each native generation successfully compiled `vlib/v3/v3.v` again.

Broad test suites were intentionally skipped for this performance-focused iteration. The fallback-disabled multi-generation self-host chain was used as the requested validation.

## Parser direction

The two backends now share source resolution, scanning, declaration indexing, and signature collection. The remaining split is body lowering: ARM64 needs concrete type IDs while FastC currently leaves many expression types to C `typeof`. A future unification can have the shared parser emit typed operations which both the C renderer and ARM64 SSA builder consume; removing generated `typeof` expressions should also reduce TinyCC work.

### DIFF
--- a/vlib/v3/README.md
+++ b/vlib/v3/README.md
@@ -148,6 +148,11 @@ times in-process so an external sampler can profile it. The compiler's own C pre
 take effect one generation later: measure the compiler built by the modified compiler, not the
 modified compiler itself.
 
+`FASTC_BENCH=1` also prints end-to-end MLOC/s. The FastC total includes TinyCC compilation and
+linking; the direct ARM64 total includes Mach-O generation, linking, and signing.
+For the optimized FastC pipeline, build the full compiler with `./v -prod -o ./vnew cmd/v`, then run
+`FASTC_BENCH=1 ./vnew -nocache -b fastc -selfhost -o /tmp/v3-fastc vlib/v3/v3.v`.
+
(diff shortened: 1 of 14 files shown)


medvednikov:
@codex review

PR #28395, opened 2026-09-05
v3: reduce self-host time and peak memory

medvednikov:
Uncached V3 C-backend self-hosting spends substantial time allocating temporary lookup data and copying AST storage for parallel transforms. This reduces the measured non-CC time to **516–529 ms** and peak memory to **1.67–1.69 GB**.

- Use copy-on-write AST snapshots on macOS, bound transform clone memory, and release worker scratch storage after merging. Other platforms use ordinary AST copies.
- Reduce repeated type parsing, normalization, signature lookups, and source-file reads. Preserve frozen interface dispatch IDs when forking checker views.
- Allocate empty map storage on demand and reserve its final metadata in one allocation.
- Avoid allocating unobservable map-guard errors in the legacy C backend used to bootstrap the production compiler. Guards whose errors are read retain their existing behavior.

Measured on an 18-core Apple Silicon Mac, using a legacy bootstrap with `-gc none -prealloc -prod`, then running the full executable self-host with `-nocache -building-v -o v4 v3.v`. Non-CC time is the reported total minus CC time.

| Build | Non-CC time | Peak memory |
| --- | ---: | ---: |
| Master baseline (`741027d462`) | 664–823 ms | 4,416–4,421 MB |
| This change, four runs | 516–529 ms | 1,674–1,687 MB |

The candidate median was 523 ms; peak memory fell by about 62%. The branch also incorporates `7006be5e41`, including adapting its new FastC statistics call to the indexed line counter. A final full self-host on the PR commit measured 524.93 ms excluding CC, 133.87 ms transform, and 1,682 MB peak memory.

Validation:

- Full legacy compiler suite: 2,345 passed, 28 skipped. Four PATH-dependent language-server failures passed on rerun with `v` available. The remaining three failures reproduce before the change: arm64 assembly constraints, a C-output fixture requiring the V3 backend, and V3 assertion-format fixtures run through the legacy backend.
- Compiler diagnostics: 1,699 passed, one skipped; eight comptime checks passed. Focused guard/map/option C-output fixtures: seven passed, one skipped.
- Targeted builtin map, V3 type/scope, transform snapshot, worker-lifetime, C-generation, source-statistics, and upstream virtual-module tests passed. Map checks passed with autofree and AddressSanitizer. Snapshot checks passed with and without preallocation and under AddressSanitizer.
- Repeated full V3→V4 self-hosts passed. The generated V4 compiled and ran the map deletion/reservation regression suite; production V3 passed borrowed/escaping method-value and sum-variant regressions.

The guard C-output expectation intentionally changes from an allocated error to the existing static `none` marker. No CLI or language behavior change is intended. Known pre-existing V3 scoped-checker/storage and legacy JSON integration failures remain unchanged; `test-all` was not run.

### DIFF
--- a/vlib/builtin/map.v
+++ b/vlib/builtin/map.v
@@ -580,8 +580,30 @@ fn (mut m map) reserve_metas(meta_bytes u32) {
 
 // reserve ensures that the map can store at least `n` entries without rehashing.
 pub fn (mut m map) reserve(n u32) {
-	for u64(n) * 5 > u64(m.even_index) * 2 {
-		m.expand()
+	if m.len == 0 {
+		old_index := m.even_index
+		for u64(n) * 5 > u64(m.even_index) * 2 {
+			m.even_index = ((m.even_index + 2) << 1) - 2
+			if m.cached_hashbits == 0 {
+				m.shift += max_cached_hashbits
+				m.cached_hashbits = max_cached_hashbits
+			} else {
+				m.cached_hashbits--
+			}
+		}
+		if m.even_index != old_index {
+			// Empty maps have no entries to rehash. Allocate the final metadata
+			// table once instead of zeroing every intermediate doubled table.
+			meta_bytes := sizeof(u32) * (m.even_index + 2 + m.extra_metas)
+			unsafe {
+				free(m.metas)
+				m.metas = &u32(vcalloc_noscan(meta_bytes))
+			}
+		}
+	} else {
+		for u64(n) * 5 > u64(m.even_index) * 2 {
+			m.expand()
+		}
 	}
 	dense_cap := u64(n) + u64(m.key_values.deletes)
 	if dense_cap > u64(max_int) {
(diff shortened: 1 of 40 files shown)


medvednikov:
No blocking correctness findings. I ignored CI/results as requested.

I specifically checked:

The lazy empty-map representation: reads short-circuit before touching metas, first writes allocate backing storage, and clear/reserve/delete/clone/free remain consistent with the zero-storage state.
The map-guard error optimization: skipping is tied to the exact outer index expression, while observable err use and nested failing lookups retain error construction; the temporary generator state is restored afterward.
The macOS COW AST snapshots: ownership of the VM mappings is separated from the .nofree array views, failures fall back cleanly, and remapped pages are independently writable via COW.
Parallel transform/scoped-storage lifetime, type/import caches, scope fast lookups, markused fast paths, and the smaller Cgen allocation-elision changes. The scoped base-write restrictions still guard which AST slots workers may mutate.
Existing review discussion/threads so I wouldn't duplicate an already-raised issue.

The highest residual risk is the Darwin-specific VM snapshot path because it depends on low-level page/remap behavior, but I don't see a concrete defect in the implementation.

Code-review verdict: LGTM / reasonable to approve from the diff alone. I did not submit an approval or comment on GitHub.

PR #28396, opened 2026-09-05
v3: accept `-ldflags` and append it to the link command

medvednikov:
The V3 driver had no `-ldflags` case, so the option fell through to the
catch-all `starts_with('-')` branch in `vlib/v3/driver/driver.v` and the build
died with:

```
unknown option `-ldflags`
```

before compiling anything. `make LDFLAGS=...` bootstraps V with
`-ldflags "<flags>"`, and `v -ownership -ldflags ... dir` reaches this on every
platform, since `v3_ownership_forwarded_args` forwards the raw arguments to the
standalone `v3` executable.

The driver already understood link flags coming from the `LDFLAGS` *environment
variable* (`parse_v3_environment_flags('LDFLAGS')`), so only the CLI option was
missing.

## Changes

- Parse `-ldflags <flags>` like `-cflags` (`cmdexec.split_args`, so
  `-L "/dir with spaces"` survives), with the matching "requires a value" error
  and an entry in `v3_driver_option_consumes_value` so the `doc` pre-scan skips
  the value.
- Append the parsed flags after the ambient `LDFLAGS` — the order V1 uses
  (`env_ldflags` before `ccoptions.ldflags`, `vlib/v/builder/cc.v`) — on every
  link path: the C backend flag plan, FastC, and the direct TinyCC one.
  Object-only builds (`-o x.o`) keep skipping link flags, and
  `-generate-c-project` picks them up through the same flag plan.
- Record them in the `crun` build identity, so a cached binary is not reused
  after the link flags change.

## Tests

New `test_driver_ldflags_are_appended_to_the_link_command` in
`vlib/v3/tests/driver_cli_test.v` covers that the option is accepted, that the
ambient `LDFLAGS` come before the explicit `-ldflags` on the link command, that
the flags really reach the linker (an unknown `-l` library still fails the
build), and that a missing value is reported.

Manually verified on the C, FastC and TinyCC paths, and that `-showcc` shows the
flags at the end of the link command.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -1793,27 +1793,27 @@ fn safe_default_bin_file_name(filename string) string {
 }
 
 struct V3CCompilerFlagOptions {
-	environment_c_flags  []string
-	environment_ld_flags []string
-	target_args          []string
-	link_c_standard      string
-	dependencies         []string
-	warn_args            []string
-	vroot                string
-	target_os            string
-	target_arch          string
-	macos_sdk_root       string
-	pic_flag             string
-	is_prod              bool
-	no_prod_options      bool
-	is_shared            bool
-	parallel_cc          bool
-	large_c_unit         bool
-	limit_inlining       bool
-	explicit_tcc         bool
-	is_c_debug           bool
-	is_o                 bool
-	is_liveshared        bool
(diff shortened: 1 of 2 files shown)


medvednikov:
@codex review

PR #28397, opened 2026-09-05
v3: let C headers own their declarations, like the V1 backend

medvednikov:
V3 emitted a prototype for every `fn C.foo` declaration it could not prove was already declared. A V declaration is only ever an approximation of the real C signature — `&char` cannot express `const char*`, and a typedef'd `Window` or `Status` cannot be spelled at all — so placing one next to the header that already declares the symbol is a `conflicting types` error.

Every gg/sokol program hits this on Linux:

```
src.c:5362:5: error: conflicting types for 'XGetWindowAttributes'; have 'int(Display *, u64,  XWindowAttributes *)'
/usr/include/X11/Xlib.h:2701:15: note: previous declaration of 'XGetWindowAttributes' with type 'int(Display *, Window,  XWindowAttributes *)'
src.c:5365:5: error: conflicting types for 'fonsAddFontMem'; have 'i32(FONScontext *, char *, u8 *, i32,  i32)'
thirdparty/fontstash/fontstash.h:921:5: note: previous definition of 'fonsAddFontMem' with type 'int(FONScontext *, const char *, unsigned char *, int,  int)'
src.c:5383:5: error: conflicting types for 'sapp_run'; have 'i32(const sapp_desc *)'
thirdparty/sokol/sokol_app.h:16821:21: note: previous definition of 'sapp_run' with type 'void(const sapp_desc *)'
```

and so does a plain `#include <iconv.h>` on macOS (`conflicting types for 'iconv_close'`).

## What changed

Adopt the V1 backend's rule (`vlib/v/gen/c/fn.v`): a `fn C.` declaration only needs a generated prototype when no header can supply one — its file links a C source/object directly, or its module links a C library without including anything. `@[c_extern]` still forces a prototype, which is what `vlib/v/builder/cc.v` already tells users to reach for when a symbol has no header.

This replaces the hand-maintained "these headers declare these symbols" table (`pwd.h`, mbedtls, brotli). That table could never keep up with header-only libraries — sokol alone exposes ~240 entry points across `sapp_`, `sg_`, `sgl_` and `sfons_`, and Xlib several hundred more.

`vlib/builtin/builtin_backtraces_nix.c.v` declares `fn C.backtrace` next to an unrelated `#include`, so nothing else would declare it any more. The preamble now owns the backtrace API the way V1's `cheaders.v` does, `<execinfo.h>` when available and the glibc-shaped fallback prototypes (musl, older toolchains) when not.

## Testing

- New `vlib/v3/tests/c_header_declared_fn_codegen_test.v`: a `fn C.` declaration whose file includes a header that defines the function `static inline` with a `const char*` parameter. It fails on master with `error: conflicting types for 'api_call'` and passes here.
- `vlib/v3/tests/c_anonymous_param_codegen_test.v` checks the parameter lowering of emitted prototypes, so its declarations are now marked `@[c_extern]` to keep asking for them.
- `vlib/v3/gen/c/names_test.v` and `source_directive_test.v` cover the new rule directly, replacing the assertions that encoded the old policy.
- `v -silent test vlib/v3/gen/c/`: 19/19 pass.
- 20 C-interop tests from `vlib/v3/tests/`: 11 pass, 9 fail — the same 9 fail on unpatched master (checked with a pristine build of this base commit).
- Built and ran on macOS: `examples/gg/rectangles.v`, `examples/gg/rotating_textured_quad.v`, `examples/2048`, `examples/sokol/particles`, `examples/term.ui/rectangles.v`, `hello_world`, `json`, `regex`, `net_raw_http`, `database/sqlite`, `process/command`, plus programs exercising `net.http` over TLS (mbedtls), `compress.brotli`, `os.executable`/`os.loginname` (libproc, pwd.h) and `#include <iconv.h>`. The gg/sokol and iconv ones do not build on master.

### DIFF
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -406,6 +406,7 @@ mut:
 	inlined_c_fns                  map[string]bool
 	inlined_c_declared_fns         map[string]bool
 	files_with_c_includes          map[string]bool
+	files_with_c_postincludes      map[string]bool
 	files_linking_c_sources        map[string]bool
 	mods_with_c_libs               map[string]bool
 	mods_with_c_includes           map[string]bool
@@ -1113,6 +1114,7 @@ pub fn FlatGen.new() FlatGen {
 		inlined_c_fns: map[string]bool{}
 		inlined_c_declared_fns: map[string]bool{}
 		files_with_c_includes: map[string]bool{}
+		files_with_c_postincludes: map[string]bool{}
 		files_linking_c_sources: map[string]bool{}
 		mods_with_c_libs: map[string]bool{}
 		mods_with_c_includes: map[string]bool{}
@@ -2826,6 +2828,7 @@ pub fn (mut g FlatGen) gen_with_used_options(a &flat.FlatAst, used_fns map[strin
 	g.inlined_c_fns.clear()
 	g.inlined_c_declared_fns.clear()
 	g.files_with_c_includes.clear()
+	g.files_with_c_postincludes.clear()
 	g.files_linking_c_sources.clear()
 	g.mods_with_c_libs.clear()
 	g.mods_with_c_includes.clear()
@@ -4557,6 +4560,15 @@ fn (mut g FlatGen) note_c_include_directive(module_name string, source_file stri
 	}
 }
 
+// note_c_postinclude_directive records a `#postinclude`. Such a header is emitted
+// after every generated declaration and call site, so it cannot declare anything
+// for them: the generated prototype is the only declaration they can get.
+fn (mut g FlatGen) note_c_postinclude_directive(source_file string) {
+	if source_file.len > 0 {
+		g.files_with_c_postincludes[source_file] = true
+	}
+}
+
 // c_flag_links_c_source reports whether a `#flag` adds a C source or object file to
 // the link. Such a file ships no header, so the V declaration is the only prototype
 // the translation unit can get.
@@ -4581,24 +4593,27 @@ fn (mut g FlatGen) collect_c_directive(module_name string, node flat.Node, sourc
 	if node.kind != .directive {
 		return false
 	}
-	if node.value in ['include', 'preinclude', 'insert'] {
-		g.note_c_include_directive(module_name, source_file)
-	}
 	if node.value in ['preinclude', 'postinclude'] {
 		if node.typ.len == 0 {
 			return true
 		}
 		include_arg := c_include_arg_for_target(node.typ, g.compiler_vroot, source_file, g.target)
+		// A `#include linux <x.h>` contributes nothing when building for another
+		// target, so it must not make the file look header backed there.
 		if include_arg.len == 0 {
 			return true
 		}
 		directive := '#include ${include_arg}'
 		if node.value == 'preinclude' {
+			g.note_c_include_directive(module_name, source_file)
 			if directive !in g.preinclude_directives {
 				g.preinclude_directives << directive
 			}
-		} else if directive !in g.postinclude_directives {
-			g.postinclude_directives << directive
+		} else {
+			g.note_c_postinclude_directive(source_file)
+			if directive !in g.postinclude_directives {
+				g.postinclude_directives << directive
+			}
 		}
 		return true
 	}
@@ -4610,6 +4625,7 @@ fn (mut g FlatGen) collect_c_directive(module_name string, node flat.Node, sourc
 		if include_arg.len == 0 {
 			return true
 		}
+		g.note_c_include_directive(module_name, source_file)
 		// These helper headers are superseded by the inline compiler helpers emitted in
 		// builtin_abi_decls(); also including them would redefine the helpers.
 		if c_include_arg_is_builtin_abi_helper(include_arg, g.compiler_vroot) {
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -405,6 +405,10 @@ mut:
 	inlined_c_typedef_names        map[string]bool
 	inlined_c_fns                  map[string]bool
 	inlined_c_declared_fns         map[string]bool
+	files_with_c_includes          map[string]bool
+	files_linking_c_sources        map[string]bool
+	mods_with_c_libs               map[string]bool
+	mods_with_c_includes           map[string]bool
 	inlined_c_active_macros        map[string]bool
 	inlined_c_static_fns           map[string]bool
 	cache_omitted_c_fns            map[string]bool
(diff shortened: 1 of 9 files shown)


medvednikov:
@codex review

medvednikov (commented):
There looks to be a regression in the scoped/parallel C-extern path. `c_extern_forward_decls_scoped()` creates a fresh worker and `c_extern_decl_has_no_header()` now depends on `files_with_c_includes`, `files_linking_c_sources`, `mods_with_c_libs`, and `mods_with_c_includes`, but those four maps are not copied into `new_parallel_worker_config()` or `configure_c_extern_scan_worker()` (only `c_extern_forced_decls` is propagated).

As a result, the scoped worker sees all four maps as empty. For an ordinary unannotated `fn C.foo` whose file has `#flag helper.o`, or whose module has `#flag -lfoo` and no include, `c_extern_decl_has_no_header()` returns false and the generated prototype is skipped. Those are exactly the cases this change says should still emit a prototype, so in scoped mode they can fall back to an implicit C declaration / wrong inferred signature.

Please propagate the four new maps into the extern-scan worker (or otherwise make the predicate read the master's collected state), and ideally add an end-to-end scoped-worker regression test for the linked-object / `-lfoo` case.

medvednikov:
Good catch — confirmed and fixed in 2bf874dccd.

The diagnosis was exactly right. Reduced to a repro before touching anything:

```v
// main.v, next to helper.c with `int helper_fn(int x) { return x * 2; }`
#flag @VMODROOT/helper.c

fn C.helper_fn(x int) int

fn main() { println(C.helper_fn(21)) }
```

```
src.c:4180:19: error: call to undeclared function 'helper_fn'; ISO C99 and later
do not support implicit function declarations [-Wimplicit-function-declaration]
```

Same for a module with `#flag -L@VMODROOT -lmathhelper` and no include: no prototype in the generated C. Since `driver.v` sets `set_scope_parallel_workers(!generic_cache_hit)`, this was the path ordinary builds take, so the predicate was effectively answering "header backed" for everything and only `@[c_extern]` still emitted anything.

The four maps now go into `new_parallel_worker_config()` next to `inlined_c_declared_fns`, and into `configure_c_extern_scan_worker()` — I added `c_extern_forced_decls` to the latter too, so the scan worker's re-seeding list can't drift from the config again.

Two regression tests, both verified to fail without the propagation:

- `vlib/v3/gen/c/parallel_worker_test.v` — the scoped extern-scan worker must agree with its master on linked-object, `-lfoo` and header-backed files. Fails with `assert worker.should_emit_c_extern_decl_from_file('helper_fn', linked_file, 'linked')`.
- `vlib/v3/tests/c_linked_extern_prototype_codegen_test.v` — end-to-end, as suggested: links a C source and runs the binary (so a missing prototype is a hard error, not a guessed signature), plus a `-lfoo` module whose generated C must still carry `int lib_triple(int x);`.

Re-checked after the fix: `vlib/v3/gen/c/` 19/19; the 21-test C-interop set from `vlib/v3/tests/` went from 9 failures to 8 — `c_flag_target_filter_codegen_test.v` now passes, and the remaining 8 all fail on unpatched master too. gg/sokol examples, `net.http`+mbedtls, brotli, libproc/pwd.h and `#include <iconv.h>` all still build and run.

medvednikov (commented):
The scoped-worker issue from my previous review is fixed: the four header/linkage maps (plus `c_extern_forced_decls`) are now propagated, and the new worker + end-to-end tests cover the linked-source and `-lfoo` cases well.

I found two remaining holes in the directive bookkeeping:

1. `collect_c_directive()` calls `note_c_include_directive()` before `c_include_arg_for_target()` decides whether the include is active. A target-inactive include therefore still marks the file/module as header-backed. For example, on Linux a file with `#flag @VMODROOT/helper.c`, `#include windows <windows.h>`, and `fn C.helper_fn(...)` will set `files_with_c_includes` even though the Windows include is dropped, so the linked-source prototype is suppressed and the call can fail as an implicit/undeclared C function. Header ownership should only be recorded after the include resolves non-empty for the current target.

2. `postinclude` is not included in `['include', 'preinclude', 'insert']`, so it never records header ownership. In a module with `#flag -lfoo` plus `#postinclude "foo.h"`, the module is still treated as headerless and V3 emits the approximate V-derived prototype before the postinclude; if `foo.h` uses `const`/typedef-qualified types, that can recreate the same conflicting-types failure this PR is fixing. The current postinclude unit test does not expose this because, without a linked source/library, the bare case suppresses the prototype anyway.

I’d move/perform the include bookkeeping after target filtering and include `postinclude` in it, with regressions for an inactive target include + linked source and for `#postinclude` + `-lfoo`.

medvednikov:
Both fixed in be26930b76 — though the second one needed the opposite change from what was suggested, and it turned out to be a regression this branch introduced. Details below.

**1. Target-inactive includes — fixed as described.** Confirmed with your scenario before touching anything: on macOS, `#flag @VMODROOT/helper.c` + `#include windows <windows.h>` + `fn C.helper_fn` gives

```
src.c:3364:19: error: call to undeclared function 'helper_fn'
```

Ownership is now recorded only after `c_include_arg_for_target()` resolves the include non-empty, in both the `preinclude`/`postinclude` and the `include`/`insert` branch.

**2. `#postinclude` — it must not take ownership.** A postincluded header is emitted as the *last line* of the unit. Generated C for a `#postinclude "api.h"` module:

```
1418:  int api_call(char* name);        <- generated prototype
4898:      return api_call((char*)(s.str));   <- call site
6958:  #include "api.h"                 <- 6958 is the final line
```

So it can never be the declaration the call sites use. Suppressing the prototype makes every call to a postincluded API an implicit declaration, and that is already happening on this branch — a bare `#postinclude` module with no other directive:

```
error: call to undeclared function 'api_call'; ISO C99 and later do not support
implicit function declarations
```

That works on master and broke when this branch flipped the default, and the test I rewrote as `test_header_backed_declarations_do_not_get_a_second_prototype` was the guard for it — the original `test_header_directives_do_not_suppress_c_prototypes` asserted `postinclude_g.should_emit_c_extern_decl_from_file(...)` is *true*, and inverting that assertion is what let the regression through. My mistake, and thanks — chasing your scenario is what surfaced it.

So postinclude now records the opposite signal: the file still needs its prototype. That restores master's behaviour for postinclude and keeps your `#postinclude` + `-lfoo` case emitting, as it does on master and in the V1 backend (whose `file_has_c_includes` likewise excludes postinclude). The conflicting-types exposure you describe is real but inherent to postinclude — the header arrives too late to be the declaration either way — so the choice is a possible conflict at the end of the unit versus a guaranteed undeclared-function error at every call site. A user who wants the header to own the symbol should use `#include`, which does take ownership.

Regression tests, all verified to fail without the fix:

- `source_directive_test.v` — an inactive target include claims no ownership while the active one does; postinclude keeps its prototype.
- `c_linked_extern_prototype_codegen_test.v` — end-to-end: links a C source behind an inactive target include and runs it, and calls a postincluded function through a linked implementation.

Re-checked: `vlib/v3/gen/c/` 19/19; the C-interop set is unchanged at 8 failures, the same 8 that fail on unpatched master; gg/sokol examples, mbedtls, brotli, libproc/pwd.h and `#include <iconv.h>` all still build and run.

PR #28399, opened 2026-09-05
v3: keep disposable-arena memo entries out of the shared cgen caches

medvednikov:
`v install <pkg>` rebuilds `cmd/tools/vpm`, and that compile crashed:

```
v install ui2
cannot compile `/home/yeva/git/v/cmd/tools/vpm`: 255
0x7f98eb975047: at ???: RUNTIME ERROR: invalid memory access
src.c:1171: by string__eq
src.c:44770: by c__StringLookupCache__get
src.c:42817: by c__FlatGen__unique_qualified_struct_c_type
src.c:43306: by c__FlatGen__value_c_type
src.c:27065: by c__FlatGen__optional_payload_c_type
src.c:45904: by c__FlatGen__concrete_optional_type_name
src.c:27257: by c__FlatGen__optional_type_name_for_context
src.c:56066: by c__FlatGen__fn_return_type_name_for_context
src.c:56370: by c__FlatGen__forward_decl_items
src.c:56458: by c__FlatGen__forward_decls
src.c:56285: by c__parallel_type_decls_thread
src.c:27359: by workers__pool_worker
```

## Cause

`forward_decls` emits each 1024-item batch from a disposable prealloc arena, and
`parallel_type_decls_thread` drives the master generator from a pool thread. Both
only detached the four *context* lookup caches. Every other memo cache stayed on
the master: the `StringLookupCache` family (`unique_struct_ct_cache`,
`struct_cname_cache`, `flattened_generic_name_cache`, …), the name-fact caches,
`generic_app_cache`, `struct_decl_pref_cache`, `sum_variant_actual_cache`,
`param_types_cache` and `array_method_cache`.

An entry written while a scratch arena is current has its key, its value *and*
its map node allocated from that arena. When the batch frees the arena the
master's cache is left pointing at unmapped memory, so the next batch's lookup
hashes or compares a freed key — the `string__eq` frame above. The same caches
written from the type-declaration pool task are also the maps the body lanes
clone concurrently.

## Fix

Give both sites a full set of scratch-local caches via
`begin_scratch_lookup_caches` / `restore_scratch_lookup_caches`, so no cache the
generator keeps can hold an entry the arena owns. A cache the generator has
disabled stays disabled — the swap only changes *where* entries are written,
never whether they are memoized.

## Verification (macOS/arm64, 18 cores, `-prealloc` self-host build)

- `cmd/tools/vpm`, cold `VCACHE`/`V3CACHE`: master crashes 4/4 runs, patched
  passes 37/37.
- Generated C for `cmd/v` is byte-identical to master's, and stable across two
  further self-host generations.
- Added two unit tests in `vlib/v3/gen/c/names_test.v` covering the swap and the
  "disabled stays disabled" invariant.

### DIFF
--- a/vlib/v3/gen/c/fn.v
+++ b/vlib/v3/gen/c/fn.v
@@ -16882,11 +16882,6 @@ fn (mut g FlatGen) forward_decls() {
 		output_sb := g.sb
 		master_tc := g.tc
 		master_c_name_cache := g.c_name_cache
-		master_import_alias_cache := g.import_alias_cache
-		master_import_type_cache := g.import_type_cache
-		master_enum_selector_cache := g.enum_selector_cache
-		master_enum_method_cache := g.enum_method_cache
-		master_qualified_enum_method_cache := g.qualified_enum_method_cache
 		mut master_concrete_optional_params := g.cur_concrete_optional_params.move()
 		mut master_needed_optional_types := g.needed_optional_types.move()
 		mut master_fn_ptr_types := g.fn_ptr_types.move()
@@ -16898,13 +16893,10 @@ fn (mut g FlatGen) forward_decls() {
 		g.c_name_cache = &CNameCache{
 			base: master_c_name_cache
 		}
-		// Context-cache entries can own strings allocated by this disposable
-		// batch. Disable them until the master caches are restored.
-		g.import_alias_cache = unsafe { nil }
-		g.import_type_cache = unsafe { nil }
-		g.enum_selector_cache = unsafe { nil }
-		g.enum_method_cache = unsafe { nil }
-		g.qualified_enum_method_cache = unsafe { nil }
+		// Memo-cache entries own strings allocated by this disposable batch, and
+		// so do the maps holding them. Emit against batch-local caches, so that
+		// nothing this arena owns is still reachable once it is freed.
+		saved_lookup_caches := g.begin_scratch_lookup_caches()
 		g.cur_concrete_optional_params = map[string]bool{}
 		g.needed_optional_types = map[string]string{}
 		g.fn_ptr_types = master_fn_ptr_types.clone()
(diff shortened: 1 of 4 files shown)


medvednikov:
@codex review

medvednikov:
Addressed: rebased onto current `master` and added `import_type_cache` to the scratch bundle.

**[P1] `import_type_cache`** — fixed. It is now a field of `ScratchLookupCaches`, saved/restored with the rest, and given a fresh instance by `scratch_import_type_cache` only when the saved cache is non-nil, so a disabled cache stays disabled. Replacing the whole cache (rather than clearing it) also drops the `ImportFileTypeCache` children that `for_file` allocates on first use, which would otherwise keep both the file key and the child alive past the arena that made them.

The rebase replaced the explicit `g.import_type_cache = unsafe { nil }` disables in `forward_decls` and `parallel_type_decls_thread` with the `begin_scratch_lookup_caches()` call, so the cache is now memoized per batch instead of switched off entirely — same lifetime guarantee, one less cold path.

I also diffed the `FlatGen` cache fields between this branch's old base and current `master` to check nothing else had been added in the meantime: `import_type_cache` was the only one.

Tests extended: `test_scratch_lookup_caches_keep_batch_entries_out_of_the_generator` now writes into `unqualified_texts` and into a `for_file` child before restoring and asserts both are gone; `test_scratch_lookup_caches_leave_disabled_caches_disabled` covers the nil case.

**Re-verified on the new base** (macOS/arm64, 18 cores, `-prealloc` self-host build, fresh `VCACHE`/`V3CACHE` per run):

- `cmd/tools/vpm`: current `master` still crashes **5/5**, patched passes **12/12**.
- Codegen identity, same input tree, two compilers: emitted C for `cmd/v` and for `cmd/tools/vcreate` is byte-identical to `master`'s.
- `names_test.v`, `types_test.v` and `parallel_worker_test.v` pass.

Thanks for catching this — the merge would have silently reintroduced the UAF for that one cache.

PR #28400, opened 2026-09-06
v3: restore `.vsh` script mode

medvednikov:
## The problem

`make` is broken on master. Its last step runs `.github/problem-matchers/register_all.vsh`, and that fails to compile:

```
V has been successfully built
V 0.5.2 b4b6d28
.github/problem-matchers/register_all.vsh:8:20: error: unknown function `getenv`
.github/problem-matchers/register_all.vsh:8:20: error: const declarations do not support multiple return values yet
.github/problem-matchers/register_all.vsh:15:16: error: cannot use `string` as argument 1 to `chdir`; expected `&char`
.github/problem-matchers/register_all.vsh:15:17: error: unexpected `!`, the function `chdir` does not return a Result
.github/problem-matchers/register_all.vsh:16:10: error: unknown function `walk_ext`
.github/problem-matchers/register_all.vsh:17:28: error: unknown function `real_path`
make: *** [GNUmakefile:222: all] Error 1
```

`register_all.vsh` is not special — **every** `.vsh` script was broken:

```
$ printf 'println(getenv("HOME"))\n' > /tmp/t.vsh && ./v run /tmp/t.vsh
/tmp/t.vsh:1:9: error: unknown function `getenv`
```

Since #28378 the `v` that `make` produces contains only the V3 frontend, and V3 had no `.vsh` handling at all (`grep -rn is_vsh vlib/v3/` matched nothing). It parsed the script as a plain `.v` file, so V script mode was gone: no implicit `import os`, and no unqualified `os` symbols.

That explains each diagnostic:

- `unknown function getenv` — `os` was never in scope. V3 then synthesizes an empty `MultiReturn` for an unknown call, which is where the unrelated-looking `const declarations do not support multiple return values yet` on the same line comes from.
- `chdir ... expected &char` / `does not return a Result` — `chdir` resolved to `fn C.chdir(path &char) i32` (`vlib/builtin/cfns.c.v:187`), whose bare name V3 also registers, instead of `os.chdir(path string) !`.
- `walk_ext` / `real_path` unknown, then `.sorted()` on a non-value and `for in: cannot index ()` — all downstream.

`vlib/v3/test_all.vsh` writes `import os` and qualifies its calls, which is why the V3 CI stayed green.

## The fix

- **`vlib/v3/parser/parser.v`** — a `.vsh` file gets a synthetic `import os` node, skipped when the script imports `os` itself, with the alias registered before the body is parsed. Not added while `is_fmt`, so `v fmt` output for `.vsh` is unchanged.
- **`vlib/v3/flat/flat.v`** + the parallel-parse merge — an AST-wide `has_vsh_source` flag, so every lookup below is behind one bool and a compilation without a script pays nothing.
- **`vlib/v3/types/checker.v`** — inside a `.vsh` file, an unqualified name that resolves to nothing else falls back to `os.<name>`: functions (including generics) through `resolve_selective_import_symbol`, constants through `const_key_for_name`. Precedence matches V1 — a local or builtin V declaration still shadows `os`, but a bare `fn C.name` declaration no longer does. `resolve_any_selective_import_fn` gets the same fallback so monomorphization instantiates `read_file_array[u8](...)` correctly.
- **`vlib/v3/types/checker_tail_stmt.v`** — the bare-ident path resolves script-mode `os` constants (`args`, `path_separator`, `max_path_len`).
- **`vlib/v3/tests/vsh_script_mode_test.v`** — new. The existing `.vsh` coverage is in `vlib/v/slow_tests/inout`, which the V3 CI does not run, which is how this landed unnoticed.

## Verification

- `make` completes; `register_all.vsh` works in both branches (`GITHUB_JOB` set and unset).
- `vlib/v/slow_tests/inout/vscript_using_constants_in_os.vsh` and `vscript_using_generics_in_os.vsh` now match their `.out` files exactly.
- `examples/v_script.vsh` runs correctly again.
- `v fmt -verify` clean on every touched file.

### DIFF
--- a/vlib/v3/flat/flat.v
+++ b/vlib/v3/flat/flat.v
@@ -273,6 +273,10 @@ pub mut:
 	// or file_index_incomplete is set (a source file failed to read).
 	file_node_ids         []int
 	file_index_incomplete bool
+	// has_vsh_source records that at least one parsed source file is a `.vsh`
+	// script. V script mode makes the `os` module global inside such files, and
+	// the checker only pays for that lookup when this flag is set.
+	has_vsh_source bool
 	// source_buffers owns the storage behind zero-copy scanner strings retained
 	// by AST nodes. Keeping the buffers on the AST makes the lifetime boundary
 	// explicit and lets parser workers transfer ownership with their nodes.
--- a/vlib/v3/parser/parser.v
+++ b/vlib/v3/parser/parser.v
@@ -288,6 +288,12 @@ pub fn (mut p Parser) parse_into(path string) {
 	p.comptime_value_undos.clear()
 	p.comptime_value_scopes.clear()
 	p.imported_module_names.clear()
+	if !p.prefs.is_fmt && path.ends_with('.vsh') {
+		// V script mode: `os` is in scope from the first statement on, so the alias
+		// has to be known before the body is parsed, not only once the synthetic
+		// `import os` node below is appended.
+		p.imported_module_names['os'] = true
+	}
 	p.local_binding_counts.clear()
 	p.local_binding_undos.clear()
 	p.local_binding_scopes.clear()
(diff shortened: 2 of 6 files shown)


medvednikov:
@codex review

PR #28401, opened 2026-09-06
v3: use narrower integer types in the AST and node-indexed tables

medvednikov:
`int` is 64-bit on 64-bit targets now, so every AST field and side table declared `int` silently doubled in size. Nothing in these tables needs more than 32 bits: node ids index `FlatAst.nodes`, and positions are file-local byte offsets.

### What changed

- `token.Pos.offset/end/id` become `i32`, so the `Pos` embedded in every node is **16 bytes instead of 32**, and `flat.Node` drops from **104 to 88 bytes**.
- `flat.NodeId` becomes `i32` (8 → 4 bytes), halving `FlatAst.children` (one entry per AST edge) and every node-id side table, including the checker's direct/rewritten parent indexes.
- `token.File.line_offsets`, `FlatAst.file_node_ids`, the checker's top-level/synthetic/inactive declaration indexes, `fn_check_costs`, `strings_builder_candidates`, `Scope.generations`, the transformer's `source_parent_ids`/`fn_scan_costs`, cgen's `top_level_node_ids` and the driver's import-scan id lists become `[]i32`.
- The transformer's per-node module/file context tables kept a 24-byte string header per node for a handful of distinct spellings; they now keep a `u32` id into a small interned table.

Reads promote to `int` implicitly, so consumers are unchanged. Constructors (`token.new_pos`, `new_span`, `File.add_line`) keep their `int` parameters and narrow internally.

### Effect

Self-hosting `cmd/v` on macOS/arm64 in the `v self` configuration (`-cc tcc -gc none -prealloc`), RSS after each stage:

| stage | before | after |
|---|---|---|
| parse | 531 MB | 495 MB |
| check | 1858 MB | 1823 MB |
| markused | 2914 MB | 2870 MB |
| transform | 3470 MB | 3366 MB |
| cgen | 3009 MB | 2912 MB |
| **peak** | **5335 MB** | **5200 MB** |

Process peak footprint drops from 5.58 GB to 5.44 GB. A plain `-gc none` build (no prealloc) shows the same shape: peak 5387 → 5186 MB. Compile time is unchanged.

### Validation

- **`v self x5` succeeds**, and the last two generations are byte-identical binaries that emit byte-identical C for `cmd/v` — the self-host is a fixpoint, not a drifting chain. (This is also the `-prealloc` configuration, so the disposable transform arenas are exercised.)
- Generated C for a non-v3 program (`cmd/tools/vpm`) matches the pre-change compiler after path normalization; the only residual difference is string-literal numbering, which is already nondeterministic on `master` for that target.
- v3 module unit tests pass. The failures seen (`eval_test`, `gen/arm64/*_test`, `ssa/builder_test`) reproduce identically on `master` with the same cold caches.
- `v fmt -verify` is clean on every touched file (`types/scope.v` was already unformatted before this change).

### Note for reviewers

The v3 checker currently accepts `[]i32` where `[]int` is expected (it rejects `[]f64` and `[]string` in the same position), so an element-width mismatch across such a boundary is silent and only shows up as a garbage index at runtime. That is a pre-existing checker gap, not introduced here, but it made this refactor need a manual audit — every array whose element type changed had its call sites checked by hand, and three real boundaries were found and fixed (`collect_fn_gen_candidates_range`, `collect_gen_info_fn_preps` plus its `&[]int(...)` worker views, and `mark_const_array_ref_safe`).

Follow-up opportunity, not done here: the checker's node-indexed `resolved_call_names`/`resolved_fn_value_names` are `[]string` (24 bytes/node, ~120 MB at self-host scale). Converting them to interned ids needs the parallel-check write path to intern under a lock, so it is left for a separate change.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -15583,16 +15583,16 @@ fn synthetic_index_shift(insertions []SyntheticInsertion, idx int) int {
 // care about (.file markers/trailers, module_decl, import_decl) for every file
 // pair at or past region_start, in ascending node order. Falls back to the
 // full id range when the parser file index is unusable.
-fn collect_import_scan_ids(a &flat.FlatAst, region_start int, pair_cursor int) ([]int, int) {
+fn collect_import_scan_ids(a &flat.FlatAst, region_start int, pair_cursor int) ([]i32, int) {
 	if !file_index_usable_for_imports(a) {
-		mut all := []int{cap: a.nodes.len - region_start}
+		mut all := []i32{cap: a.nodes.len - region_start}
 		for i in region_start .. a.nodes.len {
 			all << i
 		}
 		return all, pair_cursor
 	}
 	mut cursor := pair_cursor
-	mut ids := []int{cap: 4096}
+	mut ids := []i32{cap: 4096}
 	mut last_trailing := region_start - 1
 	for cursor + 1 < a.file_node_ids.len {
 		marker := a.file_node_ids[cursor]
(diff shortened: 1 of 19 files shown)


medvednikov:
@codex review

PR #28381, opened 2026-09-05
macos: add typed Objective-C message senders

medvednikov:
## Summary

- add Cocoa-compatible `Point` and `Range` types and constructors
- add typed Objective-C message senders for mixed object, scalar, point, range, and rectangle signatures
- keep the ABI-sensitive `objc_msgSend` casts inside the macOS C bridge
- document the API and add Foundation/AppKit round-trip tests
- mark existing const C parameters correctly for current V3 C declarations

## Testing

- `./vnew -cstrict -cc clang -silent test vlib/macos/`
- `/Users/alex/code/v/vnew check-md vlib/macos/README.md`
- ui2 macOS Objective-C bridge tests against the new API

PR #28402, opened 2026-09-06
macos, ios: compile both modules for `-os ios`, not only macOS

medvednikov:
## Problem

Neither module can be imported when the target is iOS:

```
$ v -os ios -check main.v
error: cannot import module "ios" (no .v files in vlib/ios)
error: cannot import module "macos" (no .v files in vlib/macos)
```

Every source in the two modules is named `*_darwin.c.v`, and `should_compile_c` drops those whenever the target is not macOS:

https://github.com/vlang/v/blob/master/vlib/v/pref/should_compile.v#L277

So both modules are empty for an iOS build. `vlib/ios` is UIKit-only and unusable from any other target, which means it has never been compiled at all — that is also why its bridge still had unbridged casts.

## Fix

- `vlib/macos/macos_darwin.c.v` → `macos.c.v` and `vlib/ios/ios_darwin.c.v` → `ios.c.v`, so the suffix filter no longer excludes them. Contents are unchanged.
- `vlib/macos/macos_test.v` → `macos_darwin_test.v`, so the now unsuffixed module is not pulled into a Linux or Windows test run, where `objc/message.h` does not exist. The test runner honours the suffix and reports SKIP there, so nothing compiles the module on those hosts.
- Four bridged casts in `uikit_bridge.h`. `-os ios` compiles the generated sources with `-fobjc-arc`, and ARC rejects casts between `void *` and an Objective-C pointer unless they say `__bridge`.

## Verification

- `v -os ios -check` on a program importing both modules and calling `ios.color`, `ios.index_path` and `macos.description_string`: passes on this branch; on master it fails with the two errors above.
- `v test vlib/macos/` on macOS: passes, and the renamed test is still picked up.
- `v fmt -verify` and `v vet` on both modules: clean.
- Generated iOS C compiled with `cc -x objective-c -fobjc-arc`: clean. That is how the fourth cast, `[((NSIndexPath*)index_path) row]`, turned up — it is the one going the other way, `void *` to `id`.

Nothing else in the tree imports either module, and no file referenced the old names.

Note for #28381, which touches `vlib/macos/macos_darwin.c.v`: whichever lands second needs its changes moved onto `macos.c.v`.


medvednikov:
@codex review

PR #28406, opened 2026-09-06
net.ssl/net.mbedtls/net.quic: vendor a default CA trust store, verify by default

### LINKED ISSUE #28405: net.ssl/net.mbedtls/net.quic: no default CA trust store — HTTPS/HTTP-3 requests without explicit verify: always fail cert verification
## Describe the bug

`http.fetch(url: 'https://...', enable_http3: true)` fails against **every**
real HTTP/3 server (tested: `cloudflare-quic.com`, `google.com`, `http3.is`)
unless the caller explicitly sets `verify: '<path-to-a-CA-bundle.pem>'`. The
surfaced error is generic and gives no hint of the real cause:

```
h3: request to a freshly dialed connection failed after 20 attempts
```

## Root cause

Traced by driving `net.quic.dial()` directly against a real server and
polling the resulting `QuicConn` by hand. The connection actually closes
itself almost immediately, with a much more specific reason:

```
quic.QuicEvent{
    kind: connection_closed
    error_code: Option(298)
    reason: 'net.mbedtls: failed to parse CA bundle, mbedtls ret: -8576'
}
```

`vlib/net/http/transport_h3_d_http3.v` (`h3_dial_and_do`, ~line 194-203)
passes an **empty string** as the CA bundle whenever `req.verify` is unset:

```v
ca_pem := if req.in_memory_verification {
    req.verify
} else if req.verify != '' {
    os.read_file(req.verify) or { ... }
} else {
    ''   // <-- here
}
```

That empty string reaches `net.mbedtls.verify_certificate_chain`
(`vlib/net/mbedtls/x509_standalone.c.v:196`), which calls
`mbedtls_x509_crt_parse` on it -- parsing an empty PEM buffer fails
outright (mbedtls ret `-8576`), so certificate verification can never
succeed and the handshake is aborted every time, for every host.

**This is not actually HTTP/3-specific.** `verify_certificate_chain`'s own
doc comment already states the design explicitly: *"the caller supplies
the trust anchor explicitly — there is no OS trust-store lookup anywhere
in this codebase today, for any TLS client (HTTP/1.1, HTTP/2, or this QUIC
path)."* `net.mbedtls/ssl_connection.c.v`'s `new_sslcerts_from_file`/
`_in_memory` (backing `net.ssl`, the default HTTP/1.1/2 client backend on
non-Windows platforms) has the identical requirement, with no fallback,
and the OpenSSL backend (`-d use_openssl`) never calls
`SSL_CTX_set_default_verify_paths()` either. Windows' ordinary HTTP/1.1/2
path is shielded from this only because it goes through native SChannel
(`vschannel_*.c.v`), not `net.ssl`/mbedTLS at all -- so the gap is
invisible there, but real for h1/h2 anywhere `net.ssl` is actually used
(mbedTLS or OpenSSL backend), and always-on for HTTP/3 since `net.quic`
has no alternative backend at all.

## Reproduction

```v
import net.http

fn main() {
	resp := http.fetch(
		url:          'https://cloudflare-quic.com/'
		method:       .get
		enable_http3: true
	) or {
		eprintln('ERROR: ${err}')
		return
	}
	println('status: ${resp.status_code}')
}
```

```
v -d http3 run repro.v
ERROR: h3: request to a freshly dialed connection failed after 20 attempts
```

Confirmed workaround: passing `verify: '<path to a real ca-bundle.pem, e.g. curl's cacert.pem>'` succeeds end to end (`200`, `HTTP/3.0`).

## Expected behavior

`http.fetch()` should validate against a sensible default set of trusted
CAs out of the box, the same way it already does transparently on Windows
via SChannel -- matching the ergonomics of `curl`/browsers/other language
HTTP clients, which all ship or discover a usable root store without the
caller doing anything.

## Possible solution

Vendor a maintained root CA bundle (e.g. curl's `cacert.pem`, itself a
generated snapshot of Mozilla's trusted root store) into `thirdparty/`,
matching how `thirdparty/mbedtls` is already vendored, and use it as the
default trust anchor in `net.mbedtls` whenever a caller doesn't supply an
explicit `verify`/`ca_bundle_pem` -- fixing `net.ssl` (h1/h2, non-Windows)
and `net.quic` (h3, all platforms) from one shared place. A fuller,
follow-up improvement would be OS-native trust-store integration (Windows
CryptoAPI, macOS Keychain, Linux's handful of well-known
`/etc/ssl/.../*.crt` paths -- the same approach Go's
`x509.SystemCertPool()`/Rust's `rustls-native-certs` use), which also
picks up locally-installed/corporate CAs a vendored bundle never would;
noted here as a distinct, larger follow-up rather than blocking this fix.

## V version

`vlang/v` master, commit `f3cb8ed8f9` and earlier (present since HTTP/3
client wiring landed in #28129, and pre-existing in `net.ssl`/
`net.mbedtls` before that).

> [!NOTE]
>

quaesitor-scientiam:
## Summary

Fixes #28405: `http.fetch(enable_http3: true)` failed against every real
server with no explicit `verify:` set, because the HTTP/3 dial path passed
an empty CA bundle straight to `mbedtls_x509_crt_parse`, which fails
outright on an empty buffer. The exact same underlying gap exists for
HTTP/1.1/2 on any platform where `net.ssl`/`net.mbedtls` (rather than
Windows' native SChannel) is the actual TLS backend.

**Update:** several more rounds of external review (Codex, relayed both
directly and via other reviewers) landed after the original description
below was written, extending this PR well past its initial scope — real
insecure-default and validation gaps in the OpenSSL backend, QUIC
connection-ID protocol-state requirements (RFC 9000 §5.1.1/§5.1.2/§19.15/
§19.16), truncation/bounds checks on every new QUIC frame parser, a
portable non-retryable TLS-certificate error code threaded through every
backend, and proxy TLS-setup cleanup on error paths. The "Not in scope"
section below is kept up to date with what's actually still deferred.

**Cross-platform verification is now complete.** Windows (native SChannel
and the mbedTLS backend via `-d no_vschannel`) and WSL/Linux (system
trust-store probing) were verified earlier with identical real-server
results; macOS has now been independently verified too (2026-09-07, all
suites green) — no native Keychain integration there (see "Not in scope"),
but the vendored-bundle fallback works correctly, matching Windows/Linux
behavior for both halves of the fix.

## What changed

- **Vendored a default CA trust store.** curl's `cacert.pem` (a maintained
  snapshot of Mozilla's trusted root store,
  https://curl.se/docs/caextract.html) is now vendored into
  `thirdparty/cacert/`, matching how `thirdparty/mbedtls` is already
  vendored, and embedded into the binary at compile time
  (`net/mbedtls/default_ca_bundle.v`, via `$embed_file`). On Linux, the
  distro's own system CA bundle (`/etc/ssl/certs/ca-certificates.crt` and
  the other well-known paths) is probed first and preferred over the
  vendored snapshot when present.
- **Wired it in as the client-side fallback.** `verify_certificate_chain`
  (net.quic's HTTP/3 client) and `SSLConn.init`'s client-only cert-loading
  path (net.ssl's h1/h2 client) both fall back to this default whenever a
  caller doesn't supply their own `verify`/`ca_bundle_pem`. An explicitly
  configured `verify` **replaces** the default/system trust store
  entirely rather than being merged with it, and a certificate the
  selected trust store rejects stays rejected (no silent retry against
  the default roots) — both are load-bearing, tested properties, not
  incidental. `SSLListener` (server / mTLS role) is deliberately
  untouched — a public root store must never silently satisfy a
  client-certificate trust decision.
- **`net.openssl` now enables `SSL_VERIFY_PEER`** and verifies the
  requested hostname/IP during the handshake, loading default trust paths
  when no custom CA is supplied — this backend previously never checked
  certificates at all, regardless of `validate`.
- **`Request.validate`/`FetchConfig.validate` now default to `true`**
  (was `false`). mbedTLS's `authmode` was `VERIFY_OPTIONAL` by default,
  which performs verification but never aborts the handshake on failure —
  and nothing checked the result afterward. Ordinary `http.fetch()` calls
  on a non-Windows/non-vschannel backend were silently accepting *any*
  certificate, valid or not. This now fails closed by default; set
  `validate: false` to keep the old permissive behavior (HTTP/1.1/2 only —
  HTTP/3 has no skip-verification mode and rejects `validate: false`
  explicitly rather than silently ignoring it).
- **A portable `net.err_tls_certificate_invalid_code`** now identifies any
  deterministic certificate/trust-store setup or validation failure,
  consistently across mbedTLS, OpenSSL, SChannel, and HTTP/3's async path,
  so HTTP retry logic can recognize a hopeless TLS failure and stop
  retrying instead of exhausting `max_retries` against it.

## 3 more real bugs, found verifying against real servers

Rather than trusting the fixture/loopback test suite alone, I drove the
fix end-to-end against `cloudflare-quic.com`, `google.com`, and
`http3.is`. That surfaced 3 more genuine, distinct `net.quic` interop
bugs, all fixed here:

1. **`NEW_TOKEN` (0x07) and `NEW_CONNECTION_ID`/`RETIRE_CONNECTION_ID`/
   `PATH_CHALLENGE`/`PATH_RESPONSE` (0x18-0x1b) were entirely unparsed.**
   Any server sending one — confirmed: Google's QUIC endpoints send both
   `NEW_TOKEN` and `NEW_CONNECTION_ID` as standard practice — got the
   whole connection torn down as a protocol violation. All 5 are now
   parsed and validated (bounds/truncation checks per RFC 9000's own wire
   format, plus the connection-ID state-tracking §5.1.1/§5.1.2/§19.15/
   §19.16 requires independent of connection migration, which remains out
   of v1 scope); `PATH_CHALLENGE` is answered with `PATH_RESPONSE` per
   §8.2.1.
2. **`pump_pending_uni_header` checked its 32-byte header-buffering cap
   *before* attempting to classify the stream.** A peer writing its
   Stream Type header and immediately-following data (SETTINGS, QPACK
   instructions) in one write is entirely RFC 9114-legal — nothing
   requires the header to arrive in its own STREAM frame — but was
   wrongly rejected as "sent too much without completing its header."
   Confirmed against Google's and http3.is's QUIC stacks, both of which
   write the control-stream header + SETTINGS as a single ~45-byte write;
   Cloudflare's happens to split them across two writes, which is why
   this bug never reproduced against it.
3. **`apply_peer_settings` silently dropped the Set Dynamic Table
   Capacity instruction** it generated if this connection's own QPACK
   encoder stream hadn't been opened yet — while still mutating this
   encoder's *local* capacity state regardless, desyncing its view of its
   own dynamic table capacity from the peer decoder's, which then
   rejected the first real Insert instruction with "Error inserting entry
   with name reference." Confirmed against Google's QUIC endpoints, fast
   enough to process their own SETTINGS before this connection's
   `established()`-gated own-stream-opening logic has run. Now buffered
   and flushed once the real stream opens
   (`H3Conn.pending_own_encoder_instructions`).

**`http3.is` still doesn't work, and it isn't attributable to any of the
above** — it returns zero bytes to any QUIC packet at all. Its own error
page (screenshotted by the user mid-review) references pre-standardization
IETF draft ALPN tokens (`h3-29`/`h3-27`) and Chrome flags from that era —
strong evidence it never updated its listener to the finalized RFC 9114
protocol this client (correctly) speaks, rather than a bug here.

## Test plan

- [x] `vlib/net/quic` 54/54
- [x] `vlib/net/http` 33/36 + skips (platform/OpenSSL-header gated)
- [x] `vlib/net/mbedtls` 10/10 (1 known-flaky concurrency retry ignored by
      the runner)
- [x] `vlib/net/openssl` 2-3/3 depending on platform OpenSSL availability
- [x] New tests: `NewTokenFrame`/`NewConnectionIdFrame`/
      `RetireConnectionIdFrame`/`PathChallengeFrame`/`PathResponseFrame`
      round-trip + boundary/truncation tests (`frame_test.v`)
- [x] New direct regression test reproducing the combined header+payload
      classification-ordering bug (`h3_conn_test.v`)
- [x] New direct regression test for the encoder-instruction-buffering
      fix, calling `apply_peer_settings` before the encoder stream exists
      (`h3_conn_test.v`)
- [x] New regression test: an explicitly configured but empty HTTP/3 CA
      file is rejected rather than silently widening trust to the
      system/default bundle (`transport_h3_test.v`)
- [x] One pre-existing test (`test_parse_frame_rejects_unimplemented_frame_type`)
      updated to use frame type `0x30` (RFC 9221 DATAGRAM) instead of
      `0x18`, since every RFC 9000 base frame type is now handled
- [x] End-to-end against real servers post-fix: `cloudflare-quic.com` and
      `google.com` both succeed over HTTP/3 (`200`, `HTTP/3.0`) with no
      explicit `verify:` set
- [x] A real expired-certificate site (`expired.badssl.com`) is correctly
      rejected over both the mbedTLS backend (`-d no_vschannel`) and
      OpenSSL, where it was previously silently accepted by both
- [x] macOS: full suite green (user-verified, 2026-09-07) — vendored-bundle
      fallback (no native Keychain integration) succeeds against real
      servers and correctly rejects `expired.badssl.com`, matching
      Windows/Linux behavior

## Not in scope for this PR

- **Native OS trust-store integration for macOS (Keychain) and Windows
  (CryptoAPI) on the mbedTLS/HTTP-3 code paths.** Linux's system CA
  bundle is now probed and preferred (see above); macOS/Windows have no
  equivalent flat-file trust store, so this needs real, separate platform
  API bindings, not a path probe. Windows' *default* backend (SChannel)
  already validates via the OS certificate store independently of any of
  this — only the mbedTLS-backed path (`-d no_vschannel`) lacks OS
  integration on Windows/macOS.
- **Administrator-facing trust-store configuration**: `SSL_CERT_FILE`/
  `SSL_CERT_DIR`-style environment-variable overrides with documented
  precedence, and diagnostics for which trust source was actually
  selected. Go's `x509.SetFallbackRoots`/embedded-root-package model is a
  reasonable reference design for a follow-up.
- **An embedded-bundle maintenance policy**: source/version tracking and
  an update procedure for `thirdparty/cacert/cacert.pem`, with
  documentation that refreshing it requires rebuilding existing binaries
  (`$embed_file` compiles it in).
- **Excluding the embedded CA bundle from the default Windows (SChannel)
  build's binary.** `net.http`'s TLS dispatcher (`backend.c.v`, pre-existing
  this PR) is compiled unconditionally on every platform, so `net.ssl` (and
  transitively `net.mbedtls`, and now its ~189KB embedded bundle) links
  into every Windows binary even when SChannel handles 100% of TLS at
  runtime — confirmed empirically (issuecomment-5573383622): a default
  build with no `-d no_vschannel` still contains the embedded certificate
  data. This predates this PR (net.mbedtls was already linked, just far
  smaller before the embed); fixing it needs splitting `backend.c.v`'s
  net.ssl-dependent fallback out of the default Windows compilation unit,
  following this codebase's own established `_d_x.v`/`_notd_x.v`
  convention — a real, separately-scoped refactor with its own
  build-matrix verification surface, not attempted here.


JalonSolov:
Local AI findings:

Findings

**P1: HTTP/3 ignores the documented validate: false opt-out**
`request.v:42-50` now documents `validate = false` as allowing invalid certificates, but the HTTP/3 path always verifies certificates. The PR explicitly acknowledges this in the same comment. Users opting out of validation still cannot connect to self-signed or otherwise invalid-cert HTTP/3 endpoints. Either honor the flag in QUIC TLS or reject HTTP/3 configuration clearly.

**P2: NEW_CONNECTION_ID accepts an invalid zero-length connection ID**
`frame.v:563-580` only rejects IDs longer than 20 bytes. RFC 9000 requires the length to be 1–20 bytes, so `cid_len == 0` must be rejected. Add a lower-bound check and regression test.

**P2: PATH_CHALLENGE is silently ignored instead of answered**
`conn.v:1038-1055` parses `PATH_CHALLENGE` but sends no `PATH_RESPONSE`, despite RFC 9000 requiring one. This breaks path validation and can prevent connections involving migration, NAT rebinding, or active path probing from progressing. The parser support should either enqueue the matching response or avoid claiming full support for the frame.

quaesitor-scientiam:
Thanks — all 3 confirmed and fixed in `e2d95c4693`:

- **P1 (validate: false)**: agreed this was a real inconsistency. Rather than implementing skip-verification in `net.quic`'s TLS 1.3 stack (out of scope for this PR), `h3_round_trip` now fails fast on `validate: false` with a clear error, mirroring the existing `req.cert`/`req.cert_key` fail-fast check a few lines above it. Added `test_h3_round_trip_rejects_validate_false`.
- **P2 (zero-length connection ID)**: confirmed — RFC 9000 §19.15 invalidates both `cid_len < 1` and `cid_len > 20`; the original fix only checked the upper bound. Added the lower-bound check + `test_new_connection_id_frame_rejects_zero_length`.
- **P2 (PATH_CHALLENGE unanswered)**: confirmed and fixed for real — `PATH_CHALLENGE` now queues a `PATH_RESPONSE` echoing the same 8 bytes (RFC 9000 §8.2.1's MUST), drained on the next `drain_outgoing()` call via a new `pending_path_responses` list, mirroring the existing `pending_streams_blocked` pattern. Added `encode_path_response_frame` + round-trip/rejection tests and a direct regression test (`test_path_challenge_queues_a_path_response`).

Full `net.quic` (54/54) and `net.http` (33/35 + 2 pre-existing skips) green, plus re-verified against real HTTP/3 servers (cloudflare-quic.com, google.com) post-fix to confirm no regression.

JalonSolov:
Local AI found one more thing:

**P2: HTTP/3 documentation contradicts the new default trust-store behavior.**
In `request.v:50`, the `enable_http3` documentation says `req.verify` is required and that an empty value causes every request to fail. The PR now passes an empty CA bundle to the QUIC verifier, which falls back to the Linux system bundle or vendored CA bundle. It also says `validate:false` is silently ignored, while the implementation now rejects it explicitly.

quaesitor-scientiam:
Confirmed and fixed in `c1c1a05232` — `request.v`'s own separate copy of this doc comment (distinct from `FetchConfig`'s in `http.v`, which I'd already updated) was still describing the pre-fix behavior. Both files' `verify`/`enable_http3` field comments now correctly describe `net.mbedtls.system_or_default_ca_bundle_pem` as the default (Linux system bundle if present, else vendored) and the fail-fast on `validate: false`. Also caught and fixed the same staleness in `verify_certificate_chain`'s own doc comment in `net.mbedtls` while I was at it, which still said "no OS trust-store lookup... Linux's distro-provided bundle paths" after that support had already landed.

medvednikov:
Code-only review (CI intentionally ignored). Two additional findings:

**P2: `NEW_TOKEN` incorrectly accepts an empty token.**
`parse_new_token_frame` explicitly treats `Token Length == 0` as legal, and `test_new_token_frame_accepts_zero_length_token` locks that behavior in. RFC 9000 §19.7 says the opposite: the Token field **MUST NOT be empty**, and a client **MUST** treat an empty token as a `FRAME_ENCODING_ERROR`. Please reject `length == 0` and invert that regression test.

https://www.rfc-editor.org/rfc/rfc9000.html#section-19.7

**P2: `NEW_CONNECTION_ID` / `RETIRE_CONNECTION_ID` cannot be safely parsed and then discarded.**
`dispatch_one_rtt_frame` currently treats both as no-ops because migration is out of scope, but the required state transitions are not limited to migration. In particular, RFC 9000 §5.1.2 says that when `Retire Prior To` increases, the receiver **MUST stop using** the corresponding connection IDs and retire them with `RETIRE_CONNECTION_ID`; continuing to send on a server-retired DCID can break an otherwise non-migrating connection. §19.16 also requires a `PROTOCOL_VIOLATION` if a received `RETIRE_CONNECTION_ID` names a sequence number greater than any this endpoint previously issued. This implementation only has the initial local `scid` (sequence 0), so that validation is actionable even without a CID pool.

The observed Google case with `retire_prior_to == 0` is safe to tolerate, but generic support should either track/process the required CID retirement state or otherwise explicitly handle the unsupported stateful cases rather than silently accepting them.

https://www.rfc-editor.org/rfc/rfc9000.html#section-5.1.2
https://www.rfc-editor.org/rfc/rfc9000.html#section-19.16

medvednikov:
@codex review

medvednikov:
Addressed the latest review findings in `4f6987481c` and `e9467d8288`:

- reject empty NEW_TOKEN with FRAME_ENCODING_ERROR
- reject unsupported CID-retirement transitions safely and validate RETIRE_CONNECTION_ID against the largest locally issued sequence
- preserve request validate/verify/client-certificate settings through HTTP and SOCKS5 proxy TLS paths without breaking the existing SOCKS5 helper API
- document the secure-default behavior and private-CA/temporary opt-out migration paths in README.md

Validation: net.quic 54/54; net.http 31 passed + 4 skips / 35; net.socks 1/1; README.md check-md passed.

medvednikov:
@codex review

medvednikov:
Addressed the latest review findings in e64eb76ad6:

- OpenSSL now enables SSL_VERIFY_PEER, loads default trust paths when no custom CA is supplied, and verifies the requested DNS name/IP during the handshake.
- Added an OpenSSL regression covering validation disabled, an untrusted self-signed certificate, and a trusted certificate for the wrong host.
- mbedTLS now registers client credentials only when both certificate and key are present, and reports incomplete pairs explicitly.

Validation: focused OpenSSL regression passed; net.mbedtls 9/9 passed (known-flaky retry ignored by the runner); OpenSSL net.ssl 3 passed/1 skipped; net.http 32 passed/4 skipped.

medvednikov:
@codex review

medvednikov:
Addressed the follow-up review in aef1f1376c:

- documented that the default Windows SChannel backend ignores `verify`/`in_memory_verification`, including Windows store and `-d no_vschannel` migration options;
- mbedTLS no longer loads/parses the default CA bundle when `validate: false` and no custom CA is supplied.

Validation: README `check-md`, mbedTLS compile test, and mbedTLS shutdown regression all pass.

medvednikov:
@codex review

medvednikov:
Addressed the latest review in 40de5c1eba:

- failed HTTP CONNECT and SOCKS5 TLS setup now frees partial SSL state and closes the TCP socket; mbedTLS shutdown now safely frees pre-handshake state and is idempotent, with a regression test;
- invalid NEW_CONNECTION_ID retire ordering and CID lengths now carry FRAME_ENCODING_ERROR (0x07), with error-code assertions;
- corrected PATH_CHALLENGE/PATH_RESPONSE documentation to describe the implemented queue-and-response behavior.

Validation: focused proxy/OpenSSL/mbedTLS/QUIC tests pass; net.mbedtls 10/10 and net.quic 54/54 pass.

medvednikov:
@codex review

medvednikov:
Pushed `ea95c51cb7` to address the latest review:

- added a documented portable certificate-validation error code and made HTTP retry gates stop on it
- mapped mbedTLS, OpenSSL, SChannel, and asynchronous HTTP/3 certificate failures without losing the code through pooled, waiter, or proxy paths
- made truncated PATH_CHALLENGE/PATH_RESPONSE frames return FRAME_ENCODING_ERROR
- tracked peer CID advertisements, counting the initial sequence 0 toward the default limit of 2; duplicate sequence/content consistency and CID uniqueness are enforced, and excess IDs return CONNECTION_ID_LIMIT_ERROR

Validation:
- `./vnew -silent vlib/net/quic/frame_test.v`
- `./vnew -silent vlib/net/quic/conn_test.v`
- `./vnew -silent vlib/net/http/response_test.v`
- `./vnew -silent vlib/net/mbedtls/ssl_connection_cleanup_test.v`
- `./vnew -silent -d use_openssl vlib/net/http/openssl_validation_test.v`
- `./vnew -silent test vlib/net/http/h3_mux_conn_test.v vlib/net/http/transport_h3_test.v`
- `./vnew check-md vlib/net/README.md`
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed the latest Codex findings in `57270cef3f`:

- register the preferred-address connection ID/token as sequence 1 during EncryptedExtensions processing, so it counts toward `active_connection_id_limit`
- reject truncated non-empty NEW_TOKEN payloads with `FRAME_ENCODING_ERROR` (0x07)
- add regressions for both cases

Validation:
- `./vnew -silent vlib/net/quic/frame_test.v`
- `./vnew -silent vlib/net/quic/conn_test.v`
- `./vnew -silent test vlib/net/quic/` (54/54 passed)
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed the latest Codex finding in `59402bd55c` and audited the adjacent newly added frame parsers:

- truncated NEW_CONNECTION_ID CID/reset-token payloads now return `FRAME_ENCODING_ERROR` (0x07)
- missing/truncated NEW_CONNECTION_ID Sequence Number, Retire Prior To, and Length fields now return 0x07
- missing/truncated NEW_TOKEN Token Length now returns 0x07
- missing/truncated RETIRE_CONNECTION_ID Sequence Number now returns 0x07
- semantic violations retain their existing RFC-specific codes
- added regressions covering every new truncation exit

Validation:
- `./vnew -silent vlib/net/quic/frame_test.v`
- `./vnew -silent test vlib/net/quic/` (54/54 passed)
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed the latest Codex finding in `4486f6ba49`:

- enforce RFC 9000 §5.1.1 stateless-reset-token uniqueness across peer connection-ID sequence numbers
- apply the same check when sequence 0 is populated incrementally
- preserve exact-duplicate retransmission handling and existing CID-limit behavior
- add a regression for distinct CIDs reusing one reset token

Validation:
- `./vnew -silent vlib/net/quic/conn_test.v`
- `./vnew -silent test vlib/net/quic/` (54/54 passed)
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed the latest Codex finding in `cdd947e9d1`:

- reject every nonzero mbedTLS CA bundle/file parse result, including positive partial-parse counts
- preserve the raw mbedTLS return value in the message while returning the portable nonretryable TLS certificate code
- make every client CA/certificate/key parse fail immediately, preventing a later successful parse from overwriting an earlier setup failure
- broaden the public error-code docs to include certificate/trust-store setup failures
- add a regression using the default bundle plus a malformed PEM entry; it confirms the positive skipped-cert count and portable code

Validation:
- `./vnew -silent vlib/net/mbedtls/ssl_connection_cleanup_test.v`
- `./vnew -silent vlib/net/http/response_test.v` (includes retry-gate assertion)
- `./vnew -silent test vlib/net/mbedtls/` (10/10 passed; known flaky concurrency retries ignored by suite policy)
- `./vnew -silent test vlib/net/http/` (32 passed, 4 skipped)
- `./vnew check-md vlib/net/README.md`
- `git diff --check`

medvednikov:
@codex review

medvednikov:
Addressed the latest Codex finding in `1feea49042`:

- normalize OpenSSL default/custom trust-store setup failures to `net.err_tls_certificate_invalid_code`
- normalize validated client-certificate/private-key setup and SNI/peer-name verification setup failures to the same deterministic no-retry code
- preserve permissive-mode SNI behavior
- add an invalid-CA regression that asserts the portable code and HTTP retry gate

Validation:
- `./vnew -silent -d use_openssl vlib/net/http/openssl_validation_test.v`
- `./vnew -silent test vlib/net/openssl/` (2/2 passed)
- `./vnew -silent test vlib/net/http/` (32 passed, 4 skipped; run immediately before this OpenSSL-only follow-up)
- `git diff --check`

An additional full `-d use_openssl` HTTP matrix run reached 35/36, then the existing `transport_h2_test.v` hung for several minutes and was stopped; the focused changed paths above passed.

medvednikov:
@codex review

medvednikov:
Addressed the latest Codex findings in e53eb60fcc:

- removed the arbitrary OpenSSL certificate-chain verification depth limit
- reject NEW_CONNECTION_ID with PROTOCOL_VIOLATION when the peer selected a zero-length initial connection ID, with regression coverage

Validation:
- ./vnew -silent vlib/net/quic/conn_test.v
- ./vnew -silent -d use_openssl vlib/net/http/openssl_validation_test.v
- ./vnew -silent test vlib/net/quic/ (54/54)
- ./vnew -silent test vlib/net/openssl/ (2/2)
- ./vnew -silent test vlib/net/http/ (32 passed, 4 platform skips)
- git diff --check

medvednikov:
@codex review

medvednikov:
Addressed the latest Codex findings and adjacent deterministic retry leaks in dbef24ae11:

- reject incomplete mbedTLS client credentials before trust-store parsing with the portable certificate-error code
- preserve that non-retryable code for HTTP/3 unsupported TLS options and unreadable custom CA files
- preserve it for OpenSSL in-memory CA/cert/key temp-file write failures
- added focused error-code and HTTP retry-gate regressions

Validation:
- focused mbedTLS cleanup, HTTP/3 transport, OpenSSL write-error, and HTTP retry-gate tests
- ./vnew -silent test vlib/net/mbedtls/ (10/10; known-flaky concurrent response test ignored after configured retries)
- ./vnew -silent test vlib/net/openssl/ (3/3)
- ./vnew -silent test vlib/net/http/ (32 passed, 4 platform skips)
- git diff --check

medvednikov:
@codex review

medvednikov:
Addressed the latest Codex findings in f620bfc06e:

- failed mbedTLS dials now free and null the separately allocated ALPN pointer array before marking cleanup complete
- added a failed-ALPN-dial cleanup/idempotent-shutdown regression
- corrected both public request/config docs: empty verify uses the system-or-vendored bundle for mbedTLS and HTTP/3, OpenSSL default trust paths for OpenSSL, and the Windows certificate store for SChannel

Validation:
- ./vnew -silent vlib/net/mbedtls/ssl_connection_cleanup_test.v
- ./vnew -silent vlib/net/http/response_test.v
- ./vnew -silent test vlib/net/mbedtls/ (10/10; known-flaky concurrent response test ignored after configured retries)
- formatting and git diff --check

medvednikov:
@codex review

medvednikov:
Addressed the latest Codex finding in 54d619adc5:

- OpenSSL now rejects incomplete client certificate/key pairs up front with the portable, non-retryable certificate error code, matching mbedTLS
- added focused regression coverage for the validation-disabled/in-memory case as well, so the invariant does not depend on peer-validation mode

Validation:
- ./vnew -silent vlib/net/openssl/ssl_connection_errors_test.c.v
- ./vnew -silent test vlib/net/openssl/ (3/3)
- ./vnew -silent -d use_openssl vlib/net/http/openssl_validation_test.v
- formatting and git diff --check

medvednikov:
@codex review

Maokaman1:
More from Codex:

Making certificate verification the default and preferring the Linux system bundle are welcome improvements. For corporate deployments, the desired behavior is that administrators can manage trusted CAs through the OS without rebuilding applications.

**One edge case to address here:** an explicitly configured but empty CA file appears to activate default trust in HTTP/3. The [HTTP/3 transport](https://github.com/quaesitor-scientiam/v/blob/54d619adc5c83bc0c5c9d1d78d2def708a27babf/vlib/net/http/transport_h3_d_http3.v#L207) reads the file into a string, while the [certificate verifier](https://github.com/quaesitor-scientiam/v/blob/54d619adc5c83bc0c5c9d1d78d2def708a27babf/vlib/net/mbedtls/x509_standalone.c.v#L213) interprets an empty string as “use system/default roots.” These should be distinct cases: no CA configured versus an unusable explicitly configured CA. Please reject the latter and add a regression test.

More generally, please document and test this trust contract:

- An explicitly supplied CA bundle replaces default roots; adding certificates to system roots should be a separate, explicit operation.
- A certificate rejected by the selected trust store must remain rejected. Do not retry verification against bundled roots or automatically merge them into an administrator-managed store, since that can restore trust in deliberately removed CAs.
- Distinguish absent system roots from empty or unreadable stores. The current Linux loader skips all read errors and empty files; deployments should have a way to require system/configured trust and disable embedded fallback.

Go provides a useful model: [`x509.SetFallbackRoots`](https://go.dev/pkg/crypto/x509/?m=old#SetFallbackRoots) supplies fallback roots when no system trust source is available. Its [embedded root package](https://pkg.go.dev/golang.org/x/crypto/x509roots/fallback) is an application opt-in and explicitly requires ongoing updates.

For follow-up work, I would prioritize:

1. **Native Windows/macOS verification for the mbedTLS/HTTP/3 paths**, preserving OS trust settings. Installing a corporate CA should work consistently across HTTP versions.
2. **Runtime configuration for administrators**, including `SSL_CERT_FILE`/`SSL_CERT_DIR` support with documented precedence and replacement semantics.
3. **An embedded-bundle maintenance policy:** source/version, update procedure, and clear documentation that updating embedded roots requires rebuilding existing binaries.
4. **Trust-source diagnostics and reload documentation**, so administrators can determine which store was selected and when CA changes take effect.

Also, the PR description still lists OpenSSL verification and Linux system-store integration as out of scope, although the current implementation includes them. Updating it would make the final behavior much easier to review.

quaesitor-scientiam:
Thanks for the thorough follow-up. The empty-CA-file edge case is fixed in `e51801681a`:

- Confirmed exactly as described: `req.verify` reads its file into a string for HTTP/3 (net.quic's `DialParams` carries CA trust as in-memory PEM, not a path), and `verify_certificate_chain` treats an empty result as "no CA configured" by design — so a real, explicitly-named file that happens to be empty silently widened trust to the system/default bundle instead of failing. Now rejected with the portable `net.err_tls_certificate_invalid_code`, and a regression test locks it in.
- Checked the h1/h2 mbedTLS path for the same gap (it hands the file path straight to `mbedtls_x509_crt_parse_file`, so I wasn't sure it shared the ambiguity) — it doesn't: mbedTLS itself already rejects a 0-byte file with a nonzero return, which the existing "reject every nonzero parse result" fix already turns into a hard error. Added a test proving that empirically rather than leaving it as an assumption.
- Documented the trust contract in README.md: an explicit `verify` **replaces** the default/system store (never merged), and a certificate the selected store rejects stays rejected (no retry against default roots) — both already true of the implementation, now stated explicitly.
- Updated the PR description's "Not in scope" section, which was stale — it still listed OpenSSL verification and Linux system-store integration as unimplemented even though both landed in the intervening rounds.

On the four follow-up priorities (native macOS/Windows trust-store integration, `SSL_CERT_FILE`/`SSL_CERT_DIR` admin configuration, an embedded-bundle maintenance policy, and trust-source diagnostics): agreed these are the right next steps, and I've recorded them explicitly in the PR description's "Not in scope" section rather than leaving them implicit. Not attempting them in this PR — each is a real, separately-scoped piece of work (native OS trust-store bindings in particular, per the PR's own existing scope note), and bundling them in risks masking regressions in what's already landed. Happy to open follow-up issues for these if that's useful.

Maokaman1:
At `e51801681a`, the bundle added **237,309 bytes without `-prod`** and **209,122 bytes with `-prod`** to my Windows HTTPS test executable. Even the default Windows SChannel build added **206,678 bytes** with `-prod`, despite SChannel validating against the Windows trust store and never consulting the embedded bundle.

quaesitor-scientiam:
Verified this empirically rather than just reasoning about it — built a minimal `import net.http` program with default flags (no `-d no_vschannel`) and confirmed the resulting Windows binary does contain the embedded `cacert.pem` (`BEGIN CERTIFICATE` is present in the executable), even though that build's runtime TLS path is exclusively SChannel and never reads `default_ca_bundle_pem`.

**Root cause, and why it's Windows-specific:** `backend.c.v` (`net.http`'s TLS dispatcher, pre-existing this PR) is compiled unconditionally on every platform. It contains `net_ssl_do`/`h2_do`/`do_request`/`h1_exchange_ssl`, all of which reference `ssl.SSLConn` — so `import net.ssl`, and transitively `net.mbedtls`, resolve and compile into every `net.http` binary regardless of platform, even though `ssl_do`'s own `$if windows && !no_vschannel` dispatch means those functions are only ever *called* on Windows when a caller explicitly opts out of SChannel with `-d no_vschannel`. This is genuinely pre-existing architecture, not something this PR introduced — but this PR is what made the cost material: before it, `net.mbedtls` was already linked into every Windows binary, just without a ~189KB embedded file attached to it.

I looked at fixing this now (gating `default_ca_bundle_pem`'s `$embed_file` behind a `$if windows && !no_vschannel` const expression, so the embed itself disappears on a default Windows build without touching `backend.c.v`'s broader dispatch logic) but couldn't get V's const-level `$if` to accept a combined OS-tag + custom-flag condition in my initial attempt, and I'm not comfortable pushing a half-verified change to this exact code path under time pressure — getting a CA-trust gating condition subtly wrong is precisely the class of bug this whole PR exists to close, and "silently embeds nothing" would be a regression, not a neutral no-op.

Rather than rush it, I've flagged this as its own follow-up (file-splitting `backend.c.v`'s net.ssl-dependent fallback code out of the default Windows compilation unit, following this codebase's own established `_d_x.v`/`_notd_x.v` convention already used for `server_tls_*_use_openssl.v` and similar) — a real, valuable fix, but a separate, mechanical refactor with its own build-matrix verification surface (default Windows / `-d no_vschannel` / cross-compile), not something to fold into an already-large PR under review pressure. Thanks for the concrete numbers — they made the mechanism easy to confirm rather than argue about.

medvednikov:
Let's do it like Go

quaesitor-scientiam:
??

JalonSolov:
> Go provides a useful model: [`x509.SetFallbackRoots`](https://go.dev/pkg/crypto/x509/?m=old#SetFallbackRoots) supplies fallback roots when no system trust source is available. Its [embedded root package](https://pkg.go.dev/golang.org/x/crypto/x509roots/fallback) is an application opt-in and explicitly requires ongoing updates.

@quaesitor-scientiam I think he meant this bit...

quaesitor-scientiam:
> > Go provides a useful model: [`x509.SetFallbackRoots`](https://go.dev/pkg/crypto/x509/?m=old#SetFallbackRoots) supplies fallback roots when no system trust source is available. Its [embedded root package](https://pkg.go.dev/golang.org/x/crypto/x509roots/fallback) is an application opt-in and explicitly requires ongoing updates.
> 
> @quaesitor-scientiam I think he meant this bit...

I need this PR merged in order to update it like GO

JalonSolov:
Bouncing to re-run CI...

JalonSolov:
<sigh> ... CI seems to be broken again.  :-(

medvednikov:
What I meant is use the operating system trust store by default, and make embedded roots an explicit application-provided fallback rather than silently vendoring Mozilla roots into every binary. 

I'll implement this.

medvednikov:
You’re right: Go’s standard `crypto/x509` does not embed a Mozilla bundle into every binary. It uses the native Windows/macOS verifier or OS-managed files/directories on Unix; only the separate opt-in fallback package embeds roots.

I’m implementing that split here. A normal `net.http` binary will contain no CA blob. The bundled roots are reachable only through an explicit fallback install, for minimal containers or other systems with no usable trust store. I’ll also verify the compiled binary itself, not just the source dependency graph.

medvednikov:
I'm basing my new PR on this one.

quaesitor-scientiam:
> You’re right: Go’s standard `crypto/x509` does not embed a Mozilla bundle into every binary. It uses the native Windows/macOS verifier or OS-managed files/directories on Unix; only the separate opt-in fallback package embeds roots.
> 
> I’m implementing that split here. A normal `net.http` binary will contain no CA blob. The bundled roots are reachable only through an explicit fallback install, for minimal containers or other systems with no usable trust store. I’ll also verify the compiled binary itself, not just the source dependency graph.

What I was working on addressing

PR #28407, opened 2026-09-06
v3: make the self-host ~20% faster (compiler phases 3.6x, peak RSS -67%)

medvednikov:
Three lowering/codegen fixes found by sampling the self-host build
`cd vlib/v3 && ../../v -gc none -prealloc -prod -o v3 v3.v && ./v3 -nocache -building-v -o v4 v3.v`
(338k lines).

### `arr.sort(...)` was lowered to an insertion sort

`make_array_default_sort_stmt` emitted a plain insertion sort for every element type
except primitives with a default comparator (those already call `v3_array_sort_*` =
`qsort`). So every other sort was quadratic. The parallel checker sorts its own 24k
element work list twice per run in `types.split_check_items`: **604 ms of the 2241 ms**
the compiler spent outside `cc`, all serial on the main thread.

It now lowers to a bottom-up merge sort through a scratch buffer. That keeps the
**stable** order the V1 backend produces with `v_stable_sort` (`vlib/v/gen/c/cheaders.v`),
so element order for equal keys is unchanged. Sorting 24k structs: 530 ms -> 3 ms.

### `const` maps were re-materialized at every use site

`x in m` and `m[k]` called `const_expr_for_ident()` and lowered the constant's
*initializer*, so each lookup rebuilt the whole literal: a `new_map` plus one
`map__set` (hash + key clone) per entry. `naming.c_name()` alone was rebuilding a
46-entry and a 56-entry map for **every emitted identifier**.

Constants are already emitted once into a module-scope C global (`naming__reserved_words`
existed in the output and nothing referenced it), so the identifier now lowers to that
global. Inside a constant's own initializer the global is not assigned yet, so the
initializer is still substituted there (`t.in_const_init`). This is also where most of
the peak RSS went.

### The C preamble's `wyhash` mixed one byte per iteration

`builtin_abi_decls` emitted a `wyhash` doing two multiplications *per byte* of the key.
`gen/fastc/fastc.v` already carried a version that mixes a 64-bit word per step; this
ports it to the C backend, so both preambles now agree. V maps keep insertion order
independently of the hash, so program output is unaffected.

## Numbers

Interleaved medians, five runs per binary, 18-core M5 Max, `./v3 -nocache -building-v -o v4 v3.v`:

| | master | this PR | |
|---|---|---|---|
| compiler phases (excl. `cc`) | 2241 ms | **620 ms** | −72% (3.6x) |
| non-cc throughput | 151k lines/s | **546k lines/s** | |
| peak RSS | 5949 MiB | **1977 MiB** | −67% |
| total wall (incl. `cc`) | 8286 ms | **6653 ms** | −19.7% |

Per stage: check 957 -> 132 ms, transform 618 -> 238 ms, cgen 416 -> 152 ms,
markused 123 -> 40 ms. `cc` is unchanged (6051 -> 6028 ms) and is now ~90% of the
command's wall time.

## Checks

- `v3` and the `v4` it builds emit **byte-identical C** (fixed point).
- Sort order and stability verified against a master-built compiler for structs,
  strings, ints, `sort_with_compare`, `sorted`, descending comparators, and empty /
  single-element arrays; also under `-cc tcc`.
- Const-map edge cases give identical output to master: const derived from another
  const map, `__global` initialized from a const map, cross-module const maps, nested
  const maps, const map passed as a `map[string]int` argument, `for k, v in`.
- 20 vlib tests (incl. `builtin/array_test.v`, `builtin/map_test.v`) and 40
  `examples/*.v`: identical pass/fail sets to master.
- `-d v3_no_parallel` builds.
- `v fmt -verify` clean.

Partial run of `vlib/v3/**/*_test.v` against both a master-built and a patched
compiler agreed on every file covered (identical pass/fail set); the full sweep is
left to CI. Note `vlib/v3/test_all.vsh` currently cannot run on master at all — it
invokes the host `v` with `-old-compiler`, which a V3-only executable rejects.

### DIFF
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -19209,7 +19209,34 @@ fn (mut g FlatGen) builtin_abi_decls() {
 	g.writeln('static const u64 _wyp[4] = {0x2d358dccaa6c78a5ull, 0x8bb84b93962eacc9ull, 0x4b33a62ed433d4a3ull, 0x4d5a2da51de1aa47ull};')
 	g.writeln('static inline u64 _wymix(u64 a, u64 b) { u64 ha = a >> 32, hb = b >> 32, la = (u32)a, lb = (u32)b, hi, lo; u64 rh = ha * hb, rm0 = ha * lb, rm1 = hb * la, rl = la * lb, t = rl + (rm0 << 32), c = t < rl; lo = t + (rm1 << 32); c += lo < t; hi = rh + (rm0 >> 32) + (rm1 >> 32) + c; return lo ^ hi; }')
 	g.writeln('static inline u64 wyhash64(u64 a, u64 b) { a ^= _wyp[0]; b ^= _wyp[1]; a *= 0xa0761d6478bd642full; b *= 0xe7037ed1a0b428dbull; return (a ^ (a >> 32)) ^ (b ^ (b >> 32)); }')
-	g.writeln('static inline u64 wyhash(const void* key, size_t len, u64 seed, const u64* secret) { const unsigned char* p = (const unsigned char*)key; u64 h = seed ^ secret[0] ^ (u64)len; for (size_t i = 0; i < len; i++) h = wyhash64(h ^ (u64)p[i], secret[(i + 1) & 3]); return h; }')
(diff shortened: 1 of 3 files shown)


medvednikov:
@codex review

PR #28409, opened 2026-09-06
v3: split the C compile of a large program across cores (self-host 2.4x)

medvednikov:
After #28407 the self-host spends **670 ms** in all of V's own phases and
**6.1 s** in one `cc` process. 90% of
`./v3 -nocache -building-v -o v4 v3.v` is a single-threaded C compiler working
through one 22 MB translation unit, so that is where the remaining time is.

The split C build already existed behind `-parallel-cc` (#28389). This turns it
on for the builds where it is a pure win and removes the things that were making
it cost almost as much as it saved.

### Auto-enable it, without touching the cached path

A split build is enabled when the object cache is *not* the one feeding the C
compiler — so there is nothing to reuse and the whole program is about to be
compiled from scratch anyway — and the program is big enough to be worth
splitting (400k AST nodes; hello world stays in one unit and is unchanged at
62 ms). `-prod` keeps its monolithic unit because `-flto` has to see the whole
program. `-d v3_parallel_cc_monolithic` still opts out.

### Bound the C compilers by memory, not by a constant

`v3_parallel_cc_max_jobs` was 2, which is what made the flag only worth ~25%.
A unit is the shared declaration prefix plus its share of the bodies, and clang
needs ~35 bytes of resident memory per source byte of a generated unit, so the
job count now follows `runtime.free_memory()` and the unit size. On this machine
that is 18 jobs and a measured 2.2 GiB of concurrent C compilers, against
840 MiB for the single monolithic one. On a machine with less free memory it
scales down, and it falls back to the old bound of 2 if free memory cannot be
read.

Also one unit per job instead of four: every extra unit re-parses the 2.3 MB
declaration header (0.09 s each), which costs more than the finer load balancing
returns.

### Stop paying for cache machinery a split build does not use

`-parallel-cc` had to turn on `cache_split`, which also turns on the
content-addressed renaming of `_str_N` symbols. That renaming is what lets
separately compiled *module cache* generations agree on a symbol's name; a split
build shares one generated prefix with its own units and does not need it. It
was a byte-by-byte pass over all 22 MB, on the main thread, once per worker
segment — **720 ms** of the cgen stage. It is now behind its own
`cache_stable_symbols` flag, set only when the module cache is on.

With that gone, a split build can also stream the unit markers straight into the
mapped output file instead of copying every function body through the builder
first (another ~60 ms).

### Test binaries stay monolithic

`gen_test_failure_global()` emits `static int __v3_test_failures` and the setjmp
buffer into the program *body*, so a unit that only referenced them would not
compile. Test builds keep one translation unit. (This also fixes the explicit
`-parallel-cc` flag for test files, which produced unbuildable C before.)

## Numbers

Interleaved medians, five runs per binary, 18-core M5 Max,
`./v3 -nocache -building-v -o v4 v3.v`:

| | master | this PR | |
|---|---|---|---|
| `cc` | 6127 ms | **2213 ms** | −64% |
| V's own phases | 668 ms | 674 ms | unchanged |
| **total** | **6798 ms** | **2895 ms** | **−57%** (2.35x) |

Peak resident memory of the `v3` process is unchanged (1.96 GiB); the C
compilers add 2.2 GiB concurrently at the peak instead of 840 MiB, which is what
the memory-aware job count bounds.

## Checks

- `v3` and the `v4` it builds emit byte-identical C (fixed point), and a
  compiler built through the split path emits the same C as one built with
  `-d v3_parallel_cc_monolithic`.
- Cached builds are unchanged: the C stored in the module cache is byte-identical
  to master's, content-addressed symbol names included, and cold+warm builds
  reuse it.
- `hello_world` stays in one translation unit; `-cc tcc`, `-d v3_no_parallel`,
  and explicit `-parallel-cc` all still work.
- `vlib/v3/driver/parallel_cc_test.v`, `modulecache_test.v`,
  `gen/c/source_directive_test.v`, `gen/c/types_test.v` pass.
- 20 vlib tests and 40 `examples/*.v`: identical pass/fail sets to master.
- `v fmt -verify` clean.
- A file-by-file A/B of the `vlib/v3/**/*_test.v` unit tests (each with its own
  V3CACHE) against a master-built compiler agrees on every file so far; the three
  failures seen (`eval_test.v`, `gen/arm64/gen_test.v`, `gen/arm64/linker_test.v`)
  fail identically on master.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -1823,8 +1823,26 @@ struct V3CCompilerFlagPlan {
 }
 
 const v3_parallel_cc_unit_marker = '/* V3PARALLEL_CC_UNIT */'
-const v3_parallel_cc_max_jobs = 2
-const v3_parallel_cc_units_per_job = 4
+const v3_parallel_cc_max_jobs = 32
+// One unit per job: every extra unit re-parses the shared declaration header, so
+// splitting finer than the number of concurrent compilers costs more than the
+// better load balancing returns.
+const v3_parallel_cc_units_per_job = 1
+// Used when the free memory of the machine cannot be read; this is the bound the
+// split C build shipped with before it became memory-aware.
+const v3_parallel_cc_fallback_jobs = 2
+// Peak resident size of one C compiler process, measured against the source size
+// of its translation unit (clang needs ~35 bytes per source byte on a generated
+// unit), plus a floor for the fixed per-process cost.
+const v3_parallel_cc_memory_per_source_byte = u64(40)
+const v3_parallel_cc_min_job_memory = u64(192 * 1024 * 1024)
+// Only half of the free memory is spent on C compilers; the rest stays available
+// for the compiler process itself, which still holds its arenas.
+const v3_parallel_cc_memory_share = u64(2)
+// A split unit re-parses the shared declaration header, so a small program would
+// pay for the extra headers instead of saving anything. A program with this many
+// AST nodes takes seconds in a single translation unit.
+const v3_parallel_cc_auto_node_threshold = 400000
 const v3_parallel_cc_monolithic_define = 'v3_parallel_cc_monolithic'
 const v3_parallel_cc_monolithic_exit_code = 125
 const v3_parallel_cc_monolithic_message = 'v3 parallel C source requires monolithic regeneration'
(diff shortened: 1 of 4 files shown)


medvednikov:
Not relevant to `v self`: on Apple Silicon `cmd/tools/vself.v` passes `-cc tcc`, and with tcc the whole self-host is ~950 ms, of which tcc is 293 ms — the split would only add header re-parsing there. On Linux x86_64 `v self` uses the default `cc` but keeps the object cache, which this change deliberately leaves alone. So this only affects `-nocache`/explicit-`cc`/cross builds. Moving to draft; the real self-host time is in V's own phases (660 ms of the 950 ms) and that is where I am going next.

medvednikov:
Split up: the parts that do not change what a build does by default are in #28414 (separating the module-cache symbol rewriting from `cache_split`, which takes the cgen stage of a `-parallel-cc` build from 1910 ms to 783 ms), and the test-binary fix landed as #28413.

What is left here is making the split the default and sizing its job count from free memory. Per your note that is not worth it for `v self`, which builds with tcc, so closing rather than keeping it open.

PR #28415, opened 2026-09-06
v3: fix calling fn values stored in array and map containers

medvednikov:
## Problem

This program fails to build with V3:

```v
module main

struct StructThatHaveFnArray {
mut:
	fns []fn ()
}

fn returnnothing() {
	println('called')
}

const example_repeat_count = 16

fn main() {
	mut functions := []fn ()
	for i in 0 .. example_repeat_count {
		functions << returnnothing
	}
	strct := StructThatHaveFnArray{
		fns: functions
	}
	for i := 0; i < strct.fns.len; i++ {
		strct.fns[i]()
	}
}
```

```
src.c:3394:4: error: call to undeclared function 'StructThatHaveFnArray__fns'
 3394 |    StructThatHaveFnArray__fns(strct);
```

(The driver's silent V1 fallback masks this behind an unrelated
`expression evaluated but not used`; `-new-compiler` shows the real error.)

## Cause

`recv.field[i]()` calls the function value stored at index `i`, but the brackets
parse into the same `.index` node as an explicit generic instantiation
`name[T]()`, and both the checker and the C backend assumed the generic reading.

Three distinct defects fall out of that:

1. `check_call` validated every index argument of an `.index` callee as a generic
   *type* argument, so `r.handlers[idx]()` failed with ``unknown type `idx` ``.
   Single-letter index names slipped through only because of the
   `type_name.len > 1` filter — which is why the report above reaches codegen at
   all.
2. `is_explicit_generic_method_call_selector` dropped the index and emitted a
   method call, producing the `Recv__field` C error above.
3. Independently: `arr << fn_value` lowers to `array_push(&arr, &tmp)`, but the
   `&`-of-a-function-value collapse rule in `gen_expr` stripped the address, so
   `array_push` copied the element bytes out of the function's own code. Even the
   form that parsed correctly (`functions[i]()` on a local) compiled cleanly and
   then died with SIGBUS at runtime.

## Fix

- `vlib/v3/types/checker_tail.v`: new `call_callee_indexes_container()`;
  `check_call` skips generic type-argument validation when the indexed expression
  is an array, fixed array or map value. A bare ident that does not resolve to a
  value (`apply[int](v)`) still takes the generic path.
- `vlib/v3/gen/c/fn.v`: `is_explicit_generic_method_call_selector()` returns
  false for a selector of container type, so the index is no longer dropped.
- `vlib/v3/gen/c/fn.v`: `gen_generated_storage_pointer_arg()` writes the `&`
  itself rather than delegating to `gen_expr`, which collapses `&fn_value`.

## Tests

New `vlib/v3/tests/fn_value_container_index_call_codegen_test.v` covers the
container-field call, the local push-then-call (including a multi-character index
name), and an explicit generic method call to guard against the new checks
matching too broadly.

### DIFF
--- a/vlib/v3/gen/c/fn.v
+++ b/vlib/v3/gen/c/fn.v
@@ -2314,10 +2314,21 @@ fn (g &FlatGen) concrete_generic_method_name_from_call_receiver(node flat.Node, 
 	return none
 }
 
+// cgen_type_is_indexable_container reports whether `typ` is a container value that
+// `x[i]` indexes into. A selector of such a type cannot be a generic method name.
+fn cgen_type_is_indexable_container(typ types.Type) bool {
+	mut cur := types.unalias_type(typ)
+	for cur is types.Pointer {
+		cur = types.unalias_type(cur.base_type)
+	}
+	return cur is types.Array || cur is types.ArrayFixed || cur is types.Map
+}
+
 fn (g &FlatGen) is_explicit_generic_method_call_selector(fn_node &flat.Node, resolved_target_name string, target_name string) bool {
 	if fn_node.kind != .index || fn_node.children_count < 2 {
 		return false
 	}
+	selector_id := g.a.child(fn_node, 0)
 	selector := g.a.child_node(fn_node, 0)
 	if selector.kind != .selector || selector.children_count == 0 {
 		return false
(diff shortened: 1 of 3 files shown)


medvednikov:
@codex review

PR #28416, opened 2026-09-06
v3: stop expanding an `it` interpolation encoded in a string literal

medvednikov:
### Reproducer

```v
fn main() {
	tags := ['a', 'b']
	println('p ${tags.map('--\${it}').join(',')}')
}
```

On `master` this makes V3 give up and hand the program to the stable compiler:

```
main.c:3201:66: error: use of undeclared identifier 'it'
 3201 |   string __map_val_3 = string__plus(_v3_lit_2_9afb1200c5c4a565, it);
note: V3 could not build this program, so V used the stable compiler instead.
```

The fallback is silent, so a program hits it with no indication of why beyond the
one-line note. I found it because a 400k-line project stopped building with V3
after an unrelated change; bisecting was misleading because whether V3 declines
looks non-monotonic from the outside — restoring one file flips it to success and
restoring two more flips it back — since it only depends on whether the affected
function survives dead-code elimination.

### Cause

The parser resolves the `\$` escape before building the AST, so the `map` body
arrives at the transform as a `.string_literal` whose text is `--${it}` rather
than as parsed interpolation parts:

```
MAPDBG source kind=string_literal value=`--${it}` children=0
```

`simple_nested_string_interpolation` then expands that text into `'--' + it`. But
`it` is owned by the enclosing `map`, and `lower_array_map_call` has already
renamed the `it` of the *parsed* body to the loop temporary via
`substitute_ident`. The `it` synthesized out of the literal text never passed
through that rename, so nothing declares it.

A genuine nested interpolation of `it` (`'--${it}'`, no backslash) reaches the
transform as parsed parts and already works, so an `it` still encoded in literal
text at this point can never have a binding to rename.

### Fix

Leave those spans unexpanded. This also matches what the stable compiler prints
for the escaped form:

| | stable | V3 before | V3 after |
| --- | --- | --- | --- |
| `'p ${tags.map('--\${it}').join(',')}'` | `p --${it},--${it}` | *build fails* | `p --${it},--${it}` |
| `'p ${tags.map('--${it}').join(',')}'` | `p --a,--b` | `p --a,--b` | `p --a,--b` |
| `'p ${tags.map('--' + it).join(',')}'` | `p --a,--b` | `p --a,--b` | `p --a,--b` |
| `'L ${'a ${b}'}'` | `L a B` | `L a B` | `L a B` |

### Not fixed here

V3 still interpolates `'\${x}'` nested inside an interpolation where the stable
compiler prints it literally — same root cause (the escape is resolved before the
AST exists), but for a name that *is* in scope it silently produces a different
string instead of failing to build:

```v
x := 'X'
println('p ${['a'].map('--\${x}').join(',')}')  // stable: p --${x}   V3: p --X
```

I tried fixing that root cause by preserving the escape through the parser and
resolving it at materialization. It fixes all the cases, but it corrupts string
constants while V3 compiles V3 — the resulting compiler mis-handles `'p\\q'` — so
I have left it out rather than ship a self-host regression. It wants its own
change.

### Testing

- `vlib/v3/tests/nested_escaped_map_it_interp_test.v` added; it fails on `master`
  with `use of undeclared identifier 'it'` and passes with this change.
- V3 self-host (`v3.v`) and `cmd/v` both still build with the patched compiler,
  and a standalone V3 built by it still compiles `'bs ${'p\\q'}'` correctly.
- `generic_struct_str_string_interp_codegen_test`,
  `generic_sum_str_string_interp_codegen_test` and `string_interp_char_format_test`
  pass. `string_interpolation_stringify_codegen_test` fails identically before and
  after this change.

### DIFF
--- a/vlib/v3/tests/nested_escaped_map_it_interp_test.v
+++ b/vlib/v3/tests/nested_escaped_map_it_interp_test.v
@@ -1,0 +1,48 @@
+import os
+
+const vexe = @VEXE
+const tests_dir = os.dir(@FILE)
+const v3_dir = os.dir(tests_dir)
+const vlib_dir = os.dir(v3_dir)
+const v3_src = os.join_path(v3_dir, 'v3.v')
+
+// Written as a raw string so the escapes reach the compiler under test as typed.
+const nested_it_program = r"fn main() {
+	b := 'B'
+	tags := ['a', 'b']
+	println('1:' + 'p ${tags.map('--\${it}').join(',')}')
+	println('2:' + 'p ${tags.map('--${it}').join(',')}')
+	println('3:' + 'p ${tags.map('--' + it).join(',')}')
+	println('4:' + 'p ${tags.filter(it != 'b').map('<\${it}>').join(',')}')
+	println('5:' + 'L ${'a ${b}'}')
+}
+"
+
+// A `${it}` that reaches the transform still encoded in a string literal's text has
+// no binding to rename: `it` belongs to the enclosing `map`/`filter`, whose lowering
+// renames it while walking the parsed body. Expanding it anyway synthesized an `it`
+// that no declaration reached, and the generated C failed to build with
+// `use of undeclared identifier 'it'`, which made V fall back to the stable compiler.
+fn test_escaped_it_interpolation_in_map_body_compiles() {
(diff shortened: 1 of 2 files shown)


medvednikov:
@codex review

PR #28420, opened 2026-09-06
Linux lean cmd/v build generates quadratic-memory parser code (make bootstrap OOM since #28378)

quaesitor-scientiam:
## Symptom

`make` on Linux fails during bootstrap at `GNUmakefile:219`:

```
error: v3 compiler memory usage reached 10180 MiB RSS during compilation (limit: 10176 MiB)
```

With `-no-memory-limit` it runs on to ~36 GB and `could not allocate`. **Linux + `-gc none` only** — macOS and Windows are fine.

## What actually happens

The self-hosted V3 compiler generates **quadratic-memory parser code** for its own frontend when `cmd/v` is built *lean* (Linux path, `v.builder` not in the module graph):

- `parse .v` for `cmd/v` takes **20.5 s** and reaches **10 GB RSS** before the checker starts (~305 files, 340k lines). Normal is <1 s / a few hundred MB.
- It compounds across the `v1 -> v2 -> v` bootstrap generations until it trips the ceiling.

It is a **codegen** effect, not a source bug: `cmd/v` compiled by a pre-regression compiler runs fine (~2-3 GB); the same source compiled by a current compiler is the 35 GB one. Deterministic 2-generation self-host probe (`-nocache VJOBS=1`, isolated cache):

| build | peak RSS compiling `cmd/v` |
| --- | --- |
| parent of #28378 (`c918d7ce82`) | **3 GB** |
| #28378 (`edf824295b`) .. HEAD | **35 GB** |
| HEAD **`-d v1_fallback`** (re-adds the `v.builder` import) | **3 GB** |

DCE / `markused` reachability, with `v.builder` **absent**, turns a shared parser code path into a quadratic one. #28378 (`vself: make macOS and Linux self-builds V3-only`) made it load-bearing on Linux by dropping the `v.builder` import via `$if !macos && !linux`; the underlying miscompilation predates it.

## Workaround (no patch)

```
make VFLAGS="-d v1_fallback"
```

Verified: produces a working `v` on Linux/WSL.

## This branch

Reverts the 4 dispatch files to pre-#28378 (routes Linux `cmd/v` back through V1). Works, but conflicts with #28417 and is against the "cmd/v self-builds deliberately remain V3-only" intent — kept only as a reference stopgap. The real fix belongs in the lean-build parser codegen / `markused` reachability.

### DIFF
--- a/cmd/v/macos_v3_args.c.v
+++ b/cmd/v/macos_v3_args.c.v
@@ -1,5 +1,6 @@
 module main
 
+import os
 import v.pref
 
 const macos_v3_compat_c99_flag = '-macos-v3-compat-c99'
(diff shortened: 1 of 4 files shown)


quaesitor-scientiam:
Filed as #28423 with the full diagnosis and repro. This PR is a dispatch-layer stopgap only; the real fix belongs in the lean-build parser codegen / markused reachability described there. Recommended workaround for anyone hitting this: `make VFLAGS="-d v1_fallback"`.

PR #28421, opened 2026-09-06
cgen: avoid generating keepalive helpers for pointer-free C types

maaamahAhh:
## Problem

Compiling code that uses `term.ui` with `-prod` on Windows fails under GCC:

```
error: conflicting types for '__v_boehm_collect_keepalive_..._union_uChar'; have 'int(union uChar *, void **, int)'
static inline int __v_boehm_collect_keepalive_..._union_uChar(union uChar* it, voidptr* out, int idx) {
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

## Cause

In `cgen`, `contains_ptr` conservatively returns `true` for all non-V types (`sym.language != .v`) to ensure heap allocations of C structs (which frequently only declare a subset of fields in V) are not marked `_noscan` (`GC_MALLOC_ATOMIC`).

However, `boehm_collect_keep_alive_helper_name` and `type_needs_deep_scope_gc_pin` also checked `contains_ptr(typ)`. Because `contains_ptr` returns `true` for all C types, V generated empty keepalive helpers (`return idx;`) and forward prototypes for C types whose declared fields contain no pointers (such as `C.KEY_EVENT_RECORD` and `C.uChar`).

Following #27351, C unions are emitted with a `union <name>` tag. When a C union is anonymous inside an outer struct (such as `uChar` inside `_KEY_EVENT_RECORD` in Windows SDK `wincon.h`), no file-scoped tag `union uChar` exists in C. Under C99 prototype scope rules, GCC treats parameter `union uChar *it` in the forward prototype as prototype-scoped, conflicting with the later definition.

## Fix

1. Keep `contains_ptr` untouched so conservative GC heap allocation scanning policy (`_noscan` / `GC_MALLOC_ATOMIC`) is completely preserved.
2. In `cgen.v`, added zero-allocation `c_type_has_ptr(typ)` to check whether a C struct/union declares pointer fields (inspecting direct pointers, embeds, nested structs, and fixed arrays).
3. In `boehm_collect_keep_alive_helper_name` and `type_needs_deep_scope_gc_pin`, avoid emitting keepalive helpers and deep scope pins for C types that declare no pointer fields.

## Tests

- Added regression test `vlib/v/gen/c/testdata/boehm_keepalive_c_anon_union.vv` (with `.h` and `.out`) with `// vtest vflags: -prod -gc boehm_full_opt`. (Verified to reproduce the GCC prototype-scope failure on unmodified master, and pass cleanly with this fix).
- Verified `boehm_keepalive_c_union_tag.vv` (from #27351), `vlib/v/tests/unions/`, and `vlib/v/tests/c_structs/` all pass.
- Verified real-world Windows `term.ui` applications compile with `-prod -cc gcc` without errors.


medvednikov (commented):
I think this weakens the GC-correctness guarantee for partially-declared C structs.

The PR correctly notes that `contains_ptr` is intentionally conservative for non-V types because a V declaration of a C struct may list only a subset of the real C fields. Today, even when the generated keepalive helper finds no declared pointer fields and just returns `idx`, `scope_gc_pin_pregen` still takes the address of the aggregate and emits `GC_reachable_here(tmp)` after the call. That keeps the whole C object live across the call, so Boehm can conservatively see pointer-valued fields that exist in the real C layout but are not declared to V.

With `type_needs_deep_scope_gc_pin` changed to `return g.c_type_has_ptr(resolved_typ)`, a C struct whose V declaration has no pointer fields is skipped entirely. If the actual C struct has an omitted pointer field, the optimizer is now free to end/reuse that aggregate's stack storage before a call that can collect, which can drop the hidden pointer root.

Could we keep `type_needs_deep_scope_gc_pin` conservative for non-V structs and solve the anonymous-union prototype problem only in helper generation (or add a whole-object pin path that does not require a typed helper)? A regression test where the C header contains a hidden `void *` field that is intentionally omitted from the V `C.foo` declaration would help guard this case.

medvednikov:
v1 is no longer used, see if it's an issue in vlib/v3

antono3:
I checked the V3 side and tested a narrower V1/fallback fix.

V3 has no analogous keepalive/pinning path at present: it explicitly rejects both `-gc boehm` (`currently supports only -gc none`) and `-d gcboehm` (`v3 programs must not use a garbage collector`). So this issue currently applies only when compatibility routing reaches the old C generator.

For that path, preserving conservative handling of non-V structs while avoiding the problematic typed helper works: when helper generation returns empty, emit only an untyped `voidptr tmp = &aggregate` before the call and `GC_reachable_here(tmp)` afterward. I also added a hidden `void *` field to the C test struct, intentionally omitted from its V declaration.

Focused Linux and Windows GCC checks both pass: the anonymous-union fixture compiles/runs, no `TestAnonEvent* ... keepalive` helper is generated, and the untyped reachability barrier remains.

Patch commit: https://github.com/antono3/v/commit/1e675018e899052de04261d0545ae65353dd393b
Actions: https://github.com/antono3/v/actions/runs/34102840972

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

medvednikov:
@codex review

PR #28422, opened 2026-09-06
v3: restore compatibility routing for cross builds and tools

medvednikov:
Follow-up to #28417.

The external V1 retry is working, but three CI paths exposed compatibility modes that should never enter V3 first:

- portable/cross C generation (`-os cross`) can reach both V1 import selectors while compiling `cmd/v`, causing duplicate `v.builder` imports;
- non-host C targets such as the macOS Metal test's `-os windows` probe are still compatibility-compiler modes;
- `cmd/tools` bootstraps such as VPM can currently be miscompiled by V3 (the failing macOS cross job produced a corrupt `VCS.str()` result even though the parsed value was `git`).

This change:

- explicitly excludes `v1_fallback` from the non-macOS/Linux V1 import branch;
- sends portable VC output, non-host C targets, and V's internal `cmd/tools` builds directly to the external `v1_fallback` binary;
- keeps explicit `-new-compiler` and FastC strict on V3;
- leaves native `cmd/v` self-builds V3-only;
- adds routing regression tests.

This is intentionally a routing fix, not a VPM or multiwindow workaround.


medvednikov:
@codex review

PR #28441, opened 2026-09-08
GNUmakefile: bootstrap Windows from vc/v_win.c, not vc/v.c

bscuron:
This PR fixes `make` on Windows (MSYS/MinGW), which currently fails to bootstrap.

**Wrong VC snapshot on Windows.** `GNUmakefile` hardcoded `VCFILE` to `v.c` for every platform. `vc/v.c` is produced by `./v -o vc/v.c -cross cmd/v` on Linux CI (see `gen_vc_ci.yml`), so its file selection still resolves to the host's *nix sources (`os_nix.c.v`, `termios_linux.c.v`, `closure_nix.c.v`, ...) — `-cross` only changes how some comptime conditions are lowered to C, not which OS-suffixed files get included. Compiling that snapshot with MinGW gcc fails outright, since it `#include`s POSIX-only headers (`sys/mman.h`, `termios.h`, `sys/ptrace.h`, `sys/wait.h`, `sys/utsname.h`, `sys/statvfs.h`) and calls POSIX-only APIs (`mmap`/`mprotect`, `sysconf`, `uname`, `flock`, ...) that MinGW doesn't provide.

The dedicated Windows snapshot is `vc/v_win.c`, generated with `-os windows`, and is already what `make.bat`, `vself.v`, and `vgit.v` use to bootstrap on Windows. This PR overrides `VCFILE` to `v_win.c` when `WIN32` is detected in `GNUmakefile`, matching those other tools.

**Unquoted recursive `$(MAKE)` calls.** Also quotes the recursive `$(MAKE)` invocations throughout the file, preventing word-splitting when the checkout path or `MAKE` contains spaces, matching how other paths in the file are already quoted.

Verified: `gcc -std=c99 -municode -w -o v1.exe vc/v_win.c -lws2_32` compiles cleanly, and `make` now completes the full v1 -> v2 -> v bootstrap and produces a working `v.exe` under MSYS/MinGW.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -218,8 +218,8 @@ ifdef WIN32
 	$(RM) v2$(EXE_EXT)
 else
 ifdef LEGACY
-	$(MAKE) -C $(TMPLEGACY) CPPFLAGS='$(CPPFLAGS)' CFLAGS='$(CFLAGS)' LDFLAGS='$(LDFLAGS)'
-	$(MAKE) -C $(TMPLEGACY) PREFIX=$(realpath $(LEGACYLIBS)) CPPFLAGS='$(CPPFLAGS)' CFLAGS='$(CFLAGS)' LDFLAGS='$(LDFLAGS)' install
+	'$(MAKE)' -C $(TMPLEGACY) CPPFLAGS='$(CPPFLAGS)' CFLAGS='$(CFLAGS)' LDFLAGS='$(LDFLAGS)'
+	'$(MAKE)' -C $(TMPLEGACY) PREFIX=$(realpath $(LEGACYLIBS)) CPPFLAGS='$(CPPFLAGS)' CFLAGS='$(CFLAGS)' LDFLAGS='$(LDFLAGS)' install
 	rm -rf $(TMPLEGACY)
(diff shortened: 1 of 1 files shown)


medvednikov:
Reviewed PR #28441, ignoring CI as requested. No blocking correctness issues found — LGTM.

The core fix is consistent with V’s bootstrap setup: the Makefile now selects v_win.c for MSYS/MinGW, and the Windows bootstrap recipe already compiles $(VC)/$(VCFILE). The VC-generation workflow runs on Ubuntu and explicitly generates v.c with -cross while generating v_win.c with -os windows -cc msvc, so the PR’s explanation for using the Windows-specific snapshot is accurate. Other V bootstrap code also already chooses v_win.c on Windows.

I also checked the separate $(MAKE) quoting change. GNU Make documents MAKE as the filename used to invoke make and recommends using $(MAKE) for recursive invocations. I reproduced it locally with GNU Make 4.4.1: '$(MAKE)' still receives recursive-make handling and, unlike the old unquoted form, successfully handles a make executable whose path contains spaces. So I don’t see a regression there.

The only non-blocking consideration is scope: the recursive-make quoting cleanup is logically separate from the Windows VC snapshot fix, so it could have been a separate PR/commit for easier review. It looks correct as-is, though.

Verdict: approve.

PR #28449, opened 2026-09-08
doc: document raw assembly kernel patterns

quaesitor-scientiam:
## Summary

- Document %= local labels, numeric labels, operand constraints, and clobbers for raw GNU inline assembly.
- Add an amd64 assembly slow test covering loop labels and a scalar Poly1305 kernel against the RFC 8439 vector.
- Keep the kernel as a test fixture; no production crypto API or compiler backend changes.

## Validation

- Assembly slow tests with GCC: 8 passed, 10 skipped.
- Targeted assembly test with Clang: passed.
- Targeted MSVC run: skipped as expected because GNU raw assembly is unsupported there.
- TCC run: TCC fallback to GCC, test passed.
- new fmt and git diff --check passed.

### DIFF
--- a/doc/docs.md
+++ b/doc/docs.md
@@ -9120,6 +9120,43 @@ asm amd64 raw {
 assert value == 42
 ```
 
+Raw templates are the right level for hand-written kernels that need GNU assembler features such
+as local labels or explicit operand modifiers. A label made with `%=` gets a unique numeric suffix
+for each inline-assembly statement, so prefer names such as `.Lloop%=` for loops in reusable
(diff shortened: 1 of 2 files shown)


medvednikov (commented):
Reviewed the code/documentation changes only; CI status intentionally not considered. I did not find a concrete correctness issue. The `%=` and numeric-local-label guidance matches GNU extended-asm behavior, the modified pointer/count operands are correctly read-write, and the kernel fixtures declare their scratch registers plus `memory`/`cc` clobbers.

PR #28453, opened 2026-09-09
test: cover SIMD inline assembly kernels

quaesitor-scientiam:
Summary:
- Document SIMD operand binding, vector clobbers, Win64 XMM preservation, and optional CPU feature guards.
- Add amd64 SIMD register-pressure, ChaCha20 SSE2, and PCLMUL GHASH fixtures.
- Add arm64 NEON ChaCha20 and PMULL GHASH cross-target fixture.

Validation:
- GCC assembly suite: 11 passed, 11 skipped.
- Clang assembly suite: 11 passed, 11 skipped.
- TCC targeted fixtures: 3 passed via GCC fallback.
- arm64 cross code generation: passed.
- MSVC: new GNU fixtures skipped; existing naked_attr_test.amd64.v remains a pre-existing failure because MSVC x64 does not support inline assembly.
- check-md was attempted but the environment denied writes to the default VMODULES cache for unrelated mbedtls/SQLite examples.

### DIFF
--- a/vlib/v/slow_tests/assembly/asm_raw_simd_crypto_test.arm64.v
+++ b/vlib/v/slow_tests/assembly/asm_raw_simd_crypto_test.arm64.v
@@ -1,0 +1,255 @@
+import encoding.binary
+import encoding.hex
+import os
+
+// raw_chacha20_block_neon keeps one ChaCha state row in each NEON register.
+fn raw_chacha20_block_neon(state &[16]u32) [16]u32 {
+	mut output := [16]u32{}
+	// The raw block writes the fixed array through its first element.
+	output_ptr := unsafe { &output[0] }
+	asm arm64 raw {
+		"ldr q0, [%[state]]\n\t"
+		"ldr q1, [%[state], #16]\n\t"
+		"ldr q2, [%[state], #32]\n\t"
+		"ldr q3, [%[state], #48]\n\t"
+		"mov v8.16b, v0.16b\n\t"
+		"mov v9.16b, v1.16b\n\t"
+		"mov v10.16b, v2.16b\n\t"
+		"mov v11.16b, v3.16b\n\t"
+		".rept 10\n\t"
+
+		// Column quarter round.
+		"add v0.4s, v0.4s, v1.4s\n\t"
+		"eor v3.16b, v3.16b, v0.16b\n\t"
+		"mov v4.16b, v3.16b\n\t"
+		"shl v3.4s, v3.4s, #16\n\t"
+		"ushr v4.4s, v4.4s, #16\n\t"
+		"orr v3.16b, v3.16b, v4.16b\n\t"
+		"add v2.4s, v2.4s, v3.4s\n\t"
+		"eor v1.16b, v1.16b, v2.16b\n\t"
+		"mov v4.16b, v1.16b\n\t	"
+		"shl v1.4s, v1.4s, #12\n\t"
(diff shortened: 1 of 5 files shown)


medvednikov (commented):
Reviewed the code changes only; CI status intentionally not considered. I found one portability problem in the PMULL runtime guard.

vlib/v/slow_tests/assembly/asm_raw_simd_crypto_test.arm64.v
```diff
@@ -0,0 +1,255 @@
...
+	}
+	return result
+}
+
+fn can_run_pmull_test() bool {
+	$if linux {
+		cpuinfo := os.read_file('/proc/cpuinfo') or { return false }
+		return cpuinfo.contains('pmull')
+	} $else {
+		return true
```
medvednikov: `pmull` is an optional AArch64 crypto-extension instruction, but every non-Linux arm64 host is treated as supporting it here. That can turn this test into an illegal-instruction crash on a valid arm64 machine without PMULL support (for example a generic FreeBSD/Windows ARM64 target). The docs added by this PR explicitly say optional-feature tests must check the host feature. Please either add platform-specific feature detection for the supported non-Linux arm64 hosts or skip the PMULL test when support cannot be established.

JalonSolov:
Local AI Findings

**[P1]** New tests do not compile. The test functions call `hex.decode(...)!` without returning !, causing V to reject result propagation. This affects `asm_raw_chacha20_test.amd64.v`, `asm_raw_ghash_clmul_test.amd64.v`, and `asm_raw_simd_crypto_test.arm64.v`. Declare those test functions as `fn ...() !` or handle decoding errors explicitly.

**[P1]** ARM64 raw assembly is not excluded for MSVC. The fixture uses GNU-style `asm arm64 raw`, but unlike the x86 fixtures it has no `// vtest build: !msvc` guard. The documentation states that MSVC does not support GNU inline assembly. Add the guard or provide a compatible Windows implementation.

quaesitor-scientiam:
Both addressed in 255ebb3b16.

**P1 `!` propagation** — the four test functions that call `hex.decode(...)!` are now declared `fn ... () !` (`test_raw_chacha20_simd_block_rfc8439`, `test_raw_ghash_clmul_nist_vector`, `test_raw_chacha20_neon_vector`, `test_raw_ghash_pmull_vector`). Background: a `!` inside a void `fn test_x()` is accepted for `_test.v` but not for the arch-qualified `_test.amd64.v`/`_test.arm64.v` spellings, because the V3 checker's private test-file predicate is narrower than `pref`'s (#28736, fix in #28737). Rather than wait for that, the tests now return a Result, which works with either predicate. Verified on master 2b99da47ea, Windows amd64: the pre-change ghash file fails with `to propagate the Result call, `test_raw_ghash_clmul_nist_vector` must return a Result`; after the change both amd64 tests pass under the bundled tcc and under `-cc gcc`, and a probe `fn test_e() ! { hex.decode('zz')! }` in a `.amd64.v` file is reported as a failure by the runner (so errors are not swallowed). The arm64 file passes `-os macos -arch arm64 -check` and `-os linux -arch arm64 -check`; I have no arm64 host here to run it.

**P1 MSVC guard** — `asm_raw_simd_crypto_test.arm64.v` now starts with `// vtest build: !msvc`, like the x86 raw fixtures. Note that on current master the runner honours no `// vtest build:` facts at all (#28798, fix in #28801), so the guard only takes effect once that lands.

Issue #28447, opened 2026-09-08
asm: roadmap for hand-written crypto kernels — in-asm control flow, SIMD-scale blocks, whole-function asm


quaesitor-scientiam:
## Context

`asm … raw` / `asm … intel` landed via #28358 (V1) and #28365 (v3), with AArch64
register names (#28375) and an Intel operand-width guard (#28369). A verbatim port
of OpenSSL's scalar `poly1305_blocks` inner sequence into `asm amd64 raw {}` now
assembles and produces the correct RFC 8439 tag under gcc / clang / tcc on Linux
(SysV) and Windows (Win64) — https://github.com/vlang/v/pull/28358#issuecomment-5571978682

That covers **scalar bignum** kernels (Poly1305, X25519 field arithmetic,
Montgomery multiply). Three gaps remain before inline asm can express the rest of
what a pure-V `crypto` acceleration layer needs. This issue tracks them as
separate deliverables.

---

### 1. Control flow inside a `raw` block (labels + loops)

Most real kernels loop internally over the message / limbs. Today the loop must
live in V, calling the `asm` block once per block — a function call + full operand
reload per iteration.

- Local labels usable within a `raw` template without colliding across multiple
  instantiations or inlined copies (GNU `%=`-style unique suffix, or a documented
  numeric `1:` / `1b` / `1f` convention).
- A documented, tested bound-pointer + length operand pattern for a self-contained
  loop.
- **Acceptance:** a `poly1305_blocks` looping over an arbitrary-length buffer
  entirely inside one `asm amd64 raw` block, passing RFC 8439 vectors under
  gcc/clang/tcc on x86-64 SysV + Win64.

### 2. SIMD-scale blocks (XMM/YMM/ZMM pressure)

AES-NI, GHASH-CLMUL, ChaCha20 and SHA-NI use most of the vector register file plus
several GPRs.

- Confirm the register allocator + clobber model behaves when a `raw` block
  clobbers 10+ `xmm`/`ymm` registers and the surrounding V function also uses
  floats.
- Confirm `%[name]` operand binding for vector-typed inputs
  (pointer-to-`[16]u8`, `[4]u32`, …) and alignment expectations.
- Windows: XMM6–15 are callee-saved under Win64 — verify the compilers
  save/restore around the block, or document that the kernel must.
- **Acceptance:** a ChaCha20 block function and a GHASH (CLMUL) multiply as `raw`
[...]

PR #28455, opened 2026-09-09
test: document whole-function assembly kernels (#28447)

quaesitor-scientiam:
Part 3 of #28447: whole-function / naked assembly support.

## Summary

- Document the external `.S` plus `.o` linkage path for kernels that need their own frame or
  `call` instructions.
- Add a whole-function SHA-256 compression kernel in amd64 assembly.
- Call the kernel from pure V through an explicit `@[c_extern]` C ABI declaration and verify the
  SHA-256 `abc` vector.

## Validation

- GCC full assembly suite: 8 passed, 10 skipped.
- GCC external-kernel fixture: passed on Win64.
- Clang external-kernel fixture: passed on Win64.
- SysV amd64 form: assembled with Clang Linux target.
- MASM `ml64` object plus MSVC pure-V caller: passed locally.
- Repository MSVC fixture: skipped because the checked-in source is GNU `.S`; MSVC users can
  provide a MASM-compatible `.obj` as documented.
- TCC: fixture skipped because its assembler does not support the `.macro` directives used by
  this kernel.
- `check-md doc/docs.md`: 300 OK, 0 warnings, 3 existing platform/environment errors (`dprintf`,
  SQLite headers, and malformed `ml` fence) plus 1 existing formatting error.

### DIFF
--- a/vlib/v/slow_tests/assembly/util/v_sha256_block.S
+++ b/vlib/v/slow_tests/assembly/util/v_sha256_block.S
@@ -1,0 +1,227 @@
+.text
+.globl v_sha256_block
+
+// void v_sha256_block(uint32_t state[8], const uint8_t block[64])
+// The argument registers are selected for the System V or Win64 amd64 ABI.
+#ifdef _WIN64
+#define STATE_PTR %rcx
+#define BLOCK_PTR %rdx
+#else
+#define STATE_PTR %rdi
+#define BLOCK_PTR %rsi
+#endif
+
+v_sha256_block:
+	pushq %rbx
+	pushq %r12
+	pushq %r13
+	pushq %r14
+	pushq %r15
+	subq $272, %rsp
+	movq STATE_PTR, 256(%rsp)
+
+	.macro LOAD_WORD offset
+		movl \offset(BLOCK_PTR), %eax
+		bswapl %eax
+		movl %eax, \offset(%rsp)
+	.endm
+
+	LOAD_WORD 0
+	LOAD_WORD 4
+	LOAD_WORD 8
+	LOAD_WORD 12
+	LOAD_WORD 16
+	LOAD_WORD 20
+	LOAD_WORD 24
+	LOAD_WORD 28
+	LOAD_WORD 32
+	LOAD_WORD 36
+	LOAD_WORD 40
+	LOAD_WORD 44
+	LOAD_WORD 48
+	LOAD_WORD 52
+	LOAD_WORD 56
+	LOAD_WORD 60
+
+	// Expand W[16..63] in the local stack frame.
+	movl $64, %edi
+.Lschedule:
+	movl -8(%rsp,%rdi), %eax
+	movl %eax, %ecx
+	movl %eax, %edx
+	rorl $17, %eax
+	rorl $19, %ecx
+	shrl $10, %edx
+	xorl %ecx, %eax
+	xorl %edx, %eax
+
+	movl -60(%rsp,%rdi), %r8d
+	movl %r8d, %r9d
+	movl %r8d, %r10d
+	rorl $7, %r8d
+	rorl $18, %r9d
+	shrl $3, %r10d
+	xorl %r9d, %r8d
+	xorl %r10d, %r8d
+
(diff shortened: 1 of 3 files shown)


medvednikov (commented):
Reviewed the code changes only; CI status intentionally not considered. I found one Win64 ABI violation in the new assembly fixture.

vlib/v/slow_tests/assembly/util/v_sha256_block.S
```diff
@@ -0,0 +1,227 @@
...
+	LOAD_WORD 40
+	LOAD_WORD 44
+	LOAD_WORD 48
+	LOAD_WORD 52
+	LOAD_WORD 56
+	LOAD_WORD 60
+
+	// Expand W[16..63] in the local stack frame.
+	movl $64, %edi
+.Lschedule:
```
medvednikov, line 52: On Win64, `%rdi` is nonvolatile/callee-saved, but this function overwrites it here as the schedule index (and later with the state pointer) without saving/restoring it. The prologue saves `rbx` and `r12`–`r15`, but not `rdi`, so a C/V caller that keeps a live value in RDI across the call can be corrupted even though the SHA vector test happens to pass. Please either preserve/restore RDI under `_WIN64` or use a Win64-volatile register for the schedule/state pointer.

quaesitor-scientiam: Good catch, thanks. Fixed in 3790453fa2: `%%rdi` is now saved/restored under `_WIN64` around the frame setup (all `%%rsp`-relative offsets are relative to the post-`subq` pointer, so they are unchanged). `%%rsi` is never touched on Win64 (`BLOCK_PTR` is `%%rdx` there).

Also added `test_external_sha256_block_preserves_rdi_on_win64`, which calls the fixture with a sentinel in `rdi` from a raw asm block and asserts it survives — it fails on the previous fixture and passes now (verified with gcc on Windows).

JalonSolov:
Local AI Findings

**P2** `vlib/v/slow_tests/assembly/asm_external_sha256_test.amd64.v:1` uses `amd64` in the `vtest build` condition. vtest does not recognize that architecture predicate; the `.amd64.v` filename already handles architecture selection. The new test is therefore skipped on Linux amd64, so neither the SHA-256 check nor the Win64 regression check runs. Remove `amd64 &&` from the condition.

quaesitor-scientiam:
Done in a16dc76910: the constraint is now `// vtest build: !msvc && !tinyc`, matching `asm_no_tcc_test.amd64.v`; the `.amd64.v` suffix already restricts the file to amd64 hosts via `pref.get_test_v_files_from_dir_for_target`.

One note on the cause, since it is wider than this test: `amd64` is a valid fact name for the driver (`v3_test_build_fact_name` in `vlib/v/driver/driver.v` adds `target.arch` to the facts), but the directory runner (`testing.setup_build_environment` → `pref.get_build_facts_and_defines`) reads facts only from the `VBUILD_FACTS` environment variable, and nothing has exported that since 2a7447b5e5 removed the launcher's `setup_vbuild_env_vars`. On current master (2b99da47ea, Windows amd64, `v test <dir>`) every positive fact is therefore false and every negated one true: a probe set of seven one-line tests skipped `amd64`, `windows`, `tinyc` and `gcc` and ran `!amd64`, `!windows`, `!tinyc`; `v test vlib/context` skips 5 of its 7 tests (`amd64 || arm64`). With this PR's header fixed, `v -cc gcc test …` passes and `v -cc <tcc> test …` now builds despite `!tinyc` and fails at `v_sha256_block.S:27: unknown opcode '.macro'` — the same exposure `asm_no_tcc_test.amd64.v` already has. Happy to file that separately.

quaesitor-scientiam:
Filed the runner-facts regression as #28798 (sub-issue of #28447) with a fix in #28801: the launcher exports `VBUILD_FACTS`/`VBUILD_DEFINES` again, computed by the driver's own compiler-selection code, so `v test` and `v file_test.v` evaluate `// vtest build:` identically. With that fix, this test's `!tinyc` is honoured under the bundled tcc again.

PR #28464, opened 2026-09-09
fix(cgen): deref mut pointer params passed to C functions

aki-xavier:
## Summary

A `mut` parameter referenced by value inside a V function is one dereference deep: the current value is `*o`, not `o`. `ref_or_deref_arg_ex` already emits the deref when the expected type is a non-pointer value, but when the expected type is a pointer that matches exactly one deref of the mut parameter (e.g. `mut o voidptr` passed to `C.c_read(o voidptr, ...)`), the codegen emitted the address of the parameter slot instead of the value. The C side then dereferences unrelated memory.

## Minimal repro (before/after)

```v
module main

fn C.c_id(p voidptr) voidptr

fn thunk(mut o voidptr) voidptr {
	return C.c_id(o) // generated C.c_id(&o) before; C.c_id(*o) after
}

fn main() {
	mut o := voidptr(u64(0x1234))
	println(thunk(mut o) == voidptr(u64(0x1234))) // false before, true after
}
```

Before, the generated C was `c_read(o, ...)` with `o` typed `void**`; the C function receives the address of the parameter slot. After the fix it becomes `c_read(*o, ...)`.

## Root cause

In `ref_or_deref_arg_ex`, the branch handling "mut parameter ident passed by value" only fires when `expected_type` is a non-pointer. For a pointer expected type that equals exactly one deref of the mut param (`unwrapped_expected == arg_typ with one fewer *`), no `*` is emitted, so the `void**` parameter variable is passed directly where the C function expects `void*` (C implicitly converts `void**` to `void*`, silently). Same class of bug for `mut byteptr`/`mut charptr` and any pointer-typed mut parameter.

## Fix

Only the exact-fit case is dereferenced: `expected` is a pointer and equals one deref of the mut parameter. Identical expected types (explicit `&T` references) are still passed unchanged, preserving existing reference-passing behavior.

## Verification

- New regression test `vlib/v/tests/c_mut_voidptr_param/` fails on the old compiler and passes with the fix.
- Integration case (simu project: `cpp_impulse.v`, URDF robot attach + `SetArticulatedPoseFromJoints` + `scene->Step`) crashed with SIGBUS on the release toolchain before and runs correctly after rebuilding the compiler with the fix.
- `v test vlib/v/tests/type_voidptr_test.v` passes.

### DIFF
--- a/vlib/builtin/closure/closure_nix.c.v
+++ b/vlib/builtin/closure/closure_nix.c.v
@@ -13,7 +13,10 @@ struct ClosureMutex {
 
 @[inline]
 fn closure_mtx_ptr_platform() voidptr {
-	return unsafe { voidptr(&g_closure.closure_mtx[0]) }
+	// Explicit `ClosureMutex.` (instead of the promoted `g_closure.closure_mtx`)
+	// so the V3 C backend emits the embedded sub-struct access; the promoted
+	// path emits `g_closure.closure_mtx` which does not exist at the C level.
+	return unsafe { voidptr(&g_closure.ClosureMutex.closure_mtx[0]) }
 }
 
 @[inline]
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -7587,14 +7587,30 @@ fn (mut g Gen) ref_or_deref_arg_ex(arg ast.CallArg, expected_type_ ast.Type, lan
 		g.write('*(${g.cc_type(expected_type, false)}*)&')
 	} else if arg.expr is ast.Ident {
 		if arg.expr.obj is ast.Var {
-			if arg.expr.obj.is_arg && arg.expr.obj.is_mut && !arg.is_mut && arg_typ.is_ptr()
-				&& !expected_type.is_any_kind_of_pointer() {
+			if arg.expr.obj.is_arg && arg.expr.obj.is_mut && !arg.is_mut && arg_typ.is_ptr() {
+				// A `mut` parameter referenced by value is its dereference: one level
+				// of indirection is taken off to get the current value, in V as well.
+				// The existing branch below handles value arguments (non-pointer
+				// expected type); a pointer expected type that equals exactly one
+				// deref of the mut parameter (e.g. `mut o voidptr` passed to
+				// `C.f(o voidptr)`) must also take that level off, otherwise the C
+				// side receives the address of the parameter slot instead of the value.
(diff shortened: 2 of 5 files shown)


aki-xavier:
## Verification results

Full `v test vlib/v/tests/` (2222 tests) with the patched compiler:
- 2202 passed, 16 skipped, 4 failed.
- The 4 failures (`shared_elem_test`, `struct_field_shared_test`, `struct_shared_field_with_default_init_test`, `usecache_interface_index_symbol_test`) were verified to fail identically on a compiler built from the same tree without this patch, so they are pre-existing and unrelated to this change.

Also verified that an integration case that crashed with SIGBUS before this patch (URDF robot attach + `SetArticulatedPoseFromJoints` + `scene->Step` via a C ABI shim) runs correctly after rebuilding the compiler with the fix.

medvednikov:
Thanks, but this should be fixed in v3. v1 is no longer used.


**this exact `mut voidptr`/C-call bug does not appear to be a V3 issue.**

PR #28464 fixes the legacy C generator in `vlib/v/gen/c/fn.v`: a `mut o voidptr` parameter is represented one indirection deeper, and the legacy backend incorrectly passed the slot (`void**`) to a C function expecting the value (`void*`) instead of emitting `*o`.

V3 has separate handling for this. It explicitly tracks pointer-typed mutable parameters and understands that a source `mut p &T` has a **`T**` pointer-slot ABI**.  It also distinguishes forwarding `mut p`—where the slot itself should be passed—from reading `p` by value. The V3 code specifically documents:

> `mut p &T` is represented as `T**`; forwarding `mut p` ... must pass that slot directly, not read `*p` first.

And V3 has dedicated source-mut-pointer dereference handling when determining/emitting the semantic value.

There **is a separate V3 issue exposed in this PR**, though: the first hunk changes

```v
g_closure.closure_mtx
```

to

```v
g_closure.ClosureMutex.closure_mtx
```

because V3 incorrectly emits the promoted embedded-field access at the C level. That is unrelated to the `mut voidptr` C-call dereference bug.

So I’d classify #28464 as:

**Legacy C backend:** yes, affected by the main bug.
**V3 C backend:** likely **not affected by that bug**; it already has the appropriate pointer-slot/value distinction.
**V3:** does have the separate embedded-struct-field issue revealed by the PR.

For absolute confirmation, I’d run the PR’s tiny `C.c_mut_voidptr_id(o)` regression explicitly through V3; based on the current V3 lowering code, I’d expect it to generate the equivalent of `C.c_mut_voidptr_id(*o)`.

PR #28469, opened 2026-09-09
fix(v3): keep branch type of if-expr values with generic calls

aki-xavier:
## Summary

An `if`-expression whose branches call a **generic function** (e.g. `math.min[T]` / `math.max[T]`) gets lowered to a temporary of the wrong type in the v3 transform when used in a declaration context:

```v
module main
import math

fn main() {
	a := 1e-3 / 0.02
	x := if a > 0 { math.min(1.0, a) } else { 1.0 }
	println(x) // 0, expected 0.05
}
```

Generated C (before):

```c
int __if_val_0 = 0;
if (true) {
	__if_val_0 = math__min_T_f64(1.0, a); // 0.05 truncated to int -> 0
} else {
	__if_val_0 = 1.0;
}
int x = __if_val_0;
```

Root cause: `try_expand_if_expr_value_for_type` keeps the caller-supplied `result_type` (`int`, the default target in declaration contexts) for the if-value temporary; the merged branch type is only consulted for sum-type overrides. Non-generic calls are unaffected (`math.sqrt` branches lower correctly), which is why this only fires with generic functions.

## Fix

When the branch type is a non-int, non-void type and the target is the default `int`, use the merged branch type for the temporary. The sum-type override path is unchanged, and explicit int targets (e.g. `int(expr)` casts) still work since they arrive as a real target type, not the default.

## Tests

- New regression test `vlib/v/tests/if_expr_generic_call_type_test.v`: generic `min`/`max` calls in both then/else branches and an explicit `min[f64]` instantiation; fails before the fix, passes after.
- Existing if-expr tests (`if_expr_in_or_block_test`, `nested_if_expr_method_call_test`, `orm_if_expr_value_test`, `return_error_in_if_expr_test`, `autofree_if_expr_call_arg_with_struct_init_test`, `array_map_alias_if_expr_test`, `indexexpr_with_if_expr_test`) all pass with the patch.

### DIFF
--- a/vlib/builtin/closure/closure_nix.c.v
+++ b/vlib/builtin/closure/closure_nix.c.v
@@ -13,7 +13,10 @@ struct ClosureMutex {
 
 @[inline]
 fn closure_mtx_ptr_platform() voidptr {
-	return unsafe { voidptr(&g_closure.closure_mtx[0]) }
+	// Explicit `ClosureMutex.` (instead of the promoted `g_closure.closure_mtx`)
+	// so the V3 C backend emits the embedded sub-struct access; the promoted
+	// path emits `g_closure.closure_mtx` which does not exist at the C level.
+	return unsafe { voidptr(&g_closure.ClosureMutex.closure_mtx[0]) }
 }
 
 @[inline]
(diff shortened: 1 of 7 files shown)


medvednikov:
@codex review

PR #28473, opened 2026-09-09
cgen: fix generic interface pointer arg specialization

### LINKED ISSUE #28471: Generic instantiated at both an interface type and a concrete type miscompiles
### Describe the bug

Instantiating one generic at an interface type and at a concrete type, in the
same program, miscompiles the *interface* instantiation. The two instantiations
are unrelated: neither mentions the other and either one alone is correct.

Adding the second instantiation changes how the first is compiled, which is what
makes this a codegen bug rather than invalid input.

There are two symptoms from the same cause. Which one appears depends on the
number of type parameters but the difference in danger is large:

| shape | symptom |
| --- | --- |
| one type parameter | invalid C, the build fails |
| two type parameters | compiles cleanly, then segfaults deterministically |

### 1. Build failure

```v
module main

interface Speaker {
	speak() string
mut:
	bump()
}

@[heap]
struct Dog {
mut:
	n int
}

fn (d &Dog) speak() string {
	return 'woof'
}

fn (mut d Dog) bump() {
	d.n++
}

struct Box[T] {
mut:
	item T
}

fn apply[T](b Box[T], f fn (mut it T) string) string {
	mut it := b.item
	return f(mut it)
}

fn main() {
	mut d := &Dog{}
	a := Box[Speaker]{
		item: d
	}
	c := Box[&Dog]{
		item: d
	}
	r1 := apply[Speaker](a, fn (mut it Speaker) string {
		it.bump()
		return it.speak()
	})
	r2 := apply[&Dog](c, fn (mut it &Dog) string {
		it.bump()
		return it.speak()
	})
	println('${r1} ${r2} n=${d.n}')
}
```

```
cc: In function 'main__apply_T_main__Speaker':
cc: error: incompatible type for argument 1 of 'I_main__Dog_to_Interface_main__Speaker'
cc:  7211 |         return f(HEAP(main__Speaker, I_main__Dog_to_Interface_main__Speaker(it)));
cc:       |                                                                             ^~
cc:       |                                                                             |
cc:       |                                                                             main__Speaker
```

The generated function is `main__apply_T_main__Speaker`, the instantiation at
`Speaker`. Inside it, V emits a call to `I_main__Dog_to_Interface_main__Speaker`,
the *`Dog`* conversion and passes it a value that is already a `main__Speaker`.
The `&Dog` instantiation has leaked into the `Speaker` one.

### 2. The same program with the second instantiation deleted

Byte for byte identical apart from removing `c` and `r2`:

```v
fn main() {
	mut d := &Dog{}
	a := Box[Speaker]{
		item: d
	}
	r1 := apply[Speaker](a, fn (mut it Speaker) string {
		it.bump()
		return it.speak()
	})
	println('${r1} n=${d.n}')
}
```

```
woof n=1
```

Compiles and runs correctly. `Box[&Dog]` merely existing is what breaks
`Box[Speaker]`.

### 3. Silent variant - compiles, then segfaults

With two type parameters the program builds successfully and crashes at runtime,
deterministically, 5 runs out of 5. No threads, no channels, no `unsafe`.

```v
module main

interface Entity {
	kind() string
mut:
	bump()
}

@[heap]
struct Player {
mut:
	bumps int
}

fn (p &Player) kind() string { return 'player' }
fn (mut p Player) bump() { p.bumps++ }

struct Ref[T] {
mut:
	entity T
}

fn call[T, R](r Ref[T], f fn (mut e T) R) R {
	mut ent := r.entity
	return f(mut ent)
}

fn main() {
	mut p := &Player{}
	pi := Ref[Entity]{ entity: p }
	k1 := call[Entity, string](pi, fn (mut e Entity) string { e.bump() return e.kind() })
	pc := Ref[&Player]{ entity: p }
	k2 := call[&Player, string](pc, fn (mut e &Player) string { e.bump() return Entity(e).kind() })
	println('Ref[Entity]=${k1} Ref[&Player]=${k2} bumps=${p.bumps}')
}
```

```
signal 11: segmentation fault
```

### Expected Behavior

Both programs print their results. An instantiation of a generic at one type
must not change the code generated for an instantiation at another type.

### Current Behavior

The interface instantiation is generated with the concrete instantiation's
interface conversion, producing either invalid C or a binary that faults.

Each of these works on its own. Only the combination fails:

| case | result |
| --- | --- |
| `Box[Speaker]` alone | works |
| `Box[&Dog]` alone | works |
| `Ref[Entity]` alone, value passed across a channel to another thread | works |
| `Ref[&Player]` alone, same | works |
| `Box[Speaker]` + `Box[Dog]` in one program (non pointer) | works |
| `Box[Speaker]` + `Box[&Dog]` in one program (pointer) | **fails** |

Two further conditions both needed:

The concrete instantiation must be at a pointer type. `Box[Speaker]` beside
`Box[Dog]` is fine; `Box[Speaker]` beside `Box[&Dog]` fails. Since an interface
value is itself pointer shaped, the two may be colliding somewhere the type is
keyed or the symbol is mangled.

The interface must be mutated through. The same pair of instantiations with
an immutable interface, no `mut:` section, no `mut it` binding , compiles and
runs correctly.

### Possible Solution

Not investigated. The generated symbol name `main__apply_T_main__Speaker`
containing a call to `I_main__Dog_to_Interface_main__Speaker` suggests the
interface conversion helper is selected from the wrong instantiation's type
context when the generic is specialised.

_No response_

### V version

V 0.5.2 59e148a

|V full version      |V 0.5.2 1f128c94619fb6e4863c5f81883e1b860405c8fa.59e148a
|Memory              |0.42GB/7.62GB
|                    |
|V last modified time|2026-09-09 12:19:48
|                    |
|Current working dir |OK, value: /home/scher/projects/vlang/netherrack
|                    |
|V git status        |0.5.2-617-g59e148ad
|.git/config present |true
|                    |
|emcc version        |N/A

nepinhum:
Fixes #28471.

This PR fixes old C generator argument lowering for generic function specializations where the same generic call path is instantiated with both interface and pointer argument types.

`ref_or_deref_arg_ex()` could use the stale `CallArg.typ` stored on the shared generic AST. In cases like `T = Interface` followed by `T = &Struct`, that could make cgen emit an invalid interface conversion for an already interface argument.
 
This PR also refreshes identifier argument types from the current generic specialization scope before ref/deref/interface conversion. The refresh is guarded so it only applies when the scope lookup resolves to the same bound variable declaration as the identifier, avoiding accidental replacement from same name shadowing declarations.

A regression test covers one and two generic-parameter call paths in both interface first and pointer first specialization order.

## Tests

- [x] make
- [x] v test vlib/v/tests/generics/generic_interface_pointer_specialization_recheck_test.v

### DIFF
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -7213,16 +7213,20 @@ fn (mut g Gen) ref_or_deref_arg_ex(arg ast.CallArg, expected_type_ ast.Type, lan
 				arg_typ = resolved_param_type
 			}
 		}
-		if in_generic_context {
-			resolved_scope_type := g.resolved_scope_var_type_uncached(arg.expr)
-			if resolved_scope_type != 0 {
-				resolved_arg_type := g.unwrap_generic(g.recheck_concrete_type(resolved_scope_type))
-				skip_inherited_option_storage_type := arg.expr.obj is ast.Var
-					&& arg.expr.obj.is_inherited && !expected_type.has_option_or_result()
-					&& arg_typ != 0 && !arg_typ.has_option_or_result()
-					&& resolved_arg_type.has_option_or_result()
-				if !skip_inherited_option_storage_type {
-					arg_typ = resolved_arg_type
+		if in_generic_context && arg.expr.obj is ast.Var {
+			if scope_var := arg.expr.scope.find_var(arg.expr.name) {
+				if scope_var.pos.pos == arg.expr.obj.pos.pos {
+					resolved_scope_type := g.resolved_scope_var_type_uncached(arg.expr)
+					if resolved_scope_type != 0 {
+						resolved_arg_type := g.unwrap_generic(g.recheck_concrete_type(resolved_scope_type))
+						skip_inherited_option_storage_type := arg.expr.obj.is_inherited
+							&& !expected_type.has_option_or_result() && arg_typ != 0
+							&& !arg_typ.has_option_or_result()
+							&& resolved_arg_type.has_option_or_result()
+						if !skip_inherited_option_storage_type {
+							arg_typ = resolved_arg_type
+						}
+					}
 				}
 			}
 		}
--- a/vlib/v/gen/c/fn.v
+++ b/vlib/v/gen/c/fn.v
@@ -7213,6 +7213,19 @@ fn (mut g Gen) ref_or_deref_arg_ex(arg ast.CallArg, expected_type_ ast.Type, lan
 				arg_typ = resolved_param_type
 			}
 		}
+		if in_generic_context {
+			resolved_scope_type := g.resolved_scope_var_type_uncached(arg.expr)
+			if resolved_scope_type != 0 {
+				resolved_arg_type := g.unwrap_generic(g.recheck_concrete_type(resolved_scope_type))
+				skip_inherited_option_storage_type := arg.expr.obj is ast.Var
+					&& arg.expr.obj.is_inherited && !expected_type.has_option_or_result()
+					&& arg_typ != 0 && !arg_typ.has_option_or_result()
+					&& resolved_arg_type.has_option_or_result()
+				if !skip_inherited_option_storage_type {
+					arg_typ = resolved_arg_type
+				}
+			}
+		}
 	}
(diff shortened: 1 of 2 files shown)


medvednikov:
v1 has been replaced by v3.

please fix this in v3 if it's a problem there.

Issue #28467, opened 2026-09-09
Prerequisite: harden raw assembly diagnostics and constant-time validation (#28447)


quaesitor-scientiam:
Follow-up to #28447. This is a prerequisite for #28457.

## Goal

Make raw assembly exacting but caught: every supported transcription error
should become a V compiler diagnostic at the correct source line, rather than
silent corruption or an opaque C-toolchain failure. Users still write real
assembly; the compiler supplies the safety and diagnostic boundary around it.

This work should land before converting production crypto primitives or using
OpenSSL 4.0.2 as the performance target.

## Required diagnostics and checks

1. **Clobber completeness**

   Analyze the instructions in a raw template and detect registers and flags
   written by the template but absent from the clobber list. Cover GPRs, XMM,
   YMM, ZMM, condition codes, and memory effects where the instruction table
   can prove them. Use a conservative diagnostic when the analyzer cannot
   determine the effect. The primary target is preventing `-prod` miscompiles.

2. **Operand width versus mnemonic**

   Generalize the width validation from #28369 to raw templates. Check typed
   operands, explicit register names, operand modifiers, and mnemonic width
   together. Cases such as a `movl` using an operand bound from a `u64` must
   produce a V diagnostic with the template and source location.

3. **Register-pressure feasibility**

   Model usable register classes after fixed registers, inputs, outputs, and
   clobbers are accounted for. Report impossible constraints such as a block
   requiring more GPRs or vector registers than the target ABI can provide,
   before invoking the C compiler. Include Win64 nonvolatile registers and
   target-specific reserved registers in the model.

4. **Label hygiene**

   Detect numeric or named labels and references in reusable raw templates.
   Warn or error when a label does not use GNU `%=` uniqueness, or when a
   documented local-label convention is not followed. The diagnostic must
   explain why inlining or multiple instantiations can collide.

5. **Target-capability gates**

   When GNU raw assembly is selected under `-cc msvc`, emit a direct V error
   at the assembly block: MSVC x64 does not support this form; use the external
   `.S`, MASM `.asm`, or prebuilt `.obj` path instead. Do not defer this to an
   undeclared symbol, impossible constraint, or C-toolchain error.

6. **Constant-time floor**

   Define an opt-in way to mark a crypto kernel as constant-time and decide
   which checks can be enforced statically. At minimum, reject or flag
   data-dependent branches and data-dependent memory addressing in the
   supported instruction subset. For instructions or templates outside that
[...]

PR #28477, opened 2026-09-10
v3: harden asm goto lowering (case-insensitive mnemonics, lock-scope diagnostic)

quaesitor-scientiam:
Small, related hardening fixes for the v3 backend's `asm goto` support
(follows #28460), turning silent-corruption / opaque-C-error paths into
correct behavior or positioned V diagnostics — in the spirit of #28467.

### 1. Case-insensitive branch mnemonics

`c_inline_asm_goto_branch_label_operand_index` matched mnemonics against
lowercase literals only (`call`, `j*`, `loop*` on x86; `b`/`bl`/`b.*`/
`cb*`/`tb*` on arm64). An uppercase branch such as

```v
asm goto amd64 {
    retry: JNE done
    ; ; ; ; done
}
```

was not recognised as a branch, so its label was never lowered to the
`%l[...]` placeholder and the generated C was invalid. The mnemonic is now
lower-cased before matching (GNU `as` accepts either case). Classification
only — emitted assembly keeps the author's casing.

### 2. `asm goto` across a `lock` scope → checker diagnostic

The C backend cannot lower an `asm goto` whose target label is in a
different `lock`/`rlock` scope than the block, so `gen_c_inline_asm_stmt`
wrote `#error asm goto into or out of a lock scope is not supported` into
the generated C — invisible to `v -check` and to editor tooling, surfacing
only as an opaque C-toolchain failure with no source position.

A new checker pass (`check_asm_goto_lock_scopes`) walks each function body,
records the enclosing `lock_expr` path at every label declaration and
every `asm goto` block, and emits a positioned `.compile_error` when a
target crosses a lock-scope boundary. Same-scope `asm goto` inside a
`lock` block stays allowed (the backend already lowers it). The codegen
`#error` is kept as an unreachable backstop.

Part of #28467.

### 3. Parity-gap fixes for change 2 (adversarial review)

A multi-agent adversarial review of change 2 found three cases where the new
checker pass and the codegen `#error` disagreed, all now fixed:

- **Closures / `spawn fn(){}`** — the walk bailed at the `fn_literal`
  boundary, so a crossing jump *inside a closure* was never diagnosed while
  codegen still emitted `#error`. Each nested function body is now scanned
  and reported independently against a fresh scope path, mirroring codegen's
  per-function `collect_fn_prelude_scan`.
- **Undeclared target label** — the pass skipped a label it hadn't recorded;
  codegen treats a missing label as the empty path and still flags it when
  the block sits in a lock. Matched via `label_scopes[target] or { [] }`.
- **Objectless `lock {}` / `rlock {}`** — was counted as a lock scope,
  spuriously rejecting a valid jump past it; `gen_lock_enter` skips a lock
  with no objects, and the pass now does too.

New tests: closure, `rlock`, undeclared-label, objectless-lock.

### Test plan

- [x] `./vnew -stats test vlib/v3/tests/inline_asm_c_lowering_test.v` — 34/34, incl. new mnemonic-case, within-lock, across-lock (rewritten to assert the V diagnostic), rlock, closure, undeclared-label and objectless-lock tests
- [x] Negative checks: revert each hunk in turn → the matching test fails on the pre-change code; restore → passes
- [x] `v fmt -verify` clean on all touched files
- [x] `vlib/v3/tests/{mixed_lock_codegen,checker_diagnostic_files,match_and_wasm_diagnostics}_test.v` pass (two unrelated pre-existing macOS/v3 cross-compile failures in `lock_codegen_test.v` / `type_checker_errors_test.v` confirmed against a clean baseline)
- [ ] CI green

https://claude.ai/code/session_01HgHwaPqQYDjtcWLCZJNex5

### DIFF
--- a/vlib/v3/gen/c/stmt.v
+++ b/vlib/v3/gen/c/stmt.v
@@ -3356,20 +3356,23 @@ fn (mut g FlatGen) lower_c_inline_asm_goto_raw_labels(template string, labels []
 	return lowered
 }
 
+// c_inline_asm_goto_branch_label_operand_index matches mnemonics case-insensitively,
+// since GNU as accepts `JNE`/`B.EQ` just like their lowercase forms.
 fn c_inline_asm_goto_branch_label_operand_index(instruction string, arch string, operand_count int) int {
 	if operand_count == 0 {
 		return -1
 	}
+	name := instruction.to_lower_ascii()
 	if is_c_inline_asm_x86_arch(arch) {
-		return if instruction.starts_with('call') || instruction.starts_with('j')
-			|| instruction in ['loop', 'loope', 'loopne', 'loopz', 'loopnz'] {
+		return if name.starts_with('call') || name.starts_with('j')
+			|| name in ['loop', 'loope', 'loopne', 'loopz', 'loopnz'] {
 			0
 		} else {
 			-1
 		}
 	}
-	if arch in ['arm64', 'aarch64'] && (instruction == 'b' || instruction == 'bl'
-		|| instruction.starts_with('b.') || instruction.starts_with('cb') || instruction.starts_with('tb')) {
+	if arch in ['arm64', 'aarch64'] && (name == 'b' || name == 'bl'
+		|| name.starts_with('b.') || name.starts_with('cb') || name.starts_with('tb')) {
 		return operand_count - 1
 	}
 	return -1
--- a/vlib/v3/tests/inline_asm_c_lowering_test.v
+++ b/vlib/v3/tests/inline_asm_c_lowering_test.v
@@ -425,6 +425,86 @@ fn main() {
 	assert !c_source.contains('#error asm goto'), c_source
 }
 
+fn test_asm_goto_across_rlock_scope_is_rejected() {
+	generate, _ := generate_inline_asm_c('goto_rlock_scope_program', 'struct Counter {
+mut:
+	value int
+}
+
+fn main() {
+	shared counter := Counter{}
+	rlock counter {
+		asm goto amd64 {
+			jmp done
+			; ; ; ; done
+		}
+	}
+	done:
+}
+')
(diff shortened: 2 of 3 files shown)


quaesitor-scientiam:
CI status: every failure here is inherited from master or infra — nothing in this PR.

- "All code is formatted" (×6) — vlib/crypto/scram/ is not vfmt'ed on master; fixed in #28483.
- "Test v3" → parallel_cc_test.v:51 — master's own V3 CI fails the identical assertion (and c_compiler_flags_test.v besides) on every run today, e.g. run 34579770710 on 316db75a07.
- riscv64 / s390x — error: incompatible type for argument 1 of 'v3_map_str' in generated C; master fails identically on every run of riscv64_linux_ci.yml since 5ff8ecb55f.
- clang-openbsd / tcc-openbsd — VM SSH never came up (Connection timed out during banner exchange). Linux gcc — X11 multiwindow harness timeout. Contract worker / tccbin-automation-contract — 270 s then cancelled, and its dependent.
Because test_all.vsh stops after the first failing unit batch, the batch containing vlib/v3/tests/inline_asm_c_lowering_test.v never ran, so CI hasn't exercised this PR's own tests. Ran it directly on macOS arm64 (V3 default): passes.

The format gate is the reason most of the matrix never reports. Once #28483 (scram) and 
vlang/v#28498 (check-md ```vml fence) land — plus the parser (vlang/v#28500) and build-tools fixes already in flight — this should go green on a re-run with no changes here.

medvednikov (commented):
Reviewed the code changes only; CI status intentionally not considered. I did not find a concrete correctness issue. The mnemonic normalization is classification-only, and the checker walk mirrors codegen's per-function lock-scope model, including nested function boundaries, objectless locks, and undeclared labels.

JalonSolov:
Try merging main to resolve at least some of the CI issues.

PR #28482, opened 2026-09-10
v3: fix mut generic callback params losing a pointer level

nepinhum:
A `mut` callback parameter adds one pointer level for the caller slot on top of
whatever the declared type already is. All four combinations have to stay
distinct:

| callback | effective C parameter |
| --- | --- |
| `fn (T)`, `T = Dog` | `Dog` |
| `fn (mut T)`, `T = Dog` | `Dog*` |
| `fn (T)`, `T = &Dog` | `Dog*` |
| `fn (mut T)`, `T = &Dog` | `Dog**` |

The last one didn't work. Six places disagreed about where that slot pointer
belongs and their errors cancelled out and only `mut` combined with a pointer
`T` misbehaved:

- `normalize_fn_type_param_text` refused to add the slot when the type text
  already started with `&`, collapsing `fn (mut Dog)` and `fn (mut &Dog)` onto
  the same `&Dog`.
- `Type.name()` printed `mut ` in front of the already lowered parameter, so the
  parse -> `name()` -> parse round trip that monomorphization performs in
  `specialized_signature_type_text` gained a level every time it ran.
- `lock_colliding_main_substitution_type_text` stripped a second `&` from a mut
  payload, eating the pointer that belongs to `T`.
- `fn_literal_type`, `contextual_generic_lambda_type`, `fn_literal_type_text`
  and `lift_fn_literal` ignored `op == .amp`, a `fn (mut it &Dog)` literal
  claimed one level less than the C it actually emits.
  
Each representation now defines the slot lowering rule in one place, ABI lowering is applied exactly once. The param !is Pointer checks in c_type, c_extern_abi_type and fn_type_effective_param no longer determine whether the mutable caller slot has already been lowered; they only preserve types whose slot representation has already been established.

## Tests

- vlib/v3/types/type_cache_test.v - all four combinations through text
  normalization, the parsed type, its rendered name and c_type, plus a
  parse -> name() -> parse round trip check
- vlib/v3/transform/monomorphize_test.v - collision locking keeps the declared
  & of fn (mut it T) instantiated with &Dog
- vlib/v3/tests/generic_cross_module_arg_codegen_test.v - end to end runs for
  both T = Dog (field mutation through the callback) and T = &Dog (slot
  reassignment through the callback)


medvednikov:
Checked this against the legacy/v1 compiler. I don't think this is an active v1 bug; this PR is bringing v3 back in line with behavior v1 already has.

In `vlib/v/gen/c/fn.v`, v1's `fn_decl_params` keeps the mutable caller slot separate from a pointer that comes from the generic argument. For a `mut x T` parameter it resolves `orig_typ`, and when the concrete `T` is already a pointer (`T = &int`) it uses `surface_typ.ref()` to add the extra slot level. It also tracks `orig_was_ptr` so an `&` written in the parameter signature is not confused with the pointer contributed by `T`. In other words, the same distinction here is already intentional in v1: `fn (mut T)` with `T = &Dog` lowers to `Dog**`, not `Dog*`.

`vlib/v/ast/str.v` documents the same invariant from the parser/rendering side: `mut` lowering adds one indirection, but an explicit `&` from `mut x &T` must be preserved separately.

There was a related v1 fix in `24120e7a80d798a1625a08d5cc02b96e17148bdb` (`cgen: fix mut generic param indirection under -new-generic-solver`), but that commit only broadened the already-existing lowering to the new generic solver and explicitly notes that the default compiler was unaffected. The `T = &int`/extra-`ref()` handling was already present before that commit.

I also went through the v3 changes here. The representation changes look consistent with that v1 model, and the tests cover the important controls: all four value/pointer + mut/non-mut combinations, parse -> `name()` -> parse stability, generic substitution/collision locking, and an end-to-end `&Dog` callback that reassigns the caller slot. I didn't spot another issue in the patch.

medvednikov:
@codex review

medvednikov:
Thanks!

PR #28483, opened 2026-09-10
vfmt: fix v3 formatter mut-smartcast drop, comment/brace oscillation, and attribute quote stripping; reformat vlib

quaesitor-scientiam:
## Summary

`v fmt -verify ./vlib` (the `v test-fmt` CI step) rejected 23 vlib files that matched neither the v3 nor the legacy formatter output (24 since #28487 landed — see below). Running `v fmt -w` on them was not an option because the v3 formatter had bugs, one of which silently corrupts code. This PR fixes the formatter, adds regression tests, and reformats the files.

### Bug 1 — `mut` dropped from selector smartcasts (corrupts code)

`if mut nd.child is A { … }` and `match mut nd.child { … }` were emitted without the `mut`, turning a mutable smartcast immutable — any branch that mutates through it then fails to compile. `if mut e is A` (bare ident) was fine.

Cause: the parser's prefix `mut` marks only the node parsed immediately after it, so for `mut a.b` the `is_mut` flag lives on the leftmost `a`, not on the selector node the smartcast sees. Fix (`vlib/v3/gen/v/gen.v`): `smartcast_operand_is_mut()` follows the selector/index chain; used by `is_expr`, `!is`, and `match`. `checker.v` alone has ~50 such sites, which is why it was never formatted.

### Bug 2 — trailing `{ // comment` hoisted before the brace (non-idempotent)

`if mut s is int { // c` formatted to `if mut s is int // c\n {`, which re-parsed differently on the next pass, so the file could never satisfy `-verify`. Same for `!is` and multi-line `!in [ … ] { // c`.

Cause (`vlib/v3/parser/parser.v`): `is_expr` and the `!is`/`!in` prefix wrappers were built with `add_node()`, which gives a default *point* span on the **next** token — for an `if` condition that is the `{`. The formatter's trailing-comment emission then saw the block's comment as trailing the condition. Fix: span them operand..type via `add_node_from`, exactly as `in_expr` already did.

### Bug 3 — whole-string field attribute quotes stripped

`b int @['C\x5cnD']` became `@[C\x5cnD]`, which is invalid syntax (`@["dq"]` → `@[dq]` likewise). `parse_field_attrs_with_kinds` unquotes for reflection's legacy `FieldData.attrs` form; the formatter re-emits field attributes from those same strings. Fix: keep the quotes when `is_fmt`.

### Bug 4 — comment between `}` and `else if` left a trailing space (found by CI)

`}` / `// comment` / `else if cond {` formatted to `} else if ␠` + newline + comment + condition: `v fmt -verify` accepted it (idempotent) but `v vet` rejected the trailing whitespace, failing `test-cleancode`. Fix (`if_expr` in `gen.v`): emit the comment first, then `else …` on a fresh line — the layout `checker.v` already had on master.

## Relationship to #28487

#28487 (merged) fixes three *other* v3-formatter regressions (operator-method spacing, `$if`-nested imports, `for`-header comments); none overlap with the four above, and master still fails `test-fmt` on 24 files after it. This branch includes #28487 via merge, and additionally reformats `vlib/v/tests/option_struct_eq_test.v`, which #28487's `== (` spacing fix turned red.

## Verification

- Four regression tests in `vlib/v3/gen/v/gen_test.v`; each fails without its fix and passes here.
- The 24 files were reformatted with the fixed formatter: **zero `mut` tokens lost** in any file, and the full token multiset of every reformatted file (comments stripped) equals master's.
- `v fmt -verify ./vlib` → clean; `v test-cleancode` → vet 3276/3276, fmt 6413/6413.
- The legacy compiler (`-d v1_fallback cmd/v`, which is what actually compiles `vlib/v/checker`, `transformer`, `gen/c`) builds from the reformatted tree and self-compiles.
- `v test vlib/v3/parser vlib/v3/gen/v cmd/tools/vfmt_test.v vlib/crypto/scram vlib/v/fmt vlib/v3/transform vlib/v3/types` pass, except three failures that are identical on clean master (`scanner_test.v`, `transform_parallel_test.v`, `storage_source_test.v`).

One `scanner.v` string literal (the "invalid character literal" message) is respelled without escaped backticks: the legacy and v3 scanners decode `` \` `` differently, so the formatter's re-escaped form would have changed the message's value under the legacy compiler; both scanners agree on the new spelling. The broader string-literal re-escaping behaviour of the v3 formatter (which also rewrites `\x5c`, `$`, `­` in a few test files here) is being looked at separately and is not changed by this PR.

The formatter/parser diff is ~100 lines; the rest is the mechanical reformat of the 24 files.

### DIFF
--- a/vlib/crypto/scram/attributes_test.v
+++ b/vlib/crypto/scram/attributes_test.v
@@ -18,7 +18,7 @@ fn test_escape_saslname_escapes_only_what_the_grammar_forbids() {
 }
 
 fn test_escape_saslname_refuses_a_nul_byte() {
-	escape_saslname('nul\0name') or {
+	escape_saslname('nul\x00name') or {
 		assert err.msg().contains('must not contain a NUL byte')
 		return
 	}
@@ -48,7 +48,7 @@ fn test_unescape_saslname_refuses_anything_it_did_not_produce() {
 		'a nearly truncated one':      'ab=2'
 		'a raw comma':                 'a,b'
 		'a raw comma after an escape': '=2C,'
-		'a NUL byte':                  'nul\0name'
+		'a NUL byte':                  'nul\x00name'
 	}
 	for what, name in cases {
 		unescape_saslname(name) or {
@@ -97,7 +97,7 @@ fn test_parse_attributes_accepts_a_utf8_extension_value() {
 }
 
 fn test_parse_attributes_refuses_a_nul_byte_in_any_value() {
-	for message in ['x=\0', 'r=nonce,x=before\0after'] {
+	for message in ['x=\x00', 'r=nonce,x=before\x00after'] {
 		parse_attributes(message) or {
 			assert err is MalformedMessage, message
 			assert err.msg().contains('must not contain a NUL byte'), message
@@ -188,8 +188,8 @@ fn test_parse_positive_int_refuses_anything_lenient() {
 fn test_decode_base64_refuses_non_canonical_encodings() {
 	assert decode_base64('', 'x')! == []u8{}
 	assert decode_base64('dGVzdA==', 'x')! == 'test'.bytes()
-	for value in ['dGVzdA', 'dGVzdA=', 'dGVzdA===', 'not!base64', 'dGVz dA==', '====',
-		'{AAA', 'AAA{', '=AAA', 'AA=A'] {
+	for value in ['dGVzdA', 'dGVzdA=', 'dGVzdA===', 'not!base64', 'dGVz dA==', '====', '{AAA', 'AAA{',
+		'=AAA', 'AA=A'] {
 		decode_base64(value, 'the salt') or {
 			assert err.msg() == 'scram: malformed message: the salt is not valid base64'
 			assert err is MalformedMessage, value
@@ -220,7 +220,7 @@ fn test_gs2_header_rendering() {
 }
 
 fn test_gs2_header_refuses_an_unusable_binding_name() {
-	for name in ['', 'has,comma', 'has=equals', 'has space', 'has_underscore', 'control\0byte',
+	for name in ['', 'has,comma', 'has=equals', 'has space', 'has_underscore', 'control\x00byte',
 		'café'] {
 		binding := ChannelBinding{
 			mode: .required
@@ -280,8 +280,8 @@ fn test_split_gs2_header_separates_the_three_parts() {
 }
 
 fn test_split_gs2_header_refuses_invalid_utf8_authzid() {
-	client_first := [u8(`n`), `,`, `a`, `=`, 0xff, `,`, `n`, `=`, `u`, `s`, `e`, `r`, `,`,
-		`r`, `=`, `a`, `b`, `c`].bytestr()
+	client_first := [u8(`n`), `,`, `a`, `=`, 0xff, `,`, `n`, `=`, `u`, `s`, `e`, `r`, `,`, `r`,
+		`=`, `a`, `b`, `c`].bytestr()
 	split_gs2_header(client_first) or {
 		assert err is MalformedMessage
 		assert err.msg().contains('must contain valid UTF-8')
@@ -307,7 +307,7 @@ fn test_validate_nonce_matches_the_printable_rule() {
 	validate_nonce('rOprNGfwEbeRWgbNEkqO')!
 	validate_nonce('%hvYDpWUa2RaTCAfuxFIlj)hNlF\$k0')!
 	validate_nonce('a+b/c=')!
-	for bad in ['', ' ', 'a,b', 'a b', 'tab\there', 'nul\0here', 'newline\n', 'é'] {
+	for bad in ['', ' ', 'a,b', 'a b', 'tab\there', 'nul\x00here', 'newline\n', 'é'] {
 		validate_nonce(bad) or { continue }
 		assert false, 'accepted the nonce `${bad}`'
 	}
--- a/vlib/crypto/scram/client_test.v
+++ b/vlib/crypto/scram/client_test.v
@@ -23,9 +23,9 @@ fn test_new_client_rejects_an_empty_username() {
 }
 
 fn test_new_client_rejects_nul_bytes_in_sasl_names() {
-	new_client(username: 'nul\0name', password: 'pencil') or {
+	new_client(username: 'nul\x00name', password: 'pencil') or {
 		assert err.msg().contains('must not contain a NUL byte')
-		new_client(username: 'user', password: 'pencil', authzid: 'nul\0authzid') or {
+		new_client(username: 'user', password: 'pencil', authzid: 'nul\x00authzid') or {
 			assert err.msg().contains('must not contain a NUL byte')
 			return
 		}
@@ -37,7 +37,7 @@ fn test_new_client_rejects_nul_bytes_in_sasl_names() {
 fn test_new_client_rejects_an_unusable_nonce() {
 	// An empty `nonce` is the documented request for a generated one, so the
 	// unusable values are the ones that would break the message grammar.
-	for bad in ['a,b', 'with space', 'nul\0byte', 'trailing\n'] {
+	for bad in ['a,b', 'with space', 'nul\x00byte', 'trailing\n'] {
 		new_client(username: 'user', password: 'pencil', nonce: bad) or { continue }
 		assert false, 'accepted the nonce `${bad}`'
 	}
@@ -45,8 +45,8 @@ fn test_new_client_rejects_an_unusable_nonce() {
 
 fn test_new_client_rejects_channel_binding_without_a_name() {
 	new_client(
-		username:        'user'
-		password:        'pencil'
+		username: 'user'
+		password: 'pencil'
 		channel_binding: ChannelBinding{
 			mode: .required
 			data: [u8(1), 2, 3]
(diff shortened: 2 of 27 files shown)


JalonSolov:
That's a WHOLE LOT of missing doc comments.  <sigh>

JalonSolov:
Local AI findings:

* `scanner_test.v:342` weakens scanner coverage by changing the `$` tests from unescaped `a$b` to escaped `a\$b`. This no longer verifies that an unescaped dollar without braces remains literal.

* `string_escape_opaque_backslash_test.v:11` similarly replaces opaque escapes such as `\x5c` with ordinary backslashes, removing coverage for the decoded-backslash edge case.

* Several SCRAM tests replace `\u00ad` with a literal invisible soft-hyphen, losing escape-decoding coverage and making the source harder to inspect.

quaesitor-scientiam:
> That's a WHOLE LOT of missing doc comments.

Can you be more explicit on what is missing

JalonSolov:
https://github.com/vlang/v/actions/runs/34510665406/job/102983693123?pr=28483#step:8:1

quaesitor-scientiam:
## CI status

`master` is currently deeply red. The unrelated PR #28473 (a cgen fix, same base commit) has **45 of the same failing checks** as this PR — that's the pre-existing baseline, not anything introduced here. In particular `code-formatting` is already red on `master`: `v fmt -verify` reports **24 failed** files there (the 23 this PR reformats, + `checker.v`). On this PR that step passes **6413/6413**.

Breakdown:

### Pre-existing on `master` (not caused by this PR)

| Check(s) | Evidence |
|---|---|
| `tools-linux/macos/freebsd/windows/openbsd` (`Check build-tools` / `Test tools`) | `cmd/tools/modules/testing/output_normal.v:69:22` + `:73:20` `error: accessing map value that contain pointers requires an or {} block outside unsafe` — identical on master `8b1c7b1334` ([Tools CI run 34434763666](https://github.com/vlang/v/actions/runs/34434763666)), same `7 failed, 71 passed` summary. File not touched here. |
| `time-linux`, `time-windows` | `Time CI` red on the last several master runs — `time` IANA-zoneinfo fallout. |
| `gcc-windows`, `msvc-windows`, `tcc-windows` | Windows CI red on master. |
| `clang-openbsd`, `tcc-openbsd`, `clang-freebsd`, `gcc-freebsd`, `tcc-freebsd` | BSD CI red on master. |
| `check-markdown` | `./doc/docs.md:7256:1: error: unrecognized command: "ml"` — the checker doesn't recognise the ` ```vml ` fences added in #28481. `doc/docs.md` is not touched here; also red on master head `64604901c8`. |
| `v3` | `vlib/v3/driver/parallel_cc_test.v` — `Test v3` also red on master head `64604901c8`. |
| `wasm-backend (ubuntu-22.04)` | 17/17 wasm tests fail in ~24 ms each (infra); `wasm-backend (windows-2022)` **passes** on this PR. |
| `bootstrap-v (ubuntu-latest)` (`Test v up`), `v-up-works-nix (*)` | `v up` remote-resolution checks; `bootstrap-v (macos-14)` **passes** on this PR. |
| `v-compiles-puzzle-vibes`, `vsl-compiles`, `vtl-compiles` | Fail inside third-party modules (`~/.vmodules/shy/lib/input.b.v: reading a union field requires unsafe`, etc.), unrelated to `vlib`. |
| `Build on … s390x`, `Build on … riscv64`, `Contract worker`, `tccbin-automation-contract` | Cross-arch / tcc-bootstrap infra. |
| `performance-compare-with-master` | Builds `master`, which is itself red. |
| `build-module-docs` | `vlib/v3/gen/arm64/linker.v:171:5: error: cannot use `mut` on struct field type` — reproduced on a clean `master` (`64604901c8`) worktree build with `./v doc -f html vlib/v3/gen/arm64/linker.v`. The error comes from the legacy parser (`vlib/v/parser/parse_type.v:607`) on a file neither of which this PR touches. |
| `misc-tooling` (`g++ -std=c++11` self-compile) | `expected unqualified-id before '=' token` / `expected primary-expression before 'mutable'` — reproduced identically on the same clean `master` worktree with `./v -cc g++ -no-std -cflags -std=c++11 -o v2 cmd/v`. |

The `v vet` "Function documentation seems to be missing" **warnings** in the `All code is formatted` step (e.g. `vlib/v/tests/bench/math_big_gcd/prime/maker.v`, `vlib/v/tests/builtin_arrays/array_fixed_init_node_test.v`, the 7 in `checker.v`) **existed before this PR** — they are not comments this reformat removed:

- Most of the flagged files (`maker.v`, `array_fixed_init_node_test.v`, `cmd/tools/*`, `vlib/v/ast/attr.v`, …) are not touched by this PR at all; on master those `pub fn`s simply have no doc comment above them.
- Master's own run of the same step for the base commit `64604901c8` ([job 102836510932](https://github.com/vlang/v/actions/runs/34466595775)) prints the identical warning set — the same 10 for those two test files and the same 7 `checker.v` functions. `v vet` warns on every undocumented `pub fn` in the tree (~1,489 on master); only *errors* fail the step, which is why it passes on master. The PR log counts a few higher only because the failing file's output is echoed a second time in the FAIL block.
- For the 23 files this PR does reformat, the complete set of comment texts in each file is identical to master's (the only difference anywhere is `//}` → `// }` spacing in `checker.v`). No comment was moved away from its function or deleted.

### Caused by this PR — fixed in 6a6c9c1693

`vlib/v/checker/checker.v:7273` — the reformat left `} else if ` with a trailing space (v3 formatter output for `master`'s `}` / `// prepend mod to look for fn call or const` / `else if` construct at old line ~7524). `v fmt -verify` accepts it (idempotent) but `v vet` — run by `test-cleancode` — rejects the trailing whitespace. That failed `code-formatting`, `Linux gcc/clang/tcc` (all green on master), and the `Ensure code is well formatted` step on the `sanitize-*`, `clang-macos`, `docker-*` and `space-paths-linux` jobs. Fixed: the v3 formatter now keeps a comment between `}` and `else` on its own line, followed by `else …` on a fresh line (regression test added; fails on the previous commit, passes now), and `checker.v` was re-derived from master's source with the fixed formatter — only that one hunk changed. Locally: `v test-cleancode` clean (3276/3276 vet, 6413/6413 fmt), token multiset of every reformatted file unchanged vs master (`mut`/`unsafe` counts identical), and the legacy `-d v1_fallback cmd/v` build still self-compiles.

### Known, not addressed here — string-literal re-escaping

The v3 formatter re-emits single-line string literals from their *decoded* value (`vlib/v3/gen/v/gen.v` `string_literal_text` → `quote_string`). In this reformat that turned `'A\x5cnB'` into `'A\\nB'` (`vlib/v/tests/string_escape_opaque_backslash_test.v` — the test for that very escape), `a$b` into `a\$b` (`scanner_test.v`), and `'­'` into a literal soft-hyphen (`crypto/scram/*_test.v`). The strings are byte-identical, but the tests lose the coverage they were written for. It is also not purely cosmetic: the legacy scanner decodes `` \` `` as a backtick while v3 decodes it as backslash+backtick, so the re-escaped form of `scanner.v`'s "invalid character literal" message would have changed that message's value under the legacy compiler — 6a6c9c1693 respells that one literal so both scanners agree. The formatter-level question (preserve the source spelling of a literal, as the multi-line path already does) is being looked at separately; I've left those test files as the formatter produces them rather than pre-empting that.

JalonSolov:
Looks like #28487 is fixing the same things, as well as the others you mentioned.

quaesitor-scientiam:
> Looks like #28487 is fixing the same things, as well as the others you mentioned.

Testing to verify

quaesitor-scientiam:
> Looks like #28487 is fixing the same things, as well as the others you mentioned.

Why #28487 doesn't supersede it:

- #28487 fixed three different bugs (operator-method spacing, $if-nested imports, for-header comments). It does not touch any of the four defects this PR fixes (mut dropped from selector smartcasts, is/!is/!in point-span hoisting { // comment, attribute quote stripping, comment between } and else if).

- Current master (with #28487) still fails v fmt -verify ./vlib on 24 files — the same 23 as before plus option_struct_eq_test.v, which #28487 itself turned red. I verified this on a checkout of master+#28487 earlier today.

- My branch already contains #28487 (merged in 7860464bb7), reformats that 24th file, and passes test-fmt 6413/6413 and test-cleancode locally.

JalonSolov:
I'm also seeing 24 files that fail the `-verify` check, after #28487 was merged.

quaesitor-scientiam:
Rebased onto master with #28487 merged in (`7860464bb7`): no conflicts, and the two sets of fixes are independent — all 140 `gen_test.v` tests (this PR's four + #28487's three + existing) pass together.

One follow-up: #28487's operator-method spacing fix (`fn (a T) == (b T)`) makes `vlib/v/tests/option_struct_eq_test.v` fail `test-fmt` on current master (it had been formatted with the old `==(` output). Reformatted it here in `094b0cda2a` (one line), so `v fmt -verify ./vlib` is clean again on this branch — 24 files reformatted in total now. Description updated accordingly.

quaesitor-scientiam:
Zero CI failures are attributable to this PR

@JalonSolov  anything stopping the merge for this one?

quaesitor-scientiam:
Closing as superseded — everything here is now on master by other routes:

- the four v3 formatter fixes in `vlib/v3/gen/v/gen.v` (mut-smartcast drop, comment/brace oscillation, attribute quote stripping, `}`/`else` comment placement) — absorbed by #28487 and #28491; rebasing onto `3c8278e764` leaves a zero diff in `gen.v`
- the `vlib/crypto/scram/` reformat that unblocked the format gate — landed in `e231bc95a0`
- the `scanner_test.v` five-argument fix — already on master

What survives after the rebase is 7 files / 35 lines, and all 35 are string re-spellings from the pre-#28491 formatter (`'\x61'` → `'a'`, `'\0'` → `'\x00'`, `$if` → `\`) — byte-identical strings, but in `scanner_test.v` and `string_escape_opaque_backslash_test.v` they rewrite the exact escapes those tests exist to check. #28491 landed specifically so the formatter stops doing that, so there is nothing left here worth merging.

PR #28485, opened 2026-09-10
v3: cache BSD production self-build objects

medvednikov:
## Summary

- split BSD production self-build C output into 32 balanced units
- cache those compiled units by generated source, declaration header, compiler, flags, target, and dependency contents
- keep non-BSD, non-production, and -nocache behavior unchanged

## FreeBSD benchmark

FreeBSD 15.1 arm64 VM, 8 CPUs, 16 GiB RAM; command: v -prod -parallel-cc self.

- V 0.5.1: 4.30s
- V3 first cache-populating build: 21.74s
- V3 repeated unchanged build: 2.80s

The warm V3 build reuses all 33 split objects and only relinks. Changed generated units or dependencies invalidate the corresponding cached objects.

## Validation

- ./vnew vlib/v3/driver/environment_test.v
- ./vnew vlib/v3/driver/parallel_cc_test.v
- FreeBSD: ./vbench5 vlib/v3/driver/environment_test.v
- FreeBSD: ./vbench5 vlib/v3/driver/parallel_cc_test.v
- markdown check for vlib/v3/README.md

The broad compiler_errors_test.v run was stopped after existing V3/V1 diagnostic fixture mismatches appeared; the focused driver suites pass on macOS and FreeBSD.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -69,7 +69,7 @@ const v3_fallback_native_manifest_key = '@native-input-manifest:v1'
 const v3_fallback_native_manifest_value = 'v3-native-input-manifest-v1'
 const bsd_selfhost_job_limit = 2
 const bsd_selfhost_parallel_cc_job_limit = 8
-const bsd_selfhost_parallel_cc_unit_count = 8
+const bsd_selfhost_parallel_cc_unit_count = 32
 const bsd_prod_selfhost_memory_per_job = u64(2) * 1024 * 1024 * 1024
 
 fn default_selfhost_job_count(jobs int, is_bsd_host bool, allow_overcommit bool, prod_parallel_cc bool, total_memory u64) int {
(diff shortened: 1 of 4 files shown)


medvednikov:
Cold-path follow-up pushed in 80d5425bfd: BSD production self-builds now retain the 32-unit layout only when the object cache is enabled. With `-nocache`, they use one C unit per compiler job, avoiding repeated parsing of the shared declaration header without changing production optimization.

FreeBSD 15.1 arm64 QEMU, 8 vCPUs, adjacent cold runs:

- current V1, `-prod -parallel-cc -nocache self`: 17.44s real / 82.89s user
- this PR V3, equivalent flags: 16.31s real / 74.99s user
- V 0.5.1, same user-facing command: 4.12s real / 16.75s user

The historical 0.5.1 result is real, but it is not a V1/V3 comparison on the same compiler: its generated C is about 6.7 MiB / 3,047 definitions, versus about 23 MiB / 11,273 reachable symbols today. `-profile` confirmed that the current uninstrumented V3 frontend is roughly 2s and the cold wall time is dominated by the external Clang production builds. A development-optimization experiment reached about 5.2s, but produced a materially slower compiler, so I did not weaken `-prod` semantics to chase that number.

vlib/v3/driver/driver.v
```diff
@@ -2177,44 +2271,122 @@ fn compile_v3_parallel_c(source_path string, c_compiler string, c_flag_plan &V3C
...
 			large_compile_flags.clone()
 		} else {
 			small_compile_flags.clone()
 		}
-		compile_args << ['-x', if objective_c { 'objective-c' } else { 'c' }, '-c', '-o',
-			object_name, source_name]
-		if show_command {
-			println('  > ${cmdexec.display(c_compiler, compile_args)}')
+		if use_object_cache && !owner {
+			compile_args << ['-include', cached_header_path]
```
medvednikov, line 2301: The cached header is still forced with `-include` before the unit source is parsed, but `V_PARALLEL_CC`, `_VPARALLELCC`, and `V3CACHE_PROGRAM_UNIT` are only defined *inside* that source. That changes declarations selected while the header is parsed. A concrete example is `closure_once_nix.h`: without `V_PARALLEL_CC`, every body object gets its own `static` once-state instead of `extern` references to the owner's state, so closure initialization can repeat/race. Please either pass these defines as compiler `-D` flags before `-include`, or put the cached header include back into the unit source after the defines.

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. One blocking correctness issue remains in the cached split-C path.

PR #28494, opened 2026-09-11
 net.http: use origin-form with a proxy tunnel

davidebeatrici:


vlib/net/http/http_proxy.v
```diff
@@ -147,6 +147,16 @@ fn (pr &HttpProxy) connect_tcp(host string) !&net.TcpConn {
 	}
 }
 
+fn (pr &HttpProxy) request_target(host urllib.URL, host_name string, port_part string, path string) string {
+	// RFC 9112: absolute-form only toward an HTTP proxy, and only for non-CONNECT http:// targets.
+	// SOCKS is a tunnel; the origin wants origin-form.
+	if pr.scheme in ['http', 'https'] && host.scheme == 'http' {
```
medvednikov: `http_do()` reaches the origin through `pr.dial()`, and for both `http` and `https` proxy schemes `connect_tcp()` first sends `CONNECT` and establishes a tunnel. Once CONNECT succeeds, the next request is sent to the origin through that tunnel and should use origin-form, just like SOCKS. This branch still returns absolute-form for a plain-HTTP origin whenever the proxy scheme is HTTP/HTTPS, even though this implementation is not sending that request directly to a forward proxy. Either plain HTTP should skip CONNECT and use absolute-form, or the current tunneled path should return `path` here as well.

medvednikov (commented):
Reviewed the code changes only; CI status intentionally not considered. I found one protocol mismatch in the new target-selection helper.

vlib/net/http/http_proxy.v
davidebeatrici: Let's use HTTP CONNECT and origin-form, for consistency. It also leads to a more minimal and straightforward change/diff.

PR #28496, opened 2026-09-11
x.crypto.ascon: improves Digest.absorb() by removing `.clone()` call

blackshirt:
This small PR contains an improvement to internal functions of `x.crypto.ascon` module.

On previous call of `Digest.absorb`, its calling `.clone()` by `mut msg := msg_.clone()` internally.
This is an O(n) allocation+copy on every streaming write, seemingly just so the function can reslice (msg = msg[n..]) without mutating the caller's array. Reslicing a []u8 in V doesn't require an owned copy ie, slicing in V is already a cheap, non-owning view (pointer/len only), so instead we just track how many bytes of msg have been consumed via `pos` and slice `msg[pos..]` directly — and avoid the clone entirely.

Its also removes deadly unused code on `util.v` just for small code reducing and some of `v fmt` reformatting the code.

Thanks


JalonSolov:
Several notices about unnecessary parentheses in ascon/digest.v and ascon/util.v, otherwise looks good!

blackshirt:
> Several notices about unnecessary parentheses in ascon/digest.v and ascon/util.v, otherwise looks good!

Yeah, i places a note on that with latest v 
Thank you

medvednikov (commented):
Reviewed the code changes only (ignoring CI).

The `Digest.absorb` rewrite looks correct to me: tracking `pos` preserves the consumed-byte accounting, and the existing 7-byte chunked hash tests exercise the partial-buffer/block-boundary path against the 8-byte rate.

One thing I’d change before merging: please keep `_likely_(msg.len > 0)` directly in the `if` and drop the `cond := _likely_(...)` workaround/comment. `_likely_` is documented and used as a branch hint in the condition, and V’s formatter tests cover `if _likely_(...)` directly. Keeping that canonical form also avoids relying on optimizer propagation through a temporary, and this change is unrelated to the allocation removal.

The `u64le` removal looks safe as well; I found no uses beyond its definition.

blackshirt:
> Reviewed the code changes only (ignoring CI).
> 
> The `Digest.absorb` rewrite looks correct to me: tracking `pos` preserves the consumed-byte accounting, and the existing 7-byte chunked hash tests exercise the partial-buffer/block-boundary path against the 8-byte rate.
> 
> One thing I’d change before merging: please keep `_likely_(msg.len > 0)` directly in the `if` and drop the `cond := _likely_(...)` workaround/comment. `_likely_` is documented and used as a branch hint in the condition, and V’s formatter tests cover `if _likely_(...)` directly. Keeping that canonical form also avoids relying on optimizer propagation through a temporary, and this change is unrelated to the allocation removal.
> 
> The `u64le` removal looks safe as well; I found no uses beyond its definition.

Thanks ..i will revise it

medvednikov (commented):
Reviewed the current PR diff only; CI status intentionally not considered. The Ascon `Digest.absorb` rewrite itself looks correct: `pos` preserves the old consumed-byte accounting without requiring an owned clone, including buffered partial blocks and full-block transitions. However, the current PR diff also contains ~18 unrelated compiler/Windows files and upstream fixes (for example V3 codegen regressions and `CreateProcess` reporting). That makes this PR unsafe to merge as an Ascon-only change because it would also merge unrelated behavior. Please rebase/retarget so the diff contains only the intended `x.crypto.ascon` changes (and their tests) before merging.

PR #28498, opened 2026-09-11
tools: make check-md treat ```vml fences as non-V code blocks

### LINKED ISSUE #28456: check-md misinterprets vml fenced blocks as the ml command
## Description

`check-md` misinterprets a Markdown fenced code block tagged `vml` as a V
documentation command named `ml`.

The current documentation contains this valid language-tagged block at
`doc/docs.md:7224`:

````markdown
```vml
Screen {
    id: root
}
```
````

Running:

```text
v check-md doc/docs.md
```

reports:

```text
unrecognized command: "ml", use one of: wip/ignore/compile/failcompile/okfmt/nofmt/oksyntax/badsyntax/cgen/globals/live/shared
```

## Expected behavior

`vml` should be treated as a normal Markdown language identifier for syntax
highlighting, or be explicitly supported as a non-V code block. It should not
be parsed as a V `check-md` directive.

## Additional context

The checker currently treats language tags beginning with `v` as V-specific
commands, which makes `vml` resolve to `ml`. The documentation block is not
intended to be compiled or formatted as V code.

> [!NOTE]
>

quaesitor-scientiam:
#fixes 28456

### Problem

`v check-md` fails on master:

```
doc/docs.md:7260:1: error: unrecognized command: "ml", use one of: wip/ignore/compile/...
```

`parse_line` treats every fence beginning with ```` ```v ```` as a V example and derives the check command from whatever follows the `v`, so the ```` ```vml ```` fence added with the VML docs is parsed as command `ml`. Because the block is also extracted as a V example, its contents are additionally run through `v fmt`, producing a phantom formatting error on every run.

This is currently failing the `check-markdown` job on every open PR, regardless of what the PR touches.

### Fix

Route fence languages that begin with `v` but are not V source (`vml` for now) through the ordinary non-V code-block path in `parse_line`, so they are neither counted as examples nor formatted or compiled. The list is a small const so further cases can be added without touching the parser.

I chose this over adding a no-op `'ml' {}` arm next to the existing `'mod' {}` one: that would clear the error but still push VML content through `v fmt` before the `match` is reached, leaving the phantom fmt error in place. VML isn't V syntax, so it shouldn't enter the example pipeline at all.

### Tests

New `test_check_md_vml_fence_is_not_a_v_example` in `cmd/tools/vcheck_test.v`: a ```` ```vml ```` block followed by a real ```` ```v ```` example. Asserts exit 0, no `unrecognized command`, and `Ex.: 1` (only the V example counted). Verified it fails on master with the exact error above before the fix.

`v check-md doc/docs.md` on master, before → after: the line-7260 error is gone, `Ex.` drops 344 → 343, `Fmt errors` drops 1 → 0. The remaining compile errors in that file are pre-existing and unrelated.

`v fmt -verify` clean on both files.

Fixes: #28456

### DIFF
--- a/cmd/tools/vcheck-md.v
+++ b/cmd/tools/vcheck-md.v
@@ -328,6 +328,8 @@ fn eline(file_path string, lnumber int, column int, message string) string {
 }
 
 const default_command = 'compile'
+// fence languages that begin with `v` but hold something other than V source
+const non_v_fence_languages = ['vml']
 
 struct VCodeExample {
 mut:
@@ -438,7 +440,7 @@ fn (mut f MDFile) check() CheckResult {
 }
 
 fn (mut f MDFile) parse_line(lnumber int, line string) {
-	if line.starts_with('```v') {
+	if line.starts_with('```v') && line.replace('```', '').trim_space() !in non_v_fence_languages {
 		if f.state == .markdown {
 			f.state = .vexample
 			mut command := line.replace('```v', '').trim_space()
--- a/cmd/tools/vcheck_test.v
+++ b/cmd/tools/vcheck_test.v
@@ -365,3 +365,34 @@ fn write_text_file(path string, content string) ! {
 	os.mkdir_all(os.dir(path))!
 	os.write_file(path, content)!
 }
+
+fn test_check_md_vml_fence_is_not_a_v_example() {
+	md_dir := os.join_path(os.vtmp_dir(), 'vcheck_vml_fence_${os.getpid()}')
+	os.rmdir_all(md_dir) or {}
+	os.mkdir_all(md_dir)!
+	defer {
+		os.rmdir_all(md_dir) or {}
+	}
+	md_path := os.join_path(md_dir, 'vml.md')
+	// VML markup is not V source, so a ```vml fence must not be extracted as a V example,
+	// and the parser must still pick up the V example that follows it.
+	write_text_file(md_path, '# VML fence
+
+```vml
+Screen {
+    Column {
+        Button { text: "Save" on_tap: app.save() }
+    }
+}
+```
+
+```v ignore
+fn after_vml() {}
+```
+')!
+	res :=
+		os.execute('${os.quoted_path(vexe)} check-md -hide-warnings -silent ${os.quoted_path(md_path)}')
+	assert res.exit_code == 0, res.output
+	assert !res.output.contains('unrecognized command'), res.output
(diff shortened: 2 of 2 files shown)


medvednikov:
Reviewed PR #28498 with CI ignored as requested. **Verdict: LGTM; no blocking correctness issues found.**

* The fix addresses the reported failure cleanly: exact ` ```vml` fences now bypass the V-example path and are handled as ordinary code blocks, so VML is neither formatted nor compiled. That matches issue #28456’s expected behavior.
* The regression test is well targeted: VML followed by a genuine V example, with assertions for successful exit, absence of `unrecognized command`, and exactly one V example. The PR is only 2 files / +34 −1, so the scope is appropriately narrow.
* **Non-blocking robustness note:** `line.replace('```', '').trim_space() !in non_v_fence_languages` compares the entire fence info string. Thus something like ` ```vml title=example` would still be interpreted as a V command. If such Markdown metadata is expected, extracting the first info-string token before checking `non_v_fence_languages` would make the abstraction more complete.

I would **approve** this PR as-is; the info-string case is suitable for a follow-up rather than a blocker. I did not submit a GitHub review or comment.

quaesitor-scientiam:
> Reviewed PR #28498 with CI ignored as requested. **Verdict: LGTM; no blocking correctness issues found.**
> 
> * The fix addresses the reported failure cleanly: exact ``  ```vml `` fences now bypass the V-example path and are handled as ordinary code blocks, so VML is neither formatted nor compiled. That matches issue [check-md misinterprets vml fenced blocks as the ml command #28456](https://github.com/vlang/v/issues/28456)’s expected behavior.
> * The regression test is well targeted: VML followed by a genuine V example, with assertions for successful exit, absence of `unrecognized command`, and exactly one V example. The PR is only 2 files / +34 −1, so the scope is appropriately narrow.
> * **Non-blocking robustness note:** `` line.replace('```', '').trim_space() !in non_v_fence_languages `` compares the entire fence info string. Thus something like ``  ```vml title=example `` would still be interpreted as a V command. If such Markdown metadata is expected, extracting the first info-string token before checking `non_v_fence_languages` would make the abstraction more complete.
> 
> I would **approve** this PR as-is; the info-string case is suitable for a follow-up rather than a blocker. I did not submit a GitHub review or comment.

Good catch — and it's slightly worse than that: the info string was being split on whitespace into multiple commands, so ```vml title=example produced two errors plus phantom fmt errors. Fixed in 
vlang/v#28539 by taking only the first info-string token, with a regression test for exactly this case.

PR #28502, opened 2026-09-11
parser: fix comment after a blank line before `pub mut:`/`mut:` breaking struct/interface parsing in `.parse_comments` mode (fix #28500)

### LINKED ISSUE #28500: parser: comment after a blank line before `pub mut:` / `mut:` breaks struct parsing in `.parse_comments` mode (fails `v vet`, `v_parser_test.v` on master)
### Describe the bug

The V1 parser fails on a struct declaration when a field is followed by a blank line, then a comment line, then a section keyword (`pub mut:`, `mut:`, `pub:`) — but only in `.parse_comments` mode. `.skip_comments` parses the same source cleanly.

This breaks `v vet`, and it makes `vlib/v/parser/v_parser_test.v` fail on master, because that test parses every file in the tree with `.parse_comments` and `vlib/v3/gen/arm64/linker.v` contains this pattern several times (58 errors in that one file).

### Reproduction

```v
pub struct Linker {
	macho int

	// frameworks to link
pub mut:
	frameworks []string
}
```

```
$ v vet min.v
min.v:5:5: error: cannot use `mut` on struct field type
```

`v -check-syntax min.v` (skip_comments) reports nothing. Removing **either** the blank line **or** the comment line makes `v vet` pass — it's specifically the `field → blank → comment → section keyword` sequence.

### Expected

The comment is attached/ignored and `pub mut:` is recognised as a section keyword, exactly as it is without the blank line.

### Why it hasn't shown up on master CI

Master's own CI currently fails at the "All code is formatted" gate (`vlib/crypto/scram/` is not vfmt'ed), so the Self tests that run `v_parser_test.v` never execute. #28483, which fixes the scram formatting, is the first PR to get past that gate and hits this immediately on every platform. Any PR that passes the format gate will fail on this until it's fixed.

### Environment

```
V 0.5.2 316db75, Windows (also reproduced in CI on ubuntu/macos/freebsd runners on #28483)
```

> [!NOTE]
>

quaesitor-scientiam:
Fixes #28500

### Problem

In `.parse_comments` mode (used by `v vet` and `vlib/v/parser/v_parser_test.v`), a struct field followed by a blank line, then a comment, then a section keyword failed to parse:

```v
pub struct Linker {
	macho int

	// frameworks to link
pub mut:
	frameworks []string
}
```

```
min.v:5:5: error: cannot use `mut` on struct field type
```

After a field, trailing comments are eaten with `eat_comments(follow_up: true)`, which stops at a blank line. The next loop iteration checked for section keywords (`pub mut:`, `mut:`, `pub:`, `__global:`, `module:`) *before* eating any comments, so the current token was still the comment, the section check failed, and `pub mut:` was parsed as a field named `pub` of type `mut`.

`vlib/v3/gen/arm64/linker.v` contains this pattern several times, so `v_parser_test.v` fails on master as soon as CI gets past the format gate (see #28483).

### Fix

- `struct_decl`: eat leading comments at the top of each field iteration. If a section keyword follows, attach the comments to the previous field's `next_comments` (the same owner the no-blank-line case already uses), so vfmt keeps them in front of the keyword; otherwise they stay the next field's `pre_comments` exactly as before.
- `interface_decl`: had the identical gap for its `mut:` section (`expecting type declaration`); same fix, attaching to the last parsed field/method.

### Tests

- New `vlib/v/fmt/tests/struct_decl_comment_before_section_{input,expected}.vv` covering all five struct section keywords and interface `mut:` after a field and after a method. Fails without the parser change, passes with it.
- `v test vlib/v/parser vlib/v/fmt cmd/tools/vvet` all pass; `v_parser_test.v` now parses `linker.v` with 0 errors; `v vet` is clean on the repro.

### DIFF
--- a/vlib/v/fmt/tests/struct_decl_comment_before_section_expected.vv
+++ b/vlib/v/fmt/tests/struct_decl_comment_before_section_expected.vv
@@ -1,0 +1,35 @@
+// The comment after a blank line, right before a section keyword,
+// should stay with the previous field.
+pub struct Linker {
+	macho int
+	// frameworks to link
+pub mut:
+	frameworks []string
+	// private, mutable
+mut:
+	count int
+	// public, immutable
+pub:
+	name string
+	// global
+__global:
+	flag bool
+	// module
+module:
+	handle voidptr
+}
+
+pub interface Reader {
+	name string
+	// mutable part
+mut:
+	count int
+	read() int
+}
+
+pub interface Writer {
+	flush()
+	// after a method
+mut:
+	write() int
+}
--- a/vlib/v/fmt/tests/struct_decl_comment_before_section_input.vv
+++ b/vlib/v/fmt/tests/struct_decl_comment_before_section_input.vv
@@ -1,0 +1,42 @@
+// The comment after a blank line, right before a section keyword,
+// should stay with the previous field.
+pub struct Linker {
+	macho int
+
+	// frameworks to link
+pub mut:
+	frameworks []string
+
+	// private, mutable
+mut:
+	count int
+
+	// public, immutable
+pub:
+	name string
+
+	// global
+__global:
+	flag bool
+
+	// module
+module:
+	handle voidptr
+}
+
+pub interface Reader {
+	name string
+
+	// mutable part
+mut:
+	count int
+	read() int
+}
+
+pub interface Writer {
+	flush()
+
+	// after a method
+mut:
+	write() int
+}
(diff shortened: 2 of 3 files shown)


quaesitor-scientiam:
Closing: the changes from this PR (the `struct_decl`/`interface_decl` fix and both `struct_decl_comment_before_section_*.vv` fmt tests) were incorporated verbatim into master in e231bc95a0a1c7bdf39526dde43104130010c494 ("ci: restore bootstrap and fix master failures"), which fixes #28500. The only remaining difference on this branch is the old V1-vfmt field alignment in `vlib/v/parser/struct.v`, which is why `code-formatting` still flagged it.

PR #28503, opened 2026-09-11
tools: fix `output_normal.v` map-value-with-pointers warning in the build-tools check (fix #28501)

### LINKED ISSUE #28501: tools: `output_normal.v` fails the build-tools check — map-value-with-pointers warning (as error) on `ccompiling[k]` / `crunning[k]`
### Describe the bug

`cmd/tools/modules/testing/output_normal.v` no longer builds under the CI "Check build-tools" step (`cmd/tools/vbuild-tools.v`), which compiles the tools with warnings promoted to errors:

```
cmd/tools/modules/testing/output_normal.v:69:22: error: accessing map value that contain pointers requires an `or {}` block outside `unsafe`
cmd/tools/modules/testing/output_normal.v:73:20: error: accessing map value that contain pointers requires an `or {}` block outside `unsafe`
```

That in turn fails `vbuild-examples.v`, `vbuild-tools.v` and `vbuild-vbinaries.v` (3/78) in every `tools-*` matrix job.

### Cause

`output_normal.v:69` and `:73` do

```v
cval := ccompiling[k]
...
cval := crunning[k]
```

on `map[string]LogMessage` values. `LogMessage` contains pointer-bearing fields, so the checker's map-value rule fires. That rule has been a `c.warn` since e1480b61dd (2024-10-09); on a plain `v -check` master emits it as a **warning** at exactly these two lines. These lines were last touched in 8e35f4d984 (2026-04-17), i.e. written after the rule already existed — so the tools have been failing the warnings-as-errors build since then.

### Fix

Both accesses are inside a loop over the map's own `keys()`, so the key is guaranteed present. Either form is correct:

```v
cval := ccompiling[k] or { continue }
```

or an `unsafe { ccompiling[k] }`, matching how the rule's message suggests.

### Why it hasn't shown up on master CI

Master's own CI currently fails at the "All code is formatted" gate (`vlib/crypto/scram/` is not vfmt'ed), so the tools-build check never runs. #28483 fixes the scram formatting and is the first PR to reach this step; it fails on every `tools-*` job for this reason. Any PR that passes the format gate will hit it.

### Environment

```
V 0.5.2 316db75 (reproduces locally with `v -check cmd/tools/vbuild-tools.v` as a warning; error under the CI build)
```

> [!NOTE]
>

quaesitor-scientiam:
Fixes #28501.

`cmd/tools/modules/testing/output_normal.v:69` and `:73` do `ccompiling[k]` / `crunning[k]` on `map[string]LogMessage`. `LogMessage` contains pointer-bearing fields, so the checker's *accessing map value that contain pointers requires an `or {}` block outside `unsafe`* rule fires. The CI **Check build-tools** step (`vbuild-tools.v`) compiles tools with `-W`, so the warning becomes an error and `vbuild-examples.v`, `vbuild-tools.v`, `vbuild-vbinaries.v` fail in every `tools-*` job. It has been masked on master because the *All code is formatted* gate fails earlier; any PR that passes that gate (e.g. #28483) hits it.

Both accesses iterate over the map's own `keys()`, so the key is always present — `or { continue }` satisfies the rule with no behaviour change.

Verified locally (legacy checker `v -W -check cmd/tools/vbuild-tools.v`): errors at 69:22 / 73:20 on `upstream/master`, exit 0 with this patch; `v -W cmd/tools/vtest.v` builds; `v fmt -verify` clean.

### DIFF
--- a/cmd/tools/modules/testing/output_normal.v
+++ b/cmd/tools/modules/testing/output_normal.v
@@ -66,11 +66,11 @@ fn (r &NormalReporter) report_current_running_and_compiling_status_periodically(
 		sb.writeln('')
 		sb.writeln('       >>>>> ${t.format_ss_micro()} | period ${pi:2} | started: ${t - start_t:10} ago | vjobs: ${r.njobs} | _test.v files: ${r.nfiles:5} | C: ${ckeys.len:3} | R: ${rkeys.len:3}')
 		for ik, k in ckeys {
-			cval := ccompiling[k]
+			cval := ccompiling[k] or { continue }
 			sb.writeln('       >>>>> compiling ${ik + 1:2}/${ckeys.len:-2}, T: ${cval.flow_id:2}, started: ${t - cval.when:10} ago, `${k}`')
 		}
 		for ik, k in rkeys {
-			cval := crunning[k]
+			cval := crunning[k] or { continue }
 			sb.writeln('       >>>>>   running ${ik + 1:2}/${rkeys.len:-2}, T: ${cval.flow_id:2}, started: ${t - cval.when:10} ago, `${k}`')
 		}
 		sb.writeln('')


JalonSolov:
Local check says everything looks good.  Will merge once we have as successful a CI run as we can get.  :-)

quaesitor-scientiam:
> Local check says everything looks good. Will merge once we have as successful a CI run as we can get. :-)

When my fmt PRs are merged more CI's should go green

quaesitor-scientiam:
CI status note for reviewers:

**`Check build-tools` now passes on every `tools-*` job** ([tools-linux (gcc)](https://github.com/vlang/v/actions/runs/34616104773/job/103318314866)). On master's latest run of the same workflow ([`ac986f938a`](https://github.com/vlang/v/actions/runs/34590011659)) that step fails with 7/78 tools — every tool that `import testing` (`vbuild-examples`, `vbuild-tools`, `vbuild-vbinaries`, `vtest`, `vtest-self`, `vtest-fmt`, `vtest-cleancode`), so the issue's "3/78" undercounted; all seven are fixed by this change.

The `tools-*` jobs are still red because of the **next** step, `Test tools`, which master never reaches (it is skipped after the build-tools failure). 4/54 tests fail there, none of them related to `output_normal.v`:

- `cmd/tools/vfmt_test.v` — the `FormatOptions{ ... backend: c" }` dump contains a stray `"`
- `cmd/tools/vgit-fmt-hook_test.v` — same vfmt family
- `cmd/v/macos_v3_external_fallback_test.v:70`
- `cmd/v/macos_v3_test.v:85` — expects `builder error: help doesn't exist`, gets `builder error:  doesn't exist` (the argument is dropped from the message)

These are pre-existing on master, previously masked behind the build-tools gate (the same way build-tools was masked behind the format gate). The other red checks (`vsl-compiles`, `vtl-compiles`, `v-up-works-nix` macos-arm64, `check-markdown`, `Contract worker`) also fail on master's own runs.

quaesitor-scientiam:
Filed the two newly-exposed \Test tools\ failures separately: #28511 (vfmt quote normalisation dropped by #28491, tests not updated) and #28512 (\macos_v3_*\ tests assume \@VEXE\ is the V3 dispatcher; on the \-old-compiler\ lanes it is \1_fallback\).

PR #28303, opened 2026-09-03
crypto.ecdsa: add mbedTLS backend as default, revert -d http3 gating

quaesitor-scientiam:
## Summary

Gives `crypto.ecdsa` a full mbedTLS-backed implementation alongside its
existing OpenSSL one, makes mbedTLS the **default** (matching the
`-d use_openssl` convention already established by `net.ssl` and
`net.http`'s `server_tls_*` files), reverts PR #28286's `-d http3` gating
on `net.http`'s HTTP/3 files now that the reason it existed is gone,
sweeps the resulting vestigial OpenSSL test gate out of `net.quic`'s own
test suite, and closes 2 real resource-safety bugs + a request.v
regression found by an adversarial-review pass.

`net.quic`'s TLS 1.3 stack (and therefore all of HTTP/3) previously
hard-required OpenSSL, because `crypto.ecdsa` — the only source of P-256
ECDH/ECDSA in this repo — was OpenSSL-only. That's what forced #28286 to
gate every `net.quic`-importing `net.http` file behind `-d http3`, so
plain HTTP/1.1/2 users (and `veb`) wouldn't need OpenSSL installed just to
build `net.http`. This PR removes that requirement at the source instead.

## Commit 1 — `crypto.ecdsa`: mbedTLS backend, made default

- `ecdsa.c.v/.v` → `ecdsa_d_use_openssl.c.v/.v`, `util.v` →
  `util_d_use_openssl.v` (renamed, zero content change): the existing
  OpenSSL implementation, now opt-in via `-d use_openssl`.
- `ecdsa_notd_use_openssl.c.v/.v`, `util_notd_use_openssl.v` (new): the
  mbedTLS-backed default. Self-contained C bindings (`crypto.*` never
  imports `net.*` in this repo) built on `mbedtls_ecp_export`'s one-call
  keypair extraction — no C shim needed.
- `new_key_from_seed` and PEM/DER loading are deferred with clear
  `-d use_openssl` stub errors — genuine behavioral divergences, not
  silently reimplemented.
- `.with_custom_hash` signing is explicitly rejected on the mbedTLS
  backend (RFC 6979 needs a real digest identifier its HMAC-DRBG can use,
  which an arbitrary `hash.Hash` doesn't carry). Verify has no equivalent
  problem.

## Commit 2 — `net.http`: revert `-d http3` gating

Reverts commit `28e3784751` file-for-file. `h3_server_d_http3.v` doesn't
exist on `master` yet, so nothing to revert there.

[...]

PR #28507, opened 2026-09-11
crypto.ecdsa: implement new_key_from_seed on the default mbedTLS backend (fix #28505)

### LINKED ISSUE #28505: crypto.ecdsa: new_key_from_seed is not implemented on the default mbedTLS backend, so a saved private key can't be loaded
### Describe the bug

Since #28303 made mbedTLS the default crypto.ecdsa backend, new_key_from_seed always returns an error unless you build with -d use_openssl. privkey_from_string isn't implemented on mbedTLS either, and on the default backend there's no way to load a private key you saved earlier.

I hit this with a server that saves its identity key: the first start generates and saves the key and every restart after that fails because the key can't be loaded again.

### Reproduction Steps

```v
import crypto.ecdsa

fn main() {
      key := ecdsa.PrivateKey.new(nid: .secp384r1)!
      seed := key.bytes()!
      ecdsa.new_key_from_seed(seed, nid: .secp384r1, fixed_size: true) or {
              println('reload failed: ${err}')
              return
      }
      println('reload ok')
}
```

### Expected Behavior

reload ok on the default backend without any flags.

### Current Behavior

reload failed: crypto.ecdsa: new_key_from_seed is not implemented for the default mbedTLS backend yet (OpenSSL and mbedTLS diverge on out-of-range-scalar handling here); build with -d use_openssl

### Possible Solution

The vendored mbedTLS 3.6.6 already has what's needed:

- mbedtls_ecp_read_key(grp_id, keypair, buf, len) reads the scalar, then checks it with mbedtls_ecp_check_privkey, which rejects anything outside [1, n-1];
- mbedtls_ecp_keypair_calc_public(keypair, f_rng, p_rng) computes the public point.

So new_key_from_seed could set up the context like PrivateKey.new does, call read_key instead of genkey and then calc_public with the existing ctr_drbg.

On the out-of-range divergence: anything produced by bytes() is always in range. Saved keys would reload the same on both backends. Out-of-range seeds from elsewhere would return an error on mbedTLS instead of matching OpenSSL which seems fine if it's documented.

_No response_

### V version

V 0.5.2 76c0d94

|V full version      |V 0.5.2 1f128c94619fb6e4863c5f81883e1b860405c8fa.59e148a // uh, idk why it's different
|Memory              |0.63GB/7.62GB
|                    |
|V last modified time|2026-09-11 16:54:16
|                    |
|Current working dir |OK, value: /home/scher/dev/vlang
|                    |
|V git status        |0.5.2-660-g76c0d946
|.git/config present |true
|                    |
|emcc version        |N/A

> [!NOTE]
>

quaesitor-scientiam:
Fixes #28505.

Since #28303 made mbedTLS the default `crypto.ecdsa` backend, `new_key_from_seed` always returned a "not implemented" error, so a private key saved via `PrivateKey.bytes()` could not be reloaded without `-d use_openssl`.

### Change
- Implement `new_key_from_seed` on the mbedTLS backend using the vendored mbedTLS 3.6 primitives the issue reporter pointed at: `mbedtls_ecp_read_key` (loads the group and range-checks the scalar to `[1, n-1]`) followed by `mbedtls_ecp_keypair_calc_public`. Keeps the OpenSSL backend's length checks, error messages and `fixed_size`/flexible `ks_flag`/`ks_size` contract, and frees the context on every error path.
- The one documented difference: mbedTLS rejects an out-of-range seed where OpenSSL performs no range check. Every seed `bytes()` produces is in range, so keys round-trip across both backends. `README.md` and the fn doc comment state this.
- Move the `new_key_from_seed` tests from `ecdsa_seed_use_openssl_test.v` (which forced `-d use_openssl`) back into `ecdsa_test.v` now that they are backend-neutral, and add:
  - `test_new_key_from_seed_fixed_size_roundtrip_all_curves` — the issue's exact save/reload scenario on all four curves, signing with the reloaded key and verifying with the original,
  - `test_new_key_from_seed_rejects_bad_sizes` — the length checks shared by both backends,
  - `test_new_key_from_seed_rejects_out_of_range_scalar` (mbedTLS-only, skipped under `-d use_openssl`) — zero and all-`0xff` seeds are rejected.

`privkey_from_string` (PEM loading), which the issue mentions in passing, is left as a separate follow-up.

### Testing
- The issue's repro prints `reload ok` with no flags.
- With the original stub restored, both `ecdsa_test.v` and `ecdsa_notd_use_openssl_test.v` fail; with the fix they pass (40/40 and 18/18 asserts).
- `v fmt -verify` and `v check-md vlib/crypto/ecdsa/README.md` clean.
- Not run locally: `util_test.v` / `example/ecdsa_seed_test.v` (`-d use_openssl`, no OpenSSL headers on this machine).


JalonSolov:
Local AI said:

**P1** `ecdsa_test.v:285` and `ecdsa_notd_use_openssl_test.v:97`: the new tests fail under `-d use_openssl` for P-521. OpenSSL serializes generated private scalars at minimal length, so a 65-byte scalar fails `fixed_size: true` (which requires 66 bytes), and the curve-size assertion fails. This is nondeterministic depending on the generated scalar, but blocks the OpenSSL test suite.

Validation:

* Default mbedTLS ECDSA suite: 7/7 passed.
* OpenSSL ECDSA suite: 5/7 passed, 2 failed.

quaesitor-scientiam:
> Local AI said:
> 
> **P1** `ecdsa_test.v:285` and `ecdsa_notd_use_openssl_test.v:97`: the new tests fail under `-d use_openssl` for P-521. OpenSSL serializes generated private scalars at minimal length, so a 65-byte scalar fails `fixed_size: true` (which requires 66 bytes), and the curve-size assertion fails. This is nondeterministic depending on the generated scalar, but blocks the OpenSSL test suite.
> 
> Validation:
> 
> * Default mbedTLS ECDSA suite: 7/7 passed.
> * OpenSSL ECDSA suite: 5/7 passed, 2 failed.

This is a pre-existing bug in the OpenSSL backend that predates my changes.  Looking into it

quaesitor-scientiam:
Thanks — confirmed and fixed in 44d01f4af4.

**Triage:** the claim holds, and it was pointing at a pre-existing bug in the OpenSSL backend rather than at the tests. `ecdsa_d_use_openssl.v`'s `PrivateKey.bytes()` sized a `.fixed` key's output from `BN_num_bits()` (the scalar's minimal length) instead of the curve's width, so ~1/2 of P-521 keys and 1/256 of P-256 keys came back one byte short. Beyond failing these tests, that meant the exact #28505 scenario (save with `bytes()`, reload with `fixed_size: true`) would fail nondeterministically on the OpenSSL backend too, and `bytes()` disagreed with the mbedTLS backend's for the same key. (`test_generate_key_for_every_curve` at `ecdsa_notd_use_openssl_test.v:97` predates this PR — #28303 — and had the same flake.)

**Fix:** a `.fixed` key's `bytes()` now sizes from `EVP_PKEY_get_bits()` (32/48/66), zero-padded by the existing `BN_bn2binpad`. `privkey_from_string` keys are `.fixed` too, so they get the same stable width.

**Reproduction / verification (WSL, `-d use_openssl`):**
- Before the fix: `ecdsa_test.v` failed 3/3 runs with `seed size doesnt match with curve key size`; `ecdsa_notd_use_openssl_test.v` failed 1/3 with `curve secp521r1 should produce a 66-byte private key`.
- Added a deterministic backend-neutral test (`test_new_key_from_seed_fixed_size_leading_zero_seed_keeps_width`: a fixed-size seed with two leading zero bytes on every curve). It fails on the old `bytes()` (`bytes() must be 32 wide on prime256v1, got 30`) and passes after.
- After the fix: `ecdsa_test.v` + `ecdsa_notd_use_openssl_test.v` under `-d use_openssl` passed 6/6 consecutive runs; the whole `vlib/crypto/ecdsa/` dir under `-d use_openssl` passed 3/3 runs, and `util_test.v` / `example/ecdsa_seed_test.v` (built and run directly since the `present_openssl?` vtest gate skipped them on my box) pass; default mbedTLS suite passes.

JalonSolov:
Everything looks good now.  Waiting for CI to finish...

quaesitor-scientiam:
CI is red across the board, but none of it is this PR's doing

PR #28508, opened 2026-09-11
net.openssl: honor read_timeout by making the client socket non-blocking (fix #28506)

### LINKED ISSUE #28506: net.http with -d use_openssl: HTTPS request sometimes hangs about 90s after the response arrived; read_timeout doesn't apply
### Describe the bug

With -d use_openssl, a plain http.fetch to an HTTPS endpoint sometimes takes about 90 s even though the whole response has arrived. The same request on the default mbedTLS backend takes about 300 ms every time.

### Reproduction Steps

```v
import net.http
import time

fn main() {
      url := 'https://client.discovery.minecraft-services.net/api/v1.0/discovery/MinecraftPE/builds/1.26.30'
      start := time.now()
      resp := http.fetch(url: url, method: .get, read_timeout: 20 * time.second, write_timeout: 20 * time.second)!
      println('${resp.status_code} ${resp.body.len} bytes in ${time.since(start).milliseconds()}ms')
}
```

Run it a few times with v -d use_openssl run.

### Expected Behavior

fetch returns as soon as the response is complete, the same way it does on the default mbedTLS backend (about 300 ms here). And if a read really does stall, read_timeout (20 s in the example) should end it with an error instead of the call blocking until the server closes the connection.

### Current Behavior

Some runs take 90,221 / 90,256 / 90,287 ms; others are under 300 ms. It never took long on the default backend. The 20 s read_timeout doesn't stop the wait.

strace of a slow run: the full response is read by 18:28:12. Then a background thread sits in a plain blocking read() on the socket for 89.85 s, until the server sends one more record at 18:29:42. Only then does fetch return. No poll/select comes before that read.

### Possible Solution

_No response_

In vlib/net/openssl/ssl_connection.c.v, socket_read_into_ptr has the wait before SSL_read commented out.

### V version

V 0.5.2 76c0d94

|V full version      |V 0.5.2 1f128c94619fb6e4863c5f81883e1b860405c8fa.59e148a
|Memory              |0.63GB/7.62GB
|                    |
|V last modified time|2026-09-11 16:54:16
|                    |
|Current working dir |OK, value: /home/scher/dev/vlang
|                    |
|V git status        |0.5.2-660-g76c0d946
|.git/config present |true
|                    |
|emcc version        |N/A

> [!NOTE]
>

quaesitor-scientiam:
Fixes #28506.

## What was happening

With `-d use_openssl`, the client socket stayed **blocking**, so `SSL_read()` sat in `read()` and never returned `SSL_ERROR_WANT_READ`. The backend's timeout logic only runs in the `WANT_READ` → `wait_for_read(deadline)` retry loop, so `set_read_timeout()` had no effect on reads at all: a read on a silent peer blocked until the server closed the connection.

A plain https `http.fetch()` hits this through the pooled HTTP/2 transport: its reader thread (the "background thread" in the issue's strace) parks in a blocking `SSL_read` while holding `io_mu`, and every writer — including the request whose response had already fully arrived — waits behind it until the server's idle timeout (~90 s). Whether the *first* fetch hangs is a race between the request thread's last write and the reader re-entering `SSL_read`, hence "sometimes". mbedTLS is unaffected because its recv callback (`mbedtls_net_recv_timeout`) enforces `mbedtls_ssl_conf_read_timeout`.

## Changes

- `net.openssl.SSLConn.connect` switches the socket to non-blocking (and keeps `TcpConn.is_blocking` in sync). The existing `WANT_READ`/`WANT_WRITE` loops in `complete_connect`/`socket_read_into_ptr`/`write_ptr`/`shutdown` were already written for this; they just never engaged.
- `net.openssl.ssl_remaining_timeout` clamps an elapsed deadline to 1 ns instead of returning `<= 0`, which `select()` interprets as *infinite* — mirrors the `net.mbedtls` helper (#27429), which never got ported.
- `select()` in **both** TLS backends: the `EINTR` retry branch was dead code (`net.socket_error(...)!` propagated first) and the `fd_set` wasn't re-armed between attempts. This became reachable once the OpenSSL read path waits in `select()` while other threads run (GC signals) — it showed up as `net: socket error: 4` connection failures in `transport_h2_test.v -d use_openssl` before this fix.
- New hermetic regression test `vlib/net/ssl/ssl_read_timeout_test.v`: local mbedTLS server answers once then stays silent; the client (whichever backend `net.ssl` is built with) must get its data and then time out in ≪ 6 s. Fails at exactly 6.0 s (server close) on the old OpenSSL code, passes with the fix on both backends.

## Verification

- Issue repro (`http.fetch` to the reported endpoint, 6×, `-d use_openssl`): before — run 1 hung >10 min; after — 12/12 runs in 113–662 ms with the pooled h2 connection reused.
- `v -d use_openssl test vlib/net/ssl vlib/net/openssl`, `v test vlib/net/ssl vlib/net/mbedtls vlib/net/openssl`, `v test vlib/net/http`: all pass.
- `v -d use_openssl vlib/net/http/transport_h2_test.v`: hung/timed out on master; now completes with one pre-existing, unrelated failure (`test_h2_pool_dial_failure_preserves_error_code` — `net.dial_tcp`'s aggregate error carries code 0 under the OpenSSL backend).

### DIFF
--- a/vlib/net/http/h2_pooled_transport.v
+++ b/vlib/net/http/h2_pooled_transport.v
@@ -15,6 +15,13 @@ import time
 // (see its doc comment for why), so this is the ONLY bound on how long an
 // idle reader can hold io_mu and make a writer wait — kept short for exactly
 // that reason, not because a longer internal retry would be incorrect.
+// This bound only holds if the backend's read really does return on timeout:
+// mbedTLS gets it from mbedtls_ssl_conf_read_timeout on its recv callback, and
+// OpenSSL from its socket being non-blocking (SSLConn.connect switches it), so
+// SSL_read yields WANT_READ instead of blocking in read(). Before that switch a
+// reader parked in a blocking SSL_read held io_mu until the peer sent anything
+// at all, stalling every writer — including the request that had already
+// received its full response — for the server's idle timeout (vlang/v#28506).
 //
 // Both mbedTLS and OpenSSL's write_ptr (vlib/net/mbedtls/ssl_connection.c.v,
 // vlib/net/openssl/ssl_connection.c.v) reuse this SAME configured duration for
--- a/vlib/net/mbedtls/ssl_connection.c.v
+++ b/vlib/net/mbedtls/ssl_connection.c.v
@@ -996,8 +996,6 @@ fn select(handle int, test Select, timeout time.Duration) !bool {
 		eprintln('${@METHOD} handle: ${handle}, timeout: ${timeout}')
 	}
 	set := C.fd_set{}
-	C.FD_ZERO(&set)
-	C.FD_SET(handle, &set)
 
 	is_infinite := timeout <= 0 || timeout == net.infinite_timeout
 	deadline := ssl_timeout_deadline(timeout)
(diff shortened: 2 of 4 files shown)


quaesitor-scientiam:
CI status for [#28508](https://github.com/vlang/v/pull/28508): 42 failing, but none caused by the PR.

@JalonSolov  anything blocking this one?

quaesitor-scientiam:
CI triage — none of the red checks are caused by this PR:

- **All the `*-linux` / `*-macos` / `*-windows` / docker / tools-* / code-formatting jobs** fail in the same steps on master's own last two runs (`d622f33b52`, `37684721b5`): `v test-cleancode` (`v fmt -verify` on 27 files under `vlib/v/gen/c`, `vlib/v3`, `vlib/crypto/scram`, … — none touched here), tcc `implicit declaration of '__atomic_exchange_4'` in the generated C, musl `PTHREAD_RWLOCK_PREFER_WRITER_NONRECURSIVE_NP` / `backtrace_symbols` conflicts. Same failures with or without this branch.
- **h2spec** (`6.9.2 :: Sends a SETTINGS frame for window size to be negative` → `tls: DialWithDialer timed out`) is deterministic on the 4-core runner but pre-existing: it reproduces locally with `VJOBS=4` on master's own `net.mbedtls`/`net.openssl` files, and passes with `VJOBS=8`. It is worker starvation in the TLS server (the four h2spec cases that open a preflight connection pin all handler threads), not a timing flake and not related to the backend changes here. Filed as #28517 with the repro.

The PR's own coverage: `v -d use_openssl test vlib/net/ssl vlib/net/openssl`, `v test vlib/net/ssl vlib/net/mbedtls vlib/net/openssl vlib/net/http` all pass locally on both backends, and the new `ssl_read_timeout_test.v` fails on master's OpenSSL backend / passes here.

medvednikov (commented):
Reviewed the code changes only (CI status intentionally not considered). I did not find a concrete correctness issue. The OpenSSL client socket is switched to non-blocking before the handshake, and the existing connect/read/write/shutdown retry loops all handle WANT_READ/WANT_WRITE through the deadline-aware wait helpers. Re-arming `fd_set` on EINTR and recomputing the remaining deadline also looks correct.

PR #28531, opened 2026-09-12
v3: discard cache entries the C toolchain rejects, and retry

medvednikov:
Follow-up to #28529, which fixed one way a cache entry gets poisoned. This makes v3 recover from the *class* of problem instead of only that instance.

## Problem

A poisoned cache entry is permanent. It outlives the build that published it, nothing revalidates it, and the only symptom is a C toolchain error naming a file the user never mentioned:

```
tcc: error: /tmp/v_501/v3_thirdparty_objs/atomic_<sig>.o: unrecognized file type
```

For a tcc build that error then triggers the `cc` fallback, so the build still *succeeds* — it just got several times slower, permanently, behind a one-line `warning: tcc compilation failed, falling back to cc`. That is how #28529 went unnoticed.

## What this does

Detect the case and repair it.

**Detection requires two independent signals.** The output has to name a path under one of V's persistent caches (`v3_module_cache_*`, `v3_thirdparty_objs`, `v3_fastc_unit_cache`, `v3_fastc_link_cache`, `v3_crun_cache`) **and** report a whole-file rejection — unreadable, missing, or duplicated symbols. Requiring both keeps ordinary errors out:

- a line-scoped diagnostic inside a cached unit is a real compile error, and is reported as one;
- `ld: file not found: /usr/local/lib/libfoo.a` is the user's to fix;
- the build directory is not a cache, so generated source never qualifies.

**Repair discards the sidecars too.** A stamp still certifies a deleted object, and a link plan replays the object path into the next link even after the object is gone, so both go with it.

**Then the build restarts once**, guarded by an environment variable so it cannot loop. Recovery runs *before* the tcc fallback, so a repairable cache gets repaired rather than hidden behind a slower compiler.

**If the retry hits the same rejection**, the entries are named and `v wipe-cache` is suggested — including on the fallback path, where the build still succeeds and the problem is otherwise invisible.

## Behaviour

Poisoned entry, before:

```
$ v -cc tcc -o hello hello.v
warning: tcc compilation failed, falling back to cc
```

After:

```
$ v -cc tcc -o hello hello.v
warning: the C toolchain rejected cached V build artifacts; discarding them and retrying:
  /tmp/v_501/v3_thirdparty_objs/atomic_9c646e209acb927_wrapv.o
```

…and the build completes with tcc. `v self` with a poisoned cache self-repairs the same way.

If a rebuild republishes the same bad entry:

```
warning: tcc compilation failed, falling back to cc
warning: tcc rejected cached V build artifacts that a rebuild did not repair:
  /tmp/v_501/v3_thirdparty_objs/atomic_9c646e209acb927_wrapv.o
Suggestion: run `v wipe-cache`, then repeat your compilation.
```

## Verification

- On this branch alone (without #28529) a deliberately poisoned atomic object is detected, discarded and rebuilt, and the build finishes on tcc — so this would have fixed the original `v self` report by itself.
- Negative cases confirmed not to trigger it: a checker error, a C link error for an undeclared function (cache file count unchanged, no suggestion printed), and paths in the build directory.
- Loop guard confirmed: with the guard set, recovery declines, nothing is deleted twice, and the escalation message names the entries.
- New unit tests in `vlib/v3/driver/c_compiler_flags_test.v` cover detection, path extraction across toolchain quoting styles, and sidecar/link-plan removal. All 9 `vlib/v3/driver/*_test.v` files pass.
- `v self` (cold and warm), `-prod`, threaded/`sync.stdatomic` programs, and `v wipe-cache` all verified working.

### DIFF
--- a/vlib/v3/driver/driver.v
+++ b/vlib/v3/driver/driver.v
@@ -9549,10 +9549,16 @@ pub fn run(args []string) {
 				os.rm(fastc_bin_file) or {}
 			}
 			if !fastc_result.success {
+				if v3_recover_from_cache_failure(args, fastc_result.output, '') {
+					return
+				}
 				if fastc_result.command.len == 0 {
 					eprintln('fastc requires the bundled TinyCC executable')
 				} else if !show_c_output && fastc_result.output.len > 0 {
 					eprintln(fastc_result.output.trim_space())
+				}
+				if v3_cache_failure_artifacts(fastc_result.output).len > 0 {
+					eprintln("Suggestion: the C toolchain keeps rejecting files from V's cache. Run `v wipe-cache`, then repeat your compilation.")
 				}
 				exit(1)
 			}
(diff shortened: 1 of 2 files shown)


vlib/v3/driver/driver.v
```diff
@@ -12505,6 +12531,162 @@ fn v3_is_tcc_compilation_failure(c_compiler string, output string) bool {
...
+	}
+	prefixes := v3_cache_artifact_prefixes()
+	mut artifacts := []string{}
+	for raw in output.fields() {
+		token := raw.trim('\'"`()[],;:')
+		if token.len == 0 || token in artifacts {
+			continue
+		}
+		for prefix in prefixes {
+			if token.starts_with(prefix) {
```
medvednikov, line 12583: This textual prefix test is not sufficient before deleting the returned path. A diagnostic token such as `/tmp/v3_thirdparty_objs_backup/user.o` or `/tmp/v3_thirdparty_objs/../../home/me/user.o` passes `starts_with(prefix)`, and `v3_discard_cache_artifacts()` then calls `os.rm` on it. If a user-supplied linker input with such a spelling is rejected with one of the markers, the compiler can delete a file outside V's cache. Canonicalize the candidate and cache roots, require a path-component boundary/descendant relationship, and only return paths proven to be inside an owned cache directory.

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I confirm a blocking safety issue remains in the cache-artifact ownership check.

PR #28532, opened 2026-09-12
v3: expand environment macros in C directives

antono3:
## What does this PR do?

Expand `$env(...)` macros in V3 C flags and include directives before argument splitting and pseudo-path resolution. This restores the behavior already supported by the default compiler and lets modules locate SDK headers and libraries through environment variables.

The regression test covers both `#flag -I$env(...)` and `#include "$env(...)/..."` processing.

## Verification

- Built the current V3 compiler directly without preallocation.
- `./v3np vlib/v3/gen/c/target_test.v`
- Compiled and ran a small C-interoperability program whose header is reachable only through `#flag -I$env('V3_ENV_HEADER_DIR')`.
- Ran all 20 `vlib/v3/gen/c` tests on the original reproducing V3 revision before rebasing the patch onto current `master`.

### DIFF
--- a/vlib/v3/gen/c/cleanc.v
+++ b/vlib/v3/gen/c/cleanc.v
@@ -11,6 +11,7 @@ import v3.modulecache
 import v3.pref
 import v3.types
 import v3.util
+import v.util as vutil
 
 const spread_index_expected_type_marker = '__v3_spread_index_expected_type'
 const source_mut_pointer_deref_marker = '__v3_source_mut_pointer_deref'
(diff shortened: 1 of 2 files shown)


guweigang:
I need this feature back!  @medvednikov

medvednikov (commented):
Reviewed the code changes only (CI status not considered). I did not find a concrete correctness issue. Environment expansion is applied after target/comment/default-macro processing but before flag tokenization and pseudo-path resolution, and include directives preserve their quoting. Missing/empty variables follow `resolve_env_value(..., true)` and cause the directive to be omitted rather than emitting a literal unresolved macro.

medvednikov:
Fixed in https://github.com/vlang/v/pull/28725

due to the conflicts

medvednikov:
Please confirm it works for you now.

guweigang:
> Please confirm it works for you now.

@medvednikov  works for me, thank you.

PR #28536, opened 2026-09-12
cgen: look up map-of-map values through a helper, not a statement expression

medvednikov:
`makev.bat -msvc` has failed to build V since #28466, at the final `v_stage.exe -cc msvc -o ./v_up.exe cmd/v` step:

```
checker_tail.v(9962): error C2059: syntax error: '{'
checker_tail.v(9962): error C2065: '_t8': undeclared identifier
checker_tail.v(9964): error C2065: 'label_aliases': undeclared identifier
... every later name in the function, for ~100 more lines
```

#28466 made a map-of-map index build its empty-map default lazily, so that a lookup which finds its key no longer constructs and leaks a map header. It did that with `({ ... })` — a GNU statement expression, which MSVC does not implement. One `({` derails cl.exe's parser for the rest of the function, which is why a single construct produced a hundred unrelated errors.

(The MSVC lane's *Build* step was green as recently as 09-08; it has been failing on this since #28466 landed.)

## Change

Emit the same logic as a generated helper, registered per value type the way sumtype casting fns already are, so the default is still built only on a miss:

```c
Map_string_Array_int _v_map_get_or_zero_Map_string_Array_int(map* m, void* key) {
	void* v = builtin__map_get_check(m, key);
	if (v) { return *(Map_string_Array_int*)v; }
	Map_string_Array_int zero = new_map(...);
	return zero;
}
```

The mutating form returns `ValT*` and keeps the `map_get_and_set` insert.

This was cgen's only remaining statement expression: the C emitted for `-cc msvc -os windows -o cmd/v` now contains none outside string literals (the 8 remaining hits are V3 cgen's own source text compiled into the binary).

## Generated C is otherwise unchanged

Normalising temp-variable numbering — the old form burned three `_tN` per lookup, the new one none — every line the diff removes is one of these statement expressions and nothing else:

```
$ diff <(sed -E 's/_t[0-9]+/_tN/g' before.c) <(sed -E 's/_t[0-9]+/_tN/g' after.c) \
    | grep '^<' | grep -vc '(\{ map\* _tN'
0
```

## Testing

- New `vlib/v/gen/c/map_value_lookup_msvc_codegen_test.v`; it fails on master (`assert !generated.contains('({ map*')`) and passes with the change.
- V self-hosts through the changed path on macOS.
- All 13 map / nested-map test files pass.
- 68/70 of the first `vlib/v/tests` slice pass; the two `autofree_*` failures are identical before and after.

I do not have MSVC to run the Windows lane locally, so this is verified through the generated C and the non-Windows suites — the MSVC lane itself will be the real check.

### DIFF
--- a/vlib/v/gen/c/cgen.v
+++ b/vlib/v/gen/c/cgen.v
@@ -249,6 +249,7 @@ mut:
 	array_get_types                      []ast.Type
 	auto_fn_definitions                  []string // auto generated functions definition list
 	sumtype_casting_fns                  []SumtypeCastingFn
+	map_value_lookup_fns                 []MapValueLookupFn // lazily-defaulted map-of-map lookups
 	anon_fn_definitions                  []string // anon generated functions definition list
 	anon_fns                             shared []string // remove duplicate anon generated functions
 	sumtype_definitions                  map[string]bool // `_TypeA_to_sumtype_TypeB()` fns that have been generated
@@ -582,6 +583,18 @@ pub fn gen(files []&ast.File, mut table ast.Table, pref_ &pref.Preferences) GenO
 					global_g.sumtype_casting_fns << scf
 				}
 			}
+			for mvl in g.map_value_lookup_fns {
+				mut already_exists := false
+				for existing in global_g.map_value_lookup_fns {
+					if existing.fn_name == mvl.fn_name {
+						already_exists = true
+						break
+					}
+				}
+				if !already_exists {
+					global_g.map_value_lookup_fns << mvl
+				}
+			}
 
 			for at in g.array_typedefs {
 				if at !in global_g.array_typedefs {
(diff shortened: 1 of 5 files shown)


medvednikov:
Pushed a follow-up: `coutput_test.v` and `map_value_lookup_lazy_default.c.must_have` pinned the statement-expression *spelling* (`*ADDR(..., ({`, the `_tN` carriers, `_t4 ? *((Map_string_int*)_t4) : ...`), so replacing it made them fail even though the property they exist to protect is unchanged.

They now assert that property against the new shape — a hit returns the stored map, only a miss builds the default, nothing memdups a carrier — and additionally pin the absence of `({ map*`, so the MSVC regression cannot come back unnoticed.

I found this by cherry-picking all four of my open CI fixes onto current master (3a1a3cfcf3) and running a combined sweep, which is also worth reporting: the four apply cleanly to each other and to master, V self-hosts with all of them, and the only failures left in that sweep are pre-existing on clean master — `autofree_discarded_call_return_value_test.v`, `autofree_labeled_continue_scope_test.v`, and two crashers (`closure_with_fn_ref_var_test.v` exit 138, `comptime_generic_comptime_variant_test.v` exit 139), all reproduced with a compiler built from untouched `origin/master`.

PR #28537, opened 2026-09-12
cmd, v3: forward the child's exit code when V re-executes itself on Windows

medvednikov:
The wasm lane's `scalar_option_result_error_test.v` fails on windows-2022 while passing on ubuntu, and its own assertion message shows why:

```
expected a compile error, got: .../scalar_option.v:1:8: error: option types are not implemented
```

The error is there. The exit code is 0.

Those tests run `v -old-compiler -b wasm ...`, which hands the build to the separately built V1 compatibility compiler through `os.execvp`. On POSIX that replaces the process, so the compiler's status is the one the caller sees. **Windows has no exec**: the CRT's `_execvp` starts the child and then terminates the calling process with status 0, so a failed compilation reports its error and still looks like a success to everyone upstream — `os.execute` in the test suite, `make`, a shell.

## Change

Run the child through `os.system` on Windows and exit with its status, the way `v.util`'s tool launcher (`cmd_system`/`tool_launch_command`) already sidesteps the same problem for external tools.

The two V3 driver restart paths had the same hole and are fixed alongside:
- the `-d v3_parallel_cc_monolithic` regeneration restart, and
- the generic restart, which only special-cased the JS backend.

POSIX still execs and is unchanged.

## Testing

- `cmd/v/v1_fallback_exit_code_test.v` asserts both directions — a broken build must not report success, a good one must not report failure — through the same dispatch the wasm tests use, so the property is checked on every host rather than only where the wasm backend runs.
- `vlib/v/gen/wasm/tests/scalar_option_result_error_test.v` passes locally on macOS (it already did there; Windows is the lane this unblocks).
- `cmd/v` test files: 6/6 pass.

I do not have Windows to run that lane locally, so the mechanism is established from the CI output plus the CRT's documented `_execvp` behaviour; the Windows lanes will be the real check.

### DIFF
--- a/cmd/v/macos_v3_dispatch.c.v
+++ b/cmd/v/macos_v3_dispatch.c.v
@@ -129,11 +129,26 @@ fn launch_macos_v1_fallback(executable string, args []string, is_verbose bool, r
 	if is_verbose || os.getenv('V3_CACHE_TRACE') != '' {
 		eprintln('${reason}; retrying with `${executable}`.')
 	}
-	os.execvp(executable, macos_v1_fallback_args(args)) or {
+	fallback_args := macos_v1_fallback_args(args)
+	$if windows {
+		// Windows has no exec. The CRT's `_execvp` starts the child and then
+		// terminates this process with status 0, so the compatibility
+		// compiler's exit code never reaches whoever ran `v` - a failed
+		// compilation looks like a success to every caller, including
+		// `os.execute` in the test suite. Run it as a child and forward its
+		// status, the way `v.util`'s tool launcher already does.
+		exit(os.system(exec_command(executable, fallback_args)))
+	}
+	os.execvp(executable, fallback_args) or {
 		eprintln('failed to launch the V1 compatibility compiler `${executable}`: ${err}')
 		exit(1)
 	}
 	exit(1)
+}
+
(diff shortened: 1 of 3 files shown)


medvednikov:
More evidence, from the Windows GCC lane's `Win32 W5 raw-input A1 public GREEN` step — note the two summaries:

```
 FAIL     0.042 ms     1 assert  | main.test_w5_a1_controlled_failure_exit()
 Summary for running V tests in "w5_a1_controlled_failure_test.v": 1 failed, 1 total.
...
Summary for all V _test.v files: 1 passed, 1 total.
```

The test binary's own runner reports **1 failed**. `cmd/tools/vtest`, one level up, records **1 passed** — it compiles-and-runs each `_test.v` through `v`, and that invocation goes out through the same V1-compatibility dispatch, so on Windows it returns 0 whatever the test did. The step then fails only because the PowerShell probe deliberately injects a controlled failure and refuses to believe the success it gets back.

So this is not just the wasm lane: **on Windows, a failing `_test.v` can be counted as passing**, and anything shaped "assert the compiler rejects this" is vacuous. Whatever else is red there, those results are not trustworthy until the dispatch forwards the child's status.

The three `failed CreateProcess: The system cannot find the file specified.` aborts in the same run's `gcc-windows / Test v binaries` step are a separate problem, handled in #28538.

PR #28538, opened 2026-09-12
v3: make a command that cannot start a failing result, not a hard abort

medvednikov:
The Windows TCC lane now gets all the way to `Test new v.c` (formatting and vet pass again) and dies there:

```
Run v -o v.c cmd/v
failed CreateProcess: The system cannot find the file specified.
tcc: error: file 'v.c' not found
```

`os.Process.start()` reports a failed `CreateProcess` through `failed_cfn_report_error`, which prints that line and `exit(1)`s. So on Windows **any** command V3 cannot find takes the whole compiler down, and the message does not even say which command it was — the second error is only the next line of the workflow tripping over the missing output.

POSIX never behaves that way: a failed exec happens in the forked child, the parent sees a non-zero status, and `cmdexec` already turns that into a result (it even appends `executable: ...` when the output does not name it). Probes written against that — the `#pkgconfig` existence check in the parser, the TCC usability check in the driver — quietly fall back on Linux and macOS.

## Change

Decide it in `cmdexec` instead of asking `os.Process`: resolve the program first — against the work folder for a relative path, through PATH for a bare name — and return exit code 127 with `failed to find executable: <program>` when it cannot start.

This does not change what V3 runs, only what happens when it is not there. Callers that probe fall back the way they already do on POSIX; a command that really was required now names itself in the log.

## Honest scope

I could not reproduce the Windows lane locally (no Windows), and I could not pin down *which* command is missing there: tracing `v -cc tcc -no-retry-compilation -o out.c cmd/v` on Linux shows only the bundled `tcc -v` probe and `ldd --version` (the latter is `@[if linux]`), both of which resolve.

So this is deliberately the fix for the *abort*, not a claim to have found the missing program. If the spawn was a probe, the lane goes green; if it was required, the next run says `failed to find executable: X` instead of an unattributable `failed CreateProcess`, which is what is needed to finish the job.

## Testing

- `vlib/v3/cmdexec/missing_executable_test.v` pins the contract for a missing bare name and a missing absolute path, and that a present command still runs. On POSIX it passes before and after (POSIX already degraded gracefully); the guard it adds is for Windows, where the old path could not return at all.
- `vlib/v3/cmdexec` tests: 3/3. `vlib/v3/gen/c` slice: 12/12. V self-hosts.


vlib/v/cmdexec/cmdexec.v
```diff
@@ -29,8 +29,150 @@ pub fn run_with_timeout(program string, args []string, timeout_ms i64) os.Result
...
+	if os.is_file(candidate) && os.is_executable(candidate) {
+		return candidate
+	}
+	if os.file_ext(candidate) != '' {
+		// os.is_executable only accepts the conventional extensions, but
+		// CreateProcessW runs any module it can load - including a PE binary
+		// deliberately named `clang.bin`. Existence is the honest test here.
+		return if os.is_file(candidate) { candidate } else { none }
+	}
+	for suffix in windows_implicit_suffixes {
```
medvednikov, line 153: This `.exe` fallback does not mirror the current `os.Process` launch semantics. `process_windows.c.v` makes the filename absolute and passes it as non-null `lpApplicationName` for non-batch paths; Microsoft documents that `lpApplicationName` must include the extension and **no default extension is assumed**. So an extensionless `C:\tools\cc` is tried exactly today, whereas this preflight can silently choose `C:\tools\cc.exe` (and even prefers that sibling over an existing extensionless PE because the exact-file fallback is below this loop). That violates the PR's stated goal of changing only failure handling, not what runs. Probe/accept the exact filename first and avoid suffix inference unless `os.Process` itself is changed to use command-line lookup semantics.

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I found one remaining Windows resolution mismatch that can change which executable is launched.

PR #28539, opened 2026-09-12
tools: check-md: use only the first fence info-string token as the language

quaesitor-scientiam:
Follow-up to #28498, addressing the review note there.

### Problem

The non-V fence check in `parse_line` compared the **entire** info string against `non_v_fence_languages`, so a fence with trailing metadata was still treated as a V example — and then split on whitespace into multiple bogus commands:

```
$ v check-md meta.md      # containing ```vml title=example
meta.md:3:1: error: unrecognized command: "ml", ...
meta.md:3:1: error: unrecognized command: "title=example", ...
Checked .md files: 1 | Ex.: 1 | ... | Errors: 2 | Fmt errors: 2
```

Per CommonMark, only the first word of a fence's info string is the language; the rest is arbitrary metadata.

### Fix

Add `fence_language()` which returns the first info-string token, and use it for the `non_v_fence_languages` check. The existing V-command path is untouched.

### Tests

New `test_check_md_vml_fence_with_info_string_metadata_is_not_a_v_example`, exercising ```` ```vml title=example ```` followed by a real ```` ```v ```` example. Fails on master with the exact error above; passes with the fix (`Ex.: 1`, only the V example counted). `v -silent fmt -inprocess -verify` clean on both files; full `vcheck_test.v` passes.

### DIFF
--- a/cmd/tools/vcheck-md.v
+++ b/cmd/tools/vcheck-md.v
@@ -329,6 +329,12 @@ const default_command = 'compile'
 // fence languages that begin with `v` but hold something other than V source
 const non_v_fence_languages = ['vml']
 
+// fence_language returns the first token of a code fence's info string.
+fn fence_language(line string) string {
+	fields := line.replace('```', '').fields()
+	return if fields.len > 0 { fields[0] } else { '' }
+}
+
 struct VCodeExample {
 mut:
 	text    []string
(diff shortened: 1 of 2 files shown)


medvednikov (commented):
Reviewed the code changes only (CI status not considered). I did not find a correctness issue. The helper is only used to classify known non-V fence languages, while the existing V fence command syntax remains unchanged; the regression test directly covers the metadata case that previously misclassified `vml`.

PR #28540, opened 2026-09-12
v3: fix the V1 bootstrap build of the transform worker scopes

medvednikov:
Master tip (`4089ece6a4`, #28534) does not `make`: stage 2 (`./v1 -no-parallel -o v2 -gc none cmd/v`, i.e. the V1 checker from the `vc` snapshot) rejects two new source forms in `vlib/v3/transform`.

```
vlib/v3/transform/transform_parallel_notd_v3_no_parallel.v:207:9: error: assignment mismatch: 1 variable 0 values
vlib/v3/transform/transform_parallel_notd_v3_no_parallel.v:216:3: error: `scope` used as value
vlib/v3/transform/transform_parallel_notd_v3_no_parallel.v:239:9: error: assignment mismatch: 1 variable 0 values
vlib/v3/transform/transform_parallel_notd_v3_no_parallel.v:246:3: error: `scope` used as value
vlib/v3/transform/transform_parallel_notd_v3_no_parallel.v:346:23: error: assignment mismatch: 2 variables 1 value
...
```

Two V1 limitations are involved:

* an `if` expression whose else branch is `unsafe { nil }` is typed `void`, so `scope := if cond { transform_worker_scope_begin(true) } else { unsafe { nil } }` fails and every later use of `scope` follows;
* a multi-return call wrapped in an `unsafe` *expression* counts as one value, so `scope_lo, scope_hi := unsafe { prealloc_scope_address_range(scope) }` fails (this one surfaces once the `$if prealloc` branch is checked, e.g. a `-prealloc` build).

Fixes:

* `transform_worker_scope_begin` already returns nil when `enabled` is false, so the conditional call collapses to passing the flag through — same behaviour, no `void`-typed `if` expression.
* the multi-return call moves into a `transform_scope_address_range` helper beside the existing `transform_scope_owns`, which keeps the `$if prealloc` gate and the `unsafe` block in one place and uses the statement form V1 accepts.

Generated code is unchanged (checked the emitted C for `transform__transform_scope_address_range`).

Verified on macOS/arm64 from a clean worktree at `4089ece6a4`:

* `make` fails before the patch, succeeds after (v1 → v2 → v1_fallback → v, no errors or warnings);
* `./v1 -no-parallel -prealloc -o ... cmd/v` reproduces the full error list before, is clean after;
* v3 self-host generation chain (`v` → `v_new` → `v3gen` → `v4gen`) passes, plus a `-prealloc` build of `cmd/v`;
* `vlib/v3/transform/{scoped_merge_notd_v3_no_parallel,ast_snapshot}_test.v` pass; `transform_parallel_test.v` fails identically before and after (pre-existing assert at `:919` and the segfault that follows it).

### DIFF
--- a/vlib/v3/transform/transform.v
+++ b/vlib/v3/transform/transform.v
@@ -4457,6 +4457,19 @@ fn transform_scope_owns(scope voidptr, ptr voidptr) bool {
 	return false
 }
 
+// transform_scope_address_range returns the union address range of the blocks
+// owned by `scope`, so callers can reject most pointers before the per-block
+// search in transform_scope_owns.
+fn transform_scope_address_range(scope voidptr) (usize, usize) {
+	$if prealloc {
+		unsafe {
+			lo, hi := prealloc_scope_address_range(scope)
+			return lo, hi
+		}
+	}
+	return 0, 0
+}
+
 // clone_scoped_worker_node publishes a node's owned fields through the
 // compilation text table before its helper arena is released.
 fn (mut t Transformer) clone_scoped_worker_node(idx int, scope voidptr) {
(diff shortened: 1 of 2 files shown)


medvednikov:
it's v1 being stupid

PR #28544, opened 2026-09-12
toml: avoid quadratic string accumulation

davlgd:
Parsing a document whose value is a long string is quadratic. Three places accumulate one byte at a time:

- `scanner.extract_string` and `scanner.extract_multiline_string`, with `lit += c.ascii_str()`,
- `decoder.decode_quoted_escapes`, with `decoded_s += …`, which already carried a `// TODO: use string builder`.

`extract_string` added a second O(n²) on top: it re-evaluated `lit.contains('\n')` on every byte, walking the whole literal each time.

This matters on real documents. The case that surfaced it is a corpus of more than a hundred TOML manifests with a base64 image payload, the largest ones 100 to 130 KB.

| | before | after |
|---|---|---|
| 10,000-character string | 28.5 ms | 0.32 ms |
| 40,000 | 373 ms | 0.82 ms |
| 160,000 | 7,296 ms | 1.9 ms |
| two 129 KB + 119 KB manifests | 39.6 s, 3.0 GB max RSS | 3 ms, 6.4 MB |
| > 100 manifests | did not complete: terminated by a signal the harness did not send, after reaching 8.5 GB max RSS | 50 ms, 33 MB |

Both columns come from the same program, built with the same compiler against the same revision, switching only `vlib/toml` between master and this branch, on an Apple M-series machine.

The change is a `strings.Builder` in the three functions, `write_u8` for single bytes, and a `has_newline` flag replacing `lit.contains('\n')`. The flag is equivalent: `contains` is monotone, and `lit` only ever grows by the current byte or by the string returned from `handle_escapes`, so both are folded into the flag. The order of the branches is untouched — in particular the return on a closing quote still happens before the newline check, and the escape handling still `continue`s after a skip. `handle_escapes` keeps its concatenations; they are bounded to six bytes.

Tests in `vlib/toml/tests/strings_test.v`:

- a single-line basic string may not hold a raw newline, and an unfinished one must be reported;
- a newline that reaches the literal through an escape rather than as a plain byte is caught as well, checked through the scanner's own message — removing only the `has_newline` update fed by escapes makes this case fail;
- long plain and escaped values keep their exact content.

Verification beyond the TOML suite (`vlib/toml`, 33/33): a differential harness compares master and this branch over 28 edge cases — escapes, valid and invalid unicode, literal and multiline strings, CRLF, unfinished strings, control characters, an emoji — and the output, error messages included, is identical byte for byte. A second differential over 78 real TOML files (those the old code still parses in reasonable time) compares the SHA-256 of each parsed tree: identical. An independent differential check of 1,020 synthetic inputs found byte-identical tokens, scanner positions and errors, decoded values, and parse results (6,120 records). A targeted mutation check confirmed that the new regression test detects missing newline tracking in escape fragments.


medvednikov (commented):
Reviewed the code changes only; CI intentionally ignored as requested.

No blocking issues found. The `strings.Builder` conversions look behavior-preserving: `Builder.str()` returns an owned copy and clears the builder, so the deferred `free()` is safe, and the byte-wise `write_u8`/`write_string` replacements avoid the quadratic concatenation paths. The `has_newline` flag also preserves the prior monotonic `lit.contains('\n')` behavior while avoiding repeated rescans; escaped-newline handling is covered by the added scanner-level regression test. The long-string tests exercise both plain and escaped decoding paths.

One tiny non-blocking nit: `extract_string()` allocates its 64-byte builder before the multiline check, so multiline strings allocate/free that builder before delegating to `extract_multiline_string()`. Moving builder initialization below the `is_multiline` early return would avoid that constant extra allocation, but it does not affect correctness.

davlgd:
Rebased and fix added

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I did not find a concrete correctness issue. The builder conversion preserves append order and escape handling, `has_newline` is monotonic like the old repeated `contains('\n')` check (including escape fragments), and each returned builder string is materialized once before deferred cleanup. The added long-string and newline regressions cover the behavior-sensitive paths.

PR #28554, opened 2026-09-13
json2: fix leaked indentation after an empty array or map

aki-xavier:
## What

`json2.encode(v, prettify: true)` leaked one indentation level for every empty array or map, so every member after an empty collection was indented too deep, and one level deeper again for each further empty collection.

```v
import json2

fn main() {
	value := json2.decode[json2.Any]('{"a":[],"b":[],"c":[],"tail":1}') or { panic('bad json') }
	println(json2.encode(value, prettify: true))
}
```

Before:

```
{
    "a": [
        ],
        "b": [
            ],
            "c": [
                ],
                "tail": 1
}
```

After:

```
{
    "a": [],
    "b": [],
    "c": [],
    "tail": 1
}
```

`"tail"` is a member of the outer object, so it belongs at one level of indentation; it was at four. Because the leak accumulates, the indentation a member receives depends on how many empty collections precede it, so the rendered size of a document grows with the position of a member rather than with its nesting: for a document carrying many empty collections, the rendered text is quadratic in the number of members. That shape is common in practice, since descriptions, payloads and generated configs tend to carry many collections that are empty.

## Why

The four places that open a collection raise the indentation level right after writing the opening bracket, but the matching lowering lives in the loop body:

```v
	encoder.output << `[`
	if encoder.prettify {
		encoder.increment_level()
		encoder.add_indent()
	}

	for i, item in val {
		...
		} else {
			if encoder.prettify {
				encoder.decrement_level()
				encoder.add_indent()
			}
		}
	}
```

An empty collection never enters the loop, so the level it raised is never lowered, and it stays raised for the rest of the document.

## Fix

An empty collection is written as `[]` or `{}` and raises no level, so there is nothing to lower:

```v
	if encoder.prettify && val.len > 0 {
		encoder.increment_level()
		encoder.add_indent()
	}
```

Output for non-empty collections is byte for byte what it was, so no existing expectation changes.

The same pattern is corrected in all four places that open a collection: `encode_array`, `encode_map`, `encode_array_of_sumtype_variants`, and the inline map branch of `encode_value`. (`encode_map` currently has no callers; it is corrected as well, so that reviving it cannot reintroduce the bug.)

## Test

`test_encoder_prettify_empty_collections` in `vlib/json2/tests/encoder_test.v` encodes an empty array and an empty map followed by another member. The assertion is written against the output the encoder should produce, so it fails against the code before this change (the empty collection is closed on its own over-indented line and the following members are shifted) and passes after it. `v fmt -verify` is clean on both files.


medvednikov:
@codex review

PR #28558, opened 2026-09-13
net.http: add Request.remote_addr (Go's RemoteAddr equivalent)

medvednikov:
## Problem

`net.http` has no equivalent of Go's `http.Request.RemoteAddr`. The only way to
find out who sent a request was a `Remote-Addr` header the HTTP/1.1 server
injected into the parsed request, and that has three problems:

**1. It is forgeable.** `Header.add_custom` appends and `Header.get_custom`
returns the *first* match, so the client's own header wins over the server's:

```v
// server: req.header.get_custom('Remote-Addr')
$ curl -H 'Remote-Addr: 6.6.6.6' http://localhost:8080/
first=6.6.6.6  all=['6.6.6.6', '127.0.0.1']
```

**2. It is missing on HTTP/2** — neither h2c nor h2-over-TLS set it.

**3. On TLS it comes from mbedtls' `SSLConn.ip`**, which is only filled in for
IPv4 peers, so an IPv6 client has no address at all.

## Change

Adds a first-class `Request.remote_addr`: an `ip:port` string (`[::1]:8080` for
IPv6, same shape as Go's `RemoteAddr`) read straight off the accepted socket, so
a client cannot forge it. `Request.remote_ip()` returns just the address.

```v
fn (mut h MyHandler) handle(req http.Request) http.Response {
	return http.Response{
		body: 'you are ${req.remote_ip()} (${req.remote_addr})'
	}
}
```

It is populated on all four server paths — plain HTTP/1.1, TLS HTTP/1.1, h2c and
h2-over-TLS — and is empty for a request you build yourself to send with the
client.

The `Remote-Addr` header is still set, in the same bare-IP format, so existing
code keeps working. Any copy the client sent is now removed first, in any
casing, via a new module-private `Header.remove_custom_all`. `delete_custom`
could not be reused for that: it matches a single exact casing and overwrites
matches in place with an empty value — a tombstone that `get_custom` still
returns, as an empty string, ahead of anything added afterwards.

Headers live in a fixed 50-entry array, so the mirrored header is best-effort on
a request that already filled it (previously that case would index past the end
of the array); `req.remote_addr` is always set.

`veb` has its own server and is unaffected — `ctx.ip()` remains the way to get a
client address there, and the field's doc comment says so.

## Tests

New `vlib/net/http/server_remote_addr_test.v`, plus cases added to
`server_h2c_test.v` and `server_tls_test.v`:

- the address a handler sees is the real peer `ip:port`, and `remote_ip()` drops
  the port
- a client-sent `Remote-Addr` / `remote-addr` / `REMOTE-ADDR` cannot shadow it
- a request that fills the header table still gets `remote_addr`
- HTTP/2: the field is set, and a `remote-addr` field in the client's HPACK
  block cannot shadow it
- TLS: the field is set with a port (which `SSLConn.ip` never carried)
- unit coverage for `remote_ip()` on IPv6 and for `remove_custom_all`

`vlib/net/http`'s suite passes locally on macOS (server, TLS server, h2 server,
h2c, header, request, response, http, transport_h2), as does
`vlib/veb/tests/context_ip_test.v`.


medvednikov:
@codex review

PR #28562, opened 2026-09-13
v3: let a project's `modules/` shadow vlib from a submodule import too

medvednikov:
## The bug

`resolve_local_or_project_module_path` probed the importing file's own `modules/` directory, and the project root itself, but never the project root's `modules/`. So an import written in the entry file resolved to a project module, while the very same import written one directory down did not:

```text
myapp/
  main.v                  -- `import net` -> myapp/modules/net  (ok)
  modules/net/net.v
  modules/wrapper/w.v     -- `import net` -> vlib/net           (wrong)
```

Binding to vlib silently changed what got built rather than failing outright. Compiling the Vinix kernel this way pulled vlib's `net` in beside the kernel's own module of that name, and with it vlib's `os`, `rand`, `io`, `net.conv`, `encoding.binary`, `crypto.sha256`, `math.bits` and `strings.textscanner` — 58 extra files, and 18 host POSIX headers (`<sys/un.h>`, `<netdb.h>`, `<dirent.h>`, …) emitted into a translation unit that is compiled `-nostdinc -ffreestanding` against the kernel's own headers.

This is also already the documented behaviour: doc/docs.md says the `v.mod` folder is prepended to the lookup path so that "files beside or **below** it can import sibling modules under the same tree".

## The fix

A `modules/` directory declares modules for the whole project, so it is now probed at every level between the importing file and the project root.

- The walk stops at the project root, and otherwise at the filesystem root, so it cannot climb further than V1's own parent-directory search did.
- The probe runs *after* the existing project-root probes, so no import that already resolved changes where it resolves to.
- The project root is the nearest `v.mod`, which can sit well above the directory actually being compiled, which is why this is a per-level walk rather than one extra `project_root/modules` probe.

## Tests

`vlib/v/tests/local_module_shadows_vlib_from_submodule_test.v` covers both halves:

1. the project's module wins for an import written inside a submodule, and
2. the shadowed vlib module is no longer compiled in — asserted on the generated C, so a regression cannot hide behind a project that happens to still run.

Both cases fail against a compiler without the fix.

## Checks run

- `./v self`
- every `vlib/v/tests/*module*`, `*import*`, `*submodule*`, `*vmodules*` test file
- `./v fmt -verify` on both touched files

Two failures in that set (`local_submodule_in_project_root_codegen_test.v`, `import_order_{1,2}_test.v`) reproduce identically with the fix stashed, so they are pre-existing.

## Not fixed here

With this landed, V3 still cannot build the Vinix kernel: the C preamble is chosen from whether the program has any `#include` of its own, so a freestanding target still gets the hosted libc set (`<pthread.h>`, `<sys/epoll.h>`, `<windows.h>`, …). That is a separate change and is not in this PR.


JalonSolov:
So we are keeping `modules` subdir with a special meaning?  I thought it was going the way of `src` subdir...

medvednikov:
@JalonSolov you're right, let's kill it off.

medvednikov:
Closing: extending the `modules/` special case is the wrong direction. `modules/` is a virtual directory the compiler has to know about, the same way `src/` was before V3 removed it, and it should go the same way rather than gain more reach.

The replacement needs no new mechanism: a module's import path is already its path under the nearest `v.mod`, so a module beside the entry file resolves from a submodule import today. I verified that on the Vinix kernel — moving `kernel/modules/*` up beside `main.v` makes V3 resolve the kernel's own `net` module with no compiler change at all, which is exactly the resolution bug this PR was working around.

Replacing it with a PR that removes the `modules/` lookup and reports the migration, mirroring `report_v3_removed_src_layout`.

PR #28563, opened 2026-09-13
docs: #pkgconfig runs the pkg-config command, and drops its flags when it fails

davlgd:
The `#pkgconfig` section of `doc/docs.md` says that "V uses its own pkgconfig library that is compatible with the standard freedesktop one", and that it does so because "spawning processes is not desirable for security and portability reasons". That is no longer what happens.

`c_pkgconfig_flags` (`vlib/v/gen/c/cleanc.v:10081`, called from `4628` and `4671`) runs the external `pkg-config` command and returns an empty flag list on any non-zero exit:

```v ignore
result := cmdexec.run('pkg-config', args)
if result.exit_code != 0 {
	return []string{}
}
```

The claim dates from 2020 (`fbd193b27c`), when it was accurate: `vlib/v/pkgconfig/pkgconfig.v` and `main.v` implemented a freedesktop-compatible `.pc` parser. `edf824295b` made V3 the only C compiler on macOS and Linux, and `2a7447b5e5` removed V1 and deleted those files. The section has been edited several times since 2020, but this sentence survived all of them.

On a machine that has the `.pc` files but no `pkg-config` binary, a program whose flags come only from `#pkgconfig` therefore fails to build — on a missing header, or on an undefined symbol where the header sits on the default search path. The standard library is unaffected: its bindings guard the directive with `$if $pkgconfig(...)` and a fallback branch. A program that writes the directive bare has no such fallback.

This patch describes what resolving the directive actually does, recommends the guard vlib already uses, explains why the guard works (`$pkgconfig()` runs the same command, so `$else` is reached when resolution fails), shows it with the real example from `vlib/db/sqlite/sqlite.c.v`, and notes that a directive naming several packages has to probe all of them. It also corrects "a hardcoded list of default pkg-config paths", which described the removed V implementation; `PKG_CONFIG_PATH` still works, as pkg-config's own variable.

This documents the behaviour as it stands. If the intent is to restore the `.pc` parser, or to report a failed resolution instead of ignoring it, just close this.


medvednikov (commented):
Reviewed the documentation against the current implementation (CI status not considered). I did not find a factual issue: the default C generator shells out to `pkg-config`, returns an empty flag list on non-zero status, and the documented `$pkgconfig(...)` guard/fallback pattern matches the standard-library usage. The wording also correctly scopes the silent failure behavior to the default C generator rather than FastC.

PR #28565, opened 2026-09-13
fix(v3): keep branch type of if-expr values with generic calls

aki-xavier:
Supersedes #28469 (which was closed; a force-push after closing prevents reopening it). This is now a single, focused commit rebased on current master, dropping two unrelated commits that were bundled before (a closure path fix and a cgen mut-pointer fix). (the closure-path change is not needed, since the current master v3 backend already emits the correct access; the cgen mut-pointer issue is separate.)

## Summary

An `if`-expression whose branches call a generic function is lowered to a temporary of the wrong type, truncating the value:

```v
module main
import math

fn main() {
	a := 1e-3 / 0.02
	x := if a > 0 { math.min(1.0, a) } else { 1.0 }
	println(x) // 0, expected 0.05
}
```

Generated C (before):

```c
int __if_val_0 = 0;
if (true) {
	__if_val_0 = math__min_T_f64(1.0, a); // 0.05 truncated to int -> 0
} else {
	__if_val_0 = 1.0;
}
int x = __if_val_0;
```

## Root cause

Two things had to line up to produce the truncation:

1. `try_expand_if_expr_value_for_type` keeps the caller-supplied `result_type` for the if-value temporary. In declaration contexts that is the default `int`, so the temporary is `int` even though the merged branch type is `f64`.
2. The merged branch type was itself wrong. `if_expr_branch_result_type` resolves each branch through `stmt_value_type_with_smartcasts` → `node_type_with_smartcasts`, whose `.call` arm returned the flat node type directly. A call has no type there until the checker has seen it, so a call to a generic function fell back to the default `int` and won the branch-type merge.

Non-generic calls were unaffected because `math.sqrt` resolves through the checker.

## Fix

- Use the merged branch type for the if-value temporary when the target is the default `int` (`try_expand_if_expr_value_for_type`).
- Resolve a call branch's return type in `node_type_with_smartcasts` instead of using the raw node type, and add `generic_call_declared_return_type`, which reconstructs the concrete return type of a call to a generic function from its declared signature by inferring the generic arguments from the argument types (falling back to the previous behaviour when the arguments leave a generic parameter undetermined).

## Tests

`vlib/v/compiler_tests/if_expr_generic_call_type_test.v` covers generic `min`/`max` calls in both then/else branches and an explicit `min[f64]` instantiation. The test lives under `compiler_tests/` (where the v3 test suite now resides after the directory restructure) so it is exercised by the v3 backend.

## Verification

Built the compiler from this branch and ran with the v3 backend:

- The repro above prints `0.05`.
- `vlib/v/compiler_tests/if_expr_generic_call_type_test.v` passes (it fails on the unpatched compiler).
- No new failures against a stock compiler on a comparison set of if-expr and generic tests; the two pre-existing failures under a forced v3/no-fallback build (`array_map_alias_if_expr_test.v`, `match_as_if_expr_value_propagation_codegen_test.v`) occur with and without this change.


medvednikov (commented):
Reviewed the code changes only (CI status not considered). I did not find a concrete correctness issue. The new call-return fallback is limited to generic declarations whose type parameters can be fully inferred, and the if-expression override only replaces the stale default `int` temporary when the merged branch type is a concrete non-int type; unresolved cases retain the existing fallback behavior.

medvednikov:
Thanks!

PR #28567, opened 2026-09-13
fix(cgen): pass a mut parameter read by value to C functions

aki-xavier:
## Summary

A `mut` parameter read by value is one dereference deep: its value is `*o`, not the address of the parameter slot. When such a parameter is passed to a C function, the argument was emitted as the slot itself, so the C side received the address of the parameter slot and read an unrelated pointer.

```v
#include "/tmp/c_id.c"

fn C.c_mut_voidptr_id(p voidptr) voidptr

fn c_mut_voidptr_id_thunk(mut o voidptr) voidptr {
	return C.c_mut_voidptr_id(o)
}

fn main() {
	mut o := voidptr(u64(0x1234))
	got := c_mut_voidptr_id_thunk(mut o)
	println(got == o) // false
}
```

Generated C (before):

```c
void* c_mut_voidptr_id_thunk(void** o) {
	return c_mut_voidptr_id(o);   // should be c_mut_voidptr_id(*o)
}
```

## Root cause

`gen_call_args` emits the arguments for a call that the inline `gen_call` loop does not handle, which includes `C.` functions. The plain expression path dereferences a by-value `mut` parameter through the expected type (`gen_current_mut_param_value_read`), but that helper deliberately declines when the expected type is a pointer — and a `C.` function's parameter types are either pointer-typed (`voidptr`) or not registered at all, so the argument fell through and `gen_expr` emitted the slot.

## Fix

A mutable parameter is one dereference deep only when its own value is a pointer. Every other mutable parameter already holds the address of the caller's storage in its C variable, so a callee that asks for a pointer wants the slot itself.

`gen_call_args` now emits the dereference for a by-value argument naming a mutable parameter whose own value type is pointer-like, and skips the older expected-type based handling for that case. Explicit `mut` arguments are unaffected.

Keying the check off the callee's parameter type alone is not sufficient: it also intercepts implicit references such as

```v
type Builder = []u8

fn (mut b Builder) push_byte(c u8) {
	b << c
}

fn build(mut sb Builder) int {
	sb.push_byte(u8(65))
	return sb.len
}
```

where `mut sb Builder` lowers to `Array* sb`, so the slot must be forwarded unchanged. Emitting `Builder__push_byte(*sb, ...)` passes a value where a pointer is required, which both tcc and clang reject. That form also occurs in `vlib/builtin/string_interpolation.v` (`process_str_intp_data` takes `mut sb strings.Builder`), so it broke the build of the `builtin` module as well.

## Tests

`vlib/v/compiler_tests/c_mut_voidptr_param/` covers passing a `mut voidptr` parameter to a C function by value, plus forwarding an ordinary mutable value parameter by implicit reference to a `mut` callee. Both fail without the fix and pass with it.

## Verification

Built the compiler from this branch and compared against a compiler built from current master:

- The repro above prints `true`; the generated call is `c_mut_voidptr_id(*o)`.
- The receiver case emits `Builder__push_byte(sb, (u8)(65))` instead of `Builder__push_byte(*sb, (u8)(65))`.
- `./v -silent test vlib/v/gen/c/`: 23 passed, 23 total.
- No new failures across the `mut` / pointer / C-interop codegen tests compared (`mut_param_array_clone_codegen_test`, `mut_pointer_param_field_free_codegen_test`, `pointer_receiver_rvalue_test`, `c_pointer_null_arg_codegen_test`, `loop_smartcast_mut_call_write_test`). `mut_param_return_reassign_codegen_test` and `interface_pointer_forward_codegen_test` are left to CI: both self-host the compiler with `-d ownership`, which takes upwards of 15 minutes on a single core locally.

Note: `mut_param_reassign_codegen_test` fails on current master independently of this change. Its `run_bad` cases assert on checker diagnostic text that master no longer emits — at this branch's original base the message was ``redefinition of `${lhs.value}`; redefinition of ${lhs.value}``, and the plain-text tail the test matches has since been dropped. Those errors abort at check time, before cgen.


medvednikov:
@codex review

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I confirm the existing blocking argument-lowering bug is still present: the new fast path keys off the callee parameter being pointer-like, so `fn wrapper(mut value S) { take_ref(value) }` with `take_ref(&S)` is intercepted and emitted as a value read (`*value`) instead of forwarding the pointer-backed mut-parameter storage. The intended `mut voidptr` case needs to be distinguished from an ordinary mutable value passed implicitly by reference, ideally using the argument's semantic/source type rather than only the destination parameter type.

aki-xavier:
You're right — the guard was too broad, and the argument's type is the correct discriminator. Fixed and force-pushed.

The path that actually regresses is a method receiver, which is also why it reaches `vlib/builtin`:

```v
type Builder = []u8

fn (mut b Builder) push_byte(c u8) {
	b << c
}

fn build(mut sb Builder) int {
	sb.push_byte(u8(65))
	return sb.len
}
```

`mut sb Builder` lowers to `Array* sb`, so the slot must be forwarded as-is. Against `void Builder__push_byte(Array* b, u8 c)`:

- before: `Builder__push_byte(*sb, (u8)(65));`
- after: `Builder__push_byte(sb, (u8)(65));`

`vlib/builtin/string_interpolation.v` hits this (`process_str_intp_data` takes `mut sb strings.Builder`), so the `builtin` module failed to build.

The gate now requires the mutable parameter's own value type to be pointer-like, so only `mut voidptr`-style parameters are one dereference deep and `c_mut_voidptr_id(*o)` still holds.

One note on the snippet in your comment: `fn wrapper(mut value S) { take_ref(value) }` with `take_ref(&S)` doesn't reproduce here — that call doesn't route through `gen_call_args`, and its generated C is identical to master. The method-receiver form above is the one that breaks, so it's the shape worth targeting.

The `assert p == unsafe { &a }` identity check was tautological (`mut p` is assigned once and never reassigned), so it couldn't catch this; it's replaced with a receiver-forwarding case.

Separately, `mut_param_reassign_codegen_test` fails on current master independently of this change: its `run_bad` cases assert on checker text master no longer emits (``redefinition of xs`` and ``expected `&&Item` `` versus today's wordings). Those abort at check time, before cgen, so it isn't a regression from this PR.

PR #28569, opened 2026-09-13
cgen: materialize a receiver that is a field of a call result

aki-xavier:
A method that takes its receiver by reference needs the address of that receiver. The generated C takes the address of the receiver expression, but only a receiver that was itself a call was materialized first, so a field of a call result kept its address taken directly. A field of a call result is not an lvalue, so gcc and clang reject the generated C, while the bundled tcc accepts it and hides the problem.

Minimal reproduction:

```v
struct Inner {
	values []string
}

fn (i &Inner) count() int {
	return i.values.len
}

struct Outer {
	inner Inner
}

fn make() Outer {
	return Outer{
		inner: Inner{
			values: ['a', 'b']
		}
	}
}

fn main() {
	println(make().inner.count())
}
```

The receiver is now materialized whenever it has no address in C, which also covers an element of a call result. A receiver that does have an address is emitted as before.

The regression test inspects the emitted C rather than building it, because the default C compiler in the test environment accepts the invalid expression.


medvednikov (commented):
Reviewed the code changes only (CI status not considered). The addressability check itself looks appropriate for distinguishing lvalues from call-result projections. I found one portability issue in the new regression harness.

vlib/v/compiler_tests/non_addressable_receiver_codegen_test.v
```diff
@@ -0,0 +1,85 @@
...
+const nonaddr_receiver_tests_dir = os.dir(@FILE)
+const nonaddr_receiver_v3_dir = os.dir(nonaddr_receiver_tests_dir)
+const nonaddr_receiver_vlib_dir = os.dir(nonaddr_receiver_v3_dir)
+const nonaddr_receiver_v3_src = os.join_path(nonaddr_receiver_v3_dir, 'v.v')
+
+fn nonaddr_receiver_build_v3() string {
+	v3_bin := os.join_path(os.temp_dir(), 'v3_non_addressable_receiver_test')
+	os.rm(v3_bin) or {}
+	build :=
+		os.execute('${nonaddr_receiver_vexe} -gc none -path "${nonaddr_receiver_vlib_dir}|@vlib|@vmodules" -o ${v3_bin} ${nonaddr_receiver_v3_src}')
```
medvednikov, line 13: This compiler-test helper has the same cross-platform path issue as other temp-compiler harnesses: it creates an unsuffixed output name and interpolates executable/output/source paths without `os.quoted_path`. On Windows the produced compiler should use `.exe`, and paths containing spaces can break this command (as well as the later generate/compile/run commands). Please add the platform executable suffix and quote all filesystem paths passed to `os.execute`.

aki-xavier:
Closing as superseded by #28568 (a666e410ed), which landed the equivalent fix at the same site in `vlib/v/gen/c/fn.v` while this PR was open:

```v
materialize_receiver := receiver_wants_ptr && !receiver_is_ptr
	&& !g.expr_is_addressable(receiver_id)
```

That subsumes the extra condition here: `expr_is_addressable()` already returns false for a `.call`, and for a `.selector` rooted at one, so `receiver.kind == .call` never changes the result.

Verified against master 4b8eb75 — the nested repro now emits `Inner__count(&((main__Inner[]){(make()).middle.inner})[0])` instead of `&(make()).middle.inner`, and prints `2` / `3`.

Thanks for the review note on the temp-compiler harness; the executable suffix and `os.quoted_path` points are both correct and I've applied them locally. For the record, master has no regression test for a receiver that is a field of a call result yet — I confirmed one that fails when the condition is reverted to `receiver.kind == .call`, so that coverage is still worth adding separately.

PR #28571, opened 2026-09-13
cgen: preserve defaults on specialized generic structs

nepinhum:
## What this fixes

Specialized generic structs don't have their own source declaration. Types like
`Box[int]` or their flattened CGen form only exist in the type tables.

Several v3 CGen default value paths still tried to resolve them with
`find_struct_decl()`. When that lookup failed, CGen treated the specialization as
having no declared defaults and explicit defaults from the generic source could be
silently lost.

Example:

```v
// box/box.v
pub struct Box[T] {
pub:
    x    int = 5
    size int = sizeof(T)
}

// main.v
b := box.Box[string]{}
```

Before this change, recovered/defaulted instances could lose values such as `x = 5`
or evaluate type-bearing expressions using the unresolved `T`.

## What changed

This adds one shared recovery path for generic struct defaults:

- `struct_default_decl_source()` resolves a specialized/flattened generic back to its
  source declaration and keeps the concrete generic arguments
- recovered defaults install that generic context while the original default expression
  is emitted
- `generic_default_type_text()` substitutes type-bearing expression text before semantic
  lookup

The same recovery is now used by direct struct defaults, nested/default-init checks,
globals, promoted embedded fields and promoted root initializers.

Type-bearing defaults are also specialized consistently.

## Tests

- `v test vlib/v/tests/generics/generic_struct_default_expr_cross_module_test.v` - `chan T` as a generic struct field
  keeps the placeholder (reproduces on master, no default involved)
- `v test vlib/v/gen/c/types_test.v` - doesn't compile either, unrelated: '__arr_lit_9' undeclared in test_json_helper_scan_requires_legacy_json_module
- `v test vlib/v/tests/generics/` - 281/308, same 27 failures as master


medvednikov (commented):
Reviewed the code changes only (CI status not considered). I did not find a concrete correctness issue in the generic-default recovery path. The caller-side generic arguments are canonicalized before switching into the source declaration context, the temporary default/generic context is restored at each call site, and the added coverage exercises flattened names, caller-module type collisions, type-bearing defaults, promoted fields, globals, arrays/channels, `is`, and `as`.

PR #28572, opened 2026-09-13
cgen: recognize a file's own module in a qualified name

aki-xavier:
A file can qualify a symbol with the name of the module it belongs to, as in `m.greeting` written inside `m`. No import statement records that module, so the C generator did not recognize the qualifier as a module name. It then wrote the qualified name into the C output verbatim, and the C compiler read the module name as an undeclared identifier.

Minimal reproduction, as two files of one module. `lib.v`:

```v
module m

pub const greeting = 'hi'
```

`lib_test.v`:

```v
module m

fn test_qualified_const() {
	println(m.greeting.len)
}
```

Running the tests of that module builds C in which the qualified constant is not resolved to the C name of the constant, and the build fails in the C compiler. The qualification is accepted when it is read from a test file of the module, which is why that is the shape of the reproduction. Qualifying with the module name is accepted in the other shapes too: when the module is imported by the program, and when the module is built as a shared library.

The generator now treats the module of the file it is generating as a module base, so the name is emitted as the qualified C name of the constant. Names that are shadowed by a value are unaffected, because a value base is still checked first at every use site.

The regression test builds such a module, runs its tests, and checks the generated C for the qualified C name. It disables the fallback compiler while building the fixture, because a failed build is otherwise retried with the fallback compiler, which hides the defect.


JalonSolov:
V has always referred to things in the same module without the module name.  Allowing the module name to be used goes against the "one way" philosophy of V.

medvednikov:
I agree with you @JalonSolov, V shouldn't fail at the C compilation though.

vlib/v/compiler_tests/self_qualified_module_name_codegen_test.v
```diff
@@ -0,0 +1,65 @@
...
+const selfqual_tests_dir = os.dir(@FILE)
+const selfqual_v3_dir = os.dir(selfqual_tests_dir)
+const selfqual_vlib_dir = os.dir(selfqual_v3_dir)
+const selfqual_v3_src = os.join_path(selfqual_v3_dir, 'v.v')
+
+fn selfqual_build_v3() string {
+	v3_bin := os.join_path(os.temp_dir(), 'v3_self_qualified_module_name_test')
+	os.rm(v3_bin) or {}
+	build :=
+		os.execute('${selfqual_vexe} -gc none -path "${selfqual_vlib_dir}|@vlib|@vmodules" -o ${v3_bin} ${selfqual_v3_src}')
```
medvednikov, line 13: This helper builds and later executes a temp compiler using an unsuffixed, unquoted path. On Windows compiler outputs normally need the platform executable suffix, and on any platform `@VEXE`, the temp path, or the source path can contain spaces. The later `${v3_bin}` invocations have the same quoting issue. Please derive the Windows `.exe` suffix and use `os.quoted_path(...)` consistently for the executable/output/source paths so this compiler test is portable.

medvednikov (commented):
Reviewed the code changes only (CI status not considered). The production change looks internally consistent, but the new compiler regression test has a cross-platform execution issue.

PR #28575, opened 2026-09-13
v3: remove the virtual `modules/` directory

medvednikov:
Follow-up to #28562, which I closed: extending the `modules/` special case was the wrong direction. `modules/` is a directory name the compiler has to know about, the same way `src/` was before V3 stopped treating it as a virtual source root, and it goes the same way here.

## What replaces it

Nothing new. `v.mod` already provides the module lookup root, and a module's import path is its path under that root — so the directory holding a module sits at the root rather than one level below it.

Every `import` is unchanged across the move: `modules/gpu/agx/fw` and `gpu/agx/fw` are both `import gpu.agx.fw`. Migrating a project is a directory move with no source edit.

```text
myapp/
├── v.mod
├── main.v
└── mymod/            # was myapp/modules/mymod
```

## Why it is worth removing rather than keeping

Keeping it was not neutral — it resolved imports to the wrong module, silently. `import net` written in a submodule of a project that has its own `modules/net` reached *past* it and bound to vlib's `net`, because only the importing file's own directory was probed. In the Vinix kernel that pulled vlib's `net`, `os`, `rand` and `io` into the build — 58 extra files and 18 host POSIX headers (`<sys/un.h>`, `<netdb.h>`, …) in a kernel that compiles `-nostdinc -ffreestanding`. #28562 was a workaround for that symptom; removing the convention removes the cause. I verified on Vinix that moving its modules up makes V3 resolve the kernel's own `net` with no compiler change at all.

## The change

Six probes go: the importing file's own `modules/` and the project root's, in both the V3 resolver and `pref.get_module_path`'s parent walk, the two in `import_uses_explicit_module_alias`, and the one the ancestor walk tried at every level it climbed — that last one kept the whole layout working from any file in a project, only demoted below the global module directories.

An import that a `modules/` directory would have satisfied now reports the move instead of failing with a bare "not found":

```
builder error: cannot import module "net" (not found)
the virtual `modules/` directory is no longer searched for modules.
Move it up beside the v.mod it belongs to, which keeps the import path the same:
	mv /path/myapp/modules/net /path/myapp/net
```

The tree migrates with it — cmd/tools' five tool modules, nine examples, and the test fixture modules under `vlib/v/tests` and `vlib/v/slow_tests` (124 pure renames). `vlib/v/checker/tests/modules` is deliberately untouched: it is a directory of test *cases*, not a lookup root, and its fixtures already sit beside their `main.v`.

## Checks run

- `./v self`
- the five `cmd/tools` programs that import the moved modules (`vbug`, `vbuild-tools`, `oldv`, `performance_compare`, `vtest-self`) each build
- every module / import / alias / submodule test under `vlib/v/tests`, plus the specific tests importing each moved fixture
- `./v fmt -verify` on the four edited files

Three failures in that set (`local_submodule_in_project_root_codegen_test.v`, `import_order_{1,2}_test.v`, `generics_fn_variable_3_test.v`) reproduce identically on master and are pre-existing.

## Note for downstream projects

This is a breaking layout change for any project using `modules/`. The diagnostic gives the exact `mv`, and no import statement needs editing.

`v install --local` moves with it. It installed into `<nearest-v.mod>/modules`, the one directory this branch stops searching, so a package could install and then fail to import; it now installs into the project's own lookup root, beside the modules the project writes itself. `v new`'s `.gitignore` no longer names `modules/`.

## Review follow-ups

- The migration hint is produced by the resolver rather than the checker, so it is only printed for a `modules/` directory that holds a module *this build* could have used — a directory whose only source is disabled for the target no longer advertises a move that would not help.
- The printed command runs as-is at any depth: a dotted `foo.bar` moves the top-level `modules/foo` tree, or creates the destination parents when `foo` is already taken.
- vdoc no longer resolves `<compile-dir>/modules/<name>` on its own, which documented a layout the compiler rejects.
- The printed command is quoted, so a project path holding a space is runnable as printed, and it names `move`/`mkdir` on Windows, where `mv` and `mkdir -p` do not exist. A destination that already exists asks for a merge instead: `mv` onto a directory moves the module *inside* it, leaving the import unresolved.
- The move hint applies the same ownership boundary as resolution, so an importer under `<workspace>/app` is never told to move `<workspace>/modules/foo`, which belongs to whatever lives above it.
- `resolve_ancestor_module_path` no longer tries `<level>/modules/<mod path>` either. A level holds a module at its path under that level and nowhere else; a project's own copy still beats a neighbour of the same name, since it sits in the project's lookup root, which the walk reaches first.
- `v install --local` installs into the project's own lookup root, the folder holding its `v.mod`. It used to install into `<root>/modules`, the one directory this branch stops searching, so a package could install and then fail to import.
- Since that root is shared with the modules the project writes by hand, VPM records what it installs there and only acts on a directory that still answers with that record. Version control does not prove provenance: a project may well keep a module of its own as a submodule or a manual clone. The record is a token inside the checkout VPM made, in its VCS directory so it stays out of `git status`, plus the note of that token in the global cache, never in the project — so an install deleted by hand cannot pass its authority to whatever the project writes at that path next. The note is named after the token rather than after a path, so renaming or moving a project takes its installs with it; the path the note holds is what separates a checkout that moved from a copy of it, which inherits nothing while the original still answers with the same token. An install VPM cannot record is undone rather than left behind, and a failed removal keeps its record, so it can be retried.
- Discovery goes by the same record under a local root, so `v update --local` no longer runs `git pull` in modules the project vendored itself, and they stay out of `v list --local` and `v outdated --local`. It reads the records rather than walking the project, which would mean descending through `.git`, `node_modules`, build output and every source directory to find the few directories the records already name: on a project of 2000 directories `v list --local` drops from ~0.13s to ~0.02s, and no longer grows with the project.
- A package an older `v install --local` left in the project's `modules/` directory has no record, and once moved up beside the `v.mod` it is indistinguishable from a checkout the project vendored itself. `v install --local --adopt <module>` records an existing checkout as VPM's, which is the consent nothing on disk can supply; the refusals from `remove`, `update` and `install` name it.
- vdoc's "Available modules" list no longer strips a leading `modules/`: that directory is an ordinary namespace now, so a module under it is `modules.<name>`, which is what its `-m` advice has to print.
- `v install --local vendor/dep` still means the repository at that path: the rule that a directory inside the module store is an installed module shadowing a registered name now applies to what VPM installed there, since under `--local` that store is the project itself.
- `v link --local` and `v unlink --local` are refused. Both are about the V modules directory, and pointing that at the project itself meant linking a project named `foo` into `<project>/foo -> <project>`, or unlinking whatever symlinked module the project keeps at that path.
- `v new`'s `.gitignore` no longer names `modules/`, which is an ordinary module directory now.
- The compiler-test fixtures that needed `modules/` to resolve (`missing_import_diagnostics_test.v`, `review_transform_regressions_test.v`, `sibling_project_module_resolution_test.v`) move up with everything else, as do the archived builder fixtures and the tracked test path list.


medvednikov:
@codex review

PR #28578, opened 2026-09-13
v3: add `-target-libc-headers`, for a target that supplies libc itself

medvednikov:
Groundwork for compiling a kernel with V3. Opt-in flag, so no other target's output moves.

## The problem

The C preamble is chosen from whether the program has any `#include` of its own — which reads a project's *own* headers as proof that the *host* libc is present. A kernel compiling `-nostdinc -ffreestanding` against its own header tree therefore gets the full hosted set (`<sys/un.h>`, `<pthread.h>`, `<sys/epoll.h>`, `<windows.h>`), none of which exist for it, and the generated C fails on its fifth line:

```
blob.c:5:10: fatal error: 'sys/un.h' file not found
```

It cannot be inferred from `-os` either: the same OS also hosts ordinary programs that link the host's libc, which is why gating on `-os vinix` was rejected in earlier review.

## Not `-freestanding`

`-freestanding` already exists and means *no libc at all*, with a `-bare-builtin-dir` supplying `bare_print`/`bare_panic`. Here there **is** a libc — it is the target's own. I originally reused the name and also defined a `freestanding` compile-time flag; that switched vlib into its bare-metal paths, so `bare_print`, `bare_eprint` and `bare_panic` went undeclared. That is the opposite of what such a target wants, hence the separate name and no compile-time define.

## What the flag does

1. The hosted preamble is never selected.
2. The headerless preamble includes `<stdint.h>`, `<stddef.h>`, `<stdarg.h>` instead of restating the fixed-width typedefs — the target's own `<stdint.h>` may spell `uint64_t` as `unsigned long` where V's text says `unsigned long long`, which is a typedef *conflict*, not a compatible redeclaration. It also includes the small hosted subset V's own runtime calls into. It then stops before the platform types: `FILE`, the POSIX structs and their prototypes are the target's to define, and a second `struct stat` conflicts with the real one rather than describing it. The `#ifndef`-guarded macros above that point stay.
3. A `fn C.xxx` in a file with no C include of its own gets a prototype. V does not drive the C build for such a target — the project compiles and links the generated C itself — so `#flag`/`#include` say nothing about where the symbol resolves from. A missing prototype is a hard error; a redundant one the project's header also declares is not.

## Effect

On the Vinix kernel: 80 host includes → 7, all satisfied by the kernel's own headers.

The kernel does **not** compile yet. 193 errors remain and none are about the preamble — they are V3 codegen gaps (fixed-array struct fields emitted as scalars, so `.id = (u64[4]){...}` initialises a `u64`; V functions called without a forward declaration) plus the kernel's own minimal libc headers, which declare neither `malloc` nor `qsort`. Those are separate fixes; this one unblocks reaching them.

## Checks

`./v self`, `v test`, and a module-resolution run are unchanged without the flag; every change is gated on it. `./v fmt -verify` clean on all five files. The flag is documented in `v help build-c`, including how it differs from `-freestanding`.


medvednikov:
Restructuring note, so this is reviewable.

This branch is stacked on #28575 (the `modules/` removal), and it had also accumulated five compiler fixes that have nothing to do with this flag — they were just bugs the kernel build hit next. Those are now split out into **#28591**, off master, reviewable and mergeable on their own:

- a fixed-array field's declared default emitted as a designator rather than copied
- markused pruning a method reached under its bare name
- an already-qualified imported type name being re-resolved into `gpu.agx.gpu.agx.…`
- `C.strlen` declared returning `i32`
- a `sizeof` selector not following a pointer

The commits still appear here because this branch predates the split; I have not force-pushed a re-stack, to avoid churning a branch that is already under review. Once #28591 lands I will rebase this one onto master and only the `-target-libc-headers` work will remain:

- add the flag
- propagate it to the parallel cgen workers
- alias the target's stdint types
- include `<math.h>`
- keep the thread runtime

For what the flag is still worth on its own: it takes the Vinix kernel's generated C from 80 host includes, failing at `'sys/un.h' file not found`, down to 7, all satisfied by the kernel's own headers.

medvednikov:
@codex review

medvednikov:
Re-reviewed the current head (CI ignored). I focused on target-header ownership versus generated `fn C.*` prototypes, thread/pthread type and runtime gating, preserved explicit directives, backend/cross validation, the cache salt and incremental-prefix refresh, and opaque C struct packing. The latest diff appears to address the open edge cases I could reproduce from the review history; I don't see an additional blocking issue.

PR #28580, opened 2026-09-13
v3: rename the shadowing `reported` binding, so V builds itself again

medvednikov:
master does not currently build itself.

`-dump-files` (#28568) added `mut reported := map[string]bool{}` for the build inputs that are not V sources, in the same function that already bound `reported` to the list `-print-v-files` prints. The bootstrap compiler accepts the shadowing; the compiler it then builds does not, so stage 2 fails:

```
vlib/v/driver/driver.v:10086:4: error: redefinition of `reported`
```

Reproduced on pristine `origin/master` with a plain `make`.

The inner binding is a different thing — the list to print, not the extra inputs to watch — so it is renamed to `listed` rather than merged into the outer map.

`make` and `./v self` both succeed with this.

Note: `vlib/v/driver/driver.v` is not vfmt-clean on master either (`v fmt -w` rewrites 266 lines), so I left the formatting alone rather than bury a two-line fix in churn.


medvednikov:
Fixed upstream in c5e5cb2925, which renames the outer binding to `additional_inputs` — same fix, better name. Closing.

PR #28585, opened 2026-09-13
checker: fix false "shadows a function declaration" notice

kimshrier:
fix false "shadows a function declaration" notice for bare C. fn aliases.

`vlib/v/transform/transform.v:3010` registers each C declaration
under two keys, C.name and the bare name, so return-type lookups
succeed whichever form the call site uses.

The shadow check at `checker_comptime.v:12506` cannot distinguish
    a V function named `close` exists
from
    the bare alias of fn `C.close` exists.

Every false positive follows from that one conflation.

After applying this change, the rename of `uname` to `unam` in
`cmd/tools/vself.v` lines 48 -55 is no longer necessary.


medvednikov:
Fixed in a different pr.

PR #28587, opened 2026-09-14
compiler: embed `$embed_file` payloads into portable `-os cross` output

medvednikov:
Fixes `make` dying at startup with:

```
V panic: EmbedFileData error: files "manual_stdlib_c_headers.h" and
"/home/runner/work/v/v/vlib/v/gen/c/manual_stdlib_c_headers.h" do not exist
```

### What was wrong

`$embed_file` only materialized the file's bytes for `-prod`. Every other build
recorded the file's absolute path and re-read it the first time `.data()` was
called, which is fine while the build stays on the machine that produced it.

Portable `-os cross` output is exactly the case where it is not fine: the
snapshot is compiled and run somewhere else. Since #28556 the compiler embeds
its own C header prelude, so the published `vc/v.c` contained

```c
c__manual_c_headers_source = embed_file__EmbedFileData__to_string(&((embed_file__EmbedFileData[]){
    (embed_file__EmbedFileData){.path = _str_2816, .apath = _str_2817, .len = 16166,
                                .compressed = NULL, .uncompressed = NULL, ...}})[0]);
```

with `.apath` pointing at the CI runner's checkout. That is a *const*
initializer, so anything bootstrapped from `vc/v.c` panicked before doing any
work, and `make` was broken for everyone.

### The fix

Portable output always embeds the bytes now, like `-prod` already did.

To keep that affordable, `c_byte_string_escape` escapes only what a C compiler
would otherwise misread (`"`, `\`, `?`, and non-printables as three digit octal)
instead of spending four characters on every byte, and splits its result into
adjacent string literals, because a single literal has an implementation defined
length limit that MSVC is strict about. The compiler's own snapshot grows by
~40KB, and the longest generated literal in it is under 3.8KB.

### Verification

* `./v -os cross -o vc.c cmd/v`, compiled with `cc`, then run with
  `vlib/v/gen/c/manual_stdlib_c_headers.h` moved out of the tree: `v version`
  and `v run examples/hello_world.v` both work. Before the change this is the
  panic above.
* New `vlib/v/embed_file/tests/embed_file_cross_snapshot_test.v` reproduces the
  panic on master and passes here: it cross-compiles a program embedding a file
  with quotes, a backslash, a trigraph, a NUL and high bytes, builds the
  generated C, deletes the source file, and checks the program's output.
* New unit tests in `vlib/v/gen/c/names_test.v` round-trip all 256 byte values
  through the escaper and check that a 40KB payload is split between escapes.
* `v test vlib/v/parser/` (53 passed), `v test vlib/v/embed_file/` and
  `v test vlib/v/gen/c/` show no change against the same tree without the patch.
* Self-host chain `v -> v2 -> v3` builds and runs.
* `-prod` and ordinary debug builds are unchanged.


medvednikov:
@codex review

medvednikov:
Addressed in 3340786 — thanks, the adjacent literals really were all landing on one line.

Each continuation now starts its own source line, and I lowered the chunk bound to 2048 so the text sharing the payload's first and last line still fits under the 4095 characters C99 5.2.4.1 requires (MSVC stops at 16384).

Writing the test for it turned up a second cause of the same symptom, which the first fix alone would not have covered: the payload is a string literal node, so every literal prescan interned it, and the literal table repeated the whole payload on one line — in a `_str_N` entry nothing ever references, because the embed codegen writes the bytes at the use site. A node flag keeps it out of the table. The compiler's own portable snapshot loses that 16KB duplicate, and its longest payload line is now 2251 characters.

Coverage, with a ~48KB payload (3x MSVC's line limit):

* `vlib/v/embed_file/tests/embed_file_cross_snapshot_test.v` asserts a bigger embedded file does not stretch the longest source line, measured against the same program embedding a small file. It fails at 68237 characters with the old space separator and at 52910 with the duplicate literal back, and passes now.
* The same file also checks the snapshot grows with the embedded file, and still compiles the generated C and runs it with the source file deleted.
* `vlib/v/gen/c/names_test.v` checks no escaped line reaches 4095 characters, and round-trips the payload back through a C-literal decoder.

The assertions are about what the snapshot is rather than how it is spelled, so they also hold under the V1 compatibility compiler that a directory test run can fall back to.

medvednikov:
Addressed in e38c46e — you're right, the join happens before the limit applies, so splitting was never going to get past it.

Past 32768 bytes the payload is now written as a file-scope `static const unsigned char _v_embed_blob_<node>[]`, and the initializer points at it. An array initializer has no length limit. The object is `static` rather than external because `-parallel-cc` repeats the declaration prefix in every unit.

I kept the string literal form below the threshold: it costs about one character per byte for text against five for an array, and the compiler's own 16 KB C header prelude — the case that decides `vc/v.c`'s size — is well under it. That snapshot is unchanged: still 9 lines of at most 2251 characters, no blob object.

Coverage, with a 96 KiB payload (past what any literal can hold):

* The test scans the generated C for the longest run of adjacent string literals, joining them as a compiler would and decoding escapes so the count is the resulting byte length, then asserts it stays under 65535. Raising the threshold so that payload takes the literal path makes it fail at exactly 98363 bytes, the payload's own length.
* A 16 KiB payload covers the literal path next to it, and both sizes are compiled with `cc` and run with the embedded file deleted.
* `names_test.v` pins the threshold boundary.

Also checked directly: a 200 KB embed under `-prod`, `-prod -no-parallel` and `-prod -parallel-cc` all produce a matching SHA-256, and the cross snapshot of `cmd/v` still bootstraps with `manual_stdlib_c_headers.h` removed from the tree.

medvednikov:
Addressed in 25a3750. Agreed — the cutoff was read off a particular compiler's figure, which is exactly what portable output cannot do.

It is now 4095, the C99 5.2.4.1 minimum for a literal after concatenation, so anything a conforming compiler is not obliged to accept goes to the array representation.

The compiler's own 16 KB C header prelude crosses that line, so `vc/v.c` now carries it as `_v_embed_blob_<node>[16166]` over lines of 100 characters. The snapshot grows by 64141 bytes, 0.3% of 21.9 MB, which seems a fair price for the contract in `doc/docs.md` holding.

The tests gained a payload in the range this is about — 8 KiB, past what C guarantees and nowhere near what any one compiler allows. Restoring the old 32768 cutoff now fails them at 8251 joined bytes, where before they would have passed.

One thing that surfaced while writing that assertion, and which I have **not** touched: the baseline it measures against is 7856, not 4095. Ordinary V string constants — vlib's Unicode tables among them — are already interned into `_str_N` entries past the guarantee, with no `$embed_file` involved. So the same contract is already broken independently of this PR, and `write_string_literal_def` would need the same treatment to close it. I have kept this change to embeds and held the payload to not making the longest literal any longer than it already was; happy to open a separate issue or PR for the general case if you would like.

Re-verified: a 200 KB embed under `-prod`, `-prod -no-parallel` and `-prod -parallel-cc` all produce a matching SHA-256; the cross snapshot of `cmd/v` still bootstraps with `manual_stdlib_c_headers.h` removed from the tree; self-host chain, parser 53/53, flat 12/12, `names_test.v` 43/43, and `vlib/v/embed_file/` at its master baseline.

medvednikov:
Addressed in fb93b57. Right — the array representation had a limit of its own, and one oversized object is no more portable than the literal it replaced.

Past 65535 bytes the payload is now emitted as several objects plus a table of `embed_file.EmbedFileChunk`, which `data()` joins into one buffer the first time the file is asked for. For a 96 KiB payload:

```c
static const unsigned char _v_embed_blob_35222_0[65535] = { ... };
static const unsigned char _v_embed_blob_35222_1[32828] = { ... };
static const embed_file__EmbedFileChunk _v_embed_blob_35222[3] = {
{.data = (u8*)_v_embed_blob_35222_0, .len = 65535},
{.data = (u8*)_v_embed_blob_35222_1, .len = 32828},
{.data = NULL, .len = 0},
};
...
.uncompressed = NULL, .chunks = (embed_file__EmbedFileChunk*)_v_embed_blob_35222,
```

The struct init names both `uncompressed` and `chunks` from the same payload node, so the backend fills whichever one it picks and neither is left to a default it would then have to override.

**The tradeoff, since it is a behaviour change and yours to weigh:** a payload past the limit now costs one allocation and one copy on first `.data()`, where it used to be a pointer into `.rodata`. Peak memory for such an embed is briefly doubled. Below the limit nothing changes — the compiler's own 16 KB header prelude stays a single object, and `vc/v.c` is unaffected. If you would rather keep the single-object form and treat 65535 as a limit no supported compiler enforces, this commit is the one to drop; V has emitted oversized objects since V1, so that would be the status quo.

Coverage: the test asserts no declared array in the output grows past 65535, and fails at 98363 with the split removed. That one assertion is skipped when the probed compiler's own standard library has no chunk table, because a test run can fall back to the V1 compatibility compiler, which emits one object of any size — I checked that the gate does not mask the regression. The other four assertions in the file still hold for either backend.

Also verified directly: a 200 KB embed under `-prod`, `-prod -no-parallel` and `-prod -parallel-cc` all match on SHA-256; `.to_string()` and `.to_bytes()` on a chunked embed agree with each other and with the file; the cross snapshot of `cmd/v` still bootstraps with `manual_stdlib_c_headers.h` removed from the tree.

medvednikov:
This finding was already addressed in fb93b57, pushed before this review landed — the comment is anchored at `struct.v:0`, so I think it was generated against the previous snapshot.

Verified on the current branch head. For the 96 KiB test payload, every object the generated C declares:

```
  _v_embed_blob_35222_0   unsigned char                entries=65535   object bytes=65535
  _v_embed_blob_35222_1   unsigned char                entries=32828   object bytes=32828
  _v_embed_blob_35222     embed_file__EmbedFileChunk   entries=3       object bytes=48
  payload bytes covered by the pieces: 98363
```

and the largest array declared anywhere in that file is exactly 65535. `data()` joins the pieces on first use. There is a regression test for it (`test_cross_output_never_declares_an_over_large_object`), which fails at 98363 if the split is removed.

One thing the split does not remove, which I have documented in 2d7ebf5 rather than engineered around: the chunk table is an object too, and at 16 bytes an entry it holds 4095 of them before reaching the same ceiling — about 256 MB of embedded file. Any single contiguous index has some such bound; removing it would need a chain rather than a table, and V already warns about embedded files several orders of magnitude smaller than that. Happy to take it further if you would rather have no bound at all.

medvednikov:
Both addressed in 038f6bc. They turned out to have one cause between them — the lazy materialization — so the fix for the first removes the machinery the second was about.

**The race.** You are right that this was a regression: a production embed used to hand back its static pointer and mutate nothing, and lazily filling `uncompressed`/`free_uncompressed` on first read put two threads on the same shared constant. Rather than lock it, I took the other option you offered and materialized it before concurrent access is possible: `join_chunks` is now a pure function called from the initializer, so the bytes are contiguous before the value exists and `data()` is back to not mutating anything. For a constant that runs in `_vinit`, before any thread. The `EmbedFileData.chunks` field is gone along with the lazy path that needed it.

```c
.uncompressed = embed_file__join_chunks((embed_file__EmbedFileChunk*)_v_embed_blob_35222, 98363),
```

The cost, stated plainly: a payload past the limit is now joined at startup whether or not it is read, one allocation and one copy. Below the limit nothing is allocated at all.

**The table.** Also right, and my note in 2d7ebf5 had the arithmetic one entry too generous — 4095 data chunks plus the sentinel is 65536, not 65520. A table now holds 4095 entries at most and spends its last one either on the terminator or on a link to the table that continues it, so no payload size needs an object C is not required to accept. A zero length entry that still carries a pointer is the link; one with a null pointer is the end.

**How the chained path was checked**, since it otherwise only triggers past ~256 MB: with the limits temporarily lowered to 1024-byte pieces and 4-entry tables, the same 96 KiB payload becomes 97 data objects across 33 linked tables, and reassembles byte for byte. Four threads reading it concurrently are clean under ThreadSanitizer — and I confirmed TSan was live by checking it flags a deliberate race built the same way.

Tests: the cross snapshot probe program now reads the embed from four threads and compares what they see, so the shared-constant pattern is covered end to end; `names_test.v` pins the split arithmetic at its boundaries, including that a table's entries stay within the object limit. 44/44 and 5/5. Also re-verified 200 KB embeds under `-prod`, `-prod -no-parallel` and `-prod -parallel-cc`, the `cmd/v` snapshot bootstrapping with its header removed from the tree, and the self-host chain.

medvednikov:
Addressed in f8b53f2, taking the first of your two options — materialize once and reuse the pointer.

The join now runs at the top of `_vinit`, into a buffer the expression only reads:

```c
extern u8* _v_embed_joined_35223;            /* declaration prefix */
u8* _v_embed_joined_35223 = NULL;            /* next to _vinit */
void _vinit() {
	_v_embed_joined_35223 = embed_file__join_chunks((embed_file__EmbedFileChunk*)_v_embed_blob_35223, 98363);
	...
.uncompressed = _v_embed_joined_35223,
```

That also preserves what the previous commit was after: it happens before any thread exists, and nothing is mutated afterwards, so `free_uncompressed` staying false is now correct rather than a leak — the buffer is shared program-wide, exactly like the static pointer an unsplit payload hands out.

The buffer has external linkage and is defined next to `_vinit` rather than in the declaration prefix, because `-parallel-cc` repeats that prefix in every unit and a `static` pointer would leave every unit but the one running `_vinit` holding its own null copy. Both the declaration and the `_vinit` line are derived from the AST, so the tail worker that generates `_vinit` from its own `FlatGen` arrives at the same list.

Measured, 20000 calls to a function embedding a 96 KiB file, `-prod -gc none`:

| | peak RSS |
|---|---|
| join in the expression | 2,304,606,208 bytes |
| joined once in `_vinit` | 1,753,088 bytes |

The probe programs now also check that repeated evaluations of an `$embed_file` written inside a function return the same pointer. That is portable and free of any assumption about spelling, and it fails with `the payload was materialized again on a later evaluation` if the join goes back into the expression.

Re-verified: `-prod`, `-prod -no-parallel`, `-prod -parallel-cc` and `-os cross` all reassemble the payload correctly, four concurrent readers stay clean under ThreadSanitizer, the forced-chaining check (1024-byte pieces, 4-entry tables, 97 objects across 33 linked tables) still reassembles byte for byte, and the `cmd/v` snapshot bootstraps with its header removed from the tree. 44/44, 53/53, 12/12, 5/5.

medvednikov:
Both addressed in 32b7daf.

**Prealloc.** Confirmed before fixing — under `-prealloc` the old `join_chunks` emitted `v_malloc(...)` as the first statement of `_vinit`, with `builtin_init()` 26 lines later. I took your second option and kept the buffer out of the resettable arena rather than reordering, since ordering alone would still leave a program-lifetime buffer inside something that gets reset, and a `__global` default reading the payload would run even earlier than `builtin_init()`:

```c
/* -prealloc */      u8* buffer = (u8*)(malloc((size_t)(size)));
/* otherwise */      return (u8*)(v_malloc((ptrdiff_t)(size)));
```

**Deduplication.** Also confirmed, and worse than the node count suggests. The registry is now keyed by content — the first node carrying those bytes names the objects — so identical payloads share one copy, the way the interned literal table did before this took over from it. For a 96 KiB file embedded five times, two as constants and three as specializations of one generic:

| | byte objects | joined buffers | generated C |
|---|---|---|---|
| before | 10 | 5 | 2,977,539 bytes |
| after | 2 | 1 | 988,599 bytes |

Tests: the probe programs gained a second constant embedding the same file and check that both reach the same pointer — portable, and independent of how any of this is spelled. `-prealloc` gets a probe of its own, generated, compiled and run with the embedded file deleted. That one is skipped where the probed compiler has no chunk table, since a compiler that embeds the payload as one object never reaches for the allocator during `_vinit` at all.

Re-verified: `-prod`, `-prod -no-parallel`, `-prod -parallel-cc`, `-prod -prealloc` and `-os cross` all reassemble correctly; the 20000-call loop still peaks at 1,753,088 bytes; the `cmd/v` snapshot bootstraps with its header removed; self-host chain; 44/44, 53/53, 12/12, 6/6, and `vlib/v/embed_file/` back at its master baseline.

Unrelated, but found on the way: `-prealloc` together with `spawn` and an `$embed_file` segfaults under the 0.5.2 compatibility compiler that a directory test run can fall back to. Nothing to do with this branch — that compiler has no chunk path — but it is why the prealloc probe is gated rather than simply skipped on Windows.

medvednikov:
Addressed in 739c38b. Confirmed the instability first — the same 96 KiB payload, embedded by two programs that differ only in unrelated code around it:

```
before:  dup.v -> _v_embed_joined_35250     shift.v -> _v_embed_joined_35234
after:   both  -> _v_embed_joined_98363_666a5f620b1f7c5c
```

The names now come from the payload itself, in the shape you pointed at: `cache_string_symbol` and this one share a single `content_symbol_suffix`, so both are a length next to a hash of the bytes, and the two cannot drift apart.

Content addressing also subsumes the deduplication from the previous commit, so the scan for "the first node carrying this payload" is gone — the emitters just keep one set of objects per distinct name. The `static` byte objects and chunk tables are named the same way; they have internal linkage so they were never a link hazard, but it makes the generated C the same whatever the AST looked like, which is what the cache's content addressing wants anyway.

The test generates the same embed from two programs with different surrounding declarations and compares the names each imports as `extern u8*` — which is what a cached object file would carry, and does not depend on the prefix I happened to choose. It fails against the previous commit with `the joined buffer is named after where the payload sits rather than after what it holds`.

Re-verified: `-prod`, `-prod -no-parallel`, `-prod -parallel-cc`, `-prod -prealloc` and `-os cross` all reassemble correctly; the five-way dedup still collapses to 2 objects and 1 joined buffer; the forced-chaining check (97 objects, 33 linked tables) still reassembles byte for byte; the 20000-call loop still peaks at 1,753,088 bytes; the `cmd/v` snapshot bootstraps with its header removed; self-host chain; 44/44, 53/53, 12/12, 7/7, and `vlib/v/embed_file/` at its master baseline.

I did not set up an actual two-program module-cache build to watch the link succeed — the property the cache needs is the name agreeing, and that is what the test pins.

medvednikov:
Fixed in 11dcb95. Reproduced your command first:

```
$ v -prealloc -Wimpure-v -W conc.v
vlib/v/embed_file/embed_file.v:28:18: error: C code will not be allowed in pure .v files,
please move it to a .c.v file instead
   28 |             buffer := &u8(C.malloc(usize(size)))
```

The `C.malloc` now lives in a new `vlib/v/embed_file/embed_file.c.v`, behind `join_chunks_raw_alloc`, which `join_chunks_buffer` calls from the preallocator branch. Both that command and `-prod -prealloc -Wimpure-v -W` now build clean and run correctly, as does the `v -W -Wimpure-v examples/hello_world.v` that `vtest-all` runs.

I checked the one thing this could have broken, since the module has no `.js.v`: `-b js` refuses `$embed_file` before reaching any of this, with the same message before and after the change, so nothing regresses there.

Re-verified the rest: `-prod`, `-prod -no-parallel`, `-prod -parallel-cc`, `-prod -prealloc` and `-os cross` all still reassemble the payload correctly, the `cmd/v` snapshot bootstraps with its header removed, self-host chain, 44/44, 53/53, 7/7, and `vlib/v/embed_file/` at its master baseline.

medvednikov:
Fixed in e4e626d, with one correction to the report.

You are right about the constructors: both rebuilt `flags` from `node_flags(...)`, which only knows the two bits that predate this branch, so a cloned node silently lost `node_flag_embed_payload`. That is now `clone_node_flags`, which keeps the bits describing the node itself and takes the scope-dependent one from the caller — so a flag added later has one place to be considered rather than two to be forgotten.

**Where I could not confirm the report**: I could not construct a case that actually reaches those constructors with a payload node. A generic body, a `$for` inside a generic, and a closure inside a generic all keep the payload in its own object, with nothing interned beyond the 7856-byte Unicode tables vlib already carries. Building a compiler with the marker deliberately dropped and diffing the generated C for a generic embedding 20 KB:

```
$ diff gen.c gen_drop.c
35611c35611
<     const char* v3_vexe = ".../v";
---
>     const char* v3_vexe = ".../vdrop";
```

That is the whole difference — the embed codegen is byte-identical either way, so the specialized payload is not in fact rejected today. My reading is that the payload node is carried by child id rather than re-created along these paths. The change is still right as an invariant, but I would not want it recorded as a live miscompilation, and the new test does not fail without it.

The test you asked for is there regardless: the probe programs now put an `$embed_file` inside a generic and call it at three types, so the existing literal-length, source-line, object-size and pointer-identity assertions all cover the specialized copies.

`flat.v` and `monomorphize.v` are core enough that I checked beyond the embed tests: `vlib/v/transform/` fails identically before and after (`transform_parallel_test.v:919`, pre-existing on the previous commit), and `vlib/v/flat/` 12/12, parser 53/53, `names_test.v` 44/44, cross snapshot 7/7. Plus the usual: all five build modes, a generic with a 96 KiB embed under `-prod`, the `cmd/v` snapshot bootstrapping with its header removed, and the self-host chain.

PR #28594, opened 2026-09-14
v3: finish the outstanding #28568 review fixes

medvednikov:
Everything still outstanding from the #28568 review. Three of these were pushed to that PR after it had already been squash-merged, so they are not in master; the other three were never implemented. All rebased onto current master and verified there.

### Preserve the nil operand's block scope

The statements a `nil` operand carries were spliced into the enclosing pending list one by one, discarding the block they were written in. A block-local declaration landed in the enclosing C scope, so this stopped compiling with `redefinition of 'local'`:

```v
opt == unsafe { local := 1; use(local); nil }
...
local := 2
```

The lowered effects are wrapped in a block now, one per source block, so each level keeps its own scope. Collection runs against an empty pending list, because lowering a statement drains whatever is pending: without that, the temp the option is pinned to for left-to-right evaluation was swept into the block and went out of scope before the comparison that reads it.

### Preserve the stable executable name for cached tools

The content address was in the executable's file name, so `os.executable()` reported `vshader-<hash>`. Tools use that basename as their displayed name and as their own cache directory: `cmd/tools/vshader.v` keys `~/.cache/v/<name>` on it, so every compiler or tool change downloaded another copy of `sokol-shdc`, while `vbump` and `vdownload` printed the hashed name in their help.

The hash now names a per-entry directory and the executable inside keeps the tool's name, so storage stays content addressed while the tool keeps its identity.

### Invalidate failures when nested modules reappear

An unresolved module contributes no source file to the manifest, so the only record of it is where it would have lived. Only `vlib` and `vlib/v` were stamped, and a root records its direct children: `db.sqlite` could be removed and restored with both reading the same, replaying the failure forever. Unresolved module names are read out of the failure text now, and the module's directory plus every ancestor is stamped, under every root the compiler searches — `vlib` and each `$VMODULES` entry. `VMODULES` also joins the ambient build variables.

### Record missing embed paths in manifests

`$embed_file` on an absent asset compiles; the binary just carries no payload. The dump skipped the path precisely while it was missing, so restoring the asset never invalidated the binary built without it. It is recorded now whether or not it exists.

That alone would have rebuilt forever: `last_modified` cannot date a file that does not exist and reports the end of time, which the freshness check reads as "changed" on every lookup. An input that stamps as missing skips that check.

### Serialize binary and manifest publication

Publishing the executable and the manifest are separate steps, so two processes rebuilding one key across a source edit can interleave and leave the older executable vouched for by the newer manifest — after which every invocation runs stale code.

The manifest now names the exact build it describes, taken from the staged file before publication since a rename carries the identity across. Device and inode make it exact wherever they are real; elsewhere it degrades to size and mtime, which still separates builds of different sources. A reader that finds a mismatch rebuilds.

### Cover the suite-hook file results

The fix shipped in #28568 without a test. A plain `assert false` in a hook aborts the process, so the test uses `@[assert_continues]` — the path that records a failure and keeps going, and therefore the only one that could report `OK` for a file while the process exits nonzero.

## Testing

- `cmd/v/toolcache_test.v` 32/32, `cmd/v/launcher_test.v` 5/5, `test_suite_hook_file_result_test.v` 3/3, `option_nil_compare_block_test.v` 10/10.
- Full `vlib/v/tests` sweep against a current-master baseline: no new failures. The one test outside the baseline (`generics/generic_typeof_idx_test.v`) fails on plain master too when run on its own; it is load-sensitive.
- `./v` self-hosts twice. Warm `v fmt -verify` stays at ~30ms against 4.0s cold.

The new tests are checked against a pre-fix compiler where one exists, and assert the premise rather than only the outcome: the nested-module ones verify the previously-stamped roots genuinely cannot see the change, so they fail if the ancestor stamping is trimmed back rather than merely if the result changes.


medvednikov:
@codex review

medvednikov:
Reviewed the current head (CI ignored). I rechecked nil-operand scope preservation and the tool-cache paths around publication/locking, UID namespacing, no-follow recursion, binary/manifest pairing, and missing-module invalidation (including aliases and `v.mod` source roots/subdirs). The current diff appears to cover the edge cases from the review history; I don't see another blocking code issue.

medvednikov:
@codex review

PR #28595, opened 2026-09-14
compiler, sync: fix the portable `-os cross` snapshot on macOS and Windows TCC

medvednikov:
The published `vc` snapshot stopped bootstrapping on two platforms. Both failures are in the C that `-os cross` generates, before any V code runs.

## macOS — `sem_timedwait`

`clang-macos (macos-15)`, `clang-macos (macos-26)` and `v-up-works-nix (macos-15-intel)` all fail in `make`:

```
./vc/v.c:209974:7: error: call to undeclared function 'sem_timedwait'; ISO C99 and later do not support implicit function declarations
```

`-os cross` defers target-dependent `$if`s to the C preprocessor, but it still bakes in the *generating* host's platform files. The snapshot is generated on Linux, so it carries `sync_default.c.v`'s POSIX `Semaphore`, and Apple's libc has no `sem_timedwait` symbol at all — adding a declaration only moves the failure to the linker:

```
Undefined symbols for architecture arm64:
  "_sem_timedwait", referenced from: ...
```

The call became reachable when `sync.Semaphore.timed_wait` was added to the `markused` roots for parallel compiler builds; before that it was eliminated and the snapshot happened to link.

`timed_wait` now has an Apple branch that polls `sem_trywait` until the deadline, so the `sem_timedwait` call sits behind an `#if defined(__APPLE__) ... #else` that is preprocessed out there:

```c
bool sync__Semaphore__timed_wait(sync__Semaphore* sem, i64 timeout) {
	#if ((defined(__APPLE__) && !defined(__ENVIRONMENT_IPHONE_OS_VERSION_MIN_REQUIRED__)) || ...)
	{ ... sem_trywait ... }
	#else
	{ ... sem_timedwait ... }
	#endif
}
```

## Windows — atomics redefinition

`more-tests (windows-latest)` fails compiling the msvc-generated snapshot with tcc:

```
./vc/v_win.c:9894: error: redefinition of 'ManualInterlockedExchangeAdd8'
```

`atomic_builtin_compat_decls` chose between V's `static inline` helpers and V's WinAPI atomic header from `g.ccompiler` — something a portable snapshot does not know yet, since `vc/v_win.c` is generated with `-cc msvc` and compiled with tcc. That header defines `atomic_fetch_add_byte` and friends as function-like macros, so under tcc

```c
static inline byte atomic_fetch_add_byte(void* ptr, byte delta) { ... }
```

expanded into a redefinition of the header's own `ManualInterlockedExchangeAdd8`.

The helpers now sit behind the negation of the `_WIN32 && __TINYC__` guard that `system_libc_headers()` already uses for that header, so the preprocessor picks. The early return is skipped in cross mode as well, so `-cc tcc` and `-cc msvc` produce the same portable output instead of one that only works under the compiler it was generated for.

## Not included

`periodic-network (windows-latest, msvc)` fails with

```
EmbedFileData error: files "manual_stdlib_c_headers.h" and "/home/runner/work/v/v/vlib/v/gen/c/manual_stdlib_c_headers.h" do not exist
```

That is already fixed on master by #28587 — current `master` embeds the payload as `_v_embed_blob_16166_…`. It only still fails because `vc` is pinned at `9815bf260a`, which predates that commit.

## Testing

- The Linux-targeted portable snapshot now compiles **and links** on macOS; before this it reproduced the CI error exactly.
- Full bootstrap chain runs: snapshot → `v1` → `v2` → `v`, and the result compiles programs.
- Non-cross output is **byte identical** to before for the host target and for `-os windows` with both `-cc tcc` and `-cc msvc`; only the `-cross` path changes.
- Two regression tests added to `cross_output_codegen_test.v`, each verified to fail when its fix is reverted.
- Green: `source_directive_test` (24), `preamble_test` (25), `names_test` (44), `gen/c/target_test` (48), `pref/target_test` (20), `embed_file_cross_snapshot_test` (7), `vlib/v/tests/concurrency/` (70, identical to baseline).


medvednikov:
@codex review

PR #28618, opened 2026-09-14
Create 1brc.v

changrui:


quaesitor-scientiam:
Thanks for the submission, but as it stands this cannot be merged:

1. **It breaks the existing example.** `examples/1brc/solution/main.v` already exists (same structure: `mmap_file`, `process_chunk`, `combine_results`, `process_in_parallel`, `main`). This PR adds a second file with its own `fn main()` and its own `C.mmap`/`C.munmap`/`MemoryMappedFile` declarations to the same directory, and V compiles a directory as one module. With both files present, `v -check examples/1brc/solution` fails:

   ```
   solution/main.v:8:6: error: C function `C.mmap` was already declared with a different signature in module `main` at solution/1brc.v:18:6
   solution/main.v:9:6: error: C function `C.munmap` was already declared with a different signature in module `main` at solution/1brc.v:19:6
   ```

   The `// vtest build: !windows` header means CI will build this directory on Linux/macOS, so this would turn the examples job red.

2. **The file does not compile on its own either.** Its `fn C.open(path &u8, flags int, mode int) int` conflicts with vlib's declaration:

   ```
   vlib/os/os.c.v:29:6: error: C function `C.open` was already declared with a different signature in module `main` at 1brc.v:24:6
   ```

   (`fn C.open(&char, i32, ...int) i32` is already provided by `os`; just drop the local redeclaration, or use `os.open`.)

3. **Not `vfmt`'d** — `v fmt -verify examples/1brc/solution/1brc.v` fails, which is a hard CI check.

4. The title is the GitHub default ("Create 1brc.v") and the description is the unmodified PR template, so it is not clear what this is meant to be relative to the solution that is already there. If the goal is a faster solver, please either improve `solution/main.v` in place, or put this in a separate directory with a README note explaining the difference — and in either case make sure `v examples/1brc/solution` and `v fmt -verify` pass first.

A few smaller things to look at once it compiles: `e.max = 0x80000000` on an `i32` field, `find_or_insert` loops forever if the 4096-slot table ever fills (the trailing `return &m.entries[0]` is unreachable), and the machine-specific timing comments in the header are not something the example should carry.

JalonSolov:
... and there's already `examples/1brc`...

JalonSolov:
If you can make a better/faster 1brc, that's fine, but it would be a PR against the existing example.

PR #28622, opened 2026-09-14
v3: resolve selectively imported ORM table types

### LINKED ISSUE #28576: V3: ORM statements on selectively-imported structs are silently ignored
### Describe the bug

Under V3, ORM statements referring to a struct by its **selectively-imported** name (`import mod { T }` … `sql db { select from T }`) are silently dropped: no SQL is emitted, no error is returned, and error propagation succeeds. Writing the exact same statement with the qualified name (`mod.T`) works.

This is a silent-wrong-results failure rather than a compile or runtime error: `create table` reports success while creating nothing, and `select` returns zero rows from a table that contains data. The established compiler handles both spellings identically.

### Reproduction Steps

`ent/ent.v`:

```v
module ent

pub struct Remote {
pub mut:
	id   int @[primary; sql: serial]
	name string
}
```

`main.v`:

```v
module main

import db.sqlite
import ent { Remote }

fn main() {
	sql db {
		create table ent.Remote
	}!
	r := Remote{
		name: 'a'
	}
	sql db {
		insert r into ent.Remote
	}!
	qualified := sql db {
		select from ent.Remote
	}!
	unqualified := sql db {
		select from Remote
	}!
	println('rows via ent.Remote (qualified) : ${qualified.len}')
	println('rows via Remote     (selective) : ${unqualified.len}')
}
```

```
v run .
v -old-compiler run .
```

### Expected Behavior

Both spellings refer to the same struct and the same table, so both selects must return 1 row — which is what the established compiler does:

```
rows via ent.Remote (qualified) : 1
rows via Remote     (selective) : 1
```

### Current Behavior

V3 returns an empty result for the selectively-imported spelling, with no error:

```
rows via ent.Remote (qualified) : 1
rows via Remote     (selective) : 0
```

`create table` is affected the same way. With `ent/ent.v` as above:

```v
import ent { Remote }

pub struct Local {
pub mut:
	id   int @[primary; sql: serial]
	name string
}

fn main() {
	sql db { create table Local }!
	sql db { create table Remote }!
	tables := db.exec("select name from sqlite_master where type='table' order by name")!
	for row in tables { println('table: ${row.vals[0]}') }
}
```

V3 prints only `table: local`; the established compiler prints `table: local` and `table: remote`. No error is raised for the missing statement.

### Possible Solution

The ORM lowering introduced in #27813 likely resolves the table type by unqualified name within the current module, and drops the statement when the symbol resolves to a type imported from another module. A resolution failure there should at minimum be a compile error rather than a silently skipped statement.

Selective imports are idiomatic V, and separating entity structs into their own module is a common layout for any application with a database layer. In such a project the effect is that the whole data layer becomes a no-op under V3: migrations create no tables yet report success, subsequent statements fail against the missing tables, and reads return empty through the usual `or { [] }` fallbacks — so the application starts and serves requests with no data, no log line and no error. The silent nature makes it far more damaging than a compile failure would be.

### V version

`V 0.5.2 ad15cea`

- macOS 26.6, arm64 (Darwin 25.6.0)
- Apple clang 21.0.0
- V3 compiler (default); `-old-compiler` behaves correctly

> [!NOTE]
>

medvednikov:
Fixes #28576.

V3 ORM lowering now resolves a bare table type through the current file's selective imports before looking up its struct metadata. This prevents `sql` statements using `import mod { Type }` from being silently lowered to fallback values.

The SQLite regression covers create, insert, select, and drop statements using the selectively imported name.

Tests:
- `./vnew -no-parallel -silent test vlib/v/tests/orm_selective_import_test.v`
- `./vnew -no-parallel vlib/v/tests/orm_type_alias_test.v`
- `./vnew -no-parallel vlib/v/tests/orm_generic_struct_select_test.v`
- `./vnew -no-parallel vlib/v/transform/orm_test.v`
- `git diff --check`


medvednikov (commented):
Reviewed the current code changes only; CI status intentionally not considered. I did not find a concrete correctness issue. The selective-import resolution is applied before ordinary imported-name resolution, preserves already-qualified names, and recurses through generic table arguments; the regression covers create/insert/select/drop using the selectively imported table name.

PR #28627, opened 2026-09-15
regex: allow replace on constant regexes

medvednikov:
## What this does

- allows `regex.RE.replace` to be called on immutable and constant compiled regexes
- isolates the matcher counters, captures, and backtracking buffers used during replacement
- adds a regression test for capture substitution with a constant regex and updates the README example

The matcher needs mutable scratch state internally, but replacement should not mutate the compiled regex exposed to callers.

## Tests

- `./vnew -new-compiler vlib/regex/regex_const_test.v`
- `./vnew -old-compiler vlib/regex/regex_const_test.v`
- `./vnew -silent test vlib/regex/` (21 passed)
- `./vnew check-md vlib/regex/README.md` (reports four pre-existing formatting errors in unrelated examples at lines 208, 425, 680, and 787)


medvednikov:
@codex review

medvednikov (commented):
Code review only; CI intentionally ignored.

No blocking issues found. Changing `replace` to an immutable `&RE` receiver and doing the matching through a local worker is consistent with the matcher internals: the mutable matcher-owned state (`prog` counters, captures, state/backtracking buffers, group stacks/data, and continuous-save buffer) is isolated before `find_from` runs, while the compiled char-class/name metadata is only read during matching/replacement.

One non-blocking coverage suggestion: the regression test currently proves a fresh const regex keeps `groups.len == 0`. It would be useful to also pre-populate capture state with a match, call `replace`, and assert the pre-existing group bounds/captures are unchanged. That directly locks in the new “replace does not mutate caller-visible matcher state” behavior.

PR #28628, opened 2026-09-15
cgen: cache active macro header directives

medvednikov:
## What does this PR do?

Cache the preprocessed directives and safe conventional include guards for C headers while replaying active macro state.

Repeated includes can now skip the guarded prefix while still replaying compatibility directives after the closing `#endif`. Headers that undefine their guard, use `pop_macro`, or have guard-level alternative branches continue to be fully replayed. Conditions nested in inactive branches are no longer evaluated.

This fixes the self-build regression introduced by `9546543f7ba3853ec55488000e580f6ebe11719b`, where every transitive C header was reparsed on every include. A profile showed 4,016 scans of 292 unique headers, mostly Apple Availability headers.

macOS self-build results:

| Metric | Before | After |
| --- | ---: | ---: |
| Wall time | 23.39 s | 5.80 s |
| Peak RSS | 7.41 GB | 0.95 GB |
| `collect_info` | 11.05 s | 0.60 s |

## Tests

- `./vnew -silent test vlib/v/gen/c/source_directive_test.v` (49 passed)
- Safe self-build to a separate output path
- `./vnew -silent test vlib/v/gen/c/` (312 passed; one existing `target_test.v` failure reproduced on an unmodified base worktree)

The broader compiler golden-output suites currently encounter existing V3 fallback output mismatches on the base revision.


medvednikov:
@codex review

vlib/v/gen/c/cleanc.v
```diff
@@ -8861,21 +8909,61 @@ fn (mut g FlatGen) collect_included_c_active_macros_from_file(path string, inclu
...
+		// A false dynamic value means a conditional `#undef` made the guard ambiguous.
+		if guard in g.direct_c_include_macros || g.direct_c_dynamic_macros[guard] {
+			first_directive = guard_end + 1
+		}
+	}
+	for directive_idx in first_directive .. directives.len {
+		clean := directives[directive_idx]
 		directive_name := c_directive_name(clean)
 		directive_is_active, directive_is_ambiguous := c_active_macro_directive_state(clean, mut conditionals, mut g.direct_c_include_macros, mut g.direct_c_dynamic_macros, g.c_compiler_macro_env_complete)
+		if guard.len > 0 && directive_idx == guard_end && !ambient_ambiguous {
```
medvednikov: This invariant is not safe when a nested include mutates the outer guard. For example, `guarded.h` can do `#ifndef G` / `#define G`, then `#include "inner.h"` where `inner.h` does `#undef G`, followed by `#define guarded_api(...) ...` and the closing `#endif`. The recursive scan correctly removes `G`, but at this `#endif` this block adds it back. If `guarded_api` is then undefined and `guarded.h` is included again, the real C preprocessor re-enters the guarded body because `G` is undefined, while this scanner takes `first_directive = guard_end + 1` and misses the redefinition. `c_header_guard_end_directive` only proves that this file does not directly undef/pop the guard; it cannot prove that transitive includes leave it defined. Please preserve guard mutations observed while replaying nested includes (or conservatively mark such guards unsafe) and add a regression test for this case.

medvednikov (commented):
Found one correctness issue in the include-guard cache: a nested include can undefine the outer guard, but the new guard-end invariant restores it and causes later includes to skip a body that the C preprocessor would execute. See the inline comment for a concrete reproducer. CI status intentionally not considered.

PR #28629, opened 2026-09-15
cgen: bound active C macro header scans

medvednikov:
## Summary

- release per-include C macro scan scratch allocations from the V3 preallocated arena
- cache safely skippable whole-header conditionals, including modern Clang guard expressions
- cover complex guards and guard-level alternative branches

## Root cause

The macOS Clang SDK header graph is revisited many times while replaying active C macros. With V3 preallocation, temporary header contents and scan state stayed in the cgen arena until generation finished. The no-fallback self-host reached about 28.5 GiB locally and CI killed it at the 9,984 MiB limit.

Each source-level include scan now uses a disposable nested prealloc scope. Persistent macro state is promoted before that scope is released. A conservative cache also skips repeated whole-header conditionals only when their opening condition is known to be inactive; outer else/elif branches are excluded.

## Validation

- ./vnew -silent vlib/v/gen/c/source_directive_test.v
- ./vnew -d prealloc -silent vlib/v/gen/c/source_directive_test.v
- V_MACOS_V3_NO_FALLBACK=1 ./vnew -nocache -o ./vfinalcheck -gc none -cc clang cmd/v
  - completed in 65.15s
  - peak physical footprint: 9,351,700,592 bytes, about 1.04 GiB below the CI limit
- ./vnew -silent test vlib/v/gen/c/: 311 passed; target_test.v has one unchanged current-master failure in test_cache_input_scan_rejects_ambiguous_include_macro_literal

The full compiler error suite currently has unrelated current-master V1 fallback/V3 diagnostic expectation mismatches; the same mismatches reproduce in a pristine master worktree.


medvednikov:
@codex review

medvednikov:
fixed in https://github.com/vlang/v/pull/28628

PR #28633, opened 2026-09-15
cmd/v: declare renameat in generated C

medvednikov:
## What does this PR do?

Mark the existing `fn C.renameat(...)` declaration as `@[c_extern]`. This makes V3 emit an explicit prototype in portable generated C without adding `<stdio.h>`, preventing strict C99 bootstrap compilers from rejecting the call as an implicit declaration.

## How was it tested?

- `./vnew -silent cmd/v/toolcache_test.v`
- Generated a fresh portable compiler snapshot and confirmed it contains an explicit `renameat` prototype
- Compiled that snapshot on Linux with `cc -std=c99 -Werror=implicit-function-declaration`
- `./vnew -silent vlib/v/compiler_errors_test.v` (currently fails on `master`: expected-error fixtures gain the V1 retry banner and notes)
- `./vnew -silent test vlib/v/` (same current `master` fixture-output failure)


medvednikov:
@codex review

PR #28634, opened 2026-09-15
cgen: restore Vinix kernel builds with V3

medvednikov:
## What does this PR do?

Restores Vinix kernel compilation with the new C backend.

The Vinix build exercises several backend paths that were still incomplete in V3. This change:

- implements `-no-closures` and makes selector/import classification file-aware, avoiding false closure lowering for imported functions and constants;
- preserves explicit casts, pointer declaration types, interface-field lvalues, generic-method callees, and call-argument evaluation order through transforms;
- restores the Vinix entry-point ABI, `free` mapping, target-libc header handling, and `-nofloat` runtime pruning;
- lowers x86 port-I/O operands and AArch64 register/address operands correctly; and
- adds focused regression fixtures for the affected driver, transform, C generation, and inline-assembly paths.

This is rebased on #28628. A two-stage V3 self-build now completes without `-no-memory-limit` and peaked at about 1.73 GB on macOS.

## Tests

- `vlib/v/driver/implicit_import_test.v`
- `vlib/v/gen/c/preamble_test.v`
- `vlib/v/gen/c/stmt_test.v`
- `vlib/v/compiler_tests/vinix_selector_codegen_test.v`
- `vlib/v/tests/interfaces/interface_nested_field_postfix_test.v`
- `vlib/v/compiler_tests/inline_asm_c_lowering_test.v` (passed before the #28628 rebase; after rebasing, its internal compiler process for the existing asm-goto lock-scope fixture still exceeds the default 10 GB child-process cap)
- `vlib/v/gen/c/`: 315 passed; the existing V3 `target_test.v` cache-input assertion remains the sole failure
- Vinix x86_64: generated kernel C compiles to `obj/blob.c.o`
- Vinix AArch64: generated kernel C compiles to `obj/blob.c.o`

The broad compiler-error golden suite is not yet a useful V3 signal because the backend baseline still omits many classic checker/parser diagnostics, as also documented in #28628.


medvednikov:
@codex review

medvednikov:
Ignoring CI as requested, I found one codegen issue in the x86 port-I/O lowering.

`lower_c_inline_asm_template_with_wide_aliases` currently rewrites **every** wide alias in any `in`/`out` variant to `%k[...]`. `%k` forces a 32-bit register name, which is only appropriate for the 32-bit accumulator case. For the sized variants, `inb`/`outb` need the byte accumulator (`%b[...]` / `%al`) and `inw`/`outw` need the 16-bit accumulator (`%w[...]` / `%ax`). Likewise, if the port operand itself is wide and uses an `Nd` constraint, the register form must still be `%dx` (`%w[...]`), not `%edx`.

For example, a wide `result` used with `inb result, port` would currently be rewritten to `%k[result]`, producing an invalid `inb ... %eax` form. The same blanket rewrite can also turn a wide port operand into `%edx`.

I think the modifier needs to depend on both the opcode width and whether the alias is the data/accumulator operand or the port operand. It would be worth adding coverage for `inb`/`inw`/`outb`/`outw` plus a wide `Nd` port; the new test only covers unsuffixed `in` with a `u16` port, so it doesn't exercise these cases.

PR #28635, opened 2026-09-15
parser: support comma-less array literals

medvednikov:
## Summary

- accept whitespace-separated elements in ordinary array literals
- preserve legacy prefix/infix operator boundaries in comma-less literals
- keep doubled commas invalid and cover the behavior with a regression test

## Testing

- VTEST_ONLY_FN=test_array_literal_separator_handling ./vnew -silent vlib/v/compiler_tests/review_transform_regressions_test.v
- ./vnew -silent test vlib/v/parser/ (53 passed)
- focused vfmt comparison for array_init_without_commas

The broader compiler_errors_test.v and vlib/v/ runs were attempted, but unchanged master currently fails their expected-error fixtures on macOS because the launcher adds compatibility-fallback notes. The same failure reproduces with the unmodified master bootstrap compiler.


medvednikov:
@codex review

medvednikov:
I think this introduces a parser-scope regression for nested expressions inside an array literal.

`p.in_array_literal > 0` is true for the entire time the literal is being parsed, including recursive `p.expr(...)` calls inside parentheses, call arguments, indexing, etc. The new check in `expr_with_lhs_context` therefore treats an attached `-`/`*`/`&` as an array-element boundary even when it is not at the top level of the array element.

For example, this should remain a single parenthesized subtraction expression:

```v
a := [(3 -2)]
assert a == [1]
```

While parsing the inner `3 -2`, `p.in_array_literal > 0`, there is whitespace before `-`, and `-` is attached to `2`, so the new branch breaks the inner expression at `-`. The parenthesized-expression parser then sees `-2` where it expects `)`.

The same issue should affect shapes like `[foo(3 -2)]` or `[xs[i -1]]`.

Could the comma-less boundary detection be scoped specifically to the top-level `p.expr(.lowest)` used for an array element (rather than every nested expression while an array literal is open)? A regression test with a parenthesized nested expression such as the above would cover this.

einar-hjortdal:
Do we really want this feature? Shouldn't there be only one way or writing array literals?

squidink7:
I'm happy either way. Not requiring commas puts it in line with function calls and struct definitions, both of which don't need commas however if its going to introduce ambiguous behaviour it might be best to leave out.

PR #28637, opened 2026-09-15
crypto.rand: keep cross bootstrap portable on macOS

medvednikov:
The Linux-generated portable compiler snapshot started including crypto.rand through the tool cache. Its Linux entropy implementation called SYS_getrandom unconditionally, so Apple clang could no longer compile vc/v.c.

Keep both entropy paths in cross output: Apple hosts use getentropy, while Linux keeps the existing syscall. Add a cross-output regression test that verifies the platform guard and fallback.

Validation:
- regenerated a Linux-target cross vc/v.c and compiled it with Apple clang
- make local=1 completed v1 -> v2 -> v
- ./v -silent test vlib/crypto/rand/
- focused cross-output regression test


medvednikov:
@codex review

PR #28643, opened 2026-09-15
v3: fix Office compilation regressions

medvednikov:
## Summary

- keep scoped monomorphization metadata alive after worker arenas are released
- preserve multi-return, params-struct, `or`-exit, and array-literal types through V3 lowering
- resolve generic namespace callees and module-local optional return types correctly
- keep resolved ordinary functions such as `json__decode` out of legacy JSON matching
- match established fixed-point float interpolation rounding

Office now uses the pure-V `json2` module only; this PR does not add legacy JSON compiler machinery. It is based on the merged #28628 RAM fix and does not duplicate it.

## Validation

- strict V3 builds: Office Word, Excel, and server
- Office Word: 1201/1201 tests
- Office Excel: 579/579 tests
- Office server: 22/22 tests after migration to `json2`
- all seven focused V3 regression targets pass, including a local ordinary `json__decode` call with no JSON import
- `./vnew -silent vlib/v/compiler_errors_test.v`: known V3 baseline remains (954 passed, 767 failed, 1 skipped; failures are existing diagnostic/parity gaps)
- `./vnew -silent test vlib/v/gen/c/`: 313/314 pass; existing `target_test.v` cache-input assertion remains


medvednikov:
@codex review

medvednikov:
One potential regression in the JSON decode matching:

`is_json_decode_target_name()` now treats the bare target `json__decode` as legacy JSON compiler magic in every module. `is_json_decode_call()` first checks `resolved_call_name(id)`, but if that resolves to a non-JSON function it still falls back to the textual target, so a normal/local function named `json__decode` can be misclassified. Double-underscore function names are valid V (there are many `__...` functions in `builtin`).

This is especially risky because `collect_json_decode_optional_types()` scans all calls without a `has_legacy_json_module()` guard and will then derive/synthesize optional types from that ordinary call; other magic-decode call sites can also observe the same match.

Could we either gate the `json__decode` spelling on actual legacy-`json` resolution/import context, or avoid falling back to `target` when a non-empty resolved call name is present? A regression test with a local `fn json__decode(...)` and no `json` import would cover this.

PR #28644, opened 2026-09-15
tools: fix VLS source fallback

medvednikov:
## Summary

- compile the current VLS source tree directly instead of invoking the removed `build.vsh`
- rebuild an up-to-date checkout when its executable is missing, including the Windows executable suffix
- return success when the missing-release-asset path successfully falls back to a source build
- add a regression test for recovering an incomplete source installation

## Cause

The latest VLS repository is built with `v .` and no longer contains `build.vsh`. Its latest release also has no platform VLS executable, so `v ls --update` must use the source fallback. The updater still called the removed script, skipped later rebuilds once the checkout was current, and rethrew the original missing-asset error after a successful fallback.

## Testing

- `./vnew -silent test cmd/tools/vls_test.v`
- live source recovery against the current `vlang/vls` repository in an isolated temporary home
- live default release-to-source fallback against the current GitHub release in an isolated temporary home


medvednikov:
@codex review

PR #28646, opened 2026-09-15
cgen: release header scan scratch memory

medvednikov:
## What does this PR do?

Parse each uncached C header in a short-lived preallocation scope, promote only the cached directives and include-guard metadata, then release the temporary header text, logical lines, and preprocessing strings immediately. Non-preallocated compiler builds retain the existing allocation behavior.

The directive cache merged in #28628 fixes the repeated-header scan that made the reported Vinix build cross the 10 GiB compiler limit. This follow-up also bounds scratch memory for the first scan of every unique header.

On macOS, an isolated cold-cache `vinix-website` build with an explicitly preallocated compiler changed as follows:

| Metric | master | This PR |
| --- | ---: | ---: |
| Peak RSS | 1,682,751,488 bytes | 1,618,329,600 bytes |
| Peak physical footprint | 1,182,812,656 bytes | 1,117,096,336 bytes |

That is about 63 MiB / 5.6% less peak physical footprint. The original pre-#28628 compiler reproduced the reported failure at 10,178 MiB.

## Tests

- `./vnew -silent test vlib/v/gen/c/source_directive_test.v` — 50 passed
- `./vnew -prealloc -silent test vlib/v/gen/c/source_directive_test.v` — 50 passed
- isolated cold-cache Vinix website build — passed
- `./vnew -silent test vlib/v/gen/c/` — 313 passed; the existing `target_test.v` ambiguous-include assertion failed, matching the clean-base failure documented in #28628
- `./vnew -silent vlib/v/compiler_errors_test.v` and `./vnew -silent test vlib/v/` encounter the current master fallback-diagnostic output mismatches also documented in #28628


medvednikov:
@codex review

PR #28649, opened 2026-09-15
tools: compile all external tools with V3

medvednikov:
## Summary
- build and run every `cmd/tools` program with the current V3 compiler
- port `vast`, `vdoc`, and `vvet` from the removed V1 AST/checker APIs to `v.flat` and the current parser/scanner APIs
- fix the remaining V3 incompatibilities in `vbin2v`, `vcomplete`, `vgret`/TOML, `vscan`, `vtest-parser`, and `vwipe-cache`
- prefer an executable `cmd/tools/<name>.v` over a same-named module directory, fixing `v share`
- report cached or fresh tool build failures directly instead of automatically launching V1
- preserve compatibility fallback for normal compiler inputs and explicit `-old-compiler` behavior

## Why
External tools belong to the current source tree and should compile with its current compiler and vlib. Retrying them through the V1 release tree can run stale sources, hide the original V3 error, and produce unrelated failures such as the VLS `json2` error.

## Validation
- fresh no-cache/no-fallback V3 compilation of all 51 distinct external-tool targets
- fresh no-fallback V3 compilation of upstream VLS `7a30ca152008ea5c14723c2a9d0cf4f2853efefc`
- `./vnew -silent test cmd/v/` (68 passed)
- all six vdoc test programs, including output fixtures and example checks
- `cmd/tools/vvet/vet_test.v`
- `cmd/tools/vcomplete_test.v`
- all 32 TOML test programs plus `vlib/toml/scanner/scanner_test.v`
- launcher smoke tests for `ls`, `ast`, `doc`, `scan`, `vet`, `share`, `bin2v`, and `test-parser`


medvednikov:
@codex review

medvednikov (commented):
There is a regression in the `vdoc` port around `document.generate_with_pos()`. The function still sets `with_pos`, `filename`, and `pos`, but the new flat-AST `generate()`/`parse_file()` path never consumes those fields or populates `scoped_contents`. Before this change, `file_asts()` detected the requested file and called `file_ast_with_pos()` to collect variables from the innermost scope at the requested offset. After this PR, callers of `generate_with_pos()` will always get an empty `scoped_contents`, so the position-aware API silently loses its main behavior.

Please restore equivalent scope extraction for the flat AST path, and ideally enable/add coverage for `generate_with_pos()` (the existing test stub is still commented out).

PR #28653, opened 2026-09-15
transform: preserve locked generic return types

medvednikov:
## Summary

- preserve `main.` type locks when validating specialized generic function return types
- prevent caller-owned types such as `main.Table` from being rebound to same-named types in the declaration module
- add a regression test for an ORM model named `Table` with a qualified table attribute

## Testing

- `./vnew fmt -verify vlib/v/tests/orm_table_type_name_collision_test.v`
- `./vnew -new-compiler vlib/v/tests/orm_table_type_name_collision_test.v`
- `./vnew vlib/v/tests/orm_table_type_name_collision_test.v`
- `./vnew -new-compiler vlib/v/tests/orm_table_name_test.v`
- `./vnew -new-compiler vlib/v/transform/monomorphize_test.v`
- `./vnew -new-compiler vlib/v/transform/fn_test.v`

## Broader checks

- `./vnew -new-compiler -silent test vlib/v/transform/`: 235 passed, 4 failed; the first failure reproduces on unchanged `master`
- `./vnew -silent test vlib/v/`: exits 1 without diagnostics in this checkout
- `./vnew -silent vlib/v/compiler_errors_test.v`: fails on unchanged `master` because compatibility-fallback annotations differ from the checked-in expected output


medvednikov:
@codex review

PR #28654, opened 2026-09-15
os: look up bare command names in PATH when spawning a process on Windows (fix #28516)

### LINKED ISSUE #28516: os: `new_process()` on Windows never searches PATH (`abs_path` makes `gcc` → `<cwd>\gcc`, CreateProcess fails) — breaks V3 `v -o v2.exe cmd/v` on tools-windows CI
### Describe the bug

`os.new_process(name)` on Windows never searches `PATH`. A bare program name is turned into `<cwd>\<name>` by `abs_path()` and handed to `CreateProcessW` as `lpApplicationName`, which does no `PATH` lookup — so any spawn of `gcc`, `cc`, `clang`, … by bare name fails:

```
failed CreateProcess: The system cannot find the file specified.
```

and `failed_cfn_report_error` then `exit(1)`s the whole process.

Since V3 became the default on Windows (#28479) this breaks **building V itself** on the Windows tools CI: `tools_ci.yml` → `tools-windows (gcc|msvc)` → step "Build V" → `./makev.bat -gcc && ./v -o v2.exe cmd/v` — makev.bat succeeds, then the freshly built V3 `v.exe` dies on `-o v2.exe cmd/v` with exactly that message. Every PR touching `cmd/tools/` is currently red there (e.g. #28498).

### Where

`vlib/os/process_windows.c.v`:

```v
// :67
p.filename = abs_path(p.filename) // expand the path to an absolute one, ...
// :155-157
mut application_name_ptr := &u16(unsafe { nil })
if is_abs_path(p.filename) && !filename_lc.ends_with('.bat') ... {
	// -> lpApplicationName = "<cwd>\gcc"
```

`abs_path('gcc')` is `<cwd>\gcc`, so the `is_abs_path` branch is always taken and `CreateProcessW` gets a path that does not exist.

### Reproduction

```v
import os

fn main() {
	println(os.find_abs_path_of_executable('gcc') or { 'NOT FOUND' }) // gcc IS on PATH
	println(os.abs_path('gcc'))
	mut p := os.new_process('gcc')
	p.set_args(['--version'])
	p.set_redirect_stdio()
	p.run()
	p.wait()
	println('exit: ${p.code}')
}
```

```
S:\Tools\mingw64\16.1\bin\gcc.exe
S:\repo\vlang\gcc
failed CreateProcess: The system cannot find the file specified.
```

### Why V3 hits it and V1 never did

- V1's builder runs the C compiler through `os.execute(cmd)` → `cmd.exe` → `PATH` search. Works.
- V3 runs it through `cmdexec.run(program, args)` → `os.new_process(program)` (`vlib/v3/cmdexec/cmdexec.v:18`) with the bare compiler name. Fails.

It is masked whenever `thirdparty/tcc/tcc.exe` exists: `v3_select_windows_default_c_compiler` (`vlib/v3/driver/driver.v:9136`) then returns that **full path**, which is why a normal local checkout works. CI with `-gcc` ("tcc was not used") and any fresh checkout / worktree without `thirdparty/tcc/` gets the bare name.

### Suggested fix

In `new_process`/`run` on Windows, when `filename` contains no path separator, resolve it with `os.find_abs_path_of_executable()` before the `abs_path` expansion (fall back to the current behaviour if not found). Alternatively pass `lpApplicationName = NULL` and put the program at the front of `lpCommandLine`, which makes `CreateProcessW` do its own `PATH` search — but that changes `.bat`/`.cmd` handling, so the explicit lookup is the smaller change.

### Environment

```
V 0.5.2 316db75, Windows 11, mingw-w64 gcc 16.1 on PATH. CI: tools_ci.yml tools-windows (gcc) and (msvc), every run since 5ff8ecb55f.
```

> [!NOTE]
>

quaesitor-scientiam:
Fixes #28516.

### Problem

`Process.win_spawn_process` (`vlib/os/process_windows.c.v`) turned every filename into `abs_path(filename)`, so a bare program name such as `gcc`, `cl` or `node` became `<cwd>\gcc`. `CreateProcessW`, when given an application name, does no PATH lookup, so every such spawn failed with

```
failed CreateProcess: The system cannot find the file specified.
```

and — because that path `exit(1)`s the parent — took the whole program down with it, without even naming the program. The V3 driver mostly avoided this by pre-resolving the main C compiler, but every other bare-name spawn hit it (`pkg-config`, `node`, `cl` during `-cc msvc` cross generation, the old cgen include probe, and any user code calling `os.new_process('gcc')`). The `unix` backend already resolves names via PATH (`unix_resolve_filename`); Windows never did — the existing test for it was explicitly gated off on Windows.

### Change

- New `win_resolve_filename()`, mirroring `unix_resolve_filename()`: absolute paths and paths containing a directory separator keep the old behaviour (anchored to the parent's cwd); a bare name is looked up in the [documented `CreateProcessW` order](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) — the application directory, the parent's current directory, the 32-bit system directory, `<windir>\System`, the Windows directory (`win_bare_command_search_dirs()`) — and only then in the `PATH` from the child's environment, each via the existing `find_abs_path_of_executable_in_path_env` with the `.com`/`.exe`/`.bat`/`.cmd` suffixes. Every directory is searched on its own, deliberately: the helper iterates suffixes outside directories, so a combined search would let a `tool.exe` on PATH shadow a `tool.cmd` in the cwd, or a same-named `.exe` on PATH shadow the one in System32 (both review findings, each fixed with a regression test).
- `PATH` is matched case-insensitively (`win_env_value_from_entries`), since Windows spells it `Path`.
- If nothing is found the bare name is left untouched, so `CreateProcessW`'s own lookup still runs before the spawn is reported as failed (the report names the program since #28660).
- `test_new_process_uses_path_for_bare_command_names` is un-gated on Windows and additionally spawns the extension-less spelling (`process_from_path`), which is what Windows callers normally use.
- The suffixes for a bare name are now tried in the default `PATHEXT` order, `.com`, `.exe`, `.bat`, `.cmd` (`executable_suffixes`): `.com` was missing entirely, and `.exe` came before it, unlike cmd.exe. Test: `test_new_process_tries_the_suffixes_in_pathext_order_on_windows` (`pathext_tool.com` beats `pathext_tool.exe` in the same directory, for both `find_abs_path_of_executable` and a spawn).
- `test_new_process_prefers_the_current_directory_over_path_on_windows`: cwd `tool.cmd` beats PATH `tool.exe`.
- `test_new_process_prefers_the_system_directory_over_path_on_windows`: an impostor `where.exe` in the first PATH entry must lose to `%SystemRoot%\System32\where.exe` (checked via `where /q <missing>` exiting 1 vs the impostor's 0).
- A name whose extension is already `.exe`/`.com`/`.bat`/`.cmd` is looked up exactly and never expanded (`executable_suffixes_for`), in every directory, as `CreateProcessW` does — so a `cmd.exe.exe` anywhere in the search order cannot shadow the requested `cmd.exe`. A name with some other extension (`python3.12`) keeps the usual suffix expansion, so it still resolves to `python3.12.exe`. This is in the shared helper, so `os.find_abs_path_of_executable` gets the same rule — which also fixes a pre-existing bug there: on master, `os.find_abs_path_of_executable('C:\dir\tool.exe')` returns `C:\dir\tool.exe.exe` (the absolute-path shortcut returned the first suffix unchecked); it now returns the path as given (review findings, rounds three and four). Tests: `test_new_process_runs_the_exact_spelling_of_a_suffixed_name_on_windows` (both files in one PATH directory) and `test_new_process_does_not_expand_a_suffixed_name_in_an_earlier_directory_on_windows` (`exact_tool.exe.exe` in the cwd, `exact_tool.exe` on PATH).

The `.bat`/`.cmd` application-name exclusion at the `CreateProcessW` call is unchanged, so a resolved `foo.bat` still goes through the command line as before (verified: a bare `hellobat` resolving to `…\hellobat.bat` runs, both from PATH and from cwd).

### Verification (Windows 11, bundled tcc)

- The un-gated test hard-exits with the old `failed CreateProcess` message on the unpatched tree and passes with the fix.
- `vlib/os/process_test.v` fully green (all 23 tests, including the five ordering tests; each ordering test fails against the commit before its fix). `vlib/os/find_abs_path_of_executable_test.v` green.
- `vlib/os/*_test.v` run file-by-file: the same 3 files fail with and without this change (`filepath_test`, `os_stat_windows_test`, `process_args_test`) — pre-existing on this machine, unrelated.
- Probes: bare `.bat` from PATH, bare `.bat` from cwd, and a genuinely missing name (message now names it).

Not changed here: on a spawn failure `Process` still `exit(1)`s the parent, whereas `os.execute` returns the error in its `Result` — a pre-existing divergence worth its own issue.


medvednikov (commented):
Reviewed the code changes only (ignoring CI as requested). The main PATH fix looks directionally right, but I think there is one resolver-ordering bug worth addressing:

`find_abs_path_of_executable_in_path_env()` iterates `executable_suffixes` *outside* the PATH-directory loop. Because `win_resolve_filename()` calls it with `getwd() + path_delimiter + path`, prepending the cwd does **not** guarantee the stated “current directory first, then PATH” behavior across suffixes. For example, if the cwd contains `tool.cmd` but an earlier PATH entry contains `tool.exe`, the helper will find the PATH `tool.exe` while processing the `.exe` suffix before it ever considers the cwd `tool.cmd`. That can change which program gets executed. Resolving against the cwd first and only then resolving against the child PATH (or using a directory-first search for Windows here) would preserve the intended ordering. A regression test with different suffixes in cwd vs PATH would catch this.

Also, the fallback comment says `CreateProcessW` “knows the PATHEXT suffixes”. Microsoft’s `CreateProcessW` docs say that with `lpApplicationName == NULL`, a missing extension gets `.exe` appended; it does not describe PATHEXT expansion, and batch files are documented as requiring the command interpreter. So I’d avoid relying on PATHEXT in the fallback comment/logic: https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

Aside from that, the case-insensitive `PATH` lookup and un-gating the bare-name regression test make sense for the reported bug.

quaesitor-scientiam:
Thanks — both points were right, and the ordering one reproduced exactly as described: with `preferred_tool.cmd` in the cwd and `preferred_tool.exe` on PATH, the old code ran the `.exe` (`got 7`, expected 3).

Fixed in 46bad2481d:
- `win_resolve_filename` now resolves against the current directory alone first, and only then against the child's PATH — two separate `find_abs_path_of_executable_in_path_env` calls, so the helper's suffix-outer loop can no longer let a PATH `.exe` shadow a cwd `.cmd`. The doc comment explains why the lookups are separate.
- New regression test `test_new_process_prefers_the_current_directory_over_path_on_windows`: `.cmd` in cwd (exit 3) vs `.exe` on PATH (exit 7), spawned by bare name; fails on the previous commit, passes now.
- Fallback comment corrected: `CreateProcessW` appends `.exe` and searches the system directories; no PATHEXT claim.

`vlib/os/process_test.v` is fully green on Windows with the change (27 s, 19 tests).

medvednikov (commented):
Re-reviewed the updated commit, ignoring CI. The two earlier issues are fixed: the cwd/PATH suffix-order bug is now covered by a good regression test, and the PATHEXT wording is corrected.

One follow-up correctness issue remains in the search order. `win_resolve_filename()` now explicitly resolves a bare name as **cwd -> child PATH** before `CreateProcessW` gets a chance to search. But Microsoft documents the native `CreateProcessW(NULL, command_line, ...)` order for a bare module as roughly **application directory -> parent cwd -> system directories -> Windows directory -> PATH**. So if (for example) the child PATH contains a same-named `cmd.exe`, this resolver can select that PATH copy before `%SystemRoot%\System32\cmd.exe`, while native `CreateProcessW` would select the system copy first. That is an observable behavior change and could execute a different program than Windows' normal lookup.

Docs: https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw

I think the resolver should either preserve those pre-PATH Windows search locations before consulting PATH, or otherwise delegate bare executable lookup in a way that keeps the documented CreateProcess search order. A regression test with a same-named executable on PATH versus a known system executable would make this explicit.

quaesitor-scientiam:
Agreed, and thanks for the pointer. Fixed in e41e03d487 (branch also rebased onto current master; the only conflict was the error label from #28660, upstream's wording kept):

- New `win_bare_command_search_dirs()` returns the pre-PATH locations in the documented order: the directory the application loaded from (`executable()`), the parent's current directory, `GetSystemDirectoryW`, `<windir>\System` (the 16-bit system dir, which has no query function), `GetWindowsDirectoryW`. `win_resolve_filename()` searches each of those on its own, then the child's PATH. Per-directory searching keeps the suffix loop from letting a PATH `.exe` shadow a system-directory `.exe` (same mechanism as the cwd `.cmd` case).
- Regression test `test_new_process_prefers_the_system_directory_over_path_on_windows`: an impostor `where.exe` (copy of `test_os_process`, exits 0) in the first PATH entry, spawned as bare `where /q <missing name>`; the real `System32\where.exe` exits 1. Fails on the previous commit (impostor runs, `got 0`), passes now.

One deliberate remaining difference from native: `CreateProcessW` only ever appends `.exe`, while the resolver also tries `.bat`/`.cmd` in each directory (that is what makes the cwd `tool.cmd` case work). If you'd rather have those two suffixes limited to cwd/PATH and keep the system directories `.exe`-only, that is a one-line change - say the word.

`vlib/os/process_test.v` green on Windows (20 tests).

JalonSolov:
It should act exactly as Windows does, so trying all 3 (in whatever order Windows does) is the correct way.

JalonSolov:
In fact, there should be a Windows API call to do the search for you, as Windows has to do it all the time.

According to Google AI Overview:

"Use the Windows API to locate an executable as the OS does by probing the module's path through GetModuleFileName or retrieving the full path via SearchPath, then Normalize with ResolvePath; for launching, use CreateProcess with a fully qualified path."

quaesitor-scientiam:
Thanks — the suffixes stay as they are then.

On `SearchPathW`: it's the right primitive in spirit and it is what `win_bare_command_search_dirs()` reproduces, but it isn't a drop-in here for two reasons. It takes one extension per call, so `.exe` → `.bat` → `.cmd` would be three full-order passes and bring back the shadowing that was fixed in round 1 (a `tool.exe` anywhere on PATH beating a `tool.cmd` in the cwd). And it searches the calling process's PATH, whereas this resolver, like `unix_resolve_filename`, uses the PATH the child will start with (`p.env`), so `set_environment` keeps working for lookup. `GetModuleFileName` is already what `os.executable()` uses for the application directory, and the resolved path is fully qualified before `CreateProcessW` is called. (There's no Win32 `ResolvePath` — that part of the overview is invented.)

Happy to switch to `SearchPathW` if you and @medvednikov prefer the OS implementation over exact suffix/child-PATH semantics — just say so.

JalonSolov:
Personally, I don't really care.

I think it should act as much like Windows does as possible, so nobody is "surprised" by V's code calling something that wasn't expected.  That's why I was thinking that Windows should expose some mechanism to do the lookup for you.

However, since I hope to never have to use Windows again, it doesn't matter that much to me what the final implementation looks like.

quaesitor-scientiam:
> Personally, I don't really care.
> 
> I think it should act as much like Windows does as possible, so nobody is "surprised" by V's code calling something that wasn't expected. That's why I was thinking that Windows should expose some mechanism to do the lookup for you.
> 
> However, since I hope to never have to use Windows again, it doesn't matter that much to me what the final implementation looks like.

Not everything in Windows is perfect. Or for that matter any OS.  Rechecked timings found one improvement that will be a separate PR.

medvednikov (commented):
Re-reviewed the latest commit, ignoring CI. The pre-PATH Windows search order is fixed now, and the `where.exe` regression test directly covers the case I raised.

I found one remaining filename-resolution edge that becomes observable through this change: `find_abs_path_of_executable_in_path_env()` appends every Windows `executable_suffixes` entry before it reaches the empty suffix. That means an *already suffixed* bare filename is not searched exactly first. For example, `os.new_process('cmd.exe')` will probe `cmd.exe.exe`, `cmd.exe.bat`, and `cmd.exe.cmd` before `cmd.exe`. With the new `win_resolve_filename()` calling this helper for all bare names, a `cmd.exe.exe` in an earlier searched directory can shadow the explicitly requested `cmd.exe`; native `CreateProcessW` would not add another extension to a name that already has one.

I think suffix expansion should only apply to extension-less names (or, at minimum, an explicitly suffixed name should be tried exactly before any expansion). A regression test could put `process_from_path.exe` and `process_from_path.exe.exe` in the same searched directory with different exit codes and verify that `new_process('process_from_path.exe')` runs the exact spelling.

quaesitor-scientiam:
Good catch — reproduced exactly: with `exact_tool.exe` and `exact_tool.exe.exe` in the same PATH directory, `new_process('exact_tool.exe')` ran the `.exe.exe` (`got 0`, expected 3).

Fixed in da18364a73, in the shared helper so `os.find_abs_path_of_executable('foo.exe')` is corrected too:

- New `executable_suffixes_for(exe_name)`: if the name already has an extension (`file_ext != ''`), the empty suffix is tried first, then the rest; an extension-less name keeps the existing order. I went with "exact first, then expand" rather than "no expansion at all" so that `python3.12` still resolves to `python3.12.exe`, which is what cmd.exe/PATHEXT users expect from a bare name. On non-Windows the suffix list is just `['']`, so nothing changes there.
- Regression test `test_new_process_runs_the_exact_spelling_of_a_suffixed_name_on_windows`, as you described: `exact_tool.exe` (copy of `test_os_process`, exits with the given 3) and `exact_tool.exe.exe` (ignores its arguments, exits 0) in one PATH directory. Fails on the previous commit, passes now.

`vlib/os/process_test.v` (21 tests) and `vlib/os/find_abs_path_of_executable_test.v` green on Windows.

medvednikov (commented):
Re-reviewed `da18364a73`, ignoring CI. The new `executable_suffixes_for()` fixes the same-directory case from my previous comment, but I think the earlier-directory case is still present.

`win_resolve_filename()` searches each pre-PATH directory separately. For a requested `cmd.exe`, `executable_suffixes_for('cmd.exe')` now tries `cmd.exe` first *within each directory*, but then still tries `cmd.exe.exe`/`.bat`/`.cmd` before moving to the next directory. So, for example, if the application directory (or cwd) contains `cmd.exe.exe` while System32 contains the real `cmd.exe`, the resolver can still select the earlier `cmd.exe.exe`. That is the exact-spelling shadowing case I meant by “a `cmd.exe.exe` in an earlier searched directory can shadow the explicitly requested `cmd.exe`”; the new regression test only puts both files in the same PATH directory, where exact-first is sufficient.

For recognized explicit suffixes (`.exe`, `.bat`, `.cmd`), I think lookup should be exact-only across the directory search order. If preserving expansion for dotted names such as `python3.12` is desired, that can be limited to extensions that are not already one of the executable suffixes. A regression test with `exact_tool.exe.exe` in an earlier pre-PATH directory and `exact_tool.exe` in a later one would catch this.

quaesitor-scientiam:
You're right, I under-read the earlier-directory part. Reproduced: `exact_tool.exe.exe` in the cwd and `exact_tool.exe` on PATH, `new_process('exact_tool.exe')` ran the cwd `.exe.exe` (`got 0`, expected 3).

Fixed in 9c716365ec, following your suggestion:

- `executable_suffixes_for`: a name whose extension is one of `.exe`/`.com`/`.bat`/`.cmd` (case-insensitive) is looked up exactly and nothing else, in every directory. Only a name with some other extension (`python3.12`) keeps "exact first, then the suffixes". `.com` is included because `is_executable` accepts it and the CreateProcessW docs single it out as an extension that must be given explicitly.
- Regression test `test_new_process_does_not_expand_a_suffixed_name_in_an_earlier_directory_on_windows`, as described: shadow in the current directory (searched before PATH), requested spelling on PATH. Fails on da18364a73, passes now. The same-directory test from the previous round stays.

`vlib/os/process_test.v` (22 tests) and `vlib/os/find_abs_path_of_executable_test.v` green on Windows.

JalonSolov:
When extensions need to be added, they should be done in PATHEXT (a common Windows env var) order.

By default, it is usually
```
PATHEXT=.COM;.EXE;.BAT;.CMD
```
However, different applications can extend this list.

quaesitor-scientiam:
One more small commit, f0dfc41530, from a full pass over the branch rather than a reaction to a finding:

- `executable_suffixes_for` had a dead branch: trying `python3.12` exactly before the suffixes could never match, because `is_executable` rejects every extension other than `.exe/.com/.bat/.cmd` on Windows. It is now two cases only — exact-only for those four extensions, the normal suffix list for everything else. Behavior is unchanged; the comment no longer claims something the code does not do.
- The spawn-site comment still described the round-one lookup (cwd, then PATH); refreshed.
- Noted in the PR description: the exact-only rule also fixes `os.find_abs_path_of_executable('C:\dir\tool.exe')`, which on master returns `C:\dir\tool.exe.exe` (the absolute-path shortcut returned the first suffix unchecked).

Remaining known deviations from native, both pre-existing and left alone: `executable_suffixes` has no `.com`, so a bare `foo` never resolves to `foo.com` (native `CreateProcessW` doesn't either); and a name with a non-executable extension (`script.ps1`) is still expanded (`script.ps1.exe`, ...), which native would not do but cmd.exe/PATHEXT does. Happy to change either if you want strict native behavior.

`vlib/os/process_test.v` and `vlib/os/find_abs_path_of_executable_test.v` green on Windows.

medvednikov (approved):
Re-reviewed the latest update, ignoring CI as requested. The cross-directory exact-name shadowing issue is fixed: recognized executable extensions (`.exe`, `.com`, `.bat`, `.cmd`) are now exact-only, so an earlier `exact_tool.exe.exe` can no longer shadow a later explicitly requested `exact_tool.exe`, and the new cwd-vs-PATH regression test covers that case directly.

I also checked the shared-helper impact and the follow-up cleanup in `f0dfc41530`; I don't see another substantive code issue in this pass. Looks good from code review.

quaesitor-scientiam:
Done in 14b47e83f5: `executable_suffixes` is now `.com`, `.exe`, `.bat`, `.cmd` - the default PATHEXT order. `.com` was missing altogether (so a bare `tool` never found `tool.com`, although `is_executable` accepts it), and `.exe` came first. Regression test `test_new_process_tries_the_suffixes_in_pathext_order_on_windows`: `pathext_tool.com` (PE image renamed, exits with the given 3) and `pathext_tool.exe` (exits 0) in one PATH directory; asserts both `find_abs_path_of_executable` and the spawn pick the `.com`. Fails on the previous commit, passes now.

Reading `%PATHEXT%` at runtime (for the extended lists applications add) is left for a follow-up: every added extension also needs `is_executable` and the batch-file routing to know about it.

`vlib/os/process_test.v` (23 tests) and `vlib/os/find_abs_path_of_executable_test.v` green on Windows.

quaesitor-scientiam:
Ci’s are complete.  Ready to merge

Issue #28620, opened 2026-09-14
v3: every successful build leaves its '.<out>.v3cc.<pid>.*' C build directory behind on Windows (success path uses os.rmdir on a non-empty dir)


quaesitor-scientiam:
**Describe the bug**

On Windows, every successful V3 build leaves its per-output C build directory behind, next to the output binary:

```
S:\x> v -o hello.exe hello.v
S:\x> dir /a
.hello.exe.v3cc.54012.348737100.271124032   <DIR>
hello.exe
hello.v
S:\x> dir /a .hello.exe.v3cc.54012.348737100.271124032
out.def          55 bytes
```

Each build creates a new directory (the name embeds pid + timestamps), so they accumulate: after a day of `v up` / `v self` / running one script I had 28 of them in the V root (`.v_up.exe.v3cc.*`, `.v2.exe.v3cc.*`), 9 next to the script, 16 in `~/.cache/v/tools/` (those from tool builds still contained the full `src.c` plus dozens of `dependency_scan_fallback_*.o`, ~2.6 MB each), and ~100 in a scratch directory used for repeated hello-world builds.

**Root cause**

The success path of the C build in `vlib/v/driver/driver.v` (around line 12613 at `ff57e2dc2c`) removes a fixed list of files it knows it created and then calls plain `os.rmdir`:

```v
os.mv(cc_out, bin_file) or { ... }
for temporary_object in c_object_cache_stats.temporary_objects {
    os.rm(temporary_object) or {}
}
...
os.rm(tcc_main_file) or {}
os.rm(cache_full_tcc_source) or {}
os.rm(retained_full_c_source) or {}
os.rm(cc_src) or {}
os.rmdir(cc_dir) or {}
```

Anything the C compiler itself drops into `cc_dir` is not on that list. Bundled TCC on Windows writes `out.def` next to `out` when linking an executable, and larger builds leave `dependency_scan_fallback_<pid>.*.o` objects, so `os.rmdir` fails on a non-empty directory and the error is swallowed. The directory is perfectly deletable afterwards (`rmdir /s` works), so it is not a lock problem.

The failure paths all go through `cleanup_c_build_dir`, which uses `os.rmdir_all` and does clean up correctly — only the success path is affected.

**Reproduction Steps**

```
echo fn main() { println('hi') } > hello.v
v -o hello.exe hello.v
dir /a
```

[...]

PR #28658, opened 2026-09-16
v3: accept the .exe that gcc and clang write for the -o out link target on Windows (fix #28625)

### LINKED ISSUE #28625: v3: gcc/clang builds on Windows fail with 'failed to finalize ...: Source path doesn't exist' (compiler writes out.exe, driver expects out) - breaks every -prod build incl. 'v -prod self'
**Describe the bug**

On Windows, every V3 build that goes through gcc or clang fails after a *successful* C compilation with:

```
failed to finalize <bin_file>: Source path doesn't exist
```

That covers `-cc <path-to-gcc>`, `-cc <path-to-clang>`, and — most visibly — every `-prod` build, because `-prod` moves the implicit compiler from the bundled TCC to gcc. So `v -prod self`, `v -prod -parallel-cc self` and `v up -prod` are all broken:

```
PS S:\> time {v -prod self}
failed to finalize C:\Users\john3\.cache\v\tools\vself-920af1e4...exe.staged.63096.exe: Source path doesn't exist
the V compiler cannot build `cmd/tools/vself` (recorded in `...vself-920af1e4...exe.unbuildable`); retrying with `C:\Users\john3\.cache\v\v1-fallback\0.5.2\v.exe`.
...
```

(the rest is the 0.5.2 fallback failing on its own; the first line is the actual problem.)

Minimal reproduction with the C compiler command shown:

```
PS S:\> v -prod -showcc -o hello.exe hello.v
warning: regenerating the tcc-targeted unit with S:\Tools\mingw64\16.1\bin\gcc.exe
  > S:\Tools\mingw64\16.1\bin\gcc.exe -std=gnu11 -O3 -flto -municode -Wl,--stack=33554432 -w -fwrapv -Wno-int-conversion -o out src.c -l dbghelp
failed to finalize S:\...\hello.exe: Source path doesn't exist
```

**Root cause**

The driver hard-codes the intermediate output name `out` inside the per-build directory:

- `vlib/v/driver/driver.v:11628` — `cc_out = os.join_path_single(cc_dir, 'out')`
- `:8398` — `cc_args << ['-o', 'out', 'src.c']`
- `:12387` / `:12481` — `tcc_args << ['-o', 'out', ...]`
- the non-TCC link path — `c_flag_plan.compiler_args('out', compiler_inputs, [])`

and then `os.mv(cc_out, bin_file)` at `:12613`.

MinGW gcc (and llvm-mingw clang) append `.exe` to an extension-less `-o` target, so the compiler writes `out.exe`; only TCC writes `out` literally. Verified directly:

```
> gcc -o out t.c    ->  out.exe
> tcc -o out2 t.c   ->  out2
```

`os.mv(...\out, bin_file)` therefore fails, the error is reported as "Source path doesn't exist", and the build directory is cleaned up so there is nothing left to inspect. `-parallel-cc` changes only the per-unit compile step; the final link still uses `-o out`, so it fails the same way.

**Reproduction Steps**

```
echo fn main() { println('hi') } > hello.v
v -prod -o hello.exe hello.v                                   # fails
v -cc S:\Tools\mingw64\16.1\bin\gcc.exe -o hello.exe hello.v   # fails
v -cc S:\Tools\llvm-mingw\bin\clang.exe -o hello.exe hello.v   # fails
v -prod -cc S:\repo\vlang\thirdparty\tcc\tcc.exe -o hello.exe hello.v   # works (TCC keeps the name)
v -o hello.exe hello.v                                          # works (implicit TCC)
```

**Expected Behavior**

gcc/clang builds (and therefore `-prod`) produce the binary on Windows.

**Suggested fix**

Use `out.exe` as the intermediate name on Windows in all of the places above (TCC honours an explicit `.exe` name, gcc/clang keep it), or, after the compile, accept `cc_out + '.exe'` when `cc_out` is absent. The same class of problem was #28608 (tool cache staged path without `.exe`).

**Related**

- Bare `-cc gcc` still fails with `failed CreateProcess: The system cannot find the file specified.` even though `gcc.exe` is on `PATH`; only an absolute path is accepted. #28590 fixed the *implicit* gcc fallback by passing an absolute path, but an explicit bare name is not resolved through `os.find_abs_path_of_executable`.
- Because the tool cache records this as a source-dependent failure, `~/.cache/v/tools/vself-<hash>.exe.unbuildable` now replays the error for `v -prod self` until the inputs change; users will need to delete it once this is fixed.

**Environment details (OS name and version, etc.)**

```
V 0.5.2 e0d4232 (master, 2026-09-14)
Windows 11 Pro 10.0.26200, x86_64
gcc.exe (MinGW-W64 x86_64-ucrt-posix-seh, built by Brecht Sanders, r2) 16.1.0
clang 21.1.7 (llvm-mingw 20251202 ucrt)
bundled tcc 0.9.28rc
```

quaesitor-scientiam:
Fixes #28625.

### Problem

The V3 driver links every non-TinyCC program into an extension-less `out` inside its per-build directory (`cc_out = os.join_path_single(cc_dir, 'out')`, `-o out` in every compiler invocation) and then `os.mv(cc_out, bin_file)`. MinGW gcc and llvm-mingw clang append the platform suffix to such a target, so the compilation succeeds but the file on disk is `out.exe`; the move then fails and the build directory is cleaned up:

```
failed to finalize <bin>: Source path doesn't exist
```

Since `-prod` moves the implicit compiler from the bundled tcc to gcc, this broke every `-prod` build on Windows, including `v -prod self`, `v -prod -parallel-cc self` and `v up -prod`, as well as any explicit `-cc gcc` / `-cc clang` build. Only tcc writes `out` literally.

### Change

- `v3_produced_binary(cc_out)`: returns `cc_out` if it exists, otherwise an existing `out.exe` / `out.dll` sibling, otherwise `cc_out` unchanged (so a genuinely missing output still fails with the same message). Applied at the single finalize point every non-fastc path converges on (`os.mv`), so `-cc gcc`, `-cc clang`, msvc and the `-parallel-cc` link are all covered. The tcc executable-cache path and the tcc-only fastc path are untouched.
- Unit tests for the helper (`vlib/v/driver/produced_binary_test.v`): exact name wins over a suffixed sibling; `.exe`/`.dll` accepted when the exact name is absent; a missing output stays missing.
- Windows-only driver CLI test that builds and runs a program with whichever of `gcc`/`clang` is on PATH (skips with a message if neither is). It fails on the unpatched driver with the message above and passes with the fix.
- `build_driver_cli_v3_with_flags` now returns the `.exe` name the compiler actually produces on Windows; previously the helper's path never existed there, so none of this file's tests could spawn the driver on Windows.

### Verification (Windows 11)

| build | before | after |
|---|---|---|
| `-cc clang -o hello.exe hello.v` | failed to finalize | runs, prints `hi` |
| `-cc gcc -o hello.exe hello.v` | failed to finalize | runs |
| `-prod -o hello.exe hello.v` | failed to finalize | runs |
| `-prod -parallel-cc -o hello.exe hello.v` | failed to finalize | runs |
| `-prod -o vprod.exe cmd/v` | failed to finalize | `V 0.5.2 366c0b0` (410 s, gcc `-O3 -flto`) |

- `vlib/v/driver/*_test.v` run individually: all pass, including the new file, except `c_compiler_flags_test.v` and `cache_prune_test.v`, which fail identically on unpatched master on this machine (pre-existing, unrelated).
- The new CLI test run in isolation (`VTEST_ONLY_FN`): passes with the fix, fails with `failed to finalize … Source path doesn't exist` on the unpatched driver.

A side effect worth noting: a leftover `out.exe` was one reason the non-recursive `os.rmdir(cc_dir)` after finalize left `.<out>.v3cc.*` directories behind (#28620) on gcc/clang builds; that trigger is gone.

Not verified: a real `-shared` gcc build producing `out.dll` (covered only by the unit test), and `-cc msvc` end to end (blocked locally by #28516 until #28654 lands) — the resolver applies to it the same way.


medvednikov:
Reviewed the code changes (ignoring CI as requested). I don't see a blocking issue here.

The fix is nicely scoped at the common finalize step: `v3_produced_binary()` preserves the exact `out` when the compiler produced it, then falls back to the Windows-style `.exe`/`.dll` sibling, while still returning the original path when nothing exists so the existing failure path is preserved. The per-build directory is unique, so the fallback should not pick up an unrelated stale artifact.

The unit cases cover exact-name precedence, both suffixes, and missing output, and the Windows CLI regression test exercises the original gcc/clang finalize failure end to end. The `build_driver_cli_v3_with_flags` adjustment also looks consistent with what those compilers actually emit on Windows.

No code changes requested from my review.

PR #28666, opened 2026-09-16
toml: explicitly box json2.Any container returns

medvednikov:
## What changed
- explicitly box converted TOML maps and arrays as `json2.Any` in `toml.to.json_any`
- add a regression test for converting a TOML array through `json_any`

## Why
With the new compiler, the implicit array return can enter the forwarded-array conversion path and generate a C assignment from `toml__Any` to `json2__Any`:

```text
error: initializing 'json2__Any' with an expression of incompatible type 'toml__Any'
```

The container contents have already been recursively converted to `json2.Any`, so explicitly boxing the finished container removes the ambiguous contextual conversion without changing the resulting value.

Reproducer:

```sh
v . -new-compiler -cc clang
```

## Validation
- added `test_json_any_array` in `vlib/toml/tests/json_test.v`
- repository CI should exercise the compiler/self-host path


medvednikov:
Follow-up: #28668 restores the original implicit `return jmap` / `return jarr` code and fixes the underlying V3 sum-container matching bug instead of relying on the explicit boxing workaround.

PR #28669, opened 2026-09-16
os: check the extension before touching the file system in is_executable on Windows

quaesitor-scientiam:
### Problem

On Windows `os.is_executable` resolved every path through `real_path` (`CreateFileW` + `GetFinalPathNameByHandleW` + `CloseHandle`), then called `exists` again, and only then looked at the extension. Windows decides executability by extension, so for the overwhelmingly common case — a name that already ends in `.exe`/`.com`/`.bat`/`.cmd` — that file open is wasted work. `find_abs_path_of_executable` calls `is_executable` for every hit, so it sits on the path of every bare-name lookup (and, with #28654, every bare-name `os.new_process`).

### Change

- If the name already carries an executable extension: one `exists` check, done.
- Otherwise (for example a link without an extension): resolve through `real_path` as before and test the target's extension.
- The extension test moved into a small `win_has_executable_extension` helper; the list is unchanged.

Behavioral difference, stated explicitly: a link *named* `foo.exe` whose target is not an executable used to report `false` (the target's extension was tested); it now reports `true`, which is also what `CreateProcessW` does with it. No caller in the tree depends on the old answer.

### Measurements (Windows 11, warm cache, 56 PATH entries)

| | before | after |
|---|---:|---:|
| `is_executable('C:\Windows\System32\where.exe')` | 191 µs | 56 µs |
| bare-name lookup `where` (found in System32) | 560 µs | 395 µs |
| bare-name lookup `gcc` (found on PATH) | 1262 µs | 1026 µs |
| bare-name lookup `git` (found on PATH) | 1154 µs | 898 µs |

The remaining lookup cost is the misses (`exists` on names that are not there, ~19 µs each), which this change does not touch.

### Tests

- New `vlib/os/is_executable_windows_test.v`: every executable extension (incl. mixed case and forward slashes), non-executable extensions, a file without an extension, missing files, and a link without an extension to a `.exe` (skipped when symlinks need privileges).
- `vlib/os/os_test.c.v` `test_is_executable_writable_readable` still passes; the two `test_execute*` failures in that file are pre-existing on this machine and identical without this change.


vlib/os/is_executable_windows_test.v
```diff
@@ -0,0 +1,59 @@
...
+fn testsuite_end() {
+	os.rmdir_all(tfolder) or {}
+}
+
+fn test_is_executable_decides_by_extension_on_windows() {
+	$if !windows {
+		return
+	}
+	eprintln(@FN)
+	for ext in ['exe', 'com', 'bat', 'cmd', 'EXE', 'Cmd'] {
```
JalonSolov, line 22: I would shorten this a bit by doing something like
```suggestion
	for ext.to_lower() in ['exe', 'com', 'bat', 'cmd'] {
```
which matches fine with Windows "case preserving, not case sensitive" naming scheme.

vlib/os/is_executable_windows_test.v
```diff
@@ -0,0 +1,59 @@
...
+		fpath := os.join_path(tfolder, 'tool.${ext}')
+		os.write_file(fpath, 'x')!
+		assert os.is_executable(fpath), 'tool.${ext} must be executable'
+		assert os.is_executable(fpath.replace('\\', '/')), 'tool.${ext} must be executable with forward slashes'
+	}
+	for ext in ['txt', 'ps1', 'vbs', 'js', 'msi', 'dll'] {
```
JalonSolov, line 28: Same thing here...

```suggestion
	for ext.to_lower() in ['txt', 'ps1', 'vbs', 'js', 'msi', 'dll'] {
```

vlib/os/is_executable_windows_test.v
quaesitor-scientiam: Thanks, but this one can't be applied as written, for two reasons:

1. `for ext.to_lower() in [...]` doesn't compile — the `for x in arr` loop variable has to be a plain identifier, not an expression:
   ```
   error: undefined ident: `ext`
   error: left operand to `in` does not match the array element type: expected `string`, not `void`
   ```
2. The mixed-case entries here (`EXE`, `Cmd`) are deliberate: they're the coverage that `is_executable` matches the extension case-insensitively. Normalising the loop values to lower case would drop exactly the case the test exists to check.

Happy to add a short comment noting the mixed-case entries are intentional, if that reads clearer.

vlib/os/is_executable_windows_test.v
quaesitor-scientiam: Same as above — `for ext.to_lower() in [...]` doesn't compile (the loop variable must be a plain identifier). These are all-lowercase already, so there's nothing to normalise here; I kept it symmetric with the executable-extensions loop above.

vlib/os/is_executable_windows_test.v
JalonSolov: Yes, it would make better sense if it were noted as intentional.

quaesitor-scientiam:
Are we good to Merge?

JalonSolov:
You didn't add the comment(s).  Do you plan to do so?

quaesitor-scientiam:
On the mixed-case entries: they're deliberate, so I've kept them. `is_executable` already lowercases the extension before comparing (`win_has_executable_extension`: `path.to_lower().all_after_last('.') in ['exe', 'com', 'bat', 'cmd']`), and feeding it `EXE` and `Cmd` is how the test proves that case-insensitive match actually holds. Normalising the inputs to lower case would remove exactly that coverage. The all-lowercase second loop is kept symmetric with the first on purpose.

So no change here — the check lowers before comparing, and the test is what verifies it.

JalonSolov:
"Happy to add a short comment noting the mixed-case entries are intentional, if that reads clearer."

Your words.

I guess you're not going to do it.

quaesitor-scientiam:
Added in c028ddc1b8 — the comment above the extension loop now says the mixed-case `EXE`/`Cmd` entries are intentional and must not be normalised, since that loop is what asserts `is_executable` matches case-insensitively. Sorry for the round-trip; should have committed it when I first offered. Tests still green.

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I did not find a concrete correctness issue. The fast path preserves the existing executable-extension policy while avoiding `real_path` for the common case, still requires the named path to exist, and retains target-resolution behavior for extensionless links. The mixed-case regression coverage is intentional and now documented.

PR #28677, opened 2026-09-17
net.http: silence vschannel debug logs and use SCHANNEL_NAME on Windows

maaamahAhh:
## Problem

When initializing HTTPS client credentials on Windows, `vschannel.c` prints raw debug error messages directly to `stdout`:

```text
Error 0x80090305 returned by AcquireCredentialsHandle
Error creating credentials
```

In full-screen TUI (`term.ui`) applications, unexpected stdout writes corrupt the terminal display and cause visible screen flickering.

## Cause

1. In `create_credentials()` and `get_new_client_credentials()`, `tls_ctx->sspi` is obtained via `InitSecurityInterface()` (which resolves to the ANSI `InitSecurityInterfaceA` table when compiling without `UNICODE`). However, the code passed `UNISP_NAME_W` (a wide-character `LPCWSTR`) to `AcquireCredentialsHandleA`. Passing a wide string to the ANSI function causes SSPI to read only the single-character string `"M"` due to the null byte in the UTF-16 encoding, failing with `SEC_E_SECPKG_NOT_FOUND` (`0x80090305`).
2. `thirdparty/vschannel/vschannel.c` prints directly to `stdout` via `wprintf()` on handshake and credential errors instead of returning errors to the caller.
3. In `vschannel_init()`, credential creation failures were not recorded via `vschannel_set_last_error()` and did not return early, leaving `creds_initialized = TRUE` even when `create_credentials()` failed.

## Fix

1. In `create_credentials()` and `get_new_client_credentials()`, use the standard `SCHANNEL_NAME` macro (guarded by `#ifndef SCHANNEL_NAME`), which automatically adapts to the character encoding mode of `InitSecurityInterface()`.
2. Route `wprintf` calls in `thirdparty/vschannel/vschannel.c` through a scoped `VSCHANNEL_LOG(...)` macro (`do {} while (0)` by default, active only when `VSCHANNEL_DEBUG` is defined), with `#undef VSCHANNEL_LOG` at the end of the file.
3. In `vschannel_init()`, record the failure status via `vschannel_set_last_error()` and return early on error.

## Tests

- Added `vlib/net/http/vschannel_credentials_windows_test.v` verifying `C.vschannel_init` initializes client credentials with `vschannel_last_error == 0` (fails with `-2146893051` / `0x80090305` prior to fix, passes cleanly after).
- `v test vlib/net/http/vschannel_credentials_windows_test.v` passes.
- `v test vlib/net/http/vschannel_alpn_windows_test.v` passes.
- `v test vlib/net/http/http_httpbin_test.v` and `v test vlib/net/http/head_with_content_length_test.v` pass (verifying live HTTPS requests via vschannel).
- `v fmt -verify vlib/net/http/vschannel_credentials_windows_test.v` passes.


medvednikov:
P2 — vschannel_init() failures are still ignored by the real request paths. The PR now records an initialization error and returns early, which is good. But vschannel_h1_do, vschannel_h2_do, and the pooled transport all call vschannel_init() and immediately continue without checking vschannel_last_error(). Worse, vschannel_open_and_handshake() then calls vschannel_clear_last_error() before attempting the connection.
This means an AcquireCredentialsHandle failure is discarded and the handshake proceeds with an uninitialized credential handle. If InitSecurityInterface() itself returned NULL, the later handshake can dereference tls_ctx->sspi == NULL. I’d guard vschannel_open_and_handshake() before clearing the error, or make vschannel_init() return a status and check it at every caller. The new test does not catch this because it reads last_error immediately after initialization.
Minor — capture GetLastError() before logging. In the InitSecurityInterface() == NULL branch, GetLastError() is read once for VSCHANNEL_LOG() and again afterward for vschannel_set_last_error(). With VSCHANNEL_DEBUG, wprintf() runs between those reads and could overwrite the thread's last-error value. Microsoft explicitly recommends capturing GetLastError() immediately because subsequent functions can change it. Store it in a DWORD err once.

The core UNISP_NAME_W → encoding-neutral SCHANNEL_NAME correction looks sound. Microsoft specifically warns that mixing an encoding-neutral SSPI function alias with a fixed ANSI/Unicode string can cause runtime errors, which matches the root cause described by the PR. The stdout suppression also appears complete for active wprintf calls; remaining occurrences I found are the debug macro itself and commented-out diagnostics.

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I did not find a remaining correctness issue. The current diff now captures `GetLastError()` before logging, returns immediately when SSPI/credential initialization fails, and `vschannel_open_and_handshake()` refuses to proceed when credentials were not initialized. The encoding-neutral `SCHANNEL_NAME` usage matches the `InitSecurityInterface` A/W selection.

Issue #28681, opened 2026-09-17
v3: parallel_parser_test.v fails on master (fallback parse pos mismatch; bench line asserted without -v)


quaesitor-scientiam:
**V version:** V 0.5.2 (master 94a3a56627 on Linux x64; 7cbb67b on Windows 11 x64)

## What happens

`vlib/v/compiler_tests/parallel_parser_test.v` fails on master with two assertions:

```
v -silent test vlib/v/compiler_tests/parallel_parser_test.v
FAIL vlib/v/compiler_tests/parallel_parser_test.v
> assert got.pos == node.pos                                             (test_parallel_parse_falls_back_when_workers_cannot_start, :193)
> assert compile.output.contains('parse .v (parallel)'), compile.output  (test_parallel_parser_compiles_multi_module_project, :413)
```

Same result on Linux (gcc) and Windows (tcc, with the test's `$if !windows` gates
removed locally — the gated bodies fail identically on both platforms).

1. `test_parallel_parse_falls_back_when_workers_cannot_start`: with
   `V3_TEST_PTHREAD_CREATE_FAIL=parser:all` the synchronous-fallback parse produces
   nodes whose `pos` differs from a plain serial `parse_files_with_starts` of the same
   files, while kind/value/typ/children all match. Either the fallback path's file-id
   remap (`remap_worker_file_ids`) is off, or the test's expectation is stale.

2. `test_parallel_parser_compiles_multi_module_project`: the test builds a
   standalone v3 and asserts the compile output contains `parse .v (parallel)`, but
   the bench report is only printed under `-v` (`driver.v`: `if !verbose || silent ||
   c_to_stdout { b.set_quiet() }`), and the test's command line has no `-v`, so the
   output is empty and the assertion can never pass. The program itself compiles and
   prints the expected total.

## Expected

The test passes on master. For (2), either pass `-v` in the test's compile command or
assert on something the default output contains.

## Notes

- The test also drives the child with a `VJOBS=4 <bin> ...` shell prefix, which cmd.exe
  does not understand; setting the variable with `os.setenv` before `os.execute` works on
  every platform (child inherits it).

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Issue #28679, opened 2026-09-17
v3 cgen: no prototype emitted for a \n C.x\ called from a module const initializer (custom_str_test.v fails on master)


quaesitor-scientiam:
**V version:** V 0.5.2 7cbb67b (master, 2026-09-16); also reproduced on Linux at 94a3a56627
**OS:** Windows 11 x64 (tcc) and Linux x64 (gcc)

## What happens

`vlib/v/markused/custom_str_test.v` fails on master, on both platforms:

```
v -silent test vlib/v/markused/custom_str_test.v
...
> assert c_code.contains('hidden_external(void);')
FAIL vlib/v/markused/custom_str_test.v   (test_prelude_global_initializer_seeds_calls_and_c_externs)
```

The assertion is right: the C generator emits the *call* to a `fn C.x` used from a
module-level `const` initializer, but never emits its prototype. Minimal CLI repro:

```v
// hidden/hidden.c.v
module hidden

fn C.hidden_external() int

pub const hidden_value = helper() + C.hidden_external()

fn helper() int {
	return 7
}
```

```v
// main.v
module main

import hidden

fn main() {
	println(hidden.hidden_value)
}
```

`v -o out.c main.v`, then `grep hidden_external out.c` gives only:

```c
hidden__hidden_value = hidden__helper() + hidden_external();
```

`hidden__helper` gets its `i64 hidden__helper(void);` prototype; `hidden_external` gets
nothing. The build still succeeds because the bundled tcc / gcc accept an implicit
declaration under `-w`, so the missing prototype is invisible outside this test (and
would become a hard error with `-cstrict` / a C compiler that rejects implicit
declarations, or return the wrong width for a non-`int` return type).

## Expected

An extern prototype for every `fn C.*` reached from a const initializer, as the test
expects (`int hidden_external(void);`), and the test passing.

## Notes

- Test and behaviour both come from #28556 (`2a7447b5e5`); no earlier history to bisect.
- Not related to worker count / parallelism: identical with `-no-parallel` builds.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
> Other reactions and those to comments will not be taken into account.

Issue #28678, opened 2026-09-17
v3: compiling \module types\ tests with \-d ownership\ never finishes (blocks \ test vlib/v/types/\)


quaesitor-scientiam:
**V version:** V 0.5.2 7cbb67b (master, 2026-09-16); also reproduced on Linux at 94a3a56627
**OS:** Windows 11 (x64, bundled tcc) and Linux x64 (WSL2, gcc)

## What happens

Compiling any `module types` test with `-d ownership` never finishes. The compiler gets
through the "ownership return" fixpoint rounds in ~1s and then makes no further progress
(no output, one core busy) for as long as I let it run — 10 minutes on Windows, 3 minutes
on Linux — before the stage report would print `ck semantics`.

```
v -skip-running -d ownership -v -o out vlib/v/types/checker_ownership_alias_test.v
```

Last lines printed with `-v`:

```
  [ttime]     ck c resolve     12.78 ms
  [ttime]   ownership return round 1: 5193 scanned, 762 changed
  [ttime]   ownership return round 2: 1541 scanned, 74 changed
  [ttime]   ownership return round 3: 144 scanned, 14 changed
  [ttime]   ownership return round 4: 41 scanned, 8 changed
  [ttime]   ownership return round 5: 32 scanned, 8 changed
  (nothing further; killed after 600 s)
```

The test file is 24 lines and only calls `ownership_alias_chain_borrows_indexed_storage`,
so the cost is not in the test body: it is the ownership analysis running over the
`v.types` sources that the test pulls in (`vlib/v/types/checker*.v`,
`checker_ownership_d_ownership.v` is compiled in because of the define).

Because of the `// vtest vflags: -d ownership` header, `v test vlib/v/types/` also sits
on this file indefinitely (observed >1 h before I killed it), so the whole `vlib/v/types`
test directory cannot complete on a developer machine.

## Expected

The compile finishes (the same test compiles in well under a minute without `-d ownership`).

## Notes

- The test was added in #28556 (`2a7447b5e5`).
- Reproduced with the stock `v.exe` in a clean checkout (no local changes), with the same
  binary built from a branch, and on Linux, so it is platform-independent.
- Without `-v` there is no output at all, which makes it look like a hang rather than a
  slow pass; I did not determine whether it is an infinite loop or super-linear work.

> [!NOTE]
> You can use the 👍 reaction to increase the issue's priority for developers.
>
> Please note that only the 👍 reaction to the issue itself counts as a vote.
[...]

Issue #28564, opened 2026-09-13
V3: `monomorphize` memory grows without bound until the compiler is OOM-killed


Jengro777:
## Describe the bug

When V3 compiles a `veb` + `orm` + `json2` program, the compiler's RSS grows
monotonically during the **monomorphize** pass and never plateaus. On this machine
(11.56 GiB RAM + 16 GiB swap) swap is exhausted first, the system becomes
unresponsive, and the compiler is OOM-killed without printing any diagnostic.

The issue was originally hit while building the RuoQi-v backend (368 `.v` files,
~23k lines), but the reproduction below does not need that project: it is a single
self-contained `.v` file that generates the same shape of code and runs it.

## Reproduction: one file, one run

```
// Single-file reproduction for the V3 monomorphize OOM.
//
// Usage:
//   v -new-compiler run v3_monomorphize_oom.v            (optional counts: V J O)
// (or: v v3_monomorphize_oom.v && ./v3_monomorphize_oom - `v file.v` alone only
//  builds a binary, it does not execute it)
//
// It writes repro.v next to the working directory:
//   * 126 veb controllers registered through the same generic helpers as RuoQi-v,
//   * 533 distinct json2.decode[T] handlers,
//   * 205 orm models with 4 `sql db {}` queries each,
// then compiles it with `v -new-compiler` while sampling the compiler's RSS every
// second. Once the RSS reaches 8 GiB the compile is killed, so the reproduction
// is confirmed without pushing the host into swap/OOM. If the compile does finish
// (e.g. on a fixed compiler), the generated program is run directly as well.
module main

import os
import strings
import time

const v_count = 126
const j_count = 533
const o_count = 205
const limit_mb = 8192
const timeout_s = 900
const out_file = 'repro.v'

fn generate(v_count int, j_count int, o_count int) {
	mut sb := strings.new_builder(1 << 20)
	sb.writeln('module main')
	sb.writeln('')
	sb.writeln('import veb')
	sb.writeln('import json2')
	sb.writeln('import orm')
	sb.writeln('')

	// --- veb controllers, registered through the same generic helpers as RuoQi-v
	sb.writeln('pub struct Ctx {')
	sb.writeln('\tveb.Context')
	sb.writeln('}')
	sb.writeln('')
	sb.writeln('pub struct ModelApp {')
	sb.writeln('\tveb.Middleware[Ctx]')
	sb.writeln('\tveb.Controller')
	sb.writeln('}')
	sb.writeln('')
[...]

PR #28686, opened 2026-09-17
v3: run the compiler phases and -prod C compilation in parallel on Windows

quaesitor-scientiam:
## Behavior change

On Windows the V3 compiler ran every phase single-threaded: `workers.Pool` was a serial stub and the parser, checker, markused, transform and cgen phases each had `$if windows { serial } $else { parallel }` dispatch gates. This PR ports the pool (commit 1), removes the gates (commit 2) and enables parallel C compilation for `-prod` builds (commit 3), so Windows runs the same parallel code paths Linux and macOS already run in CI.

Building `cmd/v` on a 24-core Windows 11 box (bundled tcc, `-gc none`, `VJOBS` unset):

| | before | after |
|---|---|---|
| total | 37.2 s | **6.6 s** |
| check / transform / cgen | 9.9 / 9.5 / 13.8 s | 1.3 / 1.6 / 1.8 s |
| tcc | 0.5 s | 0.5 s |

WSL on the same host is 5.8 s. `-no-parallel` used to be *faster* than the default on Windows (30 s vs 37 s), because the serial branches still paid the parallel bookkeeping.

Consequences worth knowing:
- Windows generated C now has the parallel path's shape (e.g. fixed const arrays are emitted as C arrays, as on Linux). Generated C varies with the worker count on all platforms already (pre-existing, not changed here).
- `vlib/v/checker/tests/import_unused_warning.vv` prints nothing on Windows now — matching Linux master, where the parallel unused scan already drops that warning (pre-existing; will report separately).
- `-d v3_no_parallel` remains the fully serial build. The POSIX-only mmap'd C output write is untouched.

## Design

Only thread creation/join is platform specific, so `worker_pool_nix.c.v` becomes the shared `worker_pool.c.v` and gains two ~25-line shims: `thread_nix.c.v` (pthreads, unchanged behaviour) and `thread_windows.c.v` + `vlib/v/winthread_helper.h` (`CreateThread` with `STACK_SIZE_PARAM_IS_A_RESERVATION` so the 64 MiB worker stack is reserved, not committed; `WaitForSingleObject` + `CloseHandle` to join). V's `spawn` was not used because the pool's stack size is chosen at runtime (`V3_WORKER_STACK_MB`) and `spawn` would commit the full stack per worker on Windows.

## Parallel C compilation on Windows (third commit)

`-parallel-cc` was excluded on Windows outright (`target.os != 'windows'`), so `v self -prod` — which adds `-parallel-cc` automatically — always ran one monolithic `gcc -O2 -flto` over the 22 MB translation unit. The unit compiles go through `workers.Pool`, which was the serial stub, so the exclusion was consistent at the time. The third commit lifts it, and makes the two Windows-only C headers with file-static state replicable across units (`closure_once_windows.h` gets the owner/`extern` scheme of its POSIX sibling; `toolcache_windows_helpers.h` drops a function-local static in its tcc-only loader). It also adds `V3_PARALLEL_CC_JOBS`, which can only *raise* the platform-independent cap of 2 jobs for one build (still bounded by available jobs), so a developer box can use its cores without changing the CI memory profile.

`-prod` build of `cmd/v` on the same machine (gcc 16.2):

| | total |
|---|---|
| monolithic `gcc -O2 -flto` (before) | 363 s |
| `-parallel-cc`, default cap of 2 jobs / 9 units | 153 s |
| `-parallel-cc`, `V3_PARALLEL_CC_JOBS=8` / 33 units | 63–66 s |

Compilers built through the split emit byte-identical C to the tcc-built debug compiler from the same source. Note the trade upstream already made on POSIX now applies to Windows too: split units compile at `-O3` without LTO. If some other native header on Windows turns out to have unhandled static state, the driver falls back to the monolithic build (slow, but correct) rather than failing.

## Tests run (Windows 11, tcc)

- `v test vlib/v/workers/` — 2/2
- `v test vlib/v/gen/c/` — 22/22 (includes `parallel_worker_test.v`)
- `vlib/v/types/checker_parallel_test.v`, `vlib/v/gen/c/types_test.v`, `vlib/v/compiler_tests/parallel_worker_failure_test.v`, `vlib/v/transform/transform_parallel_test.v` — pass with their Windows gates removed (the launch-failure test forces every phase's workers to fail and checks the C is identical).
- `vlib/v/driver/parallel_cc_test.v`, `vlib/v/driver/environment_test.v` — pass (the latter gains a test for the `V3_PARALLEL_CC_JOBS` contract); a tcc-built compiler was run through the tool-cache path to exercise the modified advapi32 loader.
- Self-host fixed point: the compiler built by the parallel compiler reproduces its own `cmd/v` C output byte for byte; a `-d`-gated serial build of the same source reproduces the stock compiler's output exactly.
- `vlib/v/compiler_errors_test.v` — identical pass/fail set between the parallel and serial-gated builds (1590 fixtures fail on this machine either way, all from the missing V1 fallback, not diagnostics).
- Commit 1 was also built and tested on its own (pool live, gates still serial, timing unchanged).

Pre-existing failures found along the way and filed separately: #28678 (`-d ownership` compile never finishes), #28679 (missing extern prototype in cgen), #28681 (`parallel_parser_test.v` fails on master — its Windows gates are deliberately left in place here).

Not run: `-cc gcc`/`-prealloc` explicit variants; macOS/Linux CI is the check that the POSIX side is unchanged.

## Touched files

`vlib/v/workers/{worker_pool.c.v (renamed from worker_pool_nix.c.v), thread_nix.c.v, thread_windows.c.v, worker_pool_windows.c.v (deleted), worker_pool_test.v, worker_pool_limit_test.v}`, `vlib/v/winthread_helper.h`, `vlib/v/parser/parser_parallel_notd_v3_no_parallel.v`, `vlib/v/types/{checker_parallel.v, checker_parallel_test.v}`, `vlib/v/markused/markused_parallel_notd_v3_no_parallel.v`, `vlib/v/transform/transform_parallel_notd_v3_no_parallel.v`, `vlib/v/gen/c/{fn_parallel_notd_v3_no_parallel.v, cleanc.v, types_test.v}`, `vlib/v/compiler_tests/parallel_worker_failure_test.v`, `vlib/v/driver/{driver.v, environment_test.v}`, `vlib/builtin/closure/closure_once_windows.h`, `cmd/v/toolcache_windows_helpers.h`.


vlib/v/winthread_helper.h
```diff
@@ -0,0 +1,28 @@
...
+#include <stddef.h>
+
+/* Compiler workers run deeply recursive phases, so the requested stack size
+ * is a reservation (like pthread_attr_setstacksize), not just the initial
+ * commit; without STACK_SIZE_PARAM_IS_A_RESERVATION the reserve would stay at
+ * the PE default and the workers would overflow on the same inputs the POSIX
+ * pool handles. */
+static inline void *v3_win_thread_create(size_t stack_size,
+	void *(*start_routine)(void *), void *arg) {
+	return (void *)CreateThread(NULL, stack_size, (LPTHREAD_START_ROUTINE)start_routine,
```
medvednikov: This cast is unsafe on 32-bit Windows. `LPTHREAD_START_ROUTINE` is `DWORD (WINAPI *)(LPVOID)` (`__stdcall` on x86), while `start_routine` is declared as the default C calling convention (`void *(*)(void *)`, i.e. cdecl). On Win32 the callee/caller stack-cleanup rules differ, so returning from a worker can corrupt the stack even though the argument shape happens to match. x64 hides this because the calling conventions are unified. Please use a real `WINAPI` thunk (carrying the V callback + arg in a small context), or otherwise ensure the callback itself has the Windows thread calling convention rather than casting between them.

medvednikov (commented):
Reviewed the code changes only (CI status not considered). I found one Windows correctness issue in the new thread shim.

vlib/v/winthread_helper.h
quaesitor-scientiam: Good catch, thanks. Fixed in 755565e992: the callback now runs behind a `static DWORD WINAPI v3_win_thread_thunk(LPVOID)` carrying `{start_routine, arg}` in a small context, mirroring the `__v_windows_thread_start` thunk cgen emits for `spawn`. The context is allocated from the process heap (`HeapAlloc`) rather than `malloc`, so the creating thread's `-prealloc` arena is not involved; a failed `CreateThread` frees it and the pool records a launch failure as before. The branch is also rebased onto current master (the transform conflict from #28564's scoped monomorphize was resolved by re-applying the gate removal on top of it).

medvednikov (commented):
Reviewed the current head, `6d3b982ad4942d0272618be3d54eceee4511c041`. I did not find an additional actionable correctness issue in this review.

The earlier Win32 callback-ABI concern is addressed by the real `DWORD WINAPI` thunk. The launch context is allocated independently of V's prealloc arena, freed before the callback runs, and also freed when `CreateThread` fails. The shared pool retains its synchronous fallback when no workers launch and surfaces join failures. I also checked the phase-dispatch changes, the parallel-C job cap, and the Windows headers' split-translation-unit handling.

Validation was limited: a local C harness using mocked WinAPI calls passed the allocation-failure, thread-creation-failure, callback/context-cleanup, stack-reservation-flag, wait-failure, close-failure, and successful-join checks. Clang syntax/type checks targeting both i686 Windows and x86_64 Windows also passed against mocked WinAPI declarations. These checks do **not** substitute for native Windows execution; I did not run the V build/test suite here.

The remaining validation worth adding is an actual 32-bit Windows worker-lifecycle regression test and a Windows `-prealloc` compiler build/test run. Those would protect the ABI and allocator-lifetime paths directly; I am not reporting the missing coverage as a demonstrated defect.

quaesitor-scientiam:
Thanks for the review. On the two validation gaps:

**32-bit Windows worker lifecycle** — run natively now. The driver cannot link `windows/x86` from an amd64 host, so I emitted the C with `v -arch x86 -os windows -prealloc -o wp32.c vlib/v/workers/worker_pool_test.v` (the generated call is `v3_win_thread_create(stack_size, (fn_ptr)(start_routine), arg)` going through the `DWORD WINAPI` thunk) and compiled it with llvm-mingw's `i686-w64-mingw32-clang`. The result is a genuine i386 PE (machine 0x14C) and it passes all four tests in that file (persistent batches with a forced-sync task, every injected launch-failure index including `pool:all` with the synchronous fallback, and draining a 256-task batch on one worker), exit 0. That exercises the stdcall thunk, the context free on the worker, the stack-reservation flag and the join path on a real 32-bit thread.

**Windows `-prealloc` compiler build/test** — compiler builds default to `-prealloc` on this branch (`cmd/v` gets the `prealloc` define unless `-no-prealloc`), so every stage of the three-stage bootstrap and the `v test` runs reported above (`vlib/v/workers/`, `checker_parallel_test.v`, `environment_test.v`, `parallel_worker_failure_test.v`) already ran on a `-prealloc` compiler with 23 persistent workers; the 32-bit binary above was also built with `-prealloc`.

No code changes from this; the head is still 6d3b982ad4.

medvednikov (commented):
Follow-up on `6d3b982ad4942d0272618be3d54eceee4511c041` after the validation update: the head is unchanged, and I have no outstanding actionable correctness findings from this review. CI was not considered.

The reported native i386 Windows worker-pool run with `-prealloc` addresses the native-execution question from my previous comment. I rechecked the worker tests and the `DWORD WINAPI` thunk; the tests cover persistent batches, forced-sync work, injected launch failures including the all-workers-failed fallback, and draining an oversized batch.

I also confirmed in `vlib/v/driver/driver.v` that compiler builds add the `prealloc` define by default unless `-no-prealloc` is specified. Given your clarification that the reported bootstrap and compiler test runs used that default, my earlier characterization of Windows `-prealloc` compiler validation as missing was too broad.

The validation questions from my last comment are addressed by the supplied report and source clarification. The native Windows execution results are author-reported; I did not independently rerun those tests in this follow-up.

PR #28697, opened 2026-09-18
cgen: read the TinyCC Windows thread-local ready flag with a plain load

quaesitor-scientiam:
TinyCC has no `_Thread_local`, so on Windows the C generator turns every thread-local global into a Win32 Fls/Tls slot behind a macro (`#define g_memory_block (*g_memory_block_slot())`), and every access runs `<name>_key_init()`. That function opened with

```c
if (__atomic_add_fetch(&g_memory_block_key_ready, 0, 5)) { return; }
```

— a seq_cst locked read-modify-write of one shared word, executed by every thread on every access. With `-prealloc` (the default for compiler builds) that word guards `g_memory_block`, the per-thread arena root touched several times per `malloc`. Once the compiler phases run in parallel on Windows (#28686), the workers turn that line into a cache-line ping-pong that grows with the worker count. The pthread-key slot used on Linux has no such cost: `pthread_once`'s fast path is a plain load.

This PR reads the ready flag with a plain `volatile` load on x86 (loads are acquire loads there, and the seq_cst publish already orders the key and `Fls*` pointer stores before the flag). Other architectures keep the atomic, since TinyCC has no `__atomic_load_n`. The one-time claim/publish sequence is unchanged, and the helper macro is `#undef`'d after use.

### Measurements

Building `cmd/v` on a Ryzen 9 5900X (12C/24T), Windows 11, tcc, with the parallel phases of #28686, two-stage bootstrap so the compiler itself carries the new preamble; WSL (Debian, same commit, same box) for reference:

| `VJOBS` | before | after | WSL |
|---|---|---|---|
| 4 | 10.36 s / 28.6 s CPU | 10.22 s / 27.2 s | 9.70 s / 29.6 s |
| 16 | 6.42 s / 57.2 s CPU | **5.34 s / 40.9 s** | 5.28 s / 43.8 s |
| default (23) | 6.35 s / 61.0 s CPU | **5.68 s / 42.0 s** | 4.94 s / 45.7 s |

On current master alone (serial phases on Windows) the change is not measurable; it only pays off with more than ~4 threads touching thread-locals.

### Checks

- `vlib/v/gen/c/preamble_test.v` updated (asserts the x86 plain load, the non-x86 atomic fallback, the `#undef`) and passing.
- Three-stage tcc self-build on Windows, plus a `-cc gcc -prod` stage from the result.
- `v test vlib/v/gen/c/` (22/22) with the rebuilt compiler; `vlib/v/tests/concurrency/` compiled and run per file: same three pre-existing failures as clean master (`shared_interface_test.v`, `shared_map_delete_test.v`, `volatile_vars_test.v`).
- `vlib/sync/` currently fails 7 tests on master with "cannot import `sync` into a module with the same name", unrelated to this change.


medvednikov (commented):
Reviewed the code changes only (CI status not considered). I did not find a substantive correctness issue in this diff. The x86 fast path is narrowly scoped to TinyCC/Windows, keeps the existing one-time atomic claim/publish sequence, and the regression test checks both the x86 load path and the non-x86 fallback.

PR #28698, opened 2026-09-18
v3 cgen: use pointer access for promoted methods on `mut` generic parameters

### LINKED ISSUE #28638: `v3 cgen: a promoted method on a `mut` generic parameter is emitted as `ctrl.field` instead of `ctrl->field``
### Describe the bug

`gen_embedded_method_receiver` chooses the first accessor from the V type of the
base expression (`base_type is types.Pointer`). A `mut` parameter is passed as a
pointer in C while its V type is a value, so a method that is promoted through
embedded fields is emitted as `ctrl.field` and the generated C does not compile:

```
src.c: error: 'ctrl' is a pointer; did you mean to use '->'?
```

### Reproduction

Four small files, no external dependencies (`issue-drafts/repro/cgen-mut-param-receiver`):

```sh
v -new-compiler -o app .
```

`v.mod`:
```v
Module{
	name: 'mut_param_receiver'
}
```

`veblite/mw.v`:
```v
module veblite

pub struct Mw[T] {
mut:
	used int
}

pub struct Opts[T] {
pub:
	handler fn (mut ctx T) bool
}

pub fn (mut m Mw[T]) use(o Opts[T]) {
	if o.handler != unsafe { nil } {
		m.used++
		println('use')
	}
}
```

`model/app.v`:
```v
module model

import veblite

pub struct Ctx {}

pub struct App {
	veblite.Mw[Ctx]
}
```

`route/route.v`:
```v
module route

import model
import veblite

pub struct App {
	model.App
}

pub struct Api {
	model.App
}

fn (mut app App) common_middleware[T](mut ctrl T, mut ctx model.Ctx) {
	ctrl.use(veblite.Opts[model.Ctx]{
		handler: fn (mut ctx model.Ctx) bool {
			return true
		}
	})
}

pub fn (mut app App) register_routes[T](mut ctrl T, path string, mut ctx model.Ctx) {
	app.common_middleware[T](mut ctrl, mut ctx)
	ctrl.use(veblite.Opts[model.Ctx]{
		handler: fn (mut ctx model.Ctx) bool {
			return true
		}
	})
}
```

`main.v`:
```v
module main

import model
import route

fn main() {
	mut app := &route.App{}
	mut ctx := model.Ctx{}
	app.register_routes[route.Api](mut &route.Api{}, '/api', mut ctx)
	println('ok')
}
```

### Expected behaviour

The program compiles and prints:

```
use
use
ok
```

### Actual behaviour

```
C compilation error (from cc):
src.c: In function 'route__App_route__Api__common_middleware':
src.c: error: 'ctrl' is a pointer; did you mean to use '->'?
```

### Fix

Ask the codegen about the ident's storage instead of relying on the V type:

```v
mut access_is_ptr := cgen_type_is_pointer_like(base_type)
	|| g.receiver_ident_storage_is_pointer(base_id)
```

### Environment

```
```

> [!NOTE]
>

Jengro777:
## Why

`gen_embedded_method_receiver` chose the accessor from the V type of the base
expression. A `mut ctrl T` parameter is a value in V but a pointer in C, so a
promoted method reached through an embedded field of such a parameter was
emitted as `ctrl.field` and the C compiler rejected the generated code:

```
src.c: error: 'ctrl' is a pointer; did you mean to use '->'?
```

## What changed

Ask codegen about the ident's storage before choosing the first accessor:

```v
mut access_is_ptr := cgen_type_is_pointer_like(base_type)
	|| g.receiver_ident_storage_is_pointer(base_id)
```

## Test plan

`vlib/v/compiler_tests/generic_mut_param_receiver_test.v` generates a small
`veb`-style project at run time (embedded `Middleware[T]` / `Controller` structs
and a generic `mut ctrl T` helper) and asserts that it compiles and runs. With
the fix reverted the generated C fails with `'ctrl' is a pointer`.

```sh
v -g -keepc -o vnew cmd/v
./vnew -silent test vlib/v/compiler_tests/generic_mut_param_receiver_test.v
```

## Notes

(https://github.com/RuoQi-DoDo/RuoQi-v) compile with `-new-compiler`
(`cd RuoQi-v/backend && v -new-compiler ./main`); without this fix the promoted
`ctrl.use(...)` / `ctrl.route_use(...)` calls of a generic controller helper do
not compile.

Fixes #28638.


Jengro777:
request for review

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I did not find a remaining concrete correctness issue. The earlier parenthesized-receiver and shell-path concerns are addressed: `receiver_ident_storage_is_pointer()` unwraps parens before consulting storage, and the regression harness now quotes the compiler/project/output paths. The production change stays narrowly scoped to choosing the first accessor for pointer-backed receiver storage.

Issue #28489, opened 2026-09-11
V3 reports bogus `unknown function` diagnostics for generic functions that call methods on a generic parameter, then silently falls back to V1


Jengro777:
### Describe the bug

With the V3 front-end enabled by default, a program built on `veb` that both

1. uses `veb.Middleware[T]` (generic embedded struct), and
2. calls methods on a **generic type parameter** inside generic helper functions
   (`ctrl.use(...)`, `ctrl.route_use(...)`, `app.register_controller[T, U](...)`),

makes V3 abort during the transform/monomorphization stage with an unpositioned
diagnostic:

```
type checker found 1 error(s):
unknown function `unknown.get`
```

The message carries **no file:line**, the number of such messages is
**nondeterministic** between runs (in our real project we saw 8, 10, 11, 12 and 17
messages for the exact same source), and no binary is produced by V3.

However `v` then **silently falls back to V1**, which compiles the very same code
successfully, so the command exits with code `0` and a working binary is produced.
The result is that a successful build always prints errors, and CI logs cannot be
trusted:

```
$ v -o app .
type checker found 1 error(s):
unknown function `unknown.get`
$ echo $?
0                      # binary produced, build actually succeeded
```

Forcing V3 with `-new-compiler` (no fallback) fails on the same code, and forcing V1
with `-old-compiler` is completely clean — so the diagnostic is a V3 false positive,
not a real program error.

### Reproduction Steps

Environment: `V 0.5.2 8b1c7b1` (Linux x86_64, v3 enabled by default,
veb from `vlib/veb`).

Minimal reproduction — this is a complete program, no external dependencies:

```v
module main

import veb

pub struct Ctx {
	veb.Context
}

// stands for the shared base struct embedded by every controller
pub struct ModelApp {
	veb.Middleware[Ctx]
	veb.Controller
}

// stands for a concrete controller
pub struct Item {
	ModelApp
}

// the main app
pub struct MainApp {
	veb.Middleware[Ctx]
	veb.Controller
}

fn mw() veb.MiddlewareOptions[Ctx] {
	return veb.MiddlewareOptions[Ctx]{
		handler: fn (mut ctx Ctx) bool {
			return true
		}
	}
}

// generic helper: method call on the generic parameter T
fn (mut app MainApp) common_middleware[T](mut ctrl T) {
	ctrl.use(mw())
}

fn (mut app MainApp) register_routes_no_auth[T, U](mut ctrl T, url_path string) {
[...]

PR #28699, opened 2026-09-18
v3: resolve bare and alias-qualified type names in the file that wrote them

### LINKED ISSUE #28639: `v3: bare type names in generic calls are resolved outside the writing file's scope`
### Describe the bug

When a file imports a module selectively (`import iam { Token }`) while another
imported module declares a same-named type (`json2.Token`), the monomorphizer and
codegen resolve the bare spelling `Token` through a global first-wins short-name
index instead of the writing file's imports. The call-site rewrite and the
emitted specialization then disagree, and the build dies in the C linker (or, in
larger programs, with the bogus `unknown function` diagnostics of #28489 / with
fields of the wrong homonym).

### Reproduction A - type argument (issue-drafts/repro/generic-type-arg-homonym)

`v.mod`: `Module{ name: 'ambig2' }`

`json2/tok.v`:
```v
module json2

pub struct Token {
pub mut:
	kind int
}
```

`iam/tok.v`:
```v
module iam

pub struct Token {
pub mut:
	id int
}

pub fn (mut t Token) hello() {
	t.id++
}
```

`route/route.v`:
```v
module route

import json2
import iam { Token }

fn call_hello[T](mut ctrl T) {
	ctrl.hello()
}

pub fn run() int {
	mut t := Token{}
	call_hello[Token](mut t)
	return t.id
}
```

`main.v`:
```v
module main

import route

fn main() {
	println(route.run())
}
```

```sh
v -new-compiler -o app .
```

Actual:
```
C compilation error (from cc):
src.c:(.text+0xc8e5): undefined reference to `route__call_hello_T_json2__Token'
```

### Reproduction B - struct-literal argument (issue-drafts/repro/generic-literal-arg-homonym)

`v.mod`: `Module{ name: 'litgen' }`

`aaa/a.v`:
```v
module aaa

pub struct Token {
pub mut:
	lit []u8
	kind int = 3
}
```

`bbb/b.v`:
```v
module bbb

pub struct Token {
pub mut:
	id int
}

pub fn take[T](mut ctrl T) {
	println(T.name)
}
```

`main.v`:
```v
module main

import aaa
import bbb { Token }

fn main() {
	mut t := Token{}
	bbb.take[Token](mut t)
	_ := aaa.Token{}
}
```

Actual:
```
C compilation error (from cc):
src.c:(.text+0xb7fa): undefined reference to `bbb__take_T_Token'
```

### Reproduction C - alias-qualified enum helper (3 files)

`v.mod`: `Module{ name: 'alias_enum' }`

`token/token.v`:
```v
module token

pub struct Token {
pub mut:
	id int
}
```

`main.v`:
```v
module main

import toml
import token

fn main() {
	toml.parse_text('a = 1') or { panic(err) }
	t := token.Token{
		id: 7
	}
	println(t.id)
}
```

Declaring a module called `token` next to `import toml` makes vlib's nested
`toml/token` module be named `toml.token`, while vlib's own
`toml/parser/parser.v` keeps writing the enum as `token.Kind`. The helper lookup
then fell back to the bare `Kind`:

```
C compilation error (from cc):
src.c: In function ‘parser__Parser__check_one_of’:
src.c:9935:73: error: incompatible type for argument 2 of ‘string__plus’
 9935 |    __arr_str_3 = string__plus(__arr_str_3, Kind__autostr(__arr_str_it_5));
      |                                                    ^~~~~~~~~~~~~~~~~~~~~~~~~
      |                                                    |
      |                                                    int
src.c:2944:38: note: expected ‘string’ {aka ‘struct string’} but argument is of type
                    ‘int’
 2944 | string string__plus(string s, string a);
      |                               ~~~~~~~^
```

### Expected behaviour

All three programs compile (Reproduction A prints `1`, Reproduction B prints
`bbb.Token`, Reproduction C prints `7`); the bare `Token` means the selectively
imported `iam.Token` / `bbb.Token` in the file that wrote the call, and
`token.Kind` means `toml.token.Kind` in a file that imported `toml.token`.

### Regression tests

`vlib/v/compiler_tests/selective_import_generic_arg_test.v` covers both cases:
`test_selective_import_bare_type_arg_wins_over_imported_homonym` (Reproduction A)
and `test_selective_import_bare_literal_arg_wins_over_imported_homonym`
(Reproduction B), plus `test_path_import_alias_enum_autostr_uses_declaring_module`
for Reproduction C.

### Notes

The same class of bug was observed in a larger `veb` project this way:

- a controller literal `&Token{}` was emitted as a `token.Token` cast with
  `json2.Token` field defaults (`.lit`, `.kind`) -> C error `'token__Token' has
  no member named 'lit'`;
- a bare `Context` in a synthesized node was locked to `main.Context` -> C error
  `unknown type name 'main__Context'` and a missing `_fn_ptr_*` typedef (see
  below);
- an alias-qualified enum `token.Kind` produced the helper call `Kind__autostr`
  while the definition is `toml__token__Kind__autostr`.

### Notes - `main__Context` (no small standalone reproducer yet)

`collect_generic_struct_specs_range` also walks synthesized nodes. A node that
hangs under no declaration has neither module nor file context, so a plain
generic argument in its type text (`veb.Middleware[Context]` written by a file
that did `import model { Context }`) was collected as written and the
specialization rebased `Context` onto `main`:

```
src.c:4161:41: error: unknown type name ‘main__Context’; did you mean ‘veb__Context’?
 4161 | typedef bool (*_fn_ptr_52d5bf3cc77ad60)(main__Context*);
src.c:11798:9: error: unknown type name ‘_fn_ptr_52d5bf3cc77ad60’
```

Such spellings are now skipped (`type_text_has_unqualified_generic_arg`), while
fully qualified arguments keep being collected. Reproducing this end-to-end
needed a large multi-module `veb` program; the regression test
`vlib/v/transform/monomorphize_test.v: test_contextless_generic_struct_spec_spelling_is_skipped`
pins the collection behaviour directly.

### Fix

Resolve bare names (and import-alias prefixes) through the writing file before
falling back to global short-name tables:

- `transform/fn.v`: `generic_call_type_arg_name` uses the file's selective
  imports; `enum_autostr_type_name` maps an import-alias prefix to its module;
- `transform/struct.v`: `lookup_struct_info` uses the file's selective imports;
- `gen/c/struct.v`: `find_struct_decl` uses the file's selective imports;
- `gen/c/types.v`: `enum_autostr_c_name` uses the file's imports and aliases.

### Environment

```
```

Jengro777:
## Why

A bare type spelling (`Token`, `Context`, `Kind`) and an import-alias prefix
(`token.Kind` for `import toml.token`) belong to the imports of the file that
wrote them. The rewrite and codegen looked names up in global first-wins
short-name tables instead, so a same-named type from *another* imported module
won, and the call-site rewrite disagreed with the emitted specialization:

- `undefined reference to 'token__Token__register_routes_authenticated'`
  (or, in bigger programs, the bogus `unknown function` family of #28489);
- a controller literal `&Token{}` emitted with another module's fields/defaults
  (`'token__Token' has no member named 'lit'`);
- struct fields of a homonym leaking into literals
  (`find_struct_decl` / `lookup_struct_info`);
- `Kind__autostr(...)` helper calls for an enum whose definition is
  `toml__token__Kind__autostr` (this breaks vlib's own `toml.parser` once the
  program also declares a module called `token`).

## What changed

Resolve through the writing file's selective imports and import aliases in the
transform and in cgen, where that file is known:

- `transform/expr.v`: `selective_import_type_name_for_file()`;
- `transform/fn.v`: generic call type arguments and the enum `__autostr` helper
  name (including import-alias prefixes);
- `transform/struct.v`, `gen/c/struct.v`: struct lookup for literals;
- `transform/monomorphize.v`: `generic_struct_args_in_scope()` canonicalizes the
  file-scope arguments of a generic struct specialization;
- `gen/c/types.v`: `enum_autostr_c_name()` resolves file imports for references
  and uses a declaration-only lookup when emitting the helper definitions, and
  it keeps an explicit `main.` type lock intact.

## Test plan

`vlib/v/compiler_tests/selective_import_generic_arg_test.v` generates four small
projects at run time (bare type argument over an imported homonym, bare literal
argument, alias-qualified enum helper, and a selective import that wins over a
same-named local declaration, matching the checker's `qualify_type_text_impl`).
`vlib/v/gen/c/types_test.v` covers the enum helper naming directly.

```sh
v -g -keepc -o vnew cmd/v
./vnew -silent test vlib/v/compiler_tests/selective_import_generic_arg_test.v
./vnew -silent test vlib/v/gen/c/types_test.v
```

## Notes

(https://github.com/RuoQi-DoDo/RuoQi-v) compile with `-new-compiler`; RuoQi
imports `model { Context }` in many files and also has its own `token` module,
which is exactly the collision pattern above.

Fixes #28639.


Jengro777:
request for review

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. One correctness issue remains: `generic_call_type_arg_name()` resolves every bare identifier through the writing file's selective imports before checking whether that identifier is an active generic parameter. In a generic function with parameter `T`, a same-file `import pkg { T }` can therefore rewrite the lexical type parameter to `pkg.T`, unlike the checker where the generic parameter has precedence. The earlier generic-sum/file-scope and alias/main-lock findings appear addressed on this head, but this generic-parameter precedence case still needs a guard/regression.

Jengro777:
fixed

squidink7:
I believe I encountered this when trying to use json and toml in the same project, the compiler got confused between the two Any types:
```
src.c:61092:21: error: initializing 'json2__Any' (aka 'struct json2__Any') with an expression of incompatible type 'toml__Any' (aka 'struct toml__Any')
 61092 |                                                                         json2__Any __return_array_value_8 = *(toml__Any*)(array_get(__return_array_5, __return_array_idx_7));
       |                                                                                    ^                        
```

vlib/v/gen/c/fn.v
```diff
@@ -12821,11 +12821,10 @@ fn (g &FlatGen) current_mut_param_binding_is_shadowed(name string) bool {
...
 fn (mut g FlatGen) gen_enum_str_call(fn_node &flat.Node, enum_type types.Enum) {
-	mut name := enum_type.name
-	if name.starts_with('main.') {
-		name = name[5..]
-	}
-	g.write('${g.enum_autostr_c_name(name)}__autostr(')
+	// Keep an explicit `main.` prefix: `enum_autostr_c_name` uses it as the
+	// program-module lock, so stripping it here would let the current file's
+	// selective imports capture a main-module enum.
+	g.write('${g.enum_autostr_c_name(enum_type.name)}__autostr(')
```
medvednikov: **[P2] Do not resolve an already-checked enum identity through source import aliases**

`enum_type.name` here comes from the receiver's checked `types.Enum`, but `enum_autostr_c_name()` now treats every dotted name as a spelling from the current file and applies `file_imports` before checking `enum_names`. Those are different namespaces. With `import a as real_a` and `import b as a`, a value returned by `real_a.make()` has canonical type `a.Kind`; the new lookup follows the caller's `a -> b` alias and selects `b__Kind__autostr` instead of `a__Kind__autostr`. Thus a valid alias arrangement can format a value using the wrong enum's members, or reference a helper that was not emitted. The added alias-rewrite branch in `transform/fn.v::enum_autostr_type_name` has the same risk when given a resolved type name.

Suggested regression project (with a root `v.mod`):

```v
// a/a.v
module a

pub enum Kind {
    from_a
}

pub fn make() Kind {
    return .from_a
}
```

```v
// b/b.v
module b

pub enum Kind {
    from_b
}
```

```v
// main.v
module main

import a as real_a
import b as a

fn main() {
    value := real_a.make()
    other := a.Kind.from_b
    assert other.str() == 'from_b'
    assert value.str() == 'from_a'
    assert '${value}' == 'from_a'
}
```

Please keep file-aware resolution for raw source spellings, but use an import-free naming path for already-resolved enum identities (as the new declaration-only helper already does). Apply that distinction in both transform and C generation; merely moving the global enum lookup ahead of aliases would undo the new `token.Kind` source-spelling fix. This example is derived from the call path and has not been executed in this review environment.

medvednikov (commented):
Reviewed all nine changed files and the relevant surrounding resolution/code-generation paths at `b82589720d896ae11b809f5cb9f6a52a99289207`, excluding CI status. One correctness finding is attached inline: checked enum identities can be reinterpreted through an unrelated file import alias.

Validation: source-level review only. I could not build or execute the regression tests locally because the review environment could not resolve github.com to clone the repository; the example in the finding is a proposed regression, not a claimed test run.

vlib/v/gen/c/fn.v
medvednikov: Fixed in eb96b38210.

- cgen: `enum_autostr_c_name()` is back to its import-free form (and the declaration-only split is gone). Every caller passes a checked `types.Enum` name, and the checker only builds `Enum` types from names it found in `enum_names`, so cgen never receives a raw spelling.
- transform: `enum_autostr_type_name()` no longer reads the file's imports either. The one place that really holds a source spelling is the array stringify lowering (`tokens []token.Kind` in `toml/parser/parser.v`, reached via `raw_var_type`). It now resolves that spelling through the writing file's imports (`source_enum_type_name`), but only when the binding recorded a declaration spelling that differs from its resolved type. So an inferred `values := [real_a.make()]` stays `a.Kind`.

Your project is now `test_resolved_enum_type_is_not_retargeted_by_file_import_alias`. It also checks that a parameter declared as `[]a.Kind` in the same file still formats as `b.Kind`. With `V_MACOS_V3_NO_FALLBACK=1`:

| case | before the PR | previous head | now |
|---|---|---|---|
| `value.str()`, `${value}`, `${[value]}` | from_a | from_b | from_a |
| toml plus a user `token` module | undeclared `Kind__autostr` | ok | ok |
| `[]a.Kind` / `[][]a.Kind` / `[]real_a.Kind` params | undeclared `Kind__autostr` | ok | ok |

Separately: `test_bare_name_prefers_file_selective_import_over_local_declaration` and `test_selective_import_struct_alias_wins_over_local_homonym` in the same file now fail with "cannot register struct `Token`/`Alias`, this type was already imported". That comes from master's b369e5ea4a (conflicting type declarations), with or without this commit, so those two programs are no longer valid V and need rewriting or dropping.

medvednikov:
@codex review

medvednikov:
@codex review

vlib/v/compiler_tests/selective_import_generic_arg_test.v
```diff
@@ -0,0 +1,522 @@
...
+pub struct Token {
+pub mut:
+	id int
+}
+') or { panic(err) }
+	os.write_file(os.join_path(dir, 'main.v'), 'module main
+
+import iam { Token }
+
+struct Token {
```
medvednikov: [P1] This fixture is no longer valid V: `check_type_declaration_conflict()` (b369e5e) rejects declaring `struct Token` when `Token` was already introduced by `import iam { Token }`, so the program fails in the checker before the transform/cgen path under test. The `struct Alias` fixture in `test_selective_import_struct_alias_wins_over_local_homonym` has the same problem, while both tests assert a successful compile. Please rewrite these cases so the competing homonym lives in another imported module (or drop them), rather than redeclaring the selectively imported name in the same file.

medvednikov (commented):
Reviewed the current head source changes only; CI status intentionally ignored. One blocking test-validity issue remains inline.

vlib/v/compiler_tests/selective_import_generic_arg_test.v
medvednikov: Re-checked at 115f138600. Fixed: the invalid local redeclaration fixture is gone, and the alias case now uses an imported `aaa.Alias` homonym while keeping coverage for both `Alias{}` and `take_id[Alias]`.

medvednikov (commented):
Re-reviewed the current head at `115f138600af573fc5f322c8a9dafa456f2ba181`, still ignoring CI status. The previous test-validity finding is fixed, and I did not find another source-level issue in the updated change.

Issue #28640, opened 2026-09-15
`v3 scoped monomorphize: SIGSEGV or wrong C when the memory-bounded path runs`


Jengro777:
### Describe the bug

`run_scoped_monomorphize_specs` is the only monomorphization path that bounds
memory (it releases each batch's worker arenas, needed for #28564). Exercising it
for smaller programs (lower `scoped_monomorph_node_threshold`, or a big enough
specialization batch) makes `veb` programs crash or emit wrong C. Two defects:

1. **Synthetic name collisions.** Every batch creates a fresh worker, but the
   worker's `global_temp_counter` is not continued from the master, so two
   different lifted closures get the same `__anon_fn_N` name. Signatures are
   keyed by module + name, so one closure gets the other's signature:
   `veb__Context* veb____anon_fn_5(main__Ctx* ctx) { ... return true; }` ->
   `error: incompatible types when returning type '_Bool' but 'veb__Context *'
   was expected`.
2. **Use-after-free of merged spec args.** The batch merge copies
   `[]string` tables (`generic_specialization_args`, `monomorph_cache_specs`)
   shallowly: the array is copied but the element strings still point into the
   worker arena that `transform_worker_scope_free` releases right after. A later
   pass (the driver compiles twice on this path) re-seeds from those tables and
   reads freed memory: the type arguments become NUL bytes, producing bogus
   `unknown function `ctrl.use`` / `unknown function `ctrl.route_use`` (this is
   exactly the diagnostic family of #28489) or a SIGSEGV.

### Reproduction

1. Lower the threshold that gates the bounded path (V 0.5.2 e0d4232):

```v
// vlib/v/transform/transform_parallel_notd_v3_no_parallel.v
const scoped_monomorph_node_threshold = 0   // was 1_000_000
```

2. Rebuild the compiler (`v -g -keepc -o vnew cmd/v`) and compile the small
   `veb` program from #28489 (`issue-drafts/repro/scoped-monomorph-veb/main.v`):

```v
module main

import veb

pub struct Ctx {
	veb.Context
}

pub struct ModelApp {
	veb.Middleware[Ctx]
	veb.Controller
}

pub struct Item {
	ModelApp
}

pub struct MainApp {
	veb.Middleware[Ctx]
	veb.Controller
}

fn mw() veb.MiddlewareOptions[Ctx] {
	return veb.MiddlewareOptions[Ctx]{
		handler: fn (mut ctx Ctx) bool {
			return true
		}
	}
}

[...]

PR #28700, opened 2026-09-18
v3: stop re-reading released worker args in the monomorphize merges

### LINKED ISSUE #28489: V3 reports bogus `unknown function` diagnostics for generic functions that call methods on a generic parameter, then silently falls back to V1

Jengro777:
## Why

The monomorphize worker merge copied its `[]string` tables shallowly: the array
was copied, but every element still pointed into the worker scratch arena that
`transform_worker_scope_leave()` releases right after the merge. A later pass
(the driver compiles twice on this path) re-seeds its specializations from
`generic_specialization_args` / `monomorph_cache_specs`, so it read freed type
arguments. The strings turn into NUL bytes, and V3 then reports bogus
diagnostics for perfectly valid calls:

```
error: unknown function `ctrl.use`
error: unknown function `ctrl.route_use`
```

...and silently falls back to the compatibility compiler - exactly the
diagnostic family of #28489.

## What changed

- `run_parallel_monomorphize_specs()`: deep-copy the merged
  `generic_specialization_args` entry into the parent arena instead of
  `spec_args.clone()` (which copies the slice header only).
- `run_scoped_monomorphize_specs()`: re-record every emitted specialization
  through `record_monomorph_cache_spec()` *from the master*, so the
  parent-owned copy wins over one a worker recorded for the same key.
- `record_monomorph_cache_spec()`: clone the key, declaration key, module and
  every argument string, and re-record existing keys instead of skipping them
  (the worker's copy dies with its arena).

## Test plan

`vlib/v/compiler_tests/scoped_monomorphize_closure_test.v` builds a V3 compiler
with `-prealloc` and compiles the small `veb` program from #28489 (embedded
`Middleware[Ctx]`/`Controller` structs, a generic controller helper that calls
`ctrl.use()` / `ctrl.route_use()`, and lifted closures). The fixture fails on a
shallow merge (freed arguments -> bogus diagnostics/SIGSEGV); the follow-up PR
for #28640 keeps it as the shared regression fixture for the bounded path.

```sh
v -g -keepc -o vnew cmd/v
./vnew -silent test vlib/v/compiler_tests/scoped_monomorphize_closure_test.v
```

## Notes

- Found while making the **RuoQi-v** backend
  (https://github.com/RuoQi-DoDo/RuoQi-v) compile with `-new-compiler`; the
  bogus `unknown function` diagnostics appeared there after ~100s of
  compilation followed by a silent fallback.
- Stacked PRs: #28640 and #28564 build on this branch (the bounded path has to
  be correct before it is used for every batch).

Fixes #28489.


Jengro777:
request for review

vlib/v/compiler_tests/scoped_monomorphize_closure_test.v
```diff
@@ -0,0 +1,103 @@
...
+	return os.join_path(os.temp_dir(), 'v3_scoped_monomorphize_closure_test${scoped_monomorph_bin_suffix}')
+}
+
+fn scoped_monomorph_v3_bin() string {
+	bin := scoped_monomorph_v3_bin_path()
+	if os.exists(bin) {
+		return bin
+	}
+	// `-prealloc` is what enables `scope_parallel_workers` and the scoped
+	// monomorphize path, matching how the distributed compiler is built.
```
medvednikov, line 25: `-prealloc` enables scoped worker allocation, but `run_scoped_monomorphize_specs()` is still gated by `t.a.nodes.len >= scoped_monomorph_node_threshold` (1,000,000) on this PR. This small fixture can therefore exercise the regular parallel merge and still pass while the scoped re-record/copy path remains untested. Please make the test deterministically select the scoped path (for example with a test-only threshold override) or add a focused unit test that enters `run_scoped_monomorphize_specs()`.

medvednikov (commented):
Reviewed the current code changes only; CI status intentionally not considered. The deep-copy changes themselves look coherent, but the new regression does not actually force the scoped merge path this PR also changes.

vlib/v/compiler_tests/scoped_monomorphize_closure_test.v
Jengro777: fixed

Jengro777:
fixed

PR #28701, opened 2026-09-18
v3: fix the scoped monomorphize path (closure names, context-less specs)

### LINKED ISSUE #28640: `v3 scoped monomorphize: SIGSEGV or wrong C when the memory-bounded path runs`

Jengro777:
## Why

`run_scoped_monomorphize_specs()` is the only monomorphization path that bounds
memory (it releases each batch's worker arenas). With the released-argument
issue fixed in the base branch, two defects remain:

1. **Duplicate lifted-closure names.** Every batch created a fresh worker with
   `global_temp_counter = 0`, so two different lifted closures received the same
   `__anon_fn_N` name. Signatures are keyed by module + name, so one closure got
   the other's signature and the generated C failed with:

   ```
   error: incompatible types when returning type '_Bool' but 'veb__Context *' was expected
   ```

2. **Context-less specs.** A synthesized node that hangs under no declaration
   has neither module nor file context, but its type text was still collected as
   a generic struct spec. `Middleware[Context]` (written by a file that did
   `import model { Context }`) was then rebound onto `main`:

   ```
   src.c: error: unknown type name 'main__Context'; did you mean 'veb__Context'?
   src.c: error: unknown type name '_fn_ptr_52d5bf3cc77ad60'
   ```

## What changed

- hand the master's `global_temp_counter` to every scoped batch and copy it back
  after the batch, so synthesized names keep counting across batches;
- skip context-less spellings that contain plain generic arguments
  (`type_text_has_unqualified_generic_arg`), including nested arguments
  (`Map[string, []Context]`) and the separate `generic_params()` a synthesized
  node may carry instead of a bracketed `value`;
- regression test in `vlib/v/transform/monomorphize_test.v`
  (`test_contextless_generic_struct_spec_spelling_is_skipped`).

## Test plan

```sh
v -g -keepc -o vnew cmd/v
./vnew -silent test vlib/v/transform/monomorphize_test.v
./vnew -silent test vlib/v/compiler_tests/scoped_monomorphize_closure_test.v   # from the base branch
```

Reverting the counter handoff makes the `veb` fixture fail with the
`_Bool`/`veb__Context *` error; reverting the context-less guard makes the unit
test fail at `assert 'veb.Middleware[Context]' !in specs`.

## Notes

(https://github.com/RuoQi-DoDo/RuoQi-v) compile with `-new-compiler`; RuoQi
embeds `veb.Middleware[Context]` in `model.App` and registers middleware from
`middleware/*`, which is the shape that triggers both defects.

Fixes #28640.


Jengro777:
request for review

vlib/v/transform/monomorphize.v
```diff
@@ -2711,6 +2719,85 @@ fn (mut t Transformer) collect_generic_struct_specs(decls map[string]GenericStru
...
+				payload = payload[2..].trim_space()
+				continue
+			}
+			break
+		}
+		if payload.len == 0 || payload.contains('.') || types.is_builtin_type_name(payload)
+			|| is_decimal_text(payload) {
+			continue
+		}
+		if (payload[0] >= `a` && payload[0] <= `z`) || (payload[0] >= `A` && payload[0] <= `Z`) {
```
medvednikov: `type_text_has_unqualified_generic_arg` diverges from `collect_generic_struct_spec_from_type` for container spellings. For example, `Box[map[string]int]` recurses into `map[string]int` and gets `false`, but then `payload` is still the whole `map[string]int`; since that is not a builtin *name*, this branch treats it as an unqualified type and skips a context-less spec that is actually fully resolvable. In the other direction, `Box[[2]Context]` is missed: the recursive call treats `[2]` as the bracketed portion and never inspects the trailing `Context`, while the collector explicitly strips fixed-array prefixes. That can both drop valid specs and let context-dependent specs through. Could this mirror the collector's container handling (at least `map[...]` and fixed arrays) and add nested regression cases such as `Box[map[string]int]` => false and `Box[[2]Context]` => true?

medvednikov (commented):
Reviewed the code changes only; CI status intentionally not considered. I found one correctness gap in the new context-less type-spelling guard.

vlib/v/transform/monomorphize.v
Jengro777: Fixed
Because this involves a relatively low-level issue, you can directly close the PR and use a new PR to solve it

medvednikov:
1 conflict

Jengro777:
fixed by https://github.com/vlang/v/pull/28702

PR #28702, opened 2026-09-18
v3: use the memory-bounded monomorphize path for every specialization batch

### LINKED ISSUE #28564: V3: `monomorphize` memory grows without bound until the compiler is OOM-killed

Jengro777:
## Why

`run_scoped_monomorphize_specs()` merges each batch and immediately releases the
workers' scratch arenas, which is what keeps `monomorphize` from growing without
bound. It was gated behind a 1M-node AST threshold, so ordinary projects kept
using the regular path, where every drain batch holds its workers' arenas until
the merge. Nested discoveries can turn one small batch into thousands of
specializations, and the compiler then grows until the memory guard or the OOM
killer stops the build (#28564).

Measured on the `veb` + `json2` + `orm` reproduction (RuoQi-v backend):

```
before: 19.4 GB peak RSS, killed by the 10 GiB limit
after:   7.4 GB peak RSS, build succeeds (slightly better wall time)
```

## What changed

`should_use_scoped_monomorphize(node_count, spec_count)` now selects the bounded
path for every non-empty batch (the cutoff is deliberately one specialization,
documented next to the constant), and a boundary test pins that decision.

This is only safe on top of the two stacked branches: the bounded path must stop
reading released worker arguments (#28489) and must not reuse lifted-closure
names or collect context-less specs (#28640) before it is enabled for every
batch.

## Test plan

```sh
v -g -keepc -o vnew cmd/v
./vnew -silent test vlib/v/transform/monomorphize_test.v
./vnew -silent test vlib/v/compiler_tests/scoped_monomorphize_closure_test.v
./vnew -silent test vlib/v/compiler_tests/selective_import_generic_arg_test.v
```

End-to-end, with the compiler built from this branch:

```sh
git clone https://github.com/RuoQi-DoDo/RuoQi-v
cd RuoQi-v/backend && /path/to/vnew -new-compiler ./main   # exit 0, ~2m13s, 20.7 MB
```

## Notes

The memory guard itself (`-no-memory-limit`, `vlib/v/bench`) is untouched: this
PR makes the compiler actually stay below it instead of disabling it.

Fixes #28564.


Jengro777:
request for review

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I did not find a remaining concrete correctness issue. The emitted specialization keys/args are copied before the worker scope is left, synthesized-name state is carried across scoped batches, the contextless-type guard now handles map/fixed-array containers correctly, and the selector change makes every non-empty eligible monomorphization batch use the bounded path as intended.

PR #28719, opened 2026-09-18
tests: add GUI codegen probes for #28694 (fix pending)

### LINKED ISSUE #28694: v3 fails to compile gui showcase.v example
### Describe the bug

The gui examples take extremely long to compile (over a minute), due to falling back to 0.5.2
The example does build with the fallback compiler but fails with the new compiler.

### Reproduction Steps

```
v install gui
cd ~/.vmodules/gui
v run examples/showcase.v
```

### Expected Behavior

Should compile with v3.

### Current Behavior

Falls back to v0.5.2

### Possible Solution

_No response_

_No response_

### V version

V 0.5.2 aca60a4

|V full version      |V 0.5.2 aca60a4
|Processor           |12 cpus, 64bit, little endian, 13th Gen Intel(R) Core(TM) i5-1335U
|Memory              |4.1GB/15.3GB
|                    |
|V executable        |/home/dylan/Repos/v/v
|V last modified time|2026-09-17 20:41:30
|                    |
|V home dir          |OK, value: /home/dylan/Repos/v
|VMODULES            |OK, value: /home/dylan/.vmodules
|Current working dir |OK, value: /home/dylan/.vmodules/gui
|                    |
|V git status        |0.5.2-791-gaca60a48
|.git/config present |true
|                    |
|emcc version        |N/A

> [!NOTE]
>

medvednikov:
## Status

**Work in progress: regression probes only. This PR does not fix #28694 and is not ready to merge.**

Refs #28694. The requested compiler fix has not been isolated. These cases preserve the reported failure patterns for further diagnosis; they have not been confirmed as minimal reproducers of the full GUI failure.

## Changes

Add `vlib/v/compiler_tests/issue_28694_gui_codegen_test.v` with three independent compile-and-run probes:

- `net.urllib` URL parsing and `URL.set_path` success/error paths, exercising the `!bool` wrapper implicated by the missing `Optional_bool` declaration.
- A function-initialized fixed-array constant in an imported module, accessed through a runtime index, representing the missing `vglyph__gamma_table` declaration pattern.
- Function-pointer parameters forwarded into a header-defined C callback struct. The header uses `const char *` and C `int`; negative callback arguments check that any adapter preserves signed values rather than merely making the assignment compile.

Each subprocess explicitly uses `-new-compiler -nocache`, so fallback cannot hide a compiler failure. The probes create their own temporary fixtures, need no GUI/windowing dependencies, and disable automatic compiler-error uploads while running.

There are no compiler implementation changes, CI changes, or changes to existing tests.

## Investigation

The issue's diagnostics include missing declarations and callback ABI mismatches. `URL.set_path` genuinely returns `!bool`. Current codegen already collects selected function signatures and has C callback adaptation paths; source inspection alone did not establish which path fails in the full GUI build. No speculative typedef, pointer cast, or warning suppression is included.

## Validation

- Passed: the embedded C callback header, exercised by a standalone C harness, compiled and ran with `cc -std=c11 -Wall -Wextra -Werror`.
- Passed: local trailing-whitespace and final-newline checks on the new test file.
- **Not run:** the V probes, `./v fmt`, compiler rebuild, codegen tests, or the actual GUI showcase. The execution environment has no V compiler, and both repository cloning and archive downloads were blocked. GitHub source reads and branch writes were available.
- No passing V test result or confirmed baseline reproduction is claimed.

## Required before this is ready

1. Format and run the probes with a working V checkout; establish which reproduce on the affected revision and current master.
2. Reproduce `gui/examples/showcase.v` with `-new-compiler`, including comparison of serial/parallel and cached/uncached generation as needed; implement the compiler correction and strengthen the reproducer accordingly.
3. Rebuild and run the focused compiler tests and the full showcase, then replace this draft status with actual fix and validation results.

Focused command:

```sh
V_C_ERROR_BUG_REPORT_DISABLED=1 ./v -new-compiler test vlib/v/compiler_tests/issue_28694_gui_codegen_test.v
```


vlib/v/compiler_tests/issue_28694_gui_codegen_test.v
```diff
@@ -0,0 +1,161 @@
...
+	temp_dir := os.join_path(os.temp_dir(), 'v_issue_28694_${name}_${os.getpid()}')
+	os.rmdir_all(temp_dir) or {}
+	os.mkdir_all(temp_dir) or { panic(err) }
+	defer {
+		os.rmdir_all(temp_dir) or {}
+	}
+	old_report_disabled := os.getenv('V_C_ERROR_BUG_REPORT_DISABLED')
+	os.setenv('V_C_ERROR_BUG_REPORT_DISABLED', '1', true)
+	defer {
+		os.setenv('V_C_ERROR_BUG_REPORT_DISABLED', old_report_disabled, true)
```
medvednikov, line 16: If `V_C_ERROR_BUG_REPORT_DISABLED` was originally unset, this defer restores it as a *set-but-empty* variable instead of removing it. Since this helper is used by multiple tests in the same process, that leaks ambient state into later tests (and potentially changes any code that checks presence rather than value). Please record `name in os.environ()` before setting it and `unsetenv` on restore when it was absent, like the other compiler-test environment helpers do.

medvednikov (commented):
Reviewed the draft test harness only; CI status intentionally not considered. I found one test-isolation issue.

PR #28722, opened 2026-09-18
cgen: unwrap shared map fields before map runtime calls

medvednikov:
## Problem

After #28466 reduced `map` to a pointer-sized header, direct access to a `shared map` struct field can pass the shared-field wrapper to builtin map helpers instead of the wrapper's `.val` payload.

That corrupts the map header interpretation. One macOS clean-code run at `8670fb6` panicked in:

```
VMapData__key_to_index
VMapData__set
map__set
testing__NormalReporter__report
```

with `map.hash_fn is nil`; the reported map fields contained pointer-like values, matching the shared wrapper being read as `map.data`.

`NormalReporter.report` triggers the bug through direct operations such as:

```v
lock r.compiling {
    r.compiling[message.file] = message
}
```

## Fix

Add one Cgen helper for map value expressions. It unwraps direct shared struct fields through the existing shared-field selector machinery, so map runtime helpers receive `field->val`; lock lowering continues to use the wrapper itself.

Use it for:
- map index lvalues
- map `or` lookups and if guards
- map reference arguments used by map methods

## Regression coverage

- `lock_codegen_test.v` checks that direct shared-map field indexing emits `map__get_or_set(&s->values->val, ...)` and `map__get_check(&s->values->val, ...)`, never the wrapper-address forms.
- `shared_elem_test.v` now locks a shared map field directly, writes an entry through it, and reads the entry back through the same field.


medvednikov (commented):
Reviewed the code changes only; CI status intentionally not considered. I did not find a concrete correctness issue. The new helper is narrowly applied at map-runtime call sites, preserves the shared wrapper for lock lowering, and routes direct shared-field selectors through the existing shared-field storage/value machinery before adding `.val`.

PR #28726, opened 2026-09-18
ci: preserve failures in BSD Tools CI shell steps

medvednikov:
## Summary

Publish the prepared BSD Tools CI failure-propagation patch, based on master `d701a1b9924f274d7101d2dc60409e7015bef42e`.

Fix two interacting shell defects across all five FreeBSD/OpenBSD compiler configurations:

- Enable `set -eu` in both custom VM shell blocks so failed installation, setup, formatting, and test commands cannot be hidden by successful logging commands afterward.
- Split `gmake && ./v -showcc -o v cmd/v && sudo ./v symlink` into three separately checked commands. Adding `-e` alone does not fix the first two commands in that AND-list.
- Add compiler-independent regression tests and an independent, read-only Linux job that runs them without waiting for a V bootstrap or BSD VM.

All existing compiler matrices, package installation, build, formatting, doctor, normal tests, conditional strict tests, VM configuration, workflow triggers, and concurrency are preserved. No tests, assertions, or integrity checks are disabled.

## Regression coverage

`python3 cmd/tools/tools_ci_test.py` executes the actual workflow shell strings with controlled command fixtures, rather than duplicating their shell control flow in Python. Each complete suite exercises **264 shell executions** across `sh` and Bash, with and without externally supplied `-e`: 20 successful workload traces and 244 injected failures across the five compiler configurations.

Every failure must preserve its originating nonzero exit code and stop all subsequent workload commands. Every successful execution must retain the complete existing command sequence. PATH is restricted to the fixtures; no package manager, real compiler, hostname change, or real git configuration command is executed. The shell's ulimit builtin is also replaced.

## Validation

The complete suite was rerun immediately before publication: **all three unittest methods pass**, covering all 264 shell executions. Python byte-compilation also passes.

During patch preparation, the same suite distinguished the complete fix from both incomplete fixes:

| Implementation | Failed regression assertions |
| --- | ---: |
| Unmodified workflow | 142 |
| Only add `set -eu` | 40 |
| Only split the bootstrap chain | 122 |
| Complete patch | 0 |

Before publication, revalidated:

- Current-master workflow blob is still `97aa83ff170d596a307baa74529229dbcb2cd7d0`, matching the patch baseline.
- Clean reverse/apply round trip and `git apply --check` against that verified baseline.
- YAML comparison confirms that existing jobs and workflow settings are unchanged except for the two intended BSD shell strings; the shell-contracts job is the only new job.
- `git diff --cached --check` passes.
- Both uploaded blobs exactly match the locally tested files: workflow `f85dffaaf7b04d761bd7d0495c036262e1576b8d`; regression test `263b7efadc097b916493ff1bd9d49562a4cdbb90`.

**Actual BSD VM builds and V compiler tests have not been run locally.** These tests validate workflow shell failure handling with controlled executables, not BSD compiler/runtime correctness. GitHub Actions results are pending at publication. This patch prevents false-success CI reporting; it does not claim to resolve the separate compiler crashes or Windows blockers.

## Files

- `.github/workflows/tools_ci.yml`
- `cmd/tools/tools_ci_test.py`


medvednikov (commented):
Reviewed the code changes only; CI status intentionally not considered. I did not find a concrete correctness issue. `set -eu` now applies inside the custom BSD VM shell blocks, the bootstrap commands are separated so `errexit` can observe each failure, and the regression harness executes the literal workflow scripts with controlled failures rather than reimplementing their control flow.

PR #28727, opened 2026-09-18
cgen: keep shared wrapper payloads visible to GC

medvednikov:
## Problem

Follow-up to #28722 and #28466.

After #28466 made `map` pointer-sized, a shared-field wrapper no longer contains the map header itself; its `.val` contains a `VMapData*`.

The generated `__dup__shared__...` helper still allocates the wrapper with raw libc `malloc`:

```c
shared_wrapper *dest = (shared_wrapper *)malloc(...);
```

That block is outside Boehm's managed/scanned heap. The only reference to an empty shared map's freshly allocated `VMapData` can therefore live inside memory the collector does not scan.

This showed up in the macOS clean-code run after #28722. The earlier bad-wrapper-address crash was gone, but `NormalReporter.report` then failed on the first backing allocation:

```
malloc(0xaafbb17b836ba178 < 0)
V panic: memory allocation failure
...
new_dense_array
VMapData__set
map__set
testing__NormalReporter__report
```

The requested allocation divided by the initial dense-array capacity (8) is
`0x155f762f706d742f`; in little endian its first bytes are `/tmp/v_`.
That is consistent with the `VMapData` header having been collected and its
storage subsequently reused for a temporary-path string.

## Fix

Keep the existing direct-`malloc` path for `-prealloc`. macOS already requires shared rwmutex wrappers to use direct C allocation there.

For non-prealloc builds, allocate shared wrappers with
`malloc_uncollectable` instead. Under Boehm this preserves the old
non-reclaimable wrapper lifetime while making the wrapper contents scannable,
so nested pointers such as `map.data` remain GC roots.

Because the call is synthesized by Cgen, seed `malloc_uncollectable` from
markused whenever the shared runtime is needed.

## Regression coverage

- Cgen test verifies shared dup helpers keep the raw-`malloc` prealloc arm and
  use `malloc_uncollectable` otherwise.
- Runtime shared-map test constructs an empty shared map field in a noinline
  helper, forces a collection before first map access, churns similarly sized
  path strings, then writes and reads the map field.

This specifically covers the lifetime hole that #28722's direct shared-map
access regression did not exercise.


medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I did not find a concrete correctness issue. `malloc_uncollectable` has the needed Boehm semantics here: the wrapper itself is not reclaimed while its contents remain scanned, so pointer-sized payloads such as `map.data` stay roots. The `_VPREALLOC` arm preserves the existing raw allocation path, and markused seeds the synthesized helper so non-prealloc generated C has the allocator available.

PR #28728, opened 2026-09-18
compiler: preserve interface auto-str in reflected methods

### LINKED ISSUE #34: Broken behaviour with loading interfaces on button clicks
main.v:
```
import ui2

fn main() {
	mut m := Model{}

	ui2.run_vml[Model](
		source: $embed_file('app.vml').to_string()
		model:  m
		title:  'CsvConv'
		width:  128
		height: 64
	) or {
		eprintln('Error loading UI: ${err.msg()}')
	}
}

pub struct Model {}

interface IFoo {
	name string
}

struct Foo {
	name string
}

pub fn (m Model) get_foo() IFoo {
	return Foo{
		'you should see this'
	}
}

pub fn (mut m Model) run() {
	mut f := m.get_foo()
	println(f)
}
```
app.vml:
```
Screen {
	id: root
	Button {
		id: run_button
		text: "Click Me"
		on_tap: app.run()
	}
}
```

Running the above code will print "unknown interface value" every time the button is clicked, however manually calling `m.run()` once before constructing the UI will result in all button presses successfully printing the Foo instance.

medvednikov:
Fixes vlang/ui2#34.\n\nPreserve all known interface implementations when auto-stringifying a comptime-reflected method body that has no boxed-type snapshot, so runtime VML callbacks print the concrete interface value.\n\nTests:\n- ./v self\n- ./v -silent vlib/v/tests/interfaces/comptime_method_dispatch_interface_return_test.v\n- VEXE=/Users/alex/code/v-issue-ui2-34/v VFLAGS=-nocache VTEST_ONLY=comptime_method_dispatch_interface_auto_str ./v -silent vlib/v/slow_tests/inout/compiler_test.v


medvednikov:
@codex review

medvednikov (commented):
Reviewed the current head code changes only; CI status intentionally not considered. I did not find a concrete correctness issue. The fallback is limited to the case where both boxed-type indexes are empty; once either source or late boxing information exists, the existing filtering remains in force. Open generic implementations are still excluded, and the regression covers both ordinary reflected dispatch and the auto-string path that previously rendered a live interface as `unknown`.

squidink7:
In relation to how this affects https://github.com/vlang/ui2/issues/34, I noticed it doesn't just affect printing the interface but also calling any methods on it no longer worked and caused a panic so its not just isolated to str()

PR #28741, opened 2026-09-18
cmdexec: pass PeekNamedPipe its byte count as voidptr, fixing gcc builds on Windows

### LINKED ISSUE #28731: cmdexec: PeekNamedPipe call fails to compile with gcc on Windows (incompatible pointer type, since #28708)
**Describe the bug**

Since 4c91adf1b4 (cmdexec: keep timeout active after the process leader exits, #28708, 2026-09-18) every program that links `v.cmdexec` fails to C-compile on Windows with gcc — which includes `v self -cc gcc`, `v -cc gcc test ...` and any tool that imports `v.cmdexec`.

`vlib/v/cmdexec/cmdexec_pipe.c.v`:

```v
mut available := u32(0)
if !C.PeekNamedPipe(handle, 0, 0, 0, &available, 0) {
```

`&available` is emitted as `unsigned int*`, but `PeekNamedPipe`'s fifth parameter is `LPDWORD` (`unsigned long*`). The two are distinct pointer types even though both are 32-bit on Windows, and gcc 14+ made `-Wincompatible-pointer-types` an error by default. tcc accepts it, which is why the default build and the tcc CI still pass.

**Reproduction Steps**

On Windows with MinGW gcc 14 or newer (here: `gcc.exe (MinGW-W64 x86_64-ucrt-posix-seh, built by Brecht Sanders, r1) 16.2.0`):

```
v -cc gcc vlib/v/driver/environment_test.v
```

or any program importing `v.cmdexec`.

**Expected Behavior**

Compiles.

**Current Behavior**

```
src.c: In function 'cmdexec__read_process_pipe':
src.c:127566:53: error: passing argument 5 of 'PeekNamedPipe' from incompatible pointer type [-Wincompatible-pointer-types]
127566 |                 if (!PeekNamedPipe(handle, 0, 0, 0, &available, 0)) {
```

**Possible Solution**

Pass `voidptr(&available)` (the declaration in `cmdexec_pipe.c.v` already types that parameter as `voidptr`), the way `os.fd_is_pending` in `vlib/os/fd.c.v` does for the same call.

**Additional Information/Context**

Master `01a8af98dc`. Likely the cause of the currently red `CI Windows GCC / gcc-windows` job on master.

**V version**
V 0.5.2 01a8af9

**Environment details**
Windows 11 Pro 10.0.26200, MinGW-w64 gcc 16.2.0

quaesitor-scientiam:
Fixes #28731.

Since 4c91adf1b4 (#28708), `read_process_pipe` passes `&available` (a `u32*`) as `PeekNamedPipe`'s `lpTotalBytesAvail`, whose type is `LPDWORD` (`unsigned long*`). gcc 14+ makes `-Wincompatible-pointer-types` an error, so `v -cc gcc` fails to C-compile anything that links `v.cmdexec` on Windows (`v self -cc gcc`, `v -cc gcc test ...`, every tool). tcc accepts the mismatch, which is why default builds still work.

Cast the argument to `voidptr`, as `os.fd_is_pending` already does for the same call.

Checked on Windows 11 with MinGW-w64 gcc 16.2.0: `v -cc gcc -o vgcc.exe cmd/v` and `v -cc gcc vlib/v/driver/environment_test.v` now compile (both failed before); `v test vlib/v/cmdexec/` passes with tcc. With `-cc gcc`, `cmdexec_resolve_program_test.v` fails here with `STATUS_DLL_NOT_FOUND` (0xC0000135) because the test runs a freshly built child executable without the gcc runtime DLL directory on PATH — unrelated to this change and present before it.


medvednikov (commented):
Reviewed commit `71c895aa1a0d200eda14dbbdde87e019c97d5cf5`. No actionable correctness findings in this change.

The `voidptr(&available)` cast in `vlib/v/cmdexec/cmdexec_pipe.c.v` is consistent with the local FFI declaration and the existing `os.fd_is_pending` call in `vlib/os/fd.c.v`. [PeekNamedPipe's fifth parameter is `LPDWORD`](https://learn.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-peeknamedpipe), and [`DWORD` remains 32-bit in both 32-bit and 64-bit Windows applications](https://learn.microsoft.com/en-us/windows/win32/learnwin32/windows-coding-conventions#integer-types), so the existing `u32` storage has the required size. Casting its address through `voidptr` addresses the C pointer-type diagnostic described in [GCC 14's porting guidance](https://gcc.gnu.org/gcc-14/porting_to.html#incompatible-pointer-types), without changing the storage lifetime, empty-pipe/EOF handling, or non-Windows branch.

Validation: inspected the complete diff, surrounding function, and existing `os` call, and checked the API/type documentation. CI was excluded from this review. This was a static review; I did not run a Windows/MinGW build or runtime tests.

PR #28752, opened 2026-09-18
math: fix pow exponent bounds for platform-width int

medvednikov:
## Fix

The pure-V `pow` implementation builds its lower exponent bound with:

```v
int(u32(u32(-1) << 12))
```

This relied on a 32-bit signed `int`: the `u32` bit pattern is `0xfffff000`, which became `-4096`. With a 64-bit platform-width `int`, the same conversion instead preserves the positive value `4294963200`. Ordinary `frexp` exponents therefore satisfy the supposed underflow guard, and the successive-squaring loop exits before applying the integer part of a fractional power.

Use `-(1 << 12)` instead. This produces the intended negative bound on both 32-bit and 64-bit targets without shifting a negative value. The upper bound, power algorithm, and existing test tolerances are unchanged.

## Regression coverage

`vlib/math/pow_fractional_test.v` adds:

- 12 finite cases with exactly representable expected values: bases above/below one, positive/negative fractional powers, both fractional range-reduction branches, and repeated squaring.
- 4 overflow/underflow cases exercising both exponent bounds and reciprocal results.

The final diff contains only `vlib/math/pow.v` and `vlib/math/pow_fractional_test.v`. The temporary branch-only diagnostic workflow has been removed.

## Validation status

**Draft pending V-level validation.**

A standalone C model of the fractional-power path was compiled and run locally with both Clang and GCC (`-std=c11 -Wall -Wextra -Werror -O2`). It confirms the old 32-bit/64-bit bounds are `-4096`/`4294963200`, reproduces `pow(16, 1.25) = 64` with the old 64-bit bound, and returns the expected `32` with the portable bound. All 26 numerical model cases passed with each compiler. This is an arithmetic/representation check, **not** a run of V's compiler or test suite.

No V compiler is installed in the local environment, and GitHub downloads/cloning failed because of DNS access. `v fmt` and the V tests have **not** been run locally. GitHub Actions validation was queued when this description was updated.

Commands still requiring V-level validation:

```sh
./v fmt -verify vlib/math/pow.v vlib/math/pow_fractional_test.v
./v -cc clang -exclude '@vlib/math/*.c.v' vlib/math/pow_fractional_test.v
./v -cc clang -exclude '@vlib/math/*.c.v' vlib/math/math_test.v
./v -cc clang -exclude '@vlib/math/*.c.v' vlib/math/complex/complex_test.v
./v -cc clang -exclude '@vlib/math/*.c.v' test vlib/math/
```

This fixes the width-dependent bound behind the reported `test_pow` assertion. The separate `complex_test.v` failure from the report has not yet been independently reproduced or confirmed resolved.


medvednikov (commented):
Reviewed `7533017905664203d32a73053be3ddf7ec19f2aa`, with CI excluded from the assessment. **No blocking findings in this diff.**

The change in `vlib/math/pow.v` restores the intended lower bound: `-(1 << 12)` evaluates to `-4096` with either a 32-bit or 64-bit signed `int`, without relying on the conversion of the `u32` bit pattern `0xfffff000`. It shifts a positive operand, preserves the strict comparisons at both bounds, and leaves the exponent-accumulation and reciprocal logic unchanged.

The regression cases are well targeted: the 12 finite cases avoid the `±0.5` shortcuts, cover both sides of the fractional range reduction, both signs of the `frexp` exponent, negative powers, and repeated squaring. The four extreme cases also reach both guard branches and exercise reciprocal overflow/underflow. `close` is defined in the regular `math.v` module, so the new test file does not depend on helpers from another `_test.v` file.

I independently ran a Python arithmetic model of the fractional-power path, using Python's math primitives and explicitly modeling the old 32-bit and 64-bit bound conversions. All 16 added expectations matched with the corrected bound (and with the old 32-bit bound); all 16 failed with the old 64-bit bound. In particular, `pow(16, 1.25)` changed from `64` to `32`. The extreme cases hit the upper guard at `xe = 4097` and the lower guard at `xe = -8191`.

**Validation limitation:** this was source review plus an independent arithmetic model, not execution of V's compiler or math implementation. No V compiler is installed here, and a direct clone failed because of DNS resolution, so I did not run the V tests or `v fmt`. For V-level validation, retain the documented `-exclude '@vlib/math/*.c.v'` invocation so the regression exercises the changed pure-V implementation rather than the `C.pow` wrapper.

PR #28758, opened 2026-09-18
cgen: fix function-field calls on pointer-backed mutable receivers

medvednikov:
## Summary

Fix the compiler rather than changing valid BloomFilter code. The original `b.hash_func(element)` call in `vlib/datatypes/bloom_filter.v` is restored byte-for-byte; that library implementation has no net diff in this PR.

In `FlatGen.gen_fn_field_call`, emit the existing callee expression through ordinary selector generation instead of manually reconstructing field access from `base_type`. The manual path checked only `base_type is types.Pointer`, losing the lowered selector's pointer-access information for mutable generic receivers whose V type is a struct but whose C storage is a pointer. This produced the reported invalid `b.hash_func(element)` in C.

The compiler change reuses:

```v
g.gen_expr(g.a.child(&node, 0))
```

It removes the duplicate field-selection logic while retaining callback argument conversion and function-pointer-field indirection. There are no library API changes and no documentation changes.

## Regression coverage added

- `vlib/v/gen/c/fn_field_call_test.v`: directly exercise the C generator and assert arrow access for a pointer-backed value receiver, dot access for value storage, explicit-pointer access, and function-pointer-field dereferencing. These assertions inspect the compiler under test, so compiling the test harness through V1 cannot hide a regression in its generated output.
- `vlib/v/tests/generics/generic_fn_field_mut_receiver_test.v`: direct callback calls on mutable generic receivers and parameters, value and explicit-pointer receivers, and nested receivers; integer and string specializations with observable state updates.
- `vlib/datatypes/bloom_filter_test.v`: retain the integer-callback tests for both constructors, including repeated insertion and the callback-selected bit. The existing string tests remain the original reproducer.

## Validation

- Verified the published PR diff: four files, with the compiler edit limited to `gen_fn_field_call` (2 added lines, 16 removed lines); no net change to `bloom_filter.v`.
- Patch application and `git diff --check` passed against local fixtures containing the retrieved source hunks plus the complete new test files. This is patch/whitespace validation, not a full repository build.
- **Compiler rebuild, V tests, and V formatting have not been run.** This environment has no V compiler and cannot clone/bootstrap from GitHub because outbound GitHub DNS/network access is unavailable. Changes were published through the connected GitHub API.

This remains a draft pending actual compiler validation. Suggested commands from the repository root:

```sh
./v self
./v fmt -w vlib/v/gen/c/fn.v vlib/v/gen/c/fn_field_call_test.v vlib/v/tests/generics/generic_fn_field_mut_receiver_test.v vlib/datatypes/bloom_filter_test.v
V_C_ERROR_BUG_REPORT_DISABLED=1 ./v -cc clang vlib/datatypes/bloom_filter_test.v
V_C_ERROR_BUG_REPORT_DISABLED=1 ./v vlib/v/tests/generics/generic_fn_field_mut_receiver_test.v
V_C_ERROR_BUG_REPORT_DISABLED=1 ./v test vlib/v/gen/c/
V_C_ERROR_BUG_REPORT_DISABLED=1 ./v test vlib/datatypes/
V_C_ERROR_BUG_REPORT_DISABLED=1 ./v -silent vlib/v/compiler_errors_test.v
V_C_ERROR_BUG_REPORT_DISABLED=1 ./v test vlib/v/
```

The original reproducer must succeed with the default compiler itself. A successful V1 fallback is not validation of the compiler fix. The earlier library-workaround commit is superseded by the compiler-fix commit; it is retained only in the PR history.


medvednikov (commented):
Reviewed at `48dc651fd147423a9be0327ea4d33e4e2ad3342b`.

No actionable correctness issue found in the four-file diff. Emitting the actual callee selector removes the duplicated dot/arrow decision that ignored lowered pointer-storage information. The surrounding `gen_fn_field_call` still resolves the callback signature, retains function-pointer-field dereferencing, and converts each argument against the expected callback parameter type; those behaviors are not replaced by a generic call-emission path.

The direct generator tests are particularly useful here: they distinguish a value-typed receiver stored as a pointer from ordinary value storage, and separately check explicit pointers and pointer-to-function fields. The runtime cases add mutable generic receivers/parameters, nested receivers, and two concrete argument types with observable state updates. The BloomFilter tests retain the library call as a consumer-level regression rather than hiding it behind a library workaround.

Validation: complete patch and surrounding function-field call-emission source review, including all added tests. No V compiler is available locally, so I did not rebuild or execute these tests. CI status and results were not considered.

PR #28763, opened 2026-09-19
checker: fix required field init of embedded struct literals

### LINKED ISSUE #28762: v3: false "field ... must be initialized" for @[required] fields of an initialized embedded struct
### Describe the bug

V3 (`-new-compiler`) reports `field ... must be initialized` for `@[required]`
fields of an embedded struct, even though those fields *are* initialized by the
explicit embedded struct literal. V1 accepts the exact same program, so this is
a V3/V1 divergence, not the intended `@[required]` rule.

Only the *embedded* case is affected. Initializing the fields as promoted fields
(`Derived{ a: 1 }`) or using a named non-embedded field (`base Base`) works in
both compilers.

### Reproduction Steps

`main.v`:

```v
struct Base {
pub:
	a int @[required]
}

struct Derived {
	Base
}

fn main() {
	d := Derived{
		Base: Base{
			a: 1
		}
	}
	println(d.a)
}
```

Run:

```sh
v -new-compiler run main.v   # fails
v -old-compiler run main.v   # compiles, prints 1
```

### Expected Behavior

Compiles with `-new-compiler`, exactly like V1 does. `Base.a` is initialized by
the nested literal `Base{ a: 1 }`.

### Current Behavior

```
main.v:11:7: error: field `Base.a` must be initialized
    9 |
   10 | fn main() {
   11 |     d := Derived{
      |          ~~~~~~~~
   12 |         Base: Base{
   13 |             a: 1
main.v:11:7: error: field `Derived.a` must be initialized
    9 |
   10 | fn main() {
   11 |     d := Derived{
      |          ~~~~~~~~
   12 |         Base: Base{
   13 |             a: 1
```

Note the duplicate diagnostics: the same field is reported once with its owning
type (`Base.a`) and once as a promoted field (`Derived.a`).

With the default `v`, V3 fails and then silently retries with the compatibility
compiler (exit 0), which masks the regression:

```
Compiler output from the default V compiler:
main.v:11:7: error: field `Base.a` must be initialized
...
V compilation failed (compiler_error); retrying with `/home/<user>/.cache/v/v1-fallback/0.5.2/v`.
```

Other variants of the same root cause:

```v
// 1) embedded value from a variable: V3 errors, V1 accepts
base := Base{
	a: 1
}
d := Derived{
	Base: base
}

// 2) promoted-field init: both compilers accept (workaround)
d := Derived{
	a: 1
}

// 3) named, non-embedded field: both compilers accept
struct Derived {
pub:
	base Base
}
d := Derived{
	base: Base{
		a: 1
	}
}

// 4) the nested literal itself is still checked correctly
//    (Base.b is genuinely missing here)
d := Derived{
	Base: Base{
		a: 1
	}
}
// V1:      error: field `Base.b` must be initialized   (at the inner literal)
// V3:      error: field `Base.b` must be initialized   (at the inner literal)
//      +   error: field `Base.a` / `Base.b` / `Derived.a` / `Derived.b` must be initialized
//                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                                              false positives at the outer literal
```

### Possible Solution

`vlib/v/types/checker_tail_stmt.v`, `missing_required_struct_fields()`:

```v
child_supplied := if is_embed {
	supplied
} else {
	map[string]bool{}
}
child_missing := tc.missing_required_struct_fields(field_type.name(), child_supplied, next_path)
```

For embedded fields the parent's `supplied` map is passed down, so required
fields that are only provided *inside the explicitly initialized embedded
literal* are never counted as initialized.

V1 has the opposite (intended) rule, in `vlib/v/checker/struct.v`:

```v
// all the fields of initialized embedded struct are ignored, they are considered initialized
sym := c.table.sym(init_field.typ)
if init_field.is_embed && sym.kind == .struct && sym.language == .v {
	struct_fields := c.table.struct_fields(sym)
	for struct_field in struct_fields {
		inited_fields << struct_field.name
	}
}
```

Suggested fix: when an embed field is explicitly present in the outer literal
(`is_embed && field.value in supplied`), treat the embedded struct's fields as
supplied, and let the nested literal's own check report genuinely missing
required fields (V3 already does that correctly — see variant 4). The flattened
required-field check in `vlib/v/gen/fastc/validate.v` should be verified as well
for the `-b fastc` path.

- Same behavior with `v -check`, `v run` and a full build; no dependencies
  needed, the snippet above is self-contained.
- Found while porting a real project (RuoQi-v) to `-new-compiler`; the current
  workaround in the project is to initialize the promoted fields directly
  instead of using `Embed: Embed{...}`.

### V version

V 0.5.2 0a006c2

Deepin 25 (crimson), Linux 6.18.48-amd64-desktop-rolling x86_64, gcc (Deepin
12.3.0-17deepin18) 12.3.0, V 0.5.2 0a006c2.

> [!NOTE]
>

Jengro777:
Fixes #28762.

## Problem

V3 reports false `field ... must be initialized` errors when a required field
belongs to an embedded struct that is explicitly initialized as a whole:

```v
struct Base {
pub:
	a int @[required]
}

struct Derived {
	Base
}

fn main() {
	d := Derived{
		Base: Base{
			a: 1
		}
	}
	println(d.a)
}
```

V3 emitted `field Base.a must be initialized` and
`field Derived.a must be initialized`, while V1 accepts the program.

The flattened check in `missing_required_struct_fields()` recursed into the
embedded struct with the parent's `supplied` map, so the outer literal never
saw the fields supplied by the nested literal. The same root cause affected
reference fields through `missing_reference_struct_fields()`.

## Fix

In both flattened checks, a field that is explicitly initialized at the current
literal is skipped: its own initializer is checked separately, and an explicitly
initialized embedded struct literal reports its own missing required/reference
fields once, at the nested literal.

Promoted-field initialization (`Derived{ a: 1 }`) keeps the existing behavior,
including the same diagnostics as V1.

## Tests

- `vlib/v/tests/structs/struct_embed_required_field_init_test.v` covers explicit
  embedded struct literals, embedded values from variables, promoted fields, and
  reference fields.
- `vlib/v/checker/tests/struct_embed_required_field_literal_err.vv` and its
  `.out` pin the remaining single diagnostic at the nested literal when a
  required field is genuinely missing.
- Rebuilt with `./v self`, then ran:
  - `./v vlib/v/tests/structs/struct_embed_required_field_init_test.v` - pass
  - `./v -silent vlib/v/compiler_errors_test.v` - no new failures; only
    pre-existing `-skip-unused` warning-output mismatches under
    `vlib/v/tests/skip_unused/`
  - `./v -silent test vlib/v/tests/structs/` - new test passes; remaining
    failures are pre-existing environment/host failures
  - `./v -silent vlib/v/slow_tests/inout/compiler_test.v` - 91 files:
    81 ok, 9 ok panics, 1 skipped, 0 errors

## Notes

The fastc flattened required-field check was reviewed: it only inspects direct
struct fields, so it does not produce the duplicate outer-literal diagnostics
and needs no change.


Jengro777:
request review

medvednikov (commented):
### [P2] Scope supplied embed keys to the field path before skipping recursion

`vlib/v/types/checker_tail_stmt.v:4675` — the same issue applies to the new guard in `missing_reference_struct_fields()` at line 4619.

`supplied` is passed unchanged when descending through an omitted embedded struct, so membership in this map does not necessarily mean that the current embedded field was explicitly initialized. A direct embed can shadow a same-named embed further down a different path:

```v
struct Inner {}

struct Base {
    n int @[required]
    p &Inner
}

struct Middle {
    Base
}

struct Outer {
    Base
    Middle
}

fn main() {
    value := Outer{
        Base: Base{
            n: 1
            p: &Inner{}
        }
    }
    println(value.Middle.n)
}
```

For the outer literal, `supplied` contains only `Base`. Skipping `Outer.Base` is correct, but the checker then descends into the omitted `Middle` with that same map. The new condition also skips `Middle.Base`, causing both helpers to return no missing fields for that subobject. There is no `Middle{...}` initializer to perform the separate validation described in the comment. The explicit `Base: ...` initializer targets the direct `Outer.Base`, not `Outer.Middle.Base`; the latter's required integer and reference are still omitted.

Please distinguish the resolved embedding path/owner of supplied fields instead of treating a matching name anywhere in the recursive traversal as initialization of the current embed. Add a compile-fail regression for a direct and a nested embed sharing the same name, with the nested one omitted, covering both required and reference fields. This should retain the intended single-diagnostic behavior for genuinely supplied embeds without suppressing checks on another subobject.

Validation: source-level review of `e8fff21f9eddfa473747b0f3c45b4b19da7412d5`, including the caller and C struct-initialization path. This V example was not executed locally; no V compiler was installed and the attempted checkout failed because the environment could not resolve GitHub.

Jengro777:
fixed

Jengro777:
request review

medvednikov (commented):
Re-reviewed at `ae880ea66ca275047255aa3f8239830a49c7322f`, without considering CI.

The direct-versus-nested embedding case in my previous P2 finding is addressed. In `Outer{Base: ...}`, `supplied_fields_for_embeds()` removes the direct `Base` key from the set forwarded into the omitted `Middle`, so both missing-field helpers still descend into `Middle.Base` and diagnose its omitted required/reference fields. The direct, explicitly supplied `Outer.Base` continues to be checked through its own initializer.

The added `struct_embed_shadowed_required_field_err.vv`/`.out` fixture covers the reported omission for both required and reference fields. The positive `test_same_named_embeds_keep_distinct_initializers()` case also checks that separately initialized direct and nested embeds retain distinct values. Ordinary promoted-field names remain forwarded, preserving the existing `struct_embed_required_field_err` behavior, and the nested-literal fixture retains the single diagnostic for a genuinely missing required field.

No additional actionable findings in this source-level re-review of the updated diff.

Validation limitation: I reviewed the implementation and test sources, but did not execute the V compiler or tests. No V compiler was available locally, and a fresh attempt to access the repository with Git failed because the environment could not resolve `github.com`.

Issue #28738, opened 2026-09-18
cgen: top-level `asm` block is not emitted by V3 (undefined reference to its label)


quaesitor-scientiam:
## Describe the bug

A top-level (file-scope) `asm` block is not emitted by the V3 C backend, so a symbol it defines is an undefined reference at link time. V1 emitted such blocks as a file-scope `__asm__`; `vlib/v/slow_tests/assembly/asm_test.amd64.v:171-176` relies on this (`test_rip_relative_label_u8`) and currently only compiles through the V1 fallback.

## Reproduction Steps

```v
fn main() {
	mut a := int(0)
	asm amd64 {
		mov a, [rip + word_sequence]
		; =r (a)
	}
	assert a == 0x480f3527
	println('ok')
}

asm amd64 {
	.global word_sequence
	word_sequence:
	.long 0x480f3527
}
```

```
v -cc gcc run toplevel_asm.v
```

In the generated C (`v -o out.c toplevel_asm.v`) the only occurrence of `word_sequence` is the in-function reference `"mov word_sequence(%%rip), %[a]\n\t"`; the top-level block is absent.

## Expected Behavior

`ok` — the top-level block is emitted at file scope and `word_sequence` resolves.

## Current Behavior

```
ld.exe: src.c:(.text+0x8c0d): undefined reference to `word_sequence'
collect2.exe: error: ld returned 1 exit status
```

## Additional Information/Context

Found while checking why `vlib/v/slow_tests/assembly/asm_test.amd64.v` fails under V3 (it passes in CI only via the V 0.5.2 fallback). The original test uses `.byte` in that block, which trips a second, separate problem: the checker reports the directive as the deprecated `byte` type (filed separately). Related: #28736.

## V version

V 0.5.3 01a8af9 (master, 2026-09-18)

## Environment details

Windows 11 amd64, `-cc gcc` (mingw64 16.2); V3 compiler.

PR #28764, opened 2026-09-19
checker: do not report `.byte` inside inline assembly as the deprecated `byte` type (fix #28739)

### LINKED ISSUE #28739: checker: `.byte` directive inside an `asm` block is reported as the deprecated `byte` type
## Describe the bug

Inside an `asm` block, the assembler directive `.byte` is reported as the deprecated V type `byte`:

```
byte_directive.v:14:3: error: byte is deprecated, use u8 instead
```

The diagnostic comes from `vlib/v/types/checker.v` (`'byte is deprecated, use u8 instead'`), i.e. the checker treats the directive name after `.` in an asm block as a type identifier. V1 accepted this; `vlib/v/slow_tests/assembly/asm_test.amd64.v:174` uses exactly this directive and currently only compiles through the V1 fallback.

## Reproduction Steps

```v
fn main() {
	mut a := int(0)
	asm amd64 {
		mov a, [rip + byte_sequence]
		; =r (a)
	}
	assert a == 0x480f3527
	println('ok')
}

asm amd64 {
	.global byte_sequence
	byte_sequence:
	.byte 0x27, 0x35, 0x0f, 0x48
}
```

```
v run byte_directive.v
```

The same block with `.long 0x480f3527` instead of `.byte ...` gets past the checker (and then hits #28738, the top-level asm block not being emitted).

## Expected Behavior

`.byte` is an assembler directive (like `.global`, `.long`, `.quad`); the deprecated-type check must not apply to tokens inside an `asm` block.

## Current Behavior

```
byte_directive.v:14:3: error: byte is deprecated, use u8 instead
   12 |     .global byte_sequence
   13 |     byte_sequence:
   14 |     .byte 0x27, 0x35, 0x0f, 0x48
      |      ~~~~
```

## Possible Solution

Skip the deprecated-`byte` type check for identifiers that are part of an asm block (directive names, labels, register names).

## Additional Information/Context

Found while checking why `vlib/v/slow_tests/assembly/asm_test.amd64.v` fails under V3 (it passes in CI only via the V 0.5.2 fallback). Related: #28736 (arch-qualified test files), #28738 (top-level asm block dropped).

## V version

V 0.5.3 01a8af9 (master, 2026-09-18)

## Environment details

Windows 11 amd64, `-cc gcc` (mingw64 16.2); V3 compiler.

quaesitor-scientiam:
Fixes #28739

The deprecated-`byte` check rescans the raw source text of each file, so the assembler directive `.byte` (or an Intel-syntax `byte ptr` operand) inside an `asm` block was reported as the V type. `vlib/v/slow_tests/assembly/asm_test.amd64.v:174` and the `.byte` snippet in `vlib/v/compiler_tests/inline_asm_c_lowering_test.v` both hit this under V3.

Fix: collect the span of every `asm_stmt` node per file and skip `byte` tokens inside one. Text inside string literals was already skipped, so raw templates were unaffected; this covers V-syntax and Intel-syntax blocks.

Tests:
- `vlib/v/tests/inline_asm_byte_directive_test.v` — `.byte 0x90` in an `asm amd64` block; fails on master with the error above, passes with the fix (the `$if amd64` branch is parsed and text-scanned on every arch, so the checker path is exercised everywhere).
- `vlib/v/checker/tests/use_byte_instead_of_u8_err.vv` still reports all three `byte` type uses with a compiler built from this branch.

Known limitation: a V expression in an asm operand list (e.g. `; =r (byte(x))`) is inside the span and no longer reported; that spelling is deprecated regardless and I found no instance in the tree.

The #28739 repro now passes the checker and fails at link on #28738 (top-level asm block not emitted), which is separate.


JalonSolov:
Local AI Findings:

**P2** `checker.v:6170`: the new `offset_in_asm_span` check suppresses every byte occurrence inside an `asm_stmt`, including deprecated V expressions in operand lists such as `byte(x)`. That weakens the existing byte deprecation diagnostic. Restrict the exemption to raw assembler text, excluding spans belonging to parsed asm operands.

The added test covers `.byte` but not this operand-expression case.

quaesitor-scientiam:
Addressed in 193f9836ae: the exemption now subtracts the spans of the block's operand expressions (the `asm_stmt` child nodes), so assembler text is exempt but a V expression in the operand list is still checked.

Added `vlib/v/checker/tests/use_byte_instead_of_u8_in_asm_operand_err.vv` with both a `.byte 0x90` directive and a `; r (byte(x)) as y` operand; only line 8 (the operand) is reported.

One thing I noticed while producing the `.out`: with a `-prealloc` compiler (what `v self` builds) and `-no-parallel` (what the fixture harness passes), the checker takes `check_semantics_scoped_serial`, which does not run `check_deprecated_byte_types` at all — the three existing `use_byte_instead_of_u8_*_err.vv` fixtures produce no error in that mode on master either. That is a separate, pre-existing gap; I can file it or fold a fix in here if you prefer.

JalonSolov:
Probably better to do the other fixes in this PR.  The current changes are small... it shouldn't add much to extend them.

quaesitor-scientiam:
> Probably better to do the other fixes in this PR. The current changes are small... it shouldn't add much to extend them.

Opened separate issue #28792  and creased PR #28793 28793

quaesitor-scientiam:
Rebased onto 15aa54185c (aa4921641e). The only conflict was the skip condition in `check_deprecated_byte_types_in_file`: master added `tc.deprecated_byte_is_value_ident(file_id, start)` there; it is kept, and the assembler-span exemption is added after it — the two are independent.

Still needed on master: with the PR's test and fixture copied into 15aa54185c, `inline_asm_byte_directive_test.v` fails with `7:5: error: byte is deprecated, use u8 instead` and the operand fixture reports the `.byte` directive. After the rebase: the test passes on Linux x86_64 and Windows amd64, and `use_byte_instead_of_u8_in_asm_operand_err`, `use_byte_instead_of_u8_err`, `use_byte_instead_of_u8_array_param_err`, `use_byte_instead_of_u8_array_type_err` and `asm_immutable_err` match their `.out` under the harness flags on both.

PR #28768, opened 2026-09-19
checker: allow entry writes to cloned maps with nested values

medvednikov:
## Summary

The compiler integration is now committed in `c2c6ef918258c6321cf04ee8b547ba705f6fb6c3`. This is no longer an unused-helper draft: `check_mutable_alias_assignment_lhs` now calls the new map-storage proof.

Permit insertion/replacement of whole entries in a proven builtin `map.clone()` result while preserving immutable-reference metadata for shared nested data. No UI2 workaround and no blanket alias-check exemption are added.

## Regression

With `V 0.5.2 0a006c2`, UI2's users example rejects:

```v
for index, item in items.items {
    mut item_scope := scope.clone()
    item_scope['item'] = item
    item_scope['index'] = v_number(f64(index), index.str())
}
```

Source-history comparison traced the behavior change to #28589, commit `4f915cc1ffe4d2cae6b2b39bde907c3ab2c1ac0d` (September 14, 2026). Its parent had an unconditional builtin-map-clone alias exemption. The replacement checks nested sharing as well, so a map of recursive `VValue` structs retains alias metadata. The assignment checker then incorrectly treats an entry write into the independent clone as a write through shared storage.

## Implementation

In `vlib/v/types/checker_comptime.v`:

```diff
-        if aliases {
+        if aliases && !tc.mutable_alias_has_fresh_map_storage(base_id) {
```

The new helper only proves independent top-level map storage. For local bindings it conservatively declines the exception for ambiguous declarations, whole-map reassignment (including a later loop iteration), mutable-argument escapes, address-taking, receiver/method escapes, closures, and uncertain control flow. It does not clear `immutable_reference_aliases` or exempt user-defined clone methods.

## Regression coverage added

- UI2-style recursive-value map cloned inside a loop; insertion and replacement; original map remains unchanged.
- Empty source maps and map type aliases.
- Negative compiler cases for nested mutation, reassignment, later-loop reassignment, mutable-argument escape, and a user-defined clone returning borrowed storage.

## Validation — executable results pending

Verified the final GitHub PR diff: four compiler/test files, including the actual checker hook. The existing checker file differs by exactly one added and one removed line. The temporary integration workflow is absent from the final diff, and temporary merge-scaffold history is not included in this branch.

**No compiler build, V formatter, V tests, or end-to-end UI2 run have completed in this session.** The local environment has no V toolchain or working direct repository download; the GitHub validation job remained queued without executing any steps. Repository write access is working; it is not the blocker.

Validation commands:

```sh
./v self
./v fmt -verify vlib/v/types/checker_map_clone_alias.v \
  vlib/v/tests/map_clone_struct_scope_test.v \
  vlib/v/compiler_tests/map_clone_alias_regression_test.v
./v -silent test vlib/v/tests/map_clone_struct_scope_test.v
./v -silent test vlib/v/compiler_tests/map_clone_alias_regression_test.v
./v -silent test vlib/v/compiler_tests/fresh_collection_builtin_alias_test.v
./v -silent test vlib/v/types/
./v -silent vlib/v/compiler_errors_test.v
```

Also validate the unmodified upstream `vlang/ui2` users example with the rebuilt compiler before merging.


medvednikov:
The regression is traced at source level to 4f915cc1ffe4d2cae6b2b39bde907c3ab2c1ac0d (#28589): its parent explicitly exempts builtin map clones, while the replacement exemption requires the copied values to contain no shareable storage. UI2's recursive `VValue` does not satisfy that condition, even when the subsequent write only replaces an entry in independent outer-map storage.

The current head, 9c1ca9accb5610fbc6960dd969ccb374b0d3e23e, already contains the conservative storage proof and positive/negative regression tests. The following integration patch completes the intended source change on that head:

```diff
diff --git a/vlib/v/types/checker_comptime.v b/vlib/v/types/checker_comptime.v
--- a/vlib/v/types/checker_comptime.v
+++ b/vlib/v/types/checker_comptime.v
@@ -13443,7 +13443,7 @@ fn (mut tc TypeChecker) check_mutable_alias_assignment_lhs(id flat.NodeId, rhs_id fl
 				aliases = aliases || tc.fn_context.immutable_reference_aliases[root.value]
 			}
 		}
-		if aliases {
+		if aliases && !tc.mutable_alias_has_fresh_map_storage(base_id) {
 			tc.record_error_at(.assignment_mismatch, '`${tc.source_text_for_node(base_id)}` aliases mutable data from an immutable value, clone it first (or use `unsafe`)', base_id, if base.kind in [
 				.ident,
 				.selector,
```

**This patch has not been committed to the branch.** A repository staging write was blocked by the editing tool, and direct repository downloads are unavailable in this session. I have prepared both an integration-only patch and a complete four-file patch, and verified the three copied helper/test files against their GitHub blob SHAs. No V build, V formatter, regression test, or UI2 example validation has run. The PR remains an incomplete draft; do not treat this comment as a passing validation result.

PR #28770, opened 2026-09-19
net.http: drop TE from HTTP/2 server response headers and trailers

quaesitor-scientiam:
RFC 9113 §8.2.2 permits `TE` only in requests (and only as `trailers`); a response carrying it in either its header or trailer section is malformed. `H2ServerConn.send_response` and `h2_outbound_trailer_fields` filtered only `h2_conn_specific_headers`, which deliberately omits `TE` (request-side validation allows it), so a handler-authored `TE` was forwarded unchanged.

Both filters now share `h2_response_field_is_forbidden` (= the connection-specific list + `te`). The const's doc comment no longer claims TE "is handled on the send side" — nothing did, and that claim is what let this slip past review twice.

This is the HTTP/2 sibling of the HTTP/3 server finding in #28164 (pullrequestreview-5246850132), fixed there in `e041801019`; noted as "separate PR" in that thread.

## Test plan
- [x] New end-to-end test `test_h2_server_drops_te_from_response_headers_and_trailers` (handler sets `TE` in both sections, with a body so trailers go in their own HEADERS block) **fails on master**, passes with the fix
- [x] `h2_server_test.v`, `h2_mux_conn_test.v`, `h2_conn_test.v` green
- [x] `v fmt -verify` clean on touched files

Not changed: client-side *reception* of a `TE` response field (`h2_response_field_error`) — rejecting it would be RFC-correct but is a strictness/interop policy call, left for a separate discussion.


medvednikov (commented):
Reviewed at `5a426217cdae7599d935c613d2f6e067832601a5`.

No actionable correctness issue found in the three-file diff. Both response paths lowercase names before calling the new shared predicate, so standard `TE` and custom casing are covered. Keeping `te` out of `h2_conn_specific_headers` preserves the separate request-side exception, while response headers and trailers now consistently drop it. The existing trailer pseudo-header/value checks remain intact.

I also checked the empty-body/trailers-only branch: it consumes the already-filtered `trailer_fields`, so it cannot bypass the new filter. The regression sends a body and retains an ordinary header plus an ordinary trailer, which checks that filtering does not simply discard either entire section.

Validation: complete diff and surrounding request-validation/response-emission source review. I did not execute the V tests because no V compiler is available locally. CI status and results were not considered.

PR #28791, opened 2026-09-19
tests: cover generic middleware handler aliases

medvednikov:
## Context

Reported failure:

```v
pub type MiddlewareHandler[T] = fn (mut T) bool
// error: generic type aliases are not yet implemented
```

At the base commit (`2b99da47ea7268c152d8b90aae18517b39c1f788`), the current parser already accepts generic parameters on aliases and stores them in the type declaration. Existing coverage includes `vlib/v/tests/generics/generic_alias_callback_test.v` and imported generic function aliases in `vlib/v/compiler_tests/fn_value_decl_type_test.v`.

The reported diagnostic is consistent with an older compiler executable being used against newer vlib sources. This PR does **not** claim to add a missing compiler implementation or to repair an older executable; it adds focused regression coverage for the exact middleware use case against current master.

## Changes

- Add `vlib/v/parser/generic_type_alias_test.v`: parse generic pointer aliases, the public `MiddlewareHandler[T] = fn (mut T) bool` declaration, a two-parameter function alias, and a following non-generic alias. Check generic metadata and public visibility.
- Remove `vlib/v/parser/tests/generic_type_alias_decl.vv` and `.out`: this old negative fixture expects all generic aliases to be rejected, contradicting the current parser and existing positive generic-alias tests. Its pointer declaration is retained in the new positive parser test.
- Add `vlib/veb/middleware_handler_alias_test.v`: exercise the actual middleware alias with two distinct mutable contexts, function-pointer round trips through `voidptr`, empty chains, short-circuiting, global before/after registration, and route method filtering. These tests do not start a server.

No production compiler or middleware implementation changes, public API changes, or unrelated edits.

## Validation

- Passed: `git diff --check` on the locally generated patch.
- **Not run:** V compilation, runtime tests, and `v fmt`. The execution environment has no V executable; cloning/bootstrap was blocked because the shell could not resolve `github.com`. Repository reads and writes were performed through the connected GitHub API.

Suggested validation from a freshly rebuilt compiler:

```sh
./v self
./v fmt -verify vlib/v/parser/generic_type_alias_test.v vlib/veb/middleware_handler_alias_test.v
./v -silent test vlib/v/parser/generic_type_alias_test.v
./v -silent test vlib/veb/middleware_handler_alias_test.v
./v -silent test vlib/v/tests/generics/generic_alias_callback_test.v
```

Opening as a draft until compiler/formatter validation is available.


medvednikov (commented):
Reviewed at `b34570f6a9241515a3ca257ae7f32bb410e86fad`.

No actionable correctness issue found in this tests-only diff. The middleware tests call the production `validate_middleware`, registration, and method-filtering helpers rather than recreating their behavior. Distinct mutable context layouts and explicit state assertions make the alias/cast tests meaningful; the stopping handler also verifies that a later callback is not invoked. The registration cases check both before and after collections without starting a server.

The replacement parser test retains the deleted fixture's `Pointer[T] = &T` declaration as a positive case and adds generic-parameter ordering, public visibility, and a following non-generic declaration to catch metadata leakage. The existing `generic_alias_callback_test.v` also already expresses generic function aliases as supported behavior, so removing the blanket-rejection expectation is consistent with that coverage.

Validation: complete four-file diff, the production middleware helpers, and the existing generic callback regression were inspected. I did not run the parser or runtime tests because no V compiler is available locally; this is not a claim that the new tests compile or pass. CI status and results were not considered.

